PR #3658, opened 2021-03-30
mutual TLS for the gRPC API in ws-manager

corneliusludmann:
to do:
- [x] ee/ws-scheduler: wsmanAddr
- [x] remove `localhost` from altNames of certs and use `ServerName` in gpctl client instead
- [ ] should we add a flag (or better: secretName value) in values.yaml to make this optional in the chart templates?

more ToDos:
 - [x] cache wsman clients in `ClientProvider` for a short period (~30s)
 - [ ] make sure we're not materializing `WorkspaceCluster.tls` too often as it is now ~4KB in size
 
 
 ### Test
 1. disable all static bridge config
    1. `kubectl edit cm ws-manager-bridge-config`, set `staticBridges` to `[]`, `kubectl rollout restart deployment ws-manager-bridge`
    2.  `kubectl edit deployment server`, set `WSMAN_CFG_MANAGERS` to `W10K`
 2. `cd dev/gpctl && go build -v .`
 3. `./gpctl clusters get-tls-config`
 4. `./gpctl clusters list`
 5. `./gpctl clusters register --name dev --url ws-manager:8080 --hint-govern --hint-preferability prefer --tls-path=./wsman-tls`
 6. start workspace


.werft/values.dev.yaml
```diff
@@ -37,8 +37,16 @@ components:
     wsman:
     - name: "local"
       address: "ws-manager:8080"
+      tls:
```
geropl, line 40: IMO configuring the root certificate here would be sufficient, cmp. https://grpc.io/docs/guides/auth/#with-server-authentication-ssltls-6. That would also fit into the current `gpctl`/DB implementation where we have one `certificate` field.

chart/templates/registry-facade-configmap.yaml
```diff
@@ -36,7 +36,12 @@ data:
             {{ end -}}
             {{ end -}}
             "remoteSpecProvider": {
-                "addr": "ws-manager:{{ .Values.components.wsManager.ports.rpc.containerPort }}"
+                "addr": "ws-manager:{{ .Values.components.wsManager.ports.rpc.containerPort }}",
+                "tls": {
```
geropl, line 40: same here (and all following client examples, basically).

.werft/values.dev.yaml
corneliusludmann: Correct me if I'm wrong but I think this is not true for mutual TLS because the root CA is just for checking the server cert but with mutual TLS the client needs a cert as well so that the server can check the client cert. (we have 2 certs here: one for server, one for client, both signed with the same CA)

See e.g. here for a brief example: https://krishicks.com/post/grpc-mutual-tls-golang/

components/ws-manager/cmd/run.go
```diff
@@ -133,13 +136,32 @@ var runCmd = &cobra.Command{
...
+			b, err := ioutil.ReadFile(cfg.RPCServer.TLS.CA)
+			if err != nil {
+				log.WithError(err).WithField("ca", cfg.RPCServer.TLS.CA).Fatal("could not load CA")
+			}
+			if !certPool.AppendCertsFromPEM(b) {
+				log.WithError(err).WithField("ca", cfg.RPCServer.TLS.CA).Fatal("failed to append CA")
+			}
+
+			creds := credentials.NewTLS(&tls.Config{
+				ClientAuth:   tls.RequireAndVerifyClientCert,
```
geropl, line 154: This reads like client authentication, which is also nice, but involves more certifactes/keys to be shoveled around. I hoped could get away with a `token`-based client auth for now.
Server authentication is a tad simpler and explained: https://grpc.io/docs/guides/auth/#with-server-authentication-ssltls

corneliusludmann: See my comment here: https://github.com/gitpod-io/gitpod/pull/3658#discussion_r603903638

I thought we would like to go with mutual TLS. From the design doc:

> mutual trust: we need to ensure that the ws-manager that's trying to register is one of ours, and not some impersonator. At the same time, our ws-manager must be sure that the ws-manager-bridge they're talking to is one of ours.
> We would use mutual TLS (mTLS) on the registry API interface of ws-manager-bridge to ensure mutual trust. gRPC supports mTLS.

geropl:
Nice, actually a lot more work than I expected! :bow: 
Let's sort out the client-auth/server-auth stuff and head for mergin :upside_down_face:

.werft/values.dev.yaml
geropl: Absolutely true, just thought we wanted server auth + token. After offline discussion we're going with mutual TLS.

PR #3661, opened 2021-03-30
[proxy] Improve nginx Time To First Byte - TTFB

aledbf:
**Changes:**

- Backlog Queue setting `net.core.somaxconn` from `128` to `32768`
- Ephemeral Ports setting `net.ipv4.ip_local_port_range` from `32768 60999` to `1024 65000`

In a [post from the NGINX blog](https://www.nginx.com/blog/tuning-nginx/), it is possible to see an explanation for the changes.


csweichel (approved):
LGTM

PR #3666, opened 2021-03-30
[server] added caching for open-vsx queries

svenefftinge:
We send many request to open-vsx which it cannot handle. This PR is an attempt to reduce the number of request by caching resolutions for 1h. 
Caching this is not totally cool, because people might publish something on open-vsx and add it to their gitpod.yml. When starting another workspace to test the installation, it might not find the new extension, because it returns the outdated cached answer.


akosyakov:
It did not make any different. I tried to restart several times the same workspace, but it looks that the server simply cannot cache anything, i.e. all requests times out. We could remove timeout but it is likely to make the workspace startup slow.

Issue #3630, opened 2021-03-28
 proto: file "workspace.proto" is already registered


akosyakov:
### Bug description

Supervisor logs full of:
```
2021/03/25 19:48:50 WARNING: proto: file "workspace.proto" is already registered
A future release will panic on registration conflicts. See:
https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict
```

### Steps to reproduce

start any workspace and inspect logs

PR #3669, opened 2021-03-30
Add leeway script to update protobuf generated code

aledbf:
Adds a new leeways script to run the `generate.sh` script: `leeway run components:generate-code-from-protobuf`

fixes https://github.com/gitpod-io/gitpod/issues/3630

*Issue:*
```
2021/03/30 20:54:39 WARNING: proto: file "workspace.proto" is already registered
	previously from: "github.com/gitpod-io/gitpod/content-service/api"
	currently from:  "github.com/gitpod-io/gitpod/ws-daemon/api"
A future release will panic on registration conflicts. See:
https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict
``` 

Also, implement change https://github.com/grpc/grpc-go/issues/3669


csweichel:
> Also removes pending manual code from .pb.go files into an explicit file not_autogenerated.XXXXXX.pb.go.

I don't understand this change. All of the `.pb.go` files should contain only auto-generated code. If those files don't line up with the code generation in this PR (or even on main) something fundamentally changed in the code generation and we should investigate if those files are still required.

aledbf:
> I don't understand this change. All of the `.pb.go` files should contain only auto-generated code. If those files don't line up with the code generation in this PR (or even on main) something fundamentally changed in the code generation and we should investigate if those files are still required.

Fixed. The `generate.sh` was missing the grpc plugin (that generates the grpc client)

components/content-service/pkg/service/blob-service.go
```diff
@@ -22,6 +22,8 @@ import (
 type BlobService struct {
 	cfg storage.Config
 	s   storage.PresignedAccess
+
+	api.UnimplementedBlobServiceServer
```
csweichel, line 26: I understand this change is necessary due to some deliberately non-backwards compatible, idiosyncratic design decisions (https://github.com/grpc/grpc-go/issues/3669#issuecomment-695022942).

I'm am however not convinced the unimplemented default implementation is preferable to things breaking at compile time. If someone modifies the interface, they should be made aware of the consequences of that action as early as possible.

We can't disable this behaviour though, so we'll have to live with it.

csweichel (approved):
Thank you for cleaning this up.

PR #3676, opened 2021-03-31
[chart] Fix messagebus secret name

aledbf:


jankeromnes (approved):
Thanks! Build passes + I've briefly smoke tested the deployment + verbal confirmation that message bus is okay. 🚀

PR #3677, opened 2021-03-31
Add and use OpenVSX caching proxy

corneliusludmann:
This PR introduces a caching proxy for OpenVSX at https://open-vsx.gitpod.io.

**Here are some parameters that are open for discussion:**
- [ ] With the current implementation, every GET, HEAD, and PUSH call is cashed.
- [ ] Every response is cached for 30 minutes (valid time).
- [ ] A cached response is removed when not accessed for 120 minutes.
- [ ] In case of an error (timeout http_500 http_502 http_503 http_504) a stale response in the cache (of the past 120 minutes) is served.
- [ ] Cache folder has max. size of 10 GB.
- [ ] Header from upstream like Cache-Control are ignored (everything is cached for 30 minutes).
- [ ] For VSCode the OpenVSX URL is changed in the leeway.Dockerfile, Theia uses the TheiaPluginService, right? Or does Theia need a URL change as well?

**How to test:**
- Open a workspace
- Use the extensions view to search for extensions
- Observe the proxy logs: `kubectl logs -f deployment/proxy | grep proxying`, every `"cacheStatus":"HIT"` is a cached result

### DIFF
--- a/.werft/build.js
+++ b/.werft/build.js
@@ -82,6 +82,11 @@ async function build(context, version) {
         withIntegrationTests,
     }));
 
+    const destname = version.split(".")[0];
+    const namespace = `staging-${destname}`;
+    const domain = `${destname}.staging.gitpod-dev.com`;
+    const url = `https://${domain}`;
+
     /**
      * Build
      */
@@ -102,7 +107,7 @@ async function build(context, version) {
     if (withInstaller || publishRelease) {
         exec(`leeway build --werft=true -c ${cacheLevel} ${dontTest ? '--dont-test':''} -Dversion=${version} -DimageRepoBase=${imageRepo} install:all`, buildEnv);
     }
-    exec(`leeway build --werft=true -Dversion=${version} -DremoveSources=false -DimageRepoBase=${imageRepo}`, buildEnv);
+    exec(`leeway build --werft=true -Dversion=${version} -DgitpodDomain=${domain} -DremoveSources=false -DimageRepoBase=${imageRepo}`, buildEnv);
     if (publishRelease) {
         try {
             werft.phase("publish", "checking version semver compliance...");
@@ -156,11 +161,7 @@ async function build(context, version) {
 
         return
     }
-    
-    const destname = version.split(".")[0];
-    const namespace = `staging-${destname}`;
-    const domain = `${destname}.staging.gitpod-dev.com`;
-    const url = `https://${domain}`;
+
     const deploymentConfig = {
       version,
       destname,
--- a/components/ide/code/BUILD.yaml
+++ b/components/ide/code/BUILD.yaml
@@ -7,7 +7,10 @@ packages:
       - "bin/*"
     argdeps:
       - imageRepoBase
+      - gitpodDomain
     config:
       dockerfile: leeway.Dockerfile
       image:
         - ${imageRepoBase}/ide/code:${version}
+      buildArgs:
+        OPENVSX_URL: https://open-vsx.${gitpodDomain}
--- a/components/server/src/theia-plugin/theia-plugin-service.ts
+++ b/components/server/src/theia-plugin/theia-plugin-service.ts
@@ -22,6 +22,7 @@ import * as request from 'request';
 const builtinExtensions: PluginIndexEntry[] = require('@gitpod/gitpod-protocol/data/builtin-theia-plugins.json');
 
 const userPluginsUri = 'user-plugins://';
+const localOpenVSXDomainPrefix = 'open-vsx.';
 
 export interface ResolvedPluginsResult {
     resolved: ResolvedPlugins
@@ -35,6 +36,11 @@ export class TheiaPluginService {
     @inject(StorageClient) protected readonly storageClient: StorageClient;
     @inject(TheiaPluginDB) protected readonly pluginDB: TheiaPluginDB;
     @inject(UserStorageResourcesDB) protected readonly userStorageResourcesDB: UserStorageResourcesDB;
+    protected readonly openVSXCachingProxyURL: string;
+
+    public constructor() {
+        this.openVSXCachingProxyURL = new GitpodHostUrl(process.env.HOST_URL).withDomainPrefix(localOpenVSXDomainPrefix).toString();
+    }
 
     /**
      * @returns a sanitized path to the plugin archive which can be used in signed URLs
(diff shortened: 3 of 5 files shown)


aledbf:
@corneliusludmann this should be allowed only from workspaces running in `.gitpod.io` (not an open proxy)

corneliusludmann:
> @corneliusludmann this should be allowed only from workspaces running in `.gitpod.io` (not an open proxy)

@aledbf: Good idea. Do you have a hint for me on how to do this?

svenefftinge:
Also, I remember that @akosyakov mentioned there were CORS issues with accessing open-vsx.org from within frontends loaded from *.gitpod.io. A (public) proxy could solve that issue.

akosyakov:
@svenefftinge CORS issues were sorted out on Open VSX side already: https://github.com/eclipse/openvsx/issues/268#issuecomment-801747761

components/ide/code/BUILD.yaml
```diff
@@ -7,7 +7,10 @@ packages:
       - "bin/*"
     argdeps:
       - imageRepoBase
+      - gitpodDomain
     config:
       dockerfile: leeway.Dockerfile
       image:
         - ${imageRepoBase}/ide/code:${version}
+      buildArgs:
+        OPENVSX_URL: https://open-vsx.${gitpodDomain}
```
csweichel, line 16: if we made this an explicit variable, we could set a workspace default and only produce a build when absolutely necessary. 

```suggestion
        OPENVSX_URL: ${OPENVSX_URL}
```

components/server/src/theia-plugin/theia-plugin-service.ts
```diff
@@ -35,6 +36,11 @@ export class TheiaPluginService {
     @inject(StorageClient) protected readonly storageClient: StorageClient;
     @inject(TheiaPluginDB) protected readonly pluginDB: TheiaPluginDB;
     @inject(UserStorageResourcesDB) protected readonly userStorageResourcesDB: UserStorageResourcesDB;
+    protected readonly openVSXCachingProxyURL: string;
+
+    public constructor() {
+        this.openVSXCachingProxyURL = new GitpodHostUrl(process.env.HOST_URL).withDomainPrefix(localOpenVSXDomainPrefix).toString();
+    }
```
csweichel, line 43: This bit I find most interesting. OpenVSX blocking workspace starts is a no-go.
@akosyakov with VSCode resolving its plugins itself, other than for Theia, do we even still need this?

csweichel:
> @corneliusludmann this should be allowed only from workspaces running in `.gitpod.io` (not an open proxy)

Unfortunately this would break self-hosted, unless we introduce all this variance in a "SaaS" only build - or at least produce a VS Code build for Self-Hosted that does not use `openvsx.gitpod.io`.

.werft/build.js
```diff
@@ -102,7 +107,7 @@ async function build(context, version) {
     if (withInstaller || publishRelease) {
         exec(`leeway build --werft=true -c ${cacheLevel} ${dontTest ? '--dont-test':''} -Dversion=${version} -DimageRepoBase=${imageRepo} install:all`, buildEnv);
     }
-    exec(`leeway build --werft=true -Dversion=${version} -DremoveSources=false -DimageRepoBase=${imageRepo}`, buildEnv);
+    exec(`leeway build --werft=true -Dversion=${version} -DgitpodDomain=${domain} -DremoveSources=false -DimageRepoBase=${imageRepo}`, buildEnv);
```
meysholdt, line 110: Can you make the Domain Name a parameter or env var of the IDE image? 
Introducing it as built tim variance is problematic because the build artefacts should be installable on any domain.

meysholdt (changes requested):

.werft/build.js
corneliusludmann: I totally agree. I have made some steps in that direction here: https://github.com/gitpod-io/gitpod/tree/clu/openvsx-proxy-2
However, that doesn't work yet that well.

aledbf:
> Unfortunately this would break self-hosted, 

Why? The default value should be not use a proxy.

> unless we introduce all this variance in a "SaaS" only build - or at least produce a VS Code build for Self-Hosted that does not use `openvsx.gitpod.io`.

`openvsx.gitpod.io` makes sense only for SaaS (unless we want to be in charge of maintaining a self-hosted component running outside the control of gitpod.io)

aledbf:
What's the impact of running an openvsx proxy? (resources like traffic/cpu)

corneliusludmann:
> > Unfortunately this would break self-hosted,
> 
> Why? The default value should be not use a proxy.

When the OpenVSX URL is set during runtime (see https://github.com/gitpod-io/gitpod/pull/3677#discussion_r605601092) a self-hosted installation could use its own proxy for OpenVSX, not https://open-vsx.gitpod.io but https://open-vsx.my-gitpod-installation.com.


> > unless we introduce all this variance in a "SaaS" only build - or at least produce a VS Code build for Self-Hosted that does not use `openvsx.gitpod.io`.
> 
> `openvsx.gitpod.io` makes sense only for SaaS (unless we want to be in charge of maintaining a self-hosted component running outside the control of gitpod.io)

I don't know if it's worth to artificially restrict the access to open-vsx.gitpod.io to our components unless we find an easy way to do this. However, I consider the risk of abuse to be relatively low.

roboquat:

This pull-request has been approved by: *<a href="https://github.com/gitpod-io/gitpod/pull/3677#" title="Author self-approved">corneliusludmann</a>*

<details open>

- ~~[OWNERS](https://github.com/gitpod-io/gitpod/blob/main/OWNERS)~~ [corneliusludmann]
- **[components/ide/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/ide/OWNERS)**
- ~~[components/server/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/server/OWNERS)~~ [corneliusludmann]

</details>

svenefftinge:
I wonder if we can use this approach and configure it so it can bridge longer unavailability of onen-vsx.org.
The data shouldn't be too big (it's only metadata), so we could maybe cache requests for a couple of weeks instead of 30 minutes and only use that cache if the upstream registry doesn't respond in a certain time.

csweichel:
Blobserve can replace strings when serving content to the frontend. Instead of backing in the open-vsx domain at build-time, we should make that domain configurable by virtue of blobserve.

This way we avoid the self-hosted interaction, and can enable a custom OpenVSX proxy in installations where required.

csweichel:
This almost strikes me as new PR, also considering that we're no longer using nginx as proxy.

PR #3680, opened 2021-03-31
[dashboard] route login to /

svenefftinge:


components/dashboard/src/App.tsx
```diff
@@ -70,7 +70,7 @@ function App() {
         <div className="container">
             {renderMenu()}
             <Switch>
-                <Route path={["/", "/workspaces"]} exact render={
+                <Route path={["/", "/workspaces", "/login"]} exact render={
```
gtsiolis, line 73: Works ok but could we also update the URL path to `/` or `/workspaces`? Now both anonymous and logged in users now see the workspaces list (after logging in) but see the `/login` path in the URL. Feel free to merge and open a follow up issue fo this.

:M: :V: :C:

gtsiolis (approved):

components/dashboard/src/App.tsx
svenefftinge: Good point! I pushed again.

components/dashboard/src/App.tsx
```diff
@@ -46,7 +46,7 @@ function App() {
             }
             setLoading(false);
         })();
-    }, []);
+    }, [ setUser ]);
```
AlexTugarev, line 49: @svenefftinge we need to change this :-(

this effect causes constant stream of `getLoggedInUser` requests 

Screen Shot 2021-04-01 at 05 52 56

PR #3699, opened 2021-04-01
multi origin PWA

akosyakov:


components/dashboard/src/index.tsx
```diff
@@ -18,4 +18,11 @@ ReactDOM.render(
         </UserContextProvider>
     </React.StrictMode>,
     document.getElementById('root')
-);
\ No newline at end of file
+);
+
+if (process.env.NODE_ENV === 'production' && 'serviceWorker' in navigator) {
```
svenefftinge, line 23: this is not running in node

akosyakov: It is inlined during the build: https://create-react-app.dev/docs/adding-custom-environment-variables

akosyakov:
I will close it for now. It works but closes new apps for each workspace. We will need to wait for multi origin system opener.

PR #3703, opened 2021-04-01
[ide] fixed link to upgrade subscription page

JanKoehnlein:


akosyakov (approved):
LGTM

but we still need redirects, since clients don't use newer version of gitpod-protocol

PR #3709, opened 2021-04-01
[code] install without sync from config

akosyakov:
#### What it does

It fixes a regression when extension from .gitpod.yml installed for a user, not only for a given workspace.

Changes in VS Code: https://github.com/gitpod-io/vscode/commit/0ce76acf042b084a6eb0f3c76ece0f3db2562d1c

#### How to test

- Start a workspace for https://github.com/gitpod-io/symfony-demo.
- Check that php extensions are installed but without sync.


JanKoehnlein (approved):
LGTM

PR #3722, opened 2021-04-01
[ws-daemon] Preserve file extended attributes

### LINKED ISSUE #3721: April upcoming gitpod:  cap_net_bind_service lost from docker image on restore
### Bug description

When an image is pulled that has `setcap cap_net_bind_service`, it works correctly on first use. Unfortunately, if the instance is stopped and then restarted, the `cap_net_bind_service` is lost. The same thing would result from a prebuild followed by launching an instance.

### Steps to reproduce

I prepared a repo to demonstrate this, https://github.com/rfay/cap-add-demo - it has a README with full instructions.

Basically:
1. In a gitpod workspace,  `docker-up` and pull a docker image that has cap_net_bind_service set (I'm sure this is any setcap)
2. Verify that the setcap is functional using getcap
3. Stop the gitpod workspace
4. Start the gitpod workspace. You'll find that the cap_net_bind_service is lost

### Expected behavior

An image should behave the same after restore or after prebuild as they do originally. Capabilities should not be lost.

### Example repository

https://github.com/rfay/cap-add-demo

aledbf:
**How to test:**
- Create a workspace using https://github.com/rfay/cap-add-demo as source
- Wait until the console appears and show something like this:

Screenshot from 2021-04-01 22-57-00

- Stop the workspace
- Open the workspace
- The console should display the same output
- Optionally, use port-forward with minio, download the backup, decompress the file and run `getcap` against the `nc.traditional` file

fixes #3721

**Note:** any other approach that doesn't require new capabilities is welcome


shaal:
Thank you for working on that!
I followed the instructions and can confirm it works as expected.

I also tested the build of this PR with https://github.com/shaal/ddev-gitpod/pull/21, and it works as well!
image

csweichel (approved):
Code LGTM ...

Re additional capabilities: in Q2 this code will move out of ws-daemon into supervisor anyways. Then, the required capability set is less of a concern.

PR #3730, opened 2021-04-02
Minor dashboard updates

gtsiolis:
This will:
1. Change font weight in some labels.
2. Add missing redirect URL for the `/upgrade-subscription` link triggered from the IDE.
3. ...

### DIFF
--- a/components/dashboard/src/App.tsx
+++ b/components/dashboard/src/App.tsx
@@ -94,6 +94,9 @@ function App() {
                 <Route path={["/subscription", "/usage"]} exact>
                     <Redirect to="/plans"/>
                 </Route>
+                <Route path={["/upgrade-subscription"]} exact>
+                    <Redirect to="/plans"/>
+                </Route>
             </Switch>
         </div>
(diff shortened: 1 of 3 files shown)


components/dashboard/src/App.tsx
```diff
@@ -94,6 +94,9 @@ function App() {
                 <Route path={["/subscription", "/usage"]} exact>
                     <Redirect to="/plans"/>
                 </Route>
+                <Route path={["/upgrade-subscription"]} exact>
```
svenefftinge, line 97: Could you add this to the redirect above, so we have all known aliases per destination in one list?

gtsiolis: Nice catch! Didn't notice the redirect routes above. 🙈 

Moved in cebf7c8.

gtsiolis:
Leaving the pending changes modal as is for now.

@svenefftinge could you take a look, these should be trivial changes and good to merge.

svenefftinge (approved):

Issue #3700, opened 2021-04-01
Update external buttons


svenefftinge:
We need to update the buttons to match the new logo / branding

1) gitpod.io/button/open-in-gitpod.svg (there are other rarely used variants)
2) https://img.shields.io/badge/Gitpod-ready--to--code-blue?logo=gitpod (it has blue in its name 🙄 )

gtsiolis:
For the first one, we could use the ones we have in [these specs](https://www.figma.com/file/q976RVySxCZbSaFJ9t5QiB/Shared?node-id=534%3A1460). For the shield button we could go with the common shield button style and use our logo there or maybe better the monochromatic version.

svenefftinge:
For shield, we'd need to create a PR on their repo. cc @jankeromnes

gtsiolis:
Added https://github.com/gitpod-io/gitpod/pull/3732 but I could be missing something. /cc @jankeromnes @svenefftinge

jankeromnes:
> For shield, we'd need to create a PR on their repo. cc @jankeromnes

Actually, to https://github.com/simple-icons/simple-icons (the icons library used by shields, where the current logo is). I can make a PR today, but they have very strict rules around SVG formatting so I may need some help.

gtsiolis:
I can help with the SVG icon but what about the shield badge colors:
1. Can we change them?
2. Do we want to change them?
3. Where can we change them?

jankeromnes:
> I can help with the SVG icon

Cool, thanks! 🙌 I've opened a Draft PR here: https://github.com/simple-icons/simple-icons/pull/5425

Here are the guidelines on extracting a minimal icon: https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#2-extract-the-icon-from-the-logo

I believe I've already addressed 1. 2. 3. 7. 8., but the `path` still corresponds to the old viewbox (`0 0 36 40`, while `0 0 24 24` is expected) and the icon would also need to be centered.

> but what about the shield badge colors:
>
> 1. Can we change them?

I believe simple-icons controls the logo & color on the left of [Gitpod ready-to-code](https://img.shields.io/badge/Gitpod-ready--to--code-blue?logo=gitpod), however the blue background on the right is controlled by the word `blue` in the badge URL.

So, we can change the logo & color on the left by sending a PR to simple-icons, but to fix the background color on the right we would need to update the shields badge URL everywhere it's used.

> 2. Do we want to change them?
[...]

Issue #3235, opened 2021-02-19
Gitpod's GitHub Integration feature "addComment" no longer works


jankeromnes:
### Describe the bug

When you configure your `.gitpod.yml` with:

```yml
github:
  prebuilds:
    addComment: true
```

the added comment should have a "Open/Review in Gitpod" button. But right now, it seems to have a broken image instead:

<img width="938" alt="Screenshot 2021-02-18 at 20 37 54" src="https://user-images.githubusercontent.com/599268/108491380-1e17a080-72a4-11eb-8dae-c5a7e4e0481e.png">

A few examples out in the wild:

- https://github.com/gitpod-io/workspace-images/pull/349#issuecomment-781541513
- https://github.com/gitpod-io/workspace-images/pull/350#issuecomment-781870832

This might have pushed projects to disable this feature: https://github.com/forem/forem/commit/4b5622f47bfff9a37baf75e37c6532ab8e5e7640

### Steps to reproduce

1. Open a Pull Request against any project using `addComment: true` e.g. https://github.com/gitpod-io/workspace-images

### Expected behavior

- The button image should work

### Additional information

The automated comment's HTML body seems to be generated here:

https://github.com/gitpod-io/gitpod/blob/19fb170e5f4aecef0551acbbedca24d71615bb0d/components/server/ee/src/prebuilds/github-app.ts#L373

Maybe it's `getBadgeImageURL` that no longer works properly:

https://github.com/gitpod-io/gitpod/blob/19fb170e5f4aecef0551acbbedca24d71615bb0d/components/server/ee/src/prebuilds/github-app.ts#L380-L386

### Example repository

https://github.com/gitpod-io/workspace-images/

csweichel:
We should rather remove this functionality. It comes from a time when we didn't know how this feature should work. Closed in favour of  https://github.com/gitpod-io/gitpod/issues/3303

jankeromnes:
Re-opening this -- I think it would be cool to just fix the images while deprecation work is ongoing.

(No need for significant engineering work -- just serving some SVG should be fine).

raisedadead:
Hi @jankeromnes 

Please ignore if this is not valid.

I renabled the GitPod commenting on our repo and we got this comment a little while ago: `https://github.com/freeCodeCamp/freeCodeCamp/pull/41824#issuecomment-825041202`

[...]

PR #3732, opened 2021-04-02
Add external button

gtsiolis:
This will add the external button commonly used in README files.

Related to https://github.com/gitpod-io/gitpod/issues/3700.

| BEFORE | AFTER |
|-|-|
|  light-before | light-after | 
| dark-before | dark-after |


gtsiolis:
**question**: In the meantime, we used to have four variants[[1](https://github.com/gitpod-io/gitpod/blob/0.6.0/components/dashboard/public/button/open-in-gitpod.svg)][[2](https://github.com/gitpod-io/gitpod/blob/0.6.0/components/dashboard/public/button/code-in-gitpod.svg)][[3](https://github.com/gitpod-io/gitpod/blob/0.6.0/components/dashboard/public/button/solve-in-gitpod.svg)][[4](https://github.com/gitpod-io/gitpod/blob/0.6.0/components/dashboard/public/button/review-in-gitpod.svg)]. Are these still needed?

gtsiolis:
WDYT @svenefftinge? I could be missing something. 🔍

gtsiolis:
Two things to also consider:

1. The new one is using a slightly different height `32px` v. `40px`
2. The new one is having a subtle drop shadow.

jankeromnes:
> This will add the external button commonly used in README files.

Wow, great catch! And looks awesome ✨ 

> **question**: In the meantime, we used to have four variants[1][2][3][4]. Are these still needed?

I think we broke them a while ago: https://github.com/gitpod-io/gitpod/issues/3235 but then decided not to fix them and instead deprecate the "add button to issue" feature of the Gitpod GitHub App. I think some people were sad/unhappy about these buttons no longer working.

svenefftinge:
The text in the button seems to sit a bit too high, no?

gtsiolis:
> The text in the button seems to sit a bit too high, no?

Good eyes! 👀  However, this _is_ vertically aligned but visually seems a bit off. We could hack this by adding `1px` offest on the top.

| BEFORE | BEFORE (Spacing)) | AFTER | 
|-|-|-| 
| image |image |  image | 

Here's what it could look like on light and dark GH theme.

| A (Light) | A (Dark) |
|-|-|
| image | image |

Here're some alternative variants we could go with.

| B (Light) | B (Dark) | C (Light) | C (Dark) | D (Light) | D (Dark) |
|-|-|-|-|-|-|
| image | image | image | image | image | image |

What do you think? /cc @svenefftinge @jankeromnes

svenefftinge:
Yes, that's better. I'd prefer the currently proposed variant (white bg / AFTER).

gtsiolis:
Alright, let's do it! Updated the button in https://github.com/gitpod-io/gitpod/pull/3732/commits/72707cd47ff53d30a126fbc1cd5668bf336a7cba and the PR description with some updated screenshots. 🖼️ 

Two points for future reference:
1. The white background could make the button pop up more than expected in dark themes but should be fine as is for now. We can always change this later.
2. I've used the SVG markup from MAD's specs for the button to avoid requesting font files so I've [added a small rectangle](https://github.com/gitpod-io/gitpod/pull/3732/commits/72707cd47ff53d30a126fbc1cd5668bf336a7cba#diff-594b51ad54d97ee19b885313b864b3647d16cbb828c4207c934bad2ed43c02aeR5) above the button label to introduce the `1px` offset.

svenefftinge (approved):

PR #3740, opened 2021-04-05
[example] replace flask example

svenefftinge:
how to test:

- login
- check examples list. 
- Should include the new python-flask example
- Should not include the old flask example


gtsiolis:
Works as advertised! 💯

gtsiolis (approved):

PR #3741, opened 2021-04-05
Revert the recent message-bus changes

svenefftinge:
This PR reverts the recent changes to how the message bus is deployed, as it doesn't work for staging (and possibly prod) yet.

We could alternatively fix the staging deployment.


svenefftinge:
Alejandro is going to take a look if staging can be fixed first.

PR #3747, opened 2021-04-06
[dashboard] fix pending state in Teams

AlexTugarev:
The pending purchase state ...

* should expire after a short while
* should not be persisted until the payment endpoint is hit
* should be updated (removed) on forth/back navigation

### DIFF
--- a/components/dashboard/src/settings/Teams.tsx
+++ b/components/dashboard/src/settings/Teams.tsx
@@ -4,7 +4,7 @@
  * See License-AGPL.txt in the project root for license information.
  */
 
-import React, { useEffect, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import { countries } from 'countries-list';
 import ContextMenu, { ContextMenuEntry } from "../components/ContextMenu";
 import { PageWithSubMenu } from "../components/PageWithSubMenu";
@@ -19,6 +19,7 @@ import exclamation from '../images/exclamation.svg';
 import { ErrorCodes } from "@gitpod/gitpod-protocol/lib/messaging/error";
 import { poll, PollOptions } from "../utils";
 import settingsMenu from "./settings-menu";
+import { Disposable } from "@gitpod/gitpod-protocol";
 
 export default function Teams() {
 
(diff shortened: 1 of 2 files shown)


components/dashboard/src/settings/Teams.tsx
```diff
@@ -49,29 +50,73 @@ function AllTeams() {
     const [inviteMembersModal, setInviteMembersModal] = useState<{ sub: TeamSubscription } | undefined>(undefined);
     const [addMembersModal, setAddMembersModal] = useState<{ sub: TeamSubscription } | undefined>(undefined);
 
-    const [pendingPlanPurchase, setPendingPlanPurchase] = useState<{ plan: Plan } | undefined>(getLocalStorageObject<{ plan: Plan }>("pendingPlanPurchase"));
-    const [pendingSlotsPurchase, setPendingSlotsPurchase] = useState<{ ts: TeamSubscription } | undefined>(getLocalStorageObject<{ ts: TeamSubscription }>("pendingSlotsPurchase"));
+    const restorePendingPlanPurchase = () => {
+        console.log("restorePendingPlanPurchase")
```
svenefftinge, line 54: leftover

components/dashboard/src/settings/Teams.tsx
```diff
@@ -49,29 +50,73 @@ function AllTeams() {
...
+    const pendingPlanPurchasePoller = useRef<Disposable>();
+    const pendingSlotsPurchasePoller = useRef<Disposable>();
+
+    const cancelPollers = () => {
+        pendingPlanPurchasePoller.current?.dispose();
+        pendingSlotsPurchasePoller.current?.dispose();
+    }
 
     useEffect(() => {
+        console.log("useEffect pendingPlanPurchase")
```
svenefftinge, line 92: leftover

components/dashboard/src/settings/Teams.tsx
svenefftinge: it runs 6 times after reload

svenefftinge (commented):
The poller management looks very complicated and from looking at the code I cannot understand the logic behind it.
It tested it on https://at-teams.staging.gitpod-io-dev.com/teams and the pending state never got resolved. Not sure what's going on there, though.

components/dashboard/src/settings/Teams.tsx
AlexTugarev: ah, indeed. will add a check for `pendingSlotsPurchase`. 

though, it's uncritical. `pendingSlotsPurchase` is updated a single time on reload.

AlexTugarev:
> The poller management looks very complicated and from looking at the code I cannot understand the logic behind it.

@svenefftinge, the polling machinery was copied from old dashboard. some parts are removed, e.g. warning calls. 

the new thing is about the persisted state of pending purchases, which indeed looks complicated :-( 

> the pending state never got resolved 

did you receive the confirmation email? if not, that's just the testing account of chargebee being super slow.

svenefftinge:
> did you receive the confirmation email? if not, that's just the testing account of chargebee being super slow.

Yes, I received it now. Then I went back to the screen and it was still flshing in pending state. After reload it went green.

svenefftinge (approved):

PR #3768, opened 2021-04-06
fix workspace file extension

akosyakov:
#### What it does

Rename a workspace file to use the proper extension.

#### How to test

Open this branch in gitpod.io and check explorer. It should be named without the file extension and contain all folders:
Screenshot 2021-04-06 at 14 21 23

Also if you open the workspace file you should get a proper support for the workspace file type.


gtsiolis (approved):
Thanks for picking this up @akosyakov! 🥯

svenefftinge (approved):

PR #3769, opened 2021-04-06
[code] fix donwloading extensions from GCP storage

akosyakov:
#### What it does

Provide `*/*` as a content type, otherwise GCP rejects signed URLs.

Changes in Code: https://github.com/gitpod-io/vscode/commit/a752a8a6ceb43b9add71c7d68f56b0acbdd5dcb7

#### How to test

- We can actually test only on staging, but we should at least check that installing from Open VSX is working with:
https://ak-code-ext-from-gcp.staging.gitpod-dev.com/#https://gitlab.com/gitpod/spring-petclinic


geropl:
testing...

geropl (approved):
LGTM

I can install extensions via open-vsx.org

Issue #3772, opened 2021-04-06
Server/Content Service fail when deleting non-existing workspace backup (which is a regular case)


geropl:
### Bug description

Server throws:
```
Error: 2 UNKNOWN: storage: object doesn't exist
    at Object.exports.createStatusError (/app/node_modules/grpc/src/common.js:91:15)
    at Object.onReceiveStatus (/app/node_modules/grpc/src/client_interceptors.js:1209:28)
    at InterceptingListener._callNext (/app/node_modules/grpc/src/client_interceptors.js:568:42)
    at InterceptingListener.onReceiveStatus (/app/node_modules/grpc/src/client_interceptors.js:618:8)
    at callback (/app/node_modules/grpc/src/client_interceptors.js:847:24)
```
`content-service` throws: `error deleting workspace backup: workspaces/green-gecko-zxdk2tuh/full.tar`

This is caused by a missing workspace backup. This itself is an error case for the workspace, but a regular case for the workspace GC. It should just happily move on instead.

### Steps to reproduce

---

### Expected behavior

`content-service` should happily return without throwing an error.

csweichel:
A single workspace can then block the remainder of the GC run?

PR #3773, opened 2021-04-06
[content-service] Return ErrNotFound if bucket object does not exists

aledbf:
fixes https://github.com/gitpod-io/gitpod/issues/3772


geropl (approved):
LGTM

Thx! :rocket:

geropl:
@aledbf I just checked the client code: It seems we do not properly check for `ErrNotFound`. What about going fixing that as well and move on without this today? :-)

aledbf:
> I just checked the client code

Where exactly? (I read `WorkspaceService` and the code check the error)

geropl:
@aledbf You're absolutely right, I got lost in the `WithError` noise here: https://github.com/gitpod-io/gitpod/blob/e56ed09ccb99dccd27507c3559ca64a49e857786/components/content-service/pkg/service/workspace-service.go#L70-L107

In case of `ErrNotFound` we should default to `Debug()` and not `Error()`.

aledbf:
> In case of ErrNotFound we should default to Debug() and not Error().

Done

geropl (approved):
Awesome, thank you! :pray:

PR #3774, opened 2021-04-06
[ws-manager] Remove ghost workspaces logs from HeadlessListener

aledbf:
https://cloudlogging.app.goo.gl/C9gG2UufWwZaqqAw8

Screenshot from 2021-04-06 10-21-11

### DIFF
--- a/components/ws-manager/pkg/manager/headless.go
+++ b/components/ws-manager/pkg/manager/headless.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"encoding/json"
 	"io"
+	"io/ioutil"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


components/ws-manager/pkg/manager/headless.go
```diff
@@ -172,6 +174,13 @@ func (hl *HeadlessListener) listenAndRetry(ctx context.Context, pod *corev1.Pod,
 		return err
 	}
 
+	logger := log.WithContext(ctx).WithFields(owi).Logger
```
csweichel, line 177: I reckon `ghost` workspaces should not have a headless log listener attached at all.

csweichel: see https://github.com/gitpod-io/gitpod/pull/3776

Issue #3780, opened 2021-04-06
[dashboard] Random non-active workspace considered active


jankeromnes:
### Bug description

For some reason, there is an old non-active workspace in my workspaces list, even though it's not active (not opened recently, not pinned).

<img width="1440" alt="Screenshot 2021-04-06 at 17 05 22" src="https://user-images.githubusercontent.com/599268/113738364-1fd7df00-96ff-11eb-8c6f-4a6603410fc4.png">

Workspace ID is `blue-ferret-kisxfyd7`:

```
mysql> select id, creationTime, contextURL, description, pinned, softDeleted, softDeletedTime, contentDeletedTime from d_b_workspace where id = 'blue-ferret-kisxfyd7';
+----------------------+--------------------------+------------------------------------------------+--------------------------------------+--------+-------------+-----------------+--------------------+
| id                   | creationTime             | contextURL                                     | description                          | pinned | softDeleted | softDeletedTime | contentDeletedTime |
+----------------------+--------------------------+------------------------------------------------+--------------------------------------+--------+-------------+-----------------+--------------------+
| blue-ferret-kisxfyd7 | 2021-03-24T09:04:52.429Z | https://github.com/gitpod-io/sveltejs-template | gitpod-io/sveltejs-template - master |      0 | NULL        |                 |                    |
+----------------------+--------------------------+------------------------------------------------+--------------------------------------+--------+-------------+-----------------+--------------------+
```

Surprisingly, it has zero instances 🤔 maybe that's the root cause:

```
mysql> select * from d_b_workspace_instance where workspaceId = 'blue-ferret-kisxfyd7';
Empty set (0.01 sec)
```

### Steps to reproduce

- Log in to Gitpod (staging) as me

jankeromnes:
Update: After accidentally starting it, and stopping it again, it disappeared from my list.

Then, after deleting the instance from the DB, it showed up again:

```
mysql> delete from d_b_workspace_instance where workspaceId = 'blue-ferret-kisxfyd7';
Query OK, 1 row affected (0.01 sec)
```

Issue #3635, opened 2021-03-28
Errored workspaces don't use the correct status indicator


gtsiolis:
### Problem to solve

The new dasboard surfaces errored workspaces without the correct status indicator.

### Proposal

Use `red-600` for the error status indicator.

| BEFORE | AFTER |
|-|-|
| ![image](https://user-images.githubusercontent.com/120486/112766109-fe009d00-9018-11eb-9fd6-88aa02099ff4.png) |  ![image](https://user-images.githubusercontent.com/120486/112766116-01942400-9019-11eb-81ca-f9c0d48cc779.png) |

corneliusludmann:
@gtsiolis Is it intended, that the interrupted state has color `red-400` and the errored workspace you addressed in this issue should get the color `red-600` as status indicator?

https://github.com/gitpod-io/gitpod/blob/7cc28f0acbe283ce7e6715fb3d2fc7753220f2dc/components/dashboard/src/workspaces/WorkspaceEntry.tsx#L156-L159

corneliusludmann:
What is with the `unknown` state? Currently, the status indicator is rendered with `bg-gitpod-kumquat animate-pulse` (the default, intended for workspaces that are in a transition phase). Should this state get a red indicator as well? The state description does not sound like a transition state but more like a final error state.

https://github.com/gitpod-io/gitpod/blob/9be229e4a2530e333d2cefac1754919ba539119b/components/gitpod-protocol/src/workspace-instance.ts#L85-L87

gtsiolis:
>  Is it intended, that the interrupted state has color ...

Hey @corneliusludmann! I think they can use the same red. I was going to suggest to move forward with the `red-600` but let's go with `red-500` and improve accessibility with color and shape in a later iteration in https://github.com/gitpod-io/gitpod/issues/4225. 🍊 

| `red-400` | `red-500` | `red-600` |
|-|-|-|
| <img width="330" alt="indicator-1" src="https://user-images.githubusercontent.com/120486/120542993-735f6400-c3f4-11eb-95e3-1bb69383d04f.png"> | <img width="330" alt="indicator-2" src="https://user-images.githubusercontent.com/120486/120543001-75292780-c3f4-11eb-80c0-d216bf1a673a.png"> | <img width="330" alt="indicator-3" src="https://user-images.githubusercontent.com/120486/120543003-75c1be00-c3f4-11eb-9307-3facfbf6e3c7.png"> | 

> What is with the unknown state?

[...]

PR #3784, opened 2021-04-06
[dashboard] Don't show workspaces with zero instances as 'active'

jankeromnes:
Fixes https://github.com/gitpod-io/gitpod/issues/3780

This isn't a normal case, but apparently it can happen that a workspace has zero instances if it fails at the wrong time during creation (as probably happened to me in the bug).

### DIFF
--- a/components/dashboard/src/workspaces/workspace-model.ts
+++ b/components/dashboard/src/workspaces/workspace-model.ts
@@ -126,7 +126,7 @@ export class WorkspaceModel implements Disposable, Partial<GitpodClient> {
     
     protected isActive(info: WorkspaceInfo): boolean {
         return info.workspace.pinned || 
-            info.latestInstance?.status?.phase !== 'stopped';
+            (!!info.latestInstance && info.latestInstance.status?.phase !== 'stopped');
     }
 
     public getAllFetchedWorkspaces(): Map<string, WorkspaceInfo> {


jankeromnes:
How to test:

1. Create a new workspace https://jx-active-no-instance.staging.gitpod-dev.com/#https://github.com/jankeromnes/test
2. In https://jx-active-no-instance.staging.gitpod-dev.com/workspaces, copy the workspace ID to somewhere, then stop the workspace
3. Open this PR in Gitpod https://gitpod.io/#https://github.com/gitpod-io/gitpod/pull/3784
4. Run `kubectl port-forward deployments/mysql 3306 &` and then `mysql -h 127.0.0.1 -p` (password is `test`)
6. Run `use gitpod` and then `delete from d_b_workspace_instance where workspaceId = '<ID>'` (where `<ID>` is the workspace ID from step 2.)
7. Visit https://jx-active-no-instance.staging.gitpod-dev.com/workspaces

The stopped workspace should not be visible in `Filter: Active` (only in `Filter: All`).

gtsiolis (approved):
@jankeromnes will this also resolve https://github.com/gitpod-io/gitpod/issues/3635?

jankeromnes:
Thanks @gtsiolis!

> @jankeromnes will this also resolve #3635?

Unfortunately, it won't. This only fixes the situation where workspaces that were never started got shown as "Active".

gtsiolis-test1:
Sounds also good! Thanks for clarifying and fixing this @jankeromnes! 🏀

gtsiolis:
Sounds also good! Thanks for clarifying and fixing this @jankeromnes! 🏀

Issue #3783, opened 2021-04-06
[Blocked] Redirect to /blocked when trying to create or start a workspace while blocked


jankeromnes:
If I'm blocked, this is what I currently see when I try to:

- Create a new workspace:

<img width="600" alt="Screenshot 2021-04-06 at 18 05 52" src="https://user-images.githubusercontent.com/599268/113746379-8d3b3e00-9706-11eb-84b8-f2f9f2c3d223.png">

- Start an existing workspace:

<img width="600" alt="Screenshot 2021-04-06 at 18 06 07" src="https://user-images.githubusercontent.com/599268/113746349-8280a900-9706-11eb-8064-4e6059dbd046.png">

Instead, both errors should automatically redirect to `/blocked`:

<img width="600" alt="Screenshot 2021-04-06 at 18 32 35" src="https://user-images.githubusercontent.com/599268/113746408-96c4a600-9706-11eb-866b-b018694a9d58.png">

Issue #3754, opened 2021-04-06
[Dashboard] Implement /select-account


svenefftinge:
1. First of, ensure you have two accounts with distinct provider connections, e.g. one with GitHub and one with GitLab.
2. Log in with user A.
3. Go to Integerations, try to connect with second provider (which is connected to user B)

You end up on an empty `/select-account` page.

### Designs

| Modal | Modal (Spacing) |
|-|-|
| ![image](https://user-images.githubusercontent.com/120486/113697765-52241500-96dc-11eb-9de7-6ead8f16feb6.png)| ![image](https://user-images.githubusercontent.com/120486/113697666-2f91fc00-96dc-11eb-9535-ffe8e0bf663c.png) |

See also [design specs](https://www.figma.com/file/2VSh4dCfVPTc0YEv1Kngw9/Account-Select-3754?node-id=0%3A1).

AlexTugarev:
@gtsiolis is this design spec still valid? https://github.com/gitpod-io/gitpod/issues/3072

gtsiolis:
@AlexTugarev Oh, specs in #3072 are obsolete but I can add them within the next hour if needed.

To allow some head start, this page could use the following patterns:

1. A modal.
2. Two list entries, as seen in _Existing Workspaces_ specs in https://github.com/gitpod-io/gitpod/issues/3301, with all the needed accout information on each entry. 
3. Two action buttons as seen in the existing page.

AlexTugarev:
>  I can add them within the next hour if needed.

I think this would be very helpful.

gtsiolis:
@AlexTugarev I've added the design specs for this in the description above but wanted to provide also an alterative design that would reuse existing patterns in case this could make us implement this faster.

| Proposal A | Proposal B |
|-|-|
| ![image](https://user-images.githubusercontent.com/120486/113699537-5f420380-96de-11eb-8a24-61850ca5ade9.png) | ![image](https://user-images.githubusercontent.com/120486/113699540-60733080-96de-11eb-935d-15174ca9098c.png) |

PR #3785, opened 2021-04-06
[dashboard] Redirect to /blocked when trying to create or start a workspace while blocked

jankeromnes:
Fixes https://github.com/gitpod-io/gitpod/issues/3783

### DIFF
--- a/components/dashboard/src/start/CreateWorkspace.tsx
+++ b/components/dashboard/src/start/CreateWorkspace.tsx
@@ -85,6 +85,9 @@ export default class CreateWorkspace extends React.Component<CreateWorkspaceProp
             <a href={authorizeUrl}><button className="secondary">Authorize with {error.data.host}</button></a>
           </div>;
           break;
+        case ErrorCodes.USER_BLOCKED:
+          window.location.href = '/blocked';
+          return;
         case ErrorCodes.NOT_FOUND:
           return <RepositoryNotFoundView error={error} />;
         case ErrorCodes.PLAN_DOES_NOT_ALLOW_PRIVATE_REPOS:
--- a/components/dashboard/src/start/StartWorkspace.tsx
+++ b/components/dashboard/src/start/StartWorkspace.tsx
@@ -8,6 +8,7 @@ import EventEmitter from "events";
 import React, { useEffect, Suspense } from "react";
 import { DisposableCollection, WorkspaceInstance, WorkspaceImageBuild, Workspace, WithPrebuild } from "@gitpod/gitpod-protocol";
 import { HeadlessLogEvent } from "@gitpod/gitpod-protocol/lib/headless-workspace-log";
+import { ErrorCodes } from "@gitpod/gitpod-protocol/lib/messaging/error";
 import ContextMenu, { ContextMenuEntry } from "../components/ContextMenu";
 import CaretDown from "../icons/CaretDown.svg";
 import { getGitpodService, gitpodHostUrl } from "../service/service";
@@ -95,6 +96,10 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
       if (typeof error === 'string') {
         error = { message: error };
       }
+      if (error?.code === ErrorCodes.USER_BLOCKED) {
+        this.redirectTo(gitpodHostUrl.with({ pathname: '/blocked' }).toString());
+        return;
+      }
       this.setState({ error });
     }
   }
@@ -138,7 +143,7 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
       this.setState({ hasImageBuildLogs: true });
     }
 
-    let error;
+    let error: StartWorkspaceError | undefined;
     if (workspaceInstance.status.conditions.failed) {
       error = { message: workspaceInstance.status.conditions.failed };
     }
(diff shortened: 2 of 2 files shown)


components/dashboard/src/start/CreateWorkspace.tsx
```diff
@@ -75,16 +76,10 @@ export default class CreateWorkspace extends React.Component<CreateWorkspaceProp
...
-              search: `returnTo=${encodeURIComponent(window.location.toString())}&host=${error.data.host}&scopes=${error.data.scopes.join(',')}`
-          }).toString();
-          window.location.href = authorizeUrl;
-          statusMessage = <div className="mt-2 flex flex-col space-y-8">
-            <p className="text-base w-96">Redirecting to authorize with {error.data.host} …</p>
-            <a href={authorizeUrl}><button className="secondary">Authorize with {error.data.host}</button></a>
-          </div>;
-          break;
+          const authorizeURL = `/api/authorize?returnTo=${encodeURIComponent(window.location.toString())}&host=${error.data.host}&scopes=${error.data.scopes.join(',')}`;
+          return <Redirect push to={authorizeURL}/>;
```
jankeromnes, line 80: Interesting. This does point the address bar to the correct URL, but then the page is blank & doesn't do anything (I need to refresh for the authorization to work).

components/dashboard/src/start/CreateWorkspace.tsx
```diff
@@ -75,16 +76,10 @@ export default class CreateWorkspace extends React.Component<CreateWorkspaceProp
...
+        case ErrorCodes.USER_BLOCKED:
+          return <Redirect to="/blocked"/>;
```
jankeromnes, line 82: Same here 😕

Screenshot 2021-04-07 at 09 21 00

jankeromnes: That's because `<Redirect/>` only works for `<Route/>`s in the same `<Switch/>`.

- `/api/authorize` is not a route
- `/blocked` is currently outside the `<Switch/>`

jankeromnes:
How to test:

- Log in with **GitHub** OAuth at https://jx-redirect-blocked.staging.gitpod-dev.com/login
- Open a **GitLab** workspace, e.g. https://jx-redirect-blocked.staging.gitpod-dev.com/#https://gitlab.com/gitpod/spring-petclinic -- the GitLab OAuth redirect should be fast / barely noticeable (note: you may see a GitLab login/confirmation screen, or nothing GitLab-related if you had already authorized dev-staging before)
- Stop that workspace again from https://jx-redirect-blocked.staging.gitpod-dev.com/workspaces
- Next, open this PR in Gitpod: https://gitpod.io/#https://github.com/gitpod-io/gitpod/pull/3785
- Run `kubectl port-forward deployments/mysql 3306 &` and then `mysql -h 127.0.0.1 -p` (password is `test`)
- Run `use gitpod` and then `select id, name, blocked from d_b_user;` -- this will show your user ID
- Run `update d_b_user set blocked = 1 where id = '<ID>';` (where `<ID>` is your own user ID)
- Going back to https://jx-redirect-blocked.staging.gitpod-dev.com/workspaces, try to restart your old workspace, or to create a new workspace

In both cases, you should get quickly redirected to `/blocked`:

Screenshot 2021-04-09 at 14 00 14

jankeromnes:
@gtsiolis or @AlexTugarev, please take a look if you have time 😊 (Priority: 🔅 Low)

components/dashboard/src/start/CreateWorkspace.tsx
```diff
@@ -75,16 +75,8 @@ export default class CreateWorkspace extends React.Component<CreateWorkspaceProp
           </div>;
           break;
         case ErrorCodes.NOT_AUTHENTICATED:
-          const authorizeUrl = gitpodHostUrl.withApi({
-              pathname: '/authorize',
-              search: `returnTo=${encodeURIComponent(window.location.toString())}&host=${error.data.host}&scopes=${error.data.scopes.join(',')}`
-          }).toString();
-          window.location.href = authorizeUrl;
-          statusMessage = <div className="mt-2 flex flex-col space-y-8">
```
AlexTugarev, line 83: Please undo this change 🙏🏻 

It disagrees with https://github.com/gitpod-io/gitpod/issues/3754. I'm about to add handling for another error mode: user needs to select an account.

jankeromnes: I think we can probably have both changes (faster redirect, and /select-account logic).

But I'm happy to wait for your change and rebase/re-implement mine on top. 🙂

AlexTugarev:
Screen Shot 2021-04-09 at 17 56 07

redirect on blocked state worked!

components/dashboard/src/start/CreateWorkspace.tsx
AlexTugarev: @jankeromnes no we cannot, because we need to open the extra tab instead of the redirect to the auth flow. that extra tab is executing the auth flow and is going to propagate the "select account" error.

AlexTugarev: background: redirecting away from "this page" is always a risk of getting into a dead-end, so the user cannot find back.

jankeromnes: The `returnTo` URL seems to work generally well for that.

And, if there is indeed a rare/major complication in my auth flow, I think it's expected that I don't return to the exact same URL as before, right? (Because I need to deal with the bigger issue first, then later I can try again what I originally wanted to do.)

jankeromnes: But sure, I'll remove my change for now if it makes your more complicated.

jankeromnes: I've spliced my other commit into a separate branch: https://github.com/gitpod-io/gitpod/tree/jx/faster-start-auth-redirect

Next I'll wait for your `/select-account` change to re-evaluate.

AlexTugarev: It’s not about the details of the flow, but it is integrated into the SPA after all. We agreed to try the open in new window with good reasons and knowing that it would introduce this extra step for the benefit of more control of the overall experience. 

Such an extra step is rare, that’s true. So the idea is to support user’s instead of assuming that they know how oauth works.

Also this PR doesn’t sound like it would do this 😉

AlexTugarev (approved):
Works nicely! 🙏

PR #3786, opened 2021-04-06
[chart] Migrate mysql from stable repository

aledbf:
Stable helm repository is deprecated since [November 2020](https://github.com/helm/charts#deprecation-timeline)


csweichel (approved):
Makes sense.
This should be a drop-in replacement, no? I.e. we don't expect things to behave differently because of this.

aledbf:
> This should be a drop-in replacement, no?

Yes.

PR #3791, opened 2021-04-07
[go mod] Update all go.sum files

csweichel:
This PR commits all `go.sum` changes we currently have in the repo.


geropl (approved):
TBH I don't know where this is coming from. But the result is stable, so we should merge anyway, I guess :slightly_smiling_face:

PR #3798, opened 2021-04-07
[github] remove `read:user` from default scopes

akosyakov:
#### What it does

Reverts changing default scopes for GitHub otherwise it breaks existing users, since they cannot grant such scope at all without relogin.

#### How to test

🤷‍♂️


AlexTugarev:
@akosyakov, the "how to test" would be to verify, that new signups/sessions do not have the scope. 

another check would be then to verify if one can make the GitHub PR extension running.

AlexTugarev (approved):
That worked for me.

PR #3816, opened 2021-04-07
[preview envs] Fix workspace deletion

### LINKED ISSUE #3803: [preview envs] Fix workspace deletion
- [ ] deleting workspaces breaks
 - [ ] workspaces don't delete (because finalizer or because <random reason>)
 
 part of epic: https://github.com/gitpod-io/gitpod/issues/3802

geropl:
Fixes #3803.

This should improve the robustness of our preview deployments. But more importantly, it moves our werft scripts from JS to TS. This already fixed multiple small and one bigger bug. It also opens the door for more refactorings in the near future:
 - use the same set of utilities (confidently), or
 - have one set of deployment scripts/jobs alltogether in `core`, with minimal configuration on top for `com`
 
 The PR does by design not add _all_ types but only the most basic/necessary to compile to not add even more noise.

### DIFF
--- a/.werft/build.ts
+++ b/.werft/build.ts
@@ -91,26 +91,18 @@ export async function build(context, version) {
      * Build
      */
     werft.phase("build", "build running");
-    // Build using the dev-http-cache gitpod-dev to make 'yarn install' more stable
-    const buildEnvOpts = {
-        env: {
-            ...process.env,
-            "HTTP_PROXY": "http://dev-http-cache:3129",
-            "HTTPS_PROXY": "http://dev-http-cache:3129",
-        }
-    };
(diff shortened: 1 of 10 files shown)


.werft/build.ts
```diff
@@ -95,23 +91,18 @@ async function build(context, version) {
      * Build
      */
     werft.phase("build", "build running");
-    // Build using the dev-http-cache gitpod-dev to make 'yarn install' more stable
-    const buildEnv = {
```
geropl, line 99: Note: this was already broken several months, just cleaned it up.

corneliusludmann (approved):
When I run “Format Document” in Gitpod/VSCode at least the `build.ts` file has some whitespace changes. I would suggest to run the formatter prior to merging in order to make it easier for future changes.

Everything else looks pretty good. :+1:

PR #3817, opened 2021-04-07
[blobserve] enable long term caching only on success

akosyakov:
#### What it does

Disable caching failed requests to blobserve forever.

#### How to test

- Start a workspace, open devtools, reload the page, check that content of blobserve is served from the disk cache.


geropl (approved):
LGTM

akosyakov:
tested, worked well

PR #3827, opened 2021-04-07
Move external button

gtsiolis:
Move the new external button inside a `/button` directory.


gtsiolis:
WDYT? /cc @svenefftinge because https://github.com/gitpod-io/gitpod/pull/3732.

jankeromnes (approved):
Looks good to me, thanks!

I think we should deploy this rather soon.

PR #3830, opened 2021-04-08
[dashboard] Optimize build

aledbf:
**Changes:**
- migrate `react-scripts` to `craco` https://tailwindcss.com/docs/guides/create-react-app
- remove tailwind css file
- start using `postcss` (craco)

**Screenshot of network differences (main.*.chunk.css): -450Kb**
Screenshot from 2021-04-08 06-25-20
Screenshot from 2021-04-08 06-25-36


Divlo:
Also, we could use the new "Just In Time" mode from `tailwind`.
See: https://tailwindcss.com/docs/just-in-time-mode

aledbf:
> Also, we could use the new "Just In Time" mode from tailwind.

I would prefer to delay that for a different PR. We only need to add `mode: 'jit'` once this PR is merged.

components/dashboard/package.json
```diff
@@ -14,39 +14,38 @@
...
     "postcss": "^7.0.35",
-    "react-scripts": "4.0.2",
+    "react-scripts": "^4.0.3",
     "tailwind-underline-utils": "^1.1.2",
-    "tailwindcss": "npm:@tailwindcss/postcss7-compat@^2.0.2",
+    "tailwindcss": "npm:@tailwindcss/postcss7-compat@^2.1.0",
     "tailwindcss-filters": "^3.0.0",
-    "typescript": "~4.1.2",
-    "web-vitals": "^1.0.1"
+    "typescript": "^4.2.4",
```
AlexTugarev, line 41: why bumping TS version? 
it would be best to update for all components at the same time, otherwise we end up with different TS levels in the editor and build errors might occur.

AlexTugarev:
Build is failing :-( 
@aledbf, would you mind to rebase onto origin/main? It's possible that there were some related changes. 

I'm curious to see that in action. If there are no differences, that'd be awesome! 
Is `purge` applied automatically, or how to explain the difference of the css produced? And if true, we'd need to check the `*.tsx` to ensure, there are no concatenations of class names 🙈

AlexTugarev:
From the build output I cannot see how this impacts the size of the css chunk. 

```
File sizes after gzip:

  543.31 KB  build/static/css/main.b4639cee.chunk.css
```

Looks similar to previous state. 

OTHO I ❤️ the eslint running on the code!

components/dashboard/package.json
aledbf: fixed

aledbf:
> Is purge applied automatically, or how to explain the difference of the css produced?

Yes, that what craco enables https://purgecss.com/guides/react.html#use-craco

AlexTugarev:
> Yes, that what craco enables https://purgecss.com/guides/react.html#use-craco

I though it would not be enabled, because of the missing params. Let me check again...

aledbf:
>  I double-checked, it's not applied automatically. It's currently transferring ~620KB gzipped, while the original size is still ~8megs.

My fault. When I rebased against main I forgot the purge configuration in `tailwind.config.js`.

AlexTugarev:
Awesome! The number of transferred kilobyte dropped to ~8 🎉 

Just quickly scanned what might break according to the docs, i.e. concatenation of class names. But it appears to works 🤷🏻‍♂️, at least the `Always Ready-To-Code` icon size on the Login page seems to be the expected one.

https://github.com/gitpod-io/gitpod/blob/main/components/dashboard/src/Login.tsx#L27
https://github.com/gitpod-io/gitpod/blob/main/components/dashboard/src/admin/UserDetail.tsx#L192

components/dashboard/package.json
```diff
@@ -14,39 +14,38 @@
...
     "@types/react-router-dom": "^5.1.7",
-    "@typescript-eslint/eslint-plugin": "^4.15.2",
-    "@typescript-eslint/parser": "^4.15.2",
+    "@typescript-eslint/eslint-plugin": "^4.21.0",
+    "@typescript-eslint/parser": "^4.21.0",
     "autoprefixer": "^9.8.6",
-    "classnames": "^2.2.6",
-    "eslint": "^7.21.0",
+    "classnames": "^2.3.1",
+    "eslint": "^7.23.0",
```
AlexTugarev, line 34: it would be good to use a single version per component.

```
gitpod /workspace/gitpod $ yarn why eslint
yarn why v1.22.10
[1/4] Why do we have the module "eslint"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "eslint@7.24.0"
info Reasons this module exists
   - "_project_#@gitpod#dashboard" depends on it
   - Hoisted from "_project_#@gitpod#dashboard#eslint"
info Disk size without dependencies: "9.66MB"
info Disk size with unique dependencies: "28.98MB"
info Disk size with transitive dependencies: "33.41MB"
info Number of shared dependencies: 87
=> Found "react-scripts#eslint@7.21.0"
info This module exists because "_project_#@gitpod#dashboard#react-scripts" depends on it.
```

AlexTugarev (approved):
Thanks @aledbf! I like the changes! 
See another nit regarding the eslint versions.

components/dashboard/package.json
aledbf: fixed

PR #3851, opened 2021-04-08
[dashboard] Replace nginx with caddy

aledbf:
**Reasons:**
- Using nginx for a static server is overkill
- Caddy is only one static binary without external dependencies
- Configuration is simpler


components/dashboard/conf/Caddyfile
```diff
@@ -0,0 +1,45 @@
...
+	import default
+	import discard_log
+
+	redir /.well-known/security.txt https://www.gitpod.io/.well-known/security.txt permanent
+	redir /environment-variables /settings/ permanent
+	redir /login-success / permanent
+
+	root * /www
+	file_server
+	try_files {path} /index.html
```
AlexTugarev, line 36: is that an drop-in replacement for `try_files $uri $uri/ /index.html;`?

chart/templates/dashboard-deployment.yaml
```diff
@@ -50,8 +50,8 @@ spec:
         readinessProbe:
           failureThreshold: 3
           httpGet:
-            path: /schemas/gitpod-schema.json
-            port: 80
+            path: /ready
```
AlexTugarev, line 53: that would only mean "the http service is running", the previous check means "the dashboard is served". in-between there might be error cases. 

@geropl, do you know why we started to use actual resources from the dashboard to be tested for in the readiness probe?

csweichel: I can imagine several scenarios why that was:
- convenience: by testing well known dashboard resources, we don't have to think about other paths that we might want to use
- previous breakage: we may have had some misconfiguration at some point and then moved to checking a dashboard resource

Either way, once caddy is running it would serve the dashboard. It makes sense to me to use a readiness endpoint of caddy here.

components/dashboard/conf/Caddyfile
aledbf: Yes

components/dashboard/conf/Caddyfile
```diff
@@ -0,0 +1,45 @@
...
+
+:80 {
+	import caching
+	import compression
+	import default
+	import discard_log
+
+	redir /.well-known/security.txt https://www.gitpod.io/.well-known/security.txt permanent
+	redir /environment-variables /settings/ permanent
+	redir /login-success / permanent
```
AlexTugarev, line 32: please remove this redirect.

aledbf: Without the redirect, I get a 404. Not sure if that's an error

components/dashboard/conf/Caddyfile
AlexTugarev: don't think so. 

we used to have `/foo` been tried as `/foo/` as well. 

would that be valid syntax? `try_files {path} {path}/ /index.html`

components/dashboard/conf/Caddyfile
AlexTugarev: because of https://github.com/gitpod-io/gitpod/pull/3851/files#r610569110

components/dashboard/conf/Caddyfile
aledbf: Yes. https://caddyserver.com/docs/caddyfile/directives/try_files

components/dashboard/conf/Caddyfile
aledbf: fixed

AlexTugarev (approved):
That worked nicely for me!
I checked that caching works as it used to and known redirects are still in place. 
Thanks 🙏🏻

geropl:
@aledbf @AlexTugarev @csweichel Did anybody verify that caddy emits "ETag" headers based on content-hash of the static files in this configuration? Something we rely upon [here](https://github.com/gitpod-io/gitpod/blob/main/chart/config/proxy/lib.locations.conf#L35).

components/dashboard/conf/Caddyfile
```diff
@@ -0,0 +1,44 @@
...
+	auto_https off
+	admin off
+}
+
+(compression) {
+	encode zstd gzip
+}
+
+(caching) {
+	header Cache-Control "public, max-age=604800, must-revalidate"
```
geropl, line 11: Until now caching was the responsibility of the "front-facing" `proxy` (cmp. [here](https://github.com/gitpod-io/gitpod/blob/main/chart/config/proxy/lib.locations.conf#L37-L50)).

This looks fine generally, but we should double check the results, and ideally either:
 a) cross-link with comments, or better
 b) merge into one place.

AlexTugarev:
@geropl, ups. I've seen the ETags, but didn't check for the implementation. It's indeed based on file size and date. 
cf. 
https://github.com/caddyserver/caddy/blob/3f6283b385642c56f34b479d1275095379b062d3/modules/caddyhttp/fileserver/staticfiles.go#L545

Issue #3865, opened 2021-04-09
Bug in team billing


meysholdt:
On staging, I added 100 members to my (previously 7 member strong) Team Unleashed subscription:

![image](https://user-images.githubusercontent.com/239422/114161083-f1891800-9927-11eb-841a-51f7bf103e8e.png)

In Chargebee, however, my subscription count got increased to 7100 (!!!) leading to a way-to-high-invoice.

![image](https://user-images.githubusercontent.com/239422/114161243-21382000-9928-11eb-9947-b772994c2199.png)

PR #3866, opened 2021-04-09
[Teams] Fix quantity type conversion

jankeromnes:
Fixes https://github.com/gitpod-io/gitpod/issues/3865

Explanation:
- `element.value` is always a string in HTML
- When we select a non-default value from the Team Plan Seat quantity selector, it changes `quantity` to a string in the React state
- Somewhere in checkout, something like `newQuantity = originalQuantity + extraQuantity` happens -- when `extraQuantity` is a string, it coerces `originalQuantity` to a string, and then just appends the extra digits

### DIFF
--- a/components/dashboard/src/settings/Teams.tsx
+++ b/components/dashboard/src/settings/Teams.tsx
@@ -616,7 +616,7 @@ function AddMembersModal(props: {
             <div className="flex flex-col space-y-2 pb-4">
                 <label htmlFor="quantity" className="font-medium">Members</label>
                 <select name="quantity" value={quantity} className="rounded-md w-full border-2 border-gray-400"
-                    onChange={(e) => setQuantity(e.target.value as any)}>
+                    onChange={(e) => setQuantity(parseInt(e.target.value || '1', 10))}>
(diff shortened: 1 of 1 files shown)


jankeromnes:
Waiting on https://werft.gitpod-io-dev.com/job/gitpod-build-jx-fix-teams.0

jankeromnes:
How to test:

- In https://jx-fix-teams.staging.gitpod-io-dev.com/teams create a new Team Plan with 10 Members
- Go through billing (tip: pick the proposed fake credit card)
- Once the "payment" is through, add 10 Members to the Team Plan
- Once the second "payment" is through, it should show "20 Members"
- When you click on Billing, it should also show `Team [...] x 20`, **not** `Team [...] x 1010`

AlexTugarev (approved):
LGTM! Thanks for the catch!

jankeromnes:
Thanks! Found a second occurrence of `setQuantity` (when you create a team). Probably benign but fixing as well.

gtsiolis:
Thanks for fixing this @jankeromnes! 💯

PR #3868, opened 2021-04-09
Install node.js dependencies before generating grpc code

aledbf:
**How to test:**

Running `leeway run components:generate-code-from-protobuf` or using the `./generate.sh` script should work without differences


csweichel (commented):
The proto generation changes make sense to me.

Not so sure about the `yarn.lock`. @akosyakov could you have a look?

csweichel (approved):
`yarn.lock` looks good

PR #3870, opened 2021-04-09
[supervisor] thread safe listening to terminals

akosyakov:
#### What it does

grpc is buffering messages to deal with back pressure, so we cannot reuse buffer to read data from the terminal.

#### How to test

Try `ls -l` may times (like 100 🤪 ) in some workspace with many files. It should produce the same result each time.


AlexTugarev (approved):
It works :-) 

Just tried with `top -c` & holding `space`, which is an easy way to reproduce this issue.

Issue #3800, opened 2021-04-07
Clean up token on deauthorize


AlexTugarev:
### Bug description

When disconnecting a provider, tokens should be deleted as well.

### Steps to reproduce

Go to `/integrations` page and disconnect a previously connected provider.

Related to https://github.com/gitpod-io/gitpod/issues/3754

Additional information
* on account deletion / reconnecting with the same account the tokens will be cleaned up ✅ 
* on connecting with another account (after previously disconnecting from first account) the old token remains ❌ 
  * in such cases, no tokens are returned (the DB interface and the `getToken` operation will throw an error,) thus unauthorized access is prevented ✅ 
  * further there is an re-authorization request in place when trying to open a workspace for the affected git provider, which will clean up the situation on success ✅  

**TL;DR** the UX is impacted as it looks like the authorization is missing ❌

PR #3872, opened 2021-04-09
clean up tokens on deauthorize

AlexTugarev:
# clean up tokens on deauthorize
fixes https://github.com/gitpod-io/gitpod/issues/3800

### how to test

1. authorize with two providers
2. disconnect one 
3. verify (in the db 🤷🏻‍♂️) that tokens are (marked as) deleted

# temporarily disable terms acceptance check in API

### DIFF
--- a/components/server/src/user/user-service.ts
+++ b/components/server/src/user/user-service.ts
@@ -196,9 +196,13 @@ export class UserService {
     }
 
     async checkTermsAccepted(user: User) {
-        const terms = this.termsProvider.getCurrent();
-        const accepted = await this.termsAcceptanceDb.getAcceptedRevision(user.id);
-        return !!accepted && (accepted.termsRevision === terms.revision);
+        // disabled terms acceptance check for now
+
+        return true;
+
+        // const terms = this.termsProvider.getCurrent();
+        // const accepted = await this.termsAcceptanceDb.getAcceptedRevision(user.id);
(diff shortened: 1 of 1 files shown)


jankeromnes:
Thanks Alex!

> ### how to test
> 
> 1. revoke permissions for the OAuth App of Gitpod preview environments
> 2. try to login and decline
> 3. verify to see an error instead of a hanging redirect (and timeout)

This is where I end up at (correct-looking `/sorry` URL, but since I'm not logged in, it shows the login screen instead):

Screenshot 2021-04-09 at 17 58 33

AlexTugarev:
@jankeromnes, thanks for testing!

First of, it tries to redirect is an error page (and not to login) and that's what's expected. 

An error page that is only rendered if logged in is problematic at least. :-(

In this particular case though the flow and the resulting redirect is happening in a separate browser tab, so the target for redirect should not be the sorry page. Indeed it should redirect to /login-error (symmetrical to /login-success) and propagate to the still open login page, which will close the extra tab and, that's the plan, render a message. 

One step at a time.

AlexTugarev:
@jankeromnes, I extracted https://github.com/gitpod-io/gitpod/pull/3915 from here. I turns out there is a tail to that problem.

AlexTugarev:
@csweichel, could you have a look at the changes left in this PR, please?

csweichel (approved):
Change LGTM

Issue #3795, opened 2021-04-07
[dashboard] Re-implement 'Cancel Downgrade'


jankeromnes:
Left-over from https://github.com/gitpod-io/gitpod/pull/3550

When I downgrade from a paid plan to a lower paid plan, I see the bottom label `Downgrade Scheduled` below the plan I'm downgrading to, with a `Cancel` link that shows up when the downgrade is confirmed as scheduled (i.e. no longer pending).

However, clicking on `Cancel` doesn't actually do anything. It should:

- Show a confirmation modal, explaining that you're about to cancel the scheduled downgrade to stay on your current plan

- When you accept the prompt, it should cancel the scheduled downgrade

Workaround: You can already cancel any scheduled downgrade via `Billing`, but it's a bit less convenient.

PR #3873, opened 2021-04-09
[dashboard] Re-implement 'Cancel Downgrade' in Plans page

jankeromnes:
Fixes https://github.com/gitpod-io/gitpod/issues/3795

Has no effect on Core / OSS.

How to test:
- Log in to https://jx-cancel-downgrade.staging.gitpod-io-dev.com/plans
- Upgrade to Professional (and complete billing)
- Once the upgrade is through, downgrade to Personal
- Once the downgrade is scheduled, and the "Cancel" link appears, click on it
- The modal should allow you to cancel the scheduled downgrade
- (Also, creating Team and using the invite link yourself should still work)

### DIFF
--- a/components/dashboard/src/settings/Plans.tsx
+++ b/components/dashboard/src/settings/Plans.tsx
@@ -61,45 +61,12 @@ export default function () {
             server.getAppliedCoupons().then(v => () => setAppliedCoupons(v)),
             server.getGithubUpgradeUrls().then(v => () => setGitHubUpgradeUrls(v)),
             server.getPrivateRepoTrialEndDate().then(v => () => setPrivateRepoTrialEndDate(v)),
-        ]).then(setters => setters.forEach(s => s()))
-          .then(() => {
-              handleTeamClaim();
-          });
-
+        ]).then(setters => setters.forEach(s => s()));
 
         return function cleanup() {
             clearTimeout(pollAccountStatementTimeout!);
         }
     }, []);
-
-    const handleTeamClaim = async () => {
-        const teamId = new URL(window.location.href).searchParams.get('teamid');
-        if (!teamId) {
-            return;
-        }
-        const currentlyActiveSubscriptions = (accountStatement?.subscriptions || []).filter(s => Subscription.isActive(s, new Date().toISOString()));
-        const assignedSubscriptions = currentlyActiveSubscriptions.filter(s => AssignedTeamSubscription.is(s));
-        if (assignedSubscriptions.some(s => !!s.teamSubscriptionSlotId)) {
-            return;
-        }
-
-        const freeSlot = await getGitpodService().server.tsGetUnassignedSlot(teamId);
-        if (!freeSlot) {
-            setTeamClaimModal({
-                mode: "error",
-                errorText: "This invitation is no longer valid. Please contact the team owner.",
-            });
-            return;
-        }
-
-        setTeamClaimModal({
-            mode: "confirmation",
-            teamId,
-            slotId: freeSlot.id,
-            text: "You are about to claim a seat in a team.",
-        })
-
-    };
 
     console.log('privateRepoTrialEndDate', privateRepoTrialEndDate);
 
@@ -112,15 +79,35 @@ export default function () {
     const paidPlan = paidSubscription && Plans.getById(paidSubscription.planId);
 
     const assignedTeamSubscriptions = activeSubscriptions.filter(s => AssignedTeamSubscription.is(s));
-    console.log('assignedTeamSubscriptions', assignedTeamSubscriptions);
     const getAssignedTs = (type: PlanType) => assignedTeamSubscriptions.find(s => {
         const p = Plans.getById(s.planId);
         return !!p && p.type === type
     });
-    const assignedUnleashedTs = getAssignedTs('professional');
     const assignedProfessionalTs = getAssignedTs('professional-new');
+    const assignedUnleashedTs = getAssignedTs('professional');
     const assignedStudentUnleashedTs = getAssignedTs('student');
-    const assignedTs = assignedUnleashedTs || assignedProfessionalTs || assignedStudentUnleashedTs;
+    const assignedTs = assignedProfessionalTs || assignedUnleashedTs || assignedStudentUnleashedTs;
+
+    const claimedTeamSubscriptionId = new URL(window.location.href).searchParams.get('teamid');
+    if (!!claimedTeamSubscriptionId && !assignedTeamSubscriptions.some(s => !!s.teamSubscriptionSlotId)) {
+        // Remove all query parameters from the current URL, without actually navigating anywhere
+        window.history.replaceState({}, '', window.location.pathname);
+        getGitpodService().server.tsGetUnassignedSlot(claimedTeamSubscriptionId).then(freeSlot => {
+            if (!freeSlot) {
+                setTeamClaimModal({
+                    mode: "error",
+                    errorText: "This invitation is no longer valid. Please contact the team owner.",
+                });
+            } else {
+                setTeamClaimModal({
+                    mode: "confirmation",
+                    teamId: claimedTeamSubscriptionId,
+                    slotId: freeSlot.id,
+                    text: "You are about to claim a seat in a team.",
+                });
+            }
+        });
+    }
 
     const [ pendingUpgradePlan, setPendingUpgradePlan ] = useState<PendingPlan | undefined>(getLocalStorageObject('pendingUpgradePlan'));
     const setPendingUpgrade = (to: PendingPlan) => {
@@ -444,7 +431,7 @@ export default function () {
                 </div>
                 <div className="flex justify-end mt-6">
                     {teamClaimModal.mode === "confirmation" && (
-                        <React.Fragment>
+                        <>
                             <button className="secondary" onClick={() => setTeamClaimModal(undefined)}>Cancel</button>
                             <button className="ml-2" onClick={async () => {
                                 try {
@@ -462,7 +449,7 @@ export default function () {
                                     })
                                 }
                             }}>Accept Invitation</button>
-                        </React.Fragment>
+                        </>
                     )}
                     {teamClaimModal.mode === "error" && (
                         <button className="secondary" onClick={() => setTeamClaimModal(undefined)}>Close</button>
--- a/components/dashboard/src/settings/Plans.tsx
+++ b/components/dashboard/src/settings/Plans.tsx
@@ -185,6 +185,7 @@ export default function () {
 
     const [ confirmUpgradeToPlan, setConfirmUpgradeToPlan ] = useState<Plan>();
     const [ confirmDowngradeToPlan, setConfirmDowngradeToPlan ] = useState<Plan>();
+    const [ isConfirmCancelDowngrade, setIsConfirmCancelDowngrade ] = useState<boolean>(false);
     const confirmUpgrade = (to: Plan) => {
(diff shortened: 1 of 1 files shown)


jankeromnes:
🎰
/werft run

jankeromnes:
IO preview: https://jx-cancel-downgrade.staging.gitpod-io-dev.com/workspaces/

Currently looks like this:

Screenshot 2021-04-14 at 11 04 24
Screenshot 2021-04-14 at 11 04 34

We might want to adjust the texts a bit.

@AlexTugarev and/or @gtsiolis, could you please take a look when you have time?

components/dashboard/src/settings/Plans.tsx
```diff
@@ -280,6 +268,29 @@ export default function () {
...
+        }
+        setIsConfirmCancelDowngrade(true);
+    }
+    const doCancelDowngrade = async (event: React.MouseEvent) => {
+        if (!isConfirmCancelDowngrade || !paidSubscription) {
+            return;
+        }
+        try {
+            await server.subscriptionCancelDowngrade(paidSubscription.uid);
+            // TODO: Update state somehow?
```
jankeromnes, line 287: Maybe we want to optimistically delete `scheduledDowngradePlanId` here, in order to stop showing the just-cancelled downgrade (immediately, as there is probably no need to wait for the Chargebee confirmation here).

components/dashboard/src/settings/Plans.tsx
```diff
@@ -112,15 +79,35 @@ export default function () {
...
     });
-    const assignedUnleashedTs = getAssignedTs('professional');
     const assignedProfessionalTs = getAssignedTs('professional-new');
+    const assignedUnleashedTs = getAssignedTs('professional');
     const assignedStudentUnleashedTs = getAssignedTs('student');
-    const assignedTs = assignedUnleashedTs || assignedProfessionalTs || assignedStudentUnleashedTs;
+    const assignedTs = assignedProfessionalTs || assignedUnleashedTs || assignedStudentUnleashedTs;
+
+    const claimedTeamSubscriptionId = new URL(window.location.href).searchParams.get('teamid');
+    if (!!claimedTeamSubscriptionId && !assignedTeamSubscriptions.some(s => !!s.teamSubscriptionSlotId)) {
```
AlexTugarev, line 92: it's now unclear, when this is actually executed 🤔 
could you explain the reason?

before, the `tsGetUnassignedSlot` was called on mount, right?

jankeromnes: This simplifies the code a bit (no need for an inline function that is just called once).

Also it's now called on `render()`, when there is a `teamid` in the URL query parameters. Also, as soon as there is a response, the query parameters are stripped away from the URL before showing the modal, so that visiting a team invite links effectively only shows the modal once (and the user can then accept or dismiss it).

AlexTugarev: the code itself looks cleaner indeed.

it still looks like `tsGetUnassignedSlot` may be called more than once. see, the only thing which would stop it from being recalled is the history patching, which happens after `tsGetUnassignedSlot` returns. in the meantime there is a chance that other state manipulations will cause a render cycle.

> effectively only shows the modal once
you are right about that.

components/dashboard/src/settings/Plans.tsx
```diff
@@ -112,15 +79,35 @@ export default function () {
...
+        getGitpodService().server.tsGetUnassignedSlot(claimedTeamSubscriptionId).then(freeSlot => {
+            // Remove all query parameters from the current URL, without actually navigating anywhere
+            window.history.replaceState({}, '', window.location.pathname);
```
AlexTugarev, line 95: maybe patching history before calling `tsGetUnassignedSlot`?

components/dashboard/src/settings/Plans.tsx
jankeromnes: > it still looks like `tsGetUnassignedSlot` may be called more than once.

You're right. I assumed this wasn't a problem (because it doesn't modify anything), but I agree it's probably cleaner to call it just once.

components/dashboard/src/settings/Plans.tsx
jankeromnes: Fixed by introducing a `pendingDowngradeCancellation`.

components/dashboard/src/settings/Plans.tsx
jankeromnes: Good idea, many thanks! Suggestion implemented. ✅

components/dashboard/src/settings/Plans.tsx
jankeromnes: Fixed by implementing your suggestion: https://github.com/gitpod-io/gitpod/pull/3873#discussion_r615009593

jankeromnes:
Alright, should be good to go now. 🚀

@AlexTugarev & @gtsiolis please take another look. 😊 🙏

AlexTugarev:
Just tested the downgrade. Here is what I noticed:

* `Downgrade scheduled` is rendered with a cursor pointer
* `Cancel` link does not appear immediately, actually it took a couple of reloads to get it.

Suggestion: on hover of `Downgrade scheduled` you can render `Cancel downgrade`. WDYT?

Cancellation is also not an instant thing, right? After refresh it still presents with `Cancel` button. It looks like there are a lot of intermediate states :-(

jankeromnes:
> Just tested the downgrade.

Awesome, thanks!

> - `Downgrade scheduled` is rendered with a cursor pointer

Good catch. I can fix that.

> - `Cancel` link does not appear immediately, actually it took a couple of reloads to get it.

That's expected, as we wait for the Chargebee callback here. (When there is a `pendingDowngradeToPlan`, we optimistically show that the downgrade is scheduled, and poll for an update every 10 seconds. Next, when there finally is a `scheduledDowngradePlanId`, we remove the pending state, and show the `Cancel` action.)

jankeromnes:
> Cancellation is also not an instant thing, right? After refresh it still presents with `Cancel` button. It looks like there are a lot of intermediate states with Chargebee :-(

That's a bug in my `pendingDowngradeCancellation` then (if you cancel the downgrade, it should never show the `Cancel` button again, unless Chargebee doesn't call back within 5 minutes).

components/dashboard/src/settings/Plans.tsx
```diff
@@ -294,8 +339,8 @@ export default function () {
     } else {
         const targetPlan = freePlan;
         let bottomLabel;
-        if (scheduledDowngradePlanId === targetPlan.chargebeeId) {
-            bottomLabel = <p className="text-green-600">Downgrade scheduled<br/><a className="text-blue-light leading-6" href="javascript:void(0)">Cancel</a></p>;
+        if (scheduledDowngradePlanId === targetPlan.chargebeeId && !pendingDowngradeCancellation) {
+            bottomLabel = <p className="text-green-600">Downgrade scheduled<br/><a className="text-blue-light leading-6" href="javascript:void(0)" onClick={() => confirmCancelDowngrade()}>Cancel</a></p>;
```
gtsiolis, line 343: **thought**: Wondering if there's a better copy for this one. Do you thinking the following could be better?

```suggestion
            bottomLabel = <p className="text-green-600">Downgrading ...<br/><a className="text-blue-light leading-6" href="javascript:void(0)" onClick={() => confirmCancelDowngrade()}>Cancel</a></p>;
```

gtsiolis, line 343: **issue**: Similarly, when upgrading we are adding the following copy.

```
Upgrade in progress
```

Should we align these two like the following?

```
Upgrading ...
Downgrading ...

or

Upgrade in Progress ...
Downgrade in Progress ...
```

gtsiolis (commented):
Curious why we need this cancel button in the first place? Is there a need for an undo action?

Since there's now a) confirmation to upgrade or downgrade through modals and b) adequate information on the action and potential account changes through alert components, this undo action seems to be redundant, right? 😭 

What do you think? Let me know if I'm missing something. 💭 

FWIW, an undo action on the workspaces deletion action could be great! ⭐

components/dashboard/src/settings/Plans.tsx
gtsiolis, line 343: **suggestion**: In addition to this, disabling all the buttons on all other cards could make sense as they are all broken during an upgrade or a downgrade process. However, this could be considered out of the scope of this PR and a good follow-up issue. 🏓

gtsiolis, line 343: **issue**: Also, when you click the Cancel link the downgrade process is stuck in that state when you close the modal unless you do cancel downgrade, right?

gtsiolis (commented):
Two more minor comments! 🧡

gtsiolis (commented):
~~One last comment for this round:~~

~~@jankeromnes if you think this a) covers some edge cases and b) restores some broken functionality it should be fine to merge. However, it terms of UX I'd vote for removing this undo functionality with the cancel button in the long run unless needed for technical reasons due to the integration with Chargebee. 🤷~~

See https://github.com/gitpod-io/gitpod/pull/3873#discussion_r615992156.

components/dashboard/src/settings/Plans.tsx
jankeromnes: This takes until the next billing cycle, so the thought was:

- If I see `Downgrade scheduled`, I know everything is as expected and can continue with my life
- If I see `Downgrading ...`, the `...` tell me that I should stick around and wait for something to happen

I don't think the `...` is a good idea for things that can take several weeks. 😬

gtsiolis: Ah, ok! My bad! Ignore the comment in https://github.com/gitpod-io/gitpod/pull/3873#pullrequestreview-639052561, then! 😇

gtsiolis: This is also not relevant anymore, see https://github.com/gitpod-io/gitpod/pull/3873#discussion_r615992156.

jankeromnes:
Many thanks for the reviews, @AlexTugarev and @gtsiolis!

@gtsiolis This was more about re-implementing a feature that was available in the old dashboard, but maybe it's okay to leave it out? The use case goes something like this:

- I upgrade to the Professional plan. It's great!
- Actually maybe I want the Personal plan, so I downgrade to it. However, the will only happen at the end of my billing cycle (i.e. in 3 weeks or so), so it says "Downgrade scheduled" for several days
- Actually, the Professional plan is pretty great. I now want to stick with it!

Currently: The downgrade is still schedule for 2-3 weeks from now. To cancel it, I need to navigate Billing very deep in order to find "delete pending changes" or similar.

With this feature: I can click on "Cancel" to cancel the scheduled downgrade, and stick with Professional.

gtsiolis:
Thanks for explaining this in https://github.com/gitpod-io/gitpod/pull/3873#issuecomment-822600677 @jankeromnes! 🙏 

Makes sense now. Since we're not charging per hour usage or something similar this approach seems better for now. 💯

gtsiolis (approved):

components/dashboard/src/settings/Plans.tsx
jankeromnes: It shouldn't be stuck, but rather stay as is until the next billing cycle.

gtsiolis: Yes, it makes sense after https://github.com/gitpod-io/gitpod/pull/3873#discussion_r615992156. 🏁

Issue #3645, opened 2021-03-29
[server] Support email domain base exceptions when auto-blocking new users


csweichel:
Gitpod provides a mechanism by which it can automatically block newly signed up users. We're using this for some of our development environments.

We'd need to be able to add exceptions based on the email domains of the incoming users, e.g.`@gitpod.io`

PR #3899, opened 2021-04-12
[staging] introduce blockNewUSersWhiteList

svenefftinge:


jankeromnes:
Thanks! I suspect this is somehow related to #3645 (mentioning the issue here in order to facilitate GitHub navigation)

csweichel:
The way the PR stands right now it doesn't do much. Looks like it's missing the usage in `user-controller.ts:142`.

PR #3914, opened 2021-04-13
[whats-new] update ideSettings if not set before

AlexTugarev:
# what's new / ideSettings

if `additionalData` was empty before, it should now end up with populated `ideSettings` after checking the what's new modal.


jankeromnes:
Taking a look at this now. 👀

jankeromnes (approved):
Code change looks good to me (🧙‍♂️) and works as advertised.

Thanks! Approved. 🚀

PR #3916, opened 2021-04-13
[werft] set google terraform plugin to version 3.63.0

wulfthimm:
DNS could not be overwritten in the newer version of the Google Terraform plugin.


geropl (approved):
LGTM

PR #3922, opened 2021-04-13
[dashboard] Refactoring of cache headers

### LINKED ISSUE #3818: [proxy] Make sure we do not (long-term) cache error responses
We recently introduced long-term caching for [static assets](https://github.com/gitpod-io/gitpod/blob/e56ed09ccb99dccd27507c3559ca64a49e857786/chart/config/proxy/lib.locations.conf#L25-L32).

While atm we have only a handful of resources under `/static` we should make sure that we do not accidently cache errors like 404s in the future, for instance by using a pattern like this: https://serverfault.com/a/690258

aledbf:
fixes #3818

### DIFF
--- a/components/dashboard/conf/Caddyfile
+++ b/components/dashboard/conf/Caddyfile
@@ -7,14 +7,6 @@
 	encode zstd gzip
 }
 
-(caching) {
-	header Cache-Control "public, max-age=604800, must-revalidate"
-}
-
-(default) {
-	header -Server
-}
-
 (discard_log) {
 	log {
 		output discard
@@ -22,13 +14,30 @@
 }
 
 :80 {
-	import caching
+	@static {
+		path /static/*
+	}
+
+	@notstatic {
+		not path /static/*
+	}
+
 	import compression
-	import default
 	import discard_log
 
 	redir /.well-known/security.txt https://www.gitpod.io/.well-known/security.txt permanent
 	redir /environment-variables /settings/ permanent
+
+	handle @static {
+		header Cache-Control "public, max-age=31536000"
+	}
+
+	handle @notstatic {
+		header Cache-Control "no-cache, no-transform, must-revalidate"
+	}
+
+	header -Server
+	header -Last-Modified
 
 	root * /www
 	file_server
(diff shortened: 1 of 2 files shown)


components/dashboard/conf/Caddyfile
```diff
@@ -7,29 +7,38 @@
...
+	handle @static {
+		header Cache-Control "public, max-age=31536000"
+	}
+
+	handle @notstatic {
+		header Cache-Control "no-cache, no-transform, must-revalidate"
+	}
+
+	header -Server
+	header -Last-Modified
```
geropl, line 40: nit: this could be moved to `@notstatic`. But does not harm here, either.

akosyakov (approved):
cache-control headers looking good, static resources are not fetched on reload, but non static are always checked

components/dashboard/conf/Caddyfile
```diff
@@ -7,29 +7,38 @@
...
 
 	redir /.well-known/security.txt https://www.gitpod.io/.well-known/security.txt permanent
 	redir /environment-variables /settings/ permanent
 
+	handle @static {
+		header Cache-Control "public, max-age=31536000"
+	}
+
+	handle @notstatic {
+		header Cache-Control "no-cache, no-transform, must-revalidate"
```
geropl, line 36: I'd expect at least a short comment about the two types of content and about the intent and trade-offs regarding the two caching strategies.

geropl (approved):
> fixes #3818

I'm not sure it does. What is the default behavior of caddy on error responses?

I'd love to see more comments here, but let's merge now as it:
 - works as the current setup
 - unblocks a hotfix deployment

components/dashboard/conf/Caddyfile
```diff
@@ -7,29 +7,38 @@
...
-
 (discard_log) {
 	log {
 		output discard
 	}
 }
 
 :80 {
-	import caching
+	@static {
```
akosyakov, line 17: Does it ensure that 404 won't be cached? Should we do something like here https://caddy.community/t/correct-way-to-set-expires-on-caddy-2/7914/13?

PR #3926, opened 2021-04-13
[image-builder] Change default garbage collector interval

### LINKED ISSUE #3887: Fix image builder garbage collection mechanism
the image builder keeps accumulating several hundreds of gigabytes of disk space until we run out of disk space.

~Running `docker system prune` does the trick. This should happen automatically. Maybe every hour or when 100GB disk usage is reached.~

The image builder already has a mechanism to garbage-collect old images. We should fix it.

aledbf:
The actual value is [2h](https://github.com/gitpod-io/gitpod/blob/28e154b38d9a574f57ccca573de3638494377ee0/components/image-builder/pkg/builder/builder.go#L50)

fixes #3887

### DIFF
--- a/components/image-builder/pkg/builder/gc.go
+++ b/components/image-builder/pkg/builder/gc.go
@@ -76,7 +76,12 @@ func (gc *GarbageCollector) CollectGarbage(ctx context.Context, maxAge time.Dura
 
 // Start schedules the garbage collector at regular intervals removing all artifacts older than maxAge
 func (gc *GarbageCollector) Start(ctx context.Context, maxAge time.Duration) {
-	t := time.NewTicker(maxAge / 2)
+	d := maxAge / 10
+	if d > 10*time.Minute {
+		d = 10 * time.Minute
+	}
+
+	t := time.NewTicker(d)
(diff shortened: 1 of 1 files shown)


components/image-builder/pkg/builder/gc.go
```diff
@@ -76,7 +76,7 @@ func (gc *GarbageCollector) CollectGarbage(ctx context.Context, maxAge time.Dura
 
 // Start schedules the garbage collector at regular intervals removing all artifacts older than maxAge
 func (gc *GarbageCollector) Start(ctx context.Context, maxAge time.Duration) {
-	t := time.NewTicker(maxAge / 2)
+	t := time.NewTicker(10 * time.Minute)
```
csweichel, line 79:
```suggestion
        d := maxAge / 10
        if d > 10 * time.Minute {
            d = 10 * time.Minute
        }
	t := time.NewTicker(d)
```

aledbf: done

csweichel (approved):
LGTM

Issue #3511, opened 2021-03-19
[admin dashboard] User: Add button to grant n Gitpod Hours as one-time credit


geropl:
This is mostly used to unblock users temporarily that ran into our paywall, have work left to finish but cannot/ do not want to pay

PR #3929, opened 2021-04-14
[dashboard] Allow granting a user 20 extra hours from the admin dashboard

jankeromnes:
Fixes https://github.com/gitpod-io/gitpod/issues/3511.

Has no effect on Core / OSS.

How to test:
- In a IO context, go to the Admin panel*, and search for your user
- Click on "Grant 20h", and confirm that this adds 20h to your balance

---

*Reminder: If you need to set yourself as admin:
- Open the IO branch in Gitpod
- Run `kubectl port-forward statefulset/mysql 3306 &` and then `mysql -h 127.0.0.1 -p` (the password is `test`)
- Run `use gitpod` and then `select id, name, rolesOrPermissions from d_b_user;` (this shows your user ID)
- Run `update d_b_user set rolesOrPermissions = '["admin"]' where id = <ID>;` (where `<ID>` is your user ID)
- You're now admin 🎉

### DIFF
--- a/components/dashboard/src/admin/UserDetail.tsx
+++ b/components/dashboard/src/admin/UserDetail.tsx
@@ -104,45 +104,52 @@ export default function UserDetail(p: { user: User }) {
                     <div className="flex w-full mt-6">
                         <Property name="Sign Up Date">{moment(user.creationDate).format('MMM D, YYYY')}</Property>
                         <Property name="Remaining Hours"
-                            action={
-                                accountStatement && {
+                            actions={
+                                accountStatement && [{
                                     label: 'View Account Statement',
                                     onClick: () => setViewAccountStatement(true)
-                                }
+                                }, {
+                                    label: 'Grant 20h',
+                                    onClick: async () => {
+                                        await getGitpodService().server.adminGrantExtraHours(user.id, 20);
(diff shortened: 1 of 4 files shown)


components/dashboard/src/admin/UserDetail.tsx
```diff
@@ -104,45 +104,52 @@ export default function UserDetail(p: { user: User }) {
...
                         <Property name="Remaining Hours"
-                            action={
-                                accountStatement && {
+                            actions={
+                                accountStatement && [{
                                     label: 'View Account Statement',
                                     onClick: () => setViewAccountStatement(true)
-                                }
+                                }, {
+                                    label: 'Grant 20h',
```
gtsiolis, line 112: **question**: This works fine as is. Just adding an alternative phrase in case this seems better. Feel free to leave this as is.

```suggestion
                                    label: 'Add 20 additional hours',
```

gtsiolis (approved):

components/dashboard/src/admin/UserDetail.tsx
jankeromnes: Makes sense, thanks! I've slightly edited your suggestion so that I can apply it in one click 😇 (notes: I prefer "Grant" over "Add" because I find it more explicit + capitalized like other action labels + "Extra" is shorter than "Additional" 🚀)

PR #3932, opened 2021-04-14
publish npm packages from main always

### LINKED ISSUE #3895: publish npm packages from master always
Otherwise it is not clear which version is latest for npm and installing `@gitpod/gitpod-protocol@master` installs something old.

akosyakov:
- [x] /werft publish-to-npm

#### What it does

fix #3895: It adds a new `publish-to-npm` config option for werft. If it's missing then packages should be published only if there are src or deps changes, otherwise always. For main branch this option is always present.

#### How to test

Check logs that after each werft run packages get republished from this PR.

### DIFF
--- a/.werft/build.ts
+++ b/.werft/build.ts
@@ -55,7 +55,7 @@ export async function build(context, version) {
     } catch (err) {
         werft.fail('prep', err);
     }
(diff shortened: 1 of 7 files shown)


csweichel (approved):
LGTM

geropl (approved):
:+1:

PR #3935, opened 2021-04-14
[dashboard] Improve handling of errors and cache headers

### LINKED ISSUE #3818: [proxy] Make sure we do not (long-term) cache error responses

aledbf:
fixes #3818


aledbf:
A request to non-existing resources returns 404 and no-cache headers

```console
curl http://localhost:8080/static/js/113.a4b457a801.chunk.js -vvv
*   Trying ::1:8080...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> GET /static/js/113.a4b457a801.chunk.js HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.68.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 404 Not Found
< Date: Wed, 14 Apr 2021 12:25:56 GMT
< Content-Length: 9
< 
* Connection #0 to host localhost left intact
Not found
```

akosyakov (approved):
worked well for me

Issue #3839, opened 2021-04-08
Environment Variables with the same name are hidden from the list


gtsiolis:
### Bug description

When two environment variables have the same name we only surface one of them on the environment variables list (`/variables`) regardless if the scope is different.

See [relevant discussion](https://community.gitpod.io/t/environment-variables/3183). /cc @erasmuswill

### Steps to reproduce

1. Go to Settings / Environments Variables (`/variables`).
2. Add a new variable.
3. Add a second variable with the same name. You can also try using a different scope.

### Expected behavior

You should be able to see both variables listed.

| Variable 1 | Variable 2 | List |
|-|-|-|
| ![image](https://user-images.githubusercontent.com/120486/113993446-28462c00-985d-11eb-82ab-3966849fb963.png) |![image](https://user-images.githubusercontent.com/120486/113993454-29775900-985d-11eb-9d57-d4a2679d9648.png) | ![image](https://user-images.githubusercontent.com/120486/113993461-2a0fef80-985d-11eb-80b4-9ac2b314f52a.png) |

PR #3569, opened 2021-03-24
decouple gp env from theia

akosyakov:
#### What it does

- fix #3162: decouple gp env from theia

#### How to test

- test both Theia and VS Code
- get: should return exact and wildcard matches, should not return unmatched i.e. wildcards or other repos
- set/delete: should create, updated or delete based on exact matches, should not mutate unmatched i.e. wildcards or other repos

PR #3938, opened 2021-04-14
[server] Implement a getAllEnvVars() method that doesn't filter on repository pattern precedence

jankeromnes:
Fixes https://github.com/gitpod-io/gitpod/issues/3839

Context: The repository pattern precedence filtering was hoisted from theia into server in https://github.com/gitpod-io/gitpod/pull/3569, but it's not actually relevant to show users all their environment variables.


jankeromnes:
Ok, this now works as expected again. Please take a look. 😇

gtsiolis:
Looking at this now! 👀

gtsiolis (approved):
Thank you for fixing this @jankeromnes!

**question**: Is it expected that using the sanme variable name and project scope does not work? If so, could we add a validation error message telling the user that this variable name for this scope alrady exists? Feel free to open a follow up issue for this. 🏓

jankeromnes:
Many thanks for the quick review! 💯

> Feel free to open a follow up issue for this.

Whoops, indeed. Done: https://github.com/gitpod-io/gitpod/issues/3942

PR #3945, opened 2021-04-15
[dashboard] don't delete all workspaces on enter

svenefftinge:


csweichel (approved):
LGTM

apolopena:
The accessibility is now gone. It would be better for users that rely on accessibility if the enter key actually worked like it should.

Issue #3836, opened 2021-04-08
[Admin] Blank page when viewing deleted user


jankeromnes:
### Bug description

When visiting `/admin/users/<ID-of-a-deleted-user>`, the page is full blank.

The browser console says:

```
Error: No identity with primary email for user: <ID>!         UserSearch.tsx:36:37
```

### Steps to reproduce

- Open a deleted user in the admin panel

### Expected behavior

It should load the app's shell (menu etc), and show that this user was deleted (e.g. before it showed values like "deletedName", "deletedEmail" etc).

gtsiolis:
Re-posting from the [relevant discussion](https://gitpod.slack.com/archives/C01TFSMBYQ6/p1618296123000400) (internal):

> **George**: Do we keep ANY data to our database after the user has been deleted? Surfacing explicitly that the user has been deleted would require us to store some data for this and could introduce some privacy issues, right? Alternatively, redirecting to the admin area with an alert component that mentions that user does not exist could be optimal but I could be wrong.

gtsiolis:
Still an issue. Adding never-stale label.

Devansu-Yadav:
@gtsiolis If this issue is still prevalent, could I help fix this?

PR #3950, opened 2021-04-15
Handle "Email address is taken" on Login page

### LINKED ISSUE #3801: Does decline work with GitHub?
### Bug description

Chris tried to decline the OAuth flow with GitHub, but the callback request didn't result in the expected redirect to `/sorry`

### Steps to reproduce

Try to decline the OAuth flow with GitHub

### Expected behavior

Sorry page should be shown.

### Example repository

_No response_

AlexTugarev:
This PR adds a check for already taken email addresses on signup. You supposed to see a message on the Login page that explains why a signup with an Identity Provider didn't work, because you previously created a Gitpod account with a different IdP but matching email address.

The intention of this change is to minimize the unintentional creation of second Gitpod accounts, just by using a different IdP.

Try it yourself:

  * make sure you have a GH and a GL session and the primary email address is matching
  * sign up with one provider first
  * logout
  * try to sign up with the second provider
  * see the error message

Screen Shot 2021-04-15 at 15 31 00

––

This PR also fixes #3801

How to test:
* go to GH setting and revoke access for the Gitpod Preview Environments OAuth app
* try to sign up and decline(!)
* see the error message

Screen Shot 2021-04-15 at 15 29 07


gtsiolis:
Looking at this in a moment! 👀 

Feel free to update the description if there are any particular steps to follow to help reviwing this.

AlexTugarev:
gitlab.com uses a similar strategy BTW

Screen Shot 2021-04-15 at 12 13 34

components/dashboard/src/Login.tsx
```diff
@@ -126,6 +136,12 @@ export function Login() {
                         </div>
                     </div>
                 </div>
+                {emailTaken && (<div className="bg-gitpod-kumquat-light rounded-md p-3 text-gitpod-red text-sm mb-2">
```
gtsiolis, line 139: **nitpick**: Alerts like this can use a larger radius to differentiate them from inputs or other elements as we do in the garbage collection message in the workspaces lists.

Some thoughts on this:
1. This pattern will become more clear and consistent as its product design evolves.
2. Similarly, colors used for each alert type will also become more clear as design guidelines shape up.

```suggestion
                {emailTaken && (<div className="bg-gitpod-kumquat-light rounded-xl p-3 text-gitpod-red text-sm mb-2">
```

components/dashboard/src/Login.tsx
```diff
@@ -126,6 +136,12 @@ export function Login() {
...
+                    {`Email address is already in use. Please log in with ${emailTaken.otherUser.authHost}.`}
+                </div>)}
+                {errorMessage && (<div className="bg-gitpod-kumquat-light rounded-md p-3 text-gitpod-red text-sm mb-2">
```
gtsiolis, line 142: **nitpick**: Radius fix.

```suggestion
                {errorMessage && (<div className="bg-gitpod-kumquat-light rounded-xl p-3 text-gitpod-red text-sm mb-2">
```

components/dashboard/src/Login.tsx
gtsiolis, line 139: **suggestion**: Can we make this alert component dismissable?  WDYT?

components/dashboard/src/Login.tsx
```diff
@@ -126,6 +136,12 @@ export function Login() {
                         </div>
                     </div>
                 </div>
+                {emailTaken && (<div className="bg-gitpod-kumquat-light rounded-md p-3 text-gitpod-red text-sm mb-2">
+                    {`Email address is already in use. Please log in with ${emailTaken.otherUser.authHost}.`}
```
gtsiolis, line 140: **suggestion**: Could we use the alert icon (❗) in front of the text as we do with the garbage collection message and other alert component instances? This can help make it clear to the user that this element is not a graphic affordance already existing there but something is wrong with the current state.

gtsiolis, line 140: **suggestion:** What do you think of not surfacing this kind of information to anonymous users? Feel free to rephrase.

```suggestion
                    {`Email address already exists. Log in using a different provider.`}
```

gtsiolis, line 140: **suggestion**: Ideally, we could have a title text for briefly explaining what happened and a body text to clarify what needs to be done. This way we can also make the title more prominent. We can tackle this separately and leave this as is for now in one piece.

gtsiolis (approved):
Thanks for adding this @Alex and the detailed write-up on how to review this! 🌟 

FWIW, this pattern will become useful for tackling similar issues like https://github.com/gitpod-io/gitpod/issues/3836.

Left some minor comments below, feel free to resolve here or in a follow up PR.

There are three changes that would be great to go into this PR:
1. Center align the alert component.
2. Limit alert width using `w-96`.
3. Add some margin below the alert.

Pasting below how this could ideally look like.

Error

components/dashboard/src/Login.tsx
AlexTugarev: Being more conservative sounds ok for me, thus removing those bits.

components/dashboard/src/Login.tsx
AlexTugarev: Given that the Login page is just a stop between here and there, I'd personally never try to dismiss it, but if you think this is useful 🤷🏻‍♂️ let's add it. Also it's dismissed on next try to connect.

AlexTugarev: Could we add some reuseable components for that, please? Each time I pick one example from somewhere else, it's the wrong one.

gtsiolis: It should be fine as non-dismissible for now as the task is blocked if no action is taken. Usually, an alert component is non-dismissible when the user is prevented from completing a task until a system-condition is met. Since this is user-driven and not relying on a system condition, a dismissible alert is probably the correct approach here. 💡 

However, we can tackle this in a follow up issue going through all alerts and make sure we make dismissible the ones needed. Your call! 📞

gtsiolis: Strong vote for YES!

This is common due to a) some design experimentations added on purpose in some pages and b) our MVC approach with helps us ship faster and unblock code reviews. The need to introduce a reusable component library will become more clear on the way as a) the product evolves and b) team members try to add or improve product features on more areas.

For this iteration, going with the `rounded-xl` class should suffice. WDYT?

AlexTugarev: > However, we can tackle this in a follow up issue going through all alerts and make sure we make dismissible the ones needed. Your call! 📞

sounds good!

AlexTugarev: Sure! ☝🏻 updated the screenshots in the description above.

PR #3951, opened 2021-04-15
[build] revert removal of DO_PUBLISH env var

akosyakov:
It is to fix the build in com. If the build is green then it is fine.


jankeromnes (approved):
This fixes the build issue I found, so unblocking. Thanks!

However, I have no idea if this change can have other/unwanted side effects.

PR #3959, opened 2021-04-15
[ws-manager] Remove broken phase spans

ArthurSens:
We've been heavily using Jaeger and distributed tracing while working on improving workspace startup times. During our analysis, we've noticed that `/workspace/phase-<some_phase>` spans are broken and don't represent the correct duration of such event.

At the same time, we already have another span that is emitted and closed at the exact same points,`/workspace/handle-<some_phase>`, but those seem to be a lot more representative of the real behavior.

This PR removes `/workspace/phase-<some_phase>` with the intent of removing distractions while analyzing traces.

## How to test
* Log in into the preview, start some workspaces
* Open a workspace on this PR, run the command `kubectl port-forward <jaeger_pod> 16686` and open the preview on browser
* Search for traces with:
  *  Service: ws-manager
  * Operation: StartWorkspace
* Open any trace found
* Notice that  `/workspace/phase-<some_phase>` spans are no longer there!

image

### DIFF
--- a/components/ws-manager/pkg/manager/monitor.go
+++ b/components/ws-manager/pkg/manager/monitor.go
@@ -65,9 +65,6 @@ type Monitor struct {
 	eventpool *workpool.EventWorkerPool
 	ticker    *time.Ticker
 
-	inPhaseSpans     map[string]opentracing.Span
-	inPhaseSpansLock sync.Mutex
-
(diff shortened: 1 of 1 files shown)


csweichel (approved):
LGTM

PR #3964, opened 2021-04-15
[proxy] Replace nginx with Caddy

### LINKED ISSUE #3611: Proxy seem to drop some requests
### Bug description

Working on https://github.com/gitpod-io/vscode is currently pretty unreliable, as the started webserver serves a huge amount of files and our proxy regularly seems to swallow some of the requests (no responses).
We have to try reloading multiple times until it works.

### Steps to reproduce

1) open gitpod.io/#https://github.com/gitpod-io/vscode
2) wait for the additional window to load (if it stays white you might have seen such a loading issues
3) open dev tools and see network tab.

### Expected behavior

_No response_

### Example repository

_No response_

aledbf:
**Changes:**
- Migrate proxy from NGINX to Caddy
- Remove bash script to build configuration files
- Remove chart configuration files
- Cleanup chart variables
- Remove the option to use HTTP or HTTPS and make HTTPS mandatory.
- Migrate lua scripts to Caddy middleware (Go)

**TODO:**
~~- make minio and docker-registry optional~~
~~- Private workspace ports~~
- Custom 404 error pages?

fixes #3611


akosyakov:
I tried with https://github.com/gitpod-io/vscode and it seems to load fast everything and reliable but something is failing with CORS issue, so port location does not get revealed:
Screenshot 2021-04-19 at 13 57 02

```
Access to fetch at 'https://copper-butterfly-4r28ua28.ws-dev.aledbf-proxy2caddy.staging.gitpod-dev.com/_supervisor/v1/info/workspace' from origin 'https://3000-copper-butterfly-4r28ua28.ws-dev.aledbf-proxy2caddy.staging.gitpod-dev.com' has been blocked by CORS policy: The 'Access-Control-Allow-Origin' header contains multiple values 'https://3000-copper-butterfly-4r28ua28.ws-dev.aledbf-proxy2caddy.staging.gitpod-dev.com, https://3000-copper-butterfly-4r28ua28.ws-dev.aledbf-proxy2caddy.staging.gitpod-dev.com', but only one is allowed. Have the server send the header with a valid value, or, if an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

aledbf:
@akosyakov please test again :pray:

akosyakov:
in regards to https://github.com/gitpod-io/gitpod/issues/3611, it loads first time normally, but on subsequent page reload it hangs, additional reloads don't help either.

components/proxy/conf/Caddyfile
```diff
@@ -235,6 +235,13 @@ https://*.*.{$GITPOD_DOMAIN} {
 		#gitpod.workspace_port_auth {
 		#	auth_url http://server.{$KUBE_NAMESPACE}.{$KUBE_DOMAIN}:3000/auth/workspace-port/{re.host.workspacePort}
 		#}
+		# static assets configure cache headers and do not check for changes
+		@static_asset {
+			path /out/vs/*
```
akosyakov, line 240: i don't think we should hard code any caching for port locations. Gitpod Code Server already does caching based on etags, we should make sure that we actually just don't intervene and don't drop/timeout requests.

aledbf: @akosyakov yes, this was just to test the behavior. Using vscode inside a workspace requires ~2000 requests every time you reload :upside_down_face:

akosyakov:
@aledbf It works very nice now! What was the issue? It would be good if Gero and Chris look at changes generally.

chart/templates/proxy-configmap.yaml
```diff
@@ -0,0 +1,50 @@
+# Copyright (c) 2020 Gitpod GmbH. All rights reserved.
```
akosyakov, line 1: 2021

aledbf:
> It works very nice now! What was the issue?

There is an issue in Go related to http/2 https://github.com/golang/go/issues/42534#issuecomment-790951253
(temporal workaround is to disable http/2). TODO: check if this affects `ws-proxy` when is exposed directly (pluggable workspaces)

akosyakov:
> There is an issue in Go related to http/2 golang/go#42534 (comment)
(temporal workaround is to disable http/2). TODO: check if this affects ws-proxy when is exposed directly (pluggable workspaces)

I see, wonder whether it can affect end users? e.g. if one tests an app loading time on a port location then he can get wrong impression because of disabled http2, i.e. resources will be loaded one after another, not parallel? or it does not affect them?

aledbf:
> I see, wonder whether it can affect end users? 

Maybe? I can perceive a difference only opening something like vscode (~2000 request every time)

> resources will be loaded one after another, not parallel? 

There is a default of 6 parallel downloads. This is really old but the values are similar https://www.stevesouders.com/blog/2008/03/20/roundup-on-parallel-connections/

akosyakov:
Unfortunately it makes really huge difference from perf perspective. Our website loaded with slow 3g with disabled caching in production vs the PR:
Screenshot 2021-04-24 at 22 12 15
Screenshot 2021-04-24 at 22 13 01

The thing is that optimisation techniques with http2 and without are quite different, and not having http2 on port location will harm usefulness of Gitpod in this sense.

geropl:
> The thing is that optimisation techniques with http2 and without are quite different, and not having http2 on port location will harm usefulness of Gitpod in this sense.

Hm. I'd really love to see us move away from nginx. But it'd would also nice to not have to compromise. @aledbf Would it make sense to wait just a little longer and check if either:
 a) https://github.com/golang/go/issues/42534 gets fixed upstream
 b) caddy uses the fixed `net/http` code ([this](https://github.com/caddyserver/caddy/pull/3910/) looks like it already does...?) But the question is if we really want to go with untested code here.

Personally I'm inclined to wait: As nice as a switch would be it's not blocking us AFAIK, and the nginx ~mess~ stuff is stable atm. Also, the PR is not likely to rot fast. WDYT?

aledbf:
> a) golang/go#42534 gets fixed upstream

Agree.

> b) caddy uses the fixed net/http code

I tried the proposed fix without success.

> Personally I'm inclined to wait:

Yes, I agree with that. That is why this PR is still a draft.

aledbf:
@akosyakov please try again

Screenshot from 2021-05-10 10-59-08

chart/templates/proxy-configmap.yaml
aledbf: fixed

components/proxy/plugins/workspace/go.mod
```diff
@@ -0,0 +1,8 @@
+module github.com/gitpod-io/gitpod/proxy/plugins/workspace
+
+go 1.16
+
+require (
+	github.com/caddyserver/caddy/v2 v2.4.0-beta.2
```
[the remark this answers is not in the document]
aledbf: fixed

chart/templates/proxy-configmap.yaml
```diff
@@ -0,0 +1,50 @@
...
+{{ $t := set . "username" (randAlphaNum 20) }}
+{{ $t := set . "password" (randAlphaNum 20) }}
+{{- end }}
+  vhost.docker-registry: |
+    https://minio.{$GITPOD_DOMAIN} {
+      import enable_log
+      import remove_server_header
+      import ssl_configuration
+
+      basicauth / bcrypt "Docker Registry" {
```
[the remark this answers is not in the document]
aledbf: @francislavoie thank you for the review. Fixed.

components/proxy/conf/Caddyfile
```diff
@@ -0,0 +1,261 @@
...
+			}
+		}
+	}
+}
+
+(disable_log) {
+	log {
+		output discard
+	}
+}
```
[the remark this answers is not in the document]
aledbf: @francislavoie thank you for the review. Fixed.

components/proxy/conf/Caddyfile
```diff
@@ -0,0 +1,261 @@
...
+
+(ssl_configuration) {
+	tls /etc/caddy/certificates/tls.crt /etc/caddy/certificates/tls.key {
+		#ca_root   <pem_file>
+	}
+}
+
+(upstream_headers) {
+	header_up Host {http.request.host}
```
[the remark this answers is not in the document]
aledbf: @francislavoie thank you for the review. Fixed.

akosyakov (approved):
works fast and stable for my use cases, but i did not check code

akosyakov:
@aledbf one thing which seems to be wrong it is how long does it take to download global.css with new PR in our website on slow 3g connection without caching again the prod. It is still slow compare to prod with recent changes.

components/proxy/Dockerfile
```diff
@@ -2,55 +2,11 @@
 # Licensed under the GNU Affero General Public License (AGPL).
 # See License-AGPL.txt in the project root for license information.
 
-FROM openresty/openresty:1.19.3.1-3-alpine
-
-ENV TRIGGER_REBUILD 1
+FROM aledbf/caddy-http2:0.5
```
csweichel, line 5: Can we build this caddy image as part of the Gitpod repo, or is this a fork of caddy?

components/proxy/conf/Caddyfile
```diff
@@ -0,0 +1,251 @@
...
+	respond /live 200
+	respond /ready 200
+}
+
+# TODO: is this being used?
+http://health*.{$GITPOD_DOMAIN} {
+	import remove_server_header
+
+	respond "alive"
+}
```
csweichel, line 93: not anymore. Leftover from the `cerc` days

components/proxy/plugins/workspace/download_workspace.go
```diff
@@ -0,0 +1,131 @@
...
+
+	url := fmt.Sprintf("%v%v%v", m.Service, r.URL.Path, query)
+	client := http.Client{Timeout: 5 * time.Second}
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return fmt.Errorf("Server Error: cannot download token OTS")
+	}
+
+	// pass browser headers
+	// TODO: check if it's possible to narrow the list
```
csweichel, line 58: nit
```suggestion
	// TODO(aledbf) check if it's possible to narrow the list
```

csweichel (approved):

components/proxy/Dockerfile
aledbf: This image is a fork of Caddy with the updated packages https://github.com/aledbf/caddy/commit/68fbb8eaa803b2b867fd79b230acea625e4619f4
I can do the same thing in the gitpod org, forking caddy and adding the commit, or maintain the code in my fork until the fix land in Caddy.

components/proxy/conf/Caddyfile
aledbf: fixed

components/proxy/plugins/workspace/download_workspace.go
aledbf: fixed

PR #3969, opened 2021-04-16
[code] support previous and current ubuntu LTS

### LINKED ISSUE #3832: code: support ubuntu lts as a baseline
### Bug description

I'd been having this issue during the VSCode beta too (but Theia still worked so I didn't worry too much about it).

### Steps to reproduce

Try to open the Source Control and GitHub tabs (see attached screenshots).

### Expected behavior

They work.

### Example repository

I'm not actually sure what is causing this bug to occur, I'm not having it with some repos but I have no idea what the root cause is. Could it be influenced by the choice of Docker image somehow? Git still works in the terminal.

akosyakov:
#### What it does

- fix #3832: support previous and current ubuntu LTS by building against 18.04 (previous) as a baseline

#### How to test

- In both cases that run, scm and ouput view for remote server channel are functional.
- test against ubuntu 18.04: https://akosyakov-code-support-ubuntu-lts-3832.staging.gitpod-dev.com/#https://gitlab.com/tonka3000/simple-cpp-abi-project
- test against ubuntu 20.04: https://akosyakov-code-support-ubuntu-lts-3832.staging.gitpod-dev.com/#https://github.com/gitpod-io/gitpod

### DIFF
--- a/components/ide/code/leeway.Dockerfile
+++ b/components/ide/code/leeway.Dockerfile
@@ -4,7 +4,10 @@
 
 # we use latest major version of Node.js distributed VS Code. (see about dialog in your local VS Code)
 # ideallay we should use exact version, but it has criticla bugs in regards to grpc over http2 streams
-FROM node:12.21.0 AS node_installer
+ARG NODE_VERSION=12.21.0
+
+
+FROM node:${NODE_VERSION} AS node_installer
 RUN mkdir -p /ide/node/bin \
     /ide/node/include/node/ \
     /ide/node/lib/node_modules/npm/ \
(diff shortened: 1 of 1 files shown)


markfirmware:
@akosyakov Will this provide something like

    image: gitpod/workspace-18.04-full-vnc

in .gitpod.yml?

I have a vnc use case that needs 18.04 and I would like to see the image maintained at gitpod. (My use case also requires a [small update](https://github.com/markfirmware/gtoolkit-screenshot/blob/7bc39d8d2a61aa03b793a8886b38cc90b43ec17d/mold#L29-L30) on top of 18.04 which is performed by a sudo in the workspace.)

akosyakov:
No, it is about building VS Code in such way that it can run on workspace images based on ubuntu 18.04 as well as 20.04. It is not about changing workspace images.

Could you file an issue for your request? I am not sure that it makes sense everything is moving to 20.04 and 18.04 won't be relevant in a year probably.

jankeromnes:
Taking a look now. 👀

components/ide/code/leeway.Dockerfile
```diff
@@ -19,17 +22,26 @@ RUN mkdir -p /ide/node/bin \
...
     && rm -rf /var/lib/apt/lists/*
 
+ARG NODE_VERSION
+ENV NVM_DIR /root/.nvm
+RUN curl -fsSL https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh | sh \
+    && . $NVM_DIR/nvm.sh  \
+    && nvm install $NODE_VERSION \
+    && nvm alias default $NODE_VERSION \
+    && npm install -g yarn node-gyp
+ENV PATH $NVM_DIR/versions/node/v$NODE_VERSION/bin:$PATH
```
jankeromnes, line 43: Question: Why re-install Node.js via NVM here (with `nvm.sh` being quite slow) instead of doing something like this?

```Dockerfile
COPY --from=node_installer /ide/node /ide/node
ENV PATH="/ide/node/bin:$PATH"
```

This is unimportant though, as this NVM/Node.js install doesn't seem to be captured in the final `FROM scratch` image below.

As long as this works it's good as is. 👍

jankeromnes (approved):
Code looks good and works as advertised. Many thanks @akosyakov! 💯 

I've aded a few questions below and in-line, but please feel free to merge as is. ✅ 

> In both cases that run, scm and ouput view for remote server channel are functional.

With both repos, scm works and output view for remote server channel seems to work. It does log a few errors, but I think they're unrelated. Still sharing them here FYI:

```
[2021-04-22 10:19:20.519] [main] [error] Extension 'hangxingliu.vscode-nginx-conf-hint' not found.
[2021-04-22 10:19:21.582] [main] [error] code server: failed to install intial extensions resolved by Code server: {}
[2021-04-22 10:19:40.242] [main] [error] Error: ENOENT: no such file or directory, stat '/ide/vs/workbench/contrib/output/common/outputLinkComputer.js'
```

Also, the window channel keeps logging this in a loop (also for both repos):

```
[2021-04-22 12:24:56.304] [window] [error] Not Found: WorkerScriptLoader.prototype.load/<@https://amaranth-rhinoceros-pxopwqbp.ws-dev.akosyakov-code-support-ubuntu-lts-3832.staging.gitpod-dev.com/out/vs/base/worker/workerMain.js#outputLinkComputer:707:31
```

components/ide/code/leeway.Dockerfile
akosyakov: Just did not think about it

components/ide/code/leeway.Dockerfile
```diff
@@ -4,7 +4,10 @@
 
 # we use latest major version of Node.js distributed VS Code. (see about dialog in your local VS Code)
 # ideallay we should use exact version, but it has criticla bugs in regards to grpc over http2 streams
-FROM node:12.21.0 AS node_installer
+ARG NODE_VERSION=12.21.0
```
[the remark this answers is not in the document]
akosyakov: I will bump it up with next vscode upgrade I. May, ok?

PR #3973, opened 2021-04-16
Don't log to stderr

meysholdt:
Don't pollute our logs with error message that are in fact debug messages.

### DIFF
--- a/components/registry-facade/pkg/registry/registry.go
+++ b/components/registry-facade/pkg/registry/registry.go
@@ -420,7 +420,7 @@ type dispatchFunc func(ctx context.Context, r *http.Request) http.Handler
 func dispatcher(d dispatchFunc) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		fc, _ := httputil.DumpRequest(r, false)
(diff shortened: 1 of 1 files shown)


components/registry-facade/pkg/registry/registry.go
```diff
@@ -420,7 +420,7 @@ type dispatchFunc func(ctx context.Context, r *http.Request) http.Handler
 func dispatcher(d dispatchFunc) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		fc, _ := httputil.DumpRequest(r, false)
-		fmt.Fprint(os.Stderr, string(fc))
+		log.Info(string(fc))
```
aledbf, line 423: Why not just change the writer? `fmt.Fprint(os.Stout, string(fc))`

csweichel: Or a more "traditional" structured log statement: `log.WithField("req", string(fc)).Debug("dispatching request")`

csweichel:
superseded by https://github.com/gitpod-io/gitpod/pull/4000

PR #3925, opened 2021-04-13
[docs] Generic doc update for v0.9.0

geropl:
Fixes #3893.

### Test
release: https://github.com/gitpod-io/gitpod/releases/tag/v0.9.0-alpha1
deployed to: https://gpl.gitpod-self-hosted.com/

PR #3977, opened 2021-04-16
Aws/ecr token regeneration

dimazaur:
AWS ECR token registration fix

1. fixing regenerate-ecr.tpl script Linux EOL
2. fixing regenerate-ecr.tpl script export env vars
3. adding AWS IAM user policy
4. adding K8s cluster role, resources secrets and serviceaccount
5. adding K8s cluster role binding
6. typo `esource`


aledbf:
@dimazaur please check https://github.com/gitpod-io/gitpod/pull/3925

We currently [do not support AWS as platform](https://www.notion.so/gitpod/1b9eac5cb33d42e391f86a87f0e37836?v=4f2ec7c943514ee19203b9d4fe096094) for running Gitpod.

kirooshu:
Probably fine not to support it officially but these fixes seems like a good idea regardless. Do you have plans to remove the `install/` directories for `aws`?

Issue #3898, opened 2021-04-12
[dashboard] New and w/o cxt? Redirect to www.gitpod.io


svenefftinge:
If you go to gitpod.io without any additional context or path and you are not logged-in, you should be redirected to www.gitpod.io.

FeixuRuins:
IMO `gitpod.io` should redirect to `app.gitpod.io` or some other subdomain to clear up possible confusion between the two.

svenefftinge:
The app is the more important place, especially with gitpod and its `gitpod.io/#` prefix. FWIW github and gitlab do it exactly like this as well.

AlexTugarev:
Just one thing to add: it should be `www.gitpod.io` in SaaS context, but `TBD` in self-hosted.

AlexTugarev:
We just agreed on handling this for SaaS for now. 🖖🏻 
Later on we can discuss if and how to solve this for Gitpod Self-Hosted.

svenefftinge:
For self-hosted it can stay as is, i.e. no redirect but the login screen.

ghuntley:
Some data from our community after I made a boo-boo whilst in a tweet:
- https://twitter.com/AnthonyBriggs/status/1382966449795342336?s=19
- https://twitter.com/candeira/status/1382970929437765635

ghuntley:
More data:

https://github.com/gitpod-io/website-kumquat/issues/377

### Bug description

The URL https://www.gitpod.io/docs (or any other direct link to documentation) cannot be accessed directly and results in a 500 error.
The workaround is going to any other page (like the homepage), and then click on 'Docs' on the menu.

### Steps to reproduce

Open https://www.gitpod.io/docs in your browser, you'll get a 500 error.
![image](https://user-images.githubusercontent.com/22901/114972137-e6fce000-9e4b-11eb-8821-bbb63b103929.png)


### Expected behavior

I expect to be able to open directly any page in documentation.

### Example repository

_No response_

PR #3991, opened 2021-04-19
[dashboard] redirect anonymous users to www.

AlexTugarev:
in SaaS context we want to redirect not logged in users from the root URL with no additional context provided to the website URL.

fixes https://github.com/gitpod-io/gitpod/issues/3898

### DIFF
--- a/components/dashboard/src/App.tsx
+++ b/components/dashboard/src/App.tsx
@@ -9,7 +9,7 @@ import Menu from './components/Menu';
 import { BrowserRouter } from "react-router-dom";
 import { Redirect, Route, Switch } from "react-router";
 
-import { Login } from './Login';
+import { Login, hasLoggedInBefore } from './Login';
 import { UserContext } from './user-context';
 import { getGitpodService } from './service/service';
 import { shouldSeeWhatsNew, WhatsNew } from './WhatsNew';
@@ -70,6 +70,12 @@ function App() {
             window.removeEventListener('storage', updateTheme);
         }
     }, [localStorage.theme]);
+
+    if (window.location.pathname === '/' && window.location.hash === '' && !hasLoggedInBefore()) {
+        if (window.location.host === "gitpod.io" || window.location.host === "gitpod-staging.com") {
+            window.location.href = `https://www.${window.location.host}`;
+        }
+    }
 
     if (loading) {
         return <Loading />


components/dashboard/src/App.tsx
```diff
@@ -71,6 +71,12 @@ function App() {
         }
     }, [localStorage.theme]);
 
+    if (window.location.pathname === '/' && window.location.hash === '' && !hasLoggedInBefore()) {
```
svenefftinge, line 74: Let's remove the `hasLoggedInBefore` so that you are landing on the website when you've logged out.

svenefftinge (approved):

components/dashboard/src/App.tsx
AlexTugarev: I just learned about the `markLoggedIn` which sets a HTTP cookie `gitpod-user=loggedIn`. 

But we already have a session cookie `gitpod-user=loggedIn`, which is managed on login/logout.

AlexTugarev: `hasLoggedInBefore` checks for a Cookie which is never removed, `gitpod-user=loggedIn` used to reflect the session state. 

WDYT of renaming the new Cookie, so it reflects that particular state? maybe something like "gitpod-logged-in-before". Maybe we can even set the provider host as value ;-)

svenefftinge: The cookie should be called 'seen-gitpod' or so, as it just indicates the user knows gitpod (has seen the website or the app). Anyhow, we don't want to use it here, as the login state is enough. We really should redirect everyone to www once they are logged out.

components/dashboard/src/App.tsx
```diff
@@ -71,6 +71,12 @@ function App() {
...
+        if (window.location.host === "gitpod.io" || window.location.host === "gitpod-staging.com") {
```
ghuntley, line 75: **nit:** in development the URL is https://at-redirect-www.staging.gitpod-dev.com which won't match this logic.

components/dashboard/src/App.tsx
AlexTugarev: > Anyhow, we don't want to use it here, as the login state is enough. We really should redirect everyone to www once they are logged out.

Roger that. That thing is, that the "login state" is fetched asynchronously. So, removing of the `gitpod-user=loggedIn` (session) cookie check, would mean that the redirect might happen in any case.

components/dashboard/src/App.tsx
AlexTugarev: true. in the dashboard component we still don't have a proper solution for the URL mismatch in preview environments.

components/dashboard/src/App.tsx
AlexTugarev: This is what used to work before: https://github.com/gitpod-io/gitpod/blob/24f7b609bf79b247d72d1841282639726bcd5891/components/dashboard/src/components/create/index.tsx#L56

components/dashboard/src/App.tsx
jankeromnes: Maybe something like this?

```suggestion
        if (["gitpod.io", "gitpod-staging.com", "gitpod-dev.com", "gitpod-io-dev.com"].some(suffix => window.location.host.endsWith(suffix))) {
```

svenefftinge:
continuing here #4070

Issue #3777, opened 2021-04-06
"Continue with Default Image" button doesn't work when build is still in progress


jankeromnes:
### Bug description

When creating a workspace, and it runs a Docker build, you can not skip it while the build is still running (`Continue with Default Image` button does nothing).

However, if the build fails, then the button works.

### Steps to reproduce

- Open https://github.com/jankeromnes/gitpod-build-error
- While the build is in progress, click on `Continue with Default Image`. It doesn't work.
- When the build fails, then the button works.

### Expected behavior

It should work in both cases.

jankeromnes:
Potentially relevant: The `server` pod seems to log this whenever I click on `Continue with Default Image` while the build is still running:

```
{"component":"server","severity":"DEBUG","time":"2021-04-12T10:19:44.150Z","environment":"devstaging","region":"europe-west1","message":"attempted to fetch workspace cookie, but instance has no owner token","payload":{"instanceId":"e319478c-f1be-4dbb-a688-d47ca4848408","userId":"359092b8-2ea8-469a-9303-57749a22eec2"}}
```

PR #3993, opened 2021-04-19
Allow restarting workspaces with forceDefaultImage=true even when another instance is already running

jankeromnes:
Fixes https://github.com/gitpod-io/gitpod/issues/3777

How to test:

- Open https://jx-restart-w-default-image.staging.gitpod-dev.com/#https://github.com/jankeromnes/gitpod-hanging-build (and log in if needed)
- The Docker build will call `sleep 1h`
- But, you should still be able to click on "Continue with Default Image" at any time & get a running workspace

### DIFF
--- a/components/server/src/workspace/gitpod-server-impl.ts
+++ b/components/server/src/workspace/gitpod-server-impl.ts
@@ -396,17 +396,21 @@ export class GitpodServerImpl<Client extends GitpodClient, Server extends Gitpod
             const user = this.checkAndBlockUser();
             await this.checkTermsAcceptance();
 
-            log.info({ userId: user.id, workspaceId }, 'startWorkspace');
+            const logCtx = { userId: user.id, workspaceId };
+            log.info(logCtx, 'startWorkspace');
 
             const mayStartPromise = this.mayStartWorkspace({ span }, user, this.workspaceDb.trace({ span }).findRegularRunningInstances(user.id));
             const workspace = await this.internalGetWorkspace(workspaceId, this.workspaceDb.trace({ span }));
             await this.guardAccess({ kind: "workspace", subject: workspace }, "get");
 
             const runningInstance = await this.workspaceDb.trace({ span }).findRunningInstance(workspace.id);
-            if (runningInstance) {
-                // We already have a running workspace.
-                // Note: ownership doesn't matter here as this is basically a noop. It's not StartWorkspace's concern
-                //       to guard workspace access - just to prevent non-owners from starting workspaces.
+            if (runningInstance && !options.forceDefaultImage) {
+                // We already have a running workspace, and we're not forcing the default image.
+                // Notes:
+                //   * ownership doesn't matter here as this is basically a noop. It's not StartWorkspace's concern
+                //     to guard workspace access - just to prevent non-owners from starting workspaces.
+                //   * forceDefaultImage now always interrupts a previously-running instance, even if the previous
+                //     instance already had options.forceDefaultImage === true.
 
                 await this.guardAccess({ kind: "workspaceInstance", subject: runningInstance, workspaceOwnerID: workspace.ownerId, workspaceIsShared: workspace.shareable || false }, "get");
                 return {
@@ -432,6 +436,14 @@ export class GitpodServerImpl<Client extends GitpodClient, Server extends Gitpod
             const envVars = this.userDB.getEnvVars(user.id);
 
             await mayStartPromise;
+
+            if (runningInstance) {
+                // We already had a running workspace. This may happen if we're forcing the default image.
+                // In that case, we first stop the previous workspace, and wait for it to be completely gone.
+                await this.internalStopWorkspaceAndWaitForInstance({ span }, workspaceId, workspace.ownerId).catch(err => {
+                    log.error(logCtx, "internalStopWorkspaceAndWaitForInstance error: ", err);
+                });
+            }
 
             // at this point we're about to actually start a new workspace
             return await this.workspaceStarter.startWorkspace({ span }, workspace, user, await envVars, {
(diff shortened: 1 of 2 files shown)


jankeromnes:
@csweichel Could you please take a careful look? (I really don't want to introduce new bugs in the workspace start logic.)

Note: This change seems to work well / as expected. However, I get no Docker build logs on dev-staging -- not sure why.

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -396,15 +396,16 @@ export class GitpodServerImpl<Client extends GitpodClient, Server extends Gitpod
...
 
             const mayStartPromise = this.mayStartWorkspace({ span }, user, this.workspaceDb.trace({ span }).findRegularRunningInstances(user.id));
             const workspace = await this.internalGetWorkspace(workspaceId, this.workspaceDb.trace({ span }));
             await this.guardAccess({ kind: "workspace", subject: workspace }, "get");
 
             const runningInstance = await this.workspaceDb.trace({ span }).findRunningInstance(workspace.id);
-            if (runningInstance) {
-                // We already have a running workspace.
+            if (runningInstance && !options.forceDefaultImage) {
+                // We already have a running workspace, and we're not forcing the default image.
```
jankeromnes, line 408: It would be great to add a comment explaining that:
- `forceDefaultImage` now also "forces instance restart" (behavior change from this PR)
- it will also interrupt & restart an instance even if that instance already had `forceDefaultImage: true`

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -433,6 +434,14 @@ export class GitpodServerImpl<Client extends GitpodClient, Server extends Gitpod
 
             await mayStartPromise;
 
+            if (runningInstance) {
+                // We already have a running workspace. This may happen if we're forcing the default image.
+                // In that case, we stop the previous first.
+                await this.internalStopWorkspace({ span }, workspaceId, workspace.ownerId).catch(err => {
+                    log.error(logCtx, "stopWorkspace error: ", err);
+                });
```
jankeromnes, line 442: This call only blocks until the stop request is sent.

In order to completely prevent multiple instances from running at the same time, we should probably also listen for instance updates here (e.g. via [messageBusIntegration.listenForWorkspaceInstanceUpdates()](https://github.com/gitpod-io/gitpod/blob/ec9a1922fd09fb195d5c2323c42a48b06169d051/components/server/src/workspace/gitpod-server-impl.ts#L145)) and only move on when the previous instance is definitely stopped.

There is also a case where the previous instance was in a different region, in which case to be fully pedantic we should also poll the DB, but it's probably okay to not do that here.

jankeromnes (commented):
Adding review comments from a call with @csweichel here for the record (many thanks Chris!)

jankeromnes:
Still seems to work as expected.

Two open questions:

1. Shouldn't this access guard fail:

https://github.com/gitpod-io/gitpod/blob/27056fe4481971f5c2eb2dbeb4e84f43e526fee3/components/server/src/workspace/gitpod-server-impl.ts#L426-L427

given that the previously running instance is only stopped afterwards?

https://github.com/gitpod-io/gitpod/blob/27056fe4481971f5c2eb2dbeb4e84f43e526fee3/components/server/src/workspace/gitpod-server-impl.ts#L440-L446

2. Also, an observed error logged by server:

```
{"@type":"type.googleapis.com/google.devtools.clouderrorreporting.v1beta1.ReportedErrorEvent","serviceContext":{"service":"server","version":"jx-restart-w-default-image.3"},"stack_trace":"Error: Unknown workspace manager \"\"\n    at WorkspaceManagerClientProvider.<anonymous> (/app/node_modules/@gitpod/ws-manager/lib/client-provider.js:123:51)\n    at step (/app/node_modules/@gitpod/ws-manager/lib/client-provider.js:58:23)\n    at Object.next (/app/node_modules/@gitpod/ws-manager/lib/client-provider.js:39:53)\n    at fulfilled (/app/node_modules/@gitpod/ws-manager/lib/client-provider.js:30:58)\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejections (internal/process/task_queues.js:97:5)","component":"server","severity":"ERROR","time":"2021-04-23T08:46:17.766Z","environment":"devstaging","region":"europe-west1","context":{"userId":"7792a3aa-7416-4dee-a899-e132467125d4","workspaceId":"black-goldfish-rtllylsu"},"message":"internalStopWorkspaceAndWaitForInstance error: ","error":"Error: Unknown workspace manager \"\"\n    at WorkspaceManagerClientProvider.<anonymous> (/app/node_modules/@gitpod/ws-manager/lib/client-provider.js:123:51)\n    at step (/app/node_modules/@gitpod/ws-manager/lib/client-provider.js:58:23)\n    at Object.next (/app/node_modules/@gitpod/ws-manager/lib/client-provider.js:39:53)\n    at fulfilled (/app/node_modules/@gitpod/ws-manager/lib/client-provider.js:30:58)\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejections (internal/process/task_queues.js:97:5)"}
```

Prettified error stack:

```
Error: Unknown workspace manager ""
    at WorkspaceManagerClientProvider.<anonymous> (/app/node_modules/@gitpod/ws-manager/lib/client-provider.js:123:51)
    at step (/app/node_modules/@gitpod/ws-manager/lib/client-provider.js:58:23)
    at Object.next (/app/node_modules/@gitpod/ws-manager/lib/client-provider.js:39:53)
    at fulfilled (/app/node_modules/@gitpod/ws-manager/lib/client-provider.js:30:58)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

csweichel:
> Code review feedback from @csweichel implemented ✅ and still works as expected. 🎉
> 
> Two questions:
> 
> 1. Shouldn't this access guard fail:
> 
> https://github.com/gitpod-io/gitpod/blob/27056fe4481971f5c2eb2dbeb4e84f43e526fee3/components/server/src/workspace/gitpod-server-impl.ts#L426-L427

No. It only checks if the user has the permission to execute that operation, but not if the operation makes sense.

> 
> given that the previously running instance is only stopped afterwards?
> 
> https://github.com/gitpod-io/gitpod/blob/27056fe4481971f5c2eb2dbeb4e84f43e526fee3/components/server/src/workspace/gitpod-server-impl.ts#L440-L446
> 
> 1. Also, an observed error logged by server:
> 
> Prettified error stack:
> 
> ```
> Error: Unknown workspace manager ""
>     at WorkspaceManagerClientProvider.<anonymous> (/app/node_modules/@gitpod/ws-manager/lib/client-provider.js:123:51)
>     at step (/app/node_modules/@gitpod/ws-manager/lib/client-provider.js:58:23)
>     at Object.next (/app/node_modules/@gitpod/ws-manager/lib/client-provider.js:39:53)
>     at fulfilled (/app/node_modules/@gitpod/ws-manager/lib/client-provider.js:30:58)
>     at runMicrotasks (<anonymous>)
>     at processTicksAndRejections (internal/process/task_queues.js:97:5)
> ```
> 
> Unsure if related or serious. (Could it be that's because we have no GCP region information in core-dev?)

The problem here is that the instance wasn't actually started yet, i.e. it didn't go through `actuallyStartWorkspace` yet. In `internalStopWorkspaceAndWaitForInstance` you could check if the `region` is set, and if not instead of calling ws-manager just wait for some time - you would not know which workspace manager to talk to after all. There's a chance for a race condition here (some other `server` instance could be starting the instance currently), so in lieu of proper cross-server-instance-locking waiting for say 10 seconds (please add a comment) should prevent that race.

jankeromnes:
> No. It only checks if the user has the permission to execute that operation, but not if the operation makes sense.

Aha, that makes sense. Thanks!

> The problem here is that the instance wasn't actually started yet, i.e. it didn't go through `actuallyStartWorkspace` yet.

Thanks for the pointer! Looking into `actuallyStartWorkspace`, I now think it's expected that instances don't get assigned a `region` until after their Docker build is complete, as:

https://github.com/gitpod-io/gitpod/blob/2e7cba93f1cb8702f5c8601ee13ff1b1696df4db/components/server/src/workspace/workspace-starter.ts#L132-L133

happens before:

https://github.com/gitpod-io/gitpod/blob/2e7cba93f1cb8702f5c8601ee13ff1b1696df4db/components/server/src/workspace/workspace-starter.ts#L156-L159

So in our case (long-running Docker build that we want to interrupt/skip), we should expect not to have a `region`.

How should we stop "preparing" instances that don't have a `region` yet? Should we make imageBuilder interruptible? Or "detach" the running instance by deleting its workspace ID? 🤔

jankeromnes:
After discussing this further, we've decided to not make Docker builds interruptible just yet, and instead remove the button when builds are in progress (to align with the previous dashboard design).

Superseded by https://github.com/gitpod-io/gitpod/pull/4104

Issue #3920, opened 2021-04-13
[self-hosted] UI: Re-instantiate first-steps/initial-setup page


geropl:
This is the first page you see when you visit https://your-gitpod-installation.com after installing it. It needs to accomplish two things:
 - get the installers consent on the terms
 - guide them to properly configure a Git Hoster (GitHub/Gitlab)

Fixes:
 - https://github.com/gitpod-io/gitpod/issues/3904
 - #3705

/cc @gtsiolis @jankeromnes

Issue #3904, opened 2021-04-12
[self-hosted] Empty login options


aledbf:
### Bug description

If there are no `authProviders` configured, the login doesn't show any action

![Screenshot from 2021-04-12 17-13-12](https://user-images.githubusercontent.com/161571/114463631-ac2c4b00-9bb2-11eb-9f47-97768d80047a.png)

### Steps to reproduce

-

### Expected behavior

A message indicating that there is no configured authentication provider?

### Example repository

-

AlexTugarev:
We need to reinstate a setup page for this case. It used to render the Git Provider component to allow you to setup and connect with a IdP. This is also the way the first user is created. 

cf. #3893

Issue #3705, opened 2021-04-01
[dashboard] Handle SETUP_REQUIRED


svenefftinge:
For self-hosted the server fires `SETUP_REQUIRED` errors when there is no git hoster registered.
We need to handle this case in the new dashboard.

PR #3995, opened 2021-04-19
[dashboard/self-hosted] add Setup page

AlexTugarev:
## how to test

1. either use https://at-setup.staging.gitpod-dev.com/ or create a branch from this PR
   it should start out with a `Welcome to Gitpod` modal.
2. create an OAuth Application on gitlab.com (built-in auth providers are disabled in https://github.com/gitpod-io/gitpod/commit/e5c443a05c2b11ac597c6d6f8406eca02ae65db9)
3. use `https://staging.gitpod-dev.com/<HOST>/callback` as `redirectURI` on  GitHub/GitLab.
4. follow the instructions of the New Git Integration modal.
5. expect to see the regular dashboard once you finished the setup of a git integration.
5. be sure to test the environment exclusively. remember, the setup process is a single user thing.

Screen Shot 2021-04-21 at 12 41 15

Screen Shot 2021-04-21 at 12 41 20

closes https://github.com/gitpod-io/gitpod/issues/3705
closes https://github.com/gitpod-io/gitpod/issues/3904
closes https://github.com/gitpod-io/gitpod/issues/3920

### DIFF
--- a/components/dashboard/src/Setup.tsx
+++ b/components/dashboard/src/Setup.tsx
@@ -1,0 +1,61 @@
+/**
+ * Copyright (c) 2021 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { useEffect, useState } from "react";
+import Modal from "./components/Modal";
+import { getGitpodService, gitpodHostUrl } from "./service/service";
+import { GitIntegrationModal } from "./settings/Integrations";
+
+export default function Setup() {
+
+    const [showModal, setShowModal] = useState<boolean>(false);
+
+    useEffect(() => {
+        (async () => {
+            const dynamicAuthProviders = await getGitpodService().server.getOwnAuthProviders();
+            const previous = dynamicAuthProviders.filter(ap => ap.ownerId === "no-user")[0];
+            if (previous) {
+                await getGitpodService().server.deleteOwnAuthProvider({ id: previous.id });
+            }
+        })();
+    }, []);
+
+    const acceptAndContinue = () => {
+        setShowModal(true);
+    }
+
+    const onAuthorize = (payload?: string) => {
+        // run without await, so the integrated closing of new tab isn't blocked
+        (async () => {
+            window.location.href = gitpodHostUrl.asDashboard().toString();
+        })();
+    }
+
+    const headerText = "Configure a git integration with a GitLab or GitHub instance."
+
+    return <div>
+        {!showModal && (
+            <Modal visible={true} onClose={() => { }} closeable={false}>
+                <h3 className="pb-2">Welcome to Gitpod 🎉</h3>
+                <div className="border-t border-b border-gray-200 dark:border-gray-800 mt-2 -mx-6 px-6 py-4">
+                    <p className="pb-4 text-gray-500 text-base">To start using Gitpod, you will need to set up a git integration.</p>
+
+                    <div className="flex">
+                        <span className="text-gray-500">
+                            By using Gitpod, you agree to our <a className="underline underline-thickness-thin underline-offset-small hover:text-gray-600" target="gitpod-terms" href="https://www.gitpod.io/self-hosted-terms/">terms</a>.
+                        </span>
+                    </div>
+                </div>
+                <div className="flex justify-end mt-6">
+                    <button className={"ml-2"} onClick={() => acceptAndContinue()}>Continue</button>
+                </div>
+            </Modal>
+        )}
+        {showModal && (
+            <GitIntegrationModal mode="new" login={true} headerText={headerText} userId="no-user" onAuthorize={onAuthorize} />
+        )}
+    </div>;
+}
--- a/components/dashboard/src/provider-utils.tsx
+++ b/components/dashboard/src/provider-utils.tsx
@@ -41,7 +41,7 @@ interface OpenAuthorizeWindowParams {
     scopes?: string[];
     overrideScopes?: boolean;
     onSuccess?: (payload?: string) => void;
-    onError?: (error?: string) => void;
+    onError?: (error: string | { error: string, description?: string }) => void;
 }
 
 async function openAuthorizeWindow(params: OpenAuthorizeWindowParams) {
@@ -75,12 +75,21 @@ async function openAuthorizeWindow(params: OpenAuthorizeWindowParams) {
 
         if (typeof event.data === "string" && event.data.startsWith("success")) {
             killAuthWindow();
-            onSuccess && onSuccess();
+            onSuccess && onSuccess(event.data);
         }
         if (typeof event.data === "string" && event.data.startsWith("error:")) {
-            const errorAsText = atob(event.data.substring("error:".length));
+            let error: string | { error: string, description?: string } = atob(event.data.substring("error:".length));
+            try {
+                const payload = JSON.parse(error);
+                if (typeof payload === "object" && payload.error) {
+                    error = { error: payload.error, description: payload.description };
+                }
+            } catch (error) {
+                console.log(error);
+            }
+
             killAuthWindow();
-            onError && onError(errorAsText);
+            onError && onError(error);
         }
     };
     window.addEventListener("message", eventListener);
--- a/components/dashboard/src/settings/Integrations.tsx
+++ b/components/dashboard/src/settings/Integrations.tsx
@@ -38,7 +38,7 @@ function GitProviders() {
 
     const [authProviders, setAuthProviders] = useState<AuthProviderInfo[]>([]);
     const [allScopes, setAllScopes] = useState<Map<string, string[]>>(new Map());
-    const [diconnectModal, setDisconnectModal] = useState<{ provider: AuthProviderInfo } | undefined>(undefined);
+    const [disconnectModal, setDisconnectModal] = useState<{ provider: AuthProviderInfo } | undefined>(undefined);
     const [editModal, setEditModal] = useState<{ provider: AuthProviderInfo, prevScopes: Set<string>, nextScopes: Set<string> } | undefined>(undefined);
     const [selectAccountModal, setSelectAccountModal] = useState<SelectAccountPayload | undefined>(undefined);
 
@@ -232,18 +232,18 @@ function GitProviders() {
             <SelectAccountModal {...selectAccountModal} close={() => setSelectAccountModal(undefined)} />
         )}
 
-        {diconnectModal && (
+        {disconnectModal && (
             <Modal visible={true} onClose={() => setDisconnectModal(undefined)}>
                 <h3 className="pb-2">Disconnect Provider</h3>
                 <div className="border-t border-b border-gray-200 dark:border-gray-800 mt-2 -mx-6 px-6 py-4">
                     <p className="pb-4 text-gray-500 text-base">Are you sure you want to disconnect the following provider?</p>
                     <div className="flex flex-col rounded-xl p-3 bg-gray-100 dark:bg-gray-800">
-                        <div className="text-gray-700 text-md font-semibold">{diconnectModal.provider.authProviderType}</div>
-                        <div className="text-gray-400 text-md">{diconnectModal.provider.host}</div>
+                        <div className="text-gray-700 text-md font-semibold">{disconnectModal.provider.authProviderType}</div>
+                        <div className="text-gray-400 text-md">{disconnectModal.provider.host}</div>
                     </div>
                 </div>
                 <div className="flex justify-end mt-6">
-                    <button className={"ml-2 danger secondary"} onClick={() => disconnect(diconnectModal.provider)}>Disconnect Provider</button>
+                    <button className={"ml-2 danger secondary"} onClick={() => disconnect(disconnectModal.provider)}>Disconnect Provider</button>
                 </div>
             </Modal>
         )}
@@ -358,10 +358,10 @@ function GitIntegrations() {
     return (<div>
 
         {modal?.mode === "new" && (
-            <GitIntegrationModal mode={modal.mode} userId={user?.id || "no-user"} onClose={() => setModal(undefined)} update={updateOwnAuthProviders} />
+            <GitIntegrationModal mode={modal.mode} userId={user?.id || "no-user"} onClose={() => setModal(undefined)} onUpdate={updateOwnAuthProviders} />
         )}
         {modal?.mode === "edit" && (
-            <GitIntegrationModal mode={modal.mode} userId={user?.id || "no-user"} provider={modal.provider} onClose={() => setModal(undefined)} update={updateOwnAuthProviders} />
+            <GitIntegrationModal mode={modal.mode} userId={user?.id || "no-user"} provider={modal.provider} onClose={() => setModal(undefined)} onUpdate={updateOwnAuthProviders} />
         )}
         {modal?.mode === "delete" && (
             <Modal visible={true} onClose={() => setModal(undefined)}>
@@ -397,7 +397,7 @@ function GitIntegrations() {
             <div className="w-full flex h-80 mt-2 rounded-xl bg-gray-100 dark:bg-gray-900">
                 <div className="m-auto text-center">
                     <h3 className="self-center text-gray-500 dark:text-gray-400 mb-4">No Git Integrations</h3>
-                    <div className="text-gray-500 mb-6">In addition to the default Git Providers you can authorize<br /> with a self hosted instace of a provider.</div>
+                    <div className="text-gray-500 mb-6">In addition to the default Git Providers you can authorize<br /> with a self-hosted instace of a provider.</div>
                     <button className="self-center" onClick={() => setModal({ mode: "new" })}>New Integration</button>
                 </div>
             </div>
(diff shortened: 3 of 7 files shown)


gtsiolis:
Looking at this now! 👀

components/dashboard/src/settings/Integrations.tsx
```diff
@@ -580,10 +597,10 @@ function GitIntegrationModal(props: ({
         }
     };
 
-    return (<Modal visible={!!props} onClose={close}>
-        <h3 className="pb-2">{props.mode === "new" ? "New Git Integration" : "Git Integration"}</h3>
+    return (<Modal visible={!!props} onClose={close} closeable={props.closeable}>
+        <h3 className="pb-2">{mode === "new" ? "New Git Integration" : "Git Integration"}</h3>
         <div className="space-y-4 border-t border-b border-gray-200 dark:border-gray-800 mt-2 -mx-6 px-6 py-4">
```
gtsiolis, line 602: **issue**: Just tried this in a smaller screen and realized we're going to need some scrolling overflow behavior to fit the modal. Currently, this slightly crops the modal on the bottom but should be ok for this iteration. 

| Modal Crop |
|-|
| image |

components/dashboard/src/Setup.tsx
```diff
@@ -0,0 +1,59 @@
...
+            window.location.href = gitpodHostUrl.asDashboard().toString();
+        })();
+    }
+
+    return <div>
+        {!showModal && (
+            <Modal visible={true} onClose={() => { }} closeable={false}>
+                <h3 className="pb-2">ALLMOST DONE</h3>
+                <div className="border-t border-b border-gray-200 dark:border-gray-800 mt-2 -mx-6 px-6 py-4">
+                    <p className="pb-4 text-gray-500 text-base">To complete the setup of Gitpod, you need to create a Git Integration.</p>
```
gtsiolis, line 42: **suggestion**: What do you think of the following?

```suggestion
                    <p className="pb-4 text-gray-500 text-base">To start using Gitpod, you will need to set up a Git Integration.</p>
```

components/dashboard/src/Setup.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+                    <div className="flex">
+                        <span className="text-gray-500">
+                            By continuing, you agree to our <a className="underline underline-thickness-thin underline-offset-small hover:text-gray-600" target="gitpod-terms" href="https://www.gitpod.io/self-hosted-terms/">terms</a>.
```
gtsiolis, line 46: **suggestion**: Minor rephrasing to match the one in our login page at Gitpod.io.

```suggestion
                            By using Gitpod, you agree to our <a className="underline underline-thickness-thin underline-offset-small hover:text-gray-600" target="gitpod-terms" href="https://www.gitpod.io/self-hosted-terms/">terms of service</a>.
```

gtsiolis (commented):
Hey @AlexTugarev! This is dope. 🌟 

Left some minor comments below, let me know what you think!

**question**: Does this only work with a self-hosted Gitlab installation? 
1. I couldn't set up a GitLab integration using GitLab.com. I was getting the following errors:
    - First time, I got redirected to `/sorry ...`.
    - Second time, I run into: `The redirect URI included is not valid.`
2. I couldn't set up a GitHub integration using GitHub.com. Go redirected to an empty page with a URL like `/complete-auth?message=error:eyJlcnJvciI6 ...`

---

Regarding UX, there's a lot of room for improvements on the first screen but this should be fine for this iteration. For example, we could use the first screen to promote a couple of features or prompt for an upgrade to a paid tier.

components/dashboard/src/settings/Integrations.tsx
```diff
@@ -592,10 +609,10 @@ function GitIntegrationModal(props: ({
             <div className="flex flex-col">
                 <span className="text-gray-500">Configure a git integration with a GitLab or GitHub self-hosted instance.</span>
```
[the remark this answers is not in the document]
AlexTugarev: Good catch! Hm. If setup is required we can assume there is no collision to expect.
I'll make that dynamic.

components/dashboard/src/Setup.tsx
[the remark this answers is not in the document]
AlexTugarev: Yes. https://github.com/gitpod-io/website/issues/80

AlexTugarev: ok. 

but why is this a service in case of self-hosted?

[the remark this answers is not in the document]
gtsiolis: **thought**: If we don't link these self-hosted terms from elsewhere, it would be nice to embed this here in the first step in a scrolling section. However, let's leave this as is for this iteration.

AlexTugarev:
@gtsiolis, I just realized, you didn't went through the flow, because the `hot to test` from above was incomplete. 

For preview environment we need to remove the k8s namespace from the redirect URL when testing, so it becomes: `https://staging.gitpod-dev.com/<HOST>/callback`

components/dashboard/src/Setup.tsx
gtsiolis: Ah, good point! I think this refers to the Gitpod as a service being installed. However, I could be wrong. Feel free to leave this as is if you think this could create more confusion.

gtsiolis:
> For preview environment ...

Ah, ok. Trying again now. ➿

components/dashboard/src/Setup.tsx
gtsiolis: FWIW, here are the [previous self hosted terms](https://github.com/gitpod-io/retired-gatsby-website/blob/master/src/pages/self-hosted-terms.tsx). 📖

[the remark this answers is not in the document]
gtsiolis: FWIW, here are the [previous self hosted terms](https://github.com/gitpod-io/retired-gatsby-website/blob/master/src/pages/self-hosted-terms.tsx). 📖

AlexTugarev:
@gtsiolis, please ping me if you need help with the test setup. 
I've already tackled your comments but as it would interrupt your testing, I'm waiting to push.

gtsiolis (commented):
Alright! /cc @AlexTugarev 

1. GitLab integration works as expected. ✅ 
2. GitHub integration returns to the New Integration modal and prompts the user to activate the integration. :x:

Other notes:
1. The integration is being added on the Git Providers list regardless if the integration activation failed.
2. The integration is being exposed to the login options regardless id the integration activation failed.

**question**: Is it not possible to add a Bitbucket integration on a self hosted Gitpod instance?

AlexTugarev:
> GitLab integration works as expected. ✅

Thanks for verifying @gtsiolis!

> GitHub integration returns to the New Integration modal and prompts the user to activate the integration. ❌

Could you explain how you did that? 
After the initial setup, i.e. creating & authorizing with git/identity provider, you should be redirected to the dashboard. 

Once the setup is done, you should not be able to see the Setup page again.

> The integration is being added on the Git Providers list regardless if the integration activation failed.

That should only be true for the owner.

> The integration is being exposed to the login options regardless id the integration activation failed.

That should never be the case.

gtsiolis:
@AlexTugarev when I clicked _Activate Integration_ after entering the Client ID and Client Secret I got redirect back to the same modal.

On the GH side, maybe I did something wrong but I tried entering both of the following as Homepage URLs:
1. `https://at-setup.staging.gitpod-dev.com/`
2. `https://staging.gitpod-dev.com/`

---

Regarding, the login page this is what I see when the GH activation fails on `/integrations` and `/login`:

| Integrations | Login |
|-|-|
| image | image |

AlexTugarev:
@gtsiolis, I just checked with GItHub OAuth apps for gitpod-dev environments.

If you chose to integrate with `github.com` OAuth app of yours, please configure the following `redirect URI`. Note that the k8s namespace is missing.
`https://staging.gitpod-dev.com/auth/github.com/callback`

There is a reason behind that change of `redirect URI` and it only affects preview environments. In theory you could just reuse the suggested URL from that modal, but we have a special case, in which we reuse a single OAuth app for _all_ preview deployments, thus need to distribute the requested based on k8s in the upstream proxy for preview environments.

From https://github.com/gitpod-io/gitpod/pull/3995#issuecomment-822526861, I can learn two things: 
1. the urls might be wrong
2. the error OAuth error message should be integrated into the integrations modal, the same way as we now do for the signup

AlexTugarev:
> Regarding UX, there's a lot of room for improvements on the first screen but this should be fine for this iteration. For example, we could use the first screen to promote a couple of features or prompt for an upgrade to a paid tier.

@gtsiolis that's two separate thing, right? feel free to make suggestions. I was counting on you to get some input for the rough draft, which is basically made after the What's New modal.

gtsiolis:
Taking a look a this, round two! 🥊

components/dashboard/src/Setup.tsx
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    const onAuthorize = (payload?: string) => {
+        // run without await, so the integrated closing of new tab isn't blocked
+        (async () => {
+            window.location.href = gitpodHostUrl.asDashboard().toString();
+        })();
+    }
+
+    const headerText = "Configure a git integration with a GitLab or GitHub."
```
gtsiolis, line 37: **nitpick**: Minor rephrasing.

```suggestion
    const headerText = "Configure a git integration with a GitLab or GitHub instance."
```

Alternatively, this we could just drop the `a`. You call! 📞 

```suggestion
    const headerText = "Configure a git integration with GitLab or GitHub."
```

components/dashboard/src/settings/Integrations.tsx
```diff
@@ -580,63 +612,61 @@ function GitIntegrationModal(props: ({
...
                     <span className="text-red-600 dark:text-red-400 hover:text-red-800 dark:hover:text-red-300">You need to activate this integration.</span>
                 </div>
             )}
             <div className="flex flex-col">
-                <span className="text-gray-500">Configure a git integration with a GitLab or GitHub self-hosted instance.</span>
+                <span className="text-gray-500">{props.headerText || "Configure a git integration with a GitLab or GitHub self-hosted instance."}</span>
             </div>
-            {props.mode === "new" && (
+
+            <div className="overscroll-contain max-h-96 overflow-y-auto">
```
gtsiolis, line 628: **praise**: Thanks for adding the scrolling container here!

gtsiolis, line 628: **suggestion**: What do you think of adding a small right padding here to avoid collapsing the scroll bar on top of the inoput elements? It did [something similar](https://github.com/gitpod-io/gitpod/blob/ce05f58c44d9d8ff4cf2a78e23a31b7571d42454/components/dashboard/src/workspaces/StartWorkspaceModal.tsx#L61) with the New Workspace modal (Recent/Examples).

```suggestion
            <div className="overscroll-contain max-h-96 overflow-y-auto pr-2">
```

components/dashboard/src/settings/Integrations.tsx
```diff
@@ -395,7 +395,7 @@ function GitIntegrations() {
             <div className="w-full flex h-80 mt-2 rounded-xl bg-gray-100 dark:bg-gray-900">
                 <div className="m-auto text-center">
                     <h3 className="self-center text-gray-500 dark:text-gray-400 mb-4">No Git Integrations</h3>
-                    <div className="text-gray-500 mb-6">In addition to the default Git Providers you can authorize<br /> with a self hosted instace of a provider.</div>
+                    <div className="text-gray-500 mb-6">In addition to the default Git Providers you can authorize<br /> with a self-hosted instace of a provider.</div>
                     <button className="self-center" onClick={() => setModal({ mode: "new" })}>New Integration</button>
```
gtsiolis, line 399: **issue**: I'm still getting the same behavior when trying to register a GH integration with GitHub.com. Maybe I'm doing something wrong. Pasting below some screenshots in case these help. 😭 

| OAuth Application | Modal |
|-|-|
| image | image |

gtsiolis (approved):
Thanks for all the updates here @AlexTugarev! 🥊 

We're almost there! I still can't get the GH integration on but maybe it's because I'm missing something. Let me know what you think. 💭

gtsiolis:
FYI, I'm also working on improving the first step (currently a modal). Will open a follow-up issue once I have something ready-for-development. 🎨

components/dashboard/src/settings/Integrations.tsx
AlexTugarev: Is this screenshot of the modal from `/integrations` or from the initial setup. 
From the logs, it looks like you went through the setup and created the gitlab.com integration. Then you went to the /integrations and tried to setup github.com.

I cannot find any error in the logs. Actually, the github.com integration seems to be ok. Perhaps it's an issue with a missing update in the frontend. 

Let me check this in a new deployment. 

BTW, I always tested setup with both providers by deleting the account in between, to make sure it works both ways. Integrations page has some differences in behavior. Testing this is as well is great 👍🏻

gtsiolis: This is the screenshot of the modal after clicking Activate Integration and being redirected back to the modal with the warning message at the top. See also https://github.com/gitpod-io/gitpod/pull/3995#issuecomment-822526861.

> From the logs, it looks like you went through the setup and created the gitlab.com integration. Then you went to the /integrations and tried to setup github.com.

Exactly!

> BTW, I always tested setup with both providers by deleting the account in between, to make sure it works both ways. Integrations page has some differences in behavior. 

💡

gtsiolis: **issue:** Also, regardless if the Git Integration failed I can see the Git Provider added in the Git Providers list and I'm also able to connect. Expected?

image

AlexTugarev: @gtsiolis, yes it is expected if you created the GH integration on the /integrations page ☝🏻 

From there, you can select to edit the GH integration and make it work eventually.

AlexTugarev: > This is the screenshot of the modal after clicking Activate Integration and being redirected back to the modal with the warning message at the top. See also #3995 (comment).

>> From the logs, it looks like you went through the setup and created the gitlab.com integration. Then you went to the /integrations and tried to setup github.com.

> Exactly!

Interesting, it seems to be an issue with the modal rendered on the /integrations page then! Thanks, will check.

AlexTugarev:
@gtsiolis, I figured why the unverified providers were visible on Login page, even for anonymous users. 

There is indeed a variance in gitpod-com, which handles it completely different. I though https://github.com/gitpod-io/gitpod/pull/3995/commits/fa72b2564575ca06953eff8aad382050b85b4d7a would just make sense to introduce in core as well, which changes the bits to hide not verified from the dashboard (as filtered API result)

In general, we might need to convert the compile time variance to deployment configuration, to support both cases.

.werft/build.ts
```diff
@@ -252,17 +252,17 @@ export async function deployToDev(deploymentConfig: DeploymentConfig, workspaceF
...
-        exec(`kubectl get secret preview-envs-authproviders --namespace=keys -o yaml \
-                | yq r - data.authProviders \
-                | base64 -d -w 0 \
-                > authProviders`, { slice: "authProviders" });
-        exec(`yq merge --inplace .werft/values.dev.yaml ./authProviders`, { slice: "authProviders" })
-        werft.done('authProviders');
-    } catch (err) {
-        werft.fail('authProviders', err);
-    }
+    // werft.log("authProviders", "copy authProviders")
```
AlexTugarev, line 255: TODO remove the test setup

gtsiolis:
> I figured why the unverified providers were visible on Login page, even for anonymous users.

Thanks for checking @AlexTugarev! Do you think it's worth opening a follow up issue for this? 🏓

.werft/build.ts
AlexTugarev: done

AlexTugarev:
> Do you think it's worth opening a follow up issue for this? 🏓

@gtsiolis, that's solved already.

geropl (approved):
:+1:

components/dashboard/src/provider-utils.tsx
```diff
@@ -75,12 +75,21 @@ async function openAuthorizeWindow(params: OpenAuthorizeWindowParams) {
...
         if (typeof event.data === "string" && event.data.startsWith("success")) {
             killAuthWindow();
-            onSuccess && onSuccess();
+            onSuccess && onSuccess(event.data);
         }
         if (typeof event.data === "string" && event.data.startsWith("error:")) {
-            const errorAsText = atob(event.data.substring("error:".length));
+            let error: string | { error: string, description?: string } = atob(event.data.substring("error:".length));
+            try {
+                const payload = JSON.parse(error);
```
geropl, line 83: `JSON.parse` on user-provided input is a minefield in general. We should be careful to not introduce XSS issues here..

geropl: We checked the usage and deemed it fine. The issues basically is if either:
  - you have an old browser with an polyfill `JSON.parse` -> `eval`
  - you set user-provided input to `href`/`src` attributes without validation.
 We do not do any of this here.

AlexTugarev: to summarize our discussion on that. 
`JSON.parse` isn't insecure per se, but we must pay attention to not just use parsed values as URLs for instance of `img.src`, etc. 

but even that is hard to reproduce and share, because the message event listener is only active when the auth window is open.

PR #4000, opened 2021-04-19
[registry-facade] Improve logging

corneliusludmann:
This PR improves registry facade logging.

### DIFF
--- a/components/registry-facade/pkg/registry/manifest.go
+++ b/components/registry-facade/pkg/registry/manifest.go
@@ -94,8 +94,11 @@ type manifestHandler struct {
 
 func (mh *manifestHandler) getManifest(w http.ResponseWriter, r *http.Request) {
 	span, ctx := opentracing.StartSpanFromContext(r.Context(), "getManifest")
+	logFields := log.OWI("", "", mh.Name)
+	logFields["tag"] = mh.Tag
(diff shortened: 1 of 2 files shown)


components/registry-facade/pkg/registry/manifest.go
```diff
@@ -94,8 +94,11 @@ type manifestHandler struct {
 
 func (mh *manifestHandler) getManifest(w http.ResponseWriter, r *http.Request) {
 	span, ctx := opentracing.StartSpanFromContext(r.Context(), "getManifest")
+	logFields := log.OWI("", "", mh.Name)
```
corneliusludmann, line 97: @csweichel Is `mh.Name` always the workspace intance ID? Or should we better log this as “name”?

In [lines 138](https://github.com/gitpod-io/gitpod/pull/4000/files#diff-5b66d9b7307235c24d727fa706460b5f365f2c330196905a74190e762f0ccaa5L138) and 143, it has been logged as “instanceId” before …

JanKoehnlein (approved):
LGTM

components/registry-facade/pkg/registry/manifest.go
csweichel: so far it's always the instance ID, yes :)

PR #4001, opened 2021-04-19
[registry-facade] Add buckets manifest_req_seconds

corneliusludmann:
Add additional buckets to the manifest_req_seconds metric.


ArthurSens:
Hi @corneliusludmann, I was wondering if we really need to increase those buckets today... 

Looking at those metrics in production for the last 7 days, the 99th percentile reached the highest bucket only once for a really short time. Our averages are constantly lower than those buckets, so I don't see a reason to increase it 🤔 

image

csweichel:
Considering it's only adding four more time series, and will tell us if we do hit exceptional circumstances when we reach the "10 seconds", I'd think this change is warranted.

csweichel (approved):
LGTM

PR #4018, opened 2021-04-20
[docker] Introduce socket activated docker-up

csweichel:
This PR provides `docker` every workspace without the need to call `docker-up`. Instead of simply running a Docker daemon in every workspace, we do something akin to [systemd's socket](http://0pointer.de/blog/projects/socket-activation.html) activation. When someone tries to use Docker (i.e. writes to `/var/run/docker.sock`), supervisor starts the Docker daemon and forwards the socket. The Docker daemon gets started in a supervisor terminal.

To make this work easier, and also enable supervisor's role within the workspace (reaper, terminator of processes during shutdown), supervisor now runs as UID 0. Terminals, the IDE and other child processes are still started as Gitpod user.

### How to test
1. `docker run --rm -it alpine:latest`
2. `/.supervisor/supervisor terminal list` should show the `docker-up` process


akosyakov (approved):
it looks good:
Screenshot 2021-04-21 at 13 22 40

Issue #3984, opened 2021-04-16
[dashboard] Update theme preference selector


gtsiolis:
### Problem to solve

The first part of the dark theme has been added in https://github.com/gitpod-io/gitpod/pull/3901. However, we are still missing a better styled theme selector.

| Theme Selector |
|-|
| <img width="394" alt="Screenshot 2021-04-16 at 10 51 06 PM" src="https://user-images.githubusercontent.com/120486/115077151-3ef81d00-9f06-11eb-8a90-db2d7a466a4f.png"> |

### Proposal

We could:
1. Update the theme selector style to follow the design language of the rest of the product and make the option more appealing to the user. This could use something like the card component used above in the editor preference.
2. Default to light theme to be less surprising to the user. Preference order could be _Light_, _Dark_, _System_.
3. Update the section design description to something more descriptive or maybe playful like _Early bird or night owl? Choose your side._
4. Consider introducing a secondary option to sync theme between dashboard and editor, see https://github.com/gitpod-io/gitpod/issues/3982#issuecomment-821373982.

### Designs

| Light | Dark |
|-|-|
| ![image](https://user-images.githubusercontent.com/120486/115416125-289cda80-a200-11eb-9094-aaa8f85d4d24.png) | ![image](https://user-images.githubusercontent.com/120486/115416136-2aff3480-a200-11eb-9bc0-a76add1ec183.png) |

See also [design specs](https://www.figma.com/file/8zu50RSZudHG3rtV9tzRJI/Theme-Selector-3984?node-id=0%3A1).

jankeromnes:
Looks awesome! ✨ Many thanks @gtsiolis 🙏

Additional thought: Once implemented, it could be nice to just copy/paste this Theme switcher into the "What's New" modal (thus allowing users to switch the Theme directly from What's New). Thoughts? 😇

gtsiolis:
Absolutely @jankeromnes! That was the plan, similarly to what @AlexTugarev did with the account select process in https://github.com/gitpod-io/gitpod/pull/3897. But haven't thought this through yet in terms of how this will scale, and more but it should be fine. 🤷

Issue #3982, opened 2021-04-16
[dashboard] Sync theme preference between dashboard and editor


gtsiolis:
### Problem to solve

We recently merged the first iteration of the dark theme for the dashboard in https://github.com/gitpod-io/gitpod/pull/3901. Users can now switch between a dark and light theme for both dashboard and editor.

Currently, the dashboard theme defaults to system default while the editor defaults to a light theme. We could try to sync the theme preference between dashboard and editor as long as the user has never changed the editor theme.

See [relevant discussion](https://gitpod.slack.com/archives/C01KGM9AW4W/p1618590899201400) (internal).

### Proposal

TBD (To be discussed)

svenefftinge:
We could also have a preference in the dashboard that says align with VS Code Theme. I.e. if the user switches to any dark theme in Code it will switch to dark theme in dashboard as well and switching to any light theme in code will switch to light theme in dashboard. 

I don't mean applying the color scheme to the dashboard (as we used to do) but just whether it is dark or light.

akosyakov:
Just fyi we cannot do much in VS Code, only decide which theme should be used by default if a user never touched settings.

jankeromnes:
I agree that the default theme of the dashboard and the default theme of VS Code should probably be the same. (And maybe we don't need to sync changes back-and-forth?)

> Just fyi we cannot do much in VS Code, only decide which theme should be used by default if a user never touched settings.

Ah, then we could also make VS Code use the "System" theme by default, right? (When user never touched the settings.)

You can get it like so:

```js
// Get the value once
const isDark = window.matchMedia("(prefers-color-scheme: dark)").matches;

// Listen to changes
window.matchMedia("(prefers-color-scheme: dark)").addEventListener('change', event => {
    const isDark = event.matches;
});
```

I'd vote for either:
1. "System" default in both dashboard and IDE
2. "Light" default in both dashboard and IDE

shaal:
++ for this feature!

I see it in the dashboard, but I don't think it works affects the theme for a new user that see VSCode interface for the first time.
(I changed my VSCode theme to dark theme, and it remembers that option for me across workspaces).

shaal:
I don't think this is completed yet. Perhaps add `meta: never-stale` ?

Issue #3956, opened 2021-04-15
[dashboard] Update What's New modal


gtsiolis:
### Problem

Back in https://github.com/gitpod-io/gitpod/issues/3297 and https://github.com/gitpod-io/gitpod/pull/3618, we introduced a What's New modal to inform users about new changes to the dashboard and offer the option to switch their editor to VS Code.

Some more context: 
1. The initial plan was to keep this modal for one release and remove it on the next one.
2. In this release we want to inform users about the new dashboard theme that is going to be available.
3. Users should be able to get back to this modal using a What's New button in the profile dropdown.
4. The updates that go in this feed can be quite useful for the Changelog feature we're adding for our website.

### Proposal

Let's:
1. Keep the What's New modal and figure out a scalable way to append new updates as well as hide past entries.
2. Use this modal to add an update about the dark theme availability.
3. Hide the past two entries (New Dashboard and Editor Switch) and attach the March 2021 milestone timestamp.
2. Introduce a What's New button on the profile dropdown so that users can trigger the What's New modal.
3. To avoid introducing too much noise, add a small dot indicator on the profile to indicate there's a new update.

### Designs 

TBD (To be discussed)

mikenikles:
😍 Happy to see this issue. The 4. context item above is near and dear to me. The more we align messaging between the changelog and the product updates, the better for our users.

Each changelog entry is currently a separate markdown file in https://github.com/gitpod-io/website/tree/main/src/contents/changelog. I'd like to find a way to align and reuse content used in the changelog and the What's New modal.

🤔 Food for thought: We use GitHub releases in `gitpod-io/gitpod` for all releases, not only self-hosted. The content of the new `CHANGELOG.md` file in `gitpod-io/gitpod` moves to the GitHub release page and we use the GitHub API to fetch release information to generate the What's New modal and the changelog on the website. A single source of truth (GitHub releases).

gtsiolis:
Cross-posting here a comment from a [relevant discussion](https://gitpod.slack.com/archives/C01KPEPNBD0/p1625442701085900) (internal) regarding why this modal is blocking workspace start.

> @gtsiolis: In future updates, we could skip blocking workspace startup and only show the _What’s New_ modal only when going through the dashboard.

gtsiolis:
[...]

PR #4021, opened 2021-04-20
[dashboard] Update theme preference selector

jankeromnes:
Fixes https://github.com/gitpod-io/gitpod/issues/3984

- [x] 1. Update the theme selector style
- [ ] 2. Default to light theme to be less surprising to the user (do we really want this?)
- [x] 3. Update the section design description to something more descriptive
- [ ] 4. Consider introducing a secondary option to sync theme between dashboard and editor (unsure about this)

### DIFF
--- a/components/dashboard/src/settings/Preferences.tsx
+++ b/components/dashboard/src/settings/Preferences.tsx
@@ -48,22 +48,34 @@ export default function Preferences() {
             <p className="text-base text-gray-500">Choose which IDE you want to use.</p>
             <div className="mt-4 space-x-4 flex">
                 <SelectableCard className="w-36 h-40" title="VS Code" selected={defaultIde === 'code'} onClick={() => actuallySetDefaultIde('code')}>
-                    <div className="flex-grow flex justify-center align-center">
+                    <div className="flex-grow flex justify-center items-center">
                         <img className="w-16 filter-grayscale" src={vscode}/>
                     </div>
                 </SelectableCard>
                 <SelectableCard className="w-36 h-40" title="Theia" selected={defaultIde === 'theia'} onClick={() => actuallySetDefaultIde('theia')}>
-                    <div className="flex-grow flex justify-center align-center">
+                    <div className="flex-grow flex justify-center items-center">
                         <img className="w-16 dark:filter-invert" src={theia}/>
                     </div>
                 </SelectableCard>
             </div>
             <h3 className="mt-12">Theme</h3>
-            <p className="text-base text-gray-500">Light or dark?</p>
+            <p className="text-base text-gray-500">Early bird or night owl? Choose your side.</p>
             <div className="mt-4 space-x-4 flex">
-                <label><input type="radio" name="theme" value="system" checked={theme === 'system'} onChange={() => actuallySetTheme('system')}></input> System</label>
-                <label><input type="radio" name="theme" value="dark" checked={theme === 'dark'} onChange={() => actuallySetTheme('dark')}></input> Dark</label>
-                <label><input type="radio" name="theme" value="light" checked={theme === 'light'} onChange={() => actuallySetTheme('light')}></input> Light</label>
+                <SelectableCard className="w-36 h-32" title="Light" selected={theme === 'light'} onClick={() => actuallySetTheme('light')}>
+                    <div className="flex-grow flex justify-center items-end">
+                        <svg className="h-16" xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 108 64"><rect width="68" height="40" x="40" fill="#C4C4C4" rx="8"/><rect width="32" height="16" fill="#C4C4C4" rx="8"/><rect width="32" height="16" y="24" fill="#C4C4C4" rx="8"/><rect width="32" height="16" y="48" fill="#C4C4C4" rx="8"/></svg>
+                    </div>
+                </SelectableCard>
+                <SelectableCard className="w-36 h-32" title="Dark" selected={theme === 'dark'} onClick={() => actuallySetTheme('dark')}>
+                    <div className="flex-grow flex justify-center items-end">
+                        <svg className="h-16" xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 108 64"><rect width="68" height="40" x="40" fill="#737373" rx="8"/><rect width="32" height="16" fill="#737373" rx="8"/><rect width="32" height="16" y="24" fill="#737373" rx="8"/><rect width="32" height="16" y="48" fill="#737373" rx="8"/></svg>
+                    </div>
+                </SelectableCard>
+                <SelectableCard className="w-36 h-32" title="System" selected={theme === 'system'} onClick={() => actuallySetTheme('system')}>
+                    <div className="flex-grow flex justify-center items-end">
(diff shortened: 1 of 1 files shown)


jankeromnes:
Great design! 💯 Currently looks like this:

Screenshot 2021-04-20 at 19 45 04

Screenshot 2021-04-20 at 19 45 15

jankeromnes:
Thought: Maybe the IDE cards and Theme cards should be squares?

components/dashboard/src/settings/Preferences.tsx
```diff
@@ -48,22 +48,34 @@ export default function Preferences() {
...
                 </SelectableCard>
             </div>
             <h3 className="mt-12">Theme</h3>
-            <p className="text-base text-gray-500">Light or dark?</p>
+            <p className="text-base text-gray-500">Early bird or night owl? Choose your side.</p>
             <div className="mt-4 space-x-4 flex">
-                <label><input type="radio" name="theme" value="system" checked={theme === 'system'} onChange={() => actuallySetTheme('system')}></input> System</label>
-                <label><input type="radio" name="theme" value="dark" checked={theme === 'dark'} onChange={() => actuallySetTheme('dark')}></input> Dark</label>
-                <label><input type="radio" name="theme" value="light" checked={theme === 'light'} onChange={() => actuallySetTheme('light')}></input> Light</label>
+                <SelectableCard className="w-36 h-32" title="Light" selected={theme === 'light'} onClick={() => actuallySetTheme('light')}>
```
gtsiolis, line 64: **praise**: Go re-usable components! 🎉

components/dashboard/src/settings/Preferences.tsx
```diff
@@ -48,22 +48,34 @@ export default function Preferences() {
...
                 <SelectableCard className="w-36 h-40" title="Theia" selected={defaultIde === 'theia'} onClick={() => actuallySetDefaultIde('theia')}>
-                    <div className="flex-grow flex justify-center align-center">
+                    <div className="flex-grow flex justify-center items-center">
                         <img className="w-16 dark:filter-invert" src={theia}/>
                     </div>
                 </SelectableCard>
             </div>
             <h3 className="mt-12">Theme</h3>
-            <p className="text-base text-gray-500">Light or dark?</p>
+            <p className="text-base text-gray-500">Early bird or night owl? Choose your side.</p>
```
gtsiolis, line 62: **thought**: Suggested text can be taken lightly here. If anyone has a better alternative, feel free to update! In any case, looks great for this iteration. 🦉

components/dashboard/src/settings/Preferences.tsx
```diff
@@ -48,22 +48,34 @@ export default function Preferences() {
...
-                <label><input type="radio" name="theme" value="dark" checked={theme === 'dark'} onChange={() => actuallySetTheme('dark')}></input> Dark</label>
-                <label><input type="radio" name="theme" value="light" checked={theme === 'light'} onChange={() => actuallySetTheme('light')}></input> Light</label>
+                <SelectableCard className="w-36 h-32" title="Light" selected={theme === 'light'} onClick={() => actuallySetTheme('light')}>
+                    <div className="flex-grow flex justify-center items-end">
+                        <svg className="h-16" xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 108 64"><rect width="68" height="40" x="40" fill="#C4C4C4" rx="8"/><rect width="32" height="16" fill="#C4C4C4" rx="8"/><rect width="32" height="16" y="24" fill="#C4C4C4" rx="8"/><rect width="32" height="16" y="48" fill="#C4C4C4" rx="8"/></svg>
+                    </div>
+                </SelectableCard>
+                <SelectableCard className="w-36 h-32" title="Dark" selected={theme === 'dark'} onClick={() => actuallySetTheme('dark')}>
```
gtsiolis, line 69: **question**: Curious why we named this `actuallySetTheme`? Isn't `setTheme` a sufficient alternative? Do we use `setTheme` elsewhere? 🤷

gtsiolis (approved):
This looks so good. Thanks for picking this up @jankeromnes! 🌟 🌟 🌟 

Regarding the pending tasks:

> Default to light theme to be less surprising to the user

No strong feelings for this one. We could keep defaulting to system theme, as the current setup. The two primary reasons for that suggestion are a) to avoid surprising the user, especially given that we can't sync back and forth with the editor (see https://github.com/gitpod-io/gitpod/issues/3982#issuecomment-822418098) and b) to make the dark and system setting opt-in in https://github.com/gitpod-io/gitpod/issues/3956 instead of a fallback.

> Consider introducing a secondary option to sync

We could keep this out of this iteration or wait to reach some consensus in https://github.com/gitpod-io/gitpod/issues/3982. I'd vote for getting this in as soon as possible.

> Maybe the IDE cards and Theme cards could be squares?

Also thought about this while designing the theme preference selector. Two reasons why I kept these as they are. First, a) to differentiate these options and make them distinct options from the editor preference[1] and second b) to evolve the card component by trying a new design[2].

[1] For example, we should separate tab index between editor and theme cards. Currently, you can cycle between every option when using the right or left arrows. You should able to jump between Editor and Theme options (cards) only using Tab.

[2] Our component system will naturally evolve and new designs will pop-up on the way as the system matures. Aligning the rest of the components by making them square here is also fine.

56ggq6

components/dashboard/src/settings/Preferences.tsx
jankeromnes: Haha 😂 nice catch! If you look above in the code, there is a `theme` React state (a `string`), with already a `setTheme(value)` function.

But if we just call `setTheme(value)` from the selector component, we'll just switch the React state / selected card, but not the actual web app theme itself.

That's where `actuallySetTheme(value)` comes in: it calls `setTheme(value)`, but then it also toggles the `dark` class on the body, and it also updates local storage so that the value is remembered across reloads.

If you see better names for these functions, please share. 😇

gtsiolis: Thanks for explaining @jankeromnes! Sounds good! Let's keep this as is. DOES. THE. JOB. 💯

components/dashboard/src/settings/Preferences.tsx
jankeromnes: Your text is perfect! 🌟🎯

components/dashboard/src/settings/Preferences.tsx
jankeromnes: More coming 🔜! 😇

jankeromnes:
Many thanks for the quick review! 🙏

> > Default to light theme to be less surprising to the user
> 
> No strong feelings for this one. We could keep defaulting to system theme, as the current setup. The two primary reasons for that suggestion are a) to avoid surprising the user, especially given that we can't sync back and forth with the editor (see https://github.com/gitpod-io/gitpod/issues/3982#issuecomment-822418098) and b) to make the dark and system setting opt-in in https://github.com/gitpod-io/gitpod/issues/3956 instead of a fallback.

That's a good point. In this light, I think the default theme should be the same for the dashboard and the IDE (i.e. "System" default in both, or "Light" default in both, but maybe we don't need to sync beyond the default value). Added my thoughts to the issue.

> > Consider introducing a secondary option to sync
> 
> We could keep this out of this iteration or wait to reach some consensus in https://github.com/gitpod-io/gitpod/issues/3982. I'd vote for getting this in as soon as possible.

Agreed, thanks! 🚀

PR #4024, opened 2021-04-20
[jaeger] fixed jaeger deployment in werft using jaeger-operator

wulfthimm:
This make jaeger run again in preview environments using the preinstalled `jaeger-operator`. It also uses a state file for Terraform to keep the state for redeployment and dns record removal.
You can test it by deploying a new branch on `gitpod-dev`and visit the URL https://<NAMESPACE>.jaeger.gitpod-dev.com or simply https://wth-add-jaeger-to-prev.jaeger.gitpod-dev.com/.

PR #4031, opened 2021-04-21
[jaeger] schedule jaeger to workload nodes

wulfthimm:
This is a follow up to #4024 making sure that jaeger pods are scheduled to run only on workload nodes.


JanKoehnlein (approved):
I am affine with that

PR #4033, opened 2021-04-21
[self-hosted] Set sensitive values via secrets - fixes #3094

### LINKED ISSUE #3094: [Helm chart] Configuring sensitive values via secrets
Hi,
Would it be possible to set the sensitive values in the Helm chart (like `components.server.sessionSecret` for ex., but also db password and oauth info) via either an `existingSecret` secret reference or setting an environment variable? 
Right now the `values.yaml` has sensitive information in it which is an issue for “gitops-style” management of Kubernetes.
Thanks!

cyrilcros:
Configure the chart via existingSecrets; fixes #3094 

- I fetch values via Helm `lookup` calls. This is Helm 3 only compatible.
- I do not alter the templates to use the existing secrets, I fetch from them and let gitpod create (possibly redundant) secrets. 
- In theory you can use secrets gitpod has previously generated as `existingSecretName`and be fine.
- serverProxyApiKey / db.password / components.server.sessionSecret can be set, added via secret with arbitrary key and are autogenerated otherwise
- rabbitmq secret can be set, assuming username and password are keys in the secret
- nothing is done for minio secrets because it is a subchart with a April 2021 deprecation warning
- I haven't touched the Gihub oAuth secret yet; because it is a list it would be messier?


cyrilcros:
Putting this in for consideration, I need to properly test all those configurations.

PR #4037, opened 2021-04-22
[build] Don't trigger full build on forks

csweichel:
This PR ensures we're not publishing to NPM or building the installer on forks.

/werft no-preview


akosyakov (approved):
code changes look good

PR #4042, opened 2021-04-22
[ws-manager] Metrics for control ports

ArthurSens:
When ready, this PR should introduce 2 new metrics for `ws-manager`:
* Current amount of opened ports (Gauge)
* Duration of the controlPort operation (Histogram)


ArthurSens:
Closing in favor of #4048

PR #4048, opened 2021-04-23
[ws-manager] Add metric counting total open ports

csweichel:
This PR adds a metric for counting the total number of exposed ports across all workspaces.


ArthurSens (approved):
Thanks a lot Chris! I always learn something when coding with you 🤗

PR #4065, opened 2021-04-26
[code] proxy fetch in web workers

### LINKED ISSUE #4063: [code] outputLinkComputer cannot load dependencies
### Bug description

Start a workspace with VS Code and open the output view. You should see them in devtools many errors like An unknown error occurred. Please consult the log for more details.: Error at https://salmon-hummingbird-ey9ae4ob.ws-eu02.gitpod-staging.com/out/vs/base/worker/workerMain.js:707:31

### Steps to reproduce

See above.

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

akosyakov:
#### What it does

- fix #4063: VS Code switched to use fetch instead of importScripts to load additional scripts becuase latter hangs on Safari sometimes, so we should proxy former as well

#### How to test

- Start a workspace with VS Code.
- Check that diff editors are working.
- Check that opening the output channel does not trigger load of errors, also link resolution should work.
- Test in Chrome, FF and Safari.

### DIFF
--- a/components/supervisor/frontend/public/worker-proxy.js
+++ b/components/supervisor/frontend/public/worker-proxy.js
@@ -15,8 +15,28 @@
     var originalImportScripts = self.importScripts;
     // hash contains the original worker URL to be used as a base URI to resolve script URLs
     var baseURI = decodeURI(location.hash.substr(1));
-    self.importScripts = function (scriptUrl) {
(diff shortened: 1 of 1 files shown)


svenefftinge (approved):

akosyakov:
VS Code does not open in Safari anymore.

PR #4073, opened 2021-04-27
Selective new-user blocking

csweichel:
In some Gitpod installations new users get blocked by default.  This PR adds the ability to do this new user blocking selectively based on the domain of their primary email. Supersedes #3899 

### How to test
1. Login in https://cw-newuser-passlist.staging.gitpod-dev.com/workspaces with a user with an @gitpod.io primary email - you should be able to start a workspace
2. Login in https://cw-newuser-passlist.staging.gitpod-dev.com/workspaces with a user with a different primary email - should be blocked

### ToDo before merging
- drop https://github.com/gitpod-io/gitpod/commit/7e990e0136366286033c20303d1a37f7171399fa

### DIFF
--- a/.werft/values.dev.yaml
+++ b/.werft/values.dev.yaml
@@ -35,6 +35,9 @@ components:
     makeNewUsersAdmin: true # for development
     theiaPluginsBucketName: gitpod-core-dev-plugins
     enableLocalApp: true
+    blockNewUsers: true
+    blockNewUsersPasslist:
+      - "gitpod.io"
 
   registryFacade:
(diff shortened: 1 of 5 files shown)


svenefftinge (approved):

svenefftinge:
Looks good, but the code is overwritten in -com so wouldn't be sufficient.
https://github.com/gitpod-com/gitpod/pull/5278/files#diff-da11a738582b42d187c59d607cd40d7b673311be40b6da6f161f9c65013cdd15L37

PR #4078, opened 2021-04-27
[dashboard] Standardize PendingChangesDropdown component in workspaces list (replaces Modal) and stopped page (unchanged)

jankeromnes:
Align with [spec](https://www.figma.com/file/XQOjjEHO0ALtxdIT9Blzh2/Dark-Theme-%233806) and side effect fix for dark theme:

image (2)

### DIFF
--- a/components/dashboard/src/workspaces/WorkspaceEntry.tsx
+++ b/components/dashboard/src/workspaces/WorkspaceEntry.tsx
@@ -9,8 +9,9 @@ import { GitpodHostUrl } from '@gitpod/gitpod-protocol/lib/util/gitpod-host-url'
 import ContextMenu, { ContextMenuEntry } from '../components/ContextMenu';
 import moment from 'moment';
 import Modal from '../components/Modal';
-import { MouseEvent, useState } from 'react';
+import { useState } from 'react';
 import { WorkspaceModel } from './workspace-model';
+import PendingChangesDropdown from '../components/PendingChangesDropdown';
 import Tooltip from '../components/Tooltip';
 
 function getLabel(state: WorkspaceInstancePhase) {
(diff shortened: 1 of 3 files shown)


jankeromnes:
Currently looks like this:

<table><tr><td>
Screenshot 2021-04-27 at 14 03 34
</td><td>
Screenshot 2021-04-27 at 14 03 50
</td></tr></table>

Problems:
- Vertical alignment is off
- Clicking on it doesn't work

jankeromnes:
Now looks like this:

<table><tr><td>
Screenshot 2021-04-27 at 14 25 01
</td><td>
Screenshot 2021-04-27 at 14 25 17
</td></tr><tr><td>
Screenshot 2021-04-27 at 14 25 45
</td><td>
Screenshot 2021-04-27 at 14 25 29
</td></tr></table>

jankeromnes:
@gtsiolis could you please take a look, and make a final decision? 🙏

- Either we go with this PR, replacing the Changes Modal with a dropdown

- Or we temporarily park this PR to keep the Changes Modal, and simply adjust the balloon colors for dark theme, and fix the Modal-induced scrollbar

gtsiolis:
Looking at this now! 👀

jankeromnes:
Probably not critical for the release tomorrow (since dark theme is soft-launch, and there will probably be other small things to fix here and there 😇).

So: no rush 🌴

gtsiolis (approved):
Thanks for aligning this @jankeromnes! UX looks good! 🌟 

Alright, I think this looks good to merge as is! We could probably add some hover and focus states on this new changes dropdown but this is certainly out of the scope of this PR.

components/dashboard/src/workspaces/WorkspaceEntry.tsx
```diff
@@ -155,36 +133,6 @@ export function WorkspaceEntry({ desc, model, isAdmin, stopWorkspace }: Props) {
     </div>;
 }
 
-export interface PendingChanges {
-    message: string, items: string[]
-}
-
-export function getPendingChanges(wsi?: WorkspaceInstance): PendingChanges[] {
```
gtsiolis, line 162: **praise**: Deleted code is the best code! 💯

gtsiolis:
> Either we go with this PR, replacing the Changes Modal with a dropdown
> ...

Sounds like a good plane for now! 🗺️ 

---

Posting this here for future reference:

Also, this seems like a good **two-way door decision** which allows us to come back and reverse this change in the future! 🔮

jankeromnes:
Awesome, many thanks for the review! Then let's go forward with unifying this, but I'm more than happy to re-introduce a Changes Modal later. 🚀

PR #4084, opened 2021-04-27
[dashboard] Change default theme from 'system' to 'light'

jankeromnes:
Reasons:
- Don't surprise users with a different default theme & no further explanation (we don't have a "What's New" modal for the April release)
- Align with the default theme of the IDE

In a future iteration, we can still switch both the dashboard and the IDE to "system" by default, and announce it with a "What's New" modal.

### DIFF
--- a/components/dashboard/src/App.tsx
+++ b/components/dashboard/src/App.tsx
@@ -58,7 +58,7 @@ function App() {
 
     useEffect(() => {
         const updateTheme = () => {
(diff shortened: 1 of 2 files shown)


gtsiolis:
Looking at this now! 👀

gtsiolis (approved):
Thanks for getting this in @jankeromnes! Works like a charm! 🔮 🌙 

This one also seems like a two-way door decision which we can revert later or change to `system` as the default option if needed.

PR #4091, opened 2021-04-28
[dashboard] fix images on Select Account modal

AlexTugarev:
This small change includes a fix for the avatar icon. 

Screen Shot 2021-04-28 at 15 03 25

### DIFF
--- a/components/dashboard/src/settings/SelectAccountModal.tsx
+++ b/components/dashboard/src/settings/SelectAccountModal.tsx
@@ -52,7 +52,7 @@ export function SelectAccountModal(props: SelectAccountPayload & {
         <div className="mt-10 mb-6 flex-grow flex flex-row justify-around align-center">
             <SelectableCard className="w-2/5 h-56" title="Current Account" selected={useAccount === 'current'} onClick={() => setUseAccount('current')}>
                 <div className="flex-grow flex flex-col justify-center align-center">
-                    <img className="m-auto rounded-full w-24 h-24 py-4" src={props.currentUser.avatarUrl} alt={props.currentUser.name}/>
+                    <img className="m-auto rounded-full w-24 h-24 my-4" src={props.currentUser.avatarUrl} alt={props.currentUser.name}/>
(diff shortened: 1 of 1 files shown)


components/dashboard/src/settings/SelectAccountModal.tsx
```diff
@@ -52,15 +52,15 @@ export function SelectAccountModal(props: SelectAccountPayload & {
         <div className="mt-10 mb-6 flex-grow flex flex-row justify-around align-center">
             <SelectableCard className="w-2/5 h-56" title="Current Account" selected={useAccount === 'current'} onClick={() => setUseAccount('current')}>
                 <div className="flex-grow flex flex-col justify-center align-center">
-                    <img className="m-auto rounded-full w-24 h-24 py-4" src={props.currentUser.avatarUrl} alt={props.currentUser.name}/>
+                    <img className="m-auto rounded-full w-24 h-24 my-4" src={props.currentUser.avatarUrl} alt={props.currentUser.name}/>
```
gtsiolis, line 55: **praise**: Nice catch! ⚾

gtsiolis (approved):
Thanks for catching this @AlexTugarev! Tried this and looks good! 🏀

components/dashboard/src/settings/SelectAccountModal.tsx
```diff
@@ -52,15 +52,15 @@ export function SelectAccountModal(props: SelectAccountPayload & {
...
                     <span className="m-auto text-gray-700 text-md font-semibold">{props.currentUser.authName}</span>
```
jankeromnes, line 56: Whoops, this makes me realize that Select Account probably looks bad in dark theme 😅 (good think we're only shadow-launching dark theme this time -- off by default)

gtsiolis: Thanks for noticing @jankeromnes! We're probably going to run into more small issues like this, which is fine! Opening follow-up issues for such cases should suffice! See also [relevant discussion](https://gitpod.slack.com/archives/C01KGM9BUNS/p1619517479170700) (internal). /cc @AlexTugarev

Issue #4092, opened 2021-04-28
GitHub App install confirmation flow is broken


jankeromnes:
### Bug description

When I install the [Gitpod Staging](https://github.com/apps/gitpod-staging) for one of my repositories, the installation confirmation flow is broken.

Still, the app successfully installs and then works, so it's just the confirmation that is broken.

### Steps to reproduce

(Note: Problems in **bold**)

- Start here: https://github.com/apps/gitpod-staging

<img width="600" alt="Screenshot 2021-04-28 at 15 34 15" src="https://user-images.githubusercontent.com/599268/116413250-bb103000-a837-11eb-80f8-c843d223f14f.png">

- Install it for your user, for some specific repository:

<img width="600" alt="Screenshot 2021-04-28 at 15 34 34" src="https://user-images.githubusercontent.com/599268/116413262-bd728a00-a837-11eb-9e31-0076a5590585.png">

- In the installation screen, click on "Install App":

<img width="600" alt="Screenshot 2021-04-28 at 15 34 49" src="https://user-images.githubusercontent.com/599268/116413266-be0b2080-a837-11eb-8ce1-ff62786eaf76.png">

- This opens a new tab, that says it should close itself, **but it doesn't**. Close it manually:

<img width="600" alt="Screenshot 2021-04-28 at 15 34 59" src="https://user-images.githubusercontent.com/599268/116413273-bea3b700-a837-11eb-91af-f00d20e7c71d.png">

- The confirmation screen says that **the installation has failed**. Oh no! Click "Try again":

<img width="600" alt="Screenshot 2021-04-28 at 15 35 10" src="https://user-images.githubusercontent.com/599268/116413274-bea3b700-a837-11eb-9137-ec40d2b68f1a.png">

- Clicking "Try Again" gives **a strange error message**:

<img width="600" alt="Screenshot 2021-04-28 at 15 35 18" src="https://user-images.githubusercontent.com/599268/116413276-bf3c4d80-a837-11eb-96c3-63a38ed56a76.png">

However, the app seems to work. Maybe the installation did actually work?

### Expected behavior

- The auto-closing pop-up should auto-close
- Gitpod shouldn't say "installation failed"
- "Try Again" shouldn't give a strange cryptic error

AlexTugarev:
> This opens a new tab, that says it should close itself, but it doesn't. Close it manually:

@jankeromnes did you pay attention to the JS console?

jankeromnes:
[...]

PR #4093, opened 2021-04-28
[github] fix register app

AlexTugarev:
fixes https://github.com/gitpod-io/gitpod/issues/4092

### DIFF
--- a/components/dashboard/src/prebuilds/InstallGitHubApp.tsx
+++ b/components/dashboard/src/prebuilds/InstallGitHubApp.tsx
@@ -16,13 +16,6 @@ async function registerApp(installationId: string, setModal: (modal: 'done' | st
     try {
         await getGitpodService().server.registerGithubApp(installationId);
 
-        const returnTo = encodeURIComponent(gitpodHostUrl.with({ pathname: 'complete-auth', search: 'message=success' }).toString());
-        const url = gitpodHostUrl.withApi({
-            pathname: '/authorize',
-            search: `returnTo=${returnTo}&host=github.com&scopes=repo`
-        }).toString();
-        window.open(url, "gitpod-login");
-
         const result = new Deferred<void>(1000 * 60 * 10 /* 10 min */);
 
         openAuthorizeWindow({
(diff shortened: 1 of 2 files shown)


components/dashboard/src/prebuilds/InstallGitHubApp.tsx
```diff
@@ -16,13 +16,6 @@ async function registerApp(installationId: string, setModal: (modal: 'done' | st
     try {
         await getGitpodService().server.registerGithubApp(installationId);
 
-        const returnTo = encodeURIComponent(gitpodHostUrl.with({ pathname: 'complete-auth', search: 'message=success' }).toString());
-        const url = gitpodHostUrl.withApi({
-            pathname: '/authorize',
-            search: `returnTo=${returnTo}&host=github.com&scopes=repo`
-        }).toString();
-        window.open(url, "gitpod-login");
```
AlexTugarev, line 24: this is left-over. the `window.open` call is centralized in `openAuthorizeWindow`.

components/dashboard/src/prebuilds/InstallGitHubApp.tsx
```diff
@@ -16,13 +16,6 @@ async function registerApp(installationId: string, setModal: (modal: 'done' | st
     try {
         await getGitpodService().server.registerGithubApp(installationId);
```
jankeromnes, line 17: Nit: If possible, every `window.open()` call should always happen in the exact same event loop cycle as the `click` event.

Otherwise the `window.open()` will most likely be blocked by the browser's pop-up blocker. This means, no `await`, no callbacks, etc.

jankeromnes (approved):
Thanks for the quick fix!

Code looks good to me, and I guess there is no convenient way to test it apart from gitpod-staging.com

components/dashboard/src/prebuilds/InstallGitHubApp.tsx
AlexTugarev: Can we change the order?

components/dashboard/src/provider-utils.tsx
```diff
@@ -56,12 +56,8 @@ async function openAuthorizeWindow(params: OpenAuthorizeWindowParams) {
             search: `returnTo=${encodeURIComponent(returnTo)}&host=${host}&override=true&scopes=${(scopes || []).join(',')}`
         }).toString();
 
-    const newWindow = window.open(url, "gitpod-auth-window");
-    if (!newWindow) {
```
AlexTugarev, line 60: reverting this to get to square one, where the untracked new window was opened just fine. 

let's think this through one more time, next time 🙏🏻

PR #4096, opened 2021-04-28
[dashboard/auth] don't override scopes when trying to acquire addional

AlexTugarev:
this PR fixes the authorization request for the registration of GitHub Apps.

### the problem

URL params of the authorization request were static `...&override=true&scopes=repo`, which works only for the Integration page, where we request the full set. 

### the solution

when requesting additional scopes, we should drop the `override=true` param.

### DIFF
--- a/components/dashboard/src/provider-utils.tsx
+++ b/components/dashboard/src/provider-utils.tsx
@@ -39,13 +39,15 @@ interface OpenAuthorizeWindowParams {
     login?: boolean;
     host: string;
     scopes?: string[];
+    overrideScopes?: boolean;
     onSuccess?: (payload?: string) => void;
     onError?: (error?: string) => void;
 }
 
 async function openAuthorizeWindow(params: OpenAuthorizeWindowParams) {
-    const { login, host, scopes, onSuccess, onError } = params;
+    const { login, host, scopes, overrideScopes, onSuccess, onError } = params;
     const returnTo = gitpodHostUrl.with({ pathname: 'complete-auth', search: 'message=success' }).toString();
+    const requestedScopes = scopes || [];
     const url = login
         ? gitpodHostUrl.withApi({
             pathname: '/login',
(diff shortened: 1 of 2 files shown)


components/dashboard/src/provider-utils.tsx
```diff
@@ -39,21 +39,24 @@ interface OpenAuthorizeWindowParams {
...
     onSuccess?: (payload?: string) => void;
     onError?: (error?: string) => void;
 }
 
 async function openAuthorizeWindow(params: OpenAuthorizeWindowParams) {
-    const { login, host, scopes, onSuccess, onError } = params;
+    const { login, host, scopes, additionalScopes, onSuccess, onError } = params;
     const returnTo = gitpodHostUrl.with({ pathname: 'complete-auth', search: 'message=success' }).toString();
+    const requestedScopes = scopes || additionalScopes || [];
+    const overrideScopes = !!scopes;
```
jankeromnes, line 51: This API seems a bit weird: If you provide both `additionalScopes` and `scopes`, it will pick `additionalScopes` but override?

Why introduce a different `additionalScopes` list in the first place? Maybe it's simpler to just make `overrideScopes` configurable?

AlexTugarev: Agree.

components/dashboard/src/settings/Integrations.tsx
```diff
@@ -158,6 +158,7 @@ function GitProviders() {
             await openAuthorizeWindow({
                 host,
                 scopes,
+                overrideScopes: true,
```
AlexTugarev, line 161: I think this is the single place where we actually want to override the scopes.
In all other cases we want to elevate scopes anyways.

jankeromnes (commented):
Unfortunately, the default value is not what every call-site currently expects.

Maybe also update this call-site?

https://github.com/gitpod-io/gitpod/blob/a356f9fe8659e1328b46e4ab520cd337e7071c5d/components/dashboard/src/start/CreateWorkspace.tsx#L71-L76

It also seems to me that we're mingling two different functions here, `openLoginWindow` and `openAuthorizeWindow`, but fixing that could be a follow-up.

jankeromnes (approved):
Code change looks good to me. Thanks!

AlexTugarev:
> Maybe also update this call-site?

Especially for the CreateWorkspace.tsx I'd argue, that it shouldn't be overridden at all. In that case, if it handles the NOT_AUTHENTICATED error, the provided guessed scopes should be appended.

Issue #4083, opened 2021-04-27
Improve Grant Access rendering on Start Workspace


jankeromnes:
### Bug description

When authorizing GitLab, I used to be redirected automatically.

Now, I just see confusing error messages, and often the proposed action doesn't work.

### Flow 1: Works ✅ 

- Delete GitpodStaging entry from https://gitlab.com/profile/applications if present.
- open a new private browser
- open https://gitpod-staging.com/#https://gitlab.com/AlexTugarev/gp-test
- ensure the onboarding flow works nicely.

---

### Flow 2: Doesn't work well ❌

- Delete GitpodStaging entry from https://gitlab.com/profile/applications if present.
- open a new private browser
- login to https://gitpod-staging.com/login using GitHub Auth
- open https://gitpod-staging.com/#https://gitlab.com/AlexTugarev/gp-test
- ensure the onboarding flow works nicely.

It shows me this error message, which doesn't make sense, because the repository is not private (it's [public](https://gitlab.com/AlexTugarev/gp-test)). However, clicking the button makes it work:

<img width="1440" alt="Screenshot 2021-04-27 at 16 12 15" src="https://user-images.githubusercontent.com/599268/116257244-2b557d80-a774-11eb-9769-33e5d0dfc479.png">

---

### Flow 3: Doesn't work at all ❌ 

- Delete GitpodStaging entry from https://gitlab.com/profile/applications if present.
- Disconnect from GitLab in https://gitpod-staging.com/integrations
- open a new private browser
- login to https://gitpod-staging.com/login using GitHub Auth
- open https://gitpod-staging.com/#https://gitlab.com/AlexTugarev/gp-test
- ensure the onboarding flow works nicely.

It shows me this error screen. Clicking on Authorize opens a GitLab tab, where I can authorize, but then it takes me back to the very same error page (clicking again multiple times always returns to the same error / never opens a workspace):

<img width="1440" alt="Screenshot 2021-04-27 at 16 13 52" src="https://user-images.githubusercontent.com/599268/116257971-cf3f2900-a774-11eb-9368-9b2f2e459aa7.png">

### Expected behavior

All 3 flows should redirect me + do the right thing + open a workspace successfully.

AlexTugarev:
@jankeromnes, thanks for reporting. 

Indeed `Flow 3` is broken. There seems to be no retry in that path, which lets me wonder why the path is different to `Flow 2`.

OTOH, `Flow 2` worked for me. Just double checked. 

[...]

PR #4105, opened 2021-04-29
[dashboard] fix create workspace retry

AlexTugarev:
when authorization is required, i.e. no provider identity is found on server, we request to authorize. once this is done, we should retry creating workspace.

the location based approach to reload seems to be broken.

partial fix for https://github.com/gitpod-io/gitpod/issues/4083

### DIFF
--- a/components/dashboard/src/start/CreateWorkspace.tsx
+++ b/components/dashboard/src/start/CreateWorkspace.tsx
@@ -72,7 +72,7 @@ export default class CreateWorkspace extends React.Component<CreateWorkspaceProp
             host,
             scopes,
             onSuccess: () => {
-              window.location.href = window.location.toString();
+              window.location.reload();
             },
             onError: (error) => {
                 if (typeof error === "string") {


jankeromnes (approved):
Indeed, I believe that setting an identical `window.location.href` will not navigate.

`window.location.reload()` should do the trick, and I've just confirmed that this PR fixes flow 3 from https://github.com/gitpod-io/gitpod/issues/4083

Still, I think the UX should be improved. Previously, these pages would just navigate to the authorization URL automatically, making the errors invisible & doing the right thing immediately. Now we show a somewhat confusing error, where the user doesn't know exactly what's happening and have to randomly click on the button to go forward.

Cc @gtsiolis about improving the UX for both screenshots in https://github.com/gitpod-io/gitpod/issues/4083

AlexTugarev:
> Now we show a somewhat confusing error

@jankeromnes, that's the point. Why do we show the technical details of those error modes, given that the action is always the same: `Authorize to continue`

> Previously, these pages would just navigate to the authorization URL automatically

In a discussion with @gtsiolis, we found the auto redirect problematic because:
* given that you follow a `gitpod.io#REPO` link from another (maybe external) page, this is quite surprising and unintuitive, that user's browser is redirected to a 3rd party site. 
* the actual task, starting a workspace in this case, is out of context, once something going wrong during the OAuth2 flow. there is no chance to easily retry.

Issue #4101, opened 2021-04-28
Support custom email addresses


philschatz:
Hello!

Github allows users to [Block command line pushes that expose my email](https://github.com/settings/emails) but it seems gitpod does not notice that setting and instead uses my primary GitHub email when generating the workspace `~/.gitconfig` file.

![image](https://user-images.githubusercontent.com/253202/116468321-543e4700-a836-11eb-9842-e259821d3e73.png)

It seems like there are at least 2 different ways to solve this niche problem:

1. Change something like [manager/create.go](https://github.com/gitpod-io/gitpod/blob/main/components/ws-manager/pkg/manager/create.go#L531) to allow `GIT_COMMITTER_EMAIL` if the user has set it in their "Variables" list
2. Change something like [server/.../github-auth-provider.ts](https://github.com/gitpod-io/gitpod/blob/main/components/server/src/github/github-auth-provider.ts#L65) to store the email address that GitHub expects (e.g. `253202+philschatz@users.noreply.github.com`) if no public emails are present.

Would either of these approaches (or some other ones) work? If so, it's probably just a couple of lines to fix but I'm happy to try making the changes!

AlexTugarev:
Hey @philschatz,
This is a duplicate of https://github.com/gitpod-io/gitpod/issues/387.

The comprehensive approach would be to add an optional `commitEmail` next to the `primaryEmail`, so that it would be filled in for both GitHub and GitLab users.

philschatz:
@AlexTugarev , Thanks for that feedback. I created a Pull Request (#4115) which I believe uses the comprehensive approach you mentioned.

corneliusludmann:
Fixed with #4115

Issue #3186, opened 2021-02-14
Use GitHub email alias instead of private email


k3nsei:
It would be nice to have github email alias instead of private email adress when creating pod from GitHub.
With new option to hide private emails and blocking commits which contains them it annoying to constantli changing git config.

For example mine is `190422+k3nsei@users.noreply.github.com`

jankeromnes:
Hi @k3nsei!

Please try adding these two environment variables in https://gitpod.io/settings/ : 

- Name: `GIT_AUTHOR_EMAIL`, Value: `190422+k3nsei@users.noreply.github.com`, Org/Repo: `*/*`
- Name: `GIT_COMMITTER_EMAIL`, Value: `190422+k3nsei@users.noreply.github.com`, Org/Repo: `*/*`

And then stop/restart any existing Gitpod workspaces, or just create new Gitpod workspaces.

I think that should do it.

jankeromnes:
Also, closing as a duplicate of https://github.com/gitpod-io/gitpod/issues/387. Thanks again!

k3nsei:
@jankeromnes Thanks :)

Issue #2007, opened 2020-10-16
[GitLab] Use "Commit Email" for commits


svenefftinge:
GitLab has labels for emails and we should the "commit email" instead of the primary one for commits.

PR #4115, opened 2021-04-29
[server] store separate email used for commits (GitHub and GitLab)

### LINKED ISSUE #387: Support "Keep my email address private"
Currently, `git config user.email` is forced be the one primary in GitHub.
I want to use `ID+username@users.noreply.github.com` by default when `Keep my email address private` is enabled ( https://help.github.com/en/articles/about-commit-email-addresses ).
Otherwise, users have to run `git config --global user.email "ID+username@users.noreply.github.com"` everytime starting new projects.

philschatz:
GitLab and GitHub both have a way to specify an email address for commits that is different than the primary email used for a user account.

GitLab contains a field for the [commit email](https://docs.gitlab.com/ee/api/users.html#list-current-user-for-normal-users) and while GitHub does not provide a similar field in their API, their Desktop client [constructs this email address](https://github.com/desktop/desktop/blob/7a7388edd069cc8e97ec6d3cdbe4d2747af5fbc4/app/src/lib/email.ts)

:crossed_fingers: :sweat_smile: that this works.

Fixes #387, #2007, #3186, #4101

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## July 2021
 
 - Fix `image.context` in `.gitpod.yml` ([#4715](https://github.com/gitpod-io/gitpod/pull/4715))
+- Support custom commit email address for GitHub and GitLab (keep email private). Thanks to [@philschatz](https://github.com/philschatz) for the contribution! ([#4115](https://github.com/gitpod-io/gitpod/pull/4115))
 
 ## June 2021
 
--- a/components/gitpod-db/src/typeorm/migration/1624290876856-IdentityTokensToEmails.ts
+++ b/components/gitpod-db/src/typeorm/migration/1624290876856-IdentityTokensToEmails.ts
@@ -1,0 +1,18 @@
+/**
+ * Copyright (c) 2021 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class IdentityTokensToEmails1624290876856 implements MigrationInterface {
+
+    public async up(queryRunner: QueryRunner): Promise<any> {
+        await queryRunner.renameColumn("d_b_identity", "tokens", "additionalEmails");
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<any> {
+        await queryRunner.renameColumn("d_b_identity", "additionalEmails", "tokens");
+    }
+}
--- a/components/gitpod-protocol/src/protocol.ts
+++ b/components/gitpod-protocol/src/protocol.ts
@@ -60,8 +60,6 @@ export namespace User {
         const res = { ...user };
         delete (res.additionalData);
         res.identities = res.identities.map(i => {
-            delete (i.tokens);
-
             // The user field is not in the Identity shape, but actually exists on DBIdentity.
             // Trying to push this object out via JSON RPC will fail because of the cyclic nature
             // of this field.
@@ -303,8 +301,7 @@ export interface Identity {
     authId: string;
     authName: string;
     primaryEmail?: string;
-    /** @deprecated */
-    tokens?: Token[];
+    additionalEmails?: Email[];
     /** This is a flag that triggers the HARD DELETION of this entity */
     deleted?: boolean;
     // readonly identities cannot be modified by the user
@@ -322,6 +319,26 @@ export namespace Identity {
     export function equals(id1: IdentityLookup, id2: IdentityLookup) {
         return id1.authProviderId === id2.authProviderId
             && id1.authId === id2.authId
+    }
+}
+
+export enum EmailType {
+    COMMIT = 'commit',
+    OTHER = 'other'
+}
+
+export interface Email {
+    address: string;
+    type: EmailType;
+}
+
+export namespace Email {
+    export function is(data: any): data is Email {
+        return data.hasOwnProperty('address')
+            && data.hasOwnProperty('type');
+    }
+    export function isEmailArray(data: any): data is Email[] {
+        return Array.isArray(data) && data.every(x => Email.is(x));
     }
 }
 
--- a/components/server/src/auth/auth-provider.ts
+++ b/components/server/src/auth/auth-provider.ts
@@ -6,7 +6,7 @@
 
 
 import * as express from 'express';
-import { AuthProviderInfo, User, OAuth2Config, AuthProviderEntry } from "@gitpod/gitpod-protocol";
+import { AuthProviderInfo, User, OAuth2Config, AuthProviderEntry, Email } from "@gitpod/gitpod-protocol";
 import { saveSession } from '../express-util';
 
 import { UserEnvVarValue } from "@gitpod/gitpod-protocol";
@@ -65,6 +65,7 @@ export interface AuthUser {
     readonly authId: string;
     readonly authName: string;
     readonly primaryEmail: string;
+    readonly additionalEmails?: Email[];
     readonly name?: string;
     readonly avatarUrl?: string;
 }
--- a/components/server/src/github/github-auth-provider.ts
+++ b/components/server/src/github/github-auth-provider.ts
@@ -6,7 +6,7 @@
 
 import { injectable } from 'inversify';
 import * as express from "express"
-import { AuthProviderInfo } from '@gitpod/gitpod-protocol';
+import { AuthProviderInfo, EmailType } from '@gitpod/gitpod-protocol';
 import { log } from '@gitpod/gitpod-protocol/lib/util/logging';
 import { GitHubScope } from "./scopes";
 import { AuthUserSetup } from "../auth/auth-provider";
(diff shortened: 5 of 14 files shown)


components/server/src/github/github-auth-provider.ts
```diff
@@ -104,7 +104,8 @@ export class GitHubAuthProvider extends GenericAuthProvider {
                     authName: login,
                     avatarUrl: avatar_url,
                     name,
-                    primaryEmail: userEmails.filter(e => e.primary)[0].email
```
AlexTugarev, line 107: @philschatz, I think we need to explicitly check for the primary email address to be private.

I just played with the `Keep my email addresses private` setting and the `https://api.github.com/user/emails` API, and found that if active the proxy address is included in the list. This dynamic behavior should allow to reflect the setting on each login into Gitpod. 

Could you please adopt the strategy to do this?

```
const primary = userEmails.filter(e => e.primary)[0]!;
const proxy = userEmails.find(e => e.email.endsWith(`@users.noreply.${this.config.host}`));
...
  primaryEmail: primary.email,
  commitEmail: primary.visibility === "private" ? proxy?.email : primary.email,
...
```

components/server/src/auth/auth-provider.ts
```diff
@@ -65,6 +65,7 @@ export interface AuthUser {
     readonly authId: string;
     readonly authName: string;
     readonly primaryEmail: string;
+    readonly commitEmail: string;
```
AlexTugarev, line 68: `commitEmail` might be optional here as well.

philschatz:
Thank you for the suggestions! I updated the PR with them but had trouble connecting to https://ps-commit-email-fork.staging.gitpod-dev.com/workspaces/ to see if it works. Is there something I can set up?

AlexTugarev:
@philschatz, thanks for the update. I've triggered a a new build on werft. Our public [werft](https://werft.gitpod-dev.com/) is only listening to GItpod devs. I keep an eye open on the PR to trigger it, but if you won't see a new update timely, feel free to ping me 🙏🏻

AlexTugarev:
@philschatz, we forgot to add a db migration. 

I'm happy to add this to this PR. Just, internally we've a discussion on how to add columns to db tables like `d_b_identity`. 'Keep you posted.

philschatz:
Hello @AlexTugarev :wave: . Is there any decision on how to do the db migration?

AlexTugarev:
Hi @philschatz! Sorry for the delay. 
A nice option would be to introduce `Identity.emails: { address:string, type: string }`. As we need to be careful to not introduce expensive blocking DB migration on the production environment, it would be great to rename `d_b_identity.tokens` in a migration and reuse it, because this isn't in use anymore. 
I'll try to make a PR against yours.

corneliusludmann:
Hi @philschatz!

Sorry, that it took so long to get feedback. We need to get better at this.

I would like to get your great contribution into the June release which will be deployed by the end of this month. I rebased your contribution and added the proposed changes from Alex in [this branch](https://github.com/gitpod-io/gitpod/tree/clu/additional-emails) (→ [diff](https://github.com/gitpod-io/gitpod/compare/clu/additional-emails)).

We have now 2 options:
1. We could use my branch with your commits and I'll create a PR on behalf of both of us. In that case, no action is required on your end.
2. You could update your PR with the changes in my PR.

Your call!

After that, I'll take care that we get a review soon.

What do you think?

corneliusludmann:
There is one more thing, @philschatz: For legal reasons, we need a signed contributor license agreement (CLA) from each contributor. @meysholdt will get back to you on this.

philschatz:
Hello @corneliusludmann ! I somehow missed your comment on options but did sign the CLA and rebased this Pull Request with your commits.

Please let me know if there is anything else you need to get this merged (yay!)

corneliusludmann:
Awesome! Thanks a lot!

@AlexTugarev Do you have time to review this PR? Since I've contributed as well I need a second pair of eyes.

components/server/src/auth/auth-provider.ts
```diff
@@ -65,6 +65,7 @@ export interface AuthUser {
     readonly authId: string;
     readonly authName: string;
     readonly primaryEmail: string;
+    readonly additionalEmails?: { address: string, type: string }[];
```
AlexTugarev, line 68: nit: could be typed as `Email` as well, right?

components/server/src/github/github-auth-provider.ts
```diff
@@ -98,26 +98,35 @@ export class GitHubAuthProvider extends GenericAuthProvider {
                 .map((s: string) => s.trim())
             );
 
-            const filterPrimaryEmail = (emails: typeof userEmails) => {
+            const blockPassListEmail = ((emails: typeof userEmails) => {
```
AlexTugarev, line 101: I think this should be moved to `UserService.handleNewUser`, cf. https://github.com/gitpod-io/gitpod/blob/main/components/server/src/user/user-service.ts#L140
This way, we do not need to repeat filtering.

components/server/src/github/github-auth-provider.ts
```diff
@@ -98,26 +98,35 @@ export class GitHubAuthProvider extends GenericAuthProvider {
...
 
             return <AuthUserSetup>{
                 authUser: {
                     authId: String(id),
                     authName: login,
                     avatarUrl: avatar_url,
                     name,
-                    primaryEmail: filterPrimaryEmail(userEmails)
+                    primaryEmail: primary.email,
+                    additionalEmails,
```
AlexTugarev, line 129: I think we should add **all** _verified_ email addresses to `additionalEmails` here. This could also be of use to improve the signup process.

The `proxy` is marked as `type=commit`, which is also available in the GitLab case. 👍🏻 

For the git config we would use `identity.additionalEmails.find(isCommitEmail) || identity.primaryEmail` then.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 ## June 2021
 
+- Support custom commit email address for GitHub and GitLab (keep emai private). Thanks to [@philschatz](https://github.com/philschatz) for the contribution! ([#4115](https://github.com/gitpod-io/gitpod/pull/4115))
```
AlexTugarev, line 7: nit: typo

components/gitpod-db/src/typeorm/migration/1624290876856-IdentityTokensToEmails.ts
```diff
@@ -0,0 +1,18 @@
...
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class IdentityTokensToEmails1624290876856 implements MigrationInterface {
+
+    public async up(queryRunner: QueryRunner): Promise<any> {
+        await queryRunner.renameColumn("d_b_identity", "tokens", "additionalEmails");
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<any> {
+        await queryRunner.renameColumn("d_b_identity", "additionalEmails", "tokens");
```
AlexTugarev, line 16: WDYT of a one-way migration? 
in `up` we can check if `tokens` exists and only executes if necessary.
in `down` we don't do anything about it, to not add additional costs to a potential rollback, which in this case would not be interesting at all, as `tokens` dried out already.

geropl: I'm, too, in favor of not too much mix the clean-up with the new stuff. Concretely:
 Up: create new column, drop old
 Down: drop new.

components/gitpod-protocol/src/protocol.ts
```diff
@@ -323,6 +320,21 @@ export namespace Identity {
     }
 }
 
+export interface Email {
+    address: string;
+    type: string;
```
geropl, line 325: We definitely should make `type` typesafe by having a `export type EmailType = "commit" | "block_pass"` (or similar).

components/gitpod-db/src/typeorm/migration/1624290876856-IdentityTokensToEmails.ts
AlexTugarev: OTOH, I think the point of rollback is actually bogus. So, NVM 
If we rollback and do not rename back, it might break typeorm.

corneliusludmann: > Up: create new column, drop old
> Down: drop new.

@geropl As far as I understood it, we want to re-use the old column for performance reasons. Creating a new column in production would result in long database migration, renaming not. Right?

CHANGELOG.md
corneliusludmann: Also: Since we don't get this into June release, we should move this to July.

components/gitpod-db/src/typeorm/migration/1624290876856-IdentityTokensToEmails.ts
AlexTugarev: That is correct. The current DB version on PROD would do a full table rebuild when adding columns, but renaming should be fine.

components/server/src/github/github-auth-provider.ts
corneliusludmann: When we store all verified emails (as suggested below) we can drop this here because it is already in `UserService.handleNewUser`. See:

https://github.com/philschatz/gitpod/blob/88af0c51bf7c29b55e510c24d7c825aaa28b3254/components/server/src/user/user-service.ts#L143-L144

Otherwise, this filter would be necessary to store just the needed email addresses.

components/server/src/github/github-auth-provider.ts
corneliusludmann: OK, then I would suggest adding all _verified_ and _non-private_ (public) email addresses.

The commit email gets the type `commit` (as it is now) and all others get the type `other`.

corneliusludmann:
Thanks, Alex and Gero for the feedback. 

@philschatz Please let me know if I should prepare a commit with the suggested changes (I would be happy to do so) or if you would like to do this. Your call! :slightly_smiling_face:

components/server/src/auth/auth-provider.ts
philschatz: 👍  Yep! I believe this is now resolved.

components/server/src/github/github-auth-provider.ts
philschatz: 👍  I believe this is now resolved.

components/server/src/github/github-auth-provider.ts
philschatz: 👍  All verified, non-private, and non-primary emails should now be in the `additionalEmails` array.

components/gitpod-protocol/src/protocol.ts
philschatz: 👍  `type` is now `COMMIT | OTHER` since the block_pass logic is handled in `user-service.ts`.

corneliusludmann:
Hi @philschatz! Thanks for updating this PR. We are almost there! :rocket: 

I reviewed and tested your changes: It looks great! Unfortunately, this branch does not build because we need to rebase it on top of `main` and it has also merge conflicts. The good news is: I already did it in [this branch](https://github.com/gitpod-io/gitpod/tree/clu/commit-email) and it works pretty well. I also took the liberty to squash your commits into one to have a cleaner git history.

Feel free to take the branch [`clu/commit-email`](https://github.com/gitpod-io/gitpod/tree/clu/commit-email) for this PR and we are just a click away from merging this into `main` and getting this in production soon.

Thanks again for taking the initiative to contribute to this change and thanks for your patience. I really looking forward to having this great feature in production!

corneliusludmann (approved):
Thanks to you, @philschatz, we finally have this frequently requested feature in Gitpod! :partying_face:

AlexTugarev:
🎉 Let's verify on staging that the migration and transformation works 👍🏻 

Thanks @corneliusludmann  and @philschatz!

philschatz:
Thanks everyone for the help and feedback!

Also, I'm excited to try deploying 0.9.0 on my little k3s at home and being able to edit & test more contributions!

PR #4118, opened 2021-04-30
Improve hostname validation

AlexTugarev:


components/server/src/bitbucket/bitbucket-context-parser.spec.ts
```diff
@@ -18,7 +18,7 @@ import { BitbucketTokenHelper } from "./bitbucket-token-handler";
 const expect = chai.expect;
 import { skipIfEnvVarNotSet } from "@gitpod/gitpod-protocol/lib/util/skip-if";
 
-@suite.only(timeout(10000), skipIfEnvVarNotSet("GITPOD_TEST_TOKEN_BITBUCKET"))
```
geropl, line 21: :100:

components/server/src/express-util.ts
```diff
@@ -91,7 +91,14 @@ const looksLikeWorkspaceHostname = (originHostname: URL, gitpodHostName: string)
         return false;
     }
     const url = new GitpodHostUrl(originHostname);
-    return workspaceIDRegex.test(url.workspaceId || '')
+    const workspaceId = url.workspaceId;
+    if (workspaceId) {
+        const hostname = url.url.hostname as string;
+        if (hostname.startsWith(workspaceId)) {
```
geropl, line 97: Would be nice if we had one place for the workspace URL parsing. But might be out of scope for this PR, as it was here before, too.

geropl (approved):
Tested, works fine. Also, re-enables the test that failed because of this! :heart:

components/server/src/express-util.ts
AlexTugarev: `GitpodHostUrl` does the URL parsing. Here we explicitly check, that the workspace-like hostname is not prefixed.
Ah, perhaps you mean to extract the computation of the prefix and move it to `GitpodHostUrl`?  🤷🏻‍♂️

Issue #3992, opened 2021-04-19
Admin->Workspace Search should be able to search based on workspace names/IDs


meysholdt:
Today we can search based on context URL.

However, in support requests and log messages we often don't get that information. 

It would be awesome if the search could find the workspaces based on basically any unique identifier:
* workspaces name (e.g "pink-teddybear-1245") aka workspace-ID
* instance-UUID

mads-hartmann:
I'll be looking into this as an onboarding ticket over the next couple of days :)

mads-hartmann:
When this is done we should also update the NodeFilesystemAlmostOutOfSpace runbook

JanKoehnlein:
Is that ready for a PR?

mads-hartmann:
@JanKoehnlein I have an open PR here https://github.com/gitpod-io/gitpod/pull/4111 but it's missing a review ☺️

PR #4120, opened 2021-04-30
Add tests for findAllWorkspaceAndInstances

mads-hartmann:
While working on https://github.com/gitpod-io/gitpod/issues/3992 I was trying to add a few DB tests. 

Before making changes to the code I wanted to create a basic test for the existing functionality, but unfortunately the test failed with a rather obscure error message

```
Cannot read property 'databaseName' of undefined
```

Turns out that TypeORM column names are case sensitive when used for ordering. In this case we used contextUrl and it should have been contextURL.

While column names in mysql are case insensitive I also fixed one other occurrence of contextUrl even though it is not necessary; but using the column names makes it easier to find when searching for references etc.

I also added a launch configuration for executing the db tests with the debugger enabled so you can step through them.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -19,6 +19,21 @@
             "request": "attach",
             "port": 32991,
             "mode": "remote"
+        },
+        {
+            // This will run the db-test yarn target in the gitpod-db component.
+            // This allows you to set breakpoints in your tests and step through
+            // them with the VSCode debugger.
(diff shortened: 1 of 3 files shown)


geropl (approved):
Nice find, thx for digging in and fixing it.

In general we could think about using `Object.keys` to be sure we use the exact same field name. But that's Out of scope for this PR I think.

PR #4141, opened 2021-05-04
Add telemetry to help debug workspace start failures

mads-hartmann:
As part for https://github.com/gitpod-io/observability/issues/90 I'm writing a runbook for an SLO-based alert for workspace start failures.

I believe it would be useful to be able to get a log line whenever we increment `gitpod_ws_manager_workspace_stops_total{reason="failed"}` with more context that might help shed some light on why workspaces are failing to start.


mads-hartmann:
We decided to approach this differently so I'm closing this.

PR #4147, opened 2021-05-05
Add Segment support

csweichel:
We've predominantly relied on the database to understand how Gitpod is used. This approach has hit it's limits, hence we're looking to introduce something like segment to help with this matter.

### DIFF
--- a/components/gitpod-protocol/src/util/analytics.ts
+++ b/components/gitpod-protocol/src/util/analytics.ts
@@ -1,0 +1,84 @@
+/**
+ * Copyright (c) 2021 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import Analytics = require("analytics-node");
+import { log } from './logging';
+
+export const IAnalyticsWriter = Symbol("IAnalyticsWriter");
+
+type Identity =
+    | { userId: string | number }
+    | { userId?: string | number; anonymousId: string | number };
+
+interface Message {
+    messageId?: string;
+}
+
+export type IdentifyMessage = Message & Identity & {
+    traits?: any;
+    timestamp?: Date;
+    context?: any;
+};
+
+export type TrackMessage = Message & Identity & {
+    event: string;
+    properties?: any;
+    timestamp?: Date;
+    context?: any;
+};
+
+export function newAnalyticsWriterFromEnv(): IAnalyticsWriter {
+    switch (process.env.ANALYTICS_WRITER) {
+        case "segment":
+            return new SegmentAnalyticsWriter(process.env.SEGMENT_WRITE_KEY || "");
+        case "log":
+            return new LogAnalyticsWriter();
+        default:
+            return new NoAnalyticsWriter();
+    }
+}
+
+export interface IAnalyticsWriter {
+
+    identify(msg: IdentifyMessage): void;
+
+    track(msg: TrackMessage): void;
+
+}
+
+class SegmentAnalyticsWriter implements IAnalyticsWriter {
+    
+    protected readonly analytics: Analytics;
+
+    constructor(writeKey: string) {
+        this.analytics = new Analytics(writeKey);
+    }
+
+    identify(msg: IdentifyMessage) {
+        this.analytics.identify(msg);
+    }
+
+    track(msg: TrackMessage) {
+        this.analytics.track(msg);
+    }
+
+}
+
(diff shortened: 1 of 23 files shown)


components/workspacekit/cmd/rings.go
```diff
@@ -569,7 +569,7 @@ var ring2Cmd = &cobra.Command{
 			return
 		}
 
-		err = unix.Exec(ring2Opts.SupervisorPath, []string{"supervisor", "run", "--inns"}, os.Environ())
+		err = unix.Exec(ring2Opts.SupervisorPath, []string{"supervisor", "run"}, os.Environ())
```
AlexTugarev, line 572: what did `inns` mean? related to the namespace clean up ☝🏻 ?

components/supervisor/pkg/supervisor/supervisor.go
```diff
@@ -238,10 +235,6 @@ func Run(options ...RunOption) {
 	ideWG.Wait()
 	terminateChildProcesses()
 
-	if !opts.InNamespace {
```
AlexTugarev, line 241: clean up?

components/supervisor/cmd/run.go
```diff
@@ -11,30 +11,16 @@ import (
...
-
-		supervisor.Run(opts...)
+		supervisor.Run()
 	},
 }
 
 func init() {
 	rootCmd.AddCommand(runCmd)
-
-	runCmd.Flags().BoolVar(&runCmdOpts.InNamespace, "inns", false, "disables teardown canary.")
```
AlexTugarev, line 39: I see

components/server/src/workspace/workspace-starter.ts
```diff
@@ -171,6 +173,17 @@ export class WorkspaceStarter {
             const resp = (await manager.startWorkspace({ span }, startRequest)).toObject();
             span.log({ "resp": resp });
 
+            this.analytics.track({ 
+                userId: user.id, 
+                event: "workspace-started",
```
AlexTugarev, line 178: a definition of constants for event names would make sense, or not? at least to see the names being aligned.

components/server/src/user/user-controller.ts
```diff
@@ -509,6 +511,7 @@ export class UserController {
 
         await this.userService.updateUserEnvVarsOnLogin(user, envVars);
         await this.userService.acceptCurrentTerms(user);
+        this.analytics.identify({ anonymousId: req.sessionID || "unknown", userId: user.id, });
```
AlexTugarev, line 514: it would make sense to move it to `LoginCompletionHandler.complete`, which makes me think of the `request.login` call and if there is a chance to read the session id afterwards. 

L514 might already have a non-null req.sessionID, but this might change with `request.login`

components/gitpod-protocol/src/util/analytics.ts
```diff
@@ -0,0 +1,84 @@
...
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import Analytics = require("analytics-node");
+import { log } from './logging';
+
+export const IAnalyticsWriter = Symbol("IAnalyticsWriter");
+
+type Identity =
+    | { userId: string | number }
```
AlexTugarev, line 13: why number?
why not `{ user?: { id: string }}` and handle the plumbing internally, so we save the `userId: user?.id || ""`

components/gitpod-protocol/src/util/analytics.ts
```diff
@@ -0,0 +1,84 @@
...
+    traits?: any;
+    timestamp?: Date;
+    context?: any;
+};
+
+export type TrackMessage = Message & Identity & {
+    event: string;
+    properties?: any;
+    timestamp?: Date;
+    context?: any;
```
AlexTugarev, line 30: what is the context for?

csweichel: see https://segment.com/docs/connections/spec/common/ (search for context)

components/gitpod-protocol/src/util/analytics.ts
csweichel: to make this shape match segment

components/server/src/workspace/workspace-starter.ts
csweichel: dunno. I could also imagine those names being generated in some cases.

AlexTugarev (approved):
LGTM!
Just keep in mind to move the `.identify` call into the completion handler, see https://github.com/gitpod-io/gitpod/pull/4147#discussion_r626443950

Issue #4106, opened 2021-04-29
ws-manager has slow but steady memory leak


meysholdt:
![image](https://user-images.githubusercontent.com/239422/116526273-b734ea00-a8d9-11eb-8cc9-25b0f4cd1b68.png)

Todo: Wait a few days until more memory has leaked and do `pprof` to find out more details about the leak.

meysholdt:
log entry of [kubelet evicting ws-manager](https://console.cloud.google.com/logs/query;cursorTimestamp=2021-04-29T08:20:01Z;query=%22ws-manager%22%0A%22kubelet%22%0Aseverity%20%3E%20INFO%0Atimestamp%3D%222021-04-29T08:20:01Z%22%0AinsertId%3D%22g3rqntg6cg7dj8%22?project=gitpod-191109)

JanKoehnlein:
Do we have a strategy how to proceed on this issue?

aledbf:
@JanKoehnlein test https://github.com/gitpod-io/gitpod/issues/4106 in the next deployment

aledbf:
Closing. The last deployment with #4384 fixed the issue.

![Screenshot from 2021-06-06 14-22-38](https://user-images.githubusercontent.com/161571/120935890-e3126f00-c6d2-11eb-8e11-c7a47b205369.png)

JanKoehnlein:
Thanks @aledbf !

PR #4149, opened 2021-05-05
Update jaeger and opentracing dependency

aledbf:
Update jaeger client library from [v2.16.0](https://github.com/jaegertracing/jaeger-client-go/releases/tag/v2.16.0) (released in March 2019) to latest version.


aledbf:
The update goal is to remove a memory leak reported in https://github.com/gitpod-io/gitpod/issues/4106

csweichel (approved):

PR #4150, opened 2021-05-05
Update golangci-lint

aledbf:


csweichel (approved):
LGTM

components/docker-up/docker-up/main.go
```diff
@@ -343,7 +344,7 @@ func installDocker() error {
 
 			file.Close()
 		}
-
+		//nolint:errcheck
```
csweichel, line 347: is `errcheck` part of another linter - asking because it's disabled in the workspace config

aledbf: latest golangci-lint version is more strict. That's why I had to add [more exceptions ](https://github.com/gitpod-io/gitpod/pull/4150/files#diff-44586d99da6cd7df4d771c897f236269d1bca6f4c7d455c06698f8daacf0e40aR11)
The goal is to enable this again, but we need to fix all the issues. In some cases, we need to exclude the linter or refactor, for instance, all the defer chat close something.

csweichel (approved):
LGTM (please squash before merging)

PR #4157, opened 2021-05-06
Use kube-rbac-proxy to expose prometheus metric endpoint

aledbf:


ArthurSens:
We use everything that is provided from kube-prometheus. We can see that we already have a cluster role with the right permissions 👍 

https://github.com/prometheus-operator/kube-prometheus/blob/main/manifests/prometheus-clusterRole.yaml

ArthurSens (approved):

PR #4167, opened 2021-05-07
Implement incremental prebuilds

jankeromnes:
### Design doc

https://www.notion.so/gitpod/Incremental-Prebuilds-49c0a840e54348acba05677bdc450841

### How to test

- ~~Use an `io-dev` preview (e.g. https://jx-incremental-prebuilds.staging.gitpod-io-dev.com/workspaces/)~~

- Use the `#prebuild/<repo>` URL prefix to trigger a full prebuild for some commit on some repository

- Use `#incremental-prebuild/<repo>` to trigger an incremental prebuild for some (different) commit (this will automatically select a suitable parent prebuild, or fall back to a full prebuild)

- Verify that both full and incremental prebuilds work as expected (e.g. the project is prebuilt & starts as usual)

- Testing with https://github.com/jankeromnes/gitpod-staging-prebuilds may be helpful (the `.gitpod.yml` logs every task run, along with timestamp and workspace ID, and prints a summary -- e.g. 2 different workspace IDs mean "full prebuild", 3 IDs mean "incremental prebuild")

- Check incremental prebuilds in the DB (e.g. parents, commits, durations, states) like so:

```
mysql> select p.cloneUrl, substr(p.id,1,8) id, substr(w.basedOnPrebuildId,1,8) parentId, p.creationTime, substr(p.commit,1,8) commit, p.state, time_to_sec(timediff(cast(i.stoppedTime as datetime), cast(i.creationTime as datetime))) seconds
  from d_b_prebuilt_workspace p
  left join d_b_workspace w on w.id = p.buildWorkspaceId
  left join d_b_workspace_instance i on i.workspaceId = w.id
  where w.basedOnPrebuildId is not null or p.id in (select basedOnPrebuildId from d_b_workspace where id in (select buildWorkspaceId from d_b_prebuilt_workspace))
  order by p.cloneUrl, p.creationTime desc;
+-------------------------------------------------------------+----------+----------+----------------------------+----------+-----------+---------+
| cloneUrl                                                    | id       | parentId | creationTime               | commit   | state     | seconds |
+-------------------------------------------------------------+----------+----------+----------------------------+----------+-----------+---------+
| https://github.com/gitpod-io/gitpod.git                     | 676df827 | c9b671bc | 2021-05-11 14:48:54.034660 | 066200cd | building  |    NULL |
| https://github.com/gitpod-io/gitpod.git                     | c9b671bc | NULL     | 2021-05-11 14:43:29.295607 | a0a5017b | available |     450 |
| https://github.com/gitpod-io/gitpod.git                     | 0d45a1a1 | cb7566b8 | 2021-05-11 14:40:39.234595 | 79db9d8b | available |     213 |
| https://github.com/gitpod-io/gitpod.git                     | b4a03621 | cb7566b8 | 2021-05-11 14:28:54.040767 | 3e3a6cb8 | available |     262 |
| https://github.com/gitpod-io/gitpod.git                     | 142ddbf2 | cb7566b8 | 2021-05-11 14:28:16.082448 | a0a5017b | available |     234 |
| https://github.com/gitpod-io/gitpod.git                     | 62a62628 | cb7566b8 | 2021-05-11 14:20:16.920090 | 70097b6d | available |     248 |
| https://github.com/gitpod-io/gitpod.git                     | cb7566b8 | NULL     | 2021-05-11 13:46:32.152943 | 2b2702f3 | available |     612 |
| https://github.com/jankeromnes/gitpod-staging-prebuilds.git | 92979128 | 088a5082 | 2021-05-11 13:59:07.449600 | 243426ad | available |      24 |
| https://github.com/jankeromnes/gitpod-staging-prebuilds.git | 964730fb | 088a5082 | 2021-05-11 13:56:44.305709 | 6f2aff3e | available |      32 |
| https://github.com/jankeromnes/gitpod-staging-prebuilds.git | 088a5082 | NULL     | 2021-05-11 13:49:50.546575 | 98b4bfb6 | available |      46 |
| https://gitlab.com/gitlab-org/gitlab.git                    | 2e015621 | 8eef45ed | 2021-05-11 14:25:43.513453 | af8249ab | available |     429 |
| https://gitlab.com/gitlab-org/gitlab.git                    | 76e903e8 | 8eef45ed | 2021-05-11 14:15:08.733923 | c7e05e5a | available |     393 |
| https://gitlab.com/gitlab-org/gitlab.git                    | 8eef45ed | NULL     | 2021-05-11 13:51:22.791788 | 7972bd89 | available |     545 |
+-------------------------------------------------------------+----------+----------+----------------------------+----------+-----------+---------+
13 rows in set, 25 warnings (0.01 sec)
```

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -280,6 +280,9 @@ components:
     wsman: []
     defaultBaseImageRegistryWhitelist: []
     defaultFeatureFlags: []
+    incrementalPrebuilds:
+      repositoryPasslist: []
+      commitHistory: 100
     ports:
       http:
         expose: true
--- a/components/content-service/pkg/initializer/git.go
+++ b/components/content-service/pkg/initializer/git.go
@@ -93,12 +93,13 @@ func (ws *GitInitializer) realizeCloneTarget(ctx context.Context) (err error) {
 
 	// checkout branch
 	if ws.TargetMode == RemoteBranch {
-		// create local branch based on remote
+		// create local branch based on specific remote branch
 		if err := ws.Git(ctx, "checkout", "-B", ws.CloneTarget, "origin/"+ws.CloneTarget); err != nil {
 			return err
 		}
 	} else if ws.TargetMode == LocalBranch {
-		if err := ws.Git(ctx, "checkout", "-b", ws.CloneTarget); err != nil {
+		// checkout local branch based on remote HEAD
+		if err := ws.Git(ctx, "checkout", "-B", ws.CloneTarget, "origin/HEAD"); err != nil {
 			return err
 		}
 	} else if ws.TargetMode == RemoteCommit {
@@ -106,8 +107,11 @@ func (ws *GitInitializer) realizeCloneTarget(ctx context.Context) (err error) {
 		if err := ws.Git(ctx, "checkout", ws.CloneTarget); err != nil {
 			return err
 		}
-	} else { //nolint:staticcheck
-		// nothing to do - we're already on the remote branch
+	} else {
+		// update to remote HEAD
+		if err := ws.Git(ctx, "reset", "--hard", "origin/HEAD"); err != nil {
+			return err
+		}
 	}
 	return nil
 }
--- a/components/server/ee/src/workspace/gitpod-server-impl.ts
+++ b/components/server/ee/src/workspace/gitpod-server-impl.ts
@@ -549,7 +549,7 @@ export class GitpodServerEEImpl<C extends GitpodClient, S extends GitpodServer> 
 
             const logCtx: LogContext = { userId: user.id };
             const cloneUrl = context.repository.cloneUrl;
-            const prebuiltWorkspace = await this.workspaceDb.trace({ span }).findPrebuiltWorkspaceByCommit(context.repository.cloneUrl, context.revision);
+            const prebuiltWorkspace = await this.workspaceDb.trace({ span }).findPrebuiltWorkspaceByCommit(cloneUrl, context.revision);
             const logPayload = { mode, cloneUrl, commit: context.revision, prebuiltWorkspace };
             log.debug(logCtx, "Looking for prebuilt workspace: ", logPayload);
             if (!prebuiltWorkspace) {
@@ -568,6 +568,10 @@ export class GitpodServerEEImpl<C extends GitpodClient, S extends GitpodServer> 
                 if (mode === CreateWorkspaceMode.ForceNew) {
                     // in force mode we ignore running prebuilds as we want to start a workspace as quickly as we can.
                     return;
+                    // TODO(janx): Fall back to parent prebuild instead, if it's available:
+                    //   const buildWorkspace = await this.workspaceDb.trace({span}).findById(prebuiltWorkspace.buildWorkspaceId);
+                    //   const parentPrebuild = await this.workspaceDb.trace({span}).findPrebuildByID(buildWorkspace.basedOnPrebuildId);
+                    // Also, make sure to initialize it by both printing the parent prebuild logs AND re-runnnig the before/init/prebuild tasks.
                 }
 
                 let result: WorkspaceCreationResult = {
--- a/components/server/ee/src/workspace/workspace-factory.ts
+++ b/components/server/ee/src/workspace/workspace-factory.ts
@@ -8,7 +8,7 @@ import * as uuidv4 from 'uuid/v4';
 import { WorkspaceFactory } from "../../../src/workspace/workspace-factory";
 import { injectable, inject } from "inversify";
 import { TraceContext } from "@gitpod/gitpod-protocol/lib/util/tracing";
-import { User, StartPrebuildContext, Workspace, CommitContext, PrebuiltWorkspaceContext, WorkspaceContext, WithSnapshot, WithPrebuild } from "@gitpod/gitpod-protocol";
+import { User, StartPrebuildContext, Workspace, CommitContext, PrebuiltWorkspaceContext, WorkspaceContext, WithSnapshot, WithPrebuild, TaskConfig } from "@gitpod/gitpod-protocol";
 import { log } from '@gitpod/gitpod-protocol/lib/util/logging';
 import { LicenseEvaluator } from '@gitpod/licensor/lib';
 import { Feature } from '@gitpod/licensor/lib/api';
@@ -56,7 +56,62 @@ export class WorkspaceFactoryEE extends WorkspaceFactory {
                 }
             }
 
-            let ws = await this.createForCommit({span}, user, commitContext, normalizedContextURL);
+            const config = await this.configProvider.fetchConfig({span}, user, context.actual);
+            const imageSource = await this.imageSourceProvider.getImageSource(ctx, user, context.actual, config);
+
+            // Walk back the commit history to find suitable parent prebuild to start an incremental prebuild on.
+            let ws;
+            for (const parent of (context.commitHistory || [])) {
+                const parentPrebuild = await this.db.trace({span}).findPrebuiltWorkspaceByCommit(commitContext.repository.cloneUrl, parent);
+                if (!parentPrebuild) {
+                    continue;
+                }
+                if (parentPrebuild.state !== 'available') {
+                    continue;
+                }
+                log.debug(`Considering parent prebuild for ${commitContext.revision}`, parentPrebuild);
+                const buildWorkspace = await this.db.trace({span}).findById(parentPrebuild.buildWorkspaceId);
+                if (!buildWorkspace) {
+                    continue;
+                }
+                if (!!buildWorkspace.basedOnPrebuildId) {
+                    continue;
+                }
+                if (JSON.stringify(imageSource) !== JSON.stringify(buildWorkspace.imageSource)) {
+                    log.debug(`Skipping parent prebuild: Outdated image`, {
+                        imageSource,
+                        parentImageSource: buildWorkspace.imageSource,
+                    });
+                    continue;
+                }
+                const filterPrebuildTasks = (tasks: TaskConfig[] = []) => (tasks
+                    .map(task => Object.keys(task)
+                        .filter(key => ['before', 'init', 'prebuild'].includes(key))
+                        // @ts-ignore
+                        .reduce((obj, key) => ({ ...obj, [key]: task[key] }), {}))
+                    .filter(task => Object.keys(task).length > 0));
+                const prebuildTasks = filterPrebuildTasks(config.tasks);
+                const parentPrebuildTasks = filterPrebuildTasks(buildWorkspace.config.tasks);
+                if (JSON.stringify(prebuildTasks) !== JSON.stringify(parentPrebuildTasks)) {
+                    log.debug(`Skipping parent prebuild: Outdated prebuild tasks`, {
+                        prebuildTasks,
+                        parentPrebuildTasks,
+                    });
+                    continue;
+                }
+                const incrementalPrebuildContext: PrebuiltWorkspaceContext = {
+                    title: `Incremental prebuild of "${commitContext.title}"`,
+                    originalContext: commitContext,
+                    prebuiltWorkspace: parentPrebuild,
+                }
+                ws = await this.createForPrebuiltWorkspace({span}, user, incrementalPrebuildContext, normalizedContextURL);
+                break;
+            }
+
+            if (!ws) {
+                // No suitable parent prebuild was found -- create a (fresh) full prebuild.
+                ws = await this.createForCommit({span}, user, commitContext, normalizedContextURL);
+            }
             ws.type = "prebuild";
             ws = await this.db.trace({span}).store(ws);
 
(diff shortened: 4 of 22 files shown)


components/server/ee/src/workspace/workspace-factory.ts
```diff
@@ -66,7 +88,8 @@ export class WorkspaceFactoryEE extends WorkspaceFactory {
                 cloneURL: commitContext.repository.cloneUrl,
                 commit: commitContext.revision,
                 state: "queued",
-                creationTime: new Date().toISOString()
+                creationTime: new Date().toISOString(),
+                ... (!!parentPrebuild ? { parentPrebuildId: parentPrebuild.id } : {})
```
jankeromnes, line 92: Might be slightly redundant:

```
parentPrebuildId === getWorkspaceById(buildWorkspaceId).basedOnPrebuildId
```

components/server/ee/src/workspace/workspace-factory.ts
```diff
@@ -56,7 +56,29 @@ export class WorkspaceFactoryEE extends WorkspaceFactory {
                 }
             }
 
-            let ws = await this.createForCommit({span}, user, commitContext, normalizedContextURL);
+            let ws;
+            let parentPrebuild;
+
+            // Find an older full prebuild to start an incremental prebuild
+            for (const parent of (commitContext.history || [])) {
```
jankeromnes, line 63: Thought: May want to ask {GitHub,GitLab,Bitbucket} API for commit history here, instead of in the context parser.

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -567,6 +568,7 @@ export class GitpodServerEEImpl<C extends GitpodClient, S extends GitpodServer>
             } else if (prebuiltWorkspace.state === 'queued' || prebuiltWorkspace.state === 'building') {
                 if (mode === CreateWorkspaceMode.ForceNew) {
                     // in force mode we ignore running prebuilds as we want to start a workspace as quickly as we can.
+                    // TODO: Fall back to `prebuiltWorkspace.parentPrebuildId` if it's available (i.e. finished successfully)
```
jankeromnes, line 571: Thought: Maybe don't fall back here, but earlier in the call tree (e.g. for when there is no running prebuild yet -- there we can't check for `prebuiltWorkspace.parentPrebuildId`)

jankeromnes:
Note: The `#prebuild/<repo>` manual prefix always triggers a non-incremental (i.e. full) prebuild.

So, in order to facilitate testing, I'm temporarily introducing a `#incremental-prebuild/<repo>` manual prefix.

components/server/ee/src/prebuilds/start-prebuild-context-parser.ts
```diff
@@ -23,6 +23,7 @@ export class StartPrebuildContextParser implements IPrefixContextParser {
             throw new Error("cannot start prebuilds on an issue context")
         }
 
+        // FIXME: The manual prebuild prefix forces non-incremental prebuilds. How to trigger incremental prebuilds?
```
jankeromnes, line 26: Note: Solved for now by introducing a new `#incremental-prebuild/` URL prefix.

svenefftinge:
> Make content initializer resilient to merge conflicts from older prebuilds (see #4167 (comment))

Oh, that reveals a more interesting problem with our init approach. I think we should reset --hard to the context's branch. Also looking at the code for local branch context (i.e. issue context) we don't pass the corresponding remote branch, which is needed now.

jankeromnes:
> > Make content initializer resilient to merge conflicts from older prebuilds (see #4167 (comment))
> 
> Oh, that reveals a more interesting problem with our init approach. I think we should reset --hard to the context's branch.

Indeed. I believe what happened was:
- the first (full) prebuild `c9b671bc` checked out commit `a0a5017b` and ran a full build, producing some untracked `go.sum` changes
- the second (incremental) prebuild loaded the first prebuild, then [stashed](https://github.com/gitpod-io/gitpod/blob/360759883778f3414e5c54fdfc8cae16068a19c9/components/content-service/pkg/initializer/prebuild.go#L88-L98) the `go.sum` changes, [checked out](https://github.com/gitpod-io/gitpod/blob/360759883778f3414e5c54fdfc8cae16068a19c9/components/content-service/pkg/initializer/prebuild.go#L105-L108) commit `066200cd`, then attempted to [pop the stash](https://github.com/gitpod-io/gitpod/blob/360759883778f3414e5c54fdfc8cae16068a19c9/components/content-service/pkg/initializer/prebuild.go#L110-L114), but this caused merge conflicts because the `go.sum` files had been updated between commits `a0a5017b` and `066200cd`
- then, whenever I tried to use the incremental prebuild in an interactive workspace, it would try to [stash](https://github.com/gitpod-io/gitpod/blob/360759883778f3414e5c54fdfc8cae16068a19c9/components/content-service/pkg/initializer/prebuild.go#L88-L98) changes again, but this now fails due to the unresolved conflicts (i.e. the incremental prebuild "succeeded" but can never be loaded / used by any workspaces)

The problem is that we [intentionally](https://github.com/gitpod-io/gitpod/blob/360759883778f3414e5c54fdfc8cae16068a19c9/components/content-service/pkg/initializer/prebuild.go#L110-L114) leave "stash pop" conflicts in place when loading a prebuild, delegating the resolution to the user. This no longer makes sense when we load an older prebuild to produce a newer prebuild.

I agree we might want to `git reset --hard` to the context's Git ref, at least for incremental prebuilds. Or maybe we undo a stash pop that leads to conflicts.

jankeromnes:
> Also looking [at the code](https://github.com/gitpod-io/gitpod/blob/main/components/content-service/pkg/initializer/git.go#L101) for local branch context (i.e. issue context) we don't pass the corresponding remote branch, which is needed now.

Indeed, great catch! Since `origin` is [fetched](https://github.com/gitpod-io/gitpod/blob/360759883778f3414e5c54fdfc8cae16068a19c9/components/content-service/pkg/initializer/prebuild.go#L101-L104) prior to the [checkout -b](https://github.com/gitpod-io/gitpod/blob/360759883778f3414e5c54fdfc8cae16068a19c9/components/content-service/pkg/initializer/prebuild.go#L105-L108), I believe we can achieve correctness by simply basing the new local branch on `origin/HEAD`, e.g. like so:

```
ws.Git(ctx, "checkout", "-B", ws.CloneTarget, "origin/HEAD")
```

jankeromnes:
🎰

/werft run

components/content-service/pkg/initializer/git.go
```diff
@@ -108,6 +108,7 @@ func (ws *GitInitializer) realizeCloneTarget(ctx context.Context) (err error) {
 		}
 	} else { //nolint:staticcheck
 		// nothing to do - we're already on the remote branch
+		// RemoteHead: TODO rebase/reset-hard to origin/HEAD?
```
svenefftinge, line 111: +1

svenefftinge:
Looking at the code it seems like there is no `git fetch` happening before the `realizeCloneTarget`.

jankeromnes:
> Looking at the code it seems like there is no `git fetch` happening before the `realizeCloneTarget`.

@svenefftinge Which code do you mean? In https://github.com/gitpod-io/gitpod/pull/4167#issuecomment-841343905 I pointed out:

https://github.com/gitpod-io/gitpod/blob/360759883778f3414e5c54fdfc8cae16068a19c9/components/content-service/pkg/initializer/prebuild.go#L101-L108

components/content-service/pkg/initializer/git.go
jankeromnes: Fixed

jankeromnes:
Logging a small issue I noticed in testing: Occasionally, the `#incremental-prebuild/` prefix fails and tells me "[Cannot (re-)start irregular workspace](https://github.com/gitpod-io/gitpod/blob/dd0826cc0d3432a6b601c86760387ce0afc01038/components/server/src/workspace/gitpod-server-impl.ts#L444)". Refreshing once resolves it. Maybe there is a rare secondary code path that doesn't like incremental prebuilds? (E.g. when create workspace doesn't start the instance fast enough, and we call start again explicitly?)

jankeromnes:
Alright, this is now ready for review! 🚀 @svenefftinge @geropl @csweichel please take a look when convenient. 🙏

Notes:
- A few small items remain to be fixed (in this PR, or as follow-up PRs): https://github.com/gitpod-io/gitpod/pull/4167#issuecomment-838821967
- This PR is based on branch `production`, not `main`, in order to make staging hotfix deployments easier (will rebase to `main` later)
- The `core-dev` preview env in this PR won't work (prebuilds are an Enterprise feature) -- please use https://jx-incremental-prebuilds.staging.gitpod-io-dev.com/workspaces/ for testing
- Until someone merges to `gitpod-com/master`, you can also test this PR in https://gitpod-staging.com (but check if https://gitpod-staging.com/api/version still says "jx-incremental-prebuilds")

Please ping me if you have any questions!

svenefftinge:
Works very well using the prebuild and incremental-prebuild context URLs. 
It doesn't use a previous prebuild when starting a workspace on a non-prebuild commit as described here
https://www.notion.so/gitpod/Incremental-Prebuilds-49c0a840e54348acba05677bdc450841#de09ae678ff84779940e3237c0483c04

But I think it is fine to do this later when we'Äve gained more experience with incremental prebuilds.

jankeromnes:
Rebased, and squashed as hard as I could! 😁

jankeromnes:
Fails to create new workspaces since the rebase:

Screenshot 2021-05-21 at 11 01 38

jankeromnes:
🎰

/werft run

jankeromnes:
Update: Simple re-deploy resolved the awkward content init / ws-daemon / registry-facade errors from https://github.com/gitpod-io/gitpod/pull/4167#issuecomment-845802197 👍 everything works again.

chart/values.yaml
```diff
@@ -280,6 +280,8 @@ components:
     wsman: []
     defaultBaseImageRegistryWhitelist: []
     defaultFeatureFlags: []
+    incrementalPrebuildsRepoPasslist: []
+    incrementalPrebuildsCommitHistory: 100
```
csweichel, line 284:
```suggestion
    incrementalPrebuilds:
        repoPasslist: []
        commitHistory: 100
```

components/server/src/bitbucket/bitbucket-context-parser.ts
```diff
@@ -259,4 +259,27 @@ export class BitbucketContextParser extends AbstractContextParser implements ICo
 
         return result;
     }
+
+    public async fetchCommitHistory(ctx: TraceContext, user: User, contextUrl: string, sha: string, maxDepth: number): Promise<string[]> {
+        const span = TraceContext.startSpan("BitbucketContextParser.fetchCommitHistory", ctx);
+        try {
+            // FIXME: To get more results than Bitbucket API's max pagelen (seems to be 100), pagination should be handled.
```
csweichel, line 266:
```suggestion
            // TODO(janx): To get more results than Bitbucket API's max pagelen (seems to be 100), pagination should be handled.
```

components/server/src/env.ts
```diff
@@ -110,6 +119,16 @@ export class Env extends AbstractComponentEnv {
     // maxConcurrentPrebuildsPerRef is the maximum number of prebuilds we allow per ref type at any given time
     readonly maxConcurrentPrebuildsPerRef = Number.parseInt(process.env.MAX_CONCUR_PREBUILDS_PER_REF || '10', 10) || 10;
 
+    readonly incrementalPrebuildsRepositoryPassList: string[] = (() => {
+        try {
+            const list = this.parseStringArray('INCREMENTAL_PREBUILDS_REPO_PASSLIST');
+            return list;
```
csweichel, line 125:
```suggestion
            return this.parseStringArray('INCREMENTAL_PREBUILDS_REPO_PASSLIST');
```

components/server/src/github/github-context-parser.ts
```diff
@@ -416,4 +416,56 @@ export class GithubContextParser extends AbstractContextParser implements IConte
...
+
+    public async fetchCommitHistory(ctx: TraceContext, user: User, contextUrl: string, sha: string, maxDepth: number): Promise<string[]> {
+        const span = TraceContext.startSpan("GithubContextParser.fetchCommitHistory", ctx);
+
+        try {
+            if (sha.length != 40) {
+                throw new Error(`Invalid commit ID ${sha}.`);
+            }
+
+            // FIXME: To get more results than GitHub API's max page size (seems to be 100), pagination should be handled.
```
csweichel, line 428:
```suggestion
            // TODO(janx): To get more results than GitHub API's max page size (seems to be 100), pagination should be handled.
```

csweichel: see https://www.notion.so/gitpod/How-we-develop-e8ce7e562478458a9223eb9b797415b2#86e669f8e7df4f1289c9bede0855b5f9

components/server/src/gitlab/gitlab-context-parser.ts
```diff
@@ -339,4 +339,24 @@ export class GitlabContextParser extends AbstractContextParser implements IConte
             repository,
         };
     }
+
+    public async fetchCommitHistory(ctx: TraceContext, user: User, contextUrl: string, sha: string, maxDepth: number): Promise<string[]> {
+        // FIXME: To get more results than GitLab API's max per_page (seems to be 100), pagination should be handled.
```
csweichel, line 344:
```suggestion
        // TODO(janx): To get more results than GitLab API's max per_page (seems to be 100), pagination should be handled.
```

components/supervisor/pkg/supervisor/tasks.go
```diff
@@ -378,6 +378,8 @@ func (tm *tasksManager) watch(task *task, terminal *terminal.Term) {
 		defer stdout.Close()
 
 		fileName := tm.prebuildLogFileName(task)
+		// TODO: If the file already exists (from a parent prebuild), extract its "time saved", and log that below
```
csweichel, line 381:
```suggestion
		// TODO(janx): If the file already exists (from a parent prebuild), extract its "time saved", and log that below
```

csweichel (approved):
Code looks good to me (sans some style things) :)

Issue #3879, opened 2021-04-09
Deleting an active workspace should behave the same way as deleting a stopped one.


apolopena:
### Bug description

**Background**:
When a user deletes a stopped workspace, the workspace is removed right away from the workspace dashboard and this is the desired and expected result especially since workspaces does not have a possible state/status of "Deleting".

**Result:**
This issue is when deleting an active workspace. When deleting an active workspace the status of the workspace changes to "Stopping" for a decent amount of time before disapearing from the dashboard.

This is a new behavior and was not present prior to the first live push of the March 2021 milestone. The workspace use to be removed from the dashboard instantly regardless of the state of the workspace.

### Steps to reproduce

Delete an active workspace.

The workspace is not removed right away and changes to a Status of "Stopping".

### Expected behavior

- The flow of deleting a workspace in the dashboard should be consistent regardless of workspace states/status.
- If a workspace does not have a possible state/status of Deleting then it should be removed from the dashboard right away.

### Example repository

Any

gtsiolis:
Closing this because of the one-line change workaround in https://github.com/gitpod-io/gitpod/pull/4170.  📕 

Some relevant notes for consideration in follow-up issues:
1. The fix in https://github.com/gitpod-io/gitpod/pull/4170 is moving `STOPPING` workspaces to the _All_ list.
2. Pinned workspaces are still visible in the _Active_ list when during deletion time.

Thanks @apolopena for bringing this up and @jankeromnes for the fix! 🏀

PR #4170, opened 2021-05-07
[dashboard] Hide 'stopping' & unpinned workspaces from 'Active'

jankeromnes:
Potential solution/workaround for https://github.com/gitpod-io/gitpod/issues/3879

How to test:
- Delete a running workspace
- It should immediately disappear from the workspaces list, just like when you delete an already-stopped workspace

(It actually gets hidden and not removed, so you'll still see the "stopping" workspace in `Filter: All`, but maybe that's okay for now.)

(EDIT: Another potential difference is, if you delete a running _pinned_ workspace, it will still continue to be displayed as "stopping" in `Filter: Active` for a while. A fix for that could be to immediately un-pin when you delete.)

### DIFF
--- a/components/dashboard/src/workspaces/workspace-model.ts
+++ b/components/dashboard/src/workspaces/workspace-model.ts
@@ -126,7 +126,7 @@ export class WorkspaceModel implements Disposable, Partial<GitpodClient> {
 
     protected isActive(info: WorkspaceInfo): boolean {
         return info.workspace.pinned ||
-            (!!info.latestInstance && info.latestInstance.status?.phase !== 'stopped');
+            (!!info.latestInstance && !['stopping', 'stopped'].includes(info.latestInstance.status?.phase));
     }
 
     public getAllFetchedWorkspaces(): Map<string, WorkspaceInfo> {


.werft/values.dev.yaml
```diff
@@ -35,9 +35,8 @@ components:
     makeNewUsersAdmin: true # for development
     theiaPluginsBucketName: gitpod-core-dev-plugins
     enableLocalApp: true
-    blockNewUsers: true
-    blockNewUsersPasslist:
-    - "gitpod.io"
+    blockNewUsers: false
```
gtsiolis, line 38: **reminder**: To remove this before merging! 😇

gtsiolis (approved):
Hey @jankeromnes! Thanks for giving this a spin.

Behavior for unpinned workspaces looks good. Should be fine to merge this and revert if the different behavior for deleting pinned workspaces causes any confusion.

.werft/values.dev.yaml
jankeromnes: Indeed, split out into https://github.com/gitpod-io/gitpod/pull/4211

PR #4178, opened 2021-05-07
[ws-manager] mknod /dev/net/tun

csweichel:
This change reduces the requirements towards the node a workspace runs on. Instead of mounting `/dev/net/tun` we `mknod` it.

###  How to test
`docker run --rm -it alpine:latest`


aledbf (approved):
LGTM

PR #4179, opened 2021-05-07
Avoid already exists errors for ClusterRoleBinding

aledbf:
to avoid failures installing gitpod in multiple namespaces (gitpod-com)


wulfthimm (approved):
LGTM

Issue #4166, opened 2021-05-07
Make GKE/Grafana links on admin panel work for pluggable workspaces clusters.


meysholdt:
For "new" pluggable workspace clusters, the admin panel does not yet show the links "GKE Node" etc
Example: https://gitpod.io/admin/workspaces/fuchsia-squid-nht0rfp9

![image](https://user-images.githubusercontent.com/239422/117405174-76f1ef00-af0b-11eb-8e3b-743426749cf6.png)

Expectation: Have the bottom row of links as shown in the following screenshot:
Example: https://gitpod.io/admin/workspaces/brown-owl-ydndi0ds

![image](https://user-images.githubusercontent.com/239422/117405237-9426bd80-af0b-11eb-9255-b391b4dda3cd.png)

Part of the cause is that here the mapping from cluster to hyperlinks is hardcoded:
https://github.com/gitpod-io/gitpod/blob/af51739d341bb2245598e275336ae9f730e3b41a/components/dashboard/src/admin/gcp-info.ts#L81-L91

A solution would be to make the `GcpInfo` automatically available for new pluggable workspace clusters:
https://github.com/gitpod-io/gitpod/blob/af51739d341bb2245598e275336ae9f730e3b41a/components/dashboard/src/admin/gcp-info.ts#L129-L134

A solution would involve involve surfacing this information from the the pluggable workspace cluster.
TBD: 
1. should the values come from the cluster itself or be stored as metadata in the "clusters" table in the DB?
2. should all values be surfaces from the backed on can we derive some, e.g. via a naming pattern. For example, there is a lot to derive from the name of a cluster: `gp-prod-ws-eu05-eu-west1`.

princerachit:
Mo and I discussed this and we decided that we will be using the below approach:
1. using server to expose the information from the cluster
2. the cluster name would be injected to ws-manager using helm chart
3. To decide if we need to persist the cluster name in the DB


I will update the approach if it is not feasible/violates the current design.

AlexTugarev:
Hi @princerachit!

As the `WorkspaceInstanceStatus`, which is populated in `ws-manager`, already knows the `node`, would it make sense to provide with the cluster info as well? Deriving `GcpInfo` from that would be feasible.

princerachit:
[...]

PR #4186, opened 2021-05-10
[bugfix] Add database fields to support storage of cluster and project specific details

princerachit:
## What
This is the first PR for the series of changes that will be required to fix https://github.com/gitpod-io/gitpod/issues/4166.

### Changes
1. Update Readme file for typeform changes
1. Add migration script to add `clusterName` and `projectName` fields to `d_b_workspace_instance` table
1. Add `clusterName` and `projectName` to the `WorkspaceInstance` interface and `namespaceName` to the `WorkspaceInstanceStatus` interface
1. Update tests to include the newly added fields
1. Update WorkspaceStarter to have default empty string value while creating the workspace

## Testing
Manually tested by creating a workspace in the werft preview deployment. I could see empty values for the fields in the database.

## Impact
This should not break anything as the fields are not used anywhere other than while creating a workspace.
Database performance impact: TODO


aledbf:
@princerachit  please rebase. The fix for the failure was merged minutes ago

csweichel:
We have a good part of this information in the database already, specifically each `d_b_workspace_instance` has an inaptly named `region` field which points to the cluster where the instance is running.

If we added more metadata to the corresponding `d_b_workspace_cluster` entries (e.g. generic `annotations`)  then we could provide these links without needing to introduce additional (possibly redundant) data on `d_b_workspace_instance`.

princerachit:
Hey Chris, thanks for pointing out the region field. Whereas this PR was to fix only the pluggable cluster admin links issue, I now have a separate question about database calls.

1. Do we prefer to update the models/create new views to reduce database calls? e.g. if I want both the workspace instance details and the information about the cluster, as evident in the current scenario, I would require either run a join query and have a corresponding model in my code OR I can run two db queries for each table.
2. Based on what I understand, we don't keep stale cluster information in the d_b_workspace cluster, right? Why don't we do that? If we clear the old cluster entries, we essentially lose information about where this workspace ran. I am not sure if region field provides that information clearly


Based on what you suggested above do you think this is ok:
1. Add column `annotations` of type JSON to `d_b_workspace` cluster
2. Create a model with the following structure to map to the above field:
```typescript
interface annotations {
  projectName?: string
  clusterName?: string
}

```
3. Above interface would be modified as and when we deem more fields necessary

princerachit:
I have described in detail how I am storing the [cluster urls in this comment on the original issue](https://github.com/gitpod-io/gitpod/issues/4166#issuecomment-847638186)

PR #4190, opened 2021-05-10
[PWC] Prevent deregistering non-empty cluster

csweichel:
This PR prevents accidental de-registration of active clusters:
image


meysholdt (approved):
LGTM & thx!

PR #4201, opened 2021-05-11
Enable fuse-overlayfs

aledbf:


aledbf:
@csweichel @meysholdt can we enable `fuse-overlayfs` as default and test in staging?

csweichel (approved):
LGTM

PR #4202, opened 2021-05-11
[supervisor] Distribute term logs with ringbuffer

csweichel:
instead of channels. Simplifies the code and prevents readers from blocking the terminal writer.

### How to test
1. Open a terminal and produce heaps of output

### Known Defects
- terminals "die" randomly and don't produce any more output. They're still working as they take input, but their output is broken. I've seen it once that a terminal would "die" block right from the beginning. Attempts to `pprof` this have been futile. Probably something to do with the wrap around or read blocking of the ring buffer.

### DIFF
--- a/components/supervisor/pkg/supervisor/tasks.go
+++ b/components/supervisor/pkg/supervisor/tasks.go
@@ -220,13 +220,8 @@ func (tm *tasksManager) Run(ctx context.Context, wg *sync.WaitGroup) {
 		if t.config.Env != nil {
 			openRequest.Env = *t.config.Env
 		}
-		var readTimeout time.Duration
-		if !tm.config.isHeadless() {
-			readTimeout = 5 * time.Second
-		}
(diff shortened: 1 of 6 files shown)


akosyakov:
@csweichel Is it fine if i pick it up and rebase?

csweichel:
> @csweichel Is it fine if i pick it up and rebase? I just did not understand whether it was removed from inbox because it is a PR or it is not a priority, sorry 😅

A bit of both really. I did not think this PR important as anymore because we've solved the disconnection issues differently (use gRPC-web rather than REST). However, technically this is still the better way I reckon - it's just it doesn't work so well and output gets mixed up.

Would this solve some user-impacting issue now (e.g. terminal problems)?

geropl:
> prevents readers from blocking the terminal writer.

This would in general be very nice to have, but is not a concrete issue atm.

> Would this solve some user-impacting issue now (e.g. terminal problems)?

I don't think so.

PR #4206, opened 2021-05-12
[image-builder] Rebuild when upstream has changed

corneliusludmann:
With this commit, the workspace image will be rebuilt when the upstream image (FROM in the .gitpod.Dockerfile) has changed. For this, the server extracts the FROM part of the Dockerfile and the image builder resolves the FROM image and adds the “digest” form of the image to the manifest hash that is used as image tag.

**How to test**

You can test this approach with an image in a registry that is under your control. E.g. push `gitpod/workspace-full` to `your-account/workspace-full` to Docker Hub. Add a repo with the following `.gitpod.Dockerfile`:
```
FROM your-account/workspace-full
RUN date > $HOME/date.txt
```

Open your repo in Gitpod. The workspace image will be built.

Change the image `your-account/workspace-full` on Docker Hub (e.g. push another/older tag of `workspace-full` as `latest`). Open another workspace with your repo without touching the repo content. The workspace image should be rebuilt.

**Impact of this change**

With this change, every workspace image based on a custom `.gitpod.Dockerfile` will be rebuilt since the way the hash tag is calculated has changed (by adding the upstream ref). This is somehow intentional because we want to rebuild with the latest upstream image and don't know if the workspace image is already on the latest upstream (after this change/rebuild we will know this for every workspace image).

We hope that this change will improve image pull times because this would reduce the number of different upstream images (e.g. the reference to `workspace-full` would always pull the latest `workspace-full` layers). However, that will also lead to more frequent image builds (every time an upstream image changes). Repo owners can still prevent this by using a specific image tag in FROM, e.g.:
```
docker.io/gitpod/workspace-full:latest@sha256:42181fe6917aedc92d61d9fbcfbfe28feb5c4d7b5db27cf6817e7a7c2b9df573
```
instead of just `gitpod/workspace-full`.

In fact, this behavior is more as I would expect it. When I don't pin to a specific version, I always get the latest upstream image. If I don't want this, I need to specify a specific image tag.

That also means Gitpod users don't need an artificial rebuild change (e.g. rebuild counter env var etc.) in their Dockerfile to be able to rebuild their image when the upstream image has changed.

A sample repo that is very active on Gitpod but uses a 3-month-old `workspace-full` reference is: https://github.com/tendermint/starport/blob/develop/.gitpod.Dockerfile – Workspace starts for this repo take usually ~60 seconds instead of ~10-20 seconds for repos with latest `workspace-full` image.


geropl:
Thx for the proposal @corneliusludmann ! I really like that we have the code/a PR implementing this long-discussed change.
Also conceptually, I'm 100% in line with the idea behind it, because IMO it clarifies (e.g., makes easier _and_ more straight forward) the semantics of when a workspace is rebuild.

However, I'd refrain from merging it right away: As a result of us postulating "FROM gitpod/workspace-full:latest` for far too long this would lead to many rebuilds for users in situations where they _don't expect_ it to happen.

Ideally, I'd like to see this "backend change" supported by two UX improvements:
 - on rebuild: the users get's to know _why_ they're waiting for a prebuild. Maybe it's as simple as another field on a API shape, I don't know. But I as a user would like to know why again I'm waiting for a base-image build so I can optimize for not running into it again
 - in the IDE: have a hint for `.gitpod.yml` files that reads: "Oh, you still have that `gitpod/workspace-full:latest` in there. Is that what you want? |Yes|, |Link to trade-off|, |Quick-fix to latest tag|"
 
 @jankeromnes I assume that we already produce per-build tags. That's the case, right?

jankeromnes:
Adding my thanks as well! 🙏 I think this is a fantastic change, I personally would love to see it shipped. 🚀

> However, I'd refrain from merging it right away: As a result of us postulating "FROM gitpod/workspace-full:latest` for far too long this would lead to many rebuilds for users in situations where they _don't expect_ it to happen.

Personally, I don't think "occasionally rebuilding when users don't expect it" is a problem.

Docker builds are typically not very long, and seeing them happen again from time to time wouldn't bother me at all. (In fact, I _already_ have the impression that my Docker images sometimes rebuild when I'm not exactly sure why, but this has never bothered me. However, slow workspace starts do bother me, as well as when I discover that my workspace is using very old / deprecated / insecure tools.)

As a user, I see "occasional unexpected Docker rebuilds" as more desirable than "N-years-old image & very slow start-ups for some reason".

> - in the IDE: have a hint for .gitpod.yml files that reads: "Oh, you still have that gitpod/workspace-full:latest in there. Is that what you want? |Yes|, |Link to trade-off|, |Quick-fix to latest tag|"
> 
> @jankeromnes I assume that we already produce per-build tags. That's the case, right?

Indeed, when we release new images, we [also tag them with the workspace-images commit they are based on](https://github.com/gitpod-io/workspace-images/blob/ad116b92c9d9e4f9a06932a93f7d9210a25872d2/.circleci/dazzle_build_image.sh#L40-L43).

However, I would rather not encourage users to effectively strand themselves on a forever-bitrotting image (i.e. I don't think "never seeing unexpected Docker rebuilds" is a good enough trade-off in exchange for "a workspace image that forever becomes more and more outdated, big, slow and insecure").

geropl:
> As a user, I see "occasional unexpected Docker rebuilds" as more desirable than "N-years-old image & very slow start-ups for some reason".

For me as a user, this actually bothers me a lot: I expect my workspace to be reliably reproduceable (up to my current understanding of it) - and it has been like this in the past. If we now deploy the change of behavior, I'd rather tell users about it openly, with a solution at hand, and not leaving them frustrated with a shiny new base image that breaks their stuff (happen to me personally multiple times for my personal projects. Had I not more ties with Gitpod, I had abondened it long ago, multiple times. Just because it does not seem to care to explain what it does - and thus does not respect me and my time as a user.)

IMO we're already making it very hard for the average user to "grok"/consciously use Gitpod. Just because we drive changes like that (which are awesome!!) but then "forget" or ignore the fact that we need to be more transparent about what happens behind the scenes.

A services that does not value my time is not even considered to be used.
(Of course this also holds for workspace startup as well, but there I "know" what happens, even if I'm using Gitpod the first time. This is also bad and should be improved, but, again, these features are not tight with one another.)

Also, I'm not against the change in any way (IMO I made the point above already). I just think we when we deploy it, it should alongside one or two little UX improvements which turn it from a good thing to have into a nice, round feature.

jankeromnes:
Thanks Gero, these are some fair points. I agree that env reproducibility is important, and that it would be great if Gitpod can explain why a rebuild is needed, in order to avoid surprising users with unexpected behavior.

I still feel that, fundamentally, "fresh workspace image + consistently fast workspace starts" is more important and valuable for our users than "no unexpected rebuilds + slowly deprecating workspace image".

For example: Sure, if a custom Dockerfile is built once (say on top of `ubuntu`) and then never again, the repo's workspaces will always behave the same way. However, I feel like we're "hiding" several problems to the user:
- While the setup will always appear to work the same way on that repo, maybe the underlying image will change (e.g. `ubuntu` moves to a new LTS), and after a while your workspaces will have very old tools & dependencies and things will break in surprising/counter-intuitive ways.
- This reproducibility is only per repo (e.g. if you create a new fork with the exact same setup 1 year later, suddenly you'll get very different workspaces), and only as long as we keep our caches alive (we probably don't want to keep everything indefinitely)
- Maybe `apt-get` will even stop working at all (this actually happened to Gitpod users when an old Ubuntu reached end-of-life, and all upstream packages got deleted by Ubuntu). Getting out of such a surprising situation is very tricky if you don't know the Gitpod Dockerfile checksum caveat (new workspace won't help you there).

Another thing I really like about this Pull Request is that it aligns the behavior of:

```yml
image:
  file: Dockerfile
```

```Dockerfile
FROM ubuntu
```

(currently never changes unless you change at least one character in the Dockerfile)

with the behavior of:

```yml
image: ubuntu
```

(currently changes as soon as a new base image is available)

mikenikles:
I think this is a great PR and fully support the new behaviour! We do need to communicate the change to users though to make sure there are no surprises.

With my current focus on improving documentation and thinking through the user experience in detail, here's my view on this topic.

**Environment stability**
_With a Gitpod default base image_
As a user, I expect to get _non-breaking changes_ automatically (e.g. bug fixes, security patches, etc.).
Breaking changes need to be communicated and ideally opt-in. Maybe this requires a new property in `.gitpod.yml` to specify the version of `workspace-full` 🤔.
<details>
  <summary>Example</summary>
 I work on my project today. Tomorrow, Gitpod updates the `workspace-full` image with a breaking change (e.g. Ruby v2 to v3). Next week, my dev environment is broken and I don't know why.
</details>

_With a custom `.gitpod.Dockerfile`_
As a user, I expect the `FROM` instruction to behave the way it always does, i.e. if no tag is specified, `latest` is used which means it's up to me to deal with sudden breaking changes. If I want stability, I use a tag and updating that is now my responsibility.
<details>
  <summary>Example</summary>
 I work on my project today. Tomorrow, Gitpod updates the <code>workspace-full</code> image with a breaking change (e.g. Ruby v2 to v3).

If I use `FROM gitpod/workspace-full`, I expect this breaking change and will deal with it.
If I use `FROM gitpod/workspace-full:xyz`, my dev environment continues to work and I upgrade the tag when it suits my workflow. Ideally, I see a warning that my base image is out of date.
</details>

**Workspace startup time**
As a user, I want workspaces to start as quickly as possible, regardless of whether I use a custom docker image or not. How we as Gitpod achieve this is an implementation detail and should not impact a user's decision as to what they use for their image.

**Ensure Gitpod works with custom images**
If we apply custom code on top of a plain Ubuntu image to make Gitpod work as per Slack thread, we need to document what these changes are so that anyone with a custom image can apply these changes to their docker image. The easiest way may be a `gitpod-base` docker image that is Ubuntu + changes required to run Gitpod.


I hope this doesn't hijack the PR, but given the conversations above and in Slack I felt it's worth sharing my full view here 😄.

svenefftinge:
Wow, very long discussion. At some point, a quick call could help to create alignment :-)

I'm with Gero on this and believe the main problem wouldn't be an updated image but the surprising docker build (which possibly fails). Dockerbuilds should only happen when someone actually works on a dockerfile but not out of a sudden standing in between me and the thing I actually wanted to do. I might not even be the right person to deal with a docker build fail that happens out of a sudden.

Please consider this design doc as well: https://www.notion.so/gitpod/Image-Layers-ab02c50f44ef4ad6b1a6014a5dadc00b

I.e. `workspace-full` is going to be replaced by a dynamically created image composed from pre-built layers.

svenefftinge:
We could add a prompt in the loading UI, that tells a user about the update of the parent image and lets her choose.

corneliusludmann:
Thanks for the active discussion. I can understand both sides. I'm a big fan of reproducibility and reproducible builds. In general, I like to pin versions of dependencies when I would like to have reproducibility and have the time to take care of upgrading upstream versions. But I also like to use just the latest in some cases where I can't spend the time to keep track of the upstream versions. I always want to get the latest security updates and only worry about it when it breaks.

**What I expect as a user**

> Dockerbuilds should only happen when someone actually works on a dockerfile but not out of a sudden

I don't believe that a workspace image rebuild is unexpected when the upstream image has been changed. As a user, I would expect when I use `FROM gitpod/workspace-full:latest` that my workspace image always uses the latest image. That's how Docker works. And that's how e.g. GitLab CI works (when I set `ubuntu:latest` as CI build image and the build tools I use in the CI build are not part of `ubuntu:latest` anymore at some point, my build pipeline will fail). When I want to have a reproducible Docker build I always need to pin the upstream image version. That's more or less was @mikenikles describes as expectations above.

In Gitpod, a user does not have the choice currently. Gitpod always sticks with a specific version which is in fact currently intransparent to the user. It's the version of the first workspace start of any random user after the latest commit of the `.gitpod.Dockerfile`. (OK, to be fair, usually, the user that modifies the `.gitpod.Dockerfile` also tests the changes and is actually the user that triggers the build.) And actually, we had already users that wonder why they don't have the latest workspace-full changes in their workspaces with custom Dockerfiles where we tell them, they should introduce an artificial env variable with a counter in their Dockerfile that they need to change in order to get an image re-build. Sounds wired.

With the change in this PR, the user has the choice (as I am used to it with Dockerfiles): I can stick with a specific version by using a proper image tag or I can go the lazy way with fetching the latest (security) updates automatically by using the `latest` tag.

So, in my opinion, a workspace image re-build with the upstream images changes is not surprising—I would say that the opposite is the case.

**Practical considerations regarding workspace image build failures**

Yes, there is a risk that the image build fails when we update the workspace image. However, that shouldn't happen that often and we should try to avoid it. And we can improve the situation with better documentation of what users can expect from the workspace-full image. But currently, when users get stuck with old workspace images, this breaks also from time to time (or at least leads to poorer performance). E.g. we fix/improve some issues in our image, a VSCode extension needs a new tool version, etc., every user needs to do an artificial change in their Dockerfile.

> [workspace image build failures] standing in between me and the thing I actually wanted to do. I might not even be the right person to deal with a docker build fail that happens out of a sudden.

That's indeed a good point regarding practical impacts with workspace image build failures. We as Gitpod don't want that users cannot use Gitpod since the workspace image build fails. Especially, when the user is not the maintainer of the `.gitpod.Dockerfile`. What we can do pretty easily is to add another button next to the “Start with Default Image” button that starts the workspace with the latest successfully built workspace image in case the workspace image build has been failed.

I think that would be a good compromise. However, that doesn't cover one thing: When the workspace image build finishes successfully but the changes in the upstream image e.g. introduce a new version of something that is incompatible with the repo content (see new Ruby version example of Mike). For this, maintainers need to set specific image tags (or wait for the prebuild image layers thing Sven mention above).

(It also does not cover that the user needs to wait for an image build from time to time. But like @jankeromnes, I don't really find that critical. The advantages outweigh the disadvantages, IMHO. However, the “Start with latest successfully built image” button could also skip a running workspace build.)

**Further improvements**
All this does not mean that we should not improve the situation by:
- improving documentation (Hello @mikenikles :wave:)
- improving tooling support for maintaining `.gitpod.Dockefile`
- implementing the awesome idea of rebuild image layer thing (However, this does not solve the latest vs. pinned version thing complete, IMHO. But that's another discussion.)

However, I think that his PR (together with the proposed “Start with latest successfully built image”) would be a good minimal valuable change (MVC).

**Summary**
- Conceptually, the behavior with this PR is more what I would expect as a user.
- With this PR, the user has the choice between version pinning and the lazy “just use latest”. Currently, the users do not have the choice.
- However, from a practical point of view, we do not want workspaces to be unusable when workspace image builds break.
- A good compromise would be a “Start with latest successfully built image” button next to the “Start with Default Image” button, IMHO.

And of course, this should be part of the “What's new” dialog.

What do you all think?

components/gitpod-protocol/src/protocol.ts
```diff
@@ -595,6 +595,8 @@ export interface WorkspaceImageSourceDocker {
     dockerFileHash: string
     dockerFileSource: Commit
     dockerFilePath: string
+    /** The FROM part in the Dockerfile (optional). */
+    dockerFileFrom: string | undefined
```
jankeromnes, line 599: Random thought: If you include the resolved upstream hash in the `dockerfileFileHash` function, I think you won't need to modify protocol shapes, nor will you have to remember to compare `dockerFileFrom` in all relevant places.

corneliusludmann:
Proposal for continue with latest build button: 6f3dcc8b9520656f5ba8b0470dbadaf048877702

Peek 2021-05-17 22-35

csweichel:
I'm also very much in favour of this behaviour change. 

**Re: surprising image builds**
We already have the situation where "an image build" happens without a user's change: when we change the Gitpod layer. Surely this is not the same as a base image rebuild, but from a user's perspective not easy to tell apart - annoying as it is.

**Re: reproducible builds** 
Gitpod currently tries to keep workspace configuration stable across instances (never mind the Gitpod layer). If we were to have this update mechanism only for new workspaces we would maintain the "reproducible nature" of workspaces and still get the updates. Currently the `FROM gitpod/workspace-full:latest` freezes the image at an arbitrary version, so re-freezing it at another arbitrary version is no loss of control to the user when starting a new workspace. Most users wouldn't even know the mechanism behind, or if someone else changed the Dockerfile.
TL;DR: update only for new workspaces, not for re-starting old ones.

components/gitpod-protocol/src/protocol.ts
corneliusludmann: That's indeed an interesting idea. However, the problem is that we need to resolve the FROM image somewhere (to get the revision SHA hash) and as far as I know, we cannot resolve this image in the server component easily but in the image-builder component.

PR #4207, opened 2021-05-12
Add leeway run script to unblock user in DB

corneliusludmann:
Since we currently block all new users in Gitpod preview environments that do not have set their gitpod.io email as primary email in GitHub, this change adds a small script to unblock users easily.

Just run `$ leeway run components:unblock-user` to unblock a user.

leeway-unblock-user

### DIFF
--- a/components/BUILD.yaml
+++ b/components/BUILD.yaml
@@ -107,3 +107,34 @@ scripts:
       - test/**/*
     script: |
       leeway exec --filter-type go -v -- go mod tidy -v
+  - name: unblock-user
+    deps:
+      - components/service-waiter:app
+    script: |
+      export DB_USERNAME=root
+      export DB_PASSWORD=test
+      kubectl port-forward statefulset/mysql 3306 &
+      PID=$!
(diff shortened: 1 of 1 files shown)


jankeromnes (approved):
Cool, many thanks @corneliusludmann! This makes the situation a bit more convenient, but still requires you to fire up a workspace for a PR in order to unblock yourself before you can test it (not very convenient).

I still feel like the best solution would be to:
- not block all users
- also not make the first user admin

(then maybe have a similar script to grant admin?)

But I can still approve the PR if you'd like to have it :slightly_smiling_face: (no harm in having an extra script 🚀)

corneliusludmann:
Thanks, @jankeromnes. And I totally agree. But as long as we don't have a decision here, I'm annoyed that I always have to unblock myself cumbersomely in the DB. If we change that, we can easily change the script to grant admin instead. :+1:

jankeromnes:
Fixed in https://github.com/gitpod-io/gitpod/pull/4211

PR #4211, opened 2021-05-12
[dev] Don't make new users admin and re-disable BLOCK_NEW_USERS

jankeromnes:
Implements https://github.com/gitpod-io/gitpod/pull/4207#pullrequestreview-657778708

### DIFF
--- a/components/server/src/user/user-service.ts
+++ b/components/server/src/user/user-service.ts
@@ -144,7 +144,7 @@ export class UserService {
 
             newUser.blocked = !canPass;
         }
-        if (!newUser.blocked && (isFirstUser || this.env.makeNewUsersAdmin)) {
+        if (!newUser.blocked && this.env.makeNewUsersAdmin) {
             newUser.rolesOrPermissions = ['admin'];
         }
     }
--- a/.werft/values.dev.yaml
+++ b/.werft/values.dev.yaml
@@ -28,13 +28,12 @@ tracing:
 components:
   server:
     replicas: 1
-    makeNewUsersAdmin: true # for development
+    makeNewUsersAdmin: false
     theiaPluginsBucketName: gitpod-core-dev-plugins
     enableLocalApp: true
     enableOAuthServer: true
-    blockNewUsers: true
-    blockNewUsersPasslist:
-    - "gitpod.io"
+    blockNewUsers: false
+    blockNewUsersPasslist: []
 
   registryFacade:
     daemonSet: true


svenefftinge:
so how do I become admin now?

jankeromnes:
By updating the MySQL database?

corneliusludmann (approved):
Code-wise, this change looks good. Don't know whether we need consensus to avoid an edit war, though. What do you think?

wulfthimm:
@svenefftinge You can always override that setting in your branch and enable becoming root on your first visit.

corneliusludmann:
> You can always override that setting in your branch and enable becoming root on your first visit.

Indeed. That's true. Thanks for the hint. However, I don't think that overriding such a setting in a branch is a good solution because …
a) there is a risk that you accidentally merge this eventually, and
b) if you remove this when your branch is ready, a reviewer needs this change as well.

fntlnz (approved):

jankeromnes:
Note: I haven't merged this yet, because the very first user still becomes admin even with this change:

https://github.com/gitpod-io/gitpod/blob/dc892f0333e18f0bb713e32ee71493e8c0378c5f/components/server/src/user/user-service.ts#L148-L150

components/server/src/user/user-service.ts
```diff
@@ -144,7 +144,7 @@ export class UserService {
 
             newUser.blocked = !canPass;
         }
-        if (!newUser.blocked && (isFirstUser || this.env.makeNewUsersAdmin)) {
```
AlexTugarev, line 147: this is to enable "first-user-is-admin" in self-hosted installation. how would you suggest to proceed with self-hosted?

jankeromnes: Good question. The removed behavior isn't ideal, because anyone can become admin by being the first to log into a fresh Gitpod deployment (e.g. from a PR).

Instead of always making the first user admin, I'd recommend one of the following options:
- Update your user in the DB
- Temporarily enable "make new users admin" when you install self-hosted
- Implement a "new user admin passlist" array that makes new users admin based on email domain

AlexTugarev: Actually this breaks the installation story of self-hosted installations entirely. There is no need for any UI if the installer need to edit deployments and delete pods, or connecting to the db in background.

 As far as I understand, we want to protect preview environments, right? In that case let's disable setting of the admin flag for the preview environments (aka `kubestage === 'devstaging'`). see https://github.com/gitpod-io/gitpod/blob/c5641184da7f5bb1204b2b038264742fe2f72bad/components/gitpod-protocol/src/env.ts#L18

csweichel:
considering the recent comment and that we've happily lived with the current behaviour for a few months, I'd say we stick with what we got here.

PR #4217, opened 2021-05-14
Generation of grpc code should also include typescript components

aledbf:


geropl (approved):
Thank you for the quick PR, works like it should!

PR #4219, opened 2021-05-14
[ws-daemon] Refactor full workspace backup

aledbf:


components/ws-daemon/pkg/iws/iws.go
```diff
@@ -254,7 +254,8 @@ func (wbs *InWorkspaceServiceServer) PrepareForUserNS(ctx context.Context, req *
 		}
 
 		return &api.PrepareForUserNSResponse{
```
csweichel, line 256: We should force this fsshift type when the workspace is FWB. This way we need to only run fusefs for non-fwb workspaces.

csweichel (approved):
Awesome. LGTM

PR #3704, opened 2021-04-01
[dashboard] github app installation

svenefftinge:
## How to Test

 - Go to https://github.com/apps/gitpod-dev-app
 - Click configure
 - go through the flow
 - verify that prebuilds are triggered when pushing to repos you configured the app on.

Issue #3702, opened 2021-04-01
Implement /install-github-app flow


svenefftinge:
We need to implement the flow for installing the Github app.

Currently it is:
1) https://github.com/apps/gitpod-io (click configure)
2) `/install-github-app`
<img width="1634" alt="Screenshot 2021-04-01 at 12 28 42" src="https://user-images.githubusercontent.com/372735/113281666-2f61bd00-92e6-11eb-8bb8-463268fca0a5.png">

3) Agree on github
4) confirmation
<img width="680" alt="Screenshot 2021-04-01 at 12 28 54" src="https://user-images.githubusercontent.com/372735/113281701-3688cb00-92e6-11eb-9dd7-65ea60f28f30.png">

### Designs

| Install | Warning | Success | Error |
|-|-|-|-|
| ![image](https://user-images.githubusercontent.com/120486/113303095-67c6c280-9309-11eb-87d1-2755cf4f5787.png) | ![image](https://user-images.githubusercontent.com/120486/113306057-57fcad80-930c-11eb-9e46-7b1604931453.png) | ![image](https://user-images.githubusercontent.com/120486/113303211-875deb00-9309-11eb-928b-574c7b246bb2.png) | ![image](https://user-images.githubusercontent.com/120486/113303116-6e553a00-9309-11eb-9d46-3bd6fc3b5374.png) |

See also [design specs](https://www.figma.com/file/6IsTSA9VwckZLPvHN8mJQT/Install-GitHub-App-Flow-3702?node-id=0%3A1).

svenefftinge:
There are currently three different outcomes:

```
<Dialog open={this.state.showDialog === 'cancel'}>
                    <DialogTitle>Cancel installation?</DialogTitle>
                    <DialogContent>
                        <DialogContentText>Canceling the installation here just prevents Gitpod from reacting to the events it receives from GitHub.
                        To fully undo the installation, please uninstall the <a href={`https://github.com/settings/installations/${this.installationId}`} target="_blank" rel="noopener noreferrer">Gitpod app in GitHub</a>.</DialogContentText>
                    </DialogContent>
                    <DialogActions>
                        <Button onClick={() => this.setState({ showDialog: 'none' })} variant="outlined" color="primary">Abort</Button>
                        <Button onClick={() => window.location.href = `https://github.com/settings/installations/${this.installationId}`} variant="outlined" color="secondary" autoFocus>Ok</Button>
                    </DialogActions>
                </Dialog>
[...]

PR #3693, opened 2021-04-01
[dashboard] Fix remaining issues in new start screen

jankeromnes:
Part of https://github.com/gitpod-io/gitpod/issues/3689

Fixes https://github.com/gitpod-io/gitpod/issues/3654 https://github.com/gitpod-io/gitpod/issues/3687 https://github.com/gitpod-io/gitpod/issues/3707

## Errors

Screenshot 2021-04-02 at 10 28 39

Screenshot 2021-04-02 at 10 32 33

Screenshot 2021-04-01 at 17 01 56

## Limit Reached

Screenshot 2021-04-02 at 10 39 42

Screenshot 2021-04-01 at 15 59 08

## Stopped

Screenshot 2021-04-02 at 10 41 22

Screenshot 2021-04-02 at 10 41 35

Issue #3301, opened 2021-03-01
[Dashboard] Workspace Start Page


svenefftinge:
As part of the dashboard redesign we are reimplmenting the workspace start screen.

The screen shall communicate the current state of a workspace to users. 
As a first step we need to list all possible states and create designs for them.

This is part of https://github.com/gitpod-io/gitpod/issues/3300. 🍕 

### Designs

| Step 1 | Step 2 | Step 3 |
|-|-|-|
| ![image](https://user-images.githubusercontent.com/120486/110942166-19458a00-8342-11eb-8285-559b807b098f.png) | ![image](https://user-images.githubusercontent.com/120486/110942170-19de2080-8342-11eb-8144-5601e131df7c.png) | ![image](https://user-images.githubusercontent.com/120486/110942174-1b0f4d80-8342-11eb-9ca8-e09da9b455cb.png) |

| System Error |  System Error (Spacing)
|-|-|
| ![image](https://user-images.githubusercontent.com/120486/110942203-28c4d300-8342-11eb-837e-df88d251108f.png) | ![image](https://user-images.githubusercontent.com/120486/110942216-2cf0f080-8342-11eb-9bd0-dc057704256d.png) |

| Existing Workspaces | Existing Workspaces (Spacing) |
|-|-|
| ![image](https://user-images.githubusercontent.com/120486/110942381-6c1f4180-8342-11eb-9c15-12436fbe2f35.png) | ![image](https://user-images.githubusercontent.com/120486/110942331-56aa1780-8342-11eb-97f1-e2c57041a2ac.png) |

| Parallel Workspaces | Page Not Found |
|-|-|
| ![image](https://user-images.githubusercontent.com/120486/110942435-7c372100-8342-11eb-8477-2ee2242d6901.png) | ![image](https://user-images.githubusercontent.com/120486/110942458-80633e80-8342-11eb-9f76-3ed31b83444e.png) |

See also [design specs](https://www.figma.com/file/FZVZIKi4oAOvYf9NVZnHWB/Workspace-Start-Page-3301).

svenefftinge:
states and information to surface:
https://www.notion.so/gitpod/Redesign-Loading-Screen-5e2f3a45cb5446088566af1b30cde55f

gtsiolis:
I've added a first draft of the design specs in the issue description. /cc @jankeromnes @csweichel @svenefftinge

svenefftinge:
[...]

PR #4222, opened 2021-05-17
Initial version of OAuth server to manage access to Gitpod workspaces

rl-gitpod:


rl-gitpod:
### What does this PR do?

* Adds an OAuth2 server endpoints to enable standard OAuth2 clients to interact with Gitpod and be granted access to a users workspaces
* Specifically, it utilises PKCE (Proof Key for Code Exchange) to reduce the potential for MITM attacks to gain access to tokens
* Exposes /api/oauth/authorization and /api/oauth/token for the client flow to use
* /api/oauth/authorization 
   * will open a browser and redirect to the standard Gitpod provider login (Github, Gitlab, etc) if not logged in (opening a browser works in Linux (requires xdg-open)/OSX/Windows
   * will open a browser and ask user to grant permission for the OAuth2 client to access their workspaces if not already done so
   * opened tabs are automatically closed 
* /api/oauth/token if given correct authenticaion code and PKCE params will return a Gitpod token in standard OAuth2 format:
```
{
  "access_token": "eyJhb...",
  "expires_in": 86400,
  "refresh_token": "eyJhb...",
  "scope": "function:getWorkspace, etc",
  "token_type": "Bearer"
}
```
* Extends the existing login code to add support for the required redirection - currently limited only to the new endpoints on the same server
* Adds a hacky way to test this to the existing client

components/dashboard/src/App.tsx
```diff
@@ -118,6 +119,13 @@ function App() {
     if (shouldWhatsNewShown !== isWhatsNewShown) {
         setWhatsNewShown(shouldWhatsNewShown);
     }
+    if (window.location.pathname.startsWith('/oauth-approval')) {
```
svenefftinge, line 122: I think this should be added below as a route

rl-gitpod: It was originally but you end up with the menu across the top which felt out of place in the flow, allowing people to click away w/o any resolution. 

Over time we will need to add a page to list approvals and allow people to delete them, but this part of the flow is more modal.

components/dashboard/src/Login.tsx
```diff
@@ -57,7 +57,7 @@ export function Login() {
             await openAuthorizeWindow({
                 login: true,
                 host,
-                onSuccess: () => updateUser(),
+                onSuccess: (payload?: string) => authorizeSuccessful(payload),
```
svenefftinge, line 60: minor: declaring the function after it is used seems a bit uncommon to me. Also, the value can be used directly as the signature matches (i.e. `onSuccess: authorizeSuccessful`)

svenefftinge:
@gtsiolis can you help with the design of this page?
Screenshot 2021-05-18 at 09 34 18

gtsiolis:
> @gtsiolis can you help with the design of this page?

@svenefftinge Yes! How can I see this page in aciton? I tried following the steps in https://github.com/gitpod-io/gitpod/pull/4222#issuecomment-842680975 but only see the following then opening port `8080`. /cc @rl-gitpod 

Screenshot 2021-05-18 at 12 32 29 PM

rl-gitpod:
> > @gtsiolis can you help with the design of this page?
> 
> @svenefftinge Yes! Although not needed, how could I see this page in aciton? I'd like to also see the interaction for this (if the tab is closing, etc). I tried following the steps in [#4222 (comment)](https://github.com/gitpod-io/gitpod/pull/4222#issuecomment-842680975) but only see the following then opening port `8080`. /cc @rl-gitpod
> 
You need to download the local-app in that screen and keep on with the directions.
I'll record a loom which should help.

rl-gitpod:
https://www.loom.com/share/9442ca1e2ab6408c93ce8c3390cf20e4

components/dashboard/src/Login.tsx
rl-gitpod: Thanks - updated & moved

components/gitpod-db/src/typeorm/auth-code-repository-db.ts
```diff
@@ -0,0 +1,84 @@
...
+    protected async getEntityManager(): Promise<EntityManager> {
+        return (await this.typeORM.getConnection()).manager;
+    }
+
+    async getOauthAuthCodeRepo(): Promise<Repository<DBOAuthAuthCodeEntry>> {
+        return (await this.getEntityManager()).getRepository<DBOAuthAuthCodeEntry>(DBOAuthAuthCodeEntry);
+    }
+
+    public async getByIdentifier(authCodeCode: string): Promise<OAuthAuthCode> {
+        log.info(`getByIdentifier ${authCodeCode}`);
```
csweichel, line 32: debugging leftover? I reckon this would get pretty loud in prod :)

components/gitpod-db/src/typeorm/user-db-impl.ts
```diff
@@ -350,6 +356,95 @@ export class TypeORMUserDBImpl implements UserDB {
...
+        return;
+    }
+    public async extraAccessTokenFields?(user: OAuthUser): Promise<ExtraAccessTokenFields | undefined> {
+        // No extra fields in token
+        log.info(`getUserByCredentials ${JSON.stringify(user)}`);
+        return;
+    }
+
+    // OAuthTokenRepository
+    async issueToken(client: OAuthClient, scopes: OAuthScope[], user?: OAuthUser): Promise<OAuthToken> {
```
csweichel, line 382: how does this relate to `AuthCodeRepositoryDB.issueToken`?

components/gitpod-db/src/typeorm/user-db-impl.ts
```diff
@@ -350,6 +356,95 @@ export class TypeORMUserDBImpl implements UserDB {
...
+            log.info(`persist access token update: ${JSON.stringify(dbToken)}`);
+            return repo.updateById(tokenHash, dbToken);
+        } else {
+            var user: MaybeUser;
+            if (accessToken.user) {
+                user = await this.findUserById(accessToken.user.id)
+            }
+            dbToken = {
+                tokenHash,
+                name: `local-app`,
```
csweichel, line 427: Would it make sense to include the name of the requesting application here? This way we could have multiple tokens issued using OAuth.

components/gitpod-protocol/src/protocol.ts
```diff
@@ -100,6 +100,9 @@ export interface AdditionalUserData {
     ideSettings?: IDESettings;
     // key is the name of the news, string the iso date when it was seen
     whatsNewSeen?: { [key: string]: string }
+    // key is the name of the OAuth client i.e. local app, string the iso date when it was approved
+    // TODO(rl): provide a management UX to allow rescinding of approval
```
csweichel, line 104: TIL: _rescinding_ is a synonym for _revoking_ :)

csweichel: Do I understand correctly that we have two levels of approval here: 
1. the OAuth client
2. the tokens issued for a client?

components/server/src/oauth-server/oauth-controller.ts
```diff
@@ -0,0 +1,148 @@
...
+
+@injectable()
+export class OAuthController {
+    @inject(Env) protected readonly env: Env;
+    @inject(UserDB) protected readonly userDb: UserDB;
+    @inject(AuthCodeRepositoryDB) protected readonly authCodeRepositoryDb: AuthCodeRepositoryDB;
+
+    get oauthRouter(): express.Router {
+        const router = express.Router();
+        if (this.env.enableOAuthServer) {
```
csweichel, line 26: nit (happy path on the left):

```suggestion
if (!this.env.enableOAuthServer) {
    log.warn('OAuth server disabled!');
    return router;
}
```

csweichel (commented):
Other than the generous display of logging statements the code LGTM. I've gone through the testing instructions and things work as advertised. Pending @gtsiolis's comments this could land in main I reckon :)

gtsiolis:
> Note that the gitpod.io page is only there to keep chrome on the screen during the recording ...

Thanks for clarifying @rl-gitpod!

---

I've made two variations we could use here in case one of these two makes more sense, following two relevant existing implementations. Feel free to adjust text per need. A couple of notes or questions:
1. Since this action is associated with an account, including the account avatar can help the user understand which kind of data will be exposed with this authorization.
2. If possible, we could include the workspace name (friendly identifier) in the message body.
3. Title message could also read _Authorize Gitpod App_, _Authorize Gitpod CLI_, etc.
3. The first design follows the specs from the Install GitHub App flow, see _Install_ in https://github.com/gitpod-io/gitpod/issues/3702.
4. The second design follows the modal component as used in the workspace start page when reaching a limit, see _Parallel Workspaces_ https://github.com/gitpod-io/gitpod/issues/3301.
5. If both seem appropriate, let's go with Proposal A. Otherwise, I'll defer this to @svenefftinge.

| Proposal A (Full Page) | Proposal B (Modal) |
|-|-|
| image | image |

See also [relevant specs](https://www.figma.com/file/WeaK9mLXKJYDLxwEh2iHib/Authorize-Local-App-4222?node-id=0%3A1) and relevant existing implementations in https://github.com/gitpod-io/gitpod/pull/3704 and https://github.com/gitpod-io/gitpod/pull/3693.

components/gitpod-db/src/typeorm/auth-code-repository-db.ts
rl-gitpod: That assumes lots of people will be accessing it initially ;-)

Agreed there is a lot of debug logging - I'll take a pass at removing some of them.

components/gitpod-db/src/typeorm/user-db-impl.ts
rl-gitpod: I think you meant AuthCodeRepositoryDB.issueAuthCode?

The auth code is returned by the authorization process, which the client uses to request, along with the PKCE creds for validation, the actual API token - in our case the Gitpod token.

components/gitpod-db/src/typeorm/user-db-impl.ts
rl-gitpod: Great point.

Hard-coding leftovers replaced with the client name.

components/gitpod-protocol/src/protocol.ts
rl-gitpod: Yes.

Given we are using standard OAuth flows, we have:

1. is approval for Gitpod to _issue_ api tokens for a client
2. the actual api tokens issued for a specific client. There may well be more than 1 - different laptops, etc.

components/server/src/oauth-server/oauth-authorization-server.ts
```diff
@@ -0,0 +1,62 @@
...
+import {
+  inMemoryClientRepository,
+  inMemoryScopeRepository
+} from "./repository";
+
+const clientRepository = inMemoryClientRepository;
+const scopeRepository = inMemoryScopeRepository;
+
+// TODO(rl) - get this from external secret
+const jwtService = new JwtService("secret secret secret");
```
AlexTugarev, line 18: please don't forget about this TODO. 
is it correct that this service is used to produce the issued codes in the end? see: https://github.com/jasonraimondi/typescript-oauth2-server/blob/153c83bd64940a03ba31c833d4e3b2aeb4ce03cd/src/grants/auth_code.grant.ts#L57

rl-gitpod:
Thanks @gtsiolis!

> 1. Since this action is associated with an account, including the account avatar can help the user understand which kind of data will be exposed with this authorization.

What we are doing here is the same as the 'Consent form' for any/all OAuth implementations e.g. [here](https://i.stack.imgur.com/q4RUd.png) or [here](https://rankmath.com/wp-content/uploads/2020/03/rank-math-OAuth-screen.jpg) 
They tend to be more modal in nature, but not always.

Ultimately we will want to itemise the types of access (scopes) they are approving, but we don't have a good sense of what they should be at this point so we are asking for a general access approval.

> 2. If possible, we could include the workspace name (friendly identifier) in the message body. We limit the scope of this authorization to a specific workspace, right?

It is for all workspaces atm. That will likely change in the future, but not for the first iteration and possibly more.

> 6. If both seem appropriate, let's go with Proposal A. Otherwise, I'll defer this to @svenefftinge.

The 'Consent form' in other systems tends to be modal. It is also simpler to implement :-)

If it is not modal, having the ability to jump to the dashboard/account/etc without completing an action would need to be handled. If we assume that it doesn't mean either yes or no (and therefore we don't need to update the db to reflect that) it could just be a timeout i.e. easy.

> 8. If these style changes can land into the same milestone, it should be fine to leave these out of the scope of this PR and open a new PR instead. 🛹

If we have a decision on A vs B, I'll take a look tomorrow.

components/server/src/oauth-server/oauth-authorization-server.ts
rl-gitpod: This is less important than it might appear. 

The service is used to encrypt and [decrypt](https://github.com/jasonraimondi/typescript-oauth2-server/blob/5b60bc0d4ad41a19dd6033e4d242b7db13203f58/src/grants/auth_code.grant.ts#L55) the auth code which is returned untouched by the client and the value is only used internally. In saying that, I agree it should be taken from an external source to make it less likely that external parties can create a valid auth code. Importantly, PKCE is the key here - it might be a valid code but they won't have the valid challenge/verifier that is required alongside it.

On the client side, the jwtService secret is used for _validation_ of the returned token to detect MITM among others, and involves a shared secret between the client and the server. The actual token can be decoded and used without knowing the secret (e.g. see the [online decoder](http://calebb.net/). Given that we are communicating with only our server and that having a shared secret in public code for the local app client is not secret, then @csweichel and I decided that we can safely ignore this validation step.

components/gitpod-db/src/typeorm/auth-code-repository-db.ts
```diff
@@ -0,0 +1,77 @@
...
+        const authCode = authCodes.length > 0 ? authCodes[0] : undefined;
+        if (!authCode) {
+            throw new Error(`authentication code not found`);
+        }
+        return authCode;
+    }
+    public issueAuthCode(client: OAuthClient, user: OAuthUser | undefined, scopes: OAuthScope[]): OAuthAuthCode {
+        const code = crypto.randomBytes(30).toString('hex');
+        log.info(`issueAuthCode: ${JSON.stringify(client)}, ${JSON.stringify(user)}, ${JSON.stringify(scopes)}, ${code}`);
+        // NOTE: caller (@jmondi/oauth2-server) is responsible for adding the remaining items, PKCE params, redirect URL, etc
```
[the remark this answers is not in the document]
rl-gitpod: That requires moving the entire AuthCodeRepositoryDB (OAuthAuthCodeRepository) implementation to server and since it requires storage, then adding a db service + implementation for it to use.  

I agree it meets the current, very structured implementation approach for *most* parts of the system that interact with the db and will take a look at it tomorrow.

components/server/src/oauth-server/oauth-controller.ts
```diff
@@ -0,0 +1,151 @@
...
+                    // Return the HTTP redirect response
+                    const oauthResponse = await authorizationServer.completeAuthorizationRequest(authRequest);
+                    return handleResponse(req, res, oauthResponse);
+                } catch (e) {
+                    handleError(e, res);
+                }
+            });
+
+            router.post("/oauth/token", async (req: express.Request, res: express.Response) => {
+                log.info(`TOKEN: ${JSON.stringify(req.body)}`);
```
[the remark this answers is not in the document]
rl-gitpod: Ah, I see what you meant - thanks.

As per an earlier comment I'll be doing a pass to reduce the logging and this will go away.

components/gitpod-db/src/typeorm/auth-code-repository-db.ts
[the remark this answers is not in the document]
AlexTugarev: > That requires moving the entire AuthCodeRepositoryDB (OAuthAuthCodeRepository) implementation to server

no it doesn't. 

this pattern is used in may cases here. the `DB` implementation lives in `gitpod-db` and is injected into services from `server`.

components/server/src/oauth-server/oauth-authorization-server.ts
rl-gitpod: I think you are ignoring the PKCE-side - this is why that standard was created.

Later on, it checks the [code challenge](https://github.com/jasonraimondi/typescript-oauth2-server/blob/153c83bd64940a03ba31c833d4e3b2aeb4ce03cd/src/grants/auth_code.grant.ts#L79) against the original request and will reject it.

components/server/tsconfig.json
```diff
@@ -25,5 +25,5 @@
     "include": [
         "src",
         "ee/src"
-    ]
+, "../gitpod-db/src/typeorm/db-auth-code-repository.ts"    ]
```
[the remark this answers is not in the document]
rl-gitpod: Appears that way - good catch.

components/server/src/oauth-server/oauth-authorization-server.ts
AlexTugarev: @rl-gitpod, indeed it's following the validation.

the thing is, the `authCode.codeChallenge` is looked up by `validatedPayload.auth_code_id` which – as mentioned in 2.) can be forged from a valid authorization request made in parallel, just because it's then easy to decode and recompose to a new one.

components/gitpod-db/src/typeorm/auth-code-repository-db.ts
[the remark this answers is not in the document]
rl-gitpod: > odd to implement the missing bits for the framework within gitpod-db.

Not so much a missing bit but the OAuthAuthCodeRepository interface needs to be implemented for the framework to use. Moving 1 requires moving the rest (yes - there are other, less readable ways to do so).

> this pattern is used in may cases here. the `DB` implementation lives in `gitpod-db` and is injected into services from `server`.

Agreed. That's what I was referring to with "then adding a db service + implementation for it to use"

components/server/src/oauth-server/oauth-authorization-server.ts
rl-gitpod: Sorry, I missed what you meant in 2 - long day.

if it is a publicly known secret I agree, but as it will be coming from an external source this isn't an issue.

AlexTugarev: that's why I commented with the reminder
> please don't forget about this TODO.

csweichel: I reckon we should make this come from an external source prior to merging the PR. Pushing this to `env.ts` and the `values.yaml` isn't much of an effort.

rl-gitpod: > I agree it should be taken from an external source 
That was the plan...

Done... from the logs
> 2021-05-20T04:44:40.948Z.  JWT:rdqfUiocN2Ql7fYdFKDN

It is random, but as the auth code is very short-lived it it shouldn't be a problem if it is redeployed (unless you are very unlucky).

I'll circle back to reducing the logging later, but now I've given into the virus that I've been fighting with the last couple of days.

components/dashboard/src/OauthClientApproval.tsx
```diff
@@ -0,0 +1,63 @@
...
+        const additionalData = user.additionalData = user.additionalData || {};
+        if (isApproved) {
+            additionalData.oauthClientsApproved = {
+                ...additionalData.oauthClientsApproved,
+                [clientID]: new Date().toISOString()
+            }
+        } else if (additionalData.oauthClientsApproved) {
+            delete additionalData.oauthClientsApproved[clientID];
+        }
+        await getGitpodService().server.updateLoggedInUser({
```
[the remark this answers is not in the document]
rl-gitpod: Good idea - done.

components/server/src/oauth-server/oauth-controller.ts
```diff
@@ -0,0 +1,178 @@
...
+import { Env } from "../env";
+import { clientRepository, createAuthorizationServer } from './oauth-authorization-server';
+
+@injectable()
+export class OAuthController {
+    @inject(Env) protected readonly env: Env;
+    @inject(UserDB) protected readonly userDb: UserDB;
+    @inject(AuthCodeRepositoryDB) protected readonly authCodeRepositoryDb: AuthCodeRepositoryDB;
+
+    private getValidUser = (req: express.Request, res: express.Response): User | null => {
```
csweichel, line 23:
```suggestion
    private getValidUser(req: express.Request, res: express.Response): User | null {
```

components/dashboard/src/OauthClientApproval.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { getGitpodService } from "./service/service";
+import gitpodIcon from './icons/gitpod.svg';
+import { getSafeURLRedirect } from "./provider-utils";
+
+export default function OAuthClientApproval() {
+    const { user, setUser } = useContext(UserContext);
+    const params = new URLSearchParams(window.location.search);
+    const clientID = params.get("clientID") || "";
+    const clientName = params.get("clientName") || "";
+    const redirectTo = getSafeURLRedirect(params.get("redirectTo") || undefined) || "/";
```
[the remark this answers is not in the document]
rl-gitpod: Done

csweichel (commented):
Got this when trying to authenticate:
image

rl-gitpod:
> Got this when trying to authenticate:
> image

Ah, a little chicken/egg :-(

That scope is valid but insufficient (only a single workspace) so it was changed to 'getWorkspaces' in the backend.

You need to build a local app using the commit I [linked to in our discussion](https://github.com/gitpod-io/gitpod/commit/f24dc7107edaea3373622945c3796fe6d9d3f345)

components/server/src/oauth-server/oauth-controller.ts
rl-gitpod: Sheesh, I blame the fog in my head earlier today. Clearly I was not as functional as I thought.

rl-gitpod: Fixed - thanks

components/server/src/user/user-controller.ts
```diff
@@ -227,6 +228,7 @@ export class UserController {
         });
         if (this.env.enableLocalApp) {
             router.get("/auth/local-app", async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+                log.info('AUTH');
```
csweichel, line 231: debugging leftover?

components/server/src/user/user-controller.ts
```diff
@@ -590,6 +592,7 @@ export class UserController {
 
     protected getSafeReturnToParam(req: express.Request) {
         const returnToURL: string | undefined = req.query.redirect || req.query.returnTo;
+        log.info(`getSafe: ${returnToURL}`)
```
csweichel, line 595: debugging leftover?

components/gitpod-db/src/typeorm/user-db-impl.ts
```diff
@@ -350,6 +356,92 @@ export class TypeORMUserDBImpl implements UserDB {
...
+    public async extraAccessTokenFields?(user: OAuthUser): Promise<ExtraAccessTokenFields | undefined> {
+        // No extra fields in token
+        log.info(`getUserByCredentials ${JSON.stringify(user)}`);
+        return;
+    }
+
+    // OAuthTokenRepository
+    async issueToken(client: OAuthClient, scopes: OAuthScope[], user?: OAuthUser): Promise<OAuthToken> {
+        const expiry = tokenExpiryInFuture.getEndDate();
+        log.info(`issueToken ${JSON.stringify(expiry)}`)
```
csweichel, line 384: debugging leftover?

components/gitpod-db/src/typeorm/user-db-impl.ts
```diff
@@ -350,6 +356,92 @@ export class TypeORMUserDBImpl implements UserDB {
...
+    }
+    async issueRefreshToken(accessToken: OAuthToken): Promise<OAuthToken> {
+        // NOTE(rl): this exists for the OAuth server code - Gitpod tokens are non-refreshable (atm)
+        accessToken.refreshToken = "refreshtokentoken";
+        accessToken.refreshTokenExpiresAt = new DateInterval("30d").getEndDate();
+        await this.persist(accessToken);
+        return accessToken;
+    }
+    async persist(accessToken: OAuthToken): Promise<void> {
+        log.info(`persist access token ${JSON.stringify(accessToken)}`);
```
csweichel, line 401: debugging leftover?

components/gitpod-db/src/typeorm/user-db-impl.ts
```diff
@@ -350,6 +356,92 @@ export class TypeORMUserDBImpl implements UserDB {
...
+        const tokenHash = crypto.createHash('sha256').update(accessToken.accessToken, 'utf8').digest("hex");
+        const userAndToken = await this.findUserByGitpodToken(tokenHash);
+        if (userAndToken) {
+            // Yes, update it (~)
+            // NOTE(rl): as we don't support refresh tokens yet this is not really required 
+            // since the OAuth server lib calls issueRefreshToken immediately after issueToken
+            // We do not allow changes of name, type, user or scope.
+            dbToken = userAndToken.token as GitpodToken & { user: DBUser };
+            const repo = await this.getGitpodTokenRepo();
+            log.info(`persist access token update: ${JSON.stringify(dbToken)}`);
```
csweichel, line 415: debugging leftover?

components/gitpod-db/src/typeorm/user-db-impl.ts
```diff
@@ -350,6 +356,92 @@ export class TypeORMUserDBImpl implements UserDB {
...
+            }
+            dbToken = {
+                tokenHash,
+                name: accessToken.client.id,
+                type: GitpodTokenType.MACHINE_AUTH_TOKEN,
+                user: user as DBUser,
+                scopes: scopes,
+                created: new Date().toISOString(),
+            };
+            log.info(`persist access token store: ${JSON.stringify(dbToken)}`);
```
csweichel, line 430: debugging leftover?

components/gitpod-db/src/typeorm/user-db-impl.ts
```diff
@@ -350,6 +356,92 @@ export class TypeORMUserDBImpl implements UserDB {
...
+            return this.storeGitpodToken(dbToken);
+        }
+    }
+    async revoke(accessTokenToken: OAuthToken): Promise<void> {
+        log.info(`revoke ${JSON.stringify(accessTokenToken)}`);
```
csweichel, line 435: debugging leftover?

components/gitpod-db/src/typeorm/user-db-impl.ts
```diff
@@ -350,6 +356,92 @@ export class TypeORMUserDBImpl implements UserDB {
...
+                id: user.id,
+                name: user.name,
+            };
+        }
+        log.info(`getUserByCredentials returned nothing!`)
+        return;
+    }
+    public async extraAccessTokenFields?(user: OAuthUser): Promise<ExtraAccessTokenFields | undefined> {
+        // No extra fields in token
+        log.info(`getUserByCredentials ${JSON.stringify(user)}`);
```
csweichel, line 377: debugging leftover?

components/gitpod-db/src/typeorm/user-db-impl.ts
```diff
@@ -350,6 +356,92 @@ export class TypeORMUserDBImpl implements UserDB {
     public async findUserByName(name: string): Promise<User | undefined> {
         return (await this.getUserRepo()).findOne({ name });
     }
+
+    // OAuthAuthCodeRepository
+    // OAuthUserRepository
+    public async getUserByCredentials(identifier: string, password?: string, grantType?: GrantIdentifier, client?: OAuthClient): Promise<OAuthUser | undefined> {
+        log.info(`getUserByCredentials ${identifier}`)
```
csweichel, line 363: debugging leftover?

components/gitpod-db/src/typeorm/user-db-impl.ts
```diff
@@ -350,6 +356,92 @@ export class TypeORMUserDBImpl implements UserDB {
...
+        const user = await this.findUserById(identifier);
+        if (user) {
+            log.info(`getUserByCredentials returned: ${user.id}, ${user.name}`)
```
csweichel, line 366: debugging leftover?

components/gitpod-db/src/typeorm/user-db-impl.ts
```diff
@@ -350,6 +356,92 @@ export class TypeORMUserDBImpl implements UserDB {
...
+            return {
+                id: user.id,
+                name: user.name,
+            };
+        }
+        log.info(`getUserByCredentials returned nothing!`)
```
csweichel, line 372: debugging leftover?

csweichel (commented):
There's still a bunch of log statements which I'd fret would be too loud in production. Many (most?) of them should either be on debug level, or be removed.

Other than that, after some commit history authoring this change would be ready for main.

Thank you very much for adding this! OAuth in Gitpod enables a whole host of new use-cases.

csweichel (approved):
LGTM

PR #4228, opened 2021-05-17
[werft] Use delete pod --force for stubborn workspaces

geropl:
Fixes the situation where preview environments don't get delete because `kubectl delete pod ...` hangs indefinitely on single workspaces.

### DIFF
--- a/.werft/util/kubectl.ts
+++ b/.werft/util/kubectl.ts
@@ -52,7 +52,11 @@ function deleteAllWorkspaces(namespace: string, shellOpts: ExecOptions) {
         try {
             // In most cases the calls below fails because the workspace is already gone. Ignore those cases, log others.
             exec(`kubectl patch pod --namespace ${namespace} ${o} -p '{"metadata":{"finalizers":null}}'`, { ...shellOpts });
-            exec(`kubectl delete pod --namespace ${namespace} ${o}`, { ...shellOpts });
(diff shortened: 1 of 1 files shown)


wulfthimm:
Deleting the pod by force will remove the pod from Kubernetes but not from the cluster leaving it in an unknown state. Automating the deletion could obscure the problem and break our preview cluster in the long run.

geropl:
> Deleting the pod by force will remove the pod from Kubernetes but not from the cluster leaving it in an unknown state. Automating the deletion could obscure the problem and break our preview cluster in the long run.

True, but that's a cost I'd like to pay. Especially as we're actively try to cycle nodes ideally each day - which is currently blocked by this very bug.

geropl:
```
copying certificate from "certs/staging-gpl-werft-force-delete" to "staging-gpl-werft-force-delete/proxy-config-certificates"
Error: kubectl get secret staging-gpl-werft-force-delete --namespace=certs -o yaml         | yq d - 'metadata.namespace'         | yq d - 'metadata.uid'         | yq d - 'metadata.resourceVersion'         | yq d - 'metadata.creationTimestamp'         | sed 's/staging-gpl-werft-force-delete/proxy-config-certificates/g'         | kubectl apply --namespace=staging-gpl-werft-force-delete -f - exit with non-zero status code
```
Never had _that_ before... :grimacing: Re-trying

/werft run

corneliusludmann (approved):
_May the force be with you._

Change looks good so far. Let's hope that this will improve the situation. Thank you, @geropl!

PR #4231, opened 2021-05-18
[local companion app] Add to dashboard

csweichel:
This PR makes the local companion app available from the dashboard, i.e. adds it to the `public/` assets.

It also sets the version during build by overwriting a `version.go` file instead of using ldflags. This way we can set the ldflags globally in the `WORKSPACE.yaml`.

### How to test
```bash
# Download the local companion app binary from the dashboard
curl -OL https://cw-lca-dashboard-dl.staging.gitpod-dev.com/static/bin/gitpod-local-companion-linux

# Run the app and check its version
chmod +x gitpod-local-companion-linux
./gitpod-local-companion-linux -v
```

### DIFF
--- a/WORKSPACE.yaml
+++ b/WORKSPACE.yaml
@@ -4,6 +4,7 @@ defaultArgs:
   coreYarnLockBase: ../..
   npmPublishTrigger: "false"
   publishToNPM: true
+  localAppVersion: unknown
 
 defaultVariant:
   config:
--- a/chart/templates/proxy-deployment.yaml
+++ b/chart/templates/proxy-deployment.yaml
@@ -39,6 +39,7 @@ metadata:
     component: proxy
     kind: deployment
     stage: {{ .Values.installation.stage }}
+    gitpod.io/requiresCerts: 'true'
 spec:
   selector:
     matchLabels:
@@ -46,6 +47,7 @@ spec:
       component: proxy
       kind: pod
       stage: {{ .Values.installation.stage }}
+      gitpod.io/requiresCerts: 'true'
   replicas: {{ $comp.replicas | default 1 }}
   strategy:
     type: RollingUpdate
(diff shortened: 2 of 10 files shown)


WORKSPACE.yaml
```diff
@@ -4,6 +4,7 @@ defaultArgs:
   coreYarnLockBase: ../..
   npmPublishTrigger: "false"
   publishToNPM: true
+  localAppVersion: unkown
```
rl-gitpod, line 7: /minor nit
-> unknown

chart/templates/proxy-deployment.yaml
```diff
@@ -39,13 +39,15 @@ metadata:
     component: proxy
     kind: deployment
     stage: {{ .Values.installation.stage }}
+    gitpod.io/requiresCerts: 'true'
```
rl-gitpod, line 42: Some bonus, but important additions :-)

components/local-app/BUILD.yaml
```diff
@@ -42,13 +48,26 @@ packages:
...
+        - ["cp", "_deps/components-local-app--version/version.go", "version.go"]
       config:
         packaging: app
+    - name: version
+      type: generic
+      argdeps:
+        - localAppVersion
+      config:
+        commands:
+          - ["sh", "-c", "echo package main >> version.go"]
```
rl-gitpod, line 67: fyi - the embed directive in Golang v1.16 can also be used for this e.g. https://www.buymeacoffee.com/smarterfish/dynamic-version-numbers-in-go-1-16

It has the bonus of not requiring a valid .go file and also standardises the way to embed any sort of static file

rl-gitpod (approved):

components/local-app/BUILD.yaml
csweichel: Excellent point. We're using `embed` in `docker-up` already, might as well use it here.

PR #4237, opened 2021-05-18
[github-auth-provider] Prefer pass list email over primary email

corneliusludmann:
When the deployment has a `blockNewUsersPassList`, with this change the GitHub auth provider chooses an email address that matches the passlist as Gitpod email address when such a verified email address is given in GitHub. Otherwise, it uses GitHub's default email address (as it was before).

Internal discussion: https://gitpod.slack.com/archives/C01MLB7J24T/p1621343766002600?thread_ts=1619771446.079600&cid=C01MLB7J24T

Would make https://github.com/gitpod-io/gitpod/pull/4211 obsolete.

### DIFF
--- a/components/server/src/github/github-auth-provider.ts
+++ b/components/server/src/github/github-auth-provider.ts
@@ -78,7 +78,7 @@ export class GitHubAuthProvider extends GenericAuthProvider {
             }
             return response;
         }
-        const fetchPrimaryEmail = async () => {
+        const fetchUserEmails = async () => {
             const response = await api.users.listEmails({});
             if (response.status !== 200) {
                 throw new GitHubApiError(response);
@@ -86,7 +86,7 @@ export class GitHubAuthProvider extends GenericAuthProvider {
             return response.data;
         }
         const currentUserPromise = this.retry(() => fetchCurrentUser());
-        const userEmailsPromise = this.retry(() => fetchPrimaryEmail());
+        const userEmailsPromise = this.retry(() => fetchUserEmails());
 
         try {
             const [ { data: { id, login, avatar_url, name }, headers }, userEmails ] = await Promise.all([ currentUserPromise, userEmailsPromise ]);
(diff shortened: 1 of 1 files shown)


jankeromnes:
Clever! I like this approach (for when `blockNewUsers` is used).

> Would make #4211 obsolete.

I don't think that's the case, because this Pull Request doesn't address non-Gitpod-employees wanting to try a `core-dev` deployment (e.g. starting a workspace)

components/server/src/github/github-auth-provider.ts
```diff
@@ -98,13 +98,24 @@ export class GitHubAuthProvider extends GenericAuthProvider {
                 .map((s: string) => s.trim())
             );
 
+            const filterPrimaryEmail = (emails: GitHub.UsersListEmailsResponse) => {
+                // if there is any verified email with a domain that is in the blockNewUsersPassList then use this email as primary email
+                const emailDomainInPasslist = (mail: string) => this.env.blockNewUsersPassList.some(e => mail.endsWith(`@${e}`));
+                const result = emails.filter(e => e.verified).filter(e => emailDomainInPasslist(e.email))
+                if (result.length > 0) {
+                    return result[0].email;
+                }
```
jankeromnes, line 107: Maybe move this entire block into an `if (this.env.blockNewUsers) { }` block? (To avoid confusing side-effects if you have a `blockNewUsersPassList` but `blockNewUsers` is false)

corneliusludmann: Done :+1:

AlexTugarev:
@corneliusludmann, this approach is redefining of what is actually the primary email address on GitHub, while the underlying issue is that you want to test some of the verified email addresses included in user's profile, right? If that's true, then please join a community PR to enable exactly such use cases in general, see: https://github.com/gitpod-io/gitpod/pull/4115#issuecomment-833448468
It would be best to include all verified email addresses in the OAuth process, which would make this PR easier to implement.

corneliusludmann:
@AlexTugarev I would really like to see this resolved in general with the linked PR. However, it seems that this change with multiple emails is a big thing that will not be merged soon. I think the change in this PR would solve the problem now and does not harm / has no impact on production.

The issue is now lying around again for 14 days without a solution (and interferes with me and others every day). I would still be in favor of merging this PR (regardless of the other one).

svenefftinge (approved):
LGTM! Would be better to store the emails and not treat a non-primary email as primary though.

PR #4247, opened 2021-05-19
Osf 3893 release 0.9.0

solomonope:
This is pretty much a cherry pick of the first commits from this PR #3925


JanKoehnlein (approved):
LGTM. 
Not sure whether we need @geropl's review as it's mostly deletions and docs.

chart/values.yaml
```diff
@@ -7,6 +7,10 @@ version:
 hostname: localhost
 certificatesSecret:
   secretName: https-certificates
+  certName: cert.pem
```
csweichel, line 10: This looks like a step backwards from the TLS secret type we're moving to

geropl: > This looks like a step backwards from the TLS secret type we're moving to

The goal here is to still support the certificates that come out of letsencrypt right away, by default. Otherwise we'd need to tell Self-Hosted users to do yet another step: https://www.gitpod.io/docs/self-hosted/latest/install/configure-ingress#using-lets-encrypt-to-generate-https-certificates

ghuntley:
Closing out due to age and conflicts.

PR #4267, opened 2021-05-21
Use new OAuth endpoint for login

rl-gitpod:
Remove testing code
Use correct scope for all workspaces
Clean up logging + error behaviour


components/local-app/pkg/auth/auth.go
```diff
@@ -37,21 +38,42 @@ type LoginOpts struct {
...
+	</head>
+	<body>
+		If this tab is not closed automatically, feel free to close it and proceed. <button type="button" onclick="window.open('', '_self', ''); window.close();">Close</button>
+	</body>
+</html>`
+
 // Login walks through the login flow for obtaining a Gitpod token
 func Login(ctx context.Context, opts LoginOpts) (token string, err error) {
-	rl, err := net.Listen("tcp", "localhost:0")
+	rl, err := net.Listen("tcp", "localhost:64110")
```
csweichel, line 63: I reckon the "use unused port" way is less error prone - or am I missing a reason to hardcode this?

rl-gitpod: I agree, but yes, (unfortunately) the redirect URL has to be known to the backend.

akosyakov: Could it be configurable if the port is already used?

rl-gitpod: No, not yet at least. The backend validates the full redirect uri, including port, so it would need a way to override that behaviour (which might be worth PR-ing on the lib).

In the interim I've changed it, along with the backend, to try a range of 10 ports and be more explicit about using IPV4 to workaround issues with IPV6 I saw while testing.

components/local-app/pkg/auth/auth.go
```diff
@@ -58,13 +59,28 @@ const html = `
...
+
 // Login walks through the login flow for obtaining a Gitpod token
 func Login(ctx context.Context, opts LoginOpts) (token string, err error) {
-	rl, err := net.Listen("tcp", "localhost:64110")
-	if err != nil {
-		return "", err
+	// Try a range of ports
+	var rl net.Listener
+	port := STARTING_PORT_NUM
+	for port < ENDING_PORT_NUM {
```
csweichel, line 71: A more classic `for` loop would be easier to understand here I reckon (no searching for `port++`).

csweichel (approved):
Trying on several ports is a great first step towards making this more resilient. 
I tried the login process as part of the OAuth PR. Changes LGTM.

PR #4271, opened 2021-05-21
[code] use proper icons for terminal tabs

JanKoehnlein:
Changes are in https://github.com/gitpod-io/vscode/commit/565bc735797159e8b167d4f917150c40e570ac29


akosyakov (approved):
It works for original icons, but if user picks another icon it is lost. Probably it should belong to the layout info not to active terminals.

JanKoehnlein:
It seems like they have implemented this differently on vscode/main in the meantime, so we should maybe rather abandon my solution.

akosyakov:
ok, it is not super critical and should be fixed by next rebase

PR #4274, opened 2021-05-21
[db-sync] Port from gitpod-com

csweichel:
This PR moves db-sync from gitpod-com. In a self-hosted installation db-sync is of little use, but internally it's the tool we use to synchronise databases across regions.


chart/templates/db-sync-configmap.yaml
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) 2021 Gitpod GmbH. All rights reserved.
+# Licensed under the MIT License. See License-MIT.txt in the project root for license information.
+
+{{ $comp := .Values.components.dbSync -}}
```
AlexTugarev, line 4: @csweichel this is confusing: the `dbSync` config node is not contained in `components`, see https://github.com/gitpod-io/gitpod/pull/4274/files#diff-a9d07052701ab3b718c82ac0fe74b965a8260b06bb5906362081718b0ee593b4R178
how does this work? 🤷🏻‍♂️ 
maybe that's the reason for broken deployment of kedge and payment 🤔

AlexTugarev: thanks for clarifying! 🙏🏻

PR #4276, opened 2021-05-22
[werft] Create new gcp project using terraform for preview env

princerachit:
## What?
Prelim checkin to allow the creation of terraform based preview environment gcp project and subdomain.

A gcp project would be created in the following cases:
1. Annotation "deploytarget=gke" is passed to werft context

## Next Steps
1. Once we have terraform scripts for full fledged self hosted gke installation, we will add logic to call that script for actual gitpod installation
2. Project cleanup after X hours
3. build.ts file refactoring

/werft deploytarget=gke

### DIFF
--- a/.werft/README.md
+++ b/.werft/README.md
@@ -1,0 +1,28 @@
+# Overview
+
+We use werft for our CICD pipeline. We are transforming our scripts to support various types of deployment. Here are few things you should be aware of if you decide to make changes in this directory:
+
+## Prefer Annotations over branch name based logic
+
+Using branch name to pass runtime information to werft script has several drawbacks:
+1. Branch name length has limit
+1. We endup writing complex logic around branch name parsing
+  1. It leads to bad developer experience; someone looking at the code has to rely on comments to understand cryptic parsing logic
+  1. It adds additional burden of learning about branch naming format
+
+That said, we still use branch name in special cases like releasing from a particular branch etc.
+
+Due to listed reasons we use werft annotations to pass runtime information. You can read more about annotations [here](https://github.com/csweichel/werft).
+
+## Deployments
+
+Currently all non-main branches are deployed in a preview cluster under a specific namespace. If you make changes to the source branch, the namespace is wiped and the new changes are redeployed.
+
+## Deployment target
+
+We have added a new annotation `deploytarget` to support deployment of gitpod based on the desired target. This annotation based logic is still being developed and only supports `gke` value atm.
+We will eventually deprecate the `no-preview` annotation in favour of this annotation. This readme will be updated accordingly.
+
+Future supported values for `deploytarget`:
+1. `gke`: Installs gitpod in a fresh gke cluster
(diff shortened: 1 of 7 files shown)


csweichel:
Very happy to see the ability to produce a properly isolated/closer to prod preview env in core-dev! Thank you for this effort.

I have two notes/comments:
### Branch naming conventions
For a long time we used branch-name based configuration - it's been a mess.
- branch name based config doesn't work well with Gitpod OOTB because Gitpod names branches, 
- it's not orthogonal to other build concerns because your branch names can only be so long, 
- it's yet another means of providing build/deploy time configuration.

As a replaced we introduced the build configuration that werft supports today. The only case where we still have branch-name based behaviour is the `main` build. Please, let's stick to the annotation based build config rather than re-introducing naming conventions. 

### The target environment matters, not the tech that produces it
What this PR provides is a preview environment deployed in its own GKE cluster - that's it's core functionality. The fact that this happens using terraform is not really important; that's just the tech to make it work which might even change in the future. Considering that we'll likely have a range of deployment scenarios, we could name this build config accordingly, e.g. `targetenv` or something to that tune.
    - `targetenv=gke` produces a new GKE cluster (in a GCP project using terraform) and deploys Gitpod using helm into that cluster (please don't mix the deployment in with the terraform stuff. One is infrastructure, the other is application).
    - `targetenv=namespace` is today's default behaviour where we deploy Gitpod in a namespace within a shared GKE cluster.
    - `targetenv=prod` spins up a non-GKE cluster with the stack that we ~are~ will be using in prod, e.g. k3s, Linux 5.12, containerd 1.5 + parallel pull patch, etc.
    - `targetenv=prod-split` produces a GKE cluster where we'll deploy _meta_ and a non-GKE cluster as workspace cluster. Then registers the workspace cluster using `gpctl` in the meta cluster.

Clearly some of the latter target environments are out of scope of this PR, but indicate where we'll need to be heading.

princerachit:
Thanks a lot @csweichel ! @meysholdt has done the majority of work, so credits should go to him.

Your comments make sense. I am going to do the following changes based on them:

1. Remove the branch based trigger
2. Replace tf-preview with `deploytarget` I don't particularly find `targetenv` meaningful, unless we are saying that a PR can be deployed to env like prod, staging, preprod etc.
3. Create a Readme and add the `deploytarget` that we are supporting as of now


WDYT?

csweichel:
> Thanks a lot @csweichel ! @meysholdt has done the majority of work, so credits should go to him.

Thanks @meysholdt for making this happen!

> Your comments make sense. I am going to do the following changes based on them:
> 
> 1. Remove the branch based trigger

🙏 

> 2. Replace tf-preview with `deploytarget` I don't particularly find `targetenv` meaningful, unless we are saying that a PR can be deployed to env like prod, staging, preprod etc.

That makes total sense. 

> 3. Create a Readme and add the `deploytarget` that we are supporting as of now
> 
> WDYT?

Love it. Sounds like a great way forward.

Issue #4275, opened 2021-05-22
"New Git Integration" dialog - Match field labels with the provider's naming


mikenikles:
When creating an OAuth app, GitLab and GitHub provide slightly different names for what we refer to as "Client ID" and "Client Secret" in the UI.

I suggest we display field labels depending on which provider is selected based on the following matrix:

|               | GitLab         | GitHub        |
|---------------|----------------|---------------|
| Client ID     | Application ID | Client ID     |
| Client Secret | Secret         | Client secret |

This change will make it easier to follow the instructions in the documentation as there is no need to explain the above mapping to readers.

jordanhailey:
Hey @mikenikles, following up from our chat in the Svelte Discord, hopefully this first contribution is helpful!

I couldn't figure out how to run an instance of Gitpod on my own Gitpod account. From looking over the forum I [found a relevant post](https://community.gitpod.io/t/i-would-want-bring-contribution-into-building-the-new-dashboard-but-cant-access-to-designs/2969) that seems to indicate that the team needs to initiate a build, after which in which I can get temporary access to run the application and see my changes? Let me know if that will be required for this change. 

My solution was fairly straightforward, but there is one `useEffect` function I've added that should either be pulled out or ran through a QA review.

gtsiolis:
Resolved in https://github.com/gitpod-io/gitpod/pull/4287. Thanks @jordanhailey for contributing! 🌟

PR #4285, opened 2021-05-24
Modify the "New Git Integration" experience to align with provider terminology

jordanhailey:
Modifying the `GitIntegrationModal` function per #4275

The default **Provider Type** value is still "GitLab", and when GitLab is selected as the provider, the following conditional operations are applied to the form:
- Both the field label and error message instance tied to `clientId` are shown as "**Application ID**".
- Both the field label and error message instance tied to `clientSecret` are shown as "**Secret**".

**NOTE**: When the provider is not GitLab, all strings related to `clientId` and `clientSecret` are "Client ID" and "Client Secret" accordingly.

Additionally, when creating a new integration (restricted so as not to cause unexpected side effects when editing an existing integration), if the user selects a different provider, e.g. "**GitHub**",  then the **Provider Host Name** field value changes to `"github.example.com"` and the **Redirect URL** field value changes to `"https://gitpod.io/auth/github.example.com/callback"`

### DIFF
--- a/components/dashboard/src/settings/Integrations.tsx
+++ b/components/dashboard/src/settings/Integrations.tsx
@@ -479,7 +479,12 @@ export function GitIntegrationModal(props: ({
         validate();
     }, [clientId, clientSecret])
     
-    // JH-TODO: Create new useEffect function to update the host and redirectURL values when the type value changes, restrict to running only while creating a new integration
+    useEffect(() => {
+        if (props.mode === "new") {
+            const exampleHostname = `${type.toLowerCase()}.example.com`;
+            updateHostValue(exampleHostname);
+        }
+    }, [type]);
 
     const onClose = () => props.onClose && props.onClose();
     const onUpdate = () => props.onUpdate && props.onUpdate();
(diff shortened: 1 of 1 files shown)


components/dashboard/src/settings/Integrations.tsx
```diff
@@ -655,18 +662,18 @@ export function GitIntegrationModal(props: ({
                     <div className="w-full relative">
                         <input name="redirectURL" disabled={true} readOnly={true} type="text" value={redirectURL} className="w-full pr-8" />
                         <div className="cursor-pointer" onClick={() => copyRedirectUrl()}>
-                            <img src={copy} title="Copy the Redirect URL to clippboard" className="absolute top-1/3 right-3" />
+                            <img src={copy} title="Copy the Redirect URL to clipboard" className="absolute top-1/3 right-3" />
```
gtsiolis, line 665: **praise**: Thanks for catching this typo! 😇

gtsiolis (commented):
Thanks for contributing @jordanhailey! 🧡 

Let's run this preview environment to see this in actions.

/werft run

jordanhailey:
@gtsiolis Looks like I derped the build by having too long of a branch name (It says 63 characters is the max, mine being 65 characters). Does this mean I'll need to modify the branch name and submit another draft PR?

gtsiolis:
Hey @jordanhailey, yes, could you push the same code to a branch name within the limits? For reference, this is a limitation with [Let's Encrypt](https://letsencrypt.org/) which we're using to issue new certificates.

https://github.com/gitpod-io/gitpod/blob/e2fb3db59cafca2307235c29af3f40ac404d76a0/.werft/util/certs.ts#L13

jordanhailey:
Closing this PR in favor of #4287 (has shorter branch name)

PR #4287, opened 2021-05-24
Modify the "New Git Integration" experience to align with provider terminology

jordanhailey:
Per #4275, submitting same code as #4285 on a shorter branch name to fix build error.

### DIFF
--- a/components/dashboard/src/settings/Integrations.tsx
+++ b/components/dashboard/src/settings/Integrations.tsx
@@ -479,7 +479,12 @@ export function GitIntegrationModal(props: ({
         validate();
     }, [clientId, clientSecret])
     
-    // JH-TODO: Create new useEffect function to update the host and redirectURL values when the type value changes, restrict to running only while creating a new integration
+    useEffect(() => {
+        if (props.mode === "new") {
+            const exampleHostname = `${type.toLowerCase()}.example.com`;
+            updateHostValue(exampleHostname);
+        }
+    }, [type]);
 
     const onClose = () => props.onClose && props.onClose();
     const onUpdate = () => props.onUpdate && props.onUpdate();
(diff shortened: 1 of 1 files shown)


components/dashboard/src/settings/Integrations.tsx
```diff
@@ -477,7 +477,16 @@ export function GitIntegrationModal(props: ({
...
+    }, [clientId, clientSecret, type])
+
+    useEffect(() => {
+        if (props.mode === "new") {
+            // If the host value has been modified e.g. not gitlab.example.com, assume it has been set by user and end operation
+            if (!host.includes(".example.com")) return;
+            const exampleHostname = `${type.toLowerCase()}.example.com`;
+            updateHostValue(exampleHostname);
+        }
+    }, [type]);
```
jordanhailey, line 489: This hook is for a little UX improvement to modify `host` and `redirectURL` whenever the provider changes, that being said, I understand the team is following the Minimal Valuable Change approach and perhaps this creeps a bit beyond the scope of the issue I set out to address. If that's the case I am happy to remove the hook.

jordanhailey (commented):
This may be beyond the scope of the issue, feel free to reject and I'll remove the hook.

corneliusludmann:
Hey @jordanhailey!

Thanks a lot for your contribution and sorry that it took so long to get feedback. We are still working on getting better at this. :disappointed: 

I really like your contribution and I am about to approve and merge this. Unfortunately, we need a signed [CLA](https://en.wikipedia.org/wiki/Contributor_License_Agreement) first for legal reasons and this process is not automated yet. :roll_eyes: @meysholdt will contact you for this.

Thanks for your patience!

meysholdt:
hi @jordanhailey, I have received the signed CLA. Thank you!

corneliusludmann (approved):
Thanks again for your contribution, @jordanhailey! :+1:

PR #4286, opened 2021-05-24
Removing secondary class from the Authorize Git Provider CTA button to give proper weight. Per https://github.com/gitpod-io/gitpod/issues/4197

jordanhailey:
Per https://github.com/gitpod-io/gitpod/issues/4197

Dropping the `secondary` class, from reviewing the `components/dashboard/src/index.css` file I expect this should style the button with the following rules:

```css
button {
        @apply cursor-pointer px-4 py-2 my-auto bg-green-600 dark:bg-green-700 hover:bg-green-700 dark:hover:bg-green-600 text-gray-100 dark:text-green-100 text-sm font-medium rounded-md focus:outline-none focus:ring transition ease-in-out;
    }
```

Issue #4197, opened 2021-05-10
Change git provider authorization button variant 


gtsiolis:
### Problem to solve

When opening a repository from a git provider you haven't authorized, you are prompted to authorize with that git provider. The primary button to authorize with that provider is currently not using proper weight.

### Proposal

Let's change the button to authorize with a git provider from secondary to primary.

| BEFORE | AFTER |
|-|-|
| ![image](https://user-images.githubusercontent.com/120486/117707181-e818f700-b1d6-11eb-99e0-17f2f7e3a1d4.png) | ![image](https://user-images.githubusercontent.com/120486/117707187-e94a2400-b1d6-11eb-9b8e-5d4508894e31.png) |

corneliusludmann:
Fixed by #4288. Thanks, @jordanhailey!

PR #4288, opened 2021-05-24
Removing secondary class from the main CTA button to give proper weight.

jordanhailey:
Per #4197, same code as #4286 with shorter branch name to avoid build error.


gtsiolis:
Thanks for reopening this @jordanhailey! Let's run this! 🏃 

/werft run

jordanhailey:
By toggling the css classes on the browser I was able to confirm the `secondary` class is the "muted" color, it is a grey background with darker grey text. By removing that class in the browser I could test that the default styles applied on buttons are the green background and white text as pictured in the comment at https://github.com/gitpod-io/gitpod/issues/4197#issue-884827667

NOTE: I couldn't actually verify the change was applied to the button on the "not authorized" screen due to the restrictions placed on the dev environments, not being able to add a workspace. Instead, when trying to add a workspace I ended up receiving the following message:
```
Your account has been blocked.
Please contact support if you think this is an error. See also terms of service.
```

PR #4291, opened 2021-05-25
[server] Increase timeouts for parent prebuilds

jankeromnes:
Still unsure what the best approach is here:

1. Make `WorkspaceStarter.createSpec()` fetch the prebuild workspace to check for preconditions (incremental enabled && parent prebuild) before setting the custom timeout

2. Same but move the preconditions check + custom timeout value into a helper method somewhere else

3. Add some parameter on the `Workspace` shape that indicates "is a parent prebuild that should run longer"

4. Make `PrebuildManager` call `WorkspaceStarter.startWorkspace(ctx, workspace, user, options)` with a new `options` field (e.g. `customTimeout`) that gets passed to `actuallyStartWorkspace` and then to `createSpec`

### DIFF
--- a/components/gitpod-protocol/src/protocol.ts
+++ b/components/gitpod-protocol/src/protocol.ts
@@ -800,6 +800,16 @@ export namespace WithPrebuild {
     }
 }
 
+export interface WithTimeout {
+    timeout: string;
+}
+export namespace WithTimeout {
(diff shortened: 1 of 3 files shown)


kopetsss (approved):

jankeromnes:
🎰

/werft run

PR #4292, opened 2021-05-25
[server] Don't use the full-workspace-backup feature flag for prebuilds

csweichel:
This PR makes it so that even if a user has FWB enabled, their prebuilds will not be FWB.


aledbf (approved):
LGTM

PR #4293, opened 2021-05-25
[supervisor] Better reflect incremental prebuilds in prebuild logs

jankeromnes:
If a prebuild is based on another prebuild, we currently truncate & overwrite the prebuild logs, and "forget" about how long the parent prebuild took.

Instead, we should:

- Keep the parent prebuild logs (to be more transparent about incremental prebuilds and allow debugging builds)
- Log the total saved time (parent prebuild + incremental prebuild)

That's what this PR does.

### How to test

1. Click [here](https://jx-incremental-prebuild-logs.staging.gitpod-dev.com/#prebuild/https://github.com/jankeromnes/gitpod-staging-prebuilds/commit/98b4bfb6190a3bb9fb0f513ffc2bce4009177960) to trigger a full (parent) prebuild -- the printed terminal logs should look like a normal prebuild:
    - "🤙" message
    - but no "♻️" message
2. Then click [here](https://jx-incremental-prebuild-logs.staging.gitpod-dev.com/#incremental-prebuild/https://github.com/jankeromnes/gitpod-staging-prebuilds/commit/6f2aff3ebf94cf40a0c96e87bfd00b2917fc7adb) to trigger an incremental prebuild -- now the printed terminal logs should include:
    - parent prebuild logs
    - "♻️" message
    - incremental prebuild logs
    - combined "🤙" message
3. Click [here](https://jx-incremental-prebuild-logs.staging.gitpod-dev.com/#prebuild/https://github.com/jankeromnes/gitpod-staging-prebuilds/commit/7c39c1fe4e2cb52e0dde8f8040843d3f543147b4) to trigger a **long (> 1 min)** full (parent) prebuild -- after the prebuild is done, the terminal should have:
    - "🤙"
    - "🎉" (1 min)
    - but no "♻️"
4. Then click [here](https://jx-incremental-prebuild-logs.staging.gitpod-dev.com/#incremental-prebuild/https://github.com/jankeromnes/gitpod-staging-prebuilds/commit/911f08ccd7b32d3119e66126fa55ae5e55a055c0) to trigger a **long (> 1 min)** incremental prebuild -- when it's done, terminal should have:
    - parent logs
    - "♻️"
    - new logs
    - "🤙"
    - " 🎉" (still 1 min, not 2)

### DIFF
--- a/components/supervisor/pkg/supervisor/tasks.go
+++ b/components/supervisor/pkg/supervisor/tasks.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -377,9 +378,16 @@ func (tm *tasksManager) watch(task *task, terminal *terminal.Term) {
 	go func() {
 		defer stdout.Close()
 
-		fileName := tm.prebuildLogFileName(task)
-		// TODO(janx): If the file already exists (from a parent prebuild), extract its "time saved", and log that below
-		// (instead, or in addition to, the incremental prebuild time).
+		var (
+			fileName    = tm.prebuildLogFileName(task)
+			oldFileName = fileName + "-old"
+		)
+		if _, err := os.Stat(fileName); err == nil {
+			// If the file already exists (from a parent prebuild), temporarily move it so that it doesn't get truncated.
+			// On the off chance that renaming fails here, we silently ignore that -- the new prebuild logs simply won't reflect
+			// the older logs and elapsed time (`importParentLogAndGetDuration` is always safe thanks to its initial `os.Stat`).
+			_ = os.Rename(fileName, oldFileName)
+		}
 		file, err := os.Create(fileName)
 		var fileWriter *bufio.Writer
 		if err != nil {
@@ -391,12 +399,17 @@ func (tm *tasksManager) watch(task *task, terminal *terminal.Term) {
 			fileWriter = bufio.NewWriter(file)
 			defer fileWriter.Flush()
 		}
+		// Import any parent prebuild logs and parse their total duration if available
+		parentElapsed := importParentLogAndGetDuration(oldFileName, fileWriter)
 
 		buf := make([]byte, 4096)
 		for {
 			n, err := stdout.Read(buf)
 			if err == io.EOF {
 				elapsed := time.Since(start)
+				if parentElapsed > elapsed {
+					elapsed = parentElapsed
+				}
 				duration := ""
 				if elapsed >= 1*time.Minute {
 					elapsedInMinutes := strconv.Itoa(int(elapsed.Minutes()))
@@ -423,6 +436,46 @@ func (tm *tasksManager) watch(task *task, terminal *terminal.Term) {
 			tm.reporter.write(data, task, terminal)
 		}
 	}()
+}
+
+func importParentLogAndGetDuration(fn string, out io.Writer) time.Duration {
+	if _, err := os.Stat(fn); err != nil {
+		return 0
+	}
+	defer os.Remove(fn)
+
+	file, err := os.Open(fn)
+	if err != nil {
+		return 0
+	}
+	defer file.Close()
+
+	defer out.Write([]byte("♻️ Re-running task as an incremental workspace prebuild\n\n"))
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		l := scanner.Text()
+		if strings.Contains(l, "🤙 This task ran as a workspace prebuild") {
+			break
+		}
+		out.Write([]byte(l + "\n"))
+	}
+	if !scanner.Scan() {
+		return 0
+	}
+	reg, err := regexp.Compile(`🎉 Well done on saving (\d+) minute`)
+	if err != nil {
+		return 0
+	}
+	res := reg.FindStringSubmatch(scanner.Text())
+	if res == nil {
+		return 0
+	}
+	elapsedInMinutes, err := strconv.Atoi(res[1])
+	if err != nil {
+		return 0
+	}
+	return time.Duration(elapsedInMinutes) * time.Minute
 }
 
 type composeCommandOptions struct {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 All notable changes to this project will be documented in this file.
 
+## June 2021
+
+- Better reflect incremental prebuilds in prebuilt workspace logs ([#4293](https://github.com/gitpod-io/gitpod/pull/4293))
+- Run shellcheck against scripts ([#4280](https://github.com/gitpod-io/gitpod/pull/4280))
+- Implement new Project and Team DB tables and entities ([#4368](https://github.com/gitpod-io/gitpod/pull/4368))
+- On gitpod.io 404 redirect to www.gitpod.io ([#4364](https://github.com/gitpod-io/gitpod/pull/4364))
+- Fix disk space leak in ws-manager ([#4388](https://github.com/gitpod-io/gitpod/pull/4388))
+- Fix memory leak in ws-manager ([#4384](https://github.com/gitpod-io/gitpod/pull/4384))
+- Handle GitHub issues page context URL ([#4370](https://github.com/gitpod-io/gitpod/pull/4370))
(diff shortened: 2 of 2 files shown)


components/supervisor/pkg/supervisor/tasks.go
```diff
@@ -377,9 +379,29 @@ func (tm *tasksManager) watch(task *task, terminal *terminal.Term) {
...
-		// TODO(janx): If the file already exists (from a parent prebuild), extract its "time saved", and log that below
-		// (instead, or in addition to, the incremental prebuild time).
+		var (
+			fileName                    = tm.prebuildLogFileName(task)
+			parentLog                   = ""
+			parentElapsed time.Duration = 0
+		)
+		if _, err := os.Stat(fileName); err == nil {
+			// If the file already exists (from a parent prebuild), extract its "elapsed time" before overwriting the prebuild logs.
+			content, err := ioutil.ReadFile(fileName)
```
csweichel, line 389: These log files can be arbitrarily large. Using a [`bufio.Scanner`](https://golang.org/pkg/bufio/#Scanner) to read the file line by line would be more appropriate (see [example](https://stackoverflow.com/questions/8757389/reading-a-file-line-by-line-in-go/16615559#16615559)).

components/supervisor/pkg/supervisor/tasks.go
```diff
@@ -377,9 +379,29 @@ func (tm *tasksManager) watch(task *task, terminal *terminal.Term) {
...
+			if err == nil {
```
csweichel, line 390: The block below is nested unusually deep. I'd pull the entire thing out into a function that reads the file (using a scanner) and returns the time the previous prebuild took. Something like:

```Go
func getParentBuildDuration(fn string) time.Duration {
    if _, err := os.Stat(fn); err != nil {
        return 0
    }

    file, err := os.Open(fn)
    if err != nil {
        // can't open the file - touch luck
       return 0
    }
    defer file.Close()

    scanner := bufio.NewScanner(file)
    for scanner.Scan() {
        l := scanner.Text()
        if !reg.Match(l) {
            continue
        }
        res := time.Duration(extractDurationUsingRegexp(reg, l))
       return res
    }

    return 0
}
```

components/supervisor/pkg/supervisor/tasks.go
```diff
@@ -391,12 +413,15 @@ func (tm *tasksManager) watch(task *task, terminal *terminal.Term) {
 			fileWriter = bufio.NewWriter(file)
 			defer fileWriter.Flush()
 		}
+		if parentLog != "" {
```
csweichel, line 416: if 
```suggestion
		if parentElapsed > 0 {
```

components/supervisor/pkg/supervisor/tasks.go
jankeromnes: Good point, thanks! But my goal was also to repeat the previous prebuild logs (e.g. in case incremental prebuilds cause a subtle build problem, where you need to see the old logs + the new logs).

Maybe scan & write to a new file at the same time?

csweichel: You'd have to write the log to a temp file then, as otherwise you'd be overwriting yourself :)

jankeromnes: I've elected to move a pre-existing file to `{,-old}`, then read from that while writing / looking for the end message.

components/supervisor/pkg/supervisor/tasks.go
```diff
@@ -425,6 +433,45 @@ func (tm *tasksManager) watch(task *task, terminal *terminal.Term) {
 	}()
 }
 
+func importParentLogAndGetDuration(fileName string, fileWriter *bufio.Writer) time.Duration {
```
jankeromnes, line 436:
```suggestion
func importParentLogAndGetDuration(fn string, out *io.Writer) time.Duration {
```

jankeromnes, line 436:
```suggestion
func importParentLogAndGetDuration(fn string, out *io.Writer) time.Duration {
```

components/supervisor/pkg/supervisor/tasks.go
```diff
@@ -425,6 +433,45 @@ func (tm *tasksManager) watch(task *task, terminal *terminal.Term) {
...
+	}
+	defer os.Remove(fileName)
+
+	file, err := os.Open(fileName)
+	if err != nil {
+		return 0
+	}
+	defer file.Close()
+
+	defer fileWriter.WriteString("♻️ Re-running this task as an incremental workspace prebuild\n\n")
```
jankeromnes, line 448: Handle write error?

components/supervisor/pkg/supervisor/tasks.go
```diff
@@ -425,6 +433,45 @@ func (tm *tasksManager) watch(task *task, terminal *terminal.Term) {
...
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		l := scanner.Text()
+		if strings.Contains(l, "🤙 This task ran as a workspace prebuild") {
+			break
+		}
+		fileWriter.WriteString(l + "\n")
```
jankeromnes, line 456: Handle write error?

components/supervisor/pkg/supervisor/tasks.go
```diff
@@ -425,6 +433,45 @@ func (tm *tasksManager) watch(task *task, terminal *terminal.Term) {
...
+		return 0
+	}
+	res := reg.FindStringSubmatch(scanner.Text())
+	if res == nil {
+		return 0
+	}
+	if elapsedInMinutes, err := strconv.Atoi(res[1]); err == nil {
+		return time.Duration(elapsedInMinutes) * time.Minute
+	}
+	return 0
```
jankeromnes, line 472: Swap both returns (for bail-out consistency)

components/supervisor/pkg/supervisor/tasks.go
```diff
@@ -391,12 +397,14 @@ func (tm *tasksManager) watch(task *task, terminal *terminal.Term) {
...
 		}
+		// Import any parent prebuild logs and parse their total duration if available
+		parentElapsed := importParentLogAndGetDuration(oldFileName, fileWriter)
 
 		buf := make([]byte, 4096)
 		for {
 			n, err := stdout.Read(buf)
 			if err == io.EOF {
-				elapsed := time.Since(start)
+				elapsed := time.Since(start) + parentElapsed
```
jankeromnes, line 407: Probably MAX makes more sense than SUM here.

jankeromnes (commented):
Transcribed feedback from 1:1 call.

jankeromnes:
🎰

/werft run

jankeromnes:
Can't be tested (`core-dev` deployments are currently broken) but ready for review.

jankeromnes:
🎰

/werft run

jankeromnes:
🎰

/werft run

jankeromnes:
Works as expected! 🎉

@csweichel please take another look 👀

Screenshot 2021-06-01 at 15 58 11

components/supervisor/pkg/supervisor/tasks.go
```diff
@@ -377,9 +378,14 @@ func (tm *tasksManager) watch(task *task, terminal *terminal.Term) {
...
-		fileName := tm.prebuildLogFileName(task)
-		// TODO(janx): If the file already exists (from a parent prebuild), extract its "time saved", and log that below
-		// (instead, or in addition to, the incremental prebuild time).
+		var (
+			fileName    = tm.prebuildLogFileName(task)
+			oldFileName = fileName + "-old"
+		)
+		if _, err := os.Stat(fileName); err == nil {
+			// If the file already exists (from a parent prebuild), temporarily move it so that it doesn't get truncated.
+			_ = os.Rename(fileName, oldFileName)
```
csweichel, line 387: What happens if the rename fails? Considering the error is ignored here, a comment as to why ignoring it is fine would be handy

csweichel (commented):
I know this feels a bit like coming out of the blue, but a unit test for `importParentLogAndGetDuration` would be nice. If you moved the `defer os.Remove()` out of `importParentLogAndGetDuration` you could re-write that function to use the new `io/fs` package (`testing/fstest` in the test correspondingly).

Considering the increased emphasis on testing is a recent development, I'd be ok to approve this PR as is - but tests sure would be nice :)

csweichel (approved):

PR #4297, opened 2021-05-26
fix helm chart

AlexTugarev:


geropl:
@AlexTugarev there's also https://github.com/gitpod-io/gitpod/pull/4298/files#diff-003581f24a1f58e22b9439042aed96553e6b557cd39dd059150e9f195593da45L107

PR #4300, opened 2021-05-26
[analytics] Add Track Event for Signups & Notification Changes

jakobhero:
Adds new Segment call for new signups that includes marketing information such as email, name, etc.

### DIFF
--- a/components/server/src/user/user-service.ts
+++ b/components/server/src/user/user-service.ts
@@ -18,6 +18,7 @@ import { TermsProvider } from "../terms/terms-provider";
 import { TokenService } from "./token-service";
 import { EmailAddressAlreadyTakenException, SelectAccountException } from "../auth/errors";
 import { SelectAccountPayload } from "@gitpod/gitpod-protocol/lib/auth";
+import { IAnalyticsWriter } from "@gitpod/gitpod-protocol/lib/util/analytics";
 
 export interface FindUserByIdentityStrResult {
     user: User;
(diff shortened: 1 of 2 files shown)


csweichel (commented):
Change looks good to me.
Please squash to a single commit, then I'll happily approve :)

csweichel (approved):
LGTM

components/server/src/user/user-service.ts
```diff
@@ -144,6 +146,20 @@ export class UserService {
             const canPass = newUser.identities.some(i => !!i.primaryEmail && emailDomainInPasslist(i.primaryEmail));
 
             newUser.blocked = !canPass;
+
+            //call analytics to track user signup
+            this.analytics.track({
```
AlexTugarev, line 151: @jakobhero and @csweichel, this is on the hot path of the signup process and there is no error handling around this `track` call, which we learned may throw errors. given that `handleNewUser` isn't expected to throw, such errors cases might happen under the radar. 

also, I think this analytics call should be moved out of `handeNewuser` which is there to setup a user representation. `handleTosProceedForNewUser` at https://github.com/gitpod-io/gitpod/blob/f5f99267872d8ac586967882a3ca93c388c22cb3/components/server/src/user/user-controller.ts#L495 seems to be a better home, at least there you see all of the process call.

jankeromnes: Side note: I believe a `try/catch` was recently added inside `analytics.track`:

https://github.com/gitpod-io/gitpod/blob/ea44fa8a331fa8ab062222a743773bd5ca6c7423/components/gitpod-protocol/src/util/analytics.ts#L70-L78

However, the point about `handleTosProceedForNewUser` > `handeNewuser` remains valid in any case.

AlexTugarev: @jankeromnes, thanks 💯 times for pointing this out!

PR #4307, opened 2021-05-26
Various post-deployment fixes

csweichel:
- Use same source of truth for workspace cluster in ws-manager-bridge and server
- Reduce blast radius for failing segment calls


geropl (approved):
Code LGTM

Issue #3653, opened 2021-03-29
Change page title structure for the dashboard pages


gtsiolis:
### Problem

Currently, all pages within the dashboard share the same page title that is _Gitpod - Dashboard_.

### Proposal

1. First MVC update could be to change this so that product name goes last like _Dashboard - Gitpod_.
2. Next minimum viable change could be to change page titles for each product area like the following:
   - Account - Gitpod
   - Preferences - Gitpod
   - Workspaces - Gitpod
   - ...

These two could be tackled in one step.

PR #4313, opened 2021-05-27
[dashboard] Fix page titles

jankeromnes:
Fixes https://github.com/gitpod-io/gitpod/issues/3653

Note: This PR simply sets `document.title` in relevant places (as opposed to using something like React Helmet).

### How to test

- Open every dashboard page & verify the page title is correct

### DIFF
--- a/components/dashboard/public/index.html
+++ b/components/dashboard/public/index.html
@@ -17,7 +17,7 @@
     />
     <link rel="apple-touch-icon" href="%PUBLIC_URL%/favicon192.png" />
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
-    <title>Gitpod - Dashboard</title>
+    <title>Dashboard — Gitpod</title>
   </head>
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
--- a/components/dashboard/src/components/PageWithSubMenu.tsx
+++ b/components/dashboard/src/components/PageWithSubMenu.tsx
@@ -4,6 +4,7 @@
  * See License-AGPL.txt in the project root for license information.
  */
 
+import { useEffect } from "react";
 import { useLocation } from "react-router";
 import { Link } from "react-router-dom";
 import Header from '../components/Header';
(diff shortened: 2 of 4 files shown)


AlexTugarev:
Let's use `—` instead of `-`. This would not just be good typography, but also aligned with Code:

Screen Shot 2021-05-27 at 07 45 14

jankeromnes:
> Let's use `—` instead of `-`.

I love this suggestion, many thanks!

> This would not just be good typography, but also aligned with Code:

Actually I notice that Code uses an Em Dash (`—`) surrounded by spaces. My understanding is that you should choose between:
- Using an En Dash (`–`) surrounded by spaces (e.g. `Dashboard – Gitpod`)
- Usinig an Em Dash (`—`) without any spaces around it (e.g. `Dashboard—Gitpod`)

Which one do you prefer?

AlexTugarev:
> EDIT: Went with an En Dash (–) surrounded by spaces for now.

On green field that would be my favorite. Given that we are aligning with VS Code, I'd actually vote for consistency over accuracy here.  🤷🏻‍♂️

jankeromnes:
Ok, agreed then. Go consistency! 🙌

AlexTugarev:
Seeing this while context parsing is in progress:

Screen Shot 2021-05-27 at 11 20 22

AlexTugarev (approved):
Looks much better now! ❤️ 

not sure how this fits into the picture: https://github.com/gitpod-io/gitpod/pull/4313#issuecomment-849480131

gtsiolis (approved):
Feels good to see this issue getting resolved! Thanks @jankeromnes @AlexTugarev! 🏀 

FWIW, I used to go for `mdash` for such cases but any dash should suffice! Also, noticed many have switched to a plain dash in titles recently. See also [relevant reference](https://www.grammarly.com/blog/dash/).

jankeromnes:
Many thanks for the cool reviews! 🙏

> Seeing this while context parsing is in progress:

I believe this is supervisor setting the page title to the "description" of the workspace, before the IDE changes page titles again:

https://github.com/gitpod-io/gitpod/blob/d2ece4948d19ac71732bf81ab40dc58eb79b829d/components/supervisor/frontend/src/index.ts#L36

If we want to change this, I'd be very happy to do that in a follow-up. 😁

PR #4314, opened 2021-05-27
[helm] Add server.wsmanSkipSelf

corneliusludmann:
Setting `server.wsmanSkipSelf` to `true` skips adding the ws-manager of this cluster to the list of ws-managers in server and ws-manager-bridge (and sets an empty list instead).


csweichel (approved):
Works like a charm

PR #4315, opened 2021-05-27
[proxy] fix api-server config

AlexTugarev:
how to test:

visit /admin/users first then try to hard reload in the browser. verify, you don't see an "unhandled request" error generated by the server.


JanKoehnlein (approved):
LGTM

PR #4320, opened 2021-05-27
Include rbac proxy for server and ws-manager

meysholdt:
This PR assumes https://github.com/gitpod-io/gitpod/pull/4319 is merged first.


ArthurSens:
We shouldn't include it on server yet. Server was not changed to serve metrics on localhost only

ArthurSens (approved):

PR #4322, opened 2021-05-27
Add support for Cert-Manager

wulfthimm:
This enabled the use of Certmanager to work as a CA issuer that is used to create certificates for ws-daemon, ws-manager and the docker-registry.


chart/values.yaml
```diff
@@ -631,3 +631,12 @@ rabbitmq:
         dest-uri: {{ $shovel.destUri | default "amqp://" | quote }}
         reconnect-delay: {{ $shovel.reconnectDelay | default 5 }}
     {{- end }}
+
+cert-manager:
```
geropl, line 635: in the private repo we use `certManager`; would be nice to adjust that as well to have it in-line.

geropl (approved):
:+1:

PR #4323, opened 2021-05-27
[chart] Remove deprecated prometheus annotations

aledbf:


ArthurSens (approved):
I didn't know we had so many 😮. Thanks!

PR #4324, opened 2021-05-27
Don't expose metrics when rbac proxy does that.

meysholdt:


ArthurSens (approved):
I'm not totally sure this is enough, I think we do declare metrics port on our helm values file as well. But that would be fixed on gitpod-com

PR #4327, opened 2021-05-27
[server] Enable kube-rbac-proxy

aledbf:
Change listen address in `node.js` server to localhost and enable `kube-rbac-proxy`


csweichel (approved):
LGTM

ArthurSens (approved):
Ah nice! Do we have a date when this is going to prod? I need to make the necessary adjustments on the observability repo too :)

Issue #4335, opened 2021-05-28
Workspace fails initializing


Carlos-Schrodinger:
### Bug description

Whenever i try to start a workspace i just get this message:
![image](https://user-images.githubusercontent.com/48862659/119990292-71e30580-bfc8-11eb-9eae-b6d51992b6aa.png)

### Steps to reproduce

1. Open a empty github repo
2. Open on gitpod
3. Bug

### Expected behavior

To open a new workspace and the IDE.

### Example repository

any

### Anything else?

Running chrome dev branch

Carlos-Schrodinger:
![image](https://user-images.githubusercontent.com/48862659/119990481-aa82df00-bfc8-11eb-9cc0-71d27faa919a.png)
![image](https://user-images.githubusercontent.com/48862659/119990511-b4a4dd80-bfc8-11eb-80c3-c7c26da907c2.png)

Carlos-Schrodinger:
https://share.wildbook.me/Fv0UojxLqV2Xgull.webm

A small video of what happens whenever i try running a workspace.

jankeromnes:
Many thanks for reporting this @Carlos-Schrodinger! It's due to a recent change to the workspace initializer, and only affects opening empty repositories.

We'll try to deploy a fix as soon as possible.

Carlos-Schrodinger:
> Many thanks for reporting this @Carlos-Schrodinger! It's due to a recent change to the workspace initializer, and only affects opening empty repositories.
> 
> We'll try to deploy a fix as soon as possible.

Alright, for now i'll add a README to the repo.

asthomasdk:
I am seeing the same issue on 0.9.0-alpha3 self-hosted.


Regards

Thomas Hansen
Director
____________________________
appSolutions a|s

________________________________

Carlos-Schrodinger:
Made the readme, seems to be working! Thanks ;-)
![image](https://user-images.githubusercontent.com/48862659/119994204-a953b100-bfcc-11eb-9890-4222776fee4a.png)

jankeromnes:
(We've implemented a fix, but still need to deploy it to production.)

jankeromnes:
Fix deployed and gitpod.io can open empty repositories again. 🎉

PR #4337, opened 2021-05-28
[content-init] Don't fail when origin is an empty repository

jankeromnes:
Fixes https://github.com/gitpod-io/gitpod/issues/4335

Note: This is similar to the 'git stash -u' error when repositories are empty, so the fix is mostly a copy of:

https://github.com/gitpod-io/gitpod/blob/b43fbfa22f6d3da0f1fd037ce4a686182a9c1a19/components/content-service/pkg/initializer/prebuild.go#L88-L93

### DIFF
--- a/components/content-service/pkg/initializer/git.go
+++ b/components/content-service/pkg/initializer/git.go
@@ -6,8 +6,10 @@ package initializer
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


csweichel (approved):
LGTM

PR #4342, opened 2021-05-30
Fix the security reporting URL

mikenikles:
Fixes https://github.com/gitpod-io/website/issues/512

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,4 +2,4 @@
 
 We appreciate your efforts to responsibly disclose your findings, and will make every effort to acknowledge your contributions.
 
-To report a security issue please visit https://gitpod.io/security
+To report a security issue please visit https://www.gitpod.io/security


fntlnz (approved):

Carlos-Schrodinger:
I believe this should be merged without a doubt as it does not affect any of the website structure at all.

Just an honest opinion since implementing this fix is taking 2 days even though only a word was changed

jankeromnes:
Agreed! Thanks, merging.

jankeromnes:
Well, convincing the CI first.

/werft run

jankeromnes:
Side note: We could also add a redirection from `https://gitpod.io/security` to `https://www.gitpod.io/security` just in case, for example like so:

https://github.com/gitpod-io/gitpod/blob/898b629c82bb692666b64d6cc47f306385c3b5b0/components/dashboard/conf/Caddyfile#L42

Issue #4347, opened 2021-05-31
[local-app] GITHUB_HOST has no default


rl-gitpod:
### Bug description

If you launch the local app without specifying a GITPOD_HOST env var it will fail to connect to a websocket.

### Steps to reproduce

```
$ ./gitpod-local-companion-darwin
ERRO[0001] failed to connect, trying again in 2 seconds...  error="malformed ws or wss URL" url=/api/v1
ERRO[0003] failed to connect, trying again in 3 seconds...  error="malformed ws or wss URL" url=/api/v1
ERRO[0006] failed to connect, trying again in 4 seconds...  error="malformed ws or wss URL" url=/api/v1
```

### Expected behavior

```
$ GITPOD_HOST=https://gitpod.io ./gitpod-local-companion-darwin
INFO[0001] connection was successfully established       url="wss://gitpod.io/api/v1"
```

Issue #4346, opened 2021-05-31
[local-app] Token is not saved to local keyring


rl-gitpod:
### Bug description

The local app does not save the token to the local keyring resulting in multiple round-trips & db entries.

### Steps to reproduce

Run the local app without using the mock keyring:

` GITPOD_HOST=https://gitpod.io ./gitpod-local-companion-darwin`

It will open the authentication page on gitpod.io every time it is run and nothing is written to the keyring.

### Expected behavior

Token should be written to keyring and re-used on subsequent runs

PR #4358, opened 2021-06-01
Fix issues blocking ssh from local terminal

rl-gitpod:
Deleted 'defer'ed close of terminal was main cause.

Fixed some [other](https://github.com/gitpod-io/gitpod/issues/4347) [minor](https://github.com/gitpod-io/gitpod/issues/4346) issues while there.


aledbf (approved):
LGTM

PR #4364, opened 2021-06-01
[dashboard] on 404 redirect to www

svenefftinge:
This makes `gitpod.io/whatever` redirect to `www.gitpod.io/whatever`

Try navigating to 
 - https://dashboard-404-redirect.staging.gitpod-dev.com/workspaces/
 - https://dashboard-404-redirect.staging.gitpod-dev.com/docs/
 - https://dashboard-404-redirect.staging.gitpod-dev.com/does-not-exist/

### DIFF
--- a/components/dashboard/src/App.tsx
+++ b/components/dashboard/src/App.tsx
@@ -41,6 +41,10 @@ function Loading() {
     </>;
 }
 
+function isGitpodIo() {
+    return window.location.hostname === 'gitpod.io' || window.location.hostname === 'gitpod-staging.com' || window.location.hostname.endsWith('gitpod-dev.com')
+}
+
 function App() {
     const { user, setUser } = useContext(UserContext);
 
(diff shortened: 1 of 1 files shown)


gtsiolis (approved):
Works as expected! ✔️ 

Two questions:

1. There's a small redirection delay where we print a label (`www.gitpod.io`). Expected? Would it make sense to rephrase this to `Redirecting ...` or remove it? Works fine as is, too.
2. Also, is this going to work ok when we introduce teams and users can create for example a `docs` team with a URL like `gitpod.io/docs`?

https://user-images.githubusercontent.com/120486/120515580-1a350780-c3d7-11eb-800b-853f3b47e77e.mov

svenefftinge:
> Also, is this going to work ok when we introduce teams and users can create for example a docs team with a URL like gitpod.io/docs

We need to whitelist all the slugs that we disallow for teams and that would include the slugs from the website.

Issue #4234, opened 2021-05-18
Enable "product notification" emails by default in notification settings


ChristinFrohne:
**Problem**
Currently, users need to manually navigate to `/notifications` to enable product-related emails. This probably leads most users to not enabling product emails

**Solution**
Can we enable product notifications by default?

**Context:** 
- Formerly the email settings for product (marketing) notifications were more present, as they were part of the sign-in process
- We're planning to create a drip down campaign explaining first how to configure Gitpod and second providing meaningful tips & tricks
- all users can unsubscribe/ change the email settings from each email
- legally it seems to be fine "You have implied permission to email somebody if you have an existing business relationship with them. This could mean they are a current customer, .." [[1](https://www.campaignmonitor.com/resources/guides/permission/)]

related to #4142 

cc @gtsiolis

gtsiolis:
Thanks for providing all the context above, @ChristinFrohne! 🙏 

FYI Added #4249 to resolve this. 🏓

ChristinFrohne:
**Short summary of the latest decisions:**
For now, we'll keep only `account` and `product` settings in /notification. 
Account emails include everything account related e.g. invoices, failed billing. Mostly all Chargebee emails. 
Product emails include the onboarding campaign and newsletters. 

It's likely that we extend the opt-in fields once we send more emails e.g. changelog, further campaigns. Regarding the onboarding flow, if someone unsubscribes from it, it will stop the campaign, but doesn't influence the notification settings. 

**Outstanding to-do's:**

- [ ]  product and account emails need to be active by default for all new sign-ups
- [ ] changes to the notification settings need to be picked up by Segment. Segment should forward the information to the warehouse and to customer.io 

Is this something you could take care of @jakobhero?

cc @JohannesLandgraf

jakobhero:
Notification changes are already tracked as of #4300. I just submitted a PR that changes the default preference at signup too.

jakobhero:
Woops I just saw that we already have a PR from this from @gtsiolis that was closed a couple of weeks ago here #4249 . Shouldn't we just reopen that one?

jakobhero:
#4365 was approved, closing this issue.

gtsiolis:
> Woops I just saw that we already have a PR from this from @gtsiolis that was closed a couple of weeks ago here #4249 . Shouldn't we just reopen that one?

[...]

PR #4365, opened 2021-06-01
Enable Product Comm by default for new Signups

jakobhero:
partially solves #4234.


jankeromnes (approved):
Thanks! Looks good to me. ✅

> partially solves #4234.

Why "partially"? (I.e. what is still missing?)

jakobhero:
Thanks for the swift Review!😍

> Why "partially"? (I.e. what is still missing?)

Sorry about the vagueness here. "Partially" refers to the two outstanding to dos in #4234:

> **Outstanding to-do's:**
> 
> * [ ]   product and account emails need to be active by default for all new sign-ups
> * [x]  changes to the notification settings need to be picked up by Segment. Segment should forward the information to the warehouse and to customer.io

The first is solved here. The second has been solved in #4300 already.

PR #4371, opened 2021-06-01
[proxy] Enable tls1.2 as default

aledbf:


csweichel (approved):
Makes total sense. Thank you

PR #4373, opened 2021-06-02
[dev] Enable unit tests

geropl:
This does:
- [x] enable unit tests
- [x] enable DB tests
- [x] make DB tests self-containen

#### Up to discussion
During the last meeting we thought about not just initializing but also starting the DB (as docker image) during the build. This would align "dev time" and "CI time" even more. I'm not opinionated here; so keen on what you think.

#### Test
Note that unit tests pass :heavy_check_mark: 

#### Out-of-Scope
Some tests are not run, because some personal access tokens for external services (GitHub, Gitlab, Bitbucket). Instead of making those work by adding new tokens we should try to make those self-contained.


aledbf (approved):
LGTM

PR #4375, opened 2021-06-02
[analytics] Move Signup Tracking to correct Position in Signup Flow

jakobhero:
Pointed out in #4300, The analytics track call should take place in `handleTosProceedForNewUser`, which is the only change in this PR.

### DIFF
--- a/components/server/src/user/user-controller.ts
+++ b/components/server/src/user/user-controller.ts
@@ -511,6 +511,18 @@ export class UserController {
         await this.userService.updateUserEnvVarsOnLogin(user, envVars);
         await this.userService.acceptCurrentTerms(user);
         this.analytics.identify({ anonymousId: req.sessionID || "unknown", userId: user.id, });
+        this.analytics.track({ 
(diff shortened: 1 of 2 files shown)


components/server/src/user/user-controller.ts
```diff
@@ -511,6 +511,18 @@ export class UserController {
         await this.userService.updateUserEnvVarsOnLogin(user, envVars);
         await this.userService.acceptCurrentTerms(user);
         this.analytics.identify({ anonymousId: req.sessionID || "unknown", userId: user.id, });
+        this.analytics.track({
```
AlexTugarev, line 514: please wrap both analytics calls in a `try-catch` to limit any side effects. 🙏🏻

AlexTugarev: NVM, just learned, that it's already error-handled internally. just missed that change 🤷🏻‍♂️

AlexTugarev (approved):
LGTM! Sorry for the noise!

PR #4378, opened 2021-06-02
[dashboard] Add workspace status indicator color red for failed status

### LINKED ISSUE #3635: Errored workspaces don't use the correct status indicator

corneliusludmann:
Fixes #3635

### DIFF
--- a/components/dashboard/src/workspaces/WorkspaceEntry.tsx
+++ b/components/dashboard/src/workspaces/WorkspaceEntry.tsx
@@ -4,7 +4,7 @@
  * See License-AGPL.txt in the project root for license information.
  */
 
-import { CommitContext, Workspace, WorkspaceInfo, WorkspaceInstance, WorkspaceInstancePhase } from '@gitpod/gitpod-protocol';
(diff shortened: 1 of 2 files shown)


gtsiolis (commented):
Hey @corneliusludmann, do you know if there's an easy way to reproduce this erroneous state on the workspaces list?

corneliusludmann:
> Hey @corneliusludmann, do you know if there's an easy way to reproduce this erroneous state on the workspaces list?

Unfortunately, not. :disappointed: 

What you can do is to alter the database to add a failed condition (as JSON content, unfortunately, I don't have a query handy). However, it seems to me that the value of this is pretty limited since you don't really test how a failed workspace behaves.

gtsiolis:
> However, it ...

DEAL. Shall we merge this as is?

gtsiolis (approved):

corneliusludmann:
I think it should be safe.

PR #4385, opened 2021-06-03
[cli] replace pstree in credential helper

### LINKED ISSUE #4062: Gitpod Extension warning that I have to grant necessary permissions
### Bug description

When I git-push to a GitHub repository from a Gitpod workspace I get the following warning popping up in the bottom right corner, although the push operation completes successfully.

image

In fact, my Gitpod settings are already all set properly.

image

The weird thing is that this happens only for certain repositories.

### Steps to reproduce

Fork https://github.com/icub-tech-iit/documentation and proceed with a push from the workspace.

See for example the video below:

https://user-images.githubusercontent.com/3738070/116051872-5d35e980-a679-11eb-8587-6e7037b16533.mp4

### Expected behavior

The pop-up message should not appear.

### Example repository

The repository where the pop-up message appears:
- https://github.com/icub-tech-iit/documentation.

A repository where everything runs smoothly:
- https://github.com/vvv-school/tutorial_cartesian-interface.

### Anything else?

_No response_

JanKoehnlein:
Fixes #4062

### DIFF
--- a/components/gitpod-cli/cmd/credential-helper.go
+++ b/components/gitpod-cli/cmd/credential-helper.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"regexp"
@@ -55,7 +56,7 @@ var credentialHelper = &cobra.Command{
 			fmt.Printf("username=%s\npassword=%s\n", user, token)
 		}()
 
-		repoURL, gitCommand := parsePstree()
+		repoURL, gitCommand := parseProcessTree()
 		host := parseHostFromStdin()
 		if len(host) == 0 {
 			log.Println("'host' is missing")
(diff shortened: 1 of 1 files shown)


AlexTugarev (approved):
Quickly checked with the repo from GH issue and it works as advertised!

> The remote repository "https://github.com/icub-tech-iit/documentation.git" is not accessible with the current token. Please grant the necessary permissions.

corneliusludmann:
@JanKoehnlein Is it still work in process? Asking due to the `[wip]` prefix. It's a left-over, right?

JanKoehnlein:
Changed the title to reflect that this is ready for review

components/gitpod-cli/cmd/credential-helper.go
```diff
@@ -145,39 +146,70 @@ func parseHostFromStdin() string {
 	return host
 }
 
-func logDebug(v ...interface{}) {
-	if os.Getenv("CREDENTIAL_HELPER_DEBUG_LOG") == "true" {
-		log.Println(v...)
+func parseProcessTree() (repoUrl string, gitCommand string) {
```
corneliusludmann, line 149: Any chance we could write some unit tests for this? Not just for the actual test but that would also help to understand the function because we would have a list of some example inputs and expected outputs.

Sure, the problem here is to mock the `readProc` function for the test, right? Any ideas?

corneliusludmann (approved):
Code looks good so far. Haven't tested since Alex has already.

The proposed unit test would be great, though. But since it wasn't there before either, it doesn't get any worse with this change. Therefore no blocker.

JanKoehnlein:
The crucial part is to interpret/parse what comes out of `/proc` in a very special scenario, which is hard to reproduce in a unit test. I don't see that there is more to consider beyond the happy path that is covered with manual testing, so I think we're fine.

PR #4387, opened 2021-06-03
[analytics] Include internal user creation Date on signup call

jakobhero:
When tracking the user signup, we want to send the exact datetime when the user was created in order to be consistent with the current practice


jankeromnes (approved):
Looks good to me, thanks! 👍

PR #4388, opened 2021-06-03
[ws-daemon] Remove workspace daemon directory in DisposeWorkspace

aledbf:


components/ws-daemon/pkg/content/service.go
```diff
@@ -399,6 +399,12 @@ func (s *WorkspaceService) DisposeWorkspace(ctx context.Context, req *api.Dispos
 		return nil, status.Error(codes.Internal, "cannot delete workspace from store")
 	}
 
+	// remove workspace daemon directory in the node
+	err = os.RemoveAll(sess.ServiceLocDaemon)
+	if err != nil {
+		log.WithError(err).WithField("workspaceId", req.Id).Error("unexpected error deleting workspace directory in the node")
```
csweichel, line 405: nit
```suggestion
		log.WithError(err).WithField("workspaceId", req.Id).Error("cannot delete workspace daemon directory")
```

csweichel (approved):
LGTM (sans nit)

components/ws-daemon/pkg/content/service.go
aledbf: fixed

PR #4389, opened 2021-06-03
[dev] reverted images for werft

JanKoehnlein:


JanKoehnlein:
See https://gitpod.slack.com/archives/C01KGM9AW4W/p1622732913357400 (internal discussion)

JanKoehnlein:
Seems like you cannot have different images for werft and dev, as it makes the `gitpod-config_test.go` fail.

aledbf (approved):
LGTM

PR #4395, opened 2021-06-04
Update Probot 

AlexTugarev:
This PR updates Probot and some other dependencies which seem to be quite outdated. 

What's left is, to verify, that the configured GitHub App is interacting as expected by Gitpod. To do so, the current branch deployments is deployed with a custom values file behind the scene. 

### Update on verification

- [x] The actual eventing from GH seems to work properly. I see log messages like this:
```
New installation recorded
Registered workspace prebuild: eac1be9f-b778-4301-b9da-5bf5c84fc941 for https://github.com/AlexTugarev/dactyl-cc.git:5137e7e781445205a7da12fc7ea9a09a0523db04
```

- [x] There appears to be an error happening down the road. And it doesn't looks like any limitation which is just happening on core-dev:
```
github.repos.createStatus is not a function\n    at PrebuildStatusMaintainer.<anonymous> (/app/node_modules/@gitpod/server/dist/ee/src/prebuilds/prebuilt-status-maintainer.js
```

`createStatus` was renamed to `createCommitStatus` which was hidden by a wrong type cast. It's fixed now. 

I could start a prebuild with a dummy GH App and a test repo. Server logs look fine now
```
Looking for prebuilt workspace
Found prebuilt workspace for https://github.com/AlexTugarev/clickdummy.git:d7b022204dc72fa5dabd90aaefd9bce8d9a6e27b
```

The workspace started successfully. 
```
# cat /workspace/.gitpod/ready/
{"source":"from-prebuild"}
```


components/ee/db-sync/src/export.ts
```diff
@@ -83,7 +82,7 @@ export class TableUpdate {
                     .pipe(new Transform({
                         objectMode: true,
                         transform: (data: any, encoding, cb) => {
-                            if(data[this.table.timeColumn] > now()) {
+                            if(data[this.table.timeColumn] > Date.now()) {
```
svenefftinge, line 85: Is `timeColumn` not a mysql date string (e.g. `1970-01-01 00:00:01.001`)?

components/server/ee/src/prebuilds/github-app.ts
```diff
@@ -33,46 +33,54 @@ import { Options, ApplicationFunctionOptions } from 'probot/lib/types';
...
 
-    protected async buildApp(options: ApplicationFunctionOptions) {
-        const app = options.app;
+    protected async buildApp(app: Probot, options: ApplicationFunctionOptions) {
         this.statusMaintainer.start(async id => (await app.auth(parseInt(id))) as any as Octokit);
         // this.queueMaintainer.start();
 
         // Backward-compatibility: Redirect old badge URLs (e.g. "/api/apps/github/pbs/github.com/gitpod-io/gitpod/5431d5735c32ab7d5d840a4d1a7d7c688d1f0ce9.svg")
-        options.getRouter('/pbs').get('/*', async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+        options.getRouter && options.getRouter('/pbs').get('/*', async (req: express.Request, res: express.Response, next: express.NextFunction) => {
```
svenefftinge, line 74: When is `getRouter` undefined and why would we want to silently not register the routes in such a case?

components/ee/db-sync/src/export.ts
AlexTugarev: interesting question. I didn't pay much attention to the context, just made sure that it's same value kind:

```
> moment.now()
1622818434314
```

For example L.69 also assumes, that dates column type is a unix timestamp. 
Maybe that's, what `query`'s result returns?

components/server/ee/src/prebuilds/github-app.ts
AlexTugarev: This is a misleading declaration. `getRouter` is always defined. I'll add a comment on this.
```
    async load(appFn) {
        await appFn(this.probotApp, {
            getRouter: (path) => this.router(path),
        });
    }
```

svenefftinge (approved):
Code LGTM.

AlexTugarev:
going to merge after removing https://github.com/gitpod-io/gitpod/pull/4395/commits/3f0915172f6b992bd238cbc19f2fd4bb1275b0ab (it belongs to another change)

Issue #4223, opened 2021-05-17
Show a tooltip with cycle information when hovering over remaining hours


gtsiolis:
### Problem to solve

In the plans page, the remaining hours are shown but it's not clear when the cycle resets.

### Proposal

Show a tooltip with cycle information when hovering over remaining hours. One option could be to show the start and end date of the cycle, like _May 17, 2021 - June 17, 2021_.

| BEFORE | AFTER |
|-|-|
| ![image](https://user-images.githubusercontent.com/120486/118472055-19328380-b711-11eb-949b-0d907bc2b5da.png) | ![image](https://user-images.githubusercontent.com/120486/118472060-1a63b080-b711-11eb-8593-62432a4bffc8.png) |

PR #4398, opened 2021-06-05
[dashboard] Show a tooltip with billing cycle dates when hovering over remaining hours

jankeromnes:
Fixes https://github.com/gitpod-io/gitpod/issues/4223

### DIFF
--- a/components/dashboard/src/settings/Plans.tsx
+++ b/components/dashboard/src/settings/Plans.tsx
@@ -6,7 +6,7 @@
 
 import React, { useState, useEffect, useContext } from "react";
 import { countries } from 'countries-list';
-import { AccountStatement, Subscription, UserPaidSubscription, AssignedTeamSubscription } from "@gitpod/gitpod-protocol/lib/accounting-protocol";
+import { AccountStatement, Subscription, UserPaidSubscription, AssignedTeamSubscription, CreditDescription } from "@gitpod/gitpod-protocol/lib/accounting-protocol";
 import { PlanCoupon, GithubUpgradeURL } from "@gitpod/gitpod-protocol/lib/payment-protocol";
 import { Plans, Plan, Currency, PlanType } from "@gitpod/gitpod-protocol/lib/plans";
 import { ChargebeeClient } from "../chargebee/chargebee-client";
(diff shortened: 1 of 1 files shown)


components/dashboard/src/settings/Plans.tsx
```diff
@@ -433,7 +433,7 @@ export default function () {
                 {!assignedTs && (
                     <p className="text-base w-96 m-auto">Upgrade your plan to get access to private repositories or more parallel workspaces.</p>
                 )}
-                <p className="mt-2 font-semibold text-gray-500">Remaining hours: {typeof(accountStatement?.remainingHours) === 'number'
+                <p className="mt-2 font-semibold text-gray-500" title={guessCurrentBillingCycle(currentPlan, accountStatement).map(d => d.toLocaleDateString()).join(' - ')}>Remaining hours: {typeof(accountStatement?.remainingHours) === 'number'
```
gtsiolis, line 436: **thought**: It took me a while to realize this was added as a title attribute! 😇

gtsiolis (approved):
Whoa! Thanks for picking this up @jankeromnes! 🌟 

The title attribute seems towork for now. Ideally this would look like a tooltip element as described in https://github.com/gitpod-io/gitpod/issues/4223 but feel free to leave this out of the scope of this iteration. 🛹 

Feel free to forward this to someone to review [the frontend here](https://github.com/gitpod-io/gitpod/pull/4398/files#diff-b3ab13da33d2cd0941e6ceabdaf994cbb02edee39618b5cb8eb096e1e7fd9818R604-R617) if needed. 🏓

components/dashboard/src/settings/Plans.tsx
jankeromnes: Titles are fantastic! I always hover over UI elements when I need more info, and pretty often I'm not disappointed. 😁

If we already have a tooltip component, I think we could do some JS-hacking to make it work for every element that has a `title` attribute.

gtsiolis: > Titles are fantastic!

Agree, and `title` along with `aria-label` attributes can significantly improve product accessibility. 👀 

> If we already have a tooltip component ...

We do have first [tooltip component](https://github.com/gitpod-io/gitpod/blob/4850f7c06fe45e6398b1fc53d0119d08839c8fc6/components/dashboard/src/components/Tooltip.tsx) implementation! 💬 

> I think we could do some JS-hacking ...

I'd not expect to see a tooltip element every time we're using the title attribute as this could make the user interface unnecessarily verbose, however, we can try this and see if we need to tone this down. 🏀

PR #4404, opened 2021-06-07
disable custom timeout errors in supervisor workspace web socket proxy

akosyakov:
#### What it does

- fix 3782: disable custom timeout errors in supervisor workspace web socket proxy

#### How to test

You can try to reproduce below first in production.

- Start a workspace.
- Open the network tab in devtools, select web sockets.
- Enable 3G network profile on system level, on Mac you can use https://nshipster.com/network-link-conditioner/
- Reload the page, you should see that 2 web sockets to workspace is eventually connected. Ignore web sockets to gitpod.io and localhost.
- Do it several times.
- As a bonus try on Edge connection, you should never get a dialog with `TIMEOUT` error and after many attempts 2 web socket connection will be established.


svenefftinge (approved):
I couldn't reproduce the timeout error with this anymore

PR #4405, opened 2021-06-07
[ws-manager] Backup on pod eviction

csweichel:
This PR stores the hostIP as annotation s.t. in case of eviction we know which ws-daemon to talk to. When pods get evicted they loose their association with the node they ran on, which until now broke backups.


aledbf (approved):
LGTM

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file.
 
+## July 2021
+
+- Improve backup stability when pods get evicted ([#4405](https://github.com/gitpod-io/gitpod/pull/4405))
```
jankeromnes, line 7: Nit: I think we're still in the June cycle. 😇

jankeromnes:
Many thanks for the quick fix!

If possible, it would be cool to have quick "how to test" instructions, in order to confirm that backups still work in both cases:
- normal timeout
- pod eviction

JanKoehnlein (approved):
LGTM

JanKoehnlein:
Test instructions:
- consume all of disk space via
```
dd if=/dev/zero of=/workspace/hd-fillup.zeros bs=1G
```
- watch your ws being evicted
- look at `workspace-manager` log files to see the message on successful backup. 

See also https://github.com/gitpod-io/gitpod/issues/4415

components/ws-manager/pkg/manager/monitor.go
```diff
@@ -326,14 +326,14 @@ func (m *Monitor) actOnPodEvent(ctx context.Context, status *api.WorkspaceStatus
...
-				log.WithError(err).Warn("was unable to remove traceID annotation from workspace")
-			}
+		// once a regular workspace is up and running, we'll remove the traceID information so that the parent span
+		// ends once the workspace has started.
+		//
+		// Also, in case the pod gets evicted we would not know the hostIP that pod ran on anymore.
+		// In preparation for those cases, we'll add it as an annotation.
+		err := m.manager.markWorkspace(ctx, workspaceID, deleteMark(wsk8s.TraceIDAnnotation), addMark(hostIPAnnotation, wso.HostIP()))
+		if err != nil {
+			log.WithError(err).Warn("was unable to remove traceID annotation from workspace")
```
corneliusludmann, line 336: NIT: With this change, the log message is missleading.

```suggestion
			log.WithError(err).Warn("was unable to remove traceID and/or add host IP annotation from/to workspace")
```

corneliusludmann (approved):
One small NIT (misleading log message).

Everything else looks fine! :+1:

PR #4406, opened 2021-06-07
Enable COOP in Caddy

fntlnz:
Signed-off-by: Lorenzo Fontana <lo@linux.com>

### DIFF
--- a/components/proxy/conf/Caddyfile
+++ b/components/proxy/conf/Caddyfile
@@ -33,7 +33,7 @@
 		# Enable cross-site filter (XSS) and tell browser to block detected attacks
 		X-XSS-Protection           "1; mode=block"
 		# Disallow sharing the same browsing context when opened in a popup
-		Cross-Origin-Opener-Policy same-origin
+		Cross-Origin-Opener-Policy same-origin-allow-popups
 
 		defer # delay changes
 	}
--- a/components/proxy/conf/Caddyfile
+++ b/components/proxy/conf/Caddyfile
@@ -32,6 +32,8 @@
 		Referrer-Policy            no-referrer-when-downgrade
 		# Enable cross-site filter (XSS) and tell browser to block detected attacks
 		X-XSS-Protection           "1; mode=block"
+		# Disallow sharing the same browsing context when opened in a popup
+		Cross-Origin-Opener-Policy: same-origin
 
 		defer # delay changes
 	}
--- a/components/proxy/conf/Caddyfile
+++ b/components/proxy/conf/Caddyfile
@@ -32,11 +32,18 @@
 		Referrer-Policy            no-referrer-when-downgrade
 		# Enable cross-site filter (XSS) and tell browser to block detected attacks
 		X-XSS-Protection           "1; mode=block"
-		# Disallow sharing the same browsing context when opened in a popup
-		Cross-Origin-Opener-Policy same-origin-allow-popups
 
 		defer # delay changes
 	}
+}
+
+# workspace security headers
+(workspace_security_headers) {
+	header {
+		# Disallow sharing the same browsing context when opened in a popup
+		Cross-Origin-Opener-Policy same-origin-allow-popups
+	}
+	import security_headers
 }
 
 (enable_log) {
@@ -192,7 +199,7 @@ https://{$GITPOD_DOMAIN} {
 # workspaces
 https://*.*.{$GITPOD_DOMAIN} {
 	import enable_log
-	import security_headers
+	import workspace_security_headers
 	import remove_server_header
 	import ssl_configuration
 	import debug_headers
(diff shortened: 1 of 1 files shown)


components/proxy/conf/Caddyfile
```diff
@@ -32,6 +32,8 @@
 		Referrer-Policy            no-referrer-when-downgrade
 		# Enable cross-site filter (XSS) and tell browser to block detected attacks
 		X-XSS-Protection           "1; mode=block"
+		# Disallow sharing the same browsing context when opened in a popup
+		Cross-Origin-Opener-Policy: same-origin
```
aledbf, line 36: @fntlnz please remove the `:` character

aledbf:
@fntlnz it seems this change breaks the redirect after login 
I see `If this tab is not closed automatically, feel free to close it and proceed. Close`  tab instead of the redirect.

fntlnz:
Thanks for the fix @aledbf - the header is probably too strict. I'll debug a bit more

fntlnz:
@AlexTugarev do you happen to have any idea on why this is breaking the redirect after login ? I think we could use your precious help here ☺️

AlexTugarev:
@fntlnz, I'm happy to help. 

When you select "Login with X" it opens a pop-up for an authorization URL of our server, but we do not send a document, just a redirect to another host. This is per [documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Opener-Policy) a breaking change:

> If a cross-origin document with COOP is opened in a new window, the opening document will not have a reference to it, and the window.opener property of the new window will be null. 

I wonder if we should send a document with a HTML redirect included and this would solve the issue. It's not hard to test, just replacing the `res.redirect(url)` calls with `res.send(ourCustomHTMLRedirectTo(url))`.

Another try would be to use `same-origin-allow-popups`, no? Or would it draw the whole thing meaningless?

AlexTugarev:
Quickly tested the HTML redirect thing, and it doesn't change anything (though it might be a better approach in general)

I think we should go with `same-origin-allow-popups`, this allows our app to use `window.open` properly, but disallow other websites to use `window.open` agains us.

akosyakov:
> I think we should go with same-origin-allow-popups, this allows our app to use window.open properly, but disallow other websites to use window.open agains us.

Can affect opening gitpod.io link from workspace origin?

AlexTugarev:
@akosyakov, COOP with `same-origin` policy will definitely affect `window.open`. the opener will be null. that's the issue to solve ATM

fntlnz:
As @AlexTugarev  saysi COOP is targeting window.open and similar techniques.

I just pushed the change Alex suggested of using `same-origin-allow-popups`. It should  still protect the document when opened as a popup window, but allow the communication with own popups as we do now.

fntlnz:
/werft run

Need the dev environment deployed again

fntlnz:
Seems to be working now, I verified in the preview environment. Anything else we should test?

AlexTugarev:
> Anything else we should test?

`security_headers` is used for `workspace` locations as well. we need to check, that it doesn't break anything unexpectedly.

AlexTugarev:
@fntlnz it actually breaks `window.open` when initiating the login with GH. `access to opener` changes to `no` on redirects apparently.

fntlnz:
oh yes you're right @AlexTugarev - it breaks when you aren't already logged into github for me.

fntlnz:
strange, I tried again and it worked

fntlnz:
No it keeps working for me, tried with different browsers/accounts

fntlnz:
My fault, it will work if developer tools are open. Otherwise it can't do the redirect.

AlexTugarev:
@fntlnz, that's interesting. Could you share the `access to opener` state for the opened window when you click on `Continue with GH` no the login page. Also, make sure you are not logged in, as it would not do any redirects to GH.

Screen Shot 2021-06-17 at 10 18 11

fntlnz:
This is what I get
image

fntlnz:
This is what I get after logging in in this PRs environment (the popup does not close)
image

fntlnz:
Had a conversation with @AlexTugarev - we decided that for enabling this on the dashboard it requires a rewrite in the way we communicate during the login. 

For that reason, we are rolling this change only to the workspaces for now.

fntlnz:
Pushed the change, please review again @AlexTugarev and @aledbf :)

aledbf:
@fntlnz I see the header in the workspace and no issues to login. That said, you also need to add the header in `ws-proxy`.
(there is no proxy in pluggable workspace clusters)

aledbf (approved):
LGTM

fntlnz:
Thanks for the hint @aledbf - didn't know that detail

AlexTugarev (approved):
LGTM!

fntlnz:
Thanks @aledbf @akosyakov @AlexTugarev for the reviews and the help in getting this done 🙏

PR #4412, opened 2021-06-07
wip: agent-smith

fntlnz:
More info in the README itself

State of the project

- [x] Loading of the Falco BPF probe using `cilium/ebpf`
- [x] Support for execve
- [ ] Parsing arguments for the syscalls we support
- [ ] Support for clone, vfork, fork, connect
- [ ] Glue together the data extracted from the syscalls perf loop with the Infringements
- [ ] Hook to Kubernetes and Gitpod WSS for blocking users/workspaces/pods


csweichel:
Once builds pass I'd like to merge this to `main` so that we can start iterating in parallel and in smaller PRs.

csweichel (approved):
Let's iterate on main

Issue #3594, opened 2021-03-25
[dashboard] Replace context URLs and labels


gtsiolis:
### Problem to solve

The design specs in #3426 for the new Workspaces list contained some structure and labels for the Context URL. For the time being, we're surfacing the context URl which overflows on smaller viewports and includes some duplicate information with the _Pending Changes_ column on the right.

### Proposal

More particularly, on the second column, a) the context URL should not include the branch name, b) the context URL should surface the commit message, and c) the subtitle should be more semantic like _Pull Request 324_,.

apolopena:
Would the branch information then go into the name column? I was not seeing the branch information anywhere but the context column in the V2 [design](https://www.figma.com/file/0saJp50UikQJMOB36txeZi/Workspaces-%233426?node-id=0%3A1)
The icons are helpful in the V1 design, maybe they can stay in V2 somehow?
![image](https://user-images.githubusercontent.com/3060702/112668905-430bb000-8e1c-11eb-81ad-d3e5827d3ccd.png)

svenefftinge:
I'm not sure I understand what's being discussed here, but #3627 might improve the situation by making the context url an actual link. With that you can go to the context or copy the url.

apolopena:
@svenefftinge 
Yes #3627 does help.  I was discussing placement of the data in regard to the context URL and not loosing any visual cues in the redesign that would inform the user as to what data they are actually looking at.

**Regarding the redesign**
The new design does away with the icons and has less space to display the data but being compact is good thing so:

**For Clarity**
- Could the full urls in a cell be displayed in a tool tip when rolling over a link? 

  - _Without the icons and with the urls being truncatated, it can be unclear what the urls are for._

**For Consistency and functionality**
- Cells in the Name column have two parts (workspace URL, repo URL), each are links. This is good UI.
- Context cells have two parts as well (branch URL and I am not sure what the lower link is for), however the top part (branch URL) is not a link.
  - _Could both parts in a Context 'cell' be a link?_

gtsiolis:
What do you think of this feature request, @jldec?

PR #4416, opened 2021-06-07
[dashboard] Fix workspace entry truncation

corneliusludmann:
The actual change is the commit 9f41b1d that fixes the truncation in the workspace list:

|   Before   |        This PR       | 
|:---------:|:-----------------------:|
| image | image   |

Besides this, this PR also has the commit 4476666 that adds tooltips for the values. The rationale behind this is that it should be possible to read the whole non-truncated value somehow:

image

The downside is that the tooltip is always there, no matter if truncated or not. I don't know an easy way to show the tooltip for truncated values only. If we don't like this behavior, we can drop this commit.

### DIFF
--- a/components/dashboard/src/workspaces/WorkspaceEntry.tsx
+++ b/components/dashboard/src/workspaces/WorkspaceEntry.tsx
@@ -86,19 +86,17 @@ export function WorkspaceEntry({ desc, model, isAdmin, stopWorkspace }: Props) {
                 <WorkspaceStatusIndicator instance={desc?.latestInstance} />
             </div>
             <div className="flex flex-col w-3/12">
-                <a href={startUrl.toString()}><div className="font-medium text-gray-800 dark:text-gray-100 truncate hover:text-blue-600 dark:hover:text-blue-400">{ws.id}</div></a>
-                <a href={project ? 'https://' + project : undefined}><div className="text-sm overflow-ellipsis truncate text-gray-400 hover:text-blue-600 dark:hover:text-blue-400">{project || 'Unknown'}</div></a>
+                <a href={startUrl.toString()}><div className="font-medium text-gray-800 dark:text-gray-100 overflow-ellipsis truncate w-full hover:text-blue-600 dark:hover:text-blue-400">{ws.id}</div></a>
+                <a href={project ? 'https://' + project : undefined}><div className="text-sm overflow-ellipsis truncate w-full text-gray-400 hover:text-blue-600 dark:hover:text-blue-400">{project || 'Unknown'}</div></a>
             </div>
-            <div className="flex w-4/12 truncate overflow-ellipsis">
-                <div className="flex flex-col">
-                    <div className="text-gray-500 overflow-ellipsis truncate">{ws.description}</div>
-                    <a href={ws.contextURL}>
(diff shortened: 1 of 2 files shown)


components/dashboard/src/workspaces/WorkspaceEntry.tsx
```diff
@@ -86,19 +86,31 @@ export function WorkspaceEntry({ desc, model, isAdmin, stopWorkspace }: Props) {
...
+                    <a href={project ? 'https://' + project : undefined}>
+                        <div className="text-sm overflow-ellipsis truncate w-full text-gray-400 hover:text-blue-600 dark:hover:text-blue-400">{project || 'Unknown'}</div>
+                    </a>
+                </Tooltip>
             </div>
-            <div className="flex w-4/12 truncate overflow-ellipsis">
-                <div className="flex flex-col">
-                    <div className="text-gray-500 overflow-ellipsis truncate">{ws.description}</div>
+            <div className="flex flex-col w-4/12">
+                <Tooltip content={ws.description}>
```
gtsiolis, line 101: **issue**: Adding the tooltip on all element can a) make the interface feel slightly overwelmed, b) feel redandunt on some elements like the workspace name, and introduce some visual imbalance (see screenshots).

| Context | Changes | 
|-|-|
| Screenshot 2021-06-10 at 2 22 19 AM | Screenshot 2021-06-10 at 2 22 10 AM |

components/dashboard/src/workspaces/WorkspaceEntry.tsx
```diff
@@ -86,19 +86,31 @@ export function WorkspaceEntry({ desc, model, isAdmin, stopWorkspace }: Props) {
...
-                        <div className="text-sm text-gray-400 overflow-ellipsis truncate hover:text-blue-600 dark:hover:text-blue-400">{ws.contextURL}</div>
+                        <div className="text-sm text-gray-400 overflow-ellipsis truncate w-full hover:text-blue-600 dark:hover:text-blue-400">{ws.contextURL}</div>
                     </a>
-                </div>
+                </Tooltip>
             </div>
             <div className="flex flex-col items-start w-2/12">
-                <div className="text-gray-500 truncate">{currentBranch}</div>
+                <Tooltip content={currentBranch} className="w-full">
+                    <div className="text-gray-500 overflow-ellipsis truncate w-full">{currentBranch}</div>
```
gtsiolis, line 112: **suggestion**: May I suggest to leave the truncation on and remove most, if not all, tooltips? 🎈 

Ideally, we could make the _Context_ column so that it contains more compact information (see https://github.com/gitpod-io/gitpod/issues/3594). This way, we would only need a tooltip on the first row of the context column.

Additionally, on the _Pending Changes_ column we could truncate the branch name and offer a copy-to-clipboard button without also having a tooltip.

gtsiolis (commented):
Thanks for truncating these @corneliusludmann! 🍐

components/dashboard/src/workspaces/WorkspaceEntry.tsx
corneliusludmann: Yes, absolutely. Since the truncation is fixed in #4454 already, I'll close this one.

gtsiolis: Thanks @corneliusludmann! 🏓

PR #4417, opened 2021-06-07
[dashboard] Remove margin-bottom from PageWithSubMenu

### LINKED ISSUE #3859: Horizontal scrollbar on pages that shouldn't have one
### Bug description

When you go to https://gitpod.io/plans or https://gitpod.io/integrations you can see a horizontal scrollbar that probably is there by mistake.

image
image

### Steps to reproduce

1. Sign in to your GitPod account
2. Go either https://gitpod.io/plans or https://gitpod.io/integrations
3. See the scrollbar

### Expected behavior

There shouldn't be a horizontal scrollbar in these places.

### Example repository

_Not applicable_

corneliusludmann:
Avoids unnecessary scrollbars. Fixes #3859.

Before this change, pages in the settings section do have a margin at the bottom. That leads to unnecessary scrollbars:

image

Without the margin, the scrollbars disappear:

image

Is there any good reason for the margin?

(Examples are with a canvas size of 1440x1049px.)

### DIFF
--- a/components/dashboard/src/components/PageWithSubMenu.tsx
+++ b/components/dashboard/src/components/PageWithSubMenu.tsx
@@ -42,7 +42,7 @@ export function PageWithSubMenu(p: PageWithSubMenuProps) {
                     })}
                 </ul>
             </div>
-            <div className='ml-32 w-full pt-1 mb-40'>
+            <div className='ml-32 w-full pt-1'>
                 {p.children}
             </div>
         </div>


jankeromnes:
> Is there any good reason for the margin?

FYI I believe this was to prevent awkward scrolling behavior when there are drop-downs / context menus near the bottom of the page (where the page height abruptly changes depending on collapsed state). I don't remember where that happened though.

corneliusludmann:
> FYI I believe this was to prevent awkward scrolling behavior when there are drop-downs / context menus near the bottom of the page (where the page height / scroll position abruptly changes depending on collapsed state). I don't remember where that happened though.

OK, thanks for the comment. So, I think we should merge this change and when the original concern occurs again, we should find a proper solution (and in doubt document the decision next to the code). What do y'all think?

JanKoehnlein (approved):
LGTM

PR #4419, opened 2021-06-08
[caddy] proxy requests to port locations through ws proxy

### LINKED ISSUE #4047: Websocket connections no longer work
### Bug description

A Vaadin application in development uses a websocket connection to provide live reload capabilities. The live reload server listens to port 35729, which is indicated in the lower right corner in the Gitpod UI: `Ports: 33411, 35729, 3799, 8080, 45049`.

Port 35729 should be mapped to e.g.
```
$ gp url 35729
https://35729-jade-turkey-nd11myav.ws-eu03.gitpod.io
```

However, establishing a connection to `wss://35729-jade-turkey-nd11myav.ws-eu03.gitpod.io/` fails without any other information in the browser console than `WebSocket connection to 'wss://35729-jade-turkey-nd11myav.ws-eu03.gitpod.io/' failed`.

An additional note is that a websocket connection is also established to `wss://8080-jade-turkey-nd11myav.ws-eu03.gitpod.io/vaadinServlet/?v-r=push&refresh_connection` and this one works properly

### Steps to reproduce

1. Open http://gitpod.io/#https://github.com/vaadin/skeleton-starter-flow-spring
2. Wait for the app to be shown (a text field and a button) 

### Expected behavior

The dev mode indicator in the lower right corner has a full green circle, not a half red half green.

The browser console has no errors

### Example repository

https://github.com/vaadin/skeleton-starter-flow-spring

### Anything else?
This has worked before

akosyakov:
#### What it does

- fix #4047: It is a work-around, ws-proxy does low level manipulations to web socket headers to align them with most common usages.

#### How to test

- Start a workspace for https://github.com/vaadin/skeleton-starter-flow-spring
- When the server is ready, try to create a web socket for port 35729 and connect to it from devtools. You can verify it in network tab.
   - Notice that live reloading won't work since they don't respect our preview env, only gitpod.io

### DIFF
--- a/components/proxy/conf/Caddyfile
+++ b/components/proxy/conf/Caddyfile
@@ -216,7 +216,11 @@ https://*.*.{$GITPOD_DOMAIN} {
 
 	@workspace_port header_regexp host Host ^(webview-|browser-|extensions-)?(?P<workspacePort>[0-9]{2,5})-(?P<workspaceID>[a-z0-9][0-9a-z\-]+).ws(?P<location>-[a-z0-9]+)?.{$GITPOD_DOMAIN}
 	handle @workspace_port {
-		reverse_proxy ws-{re.host.workspaceID}-ports.{$KUBE_NAMESPACE}.{$KUBE_DOMAIN}:{re.host.workspacePort} {
+		reverse_proxy https://ws-proxy.{$KUBE_NAMESPACE}.{$KUBE_DOMAIN}:9090 {
+			transport http {
+				tls_insecure_skip_verify
+			}
+
 			import upstream_headers
 
 			header_up X-Gitpod-WorkspaceId {re.host.workspaceID}
--- a/components/ws-proxy/pkg/proxy/routes.go
+++ b/components/ws-proxy/pkg/proxy/routes.go
@@ -258,6 +258,14 @@ func installWorkspacePortRoutes(r *mux.Router, config *RouteHandlerConfig) error
 	// forward request to workspace port
 	r.NewRoute().HandlerFunc(
 		func(rw http.ResponseWriter, r *http.Request) {
+			// a work-around for servers which does not respect case-insensitive headers, see https://github.com/gitpod-io/gitpod/issues/4047#issuecomment-856566526
+			for _, name := range []string{"Key", "Extensions", "Accept", "Protocol", "Version"} {
+				values := r.Header["Sec-Websocket-"+name]
+				if len(values) != 0 {
+					r.Header.Del("Sec-Websocket-" + name)
+					r.Header["Sec-WebSocket-"+name] = values
+				}
+			}
 			r.Header.Add("X-Forwarded-Proto", "https")
 			r.Header.Add("X-Forwarded-Host", r.Host+":443")
 			proxyPass(
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -47,3 +47,4 @@ vscode:
     - ms-kubernetes-tools.vscode-kubernetes-tools
     - stkb.rewrap
     - zxh404.vscode-proto3
+    - matthewpi.caddyfile-support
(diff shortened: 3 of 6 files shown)


components/proxy/conf/Caddyfile
```diff
@@ -216,7 +216,11 @@ https://*.*.{$GITPOD_DOMAIN} {
 
 	@workspace_port header_regexp host Host ^(webview-|browser-|extensions-)?(?P<workspacePort>[0-9]{2,5})-(?P<workspaceID>[a-z0-9][0-9a-z\-]+).ws(?P<location>-[a-z0-9]+)?.{$GITPOD_DOMAIN}
 	handle @workspace_port {
-		reverse_proxy ws-{re.host.workspaceID}-ports.{$KUBE_NAMESPACE}.{$KUBE_DOMAIN}:{re.host.workspacePort} {
```
aledbf, line 219: why are you removing the mapping for workspace ports?

akosyakov: ws-proxy does it after adjusting headers: https://github.com/gitpod-io/gitpod/pull/4419/files#diff-39f2e37e657f269523bb53adff2f549f05b1af3db45432b89daa1c2068847ef3R269

akosyakov: As far as i understood without this change it does not go through ws-proxy at least in prev env?

aledbf: Correct.

Just in case, I copied this configuration from what we had in NGINX previously.

akosyakov: I think in nginx we also delegated to ws-proxy: https://github.com/gitpod-io/gitpod/blob/a0a5017b588f307418b1daf160d8ccee1afbd4b5/chart/config/proxy/vhost.server.conf#L160-L192 but moved it to caddy while switching to nginx on purpose. I am not sure.

components/ws-proxy/pkg/proxy/routes.go
```diff
@@ -258,6 +258,12 @@ func installWorkspacePortRoutes(r *mux.Router, config *RouteHandlerConfig) error
 	// forward request to workspace port
 	r.NewRoute().HandlerFunc(
 		func(rw http.ResponseWriter, r *http.Request) {
+			for name, value := range r.Header {
```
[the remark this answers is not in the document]
akosyakov: I guess it would be enough to look only for listed here: https://datatracker.ietf.org/doc/html/rfc6455#section-11.3

aledbf:
@akosyakov this change LGTM but can you split the change in two different PRs?
One for the Caddyfile, to route everything through `ws-proxy` and other change headers in `ws-proxy`?

akosyakov:
Do you mean commits?

aledbf:
> Do you mean commits?

No, two PRs. The caddy change has no relation with the one in ws-proxy. Just to not assume in the future that are related in any way.

svenefftinge:
Aren't those changes related, i.e. would the fix work without the caddy change?

aledbf:
> would the fix work without the caddy change?

Yes.

The change to Caddy is for core-dev to behave similarly to production (all traffic through ws-proxy)

csweichel:
Those two changes are separate indeed. One is to bring core-dev closer to ws-proxy, the other is fixing ws-proxy itself.
While we indeed have two separate changes robed into one commit - which might create confusion in the future - I'd hope that this PR will act as a reminder that those changes should not have been conflated in the first place.

All that said, I'd propose we merge those changes because they make a whole lot of sense :)

svenefftinge:
Thanks for the clarification, Alejandro. So I guess my confusion is good proof of your concern. That said, I agree we should merge it as is to be able to get it in front of users ASAP.

akosyakov:
I split into 2 commits with detailed explanation from this PR.

aledbf (approved):
LGTM

PR #4428, opened 2021-06-08
Initialize workspaces with additional file content

svenefftinge:
To allow starting workspaces based on a config that doesn't sit in Git, this PR introduces the possibility to pass file contents when creating a workspace, which overlays/overwrites the file system.

This is going to be used for passing configuration stored in projects. A prefix context `additionalcontent` (used below) is introduced for testing purposes, only.

## Testing

A link that puts a `.gitpod.yml` and `.gitpod.Dockerfile` on top: https://se-external-config.staging.gitpod-dev.com/#/additionalcontent/eyIuZ2l0cG9kLnltbCI6ImltYWdlOlxuICBmaWxlOiAuZ2l0cG9kLkRvY2tlcmZpbGVcblxudGFza3M6XG4gIC0gY29tbWFuZDogZWNobyBcInRlc3RcIj5teWZpbGUudHh0IiwiLmdpdHBvZC5Eb2NrZXJmaWxlIjoiRlJPTSBnaXRwb2Qvd29ya3NwYWNlLWZ1bGxcblxuUlVOIC9ob21lL2dpdHBvZC8uY2FyZ28vYmluL2NhcmdvIGluc3RhbGwgbWRib29rIC0tdmVycyAwLjEuN1xuIn0=/https://github.com/gitpod-io/template-sveltejs

Same additional content but different repo, which triggers an image build on its own because the repo is different: https://se-external-config.staging.gitpod-dev.com/#/additionalcontent/eyIuZ2l0cG9kLnltbCI6ImltYWdlOlxuICBmaWxlOiAuZ2l0cG9kLkRvY2tlcmZpbGVcblxudGFza3M6XG4gIC0gY29tbWFuZDogZWNobyBcInRlc3RcIj5teWZpbGUudHh0IiwiLmdpdHBvZC5Eb2NrZXJmaWxlIjoiRlJPTSBnaXRwb2Qvd29ya3NwYWNlLWZ1bGxcblxuUlVOIC9ob21lL2dpdHBvZC8uY2FyZ28vYmluL2NhcmdvIGluc3RhbGwgbWRib29rIC0tdmVycyAwLjEuN1xuIn0=/https://github.com/gitpod-io/spring-petclinic


components/server/src/workspace/config-provider.ts
```diff
@@ -43,53 +43,64 @@ export class ConfigProvider {
...
-                /* There is no configuration for this repository. Before we fall back to the default config,
-                * let's see if there is language specific configuration we might want to apply.
-                */
-                customConfig = await inferredConfig;
-                if (customConfig) {
-                    customConfig._origin = 'derived';
+            }
+            if (!customConfig) {
+
+                // try and find config file in the context repo or remote in
```
svenefftinge, line 61: for the reviewers: section below hasn't changed besides the indent because of the if.

components/server/src/workspace/image-source-provider.ts
```diff
@@ -22,7 +24,13 @@ export class ImageSourceProvider {
             let result: WorkspaceImageSource;
 
             const imgcfg = config.image;
+            log.info(JSON.stringify(config))
+            log.info(JSON.stringify(context))
+            log.info('additionalcontext' + AdditionalContentContext.is(context))
+            log.info('configimage' + ImageConfigFile.is(config.image))
+            log.info('contents' + (AdditionalContentContext.is(context) && ImageConfigFile.is(config.image) && context.additionalFiles[config.image.file]))
```
svenefftinge, line 31: I'll remove the log lines above before merge ☝️

components/server/src/workspace/image-source-provider.ts
```diff
@@ -37,17 +45,26 @@ export class ImageSourceProvider {
...
+                        dockerFilePath: config.image.file,
+                        dockerFileContentSHA: this.getContentSHA(context.additionalFiles[config.image.file]),
+                    }
+                }
                 const hostContext = this.hostContextProvider.get(context.repository.host);
                 if (!hostContext || !hostContext.services) {
                     throw new Error(`Cannot fetch workspace image source for host: ${context.repository.host}`);
                 }
-                const lastDockerFileSha = await hostContext.services.fileProvider.getLastChangeRevision(context.repository, context.revision, user, imgcfg.file);
+                const lastDockerFileShaPromise = hostContext.services.fileProvider.getLastChangeRevision(context.repository, context.revision, user, imgcfg.file);
```
svenefftinge, line 58: we need to make two requests for the time being, to support the current resolution of image refs. Once we have rebuilt most images, we can remove the `lastChangeRevision` part.

components/server/src/workspace/workspace-starter.ts
```diff
@@ -738,13 +755,28 @@ export class WorkspaceStarter {
...
-        return result;
+            const tokenExpirationTime = new Date();
+            tokenExpirationTime.setMinutes(tokenExpirationTime.getMinutes() + 30);
+            const res = await this.otsServer.serve(traceCtx, JSON.stringify(context.additionalFiles), tokenExpirationTime);
+            additionalInit.setFilesUrl(res.token);
+            let location = workspace.config.checkoutLocation
+            if (!location && CommitContext.is(context)) {
+                location = context.repository.name;
+            }
+            additionalInit.setLocation(location || '');
```
svenefftinge, line 775: Left over. will be replaced with `this.getCheckoutLocation(workspace)`

svenefftinge:
> How are we dealing with environment variables and tasks? Right now the env vars assume that the workspace their pattern matches for can be trusted. With this change if I can make someone click on a link I can quite easily get their secrets because I can bring my own tasks.

I think this is a general problem that exists with all our workspace start URLs. The underlying issue is that supervisor runs processes and it is not obvious what those processes will be from looking at a link. I suggest we introduce something where we ask users if they trust certain sources to allow task execution.

csweichel:
> > How are we dealing with environment variables and tasks? Right now the env vars assume that the workspace their pattern matches for can be trusted. With this change if I can make someone click on a link I can quite easily get their secrets because I can bring my own tasks.
> 
> I think this is a general problem that exists with all our workspace start URLs. The underlying issue is that supervisor runs processes and it is not obvious what those processes will be from looking at a link. I suggest we introduce something where we ask users if they trust certain sources to allow task execution.

Indeed it does. What the `additionalContent` context URL does though is to explicitly circumvent the "security model" of environment variables/secrets (namely repository patterns). The assumption underpinning environment variables is that you trust the repository you specify using the repository pattern. There's a chance you trust this repo because you trust everyone you can modify its content, hence its tasks.
If we now allow arbitrary tasks to be executed in arbitrary repositories for arbitrary users (all you need is to click a link), that undermines this - already somewhat weak - idea of "trust in a repo".

Asking users if they want to execute the task might be one way - and is indeed a trade-off between friction and security.
Alternatively we could re-consider the _Workspace Secrets RFC_ which would alleviate this problem also.

components/content-service-api/initializer.proto
```diff
@@ -12,6 +12,7 @@ message WorkspaceInitializer {
         SnapshotInitializer snapshot = 3;
         PrebuildInitializer prebuild = 4;
     }
+    AdditionalContentInitializer additional_content = 5;
```
csweichel, line 15: I'd still much prefer introducing this as an orthogonal concept, something like:

```
message WorkspaceInitializer {
    oneof spec {
        EmptyInitializer empty = 1;
        GitInitializer git = 2;
        SnapshotInitializer snapshot = 3;
        PrebuildInitializer prebuild = 4;
        CompositeInitializer composite = 5;
        FileDownloadInitializer download = 6;
    }
}

// CompositeInitializer uses a collection of initializer to produce workspace content.
// All initializer are executed in the order they're provided.
message CompositeInitializer {
    repeated WorkspaceInitializer initializer = 1;
}

// FileDownloadInitializer downloads files and uses them as workspace content.
message FileDownloadInitializer {
    message FileInfo {
       string url = 1;
       // target_location is relative to the workspace root, e.g. if the workspace is initialized in `/workspace`
       // a target_location of `foobar/file` would produce a file in `/workspace/foobar/file`.
       // target_location must include the filename. The FileDownloadInitializer will create any parent directories
       // necessary to place the file.
       string target_location = 2;
       // digest is a hash of the file content in the OCI digest format (see https://github.com/opencontainers/image-spec/blob/master/descriptor.md#digests).
       // This information is used to compute subsequent
       // content versions, and to validate the file content was downloaded correctly.
       string digest = 3;
    }
    repeated FileInfo files = 1;
}
```

components/content-service-api/initializer.proto
```diff
@@ -120,3 +121,15 @@ message GitStatus {
     // the total number of unpushed changes
     int64 total_unpushed_commits = 8;
 }
+
+// An additional content initializer allows to put files on top of anything that has been initialized before.
+// This is applied on top of any other initializers
+message AdditionalContentInitializer {
+    // files_url is a URL where one can download the file contents encoded as a JSON object where the relative pathes are the properties and contents are its values
```
csweichel, line 128: After reading through the code I understand the intention of making this a "JSON" field. I don't think this is the right way going forward because the JSON format is reasonably undefined. Until now the API is completely well defined through protobuf. Going this route does away with that and defers validation and structure. We could instead define the file list in protobuf (see comment above).

Re OTS integration: what I had in mind is that we serve the individual file content using OTS and provide them to a `FileDownloadInitializer` as outlined above - rather than to provide a single file that gets downloaded.

components/image-builder-api/imgbuilder.proto
```diff
@@ -34,11 +34,13 @@ message BuildSourceReference {
     string ref = 1;
 }
 
+
 message BuildSourceDockerfile {
     contentservice.WorkspaceInitializer source = 1;
     string dockerfile_version = 2;
     string dockerfile_path = 3;
     string context_path = 4;
+    string dockerfile_content_sha = 5;
```
csweichel, line 43: We could re-use the `dockerfile_version` - I don't think we need to change the protocol here.
The only difference is that before the version was the commit hash, now it's the SHA hash.

If we moved the "legacy" behaviour to server, that is.

components/image-builder/pkg/builder/docker.go
```diff
@@ -224,6 +224,10 @@ func (b *DockerBuilder) getBaseImageRef(ctx context.Context, bs *api.BuildSource
 		return b.getAbsoluteImageRef(ctx, src.Ref.Ref, allowedAuth)
 
 	case *api.BuildSource_File:
+		if src.File.DockerfileContentSha != "" {
+			return fmt.Sprintf("%s:%x", b.Config.BaseImageRepository, src.File.DockerfileContentSha), nil
```
csweichel, line 228: The manifest approach below takes more into account than just the Dockerfile version. Using just the content hash ignores the context entirely - the same Dockerfile from two separate repositories with different context would get the same "build".

This could even leak sensitive details because one user might get the build of another (same Dockerfile content, but different context including repo).

csweichel: Note: because GitHub doesn't let me put a comment in line 238 ... you'd have to expand the `if` to include the `CompositeInitializer` etc.

components/content-service-api/initializer.proto
csweichel: (note: integrating the digest mechanism is straight forward using https://github.com/opencontainers/go-digest)

components/server/src/workspace/image-source-provider.ts
csweichel: If we resolved the base image here already we could check which case (`lastDockerFileSha` or `this.getContentSHA(dockerFileContent)`) resolves to a built image. This behaviour is what I referred to as "move the legacy behaviour to server" above.

components/server/src/workspace/workspace-starter.ts
```diff
@@ -345,33 +346,48 @@ export class WorkspaceStarter {
...
-                source.setGit(git);
+                let source: WorkspaceInitializer;
+                const disp = new DisposableCollection();
+                let checkoutLocation = this.getCheckoutLocation(workspace);
+                if (!imgsrc.dockerFileSource) {
+                    const {initializer, disposable} = await this.createInitializer({span}, workspace, workspace.context, user);
+                    source = initializer;
+                    disp.push(disposable);
+                } else {
+                    //TODO we cannot change this initializer structure now because it is part of how baserefs are computed in image-builder.
```
csweichel, line 357: nit: 
```suggestion
                    // TODO(se): we cannot change this initializer structure now because it is part of how baserefs are computed in image-builder.
```

see https://www.notion.so/gitpod/How-we-develop-e8ce7e562478458a9223eb9b797415b2#86e669f8e7df4f1289c9bede0855b5f9

components/server/src/workspace/workspace-starter.ts
```diff
@@ -345,33 +346,48 @@ export class WorkspaceStarter {
...
+                    git.setCheckoutLocation(checkoutLocation);
+                    source = new WorkspaceInitializer();
+                    source.setGit(git);
+                }
 
                 // Dockerfile and context path are both relative to the checkout location
-                let contextPath = (workspace.config.image as ImageConfigFile).context || ".";
-                let dockerFilePath = imgsrc.dockerFilePath;
+                let contextPath = checkoutLocation;
+                let dockerFilePath = checkoutLocation + '/' + imgsrc.dockerFilePath;
```
csweichel, line 375: This seems to be at odds with the comment above. Either the comment or the code is wrong :)

components/server/src/workspace/workspace-starter.ts
csweichel: I'd propose we update the behaviour in image builder to accompany the new initializer. Also, the current change on how the baseRef is computed ignores the initializer entirely.

svenefftinge:
I have updated the PR as discussed.

csweichel:
I added unit tests and implemented the digest check in the content initializer. Unfortunately the digest sent by the server does not match what we actually download:

image

components/image-builder/pkg/builder/docker.go
```diff
@@ -215,17 +215,36 @@ func (b *DockerBuilder) getAbsoluteImageRef(ctx context.Context, ref string, all
...
+	}
+
+	if ci := init.GetDownload(); ci != nil {
+		manifest[prefix+"Source"] = "download"
+
+		hash := sha256.New()
+		for _, f := range ci.Files {
+			_, _ = hash.Write([]byte(f.Digest))
 		}
+		manifest[prefix+"Files"] = fmt.Sprintf("%x", hash.Sum(nil))
```
svenefftinge, line 243: Wouldn't this trigger docker builds on every change in `.gitpod.Dockerfile` (or other files)? I believe this would be too aggressive.

csweichel (approved):
LGTM

components/server/src/workspace/config-provider.ts
jankeromnes: Tip: Adding `?w=1` at the end of the review URL ignores all whitespace changes (i.e. re-indented blocks will appear unchanged).

PR #4430, opened 2021-06-08
Cleanup tracing spans

aledbf:
Changes:
- remove unnecessaries changes of types
- remove binary content fields
- send error messages instead of error

### DIFF
--- a/components/registry-facade/pkg/registry/manifest.go
+++ b/components/registry-facade/pkg/registry/manifest.go
@@ -219,7 +219,6 @@ func (mh *manifestHandler) getManifest(w http.ResponseWriter, r *http.Request) {
 		}
 
 		dgst := digest.FromBytes(p).String()
-		span.LogKV("manifest", string(p))
 
 		w.Header().Set("Content-Type", desc.MediaType)
 		w.Header().Set("Content-Length", fmt.Sprint(len(p)))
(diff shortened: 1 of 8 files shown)


components/ws-daemon/pkg/content/initializer.go
```diff
@@ -98,12 +98,10 @@ func collectRemoteContent(ctx context.Context, rs storage.DirectAccess, ps stora
 
 // RunInitializer runs a content initializer in a user, PID and mount namespace to isolate it from ws-daemon
 func RunInitializer(ctx context.Context, destination string, initializer *csapi.WorkspaceInitializer, remoteContent map[string]storage.DownloadInfo, opts RunInitializerOpts) (err error) {
-	//nolint:ineffassign
+	//nolint:ineffassign,staticcheck
 	span, ctx := opentracing.StartSpanFromContext(ctx, "RunInitializer")
 	defer tracing.FinishSpan(span, &err)
 
-	span.LogKV("remoteContent", remoteContent)
```
csweichel, line 105: Why remove this?

components/registry-facade/pkg/registry/manifest.go
```diff
@@ -219,7 +219,6 @@ func (mh *manifestHandler) getManifest(w http.ResponseWriter, r *http.Request) {
 		}
 
 		dgst := digest.FromBytes(p).String()
-		span.LogKV("manifest", string(p))
```
csweichel, line 222: Why remove this? 
it's been rather handy at times

aledbf: > it's been rather handy at times

Do we need that in the tracing? Can we replace that with a debug log?

csweichel (approved):
As discussed offline, the span log removal is an attempt to reign in the tracing memory leak

PR #4433, opened 2021-06-08
[common-log] Enable change of log level at runtime

aledbf:
Allow the change of log level at runtime using the pprof port with a new endpoint.

*Using port-forward:*

**Get current level**
`curl localhost:6060/debug/logging`

**Change level**
`curl localhost:6060/debug/logging -d '{"level":"info"}'`

default value is `trace` (current value)

Also configures containerd log (logrus) with gitpod-io log settings

### DIFF
--- a/components/common-go/pprof/pprof.go
+++ b/components/common-go/pprof/pprof.go
@@ -38,6 +38,9 @@ func Handler() *http.ServeMux {
 	mux.HandleFunc(Path+"profile", pprof.Profile)
 	mux.HandleFunc(Path+"symbol", pprof.Symbol)
 	mux.HandleFunc(Path+"trace", pprof.Trace)
+
+	mux.HandleFunc("/debug/logging", log.LevelHandler)
+
 	return mux
 }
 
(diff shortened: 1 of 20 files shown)


components/common-go/pprof/pprof.go
```diff
@@ -33,55 +29,13 @@ func Serve(addr string) {
...
-				//nolint:gosec
-				id := rand.Uint32()
-				log.WithField("id", id).WithField("frac", frac).WithField("seconds", seconds).Debug("enabled mutex profiling")
 
-				runtime.SetMutexProfileFraction(int(frac))
-				defer func() {
-					runtime.SetMutexProfileFraction(0)
-					log.WithField("id", id).WithField("frac", frac).WithField("seconds", seconds).Debug("disabled mutex profiling")
-				}()
-			}
```
csweichel, line 69: I'm afraid that by removing this we loose the ability to set `frac` at runtime.
https://github.com/golang/go/commit/0ee4b13830572ad66cdcc045d1e4ae1e15a87ebb only adds `seconds`

aledbf: reverted

csweichel (approved):
LGTM

PR #4434, opened 2021-06-08
[dev] Add make-user-admin script

corneliusludmann:
Run `leeway run components:make-user-admin`

@gtsiolis: We discussed this and since I needed this now I implemented it. Are you available for a review (not time critical at all)?

### DIFF
--- a/components/BUILD.yaml
+++ b/components/BUILD.yaml
@@ -143,6 +143,7 @@ scripts:
       - test/**/*
     script: |
       leeway exec --filter-type go -v -- go mod tidy -v
+
   - name: unblock-user
     deps:
       - components/service-waiter:app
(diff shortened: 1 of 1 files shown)


components/BUILD.yaml
```diff
@@ -174,3 +175,35 @@ scripts:
...
+      fi
+      query="select name, rolesOrPermissions from d_b_user limit 10;"
+      result=$(mysql -e "$query" -u$DB_USERNAME -p$DB_PASSWORD -h 127.0.0.1 gitpod)
+
+      printf "\nCurrent context:           $(kubectx -c)"
+      printf "\nCurrent namespace:         $(kubens -c)\n"
+
+      printf "\nAvailable users (max 10):\n$result\n\n"
+      printf "Enter user to unblock (empty to abort): "
+      read user
```
csweichel, line 201: Leeway scripts also support arguments, i.e. instead of reading this from stdin you could have folks pass it in when calling the script.

```suggestion
      user=${user}
```

would result in
```
# fails
$ leeway run components:make-user-admin
FATA[0001] cannot run script with unresolved argument "${user}": use -Duser=value to set the argument

# works
$ leeway run components:make-user-admin -Duser=foobar
```

corneliusludmann: Indeed, that would be a nice alternative. Thanks for the hint. However, that's an additional thing that everyone that would like to use this command has to remember (the `-Duser=foobar` syntax). That's why I decided to use the current implementation. But I am not dogmatic about it.

gtsiolis (commented):
Thanks for adding this @corneliusludmann! I do run into the following error in the browser console when trying to access the Admin area after making a user an admin. Is this expected?

```js
Error: enterprise license required
```

gtsiolis:
> Yes, I think that's expected. (See also this internal discussion.)

Aha! OK, shall merge this as is for now?

gtsiolis (approved):

PR #4439, opened 2021-06-08
Headless Logs III: Upload and retrieve headless logs after the workspaces has stopped

geropl:
#### Run prebuild and download log
 - Add env var: https://gpl-headless-log-content.staging.gitpod-dev.com/variables , with `PREBUILD_PARAMS=100_1s_10s_success` for `gitpod-io/gitpod`
 - Run prebuild: https://gpl-headless-log-content.staging.gitpod-dev.com/#prebuild/https://github.com/gitpod-io/gitpod/tree/gpl/test-prebuild
 - Find out the instanceId from the prebuild (for instance `select * from d_b_prebuilt_workspace`)
 - Open Browser Console on your dashboard tabs and paste `await window._gp.gitpodService.server.getHeadlessLog("|instanceId|")`
 - Download the log file from the presented URL

#### ToDo
 - [x] upload logs
 - [x] provide access via content service
 - [x] serve from `/headless-logs/` (depends on https://github.com/gitpod-io/gitpod/pull/4262 being merged)


components/content-service/pkg/logs/logs.go
```diff
@@ -0,0 +1,91 @@
...
+	prebuildLogFilePrefix = "prebuild-log-"
+
+	legacyTerminalStoreLocation = "/workspace"
+	legacyPrebuildLogFilePrefix = ".prebuild-log-"
+
+	// UploadedHeadlessLogPathPrefix is the prefix under which headless logs are stored inside an instance
+	UploadedHeadlessLogPathPrefix = "logs"
+)
+
+// UploadedHeadlessLogPath returns the path relative to the
```
csweichel, line 29: nit: the comment doesn't help much :)

components/server/src/workspace/headless-log-service.ts
```diff
@@ -5,54 +5,92 @@
...
+            const aborted = new Deferred<boolean>();
+            setTimeout(() => aborted.resolve(true), maxTimeoutSecs * 1000);
+            const streamIds = await this.retryWhileInstanceIsRunning(wsi, () => this.supervisorListHeadlessLogs(wsi), "list headless log streams", aborted);
+            if (streamIds !== undefined) {
+                return streamIds;
+            }
         }
 
-        // render URLs
+        // we were unable to get a resonse from supervisor - let's try content service next
```
csweichel, line 44: typo 

```suggestion
        // we were unable to get a response from supervisor - let's try content service next
```

components/content-service-api/headless-log.proto
```diff
@@ -0,0 +1,36 @@
...
+    string url = 1;
+}
+
+message ListLogsRequest {
+    string owner_id = 1;
+    string workspace_id = 2;
+    string instance_id = 3;
+}
+message ListLogsResponse {
+    repeated string taskId = 1;
```
csweichel, line 35:
```suggestion
    repeated string task_id = 1;
```

components/content-service/pkg/service/headless-log-service.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/gitpod-io/gitpod/common-go/log"
+	"github.com/gitpod-io/gitpod/common-go/tracing"
+	"github.com/gitpod-io/gitpod/content-service/api"
+	"github.com/gitpod-io/gitpod/content-service/pkg/logs"
+	"github.com/gitpod-io/gitpod/content-service/pkg/storage"
+)
+
+// HeadlessLogService implements LogServiceServer
+type HeadlessLogService struct {
```
csweichel, line 24: unit tests for the `HeadlessLogService` would be awfully nice  - e.g. by mocking `PresignedAccess`

components/content-service/pkg/logs/logs.go
```diff
@@ -0,0 +1,91 @@
...
+
+const (
+	// TerminalStoreLocation is the path in the workspace where terminal related data like logs or hist-files are
+	// persisted in the file system
+	TerminalStoreLocation = "/workspace/.gitpod"
+
+	prebuildLogFilePrefix = "prebuild-log-"
+
+	legacyTerminalStoreLocation = "/workspace"
+	legacyPrebuildLogFilePrefix = ".prebuild-log-"
```
csweichel, line 23: We've made the switch from this "legacy" naming convention to what we use today about nine months ago - I reckon we could get rid of it.

Not necessarily as part of this PR, only if it makes you life/this code easier.

csweichel (commented):
This is in excellent shape! I have yet to try things out, and would much prefer more tests, but we're really rather close here :)

components/content-service/pkg/service/headless-log-service.go
geropl: I added a test for `ListLogs`. Not super convinced that the test alone justifies the LOC of the mock... but maybe useful for the other services as well.

geropl:
Not 100% this PR, but closely related:
@svenefftinge Discovered yesterday that the current implementation still needs quite some time (e.g., 1 `db-sync` cycle _after_ the pod got created) to be able to connect to a running prebuild because we need to know the `ideURL` to [derive the supervisor URL](https://github.com/gitpod-io/gitpod/blob/823952c95fb3d77f94fdcd0848f5c7a5ca7dab7e/components/server/src/workspace/headless-log-service.ts#L146-L146) from it.

We could try to improve a bit by storing the URL from the `StartWorkspaceResponse` we already get directly into the DB ([here](https://github.com/gitpod-io/gitpod/blob/53bc9e96e33ec8835a16af2f67d423e6face4490/components/server/src/workspace/workspace-starter.ts#L196-L196)) but still depend on a sync cycle.

csweichel (approved):
Thank you for sticking with this PR for so long. Let's get this in

Issue #4352, opened 2021-05-31
Install GitHub App


svenefftinge:
Allow installing the GitHub app from the gitpod.io dashboard. 
The user should be able to install it on their own user or any of their orgs.

This is the first step of the project creation flow and shall be reachable through https://gitpod.io/new.

### User Experience goal

The user should be able configure Prebuilds for a project by either installing the GitHub app for GitHub or installing a webhook for GitLab and Bitbucket.

### Proposal

In this first MVC (minimum viable change), we can do the following:

1. Introduce a full page for setting up a new project.
2. An option to reauthorize which will reconfigure the GitHub app or update the webhook.
3. An option to change the provider from the ones already connected.
3. An option to associate a team for a project.
4. An option to create a team during project import.
4. A new projects list for the team surfacing project name, git repository, active branches, and the latest prebuild status.

### Designs

| Aurthorize | Projects List | Select Account |
|-|-|-|
| ![image](https://user-images.githubusercontent.com/120486/120823543-37491200-c560-11eb-9f11-154ce68795b0.png) | ![image](https://user-images.githubusercontent.com/120486/120823553-387a3f00-c560-11eb-9659-97e5e6e50723.png) | ![image](https://user-images.githubusercontent.com/120486/120823572-3c0dc600-c560-11eb-999d-2f3a225c2341.png) |

| Select Team | Create New Team |
|-|-|
| ![image](https://user-images.githubusercontent.com/120486/120826418-17671d80-c563-11eb-98a4-f4da58fb7925.png) | ![image](https://user-images.githubusercontent.com/120486/120823650-4e87ff80-c560-11eb-9c25-8beb7aee8714.png) |

| Team Projects | Team Projects (Hover) |
|-|-|
| ![image](https://user-images.githubusercontent.com/120486/120823779-75decc80-c560-11eb-80ae-6f427c794adb.png) | ![image](https://user-images.githubusercontent.com/120486/120823905-93ac3180-c560-11eb-901b-70c0025bb48f.png) |

See also [design specs](https://www.figma.com/file/bEPGtpfH5pbx3t3uGY8r0I/Install-GitHub-App-on-Accounts-and-Teams-4352?node-id=0%3A1).

PR #4443, opened 2021-06-09
New Project > Select Repository 

AlexTugarev:
### This PR introduces the project creation flow (gitpod.io/new)

---

### GH App Change

⚠️ Please note, this change requires a significant GH App config change to be aligned with any deployment: instead of redirecting to an URL of the dashboard, it should redirect to the `/api/apps/github/setup` route introduced here. 

Screen Shot 2021-06-14 at 05 39 29

---

### Preview environment
The URL is https://at-select-repo.staging.gitpod-dev.com/new

⚠️ Please note, this preview environment requires an additional re-deployment with werft from within a Gitpod workspace using `werft run github -s chart/values.yaml -f`, where the provided values should contain the GH App config. In addition, a secret with the private key of the GH App should be created like this `kubectl create secret generic server-github-app-cert ...`

---
# New Project

The drop-down list contains the user's account by default. If the GH App is installed on Orgs of the users and the user granted `read:org` permissions, these accounts should be visible as well.

Screen Shot 2021-06-14 at 05 52 26

Screen Shot 2021-06-14 at 05 29 35

- [ ] This list should be filtered to exclude repositories already in used in any project. 

## Select Git Provider

Screen Shot 2021-06-14 at 05 29 41

## Reconfigure GH App installations

You will see the GH App configuration in a pop-up window. For the preview environment this is a test app of mine.

Screen Shot 2021-06-14 at 05 30 05

# Select a team

Listing of all teams of a user. 

- [x] Incorporate "New Team" inline

Screen Shot 2021-06-14 at 05 30 18

# List projects of a team

This is a starting point for the Projects overview. 

Screen Shot 2021-06-14 at 05 30 52

Closes https://github.com/gitpod-io/gitpod/issues/4352


components/server/ee/src/prebuilds/github-app.ts
```diff
@@ -82,6 +80,12 @@ export class GithubApp {
             res.redirect(301, this.getBadgeImageURL());
         });
 
+        options.getRouter && options.getRouter('/whoami').get('/', async (req: express.Request, res: express.Response, next: express.NextFunction) => {
```
AlexTugarev, line 83: todo: remove

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -1810,4 +1809,27 @@ export class GitpodServerImpl<Client extends GitpodClient, Server extends Gitpod
     }
     //
     //#endregion
+
+
+
+    //#region Projects concerns
+    //
+    async getProviderRepositoriesForUser(params: { provider: string }): Promise<ProviderRepository[]> {
```
AlexTugarev, line 1817: this is implemented in EE space because the github-app.ts is also there. 

@svenefftinge, it's ok to move the github-app.ts to the non-EE sources?

components/server/ee/src/prebuilds/github-app.ts
```diff
@@ -103,6 +107,23 @@ export class GithubApp {
...
+
+        options.getRouter && options.getRouter('/reconfigure').get('/', async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+            const gh = await app.auth();
+            const data = await gh.apps.getAuthenticated();
+            const slug = data.data.slug;
+
+            const state = req.query.state;
+            res.redirect(`https://github.com/apps/${slug}/installations/new?state=${state}`)
+        });
+        options.getRouter && options.getRouter('/setup').get('/', async (req: express.Request, res: express.Response, next: express.NextFunction) => {
```
AlexTugarev, line 119: todo: re-introduce `req.query.state` which will be passed on `/reconfigure` in order to identify the origins of this request. IF we get called here without `state`, we can assume, that this is call from the marketplace, thus redirecting to `/install-github-app` should be done.

components/dashboard/src/projects/NewProject.tsx
```diff
@@ -0,0 +1,314 @@
...
+                </div>
+            </>)}
+            {!showGitProviders && (<>
+                <div className="p-6">
+                    <div className="flex flex-col space-y-2">
+                        <label htmlFor="selectAccount" className="font-medium">Account</label>
+                        <select name="selectAccount" value={selectedAccount} className="w-full"
+                            onChange={(e) => onSelectAccount(e.target.value)}>
+                            {accounts.map(a => (<option key={`account-${a}`} value={a}>{`${a} (${provider})`}</option>))}
+                            {isGitHub() && (<option value="addOrg">Add GH Org</option>)}
```
gtsiolis, line 195: **nitpick**: Reminder to avoid acronyms like GH here. 😇

components/dashboard/src/projects/NewProject.tsx
```diff
@@ -0,0 +1,314 @@
...
+    const renderSelectRepository = () => (<>
+        <h3 className="pb-2 mt-8">Select Repository</h3>
+
+        <div className="mt-8 w-96 border rounded-xl border-gray-100 flex-col">
+            {showGitProviders && (<>
+                <div className="p-6">
+                    <div className="text-center text-gray-500">
+                        Select a Git provider first and continue with your repositories.
+                    </div>
+                    <div className="mt-6 flex flex-col space-y-3 items-center">
```
gtsiolis, line 176: **nitpick**: A minor spacing suggestion.

```suggestion
                    <div className="mt-6 flex flex-col space-y-3 items-center pb-8">
```

components/dashboard/src/projects/NewProject.tsx
```diff
@@ -0,0 +1,314 @@
...
+                        )}
+                    </div>
+                </div>
+                <div className="p-3 bg-gray-100">
+                    <div className="text-gray-500 text-center">
+                        Repository not found? <a href="javascript:void(0)" onClick={e => reconfigure()} className="text-gray-400 underline underline-thickness-thin underline-offset-small hover:text-gray-600">Reconfigure</a>
+                    </div>
+                    {noOrgs && (
+                        <div className="text-gray-500 mx-auto text-center">
+                            Missing organizations? <a href="javascript:void(0)" onClick={e => grantReadOrgPermissions()} className="text-gray-400 underline underline-thickness-thin underline-offset-small hover:text-gray-600">Grant permissions</a>
```
gtsiolis, line 225: **issue**: This is not applicable when switching to a GitLab provider, right?

gtsiolis (commented):
Great to see this coming to life @AlexTugarev! Left some minor UX comments below. 🏀

components/dashboard/src/projects/NewProject.tsx
AlexTugarev: Very well. Just realized how surprising the comments are, while changing code at the very same time ;-)

components/dashboard/src/projects/NewProject.tsx
```diff
@@ -0,0 +1,314 @@
...
+                                        </div>
+                                    </div>
+                                ))}
+                            </div>
+                        )}
+                    </div>
+                </div>
+                <div className="p-3 bg-gray-100">
+                    <div className="text-gray-500 text-center">
+                        Repository not found? <a href="javascript:void(0)" onClick={e => reconfigure()} className="text-gray-400 underline underline-thickness-thin underline-offset-small hover:text-gray-600">Reconfigure</a>
```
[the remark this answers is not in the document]
AlexTugarev: Yes. 

When selecting `Reconfigure` and editing the installations of the App on GH, you may have either: 
1. Install on new Account/Org
2. Selected additional Repos (or selected all)
3. Deleted the installation from an Account/Org

All these kinds of changes needs to be reflected in the action (Reconfigure) and also updated after the action is finished, which I doubt happens in all cases. At least after deletion, we're not getting redirected :-( So in that case, where the user does (need to) close the GH pop-up window, we need to update!

components/dashboard/src/projects/NewProject.tsx
AlexTugarev: At the moment this is GitHub specific. 
TODO: disable for GitLab

components/dashboard/src/projects/NewProject.tsx
```diff
@@ -0,0 +1,314 @@
...
+                                {reposToRender.map(r => (
+                                    <div key={`repo-${r.name}`} className="rounded-md whitespace-nowrap flex space-x-2 py-2 px-3 w-full justify-between hover:bg-gray-100 dark:hover:bg-gray-800 focus:bg-gitpod-kumquat-light transition ease-in-out group">
+                                        <div className="w-8/12 text-gray-400 flex flex-col">
+                                            <span className="text-gray-800 dark:text-gray-300 text-base font-semibold overflow-ellipsis truncate">{toSimpleName(r.name)}</span>
+                                            <span className="">Updated __ ago</span>
+                                        </div>
+                                        <div className="w-4/12 flex justify-end">
+                                            <div className="flex self-center hover:bg-gray-200 dark:hover:bg-gray-700 rounded-md cursor-pointer opacity-0 group-hover:opacity-100">
+                                                <button className="primary py-1" onClick={() => selectRepo(r)}>Select</button>
```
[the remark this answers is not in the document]
AlexTugarev: would that apply to the `Select Team` page as well?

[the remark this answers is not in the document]
gtsiolis: In case the project is already part of a team we could trigger a modal with a CTA button to request access to the team, instead of having conditional buttons here _Add Project_ and _Select Team_. WDYT?

AlexTugarev:
https://github.com/gitpod-io/gitpod/pull/4443#discussion_r650974879

> In case the project is already part of a team we could trigger a modal with a CTA button to request access to the team, instead of having conditional buttons here Add Project and Select Team. WDYT?

@gtsiolis, thanks for bringing this up! It's a rather conceptual question, so let's discuss. 

If a repository (aka project) is already linked to a team, what is the implication to the overall concept and the creation flow?Potentially incomplete list of options:
1. Exclude taken repositories on Create Project page.
2. Disable taken repositories on Create Project page.
3. Allow every accessible repo on Create Project page, then add an invitation method.

AlexTugarev:
The check for the state parameter on setup callbacks provides a nice migration path. When installing the app from GH, we get redirected to `/install-github-app` like before

Screen Shot 2021-06-15 at 14 13 11
Screen Shot 2021-06-15 at 14 13 22

svenefftinge:
- [ ] When I add an org, it replaces the org I previously added.

svenefftinge:
- [ ] The new project flow should not have any menus or so. See design docs.

AlexTugarev:
> When I have added the 'orgs' scope, I still don't see the orgs for which installations exist

@svenefftinge, that is true as long as the gitpod-dev OAuth app isn't approved for the org. On https://at-select-repo.staging.gitpod-dev.com/integrations you can select `Manage github.com` or directly navigate to 
 https://github.com/settings/connections/applications/51329bab66290acdf39f and make sure you can approve an organization to test with.

AlexTugarev:
@svenefftinge, I updated the PR with following changes:

- [ ] When I add an org, it replaces the org I previously added.

I'm not sure, what that means. The `Reconfigure` action (showing GitHub's installation page) combines several options: installing, updating, or deleting. 

- [x] Even when I already have a project for an org, I don't see that org anymore in the list.

That should be solved now. `knownOrgs` are preserved in `User.additionalData`, so that when you come back later to `/new`, you should be able to browse the repos of that org. 

- [x] When I land on new, nothing in the combo box is selected.

That might be just related to the missing org, I suppose. In general it takes up to seconds if you have many orgs accessible to the OAuth App of the Gitpod installation. 

- [x] When I added an app installation, the newly added installation is not selected

Done. They should be selected now properly. Also adding `updateAt` for installations and repositories. Repositories are sorted now. 

- [x] When I have added the 'orgs' scope, I still don't see the orgs for which installations exist

This is fixed in the sense that known orgs are preserved on backend now.

AlexTugarev:
> The new project flow should not have any menus or so. See design docs. 

@svenefftinge and @gtsiolis, I aligned with /new-teams on that. 

I'm happy to reduce the menus. For both?

components/gitpod-db/src/typeorm/migration/1622468446118-TeamsAndProjects.ts
```diff
@@ -11,7 +11,7 @@ export class TeamsAndProjects1622468446118 implements MigrationInterface {
...
-        await queryRunner.query("CREATE TABLE IF NOT EXISTS `d_b_project` (`id` char(36) NOT NULL, `cloneUrl` varchar(255) NOT NULL, `teamId` char(36) NOT NULL, `appInstallationId` varchar(255) NOT NULL, `creationTime` varchar(255) NOT NULL, `deleted` tinyint(4) NOT NULL DEFAULT '0', `_lastModified` timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6), PRIMARY KEY (`id`), KEY `ind_teamId` (`teamId`), KEY `ind_dbsync` (`_lastModified`)) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;");
+        await queryRunner.query("CREATE TABLE IF NOT EXISTS `d_b_project` (`id` char(36) NOT NULL, `name` varchar(255) NOT NULL, `cloneUrl` varchar(255) NOT NULL, `teamId` char(36) NOT NULL, `appInstallationId` varchar(255) NOT NULL, `creationTime` varchar(255) NOT NULL, `deleted` tinyint(4) NOT NULL DEFAULT '0', `_lastModified` timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6), PRIMARY KEY (`id`), KEY `ind_teamId` (`teamId`), KEY `ind_dbsync` (`_lastModified`)) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;");
```
svenefftinge, line 14: do you plan to manually update the staging DB?

components/gitpod-protocol/src/protocol.ts
```diff
@@ -1205,6 +1208,27 @@ export interface Project {
     deleted?: boolean;
 }
 
+export interface ProjectInfo {
```
svenefftinge, line 1211: Why copying the whole `Project` shape instead of referencing it?

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -1421,4 +1427,164 @@ export class GitpodServerEEImpl extends GitpodServerImpl<GitpodClient, GitpodSer
         await this.subscriptionService.addCredit(user.id, 50, now);
         return 'Thank you for you feedback. We have added 50 Gitpod Hours to your account. Have fun!';
     }
+
+    //#region Projects concerns
+    //
+    async getProviderRepositoriesForUser(params: { provider: string, hints?: object }): Promise<ProviderRepository[]> {
+        const user = this.checkAndBlockUser("getProviderRepositoriesForUser");
```
svenefftinge, line 1434: this method should go the respective GitProvider impl. We should not add dependencies to GitHub (octokit) across our code.

components/server/src/workspace/gitpod-server-impl.ts
svenefftinge: Yes, that is fine

svenefftinge (approved):
Works well. It needs additional passes, of course, but since it is not activated / visible by default it would be good to merge ASAP. So we can iterate in smaller batches.

components/server/ee/src/workspace/gitpod-server-impl.ts
AlexTugarev: +1 for moving out. I'm thinking of a `GitHubAppSupport` injectable.  

FYI we have two octokit dependencies
- first in our GitHub support (auth provider, context parsers, etc.)
- second comes in via probot
Both parts live in different contexts. More specifically: the GitHub support is instantiated per Git Provider host. 

BTW it's ~impossible~ impractical to sync both octokit versions both. Also, obtaining GH App's authenticated API (octokit) is done via `GitHubApp.server.probotApp.auth()`.

components/gitpod-db/src/typeorm/migration/1622468446118-TeamsAndProjects.ts
AlexTugarev: good point! this is already on staging. 
yes, it would be best to do a manual update.

components/gitpod-protocol/src/protocol.ts
AlexTugarev: Just learned that `DBProject` isn't implementing `Project` at all.
So, yeah it doesn't make sense to duplicate. 

Besides that, we introduced `xInfo` shapes to augment entities, to make it works nicer with the frontend.

components/gitpod-db/src/typeorm/project-db-impl.ts
```diff
@@ -27,4 +28,19 @@ export class ProjectDBImpl implements ProjectDB {
         const repo = await this.getRepo();
         return repo.find({ teamId });
     }
+
+    public async createProject(name: string, cloneUrl: string, teamId: string, appInstallationId: string): Promise<Project> {
+        const repo = await this.getRepo();
+
+        const project: Project = {
+            id: uuidv4(),
+            name,
```
jankeromnes, line 37: Nit: Since `name` is also used as a `slug`, maybe validate it here against a regex? For example like so:

https://github.com/gitpod-io/gitpod/blob/903c8545e5eba0365ba008465ac634b2cfd0c254/components/gitpod-db/src/typeorm/team-db-impl.ts#L74-L76

PR #4458, opened 2021-06-09
Refactor parsing of port numbers from strings

aledbf:


components/gitpod-cli/cmd/forward-port.go
```diff
@@ -26,7 +27,7 @@ var portFwdCmd = &cobra.Command{
 	Long:  ``,
 	Args:  cobra.RangeArgs(1, 2),
 	Run: func(cmd *cobra.Command, args []string) {
-		srcp, err := strconv.ParseInt(args[0], 10, 32)
+		srcp, err := num.ParseInt32(args[0])
```
csweichel, line 30: I'm not sure I follow the benefit of this change. The `strconv.ParseInt` would produce about the same error, no?

components/supervisor/cmd/tunnel.go
```diff
@@ -56,7 +57,7 @@ var closeTunnelCmd = &cobra.Command{
 	Short: "close the tunnel",
 	Args:  cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		localPort, err := strconv.Atoi(args[0])
+		localPort, err := num.ParseUint32(args[0])
```
csweichel, line 60: Why not https://golang.org/pkg/strconv/#ParseUint?

csweichel (approved):
Other than the type safety, I'm not sure I follow the benefit of this change over using `strconv` directly. 

I reckon it's likely that in the future we'd see a mix of these functions and `strconv` as new contributors (or our future selfs) might not remember these functions and default to `strconv`.

components/gitpod-cli/cmd/forward-port.go
aledbf: fixed

components/supervisor/cmd/tunnel.go
aledbf: fixed

csweichel (approved):
LGTM

PR #4461, opened 2021-06-10
[ws-proxy] tune idle connection pool

akosyakov:
### What it does

By default the connection pool is capped only by 2 connections per host.
If a client tries to open many connections then latency increased significantly, see https://stackoverflow.com/a/60114945/961588. This commit configures the connection pool be unlimited but capped by 100 connections per host. It was inspired by measurement in Caddy: https://github.com/caddyserver/caddy/issues/2805

I tried different values:
- 32 was giving a result over 1mins with cached data
- 100 was giving a result around 20s with cached data
- 1000 was giving a result around 30s with cached data

#### How to test

- Start a workspace for https://ak-proxy-test.staging.gitpod-dev.com/#https://github.com/gitpod-io/vscode
- Wait till VS Code compiles and tab on port 3000 opened.
- Open 3000 port first time and wait that everything is cached.
- After that reload several times again and check that each time VS Code pops up and requests does not time out.

We should also have some load testing.

### DIFF
--- a/chart/templates/ws-proxy-configmap.yaml
+++ b/chart/templates/ws-proxy-configmap.yaml
@@ -39,8 +39,8 @@ data:
             "transportConfig": {
                 "connectTimeout": "10s",
                 "idleConnTimeout": "60s",
-                "websocketIdleConnTimeout": "180s",
-                "maxIdleConns": 100
+                "maxIdleConns": 0,
+                "maxIdleConnsPerHost": 100
             },
             "blobServer": {
                 "scheme": "http",
--- a/components/proxy/conf/Caddyfile
+++ b/components/proxy/conf/Caddyfile
@@ -80,6 +80,14 @@
 	header X-Gitpod-Region {$GITPOD_INSTALLATION_LONGNAME}
 }
 
+(workspace_transport) {
+	transport http {
+		tls_insecure_skip_verify
+		keepalive 60s
+		keepalive_idle_conns 100
+	}
+}
+
 # Kubernetes health-check
 :8003 {
 	respond /live 200
@@ -202,10 +210,7 @@ https://*.*.{$GITPOD_DOMAIN} {
 		}
 
 		reverse_proxy https://ws-proxy.{$KUBE_NAMESPACE}.{$KUBE_DOMAIN}:9090 {
-			transport http {
-				tls_insecure_skip_verify
-			}
-
+			import workspace_transport
 			import upstream_headers
 
 			header_up X-WSProxy-Host       {http.request.host}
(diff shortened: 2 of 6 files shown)


chart/templates/ws-proxy-configmap.yaml
```diff
@@ -40,7 +40,8 @@ data:
                 "connectTimeout": "10s",
                 "idleConnTimeout": "60s",
                 "websocketIdleConnTimeout": "180s",
-                "maxIdleConns": 100
+                "maxIdleConns": 0,
```
geropl, line 43: A comment here would be nice.

components/ws-proxy/pkg/proxy/pass.go
```diff
@@ -160,6 +160,7 @@ func createDefaultTransport(config *TransportConfig) *http.Transport {
 		}).DialContext,
 		ForceAttemptHTTP2:     true,
 		MaxIdleConns:          config.MaxIdleConns,                   // default: 100
```
akosyakov, line 162: @geropl Waht is default in comment? Is it golang default or ours?

akosyakov:
Setting 0 and 32 did help, but it is slow. I would play a bit more with Caddy and ws-proxy settings to see whether it can be faster.

components/ws-proxy/pkg/proxy/pass.go
geropl: Goland std lib default.

akosyakov:
It comes from 1 mis 20s to just 20s if 100 connections per host are allowed, let's see if 1000 makes  bigger difference.

chart/templates/ws-proxy-configmap.yaml
akosyakov: I added a comment to the commit. Adding it here breaks the build.

components/proxy/conf/Caddyfile
```diff
@@ -80,6 +80,14 @@
 	header X-Gitpod-Region {$GITPOD_INSTALLATION_LONGNAME}
 }
 
+(workspace_transport) {
+	transport http {
+		tls_insecure_skip_verify
+		keepalive 2s
```
akosyakov, line 86: 60s

akosyakov:
FYI: I'm using this branch to work on VS Code please don't merge. I don't have other env which allow to open VS Code from sources reliably.

geropl:
> It comes from 1 mis 20s to just 20s if 100 connections per host are allowed, let's see if 1000 makes bigger difference.
> Updated: 1000 actually makes everything slower again, will stick with 100.

Also, this is hard to judge in preview envs because there we have 3 proxies in a row: staging proxy -> proxy -> ws-proxy
We'd have to measure in staging on a fat/ws-cluster and compare the result to have sth reliable.

akosyakov:
@geropl How should we proceed then? It can be load tested only on staging?

geropl (approved):
Works as expected. Takes more than 20s with my connection, but absolutely no timeouts whatsoever. :+1: 

@akosyakov IMO we should merge as we're lagging the prober tools (per-branch staging) to test this, and the changes are reasonable. We can still test and fine-tune on staging if necessary.

PR #4466, opened 2021-06-10
[werft] Do not fail job if there is no coverage files to upload

aledbf:


jankeromnes (approved):
Looks good to me! Many thanks.

PR #4467, opened 2021-06-10
[tests] Use TLS certs for connections to ws-manager

geropl:
Fixes the ws-manager connection issues.

The tests did not start successfully, but that's a separate issue.

#### Test
 - `cd test/tests/workspace`
 - `go test -run ^TestBackup$`
 See that it does not stop with a ws-manager connection issue but actually tries to start an instance.


aledbf (approved):
LGTM

PR #4468, opened 2021-06-10
[protocol] ContextURL.parse

geropl:
This adds code that is able to parse our `contextURL` field into an actual `URL`.

This came up here: https://github.com/gitpod-io/gitpod/pull/4262#issuecomment-857485960

#### Test
 - start prebuild: https://gpl-context-url-parser.staging.gitpod-dev.com/#prebuild/github.com/gitpod-io/gitpod/tree/gpl/test-prebuild (note the missing scheme in the URL)
 - check that the prebuild starts and the frontend streams logs successfully


components/gitpod-protocol/src/context-url.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * Copyright (c) 2020 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+export namespace ContextURL {
+  export const INCREMENTAL_PREBUILD_PREFIX = "incremental-prebuild";
+  export const PREBUILD_PREFIX = "prebuild";
```
csweichel, line 9: I'm not big on this because it centralises the otherwise server local prefix concern in gitpod-protocol.
Can't we add the protocol after the prefix parsing has happened?

geropl: > Can't we add the protocol after the prefix parsing has happened?

I'm not sure I understand.

We can indeed move this to `server`. I'm just wondering how the frontend is doing URL parsing... I bet if we dig a bit we'll find similar code.

svenefftinge: I too think this should not be centralized. We should rather test the individual prefix-context-parsers.

geropl: > We should rather test the individual prefix-context-parsers.

That's also possible. But what about code like [this](https://github.com/gitpod-io/gitpod/blob/1c6895539f5153f1e90af6e7a034665197d8fe8a/components/dashboard/src/start/StartWorkspace.tsx#L152-L153), then?

> I too think this should not be centralized

IMO having a single piece of code returning a `ContextURL` of sorts would be more suitable. The independence of the multiple index parsers feels way too artificial and brittle.

components/gitpod-protocol/src/context-url.ts
```diff
@@ -0,0 +1,50 @@
...
+   * This is the analogon to the (Prefix)ContextParser structure in "server".
+   */
+  export function parseToURL(contextUrl: string | undefined): URL | undefined {
+    if (contextUrl === undefined) {
+      return undefined;
+    }
+
+    const segments = contextUrl.split("/");
+    switch (segments.length) {
+      case 0: return undefined;             // this is definitely odd
```
corneliusludmann, line 24: As far as I know this never happens. `"".split("/") === [""]`

components/gitpod-protocol/src/context-url.ts
corneliusludmann: That's tricky.

Code places like 

https://github.com/gitpod-io/gitpod/blob/5fcaa6d52b945d0634360f877e8511c67479b38e/components/dashboard/src/start/StartWorkspace.tsx#L152-L153

are pretty awful. There are already prefixes missing there. How to know that you have to adjust this place, when you add a new `IPrefixContextParser`?

On the other hand, having the prefix centralized here counteracts the dependency injection style. Would be great if we could collect the prefixes dynamically somehow.

But all in all, I find this the lesser evil than that prefixes are hard-coded somewhere across the code.

@geropl Is there any good reason that you didn't replace the prefix strings in `StartWorkspace.tsx` with the variables?

components/gitpod-protocol/src/context-url.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * Copyright (c) 2020 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+export namespace ContextURL {
+  export const INCREMENTAL_PREBUILD_PREFIX = "incremental-prebuild";
```
corneliusludmann, line 8: Could you add at least the `imagebuild/` here as well? Don't know whether `additionalcontent` makes also sense (but that's a regex).

image

components/gitpod-protocol/src/context-url.ts
geropl: > @geropl Is there any good reason that you didn't replace the prefix strings in StartWorkspace.tsx with the variables?

I just pushed the commit that uses function above in frontend and backend.

corneliusludmann (approved):
As discussed synchronously, I'm in favor of merging this
a) to fix the underlying bug for the upcoming deployment and
b) because I think that is actually an improvement

but also creating an issue to encourage further improvements of this change with link to this discussion.

geropl:
Issue for continued discussion: https://github.com/gitpod-io/gitpod/issues/4636

PR #4474, opened 2021-06-10
[dev-environment] Do not use a pinned version and use new codecov uploader

aledbf:
**Changes:**
- Remove pulumi
- Update gcloud sdk from 245 to 344
- Switch to binary codecov uploader

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,4 +1,5 @@
-image: eu.gcr.io/gitpod-core-dev/dev/dev-environment:aledbf-trigger.4
+image:
+  file: dev/image/Dockerfile
 workspaceLocation: gitpod/gitpod-ws.code-workspace
 checkoutLocation: gitpod
 ports:
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,5 +1,4 @@
-image:
-  file: dev/image/Dockerfile
+image: eu.gcr.io/gitpod-core-dev/dev/dev-environment:aledbf-trigger.4
 workspaceLocation: gitpod/gitpod-ws.code-workspace
 checkoutLocation: gitpod
 ports:
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,4 +1,4 @@
-image: eu.gcr.io/gitpod-core-dev/dev/dev-environment:aledbf-trigger.4
+image: eu.gcr.io/gitpod-core-dev/dev/dev-environment:aledbf-devtest.5
 workspaceLocation: gitpod/gitpod-ws.code-workspace
 checkoutLocation: gitpod
 ports:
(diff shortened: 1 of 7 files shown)


.gitpod.yml
```diff
@@ -1,4 +1,4 @@
-image: eu.gcr.io/gitpod-core-dev/dev/dev-environment:aledbf-trigger.4
+image: eu.gcr.io/gitpod-core-dev/dev/dev-environment:aledbf-devtest.5
```
jankeromnes, line 1: What do you think about simplifying our dev setup like so?

```suggestion
image:
  file: dev/image/Dockerfile
```

aledbf: done

.gitpod.yml
```diff
@@ -1,4 +1,5 @@
-image: eu.gcr.io/gitpod-core-dev/dev/dev-environment:aledbf-trigger.4
+image:
+  file: dev/image/Dockerfile
```
csweichel, line 2: This move invites divergence between the environment we use for development and what we use for CI.
Please revert.

.werft/build.ts
```diff
@@ -40,12 +40,6 @@ export async function build(context, version) {
      */
     werft.phase("prepare");
 
-    const werftImg = shell.exec("cat .werft/build.yaml | grep dev-environment").trim().split(": ")[1];
-    const devImg = shell.exec("yq r .gitpod.yml image").trim();
-    if (werftImg !== devImg) {
-        werft.fail('prep', `Werft job image (${werftImg}) and Gitpod dev image (${devImg}) do not match`);
-    }
-
```
csweichel, line 48: See comment above - please revert

.gitpod.yml
jankeromnes: It actually forces one of two things:
- It becomes someone's job to manually update the dev image, frequently and forever
- We constantly sink deeper into deprecation and continuously accumulate security vulnerabilities in our dev environnement

.werft/build.ts
jankeromnes: I don't understand the value of proactively failing at this stage.

What problem is this check trying to solve? And how bad would it be to just let this problem happen and fix it then?

csweichel: The value of pro-actively failing at this stage is to prevent configuration drift. Clearly the check was not exhaustive enough.

.gitpod.yml
csweichel: > It becomes someone's job to manually update the dev image, frequently and forever

Exactly - otherwise we succumb to undefined versions and configuration drift.

> Otherwise we constantly sink deeper into deprecation and continuously accumulate disclosed security vulnerabilities in our dev environnement

I don't see how that's different with using a `Dockerfile` in the `.gitpod.yml`, except that we then have at least two places that we need to keep up to date: the `Dockerfile` and the image we use in CI and elsewhere.

.werft/build.ts
aledbf: done

csweichel (approved):
LGTM

PR #4481, opened 2021-06-11
[db] Make gitpod-db:dbtest-init self-contained by starting a mysql DB (if needed)

geropl:
This automatically starts a test-db if there is none available.

#### Reasoning
Turns out (again) we have the `testDB` sidecar in werft for a reason: It's not straight forward to access a container on the (remote!) docker host. We'd have to use `kubectl port-forward`+ some mechanism to avoid port-overlaps for that. And all this as a special case for werft, which we wouldn't do in the workspace.
Thus I opted for the simplest solution: check if a mysql instance is already available and either a) re-use that (the actual `gitpod` DB is re-created anyways, so no chance we're polluting sth) or b) starts a fresh one. Due to the locking/de-parallelization `leeway` provides us this works nicely. :upside_down_face:

### DIFF
--- a/components/gitpod-db/BUILD.yaml
+++ b/components/gitpod-db/BUILD.yaml
@@ -64,7 +64,13 @@ packages:
     ephemeral: true
     config:
       commands:
+        # Check if a DB is present. If not: start one and wait until it's up
+        # Note: In CI there is a DB running as sidecar; in workspaces we're starting it once.
+        #       Re-use of the instance because of the init scripts (cmp. next step).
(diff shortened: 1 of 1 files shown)


components/gitpod-db/BUILD.yaml
```diff
@@ -64,7 +64,11 @@ packages:
     ephemeral: true
     config:
       commands:
+        # Check if a DB is present. If not: start one and wait until it's up
```
mads-hartmann, line 67: Would be nice to include some of the description from the PR here, e.g. that the DB is present in werft jobs as it is running as a sidebar

mads-hartmann (approved):
LGTM, left a tiny comment

Issue #4470, opened 2021-06-10
[server] Optimize query behind `getTeamMembers`


jankeromnes:
`getTeamMembers(teamId)` should not get all Users one by one when building the `TeamMemberInfo[]` array. Instead, it should get them all in a single query.

As requested in https://github.com/gitpod-io/gitpod/pull/4401/files#r648201200

Follow-up to https://github.com/gitpod-io/gitpod/pull/4401

svenefftinge:
fixed as part of https://github.com/gitpod-io/gitpod/pull/4490

PR #4482, opened 2021-06-11
[server] Optimize getTeamMembers (N queries → 1 query)

jankeromnes:
Fixes https://github.com/gitpod-io/gitpod/issues/4470

I've also added a test (passed! ✅) and getting team member infos still works fine:

Screenshot 2021-06-11 at 11 18 47

### DIFF
--- a/components/gitpod-db/src/user-db.spec.db.ts
+++ b/components/gitpod-db/src/user-db.spec.db.ts
@@ -36,6 +36,7 @@ import { DBWorkspaceInstance } from './typeorm/entity/db-workspace-instance';
         const typeorm = testContainer.get<TypeORM>(TypeORM);
         const mnr = await typeorm.getConnection();
         await mnr.getRepository(DBUser).delete({});
+        await mnr.getRepository(DBIdentity).delete({});
         await mnr.getRepository(DBWorkspace).delete({});
         await mnr.getRepository(DBWorkspaceInstance).delete({});
     }
--- a/components/gitpod-db/src/team-db.spec.db.ts
+++ b/components/gitpod-db/src/team-db.spec.db.ts
@@ -15,6 +15,7 @@ import { TypeORM } from './typeorm/typeorm';
 import { DBTeam } from './typeorm/entity/db-team';
 import { DBTeamMembership } from './typeorm/entity/db-team-membership';
 import { DBUser } from './typeorm/entity/db-user';
+import { DBIdentity } from './typeorm/entity/db-identity';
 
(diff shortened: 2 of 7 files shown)


components/gitpod-db/src/user-db.spec.db.ts
```diff
@@ -36,6 +36,7 @@ import { DBWorkspaceInstance } from './typeorm/entity/db-workspace-instance';
         const typeorm = testContainer.get<TypeORM>(TypeORM);
         const mnr = await typeorm.getConnection();
         await mnr.getRepository(DBUser).delete({});
+        await mnr.getRepository(DBIdentity).delete({});
```
jankeromnes, line 39: @geropl On a side note, while manipulating identities in `TeamDBSpec`, I stumbled on a weird failure in `UserDBSpec`:


```
  1) UserDBSpec
       createUserAndFindById:

      AssertionError: expected { Object (id, creationDate, ...) } to have deep property 'identities' of [ Array(1) ], but got [ Array(1) ]
      + expected - actual

           "authId": "1234"
           "authName": "gero"
           "authProviderId": "GitHub"
           "deleted": false
      -    "primaryEmail": "tom@example.com"
      +    "primaryEmail": [undefined]
           "readonly": false
           "tokens": []
         }
       ]
```

See: https://werft.gitpod-dev.com/job/gitpod-build-jx-optimize-get-team-members.2

That's when I realized that identities are **not** deleted when users are deleted:

https://github.com/gitpod-io/gitpod/blob/a6c548135d2ae3d1b10fb24cb1a07ae828ae4b55/components/gitpod-db/src/typeorm/entity/db-user.ts#L41-L51

My drive-by fix here was to also clean up the `DBIdentity` repo after cleaning up `DBUser` (in both `UserDBSpec` and `TeamDBSpec`). But that comment makes a good point -- maybe we should cascadeRemove identities?

AlexTugarev: Wasn't the reason that we are not cascade deleting related items that it needs to go through dbsync? i.e. we need to enable soft-deletion for that.

jankeromnes:
Anyone please approve. Can be safely merged any time.

csweichel (approved):
LGTM

jankeromnes:
Thanks!

This was already merged as https://github.com/gitpod-io/gitpod/commit/236d81d49aec37538c4912b62092bca4c50d018b (as part of https://github.com/gitpod-io/gitpod/pull/4490).

PR #4485, opened 2021-06-11
[registry-facade] Reload static layer on config change

csweichel:
This PR makes registry-facade reload its static layer configuration when its configuration file changes.
The benefit of this change is simplified deployment, because no red/black deployment is required to update supervisor or docker-up.

### How to test
1. mofidy the `registry-facade-configmap` and watch the logs of registry-facade pick up the change

### Open Questions
- how do we signal that the config was updated? Right now there's a lot of verbose logging, but that doesn't integrate so well with an automated deployment process.


components/registry-facade/cmd/run.go
```diff
@@ -82,6 +85,7 @@ var runCmd = &cobra.Command{
 		if err != nil {
 			log.WithError(err).Fatal("cannot create registry")
 		}
+		go watchConfig(args[0], reg)
```
mads-hartmann, line 88: I'd prefer something like `var configPath = args[0]` at the very top of this function instead of referencing `args[0]` twice

components/registry-facade/cmd/run.go
```diff
@@ -134,3 +138,75 @@ func authorizerFromDockerConfig(cfg *configfile.ConfigFile) docker.Authorizer {
 		return
 	}))
 }
+
+// watchConfig watches the configuration file and if changed reloads the static layer
+func watchConfig(fn string, reg *registry.Registry) {
```
mads-hartmann, line 143: I have a preference towards longer variable names, so something lie `configPath` (if it is a path) instead of `fn`

components/registry-facade/cmd/run.go
```diff
@@ -134,3 +138,75 @@ func authorizerFromDockerConfig(cfg *configfile.ConfigFile) docker.Authorizer {
...
+			}
+
+			if event.Op&fsnotify.Write == fsnotify.Write {
+				reload = true
+			}
+
+			if reload {
+				err = reloadConfig()
+				if err == nil {
+					log.Warn("configuration was updated - reloaded static layer config")
```
mads-hartmann, line 199: I'd prefer info over warning as reloading the configuration is intended behaviour.

mads-hartmann (commented):
Nice, left a few preliminary comments

components/registry-facade/cmd/run.go
csweichel: Usually I'd agree, although in go `fn` is a fairly standard name choice and in combination with the `watchConfig` function its semantic is pretty clear I'd think

mads-hartmann (approved):
Sweet, LGTM.

PR #4489, opened 2021-06-11
[ws-proxy] Allow workspace router to match URLs from other clusters

### LINKED ISSUE #4374: Missing re-direct for outdated workspace clusters
We're currently creating and dropping new workspace cluster very fast, which is good but leads to the following bug on removal:
When the DNS entry `ws-eu04.gitpod.io` (+ the underlying cluster) is removed, but you still have a tab with a timed-out workspace, suddenly your request ends up on a "fat" cluster. That has two issues:
 - [x] fat clusters only have a certificate for "gitpod.io", which leads to a "Warning: Security risk, do not proceed" page: @wulfthimm solved by adding those cluster names + introduced cluster cycling (because Google cannot manage wildcard certs *.ws-eu04.gitpod.io`
 - [ ] The fat cluster has no route for requests of the form `|workspaceId|.ws-eu04.gitpod.io`, and returns a 404. Instead, it should re-direct to `/start/|workspaceId|`, just like for workspaces from the same cluster.
image

corneliusludmann:
This change allows the ws-proxy to accept workspace URLs from other clusters. Missing workspaces will be routed to `/start/<workspace ID>`.

Fixes #4374

## How to test
- Start a workspace
- Stop this workspace
- Open the workspace URL but change `ws-dev` to something like `ws-eu01` in the URL (ignore SSL error).
- The URL should be redirected and the workspace should start with `ws-dev` in the URL.

## Discussion

This is a pretty lightweight change that changes the `matchWorkspaceHostHeader` check to allow all workspace cluster host prefixes. An alternative would be a dedicated route for that. It would probably make the implementation cleaner but also more complex.

@geropl: What are your thoughts about this implementation?

### DIFF
--- a/components/ws-proxy/pkg/proxy/workspacerouter.go
+++ b/components/ws-proxy/pkg/proxy/workspacerouter.go
@@ -35,7 +35,7 @@ const (
 // WorkspaceRouter is a function that configures subrouters (one for theia, one for the exposed ports) on the given router
 // which resolve workspace coordinates (ID, port?) from each request. The contract is to store those in the request's mux.Vars
 // with the keys workspacePortIdentifier and workspaceIDIdentifier
-type WorkspaceRouter func(r *mux.Router, wsInfoProvider WorkspaceInfoProvider) (theiaRouter *mux.Router, portRouter *mux.Router, blobserveRouter *mux.Router)
+type WorkspaceRouter func(r *mux.Router, wsInfoProvider WorkspaceInfoProvider) (ideRouter *mux.Router, portRouter *mux.Router, blobserveRouter *mux.Router)
 
 // HostBasedRouter is a WorkspaceRouter that routes simply based on the "Host" header
 func HostBasedRouter(header, wsHostSuffix string) WorkspaceRouter {
@@ -51,7 +51,7 @@ func HostBasedRouter(header, wsHostSuffix string) WorkspaceRouter {
 			}
 			blobserveRouter = r.MatcherFunc(matchBlobserveHostHeader(wsHostSuffix, getHostHeader)).Subrouter()
 			portRouter      = r.MatcherFunc(matchWorkspacePortHostHeader(wsHostSuffix, getHostHeader)).Subrouter()
-			theiaRouter     = r.MatcherFunc(matchWorkspaceHostHeader(wsHostSuffix, getHostHeader)).Subrouter()
+			ideRouter       = r.MatcherFunc(matchWorkspaceHostHeader(wsHostSuffix, getHostHeader)).Subrouter()
 		)
 
 		r.NotFoundHandler = http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
@@ -59,7 +59,7 @@ func HostBasedRouter(header, wsHostSuffix string) WorkspaceRouter {
 			log.Debugf("no match for path %s, host: %s", req.URL.Path, hostname)
 			w.WriteHeader(404)
 		})
-		return theiaRouter, portRouter, blobserveRouter
+		return ideRouter, portRouter, blobserveRouter
 	}
 }
 
--- a/components/ws-proxy/pkg/proxy/workspacerouter.go
+++ b/components/ws-proxy/pkg/proxy/workspacerouter.go
@@ -38,8 +38,14 @@ const (
 type WorkspaceRouter func(r *mux.Router, wsInfoProvider WorkspaceInfoProvider) (ideRouter *mux.Router, portRouter *mux.Router, blobserveRouter *mux.Router)
 
 // HostBasedRouter is a WorkspaceRouter that routes simply based on the "Host" header
-func HostBasedRouter(header, wsHostSuffix string) WorkspaceRouter {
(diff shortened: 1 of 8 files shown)


components/ws-proxy/pkg/proxy/workspacerouter.go
```diff
@@ -51,15 +52,22 @@ func HostBasedRouter(header, wsHostSuffix string) WorkspaceRouter {
...
 			blobserveRouter = r.MatcherFunc(matchBlobserveHostHeader(wsHostSuffix, getHostHeader)).Subrouter()
 			portRouter      = r.MatcherFunc(matchWorkspacePortHostHeader(wsHostSuffix, getHostHeader)).Subrouter()
-			theiaRouter     = r.MatcherFunc(matchWorkspaceHostHeader(wsHostSuffix, getHostHeader)).Subrouter()
+
+			// Allow to match all cluster prefixes (ws-eu01, ws-eu01, ws-us01, ...)
+			// independent of the prefix of the cluster this component is
+			// deployed to. If the URL has an older WS cluster prefix that does
+			// not exist anymore, the workspaceMustExistHandler function in
+			// routes.go redirects to /start/.
+			allClusterWsHostSuffixRegex = fmt.Sprintf(`(.ws[a-zA-Z0-9-]*)?.%s`, wsHostName)
```
csweichel, line 61: This ties the domain structure down which is a stronger requirement than what we've had so far.
Until now the only requirement was that the workspace domains have to be subdomains of the Gitpod installation as a whole. Now they also need to start with `ws`.

Is there a way we could do away with this restriction, i.e. not start imposing it in the first place?

csweichel (commented):
Other than the comment, this looks good to me

components/ws-proxy/pkg/proxy/workspacerouter.go
corneliusludmann: Yes, that's right. I came from the direction that our helm chart always generates a workspace domain starting with `ws`.

https://github.com/gitpod-io/gitpod/blob/7d0986778e5294bf571c4e96ac013bf46f1319ca/chart/templates/ws-proxy-configmap.yaml#L52

---

What we can do is to move the regex as an optional value to the config file resp. `values.yaml` file and use as default the value of [`workspaceHostSuffix`](https://github.com/gitpod-io/gitpod/blob/7d0986778e5294bf571c4e96ac013bf46f1319ca/chart/templates/ws-proxy-configmap.yaml#L52) (as we do now). That would allow us to configure this for our Gitpod SaaS deployment only (in gitpod-com) and change the regex at any time so that it matches all our workspace clusters.

What do you think, @csweichel?

geropl: > Is there a way we could do away with this restriction, i.e. not start imposing it in the first place?

In general we always had a restriction that we generate in the chart: Up until now it was a suffix match, now a regex which Cornelius proposes to generate and inject from the chart as well.

But it indeed feels like a tad to complex. What if we had no restriction on the 2nd level at all, e.g. use `[^\.]+` instead of `ws[a-zA-Z0-9-]*`? IMO even the check before was too much, and served no purpose beyond a gut feeling of "these request should not go here".

csweichel: I reckon the less restriction/need for configuration we build in here the better.
If we can get away with something like `[^\.]+` we should do that.

components/ws-proxy/pkg/proxy/workspacerouter.go
```diff
@@ -51,7 +52,14 @@ func HostBasedRouter(header, wsHostSuffix string) WorkspaceRouter {
...
 			blobserveRouter = r.MatcherFunc(matchBlobserveHostHeader(wsHostSuffix, getHostHeader)).Subrouter()
 			portRouter      = r.MatcherFunc(matchWorkspacePortHostHeader(wsHostSuffix, getHostHeader)).Subrouter()
-			ideRouter       = r.MatcherFunc(matchWorkspaceHostHeader(wsHostSuffix, getHostHeader)).Subrouter()
+
+			// Allow to match all cluster prefixes (ws-eu01, ws-eu01, ws-us01, ...)
+			// independent of the prefix of the cluster this component is
+			// deployed to. If the URL has an older WS cluster prefix that does
+			// not exist anymore, the workspaceMustExistHandler function in
+			// routes.go redirects to /start/.
+			allClusterWsHostSuffixRegex = fmt.Sprintf(`(\.ws[^\.]*)?\.%s`, strings.ReplaceAll(wsHostName, `.`, `\.`))
```
csweichel, line 61: We're still imposing this convention on workspace clusters. Is there a way around that (e.g. as discussed previously)?

corneliusludmann: Moved that to the help chart and made it configurable in `values.yaml`.

geropl:
@corneliusludmann I tried to start a workspace but it ended up re-creating new workspaces over and over, each stopping immediately. Maybe we need special configuration for devstaging? Or sth that came in with the rebase?

geropl (approved):
Code looks fine, tested and works :+1:

PR #4492, opened 2021-06-12
[build] Produce and use proper version manifest

csweichel:
This PR produces a helm-chart compatible `versions.yaml` file that pins  the version for each Docker image individually. This is a precursor to faster and more flexible deployments, because a change to one component does not require a re-deployment of all the others.

### How does it work
When leeway builds a Docker package it produces an `imgnames.txt` file as build result. I've [extended leeway](https://github.com/gitpod-io/leeway/commit/bca9795d3b29982c56b2d9ca7aac2b1213ca396e) so that when the package config contains a `metadata` field, leeway serializes that metadata to a `metadata.yaml` file alongside the `imgnames.txt`. I've added appropriate metadata to all relevant Docker packages (https://github.com/gitpod-io/gitpod/commit/ecc7f8ebf963d0505cf52f4ad6ee61eefc9a0c7c).

In a previous PR (#4326) we've already started producing Docker images tagged with the leeway package version. Using this prior work, I've added a small script that consumes the `imgnames.txt` and `metadata.yaml` to produce a Helm compatible values file that pins the version of each component.

This `versions.yaml` is shipped in a separate Docker image which can be consumed during deployment. core-dev deployments use this file to decide on the image versions.

### How to test
Inspect the deployed pods in a preview env of this branch.


aledbf (approved):
LGTM

PR #4496, opened 2021-06-13
Update codecov configuration to allow partial uploads

aledbf:
Because the project is a monorepo and we use leeway to build (packages are built in isolation), we end up with a small coverage report file. Unfortunately, this creates a problem with the report in codecov.io introducing a fluctuation of the coverage (at the level project).

This PR refactors the upload of the report to include the context (package) of the change to only update a particular directory in the repository tree.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,109 @@
+# curl --data-binary @codecov.yml https://codecov.io/validate
+codecov:
+  require_ci_to_pass: yes # wait for all other statuses to pass before sending its status
+  branch: main
+  # strict_yaml_branch: main
+
+coverage:
+  precision: 2
+  round: down
+  range: "40...100"         # less than 40% having a red background
+
+comment:
+  layout: "reach,diff,flags"
+  behavior: default         #  update, if exists. Otherwise post new.
+  require_changes: true     # if true: only post the comment if coverage changes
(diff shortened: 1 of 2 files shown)


codecov.yml
```diff
@@ -0,0 +1,109 @@
...
+  range: "40...100"         # less than 40% having a red background
+
+comment:
+  layout: "reach,diff,flags"
+  behavior: default         #  update, if exists. Otherwise post new.
+  require_changes: true     # if true: only post the comment if coverage changes
+  require_base: no          # [yes :: must have a base report to post]
+  require_head: yes         # [yes :: must have a head report to post]
+
+flags:
```
csweichel, line 19: is there a way to centralise this? I.e. make it part of the Go package definition somehow?
Maybe by introducing a better naming scheme of the coverage file, or by providing this mapping from within leeway so that we can produce it when we run codecov.

csweichel (approved):
Happy to see the fluctuation go away, but I'd rather decentralise the mapping of coverage file name to component path (see comment below).

codecov.yml
aledbf: @csweichel flags is the solution provided by codecov https://docs.codecov.com/docs/flags
Also, I want to switch `strict_yaml_branch: main` to on in a future PR to ensure we don't change the configuration in a PR and only have one valid configuration

PR #4497, opened 2021-06-14
[build] Don't re-tag images, don't re-create namespace

csweichel:
This PR disables the automatic re-tagging of all Docker images during the build. Using the [new leeway flag](https://github.com/gitpod-io/leeway/commit/ac53f4e3f984537ae6d028db6931fa6217b294c7) `--dont-retag` we disable this behaviour by default, except in `main` builds or when `with-retag` is set in the build config.

Also, this PR disables the default re-creation of namespaces in core-dev. This might lead to issues down the road because config map changes might not lead to container restarts which can be unexpected. However, we would be seeing those kind of issues in prod, too - so this is a good way to clean up the chart.
If you still want a clean slate deployment, add `with-clean-slate-deployment` to your build config.

In addition, this PR contains a range of cleanup bits:
- actually remove the ws-manager integration tests from the build. We haven't run them in a long time, but still incurred the build time penalty because of its dependencies (e.g. charts)
- improves the version manifest to produce a valid YAML file when multiple components sit below each other
- moves service-waiter into `components` in the `values.yaml`, s.t. it works with the version-manifest process
- integrates gitpod-db/db-migrations into the version manifest

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,4 +1,4 @@
-image: eu.gcr.io/gitpod-core-dev/dev/dev-environment:cw-version-mf-ii.0
+image: eu.gcr.io/gitpod-core-dev/dev/dev-environment:cw-dont-retag.2
 workspaceLocation: gitpod/gitpod-ws.code-workspace
 checkoutLocation: gitpod
 ports:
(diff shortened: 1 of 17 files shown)


JanKoehnlein:
That's an awesome step towards more lightweight and more frequent deployments. Thanks!

aledbf (approved):
LGTM

PR #4498, opened 2021-06-14
[code] logging to track reconnection cause

### LINKED ISSUE #4448: code frontend fails to load sometimes
### Bug description

Sometimes the loading screen does not go away and I see following error in dev tools:
```
Uncaught Error: Check dependency list! Synchronous require cannot resolve module '@improbable-eng/grpc-web'. This is the first mention of this module!
    at ModuleManager.synchronousRequire (loader.js:1421)
    at RequireFunc (loader.js:1842)
    at Object.346 (localapp.js:1)
    at __webpack_require__ (localapp.js:1)
    at Object.674 (localapp.js:1)
    at __webpack_require__ (localapp.js:1)
    at Object.127 (localapp.js:1)
    at __webpack_require__ (localapp.js:1)
    at localapp.js:1
    at localapp.js:1
```

Not really sure how to reproduce. It seems that @improbable-eng/grpc-web is required before it is loaded.

### Steps to reproduce

See above.

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

akosyakov:
#### What it does

- It adds logging to determine a cause of web socket closing.
- It also changes the default websocket in VS Code Web:
  - by default VS Code will try to reconnect the websocket only if it was not closed cleanly (1000 code)
  - in our case it is better to reconnect always when workspace phase is running regardless of websocket close code
- fix #4448: makes sure that grpc-web is loaded before the local app

Changes in Gitpod Code: https://github.com/microsoft/vscode/compare/e2eb3ee4af68d4d14c2c62f99a64c97842f9ad02...a6ceaaf126d636d6e9499dfebaf413956256e98d

#### How to test

- Start a workspace.
- Run `WebSocket.disconnectWorkspace` and check web sockets are reconnected within a minute.
- Throttle network on system level, like with Edge profile, and see that web sockets still get connected and reconnected.

### DIFF
--- a/components/ide/code/leeway.Dockerfile
+++ b/components/ide/code/leeway.Dockerfile
@@ -42,7 +42,7 @@ RUN curl -fsSL https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh |
     && npm install -g yarn node-gyp
 ENV PATH $NVM_DIR/versions/node/v$NODE_VERSION/bin:$PATH
 
-ENV GP_CODE_COMMIT e2eb3ee4af68d4d14c2c62f99a64c97842f9ad02
+ENV GP_CODE_COMMIT a6ceaaf126d636d6e9499dfebaf413956256e98d
 RUN mkdir gp-code \
     && cd gp-code \
     && git init \


JanKoehnlein (approved):
LGTM

PR #4499, opened 2021-06-14
WORKSPACE_KERNEL environment variable

fntlnz:
Needed because the werft build executor and the final workspace are not
guaranteed to run the same kernel.

Also simplified the scripts by using the workspace kernel coming from
the workspace's package manager directly instead of re-downloading.


fntlnz:
Related broken werft build, reported by @akosyakov and investigated by @corneliusludmann 

https://werft.gitpod-io-dev.com/job/gitpod-build-ak-upgrade-core.48

corneliusludmann (approved):
Thanks for fixing this! :+1: 

The code looks good so far. I actually don't know whether something special should be tested as a reviewer. I guess as long as the build is green it's fine?

I'll leave you an approval here. Ping me, in case you still want me to test something.

PR #4517, opened 2021-06-15
Use value that k8s uses

meysholdt:
When the chart is applied, k8s changes the value from `Unmasked` to `Default`.
To prevent misunderstandings, we should use the same value in our code.

In case the value should remain `Unmasked` and k8s needs additional pampering to stop changing it - let me know.


csweichel (approved):
It's not obvious why ws-daemon works regardless of this field's value. I reckon the reason is because we're running with `hostPID: true` and mount the node mounts separately.

Issue #3976, opened 2021-04-16
[server/metrics] Cannot read property 'path' of undefined


AlexTugarev:
### Bug description

see https://github.com/gitpod-io/gitpod/pull/3110#discussion_r614814967 for context.

there appears to be an unexpected request made. 

### Steps to reproduce

---

### Expected behavior

---

corneliusludmann:
With #4518 this message should not occur once deployed to production. However, the actual problem could not be solved yet. We should watch for `route is undefined` messages after the deployment of #4518.

corneliusludmann:
In staging, we see the following:

![image](https://user-images.githubusercontent.com/24960040/123601661-3a9a8b00-d7f8-11eb-9f1d-9a4874a5a521.png)

https://cloudlogging.app.goo.gl/iWdXJTKftX9sLvS28 (internal)

AlexTugarev:
I commended here: https://github.com/gitpod-io/gitpod/pull/4518/files#r659578823

PR #3110, opened 2021-02-04
code user data sync

akosyakov:
#### What it does

This PR exposes the code sync endpoint by the server component. The endpoint stores the metadata of resources in the DB and the content in the content service.

Changes in Code: https://github.com/gitpod-io/vscode/commit/f12434cc6ef580b85fbb457daaf8935670ddacc5

Difference between Code and Theia:
In Theia we operate user and workspace level extensions. In Code all extensions are user extensions, but they can be scoped by the machine (workspace origin in our case) via disabling the extension synching.

#### How to test

- smoke test Theia startup
- enable feature preview and Code
- start 2 workspaces
- change user settings (for instance color theme) and install extensions (for instance gitlens) in one workspace and check that they are reflected in another
- start another workspace and check that new user settings and extensions are applied

PR #4518, opened 2021-06-15
[server] Fix req route metrics

corneliusludmann:
In the log files, we can see that we have some requests where `req.route` `undefined` is. That leads to the following error:

> TypeError: Cannot read property 'path' of undefined

Unfortunately, I was not able to reproduce this error in a preview environment.

This change catches this error and logs the path of the request when this happens.

See #3976 and https://github.com/gitpod-io/gitpod/pull/3110#discussion_r614814967.

### DIFF
--- a/components/server/src/server.ts
+++ b/components/server/src/server.ts
@@ -85,7 +85,10 @@ export class Server<C extends GitpodClient, S extends GitpodServer> {
             res.send = (body) => {
                 const result = originalSend(body);
                 const method = req.method;
-                const route = req.route.path;
+                const route = req.route?.path || "";
+                if (!req.route) {
+                    log.warn(`route is undefined, url: ${req.url}`);
+                }
                 observeHttpRequestDuration(method, route, res.statusCode, (Date.now() - startTime) / 1000)
                 increaseHttpRequestCounter(method, route, res.statusCode);
                 return result;


akosyakov (approved):
LGTM

components/server/src/server.ts
```diff
@@ -85,7 +85,10 @@ export class Server<C extends GitpodClient, S extends GitpodServer> {
             res.send = (body) => {
                 const result = originalSend(body);
                 const method = req.method;
-                const route = req.route.path;
+                const route = req.route?.path || "";
```
AlexTugarev, line 88: @akosyakov, why using `req.route` and not `req.path` here? is `prometheus-metrics` expecting the relative routes, or would full path' work as well? 

@corneliusludmann what is the effect of passing empty values in `prometheus-metrics`?

akosyakov: because we are not interested in concrete values, but how much some route is hit

akosyakov: Just to clarify:
- route path is `/v1/resource/:resource/:ref`
- url path is instantiation of such route, i.e. `/v1/resource/machines/xyz`

We are interested in former, not latter.

Discussed with Alex that we keep it like that, since it could be interested which URLs are hit often but don't have any route matcher.

corneliusludmann: > what is the effect of passing empty values in `prometheus-metrics`?

Hey @ArthurSens! Could you help out with your expertise? `route` could be an empty string here:

https://github.com/gitpod-io/gitpod/blob/e2bf23a914c8115ec6665a1602a3835c0f66b80d/components/server/src/server.ts#L90-L95

Which results in an empty label here:

https://github.com/gitpod-io/gitpod/blob/e2bf23a914c8115ec6665a1602a3835c0f66b80d/components/server/src/prometheus-metrics.ts#L77-L77

and here: 

https://github.com/gitpod-io/gitpod/blob/e2bf23a914c8115ec6665a1602a3835c0f66b80d/components/server/src/prometheus-metrics.ts#L89-L89

(`route` is an empty string, other labels have values).

Could this lead to problems in production?

---

```typescript
/**
* Increment for given labels
* @param labels Object with label keys and values
* @param value The number to increment with
* @param timestamp Timestamp to associate the time series with
*/
inc(labels: labelValues, value?: number, timestamp?: number | Date): void;
```

```typescript
interface labelValues {
	[key: string]: string | number;
}
```

akosyakov: btw we should be able to check it in staging metrics already

corneliusludmann: > We are interested in former, not latter.

We could URL path as fallback to get a better picture of missing routes. Don't know if it's confusing in the metrics, though.

akosyakov: I thinking it is enough to have them in logs. It can be expensive if happens often, because of high cardinality.

PR #4519, opened 2021-06-15
[build] Fix version manifest

csweichel:
for agent smith and ws-manager


corneliusludmann (approved):
Awesome! :heart_eyes:

PR #4522, opened 2021-06-16
[build] Tag images with commit that produced them

csweichel:
With this PR we tag our images using the commit hash that produced them rather than the leeway package version.
Because we don't re-tag images in branch builds anymore those versions stay stable across core-dev re-deployments.

Once we stop re-tagging the images in main builds, the versions will be stable across main builds, too.


aledbf (approved):
LGTM

PR #4523, opened 2021-06-16
[dashboard] Additional fetch pinned workspace

### LINKED ISSUE #4487: Keep stopping workspaces in the active workspaces list
### Problem to solve

Back in https://github.com/gitpod-io/gitpod/pull/4170 we added this one-line fix to resolve https://github.com/gitpod-io/gitpod/issues/3879. However, this also moved STOPPING workspaces from the Active Workspaces list which could cause some confusion to users. See also [relevant discussion](https://gitpod.slack.com/archives/C01KGM9BUNS/p1623414586055000) (internal). /cc @jankeromnes @geropl 

### Proposal

Keep stopping workspaces in the active workspaces list.

**Priority**: Low 🔅

### LINKED ISSUE #4488: Workspaces list in dashboard: limit is applied before filtering for active workspaces
My “Active” list in production has running (and starting) workspaces only. No pinned workspaces. No stopped workspaces with changes.

Could it be that the limit of 50 is applied before filtering for active workspaces? When I change to a limit of 200 the pinned workspaces appear (they are pretty old). Even when I switch back to 50, the pinned workspaces are still there. After reload, they are gone again.

corneliusludmann:
Fixes #4488

Also fixes #4487 along the way while keeping #3879 fixed.

### How to test

See issue #4488 on how to test this. Would probably make sense to change the limit to something like 3 here:

https://github.com/gitpod-io/gitpod/blob/f378e3a6d8839499604e8863db906dc4d1cc2024/components/dashboard/src/workspaces/workspace-model.ts#L15-L15

and run
```
cd components/dashboard/ && yarn telepresence
```

(otherwise starting > 50 workspaces to these this would not be that handy)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 ## June 2021
 
+- Fix active workspace list in dashboard (show also older pinned workspaces) ([#4523](https://github.com/gitpod-io/gitpod/pull/4523))
 - Adding `ItemsList` component as a more maintainable and consistent way to render a list of workspaces, git integrations, environment variables, etc. ([#4454](https://github.com/gitpod-io/gitpod/pull/4454))
 - Improve backup stability when pods get evicted ([#4405](https://github.com/gitpod-io/gitpod/pull/4405))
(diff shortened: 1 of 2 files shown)


JanKoehnlein (approved):
LGTM

PR #4527, opened 2021-06-16
Use new version of workspace-full image with updated golang setup.

rl-gitpod:


aledbf:
Closing. Already using go 1.17

PR #4532, opened 2021-06-17
[test] Fix integration tests

aledbf:
How to test:
- open build workspace
- in a terminal: `cd test && go test -v ./...`

### DIFF
--- a/test/tests/storage/storage_test.go
+++ b/test/tests/storage/storage_test.go
@@ -17,7 +17,7 @@ func TestCreateBucket(t *testing.T) {
 	it, _ := integration.NewTest(t, 30*time.Second)
 	defer it.Done()
 
-	rsa, err := it.Instrument(integration.ComponentWorkspaceDaemon, "daemon")
+	rsa, err := it.Instrument(integration.ComponentWorkspaceDaemon, "daemon", integration.WithContainer("ws-daemon"))
 	if err != nil {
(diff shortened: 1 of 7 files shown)


test/tests/storage/storage_test.go
```diff
@@ -17,7 +17,7 @@ func TestCreateBucket(t *testing.T) {
 	it, _ := integration.NewTest(t, 30*time.Second)
 	defer it.Done()
 
-	rsa, err := it.Instrument(integration.ComponentWorkspaceDaemon, "daemon")
+	rsa, err := it.Instrument(integration.ComponentWorkspaceDaemon, "daemon", integration.WithContainer("ws-daemon"))
```
csweichel, line 20: maybe we should pull this `WithContainer` bit into the `Instrument` function. It already sees which component we want to instrument, hence could know how the container is called (basically not `kube-rbac-proxy`).

test/tests/workspace/content_test.go
```diff
@@ -39,6 +39,8 @@ func TestBackup(t *testing.T) {
 	}
 	rsa.Close()
 
+	time.Sleep(3600 * time.Second)
```
csweichel, line 42: Debugging?

aledbf: yes

csweichel (approved):

PR #4533, opened 2021-06-17
use value that k8s uses no. 2

meysholdt:
Same as https://github.com/gitpod-io/gitpod/pull/4517, but for agent smith.


fntlnz:
thanks @meysholdt 

This is fine, agent-smith does not need to interact with any of [these paths](https://github.com/kubernetes/kubernetes/blob/7705b300e2085c3864bb1e49a7302bf17f080219/pkg/securitycontext/util.go#L217-L234)

PR #4536, opened 2021-06-17
Fix workspaceUrlPrefixRegex

corneliusludmann:
Fixes the “Open Workspace” button on the stopped page.

Works for deployments like:
- https://abc-nice-brunch-4224.staging.gitpod-dev.com/

but still fails if the branch prefix looks like a workspace ID, e.g. this would still fail because `abc-brunch-4224123` is a valid workspace ID:
- https://abc-brunch-4224123.staging.gitpod-dev.com/


AlexTugarev (approved):
`Open Workspace` link work again! 🎉 

🎉  More tests! 🎉

PR #4541, opened 2021-06-18
[code] fix installing extensions in workspace based on postgres image

### LINKED ISSUE #4460: [code] .gitpod.yml extensions not being installed in VS Code
### Bug description

Originally reported [here](https://community.gitpod.io/t/extensions-not-being-activated/3783) last week:

> I am using [this](https://github.com/KushagraSinha2002/Resourceium/) repository in Gitpod but for some reason, the extensions are not activated at all. Neither are some of the left sidebar features (like the port explorer)
> 
> The extensions tab is completely empty too:
> 
> b830a5d573ca912fc2bbcb33bb9d001aff4ae15e
> 
> This worked fine until a couple of weeks ago, but then I put the project on hold and when I came back, I started encountering this problem. I don’t am not really sure what has actually caused this issue.

### Steps to reproduce

1. Set VS Code as default IDE in https://gitpod.io/preferences
2. Open https://gitpod.io/#https://github.com/KushagraSinha2002/Resourceium/
3. Check Extensions panel

### Expected behavior

Extensions should be installed.

### Example repository

https://github.com/KushagraSinha2002/Resourceium/

### Anything else?

While investigating, we noticed a bunch of errors in the browser console, but they're likely unrelated (some known OpenVSX CORS errors, and some local app connection errors).

However, suspecting something might be wrong with the config, I tried to fork the repository. When I opened my **identical** fork in Gitpod, I suddenly had all extensions installed, but none of the tasks ran, which was super confusing.

akosyakov:
#### What it does

- fix #4460: it is more a work around to make VS Code work when workspaces are based on postgres image. For some reason Node.js process detecting shell env hangs because of stderr although we read everything from it. See https://github.com/nodejs/node/issues/39072 We don' really need sterr pipe in this case, so I ignore it: https://github.com/gitpod-io/vscode/commit/bd7b4627aa9426570398a32606bda2d9246a4e52

#### How to test

Start following workspaces and check that eventually extensions get installed. It can take a while. It maybe that even frontend timeout and you need to reload the page.

https://ak-code-extensions.staging.gitpod-dev.com/#https://github.com/KushagraSinha2002/Resourceium
https://ak-code-extensions.staging.gitpod-dev.com/#https://github.com/connectiveproject/connective

### DIFF
--- a/components/ide/code/leeway.Dockerfile
+++ b/components/ide/code/leeway.Dockerfile
@@ -42,7 +42,7 @@ RUN curl -fsSL https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh |
     && npm install -g yarn node-gyp
 ENV PATH $NVM_DIR/versions/node/v$NODE_VERSION/bin:$PATH
 
-ENV GP_CODE_COMMIT bcd47c10d173a0c5cb174ab7451bb5133c4e9bec
+ENV GP_CODE_COMMIT bd7b4627aa9426570398a32606bda2d9246a4e52
 RUN mkdir gp-code \
     && cd gp-code \
     && git init \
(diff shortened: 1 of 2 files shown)


JanKoehnlein (approved):
LGTM

PR #4543, opened 2021-06-18
[gitpod-schema] Add deprecation message to prebuild task

### LINKED ISSUE #4205: Deprecate `prebuild` task
We should deprecate the `prebuild` lifecycle for tasks as it complicates tasks without adding enough value.
I often see users missusing it. E.g. https://github.com/ray-project/ray/blob/56c309416e04a70d874709b9a612ffb482f3f80a/.gitpod.yml

Adding a deprecation warning to the gitpod yml schema as well as removing the documentation would do as a first step.

corneliusludmann:
Fixes #4205

### DIFF
--- a/components/gitpod-protocol/data/gitpod-schema.json
+++ b/components/gitpod-protocol/data/gitpod-schema.json
@@ -93,7 +93,8 @@
                     },
                     "prebuild": {
                         "type": "string",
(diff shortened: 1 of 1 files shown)


JanKoehnlein (approved):
LGTM

PR #4544, opened 2021-06-18
[gitlab] Handle null in avatar_url

### LINKED ISSUE #4066: Column 'avatarUrl' cannot be null
### Bug description

When some users connect, they need to accept the terms of service, and after that the interface stalls.
After +- 30 seconds the error "Sorry, something went wrong 😓 Error 504" appears.
For other users the tos accept works fine.

I'm runnig a self-hosted installation:
- Rancher 2.5
- RKE v1.19.9
- Kubernetes network layer Flannel
- Gitpod 0.8.0
- Gitlab 13.10.3-ee
- Deployment Helm and Helmfile

[gitpod-values.zip](https://github.com/gitpod-io/gitpod/files/6376726/gitpod-values.zip)

### Steps to reproduce

I log in to Gitlab as admin, impersonate on of the users that has the problem, and try to open a Gitpod session.

### Expected behavior

When accepting the terms of service, Gitpod continues to the workspace.

### Example repository

Sorry, something went wrong 😓
Error 504
Please file an issue if you think this is a bug.

### Anything else?

I have searched through the logs of Gitpod, but I can't find a reason why it should not work for some persons.
All users that authenticate on Gitlab, do so via Active directory credentials.

corneliusludmann:
Fixes #4066

### DIFF
--- a/components/server/src/gitlab/gitlab-auth-provider.ts
+++ b/components/server/src/gitlab/gitlab-auth-provider.ts
@@ -79,7 +79,7 @@ export class GitLabAuthProvider extends GenericAuthProvider {
                 authUser: {
                     authId: String(id),
                     authName: username,
-                    avatarUrl: avatar_url,
+                    avatarUrl: avatar_url || undefined,
                     name,
                     primaryEmail: email
                 },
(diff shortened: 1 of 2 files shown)


AlexTugarev (approved):
LGTM!

PR #4545, opened 2021-06-18
Add version and commit-sha1 to version manifest

meysholdt:
This adds the `version:`  field to the `version` docker image to the file `version.yaml`
This adds the `commit:` fields to the `version` docker image in a new file `commit.yaml`:

Example:
```
/ # cat versions.yaml 
version: main.981
components:
(...)
/ # cat commit.yaml 
commit: 4c612e2e52df097cba7e4bdd2d36facf8551fb17
```

We need the commit-sha1 in a downstream-repo to update a submodule to the right version.
I'm adding `version:` to have "complete" info about all versions.

edit 1: add the fields to `build.yaml` instead of `versions.yaml`
edit 2: put stuff into `versions.yaml` and `commit.yaml`

### DIFF
--- a/components/BUILD.yaml
+++ b/components/BUILD.yaml
@@ -60,7 +60,9 @@ packages:
       - dev/version-manifest:app
     config:
       commands:
-        - ["sh", "-c", "dev-version-manifest--app/version-manifest > versions.yaml"]
+        - ["sh", "-c", "echo \"commit: ${__git_commit}\" > commit.yaml"]
+        - ["sh", "-c", "echo \"version: ${version}\" > versions.yaml"]
+        - ["sh", "-c", "dev-version-manifest--app/version-manifest >> versions.yaml"]
         - ["sh", "-c", "rm -r components* dev-*"]
   - name: all-apps
     type: generic
--- a/components/leeway.Dockerfile
+++ b/components/leeway.Dockerfile
@@ -3,4 +3,5 @@
 # See License-AGPL.txt in the project root for license information.
 
 FROM alpine:3.13
+COPY components--all-docker/commit.yaml /
 COPY components--all-docker/versions.yaml /


csweichel:
beware: this will make `commit` available within helm charts - is that intentional?

Alternatively, you could create another file in that package that contains the commit.

meysholdt:
true, helm does not need to know about the commit. I've updated the code changes and PR description.

mads-hartmann:
@meysholdt I would prefer a json file just to simplify consuming it from Node/Bash, e.g.

```sh
docker run --rm "eu.gcr.io/gitpod-core-dev/build/versions:${gitpod_version}" cat /build.json \
| jq --raw-output '.commit'
        
```

mads-hartmann:
@meysholdt  Having the `version` key in the version.yaml as you had originally does make sense and would simplify our script slightly

csweichel:
> I would prefer a json file just to simplify consuming it from Node/Bash, e.g.

`yq r commit build.yaml` would do the same, no. This way we'd have "just" YAML files and needn't wonder about why one is JSON and the other one isn't.

mads-hartmann:
@csweichel I was not aware of `yq` 👍

meysholdt:
I've updated PR-desc and changes.

mads-hartmann (approved):
🎉

ArthurSens:
Hey, sorry but I'm not following why we need this. By running [this script](https://github.com/gitpod-com/gitpod/blob/master/scripts/prepare-gitpod-release.sh) we can already see the commits that generated the components, don't we?

meysholdt:
> Hey, sorry but I'm not following why we need this. By running [this script](https://github.com/gitpod-com/gitpod/blob/master/scripts/prepare-gitpod-release.sh) we can already see the commits that generated the components, don't we?

true that script did that by talking to werft. During deploy, the connection to werft failed, so I chose to eliminate the need for the connection instead of fixing the connection. See https://github.com/gitpod-com/gitpod/pull/5534

PR #4548, opened 2021-06-18
[supervisor] Make ports configured in .gitpod.yml private by default

### LINKED ISSUE #4159: Never make ports public unless explicitly configured
Currently, ports that are listed in the `.gitpod.yaml` file are public by default. That means everyone that knows the port URL can access them without authentication/authorization. We do this for backward compatibility reasons.

However, public ports can be a security reason in some cases. Therefore, we should break backward compatibility in favor of more secure defaults and make ports public only when the user has configured it explicitly with the `visibility` property in the `.gitpod.yaml` file or by clicking the “Make Public” button.

We should probably mention this change in the changelog dialog to inform our users.

corneliusludmann:
Fixes #4159

### How to test
- Open https://corneliusludmann-never-make-ports-4159.staging.gitpod-dev.com/#https://github.com/corneliusludmann/gitpod-playground/tree/test-port-visibility
- `F1` → View: Show Remote Explorer
- Verify that port 8081 is the only port that is public (it's explicitly configured to be public in the .gitpod.yml file)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 ## June 2021
 
+- Breaking Change: Make ports configured in `.gitpod.yml` private by default when no value for `visibility` is given (was public). This change is for security reasons. ([#4548](https://github.com/gitpod-io/gitpod/pull/4548))
 - Fix active workspace list in dashboard (show also older pinned workspaces) ([#4523](https://github.com/gitpod-io/gitpod/pull/4523))
 - Adding `ItemsList` component as a more maintainable and consistent way to render a list of workspaces, git integrations, environment variables, etc. ([#4454](https://github.com/gitpod-io/gitpod/pull/4454))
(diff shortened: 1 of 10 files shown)


csweichel:
You'll also have to adapt this for the ports that are exposed on workspace start where ws-manager creates the port service: https://github.com/gitpod-io/gitpod/blob/main/components/server/src/workspace/workspace-starter.ts#L609

corneliusludmann:
@csweichel: Thanks for the hint.

What about these 2 places? Do we need to change the default values here as well?

https://github.com/gitpod-io/gitpod/blob/99ec2a5c5733c7ddd04b7854896d6c6be05b3399/components/server/src/workspace/gitpod-server-impl.ts#L1074-L1082

https://github.com/gitpod-io/gitpod/blob/660c7ecd87be74a59e2affb263a2f09467963a3e/components/ws-manager/pkg/manager/manager.go#L749-L763

csweichel:
> @csweichel: Thanks for the hint.
> 
> What about these 2 places? Do we need to change the default values here as well?

Yes

JanKoehnlein (approved):
LGTM

PR #4551, opened 2021-06-20
[registry-facade] Fix config hot-reload

csweichel:
This PR fixes the registry-facade config reload mechanism. Before this PR registry-facade would reload the configuration every 30 seconds even if nothing changed.

### How to test
1. Looking at the registry-facade logs you should NOT see `configuration was updated - reloaded static layer config`
2. After modifying the registry-facade configmap you should see `configuration was updated - reloaded static layer config` in the logs


mads-hartmann (approved):
LGTM 🚀

PR #4552, opened 2021-06-20
[build] Reuse Go build cache

csweichel:
This PR configures the werft job to re-use the Go build cache across the node. It basically places the build cache on the node instead of in the build container, thereby enabling the re-use of the build cache across multiple jobs running on the node.

This should speed up the Go builds.


aledbf (approved):
LGTM

PR #4553, opened 2021-06-20
[build] Don't retag main build

csweichel:
This PR removes the Docker image retagging in the main build.
We can merge this once staging and prod no longer require `main.XXX` Docker images but use the version manifest.


meysholdt:
Can you confirm that `eu.gcr.io/gitpod-core-dev/build/versions` will continue to have the `main.XXX` tag? 

That's the only image for which staging and prod do (and will) require the old image tage.

csweichel:
> Can you confirm that `eu.gcr.io/gitpod-core-dev/build/versions` will continue to have the `main.XXX` tag?


That is the case indeed.

meysholdt (approved):

PR #4554, opened 2021-06-20
[build] Improve build start time

csweichel:
No need to recursively chown the Go cache as all files will have been created by `gitpod:gitpod` anyways.

The chown is still needed for the first run on a new node.


aledbf (approved):
LGTM

PR #4559, opened 2021-06-21
Remove the docs content; explain details in a docs/README.md.

### LINKED ISSUE #4558: Remove the `/docs` directory
For the time being, all documentation lives in https://github.com/gitpod-io/website/tree/main/src/routes/docs. To avoid confusion and potentially two places where people updates docs, let's remove the `/docs` directory in this project.

The goal is to eventually bring all docs (not only self-hosted as it's been the case so far) back into this project here.

mikenikles:
Close #4558.

/werft no-preview


csweichel (approved):
LGTM

PR #4560, opened 2021-06-21
[code] queue write/resize terminal requests

### LINKED ISSUE #4504: Terminal Window Typing Out of Order
### Bug description

This occurred when I was on a cell network.  The network appeared to have a reasonable amount of bandwidth and ping time (>10MBps and ping of around 50).  

When I would use the terminal window, my typing was lagging, but worse, it was picking up the keys in the wrong order.  So if I typed "bundle install" what I would get is something more like "bnudel intsall".  Like if it was just laggy and slow that wouldn't be so bad, but it's totally unsuable if it's corrupting the order.

This was on an ipad pro running Safari.  

### Steps to reproduce

See above.

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

akosyakov:
#### What it does

- fix #4504: queue write/resize terminal requests. Terminal API expects that clients take care to send next request after response to previous.

Changes in Gitpod Code: https://github.com/gitpod-io/vscode/commit/924ed5cf404dab58b6f01fee6f8522f86b8a93b0

It should be merged after https://github.com/gitpod-io/gitpod/pull/4565

#### How to test

- Start a workspace.
- Type something and check that characters appear in proper order without substantial delay on good connection.
- Resize and check with `stty -a` that sizes are changed.
- Emulate slow connection on system level and test again.

### DIFF
--- a/components/ide/code/leeway.Dockerfile
+++ b/components/ide/code/leeway.Dockerfile
@@ -42,7 +42,7 @@ RUN curl -fsSL https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh |
     && npm install -g yarn node-gyp
 ENV PATH $NVM_DIR/versions/node/v$NODE_VERSION/bin:$PATH
 
-ENV GP_CODE_COMMIT 5ee0238977bd53abbe6fbdddff54dfbcd6aed7a3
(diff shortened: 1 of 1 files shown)


akosyakov:
I will move it to the supervisor instead.

JanKoehnlein (approved):

akosyakov:
merged as part of https://github.com/gitpod-io/gitpod/pull/4569

PR #4564, opened 2021-06-21
[agent-smith] BPF vm mount workspace files under /workspace

fntlnz:
Introduces a couple of things:

- When you ssh into the BPF vm for agent smith you can find the workspace mounted at `/workspace`
- Removed the need of SSHing with the password, just `ssh vm`
- Use the `code` binary from inside the VM

### DIFF
--- a/dev/image/Dockerfile
+++ b/dev/image/Dockerfile
@@ -10,7 +10,8 @@ ENV WORKSPACE_KERNEL 5.4.0-1033-gke
 USER root
 
 ### QEMU (x86) and workspace kernel development tools
-RUN install-packages qemu qemu-system-x86 linux-image-$WORKSPACE_KERNEL libguestfs-tools sshpass
+RUN install-packages qemu qemu-system-x86 linux-image-$WORKSPACE_KERNEL libguestfs-tools openssh-sftp-server
+RUN ln -s /usr/lib/sftp-server /usr/libexec/sftp-server
 
 ### Clang and LLVM
 RUN install-packages clang-7 llvm-7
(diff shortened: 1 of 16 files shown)


dev/image/Dockerfile
```diff
@@ -10,7 +10,8 @@ ENV WORKSPACE_KERNEL 5.4.0-1033-gke
 USER root
 
 ### QEMU (x86) and workspace kernel development tools
-RUN install-packages qemu qemu-system-x86 linux-image-$WORKSPACE_KERNEL libguestfs-tools sshpass
+RUN install-packages qemu qemu-system-x86 linux-image-$WORKSPACE_KERNEL libguestfs-tools openssh-sftp-server
```
csweichel, line 13: When this file gets changed I'd expect to see updated `dev` images in the `.gitpod.yml` and similar places (search and replace). Is that omission intentional?

fntlnz: I just didn't do it yet, thanks for the reminder :)

fntlnz: The reason is that the branch needs to be pushed before the image gets pushed to gcr, so I was waiting for the CI to do its thing

fntlnz: Did it now!

csweichel:
Looks like the search and replace introduced a whitespace at the end of the image reference. That's why the tests fail in `gitpod-config_test.go`: https://werft.gitpod-dev.com/job/gitpod-build-lf-workspace-vm-mount.2#build:components/gitpod-protocol/go:lib

fntlnz:
Something still does not work, investigating 🔍

fntlnz:
Ready for another round of review @csweichel

csweichel (approved):
Nice one! Works like a charm.

(please squash the commits for a bit of a cleaner history on `main`)

PR #4565, opened 2021-06-22
[code/supervisor] allow retry auto exposing port

### LINKED ISSUE #3160: [ports] notify a user about failed auto port exposure and allow to restart it
With https://github.com/gitpod-io/gitpod/pull/3120 it can happen that we give up on trying to auto exposure the port if it is blocking for very long time. It should not happen often, but if it happens a user does not know about it and a port is staling in detecting mode. We should display the error state and allow a user to retry auto exposure.

akosyakov:
#### What it does

- fix #3160: display a user that ports failed to auto expose and allow to retry

Change in Gitpod Code: https://github.com/gitpod-io/vscode/commit/13852f7794240b9b6a25dfb321ac548c186212da

This PR also contains a commit which allows only one control port request in 2mins. I will drop it before merging. It is required for testing.

#### How to test

- Start a workspace and create 2 terminals.
- In one terminal run `curl lama.sh | LAMA_PORT=9000 sh`
- In another terminal run `curl lama.sh | LAMA_PORT=9001 sh`
- You should see that only one port will be detected and another will fail.
- In 2 mins after that, you should be able to click retry to expose on failed port and it will succeed as well.


JanKoehnlein (approved):
LGTM

akosyakov:
merged as part of https://github.com/gitpod-io/gitpod/pull/4569

PR #4569, opened 2021-06-22
[code] minify

### LINKED ISSUE #2117: enable minified mode for code ide
We should make sure though that source maps are still working and enable it only on staging and production.

akosyakov:
#### What it does

- fix #2117: It enable minification for Gitpod Code

Changes in Gitpod Code: https://github.com/gitpod-io/vscode/commit/5ef446b4b19381b010df7c42ba47970a8b2bed01

#### How to test

- Start a workspace.
- Open devtools, disable caching, and reload. Look at size of workbench.web.api.js.
- Do the same in gitpod-staging and compare.

My results 2.3Mb vs 3.7Mb.


JanKoehnlein (approved):
LGTM

PR #4575, opened 2021-06-22
Add script for convenient db-migrations in GCP

meysholdt:
Since gitpod-images no longer have version-based tags we can no longer run db-migrations as a side-car of a weft job (unless we pass the commit-based-tag of the db-migrations as a parameter to the job). 

This PR enables us to 
* run db-migrations via `docker run` from the werft job
* get rid of some super-nasty process-syntonisation-code that was used for job-sidecar-communication

Related: https://github.com/gitpod-com/gitpod/pull/5564, https://github.com/gitpod-com/gitpod/issues/5560, https://github.com/gitpod-com/gitpod/pull/5559

This changes introduces `node` instead of `node-slim` as the base image because `slim` misses `wget` and ssl-certs that are being used by the proxy.

To test run:

```
docker run --rm \
            --env GOOGLE_APPLICATION_CREDENTIALS_DATA='...' \
            --env GCP_DATABASE="gitpod-foobar:europe-west1:gitpod-foobar-baz" \
            --env DB_PASSWORD="..." \
            gcr.io/gitpod-core-dev/build/db-migrations:x1 /app/migrate_gcp.sh
```

### DIFF
--- a/components/gitpod-db/leeway.Dockerfile
+++ b/components/gitpod-db/leeway.Dockerfile
@@ -4,17 +4,19 @@
 
 FROM node:12.18.3-slim as builder
 COPY components-gitpod-db--migrations /installer/
-
 WORKDIR /app
 RUN /installer/install.sh
 
+FROM node:12.18.3 as proxy
+RUN wget https://storage.googleapis.com/cloudsql-proxy/v1.23.0/cloud_sql_proxy.linux.amd64 -O /bin/cloud_sql_proxy \
+ && chmod +x /bin/cloud_sql_proxy
 
 FROM node:12.18.3-slim
+COPY migrate.sh /app/migrate.sh
+COPY migrate_gcp.sh /app/migrate_gcp.sh
 RUN mkdir /home/jenkins && chown -R 10000 /home/jenkins
+COPY --from=proxy /bin/cloud_sql_proxy /bin/cloud_sql_proxy
+COPY --from=proxy /etc/ssl/certs/ /etc/ssl/certs/
 COPY --chown=10000:10000 --from=builder /app /app/
 WORKDIR /app/node_modules/@gitpod/gitpod-db
 
-RUN echo "#!/bin/bash"                                                            >> /app/migrate.sh && \
-    echo "yarn --cwd /app/node_modules/@gitpod/gitpod-db run wait-for-db"        >> /app/migrate.sh && \
-    echo "yarn --cwd /app/node_modules/@gitpod/gitpod-db typeorm migrations:run" >> /app/migrate.sh && \
-    chmod +x /app/migrate.sh
--- a/components/gitpod-db/migrate.sh
+++ b/components/gitpod-db/migrate.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+# Copyright (c) 2021 Gitpod GmbH. All rights reserved.
+# Licensed under the GNU Affero General Public License (AGPL).
+# See License-AGPL.txt in the project root for license information.
+
+set -euo pipefail
+
+yarn --cwd /app/node_modules/@gitpod/gitpod-db run wait-for-db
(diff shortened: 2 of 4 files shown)


components/gitpod-db/migrate_gcp.sh
```diff
@@ -0,0 +1,33 @@
...
+# Example usage:
+# docker run --rm \
+#            --env GOOGLE_APPLICATION_CREDENTIALS_DATA='...' \
+#            --env GCP_DATABASE="gitpod-foobar:europe-west1:gitpod-foobar-baz" \
+#            --env DB_PASSWORD="..." \
+#            gcr.io/gitpod-core-dev/build/db-migrations:x1 /app/migrate_gcp.sh
+
+set -euo pipefail
+
+echo "$GOOGLE_APPLICATION_CREDENTIALS_DATA" > /tmp/gcp.json
```
csweichel, line 22: some check for the existence of those env vars together with a sensible usage message would be great.

components/gitpod-db/leeway.Dockerfile
```diff
@@ -2,19 +2,20 @@
...
 RUN mkdir /home/jenkins && chown -R 10000 /home/jenkins
 COPY --chown=10000:10000 --from=builder /app /app/
 WORKDIR /app/node_modules/@gitpod/gitpod-db
 
-RUN echo "#!/bin/bash"                                                            >> /app/migrate.sh && \
-    echo "yarn --cwd /app/node_modules/@gitpod/gitpod-db run wait-for-db"        >> /app/migrate.sh && \
-    echo "yarn --cwd /app/node_modules/@gitpod/gitpod-db typeorm migrations:run" >> /app/migrate.sh && \
-    chmod +x /app/migrate.sh
\ No newline at end of file
+RUN wget https://dl.google.com/cloudsql/cloud_sql_proxy.linux.amd64 -O /bin/cloud_sql_proxy \
```
csweichel, line 17: This should be done before line 14 - there's no need to invalidate the cloud_sql_proxy just because the migrations changed.

components/gitpod-db/leeway.Dockerfile
```diff
@@ -2,19 +2,20 @@
...
-FROM node:12.18.3-slim as builder
+FROM node:12.18.3 as builder
 COPY components-gitpod-db--migrations /installer/
 
 WORKDIR /app
 RUN /installer/install.sh
 
 
-FROM node:12.18.3-slim
+FROM node:12.18.3
```
csweichel, line 12: instead of making the db-migrations image larger you could also resort to a multi-stage build that downloads the `cloud_sql_proxy`.

components/gitpod-db/migrate.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+# Copyright (c) 2021 Gitpod GmbH. All rights reserved.
+# Licensed under the GNU Affero General Public License (AGPL).
+# See License-AGPL.txt in the project root for license information.
+
```
mads-hartmann, line 5: This script is missing `set -euo pipefail`

mads-hartmann (approved):
Left a comment, nothing blocking

components/gitpod-db/migrate_gcp.sh
meysholdt: [`set -u`](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html) automates both for us:
> Treat unset variables and parameters other than the special parameters ‘@’ or ‘*’ as an error when performing parameter expansion. An error message will be written to the standard error, and a non-interactive shell will exit.

That's enough IMHO unless we want error messages that go beyond saying "please set this env var".

components/gitpod-db/leeway.Dockerfile
meysholdt: done

PR #4578, opened 2021-06-23
[chart] Remove registry-proxy from ws-daemon

csweichel:
which was a leftover from when registry-facade was proxied by ws-proxy.


geropl (approved):
:+1:

PR #4580, opened 2021-06-23
[code] workspace sharing support

### LINKED ISSUE #2812: [code] support "Share Running Workspace"
It should be possible to share running workspaces from Code.

### LINKED ISSUE #4528: [code] upgrade after May 2021 recovery release
See https://github.com/microsoft/vscode/milestone/157

akosyakov:

#### What it does

- fix #2812: workspace sharing support
- fix #4528

Change in Gitpod Code: https://github.com/gitpod-io/vscode/commit/11d7256b4a55983fdd7040a7ed01866e086afc85

#### How to test

- Start a workspace.
- You should see `Share` status bar indicator, click on it to start sharing. You should also be able to access owner actions like stop workspace, extend timeout and so on from the command palette and menus.
- Open a workspace link by another user and check that a workspace can be accessed. Share indicator should show `Shared by {owner}`, clicking on it should not do anything. You should not be able to access owner actions.
- Go to owner window again, share indicator should show `Shared` now, clicking on it should stop sharing.
- You can also verify that there are `Share` and `Stop Sharing` commands under the main menu and the remote menu (click on remote indicator in the status bar).

### DIFF
--- a/components/ide/code/leeway.Dockerfile
+++ b/components/ide/code/leeway.Dockerfile
@@ -42,7 +42,7 @@ RUN curl -fsSL https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh |
     && npm install -g yarn node-gyp
 ENV PATH $NVM_DIR/versions/node/v$NODE_VERSION/bin:$PATH
 
-ENV GP_CODE_COMMIT 5ef446b4b19381b010df7c42ba47970a8b2bed01
+ENV GP_CODE_COMMIT 11d7256b4a55983fdd7040a7ed01866e086afc85
 RUN mkdir gp-code \
     && cd gp-code \
     && git init \


JanKoehnlein (changes requested):
Didn't work 😔

- The "Joined" indicator didn't appear
- In both workspaces, I could click on "Share" to unshare
- Unsharing had no effect on the current connection (new connections were disabled)

I tried with two different sharing users

JanKoehnlein:
PS: The preview envs are currently pretty broken. I suggest to always deploy `with-clean-slate-deployment`

akosyakov:
@JanKoehnlein thanks for testing! The issue was that even for invited user I would use an owner user in the vscode extension. I changed now that a user is fetched from the browser session to detect whether current window own the workspace or not. I also limited commands which invited user can access. Please see test description above.

JanKoehnlein (approved):
LGTM now.

I suppose "unsharing" a workspace doesn't terminate an existing connection, or should it?

akosyakov:
> I suppose "unsharing" a workspace doesn't terminate an existing connection, or should it?

It is known a bug for long time. We don't have any channel to communicate that a workspace is unshared. Although if you refresh the page then you won't be able to access anymore.

mikenikles:
This is exciting news! Could you add an item to the `CHANGELOG.md` file please so we won't miss it when we announce the next release 🙏?

PR #4586, opened 2021-06-23
[server/auth] check for matching auth ID

AlexTugarev:
This PR adds a test for matching `authId` on permission update.

# how to test
1. ensure being logged in
2. switch GH account in another tab
3. go to the `Integrations` page and try to update permissions
  –> this should fail because you now trying to authorize with a different GH account
  –> same should be true for GitLab


fntlnz:
It fails and shows this 
immagine

fntlnz:
Works as advertised 
immagine

JanKoehnlein (approved):
LGTM. Thanks for the fast fix

fntlnz (approved):

PR #4589, opened 2021-06-23
Add dodo to animals

a2br:
very cute fat bird, ended up being killed by us introducing apex predators (read: "cats") to its island's wildlife

pretty sure the last dodo's thoughts before being eaten were about all the Gitpod workspaces that would be named after its extinct species


csweichel:
Thanks Anatole for the PR. Indeed, dodo's were a glaring omission. Thank you for rectifying that and granting that last dodo's dying wish :)

csweichel (approved):
LGTM

PR #4590, opened 2021-06-23
[core-dev] Add EE license to all core-dev installations

csweichel:
This PR adds an EE license to all core-dev preview environments. To this end I've extended licensor to support domain patterns in the form of `*.foobar.com` instead of just a direct domain match.
Then I've produced a single EE license for all `*.staging.gitpod-dev.com` domains and made it so that the `DBLicenseSource` falls back to the pre-existing `GITPOD_LICENSE` environment variable if no license is present in the database.

### How to test
1. Admin should work

### DIFF
--- a/components/licensor/ee/pkg/licensor/licensor.go
+++ b/components/licensor/ee/pkg/licensor/licensor.go
@@ -13,6 +13,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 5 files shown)


components/licensor/ee/pkg/licensor/licensor.go
```diff
@@ -162,6 +163,24 @@ func NewEvaluator(key []byte, domain string) (res *Evaluator) {
 	}
 }
 
+func matchesDomain(pattern, domain string) bool {
+	if pattern == "" {
+		return true
+	}
```
meysholdt, line 169: Does this check imply that a license without domain is valid on any domain? 

We should not allow licenses without domain, because if such licenses exists we have no control over who's using them.

csweichel: Yes, but this is not new behaviour! The "default" license works just in that way.

JanKoehnlein (approved):
LGTM

PR #4594, opened 2021-06-23
[ws-daemon] mknod /dev/fuse as S_IFCHR to consider major:minor

fntlnz:
Doing a `mount` with fuse is broken because the fuse device inside the workspace is not specifying the device type.

Actual 
```
gitpod ~ $ ls -la /dev/fuse
-rw-rw-rw- 1 root root 0 Jun 23 09:42 /dev/fuse
```

Expected

```
ls -la /dev/fuse
crw-rw-rw- 1 root root 10, 229 Jun 23 14:06 /dev/fuse
```

Note how on the actual one we miss the device type `c` and the actual mkdev `maj:min` for this reason.

### DIFF
--- a/components/ws-daemon/nsinsider/main.go
+++ b/components/ws-daemon/nsinsider/main.go
@@ -193,13 +193,32 @@ func main() {
 			{
 				Name:  "mknod-fuse",
 				Usage: "creates /dev/fuse",
+				Flags: []cli.Flag{
+					&cli.IntFlag{
+						Name:     "uid",
+						Required: true,
+					},
+					&cli.IntFlag{
+						Name:     "gid",
+						Required: true,
+					},
+				},
 				Action: func(c *cli.Context) error {
 					err := unix.Mknod("/dev/fuse", 0666|unix.S_IFCHR, int(unix.Mkdev(10, 229)))
 					if err != nil {
 						return err
 					}
 
-					return os.Chmod("/dev/fuse", os.FileMode(0666))
+					err = os.Chmod("/dev/fuse", os.FileMode(0666))
+					if err != nil {
+						return err
+					}
+					err = os.Chown("/dev/fuse", c.Int("uid"), c.Int("gid"))
+					if err != nil {
+						return err
+					}
+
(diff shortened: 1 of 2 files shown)


fntlnz:
Ok one problem is solved with this one.


```
sudo sshfs -odebug,sshfs_debug,loglevel=debug -p 23001 gitpod@127.0.0.1:/workspace /tmp/workspace
SSHFS version 2.10.0
fuse: failed to open /dev/fuse: Operation not permitted
```


Now there's an issue with uid/gid mappings on the device.

```
ls -la /dev/fuse
crw-rw-rw- 1 nobody nogroup 10, 229 Jun 23 18:09 /dev/fuse
```

csweichel:
Love it. I took the liberty and pushed a commit to your branch that fixes the ownership issue.

image

csweichel (approved):

fntlnz:
The permission issue now seems to be related to `mount_capable`

from fs/super.c
```
bool mount_capable(struct fs_context *fc)
{
	if (!(fc->fs_type->fs_flags & FS_USERNS_MOUNT))
		return capable(CAP_SYS_ADMIN);
	else
		return ns_capable(fc->user_ns, CAP_SYS_ADMIN);
}
```

However, it looks like the fuse filesystem on kenrnel 5.4.0-1033 and 5.4.0-1030 is

```
static struct file_system_type fuse_fs_type = {
	.owner		= THIS_MODULE,
	.name		= "fuse",
	.fs_flags	= FS_HAS_SUBTYPE | FS_USERNS_MOUNT,
	.init_fs_context = fuse_init_fs_context,
	.parameters	= &fuse_fs_parameters,
	.kill_sb	= fuse_kill_sb_anon,
};
```

So still investigating.

fntlnz:
Looks like mounting in the unprivileged userns works correctly as I'm able to do that accessing the namespace using nsenter.

fntlnz:
Let's merge this change on its own since the purpose of this PR is fulfilled.
I'll continue investigating the permissions issue.

PR #4598, opened 2021-06-24
Add grpc client-side reconnection

aledbf:
In case of connection errors (pod restart), reconnect.

**Example error:**
```console
{"@type":"type.googleapis.com/google.devtools.clouderrorreporting.v1beta1.ReportedErrorEvent","error":"invalid ref: rpc error: code = Unavailable desc = error reading from server: read tcp 10.0.128.131:55298-\u003e172.20.175.142:8080: read: no route to host:\n    github.com/gitpod-io/gitpod/registry-facade/pkg/registry.(*RemoteSpecProvider).GetSpec\n        github.com/gitpod-io/gitpod/registry-facade/pkg/registry/imagecfg.go:55\n  - invalid 
``` 

**How to test:**
- start a workspace and close it.
- kill `ws-manager` pod
- open a new workspace. 

Right now we only see errors in the log (and "Pulling container image …")

### DIFF
--- a/components/ws-manager/pkg/manager/manager.go
+++ b/components/ws-manager/pkg/manager/manager.go
@@ -25,6 +25,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/peer"
 	"google.golang.org/grpc/status"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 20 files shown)


components/ws-manager/pkg/manager/manager.go
```diff
@@ -1089,35 +1090,17 @@ func (m *Manager) connectToWorkspaceDaemon(ctx context.Context, wso workspaceObj
...
 // newWssyncConnectionFactory creates a new wsdaemon connection factory based on the wsmanager configuration
 func newWssyncConnectionFactory(managerConfig Configuration) (grpcpool.Factory, error) {
-	// We use client-side retry when ws-daemon is unavilable. The unavailability need just cover the time ws-daemon
-	// needs to restart. If ws-daemon is unavailable during a connection attempt, the WithBlock and WithBackoffMaxDelay
-	// configure the behaviour.
-	// Once the connection has been established, but becomes unavailable during a call, this retry mechanism takes hold.
-	//
-	// Note: the retry policy only has an effect if ws-manager is run with the GRPC_GO_RETRY=on env var set.
-	//       see https://github.com/grpc/grpc-go/blob/506b7730668b5a13465224b0d8133f974a3f843d/dialoptions.go#L522-L524
-	var retryPolicy = `{
```
csweichel, line 1099: Removing this is fine because
- we never had `GRPC_GO_RETRY` set, or because
- some other option is taking over the functionality this config provided?

aledbf: > some other option is taking over the functionality this config provided?

The changes in the PR introduce three changes:

- grpc.KeepaliveEnforcementPolicy

Allow more frequent pings. Default is 5 minutes https://github.com/grpc/grpc-go/blob/87eb5b7502493f758e76c4d09430c0049a81a557/internal/transport/defaults.go#L40
This implies that more frequent pings result in an error `HTTP/2 error code: ENHANCE_YOUR_CALM Received Goaway too_many_pings`

- grpc.WithKeepaliveParams
Adjust keepalive ping interval to avoid the default https://github.com/grpc/grpc-go/blob/87eb5b7502493f758e76c4d09430c0049a81a557/internal/transport/defaults.go#L35
and quickly to restart of pods. Also, enables `PermitWithoutStream` keepalive pings even with no active RPCs

aledbf: > we never had GRPC_GO_RETRY set, or because

Good point. This env variable was configured only in `ws-manager`. Now is present in all components with grpc

csweichel (approved):

PR #4606, opened 2021-06-24
[werft] No default value for integration test results

geropl:
This renames the GitHub check for "validate changes" to `github-check-changes` so that their no longer confused/overlap with integration tests.
 
- [ ] /werft with-integration-tests

### DIFF
--- a/.werft/util/shell.ts
+++ b/.werft/util/shell.ts
@@ -14,6 +14,9 @@ export const werft = {
         throw err;
     },
     done: (slice) => console.log(`[${slice}|DONE]`),
+    result: (description: string, channel: string, value: string) => {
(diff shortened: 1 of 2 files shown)


.werft/util/shell.ts
```diff
@@ -14,6 +14,9 @@ export const werft = {
         throw err;
     },
     done: (slice) => console.log(`[${slice}|DONE]`),
+    result: (description: string, channel: string, conclusion: "success" | "failure" | string) => {
```
csweichel, line 17: Conclusion is something that's specific to the github channels.
In it's proper form this should be:
```
result: (description: string, channel: string, value: string) => {
    exec(`werft log result -d "${description}" -c "${channel}" ${value}`);
},
```

aledbf (approved):
LGTM

PR #4613, opened 2021-06-25
fixed problem where ws-manger could not be reached by ws-proxy

wulfthimm:
`ws-proxy` was not able to talk to `ws-manager`. This solves the problem.


geropl (approved):
:+1:

PR #4617, opened 2021-06-25
[code] workspace sharing support #4580

JanKoehnlein:
Had to open another PR because terraform didn't release a certificate lock.

See https://github.com/gitpod-io/gitpod/pull/4580


JanKoehnlein:
Switching to the original PR again

PR #4620, opened 2021-06-25
Check if .gitpod.yaml was changed before reviewers open the workspace

fntlnz:
Informs reviewer on whether the current PR changed the .gitpod.yml file or the workfloow that does this check.

It only runs on PRs actually modifying one of those files.

I thought about doing this with werft but werft usually happens after the user start doing a review of the files so I thought that doing it in werft was not correct timing-wise.


JanKoehnlein (approved):
LGTM

PR #4621, opened 2021-06-25
[supervisor] Improve IDE ReadinessHTTPProbe

aledbf:
We can see in the log the time, like `"IDE readiness took 8.021 seconds",`

### DIFF
--- a/components/supervisor/pkg/supervisor/supervisor.go
+++ b/components/supervisor/pkg/supervisor/supervisor.go
@@ -567,11 +567,16 @@ func runIDEReadinessProbe(cfg *Config) {
 	case ReadinessHTTPProbe:
 		var (
 			url    = fmt.Sprintf("http://localhost:%d/%s", cfg.IDEPort, strings.TrimPrefix(cfg.ReadinessProbe.HTTPProbe.Path, "/"))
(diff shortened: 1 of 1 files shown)


components/supervisor/pkg/supervisor/supervisor.go
```diff
@@ -567,24 +567,28 @@ func runIDEReadinessProbe(cfg *Config) {
...
-			tick   = time.NewTicker(1 * time.Second)
+			client = http.Client{Timeout: 1 * time.Second}
+			tick   = time.NewTicker(500 * time.Millisecond)
 		)
 		defer tick.Stop()
+
+		t0 := time.Now()
+
 		for {
+			<-tick.C
```
csweichel, line 578: placing the tick here means we might be 500 milliseconds slower than we need to be if the IDE is already available by the time we start probing.

csweichel (approved):

components/supervisor/pkg/supervisor/supervisor.go
aledbf: yes, I am aware :)

csweichel (approved):

PR #4624, opened 2021-06-25
Remove ArgoCD label from services created by kedge

ArthurSens:
This PR aims to remove the label `app.kubernetes.io/instance` from the services created by Kedge.

'app.kubernetes.io/instance' is automatically added to kedge when it is deployed by ArgoCD. This mechanism is used so ArgoCD knows which resource it is supposed to control and which ones it isn't. If we add this label to the resources created by kedge, ArgoCD will think that ArgoCD should be controlled, and since their resource definitions are not in git, ArgoCD will try to delete them.


meysholdt (approved):
works as expected 🎉

PR #4632, opened 2021-06-28
[analytics] Add Track Event for user deletion in ee

jakobhero:
Make call to IAnalyticsWriter so that the deletion of the user account is tracked.


JanKoehnlein (approved):
LGTM

AlexTugarev:
@jakobhero just a question, but why is this traced in `UserDeletionServiceEE` and not in `UserDeletionService`?

jakobhero:
No particular reason, would you prefer it if I moved it there @AlexTugarev ?

AlexTugarev:
@jakobhero, it appears to be random to put the event for account creation into the non-EE folder, while the account deletion event is in EE folder. yes, for the sake of consistency, it would be great to align if there is no good reason (e.g. dependency) to separate.

jakobhero:
Thanks for the Insights @AlexTugarev, this makes sense to me. I'm closing this in favor of #4647.

PR #4637, opened 2021-06-28
[preview environments] Some quality-of-(dev)live improvements

geropl:
This PR:
 - ensures all pods stay on the NodePool they should
 - thus enabling us to down-scale as planned (and thus cycle nodes regularly, which greatly increases stability)
 - make sure we never fail because of "less CPU" for DaemonSets (agent-smith, ws-daemon, registry-facade)


aledbf (approved):
LGTM

PR #4663, opened 2021-06-30
[werft] Fix for failing build

princerachit:
.werft/wipe-devstaging.ts was not updated to inlude kubectl.ts function changes which caused the build to fail. Added the modifications.


geropl (approved):
:+1:

PR #4666, opened 2021-06-30
Rewrite webview URL in blobserve

csweichel:
This PR adds the ability to modify files in blobserve. We use this to replace `vscode-webview.net` in Code to make webviews work in firefox.


akosyakov (approved):
I was able to open webviews in Chrome and FF with these changes.

PR #4669, opened 2021-07-01
[code] inline static links to blobserve

### LINKED ISSUE #4568: [code] avoid redirects for static resources from blobserve
Right now in order to fetch each static resource we need to do a redirect before hitting blobserve live-long caching paying with latency for each redirect. We can get rid of redirects if there is known endpoint for blobserve which does not depend on region. In such case during the build of code image we can rewrite static URLs.

It also will allow debugging of Gitpod Code Server with the local app and chrome.

akosyakov:

#### What it does

- fix #4568

This PR allows to configure blobserve to inline static links in index.html with exposed blobserve URLs for concrete ws cluster passed from ws-proxy.

Change in Gitpod Code to allows relative source maps: https://github.com/gitpod-io/vscode/commit/ad96a5a674ce1c58635b1d4907d48100121610b9

Out-of-scope: I noticed that we have 6 sequential requests to the supervisor before showing an IDE. It would be good to reduce them or try to perform in parallel on application or transport level (use grpc-web with websocket) instead. On my connection it will save 0.3s then total loading takes around 1s.

#### How to test

- Start a workspace and check that it is functional.
- Open devtools and reload the page, you should see that there are no redirects to blobserve anymore, but index.html have already links and such links are loaded from the browser cache without any request to the server.
- Start another workspace and check in devtools that even on first opening cache is used after index.html is fetched.

### DIFF
--- a/components/ws-proxy/pkg/proxy/routes.go
+++ b/components/ws-proxy/pkg/proxy/routes.go
@@ -7,6 +7,7 @@ package proxy
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"math/rand"
@@ -163,17 +164,13 @@ func (ir *ideRoutes) HandleSupervisorFrontendRoute(route *mux.Route) {
 		})
 	})
 	// always hit the blobserver to ensure that blob is downloaded
-	r.NewRoute().HandlerFunc(proxyPass(ir.Config, func(cfg *Config, req *http.Request) (tgt *url.URL, err error) {
-		var dst url.URL
-		dst.Scheme = cfg.BlobServer.Scheme
-		dst.Host = cfg.BlobServer.Host
-		dst.Path = "/" + cfg.WorkspacePodConfig.SupervisorImage
-		return &dst, nil
+	r.NewRoute().HandlerFunc(proxyPass(ir.Config, func(cfg *Config, req *http.Request) (*url.URL, error) {
+		return resolveSupervisorURL(cfg), nil
 	}, func(h *proxyPassConfig) {
 		h.Transport = &blobserveTransport{
 			transport: h.Transport,
 			Config:    ir.Config.Config,
-			resolveImage: func(req *http.Request) string {
+			resolveImage: func(t *blobserveTransport, req *http.Request) string {
 				var (
 					image = ir.Config.Config.WorkspacePodConfig.SupervisorImage
 					path  = strings.TrimPrefix(req.URL.Path, "/"+image)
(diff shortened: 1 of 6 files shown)


components/ws-proxy/pkg/proxy/routes.go
```diff
@@ -202,13 +207,45 @@ func (ir *ideRoutes) HandleRoot(route *mux.Route) {
...
+					log.WithField("supervisorURL", supervisorURL).WithError(err).Error("could not preload supervisor")
+					return image
+				}
+				_ = resp.Body.Close()
+				if resp.StatusCode != http.StatusOK {
+					log.WithField("supervisorURL", supervisorURL).WithField("statusCode", resp.StatusCode).WithField("status", resp.Status).Error("could not preload supervisor")
+					return image
+				}
+
+				inlineVars := make(map[string]string)
```
csweichel, line 238: I'd rather this be an explicit struct, e.g.
```Go
type BlobserveInlineVars struct {
    IDE string `json:"ide"`
    SupervisorImage string `json:"supervisor"`
}
```

csweichel (commented):
Otherwise this makes sense to me

components/ws-proxy/pkg/proxy/routes.go
```diff
@@ -202,13 +207,45 @@ func (ir *ideRoutes) HandleRoot(route *mux.Route) {
...
 					// no workspace information available - cannot resolve IDE image and path
 					return ""
 				}
-				return info.IDEImage
+				image := info.IDEImage
+				imagePath := strings.TrimPrefix(req.URL.Path, "/"+image)
+				if imagePath != "/index.html" && imagePath != "/" {
+					return image
+				}
+				supervisorURL := resolveSupervisorURL(t.Config).String() + "/main.js"
```
geropl, line 221: A short comment on what follows would be nice.

components/ws-proxy/pkg/proxy/routes.go
```diff
@@ -202,13 +212,50 @@ func (ir *ideRoutes) HandleRoot(route *mux.Route) {
...
 					// no workspace information available - cannot resolve IDE image and path
 					return ""
 				}
-				return info.IDEImage
+				image := info.IDEImage
+				imagePath := strings.TrimPrefix(req.URL.Path, "/"+image)
+				if imagePath != "/index.html" && imagePath != "/" {
+					return image
+				}
+				// blobserve can inline static links in index.html for IDE and supervisor to avoid redirects for each resource
```
geropl, line 226: :heart:

akosyakov:
@csweichel @geropl I added comments and unit tests.

csweichel (approved):
Works like a charm. Code LGTM

PR #4677, opened 2021-07-01
[agent-smith] account for egress traffic

fntlnz:
Introducing the first iteration of a mechanism to count excessive egress traffic over a period of time. Based on the original code from @csweichel - the only difference is the way we account for processes using bpf.

This is what happens when a workspace uses too much egress.

```
{"level":"info","message":"Found violation","penalties":null,"serviceContext":{"service":"agent-smith","version":""},"severity":"INFO","time":"2021-07-01T15:00:58Z","violation":{"SupervisorPID":410501,"Namespace":"","Pod":"","Owner":"","InstanceID":"2eb1dab1-77c3-42fd-99b9-ba3f98800be1","WorkspaceID":"magenta-swallow-alg0r4xe","Infringements":[{"Description":"egress traffic is 144.233 megabytes over limit","Kind":" excessive egress"}],"GitRemoteURL":["https://github.com/fntlnz/bpf-harness"]}}
{"level":"info","message":"Found violation","penalties":["stop workspace"],"serviceContext":{"service":"agent-smith","version":""},"severity":"INFO","time":"2021-07-01T15:01:28Z","violation":{"SupervisorPID":410501,"Namespace":"","Pod":"","Owner":"","InstanceID":"2eb1dab1-77c3-42fd-99b9-ba3f98800be1","WorkspaceID":"magenta-swallow-alg0r4xe","Infringements":[{"Description":"egress traffic is 972.820 megabytes over limit","Kind":"very excessive egress"}],"GitRemoteURL":["https://github.com/fntlnz/bpf-harness"]}}
```


components/ee/agent-smith/pkg/agent/agent.go
```diff
@@ -138,6 +143,7 @@ func NewAgentSmith(cfg Config) (*Smith, error) {
 		notifiedInfringements: notificationCache,
 		perfHandler:           make(chan perfHandlerFunc, 10),
 		metrics:               newAgentMetrics(),
+		pidsMap:               sync.Map{},
```
csweichel, line 146: nit: this is essentially a noop, because `pidsMap` will already be initialized with the default value

components/ee/agent-smith/pkg/agent/agent.go
```diff
@@ -313,37 +319,74 @@ func (agent *Smith) Start(callback func(InfringingWorkspace, []PenaltyKind)) {
 
 	defer abpf.Close()
 
+	agent.cleanupDeadPIDS(ctx)
+
+	egressTicker := time.NewTicker(30 * time.Second)
```
csweichel, line 324: nit: missing `defer egressTicker.Stop()`

components/ee/agent-smith/pkg/agent/agent.go
```diff
@@ -313,37 +319,74 @@ func (agent *Smith) Start(callback func(InfringingWorkspace, []PenaltyKind)) {
...
-				}
+			for {
+				select {
+				case <-egressTicker.C:
+					agent.pidsMap.Range(func(key, value interface{}) bool {
+						p := key.(int)
+						t := value.(time.Time)
+						infr, err := agent.checkEgressTrafficCallback(p, t)
+						if err != nil {
+							return true
```
csweichel, line 336: I reckon at least some debug logging would be handy here

components/ee/agent-smith/pkg/agent/agent.go
```diff
@@ -357,6 +400,48 @@ func (agent *Smith) Start(callback func(InfringingWorkspace, []PenaltyKind)) {
 	}
 }
 
+func (agent *Smith) cleanupDeadPIDS(ctx context.Context) {
```
csweichel, line 403: It's a personal taste thing, but I'd prefer if `cleanupDeadPIDS` were called as Go routine rather than it spawning one itself.
While reading through I was wondering at the callsite of `cleanupDeadPIDS` if we did this only once. Had the callsite been `go cleanupDeadPIDS()` that would have been clearer.

components/ee/agent-smith/pkg/agent/agent.go
```diff
@@ -357,6 +400,48 @@ func (agent *Smith) Start(callback func(InfringingWorkspace, []PenaltyKind)) {
...
+	ticker := time.NewTicker(30 * time.Second)
```
csweichel, line 404: nit: missing `defer ticket.Stop()`

components/ee/agent-smith/pkg/agent/agent.go
```diff
@@ -711,3 +800,44 @@ func mergeInfringingWorkspaces(vws []InfringingWorkspace) (vw InfringingWorkspac
 func (agent *Smith) RegisterMetrics(reg prometheus.Registerer) error {
 	return agent.metrics.Register(reg)
 }
+
+func (agent *Smith) checkEgressTrafficCallback(pid int, pidCreationTime time.Time) (*Infringement, error) {
```
csweichel, line 804: A unit test for `checkEgressTrafficCallback` would be awesome!

components/ee/agent-smith/pkg/network/egress.go
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2021 Gitpod GmbH. All rights reserved.
+// Licensed under the Gitpod Enterprise Source Code License,
+// See License.enterprise.txt in the project root folder.
+
+package network
```
csweichel, line 5: At this point I'm not convienced that this single, quite short functuion, warants its own package

csweichel (commented):
Had a read through the code, currently deploying, will try then

csweichel:
Awesome! I was able to provoke agent smith to step in on excessive egress

components/ee/agent-smith/pkg/agent/agent.go
```diff
@@ -47,6 +51,7 @@ type Smith struct {
 
 	notifiedInfringements *lru.Cache
 	perfHandler           chan perfHandlerFunc
+	pidsMap               sync.Map
```
fntlnz, line 54: We can probably expose a prometheus metrics to see how this grows, wdyt @csweichel ?

components/ee/agent-smith/pkg/network/egress.go
fntlnz: I agree

components/ee/agent-smith/pkg/agent/agent.go
fntlnz: Was a bit tricky because of the dependency with time and with proc but this is addressed in this patch https://github.com/gitpod-io/gitpod/commit/03094bcfe64307cf410a674fd859b0e7a4cb5134.patch

components/ee/agent-smith/pkg/network/egress.go
fntlnz: Done

components/ee/agent-smith/pkg/agent/agent.go
fntlnz: I agree!

components/ee/agent-smith/pkg/agent/agent.go
fntlnz: This is in the ctx.Done, however since the goroutine is being moved out this needs to exist now.

components/ee/agent-smith/pkg/agent/agent.go
fntlnz: A log here would be extremely verbose, I can add one with a non logged by default priority.
Keep in mind that this is done for every found process every 30 seconds.

csweichel: Excellent point. We could add a metric here also, to see how these checks behave, akin to the signature check failures.

csweichel (commented):
One last nit

components/ee/agent-smith/pkg/agent/agent.go
```diff
@@ -240,7 +249,10 @@ type GradedInfringementKind string
 
 // GradeKind produces a graded infringement kind from severity and kind
 func GradeKind(kind InfringementKind, severity InfringementSeverity) GradedInfringementKind {
-	return GradedInfringementKind(fmt.Sprintf("%s %s", severity, kind))
+	if len(severity) > 0 {
```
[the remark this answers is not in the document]
fntlnz: > I'd argue for swapping the code paths to highlight what's the regular case.

Can you explain a bit more what do you mean?
I don't think I understand what is your suggestion here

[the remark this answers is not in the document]
fntlnz: Can you explain a bit more? I wasn't able to understand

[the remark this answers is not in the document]
csweichel: `GradeKind` is more likely to be called with `len(severity) > 0` than not, hence `len(severity) > 0` is the happy path of this function and should be left aligned. Put differently, `len(severity) == 0` is the exception to the rule.

Frankly, the reason I even commented is because I thought `len(severity) == 0` never ordinarily happens, but upon looking through the code again I found that [`InfringementSeverityAudit`](https://github.com/gitpod-io/gitpod/blob/7ebfcf518bf50f6ef59df28f36ac778e5462cd74/components/ee/agent-smith/pkg/agent/agent.go#L228) indeed is an empty string. The above comment still stands, although has become even more nittier than it was before :/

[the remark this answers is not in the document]
fntlnz: ah ok understood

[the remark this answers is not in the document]
fntlnz: should be done now

csweichel (approved):
Awesome. Thank you 🚀

fntlnz:
I'd love to do some performance tuning before merging.

I noticed that cpu usage is quite high compared to other services

```
NAME                                CPU(cores)   MEMORY(bytes)   
agent-smith-592nd                   280m         102Mi           
agent-smith-5k6wk                   332m         107Mi           
agent-smith-6vh5l                   321m         93Mi            
agent-smith-fvq7r                   309m         126Mi           
blobserve-7877498bcf-kw5r5          1m           105Mi           
content-service-77d7c84b7c-pqmzx    1m           5Mi             
dashboard-7f798db967-ll2tj          2m           14Mi            
image-builder-7bd7f874df-z68rp      8m           560Mi           
jaeger-7f8f8ff74-hzgkm              2m           12Mi            
messagebus-0                        177m         109Mi           
minio-58d57d5f7d-8sngl              1m           25Mi            
mysql-0                             22m          243Mi           
proxy-68cfb7c965-pmlvz              1m           16Mi            
registry-facade-dclr8               1m           22Mi            
registry-facade-fj7qg               1m           18Mi            
registry-facade-j5z69               1m           21Mi            
registry-facade-qnwbc               1m           18Mi            
server-5886b66cf7-wdrzr             30m          144Mi           
sweeper-665d98b88c-prb2h            1m           4Mi             
ws-daemon-2w265                     3m           28Mi            
ws-daemon-9gp29                     1m           9Mi             
ws-daemon-c8p6x                     1m           8Mi             
ws-daemon-n968m                     1m           8Mi             
ws-manager-7bff5c5d5d-m5rqh         3m           29Mi            
ws-manager-bridge-db56448cb-9mcpn   1m           87Mi            
ws-proxy-7955dcf85-xxrfc            2m           9Mi             
ws-scheduler-96c8bb45-k466q         9m           50Mi   
```

I need to understand if this is because of a resource leak or because of the mole of data we process.


And I added a counter of active pids that seems that we are currently processing many of them at once

```
# HELP gitpod_agent_smith_monitored_pids Current count of pids under investigation
# TYPE gitpod_agent_smith_monitored_pids counter
gitpod_agent_smith_monitored_pids{process_state="deleted"} 71124
gitpod_agent_smith_monitored_pids{process_state="stored"} 147847
```

In the case above `147847-71124` it means `76723` pids processed every 30 seconds. I'm trying to understand how this number changes in a typical scenario before deploying this.

fntlnz:
This is the top of agent smith in the current situation (before this change), there's probably something wrong - the whole pids processing pipeline needs to be improved.

```
NAME                                 CPU(cores)   MEMORY(bytes)   
agent-smith-t4knf                    9m           37Mi 
```

fntlnz:
Made a change so that only supervisor processes are taken in account for checking for egress.
This was needed because the network namespace is owned by the process tree and not by the process so we now perform the checks only once.

This reflects into a performance improvement because we handle 1 pid instead of thousands.

This can be easily seen in the metrics

```
gitpod_agent_smith_monitored_pids{process_state="deleted"} 1
gitpod_agent_smith_monitored_pids{process_state="stored"} 1
# HELP gitpod_agent_smith_penalty_attempts_total The total amount of attempts that agent-smith is trying to apply a penalty.
# TYPE gitpod_agent_smith_penalty_attempts_total counter
gitpod_agent_smith_penalty_attempts_total{penalty="stop workspace"} 1
```

This metric says, one workspace was started `process_state: stored` one workspace went above the egress limit `penalty: stop workspace` and it was garbage collected `process_state: deleted`

fntlnz:
Resource usage seems also better now, will investigate further optimizations.

```
NAME                                      CPU(cores)   MEMORY(bytes)   
agent-smith-4stl6                         233m         74Mi            
agent-smith-fxj4t                         206m         40Mi            
agent-smith-l6xng                         410m         59Mi            
agent-smith-t6skf                         371m         101Mi           
agent-smith-thjrf                         136m         68Mi            
agent-smith-wgtr6                         185m         69Mi            
agent-smith-zmk7k                         270m         60Mi  
```

fntlnz:
Looks like the performance differences between this and production are due to signature matching (which is totally not changed here). Will investigate further but I believe that this one is ready to go.

fntlnz:
Just tested against the current main. The cpu usage was not introduced by this PR, main already has it.

```
agent-smith-hjfjz                    317m         60Mi            
agent-smith-jttmr                    257m         56Mi            
agent-smith-kfwst                    233m         69Mi            
agent-smith-rgplr                    298m         74Mi            
agent-smith-rj7r4                    301m         89Mi            
agent-smith-wq8rr                    357m         70Mi            
agent-smith-zj2x9                    234m         125Mi    
```

Will investigate on main.

csweichel:
I reckon we can decouple the CPU use of agent smith and this PR, considering how unoptimised the signature checks are.

csweichel:
Agent Smith found the excessive egress event, but failed to stop my workspace.
```
{"level":"info","message":"Found violation","penalties":null,"serviceContext":{"service":"agent-smith","version":""},"severity":"INFO","time":"2021-07-08T07:47:25Z","violation":{"SupervisorPID":1155632,"Namespace":"","Pod":"","Owner":"","InstanceID":"8b6ddb44-0945-4946-93f5-5f36b8d27b34","WorkspaceID":"plum-harrier-1srrlksl","Infringements":[{"Description":"egress traffic is 718.802 megabytes over limit","Kind":"excessive egress"}],"GitRemoteURL":["https://github.com/gitpod-io/template-typescript-node"]}}
{"level":"info","message":"Found violation","penalties":["stop workspace"],"serviceContext":{"service":"agent-smith","version":""},"severity":"INFO","time":"2021-07-08T07:49:25Z","violation":{"SupervisorPID":1600186,"Namespace":"","Pod":"","Owner":"","InstanceID":"47a32706-da82-4475-9bdf-25d4d4d8b7e4","WorkspaceID":"harlequin-horse-w3q83efn","Infringements":[{"Description":"egress traffic is 4111.466 megabytes over limit","Kind":"very excessive egress"}],"GitRemoteURL":["https://github.com/gitpod-io/template-typescript-node"]}}
```

fntlnz:
That seems very strange because I didn't change anything in that code path since your last review @csweichel - is that happening consistently? I just tried it again now and it worked for me, maybe there's something I'm not considering?

csweichel:
That is most odd indeed. Let's get this in and observe the behaviour in the real world.

csweichel (approved):

PR #4687, opened 2021-07-02
[installer] Remove GCP installer

csweichel:
This PR removes the GCP installer. We haven't updated it in a while and the docs no longer refer to it.


geropl (approved):
:+1:

PR #4691, opened 2021-07-02
[agent-smith] Tone down too verbose logging

csweichel:
This PR tones down the logs produced by agent smith and replaces some of them with metrics.

### How to test
During normal operation of agent smith hardly any logs should be produced.


aledbf (approved):
LGTM

PR #4697, opened 2021-07-05
Fix data race

MatthewFagan:
There is a data race in the supervisor terminal multiplexer code. It is a simple fix.

### DIFF
--- a/components/supervisor/pkg/terminal/terminal.go
+++ b/components/supervisor/pkg/terminal/terminal.go
@@ -440,8 +440,9 @@ func (mw *multiWriter) Listen() io.ReadCloser {
 		//       block the closing because the write's blocking.
 		for b := range cchan {
 			n, err := w.Write(b)
+			blen := len(b)
 			done <- struct{}{}
-			if err == nil && n != len(b) {
+			if err == nil && n != blen {
 				err = io.ErrShortWrite
 			}
 			if err != nil {


akosyakov:
> There is a data race in the supervisor terminal multiplexer code.

Could you elaborate please?

akosyakov (approved):
i think it makes sense

MatthewFagan:
Hi

After looking a bit more into the Go specs about slices, I think I have made a mistake.

Here was my original logic:
multiWriter.Write continually gets called with the same slice argument, just with different data in the slice each time.
It pushes this slice to the listeners and waits for them to send back "done" before it returns and the slice is reused for the next data. 

So, in Listen func, anywhere after the line "done <- struct{}{}", the slice that was popped is subject to being overwritten by the other thread.

However, looking further into the go docs, it appears that the length part of the slice is passed by value, so it should not get changed even if the slice data itself gets changed.

akosyakov:
Ideally it would be good to add a unit test which reproduces the issue. It is hard to verify otherwise.

MatthewFagan:
Sure. Actually, this is not the bug I am looking for. I just noticed it as I was looking at the code.
I am looking at my terminal output getting scrambled sometimes. It seems like a race condition in the supervisor terminal code, but I cannot find it. Maybe I an close this pull request and add a bug report to issues instead?

akosyakov:
If you know how to reproduce reliably the issue with steps would help. What do you mean by scramble? The thing is that we simply stream what underlying shell is giving us. You can get scrambled results after resize for instance, but then it would happen even if you try to use such shell on the local machine in similar way

MatthewFagan:
OK, it appears that the problem is I have an old version of supervisor running. I can see the bug fix already here: https://github.com/gitpod-io/gitpod/commit/8b9ce66cb3af7e15c15cf5b6f805228e549e5a2a#diff-219b3956dbd3aaac01f85406db7840dc9aa3d44aa807a3e600adc48263a9f8e1

akosyakov:
@MatthewFagan Do you self-host? On gitpod.io it is not possible to run old version.

akosyakov:
I'm closing it, please open the issue if you know how to reproduce 🙏

PR #4707, opened 2021-07-05
[content-service] Support FromBackupInitializer

### LINKED ISSUE #4692: Restarting workspaces should fail if the backup is not found
### Bug description

We're opportunistically checking if a backup is present and if so restore it. This is independent of the lifecycle of the workspace.

On content-initializer we should tell if we expect a backup, and if don't find one fail the content init

### Steps to reproduce

yesterday's incident

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

csweichel:
which lets clients force backup restoration or fail otherwise.

fixes #4692


test/pkg/integration/workspace.go
```diff
@@ -375,8 +432,31 @@ func (it *Test) WaitForWorkspaceStop(instanceID string) (lastStatus *wsmanapi.Wo
...
+	ctx, cancel := context.WithTimeout(it.ctx, 30*time.Second)
+	defer cancel()
+	var (
+		start       = time.Now()
+		serviceGone bool
+		k8s, ns     = it.API().Kubernetes()
+	)
+	for time.Since(start) < 1*time.Minute {
+		_, err := k8s.CoreV1().Services(ns).Get(ctx, fmt.Sprintf("ws-%s-theia", workspaceID), v1.GetOptions{})
+		if err != nil {
```
aledbf, line 449: this should check for `errors.IsNotFound(err)`
https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors?utm_source=godoc#IsNotFound

aledbf (approved):
LGTM

PR #4708, opened 2021-07-05
fix(components/server): ensure a sane redirect/returnTo query param

leodido:
Signed-off-by: Leo Di Donato <leodidonato@gmail.com>

This PR ensures the value of the `redirect` (or `returnTo`) query parameter is safe.

To do so, it strictly checks the value against the allowed URLs (rather than relying on `startsWith`).
Furthermore, it uses the `URL` interface to ensure URLs are normalized before checking them.

### On a side note

My first PR! 🎈 

Also, my first Typescript code in years and years (literally!) ... 😊 
So please be gentle with me 🙇

### DIFF
--- a/components/server/src/user/user-controller.ts
+++ b/components/server/src/user/user-controller.ts
@@ -612,8 +612,7 @@ export class UserController {
             return;
         }
 
-        const hostUrl = this.env.hostUrl.url as URL;
-        if (this.urlStartsWith(returnToURL, hostUrl.toString()) || this.urlStartsWith(returnToURL, this.env.brandingConfig.homepage)) {
+        if (this.urlStartsWith(returnToURL, this.env.hostUrl.toString()) || this.urlStartsWith(returnToURL, this.env.brandingConfig.homepage)) {
             return returnToURL
         }
 
--- a/components/server/src/user/user-controller.ts
+++ b/components/server/src/user/user-controller.ts
@@ -600,23 +600,24 @@ export class UserController {
         req.query.returnTo = this.getSafeReturnToParam(req);
     }
 
+    protected urlStartsWith(url: string, prefixUrl: string): boolean {
+        prefixUrl += prefixUrl.endsWith("/") ? "" : "/";
+        return url.toLowerCase().startsWith(prefixUrl.toLowerCase());
+    }
+
     protected getSafeReturnToParam(req: express.Request) {
-        const where: string | undefined = req.query.redirect || req.query.returnTo;
-        if (!where) {
+        const returnToURL: string | undefined = req.query.redirect || req.query.returnTo;
+        if (!returnToURL) {
             log.debug({ sessionId: req.sessionID }, "Empty redirect URL");
             return;
         }
-        try {
-            const redirectUrl = new URL(where).toString();
-            const hostUrl = (this.env.hostUrl.url as URL).toString();
-            const brandingUrl = new URL(this.env.brandingConfig.homepage).toString();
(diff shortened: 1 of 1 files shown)


components/server/src/user/user-controller.ts
```diff
@@ -601,15 +601,22 @@ export class UserController {
...
+        const where: string | undefined = req.query.redirect || req.query.returnTo;
+        if (!where) {
+            log.debug({ sessionId: req.sessionID }, "Empty redirect URL");
+            return;
+        }
+        try {
+            const redirectUrl = new URL(where).toString();
+            const hostUrl = (this.env.hostUrl.url as URL).toString();
+            const brandingUrl = new URL(this.env.brandingConfig.homepage).toString();
+            if (redirectUrl === hostUrl || redirectUrl === brandingUrl) {
```
AlexTugarev, line 613: Checking for equality isn't working in the use cases this is meant for. For instance, when switching accounts the browser agent is sent to `/logout?returnTo=<login-url-with-preset-provider>`. The equality check would allow to return to the main URL only.

AlexTugarev (changes requested):
@leodido it's awesome that you're on it to fix the situation! 💯 

I believe it will be more sustainable to not allow addressing any hosts at all. We should rather rewrite all `returnTo=URL` with `returnTo=PATH` with the means of appending the path to the known Gitpod host. This way there is no need to parse user values as URL.

components/server/src/user/user-controller.ts
```diff
@@ -601,15 +601,22 @@ export class UserController {
     }
 
     protected getSafeReturnToParam(req: express.Request) {
-        const returnToURL: string | undefined = req.query.redirect || req.query.returnTo;
-        if (returnToURL) {
-            const hostUrl = this.env.hostUrl.url as URL;
-            if (returnToURL.toLowerCase().startsWith(`${hostUrl.protocol}//${hostUrl.host}`.toLowerCase())) {
```
AlexTugarev, line 607: @leodido if you are looking for a quick fix here, a change that ensures a trailing slash of the host name should work as well. 
```
returnToURL.toLowerCase().startsWith(`${hostUrl.protocol}//${hostUrl.host}/`.toLowerCase())
```

components/server/src/user/user-controller.ts
leodido: I clearly didn't know about this use case. Thanks!

leodido:
@AlexTugarev thanks for your feeback! Would you please take another look at it? :)

components/server/src/user/user-controller.ts
```diff
@@ -600,16 +600,24 @@ export class UserController {
...
-            }
-            if (returnToURL.toLowerCase().startsWith(this.env.brandingConfig.homepage.toLowerCase())) {
-                return returnToURL;
-            }
+        if (!returnToURL) {
+            log.debug({ sessionId: req.sessionID }, "Empty redirect URL");
+            return;
+        }
+
+        const hostUrl = this.env.hostUrl.url as URL;
```
AlexTugarev, line 615: nit: you can remove this and directly call `this.env.hostUrl.toString()` below.

AlexTugarev (approved):
Code looks good to me!
Let's test before merging 🙏🏻

AlexTugarev:
@leodido could you rebase please?

components/server/src/user/user-controller.ts
leodido: Done ✔️

Issue #4710, opened 2021-07-06
[self-hosted] Release 0.10.0


geropl:
After we released [0.9.0](https://github.com/gitpod-io/gitpod/releases/tag/v0.9.0) with quite a delay, we plan to have a regular cadence to not let down our Self-Hosted users.

In that vein - and after the release to gitpod.io last week - we plan to quickly release 0.10.0, which should be a straight-forward "test-and-release" because there have been no bigger changes that should impact Self-Hosted.

### Test
- [x] test on k3s
- [ ] test on GCP

### Tasks
- [x] fix access to self-hosted registry: either #4711 or #4592
- [x] fix UI to show the setup page for Self-Hosted: #4764
- [x] updates docs: https://github.com/gitpod-io/website/issues/745
- [x] fix DB migration are not executed
- [x] docker credentials
  - [x] docs: add to new `values.custom.yaml` (next to minio etc.)
  - [x] docs: migration guide

/cc @corneliusludmann @solomonope

geropl:
Dumping my personal notes here, in hope this might help whoever works on this:
 - [ ] docs:
   - [ ] "Configure Ingress": cerbot: copy'n'paste is broken
   - [ ] replace "find -exec copy broken: "find $WORKDIR/config/live -name "*.pem" -exec cp {} secrets/https-certificates \;" with
     - cp $WORKDIR/config/live/fullchain.pem secrets/https-certificates/tls.crt
     - cp $WORKDIR/config/live/privkey.pem secrets/https-certificates/tls.key
   - [ ] troubleshooting:
     - [ ] container runtime config for k3s:
        ```
        containerRuntime:
        containerd:
            socket: /var/run/k3s/containerd/containerd.sock
        nodeRoots:
            - /var/run/k3s/containerd/io.containerd.runtime.v2.task/k8s.io
        ```
     - [ ] make visible in the navigation
   - [ ] migration guide:
     - [ ] make visible in navigation
     - [ ] 0.9.0 -> 0.10.0
       - mv secrets/https-certificates/fullchain.pem secrets/https-certificates/tls.crt
       - mv secrets/https-certificates/privkey.pem secrets/https-certificates/tls.key

geropl:
[...]

PR #4592, opened 2021-06-23
[chart] Refactor chart to only use cert-manager

aledbf:
- Refactor the helm chart to remove self-generated SSL certificates using helm to certificates from `cert-manager.`
  The chart already contains a section for `cert-manager`, something that is being used in staging and prod.

  By using `cert-manager`, we can use an `Issuer` (namespaces) to generate the certificates using only one internal CA.

- Install the CA certificate in `registry-facade` to allow pulls from the internal docker registry (if enabled) without trust issues.
- Simplified configuration of components affinity.
- New Job to ensure the creation of `gitpod-sessions` database.

PR #4711, opened 2021-07-06
[helm] Fix reverse proxy for builtin registry

corneliusludmann:
With the new proxy implementation with Caddy, the access to the built-in registry is over Caddy which serves the registry with the main proxy certificates instead of the registry certificates (SSL termination). That has the following implications:

- Caddy accesses the registry via https://registry.default.svc.cluster.local. That means the self-signed cert of the registry needs this domain instead of https://registry.my-gitpod.example.com.
- Caddy needs to trust the self-singed certs of the registry (needs to know the CA).
- The dind of the image builder does not need the registry certs since the certs that Caddy uses are (need to be) trusted certs and not self-signed certs (as long as bypassProxy is not set).

I tested this on k3s with [this setting](https://github.com/corneliusludmann/gitpod-k3s-droplet/tree/gitpod-0-10-0) and verified that the preview env still works.

**Alternatives:** The alternative would be to change the behavior of Caddy so that the proxy does not terminate SSL but does a raw redirect to the registry as it was in nginx (when this is possible with Caddy).

See also #4710

/cc @solomonope due to self-hosted
/cc @aledbf since you implemented the Caddy config

### DIFF
--- a/chart/templates/builtin-registry-certs-secret.yaml
+++ b/chart/templates/builtin-registry-certs-secret.yaml
@@ -2,7 +2,7 @@
 # Licensed under the MIT License. See License-MIT.txt in the project root for license information.
 
 {{ if (index .Values "docker-registry" "enabled") }}
-{{- $regName := include "gitpod.builtinRegistry.name" . -}}
+{{- $regName := include "gitpod.builtinRegistry.internal_name" . -}}
 {{ $cm := (index .Values "cert-manager") }}
 {{- if $cm.enabled }}
 
--- a/chart/templates/_helpers.tpl
+++ b/chart/templates/_helpers.tpl
@@ -273,10 +273,16 @@ env:
 
 {{- define "gitpod.builtinRegistry.name" -}}
 {{- if .Values.components.imageBuilder.registry.bypassProxy -}}
-{{ index .Values "docker-registry" "fullnameOverride" }}.{{ .Release.Namespace }}.svc.cluster.local
+{{ include "gitpod.builtinRegistry.internal_name" . }}
 {{- else -}}
 registry.{{ .Values.hostname }}
 {{- end -}}
+{{- end -}}
+
+{{- define "gitpod.builtinRegistry.internal_name" -}}
+{{- $ := .root -}}
+{{- $gp := .gp -}}
+{{ index .Values "docker-registry" "fullnameOverride" }}.{{ .Release.Namespace }}.{{ $gp.installation.kubedomain | default "svc.cluster.local" }}
 {{- end -}}
 
 {{- define "gitpod.comp.version" -}}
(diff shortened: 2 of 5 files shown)


csweichel:
Also assigning @aledbf here due to upcoming changes on self-hosted internal cert management

chart/templates/builtin-registry-certs-secret.yaml
```diff
@@ -2,7 +2,7 @@
 # Licensed under the MIT License. See License-MIT.txt in the project root for license information.
 
 {{ if (index .Values "docker-registry" "enabled") }}
-{{- $regName := include "gitpod.builtinRegistry.name" . -}}
+{{- $regName := include "gitpod.builtinRegistry.internal_name" . -}}
 {{ $cm := (index .Values "cert-manager") }}
 {{- if $cm.enabled }}
```
corneliusludmann, line 7: @wulfthimm: The (self-signed) cert of the registry has always the internal domain `registry.default.svc.cluster.local` as domain, never the public `registry.my-gitpod.example.com` domain. Don't know how the cert-manager works but could this lead to problems with the cert-manager configuration here? Do we need to drop this?

aledbf:
@corneliusludmann I have a draft PR that improves the self-signed certificate generation and do not uses proxy to access the docker-registry. Please check https://github.com/gitpod-io/gitpod/pull/4592

corneliusludmann:
> I have a draft PR that improves the self-signed certificate generation and do not uses proxy to access the docker-registry. Please check #4592

Thanks for the hint, @aledbf! Big PR, couldn't go through everything. :innocent: Does it mean that your PR solves the issues I posted in our slack channel `self-hosted` a few days ago [[here](https://gitpod.slack.com/archives/C01KLC56NP7/p1625126415015200) and [here](https://gitpod.slack.com/archives/C01KLC56NP7/p1625479050025500) (internal)]? Is there an ETA for your PR?

aledbf:
> Does it mean that your PR solves the issues I posted in our slack channel self-hosted a few days ago 

Yes. I need to finish the PR before next Monday.

aledbf:
@corneliusludmann I updated the [PR](https://github.com/gitpod-io/gitpod/pull/4592) and split the commits to be more clear about what is being changed

geropl (approved):
:+1: 

Quick reasoning summary:

### pull
 - each registry-facade installs a re-direct in it's local `/etc/hosts`: [`reg.${domain}:port`](https://github.com/gitpod-io/gitpod/blob/c608b35509581cf3d48a0e8f3707d859d258224b/chart/templates/ws-manager-configmap.yaml#L91-L91) -> registry-facade node port
 - this works because we issue publicly verifiable certs (Let's Encrypt) which include `reg.`
 
## push:
 - prod/staging/devstaging: imagebuilder get's the properly configured GCP creds and the GCP registry has publicly verifiable certs
 - builtin: our chart has an exception that makes the registry available at `registry.domain`
 - the builtin case is broken because we never test it (neither staging nor devstaging)
 - the fix works as advertised

Issue #4693, opened 2021-07-02
Number of team members don't match on the billing


alesanchezr:
### Bug description

When I go to the [team settings](https://gitpod.io/teams) it says that we have 322 members but we deleted (disabled) some of them and the number should be 192 instead.

I'm afraid the billing will also match the 322 instead of 192.

![Teams-—-Gitpod](https://user-images.githubusercontent.com/426452/124292199-4ef5c500-db23-11eb-8eda-4c7bc337fd5e.png)

### Steps to reproduce

1. go to [team settings](https://gitpod.io/teams)
2. Open the team details
3. count the number of items on the list

### Expected behavior

The number of team members should match the number of items in the list.

jankeromnes:
Hi @alesanchezr! Thank you for reporting this problem. I can confirm that something is not right with your Team Plan. We'll investigate why there is such a discrepancy; fix any underlying bug; and of course refund any undue charges.

lechien73:
Hi @jankeromnes - sorry to pile on, but this made me check our Team Unleashed plan too.

We are showing 1649 members, but I scraped the content of the **Manage Team** modal and ran it through a script, which shows a count of 1336. I've spot checked at different places to ensure that the data is the same as the **Manage Team** modal, and all seems to be correct.

Could you add this to your check too, please? The account email is: gitpodteam@codeinstitute.net.

Thanks

jankeromnes:
Hi @lechien73, thank you for double-checking your team account, which seems to be affected as well.

We're hoping to fix this problem for every (potentially affected) team account today or tomorrow, with the fix going to production shortly thereafter. Will post updates to this issue.

lechien73:
@alesanchezr - I don't know if this is helpful to you, but I've made a Chrome extension that helps manage our teams as they get bigger. It's awaiting review in the extension store at the moment, but you can sideload it from here: https://github.com/lechien73/GitpodTeamHelper

It adds some UI elements so you can see how many active and inactive users you have, and highlights the inactive users so they're easier to find when you are scrolling.

jankeromnes:
Wow, cool, thanks for sharing this @lechien73! (This also clearly shows the limitations of our current Teams UI... 🙈)

[...]

PR #4713, opened 2021-07-06
[teams] Fix Team Plan seat de/reactivation logic

jankeromnes:
Attempts to fix https://github.com/gitpod-io/gitpod/issues/4693

---

I believe our current implementation has the following problems:

- Deactivating a Team Plan Seat sets a `cancellationDate` on the `d_b_team_subscription_slot` entry and its related `d_b_subscription` entry (**good**). However, it schedules the Team Plan to have 1 fewer Seat than it currently has, effective at the end of term (**bad** -- this doesn't take into account multiple deactivations in the same period):

https://github.com/gitpod-io/gitpod/blob/8552d9e67cb888dea41f2389e44ce760a87c699e/components/server/ee/src/workspace/gitpod-server-impl.ts#L1238-L1242

- Reactivating a Team Plan Seat **also decreases** the amount of Seats (by maximum 1 each billing cycle). This is obviously wrong (as it should increase the quantity instead):

https://github.com/gitpod-io/gitpod/blob/8552d9e67cb888dea41f2389e44ce760a87c699e/components/server/ee/src/workspace/gitpod-server-impl.ts#L1254-L1258

- Our "Manage Team" modal shows the correct number of active Team Plan Seats, because if filters out Seats that are now `cancelled` (1). By contrast, the total number of Members shown for your Team Plan doesn't filter out `cancelled` Seats (2), thus reporting a too-high number of "Members" (actually "Seats"):

(1)

https://github.com/gitpod-io/gitpod/blob/97b938150a2dc6211ea8ae549f08050bf1e8159f/components/dashboard/src/settings/Teams.tsx#L717-L718

(2)

https://github.com/gitpod-io/gitpod/blob/97b938150a2dc6211ea8ae549f08050bf1e8159f/components/dashboard/src/settings/Teams.tsx#L503-L504

---

This Pull Request implements the following fixes:

- When deactivating a Seat, decrease the Team Plan Seat quantity immediately, not at the end of term. I believe this allows multiple deactivations in the same cycle, without any side effects on billing.
- When reactivating a Seat, **increase** (not decrease) the Team Plan Seat quantity immediately as well (similar effect).
- Filter out cancelled Seats in the Members total count.

### How to test

Purchase a Team Plan. You should still be able to:
- Assign & unassign Seats (without impact to billing)
- Purchase additional Seats (should add a new deferred charge in Chargebee and also increase your next bill)
- Deactivate Seats (they should remain visible until the end of term, and should also decrease your upcoming bill in Chargebee)
- Deactivate multiple Seats in the same cycle (should decrease your next bill by the correct number of Seats, not just 1)
- Reactivate not-yet-cancelled Seats (the Seat should be usable again + you should not be charged again for this already-paid Seat + your next bill should have one more Seat on it)

- [x] /werft with-payment

### DIFF
--- a/components/dashboard/src/settings/Teams.tsx
+++ b/components/dashboard/src/settings/Teams.tsx
@@ -500,7 +500,7 @@ function AllTeams() {
                             <span className="text-sm my-auto text-gray-400 truncate overflow-ellipsis">Purchased on {formatDate(sub?.startDate)}</span>
                         </div>
                         <div className="p-0 my-auto flex flex-col w-2/12">
-                            <span className="my-auto truncate text-gray-500 overflow-ellipsis">{slots.filter(s => s.teamSubscription.id === sub.id).length || "–"}</span>
+                            <span className="my-auto truncate text-gray-500 overflow-ellipsis">{slots.filter(s => s.state !== 'cancelled' && s.teamSubscription.id === sub.id).length || "–"}</span>
                             <span className="text-sm my-auto text-gray-400">Members</span>
                         </div>
                         <div className="p-0 my-auto flex w-4/12">
@@ -726,7 +726,7 @@ function ManageTeamModal(props: {
             <div className="overscroll-contain max-h-96 overflow-y-auto">
                 {slots.map((slot, index) => {
                     return (
-                        <Slot key={slot.id} slot={slot} inputHandler={props.slotInputHandler} />
+                        <SlotInput key={slot.id} slot={slot} inputHandler={props.slotInputHandler} />
                     )
                 })}
             </div>
@@ -744,7 +744,7 @@ interface SlotInputHandler {
     reactivate: (slot: TeamSubscriptionSlotResolved) => void;
 }
 
-function Slot(props: {
+function SlotInput(props: {
     slot: Slot,
     inputHandler: SlotInputHandler;
 }) {
--- a/components/server/ee/src/workspace/gitpod-server-impl.ts
+++ b/components/server/ee/src/workspace/gitpod-server-impl.ts
@@ -1123,7 +1123,7 @@ export class GitpodServerEEImpl extends GitpodServerImpl<GitpodClient, GitpodSer
     // Get the current number of "active" slots in a team subscription (count all "assigned" and "unassigned", but not "deactivated" or "cancelled").
     protected async tsGetActiveSlotQuantity(teamSubscriptionId: string): Promise<number> {
         const slots = await this.teamSubscriptionDB.findSlotsByTeamSubscriptionId(teamSubscriptionId);
-        return slots.filter(slot => !slot.cancellationDate).length;
+        return slots.filter(TeamSubscriptionSlot.isActive).length;
     }
 
     async tsAddSlots(teamSubscriptionId: string, addQuantity: number): Promise<void> {
--- a/components/server/ee/src/workspace/gitpod-server-impl.ts
+++ b/components/server/ee/src/workspace/gitpod-server-impl.ts
@@ -7,7 +7,7 @@
 import { injectable, inject } from "inversify";
 import { GitpodServerImpl } from "../../../src/workspace/gitpod-server-impl";
 import { TraceContext } from "@gitpod/gitpod-protocol/lib/util/tracing";
-import { GitpodServer, GitpodClient, AdminGetListRequest, User, AdminGetListResult, Permission, AdminBlockUserRequest, AdminModifyRoleOrPermissionRequest, RoleOrPermission, AdminModifyPermanentWorkspaceFeatureFlagRequest, UserFeatureSettings, AdminGetWorkspacesRequest, WorkspaceAndInstance, GetWorkspaceTimeoutResult, WorkspaceTimeoutDuration, WorkspaceTimeoutValues, SetWorkspaceTimeoutResult, WorkspaceContext, CreateWorkspaceMode, WorkspaceCreationResult, PrebuiltWorkspaceContext, CommitContext, PrebuiltWorkspace, PermissionName, WorkspaceInstance, EduEmailDomain, ProviderRepository, PrebuildInfo } from "@gitpod/gitpod-protocol";
+import { GitpodServer, GitpodClient, AdminGetListRequest, User, AdminGetListResult, Permission, AdminBlockUserRequest, AdminModifyRoleOrPermissionRequest, RoleOrPermission, AdminModifyPermanentWorkspaceFeatureFlagRequest, UserFeatureSettings, AdminGetWorkspacesRequest, WorkspaceAndInstance, GetWorkspaceTimeoutResult, WorkspaceTimeoutDuration, WorkspaceTimeoutValues, SetWorkspaceTimeoutResult, WorkspaceContext, CreateWorkspaceMode, WorkspaceCreationResult, PrebuiltWorkspaceContext, CommitContext, PrebuiltWorkspace, PermissionName, WorkspaceInstance, EduEmailDomain, ProviderRepository, PrebuildInfo, Queue } from "@gitpod/gitpod-protocol";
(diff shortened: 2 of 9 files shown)


geropl:
Reviewing now.

components/dashboard/src/settings/Teams.tsx
```diff
@@ -500,7 +500,7 @@ function AllTeams() {
                             <span className="text-sm my-auto text-gray-400 truncate overflow-ellipsis">Purchased on {formatDate(sub?.startDate)}</span>
                         </div>
                         <div className="p-0 my-auto flex flex-col w-2/12">
-                            <span className="my-auto truncate text-gray-500 overflow-ellipsis">{slots.filter(s => s.teamSubscription.id === sub.id).length || "–"}</span>
+                            <span className="my-auto truncate text-gray-500 overflow-ellipsis">{slots.filter(s => s.state !== 'cancelled' && s.teamSubscription.id === sub.id).length || "–"}</span>
```
geropl, line 503: :+1:

geropl:
The frontend changes look good and make sense to me.

On the backend, the current change would improve things, allthough not completely solve it:

Good case:
 1. the user deactivates a slot
 2. Chargebee is updated
 3. we receive an Chargebee event _immediately_, and update `quantity`
 4. the user deactives another slot shortly after: :heavy_check_mark:

Bad case:
 1. the user deactivates a slot
 2. Chargebee is updated
 3. we receive an Chargebee event after _some minutes_ (happens quite often), and update `quantity`
 4. the user deactives another slot shortly after, based on the old `quantity`: :x: 
 
This is basically the same problem as before, just more unlikely. Unfortunately I don't see any other way to solve this. The alternative I thought about:
With the suggested change `end of term: false`, we could try to not use `quantity` at all, but fetch the current number from the API: This still has the conceptual TOCTOU problem, as Chargebee does not allow to atomically check `resource_version` on updates, for instance, but is another step to minimize the gap.

geropl:
@jankeromnes AFter having slept over it the explanation is weird. Would you have some time this morning to quickly discuss in sync?

geropl:
Verified my suspicion: The number of seats is _already_ updated immediately*: Shortly after the "change # of seats" action in Gitpod, we get back the Chargebee event ([example for `ed2aac6c-0635-4ca8-96a2-fb38a8c638cf`](https://gitpod.chargebee.com/events/ev_AzZdEUScQmMXT2RVT/details)).

So the lost update still happens, but only if a user clicks faster than Chargebee sends us updates.
Possible ways to improve this:
 1. directly query quantity from Chargebee API:
   - get rid of `ts.quantity` in our DB
   - request quantity from Chargebee API before sending a request with a manipulated quantity

   Pros:
    - reduces likelihood of lost updates
   Cons:
    - lost updates might still happen as the Chargebee API does not allow to conditionally-update a specific version of an object (like "update only if `resource_version` is equal)

 2. list slots that are "active" instead of using `ts.quantity`
   - request list of quantity from active slots
   - compare that to `ts.quantity` and issue a warning if there is a discrepancy
   
   Pros:
   - we break the cycle gitpod -> CB -> gitpod
   Cons:
   - not sure there is any

jankeromnes:
Post-sync summary:
- It makes sense to tell Chargebee to apply these changes immediately, not at the end of term
- Using `ts.quantity` is a bad idea (because it only changes at the end of term, or after a Chargebee callback if we make the effect "immediate", but there is still some gap in which we might loose updates)
- Instead of using `ts.quantity`, it seems safer to take the base value of activations/deactivations directly from our DB (i.e. don't use whatever Chargebee thinks is correct as a base, but use whatever we show users in their Manage Team list as a base -- that's the best source of truth we have)

Will update the PR accordingly.

jankeromnes:
I think this change is correct, and it builds successfully. Unfortunately, I can't test it:
- Hooking up chargebee-test with my `core-dev` environment doesn't work (maybe payments are disabled?)
- Trying to deploy a `io-dev` environment (by creating a new branch under gitpod-com) doesn't work ([all I get](https://werft.gitpod-io-dev.com/job/gitpod-build-jx-fix-teams.2) is "prepare" and "job config" and then Werft calls it done 🤔)

@geropl and/or @AlexTugarev, could you please review, test, and merge this change if it fixes the problem?

### How to test

1. Purchase a Team Plan with 10 seats
2. Deactivate 3 seats in quick succession
3. Teams page should still say "10 Members" (a month later it should say "7 Members" -- but that's tricky to test so don't bother)
4. Chargebee itself should say that the Team Plan has 7 Seats (and the next bill should be 7 x (price of seat), not 10x or 9x)

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -1120,6 +1120,12 @@ export class GitpodServerEEImpl extends GitpodServerImpl<GitpodClient, GitpodSer
         return slots[0];
     }
 
+    // Get the current number of "active" slots in a team subscription (count all "assigned" and "unassigned", but not "deactivated" or "cancelled").
+    protected async tsGetActiveSlotQuantity(teamSubscriptionId: string): Promise<number> {
+        const slots = await this.teamSubscriptionDB.findSlotsByTeamSubscriptionId(teamSubscriptionId);
+        return slots.filter(slot => !!slot.cancellationDate).length;
```
AlexTugarev, line 1126: @jankeromnes just to make sure I understand this correctly...
* "active" means **no cancellation date** is set
* this filter returns slots that actually have the cancellation date set

jankeromnes: Oh, indeed 🤦 excellent catch, thanks!!

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -1120,6 +1120,12 @@ export class GitpodServerEEImpl extends GitpodServerImpl<GitpodClient, GitpodSer
         return slots[0];
     }
 
+    // Get the current number of "active" slots in a team subscription (count all "assigned" and "unassigned", but not "deactivated" or "cancelled").
+    protected async tsGetActiveSlotQuantity(teamSubscriptionId: string): Promise<number> {
+        const slots = await this.teamSubscriptionDB.findSlotsByTeamSubscriptionId(teamSubscriptionId);
+        return slots.filter(slot => !slot.cancellationDate).length;
```
geropl, line 1126: Would love to see an existing method re-used for this.

AlexTugarev:
@geropl, I think we are missing `https://jx-fix-teams.staging.gitpod-dev.com/` for callbacks in the CB, because events are missing in server.

geropl:
Tested, this seems to fix it:

## UI
 - [x] correct # of seats (now: 10)
image


## Chargbee subscription:
 - [x] correct # of seats (billed for next cycle: 5)
image

## Chargebee event:
 - [x] is sent immediately
 - [x] contains correct # of seats (billed for next cycle: 5)
image

## DB:
- [x] correct nr of seats (5 and 10, resp.)
image

geropl (approved):
:+1: (details see above)

geropl (changes requested):
@AlexTugarev found that when you're fast enough, you can triger races in `server` between DB and multiple "deactivate slots" requests.

This is a more fundamental issue with our current approach. In the scope of this PR we'll try to:
 - introduce a queue to avoid this at least at the scope of a single server instance
 - think about a regular task that pushes the current state of the DB to Chargebee

AlexTugarev (approved):
Tested latest changes as described in https://github.com/gitpod-io/gitpod/pull/4713#issuecomment-875800173

It worked properly. 👍🏻

geropl (approved):
Fixed with https://github.com/gitpod-io/gitpod/pull/4713/commits/01e7ec09573b6c6a68d33f140a85efa896e7ef71 :tada:

PR #4725, opened 2021-07-07
[werft] Disable dynamic ws-manager cert generation

geropl:
This PR does two things:
 - reorders deploy stepts so we don't have to use an explicit promise in the build job
 - ties down all dependencies for the build job

The former seems* to have fixed our issue (*because we're not sure what caused it in the first place). The later is added because
 1. it's a good thing to have anyway
 2. we have a handle on the dependencies and can easily alter them in case we find an offending one


.werft/build.ts
```diff
@@ -311,23 +286,26 @@ export async function deployToDev(deploymentConfig: DeploymentConfig, workspaceF
...
+        await installMetaCertificates();
+
+        if (k3sWsCluster) {
+            await installWsCertificates();
         }
+        werft.done('certificate');
+
+        werft.done('prep');
+    } catch (err) {
+        werft.fail('prep', err);
```
princerachit, line 308: nit: What if the certificate phase errors out?

princerachit (approved):
Please merge after build

Issue #3303, opened 2021-03-01
[prebuilds] Remove prebuild labels


csweichel:
When we first introduced prebuilds we added functionality so that the bot could add comments instead of setting the commit status. This feature doesn't make much sense and is broken (see https://github.com/gitpod-io/gitpod/issues/3235). We should remove it.

jankeromnes:
The `addComment` feature is now fixed, and several users have expressed interest in continuing to use it.

Also, a PR comment is a nice, obvious way to encourage contributors and reviewers to use Gitpod, while a PR "Check" is very discreet and much easier to ignore or overlook (especially when all "Checks" are green and thus collapsed / hidden from view). Additionally, other popular tools/services also have a comment feature available (e.g. Netlify, Vercel, etc).

I would vote for:
- `addCheck`: keep ✅ 
- `addComment`: keep ✅ 
- `addBadge`: keep ✅ 
- `addLabel`: deprecate and remove ❌ (it's still broken and nobody has complained about it, as opposed to comments, so `addLabel` is probably the least useful feature here)

PR #4727, opened 2021-07-07
Link prebuilds with projects

AlexTugarev:
With this PR push events will trigger prebuilds and associate them with a project (`PrebuiltWorkspace.projectId`) when the App installation is found. 

This also includes a deduplication fix for the drop down on New Project: 

Screen Shot 2021-07-07 at 11 21 21

### DIFF
--- a/components/gitpod-db/src/typeorm/project-db-impl.ts
+++ b/components/gitpod-db/src/typeorm/project-db-impl.ts
@@ -43,4 +43,10 @@ export class ProjectDBImpl implements ProjectDB {
         await repo.save(project);
         return project;
     }
+
+    async findProjectByInstallationId(appInstallationId: string): Promise<Project | undefined> {
+        const repo = await this.getRepo();
+        const project = await repo.findOne({ appInstallationId });
+        return project;
+    }
 }
--- a/components/gitpod-db/src/typeorm/workspace-db-impl.ts
+++ b/components/gitpod-db/src/typeorm/workspace-db-impl.ts
@@ -6,7 +6,7 @@
 
 import { injectable, inject } from "inversify";
 import { Repository, EntityManager, DeepPartial, UpdateQueryBuilder } from "typeorm";
-import { AbstractWorkspaceDB, MaybeWorkspace, MaybeWorkspaceInstance, WorkspaceDB, FindWorkspacesOptions, PrebuiltUpdatableAndWorkspace, WorkspaceInstanceSessionWithWorkspace, PrebuildWithWorkspace, WorkspaceAndOwner, WorkspacePortsAuthData, WorkspaceOwnerAndSoftDeleted } from "../workspace-db";
+import { MaybeWorkspace, MaybeWorkspaceInstance, WorkspaceDB, FindWorkspacesOptions, PrebuiltUpdatableAndWorkspace, WorkspaceInstanceSessionWithWorkspace, PrebuildWithWorkspace, WorkspaceAndOwner, WorkspacePortsAuthData, WorkspaceOwnerAndSoftDeleted } from "../workspace-db";
 import { Workspace, WorkspaceInstance, WorkspaceInfo, WorkspaceInstanceUser, WhitelistedRepository, Snapshot, LayoutData, PrebuiltWorkspace, RunningWorkspaceInfo, PrebuiltWorkspaceUpdatable, WorkspaceAndInstance, WorkspaceType } from "@gitpod/gitpod-protocol";
 import { TypeORM } from "./typeorm";
 import { DBWorkspace } from "./entity/db-workspace";
@@ -27,11 +27,9 @@ interface OrderBy {
 }
 
 @injectable()
-export abstract class AbstractTypeORMWorkspaceDBImpl extends AbstractWorkspaceDB {
+export abstract class AbstractTypeORMWorkspaceDBImpl implements WorkspaceDB {
 
     protected abstract getManager(): Promise<EntityManager>;
-
-    abstract transaction<T>(code: (db: WorkspaceDB) => Promise<T>): Promise<T>;
 
     protected async getWorkspaceRepo(): Promise<Repository<DBWorkspace>> {
         return await (await this.getManager()).getRepository<DBWorkspace>(DBWorkspace);
@@ -78,6 +76,23 @@ export abstract class AbstractTypeORMWorkspaceDBImpl extends AbstractWorkspaceDB
             tries++;
         }
         throw new Error("Could not establish connection to database!");
+    }
+
+    public async transaction<T>(code: (db: WorkspaceDB) => Promise<T>): Promise<T> {
+        return code(this);
+    }
+
+    async storeInstance(instance: WorkspaceInstance): Promise<WorkspaceInstance> {
+        const inst = await this.internalStoreInstance(instance);
+        return inst;
+    }
+
+    public async findRunningInstance(workspaceId: string): Promise<MaybeWorkspaceInstance> {
+        const instance = await this.findCurrentInstance(workspaceId)
+        if (instance && instance.status.phase !== 'stopped') {
+            return instance;
+        }
+        return undefined;
     }
 
     public async store(workspace: Workspace) {
@@ -796,6 +811,18 @@ export abstract class AbstractTypeORMWorkspaceDBImpl extends AbstractWorkspaceDB
         delete res["creationTime"];
 
         return <WorkspaceAndInstance>(res);
+    }
+
+    async findPrebuiltWorkspacesByProject(projectId: string): Promise<PrebuiltWorkspace[]> {
+        const repo = await this.getPrebuiltWorkspaceRepo();
+
+        const query = repo.createQueryBuilder('pws')
+            .orderBy('pws.creationTime', 'ASC')
+            .innerJoinAndMapOne('pws.workspace', DBWorkspace, 'ws', 'pws.buildWorkspaceId = ws.id')
+            .andWhere('pws.projectId = :projectId', { projectId })
+
+        const res = await query.getMany();
+        return res;
     }
 
 }
--- a/components/server/ee/src/prebuilds/github-app.ts
+++ b/components/server/ee/src/prebuilds/github-app.ts
@@ -6,18 +6,18 @@
 
 import { Server, Probot, Context } from 'probot';
 import { getPrivateKey } from '@probot/get-private-key';
+import {WebhookEvent, EventPayloads} from "@octokit/webhooks/dist-types"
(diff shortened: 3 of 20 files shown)


components/server/ee/src/prebuilds/github-app.ts
```diff
@@ -280,7 +288,7 @@ export class GithubApp {
         updatePrPromise.catch(err => log.error(err, "Error while updating PR body", { contextURL }));
     }
 
-    protected onPrAddLabel(config: WorkspaceConfig | undefined, user: User, ctx: Context, prebuildStartPromise: Promise<StartPrebuildResult> | undefined) {
+    protected onPrAddLabel(config: WorkspaceConfig | undefined, ctx: Context, prebuildStartPromise: Promise<StartPrebuildResult> | undefined) {
```
jankeromnes, line 291: The `addLabel` feature doesn't work anymore and should likely be removed: https://github.com/gitpod-io/gitpod/issues/3303 (from the code and the docs).

components/gitpod-protocol/src/gitpod-service.ts
```diff
@@ -285,6 +285,8 @@ export namespace GitpodServer {
     }
     export interface CreateWorkspaceOptions {
         contextUrl: string;
+        branch?: string;
+        projectId?: string;
```
jankeromnes, line 289: Leftover? To me this doesn't make sense, and is not actually used anywhere in this PR -- remove for now?

```suggestion
```

jankeromnes (commented):
Awesome, many thanks for implementing this! 💯

Here a first quick pass. Will do a deeper review soon.

components/server/ee/src/prebuilds/github-app.ts
AlexTugarev: Awesome! Will tackle this.

components/gitpod-protocol/src/gitpod-service.ts
AlexTugarev: Indeed.

components/server/ee/src/prebuilds/github-app.ts
AlexTugarev: done

components/gitpod-protocol/src/gitpod-service.ts
AlexTugarev: done

components/server/ee/src/workspace/workspace-factory.ts
```diff
@@ -47,6 +46,8 @@ export class WorkspaceFactoryEE extends WorkspaceFactory {
                 throw new Error("Can only prebuild workspaces with a commit context")
             }
 
+            const { project, branch } = context;
```
jankeromnes, line 49: Do you expect to use `project` or `branch` somewhere else in the code below?

Otherwise, you could avoid the single-use local variables by using `context.project` and `context.branch` directly in the `storePrebuiltWorkspace()` call below.

components/gitpod-db/src/typeorm/workspace-db-impl.ts
```diff
@@ -798,6 +813,18 @@ export abstract class AbstractTypeORMWorkspaceDBImpl extends AbstractWorkspaceDB
         return <WorkspaceAndInstance>(res);
     }
 
+    async findPrebuiltWorkspacesByProject(projectId: string): Promise<PrebuiltWorkspace[]> {
+        const repo = await this.getPrebuiltWorkspaceRepo();
+
+        let query = repo.createQueryBuilder('pws');
+        query = query.orderBy('pws.creationTime', 'ASC');
+        query = query.innerJoinAndMapOne('pws.workspace', DBWorkspace, 'ws', 'pws.buildWorkspaceId = ws.id');
+        query = query.andWhere('pws.projectId = :projectId', { projectId })
```
jankeromnes, line 822: Nit: This looks a bit awkward with the mutable variable and frequent re-assignments. Maybe chain like in the other code locations?

```suggestion
        const query = repo.createQueryBuilder('pws')
            .orderBy('pws.creationTime', 'ASC')
            .innerJoinAndMapOne('pws.workspace', DBWorkspace, 'ws', 'pws.buildWorkspaceId = ws.id')
            .andWhere('pws.projectId = :projectId', { projectId });
```

jankeromnes (commented):
Many thanks @AlexTugarev! The code looks good to me.

How should this change be tested? (In particular, prebuilds in relation to the GitHub/GitLab/Bitbucket apps/integrations.) I've briefly tested that the `/#prebuild/` manual prefix still works, but I'd love to play with a proper App/Integration in order to confirm that everything (still) works as expected.

components/server/ee/src/prebuilds/github-app.ts
```diff
@@ -341,8 +309,24 @@ export class GithubApp {
...
 
-    protected async findUserForInstallation(ctx: Context): Promise<User | undefined> {
-        const installation = await this.appInstallationDB.findInstallation("github", ctx.payload.installation.id);
+    protected async findInstallationOwner(installationId: number): Promise<{user: User, project?: Project} | undefined> {
+
+        // Project mode
+        //
+        const project = await this.projectDB.findProjectByInstallationId(String(installationId));
+        if (project) {
+            const owner = (await this.teamDB.findMembersByTeam(project.teamId)).filter(m => m.role === "owner")[0];
```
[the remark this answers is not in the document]
AlexTugarev: > It could very well be a different user than whoever installed the app for a given project in that team.

This is a fair point. We haven't discussed the differentiation between installer and owner. I guess for now it's good enough, as there is no accounting for this.

components/server/ee/src/prebuilds/github-app.ts
```diff
@@ -341,8 +309,24 @@ export class GithubApp {
...
+                const user = await this.userDB.findUserById(owner.userId);
+                if (user) {
+                    return { user, project}
+                }
+            }
+        }
+
+        // Legacy mode
+        //
+        const installation = await this.appInstallationDB.findInstallation("github", String(installationId));
```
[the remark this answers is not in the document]
AlexTugarev: > Do we really want to trust projectDB and teamDB to implicitly guarantee the existence / validity of a given installationId?

you are absolutely right about the "uninstalled" event. It should also either clean up or disable projects. 
putting this on TODO list.

what else would you expect?

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -1442,25 +1442,27 @@ export class GitpodServerEEImpl extends GitpodServerImpl<GitpodClient, GitpodSer
 
     public async getPrebuilds(teamId: string, projectName: string): Promise<PrebuildInfo[]> {
         this.checkAndBlockUser("getPrebuilds");
+        const result: PrebuildInfo[] = [];
```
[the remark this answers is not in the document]
AlexTugarev: thanks! 'copied will be added in next turn.

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -1442,25 +1442,27 @@ export class GitpodServerEEImpl extends GitpodServerImpl<GitpodClient, GitpodSer
...
-                    id: pws.prebuild.id,
-                    startedAt: pws.prebuild.creationTime,
-                    startedBy: "Owner",
+            const pwss = await this.workspaceDb.trace({}).findPrebuiltWorkspacesByProject(project.id);
+
+            for (const pws of pwss) {
+                result.push({
+                    id: pws.id,
+                    startedAt: pws.creationTime,
+                    startedBy: "UNKNOWN",
```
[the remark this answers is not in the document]
AlexTugarev: let's consider this WIP, it's being picked up in frontend to show how triggered this prebuild. starting with the designs and seeing that the API is missing this info. now working on fixing this in parallel.

components/gitpod-db/src/typeorm/project-db-impl.ts
```diff
@@ -43,4 +43,10 @@ export class ProjectDBImpl implements ProjectDB {
         await repo.save(project);
         return project;
     }
+
+    async findProjectByInstallationId(appInstallationId: string): Promise<Project | undefined> {
+        const repo = await this.getRepo();
+        const project = await repo.findOne({ appInstallationId });
+        return project;
```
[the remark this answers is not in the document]
AlexTugarev: yes, but that makes it hard to set breakpoints

jankeromnes (approved):
After testing this some more, with an installed GitHub App, I'm more confident that this works as expected. Thus, approving. Thanks again!

However, could you please wait and only merge this after tomorrow's production deployment? This would allow more time to test on staging as well.

EDIT: Also, please squash all related commits together.

AlexTugarev:
> However, could you please wait and only merge this after tomorrow's production deployment? This would allow more time to test on staging as well.

@jankeromnes, thank you very much! I'll do so.

components/gitpod-db/src/typeorm/project-db-impl.ts
```diff
@@ -43,4 +43,10 @@ export class ProjectDBImpl implements ProjectDB {
         await repo.save(project);
         return project;
     }
+
+    async findProjectByInstallationId(appInstallationId: string): Promise<Project | undefined> {
```
jankeromnes, line 47: Maybe make this `public`?

```suggestion
    public async findProjectByInstallationId(appInstallationId: string): Promise<Project | undefined> {
```

AlexTugarev: it's public by [default](https://www.typescriptlang.org/docs/handbook/classes.html#public-by-default).

components/server/ee/src/prebuilds/github-app.ts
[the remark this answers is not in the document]
jankeromnes: > what else would you expect?

For the sanity check "does this installation ID actually correspond to anything?" to be above both "modes" (not just performed in "legacy" mode as currently). You can simply move the entire `const installation = ...` and `if (!installation)` block up several lines.

components/gitpod-db/src/typeorm/project-db-impl.ts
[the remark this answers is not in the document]
jankeromnes: Do you frequently set breakpoints immediately after `repo` method calls in DB implementations? 🤔

components/gitpod-db/src/typeorm/project-db-impl.ts
jankeromnes: Most other code locations have `public` set explicitly, so this seems inconsistent. Not very important though, so resolved for now.

PR #4729, opened 2021-07-07
[werft] Make cert creation & helm installation sequential

princerachit:
Werft deployments are broken due to some bug in one of the dependencies. Since we [install dependencies](https://github.com/gitpod-io/gitpod/blob/main/.werft/wipe-devstaging.yaml#L38) without pinning them one of the transitive dependencies could be an issue.

We see that await keyword is not returning the promise as expected, therefore, for now we are installing gitpod  sequentially followed by certificate installation.


geropl (approved):
:+1:

PR #4743, opened 2021-07-08
[code] confirm sharing

### LINKED ISSUE #4726: Confirmation popup for workspace sharing
This feature is particularly sensitive to leaking workspace data and the user needs to be extremely aware that they are enabling it when they do.

A possible solution is to add an "Are you sure?" popup that explains the implications for workspace sharing so that the user can have a chance to reflect and decide before they do the action.

This was already present in the Theia editor, we need it in Code too now.

akosyakov:
#### What it does 

- fix #4726 

Screenshot 2021-07-08 at 14 29 14

Gitpod Code Change: https://github.com/gitpod-io/vscode/commit/fde107c3bf415f31560a050d2bde356d8e7602fd

#### How to test

Start a workspace and try to share.


fntlnz (approved):

fntlnz:
Thank you @akosyakov !

PR #4746, opened 2021-07-08
[werft] Add sweeper cleanup logic for k3s ws cluster

princerachit:
# What?

When running wipe job do the following:
- determine is k3s-external.yaml exists, if it does then delete this preview env with the same logic as that of dev. And also delete the external IP that was created by werft for the ws-proxy in this cluster. Otherwise go to next step
- Delete preview env in the dev cluster

# Testing
To test this out I manually edited the sweeper deploy and added arg`--timeout=5m` to trigger wipe job instantly. The cleanup succeeded

[Reference job in a different branch](https://werft.gitpod-dev.com/job/gitpod-wipe-devstaging-prs-sweeper-n-meta-ws-disable.0/raw)

 - [ ] /werft k3s-ws

### DIFF
--- a/.werft/wipe-devstaging.ts
+++ b/.werft/wipe-devstaging.ts
@@ -1,13 +1,15 @@
-import { werft } from './util/shell';
+import { werft, exec } from './util/shell';
 import { wipePreviewEnvironment, listAllPreviewNamespaces } from './util/kubectl';
+import  * as fs from 'fs';
+import { sleep } from './util/util';
 
 
-async function wipeDevstaging() {
+async function wipeDevstaging(pathToKubeConfig: string) {
     const namespace_raw = process.env.NAMESPACE;
     const namespaces: string[] = [];
     if (namespace_raw === "<no value>" || !namespace_raw) {
         werft.log('wipe', "Going to wipe all namespaces");
-        listAllPreviewNamespaces("")
+        listAllPreviewNamespaces(pathToKubeConfig)
             .map(ns => namespaces.push(ns));
     } else {
         werft.log('wipe', `Going to wipe namespace ${namespace_raw}`);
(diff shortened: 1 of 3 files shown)


.werft/wipe-devstaging.ts
```diff
@@ -1,23 +1,73 @@
-import { werft } from './util/shell';
+import { werft, exec } from './util/shell';
 import { wipePreviewEnvironment, listAllPreviewNamespaces } from './util/kubectl';
+import  * as fs from 'fs';
+import { sleep } from './util/util';
 
 
-async function wipeDevstaging() {
+async function wipeDevstaging(pathToKubeConfig: string) {
```
geropl, line 7: I won't find time for a full review today, but: please let's try to find another way to switch between kubectl contexts. This really decreases the signal to noise ratio a lot.

I have not coherent view on this or what might be better, but will try to come up with sth tomorrow.

princerachit: One way I can think of is to merge all the kubeconfigs we have into the default one and then use context switch. That would involve downloading a kubeconfig -> updating the context name in that and then adding it to the default kubeconfig.

Since pathToKubeConfig is how I have done for all other places I would let this PR merge and raise a separate PR to use a new approach for all the files.

geropl: > Since pathToKubeConfig is how I have done for all other places I would let this PR merge and raise a separate PR to use a new approach for all the files.

Ok, don't want to block this PR as it fixes sweeper. But let's not forget this.

> One way I can think of is to merge all the kubeconfigs we have into the default one and then use context switch.

This was the way we used to do it in the past albeit it was simpler back then because clusters were static. But we can also leave the separate kubectl configs, but pass a more generic `env` around which `exec` supports (we already have that in some places).

.werft/wipe-devstaging.ts
```diff
@@ -1,23 +1,73 @@
...
     for (const namespace of namespaces) {
-        await wipePreviewEnvironment("", "gitpod", namespace, { slice: 'wipe' });
+        await wipePreviewEnvironment(pathToKubeConfig, "gitpod", namespace, { slice: 'wipe' });
     }
-    werft.done('wipe');
 }
 
-wipeDevstaging()
\ No newline at end of file
+async function deleteExternalIp(k3sWsProxyIP: string, namespace: string) {
```
geropl, line 24: Note: In general it would be good to find a common place for code like this, like a `k3s-cluster.ts` or `deployment.ts` or so. `wipe-staging.ts` should be just calling that stuff.

princerachit: makes sense. Let me move this to another file.

.werft/wipe-devstaging.ts
princerachit: >This was the way we used to do it in the past albeit it was simpler back then because clusters were static. But we can also leave the separate kubectl configs, but pass a more generic env around which exec supports (we already have that in some places).

This seems like a much better solution! I will try this out and raise a PR

geropl (approved):
LGTM!

Issue #4333, opened 2021-05-28
Prebuilts not triggered from Gitlab on branches


j0nes2k:
### Bug description

Gitpod user since some months here. I am hosting my private repository on gitlab.com, and I am experiencing issues that prebuilts for my branches are not running. From what I can see in gitlab, the corresponding webhook is triggered and the input is looking fine.

My .gitpod.yml follows a basic setup:

```
tasks:
  - init: |
      some_prebuild_task.sh
    command: |
      some_frontend_task.sh
```

On my master branch a prebuild is available, however the latest changes to master are already several days old.

### Steps to reproduce

Log file from Gitlab Webhook here:

Request headers:
```
Content-Type: application/json
User-Agent: GitLab/14.0.0-pre
X-Gitlab-Event: Push Hook
X-Gitlab-Token: my-gitlab-token
```

Request Body:
```
{
  "object_kind": "push",
  "event_name": "push",
  "before": "myoldsha",
  "after": "mynewsha",
  "ref": "refs/heads/feature/fix_pip_dependencies",
  "checkout_sha": "dfea055400435d957f1a1bcdd77283cdce48b2ac",
  "message": null,
  "user_id": 98765,
  "user_name": "Jonas Kaufmann",
  "user_username": "j0nes2k",
  "user_email": "",
  "user_avatar": "https://gitlab.com/uploads/-/system/user/avatar/747341/avatar.png",
  "project_id": 12345,
  "project": {
    "id": 12345,
    "name": "awesomecompany",
    "description": "",
    "web_url": "https://gitlab.com/awesomecompany-dev/awesomecompany",
    "avatar_url": null,
    "git_ssh_url": "git@gitlab.com:awesomecompany-dev/awesomecompany.git",
    "git_http_url": "https://gitlab.com/awesomecompany-dev/awesomecompany.git",
    "namespace": "awesomecompany-Dev",
    "visibility_level": 0,
    "path_with_namespace": "awesomecompany-dev/awesomecompany",
    "default_branch": "master",
    "ci_config_path": "",
    "homepage": "https://gitlab.com/awesomecompany-dev/awesomecompany",
    "url": "git@gitlab.com:awesomecompany-dev/awesomecompany.git",
    "ssh_url": "git@gitlab.com:awesomecompany-dev/awesomecompany.git",
    "http_url": "https://gitlab.com/awesomecompany-dev/awesomecompany.git"
  },
  "commits": [
    {
      "id": "mysha",
[...]

PR #4747, opened 2021-07-08
[server] Improve logging / error messages of GitLab app and context parser

corneliusludmann:
Improves the logging resp. error messages in the GitLab app resp. context parser. This hopefully gives us more insights into issue https://github.com/gitpod-io/gitpod/issues/4333.

### DIFF
--- a/components/server/src/gitlab/gitlab-context-parser.ts
+++ b/components/server/src/gitlab/gitlab-context-parser.ts
@@ -214,14 +214,15 @@ export class GitlabContextParser extends AbstractContextParser implements IConte
         const possibleBranches = await this.gitlabApi.run<GitLab.Branch[]>(user, async g => {
             return g.Branches.all(`${owner}/${repoName}`, { search });
         });
-        if (!GitLab.ApiError.is(possibleBranches)) {
-            for (let candidate of possibleBranches) {
-                const name = candidate.name;
-                const nameSegements = candidate.name.split('/');
-                if (segments.length >= nameSegements.length && segments.slice(0, nameSegements.length).join('/') === name) {
-                    branchOrTagObject = { type: 'branch', name, revision: candidate.commit.id };
-                    break;
-                }
+        if (GitLab.ApiError.is(possibleBranches)) {
+            throw new Error(`GitLab ApiError on searching for possible branches for ${owner}/${repoName}/tree/${segments.join('/')}: ${possibleBranches}`);
+        }
+        for (let candidate of possibleBranches) {
+            const name = candidate.name;
+            const nameSegements = candidate.name.split('/');
+            if (segments.length >= nameSegements.length && segments.slice(0, nameSegements.length).join('/') === name) {
+                branchOrTagObject = { type: 'branch', name, revision: candidate.commit.id };
+                break;
             }
         }
 
@@ -229,14 +230,15 @@ export class GitlabContextParser extends AbstractContextParser implements IConte
             const possibleTags = await this.gitlabApi.run<GitLab.Tag[]>(user, async g => {
                 return g.Tags.all(`${owner}/${repoName}`, { search });
             });
-            if (!GitLab.ApiError.is(possibleTags)) {
-                for (let candidate of possibleTags) {
-                    const name = candidate.name;
-                    const nameSegements = candidate.name.split('/');
-                    if (segments.length >= nameSegements.length && segments.slice(0, nameSegements.length).join('/') === name) {
-                        branchOrTagObject = { type: 'tag', name, revision: candidate.commit.id };
-                        break;
-                    }
+            if (GitLab.ApiError.is(possibleTags)) {
+                throw new Error(`GitLab ApiError on searching for possible tags for ${owner}/${repoName}/tree/${segments.join('/')}: ${possibleTags}`);
+            }
+            for (let candidate of possibleTags) {
+                const name = candidate.name;
+                const nameSegements = candidate.name.split('/');
+                if (segments.length >= nameSegements.length && segments.slice(0, nameSegements.length).join('/') === name) {
+                    branchOrTagObject = { type: 'tag', name, revision: candidate.commit.id };
+                    break;
                 }
             }
         }
--- a/components/server/src/gitlab/gitlab-context-parser.ts
+++ b/components/server/src/gitlab/gitlab-context-parser.ts
@@ -170,80 +170,97 @@ export class GitlabContextParser extends AbstractContextParser implements IConte
     // https://gitlab.com/AlexTugarev/gp-test/blob/wip/folder/empty.file.jpeg
     protected async handleTreeContext(user: User, host: string, owner: string, repoName: string, segments: string[]): Promise<NavigatorContext> {
 
-        const branchOrTagPromise = segments.length > 0 ? this.getBranchOrTag(user, owner, repoName, segments) : undefined;
-        const repository = await this.fetchRepo(user, `${owner}/${repoName}`);
-        const branchOrTag = await branchOrTagPromise;
-        const context = <NavigatorContext>{
-            isFile: false,
-            path: '',
-            title: `${owner}/${repoName}` + (branchOrTag ? ` - ${branchOrTag.name}` : ''),
-            ref: branchOrTag && branchOrTag.name,
-            revision: branchOrTag && branchOrTag.revision,
-            refType: branchOrTag && branchOrTag.type,
-            repository
-        };
-        if (!branchOrTag) {
-            return context;
-        }
-        if (segments.length === 1 || branchOrTag.fullPath.length === 0) {
-            return context;
-        }
-
-        const result = await this.gitlabApi.run<GitLab.TreeObject[]>(user, async g => {
-            return g.Repositories.tree(`${owner}/${repoName}`, { ref: branchOrTag.name, path: path.dirname(branchOrTag.fullPath) });
-        });
-        if (GitLab.ApiError.is(result)) {
-            log.debug(`Error reading TREE ${owner}/${repoName}/tree/${segments.join('/')}.`, result);
-        } else {
-            const object = result.find(o => o.path === branchOrTag.fullPath);
-            if (object) {
-                const isFile = object.type === "blob";
-                context.isFile = isFile;
-                context.path = branchOrTag.fullPath;
+        try {
+            const branchOrTagPromise = segments.length > 0 ? this.getBranchOrTag(user, owner, repoName, segments) : undefined;
+            const repository = await this.fetchRepo(user, `${owner}/${repoName}`);
+            const branchOrTag = await branchOrTagPromise;
+            const context = <NavigatorContext>{
+                isFile: false,
+                path: '',
+                title: `${owner}/${repoName}` + (branchOrTag ? ` - ${branchOrTag.name}` : ''),
+                ref: branchOrTag && branchOrTag.name,
+                revision: branchOrTag && branchOrTag.revision,
+                refType: branchOrTag && branchOrTag.type,
+                repository
+            };
+            if (!branchOrTag) {
+                return context;
             }
+            if (segments.length === 1 || branchOrTag.fullPath.length === 0) {
+                return context;
+            }
+
+            const result = await this.gitlabApi.run<GitLab.TreeObject[]>(user, async g => {
+                return g.Repositories.tree(`${owner}/${repoName}`, { ref: branchOrTag.name, path: path.dirname(branchOrTag.fullPath) });
+            });
+            if (GitLab.ApiError.is(result)) {
+                throw new Error(`Error reading TREE ${owner}/${repoName}/tree/${segments.join('/')}: ${result}`);
+            } else {
+                const object = result.find(o => o.path === branchOrTag.fullPath);
+                if (object) {
+                    const isFile = object.type === "blob";
+                    context.isFile = isFile;
+                    context.path = branchOrTag.fullPath;
+                }
+            }
+            return context;
+        } catch (e) {
+            log.debug("GitLab context parser: Error handle tree context.", e);
+            throw e;
         }
-        return context;
     }
 
     protected async getBranchOrTag(user: User, owner: string, repoName: string, segments: string[]): Promise<{ type: RefType, name: string, revision: string, fullPath: string }> {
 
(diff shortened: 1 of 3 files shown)


components/server/src/gitlab/gitlab-context-parser.ts
```diff
@@ -214,29 +214,31 @@ export class GitlabContextParser extends AbstractContextParser implements IConte
...
         });
-        if (!GitLab.ApiError.is(possibleBranches)) {
-            for (let candidate of possibleBranches) {
-                const name = candidate.name;
-                const nameSegements = candidate.name.split('/');
-                if (segments.length >= nameSegements.length && segments.slice(0, nameSegements.length).join('/') === name) {
-                    branchOrTagObject = { type: 'branch', name, revision: candidate.commit.id };
-                    break;
-                }
+        if (GitLab.ApiError.is(possibleBranches)) {
```
AlexTugarev, line 217: this is a quite interesting point to debug log the error

components/server/src/gitlab/gitlab-context-parser.ts
```diff
@@ -214,29 +214,31 @@ export class GitlabContextParser extends AbstractContextParser implements IConte
...
-            if (!GitLab.ApiError.is(possibleTags)) {
-                for (let candidate of possibleTags) {
-                    const name = candidate.name;
-                    const nameSegements = candidate.name.split('/');
-                    if (segments.length >= nameSegements.length && segments.slice(0, nameSegements.length).join('/') === name) {
-                        branchOrTagObject = { type: 'tag', name, revision: candidate.commit.id };
-                        break;
-                    }
+            if (GitLab.ApiError.is(possibleTags)) {
+                throw new Error(`GitLab ApiError on searching for possible tags for ${owner}/${repoName}/tree/${segments.join('/')}: ${possibleTags}`);
```
AlexTugarev, line 234: same here

components/server/src/gitlab/gitlab-context-parser.ts
```diff
@@ -214,29 +214,31 @@ export class GitlabContextParser extends AbstractContextParser implements IConte
         const possibleBranches = await this.gitlabApi.run<GitLab.Branch[]>(user, async g => {
```
AlexTugarev, line 214: why so complicated? 
there is https://docs.gitlab.com/ee/api/branches.html#get-single-repository-branch

components/server/src/gitlab/gitlab-context-parser.ts
corneliusludmann: What do you think of wrapping this in a try-catch block instead:

https://github.com/gitpod-io/gitpod/blob/e2439d8a7cbfc4174b1997e28a17c59c1181b178/components/server/src/gitlab/gitlab-context-parser.ts#L173-L205

like this:

```typescript
try {
  ...
} catch (e) {
  log.debug("GitLab context parser: Error handle tree context.", e);
  throw e;
}
```

That would catch these errors and add proper logging for all of them in the `getBranchOrTag` function.

Currently, we have something like this in the logs:

image

The try-catch change would replace this with proper logging, right?

components/server/src/gitlab/gitlab-context-parser.ts
corneliusludmann: The problem is, that we don't know what of the segments is part of the branch and what is part of the file tree.

For example this context URL:

https://gitlab.com/gp-test-group/gp-test-subgroup/gp-test-project-in-subgroup/-/tree/corneliusludmann/important-issue-1/test-dir-1

- `corneliusludmann/important-issue-1` is the branch and
-  `test-dir-1` is the file path.

But we cannot know this having just the context URL. It could also be that:
- `corneliusludmann` is the branch and
- `important-issue-1/test-dir-1` is the file path.

We don't know this until we check for all possible branches that start with the first segement.

AlexTugarev: I believe this is the root cause of the issue here. A lucky shot revealed that some/lot of projects are structuring their branches in `feature/9999_something` patterns. 
In L.212 we're searching for `^feature` branches, which in many cases will be a large set of branches. Unfortunately, there is pagination in the GL REST API unconsidered here at: https://docs.gitlab.com/ee/api/#pagination
So, we always see only some of the `feature*` branches (maybe just 20.)

corneliusludmann: That is an interesting thought! Indeed, pagination could be the problem here. :bulb:

AlexTugarev: @corneliusludmann thanks for the explanation. I forgot about this dilemma.
Still, in that case we need a retry strategy with `get-single-repository-branch` calls and path prefix' in in ascending order, i.e. call it for `seg1`, `seg1/seg2`, `seg1/seg2/seg3`, etc. This is still less expensive than filtering `n` search calls.

components/server/src/gitlab/gitlab-context-parser.ts
AlexTugarev: I just mean that `GitLab.ApiError.is(possibleBranches)` contain the error message and the like. That would be of interest to see for debugging.

components/server/src/gitlab/gitlab-context-parser.ts
AlexTugarev: I just verified `search` always returns 20 entries starting from an offset.
```
curl  "https://gitlab.com/api/v4/projects/gitlab-org%2Fgitlab/repository/branches?search=%5Efeature"
```

corneliusludmann: What we can do is:

1. implementing proper pagination or
2. using the `get-single-repository-branch` API call you mentioned above and iterate over the subsets of segments in our code, e.g. given the example context URL above:
  a) check if there is a branch called `corneliusludmann/important-issue-1/test-dir-1`, if not ...
  b) check if there is a branch called `corneliusludmann/important-issue-1` (is true so terminate here), in case if not ...
  c) check if there is a branch called `corneliusludmann`.
  or
3. doing the same as (3) but in reverse order (starting with 1 segment, then 2 segments, ...)

The question is, which approach is the best to limit the number of API calls.

(1) would need many API calls if there are a lot of branches starting with the same segment (as `feature/`). We could increase the `perPage` option but still need to implement pagination handling.

(2) would need many API calls if the context URL has a deep file tree. However, I think that most of our context URLs don't have a file tree part at all.

(3) would need some API calls if the branch has some segments. However, the number of segments for branches should be usually small. But something like a `feature/` prefix is quite common. So, it needs pretty often at least 2 calls.

My gut feeling is that with (2), in most of the cases 1 API call is sufficient. But in the worst case (a user has a pretty long path in the context URL) it takes the most API calls.

Probably a combination would be the best idea: First, try if there is a branch with all segments (which is probably in most cases successful) and continue with (3)  in case there is no match (or even better continue with (1) instead ?).

Do you have an opinion/feeling about this, @AlexTugarev?

AlexTugarev: Given that in git you cannot create "prefix" branches, I'd vote for the ascending number of segments to try with. It should be correct in any reasonable way.
Also, if you want to optimize the stuff, add a special handling for the `feature` case and begin with `seg1/seg2` ;-)

AlexTugarev: (3) 👍🏻

corneliusludmann: Regarding pagination: All info I found about the GitLab library we use, it handles pagination behind the scenes already. Newer versions of this library give more control over pagination, though.

However, that also means, if there are a lot of pages that are needed to fetch from GitLab, the library produces a lot of API calls. I'm all in favor of implementing this with pagination as fallback:

> First, try if there is a branch with all segments (which is probably in most cases successful) and continue with (3) in case there is no match (or even better continue with (1) instead ?).

However, that still not explains for sure the root cause of the actual problem.

components/server/src/gitlab/gitlab-context-parser.ts
corneliusludmann: Yes. All information available in `possibleBranches` API error are now logged here: https://github.com/gitpod-io/gitpod/blob/7785a6c2ca7b58c2ed85d02a13be39b553dad6af/components/server/src/gitlab/gitlab-context-parser.ts#L208

components/server/src/gitlab/gitlab-context-parser.ts
corneliusludmann: If added:
https://github.com/gitpod-io/gitpod/blob/7785a6c2ca7b58c2ed85d02a13be39b553dad6af/components/server/src/gitlab/gitlab-context-parser.ts#L221

That should be sufficient for most of our cases (and for all prebuilds because they never have a file path in the context URL).

To verify, that the GitLab library we use handles pagination correctly, I added a log entry that tells us the number of fetch branches/tags:

https://github.com/gitpod-io/gitpod/blob/7785a6c2ca7b58c2ed85d02a13be39b553dad6af/components/server/src/gitlab/gitlab-context-parser.ts#L291-L293

If we observe that it's always 20 when it fails, we know for sure that pagination is the problem.

What do you think?

AlexTugarev: 👍🏻  for the optimization to start with all segments joined, which might be the dominant case, at least one of them.

just to recap the situation with `feature/9999_new_feature` branches: even if pagination works nicely, it could lead to `(n/20)` API calls, which might still might eat up one sec or two for larger repos/projects. 

I believe a) it will be less API call in average and b) the code will be less verbose if we sort the array of segment prefix' and loop over a single procedure.

corneliusludmann: @AlexTugarev: I'm convinced. Here we are. The PR has been updated.

Unfortunately, the GitLab API answers in some situations with 500 instead of 404 when the branch name is incorrect. We need to check for both. Let's hope, that this will not hide other reasons for that 500 error.

AlexTugarev: Can you post such a case, please? It doesn't sound right. We can file an issue for GitLab.

corneliusludmann: > Can you post such a case, please? It doesn't sound right.

`https://gitlab.com/gp-test-group/gp-test-subgroup/gp-test-project-in-subgroup/-/tree/corneliusludmann` is one

> We can file an issue for GitLab.

Exactly. I have it already on my to-do list.

AlexTugarev:
```

  21 passing (1m)
  4 failing

  1) TestGitlabContextParser
       testTreeContext_tag_01:

      AssertionError: expected { Object (isFile, path, ...) } to have deep property 'refType' of 'tag', but got 'branch'
      + expected - actual

      -branch
      +tag
      
      at TestGitlabContextParser.<anonymous> (src/gitlab/gitlab-context-parser.spec.ts:175:36)
      at fulfilled (src/gitlab/gitlab-context-parser.spec.ts:19:58)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)

  2) TestGitlabContextParser
       testTreeContextSubgroup_tag_01:

      AssertionError: expected { Object (isFile, path, ...) } to have deep property 'refType' of 'tag', but got 'branch'
      + expected - actual

      -branch
      +tag
      
      at TestGitlabContextParser.<anonymous> (src/gitlab/gitlab-context-parser.spec.ts:423:36)
      at fulfilled (src/gitlab/gitlab-context-parser.spec.ts:19:58)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)

  3) TestGitlabContextParser
       testTreeContextTagWithSlash:

      AssertionError: expected { Object (isFile, path, ...) } to have deep property 'refType' of 'tag', but got 'branch'
      + expected - actual

      -branch
      +tag
      
      at TestGitlabContextParser.<anonymous> (src/gitlab/gitlab-context-parser.spec.ts:507:36)
      at fulfilled (src/gitlab/gitlab-context-parser.spec.ts:19:58)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)

  4) TestGitlabContextParser
       testEmptyProject:
     AssertionError: expected { Object (isFile, path, ...) } to have deep property 'ref'
      at TestGitlabContextParser.<anonymous> (src/gitlab/gitlab-context-parser.spec.ts:548:36)
      at fulfilled (src/gitlab/gitlab-context-parser.spec.ts:19:58)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)

```

@corneliusludmann this change breaks context parser tests, unfortunately.

AlexTugarev:
@corneliusludmann, you can run the tests with 
```
cd components/server
export export GITPOD_TEST_TOKEN_GITLAB="{ \"value\": \"YOURTOKEN\", \"username\": \"oauth2\", \"scopes\": [ \"api\", \"read_user\" ] }"
yarn test
```

corneliusludmann:
Ah, it skips it when the env var is not set. That explains why it didn't fail for me. :roll_eyes: Thanks for the hint.

corneliusludmann:
Stupid me. Sorry for not recognizing this and thanks for carefully reviewing the PR, @AlexTugarev. :+1: 

I've fixed this and updated the PR.

There is still a failing test but this test fails on `main` as well. I've created the issue https://github.com/gitpod-io/gitpod/issues/4773 to fix this in the course of groundwork.

components/server/src/gitlab/gitlab-context-parser.ts
corneliusludmann: GitLab issue: https://gitlab.com/gitlab-org/gitlab/-/issues/335867

AlexTugarev: 🥇  Awesome! Thanks for taking care of this and providing details to GitLab!

AlexTugarev (approved):
'Tested with positive and negative examples and it worked nicely!

Thanks for fixing this, @corneliusludmann!

🎉 🎉 🎉

components/server/src/gitlab/gitlab-context-parser.ts
AlexTugarev: quick note on the response 500.
I've seen spikes of request times in server's logs. instead of regular ~300ms it took ~5s, which after checking turned out to be the response 500.
the it was a show branch request for `foo`, where `foo` does not exist but `foo/bar` does. this appears to be a real bug in GitLab.

Issue #4700, opened 2021-07-05
[ws-scheduler] Take pod resource into account


csweichel:
In core-dev we often see `OutOfpods: Pod Node didn't have enough resource: pods, requested: 1, used: 110, capacity: 110`. ws-scheduler should take `pods` resource into account.

geropl:
`ws-scheduler` should take this into account indeed.

Note, however, the issue with `core-dev` stems from our not-well-managed dev cluster, and we run into this issue with static deployment, too.

csweichel:
Good point. For core-dev we could change the CIDR range as you suggested

meysholdt:
I looked into increasing the max pod numbers by using larger CIDR ranges, but no luck:

```
gcloud container clusters create cluster220-1 \
  --project me-cidr \
  --enable-ip-alias \
  --cluster-ipv4-cidr 10.0.0.0/21 \
  --services-ipv4-cidr 10.4.0.0/19 \
  --create-subnetwork name='my-subnet1',range=10.4.32.0/23 \
  --default-max-pods-per-node 220 \
  --zone europe-west1
```
returns 

> ERROR: (gcloud.container.clusters.create) ResponseError: code=400, message=Maximum pods per node must be at least 8 and at most 110, current number is 220.


And [the docs](https://cloud.google.com/kubernetes-engine/docs/how-to/flexible-pod-cidr#cidr_ranges_for_clusters) say 

> Although having 110 Pods per node is a hard limit, you can reduce the number of Pods on a node

meysholdt:
Let's do the math how many preview envs we can have once #4744 has been merged:

On the core-dev clusters, there are
* 8 DaemonSets in `kube-system`
* 1 DaemonSet in cloud-monitoring: `node-exporter`
* 1 DaemonSet in docker: `docker-engine`
* 3 DaemonSets per preview-env: `agent-smith`, `registry-facade`, `ws-daemon`.

110 pods max per node minus 10 (kube-system, docker, monitoring) leaves 100 for preview-envs.

While 100 slots should in theory be enough room for 33 preview envs, in practice there will probably be trouble earlier because already-running pods won't re-locate to other nodes to make room for DaemonSets.

csweichel:
That is a pity - thank you for checking this one though :) 

Looks like sooner or later we'll have to move to more isolated preview environments.
Let's hope that the ws-scheduler fix brings some help.

meysholdt:
[...]

PR #4750, opened 2021-07-08
[preview] Decrease preview env density

geropl:


princerachit:
Should we have an equivalent change in the [other file/(s)](https://github.com/gitpod-io/gitpod/blob/main/.werft/values.k3sWsCluster.yaml) too?

geropl:
> Should we have an equivalent change in the other file/(s) too?

Are we deploying multiple preview environments to them? If yes: maybe. But let's start with the problems we know about.

.werft/values.dev.yaml
```diff
@@ -25,6 +25,15 @@ tracing:
...
   samplerParam: "1"
 
+# we hit the "max. 110 pods/node" situation pretty often with our current core-dev setup.
+# the proper way to fix those would be to adjust the CIDR for workload NodePool which is a bit more work.
+# as a workaround, we blow up our RAM requests to trigger scaleup earlier.
+# Note: This only works because we tune down our DaemonSet's requests to near-0 (because DS pods don't trigger scalups!)
+resources:
+  default:
+    # as opposed to 200Mi, the default
+    memory: 500Mi
```
csweichel, line 35: Why 500Mi? Is there some kind of calculation behind this akin to https://github.com/gitpod-io/gitpod/issues/4700#issuecomment-876314437

princerachit:
> > Should we have an equivalent change in the other file/(s) too?
> 
> Are we deploying multiple preview environments to them? If yes: maybe. But let's start with the problems we know about.

Yes we are deploying multiple preview env to k3s ws cluster

.werft/values.dev.yaml
geropl: No calculation as I assumed assignment of static pods to be too dynamic to reason about.

Let's try. To come as close as possible to the theoretical limit of 33 preview envs are goals are:
 - don't consume too much RAM to drive node count up artificially
 - ensure we never reach "max nr of pods" with static deployments

Thus we make them big enough so that say 100 pods fill up the whole node (we ignore the DaemonSets here because they are quite small).
 => 32Gi / 100 ~ 328Mi
 If we choose `350Mi` we're coming close enough it seems.
 
 WDYT? @csweichel

geropl:
> Yes we are deploying multiple preview env to k3s ws cluster

But we're not having trouble with it atm, plus not all preview envs deploy to that cluster I recon. Let's wait until we have problems, we can quickly come back to this, then.

.werft/values.dev.yaml
csweichel: Sounds good to me. Let's do this

csweichel (approved):

PR #4751, opened 2021-07-08
[server] Fix HTTP endpoint metrics

### LINKED ISSUE #4741: [server] Produce metrics for all HTTP endpoints (incl. websockets)
Basically providing hit counter in the form of
```
gitpod_server_http_request_total{path="/gitpod", status=200} 1
```

csweichel:
This PR fixes the HTTP endpoint metrics on server, and makes them work for websocket connections.

### How to test
1. Produce a bunch of requests against the server, including websocket reconnects (e.g. reload the dashboard, visit `/api/version`)
2. Inspect the server metrics looking  for `gitpod_server_http_requests_*`

fixes #4741

### DIFF
--- a/components/server/src/express/ws-handler.ts
+++ b/components/server/src/express/ws-handler.ts
@@ -12,6 +12,7 @@ import * as url from 'url';
 import * as net from 'net';
 import { WsLayer } from './ws-layer';
 import { log } from '@gitpod/gitpod-protocol/lib/util/logging';
+import { increaseHttpRequestCounter } from '../prometheus-metrics';
 
 export type HttpServer = http.Server | https.Server;
 export type Route = string | RegExp;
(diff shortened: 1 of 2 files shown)


corneliusludmann (approved):
Change looks good. :+1: 

It's odd that statusCode is `0` here but I guess that's expected, right?
```
gitpod_server_http_requests_total{method="GET",route="/gitpod",statusCode="0"} 14
```

csweichel:
> Change looks good. 👍
> 
> It's odd that statusCode is `0` here but I guess that's expected, right?
> 
> ```
> gitpod_server_http_requests_total{method="GET",route="/gitpod",statusCode="0"} 14
> ```

That is expected indeed. It's the same behaviour we've seen from caddy

PR #4753, opened 2021-07-08
[server] Add trackEvent to support dashboard analytics

### LINKED ISSUE #4728: [server] Add analytics API
In order to send events from the dashboard and other places (vs code, local companion app, etc.) to our analytics end-point, the server should expose an API. I think it would be easiest to add a `trackEvent` method to the websocket protocol.

csweichel:
This PR adds a `trackEvent` call to the Gitpod server to enable analytics event propagation from the dashboard.

fixes #4728

### DIFF
--- a/components/gitpod-protocol/src/analytics.ts
+++ b/components/gitpod-protocol/src/analytics.ts
@@ -1,0 +1,40 @@
+/**
+ * Copyright (c) 2021 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { Without } from "./util/without";
+
+export const IAnalyticsWriter = Symbol("IAnalyticsWriter");
+
+type Identity =
+    | { userId: string | number }
+    | { userId?: string | number; anonymousId: string | number };
+
+interface Message {
+    messageId?: string;
+}
+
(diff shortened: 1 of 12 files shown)


svenefftinge:
Awesome! Would it make sense to use a reduced shape for the message that doesn't require adding the userid and timestamp as we can add that on the server side?

components/gitpod-protocol/src/analytics.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+export const IAnalyticsWriter = Symbol("IAnalyticsWriter");
+
+type Identity =
+    | { userId: string | number }
+    | { userId?: string | number; anonymousId: string | number };
+
+interface Message {
+    messageId?: string;
```
svenefftinge, line 14: What is the messageId and its purpose? Can it be omitted on the client-side and generically added when forwarded to segment?

csweichel: `messageId` is not required per se, but can be used to de-duplicate messages. Segment will de-dup messages with the same ID that came in in a 24h window.

csweichel:
I've updated the shape to better reflect which fields we actually use.
The `timestamp` is optional and set by the analytics writer (e.g. segment) when needed.

svenefftinge (approved):

PR #4756, opened 2021-07-08
[ws-daemon] initial implementation of cgroup customizer to support /dev/fuse

### LINKED ISSUE #4652: Device cgroup for /dev/fuse
### Bug description

`/dev/fuse` can't be opened in a gitpod workspace

```
openat(AT_FDCWD, "/dev/fuse", O_RDWR)   = -1 EPERM (Operation not permitted)
```

### Steps to reproduce

I suggest reproducing this in gitpod-staging because the mknod fix on fuse is not in production yet  (PR #4594)

Create a file named `open_fuse.c` in the workspace with the following content:

```
#define _GNU_SOURCE
#include <unistd.h>

#include <sys/syscall.h>
#include <linux/fs.h>
#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>
#include <stdio.h>


int main() {
  const char* src_path = "/dev/fuse";
  unsigned int flags = O_RDWR;
  printf("RET: %ld", syscall(SYS_openat, AT_FDCWD, src_path, flags));
}
```

Now compile it

```
gcc open_fuse.c
```

Now execute it, it will show that the return is `-1` which is `-EPERM`.

### Expected behavior

The fuse device can be opened and used.

### Example repository

_No response_

### Anything else?

# Analysis of a solution

I debugged this a bit and the error is triggered exactly in `device_cgroup.c`:
-  https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/security/device_cgroup.c?h=v5.4.128#n824

Relevant code

```c
/**
 * __devcgroup_check_permission - checks if an inode operation is permitted
 * @dev_cgroup: the dev cgroup to be tested against
 * @type: device type
 * @major: device major number
 * @minor: device minor number
 * @access: combination of DEVCG_ACC_WRITE, DEVCG_ACC_READ and DEVCG_ACC_MKNOD
 *
 * returns 0 on success, -EPERM case the operation is not permitted
 */
int __devcgroup_check_permission(short type, u32 major, u32 minor,
				 short access)
{
	struct dev_cgroup *dev_cgroup;
	bool rc;

	rcu_read_lock();
	dev_cgroup = task_devcgroup(current);
	if (dev_cgroup->behavior == DEVCG_DEFAULT_ALLOW)
		/* Can't match any of the exceptions, even partially */
		rc = !match_exception_partial(&dev_cgroup->exceptions,
					      type, major, minor, access);
	else
		/* Need to match completely one exception to be allowed */
		rc = match_exception(&dev_cgroup->exceptions, type, major,
				     minor, access);
	rcu_read_unlock();

	if (!rc)
		return -EPERM;

	return 0;
}
```

The reason is that the device cgroup that containerd/runc setup does not allow handling with that type of device, even if it's mknod inside the container (which is allowed) as we do.

From `man 7 cgroups`

```
       devices (since Linux 2.6.26; CONFIG_CGROUP_DEVICE)
              This supports controlling which processes may create
              (mknod) devices as well as open them for reading or
              writing.  The policies may be specified as allow-lists and
              deny-lists.  Hierarchy is enforced, so new rules must not
              violate existing rules for the target or ancestor cgroups.

              Further information can be found in the kernel source file
              Documentation/admin-guide/cgroup-v1/devices.rst (or
              Documentation/cgroup-v1/devices.txt in Linux 5.2 and
              earlier).
```

Here is the cgroup of the `containerd-shim` process

```
CGROUP
12:memory:/system.slice/containerd.service,9:devices:/system.slice/containerd.service,8:cpu,cpuacct:/system.slice/containerd.service,6:blkio:/system.slice/containerd.servic
```

fntlnz:
fixes #4652 

## Todo:
- [x] write a program that can customize cgroups for a given workspace
- [x] customize the cgroup to allow `rwm` for fuse devices 
- [ ] integrate the program so that when a workspace its created it is called to customize cgroups

## How to test (this will work when all the tasks above are completed)

1. Start a workspace in the environment
2. try to mount the workspace itself via sshfs

```
sudo apt update -y
sudo apt install sshfs openssh-sftp-server -y
sudo ln -s /usr/lib/sftp-server /usr/libexec/sftp-server
ssh-keygen
cp ~/.ssh/id_rsa.pub ~/.ssh/authorized_keys
mkdir -p /tmp/workspace
sshfs -odebug,sshfs_debug,loglevel=debug -p 23001 gitpod@127.0.0.1:/workspace /tmp/workspace
```

You can also verify the cgroups in the current workspace via:

```
cat /proc/1/cgroups
```

Then you can take the cgroup path and verify its content

```
 cat /sys/fs/cgroup/devices/kubepods/burstable/podab0ba05b-6232-4f7d-956b-320665b0e621/d2c70f16cab0537a5708ae059ed42748b7b776c8d7003e680802db29cf975858/devices.list
c *:* m
b *:* m
c 1:3 rwm
c 1:8 rwm
c 1:7 rwm
c 5:0 rwm
c 1:5 rwm
c 1:9 rwm
c 5:1 rwm
c 136:* rwm
c 5:2 rwm
c 10:200 rwm
c 10:229 rwm <-- this is the fuse allow
```

## Note for reviewer:

The approach is very similar to `seccomp-profile-installer` but for cgroups.


fntlnz:
closing in favor of #4762

PR #4764, opened 2021-07-09
[dashboard] Make sure the error thrown from getLoggedInUser takes pre…

geropl:
…sence before all other errors

Part of #4710 .

This ensures that we receive special error codes like "SETUP_REUQIRED" which are only emitted by getLoggedInUser, and those are not shadowed by a benign "not logged in" error.


AlexTugarev (approved):
Makes sense. Thanks!

PR #4765, opened 2021-07-09
[test] Self-Hosted 0.10.0

geropl:
/werft with-installer
/werft with-retag


geropl:
Was for testing only.

PR #4766, opened 2021-07-09
[bpf] Reduce QEMU disk size

### LINKED ISSUE #4763: [devx] Decrease size of workspace on gitpod-io/gitpod
Original Slack thread: https://gitpod.slack.com/archives/C01KGM9AW4W/p1625820943461200

Problem: Currently workspaces are around 25GB of size, from originally 2-3GB. This slows down "stop" and "restart" of a workspace.

This seems to be cause by the QEMU image that every workspaces contains ([link](https://github.com/gitpod-io/gitpod/blob/ccb09c9c4d1366763de364e80668d92e3d62ad2b/components/ee/agent-smith/scripts/prepare-bpf-dev-environment.sh)).
We should try to use a image of dynamic size as described in [this thread](https://askubuntu.com/questions/1298291/how-to-create-dynamic-disks-qcow2-as-it-happens-with-vdi).

geropl:
Fixes #4763.

This was shorter than I thought. :see_no_evil:  Reduces from 22GB -> 2.4GB:
image

### DIFF
--- a/components/ee/agent-smith/scripts/prepare-bpf-dev-environment.sh
+++ b/components/ee/agent-smith/scripts/prepare-bpf-dev-environment.sh
@@ -23,7 +23,7 @@ cd "$outdir"
 
 tar -xvf rootfs.tar.gz
 
-qemu-img resize bionic-server-cloudimg-amd64.img +20G
+qemu-img resize --preallocation=off bionic-server-cloudimg-amd64.img +20G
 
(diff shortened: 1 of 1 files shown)


fntlnz:
Cool, inside looks the same. LGTM

Nice finding and good job @geropl 

image

fntlnz (approved):

PR #4772, opened 2021-07-12
[admin] Fix adminRestoreSoftDeletedWorkspace

### LINKED ISSUE #4583: [dashboard] restore & pin on doesn't work
### Bug description

On the admin dashboard, when I click on "Restore & Pin" on a deleted workspace. Nothing happens. log says

```
construct.js:19 Uncaught (in promise) Error: operation not permitted: missing get permission on workspace
    at main.js:449
    at main.js:276
    at main.js:260
    at setImmediate.js:40
    at f (setImmediate.js:69)
    at n (setImmediate.js:109)
```

### Steps to reproduce

PS: It's a private Gitlab repo soft deleted by the user

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

corneliusludmann:
The “Restore & Pin” link in the admin portal does not work. This PR fixes the rights check.

image

### How to test

1. Login with a user A. This user is the admin user.
2. Login in another session (e.g. incognito window) with another user B. This user is a normal user.
3. Start & stop a workspace with user B. Delete this workspace.
4. Look for this workspace in the admin dashboard with user A. There should be a “Restore & Pin” link. Click it!
5. Verify that the “Soft Deleted” column updates to “No” immediately for user A.
6. Verify that user B has the workspace in the list of active workspaces (pinned).

It's important to do this with 2 users because in the previous implementation the problem was, that you can restore your own workspace only (due to the rights check).

Fixes #4583


geropl (approved):
Tested, works as advertised.

A small reminder in favor of integration tests: large parts of the test could be automated. Don't want to block the PR, just to nag a bit :upside_down_face:

PR #4775, opened 2021-07-12
[prow] Add OWNERS files

csweichel:
This PR prepares the `OWNERS` files for prow. It assigns ownership much like the `CODEOWNERS` file did, except that changes to any of the `-api` packages require approval from @csweichel or one other team member (who depends on the API). Although it should, this does not include `gitpod-protocol`, because that's too much of a hotchpotch of concerns and the [`approve` plugin does not support OWNERS `filter`](https://www.kubernetes.dev/docs/guide/owners/#owners).

Merging this PR is without effect until we enable prow for this repo.

cc @svenefftinge as proxy for @JanKoehnlein 

/werft no-preview


csweichel:
Feedback:
- only owning teams should be assigned for reviews
- second API approver should be a relevant other person (e.g. local app or supervisor: anton)

OWNERS
```diff
@@ -0,0 +1,3 @@
+approvers:
```
leodido, line 1: This will have an effect on all the nested OWNERS files.

If this is not what you want, then adding no_parent_owners: true to nested OWNERS will fix it.

leodido: Also, in case some areas/components absolutely (and always) require a person to review it (but not approve) you can define "required reviewers".

[This struct](https://github.com/kubernetes/test-infra/blob/ae26658abae0b1213549d1326b3b252422fe0215/prow/repoowners/repoowners.go#L240) helped me in the past :)

While [this one](https://github.com/kubernetes/test-infra/blob/ae26658abae0b1213549d1326b3b252422fe0215/prow/repoowners/repoowners.go#L47) is about the `no_parent_owners` thing.

leodido: Also, in case some areas/components absolutely (and always) require a person to review it (but not approve) you can define `required_reviewers` key.

[This struct](https://github.com/kubernetes/test-infra/blob/ae26658abae0b1213549d1326b3b252422fe0215/prow/repoowners/repoowners.go#L240) helped me in the past :)

csweichel:
Thank you @leodido for the insightful comments. I've updated the owners files accordingly

leodido (approved):
LGTM 💯

PR #4778, opened 2021-07-13
[code] remove oldest machine resources on max payload error

### LINKED ISSUE #3277: [code] Remove "machine resources" when deleting a workspace
A user can see a list of all synchronised machines (workspaces) in Code which grows infinitely with each new workspace. We should make sure that a machine gets removed as soon as an associated workspace is deleted.

akosyakov:

#### What it does

- fix #3277 by removing the oldest machine on too large payload

Change in Gitpod Code: https://github.com/gitpod-io/vscode/commit/a17670ba5af14e0faf3a6927983468d28fda235b

#### How to test

The second commit limits the max payload for machine resource to 2 machines. So if you start 3-4 workspaces you should see only 2 machines (#3 and #4) in synced machines view (F1 -> Open View -> Synced Machines), first 2 machines should be removed. I will drop the last commit before merging.


corneliusludmann (approved):
Works at advertised. Ready to merge after removing the “limit machines to reproduce” commit.

PR #4788, opened 2021-07-13
[gpctl] Talk to ws-manager hosts directly

csweichel:
This PR enables `gpctl workspaces` to talk to a ws-manager host directly without assuming a correct kubeconfig.

### DIFF
--- a/dev/gpctl/cmd/workspaces.go
+++ b/dev/gpctl/cmd/workspaces.go
@@ -7,7 +7,10 @@ package cmd
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"fmt"
+	"io/ioutil"
+	"path/filepath"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


dev/gpctl/cmd/workspaces.go
```diff
@@ -28,75 +31,109 @@ var workspacesCmd = &cobra.Command{
...
+		addr = "localhost:20202"
+	} else {
+		port, _ := workspacesCmd.PersistentFlags().GetString("port")
+		addr = fmt.Sprintf("%s:%s", host, port)
 	}
 
-	secopt := grpc.WithInsecure()
-	if certFromSecret, _ := workspacesCmd.Flags().GetBool("tls-from-secret"); certFromSecret {
-		certPool, err := util.CertPoolFromSecret(clientSet, namespace, "ws-manager-tls", []string{"ca.crt"})
+	if fn, _ := workspacesCmd.Flags().GetString("tls-path"); fn != "" {
```
princerachit, line 104: nit: having this as another function will reduce loc in this funtion.

princerachit (approved):

dev/gpctl/cmd/workspaces.go
csweichel: It would indeed. We've reached a point where we should pull something like this into common-go anyways

PR #4808, opened 2021-07-14
rewrite gitpod-schema location

### LINKED ISSUE #4806: json catalogue points to invalid gitpod scheme
### Bug description

It expects https://gitpod.io/schemas/gitpod-schema.json, but we move it to  https://gitpod.io/static/schemas/gitpod-schema.json. We should make sure old url still works, otherwise it breaks yaml smartness.

### Steps to reproduce

See above.

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

akosyakov:
#### What it does

- fix #4806

#### Hot to test

https://akosyakov-json-catalogue-points-4806.staging.gitpod-dev.com/schemas/gitpod-schema.json should return proper JSON

### DIFF
--- a/components/dashboard/conf/Caddyfile
+++ b/components/dashboard/conf/Caddyfile
@@ -51,6 +51,8 @@
 	redir /.well-known/security.txt https://www.gitpod.io/.well-known/security.txt permanent
 	redir /environment-variables /settings/ permanent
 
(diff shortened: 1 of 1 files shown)


aledbf (approved):
LGTM

PR #4811, opened 2021-07-14
[server] private repo access after cancellation

svenefftinge:
If a user cancels their subscription, they should still be able to use it until it ends.

### DIFF
--- a/components/ee/payment-endpoint/src/accounting/subscription-service.ts
+++ b/components/ee/payment-endpoint/src/accounting/subscription-service.ts
@@ -98,10 +98,10 @@ export class SubscriptionService {
      * @param date The date on which the subscription has to be active
      * @returns Whether the user has an active subscription (user-paid or team s.) at the given date
      */
-    async hasActiveAndNotYetCancelledPaidSubscription(userId: string, date: Date): Promise<boolean> {
+    async hasActivePaidSubscription(userId: string, date: Date): Promise<boolean> {
         const subscriptions = await this.accountingDB.findActiveSubscriptionsForUser(userId, date.toISOString());
         return subscriptions
-            .filter(s => Subscription.isActive(s, date.toISOString()) && !Subscription.isCancelled(s))
+            .filter(s => Subscription.isActive(s, date.toISOString()))
             .length > 0;
     }
 
--- a/components/server/ee/src/user/eligibility-service.ts
+++ b/components/server/ee/src/user/eligibility-service.ts
@@ -252,7 +252,7 @@ export class EligibilityService {
             return true;
         }
 
-        return this.subscriptionService.hasActiveAndNotYetCancelledPaidSubscription(user.id, date);
+        return this.subscriptionService.hasActivePaidSubscription(user.id, date);
     }
 
     /**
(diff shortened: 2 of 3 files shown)


components/ee/payment-endpoint/src/accounting/subscription-service.ts
```diff
@@ -101,7 +101,7 @@ export class SubscriptionService {
     async hasActiveAndNotYetCancelledPaidSubscription(userId: string, date: Date): Promise<boolean> {
         const subscriptions = await this.accountingDB.findActiveSubscriptionsForUser(userId, date.toISOString());
         return subscriptions
-            .filter(s => Subscription.isActive(s, date.toISOString()) && !Subscription.isCancelled(s))
+            .filter(s => Subscription.isActive(s, date.toISOString()))
```
geropl, line 104: The method does what it says on the tin (name), and is used in the other places correctly I think.

But [this call](https://github.com/gitpod-io/gitpod/blob/e8762f258e8b81d0a8da0b8062e7352722948e6c/components/server/ee/src/user/eligibility-service.ts#L255) is odd indeed: I agree that we should call `hasActivePaidSubscription` with the implementation you suggested.

geropl: @jankeromnes I faintly recall we added made the change here together, right? Do you recall what the motivation was for checking `isActive && !isCancelled` instead of `isActive`? IMO this was because of a user that:
 1. bought a subscription
 1. quickly cancelled it (and we reimbursed them)
 1. expected to be back in trial mode immediately.
 
In general the user-frriendly pattern of "easy-in-easy-out" makes sense here.

components/server/ee/src/user/eligibility-service.ts
```diff
@@ -285,24 +285,6 @@ export class EligibilityService {
         return new Date(freeTrialStartDate);
     }
 
-    /**
```
svenefftinge, line 288: I couldn't find a user, so good to remove.

components/server/ee/src/user/eligibility-service.ts
```diff
@@ -252,7 +252,7 @@ export class EligibilityService {
             return true;
         }
 
-        return this.subscriptionService.hasActiveAndNotYetCancelledPaidSubscription(user.id, date);
+        return this.subscriptionService.hasActivePaidSubscription(user.id, date);
```
geropl, line 255: :+1:

components/server/ee/src/user/eligibility-service.ts
```diff
@@ -314,7 +296,7 @@ export class EligibilityService {
         if (start === undefined) {
             return undefined;
         }
-        if (await this.subscriptionService.hasActiveAndNotYetCancelledPaidSubscription(user.id, date)) {
+        if (await this.subscriptionService.hasActivePaidSubscription(user.id, date)) {
```
geropl, line 299: This is an edge case, I'm sure we can handle the impact. We dearly need (more) tests including reasoning for this kind of policy changes. :see_no_evil:

geropl (approved):
Did not test but LGTM

PR #4816, opened 2021-07-14
[gp] env: handle multi-word values without quotes

### LINKED ISSUE #4736: gp env - fails when value contains a space or newline
### Bug description

After the fix for #4493 was deployed to production, it introduced a new bug.
Now when the value contains space character, `gp env` fails.

### Steps to reproduce

Open any Gitpod workspace.
In terminal, type:
`gp env message="hello world"`
```bash
empty string (correct format is key=value)
```

Compare it with a regular environment variable that doesn't have this issue:
1. `export message="hello world"`
1. `echo $message`
1. Confirm the result is `hello world`

### Expected behavior

`gp env` should allow the same values that Bash's `export` does.

Here's how I want to use it, but now it is broken :(

```bash
ssh-keygen -q -t rsa -b 4096 -f ~/.ssh/id_rsa
gp env MY_SSH_KEY=$(cat ~/.ssh/id_rsa)
```

### Example repository

_No response_

### Anything else?

_No response_

geropl:
Fixes #4736

### DIFF
--- a/components/gitpod-cli/cmd/env_test.go
+++ b/components/gitpod-cli/cmd/env_test.go
@@ -38,6 +38,31 @@ func TestParseArgs(t *testing.T) {
 			"",
 		},
 		{"key value empty quoted value", []string{"key=\"\""}, nil, "variable must have a value; use -u to unset a variable"},
+		{"key value containing spaces", []string{"key=\"hello world\""}, []*serverapi.UserEnvVarValue{
+			{Name: "key", Value: "hello world"},
+		},
+			"",
+		},
+		{"key value containing space, quotes stripped by shell", []string{"key=hello world"}, []*serverapi.UserEnvVarValue{
+			{Name: "key", Value: "hello world"},
+		},
+			"",
+		},
+		{"key value containing newline", []string{"key=hello\nworld"}, []*serverapi.UserEnvVarValue{
+			{Name: "key", Value: "hello\nworld"},
+		},
+			"",
+		},
+		{"value containing equals sign", []string{"key=hello=world"}, []*serverapi.UserEnvVarValue{
+			{Name: "key", Value: "hello=world"},
+		},
+			"",
+		},
+		{"value containing quoted equals sign", []string{"key=\"hello=world\""}, []*serverapi.UserEnvVarValue{
+			{Name: "key", Value: "hello=world"},
+		},
+			"",
+		},
 	}
 
 	for _, test := range tests {
--- a/components/gitpod-cli/cmd/env.go
+++ b/components/gitpod-cli/cmd/env.go
@@ -12,8 +12,6 @@ import (
 	"strings"
 	"sync"
 	"time"
-	"unicode"
-	"unicode/utf8"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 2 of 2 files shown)


components/gitpod-cli/cmd/env_test.go
```diff
@@ -38,6 +38,21 @@ func TestParseArgs(t *testing.T) {
...
+			"",
+		},
+		{"key value containing space, quotes stripped by shell", []string{"key=hello world"}, []*serverapi.UserEnvVarValue{
+			{Name: "key", Value: "hello world"},
+		},
+			"",
+		},
+		{"key value containing newline", []string{"key=hello\nworld"}, []*serverapi.UserEnvVarValue{
+			{Name: "key", Value: "hello\nworld"},
+		},
```
csweichel, line 53: Please consider the following cases:
```
{"value containing equals sign", []string{"key=hello=world"}, []*serverapi.UserEnvVarValue{
	{Name: "key", Value: "hello=world"},
},
{"value containing quoted equals sign", []string{"key=\"hello=world\""}, []*serverapi.UserEnvVarValue{
	{Name: "key", Value: "hello=world"},
},
```

shaal:
@geropl I'd love helping testing this. But I would need steps I can follow to run tests on this PR.

I found https://werft.gitpod-dev.com/job/gitpod-build-geropl-gp-env-fails-when-value-4736.2, and I can get the Gitpod interface open.
But when I try opening a repo using that interface, I am getting this error message -
image

geropl:
@shaal You're absolutely right, forgot about that I hav to unblock.

It would be awesome if you could double check the list of test cases [here](https://github.com/gitpod-io/gitpod/blob/9cad3ee1e27e5d25e560d01294c8236f839c5a26/components/gitpod-cli/cmd/env_test.go) and suggest additions if any of your use-cases it missing. Thx!

csweichel (approved):
Pending @shaal's check if the test cases cover the actual issue, this LGTM

shaal:
@geropl when I look for an available environment I can test with here - https://werft.gitpod-dev.com, the only one that work is
https://werft.gitpod-dev.com/job/gitpod-build-geropl-gp-env-fails-when-value-4736.3
When I try running my repo on it, it says I am blocked.

When I try more recent environments, like https://werft.gitpod-dev.com/job/gitpod-wipe-devstaging-geropl-gp-env-fails-when-value-4736.8 - I don't see a way to open it to test.

Can you please create an environment I can use to test?

csweichel:
You can extract the updated `gp` command and use it in any workspace (e.g. prod) using:
```
docker run --rm -it --entrypoint=sh -v $PWD:/tmp/out eu.gcr.io/gitpod-core-dev/build/image-builder:commit-9cad3ee1e27e5d25e560d01294c8236f839c5a26 -c "mkdir /tmp/out/gplayer && cd /tmp/out/gplayer && tar xf /app/workspace-image-layer.tar.gz"
```

The `gplayer/gitpod-cli` binary is the new `gp` command that includes the fixes.

geropl:
@shaal See @csweichel 's response. Also, unblocked you on: https://geropl-gp-env-fails-when-value-4736.staging.gitpod-dev.com/workspaces

shaal:
@geropl @csweichel Thank you!
I tried running all the test, as well as my use case `gp env myssh="$(cat ~/.ssh/id_rsa)"`
Everything works as expected, thank you for fixing that!

I don't know if this is related or not, there were 2 issues I noticed when unsetting an environment variable:
1. When I unset an environment variable using `gp env -u wow`, no output is displayed to confirm the action (but perhaps this is by design, similar to how `unset` command works?)
1. When I try to unset an environment variable that no longer exist `gp env -u wow`, I am getting this message
```
cannot unset wow: jsonrpc2: code 404 message: cannot delete 'wow' in scope 'shaal/drupalpod'
```
I think the error message can be simplified, but this is a very minor issue.

geropl:
Thx for confirming!

Regarding 1: Yes, that's indeed meant to be scriptable.
Regarding 2: IMO this is a bug: it should be "no output, exitcode 0" as well, but I see our code tries to be explicit there.. worth another issue if you ask me.

shaal:
Thank you for working on it and fixing it so fast!

@geropl Do you know when will this fix get deployed to gitpod.io? 
It will make my repo usable again 😁

PR #4817, opened 2021-07-14
[dev] mysql: add explicit resource request

geropl:
YAML fact you never wanted to know, fact #120323: In the bitnami mysql chart, [resources](https://github.com/bitnami/charts/blob/0a9d4c7fb9096443393b05ee2a0c7572df9e5ba2/bitnami/mysql/values.yaml#L273-L286) belongs under [primary](https://github.com/bitnami/charts/blob/master/bitnami/mysql/values.yaml#L146).


csweichel (approved):
🤦

PR #4829, opened 2021-07-15
[Analytics] Fixed wrong name in signup tracking

jakobhero:
Fixes that fullName instead of name is passed to name attribute in analytics signup tracking call.

### DIFF
--- a/components/server/src/user/user-controller.ts
+++ b/components/server/src/user/user-controller.ts
@@ -517,7 +517,7 @@ export class UserController {
             properties: {
                 "auth_provider": user.identities[0].authProviderId,
                 "email": User.getPrimaryEmail(user),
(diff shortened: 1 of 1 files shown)


components/server/src/user/user-controller.ts
```diff
@@ -517,7 +517,7 @@ export class UserController {
             properties: {
                 "auth_provider": user.identities[0].authProviderId,
                 "email": User.getPrimaryEmail(user),
-                "name": User.getName(user),
+                "name": user.identities[0].authName,
```
geropl, line 520: I see how that fixes the particular issue. But I wonder how we're sure that `identities[0]` is what we want? But outside of scope I guess.

geropl (approved):
code LGTM, signup and login still works

PR #4930, opened 2021-07-24
Revert "[ws-daemon] Remove Containerd4214 workaround"

csweichel:
Reverts gitpod-io/gitpod#4683

We've seen this issue persisting in production

PR #5153, opened 2021-08-11
[ws-daemon] Remove Containerd4214 workaround

csweichel:
Reverts gitpod-io/gitpod#4930

Turns out this "workaround" is hiding more issues that it solves.


roboquat:
LGTM label has been added.  <details>Git tree hash: c43d938a120e9185bb39ce9533763010117c497e</details>

PR #5157, opened 2021-08-11
disable userfaultfd for all unprivileged users

### LINKED ISSUE #5156: Disable userfaultfd for unprivileged users
Following up from an [internal discussion](https://gitpod.slack.com/archives/C01MLB7J24T/p1628680795008400?thread_ts=1628603354.007100&cid=C01MLB7J24T).

### Context

The `userfaultfd` syscall (and subsequent API) can be misused making it easier to exploit existing use-after-free bugs that might otherwise only make a short window or race condition available. 

By using the `userfaultfd` system call to stall a kernel thread, a malicious program can keep some state, that it wrote, stable
for an extended period, which it can then access using an existing exploit. 

While it doesn't cause the exploits itself, and while it's not the only thing (**for example**, we should also investigate `fusermount` too)  that can stall a kernel thread when accessing a memory location, I believe it is wise to (at least stochastically) reduce the attack surface.

### Notes

This **may** (I've not tested this) interfere with qemu (when using the postcopy feature that requires trapping page faults), kvm, and libvirt.

FYI: on more recent kernels (like 5.12) support to SELinux for controlling access to the `userfaultfd` interface has been added ([1](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=29cd6591ab6fee3125ea5c1bf350f5013bc615e1), [2](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=29cd6591ab6fee3125ea5c1bf350f5013bc615e1)).

leodido:
The `userfaultfd` system call (and subsequent API) can be misused making it easier to exploit existing use-after-free (and similar) bugs that might otherwise only make a short window or race condition available.

Thus, we here disable it (on the host nodes) for unprivileged users through the global sysctl knob "vm.unprivileged_userfaultfd".

Fixes #5156 

Signed-off-by: Leo Di Donato <leodidonato@gmail.com>

### DIFF
--- a/chart/templates/ws-daemon-daemonset.yaml
+++ b/chart/templates/ws-daemon-daemonset.yaml
@@ -208,6 +208,7 @@ spec:
               sysctl -w "net.ipv4.tcp_tw_reuse=1" &&
               sysctl -w fs.inotify.max_user_watches=1000000 &&
               sysctl -w "kernel.dmesg_restrict=1"
+              sysctl -w vm.unprivileged_userfaultfd=0
             ) && echo "done!" || echo "failed!"
       containers:
       - name: ws-daemon


roboquat:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please assign **rl-gitpod** after the PR has been reviewed.
You can assign the PR to them by writing `/assign @rl-gitpod` in a comment when ready.

Associated issue: *#5156*

<details open>

</details>

csweichel:
@leodido I'm afraid the branch name is too long which breaks the HTTPS certificate process. Please re-submit with a short branch name.

PR #5164, opened 2021-08-12
[ws-manager] Skip error log for pods already removed

aledbf:
Internal link https://cloudlogging.app.goo.gl/Yi4cnjdtGFfk6YXw5


csweichel:
I reckon this would hide an actual failure case. It's probably the containerd workaround that prematurely removes pods, which trips up ws-manager. I'd rather we observe the behaviour in prod for a while, now that we've removed the workaround, and see how this message behaves.

PR #4501, opened 2021-06-14
[payment] cancel all subscriptions on user deletion

JanKoehnlein:
fixes #4229

PR #5165, opened 2021-08-12
[server] don't let user deletion fail

svenefftinge:
... because the subscription deletion didn't work.


JanKoehnlein:
Hmm, I am not sure if thats enough. In particular, when there is a CB problem, the user will silently keep paying.
I've a similar PR which has not been merged yet because we found no way of testing it (same for this PR)
  https://github.com/gitpod-io/gitpod/pull/4501
I think it covers the issue a bit better and also addresses the problem with user's still signed up for a team. Maybe you could just review that, we merge it and test it on staging.

svenefftinge:
Sure, didn't see that PR!

PR #5223, opened 2021-08-16
[ws-manager] Provide ordering of status updates

csweichel:
This PR introduces a `status_version` field on workspace status which imposes a partial order of status updates. Using this field we can determine which update "came first", rather than having to rely on heuristics or best effort ordering. Due to the prior lack of such an order, we have seen otherwise functional workspaces revert back to a non-running state (e.g. "Opening IDE").

More formally speaking, this PR introduces `status_version` so that for two status updates `s1` and `s2`, `s1` was a status before `s2` if `s1.statusVersion < s2.statusVersion`.

### How could this be solved?
We have several options available for producing such a field:
- _real world time_ which, because we're only producing status updates from a single process looks like the obvious choice. However, [real wall time is a tricky beast](https://aphyr.com/posts/299-the-trouble-with-timestamps). It can go backwards (e.g. after an NTP re-sync), can run slower or faster (depending on the quartz in the machine, or NTP clock skew), and generally differs between machines even when they're synched to the same NTP server by up to 1 second.
- _logical clock/lamport clock_ is the distributed system 101 (see [paper](http://lamport.azurewebsites.net/pubs/time-clocks.pdf) or explanation [here](https://martinfowler.com/articles/patterns-of-distributed-systems/lamport-clock.html) and [here](https://people.cs.rutgers.edu/~pxk/417/notes/logical-clocks.html)). However, we have no high-frequency state storage available in ws-manager. We cannot update a Kubernetes object everytime we receive a status update as that would likely burst the k8s API rate limits quickly - it certainly would not scale. Introducing local state just for this feature seems overkill as well.
- _logical clocks_ derived from our existing state seem like a good idea. Especially because Kubernetes already offers a `resourceRevision`. However, there's a number of issues with this approach. For one the set of objects is not stable, e.g. when ports are opened a new service may be created. Some variant of a vector clock where we consider the Kubernetes objects the processes might be used - if there is such a thing as a "dynamic vector clock". However, the Kubernetes resource version must not be interpreted (they come from etcd/postgresql) - it does not even impose an order, but just says if two versions are different.
- [_hybrid logical clocks_](https://cse.buffalo.edu/tech-reports/2014-04.pdf) mix real-world time with logical clocks. While we do not (yet?) have difference processes to consider, we can use the guarantees afforded by the naive implementation to overcome the limitations of real world and logical clocks when used independently.

### How do "hybrid logical clocks" (HLC) work?
HLC work around the limitations of real-wall clocks by throwing a logical clock in the mix which steps in when the real-wall time does funny things, e.g. run backwards. Whenever the real time moves backwards we rely on the logical clock for each status update until the real clock has caught up. Real-wall time and logical time are encoded in a single `uint64` as described in the paper linked above, with 48bit for the real time and 16 bit for the logical time.

### What happens when ws-manager restarts and lands on a node with a different time?
In that case it could happen that the causally newer status updates have a "lower" `status_version`. This would violate the order we want to impose. Suppose however, that the real-wall time of all machines ws-manager could run are guaranteed (at least very likely) to not differ by more than `eps`, then waiting for `eps` prior to producing new updates is sufficient to guarantee that the new status updates will have a newer status version. In GKE machines are timesynched using Google's NTP server, hence can be expected to certainly drift less than 1 second apart (see [NTP performance](http://www.ntp.org/ntpfaq/NTP-s-algo.htm#Q-ACCURATE-CLOCK)). Hence, waiting for `eps = 2 seconds` should be more than plenty to avoid this issue.

### What happens if for some reason the real-wall clock jumped ahead/ran backwards too far?
We spend 16 bit of the HLC on the logical clock, hence can support 65535 "ticks" until the real wall time has recovered. In a workspace cluster with 1000 workspaces, each producing 10 events per minute (port opening, start, stop, prebuilds, etc). we would consume the logical clock in 6.5 minutes. Once the logical clock is "consumed", i.e. would overflow, the implementation will panic to avoid breaking the order guarantee we wish to impose. In this case, ws-manager would start afresh with a newly initialized real-wall time. If we can maintain the order guarantee then is not certain, but likely (see point above).


roboquat:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please ask for approval from **csweichel** after the PR has been reviewed.

<details open>

- **[components/ws-manager/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/ws-manager/OWNERS)**

</details>

JanKoehnlein:
Looks like we're still lacking the processing of the new `statusVersion` on the `ws-man-brigde` side, or am I missing something?

csweichel:
> Looks like we're still lacking the processing of the new `statusVersion` on the `ws-man-brigde` side, or am I missing something?

Indeed, we do not. This change just introduces the value. Once we have it we can observe its behaviour and eventually introduce logic based on it.

roboquat:
LGTM label has been added.  <details>Git tree hash: ad434b3d9fd7d54f48615e9356afdce4d36133d5</details>

roboquat:

This pull-request has been approved by: *<a href="https://github.com/gitpod-io/gitpod/pull/5223#issuecomment-930056519" title="Approved">aledbf</a>*

Associated issue requirement bypassed by: *<a href="https://github.com/gitpod-io/gitpod/pull/5223#issuecomment-930056519" title="Approved">aledbf</a>*

<details >

- ~~[components/ws-manager/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/ws-manager/OWNERS)~~ [aledbf]
- ~~[components/ws-manager-api/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/ws-manager-api/OWNERS)~~ [aledbf]

</details>

PR #5424, opened 2021-08-27
Add workspace start request counter

ArthurSens:
We're finally moving forward with our first SLOs documented at our [blogpost](https://www.gitpod.io/blog/journey-to-site-reliability-engineering-at-gitpod). 🥳 

## What this PR does

This PR implements a counter that helps us with the SLO:
* SLO: "XX% of requests for workspaces should succeed(in a user perspective)".
 ```
sum(rate(gitpod_server_workspace_starts_total{status="failed"}[30d])) 
/
sum(rate(gitpod_server_workspace_starts_total[30d]))
```

## Follow up work
After getting this PR merged, I'll:
1. Implement a dashboard focused on the SLO target and error budgets for workspace start requests.
2. Start working on a histogram that measures duration of workspace start latency, aiming for full user-experience in that request. That means going through all components involved in that process

### DIFF
--- a/components/ws-manager/pkg/manager/manager.go
+++ b/components/ws-manager/pkg/manager/manager.go
@@ -146,32 +146,41 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
 	tracing.ApplyOWI(span, owi)
 	defer tracing.FinishSpan(span, &err)
 
+	response := &api.StartWorkspaceResponse{
+		Url:        "",
+		OwnerToken: "",
+		Success:    false,
+	}
+
 	// Make sure the objects we're about to create do not exist already
 	exists, err := m.workspaceExists(ctx, req.Id)
 	if err != nil {
-		return nil, xerrors.Errorf("cannot start workspace: %w", err)
+		return response, xerrors.Errorf("cannot start workspace: %w", err)
 	}
 	if exists {
-		return nil, status.Error(codes.AlreadyExists, "workspace instance already exists")
+		return response, status.Error(codes.AlreadyExists, "workspace instance already exists")
 	}
 	span.LogKV("event", "workspace does not exist")
 	err = validateStartWorkspaceRequest(req)
 	if err != nil {
-		return nil, xerrors.Errorf("cannot start workspace: %w", err)
+		return response, xerrors.Errorf("cannot start workspace: %w", err)
 	}
 	span.LogKV("event", "validated workspace start request")
 	// create the objects required to start the workspace pod/service
 	startContext, err := m.newStartWorkspaceContext(ctx, req)
 	if err != nil {
-		return nil, xerrors.Errorf("cannot create context: %w", err)
+		return response, xerrors.Errorf("cannot create context: %w", err)
 	}
+
+	response.OwnerToken = startContext.OwnerToken
+	response.Url = startContext.WorkspaceURL
 	span.LogKV("event", "created start workspace context")
 	clog.Info("starting new workspace")
 	// we must create the workspace pod first to make sure we don't clean up the services or configmap we're about to create
 	// because they're "dangling".
 	pod, err := m.createWorkspacePod(startContext)
 	if err != nil {
-		return nil, xerrors.Errorf("cannot create workspace pod: %w", err)
+		return response, xerrors.Errorf("cannot create workspace pod: %w", err)
 	}
 	span.LogKV("event", "pod description created")
 	err = m.Clientset.Create(ctx, pod)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: components/server/src/prometheus-metrics.ts [hunk 1]
```diff
@@ -100,4 +100,16 @@ export function increaseMessagebusTopicReads(topic: string) {
     messagebusTopicReads.inc({
         topic,
     })
+}
+
+const workspaceStartHistogram = new prometheusClient.Histogram({
+    name: 'gitpod_server_workspace_start_duration_seconds',
```
csweichel, line 106: I reckon this name might be confusing, because
1. it does not reflect actual workspace startup time, but rather the time spent during a workspace start before it hits ws-manager/Kubernetes
2. it also includes image builds, hence mixes two modes of operation which will render the metric all but meaningless (you basically see two modes in the distribution, one without image builds and one with).

In addition to the name, the metric currently does not actually capture the user experience. When opening `gitpod.io/#some-context/url` we first run `createWorkspace` which - depending on the response times of the Git hoster - can take a few seconds to complete as well.

It makes perfect sense to measure the time we spend in server to create a workspace and issue the start request. Ideally though we'd don't include the image build if there is one, and to get the end-user perspective consider the workspace creation case, too.

For introducing the workspace start SLO I reckon we would get by with a simple counter vec, counting how many requests succeed. WDYT?

jankeromnes:
Looks cool! 🎉

However, it seems that the review request to me was unintentional. So, removing, but please feel free to re-add it again if you're expecting a review from me.

components/server/src/prometheus-metrics.ts
ArthurSens: > In addition to the name, the metric currently does not actually capture the user experience. When opening gitpod.io/#some-context/url we first run createWorkspace which - depending on the response times of the Git hoster - can take a few seconds to complete as well.

Thanks for this info! I didn't know about this `createWorkspace` part.

> It makes perfect sense to measure the time we spend in server to create a workspace and issue the start request. Ideally though we'd don't include the image build if there is one, and to get the end-user perspective consider the workspace creation case, too.

Yeah exactly, our current workspace startup time metric only measures time spent on ws-manager. For full user experience, we need to include all steps in between "User requests a new workspace" until "The IDE opened in the user's browser". I think that if in between these 2 moments, there is an image being built that the user needs to wait for, then it should be included in the measured duration.

> For introducing the workspace start SLO I reckon we would get by with a simple counter vec, counting how many requests succeed. WDYT?

To reduce initial complexity sounds like a good idea 🙂. I can create the dashboard based on this, and update in the future after we get the histogram working

ArthurSens:
> Looks cool! 🎉
> 
> However, it seems that the review request to me was unintentional. So, removing, but please feel free to re-add it again if you're expecting a review from me.
> 
> Tip: If you add `/uncc` to your PR comment, @roboquat will not immediately flag random reviewers.

Hey @jankeromnes, the assignment to you was not unintentional 😬. I believe that the OWNERS file contains a list of people that has the best experience with the code I'm modifying, so I assume that whoever is auto-assigned to review this PR would be a good reviewer 🙂.


If that is not true, I guess the OWNERS file needs to be reviewed?

components/server/src/workspace/workspace-starter.ts
```diff
@@ -121,11 +123,11 @@ export class WorkspaceStarter {
             // If the caller requested that errors be rethrown we must await the actual workspace start to be in the exception path.
             // To this end we disable the needsImageBuild behaviour if rethrow is true.
             if (needsImageBuild && !options.rethrow) {
-                this.actuallyStartWorkspace({ span }, instance, workspace, user, mustHaveBackup, userEnvVars, options.rethrow, forceRebuild);
+                this.actuallyStartWorkspace({ span }, instance, workspace, user, mustHaveBackup, startTime, userEnvVars, options.rethrow, forceRebuild);
```
ArthurSens, line 126: hmmm this is not necessary anymore since we're not going with a duration histogram in this PR. I'll remove it

roboquat:

<details open>

</details>

roboquat:

<details open>

- **[components/ws-manager/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/ws-manager/OWNERS)**

</details>

components/ws-manager-api/core.proto
```diff
@@ -95,6 +95,9 @@ message StartWorkspaceResponse {
 
     // OwnerToken is the token of the workspace owner used for authentication
     string owner_token = 2;
+
+    // Success tells if the workspace started successfully or not
+    bool success = 3;
```
csweichel, line 100: this does not make sense. If the workspace start (as seen from the `StartWorkspace` function) was not successful the call would return an error

ArthurSens: From what I could understand, Server does not receive the error, it only receives `StartWorkspaceResponse` from the API. And at this moment `StartWorkspaceResponse` does not have any information telling if ws-manager was able to start the workspace or not.

geropl: Hm. This covers only a small subset of possible "workspace start failures". From the top of my head a workspace can fail during the following activities during startup (once it has been created in the DB):
 - image-builder (and interaction with it): captured in this PR in `workspace-starter.ts`
 - communication with `ws-manager`: captured by this PR by adding the `success`
 - everything else that can go wrong after the call to `ws-manager` has been issued (content initialization, image pull, etc.): not covered

IMO the 1st and 3rd part are the most interesting ones. An alternative approach would be to scan the DB regularly for workspaces that failed (`status.failure`) and use that as basis for a metric. That would not be a "workspace start request counter", but far more interesting I think.

PR #5687, opened 2021-09-14
[supervisor-api] Add Java gRPC API

corneliusludmann:
## Description
This PR adds a Java gRPC API for the `supervisor-api`.

## Related Issue(s)

Solves partly #5603

## How to test
For a quick test, there is a `TestClient` class that connects to supervisor and gets the Git token. Before merging, we probably will remove this test class.

```
$ cd components/supervisor-api/java
$ ./gradlew run
```

## Release Notes

```release-note
NONE
```

## ToDo
- [ ] Publish jar artifact
- [ ] `PortVisibility` is `status.proto` uses reserved keywords (`public` and `private`) as enum values that breaks the Java classes.

## Meta

/werft no-preview
/nocc


roboquat:

Associated issue: *#5603*

<details open>

- **[components/supervisor-api/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/supervisor-api/OWNERS)**

</details>

components/supervisor-api/status.proto
```diff
@@ -122,8 +123,8 @@ message PortsStatusResponse {
 }
 
 enum PortVisibility {
-    private = 0;
-    public = 1;
+    privateVisibility = 0;
+    publicVisibility = 1;
```
csweichel, line 127:
```suggestion
    private_visibility = 0;
    public_visibility = 1;
```

The camelCase doesn't transport on the Go side.
That said, what motivates that change?

corneliusludmann: Motivation:

> PortVisibility is status.proto uses reserved keywords (public and private) as enum values that breaks the Java classes.

(see ToDo above)

For now, it's a temporary fix. Still need a good way to handle this problem. Suggestions welcome. Thought about just patching the generated Java files after generating.

akosyakov: There probably some proto annotations which can drive java generation?

akosyakov:
@corneliusludmann Is it good to review?

I looked through the code. It looks good, but I was expecting to see BUILD.yaml which publishes a new version whenever there are changes and tag it per a branch, similarly how we do for ts apis.

corneliusludmann:
> I looked through the code. It looks good, but I was expecting to see BUILD.yaml which publishes a new version whenever there are changes and tag it per a branch, similarly how we do for ts apis.

Actually, we do not run the API generator during the CI build at all. It's always a manual execution of the generator shell scripts. We could add such a job, however:

- Publishing an artifact is unique to the Java API library (the TS and Go API libs are consumed directly and not published somewhere).
- I guess we do not want to generate the code automatically to have more control over it (that's why we have the generated code under version control). Updates of gRPC generator breaks the code sometimes.
- When we don't generate the code during the CI job, it also makes no sense to publish a new version on each CI build.

That's the reason that I think adding a CI job makes no sense. WDYT?

akosyakov:
> That's the reason that I think adding a CI job makes no sense. WDYT?

If I change api I want to publish new api. With ts we detect than proto components changed and publish automatically. Without anything like that we need to document and explain how to publish and share secrets.

akosyakov:
I think we can merge it like that if it already unblocks other jetbrains issues. But we should another issue which adds a proper BUILD.yml and anyone could run something like `leeway ./components/supervisor-api/java:publish` to get a new version and tag for testing.

csweichel:
The generated java files miss the license header. I reckon there are two ways around this:
1. (my preferred way): we don't have any consumer of the Java version of the API in this repository, but all clients consume the published jar file, no? In that case, we might as well not add 58k LOC to our repo and not commit the Java files. A statement in the readme would clarify the motivation.
2. extend the addheader program to handle Java files, resp. make sure everything is tied in using leeway (also see Anton's comment above).

corneliusludmann:
> we don't have any consumer of the Java version of the API in this repository, but all clients consume the published jar file, no?

Eventually, we will. When we have the JetBrains images in this repo we have the Java code of the heartbeat plugin in this repo as well that consumes the lib directly. Then, we even don't need to deploy it to the maven registry anymore.

atduarte:
The local dependency strategy looks good to me. Even though it requires the dev to build the dependencies before being ready to build the java-test or have linting/code completion, I think we can solve that. In the worst-case scenario, we add a README.md to the component. 

Although I would advise against having the generated java code in `supervisor-api`:
1. Generated code files make it harder to search for things in the repo.
2. Their path and name don't clearly indicate they are generated code.
3. The IDE navigation (e.g. ctrl+click on a class) will lead the developer to the corresponding `.class` in the dependency jar (or `.java` if there's a sources jar), not the `.java` file in the `supervisor-api/java` folder.

Caveat: I might be missing some general guidance or reasoning. Particularly because I know we already have generated code in the case of other languages, but I assumed it's for language-specific or historical reasons.

corneliusludmann:
> Even though it requires the dev to build the dependencies before being ready to build the java-test or have linting/code completion, I think we can solve that.

For the TypeScript and Go components, we have a startup task in the `.gitpod.yaml` file that builds everything. We should add this for the Java files as well which would solve this (I'll do this).



> Although I would advise against having the generated java code in `supervisor-api`:

Yes, having generated code under version control is always a controversial topic. I actually don't know exactly why we commit the generated files (e.g. why we don't trust the code generator or want to be able to track changes when a new version of the code generator is used). However, I'm in favor of being aligned with the go and TypeScript code and therefore have it under version control as well.

> 2. Their path and name don't clearly indicate they are generated code.

Sure. We can definitely move the generate java files to another folder and/or package. What would you propose? `src/gen` folder?

> 3. The IDE navigation (e.g. ctrl+click on a class) will lead the developer to the corresponding `.class` in the dependency jar (or `.java` if there's a sources jar), not the `.java` file in the `supervisor-api/java` folder.

Any idea how to fix this?

roboquat:
LGTM label has been added.  <details>Git tree hash: 6cb0268a0d53a264f79646f89f503ed79cc5aa49</details>

PR #5742, opened 2021-09-17
[server] added an endpoint to subscribe to newsletter

### LINKED ISSUE #5598: automatically forward newsletter sign ups to customer.io
With the newsletter signup widget, website visitors have the option to sign up for emails without logging into Gitpod. 
Right now we are collecting these email addresses in a [spreadsheet](https://docs.google.com/spreadsheets/d/1DrgFXvDsLmu6kUoBHMfgYqBtqyT9kr4OHqcKX31FbXw/edit#gid=722237941). There are currently >150 email addresses. So far these email addresses are not used and forwarded to customer.io.

**What needs to be done:**
- collect emails separately depending if someone signs up for the devx email or [changelog](https://www.gitpod.io/changelog) 
- update the notification setting of existing users
- create a new user in customer.io if the email address doesn't belong to an existing user

**Next steps (according to @atduarte ):**
In order to implement this, I believe that the Gitpod app (not the website) should be responsible for the flow from form submission (or API call) to Segment updates.

1. The website would call the Gitpod app with the email
2. The app would then check whether the email is registered to decide whether to update the associated setting
3. The app publishes a Segment event (to be listened to by customer.io) with the email, user id or anonymous id, and subscription status (i.e. unsubscribed=false)
4. Customer data is upserted in Customer.io, using the email as the identifier.

**context:** 
[#696](https://github.com/gitpod-io/website/issues/696#issuecomment-894394666)
[#762](https://github.com/gitpod-io/website/pull/762)

Feel free to complement the issue in case I missed something.

JanKoehnlein:
## Description
Adds a route to the server to subscribe to newsletters

## Related Issue(s)
Fixes #5598

## How to test
TBD

## Release Notes
```release-note
NONE
```

### DIFF
--- a/components/server/src/user/newsletter-subscription-controller.ts
+++ b/components/server/src/user/newsletter-subscription-controller.ts
@@ -8,6 +8,24 @@ import * as express from 'express';
 import { inject, injectable } from "inversify";
 import { UserDB } from "@gitpod/gitpod-db/lib";
 import { IAnalyticsWriter } from "@gitpod/gitpod-protocol/lib/analytics";
+import { log } from '@gitpod/gitpod-protocol/lib/util/logging';
+
+interface SubscriptionData {
+    email: string
+    context: string
+    trackingId?: string
+    newsletterType: string
+}
+
+namespace SubscriptionData {
+    export function is(o: any): o is SubscriptionData {
+        return o !== undefined &&
+            typeof o.email === "string" &&
+            typeof o.context === "string" &&
+            typeof o.newsletterType === "string" &&
+            (o.trackingId == undefined || typeof o.trackingId === "string");
+    }
+}
 
 @injectable()
 export class NewsletterSubscriptionController {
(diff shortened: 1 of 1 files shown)


roboquat:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please ask for approval from **jankoehnlein** after the PR has been reviewed.

Associated issue: *#5598*

<details open>

</details>

components/server/src/user/newsletter-subscription-controller.ts
```diff
@@ -85,6 +104,56 @@ export class NewsletterSubscriptionController {
...
+                    res.status(400).send('Invalid message body');
+                    log.error("Invalid subscribe request", data)
+                    return;
+                }
+                if (!acceptedNewsletterTypes.includes(data.newsletterType)) {
+                    log.error("Invalid newsletter in subscribe request", data)
+                    res.status(400).send('Invalid newsletter type');
+                    return;
+                }
+                const user = (await this.userDb.findUsersByEmail(data.email))[0];
```
AlexTugarev, line 120: 1. I'm concerned about adding an open endpoint to patch any user's settings like this. Same applies to connecting to 3rd party services. Obviously, there are very simple ways how to abuse this and even disrupt our services. We should consider a verification protocol (or other suitable mechanisms) for protection.

2. Email addresses are not suitable to identify users. `findUsersByEmail` should not be used to select a user in general. We should rename it to discourage usage of it. Currently, it's used to verify email addresses already used in existing accounts.

JanKoehnlein: I understand your concerns. What do you propose? We need a way for un-registered users to un-/subscribe to the newsletter.
Should we pull newsletter subscription out of the app (and the db) entirely and handle it in segment/customer.io only?

AlexTugarev: the case of unregistered user in particular isn't clear to me: should we really introducing an open relay for a connected service? also, doesn't subscribing to newsletter require us to send verification emails? 

> Should we pull newsletter subscription out of the app (and the db) entirely and handle it in segment/customer.io only?

If we want to enable newsletter signups for unregistered users, it really looks like a separate service which we then would need to talk to from the web app to adjust settings.

Issue #5612, opened 2021-09-08
Allow users to cancel pending or running prebuilds for projects


gtsiolis:
### Problem to solve

Currently, users cannot cancel a `PENDING` or `RUNNING` prebuild on the a) project a) Configuration page, b) Prebuilds list, and c) Prebuild page.

See also https://github.com/gitpod-io/gitpod/pull/5474 and https://github.com/gitpod-io/gitpod/issues/5392#issuecomment-906271953 for initial specs. 🖍️ 

Also, canceling on the Prebuilds list triggers an alert which ideally could be avoided by removing the Cancel action on the more actions dropdown. ❗ 

| Prebuilds List | Alert | 
|-|-|
| <img width="1353" alt="Screenshot 2021-09-08 at 8 58 28 PM" src="https://user-images.githubusercontent.com/120486/132561285-519c47a7-a949-42e4-9739-7365b8bb9e30.png"> |  <img width="475" alt="Screenshot 2021-09-08 at 8 58 28 PM copy" src="https://user-images.githubusercontent.com/120486/132561316-2529f4ea-263c-4014-be1c-f5143c09c7bf.png"> |

roboquat:
@JanKoehnlein: Issue scheduled in the meta team (WIP: 0)

<details>

In response to [this](https://github.com/gitpod-io/gitpod/issues/5612#issuecomment-920769235):

>/schedule


Instructions for interacting with me using PR comments are available [here](https://git.k8s.io/community/contributors/guide/pull-requests.md).  If you have questions or suggestions related to my behavior, please file an issue against the [kubernetes/test-infra](https://github.com/kubernetes/test-infra/issues/new?title=Prow%20issue:) repository.
</details>

jankeromnes:
Happy to take a look at this again next week 👍

PR #5835, opened 2021-09-23
[server] Delete unused PrebuildQueueMaintainer and PrebuildRateLimiter

jankeromnes:
As discussed in our call today @AlexTugarev 🙂 

## Description

The `PrebuildQueueMaintainer` and `PrebuildRateLimiter` classes are currently dead code, and wouldn't work as is today anyway.

## Release Notes

```release-note
NONE
```

/uncc

PR #5865, opened 2021-09-24
[server][dashboard] Allow cancelling Prebuilds

jankeromnes:
## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/5612

## Release Notes

```release-note
Make it possible to cancel pending or running Prebuilds
```

/uncc

### DIFF
--- a/components/dashboard/src/projects/ConfigureProject.tsx
+++ b/components/dashboard/src/projects/ConfigureProject.tsx
@@ -140,6 +140,14 @@ export default function () {
     }
   }
 
+  const cancelPrebuild = async () => {
+    try {
+      // await getGitpodService().server.cancelPrebuild(project.id, TODO prebuildId);
+    } catch (error) {
+      setEditorMessage(<EditorMessage type="warning" heading="Could not cancel prebuild." message={String(error).replace(/Error: Request \w+ failed with message: /, '')}/>);
+    }
+  }
+
   const onInstanceUpdate = (instance: WorkspaceInstance) => {
     setPrebuildInstance(instance);
   }
@@ -181,7 +189,9 @@ export default function () {
           {(prebuildInstance?.status.phase === "stopped" && !prebuildInstance?.status.conditions.failed)
               ? <a className="my-auto" href={`/#${project?.cloneUrl}`}><button className="secondary">New Workspace</button></a>
               : <button disabled={true} className="secondary">New Workspace</button>}
-          <button disabled={isDetecting || (prebuildWasTriggered && prebuildInstance?.status.phase !== "stopped")} onClick={buildProject}>Run Prebuild</button>
+          {(prebuildWasTriggered && prebuildInstance?.status.phase !== "stopped")
+              ? <button className="danger" disabled={prebuildInstance?.status.phase !== "initializing" && prebuildInstance?.status.phase !== "running"} onClick={cancelPrebuild}>Cancel Prebuild</button>
+              : <button disabled={isDetecting} onClick={buildProject}>Run Prebuild</button>}
         </div>
       </div>
     </div>
--- a/components/server/ee/src/prebuilds/prebuild-manager.ts
+++ b/components/server/ee/src/prebuilds/prebuild-manager.ts
@@ -5,7 +5,7 @@
  */
 
 import { DBWithTracing, TracedWorkspaceDB, WorkspaceDB } from '@gitpod/gitpod-db/lib';
-import { CommitContext, Project, StartPrebuildContext, StartPrebuildResult, User, WorkspaceConfig, WorkspaceInstance } from '@gitpod/gitpod-protocol';
+import { CommitContext, Project, StartPrebuildContext, StartPrebuildResult, User, WorkspaceConfig, WorkspaceInstance, HeadlessWorkspaceEvent, HeadlessWorkspaceEventType } from '@gitpod/gitpod-protocol';
 import { log } from '@gitpod/gitpod-protocol/lib/util/logging';
 import { TraceContext } from '@gitpod/gitpod-protocol/lib/util/tracing';
 import { inject, injectable } from 'inversify';
@@ -14,6 +14,7 @@ import { HostContextProvider } from '../../../src/auth/host-context-provider';
 import { WorkspaceFactory } from '../../../src/workspace/workspace-factory';
 import { ConfigProvider } from '../../../src/workspace/config-provider';
 import { WorkspaceStarter } from '../../../src/workspace/workspace-starter';
+import { MessageBusIntegration } from '../../../src/workspace/messagebus-integration';
 import { Config } from '../../../src/config';
 
 export class WorkspaceRunningError extends Error {
@@ -36,6 +37,7 @@ export class PrebuildManager {
     @inject(TracedWorkspaceDB) protected readonly workspaceDB: DBWithTracing<WorkspaceDB>;
     @inject(WorkspaceFactory) protected readonly workspaceFactory: WorkspaceFactory;
     @inject(WorkspaceStarter) protected readonly workspaceStarter: WorkspaceStarter;
+    @inject(MessageBusIntegration) protected readonly messageBus: MessageBusIntegration;
     @inject(HostContextProvider) protected readonly hostContextProvider: HostContextProvider;
     @inject(ConfigProvider) protected readonly configProvider: ConfigProvider;
     @inject(Config) protected readonly config: Config;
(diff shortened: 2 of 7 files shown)


roboquat:

Associated issue: *#5612*

<details >

</details>

components/server/ee/src/prebuilds/prebuild-manager.ts
```diff
@@ -137,6 +139,30 @@ export class PrebuildManager {
...
 
+    async cancelPrebuild(ctx: TraceContext, user: User, prebuildId: string): Promise<void> {
+        const span = TraceContext.startSpan("cancelPrebuild", ctx);
+        span.setTag("prebuildId", prebuildId);
+        try {
+            const prebuild = await this.workspaceDB.trace({ span }).findPrebuildByID(prebuildId);
+            if (!prebuild || prebuild.state === 'aborted') {
+                return;
+            }
+            // TODO(janx): stop `prebuild.buildWorkspaceId`?
```
csweichel, line 150: without stopping the actual prebuild, you'll get status updates that reset the changes below.

jankeromnes: Many thanks @csweichel! I was wondering whether the headless update call below would take care of that or not. Let's also wait on stopping the prebuild workspace then :+1:

jankeromnes: Fixed

components/server/ee/src/prebuilds/prebuild-manager.ts
```diff
@@ -107,7 +110,8 @@ export class PrebuildManager {
 
             span.setTag("starting", true);
             await this.workspaceStarter.startWorkspace({ span }, workspace, user, [], {excludeFeatureFlags: ['full_workspace_backup']});
-            return { wsid: workspace.id, done: false };
+            const prebuild = (await prebuildPromise)!;
```
jankeromnes, line 113: I'm just slightly worried about this code:
- Logically, `findPrebuildByWorkspaceID(workspace.id)` should always find a `prebuild`, given that the workspace was created from a `StartPrebuildContext` context
- However, what if there is an edge case, or a new bug in the future 😱

components/server/ee/src/prebuilds/prebuild-manager.ts
```diff
@@ -120,18 +124,43 @@ export class PrebuildManager {
...
                 throw new WorkspaceRunningError('Workspace is still runnning', runningInstance);
             }
             span.setTag("starting", true);
-            const workspace = await this.workspaceDB.trace({ span }).findById(workspaceId);
+            const workspace = await workspacePromise;
             if (!workspace) {
                 console.error('Unknown workspace id.', { workspaceId });
                 throw new Error('Unknown workspace ' + workspaceId);
             }
+            const prebuild = (await prebuildPromise)!;
```
jankeromnes, line 139: Same here (see previous comment)

components/server/src/workspace/workspace-starter.ts
```diff
@@ -265,6 +265,7 @@ export class WorkspaceStarter {
                     await this.workspaceDb.trace({ span }).storePrebuiltWorkspace(prebuild)
                     await this.messageBus.notifyHeadlessUpdate({span}, workspace.ownerId, workspace.id, <HeadlessWorkspaceEvent>{
                         type: HeadlessWorkspaceEventType.Aborted,
+                        // TODO: `workspaceID: workspace.id` not needed here? (found in ee/src/prebuilds/prebuild-queue-maintainer.ts and ee/src/bridge.ts)
```
jankeromnes, line 268: @csweichel We've never added a `workspaceID` here, but I see it in all other occurrences of `notifyHeadlessUpdate`. Is it really okay to omit it here? (Couldn't find who listens to this.)

components/server/ee/src/prebuilds/prebuild-manager.ts
AlexTugarev: indeed, it seems to lack a contract here, i.e. it seems easy to break because it relies on assumptions, though it's not changed frequently.

perhaps a dedicated `workspaceFactory.createPrebuild(..., context): { ws: Workspace, pb: Prebuild }` would document the requirements more maintainable? just thinking out loud...

components/server/ee/src/prebuilds/prebuild-manager.ts
```diff
@@ -120,18 +124,43 @@ export class PrebuildManager {
...
+        const span = TraceContext.startSpan("cancelPrebuild", ctx);
+        span.setTag("prebuildId", prebuild.id);
+        try {
+            if (prebuild.state === 'aborted' || prebuild.state === 'timeout' || prebuild.state === 'available') {
+                return;
+            }
+            prebuild.state = 'aborted';
+            prebuild.error = 'Cancelled manually';
+            await this.workspaceDB.trace({ span }).storePrebuiltWorkspace(prebuild);
+            await this.messageBus.notifyHeadlessUpdate({ span }, user.id, prebuild.buildWorkspaceId, <HeadlessWorkspaceEvent>{
```
AlexTugarev, line 160: how does this play together with state transitions driven by the bridge?
the bridge is receiving status updates from ws manager and does update the prebuild workspace "in parallel". there could be races of `storePrebuiltWorkspace` calls, potentially leading to override the `Aborted` message?

shouldn't the cancellation be a request to the wsman, for which in turn the bridge would to the notifications and updates from this change?

jankeromnes: Good question, thanks! To be honest, I simply copied the prebuild cancellation logic from `bridge.ts`:

https://github.com/gitpod-io/gitpod/blob/6bc80eb92479e2c9d400d059f5ce2dc56bde304f/components/ws-manager-bridge/ee/src/bridge.ts#L77-L96

and also from how it was implemented in the (now-deleted) `PrebuildQueueMaintainer`:

https://github.com/gitpod-io/gitpod/blob/2aa2db6524c5799ee4546879fe9e33efd8330a4e/components/server/ee/src/prebuilds/prebuild-queue-maintainer.ts#L144-L175

I don't quite understand the full state machine yet, but it sounds like I should look more deeply into it. 👀

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -1554,6 +1554,28 @@ export class GitpodServerEEImpl extends GitpodServerImpl<GitpodClient, GitpodSer
         return prebuild;
     }
 
+    async cancelPrebuild(projectId: string, prebuildId: string): Promise<void> {
```
AlexTugarev, line 1557: in theory we could get the `projectId` from the workspace of that prebuild. 
what's the advantage of adding it to the list of args, besides remove a couple of db queries?

jankeromnes: > what's the advantage of adding it to the list of args, besides remove a couple of db queries?

None. Removing a few unnecessary DB queries sounds great, and this is information we typically already have when making the call, so we might as well not omit it. 🙂 Do you see it otherwise?

components/server/ee/src/prebuilds/prebuild-manager.ts
AlexTugarev: @jankeromnes, first of https://github.com/gitpod-io/gitpod/pull/5835/files

then, you see it's state propagation from the `ws-manager-bridge` which runs outside of server. which is the whole point: ws-manager running in region A and propagating event from there, but then you want to cancel a prebuild from server in region B. in such cases I'm puzzled about the stream of states a frontend would receive, thus the message for the user would be confusing.

in general, if we want something to happen with the workspace, we're sending requests to the ws-manager of the according region, which is already taken care of by the client connection as far as know. btw, this is what happens for the stop part of this new composite action called cancelPrebuild. I think what needs to be done here is to introduce a CancelPrebuild request in ws-manager, which basically does the stopping but also signals the action to the bridge to do the rest, i.e. update db and push notifications.

AlexTugarev:
Really enjoying to get some control over the running prebuild! That's an awesome feature 🎉

components/server/ee/src/prebuilds/prebuild-manager.ts
jankeromnes: After our sync discussion, it seems we'll want to remove the new `prebuildManager.cancelPrebuild` call, and instead make `stopWorkspace` also cancel the prebuild (inside the correct `ws-manager`, e.g. with a new flag on the `StopWorkspaceRequest` if necessary).

components/dashboard/src/projects/ConfigureProject.tsx
```diff
@@ -242,7 +260,11 @@ export default function () {
                     {((!isDetecting && isEditorDisabled) || (prebuildInstance?.status.phase === "stopped" && !prebuildInstance?.status.conditions.failed))
                         ? <a className="my-auto" href={`/#${project?.cloneUrl}`}><button className="secondary">New Workspace</button></a>
                         : <button disabled={true} className="secondary">New Workspace</button>}
-                    <button disabled={isDetecting || (prebuildWasTriggered && prebuildInstance?.status.phase !== "stopped")} onClick={buildProject}>Run Prebuild</button>
+                    {(prebuildWasTriggered && prebuildInstance?.status.phase !== "stopped")
+                        ? <button className="danger flex items-center space-x-2" disabled={prebuildWasCancelled || (prebuildInstance?.status.phase !== "initializing" && prebuildInstance?.status.phase !== "running")} onClick={cancelPrebuild}>
+                            <span>Cancel Prebuild</span>
```
gtsiolis, line 265: **issue**: This is most probably out of the scope of these changes but when canceling a prebuild from the project configurator, the prebuild status on the left side of the configurator falls back to `READY`. Could we introduce a `CANCELED` status and use this as a status here? Pasting below an early draft featuring the `CANCELED` status. Is this technically feasible and does this sound expected?

Screenshot 2021-10-04 at 7 01 28 PM

gtsiolis (commented):
Thanks @jankeromnes! So much user control and freedom[[1](https://www.nngroup.com/articles/user-control-and-freedom/)] in one button.

Added two comments below. Feel free to open follow up issues for these two comments if these are out of the scope of these changes and make sense. 🏓

Issue #5643, opened 2021-09-10
Implement workspace heartbeats for Jetbrains IDEs


atduarte:
To avoid the workspace timing out while the user is using the Jetbrains local IDE, we must send heartbeats to Gitpod when there's user activity locally.

**Proposed solution**
Install a plugin in the Jetbrains IDE backend that listens to activities such as "FileOpened" and sends heartbeats to the Gitpod server.

roboquat:
@akosyakov: Issue scheduled in the IDE team (WIP: 0)

<details>

In response to [this](https://github.com/gitpod-io/gitpod/issues/5643#issuecomment-925609561):

>/schedule


Instructions for interacting with me using PR comments are available [here](https://git.k8s.io/community/contributors/guide/pull-requests.md).  If you have questions or suggestions related to my behavior, please file an issue against the [kubernetes/test-infra](https://github.com/kubernetes/test-infra/issues/new?title=Prow%20issue:) repository.
</details>

PR #6152, opened 2021-10-11
Implement Jetbrains IDEs heartbeating

atduarte:
## Description
Create the IntelliJ backend plugin to provide support for Gitpod.

When installed in the headless IntelliJ running in a Gitpod workspace, this plugin monitors user activity of the client IntelliJ and sends heartbeats accordingly. Avoiding the workspace timing out.

## Related Issue(s)
Addresses #5643 

## How to test
Manually tests are required. 
1. Set up a workspace with the headless Jetbrains IDE.
2. Produce the plugin by running `./gradlew buildPlugin`.
3. Unzip `build/distributions/gitpod-intellij-plugin-1.0-SNAPSHOT.zip` to the `plugins/` folder of the headless Jetbrains IDE.
4. Start the headless Jetbrains IDE.

## Release Notes

NONE


corneliusludmann:
Hi @atduarte! I just saw that the [build fails](https://werft.gitpod-dev.com/job/gitpod-build-ad-intellij-backend-plugin.2) because there are missing license headers. You can fix this by running
```
$ leeway run components:update-license-header
```
in a Gitpod workspace and committing the changes.

corneliusludmann:
Hey @atduarte! I started to review your PR. :pray: 

Regarding the CI build:
1. You added the plugin to `:all-apps` of `components/BUILD.yaml` which makes sense. Once we have the Jetbrains IDE image in place, this Docker image will have this plugin as dependencies. At this point having it in `:all-apps` seems legit. However, during the CI build `:all` is executed which does not has `:all-apps` as a dependency (it has only `:all-docker`). That's why this plugin is not built currently. I would suggest to keep this in `:all` for now and remove it from `:all` in the PR that adds the IDE image which is built during `:all-docker` (same for supervisor and protocol Java API packages).
2. I moved it to `:all` in [this werft build](https://werft.gitpod-dev.com/job/gitpod-build-clu-test.13) for testing reasons which show that the approach with the local maven repo does not work, unfortunately. :disappointed:  The reason is that each component build is isolated and therefore they don't share a common local maven repo folder. Even if I think that the local maven repo approach would be really great, I think the easiest path to get this working is to switch back to your approach with the subproject that you had in https://github.com/gitpod-io/gitpod/commit/12d0a4867978d806093105d134313981acc9f5ff. I would propose to set the actual subproject path as property similar to this: https://github.com/gitpod-io/gitpod/commit/d8ab37f1118c50ab59c4c0bbc5977602dfa14f0d. What do you think?

akosyakov:
I missed https://github.com/gitpod-io/gitpod/pull/6152#issuecomment-940119585 first time. I liked the idea. I think we should call it `components/ide/jetbrains` top directory,  `intellij` is just one product.

akosyakov:
@atduarte ok, pick what you think what suits best to describers all such IDEs. Most important that we have them under one folder.

roboquat:
LGTM label has been added.  <details>Git tree hash: 31017ca447c057ec3e93de5fe108e004366246ae</details>

corneliusludmann:
@akosyakov You need to approve this as well because team-ide is not an owner of https://github.com/gitpod-io/gitpod/blob/291bcd9f259ae5e48ab73311e68a4970ea89a432/components/supervisor-api/OWNERS.

_(is it on purpose?)_

akosyakov:
@csweichel Is it fine if we change ownership for supervisor-api and localapp-api to IDE?

akosyakov:
Do kotlin language support work for you? I cannot use find references and navigation?

akosyakov:
/lgtm

I looked through the code and left some comments. I think we can merge but they should be addressed with next PRs. I am also not sure about using Kotlin if the language support in VS Code does not work.

/hold

please remove hold if you want to merge it immediately

Issue #5596, opened 2021-09-08
Automatically choose workspace-cluster based on lowest latency. 


meysholdt:
Context: https://github.com/gitpod-io/gitpod/issues/5534#issuecomment-914967098

## Problem Statement
We currently have workspace clusters in one region in the EU and one region in the US. To offer service at a good latency (e.g. < 100ms), we will need more clusters, maybe as many as one or two per continent. See https://gcping.com/ for your personal latency to every google cloud region. See the [GCP network map](https://cloud.google.com/about/locations) for available regions and connections between them.

## Prior Art
* [Collect 'gcping' data from the dashboard](https://www.notion.so/gitpod/Collect-gcping-data-from-the-dashboard-dacb3eb6a0de4b809e4171a38d499e30) by @jankeromnes .

## Proposed Solution

The user's web browser should measure the latency for every available workspace cluster and send the measurements to the gitpod-server, so that the server can make an informed decision about what workspace-cluster is best for the user. 

## Considerations

* latency measurement should not slow down workspace startup time
* the decision what workspace-cluster to choose should remain with the gitpod-server, because in the future, other factors besides latency may influence the decision: Example: cluster health.

## Proposed Design Choices:

* to keep workspace startup fast, the latency measurement should be cached. For example in a cookie in the web-browser.
* to keep workspace startup fast, the latency measurement should preferable not be done when a workspace starts, but when a user visits any website of gitpod.
* every workspace clusters should have a public endpoint that can be "pinged" from the web browser for latency measurement.
* the server should make a cache-key and the ws-cluster-endpoints available to the users. The cache-key should encode the public IP address of the user, so that the latency will be measured again if the user changes his/her network.

## Example Flow 1:

1. the user visit gitpod.io/workspaces. 
2. the users browser receives `{'cache-key': 'FJJDSKD', "clusters": {"us07": "https://us07.gitpod.io/ping", "sing01": "https://sing01.gitpod.io/ping" } }`
3. the user browser measures the latency to all clusters in the background and stores the result in a cookie: `{"us07": 230, "sing01": 60}`
4. When the user opens a workspace, the cookie will be send to the gitpod-server and the server will use the latency measurement to chose the best workspace cluster.

## Example Flow 2: 
1. the user opens a workspace. The cookie is already there. No delay during workspace-start.

## Example Flow 3: 
1. the user opens a workspace. The cookie is not yet there. The is the case we want to avoid, but I don't think it can be avoided all the time.
[...]

Issue #4809, opened 2021-07-14
A user (with privileges) should be able to chose a workspace cluster explicitly


meysholdt:
Use cases: If there are multiple pluggable workspace clusters...
* with admission constraints and you want to choose one
* or the cluster you want to start a workspace in has a low "score" and you don't want to start 10 workspace to have at least on end up on the right clusters. In fact, some colleagues could not use this strategy in the past because they did not have an "unleashed" plan and ran into the workspace limit too early.

Proposal:
* support an `&cluster=eu12` argument in the context urls, similar to how we do it to trigger prebuilds manually and pass ENV vars.
* since we only use this for testing this argument should only be usable by people who have the permission to do so, e.g. the `new-workspace-cluster` permission. We don't want to give users the ability to override our loadbalancing/failover/cluster-cycling strategies.

Context: @geropl brought this up in today's all-hands.

csweichel:
Other simple solution: extend the admission constraints to explicitly name users, i.e. introduce a `user=` admission constraint.

Pro: does not require plumbing cluster choice to workspace context/config.
Downside: still not as flexible as passing it in as env var.

> or the cluster you want to start a workspace in has a low "score"

This case I don't understand yet. Why can't we just raise the score while testing the cluster?

meysholdt:
> Why can't we just raise the score while testing the cluster?

it pins you to single cluster :( Sometimes it's nice to work with multiple ones: First start your "dev" workspace on a "stable" cluster and then start a test workspace on the k3s-cluster, another one on EKS, and another one on GKE just to make the comparison.

With that being said, it feels like even with a ration of 10000:50 I still end up on the "50" cluster 1/4 of the time, so I ended up starting multiple workspace in the past.

Another nice use-case for having the cluster-name in the URL: We can share such URLs in Slack "here, try this cluster".

csweichel:
> With that being said, it feels like even with a ration of 10000:50 I still end up on the "50" cluster 1/4 of the time, so I ended up starting multiple workspace in the past.

That sounds like a separate bug we should look into

> Another nice use-case for having the cluster-name in the URL: We can share such URLs in Slack "here, try this cluster".

That's a great use-case indeed

csweichel:
In conversation with @meysholdt we've decided to skip this feature for now. Removing from groundwork.

PR #6164, opened 2021-10-11
[wsman-bridge] Introduce admission preferences

csweichel:
## Description
This PR adds _admission preferences_ to the workspace cluster selection mechanism. While admission constraints limit the set of clusters a workspace can be started on, admission preferences express a preference for a set of cluster but still allow fall-back to clusters that do not match that preference.

With this change, we first try and choose a cluster from the preferred clusters (green circle below) following the usual score mechanism. If there are no preferred cluster, we'll fall back to all clusters matching the admission constraints.

image

Use-cases this addition is helpful for:
- **give paying customers their own cluster (original motivation for this change)**: this PR introduces a single admission preference named `user-level`. This in-descriptive level can be used for a variety of purposes, e.g. express "paying customer" status. In a follow-up PR (once I've figured out if a user is actually paying), we would set this level to `paying-customer`. All workspace cluster with an admission preference `user-level=paying-customer` and the (new) admission constraint `has-user-level=paying-customer` would then become exclusive for paying users.
- **explicitly select a cluster to start a workspace one**: suppose we want to test a new workspace cluster, right now we add the `has-permission=workspace-cluster` constraint and a high score, hoping that after so many attempts we land on the new cluster. We could introduce a prefix context parser which takes a cluster name, and add an admission preference for that name.
- **hyper-local workspace clusters**: if we added an admission preference with network locality or some other information we could use to predict quality of service, we could give regional/likely performant clusters preference. The admission preference need not be static, but can be negotiated on a per-user basis.

## How to test
image

## Release Notes

```release-note
Add admission constraints to support fine-grained cluster selection
```


meysholdt:
great stuff!

> explicitly select a cluster to start a workspace on

sounds like https://github.com/gitpod-io/gitpod/issues/4809

> hyper-local workspace clusters

sounds like https://github.com/gitpod-io/gitpod/issues/5596

princerachit:
Overall lgtm other than a few clarifying questions. Also, there are no unit tests, given that we don't have them for any other file this is not a blocker, but we, should definitely discuss this.

Issue #5031, opened 2021-08-01
Allow usage of private osix registry via the extension ui + gitpod file


ntziolis:
As of now the extension plugin seems hardwired to use the public open vsix registry. It would be greatly beneficial to also allow leveraging a private feed as well.

For starters it would be sufficient if it would be possible to override registry url via environment variables `GITPOD_VSIX_REGISTRY_URL` or similar, while defaulting to open vsix when not set.

This would allow users to host their own [openvsix instance ](https://github.com/eclipse/openvsx) while leveraging the public open vsix registry as upstream source.

akosyakov:
I don't think it is fixed. We have a proxy right now but it always works against open-vsx.org

corneliusludmann:
You can set the host of the OpenVSX instance in the `values.yaml` file:

https://github.com/gitpod-io/gitpod/blob/cfb528c5daf5115b666ec7404bfe25d9fc9598d6/chart/values.yaml#L137

Please give it a try!

akosyakov:
@corneliusludmann Do you mean in self hosted? I am not sure whether this issue about self hosting or allow on gitpod.io for some repos. @ntziolis could you clarify?

ntziolis:
We have enterprise customers that have a policy that only allows them to deploy extensions from sources they fully control. Meaning they will be running their own ovsx instance that largely just uses the public open vsx as upstream source but with an approval process attached.

This is not a requirement we can control and that we have seen in lots of enterprise customers in Germany.

jgallucci32:
This is a great request and there is tremendous value for self-hosted installed behind slow company firewalls. One advantage here is you can use something like Artifactory or Nexus to act as a remote repository for open-vsx.org to be able to cache on-demand at scale. I'm going to see if I can manually test this on 0.10.0 since the IDE replacement commands seem simple enough to put in manually. Can't wait to test this in the next release of self-hosted.

corneliusludmann:
Indeed, we needed to change the config. The new config looks like this:

https://github.com/gitpod-io/gitpod/blob/37435335ae86eef44eb9fee7f53f62b7e156c8e1/chart/values.yaml#L437-L462

I'll suggest a change at https://github.com/gitpod-io/website/pull/1058.

ntziolis:
@ghuntley First of let me say this requirement is born out of enterprise process and regulation stupidity.

Lets forget gitpod for a second and think node packages:
[...]

PR #6169, opened 2021-10-12
docs(contrib): add openvsx reverse nginx proxy config

ghuntley:
## Description

Archives this knowledge nugget over at https://github.com/gitpod-io/gitpod/issues/5031#issuecomment-940416029 into our contrib/ folder

## Related Issue(s)

https://github.com/gitpod-io/gitpod/issues/5031#issuecomment-940416029

## Release Notes

```release-note
NONE
```


corneliusludmann:
I'm afraid this will get orphaned pretty quickly if it sits around in the folder without further explanation. But I don't want to be the party pooper here.

/lgtm

PR #6208, opened 2021-10-14
Add cert-manager as command

### LINKED ISSUE #6207: [Installer]: add cert-manager command
[cert-manager](https://cert-manager.io) is a dependency for Gitpod. This will not necessarily need installing for every instance and certainly not be required to be installed more than once, so put as a separate command

mrsimonemms:
## Description

Creates a new `dependency cert-manager` command to install cert manager, based upon the GKE/Azure setup. The `dependency` was created as likely that there will be future third-party optional things to install that are outside the main configuration (external-dns is one such candidate).

## Related Issue(s)

Fixes #6207 

## How to test

`go run . deps cert-manager`

## Release Notes

```release-note
Create `dependency cert-manager` command
```

### DIFF
--- a/installer/cmd/dependency.go
+++ b/installer/cmd/dependency.go
@@ -1,0 +1,20 @@
+// Copyright (c) 2021 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package cmd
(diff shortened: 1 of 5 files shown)


csweichel:
I've filed the long overdue "prerequisite check" issue: https://github.com/gitpod-io/gitpod/issues/6210

PR #6270, opened 2021-10-18
[ide] Add IntelliJ desktop IDE support

### LINKED ISSUE #5641: Add local IDEs (vscode and jetbrains) options to the user settings
The user should be able to select what IDE they prefer and the options should be:
* vscode in the browser
* vscode in the desktop + vscode in the browser (fallback)
* pycharm + vscode in the browser (fallback)
* webstorm + vscode in the browser (fallback)
* intellij + vscode in the browser (fallback)
* ...

Additionally, the user should be able to select the version for each IDE. (e.g. desktop vscode latest + browser vscode stable as fallback). 

@gtsiolis Do you have any ideas on how to visually present these options?

cc @corneliusludmann @akosyakov

### LINKED ISSUE #5642: Create and use jetbrains' workspace images
As defined in https://github.com/gitpod-io/gitpod/issues/5641, users will be able to select Jetbrains IDEs as their preferred IDE. When that is the case, starting (a new) workspace should make Gitpod install and run the appropriate Jetbrains IDE backend and lead the user to a page similar to the following:

The page should automatically redirect to the Jetbrains link (for now, a _code with me_ link). And instead of "Jetbrains IDE" the button should mention the actual IDE name (e.g. Pycharm).

#### How can we detect if the IDE backend is ready and fetch the link?

One option would be to monitor the stdout of the process, but the recommended option is to either run the `status` method (see [internal discussion](https://gitpod.slack.com/archives/C02BRJLGPGF/p1632405270104700?thread_ts=1632328359.100900&cid=C02BRJLGPGF)) or do an HTTP call (see [here](https://gitpod.slack.com/archives/C02CBKZEVS8/p1632484978161600?thread_ts=1632483565.159900&cid=C02CBKZEVS8)).

#### Is the VS code web supposed to be running simultaneously in the same workspace?

Yes.

corneliusludmann:
## Description
This PR adds support for Intellij desktop IDEs.

**Feature Flag:** To be able to merge this into `main` without rolling this out to all users on the next deployment, the settings are only visible when you are an admin user. We could also introduce a dedicated feature flag for this. However, I think hiding it behind the admin flag is sufficient enough.

We could think about splitting this change into smaller PRs. However, it seems to me that parts of this change are not really viable and it would not make sense to cut them out.

**Werft Build Secret:** Because we should not leak the download URLs of the IntelliJ backends, we store them in the K8s secret `jetbrains-secrets` in the `werft` namespace and consume them during the build. Once the download URLs are public, we can remove this and add the URLs to the repo directly.

## To Do

The goal of this PR is to get a basic implementation merged into `main` and improve the implementation with smaller PRs. That means that most of the following to-do items should be moved to separate issues instead of fixing them in this PR.

We should focus on this PR to fix issues that would break something only.

- [ ] **Squash commits before merging!**
- [ ] Consider making the ws-manager-api `ide_image` filed composite as discussed here: https://github.com/gitpod-io/gitpod/issues/5642#issuecomment-926706162
- [ ] Think about it twice to make sure that we are backward compatible.
- [ ] Unit/integration tests?
- [x] Heartbeat plugin seems not to work. When I close the browser, after 5 minutes the workspace times out even when I type in the IntelliJ editor. We need to debug this. @atduarte → handled here: https://github.com/gitpod-io/gitpod/pull/6280
- [x] Use proper IDE icons in the settings.
- [ ] The password to join via Code With Me is `gitpod` (hard-coded). How should we handle this?
- [ ] Add `Stop Workspace` button?

## Related Issue(s)

Fixes #5642, fixes #5641

## How to test

**Note: The preferences settings are hidden. Only users with admin rights see the desktop IDE settings. If you are not the first user of the preview env, you need to run something like `$ leeway run components:make-admin`.**

https://clu-intellij-ide.staging.gitpod-dev.com/workspaces

1. Go to [`Settings` → `Preferences`](https://clu-intellij-ide.staging.gitpod-dev.com/preferences) and enable `Use Desktop IDE`:
image

2. Start a workspace. You should see the following:
image

3. Click on `Open IntelliJ IDEA IDE` and follow the instructions.
4. Click on `Open VSCode in Browser` and Open VSCode should open as usual.

## Release Notes

```release-note
NONE
```

Depending on the rollout strategy, we would probably don't have this in the release notes yet but add this to the PR that removes the feature flag.

## Meta
/no-cc

### DIFF
--- a/components/gitpod-protocol/src/workspace-instance.ts
+++ b/components/gitpod-protocol/src/workspace-instance.ts
@@ -212,4 +212,7 @@ export interface WorkspaceInstanceConfiguration {
 
     // ideImage is the ref of the IDE image this instance uses.
     ideImage: string;
+
+    // desktopIdeImage is the ref of the desktop IDE image this instance uses.
+    desktopIdeImage?: string
 }
--- a/components/server/src/ide-config.ts
+++ b/components/server/src/ide-config.ts
@@ -17,6 +17,7 @@ interface RawIDEConfig {
     ideVersion: string;
     ideImageRepo: string;
     ideImageAliases?: { [index: string]: string };
+    desktopIdeImageAliases?: { [index: string]: string };
 }
 const scheme = {
     "type": "object",
@@ -30,7 +31,11 @@ const scheme = {
         "ideImageAliases": {
             "type": "object",
             "additionalProperties": { "type": "string" }
-        }
+        },
+        "desktopIdeImageAliases": {
+            "type": "object",
+            "additionalProperties": { "type": "string" }
+        },
     },
     "required": [
         "ideVersion",
@@ -42,7 +47,7 @@ export interface IDEConfig {
     ideVersion: string;
     ideImageRepo: string;
     ideImageAliases: { [index: string]: string };
-
+    desktopIdeImageAliases: { [index: string]: string };
     ideImage: string;
 }
 
@@ -112,6 +117,9 @@ export class IDEConfigService {
                     ideImageAliases: {
                         ...raw.ideImageAliases,
                         "theia": ideImage,
+                    },
+                    desktopIdeImageAliases: {
+                        ...raw.desktopIdeImageAliases
                     }
                 }
             }
--- a/components/server/src/workspace/workspace-starter.ts
+++ b/components/server/src/workspace/workspace-starter.ts
@@ -323,6 +323,21 @@ export class WorkspaceStarter {
             }
         }
 
+        const useDesktopIdeChoice = user.additionalData?.ideSettings?.useDesktopIde || false;
+        if (useDesktopIdeChoice) {
+            const desktopIdeChoice = user.additionalData?.ideSettings?.defaultDesktopIde;
+            if (!!desktopIdeChoice) {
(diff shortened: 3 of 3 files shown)


loujaybee:
> I am interested in this test, can I take it? @corneliusludmann

Bear with us, it shouldn't be too long before we publish and send out any external communications on how you can join part of the beta / early release so you can experiment and give us some feedback 🙏

loujaybee:
With regards to the code with me link and splash page (screenshot below) I wonder how easily it would be to bring that step within the UX of Gitpod? It seems that the generation of the curl script is doing OS detection + pulling the ID, so we _could_ display that script directly on the workspace page to avoid the redirect. However, long term I think we'd probably want to bundle the logic into the companion app, but that's probably redundant if we eventually get rid of the Code With Me method. 



**Note:** Definitely not required within this PR, but thought I'd mention here.

loujaybee:
Looks like this also fixes #5641?

akosyakov:
@corneliusludmann we need to check that we can deploy changes independent from different teams, could you have a look at interfaces and think what will happen if another side is not updated yet?

@JanKoehnlein @csweichel It would be nice if we can coordinate on order of deployments. We could like last time workspace, meta, ide?

PR #6394, opened 2021-10-25
[werft/observability] Reuse preview certificate instead of requestin new ones

### LINKED ISSUE #6291: [preview-envs] avoid letsencrypt rate limit error
we sometimes see the following error in Werft:
> Failed to create Order: 429 urn:ietf:params:acme:error:rateLimited: Error creating new order :: too many certificates (5) already issued for this exact set of domains in the last 168 hours:
src: https://gitpod.slack.com/archives/C01KGM9EBD4/p1634246511455700?thread_ts=1634228500.454400&cid=C01KGM9EBD4
discussion: https://community.letsencrypt.org/t/5-already-issued-for-this-exact-set-of-domains-in-the-last-168-hours/153169

On https://crt.sh/?CN=gitpod-dev.com we can see wich certificates have been issues to us. Advanced Search -> mark "common name" -> search for "gitpod-dev.com". 

There I can see that we request certs for the same domains several times for grafana and prometheus:
image

Thus I assume that we're not running into the "max cert per week" rate limit for which we have an increase quota. 

We run into the "max certs for same domain" rate limit, which we could avoid, if we would not request certs redundantly.

ArthurSens:
## Description

Extends the preview certificate issuer to add monitoring-satellite subdomains while  https://github.com/gitpod-io/observability/pull/11 removes all certificates that are requested by the observability repository.

## Related Issue(s)

Fixes #6291, also needs https://github.com/gitpod-io/observability/pull/11 to make it work.

## How to test

Run the werft job, but add the appropriate annotations `with-observability` and `withObservabilityBranch=as/dns-rate-limit`

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.werft/build.ts
+++ b/.werft/build.ts
@@ -227,14 +227,12 @@ export async function build(context, version) {
     const destname = version.split(".")[0];
     const namespace = `staging-${destname}`;
     const domain = `${destname}.staging.gitpod-dev.com`;
-    const monitoringDomain = `${destname}.preview.gitpod-dev.com`;
     const url = `https://${domain}`;
     const deploymentConfig: DeploymentConfig = {
         version,
         destname,
         namespace,
         domain,
-        monitoringDomain,
         url,
         analytics,
         cleanSlateDeployment,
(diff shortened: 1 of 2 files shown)


ArthurSens:
Superseded by #6440

Issue #5231, opened 2021-08-17
reintroduce partial clone


akosyakov:
See https://github.com/gitpod-io/gitpod/pull/5152

We should reintroduce it in such way that it does not break [forking](https://github.com/gitpod-io/gitpod/issues/4914) and [blaming](https://github.com/gitpod-io/gitpod/issues/5110).

JanKoehnlein:
Though I am convinced improving performance is important, I think we should ATM focus on issues that increase stability instead. Feel free to bring this up at a later point in time. 

If you have any particular idea how to avoid the prior pitfalls, don't hesitate to add them to this issue. The clearer the path to a solution the more likely it will be scheduled.

akosyakov:
It seems the load timing is significantly high now. If it is the reason I think we should assign someone to investigate. I don't think the solution will appear othewise.

svenefftinge:
We should allow people to opt-in through the repository configuration on how a clone should happen.
I think it makes sense to allow general clone arguments. So people can use different kinds of clones. Also we might want to investigate if a shallow clone followed by an async inflation that happens when users are already on the workspace couel make sense. this could drastically reduce the size on the file system for prebuild snapshots as well.

gtsiolis:
An option to select partial clone, full clone, or clone using arguments could be something to consider adding in a projects settings area, next to features like project-level environment variables https://github.com/gitpod-io/gitpod/issues/4456. 💭

roboquat:
@csweichel: Issue scheduled in the workspace team (WIP: 0)

<details>

In response to [this](https://github.com/gitpod-io/gitpod/issues/5231#issuecomment-955704259):

>/schedule


Instructions for interacting with me using PR comments are available [here](https://git.k8s.io/community/contributors/guide/pull-requests.md).  If you have questions or suggestions related to my behavior, please file an issue against the [kubernetes/test-infra](https://github.com/kubernetes/test-infra/issues/new?title=Prow%20issue:) repository.
</details>

PR #6464, opened 2021-10-30
WIP Switch to git shallow clone and add unshallow feature

aledbf:
## Related Issue(s)

Fixes #

## Release Notes

```release-note
```


roboquat:
@aledbf: Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://git.k8s.io/community/contributors/guide/release-notes.md) to remove it.

<details>

</details>

roboquat:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please ask for approval from **aledbf** and additionally assign **jankoehnlein** after the PR has been reviewed.
You can assign the PR to them by writing `/assign @jankoehnlein` in a comment when ready.

<details open>

- **[components/content-service/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/content-service/OWNERS)**
- **[components/workspacekit/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/workspacekit/OWNERS)**

</details>

aledbf:
Another example running the script manually:

https://aledbf-unshallow.staging.gitpod-dev.com/#https://github.com/gitpod-io/openvscode-server


```
gitpod /workspace/openvscode-server $ du -sm
1487    .
```

```
gitpod /workspace/openvscode-server $ if git fetch --help | grep -c -- '--unshallow'; then git fetch --unshallow; else git fetch --depth 1000000000; fi
remote: Enumerating objects: 44275, done.
remote: Counting objects: 100% (39826/39826), done.
remote: Compressing objects: 100% (18082/18082), done.
remote: Total 37766 (delta 25267), reused 29453 (delta 18892), pack-reused 0
Receiving objects: 100% (37766/37766), 16.45 MiB | 18.00 MiB/s, done.
Resolving deltas: 100% (25267/25267), completed with 361 local objects.
```

```
gitpod /workspace/openvscode-server $ du -sm
1762    .
```

csweichel:
I very much like the unshallow idea - it nicely solves some of the issues we had seen.

I wonder though if we can up with more convenient ways to decide when to unshallow. The two things that shallow cloning broke were "forking" and "git blame":
- _forking_: when a user cannot write to a repo its likely they'll need to fork. We could naively always unshallow when a user clones a repo they cannot write to (server can decide - send an unshallow flag as part of the content initializer). Better yet, we might be able to hook into the forking process somehow and unshallow prior to the fork. To start, we could go with the first option.
- _git blame_: is much harder to detect. Short of placing something like a "git facade" in the workspace, or worse yet, relying on seccomp-notify based `execve` hacks, I cannot think of a good way to hook into that. The time-based mechanism does not help unfortunately, because there's no guarantee that a blame won't happen prior to the unshallow. Maybe @akosyakov has an idea?

The proposed time-based mechanism would certainly help reduce the load on the cluster, I fret it offloads the complexity onto the user though and does not solve the two cases above. Unshallowing the copy though IMHO is an awesome idea.

akosyakov:
> git blame: is much harder to detect. Short of placing something like a "git facade" in the workspace, or worse yet, relying on seccomp-notify based execve hacks, I cannot think of a good way to hook into that. The time-based mechanism does not help unfortunately, because there's no guarantee that a blame won't happen prior to the unshallow. Maybe @akosyakov has an idea?

If you mean whether there is a way to hook on VS Code level, unfortunately not. GitLens execs git directly. Besides we are going to have other IDEs where we don't have any control of core features.

csweichel:
> If you mean whether there is a way to hook on VS Code level, unfortunately not. GitLens execs git directly. Besides we are going to have other IDEs where we don't have any control of core features.

Can we assume that everyone is going to exec git? If so, we could add our own facade to git which triggers the unshallow when needed (e.g. prior to a git blame).

akosyakov:
> Can we assume that everyone is going to exec git? If so, we could add our own facade to git which triggers the unshallow when needed (e.g. prior to a git blame).

To my knowledge yes. Don't know what Jetbrains product will do, but maybe it is too early to worry. We can enable it first on some repos before making a default?

aledbf:
Reading the comments I think my goal is not clear:

- git repositories use shallow clone (no exceptions)
- by default, there is no unshallow.
- unshallow can be achieved:
    - using the `unshallowAfter` attribute in `.gitpod.yml`
    - running `git --unshallow` manually
    - defining a task in `.gitpod.yml` that runs `git --unshallow`
- the change in the behavior should be reflected in the docs

akosyakov:
@aledbf What you propose sounds good as the end goal, but I am not sure about usability consequences of rolling it for everybody at once. It would be better if we use it ourself for a month on gitpod-io/gitpod and gitpod/openvscode-server to learn.

JanKoehnlein:
I agree with Anton. We should start with opt-in for shallow cloning and enable it on gitpod and open-vscode-server to test ourselves.

csweichel:
We have tested shallow clones for about a month in production already, and have found two issues only: forks and blame. Also, considering the real-world performance improvements we've seen, this is a change that we don't want to delay for a month with little gain. The goal must be to make shallow clones the default, because only then do we reap their benefits.

That said, we'll want to have a good answer to the issues we've found before. Making this opt-out today would make this a breaking change; documentation would certainly help. Making this opt-in would be the low-risk path but would not give us the benefits we hope to achieve.

IMHO making this this feature opt-in makes little sense, because we're unlikely to see high uptake. I'd follow Alejandro's proposal with two alterations: 
- when a user cannot write a repo, we don't do a shallow clone. Because we know that forks and shallow clones don't go along so well, and users who cannot write to a repo are likely to fork, it wouldn't make sense to shallow clone here. 
- the time-based unshallow doesn't seem sensible to me because it doesn't solve any of the two issues we've identified. I'd get rid of it completely.

At the end it's mainly a communication issue, because with the outline above we don't expect adverse effects on most users. We'd need to make them aware of shallow clones, their implications, and how to remedy them. Also, worst case if we do see too many problems in practice, by having this a flag on the initializer (see below) we can quickly remove this behaviour.

### Technical details
Instead of providing a time-based unshallow mechanism, I reckon we'd want to make this part of the Git content initializer spec, i.e. add a `bool unshallow_clone = 7` to [`GitInitializer`](https://github.com/gitpod-io/gitpod/blob/9afad166d459c8ecd563fc0a61232142fe8413aa/components/content-service-api/initializer.proto#L46-L64). `server` would only set this flag in the _fork_ case outlined above.

geropl:
:+1: For not doing the time-based unshallow bc it adds user-facing complexity.

Regarding "opt-out vs opt-in": I agree that it does not make sense to go with opt-in only and wait for adoption. But this feature might break peoples workflows in a lot of ways, and just because we did not find more issues doesn't mean other will run into some. Especially as we cannot detect the general case and offer to unshallow.

> At the end it's mainly a communication issue

:100: Another options to solve the UX problem:
 - make the `git clone` time transparent to the user (startup screen?) and make it opt-in to generate adoption.
 - make it opt-out and have a banner (?) on workspaces/workspace start that explains the change

csweichel:
> 👍 For not doing the time-based unshallow bc it adds user-facing complexity.
> 
> Regarding "opt-out vs opt-in": I agree that it does not make sense to go with opt-in only and wait for adoption. But this feature might break peoples workflows in a lot of ways, and just because we did not find more issues doesn't mean other will run into some. Especially as we cannot detect the general case and offer to unshallow.

It's worth noting that imho Codespaces produces shallow clones, too - so it is possible to use this in production. And yes, there might be more issues we'll find down the road - as it is the case with all changes. In a month of exposing this to the real world, we've only found two issues though. I feel like we're very timid here in the face of (perceived) uncertainty.

Past experience shows that most users will not even notice the change, except that everything got a bit faster.

> 
> 💯 Another options to solve the UX problem:
> 
> * make the `git clone` time transparent to the user (startup screen?) and make it opt-in to generate adoption.
> * make it opt-out and have a banner (?) on workspaces/workspace start that explains the change

Instead of dwelling on the opt-in vs opt-out question, I'd rather we find solutions to the known problems and give this another shot:
- communicating the change: What's our usual way to communicate changes like this? Adding a notification to the IDE, the "What's New?" dialog, or the changelog?
- forks: solution is outlined above. We'd extend the GitInitializer with an `unshallow_clone` field that's set whenever the user cannot write to a repo.
- git blame: there's no clear solution yet, short of putting a facade in front of git which triggers the unshallow when someone calls `git blame`. However, we could also potentially reduce the magnitude of this issue using good comms (see first point).

In addition, we could consider adding a `shallowClone: false` field to the `.gitpod.yml` which lets users disable the shallow clone if it breaks something unexpectedly. In that case though, we should also introduce a metric that measures how many workspaces are started unshallow.

svenefftinge:
After further investigation, codespaces seems to exactly adjust to the case of a PR with more than one commit, in which case the clone is not shallow. Strange enough it is shallow when I open the codespace from the PR's branch 🤷 
I couldn't find a lot of information or complains from users about this behavior but it seems a little surprising tbh.

svenefftinge:
We should at least be able to do what codespaces do. I kind of dislike the inconsistent behavior, tbh.
What if we always fetch only the last 20 commits or so on WS start?

svenefftinge:
Sure, I have tested the following situations:

- main branch -> shallow clone
- PR with one commit -> shallow clone
- branch with on commit distance to default branch-> shallow clone
- PR with more than one commit -> deep clone (maybe inflated after the initial shallow clone, but it was immediately there, don't know)
- branch with more than one commits distance to default branch -> shallow clone (this was especially surprising, as it is technically the same git state as above)

aledbf:
> We can get repo size before clone (by using gitlab, github, bitbucket api) for small size repo we can use deep clone

That's an excellent idea

```
❯ curl -s https://api.github.com/repos/gitpod-io/gitpod|grep size
  "size": 53638,
``` 

(we need to check the rate-limiting)

svenefftinge:
I would prefer a more consistent behavior for users, so they are not confused that sometimes the clone is shallow and sometimes it is not. That is actually my concern with codespace's approach and having another condition (repo size) makes this worse. 
@aledbf what's your take on generally inflating the last 20 commits?

aledbf:
> what's your take on generally inflating the last 20 commits?

I think we will find users complaining about any change to the default behavior.
From my point of view, we should focus on the 80/20 and be **clear** in the docs **how you can `unshallow` the repository**. We could add an event/metric to track the real behavior/impact.

roboquat:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please ask for approval from **aledbf** and additionally assign **jankoehnlein** after the PR has been reviewed.
You can assign the PR to them by writing `/assign @jankoehnlein` in a comment when ready.

Associated issue: *#5231*

<details open>

- **[components/content-service/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/content-service/OWNERS)**
- **[components/workspacekit/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/workspacekit/OWNERS)**

</details>

roboquat:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please ask for approval from **aledbf** after the PR has been reviewed.

Associated issue: *#5231*

<details open>

- **[components/content-service/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/content-service/OWNERS)**
- **[components/workspacekit/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/workspacekit/OWNERS)**

</details>

aledbf:
I simplified the change to:
- Use shallow clone -> direct impact on prebuilds and opening the workspace
- Run unshallow as part of ring2 after 10s (just enough time to allow the start of the workspace)

Makes sense? This way there is no change in user workflows

csweichel:
I like @svenefftinge's idea to just fetch the last 20 commits, always. Most users would not see the difference and we would not need to add an exception for PRs.

aledbf:
> I like @svenefftinge's idea to just fetch the last 20 commits, always. Most users would not see the difference and we would not need to add an exception for PRs.

Done, depth is 20 now.

csweichel:
I was a bit unclear in my proposal I'm afraid. How about we:
1. in the content initializer always produce a shallow clone
2. in supervisor always fetch to a depth of 20 commits.

This should provide the following benefits:
- content init is quick which pays off when a workspace starts on a node with little turnaround. The IDE will open and then we pay for the fetch.
- few users would be surprised, because you have some history to go back on. E.g. (most) PR context's will just work, as will feature branches. All of this without needed to distinguish any cases.
- it's conceptually simple and easy to explain: Gitpod workspaces always come with a 20 commit history. If you want more, run `git --unshallow`.
- it opens the path for future optimisations, e.g. in the future we could choose not to unshallow in headless workspaces reducing the snapshot size of prebuilds.

I see one (very) minor downside of this approach compared to "never unshallow": on a busy node we're not saving as much CPU as we could, because every workspace will still do the fetch.

If we put this change in the documentation (the PR would need to link to the docs issue), and in the Changelog we'd also communicate the change clearly.

components/workspacekit/cmd/rings.go
```diff
@@ -546,6 +546,23 @@ var ring1Cmd = &cobra.Command{
 			}
 		}()
 
+		go func() {
```
csweichel, line 549: Just out of curiosity: why do this in workspacekit rather than supervisor? The latter has two benefits: 
1. we could forward the output to a terminal if we wanted to
2. it integrates nicer with full-workspace backup for which we'd need to move this to supervisor

csweichel: That said, I don't think that's a blocker and I'd be ok with having this go in the way it is

roboquat:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please ask for approval from **aledbf** and additionally assign **akosyakov** after the PR has been reviewed.
You can assign the PR to them by writing `/assign @akosyakov` in a comment when ready.

Associated issue: *#5231*

<details open>

- **[components/content-service/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/content-service/OWNERS)**
- **[components/supervisor/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/supervisor/OWNERS)**

</details>

csweichel:
Good catch @iQQBot. I've fixed that part now.
Opening something like https://aledbf-unshallow.staging.gitpod-dev.com/#https://github.com/gitpod-io/gitpod/commit/feaf4f6dc9610a3d01726984f7af0b3d030eefc5 works as exepcted.

csweichel:
@gitpod-io/engineering-ide please have a look as well, particularly towards the supervisor changes

csweichel:
> Can I test this in preview environment?
> 
> Currently show Your account has been blocked.
> 
> @csweichel

Sorry for that. That's the default behaviour in preview environments. I've unblocked your user.

csweichel:
> Thanks @csweichel looks good, in image builder stage, can we skip fetch to a depth of 20 commits ?

We could indeed do that, and we've discussed not fetching the history for any headless workspace (prebuilds, image-builds). However, this looks like a potential future optimisation. Considering that variance that would introduce, we'd like to see the effect of this change under load first.

corneliusludmann:
/approve

:+1:

roboquat:

This pull-request has been approved by: *<a href="https://github.com/gitpod-io/gitpod/pull/6464#issuecomment-961034328" title="Approved">corneliusludmann</a>*
To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please ask for approval from **aledbf** after the PR has been reviewed.

Associated issue: *#5231*

<details open>

- **[components/content-service/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/content-service/OWNERS)**
- ~~[components/supervisor/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/supervisor/OWNERS)~~ [corneliusludmann]

</details>

roboquat:

This pull-request has been approved by: *<a href="https://github.com/gitpod-io/gitpod/pull/6464#issuecomment-961034328" title="Approved">corneliusludmann</a>*, *<a href="https://github.com/gitpod-io/gitpod/pull/6464#issuecomment-961379710" title="Approved">csweichel</a>*

Associated issue: *#5231*

<details >

- ~~[components/content-service/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/content-service/OWNERS)~~ [csweichel]
- ~~[components/supervisor/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/supervisor/OWNERS)~~ [corneliusludmann]

</details>

roboquat:
LGTM label has been added.  <details>Git tree hash: c4004c1c50d6c1ebbc1981c68e7fd2068fb48d2f</details>

roboquat:

This pull-request has been approved by: *<a href="https://github.com/gitpod-io/gitpod/pull/6464#issuecomment-961034328" title="Approved">corneliusludmann</a>*, *<a href="https://github.com/gitpod-io/gitpod/pull/6464#issuecomment-961451924" title="LGTM">csweichel</a>*

Associated issue: *#5231*

<details >

- ~~[components/content-service/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/content-service/OWNERS)~~ [csweichel]
- ~~[components/supervisor/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/supervisor/OWNERS)~~ [corneliusludmann]

</details>

jeanp413:
Is there a way to tell gitpod to always do a deep clone, like for example defining an env variable in my gitpod settings?
I like to use the timeline view in vscode to check the latest changes in a specific file and now I have to unshallow the repo manually every time I open a new workspace :grimacing:

csweichel:
There isn't at the moment. It's not clear to me where such an "unshallow: always" setting should live: in the `.gitpod.yml` (repo scoped), or at the user-level (valid for all repos).

@svenefftinge wdyt?

svenefftinge:
For @jeanp413 use case it should be user-level. Would it make sense to build a small vscode extension that does this?
Doesn't even need to be gitpod specific.

jeanp413:
yeah, a vscode extension (or a contributed configuration option in the gitpod-web extension) that calls `git fetch --unshallow` automatically on startup should do the trick.

PR #6585, opened 2021-11-05
[supervisor] keep workspace alive via SSH

### LINKED ISSUE #4779: [local-app] heartbeating (keep workspace alive via SSH / whilst using non-browser IDE's)
We need to send heartbeats that user workspace does not timeout. We were considering to integrate on level of supervisor terminals, but implies that we have to integrate with supervisor all tools which capable to create terminal devices, i.e. it does not allow custom SSH servers. Instead we can integrate on system level using seccomp notify and intercept write and resize calls to terminal devices. It would also allow use cases when someone create new devices in the app under development, see for instance: [https://gitpod.slack.com/archives/C01KGM9DVRC/p1625443266089900](https://gitpod.slack.com/archives/C01KGM9DVRC/p1625443266089900)

**Plan:**

- *Plan A*: provide hearbeat signal based on pseudo-terminal device activity, e.g. using seccomp-notify
- *Plan B*: provide heartbeat signal based on supervisor terminal activity (requires IDE to integrate with supervisor)

iQQBot:
## Description

Provide heartbeat signal based on active ssh connection
This improve all local ide through ssh, like jetbranins gateway, vscode, and ssh self

## Related Issue(s)

Fixes #4779

## How to test

1. start a workspace
2. open gitpod-local-companion and connect into workspace ssh
3. close the browser
4. see heartbeat event on jaeger

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/supervisor/pkg/supervisor/ssh.go
+++ b/components/supervisor/pkg/supervisor/ssh.go
@@ -13,13 +13,15 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"golang.org/x/xerrors"
 
 	"github.com/gitpod-io/gitpod/common-go/log"
+	gitpod "github.com/gitpod-io/gitpod/gitpod-protocol"
 )
 
-func newSSHServer(ctx context.Context, cfg *Config) (*sshServer, error) {
+func newSSHServer(ctx context.Context, cfg *Config, gitpodService *gitpod.APIoverJSONRPC) (*sshServer, error) {
 	bin, err := os.Executable()
 	if err != nil {
 		return nil, xerrors.Errorf("cannot find executable path: %w", err)
@@ -38,9 +40,10 @@ func newSSHServer(ctx context.Context, cfg *Config) (*sshServer, error) {
 	}
 
 	return &sshServer{
-		ctx:    ctx,
-		cfg:    cfg,
-		sshkey: sshkey,
+		ctx:           ctx,
+		cfg:           cfg,
+		sshkey:        sshkey,
+		gitpodService: gitpodService,
 	}, nil
 }
 
@@ -49,6 +52,8 @@ type sshServer struct {
 	cfg *Config
 
 	sshkey string
+
+	gitpodService *gitpod.APIoverJSONRPC
 }
 
 // ListenAndServe listens on the TCP network address laddr and then handle packets on incoming connections.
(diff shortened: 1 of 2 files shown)


roboquat:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please assign **filiptronicek** after the PR has been reviewed.
You can assign the PR to them by writing `/assign @filiptronicek` in a comment when ready.

Associated issue: *#4779*

<details open>

- **[components/supervisor/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/supervisor/OWNERS)**

</details>

csweichel:
Interesting idea to use the liveness of the SSH session (which we can now gauge thanks to @aledbf's socket activation work). I wonder what would happen if someone just logged in via SSH and left it at that, but did not actually interact with the workspace.

akosyakov:
I am not sure about it, we don't want to keep hearbeating because of some ssh connection doing hearbeating. I think the only case when we are interesting if the ssh session has pty channel and there are some communications. With the suggested approach I think it would be rather costly for a user, i.e. it does not seem to stop the workspace in 2 mins if the last ssh terminal was terminated, does it?

components/supervisor/pkg/supervisor/ssh.go
```diff
@@ -125,23 +130,45 @@ func (s *sshServer) handleConn(ctx context.Context, conn net.Conn) {
...
+			timer := time.NewTimer(30 * time.Second)
+			select {
+			case <-timer.C:
+				continue
+			case <-done:
+				break
+			}
+		}
+		// send closed flag
+		_ = s.gitpodService.SendHeartBeat(ctx, &gitpod.SendHeartBeatOptions{
```
iQQBot, line 160: @akosyakov after session close, it will send closed flag
and will stop the workspace in 2 mins if the last ssh terminal was terminated
(no active HeartBeat)

akosyakov: What it is not the last session? I.e. there are other sessions still doing port forwarding which should not contribute to heartbeating. 

What if there is a terminal but it is not used 30 mins? In regards of tracking communications in terminals.

iQQBot: @akosyakov  OK, I will find a way to resolve this.

PR #6696, opened 2021-11-14
[installer]: set the number of replicas for deployments/statefulsets

### LINKED ISSUE #6357: [Installer]: add the number of replicas required for a deployment in the config
Currently, all of the deployments have a replica value of `1` (with a `todo` above them). This needs to be configurable.

Config change proposal:
 - Add a default number of replicas, which should default to `2`. Suggest putting this under `workspace.replicas.default`
 - Add in an optional map of key/value pairs under `workspace.components[$componentName].replicas`. This would then be extensible for any future configuration we need for components that cannot be inferred

MrSimonEmms:
## Description

Set the number of replicas on deployment and stateful set. This doesn't set the number of replicas in Helm dependencies (which largely can't be set anyway).

It introduces `replicas.default` to the workspace object in the config. This defaults to `2` unless set. The reason for adding like is to allow for adding a `components: map[string]struct{}` in future if we find that we need to control the number of replicas on individual components.

It's useful being able to control the number of replicas when debugging deployments.

## Related Issue(s)

Fixes #6357

## How to test

Deploy as normal, setting the value (or not) as desired

## Release Notes

```release-note
Set number of replicas for deployed components
```

### DIFF
--- a/installer/pkg/common/common.go
+++ b/installer/pkg/common/common.go
@@ -509,3 +509,17 @@ func RandomString(length int) (string, error) {
 	}
 	return string(b), nil
 }
+
+func Replicas(ctx *RenderContext) *int32 {
+	var defaultReplicas *int32
+
+	// Get the default replicas
+	if ctx.Config.Workspace.Replicas != nil && ctx.Config.Workspace.Replicas.Default != nil {
+		defaultReplicas = ctx.Config.Workspace.Replicas.Default
+	} else {
+		// Not set - default to 2
+		defaultReplicas = pointer.Int32(2)
+	}
+
+	return defaultReplicas
+}
--- a/installer/pkg/components/blobserve/deployment.go
+++ b/installer/pkg/components/blobserve/deployment.go
@@ -57,7 +57,7 @@ func deployment(ctx *common.RenderContext) ([]runtime.Object, error) {
 			},
 			Spec: appsv1.DeploymentSpec{
 				Selector: &metav1.LabelSelector{MatchLabels: labels},
-				Replicas: pointer.Int32(1),
+				Replicas: common.Replicas(ctx),
 				Strategy: common.DeploymentStrategy,
 				Template: corev1.PodTemplateSpec{
 					ObjectMeta: metav1.ObjectMeta{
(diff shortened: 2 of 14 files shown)


roboquat:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please ask for approval from **mrsimonemms** after the PR has been reviewed.

Associated issue: *#6357*

<details open>

</details>

csweichel:
/hold

I'm not sure this change makes sense this way - and I'm sorry for not reacting on the issue sooner.

1. The motivation that we need to add this "for debugging" does not warrant a config change. There are indeed cases where an increased replica-count of some components makes sense (e.g. SaaS prod), but for now the vast majority of installations will happily run with a replica count of 1.
2. Not all components can run with a replica count > 1. E.g. `ws-manager`, `ws-manager-bridge` or `ws-scheduler`.
3. Having this setting under `workspace` in the configuration seems odd. These replica counts do not affect workspaces, but rather other components which service workspaces.

Lastly - and as discussed before - configuration format like `components: map[string]struct{}` is a slippery slope; one that we're just trying escape with this installer effort. It quickly leads to a haphazard expansion of options, most of which make sense to us, but not to our users. We must design the configuration surface with our users and their use-cases in mind, not with our technical perspective on Gitpod.

I'd propose that all changes to the configuration surface in the future need to outline/describe the use-case they aim to support from a user perspective. If the user is us, there needs to be a discussion on the impact, potential confusion and way of documentation towards our users.

MrSimonEmms:
I'm sort of perfectly happy with making replicas an immutable property - those that need `1` stay as they are, the rest have whatever value they need (probably `2`). I must confess, I'd forgotten that certain components needed only `1` replica, but that's not the end of the world.

I don't particularly want to have a map of `component` options, but I think we should design a config surface that allows for that to be an option. I'm fully in agreement that it's only to be done where there's no alternative, and in a limited fashion. I certainly don't want to be spaffing all manner of config options in there.

Let's have a sync on this issue at some point this week (which is the main reason I did this - here's my understanding, lets use it as a springboard for discussion)

PR #7280, opened 2021-12-17
[installer]: grant option for disabling proxy load balancer

MrSimonEmms:
## Description

This will allow users to create their own ingress controller. An example for NGINX Ingress is shown.

The thinking behind this is due to the number of corporate users who have said that using `ServiceType: LoadBalancer` is not possible for them. If the ingress is disabled, the `ServiceType: ClusterIP` is used and the expectation is that the user will configure their own ingress controller

An example for NGINX Ingress is provided (in a gist)

This introduced the `meta` config object (similar to `workspace`) and `ingress.serviceType`, which defaults to `LoadBalancer`

## How to test

Create a deployment and set the config:

```yaml
components:
  proxy:
    serviceType: ClusterIP
```

Deploy your installation of Gitpod

Install ingress-nginx:

```shell
kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/main/docs/examples/psp/psp.yaml
helm upgrade \
    --atomic \
    --cleanup-on-fail \
    --create-namespace \
    --install \
    --namespace ingress-nginx \
    --reset-values \
    --repo https://kubernetes.github.io/ingress-nginx \
    --set defaultBackend.enabled=false \
    --set controller.extraArgs.enable-ssl-passthrough=true \
    --wait \
    ingress-nginx \
    ingress-nginx
```

Create your ingress controller:

```yaml
# Replace $DOMAIN with your domain

apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: gitpod
  annotations:
    kubernetes.io/ingress.class: nginx
    nginx.ingress.kubernetes.io/ssl-passthrough: "true"
spec:
  tls:
    - hosts:
        - "$DOMAIN"
        - "*.$DOMAIN"
        - "*.ws.$DOMAIN"
      secretName: https-certificates
  rules:
    - host: "$DOMAIN"
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: proxy
                port:
                  number: 443
    - host: "*.$DOMAIN"
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: proxy
                port:
                  number: 443
    - host: "*.ws.$DOMAIN"
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: proxy
                port:
                  number: 443
```

## Release Notes

```release-note
[installer]: grant option for disabling proxy load balancer 
```

### DIFF
--- a/installer/README.md
+++ b/installer/README.md
@@ -476,6 +476,21 @@ Kubernetes objects, such as your TLS certificate or connection secrets.
 kubectl create namespace gitpod
 ```
 
+# I need to add additional configuration to my load balancer - what do I do?
+
+By default, this will create a `proxy` service with `ServiceType: LoadBalancer`.
+If you want to specify your own ingress, you can do this easily by changing
+your Installer configuration:
+
+```yaml
+components:
+  proxy:
+    serviceType: ClusterIP
+```
+
+This will set the `ServiceType` to `ClusterIP` allowing you to create your own
+ingress controller.
+
 # Todo
 
 PRs/comments welcome
--- a/installer/pkg/components/proxy/objects.go
+++ b/installer/pkg/components/proxy/objects.go
@@ -32,7 +32,7 @@ var Objects = common.CompositeRenderFunc(
 				ServicePort:   PrometheusPort,
 			},
 		}, func(service *corev1.Service) {
-			service.Spec.Type = corev1.ServiceTypeLoadBalancer
+			service.Spec.Type = cfg.Config.Components.Proxy.ServiceType
 			service.Annotations["external-dns.alpha.kubernetes.io/hostname"] = fmt.Sprintf("%s,*.%s,*.ws.%s", cfg.Config.Domain, cfg.Config.Domain, cfg.Config.Domain)
 			service.Annotations["cloud.google.com/neg"] = `{"exposed_ports": {"80":{},"443": {}}}`
 		})(cfg)
--- a/installer/pkg/config/v1/config.go
+++ b/installer/pkg/config/v1/config.go
@@ -26,6 +26,11 @@ func (v version) Factory() interface{} {
 			Enabled:  false,
 			Passlist: []string{},
 		},
+		Components: &Components{
+			Proxy: &ProxyComponent{
+				ServiceType: corev1.ServiceTypeLoadBalancer,
+			},
+		},
 	}
 }
 func (v version) Defaults(in interface{}) error {
@@ -78,6 +83,8 @@ type Config struct {
 	Certificate ObjectRef `json:"certificate" validate:"required"`
 
 	ImagePullSecrets []ObjectRef `json:"imagePullSecrets"`
+
+	Components *Components `json:"components,omitempty"`
 
 	Workspace Workspace `json:"workspace" validate:"required"`
 
@@ -163,7 +170,7 @@ const (
 type ContainerRegistry struct {
 	InCluster *bool                      `json:"inCluster,omitempty" validate:"required"`
 	External  *ContainerRegistryExternal `json:"external,omitempty" validate:"required_if=InCluster false"`
-	S3Storage *S3Storage                 `json:"s3storage"`
+	S3Storage *S3Storage                 `json:"s3storage,omitempty"`
 }
 
 type ContainerRegistryExternal struct {
@@ -220,6 +227,14 @@ type Workspace struct {
 	Runtime   WorkspaceRuntime    `json:"runtime" validate:"required"`
 	Resources Resources           `json:"resources" validate:"required"`
 	Templates *WorkspaceTemplates `json:"templates,omitempty"`
+}
+
+type Components struct {
+	Proxy *ProxyComponent `json:"proxy,omitempty"`
+}
+
+type ProxyComponent struct {
+	ServiceType corev1.ServiceType `json:"serviceType,omitempty" validate:"k8s_service_type"`
 }
 
 type FSShiftMethod string
--- a/installer/pkg/config/v1/validation.go
+++ b/installer/pkg/config/v1/validation.go
@@ -8,10 +8,9 @@ import (
 	"fmt"
 
 	"github.com/gitpod-io/gitpod/installer/pkg/cluster"
-	"sigs.k8s.io/yaml"
-
 	"github.com/go-playground/validator/v10"
 	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/yaml"
 )
 
 var InstallationKindList = map[InstallationKind]struct{}{
@@ -39,6 +38,13 @@ var FSShiftMethodList = map[FSShiftMethod]struct{}{
 	FSShiftShiftFS: {},
 }
 
+var KubernetesServiceTypeList = map[corev1.ServiceType]struct{}{
+	corev1.ServiceTypeClusterIP:    {},
+	corev1.ServiceTypeNodePort:     {},
+	corev1.ServiceTypeLoadBalancer: {},
+	corev1.ServiceTypeExternalName: {},
+}
+
 // LoadValidationFuncs load custom validation functions for this version of the config API
 func (v version) LoadValidationFuncs(validate *validator.Validate) error {
 	funcs := map[string]validator.Func{
@@ -52,6 +58,10 @@ func (v version) LoadValidationFuncs(validate *validator.Validate) error {
 		},
 		"installation_kind": func(fl validator.FieldLevel) bool {
 			_, ok := InstallationKindList[InstallationKind(fl.Field().String())]
+			return ok
+		},
+		"k8s_service_type": func(fl validator.FieldLevel) bool {
+			_, ok := KubernetesServiceTypeList[corev1.ServiceType(fl.Field().String())]
 			return ok
 		},
 		"log_level": func(fl validator.FieldLevel) bool {


roboquat:

<details>

</details>

roboquat:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please ask for approval from **mrsimonemms** and additionally assign **jankeromnes** after the PR has been reviewed.
You can assign the PR to them by writing `/assign @jankeromnes` in a comment when ready.

<details open>

- **[installer/pkg/components/proxy/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/installer/pkg/components/proxy/OWNERS)**

</details>

MrSimonEmms:
@csweichel @corneliusludmann would be interested in your thoughts on this. We definitely need to do _SOMETHING_, but open to suggestions if this is the correct something

installer/README.md
```diff
@@ -458,6 +458,22 @@ Kubernetes objects, such as your TLS certificate or connection secrets.
...
+your Installer configuration:
+
+```yaml
+meta:
+  ingress:
+    enabled: false
+```
+
+This will set the `ServiceType` to `ClusterIP` allowing you to create your own
+ingress controller - click [here](https://gist.github.com/MrSimonEmms/480e88c149f6e725c727cc1f20b21474)
```
MrSimonEmms, line 474: Do we have a better, GP-owned place to put this?

aledbf:
> meta:
>  ingress:
>    enabled: false

@MrSimonEmms this config gives the impression that Gitpod uses Ingress, and that's not the case.
Maybe use `serviceType: ClusterIP`, saying the default is `LoadBalancer`? 

```yaml
meta:
  ingress:
    serviceType: ClusterIP
```

MrSimonEmms:
I've done the change @aledbf suggested. This will require some additional validation to ensure that the `serviceType` is acceptable - I'll do that if we agree that this is a change we want to make

installer/README.md
corneliusludmann: How about somewhere in this repo under the `contrib/` folder?

MrSimonEmms: That's a good idea

installer/README.md
```diff
@@ -458,6 +458,22 @@ Kubernetes objects, such as your TLS certificate or connection secrets.
...
+# I need to specify the load balancer - what do I do?
+
+By default, this will create a `proxy` service with `ServiceType: LoadBalancer`.
+If you need to specify your own load balancer, you can do this easily by changing
+your Installer configuration:
+
+```yaml
+meta:
+  ingress:
+    enabled: false
```
iQQBot, line 470: in code I only see `ingress` contain `serviceType` field, can't see `enabled` and it logic

iQQBot: Is there doc out of date?

installer/README.md
```diff
@@ -458,6 +458,22 @@ Kubernetes objects, such as your TLS certificate or connection secrets.
...
 ```
 
+# I need to specify the load balancer - what do I do?
+
+By default, this will create a `proxy` service with `ServiceType: LoadBalancer`.
+If you need to specify your own load balancer, you can do this easily by changing
+your Installer configuration:
+
+```yaml
+meta:
```
csweichel, line 468: Unlike `workspace` - which refers workspaces rather than the team - the term `meta` does not make sense from a user's perspective.

installer/README.md
csweichel, line 470: We should make the behaviour of this flag more obvious in its naming, e.g. directly refer to the type of the proxy service.

installer/README.md
MrSimonEmms: Fair point. Have you got a preferred name?

appcoders:
@MrSimonEmms 
Could be something wrong with the ingress? 
I tried your change but it leads to this https://github.com/gitpod-io/gitpod/issues/7324

MrSimonEmms:
@appcoders At the risk of being the worst type of engineer, "it works for me"

What version of the Installer are you using - `installer version | jq -r '.version'`?

appcoders:
Ingress can be ruled out, it is a imagebuild issue.
{"@type":"type.googleapis.com/google.devtools.clouderrorreporting.v1beta1.ReportedErrorEvent","error":"fork/exec /proc/self/exe: operation not permitted","level":"error","message":"failed to start ring0","ring":0,"serviceContext":{"service":"workspacekit","version":"commit-41879d65678692b46eac760932c3bb4b8c8714a2"},"severity":"ERROR","time":"2021-12-21T10:08:08Z"}

appcoders:
With the right distribution (Ubuntu 20.04 instead of Debian 10) everything works. To get the ingress working I added the following annotations:

```
annotations:
    kubernetes.io/ingress.class: nginx
    nginx.ingress.kubernetes.io/ssl-passthrough: "true"
    nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"
    nginx.ingress.kubernetes.io/ssl-redirect: "true"
    nginx.ingress.kubernetes.io/proxy-body-size: 1024m
```

I am not quite sure how big the proxy-body-size must be set, but 1G works. So, this could be used as a starting point.

aledbf:
@MrSimonEmms please do not include references or examples to Ingress configurations. 
By doing that, you implicitly are saying it is a supported configuration.

The main reason is related to the side effects of reload in ingress controllers like ingress-nginx and the impact on workspaces and reconnections.

aledbf:
@appcoders if you configure `nginx.ingress.kubernetes.io/ssl-passthrough: "true"` the annotation 
`nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"` has no effect because there is no NGINX upstream, only a [connection pipe](https://github.com/kubernetes/ingress-nginx/blob/main/internal/ingress/controller/tcp.go#L115)

installer/README.md
```diff
@@ -458,6 +458,22 @@ Kubernetes objects, such as your TLS certificate or connection secrets.
...
+# I need to add additional configuration to my load balancer - what do I do?
+
+By default, this will create a `proxy` service with `ServiceType: LoadBalancer`.
+If you want to specify your own ingress, you can do this easily by changing
+your Installer configuration:
+
+```yaml
+components:
+  proxy:
+    serviceType: ClusterIP
```
aledbf, line 470: Maybe add a validation checking that `serviceType` is LoadBalancer or ClusterIP only?
Both, `proxy` and `ws-proxy` are configured on ports 80 and 443

appcoders:
@aledbf You are right. This was a leftover from previous tests...

corneliusludmann:
Thanks, @MrSimonEmms for opening this pull request. I am currently trying to better understand the use case as well as the implications/risks for this change.

### Use Case

Currently, we create a service for the proxy of type `LoadBalancer`. With this change, the users have the possibility to change the proxy service to be of type `ClusterIP`. With that, you can create your own ingress controller (e.g. nginx, Traefik, etc.) that routes the traffic to the proxy service. This allows deploying Gitpod next to other apps with a single IP where the routing is based on the domain, path, etc.

### Limitations

The proxy still needs to have SSL certs. With this change, you still cannot move the SSL handling/termination to the ingress controller, right?

### Implications

Introducing this option in the installer indicates also that this is a valid use case we fully support somehow. That's why it's important to understand the implications and risks of this change. Or in other words, is the use of an ingress controller something we can support with confidence without regular testing? Or should we expect that there will be problems with it from time to time as users try it out and find cases that we haven't considered/tested? (Which ultimately leads to more variance that we need to test and support.)

@aledbf's comment above indicates that using a custom ingress controller could bring side-effect:

> The main reason is related to the side effects of reload in ingress controllers like ingress-nginx and the impact on workspaces and reconnections.

This means that we would probably have to test the case thoroughly with different ingress controllers if we want to officially support this.

### What's next?

Surely there are or will be users who have a need for this use case. However, at the moment there seems to be no one we know for sure (especially with the limitation mentioned above), right?

This pull request is, at its core, a small change in the load balancer that allows changing the type from `LoadBalancer` to `ClusterIP`. For me, at this point, it's not 100% clear if it is a use case that we can and want fully support _yet_.

I don't want to be a wet blanket. But especially because we can re-implement this change pretty quickly once we have the actual need from a customer for such a feature, I would vote for putting this pull request on hold for now. When we close this pull request now, we still have the code snippets available here once we decide that we need to implement (and test) this.

And if people from the community want to experiment with ingress controllers, they can still patch the output from the installer after the rendering pretty easily to change the service type of their installation.

What do you think?

### See also
- [Disucssion on Discord](https://discord.com/channels/816244985187008514/921322885572878367/921488009814224926)

appcoders:
@corneliusludmann  You're asking for use-cases? We are self-hosting on dedicated servers with no LB available. So, the Ingress solution is needed in this case. And certificate handling is not a problem as cert-manager handles renewal and the ingress uses the same cert secret as the proxy does.

MrSimonEmms:
@corneliusludmann tl;dr, I'm happy holding this but there is a legitimate use-case for this.

This is an oft-requested feature in the community, including from some potential large users. Technical reasons are somewhat sketchy for this, but I'm expecting that @appcoders's "we are hosting on servers with no LB available" is one such reason - I don't expect that this is the only one though.

I'm happy enough saying to people that post-processing the YAML is the recommended solution, but if we do that then we need to have better documentation on this - I would suggest as well that we would need to then have a "post-processable" output for the `render` function. This might mean that the output would be an array of objects with the apiVersion, kind, name and contents (ie, outputting the slice not the formatted data in [here](https://github.com/gitpod-io/gitpod/blob/main/installer/cmd/render.go#L128)). Expecting users to post-process may not be popular thought (I don't know, just a word of caution).

I know that the single-hosted instance would still require a cert at this stage - in my mind, that's a separate issue which, whilst related, I would like to do work on improving in the new year

Holding whilst we get different use-cases is a legitimate option.

@community - please add your use-cases down below I guess

appcoders:
After having issues bringing up the workspace at all, after removing
`nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"`
everything works again.

aledbf:
> I added the annotation again and everything works again.

Do you have the flag `--enable-ssl-passthrough` in the ingress-nginx deployment?
https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#ssl-passthrough

If not, it means `nginx.ingress.kubernetes.io/ssl-passthrough: "true"` is not going to do anything.

appcoders:
@aledbf This is how my ingress-nginx is installed:
```
helm upgrade --install ingress-nginx ingress-nginx \
  --repo https://kubernetes.github.io/ingress-nginx \
  --namespace ingress-nginx --create-namespace \
  --set defaultBackend.enabled=false \
  --set controller.extraArgs.enable-ssl-passthrough=true \
  --reset-values \
  --wait
```

MrSimonEmms:
Replaced with #7364

Issue #4456, opened 2021-06-09
Introduce project-level environment variables


gtsiolis:
### Problem to solve

The product currently supports only user-level environment variables set in user settings, where you use a scope syntax to specify for which projects the variable should be available to use. See [`/variables`](https://gitpod.io/variables) and [relevant docs](https://www.gitpod.io/docs/environment-variables).

With the upcoming [Teams & Projects](https://www.notion.so/Product-Roadmap-b9b5eac0a15147ac8d2dd25cf0519203#e80eb5330a0142568b8241c823e2055e) features, it will be able to set up a project and share this project configuration with other team members under a team.

### Proposal

Given there's also a plan to add project and team settings, it could make sense to add support for project-level environment variables or even team-level environment variables. This could also protect variables from being accidentally deleted or changed, and more as these could be read-only to non-owner members.

TBD (To be discussed)

ntziolis:
Any updates?

gtsiolis:
Hey @ntziolis! FYI, We've added this feature request on the roadmap[[1](https://github.com/gitpod-io/roadmap/projects/1)] for the next quarter. 🗺️

ntziolis:
Great to hear. This is currently blocking us from adopting the gitpod way of doing things.

Since we cannot use prebuilds (private npm registry), each dev currently creates a single pinned workspace per project and reuses it. This is because creating a new workspace per issue would take too long without prebuilds.

AlCalzone:
It would already be immensely helpful to be able to define (public) env variables in .gitpod.yml without any UI whatsover.

Our repo is using `yarn` to manage dependencies. The repo is configured to use a global `yarn` cache, but this does not work with gitpod prebuilds because only `/workspace` is persisted.

So we're changing the yarn config in the prebuilds to use a local (persisted) cache, but without extra steps does not carry over to workspaces spawned from the prebuilds. To avoid changing the config files, we need to set environment variables for each workspace. This can be done with startup scripts, but is currently very repetitive. With two startup tasks, we need to do this:
```
tasks:
  - name: "Task 1"
    before: |
      # set env variables for this terminal
      export YARN_CACHE_FOLDER=".yarn/cache"
      export YARN_ENABLE_GLOBAL_CACHE="false"

      # Persist the env variables for other terminals
      echo "export YARN_CACHE_FOLDER=\".yarn/cache\"" >> ~/.bashrc
[...]

PR #7295, opened 2021-12-17
Implement Project-level environment variables

jankeromnes:
## Description

Introduce Project-level environment variables:

- [x] [dashboard] Unify `project`-getting code into a `ProjectContext` helper
- [x] Implement a Variables page in Project Settings
- [x] Implement Project-level variables in `server` and `gitpod-db`
- [x] Expose Project-level variables in prebuilds
- [x] Expose Project-level variables in workspaces...
- [x] ...but also hiding certain variables in workspaces
- [ ] ...and censoring them in logs too (image builds, prebuild logs, etc. in order to mitigate accidental leaks of secrets) (EDIT: follow-up)
- [x] ~~Drive-by: Export `server`'s `censor` function via `gitpod-protocol` (in order to use it in `gitpod-db`)~~ (EDIT: Commit removed, censor code copy/pasted and specialized)
- [x] ~~Drive-by: Always use the actual function name (`arguments.callee.name`) in `checkUser` and `checkAndBlockUser` (instead of re-typing/copy-pasting the function name and sometimes making mistakes)~~ (EDIT: Commit removed again, `arguments.callee` is [deprecated](https://stackoverflow.com/questions/38435450/get-current-function-name-in-strict-mode) 🙄)

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/4456

## How to test

1. Log in to https://jx-project-variables.staging.gitpod-dev.com/projects
1. In your User Settings, create an environment variable called `USER` with value `user` and pattern `*/*`
1. Next, create a project
1. Navigate to Project Settings > Variables, and create a new variable (e.g. `PROJECT` with value `project`, visible in both Prebuilds and Workspaces)
1. In the Project repository, create a new branch containing a `.gitpod.yml` file that looks like this:

```yml
tasks:
  - init: echo "USER: $USER, PROJECT: $PROJECT"
    command: echo "USER: $USER, PROJECT: $PROJECT"
```

- Verify that, in the triggered prebuild (from the new branch created in 5.), only the `PROJECT` variable is defined (`USER` should still be undefined in the prebuild)
- When you open a workspace for that new branch, both `USER` and `PROJECT` should be defined (not in the printed Prebuild logs, but in the interactive command & also in live Terminals)

~~Open question: If you open a workspace for the new branch as a new user who doesn't have access to the Project (i.e. the repository is public, and you open a workspace for it, which gives you a prebuild, but you have no access to the Gitpod Project) -- should you be able to see the Project variables? (Prebuild/non-prebuild consistency says "yes", secrets say "no", maybe this should be configurable for each variable, see https://github.com/gitpod-io/gitpod/issues/4456#issuecomment-994953171)~~ (EDIT: Visibility is now configurable)

## Release Notes

```release-note
Introduce Project-level environment variables
```

## Documentation

/uncc


roboquat:

Associated issue: *#4456*

<details >

- **[components/gitpod-db/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-db/OWNERS)**

</details>

components/dashboard/src/components/Modal.tsx
```diff
@@ -41,7 +41,8 @@ export default function Modal(props: {
         return () => {
             window.removeEventListener('keydown', handler);
         };
-    }, []); // Empty array ensures that effect is only run on mount and unmount
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [props.onClose, props.onEnter]);
```
jankeromnes, line 45: Note: This was required because apparently state variables (like `name: string` and `value: string` in `AddVariableModal`) get "captured" in the `onClose` / `onEnter` lambdas, thus causing these functions to always use initial state values (i.e. empty string) instead of updated state values (i.e. whatever name/value a user typed).

The solution was to make this effect depend on `onClose` and `onEnter`, such that, if these functions change (they do change every time a captured state variable changes), the handler gets re-registered properly.

jankeromnes:
Okay, calling this ready for review 🎉

<table><tr><td>
Screenshot 2022-01-05 at 16 54 42
</td><td>
Screenshot 2022-01-05 at 16 54 55
</td></tr></table>

@geropl and/or @AlexTugarev, could you please take a look at the code changes?

@gtsiolis could you please take a look at the new Project Variables page UX?

(Happy to discuss / answer any question sync or async 🙏)

AlexTugarev:
Happy to have a closer look 👀

geropl:
Starting review now!

components/gitpod-db/src/typeorm/entity/db-project-env-vars.ts
```diff
@@ -0,0 +1,46 @@
...
+import { Transformer } from "../transformer";
+import { encryptionService } from "../user-db-impl";
+
+@Entity()
+// on DB but not Typeorm: @Index("ind_lastModified", ["_lastModified"])   // DBSync
+export class DBProjectEnvVar implements ProjectEnvVarWithValue {
+    @PrimaryColumn(TypeORM.UUID_COLUMN_TYPE)
+    id: string;
+
+    // projectId must be part of the primary key as otherwise malicious users could overwrite
```
geropl, line 19: The explanaition does not make sense to me: It has nothing to do with `TypeORM.save`, but only with the fact that wee need to track (and check) ownership of env vars, no? Having it as part of the primary key is a nice-to-have, 2nd line of defense and is fine with me.

That being said, my first thought was to re-use the existing table. The fact that it's called `...User...`, and it containing a field `userId` is a bit unfortunate, but could also be solved with migrations. WDYT? @jankeromnes

components/gitpod-db/src/typeorm/project-db-impl.ts
```diff
@@ -93,4 +106,55 @@ export class ProjectDBImpl implements ProjectDB {
             await repo.save(project);
         }
     }
+
+    public async setProjectEnvironmentVariable(projectId: string, name: string, value: string): Promise<void>{
+        const envVarRepo = await this.getProjectEnvVarRepo();
+        let envVar = await envVarRepo.findOne({ projectId, name, deleted: false });
+        if (envVar) {
+            envVar.value = value;
```
geropl, line 114: nit: this could be an `.update()` (without implicit existinance check again from typeorm). Not

components/gitpod-db/src/typeorm/project-db-impl.ts
```diff
@@ -7,22 +7,35 @@
...
+import { v4 as uuidv4 } from 'uuid';
+import { PartialProject, Project, ProjectEnvVar, ProjectEnvVarWithValue } from "@gitpod/gitpod-protocol";
+import { EncryptionService } from "@gitpod/gitpod-protocol/lib/encryption/encryption-service";
+import { censor } from '@gitpod/gitpod-protocol/lib/util/censor';
 import { ProjectDB } from "../project-db";
 import { DBProject } from "./entity/db-project";
-import { PartialProject, Project } from "@gitpod/gitpod-protocol";
+import { DBProjectEnvVar } from "./entity/db-project-env-vars";
+
+function censorEnvVar(variable: ProjectEnvVarWithValue): ProjectEnvVar {
```
geropl, line 18: IMO this is a clear "service" concern, and does not belong into the DB layer.
But maybe I'm missing the motivation here... And: why was this not necessary for Users? :thinking:

components/gitpod-protocol/src/util/censor.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Copyright (c) 2021 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+export function censor<T>(obj: T, k: keyof T): T {
```
geropl, line 7: IMO this should _not_ be part of protocol, but is a service (`server`) concern.

components/gitpod-protocol/src/protocol.ts
```diff
@@ -151,18 +151,29 @@ export interface UserFeatureSettings {
...
+    id?: string;
+    repositoryPattern: string; // DEPRECATED: Use ProjectEnvVar instead - https://github.com/gitpod-com/gitpod/issues/5322
+}
 export interface UserEnvVar extends UserEnvVarValue {
     id: string;
     userId: string;
     deleted?: boolean;
 }
 
+// DEPRECATED: Use ProjectEnvVar instead - https://github.com/gitpod-com/gitpod/issues/5322
```
geropl, line 176: This is a bit unexpected. I think I agree on the `repositoryPattern`. But user env-vars are still useful for user-wide stuff, right?

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -502,12 +502,21 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
...
         }
-        const envVars = this.userDB.getEnvVars(user.id);
+        const userEnvVars = this.userDB.getEnvVars(user.id);
+        let projectEnvVarsPromise: Promise<ProjectEnvVar[]> = Promise.resolve([]);
+        if (workspace.projectId) {
+            try {
+                await this.guardProjectOperation(user, workspace.projectId, "get");
+                projectEnvVarsPromise = this.projectsService.getProjectEnvironmentVariables(workspace.projectId);
+            } catch (error) {
+                log.warn({ userId: user.id }, "User doesn't have access to the Project, thus not loading Project environment variables:", error);
```
geropl, line 512: Is this really `warn`? IMO it should be `debug`: this might very well happen, if 1.) we have a prominten, public repo 2.) there is a Gitpod team for it only very few people are part of.

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -860,9 +869,19 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
                 throw err;
             }
 
+            let projectEnvVarsPromise: Promise<ProjectEnvVar[]> = Promise.resolve([]);
```
geropl, line 872: IMO it's the 3rd time I see this code. Could it be extracted? Would also eliminated the "noisy" let-if combination.

components/server/src/workspace/workspace-starter.ts
```diff
@@ -582,20 +584,26 @@ export class WorkspaceStarter {
...
-    protected async createSpec(traceCtx: TraceContext, user: User, workspace: Workspace, instance: WorkspaceInstance, mustHaveBackup: boolean, ideConfig: IDEConfig, userEnvVars?: UserEnvVarValue[]): Promise<StartWorkspaceSpec> {
+    protected async createSpec(traceCtx: TraceContext, user: User, workspace: Workspace, instance: WorkspaceInstance, mustHaveBackup: boolean, ideConfig: IDEConfig, userEnvVars: UserEnvVarValue[], projectEnvVars: ProjectEnvVar[]): Promise<StartWorkspaceSpec> {
         const context = workspace.context;
 
-        let allEnvVars: UserEnvVarValue[] = [];
-        if (userEnvVars) {
-            allEnvVars = allEnvVars.concat(userEnvVars);
+        let allEnvVars: EnvVarWithValue[] = [];
+        if (userEnvVars.length > 0) {
+            // DEPRECATED: Use ProjectEnvVar instead - https://github.com/gitpod-com/gitpod/issues/5322
```
geropl, line 592: Really? Same as in procotol.ts.

components/gitpod-db/src/typeorm/entity/db-project-env-vars.ts
jankeromnes: > The explanaition does not make sense to me

Ah, I simply copied it over from here: 😅

https://github.com/gitpod-io/gitpod/blob/d114f037a1c34df5206bb93fc9fe48481f28bfd7/components/gitpod-db/src/typeorm/entity/db-user-env-vars.ts#L19-L24

Happy to amend or remove it. 👍

> That being said, my first thought was to re-use the existing table.
> [...]
> Update: ah, now I think I see the indent - (partly) deprecation of the UserEnvVar table. Nvm then.

Yes, (part) deprecation of UserEnvVar is indeed the intent 👍

components/gitpod-db/src/typeorm/project-db-impl.ts
jankeromnes: Agree, this is a new practice: Instead of passing around potentially secret values through the entire meta code base (where it has many opportunities to get logged accidentally), I prefer to pass around a "handle" (which can get logged without consequence) and only use the actual value where strictly necessary (you'll notice that the dashboard and the API can never read values, only set or delete them).

This is still far from perfect, and in the future we'll likely want to use End-To-End encryption and Kubernetes secrets (thus passing a no-value "handle" further down to workspaces) but with this approach, you now have just a single place in meta that reads values (WorkspaceStarter.createSpec) as opposed to many places.

components/gitpod-protocol/src/util/censor.ts
jankeromnes: What is the correct place to share code between DB and server, if not protocol?

I was imitating other utils like `deep-partial.ts`, but I'm also happy to duplicate the `censor` function and have it just in server and in DB.

components/gitpod-protocol/src/protocol.ts
jankeromnes: Agree, maybe I was a bit heavy-handed here. My point of view is that UserEnvVar is 98% `repositoryPattern` and 2% just `name` and `value`, hence we're deprecating most of UserEnvVar.

Also, I'm not quite sure what a User-wide variable concept looks like with `repositoryPattern` removed. E.g., maybe you'll want a few variables _everywhere_ (e.g. `EDITOR=vim`), but for personal keys, I think it makes more sense to have "User-specific Project variables" (i.e. I go to the Gitpod project, and it shows "Project variables for everyone in the Team", as well as "Project variables just for my User").

That's sort of why I wanted to mark UserEnvVar as deprecated until it has been refactored, but I'm happy to mark just `repositoryPattern` for now.

components/server/src/workspace/gitpod-server-impl.ts
jankeromnes: Good catch, 100% agree. I wanted to make this more visible, as we're still working on properly segregating "public access" vs "Team/Project access" for public repos, but in this location the segregation is done properly so there is probably no need to be loud.

components/gitpod-protocol/src/util/censor.ts
jankeromnes: Removed `censor` from protocol again (instead the code is now copy/pasted and specialized in ProjectDB).

components/gitpod-db/src/typeorm/entity/db-project-env-vars.ts
geropl: > Happy to amend or remove it. +1

Let's compress it to:
```
// `projectId` is part of the PK for safety reasons: This way it's impossible that a user (maliciously or by chance) sends us an env-var that has the same PK (`id`) as the env var from anther user.
```

components/gitpod-protocol/src/protocol.ts
geropl: > User-wide variable concept

Hm. IMO the `repositoryPattern` still offers a way of flexibility I don't get otherwise, no?
If there's a service that I want to use in the majority of my projects, and I have a personal access token for: Do you see that on project level? I had a user-level view with repo selection (like GitHub) in mind.

But anyway, I'd be in favor of deprecating `repositoryPattern` only for now, and sort the rest out later.

jankeromnes: > If there's a service that I want to use in the majority of my projects, and I have a personal access token for: Do you see that on project level?

Good point, this use case doesn't seem to make sense on Project level.

> I had a user-level view with repo selection (like GitHub) in mind.

Precise repository selection sounds like a good option (and isn't affected by https://github.com/gitpod-com/gitpod/issues/5322 like `repositoryPattern`).

However, I wonder how common/important this particular use case is (i.e. whether we'll want to implement a full User-wide multi-reposelection per variable, or if we can live without this feature). In any case, we don't need to make this decision now. 😇

components/gitpod-db/src/typeorm/project-db-impl.ts
geropl: Ah, now I see what you want to achieve. I mis-read the types, and got confused by `censor`: Basically the `SELECTED` fields should match the type (precisely).

I recommend changing the function name then, maybe prefix with `filter`, or so.

Alternatively, you can "fix" the `select` in the first place. Here's an example: https://github.com/gitpod-io/gitpod/blob/7abe78c91792c18b0c22103c890b25ea3a4cdf05/components/gitpod-db/src/typeorm/workspace-cluster-db-impl.ts#L57

gtsiolis:
Looking at this now! 👀

components/dashboard/src/projects/ProjectVariables.tsx
```diff
@@ -0,0 +1,126 @@
...
+            <div>
+                <h4>Name</h4>
+                <input autoFocus className="w-full" type="text" name="name" value={name} onChange={e => setName(e.target.value)} />
+            </div>
+            <div className="mt-4">
+                <h4>Value</h4>
+                <input className="w-full" type="text" name="value" value={value} onChange={e => setValue(e.target.value)} />
+            </div>
+            <div className="mt-4">
+                <CheckBox title="Visible in Prebuilds?" desc="All Project Variables are visible in Prebuilds" checked={true} disabled={true} />
```
gtsiolis, line 112: **issue**: Including always disabled checkboxes does not help much, right? I'd suggest to drop this option and replace it with a warning alert. What do you think?

components/dashboard/src/projects/ProjectVariables.tsx
```diff
@@ -0,0 +1,126 @@
...
+        <div className="mb-2 flex">
+            <div className="flex-grow">
+                <h3>Project Variables</h3>
+                <h2 className="text-gray-500">Manage environment variables for your project.</h2>
+            </div>
+            {envVars.length > 0 && <button onClick={() => setShowAddVariableModal(true)}>Add Variable</button>}
+        </div>
+        {envVars.length === 0
+            ? <div className="bg-gray-100 dark:bg-gray-800 rounded-xl w-full py-28 flex flex-col items-center">
+                <h3 className="text-center pb-3 text-gray-500 dark:text-gray-400">No Environment Variables</h3>
```
gtsiolis, line 52: **suggestion**: I'd be nice to keep the description of the empty state, following the component pattern used elsewhere. WDYT?

| BEFORE | AFTER |
|-|-|
| Screenshot 2022-01-11 at 12 13 56 AM (2) | Screenshot 2022-01-11 at 12 13 48 AM (2) |

components/dashboard/src/projects/ProjectVariables.tsx
```diff
@@ -0,0 +1,126 @@
...
+            </div>
+            <div className="mt-4">
+                <h4>Value</h4>
+                <input className="w-full" type="text" name="value" value={value} onChange={e => setValue(e.target.value)} />
+            </div>
+            <div className="mt-4">
+                <CheckBox title="Visible in Prebuilds?" desc="All Project Variables are visible in Prebuilds" checked={true} disabled={true} />
+            </div>
+            <div className="mt-4">
+                <CheckBox title="Visible in Workspaces?" desc="Choose whether this Variable should be visible in Workspaces" checked={!censored} onChange={() => setCensored(!censored)} />
```
gtsiolis, line 115: **suggestion**: I'm probably wrong but instead of having a disabled checbox, could we reverse the defaults here and include a warning alert at the top of the modal? Does this make sense? Happy to sync[[1](https://calendly.com/gtsiolis/coworking)] to discuss more about this in depth.

| BEFORE | AFTER |
|-|-|
| Screenshot 2022-01-11 at 12 22 24 AM (2) | Screenshot 2022-01-11 at 12 21 34 AM (2) |

gtsiolis (approved):
Great work, @jankeromnes! 🔮 

I couldn't test this thoroughly as prebuilds are broken for this preview environment. 😭 

```sh
cannot initialize workspace: cannot use configured storage: Get "http://minio.staging-jx-project-variables.svc.cluster.local:9000/gitpod-user-3880fb66-b962-4374-841d-b08c07d7b9e7/?location=": dial tcp 10.100.15.120:9000: connect: connection refused; last backup failed: workspace does not exist. Please contact support if you need the workspace data.
```

This took longer than expected but left some comments below regarding UX.

Approving to unblock merging but holding in case we could address any of the UX comments below.

/hold

roboquat:
LGTM label has been added.  <details>Git tree hash: c9a59acf97ba5ac32a2dbc1fb74e36f18496824d</details>

roboquat:

This pull-request has been approved by: *<a href="https://github.com/gitpod-io/gitpod/pull/7295#pullrequestreview-847742902" title="Approved">gtsiolis</a>*

Associated issue: *#4456*

<details >

- ~~[components/dashboard/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/dashboard/OWNERS)~~ [gtsiolis]
- ~~[components/gitpod-db/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-db/OWNERS)~~ [gtsiolis]
- ~~[components/gitpod-protocol/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-protocol/OWNERS)~~ [gtsiolis]
- ~~[components/server/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/server/OWNERS)~~ [gtsiolis]

</details>

components/dashboard/src/projects/ProjectVariables.tsx
```diff
@@ -0,0 +1,126 @@
...
+            <div className="flex-grow">
+                <h3>Project Variables</h3>
+                <h2 className="text-gray-500">Manage environment variables for your project.</h2>
+            </div>
+            {envVars.length > 0 && <button onClick={() => setShowAddVariableModal(true)}>Add Variable</button>}
+        </div>
+        {envVars.length === 0
+            ? <div className="bg-gray-100 dark:bg-gray-800 rounded-xl w-full py-28 flex flex-col items-center">
+                <h3 className="text-center pb-3 text-gray-500 dark:text-gray-400">No Environment Variables</h3>
+                <button onClick={() => setShowAddVariableModal(true)}>New Variable</button>
```
gtsiolis, line 53: **thought**: It could be interesting to surface inherited user-specific environment variables below project-specific environment variables. Added follow-up https://github.com/gitpod-io/gitpod/issues/7532 to continue the discussion about this. Certainly out of the scope of this PR. 💭

components/dashboard/src/projects/ProjectVariables.tsx
```diff
@@ -0,0 +1,126 @@
...
+        try {
+            await getGitpodService().server.setProjectEnvironmentVariable(props.project.id, name, value, censored);
+            props.onClose();
+        } catch (err) {
+            setError(err);
+        }
+    }
+
+    return <Modal visible={true} onClose={props.onClose} onEnter={() => { addVariable(); return false; }}>
+        <h3 className="mb-4">Add Variable</h3>
```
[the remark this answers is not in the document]
jankeromnes: Editing is a security problem here. Team variables should be write/delete-only (i.e. no "view value", no "edit value").

Example: We have a team of 500 developers, and our project needs access to a critical service. Only I or a few people are allowed to manage tokens for that service.
- If variables are write-only, I can set up the token for our project, and be reasonably confident that others will not try to steal/re-use this token (I mean, they could, e.g. by "hacking" a Prebuild script to get the value, but someone might notice, and I generally trust my colleagues not to try and "hack" our project all the time 😝)
- If variables are read/write, anyone can go the Variables page, copy the token, and do whatever they want with it

components/dashboard/src/projects/ProjectVariables.tsx
```diff
@@ -0,0 +1,126 @@
...
+            <div className="mt-4">
+                <CheckBox title="Visible in Workspaces?" desc="Choose whether this Variable should be visible in Workspaces" checked={!censored} onChange={() => setCensored(!censored)} />
+            </div>
+            {censored && <div className="mt-4">
+                <InfoBox><strong>Never log this value</strong>: Prebuild logs are always visible in Workspaces.</InfoBox>
+            </div>}
+        </div>
+        <div className="flex justify-end mt-6">
+            <button className="secondary" onClick={props.onClose}>Cancel</button>
+            <button className="ml-2" onClick={addVariable}>Add Variable</button>
```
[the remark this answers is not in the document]
jankeromnes: Thanks for caring about consistency! 💯 But see above about "Update" being a security problem.

components/dashboard/src/projects/ProjectVariables.tsx
```diff
@@ -0,0 +1,126 @@
...
+                        <ItemField>In Workspaces?</ItemField>
+                        <ItemField></ItemField>
+                    </Item>
+                    {envVars.map(variable => {
+                        return <Item className="grid grid-cols-5 items-center">
+                            <ItemField>{variable.name}</ItemField>
+                            <ItemField>****</ItemField>
+                            <ItemField>Visible</ItemField>
+                            <ItemField>{variable.censored ? 'Censored' : 'Visible'}</ItemField>
+                            <ItemField className="text-right"><button onClick={() => deleteEnvVar(variable.id)}>x</button></ItemField>
```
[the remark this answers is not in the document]
jankeromnes: > **issue**: This one hurt! 🙈

Ah, sorry 😅 I know you wouldn't like this one 🙈

> **thought**: Eventually it should be possible to also update (edit) the environment variable, right?

Well, not really, due to the security issue explained above. Also, I don't really like context menus, or anything that "hides" common actions behind some interactive UI thing (e.g. dropdown, collapse, etc).

I must confess I copied the "Environment Variables" page in CircleCI's Project Settings, and I really liked the simplicity of their `x` button to remove variables. Maybe we could make ours less "buttony" somehow? 😋

Screenshot 2022-01-11 at 10 10 34

components/dashboard/src/projects/ProjectVariables.tsx
jankeromnes: I agree that it's a "placebo" checkbox, but in this case I felt like it really helps clarify where variables are exposed and what you can change about it.

Alternatives like having a single checkbox, and a long complicated text to explain the full situation, seem less intuitive, i.e. it requires more upfront study work & cognitive overheard for users before they can fully grasp this. A "fake" checkbox that you immediately understand seemed like the lesser evil, but I'd be curious if you have ideas to make users immediately understand the situation without needing a "fake" checkbox?

components/dashboard/src/projects/ProjectVariables.tsx
jankeromnes: Agree, and many thanks for the proposed copy! I also wanted to have a description, but I didn't know what to write in there 🙈

components/dashboard/src/projects/ProjectVariables.tsx
jankeromnes: Thanks for the visual concept! It made me change my mind about the "fake" checkbox, as I now feel the "AFTER" design could work.

- The Alert text might need a few iterations before we can convey the full story clearly / concisely / unambiguously 😋
- I like "Mask" (better than "Censor" at least) and I also like "Hide"

components/dashboard/src/projects/ProjectVariables.tsx
gtsiolis: Thanks, Jan! Let's do this.

60taoh

components/dashboard/src/projects/ProjectVariables.tsx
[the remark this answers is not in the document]
gtsiolis: Aha! 💡

components/dashboard/src/projects/ProjectVariables.tsx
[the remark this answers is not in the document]
gtsiolis: No worries, @jankeromnes! 🧡

We could go with a simple `X` button but I'd suggest to  move forward with reusing the same dropdown pattern used elsewhere and keep the `X` button for simple dismissible actions like closing a modal or later on dismissing alters. We're already using this with one action in `/integrations`.

In any case, going with the common default and hover states (no background, gray background) would be also great. WDYT? 💭 

Thanks again for helping get this closer to shipping! 🚢

components/gitpod-db/src/typeorm/project-db-impl.ts
jankeromnes: Thanks! I really like the `.select` suggestion, but feel it would make many `repo.find` look much more complicated than they really are.

If `censorEnvVar` doesn't work, I'll rename the function to `toProjectEnvVar` (similar to other `toShape` functions in other DB impls).

I personally would find the word `filter` confusing here, as to me `filter` is strongly associated with JS list filtering (i.e. I would automatically assume a `filterProjectEnvVar` function takes a list of `ProjectEnvVar`s and returns a sub-list based on some criteria -- which is not the case here).

geropl: `toProjectEnvVar` is fine with me as well. :+1:

components/dashboard/src/projects/ProjectVariables.tsx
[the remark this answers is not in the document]
geropl: > Editing values is a security problem here

I don't see how this relates to security; besides obscuring access to environment variables. :roll_eyes: See comment below.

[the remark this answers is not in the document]
jankeromnes: From my perspective:

- Making values easily visible by anyone in the UI is significantly different to having values only present in non-interactive systems (where extracting the value is significantly more complicated, and leaves traces).
- Every single CI system I know that allows setting such values also has a write-only / no-read UI. I think there must be good reasons for that, and I don't see why Gitpod should be different.
- We've had at least one customer request that values should not be visible in the UI after being set.

> See comment below.

Sorry, which comment are you referring to? (Maybe you're still writing it, or maybe it wasn't submitted?)

components/dashboard/src/projects/ProjectVariables.tsx
geropl: I'm in favor of the "warning", too.

Two questions come to my mind, though, when reading the new text:
 - `are always visible`: to whom? A: "all project members, and everyone with read access to your repository"
 - `mask`: What does this mean? A: "< long discussion about the trade-offs >. Security-wise: nothing (yet)."
 
I fear with this "double meaning" w.r.t. visibility/masking we're facing users with yet another concept - which is **not sound**: The whole idea of ["team members are not able to see environment variables"](https://github.com/gitpod-io/gitpod/pull/7295#discussion_r783096654) and "masking" environment variables breaks due to:
 - prebuild logs (not filtered atm)
 - all tool that access your env vars during prebuild that store them somewhere in the (persisted) file system
 - prebuilds being shared between all Gitpod users (atm)

IMO we all agree that this is a hard and delicate topic. In the spirit of moving forward with this I propose we split this into:
 - introduce Project env vars with only one level of visibility (all or nothing, present or not), allow editing in the UI (because you have access to it anyway = same level of trust). All while being _very clear_ of the impact with the warning introduced above. Including that all random viewers of the project might get a workspace based on one of your prebuilds - incl. all keys in there. :upside_down_face: 
 - put the discussion about "private env vars" on hold: IMO the key thing we'd need to do here is to allow for prebuilds to have the same visibility level like the env var: project-/team-only.
 
I understand - and recall your arguments @jankeromnes - that this would reduce the usefulness of this features (to private repos, or "keys" that you do not care about being shared with the world, basically). Still, it's the only way forward I see atm that feels sound and safe in a way I could explain it to a user with a good feeling.

/cc @JanKoehnlein

components/dashboard/src/projects/ProjectVariables.tsx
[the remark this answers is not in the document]
geropl: > Every single CI system I know that allows setting such values also has a write-only / no-read UI. I think there must be good reasons for that, and I don't see why Gitpod should be different.

Fair point. I mainly wanted to argue against the "false sense of security" (see the [comment below](https://github.com/gitpod-io/gitpod/pull/7295#discussion_r783113463)).

roboquat:
New changes are detected. LGTM label has been removed.

components/gitpod-protocol/src/protocol.ts
jankeromnes: Made the deprecation warnings more specific to `repositoryPattern`. ✅ Thanks for the feedback!

components/dashboard/src/projects/ProjectVariables.tsx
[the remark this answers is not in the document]
jankeromnes: Decided to remain consistent here, and replaced the awkward `X` button with our usual context menu (with only one item, `Delete`). ☺️

components/dashboard/src/projects/ProjectVariables.tsx
geropl: After dwelling a bit more about this - and re-reading my comment above - I might have overreacted a bit. :roll_eyes:  My apologies @jankeromnes  if that came across to harsh.

Maybe we compress the ramblings above to:
 - top warning: "Project specific environment variables might be (indirectly) accessible for everyone with read access to your repository"
 - warning underlines: "Secrets might leak if they 1.) are printed to the log, or 2.) persisted to the file system"
 - Option: "exclude from regular workspaces", checked by default (optional: showing a 2nd warning if unchecked? :innocent: )

Also, happy to discuss in a (quick) sync round tomorrow so we can ship this features! :muscle:

JanKoehnlein: Let's add warnings and enable users who want to use it.

jankeromnes: Many thanks @geropl for always providing consistently valuable feedback. 🙏 I feel like our discussions always remain constructive, and by working towards a consensus, the end result always becomes much better than what I could have envisioned initially. 💯

Thanks also for your helpful feedback on the warning text -- I'll be sure to reword it accordingly. 🎯

And I also agree on what you call the "false sense of security". We shouldn't ship a feature that "traps" users by encouraging them to do unsafe things, so it's important to get this right.

I'll make the proposed adjustments today, and then see whether the "exclude from regular workspaces" feature feels good to go as is (subject to review), or if it should be split into another PR (so that the non-controversial foundation and refactorings in this PR can be merged as soon as possible, while also giving us a few more cycles to improve the "exclude" feature/UX).

jankeromnes:
Okay, new UX dropped:

| No Variables | 1 Warning | 2 Warnings |
| --- | --- | --- |
| Screenshot 2022-01-13 at 17 53 18 | Screenshot 2022-01-13 at 17 53 42 | Screenshot 2022-01-13 at 17 54 13 |
| Screenshot 2022-01-13 at 17 53 01 | Screenshot 2022-01-13 at 17 57 08 | Screenshot 2022-01-13 at 17 54 38 |

How do you feel about this, @geropl and @gtsiolis? Almost ready to merge, or could the "Hide" feature use some more work? 😊

gtsiolis:
Taking another look at the UX changes here! 👀

geropl:
@jankeromnes Thx for putting in all this effort. Love the explicitness of the new messages! :heart: 

/lgtm
/hold

So @gtsiolis can review as well

roboquat:
LGTM label has been added.  <details>Git tree hash: 34d504ab9541adf0333b4aec043af2382ba2786d</details>

roboquat:

This pull-request has been approved by: *<a href="https://github.com/gitpod-io/gitpod/pull/7295#issuecomment-1013208444" title="LGTM">geropl</a>*, *<a href="https://github.com/gitpod-io/gitpod/pull/7295#pullrequestreview-847742902" title="Approved">gtsiolis</a>*

Associated issue: *#4456*

<details >

- ~~[components/dashboard/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/dashboard/OWNERS)~~ [geropl,gtsiolis]
- ~~[components/gitpod-db/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-db/OWNERS)~~ [geropl,gtsiolis]
- ~~[components/gitpod-protocol/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-protocol/OWNERS)~~ [geropl,gtsiolis]
- ~~[components/server/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/server/OWNERS)~~ [geropl,gtsiolis]

</details>

geropl:
/hold

Last got lost somehow :grimacing:

components/dashboard/src/projects/ProjectVariables.tsx
```diff
@@ -0,0 +1,132 @@
...
+            props.onClose();
+        } catch (err) {
+            setError(err);
+        }
+    }
+
+    return <Modal visible={true} onClose={props.onClose} onEnter={() => { addVariable(); return false; }}>
+        <h3 className="mb-4">Add Variable</h3>
+        <div className="border-t border-b border-gray-200 dark:border-gray-800 -mx-6 px-6 py-4 flex flex-col">
+            <AlertBox><strong>Project variables might be accessible by anyone with read access to your repository.</strong><br/>Secret values can be exposed if they are printed in the logs, persisted to the file system, or made visible in workspaces.</AlertBox>
```
gtsiolis, line 108: **thought**: It took me a while to wrap my head around the security cases here. 😇 

Some improvements could include:
1. Making the alert title brief and short in one line.
2. Avoiding complex state handling for users by default.

**suggestion**: For the second point I still think that having a sinsible default where variables can be exposed everywhere instead of requiring users to think everytime could be better. Here's how the copy could look like although I could be wrong on terminal access only:

| BEFORE | AFTER |
|-|-|
| Screenshot 2022-01-14 at 8 01 26 PM (2) | Screenshot 2022-01-14 at 8 01 10 PM (2) |

Let me know what you think!

gtsiolis (commented):
This took slightly longer than expected. 💭 

@jankeromnes I've left some final comments below regarding UX changes. Feel free to create follow up issues for each one of them if needed. 🏀

components/dashboard/src/projects/ProjectVariables.tsx
```diff
@@ -0,0 +1,132 @@
...
+            </div>
+            <div className="mt-4">
+                <h4>Value</h4>
+                <input className="w-full" type="text" name="value" value={value} onChange={e => setValue(e.target.value)} />
+            </div>
+            <div className="mt-4">
+                <CheckBox title="Hide in Workspaces" desc="Project variables are visible during prebuilds. Choose whether this variable should be visible in workspaces as well." checked={censored} onChange={() => setCensored(!censored)} />
+            </div>
+            {!censored && <div className="mt-4">
+                <InfoBox>This value will be directly visible to anyone who can open your repository in Gitpod.</InfoBox>
```
gtsiolis, line 124: **thought**: The layout change that happens on click could be quite abrupt and surpising for users. I'd suggest to not include this alert all together and default to having variables visible everywhere by default as described in https://github.com/gitpod-io/gitpod/pull/7295#discussion_r785052457. 💭

components/dashboard/src/projects/ProjectVariables.tsx
```diff
@@ -0,0 +1,132 @@
...
+                <h3 className="text-center pb-3 text-gray-500 dark:text-gray-400">No Environment Variables</h3>
+                <p className="text-center pb-6 text-gray-500 text-base w-96">Here you can set <strong>project-specific environment variables</strong> that will be visible during prebuilds and (optionally) in workspaces for this project.</p>
+                <button onClick={() => setShowAddVariableModal(true)}>New Variable</button>
+            </div>
+            : <>
+                <ItemsList>
+                    <Item header={true} className="grid grid-cols-4 items-center">
+                        <ItemField>Name</ItemField>
+                        <ItemField>Value</ItemField>
+                        <ItemField>Visible in Workspaces?</ItemField>
```
gtsiolis, line 62: **suggestion**: The question mark on the end looks friendly but slightly redandunt. What do you think?

```suggestion
                        <ItemField>Hidden in Workspaces</ItemField>
```

components/dashboard/src/projects/ProjectVariables.tsx
```diff
@@ -0,0 +1,132 @@
...
+                        <ItemField></ItemField>
+                    </Item>
+                    {envVars.map(variable => {
+                        return <Item className="grid grid-cols-4 items-center">
+                            <ItemField>{variable.name}</ItemField>
+                            <ItemField>****</ItemField>
+                            <ItemField>{variable.censored ? 'Hidden' : 'Visible'}</ItemField>
```
gtsiolis, line 69: **suggestion**: No strong feelings about this one but what do you think of using a plain `Yes`/`No` here? Alternatively, or in later iterations, we could use a checkmark icon to indicate `Yes`.

```suggestion
                            <ItemField>{variable.censored ? 'Yes' : 'No'}</ItemField>
```

gtsiolis (commented):
Three last comments. Promise! 🙂

components/dashboard/src/projects/ProjectVariables.tsx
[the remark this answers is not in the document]
jldec: +1 for _Project-specific Environment Variables_

components/dashboard/src/projects/ProjectVariables.tsx
```diff
@@ -0,0 +1,132 @@
...
+    return <Modal visible={true} onClose={props.onClose} onEnter={() => { addVariable(); return false; }}>
+        <h3 className="mb-4">Add Variable</h3>
+        <div className="border-t border-b border-gray-200 dark:border-gray-800 -mx-6 px-6 py-4 flex flex-col">
+            <AlertBox><strong>Project variables might be accessible by anyone with read access to your repository.</strong><br/>Secret values can be exposed if they are printed in the logs, persisted to the file system, or made visible in workspaces.</AlertBox>
+            {error && <div className="bg-gitpod-kumquat-light rounded-md p-3 text-gitpod-red text-sm mb-2">
+                {error}
+            </div>}
+            <div className="mt-8">
+                <h4>Name</h4>
+                <input autoFocus className="w-full" type="text" name="name" value={name} onChange={e => setName(e.target.value)} />
```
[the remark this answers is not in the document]
jldec: Yes please - I was able to add a variable with no name or value :)

components/dashboard/src/projects/ProjectVariables.tsx
jldec: - The message at the top could be a little clearer about how prebuild variables could leak into workspaces.
- The current default to **Hide in workspaces** is good IMO because it errs on the side of less exposure.
- I think we can leave out the extra instruction for what it means to uncheck the checkbox.

Suggest the following:

> **Warning**  
> Even with **Hide in Workspaces** checked, a project-specific environment variable may become exposed to anyone who can start a workspace e.g. if logged or persisted to the file system. 

> Name
> [ _ _ _ ]

> Value
> [ _ _ _ ]

> ☑️ **Hide in Workspaces**
> _Project-specific environment variables are always visible in prebuilds._

(after unchecking)

> **NOTE**
> This environment variable will be visible to anyone who starts a Gitpod workspace on this repository.

components/dashboard/src/projects/ProjectVariables.tsx
jldec: Let's leave it in - i.o.w. not make it easier for users to leak secrets. 

There is definitely a scrolling problem on landscape low-height displays. (the bottom of form becomes unreachable)
Screenshot 2022-01-14 at 20 08 56

jldec:
Awesome work @jankeromnes - thanks for going the extra mile with hiding in workspaces 🙏

roboquat:
New changes are detected. LGTM label has been removed.

components/dashboard/src/projects/ProjectVariables.tsx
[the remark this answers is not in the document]
jankeromnes: Good catch! We should indeed validate variable names here (for values though, pretty much any text is valid, including awkward symbols or leaving an empty string -- safely escaping/unescaping values is our responsibility).

I'll add that validation to the DB code, because client-side validation doesn't make sense here (because it still allows API users, or any module in our code, to use illegal values).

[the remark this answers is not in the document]
jankeromnes: Update: Seems like the specifications accept "any character except `=` or `NUL`" as valid environment variable names (including non-printable characters), but for compatibility reasons, it seems safer indeed to allow only:
- alphabeticals (upper/lowercase)
- digits (except as first character)
- underscores

Source: https://stackoverflow.com/a/2821183/3461173

(TL;DR The user env var name validation is still correct 😇 although the `+` is unnecessary)

components/dashboard/src/projects/ProjectVariables.tsx
```diff
@@ -0,0 +1,132 @@
...
+                    </Item>
+                    {envVars.map(variable => {
+                        return <Item className="grid grid-cols-4 items-center">
+                            <ItemField>{variable.name}</ItemField>
+                            <ItemField>****</ItemField>
+                            <ItemField>{variable.censored ? 'Hidden' : 'Visible'}</ItemField>
+                            <ItemField className="flex justify-end">
+                                <ItemFieldContextMenu menuEntries={[
+                                    {
+                                        title: 'Delete',
```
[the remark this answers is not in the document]
jankeromnes: Thanks for noticing! I personally think we don't need to be overly cautious here (i.e. clicking once on a red `Delete` button is likely already sufficient intent, and this action doesn't have surprising chain reactions as consequence).

Leaving this as an optional follow-up if we really feel we need the extra user confirmation.

components/dashboard/src/projects/ProjectVariables.tsx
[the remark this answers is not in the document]
jankeromnes: Sure, we can go with "Project-specific Environment Variables" as the full name. It's nice & explicit. 💯

But it's also a lot of text, so I guess we could still use just e.g. "variable" in contexts where the full name has already been introduced (e.g. error messages). Will adjust the warning (and other occurrences of "project variables" I can find) to the full name, but I hope this won't make the warning too long that users don't read it.

components/dashboard/src/projects/ProjectVariables.tsx
jankeromnes: Sure, happy to go with `Yes`/`No` for this iteration.

It's just that I really like how obvious and direct words like `Hidden`/`Visible` are (i.e. no need to mentally compute affirmation or negation of the column -- it's always crystal clear what the visibility is -- while for `Yes` no you might need to re-read the column heading to be 100% certain you're not making a mistake)

gtsiolis: @jankeromnes Your call. We can always come back and update this if we feel this could be improved. 🏀

jankeromnes: (Had `Yes`/`No`, but it felt a bit awkward, so I went back to `Visibility in Workspaces` --> `Visible`/`Hidden` 🙈 happy to iterate in follow-ups though)

components/dashboard/src/projects/ProjectVariables.tsx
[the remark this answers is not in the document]
jankeromnes: Validation implemented ✅

components/dashboard/src/projects/ProjectVariables.tsx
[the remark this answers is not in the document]
jankeromnes: Many thanks for filing this follow-up!

More than happy to implement the adjustments -- really like the new design. Please simply post a spec to the issue + remove `needs design` label + schedule, as usual 😊 (you could even directly assign me to it 🚀 I'm looking forward to have this fixed)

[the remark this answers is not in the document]
gtsiolis: Je vous remercie, @jankeromnes!

jankeromnes: Thanks for the excellent feedback here! 🙏

_Phew._ After splitting a few more hairs on this, I think I've reached an optimum consensus solution on the wording:

<table><tr><td>
Screenshot 2022-01-17 at 19 09 11
</td><td>
Screenshot 2022-01-17 at 19 09 19
</td></tr></table>

Going to ship this 🚢 (but happy to further split hairs in follow-up PRs 😅)

jankeromnes:
Just waiting for a last build to triple-check everything, then shipping this. 🚢

jankeromnes:
Everything seems to still work as expected! (Tested the variable management UI, but couldn't test prebuilds/workspaces because they're [currently broken in core-dev](https://gitpod.slack.com/archives/C01KGM9EBD4/p1642444331038600) (internal)).

Thus, releasing the breaks! 🚀 ✨

/unhold

shaal:
Thank you for working on this!
When will we be able to test Project-level environment variables in gitpod.io ?

jankeromnes:
@shaal Sure, my pleasure! This should go live in about an hour with the regular twice-a-week deployment. 📅📅

PR #7490, opened 2022-01-06
Support more regional workspace cluster

### LINKED ISSUE #5596: Automatically choose workspace-cluster based on lowest latency.

csweichel:
## Description
Today workspace cluster choice is guided solely by the location of the application cluster the workspace is started from. This PR introduces an admission preference (introduced in #6164) which adds more regional support.

The current implementation in this branch is a rather rough first start. It does two things:
1. introduce the `region` admission preferences on workspace clusters. Each such admission preference carries a "round trip time endpoint" which is used to measure the RTT of a user to a particular region. We assume that if there are multiple clusters within the same "region" that their "RTT endpoints" yield equivalent results.
2. make the `start workspace` page measure the RTT and update the cluster preference for the user. This part I'm least happy with, because it's
  - super spammy and makes many more requests than it should
  - comes too late: only after the first workspace is running will this be done

It's particularly with point 2 that I'd need some support @gitpod-io/engineering-meta :)

## Related Issue(s)

Fixes #5596

## How to test
1. remove the default cluster registration by editing the `WSMAN_CFG` env var on server setting it to `W10K` (base64 for `[]`)
2. register a the ws-manager of the preview environment with `gpctl clusters register ...` make sure it's governed
3. register a ws-manager of another preview environment
4. use `gpctl clusters update admission-preference add ...` to point one cluster to a "slow" RTT endpoint, e.g. `ws-us25.gitpod.io`, and one to a "fast" RTT endpoint, e.g. `ws-eu25.gitpod.io` (depending on where in the world you're located)
5. start a workspace and observe the correct cluster choice

## Release Notes

```release-note
Support more regional workspace cluster
```


roboquat:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please ask for approval from **csweichel** and additionally assign **geropl** after the PR has been reviewed.
You can assign the PR to them by writing `/assign @geropl` in a comment when ready.

Associated issue: *#6164*

<details open>

- **[components/ws-manager-bridge/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/ws-manager-bridge/OWNERS)**
- **[components/ws-manager-bridge-api/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/ws-manager-bridge-api/OWNERS)**
- **[dev/gpctl/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/dev/gpctl/OWNERS)**

</details>

components/gitpod-protocol/src/protocol.ts
```diff
@@ -1254,4 +1254,15 @@ export interface Terms {
 
 export interface WorkspaceClusterPreference {
     region?: string;
-}
+    // lrm == lastRegionMeasurement. Not writing the full name to save
+    // a few 100mb in the database.
+    lrm?: string;
```
jankeromnes, line 1259: Wait, for real? 🤯 (How was this calculated? Also, doesn't MySQL do compression? And sould we start using very short / unintelligible attribute names and values in every common shape? 😬 https://acronyms-suck.com/)

jankeromnes (commented):
Exciting to see progress on this! ⚡ 👏 Many thanks for moving this forward. 🙏

Added a few drive-by comments in the dashboard part. Please disregard if unhelpful.

JanKoehnlein:
My thoughts on this
1) Why not store the `lastMeasurement` date on the client (e.g. `localStorage`) instead of in the DB? It'd save one server call and allow to move this up in the component hierarchy as @jankeromnes recommends
2) Await on the measurement only if there is none yet. IMHO it's better to wait a few more seconds (AFAIKS its max 5s which could also be fine tuned) for the first workspace than being confronted with a horrible UX due to long latencies. We could show that to the user as a phase to raise acceptance if it takes longer (who would object a "Finding the best server for you")

JanKoehnlein:
Code LGTM now, also squashed it. 

I admit that  I am stuck in the testing steps as I don't know how to register a different preview env's ws-manager: What is the URL? The one of the local preview env is a `dns://` URL which I assume won't work.

PR #7241, opened 2021-12-15
community: route users to Discord for support

kylos101:
## Description

Route users to Discord for questions and support

## Related Issue(s)

Fixes #7240 

[Related conversation](https://gitpod.slack.com/archives/C01KPEPLLRY/p1639519263122900)

## How to test

Github links mentioned in the issue (bug report, issue template, and readme) route to Discord. We won't be able to test until we merge, I believe.

Also, @gitpod-io/engineering-meta there are a couple social links [[1](https://github.com/gitpod-io/gitpod/pull/7241/files#diff-a9d07052701ab3b718c82ac0fe74b965a8260b06bb5906362081718b0ee593b4R123-R124)] [[2](https://github.com/gitpod-io/gitpod/pull/7241/files#diff-001cc9444a98994948112e6868bc448d73251cc4595f816f43c0a36b36f616efR93-R94)], which I updated from Discourse to Discord. I couldn't find where they're being used, and am unsure how to test. I am unsure how to test the social type link for the configmap used by server...can you help :question: :pray: 

## Release Notes

```release-note
Route users to Discord for support
```

PR #6528, opened 2021-11-03
Replace hard-coded gitpod.io links

iQQBot:
## Description

Replace hard-coded gitpod.io links

## Related Issue(s)

Fixes #6400

## Release Notes

```release-note
NONE
```

PR #7502, opened 2022-01-07
[server] removed unused branding from config

JanKoehnlein:
## Description
Removes the branding data from the server config

## Related Issue(s)
Brought up by #7241

## Release Notes

```release-note
NONE
```


roboquat:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please ask for approval from **jankoehnlein** after the PR has been reviewed.

Associated issue: *#7241*

<details open>

- **[installer/pkg/components/server/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/installer/pkg/components/server/OWNERS)**

</details>

components/server/src/user/user-controller.ts
```diff
@@ -621,7 +620,7 @@ export class UserController {
             return;
         }
 
-        if (this.urlStartsWith(returnToURL, this.config.hostUrl.toString()) || this.urlStartsWith(returnToURL, this.config.brandingConfig.homepage)) {
```
geropl, line 624: Hm, I'm pretty sure this will break some flows. E.g., login from the website I think.

geropl: We could move this field up into `config`, and make it optional, and only set it for gitpod.io, for instance.

geropl:
@JanKoehnlein Did you check how we use this for gitpod.io? :thinking:

components/server/src/user/user-controller.ts
csweichel: it would also break self-hosted. Very few folks will want to redirect to www.gitpod.io

JanKoehnlein: @csweichel Would moving this field up in the config be an incompatible SH change?

JanKoehnlein:
> @JanKoehnlein Did you check how we use this for gitpod.io? 🤔

I couldn't find any reference on the branding there. We set the `homepage` differently (latest on Netlify) for staging and preview envs but that seems to be it.

components/server/src/user/user-controller.ts
geropl: > Would moving this field up in the config be an incompatible SH change?

No, this should work! :+1:

corneliusludmann: Currently, the installer does not allow to change the `homepage` config value. I don't know in which situation the `redirect` or `returnTo` query variable could point to a page of the homepage. If it's really a use case (@geropl ?), you need to add a `homepage` config value to the [installer config](https://github.com/gitpod-io/gitpod/blob/d9955ccc742430465b62b373baf4a01d3ba11eaf/installer/pkg/config/v1/config.go#L63) as well.

corneliusludmann: The only usage of the `homepage` value seems to check if the redirect to such a page is allowed. For self-hosted, the usage seems to be pretty limited (and not very obvious). If we really need this for SaaS, here are some additional options:

a) always allow redirects to `https://www.gitpod.io/` (hard-coded), just for our SaaS—not very smart but would probably be sufficient
b) allow always `www.$DOMAIN` (just because _we_ need it)—not very smart but would probably be sufficient
c) use a more precise name for the config value, e.g. like `allowAdditionalRedirectTarget` or similar—only if we really need it for self-hosted, but honestly, I don't think it's really a common case worth polluting the installer config for it


But at first, it would make sense to make sure if we really need this for SaaS, in my opinion.

geropl: > I don't know in which situation the redirect or returnTo query variable could point to a page of the homepage.

Hm, I thought we'd send users to www.gitpod.io on logout, but apprently we don't do that anymore.
IIRC, the "Login" button on www.gitpod.io would rely on the retunrnTo a few months ago (to send users back to some explainer), but that does not seem to be the case anymore. (maybe for 1st-time users?)

We configure(d) branding for production [here](https://github.com/gitpod-io/ops/blob/684defd9d6c0619dd3ab952b5cced3c50f0ad4ec/charts/gitpod_io/values.production.yaml#L131-L132), but:
 - the only value used is `homepage`, making this the only relevant line change in this PR
 - `redirectUrlAfterLogout` (change above in this file) is not used, and `redirectUrlIfNotAuthenticated` not used as well (subject to config drift it seems)
 
 Tl;dr:
  - Logout: we don't seem to send users to external websites anymore (stays on "Login" screen)
  - we don't seem to need it for website-flows (maybe @jldec can rule this out for first-time users?)
  - this change should be fine, then :shrug:

jldec:
We currently send unauthed users to the website at www.gitpod.io when they land on gitpod.io (if they don't have any cookie). Ideally we would only do this for gitpod.io - for all other domains including self-hosted it would be better redirect to /login instead.

components/server/src/user/user-controller.ts
JanKoehnlein: In order not to break SaaS, I've hard coded `https://www.gitpod.io` as an "allowed-return-to-URL-prefix" (option a in @corneliusludmann's proposal). I think this is not worthwhile to be part of (SH) configuration, unless we change the redirect logic more deeply.

components/server/src/user/user-controller.ts
```diff
@@ -621,7 +620,7 @@ export class UserController {
...
+        if (this.urlStartsWith(returnToURL, this.config.hostUrl.toString()) || this.urlStartsWith(returnToURL, "https://www.gitpod.io")) {
```
corneliusludmann, line 623: I don't know exactly how the attack vector looks like that we want to address with this check, however, this `urlStartsWith(...)` check would also hold for a potentially malicious domain like `https://www.gitpod.io.malicious.com`, right? Thus, you need at least add the trailing slash to the domain or do a proper host check, don't you?


```suggestion
        if (this.urlStartsWith(returnToURL, this.config.hostUrl.toString()) || this.urlStartsWith(returnToURL, "https://www.gitpod.io/")) {
```

gtsiolis: **fyi**: Cross-linking a relevant change containing dynamic host URL from https://github.com/gitpod-io/gitpod/pull/6528 (Cc @iQQBot) in case there's anything we could replicate here or there.

JanKoehnlein: `this.urlStartsWith()` already adds the trailing slash if missing

geropl:
/lgtm

I saw that https://github.com/gitpod-io/gitpod/pull/7502#discussion_r781365471 got resolve. :+1:

roboquat:

This pull-request has been approved by: *<a href="https://github.com/gitpod-io/gitpod/pull/7502#issuecomment-1010010168" title="LGTM">geropl</a>*

Associated issue: *#7241*

<details >

- ~~[OWNERS](https://github.com/gitpod-io/gitpod/blob/main/OWNERS)~~ [geropl]
- ~~[components/gitpod-protocol/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-protocol/OWNERS)~~ [geropl]
- ~~[components/server/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/server/OWNERS)~~ [geropl]
- ~~[installer/pkg/components/server/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/installer/pkg/components/server/OWNERS)~~ [geropl]

</details>

PR #7591, opened 2022-01-13
Retrieve telemetry send status from server component

MrSimonEmms:
## Description

Creates the installation admin controller in the `server` component and associated database. This database works on the basis that there is only one record created per installation - in future, this will be the backing for the installation's admin panel. At the moment, this only has one record (`sendTelemetry`), but will have more records in future.

The `sendTelemetry` boolean currently defaults to `false` and will do until the admin panel exists then it will change to default `true`.

It amends the Installer to cater for these changes. It allows the `gitpod-telemetry` job to access the `server` component. It also removed the `GITPOD_DOMAIN_HASH` and uses the randomly generated ID for installation identification.

## Related Issue(s)

Fixes #7571
Fixes #7585 

## How to test

Deploy via werft and check that you can access `GET:/installation-admin/data` when port forwarding to the server component.

## Release Notes

```release-note
[server]: Create installation admin controller
```

PR #7655, opened 2022-01-18
[Installer]: allow for the telemetry hash to be further hashed

### LINKED ISSUE #7538: Allow for optional pepper to further hash the domain hash
The `GITPOD_DOMAIN_HASH` envvar receives a hash of the Gitpod domain. Whilst cryptographically secure, it's not too difficult to guess domains where Gitpod might be running and see if a hash is in the database.

As a further protection to anonymity, allow for an optional secret to be defined which add a pepper value and then further hashes the `GITPOD_DOMAIN_HASH` value in the component.

The Installer's envvar would look something like this:

```go
[]v1.Container{
    Env: []v1.EnvVar{
        {
		Name: "GITPOD_PEPPER",
		ValueFrom: &v1.EnvVarSource{
			SecretKeyRef: &v1.SecretKeySelector{
				LocalObjectReference: v1.LocalObjectReference{
					Name: "secret-name",
				},
				Optional: pointer.Bool(true),
			},
		},
	},
}
```

In the `installation-telemetry` component, the hash would then happen like this:

```go
domainHash := os.Getenv("GITPOD_DOMAIN_HASH")

pepper := os.Getenv("GITPOD_PEPPER")
if pepper != "" {
  domainHash = fmt.Sprintf("%x", sha512.Sum512([]byte(domainHash + pepper)))
}
```

MrSimonEmms:
## Description

The `GITPOD_DOMAIN_HASH` envvar receives a hash of the Gitpod domain. Whilst cryptographically secure, it's not too difficult to guess domains where Gitpod might be running and see if a hash is in the database.

This allows for an additional and optional secret to be provided to further hash the unique token. 

## Related Issue(s)

Fixes #7538

## How to test

 - Create a secret `kubectl create secret generic telemetry --from-literal=hash=<your-custom-hash>`
 - Add to config:
```yaml
telemetry:
  kind: secret
  name: telemetry
```
 - Validate the config/cluster
 - Deploy and check that `GITPOD_CUSTOM_HASH` set to the `installation-telemetry` cronjob
## Release Notes

```release-note
[Installer]: allow for the telemetry hash to be further hashed
```

### DIFF
--- a/components/installation-telemetry/cmd/send.go
+++ b/components/installation-telemetry/cmd/send.go
@@ -5,6 +5,7 @@
 package cmd
 
 import (
+	"crypto/sha512"
 	"fmt"
 	"os"
 
@@ -38,6 +39,12 @@ var sendCmd = &cobra.Command{
 		versionId := os.Getenv("GITPOD_INSTALLATION_VERSION")
 		if versionId == "" {
 			return fmt.Errorf("GITPOD_INSTALLATION_VERSION envvar not set")
+		}
+
+		customHash := os.Getenv("GITPOD_CUSTOM_HASH")
+		if customHash != "" {
+			log.Info("Applying additional hash to the domain hash")
+			domainHash = fmt.Sprintf("%x", sha512.Sum512([]byte(domainHash+customHash)))
 		}
 
 		client, err := analytics.NewWithConfig(segmentIOToken, analytics.Config{})
--- a/installer/example-config.yaml
+++ b/installer/example-config.yaml
@@ -8,6 +8,7 @@ certificate:
   name: https-certificates
 containerRegistry:
   inCluster: true
+  s3storage: null
 database:
   inCluster: true
 domain: gitpod.example.com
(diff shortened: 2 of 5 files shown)


roboquat:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please ask for approval from **mrsimonemms** after the PR has been reviewed.

Associated issue: *#7538*

<details open>

- **[components/installation-telemetry/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/installation-telemetry/OWNERS)**

</details>

metcalfc:
Who is the envisioned user for this? If we want to achieve decoupling the install from the domain why not just further hash against the `uid` field of the kube-system namespace as a cluster identifier. In this way the config surface is small. It's truly a one way hash (no recreating it) and if the domain moves around to a different cluster that is in effect a new install. So we'd see the old one disappear. Somewhere else the new one tied to the new cluster would appear.

corneliusludmann:
Since we have the connection to the database soon, [I already proposed to use a generated ID that is stored to the database as installation ID](https://github.com/gitpod-io/gitpod/pull/7591#discussion_r786139886). I still think that would be a good idea for the following reasons:

- It would solve our problem. We make it truly anonymous. No connection to the domain.
- I find that the life cycle of the database fits pretty well with that of an installation instance. When I move, update etc. my installation, I will make a database backup. Thus, the ID survives.
- The admin does not need to know anything about the domain hash. The current implementation with the pepper was rather a stopgap solution that makes it unnecessarily complicated. Since we have a way to store an artificially generated ID now, we should make use of it.

What do you think?

MrSimonEmms:
Yeah, I'd not thought of that, but you're right that we're generating a UUID in the database so will switch to that

Issue #7564, opened 2022-01-12
Text input + autosuggest for New Workspace modal UI


jldec:
This is the first step to improve the UX for startig new workspaces in the dashboard. Specifically this removes the 2 tabs in the existing modal UI, and replaces them with a single input field with auto-suggest. The modal behavior (grey background) can remain the same. For additional context, see #7554 

![Screenshot 2022-02-01 at 21 20 48](https://user-images.githubusercontent.com/849592/152053446-d120331e-d31f-4a86-a25f-6632dd66e83b.png)

![Screenshot 2022-02-01 at 21 18 50](https://user-images.githubusercontent.com/849592/152053167-edba21f4-65ca-460a-9d3d-09b5a573685d.png)

![Screenshot 2022-02-03 at 20 26 09](https://user-images.githubusercontent.com/849592/152423280-c2b1ce68-63df-49e4-af53-201c1b936bd0.png)

### Proposed UI changes  
The new design should have only the following elements

- A single text input with the ability to paste a repo context URL or type some text
- An autosuggest feature with selectable results for a repo context which are based on what is in the text input
- Short instructions for how to use:
  `Paste a repository context URL or start typing to see suggestions.` [Learn more](https://www.gitpod.io/docs/context-urls) 

### Repo context requirements
The input should accept the same repo context URLs as currently supported for prefixed URL workspace starts.  

### Autosuggest requirements
When the input text matches, autosuggested results should include at least the following:
- recent workspaces
- examples (like today) 
- personal account repositories from your git provider(s) 

### Nice to have
- A dashboard keybinding (exact characters TBD) similar to a command pallette reachable from anywhere in the Dashboard.

jldec:
Visual design guidelines for this first implementation from the sync today:

- Keep the current modal dimensions and scrolling behavior and list size limit unchanged
- Render the auto-suggested results as simple git repo context URLs including the `http://git-provider/...` prefix 
- No need for a submit button - click on any result to submit.

Suggestions for future improvements

- Render results with URL broken down into components: git-provider, user-org/repo-name, issue / PR / commit context
- Provide additional hints in the results for different types of repos e.g examples, recently used, personal, org
- Explore alternatives to size-constrained modal UI, including non-modal page with non-nested scrolling

jldec:
See https://github.com/gitpod-io/gitpod/issues/3715#issuecomment-811947169 for additional UX ideas.

jldec:
[...]

Issue #4948, opened 2021-07-26
Update project configurator designs


gtsiolis:
Following the updates in project https://github.com/gitpod-io/gitpod/pull/4631 which added a cut through of the project configurator, the project configurator designs need to be updated.

### Designs

Adding below the first step of the project configurator.

| Authorize Provider |
|-|
| <img width="1440" alt="Pending" src="https://user-images.githubusercontent.com/120486/126963409-a8b862c3-cee3-4ebb-99b3-2e8e5738341a.png"> | 

| Repository List | Repository Hover | Repository Taken |
|-|-|-|
| <img width="1440" alt="Default" src="https://user-images.githubusercontent.com/120486/127672398-8f3407b9-bccb-4b0a-8744-4686caab845c.png"> | <img width="1440" alt="Hover" src="https://user-images.githubusercontent.com/120486/127672482-b536e3e2-cf82-4128-acf9-12bd9f7f8064.png"> | <img width="1440" alt="Taken" src="https://user-images.githubusercontent.com/120486/127672486-fae3178f-e301-433d-9275-0e198e2a35ff.png"> |

| Select Team | Create Team |
|-|-|
| <img width="1440" alt="Default" src="https://user-images.githubusercontent.com/120486/127672643-cfc64766-0208-427a-8a9d-6c05b400dc74.png"> | <img width="1440" alt="Create" src="https://user-images.githubusercontent.com/120486/127672656-bb0e4036-7ae5-44c9-93af-18492715bb6e.png"> |

| Configuration Existing in Git |
|-|
| <img width="1440" alt="Configure" src="https://user-images.githubusercontent.com/120486/127672818-c2776e7e-70b4-4702-8773-b034bb442ec6.png"> |

| Add Configuration | Pending Prebuild | Failed Prebuild | Running Prebuild | Ready prebuild |
|-|-|-|-|-|
[...]

PR #7715, opened 2022-01-20
[server][dashboard] Improve 'New Workspace' modal

jankeromnes:
## Description

Improve the 'New Workspace' modal with a search input, keyboard navigation, and a new context URL suggestion API.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/7564

## How to test

1. Click on `New Workspace` (or hit `Ctrl/Cmd + O` from anywhere in the dashboard)
2. Type to find a suggested repository* or paste any context URL and hit Enter

*currently sources suggestions from:
- Example repos
- Your recent repos
- Your repos from all your [connected integrations](https://jx-open-in-gitpod.staging.gitpod-dev.com/integrations)

## Release Notes

```release-note
[server][dashboard] Improve 'New Workspace' modal with a search input, keyboard navigation, and a new context URL suggestion API
```

## Documentation

/uncc

### DIFF
--- a/components/dashboard/src/components/Modal.tsx
+++ b/components/dashboard/src/components/Modal.tsx
@@ -51,7 +51,7 @@ export default function Modal(props: {
     return (
         <div className="fixed top-0 left-0 bg-black bg-opacity-70 z-50 w-screen h-screen">
             <div className="w-screen h-screen align-middle" style={{display: 'table-cell'}}>
-                <div className={"relative bg-white dark:bg-gray-900 border border-gray-200 dark:border-gray-800 rounded-xl p-6 max-w-lg mx-auto text-left text-gray-600 " + (props.className || '')} onClick={e => e.stopPropagation()}>
+                <div className={"relative bg-white dark:bg-gray-900 border border-gray-200 dark:border-gray-800 rounded-xl p-6 max-w-lg mx-auto text-left " + (props.className || '')} onClick={e => e.stopPropagation()}>
                     {props.closeable !== false && (
                         <div className="absolute right-7 top-6 cursor-pointer text-gray-800 dark:text-gray-100 hover:bg-gray-200 dark:hover:bg-gray-700 rounded-md p-2" onClick={props.onClose}>
                             <svg version="1.1" width="14px" height="14px"
--- a/components/dashboard/src/components/RepositoryFinder.tsx
+++ b/components/dashboard/src/components/RepositoryFinder.tsx
@@ -6,6 +6,7 @@
 
 import { User } from "@gitpod/gitpod-protocol";
 import React, { useContext, useEffect, useState } from "react";
+import { Link } from "react-router-dom";
 import { getGitpodService } from "../service/service";
 import { UserContext } from "../user-context";
 
@@ -80,9 +81,18 @@ export default function RepositoryFinder(props: { initialQuery?: string }) {
             <div className="py-4">
                 <svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 16 16" width="16" height="16"><path fill="#A8A29E" d="M6 2a4 4 0 100 8 4 4 0 000-8zM0 6a6 6 0 1110.89 3.477l4.817 4.816a1 1 0 01-1.414 1.414l-4.816-4.816A6 6 0 010 6z" /></svg>
             </div>
-            <input type="search" className="flex-grow" placeholder="Repository" autoFocus value={searchQuery} onChange={e => search(e.target.value)} onKeyDown={onKeyDown} />
+            <input type="search" className="flex-grow" placeholder="Search repositories and examples" autoFocus value={searchQuery} onChange={e => search(e.target.value)} onKeyDown={onKeyDown} />
         </div>
-        <div className="mt-3 flex flex-col space-y-2 h-64 overflow-y-auto" id="search-results">
+        <div className="mt-3 -mx-5 px-5 flex flex-col space-y-2 h-64 overflow-y-auto">
+            {searchQuery === '' && searchResults.length === 0 &&
+                <div className="mt-12 mx-auto w-96 text-gray-500">
+                    Paste a <a className="gp-link" href="https://www.gitpod.io/docs/context-urls">repository context URL</a>, or start typing to see suggestions from:
+                    <ul className="list-disc mt-4 pl-7 flex flex-col space-y-1">
+                        <li>Your recent repositories</li>
+                        <li>Your repositories from <Link className="gp-link" to="/integrations">connected integrations</Link></li>
+                        <li>Example repositories</li>
+                    </ul>
+                </div>}
             {searchResults.slice(0, MAX_DISPLAYED_ITEMS).map((result, index) =>
                 <a className={`px-4 py-3 rounded-xl` + (result === selectedSearchResult ? ' bg-gray-600 text-gray-50 dark:bg-gray-700' : '')} href={`/#${result}`} key={`search-result-${index}`} onMouseEnter={() => setSelectedSearchResult(result)}>
                     {searchQuery.length < 2
@@ -94,7 +104,7 @@ export default function RepositoryFinder(props: { initialQuery?: string }) {
                 </a>
             )}
             {searchResults.length > MAX_DISPLAYED_ITEMS &&
-                <span className="mt-3 px-4 py-2 italic text-sm">{searchResults.length - MAX_DISPLAYED_ITEMS} results not shown</span>}
+                <span className="mt-3 px-4 py-2 text-sm text-gray-400 dark:text-gray-500">{searchResults.length - MAX_DISPLAYED_ITEMS} more result{(searchResults.length - MAX_DISPLAYED_ITEMS) === 1 ? '' : 's'} found</span>}
         </div>
     </form>;
 }
@@ -147,6 +157,10 @@ async function actuallyRefreshSearchData(query: string, user: User | undefined):
 }
 
 async function findResults(query: string, onResults: (results: string[]) => void) {
+    if (!query) {
+        onResults([]);
+        return;
+    }
     const searchData = loadSearchData();
     try {
         // If the query is a URL, and it's not present in the proposed results, "artificially" add it here.
@@ -157,5 +171,5 @@ async function findResults(query: string, onResults: (results: string[]) => void
     } catch {
     }
     // console.log('searching', query, 'in', searchData);
-    onResults(searchData.filter(result => result.includes(query)));
+    onResults(searchData.filter(result => result.toLowerCase().includes(query.toLowerCase())));
 }
--- a/components/dashboard/src/components/RepositoryFinder.tsx
+++ b/components/dashboard/src/components/RepositoryFinder.tsx
@@ -1,0 +1,161 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { User } from "@gitpod/gitpod-protocol";
+import React, { useContext, useEffect, useState } from "react";
+import { getGitpodService } from "../service/service";
+import { UserContext } from "../user-context";
+
+type SearchResult = string;
+type SearchData = SearchResult[];
+
+const LOCAL_STORAGE_KEY = 'open-in-gitpod-search-data';
+const MAX_DISPLAYED_ITEMS = 15;
+
+export default function RepositoryFinder(props: { initialQuery?: string }) {
+    const { user } = useContext(UserContext);
+    const [searchQuery, setSearchQuery] = useState<string>(props.initialQuery || '');
+    const [searchResults, setSearchResults] = useState<SearchResult[]>([]);
+    const [selectedSearchResult, setSelectedSearchResult] = useState<SearchResult | undefined>();
+
+    const onResults = (results: SearchResult[]) => {
+        if (JSON.stringify(results) !== JSON.stringify(searchResults)) {
+            setSearchResults(results);
+            setSelectedSearchResult(results[0]);
+        }
+    }
+
+    const search = async (query: string) => {
+        setSearchQuery(query);
+        await findResults(query, onResults);
+        if (await refreshSearchData(query, user)) {
+            // Re-run search if the underlying search data has changed
+            await findResults(query, onResults);
+        }
+    }
+
+    // Up/Down keyboard navigation between results
+    const onKeyDown = (event: React.KeyboardEvent) => {
+        if (!selectedSearchResult) {
+            return;
+        }
+        const selectedIndex = searchResults.indexOf(selectedSearchResult);
+        const select = (index: number) => {
+            // Implement a true modulus in order to "wrap around" (e.g. `select(-1)` should select the last result)
+            // Source: https://stackoverflow.com/a/4467559/3461173
+            const n = Math.min(searchResults.length, MAX_DISPLAYED_ITEMS);
+            setSelectedSearchResult(searchResults[((index % n) + n) % n]);
+        }
+        if (event.key === 'ArrowDown') {
+            event.preventDefault();
+            select(selectedIndex + 1);
+            return;
+        }
+        if (event.key === 'ArrowUp') {
+            event.preventDefault();
+            select(selectedIndex - 1);
+            return;
+        }
+    }
+
+    useEffect(() => {
+        const element = document.querySelector(`a[href='/#${selectedSearchResult}']`);
+        if (element) {
+            element.scrollIntoView({ behavior: 'smooth', block: 'nearest' });
+        }
+    }, [selectedSearchResult]);
+
+    const onSubmit = (event: React.FormEvent) => {
+        event.preventDefault();
+        if (selectedSearchResult) {
+            window.location.href = '/#' + selectedSearchResult;
+        }
+    }
+
+    return <form onSubmit={onSubmit}>
+        <div className="flex px-4 rounded-xl border border-gray-300 dark:border-gray-500">
+            <div className="py-4">
+                <svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 16 16" width="16" height="16"><path fill="#A8A29E" d="M6 2a4 4 0 100 8 4 4 0 000-8zM0 6a6 6 0 1110.89 3.477l4.817 4.816a1 1 0 01-1.414 1.414l-4.816-4.816A6 6 0 010 6z" /></svg>
+            </div>
+            <input type="search" className="flex-grow" placeholder="Repository" autoFocus value={searchQuery} onChange={e => search(e.target.value)} onKeyDown={onKeyDown} />
+        </div>
+        <div className="mt-3 flex flex-col space-y-2 h-64 overflow-y-auto" id="search-results">
+            {searchResults.slice(0, MAX_DISPLAYED_ITEMS).map((result, index) =>
+                <a className={`px-4 py-3 rounded-xl` + (result === selectedSearchResult ? ' bg-gray-600 text-gray-50 dark:bg-gray-700' : '')} href={`/#${result}`} key={`search-result-${index}`} onMouseEnter={() => setSelectedSearchResult(result)}>
+                    {searchQuery.length < 2
+                        ? <span>{result}</span>
+                        : result.split(searchQuery).map((segment, index) => <span>
+                            {index === 0 ? <></> : <strong>{searchQuery}</strong>}
+                            {segment}
+                        </span>)}
+                </a>
+            )}
+            {searchResults.length > MAX_DISPLAYED_ITEMS &&
+                <span className="mt-3 px-4 py-2 italic text-sm">{searchResults.length - MAX_DISPLAYED_ITEMS} results not shown</span>}
+        </div>
+    </form>;
+}
+
+function loadSearchData(): SearchData {
+    const string = localStorage.getItem(LOCAL_STORAGE_KEY);
+    if (!string) {
+        return [];
+    }
+    try {
+        const data = JSON.parse(string);
+        return data;
+    } catch (error) {
+        console.warn('Could not load search data from local storage', error);
+        return [];
+    }
+}
+
+function saveSearchData(searchData: SearchData): void {
+    try {
+        window.localStorage.setItem(LOCAL_STORAGE_KEY, JSON.stringify(searchData));
+    } catch (error) {
+        console.warn('Could not save search data into local storage', error);
+    }
+}
+
+let refreshSearchDataPromise: Promise<boolean> | undefined;
+export async function refreshSearchData(query: string, user: User | undefined): Promise<boolean> {
+    if (refreshSearchDataPromise) {
+        // Another refresh is already in progress, no need to run another one in parallel.
+        return refreshSearchDataPromise;
+    }
+    refreshSearchDataPromise = actuallyRefreshSearchData(query, user);
+    const didChange = await refreshSearchDataPromise;
+    refreshSearchDataPromise = undefined;
+    return didChange;
+}
+
+// Fetch all possible search results and cache them into local storage
+async function actuallyRefreshSearchData(query: string, user: User | undefined): Promise<boolean> {
+    console.log('refreshing search data');
+    const oldData = loadSearchData();
+    const newData = await getGitpodService().server.getSuggestedContextURLs();
+    if (JSON.stringify(oldData) !== JSON.stringify(newData)) {
+        console.log('new data:', newData);
+        saveSearchData(newData);
+        return true;
+    }
+    return false;
+}
+
+async function findResults(query: string, onResults: (results: string[]) => void) {
+    const searchData = loadSearchData();
+    try {
+        // If the query is a URL, and it's not present in the proposed results, "artificially" add it here.
+        new URL(query);
+        if (!searchData.includes(query)) {
+            searchData.push(query);
+        }
+    } catch {
+    }
+    // console.log('searching', query, 'in', searchData);
+    onResults(searchData.filter(result => result.includes(query)));
+}
--- a/components/dashboard/src/workspaces/StartWorkspaceModal.tsx
+++ b/components/dashboard/src/workspaces/StartWorkspaceModal.tsx
@@ -4,18 +4,24 @@
  * See License-AGPL.txt in the project root for license information.
  */
 
-import { useContext } from "react";
+import { useContext, useEffect } from "react";
+import { useLocation } from "react-router";
 import Modal from "../components/Modal";
 import RepositoryFinder from "../components/RepositoryFinder";
 import { StartWorkspaceModalContext } from "./start-workspace-modal-context";
 
 export function StartWorkspaceModal() {
     const { isStartWorkspaceModalVisible, setIsStartWorkspaceModalVisible } = useContext(StartWorkspaceModalContext);
+    const location = useLocation();
+
+    // Close the modal on navigation events.
+    useEffect(() => {
+        setIsStartWorkspaceModalVisible(false);
+    }, [location]);
 
     return <Modal onClose={() => setIsStartWorkspaceModalVisible(false)} onEnter={() => false} visible={!!isStartWorkspaceModalVisible}>
-        <h3 className="pb-2">New Workspace</h3>
+        <h3 className="pb-2">Open in Gitpod</h3>
         <div className="border-t border-gray-200 dark:border-gray-800 mt-2 -mx-6 px-6 pt-4">
-            <h4 className="text-base">Search or paste a repository URL</h4>
             <RepositoryFinder />
         </div>
     </Modal>;
--- a/components/dashboard/src/workspaces/StartWorkspaceModal.tsx
+++ b/components/dashboard/src/workspaces/StartWorkspaceModal.tsx
@@ -4,18 +4,16 @@
  * See License-AGPL.txt in the project root for license information.
  */
 
+import { useContext } from "react";
 import Modal from "../components/Modal";
 import RepositoryFinder from "../components/RepositoryFinder";
+import { StartWorkspaceModalContext } from "./start-workspace-modal-context";
 
-interface StartWorkspaceModalProps {
-    visible: boolean;
-    onClose: () => void;
-}
+export function StartWorkspaceModal() {
+    const { isStartWorkspaceModalVisible, setIsStartWorkspaceModalVisible } = useContext(StartWorkspaceModalContext);
 
-export function StartWorkspaceModal(props: StartWorkspaceModalProps) {
-    return <Modal onClose={props.onClose} onEnter={() => false} visible={props.visible}>
+    return <Modal onClose={() => setIsStartWorkspaceModalVisible(false)} onEnter={() => false} visible={!!isStartWorkspaceModalVisible}>
         <h3 className="pb-2">New Workspace</h3>
-        {/* separator */}
         <div className="border-t border-gray-200 dark:border-gray-800 mt-2 -mx-6 px-6 pt-4">
             <h4 className="text-base">Search or paste a repository URL</h4>
             <RepositoryFinder />
--- a/components/dashboard/src/workspaces/StartWorkspaceModal.tsx
+++ b/components/dashboard/src/workspaces/StartWorkspaceModal.tsx
@@ -4,70 +4,21 @@
  * See License-AGPL.txt in the project root for license information.
  */
 
-import { useEffect, useState } from "react";
 import Modal from "../components/Modal";
-import TabMenuItem from "../components/TabMenuItem";
-
-export interface WsStartEntry {
-    title: string
-    description: string
-    startUrl: string
-}
+import RepositoryFinder from "../components/RepositoryFinder";
 
 interface StartWorkspaceModalProps {
     visible: boolean;
-    recent: WsStartEntry[];
-    examples: WsStartEntry[];
-    selected?: Mode;
     onClose: () => void;
 }
 
-type Mode = 'Recent' | 'Examples';
-
-export function StartWorkspaceModal(p: StartWorkspaceModalProps) {
-    const computeSelection = () => p.selected || (p.recent.length > 0 ? 'Recent' : 'Examples');
-    const [selection, setSelection] = useState(computeSelection());
-    useEffect(() => { !p.visible && setSelection(computeSelection()) }, [p.visible, p.recent, p.selected]);
-
-    const list = (selection === 'Recent' ? p.recent : p.examples).map((e, i) =>
-        <a key={`item-${i}-${e.title}`} href={e.startUrl} className="rounded-xl group hover:bg-gray-100 dark:hover:bg-gray-800 flex p-4 my-1">
-            <div className="w-full">
-                <p className="text-base text-gray-800 dark:text-gray-200 font-semibold">{e.title}</p>
-                <p>{e.description}</p>
-            </div>
-        </a>);
-
-    return <Modal onClose={p.onClose} visible={p.visible}>
+export function StartWorkspaceModal(props: StartWorkspaceModalProps) {
+    return <Modal onClose={props.onClose} onEnter={() => false} visible={props.visible}>
         <h3 className="pb-2">New Workspace</h3>
         {/* separator */}
-        <div className="border-t border-gray-200 dark:border-gray-800 mt-2 -mx-6 px-6 pt-2">
-            <div className="flex">
-                <TabMenuItem name='Recent' selected={selection === 'Recent'} onClick={() => setSelection('Recent')} />
-                {p.examples.length>0 && <TabMenuItem name='Examples' selected={selection === 'Examples'} onClick={() => setSelection('Examples')} />}
-            </div>
-        </div>
-        <div className="border-t border-gray-200 dark:border-gray-800 -mx-6 px-6 py-2">
-            {list.length > 0 ?
-                <p className="my-4 text-base">
-                    {selection === 'Recent' ?
-                        'Create a new workspace using the default branch.' :
-                        'Create a new workspace using an example project.'}
-                </p> : <p className="h-6 my-4"></p>}
-            <div className="space-y-2 mt-4 overflow-y-scroll h-80 pr-2">
-                {list.length > 0 ? list :
-                    (selection === 'Recent' ?
-                        <div className="flex flex-col pt-10 items-center px-2">
-                            <h3 className="mb-2 text-gray-500 dark:text-gray-400">No Recent Projects</h3>
-                            <p className="text-center">Projects you use frequently will show up here.</p>
-                            <p className="text-center">Prefix a Git repository URL with {window.location.host}/# or start with an example.</p>
-                            <button onClick={() => setSelection('Examples')} className="font-medium mt-8">Select Example</button>
-                        </div> :
-                        <div className="flex flex-col pt-10 items-center px-2">
-                            <h3 className="mb-2 text-gray-500 dark:text-gray-400">No Example Projects</h3>
-                            <p className="text-center">Sorry there seem to be no example projects, that work with your current Git provider.</p>
-                        </div>)
-                }
-            </div>
+        <div className="border-t border-gray-200 dark:border-gray-800 mt-2 -mx-6 px-6 pt-4">
+            <h4 className="text-base">Search or paste a repository URL</h4>
+            <RepositoryFinder />
         </div>
     </Modal>;
 }
--- a/components/dashboard/src/workspaces/Workspaces.tsx
+++ b/components/dashboard/src/workspaces/Workspaces.tsx
@@ -69,7 +69,7 @@ export default function () {
                                 onClick: () => { if (workspaceModel) workspaceModel.limit = 200; }
                             }]} />
                         </div>
-                        <button onClick={() => setIsStartWorkspaceModalVisible(true)} className="ml-2">New Workspace ({StartWorkspaceModalKeyBinding})</button>
+                        <button onClick={() => setIsStartWorkspaceModalVisible(true)} className="ml-2">New Workspace <span className="opacity-60 hidden md:inline">{StartWorkspaceModalKeyBinding}</span></button>
                     </div>
                     <ItemsList className="app-container pb-40">
                         <div className="border-t border-gray-200 dark:border-gray-800"></div>
@@ -99,7 +99,7 @@ export default function () {
                                 <h3 className="text-center pb-3 text-gray-500 dark:text-gray-400">No Workspaces</h3>
                                 <div className="text-center pb-6 text-gray-500">Prefix any Git repository URL with {window.location.host}/# or create a new workspace for a recently used project. <a className="gp-link" href="https://www.gitpod.io/docs/getting-started/">Learn more</a></div>
                                 <span>
-                                    <button onClick={() => setIsStartWorkspaceModalVisible(true)}>New Workspace ({StartWorkspaceModalKeyBinding})</button>
+                                    <button onClick={() => setIsStartWorkspaceModalVisible(true)}>New Workspace <span className="opacity-60 hidden md:inline">{StartWorkspaceModalKeyBinding}</span></button>
                                 </span>
                             </>
                         </div>
--- a/components/dashboard/src/workspaces/Workspaces.tsx
+++ b/components/dashboard/src/workspaces/Workspaces.tsx
@@ -44,6 +44,19 @@ export default function () {
     const showStartWSModal = () => setIsTemplateModelOpen(true);
     const hideStartWSModal = () => setIsTemplateModelOpen(false);
 
+    useEffect(() => {
+        const onKeyDown = (event: KeyboardEvent) => {
+            if ((event.metaKey || event.ctrlKey) && ['k', 'o', 'p'].includes(event.key)) {
+                event.preventDefault();
+                showStartWSModal();
+            }
+        };
+        window.addEventListener('keydown', onKeyDown);
+        return () => {
+            window.removeEventListener('keydown', onKeyDown);
+        }
+    }, []);
+
     return <>
         <Header title="Workspaces" subtitle="Manage recent and stopped workspaces." />
 
--- a/components/dashboard/src/workspaces/Workspaces.tsx
+++ b/components/dashboard/src/workspaces/Workspaces.tsx
@@ -5,16 +5,16 @@
  */
 
 import { useContext, useEffect, useState } from "react";
-import { WhitelistedRepository, Workspace, WorkspaceInfo } from "@gitpod/gitpod-protocol";
+import { WhitelistedRepository, WorkspaceInfo } from "@gitpod/gitpod-protocol";
 import Header from "../components/Header";
 import DropDown from "../components/DropDown";
 import { WorkspaceModel } from "./workspace-model";
 import { WorkspaceEntry } from "./WorkspaceEntry";
-import { getGitpodService, gitpodHostUrl } from "../service/service";
-import { StartWorkspaceModal, WsStartEntry } from "./StartWorkspaceModal";
+import { getGitpodService } from "../service/service";
+import { StartWorkspaceModal } from "./StartWorkspaceModal";
 import { ItemsList } from "../components/ItemsList";
-import { getCurrentTeam, TeamsContext } from "../teams/teams-context";
-import { useLocation, useRouteMatch } from "react-router";
+import { TeamsContext } from "../teams/teams-context";
+import { useLocation } from "react-router";
 
 export interface WorkspacesProps {
 }
@@ -29,20 +29,12 @@ export default function () {
     const location = useLocation();
 
     const { teams } = useContext(TeamsContext);
-    const team = getCurrentTeam(location, teams);
-    const match = useRouteMatch<{ team: string, resource: string }>("/(t/)?:team/:resource");
-    const projectSlug = match?.params?.resource !== 'workspaces' ? match?.params?.resource : undefined;
     const [activeWorkspaces, setActiveWorkspaces] = useState<WorkspaceInfo[]>([]);
     const [inactiveWorkspaces, setInactiveWorkspaces] = useState<WorkspaceInfo[]>([]);
-    const [repos, setRepos] = useState<WhitelistedRepository[]>([]);
     const [isTemplateModelOpen, setIsTemplateModelOpen] = useState<boolean>(false);
     const [workspaceModel, setWorkspaceModel] = useState<WorkspaceModel>();
 
     useEffect(() => {
-        // only show example repos on the global user context
-        if (!team && !projectSlug) {
-            getGitpodService().server.getFeaturedRepositories().then(setRepos);
-        }
         (async () => {
             const workspaceModel = new WorkspaceModel(setActiveWorkspaces, setInactiveWorkspaces);
             setWorkspaceModel(workspaceModel);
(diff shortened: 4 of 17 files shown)


roboquat:

Associated issue: *#7564*

<details >

</details>

roboquat:

Associated issue: *#7564*

<details >

</details>

components/dashboard/src/components/Modal.tsx
```diff
@@ -51,7 +51,7 @@ export default function Modal(props: {
     return (
         <div className="fixed top-0 left-0 bg-black bg-opacity-70 z-50 w-screen h-screen">
             <div className="w-screen h-screen align-middle" style={{display: 'table-cell'}}>
-                <div className={"relative bg-white dark:bg-gray-900 border border-gray-200 dark:border-gray-800 rounded-xl p-6 max-w-lg mx-auto text-left text-gray-600 " + (props.className || '')} onClick={e => e.stopPropagation()}>
+                <div className={"relative bg-white dark:bg-gray-900 border border-gray-200 dark:border-gray-800 rounded-xl p-6 max-w-lg mx-auto text-left " + (props.className || '')} onClick={e => e.stopPropagation()}>
```
jankeromnes, line 54: This was making all text in all modals `text-gray-600`, which was impractical.

I've removed it and checked several Modals, but I hope I haven't missed any side effects of this fix.

components/dashboard/src/components/RepositoryFinder.tsx
```diff
@@ -0,0 +1,159 @@
...
+        <div className="flex px-4 rounded-xl border border-gray-300 dark:border-gray-500">
+            <div className="py-4">
+                <svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 16 16" width="16" height="16"><path fill="#A8A29E" d="M6 2a4 4 0 100 8 4 4 0 000-8zM0 6a6 6 0 1110.89 3.477l4.817 4.816a1 1 0 01-1.414 1.414l-4.816-4.816A6 6 0 010 6z" /></svg>
+            </div>
+            <input type="search" className="flex-grow" placeholder="Repository" autoFocus value={searchQuery} onChange={e => search(e.target.value)} onKeyDown={onKeyDown} />
+        </div>
+        <div className="mt-3 flex flex-col space-y-2 h-64 overflow-y-auto" id="search-results">
+            {searchResults.slice(0, MAX_DISPLAYED_ITEMS).map((result, index) =>
+                <a className={`px-4 py-3 rounded-xl` + (result === selectedSearchResult ? ' bg-gray-600 text-gray-50 dark:bg-gray-700' : '')} href={`/#${result}`} key={`search-result-${index}`} onMouseEnter={() => setSelectedSearchResult(result)}>
+                    {result.split(searchQuery).map((segment, index) => <span>
```
jankeromnes, line 88: Possible UI performance optimization: Don't split if `searchQuery.length < 2`.

jankeromnes:
Alright, I think the current state is pretty cool. ✨

| Initial state | Type anything | Paste |
| --- | --- | --- |
| Screenshot 2022-01-21 at 16 18 40 | Screenshot 2022-01-21 at 16 18 52 | Screenshot 2022-01-21 at 16 19 05 |
| Screenshot 2022-01-21 at 16 19 23 | Screenshot 2022-01-21 at 16 19 34 | Screenshot 2022-01-21 at 16 20 06 |

@gtsiolis @jldec Could you please [take a look](https://jx-open-in-gitpod.staging.gitpod-dev.com/workspaces)? 🙏

Cool things to try:
- `Up`/`Down` arrow, `Enter`
- `Cmd`+`K` shortcut (or `P`, `O`, or with `Ctrl`)
- type `rust`, `next`, `petclinic`
- paste any context URL
- connect GitLab and Bitbucket in [/integrations](https://jx-open-in-gitpod.staging.gitpod-dev.com/integrations) then try to find your repos in the Modal

gtsiolis:
Looking at this now! 👀

components/dashboard/src/workspaces/Workspaces.tsx
```diff
@@ -52,38 +44,18 @@ export default function () {
...
-                            }
-                        }
-                    }
-                }
-                const list = Array.from(index.values());
-                list.sort((a, b) => b.lastUse.localeCompare(a.lastUse));
-                return list;
+    useEffect(() => {
+        const onKeyDown = (event: KeyboardEvent) => {
+            if (['k', 'o', 'p'].includes(event.key) && (event.metaKey || event.ctrlKey)) {
```
gtsiolis, line 49: **suggestion**: I'd suggest to go with only one shortcut here to avoid hurting memory muscle in the long run or confuse users. I'd vote for keeping only `CMD+K` for this one.

components/dashboard/src/workspaces/Workspaces.tsx
```diff
@@ -52,38 +44,18 @@ export default function () {
...
-                                entry.lastUse = lastUse;
-                            }
-                        }
-                    }
-                }
-                const list = Array.from(index.values());
-                list.sort((a, b) => b.lastUse.localeCompare(a.lastUse));
-                return list;
+    useEffect(() => {
+        const onKeyDown = (event: KeyboardEvent) => {
```
gtsiolis, line 48: **question**: Noticed we close the modal with `ESC` but what do you think of also closing the modal with the same key binding?

components/dashboard/src/workspaces/StartWorkspaceModal.tsx
```diff
@@ -4,70 +4,21 @@
...
-                        </div> :
-                        <div className="flex flex-col pt-10 items-center px-2">
-                            <h3 className="mb-2 text-gray-500 dark:text-gray-400">No Example Projects</h3>
-                            <p className="text-center">Sorry there seem to be no example projects, that work with your current Git provider.</p>
-                        </div>)
-                }
-            </div>
+        <div className="border-t border-gray-200 dark:border-gray-800 mt-2 -mx-6 px-6 pt-4">
+            <h4 className="text-base">Search or paste a repository URL</h4>
+            <RepositoryFinder />
```
gtsiolis, line 21: **question**: Could it help to keep the input state here when closing and re-opening the modal? Sometimes you could accidentally close the modal just before clicking the result.

components/dashboard/src/workspaces/Workspaces.tsx
gtsiolis, line 48: **thought**: Somehow I was expecting to be able to trigger this modal from every page in the dashboard but opening this only while in the workspaces list sounds also of for the first iteration. ➿

components/dashboard/src/components/RepositoryFinder.tsx
```diff
@@ -0,0 +1,159 @@
...
+                        {index === 0 ? <></> : <strong>{searchQuery}</strong>}
+                        {segment}
+                    </span>)}
+                </a>
+            )}
+            {searchResults.length > MAX_DISPLAYED_ITEMS &&
+                <span className="mt-3 px-4 py-2 italic text-sm">{searchResults.length - MAX_DISPLAYED_ITEMS} results not shown</span>}
```
gtsiolis, line 95: **issue**: Why the limit here? I thought we decided to go with the scrollbar because then we could always show all results. Otherwise, there's not much different with the fixed result set we discussed earlier this week, right? Cc @jldec 

**suggestion**: In any case, going with a fixed max number of results and infiniter scroll pagination would be ideal, but sounds also ok to leave out of the scope of this PR. 💭 

Screenshot 2022-01-21 at 6 11 21 PM

components/dashboard/src/components/RepositoryFinder.tsx
```diff
@@ -0,0 +1,159 @@
...
+    return <form onSubmit={onSubmit}>
+        <div className="flex px-4 rounded-xl border border-gray-300 dark:border-gray-500">
+            <div className="py-4">
+                <svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 16 16" width="16" height="16"><path fill="#A8A29E" d="M6 2a4 4 0 100 8 4 4 0 000-8zM0 6a6 6 0 1110.89 3.477l4.817 4.816a1 1 0 01-1.414 1.414l-4.816-4.816A6 6 0 010 6z" /></svg>
+            </div>
+            <input type="search" className="flex-grow" placeholder="Repository" autoFocus value={searchQuery} onChange={e => search(e.target.value)} onKeyDown={onKeyDown} />
+        </div>
+        <div className="mt-3 flex flex-col space-y-2 h-64 overflow-y-auto" id="search-results">
+            {searchResults.slice(0, MAX_DISPLAYED_ITEMS).map((result, index) =>
+                <a className={`px-4 py-3 rounded-xl` + (result === selectedSearchResult ? ' bg-gray-600 text-gray-50 dark:bg-gray-700' : '')} href={`/#${result}`} key={`search-result-${index}`} onMouseEnter={() => setSelectedSearchResult(result)}>
```
gtsiolis, line 87: **suggestion**: It would be probably good to add some right margin here to avoid squashing the highlighted reasult entry with the visible scrollbar. What do you think?

```suggestion
                <a className={`px-4 py-3 rounded-xl mr-2` + (result === selectedSearchResult ? ' bg-gray-600 text-gray-50 dark:bg-gray-700' : '')} href={`/#${result}`} key={`search-result-${index}`} onMouseEnter={() => setSelectedSearchResult(result)}>
```

| BEFORE | AFTER |
|-|-|
| Screenshot 2022-01-21 at 6 17 21 PM (2) | Screenshot 2022-01-21 at 6 17 35 PM (2) |

gtsiolis, line 87: **praise**: Loving there's an always-on and selected entry in here wherever you move the mouse pointer. ✨

gtsiolis, line 87: **issue**: There's a small vertical scroll nudge in some cases when hovering over the first or the last entry of the scrollbar area. Probably an after effect of having the entries wrapped? 🤷 

https://user-images.githubusercontent.com/120486/150562913-d9861788-57f5-473a-806e-9b0964abccd9.mov

 the scrollbar area. What do you think?

components/dashboard/src/workspaces/StartWorkspaceModal.tsx
gtsiolis, line 21: **thought**: Hopefully it's now more clear to everyone that listing recent repositories, workspaces, etc when triggering the modal for the first time makes sense! However, sounds also ok to leave this out of the scope of this iteration.

gtsiolis (approved):
Looks great, @jankeromnes! 🔮 

Added some first round review comments below with some non-blocking questions. 🥊 

Approving to unblock merging but holding in case we could tackle any of the comments below and since someone else could take a closer look at the code changes.

/hold

roboquat:
LGTM label has been added.  <details>Git tree hash: c63d3e13f3b4e80b6d013910e2fbd1b5d4ff4265</details>

roboquat:

This pull-request has been approved by: *<a href="https://github.com/gitpod-io/gitpod/pull/7715#pullrequestreview-859729132" title="Approved">gtsiolis</a>*

Associated issue: *#7564*

<details >

- ~~[components/dashboard/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/dashboard/OWNERS)~~ [gtsiolis]
- ~~[components/gitpod-protocol/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-protocol/OWNERS)~~ [gtsiolis]
- ~~[components/server/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/server/OWNERS)~~ [gtsiolis]

</details>

components/dashboard/src/workspaces/Workspaces.tsx
jldec: +1

components/dashboard/src/components/RepositoryFinder.tsx
jankeromnes: For performance reasons. I noticed the UI started struggling a bit at around 25 results. Also, I found that anything after 10-15 results is no longer relevant for my query.

Happy to adjust as needed, but there should probably always be a hard limit for performance reasons (also, Jurgen's concern was about seeing 3 elements and wanting to see more, I guess when you see 10-15 results you no longer have this problem).

components/dashboard/src/components/RepositoryFinder.tsx
jankeromnes: That's because of the `scrollIntoView()` behavior when an element becomes selected (via keyboard or mouse selection). I didn't find this to be a problem when testing (and even found it helpful, e.g. if I hover over a partially hidden element, I can now read it fully).

Did you find this behavior surprising / unwelcome?

components/dashboard/src/workspaces/StartWorkspaceModal.tsx
jankeromnes: Sorry, I'm not sure I understand this. Would you like us to make it clearer what is being displayed / what a user should type? (I agree that presenting a user with a searchbar and nothing else can be a bit confusing at first -- I wonder how VS Code solves this with their Command Palette input 🤔)

components/dashboard/src/workspaces/Workspaces.tsx
jldec: Is it a lot of work making it work from other Dashboard pages? I think it makes a big qualitative difference to the feature.

components/dashboard/src/workspaces/Workspaces.tsx
jankeromnes: Thanks! I'd personally vote for using `Cmd+P` and also `Ctrl+P` (for non Mac users).

Also, are there really downsides to having an "official" keystroke (that we advertise everywhere) + also "secretly" catching the other two keystrokes, and doing something useful with them instead of nothing?

We could always rebind these to other features in the future, but I'm a big fan of going to where our users already are, instead of trying to force/"educate" everyone into a single behavior.

components/dashboard/src/workspaces/Workspaces.tsx
jankeromnes: Interesting. Is this a standard/expected behavior for repeat keystrokes?

I notice that VS Code does not close the command palette if you press `F1` a second time (or `Cmd+P` a second time).

components/dashboard/src/workspaces/StartWorkspaceModal.tsx
jankeromnes: I would vote against this, as I prefer the modal to always have a ready-to-go fresh start as opposed to "old stuff you need to clean up first".

Can it really happen that users close the modal by mistake and then try to open it again? (Like, accidentally press `Esc`, or accidentally click the `X`? 🤔 I find this scenario a bit unlikely.)

components/dashboard/src/workspaces/Workspaces.tsx
jankeromnes: Thanks for the reminder! It's not a lot of work. I'll do it in the next iteration 👍

components/dashboard/src/components/RepositoryFinder.tsx
jankeromnes: Uh oh 😬 I'm not a big fan of either, and I think I'd prefer having the results be the same width as the input. I'll see if I can improve this somehow.

components/dashboard/src/workspaces/StartWorkspaceModal.tsx
gtsiolis: I don't have a concrete proposal yet but I meant to say that populating the empty modal white space, when users haven't typed anything yet, with recent repositories or workspaces could be useful.

Also:
> However, sounds also ok to leave this out of the scope of this iteration.

components/dashboard/src/components/RepositoryFinder.tsx
gtsiolis: I've seen both approaches work good, both scrolling or not scrolling when hovering or selecting a list entry inside a scrollable list. But I agree this small scroll nudge can be helpful. Let's leave it as is for now. 🤝

gtsiolis: Yeah, [no biggie](https://en.wiktionary.org/wiki/no_biggie). Regarding width, same width would be nice but that could require us to move the scrollbar outside the scrolling area. Posting below a slightly related design with scrollbars from https://github.com/gitpod-io/gitpod/issues/4948 to better visualize the problem. In any case, feel free to leave this as is, too. 🏓 

127672486-fae3178f-e301-433d-9275-0e198e2a35ff

components/dashboard/src/workspaces/StartWorkspaceModal.tsx
gtsiolis: I've seen both approaches work equally good depending on the application.

> Can it really happen that users close the modal by mistake and then try to open it again?

I'd think yes, but this will become much more relevant once the command palette (this modal) becomes more useful or contains more actions to take. However, let's leave this as is for now and come back to this if we run into this in action. 🤝

components/dashboard/src/workspaces/Workspaces.tsx
gtsiolis: Again, like in https://github.com/gitpod-io/gitpod/pull/7715#discussion_r790696114:

> I've seen both approaches work equally good depending on the application.

> However, let's leave this as is for now and come back to this if we run into this in action. 🤝

components/dashboard/src/workspaces/StartWorkspaceModal.tsx
jldec: Concretely, the challenge is that the new help text in the UI does not help users to understand what is being  auto-suggested or searched.

I don't think we should introduce any magic search syntax, but it would be nice to convey that the suggestions are coming from the 3 possible sources (also mentioned in the description above). This can happen either in the results themselves (indicate which result is from which source), or, my preference, in some help text below the input. It's fine to replace that when the results show up

The original suggested help text was

> Paste a repository context URL or start typing to see suggestions. [Learn more](url)

I propose adding the following text (below the input) and remove the help text above the input. 

>  Paste a repository [context URL](url) 
>  Or start typing to see suggestions from
>  ~ Your recent repositories
>  ~ Your repositories from connected integrations
>  ~ Example repositories

---

For reference here's how github does it 

**👇🏼** in in the file finder (T in code)
Screenshot 2022-01-24 at 09 58 24

**👇🏼** in the new command palette (cmd+k)

Screenshot 2022-01-24 at 09 56 23

**👇🏼** in the new beta project command palette (cmd+k in project).

Screenshot 2022-01-24 at 09 54 42

components/dashboard/src/workspaces/Workspaces.tsx
gtsiolis: For some more context, the concepts of the **command palette** and **quick switcher** overlap with what we're trying to achieve here with this modal. Some applications merge these two while other keep them distinctly separate.💡 

> I'd personally vote for using Cmd+P and also Ctrl+P (for non Mac users).

For historical context, `CMD/Ctrl+P` was predominantly the first key binding mapped for triggering search or command palettes while `CMD/Ctrl+K` recently started being used for command palettes[[1](https://philipcdavis.com/writing/command-palette-interfaces)][[2](https://blog.superhuman.com/how-to-build-a-remarkable-command-palette/)][[3](https://dribbble.com/tags/command_palette)][...]. FWIW, `CMD/Ctrl+O` has been historically used to open a file in many applications or operating systems[[4](https://en.wikipedia.org/wiki/Keyboard_shortcut#%22Sacred%22_keybindings)][[5](https://support.microsoft.com/en-us/topic/keyboard-shortcuts-for-the-file-menu-in-office-for-windows-1c4b07af-be2e-4c19-8d33-c600a8904043#bkmk_open)][...]. ⚔️ 

> .. are there really downsides to having an ...

This could mostly hurt usability in a system as **consistency and standards**[[1](https://www.nngroup.com/articles/consistency-and-standards/)] can be affected from having multiple bindings that lead to the same action. [Miller's law](https://en.wikipedia.org/wiki/Miller%27s_law) can be also helpful in this decision. Last but now least, studies[[1](https://www.researchgate.net/publication/12189991_When_Choice_is_Demotivating_Can_One_Desire_Too_Much_of_a_Good_Thing)][[2](https://www.nngroup.com/articles/minimize-cognitive-load/)] show that having too many options often leads to decision paralysis and frustration. 

🍋 🍋 🍋 🍋 

Like in https://github.com/gitpod-io/gitpod/pull/7715#discussion_r790696114, I'd suggest to leave this as is for now and come back to this if this becomes an issue unless @jldec has any strong feelings towards one specific direction. Cc @jankeromnes

svenefftinge:
I find it a little sad, that it doesn't know about the repositories I care about. So digged a bit into GitHub API.
This graphql query would return the first 100 repos I have contributed to. It's actually pretty fast query.
```
query { 
  user(login: "svenefftinge") {
      repositoriesContributedTo(includeUserRepositories: true, first: 100) {
        edges {
          node {
            nameWithOwner
            shortDescriptionHTML(limit: 120)
            url
          }
        }
      }
    }
}
```
Try [here](https://docs.github.com/en/graphql/overview/explorer).

jankeromnes:
Exciting find, many thanks @svenefftinge!

Actually this code should already fetch all your GitHub repositories (either just `svenefftinge/*` or also from orgs, I'm not sure). It uses the same code that feeds the "New Project" wizard options (just with a faster UX), which unfortunately doesn't work in this preview because I haven't connected a GitHub App yet.

You can test it by connecting GitLab and Bitbucket in `/integrations` though.

Will add a GitHub App in the next deployment.

However, I'll likely keep your query, because it can probably find more relevant results than what we already have. Thanks again!

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -961,6 +961,59 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
...
+        // Example repositories
+        promises.push(this.getFeaturedRepositories(ctx).then(exampleRepos => {
+            // log('got example repos', exampleRepos);
+            exampleRepos.forEach(r => suggestions.push(r.url));
+        }));
+
+        // User repositories
+        user.identities.forEach(identity => {
+            const provider = {
+                'Public-GitLab': 'gitlab.com',
```
svenefftinge, line 980: should this not just go through all my providers (i.e. incl. any gitlab self-managed installations)

svenefftinge (commented):
Cool! I belive it would be better if suggestions are displayed already when you first open the dialog with an empty input. WDYT?

jankeromnes:
> Cool! I belive it would be better if suggestions are displayed already when you first open the dialog with an empty input. WDYT?

I also like this idea, and you can sort of simulate it by typing then removing a letter. (The order of the suggestions is currently alphabetical, which we could probably make more relevant.)

We discussed this earlier today in our team meeting. Another option could be to have a helpful empty state, which explains what exactly is happening here / what you should do / where the results come from. But if we can immediately show a list of the most relevant suggestions, maybe we don't even need to "explain" the feature with an empty state (i.e. you just click on New Workspace, then you immediately see the repo you're interested in, and voilà, job done, no "learning" was required because the feature was intuitive).

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -961,6 +961,59 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
...
+                'Public-GitHub': 'github.com',
+                'Public-Bitbucket': 'bitbucket.org',
+            }[identity.authProviderId];
+            if (!provider) {
+                return;
+            }
+            promises.push(this.getProviderRepositoriesForUser(ctx, { provider }).then(userRepos => {
```
svenefftinge, line 987: I don't think reusing this method is a good idea as it limits the list of repos to the repos that have the app installed. Also fetches projects etc. IMHO this should be entirely about proposing great matches for a user, no matter an app is installed. We will grow out of what this method does pretty soon anyway.

roboquat:
New changes are detected. LGTM label has been removed.

jldec:
Following a sync with @gtsiolis 🙏 , here are some additional guidelines and feedback on the design comments above.
@jankeromnes Please let me know if this provides sufficient clarity to get the PR into a shippable state.

- Help text  
  Please put the help text below the input as proposed in my [comment](#discussion_r790715508) above.
  @gtsiolis can provide visual design for grey background, centering.

- Key binding - Command + O 
  OK with the action button text if we remove the parentheses and tone down the color (@gtsiolis has suggestion)

- Prepopulated results
  I would prefer to postpone this until we also have a way to include recent workspaces.

- Extra label above the input
  Can we move this into the input as placeholder to free up the space above?

- Performance and usability of the suggestions/results
  Need to test with larger numbers of repos from GitHub - could you install the GitHub app please?

components/dashboard/src/workspaces/StartWorkspaceModal.tsx
```diff
@@ -4,70 +4,19 @@
...
-        <a key={`item-${i}-${e.title}`} href={e.startUrl} className="rounded-xl group hover:bg-gray-100 dark:hover:bg-gray-800 flex p-4 my-1">
-            <div className="w-full">
-                <p className="text-base text-gray-800 dark:text-gray-200 font-semibold">{e.title}</p>
-                <p>{e.description}</p>
-            </div>
-        </a>);
-
-    return <Modal onClose={p.onClose} visible={p.visible}>
+    return <Modal onClose={() => setIsStartWorkspaceModalVisible(false)} onEnter={() => false} visible={!!isStartWorkspaceModalVisible}>
         <h3 className="pb-2">New Workspace</h3>
```
gtsiolis, line 16: **suggestion**: What do you think of keeping the Open in Gitpod term here? New Workspace seems also ok. Your call. Added below some exports from the specs. 🏓 

| BEFORE | AFTER |
|-|-|
| Frame 381 | Frame 382 |

gtsiolis (commented):
@jankeromnes I've added below some designs from the follow up comment from @jldec in https://github.com/gitpod-io/gitpod/pull/7715#issuecomment-1023156335. 🏀

components/dashboard/src/workspaces/start-workspace-modal-context.tsx
```diff
@@ -0,0 +1,37 @@
...
+            window.removeEventListener('keydown', onKeyDown);
+        }
+    }, []);
+
+    return <StartWorkspaceModalContext.Provider value={{ isStartWorkspaceModalVisible, setIsStartWorkspaceModalVisible }}>
+        {children}
+    </StartWorkspaceModalContext.Provider>;
+}
+
+export const StartWorkspaceModalKeyBinding = `${/(Mac|iPhone|iPod|iPad)/i.test(navigator.platform) ? '⌘' : 'Ctrl+'}O`;
```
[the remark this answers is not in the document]
jankeromnes: @gtsiolis `Ctrl+O` is the keybinding for non-Mac users (i.e. Windows / Linux). From your reference:

> Ctrl+X | Current Microsoft notation
> ⌃X | Classic Mac OS and macOS notation

I guess `^O` will speak mostly to Mac users. I remember seeing this notation before switching to a Mac and being very confused myself (especially since my keyboard didn't have the `^` symbol on the right key).

Also, it's worth noting that VS Code chose to use `⌘O` on Mac and `Ctrl+O` on Windows (not `^O`). I guess aligning with VS Code is good here. 💭

components/dashboard/src/workspaces/Workspaces.tsx
```diff
@@ -113,7 +69,7 @@ export default function () {
                                 onClick: () => { if (workspaceModel) workspaceModel.limit = 200; }
                             }]} />
                         </div>
-                        <button onClick={showStartWSModal} className="ml-2">New Workspace</button>
+                        <button onClick={() => setIsStartWorkspaceModalVisible(true)} className="ml-2">New Workspace ({StartWorkspaceModalKeyBinding})</button>
```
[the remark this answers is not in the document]
jankeromnes: Aha, good point. I guess a good 99% solution here would be to hide the keystroke on narrow screens. 😇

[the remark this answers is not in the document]
gtsiolis: > I guess a good 99% solution here would be to hide the keystroke on narrow screens. 

POSSIBLE. 🙂

components/dashboard/src/workspaces/start-workspace-modal-context.tsx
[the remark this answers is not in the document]
gtsiolis: Wait, I could but wrong isn't `^` a notation for the `Ctrl` key? 💭 

Re-posting references from https://github.com/gitpod-io/gitpod/pull/7715#discussion_r793649559:
[1]: https://en.wikipedia.org/wiki/Control_key#Notation
[2]: https://en.wikipedia.org/wiki/Caret_notation

jankeromnes:
Cool, many thanks for the input @jldec and @gtsiolis! 👍 

> @jankeromnes Please let me know if this provides sufficient clarity to get the PR into a shippable state.

Seems sufficient 👍

> - Prepopulated results
>   I would prefer to postpone this until we also have a way to include recent workspaces.

This Pull Request already displays repos from all recent workspaces 🤔 I guess you mean more specific context URLs like recent PRs, branches, issues, etc?

> - Extra label above the input
>   Can we move this into the input as placeholder to free up the space above?

Didn't the label already move below the input in the new spec? Assuming nothing else to do here.

> - Performance and usability of the suggestions/results
>   Need to test with larger numbers of repos from GitHub - could you install the GitHub app please? 

Alright alright, I'll set up the GitHub App here. 😇

components/dashboard/src/workspaces/Workspaces.tsx
jankeromnes: Thanks for the great references!

I ended up putting this to a poll, and for now Cmd/Ctrl+O won 😁 picking this as the current keybinding for now. We can always improve it later.

gtsiolis: FWIW, `Cmd/Ctrl+O` sounds great! Thanks for pushing this forward and going with one single keybinding for this iteration. 💯 

> We can always improve it later.

Yes! ➿

components/server/src/workspace/gitpod-server-impl.ts
jankeromnes: Good point, thanks. Added a `TODO` comment as a first start.

components/dashboard/src/workspaces/start-workspace-modal-context.tsx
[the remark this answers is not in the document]
jankeromnes: > Wait, I could be wrong but isn't `^` a notation for the `Ctrl` key? 💭 

Indeed, it seems to be, but my point is, only Mac OS users typically know it, while Windows and Linux users will very likely not recognize this notation.

jankeromnes:
Thanks @jldec for surfacing a problem with the GitHub App integration:

Sometimes, when you've enabled the app, you can see your personal repos in the list, but then they "disappear" again:

| GitHub App | No results | Console output |
| --- | --- | --- |
| Screenshot 2022-02-01 at 11 13 37 | Screenshot 2022-02-01 at 11 13 02 | Screenshot 2022-02-01 at 11 12 47 |

The console output indicates that the GitHub App results were there, but then got removed again during a refresh.

Looking at the logs, I only ever see the message `got github.com user repos` with all my personal repos (never with 0 repos), and no directly relevant error. This other error, though, seems to hint towards a GitHub API problem (e.g. maybe a rate-limit?)

```js
{
  "message": "error while checking student pack status",
  "error": `SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at EligibilityService.<anonymous> (/app/node_modules/@gitpod/server/dist/ee/src/user/eligibility-service.js:86:37)
    at fulfilled (/app/node_modules/@gitpod/server/dist/ee/src/user/eligibility-service.js:19:58)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)`
}
```

components/dashboard/src/components/RepositoryFinder.tsx
jankeromnes: @gtsiolis Okay, I believe I've found an elegant fix for this 😁 I've simply made the scroll zone wider (negative X-margin) and compensated with some padding (positive X-padding). This doesn't require changing the size of other elements, and has the effect of "simply moving the scrollbar a bit to the right":

Screenshot 2022-02-01 at 15 40 55
Screenshot 2022-02-01 at 15 43 17

gtsiolis: Thanks for giving this a go, @jankeromnes!

**suggestion**: I think going with the simple inner right padding for now sounds better.

**issue**: Moving the scrollbar outside could make the user feel like they can scroll the through the whole modal inner elements, including title, text input, etc, not just the repository list. This also becomes more visible or an issue when always showing scrollbars, see screenshots below. This is also what I see! 👀 

| Light | Dark |
|-|-|
| scrollbar-light | scrollbar-dark |

**thought**: Keeping the scrollbar near the scrollbar area could help us :a: use better layouts like the one with the gray background mentioned in https://github.com/gitpod-io/gitpod/pull/7715#discussion_r793756071, :b: potentially introduce more fixed elements like a bottom toolbar for surfacing results found, etc.

Regarding the scrollbar blindness issue we discussed a few days ago, here's a relevant thread[[1](https://news.ycombinator.com/item?id=24293421)] but unfortunately the original article has been removed. I could only find this partial screenshot[[2](https://www.are.na/block/8469033)]. 🌟

components/dashboard/src/components/RepositoryFinder.tsx
```diff
@@ -0,0 +1,171 @@
...
+            window.location.href = '/#' + selectedSearchResult;
+        }
+    }
+
+    return <form onSubmit={onSubmit}>
+        <div className="flex px-4 rounded-xl border border-gray-300 dark:border-gray-500">
+            <div className="py-4">
+                <svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 16 16" width="16" height="16"><path fill="#A8A29E" d="M6 2a4 4 0 100 8 4 4 0 000-8zM0 6a6 6 0 1110.89 3.477l4.817 4.816a1 1 0 01-1.414 1.414l-4.816-4.816A6 6 0 010 6z" /></svg>
+            </div>
+            <input type="search" className="flex-grow" placeholder="repository" autoFocus value={searchQuery} onChange={e => search(e.target.value)} onKeyDown={onKeyDown} />
```
gtsiolis, line 84: **suggestion**: What do you think of changing the palceholder here with something like the following?

```suggestion
            <input type="search" className="flex-grow" placeholder="Search repositories and examples" autoFocus value={searchQuery} onChange={e => search(e.target.value)} onKeyDown={onKeyDown} />
```

components/dashboard/src/components/RepositoryFinder.tsx
```diff
@@ -0,0 +1,171 @@
...
+        </div>
+        <div className="mt-3 -mx-5 px-5 flex flex-col space-y-2 h-64 overflow-y-auto">
+            {searchQuery === '' && searchResults.length === 0 &&
```
gtsiolis, line 87: **issue**: When you search for something and then delete the text input you are still seeing some arbitrary results below. Expected?

**suggestion**: What do you think of rolling back to the emopty (default) state when deleting the search input?

components/dashboard/src/components/RepositoryFinder.tsx
gtsiolis: Ah, I see! Ideally, we could paginate the results with an infinite scroll but sounds good to leave this out of the scope of this PR. However, there's definitely some value in surfacing the amount of results found.

**suggestion**: What do you think of rephrasing this to the following?

```suggestion
                <span className="mt-3 px-4 py-2 text-sm text-gray-400 dark:text-gray-500">{searchResults.length - MAX_DISPLAYED_ITEMS} results found</span>}
```

| BEFORE | AFTER | 
|-|-|
| results-before | results-after |

Ideally this could also use singular when only 1 more result has been found. However, feel free to leave this as is for now as more improvements could come into this area.

gtsiolis (commented):
@jankeromnes I've added below three more minor UX related comments. 🏀

components/dashboard/src/workspaces/Workspaces.tsx
```diff
@@ -113,7 +69,7 @@ export default function () {
                                 onClick: () => { if (workspaceModel) workspaceModel.limit = 200; }
                             }]} />
                         </div>
-                        <button onClick={showStartWSModal} className="ml-2">New Workspace</button>
+                        <button onClick={() => setIsStartWorkspaceModalVisible(true)} className="ml-2">New Workspace <span className="opacity-60 hidden md:inline">{StartWorkspaceModalKeyBinding}</span></button>
```
gtsiolis, line 72: **praise**: Thanks for hiding this in `md:inline` here! 🔝

components/dashboard/src/components/RepositoryFinder.tsx
```diff
@@ -0,0 +1,171 @@
...
+                <div className="mt-12 mx-auto w-96 text-gray-500">
+                    Paste a <a className="gp-link" href="https://www.gitpod.io/docs/context-urls">repository context URL</a>, or start typing to see suggestions from:
+                    <ul className="list-disc mt-4 pl-7 flex flex-col space-y-1">
+                        <li>Your recent repositories</li>
+                        <li>Your repositories from <Link className="gp-link" to="/integrations">connected integrations</Link></li>
```
gtsiolis, line 92: **issue**: The page redirect when clicking this feels a bit abrupt, no? 😬 

**suggestion**: What if, we skipped linking to the integrations on this empty state and only prompted users to connect more providers when they couldn't find the repository they were looking for? What do you think? Cc @jldec 

| Modal (Empty) | Modal (Missing Repository) |
|-|-|
|  |  |

gtsiolis: > Hmm, good catch, but I think that's how we agreed to implement links (i.e. no more target="_blank" to open links in a new tab by default).

Correct and agree! To clarify I was _not_ suggesting to open this link in a new tab but either :a: close the modal and redirect to the the integrations page or even better :b: use help links only for external links like docs and skip the link to integrations on the empty state.

filiptronicek:
@jankeromnes I'm really loving these changes, awesome job! Maybe this could be added later, as it's nothing critical, but I'd love to see something similar to what Notion is doing to issue/PR URLs from GitHub, specifically querying the provider for things like the title and state.


<table>

<tr>
	<td> Proposed
	<td> Current (as of this PR)

<tr>
	<td>

image (13)


</td>
	<td>

image


 </td>

</table>

P.S.: please pardon my horrible PR icon (color, contrast, placement, size)

jankeromnes:
> @jankeromnes I'm really loving these changes, awesome job!

Many thanks for giving it a spin, @filiptronicek! 🙌

> Maybe this could be added later, as it's nothing critical, but I'd love to see something similar to what Notion is doing [[1](https://user-images.githubusercontent.com/29888641/152358155-e4d8661e-6b41-4e2c-9f7f-b732375b366f.png)] to issue/PR URLs from GitHub, specifically querying the provider for things like the title and state.

Good idea! I sort of like these "rich results". They're especially valuable for Pull Requests (however, the current modal shows mostly repositories currently, unless you specifically paste a PR URL, but then you probably already know the content/state of that PR 💭).

See also:
- recent [Slack thread](https://gitpod.slack.com/archives/C02EN94AEPL/p1643911111366659) (internal)
- summary of earlier discussions [here](https://github.com/gitpod-io/gitpod/issues/7564#issuecomment-1012064597) and [here](https://github.com/gitpod-io/gitpod/issues/7564#issuecomment-1014922322)

Meanwhile, this Pull Request is nearly ready to be merged as is for the current iteration, except for this comment by @svenefftinge:

> I agree, with incrementally shipping things ASAP 🚀.
> To me it seems that visually this is below our threshold atm. cc @gtsiolis

Thus, will request another round of reviews:

- @gitpod-io/engineering-webapp: Please take a look at the code 🙏

- @gtsiolis / @jldec: Please take a look at the UX and make a call whether this meets our UX quality bar 🙏

(Personally, I really love URLs, so I'm quite happy of what the results look like -- but I'm also biased, and it seems a majority of Gitpodders actually prefer well-designed "rich results", so this could be a requirement for shipping.)

components/server/ee/src/user/eligibility-service.ts
```diff
@@ -93,22 +93,26 @@ export class EligibilityService {
...
 
+        const logCtx = { userId: user.id };
         try {
             const rawResponse = await fetch("https://education.github.com/api/user", {
                 headers: {
                     "Authorization": `token ${token}`,
                     "faculty-check-preview": "true"
                 }
             });
+            if (!rawResponse.ok) {
```
AlexTugarev, line 104: nice catch! I believe in that case it doesn't make sense to proceed with `await rawResponse.text()` as it will throw.

jldec:
> @gtsiolis / @jldec: Please take a look at the UX and make a call whether this meets our UX quality bar 🙏

This works for me from UX perspective - The intent of the feature is clear, the results are fast and meaningful and meet the requirements (also for pasted context URLs), and the design polish feels good - thanks @gtsiolis and @jankeromnes for iterating on that.

Seeing URLs where repositories are expected in the suggestions is not too surprising to stop us from shipping IMO, and is clearly the next thing we will address, give that @gtsiolis has already created designs.

/lgtm

PS @jankeromnes when I tried it just now, this feature in your PR preview environment was not returning any suggested URLs for me until I re-installed the github app with /new. I assume that this may be because you restarted or reconfigured something.

roboquat:
LGTM label has been added.  <details>Git tree hash: b1d5d2318e06c91211829ee530e0cada81efddff</details>

roboquat:

This pull-request has been approved by: *<a href="https://github.com/gitpod-io/gitpod/pull/7715#pullrequestreview-859729132" title="Approved">gtsiolis</a>*, *<a href="https://github.com/gitpod-io/gitpod/pull/7715#issuecomment-1029986766" title="LGTM">jldec</a>*

Associated issue: *#7564*

<details >

- ~~[components/dashboard/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/dashboard/OWNERS)~~ [gtsiolis,jldec]
- ~~[components/gitpod-protocol/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-protocol/OWNERS)~~ [gtsiolis,jldec]
- ~~[components/server/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/server/OWNERS)~~ [gtsiolis,jldec]

</details>

jankeromnes:
> PS @jankeromnes when I tried it just now, this feature in your PR preview environment was not returning any suggested URLs for me until I re-installed the github app with /new. I assume that this may be because you restarted or reconfigured something.

Hmm, this is unexpected. I believe I had properly configured the App with the latest deployment. Will have a look at the logs.

jldec:
@jankeromnes - I did some more digging and I looks like my GitHub integration token was bad (could not even start a workspace on public repo) - removing the GitHub integration and then re-connecting did the trick.

jldec (approved):

roboquat:

This pull-request has been approved by: *<a href="https://github.com/gitpod-io/gitpod/pull/7715#pullrequestreview-859729132" title="Approved">gtsiolis</a>*, *<a href="https://github.com/gitpod-io/gitpod/pull/7715#pullrequestreview-873456132" title="Approved">jldec</a>*

Associated issue: *#7564*

<details >

- ~~[components/dashboard/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/dashboard/OWNERS)~~ [gtsiolis,jldec]
- ~~[components/gitpod-protocol/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-protocol/OWNERS)~~ [gtsiolis,jldec]
- ~~[components/server/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/server/OWNERS)~~ [gtsiolis,jldec]

</details>

gtsiolis (commented):
Thanks again @jankeromnes for pushing this through the finish line! 🏁 

Cross-linking below the comments related to UX that we could improve before merging this. Up for discussing more about these or leaving any of these for the next iteration and out of the scope of this PR if needed. Cc @jldec 

1. [Empty state placeholder, text input design, modal background, and removal of the background page load.](https://github.com/gitpod-io/gitpod/pull/7715#discussion_r797990648)
2. [Results found text design](https://github.com/gitpod-io/gitpod/pull/7715#discussion_r797972968).
3. [Restoring empty state](https://github.com/gitpod-io/gitpod/pull/7715#discussion_r797969352).
4. [Text input placeholder](https://github.com/gitpod-io/gitpod/pull/7715#discussion_r797968321).

components/dashboard/src/components/RepositoryFinder.tsx
jankeromnes: Thanks @gtsiolis!

The goal of this was to give users a hint that "there are more items beyond this point" -- we actually expect the most relevant results to be among the first 15-20 items, but in some edge cases users might want to know if there are even more matches.

What do you think of an alternative phrasing like this one?

```
Showing 15/42 matches
```

components/dashboard/src/components/RepositoryFinder.tsx
jankeromnes: This is sort of an edge case, and I don't know what would be the best UX here.

I've implemented it like this because:
- The initial empty state teaches you about the feature, but once you've started typing I considered you've moved on from "what is this?" to "how does it work?" (hence no need for the empty state text anymore)
- I found it marginally helpful to know what results were available _without_ any filter: the current state allows me to know how many possible results there are, and to get an idea of what they might look like (e.g. "aha, these look like repositories -- some I've contributed to, and some examples")

Going back to the empty state could work, but this makes the "what are all possible results" edge use case harder to achieve (e.g. you'll need to type a common character like `h` or `/`, which is less discoverable).

Anyway, no strong opinion. Do you much prefer restoring the empty state on empty input?

components/dashboard/src/components/RepositoryFinder.tsx
jankeromnes: Thanks! I think this could make sense.

I generally like placeholders to look like something a user might actually type, to sort of guide possible inputs, but here I do find your proposal much clearer. 🎯

components/dashboard/src/components/RepositoryFinder.tsx
jankeromnes: Ah, I've just noticed the "2 more results found" text in https://github.com/gitpod-io/gitpod/pull/7715#discussion_r797990648

Using "{N} more results found" does work for me. 👍

gtsiolis: Thanks for tackling both comments, @jankeromnes! The designs in the comment mentioned in https://github.com/gitpod-io/gitpod/pull/7715#discussion_r800817010 are definitely more preferable.

Going with `{N} more results found` sounds perfect. 🏀 

Cross-posting from https://github.com/gitpod-io/gitpod/pull/7715#discussion_r797972968 for visibility and future reference:

> Ideally, we could paginate the results with an infinite scroll but sounds good to leave this out of the scope of this PR. However, there's definitely some value in surfacing the amount of results found.

components/dashboard/src/components/RepositoryFinder.tsx
gtsiolis: > I generally like placeholders to look like something a user might actually type, to sort of guide possible inputs, but here I do find your proposal much clearer.

Agree! 💯

components/dashboard/src/components/RepositoryFinder.tsx
jankeromnes: Thanks for making the call! 🤙 Will adjust this.

components/dashboard/src/components/RepositoryFinder.tsx
jankeromnes: Thanks! Went with 🅰️ for now. Will open a follow-up issue for the remaining design adjustments.

components/dashboard/src/workspaces/start-workspace-modal-context.tsx
gtsiolis, line 37: **suggestion**: Also, Sicne we're here one minor fine tuning we could do is use a better plus sign in terms of alignment, offset, and size. For example, `﹢` instead of `+`.

```suggestion
export const StartWorkspaceModalKeyBinding = `${/(Mac|iPhone|iPod|iPad)/i.test(navigator.platform) ? '⌘﹢' : 'Ctrl﹢'}O`;
```

| BEFORE | AFTER |
|-|-|
| plus-before | plus-after |

gtsiolis (commented):
Left two more minor comments regarding the keyboard shortcut inside the button label.

jankeromnes:
> Cross-linking below one last round of comments related to UX that we could improve before merging this.

Many thanks @gtsiolis for summarizing remaining concerns! Will address them or open follow-up issues.

> 1. [Empty state placeholder, text input design, modal background, and removal of the background page load.](https://github.com/gitpod-io/gitpod/pull/7715#discussion_r797990648)

Removed the background page load (🅰️ -- good catch!) and opened https://github.com/gitpod-io/gitpod/issues/8088 for the remaining design items.

> 2. [Results found text design](https://github.com/gitpod-io/gitpod/pull/7715#discussion_r797972968).

✅

> 3. [Restoring empty state](https://github.com/gitpod-io/gitpod/pull/7715#discussion_r797969352).

✅

> 4. [Text input placeholder](https://github.com/gitpod-io/gitpod/pull/7715#discussion_r797968321).

✅

> Left two more minor comments regarding the keyboard shortcut inside the button label.

⏳ TODO(janx)

components/dashboard/src/workspaces/start-workspace-modal-context.tsx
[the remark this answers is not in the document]
jankeromnes: When possible, I think it is wise to look at what similar tools (for example VS Code and JetBrains) are doing:

| OS \ Tool | VS Code | JetBrains |
| :---: | :---: | :---: |
| **Mac OS** | Screenshot 2022-02-08 at 14 08 43 | Screenshot 2022-02-08 at 14 08 18 |
| **Windows** | Screenshot 2022-02-08 at 14 11 39 | Screenshot 2022-02-08 at 14 13 10 |

Here, we can see that they chose to remain consistent with the local OS format, i.e.:

- Mac OS: `⌘O` (no plus, no space)
- Windows: `Ctrl+O` (no space)

I think it is a good idea to do the same in Gitpod. 😇

jankeromnes: (See [above comment](https://github.com/gitpod-io/gitpod/pull/7715#discussion_r801617528) about OS consistency.)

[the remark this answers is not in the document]
gtsiolis: DEAL 🤝

gtsiolis: Still, the better plus sign character could be used for Windows systems, right?

jankeromnes: Ah, I missed the plus sign 😅 honestly here I barely see any difference ("it's just letters" 😝)

Even my browser/system considers these characters equivalent, so it's not easy trying to differentiate them.

I believe VS Code uses a plain regular plus sign in [this file](https://github.com/microsoft/vscode/blob/6dffb2e35788e0814ef074e35df746d644b3603b/src/vs/base/common/keybindingLabels.ts#L68), right?

I'll switch to the new proposed plus sign, but I think this is asking for trouble (e.g. unreadable keybinding on some systems with less advanced unicode support) when we could be using something dead simple. 😋

jankeromnes: Done 😵

jankeromnes:
Okay, I think we've iterated enough here. 😅

Could a kind team mate please approve this Pull Request again? 🙏

JanKoehnlein (approved):
Let's ship this!

components/dashboard/src/components/RepositoryFinder.tsx
gtsiolis: Follow up issue can be found in https://github.com/gitpod-io/gitpod/issues/8088.

Issue #7608, opened 2022-01-14
Epic: Multi-repo Workspaces


svenefftinge:
## Goal

Enable workspace configuration consisting of multiple repositories without compromising on any of the contextual functionality of Gitpod. I.e. I want to be able to start a dev environment that has all the repos cloned and prebuild from any context of the participating repositories.

Additional discussion also in https://github.com/gitpod-io/gitpod/issues/7249

This should allow Gitpod to work well for scenarios where the code for different components of a project is not maintained in a monorepo, but rather in separate git repos. A common example is back-end services and front-end implementations for different target platforms.

See [this RFC](https://www.notion.so/gitpod/RFC-Multi-Repo-Support-247be8565ed14d8e88cf4147a740c9fc) for a deeper discussion about the approach.

## Proposed solution

The outlined solution will make it possible to compose multiple repos to one by introducing references in the `.gitpod.yml`. Conceptually it makes such multi-repo projects work very similar to how large mono-repos work.

- Introduce a `mainConfiguration: string` property on workspace config, that marks any sub repositories as such and points to a main repo that has a proper `.gitpod.yml`.
- Introduce a `additionalRepositories: string[]` property on workspaces config, that lists the sub repositories. They should also list the relative path to where the repo should be cloned, default is next to the main repo using the simple name of the sub-repository. So it’s a bi-directional relationship, that needs to be consistent. It’s necessary to be bi-directionally specified in order to be able to resolve all the participating repos.
- Make all clients of workspace starts aware of this composition, such that if I start a workspace on a sub repo, it creates the whole repo based on the context (see below). This includes prebuilds, i.e. users should install webhooks (resp. GH app) on all additional repositories, which will trigger a prebuild for the given context. Prebuilds are listed on the main repo, i.e. in Gitpod there is only one project for such multi-repo projects.

### Example Setup

```yaml
# Project A (https://github.com/my-org/project-a)
additionalRepositories:
 - url: https://github.com/my-org/sub1
   checkoutLocation: ./../sub1 # optional
image: 
   file: .gitpod.Dockerfile
tasks:
 - before: cd ./../sub1
   init: npm install
   command: npm run sub1
 - before: cd ../sub2
   init: npm install
   command: npm run sub2
```

```yaml
# Sub 1 (https://github.com/my-org/sub1)
mainConfiguration: https://github.com/my-org/project-a
```

```yaml
# Sub 2 (https://github.com/my-org/sub2)
[...]

PR #7785, opened 2022-01-23
Support for multi-repo projects

svenefftinge:
This PR introduces the possibility to describe dev environments that span multiple repositories. Such a project consists of a main repository, that contains the configuration for the project's dev environment using the `.gitpod.yml` as usual. The additional repositories don't have a configuration on their own but all point to the main repository's configuration. Here's a [loom video](https://www.loom.com/share/77a03b7800a340bca2c373601f021b29) showing this in action.

The pointing is done bi-directionally in the repositories  `.gitpod.yml`.
The main repository's `.gitpod.yml` contains a list of `additionalRepositories`.
### How to test (example contexts)

A team with projects is set up: [Join the team](https://se-multi-repo.staging.gitpod-dev.com/teams/join?inviteId=228ae250-a54b-4746-823b-24cf884f578f)

 | Repo | Context | |
 |---------|-|-|
 |[main repo](https://github.com/svenefftinge/multi-repo-main) | [default](https://github.com/svenefftinge/multi-repo-main) | [Open in Gitpod](https://se-multi-repo.staging.gitpod-dev.com/#https://github.com/svenefftinge/multi-repo-main)|
 |[main repo](https://github.com/svenefftinge/multi-repo-main) | [pull request](https://github.com/svenefftinge/multi-repo-main/pull/1) | [Open in Gitpod](https://se-multi-repo.staging.gitpod-dev.com/#https://github.com/svenefftinge/multi-repo-main/pull/1) |
 |[sub repo](https://github.com/svenefftinge/multi-repo-sub2) | [default](https://github.com/svenefftinge/multi-repo-sub2) | [Open in Gitpod](https://se-multi-repo.staging.gitpod-dev.com/#https://github.com/svenefftinge/multi-repo-sub2)|
 |[sub repo](https://github.com/svenefftinge/multi-repo-sub1) | [issue](https://github.com/svenefftinge/multi-repo-sub1/issues/1) | [Open in Gitpod](https://se-multi-repo.staging.gitpod-dev.com/#https://github.com/svenefftinge/multi-repo-sub1/issues/1)|
 |[sub repo](https://github.com/svenefftinge/multi-repo-sub1) | [pull request](https://github.com/svenefftinge/multi-repo-sub1/pull/4) | [Open in Gitpod](https://se-multi-repo.staging.gitpod-dev.com/#https://github.com/svenefftinge/multi-repo-sub1/pull/4)|

fixes https://github.com/gitpod-io/gitpod/issues/7608

### Things that knowingly are not perfect yet

Multi-repo projects don't have a great representation in the UI (dashboard) yet. One still has to install webhooks/GitHub App on all participating repositories to have the webhook trigger pushevents. Webhooks coming from subrepos are treated as events for the project that contains the main configuration. That's correct but i should be visually clearer, that the project spans multiple repos and also the detail information on prebuilds should list all participating commits.

The idea is to ship a first preview version to get more feedback from users and complete the feature based on what we learn from that.

### Documentation

TODO: Add Draft PR for Docs

### Release Notes
```release notes
Support for multi-repository setups
```


roboquat:

<details>

</details>

roboquat:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please assign **csweichel**, **geropl** after the PR has been reviewed.
You can assign the PR to them by writing `/assign @csweichel @geropl` in a comment when ready.

<details open>

- **[components/content-service/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/content-service/OWNERS)**
- **[components/content-service-api/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/content-service-api/OWNERS)**
- **[components/gitpod-db/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-db/OWNERS)**
- **[components/ws-daemon/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/ws-daemon/OWNERS)**

</details>

roboquat:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please assign **csweichel**, **geropl** after the PR has been reviewed.
You can assign the PR to them by writing `/assign @csweichel @geropl` in a comment when ready.

Associated issue: *#7608*

<details open>

- **[components/content-service/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/content-service/OWNERS)**
- **[components/content-service-api/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/content-service-api/OWNERS)**
- **[components/gitpod-db/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-db/OWNERS)**
- **[components/ws-daemon/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/ws-daemon/OWNERS)**

</details>

roboquat:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please assign **geropl** after the PR has been reviewed.
You can assign the PR to them by writing `/assign @geropl` in a comment when ready.

Associated issue: *#7608*

<details open>

- **[components/content-service-api/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/components/content-service-api/OWNERS)**

</details>

roboquat:

<details>

</details>

components/content-service/pkg/initializer/prebuild.go
```diff
@@ -80,61 +84,85 @@ func (p *PrebuildInitializer) Run(ctx context.Context, mappings []archive.IDMapp
...
+				anErr := runGitInit(ctx, gitinit)
+				if anErr != nil {
+					if err == nil {
+						err = anErr
+					} else {
+						log.Error("Git init for prebuild on "+gitinit.RemoteURI+" failed.", err)
+					}
+				}
+			}
+		}
```
csweichel, line 102: We can simplify the error handling here. Also, we should think about how we handle the case if the composite initializer is not what we expect it to be. Maybe, instead of using a composite initializer we should not generalise that much, and have a `repeated GitInitializer multi_repo` instead.

```suggestion
			gitinit, ok := init.(*GitInitializer)
			if !ok {
						// What do we do here? Return an error that this kind of setup isn't supported?
			}
			err := runGitInit(ctx, gitinit)
			if err != nil {
				break
			}
		}
```

components/content-service/pkg/layer/provider.go
```diff
@@ -184,7 +184,11 @@ func (s *Provider) GetContentLayer(ctx context.Context, owner, workspaceID strin
 			span.LogKV("fallback-to-git", err.Error())
 
 			// we failed creating a prebuild initializer, so let's try falling back to the Git part.
-			initializer = &csapi.WorkspaceInitializer{Spec: &csapi.WorkspaceInitializer_Git{Git: pis.Git}}
+			if pis.GetGit() == nil && pis.GetComposite() != nil {
+				initializer = &csapi.WorkspaceInitializer{Spec: &csapi.WorkspaceInitializer_Composite{Composite: pis.GetComposite()}}
+			} else {
+				initializer = &csapi.WorkspaceInitializer{Spec: &csapi.WorkspaceInitializer_Git{Git: pis.GetGit()}}
```
csweichel, line 190: `pis.Git()` can be `nil` here which would break things down the road

csweichel (commented):
Did not look through entirely yet, but I think we can make the prebuild initializer construct much simpler:

if we changed the interface to:
```proto
message PrebuildInitializer {
    SnapshotInitializer prebuild = 1;
    repeated GitInitializer git = 2;
}
```

AFAIK that would be a backwards compatible change on [serialisation level](https://developers.google.com/protocol-buffers/docs/proto3#updating), and it would drastically simplify the code.

csweichel:
Indeed, making a message `repeated` is backwards compatible:
image

csweichel:
@svenefftinge I've drafted how the proposed multi-Git prebuild initializer would look like: https://github.com/gitpod-io/gitpod/commit/2ddaa7129d9a660f2d69dd258538a2435d04336d

It indeed does simplify the code considerably.

components/gitpod-protocol/src/protocol.ts
```diff
@@ -567,7 +567,14 @@ export interface VSCodeConfig {
     extensions?: string[];
 }
 
+export interface RepositoryCloneInformation {
+    url: string;
+    checkoutLocation?: string;
+}
+
 export interface WorkspaceConfig {
+    mainConfiguration?: string;
```
csweichel, line 576: I'd propose we move this in a `multiRepo` section to keep te root namespace a bit cleaner, e.g.
```YAML
multiRepo:
  root: https://...
  children:
    - url: https://...
      checkoutLocation: foobar/
```

I'm not set on the naming at all, just tried to find something that's a tad shorter compared to what we have now. Current naming works, too:
```YAML
multiRepo:
  mainConfiguration: https://...
  additionalConfiguration:
    - url: https://...
      checkoutLocation: foobar/
```

components/gitpod-protocol/src/protocol.ts
```diff
@@ -567,7 +567,14 @@ export interface VSCodeConfig {
     extensions?: string[];
 }
 
+export interface RepositoryCloneInformation {
+    url: string;
+    checkoutLocation?: string;
```
csweichel, line 572: what happens when the checkoutLocation is undefined? Where do we clone the additional repo?

components/server/ee/src/prebuilds/github-app.ts
```diff
@@ -192,7 +205,20 @@ export class GithubApp {
...
-            let config = await this.prebuildManager.fetchConfig({ span }, user, contextURL);
+            const context = await this.contextParser.handle({ span }, user, contextURL) as CommitContext;
+            const config = await this.prebuildManager.fetchConfig({ span }, user, context);
+
+            // if it's a sub-repo of a multi-repo project, we look up the owner of the main repo
+            if (!!context.subRepositoryCheckoutInfo && (!project || context.repository.cloneUrl !== project.cloneUrl)) {
+                const owner = await this.findOwnerAndProject(installationId, context.repository.cloneUrl);
+                if (!owner) {
+                    log.info(`Did not find owner for cloneURL`, { cloneURL: context.repository.cloneUrl});
+                    return;
```
csweichel, line 216: What would the user see in this case? How would the system behave? Also, how would we observe this in the logs (metric instead?), i.e. is the log message actionable - if not, it should probably be a debug message?

components/server/ee/src/workspace/workspace-factory.ts
```diff
@@ -63,61 +62,28 @@ export class WorkspaceFactoryEE extends WorkspaceFactory {
             const imageSource = await this.imageSourceProvider.getImageSource(ctx, user, context.actual, config);
 
             // Walk back the commit history to find suitable parent prebuild to start an incremental prebuild on.
+            // TODO this doesn't work for multi-repo projects. We should fetch the last few available prebuilds and iterate over them instead.
```
csweichel, line 65: nit: please adapt to the TODO comment convention (https://www.notion.so/gitpod/How-we-develop-e8ce7e562478458a9223eb9b797415b2#86e669f8e7df4f1289c9bede0855b5f9)

components/server/ee/src/workspace/workspace-factory.ts
```diff
@@ -63,61 +62,28 @@ export class WorkspaceFactoryEE extends WorkspaceFactory {
...
+                        prebuiltWorkspace: match.prebuild,
+                    }
+                    ws = await this.createForPrebuiltWorkspace({span}, user, incrementalPrebuildContext, normalizedContextURL);
+                    // Overwrite the config from the parent prebuild:
+                    //   `createForPrebuiltWorkspace` 1:1 copies the config from the parent prebuild.
+                    //   Above, we've made sure that the parent's prebuild tasks (before/init/prebuild) are still the same as now.
+                    //   However, other non-prebuild config items might be outdated (e.g. any command task, VS Code extension, ...)
+                    //   To fix this, we overwrite the new prebuild's config with the most-recently fetched config.
+                    // See also: https://github.com/gitpod-io/gitpod/issues/7475
+                    //TODO doing side effects on objects back and forth is complicated and error-prone. We should rather make sure we pass in the config when creating the prebuiltWorkspace.
```
csweichel, line 83: nit: please adapt to the TODO comment convention (https://www.notion.so/gitpod/How-we-develop-e8ce7e562478458a9223eb9b797415b2#86e669f8e7df4f1289c9bede0855b5f9)

components/ws-daemon/pkg/content/service.go
```diff
@@ -282,8 +282,10 @@ func getCheckoutLocation(req *api.InitWorkspaceRequest) string {
 		}
 	}
 	if ir, ok := spec.(*csapi.WorkspaceInitializer_Prebuild); ok {
-		if ir.Prebuild != nil && ir.Prebuild.Git != nil {
-			return ir.Prebuild.Git.CheckoutLocation
+		if ir.Prebuild != nil {
+			if ir.Prebuild.GetGit() != nil {
```
csweichel, line 286: This does not prevent an out-of-bounds panic. The Git array not be nil, yet be empty. Better:

```
if len(ir.Prebuild.Git) > 0 {
    ...
}
```

csweichel (commented):
I very much like the overall approach - and judging from conversations with users this is likely to solve a very real problem.

There's a whole bunch of edge-cases we'll need to think through down the road, but for now, I'd say let's give this a go. Implementation-wise we're close - we'll need more tests though considering the complexity that's introduced.

components/gitpod-protocol/src/protocol.ts
svenefftinge: As it turns out these are not really bi-directional relations. `mainConfiguration` is simply delegating to a configuration using a context URL and doesn't necessarily imply that the main configuration holds a corresponding `additionalRepositories` entry. The `additionalRepositories` on the other hand is a list of repos to clone. It is thereby mainly an alternative to e.g. submodules that is aware of Gitpod context semantics (e.g. branches) and doesn't require a persistent pointer in the repo itself. So it makes sense to clone an additional repo without having a 'mainConfiguration' pointer on the other side. There are btw no use-cases I can think of where I would have both elements as in your example (but I guess it was just to clarify the schema you had in mind?).

components/gitpod-protocol/src/protocol.ts
svenefftinge: It defaults to the simple repository name, symmetrically to how we do it with the main repository.
See 
https://github.com/gitpod-io/gitpod/pull/7785/files#diff-af5f60bd3ac1a9b1771300c23f780b8fa11d821496acd6d8358739a1d802cb16R166

components/gitpod-protocol/src/protocol.ts
csweichel: > As it turns out these are not really bi-directional relations.

💯 agree. This would make the `root`/`children` naming unfavourable. I reckon we'd still want to move things into a `multiRepo` namespace, if just to keep the root of the `.gitpod.yml` clean(er).

svenefftinge: I don't see how this helps, as there is only ever one item used in a `.gitpod.yml` and also the `mainConfiguration` is just a delegation mechanism and not necessarily a `multi-repo` thing. Let's have a brief sync conversation about this later.

jankeromnes (commented):
Wow 🤯 this isn't the sort of Pull Request you review every day 👏

Thanks a lot for building native multi-repo support into Gitpod. This is a massive step forward towards building a good multi-repo story for Gitpod (whether this story ends up being native or not).

Now, whether we want to merge this native support (beyond any code nits) is, I believe, mainly a product/strategy decision. After reviewing this Pull Request, I see the following pros & cons:

👍 Pros of native multi-repo support in Gitpod:
- Helps multi-repo teams adopt Gitpod (makes Gitpod ✅ the "multi-repo" feature natively)
- Allows near out-of-the-box support for multi-repo context URLs (i.e. once the `.gitpod.yml`s are properly written and committed everywhere, opening any PR/issue should work fine)

👎 Cons of native multi-repo support in Gitpod:
- Adds significant complexity (i.e. significant risk of new bugs, and committed maintenance effort) to Gitpod's core (e.g. to workspace init, prebuilds, repository/team access control, etc)
- May have side effects on existing features (e.g. incremental prebuilds, maybe others too)
- Adds limited value to Gitpod, as multi-repo seems like a fringe use case (to me)
- I still believe multi-repo could also be adequately addressed with a simple, non-native solution*

*I'd be happy to build a proof-of-concept of a simple, non-native multi-repo solution (with minimal boilerplate code) so that we can compare what it can / cannot do, and whether the exclusive benefits of native outweigh the extra complexity and risk.

components/gitpod-protocol/data/gitpod-schema.json
```diff
@@ -136,13 +136,38 @@
             },
             "additionalProperties": false
         },
+        "additionalRepositories": {
```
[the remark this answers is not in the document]
svenefftinge: > How about childRepositories and parentRepository?

I had that and changed it because of feedback I got from others in the vain of "hey there are no primary and secondary repositories, it's just the one that contains the configuration"

components/server/src/workspace/context-parser-service.ts
```diff
@@ -70,6 +63,87 @@ export class ContextParser {
...
+        const span = TraceContext.startSpan("ContextParser.handleMultiRepositoryContext", ctx);
+        let config = await this.configProvider.fetchConfig({ span }, user, context);
+        let mainRepoContext: WorkspaceContext | undefined;
+        if (config.config.mainConfiguration) {
+            mainRepoContext = await this.internalHandleWithoutPrefix({ span }, user, config.config.mainConfiguration);
+            if (!CommitContext.is(mainRepoContext)) {
+                throw new InvalidGitpodYMLError([`Cannot find main repository '${config.config.mainConfiguration}'.`]);
+            }
+            config = await this.configProvider.fetchConfig({ span }, user, mainRepoContext);
+        }
```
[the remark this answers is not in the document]
svenefftinge: > You're probably going to ask if I'm a criminal again (😅)

oops, did I ask that? Sorry, it's an important question to raise. An initializer now can contain multiple git initializers instead of just one and if you don't have access, you cannot initialize a workspace. I have to check, though if we test this properly on prebuilt workspaces.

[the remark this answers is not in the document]
svenefftinge: > You're probably going to ask if I'm a criminal again (😅) 
😬 did I ask you that? Sorry, it is an important and very valid question. An initializer now can contain multiple git initializers instead of just one and if you don't have access, you cannot initialize a workspace. I have to check, though if we test this properly on prebuilt workspaces.

svenefftinge:
> I'd be happy to build a proof-of-concept of a simple,

Maybe just outlining your approach would be a great start 😁 
I'd love to make it as simple as possible. I do believe though that the current approach fits well with our design and is not that complicated (basically multiple repos & commits instead of a single one) and many of the code changes are in fact mostly necessary because the existing code base is a little grown. So I'd consider lots of the changes clean-ups rather than adding complexity, tbh. 😇 

Anyhow, I'm personally also "team monorepo" but I have learned that there are too many (I don't have data, but guess ~50%) teams who are not. Convincing them to change their repo-split is a hard sell (been there the last two years), so I really want us to have a solution for them rather than a religious conversation. We tried with git submodules or homegrown external solutions and they break the key product features of gitpod (context URLs, prebuilds and with that ephemeral workspaces). If you find a simpler solution that meet these requirements I'm very interested. Have you seen [the RFC](https://www.notion.so/gitpod/RFC-Multi-Repo-Support-247be8565ed14d8e88cf4147a740c9fc) where this has been discussed.

components/server/src/workspace/context-parser-service.ts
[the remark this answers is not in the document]
jankeromnes: > An initializer now can contain multiple git initializers instead of just one and if you don't have access, you cannot initialize a workspace.

Makes sense, thanks for explaining.

> I have to check, though if we test this properly on prebuilt workspaces.

👍

> 😬 did I ask you that?

Yes 😂 (but no worries, I wasn't offended 👍). I believe it was during a meeting: you had just proposed a new design idea for something, and I pointed out how it could be potentially abused by nefarious users, and so you asked if I was a criminal because the first thing I say about new ideas is always how they can be abused 😅 (note: I also have strong ethical values that prevent me from abusing things in a bad way, but I do like robust/fail-proof designs, and having a hacker/security-oriented mindset has helped me out many times).

components/gitpod-protocol/src/protocol.ts
csweichel: In practice no one will have both entries - let's skip the joint parent category.

svenefftinge:
> Shouldn't we try to checkout a branch with the same name from all repos and just fall back to main?

Yes, we should. I'll look into that.

components/server/ee/src/workspace/workspace-factory.ts
```diff
@@ -63,61 +62,28 @@ export class WorkspaceFactoryEE extends WorkspaceFactory {
...
 
             // Walk back the commit history to find suitable parent prebuild to start an incremental prebuild on.
+            // TODO this doesn't work for multi-repo projects. We should fetch the last few available prebuilds and iterate over them instead.
             let ws;
-            for (const parent of (context.commitHistory || [])) {
-                const parentPrebuild = await this.db.trace({span}).findPrebuiltWorkspaceByCommit(commitContext.repository.cloneUrl, parent);
-                if (!parentPrebuild) {
-                    continue;
+            if (context.commitHistory && context.commitHistory.length > 0) {
+                const recentPrebuilds = await this.db.trace({span}).findPrebuildsWithWorkpace(commitContext.repository.cloneUrl, 20);
```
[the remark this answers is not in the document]
svenefftinge: I have removed the limit and added a predicate to only return non deleted workspaces.

csweichel (dismissed):
Let's do this

components/gitpod-db/src/typeorm/migration/1644403519382-PrebuildUpdatableSHA.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+import {MigrationInterface, QueryRunner} from "typeorm";
+import { columnExists } from "./helper/helper";
+
+export class PrebuildUpdatableSHA1644403519382 implements MigrationInterface {
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        if (!(await columnExists(queryRunner, "d_b_prebuilt_workspace_updatable", "commitSHA"))) {
+            await queryRunner.query("ALTER TABLE d_b_prebuilt_workspace_updatable ADD COLUMN commitSHA varchar(255) NULL");
```
geropl, line 14: As we use `Transformer.MAP_EMPTY_STR_TO_UNDEFINED` to map `undefined` to `""` (which is the only sound way to encode undefined we know of) the column should be `NOT NULL`.

components/gitpod-protocol/data/gitpod-schema.json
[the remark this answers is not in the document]
geropl: Just wanted to leave an upvote for `parent`/`child`: it immediately conveys the meaning. And reflects the (default) file system layout.

components/gitpod-protocol/src/protocol.ts
```diff
@@ -686,6 +693,7 @@ export interface PrebuiltWorkspaceUpdatable {
     repo: string;
     isResolved: boolean;
     installationId: string;
+    commitSHA?: string;
```
geropl, line 696: What is this `commitSHA`? A comment would be golden. :upside_down_face:

components/gitpod-protocol/src/protocol.ts
```diff
@@ -968,9 +977,47 @@ export namespace AdditionalContentContext {
     }
 }
 
-export interface CommitContext extends WorkspaceContext, Commit {
+export interface CommitContext extends WorkspaceContext, GitCheckoutInfo {
```
geropl, line 980: :+1:

components/gitpod-protocol/src/protocol.ts
```diff
@@ -968,9 +977,47 @@ export namespace AdditionalContentContext {
...
+     * @param commitContext
+     * @returns identifier for set of commits
+     */
+    export function getCommitSHAs(commitContext: CommitContext): string {
+        if (!commitContext.additionalRepositoryCheckoutInfo || commitContext.additionalRepositoryCheckoutInfo.length === 0) {
+            return commitContext.revision;
+        }
+        // it's a multi-repo project, we need to store all commit shas and that needs to fit into 255 chars. Let's only store at most first seven.
+        const commits = [commitContext.revision, ... commitContext.additionalRepositoryCheckoutInfo.map(i => i.revision)];
+        const maxLength = Math.min(7, 255 / commits.length);
```
geropl, line 1003: Hm. It feels strange to have a hard limit here.

Why not use a hash function over all commits SHAs? For instance:
```
const hasher = createHash('sha256');
for (const revision of revisions) {
   hasher.update(revision);
}
return hasher.digest('hex');
```

components/gitpod-protocol/data/gitpod-schema.json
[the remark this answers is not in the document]
svenefftinge: The default file system layout is not hierarchical, but flat (i.e. siblings)

components/gitpod-protocol/src/protocol.ts
svenefftinge: We need to know the actual commit shas later (for incremental prebuilds), so the string is unparsed again. Ideally, we would store commit objects, that consist of the repo and the commit information.

components/gitpod-protocol/data/gitpod-schema.json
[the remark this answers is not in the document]
geropl: > The default file system layout is not hierarchical, but flat (i.e. siblings)

I thought I saw a "/workspace/parent-repo/sub-repo" folder structure in the loom video - but now those are siblings. :+1:

components/server/src/workspace/workspace-starter.ts
```diff
@@ -486,20 +488,27 @@ export class WorkspaceStarter {
...
+                    if (initializer instanceof CompositeInitializer) {
+                        // we use the first git initializer for image builds only
+                        git = initializer.getInitializerList()[0].getGit()!;
+                    } else {
+                        git = initializer;
+                    }
                     git.setCloneTaget(imgsrc.dockerFileSource.revision);
                     git.setTargetMode(CloneTargetMode.REMOTE_COMMIT);
-                    checkoutLocation = "."
-                    git.setCheckoutLocation(checkoutLocation);
```
geropl, line 502: Why do we no longer set the `checkoutLocation` for image builds? :thinking: Does it default to `.` anyway?
Also: why was the side-effect here required in the first place? :scratch:

components/server/src/workspace/workspace-starter.ts
```diff
@@ -951,17 +960,29 @@ export class WorkspaceStarter {
 
             const snapshot = new SnapshotInitializer();
             snapshot.setSnapshot(context.snapshotBucketId);
-            const { git } = await this.createGitInitializer(traceCtx, workspace, context, user);
+            const { initializer } = await this.createCommitInitializer(traceCtx, workspace, context, user);
             const init = new PrebuildInitializer();
             init.setPrebuild(snapshot);
-            init.setGit(git);
+            if (initializer instanceof CompositeInitializer) {
```
geropl, line 966: Using this API seems noisy... is there a reason that not we have non-composite initializers? And if there is: Could we encoded that with a boolean flag, and get rid of these `instanceof` branches by making all initializers composites, and just tick that flag?

components/server/src/workspace/workspace-starter.ts
```diff
@@ -486,20 +488,27 @@ export class WorkspaceStarter {
             if (WorkspaceImageSourceDocker.is(imgsrc)) {
                 let source: WorkspaceInitializer;
                 const disp = new DisposableCollection();
-                let checkoutLocation = this.getCheckoutLocation(workspace);
+                let checkoutLocation = CommitContext.is(workspace.context) && workspace.context.checkoutLocation || '.';
```
geropl, line 491: Why is it ok to drop `workspace.config.checkoutLocation` here? :thinking:

components/server/src/workspace/context-parser-service.ts
```diff
@@ -70,6 +63,88 @@ export class ContextParser {
         return result;
     }
 
+    protected async internalHandleWithoutPrefix(ctx: TraceContext, user: User, nonPrefixedContextURL: string): Promise<WorkspaceContext> {
+        const span = TraceContext.startSpan("ContextParser.internalHandle", ctx);
```
geropl, line 67: This span should be `span.finish()` at some point.

components/server/src/workspace/context-parser-service.ts
```diff
@@ -70,6 +63,88 @@ export class ContextParser {
...
+            return undefined;
+        }
+        return upstreamCloneUrl;
+    }
+
+    protected async handleMultiRepositoryContext(ctx: TraceContext, user: User, context: WorkspaceContext): Promise<WorkspaceContext> {
+        if (!CommitContext.is(context)) {
+            return context;
+        }
+        const span = TraceContext.startSpan("ContextParser.handleMultiRepositoryContext", ctx);
```
geropl, line 103: Same here: missing `span.finish()`

components/server/ee/src/prebuilds/github-app.ts
```diff
@@ -211,6 +238,16 @@ export class GithubApp {
         }
     }
 
+    private async getCommitInfo(user: User, repoURL: string, commitSHA: string) {
```
geropl, line 241: I think we're loosing context here: Before, we'd effectively also pass branch here. This part is crucial, but not deduceable just from the repo and revision (same commit can be head of multiple branches).

components/server/ee/src/prebuilds/gitlab-app.ts
```diff
@@ -124,6 +123,16 @@ export class GitLabApp {
         }
     }
 
+    private async getCommitInfo(user: User, repoURL: string, commitSHA: string) {
```
geropl, line 126: [Same](https://github.com/gitpod-io/gitpod/pull/7785#discussion_r814631171) here.

components/server/ee/src/prebuilds/bitbucket-app.ts
```diff
@@ -88,25 +93,24 @@ export class BitbucketApp {
...
             }
 
-            console.debug('Bitbucket push event: Starting prebuild.', { contextURL });
-
+            console.log('Starting prebuild.', { contextURL })
+            const {host, owner, repo} = RepoURL.parseRepoUrl(data.repoUrl)!;
+            const hostCtx = this.hostCtxProvider.get(host);
+            let commitInfo: CommitInfo | undefined;
+            if (hostCtx?.services?.repositoryProvider) {
+                commitInfo = await hostCtx.services.repositoryProvider.getCommitInfo(user, owner, repo, data.commitHash);
```
geropl, line 109: [Same](https://github.com/gitpod-io/gitpod/pull/7785#discussion_r814631171) here.

geropl:
Finished the code review, these are the important comments (by decreasing importance):
 - https://github.com/gitpod-io/gitpod/pull/7785#discussion_r814631171
 - https://github.com/gitpod-io/gitpod/pull/7785#discussion_r814591575
 - https://github.com/gitpod-io/gitpod/pull/7785#discussion_r814604250

Did some basic testing, but did **not** test prebuilds.

components/gitpod-protocol/data/gitpod-schema.json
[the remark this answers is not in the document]
jankeromnes: Orthogonally to the `parent`/`child` vs `main`/`additional` discussion (which I no longer have strong feelings about), my main gripe with the current naming is actually with the word `Configuration`:
- https://github.com/svenefftinge/multi-repo-main is _not_ a configuration, but a repository. Calling it a configuration leads to confusion.
- Trying `mainConfiguration: https://github.com/svenefftinge/multi-repo-main/blob/main/.gitpod.yml` will _not_ work as the naming suggests.
- Calling one property `...Configuration` and the other `...Repositories` is also confusing:
  - They don't appear to use the same type (one wants repository URLs, the other wants... a configuration? 🤔)
  - They don't look related at all (even though you're mostly supposed to use them together to get multi-repo support)

What do you think of alternatives like `mainRepository`, or `configuringRepository`, or `configurationRepository`?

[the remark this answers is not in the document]
svenefftinge: After this we might likely introduce multiple `.gitpod.yml` per repo (i.e. the other side) to support opening workspaces on subparts of a large repo (big mono-repos or example projects). I have a use case in mind, where I would have multiple `.gitpod.yml` in one repo where some delegate to the same `mainConfiguration` (sitting in the root of the same repo) but would overlay properties such as `workspaceLocation` to open the IDE on a subpart of the repo's contents.

components/gitpod-protocol/src/protocol.ts
jankeromnes: Today I learned that `String.substring(int, float)` actually works in JavaScript. 🙈

Still, I'd feel more reassured without floating-point length variables:

```suggestion
        const maxLength = Math.min(7, Math.floor(255 / commits.length));
```

components/server/src/workspace/context-parser-service.ts
```diff
@@ -70,6 +63,88 @@ export class ContextParser {
...
+            const subRepoCommits: GitCheckoutInfo[] = [];
+            for (const subRepo of config.config.additionalRepositories) {
+                let subContext = await this.internalHandleWithoutPrefix({ span }, user, subRepo.url) as CommitContext;
+                if (!CommitContext.is(subContext)) {
+                    throw new InvalidGitpodYMLError([`Cannot find sub-repository '${subRepo.url}'.`]);
+                }
+                if (context.repository.cloneUrl === subContext.repository.cloneUrl) {
+                    // if it's the repo from the original context we want to use that commit.
+                    subContext = JSON.parse(JSON.stringify(context));
+                }
```
jankeromnes, line 124: Is there any point in allowing a repository to specify itself as an additional repository? 🤔 (And then fetch the same commit history multiple times etc?) Naively, I would want to just "skip" this case:

```suggestion
                if (context.repository.cloneUrl === subContext.repository.cloneUrl) {
                    continue;
                }
```

components/gitpod-protocol/src/protocol.ts
```diff
@@ -968,9 +977,47 @@ export namespace AdditionalContentContext {
...
+     * Creates an identifier pointing to the commits of the CommitContext and all sub-repo commit infos.
+     * The identifier is max 255 chars long.
+     * @param commitContext
+     * @returns identifier for set of commits
+     */
+    export function getCommitSHAs(commitContext: CommitContext): string {
+        if (!commitContext.additionalRepositoryCheckoutInfo || commitContext.additionalRepositoryCheckoutInfo.length === 0) {
+            return commitContext.revision;
+        }
+        // it's a multi-repo project, we need to store all commit shas and that needs to fit into 255 chars. Let's only store at most first seven.
```
jankeromnes, line 1001: I initially misread this as "let's remember only the first 7 additional repos / commits" 🙈

```suggestion
        // it's a multi-repo project, we need to store all commit shas and that needs to fit into 255 chars. Let's only store at most the first seven chars of each commit sha.
```

components/gitpod-protocol/src/protocol.ts
jankeromnes, line 1003: By limiting our SHA length to 7 characters or less, we accept the following risks of collisions (sources: [font-v](https://github.com/source-foundry/font-v/issues/2#issuecomment-327189455), [solid](https://www.getsolid.io/blog/birthday-paradox-coding-solid/), [Wikipedia](https://en.wikipedia.org/wiki/Birthday_problem#Reverse_problem)):

```ts
function collisionRisk (probability, digits) {
  const d = Math.pow(16, digits);
  return Math.sqrt(2 * d * Math.log(1 / (1 - probability)));
}
```

| Probability of collision happening | Number of prebuilds needed with length 7 | with length 6 | with length 5 | with length 4 | with length 3 |
| --- | --- | --- | --- | --- | --- |
| 1% | 2323 | 581 | 146 | 37 | 10 |
| 10% | 7521 | 1881 | 471 | 118 | 30 |
| 50% | 19291 | 4823 | 1206 | 302 | 76 |
| 90% | 35160 | 8790 | 2198 | 550 | 138 |
| 99.99% | 70320 | 17580 | 4395 | 1099 | 275 |

For reference:
- The Gitpod repo has 15355 prebuilds
- The GitLab repo has 79099 prebuilds

If a collision happens, we very subtly break incremental prebuilds, by considering a random old prebuild to be an ancestor even though it's not (e.g. a random old prebuild from a work-in-progress Pull Request suddenly becomes the new base 💣).

svenefftinge: That should actually have been `Math.*max*(7, ...)` in which case collision probability would be the same as now. But I'll change this to store the full commit in a new field, because I think the parse/unparse contract to squeeze multiple commits into 255 chars is prone to missinterpretation and therefore errors.

components/server/ee/src/workspace/workspace-factory.ts
```diff
@@ -62,62 +61,28 @@ export class WorkspaceFactoryEE extends WorkspaceFactory {
...
 
-            // Walk back the commit history to find suitable parent prebuild to start an incremental prebuild on.
+            // Walk back the last prebuilds and check if they are valid ancestor.
             let ws;
-            for (const parent of (context.commitHistory || [])) {
-                const parentPrebuild = await this.db.trace({span}).findPrebuiltWorkspaceByCommit(commitContext.repository.cloneUrl, parent);
-                if (!parentPrebuild) {
-                    continue;
+            if (context.commitHistory && context.commitHistory.length > 0) {
+                const recentPrebuilds = await this.db.trace({span}).findPrebuildsWithWorkpace(commitContext.repository.cloneUrl);
```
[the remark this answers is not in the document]
svenefftinge: Thanks, I have adjusted the query to filter for `softDeleted` which makes it return around 600 for gitlab and ~250 for our repo.

jldec:
I spent some time comparing the experience of using Gitpod with a monorepo vs. multiple repos with (and without) the features in this PR. Some links to my test repos are included in the [draft docs preview](https://deploy-preview-1717--gitpod-kumquat.netlify.app/docs/multi-repo-workspaces).

#### TL;DR
The new features as implemented make the experience with multiple repos almost as nice and seamless as a using a monorepo. This is a significant improvement over the experience of using multiple workspaces for multiple repos.

#### Benefits
- Cheaper (1 hour = 1 hour)
- Easier to configure (esp. automatically)
- Shared file system is useful e.g. for build + test
- Easier to work on multiple components at the same time

#### Areas that could be improved or need more testing
- When starting a workspace or openig the IDE it's not easy to see that you're in a multi-repo setup.
- It's unclear what to do to set up webhooks and prebuilds with a Gitpod project.
- For reasons which are unclear, I have not been able test automatic prebuilds in my dev environment.
- I have not tested branch-name matching.

/lgtm
/hold

jldec (approved):
lgtm

components/server/ee/src/prebuilds/github-app.ts
svenefftinge: The `context` has that information. The commit info is providing additional user-facing information, that we can show in the UI.

components/server/ee/src/prebuilds/gitlab-app.ts
svenefftinge: see my reply on the original comment

components/server/src/workspace/context-parser-service.ts
svenefftinge: The case is not, that someone created a self-pointer but that the context URL points to the main repo.

components/server/src/workspace/workspace-starter.ts
svenefftinge: I'm setting it in line 498

components/server/src/workspace/workspace-starter.ts
svenefftinge: Yes, it is noisy. We need the compositeinitializer for non-prebuilds. So in case of prebuilds we decompose the structure and translate to the prebuild initializer shape.

components/server/src/workspace/workspace-starter.ts
svenefftinge: The checkoutLocation from the config is respected when the context is created.

components/gitpod-protocol/src/protocol.ts
svenefftinge: I was able change this to the suggested hash, as I now use the context of the workspace to get the commits.

csweichel (approved):

jldec:
Tested prebuilds and projects with a 2-repo setup as per [draft docs](https://deploy-preview-1717--gitpod-kumquat.netlify.app/docs/multi-repo-workspaces)  

- [fullstack-frontend](https://github.com/jldec/fullstack-frontend/blob/main/.gitpod.yml) - main configuration
- [fullstack-backend](https://github.com/jldec/fullstack-backend/blob/main/.gitpod.yml)
- created projects on a team for both repos
- pushed to branch on backend - saw prebuild in frontend project - no prebuilds visible in backend project
- correctly opened workspace using branch and PR contexts from backend repo
- pushed to matching branches on both repos
- correctly opened contexts with matching branches from both frontend and backend

## project list - note backend branch name in frontend project tile 
Screenshot 2022-03-10 at 09 43 22

## all prebuilds on main repo project
Screenshot 2022-03-10 at 09 43 50

## no prebuilds on sub-repo project
Screenshot 2022-03-10 at 10 02 58

## matching branches
Screenshot 2022-03-10 at 09 55 58

jldec (approved):
LGTM based on testing in dev environment.

components/gitpod-db/src/typeorm/migration/1646803519382-PrebuildUpdatableSHA.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+import {MigrationInterface, QueryRunner} from "typeorm";
+import { columnExists } from "./helper/helper";
+
+export class PrebuildUpdatableSHA1646803519382 implements MigrationInterface {
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        if (!(await columnExists(queryRunner, "d_b_prebuilt_workspace_updatable", "commitSHA"))) {
+            await queryRunner.query("ALTER TABLE d_b_prebuilt_workspace_updatable ADD COLUMN commitSHA varchar(255) NOT NULL");
```
jankeromnes, line 14: What happens if you add a `NOT NULL` column to an existing table, without a default value? 🤔

Maybe this should be `NULL`, since `commitSHA` can be `undefined` in `DBPrebuiltWorkspaceUpdatable`.

jldec: addressed in https://github.com/gitpod-io/gitpod/pull/8711

Issue #7023, opened 2021-12-02
Improve automated rollout of JetBrains IDE's


loujaybee:
We currently are raising a new PR [[1](https://github.com/gitpod-io/gitpod/pull/7004)] for each new JetBrains IDE configuration / version bump. This ticket is to investigate the possibility of improving the process, to ensure that we have the latest IDE's rolled out into production in an efficient, stable manner.

corneliusludmann:
This especially makes sense when we introduce early-adopters versions ([internal discussion](https://gitpod.slack.com/archives/C01KGM9BH54/p1638442118354100?thread_ts=1638436313.350800&cid=C01KGM9BH54)). Proposal how this can be achieved:
- Make sure that the PRs are merged automatically. We need to figure out how this is possible without the needed reviews/approvals.
- Let the images be tagged as `latest` (e.g. [here](https://github.com/gitpod-io/gitpod/blob/81a86c281943aafe281911e71862b33c18240e36/components/ide/jetbrains/image/BUILD.yaml#L70-L72))
- Set `ResolveImageDigest` to true (similar to what we do for [VS Code Insiders here](https://github.com/gitpod-io/gitpod/blob/81a86c281943aafe281911e71862b33c18240e36/installer/pkg/components/server/ide-configmap.go#L62))

loujaybee:
Ideally we'll split this one down. @akosyakov and @iQQBot have also discussed some plans for this work.

akosyakov:
I think it can be break down to following issues:
- [ ] Provide EAP versions support which is built and delivered nightly similarly to VS Code Insiders. It should be an optional feature from user perspective, by default off.
- [ ] Enable smoke testing of the entire integration during the build including gateway and backend plugins.
- [ ] After that we can enable automatic upgrade for stable versions.
- [ ] And as the last add all JetBrains products supporting remote development.

akosyakov:
@iQQBot I reopen it since the issue is auto rollout of all products

loujaybee:
@iQQBot / @akosyakov - can we close this one? What else is remaining here?

iQQBot:
remaining two things
- [ ] After that we can enable automatic upgrade for stable versions.
- [ ] And as the last add all JetBrains products supporting remote development.

Issue #6910, opened 2021-11-25
Improve more actions button on the workspace start page when opening in desktop IDEs


gtsiolis:
### Problem to solve

Back in https://github.com/gitpod-io/gitpod/issues/6841 and https://github.com/gitpod-io/gitpod/pull/6847, we've added a dropdown button with more options besides opening a workspace in VS Code in Browser. Cc @corneliusludmann @akosyakov @JanKoehnlein 

The current implementation brings in the following issues:
1. Redandunt right padding on the secondary button (see down chevron icon)
2. Center aligned dropdown options which is not following the pattern used in other dropdowns.
3. Inaccurate footer text as not all options above are IDE options.
4. Option order could affect memory muscle as usually this button links back to the dashboard while in the current implementation the option to open the dashboard has been moved to the bottom of the list.

### Proposal

TBD (To be discussed)

gtsiolis:
@corneliusludmann I've added the design specs in the issue description. Feedback is welcome! 🍊 

**fyi**: The dropdown button will be used in a few other places like in https://github.com/gitpod-io/gitpod/issues/6826#issuecomment-985039468. In the next iteration we could improve this page by including the context but this could be considered out of the scope of this issue.

PR #7852, opened 2022-01-26
Allow user select eap version of JetBrains IDE

iQQBot:
## Description

Allow user select eap version of JetBrains IDE

## Related Issue(s)

Relate #7023 

## How to test

We can check current Desktop IDE version by: Start a workspace, run `cat /ide-desktop/backend/product-info.json | jq`
1. Go to preference, select Pycharm, check version
2. Go to preference, select Pycharm and check **Include JetBrains EAP** then check version

## Release Notes

```release-note
NONE
```


roboquat:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please ask for approval from **iqqbot** and additionally assign **laushinka** after the PR has been reviewed.
You can assign the PR to them by writing `/assign @laushinka` in a comment when ready.

Associated issue: *#7023*

<details open>

- **[installer/pkg/components/server/ide/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/installer/pkg/components/server/ide/OWNERS)**

</details>

roboquat:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please ask for approval from **iqqbot** and additionally assign **jankeromnes** after the PR has been reviewed.
You can assign the PR to them by writing `/assign @jankeromnes` in a comment when ready.

Associated issue: *#7023*

<details open>

- **[installer/pkg/components/server/ide/OWNERS](https://github.com/gitpod-io/gitpod/blob/main/installer/pkg/components/server/ide/OWNERS)**

</details>

akosyakov:
@gtsiolis @loujaybee could you review UI/UX 🙏 please

components/server/src/ide-config.ts
```diff
@@ -190,6 +191,23 @@ export class IDEConfigService {
                 }
             }
 
+            for (const [id, option] of Object.entries(newValue.ideOptions.options).filter(([_, x]) => x.latestImage)) {
+                if (option.latestImage == null || option.latestImage === "") {
```
akosyakov, line 195: `!option.latestImage`

components/gitpod-protocol/src/protocol.ts
```diff
@@ -119,6 +119,7 @@ export type IDESettings = {
     defaultIde?: string
     useDesktopIde?: boolean
     defaultDesktopIde?: string
+    useLatestVersion?: boolean
```
akosyakov, line 122: `useLatestDesktopVersion`

Maybe we should make it even JB specific.

components/server/src/ide-config.ts
andreafalzetti, line 195:
```suggestion
                if (option.latestImage === null || option.latestImage === "") {
```

I think it's safer, given that `undefined == null` => `true`

components/server/src/ide-config.ts
```diff
@@ -190,6 +191,23 @@ export class IDEConfigService {
                 }
             }
 
+            for (const [id, option] of Object.entries(newValue.ideOptions.options).filter(([_, x]) => x.latestImage)) {
```
andreafalzetti, line 194: this for loop is fairly complex, with many things happening in a single line - did you consider breaking it down a bit?

components/server/src/workspace/workspace-starter.ts
```diff
@@ -348,7 +348,8 @@ export class WorkspaceStarter {
                 if (!!desktopIdeChoice) {
                     const mappedImage = ideConfig.ideOptions.options[desktopIdeChoice];
                     if (!!mappedImage && mappedImage.image) {
-                        configuration.desktopIdeImage = mappedImage.image;
+                        const useLatest = !!user.additionalData?.ideSettings?.useLatestVersion
```
andreafalzetti, line 351:
```suggestion
                        const useLatest = !!user.additionalData?.ideSettings?.useLatestVersion;
```

components/server/src/ide-config.ts
iQQBot: This line doesn't really work, the real check is in the previous line, here it's just to remove the typescript error, of course can replace it with @ts-ignore

iQQBot: This line doesn't really work, the real check is in the previous line, here it's just to remove the typescript error, of course can replace it with @ts-ignore

components/server/src/workspace/workspace-starter.ts
iQQBot: thanks

akosyakov:
@iQQBot Let's put a checkbox in less prominent location like under the header or even under all desktop IDEs. Most of users don't need it. 

Is it possible to show concrete versions as label? Do we have this information in the metadata or can we resolve it somehow from docker image? Can we get it?

gtsiolis:
Looking at this now! 👀

gtsiolis:
Thanks for the ping @iQQBot! FWIW, I agree with @akosyakov in https://github.com/gitpod-io/gitpod/pull/7852#issuecomment-1023164127 about moving the checkbox to a less prominent location.

This is also something we could use to consolidate all options including VS Code editors but let's leave this out of the scope of this PR. Cc @akosyakov @loujaybee 

Adding below a design for this. I've removed the rest of the preferences elements to focus on the desktop editors. Feedback is welcome! 🏀 

EAP

akosyakov:
> This is also something we could use to consolidate all options including VS Code editors but let's leave this out of the scope of this PR. Cc @akosyakov @loujaybee

Let's think about user stories here. I am not sure that's it all the same. There can be cases when you want to use stable for one, and laster for another. Maybe later we remove this big boxes and more have a drop down and add all versions there, stable and latest. Right now it is just for us to test against next version.

components/gitpod-protocol/src/protocol.ts
iQQBot: If we need to add another vendor's IDE later, do we need to get a separate variable for each one? It doesn't seem like we should.

akosyakov:
> Thanks for @gtsiolis the design plan, @akosyakov Do you think it's OK to follow the above design? If so I'll follow the design and implement it

Yes, it looks good and enable us to test latest versions.

installer/pkg/components/server/ide/configmap.go
```diff
@@ -92,36 +92,40 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
...
-					Logo:     getIdeLogoPath("phpstormLogo"),
-					Notes:    []string{"While in beta, when you open a workspace with PhpStorm you will need to use the password “gitpod”."},
-					Image:    common.ImageName(ctx.Config.Repository, ide.PhpStormDesktopIdeImage, ctx.VersionManifest.Components.Workspace.DesktopIdeImages.PhpStormImage.Version),
+					OrderKey:    pointer.String("07"),
+					Title:       "PhpStorm",
+					Type:        typeDesktop,
+					Logo:        getIdeLogoPath("phpstormLogo"),
+					Notes:       []string{"While in beta, when you open a workspace with PhpStorm you will need to use the password “gitpod”."},
+					Image:       common.ImageName(ctx.Config.Repository, ide.PhpStormDesktopIdeImage, ctx.VersionManifest.Components.Workspace.DesktopIdeImages.PhpStormImage.Version),
+					LatestImage: common.ImageName(ctx.Config.Repository, ide.PhpStormDesktopIdeImage, "latest"),
```
csweichel, line 128: Will those image references be "resolved" (e.g. using oci-tool in `server`) or are they passed to the IDE spec like this? 
In case of the latter, this is unlikely to behave as you expect it to. Becasue the IDE image is served by registry-facade and blobserve it's heavily cached, i.e. not invalidated over the lifetime of a registry-facade/blobserve pod.

Both components make the assumption that tags are immutable - a good assumption to make overall. We should not rely on "moving tags" such as `latest`.

iQQBot: https://github.com/gitpod-io/gitpod/blob/ee808018184687b84ef53d7ea4999f394b8a880d/components/server/src/ide-config.ts#L194-L209

Yes, the latest image will be "resolved"

loujaybee:
> I've removed the rest of the preferences elements to focus on the desktop editors.

No challenges with the proposed design for desktop.

However, if we implement as suggested then the preferences page (for desktop) will become inconsistent to the VS Code preferences section above (for browser) where two UX for choosing the latest/EAP versions are provided. If we go with the UX suggested, I'd suggest we also update the VS Code preferences to the same checkbox UX for insiders (or we create a ticket for it).

gtsiolis:
> I've removed the rest of the preferences elements to focus on the desktop editors.

@loujaybee To clarify, the removal was only part of the review comment to focus on the changes in the scope of thiS . The proposal was not to consolidate browser and desktop editors in this PR, as discussed last week in our meeting with @akosyakov.

> I'd suggest we also update the VS Code preferences to the same checkbox UX for insiders

Agree. Opening a follow up issue sounds also good.

🍊 🍊 🍊 🍊 

Long-term I still think consolidating the editors and using the latest and desktop options below could be better UX.

Less choices, more control[[1](https://en.wikipedia.org/wiki/Miller%27s_law)][[2](https://www.nngroup.com/articles/minimize-cognitive-load/)][[3](https://www.researchgate.net/publication/12189991_When_Choice_is_Demotivating_Can_One_Desire_Too_Much_of_a_Good_Thing)].

Happy to also discuss the limitations or benefits in a follow up issue or PR. Here's how this could look like:

editors

loujaybee:
> consolidate browser and desktop editors in this PR

Thanks for the clarification, @gtsiolis — I think we're in agreement on what you mentioned. I agree it doesn't make sense to prevent this PR progressing. I think we don't yet have an issue for what you mentioned, yet. So I'll make one now. 

In this case, I'm think I'm talking about something related, but different. EAP and insiders are essentially the same concept, i.e the user doesn't get a new IDE, but instead gets a pre-release/early version, so I think should be represented in the same way in the UI for both the desktop and browser. Merging the PR will bring a (non-critical) inconsistency. Happy to move this to another issue, but wanted to raise so that we were aware of the inconsistency before merging. 

Visual example:

| EAP selected using checkbox      | Insiders selected using tile |
| ----------- | ----------- |
| Editors      | Editors |

The screenshot you attached would tackle both:
1. Inconsistent selection of early access/insider IDE's (what I'm raising here)
2. Simplify the desktop / browser configurations (the additional simplification proposed)

I'll make an issue with the changes you suggested @gtsiolis, then let's follow up as soon as we can after this PR with those changes to simplify/align the browser/desktop preferences.

loujaybee:
Ah, I've just seen that we have: https://github.com/gitpod-io/gitpod/issues/7932, let's proceed with this as is, that issue will address the inconsistency mentioned 🙏

gtsiolis:
@loujaybee:

ok

components/gitpod-protocol/src/protocol.ts
akosyakov: After recent discussions about unifying preferences to have a check box fro all IDES, it is indeed makes sense to me.

components/server/src/workspace/workspace-starter.ts
```diff
@@ -348,7 +348,8 @@ export class WorkspaceStarter {
                 if (!!desktopIdeChoice) {
                     const mappedImage = ideConfig.ideOptions.options[desktopIdeChoice];
                     if (!!mappedImage && mappedImage.image) {
-                        configuration.desktopIdeImage = mappedImage.image;
+                        const useLatest = !!user.additionalData?.ideSettings?.useLatestVersion;
```
akosyakov, line 351: Should not `resolveReferrerIDE` also respect `useLatestVersion`?

akosyakov (commented):
UI wise it looks good, but new flag should be respected for `#referrer:jetbrains-gateway` prefix too.

akosyakov:
@iQQBot do you know why latest image still use CWM links? We never were able to rebuild latest images since Gateway plugin PR was merged?

akosyakov:
> this is the latest image is not "latest", it build failed since Gateway plugin PR was merged

Could you make sure that they are good? Together with @felladrin 

I don't want to deploy something what brings old functionality.

felladrin (approved):
/approve

Tested on all platforms, both stable and EAP.

Screenshot 2022-02-07 at 12 39 11
Screenshot 2022-02-07 at 12 30 34
Screenshot 2022-02-07 at 12 38 01
Screenshot 2022-02-07 at 12 48 29
Screenshot 2022-02-07 at 12 53 03
Screenshot 2022-02-07 at 13 15 28
Screenshot 2022-02-07 at 13 18 37
Screenshot 2022-02-07 at 13 22 53
Screenshot 2022-02-07 at 13 26 38

gtsiolis (commented):
@iQQBot UX changes look incrementally good, since this will add the checkbox to use the latest EAP version without affecting anything else. Cc @akosyakov @loujaybee 

@gitpod-io/engineering-webapp Could someone take a closer look at the code changes?

🍊 🍊 🍊 🍊 

In the next iteration, we could probably consider improving :a: this settings section or :b: the workspace start page.

For :a: we could, 
1. Merge the IDEs into a single section
2. Update card selector component and sidebar design [#7932](https://github.com/gitpod-io/gitpod/issues/7932)
2. Use the same modality for the latest VS Code option and drop the extra Insiders option
4. Update the _Latest Release_ checkbox control so that this applies to all options
5. Surface release version for each editor or IDE
6. Introducing a checkbox control to set the preference for the desktop version, including the (un)pairing use cases like when a user selects an editor that is only available in desktop, etc.
7. Mark VS Code as the default option so that we can fallback to it when users uncheck the desktop preference

For :b: we could:
1. Improve more actions button on the workspace start page when opening in desktop IDEs [#6910](https://github.com/gitpod-io/gitpod/issues/6910)

akosyakov:
@gtsiolis It would be cool to have issues for your ideas if we don't have it. The team could pull it one by one.

components/server/src/ide-config.ts
```diff
@@ -190,6 +191,20 @@ export class IDEConfigService {
...
+                try {
+                    value.ideOptions.options[id].latestImage = await this.resolveImageDigest(option.latestImage!);
```
jankeromnes, line 196: Nit: I would find it a bit "cleaner" to use a boolean like `option.useLatestImage`, and if it's true, set `option.latestImage` (just like `option.resolveImageDigest` and `option.image` above).

Here, we seem to use `option.latestImage` both as a "boolean" (i.e. `filter(([_, x]) => x.latestImage))` above) _and_ a resolved value `.latestImage = await ...`. This looks a little bit dangerous (e.g. we might accidentally resolve the same image multiple times, or maybe forget to resolve it elsewhere, etc).

However, not blocking the review due to this. Please feel free to address in a follow-up PR.

jankeromnes (approved):
Many thanks!

I've looked at the code, and found one small things, but otherwise this looks to me. I've also briefly tested as instructed and it seemed to work fine. 👍

Rubber-stamping, since this has already been approved a few times. ✅

jankeromnes:
Agree with @gtsiolis that the IDE settings are becoming a little confusing:

Screenshot 2022-02-08 at 09 21 31

- It says on top that all desktop IDEs are Beta, and near the bottom that only the JetBrains desktop IDEs are Beta (which is it? 👀)
- `[ ] Latest Release` seems to apply to all IDE options, but reading the description it seems to be only relevant for JetBrains. As a user, I wonder what happens if I check "Latest Release" but select "None" or "VS Code" as desktop IDE 🤔

But filing issues and iterating on this sounds like a good path forward. 😊

components/server/src/ide-config.ts
iQQBot: latestImage is a string not boolean, so `filter(([_, x]) => x.latestImage)` is equal than `filter(([_, x]) => x.latestImage != null && x.latestImage !== '')`
it no problem I think

jankeromnes: Thanks @iQQBot! Yes, I understand this, but my point was, I prefer something like this:

```ts
type Option = {
  resolveLatestImage: boolean;
  latestImage?: string;
}

if (option.resolveLatestImage) {
  option.latestImage = await resolve();
}
```

where the controlling condition is different from what's actually resolved.

Currently, the code seems to work like this:

```ts
type Option = {
  latestImage?: string;
}

if (option.latestImage) {
  option.latestImage = await resolve();
}
```

This makes it a bit unclear if an image has already been resolved or not, e.g.:
- If the code is called multiple times, you might accidentally resolve this image multiple times
- You cannot do `if (!option.latestImage) { option.latestImage = await resolve(); }`
- Someone might accidentally "forget" to resolve the image in some cases (i.e. leave it as `option.latestImage === "latest"`, because `if (option.latestImage)` looks "true")

Issue #7700, opened 2022-01-19
ARM64 Support


RedstoneWizard08:
## Is your feature request related to a problem? Please describe

I would like to have GitPod Self-Hosted support ARM64 machines, as I was trying to install it on a cloud-hosted VM and it won't run, due to exec format errors.

## Describe the behaviour you'd like

GitPod will run as a kubernetes application on ARM64 machines.

## Describe alternatives you've considered

None, as I have not found anything else that does what GitPod does **and** supports ARM64 machines for self-hosting.

wolf-yuan-6115:
> ## Is your feature request related to a problem? Please describe
> 
> I would like to have GitPod Self-Hosted support ARM64 machines, as I was trying to install it on a cloud-hosted VM and it won't run, due to exec format errors.
> 
> ## Describe the behaviour you'd like
> 
> GitPod will run as a kubernetes application on ARM64 machines.
> 
> ## Describe alternatives you've considered
> 
> None, as I have not found anything else that does what GitPod does **and** supports ARM64 machines for self-hosting.
> 
> ## Additional context
> 
> 

Cloud server like **Oracle Cloud** is arm64, and **Raspberry Pi** is arm based too.

I would like to hosting gitpod on my own Raspberry Pi if gitpod support it.

RedstoneWizard08:
Haha I'm on Oracle Cloud too

ZCube:
Anyone interested in solving this issue?

I think this issue will be resolved by doing the following tasks.

0. Prepare an automated build environment that can build on ARM64.
1. Modify the Dockerfile of this URL to be buildable. https://github.com/gitpod-io/workspace-images/tree/master/base
2. Modify the Dockerfile of this URL to be buildable. https://github.com/gitpod-io/workspace-images/tree/master/full
3. Modify the Dockerfile of this URL to be buildable. https://github.com/gitpod-io/gitpod/tree/main/dev/image
4. Create a distribution package for this URL. https://github.com/grpc/grpc-node/tree/master/packages/grpc-tools
5. Replace the bitnami related chart in this URL with another one. https://github.com/gitpod-io/gitpod/tree/main/installer/third_party/charts/minio
6. Replace the bitnami related chart in this URL with another one. https://github.com/gitpod-io/gitpod/tree/main/installer/third_party/charts/mysql
7. Replace the bitnami related chart in this URL with another one. https://github.com/gitpod-io/gitpod/tree/main/installer/third_party/charts/rabbitmq
[...]

PR #8053, opened 2022-02-06
[installer] Support MariaDB

ZCube:
## Description
Add MariaDB support to the installer.

## How to test
Render using below config
```
...
database:
  inCluster: true
  mariaDB: true
...
```

## Release Notes
```release-note
NONE
```

### DIFF
--- a/installer/pkg/components/database/incluster/helm.go
+++ b/installer/pkg/components/database/incluster/helm.go
@@ -12,7 +12,7 @@ import (
 	"helm.sh/helm/v3/pkg/cli/values"
 )
 
-var Helm = common.CompositeHelmFunc(
+var HelmMySQL = common.CompositeHelmFunc(
 	helm.ImportTemplate(charts.MySQL(), helm.TemplateConfig{}, func(cfg *common.RenderContext) (*common.HelmConfig, error) {
 		affinity, err := helm.AffinityYaml(cluster.AffinityLabelMeta)
 		if err != nil {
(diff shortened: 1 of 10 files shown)


roboquat:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please assign **metcalfc** after the PR has been reviewed.
You can assign the PR to them by writing `/assign @metcalfc` in a comment when ready.

<details open>

</details>

MrSimonEmms:
@ZCube thanks for this.

What problem does adding MariaDB support solve? I know MySQL and MariaDB are meant to be compatible, but there are some differences (from memory, I think MySQL supports more unquoted statements than MariaDB). My main concern is that we only officially support MySQL for the database and that this may introduce a divergence between self-hosted and SaaS which we're trying to avoid.

If you wanted to use MariaDB, my suggestion would be to use an external MariaDB - I'd suggest that we keep the in-cluster DB to be MySQL only.

ZCube:
> @ZCube thanks for this.
> 
> What problem does adding MariaDB support solve? I know MySQL and MariaDB are meant to be compatible, but there are some differences (from memory, I think MySQL supports more unquoted statements than MariaDB). My main concern is that we only officially support MySQL for the database and that this may introduce a divergence between self-hosted and SaaS which we're trying to avoid.
> 
> If you wanted to use MariaDB, my suggestion would be to use an external MariaDB - I'd suggest that we keep the in-cluster DB to be MySQL only.

okay. I would consider using the External configuration. For MySQL, only x86 is provided. Because of this, I had to use MariaDB to solve this ARM64 #7700 issue.

RedstoneWizard08:
@ZCube we could add the MySQL helm chart to the `bitnami-compat` repo using this image: [docker-library/mysql](https://github.com/docker-library/mysql) (That is the source Dockerfile for this image: [mysql](https://hub.docker.com/_/mysql))
However, we would also have to reproduce the image as it doesn't support arm64. This is easy to do.

corneliusludmann:
Thank you for your understanding. I would close the PR for now then, okay?

PR #8061, opened 2022-02-07
[workspacekit] make /etc/hosts modifiable

### LINKED ISSUE #7981: [jetbrains] many exceptions on spring petclinic example
### Bug description

Exceptions like:
```
java.io.IOException: The server sockets created using the LocalRMIServerSocketFactory only accept connections from clients running on the host where the RMI remote objects have been exported.
	at jdk.management.agent/sun.management.jmxremote.LocalRMIServerSocketFactory$1.accept(LocalRMIServerSocketFactory.java:114) ~[na:na]
	at java.rmi/sun.rmi.transport.tcp.TCPTransport$AcceptLoop.executeAcceptLoop(TCPTransport.java:394) ~[na:na]
	at java.rmi/sun.rmi.transport.tcp.TCPTransport$AcceptLoop.run(TCPTransport.java:366) ~[na:na]
	at java.base/java.lang.Thread.run(Thread.java:829) ~[na:na]
```

To reproduce start with petclinic example.

### Steps to reproduce

See above.

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

mustard-mh:
## Description

Recently we introduced workspace-wide network namespace #6409  for all workspace
This causes the network devices seen in the workspace to be independent of the workspace pod, and the hostname is expected to resolve to an accessible IP address, so it makes sense to point to 127.0.0.1

This does not cause other problems, instead the original value (workspace pod ip) causes some programs to not work properly because it not accessible

And make /ect/hosts editable make sense for users they can customize hosts in .gitpod.yaml or dotfiles

## Related Issue(s)

Fixes #7981

## How to test

this contain 2 test case
1. ensure /etc/hosts modifiable and check the hostname record is point to 127.0.0.1
    1.1 Open a workspace and edit /etc/hosts

2. ensure resolve #7981
    2.1 start [petclinic](https://github.com/gitpod-io/spring-petclinic) with IntelliJ
    2.2 change config flag to "-Dcom.sun.management.jmxremote.local.only=true"
    2.3 launch run configuration
    2.4 it should work without any jmx security errors

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/workspacekit/cmd/rings.go
+++ b/components/workspacekit/cmd/rings.go
@@ -340,12 +340,20 @@ var ring1Cmd = &cobra.Command{
 			}
 		}
 
-		// We deliberately do not bind mount `/etc/resolv.conf`, but instead place a copy
+		// We deliberately do not bind mount `/etc/resolv.conf` and `/etc/hosts`, but instead place a copy
 		// so that users in the workspace can modify the file.
-		err = copyResolvConf(ring2Root)
+
+		copyPaths := []string{"/etc/resolv.conf", "/etc/hosts"}
+		for _, fn := range copyPaths {
+			err = copyRing2Root(ring2Root, fn)
+			if err != nil {
+				log.WithError(err).Warn("cannot copy " + fn)
+			}
+		}
+
+		err = makeHostnameLocal(ring2Root)
 		if err != nil {
-			log.WithError(err).Error("cannot copy resolv.conf")
-			return
+			log.WithError(err).Warn("cannot make /etc/hosts hostname local")
 		}
 
 		env := make([]string, 0, len(os.Environ()))
@@ -565,11 +573,11 @@ var (
 		"/workspace",
 		"/sys",
 		"/dev",
-		"/etc/hosts",
 		"/etc/hostname",
 	}
 	rejectMountPaths = map[string]struct{}{
 		"/etc/resolv.conf": {},
+		"/etc/hosts":       {},
 	}
 )
 
@@ -636,9 +644,8 @@ func findBindMountCandidates(procMounts io.Reader, readlink func(path string) (d
 	return mounts, scanner.Err()
 }
 
-// copyResolvConf copies /etc/resolv.conf to <ring2root>/etc/resolv.conf
-func copyResolvConf(ring2root string) error {
-	fn := "/etc/resolv.conf"
+// copyRing2Root copies <fn> to <ring2root>/<fn>
+func copyRing2Root(ring2root string, fn string) error {
 	stat, err := os.Stat(fn)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


csweichel:
We already have this mechanism for `/etc/resolv.conf` - we should be able to reuse that mechanism

mustard-mh:
> We already have this mechanism for `/etc/resolv.conf` - we should be able to reuse that mechanism

we not only make it modifiable, also we need to make the file correct [see detail](https://gitpod.slack.com/archives/C02F19UUW6S/p1643886043639809)

the default IP address is workspace pod IP address, it not accessible in workspace, so we need to change it to 127.0.0.1

reuse means that we need to regexp find it and replace, does it make sense to reuse?

mustard-mh:
a new commit for reuse @csweichel

components/workspacekit/cmd/rings.go
```diff
@@ -340,18 +340,21 @@ var ring1Cmd = &cobra.Command{
...
-		if err != nil {
-			log.WithError(err).Error("cannot copy resolv.conf")
-			return
+
+		copyPaths := []string{"/etc/resolv.conf", "/etc/hosts"}
+		for _, fn := range copyPaths {
+			err = copyRing2Root(ring2Root, fn)
+			if err != nil {
+				log.WithError(err).Error("cannot copy " + fn)
+				return
```
csweichel, line 351: We'll still want to try and copy the other file(s) in this case.
```suggestion
				log.WithError(err).Warn("cannot copy " + fn)
```

components/workspacekit/cmd/rings.go
```diff
@@ -670,32 +672,32 @@ func copyResolvConf(ring2root string) error {
...
-fe00::0 ip6-localnet
-fe00::0 ip6-mcastprefix
-fe00::1 ip6-allnodes
-fe00::2 ip6-allrouters
-`
-	hostname, err := os.Hostname()
-	if err == nil {
-		hostsStr += fmt.Sprintf("127.0.0.1 %s\n", hostname)
+	path := filepath.Join(ring2root, "/etc/hosts")
+	stat, err := os.Stat(path)
```
csweichel, line 681: if the file does not exist, we would not necessarily want to stop the workspace

csweichel (commented):
Nice - thank you for incorporating this.

components/workspacekit/cmd/rings.go
mustard-mh: Yes, so this func only return error
https://github.com/gitpod-io/gitpod/blob/5b0af01cf950564d26e25aa525092cbfbe036300/components/workspacekit/cmd/rings.go#L353-L357

in here just log

akosyakov:
I really like that the issue gets resolved 😄 but can someone update the PR description to explain what this PR does and why it is not a workaround to list in hosts file but a proper solution.

csweichel:
I agree with Anton - the PR needs a better description :) 

That said, the change LGTM and works like a charm.
Once the description is updated, I'll happily approve.

mustard-mh:
Hi @csweichel, I already update the PR desc

csweichel (approved):
Thanks @mustard-mh

Issue #7866, opened 2022-01-27
Add number of users to the telemetry data


MrSimonEmms:
Currently, [the telemetry](https://github.com/gitpod-io/gitpod/blob/main/components/installation-telemetry/cmd/send.go#L53) only provides a randomly generated unique ID and the version number.

To this, we should add the number of registered users. The Segment data should be named something like `totalUsers` so we can expand this to include active/inactive users in the future

This will likely need some liaison from @gitpod-io/engineering-webapp to understand how best to get the total number of users from the database

JanKoehnlein:
This is what we do in the server: https://github.com/gitpod-io/gitpod/blob/fa62003417248d638df419b90e35847b037e56f9/components/server/ee/src/user/user-service.ts#L44-L45

PR #8067, opened 2022-02-07
telemetry: add more fields to data

Pothulapati:
## Description

This PR updates the `installation-admin-controller` to also retrieve
more data to send with telemetry. These are not part of the
`installationAdminDb` as we do not want to store this in the database
but lazily retrieve whenever a request is sent to `/data` endpoint
of the `installation-admin` express app unlike the `uuid` and settings
which need to be stored and updated.

The following fields are added:
- `totalUsers` : specifies the total number of users in the instance
- `totalWorkspaces`: specifies the total number of **regular** workspaces in the instance
- `totalInstances`: specifies the total number of **regular** workspace instances in the gitpod instance

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/7866

## How to test

Deploy via werft and do a `/data` `GET` on the `9000` port of `server`.
1. Run `yarn telepresence` in `/server`
2. Run `kubectl port-forward svc/server 9000`
3. Run `curl -H "Content-Type: application/json" http://localhost:9000/data`

Sample Output:

```
{"installationAdmin":{"id":"79f821ea-eee8-42cb-adff-1828b75656dd","settings":{"sendTelemetry":true}},"totalUsers":3,"totalWorkspaces":1,"totalInstances":93}
```

## Release Notes

```release-note
[server]: Add `totalUsers`, `totalWorkspaces`, and `totalInstances` fields to telemetry data
```

### DIFF
--- a/components/gitpod-db/src/typeorm/workspace-db-impl.ts
+++ b/components/gitpod-db/src/typeorm/workspace-db-impl.ts
@@ -158,7 +158,7 @@ export abstract class AbstractTypeORMWorkspaceDBImpl implements WorkspaceDB {
             .addOrderBy('GREATEST(ws.creationTime, wsi.creationTime, wsi.startedTime, wsi.stoppedTime)', 'DESC')
             .limit(options.limit || 10);
         if (options.searchString) {
-            qb.andWhere("ws.description LIKE :searchString", {searchString: `%${options.searchString}%`});
+            qb.andWhere("ws.description LIKE :searchString", { searchString: `%${options.searchString}%` });
         }
         if (!options.includeHeadless) {
             qb.andWhere("ws.type = 'regular'");
@@ -332,6 +332,15 @@ export abstract class AbstractTypeORMWorkspaceDBImpl implements WorkspaceDB {
         }
         const [rows, total] = await queryBuilder.getManyAndCount();
         return { total, rows };
+    }
+
+    public async getInstanceCount(type?: string): Promise<number> {
+        const workspaceInstanceRepo = await this.getWorkspaceInstanceRepo();
+        const queryBuilder = workspaceInstanceRepo.createQueryBuilder("wsi")
+            .leftJoinAndMapOne("wsi.workspace", DBWorkspace, "ws", "wsi.workspaceId = ws.id")
+            .where("ws.type = :type", { type: type ? type.toString() : "regular" }); // only regular workspaces by default
+
+        return await queryBuilder.getCount();
     }
 
     public async findRegularRunningInstances(userId?: string): Promise<WorkspaceInstance[]> {
@@ -578,7 +587,7 @@ export abstract class AbstractTypeORMWorkspaceDBImpl implements WorkspaceDB {
 
     public async findSnapshotsByWorkspaceId(workspaceId: string): Promise<Snapshot[]> {
         const snapshots = await this.getSnapshotRepo();
-        return snapshots.find({where: {originalWorkspaceId: workspaceId}});
+        return snapshots.find({ where: { originalWorkspaceId: workspaceId } });
     }
 
     public async storePrebuiltWorkspace(pws: PrebuiltWorkspace): Promise<PrebuiltWorkspace> {
@@ -596,7 +605,7 @@ export abstract class AbstractTypeORMWorkspaceDBImpl implements WorkspaceDB {
         }
         const repo = await this.getPrebuiltWorkspaceRepo();
         return await repo.createQueryBuilder('pws')
-            .where('pws.cloneURL = :cloneURL AND pws.commit LIKE :commit', { cloneURL, commit: commit+'%' })
+            .where('pws.cloneURL = :cloneURL AND pws.commit LIKE :commit', { cloneURL, commit: commit + '%' })
             .orderBy('pws.creationTime', 'DESC')
             .innerJoinAndMapOne('pws.workspace', DBWorkspace, 'ws', "pws.buildWorkspaceId = ws.id and ws.contentDeletedTime = ''")
             .getOne();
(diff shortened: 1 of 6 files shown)


components/gitpod-protocol/src/installation-admin-protocol.ts
```diff
@@ -23,6 +23,11 @@ export interface InstallationAdmin {
     settings: InstallationAdminSettings;
 }
 
+export interface Data {
+    installationAdmin: InstallationAdmin
+    totalUsers: number
```
Pothulapati, line 28: As we add more telemetry data, We could continue adding them here thereby separating it from the `InstallationAdmin` (which integrates into the database).

Pothulapati:
Still there is work on wiring these metrics into the `installation-telemetry` job! Will update once that is done!

components/gitpod-protocol/src/installation-admin-protocol.ts
Pothulapati: Same goes with the new `Data` struct for the golang side of things in `installation-telemetry` workspace

corneliusludmann (approved):
/lgtm
/approve

Looks good. :+1:

components/installation-telemetry/pkg/server/installationAdmin.go
```diff
@@ -17,12 +17,17 @@ type InstallationAdminSettings struct {
 	SendTelemetry bool `json:"sendTelemetry"`
 }
 
-type InstallationAdminData struct {
+type Data struct {
+	InstallationAdmin InstallationAdmin `json:"installationAdmin"`
```
corneliusludmann, line 21: Adding the new attribute `installationAdmin` also means that the data structure in Segment will change, correct? Then we would probably better copy ID and Settings to the Data struct before we send the data. Otherwise, it would break analytics with the changing data structure.

corneliusludmann (changes requested):

components/installation-telemetry/cmd/send.go
```diff
@@ -55,7 +55,9 @@ var sendCmd = &cobra.Command{
 			Event:  "Installation telemetry",
 			Properties: analytics.NewProperties().
 				Set("version", versionId).
-				Set("totalUsers", data.TotalUsers),
+				Set("totalUsers", data.TotalUsers).
+				Set("totalWorkspaces", data.TotalWorkspaces).
+				Set("totalInstnaces", data.TotalInstances),
```
corneliusludmann, line 60: Typo:
```suggestion
				Set("totalInstances", data.TotalInstances),
```

corneliusludmann (changes requested):

components/gitpod-db/src/typeorm/workspace-db-impl.ts
```diff
@@ -334,6 +334,15 @@ export abstract class AbstractTypeORMWorkspaceDBImpl implements WorkspaceDB {
         return { total, rows };
     }
 
+    public async getInstanceCount(type?: string): Promise<number> {
+        const workspaceInstanceRepo = await this.getWorkspaceInstanceRepo();
+        const queryBuilder = workspaceInstanceRepo.createQueryBuilder("wsi")
+            .leftJoinAndMapOne("wsi.workspace", DBWorkspace, "ws", "wsi.workspaceId = ws.id")
```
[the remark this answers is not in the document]
corneliusludmann: Ah, you need it for the workspace types. :bulb:

components/server/src/installation-admin/installation-admin-controller.ts
```diff
@@ -6,17 +6,25 @@
...
+    @inject(WorkspaceDB) protected readonly workspaceDb: WorkspaceDB
 
     public create(): express.Application {
         const app = express();
 
         app.get('/data', async (req: express.Request, res: express.Response) => {
-            const data = await this.installationAdminDb.getData();
+            const data: Data = {
+                installationAdmin: await this.installationAdminDb.getData(),
+                totalUsers: await this.userDb.getUserCount(false),
```
corneliusludmann, line 24: NIT: I personally would exclude the buildin user here. But that's maybe more a matter of taste. If you get a change request from the team web app review, you could consider changing this. Otherwise, I'm fine with just merging it as it.
```suggestion
                totalUsers: await this.userDb.getUserCount(true),
```

corneliusludmann (dismissed):
Looks good. Thanks for the good work! :+1:

components/server/src/installation-admin/installation-admin-controller.ts
laushinka: @corneliusludmann What's a built-in user? 🙏🏽 I tested it and despite only having one user in the preview env, I got `"totalUsers":2`.

corneliusludmann: We always add a user for “workspace probes”.

https://github.com/gitpod-io/gitpod/blob/240197f239faf9a5b1fb67c5260f3564637b5e52/components/gitpod-db/src/user-db.ts#L122

https://github.com/gitpod-io/gitpod/blob/7abe78c91792c18b0c22103c890b25ea3a4cdf05/components/gitpod-db/src/typeorm/migration/1592203031938-Baseline.ts#L84

Workspace probes are started here:

https://github.com/gitpod-io/gitpod/blob/1c0ee09cdc6a827c5e4258c1c8369eb8e6914c5f/components/server/ee/src/workspace/workspace-health-monitoring.ts#L36

This is a regular test, that the cluster is healthy and we can start workspaces.

laushinka: Thanks! In this case I would also exclude built-in users, as I believe it doesn't serve your purpose @Pothulapati. Once you have this change I'll give my thumbs up.

Pothulapati: Done! 👍🏼

Pothulapati: Interestingly, I still see the count to start from`2`. Is this because of the presence of `admin` by default? 🤔

Pothulapati: Just verified this! and It's actual users without admin! 2 of my accounts (I logged in for testing) and 1 of @laushinka! Checks out with the current count of `3` without any admin/built-in users.

This is fixed.

corneliusludmann (approved):

corneliusludmann:
@Pothulapati Since we rebase and merge, it would be good when you can clean up your commits and squash them into one commit. :heart:

Pothulapati:
@corneliusludmann Done!

corneliusludmann (approved):
Awesome! Thanks!

components/server/src/installation-admin/installation-admin-controller.ts
laushinka: Yess, you can check the actual users here https://tarun-telemetry-users.staging.gitpod-dev.com/admin/users 🤗

laushinka (approved):

Issue #8043, opened 2022-02-04
[dashboard, server] Prevent redirect loops that trigger "startWorkspace" in a loop


geropl:
Ways to solve this:
 - 1. level:
   - fix the broken assumption in the dashboard (error handling)
 - 2. level:
   - have preventive measures in place in `dashboard` (e.g., exp. backoff for `startWorkspace`)
   - have preventive measures in place in `server`: rate-limit `startWorkspace` more strictly

Proof that this issue enforced [this incident](https://app.incident.io/incidents/104): https://ui.honeycomb.io/gitpod/datasets/gitpod-production/result/xnbGHzzXiP7?tab=traces

sagor999:
Just want to add that we should try to fix it ASAP as it makes us ddos ourselves.

csweichel:
> Just want to add that we should try to fix it ASAP as it makes us ddos ourselves.

💯 agreed.

@jldec can you make sure this gets high up on the priority list?

jankeromnes:
FYI, a few links to more context about this problem, in case anyone finds this useful:

- https://github.com/gitpod-io/gitpod/issues/5401
- https://github.com/gitpod-io/gitpod/issues/6688
- [Slack thread](https://gitpod.slack.com/archives/C02EN94AEPL/p1637312973013200?thread_ts=1637310880.010600&cid=C02EN94AEPL) (internal)

geropl:
Here's the `startWorkspace` traces from Friday: https://ui.honeycomb.io/gitpod/datasets/gitpod-production/result/8xyAV6hi4Jj
It shows that for a single user the same workspace is always re-starting the same workspace.

geropl:
Reproducable in devstaging by:
 - copying the relevant `ws-daemon` label from the node, e.g. `gitpod.io/ws-daemon_ready_ns_staging-gpl-8043-redirect-loop: "true"`
 - scaling down `ws-daemon` (e.g. by requiring a non-existing node label)
 - manually re-adding said `ws-daemon_ready...` node label :+1:

geropl:
The plan is to:
 - [ ] enable basic rate-limiting (per `server` instance) and handle re-tries in `dashboard`
 - [ ] fix the underlying communication issue between the page served by `ws-proxy` / workspace cluster and the dashboard / app cluster by adding an parameter (`not found`) that helps the dashboard disting
[...]

PR #8073, opened 2022-02-07
[server, dashboard] Do basic rate limiting on startWorkspace

geropl:
## Description
This configures a rate-limit for API calls to `startWorkspace` of `1` / `10s`. The dashboard handles the error and waits until `retryAfter` to re-try the call (currently indefinitely).

Note that this is enforced per `server` instance at the moment, as the rate-limiter state is not shared between server instances. But it's a first step that already reduces the impact of future incidents.

## Related Issue(s)

Context: #8043

## How to test
 - start any workspace: https://gpl-8043-redirect-loop.staging.gitpod-dev.com/#https://github.com/gitpod-io/template-typescript-node
 - open the dev console to notice the rate limier messages

## Release Notes

```release-note
configure basic rate-limiting for `startWorkspace`
```


components/server/src/auth/rate-limiter.ts
```diff
@@ -34,7 +35,11 @@ function getConfig(config: RateLimiterConfig): RateLimiterConfig {
         default: {
             points: 60000, // 1,000 calls per user per second
             durationsSec: 60,
-        }
+        },
+        startWorkspace: {
+            points: 1,  // 1 workspace start per user per 10s
+            durationsSec: 10
+        },
```
jankeromnes, line 42: 👍

components/server/src/websocket/websocket-connection-manager.ts
```diff
@@ -356,7 +356,7 @@ class GitpodJsonRpcProxyFactory<T extends object> extends JsonRpcProxyFactory<T>
                     throw rlRejected;
                 }
                 log.warn({ userId }, "Rate limiter prevents accessing method due to too many requests.", rlRejected, { method });
-                throw new ResponseError(ErrorCodes.TOO_MANY_REQUESTS, "too many requests", { "Retry-After": String(Math.round(rlRejected.msBeforeNext / 1000)) || 1 });
+                throw new ResponseError<RateLimiterError>(ErrorCodes.TOO_MANY_REQUESTS, "too many requests", { method, retryAfter: Math.round(rlRejected.msBeforeNext / 1000) || 1 });
```
jankeromnes, line 359: Probably safer with a `Math.ceil`, but not a major issue (since the front-end retry code works well):

```suggestion
                throw new ResponseError<RateLimiterError>(ErrorCodes.TOO_MANY_REQUESTS, "too many requests", { method, retryAfter: Math.ceil(rlRejected.msBeforeNext / 1000) || 1 });
```

jankeromnes (approved):
Wow, so cool to see the rate limiter code actually working! 👀 (Last time we tried it with Team Plan methods, I think it didn't behave as expected as all and we had to revert it again.)

I guess this makes our "infinite-redirect on workspace fast crash" failure mode a little better (i.e. we make the self-DDOS 10x less severe), and the code looks good to me. 👍

I'm just curious if this can have any impact in normal situations.
- For one, I guess it will impact the use case "have 5 stopped workspaces and restart them all by clicking all the tabs" (I guess it will just make their restart `[0-4] * 10s` slower, but maybe that's okay)
- Other impact can probably be tested on staging once merged. I expect no blockers, but if starting workspaces on staging becomes unreliable, we might have to revert again before deploying (that's what happened with our Team Plan calls rate-limiting attempt last time)

In any case, this looks good for merging & further testing on staging! 🚀

geropl:
@jankeromnes Thx for the thorough review! :pray: 

> I'm just curious if this can have any impact in normal situations.

Indeed. This is first and foremost a safety measure to help with the next days. Next on the list:
 - see if we can get the "number of workspaces started by the user recently" from the DB, and maybe show a modal if it's too excessive
 - fix the underlying issue: requires a re-deploy of workspace cluster (`ws-proxy`)

PR #8074, opened 2022-02-07
[werft] add reuse last port and include nodePort for find free port

iQQBot:
## Description

[werft] add reuse last port and include nodePort for find free port
I think this is root cause core-dev start workspace loop forever, and stuck in Allocating Resources
I notice some services nodePort and registry-facad's hostPost are same
start workspace, then pull image failed ....

this PR also add a function to reuse the last port, in order to save time from build and avoid find free port every time

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.werft/build.ts
+++ b/.werft/build.ts
@@ -3,7 +3,7 @@ import * as fs from 'fs';
 import * as path from 'path';
 import { exec, ExecOptions } from './util/shell';
 import { Werft } from './util/werft';
-import { waitForDeploymentToSucceed, wipeAndRecreateNamespace, setKubectlContextNamespace, deleteNonNamespaceObjects, findFreeHostPorts, createNamespace, helmInstallName } from './util/kubectl';
+import { waitForDeploymentToSucceed, wipeAndRecreateNamespace, setKubectlContextNamespace, deleteNonNamespaceObjects, findFreeHostPorts, createNamespace, helmInstallName, findLastHostPort } from './util/kubectl';
 import { issueCertficate, installCertficate, IssueCertificateParams, InstallCertificateParams } from './util/certs';
 import { reportBuildFailureInSlack } from './util/slack';
 import * as semver from 'semver';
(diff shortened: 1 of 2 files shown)


.werft/build.ts
```diff
@@ -427,14 +427,19 @@ export async function deployToDevWithInstaller(deploymentConfig: DeploymentConfi
...
     // find free ports
-    werft.log(installerSlices.FIND_FREE_HOST_PORTS, "Check for some free ports.");
-    const [wsdaemonPortMeta, registryNodePortMeta, nodeExporterPort] = findFreeHostPorts([
-        { start: 10000, end: 11000 },
-        { start: 30000, end: 31000 },
-        { start: 31001, end: 32000 },
-    ], metaEnv({ slice: installerSlices.FIND_FREE_HOST_PORTS, silent: true }));
+    werft.log(installerSlices.FIND_FREE_HOST_PORTS, "Find last ports");
+    let wsdaemonPortMeta = findLastHostPort(namespace, 'ws-daemon', metaEnv({ slice: installerSlices.FIND_FREE_HOST_PORTS, silent: true }))
+    let registryNodePortMeta = findLastHostPort(namespace, 'registry-facade', metaEnv({ slice: installerSlices.FIND_FREE_HOST_PORTS, silent: true }))
```
ArthurSens, line 432: could you include node exporter here as well?

ArthurSens (commented):
Nice find! I believe this will also help with making the deployment faster if we don't need to rollout the deamonsets in a new hostPort :)

Just made one single comment, could you add node exporter as well? I believe there is no need to use a new port for node-exporter if it has been deployed already

.werft/build.ts
iQQBot: installer not contain node-exporter

ArthurSens: Yeah I believe you correct :)

ArthurSens (approved):

PR #8075, opened 2022-02-07
[supervisor] move reaper to another process

### LINKED ISSUE #8056: too many "wait: no child processes" in gcp log
### Bug description

Too many "wait: no child processes" in gcp log

### Steps to reproduce

https://cloudlogging.app.goo.gl/zFFZgtR8VgCPqKfr9

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

iQQBot:
## Description

Note: this is part 1, This PR only contain `supervisor init` command

This PR move reaper to `supervisor init` in order to avoid `no child process` error

This reaper not only reap zombie processes (where the parent process is not pid 1 and the parent process is dead), but also reap processes (this part will result in `no child process`)
the root cause is reaper is running in a goroutine, it will cause resource competition with `execCommand.Wait()` or other wait function

There are many places in our code where the error is ignored by determining whether the error is `no child process`, but this can mislead our judgment, and we have no way of knowing whether the process exited unexpectedly or normally

There are other alternative solutions
1. start an additional small init process (like tini) when the supervisor starts a new process and set PR_GET_CHILD_SUBREAPER so that this init process is responsible for reap its child processes
👍 No changes to the supervisor structure
👎 The number of processes will become larger and the process tree will be ugly

2. give a random delay when the supervisor receives a SIGCHLD, hoping to avoid resource competition
👍  minimal changes
👎  this is not a cure, but just a reduction of probability

3. scan all /proc/[pid]/stat when a SIGCHLD is received, and recycle only when a zombie process is identified
👍 No change to the supervisor structure
👎 slower recovery efficiency

## Related Issue(s)

Relate #8056 

## How to test

This part don't need test, please use this PR for test #8114 and #8115
~~1. start a workspace~~
~~2. watch the log~~
~~3. stop a workspace~~
~~4. see there are no `no child process` any more~~
~~5. check the stop log, the behavior should be same with before~~

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/supervisor/cmd/init.go
+++ b/components/supervisor/cmd/init.go
@@ -1,0 +1,109 @@
+// Copyright (c) 2020 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package cmd
+
+import (
+	"os"
+	"os/exec"
+	"os/signal"
+	"strings"
+	"sync"
+	"syscall"
+
+	"github.com/gitpod-io/gitpod/common-go/log"
+	"github.com/spf13/cobra"
+	"golang.org/x/sys/unix"
+)
+
+var initCmd = &cobra.Command{
+	Use:   "init",
+	Short: "init the supervisor",
+	Run: func(cmd *cobra.Command, args []string) {
+		log.Init(ServiceName, Version, true, false)
+		// Because we're reaping with PID -1, we'll catch the child process for
+		// which we've missed the notification anyways.
+		sigs := make(chan os.Signal, 1)
+		signal.Notify(sigs, syscall.SIGCHLD)
+
+		// The reaper can be turned into a terminating reaper by writing true to this channel.
+		// When in terminating mode, the reaper will send SIGTERM to each child that gets reparented
+		// to us and is still running. We use this mechanism to send SIGTERM to a shell child processes
+		// that get reparented once their parent shell terminates during shutdown.
+		var terminating bool
+
+		reaper := func() {
+			for range sigs {
+				for {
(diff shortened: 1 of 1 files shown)


csweichel:
if we go this way, need to
1. update agent-smith to recognise the new process structure (see [here](https://github.com/gitpod-io/gitpod/blob/efa4a608d91a717b6a11f318c7d919971e718438/components/ee/agent-smith/pkg/detector/proc.go#L428))
2. beware of the order in which we deploy. Once this merged, if the workspace team deploys before the supervisor change is rolled out by the IDE team, workspaces would fail to start.

csweichel:
> > if we go this way, need to
> > 
> > 1. update agent-smith to recognise the new process structure (see [here](https://github.com/gitpod-io/gitpod/blob/efa4a608d91a717b6a11f318c7d919971e718438/components/ee/agent-smith/pkg/detector/proc.go#L428))
> 
> will add
> 
> > 2. beware of the order in which we deploy. Once this merged, if the workspace team deploys before the supervisor change is rolled out by the IDE team, workspaces would fail to start.
> 
> This is really a complex issue, and to be precise they should be deployed at the same time, because if the supervisor is deployed first, then the reaper functionality will be lost, maybe we should split this operation into multiple PRs

I reckon that's the way to go, because it means we don't incur complexity in the code to make this transition. I.e. have one PR which introduces `supervisor init`, and another which makes the change to workspacekit and agent-smith, which we'd deploy once supervisor is updated.

csweichel (approved):
LGTM

csweichel:
still needs IDE approval

akosyakov (approved):
testes downstream PRs look good

PR #8048, opened 2022-02-05
[code] Build code dependencies separately

jeanp413:
## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

PR #8077, opened 2022-02-07
Update VS Code References (settings sync schema, leeway)

filiptronicek:
## Description
This PR does two things:
1. It updates the Settings Sync types and schema, including the pointer to VS Code's code base
	a. Specifically, user tasks can be synced with this PR 
2. It migrates the old name of our VS Code repo -> `vscode` to the new one -> `openvscode-server`

## Related Issue(s)

Follow-up to #8048 

## How to test
Make sure user tasks are synced correctly. In order to do that, do the following:
1. Open a workspace in the preview environment, for instance [the Node.js example](https://ft-update-vscode-refs.staging.gitpod-dev.com/#https://github.com/gitpod-io/template-typescript-node).
2. Open the command pallete (<kbd>F1</kbd>) and execute <kbd>Tasks: Open User Tasks</kbd>
3. A JSON file will open in your editor with a default config, we need to alter this config, here is an example one:
	  ```json
	  {
	      "version": "2.0.0",
	      "tasks": [
	          {
	              "label": "Test sync",
	              "type": "shell",
	              "command": "echo 'Hewo Gitpod (ᵘʷᵘ)'",
	          }
	      ]
	  }
	  ```
4. Open another workspace and repeat step two; the task should be synced in your new workspace and executable via <kbd>Tasks: Run Task</kbd> (your task will be listed first). Then confirm by selecting <kbd>Continue without scanning the task output</kbd>.

## Release Notes
```release-note
NONE
```

### DIFF
--- a/components/gitpod-db/src/typeorm/entity/db-code-sync-resource.ts
+++ b/components/gitpod-db/src/typeorm/entity/db-code-sync-resource.ts
@@ -7,7 +7,7 @@
 import { Entity, Column, PrimaryColumn, Index } from "typeorm";
 import { TypeORM } from "../typeorm";
 
-// should be aligned with https://github.com/gitpod-io/vscode/blob/75c71b49cc25554adc408e63b876b76dcc984bc1/src/vs/platform/userDataSync/common/userDataSync.ts#L113-L156
+// should be aligned with https://github.com/gitpod-io/openvscode-server/blob/a9286bef87ed21bbf108371aa1f62d9a5bc48fc4/src/vs/platform/userDataSync/common/userDataSync.ts#L110-L160
 export interface IUserData {
     ref: string;
     content: string | null;
@@ -17,14 +17,20 @@ export const enum SyncResource {
     Settings = 'settings',
     Keybindings = 'keybindings',
     Snippets = 'snippets',
+    Tasks = 'tasks',
     Extensions = 'extensions',
-    GlobalState = 'globalState'
+    GlobalState = 'globalState',
 }
-export const ALL_SYNC_RESOURCES: SyncResource[] = [SyncResource.Settings, SyncResource.Keybindings, SyncResource.Snippets, SyncResource.Extensions, SyncResource.GlobalState];
+export const ALL_SYNC_RESOURCES: SyncResource[] = [SyncResource.Settings, SyncResource.Keybindings, SyncResource.Snippets, SyncResource.Tasks, SyncResource.Extensions, SyncResource.GlobalState];
 
 export interface IUserDataManifest {
-    latest: Record<ServerResource, string>
-    session: string;
+	readonly latest?: Record<ServerResource, string>;
+	readonly session: string;
+    /**
+     * This property reflects a weak ETag for caching code sync responses,
+     * in the server, this is send in the Etag header and it's calculated by Express.js or we can override it manually.
+     */
+	//readonly ref: string;
 }
 
 export type ServerResource = SyncResource | 'machines';
--- a/components/server/src/code-sync/code-sync-service.ts
+++ b/components/server/src/code-sync/code-sync-service.ts
@@ -114,10 +114,10 @@ export class CodeSyncService {
                 return;
             }
             const manifest = await this.db.getManifest(req.user.id);
-            if (!manifest.latest.extensions) {
+            if (manifest.latest && !manifest.latest.extensions) {
                 manifest.latest.extensions = fromTheiaRev;
             }
-            if (!manifest.latest.settings) {
+            if (manifest.latest && !manifest.latest.settings) {
                 manifest.latest.settings = fromTheiaRev;
             }
             res.json(manifest);
--- a/components/gitpod-db/src/typeorm/code-sync-resource-db.spec.db.ts
+++ b/components/gitpod-db/src/typeorm/code-sync-resource-db.spec.db.ts
@@ -85,7 +85,7 @@ export class CodeSyncResourceDBSpec {
         let manifest = await this.db.getManifest(this.userId);
         expect(manifest).to.deep.eq(<IUserDataManifest>{
             session: this.userId,
-            latest: {}
+            latest: {},
         });
 
         let machinesRev = await this.db.insert(this.userId, 'machines', async () => { });
@@ -94,7 +94,7 @@ export class CodeSyncResourceDBSpec {
             session: this.userId,
             latest: {
                 machines: machinesRev
-            }
+            },
         });
 
         let extensionsRev = await this.db.insert(this.userId, SyncResource.Extensions, async () => { });
@@ -104,7 +104,7 @@ export class CodeSyncResourceDBSpec {
             latest: {
                 machines: machinesRev,
                 extensions: extensionsRev
-            }
+            },
         });
 
         machinesRev = await this.db.insert(this.userId, 'machines', async () => { });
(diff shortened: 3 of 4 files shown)


akosyakov:
I guess how to test section should explain how to define a user task which can be shared and how to verify that it is shared?

akosyakov:
@filiptronicek should it be opened for review. why is it still a draft?

components/gitpod-db/src/typeorm/entity/db-code-sync-resource.ts
```diff
@@ -17,14 +17,16 @@ export const enum SyncResource {
...
 }
-export const ALL_SYNC_RESOURCES: SyncResource[] = [SyncResource.Settings, SyncResource.Keybindings, SyncResource.Snippets, SyncResource.Extensions, SyncResource.GlobalState];
+export const ALL_SYNC_RESOURCES: SyncResource[] = [SyncResource.Settings, SyncResource.Keybindings, SyncResource.Snippets, SyncResource.Tasks, SyncResource.Extensions, SyncResource.GlobalState];
 
 export interface IUserDataManifest {
-    latest: Record<ServerResource, string>
-    session: string;
+	readonly latest?: Record<ServerResource, string>;
+	readonly session: string;
+	readonly ref: string;
```
filiptronicek, line 29: Exactly as @iQQBot said, I need to somehow get a `ref` and add it to here when syncing, haven't looked into it yet.

filiptronicek, line 29: Exactly as @iQQBot said, I need to somehow get a `ref` and add it to here when syncing, haven't looked into it yet.

components/server/src/code-sync/code-sync-service.ts
```diff
@@ -41,18 +40,6 @@ const objectPrefix = 'code-sync/';
 function toObjectName(resource: ServerResource, rev: string): string {
     return objectPrefix + resource + '/' + rev;
 }
-
-const fromTheiaRev = 'from-theia';
```
akosyakov, line 45: Could you revert it please? Theia is gone, but database records are not.

filiptronicek:
Today I dug through what the `ref` property stands for. 

This is an example object of the `manifest` during a settings sync:
```json
{
    "session": "401d5491-c3bd-4605-9175-ce33aaf8c9d5",
    "latest": {
        "extensions": "371d597e-e82e-41ed-ac89-6f497bca8872",
        "globalState": "43754c45-e198-4783-8a13-cc87d670c44c",
        "machines": "d14884d9-0f6d-48a6-90b9-664db8e9a240",
        "settings": "34459307-8332-4b50-a04e-756557358f0d"
    },
    "ref": "W/\"10a-egzfJAhstbSergamP5WP7tpzSsI\""
}
```

Aaaah, found its origin, it's an ETag! 
This is where it is generated: https://github.com/gitpod-io/openvscode-server/blob/d0b3ff8c29a181aa62d09461df5f755f479f6d2f/src/vs/server/node/webClientServer.ts#L54

jeanp413:
Good find, here's where [it's used](https://github.com/gitpod-io/openvscode-server/blob/8d283fd35f3dd74ba03cf89b5f80d319173991cc/src/vs/platform/userDataSync/common/userDataSyncStoreService.ts#L335-L353)
But the one who generates it is the settings sync server, looking at the code [here](https://github.com/gitpod-io/gitpod/blob/ffe789c123b58496b3f3a253bf1058ef53e98c8e/components/server/src/code-sync/code-sync-service.ts#L111-L124) it seems it gets generated [automatically by express](https://stackoverflow.com/a/67929691) but it's a weak etag, we need to change it to be strong and check if it's the same just respond with 304 (maybe express does this too automatically)

filiptronicek:
@jeanp413 it looks like we already indeed serve weak etags with the requests, thanks for noting the fact we don't have to have it on the server, I will remove it from there.
image

components/server/src/code-sync/code-sync-service.ts
```diff
@@ -4,303 +4,303 @@
...
+                 res.sendStatus(204);
+                 return;
+             }
+             const manifest = await this.db.getManifest(req.user.id);
+             if (manifest?.latest && !manifest.latest.extensions) {
+                 manifest.latest.extensions = fromTheiaRev;
+             }
+             if (manifest?.latest && !manifest.latest.settings) {
+                 manifest.latest.settings = fromTheiaRev;
+             }
```
jeanp413, line 122: Maybe this can be removed safely? no need to keep returning new entries with `fromTheiaRev`

filiptronicek: I think you are right, no reason to keep it since it is never going to be triggered anyways. Added it.

components/gitpod-db/src/typeorm/entity/db-code-sync-resource.ts
```diff
@@ -17,14 +17,20 @@ export const enum SyncResource {
...
+export const ALL_SYNC_RESOURCES: SyncResource[] = [SyncResource.Settings, SyncResource.Keybindings, SyncResource.Snippets, SyncResource.Tasks, SyncResource.Extensions, SyncResource.GlobalState];
 
 export interface IUserDataManifest {
-    latest: Record<ServerResource, string>
-    session: string;
+	readonly latest?: Record<ServerResource, string>;
+	readonly session: string;
+    /**
+     * This property reflects a weak ETag for caching code sync responses,
+     * but this is not needed for the server, because it is the one which calculates it.
```
jeanp413, line 31:
```suggestion
     * this is not needed for the server as it's calculated by Express.js automatically.
```

filiptronicek: @jeanp413 I like the change but would make it not say Expres but rather "our backend" or "the server", because as I mentioned [in a previous comment](https://github.com/gitpod-io/gitpod/pull/8077#issuecomment-1047243114) we calculate it instead of Expres itself.

jeanp413: Ah I missed that comment, I also forgot to mention that it's possible to override it, so in the case of `'/v1/resource/:resource/:ref'` endpoint we calculate it and in the case of `'/v1/manifest'` endpoint express does it

jeanp413 (dismissed):
LGTM

components/server/src/code-sync/code-sync-service.ts
```diff
@@ -114,12 +114,7 @@ export class CodeSyncService {
                 return;
             }
             const manifest = await this.db.getManifest(req.user.id);
-            if (!manifest.latest.extensions) {
```
akosyakov, line 117: why is it removed?

filiptronicek: Thanks for noting this, @akosyakov, I haven't really thought about the fact it could still be used. I will re-add it and we can perhaps discuss it another time, it is indeed out-of-scope for this PR.

jeanp413: > If you have a suer who used Theia and has user settines/extensions which are portable to VS Code and he start with VS Code now then we lose all his data?

@akosyakov  Is it possible to add a check so that we are sure user don't have theia data, so we can avoid returning new entries with `fromTheiaRev`

akosyakov: I think we should move this discussion to own issue and discuss it there.

It is not possible to add check here, you will need to run query on database to see how many users with Theia data which were not overridden by VS Code data.

filiptronicek: Reverted the changes ✅

jeanp413 (approved):

geropl (approved):
LGTM

PR #8087, opened 2022-02-08
werft/config: Delete core-dev namespace on branch delete

ArthurSens:

## Description

@mads-hartmann @csweichel I'm opening this one as a draft because I'm not entirely sure this is all that we need to do.

How do we tell wipe-devstaging the namespace to delete?

## Related Issue(s)

Fixes https://github.com/gitpod-io/ops/issues/981

## How to test

Also a hard one to test, since this needs to be on main to execute

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.werft/delete-preview-environments/delete-preview-environment.ts
+++ b/.werft/delete-preview-environments/delete-preview-environment.ts
@@ -1,0 +1,50 @@
+import { Werft } from '../util/werft';
+import * as fs from 'fs';
+import * as Tracing from '../observability/tracing';
+import { SpanStatusCode } from '@opentelemetry/api';
+import { wipePreviewEnvironmentAndNamespace, helmInstallName } from '../util/kubectl';
+
+// Will be set once tracing has been initialized
+let werft: Werft
+
+const context = JSON.parse(fs.readFileSync('context.json').toString());
+
+Tracing.initialize()
+    .then(() => {
+        werft = new Werft("delete-preview-environment")
+    })
+    .then(() => deletePreviewEnvironment())
+    .then(() => werft.endAllSpans())
+    .catch((err) => {
+        werft.rootSpan.setStatus({
+            code: SpanStatusCode.ERROR,
+            message: err
+        })
+        werft.endAllSpans()
+    })
+
+async function deletePreviewEnvironment() {
+    werft.phase("preparing deletion")
+    const version = parseVersion();
+    const namespace = `staging-${version.split(".")[0]}`
+    werft.log("preparing deletion", `Proceeding to delete the ${namespace} namespace`)
+    werft.done("preparing deletion")
+
+    werft.phase("delete-preview")
+    try {
+        await wipePreviewEnvironmentAndNamespace(helmInstallName, namespace, { slice: 'delete-preview' })
+    } catch (err) {
+        werft.fail("delete-preview", err)
+    }
+    werft.done("delete-prewview")
+}
+
+
+export function parseVersion() {
+    let version = context.Name;
+    const PREFIX_TO_STRIP = "gitpod-delete-preview-environment-";
+    if (version.substr(0, PREFIX_TO_STRIP.length) === PREFIX_TO_STRIP) {
+        version = version.substr(PREFIX_TO_STRIP.length);
+    }
+    return version
+}
--- a/.werft/delete-preview-environments/delete-preview-environment.yaml
+++ b/.werft/delete-preview-environments/delete-preview-environment.yaml
@@ -1,0 +1,54 @@
+pod:
+  serviceAccount: werft
+  affinity:
+    nodeAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        nodeSelectorTerms:
+        - matchExpressions:
+          - key: dev/workload
+            operator: In
+            values:
+            - "builds"
+  volumes:
+  - name: gcp-sa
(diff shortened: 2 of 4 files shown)


csweichel:
> How do we tell wipe-devstaging the namespace to delete?

We derive that from the branch the job was triggered on. 
For good measure, we should remove the "wipe all" behaviour from the job to avoid accidentally removing all preview environments.

mads-hartmann (commented):
A few early comments (In addition to Chris' suggestion of removing the the "wipe all" feature from the job)

1. Lets rename the job to delete-preview-environment and add ourselves as the CODEOWNER
2. For figuring out the namespace I the information is part of the context (e.g. you can see what such a context looks like based on the dump that's done in the build [here](https://github.com/gitpod-io/gitpod/blob/main/.werft/build.yaml#L206) - e.g. [here's an example](https://werft.gitpod-dev.com/job/gitpod-build-main.2392/raw).

ArthurSens:
> Lets rename the job to delete-preview-environment and add ourselves as the CODEOWNER

I believe we want to create a new job for this... We currently have 38 open pull requests, with some of them are open for a really long time...

I think we want to keep wipe-devstaging as it is today (deleting previews older than 2 days), but create a new job that deletes previews as soon as PRs are merged/branch is destroyed

mads-hartmann:
@ArthurSens feel free to create a new job for this PR ☺️ 

Then we can follow up with wipe-devstaging after. I'm leaning towards removing wipe-devstaging and also sweeper in favour or a nightly Werft job that performs cleanup of old preview environments. But I need to understand the use-case for wipe-devstaging and Sweeper a bit better first.

mads-hartmann:
@ArthurSens with https://github.com/gitpod-io/gitpod/pull/8116 merged I was able to remove the other teams as reviewers :)

.werft/delete-preview-environments/delete-preview-environment.yaml
```diff
@@ -0,0 +1,61 @@
...
+          name: honeycomb-api-key
+          key: apikey
+    command:
+      - bash
+      - -c
+      - |
+        sleep 1
+        set -Eeuo pipefail
+
+        export DOCKER_HOST=tcp://$NODENAME:2375
```
mads-hartmann, line 53: I don't believe you need for this job as you're not performing any Leeway builds. If so you can also remove the NODENAME env

.werft/delete-preview-environments/delete-preview-environment.yaml
```diff
@@ -0,0 +1,61 @@
...
+        sudo chown -R gitpod:gitpod /workspace
+
+        (cd .werft && yarn install && mv node_modules ..) | werft log slice prep
+        printf '{{ toJson . }}' > context.json
+
+        npx ts-node .werft/delete-preview-environments/delete-preview-environments.ts
+sidecars:
```
mads-hartmann, line 60: You don't need any sidecars for this job I believe

.werft/delete-preview-environments/delete-preview-environment.yaml
```diff
@@ -0,0 +1,61 @@
+# debug using `werft run github -f -s .werft/build.js -j .werft/build.yaml -a debug=true`
```
mads-hartmann, line 1: I'd just get rid of this :)

.werft/delete-preview-environments/delete-preview-environment.ts
```diff
@@ -0,0 +1,50 @@
...
+        await wipePreviewEnvironmentAndNamespace(helmInstallName, namespace, { slice: 'delete-preview' })
+    } catch (err) {
+        werft.fail("delete-preview", err)
+    }
+    werft.done("delete-prewview")
+}
+
+
+export function parseVersion() {
+    let version = context.Name;
```
mads-hartmann, line 44: Are you sure the name of the namespace can be derived by the Job name?

I don't see any delete-preview-* jobs on https://werft.gitpod-dev.com/ - did you run this? :) 

I was expecting you'd have to get the namespace based on Repository.ref but I might be wrong.

.werft/delete-preview-environments/delete-preview-environment.yaml
```diff
@@ -0,0 +1,61 @@
...
+        sleep 1
+        set -Eeuo pipefail
+
+        export DOCKER_HOST=tcp://$NODENAME:2375
+        sudo chown -R gitpod:gitpod /workspace
+
+        (cd .werft && yarn install && mv node_modules ..) | werft log slice prep
+        printf '{{ toJson . }}' > context.json
+
+        npx ts-node .werft/delete-preview-environments/delete-preview-environments.ts
```
mads-hartmann, line 59: You could leave this out of this PR and we could merge it to test it on main. That way we get to see the contents of context.json when branches are deleted to verify it has the information we need to compute the namespace name

mads-hartmann (changes requested):
Left a few suggestions :)

.werft/delete-preview-environments/delete-preview-environment.ts
ArthurSens: I just copied this part from `build.ts`.

Over there the namespace is being derived from the job name 🤔
I believe because werft already does the name sanitization (get rid of unsupported characters)

mads-hartmann (approved):

PR #8094, opened 2022-02-08
Create an index for contextURL column

### LINKED ISSUE #8005: Searches for workspaces never return
### Bug description

When searching for a workspaceID (e.g. `0a88a641-6bfb-4fe1-8924-84c7e98bff07`) on https://gitpod.io/admin/workspaces, the icon (magnification glass) keeps spinning for 10min+ and I never get search results. 

### Steps to reproduce

(see above)

### Workspace affected

_No response_

### Expected behavior

The search should return in less than a second.

### Example repository

_No response_

### Anything else?

Is there maybe an index on the DB missing? I assume this would need to be a full-text index over multiple columns. Not sure if MySQL can do that or if this has other unwanted side-effects.

laushinka:
## Description
The workspace search queries the contextURL column of the db_workspace table and that was not indexed.
The hypothesis is that indexing this column will improve the search, which can currently be seen in this Honeycomb query[[1](https://ui.honeycomb.io/gitpod/datasets/gitpod-production/result/nZ7mf4rWHDW)].

## Related Issue(s)
Fixes #8005 

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/gitpod-db/src/typeorm/entity/db-workspace.ts
+++ b/components/gitpod-db/src/typeorm/entity/db-workspace.ts
@@ -27,6 +27,7 @@ export class DBWorkspace implements Workspace {
     ownerId: string;
 
     @Column("text")
+    @Index('ind_contextURL')
     contextURL: string;
 
     @Column({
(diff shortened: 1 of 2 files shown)


JanKoehnlein (approved):
LGTM

PR #8095, opened 2022-02-08
Fix credential helper regression

jeanp413:
## Description

Server could return only the token and not the username, so we fallback to hardcoded `oauth2` username.
See https://github.com/gitpod-io/gitpod/pull/7889#discussion_r801670957

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/gitpod-cli/cmd/credential-helper.go
+++ b/components/gitpod-cli/cmd/credential-helper.go
@@ -46,11 +46,16 @@ var credentialHelper = &cobra.Command{
 		defer func() {
 			// Credentials not found, return `quit=true` so no further helpers will be consulted, nor will the user be prompted.
 			// From https://git-scm.com/docs/gitcredentials#_custom_helpers
-			if user == "" || token == "" {
+			if token == "" {
 				fmt.Print("quit=true\n")
-			} else {
-				fmt.Printf("username=%s\npassword=%s\n", user, token)
+				return
 			}
+			// Server could return only the token and not the username, so we fallback to hardcoded `oauth2` username.
+			// See https://github.com/gitpod-io/gitpod/pull/7889#discussion_r801670957
+			if user == "" {
+				user = "oauth2"
+			}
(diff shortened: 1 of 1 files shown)


iQQBot (dismissed):
LGTM

components/gitpod-cli/cmd/credential-helper.go
```diff
@@ -46,11 +46,15 @@ var credentialHelper = &cobra.Command{
 		defer func() {
 			// Credentials not found, return `quit=true` so no further helpers will be consulted, nor will the user be prompted.
 			// From https://git-scm.com/docs/gitcredentials#_custom_helpers
-			if user == "" || token == "" {
+			if token == "" {
 				fmt.Print("quit=true\n")
```
iQQBot, line 50: should be return?

iQQBot: Otherwise, if the username and password are both empty, then the below will be used

iQQBot: But it doesn't seem to matter, if you confirm it, you can always /unhold

akosyakov:
Could you add a comment in the code why you need to revert for next time at least? 🙏 or unit test of print function will meaningful tests names

components/gitpod-cli/cmd/credential-helper.go
```diff
@@ -46,11 +46,16 @@ var credentialHelper = &cobra.Command{
...
-			} else {
-				fmt.Printf("username=%s\npassword=%s\n", user, token)
+				return
 			}
+			// Server could return only the token and not the username, so we fallback to hardcoded `oauth2` username.
+			// See https://github.com/gitpod-io/gitpod/pull/7889#discussion_r801670957
+			if user == "" {
+				user = "oauth2"
+			}
```
jeanp413, line 57: @akosyakov This is the reason why this change

iQQBot (approved):
LGTM

PR #8099, opened 2022-02-08
[server, dashboard] Fix usage and parsing of ws.ontextURL

### LINKED ISSUE #8097: Workspace list stays blank when displaying `git` contextURLs
Example

image

/cc @jankeromnes

geropl:
## Description
 1. Never fail when trying to parse `contextURL`
 2. prefer `normalizedContextURL` if present
 3. if `normalizedContextURL` is not present, try to parse it from `contextURL`
 3. ~handle broken "git@github:" contextURLs in the DB (@jankeromnes )~ Done with #8100

## Related Issue(s)

Fixes #8097

## Release Notes

```release-note
fix dashboard contextURL handling
```

### DIFF
--- a/components/dashboard/src/admin/WorkspaceDetail.tsx
+++ b/components/dashboard/src/admin/WorkspaceDetail.tsx
@@ -63,7 +63,7 @@ export default function WorkspaceDetail(props: { workspace: WorkspaceAndInstance
                 <div className="flex w-full mt-6">
                     <Property name="Created">{moment(workspace.workspaceCreationTime).format('MMM D, YYYY')}</Property>
                     <Property name="Last Start">{moment(workspace.instanceCreationTime).fromNow()}</Property>
-                    <Property name="Context"><a className="text-blue-400 dark:text-blue-600 hover:text-blue-600 dark:hover:text-blue-400" href={ContextURL.parseToURL(workspace.contextURL)?.toString()}>{workspace.context.title}</a></Property>
+                    <Property name="Context"><a className="text-blue-400 dark:text-blue-600 hover:text-blue-600 dark:hover:text-blue-400" href={ContextURL.getNormalizedURL(workspace)?.toString()}>{workspace.context.title}</a></Property>
                 </div>
                 <div className="flex w-full mt-6">
                     <Property name="User"><Link className="text-blue-400 dark:text-blue-600 hover:text-blue-600 dark:hover:text-blue-400" to={"/admin/users/" + props.workspace.ownerId}>{user?.name || props.workspace.ownerId}</Link></Property>
--- a/components/dashboard/src/admin/WorkspacesSearch.tsx
+++ b/components/dashboard/src/admin/WorkspacesSearch.tsx
@@ -4,7 +4,7 @@
  * See License-AGPL.txt in the project root for license information.
  */
 
-import { AdminGetListResult, AdminGetWorkspacesQuery, User, WorkspaceAndInstance } from "@gitpod/gitpod-protocol";
+import { AdminGetListResult, AdminGetWorkspacesQuery, ContextURL, User, WorkspaceAndInstance } from "@gitpod/gitpod-protocol";
 import { matchesInstanceIdOrLegacyWorkspaceIdExactly, matchesNewWorkspaceIdExactly } from "@gitpod/gitpod-protocol/lib/util/parse-workspace-id";
 import moment from "moment";
 import { useContext, useEffect, useState } from "react";
@@ -133,7 +133,7 @@ function WorkspaceEntry(p: { ws: WorkspaceAndInstance }) {
             </div>
             <div className="flex flex-col w-5/12 self-center truncate">
                 <div className="text-gray-500 overflow-ellipsis truncate">{p.ws.description}</div>
-                <div className="text-sm text-gray-400 overflow-ellipsis truncate">{p.ws.contextURL}</div>
+                <div className="text-sm text-gray-400 overflow-ellipsis truncate">{ContextURL.getNormalizedURL(p.ws)?.toString()}</div>
             </div>
             <div className="flex w-2/12 self-center">
                 <div className="text-sm w-full text-gray-400 truncate">{moment(p.ws.instanceCreationTime || p.ws.workspaceCreationTime).fromNow()}</div>
--- a/components/dashboard/src/start/CreateWorkspace.tsx
+++ b/components/dashboard/src/start/CreateWorkspace.tsx
@@ -6,7 +6,7 @@
 
 import EventEmitter from "events";
 import React, { useEffect, Suspense, useContext, useState } from "react";
-import { CreateWorkspaceMode, WorkspaceCreationResult, RunningWorkspacePrebuildStarting } from "@gitpod/gitpod-protocol";
+import { CreateWorkspaceMode, WorkspaceCreationResult, RunningWorkspacePrebuildStarting, ContextURL } from "@gitpod/gitpod-protocol";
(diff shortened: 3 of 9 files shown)


components/gitpod-protocol/src/context-url.ts
```diff
@@ -4,41 +4,88 @@
...
+      const u = removePrefixes(ws.contextURL);
+      if (u) {
+        return u;
+      }
+    } catch (err) {
+      console.error(err);
+    }
+
+    return undefined;
+  }
```
jankeromnes, line 57: I agree that in most cases we should prefer `ws.context.normalizedContextURL`. However, isn't it a little overkill to have two different methods, `getNormalized` and `getNormalizedURL`?

I think I would prefer having just `getNormalizedURL` to keep this protocol helper as simple as possible. _(Side note: I find the function name `getNormalized(workspace)` a little bit ambiguous, and had to read the code to understand what it does.)_

components/dashboard/src/start/StartWorkspace.tsx
```diff
@@ -226,9 +226,9 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
 
     // Successfully stopped and headless: the prebuild is done, let's try to use it!
     if (!error && workspaceInstance.status.phase === 'stopped' && this.state.workspace?.type !== 'regular') {
-      const contextURL = ContextURL.parseToURL(this.state.workspace?.contextURL);
+      const contextURL = ContextURL.getNormalized(this.state.workspace, false);
```
jankeromnes, line 229: Shouldn't this actually "fall back to raw" when there is no normalizedContextURL? 🤔

Could also be simplified:

```ts
this.redirectTo('/' + (this.state.workspace?.context?.normalizedContextURL || this.state.workspace?.contextURL)
```

akosyakov:
Maybe also update all places in the dashboard where  `workspace.contextUrl` as user facing is used with this new method.

components/dashboard/src/admin/WorkspaceDetail.tsx
```diff
@@ -63,7 +63,7 @@ export default function WorkspaceDetail(props: { workspace: WorkspaceAndInstance
                 <div className="flex w-full mt-6">
                     <Property name="Created">{moment(workspace.workspaceCreationTime).format('MMM D, YYYY')}</Property>
                     <Property name="Last Start">{moment(workspace.instanceCreationTime).fromNow()}</Property>
-                    <Property name="Context"><a className="text-blue-400 dark:text-blue-600 hover:text-blue-600 dark:hover:text-blue-400" href={ContextURL.parseToURL(workspace.contextURL)?.toString()}>{workspace.context.title}</a></Property>
+                    <Property name="Context"><a className="text-blue-400 dark:text-blue-600 hover:text-blue-600 dark:hover:text-blue-400" href={ContextURL.getNormalized(workspace)}>{workspace.context.title}</a></Property>
```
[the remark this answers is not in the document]
geropl: IMO the expectation here is to go the context, not open the workspace.

components/gitpod-protocol/src/context-url.ts
```diff
@@ -4,41 +4,88 @@
...
+import { Workspace } from ".";
+
 export namespace ContextURL {
   export const INCREMENTAL_PREBUILD_PREFIX = "incremental-prebuild";
   export const PREBUILD_PREFIX = "prebuild";
   export const IMAGEBUILD_PREFIX = "imagebuild";
+  export const SNAPSHOT_PREFIX = "snapshot";
   export const REFERRER_PREFIX = 'referrer:';
 
+  export function getNormalized(ws: Pick<Workspace, "context" | "contextURL"> | undefined, fallbackToRaw: boolean = true): string | undefined {
```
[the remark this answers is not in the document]
geropl: Agreed. The motivation here was to make no external changes at all to minimize the PR surface. But will change now that we have more time to polish this.

components/gitpod-protocol/src/context-url.ts
geropl: Same answer as here: https://github.com/gitpod-io/gitpod/pull/8099#discussion_r801868556

components/dashboard/src/start/StartWorkspace.tsx
geropl: I don't think so: In case the workspace was a prebuild, we now want to start a fresh workspace on it's context and leverage it (cmp. comment 2 lines above).

geropl:
> Maybe also update all places in the dashboard where workspace.contextUrl as user facing is used with this new method.

@akosyakov Excellent point, will do.

components/dashboard/src/start/CreateWorkspace.tsx
```diff
@@ -163,14 +163,17 @@ export default class CreateWorkspace extends React.Component<CreateWorkspaceProp
...
-            {result?.existingWorkspaces?.map(w =>
-              <a href={w.latestInstance?.ideUrl || gitpodHostUrl.with({ pathname: '/start/', hash: '#' + w.latestInstance?.workspaceId }).toString()} className="rounded-xl group hover:bg-gray-100 dark:hover:bg-gray-800 flex p-3 my-1">
-                <div className="w-full">
-                  <p className="text-base text-black dark:text-gray-100 font-bold">{w.workspace.id}</p>
-                  <p className="truncate" title={w.workspace.contextURL}>{w.workspace.contextURL}</p>
-                </div>
-              </a>
-            )}
+            {result?.existingWorkspaces?.map(w => {
+              const normalizedContextUrl = ContextURL.getNormalizedURL(w.workspace)?.toString() || "undefined";
```
geropl, line 167: Note: This is one of the places where I was not 100% sure if we really want the `normalizedContextUrl`, but ultimately think we should always display it (except for debugging purposed in `/admin`, maybe).

geropl:
> One is: Changing the protocol API risks breaking downstream clients, right? Do we know if any other code than the dashboard relies on ContextURL.parseToURL?

The only real client here is dashboard. :+1: The Gitpod plugin for vscode server also uses parts from `gitpod-protocol`, but 1) they do not use this, and 2) they import during build, not runtime, so their build would fail and they'd have to adjust.

Most importantly, maybe, 3): We should not consider `gitpod-protocol` to be public API. It's 99% the intersection of `dashboard` and `server` for convenience sake. And hopefully replaced with something more stable, soon. :slightly_smiling_face:

components/dashboard/src/admin/WorkspacesSearch.tsx
```diff
@@ -134,6 +134,7 @@ function WorkspaceEntry(p: { ws: WorkspaceAndInstance }) {
             <div className="flex flex-col w-5/12 self-center truncate">
                 <div className="text-gray-500 overflow-ellipsis truncate">{p.ws.description}</div>
                 <div className="text-sm text-gray-400 overflow-ellipsis truncate">{p.ws.contextURL}</div>
```
JanKoehnlein, line 136: Should we get rid of this div?

JanKoehnlein (commented):
LGTM apart from the duplicate div

components/dashboard/src/admin/WorkspacesSearch.tsx
geropl: @JanKoehnlein done.

JanKoehnlein (approved):
Supi

akosyakov:
@geropl It will be cool to add a warning comment here: https://github.com/gitpod-io/gitpod/blob/95592d00d8dcce49539d39c68c22ac980d9581c6/components/gitpod-protocol/src/protocol.ts#L413

Issue #7575, opened 2022-01-12
Epic: GitHub "Governance" Processes


loujaybee:
An issue to help improve visibility on how GitHub issues are triaged, how roadmapping works, how labels are applied etc. Please use this issue to share thoughts and suggestions on how we can improve the GitHub contributing experience.

loujaybee:
I'm hoping that this small docs clarification https://github.com/gitpod-io/gitpod/pull/8120 and a full commitment to the inbox process for teams should be sufficient here. I think we also need to ensure that all folks doing issue triage also fully utilise the inbox process and that becomes the one and only way of ensuring an official update to any issue.

PR #8120, opened 2022-02-09
Clarify ways that GitHub issues get prioritised

loujaybee:
## Description

Given that we now have a newly structured roadmap [[1](https://www.gitpod.io/roadmap)], that links more closely to the team boards [[1](https://github.com/orgs/gitpod-io/projects?type=beta)], this PR attempts to clarify to externals the current roadmap process. 

## Related Issue(s)

Closes https://github.com/gitpod-io/gitpod/issues/7575


README.md
```diff
@@ -73,6 +73,10 @@ We work with quarterly roadmaps in autonomous product teams.
...
  - [Product Roadmap](https://github.com/orgs/gitpod-io/projects/27)
 
+### How do GitHub Issues get prioritised? 
+
+Most GitHub issues (except smaller or more urgent issues) relate to our [current product roadmap items](https://github.com/orgs/gitpod-io/projects/27). Gitpod teams work against these roadmap items. Each Gitpod team has [it's own project board](https://github.com/orgs/gitpod-io/projects) that follows a similar structure. You can find these project boards attached to [the GitHub organisation](https://github.com/gitpod-io). Each team board has a "GroundWork" tab which shows current GitHub issues in progress. Each team project board also has an "inbox" where issues are sent for review by the team (and should be responded to within 48 hours). "Upvoting" by [reacting](https://docs.github.com/en/rest/reference/reactions) to GitHub issues helps signal to Gitpod that issues are important to you. If you are unsure of the status of an issue, please comment and a Gitpodder should respond to you shortly. For any other questions, please utilise the [Gitpod community](https://www.gitpod.io/community).
```
loujaybee, line 78: One aspect missed here is the relation to the priority labels. 

@csweichel or @gtsiolis (as you are marked as owners of the internal labels document [[1](https://www.notion.so/gitpod/Issue-Labels-764f3823a395448ca25dec69a49f5b87)]) - are you able to make a suggested wording update to clarify the use of the priority labels? 

I would suggest here that we note that labels are used as a way to flag an issue, but not necessarily dictate when the issue might get completed, as that's the documented groundwork / roadmap process already mentioned?

csweichel: Thank you for that update!

I'm not sure we'd want to specifically add priority labels here because 
- we'd want to prioritise issues ourselves, and not externalise that. The labels might not be described well enough to be used by members of the community - and I'm not sure we'd want to encourage folks outside of Gitpod to apply priority labels.
- we'd do well to describe our labels entirely, not just priority labels. To that end, we could make the Notion page public and link it here.

svenefftinge (dismissed):
Thanks, Lou!

README.md
loujaybee: > I'm not sure we'd want to specifically add priority labels here - @csweichel 

Yeah, I was originally thinking to note here to ignore them, e.g. "Please ignore labels for the purpose of prioritisation" but wasn't sure if that would be an accurate reflection of their use. Things certainly get fuzzy when we mix the inbox / groundwork / roadmap processes with labels 😬 

> we'd do well to describe our labels entirely, not just priority labels - @csweichel 

> Just realized that labels have a description. I think we should just make good use of them. - @svenefftinge 

Agree 🙏 We can look into clarifying labels in future then and update in a subsequent PR 🙏

PR #8036, opened 2022-02-04
[ws-daemon] Refactor CPU limiting

csweichel:
## Description
This PR refactors the dynamic CPU limiting performed by ws-daemon. More details regarding motivation and solution can be found in this loom video: https://www.loom.com/share/0922bf11430c4ac084ebab6616e40b7e

This PR also adds a set of metrics to observe the new CPU limits:
- `gitpod_ws_daemon_cpulimit_workspaces_added_total`: Number of workspaces added to CPU control
- `gitpod_ws_daemon_cpulimit_workspaces_removed_total` Number of workspaces removed from CPU control
- `gitpod_ws_daemon_cpulimit_workspaces_burst_total`: Number of workspaces which received burst - `gitpod_ws_daemon_cpulimit_workspaces_cputime_seconds` CPU time of all observed workspaces
- `gitpod_ws_daemon_cpulimit_workspaces_throttled_total` Number of workspaces which ran with throttled CPU

## Related Issue(s)

Fixes #8035

## How to test
By default dynamic CPU limiting is disabled. It can be enabled using the experimental config in the installer, or by modifying the ws-daemon configmap. Therein, set the cpuLimiting.enabled to true, and configure the total available bandwidth on the node (e.g. 12 CPUs), limit and burst limit.

Then, start a few workspaces, observe their `/sys/fs/cgroup/cpu/cpu.cfs_quota_us` value while producing some CPU load.

In addition, one could try this version on a single VM and experiment with workspaces exhibiting different load profiles.

## Chaveats
- this PR removes the process priority management we had prior. It's unclear how much that actually helped, considering that it very outdated and focused on `bash` and `theia` and `gitpod-node` only. If we find user experience deteriorated too much by this removal, we'll want to bring it back in a future change.
- this change does not yet introduce QoS classes, but that would be straight forward to add down the road using annotations on the pod.
- this change removes the fixed CPU limiting that was in place as agent-smith action (limit CPU), and that was meant to afford paying customers more CPU. It's doubtful that the latter had much of an effect, but we'll want to follow up quickly with the QoS change.

## Release Notes

```release-note
Refactor dynamic CPU limiting to provide fairer scheduling.
```

PR #8124, opened 2022-02-09
Update helm installer with config for new cpu limiter

Furisto:
## Description
#8036 reworked the cpu limiting that ws-daemon is doing. We also need to update the helm installer with the new configuration for ws-daemon. This broke staging as ws-daemon expected the new configuration but the configmap still had the old configuration.

## Related Issue(s)
n.a.

## How to test
Open [preview environment](https://furisto-cpu-config.staging.gitpod-dev.com/workspaces) which has been installed with helm and check that ws-daemon is running and workspaces can be opened

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/ws-daemon/pkg/cpulimit/dispatch.go
+++ b/components/ws-daemon/pkg/cpulimit/dispatch.go
@@ -26,7 +26,7 @@ type Config struct {
 	Enabled        bool              `json:"enabled"`
 	TotalBandwidth resource.Quantity `json:"totalBandwidth"`
 	Limit          resource.Quantity `json:"limit"`
-	BurstLimit     resource.Quantity `json:"bustLimit"`
+	BurstLimit     resource.Quantity `json:"burstLimit"`
 
 	ControlPeriod  util.Duration `json:"controlPeriod"`
 	CGroupBasePath string        `json:"cgroupBasePath"`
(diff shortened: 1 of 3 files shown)


components/ws-daemon/pkg/cpulimit/dispatch.go
```diff
@@ -26,7 +26,7 @@ type Config struct {
 	Enabled        bool              `json:"enabled"`
 	TotalBandwidth resource.Quantity `json:"totalBandwidth"`
 	Limit          resource.Quantity `json:"limit"`
-	BurstLimit     resource.Quantity `json:"bustLimit"`
+	BurstLimit     resource.Quantity `json:"burstLimit"`
```
sagor999, line 29: @csweichel heh. I suspect due to this typo bursting might not be working quite right in current gen31 deploy. FYI.

sagor999 (approved):

components/ws-daemon/pkg/cpulimit/dispatch.go
Furisto: The installer also has bustLimit, so it should work. Planning on updating that next.

sagor999: Ah you are right! Thank you!

csweichel: the installer uses the same structure, hence would have the same (incorrect) behaviour

PR #8125, opened 2022-02-09
[dashboard, supervisor, ws-proxy] Break redirect loop on failing workspaces starts

### LINKED ISSUE #8043: [dashboard, server] Prevent redirect loops that trigger "startWorkspace" in a loop

geropl:
## Description
This PR breaks the restart cycles we recently oberserve for workspaces that fail quickly enough.

The basic cycle was:
 1. `StartWorkspace` (dashboard, /start) -> onload: `server.startWorkspace`
 2. `server.startWorkspace` returns :+1: inkl. `ideUrl` (1. instance)
 2. `StartWorkspace` redirects to `ideUrl`
 3. `StartWorkspace` (inframe, IDE url) -> onload: `server.startWorkspace`:
    a. expecting `server` to return the already running instance from step 1)
    b. but that was already stopped with an error, so started a fresh one (2. instance)
    c. `supervisor` frontend suspected a re-start (because it sees `stopped` updates from 1. instance, followed by `pending`/`creating` from 2.), and to properly re-connect the frontend, re-directs to `StartWorkspace` :repeat: 

This PR changes three things:
 - if workspace is not found, [`ws-proxy`](https://github.com/gitpod-io/gitpod/blob/2c302638122f9471cac25371d0b765a14aeaf22f/components/ws-proxy/pkg/proxy/routes.go#L553) now re-direct to `/start/?not_found=true`. `StartWorkspace` reads that parameter and [sets `dontAutostart`](https://github.com/gitpod-io/gitpod/blob/1b0bd82d7f43a1f244d9596f5370139649ee3aca/components/dashboard/src/start/StartWorkspace.tsx#L128-L140), which in turn ensures `startWorkspace(id)` is not called on load
 - when `StartWorkspace` is loaded inside an iframe (when already on the ideUrl) [`dontAutostart: true` is set as well](https://github.com/gitpod-io/gitpod/blob/1b0bd82d7f43a1f244d9596f5370139649ee3aca/components/dashboard/src/start/StartWorkspace.tsx#L142-L152).
 - the [reload behavior in `supervisor` frontend](https://github.com/gitpod-io/gitpod/pull/8125/files#diff-3cab7118ea66c275e7be9d00fb70116dae530eeb132a44a1fd643639748c1a11L145-L157) got removed, and replaced with a [similar](https://github.com/gitpod-io/gitpod/pull/8125/files#diff-d897e36a784992a6a4d3b76c09147a96375add6c80bd14dca49657dcff06b3a7R275-R291) mechanism in `StartWorkspace`.

### Effects on the "start workspace" flow
 - :+1:  if you come back to a (very) old workspace tabs that were unloaded, a mere "tab activation" will no longer re-start the workspace. Instead, you have to actively hit "Open Workspace".
 - :+1: we do not call `startWorkspace` that often, which should help with not confusing the rate-limiter (and potentially fasten the startup process :crossed_fingers: )
 - :-1: maybe: If there have been other errors in the startup-chain, they have so-long been covered by the "auto-restart" behavior. Now users have to actively click "Open workspace" again in such cases. Not sure how relevant this is, and IMO we have to find out
 - :-1: ... ?

## UI
Currently the cycle breaks with this screen:
image
 

Alternatively we could use a modal but I found it to be bad because:
 - it hides the error message
 - if we added the error message to the screen we'd just duplicate the original screen
**Update**: removed that code

## Alternatives

~Alternatively, and to further strengthen the process, we could try to find a way to pass the `instanceId` from `dashboard` to `supervisor` frontend. This would avoid the whole issue altogether (and make code in supervisor frontend easier), because we know upfront which instance we expect to see here.
The only way I can think of is to make it part of the url; `<workspaceId>.ws-<cluster>.gitpod.io/#<instanceId>`, for example. But I'm unsure if and how this affects IDE, and if yes, how to avoid that (/cc @akosyakov )~

Dismissed [here](https://github.com/gitpod-io/gitpod/pull/8125#issuecomment-1034694867).

**Note**: There sometimes seems to be CORS issues with redirects from `ws-proxy`, but sometimes they work. I assumed them to be out-of-scope here, because this PR does not touch CORS or any domains used.

## Related Issue(s)

Fixes #8043

### Negative
 - in this preview env all workspaces will fail: https://gpl-8043-break-redirect-loop.staging.gitpod-dev.com/workspaces
 - tab 1: start [any workspace](https://gpl-8043-break-redirect-loop.staging.gitpod-dev.com/#https://github.com/gitpod-io/template-typescript-node)
 - tab 2: open the same workspace
 - note how both end up in the error state
 - hit "open workspace" again on one tab
 - note how a) both update, and show up-to-date state, and b) both end up in the error state again

### Positive
 - in this preview env all workspaces should succeed: https://gpl-8043-break-redirect-loop-w.staging.gitpod-dev.com/workspaces
 - tab 1: start [any workspace](https://gpl-8043-break-redirect-loop-w.staging.gitpod-dev.com/#https://github.com/gitpod-io/template-typescript-node)
 - tab 2: open the same workspace
 - note how both end up in the IDE
 - stop the workspace
 - note how both end up in "Stopped"
 - open the workspace again from tab 2
 - Note how:
   - a) both workspaces show up-to-date state
   - b) only tab 2 reveals the IDE
   - c) tab 1 still shows the up-to-date state (RUNNING), incl. controls to open/reveal the IDE (with fresh credentials)
image

## Release Notes

```release-note
make workspace startup more robust
```

### DIFF
--- a/components/dashboard/src/start/StartWorkspace.tsx
+++ b/components/dashboard/src/start/StartWorkspace.tsx
@@ -8,6 +8,7 @@ import { ContextURL, DisposableCollection, GitpodServer, RateLimiterError, Start
 import { IDEOptions } from "@gitpod/gitpod-protocol/lib/ide-protocol";
 import { ErrorCodes } from "@gitpod/gitpod-protocol/lib/messaging/error";
 import EventEmitter from "events";
+import * as queryString from "query-string";
 import React, { Suspense, useEffect } from "react";
 import { v4 } from 'uuid';
 import Arrow from "../components/Arrow";
@@ -21,10 +22,54 @@ const sessionId = v4();
 const WorkspaceLogs = React.lazy(() => import('../components/WorkspaceLogs'));
 
 export interface StartWorkspaceProps {
-  workspaceId: string;
+  workspaceId: string,
+  runsInIFrame: boolean,
+  /**
+   * This flag is used to break the autostart-cycle explained in https://github.com/gitpod-io/gitpod/issues/8043
+   */
+  dontAutostart: boolean,
+}
+
+export function parseProps(workspaceId: string, search?: string): StartWorkspaceProps {
+  const params = parseParameters(search);
+  const runsInIFrame = window.top !== window.self;
+  return {
+    workspaceId,
+    runsInIFrame: window.top !== window.self,
+    // Either:
+    //  - not_found: we were sent back from a workspace cluster/IDE URL where we expected a workspace to be running but it wasn't because either:
+    //    - this is a (very) old tab and the workspace already timed out
+    //    - due to a start error our workspace terminated very quickly between:
+    //      a) us being redirected to that IDEUrl (based on the first ws-manager update) and
+    //      b) our requests being validated by ws-proxy
+    //  - runsInIFrame (IDE case):
+    //    - we assume the workspace has already been started for us
+    //    - we don't know it's instanceId
+    dontAutostart: params.notFound || runsInIFrame,
+  }
+}
+
+function parseParameters(search?: string): { notFound?: boolean } {
+  try {
+    if (search === undefined) {
+      return {};
+    }
+    const params = queryString.parse(search, {parseBooleans: true});
+    const notFound = !!(params && params["not_found"]);
+    return {
+      notFound,
+    };
+  } catch (err) {
+    console.error("/start: error parsing search params", err);
+    return {};
+  }
 }
 
 export interface StartWorkspaceState {
+  /**
+   * This is set to the istanceId we started (think we started on).
+   * We only receive updates for this particular instance, or none if not set.
+  */
   startedInstanceId?: string;
   workspaceInstance?: WorkspaceInstance;
   workspace?: Workspace;
@@ -34,8 +79,8 @@ export interface StartWorkspaceState {
     link: string
     label: string
     clientID?: string
-  }
-  ideOptions?: IDEOptions
+  };
+  ideOptions?: IDEOptions;
 }
 
 export default class StartWorkspace extends React.Component<StartWorkspaceProps, StartWorkspaceState> {
@@ -47,7 +92,7 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
 
   private readonly toDispose = new DisposableCollection();
   componentWillMount() {
-    if (this.runsInIFrame()) {
+    if (this.props.runsInIFrame) {
       window.parent.postMessage({ type: '$setSessionId', sessionId }, '*');
       const setStateEventListener = (event: MessageEvent) => {
         if (event.data.type === 'setState' && 'state' in event.data && typeof event.data['state'] === 'object') {
@@ -75,8 +120,15 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
       this.setState({ error });
     }
 
+    if (this.props.dontAutostart) {
+      this.fetchWorkspaceInfo(undefined);
+      return;
+    }
+
+    // dashboard case (no previous errors): start workspace as quickly as possible
     this.startWorkspace();
-    getGitpodService().server.getIDEOptions().then(ideOptions => this.setState({ ideOptions }))
+    // query IDE options so we can show them if necessary once the workspace is running
+    getGitpodService().server.getIDEOptions().then(ideOptions => this.setState({ ideOptions }));
   }
 
   componentWillUnmount() {
@@ -130,14 +182,13 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
       }
       console.log("/start: started workspace instance: " + result.instanceID);
       // redirect to workspaceURL if we are not yet running in an iframe
-      if (!this.runsInIFrame() && result.workspaceURL) {
+      if (!this.props.runsInIFrame && result.workspaceURL) {
         this.redirectTo(result.workspaceURL);
         return;
       }
-      this.setState({ startedInstanceId: result.instanceID });
-      // Explicitly query state to guarantee we get at least one update
+      // Start listening too instance updates - and explicitly query state once to guarantee we get at least one update
       // (needed for already started workspaces, and not hanging in 'Starting ...' for too long)
-      this.fetchWorkspaceInfo();
+      this.fetchWorkspaceInfo(result.instanceID);
     } catch (error) {
       console.error(error);
       if (typeof error === 'string') {
@@ -180,15 +231,28 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
     }
   }
 
-  async fetchWorkspaceInfo() {
+  /**
+   * Fetches initial WorkspaceInfo from the server. If there is a WorkspaceInstance for workspaceId, we feed it
+   * into "onInstanceUpdate" and start accepting further updates.
+   *
+   * @param startedInstanceId The instanceId we want to listen on
+   */
+  async fetchWorkspaceInfo(startedInstanceId: string | undefined) {
+    // this ensures we're receiving updates for this instance
+    if (startedInstanceId) {
+      this.setState({ startedInstanceId });
+    }
+
     const { workspaceId } = this.props;
     try {
       const info = await getGitpodService().server.getWorkspace(workspaceId);
       if (info.latestInstance) {
-        this.setState({
-          workspace: info.workspace
-        });
-        this.onInstanceUpdate(info.latestInstance);
+        const instance = info.latestInstance;
+        this.setState((s) => ({
+          workspace: info.workspace,
+          startedInstanceId: s.startedInstanceId || instance.id,  // note: here's a potential mismatch between startedInstanceId and instance.id. TODO(gpl) How to handle this?
+        }));
+        this.onInstanceUpdate(instance);
       }
     } catch (error) {
       console.error(error);
@@ -197,20 +261,35 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
   }
 
   notifyDidOpenConnection() {
-    this.fetchWorkspaceInfo();
+    this.fetchWorkspaceInfo(undefined);
   }
 
   async onInstanceUpdate(workspaceInstance: WorkspaceInstance) {
-    const startedInstanceId = this.state?.startedInstanceId;
-    if (workspaceInstance.workspaceId !== this.props.workspaceId || startedInstanceId !== workspaceInstance.id) {
+    if (workspaceInstance.workspaceId !== this.props.workspaceId) {
       return;
+    }
+
+    // Here we filter out updates to instances we haven't started to avoid issues with updates coming in out-of-order
+    // (e.g., multiple "stopped" events from the older instance, where we already started a fresh one after the first)
+    // Only exception is when we do the switch from the "old" to the "new" one.
+    const startedInstanceId = this.state?.startedInstanceId;
+    if (startedInstanceId !== workspaceInstance.id) {
+      // do we want to switch to "new" instance we just received an update for?
+      const switchToNewInstance = this.state.workspaceInstance?.status.phase === "stopped" && workspaceInstance.status.phase !== "stopped";
+      if (!switchToNewInstance) {
+        return;
+      }
+      this.setState({
+        startedInstanceId: workspaceInstance.id,
+        workspaceInstance,
+      });
     }
 
     await this.ensureWorkspaceAuth(workspaceInstance.id);
 
     // Redirect to workspaceURL if we are not yet running in an iframe.
     // It happens this late if we were waiting for a docker build.
-    if (!this.runsInIFrame() && workspaceInstance.ideUrl) {
+    if (!this.props.runsInIFrame && workspaceInstance.ideUrl && !this.props.dontAutostart) {
       this.redirectTo(workspaceInstance.ideUrl);
       return;
     }
@@ -255,22 +334,18 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
   }
 
   redirectTo(url: string) {
-    if (this.runsInIFrame()) {
+    if (this.props.runsInIFrame) {
       window.parent.postMessage({ type: 'relocate', url }, '*');
     } else {
       window.location.href = url;
     }
   }
 
-  runsInIFrame() {
-    return window.top !== window.self;
-  }
-
   render() {
     const { error } = this.state;
     const isHeadless = this.state.workspace?.type !== 'regular';
     const isPrebuilt = WithPrebuild.is(this.state.workspace?.context);
-    let phase = StartPhase.Preparing;
+    let phase: StartPhase | undefined = StartPhase.Preparing;
     let title = undefined;
     let statusMessage = !!error ? undefined : <p className="text-base text-gray-400">Preparing workspace …</p>;
     const contextURL = ContextURL.getNormalizedURL(this.state.workspace)?.toString();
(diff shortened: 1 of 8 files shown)


geropl:
/hold

- [ ] do not merge before todays deployment
- [ ] properly test workspace timeout/stopping as well

jankeromnes:
>  4. `StartWorkspace` (inframe, IDE url) -> onload: `server.startWorkspace`:

If we no longer wish to make the (inframe, IDE url) auto-start an instance, it seems like in this step we should call something like `server.getRunningWorkspaceInstance` instead of `server.startWorkspace`. Wouldn't that be a sufficient fix for this problem?

However, I guess in both cases, we break the behavior of "refresh stopped workspace tab to restart it" (which is probably okay though -- this eager restarting is what caused the problem, and we could instead show the "stopped" screen which has a restart button if you do want to do that).

geropl:
> If we no longer wish to make the (inframe, IDE url) auto-start an instance, it seems like in this step we should call something like server.getRunningWorkspaceInstance instead of server.startWorkspace.

:+1: That's what we do with `this.fetchWorkspaceInfo(undefined)`. It also (needs to) does authentication, error handling, etc.

> Wouldn't that be a sufficient fix for this problem?

No, because we need a "safe place" for `ws-proxy`/`supervisor` to redirect too in case of an error. They don't have the overview (access to DB), and need to delegate back to the dashboard/application cluster.
That could be a separate slug `/restart` instead of `/start`?), or an additional parameter as implemented here. Happy to switch if we can agree on something better. :+1:

jankeromnes:
> No, because we need a "safe place" for ws-proxy/supervisor to redirect too in case of an error.

Ah, I don't understand why a redirect is needed (instead of just staying on the ideURL). I'll read the code to understand better.

geropl:
> I'll read the code to understand better.

What helped me tremendously was to a) have the workspaces auto-fail and b) `yarn telepresence` dashboard to play through the different scenarios! :+1:

jankeromnes:
Thanks! 👍

My current suspicion is that we could simply delete the redirection from supervisor (because it already has an iframe that knows how to show a "Stopped" or error page, right?), however for ws-proxy I don't currently see a good alternative.

I agree that allowing to disable the "start" behavior of `StartWorkspace` (and keep only the "show instance status" part) seems like a good fix.

components/dashboard/src/start/StartWorkspace.tsx
```diff
@@ -21,10 +22,54 @@ const sessionId = v4();
...
+
+export interface StartWorkspaceParameters {
+  trigger?: StartWorkspaceTrigger,
+}
+
+const StartWorkspaceTriggers = {
+  /**
+   * A workspace cluster's ws-proxy redirected the client to "/start/#<wsId>" because it could not find that workspace
+   */
+  "redirect_from_ws_cluster": undefined,
```
akosyakov, line 37: Semehow it feels like tight coupling of components.

Maybe we shoudl rather have here a param like `autostart=false`?

akosyakov:
How can I test other flows besides with failed workspace?

akosyakov:
> The only way I can think of is to make it part of the url; <workspaceId>.ws-<cluster>.gitpod.io/#<instanceId>, for example. But I'm unsure if and how this affects IDE, and if yes, how to avoid that (/cc @akosyakov )

IDE is an owner of URL, I don't think we should mess with it.

jankeromnes (commented):
Thanks a lot for diving deeper into this problem. 🙏

I wonder if we could make this fix a bit simpler. Currently, there is:
- trigger (with a complicated type)
- dontAutostart
- dontRedirect
- showRestartModal

But:
- Do we really need to know where the redirect came from? I think it could be simpler to have just `?autostart=false` as the only (optional) query parameter, regardless of where we came from (that's not really important and introduces too much complexity)
- `dontAutostart` could be replaced by just the `autostart` prop
- is `dontRedirect` really needed? (if we never call `server.startWorkspace()`, do we ever get a `workspaceInstance.ideUrl`?)
- we don't need an extra `showRestartModal`, because the existing UI already knows what to show when the instance is stopped / doesn't exist (i.e. it has a convenient `Run Workspace` button)

Wouldn't it be enough to just have a `/start/?autostart=false#wsID` query parameter, which "simply" prevents the call to `this.startWorkspace()`? 💭 (Maybe I'm missing something)

components/dashboard/src/start/StartWorkspace.tsx
geropl: > Semehow it feels like tight coupling of components.

Yes, but `autostart` as well: IMO `ws-proxy`/`supervisor` should not directly control start behavior.

What about `?startError=true`? That would:
 - solve the tight coupling
 - simplify the code :upside_down_face:

geropl:
> How can I test other flows besides with failed workspace?

@akosyakov You could create a branch from this one, workspaces start by default.

> IDE is an owner of URL, I don't think we should mess with it.

Agreed, so I'll stay with this solution. :+1:

components/dashboard/src/start/StartWorkspace.tsx
geropl: BTW the specific labels where super helpful for debugging, that was the original intention. But we sure don't need to keep them. :+1:

geropl:
@jankeromnes Sorry if I was unclear, IMO we're mixing a few things:
 - the modal /`showRestartModal` was kept as a possible alternative and everyone sees it as bad (incl me), so will remove
 - IMO `dontAutostart` makes clear that this is the exception

> Do we really need to know where the redirect came from?

No, but IMO `?autostart=false` is also not ideal. Discussion: https://github.com/gitpod-io/gitpod/pull/8125#discussion_r803473382

> is dontRedirect really needed? (if we never call server.startWorkspace(), do we ever get a workspaceInstance.ideUrl?)

Yes it is: We get instance updates (incl. some with `ideUrl`) whenever we call `fetchWorkspaceInfo`. And this is needed as well, to solve all kinds of other situations (e.g., stopping situation: we need to eventually switch to `stopped`, and all updates contain `ideURL`)

geropl:
> My current suspicion is that we could simply delete the redirection from supervisor (because it already has an iframe that knows how to show a "Stopped" or error page, right?), however for ws-proxy I don't currently see a good alternative.

Hm. We needed that indirection because we notice a re-starting workspace, and at the moment dashboard is the only place which is allows to refresh authentication credentials. (In theory we could extend that to IDE-URLs, but IMO we have refused to do that because we wanted avoid the need to properly parse IDEUrls - excluding every path/subdomain that's potentially under user control :upside_down_face: )

With the new behavior it could be that it's superfluous... but I'm not sure, and like to avoid including that optimization here, as it would mean a lot of manual test effort, again. :slightly_smiling_face:

components/dashboard/src/start/StartWorkspace.tsx
akosyakov: > IMO ws-proxy/supervisor should not directly control start behavior.

Yeah, I was also wondering there should be better way. But could not figure out still how to pass instance id to supervisor frontend yet :(

components/supervisor/frontend/src/index.ts
```diff
@@ -153,7 +153,7 @@ const toStop = new DisposableCollection();
                 // reload the page if the workspace was restarted to ensure:
                 // - graceful reconnection of IDEs
                 // - new owner token is set
-                window.location.href = startUrl.toString();
+                window.location.href = startUrl.with({ search: "error=true" }).toString();
```
akosyakov, line 156: `error=true` looks strange here though 😬 

Does it make sense to check whether conditions of stopped event were erroneous?

geropl: > error=true looks strange here though

Agreed. :laughing: Do you have a better name?

@jankeromnes Suggested removing this redirect alltogether. It _might_ be not necessary anymore, as we don't have the "auto-restart" expectation anymore. But I'm unsure about all edge-cases: https://github.com/gitpod-io/gitpod/pull/8125#issuecomment-1034734273

> Does it make sense to check whether conditions of stopped event were erroneous?

This is not sufficient, as we don't know which instance to look at. ATM `dontAutostart` basically looks at the "last", while before we'd always look at the "last running (or created on if there was none)".

geropl:
@akosyakov @jankeromnes The PR is rebased, and cleaned up. Now ws-proxy/supervisor redirect to `/start/?error=true#...`.

Test (always fails): https://gpl-8043-break-redirect-loop.staging.gitpod-dev.com/workspaces
Test (should work): https://gpl-8043-break-redirect-loop-w.staging.gitpod-dev.com/workspaces

components/supervisor/frontend/src/index.ts
geropl: > Does loading screen now reload the page?

No, it does not. 

> Otherwise it is going to leave a user with empty page without the loading screen and IDE.

:+1:

akosyakov: Although it seems to break now if i restart a perfectly stopped workspace from the dashboard as well. It reloads and then hangs in `opennnig workspace...` and then times out without ever revealing an IDE.

geropl: Hm, that's strange for sure. That's one of the standard cases... Will re-test and debug again. :thinking: :pray:

akosyakov: Maybe it is alright to remove. ping me if you cannot reproduce it.

csweichel: how about `not-found` instead of `error`?

gtsiolis (commented):
Going through the changes from yesterday, seems that going with the non-modal approach sounds better. 🥊 

Here're some points to consider improving although these are probably out of the scope of this PR.
1️⃣ Remove the last part of the error message as it does not contain any clear action forward. 
2️⃣ Remove the No Changes element as it's probably not relevant or useful here.
3️⃣ Use the red dot indicator as the workspace failed to open.
4️⃣ Keep the same _Open_ action verb for the button label.

| BEFORE | AFTER |
|-|-|
| Screenshot 2022-02-11 at 1 58 22 PM (2) | Screenshot 2022-02-11 at 11 53 28 AM (2) |

Thanks for the ping, @geropl! 🏓

gtsiolis:
@geropl 🙏 

64u4qz

components/dashboard/src/start/StartWorkspace.tsx
```diff
@@ -75,8 +103,35 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
...
     }
 
+    // we're coming back from a workspace cluster/IDE URL where we expected a workspace to be running but it wasn't because either:
+    //  - this is a (very) old tab and the workspace already timed out
+    //  - due to a start error our workspace terminated very quickly between:
+    //    a) us being redirected to that IDEUrl (based on the first ws-manager update) and
+    //    b) our requests being validated by ws-proxy
+    // we break this potentially DDOS cycle by not calling "startWorkspace" automatically but offering a "Open workspace" button instead.
+    const params = this.props.parameters;
+    if (!!params?.not_found) {
```
svenefftinge, line 113: would be nice to name the parameter after what this component actually does (i.e. `noAutoStart`) instead of after one situation in which we would want to have that behavior. Then we also don't need the lengthy comment above.

components/dashboard/src/start/StartWorkspace.tsx
```diff
@@ -75,8 +103,35 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
...
+      this.fetchWorkspaceInfo(undefined);
+      return;
+    }
+
+    // IDE case:
+    //  - we assume the workspace has already been started for us
+    //  - we don't know it's instanceId
+    // we could also rely on "startWorkspace" to return us the "running" instance, but that opens the door for
+    // self-DDOSing as we found in the past (cmp. https://github.com/gitpod-io/gitpod/issues/8043)
+    if (this.runsInIFrame()) {
```
svenefftinge, line 124: The 'runsInIFrame` could also be asked outside of the component and then be passed as `autoStart: boolean` accordingly.

components/dashboard/src/start/StartWorkspace.tsx
geropl: Thx for sharing.

I personally prefer:
 a) to decouple ws-proxy from dashboard (by not controlling behavior)
 b) the lengthy comment: This code is already hopelessly overloaded, and explanations including contexts enable others to follow along without needing to debug every path, again.

I agree that we could map `not_found` => `dontAutostart` outside, though, and make it part of the props. :+1:

geropl: Done.

components/dashboard/src/start/StartWorkspace.tsx
```diff
@@ -317,7 +389,29 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
...
         if (!this.state.desktopIde) {
           phase = StartPhase.Running;
-          statusMessage = <p className="text-base text-gray-400">Opening Workspace …</p>;
+
+          if (this.props.dontAutostart) {
+            // hide the progress bar, as we're already running
+            phase = undefined;
+            title = 'Running';
+
+            // in case we dontAutostart the IDE we have to provide controls to do so
```
akosyakov, line 398: why this logic is not applicable when we run with desktop IDE as well?

geropl: Looking at the code below I think all controls are already in place: there are "Open in Browser" and "Stop Workspace" buttons, so both branches should be equivalent.
The `desktopIDE` case basically never had autostart behavior from it's inception, because you have the IDE choice.

@akosyakov I'd really like to merge today. If there's anything I can do to speed things up (re-testing, sync discussion), please let me know. Thanks for the thorough review! :pray:

components/dashboard/src/start/StartWorkspace.tsx
```diff
@@ -21,10 +22,51 @@ const sessionId = v4();
...
+    dontAutostart: notFound || runsInIFrame,
+  }
+}
+
+function parseParameters(search?: string) {
+  try {
+    if (search === undefined) {
+      return undefined;
+    }
+    return queryString.parse(search, {parseBooleans: true});
```
csweichel, line 58: this looks like a vulnerability in the making. Considering we're making the result of this parse operation props on an element, I'd be weary one could use this for some XSS/prototype poisoning/CSS injection type of attack.

We should make sure we don't allow props we don't expect.

csweichel (commented):
I really like the approach - nice, clean, simple change.
Once the comment is mitigated I'll happily approve :)

components/dashboard/src/start/StartWorkspace.tsx
geropl: Awh, yes! That got removed with one of the simplification steps above - will re-add. :pray:

geropl: Looking at the code again it we did not assign the result right away anymore (that was a previous version).
To avoid confusion - and minimize risk over future mis-interpretation this now looks like this:
```
function parseParameters(search?: string): { notFound?: boolean } {
  try {
    if (search === undefined) {
      return {};
    }
    const params = queryString.parse(search, {parseBooleans: true});
    const notFound = !!(params && params["not_found"]);
    return {
      notFound,
    };
  } catch (err) {
    console.error("/start: error parsing search params", err);
    return {};
  }
}
```

akosyakov:
If i change IDE settings to enable VS Code Desktop and then restart a workspace from the dashboard then:
- a new window is trying to open VS Code Desktop
- but old windows are still trying to open web version
Screenshot 2022-02-14 at 10 19 28
Screenshot 2022-02-14 at 10 19 24

In other aspects it worked flawlessly and I like simplification to the supervisor frontend 👍

components/supervisor/frontend/src/index.ts
```diff
@@ -142,19 +141,6 @@ const toStop = new DisposableCollection();
...
             }
-            if (instance.status.phase === 'stopped') {
-                stopped = true;
-            }
-            if (stopped && (
-                instance.status.phase === 'preparing' ||
-                instance.status.phase === 'pending' ||
-                instance.status.phase === 'creating' ||
-                instance.status.phase === 'initializing')) {
-                // reload the page if the workspace was restarted to ensure:
```
AlexTugarev, line 153: @geropl, how relevant is this removal?

akosyakov: @AlexTugarev see discussions here: https://github.com/gitpod-io/gitpod/pull/8125#discussion_r804569170

I think it is the proper way to go.

components/dashboard/src/start/StartWorkspace.tsx
csweichel: 🚀  Thanks :)

components/supervisor/frontend/src/index.ts
akosyakov: We do it on purpose to remove dependencies from supervisor and loading screen. Now a user will need to reload explicitly. I think it is fine.

AlexTugarev: Thanks, I missed that, as it was resolved. 

Nice and clean!

geropl:
> If i change IDE settings to enable VS Code Desktop and then restart a workspace from the dashboard then:
>  -  a new window is trying to open VS Code Desktop
>  - but old windows are still trying to open web version

@akosyakov this is relevant, and should also be easy to fix (we need to make sure IDEOptions are fetched on the `dontAutostart` paths as well`). Let's fix this in q small follow up.

components/dashboard/src/start/StartWorkspace.tsx
```diff
@@ -197,20 +261,35 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
...
       return;
     }
 
+    // Here we filter out updates to instances we haven't started to avoid issues with updates coming in out-of-order
+    // (e.g., multiple "stopped" events from the older instance, where we already started a fresh one after the first)
+    // Only exception is when we do the switch from the "old" to the "new" one.
+    const startedInstanceId = this.state?.startedInstanceId;
+    if (startedInstanceId !== workspaceInstance.id) {
+      // do we want to switch to "new" instance we just received an update for?
+      const switchToNewInstance = this.state.workspaceInstance?.status.phase === "stopped" && workspaceInstance.status.phase !== "stopped";
```
AlexTugarev, line 278: if previous instances is in `stopping` or never reaches `stopped`, this reads like an dead end.

assuming that we always have a single instance running per workspace, let's just focus on the status/phase of the new.

AlexTugarev (approved):
Awesome!

components/dashboard/src/start/StartWorkspace.tsx
geropl: I agree in general, but think it's not super important. And has been like this for years.

But in light of this change we should watch more sure! :pray:

PR #8139, opened 2022-02-10
[ws-daemon] Add page-cache reclaim

### LINKED ISSUE #7969: Oh, no! Something went wrong! OutOfmemory: Pod Node didn't have enough resource: memory, requested: 2576980378, used: 65006469130, capacity: 66388508672
### Bug description

Oh, no! Something went wrong!
OutOfmemory: Pod Node didn't have enough resource: memory, requested: 2576980378, used: 65006469130, capacity: 66388508672

### Steps to reproduce

Attempting to launch my existing Gitpod workspace.

### Workspace affected

https://ip69719-hellodjango-7mmiv2ne5bk.ws-eu30.gitpod.io/

### Expected behavior

Have my Gitpod workspace launch without any errors.

### Example repository

_No response_

### Anything else?

_No response_

csweichel:
## Description
This PR introduces a page cache reclaim mechanism for workspace cgroups. The kubelet computes memory pressure using the cgroup memory controller, which includes page-caches. This means that when a workspace does a lot of file IO, we're likely to overestimate the memory pressure, and evict pods when we should not - or fail the scheduling process.

With this change, we regularly check for the page cache use for each workspace on the node. If we consume more than 15% of the cgroup memory limit (1.8GiB) in caches, we'll trigger a page cache reclaim using `memory.force_reclaim`. This should keep the cgroup memory closer to actual memory consumption at the expense of file IO and kernel CPU time.

## Related Issue(s)

Fixes #7969

## How to test
```bash
while true; do sleep 1; (echo N; date +%s; cat /sys/fs/cgroup/memory/memory.stat | grep cache) >> /tmp/log.txt; done
```

in a new terminal
```
tar cvvfz $HOME/backup.tar.gz /workspace
```

once the tar is done, look at the entries in `log.txt`. They should climb for a few seconds (max 15) and drop sharply afterwards. Then they'll climb again.

## Release Notes

```release-note
Improved workspace memory-pressure eviction resilience
```


components/ws-daemon/pkg/daemon/cache_reclaim.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	s := strings.TrimSpace(string(fc))
+	if s == "max" {
+		return math.MaxUint64, nil
+	}
+
+	p, err := strconv.ParseInt(s, 10, 64)
+	if err != nil {
+		return 0, xerrors.Errorf("cannot parse memory.limit_in_bytes: %v", err)
```
sagor999, line 129: can you also add original value that we are trying to parse here.

components/ws-daemon/pkg/daemon/cache_reclaim.go
```diff
@@ -0,0 +1,157 @@
...
+	var didReclaim bool
+	if cache > uint64(float64(limit)*0.15) {
+		err := ioutil.WriteFile(filepath.Join(memCgroupPath, "memory.force_empty"), []byte("1"), 0644)
+		if err != nil {
+			return nil, xerrors.Errorf("cannot write memory.force_empty: %v", err)
+		}
+		didReclaim = true
+	}
+
+	nowCache, _ := readCache(memCgroupPath)
```
sagor999, line 106: sorry, just realized that if readCache fails then we would never log that error

Issue #8096, opened 2022-02-08
Failed to download OTS in US cluster (possibly happens for prebuilds, only)


csweichel:
### Bug description

Looking at the logs we're seeing a surprising amount of OTS download failures as part of workspace content initialisation: https://console.cloud.google.com/logs/query;query=%22cannot%20download%20OTS%22%0A;timeRange=PT24H;cursorTimestamp=2022-02-08T14:43:32Z?project=workspace-clusters

Each of those failures is likely to yield a failed workspace - at least if the repo was private.
Possible contributing factors:

(edited by Sven)
- as this seems to happen only in prebuilds (only US cluster and lots of this error messages in d_b_prebuild_workspace), it could be that for some reason the time between the OTS is created and when it gets requested is longer than 30min (the lifetime of a token). Prebuild clusters are sometimes heavily packed so maybe there is just too much time in scaling up etc.
- we attempt to download the OTS multiple times for some reason. That's most likely a bug in the initializer. Checking the server logs and/or adding metrics would help identifying this.

As part of a fix, we should introduce OTS download failure metrics and keep an eye on them.

### Steps to reproduce

Check the logs

kylos101:
[Also interesting.](https://cloudlogging.app.goo.gl/JTmV6uzLgYJh9FBN9)

geropl:
> As for load balancer question, that is probably for webapp team?

I think we should focus on unifying our DBs into one, so we don't need to expose app cluster identity to users. Especially given how long this has been in prod, and how resource constraint we as a team are (/cc @JanKoehnlein ).

As a bandaid, we could extend this timeout to 20s, which should in all but very rare cases plenty of time to include the db-sync roundtrip.

sagor999:
Reopening as this still occurs: 
https://twitter.com/srherobrine23/status/1497265903641116675?s=20&t=Taai1veK0dkGdYsmYqKnvw
WS id: "1af7b6dd-37c0-4c7b-b9e8-a94858e5f339"
Logs: 
```
"cannot download OTS"
"content init failed"
"cannot initialize workspace"
"InitWorkspace failed"
```

geropl:
@sagor999 Would it make sense to sync on this issue to speed up investigation? Would love to understand what you discovered so far.

sagor999:
@geropl I have not dig deep into this issue unfortunately and I think @csweichel have a deeper understanding of the problem.
But from what I understood, we have several possible scenarios:
1. OTS never gets created
[...]

PR #8148, opened 2022-02-11
add metrics to track success or failure of downloadOTS requests

sagor999:
## Description

Add metrics for content service to track download OTS requests, and to track how many have failed or succeeded. 

## Related Issue(s)

Related issue: #8096 

## How to test

I tried to spin up several workspaces from private repo in staging env, but did not see anything in content-service pod that would indicate it was doing anything. I might be missing something here. Need a second pair of :eyes: to confirm if local content-service pod is  supposed to call downloadOTS when opening a private repo from staging env.

## Release Notes

```release-note
Add metrics to content-service to track download OTS requests
```

### DIFF
--- a/components/content-service/cmd/run.go
+++ b/components/content-service/cmd/run.go
@@ -22,6 +22,7 @@ import (
 	"github.com/gitpod-io/gitpod/common-go/log"
 	"github.com/gitpod-io/gitpod/common-go/pprof"
 	"github.com/gitpod-io/gitpod/content-service/api"
+	"github.com/gitpod-io/gitpod/content-service/pkg/initializer"
 	"github.com/gitpod-io/gitpod/content-service/pkg/service"
 )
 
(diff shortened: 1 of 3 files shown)


roboquat:

<details>

</details>

Furisto:
> I tried to spin up several workspaces from private repo in staging env, but did not see anything in content-service pod that would indicate it was doing anything.

Did you set log level to debug with gpctl? A saw some "checking blob quota" messages whenever I opened a private repository.

sagor999:
Hm. Even with private repo it doesn't hit codepath that would go through OTS. I think for github private repos it does it without OTS? 
@csweichel can you provide more details as to which codepath would trigger OTS download? I cannot figure it out from looking at the code. But otherwise this PR should be good. :)

csweichel:
This is a good entry point: https://github.com/gitpod-io/gitpod/blob/main/components/ws-daemon/pkg/content/initializer.go#L114

Note: the content initializer which downloads the OTS runs out-of-process, wrapped in a runc container, started from ws-daemon. We recently introduced a unix pipe/FD based [error reporting mechanism here](https://github.com/gitpod-io/gitpod/blob/main/components/ws-daemon/pkg/content/initializer.go#L262) which we could extend to signal OTS download failure.

csweichel:
@sagor999 how do we continue with this one?

sagor999:
@csweichel I will need to figure out a way on how to redo those metrics. As you suggested, I can try to use error output for this and then add that metric to ws-daemon instead.

sagor999:
Will close this PR, as it needs to be redone differently.

PR #8150, opened 2022-02-11
werft/build: Add SIGTERM handling

ArthurSens:

## Description

Just adds SIGTERM handling so we can close spans before killing a pod.

The problem is that this isn't working as expected and I believe that the reason is the way we call `build.ts`

Exec-ing into the build container, I can see that the process PID 1 executes a `sh` process, which then calls the `node` that runs our script:
image

By killing a POD, the container runtime sends a SIGTERM to the PID 1 process of all containers of the pod. If PID 1 process has child processes, the responsibility of propagation of the SIGTERM to the child processes lies in the hands of said process.

The command `sh`, however, doesn't propagate SIGTERM, so our `node` process will never receive a SIGTERM

[Source](https://pracucci.com/graceful-shutdown-of-kubernetes-pods.html#:~:text=When%20a%20pod%20should%20be,see%20below%20to%20change%20it).

## Related Issue(s)

Should've fixed https://github.com/gitpod-io/ops/issues/865

## How to test

Run a job and kill it mid execution with 
```
kubectl delete pod -n werft gitpod-build-<sanitized-namespace>
```

## Release Notes

```release-note
NONE
```

## Documentation

N/A

### DIFF
--- a/.werft/build.ts
+++ b/.werft/build.ts
@@ -21,6 +21,12 @@ Tracing.initialize()
         werft = new Werft("build")
     })
     .then(() => run(context))
+    .then(() => {
+        process.on('SIGTERM', () => {
+            console.log('SIGTERM signal received.');
+            werft.endAllSpans()
+        });
+    })
     .then(() => VM.stopKubectlPortForwards())
     .then(() => werft.endAllSpans())
     .catch((err) => {
--- a/.werft/build.yaml
+++ b/.werft/build.yaml
@@ -205,6 +205,9 @@ pod:
         (cd .werft && yarn install && mv node_modules ..) | werft log slice prep
         printf '{{ toJson . }}' > context.json
 
-        npx ts-node .werft/build.ts
+        trap 'kill $(jobs -p)' SIGINT SIGTERM
+
+        npx ts-node .werft/build.ts &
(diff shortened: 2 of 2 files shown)


ArthurSens:
pinging @mads-hartmann for awareness, not sure how to proceed here

mads-hartmann:
@ArthurSens If you like we can pair on this Monday, but let me leave a bit more elaborate comment now that I'm not rushing into a meeting. ☺️ 

We do control the initial shell code, that's the `bash -c` that's defined in the Werft job [here](https://github.com/gitpod-io/gitpod/blob/main/.werft/build.yaml#L186). So I suggest adding a `trap` in that to verify at least the SIGNAL is sent to it. I suspect very much that this will just work.

Once you have verified that works, the next step is to make sure it's passed to the TypeScript code. We're using `npx ts-node` ([code](https://github.com/gitpod-io/gitpod/blob/main/.werft/build.yaml#L208)) which is just a very elaborate way to invoke `ts-node`. So one thing you could try is to invoke `./node_modules/.bin/ts-node` directly so we don't get the `npm exec` wrapper. If you call ts-node directly does that then end us as PID 1 so it gets the signal?

PR #8159, opened 2022-02-11
.werft/config: Fix typo on job trigger

ArthurSens:
## Description

After merging https://github.com/gitpod-io/gitpod/pull/8087, we've noticed that the configured job isn't running after branches are deleted.

I tried looking for docs under https://github.com/csweichel/werft but got no luck. I'm now wondering if we have a `=` missing. @csweichel, could you confirm that this change is what we need to make the jobs run on branch deletion?

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.werft/config.yaml
+++ b/.werft/config.yaml
@@ -5,4 +5,4 @@ rules:
   - or: ["trigger !== deleted"]
 - path: ".werft/delete-preview-environments/delete-preview-environment.yaml"
   matchesAll:
-  - or: ["trigger == deleted"]
+  - or: ["trigger === deleted"]


csweichel:
The current form would be correct: https://github.com/csweichel/werft/blob/baf5a1a8f11f23cb4341a4c3e5bcc1545c1c28d5/pkg/filterexpr/filterexpr.go#L17

Please file an issue on the werft repo

ArthurSens:
Alright, thanks for clarifying!

Issue opened https://github.com/csweichel/werft/issues/114

PR #8160, opened 2022-02-11
[supervisor] Improve init shutdown behaviour

csweichel:
## Description
This PR fixes the shutdown behaviour of `supervisor init`:
- `os.Notify` is very particular about the number of times it can be called. Prior to this change we called it twice which did not result in the expected signalling behaviour.
- The `supervisor run` wait group `Done()` call was not deferred, which could lead to situations where the waitgroup was never done and `init ` would not return.

## How to test
1. Run a prebuild with very fast task

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/supervisor/cmd/init.go
+++ b/components/supervisor/cmd/init.go
@@ -9,7 +9,6 @@ import (
 	"os/exec"
 	"os/signal"
 	"strings"
-	"sync"
 	"syscall"
 
 	"github.com/gitpod-io/gitpod/common-go/log"
@@ -24,51 +23,32 @@ var initCmd = &cobra.Command{
 		log.Init(ServiceName, Version, true, false)
 		// Because we're reaping with PID -1, we'll catch the child process for
 		// which we've missed the notification anyways.
-		sigs := make(chan os.Signal, 1)
-		signal.Notify(sigs, syscall.SIGCHLD)
-
-		// The reaper can be turned into a terminating reaper by writing true to this channel.
-		// When in terminating mode, the reaper will send SIGTERM to each child that gets reparented
-		// to us and is still running. We use this mechanism to send SIGTERM to a shell child processes
-		// that get reparented once their parent shell terminates during shutdown.
-		var terminating bool
-
-		reaper := func() {
-			for range sigs {
-				for {
-					// wait on the process, hence remove it from the process table
-					pid, err := unix.Wait4(-1, nil, 0, nil)
-					// if we've been interrupted, try again until we're done
-					for err == syscall.EINTR {
-						pid, err = unix.Wait4(-1, nil, 0, nil)
+		var (
+			sigInput      = make(chan os.Signal, 1)
+			sigReaper     = make(chan os.Signal, 1)
+			sigSupervisor = make(chan os.Signal, 1)
+		)
+		signal.Notify(sigInput, syscall.SIGCHLD, os.Interrupt, syscall.SIGTERM)
+		go func() {
+			for s := range sigInput {
+				switch s {
+				default:
+					sigSupervisor <- s
+					// the reaper needs all signals so that it can turn into
+					// a terminating reaper if need be.
+					fallthrough
+				case syscall.SIGCHLD:
+					// we don't want to blob the SIGINT/SIGTERM behaviour because
+					// the reaper is still busy.
+					select {
+					case sigReaper <- s:
+					default:
 					}
-					// The calling process does not have any unwaited-for children. Let's wait for a SIGCHLD notification.
-					if err == unix.ECHILD {
-						break
-					}
-					if err != nil {
-						log.WithField("pid", pid).WithError(err).Debug("cannot call waitpid() for re-parented child")
-					}
-					if !terminating {
-						continue
-					}
-					proc, err := os.FindProcess(pid)
-					if err != nil {
-						log.WithField("pid", pid).WithError(err).Debug("cannot find re-parented process")
-						continue
-					}
-					err = proc.Signal(syscall.SIGTERM)
-					if err != nil {
-						if !strings.Contains(err.Error(), "os: process already finished") {
-							log.WithField("pid", pid).WithError(err).Debug("cannot send SIGTERM to re-parented process")
-						}
-						continue
-					}
-					log.WithField("pid", pid).Debug("SIGTERM'ed reparented child process")
 				}
 			}
-		}
-		go reaper()
+		}()
+
+		go reaper(sigReaper)
 
 		supervisorPath, err := os.Executable()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


components/supervisor/cmd/init.go
```diff
@@ -85,20 +60,20 @@ var initCmd = &cobra.Command{
...
 				return
 			}
-			wg.Done()
 		}()
-		sigChan := make(chan os.Signal, 1)
-		signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
-		s := <-sigChan
-		terminating = true
+
+		s := <-sigSupervisor
```
iQQBot, line 76: Originally we just wanted to forward the signal once because it was the last exit signal, if we changed it like this it could end early here because of the SIGHOD received

csweichel: Did fix

csweichel:
Great test - we did see on core-dev that prebuilds would not shut down, but instead remain running with `supervisor init`. I.e. `supervisor run` had ended, but `init` remained running.

components/supervisor/cmd/init.go
```diff
@@ -107,3 +85,50 @@ var initCmd = &cobra.Command{
...
+	// When in terminating mode, the reaper will send SIGTERM to each child that gets reparented
+	// to us and is still running. We use this mechanism to send SIGTERM to a shell child processes
+	// that get reparented once their parent shell terminates during shutdown.
+	var terminating bool
+
+	for s := range sigs {
+		if s != syscall.SIGCHLD {
+			terminating = true
+			continue
+		}
```
iQQBot, line 100: if we only forward SIGCHLD, this part can be cleanup

components/supervisor/cmd/init.go
```diff
@@ -107,3 +85,50 @@ var initCmd = &cobra.Command{
...
+				continue
+			}
+			err = proc.Signal(syscall.SIGTERM)
+			if err != nil {
+				if !strings.Contains(err.Error(), "os: process already finished") {
+					log.WithField("pid", pid).WithError(err).Debug("cannot send SIGTERM to re-parented process")
+				}
+				continue
+			}
+			log.WithField("pid", pid).Debug("SIGTERM'ed reparented child process")
```
iQQBot, line 131: if we only forward SIGCHLD, this part can be cleanup

components/supervisor/cmd/init.go
```diff
@@ -85,20 +63,20 @@ var initCmd = &cobra.Command{
 			log.WithError(err).Error("supervisor run start error")
 			return
 		}
-		wg := sync.WaitGroup{}
+
+		var wg sync.WaitGroup
 		wg.Add(1)
 		go func() {
+			defer wg.Done()
```
iQQBot, line 70:
```suggestion
			defer func() {
				select {
				case sigSupervisor <- syscall.SIGTERM:
				default:
				}
				wg.Done()
			}()
```

> we did see on core-dev that prebuilds would not shut down, but instead remain running with supervisor init. I.e. supervisor run had ended, but init remained running.

In this case, because supervisor run exit proactive, but we're still expecting a SIGTERM signal (`s := <-sigSupervisor`), so the init process is always running

csweichel: Great point - I've updated the PR accordingly

csweichel:
Despite the good behaviour of `os.Notify` I opted to keep the new design because otherwise we might cause confusion down the road: if someone reads the documentation of `os.Notify` they might be lead to believe that our calling twice of `os.Notify` is errornous.

iQQBot (approved):
/hold

It an unblock issue, feel free to unhold it
 
At before we cannot kill `supervisor run` because it pid 1, but now we can, so if user kill the `supervisor run` the workspace will stop, but render a error `container xxxx completed; containers of a workspace pod are not supposed to do that`

https://github.com/gitpod-io/gitpod/blob/24f7b609bf79b247d72d1841282639726bcd5891/components/ws-manager/pkg/manager/status.go#L640-L641

kylos101 (approved):
LGTM

[Integration tests pass, too, both in core-dev and workspace-preview.](https://werft.gitpod-dev.com/job/gitpod-run-integration-tests-cw-fix-supervisor-init.1/raw)

They include tests for prebuild and image build, I don't see any lingering pods for either.

kylos101:
In prod now:
```shell
gitpod /workspace/template-python-flask $ ps -axf
    PID TTY      STAT   TIME COMMAND
      1 ?        Sl     0:00 supervisor run
     38 ?        S      0:00 sh /ide/bin/gitpod-code --port 23000 --host 0.0.0.0 --without-connection-token --server-data-dir /workspace/.vscode-remote
    787 ?        Rl     0:05  \_ /ide/node /ide/out/server-main.js --port 23000 --host 0.0.0.0 --without-connection-token --server-data-dir /workspace/.vscode-remote
    844 ?        Sl     0:00      \_ /ide/node /ide/out/bootstrap-fork --type=ptyHost
   1164 pts/1    Ss     0:00      |   \_ /usr/bin/bash
   1428 pts/1    R+     0:00      |       \_ ps -axf
   1090 ?        Sl     0:00      \_ /ide/node /ide/out/bootstrap-fork --type=fileWatcher
   1103 ?        Sl     0:01      \_ /ide/node /ide/out/bootstrap-fork --type=extensionHost --uriTransformerPath=/ide/out/vs/server/node/uriTransformer.js --useHostProxy=fals
   1124 ?        Sl     0:00          \_ /ide/node /ide/extensions/redhat.vscode-yaml/dist/languageserver.js --node-ipc --clientProcessId=1103
   1150 ?        Sl     0:00          \_ /ide/node /ide/extensions/json-language-features/server/dist/node/jsonServerMain --node-ipc --clientProcessId=1103
     48 pts/0    Ss     0:00 /bin/bash
    788 pts/0    S+     0:00  \_ /home/gitpod/.pyenv/versions/3.8.12/bin/python3 /home/gitpod/.pyenv/versions/3.8.12/bin/flask run
    858 pts/0    Sl+    0:00      \_ /home/gitpod/.pyenv/versions/3.8.12/bin/python3 /home/gitpod/.pyenv/versions/3.8.12/bin/flask run
```

In core-dev (so we're adding `supervisor init`:
```shell
gitpod /workspace/template-python-flask (main) $ ps -axf
    PID TTY      STAT   TIME COMMAND
      1 ?        Sl     0:00 supervisor init
     26 ?        Sl     0:00 supervisor run
     45 ?        S      0:00  \_ sh /ide/bin/gitpod-code --port 23000 --host 0.0.0.0 --without-connection-token --server-data-dir /workspace/.vscode-remote
    619 ?        Sl     0:06  |   \_ /ide/node /ide/out/server-main.js --port 23000 --host 0.0.0.0 --without-connection-token --server-data-dir /workspace/.vscode-remote
    630 ?        Sl     0:00  |       \_ /ide/node /ide/out/bootstrap-fork --type=ptyHost
    903 pts/1    Ss     0:00  |       |   \_ /usr/bin/bash
   1202 pts/1    R+     0:00  |       |       \_ ps -axf
    872 ?        Sl     0:02  |       \_ /ide/node /ide/out/bootstrap-fork --type=extensionHost --uriTransformerPath=/ide/out/vs/server/node/uriTransformer.js --useHostProxy=fa
    897 ?        Sl     0:00  |       |   \_ /ide/node /ide/extensions/redhat.vscode-yaml/dist/languageserver.js --node-ipc --clientProcessId=872
   1165 ?        Sl     0:00  |       |   \_ /ide/node /ide/extensions/json-language-features/server/dist/node/jsonServerMain --node-ipc --clientProcessId=872
   1183 ?        Sl     0:00  |       |   \_ /home/gitpod/.pyenv/versions/3.8.12/bin/python /workspace/.vscode-remote/extensions/ms-python.python-2022.0.1814523869/pythonFiles/
    883 ?        Sl     0:00  |       \_ /ide/node /ide/out/bootstrap-fork --type=fileWatcher
     60 pts/0    Ss     0:00  \_ /bin/bash
    554 pts/0    S+     0:00      \_ /usr/bin/python3 /workspace/.pip-modules/bin/flask run
    605 pts/0    Sl+    0:00          \_ /usr/bin/python3 /workspace/.pip-modules/bin/flask run
```

Killing the PID for `supervisor init` or `supervisor run`result in the same behavior, the workspace terminates with:

```
Oh, no! Something went wrong!
container workspace completed; containers of a workspace pod are not supposed to do that
```

Issue #8020, opened 2022-02-03
Investigate why using fuse overlayfs causes /proc not being mountable by docker in workspace


Furisto:
### Bug description

When fuse overlayfs is configured it is not possible to use docker in a workspace as mounting /proc for the container fails.

```
gitpod /workspace/template-python-flask $ docker run hello-world
Unable to find image 'hello-world:latest' locally
latest: Pulling from library/hello-world
2db29710123e: Pull complete 
Digest: sha256:975f4b14f326b05db86e16de00144f9c12257553bba9484fed41f9b6f2257800
Status: Downloaded newer image for hello-world:latest
docker: Error response from daemon: OCI runtime create failed: container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: rootfs_linux.go:76: mounting "proc" to rootfs at "/proc" caused: mount through procfd: operation not permitted: unknown.
ERRO[0003] error waiting for container: context canceled 
```

### Steps to reproduce

- Switch to fuse overlayfs
- Start a workspace
- docker run hello-world

### Expected behavior

Docker should run container successfully

### Anything else?

Here are a few thing to speed things up:
- You can use `kubectl edit cm ds ws-daemon` to edit the config map and switch quickly between shiftfs and fuse. Do not forget to restart the daemon afterward.
- nsinsider can be build with `CGO_ENABLED=0 go build -o nsinsider -a -ldflags '-extldflags "-static"' .` (if in nsinsider directory). This creates a static binary. If build without these flags it will be dynamically linked and fail to execute.
- You can copy nsinsider into ws-daemon with kubectl cp ./nsinsider ws-daemon-23u8:/app/ which saves us the redeploy
- When the error happens this is logged by ws-daemon:
```
"args":["/app/nsinsider","move-mount","--target","/tmp/supervisor2376303327/workspace/.docker-root/overlay2/d01aa7947752ee4012406c013b2bd87a7c19fad98c06dd493e40a558aeea313a/merged/proc","--pipe-fd","3"],"instanceId":"1431dff6-836b-4c87-ad40-792369174e59","level":"fatal","message":"no such file or directory","serviceContext":{"service":"nsinsider","version":""},"severity":"CRITICAL","time":"2022-02-03T14:35:06Z"}
[...]

PR #8181, opened 2022-02-13
Ensure docker in workspace works with fuse

Furisto:
## Description
This fixes an issue where trying to run a docker container within a workspace would fail when fuse-overlayfs was used as shift method for the rootfs. Chroot only affects the calling process while pivot_root affects the whole mount namespace. @csweichel Do you know why chroot was used for fuse and pivot_root for shiftfs? Trying to understand if this would have other side effects.

## Related Issue(s)
#8020

## How to test
- Ensure that fuse is selected as shift method (already done for the preview environment)
- Open workspace
- Try to run docker container in workspace 

Side note: I am currently not able to get logs from ws-daemon or exec into it in my preview environment. I do not believe this has anything to do with my changes and should not affect anything but a second set of eyes on this would be good.

## Release Notes

```release-note
Fixed a bug where docker in workspaces could not be used when overlay-fusefs was used as shift method
```
cc: @utam0k

### DIFF
--- a/components/workspacekit/cmd/rings.go
+++ b/components/workspacekit/cmd/rings.go
@@ -398,7 +398,7 @@ var ring1Cmd = &cobra.Command{
 		procLoc := filepath.Join(ring2Root, "proc")
 		err = os.MkdirAll(procLoc, 0755)
 		if err != nil {
-			log.WithError(err).Error("cannot mount proc")
+			log.WithError(err).Error("cannot create directory for mounting proc")
 			return
 		}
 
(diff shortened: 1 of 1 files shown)


utam0k:
Awesome👍 was there any man or documentation on the chroot phenomenon? Or is it a conclusion you understand from the phenomenon?

Furisto:
> Awesome was there any man or documentation on the chroot phenomenon? Or is it a conclusion you understand from the phenomenon?

I already knew about the different behavior. The man pages for chroot and pivot_root explain the different behavior, but do not specifically contrast it with each other, so you need to read both to notice the difference.

utam0k:
> > Awesome was there any man or documentation on the chroot phenomenon? Or is it a conclusion you understand from the phenomenon?
> 
> I already knew about the different behavior. The man pages for chroot and pivot_root explain the different behavior, but do not specifically contrast it with each other, so you need to have read both to notice the difference.

I remember now! Thank you very much.

PR #8189, opened 2022-02-14
[content-service] Don't let composite initializer swallow errors

### LINKED ISSUE #8156: [content-initializer] Composite content initializer shadows errors
### Bug description

When we use a composite content initializer (e.g. when server produced a `.gitpod.yml`) that shadows the errors produced by other content initializer, e.g. git. This in turn leads to hard-to-debug content init failures.

### Steps to reproduce

Open a repo for an installation we cannot clone from (yet parse the context), which does not contain a .gitpod.yml.

### Workspace affected

_No response_

### Expected behavior

The content init should fail as a whole.

### Example repository

_No response_

### Anything else?

_No response_

csweichel:
## Description
This PR makes a composite content initializer fail if any of its initializers fail. This way we don't swallow errors when a repo doesn't have a `.gitpod.yml`.

## Related Issue(s)

Fixes #8156

## How to test
This one is tough to test, because you need to make a Git content initializer fail - something we work hard to avoid :) 
One way would be to have a broken GitLab installation which reports an invalid clone URL.
Otherwise, we'll have to rely on the unit test.

## Release Notes

```release-note
Composite content initializer now correctly report errors.
```

### DIFF
--- a/components/content-service/pkg/initializer/initializer.go
+++ b/components/content-service/pkg/initializer/initializer.go
@@ -57,15 +57,16 @@ func (e *EmptyInitializer) Run(ctx context.Context, mappings []archive.IDMapping
 }
 
 // CompositeInitializer does nothing
-type CompositeInitializer struct {
-	Initializer []Initializer
-}
+type CompositeInitializer []Initializer
 
 // Run calls run on all child initializers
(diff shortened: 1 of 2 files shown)


utam0k (approved):
LGTM

PR #8199, opened 2022-02-14
Revert "Create an index for contextURL column"

geropl:
## Description
The change would break production data as seen on staging. We need reconsider the max. length want to handle based on prod DB data.
Reverting for now to unblock prod deployments.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


JanKoehnlein (approved):
LGTM

PR #8211, opened 2022-02-15
Add Replicated as a Gitpod license evaluator

MrSimonEmms:
## Description

> For the purpose of this, "enterprise" means any user that has a paid-for license

[Replicated](https://www.replicated.com/) is how we will now distribute Gitpod to enterprise teams. This adds Replicated as a licensor and is used if the secret has `type = replicated` (which sets the envvar `GITPOD_LICENSE_TYPE=replicated` in the `server`).

This PR also renames the `Evaluator` as `GitpodEvaluator` and introduces an `Evaluator` interface, which the `GitpodEvaluator` and `ReplicatedEvaluator` implements. The GitpodEvaluator will remain in-place for existing customers (eventually, will migrate to Replicated) and also for SaaS.

Like the Gitpod license, the Replicated license is only validated at runtime. If the user wishes to update their license, they would have to click "sync license" in their Kots portal which would tell them to redeploy the application - this would create a new `server` pod which would verify the new license at runtime.

Unlike the Gitpod license, the Replicated license is retrieved from a local service and has no (obvious) way of verifying the data is from a legitimate source, rather than a mocked service. For that reason, the [URL is hardcoded as per the docs](https://kots.io/vendor/entitlements/creating-entitlements/#querying-entitlements-at-runtime) and offers no way of configuring this.

The license has three custom entitlements - `domain`, `levelId` and `seats`. An entitlement is custom data that's attached to the license. These are used to provide the data that we use for validation.

### Future improvements

Talk with Replicated and establish a way of verifying that the data is valid and from an approved source. Replicated provides a signature which has an `innerSignature` which has a public key and signature but this verification is entirely undocumented.

We may also want to verify community licenses slightly differently, eg without the domain so we can just have a single community license. **NB** it may be tricky for a user to go from community to paid if we have a single license - this will need raising with Replicated

### Resources
 - [License API docs](https://help.replicated.com/api/integration-api/license-api/)
 - [License entitlements](https://kots.io/vendor/entitlements/creating-entitlements/)

## Related Issue(s)

Fixes https://github.com/gitpod-io/replicated/issues/1

## Release Notes

```release-note
Add Replicated as a Gitpod license evaluator
```

### DIFF
--- a/components/licensor/ee/cmd/validate.go
+++ b/components/licensor/ee/cmd/validate.go
@@ -22,18 +22,28 @@ var validateCmd = &cobra.Command{
 	Short: "Validates a license - reads from stdin if no argument is provided",
 	Args:  cobra.MaximumNArgs(1),
 	RunE: func(cmd *cobra.Command, args []string) (err error) {
-		var lic []byte
-		if len(args) == 0 {
-			lic, err = io.ReadAll(os.Stdin)
-			if err != nil {
-				return err
+		domain, _ := cmd.Flags().GetString("domain")
+		licensorType, _ := cmd.Flags().GetString("licensor")
+
+		var e licensor.Evaluator
+		switch licensorType {
+		case string(licensor.LicenseTypeReplicated):
+			e = licensor.NewReplicatedEvaluator(domain)
+			break
+		default:
+			var lic []byte
+			if len(args) == 0 {
+				lic, err = io.ReadAll(os.Stdin)
+				if err != nil {
+					return err
+				}
+			} else {
+				lic = []byte(args[0])
 			}
-		} else {
-			lic = []byte(args[0])
+
+			e = licensor.NewGitpodEvaluator(lic, domain)
 		}
 
-		domain, _ := cmd.Flags().GetString("domain")
-		e := licensor.NewEvaluator(lic, domain)
 		if msg, valid := e.Validate(); !valid {
 			return xerrors.Errorf(msg)
 		}
@@ -47,4 +57,5 @@ var validateCmd = &cobra.Command{
 func init() {
 	rootCmd.AddCommand(validateCmd)
 	validateCmd.Flags().String("domain", "", "domain to evaluate the license against")
+	validateCmd.Flags().String("licensor", "gitpod", "licensor to use")
 }
--- a/components/licensor/ee/pkg/licensor/licensor_test.go
+++ b/components/licensor/ee/pkg/licensor/licensor_test.go
@@ -5,41 +5,135 @@
 package licensor
 
 import (
+	"bytes"
 	"crypto/rand"
 	"crypto/rsa"
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
 	"testing"
 	"time"
 )
 
 const (
-	seats  = 5
-	domain = "foobar.com"
-	someID = "730d5134-768c-4a05-b7cd-ecf3757cada9"
+	seats                = 5
+	domain               = "foobar.com"
+	someID               = "730d5134-768c-4a05-b7cd-ecf3757cada9"
+	replicatedLicenseUrl = "http://kotsadm:3000/license/v1/license"
 )
 
 type licenseTest struct {
-	Name     string
-	License  *LicensePayload
-	Validate func(t *testing.T, eval *Evaluator)
+	Name         string
+	License      *LicensePayload
+	Validate     func(t *testing.T, eval Evaluator)
+	Type         LicenseType
+	NeverExpires bool
+}
+
+// roundTripFunc .
+type roundTripFunc func(req *http.Request) *http.Response
+
+// roundTrip .
+func (f roundTripFunc) RoundTrip(req *http.Request) (*http.Response, error) {
+	return f(req), nil
+}
+
+// newTestClient returns *http.Client with Transport replaced to avoid making real calls
+func newTestClient(fn roundTripFunc) *http.Client {
+	return &http.Client{
+		Transport: roundTripFunc(fn),
+	}
 }
 
 func (test *licenseTest) Run(t *testing.T) {
 	t.Run(test.Name, func(t *testing.T) {
-		var eval *Evaluator
-		if test.License == nil {
-			eval = NewEvaluator(nil, "")
-		} else {
-			priv, err := rsa.GenerateKey(rand.Reader, 2048)
-			if err != nil {
-				t.Fatalf("cannot generate key: %q", err)
-			}
-			publicKeys = []*rsa.PublicKey{&priv.PublicKey}
-			lic, err := Sign(*test.License, priv)
-			if err != nil {
-				t.Fatalf("cannot sign license: %q", err)
+		var eval Evaluator
+		if test.Type == LicenseTypeGitpod {
+			if test.NeverExpires {
+				t.Fatal("gitpod licenses must have an expiry date")
 			}
 
-			eval = NewEvaluator(lic, domain)
+			if test.License == nil {
+				eval = NewGitpodEvaluator(nil, "")
+			} else {
+				priv, err := rsa.GenerateKey(rand.Reader, 2048)
+				if err != nil {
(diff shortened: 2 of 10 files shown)


roboquat:

<details>

</details>

components/installer/pkg/config/v1/config.go
```diff
@@ -227,6 +227,13 @@ type Workspace struct {
 	Templates *WorkspaceTemplates `json:"templates,omitempty"`
 }
 
+type LicensorType string
+
+const (
+	LicensorTypeLegacy     LicensorType = "legacy"
```
csweichel, line 233: let's not name this `legacy` as it implies that we don't want to use this anymore in all cases, and that it's likely to go away.

That's not true - it's just a different means. How about `gitpod-signed`?

components/licensor/ee/pkg/licensor/replicated_evaluator.go
```diff
@@ -0,0 +1,25 @@
+package licensor
```
csweichel, line 1: filename nit:
- `replicated_evaluator.go` -> `replicated.go`
- `gitpod_evaluator.go` -> `gitpod.go`

Idiomatic Go style favours shorter, single-word filenames.

components/licensor/typescript/ee/main.go
```diff
@@ -7,22 +7,27 @@ package main
...
+	instances map[int]licensor.Evaluator = make(map[int]licensor.Evaluator)
+	nextID    int                       = 1
 )
 
 // Init initializes the global license evaluator from an environment variable
 //export Init
 func Init(key *C.char, domain *C.char) (id int) {
 	id = nextID
-	instances[id] = licensor.NewEvaluator([]byte(C.GoString(key)), C.GoString(domain))
+	if os.Getenv("use_replicated") == "true" {
```
csweichel, line 26: Considering this env var will be part of `server` (and all other services which might integrate with `licensor` down the road), let's think of a more apt env var name :)

E.g. `GITPOD_LICENSE_TYPE` which can be `replicated` or anything else in which case we'd fall back to the `GitpodEvaluator`

corneliusludmann:
Quick test with Replicated leads to 

> Error: Maximum number of users permitted by the license exceeded

after registering GitHub as auth provider.

MrSimonEmms:
@corneliusludmann do you want to sync on this? When I've tested it, it's downgraded to free if the URL is wrong. The only caveat is that, if you've got more than 10 users in your installation, the default license has 10 users

corneliusludmann:
@MrSimonEmms Have you tested it with a license that does not expire?

I see in the server logs:

> "message":"invalid license: falling back to default","payload":{"domain":"gitpod-698.testing.gitpod-self-hosted.com","msg":"not valid anymore"}

MrSimonEmms:
@corneliusludmann yes, that'll be it. Let me investigate and see what I need to do

corneliusludmann:
Quick test with a valid license looks good.

However, I'm pretty unhappy that an invalid license would prevent using Gitpod at all. In my opinion, with an invalid license, we should also fallback like there is no license (not only for the Replicated license but also for the Gitpod license but the latter is not so important at the moment), which means basically using the default built-in license when the license is invalid:

https://github.com/gitpod-io/gitpod/blob/558d979446473710c58b20bf3036adffb155d61e/components/licensor/ee/pkg/licensor/licensor.go#L118-L123

I think it's odd that an installation with an expired license or with a license for another domain is in a worse position than an installation with no license at all.

That means, in my opinion, a test like this one:

```
{"Replicated: invalid license within default seats", 50, 5, true, false, true, LicenseTypeReplicated, false},
```

added [here](https://github.com/gitpod-io/gitpod/blob/558d979446473710c58b20bf3036adffb155d61e/components/licensor/ee/pkg/licensor/licensor_test.go#L179) should succeed and not fail (we don't have this test with a wrong domain but that should work as well).

And there is also an actual need for such a behavior: When we want to publish a generic community license, we need to ignore the domain at least. Because this license needs to work with all domains. However, for this use case, we should probably think about making the domain value optional.

MrSimonEmms:
So, from a Replicated point-of-view, we can't get through the license gate in their UI with an invalid/expired license. I am fully in agreement that this is a retrograde step and it's one of the things I've scheduled to discuss with them in our weekly meeting.

With regards to the community license, I'm forming the opinion that we're not going to be able to have a single license for all community members. See the "future improvements" section of this PR and #8329 for details. Again, this is scheduled for discussion with Replicated.

If we've gotten through the initial Replicated license gate in their UI, an expired/invalid license gives the default license. Ticket #8328 exists to make this default license match what we advertise (either free or community, dependent upon the number of seats in use)

I've added that test in you put in

components/licensor/ee/cmd/validate.go
```diff
@@ -10,6 +10,9 @@ import (
 	"io"
 	"os"
 
+	// "io"
+	// "os"
```
corneliusludmann, line 14: nit: could this be removed?

corneliusludmann (dismissed):
I tested this heavily and it works pretty well. I think that's a very good first iteration that we can ship. However, I would like to add a few issues for follow-up pull requests. Stay tuned … :laughing:

components/licensor/ee/cmd/validate.go
MrSimonEmms: Oh yes. Sorry, didn't spot that one

MrSimonEmms:
Thanks @corneliusludmann. I'm definitely seeing this as the first iteration of many in the licensing epic

corneliusludmann (approved):

corneliusludmann:
@MrSimonEmms: You haven't touched the helm charts but changed the server deployment in the installer, right? Would that break SaaS deployment for team WebApp?

MrSimonEmms:
@corneliusludmann I don't believe so. It defaults to the Gitpod evaluator and [only uses the Replicated one if the secret tells it to](https://github.com/gitpod-io/gitpod/blob/896fe39ccf456e132a4c7dba968c18f7394e6fc8/install/installer/pkg/components/server/deployment.go#L172)

geropl:
Starting to review now...

components/licensor/ee/pkg/licensor/licensor_test.go
```diff
@@ -54,16 +148,36 @@ func TestSeats(t *testing.T) {
...
-		{"unlimited seats", 0, 1000, true, false, false},
-		{"within limited seats", 50, 40, true, false, false},
-		{"within limited seats (edge)", 50, 50, true, false, false},
-		{"beyond limited seats", 50, 150, false, false, false},
-		{"beyond limited seats (edge)", 50, 51, false, false, false},
-		{"invalid license", 50, 50, false, false, true},
-		{"within default license seats", 0, 7, true, true, false},
-		{"within default license seats (edge)", 0, 10, true, true, false},
-		{"beyond default license seats", 0, 11, false, true, false},
+		{"Gitpod: unlimited seats", 0, 1000, true, false, false, LicenseTypeGitpod, false},
```
geropl, line 154: :heart:

install/installer/pkg/components/server/deployment.go
```diff
@@ -53,7 +54,6 @@ func deployment(ctx *common.RenderContext) ([]runtime.Object, error) {
 			Name:      "gitpod-license-key",
 			MountPath: licenseFilePath,
 			SubPath:   "license",
-			ReadOnly:  true,
```
geropl, line 56: Why is this necessary? :thinking: Maybe it make sense to make that explicit here, and add a comment that states why that is required?

MrSimonEmms: Arguably not, but it feels prudent to make readonly values explicitly readonly to me

geropl (approved):
Code LG - and nicely clean - TM, thx!

/hold

In case you might want to add a comment

MrSimonEmms:
Thanks @geropl - much appreciated you checking it out

PR #7757, opened 2022-01-21
Streamline code formatting

trumbitta:
## Why

- I'm about to start working on fixing the linting problems the dashboard app logs, but ESLint and Prettier have some common gray area and it's better to start with consistently formatted code to have an easier time in handling this common area
- Basically why Prettier exists: having an automated coding style which is readable, consistent, and not particularly prone to be argued upon: again, @loujaybee this improves DevX

## What

- Adds Prettier support and configurations
- Format on save
- Batch format everything Prettier can format to obtain a good starting point
- Monorepo components / folders are batch-formatted in separate commits in case you prefer to accept this only for some of them

## TODO

- [ ] Get feedback, is this useful? (I think it's fundamental!)
- [ ] Get feedback, do we want it everywhere is supported? (again, see the commits)
- [ ] Add notes to `README` (which ones depends on the answer to the question just above) about this
- [ ] Add a pre-commit hook, just in case. But this seems better done in a separate PR after / with ESLint because Prettier suggests doing it with https://github.com/okonet/lint-staged#configuration and `lint-staged` suggests stabilizing ESLint before configuring it
- [ ] Get someone to try it with a Jetbrains IDE

## How to test

Use, build, test Gitpod to see if this broke something (it shouldn't have, since Prettier only reformats code)

Issue #3841, opened 2021-04-08
[dashboard] Lint warnings


aledbf:
### Bug description
```console
Compiled with warnings.

src/App.tsx
  Line 57:8:   React Hook useEffect has a missing dependency: 'setUser'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
  Line 67:13:  img elements must have an alt prop, either with meaningful text, or an empty string for decorative images   jsx-a11y/alt-text

src/Login.tsx
  Line 26:9:    img elements must have an alt prop, either with meaningful text, or an empty string for decorative images   jsx-a11y/alt-text
  Line 71:8:    React Hook useEffect has a missing dependency: 'setUser'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
  Line 86:25:   img elements must have an alt prop, either with meaningful text, or an empty string for decorative images   jsx-a11y/alt-text
  Line 112:29:  img elements must have an alt prop, either with meaningful text, or an empty string for decorative images   jsx-a11y/alt-text
  Line 122:41:  img elements must have an alt prop, either with meaningful text, or an empty string for decorative images   jsx-a11y/alt-text

src/admin/UserSearch.tsx
  Line 41:8:  React Hook useEffect has a missing dependency: 'searchResult.rows'. Either include it or remove the dependency array  react-hooks/exhaustive-deps

src/admin/WorkspacesSearch.tsx
  Line 51:8:  React Hook useEffect has a missing dependency: 'searchResult.rows'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
  Line 57:8:  React Hook useEffect has a missing dependency: 'search'. Either include it or remove the dependency array             react-hooks/exhaustive-deps

src/components/Menu.tsx
  Line 49:21:  img elements must have an alt prop, either with meaningful text, or an empty string for decorative images  jsx-a11y/alt-text

src/components/Modal.tsx
  Line 44:8:  React Hook useEffect has a missing dependency: 'handler'. Either include it or remove the dependency array  react-hooks/exhaustive-deps

src/settings/EnvironmentVariables.tsx
  Line 29:45:   Unexpected whitespace after spread property operator  rest-spread-spacing
  Line 137:38:  Unnecessary escape character: \*                      no-useless-escape

src/settings/Integrations.tsx
[...]

PR #8220, opened 2022-02-15
build: add Prettier support and configurations

trumbitta:
## Description

Opening this new one since https://github.com/gitpod-io/gitpod/pull/7757 was too much.

## Why

- I'm about to start working on fixing the linting problems the dashboard app logs #3841, but ESLint and Prettier have some common gray area and it's better to start with consistently formatted code to have an easier time in handling this common area
- Basically why Prettier exists: having an automated coding style which is readable, consistent, and not particularly prone to be argued upon: again, @loujaybee this improves DevX
- I just took a look at the codebase (not just at the dashboard) and I found inconsistent formatting (file A with double quotes, file B with single quotes, file C with **both**, same for indentation etc).

**This makes contributing harder than it should be. It's either contributors having to format their code without automated formatters, or having PRs full of irrelevant style changes with the actual feature or fix hidden somewhere.**

But don't take my word for it: https://prettier.io/docs/en/why-prettier.html

## What

- Adds Prettier support and configurations
- Activates format on save for the VSCode workspace
- Prettier is not invasive, it doesn't change how code works (that's ESLint), just how it looks, it doesn't break anything (that's potentially again ESLint)

## Disclaimer

This is the minimum viable contribution to get things going in the right direction and over time get to a point where all the code is formatted the same way and code formatting is automated.

Its biggest con is that after you merge this, for months, every single PR of code written with VSCode or another IDE with active Prettier support will be full of irrelevant - **but much needed for the reasons stated above** - code formatting changes.

Formatting all at once (again, this is not linting, it doesn't change how the code works and it's supposed to be safe), instead, gives you a solid starting point and doesn't impact future PRs.

## Related Issue(s)

Fixes #

## How to test

Edit a TS (for example) file in VSCode and save it. You should see the formatting change, unless your own user settings specifically turn off format on save and Prettier support.

## Release Notes

```release-note
Make it easier to contribute to Gitpod by automating code formatting via Prettier
```

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -43,7 +43,8 @@ tasks:
 vscode:
   extensions:
     - bradlc.vscode-tailwindcss
-    - EditorConfig.EditorConfig
+    - editorconfig.editorconfig
+    - esbenp.prettier-vscode
     - golang.go
     - hashicorp.terraform
     - ms-azuretools.vscode-docker
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,9 @@
+# Add files here to ignore them from prettier formatting
+
+build/
+dist/
+templates/
+
+# YAML is supported, but let's play it safe for now
+*.yaml
+*.yml
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,7 @@
+{
+  "singleQuote": true,
+  "semi": true,
+  "trailingComma": "all",
+  "printWidth": 120,
+  "tabWidth": 2
+}
--- a/components/dashboard/.eslintrc.js
+++ b/components/dashboard/.eslintrc.js
@@ -5,9 +5,9 @@
  */
 
 module.exports = {
-    root: true,
-    extends: ['react-app'],
-    rules: {
-        "import/no-anonymous-default-export": "off",
-    }
-}
+  root: true,
+  extends: ['react-app', 'prettier'],
+  rules: {
+    'import/no-anonymous-default-export': 'off',
+  },
+};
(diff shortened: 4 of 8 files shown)


geropl:
IMO the best path to migrate an existing code base to:
 - decide on linting rules: :white_check_mark: 
 - auto-format enabled (like this PR) :heavy_check_mark: 
 - make it **opt-in** on a per-file basis: this way the migration is done over time, and on the other hand guaranteed to stay stable. I really really want to avoid this to interfere with other work. [Prettier seems to support this](https://prettier.io/docs/en/options.html#require-pragma). :white_check_mark: 

What do others think?

trumbitta:
> IMO the best path to migrate an existing code base to:
> 
> * decide on linting rules: ✅
> * auto-format enabled (like this PR) ✔️
> * make it **opt-in** on a per-file basis: this way the migration is done over time, and on the other hand guaranteed to stay stable. I really really want to avoid this to interfere with other work. [Prettier seems to support this](https://prettier.io/docs/en/options.html#require-pragma). ✅
> 
> What do others think?

While we wait for the other Gitpodders to chime in, I'd like to add that I already went this route in the past with a project of comparable size, and it worked in the end but with the cons I wrote about in the description of this PR.

At the end of the day, the repository is yours and I'm just happy to help so the decision is all yours and I'm here for y'all 😊

easyCZ:
I've landed https://github.com/gitpod-io/gitpod/pull/8655 which adds a minimal prettier config and setups pre-commit to run on `git commit` as a hook.

trumbitta:
oh, ok...

trumbitta:
I guess this means I can finally start working on https://github.com/gitpod-io/gitpod/issues/3841 if that's still available? @JanKoehnlein @geropl @easyCZ

PR #8225, opened 2022-02-15
werft: Cron that deleted namespaces without related branches

ArthurSens:
## Description

We're introducing a werft job that deletes previews as soon as its related branch is deleted ([Ref](https://github.com/gitpod-io/ops/issues/981)).

As an extra safety measure in case the previews job fails, this PR deletes all previews that don't have a related branch.

This job will be run as a Werft Cron. 

## Related Issue(s)

Related to https://github.com/gitpod-io/ops/issues/1151

## How to test

```
werft run github -f -j .werft/delete-preview-environments/delete-preview-environments-cron.yaml
```

## Release Notes

```release-note
NONE
```


.werft/delete-preview-environment-cron.yaml
```diff
@@ -0,0 +1,54 @@
...
+        sleep 1
+        set -Eeuo pipefail
+
+        sudo chown -R gitpod:gitpod /workspace
+
+        (cd .werft && yarn install && mv node_modules ..) | werft log slice prep
+        printf '{{ toJson . }}' > context.json
+
+        # We're merging with the script commented so we can safely test prior to fully commit to it
+        npx ts-node .werft/delete-preview-environments-cron.ts
```
mads-hartmann, line 54: There's a typo here I'm afraid ☺️ The file name is .werft/delete-preview-environment-cron.ts (but I agree that it would be nicer if the file was named `delete-preview-environments-cron.ts` so maybe just rename the file ☺️

ArthurSens: Oh my gosh this is getting more frequent than I'd like to admit 🤦

ArthurSens:
A successful run can be seen here: https://werft.gitpod-dev.com/job/gitpod-delete-preview-environments-cron-as-cron-delete-pre.14. (The actual deletion command is commented)

Please pay attention to the "Mapping previews" log slice, where existing previews are associated with their respective branches

wulfthimm (approved):

.werft/delete-preview-environments/delete-preview-environments-cron.ts
```diff
@@ -0,0 +1,97 @@
...
+        previews = listAllPreviewNamespaces({});
+    } catch (err) {
+        werft.fail("Fetching previews", err)
+    }
+    werft.done("Fetching previews");
+
+    werft.phase("Mapping previews => branches")
+    var previewBranchMap = new Map<string, string>()
+    previews.forEach(preview => {
+        branches.forEach(branch => {
```
mads-hartmann, line 52: It not too big a deal as the we're not dealing with a ton of items in the list, but this means you'll be iterating over all the branches for every preview environment. What you could do instead is to create a Set and use the `has` method as checking for set membership is supposed to be fast ☺️ Unfortunately ECMAScript doesn't include any of the Set operations you'd expect (union, difference, etc.) so you'll still have to use filter.

But these nested forEach's and the Map could be replaced with the following (I believe)

```ts
const expectedPreviewEnvironmentNamespaces = new Set(['staging-as-tracing-sigterm' ]); // would be something like new Set([...getAllBranches().map(branch => convertToNamespace(branch))])
const actualPreviewEnvironmentNamespaces = ['staging-as-tracing-sigterm', 'staging-ak-jb-backend-debug-flow']; // would be something like listAllPreviewNamespaces({})

const nameSpacesToDelete = actualPreviewEnvironmentNamespaces.filter(ns => !expectedPreviewEnvironmentNamespaces.has(ns))
```

.werft/delete-preview-environments/delete-preview-environments-cron.ts
```diff
@@ -0,0 +1,97 @@
...
+        }
+    });
+    previewBranchMap.forEach((branch: string, preview: string) => {
+        werft.log("Mapping previews", `Preview: ${preview}, branch: ${branch}`)
+    });
+    werft.done("Mapping previews => branches")
+
+    werft.phase("deleting previews")
+    try {
+        previewBranchMap.forEach((branch: string, preview: string) => {
```
mads-hartmann, line 69: If you follow the suggestion above this would then just be

```ts
nameSpacesToDelete.forEach((namespace) => wipePreviewEnvironmentAndNamespace(helmInstallName, preview, { slice: `Deleting preview ${preview}` }))
```

mads-hartmann (commented):
Left some suggestions for making the code shorter, feel free to ignore ☺️

ArthurSens:
Thanks for the review @mads-hartmann, I wanted to use sets from the beginning but was having a hard time understanding to use them in typescript 😅 

A successful run can be seen [here](https://werft.gitpod-dev.com/job/gitpod-delete-preview-environments-cron-as-cron-delete-pre.20/raw)

.werft/delete-preview-environments/delete-preview-environments-cron.yaml
```diff
@@ -0,0 +1,53 @@
...
+      - -c
+      - |
+        sleep 1
+        set -Eeuo pipefail
+
+        sudo chown -R gitpod:gitpod /workspace
+
+        (cd .werft && yarn install && mv node_modules ..) | werft log slice prep
+
+        # We're merging with the script commented so we can safely test prior to fully commit to it
```
mads-hartmann, line 52: This comment can be deleted I think ☺️

mads-hartmann (commented):
Awesome, nice and clean. Just one tiny comment to be deleted and I'll approve ☺️

.werft/delete-preview-environments/delete-preview-environments-cron.yaml
ArthurSens: done!

mads-hartmann (approved):

PR #8229, opened 2022-02-15
[server] asyncForEach: avoid long sync loops by using setImmediate

geropl:
## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


geropl:
@AlexTugarev not sure this helps, but maybe worth a try. Are you aware of any other long-ish lists we iterate often? :thinking:

geropl:
After looking at CPU profiles again this is won't help with our problem.

Issue #8065, opened 2022-02-07
Split build.ts into multiple files


mads-hartmann:
The “build” job - `build.yaml` and `build.ts` - has grown to a point where it has become hard to maintain.

As a small step towards making it easier to main we want to split the responsibilities of build.ts into separate files.

The split could be made like so 

```
.werft
    jobs
        build
            build-and-publish.ts
            deploy-to-preview-environment.ts
    build.yaml
    build.ts
```

Part of the internal RFC ["Decomposing the build job"](https://www.notion.so/gitpod/Decomposing-the-build-job-2327d3cd8dfa4520b0ad994ab1b49dfd)

PR #8240, opened 2022-02-16
Refactor build.ts into multiple files

mads-hartmann:
## Description

This PR splits `build.ts` into smaller files so that `build.ts` becomes a fairly small file where the overall sequence and structure of the build and be easily understood.

The structure now follows this pattern

```
.werft
    jobs
        build
            build-and-publish.ts
            deploy-to-preview-environment.ts
            ... the other "phases" of the build has their own files
    build.yaml
    build.ts
```

and `build.ts` primarily looks like this

```typescript
async function run(context: any) {
    const config = jobConfig(werft, context)

    await validateChanges(werft)
    await prepare(werft)
    await buildAndPublish(werft, config)
    await coverage(werft, config)

    if (config.noPreview) {
        werft.phase("deploy", "not deploying");
        console.log("no-preview or publish-release is set");
        return
    }

    await deployToPreviewEnvironment(werft, config)
    await triggerIntegrationTests(werft, config, context.Owner)
}

```

I have tried to clean up as little as possible in this PR but I have made the following changes

1. I have consolidated the "configuration" of the job into a JobConfig type. The idea is that we compute this in the beginning of the job based on Werft attributes and other details, and then just pass that around. While that does mean we're passing a rather big object around to most functions, it means at least we do our parsing and interpretation of the "configuration" at a single place.
2. I have not tried to merge it with DeployConfig so right now the two objects share a lot of the same fields. I wanted to keep this PR as small as possible.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/8065

## How to test

- [x] Preview environment using Installer should work ([job](https://werft.gitpod-dev.com/job/gitpod-build-mads-split-build.0/results), tested a new workspace could start using the `https://github.com/gitpod-io/template-typescript-node`)
- [x] Preview environment using Helm (and observability) should work ([job failed](https://werft.gitpod-dev.com/job/gitpod-build-mads-split-build.1/results) related to observability but workspaces worked - tested workspace using same repository as above - and Prometheus / Grafana worked too) 
    ```sh
     werft job run github -a with-helm=true -a with-observability=true -a with-clean-slate-deployment=true
    ```
- [x] Preview environment using VM should work ([job](https://werft.gitpod-dev.com/job/gitpod-build-mads-split-build.2/results), tested workspace using same repository as above. Prometheus / Grafana worked, SSH worked, kubectl access worked)
    ```sh
     werft job run github -a with-vm=true
    ```

## Release Notes

```release-note
NONE
```

## Documentation

N/A


ArthurSens (approved):
Awesome work! It opens so many doors for making our CI a lot more resilient💪 and faster🚀!

What do you think of scheduling a couple of issues? 👇 
* Split `.werft/build.ts` into several different jobs, while changing `.werft/config.yaml` to run multiple jobs on every push. We can decide on what can be run separately while addressing this issue
* Split deployment code into 3 files with different functions
  * Deploy with Helm
  * Deploy with installer
  * Deploy to VM (which reuses the installer function)

Later on, we can start working on splitting build, deployment, and integration tests for each team and making them owners of those parts

PR #8243, opened 2022-02-16
ws-daemon: Enable cgroup namespace.

utam0k:
## Description
It is available from linux kernel 4.6, so I don't think it will be a problem.
https://github.com/gitpod-io/gitpod/blob/03a1524f92844b6c4ac17cac264a13d6fd44c343/docs/workspace/namespaces.drawio.svg

https://www.phoronix.com/scan.php?page=news_item&px=CGroup-Namespaces-Linux-4.6
https://man7.org/linux/man-pages/man7/cgroup_namespaces.7.html
Before
```console
gitpod /workspace/gitpod/components/workspacekit (to/enable-cgroupns) $ cat /proc/self/cgroup 
12:rdma:/
11:pids:/kubepods/burstable/pod4613b903-4fc0-46ad-b749-744d4fe8da9d/9e28361324c3d1e766cd7b7f60c255c85802f1b2ef3744fd7883e36429db905e
10:blkio:/kubepods/burstable/pod4613b903-4fc0-46ad-b749-744d4fe8da9d/9e28361324c3d1e766cd7b7f60c255c85802f1b2ef3744fd7883e36429db905e
9:devices:/kubepods/burstable/pod4613b903-4fc0-46ad-b749-744d4fe8da9d/9e28361324c3d1e766cd7b7f60c255c85802f1b2ef3744fd7883e36429db905e
8:cpuset:/kubepods/burstable/pod4613b903-4fc0-46ad-b749-744d4fe8da9d/9e28361324c3d1e766cd7b7f60c255c85802f1b2ef3744fd7883e36429db905e
7:freezer:/kubepods/burstable/pod4613b903-4fc0-46ad-b749-744d4fe8da9d/9e28361324c3d1e766cd7b7f60c255c85802f1b2ef3744fd7883e36429db905e
6:cpu,cpuacct:/kubepods/burstable/pod4613b903-4fc0-46ad-b749-744d4fe8da9d/9e28361324c3d1e766cd7b7f60c255c85802f1b2ef3744fd7883e36429db905e
5:hugetlb:/kubepods/burstable/pod4613b903-4fc0-46ad-b749-744d4fe8da9d/9e28361324c3d1e766cd7b7f60c255c85802f1b2ef3744fd7883e36429db905e
4:memory:/kubepods/burstable/pod4613b903-4fc0-46ad-b749-744d4fe8da9d/9e28361324c3d1e766cd7b7f60c255c85802f1b2ef3744fd7883e36429db905e
3:net_cls,net_prio:/kubepods/burstable/pod4613b903-4fc0-46ad-b749-744d4fe8da9d/9e28361324c3d1e766cd7b7f60c255c85802f1b2ef3744fd7883e36429db905e
2:perf_event:/kubepods/burstable/pod4613b903-4fc0-46ad-b749-744d4fe8da9d/9e28361324c3d1e766cd7b7f60c255c85802f1b2ef3744fd7883e36429db905e
1:name=systemd:/kubepods/burstable/p
```
After
```console
gitpod /workspace/template-golang-cli (main) $ cat /proc/self/cgroup 
12:perf_event:/
11:memory:/
10:freezer:/
9:cpuset:/
8:pids:/
7:blkio:/
6:rdma:/
5:devices:/
4:hugetlb:/
3:cpu,cpuacct:/
2:net_cls,net_prio:/
1:name=systemd:/
0::/
```

## Related Issue(s)

None

## How to test

In your workspace
```console
$ cat /proc/self/cgroup 
```
And get `/`

## Release Notes

```release-note
Enable cgroup namespace.
```

## Documentation

No

### DIFF
--- a/components/workspacekit/cmd/rings.go
+++ b/components/workspacekit/cmd/rings.go
@@ -52,6 +52,9 @@ const (
 	// ring2StartupTimeout is the maximum time we wait between starting ring2 and its
 	// attempt to connect to the parent socket.
 	ring2StartupTimeout = 5 * time.Second
+
+	// https://lore.kernel.org/lkml/1450844609-9194-3-git-send-email-serge.hallyn@ubuntu.com/
+	CLONE_NEWCGROUP = 0x02000000
 )
 
 var ring0Cmd = &cobra.Command{
@@ -104,7 +107,7 @@ var ring0Cmd = &cobra.Command{
 		cmd := exec.Command("/proc/self/exe", "ring1")
 		cmd.SysProcAttr = &syscall.SysProcAttr{
 			Pdeathsig:  syscall.SIGKILL,
-			Cloneflags: syscall.CLONE_NEWUSER | syscall.CLONE_NEWNS,
+			Cloneflags: syscall.CLONE_NEWUSER | syscall.CLONE_NEWNS | CLONE_NEWCGROUP,
 		}
 		cmd.Stdin = os.Stdin
 		cmd.Stdout = os.Stdout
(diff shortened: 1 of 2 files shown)


sagor999 (commented):
Looks good to me, but would like for @Furisto or @csweichel to weigh in.

csweichel:
How does that interact with cgroup v1? Once we're on v2 I can totally see the motivation - for now I'm not sure what the effect for users would be.

utam0k:
@csweichel Thanks for your review!
Currently, this does not make me particularly happy, but it can provide users with a more blank environment.

csweichel:
Thanks @utam0k - I think the question still stands though. I could well imagine a process picking up the fact that they live in a cgroup v2 namespace and trying to do something with that. Question is then, what would happen?

utam0k:
> Thanks @utam0k - I think the question still stands though. I could well imagine a process picking up the fact that they live in a cgroup v2 namespace and trying to do something with that. Question is then, what would happen?

@csweichel Sorry if I'm wrong, but are you talking about the nsdelegate option?
https://man7.org/linux/man-pages/man7/cgroups.7.html

utam0k:
> Thanks @utam0k - I think the question still stands though. I could well imagine a process picking up the fact that they live in a cgroup v2 namespace and trying to do something with that. Question is then, what would happen?

I see it now! Does this mean to do something to the root namespace cgroup at some level of the ring? If that is the case, it might be a good idea to create the cgroup namespace in ring2. Sorry, I hadn't thought of that.

Furisto:
@csweichel Not sure I follow. The cgroup namespace just hides the parent cgroups of the cgroup the process is residing in, so from a security standpoint this should even be an improvement, as a process cannot see the structure of the whole cgroup hierarchy, only its own.

csweichel:
Indeed, my original statement came from my misconception of the function of cgroup namespaces.

That said, with the upcoming k3s enablement work, I'd rather we hold off on this chance and make it part of a larger effort to make cgroupv2 work in Gitpod workspaces. WDYT?

utam0k:
@csweichel  Thanks for the great perspective. I can't think of any specific places where this would interfere with k3s or cgroup v2 at the moment. However, k3s support is very meaningful, and I understand the desire to make as much progress as possible from the current environment. Why don't you leave it as Draft for now? How about leaving it as Draft instead of close so we don't forget about it?

@Furisto I'd like to hear your opinion as well.

Furisto:
Personally I think that with the upcoming k3s work, we will have to look into using a cgroup namespace, so we will definitely not forget about it.

utam0k:
> Personally I think that with the upcoming k3s work, we will have to look into using a cgroup namespace, so we will definitely not forget about it.

Sorry, I didn't keep track of what we needed to do with working k3s.

PR #8258, opened 2022-02-16
[ws-manager] Wait for workspace pod to be ready

### LINKED ISSUE #8253: Starting new workspace causes it to fail with OOM error from kubelet
### Bug description

Sometimes workspaces fail to start due to OOM error from kubelet.
We suspect it happens when node is at capacity, but has workspaces that are still terminating.
It seems like k8s scheduler ignores terminating pods, but kubelet doesn't.
So scheduler schedules a pod on a node that it think it should be able to run on, but then kubelet rejects it with OOM error.

This seem to be related to this:
https://github.com/kubernetes/kubernetes/issues/106884
https://github.com/kubernetes/kubernetes/issues/104560

I will create a controller that will cordon node when it reached maximum capacity of workspaces on it as a temporary workaround for this issue.

### Steps to reproduce

-

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

sagor999:
## Description

Make sure to wait for container to start up, in case of out of memory or other errors.

## Related Issue(s)

Fixes #8253

## How to test

Spin up new cluster in workspace preview env:
`./new-vm.sh -v aledbf-wait.10 -z us-west1-c`
then try to start up workspaces in it.

## Release Notes

```release-note
Improve handling of "Out of Memory" error when starting up workspaces
```

### DIFF
--- a/components/ws-manager/pkg/manager/manager.go
+++ b/components/ws-manager/pkg/manager/manager.go
@@ -263,7 +263,7 @@ func podRunning(clientset client.Client, podName, namespace string) wait.Conditi
 		case corev1.PodFailed, corev1.PodSucceeded:
 			return false, fmt.Errorf("pod ran to completion")
 		case corev1.PodPending:
-			if pod.Status.Reason == "OutOfmemory" || pod.Status.Reason == "OutOfcpu" {
+			if pod.Status.Reason == "OutOfMemory" || pod.Status.Reason == "OutOfCpu" {
 				return false, xerrors.Errorf("cannot schedule pod, reason: %s", pod.Status.Reason)
 			}
 
--- a/components/ws-manager/pkg/manager/manager.go
+++ b/components/ws-manager/pkg/manager/manager.go
@@ -25,6 +25,8 @@ import (
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


roboquat:

<details>

</details>

components/ws-manager/pkg/manager/manager.go
```diff
@@ -184,27 +186,59 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
...
-	if err != nil {
-		m, _ := json.Marshal(pod)
-		safePod, _ := log.RedactJSON(m)
 
-		if k8serr.IsAlreadyExists(err) {
-			clog.WithError(err).WithField("req", req).WithField("pod", safePod).Warn("was unable to start workspace which already exists")
-			return nil, status.Error(codes.AlreadyExists, "workspace instance already exists")
+	// create the Pod in the cluster and wait until is scheduled
+	// https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.22.md#workloads-that-saturate-nodes-with-pods-may-see-pods-that-fail-due-to-node-admission
+	backoff := wait.Backoff{
```
utam0k, line 199: I'm not familiar with Go lang, but I think it should be used `const`?

components/ws-manager/pkg/manager/manager.go
```diff
@@ -217,6 +251,29 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
...
+		err := clientset.Get(ctx, types.NamespacedName{Namespace: namespace, Name: podName}, &pod)
+		if err != nil {
+			return false, nil
+		}
+
+		switch pod.Status.Phase {
+		case corev1.PodFailed, corev1.PodSucceeded:
+			return false, fmt.Errorf("pod ran to completion")
+		case corev1.PodPending:
+			if pod.Status.Reason == "OutOfmemory" || pod.Status.Reason == "OutOfcpu" {
```
utam0k, line 266: Maybe it is like this?
https://github.com/kubernetes/kubernetes/blob/a274ec0afa6202a7da53b25ff3fe22df47a1b45b/CHANGELOG/CHANGELOG-1.22.md#workloads-that-saturate-nodes-with-pods-may-see-pods-that-fail-due-to-node-admission
```suggestion
			if pod.Status.Reason == "OutOfMemory" || pod.Status.Reason == "OutOfCpu" {
```

csweichel:
This would help with the problem we're trying to solve, but most likely deteriorate the feedback mechanism towards the user by all but removing the `Pending` phase. I.e. now, users would see a very long "preparing workspace" phase until they finally get redirected. 

How much of a concern that is is something we'd see in practice. 
Certainly this is another bit of motivation to move towards a CRD based solution.

components/ws-manager/pkg/manager/manager.go
csweichel: In Go [constants](https://go.dev/blog/constants) only work for numbers, strings and booleans.

components/ws-manager/pkg/manager/manager.go
```diff
@@ -217,6 +251,29 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
...
+		err := clientset.Get(ctx, types.NamespacedName{Namespace: namespace, Name: podName}, &pod)
+		if err != nil {
+			return false, nil
+		}
+
+		switch pod.Status.Phase {
+		case corev1.PodFailed, corev1.PodSucceeded:
+			return false, fmt.Errorf("pod ran to completion")
+		case corev1.PodPending:
+			if pod.Status.Reason == "OutOfMemory" || pod.Status.Reason == "OutOfCpu" {
```
princerachit, line 266: Would it make sense to have a prefix check of `OutOf` instead of Cpu and Memory?

components/ws-manager/pkg/manager/manager.go
```diff
@@ -217,6 +251,29 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
...
+				return false, xerrors.Errorf("cannot schedule pod, reason: %s", pod.Status.Reason)
+			}
+
+			return false, fmt.Errorf("pod ran to completion")
+		}
+
+		return true, nil
```
princerachit, line 273: This might be an edge case, but what if the pod goes into `unknown` state? Should we rather add an explicit case of `corev1.PodRunning`?

components/ws-manager/pkg/manager/manager.go
```diff
@@ -184,27 +186,59 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
...
+	if err == wait.ErrWaitTimeout {
+		err = retryErr
 	}
-	span.LogKV("event", "pod created")
+
+	if err != nil {
+		return nil, xerrors.Errorf("cannot create workspace pod: %w", err)
+	}
+
+	span.LogKV("event", "pod created and scheduled")
```
princerachit, line 241: Would it make sense to say something like `"pod started successfully"` as we are already checking the podRunning status in previous steps?

sagor999:
I reduced exponential backoff factor, as otherwise in worst case scenario it would have been waiting for hours there to retry.

components/ws-manager/pkg/manager/manager.go
```diff
@@ -199,7 +199,7 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
 	backoff := wait.Backoff{
 		Steps:    10,
 		Duration: 100 * time.Millisecond,
-		Factor:   5.0,
+		Factor:   2.5,
```
kylos101, line 202: Would it make sense to leverage the `cap` property for backoff? I believe it's intent is to ensure the duration does't exceed a certain time limit.

sagor999: Yeah, we could do that as well. What would be a reasonable cap? 5 min?

kylos101: I think 5 minutes will do. 

We can always bring it down if we need, but this way it's a predictable experience. 

W/o accounting for sleep (and jitter), that would get users to iteration 8 (but not beyond) here: https://go.dev/play/p/dtUxS6TwTrw

sagor999:
Continue working in new PR (due to werft issues in this one): https://github.com/gitpod-io/gitpod/pull/8289

PR #8266, opened 2022-02-17
installer: make OpenVSX URL configurable in the openvsx proxy

### LINKED ISSUE #8224: [installer] Make OpenVSX URL configurable
If users want to use their own OpenVSX instance (e.g. in an airgap installation), they need to configure their own OpenVSX URL.

Pothulapati:
## Description

This PR adds a new `OpenVSX` object into the top-level `Config` field
to support configurations of the `openvsx-proxy` component. Currently,
Only `URL` field is present. This is needed to support air-gap
instlalations where people are expected to host their own open-vsx.

This config is top-level and not under the workspace or IDE as this
configures the proxy, but not the IDE.

## Related Issue(s)

Fixes #8224

## How to test

Set `openVSX.url` to something else and do a `./installer render -c ./config.yaml | grep <something>`

## Release Notes

```release-note
Make Open VSX upstream URL configurable in the installer for air-gap installations
```

### DIFF
--- a/install/installer/pkg/components/openvsx-proxy/configmap.go
+++ b/install/installer/pkg/components/openvsx-proxy/configmap.go
@@ -22,7 +22,7 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		LogDebug:             false,
 		CacheDurationRegular: util.Duration(time.Minute),
 		CacheDurationBackup:  util.Duration(time.Hour * 72),
-		URLUpstream:          "https://open-vsx.org", // todo(sje): make configurable
+		URLUpstream:          ctx.Config.OpenVSX.URL,
 		URLLocal:             fmt.Sprintf("https://open-vsx.%s", ctx.Config.Domain),
 		MaxIdleConns:         1000,
 		MaxIdleConnsPerHost:  1000,
--- a/install/installer/pkg/config/v1/config.go
+++ b/install/installer/pkg/config/v1/config.go
@@ -55,6 +55,7 @@ func (v version) Defaults(in interface{}) error {
 	cfg.Workspace.Runtime.FSShiftMethod = FSShiftFuseFS
 	cfg.Workspace.Runtime.ContainerDSocket = "/run/containerd/containerd.sock"
 	cfg.Workspace.Runtime.ContainerDRuntimeDir = "/var/lib/containerd/io.containerd.runtime.v2.task/k8s.io"
+	cfg.OpenVSX.URL = "https://open-vsx.org"
 
 	return nil
 }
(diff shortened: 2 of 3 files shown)


MrSimonEmms (commented):
What happens if you run it with a config that doesn't have the `openVSX` object created? I'm just thinking about user's who create a new installation without updating their config

corneliusludmann:
> What happens if you run it with a config that doesn't have the `openVSX` object created? I'm just thinking about user's who create a new installation without updating their config

In my opinion, we should make all config values that have sensible default values optional in the config file. E.g., it should be totally fine to use a config like this:
```yaml
apiVersion: v1
domain: gitpod.my-domain.example
```

That would also make generating config values for e.g. replicated easier.

I'll create an issue for this and we can discuss this there. When we agree on this, we can leave this pull request as is and handle your concern with a more general solution of using sensible default values of missing config properties.

components/installer/pkg/config/v1/load.go
```diff
@@ -50,6 +50,9 @@ func LoadMock() *Config {
 			Kind: ObjectRefSecret,
 			Name: "https-certs",
 		},
+		OpenVSX: OpenVSX{
+			URL: "openvsx-testing.com",
```
csweichel, line 54: Why not `https://open-vsx.org`?

components/installer/pkg/config/v1/config.go
```diff
@@ -55,6 +55,7 @@ func (v version) Defaults(in interface{}) error {
 	cfg.Workspace.Runtime.FSShiftMethod = FSShiftShiftFS
 	cfg.Workspace.Runtime.ContainerDSocket = "/run/containerd/containerd.sock"
 	cfg.Workspace.Runtime.ContainerDRuntimeDir = "/var/lib/containerd/io.containerd.runtime.v2.task/k8s.io"
+	cfg.OpenVSX.URL = "https://open-vsx.org"
```
csweichel, line 58: Considering we're making this an explicit default, we should add config validation which ensures the value is an actual URL (e.g. using `url.Parse`)

components/installer/pkg/config/v1/load.go
Pothulapati: This is just mock configuration and not used anywhere, right? 🤔 I followed the pattern with other fields!

csweichel: Indeed, it is just a mock. Other than the domain, the remaining values "make sense". 
The original comment is a nit, really :)

corneliusludmann:
Looks good to me. @Pothulapati Would you mind squashing the commits into one commit?

_(By the way: I usually amend changes to the previous commit and force push them directly to the PR branch so that the Git history is always clean and ready to merge. You still see the history of changes in the PR on the Gitpod PR page and everyone can see diffs for each step even when you force push.)_

components/installer/pkg/config/v1/load.go
Pothulapati: Updated! :)

corneliusludmann (approved):

corneliusludmann:
Took the liberty and resolved the conflicts after we merged #8345. Hope, you're fine with that, @Pothulapati. :pray:

akosyakov (approved):
LGTM

corneliusludmann (approved):

Pothulapati:
Thanks @corneliusludmann for doing that!

Issue #7891, opened 2022-01-28
[telemetry] Allow to see a preview of the telemetry data in the admin dashboard


corneliusludmann:
Admins should be able to see a preview of the telemetry data that we sent for self-hosted installations.

PR #8272, opened 2022-02-17
telemetry: show collected data under the toggle

Pothulapati:
## Description

Currently, The service ping can be disabled by going into the
admin settings. Users are more likely to do that if they have
no clue what is being sent.

This PR tries to improve this experience, by showcasing the
exact data that is being sent below the option. This is possible
by creating a new TelemetryDataProvider and injecting that
both in the `installation-admin-collector` controller
and the `gitpod-service` (`getTelemetryData` func is added here
that can be used).

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/7891

## How to test

1. Run `yarn telepresence` in `/server`
2. Head-over to the `admin` settings page, and view the new information below the service ping option

## Release Notes

```release-note
Show collected telemetry data below the telemetry toggle
```

### DIFF
--- a/components/dashboard/src/admin/Settings.tsx
+++ b/components/dashboard/src/admin/Settings.tsx
@@ -5,15 +5,32 @@
  */
 
 import { useContext } from "react";
-import { InstallationAdminSettings } from "@gitpod/gitpod-protocol";
+import { TelemetryData, InstallationAdminSettings } from "@gitpod/gitpod-protocol";
 import { AdminContext } from "../admin-context";
 import CheckBox from "../components/CheckBox";
 import { PageWithSubMenu } from "../components/PageWithSubMenu";
 import { getGitpodService } from "../service/service";
 import { adminMenu } from "./admin-menu";
+import { useEffect, useState } from "react";
+import InfoBox from "../components/InfoBox";
+import { Redirect } from "react-router-dom";
+import { UserContext } from "../user-context";
 
 export default function Settings() {
     const { adminSettings, setAdminSettings } = useContext(AdminContext);
+    const [telemetryData, setTelemetryData] = useState<TelemetryData>();
+    const { user } = useContext(UserContext);
+
+    useEffect(() => {
+        (async () => {
+            const data = await getGitpodService().server.adminGetTelemetryData();
+            setTelemetryData(data)
+        })();
+    });
+
+    if (!user || !user?.rolesOrPermissions?.includes('admin')) {
+        return <Redirect to="/"/>
+    }
 
     const actuallySetTelemetryPrefs = async (value: InstallationAdminSettings) => {
         await getGitpodService().server.adminUpdateSettings(value);
@@ -26,12 +43,13 @@ export default function Settings() {
                 <h3>Usage Statistics</h3>
                 <CheckBox
                     title="Enable Service Ping"
-                    desc={<span>This is used to provide insights on how you use your cluster so we can provide a better overall experience. <a className="gp-link" href="https://www.gitpod.io/privacy">Read our Privacy Policy</a></span>}
+                    desc={<span>The following usage data is sent to provide insights on how you use your Gitpod instance, so we can provide a better overall experience. <a className="gp-link" href="https://www.gitpod.io/privacy">Read our Privacy Policy</a></span>}
                     checked={adminSettings?.sendTelemetry ?? false}
                     onChange={(evt) => actuallySetTelemetryPrefs({
                         sendTelemetry: evt.target.checked,
                     })} />
-            </PageWithSubMenu>
+                <InfoBox><pre>{JSON.stringify(telemetryData, null, 2)}</pre></InfoBox>
+            </PageWithSubMenu >
         </div >
     )
 }
--- a/components/server/src/workspace/gitpod-server-impl.ts
+++ b/components/server/src/workspace/gitpod-server-impl.ts
@@ -57,8 +57,9 @@ import { PartialProject } from '@gitpod/gitpod-protocol/src/teams-projects-proto
 import { ClientMetadata, traceClientMetadata } from '../websocket/websocket-connection-manager';
 import { ConfigurationService } from '../config/configuration-service';
 import { ProjectEnvVar } from '@gitpod/gitpod-protocol/src/protocol';
-import { InstallationAdminSettings } from '@gitpod/gitpod-protocol';
+import { InstallationAdminSettings, TelemetryData } from '@gitpod/gitpod-protocol';
 import { Deferred } from '@gitpod/gitpod-protocol/lib/util/deferred';
+import { InstallationAdminTelemetryDataProvider } from '../installation-admin/telemetry-data-provider';
 
 // shortcut
 export const traceWI = (ctx: TraceContext, wi: Omit<LogContext, "userId">) => TraceContext.setOWI(ctx, wi);    // userId is already taken care of in WebsocketConnectionManager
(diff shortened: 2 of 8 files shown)


components/dashboard/src/admin/Settings.tsx
```diff
@@ -5,33 +5,45 @@
...
                 <CheckBox
                     title="Enable Service Ping"
-                    desc={<span>This is used to provide insights on how you use your cluster so we can provide a better overall experience. <a className="gp-link" href="https://www.gitpod.io/privacy">Read our Privacy Policy</a></span>}
+                    desc={<span>This is used to provide insights on how you use your Gitpod instance so we can provide a better overall experience. <a className="gp-link" href="https://www.gitpod.io/privacy">Read our Privacy Policy</a>. The following data is sent:</span>}
                     checked={adminSettings?.sendTelemetry ?? false}
                     onChange={(evt) => actuallySetTelemetryPrefs({
                         sendTelemetry: evt.target.checked,
                     })} />
-            </PageWithSubMenu>
+                <InfoBox>{JSON.stringify(telemetryData, null, 2)}</InfoBox>
```
Pothulapati, line 45: Haven't done a lot of UI yet, Hence a newbie question. What's the nicest way to display this? 😬

corneliusludmann: As a first step, I would recommend using a `pre` block.
```suggestion
                <InfoBox><pre>{JSON.stringify(telemetryData, null, 2)}</pre></InfoBox>
```

| before      | after |
| ----------- | ----------- |
| image | image |

With this proposed change, in my opinion, it would be sufficient for a first increment (as an MVC/minimal viable change).

For a next iteration, we could think about a toggle that allows collapsing the box. For that, we should ask @gtsiolis for his always terrific input on UX design.

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -2203,6 +2206,14 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
         await this.installationAdminDb.setSettings(newSettings);
     }
 
+    async adminGetTelemetryData(ctx: TraceContext): Promise<Data> {
+        traceAPIParams(ctx, {});
+
+       return await this.telemetryDataProvider.getTelemetryData();
+    }
```
corneliusludmann, line 2213: We should add a guard that only admins can access this endpoint. Just to make sure, this information is not publicly available for all Gitpod instances.

```suggestion
    async adminGetTelemetryData(ctx: TraceContext): Promise<Data> {
       traceAPIParams(ctx, {});
       await this.guardAdminAccess("adminGetTelemetryData", {}, Permission.ADMIN_API);
       return await this.telemetryDataProvider.getTelemetryData();
    }
```

components/dashboard/src/admin/Settings.tsx
```diff
@@ -5,33 +5,45 @@
  */
 
 import { useContext } from "react";
-import { InstallationAdminSettings } from "@gitpod/gitpod-protocol";
+import { Data, InstallationAdminSettings } from "@gitpod/gitpod-protocol";
```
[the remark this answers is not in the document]
Pothulapati: Yes!

components/dashboard/src/admin/Settings.tsx
Pothulapati: Updated this to use the `pre` block, and it already looks way better!

gtsiolis: Hey everyone!

**issue**: Using the alert component[[1](https://github.com/gitpod-io/gitpod/blob/bb809461b56dcc13a74d21faa1ec943fb0b8639b/components/dashboard/src/components/AlertBox.tsx)][[2](https://github.com/gitpod-io/gitpod/blob/bb809461b56dcc13a74d21faa1ec943fb0b8639b/components/dashboard/src/components/InfoBox.tsx)], also known as`InfoBox`, feels a bit misleading. Also, including the payload expanded looks a bit verbose and will not scale well in the long run as we add more information encapsulated in the payload. Lastly although a minor issue, the help links pattern (_Learn more_, _Read our Privacy Policy_, etc) is used at the end of sentences without any following text.

**suggestion**: What do you think of using a modal to encapsulate this information to avoid injecting this kind of verbose information in the dashboard layout?

**thought**: We could also :a: drop the gray background and default to the white background as well as :b: re-use the cancel action button if that's easier to do in the spirit of MVC (minimum viable change) using the existing modal component[[1](https://github.com/gitpod-io/gitpod/blob/main/components/dashboard/src/components/Modal.tsx)]. Ideally, in the long run, this could use a markdown component which we could either show under a toggle element or within a modal like the proposal below. 💭 

| Button | Modal |
|-|-|
| Button | Modal |

Thanks for the ping @corneliusludmann! 🏓

corneliusludmann: Thanks for our input, @gtsiolis! I agree with your proposal. :+1: 

If you agree, I would like to move this part to a new issue and exclude these improvements from this PR for the following reasons:
- It is functionally working now and I think it's very important that we get this into the February release.
- For @Pothulapati, it would take some time to get familiar with the modal feature. Moving this into a follow-up PR would give him more time for this without jeopardizing the release (or someone else takes the new issue to implement this improvement—Tarun can decide).

Thus, my proposal would be:
1. Let's merge this pull request as is.
2. Move the UX proposals to a new issue and do the work in a follow-up PR.

laushinka:
@Pothulapati May I have admin access please? :)

corneliusludmann:
@Pothulapati @laushinka Gave you both admin access.

components/dashboard/src/admin/Settings.tsx
corneliusludmann: Took the liberty and created https://github.com/gitpod-io/gitpod/issues/8300.

gtsiolis: Thanks @corneliusludmann for opening the follow up issue. Agree! Re-posting from https://github.com/gitpod-io/gitpod/pull/8125#issuecomment-1036261804:

153608231-bad67564-6781-477d-af76-a57d7cbc047a

gtsiolis:
Triggered by the comment in https://github.com/gitpod-io/gitpod/pull/8272#issuecomment-1044170206, cross-posting a link to some relevant docs for visibility.

💡 **tip**: You can give give another user admin rights with one-line command, see [relevant docs](https://www.notion.so/gitpod/How-we-develop-e8ce7e562478458a9223eb9b797415b2#74153318dfc74fb589d8ac12c080a187) (internal).

Cc @laushinka @Pothulapati @corneliusludmann

components/dashboard/src/admin/Settings.tsx
```diff
@@ -5,15 +5,32 @@
...
+    const { user } = useContext(UserContext);
+
+    useEffect(() => {
+        (async () => {
+            const data = await getGitpodService().server.adminGetTelemetryData();
+            setTelemetryData(data)
+        })();
+    });
+
+    if (!user || !user?.rolesOrPermissions?.includes('admin')) {
```
Pothulapati, line 31: After this new gaurd, `admin/settings` does not open when the user isn't a admin like you'd expect!

components/dashboard/src/admin/Settings.tsx
```diff
@@ -26,12 +43,13 @@ export default function Settings() {
                 <h3>Usage Statistics</h3>
                 <CheckBox
                     title="Enable Service Ping"
-                    desc={<span>This is used to provide insights on how you use your cluster so we can provide a better overall experience. <a className="gp-link" href="https://www.gitpod.io/privacy">Read our Privacy Policy</a></span>}
+                    desc={<span>This is used to provide insights on how you use your Gitpod instance so we can provide a better overall experience. <a className="gp-link" href="https://www.gitpod.io/privacy">Read our Privacy Policy</a>. The following data is sent:</span>}
```
laushinka, line 46: Agree with @gtsiolis' suggestions on follow-up improvements for the UI 🔥 Code also looks good to me. I just have a small suggestion for the text. No strong feelings about it, though - can live with the current one :)
```suggestion
                    desc={<span>The following usage data is sent to provide insights on how you use your Gitpod instance, so we can provide a better overall experience. <a className="gp-link" href="https://www.gitpod.io/privacy">Read our Privacy Policy</a>. :</span>}
```

Pothulapati: Updated!

corneliusludmann (approved):

laushinka (approved):

jankeromnes:
FYI, this is what this page currently looks like in staging:

Screenshot 2022-03-31 at 09 21 07

As a user, I find the empty info box slightly confusing/worrying. Is this okay to ship to customers as is?

PR #8277, opened 2022-02-17
Reduce CPU requests from 1 core to 100m

mads-hartmann:
/werft with-vm=true

## Description

To avoid having workspaces be unschedulable when running in a dedicated k3s cluster in a VM this PR is reducing the CPU requests from 1 full CPU to 100m. As this is just the request (and not the limit) this shouldn't have any performance impact during development. In https://github.com/gitpod-io/gitpod/pull/8276 we increased the CPUs of the VMs to 6 (instead of 4) but even so, requesting a full CPU means you won't be able to have very many workspaces running during development so I still believe we should reduce the CPU requests for workspaces during development.

## Related Issue(s)

Fixes https://github.com/gitpod-io/ops/issues/1205

## How to test

Start a VM-based preview environment and start a workspace

```sh
werft run github -a with-vm=true
```

I was able to get a workspace scheduled and verified it was using the updated CPU requests

```
gitpod /workspace/gitpod (mads/reduce-cpu-requests) $ kubectl get pod ws-e9b0fe35-77e9-4bea-a992-412ec8358679 -o yaml | grep cpu: 
        cpu: 100m
```

## Release Notes

```release-note
NONE
```

## Documentation

N/A

### DIFF
--- a/.werft/jobs/build/deploy-to-preview-environment.ts
+++ b/.werft/jobs/build/deploy-to-preview-environment.ts
@@ -310,6 +310,9 @@ async function deployToDevWithInstaller(werft: Werft, jobConfig: JobConfig, depl
         exec(`yq w -i config.yaml jaegerOperator.inCluster false`, { slice: installerSlices.INSTALLER_RENDER });
         exec(`yq w -i config.yaml workspace.runtime.containerdRuntimeDir ${CONTAINERD_RUNTIME_DIR}`, { slice: installerSlices.INSTALLER_RENDER });
 
+        // Relax CPU contraints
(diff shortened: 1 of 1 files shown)


mads-hartmann:
@gitpod-io/engineering-workspace I'm requesting review from you as well as to ensure that you're okay with this change. I have not scoped it just to VM-based preview environments so this would also affect preview environments in core-dev ☺️

iQQBot (approved):
That make sense, the workspace in the preview environment is usually only used to test

Furisto (approved):
The only negative effect I see is that workspaces could get very sluggish when many of them are scheduled to the same node, but for testing this should be good enough.

PR #8282, opened 2022-02-17
werft/observability: Reduce Prometheus resource requests

ArthurSens:

## Description

Our Prometheus instances being deployed to both core-dev and harvester previews were requesting way more resources than necessary since the load for previews is much lower than production.

This PR significantly reduces Prometheus' resource requests for all preview environments.

## Related Issue(s)

Fixes https://github.com/gitpod-io/ops/issues/1208

## How to test

* Run `werft run github -a with-vm=true`
* SSH into VM 
* Describe Prometheus' pod and look for resource requests

## Release Notes

```release-note
NONE
```


mads-hartmann (approved):
Config looks good to me. Applied hold label so you can verify the new requests in the VM (My VM is taking loooong to boot right not, possible due to the new image, will chat with Mo)

PR #7569, opened 2022-01-12
Add private dotfiles repo support

iQQBot:
## Description

Add private dotfiles repo support

## Related Issue(s)

Fixes #7565

## How to test

go to preview environment
1. set the private dotfiles repo 
2. grant `repo` scope in Git Provider in order to access private repo
3. start a workspace and check ~/.dotfiles.log

## Release Notes

```release-note
Support private dotfiles repo
```

PR #8286, opened 2022-02-17
Fix dotfiles permission

iQQBot:
## Description

In #7569 we change to use content-services API to use download dotfiles, but it is not able to set user credential, so the .dotfiles belong root

Also, it fixes some breaking changes in https://github.com/gitpod-io/gitpod/pull/7569, which is not expected (exec multiple installation scripts, the expect is exec only once: see [document](https://www.gitpod.io/docs/config-dotfiles) )

> Gitpod will recognize and run **one of** the following install scripts from your dotfiles repository.
> If there is **no install script**, your dotfiles will be symlinked into /home/gitpod.

## Related Issue(s)

Fixes #

## How to test

1. Set a dotfiles
2. open workspaces
3. check the ~/.dotfiles permission, it should belong `gitpod`
4. check the behavior, It should work as the [document](https://www.gitpod.io/docs/config-dotfiles)  says

## Release Notes

```release-note
NONE
```

## Documentation

- [x] /werft version=staging-pd-fix-dotfiles
- [x] /werft namespace=staging-pd-fix-dotfiles

### DIFF
--- a/components/supervisor/pkg/supervisor/supervisor.go
+++ b/components/supervisor/pkg/supervisor/supervisor.go
@@ -44,7 +44,6 @@ import (
 	"github.com/gitpod-io/gitpod/common-go/analytics"
 	"github.com/gitpod-io/gitpod/common-go/log"
 	"github.com/gitpod-io/gitpod/common-go/pprof"
-	"github.com/gitpod-io/gitpod/common-go/process"
 	csapi "github.com/gitpod-io/gitpod/content-service/api"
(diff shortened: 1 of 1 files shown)


Furisto (dismissed):

mustard-mh:
branch chars too much, not work to build staging 😅

mustard-mh:
image
owner not change 

cc @iQQBot

mustard-mh:
`~/.dotfile` and `~/.dotfile/*` belong to `gitpod` now

PR #8289, opened 2022-02-17
[ws-manager] Wait for workspace pod to be ready

### LINKED ISSUE #8253: Starting new workspace causes it to fail with OOM error from kubelet

sagor999:
## Description

Make sure to wait for container to start up, in case of out of memory or other errors.

## Related Issue(s)

Fixes #8253

## How to test

Spin up new cluster in workspace preview env:
`./new-vm.sh -v pavel-oom-fix.6 -z us-west1-c`

start the workspace. it should start without issues.
now cordon the node (since we cannot simulate oom error reliably)
try to start workspace. observe pod is created and pending. after 5 seconds ws-manager will delete the pending pod (since it is considered failed now), and will create a new one.
uncordon the node. 
observe pod was created. you should be able to load into your workspace now.

caveat: if you leave node cordoned for longer than 30 seconds, then startWorkspace context will get cancelled and server will attempt to create workspace again. 

## Release Notes

```release-note
Improve handling of "Out of Memory" error when starting up workspaces
```

### DIFF
--- a/components/ws-manager/pkg/manager/manager.go
+++ b/components/ws-manager/pkg/manager/manager.go
@@ -26,6 +26,8 @@ import (
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


components/ws-manager/pkg/manager/manager.go
```diff
@@ -189,27 +191,60 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
...
 
-		clog.WithError(err).WithField("req", req).WithField("pod", safePod).Error("was unable to start workspace")
-		return nil, err
+		err = wait.PollWithContext(ctx, 100*time.Millisecond, 5*time.Second, podRunning(m.Clientset, pod.Namespace, pod.Name))
+		if err != nil {
+			m, _ := json.Marshal(pod)
+			safePod, _ := log.RedactJSON(m)
+			clog.WithError(err).WithField("req", req).WithField("pod", safePod).Error("was unable to reach ready state")
+			retryErr = err
+			return true, nil
```
kylos101, line 234: If we return true here, it'll stop retrying. I assume this is intentional, to avoid creating many pods which never enter a ready state?

Were you able to confirm that when an OOM kill happens with the kubelet that it occurs [within here](https://github.com/gitpod-io/gitpod/blob/45d9b3482cd12316065dbbcb11452836eddadc9e/components/ws-manager/pkg/manager/manager.go#L219-L225)?

sagor999: >If we return true here, it'll stop retrying. 

correct, since if it gets into that if clause, that means the pod is up and running now. 

>Were you able to confirm that when an OOM kill happens 

No, I spent whole day trying to repro OOM issue and was not able to. Most probably it requires high load on the cluster and some race condition for this to happen. At least I had no luck in workspace preview env, and I did try everything that I could think of.

sagor999: Hm. I think over here we actually should return `false`:
https://github.com/gitpod-io/gitpod/blob/45d9b3482cd12316065dbbcb11452836eddadc9e/components/ws-manager/pkg/manager/manager.go#L234
since if we return true, that will break exponential backoff. 
And we want to retry creating pod again if podRunning returned error.
so I think that line actually need to be changed.

princerachit (approved):
Code changes look good to me. I have not tested this. Putting this on hold to avoid auto merge. Pavel feel free to merge this when you are ready

csweichel (approved):

PR #8305, opened 2022-02-18
[integration tests] Support Git repo fixtures

csweichel:
## Description
Today our integration tests rely on the undocumented, unspecified structure of some repo on GitHub. This makes our tests brittle and subject to random failures which are hard to recover from, because no one remembers how the repo should look like.

This PR extends integration tests to create Git repositories before they're needed. For example, if we needed a repo of a certain structure, we can now produce it using:
```Go
token, err := integration.GitHubToken(ctx, username, api)
repo, err := git.MaterialiseGitHubRepo(ctx, token, org, git.TempRepo(true), git.Ops{
    git.OpAddFile(".gitpod.yml", "image:\n  file: Dockerfile"),
    git.OpAddFile("Dockerfile", "FROM gitpod/workspace-full:latest"),
    git.OpCommitAll("initial commit"),
})
t.Cleanup(func() {
    repo.Delete(ctx)
})

integration.LaunchWorkspaceFromContextURL(ctx, repo.ContextURL(), username, api)
```
A concrete use example can be found [here](https://github.com/gitpod-io/gitpod/blob/4a3f80c38345aa045abcf0a097b493d23640734a/test/tests/workspace/git_test.go#L113-L131).

Note: 
- to use `integration.GitHubToken` the user needs at least the `repo` and `read_user` scopes.
- because Gitpod cannot request the `repo_delete` scope, the `repo.Delete()` operation is likely to fail. 

### Repo's from files
Sometimes repo's are too complicated to be described in code. Then, we can place them alongside the test and use `git.OpsFromFixture("some-path/")` to produce the test repo.

## How to test
1. log in on the Gitpod installation you'll run the test against
2. Ensure your user has the correct permissions
  image
3. Run the test

    cd test/tests/workspace
    go test  -run TestGitActions -v ./... -namespace=staging-main -kubeconfig=/home/gitpod/.kube/config -username=<your-github-user>

4. Check your account and look for a repo named something like `testing-1645181683-1443635317331776148`

## Release Notes

```release-note
NONE
```


sagor999:
>because Gitpod cannot request the repo_delete scope, the repo.Delete() operation is likely to fail.

so after every run it will be polluting your github account with temp repos?

test/pkg/git/materialise.go
```diff
@@ -0,0 +1,297 @@
...
+	"golang.org/x/xerrors"
+)
+
+// DefaultBranch is the default branch this package is using
+const DefaultBranch = plumbing.Master
+
+// TempRepo produces a repo spec with a temporary name
+func TempRepo(private bool) RepoSpec {
+	return RepoSpec{
+		Name:    fmt.Sprintf("testing-%d-%d", time.Now().Unix(), rand.Int()),
```
utam0k, line 41: Hou about including the name `gitpod` for distinction

test/pkg/git/materialise_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	// The token needs the following scopes: repo, delete_repo
+	repo, err := MaterialiseToGitHub(context.Background(), "<your-token-here>", "", TempRepo(true), spec)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	fmt.Println(repo.CloneURL())
+
+	err = repo.Delete(context.Background())
```
utam0k, line 33: Why not add this code to `cleanUp` to make sure it runs?

csweichel:
> > because Gitpod cannot request the repo_delete scope, the repo.Delete() operation is likely to fail.
> 
> so after every run it will be polluting your github account with temp repos?

That is unfortunately correct. Three ways we can handle this:
- we run the tests with a user whose account we don't care much for, i.e. polluting the repos doesn't matter much
- we introduce a repo GC, i.e. a werft cron job which cleans out repos by name
- we find a way to give the GitHub token used by Gitpod the `repo_delete` permission, e.g. by creating the user with a prepared token. IMHO creating the user in the first place is something we need to do for empty installations anyways.
WDYT?

utam0k:
I think the third option is the most straightforward if possible. I felt that the first option is not the preferred way to use github.com.

csweichel:
@utam0k @sagor999 I've addressed the comments above. How should we continue?

sagor999:
Agreed with @utam0k , can we implement option 3? How difficult it would be to give additional permissions? @csweichel

Issue #8257, opened 2022-02-16
Restrict access to snapshot URLs


jldec:
Gitpod [snapshot URLs](https://www.gitpod.io/docs/sharing-and-collaboration#sharing-snapshots), provide a way to share a snapshot of a workspace e.g. for easy reproduction of test environments.

For workspaces of private git repos, these URLs could be used to access sensitive or proprietary files.

- Run the "Gitpod Share Workspace Snapshot" command in a workspace, to receive a new snapshot URL.
- Any Gitpod user could use this URL to open a new workspace with the same state.

The proposed fix is to limit access to the workspaces from snapshot URLs to those users who have read access to the repo.

1. Determine the git remote for the snapshot
2. Ensure that the user opening the workspace with the snaphot URL has read access to the git remote.
3. If not, show the error from the git provider, and a suggest requesting access to the repo (see below).

Since this changes the behavior of existing snapshot URLs, the error should also include the following explanation
```
Snapshot URLs require read access to the underlying repository.
Please request access from the repository owner.
```

**NOTE: Breaking change**
This change may break some usage scenerios where private repo workspaces are intentionally shared via snapshots e.g. for interviews.

- [ ] https://github.com/gitpod-io/website/issues/1625
- [ ] https://github.com/gitpod-io/website/issues/1624
- [ ] Track usage with success and access-error events (segment)

jankeromnes:
Happy to give this a go tomorrow

/assign

gtsiolis:
This could also apply to running workspaces as well as snapshots, see https://github.com/gitpod-io/gitpod/issues/4841.

See also https://github.com/gitpod-io/gitpod/issues/4841#issuecomment-881277575.

> Also, maybe it could be worth considering to share only with team members by default and allow users to manually switch to share with anyone with a link. 💡

csweichel:
> This could also apply to running workspaces as well as snapshots, see https://github.com/gitpod-io/gitpod/issues/4841.

Running workspace access is mediated by ws-proxy, who has no notion of the authentication on `server` side. We'd need to introduce `guest tokens` (which we conceived of when we introduced the `owner token`, hence the name).

chuck-confluent:
whoa whoa whoa...a workspace snapshot may have sensitive credentials in files ignored by git. The owner of the workspace should reserve the right to limit access to the URL or revoke the URL at any time, regardless of what repo permissions people have.

chuck-confluent:
[...]

PR #8306, opened 2022-02-18
[server] Restrict snapshot access based on repository access

jankeromnes:
## Description

Restrict snapshot access based on repository access.

Also refactors:
- `GuardedSnapshot.workspace` (simplified)
- `WorkspaceLogAccessGuard` → `RepositoryResourceGuard`
- `RepositoryService.canAccessHeadlessLogs` → `RepositoryProvider.hasReadAccess`

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/8257

## How to test

1. Create a snapshot for a private repo
2. Verify that you can open the snapshot
3. Verify that other members of the private repo can open the snapshot
4. Verify that non-members of the private repo can **not** open the snapshot

Also:

1. Create a snapshot for a public repo
2. Verify that anyone can open the snapshot (even if they haven't connected the repo's Git provider)

## Release Notes

```release-note
Restrict snapshot access based on repository access
```


components/server/src/gitlab/gitlab-repository-provider.ts
```diff
@@ -99,4 +99,18 @@ export class GitlabRepositoryProvider implements RepositoryProvider {
...
+            const api = await this.gitlab.create(user);
+            const response = await api.Projects.show(`${owner}/${repo}`);
+            if (GitLab.ApiError.is(response)) {
+                return false;
+            }
+            return true;
+        } catch (err) {
+            return false;
+        }
+    }
```
jankeromnes, line 115: FIXME: This also fails for public GitLab repositories when you have no connected GitLab identity:

```
Oh, no! Something went wrong!
operation not permitted: missing get permission on snapshot
Unknown Error: { "code": 403 }
```

jankeromnes:
Many thanks!

Holding because I'd like to see if I can detect the missing integration, and at least show a "Connect" button on the error screen in that case.

/hold

components/server/src/github/github-repository-provider.ts
```diff
@@ -128,4 +128,21 @@ export class GithubRepositoryProvider implements RepositoryProvider {
...
+                    repository(name: "${repo}", owner: "${owner}") {
+                        viewerPermission
+                    }
+                }
+            `);
+            return result.data.repository !== null;
+        } catch (err) {
+            return false;
+        }
+    }
```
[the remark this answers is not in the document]
geropl: You mean if you don't have a GitHub account registered with Gitpod? Yes, that's expected. Would be nice to fix that, though. :+1:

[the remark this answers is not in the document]
geropl: But maybe that's a separate problem. I like @jldec suggestions: https://github.com/gitpod-io/gitpod/pull/8306#issuecomment-1044526336

geropl (approved):
Code LGTM as well

I like the WorkspaceLogAccessGuard → RepositoryResourceGuard refactoring :+1:

jankeromnes:
Left to do:
- [ ] Understand and maybe resolve the snapshot bugs mentioned above
- [ ] Surface the message `Snapshot URLs require read access to the underlying repository. Please request access from the repository owner.` in the UI when snapshot access fails

jankeromnes:
Hmm, I've added some debug logs, but could not reproduce the errors above. I guess this works as expected.

I've also added a more specific error message when you don't have repo access:

**Integration connected, repo access**

Opening the snapshot works ✅ 

**Integration connected, no repo access**

Screenshot 2022-02-19 at 14 40 13

**Integration not connected (can't verify access)**

Screenshot 2022-02-19 at 14 37 11

I feel like we could still make these error pages look nicer somehow.

components/server/src/bitbucket-server/bitbucket-server-repository-provider.ts
```diff
@@ -100,4 +96,13 @@ export class BitbucketServerRepositoryProvider implements RepositoryProvider {
         //     authorAvatarUrl: commit.author?.user?.links?.avatar?.href,
         // };
     }
+
+    getUserRepos(user: User): Promise<string[]> {
+        throw new Error("getUserRepos not implemented.");
+    }
+
+    hasReadAccess(user: User, owner: string, repo: string): Promise<boolean> {
+        throw new Error("Method not implemented.");
```
AlexTugarev, line 105: in that case `return true` would be a better default value, I guess. even if not correct in all cases, it should be stopper in general.

jankeromnes: Good point, thanks @AlexTugarev. But `return true` would mean "allow anyone to open snapshots of Bitbucket server repositories that are potentially private", which goes somewhat against the intent of this change (i.e. restrict snapshot access to only users who are allowed to read the repository).

Then again, if you have an account for a particular Bitbucket server, maybe it means you should have access to all repos/snapshots? (E.g. maybe these Bitbucket servers are for companies, and only employees can have an account there?)

components/server/src/github/github-repository-provider.ts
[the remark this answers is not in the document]
AlexTugarev: @jankeromnes, that's ok here if we can assume that it's already checked. see https://github.com/gitpod-io/gitpod/pull/8306/files#diff-b4ab99413da84ca8b9ecc7d07a462a1eb0a7bc086cbb90b90d5b59c2b170eb44R490 there is a detailed error thrown at the frontend which can be consumed properly.

components/server/src/repohost/repository-provider.ts
```diff
@@ -14,4 +14,5 @@ export interface RepositoryProvider {
     getBranches(user: User, owner: string, repo: string): Promise<Branch[]>;
     getCommitInfo(user: User, owner: string, repo: string, ref: string): Promise<CommitInfo | undefined>;
     getUserRepos(user: User): Promise<string[]>;
+    hasReadAccess(user: User, owner: string, repo: string): Promise<boolean>;
```
AlexTugarev, line 17: 👍🏻

components/server/src/bitbucket-server/bitbucket-server-repository-provider.ts
AlexTugarev: Do I read the code right, so that I see no way to open a workspace any more for BB Server if the `Method not implemented` error remains in place? If so, please change to the default `true`, and add a `TODO` in there. We're not expecting to make it feature complete very soon.

BB Server support is an alpha feature for Gitpod Self-Hosted only.

Also, the nature of the calling site is to check for a relevant git connection first, so this is not big issue. 

> goes somewhat against the intent of this Pull Request

I agree and understand. Let's call it an exception and maybe put a short `TODO` comment in there. That BB Server support is alpha, so ... 🤷🏻‍♂️

jankeromnes: > Do I read the code right, so that I see no way to open a workspace any more for BB Server if the `Method not implemented` error remains in place?

Only when you're trying to open a Snapshot made on a Bitbucket server repo.

Creating regular workspaces or prebuilds for Bitbucket server repos continues to work as before.

However, now that you mention it, `hasReadAccess` is also called when trying to watch headless logs. 🤔 I wonder what happens in that case. Maybe returning `false` is safer here.

jankeromnes: Or, actually, returning `true` and calling that an exception as you suggested.

It's a small security concern, but going from:
- snapshots can be opened by anyone

to:
- snapshots are restricted to repo readers, except for Bitbucket servers where they're still unrestricted

is already a good first step I think.

AlexTugarev (approved):
LGTM

jldec:
This looks good now @jankeromnes - thanks for improving the error message.

jldec (approved):
merci bien

jankeromnes:
Many thanks for the reviews! 🙏

Releasing the [hold-down clamps](https://www.youtube.com/watch?v=K7akDwTjRpI), go for launch 🚀

/unhold

Issue #7863, opened 2022-01-27
[vscode-web] extension host process times out on startup


thisisommore:
### Bug description

The VsCode command to stop gitpod workspace is missing sometimes
![image](https://user-images.githubusercontent.com/51229945/151329485-1c5dc054-56dd-4920-ba1b-6d007850d445.png)
Also this button doesn't work when the issue occurs
![image](https://user-images.githubusercontent.com/51229945/151329602-dabf85c3-c091-4bfb-a68d-139cec1c7e63.png)

### Steps to reproduce

Run any workspace and try to stop it after a while.
This most of the time works, but still more frequently I started seeing such issue

### Expected behavior

The gitpod button should always be clickable and the stop workspace command always be available

### Anything else?

After reloading the tab it works

akosyakov:
Do you use stable or insiders? Is anything else working, i.e. do you have terminals, git integration and so on?  It sounds like the extension host process does not start at all?

thisisommore:
@akosyakov Yeah, I am using stable and Saas (not self-hosted), also the terminal works but the git integration doesn't.

thisisommore:
Yeah, I guess some issue with the extension host process, I will try to reproduce it again and capture logs In dev tools, will it be helpful?

mustard-mh:
Does refresh this page help? @thisisommore

akosyakov:
@thisisommore You connection is good? VS Code tries to start up extensions for 1 minute and then gives up leaving such state usually. 

You can also share your startup performance in such case, maybe there is something interesting:
<img width="685" alt="Screenshot 2022-01-27 at 10 46 49" src="https://user-images.githubusercontent.com/3082655/151334110-9a0e0a36-0397-4d43-8ae6-b41a1709694a.png">

> Does refresh this page help? @thisisommore

👍 It should work like a workaround.

thisisommore:
> 

@mustard-mh  Yes it does

thisisommore:
> @thisisommore You connection is good? VS Code tries to start up extensions for 1 minute and then gives up leaving such state usually.
> 
> You can also share your startup performance in such case, maybe there is something interesting: <img alt="Screenshot 2022-01-27 at 10 46 49" width="685" src="https://user-images.githubusercontent.com/3082655/151334110-9a0e0a36-0397-4d43-8ae6-b41a1709694a.png">
> 
> > Does refresh this page help? @thisisommore
> 
> 👍 It should work like a workaround.

[...]

Issue #7409, opened 2021-12-31
Clean up initial extension installation on vscode web 


jeanp413:
Now that theia integration is being removed https://github.com/gitpod-io/gitpod/pull/7395, we should do some cleanup when we do the initial extension installation in vscode web.
Right now, we have some custom logic inside gp-code branch to do the extension installation, we need to simplify and move this logic to  the supervisor and install the needed extension using the server argument `--install-entension` and `install-builtin-extension`

akosyakov:
Supervisor should be kept agnostic to concrete IDEs. We have though startup.sh and we could move this logic there.

Issue #4630, opened 2021-06-28
remove Theia and move extension resolution to Code


akosyakov:
We don't support user uploaded extensions anymore and going to remove Theia. After that we can remove extension resolution logic from server and resolve extensions directly from .gitpod.yml on Gitpod Code Server startup.

AlexTugarev:
is there an ETA for the departure of Theia? would be good to link here.

akosyakov:
Theia is deprecated in June release and should be removed in July release.

gtsiolis:
Cross-posting the relevant issue for removing the corresponding dashboard section, see https://github.com/gitpod-io/gitpod/issues/4486.

JanKoehnlein:
On the settings page, we state

> We're deprecating the Theia editor. You can still switch back to Theia for the next few weeks but the preference will be removed by the end of August 2021.

Maybe we should wait until end of August?

akosyakov:
But we should start working on it.

csweichel:
I'm afraid we're a bit stuffed on groundwork right now. Please add back in a week's time. 

Also there are plenty of other code issues scheduled :)

bd82:
I'm just wondering.

The fact this issue is still open means gitpod is stlil using Theia under the hood?

JanKoehnlein:
>The fact this issue is still open means gitpod is stlil using Theia under the hood?

No, it doesn't. There is just some dead code and config to be cleaned up

bd82:
> No, it doesn't. There is just some dead code and config to be cleaned up

Thanks for clearing this up.

akosyakov:
@mustard-mh @jeanp413 I think now it can be done, since we don't consume env vars computed by Gitpod Server anymore. It will remove a dependency from Server to OpenVSX and should improve workspace startup time a bit if it has .gitpod.yml.

PR #8313, opened 2022-02-18
[ide] use code flag to install extensions

mustard-mh:
## Description
Track readiness for `content` `ide` `ide-desktop` ready

Upstream vscode has provided extension install with flag `--install-extension` and there are some bugs with our installation code 
~~🐛 Remove extensionIdRegex for extensions that install via file path~~
🐛 Support extensions that their id starts with http or https

It's a Golang copy of [these codes](https://github.com/gitpod-io/openvscode-server/blob/96d57b52465f59b5c0ff650d2413e99451106f83/src/vs/gitpod/node/customServerIntegration.ts#L123-L161)

And part of extension host timeout https://github.com/gitpod-io/gitpod/issues/7863 fixes, more detail see [internal slack](https://gitpod.slack.com/archives/C01KGM9BH54/p1645044794473059)

## Related Issue(s)

Relates #7863, fixes https://github.com/gitpod-io/gitpod/issues/7409

## How to test

For test, we use this version of vscode which remove our custom extensions install only https://github.com/gitpod-io/openvscode-server/commit/6d0747dbcd685cfe47251d56a452b619c9da444b

Just remove support of install with file path

1. Make sure to use insider vscode https://hw-code-flag.staging.gitpod-dev.com/preferences
2. Open a workspace, you can use this repo https://hw-code-flag.staging.gitpod-dev.com/#github.com/mustard-mh/test/tree/vs/exthost-no-file
   image
3. Click `Extensions` button of vscode check if all your extensions installed
4. Run `ps -efj fww` should have a process like 
  ```
sh /ide/bin/gitpod-code --start-server --install-builtin-extension github.vscode-pull-request-github --install-extension bierner.color-info --install-extension vscodevim.vim@1.21.10 --install-extension magicstack.magicpython --install-extension formulahendry.code-runner --port 23000 --host 0.0.0.0 --without-connection-token --server-data-dir /workspace/.vscode-remote
  ```
5. Go to [`segment.com`](https://app.segment.com/gitpod/sources/staging_untrusted/debugger) search with `supervisor_readiness` to see event tracking

## Release Notes

```release-note
NONE
```

## Documentation


components/ide/code/extensionhelper/main.go
```diff
@@ -0,0 +1,123 @@
...
+	"sync"
+	"time"
+
+	yaml "gopkg.in/yaml.v2"
+
+	gitpod "github.com/gitpod-io/gitpod/gitpod-protocol"
+)
+
+const (
+	GITPOD_REPO_ROOT             = "GITPOD_REPO_ROOT"
```
akosyakov, line 25: it is better to use supervisor info endpoint, we should not rely on env vars, they are subject to change

I guess it was debt in VS Code Server code

components/ide/code/extensionhelper/main.go
```diff
@@ -0,0 +1,123 @@
...
+func getExtensions() (extensions []string, err error) {
+	workspaceContextUrl := os.Getenv(GITPOD_WORKSPACE_CONTEXT_URL)
+	if workspaceContextUrl != "" && strings.Contains(workspaceContextUrl, "github.com") {
+		extensions = append(extensions, "github.vscode-pull-request-github")
+	}
+	repoRoot := os.Getenv(GITPOD_REPO_ROOT)
+	if repoRoot == "" {
+		return
+	}
+	data, err := os.ReadFile(filepath.Join(repoRoot, ".gitpod.yml"))
```
akosyakov, line 47: please add await here for content ready of supervisor endpoint otherwise it does not guarantee to be present on the disk

I guess it was a bug in VS Code Server customizations before, taht we don't check for it

jeanp413:
@akosyakov we should keep supporting `GITPOD_EXTERNAL_EXTENSIONS` variable right? As typefox folks are using it [context](https://gitpod.slack.com/archives/C02AK9D4R7C/p1642430294000800)
cc @mustard-mh

jeanp413 (changes requested):

akosyakov:
@jeanp413 @mustard-mh I found https://github.com/gitpod-io/gitpod/issues/4630 which I filed last year. I think this PR goes in right direction if it does not rely on env vars anymore. I think it is fine to remove both `GITPOD_EXTERNAL_EXTENSIONS` and `GITPOD_RESOLVED_EXTENSIONS`. They were never intended to be used for other purposes. External extensions means VSIX files from .gitpod.yml. We already cover them with this PR.

We should do #4630 as a follow-up after all workspaces are running this PR. It is important since it will improve startup of workspace on the server.

mustard-mh:
We can start code first then command extentions management to install extensions

mustard-mh:
But we should wait until code exthost ready, if cli command didn't await
cc @akosyakov @jeanp413

mustard-mh:
Extensions that supervisor provided, only serve once and it's reading `.gitpod.yml` file with github API, and save it to DB.

If we use it, user start and stop worksapces to trigger extensions intall not work

So we should 

🅰️
1. read `.gitpod.yml` from disk after content init ready
2. And after exthost ready, install extensions. Wait by watch `vscode.lock` file? Or @jeanp413 have other ideas?

🅱️
1. Save extensions again before workspace stopped, (need discuss with WebApp team why it designed like this, and if changes break somthing)
2. Install extensions
(a) Start code with install extensions 
(b) Same with 🅰️ step2

💡 Other ideas...

cc @akosyakov

akosyakov:
I think .gitpod.yml already parsed and stored in db as a part of [Workspace](https://github.com/gitpod-io/gitpod/blob/f41c652a8416540a8f6df41d70bfc8db60eb47c3/components/gitpod-protocol/src/protocol.ts#L418). We could fetch it from the server with `getWorkspace` or server could install an env var like `GITPOD_VSCODE_CONFIG`. So we can access it faster. It will remove requirements to wait for content load and we could immediately start server. We will still need to wait for vsix files to download, but it should be rather a corner case.

jeanp413:
I was thinking about this yesterday too, I'll write some questions I did myself:
- What does `contentAvailable` mean?, does it mean the whole `/workspace` folder is mounted/restored? does it mean only `/workspace/repo` folder is available after cloning? does it mean both?
- If it means whole `/workspace` is restored then we really want to wait for it as `.vscode-remote` is inside `/workspace` folder
- If not, then I think we should not wait for it (I think we already do this in vscode workbench.ts right?) nor reading `.gitpod.yml` from filesystem as it will delay server startup, I see the server already knows `.gitpod.yml` contents by using github/gitlab/etc  filesystem API, so I would vote for keep using env variables or calling server api to get it's contents (I think we do this to create task terminals?)
- Regarding installing extensions from URL, we first need to download it so we can pass a path to the vsix file to the server cli, but the downloading will delay server startup too, so I think we should look into other options:
  - maybe create an internal task that uses `gp await 23000` to ensure server started, download them and install them using server cli
  - install them using gitpod extension once extension host started

akosyakov:
> If it means whole /workspace is restored then we really want to wait for it as .vscode-remote is inside /workspace folder

It is a good point, `contentAvailable` means that everything under `/workspace` was restored from backup for instance. 😬 

Maybe we could move it somehow in server? i.e. to delay file system access internally?

> If not, then I think we should not wait for it (I think we already do this in vscode workbench.ts right?) nor reading .gitpod.yml from filesystem as it will delay server startup, I see the server already knows .gitpod.yml contents by using github/gitlab/etc filesystem API, so I would vote for keep using env variables or calling server api to get it's contents (I think we do this to create task terminals?)

I think we still should do it in regardless of `contentAvailable` via env var is fastest.

> Regarding installing extensions from URL, we first need to download it so we can pass a path to the vsix file to the server cli, but the downloading will delay server startup too, so I think we should look into other options:

I think we should ignore it for now and delay. I don't think that it is very popular feature. Maybe we should add some analytics for server startup time and usage of vsix files in .gitpod.yml?

jeanp413:
> Maybe we could move it somehow in server? i.e. to delay file system access internally?

You mean `/ide` folder? does it persist data? if not i would be the same as moving it back to home folder, nice thing about having it in workspace folder is that we can look into logs of previous run and extensions already installed

> I think we still should do it in regardless of contentAvailable via env var is fastest.

Agree :+1: 

> I think we should ignore it for now and delay. I don't think that it is very popular feature. Maybe we should add some analytics for server startup time and usage of vsix files in .gitpod.yml?

:+1: to adding some analytics, also small thing about options I listed is we don't need to worry about race condition in `await extensionsReady` in server on startup

akosyakov:
> You mean /ide folder? does it persist data? if not it would be the same as moving it back to home folder, nice thing about having it in workspace folder is that we can look into logs of previous run and extensions already installed on workspace restart

As for now only `/workspace` folder is persisted. I meant in VS Code Server remote file system service, but it will require changing ms code.

Let's start with `contentAvailable` check but in parallel with resolving extensions? Plus adding analytics for startup times with info about extensions in .gitpod.yml. In one month we could make a better decision then.

jeanp413:
> Let's start with contentAvailable check but in parallel with resolving extensions? 

If we resolve extensions via env variable, `contentAvailable` check should be done in supervisor?

akosyakov:
> If we resolve extensions via env variable, contentAvailable check should be done in supervisor?

no, it is on purports to let IDEs to decide when they need to have access to workspace folder, it seems that we cannot parallelize it yet for vscode, but it won’t be necessary a case in future and applicable to other IDEs

components/ide/code/extensionhelper/main.go
```diff
@@ -0,0 +1,146 @@
...
+	yaml "gopkg.in/yaml.v2"
+
+	gitpod "github.com/gitpod-io/gitpod/gitpod-protocol"
+	supervisor "github.com/gitpod-io/gitpod/supervisor/api"
+)
+
+func main() {
+	// Wait until content ready
+	contentStatus, wsInfo, err := resolveWorkspaceInfo(context.Background())
+	if err != nil || wsInfo == nil || contentStatus == nil || !contentStatus.Available {
```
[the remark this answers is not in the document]
mustard-mh: This func `resolveWorkspaceInfo` with `wait=true` will wait until contentAvailable
https://github.com/gitpod-io/gitpod/blob/8ab2f2810bb5cebf6366d1c1dbbf04a1d69a9bd3/components/supervisor-api/status.proto#L35-L44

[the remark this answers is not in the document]
akosyakov: good point 👍 

what happens if there is a connection error, do we try again? In other cases we try forever till it works

akosyakov:
@mustard-mh brought a good point that .gitpod.yml can be changed between workspace restarts, so we cannot rely on env var from server. Let's go on with a PR like is.

We should think, like a follow-up PR, how to track time from workspace startup till responsive IDE.

@mustard-mh is it good to try? or you will do additional chagnes?

mustard-mh:
@jeanp413 @akosyakov Plan A looks good https://github.com/gitpod-io/gitpod/pull/8313#issuecomment-1045965131

mustard-mh:
> > If it means whole /workspace is restored then we really want to wait for it as .vscode-remote is inside /workspace folder
> 
> It is a good point, contentAvailable means that everything under /workspace was restored from backup for instance. 😬

But if this's a race, we should always wait for `contentAvailable`. Maybe move `.vscode-remote` to another place?

akosyakov:
> But if this's a race, we should always wait for contentAvailable. Maybe move .vscode-remote to another place?

By now, I think we should just wait as you suggested in this PR. We don't want to modify MS code if possible, moving .vscode-remote folder to another place is not an option it will cause other bugs.

components/ide/code/extensionhelper/main.go
```diff
@@ -0,0 +1,146 @@
...
+		return
+	}
+	output := "--start-server"
+	if strings.Contains(wsInfo.GetWorkspaceContextUrl(), "github.com") {
```
akosyakov, line 34: we should match host here, not entire URL

components/ide/code/extensionhelper/main.go
```diff
@@ -0,0 +1,146 @@
...
+				if err != nil {
+					return
+				}
+				extensionsMu.Lock()
+				extensions = append(extensions, location)
+				extensionsMu.Unlock()
+			}(ext)
+		} else {
+			path := filepath.Join(repoRoot, ext)
+			if isVsixFileExists(path) {
```
akosyakov, line 82: @mustard-mh it is something new? I cannot remember such cases in VS Code server customisations

components/ide/code/extensionhelper/main.go
```diff
@@ -0,0 +1,146 @@
...
+				extensions = append(extensions, path)
+			} else {
+				extensions = append(extensions, lowerCaseExtension)
```
akosyakov, line 85: should not we us mutex here as well?

components/ide/code/extensionhelper/main.go
mustard-mh: > Support install extensions with file path in `.gitpod.yml` file
This, it's easy to add, and in case users want it

akosyakov: Let's remove it for now and wait for requirements. I doubt that extensions will belong to workspace folder, but rather in stalled somewhere in /tmp during docker build for workspace.

jeanp413:
> @mustard-mh brought a good point that .gitpod.yml can be changed between workspace restarts, so we cannot rely on env var from server.

Why doesn't server update env vars when workspace is restarted?

> We don't want to modify MS code if possible, moving .vscode-remote folder to another place is not an option it will cause other bugs

@akosyakov moving `.vscode-remote` is fine now that we use `extensionsSyncStateVersion`, only reason to keep in workspace folder is just to get logs from previous runs and extensions already installed when workspace is restarted, if that's not really importat let's just move back to home folder?

akosyakov:
> Why doesn't server update env vars when workspace is restarted?

As far as I remember they are frozen on creation workspace based on content from GitHub. On restart we just reinstantiate initial env vars. But maybe I'm wrong it would be good if someone double check.

components/ide/code/extensionhelper/main.go
mustard-mh: With download func below, files will be downloaded to `/tmp/vsix*.vsix`

akosyakov:
We discussed it today with @jeanp413 and @mustard-mh there are room for optimisations but there are some uncertainness about whether it is really an issue and race conditions.

So far now we decided to go with a simple solution proposed in this PR which always waits for content availability. But we also will add a new even to track readiness of different supervisor services:
```
gitpodService.TrackEvent({
    event: 'supervisor_readiness',
    properties: {
        instanceId: string
        workspaceId: string
        kind: 'content' | 'ide' | 'ide-desktop'
        timestamp: number
    }
})
```

components/ide/code/extensionhelper/main.go
```diff
@@ -0,0 +1,146 @@
...
+	if !strings.HasSuffix(path, ".vsix") {
+		return false
+	}
+	_, err := os.Stat(path)
+	return err == nil
+}
+
+func downloadExtension(url string) (location string, err error) {
+	client := &http.Client{
+		Timeout: 10 * time.Second,
```
akosyakov, line 108: try to test with java and python whether it is enough

components/ide/code/extensionhelper/main.go
```diff
@@ -0,0 +1,174 @@
...
+	"google.golang.org/grpc"
+	yaml "gopkg.in/yaml.v2"
+
+	gitpod "github.com/gitpod-io/gitpod/gitpod-protocol"
+	supervisor "github.com/gitpod-io/gitpod/supervisor/api"
+)
+
+func main() {
+	log.SetOutput(io.Discard)
+	f, err := os.OpenFile(os.TempDir()+"/.gitpod-code-helper.log", os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0644)
```
akosyakov, line 31: Can we output to stdout/stderr instead? Otherwise we won't see it in GCP logs.

akosyakov: Alternatively in startup.sh we could flush content of log file to stdout?

It will be though to write errors to stderr, not stdout. In case if downstream tools like GCP highlight them differently.

components/ide/code/startup.sh
```diff
@@ -36,8 +36,4 @@ export USER=gitpod
 grep -rl open-vsx.org /ide | xargs sed -i "s|https://open-vsx.org|$VSX_REGISTRY_URL|g"
 
 cd /ide || exit
-if [ "$SUPERVISOR_DEBUG_ENABLE" = "true" ]; then
-    exec /ide/bin/gitpod-code --inspect --log=trace "$@"
-else
-    exec /ide/bin/gitpod-code "$@"
-fi
+/ide/codehelper "$@"
```
akosyakov, line 39: 👍

components/supervisor/pkg/supervisor/supervisor.go
```diff
@@ -1554,6 +1555,45 @@ func analyseConfigChanges(ctx context.Context, wscfg *Config, w analytics.Writer
 	}
 }
 
+func trackReadiness(ctx context.Context, gitpodService *gitpod.APIoverJSONRPC, cfg *Config, cstate *InMemoryContentState, ideReady *ideReadyState, desktopIdeReady *ideReadyState) {
```
akosyakov, line 1558: ❤️

akosyakov:
@jeanp413 I think it looks good, but I don't have time to test it yet.

components/ide/code/codehelper/main.go
```diff
@@ -0,0 +1,184 @@
...
+	supervisor "github.com/gitpod-io/gitpod/supervisor/api"
+)
+
+func main() {
+	log.Println("wait until content avaliable")
+
+	// wait until content ready
+	contentStatus, wsInfo, err := resolveWorkspaceInfo(context.Background())
+	if err != nil || wsInfo == nil || contentStatus == nil || !contentStatus.Available {
+		return
```
akosyakov, line 35: @mustard-mh we should print error here

components/ide/code/codehelper/main.go
```diff
@@ -0,0 +1,184 @@
...
+		args = append(args, "--install-extension", ext)
+	}
+
+	args = append(args, os.Args...)
+	code := "/ide/bin/gitpod-code"
+	log.Printf("run %s %s\n", code, strings.Join(args, " "))
+	cmd := exec.Command(code, args...)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		log.Println("combined output failed")
```
akosyakov, line 68: we should prtin err here to stderr

components/ide/code/codehelper/main.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	log.Println("content avalible")
+
+	// start code server with --install-extension flag
+	args := []string{"--start-server"}
+
+	if os.Getenv("SUPERVISOR_DEBUG_ENABLE") == "true" {
+		args = append(args, "--inspect", "--log=trace")
+	}
+	if ctxUrl, err := url.Parse(wsInfo.GetWorkspaceContextUrl()); err == nil {
```
akosyakov, line 45: please don't swallow errs 😄 but log them to stderr

components/ide/code/codehelper/main.go
```diff
@@ -0,0 +1,184 @@
...
+	var extensionsMu sync.Mutex
+	for _, ext := range config.Vscode.Extensions {
+		lowerCaseExtension := strings.ToLower(ext)
+		if isUrl(lowerCaseExtension) {
+			wg.Add(1)
+			go func(url string) {
+				defer wg.Done()
+				location, err := downloadExtension(url)
+				if err != nil {
+					log.Println("download url", url, "failed,", err.Error())
```
akosyakov, line 134: does it print to stderr?

mustard-mh:
Yes, I'm coding with this PR now, I'll ping you if review is ready @akosyakov 
cc @jeanp413

mustard-mh:
This PR is ready for review! ✨  @akosyakov @jeanp413

components/ide/code/codehelper/main.go
```diff
@@ -0,0 +1,205 @@
...
+func main() {
+	log.Init(ServiceName, Version, true, false)
+	startTime := time.Now()
+	log.Info("wait until content available")
+
+	// wait until content ready
+	contentStatus, wsInfo, err := resolveWorkspaceInfo(context.Background())
+	if err != nil || wsInfo == nil || contentStatus == nil || !contentStatus.Available {
+		log.WithError(err).WithField("wsInfo", wsInfo).WithField("cstate", contentStatus).Error("resolve workspace info failed")
+		return
```
akosyakov, line 45: @mustard-mh if it happens we don't start IDE at all?

components/ide/code/codehelper/main.go
```diff
@@ -0,0 +1,205 @@
...
+
+	args = append(args, os.Args...)
+	log.WithField("code", code).WithField("args", args).
+		WithField("cost", time.Now().Local().Sub(startTime).Milliseconds()).
+		Info("run cmd")
+	args = append([]string{code}, args...)
+
+	// use unix.Exec to replace the calling executable in the process tree
+	if err := unix.Exec(code, args, os.Environ()); err != nil {
+		log.WithError(err).Error("unix exec code failed")
```
akosyakov, line 84: We should exit erroneously then? not with 0 code?

components/ide/code/codehelper/main.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+}
+
+func resolveWorkspaceInfo(ctx context.Context) (*supervisor.ContentStatusResponse, *supervisor.WorkspaceInfoResponse, error) {
+	resolve := func(ctx context.Context, attempt int) (contentStatus *supervisor.ContentStatusResponse, wsInfo *supervisor.WorkspaceInfoResponse, err error) {
+		supervisorAddr := os.Getenv("SUPERVISOR_ADDR")
```
akosyakov, line 90: we should be resilient here for both supervisor calls, try like 3 times for everything at least or maybe just forever with a delay

components/ide/code/codehelper/main.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+		contentStatus, err = supervisor.NewStatusServiceClient(supervisorConn).ContentStatus(ctx, &supervisor.ContentStatusRequest{Wait: true})
+		if err != nil {
+			err = errors.New("get content available failed: " + err.Error())
+		}
+		return
+	}
+	// try resolve workspace info 10 times
+	attempt := 10
+	contentStatus, wsInfo, err := resolve(ctx, attempt)
```
akosyakov, line 112: 👍

components/ide/code/codehelper/main.go
```diff
@@ -0,0 +1,205 @@
...
+		if attempt != 0 {
+			return nil, nil, errors.New("failed with attempt 10 times")
+		}
+		time.Sleep(1 * time.Second)
+		return resolve(ctx, attempt-1)
+	}
+	return contentStatus, wsInfo, err
+}
+
+func getExtensions(repoRoot string) (extensions []string, err error) {
```
akosyakov, line 124: why do we have error if a client don't use it

maybe we should return to a client and let it log once instead of in each if in this function?

akosyakov:
@mustard-mh segment is enabled right? I don't see events for my workspace 🤔

akosyakov:
stupid me, i forgot to switch to Insiders 😬

components/ide/code/codehelper/main.go
```diff
@@ -0,0 +1,205 @@
...
+	args = append([]string{code}, args...)
+
+	// use unix.Exec to replace the calling executable in the process tree
+	if err := unix.Exec(code, args, os.Environ()); err != nil {
+		log.WithError(err).Error("unix exec code failed")
+	}
+}
+
+func resolveWorkspaceInfo(ctx context.Context) (*supervisor.ContentStatusResponse, *supervisor.WorkspaceInfoResponse, error) {
+	resolve := func(ctx context.Context, attempt int) (contentStatus *supervisor.ContentStatusResponse, wsInfo *supervisor.WorkspaceInfoResponse, err error) {
```
jeanp413, line 89: `attempt` is not used inside resolve

components/ide/code/codehelper/main.go
```diff
@@ -0,0 +1,205 @@
...
+	// try resolve workspace info 10 times
+	attempt := 10
+	contentStatus, wsInfo, err := resolve(ctx, attempt)
+	if err != nil {
+		log.WithError(err).Error("resolve workspace info failed")
+		if attempt != 0 {
+			return nil, nil, errors.New("failed with attempt 10 times")
+		}
+		time.Sleep(1 * time.Second)
+		return resolve(ctx, attempt-1)
```
jeanp413, line 119: I think you meant `attempt == 0` also shouldn't this be inside a while loop

mustard-mh:
> @mustard-mh Do you always see supervisor_readiness in segment? I am not. I see always ide, but not content. It seems to work always if I enable some desktop ide as well, but not without.
It works for me, I added logs in supervisor in case it's a segment request failed. You can check logs too

WORKSPACE.yaml
```diff
@@ -7,7 +7,7 @@ defaultArgs:
   jbMarketplacePublishTrigger: "false"
   publishToJBMarketplace: true
   localAppVersion: unknown
-  codeCommit: 0175d2fd7b619fb1c34dcf2a0e17a909acd1be3e
+  codeCommit: a229ac8c8ea6f9461a5cb306d3177765b8836641
```
jeanp413, line 10: This is no longer needed https://github.com/gitpod-io/gitpod/pull/8313#issuecomment-1047921031

components/ide/code/codehelper/main.go
```diff
@@ -0,0 +1,211 @@
...
+	log.WithField("cost", time.Now().Local().Sub(startTime).Milliseconds()).Info("content available")
+
+	code := "/ide/bin/gitpod-code"
+	// start code server with --install-extension flag
+	args := []string{code, "--start-server"}
+
+	if os.Getenv("SUPERVISOR_DEBUG_ENABLE") == "true" {
+		args = append(args, "--inspect", "--log=trace")
+	}
+	wsContextUrl := wsInfo.GetWorkspaceContextUrl()
```
jeanp413, line 56: Would it be faster to use `GITPOD_WORKSPACE_CONTEXT_URL` env variable

components/ide/code/codehelper/main.go
```diff
@@ -0,0 +1,211 @@
...
+		if ctxUrl.Host == "github.com" {
+			log.Info("ws context url is from github.com, install builtin extension github.vscode-pull-request-github")
+			args = append(args, "--install-builtin-extension", "github.vscode-pull-request-github")
+		}
+	} else {
+		log.WithError(err).WithField("wsContextUrl", wsContextUrl).Error("parse ws context url failed")
+	}
+
+	uniqMap := map[string]struct{}{}
+	extensions, err := getExtensions(wsInfo.GetCheckoutLocation())
```
jeanp413, line 67: Would it be better/faster to use `GITPOD_REPO_ROOT` env variable?

components/ide/code/codehelper/main.go
akosyakov: env vars are subjects to change, we should not rely on it, workspace info endpoint is primary way to retrieve such data.

components/ide/code/codehelper/main.go
akosyakov: env vars are subjects to change and implementation details of supervisor, we should not rely on it, workspace info endpoint is primary way to retrieve such data.

mustard-mh:
I create two branchs:

1. [branch](https://github.com/gitpod-io/gitpod/commits/hw/code-flag4) `hw/code-flag4`, it will do as vscode folks recommend https://github.com/gitpod-io/gitpod/pull/8313#issuecomment-1047921031
    - after content availabel, parse `gitpod.yml` file
    - install extensions by path first
    - start code server and (with install extension by id)
2. [branch](https://github.com/gitpod-io/gitpod/commits/hw/code-flag3) `hw/code-flag3`, it will do as what plan🅰️ said https://github.com/gitpod-io/gitpod/pull/8313#issuecomment-1045965131
    - start code server first
    - after content availabel, parse `gitpod.yml` file
    - after extension host ready (wait port 23000), install extensions:
	  - install extensions by path first
	  - install extensions by id

You can test here, **make sure you choose insiders version of vscode**

- `hw/code-flag4` https://hw-code-flag4.staging.gitpod-dev.com/workspaces, you can fast start workspace [here](https://hw-code-flag4.staging.gitpod-dev.com/#github.com/mustard-mh/test/tree/vs/exthost-no-file)
- `hw/code-flag3` https://hw-code-flag3.staging.gitpod-dev.com/workspaces, you can fast start workspace [here](https://hw-code-flag3.staging.gitpod-dev.com/#github.com/mustard-mh/test/tree/vs/exthost-no-file)

akosyakov:
@mustard-mh What did you learn comparing them? What is more efficient, error prone? Which one you recommend?

akosyakov:
Also why waiting for ext host is necessary? I thought it is needed only to start them?

jeanp413:
@mustard-mh let go with `hw/code-flag4`, for option two we still need to wait for content available as server always tries to [ensure data folder exits](https://github.com/gitpod-io/openvscode-server/blob/a55dc3f03cf8a34803fa563bb989595748fac5f1/src/vs/server/node/server.main.ts#L49-L55)

mustard-mh:
@jeanp413 @akosyakov I cherry-pick branch `hw/code-flag4`, can be reviewed now

mustard-mh:
/hold 
need to clean up custom insider for test and debug log for track

components/ide/code/codehelper/main.go
```diff
@@ -0,0 +1,249 @@
...
+		if !ext.IsUrl {
+			args = append(args, "--install-extension", ext.Location)
+		} else {
+			extPathArgs = append(extPathArgs, "--install-extension", ext.Location)
+			shouldRunExtPathArgs = true
+		}
+	}
+
+	// ensure extensions install in correct dir
+	extPathArgs = append(extPathArgs, os.Args[1:]...)
```
jeanp413, line 93: I think this is not needed

components/ide/code/codehelper/main.go
```diff
@@ -0,0 +1,249 @@
...
+	args = append(args, os.Args[1:]...)
+
+	// install path extension first
+	// see https://github.com/microsoft/vscode/issues/143617#issuecomment-1047881213
+	if shouldRunExtPathArgs {
```
jeanp413, line 98:
```suggestion
	if len(extPathArgs) > 0 {
```

components/supervisor/pkg/supervisor/supervisor.go
```diff
@@ -1554,6 +1555,48 @@ func analyseConfigChanges(ctx context.Context, wscfg *Config, w analytics.Writer
...
+		Timestamp           int64  `json:"timestamp,omitempty"`
+	}
+	trackFn := func(ctx context.Context, gitpodService *gitpod.APIoverJSONRPC, cfg *Config, kind string) {
+		gitpodService.TrackEvent(ctx, &gitpod.RemoteTrackMessage{
+			Event: "supervisor_readiness",
+			Properties: SupervisorReadiness{
+				Kind:                kind,
+				WorkspaceId:         cfg.WorkspaceID,
+				WorkspaceInstanceId: cfg.WorkspaceInstanceID,
+				Timestamp:           time.Now().Unix(),
```
jeanp413, line 1572: Let's use milliseconds to have a more accurate timestamp
```suggestion
				Timestamp:           time.Now().UnixMilli(),
```

components/ide/code/codehelper/main.go
```diff
@@ -0,0 +1,249 @@
...
+
+	log.Info("wait until content available")
+
+	// wait until content ready
+	contentStatus, wsInfo, err := resolveWorkspaceInfo(context.Background())
+	if err != nil || wsInfo == nil || contentStatus == nil || !contentStatus.Available {
+		log.WithError(err).WithField("wsInfo", wsInfo).WithField("cstate", contentStatus).Error("resolve workspace info failed")
+		return
+	}
+	log.WithField("cost", time.Now().Local().Sub(startTime).Milliseconds()).Info("content available")
```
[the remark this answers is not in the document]
mustard-mh: Or you can go to [GCP](https://console.cloud.google.com/kubernetes/workload/overview?referrer=search&project=gitpod-core-dev&pageState=(%22savedViews%22:(%22i%22:%222b951a72f1c2421288e6855ea044a825%22,%22c%22:%5B%5D,%22n%22:%5B%22staging-hw-code-flag%22%5D))). Click correct workspace pod to access logs

components/supervisor/pkg/supervisor/supervisor.go
mustard-mh: I copied this `Timestamp` from somewhere inside Gitpod repo () 🤣

jeanp413:
LGTM, left a minor comment in the gp-code branch https://github.com/gitpod-io/openvscode-server/commit/6d0747dbcd685cfe47251d56a452b619c9da444b#r67646084

@mustard-mh does this still needs to be done? if not could you squash so we can merge it
> need to clean up custom insider for test and debug log for track

WORKSPACE.yaml
```diff
@@ -7,7 +7,7 @@ defaultArgs:
   jbMarketplacePublishTrigger: "false"
   publishToJBMarketplace: true
   localAppVersion: unknown
-  codeCommit: 0175d2fd7b619fb1c34dcf2a0e17a909acd1be3e
+  codeCommit: 6d0747dbcd685cfe47251d56a452b619c9da444b
```
akosyakov, line 10: It does not make it stable, right?

akosyakov:
@mustard-mh @jeanp413 Please add a new event here:  https://www.notion.so/gitpod/dd900d5177944fc38d22bb8c7d3d2cce?v=54cdae8ff0f54b4d8235429ab9893e49

components/supervisor/pkg/supervisor/supervisor.go
```diff
@@ -1554,6 +1555,48 @@ func analyseConfigChanges(ctx context.Context, wscfg *Config, w analytics.Writer
...
 
+func trackReadiness(ctx context.Context, gitpodService *gitpod.APIoverJSONRPC, cfg *Config, cstate *InMemoryContentState, ideReady *ideReadyState, desktopIdeReady *ideReadyState) {
+	type SupervisorReadiness struct {
+		Kind                string `json:"kind,omitempty"`
+		WorkspaceId         string `json:"workspaceId,omitempty"`
+		WorkspaceInstanceId string `json:"instanceId,omitempty"`
+		Timestamp           int64  `json:"timestamp,omitempty"`
+	}
+	trackFn := func(ctx context.Context, gitpodService *gitpod.APIoverJSONRPC, cfg *Config, kind string) {
+		gitpodService.TrackEvent(ctx, &gitpod.RemoteTrackMessage{
```
akosyakov, line 1566: it does not return err or something like that to log?

WORKSPACE.yaml
mustard-mh: This is for test only

akosyakov:
Does it really fix https://github.com/gitpod-io/gitpod/issues/7863 ? I think it is just a clean up we did not really identify a root cause?

mustard-mh:
@jeanp413  found out that there is some race when installing extensions[internal chat](https://gitpod.slack.com/archives/C01KGM9BH54/p1645044794473059)

This is because our extension install custom code contains both id and URL of extension [internal chat](https://gitpod.slack.com/archives/C01KGM9BH54/p1644933376827449?thread_ts=1644431626.618499&cid=C01KGM9BH54)

This PR is related to https://github.com/gitpod-io/gitpod/issues/7863

Our plans to fix https://github.com/gitpod-io/gitpod/issues/7863 are:
1. Using server arguments (This PR)
2. Remove our custom extension install code in openvscode-server

cc @akosyakov

akosyakov:
Yeah, I seen internal chat, ok, let's close it for now and keep eyes on it.

components/supervisor/pkg/supervisor/supervisor.go
mustard-mh: I print this error in a new commit, but when this reproduce, no error is returned and there is no useful information in the log, see `mustardmh-test-aabar7hcoyk`

jeanp413 (approved):
LGTM

mustard-mh:
> I print this error in a new commit, but when this reproduce, no error is returned and there is no useful information in the log, see `mustardmh-test-aabar7hcoyk`

File an issue about segment https://github.com/gitpod-io/gitpod/issues/8516

PR #8317, opened 2022-02-18
[jb] fix #8296: validate host key fingerprint

### LINKED ISSUE #8296: Remove SSH fingerprint confirm with JetBrains IDEs

During the Jetbrains remote IDE integration, we found that there is an SSH fingerprint confirmation popup that pops up every time a new workspace is opened, and the user can only click on the confirmation for this popup, because the user can’t confirm for himself if the server fingerprint is correct

We want to introduce an API to expose the public key information of the SSH Host so that the plug-in can write to the system in advance, which can avoid this pop-up window and man-in-the-middle attacks

See internal [thread](https://gitpod.slack.com/archives/C01KGM9BH54/p1645098531326929)

image

No need to display this pop-up

akosyakov:
## Description

Gateway plugin intercepts SSH fingerprint check and validates it against public ssh key of SSH gateway instead of asking a user. If SSH keys are not present then an error is presented to a user that Gitpod installation is not supported, if keys are not matched then an error presented as well.

## Related Issue(s)

fixes https://github.com/gitpod-io/gitpod/issues/8296

## How to test

- Download Gateway plugin for this branch: https://plugins.jetbrains.com/plugin/download?rel=true&updateId=159385
- Install it from the disk in your Gateway. Restart Gateway!
- Start a workpace with IntelliJ: https://gitpod.io/#referrer:jetbrains-gateway/https://github.com/gitpod-io/spring-petclinic
- You should not see a SSH fingerprint dialog anymore.
- The thin client should connect successfully. 

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/ide/jetbrains/gateway-plugin/src/main/kotlin/io/gitpod/jetbrains/gateway/GitpodConnectionProvider.kt
+++ b/components/ide/jetbrains/gateway-plugin/src/main/kotlin/io/gitpod/jetbrains/gateway/GitpodConnectionProvider.kt
@@ -4,14 +4,25 @@
 
 package io.gitpod.jetbrains.gateway
 
+import com.fasterxml.jackson.core.type.TypeReference
+import com.fasterxml.jackson.databind.DeserializationFeature
+import com.fasterxml.jackson.databind.PropertyNamingStrategies
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.intellij.ide.BrowserUtil
 import com.intellij.openapi.components.service
 import com.intellij.openapi.diagnostic.thisLogger
 import com.intellij.remote.RemoteCredentialsHolder
+import com.intellij.ssh.AskAboutHostKey
+import com.intellij.ssh.OpenSshLikeHostKeyVerifier
+import com.intellij.ssh.connectionBuilder
+import com.intellij.ui.components.JBScrollPane
+import com.intellij.ui.components.JBTextArea
 import com.intellij.ui.dsl.builder.panel
 import com.intellij.ui.dsl.gridLayout.HorizontalAlign
 import com.intellij.ui.dsl.gridLayout.VerticalAlign
 import com.intellij.util.application
+import com.intellij.util.io.DigestUtil
+import com.intellij.util.ui.JBFont
 import com.intellij.util.ui.JBUI
 import com.intellij.util.ui.UIUtil
 import com.jetbrains.gateway.api.ConnectionRequestor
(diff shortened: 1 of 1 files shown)


components/ide/jetbrains/gateway-plugin/src/main/kotlin/io/gitpod/jetbrains/gateway/GitpodConnectionProvider.kt
```diff
@@ -263,13 +266,148 @@ class GitpodConnectionProvider : GatewayConnectionProvider {
...
+        connectParams: ConnectParams
+    ): String? {
+        var resolveJoinLinkUrl = "https://24000-${ideUrl.host}/joinLink"
+        if (!connectParams.backendPort.isNullOrBlank()) {
+            resolveJoinLinkUrl += "?backendPort=${connectParams.backendPort}"
+        }
+        return fetchWS(resolveJoinLinkUrl, connectParams, ownerToken)
+    }
+
+    private fun resolveCredentials(
```
akosyakov, line 288: @iQQBot relevant functions are `resolveCredentials`, `resolveHostKeys` and most interesting is `acceptHostKey`

components/ide/jetbrains/gateway-plugin/src/main/kotlin/io/gitpod/jetbrains/gateway/GitpodConnectionProvider.kt
```diff
@@ -263,13 +266,148 @@ class GitpodConnectionProvider : GatewayConnectionProvider {
...
+        }
+    }
+
+    private fun acceptHostKey(
+        ideUrl: URL,
+        hostKeys: List<SSHHostKey>
+    ): AskAboutHostKey {
+        val hostKeysByType = hostKeys.groupBy({ it.type.lowercase() }) { it.hostKey }
+        val acceptHostKey: AskAboutHostKey = { hostName, keyType, fingerprint, _ ->
+            if (hostName != ideUrl.host) {
```
akosyakov, line 337: @iQQBot Could you have a look please whether it makes sense to you?

iQQBot: code looks good

iQQBot: Is the latest plugin URL changed?

akosyakov: good catch, now yes

akosyakov: Maybe we can print an URL in werft job somehow. We know a version and can query JB marketplace for URL, but not sure how to display it on results page.

iQQBot: try and checked, it works, The process was very smooth and there were no more pop-ups

akosyakov: ok, let's wait for workspace team to deploy and then we can merge

akosyakov:
@iQQBot changes to ws-proxy in production, it should be good to merge and deploy

filiptronicek:
Did I miss something or is the <kbd>Restart IDE</kbd> button gone by mistake, @akosyakov? It seems I still need to restart it in order for it to work.

akosyakov:
> Did I miss something or is the Restart IDE button gone by mistake, @akosyakov? It seems I still need to restart it in order for it to work.

It is on purpose, next version of GW does not require restart which is going to be released on 24th.

akosyakov:
@filiptronicek you can test against gitpod.io to be sure, you don't need prev envs

filiptronicek:
I guess this is unrelated, but is our API broken that it returns `null` for a workspace that I don't even see in my dashboard and here it looks like it is starting/stopping?

image

akosyakov:
> I guess this is unrelated, but is our API broken that it returns null for a workspace that I don't even see in my dashboard and here it looks like it is starting/stopping?

Right I don't think it is related to this PR. Let's file another issue.  It seems you get a workspace which does not normalised context URL. Please capture in the issue: How did you start a workspace? Using a link from `How to start`?

filiptronicek (approved):
Other than the issue I mentioned, it works for me a-ok. I wasn't aware that we could get rid of the dialog and I'm so happy we did!

image

I'll hold for the review from @mustard-mh and/or any potential further code review. /hold

mustard-mh (approved):

mustard-mh:
Do steps like `How to Test`, and it works well

PR #8335, opened 2022-02-21
[ws-manager] fix incorrect status when image pull is done

### LINKED ISSUE #8323: pulling container image message but actually installing vscode extensions
I noticed that when I am waiting for workspace to start and sitting on "pulling container image" message, actual container is not actually pulling an image. Instead when I looked in logs, it was installing vscode extensions. Actual image pull was about 3 seconds and then for about 2 minutes it was installing those extensions. 
Would it be possible to add a new state for that loading screen: "setting up IDE" or something like that?
And then next step would be to investigate why sometimes it takes quite a bit of time to do that.

Separate message when we are done pulling container image and doing something else.

iQQBot:
## Description

[ws-manager] fix incorrect status when image pull is done
If `pod` is running, the workspace status should be change to `initializing`

## Related Issue(s)

Fixes #8323

## How to test

1. create a workspace
2. write a big file using `dd`
3. stop workspace
4. restart this workspace
5. you will see the most time cost is `initializing` instead of `pulling image`

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/ws-manager/pkg/manager/status.go
+++ b/components/ws-manager/pkg/manager/status.go
@@ -453,15 +453,6 @@ func (m *Manager) extractStatusFromPod(result *api.WorkspaceStatus, wso workspac
 				result.Message = fmt.Sprintf("container %s was terminated unexpectedly - workspace is recovering", cs.Name)
 				return nil
 			}
-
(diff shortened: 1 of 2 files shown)


csweichel (approved):
This change makes sense. It's mostly a left-over from the days when content init wasn't tied to container readiness. 

Great find!

PR #8338, opened 2022-02-21
Integrate .gitpod.yml tasks into JetBrains IDEs

### LINKED ISSUE #6525: Build the `tasks` command API (.gitpod.yml) into JetBrains integration
In Gitpod a user can [specify tasks](https://www.gitpod.io/docs/config-start-tasks) which are shell scripts executed on workspace container startup. We would like to allow users to interact with their terminal using the commands they specific in their `tasks` command them from within the thin client. In VS Code we are [using pseudo terminal API](https://code.visualstudio.com/api/references/vscode-api#Pseudoterminal) allowing to connect tasks. 

See: [Internal link](https://gitpod.slack.com/archives/C02BRJLGPGF/p1635854581021000)

felladrin:
## Description

Integrate .gitpod.yml tasks into JetBrains IDEs.

## Related Issue(s)

Fixes #6525

## How to test

_To be defined._

Test URL: https://gp-tasks-in-jb-terminals.staging.gitpod-dev.com/#referrer:jetbrains-gateway:intellij/https://github.com/gitpod-io/spring-petclinic

## Release Notes

```release-note
NONE
```

## Documentation

_To be defined._

### DIFF
--- a/components/ide/jetbrains/backend-plugin/gradle.properties
+++ b/components/ide/jetbrains/backend-plugin/gradle.properties
@@ -16,7 +16,7 @@ platformVersion=213-EAP-SNAPSHOT
 platformDownloadSources=true
 # Plugin Dependencies -> https://plugins.jetbrains.com/docs/intellij/plugin-dependencies.html
 # Example: platformPlugins = com.intellij.java, com.jetbrains.php:203.4449.22
-platformPlugins=Git4Idea
+platformPlugins=Git4Idea, org.jetbrains.plugins.terminal
 # Opt-out flag for bundling Kotlin standard library.
 # See https://plugins.jetbrains.com/docs/intellij/kotlin.html#kotlin-standard-library for details.
 kotlin.stdlib.default.dependency=false
--- a/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/services/TerminalService.kt
+++ b/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/services/TerminalService.kt
@@ -1,0 +1,114 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package io.gitpod.jetbrains.remote.services
+
+import com.intellij.codeWithMe.ClientId
+import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.application.runInEdt
+import com.intellij.openapi.client.ClientSessionsManager
+import com.intellij.openapi.components.Service
+import com.intellij.openapi.diagnostic.thisLogger
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.project.ProjectManager
+import com.intellij.openapi.project.ProjectManagerListener
+import io.gitpod.supervisor.api.TerminalOuterClass.*
+import io.gitpod.supervisor.api.TerminalServiceGrpc
+import kotlinx.coroutines.*
+import kotlinx.coroutines.guava.asDeferred
+import org.jetbrains.plugins.terminal.TerminalView
+
+@OptIn(DelicateCoroutinesApi::class)
+@Service
+class TerminalService {
+    init {
+        val connection = ApplicationManager.getApplication().messageBus.connect()
+        connection.subscribe(ProjectManager.TOPIC, object : ProjectManagerListener {
+            override fun projectOpened(project: Project) {
+                thisLogger().warn("[Gitpod] TerminalService detected a new project: ${project.name}.")
+                GlobalScope.launch {
+                    waitUntilGuestConnectsAsync(project, this).await()
+                    runInEdt {
+                        withClient(project) { p -> p?.let { openTerminalView(it) } }
+                    }
+                }
+            }
+        })
+    }
+
+    private suspend fun waitUntilGuestConnectsAsync(project: Project, parentScope: CoroutineScope) =
+        parentScope.async {
+            if (!isActive) {
+                return@async false
+            }
+
+            thisLogger().warn("[Gitpod] TerminalService is waiting for the guest client to connect.")
+
+            waitForClientSession(project, 1000L)
+
+            true
+        }
+
+    @Suppress("UnstableApiUsage")
+    tailrec suspend fun waitForClientSession(project: Project, checkPeriod: Long): Boolean {
+        val session = ClientSessionsManager.getInstance(project).getSessions(false).firstOrNull()
+        if (session != null) return true
+        delay(checkPeriod)
+        return waitForClientSession(project, checkPeriod)
+    }
+
+    private fun openTerminalView(project: Project) {
+        try {
+            val terminalView = TerminalView.getInstance(project)
+            terminalView.createLocalShellWidget(project.basePath, "GP Terminal", true)
+                .executeCommand("gp --help")
+            thisLogger().warn("[Gitpod] TerminalService opened a terminal.")
+        } catch (e: Exception) {
+            thisLogger().error("[Gitpod] TerminalService failed to open terminal:", e)
+        }
+
+    }
+
+    @Suppress("UnstableApiUsage")
+    private fun withClient(project: Project, action: (project: Project?) -> Unit) {
+        val session = ClientSessionsManager.getInstance(project).getSessions(false).firstOrNull()
+        if (session != null) {
+            ClientId.withClientId(session.clientId) {
+                action(project)
+            }
+        } else {
+            thisLogger().warn("[Gitpod] TerminalService failed to find the client session.")
+        }
+    }
+
+    suspend fun registerTasks() {
+        val terminalSize = TerminalSize.newBuilder().setCols(1).setRows(1).build()
+
+        val listenTerminalRequest = ListenTerminalRequest.newBuilder()/*.setAlias("aliasFromSupervisor")*/.build()
+
+        val writeTerminalResponse =
+            TerminalServiceGrpc.newFutureStub(SupervisorInfoService.channel)
+                .write(WriteTerminalRequest.newBuilder().build())
+                .asDeferred()
+                .await()
+
+        val setTerminalSizeResponse =
+            TerminalServiceGrpc.newFutureStub(SupervisorInfoService.channel)
+                .setSize(SetTerminalSizeRequest.newBuilder().build())
+                .asDeferred()
+                .await()
+
+        val shutdownTerminalResponse =
+            TerminalServiceGrpc.newFutureStub(SupervisorInfoService.channel)
+                .shutdown(ShutdownTerminalRequest.newBuilder().build())
+                .asDeferred()
+                .await()
+
+        val listTerminalsResponse =
+            TerminalServiceGrpc.newFutureStub(SupervisorInfoService.channel)
+                .list(ListTerminalsRequest.newBuilder().build())
+                .asDeferred()
+                .await()
+    }
+}
--- a/components/ide/jetbrains/backend-plugin/src/main/resources/META-INF/plugin.xml
+++ b/components/ide/jetbrains/backend-plugin/src/main/resources/META-INF/plugin.xml
@@ -14,9 +14,11 @@
     <dependencies>
         <plugin id="com.intellij.modules.platform"/>
         <plugin id="Git4Idea"/>
+        <plugin id="org.jetbrains.plugins.terminal"/>
     </dependencies>
 
     <extensions defaultExtensionNs="com.intellij">
+        <applicationService serviceImplementation="io.gitpod.jetbrains.remote.services.TerminalService" preload="true"/>
         <applicationService serviceImplementation="io.gitpod.jetbrains.remote.services.HeartbeatService" preload="true"/>
         <applicationService serviceImplementation="io.gitpod.jetbrains.remote.GitpodManager" preload="true"/>
         <notificationGroup id="Gitpod Notifications" displayType="BALLOON" isLogByDefault="false" />


mustard-mh (commented):
Failed to get terminal content with https://vn-6525.staging.gitpod-dev.com/#referrer:jetbrains-gateway:goland/https://github.com/gitpod-io/template-golang-cli

Maybe it's because it was second time I connect to this workspace via JetBrains Gateway (network issue😞)

PS. I open browser IDE too
image

felladrin:
Thanks for the feedback @mustard-mh! Gonna check this case now.

felladrin:
I've just confirmed the issue @mustard-mh mentioned.

If close the Thin Client and stop the workspace. When starting the workspace again and connecting to it, the shared terminal is there.
But if I close the Thin Client, keep the workspace running and reconnect, then the terminal is not shared anymore.

image

Maybe I can fix it by re-sharing all terminal tabs when a Thin Client connects, instead of sharing them only during their creation time.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
```diff
@@ -0,0 +1,382 @@
...
+                            toolWindowManager: ToolWindowManager
+                    ) {
+                        if (ids.contains(TerminalToolWindowFactory.TOOL_WINDOW_ID)) {
+                            debug("TerminalToolWindow got registered!")
+                            action()
+                        }
+                    }
+                }
+
+        project.messageBus
```
akosyakov, line 104: Should not we check here first whether it is already registered?

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
```diff
@@ -0,0 +1,382 @@
...
+                            action(response.tasksList)
+                        }
+                    }
+
+                    override fun onCompleted() {
+                        debug("Successfully fetched tasks from Supervisor.")
+                    }
+
+                    override fun onError(throwable: Throwable) {
+                        thisLogger()
```
akosyakov, line 136: what happens if connection drops should not we reconnect and try again?

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
```diff
@@ -0,0 +1,382 @@
...
+
+                        response.tasksList.forEach { task ->
+                            if (task.state === TaskState.opening) {
+                                hasOpenedAllTasks = false
+                            }
+                        }
+
+                        if (hasOpenedAllTasks) {
+                            this.onCompleted()
+                            action(response.tasksList)
```
akosyakov, line 127: Should not we cancel streaming here?

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
```diff
@@ -0,0 +1,382 @@
...
+import org.jetbrains.plugins.terminal.TerminalToolWindowFactory
+import org.jetbrains.plugins.terminal.TerminalView
+import org.jetbrains.plugins.terminal.cloud.CloudTerminalProcess
+import org.jetbrains.plugins.terminal.cloud.CloudTerminalRunner
+import java.io.ByteArrayOutputStream
+import java.io.PipedInputStream
+import java.io.PipedOutputStream
+
+@Suppress("UnstableApiUsage")
+class GitpodTerminalService(private val project: Project) {
```
akosyakov, line 35: Should not it implement disposable and cancel async actions if it was disposed?

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
```diff
@@ -0,0 +1,382 @@
...
+                                        throwable
+                                )
+                    }
+
+                    override fun onCompleted() {
+                        debug("Successfully got the list of Supervisor terminals.")
+                    }
+                }
+
+        terminalServiceStub.list(listTerminalsRequest, listTerminalsResponseObserver)
```
akosyakov, line 172: It should be async call which returns once, correct? Why to use observer?

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
```diff
@@ -0,0 +1,382 @@
...
+        val TITLE_KEY = Key.create<String>("TITLE_KEY")
+    }
+
+    private val terminalView = TerminalView.getInstance(project)
+    private val terminalServiceStub = TerminalServiceGrpc.newStub(GitpodManager.supervisorChannel)
+    private val statusServiceStub = StatusServiceGrpc.newStub(GitpodManager.supervisorChannel)
+    private val backendTerminalManager = BackendTerminalManager.getInstance(project)
+
+    init {
+        afterTerminalToolWindowGetsRegistered {
```
akosyakov, line 46: I wonder would not it better to use coroutines + completable futures to have squential code instead of callbacks? 

For things like list terminal request callback looks like overkill.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
```diff
@@ -0,0 +1,382 @@
...
+                            }
+                        }
+                    }
+
+                    override fun onCompleted() {
+                        debug("'${supervisorTerminal.title}' terminal finished reading stream.")
+                    }
+
+                    override fun onError(throwable: Throwable) {
+                        if (containsTerminatedStatus(throwable)) return
```
akosyakov, line 256: we should handle reconnections here, looks at how notifications are handled in `GitpodManager`

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
```diff
@@ -0,0 +1,382 @@
...
+                                        writeTerminalRequest,
+                                        writeTerminalResponseObserver
+                                )
+                            }
+                        }
+                    }
+                }
+
+        shellTerminalWidget.addListener {
+            debug("Terminal '${supervisorTerminal.title}' was closed via IDE.")
```
akosyakov, line 346: It is only triggered when a user explicitly close the terminal, right? We should not call close if a user closes the entire thin client.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
```diff
@@ -0,0 +1,382 @@
...
+
+        connectSupervisorStream(
+                shellTerminalWidget,
+                supervisorTerminal,
+                terminalOutputWriter,
+                terminalInputReader
+        )
+    }
+
+    private fun connectSupervisorStream(
```
akosyakov, line 216: Where do we handle sizes?

akosyakov:
I looked throguth the code, it looks generally good, but it does not seem that cancellations and reconnections are handled, plus I could not find where we propagate client dimensions.

In regards to `How to test`. I think it should be alright to test just with in IntelliJ, but I would expect to see test cases for different properties like:
- input/output, test on scripts producting a lot of data that it goes fast and consistent
- resizes by a user
- closing when triggered from terminal
- closing when triggered by a user explicitly
- that terminal does not get closed if thin client get closed and so on

different usages should be though through and covered

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
felladrin: Yeah, this should happen only when user clicks the "X" button on a specific terminal tab.
But let me confirm.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
felladrin: You mean `project`? If so, it's not needed, as the project is guaranteed to be registered at this point. Because we're calling this class as a `projectService`, it will only be initialized after the project is registered.

```xml
<projectService serviceImplementation="io.gitpod.jetbrains.remote.GitpodTerminalService" preload="true"/>
```

But if you're talking about `messageBus`, I'm really not checking it. I thought it was always set, for any project.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
felladrin: Calling `this.onCompleted()` ends the streaming, as I understand.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
felladrin: Well observed. I think so, and will update it!

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
felladrin: I've seen JetBrains using a lot more callbacks than completable futures, so I thought it was the best. But I agree, and will replace the callbacks by completable futures.

felladrin:
@mustard-mh, the issue about terminal content not showing, when we connect the guest client for the second time, has been fixed.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
```diff
@@ -0,0 +1,162 @@
...
+        debug("Waiting for TerminalToolWindow to be registered...")
+        val toolWindowManagerListener =
+                object : ToolWindowManagerListener {
+                    override fun toolWindowsRegistered(
+                            ids: MutableList<String>,
+                            toolWindowManager: ToolWindowManager
+                    ) {
+                        if (ids.contains(TerminalToolWindowFactory.TOOL_WINDOW_ID)) {
+                            debug("TerminalToolWindow got registered!")
+                            completableFuture.complete(null)
```
akosyakov, line 89: unsubscribe here?

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
```diff
@@ -0,0 +1,162 @@
...
+                        var hasOpenedAllTasks = true
+
+                        response.tasksList.forEach { task ->
+                            if (task.state === TaskState.opening) {
+                                hasOpenedAllTasks = false
+                            }
+                        }
+
+                        if (hasOpenedAllTasks) {
+                            this.onCompleted()
```
akosyakov, line 118: are you sure that it is enough to stop streaming? I think you need to implement `beforeStart` to get an acess to `ClientCallStreamObserver<SubscribeRequest>` and be able to call cancel on it

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
```diff
@@ -0,0 +1,162 @@
...
+                                        "Got an error while trying to fetch tasks from Supervisor.",
+                                        throwable
+                                )
+                        completableFuture.completeExceptionally(throwable)
+                    }
+                }
+
+        statusServiceStub.tasksStatus(taskStatusRequest, taskStatusResponseObserver)
+
+        return completableFuture
```
akosyakov, line 139: I will assume there should be something like `while(isAcvtive)` and try catch to handle network errors, there only exception which can allow to exit is cancellation

akosyakov:
I could not try it, i fails with:
```
2022-04-01 09:30:03,021 [   2939]  ERROR - llij.ide.plugins.PluginManager - com.intellij.util.IncorrectOperationException: Sorry but parent: services of com.intellij.openapi.project.impl.ProjectExImpl@3541637 (class com.intellij.openapi.util.Disposer$2) has already been disposed (see the cause for stacktrace) so the child: com.jetbrains.rdserver.toolWindow.BackendServerToolWindowManager@206077 (class com.jetbrains.rdserver.toolWindow.BackendServerToolWindowManager) will never be disposed 
java.util.concurrent.CompletionException: com.intellij.util.IncorrectOperationException: Sorry but parent: services of com.intellij.openapi.project.impl.ProjectExImpl@3541637 (class com.intellij.openapi.util.Disposer$2) has already been disposed (see the cause for stacktrace) so the child: com.jetbrains.rdserver.toolWindow.BackendServerToolWindowManager@206077 (class com.jetbrains.rdserver.toolWindow.BackendServerToolWindowManager) will never be disposed
```

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
felladrin: Should be unsubscribed now. (Ref: https://github.com/gitpod-io/gitpod/pull/8338/commits/77e9d0e71208d77ce61a9d1b2e66c50740600473#diff-5e11fb1aae2ce12b8efdde7b6b5e097f14940fd2bc30a775f58c48241f18ed73R94)

The only way I found to unsubscribe was by disposing the project's message bus connection.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
felladrin: Yes, according to the [docs](https://grpc.io/docs/languages/java/basics/):
"We use the response observer’s `onCompleted()` method to specify that we’ve finished dealing with the RPC."

But in any case, I tried calling `clientCallStreamObserver.cancel("Reason Here", null)` but it didn't work well, as when we cancel through this method, it invokes `onError`.

felladrin:
When https://github.com/gitpod-io/gitpod/pull/8630 gets merged, we'll be able try this PR on JetBrains EAP IDEs.

felladrin:
> I've just confirmed the issue @mustard-mh mentioned.
> 
> If I close the Thin Client and stop the workspace, the shared terminal is there when starting the workspace again and connecting to it. But if I close the Thin Client, keep the workspace running and reconnect, the terminal is not shared anymore.
> 
> image
> 
> 👆 _Screenshot of the Backend IDE, containing the terminal with history, but pointing out that it's not shared._
> 
> Maybe I can fix it by re-sharing all terminal tabs when a Thin Client connects instead of sharing them only during their creation time.

This issue is not happening anymore!
So the issue is related specifically to the previous approach used: `CloudTerminalRunner`.
Now that we're using `LocalTerminalRunner` and executing  `gp tasks attach` on it, it doesn't happen anymore.

akosyakov:
I tried with EAP of GW and IntelliJ backend and I'm getting:
Screenshot 2022-04-11 at 15 27 31

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
felladrin: Understood. I'll look more into it soon. Currently `getSupervisorTasksList` is wrapped within a `retry(3)`.

felladrin:
> I tried with EAP of GW and IntelliJ backend and I'm getting: Screenshot 2022-04-11 at 15 27 31

Checked the logs:
```
2022-04-12 14:52:28,208 [      3]   INFO - #c.i.i.StartupUtil - ------------------------------------------------------ IDE STARTED ------------------------------------------------------
2022-04-12 14:52:28,291 [     86]   INFO - #c.i.i.StartupUtil - JNA library (64-bit) loaded in 14 ms
2022-04-12 14:52:28,306 [    101]   INFO - #c.i.i.p.PluginManager - Using broken plugins file from IDE distribution
2022-04-12 14:52:28,315 [    110]   INFO - #c.i.i.StartupUtil - IDE: IntelliJ IDEA (build #IU-221.4994.44, 11 Mar 2022 18:02)
2022-04-12 14:52:28,315 [    110]   INFO - #c.i.i.StartupUtil - OS: Linux (5.4.0-1059-gke, amd64)
2022-04-12 14:52:28,320 [    115]   INFO - #c.i.i.StartupUtil - JRE: 11.0.14.1+1-b2043.11 (JetBrains s.r.o.)
2022-04-12 14:52:28,320 [    115]   INFO - #c.i.i.StartupUtil - JVM: 11.0.14.1+1-b2043.11 (OpenJDK 64-Bit Server VM)
2022-04-12 14:52:28,325 [    120]   INFO - #c.i.i.StartupUtil - WM detected: null, desktop: -
2022-04-12 14:52:28,326 [    121]   INFO - #c.i.i.StartupUtil - JVM options: [-Djava.home=/workspace/.cache/JetBrains/RemoteDev-IU/_workspace_spring-petclinic/pid.1338.temp.jbr, -Xms128m, -Xmx2048m, -XX:ReservedCodeCacheSize=512m, -XX:+IgnoreUnrecognizedVMOptions, -XX:+UseG1GC, -XX:SoftRefLRUPolicyMSPerMB=50, -XX:CICompilerCount=2, -XX:+HeapDumpOnOutOfMemoryError, -XX:-OmitStackTraceInFastThrow, -XX:MaxJavaStackTraceDepth=10000, -ea, -Dsun.io.useCanonCaches=false, -Djdk.http.auth.tunneling.disabledSchemes="", -Djdk.attach.allowAttachSelf=true, -Djdk.module.illegalAccess.silent=true, -Dkotlinx.coroutines.debug=off, -Dsun.tools.attach.tmp.only=true, -Djava.home=/workspace/.cache/JetBrains/RemoteDev-IU/_workspace_spring-petclinic/pid.1338.temp.jbr, -XX:ErrorFile=/home/gitpod/java_error_in_idea_%p.log, -XX:HeapDumpPath=/home/gitpod/java_error_in_idea_.hprof, -Djb.vmOptionsFile=/workspace/.cache/JetBrains/RemoteDev-IU/_workspace_spring-petclinic/pid.1338.temp.vmoptions, -Didea.properties.file=/workspace/.cache/JetBrains/RemoteDev-IU/_workspace_spring-petclinic/pid.1338.temp.remote-dev.properties, -Djava.system.class.loader=com.intellij.util.lang.PathClassLoader, -Didea.strict.classpath=true, -Didea.vendor.name=JetBrains, -Didea.paths.selector=IntelliJIdea2022.1, -Didea.jre.check=true, -Dsplash=true]
2022-04-12 14:52:28,326 [    121]   INFO - #c.i.i.StartupUtil - args: [cwmHostNoLobby, /workspace/spring-petclinic]
2022-04-12 14:52:28,327 [    122]   INFO - #c.i.i.StartupUtil - library path: /usr/java/packages/lib:/usr/lib64:/lib64:/lib:/usr/lib
2022-04-12 14:52:28,327 [    122]   INFO - #c.i.i.StartupUtil - boot library path: /ide-desktop/backend/jbr/lib
2022-04-12 14:52:28,367 [    162]   INFO - #c.i.i.StartupUtil - locale=en_US JNU=UTF-8 file.encoding=UTF-8
  idea.config.path=/workspace/.config/JetBrains/RemoteDev-IU/_workspace_spring-petclinic
  idea.system.path=/workspace/.cache/JetBrains/RemoteDev-IU/_workspace_spring-petclinic
  idea.plugins.path=/workspace/.config/JetBrains/RemoteDev-IU/_workspace_spring-petclinic/plugins
  idea.log.path=/workspace/.cache/JetBrains/RemoteDev-IU/_workspace_spring-petclinic/log
2022-04-12 14:52:28,384 [    179]   INFO - #c.i.i.StartupUtil - CPU cores: 1; ForkJoinPool.commonPool: java.util.concurrent.ForkJoinPool@5c58b26d[Running, parallelism = 1, size = 1, active = 1, running = 1, steals = 0, tasks = 0, submissions = 1]; factory: com.intellij.concurrency.IdeaForkJoinWorkerThreadFactory@7faded87
2022-04-12 14:52:28,405 [    200]   INFO - STDOUT - This IDE build has expired. Please download a new build from JetBrains official site
2022-04-12 14:52:28,406 [    201]   INFO - #c.i.i.StartupUtil - ------------------------------------------------------ IDE SHUTDOWN ------------------------------------------------------
```

> This IDE build has expired. Please download a new build from JetBrains official site

I'll check if updating the IDE version solves it.

felladrin:
/hold

I'm holding this until we make a decision about https://github.com/gitpod-io/gitpod/pull/9268, which should solve the [issue mentioned by Anton](https://github.com/gitpod-io/gitpod/pull/8338#issuecomment-1095052604).

mustard-mh:
I cannot see full terminal content, is it correct (like we cannot do more progress now)?
|Origin Terminal in Thin Client ❌  | Main Window ✅|Run `gp tasks attach` in Thin Client ✅|
|:-:|:-:|:-:|
|image|image|image|

felladrin:
> I cannot see full terminal content, is it correct (like we cannot do more progress now)?

Thanks for reviewing it, @mustard-mh!
Unfortunately, theres't nothing else we can do about it at this moment. JetBrains Team confirmed the client-side history size is currently hardcoded within Shared Terminals feature and there's no workaround available.

mustard-mh (dismissed):
/hold

Test good

Why we need `codewithme` plugin added if `GitpodTerminalService` not use it? is it a dependency of other plugins?

felladrin:
> Why we need `codewithme` plugin added if `GitpodTerminalService` not use it? is it a dependency of other plugins?

Nice catch! I agree it's not clear in the code, but we need to add `codewithme` as a dependency to be able to import/use `com.jetbrains.rdserver.terminal.BackendTerminalManager` on `GitpodTerminalService`. ([Internal Discussion](https://gitpod.slack.com/archives/C02BRJLGPGF/p1645472052666479?thread_ts=1645431893.085369&cid=C02BRJLGPGF))

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
```diff
@@ -0,0 +1,167 @@
...
+            override fun onNext(response: TasksStatusResponse) {
+                debug("Received task list: ${response.tasksList}")
+
+                for (task in response.tasksList) {
+                    if (task.state === TaskState.opening) {
+                        return
+                    }
+                }
+
+                this.onCompleted()
```
akosyakov, line 126: I still don't understand how it works. `onCompleted` method is overriden and does nothing but print a message. Should not we call here `request.cancel` as well to stop streaming which wil lead to `onCompleted` eventually? Or maybe `onCompleted` should call `super.onCompleted` to trigger some default behaviour which stops streaming?

Could you point me please to relevant docs? 🙏

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
```diff
@@ -0,0 +1,167 @@
...
+
+    init {
+        if (!application.isHeadlessEnvironment) {
+            launch()
+        }
+    }
+
+    private fun launch() = GlobalScope.launch {
+        retry(3) { getTerminalToolWindowRegisteredEvent().await() }
+        val tasks = retry(3) { getSupervisorTasksList().await() }
```
akosyakov, line 55: What is the value if retry fails 3 times? It seems code below assume only successful case? Why is it better than retrying forever and handling cancellation?

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    init {
+        if (!application.isHeadlessEnvironment) {
+            launch()
+        }
+    }
+
+    private fun launch() = GlobalScope.launch {
+        retry(3) { getTerminalToolWindowRegisteredEvent().await() }
```
akosyakov, line 54: here retry seems to be unnecessary? getTerminalToolWindowRegisteredEvent does not complete with failure?

akosyakov:
I opened this PR with prev envs, but it showed me only 2 terminals instead of 4:
Screenshot 2022-04-14 at 10 52 37

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    private fun createSharedTerminal(supervisorTerminal: TerminalOuterClass.Terminal) {
+        debug("Creating shared terminal '${supervisorTerminal.title}' on Backend IDE")
+
+        val shellTerminalWidget = terminalView.createLocalShellWidget(project.basePath, supervisorTerminal.title)
+
+        backendTerminalManager.shareTerminal(shellTerminalWidget, supervisorTerminal.alias)
+
+        shellTerminalWidget.executeCommand("gp tasks attach ${supervisorTerminal.alias}")
```
akosyakov, line 157: we are loosing control with that, no? We need to propagate proper UI sizes to terminal, close it then a user closes, and close a frontend terminal then an underlying terminal is closed?

felladrin:
Setting it back to Draft while I analyze the last reviews comments.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
felladrin: Indeed. But I can't get back using CloudTerminalRunner until [#CWM-5523: Support CloudTerminalRunner mirroring to thin client](https://youtrack.jetbrains.com/issue/CWM-5523) is resolved. ([Internal Discussion](https://gitpod.slack.com/archives/C02BRJLGPGF/p1646752507164589?thread_ts=1646056923.510109&cid=C02BRJLGPGF))

There are like 5x more issues when using CloutTerminalRunner in place of `go tasks attach`.

Maybe the best solution for this changes is creating a custom IntelliJ Tool Window, as suggested by JetBrains Team, but that would require a totally different implementation/PR.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
felladrin: Sure, I found the documentation [here](https://grpc.io/docs/languages/java/basics/): "We use the response observer’s `onCompleted()` method to specify that we’ve finished dealing with the RPC."

And I agree that it needs to be changed to `super.onCompleted` for it to make effect. Although I couldn't find any usage of `super.onCompleted` in the documentation examples. I'll check it and update this part if possible.

And I tried calling `request.cancel` (`clientCallStreamObserver.cancel("Reason Here", null)`) but it didn't work well, because when we cancel through this method, it invokes `onError` from this object.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
felladrin: Alright, all `retry` have been replaced by `while(isAcvtive)`.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
felladrin: Well observed! Removed it.
Thanks!

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
akosyakov: > Maybe the best solution for this changes is creating a custom IntelliJ Tool Window, as suggested by JetBrains Team, but that would require a totally different implementation/PR.

Agree it is better to wrap up what we have and tackle other things in followed up.

It should at least open all task terminals, not like in https://github.com/gitpod-io/gitpod/pull/8338#issuecomment-1098878220. Fixing closing behaviour will be good but let's do it with another PR too.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
felladrin: A forever retrial is better, indeed.
I've updated the code, removing those `retry`.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
felladrin: After checking how you've done it on GitpodManager class, the idea got clearer, so started using `request.cancel`. I've updated the code.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
felladrin: I'm now trying to use `backendTerminalManager.createNewSharedTerminal` (as when I use it, all the terminals are created accordingly and displayed on the Thin Client), but the consequence is that there's no easy way to find which terminal was created by this method.
I'm looking into a way to find each terminal and run `gp attach` in each. (https://github.com/gitpod-io/gitpod/pull/8338/commits/b82c26ed84fb665d6918392c3f35e15b5fe0c2a4)

felladrin:
We decided not to merge this PR, due to the [known limitations](https://github.com/gitpod-io/gitpod/pull/8338#issue-1145544223) regarding Shared Terminals from IntelliJ Platform.

We'll get back to it when Shared Terminals feature become more stable.

Here's the list of issues from JetBrains Issue Tracker that we need to keep an eye on:

- [Stopping and restarting a Shared Terminal session from Backend IDE causes duplicated/quadrupled characters to appear on Thin Client](https://youtrack.jetbrains.com/issue/CWM-5808/Stopping-and-restarting-a-Shared-Terminal-session-from-Backend-I)
- ["Open In/Terminal" context menu is not working](https://youtrack.jetbrains.com/issue/CWM-5272)
- [Soft-wrap doesn't work in Terminal](https://youtrack.jetbrains.com/issue/CWM-5248/Soft-wrap-doesn't-work-in-Terminal)
- [Offer port forwarding for the built-in terminal](https://youtrack.jetbrains.com/issue/CWM-4821/Offer-port-forwarding-for-the-built-in-terminal)
- [Support CloudTerminalRunner mirroring to thin client](https://youtrack.jetbrains.com/issue/CWM-5523)
- [Terminal breaks line during inputting commands on controller](https://youtrack.jetbrains.com/issue/CWM-5046/Terminal-breaks-line-during-inputting-commands-on-controller)

mikenikles:
> We'll get back to it when Shared Terminals feature become more stable.

Is there an issue in this repo where you track this? I'd like to subscribe to get notified when this feature is available as the current functionality of showing a single terminal when `.gitpod.yml` defines 5 terminals isn't developer-friendly and adds friction.

felladrin:
> > We'll get back to it when Shared Terminals feature become more stable.
> 
> 
> 
> Is there an issue in this repo where you track this? I'd like to subscribe to get notified when this feature is available as the current functionality of showing a single terminal when `.gitpod.yml` defines 5 terminals isn't developer-friendly and adds friction.

Thanks for asking! We did implement it, but the feature currently available when Gitpod Settings has "Use Latest (Unstable)" option checked. We're improving it before making it available for Stable IDEs.

Here's the PR, for reference:

- https://github.com/gitpod-io/gitpod/pull/10595

mikenikles:
Awesome, thanks for the update Victor.

PR #8354, opened 2022-02-21
[Analytics] Prepare for Mixpanel Migration of website Insights

jakobhero:
## Description

PR prepares to move visitor (i.e. everyone but authenticated user) data out of mixpanel through the following two steps:
- track the qualification (whether user has previously visited the website) at signup. as it will not be possible to compute this in mixpanel anymore by checking if a website visit was tracked from the user's profile previously, this is now passed to the app through a cookie and inside the payload of the signup event
- extend the implementation of the `SegmentAnalyticsWriter` to feature the integrations object in the payload, which ensures that an event is not forwarded to Mixpanel if the user is not authenticated. This has to be specified in the app as the event filtering in Segment unfortunately does not allow to check for the existence of `userId`

These two changes are prerequisites to converting to a paying subscription with Mixpanel as we would else either risk data loss or a spike in the cost of ownership.

## How to test

- open preview environment and [Segment's Staging Untrusted](https://app.segment.com/gitpod/sources/staging_untrusted/debugger)
- Check in the raw tab of the first `page()` call whether `integrations` exists with `"Mixpanel": false` (the initial page call should occur before authenticating and thus not be forwarded to Mixpanel as the `userId` is not passed along)
- sign up
- Check in the raw tab of the `signup` track event sent to Segment that `integrations` exists with `"Mixpanel": true` (signups always contain the `userId`, so Segment should forward this to Mixpanel). Then, check that the signup event contains a `qualified` property with the value `false` (as the cookie `gitpod-marketing-website-visited` should not yet exist for this domain)
- Next, open a different browser or incognito window, and open the preview environment once again. Then, set the cookie `gitpod-marketing-website-visited=true` (current domain, expiry at least session; one way to do this is to go to the console of the developer tool and enter `document.cookie="gitpod-marketing-website-visited=true"`)
- Now, sign up with a different account to trigger another `signup` event call and make sure that the `qualified` property in the Segment Debugger is `true` this time

## Release Notes

```release-note
NONE
```

## Documentation

### DIFF
--- a/components/gitpod-protocol/src/util/analytics.ts
+++ b/components/gitpod-protocol/src/util/analytics.ts
@@ -30,7 +30,13 @@ class SegmentAnalyticsWriter implements IAnalyticsWriter {
 
         identify(msg: IdentifyMessage) {
         try {
-            this.analytics.identify(msg, (err: Error) => {
+            this.analytics.identify({
+                ...msg,
+                integrations: {
+                    "All": true,
+                    "Mixpanel": !!msg.userId
+                }
+            }, (err: Error) => {
                 if (err) {
                     log.warn("analytics.identify failed", err);
                 }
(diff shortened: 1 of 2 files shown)


roboquat:

<details>

</details>

geropl (approved):
LGTM, tested and works as advertised

PR #8361, opened 2022-02-21
Support custom docker daemon args 

### LINKED ISSUE #8103: [docker] Support arguments to Docker daemon
We automatically start the Docker daemon when it's needed in the workspace. Currently users cannot provide their own arguments to that daemon, which bars them from using things like [userns-remap](https://docs.docker.com/storage/storagedriver/vfs-driver/). We should provide a way for folks to add those arguments.

One of those options:

1. A special environment variable which we consume and add, e.g. `DOCKER_ARGS`
2. Part of the `.gitpod.yml`
3. A magic file that we read prior to starting the Docker daemon

Furisto:
## Description
This allows users to add arguments to the docker daemon running in the workspace. As a first step mapping the container user to the gitpod user is supported. 

## Related Issue(s)
Fixes #8103

## How to test
- Add environment variable DOCKER_DAEMON_ARGS to your user settings. The content should look like this:
`{ "remap-user": "1000" }` (Means container id 1000 will be mapped to gitpod user)
- Start workspace
- Run docker image which supports running as unprivileged user e.g. `sudo docker run -it -u node -v /tmp/me:/tmp node sh`
- Create file within container e.g. touch /tmp/foo
- Verify that the created file has gitpod:gitpod permissions in the workspace

## Release Notes

```release-note
Docker in workspaces now allows mapping the user id of a container user to the gitpod user
```

### DIFF
--- a/components/docker-up/.gitignore
+++ b/components/docker-up/.gitignore
@@ -2,3 +2,4 @@ bin
 docker-compose
 docker.tgz
 slirp4netns
+runc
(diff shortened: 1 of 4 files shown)


Furisto:
Closed in favor of #8435 as I had problems with my preview environment  and had to create another branch.

PR #8362, opened 2022-02-21
Fix VM image ID

meysholdt:
## Description
Follow-up PR to https://github.com/gitpod-io/gitpod/pull/8357. The image-ID of the VM image must be adjusted in another location. 

## Related Issue(s)

Fixes #

## How to test
run `/werft run with-vm=true` and confirm that the preview env is coming up.

## Release Notes

```release-note
NONE
```


mads-hartmann (approved):
Added the hold label as it looks like you're still testing (from the werft trigger comment) but the code changes look good so feel free to remove the label :)

PR #8366, opened 2022-02-21
[ws-daemon] Improve cache error handling

aledbf:
## Description

A non-existing file is not an error (in these cases)

Check logging events https://cloudlogging.app.goo.gl/FPPVKf8QeobE9PYq6

## Release Notes
```release-note
NONE
```

### DIFF
--- a/components/ws-daemon/pkg/daemon/cache_reclaim.go
+++ b/components/ws-daemon/pkg/daemon/cache_reclaim.go
@@ -7,6 +7,8 @@ package daemon
 import (
 	"bufio"
 	"context"
+	"errors"
+	"io/fs"
 	"io/ioutil"
 	"math"
 	"os"
@@ -116,6 +118,11 @@ func readLimit(memCgroupPath string) (uint64, error) {
 	fn := filepath.Join(string(memCgroupPath), "memory.limit_in_bytes")
 	fc, err := os.ReadFile(fn)
 	if err != nil {
+		// TODO(toru): find out why the file does not exists
+		if errors.Is(err, fs.ErrNotExist) {
+			return 0, nil
+		}
+
 		return 0, xerrors.Errorf("cannot read memory.limit_in_bytes: %v", err)
 	}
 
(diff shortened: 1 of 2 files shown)


aledbf:
@utam0k can you review this change, please? (you added the tests).

The current behavior returns an error, but that pollutes the logs. Is this change acceptable?

utam0k:
> @utam0k can you review this change, please? (you added the tests).
> 
> The current behavior returns an error, but that pollutes the logs. Is this change acceptable?

Thanks for your PR.  I think we can investigate this a little more carefully. This is because this update is performed when the pod is updated, so it's unnatural that these file doesn't exist. I think we can accept this PR as it stands, but I think we need to create an issue to resolve or research this problem.

utam0k:
@aledbf I've created the issue. Can I ask you to leave the `TODO` or something comment in error handling codes.?
https://github.com/gitpod-io/gitpod/issues/8368

csweichel (approved):
LGTM

aledbf:
> @aledbf I've created the issue. Can I ask you to leave the `TODO` or something comment in error handling codes.? #8368

done

components/ws-daemon/pkg/daemon/cache_reclaim.go
```diff
@@ -116,6 +118,11 @@ func readLimit(memCgroupPath string) (uint64, error) {
 	fn := filepath.Join(string(memCgroupPath), "memory.limit_in_bytes")
 	fc, err := os.ReadFile(fn)
 	if err != nil {
+		// TODO(toru): find out why the file does not exists
+		if errors.Is(err, fs.ErrNotExist) {
```
csweichel, line 122: Most likely the file doesn't exist anymore because our "dispatch" mechanism takes too long to cancel the context. I.e. the container is already removed before the dispatch mechanism shuts down this mechanism.

Furisto: Should we have todos in the code? If we have them we should at least create an issue and reference them in the comment.

utam0k: The issue has already been created.
https://github.com/gitpod-io/gitpod/issues/8368

utam0k (approved):
LGTM

components/ws-daemon/pkg/daemon/cache_reclaim.go
Furisto: I saw it :+1: , but was asking more generally if we should have todos in the code.

utam0k: Thanks for your comment 🙏 It seems quite unnatural to ignore this error. So I'd like to write something about the reason.

Furisto (approved):

csweichel (approved):

Issue #8238, opened 2022-02-16
Oh, no! Something went wrong! OutOfmemory: Pod Node didn't have enough resource: memory, requested: 2576980378, used: 65006469130, capacity: 66388008960


dkelly255:
### Bug description

Oh, no! Something went wrong!
OutOfmemory: Pod Node didn't have enough resource: memory, requested: 2576980378, used: 65006469130, capacity: 66388008960

### Steps to reproduce

Attempting to launch my existing Gitpod workspace.

### Workspace affected

https://dkelly255-djangotutorial-quxbwl9i55a.ws-eu32.gitpod.io/

### Expected behavior

Have my existing Gitpod workspace launch without any errors.

### Example repository

Receiving error on both of these repositories:

https://dkelly255-djangotutorial-quxbwl9i55a.ws-eu32.gitpod.io/

https://dkelly255-4antasyootball-x5dfnf9rm5f.ws-eu32.gitpod.io/

### Anything else?

Receiving error on both of these repositories:

https://dkelly255-djangotutorial-quxbwl9i55a.ws-eu32.gitpod.io/

https://dkelly255-4antasyootball-x5dfnf9rm5f.ws-eu32.gitpod.io/

pawlean:
Hey @dkelly255! We have an ongoing incident: https://www.gitpodstatus.com/incidents/6hc6n6zdkhnt 

Hang tight 👍🏼

dkelly255:
Hi,
I am still getting the same error – I have tried restarting my machine, and it happens when I try to open either workspace as of 10:34am:

***@***.***

mjjstockman:
Same for me on https://gitpod.io/start/#mjjstockman-setlistshare-nee26x1ripa

dkelly255:
Note - rebuilding the workspace has allowed me to access again (I am still receiving the original error when trying to open the original workspaces) - so maybe a note should be added to the solution that folks need to rebuild their workspace to resolve? I had to reinstall the dependencies & applications after rebuilding the workspaces also and am hoping there will not be any issues or conflicts when git adding, git committing and/or git pushing to the new workspace

pawlean:
Hi folks! This issue seems to have been resolved now. If you are experiencing any other issues, feel free to comment here. 🙏🏼 Thank you all for the support!

mjjstockman:
Hi pawlean!
I still can't open my workspace https://gitpod.io/start/#mjjstockman-setlistshare-nee26x1ripa

[...]

PR #8372, opened 2022-02-22
[ws-manager] Add missing check to fix OOM error

princerachit:
## Description

We were bailing out on errors when we can retry or safely ignore the error:
1.[ Pod deletion error ](https://console.cloud.google.com/logs/query;cursorTimestamp=2022-02-20T13:10:45Z;query=%22madeline8-poetryheaven-kc336gn1fet%22%0Atimestamp%3D%222022-02-20T13:10:45Z%22%0AinsertId%3D%2240tnqexuhmn7prmb%22;timeRange=P7D?folder=true&organizationId=327752801919&project=workspace-clusters)- When a pod is already deleted, any attempt to delete it would result in an error. If the error is not found error then it is safe to ignore this error and retry our attempt to create a pod.
2. [Finalizer removal error](https://console.cloud.google.com/logs/query;cursorTimestamp=2022-02-21T09:01:23Z;query=%220772a8f6-80c0-4a7c-9f0c-326030f25454%22%0Atimestamp%3D%222022-02-21T09:01:23Z%22%0AinsertId%3D%227pv0x0zh6j5q0k69%22%0Atimestamp%3D%222022-02-21T09:01:23Z%22%0AinsertId%3D%227pv0x0zh6j5q0k69%22;timeRange=2022-02-21T09:00:54.785Z%2F2022-02-21T10:30:54.785Z?project=workspace-clusters) - If a pod gets deleted before we attempt to explicitly remove the finalizer OR the pod object is changed before we attempt to remove the finalizer, then update would fail. In these cases we should check (a) the pod still exists and (b) retry updating it.

## Related Issue(s)

🤞🏾  Hopefully Fixes https://github.com/gitpod-io/gitpod/issues/8238

## How to test

We are not able to reproduce this issue in a non prod env so the only way to validate if this fixes the problem is to deploy in prod and monitor.

## Release Notes

```release-note
None
```

### DIFF
--- a/components/ws-manager/pkg/manager/manager.go
+++ b/components/ws-manager/pkg/manager/manager.go
@@ -235,22 +235,37 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
 			retryErr = err
 
 			var tempPod corev1.Pod
-			getErr := m.Clientset.Get(ctx, types.NamespacedName{Namespace: pod.Namespace, Name: pod.Name}, &tempPod)
-			if getErr != nil {
-				clog.WithError(getErr).WithField("pod.Namespace", pod.Namespace).WithField("pod.Name", pod.Name).Error("was unable to get pod")
-				// pod doesn't exist, so we are safe to proceed with retry
-				return false, nil
+			finalizerRemoved := false
+			// We do below in loop as sometimes the pod might be already deleted or modified when we attempt to remove the finalizer
+			// so we first get the pod and then attempt to remove the finalizer
+			// in case the pod is not found in the first place, we return success
+			for i := 0; i < 3 && !finalizerRemoved; i++ {
+				getErr := m.Clientset.Get(ctx, types.NamespacedName{Namespace: pod.Namespace, Name: pod.Name}, &tempPod)
+				if getErr != nil {
+					if !k8serr.IsNotFound(getErr) {
+						// pod doesn't exist, so we are safe to proceed with retry
+						return true, nil
+					}
+					clog.WithError(getErr).WithField("pod.Namespace", pod.Namespace).WithField("pod.Name", pod.Name).Error("was unable to get pod")
+					// pod get call failed so we error out
+					return false, retryErr
+				}
+				// we successfully got the pod, now we attempt to remove finalizer
+				tempPod.Finalizers = []string{}
+				updateErr := m.Clientset.Update(ctx, &tempPod)
+				if updateErr != nil {
+					clog.WithError(updateErr).WithField("pod.Namespace", pod.Namespace).WithField("pod.Name", pod.Name).Error("was unable to remove finalizer")
+					continue
+				}
+				finalizerRemoved = true
 			}
-			tempPod.Finalizers = []string{}
-			updateErr := m.Clientset.Update(ctx, &tempPod)
-			if updateErr != nil {
-				clog.WithError(updateErr).WithField("pod.Namespace", pod.Namespace).WithField("pod.Name", pod.Name).Error("was unable to remove finalizer")
-				// failed to remove finalizer, we not going to be able to create a new pod, so bail out with retry error
+			// if even after retries we could not remove finalizers, then error out
+			if !finalizerRemoved {
 				return false, retryErr
 			}
 
 			deleteErr := m.Clientset.Delete(ctx, &tempPod)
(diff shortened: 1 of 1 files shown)


princerachit:
Marked this as draft because I saw other errors not related to pod deletion e.g. [finalizer removal](https://console.cloud.google.com/logs/query;cursorTimestamp=2022-02-21T09:01:23Z;query=%220772a8f6-80c0-4a7c-9f0c-326030f25454%22%0Atimestamp%3D%222022-02-21T09:01:23Z%22%0AinsertId%3D%227pv0x0zh6j5q0k69%22;timeRange=2022-02-21T09:00:54.785Z%2F2022-02-21T10:30:54.785Z?project=workspace-clusters)

sagor999:
>When the pod does not exist and we try to delete the pod, we still get an error. 

hm. that was kinda intentional. since if we fail to delete, then we not going to be able to create new one.
did you see this error in logs?

also I would assume it would fail with delete error, and not out of memory error in this case.

princerachit:
> > When the pod does not exist and we try to delete the pod, we still get an error.
> 
> hm. that was kinda intentional. since if we fail to delete, then we not going to be able to create new one. did you see this error in logs?

If the pod does not exist then we should not be erroring out. Unsure why that would be intentional.


> 
> also I would assume it would fail with delete error, and not out of memory error in this case.

In the logs I can see that it says deletion failed and returns the pod ran to completion error. However, in the DB I still see the OOM error. I am not very sure where is that being written from.

components/ws-manager/pkg/manager/manager.go
```diff
@@ -235,22 +235,37 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
...
+						return true, nil
+					}
+					clog.WithError(getErr).WithField("pod.Namespace", pod.Namespace).WithField("pod.Name", pod.Name).Error("was unable to get pod")
+					// pod get call failed so we error out
+					return false, retryErr
+				}
+				// we successfully got the pod, now we attempt to remove finalizer
+				tempPod.Finalizers = []string{}
+				updateErr := m.Clientset.Update(ctx, &tempPod)
+				if updateErr != nil {
```
Furisto, line 256: Should we also check here if the pod does not exist because it has been deleted in the meantime?

princerachit:
> my concern here is why it was deleted already. or which service deleted it

That is a valid question, and I don't have answer to that.

components/ws-manager/pkg/manager/manager.go
Furisto: I mean is it possible that between getting the pod and updating the pod, someone deletes it? Then the update would fail because it does not exist anymore.

princerachit: I am not sure if that should happen, although I saw error messages complaining about metadata version mismatch where the existing version was empty string IIRC. So I have handled that case here.

components/ws-manager/pkg/manager/manager.go
```diff
@@ -235,22 +235,37 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
...
+			finalizerRemoved := false
+			// We do below in loop as sometimes the pod might be already deleted or modified when we attempt to remove the finalizer
+			// so we first get the pod and then attempt to remove the finalizer
+			// in case the pod is not found in the first place, we return success
+			for i := 0; i < 3 && !finalizerRemoved; i++ {
+				getErr := m.Clientset.Get(ctx, types.NamespacedName{Namespace: pod.Namespace, Name: pod.Name}, &tempPod)
+				if getErr != nil {
+					if !k8serr.IsNotFound(getErr) {
+						// pod doesn't exist, so we are safe to proceed with retry
+						return true, nil
```
sagor999, line 247: need to return `false, nil` here, as otherwise we will bail out from exponential backoff and exit without recreating the pod

sagor999 (changes requested):

components/ws-manager/pkg/manager/manager.go
sagor999: added the check.

components/ws-manager/pkg/manager/manager.go
sagor999: fixed.

sagor999 (approved):

Issue #8340, opened 2022-02-21
Track snapshot usage with success and access-error events


jldec:
This is a followup issue to #8257.

Since this change may break some existing workflows, it would be good to add usage tracking for workspace starts with snapshot URLs so that we can monitor both success and (importanly) the error cases.

jldec:
@jankeromnes do you know if there is already a log event we could track for these errors (in which case the urgency of this issue may be lower?)

jankeromnes:
I think there is no such log event. Will try to send a quick Pull Request here.

PR #8374, opened 2022-02-22
[server] Track snapshot access requests, and whether they are granted or denied

jankeromnes:
## Description

Track snapshot access requests, and whether they are granted or denied.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/8340

## How to test

1. Verify that the analytics messages are being sent (but I'm not sure how to do this?)

## Release Notes

```release-note
NONE
```

## Documentation

### DIFF
--- a/components/server/src/workspace/gitpod-server-impl.ts
+++ b/components/server/src/workspace/gitpod-server-impl.ts
@@ -836,6 +836,8 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
             ]);
 
             if (SnapshotContext.is(context)) {
+                // TODO(janx): Remove snapshot access tracking once we're certain that enforcing repository read access doesn't disrupt the snapshot UX.
+                this.trackEvent(ctx, { event: "snapshot_access_request", properties: { snapshot_id: context.snapshotId } }).catch();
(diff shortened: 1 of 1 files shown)


jldec (approved):

jldec:
Thanks for adding this @jankeromnes

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -836,6 +836,8 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
             ]);
 
             if (SnapshotContext.is(context)) {
+                // TODO(janx): Remove snapshot access tracking once we're certain that enforcing repository read access doesn't disrupt the snapshot UX.
+                this.trackEvent(ctx, { event: "snapshot_access_request", properties: { snapshot_id: context.snapshotId } }).catch();
```
jakobhero, line 840: we try to bundle track calls that follow the same context in one call and have differences contained in the properties. in order to follow this here, it would make sense to give the same name to all three track calls (e.g. `snapshot_access`; we try to follow an `object_verb` naming convention from the perspective of the user, feel free to come up with something here that follows it more closely). then, you can add a property such as `type` that contains whether the event is `requested`, `granted`  or `denied` so that we still capture the same level of information.

else, this looks good to me!

PR #8386, opened 2022-02-22
[admin] (temp) disable telemetry for gitpod.io

AlexTugarev:
```release-note
NONE
```

### DIFF
--- a/components/dashboard/src/admin/Settings.tsx
+++ b/components/dashboard/src/admin/Settings.tsx
@@ -22,11 +22,14 @@ export default function Settings() {
     const { user } = useContext(UserContext);
 
     useEffect(() => {
+        if (isGitpodIo()) {
+            return; // temporarily disable to avoid hight CPU on the DB
+        }
         (async () => {
             const data = await getGitpodService().server.adminGetTelemetryData();
             setTelemetryData(data)
         })();
-    });
+    }, []);
 
     if (!user || !user?.rolesOrPermissions?.includes('admin')) {
         return <Redirect to="/"/>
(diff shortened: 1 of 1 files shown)


roboquat:

<details>

</details>

MrSimonEmms (approved):

components/dashboard/src/admin/Settings.tsx
```diff
@@ -22,11 +22,14 @@ export default function Settings() {
     const { user } = useContext(UserContext);
 
     useEffect(() => {
+        if (isGitpodIo()) {
+            return; // temporarily disable to avoid hight CPU on the DB
```
laushinka, line 26: Do we still need to disable it since the missing dependency arg is what's causing the infinite loop?

Pothulapati: I think yes, because we should not be giving our  users insights or the permission to see/set this page for the whole SAAS instance? 🤔

AlexTugarev: It wasn't causing an infinite loop. 😌 It's just on each re-render it was called, instead of a single init call.

corneliusludmann: Not sure, if it's only the missing dependency arg. It could also be that this query is too expensive with our large instance DB:

https://github.com/gitpod-io/gitpod/blob/5f6114c2823997874352f8cd8f2207fef1253f29/components/gitpod-db/src/typeorm/workspace-db-impl.ts#L337-L344

Don't know how to figure that out, though.

AlexTugarev: You are right. The useEffect issue was not a big deal, it just amplified the actual problem.

corneliusludmann (approved):

MrSimonEmms (approved):

JanKoehnlein (approved):
LGTM

PR #8395, opened 2022-02-22
[kots]: add the KOTS installation manifests

MrSimonEmms:
## Description

This incorporates the Replicated/KOTS installation into the installation section of the repo. KOTS is a way of allowing us to manage how enterprises install our software

The change to `.gitpod.yml` is to install the [Replicated CLI](https://kots.io/vendor/cli/getting-started/) (to publish the manifests) and the [KOTS](https://kots.io/) (to install Gitpod with KOTS) plugin

### Tested on:
- [ ] ~AWS in-cluster~
- [ ] ~AWS external~
- [x] Azure in-cluster
- [x] Azure external
- [x] GCP in-cluster
- [x] GCP external 

**NB** this does not install the external dependencies, but merely that it provisions the application in those environments

I'm proposing avoiding the AWS testing at this stage until we establish whether the errors with the `2022.02.0` release are caused by this release or (more likely) an error in the AWS setup

## Related Issue(s)

Fixes #

## Release Notes

```release-note
[kots]: add the KOTS installation manifests
```


roboquat:

<details>

</details>

install/kots/Makefile
```diff
@@ -0,0 +1,20 @@
...
+.PHONY: create_unstable_release
+
+lint:
+	replicated release lint --yaml-dir ${YAML_DIR}
+.PHONY: lint
+
+helm:
+	@echo "Installing Helm dependencies"
+	@rm -f manifests/*.tgz
+	@for f in $(shell ls -d charts/*/); do cd $${f} && helm dep up && helm package . --destination ../../manifests && cd -; done
```
corneliusludmann, line 19: Instead of this helm magic, couldn't we just download the cert-manager package like this?

```
$ curl -sSLO https://charts.jetstack.io/charts/cert-manager-v1.7.0.tgz
```

Or wouldn't this work for some reason?

MrSimonEmms: That would work, but I've done it as a general function in case we ever put a second Helm chart in there - we use `external-dns` in all the guides which might be worth offering in Kots in future

corneliusludmann (changes requested):

install/installer/pkg/components/gitpod/configmap.go
```diff
@@ -24,7 +25,7 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		VersionManifest: ctx.VersionManifest,
 	}
 
-	config, err := common.ToJSONString(ctx.Config)
+	cfg, err := config.Marshal(config.CurrentVersion, ctx.Config)
```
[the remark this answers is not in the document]
MrSimonEmms: Arguably not. I spotted it and fixed it - I'll put it into a separate PR

install/kots/Makefile
MrSimonEmms: Right, I think that my original approach is the correct one. Using the `.tgz` file provided by cert-manager has everything nested in a `cert-manager` folder. For KOTS to work, it needs to be in the root of the `.tgz` file.

So, that makes there two ways of achieving this:
1. create a noddy `Chart.yaml` file and use the `helm package` function
2. download the `.tgz` file, extract it, re-compress it with the `cert-manager` and use that new `.tgz` file

Whilst both have their limitations, I think approach `1` is the simpler and more maintainable one

corneliusludmann (approved):

Issue #8267, opened 2022-02-17
[installer] Make all config values with sensible defaults optional


corneliusludmann:
In my opinion, we should make all config values that have sensible default values optional in the config file. E.g., it should be totally fine to use a config like this:
```yaml
apiVersion: v1
domain: gitpod.my-domain.example
```

That would also make generating config values for e.g. replicated easier.

_see also https://github.com/gitpod-io/gitpod/issues/8266#issuecomment-1042696678_

PR #8401, opened 2022-02-23
installer: always override `-c` config on default values 

Pothulapati:
## Description

Currently, Users run `init` command first, update it and
pass full config to `render` to generate Kubernetes manifests. The
passage of `-c` is a requirement here, and users can't skip it.

This PR makes the passage of `config` to `render` optional, and
flexible. This means
- Users can skip `-c` entirely, in which case we use the default
  values on the default config version for that installer binary
- Users can selectively override fields and *thus no need to pass full
  config* all the time. This means `-c` flag acts as a flag through
  which they can override the default fields.

For the second case, When a user explicitely sets the `apiVersion`
field in the passed config, we use the default values for that
version. If no `apiVersion` is passed, we override the passed config
onto the default values on the default config version for that
installer binary.

After this change, For users This means that they only store and use
the config that has their changes only (on the default config), and
not the entire config.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/8267

## How to test

Build the binary, with a speciifc version

```bash
VERSION="main.2514" make build
```

### Run `render` no `-c` passed
```bash
./dist/gitpod-installer render
```

### Run `render` with only `domain` set

```bash
echo "domain: gitpod-test.com" > config.yaml

./dist/gitpod-installer render -c ./config.yaml | grep gitpod-test
```

## Release Notes

```release-note
Make `-c` optional in installer, while allowing the passed config to be flexible
```


MrSimonEmms:
I think the `domain` is something that is must always be set - because no one will be running the app from `gitpod.example.com`, I think we have to enforce this property.

corneliusludmann (approved):
Works like a charm! Thanks Tarun! :rocket:

PR #7882, opened 2022-01-27
Allow project search on admin dashboard

laushinka:
## Description
1. Allows searching for projects
2. Shows project detail
3. Shows list of prebuilds for each project

| Search result | Detail with prebuilds | Detail without prebuilds
| - | - | - |
| Screenshot 2022-01-31 at 02 47 33 | Screenshot 2022-01-31 at 18 58 40 | Screenshot 2022-01-31 at 18 55 24 | 

## Related Issue(s)
Fixes #7636 and #5022

## How to test
1. Go to https://laushinka-search-project-7636.staging.gitpod-dev.com/admin/projects
2. To search all projects, click on the "Search" button without any search queries.
3. To search specific projects, enter the search query.
4. Once there are result(s), click on one of the rows to go to the project detail.

## Release Notes

```release-note
Admins can do a text search for projects and their associated prebuilds.
```

PR #8402, opened 2022-02-23
Laushinka/team admin 8060

### LINKED ISSUE #8060: [Admin] Search for team by name or slug and show team details
From discussion[[1](https://www.notion.so/gitpod/WebApp-Notes-Admin-for-Teams-Laurie-J-rgen-Feb-10-2022-f75d2e6f886a44ceb23bc57c419ab28c)](internal):

Search for teams
- by name
- by slug (helpful for team names with funny characters)

- Results - order by created desc
    - [x] Name + team URL (/t/slug) underneath
    - [x] Created date (order by desc) show arrow in heading
    - [x] Indicate deleted (nice to have in admin lists) 
- [x] Click on any result (entire row is clickable) to go to team detail

Team detail
- [x] Team name + team URL at the top
- Other fields:
    - [x] created date (either friendly with actual date/time on hover, or actual date/time)
    - [x] marked deleted
    - [x] member count
    
- Member list 
  - [x] (Not required for deleted teams)
  - [x] List columns: Member, Joined, Role
  - [x] Paginated would be nice OR ok to limit list + Search for now
  - [x] Click to link to user detail
  - [x] Nice to have actions: change role (useful for support to assign new owner)

Future
- More actions
  - Rename (should be possible for non-admin users as well - complexity with slug rename)
  - Delete 
  - Invite URL (@laushinka moved this from nice-to-have to a future task)
- Project list
  - toggled with member list using tab-like component

laushinka:
## Description
Allows managing teams as admin.

| SEARCH RESULT | TEAM DETAIL | DELETED TEAM |
| - | - | - |
Screenshot 2022-02-24 at 16 25 38 | Screenshot 2022-02-24 at 12 28 13 | Screenshot 2022-02-24 at 13 34 29 |

## Related Issue(s)

Fixes #8060 

## How to test
https://laushinka-team-admin-8060.staging.gitpod-dev.com/
1. Create a team with member(s)
2. Visit `/admin/teams` and search for teams

## Release Notes

```release-note
Admins can find teams, see team details, and change a team member's role.
```


roboquat:

<details>

</details>

laushinka:
@jldec @gtsiolis This is still a draft, but feel free to play around so we can detect issues.

gtsiolis:
Looking at this now! 👀

components/dashboard/src/admin/TeamDetail.tsx
```diff
@@ -0,0 +1,67 @@
...
+            }
+        })();
+    }, [team]);
+
+     return <>
+        <div className="flex">
+            <div className="flex-1">
+                <div className="flex"><h3>{team.name}</h3><span className="my-auto">{team.markedDeleted ? " Deleted" : ""}</span></div>
+                <p>{teamMembers && teamMembers.length} member(s)</p>
+                <p>Created {moment(team.creationTime).format('MMM D, YYYY')}</p>
```
gtsiolis, line 31: **question**: Should we align the creation timestamp information structure on a) teams list and the b) team page? For example, we show _Created Feb 23, 2022_ and _a few seconds ago_ for the same entry. Happy to move forward with this small inconsistency. In the first case, let's add an adverb, like _Created on Feb 23, 2022_. What do you think?

gtsiolis (changes requested):
Thanks for opening this PR, @laushinka! 🌟 

UX looks good. Left some comments below that are complementary to what @jldec added in https://github.com/gitpod-io/gitpod/issues/8060#issuecomment-1048859077.

components/dashboard/src/App.tsx
```diff
@@ -298,6 +299,7 @@ function App() {
                 <Route path="/from-referrer" exact component={FromReferrer} />
 
                 <Route path="/admin/users" component={UserSearch} />
+                <Route path="/admin/teams" component={TeamsSearch} />
```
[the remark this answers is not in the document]
laushinka: Ah, I actually put it under `Users` because I thought Users and Teams group nicely as they're closely related?

components/dashboard/src/admin/TeamDetail.tsx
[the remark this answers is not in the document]
laushinka: Yes! Was just doing a rough draft but planned to make it as you designed :)

laushinka: Oops good catch, thanks! Will fix the one on the search result to also be a specific date, and add "on".

components/dashboard/src/App.tsx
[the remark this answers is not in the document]
gtsiolis: Here's what I had in mind:
1. Most of these are closely related except for Settings, but Teams & Projects have a stronger connection as this feature shipped as one piece[[1](https://www.gitpod.io/changelog/introducing-teams-&-projects)].
2. We've been deliberately using the term _Users_ here instead of _Members_ to include all users across all teams in the instance which creates a small disconnection from Teams & Projects hence the suggestion to move _Teams_ a bit downwards and closer to _Projects_.
3. Teams is also a larger container as a concept that includes projects, members, billing, etc. Cross-linking relevant FigJam file[[2](https://www.figma.com/file/V2p8Rd7swkV3swykNZIHGv/Projects---Prebuilds---Teams-(Brainstorm)?node-id=0%3A1)] (screenshot below) while discussing the introduction of Teams & Projects for more context.

| Early draft for the entity relationship diagram for the dashboard |
|-|
| image |

However, no need to hold this back because of this, @laushinka please move forward with any option that looks better to you. It's a small change with little to no negative impact, any option will suffice and we can easily change it later if needed. Your call. 🏀

components/dashboard/src/admin/TeamDetail.tsx
gtsiolis: @laushinka Long-term I'd opt for going with the other option to avoid duplication of information (_Created on ..._) as this is also repeated on the table header. This is what we've practiced so far in other areas of the dashboard. You call!

laushinka: I talked about this earlier with @jldec and opted for the specific date because it would be more useful for the admin to know when a team was created. On other cases I can see where we would like to avoid duplication, but for debugging purposes the duplication would help for both when navigating the teams search result and visiting the team detail 🙏🏽

components/dashboard/src/App.tsx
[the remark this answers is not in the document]
laushinka: Ah that's right, I forgot that "Teams & Projects" was how we packaged it. It makes sense, will do the change :)

components/dashboard/src/admin/TeamDetail.tsx
```diff
@@ -0,0 +1,67 @@
...
+            if (members.length > 0) {
+                setTeamMembers(members)
+            }
+        })();
+    }, [team]);
+
+     return <>
+        <div className="flex">
+            <div className="flex-1">
+                <div className="flex"><h3>{team.name}</h3><span className="my-auto">{team.markedDeleted ? " Deleted" : ""}</span></div>
```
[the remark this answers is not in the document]
laushinka: Changed it to be the deleted label

laushinka:
@jldec @gtsiolis Ready for review (not a draft anymore).
Also now that we have a slug under the team name in the team detail, would be nice to get input about the layout 🙏🏽

gtsiolis:
Looking at this now! 👀

components/dashboard/src/admin/TeamDetail.tsx
gtsiolis: DEAL. Fair point. 🤝

components/dashboard/src/admin/TeamsSearch.tsx
```diff
@@ -0,0 +1,118 @@
...
+            {searchResult.rows.map(team => <TeamResultItem team={team} />)}
+        </div>
+    </>
+
+function TeamResultItem(props: { team: Team }) {
+    return (
+        <Link key={'pr-' + props.team.name} to={'/admin/teams/' + props.team.id} data-analytics='{"button_type":"sidebar_menu"}'>
+            <div className="rounded-xl whitespace-nowrap flex py-6 px-6 w-full justify-between hover:bg-gray-100 dark:hover:bg-gray-800 focus:bg-gitpod-kumquat-light group">
+                <div className="flex flex-col w-7/12 truncate">
+                    <div className="font-medium text-gray-800 dark:text-gray-100 truncate">{props.team.name}
```
gtsiolis, line 107: **question**: Could we include the members count below the team name here as described in the specs[[1](https://github.com/gitpod-io/gitpod/issues/8060#issuecomment-1047869502)]? Is this an expensive operation or did we decide to drop it?

Asking as :a: this can provide a visual cue when searching teams and :b: showing members below the team name is a pattern we already use in the team scope selector (dropdown) in the top navigation as well as the team selection on the onboarding wizard.

| Team scope selector (dropdown) | Team selection on the onboarding wizard |
|-|-|
| Screenshot 2022-02-24 at 12 20 19 PM | Screenshot 2022-02-24 at 12 20 45 PM |

components/dashboard/src/admin/TeamDetail.tsx
```diff
@@ -0,0 +1,105 @@
...
+        await getGitpodService().server.adminSetTeamMemberRole(team!.id, userId, role);
+        setTeamMembers(await getGitpodService().server.adminGetTeamMembers(team!.id));
+    }
+    return <>
+        <div className="flex">
+            <div className="flex-1">
+                <div className="flex"><h3>{team.name}</h3>
+                    {team.markedDeleted && <span className="mt-2"><PillLabel type="warn" className="font-semibold mt-2 py-0.5 px-2 self-center">Deleted</PillLabel></span>}
+                </div>
+                <p className="mb-6">/t/{team.slug}</p>
```
gtsiolis, line 48: **question**: Is there any value in including the `/t/` prefix here? 💭

components/dashboard/src/admin/TeamDetail.tsx
```diff
@@ -0,0 +1,105 @@
...
+                {!team.markedDeleted && <>
+                    <span className="text-gray-400">{teamMembers && teamMembers.length} member(s)</span>
```
gtsiolis, line 50: **question**: Could we use a ternary option here like we do for other instances of the members count?

https://github.com/gitpod-io/gitpod/blob/e9cd64489335d9bfd43682059400fef617c50473/components/dashboard/src/Menu.tsx#L214

components/dashboard/src/admin/TeamDetail.tsx
```diff
@@ -0,0 +1,105 @@
...
+                    <span className="text-gray-400"> · </span>
+                </>}
+                <span className="text-gray-400">Created on {moment(team.creationTime).format('MMM D, YYYY')}</span>
```
gtsiolis, line 53: **suggestion**: What do you think of moving the create date metadata near the URL slug line? We could also skip members count in the detail page for now. Long-term I'd love to have one pattern that repeats consisntenly whteher this includes the URL slug or not.

Screenshot 2022-02-24 at 12 27 02 PM

components/dashboard/src/admin/TeamDetail.tsx
```diff
@@ -0,0 +1,105 @@
...
+            </div>
+        </div>
+        <div className="flex mt-4">
+            <div className="flex">
+                <div className="py-4">
+                    <svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 16 16" width="16" height="16"><path fill="#A8A29E" d="M6 2a4 4 0 100 8 4 4 0 000-8zM0 6a6 6 0 1110.89 3.477l4.817 4.816a1 1 0 01-1.414 1.414l-4.816-4.816A6 6 0 010 6z" /></svg>
+                </div>
+                <input type="search" placeholder="Search Members" onChange={e => setSearchText(e.target.value)} />
```
gtsiolis, line 61: **suggestion**: Could we include an empty state here for the deleted teams that include no members as we did with prebuilds in https://github.com/gitpod-io/gitpod/pull/7882#discussion_r795668199?

components/dashboard/src/admin/TeamDetail.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+    const setTeamMemberRole = async (userId: string, role: TeamMemberRole) => {
+        await getGitpodService().server.adminSetTeamMemberRole(team!.id, userId, role);
+        setTeamMembers(await getGitpodService().server.adminGetTeamMembers(team!.id));
+    }
+    return <>
+        <div className="flex">
+            <div className="flex-1">
+                <div className="flex"><h3>{team.name}</h3>
+                    {team.markedDeleted && <span className="mt-2"><PillLabel type="warn" className="font-semibold mt-2 py-0.5 px-2 self-center">Deleted</PillLabel></span>}
```
gtsiolis, line 46: **issue(non-blocking)**: While using a label like this here sounds great, we could use the same design for the blocked user labed as described in https://github.com/gitpod-io/gitpod/issues/8060#issuecomment-1047869502. However, it's also fine to improve in one area and let the previous design catch-up.

**question**: Wondering how long it takes for soft-deleted items to disappear.

gtsiolis (changes requested):
Thanks for updating this and addressing all comments, @laushinka! 

Left some second review round comments below. 🥊

components/dashboard/src/admin/TeamDetail.tsx
laushinka: 1. Will change it to the same design as the Blocked label.
2. Teams, and also other entities in our project, are not technically "soft"-deleted but only "marked" as deleted. The rows will stay in the database, but they will be excluded from being shown. So the answer is, they won't disappear.

components/dashboard/src/admin/TeamDetail.tsx
laushinka: Yes - I tried this as well yesterday. What text should we show? "Members in deleted teams have been removed"?

components/dashboard/src/admin/TeamDetail.tsx
laushinka: From my understanding @jldec wanted to show the path to the team for debugging purposes, e.g. an admin can share that to the team owner.

components/dashboard/src/admin/TeamDetail.tsx
gtsiolis: :a: Going with a boring (=simple) solution here like the following sounds ok. What do you think?

> No members found.

:b: Alternatively, we could not remove any members and keep the last known state of members.

laushinka: I'll go with A, since that's also the text we show when we can't find any members. 
B is unfortunately not possible, because unlike the "marked-deletion" of a team, when a team is marked as deleted, we actually delete all the team members. I vaguely remember a discussion being made about team members being hard-deleted, but I can't remember what or why that was 🤔

components/dashboard/src/admin/TeamsSearch.tsx
laushinka: I opted to not do it here because since it's an extra call and I wanted to make sure the search result gets shown as fast as possible. In the top navigation, the dropdown isn't actually immediately shown - only when we do click on it, so there's time for it to load. I'm okay with including this if it's necessary for this iteration. Let me know @gtsiolis @jldec

gtsiolis: @laushinka If we are not paginating this list, which I think this is the case here (right?), let's skip the members count.

laushinka: I'm trying this now, since I could be optimizing early with my above thought. Coming soon!

laushinka: I wrote this comment before the page loaded with your new comment 🤣

laushinka: @gtsiolis I ended up pushing a commit with the members count. Let me know what you think once the preview env is built - I didn't rebase it with the previous one so it's easily revertible in case you'd like me to get rid of it 🤗

laushinka:
Love this 🤣 
Screenshot 2022-02-24 at 16 21 04

components/dashboard/src/admin/TeamsSearch.tsx
```diff
@@ -0,0 +1,138 @@
...
+        </div>
+    </>
+
+function TeamResultItem(props: { team: Team, members: TeamMemberInfo[]}) {
+    const { team, members } = props;
+    return (
+        <Link key={'pr-' + team.name} to={'/admin/teams/' + team.id} data-analytics='{"button_type":"sidebar_menu"}'>
+            <div className="rounded-xl whitespace-nowrap flex py-6 px-6 w-full justify-between hover:bg-gray-100 dark:hover:bg-gray-800 focus:bg-gitpod-kumquat-light group">
+                <div className="flex flex-col w-7/12 truncate">
+                    <div className="font-medium text-gray-800 dark:text-gray-100 truncate">{team.name}
```
gtsiolis, line 126: **issue**: Although the members count creates some visual balance, I'm already noticing a small delay in rendering with 16 teams on a not-so-fast network so I'm inclined to suggest dropping it for now following the concerns in https://github.com/gitpod-io/gitpod/pull/8402#discussion_r813946261. Let's cut some corners! ✂️ 

https://user-images.githubusercontent.com/120486/155563962-ed700cf2-cc8e-4bdb-b620-c4ac38149770.mov

components/dashboard/src/admin/TeamDetail.tsx
```diff
@@ -0,0 +1,95 @@
...
+        return true;
+    });
+
+    const setTeamMemberRole = async (userId: string, role: TeamMemberRole) => {
+        await getGitpodService().server.adminSetTeamMemberRole(team!.id, userId, role);
+        setTeamMembers(await getGitpodService().server.adminGetTeamMembers(team!.id));
+    }
+    return <>
+        <div className="flex">
+            <div className="flex-1">
```
gtsiolis, line 35: **issue:** Just run into an issue that affects both Projects and Teams in the admin dashboard. Certainly out of the scope of this PR, but when you reload the details page for a project or a team, parts of the information in the header disappears and does not load. This behaves also slightly different for deleted teams. Could you open a follow up issue for this if it's reproducible?

https://user-images.githubusercontent.com/120486/155565249-f8333bec-e5a1-47c5-aa90-ee9178d83c42.mov

components/dashboard/src/admin/TeamDetail.tsx
```diff
@@ -0,0 +1,95 @@
...
+                            <div className="text-base text-gray-900 dark:text-gray-50 font-medium">{m.fullName}</div>
+                            <p>{m.primaryEmail}</p>
+                        </div></Link>
+                    </ItemField>
+                    <ItemField className="my-auto">
+                        <span className="text-gray-400">{moment(m.memberSince).fromNow()}</span>
+                    </ItemField>
+                    <ItemField className="flex items-center my-auto">
+                        <span className="text-gray-400 capitalize">
+                            <DropDown contextMenuWidth="w-32" activeEntry={m.role} entries={[{
```
gtsiolis, line 83: **question**: Did we decide to bring the role change action back?

components/dashboard/src/admin/TeamsSearch.tsx
```diff
@@ -0,0 +1,138 @@
...
+                    {team.markedDeleted && <Label text='Deleted' color="red"/>}
+                    {members?.length > 0 && <p>{`${members?.length} ${members?.length > 1 ? `members` : `member`}`}</p>}
```
gtsiolis, line 128: **thought**: I still think the members count information is valuable but the current design could be slightly improved. What do you think of re-using the tabular pattern from projects, users, and workspaces here even if this is just a single column information?

| BEFORE | AFTER |
|-|-|
| members-before | members-after |

gtsiolis (changes requested):
Thanks, @laushinka! Round three! 🥊

components/dashboard/src/admin/TeamsSearch.tsx
gtsiolis: See https://github.com/gitpod-io/gitpod/pull/8402#discussion_r814045854.

components/dashboard/src/admin/TeamsSearch.tsx
laushinka: Yass

components/dashboard/src/admin/TeamDetail.tsx
laushinka: Yes, here as a nice to have for support issues[[1](https://github.com/gitpod-io/gitpod/issues/8060#issue-1125731692)]

components/dashboard/src/admin/TeamDetail.tsx
laushinka: This confused me too - I checked on prod and it's not reproducible there. I wonder if there are websocket connection limitations when it comes to preview envs 🤔

laushinka: Also not reproducible on staging. I'll make a note to check this on staging after merging, and if it's reproducible, I'll create a follow-up issue 🙏🏽

gtsiolis (dismissed):
UX changes look great, @laushinka!

Last round! 🥊 

One outstanding question is whether we'd like to include the linking back to the team from the project detail page in this PR or open a follow up issue for this. Your call. 🏓 

| Project in personal account | Project in team |
|-|-|
| project-user | project-team |

laushinka:
> One outstanding question is whether we'd like to include the linking back to the team from the project detail page in this PR or open a follow up issue for this. Your call. 🏓

@gtsiolis Thank you for remembering this! Will do it, since I'm still waiting for code reviews anyway 👏🏼

gtsiolis (approved):

jldec (approved):
Very nice work - thank you @laushinka

Issue #8375, opened 2022-02-22
High performance impact of server.trackEvent -- let's re-evaluate when it gets called


jankeromnes:
### Bug description

The API method `server.trackEvent()` is being called too frequently, putting excessive load on our servers.

It would be good to:
- [ ] find out which events are tracked the most frequently
- [ ] ... and re-evaluate whether they're still useful, or if we can stop tracking them
- [ ] ... if they're still useful, see if we can send them to our analytics backend (segment) directly instead of routing them through `server`

`server.trackEvent()` seems to be called most often by the dashboard and by the IDE backends (both VS Code and JetBrains).

---

From [Slack](https://gitpod.slack.com/archives/C02EN94AEPL/p1645515660492929?thread_ts=1645515391.622969&cid=C02EN94AEPL) (internal):

> `trackEvent` is by far our most called API method (even beating `getWorkspace`  - and has 2x the callrate of `sendHeartbeat` :grimacing: )
> 
> It puts considerable load on our server instances.

From [Grafana](https://grafana.gitpod.io/d/server/gitpod-component-server?orgId=1) (internal):

<img width="600" alt="Screenshot 2022-02-22 at 09 20 49" src="https://user-images.githubusercontent.com/599268/155091718-b45f15cd-2671-4aa2-8728-03fb2e80dd35.png">

### Steps to reproduce

1. Look at the API call rates in Grafana

AlexTugarev:
I've created a draft (https://github.com/gitpod-io/gitpod/pull/8388) it doesn't aim to solve this issue entirely, but maybe it already reduces the load significantly enough.

JanKoehnlein:
@akosyakov was concerned about rate-limits when we send data to segment directly. I don't think [60M events in 10min](https://segment.com/docs/connections/rate-limits/) are a problem yet.

PR #8418, opened 2022-02-23
WIP testing: remove track event vscode_session phase running and end

mustard-mh:
## Description

Remove track event `vscode_session` phase `running` and `end`, see changes here https://github.com/gitpod-io/openvscode-server/commit/a8cf2c01c89debd3fc8767176ad7ff2589ee12d4

## Related Issue(s)

Related #8375

## How to test

1. Use vscode insiders
2. Open a workspace
3. Go to segment `staging_untrusted` filter with `vscode_session`, you should see a track without `phase` and `focused` props

## Release Notes

```release-note
NONE
```

## Documentation

### DIFF
--- a/WORKSPACE.yaml
+++ b/WORKSPACE.yaml
@@ -7,7 +7,7 @@ defaultArgs:
   jbMarketplacePublishTrigger: "false"
   publishToJBMarketplace: true
   localAppVersion: unknown
-  codeCommit: 8428da4b36e80a122af3c3e2b58b1fcffb5d5f8e
+  codeCommit: a8cf2c01c89debd3fc8767176ad7ff2589ee12d4
   intellijDownloadUrl: "https://download.jetbrains.com/idea/ideaIU-2021.3.2.tar.gz"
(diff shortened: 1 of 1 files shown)


jeanp413:
LGTM, we''ll cherry pick it to release branch and do a stable release tomorrow cc @akosyakov

akosyakov:
> Edit: thinking about a later improvement, wouldn't it be better to decouple the gitpod extension from the gp-code release like codespaces extension and publish it to openvsx?

We will depend even more on OpenVSX availability then? Generally yes if it makes it easier to roll out and does not cause any other side effect.

akosyakov:
> Edit: thinking about a later improvement, wouldn't it be better to decouple the gitpod extension from the gp-code release like codespaces extension and publish it to openvsx?

We will depend even more on OpenVSX availability then? Generally yes if it makes it easier to roll out and does not cause any other side effect.

akosyakov:
> Edit: thinking about a later improvement, wouldn't it be better to decouple the gitpod extension from the gp-code release like codespaces extension and publish it to openvsx?

We will depend even more on OpenVSX availability then? Generally yes if it makes it easier to roll out and does not cause any other side effect.

PR #8435, opened 2022-02-24
Support custom docker daemon args

### LINKED ISSUE #8103: [docker] Support arguments to Docker daemon

Furisto:
## Description
This allows users to add arguments to the docker daemon running in the workspace. As a first step mapping the container user to the gitpod user is supported. 

## Related Issue(s)
Fixes #8103

## How to test
- Add environment variable DOCKER_DAEMON_ARGS to your user settings. The content should look like this:
`{ "remap-user": "1000" }` (Means container id 1000 will be mapped to gitpod user). 
- Start workspace
- mkdir /tmp/me
- Run docker image which supports running as unprivileged user e.g. `sudo docker run -it -u node -v /tmp/me:/tmp node sh`
- Create file within container e.g. touch /tmp/foo
- Verify that the created file has gitpod:gitpod permissions in the workspace

## Hints for reviewers
- The uidmap package needs to be installed as docker needs to create multiple user mappings which require newuidmap/newgidmap
- The most recent version of runc needs to be installed as otherwise mounting cgroupfs will fail with when starting the container
- Both dependencies will only be installed if the DOCKER_DAEMON_ARGS environment variable has been so that the startup performance of the first container will not be impacted for users that do no require a mapping.
- This does not allow specifying the dockerd argument (userns-remap) directly as this would mean the user would specify ("userns-remap" : "gitpod"). We already know that the user wants to map the container user to gitpod so this is redundant. What we do not get this way is the container user that should be mapped, which means we cannot setup the /etc/subuid and /etc/subgid files for them and they would have to do it themselves. 
- Instead they can just specify the user they want to be mapped (e.g. 1000) and we ensure that this will be mapped to the gitpod user.
- I considered modifying runc-facade instead of adding this to docker-up, but just adding the user mapping to the spec will cause file system permission errors because the rootfs needs to be chowned. This way docker will do it for us. 

## Release Notes

```release-note
Docker in workspaces now allows mapping the user id of a container user to the workspace gitpod user
```

## Documentation
Have not added documentation before, so any advice is appreciated.

### DIFF
--- a/components/docker-up/dependencies.sh
+++ b/components/docker-up/dependencies.sh
@@ -10,3 +10,4 @@ DOCKER_COMPOSE_VERSION=1.29.2
 
 curl -o docker.tgz      -fsSL https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_VERSION}.tgz
 curl -o docker-compose  -fsSL https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-Linux-x86_64
+curl -o runc            -fsSL https://github.com/opencontainers/runc/releases/download/v1.1.0/runc.amd64
--- a/components/docker-up/dependencies.sh
+++ b/components/docker-up/dependencies.sh
@@ -7,7 +7,8 @@ set -euo pipefail
 
 DOCKER_VERSION=19.03.15
 DOCKER_COMPOSE_VERSION=1.29.2
+RUNC_VERSION=v1.1.0
 
 curl -o docker.tgz      -fsSL https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_VERSION}.tgz
 curl -o docker-compose  -fsSL https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-Linux-x86_64
-curl -o runc            -fsSL https://github.com/opencontainers/runc/releases/download/v1.1.0/runc.amd64
+curl -o runc            -fsSL https://github.com/opencontainers/runc/releases/download/${RUNC_VERSION}/runc.amd64
--- a/components/docker-up/docker-up/main.go
+++ b/components/docker-up/docker-up/main.go
@@ -43,7 +43,10 @@ var opts struct {
 
 //go:embed docker.tgz
 //go:embed docker-compose
+//go:embed runc
 var binaries embed.FS
+
+var aptUpdated = false
 
 const (
 	dockerSocketFN = "/var/run/docker.sock"
@@ -171,6 +174,8 @@ var prerequisites = map[string]func() error{
 	"dockerd":        installDocker,
 	"docker-compose": installDockerCompose,
 	"iptables":       installIptables,
+	"uidmap":         installUidMap,
+	"runcV1.1.0":     installRunc,
 }
 
 func ensurePrerequisites() error {
@@ -259,20 +264,12 @@ func installDockerCompose() error {
 }
 
 func installIptables() error {
-	pth, _ := exec.LookPath("apt-get")
-	if pth != "" {
-		cmd := exec.Command("/bin/sh", "-c", "apt-get update && apt-get install -y iptables xz-utils")
-		cmd.Stdin = os.Stdin
-		cmd.Stderr = os.Stderr
-		cmd.Stdout = os.Stdout
-		cmd.SysProcAttr = &syscall.SysProcAttr{
-			Pdeathsig: syscall.SIGKILL,
-		}
-
-		return cmd.Run()
+	err := installPackages("iptables", "xz-utils")
+	if err != nil {
+		return xerrors.Errorf("could not install iptables: %w", err)
 	}
 
-	pth, _ = exec.LookPath("apk")
+	pth, _ := exec.LookPath("apk")
 	if pth != "" {
 		cmd := exec.Command("/bin/sh", "-c", "apk add --no-cache iptables xz")
 		cmd.Stdin = os.Stdin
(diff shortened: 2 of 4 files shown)


components/docker-up/dependencies.sh
```diff
@@ -10,3 +10,4 @@ DOCKER_COMPOSE_VERSION=1.29.2
 
 curl -o docker.tgz      -fsSL https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_VERSION}.tgz
 curl -o docker-compose  -fsSL https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-Linux-x86_64
+curl -o runc            -fsSL https://github.com/opencontainers/runc/releases/download/v1.1.0/runc.amd64
```
utam0k, line 13: How about making the version a variable like the others?
```suggestion
curl -o runc            -fsSL https://github.com/opencontainers/runc/releases/download/${RUNC_VERSION}/runc.amd64
```

utam0k:
First of all, thanks for the great PR.
Do you know why? I was a little curious.
> The most recent version of runc needs to be installed as otherwise mounting cgroupfs will fail with when starting the container

components/docker-up/docker-up/main.go
```diff
@@ -32,6 +35,8 @@ import (
 
 var log *logrus.Entry
 
+const DaemonArgs = "DOCKER_DAEMON_ARGS"
```
utam0k, line 38: How about this name? I feel this is simpler.
```suggestion
const DaemonArgs = "DOCKERD_ARGS"
```

components/docker-up/docker-up/main.go
```diff
@@ -161,10 +175,91 @@ func runWithinNetns() (err error) {
 	return nil
 }
 
+var allowedDockerArgs = map[string]string{
+	"remap-user": "userns-remap",
+}
+
+func setUserArgs(args []string) ([]string, error) {
```
utam0k, line 182: This function seemed simpler if it was dedicated to crate args from env vars from the user that were converted to dockerd args. In other words, instead of doing both generation and addition, why not just concentrate on generation?
```suggestion
func userArgs() ([]string, error) {
	args := []string{}
```

And,
```go
	uargs, err := userArgs()
	if err != nil {
		return xerrors.Errorf("cannot add user supplied docker args: %w", err)
	}
	args = append(args, uargs...)
```

components/docker-up/docker-up/main.go
```diff
@@ -161,10 +175,91 @@ func runWithinNetns() (err error) {
...
+			}
+
+			args = append(args, "--"+mapped, "gitpod")
+		} else {
+			args = append(args, "--"+userArg, userValue)
+		}
+	}
+
+	return args, nil
+}
```
utam0k, line 219: Given the increase in arguments, how about this code?
```suggestion
type ConvertUserArg func(arg string, value string) ([]string, error)

var allowedDockerArgs = map[string]ConvertUserArg{
	"remap-user": RemapUserConverter,
}

func RemapUserConverter(arg string, value string) ([]string, error) {
	args := []string{}
	id, err := strconv.Atoi(value)
	if err != nil {
		return args, err
	}

	for _, f := range []string{"/etc/subuid", "/etc/subgid"} {
		err := adaptSubid(f, id)
		if err != nil {
			return args, xerrors.Errorf("could not adapt subid files: %w", err)
		}
	}

	args = append(args, "--"+"userns-remap", "gitpod")
	return args, nil
}

func userArgs() ([]string, error) {
	args := []string{}
	userArgs, exists := os.LookupEnv(DaemonArgs)
	if !exists {
		return args, nil
	}

	var providedDockerArgs map[string]string
	if err := json.Unmarshal([]byte(userArgs), &providedDockerArgs); err != nil {
		return nil, xerrors.Errorf("unable to deserialize docker args: %w", err)
	}

	for userArg, userValue := range providedDockerArgs {
		mapped, exists := allowedDockerArgs[userArg]
		if !exists {
			continue
		}
		arg, err := mapped(userArg, userValue)
		if err != nil {
			return nil, err
		}
		args = append(args, arg...)
	}

	return args, nil
}
```

components/docker-up/docker-up/main.go
```diff
@@ -161,10 +175,91 @@ func runWithinNetns() (err error) {
...
+
+	if id != 0 {
+		newfile.WriteString(fmt.Sprintf("gitpod:%d:%d\n", 1, id))
+		newfile.WriteString("gitpod:33333:1\n")
+
+	} else {
+		newfile.WriteString("gitpod:33333:1\n")
+		newfile.WriteString(fmt.Sprintf("gitpod:%d:%d\n", 1, 33332))
+		newfile.WriteString(fmt.Sprintf("gitpod:%d:%d\n", 33334, 32200))
+	}
```
utam0k, line 240: How about this? I couldn't get the point what `32200` mean. What does it mean?
```suggestion
	gitpodUserId := 33333
	mappingFmt := func(username string, id int, size int) string { return fmt.Sprintf("%s:%d:%d\n", username, id, size) }

	if id != 0 {
		newfile.WriteString(mappingFmt("gitpod", 1, id))
		newfile.WriteString(mappingFmt("gitpod", gitpodUserId, 1))
	} else {
		newfile.WriteString(mappingFmt("gitpod", gitpodUserId, 1))
		newfile.WriteString(mappingFmt("gitpod", 1, gitpodUserId-1))
		newfile.WriteString(mappingFmt("gitpod", gitpodUserId+1, 33200))
	}
```

utam0k:
I agree with you, we have to add the documentation. But I didn't know 😭 
> Have not added documentation before, so any advice is appreciated.

Furisto:
@utam0k PTAL

kylos101:
Hey @mikenikles , if we wanted to document a new capability to our `docker in workspaces` feature - can you help guide on what is recommended to update related documentation?

mikenikles:
We don't have a page specifically for docker in workspaces, but could add one in the Configure section.

Alternatively, I'd add a note to https://www.gitpod.io/docs/environment-variables#user-specific-environment-variables given this PR introduces a new user-level environment variable.

The corresponding source file is at https://github.com/gitpod-io/website/blob/main/gitpod/docs/environment-variables.md.

Furisto:
> First of all, thanks for the great PR. Do you know why? I was a little curious.
> 
> > The most recent version of runc needs to be installed as otherwise mounting cgroupfs will fail with when starting the container

I did not try to verify which specific change in runc made the difference, although I have some suspicions.

components/docker-up/docker-up/main.go
```diff
@@ -191,25 +196,18 @@ func setUserArgs(args []string) ([]string, error) {
 	}
 
 	for userArg, userValue := range providedDockerArgs {
-		mapped, exists := allowedDockerArgs[userArg]
+		converter, exists := allowedDockerArgs[userArg]
```
utam0k, line 199: I love this name ❤️

components/docker-up/docker-up/main.go
```diff
@@ -229,14 +243,15 @@ func adaptSubid(oldfile string, id int) error {
...
+	if id != 0 {
+		newfile.WriteString(mappingFmt("gitpod", 1, id))
+		newfile.WriteString(mappingFmt("gitpod", gitpodUserId, 1))
 	} else {
-		newfile.WriteString("gitpod:33333:1\n")
-		newfile.WriteString(fmt.Sprintf("gitpod:%d:%d\n", 1, 33332))
-		newfile.WriteString(fmt.Sprintf("gitpod:%d:%d\n", 33334, 32200))
+		newfile.WriteString(mappingFmt("gitpod", gitpodUserId, 1))
+		newfile.WriteString(mappingFmt("gitpod", 1, gitpodUserId-1))
+		newfile.WriteString(mappingFmt("gitpod", gitpodUserId+1, 32200)) // map rest of user ids in the user namespace
```
utam0k, line 254: A good comment. Thanks.
> map rest of user ids in the user namespace

utam0k (approved):
LGTM 💯

shaoye:
```
1. Add environment variable DOCKER_DAEMON_ARGS to your user settings. The content should look like this:
{ "remap-user": "1000" } (Means container id 1000 will be mapped to gitpod user).
2. Start workspace
3. mkdir /tmp/me
4. Run docker image which supports running as unprivileged user e.g. sudo docker run -it -u node -v /tmp/me:/tmp node sh
5. Create file within container e.g. touch /tmp/foo
6. Verify that the created file has gitpod:gitpod permissions in the workspace
```

got permission denied at step 5, 
```
touch: cannot touch '/tmp/foo': Permission denied
```
the `node` user can't create any files since `/tmp/me` is owned by `gitpod` user
```
/tmp$ ls -al
total 4
drwxr-xr-x 2 33333 33333   40 Jun 21 21:59 .
drwxr-xr-x 1 root  root  4096 Jun 21 22:00 ..
```

utam0k:
@shaoye Thanks for your report. Can I ask you to make sure how under `/workspace/*` dir is the right behavior?

shaoye:
> @shaoye Thanks for your report. Can I ask you to make sure how under `/workspace/*` dir is the right behavior?

same, still got permission error if I mount /workspace/tmp/me
```
-v /workspace/tmp/me:/tmp
```

utam0k:
@shaoye Thanks! Can I ask you to create the issue? If you create it, we can schedule the issue.

shaoye:
> @shaoye Thanks! Can I ask you to create the issue? If you create it, we can schedule the issue.

np! #10844

utam0k:
@shaoye Thank you!

PR #8438, opened 2022-02-24
[server] Guarantee to run HostContextProvider.ensureInitialized just …

geropl:
…once

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/server/src/auth/host-context-provider-impl.ts
+++ b/components/server/src/auth/host-context-provider-impl.ts
@@ -15,11 +15,13 @@ import { HostContainerMapping } from "./host-container-mapping";
 import { RepositoryService } from "../repohost/repo-service";
 import { TraceContext } from "@gitpod/gitpod-protocol/lib/util/tracing";
 import { repeat } from "@gitpod/gitpod-protocol/lib/util/repeat";
+import { DisposableCollection } from "@gitpod/gitpod-protocol";
 
 @injectable()
 export class HostContextProviderImpl implements HostContextProvider {
     protected fixedHosts = new Map<string, HostContext>();
     protected dynamicHosts = new Map<string, HostContext>();
+    protected disposables = new DisposableCollection();
 
     @inject(Config)
     protected readonly config: Config;
@@ -30,15 +32,8 @@ export class HostContextProviderImpl implements HostContextProvider {
     @inject(HostContextProviderFactory)
     protected factory: HostContextProviderFactory;
 
-    async init() {
-        await this.ensureInitialized();
-    }
-
     protected initialized = false;
-    protected async ensureInitialized() {
-        if (this.initialized) {
-            return;
-        }
+    async init() {
         this.createFixedHosts();
 
         try {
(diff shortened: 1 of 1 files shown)


components/server/src/auth/host-context-provider-impl.ts
```diff
@@ -32,33 +34,41 @@ export class HostContextProviderImpl implements HostContextProvider {
...
+        const updateDynamicHostsJob = repeat(async () => {
+            const span = TraceContext.startSpan("updateDynamicHosts");
+            try {
+                await this.updateDynamicHosts({span});
+            } catch (error) {
+                log.error(`Failed to update dynamic hosts.`, error);
+            } finally {
+                span.finish();
+            }
+        }, 10000);
```
geropl, line 48: @AlexTugarev What's the use-case again where we have to sync this often?

And: can we change the "add host provider" code to:
 - add entry to the DB
 - PLUS execute the "addDynamicHostProvider" from below
so we don't rely on this timeout anymore?

Just saw the calls in honeycomg on staging, and I'd really love to set this to 30s or so, even with the optimization.

components/server/src/auth/host-context-provider-impl.ts
```diff
@@ -32,33 +34,41 @@ export class HostContextProviderImpl implements HostContextProvider {
...
         }
-        this.createFixedHosts();
-
         try {
-            await this.updateDynamicHosts({ });
-        } catch (error) {
-            log.error(`Failed to update dynamic hosts.`, error);
-        }
+            this.initializationRunning = true;
+            this.createFixedHosts();
```
geropl, line 61: @AlexTugarev Note that we do not handle the error case here, but leave it in `this.initialized = false` state, keeping the current logic. Does that really make sense? :thinking:

components/server/src/auth/host-context-provider-impl.ts
```diff
@@ -58,9 +53,17 @@ export class HostContextProviderImpl implements HostContextProvider {
                 span.finish();
             }
         }, 1999);
+        this.disposables.push(updateDynamicHostsJob);
+
         this.initialized = true;
```
AlexTugarev, line 58: let's moved this to first line of `init`. I think it's enough to signal a call to `init` was made. we don't need/want to wait for a first update.

geropl: I thought this should signal a defined state? If we do this we don't need `assertInitialized` at all, no? :thinking:

AlexTugarev: No. 

`assertInitialized` will throw if one would remove the `init` call by accident. Thus it needs to be checked in the public methods.

components/server/src/auth/host-context-provider-impl.ts
```diff
@@ -58,9 +53,17 @@ export class HostContextProviderImpl implements HostContextProvider {
                 span.finish();
             }
         }, 1999);
+        this.disposables.push(updateDynamicHostsJob);
```
AlexTugarev, line 56: @geropl, wouldn't that already do the initial `await this.updateDynamicHosts` as well? 
So, we can remove the L.40 call, no?

PR #8458, opened 2022-02-25
[ws-daemon] Fix CPU limit annotation

csweichel:
## Description
When we updated the CPU limiting mechanism, we broke the `gitpod.io/cpuLimit` annotation. This PR makes that annotation work again. It effectively controls the base and burst limits of a workspace.

## How to test
1. Start a workspace
2. Set the `gitpod.io/cpuLimit` annotation on that workspace
3. Observe how the workspace's cfs quota gets set to that value

## Release Notes

```release-note
Made the `gitpod.io/cpuLimit` annotation work again
```


csweichel:
closed in favour of https://github.com/gitpod-io/gitpod/pull/8459

Issue #7713, opened 2022-01-20
Epic: SSH access copy/paste via dashboard


loujaybee:
Now that we have the new SSH gateway and the simpler authentication flow, it is possible that a user can copy/paste an SSH command onto their local machine to facilitate quicker/easier access to a Gitpod workspace. This UX could come on the dashboard (**Figure 1**). However, whilst we also have the landing page on start of workspace (**figure 2**), we should also consider offering the SSH option here, too. Following the implementation of this feature, we also need to review documentation around local companion and SSH access, as this will provide an alternative/substitute to the local companion. 

| Figure 1      | Figure 2 |
| ----------- | ----------- |
| ![image](https://user-images.githubusercontent.com/5528307/150317817-e01da775-06bc-4ea7-93e9-ffc9bbce9fa1.png) | ![image](https://user-images.githubusercontent.com/5528307/150317955-9da99aca-297e-487f-98f4-f9937d2e512a.png)

loujaybee:
Would be good to get @gtsiolis's and @jldec / Team WebApp's thoughts on the UX of this.

iQQBot:
how is this going

gtsiolis:
Moving this to scheduled in the [product design board](https://github.com/orgs/gitpod-io/projects/29). 🥯 

[ETA](https://en.wikipedia.org/wiki/Estimated_time_of_arrival): End of this week. ⏱️

gtsiolis:
Off by one day[[1](https://github.com/gitpod-io/gitpod/issues/7713#issuecomment-1055577182)] but here're some design specs we could use here. Feedback is welcome!

Some thoughts on this MVC (minimum viable change) iteration:
1. Since we currently only support copying and pasting an SSH command to connect but do not offer an in-browser UX where we allows used to connect to SSH without a local terminal application, re-using the modal here sounds good.
2. Let's use a more specific verb here like _Connect_ instead of _Open_ to provide a better and more specific hint.
3. Hiding the dropdown option when the workspace is not running as we do with the _Stop_ option sounds good.

| Dropdown (Workspaces) | Modal | Dropdown (Start) |
|-|-|-|
[...]

PR #8464, opened 2022-02-25
Add xterm as an IDE

filiptronicek:
## Description
Adds Xterm.js as a web alternative to connecting to your workspace via SSH from your terminal. The code can be found at https://github.com/gitpod-io/xterm-web-ide and the Docker image is at [ide/xterm-web](https://console.cloud.google.com/gcr/images/gitpod-core-dev/eu/build/ide/xterm-web).

## How to test
1. Open the preview environment
2. In the preferences, select notice there is no Xterm IDE
3. In [Configcat](https://app.configcat.com/), include yourself in the enable-rule for `experimentalIdes` (non-prod)
4. Wait a couple of minutes (🫖 time)
5. Reload and see the Terminal IDE magically appear
6. Create a workspace and have a blast

## Useful commands

Re-pull the IDE: 
```sh
kubectl rollout restart deployment/ide-service
```

## Current limitations

- Does not work on mobile (tried Safari on iPadOS)

## Release Notes

```release-note
NONE
```
⬆️ we can mention it after we finish the internal dog food

## Todo

- [x] `gp stop` closes xterm.js editor and shows the "Workspace Stopping" screen 
- [x] `gp open` works
- [x] `gp preview` works
- [x] Ensure heartbreating works
- [x] Move Docker image from DockerHub to our GCP registry

### Nice-to-haves
- [ ] Local echo [[source](https://github.com/gitpod-io/openvscode-server/blob/gp-code/main/src/vs/workbench/contrib/terminal/browser/terminalTypeAheadAddon.ts)]

## Werft options:

- [x] /werft with-preview
- [x] /werft with-large-vm
- [x] /werft with-gce-vm


akosyakov:
What we are going to do about it?

felladrin:
I'd love to see it in production. I vote for continuing it, even if it's just developed sporadically.

felladrin:
Idea for a follow-up PR:

In this IDE selection screen,

image

We could have 2 Terminal options:
- Terminal Browser -> xterm.js on browser
- Terminal Desktop -> show the same dialog of the SSH connection, allowing the user to copy the ssh command line for use with SSH Keys and with Access Token.

filiptronicek:
> Reminder: Running "gp stop" should lead the user to the "Stopping Workspace" page.

Thanks for noting this, @felladrin! I tried this and it worked for the most part, but there were indeed occurrences where I got an error message which blocked the navigation. Now it should not block the main thread (we switched over from `alert()` to `<dialog>`) and just work.

felladrin:
ℹ️  Something interesting I saw yesterday: https://github.com/cloudcmd/gritty
Leaving it here as it matches what's being done in this PR. An alternative, if needed.

filiptronicek:
> ℹ️ Something interesting I saw yesterday: [cloudcmd/gritty](https://github.com/cloudcmd/gritty)

Looks interesting, @felladrin! I think that would be useful for some smaller things, but it seems like implementing stuff like custom link matching wouldn't be possible like this 🤔

filiptronicek:
Ok one issue that I would love input for:

This looks good
image

This looks very bad
image

felladrin:
I've also noticed that, at least on Brave browser on MacOS, when you hit the ESC key it loses focus from the input field of _xterm.js_, instead of registering the keypress event into the terminal app running.
It's important for us to fix it because ESC is used on VIM and other editors to exit the Insert/Visual mode.

jwpjrdev:
> Something to add to the docs later: The browser's accessibility functionality of pressing the ESC key to un-focus input fields won't allow sending the keystroke to the terminal. So, for escaping the input mode in terminal apps like VIM, please use ctrl+c.

I'd also like to note that iPads don't even have an escape key—it would be nice to have that information on the loading screen as well when you open a workspace.

axonasif:
> Something to add to the docs later: The browser's accessibility functionality of pressing the ESC key to un-focus input fields won't allow sending the keystroke to the terminal. So, for escaping the input mode in terminal apps like VIM, please use ctrl+c.

However that works on vscode-web terminal while using VIM and such, I wonder how 🤔

filiptronicek:
/werft run with-clean-slate-deployment=true with-preview=true recreate-vm=true

filiptronicek:
Status update: this PR is actively being worked on :)

filiptronicek:
/werft run with-clean-slate-deployment=true with-preview=true recreate-vm=true

filiptronicek:
/werft run with-clean-slate-deployment=true with-preview=true recreate-vm=true

filiptronicek:
/werft run with-clean-slate-deployment=true with-preview=true recreate-vm=true

filiptronicek:
/werft run with-clean-slate-deployment=true with-preview=true recreate-vm=true

filiptronicek:
/werft run with-clean-slate-deployment=true with-preview=true recreate-vm=true

filiptronicek:
/werft run with-clean-slate-deployment=true with-preview=true recreate-vm=true

filiptronicek:
/werft run with-clean-slate-deployment=true with-preview=true recreate-vm=true

filiptronicek:
/werft run with-clean-slate-deployment=true with-preview=true recreate-vm=true

filiptronicek:
/werft run with-clean-slate-deployment=true with-preview=true recreate-vm=true

filiptronicek:
/werft run with-clean-slate-deployment=true with-preview=true recreate-vm=true

filiptronicek:
/werft run with-clean-slate-deployment=true with-preview=true recreate-vm=true

filiptronicek:
/werft run with-clean-slate-deployment=true with-preview=true recreate-vm=true

gtsiolis (commented):
Tried this in the preview environment a while ago and worked like a charm! 🔮 

Adding a few thoughts below:

**thought**: Overall, UX LGTM and following the flow we're using with other editor options.

**question**: Since this has been open and active PR for a while does this mean it's not trivial to maintain as an option? If this add more friction or workload to properly maintain I'd wait to get some user feedback on this before merging it.

**suggestion**: Also, since this option only supports one terminal, this seems more closely related to the connection with [**Connect Line**](https://www.gitpod.io/docs/references/ides-and-editors/command-line) feature. Cc @loujaybee 
- This means users who would be interested in this will probably be interested in either using :a: their local terminal or :b: the browser xterm.
    - This is also related to what a customer reported once in https://github.com/gitpod-io/gitpod/issues/7713#issuecomment-1140534710.
- Maybe it's worth toning down this a bit and merging with the Command Line feature, strengthening the SSH functionality.
    - This would make SSH a primary option for the editor and when the workspace is loaded the user could have the option to copy the access token or SSH key to connect with their local terminal or fallback to opening a terminal in the browser.
    - We could also make it pop up as a smaller browser window which is the most common approach for such cases to emulate a terminal UX for users using the command line in browser.

filiptronicek:
@gtsiolis many thanks for the review! Here are my comments to the questions / thoughts you raised:

> Since this has been open and active PR for a while does this mean it's not trivial to maintain as an option? 

The PR has been open for a year, but that was mostly because it was more of a hobby project with not much time invested in it. We have put some more focus into it in the last couple of weeks and so that may have affected this. The code of the IDE is just a couple hundred lines of Typescript and therefore hopefully shouldn't be too bad to maintain - although we will mostly see after our internal dogfood.

> I'd wait to get some user feedback

That's a pretty neat idea to maybe add some really dedicated users to test this as well, I'm sure we can find someone in https://github.com/gitpod-io/gitpod/issues/11180.

> Maybe it's worth toning down this a bit and merging with the Command Line feature, strengthening the SSH functionality. 

If we after testing decide this IDE is worth keeping, this sounds really great. Having it as a fallback in environments where a desktop terminal is not available / practical was always the goal here.

akosyakov:
We agreed to use feature flag for internal consumption at Gitpod. What else is required to land it?

filiptronicek:
Those damned miss-clicks!

akosyakov:
@filiptronicek Should not be webapp part to be deployed first? Can this PR be deployed in any order?

filiptronicek:
@akosyakov I think the order should be first IDE service, because otherwise the IDE will be shown to all users (dashboard won't know yet about the feature flag)

components/supervisor/frontend/src/index.ts
```diff
@@ -64,7 +64,7 @@ const ideService = IDEFrontendService.create();
 const loadingIDE = new Promise((resolve) => window.addEventListener("DOMContentLoaded", resolve, { once: true }));
 const toStop = new DisposableCollection();
 
-document.body.style.visibility = "hidden";
```
akosyakov, line 67: @filiptronicek why do we do it?

filiptronicek: @akosyakov I'm trying to get the opening working in Safari. It doesn't respect visibility hidden on the iFrame so trying with opacity.

akosyakov: why does it work for VS Code Web?

mads-hartmann (approved):
The SID-related changes LGTM

components/supervisor/frontend/src/index.ts
filiptronicek: It looks like it's a bug / just different behavior in Safari. It comes down to the order of the elements -> the iframe needs to be first, and that presumably happens because of VS Code loading in the elements asynchronously and xterm having the container element right there from the start.

filiptronicek: Fixed and PR description updated

filiptronicek:
Alright... didn't expect roboquat to merge this automatically... let's make another PR.

vulkoingim:
> Alright... didn't expect roboquat to merge this automatically... let's make another PR. Sorry folks

When you change the base branch for the PR from main the requirements to merge are no longer the same :) If there's an approval (regardless from whom) it gets merged.

PR #8469, opened 2022-02-26
change alert period for ws stuck on starting or stopping to 20m

sagor999:
## Description

Reduce alert period from 1h to 20m for `GitpodWorkspaceStuckOnStarting` and `GitpodWorkspaceStuckOnStopping`.
Reasoning:
Currently it takes one hour of that alert firing before pager duty gets issued. This is not great customer experience.
I checked in grafana for the last 7 days and that change should not produce any false positives.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
none
```

### DIFF
--- a/operations/observability/mixins/workspace/rules/components/workspaces/alerts.libsonnet
+++ b/operations/observability/mixins/workspace/rules/components/workspaces/alerts.libsonnet
@@ -14,11 +14,11 @@
             labels: {
               severity: 'critical',
             },
-            'for': '1h',
+            'for': '20m',
             annotations: {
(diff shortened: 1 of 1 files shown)


operations/observability/mixins/workspace/rules/components/workspaces/alerts.libsonnet
```diff
@@ -31,7 +31,7 @@
             labels: {
               severity: 'critical',
             },
-            'for': '1h',
+            'for': '20m',
```
princerachit, line 34: Can you update the description of the alert. Applies elsewhere

utam0k (approved):

PR #8471, opened 2022-02-27
Enable cpu limiting using cgroup v2

Furisto:
## Description
Enable cpu limiting using cgroup v2.

## Related Issue(s)
https://github.com/gitpod-io/gitpod/issues/8498

## How to test
- Open workspace in [preview](https://ga1d12b522fcffa694f4478.workspace-preview.gitpod-io-dev.com/workspaces)
- Observe cgroup values being updated

## Notes for reviewers
Considered using containerd cgroups library but
- The api for v1 and v2 is different so we still would have duplication.
- Updating a cgroup is currently not supported. 

## Release Notes
```release-note
Support cpu limiting using cgroup v2
```

### DIFF
--- a/components/common-go/cgroups/cgroup.go
+++ b/components/common-go/cgroups/cgroup.go
@@ -1,0 +1,115 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package cgroups
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+const DefaultCGroupMount = "/sys/fs/cgroup"
+
+type CgroupSetup int
+
+const (
+	Unknown CgroupSetup = iota
+	Legacy
+	Unified
+)
+
+func (s CgroupSetup) String() string {
+	return [...]string{"Legacy", "Unified"}[s]
+}
+
+func GetCgroupSetup() (CgroupSetup, error) {
+	controllers := filepath.Join(DefaultCGroupMount, "cgroup.controllers")
+	_, err := os.Stat(controllers)
+
+	if os.IsNotExist(err) {
+		return Legacy, nil
+	}
+
+	if err == nil {
+		return Unified, nil
+	}
+
+	return Unknown, err
+}
+
+func IsUnifiedCgroupSetup() (bool, error) {
+	setup, err := GetCgroupSetup()
+	if err != nil {
+		return false, err
+	}
+
+	return setup == Unified, nil
+}
+
+func IsLegacyCgroupSetup() (bool, error) {
+	setup, err := GetCgroupSetup()
+	if err != nil {
+		return false, err
+	}
+
+	return setup == Legacy, nil
+}
+
+func EnsureCpuControllerEnabled(basePath, cgroupPath string) error {
+	targetPath := filepath.Join(basePath, cgroupPath)
+	if enabled, err := isCpuControllerEnabled(targetPath); err != nil || enabled {
+		return err
+	}
+
+	err := writeCpuController(basePath)
+	if err != nil {
+		return err
+	}
+
+	levelPath := basePath
+	cgroupPath = strings.TrimPrefix(cgroupPath, "/")
+	levels := strings.Split(cgroupPath, string(os.PathSeparator))
+	for _, l := range levels[:len(levels)-1] {
+		levelPath = filepath.Join(levelPath, l)
+		err = writeCpuController(levelPath)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func isCpuControllerEnabled(path string) (bool, error) {
+	controllerFile := filepath.Join(path, "cgroup.controllers")
+	controllers, err := os.ReadFile(controllerFile)
+	if err != nil {
+		return false, err
+	}
+
+	for _, ctrl := range strings.Fields(string(controllers)) {
+		if ctrl == "cpu" {
+			// controller is already activated
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func writeCpuController(path string) error {
+	f, err := os.OpenFile(filepath.Join(path, "cgroup.subtree_control"), os.O_WRONLY, 0)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	_, err = f.Write([]byte("+cpu"))
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/components/ws-daemon/pkg/cpulimit/cfs_v2.go
+++ b/components/ws-daemon/pkg/cpulimit/cfs_v2.go
@@ -1,0 +1,125 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package cpulimit
+
+import (
+	"bufio"
+	"math"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"time"
+
+	"golang.org/x/xerrors"
+)
+
+type CgroupV2CFSController string
+
+func (basePath CgroupV2CFSController) Usage() (CPUTime, error) {
+	usage, err := basePath.getFlatKeyedValue("usage_usec")
+	if err != nil {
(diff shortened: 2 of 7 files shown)


utam0k:
@Furisto Nice work! I created the issue and update the description.
https://github.com/gitpod-io/gitpod/issues/8497

components/ws-daemon/pkg/cpulimit/cfs_v2.go
```diff
@@ -0,0 +1,125 @@
...
+
+func (basePath CgroupV2CFSController) readCpuMax() (time.Duration, time.Duration, error) {
+	cpuMaxPath := filepath.Join(string(basePath), "cpu.max")
+	cpuMax, err := os.ReadFile(cpuMaxPath)
+	if err != nil {
+		return 0, 0, xerrors.Errorf("unable to read cpu.max: %w", err)
+	}
+
+	parts := strings.Fields(string(cpuMax))
+	if len(parts) < 2 {
```
utam0k, line 67: `parts` should always be two, is that right?
```suggestion
	if len(parts) != 2 {
```

components/ws-daemon/pkg/cpulimit/cfs_v2.go
```diff
@@ -0,0 +1,125 @@
...
+	parent := CgroupV2CFSController(filepath.Dir(string(basePath)))
+	quota, _, err := parent.readCpuMax()
+	if err != nil {
+		return time.Duration(0)
+	}
+
+	return time.Duration(quota)
+}
+
+func (basePath CgroupV2CFSController) getFlatKeyedValue(key string) (int64, error) {
```
utam0k, line 104: I was a little confused by the name of function. Why not make the function role read cpustat or receive the path to the file?
```suggestion
func (basePath CgroupV2CFSController) readCpuStat(key string) (int64, error) {
```

components/ws-daemon/pkg/cpulimit/cfs_v2.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		return 0, 0, xerrors.Errorf("unable to read cpu.max: %w", err)
+	}
+
+	parts := strings.Fields(string(cpuMax))
+	if len(parts) < 2 {
+		return 0, 0, xerrors.Errorf("cpu.max did not have expected number of fields: %s", parts)
+	}
+
+	var quota int64
```
utam0k, line 71: I think `quota` is always a positive value, right?
```suggestion
	var quota uint64
```

components/ws-daemon/pkg/cpulimit/dispatch.go
```diff
@@ -214,3 +220,18 @@ func (d *DispatchListener) WorkspaceUpdated(ctx context.Context, ws *dispatch.Wo
...
+	if os.IsNotExist(err) {
+		return CgroupV1CFSController(filepath.Join(basePath, "cpu", cgroupPath)), nil
+	}
+
+	if err == nil {
+		return CgroupV2CFSController(filepath.Join(basePath, cgroupPath)), nil
+	}
+
+	return nil, err
+}
```
utam0k, line 237: @csweichel @Furisto 
It is not very good to have scattered cgroup v2 decisions, so why not unify them since we will be using the runc library? k8s seems to use the same determination.
https://github.com/kubernetes/kubernetes/blob/6eca8cc472ee140de8fc7181a4349df643a86e04/pkg/kubelet/cm/cgroup_manager_linux.go#L152
```suggestion
func newCFSController(basePath, cgroupPath string) (CFSController, error) {
	if cgroups.IsCgroup2UnifiedMode() {
		return CgroupV2CFSController(filepath.Join(basePath, cgroupPath)), nil
	} else {
		return CgroupV1CFSController(filepath.Join(basePath, "cpu", cgroupPath)), nil
	}
}
```

utam0k: Also, since v1 and v2 can coexist, you need to make sure that cpu is enabled in `cgroup.controllers`.

utam0k (commented):
I left some comments, but I like these codes!

utam0k:
This PR is quite a great feature for users. I would love to see you write a release note.

components/ws-daemon/pkg/cpulimit/cfs_v2.go
Furisto: It can also be 3 if the burst feature of the scheduler is activated. Even if it were only 2 at the moment, it is always possible that later kernel versions add additional fields and I do not want the code to break because of that .

components/ws-daemon/pkg/cpulimit/cfs_v2.go
Furisto: It is also used for reading the number of throttled periods. Cgroup v2 has [four file formats](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#format) so I named the function after the format that it can interpret.

Furisto:
> This PR is quite a great feature for users. I would love to see you write a release note.

Wasn't sure if we already want to announce it to users :wink: . Added release note.

components/ws-daemon/pkg/cpulimit/dispatch.go
csweichel: Personally I found the runc/containerd cgroup library very hard to use. It's just so much easier for many operations to implement them on a filesystem level. rootlesskit came to the same conclusion.

Furisto: > Also, since v1 and v2 can coexist, you need to make sure that cpu is enabled in `cgroup.controllers`.

They can coexist, but if v1 has the cpu controller attached, then we would need to unmount the v1 cgroup mount in order to make the cpu controller available for v2 which would only work if we somehow managed to remove all processes from the cgroup v1 hierarchy. It would also be a pretty unusual custom setup. The usual options are either pure v1, pure v2 or hybrid mode which is basically v1 (all controllers are attached to v1) with some features available for v2 (this would be detected as v1 in this implementation which is correct because that is where the controllers are).

I agree though that it could theoretically be possible that we have cgroup v2 setup where the cpu controller has not been activated. I will handle that case.

csweichel: I'd argue we settle for pure v1 or pure v2. All other modes result in undefined behaviour.

Furisto: We are currently running hybrid mode on our own nodes. This is the default for Ubuntu. It is not a big deal though. As I said, for all intents and purposes hybrid mode is v1.

aledbf: > We are currently running hybrid mode on our own nodes.

We ran V1 in the nodes (`systemd.unified_cgroup_hierarchy=0`). At the time, v2 broke the CPU limiting

utam0k: I think this is the best.
> I'd argue we settle for pure v1 or pure v2. All other modes result in undefined behaviour.

utam0k: @Furisto @csweichel 
So that is the situation with rootless. I did not know that. I would like to take advantage of that lesson. My root concern is not the using the runc and contained libraries, but the implementation of the determination of whether or not it is cgroup v2 at every point in the future. How about we take a lesson from rootless and have a cgroup v2 check to implement original our codes somewhere(e.g. utils.go or something)?

> Personally I found the runc/containerd cgroup library very hard to use. It's just so much easier for many operations to implement them on a filesystem level. rootlesskit came to the same conclusion.

csweichel: Indeed it makes sense to group the cgroupv2 decision in one place. We could introduce a `cgroups` package and put the check in there. Alternatively, we could move it to the `cpulimit` package, as that one's a leaf in the dependency tree.

Furisto:
@utam0k @csweichel PTAL

components/common-go/cgroups/cgroup.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	if err == nil {
+		return Unified, nil
+	}
+
+	return Unknown, err
+}
+
+func IsUnifiedCgroupSetup() (bool, error) {
```
utam0k, line 42: Thanks  for reflecting my opinion❤️

utam0k (approved):
I like these codes! LGTM

PR #8336, opened 2022-02-21
Fix PREVENT_METADATA_ACCESS is not effact

iQQBot:
## Description

rename THEIA_PREVENT_METADATA_ACCESS to GITPOD_PREVENT_METADATA_ACCESS
and fix is not effect in coredev
also fix workspace `nodeAffinity` in core-dev maybe it can improve stable for core-dev

## Related Issue(s)

Relate https://github.com/gitpod-io/ops/pull/1246/files

## How to test

1. start workspace
2. you will get shutdown immediately
3. and you can find `metadata access is possible` log in GCP

image

## Release Notes

```release-note
NONE
```

PR #8476, opened 2022-02-28
[preview env] extend deny egress block

iQQBot:
## Description

From https://github.com/gitpod-io/gitpod/pull/8336 We added GCP metadata security detection to the Preview environment, so that if GCP metadata is accessible in the workspace, the supervisor will immediately end the process to prevent sensitive information from being leaked.
 
But if we only block 169.254.169.254, it still can access by HTTP request, I don't know why

Perhaps in GKE, this IP is special and may not be controlled by the network policy

After testing, I found that changing the blocked IP to 169.254.169.254/30 solves the problem, I don't know the exact reason, but it does work

I am not sure this will affect any GKE cluster or just our preview environment, if all GKE clusters affect, I think changing installer network policy is a better way

We have two ways to temporarily alleviate this problem
1. disable metadata detection in the Preview environment
2. use 169.254.169.254/30 as a mitigation measure

This PR use option 2

## Related Issue(s)

Fixes #

## How to test

Start a workspace, workspace can run

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.werft/jobs/build/installer/post-process.sh
+++ b/.werft/jobs/build/installer/post-process.sh
@@ -262,6 +262,13 @@ while [ "$i" -le "$DOCS" ]; do
       yq w -i k8s.yaml -d "$i" spec.ingress[0].ports[0].port "$WS_DAEMON_PORT"
    fi
 
+   # NetworkPolicy for workspace-default
+   if [[ "workspace-default" == "$NAME" ]] && [[ "$KIND" == "NetworkPolicy" ]]; then
+      WORK="overrides for $NAME $KIND"
+      echo "$WORK"
+      yq w -i k8s.yaml -d "$i" spec.egress[0].to[0].ipBlock.except[0] 169.254.169.254/30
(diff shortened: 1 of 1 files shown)


MrSimonEmms:
@iQQBot I'm not sure of the underlying cause, but this looks to be ok.

Looking through the code, we have this IP in twice
 - [workspace](https://github.com/gitpod-io/gitpod/blob/4820e804bb86ecee9548ce9f1f9aca761d8436b9/install/installer/pkg/components/workspace/networkpolicy.go#L92) (you've got this one in the change)
 - [image builder mk3](https://github.com/gitpod-io/gitpod/blob/4820e804bb86ecee9548ce9f1f9aca761d8436b9/install/installer/pkg/components/image-builder-mk3/networkpolicy.go#L42) (this isn't changed by your post-processing - should it be?)

mustard-mh (approved):
Test ok, approve to make preview env works

Issue #8107, opened 2022-02-08
[installer] Rename and print OR remove `--danger-use-unsupported-config` flag


sagor999:
### Bug description

when using experimental settings in installer, error message is not very helpful. 
would be good to tell user how to enable experimental options in that message.
Also, after digging through source code I did find the switch for it: 
`-danger-use-unsupported-config   enable use of unsupported config`
which doesn't mention anything that it enables experimental settings.

### Steps to reproduce

```
./gitpod-installer render --config overrride-values.yaml > manifests.yaml
config contains experimental options - ignoring them
```

metcalfc:
This is a rough UX. Can we change to flag to something that connects it to `experimental` ? 

```--enable-experimental-config```

The user of the installer is considered someone who is ok with low-level access and is often other Gitpod folks. I think the safety is good but we should still make it easy. To go with this, when we warn folks lets tell them how to fix it, ie add the flag vs just ignore it.

If folks have experimental config sections in their YAML, chances are they want to use them. Should we just bail at this point and emit the warning that they should re-run with the flag or remove the experimental config?

corneliusludmann:
> If folks have experimental config sections in their YAML, chances are they want to use them. Should we just bail at this point and emit the warning that they should re-run with the flag or remove the experimental config?

Given that the main config surface for most people out there should be the Replicated UI and that the term “experimental” in the config file should be warning enough, I would also vote for removing the flag entirely.

MrSimonEmms:
> I would also vote for removing the flag entirely.

I think I would respectfully disagree. The experimental config is not for most users (truthfully, it's mostly for SaaS) and there's a danger that people read "experimental" as "beta" or "this is some brilliant stuff that's coming in the future - use it now to be bleeding-edge" which it absolutely isn't.

It's unsupported configuration which allows users to go outside the tracks of the Installer's framework

corneliusludmann:
OK, convinced me. In particular, I think @csweichel's [argument](https://github.com/gitpod-io/gitpod/pull/8477#issuecomment-1053985130) is key:
> Say you get config sent to you, don't really understand it in detail, then this flag will make (hopefully) make you stop and think for a second.

PR #8477, opened 2022-02-28
installer: clarify experimental config flag 

Pothulapati:
## Description

This PR updates the flag descriptions, and the `render` display
notes to be more explicit on what experimental config is.

This flag is required, to make sure users are opting in explicitely/
knowlingly to use the experimental config.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/8107

## How to test

`render` cmd should have the renamed naming, and messages

## Release Notes

```release-note
Renamed `danger-use-unsupported-config` flag in the installer to `use-experimental-config`
```

## Documentation

/werft no-preview

### DIFF
--- a/install/installer/cmd/render.go
+++ b/install/installer/cmd/render.go
@@ -46,9 +46,9 @@ A config file is required which can be generated with the init command.`,
 
 		if cfg.Experimental != nil {
 			if renderOpts.UseExperimentalConfig {
-				fmt.Fprintf(os.Stderr, "rendering using experimental config - here be dragons\n")
+				fmt.Fprintf(os.Stderr, "rendering using experimental config\n")
 			} else {
-				fmt.Fprintf(os.Stderr, "config contains experimental options - ignoring them\n")
+				fmt.Fprintf(os.Stderr, "ignoring experimental config. Use `--use-experimental-config` to include the experimental section in config\n")
 				cfg.Experimental = nil
 			}
 		}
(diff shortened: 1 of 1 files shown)


csweichel:
I'm a bit late to the party - the motivation for this switch is to have a clear (and deliberately inconvenient) hurdle to use experimental config. If we don't have a cost associated with this config, what distinguishes it from regular config other than some random bit "experimental" in the config?

@metcalfc's proposal was 💯 on point imho: just make the warning/message clearer and tell users what they have to do to use the experimental bit config.

Also, the assumption that "users have experimental config in there so they must know what they're doing" is flawed. Say you get config sent to you, don't really understand it in detail, then this flag will make (hopefully) make you stop and think for a second. Granted, with the Replicated layer on top that's less of a concern - but still valid. Imagine a new hire at Gitpod needing to use the installer and stumbling over this switch, you've just taught them there's "experimental config". Add some more detail to the flag description and error message, and they understand why it's there.

Pothulapati:
> Say you get config sent to you, don't really understand it in detail, then this flag will make (hopefully) make you stop and think for a second. Granted, with the Replicated layer on top that's less of a concern - but still valid. Imagine a new hire at Gitpod needing to use the installer and stumbling over this switch, you've just taught them there's "experimental config". Add some more detail to the flag description and error message, and they understand why it's there.

Agree on the points specified here! Updated the PR to rename the flag, and have the display messages be clearer.

install/installer/cmd/render.go
```diff
@@ -46,9 +46,9 @@ A config file is required which can be generated with the init command.`,
 
 		if cfg.Experimental != nil {
 			if renderOpts.UseExperimentalConfig {
-				fmt.Fprintf(os.Stderr, "rendering using experimental config - here be dragons\n")
+				fmt.Fprintf(os.Stderr, "rendering using experimental config\n")
 			} else {
-				fmt.Fprintf(os.Stderr, "config contains experimental options - ignoring them\n")
+				fmt.Fprintf(os.Stderr, "ignoring experimental config. Use `--use-experimental-config` to also consider the experimental section in config\n")
```
MrSimonEmms, line 51: Suggest changing "to also consider" to "to include"

corneliusludmann (approved):

corneliusludmann:
build is green now, next try

/unhold

PR #8376, opened 2022-02-22
[installer]: remove jaeger operator

MrSimonEmms:
## Description

The Jaeger operator is no longer in use, so removed

## How to test

Install to a cluster

## Release Notes

```release-note
Remove Jaeger operator
```

PR #8491, opened 2022-02-28
[kots]: use Helm for the Installer job

MrSimonEmms:
## Description

This changes the KOTS Installer job to use Helm for the installation of Gitpod rather than `kubectl apply`. **IMPORTANT** this still uses the Gitpod Installer for generating the Kubernetes YAML and is not in any way us going back to using Helm for installing Gitpod.

One thing we have to be aware of with KOTS is that we control what resources third-parties/customers have installed on their clusters under the umbrella of "Gitpod". It is key that we should remove resources when they've been finished with.

An example of this:
- day 1: a user installs Gitpod with in-cluster database/registry/object storage as a way of getting up and running quickly. They use this to get familiar with Gitpod and the installation process
- day 2: once they've decided they're happy, they change their KOTS configuration to migrate to external (and more performant) database/registry/object storage.

If we use `kubectl apply`, there is no history and any resources that are no longer required (eg, the `mysql-0` statefulset) remain in the cluster. As they are no longer being used, they become orphaned from Gitpod installation YAML. More importantly, any persistent volume claims remain and the costs therein.

By switching to `helm upgrade`, we hook into Helm's history. Any resources that are removed between the previous and current installations are deleted.

Helm can be thought of as fulfilling two jobs:
1. generating the YAML from templates
2. deploying and managing resources

As we are still using the Installer to generate the YAML (ie, part 1), we're only using Helm for part 2.

The alternative way of doing this is to implement our own version of this. In reality, we'd likely end up building something very similar to how Helm works. By doing it this way, we reduce the time involved and benefit from all their testing.

## How to test

Deploy via KOTS:
 - install with in-cluster dependencies
 - upgrade to external dependencies - watch the old resources disappear

## Release Notes

```release-note
[kots]: use Helm for the Installer job
```

### DIFF
--- a/install/kots/manifests/gitpod-installer-job.yaml
+++ b/install/kots/manifests/gitpod-installer-job.yaml
@@ -43,7 +43,7 @@ spec:
       containers:
         - name: installer
           # This will normally be the release tag - using this tag as need the license evaluator
-          image: 'eu.gcr.io/gitpod-core-dev/build/installer:sje-licensing.24'
+          image: 'eu.gcr.io/gitpod-core-dev/build/installer:sje-kots-helm.7'
           volumeMounts:
             - mountPath: /config-patch
               name: config-patch
@@ -182,16 +182,39 @@ spec:
               config=$(cat "${CONFIG_FILE}")
               echo "Gitpod: ${CONFIG_FILE}=${config}"
 
-              /app/installer render -c "${CONFIG_FILE}" --namespace {{repl Namespace }} > /tmp/gitpod.yaml
+              echo "Gitpod: Create a Helm template directory"
+              rm -Rf "${GITPOD_OBJECTS}"
+              mkdir -p "${GITPOD_OBJECTS}/templates"
+              cat <<EOF >> "${GITPOD_OBJECTS}/Chart.yaml"
+              apiVersion: v2
+              name: gitpod-kots
+              description: Always ready-to-code
+              Version: "1.0.0"
+              appVersion: "$(/app/installer version | yq e '.version' -)"
+              EOF
+
+              /app/installer render -c "${CONFIG_FILE}" --namespace {{repl Namespace }} > "${GITPOD_OBJECTS}/templates/gitpod.yaml"
 
               # Workaround for #8532 and #8529
               echo "Gitpod: Remove the StatefulSet status object for OpenVSX Proxy"
               yq eval-all --inplace \
                 'del(select(.kind == "StatefulSet" and .metadata.name == "openvsx-proxy").status)' \
-                /tmp/gitpod.yaml
+                "${GITPOD_OBJECTS}/templates/gitpod.yaml"
 
+              # The long timeout is to ensure the TLS cert is created (if required)
               echo "Gitpod: Apply the Kubernetes objects"
-              kubectl apply -f /tmp/gitpod.yaml
+              helm upgrade \
+                --atomic \
+                --cleanup-on-fail \
+                --create-namespace \
+                --install \
+                --namespace="{{repl Namespace }}" \
+                --reset-values \
+                --timeout 1h \
+                --wait \
+                --wait-for-jobs \
+                gitpod \
+                "${GITPOD_OBJECTS}"
 
               echo "Gitpod: Installer job finished - goodbye"
       volumes:
--- a/install/installer/leeway.Dockerfile
+++ b/install/installer/leeway.Dockerfile
@@ -3,6 +3,7 @@
 # See License-AGPL.txt in the project root for license information.
 
 FROM alpine:3.15
+COPY --from=alpine/helm:3.8.0 /usr/bin/helm /usr/bin/helm
 COPY install-installer--app/installer install-installer--app/provenance-bundle.jsonl /app/
 RUN apk add --no-cache curl yq  \
     && curl -L "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl" -o /usr/local/bin/kubectl \


corneliusludmann:
Thanks for creating this pull request as a home for the discussion on this topic. :pray: 

In general, I like the approach of using helm to deploy and manage the resources so that we don't need to implement stuff where a solution is already available.

I would like to hear the opinions of @csweichel and @geropl about this in particular whether they are seeing downsides of this approach that we would remedy with implementing our own solution.

Open Questions:
- We want to use this approach for the Replicated installation path only, right? Do we still need the functionally for the Installer anyway (for our internal usage)?
- If yes, would be the `helm` approach just a temporary solution until we have it in the Installer? Would there be a migration path?

Link to internal discussion for more context: https://gitpod.slack.com/archives/C01KLC56NP7/p1645713248561109

csweichel:
I understand the motivation, but wonder what's the story for the installer outside of replicated? How would we solve this issue for e.g. SaaS oder future BYOC installations? Solving this issue with Helm is one way - but it feels like a step backwards almost.

How would resource deletion/upgrade be implement for a Kubernetes operator?

MrSimonEmms:
No idea how we would do it on an operator, but that would certainly require more effort and testing than it would be using Helm. Not saying that's a deal-breaker, but this is a practical way of getting this working in fairly short order.

My assumption with SaaS is that it's currently handled via ArgoCD - we've certainly had no requirement from @gitpod-io/platform to provide a native way of removing now-orphaned resources (that may be also due to the fact that we're not really removing resources).

csweichel:
> No idea how we would do it on an operator, 

We should find out

> but that would certainly require more effort and testing than it would be using Helm. 

That's not an unreasonable assumption. But it's that: an assumption.

We can use helm for this. But let's use Helm because it's the right thing to do, we're aware of the limits this imposes (e.g. limits to custom resource removal handling), and we're aware of what the alternatives would have been.

MrSimonEmms:
> We can use helm for this. But let's use Helm because it's the right thing to do, we're aware of the limits this imposes (e.g. limits to custom resource removal handling), and we're aware of what the alternatives would have been.

Since I did #8376, I don't believe we have any custom operators in the Installer anymore. The cert-manager CRDs are installed outside of the Installer and will remain that was as far as I'm concerned.

Fair point on the other team implications. I would recommend that we don't do a stable release to Replicated until we have a way of automatically removing unused resources - I do fear lots of support calls into @gitpod-io/engineering-self-hosted on "I'm not using in-cluster DB but there's a `mysql-0` pod" and the associated PVC costs

csweichel:
I was wondering, doesn't replicated help us OOTB here? I.e. they consume the manifests we produce and apply that against the cluster, no? If so, for replicated this is not relevant - and for the other users of the installer (webapp, IDE, workspace) it would be interesting to know how they handle this problem today.

MrSimonEmms:
> I was wondering, doesn't replicated help us OOTB here?

Yes and no. It does for the manifests that we create in full, but that's only things like certs, secrets and others bits of config.

For the actual Gitpod application, the [gitpod-installer-job.yaml](https://github.com/gitpod-io/gitpod/blob/406678c48e14585d4d6dd6727fa5d7feda64e430/install/kots/manifests/gitpod-installer-job.yaml#L68) runs a Bash script which creates the config file, edits it and then applies it - Replicated/KOTS doesn't know anything about those resources and so can't help us with removal. The only way we'd be able to achieve that would be to completely abandon the Installer and to write all the logic in KOTS (which we don't want to do)

csweichel:
> Replicated/KOTS doesn't know anything about those resources and so can't help us with removal

That's disappointing! It was my understanding hitherto that we render the Kubernetes manifests, and KOTS applies them. What's the point of Replicated if we end up having to implement the hard parts?

It looks like in KOTS we are already using the Helm hack. Why not keep this contained to KOTS and think of a proper solution for ourselves, i.e. the other consumer of the installer? Sounds to me like an RFC which discusses those ideas would be in order.

MrSimonEmms:
It is disappointing, but it's also a consequence of us having the Installer to generate the manifests. If you recall the initial meetings with Replicated, their initial thought was us generating the manifests from the Installer and then using KOTS to decide which is implemented - we (rightly) decided against that as something that would be unmaintainable.

The link I sent in my previous comment was from this PR - what's currently in `main` is `kubectl apply -f -`.

The proposed change is for installations via Replicated only. If we were to incorporate this into the Installer, that would be down the line (and would certainly require an RFC)

MrSimonEmms:
@gitpod-io/engineering-self-hosted are we in a position to get this merged?

install/kots/manifests/gitpod-installer-job.yaml
```diff
@@ -43,7 +43,7 @@ spec:
       containers:
         - name: installer
           # This will normally be the release tag - using this tag as need the license evaluator
-          image: 'eu.gcr.io/gitpod-core-dev/build/installer:main.2569'
+          image: 'eu.gcr.io/gitpod-core-dev/build/installer:sje-kots-helm.4'
```
Pothulapati, line 46: So, There will be a separate PR to update this back to main? :thinking:

install/kots/manifests/gitpod-installer-job.yaml
```diff
@@ -182,16 +182,39 @@ spec:
...
 
-              /app/installer render -c "${CONFIG_FILE}" --namespace {{repl Namespace }} > /tmp/gitpod.yaml
+              echo "Gitpod: Create a Helm template directory"
+              rm -Rf "${GITPOD_OBJECTS}"
+              mkdir -p "${GITPOD_OBJECTS}/templates"
+              cat <<EOF >> "${GITPOD_OBJECTS}/Chart.yaml"
+              apiVersion: v2
+              name: gitpod-kots
+              description: Always ready-to-code
+              Version: "1.0.0"
```
Pothulapati, line 192: question: The version here does not matter because this is not user facing from Replicated? and they never directly deal with the chart? Makes sense that way.

MrSimonEmms: Correct, the `appVersion` is the important bit which, crucially, doesn't have to be in semver. This is the chart version which does have to be semver. If we ever significantly change the chart (I doubt it, as it's just the `Chart.yaml`) then this number may be incremented

install/kots/manifests/gitpod-installer-job.yaml
MrSimonEmms: As part of the next release process, this will get updated to whatever the Docker image is - see [2022.02.0](https://github.com/gitpod-io/gitpod/releases/tag/2022.02.0) for example

Pothulapati (approved):
Like other people mentioned here, I'm not a big fan of using Helm internally again but not cleaning up unnecessary resources when users update config feels like a much bigger issue, and we should definitely fix that.

For now, This feels like a good fix, as users never need to interact with any of the Helm stuff here (or even notice this), and it seems to work as expected.

/hold for @corneliusludmann to approve this before we merge.

geropl:
Writing here because @corneliusludmann [pinged initially](https://github.com/gitpod-io/gitpod/pull/8491#issuecomment-1057904449), and @MrSimonEmms and I had Slack disussions about this:
 - it feels like a step backwards to me to use helm for applying the changes
 - but I see the benefit of it being easy to use right now
 - I (and we as WebApp) are also looking for ways to apply/rollout our changes to app clusters (and having SH in mind): we should definitely start tasks about this and share ideas
 - the only (minor?) concern I have is about migration: Given we go with helm now, and implement a different solution for all deployment modes later we need to take care about migrating from helm to "new method" in the replicated installer

corneliusludmann:
Thanks all for the input.

My summary of the discussion:
- It feels quite odd to use Helm for installing the Gitpod resources since we originally created the Installer to get rid of Helm. However, we dropped Helm due to the difficulties of maintaining the Helm templates with the more and more messy `values.yaml` file. Now, we just want to use Helm for managing the resources and let the Installer do the templating job.
- There are no big concerns besides that it feels odd and the migration path when we have another solution in the future. The latter can be handled in the Replicated installer job, though.
- We have a general lack of information around alternatives, requirements across teams, vision for a general solution. We will start a process that incorporates all teams on this topic next quarter. However, that needs some time and we have the need for a solution for the Replicated installer now. That shouldn't hold us back to use this approach for the Replicated installer, especially when we have a path to migrate this in the future to another approach.


It would be very interesting if Helm would handle issues [like these](https://github.com/gitpod-io/gitpod/issues/8653) as well since this would be much harder to solve on ower own than just removing unnecessary resources. Could you check this, @MrSimonEmms?

MrSimonEmms:
@corneliusludmann upsettingly, Helm does **NOT** solve the label issue you raised and for the same reasons - the field is immutable.

Personally, I don't think that makes much difference as to this PR as it's a change that's happened elsewhere that's broken the update from February to March releases. It would have been nice had Helm solved this, but it doesn't.

What it does mean is that we must all be careful when introducing potentially breaking changes against long-running Gitpod clusters. In SaaS, we tend to use a blue/green deployment pattern but many self-hosted users will want to simply upgrade their existing clusters.

corneliusludmann (approved):

PR #3723, opened 2021-04-02
[dashboard] Admin dashboard

svenefftinge:
Adds the admin dashboard to the new dashboard.

How to test:
 - create a user with 'admin' flag (default in dev staging)
 - you should see the admin menu on top.
 - try search for users or workspaces.

PR #8503, opened 2022-03-01
Remove free text search on contextURL

### LINKED ISSUE #8453: Change "Search workspaces" to "Enter workspace ID" and disable contextURL search.
Workspace ID search returns quite fast, and searches usually do use the workspace IDs. 
Therefore we will only allow workspace ID searches and disable the free text search.

### LINKED ISSUE #8454: Fix search input to trim whitespace in all searches
Searches with whitespaces (can happen when copy-pasting) would fail.

laushinka:
## Description
This change disables the free text search, and encourages searching on workspace IDs.  
It's a minimum change for now - there will be follow-up tasks to improve the search[[1](https://github.com/gitpod-io/gitpod/issues/8453#issuecomment-1050875763)][[2](https://github.com/gitpod-io/gitpod/issues/8453#issuecomment-1054716890)].
It also includes drive-by changes of removing whitespace from searches[[1](https://github.com/gitpod-io/gitpod/issues/8454)].

Screenshot 2022-03-02 at 00 54 02

## Related Issue(s)

Fixes #8453
Fixes #8454 

## How to test
1. Go to https://lm-search-ws-8453.staging.gitpod-dev.com/admin/workspaces
2. Search part of a contextURL - it should not work.

## Release Notes

```release-note
Free text search on workspace admin dashboard is not enabled anymore.
```

### DIFF
--- a/components/dashboard/src/admin/WorkspacesSearch.tsx
+++ b/components/dashboard/src/admin/WorkspacesSearch.tsx
@@ -68,27 +68,19 @@ export function WorkspaceSearch(props: Props) {
     const search = async () => {
         setSearching(true);
         try {
-            let searchTerm: string | undefined = queryTerm;
-            const query: AdminGetWorkspacesQuery = {
-                ownerId: props?.user?.id,
-            };
-            if (matchesInstanceIdOrLegacyWorkspaceIdExactly(searchTerm)) {
-                query.instanceIdOrWorkspaceId = searchTerm;
-            } else if (matchesNewWorkspaceIdExactly(searchTerm)) {
-                query.workspaceId = searchTerm;
-            }
-            if (query.workspaceId || query.instanceId || query.instanceIdOrWorkspaceId) {
-                searchTerm = undefined;
+            const query: AdminGetWorkspacesQuery = {};
+            if (matchesInstanceIdOrLegacyWorkspaceIdExactly(queryTerm)) {
+                query.instanceIdOrWorkspaceId = queryTerm;
+            } else if (matchesNewWorkspaceIdExactly(queryTerm)) {
+                query.workspaceId = queryTerm;
             }
 
-            // const searchTerm = searchTerm;
             const result = await getGitpodService().server.adminGetWorkspaces({
                 limit: 100,
                 orderBy: 'instanceCreationTime',
                 offset: 0,
                 orderDir: "desc",
                 ...query,
-                searchTerm,
             });
             setSearchResult(result);
         } finally {
@@ -104,7 +96,9 @@ export function WorkspaceSearch(props: Props) {
                             <path fillRule="evenodd" clipRule="evenodd" d="M6 2a4 4 0 100 8 4 4 0 000-8zM0 6a6 6 0 1110.89 3.477l4.817 4.816a1 1 0 01-1.414 1.414l-4.816-4.816A6 6 0 010 6z" fill="#A8A29E" />
                         </svg>
                     </div>
-                    <input type="search" placeholder="Search Workspaces" onKeyDown={(ke) => ke.key === 'Enter' && search() } onChange={(v) => { setQueryTerm(v.target.value) }} />
+                    <input type="search" placeholder="Search Workspace IDs"
+                        onKeyDown={(ke) => ke.key === 'Enter' && search() }
+                        onChange={(v) => { setQueryTerm((v.target.value).trim()) }} />
                 </div>
                 <button disabled={searching} onClick={search}>Search</button>
             </div>
--- a/components/dashboard/src/admin/ProjectsSearch.tsx
+++ b/components/dashboard/src/admin/ProjectsSearch.tsx
@@ -97,7 +97,7 @@ export function ProjectsSearch() {
                             <path fillRule="evenodd" clipRule="evenodd" d="M6 2a4 4 0 100 8 4 4 0 000-8zM0 6a6 6 0 1110.89 3.477l4.817 4.816a1 1 0 01-1.414 1.414l-4.816-4.816A6 6 0 010 6z" fill="#A8A29E" />
                         </svg>
                     </div>
-                    <input type="search" placeholder="Search Projects" onKeyDown={(k) => k.key === 'Enter' && search()} onChange={(v) => { setSearchTerm(v.target.value) }} />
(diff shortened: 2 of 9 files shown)


components/dashboard/src/admin/WorkspacesSearch.tsx
```diff
@@ -68,27 +68,19 @@ export function WorkspaceSearch(props: Props) {
...
-            } else if (matchesNewWorkspaceIdExactly(searchTerm)) {
-                query.workspaceId = searchTerm;
-            }
-            if (query.workspaceId || query.instanceId || query.instanceIdOrWorkspaceId) {
-                searchTerm = undefined;
+            const query: AdminGetWorkspacesQuery = {};
+            if (matchesInstanceIdOrLegacyWorkspaceIdExactly(queryTerm)) {
+                query.instanceIdOrWorkspaceId = queryTerm;
+            } else if (matchesNewWorkspaceIdExactly(queryTerm)) {
+                query.workspaceId = queryTerm;
```
laushinka, line 75: @geropl What was the case before where we would populate `query.instanceId`?

geropl: @laushinka Not sure I understand your question? IMO we never were able to reliably fill `instanceId` (because `workspaceId`s always looked identical).

components/dashboard/src/admin/WorkspacesSearch.tsx
```diff
@@ -104,7 +96,9 @@ export function WorkspaceSearch(props: Props) {
                             <path fillRule="evenodd" clipRule="evenodd" d="M6 2a4 4 0 100 8 4 4 0 000-8zM0 6a6 6 0 1110.89 3.477l4.817 4.816a1 1 0 01-1.414 1.414l-4.816-4.816A6 6 0 010 6z" fill="#A8A29E" />
                         </svg>
                     </div>
-                    <input type="search" placeholder="Search Workspaces" onKeyDown={(ke) => ke.key === 'Enter' && search() } onChange={(v) => { setQueryTerm(v.target.value) }} />
+                    <input type="search" placeholder="Search Workspace IDs or Instance IDs" style={{width: "20rem"}}
```
gtsiolis, line 99: **question**: Shall we go with one of these here based on https://github.com/gitpod-io/gitpod/issues/8453#issuecomment-1055427520? Cc @geropl

components/dashboard/src/admin/WorkspacesSearch.tsx
laushinka: > Maybe `instanceId` was added with the previous change that introduced this "interpretation" of user input, to pave the way for a explicit choice (dropdown).

Makes sense!

laushinka:
Thank you for taking a look @geropl @gtsiolis 🙏🏽 It's now ready for actual review (not a draft anymore).

gtsiolis (approved):
Thanks, @laushinka!

Regarding the input placeholder question in https://github.com/gitpod-io/gitpod/pull/8503#discussion_r817490980:

I think @geropl was only suggesting to go with the `workspaceId` option instead of the `instanceId` as described in https://github.com/gitpod-io/gitpod/issues/8453#issuecomment-1055424756 but sounds ok to keep the friendlier placeholder.

components/dashboard/src/admin/WorkspacesSearch.tsx
```diff
@@ -104,7 +96,9 @@ export function WorkspaceSearch(props: Props) {
                             <path fillRule="evenodd" clipRule="evenodd" d="M6 2a4 4 0 100 8 4 4 0 000-8zM0 6a6 6 0 1110.89 3.477l4.817 4.816a1 1 0 01-1.414 1.414l-4.816-4.816A6 6 0 010 6z" fill="#A8A29E" />
                         </svg>
                     </div>
-                    <input type="search" placeholder="Search Workspaces" onKeyDown={(ke) => ke.key === 'Enter' && search() } onChange={(v) => { setQueryTerm(v.target.value) }} />
+                    <input type="search" placeholder="Search Workspace IDs"
+                        onKeyDown={(ke) => ke.key === 'Enter' && search() }
+                        onChange={(v) => { setQueryTerm((v.target.value).trim()) }} />
                 </div>
                 <button disabled={searching} onClick={search}>Search</button>
```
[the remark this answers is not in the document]
laushinka: *thinking* When there is no button, would the expectation be that the search result gets updated with every key change? Or would the user be expected to hit Enter?

components/dashboard/src/admin/WorkspacesSearch.tsx
```diff
@@ -104,7 +96,9 @@ export function WorkspaceSearch(props: Props) {
                             <path fillRule="evenodd" clipRule="evenodd" d="M6 2a4 4 0 100 8 4 4 0 000-8zM0 6a6 6 0 1110.89 3.477l4.817 4.816a1 1 0 01-1.414 1.414l-4.816-4.816A6 6 0 010 6z" fill="#A8A29E" />
                         </svg>
                     </div>
-                    <input type="search" placeholder="Search Workspaces" onKeyDown={(ke) => ke.key === 'Enter' && search() } onChange={(v) => { setQueryTerm(v.target.value) }} />
+                    <input type="search" placeholder="Search Workspace IDs"
```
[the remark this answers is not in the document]
laushinka: > We also scarcely reference the term workspace or instance `ID` in our docs.

Initially I thought that Admins surely know that those are workspace IDs, but I realize that that was my biased assumption and the fact that we don't reference the term `ID` anywhere makes me see your point.
Let's see how this goes - especially when self-hosted users give us feedback, and yes we can improve this anytime. 🙏🏽

components/dashboard/src/admin/WorkspacesSearch.tsx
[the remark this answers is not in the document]
gtsiolis: **thought**: There are cases where searching or filtering on key change is valid and useful like projects or members which usually are a finite and small data set. However, it sounds optimal to _not_ search on key change when filtering large datasets like instance users, workspaces, etc but require enter to perform the search. 💭 

Thinking out loud, we added the search button back in https://github.com/gitpod-io/gitpod/pull/3723 because we require the enter key to perform a search here but seems harmless to drop the button. Your call, feel free to move this discussion in a follow up issue. 🏓

geropl (approved):
Code LGTM and tested: works and looks nice! :+1:

Issue #5176, opened 2021-08-12
Rate limit prebuilds


jankeromnes:
### Bug description

Today at around 13:24 UTC, there was a sudden jump in the number of prebuilds: Over about a minute, the repository https://gitlab.com/yo/gitlab triggered nearly 100 new prebuilds.

Unfortunately, this put excessive pressure on the US cluster, and in particular on the database, which lead to several workspaces not being able to start (for example: https://community.gitpod.io/t/django-pod-is-down/4541).

This incident lasted about 20 minutes and was logged here: https://www.gitpodstatus.com/incidents/ddyzghh3mfbm

### Steps to reproduce

Instantly trigger 100 prebuilds

### Expected behavior

There should be no outage, especially in the database

### Anything else?

Maybe it would make sense to introduce a reasonable rate limit on prebuilds, for example per repository.

AlexTugarev:
Looking into the old solution for now: https://github.com/gitpod-io/gitpod/blob/main/components/server/ee/src/prebuilds/prebuild-queue-maintainer.ts

I hope, @csweichel or @geropl find time to provide more details on this, please 🙏🏻

geropl:
This just lead to an incident: https://app.incident.io/incidents/101
Thus added it to #7812

/cc @jldec

jldec:
Scheduled this issue.
Similar to https://github.com/gitpod-io/gitpod/issues/7400

jldec:
Remove highest priority label - not 🔥.

jldec:
Back on the urgent list after another close call ([internal](https://gitpod.slack.com/archives/C01KGM9DVRC/p1645984314944559))
cc: @kylos101

jankeromnes:
Happy to look into implementing a reasonable rate limit that doesn't break normal usage but prevents incidents.

jankeromnes:
I did some [research](https://github.com/gitpod-io/ops/issues/1373#issuecomment-1055338729) (internal) on our top prebuild rates for February 2022:

> Research reveals three top rate classes:
> - "Abuse-like" (50-85+ / min)
> - "Dependabot" (25-50 / min)
> - "Popular & mirrors" (up to 25 / min)
> 
> I think we'll want to stop the "Abuse-like", but continue supporting the "Popular & mirrors".

Also, the "Dependabot" case is interesting:
[...]

PR #8504, opened 2022-03-01
Rate-limit new prebuilds to 50 / user / minute to prevent incidents

jankeromnes:
## Description

Rate-limit new prebuilds to 50 / user / minute to prevent incidents

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/5176

## How to test

1. Fork https://gitlab.com/jankeromnes/test-gitpod-prebuilds-rate-limit
2. Add your fork as a project here: https://jx-rt-lmt-prblds.staging.gitpod-dev.com/new?team=user
3. Open your fork in Gitpod (any Gitpod) and run `./push-100.sh`
4. Once all 100 commits have been pushed (20 commits to 5 branches), verify that:
    - most prebuilds actually started (but not all 100)
    - some prebuilds never started but got cancelled instead (between 0 and 50, depending on when the burst happened in the 1min rate-limit window)
    - `kubectl logs server-[TAB] server | grep -i "rate limit"` contains some rate limit warnings (as many as the cancelled prebuilds)

## Release Notes

```release-note
Rate-limit new prebuilds to 50 / user / minute to prevent incidents
```

### DIFF
--- a/components/server/src/websocket/websocket-connection-manager.ts
+++ b/components/server/src/websocket/websocket-connection-manager.ts
@@ -358,7 +358,7 @@ class GitpodJsonRpcProxyFactory<T extends object> extends JsonRpcProxyFactory<T>
                     throw rlRejected;
                 }
                 log.warn({ userId }, "Rate limiter prevents accessing method due to too many requests.", rlRejected, { method });
-                throw new ResponseError<RateLimiterError>(ErrorCodes.TOO_MANY_REQUESTS, "too many requests", { method, retryAfter: Math.round(rlRejected.msBeforeNext / 1000) || 1 });
+                throw new ResponseError<RateLimiterError>(ErrorCodes.TOO_MANY_REQUESTS, "too many requests", { method, retryAfter: Math.ceil(rlRejected.msBeforeNext / 1000) || 1 });
             }
 
             // access guard
--- a/components/server/ee/src/prebuilds/prebuild-manager.ts
+++ b/components/server/ee/src/prebuilds/prebuild-manager.ts
@@ -5,7 +5,7 @@
  */
 
 import { DBWithTracing, TracedWorkspaceDB, WorkspaceDB } from '@gitpod/gitpod-db/lib';
-import { CommitContext, Project, ProjectEnvVar, StartPrebuildContext, StartPrebuildResult, TaskConfig, User, WorkspaceConfig, WorkspaceInstance } from '@gitpod/gitpod-protocol';
+import { CommitContext, Project, ProjectEnvVar, RateLimiterError, StartPrebuildContext, StartPrebuildResult, TaskConfig, User, WorkspaceConfig, WorkspaceInstance } from '@gitpod/gitpod-protocol';
 import { log } from '@gitpod/gitpod-protocol/lib/util/logging';
 import { TraceContext } from '@gitpod/gitpod-protocol/lib/util/tracing';
 import { inject, injectable } from 'inversify';
@@ -16,6 +16,9 @@ import { ConfigProvider } from '../../../src/workspace/config-provider';
 import { WorkspaceStarter } from '../../../src/workspace/workspace-starter';
 import { Config } from '../../../src/config';
 import { ProjectsService } from '../../../src/projects/projects-service';
+import { startPrebuild, UserRateLimiter } from '../../../src/auth/rate-limiter';
+import { ErrorCodes } from '@gitpod/gitpod-protocol/lib/messaging/error';
+import { ResponseError } from 'vscode-ws-jsonrpc';
 
 export class WorkspaceRunningError extends Error {
     constructor(msg: string, public instance: WorkspaceInstance) {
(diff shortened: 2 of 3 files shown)


jankeromnes:
Re-roll 🎲

/werft run

jankeromnes:
**Expected**
- 100 commits pushed (20 to 5 branches)
- 50 prebuilds started
- 50 rate limit warnings

**Current state**
- 100 commits pushed (20 to 5 branches)
- 100 prebuilds started
- 13 rate limit warnings

Something seems off here 🤔 (do GitLab webhooks auto-retry on rate limit? 😳)

gtsiolis:
See also [relevant discussion](https://gitpod.slack.com/archives/C02EN94AEPL/p1646081001709509) (internal) in case this is related. Cc @jankeromnes @JanKoehnlein @jldec

jankeromnes:
Ah, cool, I didn't know we could see webhook errors on GitLab. Thanks @gtsiolis!

Would be curious to see if we can get the actual error messages when a webhook call fails.

gtsiolis:
> Would be curious to see if we can get the actual error messages when a webhook call fails.

@jankeromnes Yes, we can. Here're the two most recent errors seen in our fork.
1. https://gitlab.com/gitpod-io/gitlab/-/hooks/10823912/hook_logs/2455346235 (internal)
2. https://gitlab.com/gitpod-io/gitlab/-/hooks/10823912/hook_logs/2455045008 (internal)

jankeromnes:
Hey, thanks a lot for surfacing these. 💯🙏

Both seem to have failed with:

```
Internal error occurred while delivering this webhook.
Error: Net::ReadTimeout
```

I wonder if the timeout means Gitpod didn't respond in time, or if it's something inside GitLab that timed out. 🤔

In any case, this feature looks super useful for debugging webhook problems (especially now that we're inventing a new webhook problem: the rate limit 😅)

jankeromnes:
Two new errors on the current deployment, and rate-limiting doesn't work 🙄

```
TypeError: Cannot read properties of undefined (reading 'statusCode')
    at /app/node_modules/@gitbeaker/node/dist/index.js:168:43
    at step (/app/node_modules/@gitbeaker/node/dist/index.js:82:23)
    at Object.next (/app/node_modules/@gitbeaker/node/dist/index.js:63:53)
    at fulfilled (/app/node_modules/@gitbeaker/node/dist/index.js:53:58)
```

```
(node:29) UnhandledPromiseRejectionWarning: Error: NotFoundError
    at Object.<anonymous> (/app/node_modules/@gitpod/server/dist/src/errors/index.js:26:41)
    at /app/node_modules/@gitpod/server/dist/src/errors/index.js:13:71
    at new Promise (<anonymous>)
    at __awaiter (/app/node_modules/@gitpod/server/dist/src/errors/index.js:9:12)
    at Object.create (/app/node_modules/@gitpod/server/dist/src/errors/index.js:21:16)
    at GitlabContextParser.<anonymous> (/app/node_modules/@gitpod/server/dist/src/gitlab/gitlab-context-parser.js:187:52)
    at fulfilled (/app/node_modules/@gitpod/server/dist/src/gitlab/gitlab-context-parser.js:19:58)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

components/server/src/websocket/websocket-connection-manager.ts
```diff
@@ -358,7 +358,7 @@ class GitpodJsonRpcProxyFactory<T extends object> extends JsonRpcProxyFactory<T>
                     throw rlRejected;
                 }
                 log.warn({ userId }, "Rate limiter prevents accessing method due to too many requests.", rlRejected, { method });
-                throw new ResponseError<RateLimiterError>(ErrorCodes.TOO_MANY_REQUESTS, "too many requests", { method, retryAfter: Math.round(rlRejected.msBeforeNext / 1000) || 1 });
+                throw new ResponseError<RateLimiterError>(ErrorCodes.TOO_MANY_REQUESTS, "too many requests", { method, retryAfter: Math.ceil(rlRejected.msBeforeNext / 1000) || 1 });
```
geropl, line 361: :+1:

geropl:
@jankeromnes Thx for tackling this long-standing yet very annoying issue. Two thoughts:
 - the current (draft) version throttles per user. With an additional method [here](https://github.com/gitpod-io/gitpod/blob/35e97c10643734396c4368240d453ad734c82bdf/components/server/src/auth/rate-limiter.ts#L253-L255) we could turn it into per-repo
 - have you thought about using the DB as ground-truth about how many workspace are already running per repo? :thinking: This would:
   - effectively share rate-limiting state across `server` instances: the current approach allows for 6 x 50 = 300 prebuilds / 1 minutes, potentially
   - this could pave the way to support user-friendlier modes of handling over-traffic compared to "accept first, drop all others"

jankeromnes:
> > effectively share rate-limiting state across `server` instances: the current approach allows for 6 x 50 = 300 prebuilds / 1 minutes, potentially
> 
> Aha, right, I'm currently testing with a single `server` instance, which is why I didn't observe this.

I've tested this PR with 6 servers, using `kubectl scale deployments/server --replicas=6`.

```
mysql> select state, count(id) prebuilds from d_b_prebuilt_workspace where creationTime > '2022-03-02 14:42' group by state order by
prebuilds desc;
+-----------+-----------+
| state     | prebuilds |
+-----------+-----------+
| available |       100 |
+-----------+-----------+
1 row in set (0.01 sec)
```

```
mysql> select state, count(id) prebuilds from d_b_prebuilt_workspace where creationTime > '2022-03-02 14:46' group by state order by
prebuilds desc;
+-----------+-----------+
| state     | prebuilds |
+-----------+-----------+
| available |       100 |
+-----------+-----------+
1 row in set (0.01 sec)
```

Looks like this defeats the rate-limiter. 😞

jankeromnes:
While this approach (i.e. using the server rate-limiter inside the PrebuildManager) basically works, it has the problem of multiplying the allowed rate limit by the number of server instances in the cluster.

I still believe that we could fix the rate-limiter itself to use the DB in order to share a single state between all server instances of a cluster (thus fixing this approach, and also fixing every other rate-limited server method).

However, for the specific issue of prebuild spam, we've [discussed](https://github.com/gitpod-io/gitpod/issues/5176#issuecomment-1057906713) another approach (i.e. not using the rate-limiter, but making the PrebuildManager check the DB for how many prebuilds were previously started on the same cloneURL), which seems even simpler and could immediately prevent this type of outage.

Thus, closing this Pull Request for now. @andrew-farries and @easyCZ, if you do find anything useful in this PR, please feel free to recycle it as you see fit. 🙂 Many thanks for taking this over! I'm also happy to help with any questions here.

PR #8511, opened 2022-03-01
[installer/openvsx] Explicitly add statusfield of statefulset

ArthurSens:
## Description

While trying to spin up preview environments on Harvester, we were getting failures applying the gitpod config as shown [here](https://werft.gitpod-dev.com/job/gitpod-build-new-image-test-12.0):
image

It is a strange error since the status field of a stateful set should not be set before applying it, but should be populated and managed by the Kubernetes API.

If we decide to apply a statefulset with the status field though, there are some [required fields that need to be set](https://github.com/kubernetes/kubernetes/issues/78203).

I couldn't find a way to make the installer not render the openvsx-proxy statefulset without the status field, but managed to make CI pass if I add the required fields. **I'm opening this PR to ask for feedback, I'd rather prevent the status field from rendering altogether**

## Related Issue(s)

Fixes #

## How to test

Create a fresh new environment from this PR

## Release Notes

```release-note
installer: Add required fields to openvsx statefulset manifests
```

### DIFF
--- a/install/installer/pkg/components/openvsx-proxy/statefulset.go
+++ b/install/installer/pkg/components/openvsx-proxy/statefulset.go
@@ -6,6 +6,7 @@ package openvsx_proxy
 
 import (
 	"fmt"
+
 	"github.com/gitpod-io/gitpod/installer/pkg/cluster"
 
 	"github.com/gitpod-io/gitpod/installer/pkg/common"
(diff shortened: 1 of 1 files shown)


MrSimonEmms:
This PR seems to be a sensible solution to the problem. It appears that the Kubernetes `StatefulSet` struct doesn't have `omitempty` on the `Replicas`, which it does on others (see `Deployment`). This means that it adds `replicas: 0` when generating the YAML.

We want this fix in the Installer as it's likely that it'll affect other users - as it's an edge case, investigation and diagnosis will be time-consuming. Therefore, this is not something that we should post-process and treat as a Gitpod fix.

The imported Helm charts are not a problem as they manually write the YAML as a Golang template - we're converting the struct into YAML and so are bound by the Golang rules.

Other options we looked at:
1. create a custom struct with the correct `omitempty` status
2. post-process the YAML inside the Installer

Both of these seemed to be an awful lot of effort for (hopefully) one small issue. In the event that we find other things like this, we should consider one of the above two options. @gitpod-io/engineering-self-hosted

MrSimonEmms (approved):
This is a sensible approach to a misconfigured struct

@ArthurSens should we submit a bug report/PR to the Kubernetes maintainers?

ArthurSens:
> This is a sensible approach to a misconfigured struct
> 
> @ArthurSens should we submit a bug report/PR to the Kubernetes maintainers?

Still feels awkward that this hasn't been discovered yet, kubernetes client-go is one of the most used Go libraries in the world... I believe there is something off in our set up 🤔 

@meysholdt the build failed with a valid failure, I think we need to address that instead of forcing the green lights 😬 
image

ArthurSens:
Yeah, but the installer is rendering it anyways 🤔 

If possible, I'd very much rather not render status at all

ArthurSens:
Related issue https://github.com/gitpod-io/gitpod/issues/8529

PR #8518, opened 2022-03-01
try harder when attempting to add or remove annotation from workspaces

sagor999:
## Description

In GCP logs I observed this error pop up quite often: 
`cannot workspace: cannot mark workspace redacted with -gitpod/never-ready: Operation cannot be fulfilled on pods "ws-redacted": the object has been modified; please apply your changes to the latest version and try again`
https://cloudlogging.app.goo.gl/d9NjYg6x9NgPfumeA

This is caused by default backoff being too short and after 4 attempts to update pod with updated annotation, it fails and marks whole workspace as failed. 
So instead this PR make it try harder.

## Related Issue(s)

Fixes #

## How to test

There is no easy way to test this, as it happens when some other process updates pod while we are trying to call `markWorkspace` function. 

## Release Notes

```release-note
Try harder to update workspace annotation to prevent workspace from being marked as failed erroneously.
```

### DIFF
--- a/components/ws-manager/pkg/manager/annotations.go
+++ b/components/ws-manager/pkg/manager/annotations.go
@@ -7,8 +7,10 @@ package manager
 import (
 	"context"
 	"strings"
+	"time"
 
 	"golang.org/x/xerrors"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
 
 	"github.com/gitpod-io/gitpod/common-go/log"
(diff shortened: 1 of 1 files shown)


sagor999:
https://github.com/gitpod-io/gitpod/pull/8519

components/ws-manager/pkg/manager/annotations.go
```diff
@@ -79,8 +81,17 @@ const (
...
+	// would cause whole workspace to be marked as failed, hence the reason to try harder here.
+	var backoff = wait.Backoff{
+		Steps:    7,
+		Duration: 100 * time.Millisecond,
+		Factor:   2.0,
+		Jitter:   0.1,
+	}
 	// Retry on failure. Sometimes this doesn't work because of concurrent modification. The Kuberentes way is to just try again after waiting a bit.
-	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+	err := retry.RetryOnConflict(backoff, func() error {
```
Furisto, line 94: @sagor999 Lets try it. Could it be that the pod is cached by the informer and the reason for the failure is that we get a stale version on each retry?

sagor999: I don't think so, since `findWorkspacePod` pulls pod info on each retry.

PR #8530, opened 2022-03-02
Set the timeout for the server deployment to 10min

wulfthimm:
## Description

Installing Gitpod in a VM on Harvester lead to a timeout during the server deployment. Using twice the time should prevent the deployment from failing.

## Related Issue(s)

Fixes #

## How to test

Create a new VM using werft.

## Release Notes

```release-note
None
```


ArthurSens (changes requested):
The timeout we're experiencing is when checking the rollout of the `server` (the gitpod component), and not when starting the VM 😅 

This happens during the werft phase "monitor server deployment". Could you change this PR to extend the timeout of [this line](https://github.com/gitpod-io/gitpod/blob/9880270118879cc8e0aa74bd8c889b3661bd38e9/.werft/jobs/build/deploy-to-preview-environment.ts#L459)?

ArthurSens (approved):

PR #8538, opened 2022-03-02
Add Replicated/KOTS dependencies to image

MrSimonEmms:
## Description

Add Replicated and KOTS dependencies to the main image. This allows us to do the development required in `/install/kots`

- [Replicated docs](https://kots.io/vendor/cli/getting-started/#installing-the-vendor-cli)
- [KOTS docs](https://kots.io/kots-cli/getting-started/#how-to-install)

## Release Notes

```release-note
NONE
```

### DIFF
--- a/dev/image/Dockerfile
+++ b/dev/image/Dockerfile
@@ -231,5 +231,9 @@ RUN curl -LO https://github.com/brancz/gojsontoyaml/releases/download/v${GOJSONT
     tar -xzvf gojsontoyaml_${GOJSONTOYAML_VERSION}_linux_amd64.tar.gz && \
     sudo mv gojsontoyaml /usr/local/bin/gojsontoyaml
 
+# Install Replicated and KOTS
+RUN curl https://raw.githubusercontent.com/replicatedhq/replicated/v0.38.0/install.sh | sudo bash && \
+    curl https://kots.io/install/1.65.0 | bash
+
 # Copy our own tools
 COPY dev-kubecdl--app/kubecdl dev-gpctl--app/gpctl /usr/bin/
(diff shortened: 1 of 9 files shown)


dev/image/Dockerfile
```diff
@@ -231,5 +231,9 @@ RUN curl -LO https://github.com/brancz/gojsontoyaml/releases/download/v${GOJSONT
     tar -xzvf gojsontoyaml_${GOJSONTOYAML_VERSION}_linux_amd64.tar.gz && \
     sudo mv gojsontoyaml /usr/local/bin/gojsontoyaml
 
+# Install Replicated and KOTS
+RUN brew install replicatedhq/replicated/cli kubectl helm && \
+    curl https://kots.io/install | bash
```
corneliusludmann, line 236: My 2 cents on this:
- `kubectl` is already part of the image, we don't need to install it here again.
- I would prefer to avoid using `brew` and use the [Linux install step](https://github.com/replicatedhq/replicated#linux-install) instead:
  ```
  $ curl https://raw.githubusercontent.com/replicatedhq/replicated/master/install.sh | bash
  ```
- Both, the Replicated CLI install bash script and the KOTS install bash script simply download the latest release from the GitHub release page. We could also download the binaries directly (and check the sha hash etc.). However, then we need to take care that we update the version regularly.
- We used to have `helm` in the image. You could use the commit https://github.com/gitpod-io/gitpod/commit/88ff2c271deb402f1f950b68be21c0d827b88c59 as a reference.

MrSimonEmms: Yeah, sorry my bad. I copy/pasta from the old PR and forgot to remove the relevant stuff. Will update

meysholdt: I support your points, Cornelius. 
> then we need to take care that we update the version regularly.

+1 for version locking. Upgrading the versions from time to time manually is usually less painful than stuff breaking when you have the least time for and waste hours until you realize that an automated version update happened.

MrSimonEmms: Made the fixes and added version locks.

corneliusludmann (approved):
looks good to me

@MrSimonEmms can the `hold` label be removed?

ArthurSens (approved):

PR #8542, opened 2022-03-02
Temporarily comment out GitpodBranding to fix workspace not connecting through JetBrains IDEs

felladrin:
## Description

Temporarily comment out GitpodBranding to fix workspace not connecting through JetBrains IDEs.

## Related Issue(s)

NONE

## Release Notes

```release-note
NONE
```


akosyakov (approved):
lgtm

akosyakov:
@iQQBot we don't know, but JB folks confirmed that it is a root cause. We don't know more.

PR #8544, opened 2022-03-02
Fix launch-dev-server script to work also from VS Code

felladrin:
## Description

Fix launch-dev-server script to work also from VS Code.

There were some missing environment variables in the script.

Also, we now set the port in a later moment, in [ControllerStatusService.kt](https://github.com/gitpod-io/gitpod/pull/8544/files#diff-d153ea9286fb391f993f849a5883a05ef9669a2b095ada06e7f1a5f7204cdda0).

## Release Notes

```release-note
NONE
```


akosyakov (approved):
lgtm

PR #8547, opened 2022-03-02
[installer]: add annotation to make DB resources restart if changes

MrSimonEmms:
## Description

Add the database type (in-cluster, external, cloudsql) as an annotation on the long-running resources that connect to the database. This is to force a restart of these resources if the database type is changed.

This has become important to support with KOTS. This can cause a bug where the "login with GitHub" button never shows.

## Release Notes

```release-note
[installer]: add annotation to make DB resources restart if changes
```

### DIFF
--- a/install/installer/pkg/components/server/deployment.go
+++ b/install/installer/pkg/components/server/deployment.go
@@ -26,7 +26,38 @@ import (
 func deployment(ctx *common.RenderContext) ([]runtime.Object, error) {
 	labels := common.DefaultLabels(Component)
 
-	configHash, err := common.ObjectHash(configmap(ctx))
+	var hashObj []runtime.Object
+	if objs, err := configmap(ctx); err != nil {
+		return nil, err
+	} else {
+		hashObj = append(hashObj, objs...)
+	}
+
+	hashObj = append(hashObj, &corev1.Pod{
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					Env: []corev1.EnvVar{
+						// If the database type changes, this pod may stay up if no other changes are made.
+						{
+							Name: "DATABASE_TYPE",
+							Value: func() string {
+								if pointer.BoolDeref(ctx.Config.Database.InCluster, false) {
+									return "in-cluster"
+								}
+								if ctx.Config.Database.CloudSQL != nil {
+									return "cloudsql"
+								}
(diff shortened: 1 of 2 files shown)


geropl:
@MrSimonEmms makes absolute sense to add this!

Should we extend this to all DB connection information? Like user and password as well?

install/installer/pkg/components/server/deployment.go
```diff
@@ -26,7 +26,37 @@ import (
...
+
+	hashObj = append(hashObj, &corev1.Pod{
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					Env: []corev1.EnvVar{
+						{
+							Name: "DATABASE_TYPE",
+							Value: func() string {
+								if pointer.BoolDeref(ctx.Config.Database.InCluster, false) {
```
geropl, line 44: I understand that it's not required, but: would it make sense to share code between all DB-dependant deployments? :thinking:

geropl:
> The external/cloudsql creds are stored in a secret which we just refer to

Can we hash the secret then? :thinking:

install/installer/pkg/components/server/deployment.go
geropl: wrt duplication: Yeah, in this case it might actually be counter-productive, as the only dependency is within the deployment itself. If we changed a single function, we might inadvertently change other as well.

I think what I'm missing most is a method name/comment that conveys the intention: what we do, and why these values are the important ones to establish identity.

MrSimonEmms:
> Can we hash the secret then? thinking

I don't think so. Because the Installer doesn't have access to the cluster at render time, we don't know what the contents of the secrets are to create a hash from

install/installer/pkg/components/server/deployment.go
MrSimonEmms: Yeah, I'll happily add a comment

MrSimonEmms: Comment added to both instances

geropl:
> Because the Installer doesn't have access to the cluster at render time

Thx for the :bulb: , makes sense.

> put a randomly generated value in the annotations so they ALWAYS reload

That :100: won't for for SaaS. Let's merge this as-is.

geropl (approved):
LGTM :+1:

Issue #8536, opened 2022-03-02
Projects support for Gitpod Self-hosted Instances


MaxAlber:
## Is your feature request related to a problem? Please describe
Right now the Project feature is not working for self-hosted Gitpod installations. As far as i know, its simply not supported yet.

## Describe the behaviour you'd like
It would be nice to have the feature enabled for self-hosted instances, since it's kind of a core feature when working in a team.

## Additional context
Right now, when starting a project, its stuck on "loading". I don't know how difficult it is to implement this feature, since i don't know much of the internal structure of gitpod. Would be nice if someone with more expertise on this subject could add some information.

MrSimonEmms:
See:
- [#7781](https://github.com/gitpod-io/gitpod/issues/7781) for GitLab
- [#5297](https://github.com/gitpod-io/gitpod/issues/5297)+[#8501](https://github.com/gitpod-io/gitpod/issues/8501) for GitHub.

elijahworkz:
I have just installed self-hosted Gitpod by following EKS guide. I did have some issues connecting it to Bitbucket, but once I was able to do that - Projects are working for me with no issues.

jldec:
See #9163 for using Gitpod self-hosted instances with projects pointing to repos on `github.com`

PR #8550, opened 2022-03-02
Support private registries

### LINKED ISSUE #1699: Enable use of private Docker images
Please add support for private docker registries in Gitpod - especially the self-hosted version. 

We use a lot of private Docker images in our various stacks - also for development. 

Not being able to set this up for Gitpod means that we are unable to configure Gitpod workspaces for anything but simple repo editing. 

I would expect that for the self-hosted version, using secrets or docker credentials from kubernetes would be a good option to support this.

csweichel:
## Description
This PR adds rudimentary support for images from private registries. It does that by virtue of project-level environment variables.

To use it, add an environment variable named `GITPOD_IMAGE_AUTH` to your project. The content of that environment variable looks like:
`domain.com:<base64(username:password)>, anotherdomain.com:...`

This format is akin to dockerconfig.json files, where the `auth` section contains the same base64 encoded string. For example, logging into to the GitHub registry (ghcr.io) using:
```
echo $GITHUB_TOKEN | docker login ghcr.io -u USERNAME --password-stdin
```
produces an entry akin to:
```JSON
"auths": {
	"ghcr.io": {
		"auth": "WW91IGNoZWNrZWQgOikgTmVlZCBhIGpvYj8gRW1haWwgbWUgYXQgY2hyaXNAZ2l0cG9kLmlvCg=="
	}
},
```
That very base64 encoded string can be used. It would turn into:
image

Once this variable is present, private images can be used directly, e.g.
```YAML
image: my-private-repo.com/super-secret-sauce:latest
```

## Caveats
- the Docker image is copied over to Gitpod's registry. There currently is no means to delete that image.
- with this implementation, private images can only be used directly, not as part of a Dockerfile (i.e. `FROM my-private-repo.com/secret-sauce:latest`)
- ~right now the project level environment variable must NOT be "hidden". I'm looking to change that.~
- the failure modes aren't great. There are a lot of things which might not work, e.g. using the wrong format. Users don't get a good error message, it's just that their image pull fails with a "cannot resolve base image: hostname required" error.
- this PR removes the base image resolution as part of the workspace config creation. Consequently, if the image referred to in the `.gitpod.yml` refers to a "moving tag" (think `docker.io/somthing:latest`), and we need to rebuild the "gitpod workspace image", we'd re-resolve that tag, which might lead to changing workspace images across restarts for the very same workspace.
- I've been a bit coarse with the project level environment variables - we can probably clean some of that up.

## Related Issue(s)

Fixes #1699

## How to test
See PR description. I tested with ghcr.io

## Release Notes

```release-note
Add support for private registries
```

## Documentation

https://github.com/gitpod-io/website/issues/1704


components/gitpod-db/src/project-db.ts
```diff
@@ -18,7 +18,7 @@ export interface ProjectDB {
     updateProject(partialProject: PartialProject): Promise<void>;
     markDeleted(projectId: string): Promise<void>;
     setProjectEnvironmentVariable(projectId: string, name: string, value: string, censored: boolean): Promise<void>;
-    getProjectEnvironmentVariables(projectId: string): Promise<ProjectEnvVar[]>;
+    getProjectEnvironmentVariables(projectId: string): Promise<ProjectEnvVarWithValue[]>;
```
jankeromnes, line 21: **Security-in-depth**: Please revert this change.

components/gitpod-db/src/typeorm/project-db-impl.ts
```diff
@@ -162,11 +162,9 @@ export class ProjectDBImpl implements ProjectDB {
         });
     }
 
-    public async getProjectEnvironmentVariables(projectId: string): Promise<ProjectEnvVar[]> {
+    public async getProjectEnvironmentVariables(projectId: string): Promise<ProjectEnvVarWithValue[]> {
         const envVarRepo = await this.getProjectEnvVarRepo();
-        const envVarsWithValue = await envVarRepo.find({ projectId, deleted: false });
-        const envVars = envVarsWithValue.map(toProjectEnvVar);
-        return envVars;
+        return await envVarRepo.find({ projectId, deleted: false });
```
jankeromnes, line 167: **Security-in-depth**: Please revert this change.

components/server/ee/src/prebuilds/prebuild-manager.ts
```diff
@@ -162,7 +162,7 @@ export class PrebuildManager {
             if (!prebuild) {
                 throw new Error('No prebuild found for workspace ' + workspaceId);
             }
-            let projectEnvVars: ProjectEnvVar[] = [];
+            let projectEnvVars: ProjectEnvVarWithValue[] = [];
```
jankeromnes, line 165: **Security-in-depth**: Please revert this change.

components/server/src/projects/projects-service.ts
```diff
@@ -205,7 +205,7 @@ export class ProjectsService {
         return this.projectDB.setProjectEnvironmentVariable(projectId, name, value, censored);
     }
 
-    async getProjectEnvironmentVariables(projectId: string): Promise<ProjectEnvVar[]> {
+    async getProjectEnvironmentVariables(projectId: string): Promise<ProjectEnvVarWithValue[]> {
```
jankeromnes, line 208: **Security-in-depth**: Please revert this change.

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -1582,7 +1582,12 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
         traceAPIParams(ctx, { projectId });
         const user = this.checkAndBlockUser("getProjectEnvironmentVariables");
         await this.guardProjectOperation(user, projectId, "get");
-        return this.projectsService.getProjectEnvironmentVariables(projectId);
+        const res = (await this.projectsService.getProjectEnvironmentVariables(projectId)).map(ev => {
+            const envVar: ProjectEnvVar = { ...ev };
+            delete (envVar as any)['value'];
+            return envVar;
+        });
+        return res;
```
jankeromnes, line 1590: Please revert this change.

components/server/src/workspace/workspace-starter.ts
```diff
@@ -60,9 +60,8 @@ export class WorkspaceStarter {
     @inject(IAnalyticsWriter) protected readonly analytics: IAnalyticsWriter;
     @inject(TheiaPluginService) protected readonly theiaService: TheiaPluginService;
     @inject(OneTimeSecretServer) protected readonly otsServer: OneTimeSecretServer;
-    @inject(ProjectDB) protected readonly projectDB: ProjectDB;
 
-    public async startWorkspace(ctx: TraceContext, workspace: Workspace, user: User, userEnvVars: UserEnvVar[], projectEnvVars: ProjectEnvVar[], options?: StartWorkspaceOptions): Promise<StartWorkspaceResult> {
+    public async startWorkspace(ctx: TraceContext, workspace: Workspace, user: User, userEnvVars: UserEnvVar[], projectEnvVars: ProjectEnvVarWithValue[], options?: StartWorkspaceOptions): Promise<StartWorkspaceResult> {
```
jankeromnes, line 64: **Security-in-depth**: Please revert this change.

components/server/src/workspace/workspace-starter.ts
```diff
@@ -148,12 +148,13 @@ export class WorkspaceStarter {
     }
 
     // Note: this function does not expect to be awaited for by its caller. This means that it takes care of error handling itself.
-    protected async actuallyStartWorkspace(ctx: TraceContext, instance: WorkspaceInstance, workspace: Workspace, user: User, mustHaveBackup: boolean, ideConfig: IDEConfig, userEnvVars: UserEnvVar[], projectEnvVars: ProjectEnvVar[], rethrow?: boolean, forceRebuild?: boolean): Promise<StartWorkspaceResult> {
+    protected async actuallyStartWorkspace(ctx: TraceContext, instance: WorkspaceInstance, workspace: Workspace, user: User, mustHaveBackup: boolean, ideConfig: IDEConfig, userEnvVars: UserEnvVar[], projectEnvVars: ProjectEnvVarWithValue[], rethrow?: boolean, forceRebuild?: boolean): Promise<StartWorkspaceResult> {
```
jankeromnes, line 151: **Security-in-depth**: Please revert this change.

components/server/src/workspace/workspace-starter.ts
```diff
@@ -259,6 +260,21 @@ export class WorkspaceStarter {
         }
     }
 
+    protected getAdditionalImageAuth(projectEnvVars: ProjectEnvVarWithValue[]): Map<string, string> {
```
jankeromnes, line 263: I think you can "hydrate" these with values here (doing it even lower-level would be even better, but a bit more complicated, so here is fine).

```suggestion
    protected getAdditionalImageAuth(projectEnvVars: ProjectEnvVar[]): Map<string, string> {
        const projectEnvVarsWithValues = await this.projectDB.getProjectEnvironmentVariableValues(projectEnvVars);
```

components/server/src/workspace/workspace-starter.ts
```diff
@@ -674,8 +691,13 @@ export class WorkspaceStarter {
             }
         }
         if (projectEnvVars.length > 0) {
-            const projectEnvVarsWithValues = await this.projectDB.getProjectEnvironmentVariableValues(projectEnvVars);
-            allEnvVars = allEnvVars.concat(projectEnvVarsWithValues);
```
jankeromnes, line 678: Please also only "hydrate" values at this point, not earlier.

components/server/src/workspace/workspace-starter.ts
```diff
@@ -661,7 +678,7 @@ export class WorkspaceStarter {
         }
     }
 
-    protected async createSpec(traceCtx: TraceContext, user: User, workspace: Workspace, instance: WorkspaceInstance, mustHaveBackup: boolean, ideConfig: IDEConfig, userEnvVars: UserEnvVarValue[], projectEnvVars: ProjectEnvVar[]): Promise<StartWorkspaceSpec> {
+    protected async createSpec(traceCtx: TraceContext, user: User, workspace: Workspace, instance: WorkspaceInstance, mustHaveBackup: boolean, ideConfig: IDEConfig, userEnvVars: UserEnvVarValue[], projectEnvVars: ProjectEnvVarWithValue[]): Promise<StartWorkspaceSpec> {
```
jankeromnes, line 681: **Security-in-depth**: Please revert this change.

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -518,7 +518,7 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
             throw new ResponseError(ErrorCodes.PERMISSION_DENIED, "Cannot (re-)start a deleted workspace.");
         }
         const userEnvVars = this.userDB.getEnvVars(user.id);
-        let projectEnvVarsPromise = this.internalGetPublicProjectEnvVars(workspace.projectId);
+        let projectEnvVarsPromise = this.internalGetProjectEnvVars(workspace.projectId);
```
jankeromnes, line 521: With the filtering moved to another level, I believe you can delete the `internalGet[...]ProjectEnvVars` helper function, and just call this here:

```suggestion
        let projectEnvVarsPromise = this.projectsService.getProjectEnvironmentVariables(workspace.projectId);
```

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -908,7 +908,7 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
                 throw err;
             }
 
-            let projectEnvVarsPromise = this.internalGetPublicProjectEnvVars(workspace.projectId);
+            let projectEnvVarsPromise = this.internalGetProjectEnvVars(workspace.projectId);
```
jankeromnes, line 911: With the filtering moved to another level, I believe you can delete the `internalGet[...]ProjectEnvVars` helper function, and just call this here:

```suggestion
            let projectEnvVarsPromise = this.projectsService.getProjectEnvironmentVariables(workspace.projectId);
```

components/server/src/workspace/workspace-starter.ts
```diff
@@ -674,8 +691,13 @@ export class WorkspaceStarter {
             }
         }
         if (projectEnvVars.length > 0) {
-            const projectEnvVarsWithValues = await this.projectDB.getProjectEnvironmentVariableValues(projectEnvVars);
-            allEnvVars = allEnvVars.concat(projectEnvVarsWithValues);
+            // Instead of using an access guard for Project environment variables, we let Project owners decide whether
+            // a variable should be:
+            //   - exposed in all workspaces (even for non-Project members when the repository is public), or
+            //   - censored from all workspaces (even for Project members)
+            const publicPrjEnvVars = projectEnvVars.filter(variable => !variable.censored);
```
jankeromnes, line 698: Since you do need a few special env vars in this file, I'm okay with moving the censoring/filtering code here (although the "hydration" with values can still happen right above).

One question though: Does it ever make sense to allow private registries to not be `censored`? (I.e. expose the value in workspaces.) If it never makes sense, I guess we could always filter out private registries here. 💭

jankeromnes (commented):
Many thanks for implementing this super valuable feature! 🙏 💯

Briefly reviewed the environment variable code, and left a few suggestions in-line. TL;DR -- please only "hydrate" values just when they're actually needed, in order to not pass them around clear-text too much.

components/server/src/workspace/workspace-starter.ts
csweichel: > Does it ever make sense to allow private registries to not be censored? (I.e. expose the value in workspaces.) 

It does in a future iteration when we'd merge this value with the `~/docker/config.json` file, hence provide access to private repos in the workspace.

> If it never makes sense, I guess we could always filter out private registries here. 💭

If we did this now, we'd introduce another failure mode - without telling anyone. I.e. if a user were to make the variable "visible" things just wouldn't work, and they wouldn't know why.

components/server/src/workspace/gitpod-server-impl.ts
csweichel: It's still helpful to handle the "outside of a project" case

csweichel:
@jankeromnes Thank you for your feedback. I've reverted the early "hydration" of the project-level environment variables and pushed it back to the actual point of use. PTAL.

jankeromnes (commented):
Awesome, many thanks for the quick fix!

I've reviewed the environment variables handling code, and it looks good to me. 👍

I haven't looked too closely at the Go code or the `image-source-provider.ts` changes.

Please let me know if you'd like me to test this PR.

> To use it, add an environment variable named `GITPOD_IMAGE_AUTH` to your project. The content of that environment variable looks like:
> `domain.com:<base64(username:password)> anotherdomain.com:...`

Two minor thoughts:
- The code seems to split on `,` instead of ` ` -- I assume it's a typo here
- Why base64-encode the `username:password` part? (As a user, I'd find it more convenient to specify `domain.com:username:password` directly)

sagor999 (approved):
reviewed go related changes. only one minor comment.

sagor999:
>the Docker image is copied over to Gitpod's registry. There currently is no means to delete that image.

is that a local cluster level registry? or registry at eu.gcr.io? 
and can someone else download that image from local registry somehow later on?

components/image-builder-mk3/debug.sh
```diff
@@ -5,8 +5,12 @@
...
 
-kubectl patch deployment image-builder --patch '{"spec": {"template": {"spec": {"containers": [{"name": "service","image": "'$devImage'"}]}}}}'
-kubectl get pods --no-headers -o=custom-columns=:metadata.name | grep image-builder | xargs kubectl delete pod
\ No newline at end of file
+kubectl patch deployment image-builder-mk3 --patch '{"spec": {"template": {"spec": {"containers": [{"name": "image-builder-mk3","imagePullPolicy":"Always","image": "'$devImage'"}]}}}}'
+kubectl rollout restart deployment/image-builder-mk3
+kubectl rollout status -w deployment/image-builder-mk3
+# give the old pod time to disappear
+sleep 20
+gpctl debug logs image-builder-mk3
```
utam0k, line 16:
```suggestion
gpctl debug logs image-builder-mk3

```

components/image-builder-bob/pkg/proxy/auth.go
```diff
@@ -54,17 +54,44 @@ func (a authorizerImpl) Authorize(host string) (user, pass string, err error) {
...
+
+func NewAuthorizerFromEnvVar(content string) (auth MapAuthorizer, err error) {
+	if content == "" {
+		return nil, nil
+	}
+
+	var res map[string]authConfig
+	err = json.Unmarshal([]byte(content), &res)
+	if err != nil {
+		return
```
utam0k, line 94: Is there any reason for ignoring this error?
```suggestion
		return nil, err
```

csweichel: it's not ignored. Because `err` is a named return argument, the `return` will just return the value last assigned to `err`

csweichel:
> 

I just used `ghcr.io` and a GitHub access token.

components/image-builder-mk3/debug.sh
csweichel: I'm not sure I understand the comment :)

components/image-builder-mk3/debug.sh
```diff
@@ -5,8 +5,12 @@
 docker ps &> /dev/null || (echo "You need a working Docker daemon. Maybe set DOCKER_HOST?"; exit 1)
 gcloud auth list | grep typefox &>/dev/null || (echo "Login using 'gcloud auth login' for the docker push to work"; exit 1)
 
-leeway build .:docker -Dversion=dev
-devImage=eu.gcr.io/gitpod-dev/image-builder:dev
+leeway build -v .:docker -Dversion=cw-dev -DimageRepoBase=eu.gcr.io/gitpod-core-dev/build
+devImage=eu.gcr.io/gitpod-core-dev/build/image-builder-mk3:cw-dev
```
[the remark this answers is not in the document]
csweichel: nope - was a leftover. Fixed now :)

components/image-builder-mk3/debug.sh
utam0k: This is the fix that the github warns about because there is no new line. There is no specific problem, so it is a minor nits.
https://thoughtbot.com/blog/no-newline-at-end-of-file

components/image-builder-bob/pkg/proxy/auth.go
utam0k: @csweichel Oh, I didn't know about it. I just learned, thanks.

csweichel:
build is green again.

@geropl this needs your approval due to API changes.

geropl (approved):
:+1:

sagor999 (approved):

jmls:
looking forward to this .. a couple of questions

1) what happens if the custom image is updated ? Will it be re-copied to Gitpod's registry ?
2) Will this work with prebuilds ?

csweichel:
> looking forward to this .. a couple of questions
> 
> 1. what happens if the custom image is updated ? Will it be re-copied to Gitpod's registry ?
> 2. Will this work with prebuilds ?

yes :)
(it should at least)

jmls:
so now the million dollar question ... when ... ? :)

jmls:
ok, so hoping that "deployed" means it's deployed :) ... 

I do have a questtion:

does 

domain.com:<base64(username:password)>

mean  domain.com:<base64datahere> 
or
domain.com:base64datahere.

csweichel:
@jmls it's the latter: `domain.com:base64datahere`, for example: 
```bash
$ echo -n "domain.com:"; echo username:baseword | base64 -w0

# output
domain.com:dXNlcm5hbWU6YmFzZXdvcmQK
```

jmls:
@csweichel  thanks :) 

Any idea when this will be available to start using ?

csweichel:
@jmls beware: the change is not actually live yet - it's deployed by one of the teams. Once there's a `deployed` label (without a suffix) attached, it should be available.

jmls:
well, this is working and works great ! Saved us TB in bandwidth and startup times are much quicker. Great result.

ccll:
Does this suppose to work with Gitlab Registry?
I got an error trying to use image from my self-hosted Gitlab Registry from a self-hosted Gitpod.

Might be related with #8938 and #8795.

```
Request createWorkspace failed with message: 13 INTERNAL: cannot resolve base image: hostname required

Unknown Error: { "code": -32603 }
```

image


Due to issue #8536 I could not setup variable per project, so I added the variable to global settings.
image

jhulance:
I believe this doesn't work for self-hosted (yet?) because it makes use of Project vars, and the global Variables are ignored?

In order to get this working in the FROM line of a Dockerfile for a workspace, I believe the BOB_WSLAYER_AUTH env var needs to be set up by Image Builder MK3 when it starts a Image Build. This is the whole of the "auths" var from a standard Docker config.json with auth/creds for private registries. Not sure how that will get passed over to the Image Builder MK3 -- use ImagePullSecrets somewhere?

Keith-Hon:
I am unable to use a private image from hub.docker.com even after I added GITPOD_IMAGE_AUTH accordingly. Any help?

csweichel:
> I am unable to use a private image from hub.docker.com even after I added GITPOD_IMAGE_AUTH accordingly. Any help?

What do you see when you open the workspace?
Do you use the private image directly, or as part of a `Dockerfile`?

Keith-Hon:
These are the steps I took.

1. I logged in docker hub via cli and printed the token

Screenshot 2022-08-04 at 5 03 12 PM

2. Add a variable named GITPOD_IMAGE_AUTH in gitpod settings
 
Screenshot 2022-08-04 at 5 05 31 PM

3. Here is the result I get

Screenshot 2022-08-04 at 5 06 16 PM

jmls:
you need to add the repo to a gitpod project , then add the variable to the project, not personal

Keith-Hon:
I just tried create a project at https://gitpod.io/projects, and added the variable GITPOD_IMAGE_AUTH to the project's Settings 

Screenshot 2022-08-04 at 5 16 03 PM

and then create a new workspace from that project

Screenshot 2022-08-04 at 5 16 25 PM

jmls:
double and triple check the value of the token - I know this works because we're using this for every single one of our repos across dozens of developers every day

Getting the *value* of the token right was our biggest hurdle ;)

jmls:
The value is domain:<base64 encoded value>

echo -n "<domain>:"; echo <username>:<password> | base64 -w0

are you sure the domain is index.docker.io ?

Keith-Hon:
@jmls  I am not sure if the domain is index.docker.io. But that's what I get from the docker login command. I found that the token is not the same with "echo <username>:<password> | base64 -w0" and the one from .json.

csweichel:
@Keith-Hon 

Do you use the private image directly, or as part of a Dockerfile?
How do you refer to the image: `dockerhub-user/imagename:latest` or `index.docker.io/dockerhub-user/imagename:latest`?

jmls:
oh, I forgot that a custom image on the dockerfile is not supported atm

You have to specify a custom image in the gitpod.yml

Keith-Hon:
@csweichel 

dockerhub-user/imagename:latest

csweichel:
Please try using `docker.io` as host, i.e.

```
echo -n "docker.io:"; echo <username>:<password> | base64 -w0
```

Keith-Hon:
@csweichel, @jmls  it's working now. Thanks a lot!

However, this line of command didn't work

`echo -n "docker.io:"; echo <username>:<password> | base64 -w0`

I succeeded by using the token obtained from 

`docker login`
`cat  /home/gitpod/.docker/config.json`

and add GITPOD_IMAGE_AUTH to the project's Variables by

docker.io:TOKEN_OBTAINED

Keith-Hon:
Btw, it is a bit confusing that "Projects" tab resides inside user's Setting. I didn't notice there was a "project" specific scope setting and environment variables....

Keith-Hon:
> @csweichel, @jmls it's working now. Thanks a lot!
> 
> However, this line of command didn't work
> 
> `echo -n "docker.io:"; echo <username>:<password> | base64 -w0`
> 
> I succeeded by using the token obtained from
> 
> `docker login` `cat /home/gitpod/.docker/config.json`
> 
> and added GITPOD_IMAGE_AUTH to the project's Variables by
> 
> docker.io:TOKEN_OBTAINED

This method stopped working today...

with this error

[error.txt](https://github.com/gitpod-io/gitpod/files/9281167/error.txt)

Keith-Hon:
> @csweichel, @jmls it's working now. Thanks a lot!
> 
> However, this line of command didn't work
> 
> `echo -n "docker.io:"; echo <username>:<password> | base64 -w0`
> 
> I succeeded by using the token obtained from
> 
> `docker login` `cat /home/gitpod/.docker/config.json`
> 
> and added GITPOD_IMAGE_AUTH to the project's Variables by
> 
> docker.io:TOKEN_OBTAINED

I tried again with this approach today. It only works if 

1. the targeted docker image was public when the workspace was prebuilt, and then set to private in dockerhub
2. the tag in the .gitpod.yml must match that once-public-then-private image's tag

It doesn't work if the docker image was never public @csweichel

utam0k:
@Keith-Hon @jmls Thanks for your report! Can I ask you to create [a new issue](https://github.com/gitpod-io/gitpod/issues/new?assignees=&labels=type%3A+bug&template=bug_report.yml) to prioritize to address?

Keith-Hon:
https://github.com/gitpod-io/gitpod/issues/11982

Keith-Hon:
Ok created

> @Keith-Hon @jmls Thanks for your report! Can I ask you to create [a new issue](https://github.com/gitpod-io/gitpod/issues/new?assignees=&labels=type%3A+bug&template=bug_report.yml) to prioritize to address?

utam0k:
@Keith-Hon Thanks!

mrzarquon:
As a note here until docs are live - this snippet needs the second echo to have -n also, it's throwing a carriage return onto the string. Using the below example you don't need to extract the credential from config.json.

`echo -n "docker.io:"; echo -n <username>:<password> | base64 -w0`

Issue #8066, opened 2022-02-07
Run prettier formatting and test for formatting in builds


mads-hartmann:
To ensure consistent formatting of the build job TypeScript code we should:

1. Format all `ts` code in .werft with prettier
2. Add a prettier format check to our build and block PRs on it

Part of the internal RFC ["Decomposing the build job"](https://www.notion.so/gitpod/Decomposing-the-build-job-2327d3cd8dfa4520b0ad994ab1b49dfd)

mads-hartmann:
Grabbing this as it's blocked by https://github.com/gitpod-io/gitpod/issues/8065

mads-hartmann:
I won't finish this before I leave on vacation so moving it to a later iteration. The next step is to try and use the formatter that vscode ships with (see [here](https://github.com/gitpod-io/gitpod/pull/8245#issuecomment-1047198048) and [here](https://gitpod.slack.com/archives/C01KGM9BH54/p1645089446217389) for context)

mads-hartmann:
The meta team is working on adding prettier (https://github.com/gitpod-io/gitpod/pull/8655) - I will wait till they have implemented it and then extend it to cover werft. Marking as blocked and moving out of this iteration.

mads-hartmann:
Prettier is now being used by the WebApp team (see PR https://github.com/gitpod-io/gitpod/pull/8655) so it should be easy to add for everything in the .werft now ☺️

PR #8568, opened 2022-03-03
server: rate limit prebuilds by cloneURL

andrew-farries:
## Description

Introduce a rate-limit for prebuilds.

Prebuilds are rate limited by *cloneURL* in the last 1 minute to X. X either comes from Config, or defaults to 50.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/5176

## How to test

Same instructions as https://github.com/gitpod-io/gitpod/pull/8504
* To have an easier time, you can modify the `server-configmap.yaml` on your preview and a special rate limit for your clone URL
	```
	"https://gitlab.com/<USER>/test-gitpod-prebuilds-rate-limit.git": 10,
    ```

### Tests
With a rate limit of 10:
Screenshot 2022-03-07 at 11 59 59

## Release Notes

```release-note
Rate-limit workspace prebuilds to 50 per minute (rolling-window) by default, configurable through config.
```

## Documentation


roboquat:
@andrew-farries: Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://git.k8s.io/community/contributors/guide/release-notes.md) to remove it.

<details>

</details>

components/server/ee/src/prebuilds/prebuild-manager.ts
```diff
@@ -113,23 +117,32 @@ export class PrebuildManager {
...
-            log.debug("Created prebuild context", prebuildContext);
+
 
             const projectEnvVarsPromise = project ? this.projectService.getProjectEnvironmentVariables(project.id) : [];
             const workspace = await this.workspaceFactory.createForContext({span}, user, prebuildContext, contextURL);
-            const prebuildPromise = this.workspaceDB.trace({span}).findPrebuildByWorkspaceID(workspace.id)!;
+            const prebuild = await this.workspaceDB.trace({span}).findPrebuildByWorkspaceID(workspace.id)!;
+            if (!prebuild) {
+                throw new Error(`Failed to create a prebuild for: ${contextURL}`);
+            }
```
jankeromnes, line 127: 👍

(You can also remove the identical check a few lines below)

chart/values.yaml
```diff
@@ -332,6 +332,8 @@ components:
       timeoutSeconds: 300
     insecureNoDomain: false
     chargebeeProviderOptionsFile: "/chargebee/providerOptions"
+    prebuildLimiter:
+      '*': 10
```
jankeromnes, line 336: Not very important, but this low value looks a little bit risky for an example config. 100 could be safer. 💭

components/gitpod-db/src/workspace-db.spec.db.ts
```diff
@@ -504,5 +505,60 @@ import { DBWorkspaceInstance } from './typeorm/entity/db-workspace-instance';
...
+    private async storePrebuiltWorkspace(pws: PrebuiltWorkspace) {
+        await this.db.storePrebuiltWorkspace(pws)
+
+        const conn = await this.typeorm.getConnection();
+        const repo = conn.getRepository(DBPrebuiltWorkspace);
+
+        if (!!pws.creationTime) {
+            // MySQL requires the time format to be 2022-03-07 15:44:01.746141
+            // Looks almost like an ISO time string, hack it a bit.
+            const mysqlTimeFormat = pws.creationTime.replace("T", " ").replace("Z", "")
```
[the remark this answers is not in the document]
easyCZ: Intentionally only done in the test here as doing it properly and catching edge cases is gonna be much more work and out of scope. I'll cut a follow-up ticket to revisit the default value in the column.

components/server/ee/src/prebuilds/prebuild-manager.ts
```diff
@@ -113,26 +117,32 @@ export class PrebuildManager {
...
-            //     // we cannot start building now because the rate limiter prevents it.
-            //     span.setTag("starting", false);
-            //     return { wsid: workspace.id, done: false };;
-            // }
+            if (await this.shouldRateLimitPrebuild(span, cloneURL)) {
+                prebuild.state = "aborted";
+                prebuild.error = "Prebuild is rate limited. Please contact Gitpod if you believe this happened in error.";
+
+                await this.workspaceDB.trace({ span }).storePrebuiltWorkspace(prebuild)
+                span.setTag("starting", false);
```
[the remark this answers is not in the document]
easyCZ: Makes sense. I'm happy to change this in a follow-up PR. The existing code already uses `starting` so there may be more placed to update.

components/server/ee/src/prebuilds/prebuild-manager.ts
```diff
@@ -228,4 +238,33 @@ export class PrebuildManager {
         }
         return hostContext.contextParser;
     }
-}
\ No newline at end of file
+
+    private async shouldRateLimitPrebuild(span: opentracing.Span, cloneURL: string): Promise<boolean> {
+        const minuteAgo = secondsBefore(new Date().toISOString(), 60)
```
geropl, line 243: This magic number is burried rather deep in here. It's ok to make this a constant if:
 - we move it more to the top, maybe by making it a unit-level `const` with a self-explanatory name
 - it's easily discoverable when looking for the config field. e.g., by adding a comment to the field description

components/server/ee/src/prebuilds/prebuild-manager.ts
```diff
@@ -228,4 +238,33 @@ export class PrebuildManager {
...
+        const unabortedCount = await this.workspaceDB.trace({span}).countUnabortedPrebuildsSince(cloneURL, new Date(minuteAgo))
+        const limit = this.getPrebuildRateLimitForCloneURL(cloneURL)
```
geropl, line 245: nit:  we don't have a formatter yet, but if we were to have one we'd required `;`to be used :upside_down_face:

components/server/ee/src/prebuilds/prebuild-manager.ts
```diff
@@ -228,4 +238,33 @@ export class PrebuildManager {
...
+        if (unabortedCount >= limit) {
+            log.debug("Prebuild exceeds rate limit", { limit, unabortedPrebuildsCount: unabortedCount, cloneURL })
+            return true;
+        }
+        return false;
+    }
+
+    private getPrebuildRateLimitForCloneURL(cloneURL: string): number {
+        // First we use any explicit overrides for a given cloneURL
+        let limit = Number(this.config.prebuildLimiter[cloneURL]);
```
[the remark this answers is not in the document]
geropl: Alright, scrape the type, just found [it](https://github.com/gitpod-io/gitpod/pull/8568/files#diff-b53cd6a3e7c6f723a548f8c5560dc26b34591c201f4605cdf7c6ca1ced8e651bR158). :+1:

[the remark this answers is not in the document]
easyCZ: So the type is pushed in there, but the type system doesn't guarantee that the value that's in the config is in fact a number. For that reason, the extra guards to ensure that we only operate with numbers.

jankeromnes:
Reminder before this gets merged: Please try to squash your commits into one or a few logical commits with a descriptive commit message. 🙏

components/gitpod-db/src/workspace-db.spec.db.ts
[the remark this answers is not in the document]
easyCZ: Created https://github.com/gitpod-io/gitpod/issues/8647

components/server/ee/src/prebuilds/prebuild-manager.ts
easyCZ: Will fix. Also cutting a ticket for a formatter. I need a smart computer to do the boring stuff for me.

easyCZ: Already exists here: https://github.com/gitpod-io/gitpod/issues/8066

components/server/ee/src/prebuilds/prebuild-manager.ts
easyCZ: Updated.

easyCZ:
Squashed commits.

easyCZ:
I suppose this PR will also need documentation - around configuration of the prebuild rate limits.

components/server/ee/src/prebuilds/prebuild-manager.ts
easyCZ: Thanks. Created https://github.com/gitpod-io/gitpod/issues/8654

components/server/src/config.ts
```diff
@@ -150,6 +150,12 @@ export interface ConfigSerialized {
      */
     chargebeeProviderOptionsFile?: string;
     enablePayment?: boolean;
+
+    /**
+     * Number of prebuilds that can be started in the last 1 minute.
+     * Key '*' specifies the default rate limit for a cloneURL, unless overriden by a specific cloneURL.
+     */
+    prebuildLimiter: {[cloneURL: string]: number } & {'*': string};
```
geropl, line 158: It we change the type to be `{[cloneURL: string]: number } & {'*': number};` we don't need the `Number()` parsing anymore.

components/server/ee/src/prebuilds/prebuild-manager.ts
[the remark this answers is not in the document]
geropl: True. What I wanted to get at is, that we typically don't do this kind of low-level verification. If we were to introduce type-verification it would be a schema for the config.

If you adjust the type as suggested in the other comment, at least the compiler is happy - and that'd be fine with me as well.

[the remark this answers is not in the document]
easyCZ: Sure, happy to make that change.

components/gitpod-db/src/typeorm/workspace-db-impl.ts
```diff
@@ -646,6 +646,17 @@ export abstract class AbstractTypeORMWorkspaceDBImpl implements WorkspaceDB {
         });
     }
 
+    public async countUnabortedPrebuildsSince(cloneURL: string, date: Date): Promise<number> {
+        const abortedState: PrebuiltWorkspaceState = 'aborted';
+        const repo = await this.getPrebuiltWorkspaceRepo();
+
+        let query = repo.createQueryBuilder('pws');
+        query = query.where('pws.cloneURL = :cloneURL', { cloneURL })
```
geropl, line 654: Checking [DBPrebuiltWorkspace](https://github.com/gitpod-io/gitpod/blob/6081ee1dad74f863a45f4e3fc329a830e59ec5a5/components/gitpod-db/src/typeorm/entity/db-prebuilt-workspace.ts#L14) I noticed we don't have an index for `(cloneURL, creationTime, state)`. We should definitely add one before merging. Here's an example of how to add a migration for a secondary index: https://github.com/gitpod-io/gitpod/commit/1c4cd73f47e6c92941a0516ea3f746d5a7ef900b

easyCZ: Great spot! Will go through this shortly.

geropl:
@andrew-farries Alright, thanks a lot for tackling this! :pray: I expected I would manage to finish the review quicker... but anyway, I'm through now.

@easyCZ thx for the adjustments and squashing (we should really think about allowing for this/make it the default).

The only thing left I see here is the [migration for an index](https://github.com/gitpod-io/gitpod/pull/8568#discussion_r821536462).

components/gitpod-db/src/typeorm/workspace-db-impl.ts
easyCZ: Added the migrations.

geropl (approved):
Awesome, thanks to both of you! :pray:

Issue #8501, opened 2022-03-01
Epic: Support GitHub Enterprise


jldec:
Gitpod users would like to use their own GitHub Enterprise instance to work with Gitpod.

The experience of registering and connecting users to GHE should be at least on parity with private GitLab instances.

Support for registering a GHE instance and performing authentication is already in place. 

**Gaps** (_additional validation required_)

- [x] Registration of the GitHub App for Gitpod, for GHE instances. 
- [x] Context URL parsing
- [x] Webhooks integration for prebuilds and projects
- [x] #8429

gtsiolis:
See also https://github.com/gitpod-io/gitpod/issues/5297.

jankeromnes:
_Logging my progress. Please let me know if I should do this somewhere else. 🙏_

### Current state

- We have a GitHub Enterprise installation for testing (thanks @metcalfc!)
- There are already [docs for connecting GitHub Enterprise](https://www.gitpod.io/docs/github-integration#github-enterprise-server):
  - Seems to be written for Self-Hosted (e.g. "starting your Gitpod installation, you’ll get redirected to /first-steps")
  - However, it seems to work with gitpod.io
- I was able to connect our GitHub Enterprise instance with gitpod.io :

<img width="600" alt="Screenshot 2022-03-03 at 09 47 33" src="https://user-images.githubusercontent.com/599268/156558753-a2ca0c1e-c301-4eba-98e2-ad1539962512.png">

### Problems

- However, the final step of adding the GitHub Enterprise integration is to Connect with it, and this currently fails:

<img width="600" alt="Screenshot 2022-03-03 at 09 47 02" src="https://user-images.githubusercontent.com/599268/156559070-281c2fe3-9e48-47d1-b488-f8236a391edf.png">

Maybe it's because I've configured it badly, or there is a bug somewhere. Currently investigating this.

jankeromnes:
This logged error seems relevant:

```
TokenError: The client_id and/or client_secret passed are incorrect.
    at GenericOAuth2Strategy.OAuth2Strategy.parseErrorResponse (/app/node_modules/passport-oauth2/lib/strategy.js:373:12)
    at GenericOAuth2Strategy.OAuth2Strategy._createOAuthError (/app/node_modules/passport-oauth2/lib/strategy.js:420:16)
    at /app/node_modules/passport-oauth2/lib/strategy.js:177:45
[...]

Issue #8429, opened 2022-02-24
Gitpod self-hosted create workspace parse `.gitpod.yml` error from Github Enterprise


neilkuantm:
### Bug description

Gitpod create workspace parse `.gitpod.yml` from Github Enterprise repo got `Invalid gitpod.yml: Unparsable Gitpod configuration: YAMLException: end of the stream or a document separator is expected at line 27, column 25: ... authentication" style="word-wrap: break-word;"> ^`

![Screen Shot 2022-02-22 at 11 19 19 AM](https://user-images.githubusercontent.com/98370258/155476557-5424da84-bb57-420d-83ab-137c6fc1f4b3.png)

component `server` error message
```bash
{"component":"server","severity":"INFO","time":"2022-02-22T03:19:06.231Z","context":{"userId":"83488da8-9fd8-470b-aca9-23100ad8492a"},"message":"admin access","payload":{"authorised":true,"method":"adminGetSettings","params":{}}}
[...]

PR #8574, opened 2022-03-03
Support GitHub Enterprise

jankeromnes:
## Description

- [x] Fix raw file URL for GitHub Enterprise to allow fetching `.gitpod.yml` files successfully
- [x] Fix the `Grant Access` button/flow for private repositories on GitHub Enterprise
- [x] Allow creating Projects and running Prebuilds for GitHub Enterprise repositories
- [x] Drive-by: Trim `clientId` and `clientSecret` values when creating a new Integration
- [x] Drive-by: In dev-staging, don't truncate the callbackURL of new Integrations

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/8501 https://github.com/gitpod-io/gitpod/issues/8429

## How to test

1. Have a GitHub Enterprise server account (for example, on https://ghe.gitpod-self-hosted.com )
2. Have a public repository there (e.g. https://ghe.gitpod-self-hosted.com/roboquat/public-repo) and a private repository (e.g. https://ghe.gitpod-self-hosted.com/roboquat/top-secret)
3. Log in to https://jx-fix-ghe-yml.staging.gitpod-dev.com/integrations (e.g. using your github.com account)
4. Follow https://www.gitpod.io/docs/github-integration#github-enterprise-server to create a new Integration for your GitHub Enterprise server, and activate it
5. Open your public repository from 2. in https://jx-fix-ghe-yml.staging.gitpod-dev.com/open -- it should work
6. Open your private repository from 2. -- it should ask you to `Grant Access`, then also work
7. Add your two repositories as Projects in https://jx-fix-ghe-yml.staging.gitpod-dev.com/new
8. Once the Projects are created, a first Prebuild should run and complete successfully
9. Push a new commit to one of your repositories -- the new commit should be detected by Gitpod, and trigger a new Prebuild

## Release Notes

```release-note
Support GitHub Enterprise
```

### DIFF
--- a/components/dashboard/src/projects/NewProject.tsx
+++ b/components/dashboard/src/projects/NewProject.tsx
@@ -525,8 +525,7 @@ function GitProviders(props: {
         });
     }
 
-    // for now we exclude GitHub Enterprise
-    const filteredProviders = () => props.authProviders.filter(p => p.host === "github.com" || p.host === "bitbucket.org" || p.authProviderType === "GitLab");
+    const filteredProviders = () => props.authProviders.filter(p => p.authProviderType === "GitHub" || p.host === "bitbucket.org" || p.authProviderType === "GitLab");
 
     return (
         <div className="mt-8 border rounded-t-xl border-gray-100 dark:border-gray-800 flex-col">
--- a/components/server/ee/src/prebuilds/github-enterprise-app.ts
+++ b/components/server/ee/src/prebuilds/github-enterprise-app.ts
@@ -1,0 +1,197 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the Gitpod Enterprise Source Code License,
+ * See License.enterprise.txt in the project root folder.
+ */
+
+import * as express from 'express';
+import { createHmac } from 'crypto';
+import { postConstruct, injectable, inject } from 'inversify';
+import { ProjectDB, TeamDB, UserDB } from '@gitpod/gitpod-db/lib';
+import { PrebuildManager } from '../prebuilds/prebuild-manager';
+import { TraceContext } from '@gitpod/gitpod-protocol/lib/util/tracing';
+import { TokenService } from '../../../src/user/token-service';
+import { HostContextProvider } from '../../../src/auth/host-context-provider';
+import { log } from '@gitpod/gitpod-protocol/lib/util/logging';
+import { Project, StartPrebuildResult, User } from '@gitpod/gitpod-protocol';
+import { GitHubService } from './github-service';
+import { URL } from 'url';
+
+@injectable()
+export class GitHubEnterpriseApp {
+
+    @inject(UserDB) protected readonly userDB: UserDB;
+    @inject(PrebuildManager) protected readonly prebuildManager: PrebuildManager;
+    @inject(TokenService) protected readonly tokenService: TokenService;
+    @inject(HostContextProvider) protected readonly hostContextProvider: HostContextProvider;
+    @inject(ProjectDB) protected readonly projectDB: ProjectDB;
+    @inject(TeamDB) protected readonly teamDB: TeamDB;
+
+    protected _router = express.Router();
+    public static path = '/apps/ghe/';
+
+    @postConstruct()
+    protected init() {
+        this._router.post('/', async (req, res) => {
+            const event = req.header('X-Github-Event');
+            if (event === 'push') {
+                const payload = req.body as GitHubEnterprisePushPayload;
+                const span = TraceContext.startSpan("GitHubEnterpriseApp.handleEvent", {});
+                span.setTag("payload", payload);
+                let user: User | undefined;
+                try {
+                    user = await this.findUser({ span }, payload, req);
+                } catch (error) {
+                    log.error("Cannot find user.", error, { req })
+                }
+                if (!user) {
+                    res.statusCode = 401;
+                    res.send();
+                    return;
+                }
+                await this.handlePushHook({ span }, payload, user);
+            } else {
+                log.info("Unknown GitHub Enterprise event received", { event });
+            }
+            res.send('OK');
+        });
+    }
+
+    protected async findUser(ctx: TraceContext, payload: GitHubEnterprisePushPayload, req: express.Request): Promise<User> {
+        const span = TraceContext.startSpan("GitHubEnterpriseApp.findUser", ctx);
+        try {
+            const host = req.header('X-Github-Enterprise-Host');
+            const hostContext = this.hostContextProvider.get(host || '');
+            if (!host || !hostContext) {
+                throw new Error('Unsupported GitHub Enterprise host: ' + host);
+            }
+            const { authProviderId } = hostContext.authProvider;
+            const authId = payload.sender.id;
+            const user = await this.userDB.findUserByIdentity({ authProviderId, authId });
+            if (!user) {
+                throw new Error(`No user found with identity ${authProviderId}/${authId}.`);
+            } else if (!!user.blocked) {
+                throw new Error(`Blocked user ${user.id} tried to start prebuild.`);
+            }
+            const gitpodIdentity = user.identities.find(i => i.authProviderId === TokenService.GITPOD_AUTH_PROVIDER_ID);
+            if (!gitpodIdentity) {
+                throw new Error(`User ${user.id} has no identity for '${TokenService.GITPOD_AUTH_PROVIDER_ID}'.`);
+            }
+            // Verify the webhook signature
+            const signature = req.header('X-Hub-Signature-256');
+            const body = (req as any).rawBody;
+            const tokenEntries = (await this.userDB.findTokensForIdentity(gitpodIdentity)).filter(tokenEntry => {
+                return tokenEntry.token.scopes.includes(GitHubService.PREBUILD_TOKEN_SCOPE);
+            });
+            const signingToken = tokenEntries.find(tokenEntry => {
+                const sig = 'sha256=' + createHmac('sha256', user.id + '|' + tokenEntry.token.value)
+                    .update(body)
+                    .digest('hex');
+                return sig === signature;
+            });
+            if (!signingToken) {
+                throw new Error(`User ${user.id} has no token matching the payload signature.`);
+            }
+            return user;
+        } finally {
+            span.finish();
+        }
+    }
+
+    protected async handlePushHook(ctx: TraceContext, payload: GitHubEnterprisePushPayload, user: User): Promise<StartPrebuildResult | undefined> {
+        const span = TraceContext.startSpan("GitHubEnterpriseApp.handlePushHook", ctx);
+        try {
+            const contextURL = this.createContextUrl(payload);
+            span.setTag('contextURL', contextURL);
+            const config = await this.prebuildManager.fetchConfig({ span }, user, contextURL);
+            if (!this.prebuildManager.shouldPrebuild(config)) {
+                log.info('GitHub Enterprise push event: No config. No prebuild.');
+                return undefined;
+            }
+
+            log.debug('GitHub Enterprise push event: Starting prebuild.', { contextURL });
+
+            const cloneURL = payload.repository.clone_url;
+            const projectAndOwner = await this.findProjectAndOwner(cloneURL, user);
+
+            const ws = await this.prebuildManager.startPrebuild({ span }, {
+                user: projectAndOwner.user,
+                project: projectAndOwner?.project,
+                branch: this.getBranchFromRef(payload.ref),
+                contextURL,
+                cloneURL,
+                commit: payload.after,
+            });
+            return ws;
+        } finally {
+            span.finish();
+        }
+    }
+
+    /**
+     * Finds the relevant user account and project to the provided webhook event information.
+     *
+     * First of all it tries to find the project for the given `cloneURL`, then it tries to
+     * find the installer, which is also supposed to be a team member. As a fallback, it
+     * looks for a team member which also has a connection with this GitHub Enterprise server.
+     *
+     * @param cloneURL of the webhook event
+     * @param webhookInstaller the user account known from the webhook installation
+     * @returns a promise which resolves to a user account and an optional project.
+     */
+    protected async findProjectAndOwner(cloneURL: string, webhookInstaller: User): Promise<{ user: User, project?: Project }> {
+        const project = await this.projectDB.findProjectByCloneUrl(cloneURL);
+        if (project) {
+            if (project.userId) {
+                const user = await this.userDB.findUserById(project.userId);
+                if (user) {
+                    return { user, project };
+                }
+            } else if (project.teamId) {
+                const teamMembers = await this.teamDB.findMembersByTeam(project.teamId || '');
+                if (teamMembers.some(t => t.userId === webhookInstaller.id)) {
+                    return { user: webhookInstaller, project };
+                }
+                const hostContext = this.hostContextProvider.get(new URL(cloneURL).host);
+                const authProviderId = hostContext?.authProvider.authProviderId;
+                for (const teamMember of teamMembers) {
+                    const user = await this.userDB.findUserById(teamMember.userId);
+                    if (user && user.identities.some(i => i.authProviderId === authProviderId)) {
+                        return { user, project };
+                    }
+                }
+            }
+        }
+        return { user: webhookInstaller };
+    }
+
+    protected getBranchFromRef(ref: string): string | undefined {
+        const headsPrefix = "refs/heads/";
+        if (ref.startsWith(headsPrefix)) {
+            return ref.substring(headsPrefix.length);
+        }
+
+        return undefined;
+    }
+
+    protected createContextUrl(payload: GitHubEnterprisePushPayload) {
+        return `${payload.repository.url}/tree/${this.getBranchFromRef(payload.ref)}`;
+    }
+
+    get router(): express.Router {
+        return this._router;
+    }
+}
+
+interface GitHubEnterprisePushPayload {
+    ref: string;
+    after: string;
+    repository: {
+        url: string;
+        clone_url: string;
+    };
+    sender: {
+        login: string;
+        id: string;
+    };
+}
--- a/components/server/ee/src/prebuilds/github-service.ts
+++ b/components/server/ee/src/prebuilds/github-service.ts
@@ -1,0 +1,87 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the Gitpod Enterprise Source Code License,
+ * See License.enterprise.txt in the project root folder.
+ */
+
+import { RepositoryService } from "../../../src/repohost/repo-service";
+import { inject, injectable } from "inversify";
+import { GitHubGraphQlEndpoint, GitHubRestApi } from "../../../src/github/api";
+import { GitHubEnterpriseApp } from "./github-enterprise-app";
+import { AuthProviderParams } from "../../../src/auth/auth-provider";
+import { GithubContextParser } from "../../../src/github/github-context-parser";
+import { ProviderRepository, User } from "@gitpod/gitpod-protocol";
+import { Config } from "../../../src/config";
+import { TokenService } from "../../../src/user/token-service";
+
+@injectable()
+export class GitHubService extends RepositoryService {
+
+    static PREBUILD_TOKEN_SCOPE = 'prebuilds';
+
+    @inject(GitHubGraphQlEndpoint) protected readonly githubQueryApi: GitHubGraphQlEndpoint;
+    @inject(GitHubRestApi) protected readonly githubApi: GitHubRestApi;
+    @inject(Config) protected readonly config: Config;
+    @inject(AuthProviderParams) protected authProviderConfig: AuthProviderParams;
+    @inject(TokenService) protected tokenService: TokenService;
+    @inject(GithubContextParser) protected githubContextParser: GithubContextParser;
+
+    async getRepositoriesForAutomatedPrebuilds(user: User): Promise<ProviderRepository[]> {
+        const repositories = (await this.githubApi.run(user, gh => gh.repos.listForAuthenticatedUser({}))).data;
+        const adminRepositories = repositories.filter(r => !!r.permissions?.admin)
+        return adminRepositories.map(r => {
+            return <ProviderRepository>{
+                name: r.name,
+                cloneUrl: r.clone_url,
+                account: r.owner?.login,
+                accountAvatarUrl: r.owner?.avatar_url,
+                updatedAt: r.updated_at
+            };
+        });
+    }
+
+    async canInstallAutomatedPrebuilds(user: User, cloneUrl: string): Promise<boolean> {
+        const { host, owner, repoName: repo } = await this.githubContextParser.parseURL(user, cloneUrl);
+        if (host !== this.authProviderConfig.host) {
+            return false;
+        }
+        try {
+            // You need "ADMIN" permission on a repository to be able to install a webhook.
+            // Ref: https://docs.github.com/en/organizations/managing-access-to-your-organizations-repositories/repository-roles-for-an-organization#permissions-for-each-role
+            // Ref: https://docs.github.com/en/graphql/reference/enums#repositorypermission
+            const result: any = await this.githubQueryApi.runQuery(user, `
+                query {
+                    repository(name: "${repo}", owner: "${owner}") {
+                        viewerPermission
+                    }
+                }
+            `);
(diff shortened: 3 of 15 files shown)


jankeromnes:
Taking this back to draft: Should work fine as is, but needs a GitHub Enterprise integration with the preview environment to test it, and once I have that, I might as well add in a few more fixes.

components/server/ee/src/prebuilds/github-service.ts
```diff
@@ -0,0 +1,52 @@
...
+export class GitHubService extends RepositoryService {
+
+    static PREBUILD_TOKEN_SCOPE = 'prebuilds';
+
+    @inject(GitHubGraphQlEndpoint) protected readonly githubQueryApi: GitHubGraphQlEndpoint;
+    @inject(GitHubRestApi) protected readonly githubApi: GitHubRestApi;
+    @inject(Config) protected readonly config: Config;
+    @inject(AuthProviderParams) protected authProviderConfig: AuthProviderParams;
+	@inject(TokenService) protected tokenService: TokenService;
+	@inject(GithubContextParser) protected githubContextParser: GithubContextParser;
```
andrew-farries, line 27: Looks like some wonky formatting here.

components/server/src/bitbucket-server/bitbucket-server-repository-provider.ts
```diff
@@ -102,6 +102,11 @@ export class BitbucketServerRepositoryProvider implements RepositoryProvider {
         return [];
     }
 
+    async getProviderRepos(user: User): Promise<ProviderRepository[]> {
+        // FIXME(janx): Not implemented yet
+        return [];
+    }
+
```
andrew-farries, line 109: Assume this FIXME is not intended to be addressed in this PR?

components/server/src/bitbucket/bitbucket-repository-provider.ts
```diff
@@ -104,6 +104,11 @@ export class BitbucketRepositoryProvider implements RepositoryProvider {
         return [];
     }
 
+    async getProviderRepos(user: User): Promise<ProviderRepository[]> {
+        // FIXME(janx): Not implemented yet
+        return [];
+    }
+
```
andrew-farries, line 111: Same question about the FIXME.

components/server/src/gitlab/gitlab-repository-provider.ts
```diff
@@ -100,6 +100,11 @@ export class GitlabRepositoryProvider implements RepositoryProvider {
         return [];
     }
 
+    async getProviderRepos(user: User): Promise<ProviderRepository[]> {
+        // FIXME(janx): Not implemented yet
+        return [];
+    }
+
```
andrew-farries, line 107: Same question about the FIXME

andrew-farries (commented):
We went through some of this on the call earlier, but could the 'how to test' section be filled in so that I can test an integration with GHE using our installation at https://ghe.gitpod-self-hosted.com/ ?

components/server/src/bitbucket-server/bitbucket-server-repository-provider.ts
jankeromnes: You're right. 👍

As discussed earlier, I'd love to re-unify these "repository-handling" classes soon:
- [RepositoryProvider](https://github.com/gitpod-io/gitpod/blob/main/components/server/src/repohost/repository-provider.ts)
- [RepositoryService](https://github.com/gitpod-io/gitpod/blob/main/components/server/src/repohost/repo-service.ts)
- {GitHub,GitLab,Bitbucket}AppSupport (single method, no parent class, see e.g. [GitLabAppSupport](https://github.com/gitpod-io/gitpod/blob/main/components/server/ee/src/gitlab/gitlab-app-support.ts))

An in fact, this method `getProviderRepos` is actually already implemented in [BitbucketAppSupport](https://github.com/gitpod-io/gitpod/blob/main/components/server/ee/src/bitbucket/bitbucket-app-support.ts#L18-L78).

But I'd prefer leaving this refactoring for a follow-up Pull Request, so adding a FIXME for now.

andrew-farries:
Thanks for adding the testing steps. 

When I push a new commit to https://ghe.gitpod-self-hosted.com/roboquat/public-repo, I see the commit in the `Branches` tab but I don't get a corresponding prebuild on the `Prebuilds` tab:

https://user-images.githubusercontent.com/8225907/157419757-810d4e12-ffbf-460d-b38c-dc646cc1e7b6.mov

jankeromnes:
Awesome, many thanks for testing @andrew-farries! 🙏

Indeed, the triggering of new Prebuilds on pushes is currently broken, as I'm still adjusting the webhook signature verification. I'm currently iterating on it like so:
- pushing a new commit to the test repo
- observing the server logs for any errors
- fixing errors, then waiting for a new deployment of this PR

Happy to have a chat about where things are. 🙂 I'm hopeful that it will be working and ready for review quite soon 🤞 (at which point we'll be able to take the PR out of Draft).

jankeromnes:
Thanks @andrew-farries for helping discuss, review, and test this. 💯

It now works as expected, and should be about ready for review. 🚀

jankeromnes:
Holding because of the minor TODOs above:

/hold

jankeromnes (commented):
(Forgot to actually submit the minor TODO comments)

components/server/ee/src/prebuilds/github-enterprise-app.ts
```diff
@@ -0,0 +1,195 @@
...
+            span.finish();
+        }
+    }
+
+    /**
+     * Finds the relevant user account and project to the provided webhook event information.
+     *
+     * First of all it tries to find the project for the given `cloneURL`, then it tries to
+     * find the installer, which is also supposed to be a team member. As a fallback, it
+     * looks for a team member which also has a bitbucket.org connection.
```
geropl, line 136: > bitbucket.org

Copy'n'paste

geropl:
@jankeromnes Thank you for tackling this, and providing a working solution this fast! :100: :pray: 

We discussed this earlier, just want to point it out again:
 - This PR breaks with the existing pattern we use for integrations `type == 'GitHub'`, where we install GitHub Apps on the org/repo level. Instead, it uses the OAuth-app (user-impersonation) to install the webhook (see ["Differences between GitHub Apps and OAuth Apps"](https://docs.github.com/en/developers/apps/getting-started-with-apps/differences-between-github-apps-and-oauth-apps)).
 - I'm fine with deviating for GitHub Enterprise, especially if it's temporary for the sake of progress. But then someone must own the responsibility to make this coherent again a) technically (by implementing a any required changes), and b) product-wise (by re-discovering the differences between both approaches). (/cc @jldec )
 - I want to
   a) ensure we don't take up additional long-term technical debt by introducing new integration modes
   b) make sure someone (not me :upside_down_face: ) follows up on this, because I see no value being added to our goals in this move.

If I'm missing something here - be it technical, or other - please let me know! Also, really don't want to hinder the good progress here, just want to make sure by point is clear. :slightly_smiling_face:

geropl:
On the technical side: How do we ensure we have the right scopes for:
 - registering webhooks (`write:repo_hook` or `write:org_hook` scope, [ref](https://docs.github.com/en/developers/apps/getting-started-with-apps/differences-between-github-apps-and-oauth-apps#requesting-permission-levels-for-resources))
 - updating Commit statuses (`repo:status` scope, [ref](https://docs.github.com/en/developers/apps/getting-started-with-apps/differences-between-github-apps-and-oauth-apps#requesting-permission-levels-for-resources))

jankeromnes:
Thanks @geropl for the quick review! 🙏

>  - This PR breaks with the existing pattern we use for integrations `type == 'GitHub'`, where we install GitHub Apps on the org/repo level. Instead, it uses the OAuth-app (user-impersonation) to install the webhook (see ["Differences between GitHub Apps and OAuth Apps"](https://docs.github.com/en/developers/apps/getting-started-with-apps/differences-between-github-apps-and-oauth-apps)).

I don't think it "breaks the pattern", because the GitHub App is [a singleton](https://github.com/gitpod-io/gitpod/blob/54a765d321236ef42dceeb154cf477b17a8aaf32/components/server/src/config.ts#L72-L80) in Gitpod (i.e. you can only have one GitHub App per Gitpod installation), while we already allow an arbitrary number of Integrations (of type `GitHub` or `GitLab`) which are already OAuth apps. As we've seen, there are already 13 GitHub Enterprise integrations on gitpod.io. This Pull Request makes them actually work.

Also, webhooks and the GitHub App are not mutually exclusive. As a next step to this PR, I'd love to explore what it could look like if you have a single Gitpod Self-Hosted installation, and connect its GitHub App singleton to a single GitHub Enterprise server (as opposed to connecting it to github.com as we do on SaaS). Ironing out any bugs in that process could create a nice alternative integration point specifically for Self-Hosted / GHE.

> - I want to
>   a) ensure we don't take up additional long-term technical debt by introducing new integration modes

- We already allow arbitrary OAuth App integrations for GitLab and GitHub Enterprise in Gitpod (that's how the custom Integrations in `/integrations` work)
- This Pull Request just makes the existing GitHub Enterprise integrations work properly, by using the same mechanism as GitLab and Bitbucket (webhooks)
- In the future, I would love to improve these integrations (GitLab, Bitbucket and GitHub Enterprise) to make them post visible status updates on repositories, commits and Pull Requests, to make GitLab Bitbucket and GitHub Enterprise reach parity with our current `github.com` support

>   b) make sure someone (not me 🙃 ) follows up on this, because I see no value being added to our goals in this move.

As mentioned above, I'm happy to follow up, e.g. by testing a Gitpod Self-Hosted using a GitHub App connected to a GitHub Enterprise server. As mentioned above, I don't think webhooks and the GitHub App are mutually exclusive.

jankeromnes:
> On the technical side: How do we ensure we have the right scopes for:
>  - registering webhooks (`write:repo_hook` or `write:org_hook` scope, [ref](https://docs.github.com/en/developers/apps/getting-started-with-apps/differences-between-github-apps-and-oauth-apps#requesting-permission-levels-for-resources))
>  - updating Commit statuses (`repo:status` scope, [ref](https://docs.github.com/en/developers/apps/getting-started-with-apps/differences-between-github-apps-and-oauth-apps#requesting-permission-levels-for-resources))

I believe these scopes are all included in the `repo` (or `public_repo`) scopes we already request (see mentions of "webhooks" and "commit status" in [available scopes](https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps#available-scopes))

jankeromnes:
For some reason Werft decided to hide the preview URL of this Pull Request, but the preview itself is still live FYI:

https://jx-fix-ghe-yml.staging.gitpod-dev.com/

components/server/src/bitbucket-server/bitbucket-server-repository-provider.ts
jankeromnes: Ah, given that the EE vs not-EE split might still be relevant (due to licensing concerns), maybe it does make sense to keep two seperate classes (for example, unify the code into `RepositoryProvider` in not-EE and `RepositoryService` in EE -- in that case I think we'd need to move this `getProviderRepos` method into `RepositoryService`, which sort of makes sense because "get webhook-compatible repos" and "is repo webhook-compatible" and "install webhook on repo" are all related)

components/server/ee/src/prebuilds/github-service.ts
```diff
@@ -0,0 +1,60 @@
...
+        const { host } = await this.githubContextParser.parseURL(user, cloneUrl);
+        if (host === 'github.com' || host !== this.authProviderConfig.host) {
+            return false;
+        }
+        return true;
+    }
+
+    async installAutomatedPrebuilds(user: User, cloneUrl: string): Promise<void> {
+        const { owner, repoName: repo } = await this.githubContextParser.parseURL(user, cloneUrl);
+        const webhooks = (await this.githubApi.run(user, gh => gh.repos.listWebhooks({ owner, repo }))).data;
```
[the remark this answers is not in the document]
AlexTugarev: as long as they are bound in the container module which are loaded per host (see HostContextProvider) this should be fine, at least that exactly the purpose of host contexts.

components/server/ee/src/prebuilds/github-service.ts
```diff
@@ -0,0 +1,60 @@
...
+    @inject(AuthProviderParams) protected authProviderConfig: AuthProviderParams;
+    @inject(TokenService) protected tokenService: TokenService;
+    @inject(GithubContextParser) protected githubContextParser: GithubContextParser;
+
+    async canInstallAutomatedPrebuilds(user: User, cloneUrl: string): Promise<boolean> {
+        const { host } = await this.githubContextParser.parseURL(user, cloneUrl);
+        if (host === 'github.com' || host !== this.authProviderConfig.host) {
+            return false;
+        }
+        return true;
```
AlexTugarev, line 34: that's not enough. see `GitLabService.canInstallAutomatedPrebuilds`, we need to check if `user` has admin/maintainer permissions for this repo.

components/server/ee/src/prebuilds/github-enterprise-app.ts
```diff
@@ -0,0 +1,197 @@
...
+                throw new Error(`Blocked user ${user.id} tried to start prebuild.`);
+            }
+            const gitpodIdentity = user.identities.find(i => i.authProviderId === TokenService.GITPOD_AUTH_PROVIDER_ID);
+            if (!gitpodIdentity) {
+                throw new Error(`User ${user.id} has no identity for '${TokenService.GITPOD_AUTH_PROVIDER_ID}'.`);
+            }
+            // Verify the webhook signature
+            const signature = req.header('X-Hub-Signature-256');
+            const body = (req as any).rawBody;
+            const tokenEntries = (await this.userDB.findTokensForIdentity(gitpodIdentity)).filter(tokenEntry => {
```
AlexTugarev, line 83: how many `tokenEntry`s are expected to be there? it would be great to pick the current one if there are multiple.

components/dashboard/src/projects/NewProject.tsx
```diff
@@ -525,17 +525,14 @@ function GitProviders(props: {
...
-    const filteredProviders = () => props.authProviders.filter(p => p.host === "github.com" || p.host === "bitbucket.org" || p.authProviderType === "GitLab");
-
     return (
         <div className="mt-8 border rounded-t-xl border-gray-100 dark:border-gray-800 flex-col">
             <div className="p-6 p-b-0">
                 <div className="text-center text-gray-500">
                     Select a Git provider first and continue with your repositories.
                 </div>
                 <div className="mt-6 flex flex-col space-y-3 items-center pb-8">
-                    {filteredProviders().map(ap => {
```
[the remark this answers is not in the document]
jankeromnes: Good catch!

I've removed this filter from the UI, because I consider that:
- Your unsupported provider is visible in the UI, but it might potentially fail somewhere down the line because we haven't fully tested it

is a better UX than:
- Your unsupported provider is not visible in the UI, because we haven't tested it yet

In other words, maybe it's better to try and give users more agency here. If things work out of the box, great! If they break in some specific way, we can file an issue and try to solve it quickly. (But if we completely hide these, we're sort of ensuring that they'll never work.)

components/server/src/server.ts
```diff
@@ -117,9 +117,9 @@ export class Server<C extends GitpodClient, S extends GitpodServer> {
         });
 
         // Express configuration
-        // Read bodies as JSON
-        app.use(bodyParser.json())
-        app.use(bodyParser.urlencoded({ extended: true }))
+        // Read bodies as JSON (but keep the raw body just in case)
+        app.use(bodyParser.json({ verify: (req, res, buffer) => { (req as any).rawBody = buffer; }}));
```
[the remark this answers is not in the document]
jankeromnes: You're right, with this change we'll carry the raw body of JSON requests we receive. I'm assuming that, apart from some webhooks, we don't receive too many / very large JSON payloads.

components/server/ee/src/prebuilds/github-service.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    async installAutomatedPrebuilds(user: User, cloneUrl: string): Promise<void> {
+        const { owner, repoName: repo } = await this.githubContextParser.parseURL(user, cloneUrl);
+        const webhooks = (await this.githubApi.run(user, gh => gh.repos.listWebhooks({ owner, repo }))).data;
+        for (const webhook of webhooks) {
+            if (webhook.config.url === this.getHookUrl()) {
+                await this.githubApi.run(user, gh => gh.repos.deleteWebhook({ owner, repo, hook_id: webhook.id }));
+            }
+        }
+        // TODO(janx): Also delete old tokens with scopes `GitHubService.PREBUILD_TOKEN_SCOPE` and `cloneUrl`?
```
AlexTugarev, line 45: please, yes! unless it's already done in `createGitpodToken` op.

AlexTugarev: isn't reusing a token another option here?
just in case, I'm thinking of the `db-sync` dilemma.

components/server/src/github/api.ts
```diff
@@ -174,7 +174,7 @@ export class GitHubRestApi {
         try {
             const response = (await operation(userApi));
             const statusCode = response.status;
-            if (statusCode !== 200) {
+            if (!(statusCode >= 200 && statusCode < 300)) {
```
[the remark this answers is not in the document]
jankeromnes: There are actually a bunch of different HTTP success codes: [Wikipedia](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#2xx_success)

And indeed, when you create a webhook using the GitHub API, it responds with `201 Created`, which is obviously not an error. 😇

components/server/ee/src/prebuilds/github-enterprise-app.ts
jankeromnes: As many `tokenEntry`s as repositories you've installed prebuilds on.

I didn't filter on the `cloneUrl` scope we also add to tokens, because we don't update the `cloneUrl`s in tokens when repositories get renamed.

components/server/ee/src/prebuilds/github-service.ts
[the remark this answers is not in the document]
jankeromnes: Awesome, many thanks for this explanation @AlexTugarev. 🙏

components/server/src/github/github-repository-provider.ts
```diff
@@ -129,6 +129,20 @@ export class GithubRepositoryProvider implements RepositoryProvider {
         return (result.data.viewer?.repositoriesContributedTo?.edges || []).map((edge: any) => edge.node.url)
     }
 
+    async getProviderRepos(user: User): Promise<ProviderRepository[]> {
```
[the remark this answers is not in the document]
jankeromnes: Refactored `RepositoryProvider.getProviderRepos` to `RepositoryService.getRepositoriesForAutomatedPrebuilds`.

Resolving here, please feel free to re-open a conversation if the new name still isn't right. 😁

components/server/ee/src/prebuilds/github-service.ts
jankeromnes: Who put _tabs_ into my soup? 😳 Fixed ✅

components/server/src/bitbucket-server/bitbucket-server-repository-provider.ts
jankeromnes: Done.

components/dashboard/src/projects/NewProject.tsx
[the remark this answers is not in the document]
AlexTugarev: > is a better UX than:

really? 

how is a (untested ===) broken state better than hiding untested paths? 

> But if we completely hide these, we're sort of ensuring that they'll never work.

We got a lot of requests and scheduled several tasks to extend the support in this area in the past year, and this PR is an example for it. So this isn't true.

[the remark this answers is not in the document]
jankeromnes: I'm still biased towards leaving all doors open and all panels unlocked, so that savvy users can help us test and find their way around / fix their own problems, as opposed to shutting every door and having just one supported path.

However, I also understand your point of view, so I'll re-instate the UI filter I removed (and adjust it to show GitHub Enterprise as well).

components/server/ee/src/prebuilds/github-enterprise-app.ts
AlexTugarev: oh! that sounds really expensive. if we get many of them, such operations will drive the event loop lag even higher.

components/server/ee/src/prebuilds/github-service.ts
jankeromnes: > unless it's already done in `createGitpodToken` op.

Ah, good thought! How nice that this is actually the case: 😄

https://github.com/gitpod-io/gitpod/blob/3bd683e6eaf391deeb3b3b7803d7fe4d4e513d9d/components/server/src/user/token-service.ts#L64-L67

> isn't reusing a token another option here?
> just in case, I'm thinking of the `db-sync` dilemma.

Could you please explain the `db-sync` dilemma regarding whether to reuse a token vs creating a new one? 🤔

FYI, I followed the same pattern as GitLab and Bitbucket webhooks, i.e. generate a new token for every repo (with the scopes `"prebuilds"` and `cloneUrl`).

geropl:
@jankeromnes Thanks for the [explanation](https://github.com/gitpod-io/gitpod/pull/8574#issuecomment-1062938200).

> As we've seen, there are already 13 GitHub Enterprise integrations on gitpod.io. This Pull Request makes them actually work.

:bulb: It's news to me that we allow/already support these. I saw you slack post earlier this week, but thought the entries where due to bugs.

> I don't think it "breaks the pattern"

It's bit complicated because of the two use cases:
 - qualified, by admin, at config-time: connect 1 integration with a single, statically configured GitHub app
 - unqualified, by user, at runtime: connect n integrations; so far we only had GitLab/BitBucket, so used webhooks (and now for GHE as well)

My main concern was that I understood that we want to achieve 1.), but this PR uses the techniques we employ for 2.). This is what I meant with "breaking the pattern". Instead of focusing on the problem at hand (having a customer configure they own GitHub App for their own installation) we seem to be fixing the GHE integration in general. Which is also an interesting goal, but feels a bit out of scope here.

> Also, webhooks and the GitHub App are not mutually exclusive

Not at runtime, yes, but I'd really like us to see it this way, because it additional maintenance burden, and now we have to schedule the task of cleaning it up (e.g., deciding for one way or the other).

> As a next step to this PR, I'd love to explore what it could look like if you have a single Gitpod Self-Hosted installation, and connect its GitHub App singleton to a single GitHub Enterprise server (as opposed to connecting it to github.com as we do on SaaS)

Awesome, really looking forward to this. :muscle:

components/server/ee/src/prebuilds/github-service.ts
jankeromnes: Added a GraphQL query to check that `viewerPermission === 'ADMIN'` (`'MAINTAIN'` is not sufficient -- [source](https://docs.github.com/en/organizations/managing-access-to-your-organizations-repositories/repository-roles-for-an-organization#permissions-for-each-role))

jankeromnes:
Many thanks @andrew-farries, @geropl and @AlexTugarev for your super helpful reviews! 🙏

I've addressed all nits, so this seems ready to be merged. 🚢

/unhold

components/server/ee/src/prebuilds/github-enterprise-app.ts
```diff
@@ -0,0 +1,197 @@
...
+                return tokenEntry.token.scopes.includes(GitHubService.PREBUILD_TOKEN_SCOPE);
+            });
+            const signingToken = tokenEntries.find(tokenEntry => {
+                const sig = 'sha256=' + createHmac('sha256', user.id + '|' + tokenEntry.token.value)
+                    .update(body)
+                    .digest('hex');
+                return sig === signature;
```
andrew-farries, line 90: This is probably pedantry at this stage, but comparing the HMAC signatures using a simple string comparison is considered insecure because of the potential for timing attacks. 

See [the github webhook docs](https://docs.github.com/en/developers/webhooks-and-events/webhooks/securing-your-webhooks). Seems the best way to do it is with `crypto.timingSafeEqual` as suggested [here](https://stackoverflow.com/questions/51486432/what-is-the-preferred-way-of-comparing-hmac-signatures-in-node).

andrew-farries (approved):

components/server/ee/src/prebuilds/github-service.ts
AlexTugarev: > Could you please explain the db-sync dilemma regarding whether to reuse a token vs creating a new one? 🤔

If you override tokens in the DB for a particular repo, they won't be available in the other cluster until the following db-sync cycle. But that might be a general problem with self-managed webhooks.

PR #8576, opened 2022-03-03
[installer] Allow to set default workspace timeout

### LINKED ISSUE #8528: Disable Workspace Timeout When Running in a Self-Hosted Manner
I'm having the issue of workspaces timing out and thus shutting down even though I still want to use it.

Interactive way to change the timeout of a workspace to allow it to run continuously even after a script has finished and/or there is no user input/activity.

Put a sleep at the end of the script, which is more of a hack

## Additional context
Connecting to the workspace via SSH & using self-hosted Gitpod

corneliusludmann:
## Description

This is a proposal that allows to set custom timeouts for the workspaces. It adds the following config values:
```yaml
workspace:
  timeoutDefault: 30m
  timeoutExtended: 60m
  maxLifetime: 36h
  timeoutAfterClose: 2m
```

## Related Issue(s)

Fixes #8528

## How to test

Add the following patch to `deploy-to-preview-environment.ts`:
```diff
diff --git a/.werft/jobs/build/deploy-to-preview-environment.ts b/.werft/jobs/build/deploy-to-preview-environment.ts
index 4c82fa8d..b963703a 100644
--- a/.werft/jobs/build/deploy-to-preview-environment.ts
+++ b/.werft/jobs/build/deploy-to-preview-environment.ts
@@ -392,6 +392,12 @@ async function deployToDevWithInstaller(werft: Werft, jobConfig: JobConfig, depl
             exec('exit 0')
         }
 
+        // TODO: remove me – just for testing purposes
+        exec(`yq w -i ./config.yaml workspace.timeoutDefault "3m"`);
+        exec(`yq w -i ./config.yaml workspace.timeoutExtended "6m"`);
+        exec(`yq w -i ./config.yaml workspace.maxLifetime "9m"`);
+        exec(`yq w -i ./config.yaml workspace.timeoutAfterClose "10s"`);
+
         // validate the config
         exec(`/tmp/installer validate config -c config.yaml`, { slice: installerSlices.INSTALLER_RENDER });
```

And run a werft build like this:
```
$ werft run github -s .werft/jobs/build/deploy-to-preview-environment.ts
```
**NOTE:** I did this already. As long as the preview env hasn't been removed in the meantime, the preview should be deployed like this.

With this configuration you can observe the following timeouts:
- When you start a workspace, it times out after 3 minutes of inactivity. You can see this value when you run (change your workspace ID):
   ```
   $ kubectl describe pod ws-50950ba6-eea9-4d06-8aa9-a98f6a19811d  | grep customTimeout
                 gitpod/customTimeout: 3m0s
   ```
  You'll also see the current timeout when you hover over the clock in your workspace status bar:
image

- When you extend your workspace timeout (click on the clock), the new timeout will be 6 minutes (check again with `kubectl` or the hover message).
- When you close the browser, the workspace times out after 10 seconds. (Note: The value that you see with `kubectl` will not change but you can see that the pod disappears after 10 seconds and the Gitpod dashboard shows stopping.)
- The workspace will be stopped after 9 minutes. Even if there is activity. It will stop in every case.

## Discussion and Limitations

- I tried to eliminate the hard-coded values in the code and add “classes” of timeouts (short, long, and extended). However, [the VS Code IDE sends always `180m` as a timeout value](https://github.com/gitpod-io/openvscode-server/blob/5658446f7bef3662e805df93669b0b4cb79349c5/extensions/gitpod-shared/src/features.ts#L598) when a user clicks on the clock. That's why I kept the `180m` constant as an alternative value for `extended`. In the long term, it would make sense to send `extended` instead.

- When the user extends the workspace, the message always says [it's extended to “three hours”](https://github.com/gitpod-io/openvscode-server/blob/5658446f7bef3662e805df93669b0b4cb79349c5/extensions/gitpod-shared/src/features.ts#L599-L602). That should be changed to a more general wording.

- The hover text always shows the proper timeout value but the color of the clock is checked by comparing it [with the hard-coded value of `180m`](https://github.com/gitpod-io/openvscode-server/blob/5658446f7bef3662e805df93669b0b4cb79349c5/extensions/gitpod-shared/src/features.ts#L627). The problem is, that we don't know at this point what the extended timeout value is. Instead of [reading `listener.info.latestInstance`](https://github.com/gitpod-io/openvscode-server/blob/5658446f7bef3662e805df93669b0b4cb79349c5/extensions/gitpod-shared/src/features.ts#L621), we could also call 
   ```
  const workspaceTimeout = await context.gitpod.server.getWorkspaceTimeout(context.info.getWorkspaceId());
  ```
   instead. That [always returns `180m` as timeout as constant for the extended value](https://github.com/gitpod-io/gitpod/blob/9e736913586482b75d43cc68454304e1e7ac477a/components/server/src/user/user-service.ts#L167-L169) (should be probably changed to `extended` in the long term, see above). However, I don't know if that has a performance impact.

All of these are little UI flaws that shouldn't be real blockers, in my opinion.

/cc @akosyakov 

## Release Notes

```release-note
Allow setting custom workspace timeouts in the Gitpod Installer for self-hosted installations.
```

### DIFF
--- a/components/server/ee/src/user/user-service.ts
+++ b/components/server/ee/src/user/user-service.ts
@@ -5,7 +5,7 @@
  */
 
 import { UserService, CheckSignUpParams, CheckTermsParams } from "../../../src/user/user-service";
-import { User, WorkspaceTimeoutDuration } from "@gitpod/gitpod-protocol";
+import { User, WorkspaceTimeoutDuration, WORKSPACE_TIMEOUT_EXTENDED, WORKSPACE_TIMEOUT_EXTENDED_ALT, WORKSPACE_TIMEOUT_DEFAULT_LONG, WORKSPACE_TIMEOUT_DEFAULT_SHORT } from "@gitpod/gitpod-protocol";
 import { inject } from "inversify";
 import { LicenseEvaluator } from "@gitpod/licensor/lib";
 import { Feature } from "@gitpod/licensor/lib/api";
@@ -14,6 +14,7 @@ import { EligibilityService } from "./eligibility-service";
 import { SubscriptionService } from "@gitpod/gitpod-payment-endpoint/lib/accounting";
 import { OssAllowListDB } from "@gitpod/gitpod-db/lib/oss-allowlist-db";
 import { HostContextProvider } from "../../../src/auth/host-context-provider";
+import { Config } from "../../../src/config";
 
 export class UserServiceEE extends UserService {
     @inject(LicenseEvaluator) protected readonly licenseEvaluator: LicenseEvaluator;
@@ -21,6 +22,7 @@ export class UserServiceEE extends UserService {
     @inject(SubscriptionService) protected readonly subscriptionService: SubscriptionService;
     @inject(OssAllowListDB) protected readonly OssAllowListDb: OssAllowListDB;
     @inject(HostContextProvider) protected readonly hostContextProvider: HostContextProvider;
+    @inject(Config) protected readonly config: Config;
 
     async getDefaultWorkspaceTimeout(user: User, date: Date): Promise<WorkspaceTimeoutDuration> {
         if (this.config.enablePayment) {
@@ -32,10 +34,35 @@ export class UserServiceEE extends UserService {
 
         // the self-hosted case
         if (!this.licenseEvaluator.isEnabled(Feature.FeatureSetTimeout, userCount)) {
-            return "30m";
+            return WORKSPACE_TIMEOUT_DEFAULT_SHORT;
         }
 
-        return "60m";
+        return WORKSPACE_TIMEOUT_DEFAULT_LONG;
+    }
+
+    public workspaceTimeoutToDuration(timeout: WorkspaceTimeoutDuration): string {
+        switch (timeout) {
+            case WORKSPACE_TIMEOUT_DEFAULT_SHORT:
+                return "30m";
+            case WORKSPACE_TIMEOUT_DEFAULT_LONG:
+                return this.config.workspaceDefaults.timeoutDefault || "60m";
+            case WORKSPACE_TIMEOUT_EXTENDED:
+            case WORKSPACE_TIMEOUT_EXTENDED_ALT:
+                return this.config.workspaceDefaults.timeoutExtended || "180m";
+        }
+    }
+
+    public durationToWorkspaceTimeout(duration: string): WorkspaceTimeoutDuration {
+        switch (duration) {
+            case "30m":
+                return WORKSPACE_TIMEOUT_DEFAULT_SHORT;
+            case this.config.workspaceDefaults.timeoutDefault || "60m":
+                return WORKSPACE_TIMEOUT_DEFAULT_LONG;
+            case this.config.workspaceDefaults.timeoutExtended || "180m":
+                return WORKSPACE_TIMEOUT_EXTENDED_ALT;
+            default:
+                return WORKSPACE_TIMEOUT_DEFAULT_SHORT;
+        }
     }
 
     async userGetsMoreResources(user: User): Promise<boolean> {
--- a/components/server/ee/src/workspace/gitpod-server-impl.ts
+++ b/components/server/ee/src/workspace/gitpod-server-impl.ts
@@ -45,6 +45,7 @@ import {
     Workspace,
     FindPrebuildsParams,
     TeamMemberRole,
+    WORKSPACE_TIMEOUT_DEFAULT_SHORT,
 } from "@gitpod/gitpod-protocol";
 import { ResponseError } from "vscode-jsonrpc";
 import {
@@ -77,7 +78,7 @@ import {
 import { Plans } from "@gitpod/gitpod-protocol/lib/plans";
 import * as pThrottle from "p-throttle";
 import { formatDate } from "@gitpod/gitpod-protocol/lib/util/date-time";
-import { FindUserByIdentityStrResult } from "../../../src/user/user-service";
+import { FindUserByIdentityStrResult, UserService } from "../../../src/user/user-service";
 import {
     Accounting,
     AccountService,
@@ -132,6 +133,8 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
     @inject(SnapshotService) protected readonly snapshotService: SnapshotService;
 
     @inject(UserCounter) protected readonly userCounter: UserCounter;
+
+    @inject(UserService) protected readonly userService: UserService;
 
     initialize(
         client: GitpodClient | undefined,
@@ -312,7 +315,7 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
             instancesWithReset.map((i) => {
                 const req = new SetTimeoutRequest();
                 req.setId(i.id);
-                req.setDuration(defaultTimeout);
+                req.setDuration(this.userService.workspaceTimeoutToDuration(defaultTimeout));
 
                 return client.setTimeout(ctx, req);
             }),
@@ -320,7 +323,7 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
 
         const req = new SetTimeoutRequest();
         req.setId(runningInstance.id);
-        req.setDuration(duration);
+        req.setDuration(this.userService.workspaceTimeoutToDuration(duration));
         await client.setTimeout(ctx, req);
 
         return {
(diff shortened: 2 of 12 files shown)


roboquat:
@corneliusludmann: Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://git.k8s.io/community/contributors/guide/release-notes.md) to remove it.

<details>

</details>

components/server/ee/src/user/user-service.ts
```diff
@@ -30,10 +32,35 @@ export class UserServiceEE extends UserService {
...
+        switch (duration) {
+            case "30m":
+                return WORKSPACE_TIMEOUT_DEFAULT_SHORT;
+            case this.config.workspaceDefaults.timeoutDefault || "60m":
+                return WORKSPACE_TIMEOUT_DEFAULT_LONG;
+            case this.config.workspaceDefaults.timeoutExtended || "180m":
+                return WORKSPACE_TIMEOUT_EXTENDED_ALT;
+            default:
+                return WORKSPACE_TIMEOUT_DEFAULT_SHORT;
+        }
```
corneliusludmann, line 63: I don't know if `user-service` is a good place for these functions. I was too miserly to create a dedicated service for this that does nothing but provide these values. However, I'm very open to proposals for better places.

akosyakov:
> All of these are little UI flaws that shouldn't be real blockers, in my opinion.

Could you file a follow-up issue for VS Code please? 🙏 and add to IDE team inbox

akosyakov:
I wonder why the PR is marked wit `team: IDE` label? There are no changes for IDE components. 

@mads-hartmann Should we remove this label? Or the deploy labeller is a smart enough to check only for teams which actually affected?

corneliusludmann:
@akosyakov I requested a review from the IDE team due to the IDE flaws with timeout extends. Just wanted to make sure that the IDE team has a look that is does not break the communication with the IDE.

image

mads-hartmann:
@akosyakov The label gets added when a team is requested as reviewers - the intention is that CODEOWNERS defines reviewers, but for cases like this where you just want a 2nd pair of eyes I agree it isn't ideal. Removing the label is these cases is the best way forward :)

MrSimonEmms (approved):

akosyakov:
A review from webapp team is required here to land correct?

MrSimonEmms:
@akosyakov I don't think there's much point at the moment as there's some conflicted files. I'll leave this today in the hope that @corneliusludmann is feeling better by tomorrow, otherwise I'll do my best at resolving the conflicts

MrSimonEmms:
Looks good. Can you update the `example-config.yaml` with the result of the `init` command? It adds the `workspace.maxLifetime` to it

corneliusludmann:
> Looks good. Can you update the `example-config.yaml` with the result of the `init` command? It adds the `workspace.maxLifetime` to it

Done.

MrSimonEmms (approved):

princerachit (approved):

jankeromnes:
Many thanks for this contribution! Taking a look on behalf of team WebApp 👀

components/server/ee/src/user/user-service.ts
```diff
@@ -32,10 +34,35 @@ export class UserServiceEE extends UserService {
...
-        return "60m";
+        return WORKSPACE_TIMEOUT_DEFAULT_LONG;
+    }
+
+    public workspaceTimeoutToDuration(timeout: WorkspaceTimeoutDuration): string {
+        switch (timeout) {
+            case WORKSPACE_TIMEOUT_DEFAULT_SHORT:
+                return "30m";
+            case WORKSPACE_TIMEOUT_DEFAULT_LONG:
+                return this.config.workspaceDefaults.timeoutDefault || "60m";
```
jankeromnes, line 48: Sorry if you've already explained this elsewhere, but I wonder what the reasoning is behind:
- Allowing configuration of `WORKSPACE_TIMEOUT_DEFAULT_LONG`
- Hard-coding `WORKSPACE_TIMEOUT_DEFAULT_SHORT` to `30m`

(Also, in your Pull Request description, the example has `DEFAULT_LONG` as `30m` and `EXTENDED` as `60m`, which I found a bit confusing.)

components/server/ee/src/user/user-service.ts
jankeromnes: I also have no idea if `user-service` is a good place for this, but if nobody has an opinion on this, then let's say it is a good place. 🙌

components/server/ee/src/user/user-service.ts
```diff
@@ -32,10 +34,35 @@ export class UserServiceEE extends UserService {
...
+            case WORKSPACE_TIMEOUT_EXTENDED:
+            case WORKSPACE_TIMEOUT_EXTENDED_ALT:
+                return this.config.workspaceDefaults.timeoutExtended || "180m";
+        }
+    }
+
+    public durationToWorkspaceTimeout(duration: string): WorkspaceTimeoutDuration {
```
jankeromnes, line 55: I find the types a little confusing in these method signatures:
- `duration` is a `string`, and the returned `Timeout` is a `WorkspaceTimeoutDuration`? 🤔
- Also `workspaceTimeoutToDuration` takes a `WorkspaceTimeoutDuration` and turns it into a `string` (would expect the other way around)

I feel like `duration` should be of type `...Duration`, and `timeout` of type `...Timeout`. 💭

Maybe this could be a follow-up issue though. ⏩

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -312,15 +315,15 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
+                req.setDuration(this.userService.workspaceTimeoutToDuration(defaultTimeout));
 
                 return client.setTimeout(ctx, req);
             }),
         );
 
         const req = new SetTimeoutRequest();
         req.setId(runningInstance.id);
-        req.setDuration(duration);
+        req.setDuration(this.userService.workspaceTimeoutToDuration(duration));
```
jankeromnes, line 326: Nit: Unfortunate variable naming 🙈 `somethingElseToDuration(duration)` ("...wait, isn't it already a duration?")

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -352,7 +355,7 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
 
         const client = await this.workspaceManagerClientProvider.get(runningInstance.region);
         const desc = await client.describeWorkspace(ctx, req);
-        const duration = desc.getStatus()!.getSpec()!.getTimeout() as WorkspaceTimeoutDuration;
+        const duration = this.userService.durationToWorkspaceTimeout(desc.getStatus()!.getSpec()!.getTimeout());
```
jankeromnes, line 358: Micro-nit: Same here, we're calling `somethingElseToTimeout(...getTimeout())` 🙈 (as in, turn a timeout into a timeout?)

components/server/src/user/user-service.ts
```diff
@@ -155,7 +155,32 @@ export class UserService {
...
+        switch (timeout) {
+            case WORKSPACE_TIMEOUT_DEFAULT_SHORT:
+                return "30m";
+            case WORKSPACE_TIMEOUT_DEFAULT_LONG:
+                return "60m";
+            case WORKSPACE_TIMEOUT_EXTENDED:
+            case WORKSPACE_TIMEOUT_EXTENDED_ALT:
+                return "180m";
+        }
+    }
```
jankeromnes, line 171: Aha, I see that you've made Gitpod Enterprise Edition configurable, but the non-Enterprise edition still has only hard-coded values. Just to be overly safe: That's intended, right?

jankeromnes (approved):
The WebApp component changes look good to me! ✨

I've also tested the latest deployment, and it seems to work flawlessly. 🥇

I've added a few comments/thoughts/questions in-line (mostly about naming things and better understanding this change, but please feel free to dismiss my comments and/or open follow-up issues to be addressed later without blocking your PR).

As a side note, this reminds me that I've always wanted to also make the Prebuild timeouts more "dynamic" (i.e. not necessarily configurable by installers as in your PR, but it would be nice -- in addition to the default 1h timeout -- to also have a 4h timeout for long-building projects, e.g. only for base prebuilds when you have Incremental Prebuilds enabled). No action required here, but sharing my thoughts here as they seem somewhat related.

Adding a hold in case you'd like to address some comments, but feel free to unhold any time:

/hold

corneliusludmann:
Thanks, @jankeromnes for your great feedback!

Since your comments are more very useful improvements and no blockers, as suggested, I'll let this pull request merge now and handle your comments in follow-up issues/pull requests. That makes the review process easier because we don't need approval from the other teams for the proposed changes.

Thanks again! :pray: 

/unhold

components/server/src/user/user-service.ts
corneliusludmann: That's intended in that way that it is similar to what we have before. If it makes sense or not is more a business decision (see also [this internal discussion](https://gitpod.slack.com/archives/C01KGM9DVRC/p1646666523461289)).

components/server/ee/src/user/user-service.ts
corneliusludmann: (see below, it's more or less what we had before with EE and non-EE)

neilkuantm:
Hi @corneliusludmann , how can I setting Default workspace CustomTime after install via kots?

shaoye:
#10270

PR #8580, opened 2022-03-03
[installer] Registry facade should not use a port from node ports range

aledbf:
## Description

Some CNI providers (as Cilium) do not allow overlap between `hostPorts` and `nodePorts`

## Release Notes

```release-note
[installer] Registry facade should not use a port from node ports range
```


install/installer/pkg/components/cluster/podsecuritypolicies.go
```diff
@@ -100,7 +101,7 @@ func podsecuritypolicies(ctx *common.RenderContext) ([]runtime.Object, error) {
 				},
 				HostNetwork: true,
 				HostPorts: []v1beta1.HostPortRange{{
-					Min: 30000,
+					Min: 20000,
```
sagor999, line 104: Hm. But this would add all ports from 20000-33000 range into SecPolicy. 
While registry facade is only using port 20000?

aledbf: fixed

sagor999 (approved):

MrSimonEmms (approved):

install/installer/pkg/components/image-builder-mk3/networkpolicy.go
```diff
@@ -16,34 +16,40 @@ import (
...
+									"169.254.169.254/32",
+								},
+							},
+						}},
 					},
-				}},
-			}},
+					common.AllowKubeDnsEgressRule(),
+					common.AllowWSManagerEgressRule(),
+				},
```
csweichel, line 51: Because mk3 doesn't run user workloads we could consider just allowing arbitrary egress. The current version is more secure though :)

csweichel (approved):

install/installer/pkg/components/image-builder-mk3/networkpolicy.go
aledbf: done

aledbf:
with the changes, we can run Gitpod in a cluster using Cilium

Screenshot from 2022-03-04 08-33-55

and use Hubble to see dropped packages

Screenshot from 2022-03-04 08-37-42

csweichel:
> with the changes, we can run Gitpod in a cluster using Cilium
> 
> Screenshot from 2022-03-04 08-33-55
> 
> and use Hubble to see dropped packages

Nice. Is there alerting that could use? E.g. an alert if a package makes its way through to the Google Metadata API server?

aledbf:
> Nice. Is there alerting that could use? E.g. an alert if a package makes its way through to the Google Metadata API server?

We can build one using the drop metric https://docs.cilium.io/en/stable/operations/metrics/#drop

csweichel (approved):

install/installer/pkg/components/registry-facade/podsecuritypolicy.go
```diff
@@ -38,8 +38,8 @@ func podsecuritypolicy(ctx *common.RenderContext) ([]runtime.Object, error) {
 			HostIPC:     false,
 			HostPID:     false,
 			HostPorts: []v1beta1.HostPortRange{{
-				Min: 30000,
-				Max: 33000,
+				Min: 20000,
+				Max: 20000,
```
MrSimonEmms, line 42: Presume this is fine with the min/max being the same value?

MrSimonEmms (approved):
/hold

Happy with that, just on the question raised. If that's fine, just remove the hold

install/installer/pkg/components/registry-facade/podsecuritypolicy.go
aledbf: We are exposing only one port. Also this was changed after feedback from @sagor999 https://github.com/gitpod-io/gitpod/pull/8580#discussion_r819019732

sagor999:
Ah, I think preview env is still using helm chart? So maybe that change needs to be done in there too now

aledbf:
Now is clear why we had a range in the policy

PR #8590, opened 2022-03-04
A little follow-up on cgroup v2 support

utam0k:
## Description

* Update the docs for cgroup v2
* Use runc's cgroup v2 determination

## Related Issue(s)

Fixes #

## How to test

No

## Release Notes

```release-note
Update the docs for cgroup v2
```

## Documentation

No

### DIFF
--- a/components/ws-daemon/pkg/iws/iws.go
+++ b/components/ws-daemon/pkg/iws/iws.go
@@ -818,6 +818,11 @@ func (wbs *InWorkspaceServiceServer) WriteIDMapping(ctx context.Context, req *ap
 	return &api.WriteIDMappingResponse{}, nil
 }
 
+// Allow workspace users to manipulate the cgroups to which the user process belong by constructing the cgroups of the following form
+//
+// <container-cgorup>  drwxr-xr-x 3 root      root
+// └── workspace       drwxr-xr-x 5 gitpodUid gitpodGid
+//     └── user        drwxr-xr-x 5 gitpodUid gitpodGid
 func (wbs *InWorkspaceServiceServer) EvacuateCGroup(ctx context.Context, req *api.EvacuateCGroupRequest) (*api.EvacuateCGroupResponse, error) {
 	if cgroups.Mode() != cgroups.Unified {
 		return &api.EvacuateCGroupResponse{}, nil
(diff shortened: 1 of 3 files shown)


utam0k:
@csweichel Thanks for previous your PR. Great! I have a question about the cgroup v2 hierarchy. Why do we need to create the user cgroup? I guess that if we just wanted to allow workspace users to manipulate the cgroup, we could just create the workspace cgroup.

utam0k:
If you want to review SVG, the rich diff is very helpful to you.
https://github.blog/2014-10-06-svg-viewing-diffing/

csweichel:
Thanks @utam0k for updating the diagram. At the moment it doesn't make the first cgroup evacuation explicit. The process is as follows:

1. `PrepareForUserNS`: evacuate container cgroup to `<containerCGroup>/workspace` which moves the ring0 process to this new `workspace` cgroup
2. start ring2 in its own cgroup namespace (i.e. `unshare(CLONE_NEWCGROUP)` in ring2)
3. Evacuate `<containerCGroup>/workspace` to `<containerCGroup>/workspace/user` and chown `<containerCGroup>/workspace` to the `gitpod` user

Motivation for this process is:
- the "root" cgroup that the user sees (i.e. the cgroup ring2 was in when we called unshare) must not contain processes. If it did, we could not enable threaded subtree controllers (e.g. `pid`) and have that cgroup be of type `domain`. It would necessarily be `domain threaded`. For k3s support the "ring2 cgroup" must be of type `domain`.
- by introducing the `workspace` cgroup, we can safely chown it to `gitpod` and give the user access because they're still "confined" by the container runtime's outer cgroup. If we didn't introduce `workspace` we'd need to give the user full control over the container cgroup so that new child cgroups could be created.

---

I was wondering: what's the motivation for using runc's cgroup implementation? It introduces a new dependency and with unclear benefit over the current implementation.

utam0k:
I got your explanation. Thanks. It makes sense. 
> Thanks @utam0k for updating the diagram. At the moment it doesn't make the first cgroup evacuation explicit. The process is as follows:
> 
> PrepareForUserNS: evacuate container cgroup to <containerCGroup>/workspace which moves the ring0 process to this new workspace cgroup
> start ring2 in its own cgroup namespace (i.e. unshare(CLONE_NEWCGROUP) in ring2)
> Evacuate <containerCGroup>/workspace to <containerCGroup>/workspace/user and chown <containerCGroup>/workspace to the gitpod user
> Motivation for this process is:
> 
> the "root" cgroup that the user sees (i.e. the cgroup ring2 was in when we called unshare) must not contain processes. If it did, we could not enable threaded subtree controllers (e.g. pid) and have that cgroup be of type domain. It would necessarily be domain threaded. For k3s support the "ring2 cgroup" must be of type domain.
> by introducing the workspace cgroup, we can safely chown it to gitpod and give the user access because they're still "confined" by the container runtime's outer cgroup. If we didn't introduce workspace we'd need to give the user full control over the container cgroup so that new child cgroups could be created.

---

Sorry, this was not clear enough to convey my intent. I'm worried about this https://github.com/gitpod-io/gitpod/pull/8471#discussion_r820314115. But I left that out of this PR.
> I was wondering: what's the motivation for using runc's cgroup implementation? It introduces a new dependency and with unclear benefit over the current implementation.

csweichel (approved):

Issue #8412, opened 2022-02-23
[licensing] Treat any case where we do not have a valid license as if no license is installed


corneliusludmann:
Currently, an invalid license prevents any users to sign up. An installation with an invalid license should not be in a worse position than an installation with no license at all.

We have this behavior already to some extent so that we don't return an invalid license but a default license for some cases, e.g. here:

https://github.com/gitpod-io/gitpod/blob/896fe39ccf456e132a4c7dba968c18f7394e6fc8/components/licensor/ee/pkg/licensor/replicated.go#L117-L119

However, there are some cases where we don't have a valid license, e.g. when the license server from Replicated is not reachable etc.

I would propose to change the implementation like this:
- Don't return the default license in `newReplicatedEvaluator()` but return the invalid license with the proper reason/explanation why the license is invalid.
- In the function `HasEnoughSeats`, don't respond with `false` when the license is invalid but call `HasEnoughSeats` from the default license in that case (same for other methods like `Enabled(...)`).

That has the following advantages:
- We still have the proper (invalid) license in place and could e.g. log why the license is invalid.
- We could add a dashboard that shows the (invalid) license (see #8413) .
- No installation acts worse than the default license behavior.

gtsiolis:
See also https://github.com/gitpod-io/gitpod/issues/3918 for historical context and in case we can close one of these two issues as duplicate. Cc @corneliusludmann @AlexTugarev

corneliusludmann:
Thanks, @gtsiolis. I think the other issue is more about handling error messages. I added this to our project board since this is something that I had on my radar as well.

MrSimonEmms:
See https://github.com/gitpod-io/gitpod/pull/8593#issuecomment-1062851893

gtsiolis:
Closing as the `licensor` component has been removed in https://github.com/gitpod-io/gitpod/pull/16983. Cc @aledbf @AlexTugarev @geropl

PR #8593, opened 2022-03-04
 Make the licensing match what's advertised 

### LINKED ISSUE #6932: [Self-hosted]: cannot search users/workspaces without enterprise license and error is console only
### Bug description

When I go to `/admin/users` in my self-hosted instance and click "search", I'm not able to search without an enterprise license. Whilst this is fine, the error only appears in the console.

It would be nice to have the error in the actual dashboard - an alert-style error

image

### Steps to reproduce

Install self-hosted, go to admin and click "search"

### Workspace affected

_No response_

### Expected behavior

Should display the "enterprise license required" in the dashboard, not just the console

### Example repository

_No response_

### Anything else?

_No response_

### LINKED ISSUE #8328: Make the licensing match what's advertised
On [Gitpod](https://www.gitpod.io/self-hosted), we advertise three self-hosted plans - community, free and professional.

In the [application](https://github.com/gitpod-io/gitpod/blob/main/components/licensor/ee/pkg/licensor/licensor.go#L81), we only provide two plan types - team and enterprise

This ticket is to make the application match what we advertise.

---

There are a few potential things that need to be thought about here

- Any changes need to maintain backwards compatibility with existing customers (including SaaS)
- We should support automatic downgrading to the community license if a user has a free license and the number of users increases from 10 to 11
- I would suggest that the admin dashboard is enabled for all license levels. This will allow us to put useful things in here in future

Linked to #8329

MrSimonEmms:
## Description

The self-hosted licensing now matches [what we advertise](https://www.gitpod.io/self-hosted), namely:

- Default to free/professional license which allows full access for a specific number of users. Free is for 10 users, pro is for however many their license allows. Internally, this is the "Enterprise" license.
- Fallback to community license, which allows for none of the EE features (except admin dashboard) for an unlimited number of users. Internally, this is the "Team" license.

This introduces the concept of a fallback license, in addition to the default license. It also adds the number of seats to the `Enabled` call in the licensor. As a workflow, if the license exceeds the number of seats requested in the call, it will try again with the fallback license, which is the community license.

### Fallback rules

| License type | Fallback allowed |
| --- | --- |
| No license (ie, installed via Installer) | Y |
| Gitpod paid license | N |
| Replicated community license | Y |
| Replicated paid license | N |

This table shows the fallback rules. It is done so that a user who is paying for Gitpod will have additional users rejected rather than having their service downgraded. It would be a pretty poor user experience if someone was paying for Gitpod, had an additional user sign up and then everyone lost all their premium services.

Gitpod licenses are always considered "paid". Now we are moving to Replicated for most licensing, Gitpod licenses will only be issued to paying customers who cannot use Replicated (eg, SaaS). Replicated has the concept of a community license - if their license is marked as "community", it will allow for fallback.

## Related Issue(s)

Fixes #8328
Fixes #6932

## How to test

Deploy to a self-hosted instance with a Replicated license. Tests already done:
 - Test prebuilds work with <=10 users
 - Test new user rejected when paid license has run out of seats
 - Test new user accepted when paid license changed to a community license
 - Test prebuilds rejected when on community license and > 10 users
 - Test prebuilds accepted when community license changed to a paid license with 20 users

## Release Notes

```release-note
 Make the licensing match what's advertised
```

### DIFF
--- a/components/server/ee/src/user/user-counter.ts
+++ b/components/server/ee/src/user/user-counter.ts
@@ -1,0 +1,31 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { injectable } from 'inversify';
+
+@injectable()
+export class UserCounter {
+    public expires: Date | null = null;
+
+    protected data: number | null = null;
+
+    protected readonly timeout: number = 60 * 1000; // Cache data for 1 minute
+
+    get count(): number | null {
+        if (this.expires !== null && Date.now() >= this.expires.getTime()) {
+            // The timestamp is in range - return the data
+            return this.data;
+        }
+        // Not in range - return null
+        return null;
+    }
+
+    set count(userCount: number | null) {
+        this.expires = new Date(Date.now() + this.timeout);
+
+        this.data = userCount;
+    }
+}
--- a/components/licensor/ee/cmd/validate.go
+++ b/components/licensor/ee/cmd/validate.go
@@ -25,7 +25,7 @@ var validateCmd = &cobra.Command{
 		domain, _ := cmd.Flags().GetString("domain")
 		licensorType, _ := cmd.Flags().GetString("licensor")
 
-		var e licensor.Evaluator
+		var e *licensor.Evaluator
 		switch licensorType {
 		case string(licensor.LicenseTypeReplicated):
 			e = licensor.NewReplicatedEvaluator(domain)
--- a/components/licensor/ee/pkg/licensor/gitpod.go
+++ b/components/licensor/ee/pkg/licensor/gitpod.go
@@ -61,6 +61,7 @@ func NewGitpodEvaluator(key []byte, domain string) (res *Evaluator) {
 	}
 
 	return &Evaluator{
-		lic: lic.LicensePayload,
+		lic:           lic.LicensePayload,
+		allowFallback: false, // Gitpod licenses cannot fallback - assume these are always paid-for
 	}
 }
--- a/components/licensor/ee/pkg/licensor/gitpod.go
+++ b/components/licensor/ee/pkg/licensor/gitpod.go
@@ -14,53 +14,11 @@ import (
 	"time"
 )
 
-// GitpodEvaluator determines what a license allows for
-type GitpodEvaluator struct {
-	invalid string
-	lic     LicensePayload
-}
-
-// Validate returns false if the license isn't valid and a message explaining why that is.
-func (e *GitpodEvaluator) Validate() (msg string, valid bool) {
-	if e.invalid == "" {
-		return "", true
-	}
-
-	return e.invalid, false
-}
-
-// Enabled determines if a feature is enabled by the license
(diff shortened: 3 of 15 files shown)


roboquat:

<details>

</details>

MrSimonEmms:
@geropl I kept the caching of the `getUsersCount` call as a [separate commit](https://github.com/gitpod-io/gitpod/pull/8593/commits/28b8f8d586e1b499b3933c036edaaaa337929b84), to make it easier for you to check I'm on the right track with what you suggested. Happy to keep as-is or to merge into the other `[server]` commit - your call

corneliusludmann:
Great progress, @MrSimonEmms! :pray: 

### Suggestion
I'm totally fine with moving this suggestion to a follow-up PR (since this PR is already a minimum viable change). However, in order to fix https://github.com/gitpod-io/gitpod/issues/8412 I would suggest the following change:

1. Allow **fallback** here:
https://github.com/gitpod-io/gitpod/blob/9c5a4d017f454744856ea2c2e11f886813aebfb8/components/licensor/ee/pkg/licensor/replicated.go#L90-L100
1. Return an **invalid license** that allows fallback here instead of the default replicated license:
https://github.com/gitpod-io/gitpod/blob/9c5a4d017f454744856ea2c2e11f886813aebfb8/components/licensor/ee/pkg/licensor/replicated.go#L118-L128 
3. Allow **fallback in case of an invalid license** here instead of returning `false`:
https://github.com/gitpod-io/gitpod/blob/dd508e3ffb903fb1869d6190626a4cdd72645a77/components/licensor/ee/pkg/licensor/licensor.go#L168-L171
https://github.com/gitpod-io/gitpod/blob/dd508e3ffb903fb1869d6190626a4cdd72645a77/components/licensor/ee/pkg/licensor/licensor.go#L186-L189


With this, we can display the result of
https://github.com/gitpod-io/gitpod/blob/dd508e3ffb903fb1869d6190626a4cdd72645a77/components/licensor/ee/pkg/licensor/licensor.go#L158-L165
in the admin dashboard and users know why the license is invalid and they still have basic functionality (as you would have with no license).

corneliusludmann (dismissed):
Licensor part looks good to me.

:information_source:  I would like to leave the final review of the server part to **Team WebApp**.

MrSimonEmms:
@corneliusludmann I'll remove the "Fixes" for #8412 and we can continue the discussion there

components/server/ee/src/user/user-counter.ts
```diff
@@ -0,0 +1,31 @@
...
+
+@injectable()
+export class UserCounter {
+    public expires: Date | null = null;
+
+    protected data: number | null = null;
+
+    protected readonly timeout: number = 60 * 1000; // Cache data for 1 minute
+
+    get count(): number | null {
```
geropl, line 17: nit: we usually do not use `null` but `undefined` instead. But ignore this for how, we'll have a lint for it shortly.

geropl (approved):
Thx @MrSimonEmms , looks good to me! 

Did not test the self-hosted setup, though. :wink:

components/server/ee/src/user/user-counter.ts
MrSimonEmms: @geropl Would you like me to change this? I don't mind - normally, I use `undefined` as well so I can't understand why I used `null`

corneliusludmann (approved):

Issue #8592, opened 2022-03-04
[prebuilds] Failed Prebuilds without Snapshot are marked as 'available'


svenefftinge:
There are many failing prebuilds without a snapshot that are still marked as `available` and therefore used for new workspace starts.
```SQL
SELECT * FROM d_b_prebuilt_workspace WHERE snapshot='' and state = 'available' order by creationTime DESC
```

- We need to figure out why these workspaces fail (many have an error message (Out of Memory) but some don't).
- Make sure they are marked as failed, so they are not considered when starting a workspace.

svenefftinge:
I'm looking into filtering those prebuilds out.

geropl:
Discovered yesterday [here](https://github.com/gitpod-io/gitpod/issues/8096#issuecomment-1058175304).

@svenefftinge I don#t think it's possible. Quick sync?

geropl:
We figured out there are two issues here:
 - @svenefftinge figured that we're failing to check the `prebuild.error` state when trying to determine whether a snapshot is `available` or not
 - there is a problem with failing prebuilds overriding their failed state (#8096)

geropl:
Removing the workspace label as the interpretation is webapps responsibility.

PR #8601, opened 2022-03-04
[prebuilds] Introduce 'failed' state for prebuilds

svenefftinge:
## Description

This PR introduces a "failure" state for prebuilds that have entirely failed to finish.
This is to distinguish them from failed tasks but finished rebuilds.

## Related Issue(s)
fixes https://github.com/gitpod-io/gitpod/issues/8592

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/dashboard/src/projects/Prebuilds.tsx
+++ b/components/dashboard/src/projects/Prebuilds.tsx
@@ -248,6 +248,8 @@ export function prebuildStatusLabel(prebuild?: PrebuildWithStatus) {
             return (<span className="font-medium text-blue-500 uppercase">running</span>);
         case "aborted":
             return (<span className="font-medium text-gray-500 uppercase">canceled</span>);
+        case "failed":
+            return (<span className="font-medium text-red-500 uppercase">failed</span>);
         case "timeout":
             return (<span className="font-medium text-red-500 uppercase">failed</span>);
         case "available":
--- a/components/ws-manager-bridge/ee/src/bridge.ts
+++ b/components/ws-manager-bridge/ee/src/bridge.ts
@@ -77,9 +77,9 @@ export class WorkspaceManagerBridgeEE extends WorkspaceManagerBridge {
                     prebuild.error = status.conditions!.timeout;
                     headlessUpdateType = HeadlessWorkspaceEventType.AbortedTimedOut;
                 } else if (!!status.conditions!.failed) {
-                    prebuild.state = "aborted";
+                    prebuild.state = "failed";
                     prebuild.error = status.conditions!.failed;
-                    headlessUpdateType = HeadlessWorkspaceEventType.Aborted;
+                    headlessUpdateType = HeadlessWorkspaceEventType.Failed;
                 } else if (!!status.conditions!.stoppedByRequest) {
                     prebuild.state = "aborted";
                     prebuild.error = "Cancelled";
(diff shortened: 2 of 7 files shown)


components/dashboard/src/projects/Prebuilds.tsx
```diff
@@ -248,6 +248,8 @@ export function prebuildStatusLabel(prebuild?: PrebuildWithStatus) {
             return (<span className="font-medium text-blue-500 uppercase">running</span>);
         case "aborted":
             return (<span className="font-medium text-gray-500 uppercase">canceled</span>);
+        case "failed":
```
jankeromnes, line 251: Drive-by comment: You probably want to handle `"failed"` in `prebuildStatusIcon` below as well (there is already a `StatusFailed` icon you can re-use).

JanKoehnlein (approved):
LGTM

components/ws-manager-bridge/ee/src/bridge.ts
```diff
@@ -89,6 +89,9 @@ export class WorkspaceManagerBridgeEE extends WorkspaceManagerBridge {
                     prebuild.error = status.conditions!.headlessTaskFailed;
                     prebuild.snapshot = status.conditions!.snapshot;
                     headlessUpdateType = HeadlessWorkspaceEventType.FinishedButFailed;
+                } else if (!status.conditions!.snapshot) {
+                    prebuild.state = "failed";
```
jankeromnes, line 93: Suggestion: Here you could also set `prebuild.error` to something like "Prebuild does not have a snapshot".

components/dashboard/src/projects/Prebuilds.tsx
easyCZ: As the number of error cases grows, it becomes harder to enumerate them. One possible way to simplify the API UX is to have a single `error` state and extend it with a `cause` which is set to a computer readable reason for the failure - TimedOut, RateLimited, FailedToStart, etc. The error message then becomes a human readable string to show in the UI etc.

jankeromnes: Good point @easyCZ. Alternatively, I think we could also special-case the few non-error cases, then consider "all the rest" as errors?

svenefftinge:
wow, that was quick feedback 😁 The branchname was too long so I had to crate [another PR](https://github.com/gitpod-io/gitpod/pull/8602).

components/ws-manager-bridge/ee/src/bridge.ts
svenefftinge: That is a symptom not the actual error. I've discussed with Gero that we will also relay overwriting actual errors if they exist (in case of multiple stop messages).

components/dashboard/src/projects/Prebuilds.tsx
easyCZ: Depends how far we want to go. Some failures are due to user issues, some can be due to system issues. I think the it mostly depends if we control all the `clients` which will be interpreting the status (Dashboard, etc) or if some of the clients are/will be out of our control (implementing against an API). In the former case, it doesn't matter much as we can largely reach a shared understanding of what is an error and what isn't. But if an external frontend/server needs to `switch` on these, it may be less obivous to them with a large number of cases.

I think this is probs not something we need/should decide now and I don't think it blocks the change.

PR #8602, opened 2022-03-04
[prebuilds] Introduce 'failed' state for prebuilds

svenefftinge:
## Description

This PR introduces a "failure" state for prebuilds that have entirely failed to finish.
This is to distinguish them from failed tasks but finished rebuilds.

## Related Issue(s)
fixes https://github.com/gitpod-io/gitpod/issues/8592

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/dashboard/src/projects/Prebuilds.tsx
+++ b/components/dashboard/src/projects/Prebuilds.tsx
@@ -103,7 +103,7 @@ export default function (props: { project?: Project, isAdminDashboard?: boolean 
     }, [prebuilds])
 
     const prebuildContextMenu = (p: PrebuildWithStatus) => {
-        const isFailed = p.status === "aborted" || p.status === "timeout" || !!p.error;
+        const isFailed = p.status === "aborted" || p.status === "timeout" || p.status === "failed"|| !!p.error;
         const isRunning = p.status === "building";
         const entries: ContextMenuEntry[] = [];
         if (isFailed) {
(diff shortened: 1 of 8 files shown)


components/ws-manager-bridge/ee/src/bridge.ts
```diff
@@ -77,18 +77,23 @@ export class WorkspaceManagerBridgeEE extends WorkspaceManagerBridge {
                     prebuild.error = status.conditions!.timeout;
                     headlessUpdateType = HeadlessWorkspaceEventType.AbortedTimedOut;
                 } else if (!!status.conditions!.failed) {
-                    prebuild.state = "aborted";
-                    prebuild.error = status.conditions!.failed;
-                    headlessUpdateType = HeadlessWorkspaceEventType.Aborted;
+                    prebuild.state = "failed";
+                    if (status.conditions!.failed)
```
AlexTugarev, line 81: this is always true, given we made it here.

components/dashboard/src/projects/Prebuilds.tsx
```diff
@@ -248,6 +248,8 @@ export function prebuildStatusLabel(prebuild?: PrebuildWithStatus) {
             return (<span className="font-medium text-blue-500 uppercase">running</span>);
         case "aborted":
             return (<span className="font-medium text-gray-500 uppercase">canceled</span>);
+        case "failed":
+            return (<span className="font-medium text-red-500 uppercase">failed</span>);
```
AlexTugarev, line 252: > This is to distinguish them from failed tasks but finished rebuilds.

How is this different from `available` & `prebuild.error !== ''` ?

AlexTugarev:
The error handling makes sense and the changes look good. I'm just not sure about the presentation on the dashboard, it seems to be the same as the "headless task failed" case. Could we render the "system error" case differently?

PR #8608, opened 2022-03-04
Introduce TeamSubscription.noMoreResources to be able to exclude TeamSubscriptions from bigger clusters

### LINKED ISSUE #8472: Limit who can receive has-more-resources permission
### Bug description

Users who have student discounts can receive the has-more-resources permission, if their plan is professional or unleashed.

### Steps to reproduce

Grant your user an unleashed plan who is also considered a student.

### Workspace affected

_No response_

### Expected behavior

`[userGetsMoreResources](https://github.com/gitpod-io/gitpod/blob/f36d44f6e6f940a7e14368bcbb876f100a013551/components/server/ee/src/user/eligibility-service.ts#L259)` should return false if the user has a student discount.

### Example repository

n/a

### Anything else?

@jldec can you think of any other scenarios where we apply discounts, aside from the student use case?

geropl:
## Description
This introduces a `noMoreResources` column on `TeamSubscription` that can be used to exclude it's members from receiving `more-resources`.

## Related Issue(s)

Fixes #8472 

## How to test
 - note how the unit tests are green.

## Release Notes

```release-note
NONE
```


geropl:
/werft run

Re-try certificate

components/gitpod-db/src/typeorm/entity/db-team-subscription.ts
```diff
@@ -51,4 +51,9 @@ export class DBTeamSubscription implements TeamSubscription {
         default: false
     })
     deleted?: boolean;
+
+    @Column({
+        default: false
+    })
+    noMoreResources?: boolean;
```
jankeromnes, line 58: I feel like I always point out the silly things, but, as always, I'm not sure about the naming. 😅

- `noMoreResources` sounds a bit confusing (e.g. it means the team doesn't have any resources anymore? 🤔)

How about something like:

- `noPremiumResources`
- `noExtraResources`
- `exemptFromPremiumResources`
- ... ?

geropl: Yeah, I feel you. I'm no fan of it, too. But it fits with other usage in our code-base, especially the relevant marker on clusters.

I also wondered if it make sense to make this field a generic JSON bag, with this field just being one of multiple possible tags. WDYT? :thinking:  This way we could more easily adjust once we explored what we want to support beyond "more-resources".

components/server/ee/src/user/eligibility-service.ts
```diff
@@ -270,7 +269,26 @@ export class EligibilityService {
...
+            return false;
+        }
+
+        // some TeamSubscriptions are marked with 'noMoreResources' to convey that those are _not_ receiving more resources
+        const noMoreResources = await Promise.all(relevantSubscriptions.map(async (s): Promise<boolean> => {
+            if (!s.teamSubscriptionSlotId) {
+                return false;
+            }
+            const ts = await this.teamSubscriptionDb.findTeamSubscriptionBySlotId(s.teamSubscriptionSlotId);
+            return !!ts?.noMoreResources;
```
jankeromnes, line 284: Thought: I feel like this could be more "elegant" a second filter right after the `eligiblePlans.includes(s.planId)` above, thus leading to a single `if (relevantSubscriptions.length === 0) return false` (i.e. without needing a `noMoreResources` interim variable or a `noMoreResources.every()` call)

components/gitpod-db/src/typeorm/entity/db-team-subscription.ts
jankeromnes: > Yeah, I feel you. I'm no fan of it, too. But it fits with other usage in our code-base, especially the relevant marker on clusters.

Ah 😬 I wasn't aware of these. Still, I don't think team DB should strictly follow the naming conventions of cluster DB (they're somewhat unrelated), and in any case, I think it's always better to use clear self-evident naming as opposed to following "unfortunate" per-existing naming conventions. 🙈

jankeromnes: > I also wondered if it make sense to make this field a generic JSON bag, with this field just being one of multiple possible tags. WDYT? 🤔

I think this would be premature optimization:
- We don't know yet whether we'd even use that
- The `d_b_team_subscription` is about to get deprecated anyway
- Aren't JSON bags done to avoid significant table modifications for tables with many rows? (E.g. would be okay with `d_b_workspace_instance`, but I don't think altering `d_b_team_subscription` is significant performance-wise)

geropl: I renamed to `excludeFromMoreResources` to better fit into the `TeamSubscription` perspective/model.

> I think this would be premature optimization:

True. Alos, really hope we can get rid of it rather sooner than later. To retiterate: The whole point of this is to avoid the (tremendous) overhead of (re-)adding a distinct Chargebee subscription for these kind of clients. We had that once, but decided to drop it... :confused:

components/server/ee/src/user/eligibility-service.ts
geropl: This is certainly partly a subjective thing. The problem here is that due to TS lagging proper combinators for async (like `filterAsync`), you'd need to encode both filters as `map` returning `Subscription | undefined`, followed by a `every` (or `filter` + length), which I feel is counterintuitive. So I decided to:
 - keep the filter where it feels natural
 - only use the complex `map` for async stuff

components/gitpod-db/src/typeorm/entity/db-team-subscription.ts
```diff
@@ -51,4 +51,9 @@ export class DBTeamSubscription implements TeamSubscription {
         default: false
     })
     deleted?: boolean;
+
+    @Column({
+        default: false
+    })
+    excludeFromMoreResources?: boolean;
```
jankeromnes, line 58: Can this really be `undefined`, given that the DB migration uses `tinyint(4) NOT NULL DEFAULT '0'`?

geropl: Fixed and pushed.

components/gitpod-protocol/src/team-subscription-protocol.ts
```diff
@@ -18,6 +18,8 @@ export interface TeamSubscription {
     paymentReference: string;
     cancellationDate?: string;
     deleted?: boolean;
+    /** If this flag is set slots are not eligibile for clusters with "more-resources" - even if their plan might be */
+    excludeFromMoreResources?: boolean;
```
jankeromnes, line 22: Nit: Can't be `undefined`.

```suggestion
    excludeFromMoreResources: boolean;
```

jankeromnes (approved):
Many thanks for the quick fix @geropl! 🚀

Code looks good to me, and the unit test adds a lot of confidence. Also, the migrated (empty) DB looks good. 👍

Still, I'd feel a bit safer testing this with an actual chargebee integration. Adding a hold in case you want to add an integration in this PR, but I'm also happy to fully test this on staging (i.e. please feel free to unhold 🚢).

/hold

components/gitpod-protocol/src/team-subscription-protocol.ts
geropl: Auto-save did not auto-save :rolling_eyes:

geropl:
/unhold

> Still, I'd feel a bit safer testing this with an actual chargebee integration.

As this is only about how we interpret our own model I think it's fine, especially with the test.

@jankeromnes Need re-approval for [the removed `?`](https://github.com/gitpod-io/gitpod/pull/8608#discussion_r820596668).

jankeromnes (approved):
Thanks, re-approved! 🚀

jankeromnes:
Oh, oops:

```
src/chargebee/team-subscription-handler.ts(69,52): error TS2345: Argument of type '{ userId: string; paymentReference: string; planId: string; startDate: string; endDate: string | undefined; quantity: number; }' is not assignable to parameter of type 'Omit<TeamSubscription, "id">'.
  Property 'excludeFromMoreResources' is missing in type '{ userId: string; paymentReference: string; planId: string; startDate: string; endDate: string | undefined; quantity: number; }' but required in type 'Omit<TeamSubscription, "id">'.
```

jankeromnes (approved):
Still good to go! 🚀

PR #8612, opened 2022-03-04
[installer] Allow range of ports for registry-facade pod security policy

aledbf:
## Description

Fix regression introduced by pod security policy (break dev-env)

## Release Notes
```release-note
[installer] Allow range of ports for registry-facade pod security policy
```


aledbf:
@iQQBot yeah, I would prefer that. Can you open a PR?

PR #8617, opened 2022-03-05
[server] Temp. disable expensive telemetry queries

geropl:
## Description
This change disables the execution of all telemetry queries. As discussed with @corneliusludmann  this it's meant as a temporary safety measure until a fix (#8638) is merged.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Documentation

 - [x] /werft no-preview=true

### DIFF
--- a/components/server/src/installation-admin/telemetry-data-provider.ts
+++ b/components/server/src/installation-admin/telemetry-data-provider.ts
@@ -18,9 +18,9 @@
      async getTelemetryData(): Promise<TelemetryData> {
             const data: TelemetryData = {
                 installationAdmin: await this.installationAdminDb.getData(),
-                totalUsers: await this.userDb.getUserCount(true),
-                totalWorkspaces: await this.workspaceDb.getWorkspaceCount(),
-                totalInstances: await this.workspaceDb.getInstanceCount(),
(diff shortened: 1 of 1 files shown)


geropl:
/hold

Don't merge

components/server/src/installation-admin/telemetry-data-provider.ts
```diff
@@ -18,9 +18,9 @@
      async getTelemetryData(): Promise<TelemetryData> {
             const data: TelemetryData = {
                 installationAdmin: await this.installationAdminDb.getData(),
-                totalUsers: await this.userDb.getUserCount(true),
-                totalWorkspaces: await this.workspaceDb.getWorkspaceCount(),
-                totalInstances: await this.workspaceDb.getInstanceCount(),
+                totalUsers: 0,  //await this.userDb.getUserCount(true),
```
jankeromnes, line 21: Why disable the user count query? (This seems unrelated to an index on workspace.type)

geropl: It was Saturday morning - and retrospectively, I would argue to be consistent :wink:

PR #8622, opened 2022-03-06
[installer] Add network policy for registry-facade access to coredns

aledbf:
## Description

kube-dns and coredns (helm chart) use different labels.

## Release Notes

```release-note
[installer] Add network policy for registry-facade access to coredns
```


Pothulapati (approved):
LGTM, and seems to [match the `core-dns` label that we are using](https://github.com/coredns/helm/blob/master/charts/coredns/templates/deployment.yaml#L41) !

PR #8629, opened 2022-03-07
Disable cache reclaim when cgroup v2.

### LINKED ISSUE #8497: Make cache reclaim support v2

utam0k:
## Description

Because cache reclaim supports only when using cgroup v1.

## Related Issue(s)

Fixes #8497

## How to test

1. Prepare an environment with cgroup v2 enabled e.g. `gitpod-k3s-202202181722`
2. Start a workspace
3. Confirm that ws-daemon doesn't complain error message about the cache reclaim

## Release Notes

```release-note
Disable cache reclaim when cgroup v2
```

## Documentation

No

### DIFF
--- a/components/ws-daemon/pkg/daemon/daemon.go
+++ b/components/ws-daemon/pkg/daemon/daemon.go
@@ -17,6 +17,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 
+	"github.com/gitpod-io/gitpod/common-go/cgroups"
 	"github.com/gitpod-io/gitpod/common-go/log"
 	"github.com/gitpod-io/gitpod/ws-daemon/api"
 	"github.com/gitpod-io/gitpod/ws-daemon/pkg/container"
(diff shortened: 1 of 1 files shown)


utam0k:
This PR is a draft because this pr is waiting for solving this discussion.
https://github.com/gitpod-io/gitpod/pull/8471#discussion_r819202598

components/ws-daemon/pkg/daemon/daemon.go
```diff
@@ -52,12 +53,22 @@ func NewDaemon(config Config, reg prometheus.Registerer) (*Daemon, error) {
...
 		cgCustomizer,
 		markUnmountFallback,
-	)
+	}
+
+	unified, err := cgroups.IsUnifiedCgroupSetup()
+	if err != nil {
+		return nil, xerrors.Errorf("could not determine cgroup setup: %w", err)
+	}
+	if unified {
```
Furisto, line 67: Shouldn't this be the other way around? That is add cache reclaim if we are on a system with a legacy (v1) cgroup setup?

utam0k: you are right! Thanks.

utam0k:
@Furisto PTAL

Furisto (approved):

PR #8632, opened 2022-03-07
Only allow exact-matching of workspace search

### LINKED ISSUE #8614: Disable empty string search on admin workspace search
Currently, all the admin dashboard searches allow an empty string search, which would return all the rows. 
Unfortunately, an empty string search on the admin workspace will spike our DB load and CPU usage, therefore can cause incidents[[1](https://app.incident.io/incidents/114)].

We should disable it.

laushinka:
## Description
This PR disables empty string and partial match search.

## Related Issue(s)

Fixes #8614

## How to test
1. Go to https://lm-empty-8614.staging.gitpod-dev.com/admin/workspaces
2. Click the Search button without any search queries. It should not do anything.
3. Enter a partial match of a workspace ID. It should not do anything.
4. Enter exact match of a workspace ID. It should return a result.

## Release Notes

```release-note
Admins cannot search empty strings or partial matches on workspace search.
```

### DIFF
--- a/components/dashboard/src/admin/WorkspacesSearch.tsx
+++ b/components/dashboard/src/admin/WorkspacesSearch.tsx
@@ -16,6 +16,7 @@ import { UserContext } from "../user-context";
 import { getProject, WorkspaceStatusIndicator } from "../workspaces/WorkspaceEntry";
 import { adminMenu } from "./admin-menu";
 import WorkspaceDetail from "./WorkspaceDetail";
+import info from '../images/info.svg';
 
 interface Props {
     user?: User
@@ -66,6 +67,11 @@ export function WorkspaceSearch(props: Props) {
     }
 
     const search = async () => {
+        // Disables empty search on the workspace search page
+        if (!props.user && queryTerm.length === 0) {
+            return;
+        }
+
         setSearching(true);
         try {
             const query: AdminGetWorkspacesQuery = {
@@ -75,6 +81,9 @@ export function WorkspaceSearch(props: Props) {
                 query.instanceIdOrWorkspaceId = queryTerm;
             } else if (matchesNewWorkspaceIdExactly(queryTerm)) {
                 query.workspaceId = queryTerm;
+            }
+            if (!query.ownerId && !query.instanceIdOrWorkspaceId && !query.workspaceId) {
+                return;
             }
 
             const result = await getGitpodService().server.adminGetWorkspaces({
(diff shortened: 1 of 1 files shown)


components/dashboard/src/admin/WorkspacesSearch.tsx
```diff
@@ -66,6 +67,15 @@ export function WorkspaceSearch(props: Props) {
     }
 
     const search = async () => {
+        if (!props.user) { // In the workspace search page
+            // This only allows searching for exact-matches of workspace or instance IDs
+            if ((queryTerm.length === 0) ||
+                (!matchesNewWorkspaceIdExactly(queryTerm) &&
```
geropl, line 73: IMO we should try to avoid calling the `matches` functions multiple times.

One way do to that is to move it too line 79 :point_down:

laushinka: Yes, good point!

laushinka: I was thinking we could return early when there is an empty search, and then check further down if they're partial matches, like so:

```
    const search = async () => {
        // Disables empty search on the workspace search page
        if (!props.user && queryTerm.length === 0) {
            return;
        }

        setSearching(true);
        try {
            ...
            if (!query.ownerId && !matchesNewWorkspaceIdExactly(queryTerm) && !matchesInstanceIdOrLegacyWorkspaceIdExactly(queryTerm)) {
                return;
            }
```

What do you think?

geropl (approved):
LGTM - thx for caring and improving safety! :pray:

components/dashboard/src/admin/WorkspacesSearch.tsx
```diff
@@ -105,6 +114,10 @@ export function WorkspaceSearch(props: Props) {
                 <button disabled={searching} onClick={search}>Search</button>
             </div>
         </div>
+        <div className={'flex rounded-xl bg-gray-200 dark:bg-gray-800 text-gray-600 dark:text-gray-400 p-2 w-2/3 mb-2'}>
+            <img className="w-4 h-4 m-1 ml-2 mr-4" alt="info" src={info} />
+            <span>Please enter complete IDs - this search does not perform partial-matching.</span>
```
gtsiolis, line 119: This one hurt! Opened https://github.com/gitpod-io/gitpod/issues/8701 to improve this.

Screenshot 2022-03-09 at 10 52 00 PM

Issue #8559, opened 2022-03-03
Support own CA certificate / self-signed certificated for self-hosted installations


corneliusludmann:
- [ ] Test what works and was not
- [ ] Write docs on how to use your own CA certificate
- [ ] Maybe add support to the installer if specific settings are needed
- [ ] Add support for CA in KOTS / Replicated installer

## See also:
- https://github.com/gitpod-io/gitpod/issues/2615
- https://github.com/gitpod-io/gitpod/pull/2984
- ([internal](https://gitpod.slack.com/archives/C01KLC56NP7/p1646286675863249))

corneliusludmann:
Comment from @MrSimonEmms ([internal](https://gitpod.slack.com/archives/C01KLC56NP7/p1646295351494919?thread_ts=1646286675.863249&cid=C01KLC56NP7)):

> one thing that would make startup so much faster is if we didn't require a cert (or would allow a self-signed one). I did a little bit of exploration around this - we need to be able to turn off the HSTS headers emitted by proxy. That alone would allow us to run with self-signed ones

csweichel:
If we could implement this without introducing too much variance (e.g. a special "custom CA cert" mode), that would be awesome! Those categorically variant modes don't age well (think "stage" as "prod" and "dev") :)

MrSimonEmms:
Agreed. My understanding is that we can get this working purely by allowing the HSTS headers in `proxy` to be configurable at deploy-time

SirLemyDanger:
I saw (blocking) errors related to this:

1. registry facade (needs to connect to our gitlab container registry, which has a TLS cert signed by the company CA)
2. server (oauth towards gitlab, which has a TLS cert signed by the company CA)
3. ??? when opening a workspace

VouDoo:
Hello 👋

I am facing the same issue that @SirLemyDanger explained.
In my environment, I have the following services:

- Self-hosted Gitpod
- Self-hosted GitLab server integrated with Gitpod
- Harbor used as image registry by Gitpod

All of those services use Company custom certificates (i.e. not trusted by a known public issuer).
When I try to create a workspace, I get this error (which does not say much):
![image](https://user-images.githubusercontent.com/16014882/174211625-63954103-cb29-491e-bb28-7eea18b0e366.png)

After investigation, I was able to find error entries in the log of the `registry-facade` pod:

```txt
http: TLS handshake error from XXX.XXX.XXX.XXX:XXXXX: remote error: tls: bad certificate
[...]

PR #8633, opened 2022-03-07
[kots]: make the self-hosted cert more explicitly selectable

MrSimonEmms:
## Description

The previous "in-cluster" naming still required the config to receive LetsEncrypt variables. This has now been switched to a "select self-signed" or "use a valid cert" flow

**IMPORTANT** this does nothing to make Gitpod run with a self-signed certificate (#8559). It is still a requirement to have a valid TLS cert. This flow would be used if generating a valid TLS cert and terminating against the load balancer, such as the [EKS Guide](https://github.com/gitpod-io/gitpod-eks-guide)

## How to test

Run in KOTS

## Release Notes

```release-note
[kots]: make the self-hosted cert more explicitly selectable
```

### DIFF
--- a/install/kots/manifests/gitpod-certificate.yaml
+++ b/install/kots/manifests/gitpod-certificate.yaml
@@ -3,12 +3,12 @@ kind: Certificate
 metadata:
   name: https-certificates
   annotations:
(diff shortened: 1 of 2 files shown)


Pothulapati (approved):
This option works as expected!

(I was struck on a AWS thing, but is good to go)

PR #8635, opened 2022-03-07
werft/build/job-config: Add withVM annotation if present on branch name

ArthurSens:
## Description

Since we can't pass annotations on commits anymore, this is needed to make Harvester load tests possible

## Related Issue(s)

Fixes https://github.com/gitpod-io/ops/issues/1444

## How to test

Create a new branch from this PR with `with-vm` present on its name 

Start a new werft job, it should trigger the creation of a Harvester VM.

Or just run `./dev/preview/test/load-test.sh`

## Release Notes

```release-note
NONE
```

### DIFF
--- a/dev/preview/test/load-test.sh
+++ b/dev/preview/test/load-test.sh
@@ -21,7 +21,7 @@ BASE_BRANCH_NAME="vm-load-test"
 BRANCH_SUFFIX="with-vm"
 
 opts=$(getopt \
-  --longoptions "start:,end:,base-branch-name:" \
+  --longoptions "start:,end:,base-branch-name:,interval:" \
   --name "$(basename "$0")" \
   --options "" \
   -- "$@"
--- a/dev/preview/test/load-test.sh
+++ b/dev/preview/test/load-test.sh
@@ -5,7 +5,7 @@
 #
 # Usage:
 #   ./dev/preview/test/load-test.sh
-#   ./dev/preview/test/load-test.sh --base-branch-name my-load-test-base-branch-name --start 1 --end 5
+#   ./dev/preview/test/load-test.sh --base-branch-name test --start 1 --end 5 --interval 30
 #
 
 set -euo pipefail
@@ -16,7 +16,9 @@ function log {
 
 START=1
 END=15
+INTERVAL=30
 BASE_BRANCH_NAME="vm-load-test"
+BRANCH_SUFFIX="with-vm"
 
 opts=$(getopt \
   --longoptions "start:,end:,base-branch-name:" \
@@ -32,6 +34,7 @@ while [[ $# -gt 0 ]]; do
     --start) START=$2 ; shift 2 ;;
     --end) END=$2 ; shift 2 ;;
     --base-branch-name) BASE_BRANCH_NAME=$2 ;  shift 2 ;;
+    --interval) INTERVAL=$2 ; shift 2 ;;
     *) break ;;
   esac
 done
@@ -42,6 +45,7 @@ Are you sure you want to run the load test with the following configuration?
 
 START=${START}
 END=${END}
+INTERVAL=${INTERVAL}
 BASE_BRANCH_NAME=${BASE_BRANCH_NAME}
 
 y/n: " yn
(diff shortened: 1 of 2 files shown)


.werft/jobs/build/prepare.ts
```diff
@@ -17,8 +18,23 @@ export async function prepare(werft: Werft) {
...
     } catch (err) {
         werft.fail('prep', err);
     }
+
+    werft.done('prep');
+    return config
+}
+
+function lookForBranchNameWithVM(branchName: string): boolean {
+    return branchName.includes("with-vm") ||
```
mads-hartmann, line 34: Just just stick with `with-vm`. Currently the load-test is the only consumer of this so lets keep it simple ☺️

dev/preview/test/load-test.sh
```diff
@@ -17,6 +17,7 @@ function log {
 START=1
 END=15
 BASE_BRANCH_NAME="vm-load-test"
+BRANCH_SUFFIX="withVM"
```
mads-hartmann, line 20: I'd prefer to stick with using hyphens and use `with-vm`

dev/preview/test/load-test.sh
```diff
@@ -52,19 +53,20 @@ y/n: " yn
     esac
 done
 
-log "Creating base branch ${BASE_BRANCH_NAME}"
-git checkout -b "${BASE_BRANCH_NAME}"
-git commit --allow-empty -m "Prepare load test" -m "/werft with-vm=true"
+FULL_BRANCH_NAME="${BASE_BRANCH_NAME}-${BRANCH_SUFFIX}"
+log "Creating base branch ${FULL_BRANCH_NAME}"
+git checkout -b "${FULL_BRANCH_NAME}"
+git commit --allow-empty -m "Prepare load test"
```
mads-hartmann, line 59: We don't need to create a commit anymore now that we don't need to add the Werft commands ☺️

.werft/build.ts
```diff
@@ -58,7 +58,7 @@ async function run(context: any) {
 
 
     await validateChanges(werft)
-    await prepare(werft)
+    config = await prepare(werft, config)
```
mads-hartmann, line 61: I would very much like to keep the config a const and not mutate the config options anywhere ☺️ It makes it easier to debug the the script if you keep the configuration parsing early and treat it as immutable for the rest of the job. So if possible, please move `lookForBranchNameWithVM` to job-config.ts and invoke the function there instead ☺️

.werft/jobs/build/prepare.ts
```diff
@@ -17,8 +18,23 @@ export async function prepare(werft: Werft) {
         exec("gcloud auth configure-docker --quiet");
         exec("gcloud auth configure-docker europe-docker.pkg.dev --quiet");
         exec('gcloud container clusters get-credentials core-dev --zone europe-west1-b --project gitpod-core-dev');
-        werft.done('prep');
+
+        if (lookForBranchNameWithVM(config.repository.branch)) {
```
mads-hartmann, line 22: When you move this to job-config.ts please also log a line that says that we're overriding the value ☺️

mads-hartmann (changes requested):
Thanks! I have a few change requests before I can approve ☺️

ArthurSens:
@mads-hartmann I ended up changing the strategy completely.

I've just added another conditional when creating the job config

dev/preview/test/load-test.sh
mads-hartmann: especially now that the ts code looks for with-vm :)

mads-hartmann (commented):
@ArthurSens The TS changes are much nicer, great job! I still have two unresolved comments though ☺️

ArthurSens:
Whoops, I guess I was a bit distracted when reading the review and missed a couple of things 😬 

I've addressed all comments, and also added a new flag for the load test script.

`--interval` can be useful here in case 30s is too fast or too slow for different scenarios

dev/preview/test/load-test.sh
```diff
@@ -5,7 +5,7 @@
 #
 # Usage:
 #   ./dev/preview/test/load-test.sh
-#   ./dev/preview/test/load-test.sh --base-branch-name my-load-test-base-branch-name --start 1 --end 5
+#   ./dev/preview/test/load-test.sh --base-branch-name test --start 1 --end 5 --interval 30
```
ArthurSens, line 8: I've changed the example branch name because of the 20 characters restriction 🤔

mads-hartmann (approved):
Awesome, LGTM. Added the hold label as CI hasn't completed running yet and don't want to mess with Tide :D

dev/preview/test/load-test.sh
```diff
@@ -16,7 +16,9 @@ function log {
 
 START=1
 END=15
+INTERVAL=30
 BASE_BRANCH_NAME="vm-load-test"
+BRANCH_SUFFIX="with-vm"
 
 opts=$(getopt \
   --longoptions "start:,end:,base-branch-name:" \
```
ArthurSens, line 24:
```suggestion
  --longoptions "start:,end:,base-branch-name:,interval:" \
```

ArthurSens:
Sorry mads, just noticed something was missing 😅

mads-hartmann (approved):
🎉

PR #8636, opened 2022-03-07
.werft/build: Cleaning code

ArthurSens:
## Description

Keep built.ts cleaner by removing branch length check from it

Add branch length check to validation phase, as it seems more appropriate, while making logic clearer with better error handling

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.werft/jobs/build/validate-changes.ts
+++ b/.werft/jobs/build/validate-changes.ts
@@ -1,12 +1,38 @@
 import { exec } from "../../util/shell";
 import { Werft } from "../../util/werft";
+import { JobConfig } from "./job-config";
 
-export async function validateChanges(werft: Werft) {
+export async function validateChanges(werft: Werft, config: JobConfig) {
     werft.phase('validate-changes', 'validating changes');
     try {
-        exec(`pre-commit run --all-files --show-diff-on-failure`);
-        werft.done('validate-changes');
+        branchNameCheck(werft, config)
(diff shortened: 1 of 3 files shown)


.werft/jobs/build/validate-changes.ts
```diff
@@ -1,12 +1,38 @@
...
+        if (config.previewEnvironment.destname.length > maxBranchNameLength) {
+            throw new Error(`The branch name ${config.previewEnvironment.destname} is more than ${maxBranchNameLength} character. Please choose a shorter name!`)
+        }
+        werft.done("check-branchname")
+    }
+}
+
+async function preCommitCheck(werft: Werft) {
+    werft.log("pre-commit checks", "Running pre-commit hooks.")
+    try {
```
mads-hartmann, line 32: As you're re-throwing the error I would suggest removing the try/catch logic here all-together ☺️

mads-hartmann (changes requested):
A tiny change request, otherwise I agree that the check fits better there ☺️

.werft/jobs/build/validate-changes.ts
ArthurSens: Ah nice catch...

I think we could follow this pattern on other areas too:
* One public function that is directly called by build.ts
* The public function would call private functions on a try-catch block
* Private functions will throw errors instead of running `werft.fail`

meysholdt (approved):

PR #8643, opened 2022-03-07
werft/build/prepare: Cleanup code

ArthurSens:
## Description

A couple of changes to make code cleaner:

* Only uses `exec` from `util/shell` instead of using it from `shelljs` at some places.
* Refactor the `prepare` function to no longer hold any logic, only calling several smaller private functions.
* Write small functions following the ["Do one thing"](https://blog.codinghorror.com/curlys-law-do-one-thing/#:~:text=Curly's%20Law%20is%20about%20choosing,your%20code%20won't%20do.) pattern that replaces the previous logic inside the `prepare` function

## Related Issue(s)

Fixes #

## How to test

Start a new job, you shouldn't notice any difference

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.werft/jobs/build/prepare.ts
+++ b/.werft/jobs/build/prepare.ts
@@ -1,24 +1,52 @@
-import * as shell from 'shelljs';
 import { exec } from '../../util/shell';
 import { Werft } from "../../util/werft";
 import { GCLOUD_SERVICE_ACCOUNT_PATH } from "./const";
 
+const phaseName = "prepare";
+
 export async function prepare(werft: Werft) {
-    werft.phase("prepare");
-
-    const werftImg = shell.exec("cat .werft/build.yaml | grep dev-environment").trim().split(": ")[1];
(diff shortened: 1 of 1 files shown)


.werft/jobs/build/prepare.ts
```diff
@@ -1,24 +1,52 @@
...
+        configureCoreDevAccess()
+    } catch (err) {
+        werft.fail(phaseName, err);
+    }
+    werft.done(phaseName);
+}
 
-    const werftImg = shell.exec("cat .werft/build.yaml | grep dev-environment").trim().split(": ")[1];
-    const devImg = shell.exec("yq r .gitpod.yml image").trim();
+// TODO: The reasoning behind this step should be clarified
```
meysholdt, line 20: it's because both images need to be updated separately and letting them go out of sync may cause weird follow-up errors.

meysholdt (approved):

PR #8607, opened 2022-03-04
Created a Global Prettier File

Harry-Hopkinson:
## Description
I have created a global .prettierrc.js file for Typescript and Js Files in order to remove inconsistent linting configurations which spanned across files.

PR #8644, opened 2022-03-07
[server] remove setting sync record with limit

### LINKED ISSUE #8589: Settings sync stores more than 20 resource revisions
### Bug description

Settings sync stores more than 20 resource revisions when it should only be 20 rev max

### Steps to reproduce

gp_setting_sync

mustard-mh:
## Description

Origin code looks fine. So sync-data more than 20, maybe it's caused by `db-sync`.

This PR will alleviate this problem.

## Related Issue(s)

Fixes #8589

## How to test

1. Start a workspace
2. Adjust UI and toggle setting sync to reach config limit 2 (debug only)
3. Repeat prev step, watch minio files and DB record

### How to access minio with UI
```shell
# make sure using correct namespace
kubens staging-hw-fix-8589

# get user and psw
kubectl get secrets minio -oyaml | yq - r "data.root-user" | base64 -d
kubectl get secrets minio -oyaml | yq - r "data.root-password" | base64 -d

# port forward minio console to 1234 (whatever you want)
kubectl pord-forward service/minio 1234:9001

# open port in your browser!! 
```
How to access your configuration?

Buckets -> search with your userId (Chrome DevTools to get it) -> Browser details -> blobs -> code-data -> globalState(which store your UI setting)

### How to access setting sync table
```shell
# get psw
kubectl get secrets mysql -oyaml | yq - r "data.password" | base64 -d

# exec into mysql
kubectl exec -it mysql-0 -c mysql sh

# get your data
mysql -u gitpod -p # then enter your password
use gitpod;
SELECT * FROM d_b_code_sync_resource WHERE userId=<your_user_id>;
```

### How to trigger UI update request to server
Drag your *DEBUG CONSOLE* to *TERMINAL* (any of them)
|Image|
|:-:|
|image|

## Release Notes

```release-note
Fix setting sync limit failure in some cases
```

### DIFF
--- a/components/gitpod-db/src/typeorm/code-sync-resource-db.ts
+++ b/components/gitpod-db/src/typeorm/code-sync-resource-db.ts
@@ -70,38 +70,44 @@ export class CodeSyncResourceDB {
         });
     }
 
-    async insert(userId: string, kind: ServerResource, doInsert: (rev: string, oldRev?: string) => Promise<void>, params?: CodeSyncInsertOptions): Promise<string | undefined> {
+    async deleteResource(userId: string, kind: ServerResource, rev: string, doDelete: (rev: string) => Promise<void>): Promise<void> {
+        const connection = await this.typeORM.getConnection();
+        await connection.transaction(async manager => {
+            await manager.createQueryBuilder()
+                .delete()
+                .from(DBCodeSyncResource)
+                .where("userId = :userId AND kind = :kind AND rev = :rev", { userId, kind, rev: rev })
+                .execute();
+            await doDelete(rev);
+        });
+    }
+
+    async insert(userId: string, kind: ServerResource, doInsert: (rev: string, oldRevs: string[]) => Promise<void>, params?: CodeSyncInsertOptions): Promise<string | undefined> {
         const connection = await this.typeORM.getConnection();
         return await connection.transaction(async manager => {
             let latest: DBCodeSyncResource | undefined;
-            let toUpdated: DBCodeSyncResource | undefined;
+            let toDeleted: DBCodeSyncResource[] = [];
             if (params?.revLimit) {
                 const resources = await this.doGetResources(manager, userId, kind);
                 latest = resources[0];
                 if (resources.length >= params.revLimit) {
-                    toUpdated = resources[resources.length - 1];
+                    // delete + 1 to insert instead of update
+                    toDeleted = resources.splice(params?.revLimit - 1);
                 }
             } else {
                 latest = await this.doGetResource(manager, userId, kind, 'latest');
             }
+            // user setting always show with diff so we need to make sure it’s changed from prev revision
             if (params?.latestRev && latest?.rev !== params.latestRev) {
                 return undefined;
             }
             const rev = uuid.v4();
-            await doInsert(rev, toUpdated?.rev);
-            if (toUpdated) {
-                await manager.createQueryBuilder().
-                    update(DBCodeSyncResource)
-                    .set({ rev })
-                    .where("userId = :userId AND kind = :kind AND rev = :rev", { userId, kind, rev: toUpdated.rev })
-                    .execute();
-            } else {
-                await manager.createQueryBuilder()
-                    .insert()
-                    .into(DBCodeSyncResource)
-                    .values({ userId, kind, rev })
-                    .execute();
-            }
+            await manager.createQueryBuilder()
+                .insert()
+                .into(DBCodeSyncResource)
+                .values({ userId, kind, rev })
+                .execute();
+            await doInsert(rev, toDeleted.map(e => e.rev));
             return rev;
         });
     }
--- a/components/server/src/code-sync/code-sync-service.ts
+++ b/components/server/src/code-sync/code-sync-service.ts
@@ -4,8 +4,8 @@
  * See License-AGPL.txt in the project root for license information.
  */
 
-import { status, ServiceError } from '@grpc/grpc-js';
-import fetch from "node-fetch";
+import { status } from '@grpc/grpc-js';
+import fetch from 'node-fetch';
 import { User } from '@gitpod/gitpod-protocol/lib/protocol';
 import bodyParser = require('body-parser');
 import * as util from 'util';
@@ -14,7 +14,7 @@ import { inject, injectable } from 'inversify';
 import { BearerAuth } from '../auth/bearer-authenticator';
 import { isWithFunctionAccessGuard } from '../auth/function-access';
 import { CodeSyncResourceDB, UserStorageResourcesDB, ALL_SERVER_RESOURCES, ServerResource, SyncResource } from '@gitpod/gitpod-db/lib';
-import { DeleteRequest, DownloadUrlRequest, DownloadUrlResponse, UploadUrlRequest, UploadUrlResponse } from '@gitpod/content-service/lib/blobs_pb';
+import { DeleteRequest, DeleteResponse, DownloadUrlRequest, DownloadUrlResponse, UploadUrlRequest, UploadUrlResponse } from '@gitpod/content-service/lib/blobs_pb';
 import { log } from '@gitpod/gitpod-protocol/lib/util/logging';
 import { v4 as uuidv4 } from 'uuid';
 import { accessCodeSyncStorage, UserRateLimiter } from '../auth/rate-limiter';
@@ -23,17 +23,17 @@ import { Config } from '../config';
 import { CachingBlobServiceClientProvider } from '@gitpod/content-service/lib/sugar';
 
 // By default: 5 kind of resources * 20 revs * 1Mb = 100Mb max in the content service for user data.
-const defautltRevLimit = 20;
+const defaultRevLimit = 20;
 // It should keep it aligned with client_max_body_size for /code-sync location.
 const defaultContentLimit = '1Mb';
 export type CodeSyncConfig = Partial<{
-    revLimit: number
-    contentLimit: number
+    revLimit: number;
+    contentLimit: number;
     resources: {
         [resource: string]: {
-            revLimit?: number
-        }
-    }
+            revLimit?: number;
+        };
+    };
 }>;
 
 const objectPrefix = 'code-sync/';
@@ -125,14 +125,14 @@ export class CodeSyncService {
                 return;
             }
             const resourceKey = ALL_SERVER_RESOURCES.find(key => key === req.params.resource);
-            const revs = resourceKey && await this.db.getResources(req.user.id, resourceKey)
+            const revs = resourceKey && (await this.db.getResources(req.user.id, resourceKey));
             if (!revs || !revs.length) {
                 res.sendStatus(204);
                 return;
             }
-            const result: { url: string, created: number }[] = revs.map(e => ({
+            const result: { url: string; created: number }[] = revs.map(e => ({
                 url: req.originalUrl + '/' + e.rev,
-                created: Date.parse(e.created) / 1000 /* client expects in secondsm */
+                created: Date.parse(e.created) / 1000 /* client expects in secondsm */,
             }));
             res.json(result);
             return;
@@ -189,8 +189,8 @@ export class CodeSyncService {
                     const response = await fetch(urlResponse.getUrl(), {
                         timeout: 10000,
                         headers: {
-                            'content-type': contentType
-                        }
+                            'content-type': contentType,
+                        },
                     });
                     if (response.status !== 200) {
                         throw new Error(`code sync: blob service: download failed with ${response.status} ${response.statusText}`);
@@ -224,11 +224,11 @@ export class CodeSyncService {
             if (latestRev === fromTheiaRev) {
                 latestRev = undefined;
             }
-            const revLimit = resourceKey === 'machines' ? 1 : config.resources?.[resourceKey]?.revLimit || config?.revLimit || defautltRevLimit;
+            const revLimit = resourceKey === 'machines' ? 1 : config.resources?.[resourceKey]?.revLimit || config?.revLimit || defaultRevLimit;
             const userId = req.user.id;
-            let oldObject: string | undefined;
             const contentType = req.headers['content-type'] || '*/*';
-            const rev = await this.db.insert(userId, resourceKey, async (rev, oldRev) => {
+            let oldRevList: string[] = [];
+            const rev = await this.db.insert(userId, resourceKey, async (rev, oldRevs) => {
                 const request = new UploadUrlRequest();
                 request.setOwnerId(userId);
                 request.setName(toObjectName(resourceKey, rev));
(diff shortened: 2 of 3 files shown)


jeanp413:
@mustard-mh For globalState resource I see 30 revs, did you found something than causes this?
image

mustard-mh:
> @mustard-mh For globalState resource I see 31 revs, did you found something that causes this?

@jeanp413 I think it's because of `db-sync`. 

Original code looks fine, and it's correct with self-hosted Gitpod

mustard-mh:
The original code, only in case of high concurrency, may have this problem. But according to the time of database rows, it is not a high concurrency.

Original code only updates the last record and deletes the last record JSON file. 

If there is no high concurrency and only one database, it'll work fine. But we have `db-sync`, it'll sync other records from another DB, and sometimes causes some bug, then we get more and more records.

cc @jeanp413

components/gitpod-db/src/typeorm/code-sync-resource-db.ts
```diff
@@ -70,38 +70,43 @@ export class CodeSyncResourceDB {
...
                 latest = resources[0];
                 if (resources.length >= params.revLimit) {
-                    toUpdated = resources[resources.length - 1];
+                    // delete + 1 to insert instead of update
+                    toDeleted = resources.splice(params?.revLimit - 1);
                 }
             } else {
                 latest = await this.doGetResource(manager, userId, kind, 'latest');
             }
+            // TODO(hw): why we use `latest?.rev !== params.latestRev` ? what's latestRev?
```
jeanp413, line 99: latestRev is the `etag` of the resource, if the `etag` send from the client is different from the server then it means it got updated already by some other client so we return 412 error, here's an example of how [they mock it](https://github.com/microsoft/vscode/blob/98c97abb8a33a34a3b82434c91ef93fe5ac819fc/src/vs/platform/userDataSync/test/common/userDataSyncClient.ts#L250-L252) in a test

components/server/src/code-sync/code-sync-service.ts
```diff
@@ -303,4 +290,24 @@ export class CodeSyncService {
...
+    protected async deleteObjects(userId: string, resourceKey: ServerResource, revs: string[]) {
+        for (let rev of revs) {
+            const obj = toObjectName(resourceKey, rev)
+            const request = new DeleteRequest();
+            request.setOwnerId(userId);
+            request.setExact(obj);
+            const blobsClient = this.blobsProvider.getDefault();
+            try {
+                await util.promisify<DeleteRequest, DeleteResponse>(blobsClient.delete.bind(blobsClient))(request);
+                await this.db.deleteResource(userId, resourceKey, rev);
```
akosyakov, line 302: I wonder whether it will be better to perform deletion of blob in the transaction of deleting a resource, i.e.:
- open transaction
- delete record from DB
- delete blob from storage (ignore not found exception)
- close transaction

components/server/src/code-sync/code-sync-service.ts
```diff
@@ -253,24 +252,12 @@ export class CodeSyncService {
                 if (response.status !== 200) {
                     throw new Error(`code sync: blob service: upload failed with ${response.status} ${response.statusText}`);
                 }
-                oldObject = oldRev && toObjectName(resourceKey, oldRev);
+                // delete old revs from storage
+                this.deleteObjects(userId, resourceKey, oldRevs).then().catch(e => {
```
akosyakov, line 256: this part should be executed only after successful insert transaction, not before, otherwise it can happen that we delete data but not insert

components/gitpod-db/src/typeorm/code-sync-resource-db.ts
akosyakov: @mustard-mh that's a good reference, i used this test impl as a blueprint while implementing our code sync endpoing

akosyakov:
I removed this PR from IDE project and added the corresponding issue instead.

components/server/src/code-sync/code-sync-service.ts
mustard-mh: I think it's the same with steps now

components/server/src/code-sync/code-sync-service.ts
mustard-mh: change DB insert and doInsert order can resolve

components/gitpod-db/src/typeorm/code-sync-resource-db.ts
mustard-mh: Got it! 🙏

components/server/src/code-sync/code-sync-service.ts
akosyakov: I don't think so, since we call from within transaction still, no?

mustard-mh: Yes, we trust that the transaction commit will succeed so we put the object to GCP, and we delete objects in this step too. (I added a commit now)

We need to resolve problems if things we trust don't do like we expected, that's why we need some GC things, maybe workspace images also have some leak

akosyakov: why cannot we just move out of transaction before, so we don't trust, but know for sure?

components/server/src/code-sync/code-sync-service.ts
akosyakov: Is not it that if deleteResource fails then we have a record without blob?

components/server/src/code-sync/code-sync-service.ts
```diff
@@ -253,28 +253,17 @@ export class CodeSyncService {
...
-                        }
-                        log.error({ userId }, 'code sync: failed to delete', err, { object: oldObject });
-                    }
-                });
-            }
             if (!rev) {
                 res.sendStatus(412);
                 return;
             }
+            // delete old revs from storage
```
akosyakov, line 262: Should not it be performed regardless of if above?

components/server/src/code-sync/code-sync-service.ts
akosyakov: this one does not seem to be addressed, deletion of blob happens outside of delete DB transaction, that can deleted blob, but no a DB record

components/gitpod-db/src/typeorm/code-sync-resource-db.ts
```diff
@@ -70,38 +70,43 @@ export class CodeSyncResourceDB {
...
 
-    async insert(userId: string, kind: ServerResource, doInsert: (rev: string, oldRev?: string) => Promise<void>, params?: CodeSyncInsertOptions): Promise<string | undefined> {
+    async deleteResource(userId: string, kind: ServerResource, rev: string): Promise<void> {
+        const connection = await this.typeORM.getConnection();
+        await connection.transaction(async manager => {
+            await manager.createQueryBuilder()
+                    .delete()
+                    .from(DBCodeSyncResource)
+                    .where("userId = :userId AND kind = :kind AND rev = :rev", { userId, kind, rev: rev })
+                    .execute();
```
akosyakov, line 80: we should delete blob here

mustard-mh: updated

components/server/src/code-sync/code-sync-service.ts
```diff
@@ -303,4 +292,25 @@ export class CodeSyncService {
...
+                await util.promisify<DeleteRequest, DeleteResponse>(blobsClient.delete.bind(blobsClient))(request);
+            } catch (err) {
+                if (err.code === status.NOT_FOUND) {
+                    return;
+                }
+                log.error({ userId }, 'code sync: failed to delete obj', err, { object: obj });
+                throw err;
+            }
+        }).catch(err => {
+            log.error({ userId }, 'code sync: failed to delete', err);
```
jeanp413, line 312: Remove this catch and handle it on caller function?

components/server/src/code-sync/code-sync-service.ts
```diff
@@ -253,24 +253,13 @@ export class CodeSyncService {
...
-                        if (err.code === status.NOT_FOUND) {
-                            // we're good here
-                            return;
-                        }
-                        log.error({ userId }, 'code sync: failed to delete', err, { object: oldObject });
-                    }
-                });
-            }
+            // delete old revs from storage
+            this.deleteObjects(userId, resourceKey, oldRevList).then().catch(e => {
```
jeanp413, line 259: Remove `.then()`

components/server/src/code-sync/code-sync-service.ts
```diff
@@ -303,4 +292,25 @@ export class CodeSyncService {
         return router;
     }
 
+    protected async deleteObjects(userId: string, resourceKey: ServerResource, revs: string[]) {
+        revs.forEach(rev => this.db.deleteResource(userId, resourceKey, rev, async (rev: string) => {
```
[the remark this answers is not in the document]
mustard-mh: 😅 I use it because I only want to exec with all this array, forgot it's an `async` function, will do it

components/server/src/code-sync/code-sync-service.ts
mustard-mh: Any of them failed is ok, we should continue to delete other items

components/server/src/code-sync/code-sync-service.ts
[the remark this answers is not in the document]
mustard-mh: @jeanp413 updated

components/server/src/code-sync/code-sync-service.ts
jeanp413: If you catch here then this function will never throw and catch below :arrow_down: will never execute

components/server/src/code-sync/code-sync-service.ts
```diff
@@ -303,4 +292,26 @@ export class CodeSyncService {
...
+                if (err.code === status.NOT_FOUND) {
+                    return;
+                }
+                log.error({ userId }, 'code sync: failed to delete obj', err, { object: obj });
+                throw err;
+            }
+        }).catch(err => {
+            log.error({ userId }, 'code sync: failed to delete', err);
+        }))
+        return Promise.all(tasks)
```
jeanp413, line 314: > Any of them failed is ok, we should continue to delete other items

Then is better to use `Promise.allSettled`

mustard-mh: 👍 Great! I have never used this func before! but in this case, we need some logs which may relate to DB right? We can use `Promise.allSettled` to make it more readable.

mustard-mh: 🙏 updated @jeanp413

jeanp413: If we keep the catch here then `Promise.all(tasks) will always resolve successfully and no need to change it to `Promise.allSettled`, other option is to remove catch here and use `Promise.allSettled` handle any error in caller function

components/server/src/code-sync/code-sync-service.ts
```diff
@@ -253,24 +253,10 @@ export class CodeSyncService {
...
-                        if (err.code === status.NOT_FOUND) {
-                            // we're good here
-                            return;
-                        }
-                        log.error({ userId }, 'code sync: failed to delete', err, { object: oldObject });
-                    }
-                });
-            }
+            // sync delete old revs from storage
+            this.deleteObjects(userId, resourceKey, oldRevList)
```
jeanp413, line 259: await?

mustard-mh: We don't need to await this delete func, unless client will fetch list immediately after post their configuration

components/server/src/code-sync/code-sync-service.ts
jeanp413: I was thinking of something like this, but it's fine either way, feel free to resolve this
```typescript
this.deleteObjects(userId, resourceKey, oldRevList).then(results => {
	results.forEach(p => {
		if (p.status === 'rejected') {
			log.error({ userId }, 'code sync: failed to delete', err);
		} 
	});
});
```

components/server/src/code-sync/code-sync-service.ts
```diff
@@ -303,4 +289,27 @@ export class CodeSyncService {
...
+                if (err.code === status.NOT_FOUND) {
+                    return;
+                }
+                log.error({ userId }, 'code sync: failed to delete obj', err, { object: obj });
+                throw err;
+            }
+        }).catch(err => {
+            log.error({ userId }, 'code sync: failed to delete', err);
+        }))
+        await Promise.allSettled(tasks)
```
jeanp413, line 311: missing semicolon, no need to await the last promise just return it

components/server/src/code-sync/code-sync-service.ts
```diff
@@ -24,7 +24,7 @@ import { Config } from '../config';
 import { CachingBlobServiceClientProvider } from '@gitpod/content-service/lib/sugar';
 
 // By default: 5 kind of resources * 20 revs * 1Mb = 100Mb max in the content service for user data.
-const defautltRevLimit = 20;
+const defaultRevLimit = 2;
```
jeanp413, line 27: revert this :smile:

components/server/src/code-sync/code-sync-service.ts
mustard-mh: await first and return void to avoid caller use this to check result 😄 https://github.com/gitpod-io/gitpod/pull/8644#discussion_r822938267

components/server/src/code-sync/code-sync-service.ts
mustard-mh: It's only for debug, I'll revert this after review ok https://github.com/gitpod-io/gitpod/pull/8644/commits/6bd58d59ed5eb15ecf676e77939622748e252780

components/server/src/code-sync/code-sync-service.ts
```diff
@@ -253,24 +253,10 @@ export class CodeSyncService {
                 if (response.status !== 200) {
                     throw new Error(`code sync: blob service: upload failed with ${response.status} ${response.statusText}`);
                 }
-                oldObject = oldRev && toObjectName(resourceKey, oldRev);
+                oldRevList = oldRevs
```
jeanp413, line 256: missing semicolon

components/server/src/code-sync/code-sync-service.ts
jeanp413, line 259: missing semicolon

mustard-mh: That's why we need some lint !!

mustard-mh: I create a temp .prettierrc file like this to reformat temporarily
```json
{
  "tabWidth": 4,
  "useTabs": false,
  "printWidth": 400,
  "singleQuote": true,
  "arrowParens": "avoid"
}
```

mustard-mh: @Harry-Hopkinson created a global lint for us 🍊 https://github.com/gitpod-io/gitpod/pull/8607 and @easyCZ created one too https://github.com/gitpod-io/gitpod/pull/8655!!

Hope them can be merged soon 👍

jeanp413 (approved):

mustard-mh:
I added `How to access minio UI` and `How to access setting sync table` section. 

cc @jeanp413

components/server/src/code-sync/code-sync-service.ts
```diff
@@ -253,24 +252,10 @@ export class CodeSyncService {
...
-                        if (err.code === status.NOT_FOUND) {
-                            // we're good here
-                            return;
-                        }
-                        log.error({ userId }, 'code sync: failed to delete', err, { object: oldObject });
-                    }
-                });
-            }
+            // sync delete old revs from storage
+            this.deleteObjects(userId, resourceKey, oldRevList)
```
jeanp413, line 258: @mustard-mh I just found we have lint check [server package.json](https://github.com/gitpod-io/gitpod/blob/a81c23ee504048414865a53de7af81d2a33b21cf/components/server/package.json#L20-L21)
If you run it we get an errorin this line `ERROR: 258:13  no-floating-promises  Promises must be handled appropriately` so maybe we should go with https://github.com/gitpod-io/gitpod/pull/8644#discussion_r822938267 ?

mustard-mh: I think we can just pass `.catch(e => {})` to it to pass this lint

jeanp413: I think that's kinda misleading but your call :+1:

jeanp413 (approved):

components/server/src/code-sync/code-sync-service.ts
easyCZ: This has now landed. Not all components are enabled by default but `server` is.

geropl (approved):

mustard-mh:
removed debug commit

jeanp413:
need to resolve merge conflicts now :eye

geropl (approved):

PR #8648, opened 2022-03-07
Namespace prebuild tracing tags

easyCZ:
## Description

Namespace `startPrebuild` tags as mentioned in https://github.com/gitpod-io/gitpod/pull/8568#discussion_r820887740

## Open Questions
* Fields like `contextURL`, `project` and few others feel like shared domain tags between many RPCs - do we have any unified tagging for domain properties which are wide-reaching? In practice, we want to be able to find relevant traces for any of these properties and for that we need a common tag schema.

## Release Notes

```release-note
Namespace prebuild tags
```

### DIFF
--- a/components/server/ee/src/prebuilds/prebuild-manager.ts
+++ b/components/server/ee/src/prebuilds/prebuild-manager.ts
@@ -69,9 +69,11 @@ export class PrebuildManager {
 
     async startPrebuild(ctx: TraceContext, { contextURL, cloneURL, commit, branch, project, user }: StartPrebuildParams): Promise<StartPrebuildResult> {
         const span = TraceContext.startSpan("startPrebuild", ctx);
-        span.setTag("contextURL", contextURL);
-        span.setTag("cloneURL", cloneURL);
-        span.setTag("commit", commit);
+        span.setTag("prebuild.contextURL", contextURL);
+        span.setTag("prebuild.cloneURL", cloneURL);
+        span.setTag("prebuild.commit", commit);
+        span.setTag("prebuild.branch", branch);
+        span.setTag("prebuild.project", project);
 
         try {
             if (user.blocked) {
@@ -134,8 +136,8 @@ export class PrebuildManager {
                 prebuild.error = "Prebuild is rate limited. Please contact Gitpod if you believe this happened in error.";
 
                 await this.workspaceDB.trace({ span }).storePrebuiltWorkspace(prebuild);
-                span.setTag("starting", false);
-                span.setTag("ratelimited", true);
+                span.setTag("prebuild.starting", false);
+                span.setTag("prebuild.ratelimited", true);
                 return {
                     wsid: workspace.id,
                     prebuildId: prebuild.id,
@@ -143,7 +145,7 @@ export class PrebuildManager {
                 };
             }
 
-            span.setTag("starting", true);
+            span.setTag("prebuild.starting", true);
             const projectEnvVars = await projectEnvVarsPromise;
             await this.workspaceStarter.startWorkspace({ span }, workspace, user, [], projectEnvVars, {excludeFeatureFlags: ['full_workspace_backup']});
             return { prebuildId: prebuild.id, wsid: workspace.id, done: false };


components/server/ee/src/prebuilds/prebuild-manager.ts
```diff
@@ -69,9 +69,11 @@ export class PrebuildManager {
...
     async startPrebuild(ctx: TraceContext, { contextURL, cloneURL, commit, branch, project, user }: StartPrebuildParams): Promise<StartPrebuildResult> {
         const span = TraceContext.startSpan("startPrebuild", ctx);
-        span.setTag("contextURL", contextURL);
-        span.setTag("cloneURL", cloneURL);
-        span.setTag("commit", commit);
+        span.setTag("prebuild.contextURL", contextURL);
+        span.setTag("prebuild.cloneURL", cloneURL);
+        span.setTag("prebuild.commit", commit);
+        span.setTag("prebuild.branch", branch);
+        span.setTag("prebuild.project", project);
```
easyCZ, line 76: Project, for example, is a first class domain property which exists in other features. It feels _odd_ to namespace it with `prebuild`. This effectively prevents tracing across services and features.

mads-hartmann: How to namespace your attributes depends on how you'd like to query for it.

1. Find all spans related to a specific entity (e.g. user). Then you'd do something like `COUNT where user.id = "foobar" group by name`. A query like that is only possible if we use `user.id` for the user id everywhere.
2. Filter/group a specific span based on attributes. For this it's a much nicer experience in Honeycomb if the attributes are namespaced using the name of the span. E.g. in this case `startPrebuild.cloneURL`. When you write your Honeycomb query you can then do `COUNT where name = startPrebuild, startPrebuild.<autocomplete>` and it will show all the attributes that it has ever seen for that span. If we didn't namespace it then you'd have to know attributes might be available on the span, e.g. that `context.cloneURL` happens to be a valid attribute for that span ☺️

So for this PR I would use the `startPrebuild` namespace for all the attributes (so rename `prebuild` to `startPrebuild`). If you believe the use-case (1) is valid for any of the attributes then just add those as well. A bit of duplication is fine, we pay per event, not by the number of attributes.

If you do add a new top-level namespace please do update [the document](https://www.notion.so/gitpod/Naming-conventions-af9084c4743349089cf2450a3c860733) though! ☺️

easyCZ: From my experience, I'd actually expect both to work. Let me elaborate:

I would indeed expect to be able to find all interactions by a given user, or everything related to a given workspace. Here I'd use the `user.id` or `workspace.id` fields.

But I'd also expect to be able to drill deeper and get all prebuild (startPrebuild) traces and do operations on them which aggregate over (or away) the `user.id` or `workspace.id` fields. I'd expect to query this with a more specific tag - `component: prebuild` or something along the lines. This would be yet another field which helps me go deeper. Each span could then also have very custom attributes of course but those would be required to be namespaced as I wouldn't expect these to be cross-cutting, or have semantics outside of a given `component: prebuild` trace.

I'm happy to follow whichever guidence you provide as I don't yet have enough context but it _feels_ we're preventing ourselves from leveraging some features of distributed tracing.

mads-hartmann: > I'm happy to follow whichever guidence you provide as I don't yet have enough context but it feels we're preventing ourselves from leveraging some features of distributed tracing.

We are not, it's a trade-off. If you want to get discoverability of what attributes are available on a specific span as you write your query for that specific span, then have to go with (2). 

If you want to find all spans for a specific entity go with (1).

If you need both, go with (1) and (2).

In my experience, when you're debugging you're mostly trying to figure out why a specific operation (span) is slow. For that you want to run some aggregates over all of those spans and do some grouping on the attributes to help you drill down. So providing a good experience when writing queries of the form `<AGGREGATE> WHERE name = startPrebuild GROUP BY startPrebuild.XXX <CODE COMPLETE>` is a worthwhile thing to optimise for.

And as I said, you can just add both startPrebuild.context.id and context.id. A bit of duplication doesn't hurt. ☺️

PR #8660, opened 2022-03-08
[Analytics] Track current user's subscription(s) at login

jakobhero:
## Description

tracks the current subscription(s) of a user at each login

## How to test

?

## Release Notes

```release-note
NONE
```

## Documentation

### DIFF
--- a/components/server/src/analytics.ts
+++ b/components/server/src/analytics.ts
@@ -6,10 +6,11 @@
 import { User } from '@gitpod/gitpod-protocol';
 import { Request } from 'express';
 import { IAnalyticsWriter } from '@gitpod/gitpod-protocol/lib/analytics';
+import { SubscriptionService } from '@gitpod/gitpod-payment-endpoint/lib/accounting';
 
-export async function trackLogin(user: User, request: Request, authHost: string, analytics: IAnalyticsWriter) {
+export async function trackLogin(user: User, request: Request, authHost: string, analytics: IAnalyticsWriter, subscriptionService: SubscriptionService) {
     //make new complete identify call for each login
-    fullIdentify(user, request, analytics);
+    fullIdentify(user, request, analytics, subscriptionService);
 
     //track the login
     analytics.track({
@@ -23,25 +24,26 @@ export async function trackLogin(user: User, request: Request, authHost: string,
 }
 
 export async function trackSignup(user: User, request: Request, analytics: IAnalyticsWriter) {
-        //make new complete identify call for each signup
-        fullIdentify(user,request,analytics);
+    //make new complete identify call for each signup
+    fullIdentify(user,request,analytics);
 
-        //track the signup
-        analytics.track({
-            userId: user.id,
-            anonymousId: stripCookie(request.cookies.ajs_anonymous_id),
-            event: "signup",
-            properties: {
-                "auth_provider": user.identities[0].authProviderId,
-                "qualified": !!request.cookies["gitpod-marketing-website-visited"]
-            }
-        });
+    //track the signup
+    analytics.track({
+        userId: user.id,
+        anonymousId: stripCookie(request.cookies.ajs_anonymous_id),
+        event: "signup",
+        properties: {
+            "auth_provider": user.identities[0].authProviderId,
+            "qualified": !!request.cookies["gitpod-marketing-website-visited"]
+        }
+    });
 }
 
-function fullIdentify(user: User, request: Request, analytics: IAnalyticsWriter) {
+function fullIdentify(user: User, request: Request, analytics: IAnalyticsWriter, subscriptionService?: SubscriptionService) {
     //makes a full identify call for authenticated users
     const coords = request.get("x-glb-client-city-lat-long")?.split(", ");
     const ip = request.get("x-forwarded-for")?.split(",")[0];
+    const subscriptions = subscriptionService ? subscriptionService.getNotYetCancelledSubscriptions(user,new Date(Date.now()).toISOString()) : undefined;
     analytics.identify({
         anonymousId: stripCookie(request.cookies.ajs_anonymous_id),
         userId:user.id,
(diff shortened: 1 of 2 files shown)


roboquat:

<details>

</details>

jankeromnes:
Reviving this PR 🧙✨🧟

jankeromnes:
Re-running Werft to pick up the additional payment config (and also because it failed):

/werft run

components/server/src/auth/login-completion-handler.ts
```diff
@@ -88,7 +88,7 @@ export class LoginCompletionHandler {
         if (authHost) {
             increaseLoginCounter("succeeded", authHost);
 
-            /** no await */ trackLogin(user, request, authHost, this.analytics).catch((err) =>
+            /** no await */ trackLogin(user, request, authHost, this.analytics, this.subscriptionService).catch((err) =>
                 log.error({ userId: user.id }, err),
```
easyCZ, line 92: [nit] Does it make sense to add more human context to the logged error? Here we might just get the user ID and the raw error, but we could say "Failed to track login: <error>"

components/server/src/analytics.ts
```diff
@@ -38,10 +46,23 @@ export async function trackSignup(user: User, request: Request, analytics: IAnal
...
     const coords = request.get("x-glb-client-city-lat-long")?.split(", ");
     const ip = request.get("x-forwarded-for")?.split(",")[0];
+    const subscriptions = subscriptionService
+        ? (
+              (await subscriptionService.getNotYetCancelledSubscriptions(
+                  user,
+                  new Date().toISOString(),
+              )) as Subscription[]
+          ).map((sub) => sub.planId)
+        : undefined;
```
easyCZ, line 65: I think here we either shift the `undefined` logic into the `getNotYetCancelledSubscriptions` method, such that it never throws and always returns either `undefined` or `Subscription[]`. That way, you can remove the logic here.

Alternatively, I'd find this slightly easier to visually grep, but that's just a suggestions and doesn't block this PR.
```typescript
var subscriptionIDs: string[] | undefined = undefined;
const subscriptions = await subscriptionService.getNotYetCancelledSubscriptions(user, new Date().toISOString());
if (subscriptions) {
	subscriptionIDs = subscriptions.map(...)
}
```

easyCZ (approved):
LGTM, approving but adding hold in case you want to update the PR with suggestions (not required).

/hold

components/server/src/auth/login-completion-handler.ts
jakobhero: hi @easyCZ, thanks for taking the time to leave these comments! great point, i added a little bit of context.

components/server/src/analytics.ts
jakobhero: thanks, your suggestion makes it much easier to follow! i used it, but had to account for cases in which planId is undefined - hope this does not destroy all of the gained clarity😃

components/server/src/analytics.ts
```diff
@@ -38,10 +45,20 @@ export async function trackSignup(user: User, request: Request, analytics: IAnal
...
+    subscriptionService?: SubscriptionService,
+) {
+    // makes a full identify call for authenticated users
     const coords = request.get("x-glb-client-city-lat-long")?.split(", ");
     const ip = request.get("x-forwarded-for")?.split(",")[0];
+    var subscriptionIDs: string[] | undefined = undefined;
+    const subscriptions = await subscriptionService?.getNotYetCancelledSubscriptions(user, new Date().toISOString());
+    if (subscriptions) {
+        subscriptionIDs = subscriptions.map((sub) => sub.planId).filter((x) => typeof x === "string") as string[];
+    }
```
easyCZ, line 61:
```suggestion
    var subscriptionIDs: string[] = [];
    const subscriptions = await subscriptionService?.getNotYetCancelledSubscriptions(user, new Date().toISOString());
    if (subscriptions) {
        subscriptionIDs = subscriptions.filter(sub => !!sub.planId).map((sub) => sub.planId!);
    }
```
We can actually simplify a bit further by making `subscriptionIDs` the empty list initially, which is a good default value when it's missing.

The actual mapping does filter first, to reduce the dataset and then does the map second to extract the IDs. It uses `!` to say it's definitely there as sugar because TS progressive type-check can't infer we've filtered out the items which are not set.

components/server/src/analytics.ts
```diff
@@ -38,10 +45,20 @@ export async function trackSignup(user: User, request: Request, analytics: IAnal
     });
 }
 
-function fullIdentify(user: User, request: Request, analytics: IAnalyticsWriter) {
-    //makes a full identify call for authenticated users
+async function fullIdentify(
+    user: User,
+    request: Request,
+    analytics: IAnalyticsWriter,
+    subscriptionService?: SubscriptionService,
```
easyCZ, line 52: I think you can safely drop the optinal on `subscriptionService` since the only caller is from `trackLogic` which as the `subscriptionService` as required.

easyCZ:
```suggestion
    subscriptionService: SubscriptionService,
```

easyCZ: This also means you can change a line below to 
```
const subscriptions = await subscriptionService.getNotYetCancelledSubscriptions(user, new Date().toISOString());
```

components/server/src/analytics.ts
jakobhero: ah yes, this is much more elegant! thanks!

easyCZ (approved):
Looks good. Approving. You may want to squash the commits in this PR, so adding a hold if you do.

/hold

components/server/src/analytics.ts
jakobhero: woops, the latest build failed as we also use `fullIdentify` in `trackSignup` and we don't have access to the subscription service from the calling context there. reverting this change and squashing the commits then!

easyCZ (approved):

Issue #8652, opened 2022-03-08
[previews/Harvester] Trigger deployment of Harvester VM before leeway build


ArthurSens:
Today, we build all Gitpod components before deploying a VM on Harvester. The build can take quite a while and starting a VM also might take quite some time.

![image](https://user-images.githubusercontent.com/24193764/157212807-2eed5574-8a95-431f-8500-0e08bb7226ad.png)

We can optimize this by triggering the VM creation before we build Gitpod components, then checking for VM readiness after everything was built

PR #8661, opened 2022-03-08
werft/deploy: Trigger VM creation before container builds

ArthurSens:

## Description

Triggers VM creation before we build our containers so it can boot in the background as we run `leeway build`.

This way we can make previews available faster 🔥 

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/8652

In the example below we made the whole pipeline 84 seconds faster
image

## Release Notes

```release-note
NONE
```


meysholdt (approved):
code changes LGTM!

Issue #8662, opened 2022-03-08
Accelerate VM start speed


meysholdt:
Analyze the time from VM-creation until the readiness probe is ready (because ssh is is-and-running):

Ideas:
* Confirm that the GCP Agent did never run. see https://github.com/gitpod-io/gitpod/pull/8639.
* run `systemd-analyze blame` so see where the time went.
* Confirm that the VM has no unneeded hardware that takes time to initialize

ArthurSens:
A couple of extra efforts are being done here:
* Require less successful probes to make VM ready https://github.com/gitpod-io/gitpod/pull/8667
* Boot VM in parallel with `leeway build`
* Issue SSL certificate in parallel with `leeway build`

meysholdt:
IMHO VM startup speed is good enough for now. The new bottlenecks are :
* https://github.com/gitpod-io/gitpod/issues/8663.
* https://github.com/gitpod-io/gitpod/pull/8690

PR #8667, opened 2022-03-08
.werft/vm/manifest: Faster boots by requiring less readiness probes

ArthurSens:
## Description

After reading [k8s docs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes):
```
successThreshold: Minimum consecutive successes for the probe to be considered successful after having failed. Defaults to 1. Must be 1 for liveness and startup Probes. Minimum value is 1.
```

It looks like we're shooting ourselves in the foot by requiring way too many successful probes before setting VMs to ready state. Since we probe every 20 seconds, this change should reduce readiness by 200~ seconds 

## Related Issue(s)

Related to https://github.com/gitpod-io/gitpod/issues/8662

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.werft/vm/manifests.ts
+++ b/.werft/vm/manifests.ts
@@ -48,7 +48,7 @@ spec:
         periodSeconds: 20
         timeoutSeconds: 10
         failureThreshold: 10
-        successThreshold: 10
+        successThreshold: 1
       domain:
         hostname: ${vmName}
         machine:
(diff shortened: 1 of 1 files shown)


wulfthimm:
I suggest to reduce the time for `initialDelaySeconds: 120` and setting a higher value for `timeoutSeconds` and `failureThreshold`.

meysholdt (approved):
Code changes LGTM!

PR #8679, opened 2022-03-09
[WIP][Installer] (in)validate the unknown fields

### LINKED ISSUE #8268: [installer] Handle additional (unknown) config values as error during config validation
Currently, the config validator checks only if necessary config values are present. It does not check, if there are additional values, that are unknown. We should treat additional / unknown config values as validation error.

**Why is this useful?**

I recently had a typo in my config file (typo in a config value name) and was wondering why the config changes didn't work. The proposed change would help in this case and warn a user that there are unknown config values in the config file.

nandajavarma:
## Description

This PR adds a tiny modification to reject the config file if there is a typo or unknown fields. This utilizes `UnmarshalStrict` to strictly parse the config. An expected extra field in the config is `apiVersion` which is being trimmed using regexp.

In addition the PR also does the following related changes:
- Remove deprecated `jaegarOperator` field from CI along
- Fix the field name of `DynamicLimits` key to match the one used in the values file
- Fix the values file to match the config structure
 
## Related Issue(s)

Fixes #8268

## Release Notes

```release-note
TODO
```


corneliusludmann (approved):
Thanks for this improvement! :rocket:

Issue #7900, opened 2022-01-28
Epic: Public Gitpod API


JanKoehnlein:
# Summary

In this epic we lay the foundation for a new public Gitpod API that can be consumed by Gitpod components as well as external clients. See this [design doc](https://www.notion.so/gitpod/Design-Doc-Public-Gitpod-API-15cdeb5d3496448ca3b61187a8aac53f) for details.

# Context

Extensively summed up in the [design doc](https://www.notion.so/gitpod/Design-Doc-Public-Gitpod-API-15cdeb5d3496448ca3b61187a8aac53f).

# Value
- provide a uniform, reliable, stable and versioned API
- consolidate our tech stack (e.g. replace JSON-RPC and REST interfaces)
- centralise common aspects such as authentication, rate-limiting, circuit-breaking etc.
- more in the [design doc](https://www.notion.so/gitpod/Design-Doc-Public-Gitpod-API-15cdeb5d3496448ca3b61187a8aac53f)

# Measurement
This epic has been successfully finished when
- the JetBrains Gateway and the dashboard successfully use the new API
- we have hardened and proven our API Guidelines
- we feel confident about our choice of technologies (gRPC, Envoy, ...)
- we are convinced the API is ready for public adoption

# Steps

## v0.1: offer basic functionality for the IDE team
- [ ] Define gRPC interface files for a workspace API
  - define `proto` files covering services to list workspaces, create and start workspace, and listen to status updates (including pre-builds and image builds)
  - gather experiences with/revise API design guidelines from the [design doc](https://www.notion.so/gitpod/Design-Doc-Public-Gitpod-API-15cdeb5d3496448ca3b61187a8aac53f)
  - define and separate re-usable concepts, e.g. for paging, structured errors
  - generate implementations in TypeScript, Go and Java
- [ ] implement the gRPC services in server 
- [ ] provide workspace API via proxy
- [ ] Introduce Envoy and plumb through proxy
   Configure
    - authorization
    - rate limiting
    - circuit breaking
    - metrics for HTTP and gRPC
    - distributed tracing

## v0.2: transition to Envoy
- [ ] Completely replace Caddy with Envoy proxy
  Migrate all the custom logic (Tiny Go/WebAssembly)

## v0.3: transition workspace startup in the dashboard
- [ ] Replace JSON-RPC with new public API in dashboard 
  Only for the workspace services implemented in v0.1

# Out of scope
- Complete migration of the current JSON-RPC API
- Define separation between public and private API

[...]

PR #8683, opened 2022-03-09
[public-api] Initial scaffolding and draft

### LINKED ISSUE #8681: Define gRPC interface files for a workspace API
- define proto files covering services to list workspaces, create and start workspace, and listen to status updates (including pre-builds and image builds)
- gather experiences with/revise API design guidelines from the [design doc](https://www.notion.so/gitpod/Design-Doc-Public-Gitpod-API-15cdeb5d3496448ca3b61187a8aac53f)
- define and separate re-usable concepts, e.g. for paging, structured errors
- generate implementations in TypeScript, Go and Java

Part of #7900

csweichel:
# Description
This PR produces an initial draft of a public API for Gitpod. We will merge this PR once we're happy with a first draft design of the workspaces side of the public API.

# Related Issue(s)

Fixes #8681
Part of #7900

# Open Questions
We have a bunch of open questions we should explore/answer as part of this PR.

## Error Handling
By default gRPC offers [error codes.](https://grpc.github.io/grpc/core/md_doc_statuscodes.html)  While the categories they introduce work well in practice, categorical errors alone (even with a message) often isn't enough.

### Errors as part of the protocol definition
Failure modes and "unhappy paths" are very APIs get interesting. Callers need to know what to expect, and which error cases to handle. If we found a way how we could communicate those failure modes as part of the API spec, we'd enable future automatic tests and better documentation.

Right now the draft uses comments to describe possible errors. We could consider using custom "option"s for this use-case. Either way, we'd have to build tooling to lint and test this stuff.

## Co-locating Workspace and Workspace Instance
In some places it's convenient to retrieve the workspace and last active instance at the same time, e.g. `ListWorkspaces`. Without this combination we're likely to see 1+n access patterns, where users `ListWorkspaces` and then `GetLastActiveInstance` for each workspace. Retrieving both at the same time can be reasonably expensive though and is not always necessary.

[`FieldMasks`](https://netflixtechblog.com/practical-api-design-at-netflix-part-1-using-protobuf-fieldmask-35cfdc606518) allow clients to specify what they expect in the response, akin to GraphQL. They are however a bit cumbersome to work with, and support for them in all languages is unclear.

## Pagination
All `List*` calls currently support pagination, following a simple pagination model: in the response we return a `next_page_token` which can be passed back into the `List*` call to advance to the next page of the paginated result set.

It's unclear if we need such capabilities today - we might just as well get away with a result limit and not be able to read "further pages".

## Linting
We'll have many authors working on this API in the future. Keeping a consistent style, and making sure we're not breaking things down the road is important. [Buf](https://buf.build/) offers such capabilities OOTB. I've added buf liniting to the code generation script. We could just as well consider making that part of our CI build. 

We will want to add means to detect breaking changes (using buf) in our CI pipeline sooner than later.

## Rich Errors
Instead of relying on the official error model, we could sport a [richer error model](https://www.grpc.io/docs/guides/error/#richer-error-model). Google APIs which rely heavily on gRPC [sport such rich error model](https://cloud.google.com/apis/design/errors#error_model). We could re-use this rich error model, which to some extend is supported by client libraries.

We will want to decide on the model early on, because introducing it later on (while maintaining backwards compatibility) would be painful.

**:+1: Pro:**
- gives us a way to detail the categorical errors and let's us introduce more fine-grained handling

**:-1: Con:**
(copied from the gRPC docs)
- Library implementations of the extended error model may not be consistent across languages in terms of requirements for and expectations of the error details payload
- Existing proxies, loggers, and other standard HTTP request processors don’t have visibility into the error details and thus wouldn’t be able to leverage them for monitoring, metrics or other purposes
- Additional error detail in the trailers interferes with head-of-line blocking, and will decrease HTTP/2 header compression efficiency due to more frequent cache misses
- Larger error detail payloads may run into protocol limits (like max headers size), effectively losing the original error

# How to test

Not much to test - more to discuss

# Release Notes

```release-note
Added first draft of a public API
```


easyCZ:
Looking pretty good! Below are couple of ideas (and reasoning for them)

## Domain model
I would recommend using the kubernetes resource style for structure the API - having a first class `spec` - this is what the user/caller specifies and configures and a `status` which the system uses to report information back. Here's an example

```protobuf
message Workspace {
   // House-keeping information, like first class ownership, project it belongs to etc..
   string id = 1;
   string owner_id = 1;
   string project_id = 2;
   Timestamp created = 1;
  
    // spec defines the requested specification of the Workspace - it is only modifiable by the caller (user/system) and epxresses the desired state of the resource
    WorkspaceSpec spec = 2;

	// status is only modifiable/settable by the system to communicate the currently observed (not necessarily the most up to date - but eventually consistent) status of the system
    WorkspaceStatus status = 3;
}

message WorkspaceSpec {
    // context_url reports the original context URL the workspace was created from
    string context_url = 4;

    // description is a human readable description of the workspace
    string description = 5;

    // Only start if set to true
    bool start = 6;

   // when set, expresses preference for regional affinity etc
   string region = 6;

   // other
   ...
}

message WorkspaceStatus {
    // The workspace instance should be what the system "assigns" to the workspace, user shouldn't set it so goes in the status
    WorkspaceInstance instance = 1;

    // Encodes the currently observed state of the Workspace - NEW, QUEUED, IMAGE_BUILDING etc...
    WorkspaceState = 2;

   ...
}
```
Following this model does a couple of things:
1. Makes it very explicit what the user can/should set
2. Delineates responsibilities for resource reconciliation in the system
3. The specification is very declarative, with the system doing magic to get to that state
4. Enables you to push configurability to the spec - the system is then responsible for making it happen - for example spec specifies `start = false` so this only creates a record but doens't actually provision resources.
5. Allows you to isolate the service from the rest of the system - largely it has most of the information it needs to operate in it's own domain model.
6. Explicitly communicates to the user that this resource is long running and that the resources progresses through states as it runs.

## RPCs
I would encourage to trim the RPC surface a bit more. Concretely:
```protobuf
service WorkspacesService {
	// Basic CRUD operations on the resource spec - this is quite easy to grasp for most people familiar with REST

	// ListWorkspaces lists workspaces. You can filter for currently running ones by expressing constraints/filters in the ListWorkspacesRequest
    rpc ListWorkspaces(ListWorkspacesRequest) returns ()
    rpc CreateWorkspace() returns ()

    // Some updates may not be possible - like moving a region within a given Workspace ID (at least for now), in that case this would be rejected with an error
    rpc UpdateWorkspace() returns ()

   // Delete should be a soft delete - mark it as "to-Delete" and the system begins termination, using status to update the consumer
    rpc DeleteWorkspace() returns ()

	// Subscribe
	rpc SubscribeToStatusUpdates(SubscribeReq) returns (WorkspaceStatus)

    rpc GetLogs(..) returns (...)
	rpc StreamLogs(...) returns (...)
}
```
Keeping the RPC surface trim eases integration - I spend less time figuring out which method to use (for example Create vs Start) and makes it easier to transactionise it.

Happy to go into more details or discuss this further. I still have limited context of the whole system so definitely looking for context and details.

csweichel:
Thank you for the feedback - much appreciated :) 

I had a read and tried to map this back. 
- spec vs status: for workspaces there isn't much of a spec that users could modify. It's all derived from the `context_url`. There is the workspace config which comes closest to the spec, and which can make sense to modify, but I've deliberately left that out of the API for now (it's not needed in 90% of use-cases).
- for workspace instances that's a slightly different story. Here a spec makes sense at start time, but modifying it at runtime doesn't. That spec would contain things like the IDE images, environment variables, or the region to start in. We could well wrap that in a Spec message.

Re CRUD/RPC: 
- I'm with you, we'll want to model things closely to known nomenclature (basically follow the principles outlined in the DD, which are CRUD). We have interesting edge cases though: StartWorkspace for example. In CRUD terms it's CreateWorkspaceInstance and then some, but with a host of constraints. StartWorkspace is much more descriptive of the operation.
- I'm a proponent of following the `rpc Foo (FooRequest) returns (FooResponse) {}` convention. If that `FooResponse` just contains a single field that's fine. Benefits: those `FooResponse` messages can grow down the line; this way we can do that and remain backwards compatible. Also, it is very regular, i.e. as a user of the API I always know how the request/response types are called. Buf nicely enforces this convention with its linter.
- we should find a naming convention for listening/subscribing to events and logs. Options are:
  1. `ListenTo...` which is the naming convention used in the old API
  2. `SubscribeTo...` which is the naming convention used in ws-manager
  3. `Get...` which would follow a more CRUD like convention, but fails to convey the streaming response nature of the call.
  
  I have no strong preference for any of those names, so long as we pick one and use it consistently.

components/public-api/gitpod/v1/workspaces.proto
```diff
@@ -0,0 +1,277 @@
...
+    string workspace_id = 1;
+
+    // owner_id is the ID of the user who created this workspace
+    string owner_id = 2;
+
+    // project_id is the ID of the project which this workspace belongs to
+    string project_id = 3;
+
+    // context_url reports the original context URL the workspace was created from
+    string context_url = 4;
```
akosyakov, line 149: Original means with prefixes? It would be nice to have context URL which can be also displayed in UI without prefixes. Dashboard right now has quite complicated logic to get such and Gateway UI we see null from time to time, since `normalizedContextUrl` is missing.

components/public-api/gitpod/v1/workspaces.proto
```diff
@@ -0,0 +1,277 @@
...
+
+package gitpod.v1;
+
+option go_package = "github.com/gitpod-io/gitpod/public-api/v1";
+
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/field_mask.proto";
+import "gitpod/v1/pagination.proto";
+
+service WorkspacesService {
```
akosyakov, line 11: How can one get an owner token?

akosyakov: I see it is from `GetActiveWorkspaceInstance` should work.

csweichel: Right now it would just be part of the workspace instance. I.e. we don't treat that token any different from the other data.

Considering our authorisation model (function scopes), we might want to reconsider that though.

components/public-api/gitpod/v1/workspaces.proto
csweichel: we could just report the `normalized_context_url`. AFAIK there isn't a use-case for the original context URL, no?

akosyakov: > we could just report the normalized_context_url.

It is a bit complicated, since not all workspaces have it and then we need to recompute normalized.

> AFAIK there isn't a use-case for the original context URL, no?

In IDE integrations I did not see any, not sure about dashboard @geropl ?

csweichel: Could we not "just" produce a normalised form for all contexts?

akosyakov:
I think it looks good for first attempt. We will need APIs around user IDE settings management, but can be added later.

components/public-api/gitpod/v1/workspaces.proto
geropl: > Could we not "just" produce a normalised form for all contexts?

And how to handle non-URl contexts like snapshots, then?

I have not find time to look at this PR in total, yet. But wrt `contextURL` what we learned in the past is that we need a clear separation between :
 - what was pasted into the browser URL field to start the workspace
 - (optional) a URL of the context, e.g. github.com/my/repo/tree/my-branch
 - a nicely displayable version of the context, e.g. "https://github.com/my/repo/tree/my-branch", or "Snapshot: xyz"

A single `string` field won't cut it I fear. Strongly voting for entity that we can extend as we evolve what a "Context" actually is.

csweichel: I've extended the `context_url` to a `context` type which can provide more details about the workspace context.

components/public-api/gitpod/v1/workspaces.proto
csweichel: I've introduced a `GetWorkspaceInstanceOwnerToken` call and removed the owner token from the workspace instance status.

csweichel:
After a discussion/conversation with @easyCZ I've added structured errors using `google.rpc.Status` to the calls, and added linting rules which enforce those for all response messages.

easyCZ:
Regarding the structured errors, my thinking has been to use a custom status represetnation rather than the google RPC. Let me illustrate in an example.
I would expect that an RPC returns a standard grpc error (possibly mappable to HTTP status codes) when the underlying request cannot be serviced, for example due to:
* Permissions
* Resource not existing
* Rate limits
* Invalid Arguments

However, for requests which are valid, for example for something like `Resource` with `ID: 123` I would expect the following response:
```proto
message ResourceSpec { 
	...
	bool delete = 10;
 }

message ResourceStatus {
  // state is a machine readable high level state of the resource - last observed state by the controller
  string state = 1; // NEW, QUEUED, RUNNING, DELETING, COMPLETE, ERROR
  string details = 2;  // human redable details of each status - when state == ERROR, this can provide hints what to do to fix it
  ...
}

message Response {
  ResourceSpec spec = 1;
  ResourceStatus status = 2;
}
```
All valid requests return this as the underlying Resource - that is Get, Create, Delete, Update and List returns a repeated of this.
So now for an example:
`GetResource(123)` -> Returns OK, even when the Resource is in an ErrorState
`DeleteResource(999)` -> Returns NotFound
`DeleteResource(123)` - > Returns OK, `{ spec: { delete = true, ...}, status: { state: DELETING }`

So instead of using the default `google.rpc.Status`, we define whichever states make sense for our particular resource. Just like k8s Pods define their statuses.
Of course, we can use `oneof` definitions to better encapsulate an ErrorState with extra messaging etc.
I think using `google.rpc.Status` constraints us too much with respect to encoding the relevant domain of gitpod into the API.

components/public-api/README.md
```diff
@@ -0,0 +1,101 @@
...
+})
+```
+
+### Prebuild aware but no log support
+```go
+workspaces.CreateWorkspace({
+    idempotency_token: "<random_string>",
+    context_url: "https://github.com/gitpod-io/gitpod",
+    prebuild: {
+        if_available: true,
```
svenefftinge, line 22: this should be the default behavior.

components/public-api/README.md
```diff
@@ -0,0 +1,101 @@
...
+    }
+})
+```
+
+### Prebuild aware with log support
+```go
+contextURL := "https://github.com/gitpod-io/gitpod"
+prb := prebuilds.GetRunningPrebuild({ context_url: contextURL })
```
svenefftinge, line 30: I believe starting prebuilds and starting workspaces have so much in common, that we should rather generalize over them and make 'prebuild' a property on the workspaces API rather than duplicating everything in a prebuilds API.

components/public-api/README.md
```diff
@@ -0,0 +1,101 @@
...
+logs := prebuilds.ListenToPrebuild({ context_url: contextURL, prebuild_id: prb.PrebuildID })
```
svenefftinge, line 31: Is `context_url` a leftover or why would I provide it in addition to the prebuild id?

components/public-api/README.md
svenefftinge: At least for listening to logs and status updates, we could use a prebuild's workspace instance. Also there might be an image build ongoing.

components/public-api/gitpod/v1/workspaces.proto
```diff
@@ -0,0 +1,328 @@
...
+    string idempotency_token = 1;
+
+    string context_url = 2;
+
+    oneof prebuild {
+        // if true, and there's a prebuild available (not running), use that prebuild
+        bool if_available = 3;
+        // uses a particular prebuild - the prebuild ID is scoped by the context URL.
+        // Use the PrebuildService to get ahold of the prebuild_id.
+        string prebuild_id = 4;
```
svenefftinge, line 87: Why would I expicitly pass a prebuild_id?

components/public-api/gitpod/v1/workspaces.proto
```diff
@@ -0,0 +1,328 @@
...
+}
+
+message CreateAndStartWorkspaceRequest {
+    string idempotency_token = 1;
+
+    string context_url = 2;
+
+    oneof prebuild {
+        // if true, and there's a prebuild available (not running), use that prebuild
+        bool if_available = 3;
```
svenefftinge, line 84: Using an existing prebuild should be the default and I wonder if forcing "no prebuild" is important enough to be included in such an early draft.

components/public-api/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+# Example Flows
+This section gives examples how clients would use this API.
+
+## Create Workspace
+
+### Not prebuild aware
+```go
+workspaces.CreateWorkspace({
+    idempotency_token: "<random_string>",
```
jankeromnes, line 11: Small nit: We could call this `request_id` in order to use shorter & less complicated words on every call. 💡 (Advantage: More easily understood by all users, especially non-native speakers.)

components/public-api/gitpod/v1/workspaces.proto
csweichel: That's a fair point - do we even have a case where we want to ignore an existing available prebuild? Arguably, whenever there's a prebuild available we want to use it. Not doing so does feel like an edge case.

csweichel: will update this accordingly

components/public-api/gitpod/v1/workspaces.proto
csweichel: Today's `createWorkspace` call is rather complicated because it mixes prebuild selection and workspace creation to provide a form of "transactional behaviour". By making prebuilds an explicit conern this becomes a lot easier, and we can factor prebuild selection out of workspace creation.

components/public-api/README.md
csweichel: Good point you're raising here. There are several choices we have:
1. `idempotency token` or `key`. This accurately describes what's happening. Is used by [IETF](https://tools.ietf.org/id/draft-idempotency-header-01.html#name-the-idempotency-http-reques) and [Stripe](https://stripe.com/docs/api/idempotent_requests). Is long and fancy, hence maybe harder to understand.
2. `client token` is used by AWS. Downside: doesn't relate to the request.
3. `request_id` is short and not fancy (that's good). But it's not self-explanatory either: users would need to read the docs to understand what this field is for and how to use it.

easyCZ: `request_id` is often used to mean a trace id. Imo, idempotency token is the most explicit. Users will consume this from sdks and will therefore also see the doc string on this field. That should provide enough clarity (together with website examples)

csweichel: Let's stick to idempotency_token for the reasons above.

components/public-api/gitpod/v1/workspaces.proto
csweichel: In the original design the prebuild_id would be matched against the context URL (basically scoped by the context URL). That feels very complicated. 

In the most recent draft the "source" of the workspace is a oneof `context_url` or `prebuild_id`.

easyCZ: _I think the Workspace APIs should NOT accept a prebuild ID_. Here's why:

The API shouldn't necessarily encode how things work today, but much more the experience we want a user to have. We can use the "time to first 200" as a metric for the API UX.

With this in mind, we should eliminate ambiguity and concepts from the API calls which are not needed, or which the system can compute/discover for itself. In this case, I believe the system should be able to lookup if a prebuild for a context exists, and use it when it does - without the explicitly defining the prebuild ID. 

As a user, it's easier for me to think of workspaces in terms of contexts, and not prebuilds. This means that it will be easier for me to only supply the context and not the prebuild id when I'm integrating. This in turn reflects badly on us because the user is not getting the benefits of prebuilds.

geropl:
wrt to "encoding errors in the type system": Did we explore encoding error messages as return types? Like, treating protobuf as if it would support algebraic data types? as an example `GetWorkspaceResponse` could become:
```
message GetWorkspaceResponse {
    oneof response {
        Workspace result = 1;
        NotFoundError not_found = 2;
        PermissionError permission_error = 3;
    }
}

message NotFoundError {
    string resource_type = 1;
    string id = 2;
}

message PermissionError {
    string resource_type = 1;
    string id = 2;
    string message = 3;
}
```
On client side we would always have to explicitly check all errors before accessing the "actual" response. And that would have to also be checked to be present for forward compatibility: in case we added an error later that means old definitions have all fields empty. IMO that nicely encodes the "unknown error" case and establishes good client-side code patterns. :upside_down_face: 

What I like as well is that we clearly separate "domain errors" ("the id you requests does not return a result") from the generic protobuf errors which are more geared towards service availability / a lower abstraction layer.

components/public-api/gitpod/v1/workspaces.proto
```diff
@@ -0,0 +1,322 @@
...
+    string workspace_id = 1;
+}
+message GetWorkspaceResponse {
+    google.rpc.Status response_status = 1;
+
+    Workspace result = 2;
+}
+
+message CreateAndStartWorkspaceRequest {
+    string idempotency_token = 1;
```
geropl, line 78: :heart: THE single most important change that bring immediate benefit. Maybe I'm exaggerating... but maybe not. :stuck_out_tongue:

components/public-api/gitpod/v1/workspaces.proto
```diff
@@ -0,0 +1,322 @@
...
+
+    oneof source {
+        string context_url = 2;
+        string prebuild_id = 3;
```
geropl, line 82: Hm, this feels sligthly odd... :thinking: cannot pinpoint, though. Have to sleep over it.

AlexTugarev: Why so? 
As we now expose the prebuild ids on the dashboard, it would be nice to actually make use of them instead of going though all of the indirections.

components/public-api/gitpod/v1/workspaces.proto
```diff
@@ -0,0 +1,322 @@
...
+
+    // description is a human readable description of the workspace
+    string description = 5;
+}
+
+// WorkspaceContext describes the context a workspace was created from
+message WorkspaceContext {
+    // Explicit Git context
+    message Git {
+        string normalized_context_url = 1;
```
AlexTugarev, line 177: what is the normalized context url? 
let's use the clone url instead.
I just found out, that some providers might use different url patterns in comparison to the context url, so you cannot just compute the clone url from it, instead it's some result from external api services which we need to store.

geropl (approved):

aledbf (approved):

PR #8685, opened 2022-03-09
Remove Theia extension related code

### LINKED ISSUE #4630: remove Theia and move extension resolution to Code

mustard-mh:
## Description

Remove Theia extension related code

## Related Issue(s)

Fixes #4630

## How to test

Start A workspace to do the smoke test like release do https://github.com/gitpod-io/gitpod/pull/8609 
I have did this all (although I think we don't need to do this many steps 😄)
- Start a workspace.
- Test following:
  - [x] terminals are preserved and resized properly between window reloads
  - [x] WebViews are working
  - [x] extension host process: check language smartness and debugging 
  - [x] extension management (installing/uninstalling)
  - [x] install the [VIM extension](https://open-vsx.org/extension/vscodevim/vim) to test web extensions
  - that user data are sync across workspaces as well as on workspace restart, especially for extensions
   - [x] extensions from .gitpod.yml are not installed as sync
   - [x] extensions installed as sync are actually sync in new workspaces
  - [x] settings should not contain any mentions of MS telemetry
  - websockets and workers are properly proxied
    - [x] diff editor should be operatable
    - [x] trigger reconnection with `window.WebSocket.disconnectWorkspace()`, check that old websockets are closed and new opened of the same amount
  - [x] workspace specific commands should work, i.e. F1 → type Gitpod prefix
  - [x] that a PR view is preloaded on the PR url
  - [x] test gp open and preview
  - [x] test open in VS Code Desktop, check gp open and preview in task/user terminals
  - [ ] ~~telemetry data is collected in [Segment](https://app.segment.com/gitpod/sources/staging_untrusted/debugger)~~

## Release Notes

```release-note
NONE
```


components/gitpod-protocol/BUILD.yaml
```diff
@@ -17,13 +17,6 @@ packages:
       packaging: library
       yarnLock: ${coreYarnLockBase}/yarn.lock
       tsconfig: tsconfig.json
-  - name: theia-plugin-list
```
mustard-mh, line 20: global searched `theia-plugin-list` -> no usage

components/server/src/theia-plugin/theia-plugin-service.ts
```diff
@@ -4,28 +4,18 @@
...
 
 const userPluginsUri = 'user-plugins://';
 
-export interface ResolvedPluginsResult {
-    resolved: ResolvedPlugins
-    external: string[]
-}
-
 @injectable()
 export class TheiaPluginService {
```
mustard-mh, line 20: No idea other functions should be deleted or not. Should we change the whole class name and file name/path to something not containing `theia`?

akosyakov: Could you list places where it is still used?

We still have data for user settings/extensions which should be migrated to VS Code, so if they are accessed from user data sync then we should keep it. But there should not be way to generate new records, so most likely most can be removed.

mustard-mh: Only here use `theia-plugin-service` now

https://github.com/gitpod-io/gitpod/blob/0d05dcf813d70d5ae4678e69f61b7bd12ad1ffe8/components/server/src/code-sync/code-sync-service.ts#L176

akosyakov:
@mustard-mh Could you provide `How to test` section? We should smoke test that workspaces still start and extensions are there. Do you do any other testing?

components/server/src/theia-plugin/theia-plugin-service.ts
akosyakov: maybe can be inlined into code-sync endpoint to get rid of this file completely?

mustard-mh:
How to test section updated ✅

felladrin (approved):
Code reviewed and also tested with a production app that uses most of the features mentioned in the How To Test section. Everything's working great! ✅ 

I also verified each checklist item in the How To Test section.

andreafalzetti:
Do we still need this file? https://github.com/gitpod-io/gitpod/blob/hw/fix-4630/components/gitpod-cli/generate-theia-protocol.go

andreafalzetti:
Do we still need this file? https://github.com/gitpod-io/gitpod/blob/hw/fix-4630/components/gitpod-cli/generate-theia-protocol.go

geropl (approved):
LGTM

mustard-mh:
> Shall we also remove this file? [components/gitpod-cli/generate-theia-protocol.go](https://github.com/gitpod-io/gitpod/blob/hw/fix-4630/components/gitpod-cli/generate-theia-protocol.go)

@andreafalzetti  maybe other PR to remove theia things

PR #8689, opened 2022-03-09
[kots]: create dev channels in KOTS and formalise the release process

### LINKED ISSUE #8688: Enable users to push to their own KOTS channel for development purposes
Currently, the `make` command pushes everything to `Unstable` channel. This should change so that it pushes to a channel per-developer. This would avoid clashes when multiple people developing the KOTS ecosystem

MrSimonEmms:
## Description

Formalise the development and release process in the KOTS ecosystem. This came as a result of a conversation with @Pothulapati where it was clear a lot of the process was in my head rather than formally documented

## Related Issue(s)

Fixes #8688

## How to test

Follow readme

## Release Notes

```release-note
[kots]: create dev channels in KOTS and formalise the release process
```

### DIFF
--- a/install/kots/README.md
+++ b/install/kots/README.md
@@ -20,12 +20,15 @@ commercially supported project.
 
 # Development
 
+> tl;dr the `make` command will build and push a release to your development channel
+
 ## Authentication
 
-Two environment variables are required to be able to publish to our Replicated account:
+The following environment variables are required to be able to publish to our Replicated account:
 
- - `REPLICATED_APP`: the unique application slug
+ - `REPLICATED_APP`: the unique application slug. If in doubt, use `gitpod-pov`.
  - `REPLICATED_API_TOKEN`: a [User API Token](https://vendor.replicated.com/account-settings) with `Read/Write` permissions
+ - `REPLICATED_DEV_CHANNEL`: the channel to push dev releases to. Use the naming convention `dev-<initials>` (eg, `dev-sje`).
 
 ## Naming conventions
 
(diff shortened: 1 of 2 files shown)


MrSimonEmms:
**NB** I have kept the envvars with `REPLICATED` rather `KOTS` convention because this is relevant to the paid-for ecosystem, rather than the KOTS open source package

install/kots/README.md
```diff
@@ -20,12 +20,15 @@ commercially supported project.
...
 # Development
 
+> tl;dr the `make` command will build and push a release to your development channel
+
 ## Authentication
 
-Two environment variables are required to be able to publish to our Replicated account:
+The following environment variables are required to be able to publish to our Replicated account:
 
  - `REPLICATED_APP`: the unique application slug
```
corneliusludmann, line 29: We still use `gitpod-pov`, right? Eventually, we will use `gitpod` for prod but still keep a different dev app slug, right? Should we mention the app slugs in use here?

By the way: What do you think of renaming `gitpod-pov` to `gitpod-dev` (or create if renaming does not work)?

corneliusludmann (approved):
/hold

looks good to me, adding hold in case @Pothulapati will check this as well or if you would like to add something for https://github.com/gitpod-io/gitpod/pull/8689#discussion_r822702800

install/kots/README.md
MrSimonEmms: Yes, we are using `gitpod-pov` which is what's defined by the `REPLICATED_APP` value. I'm not against documenting in there, but I don't think it's necessary

I can see why `gitpod-dev` would be preferable, but I'm not that bothered by renaming it. If the majority want to rename it, that's fine, but it's effectively our internal dev service so it's fine as is.

I think we'll likely abandon `gitpod-pov` in future and just use `gitpod` but that's a future discussion

Pothulapati (dismissed):
Change Looks good to me! Thanks for taking the time @MrSimonEmms

install/kots/README.md
corneliusludmann: > I'm not against documenting in there, but I don't think it's necessary

How does a new colleague know which value to set here? Maybee, we need also a link to the vendor portal where an API key can be generated. Would make the onboarding of new folks easier when we have this written down somewhere and why not here in the readme?

corneliusludmann:
@MrSimonEmms: Installing preview env fails due to [exceeded quota](https://gitpod.slack.com/archives/C032A46PWR0/p1646839029425789?thread_ts=1646837123.545349&cid=C032A46PWR0).

Why not using … ?

/werft run no-preview

_(since this change has no effect on the preview env (does it?) a preview env would indeed unnecessary)_

MrSimonEmms:
(I'm tired and just want Werft to work 😄 )

install/kots/README.md
corneliusludmann: > Ok. I've changed the wording slightly.
> 
> The vendor portal is linked in the `REPLICATED_API_TOKEN`. I'm reluctant to duplicate too much of their development readme in ours as it means we then have to maintain that as well as our own application.

Awesome! Thanks a lot! :+1:

PR #8690, opened 2022-03-09
werft/deploy: Issue certificates before container builds 

ArthurSens:
## Description

Using the same strategy as https://github.com/gitpod-io/gitpod/pull/8661

This PR triggers the creation of Let's Encrypt certificates before we build all containers with Leeway, so they can be processed in parallel. Further down during the deployment, we check for Certificate readiness before proceeding to install Gitpod in our previews

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.werft/util/certs.ts
+++ b/.werft/util/certs.ts
@@ -1,7 +1,7 @@
 import { exec, ExecOptions } from './shell';
 import { sleep } from './util';
-import { readFileSync, writeFileSync } from 'fs';
 import * as path from 'path';
+import { getGlobalWerftInstance, Werft } from './werft';
 
 
 export class IssueCertificateParams {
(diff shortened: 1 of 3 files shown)


ArthurSens:
Converting to draft because it is failing for harvester previews

.werft/util/certs.ts
```diff
@@ -25,49 +25,47 @@ export class InstallCertificateParams {
...
+    werft.log(shellOpts.slice, "Kubectl command for cert creation: " + cmd)
+    exec(cmd, { ...shellOpts, slice: shellOpts.slice });
 }
 
-
 export async function installCertficate(werft, params: InstallCertificateParams, shellOpts: ExecOptions) {
     let notReadyYet = true;
-    werft.log('certificate', `copying certificate from "${params.certNamespace}/${params.certName}" to "${params.destinationNamespace}/${params.certSecretName}"`);
+    werft.log(shellOpts.slice, `copying certificate from "${params.certNamespace}/${params.certName}" to "${params.destinationNamespace}/${params.certSecretName}"`);
     const cmd = `kubectl get secret ${params.certName} --namespace=${params.certNamespace} -o yaml \
```
meysholdt, line 69: I assume this uses the wrong kubectl context by default: it uses k3s but it should get the cert from core-dev.

ArthurSens: it depends on when you're calling this method...

The k3s kubeconfig is never downloaded if deploying without the `with-vm` annotation isn't passed, so the certificates are installed on the correct place

PR #8696, opened 2022-03-09
Make waiting for pods more robust

meysholdt:
## Description
The new `kubectl wait --for=condition=Ready pods --all -n kube-system --timeout=300s` fails with `error: no matching resources found` :astonished:
The api-server is already ready, but that does not seem to imply that there are pods in kube-system. 

This PR fixes the error by introducing more robust logic for waiting.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.werft/jobs/build/deploy-to-preview-environment.ts
+++ b/.werft/jobs/build/deploy-to-preview-environment.ts
@@ -3,7 +3,7 @@ import * as shell from 'shelljs';
 import * as fs from 'fs';
 import { exec, ExecOptions } from '../../util/shell';
 import { InstallMonitoringSatelliteParams, installMonitoringSatellite } from '../../observability/monitoring-satellite';
-import { wipeAndRecreateNamespace, setKubectlContextNamespace, deleteNonNamespaceObjects, findFreeHostPorts, createNamespace, helmInstallName, findLastHostPort } from '../../util/kubectl';
+import { wipeAndRecreateNamespace, setKubectlContextNamespace, deleteNonNamespaceObjects, findFreeHostPorts, createNamespace, helmInstallName, findLastHostPort, waitUntilAllPodsAreReady, waitForApiserver } from '../../util/kubectl';
 import { issueCertficate, installCertficate, IssueCertificateParams, InstallCertificateParams } from '../../util/certs';
 import { sleep, env } from '../../util/util';
 import { GCLOUD_SERVICE_ACCOUNT_PATH } from "./const";
@@ -43,6 +43,8 @@ const vmSlices = {
     COPY_CERT_MANAGER_RESOURCES: 'Copy CertManager resources from core-dev',
     INSTALL_LETS_ENCRYPT_ISSUER: 'Install Lets Encrypt issuer',
     KUBECONFIG: 'Getting kubeconfig',
+    WAIT_K3S: 'Waiting for k3s',
+    WAIT_CERTMANAGER: 'Waiting for Cert-Manager',
     EXTERNAL_LOGGING: 'Install credentials to send logs from fluent-bit to GCP'
 }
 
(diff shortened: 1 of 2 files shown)


meysholdt:
Build successful: https://werft.gitpod-dev.com/job/gitpod-build-me-robustwaiting.20/logs

Note the two new slices:
image

.werft/util/kubectl.ts
```diff
@@ -240,3 +241,62 @@ export function findFreeHostPorts(ranges: PortRange[], shellOpts: ExecOptions):
...
     exec(`kubectl rollout status ${type} ${name} -n ${namespace}`, shellOpts);
 }
+
+export async function waitUntilAllPodsAreReady(namespace: string, shellOpts: ExecOptions) {
+    interface Pod {
+        name: string
+        owner: string
+        phase: string
+    }
+    const werft = getGlobalWerftInstance();
```
ArthurSens, line 251:
```suggestion
```

I've noticed that the trace result is broken: https://ui.honeycomb.io/gitpod/datasets/werft/trace/oH1d5CeEh9E?span=55953ea97f9ec6b5

Not sure about the reason, but could it be because we're not passing it as a parameter for this function and also for `waitForApiserver`? cc @mads-hartmann

.werft/util/kubectl.ts
```diff
@@ -240,3 +241,62 @@ export function findFreeHostPorts(ranges: PortRange[], shellOpts: ExecOptions):
...
+    }
+    exec(`kubectl get pods -n ${namespace}`, { ...shellOpts, async: false })
+    throw new Error(`Not all pods in namespace ${namespace} transitioned to 'Running' or 'Succeeded/Completed' during the expected time.`)
+}
+
+export async function waitForApiserver(shellOpts: ExecOptions) {
+    const werft = getGlobalWerftInstance();
+    for (let i = 0; i < 300; i++) {
+        werft.log(shellOpts.slice, 'Checking that k3s apiserver is ready...')
+        const result = (exec(`kubectl get --raw='/readyz?verbose'`, { ...shellOpts, dontCheckRc: true, async: false }));
```
ArthurSens, line 294:
```suggestion
        const result = exec(`kubectl get --raw='/readyz?verbose'`, { ...shellOpts, dontCheckRc: true, async: false });
```

ArthurSens (commented):
Just a quick look while still waking up 😅 

I've got to do something and will take another look when I start working tooday

.werft/util/kubectl.ts
meysholdt: > could it be because we're not passing it as a parameter

not sure I understand what should be passed as a parameter here. Do you have an example? My understanding is that `werft` is a global static singleton and therefore does not need to be passed as a parameter.

mads-hartmann: The `werft` instance is a singleton at the moment so passing it as parameter or fetching the global instance should be the same (I don't see any race conditions at a first glance at least). It is weird that your slices are orphaned and the `Ensuring VM is ready for deployment` phase doesn't show up in Honeycomb - so something is wrong in the instrumentation.

.werft/jobs/build/deploy-to-preview-environment.ts
```diff
@@ -454,16 +447,8 @@ async function deployToDevWithInstaller(werft: Werft, jobConfig: JobConfig, depl
...
-        werft.log(installerSlices.DEPLOYMENT_WAITING, "Server not ready. Let the waiting...commence!");
-        exec(`kubectl -n ${namespace} rollout status deployment/server --timeout=10m`, { slice: installerSlices.DEPLOYMENT_WAITING });
-        werft.done(installerSlices.DEPLOYMENT_WAITING);
-    } catch (err) {
-        if (!jobConfig.mainBuild) {
-            werft.fail(installerSlices.DEPLOYMENT_WAITING, err);
-        }
-        exec('exit 0')
-    }
+    await waitUntilAllPodsAreReady(deploymentConfig.namespace, { slice: installerSlices.DEPLOYMENT_WAITING })
```
mads-hartmann, line 450: Maybe you're missing a `werft.log(vmSlices.DEPLOYMENT_WAITING, 'Waiting for deployment')` or something to that effect here.

In any case, Werft is displaying the phases/logs correctly, so could be that the instrumentation in the Werft class isn't following excatly the semantics that Werft has for when to close phases.

meysholdt: I've added that line.

meysholdt:
merged via https://github.com/gitpod-io/gitpod/pull/8706

PR #8713, opened 2022-03-10
[kots]: switch azure cluster issuer to issuer

MrSimonEmms:
## Description

Related to a [cert-manager issue](https://github.com/cert-manager/cert-manager/issues/4867), in order for Azure issuers to use the ambient credentials an additional argument is required to be passed in.

This changes the Azure `ClusterIssuer` to an `Issuer`

## How to test

Deploy to Azure via KOTS. Search for:

```shell
kubectl get deployments.apps -n gitpod cert-manager -o yaml | grep '\-\-issuer-ambient-credentials=true'
```

## Release Notes

```release-note
[kots]: switch azure cluster issuer to issuer
```


Pothulapati:
Closing this as we discussed it was better to remove it, and also raised https://github.com/gitpod-io/gitpod/issues/8749

Issue #8727, opened 2022-03-10
Traces for Werft jobs are missing phases and spans


meysholdt:
## Is your feature request related to a problem? Please describe
Traces are missing phases and spans. Example: 
https://ui.honeycomb.io/gitpod/datasets/werft/trace/6oNaVv7xGTR?span=missing%20root
Related discussion: https://gitpod.slack.com/archives/C01KGM9EBD4/p1646904136525909
We observed the errors on https://github.com/gitpod-io/gitpod/pull/8696. That PR got superseded by https://github.com/gitpod-io/gitpod/pull/8706 on which tracing was working fine and thus we merged it.
Now, however, we see broken traces on `main`.

It's not clear yet what causes this and I filed https://github.com/gitpod-io/gitpod/issues/8707 to rule out potential causes. 

## Describe the behaviour you'd like
Traces without missing spans/phases.

mads-hartmann:
I haven't been able to reproduce this since https://github.com/gitpod-io/gitpod/pull/8743 so I'm closing this for now. We an reopen it if we see it again.

PR #8743, opened 2022-03-11
Werft tracing improvements

mads-hartmann:
## Description

This moves the graceful shutdown logic to a "finally" block and introduces separate phases for the shutdown logic. Previously the shutdown logs would just to to the last open phase which was a bit confusing.

With this change I haven't been able to reproduce https://github.com/gitpod-io/gitpod/issues/8727 but I don't see how this change should fix it. However, as this is an improvement I'd like to get it into main and then see if I can reproduce #8727 off a new branch.

## Related Issue(s)

Part of #8727 but suspect not the solution.

## How to test

I was able to reproduce the #8727 error here ([job 3](https://werft.gitpod-dev.com/job/gitpod-build-mads-werft-tracing.3/logs), [trace](https://ui.honeycomb.io/gitpod/datasets/werft/trace?trace_id=8ee830780829772e20719b4d4a6818c4)) but never again after introducing the change ([job 4](https://werft.gitpod-dev.com/job/gitpod-build-mads-werft-tracing.4/results), [trace](https://ui.honeycomb.io/gitpod/datasets/werft/trace/AsLa8F7W7gZ?span=bfa0f8ebc265cc68)). And that's even when I tried clean-slate-deployment ([job 6](https://werft.gitpod-dev.com/job/gitpod-build-mads-werft-tracing.6/results), [trace](https://ui.honeycomb.io/gitpod/datasets/werft/trace/dEe5XWgpoiC?span=c105aa50a6df3f81)). Even so, I don't this change is the fix and suspect there's some other race condition or similar that I just didn't hit.

## Release Notes

```release-note
NONE
```

## Documentation

N/A

### DIFF
--- a/.werft/build.ts
+++ b/.werft/build.ts
@@ -21,14 +21,11 @@ Tracing.initialize()
         werft = new Werft("build")
     })
     .then(() => run(context))
-    .then(() => VM.stopKubectlPortForwards())
-    .then(() => werft.endAllSpans())
     .catch((err) => {
         werft.rootSpan.setStatus({
             code: SpanStatusCode.ERROR,
             message: err
         })
-        werft.endAllSpans()
 
         if (context.Repository.ref === "refs/heads/main") {
             reportBuildFailureInSlack(context, err, () => process.exit(1));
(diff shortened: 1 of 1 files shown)


meysholdt (approved):
Code changes LGTM!

PR #8748, opened 2022-03-11
Fixing the clean-slate-deployment of new VM base preview-environments

### LINKED ISSUE #8747: VMs are not removed during a clean-slate-deployment
### Bug description

I started a preview-environment in a VM and tried a `clean-slate-deployment`. The VM has not been removed before the deployment of Gitpod.

### Steps to reproduce

Open a new workspace and run `werft run github -a with-vm=true` which will trigger the deployment of Gitpod on a VM. After the deployment has succeeded run `werft run github -a with-vm=true -a with-clean-slate-deployment` and follow the logs in werft.gitpod-dev.com to check if the VM is removed in the `prepare` phase.

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

wulfthimm:
## Description

This fixes a problem where VMs are not removed during a `clean-slate-deployment`.

UPDATE: Thanks for the support. I dug deeper and refactored a bit. The `prepareVM` function is now only used when we really want a VM. Then it checks in `prepareVM` if the the VM is present. If it is not present `createVM` otherwise the VM exists and we need to check if a `clean-slate-deployment` is triggered. Then the VM which existence has been verified in the previous step will be deleted and recreated.

## Related Issue(s)

Fixes #8747

## How to test

Start a workspace and run `werft run github -a with-vm=true`. After the deployment of the VM run `werft run github -a with-vm=true -a with-clean-slate-deployment`. The VM will be removed during the `prepare` phase and watch out for the string "Cleaning previously created VM" in the logs. An additional test would be to run `werft run github -a with-vm=true` and make sure that the VM will not be removed.

## Release Notes

```release-note
None
```

### DIFF
--- a/.werft/jobs/build/prepare.ts
+++ b/.werft/jobs/build/prepare.ts
@@ -63,26 +63,30 @@ function configureCoreDevAccess() {
 }
 
 function decideHarvesterVMCreation(werft: Werft, config: JobConfig) {
-    if (config.withVM && !VM.vmExists({ name: config.previewEnvironment.destname })) {
-        prepareVM(werft, config)
+    if (shouldCreateVM(config)) {
+        createVM(werft, config)
     } else {
         werft.currentPhaseSpan.setAttribute("werft.harvester.created_vm", false)
     }
     werft.done(prepareSlices.BOOT_VM)
 }
 
-function prepareVM(werft: Werft, config: JobConfig) {
-    if (config.cleanSlateDeployment) {
-        werft.log(prepareSlices.BOOT_VM, "Cleaning previously created VM")
-        VM.deleteVM({ name: config.previewEnvironment.destname })
-    }
-    createVM(werft, config, prepareSlices.BOOT_VM)
+function shouldCreateVM(config: JobConfig) {
+    return config.withVM && (
+        !VM.vmExists({ name: config.previewEnvironment.destname }) ||
+        config.cleanSlateDeployment
+    )
 }
 
 // createVM only triggers the VM creation.
 // Readiness is not guaranted.
-function createVM(werft: Werft, config: JobConfig, slice: string) {
-    werft.log(slice, 'Booting  VM')
+function createVM(werft: Werft, config: JobConfig) {
+    if (config.cleanSlateDeployment) {
+        werft.log(prepareSlices.BOOT_VM, "Cleaning previously created VM")
+        VM.deleteVM({ name: config.previewEnvironment.destname })
+    }
+
+    werft.log(prepareSlices.BOOT_VM, 'Creating  VM')
     VM.startVM({ name: config.previewEnvironment.destname })
     werft.currentPhaseSpan.setAttribute("werft.harvester.created_vm", true)
 }


roboquat:
@wulfthimm: Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://git.k8s.io/community/contributors/guide/release-notes.md) to remove it.

<details>

</details>

.werft/vm/vm.ts
```diff
@@ -114,8 +114,7 @@ export function startVM(options: { name: string }) {
  * @returns true if the VM already exists
  */
 export function vmExists(options: { name: string }) {
```
wulfthimm, line 116: I would expect that `name` is the `destname` which will be without the leading `preview-`. What am I missing here?

wulfthimm: Is it worth digging deeper?

mads-hartmann (changes requested):
The changes don't look right to me. You're passing in the namespace and then also appending `preview` in the `-n` argument to kubectl.

The code looks right before though, so I suspect whatever issue you're looking for is hiding somewhere else ☺️

mads-hartmann:
@wulfthimm I suspect the real problem is that we only invoke `prepareVM` if the VM doesn't exist. And the logic to delete the VM is inside prepareVM ☺️

wulfthimm:
> The changes don't look right to me. You're passing in the namespace and then also appending `preview` in the `-n` argument to kubectl.
> 
> The code looks right before though, so I suspect whatever issue you're looking for is hiding somewhere else relaxed

That is what I also suspect. I started with this approach and the job did what I expected to do. Therefore this premature PR. But I will take a deeper look to ensure that the solution is sustainable.

wulfthimm:
> @wulfthimm I suspect the real problem is that we only invoke `prepareVM` if the VM doesn't exist. And the logic to delete the VM is inside prepareVM relaxed

Having the deletion of the previous VM inside `prepareVM` is valid as we check for a `cleanSlateDeployment`. The problem is the function `existsVM` that return a `false` when it should return a `true`.

mads-hartmann:
> The problem is the function existsVM that return a false when it should return a true.

Maybe we're talking past each other. Your changes has introduced a bug which means `VM.vmExists` will always return false, but was that also the case on main?

If we're looking at main there's a bug even if `VM.vmExists` does the right thing. If VM.vmExists return true we will never invoke prepareVM because of [this conditional](https://github.com/gitpod-io/gitpod/blob/main/.werft/jobs/build/prepare.ts#L66). Because we never invoke prepareVM, we will never try to delete the VM.

I think the best way forward is to

1. Revert your changes
2. Ensure VM.vmExists does the right thing
3. Fix the issue I described above

wulfthimm:
> > The problem is the function existsVM that return a false when it should return a true.
> 
> Maybe we're talking past each other. Your changes has introduced a bug which means `VM.vmExists` will always return false, but was that also the case on main?
> 
> If we're looking at main there's a bug even if `VM.vmExists` does the right thing. If VM.vmExists return true we will never invoke prepareVM because of [this conditional](https://github.com/gitpod-io/gitpod/blob/main/.werft/jobs/build/prepare.ts#L66). Because we never invoke prepareVM, we will never try to delete the VM.
> 
> I think the best way forward is to
> 
> 1. Revert your changes
> 2. Ensure VM.vmExists does the right thing
> 3. Fix the issue I described above

.werft/vm/vm.ts
```diff
@@ -114,8 +114,7 @@ export function startVM(options: { name: string }) {
...
-    const namespace = `preview-${options.name}`
-    const status = exec(`kubectl --kubeconfig ${KUBECONFIG_PATH} -n ${namespace} get vmi ${options.name}`, { dontCheckRc: true, silent: true })
+    const status = exec(`kubectl --kubeconfig ${KUBECONFIG_PATH} -n preview-${options.name} get vmi ${options.name}`, { dontCheckRc: true, silent: true })
```
ArthurSens, line 117: I believe you want to revert this :)

.werft/jobs/build/prepare.ts
```diff
@@ -62,21 +63,15 @@ function configureCoreDevAccess() {
...
 
-function decideHarvesterVMCreation(werft: Werft, config: JobConfig) {
-    if (config.withVM && !VM.vmExists({ name: config.previewEnvironment.destname })) {
-        prepareVM(werft, config)
-    } else {
-        werft.currentPhaseSpan.setAttribute("werft.harvester.created_vm", false)
-    }
-    werft.done(prepareSlices.BOOT_VM)
-}
-
```
ArthurSens, line 73: Any reason for this function to be deleted?

If you delete it, you're not closing the "BOOT_VM" slice anymore, while also not setting the `werft.harvester.created_vm` span attribute to false

ArthurSens (changes requested):

.werft/jobs/build/prepare.ts
wulfthimm: I am still working on it.

ArthurSens: Ah sorry 👍

ArthurSens (commented):
Thanks for tackling this bug introduced by me wulf! 🤗 

I feel like your approach is overcomplicating things a bit though... Could we stick with the original implementation, while adding your check following the [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single-responsibility_principle)?

Does this approach looks better to read?
```
function decideHarvesterVMCreation(werft: Werft, config: JobConfig) {
    if (config.withVM && shouldCreateVM(config)) {
        prepareVM(werft, config)
    } else {
        werft.currentPhaseSpan.setAttribute("werft.harvester.created_vm", false)
    }
    werft.done(prepareSlices.BOOT_VM)
}

function shouldCreateVM(config: JobConfig) {
    return !VM.vmExists({ name: config.previewEnvironment.destname }) || config.cleanSlateDeployment
}
```

wulfthimm:
@ArthurSens I find my approach much easier to read because it does not use as much functions and is still relatively compact. But I am not a TS developer and do not have any preferences.

ArthurSens:
Using conditionals is nice because it's easy to implement things when in a hurry, but they add complexity as we need to keep track of how all those conditionals correlate between different parts of the code. When reading the code, we need to scroll up and down the code, we need to look and different levels of nested conditionals and it does add cognitive noise 😵 

Functions, on the other hand, are a powerful thing that can reduce complexity 🙂. We can read code only once in a top-to-bottom approach, while not entering into nested conditionals.

ArthurSens:
@mads-hartmann, wulf and I just had a pair programming session to be more aligned around code style. Do you mind giving it a review then? 🙂

ArthurSens (approved):
@wulfthimm and I just did a pair-programming session to align coding style 🙂

mads-hartmann (approved):
Looks good to me, thanks!

ArthurSens:
/hold

Looks like the job failed due to core-dev flakyness?

/werft run

PR #8750, opened 2022-03-11
Only flush traces once

mads-hartmann:
## Description

This ensure that we only try to flush traces once. We are triggering the flushing in the beforeExit handler which is invoked when Node doesn't have any more code to execute (or Promises in its queue). However, sdk.shutdown generates a new Promise which means that once that's done executing the beforeExit handler is invoked again. This seems to be okay (e.g. we have previous seen:

```
About to exit with code 0. Shutting down tracing.
Tracing terminated
About to exit with code 0. Shutting down tracing.
Tracing terminated
```

So presumably the second shutdown happens before the exitHandler is done executing and thus doesn't introduce an infinite loop. Even so, only trying to invoke sdk.shutdown once seems the better open, hence this PR.

## Related Issue(s)

No previous issue.

## How to test

Just trigger a job and look at the output. We now see

```
About to exit with code 0. Shutting down tracing.
Tracing terminated
About to exit with code 0. Traces already flushed, no further action needed.
```

## Release Notes

```release-note
NONE
```

## Documentation

N/A

### DIFF
--- a/.werft/observability/tracing.ts
+++ b/.werft/observability/tracing.ts
@@ -37,10 +37,17 @@ export async function initialize() {
         process.exit(1)
     }
 
+    let didFlushTraces = false
     process.on('beforeExit', (code) => {
-        console.log(`About to exit with code ${code}. Shutting down tracing.`)
-        sdk.shutdown()
-            .then(() => console.log('Tracing terminated'))
-            .catch((error) => console.log('Error terminating tracing', error))
+        const sliceID = 'tracing shutdown'
+        if (!didFlushTraces) {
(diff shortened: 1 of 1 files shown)


roboquat:
@mads-hartmann: Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://git.k8s.io/community/contributors/guide/release-notes.md) to remove it.

<details>

</details>

ArthurSens (approved):
Ah nice, I always wondered why shutting down tracing was confusing on our logs 😅

PR #8754, opened 2022-03-11
Use placeholder text for Provider Host input

andrew-farries:
## Description

Change the default value in the **Provider Host** field on the new integration modal from a value that the user must delete before entering a valid host, to a [placeholder](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-placeholder) that disappears when the input field has focus.

Before:

https://user-images.githubusercontent.com/8225907/157693897-9cd571e7-2fcf-45c3-9e81-53d401cf1e21.mov

After:

https://user-images.githubusercontent.com/8225907/157693909-694a5167-0418-43e4-8d41-de020864e08f.mov

## How to test

Open the modal at **Settings** -> **Integrations** -> **Add integration** and check that the value in the **Provider Host** field is now a placeholder not a value that must be deleted.

## Release Notes

```release-note
NONE
```

## Documentation

No changes needed.

### DIFF
--- a/components/dashboard/src/settings/Integrations.tsx
+++ b/components/dashboard/src/settings/Integrations.tsx
@@ -448,7 +448,7 @@ export function GitIntegrationModal(props: ({
     const [providerEntry, setProviderEntry] = useState<AuthProviderEntry | undefined>(undefined);
 
     const [type, setType] = useState<string>("GitLab");
-    const [host, setHost] = useState<string>("gitlab.example.com");
+    const [host, setHost] = useState<string>("");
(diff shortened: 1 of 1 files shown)


jankeromnes (approved):
Looks good to me, and works as advertised. Thanks! 🙌

One small nit is that the callback URL no longer updates according to the placeholder value, but since you do need to input an actual hostname, that's actually okay. 👍

gtsiolis:
> One small nit is that the callback URL no longer updates according to the placeholder value, but since you do need to input an actual hostname, that's actually okay.

Nice catch, @jankeromnes!

Thanks for making this change @andrew-farries!

Issue #8719, opened 2022-03-10
Project creation with GitHub repo sometimes fails with 404


svenefftinge:
I see a 404 error coming from GitHub when creating a project for any repo of a certain org. Works with other orgs.

![Screenshot 2022-03-10 at 15 21 10](https://user-images.githubusercontent.com/849592/157695092-84c0410f-d4cf-43fb-ae09-e15f6887c353.png)

So far only reproducing on public repos.

After error, the new project appears to be created (unclear what state it is in)

Issue #8756, opened 2022-03-11
Prebuilds not triggered on GitHub Enterprise projects


jldec:
Projects associated with a GitHub Enterprise repo are not receiving webhooks or triggering prebuilds.
This is ~~probably~~ because of https://github.com/gitpod-io/gitpod/pull/8722

PR #8757, opened 2022-03-11
Fix and re-enable automated prebuilds for GitHub Enterprise

jankeromnes:
## Description

Okay, here is my theory on why project creation sometimes failed with `404` errors. The short version is: We never ensured that the user's GHE token is actually allowed to install a webhook.

1. User logs connects with GitHub Enterprise. Gitpod receives a token with minimum scopes (i.e. [user:email](https://github.com/gitpod-io/gitpod/blob/5e3897debdc361aea07575c4bf6b0529483d5d63/components/server/src/github/scopes.ts#L22))
2. User goes to `/new` to create a new Project, and sees all their repositories as available, because [gh.repos.listForAuthenticatedUser](https://github.com/gitpod-io/gitpod/blob/5e3897debdc361aea07575c4bf6b0529483d5d63/components/server/ee/src/prebuilds/github-service.ts#L29-L41) works without special scopes
3. User selects a repository, Project gets created, then [onDidCreateProject](https://github.com/gitpod-io/gitpod/blob/b10bdc67d96ed0f8c559cc654b969c3bd68c9631/components/server/src/projects/projects-service.ts#L129-L150) looks if it can/should install a webhook
  a. [canInstallAutomatedPrebuilds](https://github.com/gitpod-io/gitpod/blob/5e3897debdc361aea07575c4bf6b0529483d5d63/components/server/ee/src/prebuilds/github-service.ts#L43-L63) returns `true`, because the user has indeed `ADMIN` permission on the repo (and GraphQL tells you that without special scopes)
  b. However, [installAutomatedPrebuilds](https://github.com/gitpod-io/gitpod/blob/5e3897debdc361aea07575c4bf6b0529483d5d63/components/server/ee/src/prebuilds/github-service.ts#L65-L80) then fails, presumably with a `404` error (to not leak the existence of a repo you shouldn't be able to access?), because you do need at least `public_repo` or `repo` scope to actually install a webhook

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/8719

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/dashboard/src/projects/NewProject.tsx
+++ b/components/dashboard/src/projects/NewProject.tsx
@@ -500,13 +500,14 @@ function GitProviders(props: {
         setErrorMessage(undefined);
 
         const token = await getGitpodService().server.getToken({ host: ap.host });
-        if (token) {
+        const isGitHubEnterprise = ap.authProviderType === "GitHub" && ap.host !== "github.com";
+        if (token && !(isGitHubEnterprise && !token.scopes.includes("public_repo"))) {
             props.onHostSelected(ap.host);
             return;
         }
         await openAuthorizeWindow({
             host: ap.host,
-            scopes: ap.requirements?.default,
+            scopes: isGitHubEnterprise ? ["public_repo"] : ap.requirements?.default,
             onSuccess: async () => {
                 props.onHostSelected(ap.host, true);
             },
--- a/components/server/src/projects/projects-service.ts
+++ b/components/server/src/projects/projects-service.ts
@@ -6,15 +6,22 @@
 
 import { inject, injectable } from "inversify";
 import { DBWithTracing, ProjectDB, TeamDB, TracedWorkspaceDB, UserDB, WorkspaceDB } from "@gitpod/gitpod-db/lib";
-import { Branch, PrebuildWithStatus, CreateProjectParams, FindPrebuildsParams, Project, ProjectEnvVar, User } from "@gitpod/gitpod-protocol";
+import {
+    Branch,
+    PrebuildWithStatus,
+    CreateProjectParams,
+    FindPrebuildsParams,
+    Project,
+    ProjectEnvVar,
+    User,
+} from "@gitpod/gitpod-protocol";
 import { HostContextProvider } from "../auth/host-context-provider";
 import { RepoURL } from "../repohost";
-import { log } from '@gitpod/gitpod-protocol/lib/util/logging';
+import { log } from "@gitpod/gitpod-protocol/lib/util/logging";
 import { PartialProject } from "@gitpod/gitpod-protocol/src/teams-projects-protocol";
 
 @injectable()
 export class ProjectsService {
-
     @inject(ProjectDB) protected readonly projectDB: ProjectDB;
     @inject(TeamDB) protected readonly teamDB: TeamDB;
     @inject(UserDB) protected readonly userDB: UserDB;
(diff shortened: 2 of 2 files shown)


jankeromnes:
@andrew-farries @AlexTugarev What do you think about this approach, i.e. ensure at least `public_repo` scope when attempting to select GitHub Enterprise in the New Project flow?

Hint: One commit contains auto-format changes from our pre-commit hook. Please feel free to ignore it to limit noise. [Relevant commit](https://github.com/gitpod-io/gitpod/pull/8757/commits/b9fd06e943f9bb5c77c99c351366f48be575c7e3).

jankeromnes:
> Amended the related issue to https://github.com/gitpod-io/gitpod/issues/8756

I saw 😂 (when I saw the new issue, I started editing my PR description, then wondered "wait this is the same number?!" -- so I cancelled my edit and saw "edited by jldec" 😁)

jldec (changes requested):
Default oauth scope shows list of public repos only in /new, and adding a project for a public repo does not appear to configure the webhook properly for triggering prebiulds (even though I see the webhook on the public repo). There is no indicator in the /new project flow to suggest adding repo scope.

Opening a workspace on a private repo DOES ask for repo scope, and then the /new project page shows both public and private repos, and the private repo project appears to receive webhooks properly. 

I think it would be better for /new to ask for repo scope before listing repos.

jankeromnes:
Many thanks for your feedback @jldec!

> Default oauth scope shows list of public repos only in /new

Yes, that's expected (if you don't grant `repo` scope, you won't see private repos in `/new`). Although, I agree with you that there is no clear UX for how to "see my private repositories".

> and adding a project for a public repo does not appear to configure the webhook properly for triggering prebiulds (even though I see the webhook on the public repo).

That's unexpected. Testing my PR, I was able to successfully:
1. Connect GitHub Enterprise with minimum scope
2. Use the `/new` flow to add a public repository
3. Gitpod then detected new pushes to the public repository, and successfully started new Prebuilds for them

Are you saying that for you my PR did not trigger new prebuilds?

> I think it would be better for /new to ask for repo scope before listing repos.

That's indeed a product choice we can make. I believe our options are:
a. Make the minimum scope for New Projects `public_repo` (thus you'll see only public repos by default, but webhooks should work on them)
b. Make the minimum scope for New Projects `repo` (thus forcing users to also grant private repo access, which they might not want, but makes for a simpler UX)

components/dashboard/src/projects/NewProject.tsx
```diff
@@ -500,13 +500,14 @@ function GitProviders(props: {
         setErrorMessage(undefined);
 
         const token = await getGitpodService().server.getToken({ host: ap.host });
-        if (token) {
+        const isGitHubEnterprise = ap.authProviderType === "GitHub" && ap.host !== "github.com";
```
andrew-farries, line 503: Not introduced on this PR, but we should have a narrower type here than a `string`, ie a union like `"GitHub" | "Gitlab" | ... `.

components/server/src/projects/projects-service.ts
```diff
@@ -148,7 +148,7 @@ export class ProjectsService {
         const parsedUrl = RepoURL.parseRepoUrl(project.cloneUrl);
         const hostContext = parsedUrl?.host ? this.hostContextProvider.get(parsedUrl?.host) : undefined;
         const type = hostContext && hostContext.authProvider.info.authProviderType;
-        if (type === "GitLab" || type === "Bitbucket") {
+        if (type === "GitLab" || type === "Bitbucket" || (type === "GitHub" && parsedUrl?.host !== "github.com")) {
```
andrew-farries, line 151: Feels like we should be able to pull out a function here called something like `isGithubEnterprise`.

andrew-farries (changes requested):

components/dashboard/src/projects/NewProject.tsx
jankeromnes: Good point @andrew-farries! I agree with you, but since this implies modifying a central protocol symbol, I'll not address it here, but leave it for this follow-up issue: https://github.com/gitpod-io/gitpod/issues/8788

components/server/src/projects/projects-service.ts
jankeromnes: Good point, thanks! Done ✅

andrew-farries (approved):

andrew-farries (approved):

andrew-farries (approved):

PR #8759, opened 2022-03-11
[installer] set the ServiceType as LB for ws-proxy

### LINKED ISSUE #7734: [installer] Add support for Ingress traffic in cloud providers
### Bug description

The installer uses services `type=ClusterIP`. In particular for workspace clusters we need to use `type=LoadBalancer`, using standard ports 80 and 443. If we could also add support for additional annotations it would be ideal.

The goal here is to reduce the need for post-processing.

### Steps to reproduce

-

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

nandajavarma:
## Description

Currently, the type of the service created for `ws-proxy` is `ClusterIP` by default. When setting up in cloud in just the `Workspace` installation, this creates a requirement for having to post process the YAML to change the type to `LoadBalancer`. With this PR, the default value of `ws-proxy` service type is set as `LoadBalancer` when the installation is of the kind `Workspace`

## Related Issue(s)

Fixes #7734

## How to test

Set the value of `kind` in config file `Workspace`. Run the render command, `installer render --config gitpod.config.yaml > gitpod.yaml`. This will render `Service` definition of the app `ws-proxy` as type `LoadBalancer`

## Release Notes

```release-note
NONE
```

## Documentation

No

### DIFF
--- a/install/installer/README.md
+++ b/install/installer/README.md
@@ -163,12 +163,21 @@ yq eval-all --inplace \
   gitpod.yaml
 ```
 
+Similarly, if you are doing a `Workspace` only install (specifying `Workspace`
+as the kind in config) you might want to change the service type of `ws-proxy` to `ClusterIP`
+instead of the default `LoadBalancer`. You can post-process the YAML to change that.
+
+```shell
+yq eval-all --inplace \
+  '(select(.kind == "Service" and .metadata.name == "ws-proxy") | .spec.type) |= "ClusterIP"' \
+  gitpod.yaml
+```
+
 ## Error validating `StatefulSet.status`
 
 ```shell
 error: error validating "gitpod.yaml": error validating data: ValidationError(StatefulSet.status): missing required field "availableReplicas" in io.k8s.api.apps.v1.StatefulSetStatus; if you choose to ignore these errors, turn validation off with --validate=false
 ```
-
 Depending upon your Kubernetes implementation, you may receive this error. This is
 due to a bug in the underlying StatefulSet dependency, which is used to generate the
 OpenVSX proxy (see [#8529](https://github.com/gitpod-io/gitpod/issues/8529)).
--- a/install/installer/pkg/components/ws-proxy/objects.go
+++ b/install/installer/pkg/components/ws-proxy/objects.go
@@ -6,6 +6,9 @@ package wsproxy
 
 import (
 	"github.com/gitpod-io/gitpod/installer/pkg/common"
+	"github.com/gitpod-io/gitpod/installer/pkg/config/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 )
 
 var Objects = common.CompositeRenderFunc(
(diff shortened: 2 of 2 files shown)


install/installer/pkg/components/ws-proxy/objects.go
```diff
@@ -14,23 +17,33 @@ var Objects = common.CompositeRenderFunc(
...
+				ContainerPort: SSHTargetPort,
+				ServicePort:   SSHServicePort,
+			},
+		}
+		return common.GenerateService(Component, ports, func(service *corev1.Service) {
+			// In the case of Workspace only setup, `ws-proxy` service is the entrypoint
+			// Hence we use LoadBalancer type for the service
+			if cfg.Config.Kind == config.InstallationWorkspace {
+				service.Spec.Type = corev1.ServiceTypeLoadBalancer
+				service.Annotations["cloud.google.com/neg"] = `{"exposed_ports": {"80":{},"443": {}}}`
```
princerachit, line 44: Do we need to hardcode google specific annotations here?

MrSimonEmms: These annotations have been hardcoded for some time on the basis that a) they're ignored if not used and b) we don't really have an effective (and non-verbose) way of adding custom annotations to services. Elsewhere, there's also an annotation for `external-dns` with the same thinking

install/installer/README.md
```diff
@@ -163,12 +163,21 @@ yq eval-all --inplace \
   gitpod.yaml
 ```
 
+Similarly, if you are doing a `Workspace` only install (specifying `Workspace`
+as the kind in config) you might want to change the service type of `ws-proxy` to `ClusterIP`
+instead of the default `LoadBalancer`. You can post-process the yaml to change that.
```
MrSimonEmms, line 168: nit: `YAML`, not `yaml`

nandajavarma: done!

princerachit (approved):
I did not test the changes but code lgtm

install/installer/README.md
```diff
@@ -163,12 +163,21 @@ yq eval-all --inplace \
...
 
+Similarly, if you are doing a `Workspace` only install (specifying `Workspace`
+as the kind in config) you might want to change the service type of `ws-proxy` to `ClusterIP`
+instead of the default `LoadBalancer`. You can post-process the YAML to change that.
+
+```shell
+yq eval-all --inplace \
+  '(select(.kind == "Service" and .metadata.name == "ws-proxy") | .spec.type) |= "ClusterIP"' \
+  gitpod.yaml
+```
```
corneliusludmann, line 174: Off-topic: Do we have already an issue to get rid of this post-processing step and having a config value for the service type?

nandajavarma: @corneliusludmann there is no such issue open specifically for making the service type configurable. I guess the question is do we plan on making this configurable, since our thought process [was to not make it configurable.](https://github.com/gitpod-io/gitpod/pull/7280)

corneliusludmann: I'm glad you brought that up. :rocket: Indeed, we decided like this in the past. However, 2 things have changed since then:
- We introduced the `experimental` section. Everything that we won't support for some reason but need internally for SaaS should go there. We shouldn't need to rely on post-processing.
- Since our official supported installation path will be the KOTS installation, we are a little more relaxed about the config surface.

What do you think, @MrSimonEmms?

nandajavarma: Sorry for the inactivity on this issue. I have created[ this issue ](https://github.com/gitpod-io/gitpod/issues/8838) to track this requirement. Shall we merge this PR and continue the work there?

princerachit (approved):
code changes lgtm. did not test this though

Furisto:
Hey @MrSimonEmms, if I understood correctly you agree with the changes in this PR. If so, can you please approve it?

MrSimonEmms (approved):
LGTM

Issue #8493, opened 2022-02-28
Document SSH Gateway in the installer


loujaybee:
Update SSH installation steps in the installer (for JetBrains). 

https://github.com/gitpod-io/gitpod/tree/main/install/installer

PR #8761, opened 2022-03-11
[installer] Add docs for workspace SSH access

corneliusludmann:
## Description
This pull request adds docs on how to configure the Gitpod installer to enable the SSH gateway.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/8493

## Release Notes

```release-note
NONE
```

## Meta

/werft no-preview

### DIFF
--- a/install/installer/docs/workspace-ssh-access.md
+++ b/install/installer/docs/workspace-ssh-access.md
@@ -1,0 +1,26 @@
+# Installing Gitpod with SSH access to workspaces with the Gitpod Installer
+**_Needed by some desktop IDEs to connect to a workspace._**
+
+> **IMPORTANT:** When you use k3s, this will open port 22 on your Kubernetes nodes for accessing the workspaces. This will prevent login to the cluster via SSH. If you wish to maintain SSH access to your cluster, please configure another SSH port on your nodes.
+
+To enable the SSH gateway you need to generate a host key, need to add it as secret to your Kubernetes cluster, and need to configure the Gitpod Installer to use this secret.
+
+You can use `ssh-keygen` to generate a host key like this:
+
+```
+ssh-keygen -t rsa -q -N "" -f host.key
+```
+
+Add it to your Kubernetes cluster like this:
+```
+kubectl create secret generic ssh-gateway-host-key --from-file=host.key
+```
+
+Add the following to your Gitpod config `gitpod.config.yaml`:
+```yaml
+sshGatewayHostKey:
+  kind: secret
+  name: ssh-gateway-host-key
+```
+
+That's it. Install Gitpod as usual, consider adding firewall rules that allow the connection on port 22 to your workspace nodes, and you can use desktop IDEs that need SSH access to connect to your workspace.


install/installer/docs/workspace-ssh-access.md
```diff
@@ -0,0 +1,26 @@
...
+**_Needed by some desktop IDEs to connect to a workspace._**
+
+> **IMPORTANT:** This uses port 22 on your Kubernetes nodes. When enabled, this will prevent login to the cluster via SSH. If you wish to maintain SSH access to your cluster, please configure another SSH port on your nodes.
+
+To enable the SSH gateway you need to generate a host key, need to add it as secret to your Kubernetes cluster, and need to configure the Gitpod Installer to use this secret.
+
+You can use `ssh-keygen` to generate a host key like this:
+
+```shell
+$ ssh-keygen -t rsa -q -N "" -f host.key
```
MrSimonEmms, line 11: Might be a nit, but I never like bash that starts with `$` as you can't just copy/paste

install/installer/docs/workspace-ssh-access.md
```diff
@@ -0,0 +1,26 @@
...
+```
+
+Add it to your Kubernetes cluster like this:
+```shell
+$ kubectl create secret generic ssh-gateway-host-key --from-file=host.key
```
MrSimonEmms, line 16: As above

MrSimonEmms (approved):
/hold

I've put a nit in there. Happy for you to merge as-is though if you want

MrSimonEmms (dismissed):
👍🏻

install/installer/docs/workspace-ssh-access.md
```diff
@@ -0,0 +1,26 @@
+# Installing Gitpod with SSH access to workspaces with the Gitpod Installer
+**_Needed by some desktop IDEs to connect to a workspace._**
+
+> **IMPORTANT:** This uses port 22 on your Kubernetes nodes. When enabled, this will prevent login to the cluster via SSH. If you wish to maintain SSH access to your cluster, please configure another SSH port on your nodes.
```
iQQBot, line 4: In fact, only the use of hostPort (pod) and nodePort (services) will affect the SSH port of the host

iQQBot (approved):
This is enough for GKE/AKS/EKS which use loadbalancer, users who use loadbalancer do not need to modify the node ssh port

install/installer/docs/workspace-ssh-access.md
```diff
@@ -0,0 +1,26 @@
+# Installing Gitpod with SSH access to workspaces with the Gitpod Installer
+**_Needed by some desktop IDEs to connect to a workspace._**
+
+> **IMPORTANT:** If you use hostPort or nodePort as the method for exposing ports to the public, This will uses port 22 on your Kubernetes nodes. When enabled, this will prevent login to the cluster via SSH. If you wish to maintain SSH access to your cluster, please configure another SSH port on your nodes.
```
corneliusludmann, line 4: Could you explain this from a user perspective? I'm not aware that I changed the exposing port method somewhere but had a problem with my SSH connection after enabling the SSH gateway. I used k3s on a VM and I was not able to connect to the VM via SSH anymore. See also [this internal discussion](https://gitpod.slack.com/archives/C01KGM9BH54/p1645522647464259).

How do I now as a user of the Gitpod Installer whether I'm affected?

iQQBot: The exception is k3s, which has a loadbalancer that opens the port directly on the host 😂

iQQBot: This whole mechanism and situation is very complicated, I would prefer to
1. be able to configure the SSH Gateway port, so as to avoid a lot of very complicated situations
2. reuse port 443, which is very complex and has some performance loss

iQQBot: for k3s see https://rancher.com/docs/k3s/latest/en/networking/#how-the-service-lb-works for detail

corneliusludmann: Ah, ok. What do you think about this?

```suggestion
> **IMPORTANT:** When you use k3s, this will open port 22 on your Kubernetes nodes for accessing the workspaces. This will prevent login to the cluster via SSH. If you wish to maintain SSH access to your cluster, please configure another SSH port on your nodes.
```

iQQBot: LGTM

iQQBot: Given that our README and installation guides only provide installation guides for GKE/AKS/EKS (loadbalancer) and k3s, I think this basically covers most cases

iQQBot (approved):

MrSimonEmms (approved):

PR #8769, opened 2022-03-11
[cgroupv2] Enable fuse device

### LINKED ISSUE #8417: Make cgroup customizer support v2
In cgroup v2 the `device` controller no longer exists, but the device access is mediated using BPF. We don't want to modify the BPF program at runtime, but instead add the device to the OCI runtime spec.

To this end we'd ship something akin to the `runc-facade` using an init container to the node, and use a RuntimeClass to configure the use of that runc-facade.

Furisto:
## Description
Enable fuse device on nodes that are using cgroup v2. 

## Related Issue(s)
Fixes #8417

## How to test
- Open workspace on a cgroup v2 system
- Compile this program (gcc -o fuse_test fuse_test.c)
```
#define _GNU_SOURCE
#include <unistd.h>

#include <sys/syscall.h>
#include <linux/fs.h>
#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>
#include <stdio.h>


int main() {
  const char* src_path = "/dev/fuse";
  unsigned int flags = O_RDWR;
  printf("RET: %ld", syscall(SYS_openat, AT_FDCWD, src_path, flags));
}
```
- fuse_test should return 3

## Notes for reviewers
My initial implementation was using a runc facade like we discussed in #8471, but I switched to instead replacing the ebpf program that runc generates with our own.
- No need to touch the host (copying the runtime facade, modifying the config of containerd or of other runtimes if we want to support them in the future)
- No need to specify a runtime class when creating the workspace
- Implementation uses the same approach as the cgroup v1 implementation
- Less code

## Release Notes
```release-note
Enable the use of fuse device on cgroup v2 systems
```

### DIFF
--- a/components/ws-daemon/pkg/daemon/cgroup_customizer.go
+++ b/components/ws-daemon/pkg/daemon/cgroup_customizer.go
@@ -6,10 +6,17 @@ package daemon
 
 import (
 	"context"
+	"path/filepath"
 
 	"github.com/containerd/cgroups"
+	"github.com/gitpod-io/gitpod/ws-daemon/pkg/container"
 	"github.com/gitpod-io/gitpod/ws-daemon/pkg/dispatch"
+	"github.com/opencontainers/runc/libcontainer/cgroups/ebpf"
+	"github.com/opencontainers/runc/libcontainer/cgroups/ebpf/devicefilter"
+	"github.com/opencontainers/runc/libcontainer/devices"
+	"github.com/opencontainers/runc/libcontainer/specconv"
 	"github.com/opencontainers/runtime-spec/specs-go"
+	"golang.org/x/sys/unix"
 	"golang.org/x/xerrors"
 )
 
@@ -18,24 +25,31 @@ var (
 	fuseDeviceMinor int64 = 229
 )
 
-type CgroupCustomizer struct {
+func NewCGroupCustomizer(basePath string, unified bool) dispatch.Listener {
+	if unified {
+		return &CgroupCustomizerV2{
+			cgroupBasePath: basePath,
+		}
+	} else {
+		return &CgroupCustomizerV1{
+			cgroupBasePath: basePath,
+		}
+	}
+}
+
+type CgroupCustomizerV1 struct {
 	cgroupBasePath string
 }
 
-func (c *CgroupCustomizer) WithCgroupBasePath(basePath string) {
+func (c *CgroupCustomizerV1) WithCgroupBasePath(basePath string) {
 	c.cgroupBasePath = basePath
 }
 
 // WorkspaceAdded will customize the cgroups for every workspace that is started
-func (c *CgroupCustomizer) WorkspaceAdded(ctx context.Context, ws *dispatch.Workspace) error {
-	disp := dispatch.GetFromContext(ctx)
-	if disp == nil {
-		return xerrors.Errorf("no dispatch available")
-	}
-
-	cgroupPath, err := disp.Runtime.ContainerCGroupPath(context.Background(), ws.ContainerID)
+func (c *CgroupCustomizerV1) WorkspaceAdded(ctx context.Context, ws *dispatch.Workspace) error {
+	cgroupPath, err := retrieveCGroupPath(ctx, ws.ContainerID)
 	if err != nil {
-		return xerrors.Errorf("cannot start governer: %w", err)
+		return err
 	}
 
 	control, err := cgroups.Load(c.customV1, cgroups.StaticPath(cgroupPath))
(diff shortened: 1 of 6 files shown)


Furisto:
@csweichel Any idea why this fails in CI? I can build it with leeway in my workspace without issue.

components/ws-daemon/pkg/daemon/cgroup_customizer.go
```diff
@@ -18,24 +25,31 @@ var (
...
+}
+
+type CgroupCustomizerV1 struct {
 	cgroupBasePath string
 }
 
-func (c *CgroupCustomizer) WithCgroupBasePath(basePath string) {
+func (c *CgroupCustomizerV1) WithCgroupBasePath(basePath string) {
 	c.cgroupBasePath = basePath
 }
```
utam0k, line 47: I think we don't need this function now.
```suggestion
```

components/ws-daemon/pkg/daemon/cgroup_customizer.go
```diff
@@ -64,8 +78,72 @@ func (c *CgroupCustomizer) WorkspaceAdded(ctx context.Context, ws *dispatch.Work
...
+		Type:        'c',
+		Major:       fuseDeviceMajor,
+		Minor:       fuseDeviceMinor,
+		Permissions: "rwm",
+		Allow:       true,
+	}
+
+	deviceRules := make([]*devices.Rule, 0)
+	deviceRules = append(deviceRules, &denyAll, &allowFuse)
+	for _, device := range specconv.AllowedDevices {
```
utam0k, line 144: In our current use case, we don't specify cgroup devices. In other words, this is all we need to specify for now. However, if we want to add some devices about the runc layer(e.g. kubelet, contained) in the future, this part will arise a problem. However, it may be an option to decide that the cgroup devices are managed here. In fact, we have been used to overwrite `/dev/fuse` here.
@ChristinFrohne WDYT?

utam0k:
Thanks for your PR. Super. I have two questions.
1. I wonder if the `CgroupCustomizerV2` takes effect on the performance because ws-daemon loads eBPF code to kernel every 15 seconds.
2. I wonder if there may be a time when this override is not reflected for a few seconds since it is a dispatch.

components/ws-daemon/pkg/daemon/cgroup_customizer.go
```diff
@@ -64,8 +78,72 @@ func (c *CgroupCustomizer) WorkspaceAdded(ctx context.Context, ws *dispatch.Work
...
+		return xerrors.Errorf("cannot get directory fd for %s", fullCgroupPath)
+	}
+	defer unix.Close(cgroupFD)
+
+	insts, license, err := devicefilter.DeviceFilter(composeDeviceRules())
+	if err != nil {
+		return xerrors.Errorf("failed to generate device filter: %w", err)
+	}
+
+	_, err = ebpf.LoadAttachCgroupDeviceFilter(insts, license, cgroupFD)
```
utam0k, line 109: If there is an err, it is necessary to call `closer` of the return value. Or you may need to revert back to the original ebpf program. the `closer` will only detach and will not revert the original ebpf code. Also, I should be a bit concerned that the workspace will probably stop working if this code fails, but I don't have a good idea... 😭 
@Furisto @csweichel Are there any good ideas?

components/ws-daemon/pkg/daemon/cgroup_customizer.go
utam0k: What about applying it to the pod cgroup instead of the container cgroup and erasing the ebpf in the container cgroup? Maybe elimination may not be necessary due to inheritance from parents. In this case, even if it fails, the ebpf code for the container's cgroup device will remain, so there should be no such thing as the user being able to access all devices.

components/ws-daemon/pkg/daemon/cgroup_customizer.go
utam0k: What about applying it to the pod cgroup instead of the container cgroup and erasing the ebpf in the container cgroup? Maybe elimination may not be necessary due to inheritance from parents. In this case, even if it fails, the ebpf code for the container's cgroup device will remain, so there should be no such thing as the user being able to access all devices.

utam0k:
Oh, I misunderstood. There is no problem.
> Are you thinking of the cpu limiter? The CgroupCustomizer should only be called when a new workspace is added.

I think the worst-case scenario is that the absence of the eBPF code for the devices controller is the item of greatest concern. Was the eBPF code loaded in the pod cgroup?
> It is possible but would there be any serious risk? We replace a more restrictive program with a less restrictive program. At worst the customer would not be able to use the fuse device some time after workspace start (or not at all if the event is somehow lost)

components/ws-daemon/pkg/daemon/cgroup_customizer.go
Furisto: There are two cases with the closer:
- There is only one ebpf program attached to the cgroup (this is the current case). In this case the closer is a dummy and does not do anything -> No need to call the closer
- There are multiple ebpf programs attached to the cgroup. In this case the closer will detach the program we have attached when an error occurs while detaching the other programs. That is the order is -> attach new program -> setup closer -> detach all other programs. In that case I would not want to call the closer either because it could mean that we have detached the old runc device rules, then some error happens while detaching another program and then if we call the closer we would detach our newly attached program leaving us with no device rules to protect us.

> Also, I should be a bit concerned that the workspace will probably stop working if this code fails, but I don't have a good > idea... sob

Why would the workspace stop working?

My understanding was that ebpf programs are atomically replaced, so either the old program is attached or the new program,

Furisto: > My understanding was that ebpf programs are atomically replaced, so either the old program is attached or the new program.

I have asked in the ebpf slack and got the confirmation that this is indeed the behavior.

utam0k: Oops, I was reading the code backward for detach and attach in the old eBPF code. My brain seems to be not good today. Looks like it's not a problem.
https://github.com/opencontainers/runc/blob/067aaf8548d78269dcb2c13b856775e27c410f9c/libcontainer/cgroups/ebpf/ebpf_linux.go#L242

Furisto:
@csweichel What do you think?

components/ws-daemon/pkg/daemon/cgroup_customizer.go
utam0k: @Furisto Since there is currently a way to get the contents of `config.json`, why don't we get the original information from here? To be honest, I want to get `status.json`, but `config.json` should be sufficient.
https://github.com/gitpod-io/gitpod/blob/60a9c550444f10dca15d4e39d12dc8d561500bc9/components/ws-daemon/pkg/container/containerd.go#L460-L470

kylos101: Hey @utam0k , that sounds like an optimization we could potentially add later, yes? In other words, let's try to keep this PR smaller, ship the skateboard, and ship another one in the future (if we need).

Aside from this observation, do you have any other concerns which are blocking? I ask so that we can get @Furisto the feedback needed to have this approved and merged.

utam0k: I did a little research and as it stands it shouldn't be a problem, because there is no API to change devices normally from k8s. However I think this should be improved. It can create some pretty strange bugs. The skateboard is a very good watchword 🛹

utam0k (approved):
Looks good

PR #8785, opened 2022-03-14
[kots] remove cert-manager from Gitpod package

### LINKED ISSUE #8749: [installer] Remove `cert-manager` Helm chart from Kots
With Replicated, Packaging `cert-manager` along with Gitpod seems to give us more
problems than benefit. Starting with us having to show configurations to make it work
in the popular cloud provider Kubernetes distributions i.e EKS, AKS, etc and that will
only grow as we will have more of these diverse set of Kubernetes clusters.

After talking to folks at replicated, It was suggested to instead ask users to get it up
and running (some users could have already been doing that). We could also have
documentation to go along with this, to make it explicit.

Pothulapati:
## Description

Currently, We include `cert-manager` instance into the
Gitpod `kots` package, and install it whenever Gitpod installation
is requested.

This causes the following problems:

- `cert-manager` is a separate beast on its own, There are numerous
configurations that might be needed to be set to make it run correctly
based on the cluster config.
- A lot users could have already have it installed, and have to
  struggle to get Gitpod up and running in those cases.

To solve them, We could either provide all the configurations of
`cert-manager` into the Gitpod package to cater users with all
config needs or remove the package and expect `cert-manager` as a
pre-requisite and make sure its documented.

The latter feels better as it removes the maintenance of cert-manager for
us while also allowing users to have specific configuration. The install
path for `cert-manager` is [also pretty well documented](https://cert-manager.io/docs/installation/)
with all the [changes needed based on the cluster environment](https://cert-manager.io/docs/installation/compatibility/).

This PR also removes the cert-manager issuers and updates the config to allow
users to use pre-configured resources

## Related Issue(s)

Fixes #8749 

## How to test

- Install `Gitpod` using `k kots install gitpod-pov/dev-tar`

## Release Notes

```release-note
[kots] Remove `cert-manager` from the Gitpod package
```


install/kots/manifests/gitpod-installer-job.yaml
```diff
@@ -35,11 +35,11 @@ spec:
               echo "Gitpod: Perform the check"
               while [ "$(helm status -n {{repl Namespace }} cert-manager -o json | jq '.info.status == "deployed"')" = "false" ];
               do
-                echo "Gitpod: Release not found - will retry in 10s"
+                echo "Gitpod: cert-manager release not found - will retry in 10s"
```
MrSimonEmms, line 38: Personally, I would remove the whole `initContainer` section. Gitpod installer will handle having no cert-manager by erroring, which is probably an easier way to debug.

If we did keep this in, this command would only check in the namespace that Gitpod will be installed it - it's likely (but not guaranteed) to be in the `cert-manager` namespace.

install/kots/manifests/kots-config.yaml
```diff
@@ -6,7 +6,7 @@ spec:
   groups:
     - name: general
       title: General settings
-      description: General settings for your Gitpod instance
+      description: General settings for your Gitpod instance. Please make sure to have a [running cert-manager instance](https://cert-manager.io/docs/installation/) before proceeding.
```
MrSimonEmms, line 9: I would have a simple boolean, defaulting to `false` (and, if possible, requiring `true`) that says "cert-manager is installed on this system"

MrSimonEmms (commented):
Good start, but a couple of thoughts on it

install/kots/manifests/gitpod-installer-job.yaml
Pothulapati: Agree! Removing this!

install/kots/manifests/kots-config.yaml
```diff
@@ -253,7 +263,7 @@ spec:
           help_text: A self-signed certficate should only be used if applying TLS termination to your load balancer or other proxy.
 
         - name: cert_manager_enabled
-          title: Use cert-manager
+          title: Use existing cert-manager
```
Pothulapati, line 266: Also, I think we have to update this section. Not enabling this does not mean we aren't using cert-manager right? It means we are only using it for internal communication.

cc: @MrSimonEmms

install/kots/manifests/kots-config.yaml
```diff
@@ -4,9 +4,19 @@ metadata:
   name: gitpod
 spec:
   groups:
+    - name: prerequisites
```
Pothulapati, line 7: Had to separate this out, as it was confusing to be in general settings. Currently, Trying to make this config screen pass, only if the items here are ticked off 🤔

Pothulapati:
Updated the PR to not update the pre-requisite requirement of `cert-manager` as we have multiple ways to do that, and are discussing with replicated folks on the best way to have it i.e either a pre-flight check, or in the config itself.

Will raise a PR once we take a decision there. We can review and merge this removal meanwhile.

MrSimonEmms (dismissed):
/hold

This is good work @Pothulapati. I'd like to get the work in to remove the (cluster)issuer setup as well which is why I've put this on hold - if you want me to do it, remove the hold and we'll merge yours in. If you want to add it to this PR, please do so and I'll add that to the review.

Pothulapati:
@MrSimonEmms I've updated the commit log to include a single commit for both this and the other PR that got merged into this.

MrSimonEmms (approved):
LGTM

PR #8793, opened 2022-03-14
Fix caching of GitHub queries

### LINKED ISSUE #7307: Bugs in github request caching
### Bug description

In the file components/server/src/github/api.ts, lines 244, 250, 264, 273, and 288, the cache key is defined as ${params.owner}/${params.owner} instead of ${params.owner}/${params.repo}. This causes the caching of these requests to fail

Fix: Change ${params.owner}/${params.owner} to ${params.owner}/${params.repo}

### Steps to reproduce

Go to /workspaces page and hit F5. All the cached results squash each other and caching fails for all requests.

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

geropl:
## Related Issue(s)

Fixes #7307 

## Release Notes

```release-note
fix caching of GitHub server<->server queries
```

### DIFF
--- a/components/server/src/github/api.ts
+++ b/components/server/src/github/api.ts
@@ -266,7 +266,7 @@ export class GitHubRestApi {
         user: User,
         params: RestEndpointMethodTypes["repos"]["get"]["parameters"],
     ): Promise<Repository> {
-        const key = `getRepository:${params.owner}/${params.owner}:${user.id}`;
+        const key = `getRepository:${params.owner}/${params.repo}:${user.id}`;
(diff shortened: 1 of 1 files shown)


andrew-farries (approved):
Looks good but is there something we can do using typescript's [template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) to make this kind of mistake impossible at compile time?

Perhaps not, as the format of the string is the same whether or not we mix up `owner` and `repo`.

Issue #8790, opened 2022-03-14
Propagate job attributes to all span produced by the job


mads-hartmann:
Currently we set job attributes on the root span [here](https://github.com/gitpod-io/gitpod/blob/main/.werft/jobs/build/job-config.ts#L139-L143). However, we'd like to have these attributes on all spans produced by the job as that will allow us to write much better queries.

One example query we'd like to write is to measure the time the deploy span takes and group by whether withVM was enabled.

Proposed implementation

1. See if the OpenTelemetry Node SDK has an API we can use to set traces to all spans.
2. If not, we probably have to add some custom instrumentation to werft.ts so that we can add the attributes to all spans that we create there.

PR #8797, opened 2022-03-14
Add job attributes to all spans

mads-hartmann:
## Description

This adds Werft job configuration span attributes - the ones in the namespace `werft.job.config.` - to all spans instead of just the root span. This is helpful as we often want to know how a job config influences a specific slice or phase, e.g. how does `werft.job.config.withVM` impact the deploy phase (for duration, error rates, etc.).

A note on the implementation. As far as I could see, the only way to add "global attributes" using the OpenTelemetry SDK for Node would be to provide a custom [SpanProcessor](https://open-telemetry.github.io/opentelemetry-js/interfaces/_opentelemetry_sdk_trace_base.SpanProcessor.html) which would add the attributes as it "processes" the spans - we would then add that along-side the one we use already. However, the SpanProcessor interface doesn't allow you to modify spans onEnd (it takes a ReadableSpan) only in onStart. That means we wouldn't get the attributes on all spans, only the ones that are started after we have parsed the job configuration (this means we wouldn't get it on the root span unless we also added logic to add it there).

So instead I went with adding the attributes to all the spans in our Werft class. This will still only allow us to add the attributes to spans in the currently open phase and all future phases/slices. As parsing the job configuration is the first phase, this effectively means we get it on the spans for all Werft spans.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/8790

## How to test

Here's [the trace](https://ui.honeycomb.io/gitpod/datasets/werft/trace/2tL6meww9YA?span=a02c8bdca5f4afc1) for [gitpod-build-mads-hartmann-propagate-job-attributes-8790.5](https://werft.gitpod-dev.com/job/gitpod-build-mads-hartmann-propagate-job-attributes-8790.5). The screenshot below shows that the werft attributes are not added to other spans than just the root span.

Screenshot 2022-03-17 at 20 26 48

## Release Notes

```release-note
NONE
```

## Documentation

N/A

### DIFF
--- a/.werft/jobs/build/job-config.ts
+++ b/.werft/jobs/build/job-config.ts
@@ -49,9 +49,9 @@ export interface Observability {
 }
 
 export function jobConfig(werft: Werft, context: any): JobConfig {
-    const slideId = 'Parsing job configuration'
+    const sliceId = 'Parsing job configuration'
     werft.phase('Job configuration')
-    werft.log(slideId, "Parsing the job configuration")
+    werft.log(sliceId , "Parsing the job configuration")
     const version = parseVersion(context)
     const repo = `${context.Repository.host}/${context.Repository.owner}/${context.Repository.repo}`;
     const mainBuild = repo === "github.com/gitpod-io/gitpod" && context.Repository.ref.includes("refs/heads/main");
@@ -144,9 +144,9 @@ export function jobConfig(werft: Werft, context: any): JobConfig {
         return [`werft.job.config.${key}`, value]
     }))
     globalAttributes['werft.job.config.branch'] = context.Repository.ref
-    werft.addGlobalAttributes(globalAttributes)
+    werft.addAttributes(globalAttributes)
 
-    werft.done(slideId)
+    werft.done(sliceId)
 
     return jobConfig
 }
--- a/.werft/jobs/build/job-config.ts
+++ b/.werft/jobs/build/job-config.ts
@@ -49,6 +49,9 @@ export interface Observability {
 }
 
 export function jobConfig(werft: Werft, context: any): JobConfig {
+    const slideId = 'Parsing job configuration'
+    werft.phase('Job configuration')
+    werft.log(slideId, "Parsing the job configuration")
     const version = parseVersion(context)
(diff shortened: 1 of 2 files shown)


.werft/util/werft.ts
```diff
@@ -115,4 +118,23 @@ export class Werft {
         this.endPhase()
         this.rootSpan.end()
     }
+
+    /**
+     * This allows you to set attributes on all open and future Werft spans.
+     * Any spans in phases that have already been closed won't get the attributes.
+     */
+    public addGlobalAttributes(attributes: SpanAttributes): void {
```
ArthurSens, line 126: We are currently using it to add attributes that we call "global", but the method can be used to add any attribute, right?

```suggestion
    public addAttributes(attributes: SpanAttributes): void {
```

ArthurSens (commented):
Just some nit comments, overall looks good :)

.werft/jobs/build/job-config.ts
```diff
@@ -136,11 +139,15 @@ export function jobConfig(werft: Werft, context: any): JobConfig {
...
     werft.log("job config", JSON.stringify(jobConfig));
-    werft.rootSpan.setAttributes(Object.fromEntries(Object.entries(jobConfig).map((kv) => {
+    const globalAttributes = Object.fromEntries(Object.entries(jobConfig).map((kv) => {
         const [key, value] = kv
         return [`werft.job.config.${key}`, value]
-    })))
-    werft.rootSpan.setAttribute('werft.job.config.branch', context.Repository.ref)
+    }))
+    globalAttributes['werft.job.config.branch'] = context.Repository.ref
+    werft.addGlobalAttributes(globalAttributes)
```
mads-hartmann, line 147: Matching the [suggestion below](https://github.com/gitpod-io/gitpod/pull/8797/files#r829792620)

```suggestion
    werft.addAttributes(globalAttributes)
```

mads-hartmann:
@ArthurSens Thanks for the suggestions. I applied them and verified everything still works :)

ArthurSens (approved):
noice!

PR #8801, opened 2022-03-14
Format components/server with prettier

easyCZ:
## Description

* Reformat files under components/server with prettier
* Done as a bulk reformat to avoid PRs which mix logic and style

## Release Notes

```release-note
NONE
```


geropl (approved):
:+1:

PR #8802, opened 2022-03-14
Format components/dashboard with prettier

easyCZ:
## Description

* Reformat files under components/dashboard with prettier
* Done as a bulk reformat to avoid PRs which mix logic and style

## Release Notes

```release-note
NONE
```


geropl (approved):
One thing to note here: With this change, we invalidate all website artifacts in all caches derived from dashboard code.
We just have to be aware that we might see increased traffic after deployment.

PR #8817, opened 2022-03-15
Extend build to typecheck Typescript files in .werft

mads-hartmann:
## Description

Currently we don't validate any of the Typescript files we have in .werft as part of our build. We indirectly test build.ts of course as we run it for our build, but if you're modifying any of the other jobs it's possible to get a green build even when you have type errors (as I have done in the past, oops!)

This runs `tsc --noEmit` as part of our build and also fixes the TS error I had introduced previously.

## Related Issue(s)

(Indirectly) Part of https://github.com/gitpod-io/ops/issues/1255 as that's where I introduced the TS errors.

## How to test

- See job that failed when I had introduced the typechecking, but not yet fixed the errors ([link](https://werft.gitpod-dev.com/job/gitpod-build-mads-extend-build-to-tsc-werft-jobs.1))
- See job that succeeded after having fixed the TS errors ([link](https://werft.gitpod-dev.com/job/gitpod-build-mads-extend-build-to-tsc-werft-jobs.4))
- Job that I triggered manually using `werft run github -j platform-delete-preview-environments-cron.yaml -s platform-delete-preview-environments-cron.ts` ([link](https://werft.gitpod-dev.com/job/gitpod-custom-mads-extend-build-to-tsc-werft-jobs.0))

## Release Notes

```release-note
NONE
```

## Documentation

N/A

### DIFF
--- a/.werft/build.ts
+++ b/.werft/build.ts
@@ -10,6 +10,7 @@ import { prepare } from './jobs/build/prepare';
 import { deployToPreviewEnvironment } from './jobs/build/deploy-to-preview-environment';
 import { triggerIntegrationTests } from './jobs/build/trigger-integration-tests';
 import { jobConfig } from './jobs/build/job-config';
+import { typecheckWerftJobs } from './jobs/build/typecheck-werft-jobs';
 
 // Will be set once tracing has been initialized
 let werft: Werft
@@ -47,6 +48,7 @@ async function run(context: any) {
 
     await validateChanges(werft, config)
     await prepare(werft, config)
+    await typecheckWerftJobs(werft)
     await buildAndPublish(werft, config)
 
     if (config.noPreview) {
--- a/.werft/jobs/build/typecheck-werft-jobs.ts
+++ b/.werft/jobs/build/typecheck-werft-jobs.ts
@@ -1,0 +1,17 @@
+import { Werft } from "../../util/werft";
+import { exec } from "../../util/shell";
+
+/**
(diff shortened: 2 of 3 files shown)


meysholdt:
Nice initiative!

Would there be benefits of doing the compile check at the time when the Werft build is executed? i.e. here?
https://github.com/gitpod-io/gitpod/blob/8dd2ddd844f30a4ff66d2704f4714e9da875c7d5/.werft/build.yaml#L205-L208

E.g. do we now compile all `.werft/*.ts` twice? (Maybe we don't if npx stores .js files and tsc is incremental - I didn't check).

mads-hartmann:
@meysholdt I'd like to keep the bash part of the job as minimal as possible and leave the "build orchestration" in Typescript. We could pass `--transpileOnly` to ts-node to disable typechecking (which means we'd only compile once) but I'm not in favour of that because

1. I like that the build job fails early in case of TS errors.
2. Right now `tsc --noEmit` only takes 2.3s ([ref](https://ui.honeycomb.io/gitpod/datasets/werft/trace/boH4dbgo8Gz?span=19ba01145e9a130f)) so it's not super important to optimise that away at the moment ☺️

ArthurSens (approved):
In my opinion the extra 2s is fine, but I'll let mo have the final say on this in case there are other concerns besides the extra 2 seconds.

/hold

mads-hartmann:
I'm merging this, we can always have a follow up PR. Right now the clean preview environments job doesn't run so I'd like to get this in ☺️

Issue #8604, opened 2022-03-04
Add additional DNS01 challenge resolvers


MrSimonEmms:
The current DNS01 challenge resolvers are Google and Azure, with an optional in-cluster (self-signed cert - see #8559). 

# Required
- [ ] [Route53](https://cert-manager.io/docs/configuration/acme/dns01/route53/)

# Recommended
- [ ] [DigitalOcean](https://cert-manager.io/docs/configuration/acme/dns01/digitalocean/)
- [ ] [CloudFlare](https://cert-manager.io/docs/configuration/acme/dns01/cloudflare/)

# Optional
- [ ] [ACME-DNS](https://cert-manager.io/docs/configuration/acme/dns01/acme-dns/) - may be affected by the documented limitations as would require 3 domains (`$DOMAIN`, `*.$DOMAIN` and `*.ws.$DOMAIN`)
- [ ] [Akamai](https://cert-manager.io/docs/configuration/acme/dns01/akamai/)
- [ ] [RFC-2136](https://cert-manager.io/docs/configuration/acme/dns01/rfc2136/)
- [ ] [Webhook](https://cert-manager.io/docs/configuration/acme/dns01/webhook/)

MrSimonEmms:
Closing as cert-manager is now no longer configured by KOTS /cc @gitpod-io/engineering-self-hosted

Pothulapati:
@MrSimonEmms We don't want to install `cert-manager`  but we still create the `Certificate` object and there we need the DNS01 challenge right? 🤔 

Even #8785 doesn't remove the DNS101 config, right now!

MrSimonEmms:
Good call @Pothulapati

PR #8822, opened 2022-03-15
[kots]: removed the cert-manager issuers

MrSimonEmms:
## Description

Removed the cert-manager issuers and used pre-configured resources.

This now expects that the `Issuer` or `ClusterIssuer` is already pre-configured on the cluster if someone wants to use cert-manager to generate their TLS certs. 

image

## How to test

Deploy via KOTS

## Release Notes

```release-note
[kots]: removed the cert-manager issuers
```


Pothulapati (approved):
I'm +1 to this separation of concerns, as it allows us to not install `Issuers` (which can be shared between components by cert-manager) and also not have to worry about doing https://github.com/gitpod-io/gitpod/issues/8604

/hold As It's a user facing change (as we would expect users to set it up), I'm going to ask @corneliusludmann for a review.

Issue #8755, opened 2022-03-11
Avatar access restriction causes overlaid text in project branch list.


jldec:
- Connect gitpod.io to a self-hosted GitHub enterprise
- Create a project
- The branch list shows overlaid text where there are normally user avatars

![Screenshot 2022-03-11 at 13 41 23](https://user-images.githubusercontent.com/849592/157879816-1183e81d-0e5a-4674-9d07-d61b72c48c79.png)

jldec:
The image GET on the avatar url returns a 302 redirecting to the the login screen.
The CSS should be fixed to set a max size on the avatar width.

gtsiolis:
Nice catch, @jldec! ⚾

jankeromnes:
Will take a look today 😁

Issue #8815, opened 2022-03-15
Group and user avatars not loading for GitHub Enterprise new project repo selector 


jldec:
![Screenshot 2022-03-15 at 08 24 15](https://user-images.githubusercontent.com/849592/158336676-c533e010-66b8-4735-bd44-6d028fe32de5.png)

May be related to #5330

jankeromnes:
Will take a look today 😁

gtsiolis:
This is still relevant.

PR #8825, opened 2022-03-15
[server] Improve GitHub Enterprise avatars handling

jankeromnes:
## Description

GitHub Enterprise doesn't seem to allow cross-origin requests for user avatars, leading to a broken Gitpod UI.

Also, I initially wanted to fetch the avatars, and encode them as data URIs, however GitHub Enterprise also doesn't seem to have an API to download avatars (only browser authentication via cookies seems to work).

Thankfully, GitHub Enterprise does provide a Gravatar ID that we can use. This Pull Request does just that whenever possible, and also improves the display of avatars that cannot be found or loaded.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/8755
Partially improves https://github.com/gitpod-io/gitpod/issues/8815

## How to test

1. Connect with GitHub Enterprise
2. Open the org selector in `/new`
3. The avatars should work for all accounts/orgs that use Gravatar
4. Add a project and start a prebuild for a repo where the owner does **not** use Gravatar
5. In the Project Branches & Prebuilds pages, the broken avatar will show the alt text, but it should be truncated such that the rest of the details are still legible

## Release Notes

```release-note
[server] Improve GitHub Enterprise avatars handling
```

### DIFF
--- a/components/server/ee/src/prebuilds/github-service.ts
+++ b/components/server/ee/src/prebuilds/github-service.ts
@@ -32,8 +32,10 @@ export class GitHubService extends RepositoryService {
             return <ProviderRepository>{
                 name: r.name,
                 cloneUrl: r.clone_url,
-                account: r.owner?.login,
-                accountAvatarUrl: r.owner?.avatar_url,
+                account: r.owner.login,
+                accountAvatarUrl: r.owner.gravatar_id
+                    ? `https://www.gravatar.com/avatar/${r.owner.gravatar_id}?size=128`
+                    : r.owner.avatar_url,
                 updatedAt: r.updated_at,
             };
         });
--- a/components/dashboard/src/projects/Prebuilds.tsx
+++ b/components/dashboard/src/projects/Prebuilds.tsx
@@ -280,7 +280,7 @@ export default function (props: { project?: Project; isAdminDashboard?: boolean 
                                         <p>
                                             {p.info.changeAuthorAvatar && (
                                                 <img
-                                                    className="rounded-full w-4 h-4 inline-block align-text-bottom mr-2"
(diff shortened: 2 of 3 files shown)


jankeromnes:
Hmm, of course, neither our GHE user nor its org have Gravatars, so that doesn't help 😅

Screenshot 2022-03-15 at 15 14 05
Screenshot 2022-03-15 at 15 13 47

But I guess this is the best we can do. I'll also fix the avatar width in the workspaces list and call it done. 😕

components/server/ee/src/prebuilds/github-service.ts
```diff
@@ -32,8 +32,10 @@ export class GitHubService extends RepositoryService {
             return <ProviderRepository>{
                 name: r.name,
                 cloneUrl: r.clone_url,
-                account: r.owner?.login,
-                accountAvatarUrl: r.owner?.avatar_url,
+                account: r.owner.login,
+                accountAvatarUrl: r.owner.gravatar_id
+                    ? `https://www.gravatar.com/avatar/${r.owner.gravatar_id}?size=128`
```
gtsiolis, line 37: **issue**: This is still failing to load the avatar when there's no gravatar set, right?

| New project | Project branches |
|-|-|
| Screenshot 2022-03-15 at 8 00 55 PM | Screenshot 2022-03-15 at 8 00 43 PM |

gtsiolis (commented):
I was in the neighborhood and left one comment below.
Thanks for tackling this, @jankeromnes!

components/server/ee/src/prebuilds/github-service.ts
gtsiolis: Ah, I see.

jldec (approved):
merci @jankeromnes

PR #8827, opened 2022-03-15
Fix docker run --privileged

### LINKED ISSUE #2459: [docker] docker run --privileged does not work
### Describe the bug
Running anything with `docker run --privileged` fails because
```
ERRO[2020-12-08T13:21:18.928108663Z] Handler for POST /v1.40/containers/c9b7752a0e1f3a23baa7739b591814eb368c109cb6d0fa480b89ac7f783613c7/start returned error: OCI runtime create failed: container_linux.go:349: starting container process caused "process_linux.go:449: container init caused \"rootfs_linux.go:58: mounting \\\"cgroup\\\" to rootfs \\\"/var/lib/docker/vfs/dir/6e5f280f3ba91c9e9478248f1e69c8136ba7b82186adb72e8eee14253c9790d8\\\" at \\\"/sys/fs/cgroup\\\" caused \\\"operation not permitted\\\"\"": unknown 
```

This is in part because `/sys` is just a bind mount the workspace container's `/sys`, but also because we're using cgroups v1 (i.e. they're not namespaced).

### Steps to reproduce
```
sudo docker-up
docker run --privileged --rm -it alpine:latest
```

### Expected behavior
The container should start and run as root

Furisto:
## Description
Allow use of the privileged flag with docker. Previously this would cause an error. The fix is to ensure that the latest version of runc is installed.

## Related Issue(s)
Fixes #2459 

## How to test
- docker run --privileged -it busybox sh
--> No error should be happening

## Release Notes
```release-note
Allow use of the --privileged flag with docker.
```

### DIFF
--- a/components/docker-up/docker-up/main.go
+++ b/components/docker-up/docker-up/main.go
@@ -414,11 +414,6 @@ func installUidMap() error {
 }
 
 func installRunc() error {
-	_, exists := os.LookupEnv(DaemonArgs)
-	if !exists {
-		return nil
-	}
-
 	runc, _ := exec.LookPath("runc")
 	if runc != "" {
 		// if the required version or a more recent one is already


utam0k (approved):
So far, we were using an old runc(i.e., runc installed by image) and it didn't work.

components/docker-up/docker-up/main.go
```diff
@@ -414,11 +414,6 @@ func installUidMap() error {
 }
 
 func installRunc() error {
-	_, exists := os.LookupEnv(DaemonArgs)
```
kylos101, line 417: @Furisto I love when small code changes result in gains like this. Well done!

PR #8831, opened 2022-03-15
[dashboard] Remove Prebuild logs and button from Project Settings

easyCZ:
## Description

Removes the following:
* Prebuild log view from Project Configuration
* Ability to trigger prebuilds from Project configuration. See Before & After screenshots below.

### Old:
Screenshot 2022-03-16 at 16 33 42

### New:
Screenshot 2022-03-16 at 16 33 18

## How to test

1. Open up preview environment
2. Create a project
3. Navigate to project settings
4. Observe no toggle to run prebuilds or a log viewer

## Release Notes

```release-note
Remove Prebuild action and log view from Project Settings view
```

/hold

### DIFF
--- a/components/dashboard/src/projects/ConfigureProject.tsx
+++ b/components/dashboard/src/projects/ConfigureProject.tsx
@@ -5,16 +5,11 @@
  */
 
 import React, { Suspense, useContext, useEffect, useState } from "react";
-import { Project, StartPrebuildResult, WorkspaceInstance } from "@gitpod/gitpod-protocol";
-import PrebuildLogs from "../components/PrebuildLogs";
+import { Project } from "@gitpod/gitpod-protocol";
 import TabMenuItem from "../components/TabMenuItem";
 import { getGitpodService } from "../service/service";
 import Spinner from "../icons/Spinner.svg";
 import NoAccess from "../icons/NoAccess.svg";
-import PrebuildLogsEmpty from "../images/prebuild-logs-empty.svg";
-import PrebuildLogsEmptyDark from "../images/prebuild-logs-empty-dark.svg";
-import { ThemeContext } from "../theme-context";
-import { PrebuildInstanceStatus } from "./Prebuilds";
 import { ErrorCodes } from "@gitpod/gitpod-protocol/lib/messaging/error";
 import { openAuthorizeWindow } from "../provider-utils";
 import { ProjectSettingsPage } from "./ProjectSettings";
@@ -30,15 +25,6 @@ const TASKS = {
       echo 'TODO: start app'`,
 };
 
-// const IMAGES = {
-//   Default: 'gitpod/workspace-full',
-//   '.NET': 'gitpod/workspace-dotnet',
-//   MongoDB: 'gitpod/workspace-mongodb',
-//   MySQL: 'gitpod/workspace-mysql',
-//   PostgreSQL: 'gitpod/workspace-postgres',
-//   'Virtual Desktop (VNC)': 'gitpod/workspace-full-vnc',
-// }
-
 export default function () {
     const { project } = useContext(ProjectContext);
     const [gitpodYml, setGitpodYml] = useState<string>("");
@@ -47,12 +33,7 @@ export default function () {
     const [selectedEditor, setSelectedEditor] = useState<".gitpod.yml" | ".gitpod.Dockerfile">(".gitpod.yml");
     const [isEditorDisabled, setIsEditorDisabled] = useState<boolean>(true);
     const [isDetecting, setIsDetecting] = useState<boolean>(true);
-    const [prebuildWasTriggered, setPrebuildWasTriggered] = useState<boolean>(false);
-    const [prebuildWasCancelled, setPrebuildWasCancelled] = useState<boolean>(false);
-    const [startPrebuildResult, setStartPrebuildResult] = useState<StartPrebuildResult | undefined>();
-    const [prebuildInstance, setPrebuildInstance] = useState<WorkspaceInstance | undefined>();
-    const { isDark } = useContext(ThemeContext);
-
+    const [progressMessage, setProgressMessage] = useState<string>("");
     const [showAuthBanner, setShowAuthBanner] = useState<{ host: string; scope?: string } | undefined>(undefined);
     const [buttonNewWorkspaceEnabled, setButtonNewWorkspaceEnabled] = useState<boolean>(true);
 
@@ -108,7 +89,7 @@ export default function () {
                 <EditorMessage
                     type="warning"
                     heading="Configuration already exists in git."
-                    message="Run a prebuild or open a new workspace to edit project configuration."
+                    message="Open a new workspace to edit project configuration."
                 />,
             );
             setGitpodYml(repoConfigString);
@@ -128,7 +109,7 @@ export default function () {
                 <EditorMessage
                     type="success"
                     heading="Project type detected."
-                    message="You can edit project configuration below before running a prebuild"
+                    message="You can edit project configuration below before starting a workspace."
                 />,
             );
             setGitpodYml(guessedConfigString);
@@ -138,7 +119,7 @@ export default function () {
             <EditorMessage
                 type="warning"
                 heading="Project type could not be detected."
-                message="You can edit project configuration below before running a prebuild."
+                message="You can edit project configuration below before starting a workspace."
             />,
         );
         setGitpodYml(TASKS.Other);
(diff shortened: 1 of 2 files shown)


roboquat:

<details>

</details>

easyCZ:
@jldec Let me know if this matches your expectation for the prebuild removal.

AlexTugarev (dismissed):
~Works as advertised!~

Just confused with another PR of yours. This is next.

components/dashboard/src/projects/ConfigureProject.tsx
```diff
@@ -108,7 +89,7 @@ export default function () {
                 <EditorMessage
                     type="warning"
                     heading="Configuration already exists in git."
-                    message="Run a prebuild or open a new workspace to edit project configuration."
+                    message="Oopen a new workspace to edit project configuration."
```
AlexTugarev, line 92: typo

easyCZ: Thanks. Fixed now.

gtsiolis:
Taking a look now! 👀

components/dashboard/src/projects/ConfigureProject.tsx
```diff
@@ -338,56 +255,10 @@ export default function () {
...
-                                Run Prebuild
-                            </button>
-                        )}
-                        <button disabled={isDetecting && buttonNewWorkspaceEnabled} onClick={onNewWorkspace}>
-                            New Workspace
-                        </button>
-                    </div>
+                <div className="h-20 px-6 bg-gray-50 dark:bg-gray-800 border-gray-200 dark:border-gray-600 flex space-x-2">
+                    <button disabled={isDetecting && buttonNewWorkspaceEnabled} onClick={onNewWorkspace}>
+                        New Workspace
```
gtsiolis, line 260: **suggestion**: What do you think of moving the action button to the right following the pattern we already use across the product for having the primary action on the outer right side of things?

| BEFORE | AFTER |
|-|-|
| config-before | config-after |

easyCZ:
Thanks for the feedback @gtsiolis. I've made the changes you've suggested. Currently looks like this:
Screenshot 2022-03-29 at 12 28 57

gtsiolis (dismissed):
Thanks, @easyCZ! UX changes LGTM! 🏁

jldec:
@easyCZ it appears that the behavior of the config editor is different now (when there is no .gitpod.yml in the repo)

- before: editing the suggested config would save the changes and they appear in an uncommitted gitpod.yml when the workspace is started.

- now: any edits appear to be lost or ignored - this makes it quite confusing.

I was expecting the save-edits behavior to be preserved.

gtsiolis:
Can confirm the issue @jldec described in https://github.com/gitpod-io/gitpod/pull/8831#issuecomment-1081740517. ❗ 

Plus, the initial auto-generated configuration (see code below) is lost when the project type could not be detected.

```yml
tasks:
  - init: |
      echo 'TODO: build project'
    command: |
      echo 'TODO: start app'
```

easyCZ:
I see where the problem is. Originally, if you edited the `gitpod.yml` and clicked Start Workspace, it also wouldn't be preserved. However, if you first prebuilt, it would as it sets the `gitpod.yml` for the project before it runs the prebuild.

Will dig into it. Moving back to draft as it's a regression at this stage.

easyCZ:
@gtsiolis @jldec I've now pushed a fix for the issue you've discovered, thanks! I'm moving this back into review.

The behaviour is now as follows when there is no `gitpod.yml` in the repo:
1. Edit the config
2. Click `New Workspace` which saves the config.

This doesn't feel like the most intuitive, but matches the original behaviour (which was `Run prebuild` rather `New Workspace`).
We could explicitly introduce a button to save the config, which may be nicer.

gtsiolis:
Thanks, @easyCZ! Works as before now.

> This doesn't feel like the most intuitive, but matches the original behaviour

Agree, this is also a bit confusing as we auto-save but not fast enough to include all latest changes in the updated and injected `.gitpod.yml` file. Maybe a subtle auto-save visual feedback on the editor along with a configuration validation action could help. 💭 

**issue**: One seemingly important issue but probably not worth blocking merging this is that now when the configuration is invalid we're blocking the users from starting a new workspace without any error or visual feedback as we did with when having the ability to trigger a prebuild, see screenshot below. ❗ 

| (BEFORE) Invalid configuration (Run Prebuild) | (AFTER) Invalid configuration (New Workspace) | 
|-|-|
| Screenshot 2022-03-29 at 3 23 12 PM | Screenshot 2022-03-29 at 3 20 43 PM |

easyCZ:
I'm digging into this some more just now. It looks like for this behaviour to remain the same, the following must happen in order:
1. Click start workspace
2. The current `gitpod.yml` config is stored on the server
3. A prebuild is triggered -> this then uses the previously saved gitpod.yml
4. Redirect to the context URL to start a new workspace. Because there was a prebuild in step 3, it will show the prebuild steps.

In the current implementation, we're not triggering the prebuild and as such the workspace start drops you into the IDE with the `before` and `init` tasks being run in the current terminal.

The fix is to re-add the trigger of a prebuild when you're starting a workspace. However, now with the prebuild view gone, we don't have a place to report progress. How do you want to handle this @jldec?

easyCZ:
I've now updated this PR with the following:
* Triggering a new workspace saves the config and starts a prebuild, once the prebuild request is accepted, we redirect to the workspace detail page where the prebuild logs will show
* The button becomes disabled once clicked and shows the status - "Saving configration..." or "Starting a prebuild"

Once staging with new images becomes usable again, I'll put this back into review.

easyCZ:
Screenshot 2022-04-04 at 16 56 38

jldec (approved):
LGTM
Thanks for starting the cleanup
I think this is good for now - I expect we will remove this dashboard editor feature entirely when we deprecate the config in DB with #9102

/hold for other reviews

gtsiolis:
Looking at this now! 👀

components/dashboard/src/projects/Prebuilds.tsx
```diff
@@ -386,113 +381,3 @@ export function PrebuildStatus(props: { prebuild: PrebuildWithStatus }) {
...
-            status = (
-                <div className="flex space-x-1 items-center text-blue-600">
-                    <img alt="" className="h-4 w-4" src={StatusRunning} />
-                    <span>RUNNING</span>
-                </div>
-            );
-            details = (
-                <div className="flex space-x-1 items-center text-gray-400">
-                    <img alt="" className="h-4 w-4 animate-spin" src={Spinner} />
-                    <span>Prebuild in progress ...</span>
```
gtsiolis, line 426: **suggestion**: Probably not the best line to comment this and sounds like out of the scope of this PR but what do you think of changing the prebuild phase of the workspace start page by reusing the prebuild elements? Cc @jankeromnes @AlexTugarev because of relevant past discusions. Probably also best as a follow up issue.

| BEFORE | AFTER |
|-|-|
| Screenshot 2022-04-05 at 2 00 24 PM | Screenshot 2022-04-05 at 2 09 51 PM |

gtsiolis (approved):
UX LGTM.

Left some minor comments below, some of these probably could be best as follow up issues.

Thanks for the ping, @easyCZ! 🏓

components/dashboard/src/projects/ConfigureProject.tsx
```diff
@@ -176,66 +157,39 @@ export default function () {
...
-    };
+        if (!isEditorDisabled) {
+            try {
+                setProgressMessage("Saving project configuration...");
+                await getGitpodService().server.setProjectConfiguration(project.id, gitpodYml);
 
-    useEffect(() => {
-        document.title = "Configure Project — Gitpod";
-    }, []);
+                setProgressMessage("Starting a new prebuild...");
```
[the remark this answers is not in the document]
easyCZ: Good point! Related to the other point raised around:
> question: Is there a limitation that does not allow us to simply redirect to the workspace start page which automatically surfaces when a prebuild is running?

Will investigate!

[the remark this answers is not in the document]
easyCZ: In my testing, starting a prebuild was required to actually use the new config. It wouldn't get picked up after the save without the prebuild. However, this may be a red-herring. I need to dig into the config & prebuild flow a bit more to understand why this is, and if in fact this is the case.

Will investigate!

components/dashboard/src/projects/Prebuilds.tsx
easyCZ: This makes sense, would definitely help with increased perceived reliability as the UI would be consistent. I've cut an issue for this. https://github.com/gitpod-io/gitpod/issues/9132

gtsiolis: Je vous remercie, @easyCZ! ✨

components/dashboard/src/projects/ConfigureProject.tsx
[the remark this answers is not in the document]
easyCZ: I've attempted to remove the prebuild step but the saved configuration is not reflected in the workspace. I will continue to investigate but in the meantime, this PR is functionally complete and will land.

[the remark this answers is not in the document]
gtsiolis: Sounds good! Thanks for taking a closer look, @easyCZ!

PR #8833, opened 2022-03-16
[server] always use `.` for image builder checkoutLocation

iQQBot:
## Description

all preview environment can't build image, this is because
https://github.com/gitpod-io/gitpod/blob/45f13cf8e9214db664adc23a418e53d92364fbb7/components/server/src/workspace/workspace-starter.ts#L721-L735
L721 using workspace.context.checkoutLocation and L735 is . 

In image build stage we don't care checkoutLocation, this PR always use `.` for checkoutLocation
## Related Issue(s)

Fixes #

## How to test

1. start a workspace which need image build, like `https://github.com/iQQBot/test.test` you will see image builder is work
2. and start another workspace check everything is like before

## Release Notes

```release-note
Fixes image builder not work
```


geropl:
@iQQBot I will have a look later today 


Please

PR #8844, opened 2022-03-16
[kots] Add a pre-flight check for `cert-manager`

### LINKED ISSUE #8823: Add pre-flight check to verify cert-manager validity
There have been various discussions with Barry@Replicated about whether it's possible (and wise) about getting a pre-flight check to ensure that cert-manager is present in the cluster.

As cert-manager is a requirement, it is something that we should check. Adding a KOTS pre-flight check to verify this (similar to in the [Installer](https://github.com/gitpod-io/gitpod/blob/main/install/installer/pkg/cluster/validation.go#L74)) is wise.

There should be two checks:

1. Ensure that cert-manager is detected. This could be achieved as-per the Installer where the `Certificate` CRD is checked that it exists. **Fail if not present.**
2. Ensure that the desired Issuer or ClusterIssuer is present. This should only be checked if this is included as part of the configuration. **Fail if not present**

## Limitations

It's unlikely at this stage that we can actually check whether the resource is valid and can create a certificate. That may come as a future improvement

Pothulapati:
## Description

With https://github.com/gitpod-io/gitpod/pull/8785, We expect users
to install their own `cert-manager` install. This PR adds a new
pre-flight check to verify the existence of the `cert-manager`.

As there is no general way to check for the existence of `cert-manager`,
We check for the `certificates.cert-manager.io` CRD instead.

## Related Issue(s)

Fixes #8823 

## How to test

Install it by running
```
k kots install gitpod-pov/dev-tar
```

and you should see

image

## Release Notes

```release-note
[kots] Add a pre-flight check for `cert-manager`
```

### DIFF
--- a/install/kots/manifests/kots-preflight.yaml
+++ b/install/kots/manifests/kots-preflight.yaml
@@ -95,3 +95,11 @@ spec:
               message: The cluster must contain at least 4 cores
           - pass:
               message: There are at least 2 cores in the cluster
+    - customResourceDefinition:
+        checkName: Cert-manager is installed
+        customResourceDefinitionName: certificates.cert-manager.io
+        outcomes:
+          - fail:
+              message: Custom resource definition `certificates.cert-manager.io` was not found in the cluster. Please [install `cert-manager`](https://cert-manager.io/docs/installation/) to proceed.
+          - pass:
+              message: Cert-manager is installed and available.


Pothulapati:
Reg

> Ensure that the desired Issuer or ClusterIssuer is present. This should only be checked if this is included as part of the configuration. Fail if not present

@MrSimonEmms Currently, From the docs, I don't think there is a intitutive/simple way to do this. We can use the exec functionality to run a command (in some pod that has access) and get the data and verify but that seems non-intutive, etc. Though the doc mentions about [cluster resources](https://troubleshoot.sh/docs/collect/cluster-resources/), FWIU I don't think it supports custom resource objects.

install/kots/manifests/kots-preflight.yaml
```diff
@@ -95,3 +95,10 @@ spec:
               message: The cluster must contain at least 4 cores
           - pass:
               message: There are at least 2 cores in the cluster
+    - customResourceDefinition:
+        customResourceDefinitionName: certificate.cert-manager.io
```
MrSimonEmms, line 99:
```suggestion
        customResourceDefinitionName: certificates.cert-manager.io
```

install/kots/manifests/kots-preflight.yaml
```diff
@@ -95,3 +95,10 @@ spec:
...
+        outcomes:
+          - fail:
+              message: Certificate CRD of Cert Manager was not found in the cluster. Please [install `cert-manager`](https://cert-manager.io/docs/installation/) to proceed.
```
MrSimonEmms, line 102:
```suggestion
              message: Custom resource definition `certificates.cert-manager.io` was not found in the cluster. Please [install `cert-manager`](https://cert-manager.io/docs/installation/) to proceed.
```

install/kots/manifests/kots-preflight.yaml
```diff
@@ -95,3 +95,10 @@ spec:
...
+          - pass:
+              message: Cert Manager is installed and available.
```
MrSimonEmms, line 104:
```suggestion
              message: Cert-manager is installed and available.
```

This we should use their naming convention of "cert-manager"

install/kots/manifests/kots-preflight.yaml
MrSimonEmms, line 99: Consider adding `checkName: Cert-manager is installed` to give a friendly name to the check

MrSimonEmms (commented):
Couple of suggested changes. The check works well otherwise 👍🏻

MrSimonEmms (approved):

Issue #8843, opened 2022-03-16
[agent-smith] Concurrent map write during start


aledbf:
### Bug description

```
{"level":"info","message":"procfs detector started","serviceContext":{"service":"agent-smith","version":"commit-526ed855bb509e6523f1318a7f207a55cefdb054"},"severity":"INFO","time":"2022-03-16T17:18:38Z"}
fatal error: concurrent map read and map write

goroutine 31 [running]:
runtime.throw({0x17950a1, 0x1})
	runtime/panic.go:1198 +0x71 fp=0xc000560e40 sp=0xc000560e10 pc=0x436b31
runtime.mapaccess2_fast64(0xc0002a7d00, 0xc00059c150, 0xa485)
	runtime/map_fast64.go:61 +0x176 fp=0xc000560e60 sp=0xc000560e40 pc=0x411536
github.com/gitpod-io/gitpod/agent-smith/pkg/agent.(*Smith).Start.func1()
	github.com/gitpod-io/gitpod/agent-smith/pkg/agent/agent.go:231 +0x175 fp=0xc000560fe0 sp=0xc000560e60 pc=0x13c0815
runtime.goexit()
	runtime/asm_amd64.s:1581 +0x1 fp=0xc000560fe8 sp=0xc000560fe0 pc=0x46a701
created by github.com/gitpod-io/gitpod/agent-smith/pkg/agent.(*Smith).Start
	github.com/gitpod-io/gitpod/agent-smith/pkg/agent/agent.go:222 +0x2b4

goroutine 1 [select, 17 minutes]:
github.com/gitpod-io/gitpod/agent-smith/cmd.glob..func3(0x2570cc0, {0x1776031, 0x2, 0x2})
	github.com/gitpod-io/gitpod/agent-smith/cmd/run.go:107 +0xbff
github.com/spf13/cobra.(*Command).execute(0x2570cc0, {0xc00053c0c0, 0x2, 0x2})
	github.com/spf13/cobra@v1.1.3/command.go:856 +0x5f8
github.com/spf13/cobra.(*Command).ExecuteC(0x2570a40)
	github.com/spf13/cobra@v1.1.3/command.go:960 +0x3ad
github.com/spf13/cobra.(*Command).Execute(...)
	github.com/spf13/cobra@v1.1.3/command.go:897
github.com/gitpod-io/gitpod/agent-smith/cmd.Execute()
[...]

PR #8849, opened 2022-03-17
[agent-smith] lock workspaces map before read

princerachit:
## Description

The go map does not allow concurrent access i.e. both read and write. So, take lock in the following scenarios:
1. Before reading an entry from the map
2. Before ranging over the map
3. Before updating an entry in the map

### Impact
Since both process classification and egress checks are performed in the same go routine, repeated locking attempt before reading will cause a delay in execution of classification method.

The contention impact on performance can be evaluated when this runs in production.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/8843

## How to test

NA

## Release Notes

```release-note
None
```

### DIFF
--- a/components/ee/agent-smith/pkg/agent/agent.go
+++ b/components/ee/agent-smith/pkg/agent/agent.go
@@ -214,7 +214,7 @@ func (agent *Smith) Start(ctx context.Context, callback func(InfringingWorkspace
 	agent.metrics.RegisterClassificationQueues(cli, clo)
 
 	workspaces := make(map[int]*common.Workspace)
-	wssWriteMutex := &sync.Mutex{}
+	wsMutex := &sync.Mutex{}
 
 	defer wg.Wait()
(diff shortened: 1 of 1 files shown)


utam0k (approved):

components/ee/agent-smith/pkg/agent/agent.go
```diff
@@ -214,7 +214,7 @@ func (agent *Smith) Start(ctx context.Context, callback func(InfringingWorkspace
 	agent.metrics.RegisterClassificationQueues(cli, clo)
 
 	workspaces := make(map[int]*common.Workspace)
-	wssWriteMutex := &sync.Mutex{}
+	wsMutex := &sync.Mutex{}
```
kylos101, line 217: Hey there, you may also be able to try this `sync.RWMutex`, it should allow for a single writer and one or more readers.

@princerachit @utam0k

PR #8853, opened 2022-03-17
[server] Finish spans

geropl:
## Description
Finish dangling spans

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/server/src/workspace/workspace-starter.ts
+++ b/components/server/src/workspace/workspace-starter.ts
@@ -451,12 +451,19 @@ export class WorkspaceStarter {
 
     protected async notifyOnPrebuildQueued(ctx: TraceContext, workspaceId: string) {
         const span = TraceContext.startSpan("notifyOnPrebuildQueued", ctx);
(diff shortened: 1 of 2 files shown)


components/server/src/workspace/workspace-starter.ts
```diff
@@ -1389,27 +1398,34 @@ export class WorkspaceStarter {
         user: User,
     ): Promise<{ initializer: GitInitializer | CompositeInitializer; disposable: Disposable }> {
         const span = TraceContext.startSpan("createInitializerForCommit", ctx);
```
jankeromnes, line 1400: Shouldn't spans have the same name as their method? (If they should, we could probably enforce this with a custom linting rule):

```suggestion
        const span = TraceContext.startSpan("createCommitInitializer", ctx);
```

jankeromnes (approved):
Looks good to me, many thanks! 👍🧹

Left one question about pre-existing code, otherwise good for merge.

PR #8858, opened 2022-03-17
[gitpod-db] add index on `workspaceDB.Type` Column

### LINKED ISSUE #8638: [telemetry] Fix telemetry causing high DB CPU load
Currently we're executing an [expensive query](https://github.com/gitpod-io/gitpod/blob/29df192bd0463473390716567fad2985d696f43b/components/gitpod-db/src/typeorm/workspace-db-impl.ts#L337-L344) more often than we expect. This caused 3 incidents over the course of the last 2,5 weeks. We still are not sure why it's triggered this often (there are no traces/logs), but still we can improve the situation by:

 1. DB:
   1. improve the query itself: our hypothesis is that the ORM generates a query like this `SELECT COUNT(1) FROM (SELECT ... JOIN...)` where the subquery is the reason for the slowness (MySQL tries to materialize the table). Try writing direct SQL ala `SELECT COUNT(1) FROM d_b_workspace_instance AS wsi JOIN d_b_workspace AS ws ON ws.id = wsi.workspaceId WHERE ws.type = 'regular'. Test this against a failover prod DB.
   2. double-check we have an index on `workspace.type` _in both prod DBs_
 2. API: use different API/HTTP calls/requests for "config" and "telemetry data": e.g., don't execute the queries in case we are not sending the result anyway
 3. better observability: add tracing to the [HTTP endpoint](https://github.com/gitpod-io/gitpod/blob/3be4e0b7a56a1249ff00c015efd28a5ce715c161/components/server/src/installation-admin/installation-admin-controller.ts#L17-L20)

/cc @corneliusludmann

Pothulapati:
## Description

For Telemetry queries to be less intensive on the CPU, We need to
build Indexes to make the querying faster. This PR creates
a index on the `type` column for the `workspaceDB` as
we filter on `type=='regular'`

## Related Issue(s)

Fixes #8638 

## How to test

After creating the index on prod, The query seems to be faster 
```
mysql> select count(*) from d_b_workspace_instance join d_b_workspace on d_b_workspace_instance.workspaceId = d_b_workspace.id where d_b_workspace.type = 'regular';
+----------+
| count(*) |
+----------+
|  5665695 |
+----------+
1 row in set (11.80 sec)
```

## Release Notes

```release-note
[gitpod-db] add index on `workspaceDB.Type` Column
```

### DIFF
--- a/components/gitpod-db/src/typeorm/entity/db-workspace.ts
+++ b/components/gitpod-db/src/typeorm/entity/db-workspace.ts
@@ -68,6 +68,7 @@ export class DBWorkspace implements Workspace {
     })
     shareable?: boolean;
 
+    @Index("ind_type")
     @Column({
         default: "regular",
     })
--- a/components/gitpod-db/src/typeorm/migration/1647290507971-IndexWorkspaceType.ts
+++ b/components/gitpod-db/src/typeorm/migration/1647290507971-IndexWorkspaceType.ts
@@ -1,0 +1,23 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import {MigrationInterface, QueryRunner} from "typeorm";
+import { indexExists } from "./helper/helper";
+
+export class IndexWorkspaceType1647290507971 implements MigrationInterface {
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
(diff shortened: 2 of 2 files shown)


components/gitpod-db/src/typeorm/entity/db-workspace.ts
```diff
@@ -68,6 +68,7 @@ export class DBWorkspace implements Workspace {
     })
     shareable?: boolean;
 
+    @Index("ind_type")
```
geropl, line 71: :+1:

geropl (approved):
Tested and works as expected :+1:

PR #8871, opened 2022-03-17
[kots]: remove wait-for-jobs

MrSimonEmms:
## Description

Remove `wait-for-jobs` flag in Helm.

As we're deleting jobs on completion, the Helm command can error
wrongly if the job is deleted before Helm detects it's completed.

## How to test

Deploy via KOTS

## Release Notes

```release-note
[kots]: remove wait-for-jobs
```

### DIFF
--- a/install/kots/manifests/gitpod-installer-job.yaml
+++ b/install/kots/manifests/gitpod-installer-job.yaml
@@ -191,7 +191,6 @@ spec:
                 --reset-values \
                 --timeout 1h \
                 --wait \
-                --wait-for-jobs \
                 gitpod \
                 "${GITPOD_OBJECTS}"
 


install/kots/manifests/gitpod-installer-job.yaml
```diff
@@ -191,7 +191,6 @@ spec:
                 --reset-values \
                 --timeout 1h \
                 --wait \
-                --wait-for-jobs \
```
nandajavarma, line 194: nit: Just wondering if this would falsely give a "Installer job finished" message. Is increasing the `ttl` of the job a possibility?

MrSimonEmms: I don't think so. The jobs are decoupled from the install/upgrade process really. Even if it does fail, it doesn't allow us to feedback to the GUI (I don't think KOTS has anything that allows that in fact)

nandajavarma (approved):
LGTM!

PR #8873, opened 2022-03-17
Avoid using process.exit as we need to flush telemetry

mads-hartmann:
## Description

We don't always get traces from gitpod-io/gitod when the main build fails (see https://github.com/gitpod-io/ops/issues/865 for more context).

We're relying on Node's beforeExit event to ensure that we flush our traces before stopping the node process (see [here](https://github.com/gitpod-io/gitpod/blob/main/.werft/observability/tracing.ts#L41)). This means that we can't use process.exit in our TS scripts, as that doesn't invoke the beforeExit handlers.

We have previously gotten rid of `process.exit` invocations but had overlooked one. This PR removes it. Previously `reportBuildFailureInSlack` took an onErr function which it invoked once the request to slack completed (or failed). We passed in `() => process.exit(1)`. I've removed the onErr argument and converted it to a promise-based API instead. We now don't do anything if the Slack requests succeeds but do log the error if it fails. The process.exitCode is moved out of the else-branch so it's always set if the build fails.

## Related Issue(s)

Fixes https://github.com/gitpod-io/ops/issues/865

## How to test

I haven't tested this as I couldn't find an easy way to test it. Given the code change is rather trivial (though that can be deceiving) I'm happier merging and reverting if it turns out I broke the Slack notification ☺️

## Release Notes

```release-note
NONE
```

## Documentation

N/A

### DIFF
--- a/.werft/build.ts
+++ b/.werft/build.ts
@@ -28,12 +28,15 @@ Tracing.initialize()
         })
 
         if (context.Repository.ref === "refs/heads/main") {
-            reportBuildFailureInSlack(context, err, () => process.exit(1));
+            reportBuildFailureInSlack(context, err).catch((error: Error) => {
+                console.error("Failed to send message to Slack", error)
+            });
         } else {
             console.log('Error', err)
-            // Explicitly not using process.exit as we need to flush tracing, see tracing.js
-            process.exitCode = 1
         }
+
+        // Explicitly not using process.exit as we need to flush tracing, see tracing.js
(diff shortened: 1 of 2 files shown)


mads-hartmann:
@geropl I requested review from you as well you have touched this code before ☺️

ArthurSens (approved):
/hold

LGTM, but I'm adding a hold in case you're waiting for Gero's review

Harry-Hopkinson (approved):

PR #8881, opened 2022-03-18
[installer]: handle different line endings on apiVersion extraction

MrSimonEmms:
## Description

I've just found that, even on Linux, you can get Windows line endings. This removes the `\n` from the `apiVersion` extraction and makes the comma optional

## How to test

Generate config and render with Installer

## Release Notes

```release-note
NONE
```


nandajavarma (approved):
LGTM!

PR #8884, opened 2022-03-18
.werft/build: Refactor installer steps into dedicated class

ArthurSens:
## Description

This PR tries to reduce amount of code under `.werft/jobs/build/deploy-to-preview-environments.ts` by moving all code related to installer to a dedicated `Installer` interface.

That way, we can make the deploy phase as simple as:
```typescript
werft.phase(deploy)
const installer = newInstaller(params...)
try {
        installer.init(installerSlices.INSTALLER_INIT)
        installer.addPreviewConfiguration(installerSlices.PREVIEW_CONFIG)
        installer.validateConfiguration(installerSlices.VALIDATE_CONFIG)
        installer.render(installerSlices.INSTALLER_RENDER)
        installer.postProcessing(installerSlices.INSTALLER_POST_PROCESSING)
        installer.install(installerSlices.APPLY_INSTALL_MANIFESTS)
    } catch (err) {
        if (!jobConfig.mainBuild) {
            werft.fail(phases.DEPLOY, err)
        }
        exec('exit 0')
    }
werft.done(deploy)
```

I'm opening it as a draft just to collect early feedback on the approach. @gitpod-io/platform do you agree with the approach here?

## How to test

- [x] Start a preview on core-dev and make sure workspaces start
- [x] Start a preview on Harvester and make sure workspaces start

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.werft/jobs/build/installer/installer.ts
+++ b/.werft/jobs/build/installer/installer.ts
@@ -1,0 +1,237 @@
+import { exec } from "../../../util/shell";
+import { getGlobalWerftInstance } from "../../../util/werft";
+import { getNodePoolIndex } from "../deploy-to-preview-environment";
+
+const blockNewUserConfigPath = './blockNewUsers';
+const workspaceSizeConfigPath = './workspaceSizing';
+const PROJECT_NAME = "gitpod-core-dev";
+const CONTAINER_REGISTRY_URL = `eu.gcr.io/${PROJECT_NAME}/build/`;
+const CONTAINERD_RUNTIME_DIR = "/var/lib/containerd/io.containerd.runtime.v2.task/k8s.io";
+
+export interface Installer {
+    configPath: string,
+    version: string,
+    proxySecretName: string
+    domain: string
+    previewName: string
+    imagePullSecretName: string
+    deploymentNamespace: string
+    analytics: Analytics,
+    withEELicense: boolean,
+    withVM: boolean,
+    workspaceFeatureFlags: string[],
+    gitpodDaemonsetPorts: GitpodDaemonsetPorts
+
+    init(slice: string): void
+    addPreviewConfiguration(slice: string): void
+    validateConfiguration(slice: string)
+    render(slice: string): void
+    postProcessing(slice: string): void
+    install(slice: string): void
+}
+
+export interface Analytics {
+    type: string,
+    token: string
+}
+
+export interface GitpodDaemonsetPorts {
+    registryFacade: number,
+    wsDaemon: number,
+}
+
+export function newInstaller(configPath: string, version: string, proxySecretName: string, domain: string, previewName: string, imagePullSecretName: string, deploymentNamespace: string, analytics: Analytics, withEELicense: boolean, withVM: boolean, workspaceFeatureFlags: string[], gitpodDaemonsetPorts: GitpodDaemonsetPorts): Installer {
+
+    const init = function init(slice) {
+        const werft = getGlobalWerftInstance()
+        werft.log(slice, "Downloading installer and initializing config file");
+        exec(`docker run --entrypoint sh --rm eu.gcr.io/gitpod-core-dev/build/installer:${this.version} -c "cat /app/installer" > /tmp/installer`, { slice: slice });
+        exec(`chmod +x /tmp/installer`, { slice: slice });
+        exec(`/tmp/installer init > ${this.configPath}`, { slice: slice });
+        werft.done(slice);
+    }
+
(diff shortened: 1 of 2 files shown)


ArthurSens:
Tested on both core-dev and Harvester, so turning it into ready for review 🙂

.werft/jobs/build/installer/installer.ts
```diff
@@ -0,0 +1,237 @@
...
+import { getGlobalWerftInstance } from "../../../util/werft";
+import { getNodePoolIndex } from "../deploy-to-preview-environment";
+
+const blockNewUserConfigPath = './blockNewUsers';
+const workspaceSizeConfigPath = './workspaceSizing';
+const PROJECT_NAME = "gitpod-core-dev";
+const CONTAINER_REGISTRY_URL = `eu.gcr.io/${PROJECT_NAME}/build/`;
+const CONTAINERD_RUNTIME_DIR = "/var/lib/containerd/io.containerd.runtime.v2.task/k8s.io";
+
+export interface Installer {
```
mads-hartmann, line 11: In Typescript you generally only use an interface if you intend there to be multiple implementations of the interface. In this case we only have the one so it is preferred to instead just export a class and make your `newInstaller` function the constructor of the class.

.werft/jobs/build/installer/installer.ts
```diff
@@ -0,0 +1,237 @@
...
+
+    init(slice: string): void
+    addPreviewConfiguration(slice: string): void
+    validateConfiguration(slice: string)
+    render(slice: string): void
+    postProcessing(slice: string): void
+    install(slice: string): void
+}
+
+export interface Analytics {
```
mads-hartmann, line 33: I'd use `type` here instead of interface. This is describing the shape of some data, so we don't intend to use `extends` or `implements` for this ☺️

.werft/jobs/build/installer/installer.ts
```diff
@@ -0,0 +1,237 @@
...
+        validateConfiguration,
+        render,
+        postProcessing,
+        install,
+    }
+
+    return installer
+}
+
+function getDevCustomValues(slice: string, configPath: string): void {
```
mads-hartmann, line 139: All of these functions could be private function or the class or you could leave them here. I'd prefer making them private functions on the class so you can access `this.werft`

.werft/jobs/build/installer/installer.ts
```diff
@@ -0,0 +1,237 @@
...
+
+export interface GitpodDaemonsetPorts {
+    registryFacade: number,
+    wsDaemon: number,
+}
+
+export function newInstaller(configPath: string, version: string, proxySecretName: string, domain: string, previewName: string, imagePullSecretName: string, deploymentNamespace: string, analytics: Analytics, withEELicense: boolean, withVM: boolean, workspaceFeatureFlags: string[], gitpodDaemonsetPorts: GitpodDaemonsetPorts): Installer {
+
+    const init = function init(slice) {
+        const werft = getGlobalWerftInstance()
```
mads-hartmann, line 46: I would make `werft` a private field on the class that you pass in the constructor. That was you can use `this.werft` everywhere in the class.

mads-hartmann (changes requested):
Awesome Arthur. I think this could be improved further, I have left some comments in the review. I'm happy to talk through this more if that's helpful ☺️

mads-hartmann (dismissed):
I left a few comments that you can follow up on in a later PR ☺️ I added the hold label in case you needed to do some last testing before merging

.werft/jobs/build/installer/installer.ts
```diff
@@ -0,0 +1,218 @@
...
+    previewName: string
+    imagePullSecretName: string
+    deploymentNamespace: string
+    analytics: Analytics
+    withEELicense: boolean
+    withVM: boolean
+    workspaceFeatureFlags: string[]
+    gitpodDaemonsetPorts: GitpodDaemonsetPorts
+
+    constructor(werft: Werft, configPath: string, version: string, proxySecretName: string, domain: string, previewName: string, imagePullSecretName: string, deploymentNamespace: string, analytics: Analytics, withEELicense: boolean, withVM: boolean, workspaceFeatureFlags: string[], gitpodDaemonsetPorts: GitpodDaemonsetPorts) {
```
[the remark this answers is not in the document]
ArthurSens: Ah nice, I really wasn't enjoying that method signature 😅 

I'll change that real quick

mads-hartmann (approved):
Awesome, looks great 🛹

Issue #7016, opened 2021-12-02
[gp cli] tasks support


akosyakov:
[Gitpod CLI](https://github.com/gitpod-io/gitpod/tree/main/components/gitpod-cli) should provide following commands to access tasks:
- `gp task list` to list [workspace tasks](https://github.com/gitpod-io/gitpod/blob/fc0c0a613c354444056ff43dfcc55c5acc9052a6/components/supervisor-api/status.proto#L65-L75) and their state
- `go task attach <id>` to attach to a workspace task by [an id](https://github.com/gitpod-io/gitpod/blob/fc0c0a613c354444056ff43dfcc55c5acc9052a6/components/supervisor-api/status.proto#L200)

In order to attach we should use [supervisor terminal service](https://github.com/gitpod-io/gitpod/blob/fc0c0a613c354444056ff43dfcc55c5acc9052a6/components/supervisor-api/terminal.proto#L14) similar how it is done [here](https://github.com/gitpod-io/gitpod/blob/main/components/supervisor/cmd/terminal-attach.go) but always with resizable and interactive session.

axonasif:
> gp task attach <id> to attach to a workspace task by an id

So this will be something like `tmux` I assume.
For that case I have a suggestion, I think it would be better to have a separate UI for the gitpod tasks, more like how the `Remote Explorer` works on Gitpod.

akosyakov:
> For that case I have a suggestion, I think it would be better to have a separate UI for the gitpod tasks, more like how the Remote Explorer works on Gitpod.

It is for terminal users via SSH. Generally whatever we have in UI should be reflected in Gitpod CLI as well.

In VS Code we don't need UI, gitpod tasks are automatically opened in terminals on workspace startup.

axonasif:
> It is for terminal users via SSH and for IDEs there we don't provide deep integration yet, i.e. JetBrains products. Generally whatever we have in UI should be reflected in Gitpod CLI as well.

I agree, I like to do most of my tasks from cli when possible.

> In VS Code we already have UI, Gitpod tasks are automatically opened in terminals on workspace startup.

Here I was thinking about an UI for the detached background gitpod `command` tasks in the web version of Gitpod. But if course, cli should be there too.

axonasif:
[...]

PR #8890, opened 2022-03-18
feat(gitpod-cli): add tasks cmd

andreafalzetti:
## Description
Adds a command with 2 sub-commands to gitpod-cli.

- gp tasks list
- gp tasks attach

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/7016

## Release Notes

```release-note
```

### DIFF
--- a/components/gitpod-cli/pkg/supervisor/terminal-attach.go
+++ b/components/gitpod-cli/pkg/supervisor/terminal-attach.go
@@ -1,0 +1,129 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package supervisor
+
+import (
+	"context"
+	"io"
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/creack/pty"
+	"github.com/gitpod-io/gitpod/supervisor/api"
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/term"
+)
+
+type AttachToTerminalOpts struct {
+	Interactive bool
+	ForceResize bool
+	Token       string
+}
+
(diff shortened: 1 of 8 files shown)


roboquat:

<details>

</details>

components/gitpod-cli/pkg/supervisor/terminal-attach.go
```diff
@@ -0,0 +1,129 @@
...
+	"golang.org/x/term"
+)
+
+type AttachToTerminalOpts struct {
+	Interactive bool
+	ForceResize bool
+	Token       string
+}
+
+func AttachToTerminal(ctx context.Context, client api.TerminalServiceClient, alias string, opts AttachToTerminalOpts) {
```
andreafalzetti, line 26: This is the same as https://github.com/gitpod-io/gitpod/blob/5934558596cebbfd179baa897b9126bec66a4c3c/components/supervisor/cmd/terminal-attach.go#L79

We agreed with @akosyakov to duplicate the logic instead of trying to share it between the two CLIs.

andreafalzetti: Also, this function comes with a problem - if a terminal attaches to itself, it will go into a loop. In tasks cmd, we try avoiding entering in that situation but it's not easy to detect.

felladrin (commented):
I've checked and tested it, and it looks great!

I think this could even be used in https://github.com/gitpod-io/gitpod/pull/8338

components/gitpod-cli/pkg/supervisor/terminal-attach.go
felladrin: Maybe, in the context of the running terminal, there could be an environment variable indicating it's a Task Terminal, so the attach command could check this env variable, and if detected someone trying run `gp tasks attach` on it, it would just print a message saying, for example: "It's not possible to attach to a task terminal from another task terminal. Please start a new terminal and run 'gp tasks attach' from there."

felladrin (approved):

PR #8891, opened 2022-03-18
Iterate through pages

### LINKED ISSUE #8650: Cannot see all our repositories with BitBucket (Pagination issue?)
Hi

We are connected with BitBucket and we cannot see all our repositories; if I check the network log I can see that the response contains 100 records. I wonder if we are hitting a pagination limit or similar? We currently have 130 repositories and the ones we want to add to GitPod of course are not showing.

Thanks

_Originally posted by @jmortlock in https://github.com/gitpod-io/gitpod/issues/5095#issuecomment-1061381262_

laushinka:
## Description
This change allows iterating through all the pages of Bitbucket API's response.

## Related Issue(s)
Fixes #8650

## How to test
1. Have some BitBucket repositories.
2. Go to https://laushinka-0486385fa8.staging.gitpod-dev.com/new
3. See all repositories

## Release Notes

```release-note
Allows fetching more than 100 BitBucket repositories.
```

### DIFF
--- a/components/server/ee/src/bitbucket/bitbucket-app-support.ts
+++ b/components/server/ee/src/bitbucket/bitbucket-app-support.ts
@@ -4,10 +4,11 @@
  * See License.enterprise.txt in the project root folder.
  */
 
+// @ts-nocheck
 import { AuthProviderInfo, ProviderRepository, User } from "@gitpod/gitpod-protocol";
 import { inject, injectable } from "inversify";
 import { TokenProvider } from "../../../src/user/token-provider";
-import { Bitbucket } from "bitbucket";
+import { Bitbucket, Schema } from "bitbucket";
 import { URL } from "url";
 
 @injectable()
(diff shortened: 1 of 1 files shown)


components/server/ee/src/bitbucket/bitbucket-app-support.ts
```diff
@@ -41,44 +42,56 @@ export class BitbucketAppSupport {
...
+                const response = await api.repositories
                     .list({
                         workspace,
-                        pagelen: 100,
+                        pagelen,
+                        page,
                         role: "admin", // installation of webhooks is allowed for admins only
                     })
                     .catch((e) => {
                         console.error(e);
```
jldec, line 60: Is there a simple way to show errors like this to users without requiring user interaction like alert() or confirm() do? I like the pattern in GitHub where it adds a blue banner at the top with error or success messages.

laushinka: Do you have a screenshot of an example? Either on our own platform or on GitHub.

AlexTugarev: That's a nice idea. I think we were discussing this already and the toast notification was included as well. 
cc. @gtsiolis

components/server/ee/src/bitbucket/bitbucket-app-support.ts
```diff
@@ -41,44 +42,57 @@ export class BitbucketAppSupport {
         const workspaces =
             (await api.workspaces.getWorkspaces({ pagelen: 100 })).data.values?.map((w) => w.slug!) || [];
 
-        const reposPromise = Promise.all(
-            workspaces.map((workspace) =>
-                api.repositories
+        const fetchAllRepos = async (workspace: string) => {
+            const reposResponse: (Schema.Repository[] | undefined)[] = [];
```
AlexTugarev, line 46: nit: type of `Schema.Repository[]` should be enough

components/server/ee/src/bitbucket/bitbucket-app-support.ts
```diff
@@ -41,44 +42,57 @@ export class BitbucketAppSupport {
...
-
-                    (account === usersBitbucketAccount ? ownersRepos : result).push({
-                        name: repo.name!,
-                        account,
-                        cloneUrl,
-                        updatedAt,
-                        accountAvatarUrl,
                     });
+                if (response) {
+                    reposResponse.push(response.data.values);
```
AlexTugarev, line 63: ... then you can do `reposResponse.push(...response.data.values);`

AlexTugarev:
@laushinka if you want to have this tested in a sustainable way, I'd suggest to make the `pagelen` configurable, e.g. moving the code to a function with such a param. This way we could add a "unit test" which would not require you to create 100 repos to test with ;-)

components/server/ee/src/bitbucket/bitbucket-app-support.ts
jldec: I have opened separate issue #8933.
Keen to ship this fix.

jldec (approved):
works for me in testing - thanks @laushinka

laushinka:
> /hold (to reset pagelen)

Pagelength reset, will /unhold

PR #8892, opened 2022-03-18
[kots]: escape golang template variables for Helm resources

MrSimonEmms:
> This is quite annoying it got missed as (I thought) I had tested this deployment fully. Huge thanks to @geropl for helping me debug this one 🙏🏻 

## Description

Helm parses Golang template variables when rendering the Kubernetes objects. For the ws-manager config (and potentially others), this breaks as this includes Golang template variables which is interpolated at runtime.

Helm allows for escaping of Golang template variables if wrapped in ``{{`x`}}``. The sed command searches for any line containing `{{` and escapes the whole line - because we use backticks and it's (currently) only in a JSON configmap, this escapes everything adequately.

## How to test

Deploy via KOTS and run a workspace

## Release Notes

```release-note
[kots]: escape golang template variables for Helm resources
```

### DIFF
--- a/install/kots/manifests/gitpod-installer-job.yaml
+++ b/install/kots/manifests/gitpod-installer-job.yaml
@@ -167,7 +167,7 @@ spec:
               apiVersion: v2
               name: gitpod-kots
               description: Always ready-to-code
-              Version: "1.0.0"
+              version: "1.0.0"
               appVersion: "$(/app/installer version | yq e '.version' -)"
               EOF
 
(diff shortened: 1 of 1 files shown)


Pothulapati (approved):
LGTM! Also, Good Find!

Was able to find that, the `ws-manager` requires the following templates in its config

```
        "urlTemplate": "https://{{ .Prefix }}.ws.tarun.self-hosted.com",
        "portUrlTemplate": "https://{{ .WorkspacePort }}-{{ .Prefix }}.ws.tarun.self-hosted.com",
```

which is then replaced by the following after the `sed` command:

```
{{`        "urlTemplate": "https://{{ .Prefix }}.ws.tarun.self-hosted.com",`}}
{{`        "portUrlTemplate": "https://{{ .WorkspacePort }}-{{ .Prefix }}.ws.tarun.self-hosted.com",`}}
```

MrSimonEmms:
@Pothulapati correct analysis 👍🏻

Issue #8603, opened 2022-03-04
Include a button icon on the login page for Bitbucket Server provider


gtsiolis:
### Problem to solve

Bitbucket Server (https://github.com/gitpod-io/gitpod/issues/8499) support was added in https://github.com/gitpod-io/gitpod/pull/8086, however, the login button for this type of provider is not using the button layout as the rest of the providers.

See https://github.com/gitpod-io/gitpod/pull/8086#issuecomment-1036352689 and [relevant discussion](https://gitpod.slack.com/archives/C01LDA28WV6/p1646338858901439) (internal). Cc @metcalfc @corneliusludmann @jldec 

![153623370-f59601d5-1170-4133-9784-8d624b1cf589](https://user-images.githubusercontent.com/120486/156772384-4c183d66-9521-433b-b401-6ab72ccaa7b7.png)

### Proposal

This could be resolved by simply re-using the Bitbucket icon as the button icon for this type of provider.

jldec:
@gtsiolis I think we need a button layout which is not truncated.  
E.g. something more like the wider error-page auth button below (but consistent with the other buttons)

![Screenshot 2022-03-15 at 08 08 08](https://user-images.githubusercontent.com/849592/158335098-aa714597-4acd-49b8-b44b-77931cf3fc48.png)

AlexTugarev:
@gtsiolis are you working on this? it's a recurring thing mentioned on other PRs.

AlexTugarev:
@gtsiolis, reusing the icon from Bitbucket seems fairly easy, right? The more interesting part is render the labels properly. Do you have a suggestion for that?

gtsiolis:
@AlexTugarev I'm not actively working on any designs for this. Like you mentioned in https://github.com/gitpod-io/gitpod/issues/8603#issuecomment-1087303433, reusing the Bitbucket icon is a great MVC (minimum viable change) for this issue. 🛹 

**thought**: The truncation issue mentioned in https://github.com/gitpod-io/gitpod/issues/8603#issuecomment-1067691775 is a different issue and can probably be tackled separately. Also, feature requests like https://github.com/gitpod-io/gitpod/issues/7061 could make the truncation issue obsolete.

**suggestion**: For resolving this issue, I'd suggest to simply re-use the Bitbucket icon and optionally use a title attribute for making the button more accessible. We could open a follow up issue to tackle the truncation issue separately. ➿

PR #8896, opened 2022-03-19
[bitbucket-server] support for projects and prebuilds

### LINKED ISSUE #8455: [Bitbucket Server] Add support for prebuilds
The Bitbucket server integration should support installing webhooks and prebuilds.

AlexTugarev:
## Description
This enables prebuilds for Bitbucket Server integrations.

Screenshot 2022-03-19 at 06 00 10

## Breaking Changes
* Admin scopes are required to make the webhook creation work. Full set of scopes to enable for the OAuth App: `PUBLIC_REPOS, REPO_READ, REPO_WRITE, REPO_ADMIN, PROJECT_ADMIN`
* Due to a change in `Identity.authName` mapping, users need to re-login to get that updated, otherwise repository services wont be able to read permissions properly.

## Related Issue(s)

Fixes #8455

## How to test
- [ ] Make sure to create a new account in the preview environment
- [ ] Make sure to create an account in https://bitbucket.gitpod-self-hosted.com (hint: log in as "roboquat" with the credentials in 1password, then create a new user account for yourself)
- [ ] Try to add repos with admin permission as projects. Test with personal repos as well as repos in repos.
  - [ ] Verify that webhooks are created on the BBS.
- [ ] Open workspaces for those repos and make changes.
  - [ ] Verify that prebuilds are triggered.
  - [ ] Verify that prebuilds ran on the proper branch for example.

### TODO
- [ ] manage and use webhook secret to verify events

### Known issues or out of scope
* No support for Issue and PR contexts
* No UI for setting up a BBS integration (this has to be done via config at the moment)

## Release Notes
```release-note
Adding support for Projects and Prebuilds for Bitbucket Server. 
```

### DIFF
--- a/components/server/ee/src/prebuilds/bitbucket-server-service.ts
+++ b/components/server/ee/src/prebuilds/bitbucket-server-service.ts
@@ -1,0 +1,135 @@
+/**
+ * Copyright (c) 2020 Gitpod GmbH. All rights reserved.
+ * Licensed under the Gitpod Enterprise Source Code License,
+ * See License.enterprise.txt in the project root folder.
+ */
+
+import { RepositoryService } from "../../../src/repohost/repo-service";
+import { ProviderRepository, User } from "@gitpod/gitpod-protocol";
+import { inject, injectable } from "inversify";
+import { BitbucketServerApi } from "../../../src/bitbucket-server/bitbucket-server-api";
+import { AuthProviderParams } from "../../../src/auth/auth-provider";
+import { BitbucketServerContextParser } from "../../../src/bitbucket-server/bitbucket-server-context-parser";
+import { Config } from "../../../src/config";
+import { TokenService } from "../../../src/user/token-service";
+import { BitbucketServerApp } from "./bitbucket-server-app";
+
+@injectable()
+export class BitbucketServerService extends RepositoryService {
+    static PREBUILD_TOKEN_SCOPE = "prebuilds";
+
+    @inject(BitbucketServerApi) protected api: BitbucketServerApi;
+    @inject(Config) protected readonly config: Config;
+    @inject(AuthProviderParams) protected authProviderConfig: AuthProviderParams;
+    @inject(TokenService) protected tokenService: TokenService;
+    @inject(BitbucketServerContextParser) protected contextParser: BitbucketServerContextParser;
+
+    async getRepositoriesForAutomatedPrebuilds(user: User): Promise<ProviderRepository[]> {
+        const repos = await this.api.getRepos(user, { limit: 100, permission: "REPO_ADMIN" });
+        return (repos.values || []).map((r) => {
+            const cloneUrl = r.links.clone.find((u) => u.name === "http")?.href!;
+            // const webUrl = r.links?.self[0]?.href?.replace("/browse", "");
+            const accountAvatarUrl = this.api.getAvatarUrl(r.project.key);
+            return <ProviderRepository>{
+                name: r.name,
+                cloneUrl,
+                account: r.project.key,
+                accountAvatarUrl,
+                // updatedAt: TODO(at): this isn't provided directly
+            };
+        });
+    }
+
+    async canInstallAutomatedPrebuilds(user: User, cloneUrl: string): Promise<boolean> {
+        const { host, repoKind, owner, repoName } = await this.contextParser.parseURL(user, cloneUrl);
+        if (host !== this.authProviderConfig.host) {
+            return false;
+        }
+
+        const identity = user.identities.find((i) => i.authProviderId === this.authProviderConfig.id);
+        if (!identity) {
+            console.error(
+                `Unexpected call of canInstallAutomatedPrebuilds. Not authorized with ${this.authProviderConfig.host}.`,
+            );
+            return false;
+        }
+
+        try {
+            await this.api.getWebhooks(user, { repoKind, repositorySlug: repoName, owner });
+            // reading webhooks to check if admin scope is provided
+        } catch (error) {
+            return false;
+        }
+
+        if (repoKind === "users") {
+            const ownProfile = await this.api.getUserProfile(user, identity.authName);
+            if (owner === ownProfile.slug) {
+                return true;
+            }
+        }
+
+        let permission = await this.api.getPermission(user, { username: identity.authName, repoKind, owner, repoName });
+        if (!permission && repoKind === "projects") {
+            permission = await this.api.getPermission(user, { username: identity.authName, repoKind, owner });
+        }
+
+        if (this.hasPermissionToCreateWebhooks(permission)) {
+            return true;
+        }
+
+        console.debug(
+            `User is not allowed to install webhooks.\n${JSON.stringify(identity)}\n${JSON.stringify(permission)}`,
+        );
+        return false;
+    }
+
+    protected hasPermissionToCreateWebhooks(permission: string | undefined) {
+        return permission && ["REPO_ADMIN", "PROJECT_ADMIN"].indexOf(permission) !== -1;
+    }
+
+    async installAutomatedPrebuilds(user: User, cloneUrl: string): Promise<void> {
+        const { owner, repoName, repoKind } = await this.contextParser.parseURL(user, cloneUrl);
+
+        const existing = await this.api.getWebhooks(user, {
+            repoKind,
+            repositorySlug: repoName,
+            owner,
+        });
+        const hookUrl = this.getHookUrl();
+        if (existing.values && existing.values.some((hook) => hook.url && hook.url.indexOf(hookUrl) !== -1)) {
+            console.log(`BBS webhook already installed on ${cloneUrl}`);
+            return;
+        }
+        const tokenEntry = await this.tokenService.createGitpodToken(
+            user,
+            BitbucketServerService.PREBUILD_TOKEN_SCOPE,
+            cloneUrl,
+        );
+        try {
+            await this.api.setWebhook(
+                user,
+                { repoKind, repositorySlug: repoName, owner },
+                {
+                    name: `Gitpod Prebuilds for ${this.config.hostUrl}.`,
+                    active: true,
+                    configuration: {
+                        secret: "foobar123-secret",
+                    },
+                    url: hookUrl + `?token=${encodeURIComponent(user.id + "|" + tokenEntry.token.value)}`,
+                    events: ["repo:refs_changed"],
+                },
+            );
+            console.log("Installed Bitbucket Server Webhook for " + cloneUrl);
+        } catch (error) {
+            console.error(`Couldn't install Bitbucket Server Webhook for ${cloneUrl}`, error);
+        }
+    }
+
+    protected getHookUrl() {
+        return this.config.hostUrl
+            .with({
+                pathname: BitbucketServerApp.path,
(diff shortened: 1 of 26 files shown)


csweichel:
Awesome work - we really need this change.
/cc @geropl what do we need to do to get this merged?

geropl:
@AlexTugarev Can you raise this within the team to get a review this week? Maybe from Jan?

jldec:
Validated adding integration as a user by opening a workspace on a prefixed URL.
Works as expected  ✅ 

Screenshot 2022-03-30 at 10 09 09

Screenshot 2022-03-30 at 10 10 04

Screenshot 2022-03-30 at 10 12 00

layout of the scopes is a bit wide, but not a big problem (look similar to existing BB integration)

jldec:
New project flow fails to fetch branches (upper-case/lower-case issue?) for repo at 
https://bitbucket.gitpod-self-hosted.com/projects/JLDEC/repos/jldec-repo-march-30/browse

Screenshot 2022-03-30 at 10 56 19

Screenshot 2022-03-30 at 10 15 30

jldec:
Project on personal repo is able to fetch branches.  ✅   
Branch list formatting is off (because of avatar - see similar issue for GHE in #8755)

Screenshot 2022-03-30 at 10 54 32

jldec:
Project on same personal repo shows error when looking for .gitpod.yml under configuration page.

Screenshot 2022-03-30 at 11 19 57

jldec:
The button to start a new workspace after creating a new project on a personal repo, appears to create a broken context url since it uses the clone url path `scm/~jldec/test-repo2.git`. Replacing that context url with the actual url path from BBS does work `users/jldec/repos/test-repo2/browse`

Screenshot 2022-03-30 at 12 25 54

Screenshot 2022-03-30 at 12 26 06

Screenshot 2022-03-30 at 12 26 22

jldec:
There does not appear to be any installed webhook on any of the repos for which I have created a project.

Screenshot 2022-03-30 at 12 37 09

AlexTugarev:
@jldec, please check again 🙏🏻 

Also, pay attention to the `Breaking Changes`, esp. to the re-login step. Casing of the auth name was wrong in my account. I've added tests for that, it should meet the expectations in both directions and match the permission queries.

In your case, which is assumingely uncommon, there is a project and a user account named equally. Something you would not expect to be possible. I've change the workaround to try projects first, this should resolve the [issues](https://github.com/gitpod-io/gitpod/pull/8896#issuecomment-1083020300) from above as well.

Let's see if we need a 3rd round 🤞🏻

AlexTugarev:
Just found an issue with the New Project wizard, which was actually reported several times, but we seem to not have found a repro for: `User.identities` contain a `Gitpod` identity, which AFAIU is used to authorize webhook event handler (i.e. for prebuild events) and this is not handled nor expected. 

I'll make a drive-by-change to exclude them from the results of `GitpodService.getAuthProviders` as it's likely unexpected/unhandled in other places.

Screen Shot 2022-04-01 at 08 47 21

easyCZ:
I'm unable to see prebuilds for my project:

Steps:
1. Create a bitbucket project
2. Import https://gitlab.com/gitpod-milan/gitpod-large-image as a repo
3. Create gitpod project
4. Navigate to Branches - nothing visible
Screenshot 2022-04-01 at 9 38 06
5. Try `Run prebuild` - nothing happens, request response contains `Request triggerPrebuild failed with message: 404 / Not Found`
6. Project Settings also seem to not have access to the repo properly
Screenshot 2022-04-01 at 9 42 11

components/server/ee/src/prebuilds/bitbucket-server-service.ts
```diff
@@ -0,0 +1,123 @@
...
+    static PREBUILD_TOKEN_SCOPE = "prebuilds";
+
+    @inject(BitbucketServerApi) protected api: BitbucketServerApi;
+    @inject(Config) protected readonly config: Config;
+    @inject(AuthProviderParams) protected authProviderConfig: AuthProviderParams;
+    @inject(TokenService) protected tokenService: TokenService;
+    @inject(BitbucketServerContextParser) protected contextParser: BitbucketServerContextParser;
+
+    async getRepositoriesForAutomatedPrebuilds(user: User): Promise<ProviderRepository[]> {
+        const repos = await this.api.getRepos(user, { limit: 100, permission: "REPO_ADMIN" });
```
easyCZ, line 28: [nit] Would be good to comment that we're requesting REPO_ADMIN because that's the minimum role required to create webhooks.

components/server/ee/src/prebuilds/bitbucket-server-service.ts
```diff
@@ -0,0 +1,123 @@
...
+            return false;
+        }
+
+        const permission =
+            (await this.api.getPermission(user, { username: identity.authName, repoKind, owner, repoName })) ||
+            (await this.api.getPermission(user, { username: identity.authName, repoKind, owner }));
+
+        if (permission && ["REPO_ADMIN", "PROJECT_ADMIN"].indexOf(permission) !== -1) {
+            return true;
+        }
```
easyCZ, line 70: [nit] It can be sometimes helpful to communicate the reasoning for a particular check though a method name. For example:
```
if (hasPermissionToCreateWebhooks(permission) {
```

components/server/src/bitbucket-server/bitbucket-server-api.ts
```diff
@@ -15,71 +15,304 @@ export class BitbucketServerApi {
...
+
+            if (!response.ok) {
+                throw Error(`${response.status} / ${response.statusText}`);
+            }
+            return await response.text();
+        } catch (error) {
+            result = "error " + error?.message;
+            throw error;
+        } finally {
+            console.debug(`BBS GET ${fullUrl} - ${result}`);
```
easyCZ, line 80:
```suggestion
            console.debug(`BitBucketServer GET ${fullUrl} - ${result}`);
```
Just to reduce mental overhead model.

components/server/src/bitbucket-server/bitbucket-server-api.ts
```diff
@@ -15,71 +15,304 @@ export class BitbucketServerApi {
...
+    }
+
+    async getBranches(
+        user: User,
+        params: { repoKind: "projects" | "users"; owner: string; repositorySlug: string },
+    ): Promise<BitbucketServer.BranchWithMeta[]> {
+        const result = await this.runQuery<BitbucketServer.Paginated<BitbucketServer.BranchWithMeta>>(
+            user,
+            `/${params.repoKind}/${params.owner}/repos/${params.repositorySlug}/branches?details=true&orderBy=MODIFICATION&limit=1000`,
+        );
```
easyCZ, line 238: We seem to be crafting a lot of these URLs ourselves where an SDK could simplify our lives a bit. I found https://www.npmjs.com/package/bitbucket but haven't tried it.

AlexTugarev: "bitbucket.org" is not "bitbucket server", unfortunately. there is nothing out that which covers what we need.

components/server/src/bitbucket-server/bitbucket-server-api.ts
```diff
@@ -131,4 +377,45 @@ export namespace BitbucketServer {
...
+    export interface Webhook {
+        id: number;
+        name: "test-webhook";
+        createdDate: number;
+        updatedDate: number;
+        events: any;
+        configuration: any;
+        url: string;
+        active: boolean;
+    }
```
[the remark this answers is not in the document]
AlexTugarev: same here, it's not for the self-managed bitbucket. it's completely different APIs.

AlexTugarev:
@easyCZ, please invite me to your project to reproduce https://github.com/gitpod-io/gitpod/pull/8896#issuecomment-1085541314

jankeromnes:
Many thanks for implementing BBS! 🏭

Gave this a spin, and was able to successfully add a Bitbucket project! 🎉

---

Minor: The UI looks a little bit unpolished still --

| 1. No BBS provider icon? 2. Should BBS really be visible by default on Login? | 3. Uppercase scopes? 4. Without scope descriptions? | 5. Why the `~` in my name? 6. Shouldn't the project name be vertically centered? |
| --- | --- | --- |
| Screenshot 2022-04-01 at 11 35 15 | Screenshot 2022-04-01 at 11 46 01 | Screenshot 2022-04-01 at 11 49 21 |

---

Also, I wasn't able to start a workspace for some reason:

| Add empty repo, click on New Workspace | Fails with 404? 😳 |
| --- | --- |
| Screenshot 2022-04-01 at 11 50 28 | Screenshot 2022-04-01 at 11 50 14 |

AlexTugarev:
@jankeromnes thanks for asking on the `~`

> Why the ~ in my name?

That indicated personal projects. As @jldec managed to create a personal and a regular project of the same name, I'd consider that a "feature"! Seriously, I've no idea of how to treat that properly. Ambiguity of project name is hard. Let's have the `~` until we learn, how if there is a better way to handle it. Having that said, I see no way in removing it without breaking the other project kinds. Ok?

jankeromnes:
Okay, I was now able to push changes to my BBS repository. Unfortunately, no Prebuilds are being triggered.

When I try the `Run Prebuild` button, it fails with the same `404` message. I guess this is the `~` problem as well?

Screenshot 2022-04-01 at 12 15 38

easyCZ (dismissed):
Thanks, I've tested this as far as I can and seems to work. There are still some rough edges on the integration (largely due to BitBucket Server being a bit crazy) but I think this is a good enough 🛹 .

Code-wise, looks sensible but I don't yet have full context to know if there are skeletons.

jankeromnes:
Nice! Feels like this is closing in on the finish line. 🏁

I've noticed a few remaining rough edges though:

- [ ] This button looks awkward (in "New Project", but I guess it's the same as https://github.com/gitpod-io/gitpod/issues/8603)

Screenshot 2022-04-01 at 17 16 21

- [ ] When I add my project "YOLO" (https://bitbucket.gitpod-self-hosted.com/scm/~jan/yolo.git) as a new project, it fails with:

Screenshot 2022-04-01 at 17 16 36

(The project is actually still created, but I guess without a webhook, because prebuilds do not trigger automatically on pushes)

jldec:
I was a bit surprised to get the extra login button when I start a totally new session - if we ship this and configure our BBS integration for testing, we don't want everyone else (non-gitpod) who lands on gitpod.io for the first time to see the extra button. Can you confirm that this will not be the case for SaaS?
 
Screenshot 2022-04-04 at 00 31 39

jldec:
Regarding the button formatting - the lack of image, and the way the button text bleeds to the edges does look unpolished. I understand that it's cosmestic, but the first impression users see is important as well.

AlexTugarev:
> Can you confirm that this will not be the case for SaaS?

@jldec, don't worry, one need to do a backflip to get it running on self-hosted, there is no path for this to get initialized on SaaS.

AlexTugarev:
Thanks @jankeromnes, overlooked your [issue](https://github.com/gitpod-io/gitpod/pull/8896#issuecomment-1086033153) with the 404 error. I could reproduce it on a fork. Trying to get this sorted out before proceeding.

easyCZ (approved):

PR #8899, opened 2022-03-19
Remove console statements introduced in #7715

filiptronicek:
## Description
I discovered that in parts of the dashboard my console is full of a very long array of my repo recommendations, so this PR fixes that by removing all `console.log()`s introduced in #7715.

## How to test
Should have no effect on anything at all :)

## Release Notes

```release-note
NONE
```

I hope these really are not needed, I assumed they would only be needed in debugging.

cc @jankeromnes


jankeromnes (approved):
Aha, yes these would be my debugging logs. No need to keep them active for every Gitpod user. 😅

Good catch, many thanks for fixing!

Issue #8303, opened 2022-02-18
[admin] Move admin check for pages to `App.tsx`


corneliusludmann:
On admin pages, we have checks like this:

https://github.com/gitpod-io/gitpod/blob/3f3ca7c01d96c16b7b68b04c36c22400f9d3efa0/components/dashboard/src/admin/UserSearch.tsx#L43-L45

It would be better to have something like this in `App.tsx`:

```typescript
if (location.startsWith('/admin') && !userIsAdmin) {
    history.replace('/');
}
```

(based on the proposal of @jankeromnes, see [internal](https://gitpod.slack.com/archives/C01KLC56NP7/p1645177742553389?thread_ts=1645174903.467919&cid=C01KLC56NP7))

PR #8925, opened 2022-03-22
[webapp] move admin redirects to `App.tsx`

Pothulapati:
## Description

Currently, Each page in `admin` has its own redirect to check if
the user has admin access. This is error-prone as some pages
can miss these tags causing unexpected privelleges for non-admin
users.

This is fixed by moving these redirects into admin, where the URL router
is added. We create a new wrapper called `AdminRoute` which has to
be used for all the `/admin` routes.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/8303

## How to test

- Login to https://tar-admin-gate.staging.gitpod-dev.com/workspaces
- Try to access `/admin` pages as a non-admin user.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/dashboard/src/App.tsx
+++ b/components/dashboard/src/App.tsx
@@ -119,6 +119,28 @@ function isWebsiteSlug(pathName: string) {
     return slugs.some((slug) => pathName.startsWith("/" + slug + "/") || pathName === "/" + slug);
 }
 
+// A wrapper for <Route> that redirects to the workspaces screen if the user isn't a admin.
+// This wrapper only accepts the component property
+function AdminRoute({ component }: any) {
+    const { user } = useContext(UserContext);
+    return (
+        <Route
+            render={({ location }: any) =>
+                user?.rolesOrPermissions?.includes("admin") ? (
+                    <Route component={component}></Route>
+                ) : (
+                    <Redirect
+                        to={{
+                            pathname: "/workspaces",
+                            state: { from: location },
+                        }}
+                    />
+                )
+            }
+        />
+    );
+}
+
 export function getURLHash() {
     return window.location.hash.replace(/^[#/]+/, "");
 }
@@ -281,6 +303,7 @@ function App() {
     if (!user) {
         return <Login />;
     }
+
     if (window.location.pathname.startsWith("/blocked")) {
         return (
             <div className="mt-48 text-center">
@@ -340,11 +363,11 @@ function App() {
                     <Route path={projectsPathInstallGitHubApp} exact component={InstallGitHubApp} />
                     <Route path="/from-referrer" exact component={FromReferrer} />
 
-                    <Route path="/admin/users" component={UserSearch} />
-                    <Route path="/admin/teams" component={TeamsSearch} />
-                    <Route path="/admin/workspaces" component={WorkspacesSearch} />
-                    <Route path="/admin/projects" component={ProjectsSearch} />
-                    <Route path="/admin/settings" component={AdminSettings} />
+                    <AdminRoute path="/admin/users" component={UserSearch} />
+                    <AdminRoute path="/admin/teams" component={TeamsSearch} />
+                    <AdminRoute path="/admin/workspaces" component={WorkspacesSearch} />
+                    <AdminRoute path="/admin/projects" component={ProjectsSearch} />
+                    <AdminRoute path="/admin/settings" component={AdminSettings} />
 
                     <Route path={["/", "/login"]} exact>
                         <Redirect to={workspacesPathMain} />
--- a/components/dashboard/src/admin/ProjectsSearch.tsx
+++ b/components/dashboard/src/admin/ProjectsSearch.tsx
@@ -6,12 +6,11 @@
 
 import moment from "moment";
 import { useLocation } from "react-router";
-import { Link, Redirect } from "react-router-dom";
-import { useContext, useState, useEffect } from "react";
+import { Link } from "react-router-dom";
+import { useState, useEffect } from "react";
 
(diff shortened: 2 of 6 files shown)


roboquat:

<details>

</details>

Pothulapati:
With the redirect to `/workspaces` logic, We seem to be redirected (when not a `admin` user) but the page seems to be empty, and only gets elements when re-loaded. Any idea how to fix this?

easyCZ (approved):
Hey, thanks for the contribution. This looks good to me but I'm definitely not an expert on the area.

Adding hold in case you want to wait for someone else to review, otherwise feel free to unhold.

/hold

laushinka:
> With the redirect to `/workspaces` logic, We seem to be redirected (when not a `admin` user) but the page seems to be empty, and only gets rendered when re-loaded. Any idea how to fix this?

@Pothulapati Was going to suggest explicitly returning the component, but I saw that you just did this change. Hope that works!

Pothulapati:
> @Pothulapati Was going to suggest explicitly returning the component, but I saw that you just did this change. Hope that works!

Somehow even that does not seem to work!  🤔

trumbitta:
@Pothulapati this is a workaround: https://stackoverflow.com/a/60825615

I'd suggest reworking the solution a bit though: at my daily job we implemented a handy `PrivateRoute` component based on https://v5.reactrouter.com/web/example/auth-workflow
This one could be named `AdminGuard` or something even more close to its intended use.
Or it could stay `PrivateRoute` and have props to specialize it.

Anyway, such component should be used instead of the usual `Route` component, here: https://github.com/gitpod-io/gitpod/blob/fad6a79a3228e92b24fe2ba56b2fdb443df9b27c/components/dashboard/src/App.tsx#L348

Hope this helps!

Pothulapati:
@trumbitta Thanks for taking the time, and explaining a better way to build this in react! (React newbie here)

I updated the PR to have a new `AdminRoute` wrapper that accepts a `component` arguement, and uses that if a user is a `admin`, or redirects them to `/workspaces`. This also[ works well with the `/admin` redirect to `admin/users`](https://github.com/gitpod-io/gitpod/blob/2a56cc4a106fac383e37c64dd6c6ffe69a4070ad/components/dashboard/src/App.tsx#L384) as we still check admin status at `/admin/users`.

components/dashboard/src/App.tsx
```diff
@@ -119,6 +119,28 @@ function isWebsiteSlug(pathName: string) {
...
 
+// A wrapper for <Route> that redirects to the workspaces screen if the user isn't a admin.
+// This wrapper only accepts the component property
+function AdminRoute({ component }: any) {
+    const { user } = useContext(UserContext);
+    return (
+        <Route
+            render={({ location }: any) =>
+                user?.rolesOrPermissions?.includes("admin") ? (
+                    <Route component={component}></Route>
```
trumbitta, line 130: 🤔 the nested `Route` seems suspicious: shouldn't this just be `{component}`?

Look at how they use `{children}` in the original example:

```tsx
function PrivateRoute({ children, ...rest }) {
  let auth = useAuth();
  return (
    <Route
      {...rest}
      render={({ location }) =>
        auth.user ? (
          children
        ) : (
          <Redirect
            to={{
              pathname: "/login",
              state: { from: location }
            }}
          />
        )
      }
    />
  );
```

components/dashboard/src/App.tsx
```diff
@@ -119,6 +119,28 @@ function isWebsiteSlug(pathName: string) {
     return slugs.some((slug) => pathName.startsWith("/" + slug + "/") || pathName === "/" + slug);
 }
 
+// A wrapper for <Route> that redirects to the workspaces screen if the user isn't a admin.
+// This wrapper only accepts the component property
+function AdminRoute({ component }: any) {
```
trumbitta, line 124: It's awesome that you found my suggestion useful 🥳 

Shouldn't this component also get the other props from `Route`?
In the original example they do it like this with `...rest`:

```tsx
function PrivateRoute({ children, ...rest }) {
  let auth = useAuth();
  return (
    <Route
      {...rest}
```

components/dashboard/src/App.tsx
Pothulapati: Nope, That doesn't work! Looks like we need to pass along with the `Route` tag. It works in the example for children because its some HTML code there but here we need this properties of the current tag i.e `component`

components/dashboard/src/App.tsx
Pothulapati: Just want to be restrictive here, as we know we only send `component` here based on the usages in the admin routes.

We can always do `...rest` if we add something at the references! :)

components/dashboard/src/App.tsx
trumbitta: It *should* work... 🤔 maybe I'll take it for a spin some time in the future and see what happens 🤷

components/dashboard/src/App.tsx
trumbitta: Makes sense :) but how does it work without even just the `path` prop?

Pothulapati: Good question! probably because we don't use `path` anywhere lower down! It's used with higher level Routers but not here, maybe?

trumbitta (dismissed):

trumbitta:
@Pothulapati I just approved your changes only to make GitHub's UI happy. I know me approving them looks weird, because it is! Sorry! 🙈

geropl (approved):
Tested, and works nicely, thx! :pray: 

I like how it simplifies the code.

geropl (approved):

PR #8932, opened 2022-03-23
[observability] Update agent smith graph

princerachit:
## Description

Update agent-smith `Egress Violation` graph to display the count of violations per severity per cluster.
This gives a better intuitive way to understand what is going on in the cluster compared to the previous `rate` query.

## Related Issue(s)

Fixes NA

## How to test

I updated the graph in the agent-smith dashboard and could see correct render. You can test using the same query and legend.
image

## Release Notes

```release-note
None
```


utam0k (approved):
Looks good

PR #8949, opened 2022-03-24
[Analytics] Include Invite ID with Team created Call

jakobhero:
## Description

Generates and includes an invitation ID with each team created Segment call. This is done in order to have an ID available in downstream tools after a team was created so that it can be sent to users outside of the product.

## How to test

1. Open a preview environment, sign up and create a team
2. Go to the [debugger of Staging Untrusted
](https://app.segment.com/gitpod/sources/staging_untrusted/debugger) and ensure that the `team_created` call includes an `invite_id`
## Release Notes

```release-note
NONE
```

## Documentation

### DIFF
--- a/components/server/src/workspace/gitpod-server-impl.ts
+++ b/components/server/src/workspace/gitpod-server-impl.ts
@@ -1974,6 +1974,7 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
         // Note: this operation is per-user only, hence needs no resource guard
         const user = this.checkAndBlockUser("createTeam");
         const team = await this.teamDB.createTeam(user.id, name);
+        const invite = await this.getGenericInvite(ctx, team.id);
         ctx.span?.setTag("teamId", team.id);
         this.analytics.track({
             userId: user.id,
@@ -1983,6 +1984,7 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
                 name: team.name,
                 slug: team.slug,
                 created_at: team.creationTime,
+                invite_id: invite.id,
             },
         });
         return team;


roboquat:

<details>

</details>

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -1974,6 +1974,7 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
         // Note: this operation is per-user only, hence needs no resource guard
         const user = this.checkAndBlockUser("createTeam");
         const team = await this.teamDB.createTeam(user.id, name);
+        const invite = await this.getGenericInvite(ctx, team.id);
```
easyCZ, line 1977: Retrieval of the `genericInvite` seems to also reset the invite link. Is this potentially a problem if the invite has already been sent? I'm not intimately familiar with the product/code behaviour in this area so seeking clarification.

https://github.com/gitpod-io/gitpod/blob/db8d56c49157d6f242778a4b69fbc5b8fc5f0d4b/components/server/src/workspace/gitpod-server-impl.ts#L2059

easyCZ (commented):
Looks sensible, just a question to clarify my understanding.

components/server/src/workspace/gitpod-server-impl.ts
jakobhero: hi @easyCZ , thanks for having a look! i went back into the preview environment and checked, the invite URL attached to the team created call here does not change unless the reset invite button is clicked below (at which point any previously sent invites lose their validity). thus, using `this.getGenericInvite()` is fine here imo as:

- no invites to a team can be sent prior to its creation
- the behaviour after this call is consistent with what's happening now as the invite URL remains the same unless it's manually reset

Screenshot 2022-03-28 at 10 54 34

easyCZ (approved):

PR #8955, opened 2022-03-25
ws-daemon: Use a pair of veths instead of slirp4netns

### LINKED ISSUE #8106: [workspacekit] Use veth instead of slirp4netns
Today we use slirp4netns to connect the network namespace of our workspaces. While convenient and easy to use, this comes at a runtime cost because we make networking a userland issue.

Instead, we should defer to the ws-dameon IWS to create `veth` pairs between ring1 and the workspace pod.

[Also, if there's an opportunity to improve the logging, that would be great.](https://gitpod.slack.com/archives/C01KGM9BH54/p1645792024776959)

Maybe IPVLAN or MACVLAN work nicer because we don't have to setup routing in the workspace pod - but the interaction with CNI is unclear.

utam0k:
## Description

This PR creates a pair of veth between the workspace and ring1.

## Related Issue(s)

Fixed #8106 

## How to test

in gitpod workspace
https://to-setup-veth.staging.gitpod-dev.com/workspaces

```console
$ docker run hello-world
```

maybe you have to kill slirp4netns's process.

## Release Notes

```release-note
Use a pair of veths instead of slirp4netns
```

## Documentation

No


components/ws-daemon-api/workspace_daemon.proto
```diff
@@ -45,6 +45,9 @@ service InWorkspaceService {
     // Teardown prepares workspace content backups and unmounts shiftfs mounts. The canary is supposed to be triggered
     // when the workspace is about to shut down, e.g. using the PreStop hook of a Kubernetes container.
     rpc Teardown(TeardownRequest) returns (TeardownResponse) {}
+
+    // Set up a pair of veths that interconnect the specified PID and the workspace container's network namename.
```
princerachit, line 49:
```suggestion
    // Set up a pair of veths that interconnect the specified PID and the workspace container's network namespace.
```

components/workspacekit/cmd/rings.go
```diff
@@ -495,6 +495,18 @@ var ring1Cmd = &cobra.Command{
 		//nolint:errcheck
 		defer slirpCmd.Process.Kill()
 
+		client, err = connectToInWorkspaceDaemonService(ctx)
```
princerachit, line 498: Can we reuse the client created in previous steps?

components/workspacekit/cmd/rings.go
```diff
@@ -495,6 +495,18 @@ var ring1Cmd = &cobra.Command{
...
+		if err != nil {
+			log.WithError(err).Error("cannot connect to daemon")
+			return
+		}
+		_, err = client.SetupPairVeths(ctx, &daemonapi.SetupPairVethsRequest{Pid: int64(cmd.Process.Pid)})
+		if err != nil {
+			log.WithError(err).Error("can not set up pair of veths")
```
princerachit, line 505:
```suggestion
			log.WithError(err).Error("cannot setup pair of veths")
```

components/ws-daemon/nsinsider/main.go
```diff
@@ -251,6 +252,162 @@ func main() {
...
+					out, err = cmd.CombinedOutput()
+					if err != nil {
+						return xerrors.Errorf("bring up the lo (%v) failed: %q\n%v",
+							cmd.Args,
+							string(out),
+							err,
+						)
+					}
+
+					cmd = exec.Command(iptablesCmd, "-A", "FORWARD", "-o", containerIf, "-i", vethIf, "-j", "ACCEPT")
```
aledbf, line 370: Why are you adding these FORWARD rules? The default policy is ACCEPT
```
$ sudo iptables -L
Chain INPUT (policy ACCEPT)
target     prot opt source               destination         

Chain FORWARD (policy ACCEPT)
target     prot opt source               destination         

Chain OUTPUT (policy ACCEPT)
target     prot opt source               destination 
```

utam0k:
I will have more drawings and other information on Monday that will show more of what I intend to do.
/hold

components/ws-daemon/nsinsider/main.go
```diff
@@ -251,6 +252,162 @@ func main() {
...
+						Mask: mask,
+					}
+					masqueradeAddr := net.IPNet{
+						IP:   vethAddr.IP.Mask(mask),
+						Mask: mask,
+					}
+					ipCmd, iptablesCmd := "/usr/sbin/ip", "/usr/sbin/iptables"
+					netns := "workspace-ns"
+
+					cmd := exec.Command(ipCmd, "link", "add", vethIf, "type", "veth", "peer", "name", cethIf)
```
aledbf, line 289: @utam0k why not use "github.com/vishvananda/netlink"
```
veth := &netlink.Veth{
    LinkAttrs: netlink.LinkAttrs{
        Name:  vethIf,
        Flags: net.FlagUp,
        MTU:   1500,
        PeerName: cethIf,
    },
    PeerNamespace: netlink.NsFd(int(<hostNS>.Fd())),
}
```

to remove any package dependency in the workspace?

components/ws-daemon/nsinsider/main.go
```diff
@@ -251,6 +252,162 @@ func main() {
...
+					out, err = cmd.CombinedOutput()
+					if err != nil {
+						return xerrors.Errorf("bring up the cethIf (%v) failed: %q\n%v",
+							cmd.Args,
+							string(out),
+							err,
+						)
+					}
+
+					cmd = exec.Command(ipCmd, "netns", "exec", netns, "ip", "link", "set", "lo", "up")
```
aledbf, line 360: Why do we need to set lo up?

components/ws-daemon/nsinsider/main.go
utam0k: That's right. We don't need it. Before this code, I tried to add rules to the original chain and I forget removing the codes.

components/ws-daemon/nsinsider/main.go
utam0k: It is just what I was looking for! Thanks. I redid the implementation.

kylos101:
Hey @utam0k , given the commit message in [23346cf](https://github.com/gitpod-io/gitpod/pull/8955/commits/23346cfb9823141c9f9acabaf391c2119f1d4073), I am going to mark this back as draft (it seems like you're unsure if netlink will work because you use the word "try"). May I ask you to confirm if this is ready for review? If yes, feel free to mark as ready for review.

kylos101:
Hi @utam0k , while this is in draft, if you would like "an early review". that is totally okay to request. :love_letter: :heart_eyes: See you tomorrow!

utam0k:
@kylos101 Thanks for organizing the PR. This PR will be available for review soon. I had a little trouble verifying it.

components/ws-daemon/nsinsider/main.go
utam0k: @aledbf Thanks for your review!
For now, I guess slirp4netns enable `lo` automatically. So, when we want to remove `slirp4netns`,  we have to enable `lo` manually.

utam0k:
@princerachit @aledbf @iQQBot @kylos101 Sorry to keep you waiting. I had finished brushing up on this PR. PTAL.
I recommended [this picture](https://github.com/gitpod-io/gitpod/issues/8106#issuecomment-1080250970) before you start reviewing. This whole attempt is too large to be in one PR, so it is divided into two phases(PRs). This PR is the first these. There should be no problem if only this PR is merged and deployed.

components/ws-daemon/nsinsider/main.go
```diff
@@ -254,6 +256,101 @@ func main() {
...
+						return xerrors.Errorf("cannot found %q netns failed: %v", vethIf, err)
+					}
+					if err := netlink.AddrAdd(vethLink, &netlink.Addr{IPNet: &vethIp}); err != nil {
+						return xerrors.Errorf("failed to add IP address to %q: %v", vethIf, err)
+					}
+					if err := netlink.LinkSetUp(vethLink); err != nil {
+						return xerrors.Errorf("failed to enable %q: %v", vethIf, err)
+					}
+
+					cmd := exec.Command(iptablesCmd, "-t", "nat", "-A", "POSTROUTING", "-s", masqueradeAddr.String(), "-o", containerIf, "-j", "MASQUERADE")
```
csweichel, line 306: `nsinsider` should never exec out like this because it runs as actual root (i.e. has all capabilities on the node), yet executes within the mount namespace of the container. Hence, we're running a user-supplied `iptables` as actual root. It's easy to see how this is a security issue.

If we want to keep using iptables, we might have to resort to the iptables command (unfortunately netfilter doesn't have too great a userspace API). To do this we must:
- ensure within nsinsider that we're running in the original ws-daemon mount namespace. This is a safety precaution so that we don't inadvertently execute `iptables` from a different context in the future. I'm not sure how to do this reliably though :/ 
- use `iptables` that we ship within `ws-daemon`.


Alternatively, we could use `nftables` (the successor to iptables) for which [better library support exists](https://github.com/google/nftables).

csweichel (commented):
Awesome to see work on this :)

For the port-forwarding, we could use `nftables` to [forward all traffic](https://ungleich.ch/u/blog/nftables-magic-redirect-all-ports-to-one-port/). If that doesn't work, we might have to resort to "auto-tunneling" triggered by workspacekit, where we observe the open ports and call on IWS to add the appropriate rules. I'm not sure how that would scale though - and it's considerably more complex than a single rule.

utam0k:
@csweichel Thanks for your review. I'll check the detail tomorrow, but I want to ask you one question quickly.
My understanding is that the supervisor is currently exposing the port using the slirp4nets functionality. I am trying to implement functionality similar to this. Wouldn't exposing all ports lead to users not being able to choose to keep their ports private? I am wondering if perhaps my understanding is incorrect.
https://github.com/gitpod-io/gitpod/blob/3d4e0e4ec4441c72566f2e0e451ea52063a3c50d/components/supervisor/pkg/ports/slirp4netns.go#L31-L55

components/ws-daemon/nsinsider/main.go
utam0k: @csweichel  Thanks for the great findings. You are absolutely right. So `iptablesCmd` just happens to exist depending on the user, right? And thanks for the great suggestions. I'm giving it some thought, I didn't know about `nftables` so I'll give it a bit of a try.
I looked into `iptables` command too, but it looks like all I can do is use the command.

csweichel:
> @csweichel Thanks for your review. I'll check the detail tomorrow, but I want to ask you one question quickly. My understanding is that the supervisor is currently exposing the port using the slirp4nets functionality. I am trying to implement functionality similar to this. Wouldn't exposing all ports lead to users not being able to choose to keep their ports private? I am wondering if perhaps my understanding is incorrect.

supervisor exposing ports through slirp4netns is more of a workaround/burden than a feature. The default behaviour we want to provide is that all served ports are exposed privately by default - always. There is no "workspace internal private mode only"; all access control happens within ws-proxy. If we can do away with the need to expose ports on a "port by port" basis, that's a benefit :)

utam0k:
> > @csweichel Thanks for your review. I'll check the detail tomorrow, but I want to ask you one question quickly. My understanding is that the supervisor is currently exposing the port using the slirp4nets functionality. I am trying to implement functionality similar to this. Wouldn't exposing all ports lead to users not being able to choose to keep their ports private? I am wondering if perhaps my understanding is incorrect.
> 
> supervisor exposing ports through slirp4netns is more of a workaround/burden than a feature. The default behaviour we want to provide is that all served ports are exposed privately by default - always. There is no "workspace internal private mode only"; all access control happens within ws-proxy. If we can do away with the need to expose ports on a "port by port" basis, that's a benefit :)

Oh, my understanding was wrong, Thanks for your guide, I go it. It sounds good.

components/ws-daemon/nsinsider/main.go
```diff
@@ -254,6 +258,193 @@ func main() {
...
+					// 			Data:     cethIp.IP.To4(),
+					// 		},
+					// 		&expr.NAT{
+					// 			Type:        expr.NATTypeDestNAT,
+					// 			Family:      unix.NFPROTO_IPV4,
+					// 			RegAddrMin:  2,
+					// 			RegProtoMin: 1,
+					// 		},
+					// 	},
+					// })
```
utam0k, line 404: This is the result of applying everything. I have applied (commented out) everything, changed the default gw, deleted the slirp4netns process, and confirmed that the health check of `supervisor` succeeds and the workspace starts.
```console
root@gitpodio-templategolangc-7xs0qksc7q0:/# nft list table nat
table ip nat {
        chain postrouting {
                type nat hook postrouting priority srcnat; policy accept;
                ip saddr 10.0.5.0/24 oifname "eth0" masquerade
        }

        chain prerouting {
                type nat hook prerouting priority filter; policy accept;
                tcp dport 1-65535 dnat to 10.0.5.2:tcp dport
        }
}
```

components/ws-daemon/nsinsider/main.go
```diff
@@ -254,6 +258,193 @@ func main() {
...
+					postrouting := nc.AddChain(&nftables.Chain{
+						Name:     "postrouting",
+						Hooknum:  nftables.ChainHookPostrouting,
+						Priority: nftables.ChainPriorityNATSource,
+						Table:    nat,
+						Type:     nftables.ChainTypeNAT,
+					})
+
+					// ip saddr 10.0.5.0/24 oifname "eth0" masquerade
+					nc.AddRule(&nftables.Rule{
```
utam0k, line 323: If you are interested in the rule commands, I would recommend the following resources. However, these may be a bit tougher to understand. If you have experience with eBPF, maybe you are familiar with these rule commands, because I feel it is similar to eBPF.
- [Quick reference-nftables in 10_minutes](https://wiki.nftables.org/wiki-nftables/index.php/Quick_reference-nftables_in_10_minutes)
- [test codes of google/nftables](https://github.com/google/nftables/blob/master/nftables_test.go)

utam0k:
@princerachit @aledbf @iQQBot @kylos101 @csweichel This PR is ready for review again. PTAL. And I updated the picture.
https://github.com/gitpod-io/gitpod/issues/8106#issuecomment-1080250970

components/workspacekit/cmd/rings.go
```diff
@@ -495,6 +495,18 @@ var ring1Cmd = &cobra.Command{
 		//nolint:errcheck
 		defer slirpCmd.Process.Kill()
 
+		client, err = connectToInWorkspaceDaemonService(ctx)
+		if err != nil {
+			log.WithError(err).Error("cannot connect to daemon")
+			return
+		}
+		_, err = client.SetupPairVeths(ctx, &daemonapi.SetupPairVethsRequest{Pid: int64(cmd.Process.Pid)})
```
aledbf, line 503: Please move `client.Close()` to line 503 and use something like `defer client.Close()`

csweichel:
Thanks @utam0k for moving to nftables.

I'm not sure I understand why we need to keep slirp4netns around. As far as I understand the nft rules would forward all ports now, without the use/need of/for slirp4netns. Is that not correct?

csweichel:
I reckon having both (adding veth and removing slirp4netns) in one PR would be easier to review. Otherwise we'd be left with some in-between state. At the least, we should disable slirp4netns (if not remove it).

components/ws-daemon/seccomp-profile-installer/main.go
```diff
@@ -42,7 +42,7 @@ func main() {
 			Action: specs.ActAllow,
 		},
 
-		// slirp4netns requires setns, as do we for debugging
+		// Running docker on a workspace requires setns
```
utam0k, line 45: docker is still using slirp4netns.

components/workspacekit/cmd/rings.go
utam0k: Is there any particular reason why you don't use defer in other places?
https://github.com/gitpod-io/gitpod/blob/e30502ed12142a4125e63fda868f09001d8a3b53/components/workspacekit/cmd/rings.go#L74-L85

utam0k:
@princerachit @aledbf @iQQBot @kylos101 @csweichel This PR is ready for review again. PTAL. Load and other measurements will be done tomorrow, but the code itself is available for review.

utam0k:
## Network IO metrics

I measured the limits of traffic from workspace to ring1 using `iperf3`.

### slirp4netns

```console
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec  9.78 GBytes  8.40 Gbits/sec    0             sender
[  5]   0.00-10.00  sec  9.78 GBytes  8.40 Gbits/sec                  receiver
```

### veth

```console
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec  50.5 GBytes  43.4 Gbits/sec  284             sender
[  5]   0.00-10.00  sec  50.5 GBytes  43.4 Gbits/sec                  receiver
```

## CPU metrics

This is the CPU utilization when 8 Gbit/sec of traffic continues to flow(workspace -> ring1)

|               | slirp4netns | veth   | 
| ------------- | ----------- | ------ | 
| avg of user   | 4.46 %      | 2.64 % | 
| avg of system | 10.07 %     | 1.69%  | 

More details can be found in the `vmstat` results in [this spreadsheet](https://docs.google.com/spreadsheets/d/1V9TVg6BjebG5vY4BSGj1RuNkum04KbtNWqDixy9GalE/edit#gid=1857623308).

utam0k:
@csweichel I think you are probably interested in these above metrics.

components/ws-daemon/seccomp-profile-installer/main.go
csweichel: We disabled the docker slirp4netns use when we introduced slirp4netns for the workspace as a whole. We have however not yet removed that code from docker-up. We should do that in a follow-up PR.

csweichel (approved):
Works like a charm - love the new performance!
Really great work @utam0k.

aledbf (approved):

kylos101:
> Another thing to consider is how to deploy, supervisor / workspacekit and ws-daemon has different deploy cycle

@iQQBot We use installer to ship `workspace-clusters`. When we do this, it uses the `versions.yaml` to parse necessary versions. For example: `docker run --rm -it eu.gcr.io/gitpod-core-dev/build/versions:main.2836 cat versions.yaml`.`ws-daemon` and `registry-facade`, as well as `supervisor` and `workspacekit`, will use versions from that file. The configmap for `registry-facade` will specify `supervisor` and `workspacekit` versions. 

For WebApp, they currently use helm, but I believe it uses a similar versions strategy, so if we merge to main, all these components should ship too. @geropl , can you confirm?

How do you deploy IDE changes? I see this [job](https://github.com/gitpod-io/ops/blob/18c819ceec7468d9c5fd5f47522a67b0c7e8e0d2/.werft/ide-deploy-prod.ts#L43) updates web app clusters using argo. So it seems like, if we merge to main, IDE should avoid doing IDE deploy until web app has deployed. WDYT?

akosyakov (approved):
i have not tried, clean up in supervisor looks good.

PR #8956, opened 2022-03-25
[local-app] Cleanup run command, apply recommended go style

easyCZ:
## Description

* Restructure run arguments into an object, easier extension and visual grepping
* Use CamelCase for constants
* Use the same logic of temp definition across platforms, `os.Temp` already produces `/tmp` on darwin and linux.

## How to test

```bash
cd components/local-app
go run main.go
```

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/local-app/pkg/auth/auth.go
+++ b/components/local-app/pkg/auth/auth.go
@@ -120,25 +120,24 @@ const html = `
 </html>`
 
 // NOTE: the port ranges all need to be valid redirect URI's in the backend
-const STARTING_PORT_NUM = 63110
-const ENDING_PORT_NUM = 63120
+const (
+	StartingPortNum = 63110
+	EndingPortNum   = 63120
+)
 
 // Login walks through the login flow for obtaining a Gitpod token
 func Login(ctx context.Context, opts LoginOpts) (token string, err error) {
 	// Try a range of ports for local redirect server
-	var rl net.Listener
-	var port int
-	for port = STARTING_PORT_NUM; port < ENDING_PORT_NUM || rl == nil; port++ {
-		rl, err = net.Listen("tcp4", fmt.Sprintf("127.0.0.1:%d", port))
-		if err != nil {
-			logrus.Infof("Could not open port:%d got error:%s\nTrying next port\n", port, err)
-			continue
+	rl, port, err := findOpenPortInRange(StartingPortNum, EndingPortNum)
+	if err != nil {
+		return "", err
+	}
+
+	defer func() {
+		if closeErr := rl.Close(); closeErr != nil {
+			logrus.WithField("port", port).WithError(closeErr).Warn("Failed to to close listener")
 		}
-		defer rl.Close()
-	}
-	if rl == nil {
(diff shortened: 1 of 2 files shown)


roboquat:

<details>

</details>

components/local-app/pkg/auth/auth.go
```diff
@@ -120,24 +120,29 @@ const html = `
...
 	var rl net.Listener
 	var port int
-	for port = STARTING_PORT_NUM; port < ENDING_PORT_NUM || rl == nil; port++ {
+	for port = StartingPortNum; port < EndingPortNum || rl == nil; port++ {
 		rl, err = net.Listen("tcp4", fmt.Sprintf("127.0.0.1:%d", port))
 		if err != nil {
 			logrus.Infof("Could not open port:%d got error:%s\nTrying next port\n", port, err)
 			continue
 		}
-		defer rl.Close()
```
easyCZ, line 137: `defer` inside a for loop gets only queued up and executed once the function exits. This way, we're releasing the port immediately.

mustard-mh: We should use `defer`, because we use `rl` below.

To make code more readable, we can

1. `break` after listening to one port success and remove condition `rl == nil`
2. defer close `rl` after check `rl` is not nil (below line 144)

I pushed a commit, feel free to revert or rebase

akosyakov:
@mustard-mh Could you finish a review please?

mustard-mh:
> @mustard-mh Could you finish a review please?

Need more input from @easyCZ  about https://github.com/gitpod-io/gitpod/pull/8956#discussion_r836250727

cc @akosyakov

components/local-app/pkg/auth/auth.go
easyCZ: Thanks, that makes sense. I'll pull the listener acquistion into a seperate method just to clean up the continue/break a bit.

easyCZ:
@mustard-mh Thanks for the review. I've incorporated your changes and split out the port acquire into a seperate method to make it a bit cleaner.

components/local-app/pkg/auth/auth.go
```diff
@@ -120,26 +120,25 @@ const html = `
...
-	if rl == nil {
-		return "", xerrors.Errorf("could not open any valid port in range %d - %d", STARTING_PORT_NUM, ENDING_PORT_NUM)
+	rl, port, err := findOpenPortInRange(StartingPortNum, EndingPortNum)
+	if err != nil {
+		return "", err
 	}
 
+	defer func() {
+		if closeErr := rl.Close(); closeErr != nil {
+			logrus.WithField("port", port).WithError(closeErr).Warn("Failed to to close listener")
```
mustard-mh, line 138: > WARN[0005] Failed to to close listener                   error="close tcp4 127.0.0.1:63110: use of closed network connection" port=63110

After test, we can see this warning every time. we don't need this warning, because it will always appear if auth works(it will close by [this one](https://github.com/gitpod-io/gitpod/pull/8956/files#diff-45dc697108fc2017a12d5505d0cfe7279372d64248c62238a15d1c4ade7be7f7R166) too).

mustard-mh (approved):
LGTM

We can remove this warning in another PR if it's necessary

PR #8957, opened 2022-03-25
[agent-smith] use tap0 to account egress traffic

princerachit:
## Description

Use only tap0 interface to account for egress traffic in agent-smith. We (will in future) penalize worspaces and users based on the egress traffic. With existing implementation we were accounting for all the interfaces to calculate egress, however, this should not be a case as local running servers, local docker registries etc will cause the workspace to be penalized when no actual violation has occurred.

We should use only `tap0` interface to calculate egress traffic.

## Related Issue(s)

Related to https://github.com/gitpod-io/ops/issues/1384#issuecomment-1078900091

## How to test

1. Start a workspace using this branch
2. Start a local http server/ docker registry e.g. `python3 -m http.server`
3. Upload/Download a huge file e.g. `wget 0.0.0.0:8000/testfile`
4. You should not see logs in agent smith about infringing workspace.
5. 
Tip: You can create testfile using `fallocate testfile -l 5G`

## Release Notes

```release-note
None
```

### DIFF
--- a/components/ee/agent-smith/pkg/agent/egress.go
+++ b/components/ee/agent-smith/pkg/agent/egress.go
@@ -20,5 +20,5 @@ func getEgressTraffic(pid int) (int64, error) {
 		return -1, err
 	}
 
-	return int64(nd.Total().TxBytes), nil
+	return int64(nd["tap0"].TxBytes), nil
 }


princerachit:
See https://github.com/gitpod-io/ops/issues/1384#issuecomment-1079372524 for context. Closing this.

PR #8970, opened 2022-03-28
[dashboard] replace dev configuration with env vars

trumbitta:
## Description

Set 2 variables once instead of editing `craco.config.js` every time you want to contribute to the dashboard app.

## How to test

- Follow the new instructions in `components/dashboard/README.md`
- Check if the dashboard loads properly
- 🚨 **IMPORTANT!** Do a production build and check that one as well: I never used CRACO's `whenDev` before

## Release Notes

```release-note
[dashboard] It's now even easier to contribute: set a couple variables once, contribute for as long as you like.
```

### DIFF
--- a/components/dashboard/craco.config.js
+++ b/components/dashboard/craco.config.js
@@ -3,11 +3,27 @@
  * Licensed under the GNU Affero General Public License (AGPL).
  * See License-AGPL.txt in the project root for license information.
  */
+const { when } = require("@craco/craco");
 
 module.exports = {
-  style: {
-    postcss: {
-      plugins: [require("tailwindcss"), require("autoprefixer")],
+    style: {
+        postcss: {
+            plugins: [require("tailwindcss"), require("autoprefixer")],
+        },
     },
-  },
+    ...when(process.env.GP_DEV_HOST && process.env.GP_DEV_COOKIE, () => ({
+        devServer: {
+            proxy: {
+                "/api": {
+                    target: "https://" + process.env.GP_DEV_HOST,
+                    ws: true,
+                    headers: {
+                        host: process.env.GP_DEV_HOST,
+                        origin: "https://" + process.env.GP_DEV_HOST,
+                        cookie: process.env.GP_DEV_COOKIE,
+                    },
+                },
+            },
+        },
+    })),
 };
--- a/components/dashboard/README.md
+++ b/components/dashboard/README.md
@@ -6,14 +6,14 @@ The `App.tsx` is the entry point for the SPA and it uses React-Router to registe
 
 ```ts
 <Switch>
-  <Route path="/" exact component={Workspaces} />
-  <Route path="/profile" exact component={Profile} />
-  <Route path="/notifications" exact component={Notifications} />
-  <Route path="/subscriptions" exact component={Subscriptions} />
-  <Route path="/env-vars" exact component={EnvVars} />
-  <Route path="/git-integration" exact component={GitIntegration} />
-  <Route path="/feature-preview" exact component={FeaturePreview} />
-  <Route path="/default-ide" exact component={DefaultIDE} />
+    <Route path="/" exact component={Workspaces} />
+    <Route path="/profile" exact component={Profile} />
+    <Route path="/notifications" exact component={Notifications} />
+    <Route path="/subscriptions" exact component={Subscriptions} />
+    <Route path="/env-vars" exact component={EnvVars} />
+    <Route path="/git-integration" exact component={GitIntegration} />
+    <Route path="/feature-preview" exact component={FeaturePreview} />
+    <Route path="/default-ide" exact component={DefaultIDE} />
 </Switch>
 ```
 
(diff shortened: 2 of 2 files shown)


trumbitta:
🆙 I'm quite fond of this one, so 🤞

laushinka:
Have to run again due to Werft issues earlier. 

/werft run

laushinka:
Again 😭 [[internal Slack](https://gitpod.slack.com/archives/C032A46PWR0/p1649248958458139)]

/werft run

components/dashboard/craco.config.js
```diff
@@ -3,11 +3,27 @@
...
-  style: {
-    postcss: {
-      plugins: [require("tailwindcss"), require("autoprefixer")],
+    style: {
+        postcss: {
+            plugins: [require("tailwindcss"), require("autoprefixer")],
+        },
     },
-  },
+    ...whenDev(() => ({
```
iQQBot, line 14: devServer is always using in dev, you can add some check for `process.env.GP_DEV_HOST` and `process.env.GP_DEV_COOKIE` to make sure someone want to enable this feature

trumbitta: Is this necessary? The instructions before this change were like: if you want to contribute to the dashboard app, you have to add a `devServer` block with this and that.
The changes in this PR just make those instructions easier: you don't have to add that block every time you open a workspace, and you don't have to make sure every time that you don't commit and push that block.

Are there more, undocumented, situations where one doesn't need to add a `devServer` block in order to work on the dashboard app?

iQQBot: member can use `telepresence` for development, it need permission for preview environment kubectl

laushinka: @iQQBot Will we hit this whenDev() block when we use telepresence?

iQQBot: image
I think it will hit, because telepresence also use `craco start`

mustard-mh:
Good idea🤩 , I used to add `const COOKIE = require('fs').readFileSync(".dev.local").toString()` and apply it before 😂

components/dashboard/craco.config.js
trumbitta: I don't know what `telepresence` is 😬 but I'll try to act on your feedback as soon as I am able. 💪 

Meanwhile:

Are you Gitpodders using `telepresence` instead of the "public" development flow?
If that's the case, do you think I should add a couple lines about it in the README, so that new Gitpodders can onboard themselves?
And if yes, can you share some details about it? I'll do my best to digest them and add them to the README.

iQQBot: we have a notion document about using `telepresence`, current is internal, But to be honest, I didn't use it 😂

image
I do some test, when someone use `telepresence` is will in `whenDev`

mustard-mh: Isn't `devServer` is only work with dev? Why we wrap it with `whenDev`? 

cc @trumbitta

iQQBot, line 14:
```suggestion
    ...when(process.env.GP_DEV_HOST && process.env.GP_DEV_COOKIE, () => ({
```

laushinka: Thanks for proposing this change, @iQQBot! I would go along with being explicit whenever possible.

trumbitta: > we have a notion document about using `telepresence`, current is internal, But to be honest, I didn't use it 😂
> 
> image I do some test, when someone use `telepresence` is will in `whenDev`

@iQQBot  @laushinka (I'm still going to act on your initial feedback ASAP, but I'd like to understand) If it's correct for `telepresence` to run in dev mode, then why isn't it correct for `telepresence` to also use the `devServer`?

iQQBot: For telepresence, which works by replacing pods in a kubernetes cluster and redirecting traffic to access them, it has strong limitations, for example, it can only debug the current preview environment, it cannot debug gitpod.io

The reason I'm using devServer here is that our server has access restrictions on origin, and ws-proxy automatically filters out sensitive cookies, making it impossible to skip these restrictions at the browser level without using devServer's forwarding capabilities

iQQBot: Another reason I wanted to add the check is that if a user doesn't have `GP_DEV_HOST` configured then bringing in devServer results in all his requests being directed to `https://` which is an address that can never be reached and can be confusing

laushinka: Thanks @iQQBot for jumping in and thanks @trumbitta for this implementation! This change will greatly benefit our contributors 🙏🏽 
The use of telepresence is explained in an internal doc, but it's a good point for us to reconsider moving it closer to our code. 
And for now as @iQQBot explained, for this change, we should make sure that only contributors will hit the devServer block by explicitly checking for the env variables.
Looking forward to that change and shipping this 🔥

trumbitta:
@laushinka I followed the example by @iQQBot and updated the code 🧡

laushinka (approved):

laushinka:
Seems like I have to do this again 🤔

/werft run

PR #8978, opened 2022-03-28
Refresh web page when instanceId changed

### LINKED ISSUE #8226: [supervisor] Ensure we fetch IDEOptions/desktopIDEConfig on re-start
So far we set [dekstopIDEConfig](https://github.com/gitpod-io/gitpod/blob/c0bbf911db3a13c23cd7da8a124c17b5046e245c/components/supervisor/frontend/src/index.ts#L119-L123) and [IDEOptions](https://github.com/gitpod-io/gitpod/blob/0fa9200ac83835463267630ab01f79a01a1db5fb/components/dashboard/src/start/StartWorkspace.tsx#L131-L132) onload. With the [recent change away from auto-reload behavior](https://github.com/gitpod-io/gitpod/pull/8125#issuecomment-1038846404) this does no longer show "Open in < desktopIDE >" in cases where:
 - you open a workspace in tab 1
 - stop it in tab 2 (dashboard, for example)
 - re-start it in tab 2 (dashboard, for example)

We should see
image
but instead see this:
image

The dashboard change is quite [straight forward](https://github.com/gitpod-io/gitpod/pull/8188). In supervisor-frontend we'd need to re-fetch the desktopIDEConfig on re-start as well.

/cc @akosyakov

mustard-mh:
## Description

https://github.com/gitpod-io/gitpod/issues/8226#issuecomment-1080807298 Option3 with refresh page

Refresh web page when instanceId changed to get ide config again

## Related Issue(s)

Fixes #8226

## How to test

1. Start a workspace A (ie. browser IDE only) in tab 1
2. Go to Dashboard in tab2 to stop A, go to Preferences to choose one desktop IDE, and re-open A
3. Note tab1 to see if display "Open in xxx"

## Release Notes

```release-note
Fix IDE options display incorrectly when restarting the workspace
```

### DIFF
--- a/components/supervisor/frontend/src/index.ts
+++ b/components/supervisor/frontend/src/index.ts
@@ -108,9 +108,20 @@ const toStop = new DisposableCollection();
     //#region current-frame
     let current: HTMLElement = loading.frame;
     let desktopRedirected = false;
+    let currentInstanceId = "";
     const nextFrame = () => {
         const instance = gitpodServiceClient.info.latestInstance;
         if (instance) {
+            // refresh web page when instanceId changed
+            if (currentInstanceId !== "") {
+                if (instance.id !== currentInstanceId && instance.ideUrl !== "") {
+                    currentInstanceId = instance.id;
+                    window.location.href = instance.ideUrl;
+                    return;
+                }
+            } else {
+                currentInstanceId = instance.id;
+            }
             if (instance.status.phase === 'running') {
                 if (!hideDesktopIde) {
                     if (isDesktopIde == undefined) {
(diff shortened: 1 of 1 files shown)


akosyakov:
The page refresh was remove on purpose in the original PR.

geropl:
@akosyakov Hm. But I wonder if this signal could be used to trigger the logic again that is currently only run "onload":
 - re-fetch options
 - reveal the IDE if running

This would fix the re-start case, which I (tried to :roll_eyes: ) explain in #8226 :+1: 

@mustard-mh I would want to avoid re-load (to not trigger a re-direct from `ws-proxy`, which in turns closes the redirect-loop again), but rather try to run it's code again.

akosyakov:
As far as I remember on stopping/stopped event IDE is unloaded any only reliable way to load it again is to reload the page. Maybe it can be improved.

geropl:
> As far as I remember on stopping/stopped event IDE is unloaded any only reliable way to load it again is to reload the page. Maybe it can be improved.

Hm, ok. Also, maybe re-loading is not as bad as I thought when
 1.) we already are on the IDE url (as in this case)
 2.) we already know the workspace is actually running and we're connected (as in this case)

PR #8842, opened 2022-03-16
[dashboard] lint warnings: NewProject.tsx

trumbitta:
## Description

As per https://discord.com/channels/816244985187008514/885406100436951080/953232219261009950

This fixes the first bits of lint warnings.

Specifically: 
- adds ESLint and Prettier to the auto-installed VSCode extensions
- activates format on save for TypeScript and JavaScript
- adds a `.gp-link` Tailwind class to make buttons mimic links and avoid using `<a href="javascript:void(0)" ...>`
- fixes all the linting warnings for `projects/NewProject.tsx`

## Related Issue(s)

Re #3841 

## How to test

- launch the dashboard app
- try to create a new project: it should still work
- look in the console logs or in the terminal: there shouldn't be any lint warning for `projects/NewProject.tsx`
- if you are using VSCode, try opening a TypeScript file and check if "format on save" via the Prettier extension is working (remember Prettier won't work if there are syntax errors - red wiggles - in the file)

## Release Notes

```release-note
- [devx]: Added ESLint and Prettier to the auto-installed VSCode extensions
- [devx]: Activated "format on save" for TypeScript and JavaScript
- [dashboard]: Added a `.gp-link` Tailwind class to make buttons mimic anchor tags
- [dashboard]: Fixed all the linting warnings for `projects/NewProject.tsx`
```

PR #8995, opened 2022-03-29
[dashboard] lint warnings: NewProject.tsx

trumbitta:
## Description

As per https://discord.com/channels/816244985187008514/885406100436951080/953232219261009950

This fixes the first bits of lint warnings.

Specifically: 
- adds ESLint and Prettier to the auto-installed VSCode extensions
- activates format on save for TypeScript and JavaScript
- adds a `.gp-link` Tailwind class to make buttons mimic links and avoid using `<a href="javascript:void(0)" ...>`
- fixes all the linting warnings for `projects/NewProject.tsx`

## Related Issue(s)

Re #3841 

## How to test

- launch the dashboard app
- try to create a new project: it should still work
- look in the console logs or in the terminal: there shouldn't be any lint warning for `projects/NewProject.tsx`
- if you are using VSCode, try opening a TypeScript file and check if "format on save" via the Prettier extension is working (remember Prettier won't work if there are syntax errors - red wiggles - in the file)

## Release Notes

```release-note
- [devx]: Added ESLint and Prettier to the auto-installed VSCode extensions
- [devx]: Activated "format on save" for TypeScript and JavaScript
- [dashboard]: Added a `.gp-link` Tailwind class to make buttons mimic anchor tags
- [dashboard]: Fixed all the linting warnings for `projects/NewProject.tsx`
```

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -56,3 +56,5 @@ vscode:
     - timonwong.shellcheck
     - vscjava.vscode-java-pack
     - fwcd.kotlin
+    - dbaeumer.vscode-eslint
+    - esbenp.prettier-vscode
--- a/components/dashboard/src/projects/NewProject.tsx
+++ b/components/dashboard/src/projects/NewProject.tsx
@@ -4,7 +4,7 @@
  * See License-AGPL.txt in the project root for license information.
  */
 
-import { useContext, useEffect, useState } from "react";
+import { useCallback, useContext, useEffect, useState } from "react";
 import { getGitpodService, gitpodHostUrl } from "../service/service";
 import { iconForAuthProvider, openAuthorizeWindow, simplifyProviderName } from "../provider-utils";
 import { AuthProviderInfo, Project, ProviderRepository, Team, TeamMemberInfo, User } from "@gitpod/gitpod-protocol";
@@ -42,6 +42,29 @@ export default function NewProject() {
 
     const [authProviders, setAuthProviders] = useState<AuthProviderInfo[]>([]);
 
+    const updateReposInAccounts = useCallback(
+        async (installationId?: string) => {
+            setLoaded(false);
+            setReposInAccounts([]);
+            if (!selectedProviderHost) {
+                return [];
+            }
+            try {
+                const repos = await getGitpodService().server.getProviderRepositoriesForUser({
+                    provider: selectedProviderHost,
+                    hints: { installationId },
+                });
+                setReposInAccounts(repos);
+                setLoaded(true);
+                return repos;
+            } catch (error) {
+                console.log(error);
+            }
+            return [];
+        },
+        [selectedProviderHost],
+    );
+
     useEffect(() => {
         if (user && selectedProviderHost === undefined) {
             if (user.identities.find((i) => i.authProviderId === "Public-GitLab")) {
@@ -55,7 +78,7 @@ export default function NewProject() {
                 setAuthProviders(await getGitpodService().server.getAuthProviders());
             })();
         }
-    }, [user]);
+    }, [selectedProviderHost, user]);
 
     useEffect(() => {
         const params = new URLSearchParams(location.search);
@@ -69,7 +92,7 @@ export default function NewProject() {
             window.history.replaceState({}, "", window.location.pathname);
             setSelectedTeamOrUser(user);
         }
-    }, []);
+    }, [location.search, teams, user]);
 
     const [teamMembers, setTeamMembers] = useState<Record<string, TeamMemberInfo[]>>({});
     useEffect(() => {
@@ -123,10 +146,34 @@ export default function NewProject() {
     }, [selectedRepo]);
 
     useEffect(() => {
+        const createProject = async (teamOrUser: Team | User, repo: ProviderRepository) => {
+            if (!selectedProviderHost) {
+                return;
+            }
+            const repoSlug = repo.path || repo.name;
+
+            try {
+                const project = await getGitpodService().server.createProject({
+                    name: repo.name,
+                    slug: repoSlug,
+                    cloneUrl: repo.cloneUrl,
+                    account: repo.account,
+                    provider: selectedProviderHost,
+                    ...(User.is(teamOrUser) ? { userId: teamOrUser.id } : { teamId: teamOrUser.id }),
+                    appInstallationId: String(repo.installationId),
+                });
+
+                setProject(project);
+            } catch (error) {
+                const message = (error && error?.message) || "Failed to create new project.";
+                window.alert(message);
+            }
+        };
+
         if (selectedTeamOrUser && selectedRepo) {
             createProject(selectedTeamOrUser, selectedRepo);
         }
-    }, [selectedTeamOrUser, selectedRepo]);
+    }, [selectedTeamOrUser, selectedRepo, selectedProviderHost]);
 
     useEffect(() => {
         if (reposInAccounts.length === 0) {
@@ -155,7 +202,7 @@ export default function NewProject() {
         (async () => {
             await updateReposInAccounts();
         })();
-    }, [selectedProviderHost]);
+    }, [selectedProviderHost, updateReposInAccounts]);
 
     useEffect(() => {
         if (project && sourceOfConfig) {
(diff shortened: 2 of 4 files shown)


trumbitta:
@jankeromnes this is the new #8842 🤞

components/dashboard/src/projects/NewProject.tsx
```diff
@@ -42,6 +42,29 @@ export default function NewProject() {
 
     const [authProviders, setAuthProviders] = useState<AuthProviderInfo[]>([]);
 
+    const updateReposInAccounts = useCallback(
```
jankeromnes, line 45: Is it really necessary to use `useCallback` here? I believe we haven't used it elsewhere in the dashboard yet, and if I understand it correctly, calling `updateReposInAccounts` from the `Effect` with the dependency `selectedProviderHost` already ensures that we call it only once every time the dependency changes.

I think I'd prefer not using this feature here, to limit the "magic" in our code. What do you think? 🙂

.gitpod.yml
```diff
@@ -56,3 +56,5 @@ vscode:
     - timonwong.shellcheck
     - vscjava.vscode-java-pack
     - fwcd.kotlin
+    - dbaeumer.vscode-eslint
+    - esbenp.prettier-vscode
```
jankeromnes, line 60: @filiptronicek Would you happen to know if any of these extensions are already installed by default in VS Code?

components/dashboard/src/projects/NewProject.tsx
```diff
@@ -69,7 +92,7 @@ export default function NewProject() {
             window.history.replaceState({}, "", window.location.pathname);
             setSelectedTeamOrUser(user);
         }
-    }, []);
+    }, [location.search, teams, user]);
```
jankeromnes, line 95: I'm not 100% sure whether this change makes sense:
- Currently, we only call this effect once, and we immediately "consume" the search parameter if found (i.e. if you visit `/new?team=1234`, we immediately select team `1234` and overwrite the URL to `/new`). This means that, if we call this code multiple times, only the first time can have any effect.
- Calling it again if `location.search` changes sort of makes sense, but can it really happen that `location.search` changes without a reload? (I guess it could happen if we have something like `<Link to="/new?team=1234"/>` 🤔)
- Can `teams` or `user` really change here? If `teams` or `user` can be something unexpected, I guess we'd need to check for that in the beginning of the effect, and return early if the values are unexpected (e.g. undefined or empty array?)

All in all, since this code currently seems to work in all cases, I'm a bit hesitant to change the effect logic here.

components/dashboard/src/projects/NewProject.tsx
```diff
@@ -166,30 +213,10 @@ export default function NewProject() {
                 await getGitpodService().server.triggerPrebuild(project.id, null);
             })();
         }
-    }, [project, sourceOfConfig]);
+    }, [guessedConfigString, project, sourceOfConfig]);
```
jankeromnes, line 216: Reading the code, I think depending on `guessedConfigString` in addition to `sourceOfConfig` has the same effect. No objection here if it makes the linter happy. 🙂

gitpod-ws.code-workspace
```diff
@@ -1,60 +1,69 @@
...
+    ],
+    "settings": {
+        "typescript.tsdk": "gitpod/node_modules/typescript/lib",
+        "[javascript]": {
+            "editor.defaultFormatter": "esbenp.prettier-vscode",
+            "editor.formatOnSave": true
+        },
+        "[typescript]": {
+            "editor.defaultFormatter": "esbenp.prettier-vscode",
+            "editor.formatOnSave": true
```
jankeromnes, line 39: Thank you for this ❤️ I've been wanting to have this team-wide for a long time.

jankeromnes (commented):
Hi @trumbitta, many thanks for improving Gitpod! 🔥 (and also for your patience on our code reviews 😇)

I really like most of the changes proposed here, and think they are quite valuable improvements. 👍

But, since I'm a bit uncertain about a few items, I've left some questions/remarks in-line. Please take a look at them when you have some time, I'm looking forward to your thoughts on these.

.gitpod.yml
filiptronicek: @jankeromnes neither of these are bundled with VS Code by default [[1](https://github.com/gitpod-io/openvscode-server/blob/6cab367ef24e7f0ea0f1600b237519a56b75dcb5/product.json#L678-L739)] (not even in Microsoft-distributed VS Code), but `dbaeumer.vscode-eslint` is recommended for users when they first open JS-like files [[1](https://github.com/gitpod-io/openvscode-server/blob/6cab367ef24e7f0ea0f1600b237519a56b75dcb5/product.json#L110)]

jankeromnes: Many thanks! Good as is then 😁 👍

components/dashboard/src/projects/NewProject.tsx
trumbitta: Oh it was actually a cascading effect 🙂

- Without the `useCallback`, we have this linting warning: "The 'updateReposInAccounts' function makes the dependencies of useEffect Hook (at line 203) change on every render. To fix this, wrap the definition of 'updateReposInAccounts' in its own useCallback() Hook"
- Leaving `updateReposInAccounts` out of the dependencies of that `useEffect` to silence the previous warning, we have this other warning: "React Hook useEffect has a missing dependency: 'updateReposInAccounts'. Either include it or remove the dependency array."
- Removing the dependency array altogether to silence the previous warning, of course makes the hook execute on every render, while keeping it empty generates the linting warning about not having empty dependency arrays when we are using some dependencies inside the hook.

The only alternative seems to silence the ESLint rule for this line, but as long as the code still works as intended I'd rather follow the lead of the linting rules and move on.

Let me know what you want me to do!

components/dashboard/src/projects/NewProject.tsx
trumbitta: Yup, same reasoning as in https://github.com/gitpod-io/gitpod/pull/8995/files#r839619341

components/dashboard/src/projects/NewProject.tsx
jankeromnes: Thanks! I generally agree with you, but I believe I had one bad experience where I just followed ESLint and it broke my component 😬 (or, maybe the bug was my fault and I now wrongly blame ESLint for it). I guess this calls for more testing. 👀

components/dashboard/src/projects/NewProject.tsx
jankeromnes: If I'm not mistaken, you've already added the `updateReposInAccounts` closure as a dependency to the `useEffect`. Is that not enough to resolve the warning? (I.e. without needing the extra `useCallback`?)

jankeromnes: Or, actually I really don't like this pattern of implementing functions that are only called once. What about moving all the `updateReposInAccounts` code back inline into the `useEffect`? (Maybe as an in-line async function if needed.)

trumbitta: > What about moving all the updateReposInAccounts code back inline into the useEffect? (Maybe as an in-line async function if needed.)

Oh that's also my usual first move 🤝 but in this case `updateReposInAccounts` is also used elsewhere outside the `useEffect`.
Avoidable? Probably yes, but I think it will come with future refactorings 🔮

trumbitta: > If I'm not mistaken, you've already added the `updateReposInAccounts` closure as a dependency to the `useEffect`. Is that not enough to resolve the warning? (I.e. without needing the extra `useCallback`?)

Yes, but we have another warning :D look at the first of my bullet points above.

> Without the useCallback, we have this linting warning: "The 'updateReposInAccounts' function makes the dependencies of useEffect Hook (at line 203) change on every render. To fix this, wrap the definition of 'updateReposInAccounts' in its own useCallback() Hook"

trumbitta: That's because without `useCallback` the function get created on every render, and that triggers the dependency array of the `useEffect`

components/dashboard/src/projects/NewProject.tsx
trumbitta: I think it happened to me once!
I don't remember the specifics but like you said turns out it was my "fault" meaning I then applied "more React" to the problem (IIRC more composition and maybe a custom hook 🤔 ) and that fixed both the warning and the bug 🐞

trumbitta:
🎶 And we're back! 🎶 Welcome to this new episode of "Weekly pings to @jankeromnes!"

💖

Hey, any news? I see we also managed to wait enough for the usual merge conflicts to show up 🤣

jankeromnes:
Hi @trumbitta! Thank you again for all these improvements, and for keeping up the pings to make sure that they get the attention they very much deserve.

Most of these changes are actually fine and could be merged as is, but I'm very nervous about the effect dependency changes, so I would like to spend more time investigating & testing those specifically.

(On a side note, if this had been a few separate Pull Requests, I believe I could have approved and merged most of them, while only needing more time for reviewing the effect dependency changes. 💭)

trumbitta:
No problem, how do you want me to split them?

I can close this one and open the new ones. 🤝

trumbitta:
Unstale

trumbitta:
Hey 😊 

I think I'm going to close this and redo everything bit by bit (see #9269 and #9641), with the tiniest possible PRs while @gtsiolis drinks wine and smiles at my general direction, because the plan we hatched with @geropl on https://discord.com/channels/816244985187008514/953232219261009950/953667195089588305 didn't stand the test of time 👴

gtsiolis:
@trumbitta 🙂 
6ear4m

PR #8996, opened 2022-03-29
[install]: document installing via kots

### LINKED ISSUE #8329: Create and publish a "free" license
A generic "free" Replicated license needs to created for users to download and consume. This will then need to be published to the Gitpod self-hosted docs.

In Replicated parlance, this will be a community license. However, this will be the Gitpod free license for 10 users and the Gitpod community license for more than 10 users - see [docs](https://www.gitpod.io/self-hosted)

As this will be used by many, unlinked installations, this needs to be independent of the domain

### LINKED ISSUE #8410: [license] Make the domain field optional
Currently, we have a domain field in the license and we check that the domain matches the Gitpod domain. @metcalfc already mentioned that he would like to not ask for a domain when a customer approaches and with Replicated, we have at least some control about the distribution of the license as we see the number of installations in the vendor portal (I guess as long as it's not an air-gap installation).

I would propose to implement at least the Replicated license like this:
- Make the domain field in the license optional.
- When there is a domain given, the domain needs to match. Otherwise, the license is invalid.
- When there is no domain, this license is valid for all domains (no restriction).

With this implementation, we are quite flexible as we can set the domain field in the license or not (sales decision) and can change this at any time without code changes.

### LINKED ISSUE #8872: Publish first Stable release
Publish to `gitpod` slug (not `gitpod-pov`)

Once a release is published to `stable`,  #8329 and #8410 can be done

MrSimonEmms:
## Description

Add some base documentation for how to install Gitpod via kots. This also publishes the licenses for `stable`, `beta` and `unstable` channels.

This is not designed as a public-facing document (ie, for the website) but a technical quick-start. The website documentation will come later when we are confident in this process and our ability to support releases through kots.

The embedded installer (ie, [kURL](https://kurl.sh/)) is marked as a todo. I've left this in rather than staying silent on the subject as the kots docs talk about this, so seems a way of avoiding lots of "that's great, but how do I use kURL?" questions coming in.

The Installer docs are linked in here as an "advanced" method.

## Related Issue(s)

Fixes #8410
Fixes #8329 
Fixes #8872 

## How to test

Read docs and follow instructions.

## Release Notes

```release-note
[install]: document installing via kots
```


roboquat:

<details>

</details>

install/README.md
```diff
@@ -1,3 +1,48 @@
 # Install
 
 Resources for installing Gitpod
+
+## Replicated/KOTS
+
+Gitpod has teamed up with Replicated to simplify the installation of Gitpod for self-hosted users. This provides an installation interface, simplified license delivery for both community and paying users, easier upgrades and improved
```
Pothulapati, line 7: Looks like the sentence got cut-off in the middle 🤔

install/README.md
```diff
@@ -1,3 +1,48 @@
...
+ - [Beta](./licenses/Community%20(Beta).yaml)
+ - [Unstable](./licenses/Community%20(Unstable).yaml)
+
+### Installation
+
+#### Existing Cluster
+
+```shell
+curl https://kots.io/install | bash
+kubectl kots install gitpod
```
Pothulapati, line 37: Guessing it would take `stable` by default. How about adding a reference on using the channel here i.e `gitpod/unstable`. Not sure how users would know otherwise 🤔

Pothulapati (dismissed):
LGTM, and works well!

Also, Nice community edition flag in the UI when these licenses are used

image

install/README.md
MrSimonEmms: Do you know what, even though the KOTS docs actually say that I tried putting `kubectl kots install gitpod/unstable` and using the `beta` license - and it seems to install whatever the license says, rather than whatever the install command says

install/README.md
MrSimonEmms: Lucky I'm not a native English speaker...

Alt Text

install/README.md
Pothulapati: AH NICE! Makes sense! Then I think the installation command doesn't need any of this!

MrSimonEmms: Yeah, agreed. I've asked Barry and Todd from Replicated to cast their eyes over this to ensure I've not made any factual errors.

PR #8783, opened 2022-03-14
WIP: [dashboard] display warning for latest IDE versions

mustard-mh:
## Description

display warning for latest IDE versions on the loading screen for latest versions like VS Code Insiders and EAP of JB products

- [ ] Add test cases
- [ ] Need design (message detail, display position, adjust if content height too long

~~I can split commits into `format` and `code` if you want. done~~

## Related Issue(s)

Fixes #8028 

## How to test

1. Switch your Preferences
2. Open workspace, see if warning display
image

## Release Notes

```release-note
Display warning message when users choose latest IDE
```

Issue #7613, opened 2022-01-14
Improve warning alert component design


gtsiolis:
### Problem to solve

:a: The first version of the warning alert component design was based on high saturated colors[[1](https://en.wikipedia.org/wiki/Colorfulness)] and didn't adapt on dark theme colors (see screenshots below) as we did with labels in https://github.com/gitpod-io/gitpod/pull/4607.

| Delete deletion modal on light theme ☀️ | Delete deletion modal on dark theme  🌔 |
|-|-|
| <img width="1440" alt="Screenshot 2022-01-14 at 6 02 14 PM (2)" src="https://user-images.githubusercontent.com/120486/149554777-9874372f-1280-4840-b7a9-9ddb4ca198db.png"> | <img width="1440" alt="Screenshot 2022-01-14 at 6 02 04 PM (2)" src="https://user-images.githubusercontent.com/120486/149554780-28004770-4e00-43c2-9fcb-f59d349c0931.png"> |

| Labels on light theme ☀️ | Labels on dark theme ☀️ |
|-|-|
| ![image](https://user-images.githubusercontent.com/24960040/123451112-b2409e00-d5dd-11eb-8ae2-f2d581a7572d.png) | ![image](https://user-images.githubusercontent.com/24960040/123451032-9ccb7400-d5dd-11eb-9b96-57f48128eedc.png) |

:b: In the latest design specs using ther warning alert component, we've tried to improve the colors of the warning alert component, see example in https://github.com/gitpod-io/gitpod/pull/7295#discussion_r781620284.

| Proposed design for the warning alert component |
|-|
|  <img width="1440" alt="148848588-dddd6c66-8e7f-4a48-a0c6-61e8055d8574" src="https://user-images.githubusercontent.com/120486/149555791-b6d53963-ba36-467d-9832-e7119f1f77fe.png"> |

### Proposal

Since this is using the [`AlertBox`](https://github.com/gitpod-io/gitpod/blob/main/components/dashboard/src/components/AlertBox.tsx) component, we could update this once and improve the UX of this component across the product.

Next steps could include improving the rest of alerts by clearly defining the available types (info, warning, error, etc) and documenting when to use which.

jldec:
Assigned to @jankeromnes per https://github.com/gitpod-io/gitpod/pull/7295#discussion_r786166821

jldec:
Removing from groundwork for now.

gtsiolis:
[...]

PR #9003, opened 2022-03-29
[dashboard] support connect via SSH

### LINKED ISSUE #7713: Epic: SSH access copy/paste via dashboard

iQQBot:
## Description

Now that we have the new SSH gateway and the simpler authentication flow, it is possible that a user can copy/paste an SSH command onto their local machine to facilitate quicker/easier access to a Gitpod workspace. 

This PR introduces a way to direct connect ssh to workspace 

## Related Issue(s)

Fixes #7713

## How to test

start a workspace in preview environment, you can see Connect via SSH in workspace list and workspace start page(desktop IDE)

image

image

image

image

## Release Notes

```release-note
support direct connect workspace via ssh command
```

### DIFF
--- a/components/dashboard/src/start/StartWorkspace.tsx
+++ b/components/dashboard/src/start/StartWorkspace.tsx
@@ -27,6 +27,7 @@ import PendingChangesDropdown from "../components/PendingChangesDropdown";
 import { watchHeadlessLogs } from "../components/PrebuildLogs";
 import { getGitpodService, gitpodHostUrl } from "../service/service";
 import { StartPage, StartPhase, StartWorkspaceError } from "./StartPage";
+import ConnectToSSHModal from "../workspaces/ConnectToSSHModal";
 const sessionId = v4();
 
 const WorkspaceLogs = React.lazy(() => import("../components/WorkspaceLogs"));
@@ -91,6 +92,8 @@ export interface StartWorkspaceState {
         clientID?: string;
     };
     ideOptions?: IDEOptions;
+    isSSHModalVisible?: boolean;
+    ownerToken?: string;
 }
 
 export default class StartWorkspace extends React.Component<StartWorkspaceProps, StartWorkspaceState> {
@@ -520,6 +523,15 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
                                                 getGitpodService().server.stopWorkspace(this.props.workspaceId),
                                         },
                                         {
+                                            title: "Connect via SSH",
+                                            onClick: async () => {
+                                                const ownerToken = await getGitpodService().server.getOwnerToken(
+                                                    this.props.workspaceId,
+                                                );
+                                                this.setState({ isSSHModalVisible: true, ownerToken });
+                                            },
+                                        },
+                                        {
                                             title: "Go to Dashboard",
                                             href: gitpodHostUrl.asDashboard().toString(),
                                             target: "_parent",
@@ -556,6 +568,14 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
                                 </a>
                                 .
                             </div>
+                            {this.state.isSSHModalVisible === true && this.state.ownerToken && (
+                                <ConnectToSSHModal
+                                    workspaceId={this.props.workspaceId}
+                                    ownerToken={this.state.ownerToken}
+                                    ideUrl={this.state.workspaceInstance?.ideUrl.replaceAll("https://", "")}
+                                    onClose={() => this.setState({ isSSHModalVisible: false, ownerToken: "" })}
+                                />
+                            )}
                         </div>
                     );
                 }
--- a/components/dashboard/src/workspaces/ConnectToSSHModal.tsx
+++ b/components/dashboard/src/workspaces/ConnectToSSHModal.tsx
@@ -1,0 +1,109 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { useState } from "react";
+import Modal from "../components/Modal";
+import Tooltip from "../components/Tooltip";
+import copy from "../images/copy.svg";
+import AlertBox from "../components/AlertBox";
+import InfoBox from "../components/InfoBox";
+
+function InputWithCopy(props: { value: string; tip?: string; className?: string }) {
+    const [copied, setCopied] = useState<boolean>(false);
+    const copyToClipboard = (text: string) => {
+        const el = document.createElement("textarea");
+        el.value = text;
+        document.body.appendChild(el);
+        el.select();
+        try {
+            document.execCommand("copy");
+        } finally {
+            document.body.removeChild(el);
+        }
+        setCopied(true);
+        setTimeout(() => setCopied(false), 2000);
+    };
+    const tip = props.tip ?? "Click to copy";
+    return (
+        <div className={`w-full relative ${props.className ?? ""}`}>
+            <input
+                disabled={true}
+                readOnly={true}
+                autoFocus
+                className="w-full pr-8 overscroll-none"
+                type="text"
+                defaultValue={props.value}
+            />
+            <div className="cursor-pointer" onClick={() => copyToClipboard(props.value)}>
+                <div className="absolute top-1/3 right-3">
+                    <Tooltip content={copied ? "Copied" : tip}>
+                        <img src={copy} alt="copy icon" title={tip} />
+                    </Tooltip>
+                </div>
+            </div>
+        </div>
+    );
+}
+
+interface SSHProps {
+    workspaceId: string;
+    ownerToken: string;
+    ideUrl: string;
+}
+
+function SSHView(props: SSHProps) {
+    const sshCommand = `ssh ${props.workspaceId}#${props.ownerToken}@${props.ideUrl}`;
+    return (
+        <div className="border-t border-b border-gray-200 dark:border-gray-800 mt-2 -mx-6 px-6 py-6">
+            <div className="mt-1 mb-4">
+                <AlertBox>
+                    <p className="text-red-500 whitespace-normal text-base">
+                        <b>Anyone</b> on the internet with this command can access the running workspace. The command
+                        includes a generated access token that resets on every workspace restart.
+                    </p>
+                </AlertBox>
+                <InfoBox className="mt-4">
+                    <p className="text-gray-500 whitespace-normal text-base">
+                        Before connecting via SSH, make sure you have an existing SSH private key on your machine. You
+                        can create one using&nbsp;
+                        <a
+                            href="https://en.wikipedia.org/wiki/Ssh-keygen"
+                            target="_blank"
+                            rel="noopener noreferrer"
+                            className="gp-link"
+                        >
+                            ssh-keygen
+                        </a>
+                        .
+                    </p>
+                </InfoBox>
+                <p className="mt-4 text-gray-500 whitespace-normal text-base">
+                    The following shell command can be used to SSH into this workspace.
+                </p>
+            </div>
+            <InputWithCopy value={sshCommand} tip="Copy SSH Command" />
+        </div>
+    );
+}
+
+export default function ConnectToSSHModal(props: {
+    workspaceId: string;
+    ownerToken: string;
+    ideUrl: string;
+    onClose: () => void;
+}) {
+    return (
+        <Modal visible={true} onClose={props.onClose}>
+            <h3 className="mb-4">Connect via SSH</h3>
+            <SSHView workspaceId={props.workspaceId} ownerToken={props.ownerToken} ideUrl={props.ideUrl} />
+            <div className="flex justify-end mt-6">
+                <button className={"ml-2 secondary"} onClick={() => props.onClose()}>
+                    Close
+                </button>
+            </div>
+        </Modal>
+    );
+}
--- a/components/dashboard/src/workspaces/WorkspaceEntry.tsx
+++ b/components/dashboard/src/workspaces/WorkspaceEntry.tsx
@@ -24,6 +24,7 @@ import PendingChangesDropdown from "../components/PendingChangesDropdown";
 import Tooltip from "../components/Tooltip";
 import { WorkspaceModel } from "./workspace-model";
 import { getGitpodService } from "../service/service";
+import ConnectToSSHModal from "./ConnectToSSHModal";
 
 function getLabel(state: WorkspaceInstancePhase, conditions?: WorkspaceInstanceConditions) {
     if (conditions?.failed) {
@@ -42,6 +43,7 @@ interface Props {
 export function WorkspaceEntry({ desc, model, isAdmin, stopWorkspace }: Props) {
     const [isDeleteModalVisible, setDeleteModalVisible] = useState(false);
     const [isRenameModalVisible, setRenameModalVisible] = useState(false);
+    const [isSSHModalVisible, setSSHModalVisible] = useState(false);
     const renameInputRef = useRef<HTMLInputElement>(null);
     const [errorMessage, setErrorMessage] = useState("");
     const state: WorkspaceInstancePhase = desc.latestInstance?.status?.phase || "stopped";
@@ -49,6 +51,7 @@ export function WorkspaceEntry({ desc, model, isAdmin, stopWorkspace }: Props) {
         desc.latestInstance?.status.repo?.branch || Workspace.getBranchName(desc.workspace) || "<unknown>";
     const ws = desc.workspace;
     const [workspaceDescription, setWsDescription] = useState(ws.description);
+    const [ownerToken, setOwnerToken] = useState("");
 
     const startUrl = new GitpodHostUrl(window.location.href).with({
         pathname: "/start/",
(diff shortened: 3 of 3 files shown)


gtsiolis:
Looking at this now! 👀

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
```diff
@@ -0,0 +1,147 @@
...
+    }
+    const menuList: MenuItem[] = [
+        {
+            key: "manual",
+            component: (
+                <SSHManualView workspaceId={props.workspaceId} ownerToken={props.ownerToken} ideUrl={props.ideUrl} />
+            ),
+        },
+        {
+            key: "sshpass",
```
gtsiolis, line 117: **issue**: While this seems efficient do we really want to promote tools like this? Are you regularly using this, @iQQBot? I could be wrong but I assume many users will be against installing tools like this simply to use SSH access.

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
```diff
@@ -0,0 +1,147 @@
...
+                </p>
+            </div>
+            <div className="flex flex-col">
+                <div className="mt-2">
+                    <h4>Host</h4>
+                    <InputWithCopy value={props.ideUrl} />
+                </div>
+                <div className="mt-2">
+                    <h4>Username</h4>
+                    <InputWithCopy value={props.workspaceId} />
```
gtsiolis, line 71: **suggestion**: The user id can be part of the ssh command, right?

```sh
ssh user@xyz...
```

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
```diff
@@ -0,0 +1,147 @@
...
+                </div>
+                <div className="mt-2">
+                    <h4>Password</h4>
+                    <InputWithCopy value={props.ownerToken} />
```
gtsiolis, line 75: **issue**: Printing the password here feels like we're exposing information we could better hide, no?

components/dashboard/src/start/StartWorkspace.tsx
```diff
@@ -519,6 +522,15 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
                                             onClick: () =>
                                                 getGitpodService().server.stopWorkspace(this.props.workspaceId),
                                         },
+                                        {
+                                            title: "Connect via SSH",
```
gtsiolis, line 526: **praise**: Thanks for adding this here, too! ✨

gtsiolis (commented):
@iQQBot Thanks for pushing this forward! 🏀 

UX changes using the modal lood good.

However, I left some comments about UX and security practices here as these could affect the adoption of this feature.

Long-term exposing the password like this seems unessasary and can be avoided. I'd like to suggest the following approaches but I'm not sure if these will increase significantly the scope of this PR:

:a: Use the standard commands or binaries that ship with most operating system like `ssh`, `curl`, or something similar that could allow use to have a single command to copy and paster as well as authenticate via browser. Thus, avoiding and dropping all together `sshpass` support.
:b: Introduce an SSH Keys section in user settings so that users can upload their public SSH key as they usually do with GitLab, GitHub, etc sto that there's a single command to cipy and paste that is using the public key to connect securely to the workspace. This would require a new section in user settings as well as alert in the connect via SSH modal that warns this is a requirement or not set.

Looping in @akosyakov and @loujaybee to decide whether we'd like to a) ship this as is and iterate or b) step back and go with one of the :a: or :b: option described above. 🛹

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
iQQBot: Yes, the username can indeed be part of the ssh command, but what we need to know is that many tools split these

image
image

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
iQQBot: There are some complex reasons here
1. these passwords are one-time and will be changed when workspace is restarted
2. some browsers forbid js from copying, so we need a way for users to copy

csweichel:
> Long-term exposing the password like this seems unessasary and can be avoided. 

💯 

> I'd like to suggest the following approaches but I'm not sure if these will increase significantly the scope of this PR:

🛹 🛹 🛹 🛹 🛹 🛹 🛹 🛹 🛹 

> 
> 🅰️ Use the standard commands or binaries that ship with most operating system like `ssh`, `curl`, or something similar that could allow use to have a single command to copy and paster as well as authenticate via browser. Thus, avoiding and dropping all together `sshpass` support. 

We could offer some `curl https://... | sh` like experience. The URL folks would download would be a one-time secret (i.e. it can be downloaded only once). This way we could protect the owner token and would not make it part of logs and terminal history. We'd still need to find a way to pass the password to SSH - [sshpass is surprisingly complex](https://sourceforge.net/p/sshpass/code/HEAD/tree/tags/V1.09/main.c), but we could consider offering a static build so that people don't need to install it on their own.

Security-minded users would not execute anything `curl ... | sh` though. In that sense sshpass is way more transparent.

> 🅱️ Introduce an SSH Keys section in user settings so that users can upload their public SSH key as they usually do with GitLab, GitHub, etc sto that there's a single command to cipy and paste that is using the public key to connect securely to the workspace. This would require a new section in user settings as well as alert in the connect via SSH modal that warns this is a requirement or not set.

This is definitely where we want to go. It's also a considerable change, we'd need to
- add UI and API to accept the user's public key. Maybe we can pull this from GitHub (what about the other hoster)
- push the through server -> ws-manager -> pod -> ws-proxy
- support that key in the SSH gateway functionality of ws-proxy

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
iQQBot: We actually support workspaceID#ownerToken@ideUrl for single command connections

But this has to be based on the fact that the user implicitly provides a priavate_key which is the default id_rsa for authentication, we don't authenticate the user's private key, we just skip the step of having to enter a password through this mechanism. That said, if the user does not have any default privatekey, then the user will still be asked to enter the password.

We can change the behavior of the SSHGateway so that it can log in with any password if it provides full authentication information like workspaceID#ownerToken, but users must also be told this

https://github.com/gitpod-io/gitpod/pull/9003#discussion_r837702673 And with a program like this, the user will be confused as to how the input should be done

To ease the burden of education in this area, I chose to provide sshpass and `host` `username` `password` respectively.

CC @csweichel

csweichel: Wow, that's a neat hack :)

It is also a bit indirect and harder to understand.
Frankly I'd vote for going for the manual/sshpass options as they stand and make sure we have really good documentation for both cases in place.

akosyakov:
I would start with something simple like single `ssh` command  with a warning about private keys. As a next step add SSH keys to improve on it. I am not sure about providing other options I would prefer to see users using the simple method, i.e. run some usability testing.

akosyakov:
> Who can help me write this warning text? Do we need to explain to the user why a default private key is needed? @akosyakov

I remember you mentioned also that ssh supports an option to use only password based auth without keys? 

> https://github.com/gitpod-io/gitpod/pull/9003#discussion_r837702673 I think these situations still need to be considered, if the user is a professional user that is no problem, but if the user is not a professional user they will be very confused

I am not sure yet what our users are going to do with it. We could do later it too, no? We should get out something simple and run some usability tests.

akosyakov:
> Yes, ssh supports password-only authentication, but the ssh command does not support passing passwords, which is the biggest problem, If you need to pass a password, you must use a tool like sshpass. StrictHostKeyChecking is used to skip the host fingerprint confirmation

ok, could we then instead provide a warning which says that a user should have at least one private key and point to some docs how to create it (maybe there is some nice section on ssh.com)?

akosyakov:
@iQQBot It worked nicely for me. 🙏  @gtsiolis Could you help with styling of dialog please?

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
```diff
@@ -0,0 +1,91 @@
+/**
+ * Copyright (c) 2021 Gitpod GmbH. All rights reserved.
```
akosyakov, line 2: 2022

components/dashboard/src/start/StartWorkspace.tsx
```diff
@@ -556,6 +568,14 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
                                 </a>
                                 .
                             </div>
+                            {this.state.isSSHModalVisible === true && this.state.ownerToken && (
```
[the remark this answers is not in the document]
iQQBot: May be redundant in actual role, but not in semantics

akosyakov:
@iQQBot please update `How to test` section, it is misleading right now.

loujaybee:
@iQQBot before we ship this, can you ensure to add an analytics tracking event? Ideally with a property that distinguishes between the usage on the dashboard and from the workspace loading/start page? 🙏

akosyakov:
> @gtsiolis can you help me improve this UX?

I would probably move a security warning below the field.

gtsiolis:
Looking at this now! 👀

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+                disabled={true}
+                readOnly={true}
+                autoFocus
+                className="w-full pr-8 overscroll-none"
+                type="text"
+                defaultValue={props.value}
+            />
+            <div className="cursor-pointer" onClick={() => copyToClipboard(props.value)}>
+                <div className="absolute top-1/3 right-3">
+                    <Tooltip content={copied ? "Copied!" : tip}>
```
gtsiolis, line 41: **suggestion**: Minor rephrase! 😁 

```suggestion
                    <Tooltip content={copied ? "Copied" : tip}>
```

We should probably also align other instances of this or create a reusable component in future iterations. However, feel free to leave this as is!

https://github.com/gitpod-io/gitpod/blob/00e08df666edff200edf760ba26f40803c5ebfde/components/dashboard/src/teams/Members.tsx#L296

https://github.com/gitpod-io/gitpod/blob/2f0b9462b0451f0974434fa36bd757420ef04e34/components/dashboard/src/settings/Teams.tsx#L682

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+        <div className="border-t border-b border-gray-200 dark:border-gray-800 mt-2 -mx-6 px-6 py-6">
+            <div className="mt-1 mb-4">
+                <p className="text-gray-500 whitespace-normal text-base">
+                    The following shell command can be used to SSH into this workspace.
+                </p>
+                <p className="mt-2 text-gray-500 whitespace-normal text-base">
+                    If you encounter a prompt to enter a password, please make sure you have a private key in&nbsp;
+                    <CodeText>~/.ssh/id_rsa</CodeText>. You can use command&nbsp;<CodeText>ssh-keygen</CodeText>&nbsp;to
+                    generate one.
+                </p>
```
gtsiolis, line 68: **suggestion**: What do you think of using an alert component here for sufacing this information and rephrasing the copy to avoid any confusion because of the different options available (different SSH key types) as well as different set ups (where different systems store SSH keys).

| BEFORE | AFTER |
|-|-|
| Screenshot 2022-04-05 at 12 40 19 PM | Screenshot 2022-04-05 at 1 04 48 PM |

**fyi**: The alert component is being refactored in https://github.com/gitpod-io/gitpod/pull/8783 by @mustard-mh, but it should be fine to use the existing info alert component variant. Re-posting from https://github.com/gitpod-io/gitpod/pull/8783#discussion_r833317085 for visibility:

> [FWIW](https://en.wiktionary.org/wiki/FWIW), we also have https://github.com/gitpod-io/gitpod/issues/7613 to tackle this component separately and also a few similar components that we could deprecate in favor of this, see [AlertBox](https://github.com/gitpod-io/gitpod/blob/main/components/dashboard/src/components/AlertBox.tsx) and [InfoBox](https://github.com/gitpod-io/gitpod/blob/main/components/dashboard/src/components/InfoBox.tsx). However deprecating those other component could be tackled separately.

gtsiolis: **suggestion**: Following the comment in https://github.com/gitpod-io/gitpod/pull/9003#issuecomment-1085719584, if we'd like to include the security warning about access, we could restructure the modal like the following:

| Proposed Modal |
|-|
| Screenshot 2022-04-05 at 1 38 51 PM |

I've also included a sentence about the temporary generated token used in the command. Let me know what you think. We could also move that sentence out of the alerts. 💭

gtsiolis (approved):
@iQQBot This works like black magic! 🔮 

We're almost at the finish line! 🏁

Left some UX comments below. I think we've made the right decision to drop other options like `sshpass` in favor of the simple and one-liner-command user can copy and paste to connect. 🤝 

Regarding the comment in https://github.com/gitpod-io/gitpod/pull/9003#discussion_r837702673, I think the goal of this feature is to access a workspace via a terminal, right? If someone wants to break down manually the command to figure out the username or host they are free to do so. 🤷

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+function SSHView(props: SSHProps) {
+    const sshCommand = `ssh ${props.workspaceId}#${props.ownerToken}@${props.ideUrl}`;
+    return (
+        <div className="border-t border-b border-gray-200 dark:border-gray-800 mt-2 -mx-6 px-6 py-6">
+            <div className="mt-1 mb-4">
+                <p className="text-gray-500 whitespace-normal text-base">
+                    The following shell command can be used to SSH into this workspace.
+                </p>
+                <p className="mt-2 text-gray-500 whitespace-normal text-base">
+                    If you encounter a prompt to enter a password, please make sure you have a private key in&nbsp;
```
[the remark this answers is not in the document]
iQQBot: just privatekey

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+                    <CodeText>~/.ssh/id_rsa</CodeText>. You can use command&nbsp;<CodeText>ssh-keygen</CodeText>&nbsp;to
```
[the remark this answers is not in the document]
iQQBot: yes 🙏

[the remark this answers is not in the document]
gtsiolis: Thanks for confirming, @iQQBot! Resolving this as the proposal in https://github.com/gitpod-io/gitpod/pull/9003#discussion_r842685302 no longer mentions a specific file name.

akosyakov:
@iQQBot new analytics events have to be added here: https://www.notion.so/gitpod/Tracking-Plan-ed27d1492da745b9b31711dd73dcbe57#6c68c725b7d744ebb0bd1593f37ed623

akosyakov:
I think it is alright to reuse events if they are stable, but we have to update [the epic ](https://github.com/gitpod-io/gitpod/issues/7713) to have proper info about how we can verify that a feature got discovered and used. It would be also cool to have a dashboard for it in Mixpanel.

loujaybee:
@iQQBot ah yep, I think that event should work, thanks! 🚀

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
gtsiolis: Marking this one as resolved, too.

gtsiolis (approved):
Looks great, @iQQBot! Left one last comment above.

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
```diff
@@ -0,0 +1,109 @@
...
+}
+
+function SSHView(props: SSHProps) {
+    const sshCommand = `ssh ${props.workspaceId}#${props.ownerToken}@${props.ideUrl}`;
+    return (
+        <div className="border-t border-b border-gray-200 dark:border-gray-800 mt-2 -mx-6 px-6 py-6">
+            <div className="mt-1 mb-4">
+                <p className="text-gray-500 whitespace-normal text-base">
+                    The following shell command can be used to SSH into this workspace.
+                </p>
```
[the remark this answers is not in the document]
iQQBot: go for A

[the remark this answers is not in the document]
gtsiolis: Sounds good! [Impeccable agreements](https://www.notion.so/gitpod/Core-Values-Differentiators-2ed4c2f93c84499b98e3b5389980992e#5c98845e99264744bf9c70ad710a1397)! 🤝

gtsiolis (approved):
UX looks great, @iQQBot! Let's [ship it](https://www.notion.so/gitpod/Core-Values-Differentiators-2ed4c2f93c84499b98e3b5389980992e#d766b4a2f4664dea9905ac29e65a5ccf)! 🛹

mustard-mh:
> UX looks great, @iQQBot! Let's [ship it](https://www.notion.so/gitpod/Core-Values-Differentiators-2ed4c2f93c84499b98e3b5389980992e#d766b4a2f4664dea9905ac29e65a5ccf)! 🛹

Oooh, we need robot to cover related keywords to our notion define. XD

geropl (approved):
LGTM

PR #9004, opened 2022-03-29
Redirect web page when instance changed

### LINKED ISSUE #8226: [supervisor] Ensure we fetch IDEOptions/desktopIDEConfig on re-start

### LINKED ISSUE #8990: One frame running phase for every workspace start
### Bug description

One frame running phase for every workspace start

This ~~is~~ part because of the time diff between `contentAvailable` and `ideReady`

This PR https://github.com/gitpod-io/gitpod/pull/8313 will wait until `contentAvailable` to get `.gitpod.yml` file to parse extensions and download extensions via URL first, and then start IDE

So `ideReady` will always be after `contentAvailable`, but after `contentAvailable`, workspace phase turns to `Running`.

We can verify it here https://gitpod.io/#github.com/mustard-mh/test/tree/hw/large-file which will download some large files

```yaml
vscode:
  extensions:
    - http://ipv4.download.thinkbroadband.com/1GB.zip
    - http://212.183.159.230/1GB.zip
    - https://speed.hetzner.de/10GB.bin
```

|Image|
|:-:|
|iShot2022-03-29 15 50 04|

### Steps to reproduce

Start a workspace

### Anything else?

_No response_

mustard-mh:
## Description

Display opening workspace when ide is not ready yet. Relates PR https://github.com/gitpod-io/gitpod/pull/8125

Refresh web page when instanceId changed to get ide config and ws works again

Need to verify
1. One frame running phase does not display (will display `Starting`, `Opening workspace...`
2. Stopped workspace URL contains `not_found` works after ws started
3. Stopped workspace works after ws started https://github.com/gitpod-io/gitpod/pull/9004#issuecomment-1084573937

## Related Issue(s)

Fixes #8990
Fixes #8226

## How to test

1. Start same workspace in tab 1\2\3 (tab1 <-> verify1/3, tab2 <-> verify2
2. After they started, stop workspace in dashboard
3. Wait until tabs 1\2\3 Stopped
4. Refresh tab2
5. Go to tab3 click `Open Workspace`
6. See if tabs 1\2\3 work fine

We can do negative test like this PR https://github.com/gitpod-io/gitpod/pull/8125 in staging here https://hw-fix-8990-crash.staging.gitpod-dev.com/workspaces which will make Insiders VSCode crash https://github.com/gitpod-io/gitpod/commit/bb705e2d9d851b0c3cefb3c6f84f4a392722edba

## Release Notes

```release-note
Fixed one frame running phase when starting workspace
Fixed IDE options display incorrectly when restarting the workspace
Fixed stopped workspace does not work again if it started in another place
```

### DIFF
--- a/components/dashboard/src/start/StartWorkspace.tsx
+++ b/components/dashboard/src/start/StartWorkspace.tsx
@@ -324,7 +324,11 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
 
         // Redirect to workspaceURL if we are not yet running in an iframe.
         // It happens this late if we were waiting for a docker build.
-        if (!this.props.runsInIFrame && workspaceInstance.ideUrl && !this.props.dontAutostart) {
+        if (
+            !this.props.runsInIFrame &&
+            workspaceInstance.ideUrl &&
+            (!this.props.dontAutostart || workspaceInstance.status.phase === "running")
+        ) {
             this.redirectTo(workspaceInstance.ideUrl);
             return;
         }
@@ -431,48 +435,7 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
                 }
                 if (!this.state.desktopIde) {
                     phase = StartPhase.Running;
-
-                    if (this.props.dontAutostart) {
-                        // hide the progress bar, as we're already running
-                        phase = undefined;
-                        title = "Running";
-
-                        // in case we dontAutostart the IDE we have to provide controls to do so
-                        statusMessage = (
-                            <div>
-                                <div className="flex space-x-3 items-center text-left rounded-xl m-auto px-4 h-16 w-72 mt-4 mb-2 bg-gray-100 dark:bg-gray-800">
-                                    <div className="rounded-full w-3 h-3 text-sm bg-green-500">&nbsp;</div>
-                                    <div>
-                                        <p className="text-gray-700 dark:text-gray-200 font-semibold w-56 truncate">
-                                            {this.state.workspaceInstance.workspaceId}
-                                        </p>
-                                        <a target="_parent" href={contextURL}>
-                                            <p className="w-56 truncate hover:text-blue-600 dark:hover:text-blue-400">
-                                                {contextURL}
-                                            </p>
-                                        </a>
-                                    </div>
-                                </div>
-                                <div className="mt-10 justify-center flex space-x-2">
-                                    <a target="_parent" href={gitpodHostUrl.asDashboard().toString()}>
-                                        <button className="secondary">Go to Dashboard</button>
-                                    </a>
-                                    <a
-                                        target="_parent"
-                                        href={
-                                            gitpodHostUrl
-                                                .asStart(this.props.workspaceId)
-                                                .toString() /** move over 'start' here to fetch fresh credentials in case this is an older tab */
-                                        }
-                                    >
-                                        <button>Open Workspace</button>
-                                    </a>
-                                </div>
-                            </div>
-                        );
-                    } else {
-                        statusMessage = <p className="text-base text-gray-400">Opening Workspace …</p>;
-                    }
+                    statusMessage = <p className="text-base text-gray-400">Opening Workspace …</p>;
                 } else {
                     phase = StartPhase.IdeReady;
                     const openLink = this.state.desktopIde.link;
--- a/components/supervisor/frontend/src/index.ts
+++ b/components/supervisor/frontend/src/index.ts
@@ -108,9 +108,19 @@ const toStop = new DisposableCollection();
     //#region current-frame
     let current: HTMLElement = loading.frame;
     let desktopRedirected = false;
+    let currentInstanceId = "";
     const nextFrame = () => {
         const instance = gitpodServiceClient.info.latestInstance;
         if (instance) {
+            // refresh web page when instanceId changed
+            if (currentInstanceId !== "") {
+                if (instance.id !== currentInstanceId && instance.ideUrl !== "") {
+                    currentInstanceId = instance.id;
+                    window.location.href = instance.ideUrl;
+                }
+            } else {
+                currentInstanceId = instance.id;
+            }
             if (instance.status.phase === 'running') {
                 if (!hideDesktopIde) {
                     if (isDesktopIde == undefined) {


geropl (dismissed):
We introduced this change to avoid endless loops where we would start a new workspace on each page load.

IMO the problem here is somewhere else.

mustard-mh:
> We introduced this change to avoid endless loops where we would start a new workspace on each page load.
> 
> IMO the problem here is somewhere else.

@geropl  Code this PR changed, only effect display, logic not changed

geropl:
@mustard-mh Sorry for not being more precise here (still struggling with the virous :dizzy_face: ).

We _need_ the buttons to be there in case of a restart. Also cmp. my later comment [here](https://github.com/gitpod-io/gitpod/issues/8990#issuecomment-1082735140).

akosyakov:
Eh, now we have a problem in another place:
- tab 1: start any workspace
- tab 2: open the same workspace
- note how both end up in the IDE
- stop the workspace
- note how both end up in "Stopped"
- open the workspace again from tab 2
- Note how:
  - both workspaces show up-to-date state
  - only tab 2 reveals the IDE
  - tab 1 still shows the up-to-date state and controls to open/reveal the IDE (with fresh credentials), but it just hangs in opening workspace... forever

geropl:
> tab 1 still shows the up-to-date state and controls to open/reveal the IDE (with fresh credentials), but it just hangs in opening workspace... forever

@akosyakov Ah, yes. That's why I originally combined this with #8226 . If we combine this with https://github.com/gitpod-io/gitpod/pull/8978 I think we should be good! :upside_down_face:

akosyakov:
ok, let's try, @mustard-mh Could you bring other changes?

components/dashboard/src/start/StartWorkspace.tsx
```diff
@@ -431,48 +435,7 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
                 }
                 if (!this.state.desktopIde) {
                     phase = StartPhase.Running;
-
```
mustard-mh, line 434: We can delete this condition because it has already been redirected

components/supervisor/frontend/src/index.ts
```diff
@@ -108,9 +108,20 @@ const toStop = new DisposableCollection();
...
+    let currentInstanceId = "";
     const nextFrame = () => {
         const instance = gitpodServiceClient.info.latestInstance;
         if (instance) {
+            // refresh web page when instanceId changed
+            if (currentInstanceId !== "") {
+                if (instance.id !== currentInstanceId && instance.ideUrl !== "") {
+                    currentInstanceId = instance.id;
+                    window.location.href = instance.ideUrl;
+                    return;
```
akosyakov, line 120: you don't need to return here, it will never happen

mustard-mh:
Also updated how to test section

mustard-mh:
@geropl Please have a look 🙏

andreafalzetti (approved):
I've tested all scenarios and it works as expected.

Great work @mustard-mh 🚀

components/dashboard/src/start/StartWorkspace.tsx
```diff
@@ -324,7 +324,11 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
 
         // Redirect to workspaceURL if we are not yet running in an iframe.
         // It happens this late if we were waiting for a docker build.
-        if (!this.props.runsInIFrame && workspaceInstance.ideUrl && !this.props.dontAutostart) {
+        if (
+            !this.props.runsInIFrame &&
+            workspaceInstance.ideUrl &&
+            (!this.props.dontAutostart || workspaceInstance.status.phase === "running")
```
geropl, line 330: :+1:

geropl: Unfortunately I don't have a reproducer for such a scenario... :thinking:

components/supervisor/frontend/src/index.ts
```diff
@@ -108,9 +108,19 @@ const toStop = new DisposableCollection();
...
     let desktopRedirected = false;
+    let currentInstanceId = "";
     const nextFrame = () => {
         const instance = gitpodServiceClient.info.latestInstance;
         if (instance) {
+            // refresh web page when instanceId changed
+            if (currentInstanceId !== "") {
+                if (instance.id !== currentInstanceId && instance.ideUrl !== "") {
+                    currentInstanceId = instance.id;
+                    window.location.href = instance.ideUrl;
```
geropl, line 119: :100:

components/dashboard/src/start/StartWorkspace.tsx
mustard-mh: running -> redirect ->   ws-proxy find pod -> ide -> failed -> stopped
                                ｜
                                └>  ws-proxy find no pod -> add not_found -> loop break

@geropl

mustard-mh: Loop only happens when DB shows workspace status as `running` forever

mustard-mh: Steps like Anton said https://github.com/gitpod-io/gitpod/pull/9004#issuecomment-1084573937 will repro this scenario, others has been test in prev PR https://github.com/gitpod-io/gitpod/pull/8125

components/dashboard/src/start/StartWorkspace.tsx
mustard-mh: If we are going to use `redirect` for running phase, we can remove this safely. If not, we can change it @geropl

components/dashboard/src/start/StartWorkspace.tsx
geropl: > I think after this PR, flow will be like this

@mustard-mh Thx for the flow picture, that explains it! :pray:  We have one failed step, but will break because we still respect `dontAutostart` when triggering the actual workspace start. :heavy_check_mark: :lotus_position:

geropl (approved):
LGTM - thx for fixing this! :muscle:

PR #9005, opened 2022-03-29
[registry-facade] Add IPFS support

### LINKED ISSUE #8186: Speed up image pulls using IPFS
Today we pull all images from a central registry. While this is the standard way, it can become a performance bottleneck when starting many different, possibly large images. We already optimise for layer re-use, yet image pull remains one of most time-consuming tasks during workspace startup.

One way we could try and improve the situation is by using IPFS to share layers across nodes within a cluster. This approach
- [is gaining tracking in the community](https://fosdem.org/2022/schedule/event/container_ipfs_image/)
- is compatible with eStargz
- [uses open standards and technology](https://github.com/ipfs/go-ipfs#some-things-to-try), i.e. we don't have to implement it all.

We need to solve two paths:
1. pushing layers into IPFS
  - [nerdctl](https://github.com/containerd/nerdctl) can push to IPFS. We could make the image-builder push to our local IPFS cluster once the image-builder lives on the workspace side. Pro: we just have to call a CLI as part of the image build. Downside: we need to give the image build workspaces access to our local cluster
  - registry facade can push layers on demand to IPFS ([client libraries exist](https://github.com/ipfs/go-ipfs-api)). Pro: straight forward caching implementation whereby IPFS becomes a mere optimisation. Downside: we have to implement this
2. pulling layers from IPFS
  - the stargz snapshotter can [pull from IPFS directly](https://github.com/containerd/stargz-snapshotter/pull/496) - registry-facade could add `ipfs://` URLs to the manifest.
  - we could make registry-facade try and pull from IPFS, thereby providing a standard-compliant path towards the container runtime

csweichel:
## Description
This PR adds IPFS support to registry-facade. This is an optional feature and
- must be explicitly enabled in the config,
- requires IPFS and Redis to run within the cluster,
- has no support for GC, i.e. everything that's uploaded to IPFS remains there,
- has no observability for this feature,
- has unclear performance and stability characteristics: it's unclear how much overhead the Redis `GET` on each manifest production incurs.

Really, this is something that came out of a four our session from a hackathon.

### How it works
Whenever a blob gets pulled through registry-facade, registry-facade will upload that blob to IPFS and record the IPFS CID in Redis with the sha hash of the blob as key. The next time a manifest which contains the blob is produced, we'll add an `ipfs://` URL to the OCI descriptor of that blob. 

If the snapshotter supports IPFS URLs (like the [stargz snapshotter](https://github.com/containerd/stargz-snapshotter) does), image pulls should be accelerated. If the snapshotter does not support IFPS (like most other snapshoters), the image pull will likely fail because the snapshotter cannot handle ipfs URLs.

## Related Issue(s)

Fixes #8186 

## How to test
<a href="https://www.loom.com/share/64eee0456ee94e9689526a11229e925a">
    
  </a>

This PR is best tested locally (i.e. in a Gitpod workspace):

IPFS setup:
```bash
# install ipfs
cd /tmp
curl -L https://dist.ipfs.io/go-ipfs/v0.12.0/go-ipfs_v0.12.0_linux-amd64.tar.gz | tar xzv
cd go-ipfs
sudo sh install.sh

# setup and start daemon
ipfs init
jq '.Discovery.MDNS.Enabled=false'  ~/.ipfs/config > ~/.ipfs/config-new && mv ~/.ipfs/config-new ~/.ipfs/config
ipfs daemon
```

Redis setup:
```bash
docker run --name redis -p 6379:6379 -d redis
```

Tool setup:
```bash
go install github.com/csweichel/oci-tool@latest
```

Start registry-facade
```bash
cd components/registry-facade
go run main.go run example-config.json

# MANUAL STEP: please expose port 9090 publicly
```

Pull an image for the first time (make sure your docker damon cache is empty so that Docker actually pulls the blobs):
```bash
ref="$(gp url 9090 | cut -d '/'  -f 3)/fixed/foo:latest"

# inspect the manifest and observe that there are no IPFS URLs
oci-tool fetch manifest $ref

# pull the blobs which triggers the caching
docker pull $ref

# inspect the manifest and observe that now there are IPFS URLs
oci-tool fetch manifest $ref
```

## Release Notes

```release-note
Experimental IPFS support for cluster-local image caching
```


components/registry-facade/pkg/registry/ipfs.go
```diff
@@ -0,0 +1,88 @@
...
+	if store == nil || store.IPFS == nil || store.Redis == nil {
+		return nil
+	}
+
+	p, err := store.IPFS.Unixfs().Add(ctx, files.NewReaderFile(content), options.Unixfs.Pin(true), options.Unixfs.CidVersion(1))
+	if err != nil {
+		return err
+	}
+
+	res := store.Redis.Set(ctx, dgst.String(), p.Cid().String(), 36*time.Hour)
```
utam0k, line 80: Why do we set 36 hours?

sagor999, line 80: just curious, why expiration on the key is set to 36 hours?

components/registry-facade/pkg/registry/blob.go
```diff
@@ -146,6 +148,23 @@ func (bh *blobHandler) getBlob(w http.ResponseWriter, r *http.Request) {
...
+			if err != nil {
+				log.WithError(err).WithField("digest", bh.Digest).Warn("cannot push to IPFS")
+			}
+			if rc == nil {
+				log.WithField("digest", bh.Digest).Warn("cannot push to IPFS - blob is nil")
+				return
+			}
+			err = bh.IPFS.Store(context.Background(), bh.Digest, rc)
+			if err != nil {
+				log.WithError(err).WithField("digest", bh.Digest).Warn("cannot push to IPFS")
```
princerachit, line 164: Should we print err here?

components/registry-facade/pkg/registry/imagecfg.go
```diff
@@ -27,6 +27,14 @@ type ImageSpecProvider interface {
 	GetSpec(ctx context.Context, ref string) (*api.ImageSpec, error)
 }
 
+// FixedImageSpecProvider provides a single spec
+type FixedImageSpecProvider api.ImageSpec
+
+func (p FixedImageSpecProvider) GetSpec(ctx context.Context, ref string) (*api.ImageSpec, error) {
```
princerachit, line 33: Should we make p a pointer? `*p`?

components/registry-facade/pkg/registry/manifest.go
```diff
@@ -208,6 +210,14 @@ func (mh *manifestHandler) getManifest(w http.ResponseWriter, r *http.Request) {
 			manifest.Config.URLs = nil
 			manifest.Config.Size = int64(len(rawCfg))
 
+			// We might have additional modifications, e.g. adding IPFS URLs to the layers
+			if mh.ManifestModifier != nil {
+				err = mh.ManifestModifier(manifest)
+				if err != nil {
+					log.WithError(err).WithFields(logFields).Warn("cannot modify manifest")
```
princerachit, line 217: worth printing the error in the log statement?

MrSimonEmms (dismissed):

components/registry-facade/pkg/registry/ipfs.go
princerachit: is it because the max lifetime is 36 hours?

princerachit:
I see an error when trying to test
```console
gitpod /tmp/go-ipfs $ ipfs daemon
Initializing daemon...
go-ipfs version: 0.12.0
Repo version: 12
System version: amd64/linux
Golang version: go1.16.12
2022/03/30 07:41:34 failed to sufficiently increase receive buffer size (was: 208 kiB, wanted: 2048 kiB, got: 416 kiB). See https://github.com/lucas-clemente/quic-go/wiki/UDP-Receive-Buffer-Size for details.
2022/03/30 07:41:34 [zeroconf] no suitable IPv4 interface: udp4: failed to join any of these interfaces: []
2022/03/30 07:41:34 [zeroconf] no suitable IPv6 interface: udp6: failed to join any of these interfaces: []
2022-03-30T07:41:34.105Z        ERROR   cmd/ipfs        ipfs/daemon.go:430      error from node construction: no supported interface

Error: no supported interface
```

also, there might be a mismatch in jq version as it complains that the flag `-i` is not supported. However, this works:
```console
gitpod /tmp/go-ipfs $ jq '.Discovery.MDNS.Enabled=true'  ~/.ipfs/config > ~/.ipfs/config-new && mv ~/.ipfs/config-new ~/.ipfs/config
```

components/registry-facade/pkg/registry/manifest.go
csweichel: We are printing the error at the moment - should we not do that? If so, why?

components/registry-facade/pkg/registry/imagecfg.go
csweichel: That would make the use of the `FixedImageSpecProvider` more cumbersome. It's unfortunate that the message contains a mutex, but copying it is inconsequential here.

csweichel (commented):
Thanks all for the reviews :)
Had another go.

components/registry-facade/pkg/registry/ipfs.go
csweichel: changed it to 0 (indefinite) because we're not GC'in the IPFS content either.

princerachit:
> worth printing the error in the log statement? 

my bad. Wrong line comment.

components/registry-facade/pkg/registry/registry.go
```diff
@@ -170,12 +176,51 @@ func NewRegistry(cfg config.Config, newResolver ResolverProvider, reg prometheus
...
+
+		maddr, err := ma.NewMultiaddr(strings.TrimSpace(addr))
+		if err != nil {
+			return nil, xerrors.Errorf("cannot connect to IPFS: %w", err)
+		}
+		core, err := httpapi.NewApi(maddr)
+		if err != nil {
+			return nil, xerrors.Errorf("cannot connect to IPFS: %w", err)
+		}
+		rdc := redis.NewClient(&redis.Options{
```
aledbf, line 208: Please change it to `redis.NewSentinelClient`. If we commit to this, a single Redis instance is a no go from the start.
Also, please add Username and Password (same reason)

utam0k:
Looks good to me.

sagor999:
Code looks good, but like Prince I cannot test this for some reason, getting same error:
```
$ ipfs daemon
Initializing daemon...
go-ipfs version: 0.12.0
Repo version: 12
System version: amd64/linux
Golang version: go1.16.12
2022/03/31 00:11:56 failed to sufficiently increase receive buffer size (was: 208 kiB, wanted: 2048 kiB, got: 416 kiB). See https://github.com/lucas-clemente/quic-go/wiki/UDP-Receive-Buffer-Size for details.
2022/03/31 00:11:56 [zeroconf] no suitable IPv4 interface: udp4: failed to join any of these interfaces: []
2022/03/31 00:11:56 [zeroconf] no suitable IPv6 interface: udp6: failed to join any of these interfaces: []
2022-03-31T00:11:56.357Z        ERROR   cmd/ipfs        ipfs/daemon.go:430      error from node construction: no supported interface

Error: no supported interface
```

@csweichel am I missing a step? I an doing this in gitpod workspace, opening this PR.

csweichel:
My bad: I had the ipfs settings wrong in the description. The line should read:
```
jq '.Discovery.MDNS.Enabled=false'  ~/.ipfs/config > ~/.ipfs/config-new && mv ~/.ipfs/config-new ~/.ipfs/config
```

sagor999 (dismissed):
Tested. Looks good.

install/installer/pkg/config/v1/experimental/experimental.go
```diff
@@ -29,6 +29,19 @@ type WorkspaceConfig struct {
 		Limit            resource.Quantity `json:"limit"`
 		BurstLimit       resource.Quantity `json:"burstLimit"`
 	}
+
+	RegistryFacade struct {
```
aledbf, line 33: Why `registryFacade` is inside `workspace`?

csweichel: here `workspace` refers to the workspace team

MrSimonEmms (approved):
Looks good. From an Installer point of view, do we want to add an in-cluster Redis, or is that something that we're only explicitly supporting an external resource?

Also, I think this will be a good candidate as the first component that has "graduated" from `experimental` to the stable config. We're probably not there yet, but this is definitely going to be a useful addition for many of the bigger/enterprise self-hosted users

Furisto:
Rebased this on top of Simon's replicated changes, so that the werft build passes again.

aledbf (approved):
LGTM

MrSimonEmms (approved):

nandajavarma (approved):
LGTM :shipit:

PR #9014, opened 2022-03-30
WIP: Push to KOTS on each branch

MrSimonEmms:
## Description

It either pushes to the "unstable" channel if on the main branch, or the branch name if not. This makes KOTS a first-class citizen in the Gitpod testing process

Branches are cleaned up when no longer used via a cronjob

## Release Notes

```release-note
Push to KOTS on each branch
```


Pothulapati (approved):
LGTM! Initially, definitely worried that we'll get a lot of channels this way! but as each branch is a single PR, and a channel is how you install things (latest release on the channel), This way makes more sense!

> A future PR to auto-delete channels when unused may need to be done in future

We should definitely do this!

PR #9015, opened 2022-03-30
[idea] Ignore .idea/misc.xml which contains opened editor config

easyCZ:
## Description

When the IDE has panes open, it checks in cursor information into `.idea/misc.xml` with contents like 
```
 <component name="UnattendedHostPersistenceState">
    <option name="openedFilesInfos">
      <list>
        <OpenedFileInfo>
          <option name="caretOffset" value="0" />
          <option name="fileUrl" value="file://$PROJECT_DIR$/components/content-service/pkg/initializer/prebuild.go" />
        </OpenedFileInfo>
      </list>
    </option>
  </component>
```
This information is only meaningful to the local opened instance of the IDE and shouldn't be checked in. 

## Related Issue(s)

Fixes #

## How to test

1. Edit a file, run `git status` and see no changes are detected by git.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,6 @@ components/ws-proxy/ws-proxy
 
 # Logs
 components/server/*.log
+
+# Jetbrains workspace files
+.idea/misc.xml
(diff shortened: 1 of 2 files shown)


akosyakov:
@easyCZ I don't really looked at content of idea folder, as far as I understand this folder already has .gitignore file which should remove files unrelated to project config.

The main motivation of having `idea` folder to be checked in to enable discovery of gradle modules. I'm fine to ignore more files if backend-plugin and its dependencies are still auto discovered on start.

felladrin (approved):

PR #9016, opened 2022-03-30
.werft: Refactoring to use constant kubeconfig files instead of overrides

### LINKED ISSUE #8997: Refactor deploy-to-preview job to set explicit kubectx
Right now, we set the kubectx config for k3s in VMs to be the default:
https://github.com/gitpod-io/gitpod/blob/3d4e0e4ec4441c72566f2e0e451ea52063a3c50d/.werft/jobs/build/deploy-to-preview-environment.ts#L122-L125

This prevent us from accessing core-dev later.

Proposed solution:
* all invocations of `kubectl` should have an explicit `--kubeconfig <kubectxfile>` arg.
* there should not be a default kubecontext so that invocations of `kubectl` without explicit `--kubeconfig <kubectxfile>` fail. This is for robustness: don't accidentally use the wrong context.

ArthurSens:
## Description

This PR refactors the whole deployment job so all interactions with any cluster have to be made with the appropriate kubeconfig file explicit set. This is being done because we've been constantly overriding kubecontexts throughout the job's logic and we've got to a point where it is super confusing to know what is being applied where.

It is a huge change and too much effort to review only looking at the changes. I'm making sure it works by starting multiple jobs using different configurations and checking if Gitpod runs and is able to start workspaces afterwards.

## Related Issue(s)

Fixes #8997 

## How to test

- [ ] Start a preview on core-dev with Helm. Workspaces start.
- [x] Start a preview on core-dev with Installer. Workspaces start.
- [x] Start a preview on Harvester. Workspaces start.

## Release Notes

```release-note
NONE
```


.werft/jobs/build/prepare.ts
```diff
@@ -54,10 +54,11 @@ function configureDocker() {
...
 }
 
-function configureCoreDevAccess() {
-    const rc = exec('gcloud container clusters get-credentials core-dev --zone europe-west1-b --project gitpod-core-dev', { slice: prepareSlices.CONFIGURE_CORE_DEV }).code;
+function configureStaticClustersAccess() {
+    const rcCoreDev = exec(`KUBECONFIG=${CORE_DEV_KUBECONFIG_PATH} gcloud container clusters get-credentials core-dev --zone europe-west1-b --project gitpod-core-dev`, { slice: prepareSlices.CONFIGURE_CORE_DEV }).code;
+    const rcHarvester = exec(`cp /mnt/secrets/harvester-kubeconfig/harvester-kubeconfig.yml ${HARVESTER_KUBECONFIG_PATH}`, { slice: prepareSlices.CONFIGURE_CORE_DEV }).code;
 
-    if (rc != 0) {
+    if (rcCoreDev != 0 || rcHarvester != 0) {
```
mads-hartmann, line 61: I think it would be nice to split this into two separate checks so you can provide a more detailed error message when it fails.

So could be something like

```ts
const rcCoreDev = exec(`KUBECONFIG=${CORE_DEV_KUBECONFIG_PATH} gcloud container clusters get-credentials core-dev --zone europe-west1-b --project gitpod-core-dev`, { slice: prepareSlices.CONFIGURE_CORE_DEV }).code;

if (rcCoreDev != 0) {
  throw new Error("Failed to get core-dev kubeconfig credentials.")
}

const rcHarvester = exec(`cp /mnt/secrets/harvester-kubeconfig/harvester-kubeconfig.yml ${HARVESTER_KUBECONFIG_PATH}`, { slice: prepareSlices.CONFIGURE_CORE_DEV }).code;

if (rcHarvester != 0) {
  throw new Error("Failed to get Harvester kubeconfig credentials.")
}

```

.werft/observability/monitoring-satellite.ts
```diff
@@ -7,7 +7,7 @@ import * as fs from 'fs';
  * Monitoring satellite deployment bits
  */
  export class InstallMonitoringSatelliteParams {
```
mads-hartmann, line 9: This would have been nicer to just have as a `type` instead of class.

.werft/observability/monitoring-satellite.ts
```diff
@@ -76,39 +76,39 @@ export async function installMonitoringSatellite(params: InstallMonitoringSatell
 
     // The correct kubectl context should already be configured prior to this step
     // Only checks node-exporter readiness for harvester
-    ensureCorrectInstallationOrder(params.satelliteNamespace, params.withVM)
+    ensureCorrectInstallationOrder(params.kubeconfigPath, params.satelliteNamespace, params.withVM)
 }
 
-async function ensureCorrectInstallationOrder(namespace: string, checkNodeExporterStatus: boolean){
+async function ensureCorrectInstallationOrder(kubeconfig: string, namespace: string, checkNodeExporterStatus: boolean){
```
mads-hartmann, line 82: For this file you could have exposed a single class and then have all the functions as public/private methods. That way you wouldn't need to pass the kubeconfig around.

I don't think it's worth doing now, just as inspiration for future refactoring if you're feeling like it 😉

.werft/jobs/build/deploy-to-preview-environment.ts
```diff
@@ -99,18 +99,18 @@ export async function deployToPreviewEnvironment(werft: Werft, jobConfig: JobCon
         withVM,
     };
 
-    exec(`kubectl --namespace keys get secret host-key -o yaml > /workspace/host-key.yaml`)
+    exec(`kubectl --kubeconfig ${CORE_DEV_KUBECONFIG_PATH} --namespace keys get secret host-key -o yaml > /workspace/host-key.yaml`)
```
mads-hartmann, line 102: To be on the 100% safe side it would be nicest to quote the path, so it is `--kubeconfig "${CORE_DEV_KUBECONFIG_PATH}"`. This goes for all of the places where we are using a path in an `exec` call

mads-hartmann (approved):
This is really impressive Arthur 👏  I left a few comments, but I don't think either of them are blocking. The biggest one is about quoting the paths we pass to exec. As we have full control over the paths in our source code I don't think it's worth making the change but I'll leave it up to you to decide.

I added the blocked label so you can test all the various scenarios.

.werft/observability/monitoring-satellite.ts
ArthurSens: Yeah, that definitely crossed my mind while doing this, maybe its worth raising an issue to refactor this later

ArthurSens:
Oh damn, I spent way too much time trying to understand how my changes could have made helm to fail, but it is also failing in `main` 🥲

ArthurSens:
Sorry team-webapp, I'm just cherrypicking Janx's commit to test a deployment with Helm, I'll drop the commit afterwards

meysholdt (approved):

Issue #9020, opened 2022-03-30
VS Code Release 1.66


filiptronicek:
Endgame: https://github.com/microsoft/vscode/issues/145549
Release: https://github.com/microsoft/vscode/releases/tag/1.66.0

MS release date: March 30, 2022

DRI: @filiptronicek @mustard-mh 
Plan:

- [x]  Create corresponding release branches for `main` and `gp-code/main`
    - [x] Create release branch `release/1.66`
    	- Execute the rebase from upstream: `./scripts/sync-with-upstream.sh upstream/release/1.66 release/1.66`
    - [x] Create release branch `gp-code/release/1.66`
     	- Execute the rebase from upstream: `./scripts/sync-with-upstream.sh upstream/release/1.66 gp-code/release/1.66`
- [x]  Switch nightly jobs in https://github.com/gitpod-io/openvscode-releases and Gitpod Code-Nightly GitHub actions to point to the release branches
	- [x] https://github.com/gitpod-io/openvscode-releases/blob/1b60a53a1a34b61dfb41ca13b65bfbad4115bda4/.github/workflows/insiders-gp.yml#L19

    - [x] https://github.com/gitpod-io/openvscode-releases/blob/1b60a53a1a34b61dfb41ca13b65bfbad4115bda4/.github/workflows/insiders.yml#L19

    - [x] https://github.com/gitpod-io/gitpod/pull/9027 https://github.com/gitpod-io/gitpod/blob/4bdee21961c5390e1dc61606b7070af3b0f65971/.github/workflows/code-nightly.yaml#L34
- [x]  Create and merge PR in Gitpod repo to generate stable image for VS Code https://github.com/gitpod-io/gitpod/pull/9044
    - Use the template from https://gist.github.com/filiptronicek/be19dcab639a1cdf08089cc762377a41
    - [x]  [Smoke test](https://www.notion.so/Gitpod-VS-Code-1aa1dfcfdc5147869ec5ffcf86f430a6) version 1.66 in VS Code Insiders
- [x]  Create and merge PR in Gitpod repo updating VS Code stable image tags to image generated in previous step https://github.com/gitpod-io/gitpod/pull/9048
		https://github.com/gitpod-io/gitpod/blob/14a9636f476fc69294c5d0d9e7f831f16b39074f/chart/values.yaml#L361
[...]

PR #9024, opened 2022-03-30
Add new VS Code domain to config maps

filiptronicek:
## Description
This change is needed for VS Code versions after `1.66.0`, because they no longer use `vscode-webview.net` as the host domain for their webviews. Instead, they started using `vscode-cdn.net`. 

## Related Issue(s)
~This is needed for https://github.com/gitpod-io/gitpod/issues/9020~
This is not needed for VS Code 1.66 immediately, but it will be needed in the next release for stable and for insiders right after the release of 1.66.

## Release Notes

```release-note
NONE
```


jeanp413:
Also need to update in ops repo [here](https://github.com/gitpod-io/ops/blob/20029adf580c2b678001dc61d4987380928019ad/deploy/workspace/manifests/blobserve.yaml#L25)

aledbf (approved):
LGTM

PR #9045, opened 2022-03-31
[dashboard] Remove prebuild status errors from prebuild logs

easyCZ:
## Description

Removes injection of prebuild/workspace instance status errors into the logs. The error/status is already shown below the logs view.

Screenshot 2022-03-31 at 11 22 17
Screenshot 2022-03-31 at 11 19 27

Doing this addresses the following problems:
* Simplifies state checking logic
* Removes a race-condition where sometimes (fairly frequently) the error is injected at the top of the logs but with enough log output, the error is not in the view.

## Related Issue(s)

Fixes #

## How to test

1. Clone https://gitlab.com/gitpod-milan/gitpod-large-image
2. Setup the repo in step 1. as a project
3. Trigger pre-builds for the following example branches:
	* main
	* timeout
	* fail
4. Observe the error is not injected into the logs
5. Image build & logs continue to be streamed
6. Once built, logs should continue to be accessible

## Release Notes

```release-note
[dashboard] Remove prebuild error message from prebuild log view
```

## Documentation

/uncc
/hold

### DIFF
--- a/components/dashboard/src/components/PrebuildLogs.tsx
+++ b/components/dashboard/src/components/PrebuildLogs.tsx
@@ -85,58 +85,12 @@ export default function PrebuildLogs(props: PrebuildLogsProps) {
             props.onInstanceUpdate(workspaceInstance);
         }
         switch (workspaceInstance?.status.phase) {
-            // unknown indicates an issue within the system in that it cannot determine the actual phase of
-            // a workspace. This phase is usually accompanied by an error.
-            case "unknown":
-                break;
-
             // Preparing means that we haven't actually started the workspace instance just yet, but rather
             // are still preparing for launch. This means we're building the Docker image for the workspace.
             case "preparing":
-                getGitpodService().server.watchWorkspaceImageBuildLogs(workspace!.id);
-                break;
-
-            // Pending means the workspace does not yet consume resources in the cluster, but rather is looking for
-            // some space within the cluster. If for example the cluster needs to scale up to accomodate the
-            // workspace, the workspace will be in Pending state until that happened.
-            case "pending":
-                break;
-
-            // Creating means the workspace is currently being created. That includes downloading the images required
-            // to run the workspace over the network. The time spent in this phase varies widely and depends on the current
-            // network speed, image size and cache states.
-            case "creating":
-                break;
-
-            // Initializing is the phase in which the workspace is executing the appropriate workspace initializer (e.g. Git
-            // clone or backup download). After this phase one can expect the workspace to either be Running or Failed.
-            case "initializing":
-                break;
-
-            // Running means the workspace is able to actively perform work, either by serving a user through Theia,
-            // or as a headless workspace.
-            case "running":
-                break;
-
-            // Interrupted is an exceptional state where the container should be running but is temporarily unavailable.
-            // When in this state, we expect it to become running or stopping anytime soon.
-            case "interrupted":
-                break;
-
-            // Stopping means that the workspace is currently shutting down. It could go to stopped every moment.
-            case "stopping":
-                break;
-
-            // Stopped means the workspace ended regularly because it was shut down.
             case "stopped":
                 getGitpodService().server.watchWorkspaceImageBuildLogs(workspace!.id);
                 break;
-        }
-        if (workspaceInstance?.status.conditions.headlessTaskFailed) {
-            setError(new Error(workspaceInstance.status.conditions.headlessTaskFailed));
-        }
-        if (workspaceInstance?.status.conditions.failed) {
-            setError(new Error(workspaceInstance.status.conditions.failed));
         }
     }, [props.workspaceId, workspaceInstance?.status.phase]);
 


roboquat:

<details>

</details>

components/dashboard/src/components/PrebuildLogs.tsx
```diff
@@ -85,59 +85,13 @@ export default function PrebuildLogs(props: PrebuildLogsProps) {
...
         switch (workspaceInstance?.status.phase) {
-            // unknown indicates an issue within the system in that it cannot determine the actual phase of
-            // a workspace. This phase is usually accompanied by an error.
-            case "unknown":
-                break;
-
             // Preparing means that we haven't actually started the workspace instance just yet, but rather
             // are still preparing for launch. This means we're building the Docker image for the workspace.
             case "preparing":
-                getGitpodService().server.watchWorkspaceImageBuildLogs(workspace!.id);
```
jankeromnes, line 96: I don't understand this change. Why would you like to not subscribe to image build logs when watching the prebuild logs while an image build is in progress? 🤔

components/dashboard/src/components/PrebuildLogs.tsx
```diff
@@ -85,59 +85,13 @@ export default function PrebuildLogs(props: PrebuildLogsProps) {
...
-            case "running":
-                break;
-
-            // Interrupted is an exceptional state where the container should be running but is temporarily unavailable.
-            // When in this state, we expect it to become running or stopping anytime soon.
-            case "interrupted":
-                break;
-
-            // Stopping means that the workspace is currently shutting down. It could go to stopped every moment.
-            case "stopping":
```
jankeromnes, line 127: I'm curious to know your thoughts on removing all these "empty" cases. I get that they currently have no effect here, but we also have a history of "forgetting" to handle specific instance phases which are important to handle, so I initially felt like listing them all here, with appropriate comments, would make it easier not to forget about them in future refactors. 💭

AlexTugarev (approved):
Works as advertised!

components/dashboard/src/components/PrebuildLogs.tsx
```diff
@@ -85,59 +85,13 @@ export default function PrebuildLogs(props: PrebuildLogsProps) {
...
             case "stopped":
                 getGitpodService().server.watchWorkspaceImageBuildLogs(workspace!.id);
                 break;
         }
-        if (workspaceInstance?.status.conditions.headlessTaskFailed) {
-            setError(new Error(workspaceInstance.status.conditions.headlessTaskFailed));
-        }
-        if (workspaceInstance?.status.conditions.failed) {
-            setError(new Error(workspaceInstance.status.conditions.failed));
-        }
```
jankeromnes, line 140: If we no longer set these errors here (and no longer display them in the logs), I'm worried that we'll no longer display them anywhere.

If we're removing the explicit `conditions.headlessTaskFailed` and `conditions.failed` messages from here, wouldn't it make sense to explicitly include them in the status message? (Currently, that message seems to always be just "Prebuild failed")

https://github.com/gitpod-io/gitpod/blob/9cd88686371b26d727c6807bd4153d4b26a3f12a/components/dashboard/src/projects/Prebuilds.tsx#L464-L478

jankeromnes (commented):
Many thanks for this Pull Request! Had a pass through the code changes, and left a few thoughts/questions. Will now try to test this.

components/dashboard/src/components/PrebuildLogs.tsx
easyCZ: So we still do, we're just collapsing the cases. The diff looks confusing as it has no additions but we effectively remove all the empty `break` cases and only leave the "preparing" and "stopping" cases that had some logic

jankeromnes: Ah, right, it's a fall through. 👍

components/dashboard/src/components/PrebuildLogs.tsx
easyCZ: They are already shown in the footer of the Prebuild logs (by the parent) - see screenshot.

components/dashboard/src/components/PrebuildLogs.tsx
easyCZ: The reasoning:
* We have types that allow us to discover the cases
* They are documented against the underlying `gitpod-protocol` already, this is just duplication here.

components/dashboard/src/components/PrebuildLogs.tsx
jankeromnes: Ah, my mistake, my code snippet is from the deprecated `PrebuildInstanceStatus`.

I now see that the new `PrebuildStatusDescription` does embed `prebuild.error` in the message:

https://github.com/gitpod-io/gitpod/blob/9cd88686371b26d727c6807bd4153d4b26a3f12a/components/dashboard/src/projects/Prebuilds.tsx#L354-L360

and this error will include `conditions.failed` or `conditions.headlessTaskFailed` when appropriate:

https://github.com/gitpod-io/gitpod/blob/5ea808e7d43464829e183c1f93f95d3efe784c6f/components/ws-manager-bridge/ee/src/bridge.ts#L79-L81

https://github.com/gitpod-io/gitpod/blob/5ea808e7d43464829e183c1f93f95d3efe784c6f/components/ws-manager-bridge/ee/src/bridge.ts#L87-L90

(Also for what it's worth, that repeated `if` doesn't seem to make sense ☝️)

jankeromnes (commented):
Many thanks for responding to my comments/questions! They're now all addressed.

The code changes look good to me now. 👍

I also tried testing this, but for some reason, when I try my go-to image build error repo (https://gitlab.com/jankeromnes/gitpod-build-error), I never get any image build logs.

It always goes straight from "Queued" to "System Error", without showing any logs:

| BEFORE | AFTER |
| --- | --- |
| Screenshot 2022-04-01 at 10 33 12 | Screenshot 2022-04-01 at 10 30 43 |

jankeromnes (approved):
Approving, because the observed logs problem seems unrelated to this change, and you did get logs for my test repo.

Thanks also for the call! I can't wait for us to have much more robust Prebuild status and logs. 😊

easyCZ:
> It always goes straight from "Queued" to "System Error", without showing any image build logs:

@jankeromnes using your repo, I've managed to get the image build logs.

The case you're observing is a race condition that happens due to the following:
1. We only subscribe for image build logs if the WorksapceInstance is in "preparing" or "stopping" states. Of course, we're running a distributed system so we can easily not receive the update as fast, or skip it entirely. This means that we wouldn't always detect the "preparing" state so we wouldn't start the image build stream :/
2. If you refresh the browser, you may have already moved to the "prebuild failed" status in which case you don't get the image build logs.

I'm looking to tackle both by making the prebuild logs deterministic, but that's outside of the scope of this PR.

PR #9046, opened 2022-03-31
At/zzz-auth

AlexTugarev:


roboquat:

<details>

</details>

AlexTugarev:
/hold

... until https://github.com/gitpod-io/gitpod/pull/9046/commits/9ee97f3d5c32ec115ecb10b50ecbf8fcf0eac1ad is removed.

components/server/src/auth/host-context-provider-impl.ts
```diff
@@ -82,16 +82,19 @@ export class HostContextProviderImpl implements HostContextProvider {
...
 
             const existingContext = this.dynamicHosts.get(host);
-            const existingConfig = existingContext && existingContext.authProvider.params;
-            if (existingConfig && config.id === existingConfig.id) {
-                if (existingConfig.host !== config.host) {
+            if (existingContext) {
+                const existingConfig = existingContext.authProvider.params;
+                const sameId = config.id === existingConfig.id;
+                const sameHost = config.host === existingConfig.host;
+                if (!sameHost || !sameId) {
```
geropl, line 89: Idee that we now treat "different IDs" to be a broken state. Correct? Before we would treat them as different provider.

components/server/src/auth/host-context-provider-impl.ts
```diff
@@ -82,16 +82,19 @@ export class HostContextProviderImpl implements HostContextProvider {
...
                     log.warn("Ignoring host update for dynamic Auth Provider: " + host, { config, existingConfig });
                     continue;
                 }
-                if (existingConfig.status === config.status) {
-                    if (!!config.oauthRevision && existingConfig.oauthRevision === config.oauthRevision) {
-                        continue;
-                    }
+                const sameOwner = config.ownerId === existingConfig.ownerId;
+                const sameStatus = config.status === existingConfig.status;
+                const hasOAuthRevisionChanged =
```
geropl, line 95: hasOauthRevisionNOTChanged (or similar)

components/server/src/config.ts
```diff
@@ -195,6 +195,7 @@ export namespace ConfigFile {
             authProviderConfigs.push(...authProviderConfigFiles);
         }
         authProviderConfigs = normalizeAuthProviderParams(authProviderConfigs);
+        authProviderConfigs = []; // disable built-ins
```
geropl, line 198: Is this for debugging I assume?

geropl:
Looks good in general, please have a look at the comments for small fixes. Please ping again for ✔️

components/gitpod-db/src/typeorm/auth-provider-entry-db-impl.ts
```diff
@@ -91,8 +91,10 @@ export class AuthProviderEntryDBImpl implements AuthProviderEntryDB {
         return query.getMany();
     }
 
-    protected oauthContentHash(oauth: AuthProviderEntry["oauth"]): string {
-        const result = createHash("sha256").update(JSON.stringify(oauth)).digest("hex");
+    protected oauthContentHash(entry: AuthProviderEntry): string {
+        const result = createHash("sha256")
+            .update(JSON.stringify({ oauth: entry.oauth, ownerId: entry.ownerId, status: entry.status }))
```
geropl, line 96: Currently the code has no "update hash" path. If we change the meaning here we have to introduce one, or we get back into a state where we have not cashing at all!

components/server/src/auth/host-context-provider-impl.ts
```diff
@@ -82,16 +82,19 @@ export class HostContextProviderImpl implements HostContextProvider {
...
+                    !!config.oauthRevision && existingConfig.oauthRevision === config.oauthRevision;
+                if (sameOwner && sameStatus && hasOAuthRevisionChanged) {
```
geropl, line 97: Why do we jump over the fast path here (Jason.stringify), if we just checked all properties?

components/server/src/auth/host-context-provider-impl.ts
AlexTugarev: Right, that wasn't though through. Indeed it could happen, that one provider is removed and another (legit) was created in the same interval. 

Going to revert that.

components/server/src/auth/host-context-provider-impl.ts
AlexTugarev: going with `sameOAuthRevision`

components/server/src/auth/host-context-provider-impl.ts
AlexTugarev: good question. I don't recall, why the `JSON.stringify(existingConfig.oauth) === JSON.stringify(config.oauth)` was left in place. in theory, checking for the `sameOAuthRevision` should include it already.

components/server/src/config.ts
AlexTugarev: that's the test commit to be removed.

components/gitpod-db/src/typeorm/auth-provider-entry-db-impl.ts
AlexTugarev: Don't get the comment.

HostContextProviderImpl is cashing the runtime handlers of auth providers. The original optimization was to reduce costs of decryption of db entries and instantiation of those runtime handlers. 

The problem here to be solved is, that the revision didn't consider rarely `ownerId`, so we overshoot and ignored necessary updates. So this basically adds the missing update, but it does not re-introduce re-reading of everything.

components/server/src/auth/host-context-provider-impl.ts
AlexTugarev: I removed the `JSON.stringify` comparison, as I see no other paths which would tamper with the config object here, i.e. it's kind of useless double-checking.

components/server/src/config.ts
AlexTugarev: removed.

components/server/src/auth/host-context-provider-impl.ts
geropl: Originally the `if (nothingChanged) { continue; }` was preferrred, and the JSON compare served as fallback.

With your last change both (accidentally) got merged.

IMO it's ok to drop the fallback, as it should not be required anymore. :+1:

components/gitpod-db/src/typeorm/auth-provider-entry-db-impl.ts
geropl: > Don't get the comment.

Sorry for the confusion: did not read the code but answered from the top of my head - and that was outdated :see_no_evil: 

Hash gets updated [here](https://github.com/gitpod-io/gitpod/blob/45f13cf8e9214db664adc23a418e53d92364fbb7/components/server/src/auth/auth-provider-service.ts#L109) and [here](https://github.com/gitpod-io/gitpod/blob/45f13cf8e9214db664adc23a418e53d92364fbb7/components/server/src/auth/auth-provider-service.ts#L148).

MrSimonEmms:
Have checked and this works perfectly 👍🏻

MrSimonEmms (approved):

geropl (approved):
LGTM - thx for fixing this! :pray:

PR #9070, opened 2022-04-01
Add "publish-to-kots" flag in werft

mrsimonemms:
## Description

Add a `publish-to-kots=true` to Werft. This makes the publish to KOTS only happen when requested or on the `main` branch

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.werft/jobs/build/job-config.ts
+++ b/.werft/jobs/build/job-config.ts
@@ -16,6 +16,7 @@ export interface JobConfig {
     publishRelease: boolean;
     publishToJBMarketplace: string
     publishToNpm: string
+    publishToKots: boolean;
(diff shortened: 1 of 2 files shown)


meysholdt (approved):

ArthurSens:
👋 coming for a late review 😅 

Does it make sense to move `publish-to-kots` to be a method under `buildAndPublish`?

I'm thinking of keeping `built.ts` as small as possible, while putting logic where it makes sense 🙂 (I can create an issue for that if that's agreed

cc @gitpod-io/platform

MrSimonEmms:
Yeah, I'm happy with that as a proposed change

PR #9071, opened 2022-04-01
[kots]: change the CPU and memory checks to be min not sum

MrSimonEmms:
## Description

Whilst helping someone in the community, noticed that the [memory/CPU recommendations](https://www.gitpod.io/docs/self-hosted/latest/requirements#minimum-compute-resources) and "per node" rather than "per cluster".

This updates the preflight checks to reflect that.

## How to test

[Install](https://troubleshoot.sh/docs/#installation) the `preflight` plugin and run `kubectl preflight kots-preflight.yaml`

## Release Notes

```release-note
[kots]: change the CPU and memory checks to be min not sum
```

### DIFF
--- a/install/kots/manifests/kots-preflight.yaml
+++ b/install/kots/manifests/kots-preflight.yaml
@@ -103,24 +103,24 @@ spec:
           - pass:
               message: Cert-manager is installed and available.
     - nodeResources:
-        checkName: Total CPU Cores in the cluster
+        checkName: CPU Cores per node
         outcomes:
           - fail:
-              when: "sum(cpuCapacity) < 2"
+              when: "min(cpuCapacity) < 2"
               message: The cluster must contain at least 2 cores
           - warn:
-              when: "sum(cpuCapacity) < 4"
+              when: "min(cpuCapacity) < 4"
               message: The cluster must contain at least 4 cores
           - pass:
               message: There are at least 4 cores in the cluster
     - nodeResources:
-        checkName: Total memory in the cluster
+        checkName: Memory per node
         outcomes:
           - fail:
-              when: "sum(memoryCapacity) < 8G"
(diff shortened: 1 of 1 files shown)


install/kots/manifests/kots-preflight.yaml
```diff
@@ -106,21 +106,21 @@ spec:
...
+              when: "min(cpuCapacity) < 2"
               message: The cluster must contain at least 2 cores
           - warn:
-              when: "sum(cpuCapacity) < 4"
+              when: "min(cpuCapacity) < 4"
               message: The cluster must contain at least 4 cores
           - pass:
               message: There are at least 4 cores in the cluster
     - nodeResources:
         checkName: Total memory in the cluster
```
Pothulapati, line 117: These aren't checking for the total's of the cluster anymore, right? Should be named around the nodes?

corneliusludmann: @MrSimonEmms Any thoughts on this?

MrSimonEmms: @corneliusludmann sorry, I missed this.

As I see it, we can do one of two things:
1. put a filter in so that the workspace nodes act as above and the "meta" nodes have no `warn` section
2. leave as-is

Pothulapati: @MrSimonEmms I don't think my question was around `meta` vs `workspace` nodes. It was just to update the `checkName` to reflect it to be more about induvidual node's min requirement than the total. 🤔

MrSimonEmms: Oh right, yes I see

MrSimonEmms:
@Pothulapati I've also changed the wording on the CPU check as similarly wrong focus

Pothulapati (approved):
LGTM

PR #9076, opened 2022-04-01
.werft/build: Issue certificates during preparation phase

### LINKED ISSUE #8998: preview envs on Harvester: request certs on core-dev and copy them into VMs
Prerequisite: https://github.com/gitpod-io/gitpod/issues/8998

1. preview envs can launch faster if cert-requests run in parallel to the build and are not on the hot path
2. certs need to be cached, otherwise we run into rate limits after requesting a cert for the same subdomain 5x+.

1. Request certs at the same time when we launch the VM and have the cert-request run in the background. The cert should go into the `certs` namespace on core-dev, just like it does for preview envs on core-dev.
2. when the k3s cluster on the VM is ready, copy the certs into that cluster.

## Additional context
If we install certs as described above, should we no longer install cert-manager in preview envs? 
* pro: simplicity
* pro: security
* con: can't test cert-manager in preview envs.

ArthurSens:

## Description

Moves the issuing of certificates to the `prepare` phase (before the containers' build). The certificate can be issued while build is still running and will be cached in core-dev's `certs` namespace.

The cached certificate will be copied to the appropriate preview environment after it is ready to use.

## Known problems (at the moment)
Certificates for both core-dev and harvester are sharing (and overriding) the same Certificate CR in core-dev. When transitioning from core-dev to harvester (or vice-versa) the preview domain changes and the certificate is not valid anymore.

I'll remove the draft after and finish making each preview type to have their own Certificate resource

## Related Issue(s)

Fixes #8998 

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.werft/util/certs.ts
+++ b/.werft/util/certs.ts
@@ -2,17 +2,18 @@ import { exec, ExecOptions } from './shell';
 import { sleep } from './util';
 import * as path from 'path';
 import { CORE_DEV_KUBECONFIG_PATH } from '../jobs/build/const';
+import { Werft } from './werft';
 
 
 export class IssueCertificateParams {
     pathToTemplate: string
     gcpSaPath: string
-    namespace: string
     dnsZoneDomain: string
     domain: string
     ip: string
     additionalSubdomains: string[]
     bucketPrefixTail: string
+    certName: string
     certNamespace: string
 }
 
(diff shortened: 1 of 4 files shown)


ArthurSens:
Currently struggling to issue certificates due to:

.werft/util/certs.ts
```diff
@@ -26,52 +27,72 @@ export class InstallCertificateParams {
...
 export async function issueCertificate(werft, params: IssueCertificateParams, shellOpts: ExecOptions) {
     var subdomains = [];
-    werft.log("certificate", `Subdomains: ${params.additionalSubdomains}`)
+    werft.log(shellOpts.slice, `Subdomains: ${params.additionalSubdomains}`)
     for (const sd of params.additionalSubdomains) {
         subdomains.push(sd);
     }
 
+    try {
+        exec(`echo "Domain: ${params.domain}, Subdomains: ${subdomains}"`, {slice: shellOpts.slice})
```
ArthurSens, line 36:
```suggestion
```

meysholdt:
> Currently struggling to issue certificates due to:

It sounds like the GCP service account used by CertManager does not have access to the domain preview.gitpod-dev.com

ArthurSens:
> > Currently struggling to issue certificates due to:
> 
> It sounds like the GCP service account used by CertManager does not have access to the domain preview.gitpod-dev.com

Thanks Mo, do you know where I can change this configuration? I get kinda lost when navigating through DNS configuration in GCP 😅

wulfthimm:
@ArthurSens The cluster-issuer that manages `preview.gitpod-dev.com` is `letsencrypt-issuer-gitpod-core-dev`. The problem is that the cluster-issuer is bound to a GCP-project and the DNS-Zone for preview.gitpod-dev.com is in the project `gitpod-core-dev`.

ArthurSens:
> @ArthurSens The cluster-issuer that manages `preview.gitpod-dev.com` is `letsencrypt-issuer-gitpod-core-dev`. The problem is that the cluster-issuer is bound to a GCP-project and the DNS-Zone for preview.gitpod-dev.com is in the project `gitpod-core-dev`.

Hmmmm I see... do you see a way to move forward? 🤔

ArthurSens:
Thanks for all the help, just tested and it worked for both core-dev and harvester previews :)

Ready for a review round

wulfthimm (approved):

PR #9081, opened 2022-04-01
[ws-manager] Support custom CA certs

### LINKED ISSUE #9079: Add CA cert support to ws-manager
When using "self-signed certs" the CA cert with which those certs were signed need to be available in workspaces (no matter if regular, prebuild or image-build).

To this end, ws-manager should gain a config field which carries the name of a secret which ought to be mounted into workspaces. See #9078 for more details as to how this should be mounted.

We should refrain from using the workspace templates for this, because that would mean we'd need to merge templates within the installer. Today, the workspace templates are used solely at the discretion of the installing user.

csweichel:
## Description
Adds custom CA cert support to workspaces.

## Related Issue(s)

Fixes #9079

## How to test
1. Ensure that ws-manager's config carries `caCertSecret` entry (done on this branch)
2. Start a workspace on https://cw-fix-9079.staging.gitpod-dev.com/workspaces
3. Cat the custom cert from `cat /etc/ssl/certs/gitpod-ca.crt`

## Release Notes

```release-note
Add custom CA cert support to workspaces
```

### DIFF
--- a/components/workspacekit/cmd/rings.go
+++ b/components/workspacekit/cmd/rings.go
@@ -582,6 +582,7 @@ var (
 		"/sys",
 		"/dev",
 		"/etc/hostname",
+		"/etc/ssl/certs/gitpod-ca.crt",
 	}
 	rejectMountPaths = map[string]struct{}{
 		"/etc/resolv.conf": {},


sagor999 (commented):
@csweichel code looks good, but when I do `cat /etc/ssl/certs/gitpod-ca.crt` it shows me `BEGIN RSA PRIVATE KEY`, which I think is wrong, since `crt` should be public part of the cert, and `gitpod-ca.key` should be the private key. 
So want to make sure we map this correctly.

csweichel:
> @csweichel code looks good, but when I do `cat /etc/ssl/certs/gitpod-ca.crt` it shows me `BEGIN RSA PRIVATE KEY`, which I think is wrong, since `crt` should be public part of the cert, and `gitpod-ca.key` should be the private key. So want to make sure we map this correctly.

That's right - my bad. I just stuffed the wrong file into the secret. We're just mounting the secret directly into the workspace, irregardless of its content. It's up to the user to ensure there's a valid CA cert (not a key) in there.

We could add validation to the installer to ensure it's a CA cert.

sagor999 (approved):

PR #9082, opened 2022-04-01
[installer] Add custom CA secret support

### LINKED ISSUE #9077: Add CA cert secret to the installer
To support "self-signed" certs within Gitpod we need to provide means for users to supply their CA cert.

To this end, we'd introduce a new field in the installer config where users can reference a secret which carries their CA cert chain. We'd add cluster validation which ensures that if this secret is supplied it carries a `ca.crt` entry.

This secret would be consumed by #9078 and #9079

csweichel:
## Description
Adds custom CA cert support to the installer. When specified, the CA certs are mounted into `server`, `image-builder` and `ws-daemon`.

## Related Issue(s)

Fixes #9077

## How to test
(proper test is to install Gitpod on a cluster trying to talk to a Docker registry and SCM which use custom certs).

```bash
# produce new config
installer init > cfg.yaml

# set required values
yq write -i config.yaml domain foobar.com
yq write -i config.yaml customCACert.kind secret
yq write -i config.yaml customCACert.name custom-ca-cert

# validate cluster (should fail)
installer validate cluster --config config.yaml # reports missing secret

# produce secret and validate (should succeed)
openssl genrsa -out ca.crt 2048
kubectl create secret generic custom-ca-certs --from-file=ca.crt=ca.crt
installer validate cluster --config config.yaml

# render
installer render --config config.yaml > objs.yaml

# make sure we got the certs rendered in
grep -B 5 objs.yaml custom-ca-certs
```

## Release Notes

```release-note
Add custom CA cert support to Gitpod services
```

### DIFF
--- a/install/installer/pkg/common/ca.go
+++ b/install/installer/pkg/common/ca.go
@@ -64,3 +64,41 @@ func InternalCAContainer(ctx *RenderContext, mod ...func(*corev1.Container)) *co
 
 	return res
 }
+
+// CustomCACertVolume produces the objects required to mount custom CA certificates
+func CustomCACertVolume(ctx *RenderContext) (vol *corev1.Volume, mnt *corev1.VolumeMount, env *corev1.EnvVar, ok bool) {
+	if ctx.Config.CustomCACert == nil {
+		return nil, nil, nil, false
+	}
+
+	const (
+		volumeName = "custom-ca-cert"
+		mountPath  = "/etc/ssl/certs/custom-ca.crt"
+	)
+	vol = &corev1.Volume{
+		Name: volumeName,
+		VolumeSource: corev1.VolumeSource{
+			Secret: &corev1.SecretVolumeSource{
+				SecretName: ctx.Config.CustomCACert.Name,
+				Items: []corev1.KeyToPath{
+					{
+						Key:  "ca.crt",
+						Path: "ca.crt",
+					},
+				},
+			},
+		},
+	}
+	mnt = &corev1.VolumeMount{
+		Name:      volumeName,
+		ReadOnly:  true,
+		MountPath: mountPath,
+		SubPath:   "ca.crt",
+	}
+	env = &corev1.EnvVar{
+		Name:  "NODE_EXTRA_CA_CERTS",
+		Value: mountPath,
+	}
+	ok = true
+	return
+}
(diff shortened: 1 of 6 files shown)


install/installer/pkg/common/ca.go
```diff
@@ -64,3 +64,41 @@ func InternalCAContainer(ctx *RenderContext, mod ...func(*corev1.Container)) *co
 
 	return res
 }
+
+// CustomCACertVolume produces the objects required to mount custom CA certificates
+func CustomCACertVolume(ctx *RenderContext) (vol *corev1.Volume, mnt *corev1.VolumeMount, env *corev1.EnvVar, ok bool) {
```
princerachit, line 69: nit: Would it make sense to have the return values in a wrapper struct

princerachit:
@csweichel I think something is missing in the test description around `yq`. Are you using the default shipped version of yq with workspace full?

I believe `cfg.yaml` and `config.yaml` are the same files. And is the `write` keyword in `yq write -i config.yaml domain foobar.com` command supposed to be there? It is failing for me.

csweichel:
> @csweichel I think something is missing in the test description around `yq`. Are you using the default shipped version of yq with workspace full?

Yes. I just used `yq` as it was available in my workspace on this branch.

> 
> I believe `cfg.yaml` and `config.yaml` are the same files. 

Yep, fixed it in the description.

> And is the `write` keyword in `yq write -i config.yaml domain foobar.com` command supposed to be there? It is failing for me.

AFAIK there are two incompatible versions of `yq` going 'round. I just used the `yq` as I found it in the workspace on this branch.

corneliusludmann (approved):
looks good :+1:  :skateboard:

csweichel:
@easyCZ that's a UX bug in the installer. The installer requires users to pass the namespace explicitly, i.e. `./installer  validate cluster -n staging-cw-fix-9077 config.yaml` works.

@corneliusludmann @lucasvaltl do we have an issue for this already?

iamulya:
> (proper test is to install Gitpod on a cluster trying to talk to a Docker registry and SCM which use custom certs).
@csweichel Doesn't the workspace pod configuration have to be changed/updated as well so that git inside workspace can talk to an SCM with custom CA?

princerachit (approved):

easyCZ (approved):

PR #9083, opened 2022-04-01
[kots]: add a kernel version check to the preflights

### LINKED ISSUE #8862: Kernel version
Ensure that the Kernel version is at least 5.4.0. Dependent upon whether the check is semver, this may need to support `5.4.0-0` so that examples such as `5.4.0-gke` are valid

See [check](https://github.com/gitpod-io/gitpod/blob/main/install/installer/pkg/cluster/checks.go#L23) in the Installer for details

MrSimonEmms:
## Description

Add a preflight check to ensure that the kernel version is 5.4.0 or above. It includes preflight checks, so that `5.4.0-gke` would return as valid. 

## Related Issue(s)

Fixes #8862

## How to test

Deploy via KOTS.

To check the functionality, run some of the followings:

```shell
docker run --rm alpine/semver semver --coerce --range '>=5.4.0' 5.13.0-1019-gcp // valid
docker run --rm alpine/semver semver --coerce --range '>=5.4.0' 5.4.0-1009-gcp // valid
docker run --rm alpine/semver semver --coerce --range '>=5.4.0' 5.3.9-1009-gcp // errors - would return "invalid"
```

## Release Notes

```release-note
[kots]: add a kernel version check to the preflights
```

### DIFF
--- a/install/kots/manifests/kots-preflight.yaml
+++ b/install/kots/manifests/kots-preflight.yaml
@@ -6,6 +6,16 @@ kind: Preflight
 metadata:
   name: gitpod
 spec:
+  collectors:
+    - run:
+        collectorName: "kernel"
+        image: alpine/semver
+        name: kernel
+        command:
+          - /bin/sh
+          - -c
+        args:
+          - semver --coerce --range '>=5.4.0' $(uname -r) || echo invalid
   analyzers:
     - clusterVersion:
         outcomes:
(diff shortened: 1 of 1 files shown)


MrSimonEmms:
As a follow-up, [have checked](https://gitpod.slack.com/archives/C02SWJ49JQ2/p1648831983190249) with Barry @ Replicated whether this image will automagically get added to the list of airgapped images or whether we need to add it to `additionalImages` list in `kots-app.yaml`

corneliusludmann (approved):
looks good to me

Issue #8740, opened 2022-03-11
[installer] Disable definitely-gp by default


corneliusludmann:
The installer has a config value `definitelyGpDisabled` that is currently `false` by default. Since we [deprecate `definitely-gp` soon](https://github.com/gitpod-io/gitpod/issues/8473), we should set it to `true` by default.

Things to consider:
- [ ] Change the[ default config value in the installer](https://github.com/gitpod-io/gitpod/blob/b10bdc67d96ed0f8c559cc654b969c3bd68c9631/install/installer/pkg/config/v1/config.go#L57)
- [ ] Update [`air-gap.md`](https://github.com/gitpod-io/gitpod/blob/b10bdc67d96ed0f8c559cc654b969c3bd68c9631/install/installer/docs/air-gap.md?plain=1#L29)
- [ ] Check if we need to set the value to `false` for preview envs and SaaS installations

PR #9094, opened 2022-04-04
[installer] Disable `definitely-gp` by default

Pothulapati:
## Description

`definitely-gp` is being deprecated soon, and hence
it is important to be disabled by default. Once it
is fully deprecated, even this option has to be removed.

For now, The following changes have been made:

- Update the `definitelyGpDisabled` option to be true, Thus
  disabling it by default
- Update the airgap docs to remove a mention of this

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/8740

## Release Notes

```release-note
[installer] Disable `definitely-gp` by default
```

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -300,7 +300,7 @@ components:
       minAgePrebuildDays: 7
       contentRetentionPeriodDays: 21
       contentChunkLimit: 1000
-    definitelyGpDisabled: "false"
+    definitelyGpDisabled: true
     enableLocalApp: false
     disableDynamicAuthProviderLogin: false
     maxEnvvarPerUserCount: 4048
--- a/install/installer/docs/air-gap.md
+++ b/install/installer/docs/air-gap.md
@@ -22,11 +22,10 @@ done
 
 ## Install Gitpod in Air-Gap Mode
 
-To install Gitpod in an air-gap network, you need to configure the repository of the images needed by Gitpod (see previous step) and disable the `definitely-gp` feature. Add this to your Gitpod config:
+To install Gitpod in an air-gap network, you need to configure the repository of the images needed by Gitpod (see previous step). Add this to your Gitpod config:
 
 ```yaml
 repository: your-registry.example.com
-disableDefinitelyGp: true
 ```
 
 That's it. Run the following commands as usual and Gitpod fetches the images from your registry and does not need internet access to operate:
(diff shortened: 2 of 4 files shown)


MrSimonEmms (approved):

Pothulapati:
Was able to verify that after this change, We don't run `.gitpod.yml`'s from the [definitely-gp]() repository, and instead run the commands from recognition. [Open Vue Workspace on preview branch](https://tar-disabl6670848dbe.staging.gitpod-dev.com/#https://github.com/vuejs/vue) runs `yarn install && yarn run build` instead of what's in [definitely-gp](https://github.com/gitpod-io/definitely-gp/blob/master/vue/.gitpod.yml) like on [main's environment](https://main.staging.gitpod-dev.com/#https://github.com/vuejs/vue)

With this change, preview environments will not be using `definitely-gp` configs anymore.

geropl:
@Pothulapati We should also adjust:
 - the default in the helm chart: https://github.com/gitpod-io/gitpod/blob/main/chart/values.yaml#L303
 - override that with "false" for:
   - [staging](https://github.com/gitpod-io/ops/blob/main/charts/gitpod_io/values.staging.yaml#L175)
   - and [production](https://github.com/gitpod-io/ops/blob/main/charts/gitpod_io/values.production.yaml#L175)

geropl (approved):

chart/values.yaml
```diff
@@ -300,7 +300,7 @@ components:
       minAgePrebuildDays: 7
       contentRetentionPeriodDays: 21
       contentChunkLimit: 1000
-    definitelyGpDisabled: "false"
+    definitelyGpDisabled: true
```
MrSimonEmms, line 303: Is this ok changing the type from a string to a boolean?

MrSimonEmms (approved):
/hold

Asked a question about changing the Helm chart type. If it's fine, happy for you to unhold

chart/values.yaml
Pothulapati: Yep, Confirmed that we aren't using it anywhere to toggle something, but [just passing it down into the `server-configmap`, and follow the same pattern there like other boolean values ](https://github.com/gitpod-io/gitpod/blob/54a765d321236ef42dceeb154cf477b17a8aaf32/chart/templates/server-configmap.yaml#L38)(soemthing like `enableLocalApp`)

PR #9096, opened 2022-04-04
WIP: Base server for go

easyCZ:
## Description

Introduces a library called `baseserver`. It is a combination of an HTTP and gRPC server designed to make it dead-easy to get sensible defaults and be able to use a server.

Using a common base server can give us the following benefits:
* Reduced learning curve for each server engineers interact with
* Drive good sensible defaults
* Provide security out of the box
* Provide authentication & authorization out of the box
* Metrics, logging, profiling all out of the box
* Engineers can focus on building services, not servers
* No reason to not write good client-to-server tests as unit tests when standing up a server is so easy

To fulfill the role outlined above, the following will be needed as well. These will be added incrementally as we need them.
- [ ] Prometheus metrics server
- [ ] gRPC Interceptors
- [ ] HTTP Interceptors
- [ ] Use k8s health handler
- [ ] pprof

## Related Issue(s)

Fixes #

## How to test

* go test ./...

## Release Notes

```release-note
NONE
```

## Documentation

/uncc
/hold

### DIFF
--- a/components/common-go/baseserver/http.go
+++ b/components/common-go/baseserver/http.go
@@ -1,0 +1,11 @@
+package baseserver
+
+import "net/http"
+
+var healthyResponse = []byte(`healthy`)
+
+func ReadyHandler() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		_, _ = w.Write(healthyResponse)
+	})
+}
--- a/components/common-go/baseserver/http.go
+++ b/components/common-go/baseserver/http.go
@@ -1,3 +1,7 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
 package baseserver
 
 import "net/http"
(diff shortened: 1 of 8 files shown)


roboquat:

<details>

</details>

components/common-go/baseserver/server.go
```diff
@@ -0,0 +1,213 @@
...
+	"context"
+	"fmt"
+	"github.com/gitpod-io/gitpod/common-go/log"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+	"net"
+	"net/http"
+)
+
+type config struct {
```
csweichel, line 13: We should export this so that we can reuse this config in the installer and component config.

We'd want to pull in pprof and prometheus, too.

components/common-go/baseserver/http.go
```diff
@@ -0,0 +1,11 @@
+package baseserver
+
+import "net/http"
+
+var healthyResponse = []byte(`healthy`)
+
+func ReadyHandler() http.Handler {
```
csweichel, line 7: registry-facade and others use a Kubernetes package for their readiness probes. We should re-use that and promote their use rather than rolling our own.

easyCZ: Yep, makes sense. The idea here is that when you get a server, it already comes with sensible defaults. What does are is entirely up to us.

easyCZ:
Landed incrementally as part of https://github.com/gitpod-io/gitpod/issues/9229, closing.

PR #9106, opened 2022-04-04
[registry-facade] Override IPFSAddr value using the IPFS_HOST env var…

aledbf:
…iable

## Related Issue(s)

Fixes #

## Release Notes

```release-note
```


csweichel:
The sentinel change makes 💯 sense.

I'm not sure I follow the need to override the config field value though. This strikes me as a less flexible way compared to what we have now. Also it's a departure from current behaviour where AFAIK we have no other place where env vars would take precedence over config values.

aledbf:
> I'm not sure I follow the need to override the config field value though. 

Ok. How do you propose to set the value?
This is what I have in the installer yaml file

```
    registryFacade:
      ipfsCache:
        enabled: true
        redis:
          masterName: kumquat
          sentinelAddrs:
            - redis.storage.svc.cluster.local:26379
          username: default
          passwordSecret: redis-credentials
```

csweichel:
> Ok. How do you propose to set the value? (I want to use the local node IPFS instance)
 
 ```YAML
     registryFacade:
       ipfsCache:
         enabled: true
         ipfs: /ip4/$IPFS_HOST/tcp/5001
         redis:
           masterName: kumquat
           sentinelAddrs:
             - redis.storage.svc.cluster.local:26379
           username: default
           passwordSecret: redis-credentials
 ```

csweichel (approved):
LGTM

PR #9113, opened 2022-04-04
Properly update response in openvsx proxy

jeanp413:
## Description

Updates response when it's compressed

## Related Issue(s)

Fixes #

## How to test

1. Open a workspace
2. Run `NODE_DEBUG=http,https,tls /ide/node /ide/out/server-main.js --install-extension znck.vue --do-not-sync --log trace 2> debug.log` in terminal
3. Check in `debug.log` file that only requests to the openvsx proxy are made and extensions are installed sucessfully
4. Install extensions manually using UI, check extensions are installed sucessfully

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/openvsx-proxy/pkg/modifyresponse.go
+++ b/components/openvsx-proxy/pkg/modifyresponse.go
@@ -6,6 +6,7 @@ package pkg
 
 import (
 	"bytes"
+	"compress/gzip"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -48,19 +49,19 @@ func (o *OpenVSXProxy) ModifyResponse(r *http.Response) error {
 		return nil
 	}
 
-	body, err := ioutil.ReadAll(r.Body)
+	rawBody, err := ioutil.ReadAll(r.Body)
 	if err != nil {
-		log.WithFields(logFields).WithError(err).Error("error reading response body")
+		log.WithFields(logFields).WithError(err).Error("error reading response raw body")
 		return err
 	}
 	r.Body.Close()
-	r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+	r.Body = ioutil.NopCloser(bytes.NewBuffer(rawBody))
 
 	if r.StatusCode >= 500 || r.StatusCode == http.StatusTooManyRequests || r.StatusCode == http.StatusRequestTimeout {
 		// use cache if exists
 		bodyLogField := "(binary)"
-		if utf8.Valid(body) {
-			bodyStr := string(body)
+		if utf8.Valid(rawBody) {
+			bodyStr := string(rawBody)
 			truncatedSuffix := ""
 			if len(bodyStr) > 500 {
 				truncatedSuffix = "... [truncated]"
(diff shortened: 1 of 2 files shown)


akosyakov:
@jeanp413 `openvsxorg.blob.core.windows.net` is reliable and it is fine to connect to it directly to fetch vsix files, only requests to open-vsx.org are problematic and should always go through OpenVSX proxy. We don't watch to cache vsix files, but only requests to open-vsx.org necessary to resolve them.

akosyakov:
> @jeanp413 openvsxorg.blob.core.windows.net is reliable and it is fine to connect to it directly to fetch vsix files, only requests to open-vsx.org are problematic and should always go through OpenVSX proxy. We don't watch to cache vsix files, but only requests to open-vsx.org necessary to resolve them.

Sorry for the noise, after testing it does behave as I described. I was confused by mentioning of openvsxorg.blob.core.windows.net.

akosyakov:
@iQQBot @mustard-mh the changes seem to do the job, it would be good if you could review go code

components/openvsx-proxy/pkg/modifyresponse.go
```diff
@@ -48,19 +50,53 @@ func (o *OpenVSXProxy) ModifyResponse(r *http.Response) error {
...
 	if err != nil {
-		log.WithFields(logFields).WithError(err).Error("error reading response body")
+		log.WithFields(logFields).WithError(err).Error("error reading response raw body")
 		return err
 	}
 	r.Body.Close()
-	r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+	r.Body = ioutil.NopCloser(bytes.NewBuffer(rawBody))
+
+	if r.StatusCode >= 300 && r.StatusCode < 400 && o.Config.FollowRedirects > 0 && r.Header.Get("Location") != "" {
```
akosyakov, line 61: @jeanp413 are these changes necessary? The problem was that open-vsx.org has a response with URLs to open-vsx.org and since compressions is enabled we did not replace them?

Why do we need to tackle redirects?

components/openvsx-proxy/pkg/modifyresponse.go
```diff
@@ -93,8 +129,26 @@ func (o *OpenVSXProxy) ModifyResponse(r *http.Response) error {
...
+			}
+
+			body, err = ioutil.ReadAll(gzipReader)
+			if err != nil {
+				log.WithFields(logFields).WithError(err).Error("error reading response body")
+				return nil
+			}
+			gzipReader.Close()
+
+			r.Header.Del("Content-Encoding")
```
akosyakov, line 149: Should not we compress the final result then to reduce load on clients?

jeanp413: I wans't sure about it, but now that you mention it I think it makes sense

components/openvsx-proxy/pkg/modifyresponse.go
akosyakov: All workspaces are configured to use  `http_proxy` and `no_proxy` in their setup. It is fine. We don't fix this issue for air-gapped but for general case:
- open-vsx.org is not reliable, request to it should go through OpenVSX proxy
- requests to openvsxorg.blob.core.windows.net (and also big), so we don't cache them and go directly

The cache is only 200Mb, we don't want to populate it with something like 50Mb vsix for java extension, but fast responses for common questions to open-vsx.org.

jeanp413: Thanks for the explanation, I'll remove the redirect logic then

components/openvsx-proxy/pkg/modifyresponse.go
```diff
@@ -93,13 +94,42 @@ func (o *OpenVSXProxy) ModifyResponse(r *http.Response) error {
 	}
 
 	// no error (status code < 500)
+	body := rawBody
 	contentType := r.Header.Get("Content-Type")
 	if strings.HasPrefix(contentType, "application/json") {
+		compressedResponse := strings.EqualFold(r.Header.Get("Content-Encoding"), "gzip")
```
andreafalzetti, line 100: for me, it'd be easier to understand that this is a bool, if we name this: `isCompressedResponse`

iQQBot (approved):
go code looks good

/hold

Only in the original code there are a lot of conversions between []byte and string, some of which are not necessary, bytes have almost the same methods as strings, as well as Count, Replace, etc., but these are not included in this PR

feel free to unblock

akosyakov (approved):
It also works well for me.

Issue #9107, opened 2022-04-04
Use WorkspaceStatus.status_version to ignore old status updates


easyCZ:
## Is your feature request related to a problem? Please describe

When the `bridge` handles `WorkspaceStatus` updates, it doesn't ignore events which have an older `status_version`. The `status_version` can allow us to apply a partial order on the WorkspaceStatus events.

To do this, we need to do the following:
* Store the status_version we've processed in the WorkspaceInstance (and Prebuild) records
* When a new update arrives, check if this is an older update by comparing with the stored version.

Doing this allows us to ensure we do not rely on event ordering inside the message bus and we prevent inconsistent state compared to reality.

easyCZ:
This has now landed with #9116.

Issue #8596, opened 2022-03-04
[bridge] Ensure we do not override "failed" instance states (affects prebuilds!)


geropl:
Currently we're seeing that `failed` prebuild instances' status is override by newer instance updates lagging that state ([ref](https://github.com/gitpod-io/gitpod/issues/8096#issuecomment-1058175304)).

While [this is a bug](https://github.com/gitpod-io/gitpod/issues/8096#issuecomment-1059046204), we should be failsafe here, and basically:
 - never overwrite the `failed` condition
 - add additional logging for unexpected updates
 - accept and forward `ws-manager` `version` field

geropl:
@easyCZ This was the issue I tried to refer to: it's relevant for prebuild reliability, but also for other workspaces

The `version` field won't help with this concrete problem (because ws-manager happily gives out a higher version for earlier, already gone-thorough phases). But it's nice to forward anyway, and should help with reconnect issues in ws-manager-bridge (and potentially, all downstream channels).

geropl:
> Are you referring to the status_version field or is this some other version field?

No, that's the field I meant! :100:

easyCZ:
The prebuild part of this issue has now landed in #9116. I am going to first validate the effects of this change on Prebuilds before making the corresponding change for Workspace Instance updates.

Leaving this open as we still need to make the same change for WorkspaceInstanceStatus updates.

easyCZ:
Using prior work to detect stale events, we can inspect if stale events are a problem for WorkspaceInstance updates also. In practice, we haven't observed stale event updates for prebuilds. The code-flow is as follows:

1. Receive a workspace instance update
2. Process it, and re-map to a db-record so that we can communicate the WorkspaceInstance status
3. Invoke a controller loop for prebuilds. If the update is not for a Prebuild workspace instance, we ignore it.

This means that prebuild updates are a reasonable (but complete) baseline for "stale updates".

<img width="1609" alt="Screenshot 2022-04-20 at 10 32 33" src="https://user-images.githubusercontent.com/1419286/164186289-d7df1081-52f3-4580-aa62-376f66c2b3b3.png">

geropl:
We'# fix this in two steps:
 1. roll out https://github.com/gitpod-io/gitpod/pull/10900 and observe logs for ~1 week to see if how often we're hitting that situation
 2. based on the data we gathered, decide how to proceed with this: Ideally by ignoring destructive updates to the "failed" condition :+1:

[...]

PR #9114, opened 2022-04-04
[db] Add statusVersion to prebuilds to track status version

easyCZ:
## Related Issue(s)

* https://github.com/gitpod-io/gitpod/issues/8596
* https://github.com/gitpod-io/gitpod/issues/9107

## How to test

1. Preview works
2. New prebuilds can start
3. Column statusVersion appears in the DB

## Release Notes

```release-note
NONE
```

## Documentation

/hold
/uncc

### DIFF
--- a/components/gitpod-protocol/src/protocol.ts
+++ b/components/gitpod-protocol/src/protocol.ts
@@ -634,6 +634,7 @@ export interface PrebuiltWorkspace {
     buildWorkspaceId: string;
     creationTime: string;
     state: PrebuiltWorkspaceState;
+    stateVersion: number;
     error?: string;
     snapshot?: string;
 }
--- a/components/gitpod-protocol/src/protocol.ts
+++ b/components/gitpod-protocol/src/protocol.ts
@@ -634,7 +634,7 @@ export interface PrebuiltWorkspace {
     buildWorkspaceId: string;
     creationTime: string;
     state: PrebuiltWorkspaceState;
-    stateVersion: number;
+    statusVersion: number;
     error?: string;
     snapshot?: string;
 }
--- a/components/gitpod-db/src/typeorm/entity/db-prebuilt-workspace.ts
+++ b/components/gitpod-db/src/typeorm/entity/db-prebuilt-workspace.ts
@@ -69,10 +69,10 @@ export class DBPrebuiltWorkspace implements PrebuiltWorkspace {
     })
     error?: string;
 
-    // stateVersion defines the last observed stateVersion from a WorkspaceStatus. See ws-manager-api/core.proto.
-    // stateVersion must only be set by controller/observer.
+    // statusVersion defines the last observed stateVersion from a WorkspaceStatus. See ws-manager-api/core.proto.
+    // statusVersion must only be set by controller/observer.
     @Column({
         default: 0,
     })
-    stateVersion: number;
+    statusVersion: number;
 }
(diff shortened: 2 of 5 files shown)


roboquat:

<details>

</details>

Sunset17 (dismissed):

components/gitpod-protocol/src/protocol.ts
```diff
@@ -634,6 +634,7 @@ export interface PrebuiltWorkspace {
     buildWorkspaceId: string;
     creationTime: string;
     state: PrebuiltWorkspaceState;
+    statusVersion: number;
```
easyCZ, line 637: Not entirely happy that the statusVersion needs to be exposed in the protocol and sent to clients. Is there a way (without significantly changing the system) to have this data only available to internal components and continue not having it in clients?

geropl: > continue not having it in clients?

You could :
 1. introduce a sub-shape of that we use in protocol, and a more expanded shape we expose in `GitpodService`
 2. and hand-remove this field when sending it to clients (we do this with some sensitive field of `WorkspaceInstance` in `GitpodServerImpl`)
 3. but I'm not sure it warrant's the effort, now and for this single field. Let's hope make it better with the new API :wink:

geropl:
@easyCZ I wonder if we could place the field into `WorkspaceInstance` generically, because there we'll have to have it as well. And then we have duplication. :thinking:

easyCZ:
> @easyCZ I wonder if we could place the field into WorkspaceInstance generically, because there we'll have to have it as well. And then we have duplication. 🤔

I'd rather avoid generics for things like the DB schema and core protocol. It introduces coupling which can create unintended side-effects (especially on the DB level) when you try to break it up.

I'm happy to compose it into both, however, if you think it's needed.

components/gitpod-protocol/src/protocol.ts
easyCZ: Thanks, that's what I was thinking as well. Unless there is a strong objection, I'll leave it as is for now.

geropl:
> I'd rather avoid inheritance for things like the DB schema and core protocol.

Hm, I'd rather stick to the current idea of "data deduplication" (inheritance :ghost: ) even if maintain the tight coupling between both tables. If we want to split it out we have to look at it more holistically I fear.

easyCZ:
From what I can see, the ORM definition of Prebuilds does not currently have any inheritance specified [[file](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-db/src/typeorm/entity/db-prebuilt-workspace.ts#L21)].

And neither does the protocol [[file]](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-protocol/src/protocol.ts#L628). Where exactly do you want to add inheritance?

I'd be open to defining an interface which has a field `statusVersion` which we can use to type the objects but YAGNI?

geropl:
**Summary**: We're going with the statusVersion field in the prebuilds tabe
**Reasoning**:
we want to have statusVersion the API-facing shape
we agree that WorkspaceInstance will use the same field at some point
it's unpractical to read statusVersion from prebuiold (2-join Workspace->WorkspaceInstance.statusVersion)
we're taking the slight risk of data-duplication/-races in favor of
following the current pattern for API shapes (same as internal)
following the current pattern by having all fields in the shape in the DB table
Also, statusVersion tells us something about the data quality of all fields in the table (data provenance), so might make sense to duplicate it in the first place :upside_down_face:

geropl (approved):
LGTM

easyCZ:
Thanks for the review and discussion @geropl!

/unhold

PR #9118, opened 2022-04-04
add PVC feature flag

### LINKED ISSUE #9117: add a feature flag (set at user level) into workspace creation to support PVC\Snapshot pathway
Add a dedicated feature flag "PVC" to enable code path to start implementing PVC\Snapshot DD

sagor999:
## Description

Add PVC feature flag to be able to start working on PVC\Snapshot code path.

## Related Issue(s)

Fixes #9117

## Release Notes

```release-note
NONE
```


sagor999:
@geropl do I need to do anything extra to make sure that new feature flag will show up in Admin panel?

components/gitpod-protocol/src/protocol.ts
```diff
@@ -151,7 +151,7 @@ export interface UserFeatureSettings {
  * The values of this type MUST MATCH enum values in WorkspaceFeatureFlag from ws-manager/client/core_pb.d.ts
  * If they don't we'll break things during workspace startup.
  */
-export const WorkspaceFeatureFlags = { full_workspace_backup: undefined, fixed_resources: undefined };
+export const WorkspaceFeatureFlags = { full_workspace_backup: undefined, fixed_resources: undefined, pvc: undefined };
```
geropl, line 154: Personally I'm a big fan of more descriptive names, especially if it's in an "interface". E.g., it will be impossible in the admin UI to tell what exactly "pvc" stands for without prior knowledge/lookup in the code.

Does it mean "persistent_volume_claim"? Can we use that as feature flag? :thinking:

sagor999: @geropl Sounds good. Even that name is not quite right, but I do agree it is better than just `PVC`.

easyCZ (approved):
LGTM. Worth pointing out different deployment cadence between WS Clusters and WebApp clusters. It won't break due to proto compatibility protocol but it may not fully work until both are deployed.

Adding hold incase others want to review also.
/hold

components/gitpod-protocol/src/protocol.ts
easyCZ: Personally, I'd even go further and name `workspaces_use_persistent_volume_claim_for_storage` as it's real fun deprecating a feature flag two years later after it was forgotten through team reshuffles etc :)

geropl (approved):
👍

aledbf (approved):
LGTM

PR #9121, opened 2022-04-05
[gitpod-db] Don't consider garbage-collected prebuilds as potential bases for incremental prebuilds

jankeromnes:
## Description

Don't consider garbage-collected prebuilds as potential bases for incremental prebuilds.

## Related Issue(s)

Fixes high cardinality on frequently-run SQL queries.

For example, for the [Gitpod](https://github.com/gitpod-io/gitpod) repository,
- The current query returns 16K+ entries
- The fixed query returns 335 entries

For the official [GitLab](https://gitlab.com/gitlab-org/gitlab) repository,
- The current query returns 81K+ entries
- The fixed query returns 6 entries (😳)

## How to test

1. Run the query in a copy of the production database (e.g. eu-failover) for a few highly-active repositories

## Release Notes

```release-note
[gitpod-db] Don't consider garbage-collected prebuilds as potential bases for incremental prebuilds
```

### DIFF
--- a/components/gitpod-db/src/typeorm/workspace-db-impl.ts
+++ b/components/gitpod-db/src/typeorm/workspace-db-impl.ts
@@ -754,6 +754,7 @@ export abstract class AbstractTypeORMWorkspaceDBImpl implements WorkspaceDB {
         query = query.orderBy("pws.creationTime", "DESC");
         query = query.innerJoinAndMapOne("pws.workspace", DBWorkspace, "ws", "pws.buildWorkspaceId = ws.id");
         query = query.andWhere("ws.deleted = false");
+        query = query.andWhere("ws.contentDeletedTime = ''");
 
         const res = await query.getMany();
         return res.map((r) => {
--- a/components/server/ee/src/workspace/workspace-factory.ts
+++ b/components/server/ee/src/workspace/workspace-factory.ts
@@ -106,6 +106,8 @@ export class WorkspaceFactoryEE extends WorkspaceFactory {
             // Walk back the last prebuilds and check if they are valid ancestor.
             let ws;
             if (context.commitHistory && context.commitHistory.length > 0) {
+                // Note: This query returns only not-garbage-collected prebuilds in order to reduce cardinality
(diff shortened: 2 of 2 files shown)


components/gitpod-db/src/typeorm/workspace-db-impl.ts
```diff
@@ -754,6 +754,7 @@ export abstract class AbstractTypeORMWorkspaceDBImpl implements WorkspaceDB {
         query = query.orderBy("pws.creationTime", "DESC");
         query = query.innerJoinAndMapOne("pws.workspace", DBWorkspace, "ws", "pws.buildWorkspaceId = ws.id");
         query = query.andWhere("ws.deleted = false");
+        query = query.andWhere("ws.contentDeletedTime = ''");
```
geropl, line 757: Will try this in prod (offline) to get a feeling for cardinality...

jankeromnes: Currently trying this in prod-failover with the official GitLab repo (with and without the fix). Will add findings to the PR description.

geropl: Reduced cardinality from 16919 to 337. :muscle:

geropl:
@jankeromnes Thx for tackling this! One small nit: Could you add a comment to [this line](https://github.com/gitpod-io/gitpod/blob/8ad398ecd48ec66ff3b1c28d3b669e48e8845b23/components/server/ee/src/workspace/workspace-factory.ts#L109) about the reasoning behind cardinality? That would be golden!

jankeromnes:
Thanks!

> One small nit: Could you add a comment to [this line](https://github.com/gitpod-io/gitpod/blob/8ad398ecd48ec66ff3b1c28d3b669e48e8845b23/components/server/ee/src/workspace/workspace-factory.ts#L109) about the reasoning behind cardinality? That would be golden!

Sure! [Done](https://github.com/gitpod-io/gitpod/compare/87f8858866ee23803c50da52f6b6ffef881efed5..d7939b2627a57ea4d38ac4afc2b6609e8b7e27c5)

geropl (approved):
Thx!

Issue #8445, opened 2022-02-24
.gitpod.yml in project gets overwritten by gitpod on launch


rfay:
### Bug description

If I launch a project, the existing .gitpod.yml is overwritten with a dummy .gitpod.yml.

This is a serious problem - The existing .gitpod.yml should be respected. This completely breaks everything that has a .gitpod.yml. Is there any workaround? This completely breaks https://github.com/drud/ddev-gitpod-launcher and really... everybody that ever had a .gitpod.yaml. I don't need this silly replacement .gitpod.yml:

```yaml
tasks:
  - init: |
      echo 'TODO: build project'
    command: |
      echo 'TODO: start app'
```

This is  a super-critical bug, gitpod should never be replacing existing files.

### Steps to reproduce

Just launch https://drud.github.io/ddev-gitpod-launcher/ with the defaults. The code is in https://github.com/drud/ddev-gitpod-launcher

### Expected behavior

.gitpod.yml should not be overwritten.

### Example repository

https://github.com/drud/ddev-gitpod-launcher

rfay:
I apologize for making this work now. I went into settings and forced a prebuild, and now it does the right thing. 

I assume that this problem was caused by moving ddev-gitpod-launcher into a new team, which may have caused an invalid prebuild? So maybe that's how to recreate it... Unfortunately, now the problem won't happen with ddev-gitpod-launcher.

Interestingly enough, the problem seemed not to happen with the other project added to the team, https://github.com/drud/ddev

geropl:
I can verify that it works now as expected.

> I went into settings and forced a prebuild, and now it does the right thing. 
> I assume that this problem was caused by moving ddev-gitpod-launcher into a new team, which may have caused an invalid prebuild? So maybe that's how to recreate it... Unfortunately, now the problem won't happen with ddev-gitpod-launcher.

@jankeromnes Can you investigate if there might be a gap that we have to close when trying to answer the "does this repo have a config" question? :thinking:

jankeromnes:
Uh oh, many thanks @rfay for reporting this quite scary-sounding bug!

---

Let's have a look in the code to see what the intended logic is, and if there can be any edge cases:

https://github.com/gitpod-io/gitpod/blob/2f0b9462b0451f0974434fa36bd757420ef04e34/components/server/src/workspace/workspace-factory.ts#L179-L182

[...]

PR #9126, opened 2022-04-05
[server] Throw an error when trying to fetch a file from GitHub using an unknown revision

jankeromnes:
## Description

When we fetch the file content of a GitHub repository's `.gitpod.yml` file, but we don't know which `revision` (i.e. commit hash) to fetch, we currently act like there is no `.gitpod.yml` in the repo and continue without throwing any error:

https://github.com/gitpod-io/gitpod/blob/2f0b9462b0451f0974434fa36bd757420ef04e34/components/server/src/github/file-provider.ts#L53-L56

I believe this might potentially lead to us accidentally consider there is no `.gitpod.yml` in a repo, and overwrite it with something else when starting a workspace, even if there was actually a `.gitpod.yml`: https://github.com/gitpod-io/gitpod/issues/8445

Thus, I believe it might be better to throw an error when trying to fetch a GitHub file for an unknown revision.

## Related Issue(s)

Proof-of-concept related to https://github.com/gitpod-io/gitpod/issues/8445

## Release Notes

```release-note
[server] Throw an error when trying to fetch a file from GitHub using an unknown revision
```

## Anything else?

I believe `commit.revision` might accidentally be set to `undefined` or `""` (or `null` / other "falsy" value) in one of these places in the GitHub context parser:

https://github.com/gitpod-io/gitpod/blob/f6feb961ed1f4b5987ca0af71bc7929b33e73d26/components/server/src/github/github-context-parser.ts#L156

https://github.com/gitpod-io/gitpod/blob/f6feb961ed1f4b5987ca0af71bc7929b33e73d26/components/server/src/github/github-context-parser.ts#L235

https://github.com/gitpod-io/gitpod/blob/f6feb961ed1f4b5987ca0af71bc7929b33e73d26/components/server/src/github/github-context-parser.ts#L240

https://github.com/gitpod-io/gitpod/blob/f6feb961ed1f4b5987ca0af71bc7929b33e73d26/components/server/src/github/github-context-parser.ts#L421

https://github.com/gitpod-io/gitpod/blob/f6feb961ed1f4b5987ca0af71bc7929b33e73d26/components/server/src/github/github-context-parser.ts#L505

(For example, if there are edge cases we didn't foresee, or if the GitHub API misbehaves, or if they slightly change their response format...)

### DIFF
--- a/components/server/src/github/file-provider.ts
+++ b/components/server/src/github/file-provider.ts
@@ -52,7 +52,7 @@ export class GithubFileProvider implements FileProvider {
 
     public async getFileContent(commit: Commit, user: User, path: string) {
         if (!commit.revision) {
-            return undefined;
+            throw new Error(`Cannot fetch file content of ${path} (unknown revision: ${commit.revision})`);
         }
 
         try {


jankeromnes:
This might not be the correct fix, so closing for now. [More context on Slack](https://gitpod.slack.com/archives/C02EN94AEPL/p1649155558332379) (internal)

PR #9129, opened 2022-04-05
[ws-manager-bridge] Fix counting of stale prebuild events

easyCZ:
## Description

We'd count records with statusVersion = 0 stale, but this is the default value set in the DB. We want to exclude these from metrics to get an accurate count of what is actually stale.

## How to test

1. Start prebuild
2. No metric initially reported

## Release Notes

```release-note
NONE
```


AlexTugarev (approved):
Change looks good to me.

PR #9133, opened 2022-04-05
werft: Replace terraform with gcloud nodejs lib

ArthurSens:
## Description

This PR should replace all terraform scripts used to create DNS with the official google cloud nodejs library.

## Related Issue(s)

Fixes https://github.com/gitpod-io/ops/issues/1441

## Release Notes

```release-note
NONE
```


.werft/util/gcloud.ts
```diff
@@ -33,3 +35,28 @@ export async function deleteExternalIp(phase: string, name: string, region = "eu
 function getExternalIp(name: string, region = "europe-west1") {
     return exec(`gcloud compute addresses describe ${name} --region ${region}| grep 'address:' | cut -c 10-`, { silent: true }).trim();
 }
+
+export function createDNSRecord(domain: string, dnsZone: string, IP: string, slice: string) {
```
mads-hartmann, line 39: I'd change this to be an `async` function and then explicitly set the return type. So it would be the following:

```
export async function createDNSRecord(domain: string, dnsZone: string, IP: string, slice: string): Promise<void> {
```

.werft/util/gcloud.ts
```diff
@@ -33,3 +35,28 @@ export async function deleteExternalIp(phase: string, name: string, region = "eu
...
+    })
+    const zone = dnsClient.zone(dnsZone)
+    const record = new Record(zone, 'a', {
+        name: domain,
+        ttl: 300,
+        signatureRrdatas: [IP],
+        data: IP
+    })
+
+    zone.addRecords(record).then(rsp => {
```
mads-hartmann, line 55: Now that this is an async function, you can await this promise and then perform the log. So it becomes

```
await zone.addRecords(record)
werft.log(slice, `DNS Record created`)
```

As you're just re-throwing the error you can just delete those. When you use `await` on a promise it will automatically throw the error if the promise is rejected (fails) ☺️

.werft/jobs/build/deploy-to-preview-environment.ts
```diff
@@ -659,22 +661,12 @@ async function addDNSRecord(werft: Werft, namespace: string, domain: string, isL
...
-    && terraform init -backend-config='prefix=${namespace}' -migrate-state -upgrade \
-    && terraform apply -auto-approve \
-        -var 'dns_zone_domain=gitpod-dev.com' \
-        -var 'domain=${domain}' \
-        -var 'ingress_ip=${getCoreDevIngressIP()}' \
-        -var 'ws_proxy_ip=${wsProxyLBIP}'`;
-
-    werft.log(installerSlices.DNS_ADD_RECORD, "Terraform command for create dns record: " + cmd)
-    exec(cmd, { ...metaEnv(), slice: installerSlices.DNS_ADD_RECORD });
+    createDNSRecord(domain, 'gitpod-dev-com', coreDevIngressIP, installerSlices.DNS_ADD_RECORD)
```
mads-hartmann, line 667: If you want to run all of these in parallel and then wait for them all to finish (or fail if one of them does) you can use [Promise.all ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all)

```ts
await Promise.all([
  createDNSRecord(domain, 'gitpod-dev-com', coreDevIngressIP, installerSlices.DNS_ADD_RECORD),
  createDNSRecord(`*.${domain}`, 'gitpod-dev-com', coreDevIngressIP, installerSlices.DNS_ADD_RECORD),
  createDNSRecord(`*.ws-dev.${domain}`, 'gitpod-dev-com', wsProxyLBIP, installerSlices.DNS_ADD_RECORD),
])
```

.werft/util/gcloud.ts
```diff
@@ -33,3 +35,28 @@ export async function deleteExternalIp(phase: string, name: string, region = "eu
...
+    const dnsClient = new DNS({
+        projectId: 'gitpod-dev',
+        keyFilename: GCLOUD_SERVICE_ACCOUNT_PATH
+    })
+    const zone = dnsClient.zone(dnsZone)
+    const record = new Record(zone, 'a', {
+        name: domain,
+        ttl: 300,
+        signatureRrdatas: [IP],
+        data: IP
```
mads-hartmann, line 52: The error you're getting sounds like a bug in @google-cloud/dns to be honest. You could try to remove the `signatureRrdatas` from here just because that's the only array you're passing in

mads-hartmann (commented):
Left a few promise-related suggestions ☺️

ArthurSens:
Everything seems to work well, except for ONE thing!

For some reason that I couldn't understand, creating the DNS record for the ws-proxy load balancer is failing. The error seems to be that the given IP is invalid for DNS records of type "A".

image

I'm currently troubleshooting by following [this guide](https://cloud.google.com/dns/docs/error-messages#invalidvalue), but it's not very helpful 😅 
image

ArthurSens:
> Everything seems to work well, except for ONE thing!
> 
> For some reason that I couldn't understand, creating the DNS record for the ws-proxy load balancer is failing. The error seems to be that the given IP is invalid for DNS records of type "A".
> 
> image
> 
> I'm currently troubleshooting by following [this guide](https://cloud.google.com/dns/docs/error-messages#invalidvalue), but it's not very helpful 😅 image

Actually, it was a typescript type validation error 🤡 

Alright, just testing it with Harvester previews and it should be good to go

meysholdt (approved):
Code changes LGTM

PR #9155, opened 2022-04-06
[kots]: enable use of a local registry

### LINKED ISSUE #9041: Allow for air-gapped installation
Airgapped installation is a key business driver. This needs the following to happen:

- [x] #9042
- [x] #9180
- [ ] #8559

### LINKED ISSUE #9180: Change the container images to use the KOTS registry format
In order to allow for airgapped installation, KOTS allows for `.airgap` files to be uploaded to a "local" container registry. These images are published to the image level only - for example, the image `quay.io/jetstack/cert-manager/cert-manager-image` would be published to `myregistry.com/cert-manager-image`.

The Installer has an [ImageName](https://github.com/gitpod-io/gitpod/blob/584f8d925406ed16bc9e813315436fa2b273ead9/install/installer/pkg/common/common.go#L356) function which all container images are passed through. Currently, if the configuration for the `repository` is set to the default Gitpod registry, it returns the image name given. However, if the `repository` is different, it prepends it so it would return `myregistry.com/jetstack/cert-manager/cert-manager-image`

In order to use the KOTS registry, this needs to switch it to the new format if a non-Gitpod registry is detected.

MrSimonEmms:
## Description

This will enable use of Replicated's airgapped functionality.

## Related Issue(s)

Fixes #9180
Fixed #9041

## How to test

1. Use the [self-hosted infra repo](https://github.com/MrSimonEmms/gitpod-self-hosted-infrastructure) to create an Azure airgapped installation. Set the `TF_VAR_domain_name` envvar to the domain name you want to use and `TF_VAR_enabled_airgapped=true`.
2. Download the latest airgapped bundle on the `dev-sje` channel
3. Run `make install` to get the command to install Gitpod via KOTS. This will also configure the local registry
4. In [localhost:8800](http://localhost:8800), configure and install the application. **IMPORTANT** this will hide the Gitpod registry config - this is because we must use the same registry that images are pulled from.

Important, this has been tested with a valid TLS cert issued via LetsEncrypt. Self-signed/corporate CA certs are outside the scope of this ticket

This workflow doesn't include anything past login as that requires a self-hosted GitLab instance on the same network which I've not yet done. In theory, this should work fine but any further stuff for that can be done as a future PR.

## Release Notes

```release-note
[kots]: enable use of a local registry
```


roboquat:

<details>

</details>

install/kots/manifests/gitpod-installer-job.yaml
```diff
@@ -24,7 +24,7 @@ spec:
       containers:
         - name: installer
           # This will normally be the release tag - using this tag as need the license evaluator
-          image: 'eu.gcr.io/gitpod-core-dev/build/installer:sje-installer-mini-config.0'
+          image: 'eu.gcr.io/gitpod-core-dev/build/installer:sje-airgapped.0'
```
MrSimonEmms, line 27: This is why I want to have a `latest` tag for the Installer

install/kots/Makefile
```diff
@@ -16,7 +16,7 @@ create_dev_release:
 		exit 1; \
 	fi
 
-	replicated release create --lint --ensure-channel --yaml-dir ${YAML_DIR} --promote ${REPLICATED_DEV_CHANNEL}
+	replicated release create --lint --ensure-channel --yaml-dir ${YAML_DIR} --promote ${REPLICATED_DEV_CHANNEL} --version="$(shell date +%s)"
```
geropl, line 19: Is this temporary or meant to be merged?

MrSimonEmms: To be merged. This is only for when a developer wants to push a release from their dev branch (eg, `dev-sje`) - airgapped releases need unique version numbers, so this is what this is for.

It won't be used for releases to the stable/beta/unstable channels

install/installer/pkg/common/common.go
```diff
@@ -339,7 +339,7 @@ func Affinity(orLabels ...string) *corev1.Affinity {
 	}
 }
 
-func RepoName(repo, name string) string {
+func RepoName(repo, name string, cfg *config.Config) string {
```
geropl, line 342: @MrSimonEmms Have you thought about adding tests for this? I wonder if we could have tests that render some config, and then parse those YAML to validate that:
 - all image refs are valid
 - if airgapped is enabled: all image refs are fully qualified
 - whatever other variant there might be

geropl: Alright, stupid me: It looks like there are [unit tests](https://github.com/gitpod-io/gitpod/pull/9155/files#diff-c7027dcc8c06120d3bd319bd1e23b4ed93915ffa83917b729cc091226dc72efaR47) already. :heavy_check_mark:

geropl (dismissed):
did _not_ test, but code changes LGTM

install/installer/pkg/common/common.go
MrSimonEmms: 🤣

install/installer/pkg/common/common.go
```diff
@@ -350,11 +350,26 @@ func RepoName(repo, name string) string {
...
+		}
+		return strings.Join(noNamespace, "/")
+	}
+
+	return prefString
 }
 
-func ImageName(repo, name, tag string) string {
-	ref := fmt.Sprintf("%s:%s", RepoName(repo, name), tag)
+func ImageName(repo, name, tag string, cfg *config.Config) string {
```
nandajavarma, line 371: nit question: why is `cfg` param added considering that only `cfg.Repository` is used, which in all calls  to the function seem to be same as `repo`?

MrSimonEmms: Fair question. It's done more as a nod to future-proofing the function - whilst I'm not expecting to need anything in future, having the config struct in there allows for avoiding having to do any change in future to the function interface

csweichel: Adding `cfg` here bloats the interface considerably (see comment above), especially compared to what it was. Let's try and get away without needing this.

install/installer/pkg/common/common.go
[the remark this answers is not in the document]
MrSimonEmms: It's not actually. The `repo` is wherever the original image is hosted, which might be Gitpod repo or `docker.io` (or something else). The other one is what the value in the `config.yaml` is.

As when this was raised by @nandajavarma, I used the Config struct to avoid having to change this in future.

install/installer/cmd/mirror_list.go
```diff
@@ -85,6 +86,117 @@ func init() {
 	mirrorListCmd.Flags().StringVarP(&mirrorListOpts.ConfigFN, "config", "c", os.Getenv("GITPOD_INSTALLER_CONFIG"), "path to the config file")
 }
 
+func renderAllKubernetesObject(cfgVersion string, cfg *configv1.Config) ([]string, error) {
```
csweichel, line 89: If I understand this function correctly it tries to exercise all code paths which might ever require an image. How do we make sure we don't inadvertently introduce a code path down the road which is not covered here, and end up with a broken airgapped installer?

install/installer/pkg/common/common.go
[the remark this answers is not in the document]
csweichel: We introduced the `RepoName` for exactly this reason: swap out the name of a repo with something set in the config. Looking at how `RepoName` is called, it's usually called with `RepoName(cfg.Repository, ...)` (through `ImageName`). There are cases where third-party repos are involved, but that's encapsulated in the `ThirdParty` repo function already.

install/kots/manifests/gitpod-installer-job.yaml
MrSimonEmms: I know. This line is actually changed by the werft job when it pushes to KOTS. It's just for avoidance for having to change line in PRs

install/installer/pkg/common/common.go
MrSimonEmms: We still need the `cfg.Repository` value in here. I'll change to just the single value rather than the structu

install/installer/pkg/common/common.go
```diff
@@ -350,11 +350,26 @@ func RepoName(repo, name string) string {
...
+		// If not in the Gitpod registry, don't use namespaces unless specified in the config.
+		// This is to match the KOTS registry format of no namespace in the name
+		s := strings.Split(prefString, "/")
+
+		noNamespace := []string{
+			cfg.Repository,
+			s[len(s)-1],
+		}
+		return strings.Join(noNamespace, "/")
+	}
```
csweichel, line 366: Could you give some background on what's happening here and what problem it solves?
It strikes me that this
1. is the cause of the `cfg` inclusion/need for passing `cfg.Repository` along,
2. possibly unsafe modification of image references without the use of the [reference](https://pkg.go.dev/github.com/docker/distribution/reference) package

MrSimonEmms: This is driven by KOTS. Their `.airgap` package changes the format of the images, so `docker.io/library/ubuntu:20.04` becomes `<url>/ubuntu:20.04` - note that the "namespace" `library` has been removed.

This is indeed the reason for requiring the `cfg.Repository` value. Because we decided (rightly) to make the use of airgapping implicit in the config, the established pattern has been "if value not equal to the default Gitpod registry".

I looked in the `reference` package and couldn't find a "remove namespace from container URL" function. If I've missed it, please let me know where it is as I'd love to use it

csweichel: Does this affect all images or just the ones from `docker.io/library`?

MrSimonEmms: All. I just used that as example

csweichel: That's really bad behaviour on Replicated's part 🤦  We should just get a registry that behaves like a proper mirror.

In that case, rather than making such a blatant exception implicit behaviour, what if we
- tied `RepoName` and `ImageName` to the render context,
- made the "drop repo name" explicit on the render context,
- ask the installer to implement the "drop repo name" behaviour using a command line flag.

Motivation and benefits:
- Makes this odd behaviour explicit: dropping the repository is really unusual. We're only doing this because replicated requires us to. By making this explicit, we'll know in a few months why we're doing this and what triggers it.
- Does not inflate the `ImageName` and `RepoName` API, which means we're not spreading complexity throughout the installer codebase and maintain testability.
- This "drop repo name" behaviour is needed by replicated, in which case we're running the installer from within a container we control. We can pass that flag ourselves. We still keep the config free from this concern.

MrSimonEmms: > That's really bad behaviour on Replicated's part facepalm We should just get a registry that behaves like a proper mirror.

Yup

Good idea about dropping the repo name in the config. We're still going to need to change the interface of the `ImageName` and/or `RepoName` functions, but I'll see what I can do 👍🏻

csweichel: I was thinking something akin to:

```Go

func (r *RenderContext) RepoName(repo, name string) string {
	var ref string
	if repo == "" {
		ref = name
	} else {
		ref = fmt.Sprintf("%s/%s", strings.TrimSuffix(repo, "/"), name)
	}
	pref, err := reference.ParseNormalizedNamed(ref)
	if err != nil {
		panic(fmt.Sprintf("cannot parse image repo %s: %v", ref, err))
	}

        if r.DropImageRepo {
		domain := reference.Domain(pref)
                segs := strings.Split(reference.Path(pref), "/")
                return fmt.Sprintf("%s/%s", domain, segs[len(segs) - 1])
        }

	return pref.String()
}

func (r *RenderContext) ImageName(repo, name, tag string) string {
	ref := fmt.Sprintf("%s:%s", r.RepoName(repo, name), tag)
	pref, err := reference.ParseNamed(ref)
	if err != nil {
		panic(fmt.Sprintf("cannot parse image ref %s: %v", ref, err))
	}
	if _, ok := pref.(reference.Tagged); !ok {
		panic(fmt.Sprintf("image ref %s has no tag: %v", ref, err))
	}

	return ref
}
```

MrSimonEmms:
/werft run publish-to-kots with-clean-slate-deployment no-preview

MrSimonEmms:
/werft run publish-to-kots with-clean-slate-deployment no-preview

MrSimonEmms:
/werft run publish-to-kots with-clean-slate-deployment no-preview

install/installer/pkg/common/common.go
MrSimonEmms: Changes made and tested

geropl (approved):
Reviewed again, looks even better now. Adding my rubber-stamp here, again. :crossed_fingers:

corneliusludmann:
@MrSimonEmms Started with the review but need more time testing it. Will finish my review tomorrow. :pray:

corneliusludmann (approved):

mustard-mh (approved):
I am not sure how to test, but code LGTM

PR #9175, opened 2022-04-07
[licensor] fix fallback for `defaultlicense`

nandajavarma:
## Description

This fixes the bug that stops the fallback (usage with limited features when number of users exceed 10) when no licence is installed, or in the case of replicated when licence domain does not match with the one issued (community licence). 

## Related Issue(s)

NA

## How to test

Currently, if one setup Gitpod via the `installer` or `replicated` using the community licence, if you increase the number of users before 10, admin dashboard will fail to be available. With this fix, once can try setting up replicated and admin dashboard will continue to be accessible after 10 users, whereas features like prebuild, snapshot etc. will not be available.

## Release Notes

```release-note
NONE
```


geropl (approved):
Thx, change make sense to me, tests look reasonable and are green, so: :+1:

Issue #8448, opened 2022-02-25
Alert when workspaces starting is zero but there are requests to start


kylos101:
## Is your feature request related to a problem? Please describe
There could be requests to start workspaces, but no workspaces may be starting.

## Describe the behaviour you'd like

If webapp requests to start a workspace, and workspaces haven't been started after some criteria is met, we must send an alert to page on-call staff.

## Describe alternatives you've considered

If there are [requests to startWorkspaces in this region's meta cluster](https://grafana.gitpod.io/explore?orgId=1&left=%5B%22now-1h%22,%22now%22,%22VictoriaMetrics%22,%7B%22refId%22:%22A%22,%22exemplar%22:false,%22instant%22:true,%22range%22:true,%22expr%22:%22histogram_quantile(0.55,%20sum%20(rate(gitpod_server_api_calls_duration_seconds_bucket%7Bmethod%3D~%5C%22startWorkspace%5C%22%7D%5B5m%5D))%20by%20(cluster,%20method,%20le))%22%7D%5D), and there are [no recent startup seconds](https://grafana.gitpod.io/explore?left=%5B%22now-1h%22,%22now%22,%22VictoriaMetrics%22,%7B%22expr%22:%22sum(rate(gitpod_ws_manager_workspace_startup_seconds_bucket%7Bcluster%3D~%5C%22us34%5C%22,type%3D%5C%22REGULAR%5C%22%7D%5B$__rate_interval%5D))%20by%20(le)%22,%22format%22:%22heatmap%22,%22intervalFactor%22:2,%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22P4169E866C3094E38%22%7D,%22interval%22:%22%22,%22exemplar%22:false%7D%5D&orgId=1), send an alert.

## Additional context

https://gitpod.slack.com/archives/C0345L4CGHM

princerachit:
>If webapp requests to start a workspace, and workspaces haven't been started after some criteria is met, we must send an alert to page on-call staff.

What do you mean by `workspaces haven't been started after some criteria is met`? We cannot map 1:1 each request from webapp to a workspace being started in workspace cluster.



### I find the alternative you mentioned better with slight modification

>If there are startWorkspaces requests, and there are no recent workspace startups, send an alert.

WDYT?

###  We can use the following alert query with above approach
```promql
[...]

PR #9190, opened 2022-04-08
[observability] Add alert when workspaces don't start

princerachit:
## Description

For the given formula the alerts can fire very often for xl clusters. To disable it I have created a custom pagerduty eventrule to suppress alerts being fired from the xl clusters. Refer to this [internal thread](https://gitpod.slack.com/archives/C0201TXMV54/p1650443926088999)

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/8448

## How to test

The alert generation itself does not need to be tested as this is a standard addition to alerts. However , you can review this as part of review:

### If alerts are suppressed for xl clusters
**NOTE**: Please make sure you inform the oncall about your tests, or override yourself to be the oncall before running the test

* Start a workspace and fetch kubeconfig of a cluster from which you want to generate the alert.
* port forward alertmanager service in the terminal 

```
kubectl port-forward svc/alertmanager-main 9093:9093
```

* From a different terminal run these one by one and check pager duty.

1. Create an alert with right name but cluster is not xl -> You should see an incident in pagerduty

```console
curl -d '[{"labels": {"Alertname": "GitpodWorkspacesNotStarting", "severity": "critical", "cluster": "us41", "container": "ws-daemon", "instance": "10.20.169.66:8443"}}]' http://localhost:9093/api/v1/alerts
```

2. Create an alert with right name and cluster set to xl -> You should NOT see an incident in pagerduty

```console
curl -d '[{"labels": {"Alertname": "GitpodWorkspacesNotStarting", "severity": "critical", "cluster": "us41xl", "container": "ws-daemon", "instance": "10.20.169.66:8443"}}]' http://localhost:9093/api/v1/alerts
```

3. Create an alert with wrong name but cluster is xl -> You should see an incident in pagerduty

```console
curl -d '[{"labels": {"Alertname": "G1itpodWorkspacesNotStarting", "severity": "critical", "cluster": "us41xl", "container": "ws-daemon", "instance": "10.20.169.66:8443"}}]' http://localhost:9093/api/v1/alerts
```

### Additional References
You can use this formula to plot graph (you would need to set cluster variable in dashboard settings, refer to the success criteria dashboard for the formula)

```
count(avg_over_time(gitpod_workspace_regular_not_active_percentage{cluster=~"$cluster"}[10m])) * count(rate(gitpod_ws_manager_workspace_startup_seconds_sum{type="REGULAR",cluster=~"$cluster"}[10m]) == 0)
```

<details>
    <summary>You can alternatively just import this dashboard in Grafana.</summary>
https://gist.github.com/princerachit/c28960e5afb4a316e0cfc35d3d674987
</details>

#### Past Data
Here is a list of various cluster screenshots. Each point in the cluster indicates that this alert would have been fired

<details>
    <summary>us38</summary>
image
</details>

<details>
    <summary>eu38</summary>

image
</details>

<details>
    <summary>us40</summary>
image
</details>

<details>
    <summary>eu40</summary>
image
</details>

<details>
    <summary>us39</summary>

image
</details>

<details>
    <summary>eu39</summary>

image
</details>

## Release Notes

```release-note
None
```

### DIFF
--- a/operations/observability/mixins/workspace/rules/components/workspaces/alerts.libsonnet
+++ b/operations/observability/mixins/workspace/rules/components/workspaces/alerts.libsonnet
@@ -73,6 +73,23 @@
               gitpod_workspace_regular_not_active_percentage > 0.15 AND sum(gitpod_ws_manager_workspace_activity_total) > 100
             |||,
           },
+          {
+            alert: 'GitpodWorkspacesNotStarting',
+            labels: {
+              severity: 'critical',
+            },
+            'for': '10m',
+            annotations: {
+              runbook_url: 'https://github.com/gitpod-io/runbooks/blob/main/runbooks/GitpodWorkspaceNotStarting.md',
+              summary: 'workspaces are not starting',
+              description: 'inactive regular workspaces exists but workspaces are not being started',
+            },
+            expr: |||
+              avg_over_time(gitpod_workspace_regular_not_active_percentage[1m]) > 0
+              AND
+              rate(gitpod_ws_manager_workspace_startup_seconds_sum{type="REGULAR"}[1m]) == 0
+            |||,
+          },
         ],
       },
     ],


operations/observability/mixins/workspace/rules/components/workspaces/alerts.libsonnet
```diff
@@ -73,6 +73,23 @@
               gitpod_workspace_regular_not_active_percentage > 0.15 AND sum(gitpod_ws_manager_workspace_activity_total) > 100
             |||,
           },
+          {
+            alert: 'GitpodWorkspacesNotStarting',
+            labels: {
+              severity: 'critical',
+            },
+            'for': '10m',
```
kylos101, line 81: Hi @princerachit does it make sense to leave this at 10m, when the expression is inspecting rate for 5m? :thinking: Seems odd to me, but, maybe I am missing something.

princerachit: I can change this but it does not make any difference, afterall we are checking if the value is zero

kylos101: Can you share if this would have triggered in the last 30 or 60 days with support detail? 

My understanding is `for` controls how long an expression must be true before the alert fires. And the underlying expression is looking for a window of time where the sum of starts is 0.

I would expect this alert to fire after 20 minutes. For 10 minutes, the following must be true:
1. regular not active has been > than zero
2. no workspaces have started in the last 10 minutes

I think it would have triggered if we were using 5m and 5m, triggered [here](https://grafana.gitpod.io/?orgId=1&var-datasource=VictoriaMetrics&var-cluster=us38xl2&from=1649419599314&to=1649422342364).

image

kylos101: @princerachit marked as unresolved so I can get your [feedback](https://github.com/gitpod-io/gitpod/pull/9190#discussion_r846264519).

operations/observability/mixins/workspace/rules/components/workspaces/alerts.libsonnet
```diff
@@ -73,6 +73,23 @@
...
+            annotations: {
+              runbook_url: 'https://github.com/gitpod-io/runbooks/blob/main/runbooks/GitpodWorkspaceNotStarting.md',
+              summary: 'workspaces are not starting',
+              description: 'inactive regular workspaces exists but workspaces are not being started',
+            },
+            expr: |||
+              gitpod_workspace_regular_not_active_percentage > 0.0
+              AND
+              sum(rate(gitpod_ws_manager_workspace_startup_seconds_bucket{type="REGULAR",le="+Inf"}[10m])) == 0
```
ArthurSens, line 90:
```suggestion
              sum(rate(gitpod_ws_manager_workspace_startup_seconds_count{type="REGULAR"}[10m])) == 0
```

Every prometheus histogram comes with a counter 🙂

operations/observability/mixins/workspace/rules/components/workspaces/alerts.libsonnet
ArthurSens: > I would expect this alert to fire after 20 minutes.

That is 100% correct, 

When using `rate(metric[10m])`, the value will be calculated using the latest 10 minutes of data. This means that the value cannot increase for the last 10 minutes for this value ever becomes 0. Adding a `for` to the alert will just increase the time for the alert to be triggered.

For the alert to be triggered in 10 minutes you can:
* rate[5m] + for: 5m
* rate[10m]

ArthurSens: [Some extra info on rates](https://promlabs.com/blog/2021/01/29/how-exactly-does-promql-calculate-rates) :)

kylos101:
@ArthurSens :pray: thank you for taking a gander!

princerachit:
I have fixed the calculation issues and made it similar to the way we have been doing e.g. 1m for 10m.

I have also updated metrics so that formula becomes simple and intutive.

You can refer to the graph url that I just added in the description @kylos101 to see how often this alert can get triggered

kylos101:
@princerachit we must make it large enough so that it doesn't trigger for XL clusters. Is 20 minutes enough time? Please look in Grafana and update accordingly.

kylos101:
> @princerachit we must make it large enough so that it doesn't trigger for XL clusters. Is 20 minutes enough time? Please look in Grafana and update accordingly the alert accordingly, sharing a picture so the reviewer has context as to why you set the value.

Hey @princerachit , sounds like the XL clusters have gaps that are too large. Can we exclude them from this alert for now? That way you can ship a :skateboard:

princerachit:
I have reached out to the team internally on how to implement this filtering of xl clusters. Based on the promethus configuration that we have, it seems impossible to disable alert from prometheus running in the cluster itself. This is because we use external_labels to add cluster information.

kylos101:
@princerachit can you share a short loom demonstrating the test? I'm not able to recreate.

princerachit:
@kylos101 I have added all the information in the description. I have also added screenshots.
I have not captured this in loom as the critical channel was already polluted with my tests.

kylos101:
Thank you, @princerachit . This is an excellent start, and we can tune from here. 

Can you update the runbook, to say something like, please try starting workspaces in the context cluster? If the on-caller can start workspaces in the context cluster (or not) I think that will influence whether or not they escalate further. It doesn't mention to do this, and it seems like a pretty good first step. For example, they might have to use a VPN to access the US or EU cluster.

kylos101 (approved):
:+1:

PR #9218, opened 2022-04-11
[installer] (optionally) surface "metadata.shortname"

geropl:
## Related Issue(s)

Fixes #

## Release Notes

```release-note
installer: make shortname configurable
```


easyCZ (dismissed):
Makes sense.

install/installer/pkg/common/render.go
```diff
@@ -123,10 +124,11 @@ func NewRenderContext(cfg config.Config, versionManifest versions.Manifest, name
...
 
 	ctx := &RenderContext{
-		Config:             cfg,
-		VersionManifest:    versionManifest,
-		Namespace:          namespace,
-		experimentalConfig: us,
+		Config:                cfg,
+		VersionManifest:       versionManifest,
+		Namespace:             namespace,
+		InstallationShortname: cfg.Metadata.InstallationShortname,
```
MrSimonEmms, line 130: Is there a reason why you put this at the root level rather than getting it from the `Config` object?

install/installer/pkg/components/ws-manager-bridge/configmap.go
```diff
@@ -16,7 +16,7 @@ import (
 
 func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 	wsmbcfg := Configuration{
-		Installation:                        "",
+		Installation:                        ctx.InstallationShortname,
```
MrSimonEmms, line 19: Related to above, we should probably use `ctx.Config.Metadata.InstallationShortname`

install/installer/pkg/config/v1/config.go
```diff
@@ -99,6 +100,8 @@ type Config struct {
 
 type Metadata struct {
 	Region string `json:"region" validate:"required"`
+	// InstallationShortname establishes the "identity" of the (application) cluster.
+	InstallationShortname string `json:"shortname,omitempty"`
```
MrSimonEmms, line 104:
```suggestion
	InstallationShortname *string `json:"shortname,omitempty"`
```

May be worth replacing this with a pointer so the `omitempty` is respected? If you do this, the value will default to `nil` rather than an empty string.

If you want to stay with this, can you run `go run . init > example-config.yaml` to generate the new example config please?

install/installer/pkg/common/render.go
geropl: Ah, no, just overlooked that. Will change.

install/installer/pkg/config/v1/config.go
geropl: Thx for the hint, will do!

install/installer/pkg/components/ws-manager-bridge/configmap.go
MrSimonEmms: If you change to a pointer, use the `pointer.StringDeref` function. That will convert a pointer to a string and allow you specify a default value

install/installer/pkg/common/render.go
MrSimonEmms: Don't worry - I did the same as well

install/installer/pkg/components/ws-manager-bridge/configmap.go
geropl: Where should that function be used? It only makes sense to set a default in a central location, so I opted for `config.go`.

MrSimonEmms: When you've finished lunch, do you want ~10 mins pairing on it? Probably easier than async

geropl: 👍

geropl: Resolved in code.

install/installer/pkg/config/v1/config.go
```diff
@@ -99,6 +100,8 @@ type Config struct {
 
 type Metadata struct {
 	Region string `json:"region" validate:"required"`
+	// InstallationShortname establishes the "identity" of the (application) cluster.
+	InstallationShortname *string `json:"shortname,omitempty"`
```
MrSimonEmms, line 104:
```suggestion
	InstallationShortname *string `json:"shortname,omitempty" validate:"required"`
```

If this is a required field, please add this. The `omitempty` can probably be removed and, strictly speaking, it doesn't need to be a pointer (apologies for that, I misunderstood the requirement).

If you'd prefer me to make the change, happy to do so

MrSimonEmms (approved):
🎉

install/installer/pkg/config/v1/config.go
geropl: We sorted out in sync:
 - `defaults()` is currently only used to initialize the very first config file
 - but _not_ to merge it as base "under" an already existing one
 - 
So we mark this as "required" to be in line with current semantics and usage. :heavy_check_mark:

geropl:
@andrew-farries Could you :heavy_check_mark: ? :pray:

Issue #8911, opened 2022-03-21
[prebuilds] Automatically disable prebuilds for projects without active Workspace usage


easyCZ:
## Is your feature request related to a problem? Please describe

In some instances, users configure Gitpod for a repository but do not actively start workspaces to develop in Gitpod.

Because of our prebuilds - optimization to get you to run workspaces faster, we continue to react to pushes to the repository and create prebuilds. However, these prebuilds are not useful as the users do not actively use gitpod to develop these.

In this scenario, we should consider reduce the number of prebuilds triggered from these repositories. Some options for doing this:
* disable prebuilds entirely if no workpsace started in the last 7 days
* Limit prebuilds to main only (and once a day) if no workspace started in 7 days

With this change, we need to strike the balance of convenience for the user and fairness in a multi-tenant system.

Originally raised by @kylos101.

jankeromnes:
Many thanks @easyCZ for filing this issue! 💯 I'm very much in favour of this.

> if no workpsace started in the last 7 days

- I wonder if 7 days is maybe a bit too short as a checking period (e.g. imagine a project with just a few developers, and they all happen to take a week of holidays at the same time -- disabling their prebuilds might be a bit too disruptive). How about 3 months?

- Also, I wonder whether it makes sense to check for "did any workspace start for this project", or if we should more precisely look at "did any of the prebuilds get used as a base or a workspace" (somewhat similar, but I'd argue we're looking precisely for prebuilds that are 100% unused here -- although, looking for projects that are 100% unused could also be interesting, e.g. if we want to garbage collect projects one day? 🤔)

---

As for a technical approach, I guess we could implement a `uninstallAutomatedPrebuilds` method in `RepositoryService`, similar to:

https://github.com/gitpod-io/gitpod/blob/2f0b9462b0451f0974434fa36bd757420ef04e34/components/server/src/repohost/repo-service.ts#L20

jankeromnes:
See also [research issue](https://github.com/gitpod-io/ops/issues/1590) (internal)

kylos101:
@jankeromnes @easyCZ Another dimension to consider is tuning how aggressively we stop doing prebuilds based on the plan for the user. For example, we might want to be clean-up more aggressively for free plans, but, less aggressively with pay plans.

kylos101:
@easyCZ I seem to recall this (or something like it) being socialized recently, and just wanted to bump this issue for awareness. :wave: Hope you're well!

[...]

Issue #7010, opened 2021-12-02
Show feedback on the project card when there's need to reconfigure the app for GitHub or reinstall the webhook for GitiLab


gtsiolis:
### Problem to solve

Currently, when removing (or reconfiguring) the GitHub app or removing the GitLab webhook the project, there's no feedback to the user the GitHub app needs to be reconfigured or the webhook is missing. See also https://github.com/gitpod-io/gitpod/issues/7009.

### Proposal

On the project card, show some feedback there's project needs to be reconfigured, whether this is GitHub app reconfiguration or adding the webhook back.

Re-posting the design specs from https://github.com/gitpod-io/gitpod/issues/4420, see project with configuration error:

| Projects |
|-|
| <img width="1440" alt="121364262-4c4aea00-c940-11eb-8f84-135402d14066 (3)" src="https://user-images.githubusercontent.com/120486/144334200-0632d6da-15d5-4c9f-8cbb-ede442ea3ced.png"> | 

| Configuration Error |
|-|
| <img width="454" alt="121364262-4c4aea00-c940-11eb-8f84-135402d14066 (3) copy" src="https://user-images.githubusercontent.com/120486/144334238-6fed1085-9025-409b-9378-d5443c54cfee.png"> |

jldec:
Renaming this issue to detect and show some feedback.

Note: Hovers/tooltips can be more difficult to use on touch devices - do you have a suggestion for working around that?

jldec:
@gtsiolis - as mentioned in #7011 - let's track this particular scenario in this one issue for now and try to deliver just the MVC to detect and indicate **somewhere** first. 

It may make more sense to followup with something on the project settings page.

jldec:
@AlexTugarev - Let's figure out an MVC to help users understand the understand the state of a prebuild configuration for a specific project. Could we show a repository prebuild configuration status on `/projects/<project-name>/prebuilds` with one of the states below? 

### Prebuild configuration states

#### Any repo
-  `OK - last webhook received at <date/time> for <event type>`
- `No prebuild configured for this repo` (no prebuild in gitpod.yml - file or db) 

#### GitHub
- `Gitpod app on GitHub not installed for org or user <name>`
- `Gitpod app on GitHub not configured to access <repo>`

#### Other git integrations
- `No webhook installed for <repo>` (is this possible to detect?)  



cc: @gtsiolis

PR #9219, opened 2022-04-11
Stop running prebuilds for inactive projects (10+ weeks)

jankeromnes:
## Description

Stop running prebuilds for projects that did not start a workspace in the last 10+ weeks.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/8911
Drive-by fixes prebuild rate limits again

## How to test

1. Add a Project
2. Verify that Prebuilds are triggered successfully and workspaces can be started
3. Verify that `lastWorkspaceStart` and `lastWebhookReceived` are properly set in `d_b_project_usage`
4. Manually set `lastWorkspaceStart` to a date more than 10 weeks ago in `d_b_project_usage`
5. Trigger new Prebuilds -- verify that they now get auto-cancelled on start due to Project inactivity

## Release Notes

```release-note
Stop running prebuilds for projects that did not start a workspace in the last 10+ weeks
```

### DIFF
--- a/components/gitpod-db/src/typeorm/entity/db-project-usage.ts
+++ b/components/gitpod-db/src/typeorm/entity/db-project-usage.ts
@@ -1,0 +1,26 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the Gitpod Enterprise Source Code License,
+ * See License.enterprise.txt in the project root folder.
+ */
+
+import { Entity, Column, PrimaryColumn } from "typeorm";
+
+import { TypeORM } from "../../typeorm/typeorm";
+
+@Entity()
+// on DB but not Typeorm: @Index("ind_dbsync", ["_lastModified"])   // DBSync
+export class DBProjectUsage {
+    @PrimaryColumn(TypeORM.UUID_COLUMN_TYPE)
+    projectId: string;
+
+    @Column("varchar")
+    lastWebhookReceived: string;
+
+    @Column("varchar")
+    lastWorkspaceStart: string;
+
+    // This column triggers the db-sync deletion mechanism. It's not intended for public consumption.
+    @Column()
+    deleted: boolean;
+}
--- a/components/gitpod-db/src/typeorm/migration/1649667202321-ProjectUsage.ts
+++ b/components/gitpod-db/src/typeorm/migration/1649667202321-ProjectUsage.ts
@@ -1,0 +1,17 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class ProjectUsage1649667202321 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(
+            "CREATE TABLE IF NOT EXISTS `d_b_project_usage` (`projectId` char(36) NOT NULL, `lastWebhookReceived` varchar(255) NOT NULL DEFAULT '', `lastWorkspaceStart` varchar(255) NOT NULL DEFAULT '', `deleted` tinyint(4) NOT NULL DEFAULT '0', `_lastModified` timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6), PRIMARY KEY (`projectId`), KEY `ind_dbsync` (`_lastModified`)) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;",
+        );
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {}
+}
--- a/components/server/ee/src/prebuilds/github-enterprise-app.ts
+++ b/components/server/ee/src/prebuilds/github-enterprise-app.ts
@@ -116,6 +116,15 @@ export class GitHubEnterpriseApp {
     ): Promise<StartPrebuildResult | undefined> {
         const span = TraceContext.startSpan("GitHubEnterpriseApp.handlePushHook", ctx);
         try {
+            const cloneURL = payload.repository.clone_url;
+            const projectAndOwner = await this.findProjectAndOwner(cloneURL, user);
+            if (projectAndOwner.project) {
+                /* tslint:disable-next-line */
+                /** no await */ this.projectDB.updateProjectUsage(projectAndOwner.project.id, {
+                    lastWebhookReceived: new Date().toISOString(),
+                });
+            }
+
             const contextURL = this.createContextUrl(payload);
             span.setTag("contextURL", contextURL);
             const context = (await this.contextParser.handle({ span }, user, contextURL)) as CommitContext;
@@ -127,15 +136,13 @@ export class GitHubEnterpriseApp {
 
             log.debug("GitHub Enterprise push event: Starting prebuild.", { contextURL });
 
-            const cloneURL = payload.repository.clone_url;
-            const projectAndOwner = await this.findProjectAndOwner(cloneURL, user);
             const commitInfo = await this.getCommitInfo(user, payload.repository.url, payload.after);
             const ws = await this.prebuildManager.startPrebuild(
                 { span },
                 {
                     context,
                     user: projectAndOwner.user,
-                    project: projectAndOwner?.project,
+                    project: projectAndOwner.project,
                     commitInfo,
                 },
             );
--- a/components/gitpod-db/src/project-db.ts
+++ b/components/gitpod-db/src/project-db.ts
@@ -4,7 +4,7 @@
  * See License.enterprise.txt in the project root folder.
  */
 
-import { PartialProject, Project, ProjectEnvVar, ProjectEnvVarWithValue } from "@gitpod/gitpod-protocol";
+import { PartialProject, Project, ProjectEnvVar, ProjectEnvVarWithValue, ProjectUsage } from "@gitpod/gitpod-protocol";
 
 export const ProjectDB = Symbol("ProjectDB");
 export interface ProjectDB {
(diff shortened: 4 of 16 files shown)


gtsiolis:
**fyi**: I've opened a follow up issue with a feature request to surface this automatic pause action better on the dashboard as this could be confusing to users coming back to inactive projects in the dashboard, see https://github.com/gitpod-io/gitpod/issues/9232. Feedback is welcome!

Cc @jankeromnes @easyCZ @jldec

jankeromnes:
Looks like this now when a project becomes inactive, but still triggers new prebuilds:

<table><tr><td>
Screenshot 2022-04-12 at 11 30 54
</td><td>
Screenshot 2022-04-12 at 11 31 06
</td></tr></table>

components/gitpod-db/src/typeorm/entity/db-project-usage.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import { TypeORM } from "../../typeorm/typeorm";
+
+@Entity()
+// on DB but not Typeorm: @Index("ind_dbsync", ["_lastModified"])   // DBSync
+export class DBProjectUsage {
+    @PrimaryColumn(TypeORM.UUID_COLUMN_TYPE)
+    projectId: string;
+
+    @Column("varchar")
```
easyCZ, line 17: What's historically the reason for using varchar to store timestamps? Is it possible to use a DATETIME?

jankeromnes: Good question -- I don't have a strong opinion here, and was just following established practice 😊

jankeromnes: > The reason I ask is that when prototyping an ORM for Golang, Varchar timestamps make it much more complicated - this is further coupled with the somewhat non-standard usage of ISO8601 in javascript to represent a timestamp, which isn't quite RFC3339 compatible.

I think your concerns are valid, and there is definitely room for improvement in our DB types. 👍

> Definitely not trying to block this PR on this, mostly looking to understand historical reasoning for the varchar choice. Feel free to ignore this comment to land the change.

Thanks! I would love to, but first I need a review. 😇

easyCZ: Still reviewing :)

jankeromnes: That's awesome, thanks a bunch 🙏

I'll try to practice our new review process by assigning this Pull Request to you then (but please feel free to un-assign yourself again if that didn't make sense!)

easyCZ:
One alternative approach we could consider is to load the `d_b_project_usage` async from existing tables (Prebuild, WorkspaceInstance) to treat it like more of a cache. Doing this would achieve the following:
1. The handling code for a project doesn't need to be aware of the usage tracking - this eliminates opportunity for a bug where we forget to add the call to track
2. It derives the usage from a single place, rather than multiple
3. It makes the update process async which does not impact our latency on project related calls (the setting of the value).

WDYT?

components/gitpod-db/src/typeorm/entity/db-project-usage.ts
```diff
@@ -0,0 +1,26 @@
...
+    lastWebhookReceived: string;
```
easyCZ, line 18: Is this field strictly needed for our first version? We do set it, but we're not using it in read paths.

components/gitpod-db/src/typeorm/migration/1649667202321-ProjectUsage.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class ProjectUsage1649667202321 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(
+            "CREATE TABLE IF NOT EXISTS `d_b_project_usage` (`projectId` char(36) NOT NULL, `lastWebhookReceived` varchar(255) NOT NULL DEFAULT '', `lastWorkspaceStart` varchar(255) NOT NULL DEFAULT '', `deleted` tinyint(4) NOT NULL DEFAULT '0', `_lastModified` timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6), PRIMARY KEY (`projectId`), KEY `ind_dbsync` (`_lastModified`)) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;",
```
easyCZ, line 12: Here's we're setting a limit on `projectId` of 36 chars, should this also be reflected in the `precision: 36` field on the Column of the ORM model?

components/server/ee/src/prebuilds/github-enterprise-app.ts
```diff
@@ -116,6 +116,15 @@ export class GitHubEnterpriseApp {
     ): Promise<StartPrebuildResult | undefined> {
         const span = TraceContext.startSpan("GitHubEnterpriseApp.handlePushHook", ctx);
         try {
+            const cloneURL = payload.repository.clone_url;
+            const projectAndOwner = await this.findProjectAndOwner(cloneURL, user);
+            if (projectAndOwner.project) {
```
easyCZ, line 121: At the point of this call, how do we know that the project specified in the payload in fact belongs to the correct user? Is it possible to poison our data across projects here?

Mostly lacking context, so trying to understand better how the payload from the webhook is authhorized

easyCZ (approved):
PR looks good, and worked for me based on the test specified. Leaving a couple of questions for my own context. Approving with a hold.

/hold

Posting our slack conversation about splitting PRs here as well for visibility.
For the PR, with regards to tiny PRs. I'd split it into the following:

1. Add a method to call to updateProjectUsage but don't write to the DB, only increment a metric
2. Create the table, it's unused
Get data that this works
3. Wire them together - updateProjectUsage actually writes to the DB

components/gitpod-db/src/typeorm/entity/db-project-usage.ts
jankeromnes: Good catch, it's not strictly needed here indeed. However, since I modified very related code, I thought this small drive-by change could be a good first step toward fixing https://github.com/gitpod-io/gitpod/issues/7010

easyCZ: I'll leave it up to you. 

For surfacing which webhook we've received, I'd expose it as a first class-table with webhooks (and webhook payloads), such that we can show what we've received in the webhook as well (would definitely be useful for our debugging).

jankeromnes: Good idea. I'll leave this as is, and we can decide later whether we want a full-fledged table of webhook events, or whether showing a simple timestamp of the last event is good enough. 🛹

components/gitpod-db/src/typeorm/migration/1649667202321-ProjectUsage.ts
jankeromnes: I'm not too familiar with the intricacies of our ORM model, so I wouldn't know the pros and cons of doing this.

Briefly looking through our code, it seems that we typically use `precision` for timestamps, but not for IDs. If you think this should be changed, could you please open an issue for this? 🙏

components/server/ee/src/prebuilds/github-enterprise-app.ts
jankeromnes: > At the point of this call, how do we know that the project specified in the payload in fact belongs to the correct user?

That's a good question. I don't know about webhook events from the GitHub App, but for GitLab, we look at the provided (user + project) token in order to make sure that the project specified in the payload actually belongs to the user who installed this webhook:

https://github.com/gitpod-io/gitpod/blob/e8fa6d20d2c18cdb0f02ff2cfd483388c5c5dac9/components/server/ee/src/prebuilds/gitlab-app.ts#L61-L97

We do something similar for GitHub Enterprise webhooks, where the secret token is not sent as part of the webhook payload, but is used to sign the webhook payload (which we can verify against the user's tokens):

https://github.com/gitpod-io/gitpod/blob/e8fa6d20d2c18cdb0f02ff2cfd483388c5c5dac9/components/server/ee/src/prebuilds/github-enterprise-app.ts#L89-L106

> Is it possible to poison our data across projects here?

I don't think it is possible to send forged webhook events to Gitpod.

However, your question does remind me of an issue where Pull Requests can come from forks, but trigger a Prebuild for the main repository, which can be problematic in some cases: https://github.com/gitpod-io/security/issues/26

jankeromnes:
Many thanks for your very helpful review and suggestions @easyCZ! 🙏

> One alternative approach we could consider is to load the `d_b_project_usage` async from existing tables (Prebuild, WorkspaceInstance) to treat it like more of a cache. Doing this would achieve the following:
> 1. The handling code for a project doesn't need to be aware of the usage tracking - this eliminates opportunity for a bug where we forget to add the call to track
> 2. It derives the usage from a single place, rather than multiple
> 3. It makes the update process async which does not impact our latency on project related calls (the setting of the value).
> 
> WDYT?

I like the idea of doing these things more async, because it could make our system simpler and much less subtle/magical.

(For example, I initially updated `lastWorkspaceStart` only when starting from a prebuild, so starting a workspace not from a prebuild wouldn't count -- I've since added the missing update code, but it's pretty easy to imagine a future refactor where someone accidentally forgets one of these updates again.)

However, it feels like we're lacking some good framework code for such asynchronous jobs that regularly process large amounts of data in the background. I agree that we have to start somewhere, but I'm a bit hesitant to volunteer this Pull Request to be the first "async background processing job", given that it can already provide value to users as is. 🛹

---

I've also replied to your questions in-line. Many thanks for diving deeper into all these systems! I find your questions and instincts for improvements quite relevant, and am keen to help improve things or provide context whever I can. 🙂

Also, removing the hold, as I think this is good to go as is. Happy to improve anything in follow-ups. 🚀

/unhold

Issue #6859, opened 2021-11-24
[observability] Add alert when workspace failure rate increases


csweichel:
We need an alert when the workspace failure rate in any cluster exceeds 1 failure per second for at least 5 minutes.
As a follow up, we'll want to write a runbook for this case.

kylos101:
@princerachit as you're knee deep in alerts now, may I ask you to queue this one as well? It'll also need a related runbook as Chris mentions.

This will complete what @sagor999 mentioned a few days ago to me in a conversation as to what he looks for:
1. High normalized averages <-- covered now with warnings and on-call
2. High error rate <-- not covered, but this issue will do
3. Regular not active coupled with no workspaces starting <-- you are working this now in another issue

Are the trends we look for.

princerachit:
When did we such an issue last time? I want to understand how did we decided on 1failure per second. I have raised this PR to address the issue https://github.com/gitpod-io/gitpod/pull/9220

kylos101:
Hmm, good question, @princerachit ! 

What do you see over the last 60 days in Gitpod for how often this rule would have fired, if it were configured at 1 per second? Depending on the frequency you see, you might want to change the condition higher or lower.

What else can you check for clues? I am thinking description in incident.io, and maybe Slack threads for strings like "high failure rate" or "failures per second".

PR #9220, opened 2022-04-11
[observability] Add alert rule for high ws failure

princerachit:
## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/6859

## How to test

NA

## Release Notes

```release-note
None
```

### DIFF
--- a/operations/observability/mixins/workspace/rules/components/workspaces/alerts.libsonnet
+++ b/operations/observability/mixins/workspace/rules/components/workspaces/alerts.libsonnet
@@ -44,6 +44,20 @@
             |||,
           },
           {
+            alert: 'GitpodWorkspaceHighFailureRate',
+            labels: {
+              severity: 'critical',
+            },
+            annotations: {
+              runbook_url: 'https://github.com/gitpod-io/runbooks/blob/main/runbooks/GitpodWorkspaceHighFailureRate.md',
+              summary: 'Workspaces are failing',
+              description: 'Multiple workspaces are failing for the last 5 minutes',
+            },
+            expr: |||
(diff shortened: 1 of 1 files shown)


operations/observability/mixins/workspace/rules/components/workspaces/alerts.libsonnet
```diff
@@ -43,6 +43,20 @@
               ) without(phase) > 5
             |||,
           },
+          {
+            alert: 'GitpodWorkspaceHighFailureRate',
+            labels: {
+              severity: 'critical',
+            },
+            annotations: {
+              runbook_url: '',
```
sagor999, line 52: I think we need runbook for all our alerts, no?

kylos101: Hi @princerachit I shared a response yesterday, do you need anything else from the team?

princerachit: Nope, I have updated the runbook url.

operations/observability/mixins/workspace/rules/components/workspaces/alerts.libsonnet
```diff
@@ -43,6 +43,20 @@
               ) without(phase) > 5
             |||,
           },
+          {
+            alert: 'GitpodWorkspaceHighFailureRate',
```
sagor999, line 47: hm. this alert seems to be missing a `for` statement. I think it might default to 1min or something in that case?

princerachit: we are calculating rate for five minutes itself so we don't need a for statement.

sagor999 (approved):

Issue #8191, opened 2022-02-14
[code] fix gp cli while running next to jb product


akosyakov:
Since https://github.com/gitpod-io/gitpod/pull/8025 gp open/preview always exepct JB product if it is enabled, so they break in VS Code Web. We should redesign it by relying on notification service on supervisor instead of env vars.

andreafalzetti:
after a chat with @akosyakov we concluded that we could start working on this feature incrementally, by adding initially only the endpoint in supervisor and testing it internally by supervisor CLI.

Then, we could have a follow-up ticket to integrate with gp-cli and so on.

andreafalzetti:
@mustard-mh and I started working on this, we pair programmed for a couple of hours today and made some progress. Besides adding two new endpoints to `supervisor-api` we have discussed the integration with `gp cli`. 

We've considered 2 approaches, when users use `gp open` or `gp preview`:
A. `gp  cli` calls `supervisor` which is now aware of the latest active IDE/client and supervisor executes the IDE's binary directly.
B. `gp cli` calls `supervisor` to retrieve the latest active IDE/client, and `gp cli` executes the IDE's binary.

The biggest downside of approach A is that we need to implement in supervisor 2 endpoints, one for `open` one for `preview` mapping all parameters and flags of those 2 commands. For example `gp open` has a `--wait` flag which we should make sure to pass to supervisor. We would stricly couple gp-cli commands and supevisor API.

Another small downside for approach A, is that potentially makes harder returning any error to the user. If the `open` and `preview` logic is executed from `supervisor`, we would need to pro-actively make sure we can report back to the user, in case of failure. Instead, if the logic lives in `gp cli` the error would be returned directly to the user.

mustard-mh:
What if we do not use `<ide_identity>` like `vscode` `jetbrains_gateway` etc enum, just use `<ide_location>` like `/ide-desktop/bin/idea-cli` so that custom editor can be used here (eg. Users open workspaces without choose any IDE, but use vim with ssh)

And provide cmd like `gp set-editor <editor_cli_location>`

And we should also provide **How to Use Custom IDE (via ssh)** section in our doc and here https://github.com/gitpod-io/gitpod/issues/6707

akosyakov:
I think the upside of A that we have clean interface and hide implementation details of working with concrete IDEs behind them instead of relying on other ways. B does not sound so clean, since there will be one implicit protocol via interface and another implicit via binaries.

[...]

PR #9224, opened 2022-04-11
[supervisor-api] New API endpoint to set active client

andreafalzetti:
## Description
This is the first step toward fixing #8191. The problem we're trying to solve is: make `gp open` and `gp preview` work when: User's preferred IDE in settings is a JetBrains product, however, the user decides to open a workspace with VS Code/ VS Code browser. When this occurs, the CLI commands will still open or preview in JetBrains instead of VS Code.

To solve the issue, we need to keep track of the actual latest IDE used.

To achieve that, in this PR we propose to extend `notifications.proto`, adding a new message type called `SubscribeActive`. This will be used by clients (i.e. IDEs) to subscribe as "latest active client" with supervisor. They will be listening to the returned stream waiting for actions to run (e.g. `open`, `preview`).

This stream will become useful when we continue working on #8191 and update `gp-cli` `open` and `preview` commands to notify supervisor via the other new message type `Action` and trigger a notification from `supervisor` to the active client, to run `open` or `preview`. Finally, the last new message type `ActionRespond` will be used to inform supervisor about outcome of the Action (success/fail).

Note:
- _@mustard-mh and @andreafalzetti worked in pair on this PR_
- We did consider using the method `Subscribe` already available in `notifications.proto`, however current notifications are currently "actual text notifications" that show up in the IDE with a toast UI element. We would need to make sure this new type of notifications are "silent" and do not appear to the user (as they are not really notifications). Therfore we opted for not reusiung `Subscribe` and create a new but similar message type.
- This also opens a debate whether we should actually extend the `notification.proto` file at all, or create a new one, or use something like `control.proto`.

## Related Issue(s)

Starts #8191

## How to test
N/A

## Release Notes
```release-note
NONE
```


roboquat:

<details>

</details>

components/supervisor-api/notification.proto
```diff
@@ -11,60 +11,93 @@ import "google/api/annotations.proto";
...
-    uint64 requestId = 1;
-    NotifyResponse response = 2;
+  uint64 requestId = 1;
+  NotifyResponse response = 2;
+}
+
+message RespondResponse {}
+
+enum ActiveClient {
+  code = 0;
```
akosyakov, line 92: I think identity should be just a string, that it can be extended by 3rd party IDEs later.

Most likely it also won't be `code`, but more like someting like `code-${uuid}`, where is an uuid is session ID of VS Code window.

akosyakov: We should also allow clients to unset active, i.e. if we detect that VS Code window is not focused anymore then it should indicate it to supervisor.

mustard-mh: @akosyakov How about https://github.com/gitpod-io/gitpod/issues/8191#issuecomment-1096110343 this?

components/supervisor-api/notification.proto
```diff
@@ -11,60 +11,129 @@ import "google/api/annotations.proto";
...
+      post : "/v1/notification/respond"
+    };
+  }
+
+  // SetActiveClient to inform supervisor about
+  // which is the latest client actively used by the user.
+  // Typically called by the IDE.
+  // Will response what client should do like `open` `preview`
+  rpc SetActiveClient(SetActiveClientRequest)
+      returns (stream SetActiveClientResponse) {
```
akosyakov, line 47: why is it stream?

components/supervisor-api/notification.proto
```diff
@@ -11,60 +11,129 @@ import "google/api/annotations.proto";
...
+  }
+  PreviewData preview = 3;
+}
+
+message ActionResponse {
+  uint64 code = 1;
+  string message = 2;
+}
+
+message SetActiveClientRequest { string clientDesc = 1; }
```
akosyakov, line 127: `string id = 1;`

components/supervisor-api/notification.proto
```diff
@@ -11,60 +11,129 @@ import "google/api/annotations.proto";
...
+  message PreviewData {
+    string url = 1;
+    // open the URL in a new browser tab
+    bool external = 2;
+  }
+  PreviewData preview = 3;
+}
+
+message ActionResponse {
+  uint64 code = 1;
```
akosyakov, line 123: Why do we need `code` and `message`?

components/supervisor-api/notification.proto
```diff
@@ -11,60 +11,129 @@ import "google/api/annotations.proto";
...
+  // Typically called by the IDE.
+  // Will response what client should do like `open` `preview`
+  rpc SetActiveClient(SetActiveClientRequest)
+      returns (stream SetActiveClientResponse) {
+    option (google.api.http) = {
+      post : "/v1/notification/active-client"
+    };
+  }
+
+  rpc Action(ActionRequest) returns (ActionResponse) {
```
akosyakov, line 53: Did you consider to extend `NotifyRequest` in backward compatible way instead of introducing new rpc methods?

We should make sure that both `Notify` and `Subscribe` methods support this shape.

components/supervisor-api/notification.proto
mustard-mh: We use stream because we want to ask client to do action here, so same as `Subscribe`.

If we choose to extend `NotifyRequest` we don't need this `stream` here

andreafalzetti: yeah, the idea behind `SetActiveClient` is for the client to "subscribe" as active client and by returning a stream we can then later (when an Action request arrives), ask the active client to perform the action.

components/supervisor-api/notification.proto
mustard-mh: We choose to make a new message like `ActionRequest` because `NotifyRequest` 's name is not so not fit as what it'll do

components/supervisor-api/notification.proto
mustard-mh: Consider it like http response, cmd like `code open` something's failed in some reason, we use code to check if it's ok, use message to notice users why it's not ok etc

components/supervisor-api/notification.proto
andreafalzetti: we wanted to avoid refactoring the clients to exclude these notifications from being displayed in the IDE, also in terms of supervisor, it seems cleaner to have dedicated endpoint considering it's an important feature that might also grow over time with more actions

components/supervisor-api/notification.proto
akosyakov: ok, not i'm starting getting it, so you we could have 2 new APIs like `SubscribeActive` and `NotifyActive`? The last caller of `SubscribeActive` wins and close currently active? If the client becomes inactive it simply drops `SubscribeActive`?

components/supervisor-api/notification.proto
```diff
@@ -11,60 +11,126 @@ import "google/api/annotations.proto";
...
+    option (google.api.http) = {
+      post : "/v1/notification/respond"
+    };
+  }
+
+  // SubscribeActive to inform supervisor about
+  // which is the latest client actively used by the user.
+  // Typically called by the IDE.
+  // Will response what client should do like `open` `preview`
+  rpc SubscribeActive(SubscribeActiveRequest)
```
akosyakov, line 46: Please clarify contract:
- Active means that a client is last focused.
- If a client should not respond after loosing focus then it should close the stream.
- There can be only one active client. The last one wins. For previous one the stream gets closed.

components/supervisor-api/notification.proto
```diff
@@ -11,60 +11,126 @@ import "google/api/annotations.proto";
...
+      returns (stream SubscribeActiveResponse) {
+    option (google.api.http) = {
+      post : "/v1/notification/subscribe-active"
+    };
+  }
+
+  rpc NotifyActive(NotifyActiveRequest) returns (NotifyActiveResponse) {
```
akosyakov, line 53: please provide docs for all rpc calls

components/supervisor-api/notification.proto
```diff
@@ -11,60 +11,126 @@ import "google/api/annotations.proto";
...
+  }
+
+  oneof action_data {
+    OpenData open = 1;
+    PreviewData preview = 2;
+  }
+}
+
+message NotifyActiveResponse {
+  uint64 code = 1;
```
akosyakov, line 120: it should be also oneof and correspond to kind of aciton in request.

components/supervisor-api/notification.proto
akosyakov: I think you are still affected by thiking about exec. Clients won't exec anthing, they will call internal APIs to do something. There is no code.

components/supervisor-api/notification.proto
andreafalzetti: done! 

```
// Called by the IDE to inform supervisor about
  // which is the latest client actively used by the user.
  // Will respond with a stream to which the IDE will listen
  // waiting to receive actions to run, for example: `open` or `preview`
  rpc SubscribeActive(SubscribeActiveRequest)

  // Used by gp-cli to ask supervisor to request the active client
  // to run a given command (eg. open or preview)
  rpc NotifyActive(NotifyActiveRequest) returns (NotifyActiveResponse) {

  // Used by the IDE to inform supervisor about the result (eg. success or failure)
  // of the action (eg. open or preview) requested via NotifyActive
  rpc NotifyActiveRespond(NotifyActiveRespondRequest)
  ```

cc @mustard-mh

File: components/supervisor-api/notification.proto (hunk 4 above)
Comment (mustard-mh) [reply components/supervisor-api/notification.proto:124]: No exec, code is something like status code

File: components/supervisor-api/notification.proto (hunk 6 above)
Comment (andreafalzetti) [reply components/supervisor-api/notification.proto:48]: good shout, I've extended the comment clarifying the behaviour 👌

File: components/supervisor-api/notification.proto (hunk 8 above)
Comment (andreafalzetti) [reply components/supervisor-api/notification.proto:126]: @akosyakov do you mean something like this?

```go
message NotifyActiveResponse {
  enum ActionType {
    OPEN = 0;
    PREVIEW = 1;
  }

  uint64 code = 1;
  string message = 2;
  ActionType action = 3;
}
```

cc @mustard-mh

mustard-mh: Not sure what to do with, if we do not care about result of action, it should be empty message.

If we do care about it, maybe something like this?

🅰️ 
```
message NotifyActiveResponse {
  message OpenData {
    uint64 code = 1;
    string message = 2;
  }

  message PreviewData {
    uint64 code = 1;
    string message = 2;
  }

  oneof data {
    OpenData open = 1;
    PreviewData preview = 2;
  }
}
```

🅱️ 
```
message NotifyActiveResponse {
  message OpenData {
    xxx
  }
  message PreviewData {
    xxx
  }

  uint64 code = 1;
  string message = 2;
  oneof data {
    OpenData open = 3;
    PreviewData preview = 4;
  }
}
```

I thought `code (status code)` and `message` is enough, so use one type for it

akosyakov: Honestly for now I will leave `NotifyActiveResponse` just empty. I don't existing clients propagate anything back. For error reporting should not rather propagate using grpc errors?

akosyakov (approved):
LGTM

Issue #9163, opened 2022-04-06
[Self-hosted] Allow projects to work with github.com the same way they do for GHE


jldec:
Self-hosted users cannot easily use Gitpod projects with `github.com` because there is no straightforward way to configure the Gitpod GitHub App to work with a self-hosted Gitpod instance.

> Would it be possible to enable SH Gitpod users to work with [github.com](http://github.com/) using the GHE flavor instead of always requiring a full GitHub App for any Gitpod that talks to [github.com](http://github.com/)
> 
> Yes -- we currently have a hard-coded if (host === "[github.com](http://github.com/)") that could be made smarter, e.g.
> 

```
if (host === "github.com" && hasApp) {
  useApp();
} else {
  useWebhooks();
}
```

jankeromnes:
Happy to implement this if/when we schedule it.

jldec:
Another customer is asking (see [internal](https://gitpod.slack.com/archives/C01KLC56NP7/p1649390205595319))
The alternative until we have a GH App config admin UI would be to document the [steps from here](https://www.notion.so/gitpod/How-to-deploy-a-PR-with-a-working-GitHub-App-integration-d297a1ef2f7b4b3aa8483b2ae9b47da2).
cc: @geropl

geropl:
@jldec @jankeromnes I talked to Chris about this yesterday. There are ideas to externalize all kind of "admin" config surface from the app. Those have not been populated/discussed, yet, so take this with a grain of salt.

For this issue it basically means: Let's do the minimum to make it work for self-hosted for now. And I learned yesterday that we in fact still _have_ the view that would let you configure this "on installation". Maybe that's good enough for now, and the examples @MrSimonEmms shared are mis-configured? :thinking:

jldec:
scheduled, since this unblocks self-hosted users who want to try working with github.com.
cc: @jankeromnes

rgoldfinger-quizlet:
> The alternative until we have a GH App config admin UI would be to document the [steps from here](https://www.notion.so/gitpod/How-to-deploy-a-PR-with-a-working-GitHub-App-integration-d297a1ef2f7b4b3aa8483b2ae9b47da2).

Any chance you should share these steps? It would help us get unblocked.

PR #9231, opened 2022-04-11
[Self-Hosted] Allow integrating with 'github.com' without a GitHub App

jankeromnes:
## Description

Allow integrating with 'github.com' without requiring a GitHub App. This simply reuses the same common code as GitLab, Bitbucket and GitHub Enterprise, i.e. API and webhooks.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/9163

## How to test

1. Don't follow any of the steps from https://www.notion.so/gitpod/How-to-deploy-a-PR-with-a-working-GitHub-App-integration-d297a1ef2f7b4b3aa8483b2ae9b47da2
2. You should still be able to add Projects and run Prebuilds for `github.com` repos 🎉

## Release Notes

```release-note
Allow integrating with 'github.com' without a GitHub App
```


jankeromnes:
TODO:
- [ ] Ensure `"repo"` scope for `github.com` in New Project flow (when there is no GitHub App)

jankeromnes:
Alrighty, this is now ready for review! 🎉 Please take a look. 🙏

_(Also, please disregard the merge conflict -- I had to base my PR on an older commit because preview envs got broken, but I'll rebase again before merging.)_

geropl:
Thank you @jankeromnes ! :rocket: Will have a look later today :eyes:

jankeromnes:
Friendly ping @geropl 😇 🙏

geropl:
> Friendly ping @geropl 

Thx @jankeromnes ! still have it on the radar, but might get delayed until tomorrow.

geropl:
@jankeromnes Starting now.

components/server/ee/src/prebuilds/github-service.ts
```diff
@@ -26,7 +26,8 @@ export class GitHubService extends RepositoryService {
     @inject(GithubContextParser) protected githubContextParser: GithubContextParser;
 
     async getRepositoriesForAutomatedPrebuilds(user: User): Promise<ProviderRepository[]> {
-        const repositories = (await this.githubApi.run(user, (gh) => gh.repos.listForAuthenticatedUser({}))).data;
+        const octokit = await this.githubApi.create(user);
+        const repositories = await octokit.paginate(octokit.repos.listForAuthenticatedUser, { per_page: 100 });
```
geropl, line 30: Cool to see there is a library method for this! :+1:

We could try to push our `map` function into `paginate` to be more conscious about memory usage: https://github.com/octokit/plugin-paginate-rest.js/#octokitpaginate.
Because we do not have a `filterMap`, it could return `ProviderRepository | undefined`, followed by a `filter(repo => !!)` afterwards.

The reason I'm raising this seemingly minor issue is that when looking at COU profiles, our code is riddled with uncertainties about cardinality, leading to performance drops/spikes.
Also happy to review a follow up for this.

geropl (dismissed):
Approving now despite the conflic because my test worked flawlessly, and I like the code as is!

Please ping once the conflict is resolved.

components/server/ee/src/prebuilds/github-service.ts
jankeromnes: Good point! I believe we can simply move our after-the-fact `filter` into the `octokit.paginate` call (not sure what you mean with a `filterMap`, but I think a simple `filter` might do the trick).

geropl (approved):
https://github.com/gitpod-io/gitpod/pull/9231#pullrequestreview-948528463

components/server/ee/src/prebuilds/github-service.ts
jankeromnes: Ah, whoops, the `map` would indeed have a much stronger memory reduction effect than the `filter` alone. 💡 Follow-up issue to the rescue!

PR #9233, opened 2022-04-11
[ws-manager-bridge] Fix statusVersion comparison

easyCZ:
## Description

The comparison logic was the wrong way. We'd check the statusVersion stored in the DB <= new update and exit early if it was. That's the incorrect behaviour as status.statusVersion is monotonically **increasing**

## Related Issue(s)

Fixes #

## How to test

1. Create a prebuild
2. Observe it transitions to Done (previously it would hang on Preparing)

## Release Notes

```release-note
[ws-manager-bridge] Fix statusVersion comparison.
```

## Documentation

NONE

### DIFF
--- a/components/ws-manager-bridge/ee/src/bridge.ts
+++ b/components/ws-manager-bridge/ee/src/bridge.ts
@@ -65,15 +65,13 @@ export class WorkspaceManagerBridgeEE extends WorkspaceManagerBridge {
             span.setTag("updatePrebuiltWorkspace.workspaceInstance.statusVersion", status.statusVersion);
             log.info("Found prebuild record in database.", prebuild);
 
-            if (prebuild.statusVersion <= status.statusVersion) {
-                // prebuild.statusVersion = 0 is the default value in the DB, these shouldn't be counted as stale in our metrics
-                if (prebuild.statusVersion > 0) {
-                    // We've gotten an event which is younger than one we've already processed. We shouldn't process the stale one.
-                    span.setTag("updatePrebuiltWorkspace.staleEvent", true);
-                    this.prometheusExporter.recordStalePrebuildEvent();
-                    log.info(logCtx, "Stale prebuild event received, skipping.");
-                    return;
-                }
+            // prebuild.statusVersion = 0 is the default value in the DB, these shouldn't be counted as stale in our metrics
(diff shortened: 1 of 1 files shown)


easyCZ:
Screenshot 2022-04-11 at 19 36 28
Just ran an extra test and got Ready!

components/ws-manager-bridge/ee/src/bridge.ts
```diff
@@ -65,15 +65,13 @@ export class WorkspaceManagerBridgeEE extends WorkspaceManagerBridge {
...
-                // prebuild.statusVersion = 0 is the default value in the DB, these shouldn't be counted as stale in our metrics
-                if (prebuild.statusVersion > 0) {
-                    // We've gotten an event which is younger than one we've already processed. We shouldn't process the stale one.
-                    span.setTag("updatePrebuiltWorkspace.staleEvent", true);
-                    this.prometheusExporter.recordStalePrebuildEvent();
-                    log.info(logCtx, "Stale prebuild event received, skipping.");
-                    return;
-                }
+            // prebuild.statusVersion = 0 is the default value in the DB, these shouldn't be counted as stale in our metrics
+            if (prebuild.statusVersion > 0 && prebuild.statusVersion >= status.statusVersion) {
```
easyCZ, line 69: This is the main change. We skip any records which have `prebuild.statusVersion == 0` (this comes from the db) because 0 is the default DB value. And we then check that prebuild status is greater than our new event, so that we  bail early.

jankeromnes (approved):
Logic change looks good to me, and trusting your test run. Many thanks for the quick fix!

components/ws-manager-bridge/ee/src/bridge.ts
jankeromnes: But, just to make sure I understand this properly, `status.statusVersion` can never be `0`, right?

easyCZ: I believe that's correct.

If it were, the following would be executed:
The only way we get to a comparison of `prebuild db record` and `event` is if `prebuild.statusVersion > 0`. 

If this is true, than `prebuild.statusVersion` must be at least 1, or greater. In this case, that means that we must have already processed an event once to change it from 0, to 1+ (0 is the DB default value, set when the Prebuild record is initially created by the server). In this case, that means that we're good even if `status.statusVersion` can be 0.

PR #9238, opened 2022-04-11
[ws-manager] Introduce workspace classes

csweichel:
## Description
This PR adds support for workspace classes. Workspace classes add per-workspace support for different resource requests and limits, as well as different templates.

## Related Issue(s)

works on #8261 

## How to test
See unit tests

## Release Notes

```release-note
[ws-manager] Add support for workspace classes
```

PR #9242, opened 2022-04-12
Add persistent volume support for workspaces

### LINKED ISSUE #9142: ws-manager: create PVC before creating workspace pod and specify it in pod spec mounted at /workspace

### LINKED ISSUE #9442: add support for restoring workspace from backup that was stored in GCS when using PVC
This will allow smooth transition from non PVC approach to PVC approach.
Implement this by injecting a layer in registry facade that would download and untar backup.

sagor999:
## Description

Add support for persistent volume claim use for workspaces instead of using local storage.
Usage is gated behind "persistent volume claim" feature flag.

## Related Issue(s)

Fixes #9142
Fixed #9442 

## How to test

Spin up new workspace preview environment. 
go into admin panel and enable feature flag
start new workspace. it will start using dedicated persistent volume.
observe new pvc created (`kubectl get pvc -A`)
delete workspace. pvc will be deleted.

image build is using old codepath and not affected by this change.
prebuilds are using old codepath and are not affected by this change.

limitations:
current implementation doesn't support restoring from previous backup. That will be added separately. 

## Release Notes

```release-note
[experimental] Add persistent volume support for workspaces
```


install/installer/pkg/config/v1/config.go
```diff
@@ -236,6 +238,12 @@ type Workspace struct {
 	Resources Resources           `json:"resources" validate:"required"`
 	Templates *WorkspaceTemplates `json:"templates,omitempty"`
 
+	// PVCSize is a size of persistent volume claim to use when creating workspace
+	PVCSize string `json:"pvcSize" validate:"required"`
```
MrSimonEmms, line 242: I think it would be worth having `pvc` as an object

```yaml
workspace:
  pvc:
    size: 30Gi
    storageClass: csi-gce-pd
```

MrSimonEmms (commented):
@sagor999 can you update the `example-config.yaml` to the include the changes please - `go run . init > example-config.yaml`?

Also, are there any situations where you can imagine that having a PVC would be undesirable? We currently have PVCs for `docker-registry` and `minio` (in-cluster), but we have had a couple of people in the past who have said that they can't use a PVC in their installation (admittedly, not paying customers). With those scenarios, there's a fairly established pattern of getting around it (run the dependency as a Helm chart on your machine and link through as an "external" dependency) but with a workspace PVC, we won't be able to do that.

Would a `workspace.pvcEnabled` boolean be worthwhile for wider support?

sagor999:
@MrSimonEmms updated and made suggested fix. Thank you!

As for PVC: once we (workspace) switches to PVC completely, old way of doing things will be removed from codebase. Reason for this is that it will be extremely tedious and error prone to maintain both code paths. 

What is the reason for not being able to use PVC? Since you can use PVC with "local" storage as well (in case if customer doesn't have access to remote storage or would like to use local storage for added speed).
We will have a requirement for supporting Snapshot CRD though, as workspaces will be backed up and restored through the use of snapshotter (k8s snapshot term, not gitpod snapshot).

MrSimonEmms:
> @MrSimonEmms updated and made suggested fix. Thank you!

Excellent. Thanks.
 
> As for PVC...

I don't necessarily have a use-case against it, but I'm just thinking about all the self-hosted/non-GCP users we have. It seems that the `csi-gce-pd` storage class is specific to GCP - I'm not that happy about putting what we need as the default value and having to get everyone else to remember to change it. What I'd much prefer to be able to do is to put a storage class that works generically and then for us to change to the `csi-gce-pd` class in our deployment script.

Looking at the [documentation](https://kubernetes.io/docs/concepts/storage/storage-classes/), it seems most (all?) of them are tied to a cloud provider in some way. Are we able to let this be blank by default as is done in our [Registry chart](https://github.com/twuni/docker-registry.helm/blob/main/templates/pvc.yaml#L19)?

sagor999:
@MrSimonEmms you are :100:% correct. That was oversight on my part. Removed `csi-gce-pd`

MrSimonEmms:
Thanks @sagor999. It's definitely is tricky making things easy for SaaS, enterprise and hobbyist self-hosted

components/content-service/pkg/layer/provider.go
```diff
@@ -227,6 +227,116 @@ func (s *Provider) GetContentLayer(ctx context.Context, owner, workspaceID strin
...
+	manifest, _, err = s.downloadContentManifest(ctx, bucket, mfobj)
+	if err != nil && err != storage.ErrNotFound {
+		return nil, nil, err
+	}
+	if manifest != nil {
+		span.LogKV("backup found", "full workspace backup")
+
+		l, err = s.layerFromContentManifestPVC(ctx, manifest, csapi.WorkspaceInitFromBackup, true)
+		return l, manifest, err
+	}
```
csweichel, line 260: We probably want to keep the FWB and PVC paths separate for now.
Please remove this.

components/content-service/pkg/layer/provider.go
```diff
@@ -368,6 +508,17 @@ func contentDescriptorToLayer(cdesc []byte) (*Layer, error) {
...
+// version of this function for persistent volume claim feature
+// we cannot use /workspace folder as when mounting /workspace folder through PVC
+// it will mask anything that was in container layer, hence we are using /.workspace instead here
+func contentDescriptorToLayerPVC(cdesc []byte) (*Layer, error) {
+	return layerFromContent(
+		fileInLayer{&tar.Header{Typeflag: tar.TypeDir, Name: "/.workspace", Uid: initializer.GitpodUID, Gid: initializer.GitpodGID, Mode: 0755}, nil},
+		fileInLayer{&tar.Header{Typeflag: tar.TypeDir, Name: "/.workspace/.gitpod", Uid: initializer.GitpodUID, Gid: initializer.GitpodGID, Mode: 0755}, nil},
+		fileInLayer{&tar.Header{Typeflag: tar.TypeReg, Name: "/.workspace/.gitpod/content.json", Uid: initializer.GitpodUID, Gid: initializer.GitpodGID, Mode: 0755, Size: int64(len(cdesc))}, cdesc},
+	)
+}
```
csweichel, line 520: Considering that we're not using FWB anywhere, we could harmonise this path and make `/.workspace` the default for those files. Removes the duplication herein and and in supervisor. Same goes for `workspaceReadyLayerPVC`.

components/workspacekit/cmd/rings.go
```diff
@@ -112,6 +112,7 @@ var ring0Cmd = &cobra.Command{
 		cmd.Env = append(os.Environ(),
 			"WORKSPACEKIT_FSSHIFT="+prep.FsShift.String(),
 			fmt.Sprintf("WORKSPACEKIT_FULL_WORKSPACE_BACKUP=%v", prep.FullWorkspaceBackup),
+			fmt.Sprintf("WORKSPACEKIT_PERSISTENT_VOLUME_CLAIM=%v", prep.PersistentVolumeClaim),
```
csweichel, line 115:
```suggestion
			fmt.Sprintf("WORKSPACEKIT_NO_WORKSPACE_MOUNT=%v", prep.FullWorkspaceBackup || prep.PersistentVolumeClaim),
```

components/workspacekit/cmd/rings.go
```diff
@@ -303,7 +304,8 @@ var ring1Cmd = &cobra.Command{
 
 		// FWB workspaces do not require mounting /workspace
 		// if that is done, the backup will not contain any change in the directory
-		if os.Getenv("WORKSPACEKIT_FULL_WORKSPACE_BACKUP") != "true" {
+		// same applies to persistent volume claims, we cannot mount /workspace folder when PVC is used
+		if os.Getenv("WORKSPACEKIT_FULL_WORKSPACE_BACKUP") != "true" && os.Getenv("WORKSPACEKIT_PERSISTENT_VOLUME_CLAIM") != "true" {
```
csweichel, line 308:
```suggestion
		if os.Getenv("WORKSPACEKIT_NO_WORKSPACE_MOUNT") != "true" {
```

components/ws-daemon/pkg/content/service.go
```diff
@@ -165,6 +165,10 @@ func (s *WorkspaceService) InitWorkspace(ctx context.Context, req *api.InitWorks
 		if err != nil {
 			return nil, status.Errorf(codes.InvalidArgument, "invalid content manifest: %s", err.Error())
 		}
+		if req.PersistentVolumeClaim {
+			// pavel: setting wsloc as otherwise mkdir fails later on.
```
csweichel, line 169: nit

```suggestion
			// TODO(pavel): setting wsloc as otherwise mkdir fails later on.
```

see https://www.notion.so/gitpod/How-we-develop-e8ce7e562478458a9223eb9b797415b2#86e669f8e7df4f1289c9bede0855b5f9

components/ws-daemon/pkg/content/service.go
```diff
@@ -153,7 +153,7 @@ func (s *WorkspaceService) InitWorkspace(ctx context.Context, req *api.InitWorks
 	var (
 		wsloc string
 	)
-	if req.FullWorkspaceBackup {
+	if req.FullWorkspaceBackup || req.PersistentVolumeClaim {
 		if s.runtime == nil {
 			return nil, status.Errorf(codes.FailedPrecondition, "full workspace backup is not available - not connected to container runtime")
 		}
```
csweichel, line 159: I know this wasn't introduced in this PR, but this hasn't made sense in a while; even less so.

Neither FWB nor PVC require this connection anymore. Please remove.

components/ws-daemon/pkg/content/service.go
```diff
@@ -386,6 +392,12 @@ func (s *WorkspaceService) uploadWorkspaceContent(ctx context.Context, sess *ses
 		mf   csapi.WorkspaceContentManifest
 	)
 
+	if sess.PersistentVolumeClaim {
+		// currently not supported (will be done differently via snapshots)
+		log.Error("uploadWorkspaceContent not supported yet when PVC feature is enabled")
+		return nil
```
csweichel, line 398:
```suggestion
		return nil, status.Error(codes.FailedPrecondition, "uploadWorkspaceContent not supported yet when PVC feature is enabled")
```

components/ws-manager-api/go/config/config.go
```diff
@@ -118,6 +118,7 @@ type Configuration struct {
 // AllContainerConfiguration contains the configuration for all container in a workspace pod
 type AllContainerConfiguration struct {
 	Workspace ContainerConfiguration `json:"workspace"`
+	PVC       PVCConfiguration       `json:"pvc"`
```
csweichel, line 121: Not sure this is the right place. I'd move this to `Configuration` for now.
With the introduction of [workspace classes](https://github.com/gitpod-io/gitpod/pull/9238) the `AllContainerConfiguration` struct goes away.

components/ws-manager/pkg/manager/annotations.go
```diff
@@ -63,6 +63,9 @@ const (
 	// fullWorkspaceBackupAnnotation is set on workspaces which operate using a full workspace backup
 	fullWorkspaceBackupAnnotation = "gitpod/fullWorkspaceBackup"
 
+	// pvcWorkspaceFeatureAnnotation is set on workspaces which are using persistent_volume_claim feature
+	pvcWorkspaceFeatureAnnotation = "gitpod/pvcFeature"
```
csweichel, line 67: (newer annotations use the more "correct" domain-style naming)

```suggestion
	pvcWorkspaceFeatureAnnotation = "gitpod.io/pvcFeature"
```

components/ws-manager/pkg/manager/create.go
```diff
@@ -212,6 +213,38 @@ func mergeProbe(dst, src reflect.Value) (err error) {
 	return nil
 }
 
+func (m *Manager) createPVCForWorkspacePod(startContext *startWorkspaceContext) (*corev1.PersistentVolumeClaim, error) {
```
csweichel, line 216: We'll want some basic test coverage of this. Introducing a fixture test akin to the [CDWP tests](https://github.com/gitpod-io/gitpod/blob/dcf82f628eac7c380d0677b5c3f4485e293ee560/components/ws-manager/pkg/manager/create_test.go#L22) should be straight forward.

(doesn't necessarily have to be in this PR)

components/ws-manager/pkg/manager/create.go
```diff
@@ -212,6 +213,38 @@ func mergeProbe(dst, src reflect.Value) (err error) {
...
+			Name:      fmt.Sprintf("%s-%s", prefix, req.Id),
+			Namespace: m.Config.Namespace,
+		},
+		Spec: corev1.PersistentVolumeClaimSpec{
+			AccessModes:      []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce},
+			StorageClassName: &storageClassName,
+			Resources: corev1.ResourceRequirements{
+				Requests: corev1.ResourceList{
+					// todo: fix pvc size, should come from ws manager config now
+					corev1.ResourceName(corev1.ResourceStorage): resource.MustParse(m.Config.Container.PVC.Size),
```
csweichel, line 241: `MustParse` would panic. If the `PVC.Size` type were `resource.Quantity` already this parse operation would not be necessary.

`resource.Quantity` implements `UnmarshalJSON`.

components/ws-manager/pkg/manager/create.go
```diff
@@ -483,6 +518,26 @@ func (m *Manager) createDefiniteWorkspacePod(startContext *startWorkspaceContext
...
+					ClaimName: pvcName,
+				},
+			}
+
+			// SubPath so that lost+found is not visible
+			pod.Spec.Containers[0].VolumeMounts[0].SubPath = "workspace"
+			// not needed, since it is using dedicated disk
+			pod.Spec.Containers[0].VolumeMounts[0].MountPropagation = nil
+
+			// pavel: magical ID to make sure that gitpod user inside the workspace can write into /workspace folder mounted by PVC
```
csweichel, line 537:
```suggestion
			// pavel: 133332 is the Gitpod UID (33333) shifted by 99999. The shift happens inside the workspace container due to the user namespace use. We set this magical ID to make sure that gitpod user inside the workspace can write into /workspace folder mounted by PVC
```

components/ws-manager/pkg/manager/create.go
```diff
@@ -483,6 +518,26 @@ func (m *Manager) createDefiniteWorkspacePod(startContext *startWorkspaceContext
...
+
+			// update volume to use persistent volume claim, and name of it is the same as pod's name
+			pvcName := pod.ObjectMeta.Name
+			pod.Spec.Volumes[0].VolumeSource = corev1.VolumeSource{
+				PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+					ClaimName: pvcName,
+				},
+			}
+
+			// SubPath so that lost+found is not visible
```
csweichel, line 532: ❤️

components/ws-manager/pkg/manager/create.go
```diff
@@ -569,6 +624,20 @@ func (m *Manager) createWorkspaceContainer(startContext *startWorkspaceContext)
...
+			MountPath:        workspaceDir,
+			ReadOnly:         false,
+			MountPropagation: &mountPropagation,
+		},
+		{
+			MountPath:        "/.workspace",
+			Name:             "daemon-mount",
+			MountPropagation: &mountPropagation,
+		},
+	}
```
csweichel, line 639: nit: why move this to a variable beforehand and not leave it inline?

components/ws-manager/pkg/manager/manager.go
```diff
@@ -200,6 +200,26 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
...
+			err = m.Clientset.Create(ctx, pvc)
+			if err != nil {
+				return nil, xerrors.Errorf("cannot create pvc object for workspace pod: %w", err)
+			}
+			created = true
+		}
+		if created {
+			break
+		}
+	}
```
csweichel, line 221: nit: (can be simplified a bit)
```suggestion
    var createPVC bool
	for _, feature := range startContext.Request.Spec.FeatureFlags {
	    if feature == api.WorkspaceFeatureFlag_PERSISTENT_VOLUME_CLAIM {
	        createPVC = true
	        break
	    }
	}
		if createPVC {
			clog.Info("PVC feature detected, creating PVC object")
			pvc, err := m.createPVCForWorkspacePod(startContext)
			if err != nil {
				return nil, xerrors.Errorf("cannot create pvc for workspace pod: %w", err)
			}
			err = m.Clientset.Create(ctx, pvc)
			if err != nil {
				return nil, xerrors.Errorf("cannot create pvc object for workspace pod: %w", err)
			}
		}
```

components/ws-manager/pkg/manager/monitor.go
```diff
@@ -856,6 +859,30 @@ func (m *Monitor) finalizeWorkspaceContent(ctx context.Context, wso *workspaceOb
 			return false, nil, nil
 		}
 
+		// todo: once we add snapshot objects, this will be moved to a better place
```
csweichel, line 862:
```suggestion
		// TODO(pavel): once we add snapshot objects, this will be moved to a better place
```

components/content-service/pkg/layer/provider.go
sagor999: I agree! I thought of doing that change, but was concerned about some unknown side effect to regular workspaces, so instead concentrated on adding PVC path that would not affect normal workflow right now. 
I added an issue to make this happen as part of separate PR though: https://github.com/gitpod-io/gitpod/issues/9496

components/ws-manager/pkg/manager/create.go
sagor999: bad refactor on my part, was cleaning up after all my changes and forgot to undo this one.

components/ws-manager/pkg/manager/create.go
sagor999: :100: 
I think once everything is working (and code is no longer changing with every commit that I make :sweat_smile: ), I will add all missing tests (to make sure I don't forget I added a sub section into epic regarding this)

sagor999:
@csweichel @aledbf @Furisto @gitpod-io/engineering-self-hosted @gitpod-io/engineering-ide 
Please review. This PR has been sitting ready for review for over 10 days now, and it takes a lot of my time on keeping it up to date with all the changes that are coming in. So would greatly appreciate :eyes: on this PR. 
If there are any issues, please raise them, and I will take care of them right away.

@kylos101 for visibility.

utam0k:
I tried it in the preview environment, but it stopped at the Preparing workspace and did not proceed.
PVC was pending all the time.
```console
staging-pavel-9142-pvc2        ws-db9ba9eb-2670-4f1e-9d68-a17e45816063   Pending                                                                                       10m
```

components/ws-manager/pkg/manager/imagespec.go
```diff
@@ -70,7 +74,12 @@ func (m *Manager) GetImageSpec(ctx context.Context, req *regapi.GetImageSpecRequ
 		if err != nil {
 			return nil, xerrors.Errorf("cannot unmarshal init config: %w", err)
 		}
-		cl, _, err := m.Content.GetContentLayer(ctx, owner, workspaceID, &initializer)
+		var cl []layer.Layer
+		if pvcFeatureEnabled {
+			cl, _, err = m.Content.GetContentLayerPVC(ctx, owner, workspaceID, &initializer)
+		} else {
+			cl, _, err = m.Content.GetContentLayer(ctx, owner, workspaceID, &initializer)
+		}
```
utam0k, line 82: How about raising the abstraction level of the Content layer? In other words, how about installing an interface like ContentLayerProvider, and implementing `Get()` for both the conventional conventional method and the PVC Content method? Is this a bit much?

components/content-service/pkg/layer/provider.go
```diff
@@ -227,6 +227,105 @@ func (s *Provider) GetContentLayer(ctx context.Context, owner, workspaceID strin
 	return nil, nil, xerrors.Errorf("no backup or valid initializer present")
 }
 
+// GetContentLayerPVC provides the content layer for a workspace that uses PVC feature
+func (s *Provider) GetContentLayerPVC(ctx context.Context, owner, workspaceID string, initializer *csapi.WorkspaceInitializer) (l []Layer, manifest *csapi.WorkspaceContentManifest, err error) {
+	span, ctx := tracing.FromContext(ctx, "GetContentLayer")
```
utam0k, line 232: Perhaps this is more correct?
```suggestion
	span, ctx := tracing.FromContext(ctx, "GetContentLayerPVC")
```

components/content-service/pkg/layer/provider.go
```diff
@@ -227,6 +227,105 @@ func (s *Provider) GetContentLayer(ctx context.Context, owner, workspaceID strin
...
+			manifest = &csapi.WorkspaceContentManifest{
+				Type: csapi.TypeFullWorkspaceContentV1,
+			}
+		}
+	}()
+
+	// check if workspace has an FWB
+	var (
+		bucket = s.Storage.Bucket(owner)
+	)
```
utam0k, line 248:
```suggestion
    bucket := s.Storage.Bucket(owner)
```

akosyakov (approved):
I did not try, but I looked at supervisor changes and they look backward compatible to how we check for content readiness.

components/ws-daemon/pkg/content/service.go
```diff
@@ -153,10 +153,7 @@ func (s *WorkspaceService) InitWorkspace(ctx context.Context, req *api.InitWorks
 	var (
 		wsloc string
 	)
-	if req.FullWorkspaceBackup {
-		if s.runtime == nil {
-			return nil, status.Errorf(codes.FailedPrecondition, "full workspace backup is not available - not connected to container runtime")
-		}
+	if req.FullWorkspaceBackup || req.PersistentVolumeClaim {
 		var mf csapi.WorkspaceContentManifest
```
csweichel, line 157: PVC workspaces have no content manifest - I'd be surprised if this code worked.

csweichel (approved):
LGTM - thank you for incorporating the feedback.

In a follow-up PR we need to have tests CDWP and status tests for the PVC workspaces.

components/ws-daemon-api/daemon.proto
```diff
@@ -55,6 +55,11 @@ message InitWorkspaceRequest {
 
     // storage_quota_bytes enforces a storage quate for the workspace if set to a value != 0
     int64 storage_quota_bytes = 8;
+
+    // persistent_volume_claim means that we use PVC instead of HostPath to mount /workspace folder and content-init
+    // happens inside workspacekit instead of in ws-daemon. We also use k8s Snapshots to store\restore workspace content
+    // instead of GCS\tar.
+    bool persistent_volume_claim = 9;
```
utam0k, line 62: How about using enumerations instead of using bool?
https://developers.google.com/protocol-buffers/docs/proto3#enum

components/supervisor/pkg/supervisor/supervisor.go
```diff
@@ -1221,6 +1221,12 @@ func startContentInit(ctx context.Context, cfg *Config, wg *sync.WaitGroup, cst
 	}()
 
 	fn := "/workspace/.gitpod/content.json"
+	fnReady := "/workspace/.gitpod/ready"
+	if _, err := os.Stat("/.workspace/.gitpod/content.json"); !os.IsNotExist(err) {
+		fn = "/.workspace/.gitpod/content.json"
```
utam0k, line 1226: To make it explicit that these paths come from a PVC mount, why not make it refer to a variable shared with the following location? This would surely help the code reader.
https://github.com/gitpod-io/gitpod/pull/9242/files#diff-46a025104f556cc7faed4319cb1c9d2bd0b5097231b42707825aaf0df24d7337R477

Furisto:
I tried it [here](https://fo-pvc.preview.gitpod-dev.com/workspaces), but the workspace never gets ready. `timed out waiting for the condition`

sagor999:
@Furisto @utam0k thank you for reviews! For preview environment, please use this branch:
https://github.com/gitpod-io/workspace-preview/pull/33
As it setups storage class to use for PVC (without it k8s doesn't know which storage class to use, and that's why PVC is stuck in pending)

components/ws-daemon-api/daemon.proto
sagor999: Not sure if this is applicable here, since this bool flag is used as feature flag. Once PVC feature is stable and working, we will remove old code and remove this feature flag and this new code will become default code path.

components/supervisor/pkg/supervisor/supervisor.go
sagor999: I think we will be switching default to always use `/.workspace` folder, so that we don't need to do all those checks in the future

Furisto (approved):

components/ws-daemon/pkg/content/service.go
sagor999: Indeed it doesn't. I will remove it in a follow up PR (which is currently in draft state)

jenting:
> And you cannot use PVC feature just yet in our dev env, as platform team would need to setup storageclass\snapshot CRD in there first. I will be reaching out to them regarding this once this feature (PVC support) is a bit more feature complete.

I thought the core-dev have the pre-builtin CRD already.
image

components/content-service/pkg/layer/provider.go
```diff
@@ -227,6 +227,103 @@ func (s *Provider) GetContentLayer(ctx context.Context, owner, workspaceID strin
 	return nil, nil, xerrors.Errorf("no backup or valid initializer present")
 }
 
+// GetContentLayerPVC provides the content layer for a workspace that uses PVC feature
+func (s *Provider) GetContentLayerPVC(ctx context.Context, owner, workspaceID string, initializer *csapi.WorkspaceInitializer) (l []Layer, manifest *csapi.WorkspaceContentManifest, err error) {
```
jenting, line 231: NOTE: we could refactor these functions `GetContentLayer` and `GetContentLayerPVC` into one in the future.

jenting (commented):
Questions:
- What if the PVC storage full, will the workspace hang? Or will it auto increase the PVC size? (I guess no)?
- If I understand correctly, once the workspace stop, it'd still backup the content to the remote like S3, correct?

MrSimonEmms (approved):

sagor999:
@jenting 

>     * What if the PVC storage full, will the workspace hang? Or will it auto increase the PVC size? (I guess no)?

It depends on workspace, but it might experience issues. It is the same behavior as pre PVC though (since workspaces have same 30GB storage quota set on them). And we wouldn't auto increase PVC size.

>     * If I understand correctly, once the workspace stop, it'd still backup the content to the remote like S3, correct?

When workspace stops, we will create snapshot volume out of PVC and that will be our backup. (part of a different PR). We are moving away from S3\Object storage.

components/ws-manager/config-schema.json
```diff
@@ -126,6 +131,22 @@
...
+    "PVCConfiguration": {
+      "required": [
+        "size",
+        "storage-class"
+      ],
+      "properties": {
+        "size": {
+          "type": "string"
+        },
+        "storage-class": {
```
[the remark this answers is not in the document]
sagor999: Good point! :heart: 
Will address in a follow up PR.

components/ws-daemon/pkg/content/service.go
```diff
@@ -165,6 +162,10 @@ func (s *WorkspaceService) InitWorkspace(ctx context.Context, req *api.InitWorks
 		if err != nil {
 			return nil, status.Errorf(codes.InvalidArgument, "invalid content manifest: %s", err.Error())
 		}
+		if req.PersistentVolumeClaim {
```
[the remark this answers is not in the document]
sagor999: Already fixed in a follow up PR! Thank you!

components/content-service/pkg/layer/provider.go
sagor999: Indeed. Once PVC is stable and we switched over to it, old code will be removed. (so that we will have only one version of that function)

sagor999:
@aledbf just need your approval.

aledbf (approved):

jenting:
> When workspace stops, we will create snapshot volume out of PVC and that will be our backup. (part of a different PR). We are moving away from S3\Object storage.

If I understand correct, create a snapshot volume is leverage the Kubernetes snapshot controller + volumesnapshot/volumesnapshotcontent/volumesnapshotclass CRDs. However, it's the in-cluster CRDs and the volume snapshot could be stored in the external storage disk, then how can we handle a case that we launch a new cluster and shift the traffic to new cluster? If means that the old and new cluster _must_ uses the same external storage, but it also means that the external storage needs to support mounts by multiple node/zone/...

sagor999:
>However, it's the in-cluster CRDs and the volume snapshot could be stored in the external storage disk

Not quite. We are using GCP's snapshot feature for this. So when we create volumesnapshot, it actually creates GCP`s snapshot of the disk from which PVC was mounted from.
So snapshots live outside of cluster lifecycle.
When we bring up new cluster, we create snapshot volume object referencing the one from GCP. 
And also due to that we can mount snapshot from one region to another region (though incurring trans-region cost).

jenting:
> > However, it's the in-cluster CRDs and the volume snapshot could be stored in the external storage disk
> 
> Not quite. We are using GCP's snapshot feature for this. So when we create volumesnapshot, it actually creates GCP`s snapshot of the disk from which PVC was mounted from. So snapshots live outside of cluster lifecycle. When we bring up new cluster, we create snapshot volume object referencing the one from GCP. And also due to that we can mount snapshot from one region to another region (though incurring trans-region cost).

Awesome 💯 , then we must make sure the self-hosted works as well 🙏

components/ws-manager/config-schema.json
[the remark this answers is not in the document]
jenting: Address in #9806.

components/ws-daemon/pkg/content/service.go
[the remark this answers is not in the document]
jenting: #9475

components/ws-manager/pkg/manager/monitor.go
```diff
@@ -867,6 +870,30 @@ func (m *Monitor) finalizeWorkspaceContent(ctx context.Context, wso *workspaceOb
...
+				// todo: once we add snapshot objects, this will be changed to create snapshot object first
+				pvcErr := m.manager.Clientset.Delete(ctx,
+					&corev1.PersistentVolumeClaim{
+						ObjectMeta: metav1.ObjectMeta{
+							Name:      pvcName,
+							Namespace: m.manager.Config.Namespace,
+						},
+					},
+				)
+				span.LogKV("event", "pvc deleted")
```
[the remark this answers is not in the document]
jenting: #9475

PR #9266, opened 2022-04-12
[ws-daemon] Enable workspace IP forwarding

csweichel:
## Description
To make the new nftables setup work.
It worked on core-dev because there IP forwarding is globally enabled.

## How to test
Run on an ephemeral cluster.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/ws-daemon/nsinsider/main.go
+++ b/components/ws-daemon/nsinsider/main.go
@@ -464,6 +464,13 @@ func main() {
 					return nil
 				},
 			},
+			{
+				Name:  "enable-ip-forward",
+				Usage: "enable IPv4 forwarding",
+				Action: func(c *cli.Context) error {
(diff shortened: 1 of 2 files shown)


aledbf (approved):

utam0k:
@aledbf @csweichel Thanks for taking care of it. I have a question. When I introduce it, I checked this and figured out the below setting so I didn't write this PR function. Is a different image used than this one?
https://github.com/gitpod-io/gitpod-packer-gcp-image/blob/f033009d070bf2d4b180153744d43201b4c64f7b/linux/sysctl/99-gce.conf#L25

aledbf:
@utam0k yes, that's the image and the `sysctl` is correct. That said, the value we see in the pod, by default, is 0.

PR #9290, opened 2022-04-13
[public-api-server] Simple gRPC server

easyCZ:
## Description

Extends public-api-server with a simple gRPC server. The server doesn't have any services any starts up. This is intended to move us along and allow us to iterate, rather than be a complete and final result.

Graceful shutdown of servers is outside of the scope of this PR and will be tackled in subsequent ones.

## Related Issue(s)

Part of https://github.com/gitpod-io/gitpod/issues/9229

## How to test

`cd components/public-api-server && go run main.go`
```
Serving HTTP :9000
Serving gRPC [::]:9001
^CReceived termination signal, shutting down.
```

## Release Notes

```release-note
NONE
```

## Documentation

NONE


components/public-api-server/main.go
```diff
@@ -6,15 +6,66 @@ package main
...
-	}))
+	errors := make(chan error)
+	defer close(errors)
+
+	go func() {
+		addr := ":9000"
+		fmt.Println("Serving HTTP", addr)
+		err := http.ListenAndServe(addr, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			_, _ = w.Write([]byte(`hello world\n`))
+		}))
```
andrew-farries, line 27: Do we still need this http server in addition to the GRPC server below?

components/public-api-server/main.go
```diff
@@ -6,15 +6,66 @@ package main
...
+		if err != nil {
+			fmt.Println("Failed to server on", addr, err)
```
andrew-farries, line 29: 'failed to start'

andrew-farries (changes requested):

components/public-api-server/main.go
easyCZ: Yes. We'll need to serve metrics, ready/health endpoints etc

andrew-farries (approved):

components/public-api-server/main.go
easyCZ: pushed a fix, thanks!

andrew-farries (approved):

Issue #4050, opened 2021-04-23
Improve responsive layout for the dashboard


gtsiolis:
### Problem to solve

Although the new dashboard layout was not optimized for being responsive on smaller viewports, layout is relatively acessible for medium and large viewports.

Still, there are some minor issues that are visible also on medium viewports, see https://github.com/gitpod-io/gitpod/issues/3859 and https://github.com/gitpod-io/gitpod/issues/3829.

FWIW, we're currently introducing a scrollable modal in https://github.com/gitpod-io/gitpod/pull/3995 which could be useful to componetization of the modal component.

### Proposal

MVC steps here could be the following:
- [ ] Resolve the scrollbar issues on Plans and Integrations pages, https://github.com/gitpod-io/gitpod/issues/3859.
- [ ] Resolve the accessibility issues with the scrolling container in the modals, https://github.com/gitpod-io/gitpod/issues/3829.
- [ ] Design and implement a responsive layout for component like settings sidebar, workspaces list, top navigation, etc.

jankeromnes:
I would absolutely _love_ to add minimum viable responsiveness to our dashboard. This unfinished business has been gnawing at me for months now.

I can time-box this e.g. to one or two days if that's helpful.

pri1311:
The responsiveness issue also arises for the edit permissions modal. I didn't see this specifically mentioned anywhere, hence adding it here. Have to resize the window to 90% to see the update permissions button.
![image](https://user-images.githubusercontent.com/64613009/144757194-98af1185-947b-4b29-b174-e60af11e117b.png)

gtsiolis:
Thanks @pri1311! I've reopened https://github.com/gitpod-io/gitpod/issues/3829 for this.

jankeromnes:
Not stale 😇

PR #9293, opened 2022-04-13
[dashboard] add notice for stable back and change latest desc

mustard-mh:
## Description

add notice for stable back and change latest desc

|IDE Version|Dark|Light|
|:-:|:-:|:-:|
|stable|image|image|
|latest| image |image|

## Related Issue(s)

Fixes #

## How to test

1. Open a workspace with stable IDE, to see if have warning for change Preferences
2. Open a workspace with latest IDE, to see if warning changed to `This workspace is configured with the latest release (unstable) for the editor`

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/dashboard/src/start/StartWorkspace.tsx
+++ b/components/dashboard/src/start/StartWorkspace.tsx
@@ -28,6 +28,7 @@ import { watchHeadlessLogs } from "../components/PrebuildLogs";
 import { getGitpodService, gitpodHostUrl } from "../service/service";
 import { StartPage, StartPhase, StartWorkspaceError } from "./StartPage";
 import ConnectToSSHModal from "../workspaces/ConnectToSSHModal";
+import Alert from "../components/Alert";
 const sessionId = v4();
 
 const WorkspaceLogs = React.lazy(() => import("../components/WorkspaceLogs"));
@@ -391,6 +392,7 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
         let title = undefined;
         let statusMessage = !!error ? undefined : <p className="text-base text-gray-400">Preparing workspace …</p>;
         const contextURL = ContextURL.getNormalizedURL(this.state.workspace)?.toString();
+        const useLatest = !!this.state.workspaceInstance?.configuration?.ideConfig?.useLatest;
 
         switch (this.state?.workspaceInstance?.status.phase) {
             // unknown indicates an issue within the system in that it cannot determine the actual phase of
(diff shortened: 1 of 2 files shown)


mustard-mh:
@gtsiolis How about use `Alert` component with `info` type at the same position with latest warning?

gtsiolis:
> How about use Alert component with info type at the same position with latest warning?

Sounds great, @mustard-mh! This was also the intent initially, see [relevant discussion](https://gitpod.slack.com/archives/C01KGM9BH54/p1649842179382579?thread_ts=1649839540.277159&cid=C01KGM9BH54) (internal) and linked review comment in https://github.com/gitpod-io/gitpod/pull/8783#discussion_r828522647.

mustard-mh:
Yeah, I missed it, thx @gtsiolis

mustard-mh:
@gtsiolis Preview environment is ok now, you can have a look. I also capture images to PR description

And we have 

1. margin-top too big for small screen (Macbook Pro 14')
2. After click drag down button, we need to scroll down to see all options
3. Option **Stop workspace** maybe need to change text color and (move it to bottom?), I often accidentally click close

Maybe we should file issues to solve them

image

gtsiolis:
Looking at this now! 👀

components/dashboard/src/start/StartWorkspace.tsx
```diff
@@ -524,6 +526,19 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
                                     {openLinkLabel}
                                 </button>
                             </div>
+                            {!useLatest && (
+                                <Alert type="info" className="mt-4 w-96">
```
gtsiolis, line 530: **thought**:  This is out of the scope of this PR but in the future we could probably introduce better breakpoints in general within the dashboard so that a) components like this expand with the container width and b) do not need to use a fixed width attribute.

However, that's minor for now and the effort/impact of doing this sounds low, but probably something to consider when we pick up https://github.com/gitpod-io/gitpod/issues/4050 for improving the responsive layout of the dashboard. Cc @jankeromnes as we've talked about this a couple of times.

components/dashboard/src/start/StartWorkspace.tsx
```diff
@@ -524,6 +526,19 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
...
+                                    You can change the default editor for opening workspaces in{" "}
```
gtsiolis, line 531: **thought**: This alert could become obsolete or less useful so we could remove it in the future if the actions were clearer like in https://github.com/gitpod-io/gitpod/issues/6910 or more flexiible as seen in the design exploration mentioned in https://github.com/gitpod-io/gitpod/pull/8783#discussion_r828522647. Re-posting the design exploration for visibility. Feedback is always welcome!

dropdown

gtsiolis (approved):
UX changes look good, @mustard-mh! 🔮 

Took me longer than expected to get back to this because lunch! 🌮 🌮 🌮 🌮 🌮 

Left below some non-blocking comments for visibility and opened a follow up issue in https://github.com/gitpod-io/gitpod/issues/9326 for the overflowing dropdown mentioned in https://github.com/gitpod-io/gitpod/pull/9293#issuecomment-1097985605 which has been bothering me for a while. 

Regarding, the move of the action to stop a workspace below, I think improving this dropdown as described in https://github.com/gitpod-io/gitpod/issues/6910 could help resolve most issues one action (probably going back to dashboard) would always surface eliminating the need to click the dropdown options menu and decreasing the list of options there.

Thanks so much for noticing, @mustard-mh! 🙏 💯

geropl (approved):
Looks good and code LGTM

Issue #8836, opened 2022-03-16
Implement Cookie Consent banner


geropl:
Internal thread: https://gitpod.slack.com/archives/C02EN94AEPL/p1646071841755889

We should align with @mikenikles on a implementation. Suggestions:
 - https://www.osano.com/cookieconsent
 - https://brainsum.github.io/cookieconsent/
 - https://github.com/beyonk-adventures/gdpr-cookie-consent-banner (example: https://beyonk.com/)

geropl:
- RFC: https://www.notion.so/gitpod/835214fc21224ba78c3a2d1df95f4faf?v=c3e0899e034040fd889db011241e4733&p=525099227a8d4cb0974032c8232d7af9
 - Data Privacy Guideline: https://www.notion.so/gitpod/Data-Privacy-Guideline-42e96258e7594841a1706ba832c5ac16
 - List of Cookies and Categories: https://www.notion.so/gitpod/a8f22878433e4c00a98dc5affe08f42d?v=2ec614009b1749c2a16401300ed1208d

geropl:
things we need:
 - banner to ask (should be as slick as possible)
 - page to get back, review and change settings

MircoatGitpod:
- We need to allow our users to go back to the cookie consent settings, this could be done with a link inside our website footer

gtsiolis:
@geropl
1. Here's a screenshot below of the cookie consent banner we already have on the website, implemented in https://github.com/gitpod-io/website/pull/298. Cc @mikenikles 
2. We also recently changed the footnote when logging in the dashboard so that you also accept [terms of service](https://www.gitpod.io/terms/) and [privacy policy](https://www.gitpod.io/privacy), see relevant PR in https://github.com/gitpod-io/gitpod/pull/8546. Cc @jldec 

| Existing cookine consent banner in `www.gitpod.io` |
|-|
| <img width="1440" alt="Screenshot 2022-04-06 at 5 56 36 PM" src="https://user-images.githubusercontent.com/120486/162017114-5b74a431-dc95-47bf-a16a-932a5e6c6469.png"> |

geropl:
@gtsiolis Thx for the reference! The motivation to touch this is that the current impl/footnote is not GDPR compliant as we learned from @MircoatGitpod . But the current banner looks like sth we could easily use and extend. WDYT @mikenikles ?

MircoatGitpod:
@laushinka Yes we could implement it like this, great suggestion!

[...]

PR #9299, opened 2022-04-13
Implement Cookie Consent Banner and Form (#8836)

laushinka:
## Description
1. Shows a cookie consent banner
2. Allows seeing and modifying cookie settings. The modifications don't affect the Dashboard because we only use strictly necessary cookies. Non-necessary cookies affect the website, which will be implemented by @nisarhassan12. 

Set cookies[[1](https://www.notion.so/gitpod/a8f22878433e4c00a98dc5affe08f42d)][[2](https://www.notion.so/gitpod/Cookie-consent-banner-9dfcb9a50b3444c99c2d4cc1947a8f68#88e4c5e5e60149b0bee20bda2de1cfe7)]:
```
"gitpod-io-consent": {
    necessary: true,
    analytical: false,
    targeting: false
}
```

| BANNER | SETTINGS |
| - | - |
| Screenshot 2022-04-13 at 15 03 38 | Screenshot 2022-04-20 at 16 25 20 |

Follow-up task with @nisarhassan12:
- [ ] Create an identical file that contains the format of the cookie settings, to be kept in-sync by both website and app.

## Related Issue(s)

Relates #8836 
Another implementation will be done on the website.

## How to test
Visit https://laushinka-96c11f76fa.staging.gitpod-dev.com/workspaces.
1. Check your cookies either by going to Application in the dev tools or by running `document.cookie` in the console.
2. If `gp-analytical` and `gp-targeting` are not set, the banner should show.
3. Modify settings or click the buttons and run `document.cookie` to see that they are set properly.

## Release Notes

```release-note
Allow users to accept or reject cookies, compliant with GDPR.
```

### DIFF
--- a/components/dashboard/src/App.tsx
+++ b/components/dashboard/src/App.tsx
@@ -5,9 +5,9 @@
  */
 
 import React, { Suspense, useContext, useEffect, useState } from "react";
-import Menu from "./Menu";
 import { Redirect, Route, Switch } from "react-router";
 
+import Menu from "./Menu";
 import { Login } from "./Login";
 import { UserContext } from "./user-context";
 import { TeamsContext } from "./teams/teams-context";
@@ -44,6 +44,9 @@ import {
 import { refreshSearchData } from "./components/RepositoryFinder";
 import { StartWorkspaceModal } from "./workspaces/StartWorkspaceModal";
 import { parseProps } from "./start/StartWorkspace";
+import CookieBanner from "./CookieBanner";
+import CookieSettingsModal from "./CookieSettingsModal";
+import { cookieConsentDefaultValue, getConsentCookies } from "./consent-cookie";
 
 const Setup = React.lazy(() => import(/* webpackPrefetch: true */ "./Setup"));
 const Workspaces = React.lazy(() => import(/* webpackPrefetch: true */ "./workspaces/Workspaces"));
@@ -154,6 +157,9 @@ function App() {
     const [loading, setLoading] = useState<boolean>(true);
     const [isWhatsNewShown, setWhatsNewShown] = useState(false);
     const [isSetupRequired, setSetupRequired] = useState(false);
+    const [showCookieBanner, setShowCookieBanner] = useState<boolean>(true);
+    const [showCookieSettings, setShowCookieSettings] = useState<boolean>(false);
+    const [cookiePreferences, setCookiePreferences] = useState(cookieConsentDefaultValue);
     const history = useHistory();
 
     useEffect(() => {
@@ -274,6 +280,23 @@ function App() {
         }
     }, [user]);
 
+    const checkCookie = () => {
+        if (isGitpodIo()) {
+            const consent = getConsentCookies();
+            if (consent === undefined) {
+                setShowCookieBanner(true);
+                return;
+            } else {
+                setCookiePreferences(JSON.parse(consent));
+                setShowCookieBanner(false);
+            }
+        }
+    };
+
+    useEffect(() => {
+        checkCookie();
+    }, []);
+
     // redirect to website for any website slugs
     if (isGitpodIo() && isWebsiteSlug(window.location.pathname)) {
         window.location.host = "www.gitpod.io";
@@ -345,10 +368,49 @@ function App() {
         false,
     );
 
+    const acceptAllCookies = () => {
+        setCookiePreferences({
+            ...cookieConsentDefaultValue,
+            analytical: true,
+            targeting: true,
+        });
+        setShowCookieBanner(false);
+    };
+
+    const rejectAllCookies = () => {
+        setCookiePreferences({
+            ...cookieConsentDefaultValue,
+            analytical: false,
+            targeting: false,
+        });
+        setShowCookieBanner(false);
+    };
+
+    const openCookieSettings = () => {
+        setShowCookieSettings(true);
+    };
+
+    const saveCookiePreferences = () => {
+        setShowCookieSettings(false);
+        setShowCookieBanner(false);
+    };
+
+    const handleOnClose = () => {
+        setShowCookieSettings(false);
+    };
+
     let toRender: React.ReactElement = (
         <Route>
             <div className="container">
                 <Menu />
+                {isGitpodIo() && showCookieSettings && (
+                    <CookieSettingsModal
+                        showCookieSettings={showCookieSettings}
+                        cookiePreferences={cookiePreferences}
+                        onClose={handleOnClose}
+                        onSave={saveCookiePreferences}
+                    />
+                )}
                 <Switch>
                     <Route path={projectsPathNew} exact component={NewProject} />
                     <Route path="/open" exact component={Open} />
@@ -473,6 +535,13 @@ function App() {
                     ></Route>
                 </Switch>
                 <StartWorkspaceModal />
+                {showCookieBanner && (
+                    <CookieBanner
+                        onAcceptAllCookies={acceptAllCookies}
+                        onOpenCookieSettings={openCookieSettings}
+                        onRejectAllCookies={rejectAllCookies}
+                    />
+                )}
             </div>
         </Route>
     );
--- a/components/dashboard/src/CookieBanner.tsx
+++ b/components/dashboard/src/CookieBanner.tsx
@@ -1,0 +1,55 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { setCookie } from "./consent-cookie";
+
+function CookieBanner(props: {
+    onOpenCookieSettings: () => void;
+    onAcceptAllCookies: () => void;
+    onRejectAllCookies: () => void;
+}) {
+    const onAccept = () => {
+        setCookie(true, true);
+        props.onAcceptAllCookies();
+    };
+
+    const onReject = () => {
+        setCookie(false, false);
+        props.onRejectAllCookies();
+    };
+
+    return (
+        <div className="flex justify-between items-center mx-auto h-12 px-4 py-2 text-center text-xs text-gray-600 bg-gray-200 w-full bottom-0 left-0 fixed">
+            <p className="text-gray-600">
+                The website uses cookies to enhance the user experience. Read our{" "}
+                <a
+                    className="gp-link hover:text-gray-600"
+                    target="gitpod-privacy"
+                    href="https://www.gitpod.io/privacy/"
+                >
+                    privacy policy{" "}
+                </a>
+                for more info.
+            </p>
+            <div className="flex gap-1">
+                <button
+                    className="py-3 bg-sand-dark underline text-xs text-gray-400 hover:text-gray-600"
+                    onClick={() => props.onOpenCookieSettings()}
+                >
+                    Modify settings
+                </button>
+                <button className="bg-gray-100 rounded-lg hover:bg-white text-xs text-gray-600" onClick={onAccept}>
+                    Accept Cookies
+                </button>
+                <button className="bg-gray-100 rounded-lg hover:bg-white text-xs text-gray-600" onClick={onReject}>
+                    Reject All
+                </button>
+            </div>
+        </div>
+    );
+}
+
+export default CookieBanner;
--- a/components/dashboard/src/CookieSettingsModal.tsx
+++ b/components/dashboard/src/CookieSettingsModal.tsx
@@ -1,0 +1,117 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { useState } from "react";
+import CheckBox from "./components/CheckBox";
+import Modal from "./components/Modal";
+import { cookieConsentDefaultValue, getConsentCookies, setCookie } from "./consent-cookie";
+
+function CookieSettingsModal(props: {
+    showCookieSettings: boolean;
+    cookiePreferences: { necessary: boolean; analytical: boolean; targeting: boolean };
+    onClose: () => void;
+    onSave: () => void;
+}) {
+    const [cookiePreferences, setCookiePreferences] = useState(cookieConsentDefaultValue);
+
+    const handleAnalyticalCheckboxChange = () => {
+        setCookiePreferences({
+            ...cookiePreferences,
+            analytical: !cookiePreferences.analytical,
+        });
+    };
+
+    const handleTargetingCheckboxChange = () => {
+        setCookiePreferences({
+            ...cookiePreferences,
+            targeting: !cookiePreferences.targeting,
+        });
+    };
+
+    const onSave = () => {
+        setCookie(cookiePreferences.analytical, cookiePreferences.targeting);
+        setCookiePreferences({
+            ...cookieConsentDefaultValue,
+            analytical: cookiePreferences.analytical,
+            targeting: cookiePreferences.targeting,
+        });
+
+        props.onSave();
+    };
+
+    const onClose = () => {
+        const value = getConsentCookies();
+        if (value) {
+            setCookiePreferences(JSON.parse(value));
+        }
+
+        props.onClose();
+    };
+
+    return (
+        <Modal visible={props.showCookieSettings} onClose={onClose}>
+            <h3 className="mb-4">Cookie settings</h3>
+            <div className="flex flex-col -mx-6 px-6 py-4 border-t border-b border-gray-200 dark:border-gray-800">
+                <div className="flex gap-2">
+                    <div className="-mt-4">
+                        <CheckBox title="" desc={undefined} checked={true} disabled={true} name="necessary" />
+                    </div>
+                    <h4>Strictly necessary cookies</h4>
+                </div>
+                <p className="mb-4">
+                    These are cookies that are required for the operation of our website and under our terms with you.
+                    They include, for example, cookies that enable you to log into secure areas of our website or make
+                    use of our teams and subscription services.
+                </p>
+                <div className="flex gap-2">
+                    <div className="-mt-4">
+                        <CheckBox
+                            title=""
+                            desc={undefined}
+                            checked={cookiePreferences.analytical}
+                            name="analytical"
+                            onChange={handleAnalyticalCheckboxChange}
+                        />
+                    </div>
+                    <h4>Analytical & performance cookies</h4>
+                </div>
+                <p className="mb-4">
+                    These allow us to recognise and count the number of visitors and to see how visitors move around our
+                    website when they are using it. This helps us improve the way our website works, for example, by
+                    ensuring that users are finding what they are looking for easily.
+                </p>
+                <div className="flex gap-2">
+                    <div className="-mt-4">
+                        <CheckBox
+                            title=""
+                            desc={undefined}
+                            checked={cookiePreferences.targeting}
+                            name="targeting"
+                            onChange={handleTargetingCheckboxChange}
+                        />
+                    </div>
+                    <h4>Targeting & advertising cookies</h4>
+                </div>
+                <p className="mb-4">
+                    These cookies record your visit to our website, the pages you have visited and the links you have
+                    followed. We will use this information subject to your choices and preferences to make our website
+                    and the advertising displayed on it more relevant to your interests. We may also share this
+                    information with third parties for this purpose.
+                </p>
+            </div>
+            <div className="flex justify-end mt-6">
+                <button className="secondary" onClick={onClose}>
+                    Cancel
+                </button>
+                <button className="ml-2" onClick={onSave}>
+                    Save
+                </button>
+            </div>
+        </Modal>
+    );
+}
+
+export default CookieSettingsModal;
--- a/components/dashboard/src/Login.tsx
+++ b/components/dashboard/src/Login.tsx
@@ -248,15 +248,6 @@ export function Login() {
                             >
                                 terms of service
                             </a>{" "}
-                            and{" "}
-                            <a
-                                className="gp-link hover:text-gray-600"
-                                target="gitpod-privacy"
-                                href="https://www.gitpod.io/privacy/"
-                            >
-                                privacy policy
-                            </a>
-                            .
                         </span>
                     </div>
                 </div>
--- a/components/dashboard/src/consent-cookie.ts
+++ b/components/dashboard/src/consent-cookie.ts
@@ -1,0 +1,51 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import Cookies, { CookieAttributes } from "js-cookie";
+
+export const COOKIE_NAME = `${window.location.hostname}-consent`;
+
+export let cookieConsentDefaultValue = {
+    necessary: true,
+    analytical: false,
+    targeting: false,
+};
+
+export function cookieSettingOptions(domain: string): CookieAttributes {
+    return {
+        domain: `.${domain}`,
+        expires: 365,
+        path: "/",
+        secure: false,
+        sameSite: "lax",
+    };
+}
+
+export function setNonNecessaryValues(analyticalValue: boolean, targetingValue: boolean) {
+    return {
+        ...cookieConsentDefaultValue,
+        analytical: analyticalValue,
+        targeting: targetingValue,
+    };
+}
+
+export function serialize(value: any) {
+    const stringified = JSON.stringify(value);
+    const quotesRemoved = stringified.replace("%22", '"');
+    return quotesRemoved;
+}
+
+export function setCookie(analyticalValue: boolean, targetingValue: boolean) {
+    return Cookies.set(
+        COOKIE_NAME,
+        serialize(setNonNecessaryValues(analyticalValue, targetingValue)),
+        cookieSettingOptions(window.location.hostname),
+    );
+}
+
+export function getConsentCookies() {
+    return Cookies.get(COOKIE_NAME);
+}
--- a/components/dashboard/package.json
+++ b/components/dashboard/package.json
@@ -25,7 +25,7 @@
     "@testing-library/react": "^11.2.6",
     "@testing-library/user-event": "^12.1.10",
     "@types/jest": "^26.0.22",
-    "@types/js-cookie": "^2.2.7",
+    "@types/js-cookie": "^3.0.1",
     "@types/node": "^16.11.0",
     "@types/react": "^17.0.0",
     "@types/react-dom": "^17.0.3",
--- a/components/dashboard/src/consent-cookie.test.ts
+++ b/components/dashboard/src/consent-cookie.test.ts
@@ -1,0 +1,29 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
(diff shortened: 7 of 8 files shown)


components/gitpod-protocol/src/util/gitpod-cookie.ts
```diff
@@ -3,7 +3,10 @@
  * Licensed under the GNU Affero General Public License (AGPL).
  * See License-AGPL.txt in the project root for license information.
  */
+
 import * as cookie from "cookie";
+import * as Cookies from "js-cookie";
+import { CookieAttributes } from "js-cookie";
```
laushinka, line 9: I ended up using the library `js-cookie` that we are also using in the frontend because this worked better in setting multiple cookies.

components/dashboard/src/App.tsx
```diff
@@ -345,10 +362,73 @@ function App() {
         false,
     );
 
+    const acceptAllCookies = () => {
+        GitpodCookie.setCookie(ANALYTICAL_COOKIES, "true");
```
geropl, line 366: @laushinka It would be nice if we would limit the amount of cookies involved.

I imagined we'd use a single cookie that we use to store a `JSON.stringify`d (propably even `base64` encoded) shape of this form to:
```
{
  analytics: bool | undefined;
  targeting: bool | undefined;
}
```

components/gitpod-protocol/src/util/gitpod-cookie.ts
```diff
@@ -12,6 +15,10 @@ import * as cookie from "cookie";
 export const NAME = "gitpod-user";
 export const VALUE = "true";
 
+export const NECESSARY_COOKIES = "gp-necessary";
```
geropl, line 18: The (single) cookie name needs to contain the domain name. E.g. for `gitpod.io`, it could be prefixed with `gitpod-io-consent`.

components/dashboard/src/App.tsx
laushinka: Hmm, even when stringified in our code, it would end up being two different cookies in the browser if I understand correctly?

geropl: > it would end up being two different cookies in the browser if I understand correctly?

The idea is to just have one cookie. basically doing:
```
const consent = { analytics: true, targeting: false };
ConsentCookie.set(domain, JSON.stringify(consent));
...
const consent = JSON.parse(ConsentCookie.get(domain));
```

components/gitpod-protocol/src/util/gitpod-cookie.ts
laushinka: cc @nisarhassan12 so we will align

components/dashboard/src/App.tsx
laushinka: Have synced with @geropl and will adjust so that it's one cookie with an object for a value. Something like:
```
"gitpod-io-consent": {
    necessary: true,
    analytical: true,
    targeting: true
}
```

laushinka: cc @nisarhassan12

laushinka: The change to object was trickier than anticipated, because special characters of the values end up being encoded by the libraries. Ideally @nisarhassan12 and I would like the cookies to already be parsed and human-readable, but we decided to leave it to the consumer of the value, i.e. website and other parts of the app code, to parse it.

components/dashboard/src/consent-cookie.tsx
```diff
@@ -0,0 +1,51 @@
+/**
```
geropl, line 1: :point_up: nit: this file can be `.ts`

components/dashboard/src/App.tsx
laushinka: ✅

components/dashboard/src/consent-cookie.tsx
```diff
@@ -0,0 +1,51 @@
...
+    return {
+        ...cookieConsentDefaultValue,
+        analytical: analyticalValue,
+        targeting: targetingValue,
+    };
+}
+
+export function serialize(value: any) {
+    const stringified = JSON.stringify(value);
+    const quotesRemoved = stringified.replace("%22", '"');
```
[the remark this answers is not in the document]
laushinka: Yes this was quite a hair-pulling moment[[1](https://github.com/gitpod-io/gitpod/pull/9299#discussion_r850544862)]. Will this be ok to ship or should we fix it?

components/dashboard/src/App.tsx
```diff
@@ -274,6 +281,32 @@ function App() {
         }
     }, [user]);
 
+    const checkCookie = () => {
+        if (isGitpodIo()) {
+            if (Cookies.get(COOKIE_NAME) === undefined) {
```
geropl, line 286: This code would read easier in a form a-la:
```
const consent = ConsentCookie.get(domain);
if (consent === undefined) {
   showCookieBanner();
}
```
Everything else should be an implementation detail.

laushinka: Thanks! Definitely better :)

components/dashboard/src/App.tsx
```diff
@@ -345,10 +378,77 @@ function App() {
...
+            setCookiePreferences(JSON.parse(value));
+        }
+        setShowCookieSettings(false);
+    };
+
     let toRender: React.ReactElement = (
         <Route>
             <div className="container">
                 <Menu />
+                {isGitpodIo() && showCookieSettings && (
```
geropl, line 442: React question: is there a way to read this `showCookieSettings` from `CookieSettingsModal`? :thinking: 

because then we could push all the handlers into that modal, with it polluting `App.tsx` too much.

components/dashboard/src/Login.tsx
```diff
@@ -34,7 +35,13 @@ function Item(props: { icon: string; iconSize?: string; text: string }) {
 }
 
 export function markLoggedIn() {
-    document.cookie = GitpodCookie.generateCookie(window.location.hostname);
+    document.cookie = GitpodCookie.generateCookie(GitpodCookie.NAME, GitpodCookie.VALUE, window.location.hostname);
+    // setting document.cookie more than once doesn't replace, only appends
+    document.cookie = GitpodCookie.generateCookie(
```
geropl, line 40: I don't understand why we need to set it here. :thinking:

components/dashboard/src/App.tsx
```diff
@@ -154,6 +158,9 @@ function App() {
     const [loading, setLoading] = useState<boolean>(true);
     const [isWhatsNewShown, setWhatsNewShown] = useState(false);
     const [isSetupRequired, setSetupRequired] = useState(false);
+    const [showCookieBanner, setShowCookieBanner] = useState<boolean>(true);
+    const [showCookieSettings, setShowCookieSettings] = useState<boolean>(false);
+    const [cookiePreferences, setCookiePreferences] = useState(cookieConsentDefaultValue);
```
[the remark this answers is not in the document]
laushinka: Do you mean the cookiePreferences? Or all the local states?

components/dashboard/src/App.tsx
laushinka: It's true that App.tsx is already quite convoluted. But since the banner and settings both need to live in the App.tsx so that they will be shown across the app, and since both the settings and the banner need to inform the component that holds them both (App.tsx), then the App.tsx needs to be the one handling the state changes and reflecting the changing states[[1](https://reactjs.org/docs/lifting-state-up.html)]. 
I can sleep on it and think if other things can be extracted out..

components/dashboard/src/Login.tsx
laushinka: I realize I overthought this and you're right! We only need to set all the consent cookies once there is an explicit choice made by the user. Will make the change. Thank you!

components/dashboard/src/App.tsx
laushinka: I extracted the non-shared states. It was a good suggestion 🙏🏽

gtsiolis:
[Incoming changes](https://github.com/gitpod-io/gitpod/pull/9299#event-6454332495), coming back to this later after the build has passed! ⏱️

components/dashboard/src/CookieSettingsModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+                        desc={undefined}
+                        checked={cookiePreferences.targeting}
+                        name="targeting"
+                        onChange={handleTargetingCheckboxChange}
+                    />
+                </div>
+            </div>
+            <p className="mb-4">
+                These cookies record your visit to our Website, the pages you have visited and the links you have
+                followed. We will use this information subject to your choices and preferences to make our Website and
```
gtsiolis, line 99: **thought**: Not sure if this is also going to be re-used on the website, but using both Site and Website terms within the same modal feels a bit confusing, no?

components/dashboard/src/CookieSettingsModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+                the advertising displayed on it more relevant to your interests. We may also share this information with
+                third parties for this purpose.
+            </p>
+            <button onClick={onSave}>Save</button>
```
gtsiolis, line 103: **issue**: In most, if not, all other modals, we're grouping the primary and secondary actions towards the right side of the modal. Aligning the primary action here on the left could affect memory muscle.

gtsiolis, line 103: **suggestion**: Moving this to the right side of the modal and including Cancel as a secondary action button could be nice.

gtsiolis, line 103: **suggestion**: In all other modals, we're currently using a border line as a separator between the modal content and the action buttons on the bottom.

- [ ] Could we also use this here?
- [ ] The sample applies also for the header section of this modal.

gtsiolis (commented):
Thanks, @laushinka!

I've added a first round of comments below while taking a closer look at the preview environment of the last successful build, not sure if the recent updates changed anything mentioned in the following comments.

**issue**: However, it's still not clear to me whether showing the cookie consent banner is _needed_ for the dashboard, see [relevant disccussion](https://gitpod.slack.com/archives/C01KPEPQYE6/p1649435475733359?thread_ts=1649430296.927449&cid=C01KPEPQYE6):

> For me it’s still not clear whether this is something we want or have to add in the dashboard. Usually such controls exist only in the marketing website before signing in, not within the product (dashboard) itself. 

**question:** Do we have any example use cases where other products show a cookie consent banner after the user has logged in? Even the examples used in the [relevant document](https://www.notion.so/gitpod/Cookie-consent-banner-9dfcb9a50b3444c99c2d4cc1947a8f68#aa215383d5064eecab18d6ffdc8069ce) (internal) are also _not_ using a cookie consent banner after logging in. Cc @MircoatGitpod

components/dashboard/src/CookieBanner.tsx
```diff
@@ -0,0 +1,55 @@
...
+            </p>
+            <div className="flex gap-1">
+                <button
+                    className="py-3 bg-sand-dark underline text-xs text-gray-400 hover:text-gray-600"
+                    onClick={() => props.onOpenCookieSettings()}
+                >
+                    Modify settings
+                </button>
+                <button className="bg-off-white rounded-lg hover:bg-white text-xs text-gray-600" onClick={onAccept}>
+                    Accept Cookies
```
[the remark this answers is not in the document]
MircoatGitpod: @gtsiolis From the compliance perspective it's no problem to have one cookie banner that stores preferences across our website and product. I also believe this is the more user-friendly approach.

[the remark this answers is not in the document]
laushinka: @gtsiolis If the user has accepted it on the website, they will not see it anymore inside the dashboard. The cookies are tracked along the gitpod-io domain, so both the website and the dashboard knows when the user has made a choice. (The implementation on the website will follow)

components/dashboard/src/CookieSettingsModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        props.onClose();
+    };
+
+    return (
+        <Modal visible={props.showCookieSettings} onClose={onClose}>
+            <h3 className="mb-4">Cookie settings</h3>
+            <div className="flex gap-2">
+                <h4>Strictly necessary cookies</h4>
```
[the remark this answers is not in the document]
laushinka: Yes. @nisarhassan12 will implement the same based on this document[[1](https://www.notion.so/gitpod/a8f22878433e4c00a98dc5affe08f42d?v=2ec614009b1749c2a16401300ed1208d)]

MircoatGitpod:
> Thanks, @laushinka!
> 
> I've added a first round of comments below while taking a closer look at the preview environment of the last successful build, not sure if the recent updates changed anything mentioned in the following comments.
> 
> **issue**: However, it's still not clear to me whether showing the cookie consent banner is _needed_ for the dashboard, see [relevant disccussion](https://gitpod.slack.com/archives/C01KPEPQYE6/p1649435475733359?thread_ts=1649430296.927449&cid=C01KPEPQYE6):
> 
> > For me it’s still not clear whether this is something we want or have to add in the dashboard. Usually such controls exist only in the marketing website before signing in, not within the product (dashboard) itself.
> 
> **question:** Do we have any example use cases where other products show a cookie consent banner after the user has logged in? Even the examples used in the [relevant document](https://www.notion.so/gitpod/Cookie-consent-banner-9dfcb9a50b3444c99c2d4cc1947a8f68#aa215383d5064eecab18d6ffdc8069ce) (internal) are also _not_ using a cookie consent banner after logging in. Cc @MircoatGitpod

In the past I have seen a few products that just display the same banner inside the product. In some products you can find cookie selection inside the user settings - both can work to comply with privacy laws.

components/dashboard/src/CookieSettingsModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+                    <CheckBox title="" desc={undefined} checked={true} disabled={true} name="necessary" />
+                </div>
+            </div>
+            <p className="mb-4">
+                These are cookies that are required for the operation of our Site and under our terms with you. They
+                include, for example, cookies that enable you to log into secure areas of our Site or (on other sites)
+                use a shopping cart or make use of e-billing services.
+            </p>
+            <div className="flex gap-2">
+                <h4>Analytical / Performance cookies</h4>
```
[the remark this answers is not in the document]
laushinka: The options that the user makes apply for both website and dashboard - they can't choose only for either one.

components/dashboard/src/CookieSettingsModal.tsx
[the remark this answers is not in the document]
laushinka: I realize I've been replying without reloading first so I didn't see @MircoatGitpod's replies 😆

components/dashboard/src/CookieBanner.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+    const onReject = () => {
+        setCookie(false, false);
+        props.onRejectAllCookies();
+    };
+
+    return (
+        <div className="flex justify-between items-center mx-auto h-12 px-4 py-2 text-center text-xs text-gray-600 bg-sand-dark w-full bottom-0 left-0 fixed">
+            <p className="text-gray-600">
+                The website uses cookies to enhance the user experience. Read our{" "}
```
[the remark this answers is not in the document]
laushinka: Hmm, since this will explain our usage across both website and dashboard, we should reflect it in a more unified way. How about "We use cookies" or "The website and dashboard..." cc @MircoatGitpod

laushinka:
Apologies to reviewers for my late replies - currently on sick leave 🙏🏽

components/dashboard/src/CookieSettingsModal.tsx
laushinka: Definitely. Will use "website" (lowercase) instead.

components/dashboard/src/CookieSettingsModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+            <div className="flex gap-2">
+                <h4>Strictly necessary cookies</h4>
+                <div className="-mt-4">
+                    <CheckBox title="" desc={undefined} checked={true} disabled={true} name="necessary" />
+                </div>
+            </div>
+            <p className="mb-4">
+                These are cookies that are required for the operation of our Site and under our terms with you. They
+                include, for example, cookies that enable you to log into secure areas of our Site or (on other sites)
+                use a shopping cart or make use of e-billing services.
```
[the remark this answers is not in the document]
laushinka: You're right, we should reflect the wording better to reflect our products. How about "They include, for example, cookies that enable you to log into secure areas of our website or make use of our teams and subscription services." (aligned with our cookies list [here](https://www.notion.so/gitpod/a8f22878433e4c00a98dc5affe08f42d?v=2ec614009b1749c2a16401300ed1208d)). cc @MircoatGitpod

components/dashboard/src/CookieSettingsModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+        <Modal visible={props.showCookieSettings} onClose={onClose}>
+            <h3 className="mb-4">Cookie settings</h3>
+            <div className="flex gap-2">
+                <h4>Strictly necessary cookies</h4>
+                <div className="-mt-4">
+                    <CheckBox title="" desc={undefined} checked={true} disabled={true} name="necessary" />
+                </div>
+            </div>
+            <p className="mb-4">
+                These are cookies that are required for the operation of our Site and under our terms with you. They
```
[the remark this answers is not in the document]
laushinka: (Related to [this thread](https://github.com/gitpod-io/gitpod/pull/9299#discussion_r854170095))

gtsiolis:
> In the past I have seen a few products that just display the same banner inside the product. In some products you can find cookie selection inside the user settings - both can work to comply with privacy laws.

Thanks for clarifying, @MircoatGitpod!

1️⃣ **suggestion**: I'd avoid introducing a preference for this in user settings. 💭 

2️⃣ **question**: Regarding the banner, could we find or list _one_ or _two_ example products that are diplaying the same banner inside the product? Sorry for the small pushback here but I'd like to make sure we are solving the right problem as the current approach could a) affect the overall user experience of the product, b) introduce points of technical maintainance that we need to later sync between the website and the dashboard on future updates, c) confuse users as the pattern of nudging users to accept cookies within products is not well established, and more.

3️⃣ **thought**: During the dashboard redesign we decided to go with the [browse wrap](https://en.wikipedia.org/wiki/Browse_wrap) instead of [clickwrap](https://en.wikipedia.org/wiki/Clickwrap) to cover user agreement regarding terms of service and privacy policy when using the dashboard, see [relevant discussion](https://gitpod.slack.com/archives/C01KGM9BUNS/p1615544432040800) (internal). Other products include an extra cookie policy link or include the cookie policy within the privacy policy and link once. From what I've seen, we're also already referencing cookies in our [privacy policy](https://www.gitpod.io/privacy/#details-III), no?

4️⃣ **question**: Could we continue with the browse wrap approach and simply extend the footer notice or the cookies' section inside our privacy policy? Asking as, from what I've seen in the [cookies list document](https://www.notion.so/gitpod/a8f22878433e4c00a98dc5affe08f42d?v=2ec614009b1749c2a16401300ed1208d) (internal), the cookies related to the product (dashboard) include only strictly necessary cookies and opting in or out of the rest of the cookie categories has no real effect on the dashboard, right?

| BEFORE | AFTER |
|-|-|
| Screenshot 2022-04-20 at 2 18 28 PM (2) | Screenshot 2022-04-20 at 2 18 23 PM (2) |

components/dashboard/src/CookieSettingsModal.tsx
trumbitta: Please don't let me hijack the discussion 🙏 

I just wanted to point out that this comment and [the one about checkboxes](https://github.com/gitpod-io/gitpod/pull/9299#discussion_r852992789) immediately make the words ["encapsulation" and "design system"](https://discord.com/channels/816244985187008514/914836798830690344/915363278031577109) light up in my mind like [big fire signals on the sands of a remote island at night](https://bluejayblog.files.wordpress.com/2014/06/sos-aregua.jpg) 😄

gtsiolis: Thanks for chiming in, @trumbitta! ✨ 

Here are some benefits of the current status quo:
1. Running into such cases through code reviews helps improve our team by exposing team members to design thinking and why some options are better than others.
2. Keeping existing components loose can breed new component structures that are needed as the product matures. Using this approach means embracing technical debt and continuously iterating. For example, the alert component improved over time because of these loose ends, and eventually we managed to introduce a better and more flexible component instead of maintaining two separate components, see https://github.com/gitpod-io/gitpod/issues/9411.
3. Focusing on shipping this as a [minimum viable change (MVC)](https://www.notion.so/gitpod/Core-Values-Differentiators-2ed4c2f93c84499b98e3b5389980992e#0c2e0d6c8a99404b97ccdc206b573d72), even if the button was never aligned to the right before merging this, sounds optimal as we're always making the quickest change possible to improve the user's outcome.

MircoatGitpod:
Hi @gtsiolis,

no problem at all, thank you for your feedback! 👍 

1️⃣ suggestion: I'd avoid introducing a preference for this in user settings. 💭

-> that's fine, it's just a way in avoiding to disturb our users with banners.

2️⃣ question: Regarding the banner, could we find or list one or two example products that are displaying the same banner or at least the same cookie setting options across their product and marketing website?

-> will keep my eyes open, I have seen a few. Recently I have just see this in Notion.so, they seem to display the banner to product users once and then move cookie related things to the settings
image


3️⃣ thought: During the dashboard redesign we decided to go with the [browse wrap](https://en.wikipedia.org/wiki/Browse_wrap) instead of [clickwrap](https://en.wikipedia.org/wiki/Clickwrap) to cover user agreement regarding terms of service and privacy policy when using the dashboard, see [relevant discussion](https://gitpod.slack.com/archives/C01KGM9BUNS/p1615544432040800) (internal). Other products include an extra cookie policy link or expand on the cookie policy within the [privacy policy](https://www.gitpod.io/privacy/#details-III) and link once. From what I've seen, we're also already referencing cookies in our privacy policy, no?

-> Yes our privacy policy will reference to our to-be-published cookie policy, we wouldn't need to mention the cookie policy itself on the login form. Inside the cookie banner we should refer to one of those, inside the cookie settings dialog it could make sense to have a cookie policy reference. I plan to upload our policy at gitpod.io/cookies

4️⃣ question: Could we continue with the browse wrap approach and simply extend the footer notice or the cookies' section inside our privacy policy, see BEFORE and AFTER screenshots below? Asking as, from what I've seen in the [cookies list document](https://www.notion.so/gitpod/a8f22878433e4c00a98dc5affe08f42d?v=2ec614009b1749c2a16401300ed1208d) (internal), the cookies related to the product (dashboard) include only strictly necessary cookies and opting in or out of the rest of the cookie categories has no real effect on the dashboard, right?

-> Cookie List will be a part of our future Cookie Policy. As mentioned above, it's fine to only reference to our ToS and Privacy Policy.

Please let me know if you have any more questions 🙏

MircoatGitpod:
@gtsiolis @laushinka I forgot to add the following:

You are right, at the moment our product is only processing "strictly necessary" cookies - these don't require any opt-in. Unless this aspect is subject to change somewhere in the future, when you only visit our product you wouldn't need to be presented with banner as of today.

geropl:
@laushinka will have another round this afternoon! :pray:

components/dashboard/src/CookieBanner.tsx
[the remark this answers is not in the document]
geropl: > "We use cookies"

I like this one :point_up:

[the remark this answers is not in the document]
geropl: or yet another alternative: "Gitpod uses cookies"

components/dashboard/src/consent-cookie.ts
```diff
@@ -0,0 +1,51 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import Cookies, { CookieAttributes } from "js-cookie";
+
+export const COOKIE_NAME = `${window.location.hostname}-consent`;
```
geropl, line 9: :+1:

components/dashboard/src/consent-cookie.ts
```diff
@@ -0,0 +1,51 @@
...
+
```
geropl, line 10: It would be nice to have an 
```
interface ConsentSettings {
    necessary: true,
    analytical: false,
    targeting: false,
}
```
here.

components/dashboard/src/consent-cookie.ts
```diff
@@ -0,0 +1,51 @@
...
+    return {
+        domain: `.${domain}`,
+        expires: 365,
+        path: "/",
+        secure: false,
+        sameSite: "lax",
+    };
+}
+
+export function setNonNecessaryValues(analyticalValue: boolean, targetingValue: boolean) {
```
geropl, line 27: nit: if we had a [CookieSettings](https://github.com/gitpod-io/gitpod/pull/9299/files#r855087946) we could make the interface even nicer by taking `ConsentCookie.setValues(values: Drop<CookieSettings, "necessary">)`.

components/dashboard/src/App.tsx
```diff
@@ -345,10 +368,49 @@ function App() {
         false,
     );
 
+    const acceptAllCookies = () => {
```
geropl, line 371: If we introduced a `ConsentCookieContext` I think all of these methods could move into the modal/banner/consent-cookie.ts - not sure though. :thinking: 
I suggest we sync next week and play around with it together. :keyboard:  :pray:

gtsiolis:
> You are right, at the moment our product is only processing "strictly necessary" cookies - these don't require any opt-in. Unless this aspect is subject to change somewhere in the future, when you only visit our product you wouldn't need to be presented with banner as of today.

**question**: @MircoatGitpod Do I understand correctly here that it's _no_ longer needed to work on adding the cookie consent banner within the dashboard, since the product (dashboard) is only processing _strictly necessary_ cookies?

**suggestion**: If yes, I'd suggest closing this PR and picking it up again _once_ and _if_ we introduce more categories for cookies like analytical and targeting cookies that would be processed for the product (dashboard).

geropl:
The fact that there is a extensive discussion going on here (76+ counting :exploding_head: ) shows we that there is some additional alignment required, and the fact @laushinka is ooo (and she really should be part of these discussions) I think we should pull the brake here.

I will close the PR for now, so:
 a) we can re-groupd and create additional alignment on the why and how before wasting more hours on iterations
 b) fine-tune the design and implementation (separately)
 c) most importantly: avoid @laushinka feeling forced to participate despite being out of office

gtsiolis:
Thanks for taking action [above](https://github.com/gitpod-io/gitpod/pull/9299#issuecomment-1105241167) and closing this, @geropl! 🙏 

> @MircoatGitpod @gtsiolis What about syncing on Monday when Laurie is hopefully around again?

@geropl I'll be offline on Monday but any other day throughout the week is ok for me.

MircoatGitpod:
> > You are right, at the moment our product is only processing "strictly necessary" cookies - these don't require any opt-in. Unless this aspect is subject to change somewhere in the future, when you only visit our product you wouldn't need to be presented with banner as of today.
> 
> **question**: @MircoatGitpod Do I understand correctly here that it's _no_ longer needed to work on adding the cookie consent banner within the dashboard, since the product (dashboard) is only processing _strictly necessary_ cookies?
> 
> **suggestion**: If yes, I'd suggest closing this PR and picking it up again _once_ and _if_ we introduce more categories for cookies like analytical and targeting cookies that would be processed for the product (dashboard), as this change could bring in little to no value for users.

@gtsiolis - Yes you are right, we would only need to add the banner here if we choose to store cookies or local browser storage that is non necessary to keep our product functioning.  - Sorry, I should have clarified this earlier.

Sync on Monday would be fine to me.

cc @geropl @laushinka

PR #9300, opened 2022-04-13
[public-api-server] Group http and grpc servers into a baseserver package

easyCZ:
## Description

Subsequent PRs will add (as needed):
* Cert support
* Metrics collector and endpoint
* Health/Ready endpoints
* Additional configuration as required

This change is primarily for WebApp, but Github tags Workspace also due to a new library in common-go. Unfortunately, this change is bigger than my regular changes but it needed to remain together to show a working example.

## Related Issue(s)

Part of https://github.com/gitpod-io/gitpod/issues/9229

## How to test

```
cd components/common-go/baseserver && go test ./...
```

```
➜ cd components/public-api-server 
➜ go run main.go 
INFO[0000] Serving gRPC on [::]:9001                     protocol=grpc
INFO[0000] Serving http on [::]:9000                     protocol=http
^CINFO[0004] Received system signal interrupt, closing server. 
INFO[0004] Received graceful shutdown request.          
INFO[0004] GRPC server terminated.                      
INFO[0004] HTTP server terminated.     
```

## Release Notes

```release-note
NONE
```

## Documentation

NONE

/hold


roboquat:

<details>

</details>

easyCZ:
Werft build keeps flaking, but this can be reviewed and I'll re-run the build at a later point. The preview build does work.

components/common-go/baseserver/server.go
```diff
@@ -0,0 +1,211 @@
...
+	return fmt.Sprintf("%s://%s:%d", protocol, s.cfg.hostname, s.cfg.grpcPort)
+}
+
+func (s *Server) HTTPMux() *http.ServeMux {
+	return s.httpMux
+}
+
+func (s *Server) close(ctx context.Context) error {
+	if s.listening == nil {
+		return fmt.Errorf("server is not running, invalid close operaiton")
```
andrew-farries, line 151: nit: typo here (operaiton)

components/common-go/baseserver/server.go
```diff
@@ -0,0 +1,211 @@
...
+		if serveErr := s.http.Serve(s.httpListener); serveErr != nil {
+			if s.isClosing() {
+				return
+			}
+
+			errors <- serveErr
+		}
+	}()
+
+	signals := make(chan os.Signal, 1)
```
andrew-farries, line 101: I don't think you need to specify a channel buffer size here.

components/common-go/baseserver/server.go
```diff
@@ -0,0 +1,211 @@
...
+			WithField("protocol", "grpc").
+			Infof("Serving gRPC on %s", s.grpcListener.Addr().String())
+		if serveErr := s.grpc.Serve(s.grpcListener); serveErr != nil {
+			if s.isClosing() {
+				return
+			}
+
+			errors <- serveErr
+		}
+	}()
```
andrew-farries, line 86: Why does this server (and the http one below) need to be started in a goroutine? Could we not simply error out if we fail to start either of them without having them send their errors on a channel only to collect them on line 105?

andrew-farries (changes requested):

components/common-go/baseserver/server.go
easyCZ: Needs to be a go-routine because grpc.Serve and `http.ListenAndServe` are blocking. This follows the standard golang practice where if you want things to be in the background, the caller can do that themselves.

We need to put them in go-routines because we need to start both of them, and then synchronously await termination of either (or both) of the servers before exiting to maintain the semantics of listen and serve.

components/common-go/baseserver/options_test.go
```diff
@@ -0,0 +1,38 @@
...
+package baseserver
+
+import (
+	"github.com/gitpod-io/gitpod/common-go/log"
+	"github.com/stretchr/testify/require"
+	"testing"
+	"time"
+)
+
+func TestOptions(t *testing.T) {
```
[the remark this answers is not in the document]
easyCZ: Sure, happy to add them.

components/common-go/baseserver/server.go
easyCZ: I was going by this example: https://gobyexample.com/signals

I think it doesn't hurt to specify it as that we're ensuring that repeated signals (unsure if we'd receive them) don't get buffered and block instead. That's desirable as we're looking to limit memory usage and complete our termination procedure. In practice, in k8s, we'll receive SIGINT, and if we don't shut down in time k8s will just terminate the pod entirely

components/common-go/baseserver/options_test.go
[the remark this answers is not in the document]
easyCZ: Added tests and pushed.

andrew-farries (dismissed):

princerachit:
@easyCZ would it make sense if workspace team approves this change only after webapp has approved? As you mentioned the changes are related to webapp and we are only getting tagged for a new library in common.go

princerachit:
@easyCZ would it make sense if workspace team approves this change only after webapp has approved? As you mentioned the changes are related to webapp and we are only getting tagged for a new library in common.go

easyCZ:
@princerachit Thanks for taking a look. @andrew-farries previously approved but I had to rebase due to a downstream change. If you're happy with the changes, I'll put on a hold. In this change, both approvals (WebApp, Workspace) so even with your approval, it won't get auto-merged.

/hold

andrew-farries (approved):

components/common-go/baseserver/options.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+
+		cfg.httpPort = port
+		return nil
+	}
+}
+
+func WithGRPCPort(port int) Option {
+	return func(cfg *config) error {
+		if port < 0 {
```
princerachit, line 60: nit: Should it be ?
```suggestion
		if port <= 0 {
```

components/common-go/baseserver/options.go
```diff
@@ -0,0 +1,95 @@
...
+
+		cfg.grpcPort = port
+		return nil
+	}
+}
+
+func WithLogger(logger *logrus.Entry) Option {
+	return func(cfg *config) error {
+		if logger == nil {
+			return fmt.Errorf("nil logger specified")
```
princerachit, line 72: maybe?
```suggestion
			return fmt.Errorf("no logger specified")
```

princerachit (approved):

components/common-go/baseserver/options.go
easyCZ: This is actually a bug in the HTTP Port option. When you specify port `0`, `net.Listen` actually allocates an open port. I'll fix this in a follow-up PR to avoid tangling you in the review again. Thanks for flagging!

components/common-go/baseserver/options.go
easyCZ: Intent was to be more developer friendly and actually tell them it's nil when we know it is.

PR #9309, opened 2022-04-13
[ws-daemon] Only limit storage device classes

aledbf:
## Description

Some `io.stat` file can return more classes. We need to apply limits only to storage.

```
# cat /sys/fs/cgroup/kubepods/besteffort/pod7d34503e-d63c-427c-bcd9-5f339039d55e/612035c1d87101e7c51997fab8184d966c6b9da851196c622a305246d1397459/io.stat 
8:0 rbytes=19054592 wbytes=0 rios=1240 wios=0 dbytes=0 dios=0
259:2 rbytes=77824 wbytes=3719168 rios=8 wios=809 dbytes=0 dios=0
259:3 rbytes=167936 wbytes=6172672 rios=17 wios=1362 dbytes=0 dios=0
9:43 rbytes=245760 wbytes=9891840 rios=25 wios=2311 dbytes=0 dios=0
``` 

## Release Notes
```release-note
[ws-daemon] Only limit storage device classes
```

### DIFF
--- a/components/ws-daemon/pkg/cgroup/plugin_iolimit_v2.go
+++ b/components/ws-daemon/pkg/cgroup/plugin_iolimit_v2.go
@@ -33,6 +33,9 @@ func (c *IOLimiterV2) Apply(ctx context.Context, basePath, cgroupPath string) er
 		// devices yet, and hence would not impose IO limits on them.
 		ticker := time.NewTicker(30 * time.Second)
 		defer ticker.Stop()
+
+		ioMaxFile := filepath.Join(basePath, cgroupPath)
+
 		for {
 			select {
 			case <-ctx.Done():
@@ -44,14 +47,14 @@ func (c *IOLimiterV2) Apply(ctx context.Context, basePath, cgroupPath string) er
 				c.WriteIOPs = 0
 				c.ReadIOPs = 0
 
-				err := c.writeIOMax(filepath.Join(basePath, cgroupPath))
+				err := c.writeIOMax(ioMaxFile)
 				if err != nil {
 					log.WithError(err).WithField("cgroupPath", cgroupPath).Error("cannot write IO limits")
 				}
 				log.WithField("cgroupPath", cgroupPath).Debug("stopping io limiting")
 				return
 			case <-ticker.C:
-				err := c.writeIOMax(filepath.Join(basePath, cgroupPath))
+				err := c.writeIOMax(ioMaxFile)
 				if err != nil {
 					log.WithError(err).WithField("cgroupPath", cgroupPath).Error("cannot write IO limits")
 				}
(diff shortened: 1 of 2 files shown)


csweichel:
We'll probably wanna include `9` as well:

```
9 block	Metadisk (RAID) devices
		  0 = /dev/md0		First metadisk group
		  1 = /dev/md1		Second metadisk group
		    ...

		The metadisk driver is used to span a
		filesystem across multiple physical disks.
```
([source](https://www.kernel.org/doc/Documentation/admin-guide/devices.txt))

csweichel:
The original thinking was that if you're using it, and you're expending IO bandwidth on the device, we'll want to limit it because it's a shared resource.

aledbf:
> We'll probably wanna include 9 as well:

Fixed

components/ws-daemon/pkg/cgroup/plugin_iolimit_v2.go
```diff
@@ -61,38 +64,52 @@ func (c *IOLimiterV2) Apply(ctx context.Context, basePath, cgroupPath string) er
 	return nil
 }
 
-func (c *IOLimiterV2) writeIOMax(loc string) error {
-	iostat, err := os.ReadFile(filepath.Join(string(loc), "io.stat"))
+func (c *IOLimiterV2) writeIOMax(cgroupPath string) error {
+	iostat, err := os.ReadFile(filepath.Join(string(cgroupPath), "io.stat"))
 	if os.IsNotExist(err) {
+		log.Error("Pod is gone")
```
utam0k, line 70: I feel it isn't an error.  Timing issues may arise as per the comment below.
https://github.com/gitpod-io/gitpod/pull/9309/files#diff-9924bdc24dd06931f08b754cc11839d09c8ad010ce1fb4cc2c376c8b752b0538R71

components/ws-daemon/pkg/cgroup/plugin_iolimit_v2.go
```diff
@@ -61,38 +64,52 @@ func (c *IOLimiterV2) Apply(ctx context.Context, basePath, cgroupPath string) er
...
 		// cgroup gone is ok due to the dispatch/container race
 		return nil
 	}
+
 	if err != nil {
 		return err
 	}
+
+	var classesToLimit = []string{"8", "9"}
```
utam0k, line 79: Since these numbers is too magic, how about add some comments to make readable for humans.
```suggestion
   // 8: block	SCSI disk devices (0-15)
   // 9: char	SCSI tape devices
	var classesToLimit = []string{"8", "9"}
```

components/ws-daemon/pkg/cgroup/plugin_iolimit_v2.go
aledbf: removed

components/ws-daemon/pkg/cgroup/plugin_iolimit_v2.go
aledbf: done

utam0k (approved):
LGTM

PR #9317, opened 2022-04-14
[server] Add priorization to repo URLs

svenefftinge:
## Description
Adds prioritization to suggested repo URLs in new workspace modal, so that new users get their own
repositories suggested before the generic templates.

## How to test
Login in and start a workspace through the new workspace modal.
You should find your own repositories listed above the gitpod templates.

## Release Notes
```release-note
NONE
```

### DIFF
--- a/components/server/src/workspace/gitpod-server-impl.ts
+++ b/components/server/src/workspace/gitpod-server-impl.ts
@@ -1256,7 +1256,7 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
 
     public async getSuggestedContextURLs(ctx: TraceContext): Promise<string[]> {
         const user = this.checkAndBlockUser("getSuggestedContextURLs");
-        const suggestions: Array<{ url: string; lastUse?: string }> = [];
+        const suggestions: Array<{ url: string; lastUse?: string; priority: number }> = [];
         const logCtx: LogContext = { userId: user.id };
 
         // Fetch all data sources in parallel for maximum speed (don't await in this scope before `Promise.allSettled(promises)` below!)
@@ -1266,7 +1266,7 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
         promises.push(
             this.getFeaturedRepositories(ctx)
                 .then((exampleRepos) => {
-                    exampleRepos.forEach((r) => suggestions.push({ url: r.url }));
+                    exampleRepos.forEach((r) => suggestions.push({ url: r.url, priority: 0 }));
                 })
                 .catch((error) => {
                     log.error(logCtx, "Could not get example repositories", error);
@@ -1281,7 +1281,9 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
                         authProviders.map(async (p) => {
                             try {
                                 const userRepos = await this.getProviderRepositoriesForUser(ctx, { provider: p.host });
-                                userRepos.forEach((r) => suggestions.push({ url: r.cloneUrl.replace(/\.git$/, "") }));
+                                userRepos.forEach((r) =>
+                                    suggestions.push({ url: r.cloneUrl.replace(/\.git$/, ""), priority: 5 }),
+                                );
                             } catch (error) {
                                 log.debug(logCtx, "Could not get user repositories from App for " + p.host, error);
                             }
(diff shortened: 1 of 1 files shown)


easyCZ (commented):
This function could be improved in terms of readability by doing the following:
1. First fetch featured repos and store them raw.
2. Also (in parallel) fetch the user repos (from Apps) and store them in a different variable
3. Fetch the user repos (in parallel) from user repos (git hosts) and store in third variable
4. Await all the setlle
5. Map/filter/sort on the results

This helps break the nesting and reduces the volume of transformations done in async flows such that the code is easier to follow. It also allows for the map/filter/sort to be tested without async dependencies.
It would also allow to not need the priorities as the suggestions from various (prioritized) sources can be appended onto the final result directly.

Waiting for preview to test this.

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -1307,7 +1309,9 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
                                     return;
                                 }
                                 const userRepos = await services.repositoryProvider.getUserRepos(user);
-                                userRepos.forEach((r) => suggestions.push({ url: r.replace(/\.git$/, "") }));
+                                userRepos.forEach((r) =>
+                                    suggestions.push({ url: r.replace(/\.git$/, ""), priority: 5 }),
```
jankeromnes, line 1313: Random side note: Would be cool to add `lastUse` here too (e.g. in order to first show repositories I've most recently contributed to, even if I've never opened them in Gitpod before).

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -1343,7 +1347,11 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
         const uniqueURLs = new Set();
         return suggestions
             .sort((a, b) => {
-                // Most recently used first
+                // priority first
+                if (a.priority != b.priority) {
```
jankeromnes, line 1351: I generally try to avoid "loose equality" (e.g. `"5" == 5`) -- is there any reason not to use `!==` here?

```suggestion
                if (a.priority !== b.priority) {
```

jankeromnes (dismissed):
Apologies for jumping on this @easyCZ -- this change really caught my attention. 😅

For what it's worth, it looks good to me. 👍 However I couldn't test it, as https://se-sorting600cb9b1f1.staging.gitpod-dev.com/workspaces seems to be down for me.

Approving, but adding a hold to give you a chance to review @easyCZ

/hold

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -1343,7 +1347,11 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
...
+                    return a.priority > b.priority ? 1 : -1;
```
jankeromnes, line 1352: Should this be sorted in descending order? (I.e. highest priority first?)

```suggestion
                    return a.priority < b.priority ? 1 : -1;
```

svenefftinge:
@easyCZ I understand we always want our codebase to become cleaner and simpler so just adding code without cleaning up now and then isn't sustainable, but I'm generally a little reluctant to rewrite large bits of working code when adding a minor aspect like in this case.

easyCZ:
> @easyCZ I understand we always want our codebase to become cleaner and simpler so just adding code without cleaning up now and then isn't sustainable...

Perfectly valid. My comment was largely a suggestion on an alternative way to think through the changes and possibly suggestions how we can improve this particular section in the future.

To clarify this further, I am happy with the current change to remain. The suggestion is not a requirement for this PR.

svenefftinge:
@easyCZ Awesome, thanks for clarifying. After rereading your comment once more I realize that your suggestion was maybe mainly to have implicit prioritization by maintaining three different arrays instead of adding everything to the same and adding and sorting through an explicit `priorization` property? I don't see how that is much better, especially because we already have other means of explicitly sorting, but I also somehow didn't realize that when I first read the comment.

easyCZ:
I've resolved the merge conflict on this PR, waiting for preview to retest.

easyCZ:
I've also rebased to pull in fixes for broken CI from yesterday.

easyCZ:
When testing, I did the following:
1. Preview env
2. Sign-in with Github
3. Link Gitlab account
4. Open a new workspace with modal
Screenshot 2022-04-20 at 9 28 27

I got `werft` as a the first suggestion, but expected my repos `easyCZ/*` to be first. I did raise a PR against werft in the past. Is this expected @svenefftinge?

svenefftinge:
Yes, it is expected. The result list from all contributions is not ordered, yet. This change basically only moves down the gitpod templates.

easyCZ (approved):
LGTM. Adding hold to land at your leisure.

/hold

Issue #9097, opened 2022-04-04
Epic: Migrate to Installer


geropl:
### Summary

WebApp is the last team that relies on the helm chart for deployments, we should get rid of it.

#### General approach

Go through components one by one and decide for each config to either:
   - work with a default
   - find a way to place variance in the installer

#### Tasks 
 - [x] Setup a diff tool between chart and installer to identify config diff https://github.com/gitpod-io/ops/pull/1558
   - [x] `open-vsx`
     - https://github.com/gitpod-io/gitpod/pull/9215
   - [x] `server`
     - [x] #9227
     - [x] #9297
     - [x] #9305
     - [x] remove `logLevel` from config, and take it from `LOG_LEVEL` env var
     - [x] make `defaultBaseImageRegistryWhitelist` configurable (or is this an imageBuilder concern?)
     - [x] make `definitelyGpDisabled` configurable
     - [x] make `disableDynamicAuthProviderLogin` configurable
     - [x] make `enableLocalApp` configurable
     - [x] `runDbDeleter`: = !dbSync.enabled
     - [x] `builtinAuthProvidersConfigured`: https://github.com/gitpod-io/gitpod/pull/9612
     - [x] payment
       - [x] ~add chargebee config to outer wrapper chart (cmp. [old](https://github.com/gitpod-io/ops/blob/gpl/webapp-installer/charts/gitpod_io/charts/gitpod/templates/chargebee-config-secret.yaml#L7-L18) [chart](https://github.com/gitpod-io/ops/blob/0ef3b446536cafa90598ae808fffe2b0aadc80c0/charts/gitpod_io/values.staging.yaml#L225-L233))~ already done by terraform
       - [x] add option to specify a secret to config.experimental.server.chargebeeConfig, and configure server to [find it](https://github.com/gitpod-io/gitpod/blob/948ae5192be0dd3d35431e6b0f5b443b1c61cfc1/install/installer/pkg/components/server/configmap.go#L142)
       - [x] set [server.config.EnablePayment](https://github.com/gitpod-io/gitpod/blob/948ae5192be0dd3d35431e6b0f5b443b1c61cfc1/install/installer/pkg/components/server/configmap.go#L140) when that secret is configured
     - [x] analytics: configure `GITPOD_ANALYTICS_WRITER` and `GITPOD_ANALYTICS_SEGMENT_KEY`
     - [x] license secret: set `GITPOD_LICENSE_TYPE` in wrapper chart as expected
[...]

PR #9321, opened 2022-04-14
Allow configuration of cross-cutting pod-level concerns to the installer

andrew-farries:
## Description

One of the Webapp team's [epics for Q2](https://github.com/gitpod-io/gitpod/issues/9097) is to use the Gitpod installer to deploy to Gitpod SaaS. In order to do that we may need to add additional configuration to the installer to make the output suitable for a SaaS deployment as opposed to a self-hosted deployment.

This PR adds a way to configure cross-cutting pod-level concerns like number of replicas, pod affinity and resource requests/limits. The idea is to add a new top-level `experimental` config section called `podConfig`, under which components can configure these details. 

## Related Issue(s)

Part of https://github.com/gitpod-io/gitpod/issues/9097

## How to test

Add an experimental section to the bottom of an installer config file like this:

```yaml
experimental:
  podConfig:
    server:
      replicas: 3
```

and invoke this installer:

```
installer render --config /path/to/config --use-experimental-config
```

The `server` deployment should then have the specified number of replicas set:

```yaml
- apiVersion: apps/v1
  kind: Deployment
  metadata:
    creationTimestamp: null
    labels:
      app: gitpod
      component: server
    name: server
  spec:
    replicas: 3
    ...
```

If the `experimental` config section is removed, the number of replicas defaults back to 1.

## Release Notes

```release-note
Add ability to configure cross-cutting pod-level concerns like replicas to the installer.
```

## Documentation

No changes required as this is experimental configuration for deploying Gitpod SaaS, not intended to be used by self-hosted installations.

### DIFF
--- a/install/installer/pkg/components/server/deployment.go
+++ b/install/installer/pkg/components/server/deployment.go
@@ -173,8 +173,7 @@ func deployment(ctx *common.RenderContext) ([]runtime.Object, error) {
 			},
 			Spec: appsv1.DeploymentSpec{
 				Selector: &metav1.LabelSelector{MatchLabels: labels},
-				// todo(sje): receive config value
-				Replicas: pointer.Int32(1),
+				Replicas: common.Replicas(ctx, Component),
 				Strategy: common.DeploymentStrategy,
 				Template: corev1.PodTemplateSpec{
 					ObjectMeta: metav1.ObjectMeta{
--- a/install/installer/pkg/config/v1/experimental/experimental.go
+++ b/install/installer/pkg/config/v1/experimental/experimental.go
@@ -10,13 +10,23 @@
 // If you use any setting herein, you forfeit support from Gitpod.
 package experimental
 
-import "k8s.io/apimachinery/pkg/api/resource"
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
 
 // Config contains all experimental configuration.
 type Config struct {
-	Workspace *WorkspaceConfig `json:"workspace,omitempty"`
-	WebApp    *WebAppConfig    `json:"webapp,omitempty"`
-	IDE       *IDEConfig       `json:"ide,omitempty"`
+	Workspace *WorkspaceConfig      `json:"workspace,omitempty"`
+	WebApp    *WebAppConfig         `json:"webapp,omitempty"`
+	IDE       *IDEConfig            `json:"ide,omitempty"`
+	PodConfig map[string]*PodConfig `json:"podConfig,omitempty"`
+}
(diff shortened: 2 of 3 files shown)


install/installer/pkg/config/v1/experimental/experimental.go
```diff
@@ -10,13 +10,23 @@
...
 
 // Config contains all experimental configuration.
 type Config struct {
-	Workspace *WorkspaceConfig `json:"workspace,omitempty"`
-	WebApp    *WebAppConfig    `json:"webapp,omitempty"`
-	IDE       *IDEConfig       `json:"ide,omitempty"`
+	Workspace *WorkspaceConfig      `json:"workspace,omitempty"`
+	WebApp    *WebAppConfig         `json:"webapp,omitempty"`
+	IDE       *IDEConfig            `json:"ide,omitempty"`
+	PodConfig map[string]*PodConfig `json:"podConfig,omitempty"`
```
geropl, line 23: :+1: leaving a pointer here bc we might want to discuss whether this is string-indexed or an explicit struct.

install/installer/pkg/components/server/deployment.go
```diff
@@ -173,8 +173,7 @@ func deployment(ctx *common.RenderContext) ([]runtime.Object, error) {
 			},
 			Spec: appsv1.DeploymentSpec{
 				Selector: &metav1.LabelSelector{MatchLabels: labels},
-				// todo(sje): receive config value
-				Replicas: pointer.Int32(1),
+				Replicas: common.Replicas(ctx, Component),
```
geropl, line 176: Once we agreed that this makes sense we should replicate it to `StatefulSet`s as well.

install/installer/pkg/config/v1/experimental/experimental.go
andrew-farries: I think string indexed is nice because every component defines its name via a `Component` constant, like here for the server: 

https://github.com/gitpod-io/gitpod/blob/db07326867158a3ed62c99a3e57310fb1f04e57f/install/installer/pkg/components/server/constants.go#L12

install/installer/pkg/components/server/deployment.go
andrew-farries: Every deployment/statefulset that needs to be able to specify its replica count via `experimental` config  will need this change - `Replicas` is hard-coded to `1` everywhere. This change to `server` is just to illustrate the pattern.

andrew-farries:
Closing this now that's served it's purpose in communicating what we intend to do. Will raise smaller PRs for each of the cross-cutting concerns:

* replicas
* pod affinity
* resources

PR #9333, opened 2022-04-14
[server] Cleanup (effectively) unused config: stage/KUBE_STAGE

geropl:
## Related Issue(s)

Related to this epic: https://github.com/gitpod-io/gitpod/issues/9097

## How to test
 - open a workspace on this branch
 - search the whole repo for these words, and make sure we no longer use them:
   - GITPOD_INSTALLATION_LONGNAME
   - theiaPluginsBucketNameOverride
   - stage (exact case, full word, limit to `*.tsx?` files)

## Release Notes

```release-note
none
```


components/server/src/server.ts
```diff
@@ -154,14 +153,6 @@ export class Server<C extends GitpodClient, S extends GitpodServer> {
...
             const verifyCSRF = (origin: string) => {
                 let allowedRequest = isAllowedWebsocketDomain(origin, this.config.hostUrl.url.hostname);
-                if (this.config.stage === "prodcopy" || this.config.stage === "staging") {
-                    // On staging and devstaging, we want to allow Theia to be able to connect to the server from this magic port
-                    // This enables debugging Theia from inside Gitpod
-                    const url = new URL(origin);
-                    if (url.hostname.startsWith("13444-")) {
-                        allowedRequest = true;
-                    }
-                }
```
andrew-farries, line 164: Does it matter that we are removing this debug port?

andrew-farries (commented):
There are still some references to `stage` in some Go files, at least:

* `install/installer/pkg/config/v1/experimental/experimental.go`
* `components/content-service/pkg/storage/gcloud.go`

components/server/src/server.ts
geropl: > Does it matter that we are removing this debug port?

AFAIK we have not clients anymore. This way conceived for Theia, and is not used anymore.

geropl:
> There are still some references to stage in some Go files, at least:

@andrew-farries That's right. This PR removes all usages of `(KUBE_)STAGE` for typescript components - especially `server` - but is meant to not touch the `content-service` dependency.

For that I created a [separate PR for that but closed it](https://github.com/gitpod-io/gitpod/pull/9339#issuecomment-1102366217), as:
 a) there no benefit for us nor anyone atm
 b) requires a deprecation cycle as we'd need to change public config surface in the installer
 c) has the potential to wreckage installations :boom:

andrew-farries (approved):

components/server/src/server.ts
geropl: Nope, this should not be in use anymore. It's theia-specific.

MrSimonEmms (approved):

PR #9336, opened 2022-04-14
Add some missing `component:` labels to various helm chart templates

andrew-farries:
## Description

Add missing component labels to the helm chart.

## Related Issue(s)

Relates to  #9097

## How to test

n/a

## Release Notes

```release-note
NONE
```

## Documentation

n/a


andrew-farries:
Not sure if we want the deletion of `chart/templates/auth-providers-configmap.yaml`. Also the addition of `name:` to the `agent-smith` daemonset is not consistent with the rest of the changes.

chart/templates/agent-smith-daemonset.yaml
```diff
@@ -28,6 +28,7 @@ spec:
         component: agent-smith
         kind: daemonset
         stage: {{ .Values.installation.stage }}
+      name: agent-smith
```
geropl, line 31: This is just a cheap change to make it more identical to the installer ouput. And it does no harm.

PR #9339, opened 2022-04-14
[content-service] Deprecate KUBE_STAGE/Stage, and replace it with bucketPrefix

geropl:
## Description
Based on: https://github.com/gitpod-io/gitpod/pull/9333
ToDo:
 - test
 - think about and discuss Self-Hosted implications
 - make adjacent ops changes (use it for prod/staging)

## Related Issue(s)

Fixes #

## Release Notes

```release-note
none
```

### DIFF
--- a/install/installer/pkg/config/v1/config.go
+++ b/install/installer/pkg/config/v1/config.go
@@ -153,6 +153,8 @@ type ObjectStorageS3 struct {
 type ObjectStorageCloudStorage struct {
 	ServiceAccount ObjectRef `json:"serviceAccount" validate:"required"`
 	Project        string    `json:"project" validate:"required"`
+	// BucketPrefix is an optional prefix that's prepended to all bucket names
+	BucketPrefix   string    `json:"bucketPrefix,omitempty"`
 }
 
 type ObjectStorageAzure struct {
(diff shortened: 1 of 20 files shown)


install/installer/pkg/config/v1/config.go
```diff
@@ -153,6 +153,8 @@ type ObjectStorageS3 struct {
 type ObjectStorageCloudStorage struct {
 	ServiceAccount ObjectRef `json:"serviceAccount" validate:"required"`
 	Project        string    `json:"project" validate:"required"`
+	// BucketPrefix is an optional prefix that's prepended to all bucket names
+	BucketPrefix   string    `json:"bucketPrefix,omitempty"`
```
csweichel, line 157: !! This is a backwards incompatible change. If a (self-hosted) user were to upgrade with this new version they would not find their backups anymore (`stage` was fixed to `production` before).

Please make sure that we maintain backwards compatibility.

geropl: Exactly, I'm aware. That's why I fixed that to "dev" (should indeed be "production").
The exact way how to ensure backwards compatibility will be discussed with Self-Hosted next week. Until then this will remain a draft.

geropl:
As this PR is not required atm and just and improvement, and has some potential to break stuff (see [comment](https://github.com/gitpod-io/gitpod/pull/9339/commits/519763ed124056946f3e8ead844d097594ed0014#r850627962)), we drop it for now.

Issue #7597, opened 2022-01-13
Epic: Admin dashboard enhancements to avoid SQL queries for support


jldec:
**Summary**
The goal of this epic is to enable the use of the admin dashboard, to resolve support scenarios which currently require SQL queries for inspection or direct SQL manipulation of data.

**Context**
Features such as teams, or projects and prebuilds cannot currently be inspected via the admin dashboard. This makes it difficult to support users in their use of these features.

Issues with infrastructure scheduling may result in database inconsistencies which can currently only be resolved through direct SQL manipulation of the affected records. One examples of this is #7255.

**Description**

general
- [x] #6932
- [x] #8454
- [ ] Admin date/time values should show exact date/time on hover

workspaces

- [x] https://github.com/gitpod-io/gitpod/issues/8005
- [ ] force backup
- [ ] force stop when stuck in stopping state
- [ ] list recent workspace instances with times, duration, status
- [ ] link-to and show prebuild workspaces
- [x] #8453
- [ ] Show list of instances under workspace detail
- [ ] Distinguish between prebuild and other kinds of workspaces in workspaces list 

user accounts

- [x] [fix account statement to download JSON](https://github.com/gitpod-io/gitpod/issues/7583) 
- [ ] #8451
- [ ] recover account e.g. after deleting the last authorized git provider account
- [ ] list projects under a user account

projects

- [x] #7636
  - [x] show project team or owner
  - [x] show recent project prebuilds below
- [ ] delete project
- [ ] restore project
- [ ] #8452
- [ ] show project configuration

teams

- [x] #8060
  - [x] list team members and owners below
  - [x] change roles between member and admin 
 - [ ] #8523
- [ ] list projects in a team

nice to have (not in scope for this epic)

- [ ] show team activity log (changes in memberships, projects)
- [ ] report team usage (workspace hours by project, team member)

**Value**
Visibility into Gitpod feature usage is important for customers and operators to understand how users are getting value from the platform.

geropl:
@jldec Once you're back: can you look into splitting this up in follow ups? :pray:

jldec:
Closing this because the Epic process requires epics to end.

- scheduled #8451 as urgent support followup.
- added #8452 to Epic #9291.
- treating other tasks including #8532 as lower priority for now.

laushinka:
[...]

Issue #7232, opened 2021-12-14
Consider differentiating external links from internal links in the dashboard


gtsiolis:
### Problem to solve

By default, any (internal) links that keep the user within Gitpod open in the same page rather than in a new tab or window. Currently, the dashboard includes internal links but also external links to repositories, branches, etc and this can sometime become confusing as :a: there's no clear differentiation of those links and the expected interaction as well as :b: the design of those links can sometimes be inconsistently the same as internal links which could surprise users. 🎁 

@AlexTugarev recently mentioned this during a deployment, see [relevant discussion](https://www.notion.so/gitpod/Meta-Deployment-Log-fb66ef0b8d47447095e017a1491475ca#b068e3ceaaa149e3be18b00cee0191df) (internal).

### Proposal

We could consider:
1. Limiting the amount of external links as much as possible, or especially avoid using the same design with internal links
2. Introducing an external links icon that can be shown at the end of the link text
3. Consolidating the design of the links so that it's a bit more clear when a link is an external link across the product

gtsiolis:
Reopening as this is still relevant.

PR #9343, opened 2022-04-14
[dashboard] add license tab to the admin dashboard

### LINKED ISSUE #8413: [licensing] Add a license page to the admin dashboard
Add a page to the admin dashboard that shows the installed license with properties (number allowed users, actual users, domain restriction etc.).

nandajavarma:
## Description

This PR adds a Licence page to the `admin` dashboard. This will help the self-hosted user to understand the features they have available, user limit, etc. The information that are currently available on the UI are:
* Licence level (Community, Professional, or Trial)
* Licence type (Replicated or Gitpod)
* List of features and the ones enabled
* User limit, current user count

Design specs for this can be found under the issue [8413](https://github.com/gitpod-io/gitpod/issues/8413#issuecomment-1089213673)

## Related Issue(s)

Fixes #8413

## How to test

`Admin` > `License` tab has been tested for the following scenarios on a self-hosted setup (it won't be available for SaaS):
- [Replicated community license](https://github.com/gitpod-io/gitpod/tree/main/install/licenses):
  - if they exceeded user limit(lesser features) or is within
- Replicated Professional license:
  - if they exceeded user limit(lesser features) or is within
- Gitpod paid license - [internal](https://www.notion.so/gitpod/Create-a-Gitpod-Self-Hosted-License-fef6b0e3548b4676a9d782f6d5a3df9a):
  - if they exceeded user limit(lesser features) or is within
- Installation with `installer` without a license key (default license)
  - if they exceeded user limit(lesser features) or is within
  
The testing can be done by manually inserting users to the `d_b_user` table of the `mysql` pod to exceed the user limits to test all scenarios.

## Release Notes

```release-note
[dashboard] add license tab to the admin dashboard
```

### DIFF
--- a/components/dashboard/src/admin/License.tsx
+++ b/components/dashboard/src/admin/License.tsx
@@ -1,0 +1,212 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { PageWithSubMenu } from "../components/PageWithSubMenu";
+import { adminMenu } from "./admin-menu";
+
+import { LicenseContext } from "../license-context";
+import { ReactElement, useContext, useEffect } from "react";
+import { getGitpodService } from "../service/service";
+
+import { ReactComponent as Alert } from "../images/exclamation.svg";
+import { ReactComponent as Success } from "../images/check-circle.svg";
+import { LicenseInfo } from "@gitpod/gitpod-protocol";
+import { ReactComponent as XSvg } from "../images/x.svg";
+import { ReactComponent as CheckSvg } from "../images/check.svg";
+import { ReactComponent as LinkSvg } from "../images/external-link.svg";
+import SolidCard from "../components/SolidCard";
+import Card from "../components/Card";
+
+export default function License() {
+    const { license, setLicense } = useContext(LicenseContext);
+
+    useEffect(() => {
+        if (isGitpodIo()) {
+            return; // temporarily disable to avoid hight CPU on the DB
+        }
+        (async () => {
+            const data = await getGitpodService().server.adminGetLicense();
+            setLicense(data);
+        })();
+    }, []);
+
+    const featureList = license?.enabledFeatures;
+    const features = license?.features;
+
+    // if user seats is 0, it means that there is no user limit in the license
+    const userLimit = license?.seats == 0 ? "Unlimited" : license?.seats;
+
+    const [licenseLevel, paid, statusMessage] = license ? getSubscriptionLevel(license) : defaultMessage();
+
+    return (
+        <div>
+            <PageWithSubMenu
+                subMenu={adminMenu}
+                title="License"
+                subtitle="License associated with your Gitpod installation"
+            >
+                <div className="flex flex-row space-x-4">
+                    <Card>
+                        {licenseLevel}
+                        {paid}
+                        <div className="mt-4 font-semibold text-sm">Available features:</div>
+                        <div className="flex flex-col items-start text-sm">
+                            {features &&
+                                features.map((feat: string) => (
+                                    <span className="inline-flex space-x-1">
+                                        {featureList?.includes(feat) ? (
+                                            <CheckSvg fill="currentColor" className="self-center mt-1" />
+                                        ) : (
+                                            <XSvg fill="currentColor" className="self-center h-2 mt-1" />
+                                        )}
+                                        <span>{capitalizeInitials(feat)}</span>
+                                    </span>
+                                ))}
+                        </div>
+                    </Card>
+                    <SolidCard>
+                        <div className="my-2">{statusMessage}</div>
+                        <p className="dark:text-gray-500 font-semibold">Registered Users</p>
+                        <span className="dark:text-gray-300 text-lg">{license?.userCount || 0}</span>
+                        <span className="dark:text-gray-500 text-gray-400 pt-1 text-lg"> / {userLimit} </span>
+                        <p className="dark:text-gray-500 pt-2 font-semibold">License Type</p>
+                        <h4 className="dark:text-gray-300 text-lg">{capitalizeInitials(license?.type || "")}</h4>
+                        <a
+                            className="gp-link flex flex-row mr-2 justify-end font-semibold space-x-2 mt-6"
+                            href="https://www.gitpod.io/self-hosted"
+                            target="_blank"
+                        >
+                            <span>Compare Plans</span>
+                            <div className="self-end">
+                                <LinkSvg />
+                            </div>
+                        </a>
+                    </SolidCard>
+                </div>
+            </PageWithSubMenu>
+        </div>
+    );
+}
+
+function capitalizeInitials(str: string): string {
+    return str
+        .split("-")
+        .map((item) => {
+            return item.charAt(0).toUpperCase() + item.slice(1);
+        })
+        .join(" ");
+}
+
+function getSubscriptionLevel(license: LicenseInfo): ReactElement[] {
+    switch (license.plan) {
+        case "prod":
+        case "trial":
+            return professionalPlan(license.userCount || 0, license.seats, license.plan == "trial", license.validUntil);
+        case "community":
+            return communityPlan(license.userCount || 0, license.seats, license.fallbackAllowed);
+        default: {
+            return defaultMessage();
+        }
+    }
+}
+
+function licenseLevel(level: string): ReactElement {
+    return <div className="text-white dark:text-black font-semibold mt-4"> {level}</div>;
+}
+
+function additionalLicenseInfo(data: string): ReactElement {
+    return <div className="dark:text-gray-500 text-gray-400 font-semibold text-sm">{data}</div>;
+}
+
+function defaultMessage(): ReactElement[] {
+    const alertMessage = () => {
+        return (
+            <span className="text-gray-600 dark:text-gray-50 flex font-semibold items-center">
+                <div>Inactive or unknown license</div>
+                <div className="flex justify-right my-4 mx-1">
+                    <Alert fill="grey" className="h-8 w-8" />
+                </div>
+            </span>
+        );
+    };
+
+    return [licenseLevel("Inactive"), additionalLicenseInfo("Free"), alertMessage()];
+}
+
+function professionalPlan(userCount: number, seats: number, trial: boolean, validUntil: string): ReactElement[] {
+    const alertMessage = (aboveLimit: boolean) => {
+        return aboveLimit ? (
+            <span className="text-red-700 dark:text-red-400 flex font-semibold items-center">
+                <div>You have exceeded the usage limit.</div>
+                <div className="flex justify-right my-4 mx-1">
+                    <Alert className="h-6 w-6" />
+                </div>
+            </span>
+        ) : (
+            <span className="text-green-600 dark:text-green-400 flex font-semibold items-center">
+                <div>You have an active professional license.</div>
+                <div className="flex justify-right my-4 mx-1">
+                    <Success className="h-8 w-8" />
+                </div>
+            </span>
+        );
+    };
+
+    const aboveLimit: boolean = userCount > seats;
+
+    const licenseTitle = () => {
+        const expDate = new Date(validUntil);
+        if (typeof expDate.getTime !== "function") {
+            return trial ? additionalLicenseInfo("Trial") : additionalLicenseInfo("Paid");
+        } else {
+            return additionalLicenseInfo(
+                "Expires on " +
+                    expDate.toLocaleDateString("en-DB", { year: "numeric", month: "short", day: "numeric" }),
+            );
+        }
+    };
+
+    return [licenseLevel("Professional"), licenseTitle(), alertMessage(aboveLimit)];
+}
+
+function communityPlan(userCount: number, seats: number, fallbackAllowed: boolean): ReactElement[] {
+    const alertMessage = (aboveLimit: boolean) => {
+        if (aboveLimit) {
+            return fallbackAllowed ? (
+                <div className="text-gray-600 dark:text-gray-50 flex font-semibold items-center">
+                    <div>No active license. You are using community edition.</div>
+                    <div className="my-4 mx-1 ">
+                        <Success className="h-8 w-8" />
+                    </div>
+                </div>
+            ) : (
+                <span className="text-red-700 dark:text-red-400 flex font-semibold items-center">
+                    <div>No active license. You have exceeded the usage limit.</div>
+                    <div className="flex justify-right my-4 mx-1">
+                        <Alert className="h-8 w-8" />
+                    </div>
+                </span>
+            );
+        } else {
+            return (
+                <span className="text-green-600 dark:text-green-400 flex font-semibold items-center">
+                    <div>You are using the free community edition.</div>
+                    <div className="flex justify-right my-4 mx-1">
+                        <Success fill="green" className="h-8 w-8" />
+                    </div>
+                </span>
+            );
+        }
+    };
+
+    const aboveLimit: boolean = userCount > seats;
+
+    return [licenseLevel("Community"), additionalLicenseInfo("Free"), alertMessage(aboveLimit)];
+}
+
+function isGitpodIo() {
+    return window.location.hostname === "gitpod.io" || window.location.hostname === "gitpod-staging.com";
+}
--- a/components/dashboard/src/admin/License.tsx
+++ b/components/dashboard/src/admin/License.tsx
@@ -79,7 +79,7 @@ export default function License() {
                             href="https://www.gitpod.io/self-hosted"
                             target="_blank"
                         >
-                            <span>Compare Plans</span>
+                            <span className="text-sm">Compare Plans</span>
                             <div className="self-end">
                                 <LinkSvg />
                             </div>
--- a/components/dashboard/src/components/SolidCard.tsx
+++ b/components/dashboard/src/components/SolidCard.tsx
@@ -1,0 +1,20 @@
+/**
+ * Copyright (c) 2021 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+function SolidCard(p: { className?: string; children?: React.ReactNode }) {
+    return (
+        <div
+            className={
+                "flex rounded-xl w-72 h-64 px-4 bg-gray-100 dark:bg-gray-800 text-gray-600 dark:text-gray-600" +
+                (p.className || "")
+            }
+        >
+            <span>{p.children}</span>
+        </div>
+    );
+}
+
+export default SolidCard;
--- a/components/dashboard/src/App.tsx
+++ b/components/dashboard/src/App.tsx
@@ -13,13 +13,14 @@ import { UserContext } from "./user-context";
 import { TeamsContext } from "./teams/teams-context";
 import { ThemeContext } from "./theme-context";
 import { AdminContext } from "./admin-context";
+import { LicenseContext } from "./license-context";
 import { getGitpodService } from "./service/service";
 import { shouldSeeWhatsNew, WhatsNew } from "./whatsnew/WhatsNew";
 import gitpodIcon from "./icons/gitpod.svg";
 import { ErrorCodes } from "@gitpod/gitpod-protocol/lib/messaging/error";
 import { useHistory } from "react-router-dom";
 import { trackButtonOrAnchor, trackPathChange, trackLocation } from "./Analytics";
-import { User } from "@gitpod/gitpod-protocol";
+import { LicenseInfo, User } from "@gitpod/gitpod-protocol";
 import * as GitpodCookie from "@gitpod/gitpod-protocol/lib/util/gitpod-cookie";
 import { Experiment } from "./experiments";
 import { workspacesPathMain } from "./workspaces/workspaces.routes";
@@ -77,6 +78,7 @@ const AdminSettings = React.lazy(() => import(/* webpackPrefetch: true */ "./adm
 const ProjectsSearch = React.lazy(() => import(/* webpackPrefetch: true */ "./admin/ProjectsSearch"));
 const TeamsSearch = React.lazy(() => import(/* webpackPrefetch: true */ "./admin/TeamsSearch"));
 const OAuthClientApproval = React.lazy(() => import(/* webpackPrefetch: true */ "./OauthClientApproval"));
+const License = React.lazy(() => import(/* webpackPrefetch: true */ "./admin/License"));
 
 function Loading() {
     return <></>;
@@ -150,6 +152,7 @@ function App() {
     const { teams, setTeams } = useContext(TeamsContext);
     const { setAdminSettings } = useContext(AdminContext);
     const { setIsDark } = useContext(ThemeContext);
+    const { setLicense } = useContext(LicenseContext);
 
     const [loading, setLoading] = useState<boolean>(true);
     const [isWhatsNewShown, setWhatsNewShown] = useState(false);
@@ -186,6 +189,9 @@ function App() {
                 if (user?.rolesOrPermissions?.includes("admin")) {
                     const adminSettings = await getGitpodService().server.adminGetSettings();
                     setAdminSettings(adminSettings);
+
+                    var license: LicenseInfo = await getGitpodService().server.adminGetLicense();
+                    setLicense(license);
                 }
             } catch (error) {
                 console.error(error);
@@ -367,6 +373,7 @@ function App() {
                     <AdminRoute path="/admin/teams" component={TeamsSearch} />
                     <AdminRoute path="/admin/workspaces" component={WorkspacesSearch} />
                     <AdminRoute path="/admin/projects" component={ProjectsSearch} />
+                    <AdminRoute path="/admin/license" component={License} />
                     <AdminRoute path="/admin/settings" component={AdminSettings} />
 
                     <Route path={["/", "/login"]} exact>
--- a/components/dashboard/src/admin/admin-menu.ts
+++ b/components/dashboard/src/admin/admin-menu.ts
@@ -22,6 +22,10 @@ export const adminMenu = [
         link: ["/admin/teams"],
     },
     {
+        title: "License",
+        link: ["/admin/license"],
+    },
+    {
         title: "Settings",
         link: ["/admin/settings"],
     },
--- a/components/dashboard/src/components/Card.tsx
+++ b/components/dashboard/src/components/Card.tsx
@@ -1,0 +1,20 @@
+/**
+ * Copyright (c) 2021 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+function Card(p: { className?: string; children?: React.ReactNode }) {
+    return (
+        <div
+            className={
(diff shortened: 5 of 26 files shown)


roboquat:

<details>

</details>

gtsiolis:
Looking at this now! 👀

components/dashboard/src/admin/License.tsx
```diff
@@ -0,0 +1,177 @@
...
+                subMenu={adminMenu}
+                title="License"
+                subtitle="License associated with your Gitpod Installation"
+            >
+                <div className="flex flex-row space-x-4">
+                    <Card className="bg-gray-800 dark:bg-white text-white dark:text-gray-400">
+                        <p className="text-white dark:text-black font-bold pt-4 text-sm"> {licenseLevel}</p>
+                        <p className="dark:text-gray-500">{paid}</p>
+                        <p className="text-gray-400 pt-4 pb-2">Available features:</p>
+                        {features &&
```
gtsiolis, line 57: **issue**: This is minor and could be ignored for this iteration, but these are fetched from ther backend but feels like there's a typo for some of them. Could we safely rename the following or create follow up issues for cleaning up these?

| BEFORE | AFTER |
|-|-|
| Prebuild | Prebuilds |
| Set Timeout | Timeouts |
| Snapshot | Snapshots |

components/dashboard/src/admin/License.tsx
```diff
@@ -0,0 +1,177 @@
...
+                                </div>
+                            ))}
+                    </Card>
+                    <Card className="bg-gray-200 dark:bg-gray-900 text-gray-600 dark:text-gray-600">
+                        <div className="text-gray-600 dark:text-gray-200 text-sm py-4 flex-row flex items-center">
+                            <div>{msg}</div>
+                            <div className="px-4">{getLicenseValidIcon(tick)}</div>
+                        </div>
+                        <p className="dark:text-gray-500">Registered Users</p>
+                        <span className="dark:text-gray-300 pt-1 text-lg">{license?.userCount || 0}</span>
```
gtsiolis, line 75: **question**: Could we align and fix the action button on bottom right area of the card as described in the design specs? Using flex layout could help or simply using some bottom margin here if that sounds safe. Your call. 🏓 

| BEFORE | AFTER |
|-|-|
| button-before | button-after |

components/dashboard/src/admin/License.tsx
```diff
@@ -0,0 +1,177 @@
...
+    return (
+        <div>
+            <PageWithSubMenu
+                subMenu={adminMenu}
+                title="License"
+                subtitle="License associated with your Gitpod Installation"
+            >
+                <div className="flex flex-row space-x-4">
+                    <Card className="bg-gray-800 dark:bg-white text-white dark:text-gray-400">
+                        <p className="text-white dark:text-black font-bold pt-4 text-sm"> {licenseLevel}</p>
```
gtsiolis, line 54: **suggestion**: Could we use here normal font size as described in the specs? Going with the `text-base` or removing the `text-sm` above that is inhereted could work. Your call. 🏀

components/dashboard/src/admin/License.tsx
```diff
@@ -0,0 +1,177 @@
...
+                                        <Tick className="h-4" />
+                                    ) : (
+                                        <Cross className="h-4 w-4 px-1" />
+                                    )}
+                                    {capitalizeInitials(feat)}
+                                </div>
+                            ))}
+                    </Card>
+                    <Card className="bg-gray-200 dark:bg-gray-900 text-gray-600 dark:text-gray-600">
+                        <div className="text-gray-600 dark:text-gray-200 text-sm py-4 flex-row flex items-center">
```
gtsiolis, line 70: **fyi**: Generally, text-sm should be scarcely used. Defaulting to `text-base` or inheriting the base font size when possible would be ideal. What do you think?

components/dashboard/src/admin/License.tsx
```diff
@@ -0,0 +1,177 @@
...
+        <div>
+            <PageWithSubMenu
+                subMenu={adminMenu}
+                title="License"
+                subtitle="License associated with your Gitpod Installation"
+            >
+                <div className="flex flex-row space-x-4">
+                    <Card className="bg-gray-800 dark:bg-white text-white dark:text-gray-400">
+                        <p className="text-white dark:text-black font-bold pt-4 text-sm"> {licenseLevel}</p>
+                        <p className="dark:text-gray-500">{paid}</p>
```
gtsiolis, line 55: **question**: Probably not the best line to add this comment, but curious what _Paid_ means under the _Professional_ license header. Could we use a more useful subtitle like the following?

> Expires on Dec 22, 2023

Not sure if this information is available though. If not, let's go with the existing chage.

gtsiolis (commented):
Thanks for adding this, @nandajavarma! ✨ 

Looks great so far! Added some comments below.

From UX perspective this is getting closer to the finish line. We should probably ask for a review from someone to take a closer look at the code before merging, too. 🏁

components/dashboard/src/admin/License.tsx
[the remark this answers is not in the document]
nandajavarma: @gtsiolis I added users from database without github login, etc. Hence, they are ignored by our users table, it was the best way for me to test.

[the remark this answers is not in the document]
gtsiolis: Ah, ok! Sounds good! Resolving this. 📕

components/dashboard/src/admin/License.tsx
nandajavarma: This info might be available in certain cases, like a `Trial` Licence from replicated. But yeah might not be available in the case of a professional licence. The amount they paid for the licence(as suggested in your initial design specs) is an information we currently cannot gather from the licence.

nandajavarma: May be in the case of Trial I can try and bring the expiration info, since the account is technically not paid for.

components/dashboard/src/admin/License.tsx
nandajavarma: This is a great question! I think I will create another issue for this and deal with it outside the scope, since it might require other code changes in the server/licensor components (I am getting the feature names from an enum in licensor). I don't know if I want to hack more on the front end for this.

components/dashboard/src/admin/License.tsx
```diff
@@ -0,0 +1,177 @@
...
+        default:
+            return <Alert fill="gray" className="h-8 w-8" />;
+    }
+}
+
+function isGitpodIo() {
+    return window.location.hostname === "gitpod.io" || window.location.hostname === "gitpod-staging.com";
+}
+
+function Card(p: { className?: string; children?: React.ReactNode }) {
```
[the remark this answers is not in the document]
nandajavarma: Yes exactly! I tried to make it work, but ended up becoming a bit hard to achieve with my very limited front end knowledge. I can move this card component next to the `SelectableCard` components and make a note to standardize in the future!

components/dashboard/src/admin/License.tsx
```diff
@@ -0,0 +1,177 @@
...
+import { PageWithSubMenu } from "../components/PageWithSubMenu";
+import { adminMenu } from "./admin-menu";
+
+import { LicenseContext } from "../license-context";
+import { ReactElement, useContext, useEffect } from "react";
+import { getGitpodService } from "../service/service";
+
+import { ReactComponent as Alert } from "../images/exclamation.svg";
+import { ReactComponent as Success } from "../images/tick.svg";
+import { ReactComponent as Tick } from "../images/check.svg";
```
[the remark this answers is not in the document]
nandajavarma: you can find it here https://gitpod.nanpod.fun/admin/license, I have granted you `admin` role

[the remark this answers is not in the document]
nandajavarma: Ah okay, I assumed we use icons. Now that you mention it, I think it might make it much simpler to use HTML codes for check and cross. I have changed the code to reflect this. LMK how the new changes look in design https://gitpod.nanpod.fun/admin/license and https://nvn-license-dash.staging.gitpod-dev.com/admin/license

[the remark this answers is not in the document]
nandajavarma: `x` is not being used in the latest update, instead the HTML codes `&#10003;` and `&#10007;` are being used. eg: https://gitpod.nanpod.fun/admin/license, https://nvn-license-dash.staging.gitpod-dev.com/admin/license

components/dashboard/src/admin/License.tsx
```diff
@@ -0,0 +1,177 @@
...
+                        <h4 className="dark:text-gray-300 pt-1 text-lg">{capitalizeInitials(license?.type || "")}</h4>
+                        <button
+                            type="button"
+                            onClick={(e) => {
+                                e.preventDefault();
+                                window.location.href = "https://www.gitpod.io/self-hosted";
+                            }}
+                            className="ml-2 float-right"
+                        >
+                            {license?.plan == "prod" ? "Contact Sales" : "Request License"}
```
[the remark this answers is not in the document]
nandajavarma: Agreed! updating!

components/dashboard/src/admin/License.tsx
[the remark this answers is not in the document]
gtsiolis: > I assumed we use icons.

Most of the times we go with a custom icon. ✔️ 

**issue**: Looks good overall on macOS but I'm concerned that going with this could slightly affect UX as rendering could look quite different on different browsers and operating systems. Do you know how this will render in Linux a system?

**suggestion**: Sounds safer to go with a custom icon taking also into consideration the increased complexity for supporting different icons for light and dark theme. Maybe we could also simply use the `dark:filter-invert` class which we're using elsewhere. Sorry for the back and forth, @nandajavarma! What do you think?

components/dashboard/src/admin/License.tsx
```diff
@@ -0,0 +1,177 @@
...
+}
+
+function professionalPlan(userCount: number, seats: number): string[] {
+    const aboveLimit: boolean = userCount > seats;
+    let msg: string, tick: string;
+    if (aboveLimit) {
+        msg = "You have exceeded the usage limit.";
+        tick = "red-cross";
+    } else {
+        msg = "You have an active professional license.";
```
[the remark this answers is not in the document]
gtsiolis: > Do you suggest we do it?

I'd think yes, following the design specs. In the next iterations we could improve how we surface this information or drop the idea of the colored text following the colored icon.

components/dashboard/src/admin/License.tsx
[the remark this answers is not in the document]
gtsiolis: See https://github.com/gitpod-io/gitpod/pull/9343#discussion_r854079292.

components/dashboard/src/admin/License.tsx
gtsiolis: **suggestion**: Since we're here could we also move the checkmark to that the right margin matches the right margin of the button at the bottom of this card? 😇

components/dashboard/src/admin/License.tsx
nandajavarma: updated it to reflect expires value for trial license

components/dashboard/src/admin/License.tsx
nandajavarma: I have made the changes!

components/dashboard/src/admin/License.tsx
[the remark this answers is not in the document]
nandajavarma: done

components/dashboard/src/admin/License.tsx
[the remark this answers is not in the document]
nandajavarma: moved [`Card` component to a file ](https://github.com/gitpod-io/gitpod/pull/9343/files#diff-6873878bfc81131ca2d4a0cc6e3165f6ad1bbefc418cfa89a0df56ba5e07b4beR1)

components/dashboard/src/admin/License.tsx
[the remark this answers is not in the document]
nandajavarma: agree! I have added the icons back! changed the `Check` icon to one you suggested. It should be [pretty uniform now.](https://nvn-license-dash.staging.gitpod-dev.com/admin/license)

components/dashboard/src/components/SolidCard.tsx
```diff
@@ -0,0 +1,20 @@
...
+ * Copyright (c) 2021 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+function SolidCard(p: { className?: string; children?: React.ReactNode }) {
+    return (
+        <div
+            className={
+                "flex rounded-xl text-base w-72 h-64 px-4 bg-gray-100 dark:bg-gray-900 text-gray-600 dark:text-gray-600" +
```
gtsiolis, line 11: **suggestion**: Skipped to mentioned this change in https://github.com/gitpod-io/gitpod/pull/9343#discussion_r853372863. 😅 

```suggestion
                "flex rounded-xl text-base w-72 h-64 px-4 bg-gray-100 dark:bg-gray-800 text-gray-600 dark:text-gray-600" +
```

components/dashboard/src/admin/License.tsx
```diff
@@ -0,0 +1,215 @@
...
+        return aboveLimit ? (
+            <>
+                <div className="text-red-600">You have exceeded the usage limit.</div>
+                <div className="flex justify-right my-4 mx-1">
+                    <Alert fill="red" className="h-6 w-6" />
+                </div>
+            </>
+        ) : (
+            <>
+                <div className="text-green-600">You have an active professional license.</div>
```
gtsiolis, line 152: **nitpick**: Adding below a different green color to use on the dark theme following the initial specs.

```suggestion
                <div className="text-green-600 dark:text-green-400">You have an active professional license.</div>
```

components/dashboard/src/admin/License.tsx
```diff
@@ -0,0 +1,215 @@
...
+                                        ) : (
+                                            <XSvg fill="currentColor" className="self-center h-2" />
+                                        )}
+                                        <span>{capitalizeInitials(feat)}</span>
+                                    </span>
+                                ))}
+                        </div>
+                    </Card>
+                    <SolidCard>
+                        <div className="text-gray-600 dark:text-gray-200 py-4 flex-row flex font-semibold items-center">
```
gtsiolis, line 70: **nitpick**: Moving this back would be great! Sorry if the suggestion in https://github.com/gitpod-io/gitpod/pull/9343#discussion_r853460822 was confusing, @nandajavarma! I meant to keep the top margin and padding as is but move the button after the text elements.

```suggestion
                        <div className="text-gray-600 dark:text-gray-200 py-2 flex-row flex font-semibold items-center">
```

components/dashboard/src/admin/License.tsx
```diff
@@ -0,0 +1,215 @@
...
+                        <h4 className="dark:text-gray-300 pt-1 text-lg">{capitalizeInitials(license?.type || "")}</h4>
+                        <div className="flex justify-end bottom-2 relative">
+                            <button
+                                type="button"
+                                onClick={(e) => {
+                                    e.preventDefault();
+                                    window.location.href = "https://www.gitpod.io/self-hosted";
+                                }}
+                            >
+                                <div className="font-semibold">Compare Plans</div>
```
gtsiolis, line 86: **issue**: Maybe it's worth opening this in a new tab since it's a link outside the dashboard. WDYT?

Eventually we could improve how we differentiate external links (see https://github.com/gitpod-io/gitpod/issues/7232) but that's a different discussion and out of the scope of this PR.

components/dashboard/src/admin/License.tsx
```diff
@@ -0,0 +1,215 @@
...
+}
+
+function communityPlan(userCount: number, seats: number, fallbackAllowed: boolean): ReactElement[] {
+    const alertMessage = (aboveLimit: boolean) => {
+        if (aboveLimit) {
+            return fallbackAllowed ? (
+                <>
+                    <div>No active license. You are using community edition.</div>
+                    <div className="flex justify-right my-4 mx-1">
+                        <Success fill="grey" className="h-8 w-8" />
```
gtsiolis, line 185: **question**: This is minor but could we safely use the same color for the icon as the text next to the icon here as seen in the specs?

- For the gray version, we could use `gray-600` for the light theme and `gray-50` for the dark theme.
- For the green version, we could use `green-600` for the light theme and `green-400` for the dark theme.
- For the red version, we could use  `red-700` for the light theme and `red-400` for the dark theme.

components/dashboard/src/admin/License.tsx
```diff
@@ -0,0 +1,215 @@
...
+                        <div className="flex justify-end bottom-2 relative">
+                            <button
+                                type="button"
+                                onClick={(e) => {
+                                    e.preventDefault();
+                                    window.location.href = "https://www.gitpod.io/self-hosted";
+                                }}
+                            >
+                                <div className="font-semibold">Compare Plans</div>
+                            </button>
```
corneliusludmann, line 87: @gtsiolis Don't we also have a link that looks like a button (`a` tag with `href` etc)? Since this is just a link to another page, in my opinion that would be preferable so that we can get rid of the `onClick` JavaScript event.

Advantages:
- Users can see the link target on hover.
- Users can open the page in a new tab.

components/dashboard/src/admin/License.tsx
```diff
@@ -0,0 +1,215 @@
...
+    const userLimit = license?.seats == 0 ? "Unlimited" : license?.seats;
+
+    const [licenseLevel, paid, statusMessage] = license ? getSubscriptionLevel(license) : defaultMessage();
+
+    return (
+        <div>
+            <PageWithSubMenu
+                subMenu={adminMenu}
+                title="License"
+                subtitle="License associated with your Gitpod Installation"
```
corneliusludmann, line 48: NIT: tiny typo
```suggestion
                subtitle="License associated with your Gitpod installation"
```

components/dashboard/src/admin/License.tsx
```diff
@@ -0,0 +1,215 @@
...
+        (async () => {
+            const data = await getGitpodService().server.adminGetLicense();
+            setLicense(data);
+        })();
+    }, []);
+
+    const featureList = license?.enabledFeatures;
+    const features = license?.features;
+
+    // if user seats is 0, it means that there is user limit in the license
```
corneliusludmann, line 38: NIT: Shouldn't it mean “no user limit”? :thinking: 
```suggestion
    // if user seats is 0, it means that there is no user limit in the license
```

gtsiolis (commented):
Thanks for the ping, @nandajavarma! 🏓 

This is almost at the finish line! 🏁 

Left a few more comments below.

components/dashboard/src/admin/License.tsx
```diff
@@ -0,0 +1,215 @@
...
+                    <Alert fill="grey" className="h-8 w-8" />
+                </div>
+            </>
+        );
+    };
+
+    return [licenseLevel("Inactive"), additionalLicenseInfo("Free"), alertMessage()];
+}
+
+function professionalPlan(userCount: number, seats: number, trial: boolean, validUntil: string): ReactElement[] {
```
corneliusludmann, line 141: In this code, the `validUntil` param is used for Trial licenses only. However, every license could have a limited validation (I expect that we usually issue 1-year licenses that need to be renewed). Right?

corneliusludmann (commented):
Thanks for the great work, Nandaja! :rocket: 

I left a few comments.

Together with the flow chart @lucasvaltl is working on about the different states the license could reach (valid, expired, too many users, wrong domain, etc.) we need to think about how we could test every case (unit tests?) and how the license page actually looks like for each case. But that's follow-up work.

components/dashboard/src/admin/License.tsx
corneliusludmann: Related: https://github.com/gitpod-io/gitpod/pull/9343#discussion_r855126635

gtsiolis:
> ... we need to think about how we could test every case (unit tests?) and how the license page actually looks like for each case. But that's follow-up work.

@corneliusludmann Triggered also from the comment in https://github.com/gitpod-io/gitpod/pull/9343#discussion_r853845336 (Cc @nandajavarma), I'd also like to better understand the options that exist and the matrix of possible use cases depending on which option a user picks to install Gitpod, so that we make the license page flexible to support all cases. Please, keep me in the loop or ping me in a follow up issue. 🏓

components/dashboard/src/admin/License.tsx
gtsiolis: Thanks for noticing, @corneliusludmann! The intent of using a primary (confirm) action button here was to provide at least one primary action users can take. However, since there is no self-serve flow (yet) where users can request or manage their license, going with a link sounds good.

We could keep the button as is and introduce an appending icon to indicate the external link or simply go with a plain `<a>` link. Hint: Using the `gp-link` class could help as we need a different blue color for the light and dark theme.

| Button | Link (Light) | Link (Dark) |
|-|-|-|
| Community | Community | Community-1 |


I've also updated the specs linked in https://github.com/gitpod-io/gitpod/issues/8413#issuecomment-1089213673 to include these new designs. Cc @nandajavarma

jldec:
Hi, would you mind making me (jldec) admin on the preview env so that I can look at the admin UI.
(happy to follow instructions if there is a way for me to do this myself)

corneliusludmann:
> Hi, would you mind making me (jldec) admin on the preview env so that I can look at the admin UI. (happy to follow instructions if there is a way for me to do this myself)

@jldec Open a Gitpod workspace from this PR and run `leeway run components:make-user-admin` ([source](https://www.notion.so/gitpod/How-we-develop-e8ce7e562478458a9223eb9b797415b2#74153318dfc74fb589d8ac12c080a187)).

jldec:
Here's what I see when I look at the preview env (I have narrowed my browser viewport). The comments are just first impressions to someone non-familiar with all the details - not intended as criticism 🙏. Generally speaking I do like the look of the design.

Screenshot 2022-04-21 at 21 59 59


- The 2 boxes are not responsive, and don't wrap when the viewport is narrower - I would have expected wrapping behavior given the design (cc: @gtsiolis)

- The baseline of the X's in the left box are not aligned with the text next to them 

- The vertical position of the "Compare Plans" link seems too high, leaving a large space between the text and the bottom of the box.

- (suggestion) I assume that the X's in the left box mean that the features are unavailable.  This is slightly unclear because some web forms show checked checkboxes with an X and unchecked checkboxes as empty but with the checkbox frame. It would be better if there were at least one item with a checkbox, to make it easier to understand the meaning of the X vs. check iconography. (cc: @gtsiolis)

- I interpret the X's in the following way: _all users have lost access to licensed features because the number of seats was exceeded_ -- Is this the intended interpretation?

- (suggestion) Is the 6/5 exceeded usage limit message correct? I guess this is meant to convey that this instance has a professional licence with 5 seats, and 6 users have registered. It would be clearer if the 5 purchased seats were also explicitly mentioned together with the "Professional" license description on the left, since that feels like one of the most important details. 

- I don't see the "Gitpod" License type in the [linked self-hosted license page](https://www.gitpod.io/self-hosted)) and can't really interpret what that means.

jldec:
I was going through another [admin epic](https://github.com/gitpod-io/gitpod/issues/7597), and noticed that the Admin Dashboard should no longer be limited to license holders. This means that Admin Dashboard could always be shown as checked in the left box :)

See https://github.com/gitpod-io/gitpod/pull/8593/commits/ebb90c600efd7adf72c087260cb0ef40648d2e89 and https://github.com/gitpod-io/gitpod/issues/6932#issuecomment-1062837883 - cc: @MrSimonEmms

nandajavarma:
@jldec The bit about `admin` dashboard always being shown is only true in the case of `community` license. As in, a person setting up self-hosted Gitpod with a `Starter` plan(until 10 users they get all the enterprise features) gets a fallback upon crossing the 10 user limit and they get only `admin` dashboard. This is not the case with a paid license. We do not offer fallback. We disable signups and do not offer any of the features(even admin dashboard).

corneliusludmann:
Whether the admin dashboard is available or not is outside of the scope of this pull request, in my opinion. We need to fix this on the license code level (if needed) and not in the UI.

Nevertheless, good heads up, @jldec! :pray: 

---

@lucasvaltl We need to consider this in our license evaluation work and check implementation vs. what we really want. Homepage tells currently that admin dashboard is a EE feature but I think Jürgen is right here that the admin dashboard is intentional always available (code-wise).

components/dashboard/src/admin/License.tsx
```diff
@@ -0,0 +1,215 @@
...
+        case "community":
+            return communityPlan(license.userCount || 0, license.seats, license.fallbackAllowed);
+        default: {
+            return defaultMessage();
+        }
+    }
+}
+
+function licenseLevel(level: string): ReactElement {
+    return <p className="text-white dark:text-black font-bold pt-4"> {level}</p>;
```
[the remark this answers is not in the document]
nandajavarma: Sorry about the confusion @gtsiolis , I believe I have fixed this [now](https://nvn-license-dash.staging.gitpod-dev.com/admin/license)

components/dashboard/src/admin/License.tsx
nandajavarma: I have changed it to be a link instead of a button according to your designs.

components/dashboard/src/admin/License.tsx
nandajavarma: done!

components/dashboard/src/admin/License.tsx
```diff
@@ -0,0 +1,212 @@
...
+                        <span className="dark:text-gray-300 text-lg">{license?.userCount || 0}</span>
+                        <span className="dark:text-gray-500 text-gray-400 pt-1 text-lg"> / {userLimit} </span>
+                        <p className="dark:text-gray-500 pt-2 font-semibold">License Type</p>
+                        <h4 className="dark:text-gray-300 text-lg">{capitalizeInitials(license?.type || "")}</h4>
+                        <a
+                            className="gp-link flex flex-row mr-2 justify-end font-semibold space-x-2 mt-6"
+                            href="https://www.gitpod.io/self-hosted"
+                            target="_blank"
+                        >
+                            <span>Compare Plans</span>
```
gtsiolis, line 82: **suggestion**: What do you think of using text-sm here or the parent <a> element so that link text is not larger than the rest of the information on this card as also match the line height to the height of the icon used, `20px`?

```suggestion
                            <span className="text-sm">Compare Plans</span>
```

| BEFORE | AFTER |
|-|-|
| link-before | link-after |

gtsiolis (approved):
Thanks @nandajavarma for tackling all pending discussions!

FWIW, regarding responsive layout mentioned in https://github.com/gitpod-io/gitpod/pull/9343#issuecomment-1111267703 is part of a larger topic affecting more parts of the dashboard, see https://github.com/gitpod-io/gitpod/issues/4050. In case there's anything we could do ad-hoc to improve the responsive layout in every PR that introduces new components or patterns we could do that! However, it's also ok or even preferred to leave this out of the scope of this PR.

UX in this PR now looks mergable! 🛹 

Do you think it's worth passing this to someone else to take a closer look at the code changes? 🏀

components/dashboard/src/admin/License.tsx
nandajavarma: love it!

jldec (approved):
Thank you @nandajavarma - this looks much more polished now.
You have addressed all of my feedback in your comments or in fixes.
LGTM

components/dashboard/src/admin/License.tsx
lucasvaltl: Hooking into this conversation: It would be great if we can match the naming here to what is on the pricing page - e.g. it’s workspace sharing here in the admin UI but shared workspaces in the pricing page. Not urgent but makes things tidier.

gtsiolis: Opened https://github.com/gitpod-io/gitpod/issues/9785 to track this discussion and avoid discussions on closed PRs. 🏓 Cc @nandajavarma @lucasvaltl

PR #9355, opened 2022-04-15
[ws-manager] add metrics to track initialize and finalize of workspaces

### LINKED ISSUE #9353: observability: add metrics so we can measure frequency and duration of tar backup and restore now (with tar).

sagor999:
## Description

Add metrics to track initialize (time it takes to download and untar backup for example or to load prebuild) and finalize (time it takes to create and upload backup of workspace).
This metrics will be used later on to compare them to PVC implementation.

## Related Issue(s)

Fixes #9353

## How to test

Spin up new workspace preview environment using installer from this PR
Create new workspace, create large file in workspace and stop workspace.
Then open that workspace again.
Port forward metrics port of ws-manager and look for new metrics.
Times reported will match the time it took to finalize or initialize the workspace.

## Release Notes

```release-note
[ws-manager] add metrics to track initialize and finalize of workspaces
```

### DIFF
--- a/components/ws-manager/pkg/manager/monitor.go
+++ b/components/ws-manager/pkg/manager/monitor.go
@@ -734,7 +734,7 @@ func (m *Monitor) initializeWorkspaceContent(ctx context.Context, pod *corev1.Po
 		// we are already initialising
 		return nil
 	}
-
+	t := time.Now()
 	err = retryIfUnavailable(ctx, func(ctx context.Context) error {
 		_, err = snc.InitWorkspace(ctx, &wsdaemon.InitWorkspaceRequest{
 			Id: workspaceID,
@@ -755,6 +755,12 @@ func (m *Monitor) initializeWorkspaceContent(ctx context.Context, pod *corev1.Po
 	} else {
 		err = handleGRPCError(ctx, err)
 	}
+	wsType := pod.Labels[wsk8s.TypeLabel]
+	hist, errHist := m.manager.metrics.initializeTimeHistVec.GetMetricWithLabelValues(wsType)
+	if errHist != nil {
+		log.WithError(errHist).WithField("type", wsType).Warn("cannot get initialize time histogram metric")
+	}
+	hist.Observe(time.Since(t).Seconds())
 	if err != nil {
 		return xerrors.Errorf("cannot initialize workspace: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


components/ws-manager/pkg/manager/monitor.go
```diff
@@ -959,6 +966,12 @@ func (m *Monitor) finalizeWorkspaceContent(ctx context.Context, wso *workspaceOb
 		}
 		break
 	}
+	wsType := api.WorkspaceType_name[int32(tpe)]
+	hist, errHist := m.manager.metrics.finalizeTimeHistVec.GetMetricWithLabelValues(wsType)
```
utam0k, line 970: This may be an elementary question, but why not use `err` here?

components/ws-manager/pkg/manager/monitor.go
```diff
@@ -916,6 +922,7 @@ func (m *Monitor) finalizeWorkspaceContent(ctx context.Context, wso *workspaceOb
 		backupError error
 		gitStatus   *csapi.GitStatus
 	)
+	t := time.Now()
```
jenting, line 925: Would it be better that calculate the start time at finalizeWorkspaceContent and have a defer function to calculate the end time? For example:
```go
func (m *Monitor) finalizeWorkspaceContent(ctx context.Context, wso *workspaceObjects) {
    t := time.Now()
    ...
    tpe, err := wso.WorkspaceType()
    if err != nil {
        tracing.LogError(span, err)
        log.WithError(err).Warn("cannot determine workspace type - assuming this is a regular")
        tpe = api.WorkspaceType_REGULAR
    }
    defer func() {
        wsType := api.WorkspaceType_name[int32(tpe)]
	hist, errHist := m.manager.metrics.finalizeTimeHistVec.GetMetricWithLabelValues(wsType)
	if errHist != nil {
		log.WithError(errHist).WithField("type", wsType).Warn("cannot get finalize time histogram metric")
	}
	hist.Observe(time.Since(t).Seconds())
    }()
    ...
}
```

components/ws-manager/pkg/manager/monitor.go
sagor999: you are right, that was copy paste mistake. fixed!

components/ws-manager/pkg/manager/monitor.go
sagor999: Yeah, I thought of that, but that would catch incorrect metrics in case of this `return` statement:
https://github.com/gitpod-io/gitpod/blob/054bd0be5eb1f9e3fff153cf49093c0f424cda35/components/ws-manager/pkg/manager/monitor.go#L929-L933
I only want to catch timing for workspaces where we actually did some work.

utam0k (approved):

Issue #9354, opened 2022-04-15
Issues accessing Fastly endpoints from within docker that runs in Gitpod


shaal:
### Bug description

Issues accessing Fastly endpoints from within docker that runs in Gitpod

When running `curl https://www.drupal.org/files/issues/2020-03-30/2759397-25.patch -vvv`, I am getting -
```
*   Trying 151.101.2.217:443...
* Connected to www.drupal.org (151.101.2.217) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*  CAfile: /etc/ssl/certs/ca-certificates.crt
*  CApath: /etc/ssl/certs
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
```
And then it hangs for a minute, and display
```
* OpenSSL SSL_connect: Connection reset by peer in connection to www.drupal.org:443 
* Closing connection 0
curl: (35) OpenSSL SSL_connect: Connection reset by peer in connection to www.drupal.org:443
```

### Steps to reproduce

* Open any Gitpod workspace (tested in US Gitpod workspace)
* Confirm 10 times that you can run `curl https://www.fastly.com` with no issues ✅ 
* `docker run -it debian:11`
* `apt update`
* `apt install curl -y`
* Confirm 10 times that you can run `curl https://www.fastly.com` with no issues

You'll notice that inside docker, sometimes it works perfect, and sometimes it just hang up

### Anything else?

(This was found because of new errors when running composer requests to drupal.org (which uses Fastly CDN) through DrupalPod.

This might be a temporary Fastly issue? but it's weird why I cannot replicate this in Gitpod itself (not inside docker inside Gitpod), nor I can replicate that on local machine with or without Docker.

shaal:
@misterfalcon posted this on Gitpod Discord, not sure if it's related or not -
https://discord.com/channels/816244985187008514/839379835662368768/964597530648252476
```
have a question that I also sent to support, but maybe someone here has some insight....
Outbound connections to google over https are not working all of a sudden (starting today) from within our docker
containers. These same docker containers are also running on EC2 instances that we manage and have setup through a
CI/CD pipeline, so effectively the same setup. We are not having any issues on our EC2 instances so we suspect there's
something going on with Gitpod. Yesterday we had no issues, and today we did with no change to our code or config.

These are strictly issues reaching google, which affects our API since that relies on making external calls to Firebase. All
[...]

PR #9356, opened 2022-04-15
[docker-up] Set the MTU using ceth0 value

aledbf:
## Description

The default dockerd MTU value is 1500. If the container interface value is lower than that, the behavior of the containers could end up in intermittent network behavior.

## Release Notes
```release-note
[docker-up] Set the MTU using ceth0 value
```

fixes https://github.com/gitpod-io/gitpod/issues/9354


components/docker-up/go.mod
```diff
@@ -4,9 +4,12 @@ go 1.17
 
 require (
 	github.com/opencontainers/runtime-spec v1.0.2
-	github.com/rootless-containers/rootlesskit v0.14.1
+	github.com/rootless-containers/rootlesskit v1.0.0
```
utam0k, line 7: 👍

install/installer/go.mod
```diff
@@ -21,6 +21,7 @@ require (
 	github.com/jetstack/cert-manager v1.4.4
 	github.com/replicatedhq/kots v1.67.0
 	github.com/spf13/cobra v1.2.1
+	github.com/stretchr/testify v1.7.0
```
utam0k, line 24: It does not appear to have anything to do with this PR 🤔

aledbf: correct, just a side effect of running `leeway run components:update-go-modules`

components/docker-up/docker-up/main.go
```diff
@@ -122,6 +123,14 @@ func runWithinNetns() (err error) {
 	}
 	args = append(args, userArgs...)
 
+	containerIf := "ceth0"
+	eth0, err := netlink.LinkByName(containerIf)
```
utam0k, line 127: I think the name `eth0` get confuse, how about this?
```suggestion
	interface, err := netlink.LinkByName(containerIf)
```

utam0k (commented):
@aledbf  Looks good with nits. Please take look at my comments.

components/docker-up/docker-up/main.go
aledbf: `interface` is a reserved word. Changed to `netIface`

components/docker-up/go.mod
kylos101: Initially this had me worried, but, after looking, [is maintenance.](https://github.com/rootless-containers/rootlesskit/compare/v0.14.6...v1.0.0)

utam0k (approved):
:100:

components/docker-up/docker-up/main.go
```diff
@@ -122,6 +123,14 @@ func runWithinNetns() (err error) {
...
 	args = append(args, userArgs...)
 
+	containerIf := "ceth0"
+	netIface, err := netlink.LinkByName(containerIf)
+	if err != nil {
+		return xerrors.Errorf("cannot get container network device %s: %w", containerIf, err)
+	}
+
+	args = append(args, fmt.Sprintf("--mtu=%v", netIface.Attrs().MTU))
+
```
kylos101, line 133: Is it worth while to run a perf test again, to see if we get similar to the results that @utam0k shared initially?

I ask because `netlink` is described as `Netlink is the interface a user-space program in linux uses to communicate with the kernel.`

kylos101 (commented):
Just a question and a comment

components/docker-up/docker-up/main.go
aledbf: > Is it worth while to run a perf test again,

No. We are only referencing the interface we already have by name. No changes were introduced.

aledbf:
The change fixes the issue with docker but docker-compose does not use the dockerd MTU value. 
We need to create a document indicating the required value
``` 
networks:                                
  default:                               
    driver: bridge                       
    driver_opts:                         
      com.docker.network.driver.mtu: 1440
```

aledbf:
Fix and test for docker-compose already merged https://github.com/gitpod-io/template-docker-compose/pull/4

sagor999 (approved):

csweichel (approved):
LGTM

jenting (approved):

MrSimonEmms (approved):

Issue #483, opened 2019-04-15
[theia] command to create a gitpod link for the current context


akosyakov:
As a user I want to add nice links and badges to README file for example. Right now i had to copy it from other repos, it would be nice to have a command to generate them.

jankeromnes:
Agreed! My current (completely broken) workflow is:

1. Click top-right avatar
2. "Share Workspace Snapshot"
3. Copy the Markdown code, but replace snapshot URL with what you want

(I'm aware that this creates a snapshot that's never used, but... it's convenient 😅)

svenefftinge:
https://github.com/gitpod-io/gitpod/issues/1023

PR #9366, opened 2022-04-17
Automatically block users that are running blacklisted workloads

Furisto:
## Description
Users that are running workloads listed in the **very** category will be blocked by agent-smith. This PR also adds support for testing blocking users in preview environments by adding a token and a user for agent-smith (only in preview environments). Before testing had to be done against gitpod.io.

## Related Issue(s)

Fixes [#483](https://github.com/gitpod-io/ops/issues/483)
https://github.com/gitpod-io/ops/pull/1863
https://github.com/gitpod-io/ops/pull/1862

## How to test
- Open workspace
- Run watch date (I have added watch to the list of banned executables in this preview environment)
-> Your workspace should be stopped after a few seconds and your account should have been blocked (in the preview environment, your gitpod.io account should not be affected)

## Notes for reviewers 
- Web App has been added as reviewer because I had to exclude one additional user (the new agent-smith user) from user count, otherwise the first actual user in the preview environment will not have the admin role.

## Release Notes
```release-note
Automatically block users that are running blacklisted workloads
```

### DIFF
--- a/components/gitpod-db/src/user-db.ts
+++ b/components/gitpod-db/src/user-db.ts
@@ -143,6 +143,8 @@ export type PartialUserUpdate = Partial<Omit<User, "identities">> & Pick<User, "
 
 export const BUILTIN_WORKSPACE_PROBE_USER_ID = "builtin-user-workspace-probe-0000000";
 
+export const BUILTIN_WORKSPACE_USER_AGENT_SMITH = "buildin-user-agent-smith-0000000";
+
 export interface OwnerAndRepo {
     owner: string;
     repo: string;
--- a/components/gitpod-db/src/user-db.ts
+++ b/components/gitpod-db/src/user-db.ts
@@ -143,7 +143,7 @@ export type PartialUserUpdate = Partial<Omit<User, "identities">> & Pick<User, "
 
 export const BUILTIN_WORKSPACE_PROBE_USER_ID = "builtin-user-workspace-probe-0000000";
 
-export const BUILTIN_WORKSPACE_USER_AGENT_SMITH = "buildin-user-agent-smith-0000000";
+export const BUILTIN_WORKSPACE_USER_AGENT_SMITH = "builtin-user-agent-smith-0000000";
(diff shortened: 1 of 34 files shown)


utam0k (approved):
❤️

components/gitpod-db/src/user-db.ts
```diff
@@ -143,6 +143,8 @@ export type PartialUserUpdate = Partial<Omit<User, "identities">> & Pick<User, "
 
 export const BUILTIN_WORKSPACE_PROBE_USER_ID = "builtin-user-workspace-probe-0000000";
 
+export const BUILTIN_WORKSPACE_USER_AGENT_SMITH = "buildin-user-agent-smith-0000000";
```
geropl, line 146: :heart:

atduarte: Should it be `builtin` instead of `buildin`?

geropl: nit: the first user is called `buil**t**in`. We should try to avoid future confusion and align the agent-smith user name :+1:

geropl:
Tested and works as advertised. Let's fix [this](https://github.com/gitpod-io/gitpod/pull/9366#discussion_r851965669) before merging, though.

@Furisto Also, we need to configure this for prod before we deploy this (webapp still runs a workspace stack from helm chart!)
Placing a hold here until we have that PR.
/hold

components/gitpod-db/src/user-db.ts
Furisto: Good catch!

Furisto:
[PR for helm chart update](https://github.com/gitpod-io/ops/pull/1863)
Username has been fixed.
@geropl PTAL

geropl (approved):

geropl:
/unhold
Will make sure we will merge the helm PR before deployment tomorrow. 👍

sagor999 (approved):

PR #9372, opened 2022-04-18
[server] disallow deauthorization if account is blocked

geropl:
## Description
Note: This is already deployed in prod.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/server/src/auth/authenticator.ts
+++ b/components/server/src/auth/authenticator.ts
@@ -151,7 +151,7 @@ export class Authenticator {
 
     async deauthorize(req: express.Request, res: express.Response, next: express.NextFunction) {
         const user = req.user;
-        if (!req.isAuthenticated() || !User.is(user)) {
+        if (!req.isAuthenticated() || !User.is(user) || user.blocked) {
             log.info({ sessionId: req.sessionID }, `User is not authenticated.`);
             res.redirect(this.getSorryUrl(`Not authenticated. Please login.`));
             return;
@@ -191,7 +191,7 @@ export class Authenticator {
             return;
         }
         const user = req.user;
-        if (!req.isAuthenticated() || !User.is(user)) {
+        if (!req.isAuthenticated() || !User.is(user) || user.blocked) {
             log.info({ sessionId: req.sessionID }, `User is not authenticated.`, { "authorize-flow": true });
             res.redirect(this.getSorryUrl(`Not authenticated. Please login.`));
             return;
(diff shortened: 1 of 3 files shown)


jankeromnes:
Will take a look 👀

components/server/src/auth/authenticator.ts
```diff
@@ -151,7 +151,7 @@ export class Authenticator {
 
     async deauthorize(req: express.Request, res: express.Response, next: express.NextFunction) {
         const user = req.user;
-        if (!req.isAuthenticated() || !User.is(user)) {
+        if (!req.isAuthenticated() || !User.is(user) || user.blocked) {
             log.info({ sessionId: req.sessionID }, `User is not authenticated.`);
             res.redirect(this.getSorryUrl(`Not authenticated. Please login.`));
```
jankeromnes, line 156: This is not the correct failure mode for blocked users, right?

jankeromnes (commented):
The `checkAndBlockUser` changes mostly look fine to me.

However, I'm unsure about the `deauthorize` changes. Seems a bit strange to tell logged in but blocked users to please log in. 😬 Also -- what is the benefit of preventing blocked users from disconnecting their Integrations?

geropl:
> what is the benefit of preventing blocked users from disconnecting their Integrations?

@jankeromnes That's the point of this PR - blocking blocked users from removeing their identities. :upside_down_face:

components/server/src/auth/authenticator.ts
geropl: > This is not the correct failure mode for blocked users, right?

Nope, but I think we have bigger problems atm then caring about UX for blocked users.

Let's take and note/follow up PR and merge this one asap. :pray:

jankeromnes: I wonder if a "cleaner" location for this fix might be:

https://github.com/gitpod-io/gitpod/blob/ed30d9696fd1c0ca5775a2d44d6b4a9e5709ba63/components/server/src/user/user-controller.ts#L147-L150

jankeromnes:
@geropl Do you think [this commit](https://github.com/gitpod-io/gitpod/commit/3b806e8edb61d57cbd51327e5cf6f8112afc49ec) could work? Should I push it to this PR?

I think this would be a bit cleaner. (It's not about blocked user UX, but keeping consistent failure modes.)

jankeromnes:
Works "as intended", but the failure mode looks a bit surprising:

| De-authorizing shows inaccurate error | Deleting shows nothing | Authorizing opens a silly pop-up |
| --- | --- | --- |
| Screenshot 2022-04-19 at 14 47 24 | Screenshot 2022-04-19 at 14 47 51 | Screenshot 2022-04-19 at 14 49 34 |

jankeromnes (approved):
PR looks good as is. 👍 Leaving the aforementioned [UX problems](https://github.com/gitpod-io/gitpod/pull/9372#issuecomment-1102610850) as follow-ups.

PR #9379, opened 2022-04-18
[installer] fix redis pvc name incorrect

mustard-mh:
## Description

Should assign to
https://github.com/gitpod-io/gitpod/blob/37435335ae86eef44eb9fee7f53f62b7e156c8e1/chart/templates/openvsx-proxy-statefulset.yaml#L78-L90

## Related Issue(s)

Fixes #

## How to test

See if preview env openvsx-proxy started

## Release Notes

```release-note
NONE
```


iQQBot (approved):
@mustard-mh  Good catch, thanks very much!

PR #9385, opened 2022-04-18
[server] Stop signup for blocklisted email domains

### LINKED ISSUE #6285: Filtered email domains are not working
### Bug description

I added entry to `d_b_email_domain_filter`. The entry was synced across DBs. However, new users from that domain name were able to signup.

### Steps to reproduce

I haven't reproduced it by explicitly adding a domain but I believe it should be:
1. Add an entry to the table
2. Try signing up with that domain email id

### Workspace affected

_No response_

### Expected behavior

Users are not allowed to signup

### Example repository

_No response_

### Anything else?

_No response_

geropl:
## Related Issue(s)

Fixes #6285 

## How to test
 - try to log-in to https://gpl-6285-edomains.staging.gitpod-dev.com/workspaces with gitpod.io user
 - see that you are being blocked:
image

## Release Notes

```release-note
re-enables blocklisting of email domains
```

### DIFF
--- a/components/server/ee/src/auth/email-domain-service.ts
+++ b/components/server/ee/src/auth/email-domain-service.ts
@@ -25,7 +25,7 @@ export class EMailDomainServiceImpl implements EMailDomainService {
 
     async isBlocked(email: string): Promise<boolean> {
         const { domain } = this.parseMail(email);
-        return !this.domainFilterDb.filter(domain);
+        return !(await this.domainFilterDb.filter(domain));
     }
 
     async hasEducationalInstitutionSuffix(email: string): Promise<boolean> {
--- a/components/server/src/analytics.ts
+++ b/components/server/src/analytics.ts
@@ -34,6 +34,7 @@ export async function trackSignup(user: User, request: Request, analytics: IAnal
         properties: {
             auth_provider: user.identities[0].authProviderId,
             qualified: !!request.cookies["gitpod-marketing-website-visited"],
+            blocked: user.blocked,
         },
     });
 }
--- a/components/server/src/user/user-controller.ts
+++ b/components/server/src/user/user-controller.ts
@@ -575,7 +575,9 @@ export class UserController {
             await this.userService.updateUserIdentity(user, additionalIdentity, additionalToken);
         }
 
-        // const { isBlocked } = tosFlowInfo; // todo@alex: this setting is in conflict with the env var
+        if (user.blocked) {
+            log.warn({ user: user.id }, "user blocked on signup");
+        }
 
         await this.userService.updateUserEnvVarsOnLogin(user, envVars);
         await this.userService.acceptCurrentTerms(user);
@@ -592,6 +594,7 @@ export class UserController {
         newUser.name = authUser.authName;
         newUser.fullName = authUser.name || undefined;
         newUser.avatarUrl = authUser.avatarUrl;
+        newUser.blocked = tosFlowInfo.isBlocked;
     }
 
     protected getSorryUrl(message: string) {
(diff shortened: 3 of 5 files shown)


components/server/src/user/user-controller.ts
```diff
@@ -575,7 +575,13 @@ export class UserController {
             await this.userService.updateUserIdentity(user, additionalIdentity, additionalToken);
         }
 
-        // const { isBlocked } = tosFlowInfo; // todo@alex: this setting is in conflict with the env var
+        const { isBlocked } = tosFlowInfo;
+        if (isBlocked) {
+            await TosFlow.clear(req.session);
+            await AuthFlow.clear(req.session);
+
+            res.status(403);
```
AlexTugarev, line 583: @geropl, this won't be sticky.
the user account was already created in L.567. 
there the called `updateNewUserAfterTos` was the original place to set `User.isBlocked` attribute from `tosFlowInfo.isBlocked`, AFAIR. reinstating this should do the trick.

also, sending response here should be followed by a return, otherwise there will be more attempts to send in the followings.

this is a confusing line of comment which was a left over from the blocking attempts on staging envs.

geropl: Talked to Alex , the result is pushed as condensed [commit](https://github.com/gitpod-io/gitpod/pull/9385/commits/2a113294d3cace825d19abfeb5d2a5644c381285).

components/server/src/analytics.ts
```diff
@@ -34,6 +34,7 @@ export async function trackSignup(user: User, request: Request, analytics: IAnal
         properties: {
             auth_provider: user.identities[0].authProviderId,
             qualified: !!request.cookies["gitpod-marketing-website-visited"],
+            blocked: user.blocked,
```
geropl, line 37: @jakobhero I added this property. Does that require any work from your side?

jankeromnes:
Stealing this review as Alex is on holidays 😁

jankeromnes:
> - try to log-in to https://gpl-6285-edomains.staging.gitpod-dev.com/workspaces with gitpod.io user
> - see that you are being blocked:

I did not get blocked 😬 (even though my `/account` shows a `@gitpod.io` primary email address)

Did this accidentally get redeployed, or am I doing something wrong?

csweichel:
@jankeromnes please check the DB to see if the branch is properly set up. @geropl, @aledbf and I are currently busy blocking miners. This would really really help.

Can you try and push this change forward as quickly as possible?

jankeromnes:
Okay, given the urgency, will push this forward.

geropl:
/werft run

:crossed_fingers:

components/server/src/user/user-service.ts
```diff
@@ -142,7 +152,8 @@ export class UserService {
                 this.config.blockNewUsers.passlist.some((e) => mail.endsWith(`@${e}`));
             const canPass = newUser.identities.some((i) => !!i.primaryEmail && emailDomainInPasslist(i.primaryEmail));
 
-            newUser.blocked = !canPass;
+            // blocked = if user already blocked OR is not allowed to pass
+            newUser.blocked = newUser.blocked || !canPass;
```
jankeromnes, line 156: you-shall-not-pass

components/server/ee/src/auth/email-domain-service.ts
```diff
@@ -25,7 +25,7 @@ export class EMailDomainServiceImpl implements EMailDomainService {
 
     async isBlocked(email: string): Promise<boolean> {
         const { domain } = this.parseMail(email);
-        return !this.domainFilterDb.filter(domain);
+        return !(await this.domainFilterDb.filter(domain));
```
jankeromnes, line 28: 😵

jankeromnes (approved):
Code looks good and works as intended!

But we should really rename `filter` to something much less subtle. 🤪

components/server/src/analytics.ts
jakobhero: hi @geropl, thanks for looping me in! this does not cause any issues downstream, i will update the tracking plan with the additional property.

PR #9391, opened 2022-04-19
make vm support ssh gateway

iQQBot:
## Description

This PR make harvester VM preview environment support  SSH Gateway

How to archive this?
1. we setup [RouterOS](https://mikrotik.com/software) both in GCP and harvester side
2. setup wireguard between GCP and harvester routeros
3. setup VPC route for `harvester cluster service IP range` to RouterOS GCP side
4. setup route in RouterOS to route this range IP to harvester side RouterOS
5. for every preview environment in harvester, create a `pod` for forwarding traffic and create a `loadbalancer` type `service` for get `public IP address`

This PR only do step 5, Everything else is already done in core-dev and harvester
see internal notion page 

[How to setup routeros in GCP side](https://www.notion.so/gitpod/How-to-setup-routeros-in-GCP-side-949aa866383646758acba76dbd2c2d54)
[How to setup routeros in harvester side](https://www.notion.so/gitpod/How-to-setup-routeros-in-harvester-side-465defa5e6fd4c039c9aef303178bca4)

TODO: 
- [ ] clean loadbalancer after VM delete

## Related Issue(s)

Fixes #

## How to test

run werft job in this PR
start a workspace, check ssh gateway is work

## Release Notes

```release-note
NONE
```

## Documentation

- [x] /werft with-vm=true

PR #9416, opened 2022-04-20
[werft] change routeros to kubectl port-forward

iQQBot:
## Description

In #9391 We introduce a way to use RouterOS to relay traffic to access the harvester internal sub-network

While the current solution is based on RouterOS we’re now evaluating if we want to have this in our tech stack in the long term or move to another routing/tunneling solution. 

This PR use `kubectl port-forward` to replace RouterOS

It has the following benefits
- the addition to our tech stack is very minimal
- we don’t need an additional `RouterOS` / `publicIP` on the Harvester side
- we don’t need an additional `vpc router` / `RouterOS` on the GCP side

## Related Issue(s)

Fixes #

## How to test

1. run werft job in `this PR` or `create a new branch base on this PR`
2. start a workspace, check ssh gateway is work

## Release Notes

```release-note
NONE
```

## Documentation

- [x] /werft with-vm=true

### DIFF
--- a/.werft/jobs/build/prepare.ts
+++ b/.werft/jobs/build/prepare.ts
@@ -85,9 +85,11 @@ function issueCertificate(werft: Werft, config: JobConfig) {
 function decideHarvesterVMCreation(werft: Werft, config: JobConfig) {
     if (shouldCreateVM(config)) {
         createVM(werft, config)
-        applyLoadBalancer({ name: config.previewEnvironment.destname })
     } else {
         werft.currentPhaseSpan.setAttribute("werft.harvester.created_vm", false)
+    }
+    if (config.withVM) {
+        applyLoadBalancer({ name: config.previewEnvironment.destname })
     }
     werft.done(prepareSlices.BOOT_VM)
 }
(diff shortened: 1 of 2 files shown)


.werft/vm/manifests.ts
```diff
@@ -156,11 +156,11 @@ spec:
     - name: ssh-gateway
       protocol: TCP
       port: 22
-      targetPort: 22
+      targetPort: 2200
```
iQQBot, line 159: change to this port in order to remove NET_BIND_SERVICE and NET_ADMIN cap

.werft/jobs/build/prepare.ts
```diff
@@ -85,10 +85,12 @@ function issueCertificate(werft: Werft, config: JobConfig) {
 function decideHarvesterVMCreation(werft: Werft, config: JobConfig) {
     if (shouldCreateVM(config)) {
         createVM(werft, config)
-        applyLoadBalancer({ name: config.previewEnvironment.destname })
     } else {
         werft.currentPhaseSpan.setAttribute("werft.harvester.created_vm", false)
     }
+    if (config.withVM) {
```
iQQBot, line 91: If the preview environment is running in a vm, you need to make sure the loadbalancer exists, regardless of whether the vm is recreated

.werft/vm/manifests.ts
meysholdt: nice, that's good security practice

meysholdt (approved):
Thank you for this PR, @iQQBot !

I could connect via SSH flawlessly. Also, the code changes LGTM.

PR #9422, opened 2022-04-20
[ws-man-bridge] Add started and completed metrics to track health

easyCZ:
## Description

Introduce metrics for when an update is received, and when an update is finished processing. The combination of these gives us something like <inputs> and <outputs> measurements for the system. This allows us to construct graphs where we can clearly see the following:
* How much data flows in
* How much data is completed
* How much of it errored
* How long it took to process

It's important to note these are seperate metrics, and should be. Sometimes a system doesn't output anything (due to error) and only reporting metrics on output would prevent us from capturing this case.

This PR also introduces a sub-method which does most of the existing logic, the original method is used to wrap the private one with metircs.

## Related Issue(s)

* Part of https://github.com/gitpod-io/gitpod/issues/8596

## How to test

* Workspace starts work
* Prebuilds work

## Release Notes

```release-note
NONE
```

## Documentation

NONE

/hold

### DIFF
--- a/components/ws-manager-bridge/src/bridge.ts
+++ b/components/ws-manager-bridge/src/bridge.ts
@@ -36,6 +36,7 @@ import { Configuration } from "./config";
 import { WorkspaceCluster } from "@gitpod/gitpod-protocol/lib/workspace-cluster";
 import { repeat } from "@gitpod/gitpod-protocol/lib/util/repeat";
 import { PreparingUpdateEmulator, PreparingUpdateEmulatorFactory } from "./preparing-update-emulator";
+import { performance } from "perf_hooks";
 
 export const WorkspaceManagerBridgeFactory = Symbol("WorkspaceManagerBridgeFactory");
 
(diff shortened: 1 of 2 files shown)


roboquat:

<details>

</details>

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -166,13 +167,52 @@ export class WorkspaceManagerBridge implements Disposable {
...
+            throw e;
+        } finally {
+            const durationMs = performance.now() - start;
+            this.prometheusExporter.reportWorkspaceInstanceUpdateCompleted(
+                durationMs / 1000,
+                writeToDB,
+                this.cluster.name,
+                status.spec.type,
+            );
+            log.info("Successfully completed WorkspaceInstance status update");
```
geropl, line 205: What about adding owi/logcontext to these log lines (like [here](https://github.com/gitpod-io/gitpod/pull/9422/files#diff-5c377b2cad37712ee6f512a499f1069f7641af5aab7dceaa7e8d033fc23d4780L197))?

geropl:
BTW: If a PR seems otherwise ready, but still has a "hold" label, I tend to be slightly put off to just start reviewing it.
Is there a reason this PR still has it? Would it make sense to add sth akin "make sure to remove the label before marking a PR as "review ready"" to our "how we work"? :thinking:

easyCZ:
> Is there a reason this PR still has it? Would it make sense to add sth akin "make sure to remove the label before marking a PR as "review ready"" to our "how we work"? 🤔

I want to land them myself, so that I'm in control of the order of PRs going in and can ensure I know what has landed and what hasn't. It's a personal workflow. I tried to change it such that hold is on PRs by default but didn't get consensus as it seems it'd interfere with workflows of others.

IMO, we'll need to move this down the line in particular with stacked PRs as it's unlikely others will have the right context to know when exactly changes should land. In particular with feature flags, there may be some additional validation the author may want to do ahead of landing the change (ensure prod is disabled in the feature, for example). This in general goes back to putting more responsiblity on the author of the PR, and less on the reviewer - trust by default.

geropl:
Got this while testing: 
:100: unrelated, but blocks testing.
image

easyCZ:
Thanks @geropl, will investigate. Sorry about that, preview looks to be broken more often than not..

components/ws-manager-bridge/src/bridge.ts
easyCZ: Added.

geropl:
/werft run with-vm=true

@easyCZ Got the same error, and just made very good experience with `with-vm=true`, so took the liberty to re-trigger :+1:

geropl (approved):
After minutes of image pull I was able to use my workspace :relieved:

easyCZ:
Thanks @geropl!

PR #9388, opened 2022-04-19
Fix the shell error

jenting:
## Description
Fix the shell outputs
```
return: can only `return' from a function or sourced script
```

## Related Issue(s)
N/A

## How to test
1. Manually move the `/home/gitpod/.config/gcloud/application_default_credentials.json` to `/home/gitpod/.config/gcloud/application_default_credentials.json.bak`
2. Unset the environment variable `GCP_ADC_FILE`
   ```shell
   unset GCP_ADC_FILE
   ```
4. Run the script ./scripts/setup-google-adc.sh

## Release Notes

```release-note
NONE
```

PR #9429, opened 2022-04-20
Revert "Fix the shell error"

mustard-mh:
## Description

Reverts gitpod-io/gitpod#9388

Fixes cannot open new terminal bug. this script is write for `.bashrc` if it's `exit 0` terminal will exit immediately after it open

## Related Issue(s)

Fixes #

## How to test

1. Delete your env var `GCP_ADC_FILE`
2. Open Gitpod with this branch
3. See if terminal created

## Release Notes

```release-note
NONE
```

### DIFF
--- a/scripts/setup-google-adc.sh
+++ b/scripts/setup-google-adc.sh
@@ -13,7 +13,7 @@ GCLOUD_ADC_PATH="/home/gitpod/.config/gcloud/application_default_credentials.jso
 if [ ! -f "$GCLOUD_ADC_PATH" ]; then
     if [ -z "$GCP_ADC_FILE" ]; then
         echo "GCP_ADC_FILE not set, doing nothing."
-        exit 0
+        return;
     fi
     echo "$GCP_ADC_FILE" > "$GCLOUD_ADC_PATH"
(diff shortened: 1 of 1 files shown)


iQQBot (approved):

jeanp413:
Thanks @mustard-mh 
@jenting to supress the error message you can use `return <x> 2> /dev/null` see: https://stackoverflow.com/a/25811163

jeanp413:
yeah but somehow is not defined in some cases like when Chris reported it :shrug:, anyway just adding `2> /dev/null` to the return statement will prevent showing `return: can only return from a function or sourced script` as it seems annoying :slightly_smiling_face:

jenting:
> Thanks @mustard-mh @jenting to supress the error message you can use `return <x> 2> /dev/null` see: https://stackoverflow.com/a/25811163

Thank you @mustard-mh to help fix 🙇‍♂️  I did not realize it has other side effects because this script is in `.bashrc`.

Since the current design, the script is in `.bashrc`, the workaround way is like @jeanp413 mentioned because the environment variable `GCP_ADC_FILE` is optional.

Issue #6707, opened 2021-11-15
Epic: Allow user to configure preferred IDE configurations as part of onboarding


loujaybee:
**Goal/Intended Outcome** 

Users are aware of JetBrains IDE’s when they first use Gitpod so that if they are a regular JetBrains user they can configure all future workspaces to use JetBrains. 

**Proposal**
- Introduce a prompt for onboarding users (e.g. first workspace start)
- Present users with the current alternative IDE (JetBrains, etc) options, but also allow to dismiss
- If possible, we can directly update the preferences (or alternatively routing to the preferences page)
- On subsequent workspace starts we revert to the existing flow / respect existing preferences

**Consideration(s)**
- Embedding of logo's might help user perception (but not mandatory as part of the design)
- We should see this work as iterative, implement an experiment, and we can adapt/alter based on the results

**Measurement ([Experiment Dashboard](https://eu.mixpanel.com/project/2482401/view/3024765/app/dashboards#id=3173296))**

- Increased conversion rate through the onboarding funnel
- Increase in retention for users who are more likely to use JetBrains as a daily driver
- Increased general adoption of the JetBrains feature

**Note:** Some dropoff to be expected by adding additional friction/step into the onboarding flow. 

**Out of scope**
- VS Code Desktop (unless v.easy)
- Vim / SSH flows
- Implementing as A/B test

**Issues**
- [x] https://github.com/gitpod-io/gitpod/issues/9433
- [x] https://github.com/gitpod-io/gitpod/issues/9534
- [x] https://github.com/gitpod-io/gitpod/issues/9791
- [x] https://github.com/gitpod-io/gitpod/issues/9790
- [x] https://github.com/gitpod-io/gitpod/issues/9792

~~https://github.com/gitpod-io/gitpod/issues/9520~~

See Figma [[1](https://www.figma.com/file/1ofpgCa3XI0F4zZKhcgiGg/%236707?node-id=0%3A1)]

svenefftinge:
I think generally it could make sense to ask the user when we detect certain IDE metafiles (e.g. `.idea`) in a project.
Including the "Don't ask me again for this project" or even "Don't ask me again for similar projects".
We can surface these settings in the user preferences so users could manage them there.

loujaybee:
Moving this under JetBrains, removing epic label for now.

andreafalzetti:
Shall we consider this Epic as complete? cc @loujaybee

loujaybee:
[...]

Issue #9323, opened 2022-04-14
Vim as an explicit IDE preference


loujaybee:
We currently have access to workspaces via SSH, however adding Vim would make it a first-class citizen as part of the Gitpod experience. Alongside this we should consider what we can do to ensure things like dotfiles are also setup, and that workspaces potentially already load with vim + other similar tools such as tmux. This would support experiments such as https://github.com/gitpod-io/gitpod/issues/6707 where vim configuration can be presented to the user as part of onboarding.

gtsiolis:
@loujaybee I haven't researched in depth about this yet, but here are some early designs on what I had in mind forhttps://github.com/gitpod-io/gitpod/issues/7713#issuecomment-1140534710. Cc @josepjaume

| Editors :a: | Editors :b: | Workspace Start |
|-|-|-|
| ![Editors](https://user-images.githubusercontent.com/120486/171058037-0814909d-cd9c-412c-a620-75dce43468d7.png) | <img width="1440" alt="Editors" src="https://user-images.githubusercontent.com/120486/171058041-68f86aaf-9cd0-4b29-8d76-91145e38b6fc.png"> | <img width="1440" alt="StartDropdown" src="https://user-images.githubusercontent.com/120486/171058363-54b74474-2569-443f-b021-9df8076da5c5.png"> |

josepjaume:
I love this @gtsiolis! I would personally go for the first option for consistency with the rest of the IDE options :). Can't wait to be able to use this!

fcsonline:
The first option looks great! And when the workspace starts, the user would be able to choose an application to open the ssh links instead Vscode.

loujaybee:
@fcsonline - what application(s) are you opening for your local terminal, currently? Do you open just one, or multiple applications (like your terminal) on desktop? Having this option certainly removes some friction for command-line users, but I'm also curious about what other desktop applications you're configuring outside of the IDE / editor itself.

axonasif:
> Right now none, but this would let me to associate a script to open a Terminal and enter to the Gitpod automatically. It would awesome.

That indeed would be very awesome!
I'm just thinking, in order for such a thing to work, gitpod will need to use a trigger a custom URL scheme registered on the users local machine to pass the ball to the user for deciding what to do 🤔

axonasif:
> @loujaybee: alongside this we should consider what we can do to ensure things like dotfiles are also setup.

[...]

Issue #8573, opened 2022-03-03
Open workspace in ad-hoc (per-workspace) editor or IDE


loujaybee:
Currently, if a user wants to use a different editor they are required to: 

1. Go to preferences
2. Restart a workspace

**Example use cases:** 
1. You want to use VS Code (web) editor for making "smaller" changes to documentation, markdown etc and having the in-browser functionality makes more sense. 
2. You use a mono-repo with multiple languages or technology stacks and need or want an IDE for multiple aspects of that tech stack. 
3. You have a mono-repo with multiple languages, but you also need the ability to have two editors or IDE's open for the same workspace for doing two different types of work simultaneously. 

This issue is to add a method to directly open a workspace using X editor, overriding the general settings, allowing the user to choose an editor/IDE at workspace start time. This will also speed up the workflow for opening VS Code Desktop users, who currently have to go to the browser, but then use the command palette to open VS Code Desktop. 

Relates to: 
- https://github.com/gitpod-io/gitpod/issues/6874
- https://github.com/gitpod-io/gitpod/issues/9434
- https://github.com/gitpod-io/browser-extension/pull/59
- https://github.com/gitpod-io/gitpod/issues/7525
- https://github.com/gitpod-io/gitpod/pull/9467

loujaybee:
Closing as duplicate of: https://github.com/gitpod-io/gitpod/issues/6874

loujaybee:
Since we now have the IDE preference check for new users, a simple way of implementing this could be to have an "always ask me" setting for IDE on workspace start, which is then shown for all new workspaces.

![image](https://user-images.githubusercontent.com/5528307/171586676-50fe7e8d-a8b8-4712-ab8f-b03d7aaaa4df.png)

gtsiolis:
Thanks for reopening this, @loujaybee! I also think this is still relevant.

I’d :a: take into account the potential entry points to nudge or allow users to select a different editor and :b: try to understand the real user need here to decide whether a _remember me_ option or an _always ask me_ option may be better long-term or as a MVC (minimum viable change) solution. :thought_balloon:  

🍊 🍊 🍊 🍊 

The option to ad-hoc start a workspace from the dashboard or the editor with a different editor sounds interesting and useful. Re-posting from https://github.com/gitpod-io/gitpod/issues/9131#issuecomment-1093000858:

> 1. Introduce a command palette action with two steps that allow users to first select workpace restart using a different editor and then select the editor before restarting the workspace.
[...]

PR #9432, opened 2022-04-20
[Dashboard] Increase IDE awareness during onboarding

### LINKED ISSUE #9433: Property aware users that they can choose IDEs with onboarding dashboard
Property aware users that they can choose IDEs with onboarding dashboard

Further opinion: Display recommended template project after they choose some IDE in this page (i.e Goland => template-golang-cli)

It's a IDE feature in our [workshop](https://www.notion.so/gitpod/IDE-Discovery-Increase-awareness-of-desktop-and-JetBrains-IDE-features-to-new-and-returning-users-c98914780b394101bde0fa4be4842de0#6d6dac56acae4054ae0e9ec0e98a7ae6)

See draft [Figma](https://www.figma.com/file/1ofpgCa3XI0F4zZKhcgiGg/%236707?node-id=0%3A1)

|Before|After (onboarding)| After 2(after onboarding)|
|:-:|:-:|:-:|
|image|image| image|

andreafalzetti:
## Description
An experiment to try to increase the awareness of other IDEs besides VS Code.

**Every selection from user will change User setting immediately**

The approach used: 

- Extract the "IDE selection" into a standalone component `<SelectIDE>`
- Introduced a new component `<SelectIDEModal>`
- ~~Extracted the util `updateUserIDEInfo` from `<SelectIDE>` to its own file~~
- Re-use the `<SelectIDE>` in both the settings page and in SelectIDEModal
- `<SelectIDEModal>` will appear in the workspaces page, if the user [isOnboardingUser](https://github.com/gitpod-io/gitpod/pull/9432/files#diff-957d9dd15b947e3969539ebd0dabcb8b40ce8b5564e5820cd9bb27d88be1dbc7R106)

## Related Issue(s)

Fixes #9433
Relates to #6707

## How to test

The first thing to test is that workspaces started without this new preference, will default to VS Code browser. If you want to only test the onboarding flow, skip all `[optional]` steps below.

1. Sign-up in [this preview environment](https://af-ide-disbb5b4bbe7a.staging.gitpod-dev.com/workspaces)
2. `[optional]` Before selecting an IDE, try creating a workspace appending the repo to the preview env URL. This will make sure that users who skip this selection, can still launch workspaces with `code` as default IDE.
3. `[optional]`Once the workspace starts, verify that VS Code Browser is the default IDE
4. `[optional]`Stop the workspace `gp stop` and go back to the dashboard and delete your workspace
5. `[optional]`Go back to the workspaces page
6. Choose your preferred IDE (e.g IntelliJ)
7. Start a workspace and verify that IntelliJ opens or options to download the JetBrains Gateway are presented

### Test Cases

1. Choose nothing, refresh page will see modal again
2. Choose nothing and click `continue` or `X`, DB go with `code latest:false`
3. Check `use latest` and click continue or `X`, DB go with `code latest:true`
4. Choose any IDE, DB will change

#### FYI: Check and reset your ide option
```mysql
# check
SELECT id, name, additionalData->'$.ideSettings.defaultIde', additionalData->'$.ideSettings.useLatestVersion' FROM d_b_user WHERE id = '<your_user_id>';

# reset
UPDATE d_b_user SET additionalData='{"ideSettings":{},"emailNotificationSettings":{"allowsChangelogMail":true,"allowsDevXMail":true,"allowsOnboardingMail":true}}' WHERE id = '<your_user_id>';
```

## Release Notes

```release-note
Prompt first-time users to choose their default IDE 
```

## Screenshots

|Light|Dark|
|:-:|:-:|
|Screenshot 2022-04-27 at 15 34 49|Screenshot 2022-04-27 at 15 35 32|

### DIFF
--- a/components/dashboard/src/settings/Preferences.tsx
+++ b/components/dashboard/src/settings/Preferences.tsx
@@ -4,22 +4,17 @@
  * See License-AGPL.txt in the project root for license information.
  */
 
-import { IDEOption, IDEOptions } from "@gitpod/gitpod-protocol/lib/ide-protocol";
-import { useContext, useEffect, useState } from "react";
-import InfoBox from "../components/InfoBox";
+import { useContext, useState } from "react";
 import { PageWithSubMenu } from "../components/PageWithSubMenu";
 import PillLabel from "../components/PillLabel";
 import SelectableCard from "../components/SelectableCard";
-import SelectableCardSolid from "../components/SelectableCardSolid";
-import Tooltip from "../components/Tooltip";
 import { getGitpodService } from "../service/service";
 import { ThemeContext } from "../theme-context";
 import { UserContext } from "../user-context";
 import getSettingsMenu from "./settings-menu";
 import { trackEvent } from "../Analytics";
 import { PaymentContext } from "../payment-context";
-import CheckBox from "../components/CheckBox";
-import { IDESettings } from "@gitpod/gitpod-protocol";
+import SelectIDE from "./SelectIDE";
 
 type Theme = "light" | "dark" | "system";
 
@@ -27,79 +22,6 @@ export default function Preferences() {
     const { user } = useContext(UserContext);
     const { showPaymentUI } = useContext(PaymentContext);
     const { setIsDark } = useContext(ThemeContext);
-
-    const migrationIDESettings = () => {
-        if (!user?.additionalData?.ideSettings || user.additionalData.ideSettings.settingVersion === "2.0") {
-            return;
-        }
-        const newIDESettings: IDESettings = {
-            settingVersion: "2.0",
-        };
-        const ideSettings = user.additionalData.ideSettings;
-        if (ideSettings.useDesktopIde) {
-            if (ideSettings.defaultDesktopIde === "code-desktop") {
-                newIDESettings.defaultIde = "code-desktop";
-            } else if (ideSettings.defaultDesktopIde === "code-desktop-insiders") {
-                newIDESettings.defaultIde = "code-desktop";
-                newIDESettings.useLatestVersion = true;
-            } else {
-                newIDESettings.defaultIde = ideSettings.defaultDesktopIde;
-                newIDESettings.useLatestVersion = ideSettings.useLatestVersion;
-            }
-        } else {
-            const useLatest = ideSettings.defaultIde === "code-latest";
-            newIDESettings.defaultIde = "code";
-            newIDESettings.useLatestVersion = useLatest;
-        }
-        user.additionalData.ideSettings = newIDESettings;
-    };
-    migrationIDESettings();
-
-    const updateUserIDEInfo = async (selectedIde: string, useLatestVersion: boolean) => {
-        const additionalData = user?.additionalData ?? {};
-        const settings = additionalData.ideSettings ?? {};
-        settings.settingVersion = "2.0";
-        settings.defaultIde = selectedIde;
-        settings.useLatestVersion = useLatestVersion;
-        additionalData.ideSettings = settings;
-        getGitpodService()
-            .server.trackEvent({
-                event: "ide_configuration_changed",
-                properties: settings,
-            })
-            .then()
-            .catch(console.error);
-        await getGitpodService().server.updateLoggedInUser({ additionalData });
-    };
-
-    const [defaultIde, setDefaultIde] = useState<string>(user?.additionalData?.ideSettings?.defaultIde || "");
-    const actuallySetDefaultIde = async (value: string) => {
-        await updateUserIDEInfo(value, useLatestVersion);
-        setDefaultIde(value);
-    };
-
-    const [useLatestVersion, setUseLatestVersion] = useState<boolean>(
-        user?.additionalData?.ideSettings?.useLatestVersion ?? false,
-    );
-    const actuallySetUseLatestVersion = async (value: boolean) => {
-        await updateUserIDEInfo(defaultIde, value);
-        setUseLatestVersion(value);
-    };
-
-    const [ideOptions, setIdeOptions] = useState<IDEOptions | undefined>(undefined);
-    useEffect(() => {
-        (async () => {
-            const ideopts = await getGitpodService().server.getIDEOptions();
-            // TODO: Compatible with ide-config not deployed, need revert after ide-config deployed
-            delete ideopts.options["code-latest"];
-            delete ideopts.options["code-desktop-insiders"];
-
-            setIdeOptions(ideopts);
-            if (!defaultIde || ideopts.options[defaultIde] == null) {
-                setDefaultIde(ideopts.defaultIde);
-            }
-        })();
-    }, []);
 
     const [theme, setTheme] = useState<Theme>(localStorage.theme || "system");
     const actuallySetTheme = (theme: Theme) => {
@@ -114,8 +36,6 @@ export default function Preferences() {
         setIsDark(isDark);
         setTheme(theme);
     };
-
-    const allIdeOptions = ideOptions && orderedIdeOptions(ideOptions);
 
     const [dotfileRepo, setDotfileRepo] = useState<string>(user?.additionalData?.dotfileRepo || "");
     const actuallySetDotfileRepo = async (value: string) => {
@@ -136,80 +56,9 @@ export default function Preferences() {
                 title="Preferences"
                 subtitle="Configure user preferences."
             >
-                {ideOptions && (
-                    <>
-                        {allIdeOptions && (
-                            <>
-                                <h3>Editor</h3>
-                                <p className="text-base text-gray-500 dark:text-gray-400">
-                                    Choose the editor for opening workspaces.
-                                </p>
-                                <div className="my-4 gap-4 flex flex-wrap max-w-2xl">
-                                    {allIdeOptions.map(([id, option]) => {
-                                        const selected = defaultIde === id;
-                                        const onSelect = () => actuallySetDefaultIde(id);
-                                        return renderIdeOption(option, selected, onSelect);
-                                    })}
-                                </div>
-                                {ideOptions.options[defaultIde]?.notes && (
-                                    <InfoBox className="my-5 max-w-2xl">
-                                        <ul>
-                                            {ideOptions.options[defaultIde].notes?.map((x, idx) => (
-                                                <li className={idx > 0 ? "mt-2" : ""}>{x}</li>
-                                            ))}
-                                        </ul>
-                                    </InfoBox>
-                                )}
-                                <p className="text-left w-full text-gray-500">
-                                    The <strong>JetBrains desktop IDEs</strong> are currently in beta.{" "}
-                                    <a
-                                        href="https://github.com/gitpod-io/gitpod/issues/6576"
-                                        target="gitpod-feedback-issue"
-                                        rel="noopener"
-                                        className="gp-link"
-                                    >
-                                        Send feedback
-                                    </a>{" "}
-                                    ·{" "}
-                                    <a
-                                        href="https://www.gitpod.io/docs/integrations/jetbrains"
-                                        target="_blank"
-                                        rel="noopener noreferrer"
-                                        className="gp-link"
-                                    >
-                                        Documentation
-                                    </a>
-                                </p>
-                            </>
-                        )}
-                        <CheckBox
-                            title="Latest Release (Unstable)"
-                            desc={
-                                <span>
-                                    Use the latest version for each editor.{" "}
-                                    <a
-                                        className="gp-link"
-                                        target="_blank"
-                                        href="https://code.visualstudio.com/blogs/2016/02/01/introducing_insiders_build"
-                                    >
-                                        Insiders
-                                    </a>{" "}
-                                    for VS Code,{" "}
-                                    <a
-                                        className="gp-link"
-                                        target="_blank"
-                                        href="https://www.jetbrains.com/resources/eap/"
-                                    >
-                                        EAP
-                                    </a>{" "}
-                                    for JetBrains IDEs.
-                                </span>
-                            }
-                            checked={useLatestVersion}
-                            onChange={(e) => actuallySetUseLatestVersion(e.target.checked)}
-                        />
-                    </>
-                )}
+                <h3>Editor</h3>
+                <p className="text-base text-gray-500 dark:text-gray-400">Choose the editor for opening workspaces.</p>
+                <SelectIDE />
                 <h3 className="mt-12">Theme</h3>
                 <p className="text-base text-gray-500 dark:text-gray-400">Early bird or night owl? Choose your side.</p>
                 <div className="mt-4 space-x-4 flex">
@@ -293,42 +142,4 @@ export default function Preferences() {
             </PageWithSubMenu>
         </div>
     );
-}
-
-function orderedIdeOptions(ideOptions: IDEOptions) {
-    // TODO: Maybe convert orderKey to number before sort?
-    return Object.entries(ideOptions.options)
-        .filter(([_, x]) => !x.hidden)
-        .sort((a, b) => {
-            const keyA = a[1].orderKey || a[0];
-            const keyB = b[1].orderKey || b[0];
-            return keyA.localeCompare(keyB);
-        });
-}
-
-function renderIdeOption(option: IDEOption, selected: boolean, onSelect: () => void): JSX.Element {
-    const label = option.type === "desktop" ? "" : option.type;
-    const card = (
-        <SelectableCardSolid className="w-36 h-40" title={option.title} selected={selected} onClick={onSelect}>
-            <div className="flex justify-center mt-3">
-                <img className="w-16 filter-grayscale self-center" src={option.logo} alt="logo" />
-            </div>
-            {label ? (
-                <div
-                    className={`font-semibold text-sm ${
-                        selected ? "text-gray-100 dark:text-gray-600" : "text-gray-600 dark:text-gray-500"
-                    } uppercase mt-2 px-3 py-1 self-center`}
-                >
-                    {label}
-                </div>
-            ) : (
-                <></>
-            )}
-        </SelectableCardSolid>
-    );
-
-    if (option.tooltip) {
-        return <Tooltip content={option.tooltip}>{card}</Tooltip>;
-    }
-    return card;
 }
--- a/components/dashboard/src/settings/SelectIDE.tsx
+++ b/components/dashboard/src/settings/SelectIDE.tsx
@@ -1,0 +1,194 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { IDEOption, IDEOptions } from "@gitpod/gitpod-protocol/lib/ide-protocol";
+import { useContext, useEffect, useState } from "react";
+import InfoBox from "../components/InfoBox";
+import SelectableCardSolid from "../components/SelectableCardSolid";
+import Tooltip from "../components/Tooltip";
+import { getGitpodService } from "../service/service";
+import { UserContext } from "../user-context";
+import CheckBox from "../components/CheckBox";
+import { User } from "@gitpod/gitpod-protocol";
+
+interface SelectIDEProps {
+    updateUserContext?: boolean;
+}
+
+export const updateUserIDEInfo = async (user: User, selectedIde: string, useLatestVersion: boolean) => {
+    const additionalData = user?.additionalData ?? {};
+    const settings = additionalData.ideSettings ?? {};
+    settings.settingVersion = "2.0";
+    settings.defaultIde = selectedIde;
+    settings.useLatestVersion = useLatestVersion;
+    additionalData.ideSettings = settings;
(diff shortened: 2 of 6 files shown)


roboquat:

<details>

</details>

components/dashboard/src/settings/Preferences.tsx
```diff
@@ -55,52 +51,6 @@ export default function Preferences() {
     };
     migrationIDESettings();
 
-    const updateUserIDEInfo = async (selectedIde: string, useLatestVersion: boolean) => {
```
andreafalzetti, line 58: I've extracted the logic to present & save the IDE preference into a dedicated component `<SelectIDE>`

components/gitpod-db/src/typeorm/user-db-impl.ts
```diff
@@ -101,7 +101,7 @@ export class TypeORMUserDBImpl implements UserDB {
             creationDate: new Date().toISOString(),
             identities: [],
             additionalData: {
-                ideSettings: { defaultIde: "code" },
+                ideSettings: {},
```
andreafalzetti, line 104: By removing `defaultIde: "code"` from the user default preferences, we can actually detect when the user actually select their `defaultIde`.

A workspace start will fallback anyway on `code` if the user does not select any preferred IDE.

On another note, we could entirely remove `ideSettings` instead of keeping it as empty object. My preference is being explicit for code readability purposes and keep it, but I am happy to get rid of it too, if we want to keep it clean.

components/dashboard/src/settings/SelectIDE.tsx
```diff
@@ -0,0 +1,187 @@
...
+            })
+            .then()
+            .catch(console.error);
+        await getGitpodService().server.updateLoggedInUser({ additionalData });
+    };
+
+    const [defaultIde, setDefaultIde] = useState<string>(user?.additionalData?.ideSettings?.defaultIde || "");
+    const actuallySetDefaultIde = async (value: string) => {
+        // force context to refresh
+        setUser({ ...user! });
```
andreafalzetti, line 45: This `hack` was required to make the components re-render due to state change. If anyone can suggests a nicer way, happy to change it

components/dashboard/src/workspaces/Workspaces.tsx
```diff
@@ -40,6 +43,8 @@ export default function () {
         })();
     }, [teams, location]);
 
+    const hasPreferredIde = typeof user?.additionalData?.ideSettings?.defaultIde !== "undefined";
```
mustard-mh, line 46: The first time I visit Perferences, `code` not selected. maybe we can init `hasPreferredIde`  before migrate (or make user setting copied) so that it’ll not affect your checking conditions and display of `SelectIDE`

andreafalzetti: we could show "code" as default but I am concerned that it creates more confusion, considering that if you are a new user, in the homepage you would see the same component without any default, but in Preferences, if you do this, you would see "Code" as your default.. 🤔 

any more thoughts from @loujaybee @gtsiolis?

mustard-mh:
Maybe we can also track here about how many people visit this IDE options page in dashboard etc

cc @loujaybee @andreafalzetti

components/dashboard/src/workspaces/Workspaces.tsx
mustard-mh: We can show both `dashboard` and `preference` with default `code`

components/dashboard/src/workspaces/Workspaces.tsx
```diff
@@ -144,30 +149,53 @@ export default function () {
...
-                                            <span className="opacity-60 hidden md:inline">
-                                                {StartWorkspaceModalKeyBinding}
+                    <>
+                        {!hasPreferredIde && (
+                            <div className="px-6 py-10 flex flex-col text-gray-400 border-t border-gray-200 dark:border-gray-800">
+                                <div className="flex flex-col items-center justify-center mx-auto">
+                                    <>
+                                        <h3 className="text-center pb-3">Choose your Default IDE</h3>
+                                        <p className="text-base text-gray-500 dark:text-gray-400">
+                                            Gitpod is using Code as the default IDE. You can choose a different IDE or
```
akosyakov, line 159: In user facing text we should use proper product names, i.e. VS Code, not Code

akosyakov: Let’s also prefer `editor` over IDE

components/dashboard/src/workspaces/Workspaces.tsx
```diff
@@ -144,30 +149,53 @@ export default function () {
...
+                                            use a custom IDE image.
```
akosyakov, line 160: Custom editor images are not exposed to any user right now

components/dashboard/src/workspaces/Workspaces.tsx
mustard-mh: Also, we can provide a interface in `User` namespace and implement it, so that others (me 🙈) can reuse the same logic.

We can add logic in `isOnboarding` if @loujaybee want to change onboarding condition
```ts
export namespace User {
    ...
    export function hasPerferredIde(user: User) {
        return !user?.additionalData?.ideSettings?.defaultIde;
    }
    export function isOnboarding(user: User) {
        return !hasPerferredIde(user);
    }
}
```

https://github.com/gitpod-io/gitpod/commit/24630fcacd7278bd1c047eb640d712a15a9065e9#diff-957d9dd15b947e3969539ebd0dabcb8b40ce8b5564e5820cd9bb27d88be1dbc7R98-R103

mustard-mh, line 46: Also, we can provide a interface in `User` namespace and implement it, so that others (me 🙈) can reuse the same logic.

We can add logic in `isOnboarding` if @loujaybee want to change onboarding condition
```ts
export namespace User {
    ...
    export function hasPerferredIde(user: User) {
        return !user?.additionalData?.ideSettings?.defaultIde;
    }
    export function isOnboarding(user: User) {
        return !hasPerferredIde(user);
    }
}
```

https://github.com/gitpod-io/gitpod/commit/24630fcacd7278bd1c047eb640d712a15a9065e9#diff-957d9dd15b947e3969539ebd0dabcb8b40ce8b5564e5820cd9bb27d88be1dbc7R98-R103

components/dashboard/src/workspaces/Workspaces.tsx
```diff
@@ -144,30 +149,53 @@ export default function () {
...
+                                                <a
+                                                    className="gp-link"
+                                                    href="https://www.gitpod.io/docs/getting-started/"
+                                                >
+                                                    Learn more
+                                                </a>
+                                            </div>
+                                            <span>
+                                                <button onClick={() => setIsStartWorkspaceModalVisible(true)}>
+                                                    New Workspace{" "}
```
mustard-mh, line 187: We also need to provide `New Workpsace` button here in some way if `!hasPerferredIde`

components/dashboard/src/settings/SelectIDE.tsx
```diff
@@ -0,0 +1,187 @@
...
+import SelectableCardSolid from "../components/SelectableCardSolid";
+import Tooltip from "../components/Tooltip";
+import { getGitpodService } from "../service/service";
+import { UserContext } from "../user-context";
+import CheckBox from "../components/CheckBox";
+
+export interface SelectIDEProps {
+    showLatest?: boolean;
+}
+
```
mustard-mh, line 19: `migrationIDESettings` should move to this file too.

And will do this https://github.com/gitpod-io/gitpod/pull/8783#discussion_r845141061 in the future

components/gitpod-db/src/typeorm/user-db-impl.ts
loujaybee: > By removing defaultIde: "code" from the user default preferences when we insert the user into the DB, we can actually detect when the user actually select their defaultIde.

Knowing when a user makes an explicit choice will be useful 🙏

loujaybee:
Thanks for raising the PR @andreafalzetti ! For the benefit of others looking into the PR, the decision was to proceed with an experiment on the dashboard as it was a simpler UX to update to get an indication of the appetite for users to change their editor/IDE preferences, and the impact of the change. Which comes at the cost of not impacting users who onboard in gitpod via direct URL, e.g. a context URL or "Open in Gitpod" button. 

We can look to extend this experiment with other extensions, such as:

- Per project IDE selection(s)
    - #8573
    - #6874
- Reduced friction for JetBrains / Desktop setup 
    - #8572
- Introduction of explicit Vim experience 
    - #9323

loujaybee:
Before we ship 🛹  , I think we should update the copy/experience to be more explicit to the user about: 

1. Letting the user know that IDE selection is the first step towards starting a workspace (i.e "lead" the user). So we avoid the potential of users being clicking to different pages confused about why they're seeing IDE settings.
1. **Optionally:** Letting the user know they can skip the step (implicitly or explicit)

A couple options...

**1) More explicit "onboarding" experience** 

e.g. where the UX has definitive "steps" where IDE selection is part 1. Here's a rough wireframe of what I mean, however I'd want to get @gtsiolis input about which UI components we can re-use for this type of experience. 


| Before      | After |
| ----------- | ----------- |
| image      | image |

**2) Update copy text** 

e.g. "lead the user" by making reference to this step being the first step in starting a workspace (this makes it clearer why the selection is on the "workspace" page, etc). Also, we currently tell the user that VS Code is the default, but we're not super clear what the user should do next.

| Before      | After |
| ----------- | ----------- |
| image      | image |

**3) Embed IDE choice into project selection**

e.g. We present the current start a workspace as it is currently is, but we let the user know that they'll be opening in VS Code (Browser) and grant them the opportunity to update that. 

I don't have a wireframe for this just yet... let me create one and I'll update the comment 🙏

loujaybee:
Small bug 🐛  - There seems to be some shared state between preference pages and this page. 

**Steps to replicate:** 
1. Login: https://af-ide-disbb5b4bbe7a.staging.gitpod-dev.com
1. Navigate to https://af-ide-disbb5b4bbe7a.staging.gitpod-dev.com/preferences (by clicking, no refresh)
1. Return to workspaces list

**Expected result:**  IDE preferences select is still shown.
**Actual result:** Start workspace button is shown.

andreafalzetti:
> Small bug 🐛 - There seems to be some shared state between preference pages and this page.

Did you actually start your first workspace? In that case, the IDE/editor selection is skipped

components/dashboard/src/workspaces/Workspaces.tsx
andreafalzetti: done 👍

components/dashboard/src/workspaces/Workspaces.tsx
andreafalzetti: the user can skip in this case, and the regular flow for opening new workspaces will be there

components/dashboard/src/App.tsx
```diff
@@ -271,6 +271,7 @@ function App() {
     useEffect(() => {
         if (user) {
             refreshSearchData("", user);
+            User.migrationIDESettings(user);
```
andreafalzetti, line 274: by calling this here, it affects both Workspaces and Preferences pages

gtsiolis:
Looking at this now! 👀

components/dashboard/src/workspaces/Workspaces.tsx
andreafalzetti: I think it's very confusing if we show a default while we ask for a selection

andreafalzetti: personally, I would leave unselected on the first screen, and, when we update the copy on the Preferences page, we can clarify that VS Code browser will be the default

components/gitpod-protocol/src/protocol.ts
```diff
@@ -128,6 +162,7 @@ export type IDESettings = {
     // DEPRECATED: Same with useDesktopIde.
     defaultDesktopIde?: string;
     useLatestVersion?: boolean;
+    skipIdeSelection?: boolean;
```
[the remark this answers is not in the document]
mustard-mh: I think we should show `code` as the default value on both the `Preference` and `Dashboard` pages, because without actually making any selections, workspaces will start with code. (code as the default option without modifying the user setting in DB)

In this case, when we click `skip`, it is equivalent to the user choosing code by default. So there is no need to add `skipIdeSelection`.

If we want to know how many people chose `skip`, we should add a `track event`

components/dashboard/src/workspaces/Workspaces.tsx
mustard-mh: Same here https://github.com/gitpod-io/gitpod/pull/9432#discussion_r858715892

> I think we should show `code` as the default value on both the `Preference` and `Dashboard` pages, because without actually making any selections, workspaces will start with code, code is the real default. (code as the default option without modifying the user setting in DB)
> 
> In this case, when we click `skip`, it is equivalent to the user choosing code by default. So there is no need to add `skipIdeSelection`.
> 
> If we want to know how many people chose `skip`, we should add a `track event`

components/gitpod-protocol/src/protocol.ts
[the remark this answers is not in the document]
andreafalzetti: We'll be able to know how many users skipped also with the new event `ide_configuration_skipped`.

I am not strongly against saving `defaultIde` in the DB the information, but at the moment I don't see the benefit of doing it? 

If the `defaultIde` is missing, we would default to `code`, as you said. With the benefit that in the DB we can clearly know who has consciously chose `code`.

Perhaps, could you try to elaborate more on why you think we should save this info in the DB? :)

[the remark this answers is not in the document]
mustard-mh: > In this case, when we click `skip`, it is equivalent to the user choosing code by default. So there is no need to add `skipIdeSelection`.

Skip means users like our default setting. So we store `code` as default

If not, will user see this prompt again if he refreshes page, 
=> if so, it's not a good experience.
=> if not, what should we start as IDE when user choose to skip and start a workspace?

IMO, `skip` means skip to use default, and don't alert this again.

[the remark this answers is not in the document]
andreafalzetti: I am assuming that if a user is happy with VS Code, they would click on it :) 

In my mind, "skip" means "I don't want to take a decision now" rather than "I want VS Code"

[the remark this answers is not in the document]
mustard-mh: I don't think this is resolved(still thought `skipIdeSelection` is not necessary). unresolved now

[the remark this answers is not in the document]
andreafalzetti: @mustard-mh I resolved it because there is not going to be A `Skip` button anymore.

components/dashboard/src/App.tsx
andreafalzetti: moved in `SelectIDE.tsx` 👍

components/dashboard/src/settings/SelectIDEModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+import { useState, useContext } from "react";
+import { Link } from "react-router-dom";
+import SelectIDE from "./SelectIDE";
+import Modal from "../components/Modal";
+
+import { getGitpodService } from "../service/service";
+import { UserContext } from "../user-context";
+
+export interface SelectEditorModal {
```
mustard-mh, line 15: Go with a close function so that other components can use it like `await modal`

components/dashboard/src/settings/SelectIDEModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+    showModal?: boolean;
+}
+
+export default function (props: SelectEditorModal) {
+    const [visible, setVisible] = useState(props.showModal ?? false);
+    const { user, setUser } = useContext(UserContext);
+
+    const handleContinue = async () => {
```
mustard-mh, line 23: We can use `SelectIDE` component here, and delete these code, since they already exist in `SelectIDE`

mustard-mh: Every user selection will change user setting in DB, we can just wait until they click `continue`. It will make code easier

components/dashboard/src/settings/SelectIDEModal.tsx
mustard-mh: And will be useful in this issue https://github.com/gitpod-io/gitpod/issues/9534

mustard-mh:
https://github.com/gitpod-io/gitpod/issues/9534 and https://github.com/gitpod-io/gitpod/issues/9433 both can be resolved in this PR, after this change https://github.com/gitpod-io/gitpod/pull/9432/commits/f66da4f7cf174c24c2928e80cf455a6cda1234bc#r859661101

components/dashboard/src/settings/SelectIDEModal.tsx
mustard-mh: Deal with re-use login in `SelectIDE` and will only be called if user not select any of them👍 cc @andreafalzetti

components/dashboard/src/settings/Preferences.tsx
andreafalzetti: `updateUserIDEInfo` was also moved to its own file, to allow reusability

AlexTugarev (commented):
Hi @andreafalzetti!

I tried to test this changes, but was not able to trigger the modal to be shown.

'Verified several times and re-created accounts on the preview env, and also deleted cookies in between. 

'Trying to debug I could see that `User.isOnboardingUser(user!) === true` in Workspaces.tsx initially and the modal's frame is about to render, but the next thing this is updated and some default IDE settings are set. Is that expected? I suppose this is not intended.

gtsiolis:
> I tried to test this changes, but was not able to trigger the modal to be shown.

> 'Verified several times and re-created accounts on the preview env, and also deleted cookies in between.

Same here! Cc @AlexTugarev

components/dashboard/src/settings/SelectIDE.tsx
```diff
@@ -0,0 +1,194 @@
...
+    const { user, setUser } = useContext(UserContext);
+
+    // Only exec once when we access this component
+    useEffect(() => {
+        user && User.migrationIDESettings(user);
+    }, []);
+
+    const actualUpdateUserIDEInfo = async (user: User, selectedIde: string, useLatestVersion: boolean) => {
+        const newUserData = await updateUserIDEInfo(user, selectedIde, useLatestVersion);
+        props.updateUserContext && setUser({ ...newUserData });
```
AlexTugarev, line 48: @andreafalzetti and @mustard-mh, please explain this, it's not clear what `updateUserContext` is for here. it looks like a switch to make changes on backend but not to reflect on frontend, it's hard to see why this is wanted w.r.t to UX.

components/dashboard/src/settings/SelectIDEModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+                    >
+                        user preferences
+                    </Link>
+                    .
+                </p>
+                <SelectIDE updateUserContext={false} />
+            </div>
+            <div className="flex justify-end mt-6">
+                <button onClick={handleContinue} className="ml-2">
+                    Continue
```
AlexTugarev, line 51: What is the Continue button for? 

When I start out with the modal, and instead of dismissing, directly navigate to e.g. /settings I see the changes being effective.

AlexTugarev: Also, if I dismiss via `esc` key any selection is seen as applied.

AlexTugarev (approved):
LGTM!

I tested the new onboarding modal several times and it worked as advertised.

components/dashboard/src/settings/SelectIDEModal.tsx
AlexTugarev: That's now clarified in the PR description.

components/gitpod-protocol/src/protocol.ts
```diff
@@ -95,6 +95,47 @@ export namespace User {
...
+            typeof user?.additionalData?.ideSettings?.defaultIde !== "undefined" ||
+            typeof user?.additionalData?.ideSettings?.useLatestVersion !== "undefined"
+        );
+    }
+
+    export function isOnboardingUser(user: User) {
+        return !hasPreferredIde(user);
+    }
+
+    export function migrationIDESettings(user: User) {
```
akosyakov, line 110: @mustard-mh @andreafalzetti Why did we pull it to protocol from dashboard?

components/dashboard/src/settings/SelectIDE.tsx
mustard-mh: 🙏  Will do this in following up PR that works for issue https://github.com/gitpod-io/gitpod/issues/6707

components/gitpod-protocol/src/protocol.ts
mustard-mh: @akosyakov This? https://github.com/gitpod-io/gitpod/pull/9432#discussion_r858918692

akosyakov: reads like you agree to put it in dashboard though, not in protocol

andreafalzetti: we agreed to put it in protocol

Issue #8274, opened 2022-02-17
Dashboard: Can't delete workspaces which failed to be created due to lack of memory on pod


felladrin:
### Bug description

After trying to open https://github.com/gitpod-io/gitpod repository on Gitpod and failing due to lack of memory, the workspaces that failed to open (it failed two times, as you can see in the screenshot below) got stuck in the dashboard, with “Failed” status. Then when I click the Delete Workspace button, it's not deleting them (triggering the error in the screenshot below).

<img width="1021" alt="Screenshot 2022-02-16 at 11 14 26" src="https://user-images.githubusercontent.com/418083/154478034-287a1a84-8df8-4135-9c9c-63ab3e199ac8.png">

![image (2)](https://user-images.githubusercontent.com/418083/154478027-899d62d1-fa03-4c91-99da-90c65fcfd541.png)

![image (1)](https://user-images.githubusercontent.com/418083/154478037-345d83ce-16e9-408d-a525-ae0fa78f9548.png)

### Steps to reproduce

- Try reproducing the OutOfMemory error by opening a big repository, like https://github.com/gitpod-io/gitpod
- After it failed to create, go to your dashboard, click the three-dots icon, and select "Delete Workspace"
- You'll see that nothing happens, and an error is displayed on Dev Tools Console.

### Workspaces affected

1. gitpodio-gitpod-wdvp0x65v3a
2. gitpodio-gitpod-px58o9iw73v

Note: Both workspaces were automatically deleted by the garbage collector on 2022-03-02. [[1](https://www.gitpod.io/docs/life-of-workspace/#garbage-collection)]

### Expected behavior

The workspace should disappear from my dashboard after clicking the Delete Workspace button.

### Example repository

I can't share the workspace either. As it failed to be created, when I click the Share button, it triggers the following error:

<img width="733" alt="image" src="https://user-images.githubusercontent.com/418083/154478498-2fcd5a2e-914a-4747-b17f-f994f6816b88.png">

sagor999:
I think this is for @gitpod-io/engineering-webapp to handle this case appropriately. 
I will remove workspace for now from this issue, but feel free to tag us in if needed.

geropl:
@felladrin How long did you wait between 1) workspace failed and 2) try to delete?
[...]

PR #9438, opened 2022-04-20
[ws-manager] fix workspace status flipping pending to deleted 

### LINKED ISSUE #8703: Accessing server API at gitpod.io from within workspaces might fail sometimes
### Bug description

Sometimes, when you start a workspace ~~around the same time as a Gitpod WebApp deployment (typically on Tuesdays and Thursdays around ~08:30 UTC)~~, it can get into a bad state.

Symptoms:
- The IDE theme is always Light (quite noticeable if you normally use Dark theme)
- The Git credential-helper doesn't work (e.g. if you try to push to a new branch, it will ask you for username and password)
- Some Gitpod features are missing (e.g. if you look for `Gitpod: Stop Workspace` in `F1`, you won't find it)

The only solution is to stop the workspace (e.g. from your workspaces list) and restart it.

Screenshot 2022-03-10 at 09 29 12

### Steps to reproduce

1. Start a workspace ~~on a Tuesday or Thursday around ~08:30 UTC~~
2. Try to push something to a branch
3. It will ask for username and password

### Workspace affected

gitpodio-gitpod-ykfphcb8pwd

### Expected behavior

1. The workspace should not land into such a bad state or require a restart

### Example repository

_No response_

### Anything else?

_No response_

sagor999:
## Description

Fixes an issue where workspace status can get flipped from PENDING to TERMINATED back and forth if workspace pod is stuck in pending state (can happen if we are scaling up). When this happens, ws-manager will delete the pod after 5s and try again. Deletion of the pod would trigger workspace status update to get flipped into TERMINATED state causing webapp to delete workspace token from db. 
This PR fixes this issue.

@geropl there is a small UX issue when pod is stuck in pending: it doesn't seem like there is any way to signal to stop the workspace. It will be stuck in pending with no way to stop it unless ws-manager will be able to schedule it.

## Related Issue(s)

Fixes #8703 

## How to test

Spin up workspace preview env
Start workspace and observe it starts normally
Stop workspace
Cordon the node
Start workspace again
Observe that ws-manager keeps re-creating the pod every 5s as it is stuck in pending state but dashboard will remain now on Preparing workspace phase. Also observe that ws-manager-bridge is now only getting status updates that indicate phase is pending.

## Release Notes

```release-note
[ws-manager] fix a bug when opening workspace you would be signed out from git and not able to do git commands
```

### DIFF
--- a/components/ws-manager/pkg/manager/manager.go
+++ b/components/ws-manager/pkg/manager/manager.go
@@ -290,6 +290,13 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
 		return nil, xerrors.Errorf("cannot create workspace pod: %w", err)
 	}
 
+	// remove annotation to signal that workspace pod was indeed created and scheduled on the node
+	err = m.markWorkspace(ctx, req.Id, deleteMark(attemptingToCreatePodAnnotation))
+	if err != nil {
+		clog.WithError(err).WithField("pod.Namespace", pod.Namespace).WithField("pod.Name", pod.Name).Error("failed to remove annotation after creating workspace pod. this will break things")
+		return nil, xerrors.Errorf("couldn't remove annotation after creating workspace pod: %w", err)
+	}
+
 	span.LogKV("event", "pod started successfully")
 
 	// all workspaces get a service now
--- a/components/ws-manager/pkg/manager/annotations.go
+++ b/components/ws-manager/pkg/manager/annotations.go
@@ -77,6 +77,9 @@ const (
 
 	// stoppedByRequestAnnotation is set on a pod when it was requested to stop using a StopWorkspace call
 	stoppedByRequestAnnotation = "gitpod.io/stoppedByRequest"
+
+	// attemptingToCreatePodAnnotation is set when ws-manager is trying to create pod and is removed when pod is successfully scheduled on the node
+	attemptingToCreatePodAnnotation = "gitpod.io/attemptingToCreate"
 )
 
 // markWorkspaceAsReady adds annotations to a workspace pod
--- a/components/ws-manager/pkg/manager/create.go
+++ b/components/ws-manager/pkg/manager/create.go
@@ -306,6 +306,7 @@ func (m *Manager) createDefiniteWorkspacePod(startContext *startWorkspaceContext
 		kubernetes.WorkspaceImageSpecAnnotation: imageSpec,
 		kubernetes.OwnerTokenAnnotation:         startContext.OwnerToken,
 		wsk8s.TraceIDAnnotation:                 startContext.TraceID,
+		attemptingToCreatePodAnnotation:         "true",
 		// TODO(cw): post Kubernetes 1.19 use GA form for settings those profiles
 		"container.apparmor.security.beta.kubernetes.io/workspace": "unconfined",
 		// We're using a custom seccomp profile for user namespaces to allow clone, mount and chroot.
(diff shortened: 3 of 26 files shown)


roboquat:

<details>

</details>

components/ws-manager/pkg/manager/manager.go
```diff
@@ -200,6 +200,9 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
 	}
 	span.LogKV("event", "pod description created")
 
+	// add an annotation to the pod to signal that ws-manager will now attempt to create this pod
+	pod.Annotations[attemptingToCreatePodAnnotation] = "true"
```
csweichel, line 204: this should be part of the `createDefiniteWorkspacePod` and corresponding tests (`cd pkg/manager && go test -update -force .`)

components/ws-manager/pkg/manager/manager.go
```diff
@@ -290,6 +293,33 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
...
+		if err != nil {
+			clog.WithError(err).WithField("pod.Namespace", pod.Namespace).WithField("pod.Name", pod.Name).Error("was unable to get pod to remove annotation")
+			return err
+		}
+		delete(tempPod.Annotations, attemptingToCreatePodAnnotation)
+		err = m.Clientset.Update(ctx, &tempPod)
+		if err != nil {
+			return err
+		}
+		return nil
```
csweichel, line 316: this should use `markWorkspace` instead which carries the concurrent modification retry logic

components/ws-manager/pkg/manager/manager.go
sagor999: I disagree. :sweat_smile: 
Adding it into `createDefiniteWorkspacePod` will lose some context as to why it is added, and also this whole retry logic is a temporary workaround for OutOfMemory error. Once we have 1.23.6 k8s and confirm it is working, we might consider removing retry logic all together.
Another reason not to add this into `createDefiniteWorkspacePod` is that it is not obvious that someone has to remove that annotation when pod is created. So if that function will be used later somewhere else, it will have unexpected side effects for someone who is using it.

sagor999:
@csweichel PTAL :pray:

geropl:
> @geropl there is a small UX issue when pod is stuck in pending: it doesn't seem like there is any way to signal to stop the workspace. It will be stuck in pending with no way to stop it unless ws-manager will be able to schedule it.

@sagor999 Agreed, thx for hinting: Will add a note to #8274

components/ws-manager/pkg/manager/manager.go
csweichel: I understand the intent to keep the code together - it is however a break in current style.

Today, all pod creation is covered by the very same createDefiniteWorkspacePod (CDWP) function and tests. I.e. no modifications happen outside of the initial pod struct production prior to creating it in Kubernetes. Adding this annotation outside of the regular pod creation path means we have no coverage that it's added, and we can no longer be sure that what our fixtures say will be used is what's actually used.

`createDefiniteWorkspacePod` only makes sense in the path of creating a pod, and is very unlikely to be used elsewhere.

Also, I don't think this workaround will go away any time soon. When `1.23.6` comes around, we might be able to upgrade in SaaS. But self-hosted users might face the very same problem still. IMHO this code is here to stay for a good while longer.

I have no hard feelings on this (would be ok either way), but a clear preference towards making this part of CDWP for the reasons outlined above.

sagor999: @csweichel I see your point. Fair enough, code and tests updated.

csweichel (approved):

geropl:
:tada: :rocket: :pray:

geropl:
Is there an ETA for this? I guess sometime next week?

PR #9440, opened 2022-04-20
[ws-daemon] Support config reload for IO limits

csweichel:
## Description
This PR adds support for hot config reloading of the IO bandwidth limiting in ws-daemon.

## How to test
- change the config
- check if the v1 IO limits change

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/ws-daemon/cmd/root.go
+++ b/components/ws-daemon/cmd/root.go
@@ -5,8 +5,6 @@
 package cmd
 
 import (
-	"bytes"
-	"encoding/json"
 	"fmt"
 	"os"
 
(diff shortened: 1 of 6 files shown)


aledbf (approved):

sagor999:
why not use fsnotify to watch config file for changes?

csweichel:
Kubernetes configMap changes aren't just a file modification, but a delete/move of a symlink. This is hard to keep track of using fsnotify. That said, 💯 agree: fsnotify is the way to go.

We're not using fsnotify here because:
- we wanted to get this change out, and registry-facade has proven working code which we took over,
- in registry-facade prior attempts at making this work with fsnotify had failed, hence this implementation.

PR #9446, opened 2022-04-21
[server, dashboard] Refactor User.getPrimaryEmail to return "string | undefined" instead of throwing an error

### LINKED ISSUE #8359: Request getWorkspace failed with message: No identity with primary email for user: [...]!
### Bug description

A recent incident affected all members of the Gitpod team and caused starting workspaces to fail with:

```
Request getWorkspace failed with message: No identity with primary email for user: [...]!
```

No identity with primary email for user

This looks like a come back of this issue: https://github.com/gitpod-io/gitpod/issues/6035

### Steps to reproduce

1. Be part of a team where another team member has no identity in `d_b_identity` (which should itself not really happen)
2. Starting a workspace always fails with the above error until the no-identity member is removed again

### Workspace affected

_No response_

### Expected behavior

Several expectations were unmet:
- A user should never have zero identities in `d_b_identity` (already mostly addressed by https://github.com/gitpod-io/gitpod/pull/7946, but I believe this issue reveals another way this can happen)
- If `User.getPrimaryEmail(member)` throws an error for even just one member in `TeamDB.findMembersByTeam(teamId)`, it should not make the entire method fail. Instead, it should probably catch the error, and set `primaryEmail` to `(unknown)`.

https://github.com/gitpod-io/gitpod/blob/2479e485496cd29af37870f1be5bd4e56561d56d/components/gitpod-db/src/typeorm/team-db-impl.ts#L59

### Example repository

_No response_

### Anything else?

The user doesn't remember doing anything in [/integrations](https://gitpod.io/integrations).

Instead, here is what we believe happened:
1. User created a Gitpod account (let's call it `5773b3a1`)
2. User joined the Gitpod team
3. User noticed the wrong primary email displayed in the Members list
4. User removed themselves from the team again
5. User changed their primary email on GitHub, and added their `@gitpod.io` email as well
6. User tried to join the team again, which seems to have worked
7. (User added some user environment variables)

However, for some reason:
- Somewhere along this flow, a new Gitpod account was created (let's call it `71f05ca8`), and it somehow "took over" or "stole" the GitHub identity of the account `5773b3a1` (which was itself left with zero identities in the DB)
- We know steps 6. and 7. were completed as the old account (`5773b3a1`), thus adding a member without any identity or email address to the Gitpod team (breaking getWorkspace for everyone in the Gitpod team)

Questions:
- Why did Gitpod create a new account?
- Can this possibly happen if you create an account, then log out, then change your primary email in GitHub, then log in again? 🤔
- Or maybe if you use a different browser to change the GitHub primary email and log into Gitpod, thus leaving your other browser still logged into the now-identity-less account?

geropl:
## Related Issue(s)

Fixes #8359 

## How to test
 - signup with two users: 1st becomes admin, 2 is our test account
 - login with 2, and remove all git integrations. Verify in the DB that there are no more indentities for this account (if there still are, just delete them from the DB
 - login with 1, and open the user admin page for 2
 - verify that you see more than a blank page

## Release Notes

```release-note
fixed a bug around email address rendering
```

### DIFF
--- a/components/dashboard/src/admin/UserDetail.tsx
+++ b/components/dashboard/src/admin/UserDetail.tsx
@@ -50,7 +50,7 @@ export default function UserDetail(p: { user: User }) {
     }, [p.user]);
 
     const email = User.getPrimaryEmail(p.user);
-    const emailDomain = email.split("@")[email.split("@").length - 1];
+    const emailDomain = email ? email.split("@")[email.split("@").length - 1] : undefined;
 
     const updateUser: UpdateUserFunction = async (fun) => {
         setActivity(true);
@@ -62,6 +62,11 @@ export default function UserDetail(p: { user: User }) {
     };
 
     const addStudentDomain = async () => {
+        if (!emailDomain) {
+            console.log("cannot add student's email domain because there is none!");
+            return;
+        }
+
         await updateUser(async (u) => {
             await getGitpodService().server.adminAddStudentEmailDomain(u.id, emailDomain);
             await getGitpodService()
@@ -221,7 +226,9 @@ export default function UserDetail(p: { user: User }) {
                             <Property
                                 name="Student"
                                 actions={
-                                    !isStudent && !["gmail.com", "yahoo.com", "hotmail.com"].includes(emailDomain)
+                                    !isStudent &&
+                                    emailDomain &&
+                                    !["gmail.com", "yahoo.com", "hotmail.com"].includes(emailDomain)
                                         ? [
                                               {
                                                   label: `Make '${emailDomain}' a student domain`,
--- a/components/dashboard/src/Menu.tsx
+++ b/components/dashboard/src/Menu.tsx
@@ -379,7 +379,7 @@ export default function Menu() {
                             <ContextMenu
                                 menuEntries={[
                                     {
-                                        title: (user && User.getPrimaryEmail(user)) || "",
+                                        title: (user && (User.getPrimaryEmail(user) || user?.name)) || "User",
                                         customFontStyle: "text-gray-400",
                                         separator: true,
                                     },
--- a/components/dashboard/src/admin/UserSearch.tsx
+++ b/components/dashboard/src/admin/UserSearch.tsx
@@ -5,7 +5,6 @@
  */
 
 import { AdminGetListResult, User } from "@gitpod/gitpod-protocol";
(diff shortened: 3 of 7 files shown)


easyCZ (commented):
It feels like in various places, we treat an empty or undefined email differently. Sometimes we use empty string, sometimes we use "---", sometimes we keep it as undefined. Is it at all possible to choose one? In practice, we could encode that logic inside `getPrimaryEmail()` and not have to cast or convert in so many places.

geropl:
> Sometimes we use empty string, sometimes we use "---", sometimes we keep it as undefined. Is it at all possible to choose one?

I think in general it is not possible, because depending on your context, this has different meaning. `getPrimaryEmail` is also not about rendering, but status retrieval.

Wrt to the different options to render in the dashboard("---" vs ""): I basically kept the previous decisions in place, scoping this PR solely on the reliability parts to never break the renderning.
If we want to align this I'm all open for a follow-up.

components/dashboard/src/admin/UserDetail.tsx
```diff
@@ -50,7 +50,7 @@ export default function UserDetail(p: { user: User }) {
     }, [p.user]);
 
     const email = User.getPrimaryEmail(p.user);
-    const emailDomain = email.split("@")[email.split("@").length - 1];
+    const emailDomain = email ? email.split("@")[email.split("@").length - 1] : undefined;
```
easyCZ, line 53: [nit] Nothing to fix here, just a common. This is in general quite fishy, but oh well. In practice, it's extremely hard to parse email addresses well as the spec is so very broad, the best we can often do is check the email is valid by sending an email.

easyCZ (approved):
LGTM. I'd personally split this across 2 PRs - one for server and one for dashboard to keep the changes independent and better communicate nature of distributed systems but this is only a suggestion.

components/dashboard/src/admin/UserDetail.tsx
geropl: Absolutely agree in theory. Currently the only assumptions we make is that a) it has a '@' and b) non-empty domain part.
I know that both are not guaranteed by the spec, but held in practice well enough.
And then there is the other point that we should not be doing this randomly in the frontend, but should try to keep in a central place (or library).... :+1:

PR #9449, opened 2022-04-21
[installer] reduce openvsx-proxy request memory

mustard-mh:
## Description

this PR reduce the Resource requirement for `openvsx-proxy` and `redis` 2.25G looks like too big

As you can see, in a production environment, the memory of `openvsx` is basically stable at 150MB, so a request of 150MB is reasonable.

image

same with redis
image

## Related Issue(s)

Fixes #

## How to test

no need

## Release Notes

```release-note
NONE
```


iQQBot (approved):
Thanks @mustard-mh

PR #9455, opened 2022-04-21
[kots]: add an installation status pod

### LINKED ISSUE #9195: Configure a full status checker for installations
This ticket covers two related problems we have with the [KOTS status informers](https://docs.replicated.com/vendor/admin-console-display-app-status).

1. We make use of `DaemonSets` and there's no support for this (yet). See [Slac] discussion](https://gitpod.slack.com/archives/C02SWJ49JQ2/p1643651665487759)
2. As we're using the Installer to generate the manifests, the status won't be correctly reflected in updates.

To explain `2` further, consider the workflow. For an installation, the `statusInformers` will report the status correctly as they will either be healthy or not. For an upgrade, they don't because, as far as KOTS is concerned, the resources exist.

## Proposed solution

1. Add a randomly generated annotation/label for all Installer resources. This will be the same for all resources.
2. Create an application that is able to search for the existence of a resource _INCLUDING_ the annotation/label
3. Add this new deployment to the list of `statusInformers`

## Resources to support

Resource in _italics_ are not currently supported by KOTS `statusInformers`

 - deployment
 - _daemonset_
 - _statefulset_
 - service

## Other considerations

There may already be an open source application which does this. If so, this becomes an implementation ticket. If not, this could be a candidate for a future project separate from the Gitpod application.

MrSimonEmms:
## Description

Under normal circumstances, a KOTS application is able to manage the [status of a deployment](https://docs.replicated.com/vendor/admin-console-display-app-status) - presumably through checking an annotation exists on a pod. As we are using the Installer rather than Kustomize/Helm and we cannot add the required annotations, the status will simply check that the resource exists which means that upgrades will find the _old_ resource rather than the new, falsely reporting that the application is deployed.

This uses the Helm status to find out the status of the Helm installation and crashes if it is not "deployed". This gives a KOTS user feedback as to the status of the Gitpod installation job. Unfortunately, it's a little bit brute-force so it show "unavailable" rather than "upgrading", but we're limited by what KOTS provides us.

image

When the Helm deployment has been successfully completed, it deletes the pod. This is important as the Kubernetes job works on an exponential back-off time - if the installation takes a while, this can result in many minutes until it's marked as "deployed". This makes the feedback as fast as possible.

This is the first of various tasks to make KOTS a more reliable and enjoyable experience, especially when not following the Happy Path.

## Related Issue(s)

Fixes #9195 

## How to test

Deploy via KOTS and check the status

## Release Notes

```release-note
[kots]: add an installation status pod
```

### DIFF
--- a/.werft/jobs/build/build-and-publish.ts
+++ b/.werft/jobs/build/build-and-publish.ts
@@ -111,6 +111,7 @@ function publishKots(werft: Werft, jobConfig: JobConfig) {
 
     // Set the tag to the current version
     exec(`yq w -i ${REPLICATED_YAML_DIR}/gitpod-installer-job.yaml ${INSTALLER_JOB_IMAGE} ${image}:${jobConfig.version}`, { slice: phases.PUBLISH_KOTS });
+    exec(`yq w -i ${REPLICATED_YAML_DIR}/gitpod-installation-status.yaml ${INSTALLER_JOB_IMAGE} ${image}:${jobConfig.version}`, { slice: phases.PUBLISH_KOTS });
 
     // Generate the logo
     exec(`make logo -C ${REPLICATED_DIR}`, { slice: phases.PUBLISH_KOTS });
--- a/install/kots/manifests/gitpod-installation-status.yaml
+++ b/install/kots/manifests/gitpod-installation-status.yaml
@@ -1,0 +1,57 @@
+# Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+# Licensed under the MIT License. See License-MIT.txt in the project root for license information.
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: installation-status
+  labels:
+    app: gitpod
+    component: gitpod-installer-status
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: gitpod
+      component: gitpod-installer-status
+  strategy:
+    rollingUpdate:
+      maxSurge: 1
+      maxUnavailable: 0
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: gitpod
+        component: gitpod-installer-status
+    spec:
+      restartPolicy: Always
+      serviceAccountName: installer
+      containers:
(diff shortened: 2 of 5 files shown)


install/kots/manifests/gitpod-installation-status.yaml
```diff
@@ -0,0 +1,57 @@
...
+              set -e
+
+              while true
+              do
+                echo "Checking installation status"
+
+                if [ "$(helm status -n {{repl Namespace }} gitpod -o json | jq '.info.status == "deployed"')" != "true" ];
+                then
+                  echo "Gitpod: Installation not complete"
+                  exit 1
```
nandajavarma, line 48: Question: (I haven't tested this yet) when an upgrade is happening wouldn't a new deployment be created of the `installation-status` deployment too? wouldn't KOTS UI show its status as `Ready` referring from the old deployment?

MrSimonEmms: Theoretically yes, but only for 10 seconds whilst the sleep waits for the next try. I think that's an acceptable compromise.

In practice, I've not found that to be the case though

.werft/jobs/build/build-and-publish.ts
```diff
@@ -111,6 +111,7 @@ function publishKots(werft: Werft, jobConfig: JobConfig) {
 
     // Set the tag to the current version
     exec(`yq w -i ${REPLICATED_YAML_DIR}/gitpod-installer-job.yaml ${INSTALLER_JOB_IMAGE} ${image}:${jobConfig.version}`, { slice: phases.PUBLISH_KOTS });
+    exec(`yq w -i ${REPLICATED_YAML_DIR}/gitpod-installation-status.yaml ${INSTALLER_JOB_IMAGE} ${image}:${jobConfig.version}`, { slice: phases.PUBLISH_KOTS });
```
MrSimonEmms, line 114: Will replace with a job in the KOTS `Makefile`

nandajavarma (approved):
LGTM! :shipit:

PR #9458, opened 2022-04-21
[kots]: move dropImageRepo config to when using local registry

mrsimonemms:
## Description

Move the `dropImageRepo` config to only when using local registry. This breaks non-airgapped installations

## Release Notes

```release-note
[kots]: move dropImageRepo config to when using local registry
```


MrSimonEmms:
/werft run no-preview publish-to-kots with-clean-slate-deployment

corneliusludmann (approved):

PR #9463, opened 2022-04-21
[server] fix "floating promises" error

svenefftinge:


roboquat:

<details>

</details>

components/server/src/analytics.ts
```diff
@@ -16,7 +16,7 @@ export async function trackLogin(
     subscriptionService: SubscriptionService,
 ) {
     // make new complete identify call for each login
-    fullIdentify(user, request, analytics, subscriptionService);
+    await fullIdentify(user, request, analytics, subscriptionService);
```
jankeromnes, line 19: I don't think we should make everyone wait for this -- i.e. this call is probably fine to leave async.

We could silence the warning though (don't remember if that's with `/** no await */` or something else)

jankeromnes (changes requested):
Many thanks for fixing these warnings! 🙏

components/server/src/analytics.ts
jankeromnes: Ah, indeed, thanks for pointing this out. 👍

jankeromnes (approved):
Adding `await` there is okay because the calling functions are themselves not being awaited.

Issue #7525, opened 2022-01-10
Create "Open In Gitpod (JetBrains)" button


loujaybee:
Create an "Open In JetBrains" button, which allows users to force open a project within JetBrains. This UX would be useful to support our work on the JetBrains integration. In terms of solution, I believe this can be done via URL parameters. 

Related to: 
- https://github.com/gitpod-io/gitpod/pull/9467

akosyakov:
I reopen it, since i thought to reuse new referrer prefix for that, but it is not really great, since it will disturb our data about local clients. I will add a dedicated prefix for that instead.

loujaybee:
Removing from scheduled, unlikely to be completed as part of the next release, but could be picked up in future.

mustard-mh:
close, will continue with this issue https://github.com/gitpod-io/gitpod/issues/9434

akosyakov:
@mustard-mh I don't think it is a duplicate of https://github.com/gitpod-io/gitpod/issues/9434.

svenefftinge:
not sure where that button should appear, but I think this has been somewhat fixed by https://github.com/gitpod-io/gitpod/pull/15389

PR #9467, opened 2022-04-21
[server] add editor prefix to choose custom IDE

### LINKED ISSUE #9434: Open workspace with custom IDE via URL
Open workspace with custom IDE via URL. related PR https://github.com/gitpod-io/browser-extension/pull/59#discussion_r853098135

mustard-mh:
## Description

Support choose custom IDE with prefix `/#editor:<ide>:<version?>/<repo>`

|Name|Desc|Required|values|Default|Example|
|:-:|:-:|:-:|:-:|:-:|:-:|
|ide| IDE id choose to open| true | see [here](https://github.com/gitpod-io/gitpod/blob/b2d07b8bf0d3364cd8b831e487a79fa7ab1fc638/chart/templates/server-ide-configmap.yaml#L45-L89) | code | `code` `goland` ... |
|version| IDE version | false| stable or latest | stable | latest |

## Related Issue(s)

Fixes #9434

https://github.com/gitpod-io/template-sveltejs
## How to test

Open with 

1. code stable => https://hw-9434-open-via-url.staging.gitpod-dev.com/#editor:code/https://github.com/gitpod-io/template-sveltejs
2. code desktop stable => https://hw-9434-open-via-url.staging.gitpod-dev.com/#editor:code-desktop/https://github.com/gitpod-io/template-sveltejs
3. goland stable => https://hw-9434-open-via-url.staging.gitpod-dev.com/#editor:goland/https://github.com/gitpod-io/template-sveltejs
4. goland latest => https://hw-9434-open-via-url.staging.gitpod-dev.com/#editor:goland:latest/https://github.com/gitpod-io/template-sveltejs
5. **unknown ide** will use default code => https://hw-9434-open-via-url.staging.gitpod-dev.com/#editor:not-exists:latest/https://github.com/gitpod-io/template-sveltejs

## Release Notes

```release-note
NONE
```

## Documentation

### DIFF
--- a/components/gitpod-protocol/src/protocol.ts
+++ b/components/gitpod-protocol/src/protocol.ts
@@ -863,6 +863,17 @@ export namespace PrebuiltWorkspaceContext {
     }
 }
 
+export interface WithEditorContext extends WorkspaceContext {
+    ide: string;
+    useLatest?: boolean;
+}
+
+export namespace WithEditorContext {
+    export function is(context: any): context is WithEditorContext {
+        return context && "ide" in context;
+    }
+}
+
 export interface WithReferrerContext extends WorkspaceContext {
     referrer: string;
     referrerIde?: string;
--- a/components/server/src/workspace/workspace-starter.ts
+++ b/components/server/src/workspace/workspace-starter.ts
@@ -156,7 +156,7 @@ export const chooseIDE = (
     const data: { desktopIdeImage?: string; ideImage: string } = {
         ideImage: defaultIdeImage,
     };
-    const chooseOption = ideOptions.options[ideChoice];
+    const chooseOption = ideOptions.options[ideChoice] ?? defaultIDEOption;
     const isDesktopIde = chooseOption.type === "desktop";
     if (isDesktopIde) {
         data.desktopIdeImage = useLatest ? chooseOption?.latestImage ?? chooseOption?.image : chooseOption?.image;
--- a/components/server/src/workspace/workspace-starter.ts
+++ b/components/server/src/workspace/workspace-starter.ts
@@ -103,7 +103,7 @@ import * as path from "path";
 import * as grpc from "@grpc/grpc-js";
 import { IDEConfig, IDEConfigService } from "../ide-config";
 import { EnvVarWithValue } from "@gitpod/gitpod-protocol/src/protocol";
-import { WithReferrerContext } from "@gitpod/gitpod-protocol/lib/protocol";
+import { WithEditorContext, WithReferrerContext } from "@gitpod/gitpod-protocol/lib/protocol";
 import { IDEOption, IDEOptions } from "@gitpod/gitpod-protocol/lib/ide-protocol";
 import { Deferred } from "@gitpod/gitpod-protocol/lib/util/deferred";
 import { ExtendedUser } from "@gitpod/ws-manager/lib/constraints";
@@ -607,8 +607,12 @@ export class WorkspaceStarter {
             user.additionalData.ideSettings = migratted;
         }
 
-        const ideChoice = user.additionalData?.ideSettings?.defaultIde;
-        const useLatest = !!user.additionalData?.ideSettings?.useLatestVersion;
+        const preferenceIDE = user.additionalData?.ideSettings?.defaultIde;
+        const perferenceUseLatest = !!user.additionalData?.ideSettings?.useLatestVersion;
+        const ideChoice = WithEditorContext.is(workspace.context) ? workspace.context.ide : preferenceIDE;
+        const useLatest = WithEditorContext.is(workspace.context)
+            ? workspace.context.useLatest ?? perferenceUseLatest
+            : perferenceUseLatest;
 
         // TODO(cw): once we allow changing the IDE in the workspace config (i.e. .gitpod.yml), we must
         //           give that value precedence over the default choice.
@@ -618,7 +622,7 @@ export class WorkspaceStarter {
             ideConfig: {
                 // We only check user setting because if code(insider) but desktopIde has no latestImage
                 // it still need to notice user that this workspace is using latest IDE
-                useLatest: user.additionalData?.ideSettings?.useLatestVersion,
+                useLatest,
             },
         };
 
@@ -635,9 +639,10 @@ export class WorkspaceStarter {
 
         const referrerIde = this.resolveReferrerIDE(workspace, user, ideConfig);
         if (referrerIde) {
-            configuration.desktopIdeImage = useLatest
+            configuration.desktopIdeImage = perferenceUseLatest
                 ? referrerIde.option.latestImage ?? referrerIde.option.image
                 : referrerIde.option.image;
+            configuration.ideConfig!.useLatest = perferenceUseLatest;
             if (!user.additionalData?.ideSettings) {
                 // A user does not have IDE settings configured yet configure it with a referrer ide as default.
                 const additionalData = user?.additionalData || {};
--- a/components/gitpod-protocol/src/context-url.ts
+++ b/components/gitpod-protocol/src/context-url.ts
@@ -20,6 +20,7 @@ export namespace ContextURL {
     export const IMAGEBUILD_PREFIX = "imagebuild";
     export const SNAPSHOT_PREFIX = "snapshot";
     export const REFERRER_PREFIX = "referrer:";
+    export const EDITOR_PREFIX = "editor:";
 
     /**
      * This function will (try to) return the HTTP(S) URL of the context the user originally created this workspace on.
@@ -94,7 +95,8 @@ export namespace ContextURL {
             firstSegment === INCREMENTAL_PREBUILD_PREFIX ||
             firstSegment === IMAGEBUILD_PREFIX ||
             firstSegment === SNAPSHOT_PREFIX ||
-            firstSegment.startsWith(REFERRER_PREFIX)
+            firstSegment.startsWith(REFERRER_PREFIX) ||
+            firstSegment.startsWith(EDITOR_PREFIX)
         ) {
             return segmentsToURL(1);
         }
--- a/components/server/src/container-module.ts
+++ b/components/server/src/container-module.ts
@@ -96,6 +96,7 @@ import { DebugApp } from "./debug-app";
 import { LocalMessageBroker, LocalRabbitMQBackedMessageBroker } from "./messaging/local-message-broker";
 import { contentServiceBinder } from "@gitpod/content-service/lib/sugar";
 import { ReferrerPrefixParser } from "./workspace/referrer-prefix-context-parser";
+import { EditorPrefixParser } from "./workspace/editor-prefix-context-parser";
 import { InstallationAdminTelemetryDataProvider } from "./installation-admin/telemetry-data-provider";
 
 export const productionContainerModule = new ContainerModule((bind, unbind, isBound, rebind) => {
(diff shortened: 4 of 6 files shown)


components/server/src/workspace/workspace-starter.ts
```diff
@@ -702,6 +719,18 @@ export class WorkspaceStarter {
                     referrerIde: workspace.context.referrerIde,
                 },
             });
+        } else if (WithEditorContext.is(workspace.context)) {
+            this.analytics.track({
+                userId: user.id,
+                event: "custom_editor_referrer",
```
akosyakov, line 725: I guess we can reuse `ide_referrer` here and extend code above to support `useLatest` as well?

components/server/src/workspace/workspace-starter.ts
```diff
@@ -633,6 +633,23 @@ export class WorkspaceStarter {
             configuration.desktopIdeImage = choose.desktopIdeImage;
         }
 
+        log.info(
```
akosyakov, line 636: Is it only for debugging? I think it should have `debug` log. There is a tool to enable debug log in prev envs.

I usually use `yarn telepresence` + `yarn watch` on server, and then attach with F5 and debug it.

components/server/src/workspace/workspace-starter.ts
```diff
@@ -633,6 +633,23 @@ export class WorkspaceStarter {
...
+            { context: workspace.context, isEditorContext: WithEditorContext.is(workspace.context) },
+            "check context",
+        );
+        if (WithEditorContext.is(workspace.context)) {
```
akosyakov, line 640: Should not we rather merge it with if above and only compute `ideChoice` and `useLatest` based on it?

components/gitpod-protocol/src/protocol.ts
```diff
@@ -863,14 +863,26 @@ export namespace PrebuiltWorkspaceContext {
...
+
 export interface WithReferrerContext extends WorkspaceContext {
     referrer: string;
     referrerIde?: string;
 }
 
 export namespace WithReferrerContext {
     export function is(context: any): context is WithReferrerContext {
-        return context && "referrer" in context;
+        return context && "referrer" in context && "referrerIde" in context;
```
akosyakov, line 885: Won't it break?

components/server/src/workspace/workspace-starter.ts
mustard-mh: will remove these, they are debug log here https://github.com/gitpod-io/gitpod/pull/9467/commits/23308de53dc84a6ddca92dde75d58a1e3e34d9c5

components/gitpod-protocol/src/protocol.ts
```diff
@@ -863,14 +863,26 @@ export namespace PrebuiltWorkspaceContext {
...
+
 export interface WithReferrerContext extends WorkspaceContext {
     referrer: string;
     referrerIde?: string;
 }
 
 export namespace WithReferrerContext {
     export function is(context: any): context is WithReferrerContext {
-        return context && "referrer" in context;
+        return context && "referrer" in context && !("ide" in context);
```
akosyakov, line 885: `&& !WithEditorContext.is(context)`?

components/server/src/workspace/workspace-starter.ts
```diff
@@ -607,8 +607,12 @@ export class WorkspaceStarter {
             user.additionalData.ideSettings = migratted;
         }
 
-        const ideChoice = user.additionalData?.ideSettings?.defaultIde;
-        const useLatest = !!user.additionalData?.ideSettings?.useLatestVersion;
+        const preferenceIDE = user.additionalData?.ideSettings?.defaultIde;
+        const perferenceUseLatest = !!user.additionalData?.ideSettings?.useLatestVersion;
+        const ideChoice = WithEditorContext.is(workspace.context) ? workspace.context.ide : preferenceIDE;
```
akosyakov, line 612: what happens if `workspace.context.ide` is invalid? Should we fall back to `preferenceIDE`?

components/gitpod-protocol/src/protocol.ts
mustard-mh: How about using `#referrer` and remove checkes above?

components/server/src/workspace/workspace-starter.ts
```diff
@@ -691,17 +696,24 @@ export class WorkspaceStarter {
             },
             configuration,
         };
-        if (WithReferrerContext.is(workspace.context)) {
+
+        const trackIDEPreferrer = (referrer?: string, referrerIde?: string, useLatest?: boolean) => {
```
akosyakov, line 700: IDE**P** -> IDE ?

components/server/src/workspace/workspace-starter.ts
mustard-mh: chooseIDE will do this step, see `workspace-starter.spec.ts`

components/server/src/workspace/workspace-starter.ts
```diff
@@ -607,8 +607,12 @@ export class WorkspaceStarter {
             user.additionalData.ideSettings = migratted;
         }
 
-        const ideChoice = user.additionalData?.ideSettings?.defaultIde;
-        const useLatest = !!user.additionalData?.ideSettings?.useLatestVersion;
+        const preferenceIDE = user.additionalData?.ideSettings?.defaultIde;
+        const perferenceUseLatest = !!user.additionalData?.ideSettings?.useLatestVersion;
+        const ideChoice = WithEditorContext.is(workspace.context) ? workspace.context.ide : preferenceIDE;
+        const useLatest = WithEditorContext.is(workspace.context)
+            ? workspace.context.useLatest ?? false
```
akosyakov, line 614: actually maybe it makes sense that version falls back to user preferences if not provided?

components/gitpod-protocol/src/protocol.ts
akosyakov: Could you elaborate please? I did not grasp an idea, sorry.

mustard-mh: `#editor` works like `#referrer` we can extend `#referrer` and remove `#editor`.

Now `#referrer` only works for client (referrer) `vscode` `vscode-insiders` and `jetbrains-gateway`, and IDE option can only be one of these below 👇 . 
https://github.com/gitpod-io/gitpod/blob/330ddb2fd2d27e83e9eae400bc4ec96af91b98d2/chart/templates/server-ide-configmap.yaml#L109
We can remove this check and make referrer to be anything like `bowser-extension` `dashboard-project` `readme` etc and IDEs with any IDE. They are worth tracking too

Or extend `ide-configmap` to support `browser-extension`

mustard-mh: Or with `#editor` if referrer is not define, we do not send track here, and consider `#eidor` as make to choose IDE

akosyakov: I think we conflate 2 issue here:
- https://github.com/gitpod-io/gitpod/issues/7525 is about allowing a user to open a workspace with given editor, we don't care about referrer
- https://github.com/gitpod-io/gitpod/issues/9434 seems to be just to enable `browser-extension`

For first issue we don't care about referrer and should not reuse it, it should be very simple `#editor:code` user is done.

For second indeed maybe we can consider to lift off some constraints. Initially we introduced to ensure that JB users via Gateway never can start a workspace with no JB product. It still should hold even we remove a constraint. You will need to check GW plugin to ensure it.

Maybe let's reopen https://github.com/gitpod-io/gitpod/issues/7525, drop `referrer` from this PR for the editor context and fix it? And then open another PR which lifts off constraints of  `referrer` prefix in a way that it does not break JB users and enables `browser-extension`?

mustard-mh: This PR introduces the `editor` parameter, which have three parameters, `ide`, `version`, and `referrer`, but except `ide` all of them are **optional**

So it can be used to solve #7525, just use `#ediror:code/http://xxxx`
It can also be used to solve #9434 with just `#editor:code:stable:browser-extension/http://xxxx`
It can also be used to solve any need to temporarily switch IDEs, for example, we may introduce another feature later, where we can hold down the option key in the Open workspace (select url) to temporarily select the current open IDE

akosyakov: Let's solve one issue at the time instead one technical solution for everything and don't be to worry about all possible which we will need somewhen 🙏 

> It can also be used to solve https://github.com/gitpod-io/gitpod/issues/9434 with just #editor:code:stable:browser-extension/http://xxxx

Let's use `referrer` still there, but introduce a new kind `browser-extension` which allows any IDE and use `code` as default. We don't want to allow users to specify random referrers. Please also prioritise browser-extension support as minor. It is not a primary onboarding experience.

> So it can be used to solve https://github.com/gitpod-io/gitpod/issues/7525, just use #ediror:code/http://xxxx

We don't need uncontrolled referrer here at all. It also not nice syntax I would go with something like `#editor:code&referrer=.../https://`, but then it is better to use `referrer` prefix with new kinds. I suggest we go with an initial proposal here `editor:<ide>(:<version>)?`. Adding analytics is not bad then to see whether it is used, but it should not be mixed with `ide_referrer` event anymore then, something like `ide_editor_prefix`.

> It can also be used to solve any need to temporarily switch IDEs, for example, we may introduce another feature later, where we can hold down the option key in the Open workspace (select url) to temporarily select the current open IDE https://github.com/gitpod-io/gitpod/issues/9433#issuecomment-1105694971

From the dashboard, we should not rely on prefix parsers, but improve APIs to allow passing concrete IDE to create or start workspace methods. This one actually sounds like most important issue of all mentioned above. But should rather align with @loujaybee and @andreafalzetti what is goal for onboarding right now.

mustard-mh: > I suggest we go with an initial proposal here editor:<ide>(:<version>)?

Go with this. will create a PR to open workspace with custom IDE via API later. 

If PR above is created, and PR to add referrer `browser-extension` is created, why we need this PR? No one will use it.🙈

Feel free to close this PR @akosyakov

akosyakov: > If PR above is created, and PR to add referrer browser-extension is created, why we need this PR? No one will use it.🙈

- People will add to README if a project is configured only to work nicely with VS Code for instance.
- I already would like to use it on gitpod-io/gitpod, since I usually have IntelliJ one, but would like to have a workspace only with VS Code 🙏 

I think we will need to add some docs about this prefix somewhere. @loujaybee

akosyakov: Ah, sorry I misunderstand. You say you will do 2 PRs instead? Then yes just close this PR and open 2 others.

jeanp413:
I have a question:
Why are we using this format `/#editor:<ide>:<version?>/<repo>` instead of just using the query parameters in the url?
I get the idea of using the fragment part to pass the repo url as it's really easy to just prepend `gitpod.io#` to some repo url, but for more complex stuff like this a don't see a point, it just makes it more complex and cumbersome to use, for both end users and us :thinking:

geropl (changes requested):
Apart from the general question "what is meant to be part of the start workpace URL": We should really have a basic test converage before merging a `ContextParser`.

Two things I would be interested in:
 - the "internal" stuff works, and does not throw exceptions (extensive)
 - the "external" facet: it does play nicely with other context parsers, e.g. EnvVars (some examples)

mustard-mh:
> I have a question:
> Why are we using this format /#editor:<ide>:<version?>/<repo> instead of just using the query parameters in the url?
I get the idea of using the fragment part to pass the repo url as it's really easy to just prepend gitpod.io# to some repo url, but for more complex stuff like this I don't see a point, it just makes it more complex and cumbersome to use, for both end users and us 🤔

@jeanp413  I think this question should back to why we use fragment for`/#prebuild` and env var😂

In my opinion, we may just want to split input into `Repo from user` and `Gitpod params`

mustard-mh:
Close, will create a new one for it in the future

PR #9471, opened 2022-04-21
[kots]: allow deployment whilst installer job running

MrSimonEmms:
## Description

> In this PR, install means "install or upgrade" unless otherwise stated

Allow deployment whilst an existing Installer job running. This should allow for recovery in the following scenarios:

1. Deployment failed/stopped mid-install and user wants to redeploy same version
2. Deployment failed/stopped mid-install and user wants to deploy a different version
3. User decides to trigger a new deployment whilst an existing deployment is in progress

This appends the [cursor](https://docs.replicated.com/reference/template-functions-license-context#cursor) to the job name. This is necessary to allow us to change the `image` on the container, effectively replacing the existing job. The first part of this script then removes any already running jobs/pods, thus clearing up the state.

It then rolls back (or uninstalls if it's the first deployment) before starting the next `helm upgrade`. For a user, this should then allow them to unstick their issues.

Setting of the `ttlSecondsAfterFinished` to `0` means that the "redeploy" button works as soon as the job as finished

## How to test

Get your Helm deployment into either a `pending-install` or `pending-upgrade` state. The script does this anyway for an upgrade, but to do this manually, suggest starting an install job then running:

```shell
kubectl delete jobs.batch -n gitpod -l component=gitpod-installer --force --grace-period 0
kubectl delete pod -n gitpod -l component=gitpod-installer --force --grace-period 0
```

> The reason it doesn't work for a redeployment is because Kubernetes sees the request to deploy a job called `installer-1234` and, as that's already deployed, it thinks "oooh, nothing to do". There seems to be no way of generating a random value at deploy-time with KOTS - have asked as a [question in our channel](https://gitpod.slack.com/archives/C02SWJ49JQ2/p1650571390051829)

Ensure by running `helm ls -aA`.

Now, run the following - all of these should result in a successful deployment, and will need you to get it to enter the same "pending" state as above:
1. Redeploy the existing deployment.
4. Publish a new version to KOTS and deploy that

For regression, we should also test:
1. A fresh installation (run `helm un -n gitpod gitpod`)
2. An upgrade that's not in a "pending" state
5. A redeployment that's not in a "pending" state

## Release Notes

```release-note
[kots]: allow deployment whilst installer job running
```

### DIFF
--- a/install/kots/manifests/gitpod-installer-job.yaml
+++ b/install/kots/manifests/gitpod-installer-job.yaml
@@ -7,24 +7,27 @@
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: installer
+  # Appending cursor allows us to replace with new versions
+  name: installer-{{repl Cursor }}
   labels:
     app: gitpod
     component: gitpod-installer
+    cursor: "{{repl Cursor }}"
 spec:
-  ttlSecondsAfterFinished: 60
+  ttlSecondsAfterFinished: 0
   template:
     metadata:
       labels:
         app: gitpod
         component: gitpod-installer
+        cursor: "{{repl Cursor }}"
     spec:
       serviceAccountName: installer
       restartPolicy: OnFailure
       containers:
         - name: installer
           # This will normally be the release tag
-          image: 'eu.gcr.io/gitpod-core-dev/build/installer:main.2968'
+          image: "eu.gcr.io/gitpod-core-dev/build/installer:main.2968"
           volumeMounts:
             - mountPath: /config-patch
               name: config-patch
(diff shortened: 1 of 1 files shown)


roboquat:

<details>

</details>

install/kots/manifests/gitpod-installer-job.yaml
```diff
@@ -50,6 +53,26 @@ spec:
             - |
               set -e
 
+              echo "Gitpod: Killing any in-progress installations"
+
+              kubectl delete jobs.batch -n {{repl Namespace }} -l component=gitpod-installer,cursor!={{repl Cursor }} --force || true
+              kubectl delete pod -n {{repl Namespace }} -l component=gitpod-installer,cursor!={{repl Cursor }} --force || true
```
nandajavarma, line 59: Maybe a ` --grace-period=0` could be added too, just to be sure?

install/kots/manifests/gitpod-installer-job.yaml
```diff
@@ -50,6 +53,26 @@ spec:
...
+
+              if [ "$(helm status -n {{repl Namespace }} gitpod -o json | jq '.info.status == "deployed"')" = "false" ];
+              then
+                echo "Gitpod: Deployment in-progress - clearing"
+
+                VERSION="$(helm status -n {{repl Namespace }} gitpod -o json | jq '.version')"
+                if [ "${VERSION}" -le 1 ];
+                then
+                  echo "Gitpod: Uninstall application"
+                  helm uninstall -n {{repl Namespace }} gitpod --wait || true
```
nandajavarma, line 69: Should we be worried about deleting the PVCs?

nandajavarma:
Tested and works like a charm! LGTM! Merging on Friday early evening is bad luck. Will approve on Monday 🤓

MrSimonEmms:
>  Merging on Friday early evening is bad luck

You are very wise

install/kots/manifests/gitpod-installer-job.yaml
MrSimonEmms: Good call

corneliusludmann (approved):
Since @nandajavarma is OOO today, I'm happy to add my approval. I am very curious to see how well this works in practice.

Issue #9407, opened 2022-04-19
Why did fixing CPU limit annotation break CPU limiting with cgroup v2?


kylos101:
### Bug description

We broke the gitpod.io/cpuLimit annotation when we introduced cpu limiting. We then broke cpu limiting [in this PR](https://github.com/gitpod-io/gitpod/pull/8459). As a result, many workspaces were trying to use 6 cores at the same time, and this resulted in a poor experience for users.

The intent of this issue is to:
1. Examine what [in this PR](https://github.com/gitpod-io/gitpod/pull/8459) broke cpu limiting in gen39 and gen40, and why [reverting it](https://github.com/gitpod-io/gitpod/pull/9382) worked for gen34 (which used cgroup v2)
2. See if we can truly fix the CPU limit annotation, without breaking CPU limiting for cgroup v2

### Steps to reproduce

Revert [this revert](https://github.com/gitpod-io/gitpod/pull/9382). You should be able to have many workspaces in the cluster that use up to 6 cores of CPU.[](url)

### Workspace affected

n/a

### Expected behavior

1. CPU limiting should rotate CPU fairly, so it is not possible for every workspace pod to try and use 6 CPUs. For example, the 6 cores should 'rotate' from workspace to workspace - not all workspaces should be able to use 6 cores at the same time.
2. When the `gitpod.io/cpuLimit` annotation is applied, `ws-daemon` should enforce it. For example, if it is set to 1, you get 1.

### Example repository

n/a

utam0k:
> Revert this revert. You should be able to have many workspaces in the cluster that use up to 6 cores of CPU.

@kylos101 Maybe #8459 is wrong? Does `this revert` point https://github.com/gitpod-io/gitpod/pull/9382?

utam0k:
Hi, @csweichel. I'd like to confirm for you that `fixedLimiter` of CPU limit isn't implemented, right? Do you remember them?
This feels unimplemented to me, is that correct?
https://github.com/gitpod-io/gitpod/pull/8036/files#diff-49497202e6fb01e6a2d9fe29e32980d3c4f25dcdacfffd9b41e6447895ed27a5R213

csweichel:
@utam0k I'm not sure I follow. `fixedLimiter` is implemented [here](https://github.com/gitpod-io/gitpod/pull/8036/files#diff-49497202e6fb01e6a2d9fe29e32980d3c4f25dcdacfffd9b41e6447895ed27a5R221-R223).

[...]

PR #8459, opened 2022-02-25
[ws-daemon] Fix CPU limit annotation

csweichel:
## Description
When we updated the CPU limiting mechanism, we broke the `gitpod.io/cpuLimit` annotation. This PR makes that annotation work again. It effectively controls the base and burst limits of a workspace.

## How to test
1. Start a workspace
2. Set the `gitpod.io/cpuLimit` annotation on that workspace
3. Observe how the workspace's cfs quota gets set to that value

## Release Notes

```release-note
Made the `gitpod.io/cpuLimit` annotation work again
```

PR #9479, opened 2022-04-22
[ws-daemon] Fix CPU limit annotation

utam0k:
## Description

Fix CPU limit annotation.
I actually annotated it manually and it worked.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/9407

## How to test

Annotate as follows and confirm that `cpu.max` of cgroup is set. 
Note: you have to enable CPU limit configuration in ws-daemon.

```console
$ kubectl annotate --overwrite pods $ws-pod gitpod.io/cpuLimit=50000m
```

## Release Notes

```release-note
[ws-daemon] Fix CPU limit annotation
```

### DIFF
--- a/components/ws-daemon/pkg/cpulimit/cpulimit.go
+++ b/components/ws-daemon/pkg/cpulimit/cpulimit.go
@@ -18,6 +18,9 @@ type Workspace struct {
 	NrThrottled uint64
 	Usage       CPUTime
 	QoS         int
+
+	BaseLimit  Bandwidth
+	BurstLimit Bandwidth
 }
 
 type WorkspaceHistory struct {
@@ -113,7 +116,7 @@ func (d *Distributor) Tick(dt time.Duration) (DistributorDebug, error) {
 		return DistributorDebug{}, err
 	}
 
-	f := make(map[string]struct{}, len(ws))
+	wsidx := make(map[string]Workspace, len(ws))
 	for _, w := range ws {
 		h, ok := d.History[w.ID]
 		if !ok {
@@ -123,10 +126,10 @@ func (d *Distributor) Tick(dt time.Duration) (DistributorDebug, error) {
 			d.History[w.ID] = h
 		}
 		h.Update(w)
-		f[w.ID] = struct{}{}
+		wsidx[w.ID] = w
 	}
 	for oldWS := range d.History {
-		if _, found := f[oldWS]; !found {
+		if _, found := wsidx[oldWS]; !found {
 			delete(d.History, oldWS)
 		}
 	}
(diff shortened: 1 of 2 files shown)


utam0k:
@aledbf @sagor999 Do you remember why you reverted last time? I felt like this was working fine.
https://github.com/gitpod-io/gitpod/pull/9382

components/ws-daemon/pkg/cpulimit/cpulimit.go
```diff
@@ -172,15 +175,23 @@ func (d *Distributor) Tick(dt time.Duration) (DistributorDebug, error) {
...
 
 		// if we didn't get the max bandwidth, but were throttled last time
 		// and there's still some bandwidth left to give, let's act as if had
 		// never spent any CPU time and assume the workspace will spend their
 		// entire bandwidth at once.
 		var burst bool
 		if totalBandwidth < d.TotalBandwidth && ws.Throttled() {
-			limit = d.BurstLimiter.Limit(ws.Usage())
+			limiter := d.BurstLimiter
+			if w := wsidx[id]; w.BaseLimit > 0 {
```
csweichel, line 191: shouldn't this be `w.BurstLimiter > 0`?

sagor999:
@utam0k it was reverted because it was breaking CPU limiting for all workspaces. CPU limiting was not working at all with this change, not clear why.

kylos101:
@utam0k my recollection is that it did not work with cgroup v2 and gen34. 
We are currently using cgroup v1 in production with gen41.
How did you do your test (above)? Did you test with cgroup v2 in workspace-preview?

utam0k:
> @utam0k my recollection is that it did not work with cgroup v2 and gen34. We are currently using cgroup v1 in production with gen41. How did you do your test (above)? Did you test with cgroup v2 in workspace-preview?

@kylos101 
That's right. I used the workspace-preview with cgroup v2 for the tests.

utam0k:
@sagor999 @kylos101 
Sorry, I haven't figured out how to break it yet, so let me confirm a few things. 
1. What are the situations of the breakage? When annotated?
2. How was it broken? CPU limits not working, cgroup not reflecting values as expected?
3. Has the situation recovered with the revert?

kylos101:
@utam0k I just updated https://github.com/gitpod-io/gitpod/issues/9407 slightly, my apologies for being unclear.

> @sagor999 @kylos101 Sorry, I haven't figured out how to break it yet, so let me confirm a few things.
> 
> 1. What are the situations of the breakage? When annotated?
Ignore annotations for a moment. The problem was that cpu limiting was not working effectively after [this PR](https://github.com/gitpod-io/gitpod/pull/8459) was deployed to production. While we observed that CPU limiting was not working, I do not recall if this annotation was set on workspaces. @Furisto do you recall? 

> 2. How was it broken? CPU limits not working, cgroup not reflecting values as expected?
More than one workspace was getting 6 cores. cgroup was reflecting this, confirming that we were allocating too much.

> 3. Has the situation recovered with the revert?
Yes

utam0k:
I was able to experience this once during my trial and error. However, I felt the dispatcher did not seem to be fully operational. I will investigate a little more, but the recovery of the state may not be due to the revert the PR but to restarting ws-daemon at the time. I will investigate a little more overhead. However, there are many riddles...

kylos101:
Flipping this to draft, @utam0k , as you are still investigating.

utam0k:
Please close this PR once because I investigate this branch.

PR #9488, opened 2022-04-22
[dashboard] Make sure we have the auth cookie set before redirecting to the workspace URL

### LINKED ISSUE #9145: [StartWorkspace] Make sure ensureWorkspaceAuth is called
Today I reopened  two workspaces from yesterday which did not make into the IDE, but the workspace pod actually were there and reported readiness. After investigating with @geropl and @akosyakov, we came to the conclusion that the owner token might not have been set during the workspace startup flow, because all of the supervisor request were returned with code 403. 

Having a look at the StartWorkspace.tsx reveals that we might need to reiterate on the process `ensureWorkspaceAuth` is actually called. Cases to consider:
* workspace will be re-opened in a tab in background and browser window get blurred
* a timed-out workspace tab remained in background

geropl:
## Related Issue(s)

Fixes #9145 

## How to test
 - [starting a workspace](https://gpl-9145-ensureauth.preview.gitpod-dev.com/#https://github.com/gitpod-io/template-rust-cli) should work as usual :green_circle: 

## Release Notes

```release-note
fix a dashboard bug where you might end up on a workspace without having access to it
```

### DIFF
--- a/components/dashboard/src/start/StartWorkspace.tsx
+++ b/components/dashboard/src/start/StartWorkspace.tsx
@@ -199,8 +199,11 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
                 throw new Error("No result!");
             }
             console.log("/start: started workspace instance: " + result.instanceID);
+
             // redirect to workspaceURL if we are not yet running in an iframe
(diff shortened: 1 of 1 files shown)


easyCZ (approved):
works

Issue #9059, opened 2022-04-01
.werft/monitoring-satellite: Clean up logic


ArthurSens:
Following up from https://github.com/gitpod-io/gitpod/pull/9016#discussion_r838424340

The logic that deploys monitoring-satellite is getting pretty hard to read, and it's using `classes`, `types` and `interfaces` in unconventional ways.

We should clean that up 🙂

mads-hartmann:
Moving to the "Reduce Technical Debt" epic and adding it to the next iteration.

PR #9490, opened 2022-04-22
Convert to class to reduce argument passing

mads-hartmann:
## Description

This promotes `installMonitoringSatellite` to be a class to reduce the need for passing the same arguments around to the functions.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/9059

## How to test

- [x] in core-dev with observability enabled `werft run github -f -a with-helm=true -a with-observability=true` - [job](https://werft.gitpod-dev.com/job/gitpod-build-mads-classify-monitoring-satellite.3) - was able to port-forward to Prometheus and Grafana.
- [x] with-vm ` werft run github -f -a with-vm=true` - [job](https://werft.gitpod-dev.com/job/gitpod-build-mads-classify-monitoring-satellite.4) - was able to port-forward to Prometheus and Grafana but had to use the following commands instead of our own script. I suspect there's a bug in our proxy. Will open a follow up issue.
  ```
  kubectl --context=k3s-preview-environment port-forward svc/prometheus-k8s "9090:9090"
  kubectl --context=k3s-preview-environment port-forward svc/grafana "3000:3000"
  ```

## Release Notes

```release-note
NONE
```

## Documentation

N/A

### DIFF
--- a/.werft/jobs/build/deploy-to-preview-environment.ts
+++ b/.werft/jobs/build/deploy-to-preview-environment.ts
@@ -462,7 +462,7 @@ async function deployToDevWithHelm(werft: Werft, jobConfig: JobConfig, deploymen
     werft.log(`observability`, "Installing monitoring-satellite...")
     if (deploymentConfig.withObservability) {
         try {
-            await installMonitoring(werft, CORE_DEV_KUBECONFIG_PATH, namespace, nodeExporterPort, monitoringDomain, STACKDRIVER_SERVICEACCOUNT, false, jobConfig.observability.branch);
+            installMonitoring(werft, CORE_DEV_KUBECONFIG_PATH, namespace, nodeExporterPort, monitoringDomain, STACKDRIVER_SERVICEACCOUNT, false, jobConfig.observability.branch);
         } catch (err) {
             if (!jobConfig.mainBuild) {
                 werft.fail('observability', err);
@@ -771,7 +771,7 @@ async function installMetaCertificates(werft: Werft, branch: string, withVM: boo
     await installCertificate(werft, metaInstallCertParams, { ...metaEnv(), slice: slice });
 }
 
-async function installMonitoring(werft: Werft, kubeconfig: string, namespace: string, nodeExporterPort: number, domain: string, stackdriverServiceAccount: any, withVM: boolean, observabilityBranch: string) {
+function installMonitoring(werft: Werft, kubeconfig: string, namespace: string, nodeExporterPort: number, domain: string, stackdriverServiceAccount: any, withVM: boolean, observabilityBranch: string) {
     const installMonitoringSatellite = new MonitoringSatelliteInstaller({
         kubeconfigPath: kubeconfig,
         branch: observabilityBranch,
--- a/.werft/jobs/build/deploy-to-preview-environment.ts
+++ b/.werft/jobs/build/deploy-to-preview-environment.ts
@@ -161,8 +161,19 @@ export async function deployToPreviewEnvironment(werft: Werft, jobConfig: JobCon
         // success rate or installing monitoring satellite, but we can at least count and debug errors.
         // In the future we can consider not closing spans when closing phases, or restructuring our phases
         // based on parallelism boundaries
+        const monitoringSatelliteInstaller = new MonitoringSatelliteInstaller({
+            kubeconfigPath: PREVIEW_K3S_KUBECONFIG_PATH,
+            branch: jobConfig.observability.branch,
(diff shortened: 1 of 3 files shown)


.werft/observability/monitoring-satellite.ts
```diff
@@ -1,125 +1,161 @@
...
+                { slice: sliceName, async: true },
+            );
+        }
     }
-}
 
-async function deployGitpodServiceMonitors(kubeconfig: string) {
-    const werft = getGlobalWerftInstance()
+    private async deployGitpodServiceMonitors() {
+        const { werft, kubeconfigPath } = this.options;
```
ArthurSens, line 139: back when I implemented this I had no idea what was this `async` and was just adding because it was being used everywhere.

Does it make sense in this class? Almost function is using async, although we're not using Promises anywhere. Would it make sense to just remove it?

.werft/jobs/build/deploy-to-preview-environment.ts
```diff
@@ -771,17 +771,19 @@ async function installMetaCertificates(werft: Werft, branch: string, withVM: boo
...
-    installMonitoringSatelliteParams.kubeconfigPath = kubeconfig
-    installMonitoringSatelliteParams.branch = observabilityBranch;
-    installMonitoringSatelliteParams.satelliteNamespace = namespace
-    installMonitoringSatelliteParams.clusterName = namespace
-    installMonitoringSatelliteParams.nodeExporterPort = nodeExporterPort
-    installMonitoringSatelliteParams.previewDomain = domain
-    installMonitoringSatelliteParams.stackdriverServiceAccount = stackdriverServiceAccount
-    installMonitoringSatelliteParams.withVM = withVM
-    installMonitoringSatellite(installMonitoringSatelliteParams);
+async function installMonitoring(werft: Werft, kubeconfig: string, namespace: string, nodeExporterPort: number, domain: string, stackdriverServiceAccount: any, withVM: boolean, observabilityBranch: string) {
```
ArthurSens, line 774: same thing here, we declare an `async` function only to use `await` when calling it. Would it make sense to just remove it?

ArthurSens (approved):
Noice! I'm really liking this pattern for using those "installer" classes. 

I have a few questions but I don't know if they make sense 😅, so I'm approving with a hold. Please decide if we still need to make changes here, leave them another PR or not do them at all

/hold

mads-hartmann:
@ArthurSens Thanks for the pointer! I removed the async keyword from the methods and also the async option for the exec invocations - in effect making observability installation blocking. Previously we were blocking using `await` in core-dev but not in Harvester. As we generally want observability ready before we install Gitpod to avoid errors about Gitpod not being able to send telemetry I think it makes sense to always have the function be blocking. But I will test the code tomorrow and do another pass to see if it makes sense ☺️

ArthurSens:
I believe this is good to go? Can I just remove the hold? 🙂

mads-hartmann:
@ArthurSens When testing earlier it looked like it had broken tracing but I didn't have time to debug further. I will move this back to draft ☺️

.werft/jobs/build/deploy-to-preview-environment.ts
mads-hartmann: I'm keeping this async as it's my understand that we want to block in core-dev (when observability is enabled) but not block when installing it in VM-based preview environments ☺️

mads-hartmann:
Okay @ArthurSens this is finally ready for another review. In the future I think we want to separate the installation of monitoring satellite from the waiting for it to be ready. That way we'd avoid the problem of having a slice potentially spanning multiple phases, but I didn't want to make that change now as I've already spent too long on this 😅

This should PR should also fix the problem Chris has been experiencing with failed monitoring satellite deployments breaking the build as explained [here](https://gitpod.slack.com/archives/C01KGM9EBD4/p1651144271318639?thread_ts=1651061591.320309&cid=C01KGM9EBD4).

The last job has some missing spans, but I think that's more likely to to be due to burst protection as you also called out [here](https://gitpod.slack.com/archives/C01KGM9EBD4/p1651053173542839) - that also explains why I saw issue sometime yesterday but not always.

mads-hartmann:
@ArthurSens lol sorry, I have to get used to commits not dismissing previous reviews - I did want you to have a look at this before merging 😅  I don't want to revert it as I tested it pretty well, but if you have time please have a look and I can create a follow up PR with any changes you'd like to see

ArthurSens (commented):
No worries about the merge, nothing really important to change :)

PR #9492, opened 2022-04-22
Remove stargz from builder bob

Furisto:
## Description
Remove stargz from builder bob

## Related Issue(s)
n.a.

## How to test
n.a

## Release Notes

```release-note
Remove stargz snapshotter from image build
```

### DIFF
--- a/components/image-builder-bob/pkg/builder/builder.go
+++ b/components/image-builder-bob/pkg/builder/builder.go
@@ -138,7 +138,7 @@ func buildImage(ctx context.Context, contextDir, dockerfile, authLayer, target s
 		// "--debug",
 		"build",
 		"--progress=plain",
-		"--output=type=image,name=" + target + ",push=true,oci-mediatypes=true,compression=estargz,force-compression=true",
+		"--output=type=image,name=" + target + ",push=true,oci-mediatypes=true",
 		//"--export-cache=type=inline",
 		"--local=context=" + contextdir,
 		//"--export-cache=type=registry,ref=" + target + "-cache",
@@ -209,7 +209,7 @@ func StartBuildkit(socketPath string) (cl *client.Client, teardown func() error,
 	cmd := exec.Command("buildkitd",
 		"--debug",
 		"--addr="+socketPath,
-		"--oci-worker-net=host", "--oci-worker-snapshotter=stargz",
+		"--oci-worker-net=host",
 		"--root=/workspace/buildkit",
 	)
 	cmd.SysProcAttr = &syscall.SysProcAttr{Credential: &syscall.Credential{Uid: 0, Gid: 0}}


csweichel:
Can we move this along? We're still seeing stuck image builds in prod.

cc @Furisto @aledbf

aledbf:
@csweichel I am not sure this will help. I prefer to finish the image builder transition to workspace clusters

csweichel:
> I prefer to finish the image builder transition to workspace clusters

Same here - that's still a bit out though.
@Furisto if I understood you correctly, a possible cause of this issue is the use of FUSE when building the image, no? If so, remove this and reverting later seems like a low-hanging way to sort this out.

aledbf:
Can we reproduce the issue in a workspace cluster?

Furisto:
> Can we reproduce the issue in a workspace cluster?

Checking now...

aledbf:
@csweichel can we update the meta clusters without merging this change?

csweichel:
> @csweichel can we update the meta clusters without merging this change?

Good call.
@geropl wdyt?

csweichel:
it would be a change to the registry-facade configmap which automatically reloads. No service restart is required.

csweichel:
> I looked at registry-facade, and could not find a reference to image-builder-bob. Do you mean the image-builder-mk3 configmap, and its builderImage property?

🤦 
My bad - it is in fact the [image-builder config](https://github.com/gitpod-io/gitpod/blob/main/components/image-builder-api/go/config/config.go#L84) where we choose this value, not registry-facade.

geropl:
>>  @csweichel can we update the meta clusters without merging this change?
>
>Good call.
>@geropl wdyt?

@csweichel Sure, we can update right away.

It feels I'm missing some context, though: Since when do we see stuck image builds? Yesterdays deployment? :thinking:

geropl:
> Not entirely sure - but since Friday last week for sure

:+1: 

@csweichel @kylos101 Going to bump bob to: `eu.gcr.io/gitpod-core-dev/build/image-builder-mk3/bob:9c0c6ae5a34dab7507fa94ab357a004868111fdc` from the [last green build on this branch](https://werft.gitpod-dev.com/job/gitpod-build-fo-remove-stargz.2)

aledbf:
This change is not enough. The daemon still uses `--oci-worker-snapshotter=stargz` (the source of the error)

AlexTugarev:
@Furisto, what's the timeline on this fix? I'm asking, because unintentionally we did an undo of the workaround with the recent deployment of WebApp.

Furisto:
@aledbf I have removed the stargz snapshotter from the daemon args as well.

geropl:
@aledbf @Furisto Thank you both! So we're going to apply the bob image as last time to both prod app clusters, right?

aledbf:
@geropl yes please :pray:

csweichel (approved):

PR #9495, opened 2022-04-22
[installer]: Allow replica counts to be specified for all components

andrew-farries:
## Description

This PR adds the ability to configure the number of replicas for each component, rather than using to a hard-coded value of `1`.

## Related Issue(s)

Part of #9097 

## How to test

```yaml
experimental:
  common:
    podConfig:
      server:
        replicas: 123
      content-service:
        replicas: 456
```

```
```

Then invoke the installer as:

```
installer render --debug-version-file versions.yaml --config /path/to/config --use-experimental-config
```

The rendered output will have the values from the config for the `server` and `content-service` components and the old default of `1` for all other components.

## Release Notes

```release-note
Allow replica counts for each component to be configurable through the installer
```

## Documentation

None.

### DIFF
--- a/install/installer/pkg/common/common.go
+++ b/install/installer/pkg/common/common.go
@@ -337,6 +337,20 @@ func NodeAffinity(orLabels ...string) *corev1.Affinity {
 	}
 }
 
+func Replicas(ctx *RenderContext, component string) *int32 {
+	replicas := int32(1)
+
+	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
+		if cfg.Common != nil && cfg.Common.PodConfig[component] != nil {
+			if cfg.Common.PodConfig[component].Replicas != nil {
+				replicas = *cfg.Common.PodConfig[component].Replicas
+			}
+		}
+		return nil
+	})
+	return &replicas
+}
+
 // ObjectHash marshals the objects to YAML and produces a sha256 hash of the output.
 // This function is useful for restarting pods when the config changes.
 // Takes an error as argument to make calling it more conventient. If that error is not nil,
--- a/install/installer/pkg/common/render_test.go
+++ b/install/installer/pkg/common/render_test.go
@@ -36,18 +36,22 @@ func TestCompositeRenderFunc_NilObjectsNilError(t *testing.T) {
 func TestReplicas(t *testing.T) {
 	testCases := []struct {
 		Component        string
+		Name             string
 		ExpectedReplicas int32
 	}{
 		{
 			Component:        server.Component,
+			Name:             "server takes replica count from common config",
 			ExpectedReplicas: 123,
 		},
 		{
 			Component:        dashboard.Component,
+			Name:             "dashboard takes replica count from common config",
 			ExpectedReplicas: 456,
 		},
 		{
 			Component:        content_service.Component,
+			Name:             "content_service takes default replica count",
 			ExpectedReplicas: 1,
 		},
 	}
@@ -64,12 +68,14 @@ func TestReplicas(t *testing.T) {
 	require.NoError(t, err)
 
 	for _, testCase := range testCases {
-		actualReplicas := common.Replicas(ctx, testCase.Component)
+		t.Run(testCase.Name, func(t *testing.T) {
+			actualReplicas := common.Replicas(ctx, testCase.Component)
 
-		if *actualReplicas != testCase.ExpectedReplicas {
-			t.Errorf("expected %d replicas for %q component, but got %d",
-				testCase.ExpectedReplicas, testCase.Component, *actualReplicas)
-		}
+			if *actualReplicas != testCase.ExpectedReplicas {
+				t.Errorf("expected %d replicas for %q component, but got %d",
+					testCase.ExpectedReplicas, testCase.Component, *actualReplicas)
+			}
+		})
 	}
 }
 
--- a/install/installer/pkg/common/render_test.go
+++ b/install/installer/pkg/common/render_test.go
@@ -6,6 +6,9 @@ package common
 import (
 	"testing"
 
+	"github.com/gitpod-io/gitpod/installer/pkg/components/content_service"
+	"github.com/gitpod-io/gitpod/installer/pkg/components/dashboard"
+	"github.com/gitpod-io/gitpod/installer/pkg/components/server"
 	"github.com/gitpod-io/gitpod/installer/pkg/config/v1"
 	"github.com/gitpod-io/gitpod/installer/pkg/config/v1/experimental"
 	"github.com/gitpod-io/gitpod/installer/pkg/config/versions"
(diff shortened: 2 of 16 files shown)


sagor999 (approved):

mustard-mh (commented):
this? https://github.com/gitpod-io/gitpod/blob/948ae5192be0dd3d35431e6b0f5b443b1c61cfc1/install/installer/pkg/components/openvsx-proxy/statefulset.go#L43-L44

MrSimonEmms:
@andrew-farries I hope I'm not coming across as Mr Negative, but I'm not sure this is the right approach. The `experimental` section is designed as "Gitpod SaaS-only" stuff but there are going to definitely be large self-hosted instances that will need to have multiple replicas. Generally speaking, we're not going to be supporting the use of the `experimental` section for self-hosted users (with some exceptions).

What might be a better approach is a more nuanced approach.

1. Adding a section to the main config, something like `lite` (I don't like that name, but can't think of anything better at the moment) where it defaults to `false` - if it's enabled, it puts everything into a single replica and if disabled it puts everything that can have multiple replicas to `2` (I believe there are deployments that cannot be replicated, but not sure)
2. Keep the `experimental` section as you've done for more fine-grained configuration

Then the value can be `1` or `2` for self-hosted users with a fairly small amount of configuration or `n` for SaaS users.

I'd also prefer that the `experimental` section didn't have a [CommonConfig](https://github.com/gitpod-io/gitpod/blob/5df4cc277a6c89034ff94c0cd4f08c14717f2764/install/installer/pkg/config/v1/experimental/experimental.go#L23) struct - I think it may be better having `experimental.webapp.dashboard.podConfig.replicas = 2`. That said, as this is for Gitpod SaaS, I'm not going to put up a fight if you want to stick with the current approach.

install/installer/pkg/common/common.go
```diff
@@ -337,6 +337,20 @@ func NodeAffinity(orLabels ...string) *corev1.Affinity {
 	}
 }
 
+func Replicas(ctx *RenderContext, component string) *int32 {
```
easyCZ, line 340: IMO, I feel this makes the replica config more implicit, than explicit. For it to work, you've gotta (somewhat opaquely) specify it in the `PodConfig[component]` config which can be harder to find. 

This will also become somewhat problematic when we actually want to use a pod-autoscaler as this config will no longer make sense as the AutoScaler has a lot more config.

I'm basing this on a similar system we had in the past and it turned out to be very beneficial to just be extremely explicit, without helpers. The power of a full programming language to render config is awesome, but also quite a challenge to ensure we do keep it simple.

easyCZ (commented):
PR looks good, left a comment around general approach to config in go. Would you be able to please also add tests for the `common.Replicas` function? It's shared across many components and I'd rather we did have some tests to start with as we evolve it.

install/installer/pkg/common/common.go
geropl: > IMO, I feel this makes the replica config more implicit, than explicit. For it to work, you've gotta (somewhat opaquely) specify it in the PodConfig[component] config which can be harder to find.

I generally agree on the implicit connection via `component string` part. Here it's helpful that because using the "experimental" section is a bit noisy atm. And it follows the existing pattern, so ok for me, especially as it allows us to move forward. :+1:

geropl (approved):

install/installer/pkg/common/common.go
easyCZ, line 340: Unit test, please.

geropl:
> but I'm not sure this is the right approach. The experimental section is designed as "Gitpod SaaS-only" stuff but there are going to definitely be large self-hosted instances that will need to have multiple replicas

@MrSimonEmms this approach was taken after a sync with @corneliusludmann last week which we scheduled for exactly [this discussion](https://www.google.com/url?q=https://www.notion.so/gitpod/Installer-WebApp-x-Self-Hosted-67cc7661c13c47e98ecd61a47383fe1f&sa=D&source=calendar&ust=1651046832803422&usg=AOvVaw3gcJaUAVtUtcbf2y6uqK2E). In general I agree: This is an option that we _will_ need in for self-hosted at some point. But we decided to solve this in this context, and put it into `experimental` because it's far easier to change afterwards. There is a whole host of config that could go into such a "common"/"generic" podConfig section. But unless we know how that should look like I suggest we decouple that decision from this PR/change. :+1:

geropl:
/hold 

@andrew-farries Could you add [unit tests](https://github.com/gitpod-io/gitpod/pull/9495#discussion_r857367598) as suggested by @easyCZ ? Thank you! :pray:

install/installer/pkg/common/render_test.go
```diff
@@ -28,6 +29,46 @@ func TestCompositeRenderFunc_NilObjectsNilError(t *testing.T) {
...
+	testCases := []struct {
+		Component        string
+		ExpectedReplicas int32
+	}{
+		{
+			Component:        "server",
+			ExpectedReplicas: 123,
+		},
+		{
+			Component:        "dashboard",
```
easyCZ, line 42:
```suggestion
			Component:        dashboard.Component,
```
and similar in other component entries. The test can be used as an example how to specify it so should also "do it as we'd like it to be done"

install/installer/pkg/common/render_test.go
```diff
@@ -28,6 +29,46 @@ func TestCompositeRenderFunc_NilObjectsNilError(t *testing.T) {
...
+			ExpectedReplicas: 456,
+		},
+		{
+			Component:        "content-service",
```
easyCZ, line 46: [recommendation]: Give each test a name as a string, like 
```
name: "defaults to 1 for components without experimental pod config"
```

You can then even do 
```
for _, scenario := range scenarios {
  t.Run(scenario.name, func(t *testing.T) {
     // your test logic
  }))
}
```
Which prints nice errors when a particular table test fails

easyCZ (approved):
Approving as I'm happy with the changes. Left a couple of suggestions but up to you if you want to tackle them or how.

andrew-farries:
> this?
> 
> https://github.com/gitpod-io/gitpod/blob/948ae5192be0dd3d35431e6b0f5b443b1c61cfc1/install/installer/pkg/components/openvsx-proxy/statefulset.go#L43-L44

Thanks, I missed this one.

andrew-farries:
> Also, we'll need to have some validation to make sure that the values are an integer. It's worth seeing what happens if you set the value to a string/boolean - hopefully it'll error/panic, but I don't think it would from reading the code

If the value for `replicas` is set to a string, the installer fails with:

image

install/installer/pkg/common/render_test.go
andrew-farries: [29d51e5](https://github.com/gitpod-io/gitpod/pull/9495/commits/29d51e56a404ab9f96eb55f0b00e888bebf02509)

corneliusludmann (approved):
Looks good to me.

Regarding `experimental` or not: As long as we don't have customers asking for this and we are not sure yet how exactly this config should look like it's much more convenient for us to use the `experimental` section as incubation for this config and move it to the general (public) section once we are fine with that and want to support this officially. That was the reason for the decision.

andrew-farries:
@easyCZ requested a re-review from you as I had to move the tests into an `_test` package ([725fbd0](https://github.com/gitpod-io/gitpod/pull/9495/commits/725fbd066463194c98b6f9981e112768185d36cc)) in order to address [this](https://github.com/gitpod-io/gitpod/pull/9495#discussion_r857530620)

install/installer/pkg/common/render_test.go
andrew-farries: Done in dedf7d3d53e366f93e01a0c804450bab96d8ee19

easyCZ (approved):

mustard-mh (approved):

PR #9498, opened 2022-04-22
[ws-daemon] Fix check logic for I/O limits

aledbf:
## Release Notes
```release-note
<!--
  Add entries for the CHANGELOG.md or "NONE" if there aren't any user facing changes.
  Each line becomes a separate entry.
  Format: [!<optional for breaking>] <description>
  Example: !basic auth is no longer supported
  See https://www.notion.so/gitpod/Release-Notes-513a74fdd23b4cb1b3b3aefb1d34a3e0
-->

```

### DIFF
--- a/components/ws-daemon/pkg/cgroup/plugin_iolimit_v1.go
+++ b/components/ws-daemon/pkg/cgroup/plugin_iolimit_v1.go
@@ -96,13 +96,13 @@ func (c *IOLimiterV1) Apply(ctx context.Context, basePath, cgroupPath string) er
 	}
 
 	writeLimit := func(limitPath string, content []string) error {
-		for _, l := range content {
-			_, err := os.Stat(limitPath)
-			if errors.Is(err, os.ErrNotExist) {
-				log.WithError(err).WithField("limitPath", limitPath).Debug("blkio file does not exist")
-				continue
-			}
+		_, err := os.Stat(limitPath)
+		if errors.Is(err, os.ErrNotExist) {
+			log.WithError(err).WithField("limitPath", limitPath).Debug("blkio file does not exist")
+			return nil
+		}
 
+		for _, l := range content {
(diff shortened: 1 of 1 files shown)


components/ws-daemon/pkg/cgroup/plugin_iolimit_v1.go
```diff
@@ -55,80 +64,49 @@ const (
...
-
-	c.limits = limits{
+func buildLimits(writeBytesPerSecond, readBytesPerSecond, writeIOPs, readIOPs int64) limits {
+	return limits{
 		WriteBytesPerSecond: writeBytesPerSecond,
 		ReadBytesPerSecond:  readBytesPerSecond,
 		WriteIOPS:           writeIOPs,
 		ReadIOPS:            readIOPs,
+
+		cache:   make(map[string][]string),
```
utam0k, line 74: As far as I know, `IOLimiterV1` is a singleton. It means this `cache` is shared by all instances managed by a given ws-damon. Is this as intended?

aledbf: Yes. We don't have different configurations per workspace

utam0k:
Can I ask you to organize your commits on this PR?

aledbf:
> Can I ask you to organize your commits on this PR?

Sure. What do you expect? What you see now is the progression of the tests and issues we found.

utam0k:
> > Can I ask you to organize your commits on this PR?
> 
> Sure. What do you expect? What you see now is the progression of the tests and issues we found.

Thank you. How about summarizing these in the first commit?

components/ws-daemon/pkg/cgroup/plugin_iolimit_v1.go
utam0k: It would be good to comment on why that is okay. Perhaps, but because ws-daemon is on each node, right?

PR #9511, opened 2022-04-24
Add real owner to imagebuild metadata

Furisto:
## Description
Add real owner to imagebuild metadata

## Related Issue(s)
n.a.

## How to test
- Start image build in preview environment e.g. https://fo-imgbuild-owner.staging.gitpod-dev.com/#imagebuild/https://github.com/gitpod-io/gitpod
- Verify that owner label on the pod is the id of the user that triggered the imagebuild and not image-builder
```
Labels:       app=gitpod
              component=workspace
              gitpod.io/networkpolicy=default
              gpwsman=true
              headless=true
              metaID=db0de155-4a44-4d24-874a-f28d219dfeb1
              owner=b8a478cf-8488-44e1-a735-c155086c5095
              workspaceID=db0de155-4a44-4d24-874a-f28d219dfeb1
              workspaceType=imagebuild
```
- Use gcptl imagebuilds list -> imagebuild should still be listable

## Notes for reviews
The push of the imagebuild to the registry fails in the preview environment, but this also happens without my changes.

## Release Notes

```release-note
None
```


components/image-builder-mk3/pkg/orchestrator/monitor.go
```diff
@@ -70,14 +71,17 @@ func (m *buildMonitor) Run() {
 	for {
 		wss, err := m.wsman.GetWorkspaces(ctx, &wsmanapi.GetWorkspacesRequest{
 			MustMatch: &wsmanapi.MetadataFilter{
-				Owner: buildWorkspaceOwnerID,
+				Annotations: map[string]string{
+					annotationManagedBy: buildWorkspaceManagerID,
```
princerachit, line 75: I have not tested the PR. Overall it looks good but I have a question about performance with this change. AFAIU filtering with owner is much faster and done on server side, but annotation based filters are done on client side.

If my understanding is correct, then how much this impacts the performance and in which cases

components/server/src/workspace/workspace-starter.ts
```diff
@@ -931,6 +931,7 @@ export class WorkspaceStarter {
             req.setSource(src);
             req.setAuth(auth);
             req.setForceRebuild(forceRebuild);
+            req.setTriggeredBy(user.id);
```
geropl, line 934: Note that it's not defined what the "real owner" of an image build is, exactly. Calling it "triggered by" though makes sense.

components/image-builder-mk3/pkg/orchestrator/orchestrator.go
```diff
@@ -345,12 +345,11 @@ func (o *Orchestrator) Build(req *protocol.BuildRequest, resp protocol.ImageBuil
...
-					annotationRef:     wsrefstr,
-					annotationBaseRef: baseref,
+					annotationRef:       wsrefstr,
+					annotationBaseRef:   baseref,
+					annotationManagedBy: buildWorkspaceManagerID,
 				},
-				// TODO(cw): use the actual image build owner here and move to annotation based filter
-				//           when retrieving running image builds.
-				Owner: buildWorkspaceOwnerID,
+				Owner: req.GetTriggeredBy(),
```
geropl, line 352: @Furisto Hm, I'm not sure I agree. Having the userId who triggered the image build might make sense to have for debugging purposes. But in every other way it's different to the owner of a regular or prebuild workspace.
What's the motivation to change this? :thinking:

Furisto: @geropl The motivation for this was that Prince and I were debugging imagebuilds last week and we found it difficult to identify which imagebuild we need to look at. I then implemented it like that because based on that comment there was already a plan to do it this way. I am open to do it differently e.g. keeping the image builder as the owner and adding a label or annotation with the user that triggered the imagebuild.

components/image-builder-mk3/pkg/orchestrator/monitor.go
Furisto: Interesting, do you have a source for this? I haven't been able to find one.

princerachit: NVM this GetWorkspaces internally calls server component. I assumed that it was calling k8s api server directly.

BTW here is the [client go](https://github.com/kubernetes/client-go/blob/28ccde769fc5519dd84e5512ebf303ac86ef9d7c/listers/core/v1/pod.go#L50) implementation

components/image-builder-mk3/pkg/orchestrator/orchestrator.go
csweichel: IMHO it makes sense to make use the correct owner (i.e. the user originally triggering the image build). It's very much like prebuilds in that way.

geropl (approved):
Code LGTM, sans the off-topic disucssion about "workspace ownership".

components/image-builder-mk3/pkg/orchestrator/orchestrator.go
csweichel: > So "who owns a prebuiöd/image build" is somewhat in a limbo state atm, at least that's how it feels to me.

Totally agree.

Furisto:
Hey @csweichel can you approve this PR if you do not have any further concerns?

geropl:
@csweichel friendly pin so we can get this one in. :upside_down_face:

csweichel (approved):

csweichel:
sorry for the delay

Issue #9074, opened 2022-04-01
Epic: Allow for Self Signed Certs in Self-Hosted


lucasvaltl:
### Summary

We want users to be able to use self-signed certs in their self-hosted Gitpod installation

### Context

ℹ️ Defining Self-Signed Certs
There are many ways to define self-signed certs, but this is the definition we are using: self-singed certs are any certs that are **not** signed by a public certificate authority (ca) that is already known to linux, i.e. the CA’s certs are already shipped with the linux build. 

Some users want to use components or run in environments that require self-signed certs. We want to allow for this. 

--> This is an epic for #8559 . We think this will require further collaboration between teams, which is why we want to bring it to the epic level. 

### Value

- This unlocks a new type of user / customer for us which previously was unable to use self-hosted Gitpod due to the certificates they used. Further, it facilitates installing Gitpod on your local machine. 

### Acceptance Criteria

Self-signed certs can be propagated into:
1. server-pod
2. ws-daemon container
3. workspace container
4. on the node  to use for containerd
5. Image builder (assuming the registry it talks to is using self signed certs)

We have basic documentation in place and this feature is tested with at least one user. 

### Measurement

- We run nightly tests that use self-signed certs and these run successfully
- At least one customer has successfully used self-signed certs in their installation

### Complexities

On GKE you cannot allow containerd to trust other certificates without restarting containerd. This hinders us from having self-signed certs on GKE right now.

### Child Issues
- [x] https://github.com/gitpod-io/gitpod/issues/8559
- [x] #9077
- [x] #9078
- [x] #9079

MrSimonEmms:
Looks like all the tasks are done so will close.

Reopen if necessary (and add more tasks)

tysonrm:
Why not build in an ACME service that provides trusted certs?

PR #9518, opened 2022-04-25
Fix issues with `self-signed` certificates

Pothulapati:
## Description

This PR includes temporary fixes to get `Certificate` option working with `self-signed` certificates.

Currently, Whenever the `Certificate` config option is set to a specific secret, we seem to assume **by default**
 that they are publicly signed. This currently has the following issues when they are `self-signed` :
- Due to the usage of `Certificate` in internal facing services like `registry-facade`, Internal clients fail to communicate as the self-signed CA is not in their trust store.  (Fix: Update `registry-facade` to use `builtin-registry-facade-cert`, generated by `cert-manager`)
- Due to the fact that internal `registry` communication also flows through the `proxy`, This communication fails as the `proxy` cert is self-signed as its not available in the trust store. (Fix: Update the `registry-facade` daemonset to also incldue `crt` of `Certificate` in its trust store).

This PR also fixes an issue with `image-builder-mk3` when the `customCACert` option is set, by removing `InternalVolumeCAMount` as it makes the `etc/ssl/certs` empty and readOnly (which prevents from ounting the customCACert cert into `etc/ssl/certs`

## Related Issue(s)

Part of #9074 

## How to test

Generate a CA, and self-signed SSL by following https://gist.github.com/Pothulapati/a7c68e073cdbf8573f12b64320711ce5 (or any other online tutorial for that matter)

```bash
# Create the relevant secrets before installing Gitpod
# myCa.pem is what you want to propogate across components to access things external to Gitpod
kubectl create secret generic ca-certificate --from-file=ca.crt=./certs/myCa.pem
# these are the ones created for the domain
kubectl create secret tls https-cert --cert=./certs/ssl.crt --key=./certs/ssl.key

# Set the relevant config options in config file
yq e -i ".certificate.name = \"https-cert\"" "${CONFIG_FILE}"
yq e -i ".certificate.kind = \"secret\"" "${CONFIG_FILE}"
yq e -i ".customCACert.kind = \"secret\"" "${CONFIG_FILE}"
yq e -i ".customCACert.name = \"ca-certificate\"" "${CONFIG_FILE}"

# Install GItpod
gitpod-installer \
        render \
        --config="${CONFIG_FILE}" > gitpod.yaml

# See https://github.com/gitpod-io/gitpod/tree/main/install/installer#error-validating-statefulsetstatus
yq eval-all --inplace \
        'del(select(.kind == "StatefulSet" and .metadata.name == "openvsx-proxy").status)' \
        gitpod.yaml

kubectl apply -f gitpod.yaml
```

## Release Notes

```release-note
Fix multiple issues with `self-signed` certs
```

### DIFF
--- a/install/installer/pkg/common/ca.go
+++ b/install/installer/pkg/common/ca.go
@@ -10,17 +10,8 @@ import (
 
 const (
 	cacerts      = "cacerts"
-	etcSSLCerts  = "/etc/ssl/certs"
 	caVolumeName = "gitpod-ca-certificate"
 )
-
-func InternalCAVolumeMount() *corev1.VolumeMount {
-	return &corev1.VolumeMount{
-		Name:      cacerts,
-		ReadOnly:  true,
-		MountPath: etcSSLCerts,
-	}
-}
 
 func InternalCAVolume() *corev1.Volume {
 	return &corev1.Volume{
--- a/install/installer/pkg/components/image-builder-mk3/deployment.go
+++ b/install/installer/pkg/components/image-builder-mk3/deployment.go
@@ -103,7 +103,6 @@ func deployment(ctx *common.RenderContext) ([]runtime.Object, error) {
 			MountPath: PullSecretFile,
 			SubPath:   ".dockerconfigjson",
 		},
-		*common.InternalCAVolumeMount(),
 	}
 	if vol, mnt, _, ok := common.CustomCACertVolume(ctx); ok {
 		volumes = append(volumes, *vol)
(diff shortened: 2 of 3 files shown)


csweichel (approved):

MrSimonEmms (approved):
Works really well. It also works when using the `ca-issuer-ca`, which is the internal CA that's used to create all the internal certs.

Pothulapati:
@iQQBot This is weird, as we pass the same CA cert is [loaded into kubelets](https://github.com/gitpod-io/gitpod/blob/main/install/installer/pkg/components/registry-facade/daemonset.go#L184) by the same registry-facade daemonset! Wondering why that cert is not being used! 🤔

PR #9541, opened 2022-04-26
[server] fix ide option not exists bug

mustard-mh:
## Description

Fix function `chooseIDE` func not cover not exists IDE id

## Related Issue(s)

Fixes #

## How to test

Open Gitpod with this PR, exec 
```
cd components/server && yarn test
```

## Release Notes

```release-note
NONE
```

## Documentation

- [x] /werft no-preview

### DIFF
--- a/components/server/src/workspace/workspace-starter.spec.ts
+++ b/components/server/src/workspace/workspace-starter.spec.ts
@@ -249,5 +249,19 @@ describe("workspace-starter", function () {
             const result = chooseIDE("unknown-custom", customOptions, useLatest, hasPerm);
             expect(result.ideImage).to.equal(ideOptions.options["code"].latestImage);
         });
+
+        it("not exists ide with custom permission", function () {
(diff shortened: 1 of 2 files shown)


mustard-mh:
> How to reproduce the original issue from UI?

We can't. This is part of https://github.com/gitpod-io/gitpod/pull/9467 RP's commit

mustard-mh:
I wrote function `chooseIDE` first and add test cases. So test cases can be passed of course. But other test cases I don't cover may not pass (like this🙈). 

How can we avoid this to happen? Write test cases first to drive code? Ask someone else to input more test cases?

cc @akosyakov

akosyakov:
@geropl could you have a look?

geropl:
> @geropl could you have a look?

@akosyakov  Will do after all-hands. :+1:

geropl (approved):
Code LGTM, tests are green :heavy_check_mark:

PR #9542, opened 2022-04-26
Use the right `registry-facade` port in preview environments

Pothulapati:
## Description

After #9518 , `StartWorkspace` seems to fail on preview environments, as we randomize ports in the preview environment but we specifically use the port number in the `update-containerd-certificates` init Contianer to specify the cert for that `URL:PORT`.

This PR fixes that by passing the same into the `update-containerd-ceritificates` during post processing.

## Related Issue(s)

Follow up to #9518 

## How to test

Preview environments should be able to start and run workspaces.

## Release Notes

```release-note
NONE
```


Pothulapati:
@iQQBot It doesn't seem to update the post processing script to point to the right port for some reason!

PR #9547, opened 2022-04-26
[common-go] Add keyed gRPC rate limits

csweichel:
## Description
This PR adds support for message-keyed gRPC rate limits to common-go.
Using this functionality, we can e.g. add per-user rate limits in ws-manager.

For example:
```JSON
{
    "rateLimits": {
        "/wsman.WorkspaceManager/StartWorkspace": {
            "bucketSize": 5,
            "refillInterval": "1s",
            "key": "metadata.owner",
            "keyCacheSize": 200
        }
    }
}
```

would impose a 5 workspaces per second rate limit per workspace owner.

## How to test
- See unit test
- Manually add the config to ws-manager and use loadgen

## Release Notes

```release-note
[ws-manager] Support more fine-grained rate limits
```

### DIFF
--- a/components/common-go/grpc/ratelimit.go
+++ b/components/common-go/grpc/ratelimit.go
@@ -6,60 +6,143 @@ package grpc
 
 import (
 	"context"
+	"strings"
 	"time"
 
 	"golang.org/x/time/rate"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
 
 	"github.com/gitpod-io/gitpod/common-go/util"
+	lru "github.com/hashicorp/golang-lru"
 )
+
+type keyFunc func(req interface{}) (string, error)
 
 // RateLimit configures the reate limit for a function
 type RateLimit struct {
 	Block          bool          `json:"block"`
 	BucketSize     uint          `json:"bucketSize"`
 	RefillInterval util.Duration `json:"refillInterval"`
+
+	KeyCacheSize uint   `json:"keyCacheSize,omitempty"`
+	Key          string `json:"key,omitempty"`
+}
+
+func (r RateLimit) Limiter() *rate.Limiter {
+	return rate.NewLimiter(rate.Every(time.Duration(r.RefillInterval)), int(r.BucketSize))
 }
 
 // NewRatelimitingInterceptor creates a new rate limiting interceptor
 func NewRatelimitingInterceptor(f map[string]RateLimit) RatelimitingInterceptor {
-	funcs := make(map[string]ratelimitedFunction, len(f))
+	funcs := make(map[string]*ratelimitedFunction, len(f))
 	for name, fnc := range f {
-		funcs[name] = ratelimitedFunction{
-			Block: fnc.Block,
-			L:     rate.NewLimiter(rate.Every(time.Duration(fnc.RefillInterval)), int(fnc.BucketSize)),
+		var (
+			keyedLimit *lru.Cache
+			key        keyFunc
+		)
+		if fnc.Key != "" && fnc.KeyCacheSize > 0 {
+			keyedLimit, _ = lru.New(int(fnc.KeyCacheSize))
+			key = fieldAccessKey(fnc.Key)
+		}
+
+		funcs[name] = &ratelimitedFunction{
+			RateLimit:   fnc,
+			GlobalLimit: fnc.Limiter(),
+			Key:         key,
+			KeyedLimit:  keyedLimit,
 		}
 	}
 	return funcs
 }
 
+func fieldAccessKey(key string) keyFunc {
+	return func(req interface{}) (string, error) {
+		msg, ok := req.(proto.Message)
+		if !ok {
+			return "", status.Errorf(codes.Internal, "request was not a protobuf message")
+		}
+
+		val, ok := getFieldValue(msg.ProtoReflect(), strings.Split(key, "."))
+		if !ok {
+			return "", status.Errorf(codes.Internal, "Field %s does not exist in message. This is a rate limiting configuration error.", key)
+		}
+		return val, nil
+	}
+}
+
+func getFieldValue(msg protoreflect.Message, path []string) (val string, ok bool) {
+	if len(path) == 0 {
+		return "", false
+	}
+
+	field := msg.Descriptor().Fields().ByName(protoreflect.Name(path[0]))
+	if field == nil {
+		return "", false
+	}
+	if len(path) > 1 {
+		if field.Kind() != protoreflect.MessageKind {
+			// we should go deeper but the field is not a message
+			return "", false
+		}
+		child := msg.Get(field).Message()
+		return getFieldValue(child, path[1:])
+	}
+
+	if field.Kind() != protoreflect.StringKind {
+		// we only support string fields
+		return "", false
+	}
+
+	return msg.Get(field).String(), true
+}
+
 // RatelimitingInterceptor limits how often a gRPC function may be called. If the limit has been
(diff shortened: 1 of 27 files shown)


test/go.sum
```diff
@@ -1429,15 +1426,16 @@ k8s.io/kubectl v0.23.5/go.mod h1:lLgw7cVY8xbd7o637vOXPca/w6HC205KsPCRDYRCxwE=
...
 k8s.io/utils v0.0.0-20210802155522-efc7438f0176/go.mod h1:jPW/WVKK9YHAvNhRxK0md/EJ228hCsBRufyofKtW8HA=
 k8s.io/utils v0.0.0-20211116205334-6203023598ed h1:ck1fRPWPJWsMd8ZRFsWc6mh/zHp5fZ/shhbrgPUxDAE=
 k8s.io/utils v0.0.0-20211116205334-6203023598ed/go.mod h1:jPW/WVKK9YHAvNhRxK0md/EJ228hCsBRufyofKtW8HA=
 rsc.io/binaryregexp v0.2.0/go.mod h1:qTv7/COck+e2FymRvadv62gMdZztPaShugOCi3I+8D8=
 rsc.io/quote/v3 v3.1.0/go.mod h1:yEA65RcK8LyAZtP9Kv3t0HmxON59tX3rD+tICJqUlj0=
 rsc.io/sampler v1.3.0/go.mod h1:T1hPZKmBbMNahiBKFy5HrXp6adAjACjK9JXDnKaTXpA=
 sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.30/go.mod h1:fEO7lRTdivWO2qYVCVG7dEADOMo/MLDCVr8So2g88Uw=
-sigs.k8s.io/controller-runtime v0.11.2 h1:H5GTxQl0Mc9UjRJhORusqfJCIjBO8UtUxGggCwL1rLA=
-sigs.k8s.io/controller-runtime v0.11.2/go.mod h1:P6QCzrEjLaZGqHsfd+os7JQ+WFZhvB8MRFsn4dWF7O4=
+sigs.k8s.io/controller-runtime v0.9.0 h1:ZIZ/dtpboPSbZYY7uUz2OzrkaBTOThx2yekLtpGB+zY=
```
aledbf, line 1437: Please revert this change

test/go.mod
```diff
@@ -139,10 +140,6 @@ require (
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
-replace sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.11.0
```
aledbf, line 142: Please revert this change

csweichel: Which one is correct?
```

replace sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.11.0

replace sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.11.2
```

aledbf: v0.11.2

easyCZ:
https://github.com/grpc-ecosystem/go-grpc-middleware offers a rate limit implementation, this may be possibly cheaper to use than building our own

csweichel:
> https://github.com/grpc-ecosystem/go-grpc-middleware offers a rate limit implementation, this may be possibly cheaper to use than building our own.
> 
> I've used it in the past and worked well. It has built in support for sharing the rate limits across instances (but can be used without it).

We looked at that implementation prior to rolling our own.
AFAIK it does not support
- per method rate limiting
- "keyed" rate limiting as is introduced in this PR.

easyCZ (commented):
Looking good, left a couple comments. Could you please also add metrics to expose the following:
1. Rate limited count, by method
2. Hit rate on the LRU

Can be in seperate follow-up PRs, but we should be able to observe this before we deploy it to prod.

components/common-go/grpc/ratelimit.go
```diff
@@ -6,60 +6,143 @@ package grpc
...
+		var limit *rate.Limiter
+		if f.Key == nil {
+			limit = f.GlobalLimit
+		} else {
+			key, err := f.Key(req)
+			if err != nil {
+				return nil, err
+			}
+
+			f.KeyedLimit.ContainsOrAdd(key, f.RateLimit.Limiter())
```
geropl, line 132: question: this immediately throws creates and trashes a limiter struct on stack AFAIU which could be avoided by not using `ContainsOrAdd`.

csweichel: Agreed - not doing it this way requires extra locking because this operation can happen concurrently. 

We could also ignore the concurrency for the time being and go with a `Peek` followed by an `Add`, running this risk to mis-account for a user during very high call-rate situations.

geropl (approved):
Code LGTM, and seems to work. Left minor comments.

components/common-go/grpc/ratelimit.go
```diff
@@ -6,60 +6,143 @@ package grpc
...
+	RateLimit RateLimit
+
+	GlobalLimit *rate.Limiter
+	Key         keyFunc
+	KeyedLimit  *lru.Cache
 }
 
 // UnaryInterceptor creates a unary interceptor that implements the rate limiting
 func (r RatelimitingInterceptor) UnaryInterceptor() grpc.UnaryServerInterceptor {
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
```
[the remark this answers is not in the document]
csweichel: they indicate what's returned though :)

csweichel:
Added metrics

components/common-go/grpc/ratelimit.go
```diff
@@ -6,60 +6,201 @@ package grpc
...
+}
+
+func (r RateLimit) Limiter() *rate.Limiter {
+	return rate.NewLimiter(rate.Every(time.Duration(r.RefillInterval)), int(r.BucketSize))
 }
 
 // NewRatelimitingInterceptor creates a new rate limiting interceptor
 func NewRatelimitingInterceptor(f map[string]RateLimit) RatelimitingInterceptor {
-	funcs := make(map[string]ratelimitedFunction, len(f))
+	callCounter := prometheus.NewCounterVec(prometheus.CounterOpts{
```
easyCZ, line 42: I believe the metrics here also need to be registered, to be picked up by prometheus, otherwise they are just local counters which won't show up on `/metrics` endpoint.

See https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Register

easyCZ: In this case, I'd actually hoist the counter definition outside of this instantiation and make it package private variable. You can use https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/promauto to simplify the registration.

sagor999 (approved):
just need to fix metrics name copy paste error.

components/common-go/grpc/ratelimit.go
easyCZ: My bad, missed this. Normally, I'd expect the registration to happen in this file and automatically without requiring an extra call. The risk here is that a component uses the rate limiting interceptor but doesn't register the metric.

The other issue is that this would panic on startup, but CI doesn't catch this.

csweichel: For a while now we've adopted the pattern to make things implement the `prometheus.Collector` interface so that they can be registered.
Testing is just as easy as passing in the registry, but it doesn't mandate that someone uses metrics.
The "registration error at runtime" behaviour is also the same.

csweichel:
fixed the metrics name

components/common-go/grpc/ratelimit.go
easyCZ: Thanks. I'd personally like the need for registration to be more explicit (and to use DI to pass the registry in) but given this is an existing pattern, better to stick to it than introduce yet another. Thanks for the info.

PR #9553, opened 2022-04-26
Update `registry-facade` to use `https-certificate`

Pothulapati:
## Description

This updates the `GITPOD_CA_CERT` loading to also use the same customCA
cert, instead of loading `builtin-registry-facade-cert` which is not
used anywhere for now.

This is done to fix the GKE regression of loading certs into
`docker` and `containerd`.

## Related Issue(s)

Fixes the current issues in the preview environment 

Follow up to https://github.com/gitpod-io/gitpod/pull/9518

## How to test

Preview Environments should work as expected!

## Release Notes

```release-note
NONE
```

### DIFF
--- a/install/installer/pkg/components/registry-facade/daemonset.go
+++ b/install/installer/pkg/components/registry-facade/daemonset.go
@@ -37,33 +37,34 @@ func daemonset(ctx *common.RenderContext) ([]runtime.Object, error) {
 		volumeMounts []corev1.VolumeMount
 	)
 
-	name := "config-certificates"
-	volumes = append(volumes, corev1.Volume{
-		Name: name,
-		VolumeSource: corev1.VolumeSource{
-			Secret: &corev1.SecretVolumeSource{
-				SecretName: common.RegistryFacadeTLSCertSecret,
-			},
-		},
-	})
-
-	volumeMounts = append(volumeMounts, corev1.VolumeMount{
-		Name:      name,
-		MountPath: "/mnt/certificates",
-	})
-
-	// Attach the proxy CA certificate as registry-facade seems to talk to
-	// the registry through the `proxy`
 	if ctx.Config.Certificate.Name != "" {
-		volumeName := "proxy-ca-cert"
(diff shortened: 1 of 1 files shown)


csweichel:
Is the purpose of this change to enable registry-facade in GKE for all but the self-signed case?

Pothulapati:
@csweichel Yep, It makes it work in all cases except for GKE by 
- Updating `registry-facade` to coninue using `https-certificate` (and thus fixing the problem with GKE)
- Update `registry-facade` to also load `https-certificate`'s CA i.e `customCACert` into `docker` and `contaienrd` and thus making self-signed work in other cases.

MrSimonEmms (approved):
Have tested in GKE (public-signed certs) and Azure (public and self-signed certs) and was able to build an image and start a workspace 🎉

csweichel:
If we can load the `customCACert` in GKE, then why can't we use the cert-manager produced cert in all cases?

cc @aledbf

aledbf:
> This was caused because the update-containerd-ceritifcates initContianer doesn't really work in the GKE case, and hence GKE with self-signed doesn't really work.

What's the issue? Please post some logs

Pothulapati:
> If we can load the customCACert in GKE, then why can't we use the cert-manager produced cert in all cases?

@csweichel We can't load even in that case! Hence, GKE works only in the publicly signed case! (Loading is not even needed here)

We load `customCACert` for other self-signed cases (i.e EKs, etc) as we  use `Certificate` as the TLS cert of `registry-facade`! (and Kubelets need to be able to talk to `registry-facade`)

csweichel:
@aledbf the GKE problem is the same as usual: containerd is not picking up the certs from `certs.d` because CRI is not configured properly.

Basically this change enables Gitpod on GKE when publicly signed certs are used. Today (with the cert-manager produced certs) we cannot use Gitpod on GKE. Prod uses `https-certificates` as well.

csweichel (approved):

PR #9555, opened 2022-04-26
Use configured reverse proxy port rather than fixed port

jenting:
## Description
The reverse proxy port is passed into the script `ssh-proxy-command.sh`, it shouldn't be a fixed port `8022`.

## Related Issue(s)
No

## Release Notes

```release-note
No
```

## Documentation

No


mads-hartmann (approved):
Good catch, thanks!

PR #9557, opened 2022-04-26
Build: never lose STDOUT or STDERR when processes fail

meysholdt:
## Description
When a Werft job fails, we always need to have all information available that hint at potential causes. 
Thus:
* don't lose STDOUT when there is STDERR.
* don't lose STDOUT and STDERR when no slice is given.

## Release Notes

```release-note
NONE
```


mads-hartmann (approved):
Good catch, changes LGTM 🛹

PR #9566, opened 2022-04-26
[kots]: add configuration for using a custom CA certificate

### LINKED ISSUE #8559: Support own CA certificate / self-signed certificated for self-hosted installations

MrSimonEmms:
## Description

Add support for custom CA certs to the KOTS interface. There are three ways of configuring a cert with KOTS

1. Use cert-manager
2. Use a self-signed cert
3. Upload your own certificate - may be made with public or private CA

For `1`, no option is provided for a custom CA. It is assumed that this will be made against a public CA - we can keep tabs on whether there's a requirement for this in future. 🛹🛹🛹

image

For `2`, this self-signed cert is generated with the same CA used to generate the internal certs (`ca-issuer-ca`). This PR adds a note to explain how to extract the CA cert to import into the browser. It's expected that this will be used for limited numbers of users.

image

For `3`, this adds an optional parameter to add the CA when uploading your certs.

image

## Related Issue(s)

Fixes #8559 

## How to test

Deploy with KOTS. Use version `sje-kots-self-signed-certs.4` in `dev-sje` channel (currently the most recent version in there).

## Release Notes

```release-note
[kots]: add configuration for using a custom CA certificate
```

### DIFF
--- a/install/kots/manifests/gitpod-ca-secret.yaml
+++ b/install/kots/manifests/gitpod-ca-secret.yaml
@@ -1,0 +1,14 @@
+# Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+# Licensed under the MIT License. See License-MIT.txt in the project root for license information.
+
+apiVersion: v1
+kind: Secret
+metadata:
+  name: ca-certificate
+  labels:
+    app: gitpod
+    component: gitpod-installer
+  annotations:
+    kots.io/when: '{{repl and (ConfigOptionEquals "cert_manager_enabled" "0") (ConfigOptionNotEquals "tls_ca_crt" "") }}'
+data:
+  ca.crt: '{{repl ConfigOption "tls_ca_crt" }}'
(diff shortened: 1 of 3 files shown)


corneliusludmann (approved):
Changes in this PR look good.

Tested both self-signed option as well as own self-signed generated certs with CA. All pods are running, and access to dashboard works. For k3s I needed to configure [private registry](https://rancher.com/docs/k3s/latest/en/installation/private-registry/). However, registry-facade has a TLS error during image build. Need to investage this further but I think it is not an issue of this PR.

PR #9579, opened 2022-04-27
Overwrite existing k3s entries in the kubectx in install-k3s-kubeconfig.sh

mads-hartmann:
## Description

Currently install-k3s-kubeconfig.sh would not update the .kube/config if it already had an entry to the k3s cluster. This means that your kubectx config can become stale if you change branches or use with-clean-slate-deployment.

The fix was the ensure that K3S_KUBECONFIG_PATH entries takes precedence over KUBECONFIG_PATH when merging the kubeconfig.

## Related Issue(s)

No issue, was reported in our internal Slack [here](https://gitpod.slack.com/archives/C032A46PWR0/p1651044970509919).

## How to test

I tested this manually by using two branches with VMs and re-running the script after having changed branches. Before this changeset it didn't update the kubectx after the first invocation. With this changeset it did.

## Release Notes

```release-note
NONE
```

## Documentation

N/A


easyCZ (approved):

meysholdt (approved):
Code changes LGTM

PR #9588, opened 2022-04-27
[public-api] Simple CLI tool to interact with Public API

easyCZ:
## Description

Adds `api workspaces get --id 123 --token "foo-bar" to interact with Public API. Creates a dir `api` under `cmd` and gives code-ownership to WebApp.

```bash
 ./gpctl 
Gpctl controls a Gitpod installation

Usage:
  gpctl [command]

Available Commands:
  api         Interact with Public API services
  clusters    Controls and inspects cluster
  completion  Generate completion script
  debug       Helps with debugging a component
  help        Help about any command
  imagebuilds Controls and inspects workspace Docker image builds
  workspaces  Controls and inspects workspaces in the Gitpod installation

Flags:
  -h, --help                     help for gpctl
  -c, --kubeconfig string        Path to the kubeconfig file used for connecting to the cluster (default "$HOME/.kube/config")
  -o, --output-format string     Output format. One of: string|json|jsonpath|template (default "template")
      --output-template string   Output format Go template or jsonpath. Use with -o template or -o jsonpath

Use "gpctl [command] --help" for more information about a command.

```

```bash
./gpctl api workspaces get --id 123 --token foo-bar --address api.mp-papi-cmd-yolo.staging.gitpod-dev.com:443
INFO[0000] Estabilishing gRPC connection against api.mp-papi-cmd-yolo.staging.gitpod-dev.com:443 
{"result":{"workspaceId":"123","ownerId":"mock_owner","projectId":"mock_project_id","context"{"contextUrl":"https://github.com/gitpod-io/gitpod"},"description":"This is a mock response"}}%  
```

## Related Issue(s)

Fixes #

## How to test

**Test the CLI works**
```
cd dev/gpctl
go run main.go api workspaces get --id 123 --token foo-bar --address api.mp-papi-cmd-yolo.staging.gitpod-dev.com:443
```

**Test with a local instance of Public API**
1. Navigate to [preview env on main](https://main.preview.gitpod-dev.com/workspaces)
2. Open dev console and run 
```js
	"function:getWorkspace",
	"function:getWorkspaces",
	"resource:default",]})
```
, this will obtain a token you can use to access the through a CLI
4. Start the Public API locally with `cd components/public-api-server && go run main.go`
5. Hit the local API using the CLI, here we need to use `--insecure` because we're not serving TLS certs
```
go run main.go api workspaces get --id $WORKSPACE_ID --token $TOKEN --insecure --address localhost:9501
```
You'll likely get Not Found. To get a response, start a workspace on the main preview env and use its ID as `$WORKSPACE_ID`

## Release Notes

```release-note
NONE
```

## Documentation

NONE

/uncc

- [x] /werft without-vm=true

### DIFF
--- a/dev/gpctl/cmd/api/workspaces.go
+++ b/dev/gpctl/cmd/api/workspaces.go
@@ -1,0 +1,94 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package api
+
+import (
+	"context"
+	"fmt"
+	"github.com/gitpod-io/gitpod/common-go/log"
+	v1 "github.com/gitpod-io/gitpod/public-api/v1"
+	"github.com/spf13/cobra"
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
+	"io"
+	"os"
+)
+
+func newWorkspacesCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "workspaces",
+		Short: "Retrieve information about workspaces",
+	}
+
+	cmd.AddCommand(newWorkspacesGetCommand())
+
+	return cmd
+}
+
+func newWorkspacesGetCommand() *cobra.Command {
+	var workspaceID string
+
+	cmd := &cobra.Command{
+		Use:     "get",
+		Short:   "Retrieve details about a workspace by ID",
+		Example: "  get --workspace-id 1234",
+		Run: func(cmd *cobra.Command, args []string) {
+			if workspaceID == "" {
+				log.Log.Fatal("no workspace id specified, use --id to set it")
+			}
+
+			if err := connOpts.Validate(); err != nil {
+				log.Log.WithError(err).Fatal("Invalid connections options.")
+			}
+
+			conn, err := newConn(connOpts)
+			if err != nil {
+				log.Log.WithError(err).Fatal("Failed to establish gRPC connection")
+			}
+
+			workspace, err := getWorkspace(cmd.Context(), conn, workspaceID)
+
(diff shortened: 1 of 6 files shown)


csweichel:
should be part of `gpctl` imho

easyCZ:
@csweichel That's a good point. How do we want to use the first iteration? Should it eventually be a CLI that externals should be able to interact with to consume the API? Or do we want to merge it with `gpctl` which has an explicit purpose of being used by Gitpod engineers to operate the system.

My inclination is to keep it separate and gear more towards a CLI for the API only, which can be published and consumed by externals.

csweichel:
@easyCZ I'd start with the latter - go for `gpctl`. It gives you more freedom to move quickly without having to consider breaking changes or the experience of others. If we see a need for a CLI for our public API we can always transform that part of `gpctl`.

Also, we're noticing a trend where `gpctl` might become part of the product down the road, if just for "integrators". There's a lot of functionality in there already that's handy in e.g. a self-hosted context.

easyCZ:
@csweichel @geropl I've now moved this under `gpctl`. I've put it under an `api` directory to slightly better manage code-ownership and allow us to make changes without needing workspace reviews.

laushinka:
I've tested this and it works like a charm. Can't wait for this to be merged 😍 Thanks for this, @easyCZ!

dev/gpctl/cmd/api/workspaces.go
```diff
@@ -0,0 +1,94 @@
...
+	return cmd
+}
+
+func newWorkspacesGetCommand() *cobra.Command {
+	var workspaceID string
+
+	cmd := &cobra.Command{
+		Use:     "get",
+		Short:   "Retrieve details about a workspace by ID",
+		Example: "  get --workspace-id 1234",
```
sagor999, line 37: Fix param name:
```suggestion
		Example: "  get --id 1234",
```

sagor999 (approved):
LGTM except for a minor nit.

/hold

PR #9596, opened 2022-04-27
ws-manager-mk2 - or what if we built it today?

csweichel:
## Description
This PR is the initial work for getting ws-manager-mk2 off the ground. It can start workspaces and do initialize the content. Communication between ws-manager, ws-daemon and ws-proxy is done through a workspace CRD. This is not a production grade implementation yet. For milestones see the [milestone plan](https://www.notion.so/gitpod/Milestones-ws-manager-mk2-fb471d7e01e4494b9bca9ad1cd50b3b2). Features not yet working will be added in subsequent PRs.

<a href="https://www.loom.com/share/4cfb4cc8dceb4153b1048d01544f4898">
    <p>PoC: ws-manager-mk2 - Watch Video</p>
    
  </a>

When we built ws-manager more then three years ago, kubebuilder wasn't a thing and writing Kubernetes controllers was hard(er). Also, we didn't know too much about Kubernetes at the time.

This PR shows how ws-manager would look like if it were written as a Kubernetes controller, CRDs and all. The main takeway: **it terrifyingly simplifies things**. The code is orders of magnitudes easier to read and maintain compared to what we have today.

If we just look at the lines of code (a weak, but easily quantifiable proxy for complexity):
```bash
# ws-manager-mk2: 7906 SLOC
$ find ws-manager/pkg/manager -name '*.go' | xargs cat | wc -l
7906

# ws-manager-mk2: 2625 SLOC
$ find ws-manager-mk2/controllers/ -name '*.go' | xargs cat | wc -l
1773
$ find ws-manager-mk2/service/ -name '*.go' | xargs cat | wc -l
852
```

The version in this branch is nearly happy-path feature complete:
- it is able to start and stop workspaces
- supports headless workspaces, e.g. image builds and maybe prebuilds
- it talks to ws-daemon to do content init and disposal
- it can handle ports,
- it acts as imagespec provider for registry-facade,
- it's a drop-in replacement for ws-manager; uses the exact same config an provides the very same API.

I've also extended ws-proxy to use the workspace CRs as workspace info source.

There's a bunch of things missing in this branch:
- tests of any kind really
- better non-happy path handling
- did I mention tests?
- it's not integrated into the installer yet

### Ok, but why?
Other than the considerable reduction in complexity this approach gets us in ws-manager,
- it simplifies the workspace side as a whole, because we can build the in-cluster interactions using CRs rather than gRPC. Much like ws-proxy functions today, ws-daemon and image-builder could interact on the custom resource rather than speak gRPC. This greatly simplifies state management.
- it lets us implement functionality that's independent of the pod's lifecycle, e.g. image builds become much easier to do.
- it lets us implement workspaces which are not pods. Today, all workspaces must be a pod. If we ever want to move to other means of executing workspaces, we'd have a ton of work ahead (think workspaces as harvester VMs? :D - not an actual suggestion).

### Do you not have better things to do?
This was primarily an evening/weekend project - less than 20h have gone into this.

## How to review
- Check for adverse side effects. This PR should not affect the function of Gitpod in any way. It's merely the foundation for things to come.
- Workspaces don't actually start yet using mk2. This is expected. We need to get this changeset off a feature branch though.

## Release Notes

```release-note
NONE
```

Relates with https://github.com/gitpod-io/gitpod/issues/11416

/werft with-vm=true
/werft with-preview=true
/werft with-wsman-mk2=true


components/ws-manager-api/go/crd/v1/workspace_types.go
```diff
@@ -0,0 +1,226 @@
...
+	// the total number of untracked files
+	TotalUntrackedFiles int64 `json:"total_untracked_files,omitempty"`
+	// unpushed_commits is the number of unpushed changes in the workspace, possibly truncated
+	UnpushedCommits []string `json:"unpushed_commits,omitempty"`
+	// the total number of unpushed changes
+	TotalUnpushedCommits int64 `json:"total_unpushed_commits,omitempty"`
+}
+
+type WorkspaceRuntimeStatus struct {
+	NodeName string `json:"nodeName,omitempty"`
```
aledbf, line 197: Here we should use references to the Pod object, like using the [same approach as k8s Gateway](https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1beta1/object_reference_types.go#L94-L133)

components/ws-manager-api/go/crd/v1/workspace_types.go
```diff
@@ -0,0 +1,226 @@
...
+	// +kubebuilder:validation:Required
+	Visibility AdmissionLevel `json:"visibility"`
+}
+
+// WorkspaceStatus defines the observed state of Workspace
+type WorkspaceStatus struct {
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	PodStarts  int    `json:"podStarts"`
```
aledbf, line 110: what does this mean?

components/ws-manager-api/go/crd/v1/workspace_types.go
csweichel: This version of the rewrite still attempts to integrate the "legacy content init" using ws-daemon. We could simplify the thing if we just went for PVC straight away.

components/ws-manager-api/go/crd/v1/workspace_types.go
csweichel: the number of pod restarts

components/ws-manager-api/go/crd/v1/workspace_types.go
```diff
@@ -0,0 +1,226 @@
...
+	Headless   bool   `json:"headless"`
+	URL        string `json:"url,omitempty"`
+	OwnerToken string `json:"ownerToken,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	Phase WorkspacePhase `json:"phase,omitempty"`
```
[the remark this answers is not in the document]
csweichel: IMHO the phase is a rough description of where the workspace is at, and conditions detail that. Akin to how there's a phase and conditions on a pod.

components/ws-manager-api/go/crd/v1/workspace_types.go
csweichel: For e.g. implementing prebuilds on spot instances. Whenever we need to control workspace restarts on the workspace side.

components/ws-manager-api/go/crd/v1/workspace_types.go
```diff
@@ -0,0 +1,226 @@
...
+	WorkspacePhasePending      WorkspacePhase = "pending"
+	WorkspacePhaseImageBuild   WorkspacePhase = "imagebuild"
+	WorkspacePhaseCreating     WorkspacePhase = "creating"
+	WorkspacePhaseInitializing WorkspacePhase = "initializing"
+	WorkspacePhaseRunning      WorkspacePhase = "running"
+	WorkspacePhaseStopping     WorkspacePhase = "stopping"
+	WorkspacePhaseStopped      WorkspacePhase = "stopped"
+)
+
+type WorkspaceConditions struct {
```
aledbf, line 138: I think we need to talk about this. We should follow the same behavior than k8s proposes, an array, like we can see here https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1beta1/gateway_types.go#L497 (or any k8s objecT)

components/ws-manager-api/go/crd/v1/workspace_types.go
aledbf: >  if we just went for PVC straight away.

I think this is not optional ;)

components/ws-manager-api/go/crd/v1/workspace_types.go
csweichel: That makes sense - getting closer to common practice is the right way.
What's the convention for making consumers of the API aware of the expected conditions?

aledbf: We need to define that and use `metav1.Condition` `Type` (to define) and use the `Status` to signal in which part of the lifecycle we are and the status, like, WaitingForSupervisor=true or Ring0=false (to signal something wrong with the pod)

I think this is what we need to focus as the first step.

components/ws-daemon/pkg/container/containerd.go
```diff
@@ -198,12 +198,25 @@ func (s *Containerd) handleNewContainer(c containers.Container) {
...
+			}
+		} else {
+			// this is a ws-manager-mk2 workspace
+			info = &containerInfo{
+				InstanceID:  c.Labels["gitpod.io/instanceID"],
+				OwnerID:     c.Labels[wsk8s.OwnerLabel],
+				WorkspaceID: c.Labels[wsk8s.WorkspaceIDLabel],
+				PodName:     podName,
+			}
 		}
```
jenting, line 218:
```go
                var instanceID string
                if _, ok := c.Labels["gpwsman"]; ok {
                    // this is a ws-manager-mk1 workspace
                    instanceID = c.Labels[wsk8s.WorkspaceIDLabel]
                } else {
                    // this is a ws-manager-mk2 workspace
                    instanceID = c.Labels["gitpod.io/instanceID"]
                }

                info := &containerInfo{
                    InstanceID:  instanceID,
                    OwnerID:     c.Labels[wsk8s.OwnerLabel],
		    WorkspaceID: c.Labels[wsk8s.MetaIDLabel],
		    PodName:     podName,
                }
```

components/ws-manager-api/go/crd/v1/workspace_types.go
```diff
@@ -0,0 +1,256 @@
...
+	Snapshot string `json:"snapshot,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	GitStatus *GitStatus `json:"git,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	Runtime *WorkspaceRuntimeStatus `json:"runtime,omitempty"`
+}
+
+// +kubebuilder:validation:Enum=Deployed;Failed;Timeout;UserActivity;HeadlessTaskFailed;StoppedByRequest;EverReady;BackupComplete;BackupFailure
```
jenting, line 133:
```suggestion
// +kubebuilder:validation:Enum=Deployed;Failed;Timeout;UserActivity;HeadlessTaskFailed;StoppedByRequest;EverReady;BackupComplete;BackupFailure;PVCExists
```

components/ws-manager-api/go/crd/v1/workspace_webhook.go
```diff
@@ -0,0 +1,65 @@
...
+func (r *Workspace) Default() {
+	workspacelog.Info("default", "name", r.Name)
+
+	// TODO(user): fill in your defaulting logic.
+}
+
+// TODO(user): change verbs to "verbs=create;update;delete" if you want to enable deletion validation.
+//+kubebuilder:webhook:path=/validate-workspace-gitpod-io-v1-workspace,mutating=false,failurePolicy=fail,sideEffects=None,groups=workspace.gitpod.io,resources=workspaces,verbs=create;update,versions=v1,name=vworkspace.kb.io,admissionReviewVersions=v1
+
+var _ webhook.Validator = &Workspace{}
```
jenting, line 39: We implement the validation webhook.
Is there any scenario in which we probably need to mutation the webhook?

components/ws-manager-api/go/crd/v1/workspace_types.go
```diff
@@ -0,0 +1,256 @@
...
+	WorkspaceConditionEverReady WorkspaceCondition = "EverReady"
+
+	// BackupComplete is true once the backup has happened
+	WorkspaceConditionBackupComplete WorkspaceCondition = "BackupComplete"
+
+	// BackupFailure contains information about the backup failure
+	WorkspaceConditionBackupFailure WorkspaceCondition = "BackupFailure"
+
+	// WorkspaceConditionPVCExists is true once the PVC of a workspace exists
+	WorkspaceConditionPVCExists WorkspaceCondition = "PVCExists"
```
jenting, line 166: What's the reason we need this condition?

components/ws-manager-api/go/crd/v1/workspace_webhook.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+// log is for logging in this package.
+var workspacelog = logf.Log.WithName("workspace-resource")
```
jenting, line 15:
```suggestion
var workspacelog = logf.Log.WithName("workspace-webhook")
```

components/ws-manager-mk2/controllers/workspace_controller.go
```diff
@@ -0,0 +1,876 @@
...
+
+	var workspacePods corev1.PodList
+	err = r.List(ctx, &workspacePods, client.InNamespace(req.Namespace), client.MatchingFields{wsOwnerKey: req.Name})
+	if err != nil {
+		log.Error(err, "unable to list workspace pods")
+		return nil, err
+	}
+	switch len(workspacePods.Items) {
+	case 0:
+		// no PVC, that's ok
```
jenting, line 867: // no Pod, that's ok

components/ws-manager-mk2/controllers/workspace_controller.go
```diff
@@ -0,0 +1,876 @@
...
+		return []string{owner.Name}
+	}
+	err := mgr.GetFieldIndexer().IndexField(context.Background(), &corev1.Pod{}, wsOwnerKey, idx)
+	if err != nil {
+		return err
+	}
+	err = mgr.GetFieldIndexer().IndexField(context.Background(), &corev1.PersistentVolumeClaim{}, wsOwnerKey, idx)
+	if err != nil {
+		return err
+	}
```
jenting, line 733: I think we could remove this part, don't we?

csweichel: What is broken/unnecessary with this code?

easyCZ:
Moving to draft as it shows up in Review Requests but isn't actionable at this point in time. Feel free to bring back into review if you're actively seeking a review on this.

install/installer/pkg/config/v1/experimental/experimental.go
```diff
@@ -128,6 +128,7 @@ type WorkspaceConfig struct {
 	} `json:"contentService"`
 
 	EnableProtectedSecrets *bool `json:"enableProtectedSecrets"`
+	UseWsmanagerMk2        bool  `json:"useWsmanagerMk2,omitempty"`
```
mrsimonemms, line 131: I would strongly advocate not putting into the `experimental` section as I'm currently in the process of deprecating as much as possible from here.

The only exception would be if this is a truly experimental feature that is only going to be enabled in a limited fashion **AND** a feature that we're unlikely to promote to LTS - for example, if the intention is to use ws-manager2 once we've gotten through testing. If the plan is to support both ws-manager 1 and 2 concurrently, this should be in the main config.

mrsimonemms (changes requested):

install/installer/pkg/config/v1/experimental/experimental.go
Furisto: @mrsimonemms This property will be removed once we have rolled out mk2 to production. Mk1 will be deleted from the repository and mk2 will be the only relevant workspace manager going forward. See [milestone plan](https://www.notion.so/gitpod/Milestones-ws-manager-mk2-fb471d7e01e4494b9bca9ad1cd50b3b2).

mrsimonemms: Perfecto!

meysholdt:
I do have a question (unrelated to my approval), though :) How are the Kubernetes resources in [components/ws-manager-mk2/config](https://github.com/gitpod-io/gitpod/tree/7faac785a4471bd60fa54b8fd7dd386550a2594a/components/ws-manager-mk2/config) being used? I'd like to understand better how they relate to the [installer](https://github.com/gitpod-io/gitpod/tree/main/install/installer) and if they are being used in dev-loops.

csweichel:
> I do have a question (unrelated to my approval), though :) How are the Kubernetes resources in [components/ws-manager-mk2/config](https://github.com/gitpod-io/gitpod/tree/7faac785a4471bd60fa54b8fd7dd386550a2594a/components/ws-manager-mk2/config) being used? I'd like to understand better how they relate to the [installer](https://github.com/gitpod-io/gitpod/tree/main/install/installer) and if they are being used in dev-loops.

Those are kubebuilder generated resources which eventually result in YAML that gets injected into the installer, primarily the CRD part of it. I opted not to delete the unused bits to remain close to a "standard kubebuilder project".

mrsimonemms (approved):

akosyakov (approved):
I did not review the entire code, check out changes to ws-proxy and tested that different connectivity methods still work, i.e. VS Code Browser, IntelliJ IDEA and SSH.

components/ws-daemon/pkg/controller/controller.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	log.WithValues("workspaceID", workspace.Name, "runtime", workspace.Status.Runtime).Info("seen workspace we have to manage")
+
+	if workspace.Status.Phase == workspacev1.WorkspacePhaseCreating ||
+		workspace.Status.Phase == workspacev1.WorkspacePhaseInitializing ||
+		workspace.Status.Phase == workspacev1.WorkspacePhaseRunning {
+
+		if c := wsk8s.GetCondition(workspace.Status.Conditions, string(workspacev1.WorkspaceConditionContentReady)); c == nil {
+			failure, err := wsc.initWorkspaceContent(ctx, &workspace)
```
jenting, line 105: Could the `initWorkspaceContent` function reentrant?

I ask because is there a case that the workspace is running, but the `workspace.Status.Conditions != WorkspaceConditionContentReady` because the ws-manager overwrites the `workspace.Status.Conditions`? So, the workspace content will be reinitialized?

aledbf (approved):

easyCZ (approved):
Approving largely based on us having a feature flag and a gradual way to roll this out. The changeset is way too big to actually provide a meaningful architectural,or code, review.

More details are in https://gitpod.slack.com/archives/C02EN94AEPL/p1673947226543299

components/ws-manager-api/go/BUILD.yaml
```diff
@@ -10,4 +10,4 @@ packages:
       - "go.sum"
     config:
       packaging: library
-      dontTest: false
+      dontTest: true
```
kylos101, line 13: Non-blocking, question

@Furisto why disable? Just curious.

Furisto: Ah was not aware of that but it makes sense. Currently the test we have (for the webhook) fails. Webhooks will be removed with the next PR.

aledbf: > Webhooks will be removed with the next PR which means the test will be removed as well.

Do we have a written reason why is this being done? What's the replacement?

Furisto: This was a suggestion from @csweichel. Currently there is just nothing that we would want to validate.

aledbf: Let's talk about that in the next sync. One of the goals of the webhook is to avoid persistence with invalid values and wasting sync loops in the controller.

kylos101: @Furisto was the thought to use CEL, instead of webhooks to do validation? [This article](https://blog.rewanthtammana.com/kubernetes-crd-validation-with-cel-and-kubebuilder-marker-comments) suggests we cannot use CEL until Kubernetes 1.25 (we're currently on 1.23, and cannot move to 1.24 yet).

PR #9605, opened 2022-04-28
[public-api] Fix codegen

easyCZ:
## Description

With the move to Go 1.18, we can no longer do `go get` outside of a module file. Our dependencies are also not runnable tools so can't do `go install`. See [release notes](https://go.dev/doc/go1.18#go-command).

* `tools.go` is introduced to reference the dependencies and enable `go mod tidy` to work.

## Related Issue(s)

Fixes public-api codegen.

## Release Notes

```release-note
Fix public-api codegeneration.
```

## Documentation

NONE


easyCZ:
Some generates files are also changed. This is due to the upgraded dependencies in previous PRs. Unfortunately, we don't have a codegen-drift-check on CI so this slipped through.

csweichel (approved):
We should extend this to the other API packages

PR #9608, opened 2022-04-28
[server] reduce project inactivity time to 1 week (was 10)

svenefftinge:
This PR marks projects as inactive when there has been no workspace for more than a week. As a result, Gitpod will stop preparing workspaces through prebuilds.

I believe we can and should be more aggressive here, to not waste money (and energy). If there are pushes but no workspaces for a repo we can consider it inactive.

```release-notes
Projects are considered inactive after one week without a workspace. After that, incoming push events will no longer trigger prebuilds until a regular workspace has been started again.
```

### DIFF
--- a/components/server/ee/src/prebuilds/prebuild-manager.ts
+++ b/components/server/ee/src/prebuilds/prebuild-manager.ts
@@ -353,7 +353,7 @@ export class PrebuildManager {
         }
         const now = Date.now();
         const lastUse = new Date(usage.lastWorkspaceStart).getTime();
-        const inactiveProjectTime = 1000 * 60 * 60 * 24 * 7 * 10; // 10 weeks
+        const inactiveProjectTime = 1000 * 60 * 60 * 24 * 7 * 1; // 1 week
         return now - lastUse > inactiveProjectTime;
     }
 }


geropl:
@jankeromnes Could you PTAL? :pray:

components/server/ee/src/prebuilds/prebuild-manager.ts
```diff
@@ -353,7 +353,7 @@ export class PrebuildManager {
         }
         const now = Date.now();
         const lastUse = new Date(usage.lastWorkspaceStart).getTime();
-        const inactiveProjectTime = 1000 * 60 * 60 * 24 * 7 * 10; // 10 weeks
+        const inactiveProjectTime = 1000 * 60 * 60 * 24 * 7 * 1; // 1 week
```
jankeromnes, line 356: Nit: `* 1` is unnecessary.

```suggestion
        const inactiveProjectTime = 1000 * 60 * 60 * 24 * 7; // 1 week
```

jankeromnes (approved):
Many thanks for this Pull Request! FYI, 10 weeks was intentionally very lax, in order to test this new behavior on a smaller scale, before making it more aggressive.

The 10 weeks change was effectively enabled in production this Monday (3 days ago), and we should see the first results/saving in our internal Prebuilds dashboard starting next week.

We can already see in the DB that 4384 prebuilds got auto-cancelled since `2022-04-25T08:34`, or about 1 prebuild every minute that we decide not to run.

I agree with making this more aggressive, and now seems to be a good time.

Note: 1 week seems quite aggressive, and will likely cause some "false positives", for example most Projects might become "inactive" over the summer or winter holidays, or smaller Projects if the team goes on holidays, or becomes somewhat inactive, etc. However, the impact is quite minimal, and a brief visit to the Prebuild details (or a simple workspace start) are enough to resolve this. So, 2 weeks would seem safer, but 1 week should be fine.

PR #9614, opened 2022-04-28
Remove unneeded privileges for server

meysholdt:
## Description
A security scan highlighted that `verbs` `"delete","deletecollection"` should only be granted via k8s Roles when necessary. See the linked issue for details. 

Our `server` component has these verbs and I assume it does not need them because I assume it's `ws-manger` that creates and deletes k8s services and pods. However, maybe I'm overlooking something. If I do, please close this PR.

## Related Issue(s)
Fixes # https://github.com/gitpod-io/security/issues/39

## Release Notes

```release-note
Improved security by removing unneeded privileges from the server component.
```

### DIFF
--- a/install/installer/pkg/components/server/role.go
+++ b/install/installer/pkg/components/server/role.go
@@ -31,8 +31,6 @@ func role(ctx *common.RenderContext) ([]runtime.Object, error) {
 					"update",
 					"patch",
 					"watch",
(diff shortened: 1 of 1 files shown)


geropl (approved):
Absolutely correct. We could even think about removing all permissions. But let's do that as follow-up 👍

PR #9281, opened 2022-04-13
mess with tracing

AlexTugarev:
Trying to get rid of the millions of error logged in prod. 

- [x] /werft with-observability=true
- [x] /werft with-helm=true
- [x] /werft with-clean-slate-deployment

```release-notes
NONE
```

PR #9625, opened 2022-04-28
try to fix tracing

AlexTugarev:
follow-up to https://github.com/gitpod-io/gitpod/pull/9281

- [x] /werft with-observability=true
- [x] /werft with-helm=true

```release-notes
NONE
```

### DIFF
--- a/components/gitpod-protocol/src/util/tracing.ts
+++ b/components/gitpod-protocol/src/util/tracing.ts
@@ -28,7 +28,8 @@ export namespace TraceContext {
         if (!!parentCtx?.span) {
             const ctx = parentCtx?.span?.context();
             if (ctx && !!ctx.toTraceId() && !!ctx.toSpanId()) {
-                options.references = [opentracing.followsFrom(ctx)];
+                // options.references = [opentracing.followsFrom(ctx)];
+                options.childOf = ctx;
             }
         }
 


components/gitpod-protocol/src/util/tracing.ts
```diff
@@ -28,7 +28,8 @@ export namespace TraceContext {
         if (!!parentCtx?.span) {
             const ctx = parentCtx?.span?.context();
             if (ctx && !!ctx.toTraceId() && !!ctx.toSpanId()) {
-                options.references = [opentracing.followsFrom(ctx)];
+                // options.references = [opentracing.followsFrom(ctx)];
+                options.childOf = ctx;
```
AlexTugarev, line 32: @geropl, wdyt, should we remove these entirely? 
frankly, I don't understand the issue at hand, but given I started https://github.com/gitpod-io/gitpod/pull/9281, I'd like to find an end to this story. 
what option do we have? if we revert, we end up with messed up logs.

AlexTugarev: @geropl, is this still an issue?

geropl: No, I think we should leave it as-is. The problems we faced were mostly solved by Mad's attempts to find dangling spans that were not `finish()`ed. :+1:

AlexTugarev:
Closing, because I'm out of ideas on this.

PR #9638, opened 2022-04-29
[server][github] list repos with admin permissions only

AlexTugarev:
Avoid timeout on fetching *all* of user's repos, instead limit to those with *admin* permissions only.

```release-notes
NONE
```

### DIFF
--- a/components/server/ee/src/prebuilds/github-service.ts
+++ b/components/server/ee/src/prebuilds/github-service.ts
@@ -29,21 +29,19 @@ export class GitHubService extends RepositoryService {
         const octokit = await this.githubApi.create(user);
         const repositories = await octokit.paginate(
             octokit.repos.listForAuthenticatedUser,
-            { per_page: 100 },
+            { per_page: 100, type: "owner" },
             (response) =>
-                response.data
-                    .filter((r) => !!r.permissions?.admin)
-                    .map((r) => {
-                        return <ProviderRepository>{
-                            name: r.name,
-                            cloneUrl: r.clone_url,
-                            account: r.owner.login,
-                            accountAvatarUrl: r.owner.gravatar_id
-                                ? `https://www.gravatar.com/avatar/${r.owner.gravatar_id}?size=128`
-                                : r.owner.avatar_url,
-                            updatedAt: r.updated_at,
-                        };
-                    }),
+                response.data.map((r) => {
+                    return <ProviderRepository>{
+                        name: r.name,
+                        cloneUrl: r.clone_url,
+                        account: r.owner.login,
+                        accountAvatarUrl: r.owner.gravatar_id
+                            ? `https://www.gravatar.com/avatar/${r.owner.gravatar_id}?size=128`
+                            : r.owner.avatar_url,
+                        updatedAt: r.updated_at,
+                    };
+                }),
         );
         return repositories;
     }


easyCZ:
Do you know if our providers do offer a search API? If they did, we could consider only searching for the first 100 (based on the input string) and rely on the search query getting more specific to actually find what you're looking for.

components/server/ee/src/prebuilds/github-service.ts
```diff
@@ -29,21 +29,19 @@ export class GitHubService extends RepositoryService {
         const octokit = await this.githubApi.create(user);
         const repositories = await octokit.paginate(
             octokit.repos.listForAuthenticatedUser,
-            { per_page: 100 },
+            { per_page: 100, admin: true },
```
jankeromnes, line 32: Looks useful! However, I've never seen this parameter documented anywhere:

- Not in Octokit: https://octokit.github.io/rest.js/v18#repos-list-for-authenticated-user
- Not in GitHub API: https://docs.github.com/en/rest/repos/repos#list-repositories-for-the-authenticated-user

Intrigued to see whether it works. 👀

jankeromnes:
> Do you know if our providers do offer a search API? If they did, we could consider only searching for the first 100 (based on the input string) and rely on the search query getting more specific to actually find what you're looking for.

That's a good idea, which would allow us to not have to fetch all repositories (problematic for people who are admin of 3K+ repositories).

But another solution which might be simpler could be to simply allow users to paste a repository URL while creating a project (instead of painstakingly fetching all possible repos, then letting users browse through a slightly awkward UI to hunt for the one thing they want -- we don't need to build a new search feature, there are already good ones out there e.g. GitHub, Google Search...).

components/server/ee/src/prebuilds/github-service.ts
AlexTugarev: 🤦🏻‍♂️  good point! my bad, the param's name is different. `type=owner` it is. 

The very request can be scoped ...

Screen Shot 2022-04-29 at 10 16 00

And the docs on the `type` param ... 

Screen Shot 2022-04-29 at 10 39 55

The `owner` type translates to `admin: true` on GitHub. 

Feel free to verify yourself.

AlexTugarev:
>  Do you know if our providers do offer a search API? If they did, we could consider only searching for the first 100 (based on the input string) and rely on the search query getting more specific to actually find what you're looking for.

@easyCZ, yes there is a search API for repositories: https://docs.github.com/en/rest/search#search-repositories

You can verify yourself, that this does not provide search criteria to filter for permission type.

For the problem at hand we want to find *all* repos which the current user has admin permissions for. Because interaction with 3rd party APIs is slow in general, we aim for two goals:
1. push the filtering part to the remote service, instead of getting *ALL* and filtering in our service
2. making use of pagination to fetch the data async, and ideally do the paginated requests on our API as well as to the remote service (which we need to do properly and broadly available in the public API 🙏🏻 )

This PR is about 1.

components/server/ee/src/prebuilds/github-service.ts
jankeromnes: Thanks @AlexTugarev!

> The `owner` type translates to `admin: true` on GitHub.

Well, I initially did not use this, because I wasn't able to find the information that `owner` type (here) === `admin: true` for every repository.

Where can I find more documentation on the `type` parameter here?

jankeromnes:
Thanks for the update @AlexTugarev!

Before your latest change, I got many repositories, including ones where I'm not an admin:

<table><tr><td>
Screenshot 2022-04-29 at 09 01 23
</td><td>
Screenshot 2022-04-29 at 09 01 39
</td><td>
Screenshot 2022-04-29 at 09 02 02
</td><td>
Screenshot 2022-04-29 at 09 02 43
</td></tr></table>

However, now, with `type = owner`, I seem to only get repositories where I'm indeed the owner (i.e. my personal repos), but I no longer see any repositories from other orgs where I'm admin:

Screenshot 2022-04-29 at 11 16 53

Does the "New Project" view work as expected for you? 🤔

AlexTugarev:
> However, now, with type = owner, I seem to only get repositories where I'm indeed the owner (i.e. my personal repos), but I no longer see any repositories from other orgs where I'm admin:

You are right. Thanks for checking! 🙏🏻 

This was a long shot, thus closing it. 🛑 

I think the way to proceed would be to write a graphql query which would iterate over all orgs. This way we could likely catch most cases in a single request. Quickly tested this in the graphql explorer on github.com:
```
{
  viewer {
    organizations(first: 100) {
      nodes {
        repositories(first: 100, affiliations: OWNER) {
          totalCount
          
          nodes {
            nameWithOwner
            viewerPermission
          }
        }
        viewerCanAdminister
      }
    }
  }
}
```

PR #9269, opened 2022-04-12
Pre-install Prettier and ESLint VSCode extensions

trumbitta:
## Description

In an effort to further simplify #8995, this only adds pre-installed and pre-activated ESLint & Prettier extensions for VSCode with "format on save".

It's a cherry-pick from #8895 so it's expected to vanish from that PR once this gets merged.

## Related Issue(s)

Re #3841
This is a prerequisite to ensure the work on the linting issues can proceed with sane defaults. 

## How to test

For Prettier: open a TS file, and see if the code you write gets formatted on save.
For ESLint: open a TS file, produce a linting error, and see if VSCode marks it with a squiggly yellow line.

## Release Notes

```release-note
- [devx]: Added ESLint and Prettier to the auto-installed VSCode extensions
- [devx]: Activated "format on save" for TypeScript and JavaScript
```

PR #5112, opened 2021-08-09
Update projects empty state

gtsiolis:
### What does this PR do?

1. Fix the help link (_Learn more_) on the Projects empty state.
2. Refactor help links to use a single class named `.gp-link` as potentially a good name for a link component.

PR #9641, opened 2022-04-29
[dashboard] add button.gp-link CSS class

trumbitta:
## Description

Adds a `.gp-link` Tailwind class to make buttons mimic links and avoid using `<a href="javascript:void(0)" ...>`

To be used for anchor tags that should've been buttons because they don't really bring the user to any new location like an anchor tag usually does.

Just like #9269, this is about #3841 and its purpose is to simplify to the max #8995.

## Related Issue(s)

Re #3841

## How to test

- launch the dashboard app
- the two updated links which are now buttons should look and work just like before

## Release Notes

```release-note
- [dashboard]: Added a `.gp-link` Tailwind class to make buttons mimic anchor tags
```

### DIFF
--- a/components/dashboard/src/index.css
+++ b/components/dashboard/src/index.css
@@ -9,7 +9,8 @@
 @tailwind utilities;
 
 @layer base {
-    html, body {
+    html,
+    body {
         @apply h-full;
     }
     body {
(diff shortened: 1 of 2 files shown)


laushinka:
Thanks, @trumbitta! Let me loop in @gtsiolis here for this one :)

gtsiolis:
Looking at this now! 👀 

/werft run

gtsiolis:
@trumbitta could you rebase this to `main` branch to fix the failing build?

components/dashboard/src/index.css
```diff
@@ -61,8 +61,13 @@
         @apply cursor-default opacity-50 pointer-events-none;
     }
 
-    a.gp-link {
-        @apply text-blue-500 hover:text-blue-600  dark:text-blue-400 dark:hover:text-blue-500;
+    button.gp-link {
+        @apply bg-transparent hover: bg-transparent p-0 rounded-none;
```
trumbitta, line 65: wait a minute... where does this extra white-space between `hover:` and `bg-transparent` come from 🤔 

fixing it now (also in the next selector)

trumbitta: @gtsiolis should be fixed now. Sorry for the hiccup, I don't really know how it happened 😮

trumbitta:
Is werft having a bad day?

components/dashboard/src/projects/NewProject.tsx
```diff
@@ -493,13 +493,12 @@ export default function NewProject() {
                     <div>
                         <div className="text-gray-500 text-center w-96 mx-8">
                             Repository not found?{" "}
-                            <a
-                                href="javascript:void(0)"
```
gtsiolis, line 497: **praise**: Nice!

gtsiolis, line 497: **suggestion(non-blocking)**: We have a few more actions in **/plans** masked as links that could benefit from this. However, I wouldn't block this from merging it as we're going to update and replace plans entirly soon.

https://github.com/gitpod-io/gitpod/blob/08954965c673d38e5aa042de9146bf5406929dd8/components/dashboard/src/settings/Plans.tsx#L390-L396

https://github.com/gitpod-io/gitpod/blob/08954965c673d38e5aa042de9146bf5406929dd8/components/dashboard/src/settings/Plans.tsx#L452-L458

...

components/dashboard/src/projects/NewProject.tsx
```diff
@@ -368,9 +368,9 @@ export default function NewProject() {
                 <p className="text-gray-500 text-center text-base mt-12">
                     {loaded && noReposAvailable ? "Select account on " : "Select a Git repository on "}
                     <b>{selectedProviderHost}</b> (
-                    <a className="gp-link cursor-pointer" onClick={() => setShowGitProviders(true)}>
+                    <button className="gp-link cursor-pointer" onClick={() => setShowGitProviders(true)}>
                         change
```
gtsiolis, line 372: **issue(non-blocking)**: This will slightly affect how the link style is rendered but that's minor for now.

| BEFORE | AFTER |
|-|-|
| link-before | link-after |

gtsiolis (approved):
Thanks for opening this PR, your patience, and caring about accessibility with replacing these links with buttons, @trumbitta! 💯 

To add some background context here, the introduction of the `gp-link` class in https://github.com/gitpod-io/gitpod/pull/5112 initially aimed to abstract the style used for help links. 🔗 

However, the long-term goal is to have a button component _variant_ styled as a link for _actions_ like the ones changed in this PR, instead of maintaining CSS styles in the `index.css`. 💡 

Ideally, we'd like to introduce a button component for what we need here but I don't want to block or close this pull request since we're going to update and replace the pages affected here soon. 🚀 

UX LGTM!

@laushinka could you take another look and merge this since this requires also an approval from the WebApp team?

components/dashboard/src/projects/NewProject.tsx
trumbitta: Good catch 🏅 !

I guess it stems from [`text-sm` in the Tailwind styles applied to the `button` element](https://github.com/gitpod-io/gitpod/blob/7b0acb66f47d47ee5533a8a6c3cba545a677272d/components/dashboard/src/index.css#L50) while `a` seems to just inherit as usual.

The upcoming button component should take it into account. If you need a volunteer for working on the component, you know where to find me 😊

trumbitta:
> @trumbitta I need to stop apologizing to you and just review faster! This fell through the cracks of my reviewing cycle 🙇🏻‍♀️
> 
> LGTM

🙏  @laushinka I'm yet to find anything you should apologize for. If anything, I owe you because sometimes I just find bits of your code in the dashboard codebase that are so good I want to scream.

PR #9651, opened 2022-04-29
[server] Remove logLevel from server config

andrew-farries:
## Description

Remove `logLevel` from the `server` config and take it from the environment instead. `server` deployments in Gitpod releases installed with the installer have the `LOG_LEVEL` env var set here:

https://github.com/gitpod-io/gitpod/blob/1f3919ea76d70e11a2989b7346bd3c885f6580ea/install/installer/pkg/components/server/deployment.go#L74

and applied here:

https://github.com/gitpod-io/gitpod/blob/1f3919ea76d70e11a2989b7346bd3c885f6580ea/install/installer/pkg/components/server/deployment.go#L279

## Related Issue(s)
Part of #9097 

## Release Notes

```release-note
NONE
```

## Documentation

none


geropl (approved):
Nice! :footprints:

Issue #9652, opened 2022-04-29
Extend the DB schema and entities to support team subscriptions in teams


jankeromnes:
Part of https://github.com/gitpod-io/gitpod/issues/7759 (public) / [RFC](https://www.notion.so/gitpod/Aligning-Team-Plans-and-Teams-9073db94359549918ab5c85cc364de46) (internal)

PR #8041, opened 2022-02-04
Implement a new Team Billing where Owners can conveniently manage a paid plan for their Team

jankeromnes:
## Description

Allow teams to easily purchase and manage a Team Plan directly from their team's Settings.

TODO:
- [x] Implement a new Team Plan shape that's directly linked to / owned by a team: https://github.com/gitpod-io/gitpod/pull/9655
- [x] Allow subscribing via Chargebee directly from your team Settings
- [x] Create a "synthetic" user subscription when you're a member of a paid Team Plan
- [x] Transparently update the Team Plan quantity in Chargebee when users join or leave a team
- [x] Pro-rate charge for new members joining in the middle of a billing cycle
- [x] Allow managing your Billing in team Settings via a "Team Chargebee Portal"
- [x] Align the new UI with #8352 

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/7759

## How to test

1. Create a Team
2. Invite other team members
3. Go to Team Settings > Plans
3. Buy a Team Plan, verify billing amount matches member count, verify all members are indeed upgraded
4. Add/remove members, verify that the Team Plan billing amount is updated accordingly
5. Verify that cancelling, re-upgrading, deleting the team all work as expected

## Release Notes

```release-note
Implement a new Team Billing where Team Owners can conveniently manage a paid plan for their Team
```

## Documentation

- [ ] /werft without-vm=true
- [ ] /werft with-helm=true
- [x] /werft with-payment=true
- [x] /werft with-clean-slate-deployment

Issue #7759, opened 2022-01-21
Epic: Simplify team plans and align with project teams


jldec:
[Updated 2022-01-28 by @JanKoehnlein]
[Updated 2022-02-01 by @JanKoehnlein]
[Updated 2022-05-13 by @jldec]

**Summary**  
The goal of this epic is to simplify the administration of [**team plans**](https://www.gitpod.io/docs/teams) by making team plans behave like the new [**teams**](https://www.gitpod.io/docs/teams-and-projects#create-a-team) introduced with teams and projects. There's a detailed discussion in this [RFC](https://www.notion.so/gitpod/Aligning-Team-Plans-and-Teams-9073db94359549918ab5c85cc364de46?d=cfbea403a6734db8a3049add816c6295#e24d0882bd174016a1499a424c73e662) (internal)

**Challenges**  
- There are now 2 kinds of teams in Gitpod, one for team plan subscriptions, one for projects. This is confusing for users.
- Managing the subscription slots (members) in a Team Plan can be annoying.
- Team plans are owned by a single user. Administration of team plans cannot be delegated.
- Team plan owners do not need to be a member of the team plan themselves. ~~This capability needs to be preserved.~~
- Team plan members cannot remove themselves from a team.
- There is no API for managing team plan members.

**Task breakdown**  
- [x] #9652
- [x] #8041
  - Allow subscribing via Chargebee directly from your team Settings
  - Create a "synthetic" user subscription when you're a member of a paid Team Plan
  - Transparently update the Team Plan quantity in Chargebee when users join or leave a team
  - Pro-rate charge for new members joining in the middle of a billing cycle
  - Allow managing your Billing in team Settings via a "Team Chargebee Portal"
  - Align the new UI with #8352
  - Ship without new 'Billing' menu in team settings
- [ ] ~Only allow team owners to invite new users to a team with a paid plan~
- [ ] ~Add a membership flag for non-billed members~
- [x] Add new 'Billing' menu into team settings
- [ ] ~https://github.com/gitpod-io/ops/issues/2543~
- [x] https://github.com/gitpod-io/website/issues/2216
- [x] #10741

**Related work**
[Usage-based Billing](https://github.com/gitpod-io/gitpod/issues/9036) will provide an alternative billing mechanism which is independent of user-count or plan-selection.

jankeromnes:
> - Team plan owners do not need to be a member of the team plan themselves. This capability needs to be preserved.

[...]

PR #9655, opened 2022-04-29
[db][protocol] Implement TeamSubscription2 DB shapes and migration

jankeromnes:
## Description

Implement TeamSubscription2 DB shapes and migration.

Part of https://github.com/gitpod-io/gitpod/issues/7759, split out of https://github.com/gitpod-io/gitpod/pull/8041

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/9652

## How to test

1. Should build
2. DB should contain new table `d_b_team_subscription2` + an additional column in both `d_b_subscription` and `d_b_team_membership` tables

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/gitpod-db/src/typeorm/migration/1650526577994-TeamSubscrition2.ts
+++ b/components/gitpod-db/src/typeorm/migration/1650526577994-TeamSubscrition2.ts
@@ -1,0 +1,38 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { MigrationInterface, QueryRunner } from "typeorm";
+import { columnExists, tableExists } from "./helper/helper";
+
+export class TeamSubscrition21650526577994 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(
+            "CREATE TABLE IF NOT EXISTS `d_b_team_subscription2` (`id` char(36) NOT NULL, `teamId` char(36) NOT NULL, `paymentReference` varchar(255) NOT NULL, `startDate` varchar(255) NOT NULL, `endDate` varchar(255) NOT NULL DEFAULT '', `planId` varchar(255) NOT NULL, `quantity` int(11) NOT NULL, `cancellationDate` varchar(255) NOT NULL DEFAULT '', `deleted` tinyint(4) NOT NULL DEFAULT '0', `_lastModified` timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6), PRIMARY KEY (`id`), KEY `ind_team_paymentReference` (`teamId`, `paymentReference`), KEY `ind_team_startDate` (`teamId`, `startDate`), KEY `ind_dbsync` (`_lastModified`)) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;",
+        );
+        if (!(await columnExists(queryRunner, "d_b_subscription", "teamMembershipId"))) {
+            await queryRunner.query(
+                "ALTER TABLE `d_b_subscription` ADD COLUMN `teamMembershipId` char(36) NOT NULL DEFAULT ''",
+            );
+        }
+        if (!(await columnExists(queryRunner, "d_b_team_membership", "subscriptionId"))) {
+            await queryRunner.query(
+                "ALTER TABLE `d_b_team_membership` ADD COLUMN `subscriptionId` char(36) NOT NULL DEFAULT ''",
+            );
+        }
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        if (await tableExists(queryRunner, "d_b_team_subscription2")) {
+            await queryRunner.query("DROP TABLE `d_b_team_subscription2`");
+        }
+        if (await columnExists(queryRunner, "d_b_subscription", "teamMembershipId")) {
+            await queryRunner.query("ALTER TABLE `d_b_subscription` DROP COLUMN `teamMembershipId`");
+        }
+        if (await columnExists(queryRunner, "d_b_team_membership", "subscriptionId")) {
+            await queryRunner.query("ALTER TABLE `d_b_team_membership` DROP COLUMN `subscriptionId`");
+        }
+    }
+}
--- a/components/gitpod-db/src/typeorm/team-db-impl.ts
+++ b/components/gitpod-db/src/typeorm/team-db-impl.ts
@@ -83,6 +83,11 @@ export class TeamDBImpl implements TeamDB {
         return infos.sort((a, b) => (a.memberSince < b.memberSince ? 1 : a.memberSince === b.memberSince ? 0 : -1));
     }
 
+    public async findTeamMembership(userId: string, teamId: string): Promise<DBTeamMembership | undefined> {
+        const membershipRepo = await this.getMembershipRepo();
+        return membershipRepo.findOne({ userId, teamId, deleted: false });
+    }
+
     public async findTeamsByUser(userId: string): Promise<Team[]> {
         const teamRepo = await this.getTeamRepo();
         const membershipRepo = await this.getMembershipRepo();
@@ -189,6 +194,21 @@ export class TeamDBImpl implements TeamDB {
             throw new Error("The user is not currently a member of this team");
         }
         membership.role = role;
+        await membershipRepo.save(membership);
+    }
+
+    public async setTeamMemberSubscription(userId: string, teamId: string, subscriptionId: string): Promise<void> {
+        const teamRepo = await this.getTeamRepo();
+        const team = await teamRepo.findOne(teamId);
+        if (!team || !!team.deleted) {
+            throw new Error("A team with this ID could not be found");
+        }
+        const membershipRepo = await this.getMembershipRepo();
+        const membership = await membershipRepo.findOne({ teamId, userId, deleted: false });
+        if (!membership) {
+            throw new Error("The user is not currently a member of this team");
+        }
+        membership.subscriptionId = subscriptionId;
         await membershipRepo.save(membership);
     }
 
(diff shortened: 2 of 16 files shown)


AlexTugarev:
Looking into this now, and reading up the plan.

jankeromnes:
Many thanks @AlexTugarev! 😊 Let me know if you have any questions 🙏

jankeromnes:
Friendly ping @AlexTugarev 😇 🧡

AlexTugarev (commented):
`DBTeamMembership.subscriptionId` and `DBSubscription.teamMembershipId` seems to be redundant.

Looking at the usage in https://github.com/gitpod-io/gitpod/pull/8041/commits/81b41f18b0f00ce58bc8af4ed9c5ca76b5020c54#diff-349769928506cc98ef99e13b1d146ae991fd2755d6076cda2ed87b1693eec071R36-R37

What's the benefit?

jankeromnes:
Thanks for the feedback @AlexTugarev!

> `DBTeamMembership.subscriptionId` and `DBSubscription.teamMembershipId` seems to be redundant.
> [...]
> What's the benefit?

This is modelled on `DBTeamSubscriptionSlot.subscriptionId` and `DBSubscription.teamSubscriptionSlotId` being also redundant (or rather, they materialize a relationship in both directions), which has been helpful on several occasions (e.g. for bugs, or auditing purposes, when you lose one side of the link but not the other).

For example, when you leave a team, the membership is automatically deleted (leaving no trace), but the subscription is merely cancelled at end of term. This leaves a trace (e.g. for users wondering why they have unlimited hours without a paid subscription, we could see that the still-active subscription came from a team membership that has since been deleted).

AlexTugarev:
@jankeromnes, I find the cut of the PRs less useful for a review. I understand #8041 is a 🐋, but it seems cutting out the DB shapes doesn't much, as one still need to understand the call sites. 

How else do you expect this to be reviewed?

components/gitpod-db/src/typeorm/migration/1650526577994-TeamSubscrition2.ts
```diff
@@ -0,0 +1,28 @@
...
+            );
+        }
+        if (!(await columnExists(queryRunner, "d_b_team_membership", "subscriptionId"))) {
+            await queryRunner.query(
+                "ALTER TABLE `d_b_team_membership` ADD COLUMN `subscriptionId` char(36) NOT NULL DEFAULT ''",
+            );
+        }
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {}
```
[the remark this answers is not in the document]
jankeromnes: Good catch! But, it should be pretty safe to omit this tear down, no? (I.e. up and down migrations can safely happen as many times and in any order as we like, the end result is still what we want.)

components/gitpod-db/src/typeorm/team-db-impl.ts
```diff
@@ -192,6 +197,21 @@ export class TeamDBImpl implements TeamDB {
...
+        const team = await teamRepo.findOne(teamId);
+        if (!team || !!team.deleted) {
+            throw new Error("A team with this ID could not be found");
+        }
+        const membershipRepo = await this.getMembershipRepo();
+        const membership = await membershipRepo.findOne({ teamId, userId, deleted: false });
+        if (!membership) {
+            throw new Error("The user is not currently a member of this team");
+        }
+        membership.subscriptionId = subscriptionId;
```
AlexTugarev, line 211: from #9655 I can read that the subscription already knows the membership. why storing this link in both directions?

components/gitpod-protocol/src/accounting-protocol.ts
```diff
@@ -158,6 +159,15 @@ export namespace AssignedTeamSubscription {
     }
 }
 
+export interface AssignedTeamSubscription2 extends Subscription {
+    teamMembershipId: string;
+}
+export namespace AssignedTeamSubscription2 {
+    export function is(data: any): data is AssignedTeamSubscription2 {
+        return !!data && data.hasOwnProperty("teamMembershipId");
```
AlexTugarev, line 167: instead of `!data`, please use `typeof data === "object"` instead.

components/gitpod-db/src/typeorm/team-db-impl.ts
jankeromnes: See https://github.com/gitpod-io/gitpod/pull/9655#issuecomment-1120769840

components/gitpod-db/src/typeorm/migration/1650526577994-TeamSubscrition2.ts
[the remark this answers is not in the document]
AlexTugarev: In general, no. In case we'd need to revert, a following changed application is more complicated.

components/gitpod-db/src/typeorm/team-db-impl.ts
AlexTugarev: (Let's use this as a thread in here.)

> For example, when you leave a team, the membership is automatically deleted (leaving no trace), but the subscription is merely cancelled at end of term. This leaves a trace (e.g. for users wondering why they have unlimited hours without a paid subscription, we could see that the still-active subscription came from a team membership that has since been deleted).

So, what is the `DBTeamMembership.subscriptionId` need for then? I couldn't spot any usage.

jankeromnes: Sure! So, while this direction of the link is indeed not used in the current code, it was requested in the RFC that this link be materialized in both directions:

> - We add a 0..1 reference from `d_b_team_member[ship]`  to `d_b_subscription` and the opposite reference in analogy to the reference to `d_b_team_subscription_slot`. 

[Source](https://www.notion.so/gitpod/Aligning-Team-Plans-and-Teams-9073db94359549918ab5c85cc364de46#e207bc6c7bf24b578fae8f7b9bd9d5d8) (internal)

And, this made sense to me at the time, because the link between `d_b_team_subscription_slot` and `d_b_subscription` also went both ways.

As mentioned above, this slight duplication has helped/saved us a few times, for example in customer requests, or to investigate recover from bugs.

Are you proposing to drop the reference from `d_b_team_membership` to `d_b_subscription` now?

components/gitpod-db/src/typeorm/migration/1650526577994-TeamSubscrition2.ts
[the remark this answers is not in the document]
jankeromnes: I'd prefer to leave this as is for now, but if we do need to permanently revert, I volunteer myself to drop the extra tables/columns manually.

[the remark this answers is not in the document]
jankeromnes: Hmm, well okay, I might as well implement the tear down while we're here. 👍

jankeromnes:
Addressed all nits, DB still looks good!

<details><summary>(see tables)</summary>

```
mysql> describe d_b_team_subscription2;
+------------------+--------------+------+-----+----------------------+--------------------------------+
| Field            | Type         | Null | Key | Default              | Extra                          |
+------------------+--------------+------+-----+----------------------+--------------------------------+
| id               | char(36)     | NO   | PRI | NULL                 |                                |
| teamId           | char(36)     | NO   | MUL | NULL                 |                                |
| paymentReference | varchar(255) | NO   |     | NULL                 |                                |
| startDate        | varchar(255) | NO   |     | NULL                 |                                |
| endDate          | varchar(255) | NO   |     |                      |                                |
| planId           | varchar(255) | NO   |     | NULL                 |                                |
| quantity         | int(11)      | NO   |     | NULL                 |                                |
| cancellationDate | varchar(255) | NO   |     |                      |                                |
| deleted          | tinyint(4)   | NO   |     | 0                    |                                |
| _lastModified    | timestamp(6) | NO   | MUL | CURRENT_TIMESTAMP(6) | on update CURRENT_TIMESTAMP(6) |
+------------------+--------------+------+-----+----------------------+--------------------------------+
10 rows in set (0.01 sec)

mysql> describe d_b_subscription;
+------------------------+--------------+------+-----+----------------------+--------------------------------+
| Field                  | Type         | Null | Key | Default              | Extra                          |
+------------------------+--------------+------+-----+----------------------+--------------------------------+
| id                     | int(11)      | YES  |     | NULL                 |                                |
| userId                 | char(36)     | NO   | MUL | NULL                 |                                |
| startDate              | varchar(255) | NO   |     | NULL                 |                                |
| endDate                | varchar(255) | NO   |     |                      |                                |
| amount                 | double       | NO   |     | NULL                 |                                |
| uid                    | char(36)     | NO   | PRI | NULL                 |                                |
| _lastModified          | timestamp(6) | NO   | MUL | CURRENT_TIMESTAMP(6) | on update CURRENT_TIMESTAMP(6) |
| planId                 | varchar(255) | NO   | MUL | free                 |                                |
| paymentReference       | varchar(255) | NO   |     |                      |                                |
| deleted                | tinyint(4)   | NO   |     | 0                    |                                |
| cancellationDate       | varchar(255) | NO   |     |                      |                                |
| paymentData            | text         | YES  |     | NULL                 |                                |
| teamSubscriptionSlotId | char(255)    | NO   | MUL |                      |                                |
| firstMonthAmount       | double       | YES  |     | NULL                 |                                |
| teamMembershipId       | char(36)     | NO   |     |                      |                                |
+------------------------+--------------+------+-----+----------------------+--------------------------------+
15 rows in set (0.05 sec)

mysql> describe d_b_team_membership;
+----------------+--------------+------+-----+----------------------+--------------------------------+
| Field          | Type         | Null | Key | Default              | Extra                          |
+----------------+--------------+------+-----+----------------------+--------------------------------+
| id             | char(36)     | NO   | PRI | NULL                 |                                |
| teamId         | char(36)     | NO   | MUL | NULL                 |                                |
| userId         | char(36)     | NO   | MUL | NULL                 |                                |
| role           | varchar(255) | NO   |     | NULL                 |                                |
| creationTime   | varchar(255) | NO   |     | NULL                 |                                |
| deleted        | tinyint(4)   | NO   |     | 0                    |                                |
| _lastModified  | timestamp(6) | NO   | MUL | CURRENT_TIMESTAMP(6) | on update CURRENT_TIMESTAMP(6) |
| subscriptionId | char(36)     | NO   |     |                      |                                |
+----------------+--------------+------+-----+----------------------+--------------------------------+
8 rows in set (0.02 sec)
```

</details>

AlexTugarev (approved):
LGTM!

PR #9663, opened 2022-04-29
[dashboard] increase IDE awareness for onboarding users when starting workspace

### LINKED ISSUE #9534: Allow user to configure IDE configurations when onboarding start workspace
Allow user to configure IDE configurations when onboarding start workspace.

Works for aware users Gitpod not only works with VSCode Browser. We got many IDEs.

 See also https://github.com/gitpod-io/gitpod/issues/6707, [Draft Figma](https://www.figma.com/file/1ofpgCa3XI0F4zZKhcgiGg/%236707?node-id=41%3A277)

|Draft|
|:-:|
|image|

mustard-mh:
## Description

🧪 This is a PLG experiment that aims to increase the awareness of IDE/editor variety on Gitpod, for first-time users. 

With https://github.com/gitpod-io/gitpod/pull/9432, we've introduced the IDE/editor selection as one of the very first steps for new users. However, some users start using Gitpod without going through the Dashboard - for example by using the context URL: `gitpod.io#REPO-URL`

With this PR, we introduce the same modal UI component, but in the "Create Workspace" page (aka landing page when a context URL is used). This is only impacting users who want to start their first workspace, but don't have a preferred IDE/editor.

See GIF below, to preview the new UI

## Related Issue(s)

Fixes #9534
Relates to #6707

## How to test

Using this preview env: `https://hw-9534-onboarding.staging.gitpod-dev.com#<REPO_URL>`

Open workspace via URL 
1. `[After Reset Account]` Default prefix `/#` will wait until Modal close or choose. After select, will open workspace with choosed IDE.
   - Popup Modal and wait until close or select or continue
   - Detail cases see **TestCases** section
   - i.e. https://hw-9534-onboarding.staging.gitpod-dev.com/#https://github.com/gitpod-io/template-sveltejs
2. `[After Reset Account]` Prefix with special IDE `/#referrer` will set up user's defaultIde if it's **not been set yet**. Open with `/#referrer` is behave like open <repo> via JetBrains Gateway UI with choosen IDE. 
    - No modal, because gateway specified one
    - If no `defaultIde` for user's setting, go with the one specified in URL
    - If have `defaultIde` behave like before
    - i.e. https://hw-9534-onboarding.staging.gitpod-dev.com/#referrer:jetbrains-gateway:pycharm/https://github.com/gitpod-io/template-python-flask (this one go with `pycharm`)
3. `[After Reset Account]` Prefix with special IDE `/#referrer` but args not correctly fit referrer required will use default ide `code` to start workspace
    - No modal, because hash starts with referrer
    - Workspace starts with IDE `code`
    - User ideSettings still undefined
    - Will show ide selection next time start a workspace too
    - i.e. https://hw-9534-onboarding.staging.gitpod-dev.com/#referrer:/https://github.com/gitpod-io/template-python-flask
4.  After setup defaultIde via prefix workspace starting page, onboarding aware on page `/workspaces` will not appear.
    - Detail `/workspaces` PR see https://github.com/gitpod-io/gitpod/pull/9432

### Test Cases

**Every selection from user will change User setting immediately**

1. Choose nothing, refresh page will see modal again
2. Choose nothing and click `continue` or `X`, DB go with `code latest:false`
6. Check `use latest` and click continue or `X`, DB go with `code latest:true`
7. Choose any IDE, DB will change

#### FYI: Check and reset your ide option

**Check**: go to preferences page to see current selection IDE

**Reset**: delete account on settings account page.

**OR**

```mysql
# check
SELECT id, name, additionalData->'$.ideSettings' FROM d_b_user WHERE id = '<your_user_id>';

# reset
UPDATE d_b_user SET additionalData='{"emailNotificationSettings":{"allowsChangelogMail":true,"allowsDevXMail":true,"allowsOnboardingMail":true}}' WHERE id = '<your_user_id>';
```

## Release Notes

```release-note
Prompt onboarding users to choose default IDE during workspace starting
```

## Screenshots / GIF

- [x] /werft without-vm=true

### DIFF
--- a/components/dashboard/src/App.tsx
+++ b/components/dashboard/src/App.tsx
@@ -20,7 +20,7 @@ import gitpodIcon from "./icons/gitpod.svg";
 import { ErrorCodes } from "@gitpod/gitpod-protocol/lib/messaging/error";
 import { useHistory } from "react-router-dom";
 import { trackButtonOrAnchor, trackPathChange, trackLocation } from "./Analytics";
-import { LicenseInfo, User } from "@gitpod/gitpod-protocol";
+import { ContextURL, LicenseInfo, User } from "@gitpod/gitpod-protocol";
 import * as GitpodCookie from "@gitpod/gitpod-protocol/lib/util/gitpod-cookie";
 import { Experiment } from "./experiments";
 import { workspacesPathMain } from "./workspaces/workspaces.routes";
@@ -45,6 +45,8 @@ import {
 import { refreshSearchData } from "./components/RepositoryFinder";
 import { StartWorkspaceModal } from "./workspaces/StartWorkspaceModal";
 import { parseProps } from "./start/StartWorkspace";
+import SelectIDEModal from "./settings/SelectIDEModal";
+import { StartPage, StartPhase } from "./start/StartPage";
 
 const Setup = React.lazy(() => import(/* webpackPrefetch: true */ "./Setup"));
 const Workspaces = React.lazy(() => import(/* webpackPrefetch: true */ "./workspaces/Workspaces"));
@@ -495,12 +497,24 @@ function App() {
         );
         return <div></div>;
     }
+    // Prefix with `/#referrer` will specify an IDE for workspace
+    // We don't need to show IDE preference in this case
+    const shouldUserIdePreferenceShown = User.isOnboardingUser(user) && !hash.startsWith(ContextURL.REFERRER_PREFIX);
+
     const isCreation = window.location.pathname === "/" && hash !== "";
     const isWsStart = /\/start\/?/.test(window.location.pathname) && hash !== "";
     if (isWhatsNewShown) {
         toRender = <WhatsNew onClose={() => setWhatsNewShown(false)} />;
     } else if (isCreation) {
-        toRender = <CreateWorkspace contextUrl={hash} />;
+        if (shouldUserIdePreferenceShown) {
+            toRender = (
+                <StartPage phase={StartPhase.Checking}>
+                    <SelectIDEModal />
+                </StartPage>
+            );
+        } else {
+            toRender = <CreateWorkspace contextUrl={hash} />;
+        }
     } else if (isWsStart) {
         toRender = <StartWorkspace {...parseProps(hash, window.location.search)} />;
     } else if (/^(github|gitlab)\.com\/.+?/i.test(window.location.pathname)) {
--- a/components/dashboard/src/settings/SelectIDEModal.tsx
+++ b/components/dashboard/src/settings/SelectIDEModal.tsx
@@ -11,7 +11,7 @@ import SelectIDE, { updateUserIDEInfo } from "./SelectIDE";
 import Modal from "../components/Modal";
 import { UserContext } from "../user-context";
 
-export default function () {
+export default function (props: { onClose?: () => void }) {
     const { user, setUser } = useContext(UserContext);
     const [visible, setVisible] = useState(true);
 
@@ -23,11 +23,13 @@ export default function () {
     const handleContinue = async () => {
         setVisible(false);
         if (!user || User.hasPreferredIde(user)) {
+            props.onClose && props.onClose();
             return;
         }
         // TODO: We need to get defaultIde in ideOptions..
         const defaultIde = "code";
         await actualUpdateUserIDEInfo(user, defaultIde, false);
+        props.onClose && props.onClose();
     };
 
     return (


components/server/src/workspace/context-parser.ts
```diff
@@ -96,12 +96,24 @@ export interface URLParts {
  * and expects "othercontext" to be parsed and passed back.
  */
 export interface IPrefixContextParser {
+    // readonly property to see if prefix context will specify an IDE
```
mustard-mh, line 99: Update interface `IPrefixContextParser` so that other prefix context parsers that may be added in the future will not break this PR

components/server/src/workspace/context-parser.ts
```diff
@@ -96,12 +96,24 @@ export interface URLParts {
...
+    specifiedIDE: boolean;
```
mustard-mh, line 100: Update interface `IPrefixContextParser` so that other prefix context parsers that may be added in the future will not break this PR

akosyakov:
@mustard-mh I don't understand which problem we are solving here? Could you elaborate in description what this PR does?

andreafalzetti:
I am rebasing and fixing merge conflicts

andreafalzetti:
> @mustard-mh I don't understand which problem we are solving here? Could you elaborate in description what this PR does?

@akosyakov I've updated the PR description, in short: it's a follow-up improvement for our PLG experiment around IDE awareness.

components/dashboard/src/start/CreateWorkspace.tsx
```diff
@@ -250,6 +251,10 @@ export default class CreateWorkspace extends React.Component<CreateWorkspaceProp
                     onPrebuildSucceeded={() => this.createWorkspace(CreateWorkspaceMode.UsePrebuild)}
                 />
             );
+        } else if (result?.needOnboardingIde) {
+            statusMessage = (
+                <SelectIDEModal onClose={() => this.createWorkspace(CreateWorkspaceMode.Default)}></SelectIDEModal>
```
andreafalzetti, line 256:
```suggestion
                <SelectIDEModal onClose={() => this.createWorkspace(CreateWorkspaceMode.Default)}/>
```

andreafalzetti (approved):
It works as advertised: As a first-time user, using a context URL, I am prompted to select my preferred IDE/editore before the workspace starts.

When I initially approached this task before this PR was even created, I tried using the information in context or making an additional call to know if workspaces existed, but I admit that this approach is better because it extends the current architecture.

Perhaps I would rename `needOnboardingIde` in `needUserIdePreference`. What do you think, @mustard-mh?

Approved - the rename is not a blocker for me, only a preference and suggestion. I look forward to seeing this in production 🚀

gtsiolis (approved):
UX looks flawless! ✨ 

Thanks, @mustard-mh! 💯

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -1063,6 +1064,10 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
                 }).catch();
             }
 
+            if (User.isOnboardingUser(user) && (!IPrefixContextParserContext.is(context) || !context.specifiedIDE)) {
```
akosyakov, line 1067: why `User.isOnboardingUser(user) && !WithReferrerContext.is(context)` does not work?

components/server/src/workspace/context-parser.ts
akosyakov: I agree that leaking info about a single context on generic interface is not really great. Why don't we only check for `WithReferrerContext`?

mustard-mh: @akosyakov  `WithReferrerContext` works, but if someone add a new prefix like `/#editor` () in the future and don't know this experiment, the prefix will not work for first-time users. That's why we extend `IPrefixContextParser`.

If you don't think so, check with `WithReferrerContext` is also good for me

akosyakov: When we can discuss it now? And we get away then with a super context like `IDEAwareContext = WithReferrerContext | WithEditorContext´ and one is check which delegates both? 

I don't think a result should be propagated via boolean properties.

loujaybee:
Looks great @mustard-mh and co! I'm currently looking into measurement of the feature. I think there are some improvements that we can make, but I don't think it's necessary to block this from being merged / deployed, we can apply analytics improvements as an addition / follow up change.

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -1063,6 +1063,10 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
                 }).catch();
             }
 
+            if (User.isOnboardingUser(user) && !IDESpecifiedContext.is(context)) {
+                return { showUserIdePreference: true };
```
AlexTugarev, line 1067: Why sending back such a dashboard-specific signal?
That doesn't fit into the existing `WorkspaceCreationResult` shape. 

Is it possible to rewrite in a way, that this is checked on frontend, so that `createWorkspace` doesn't need to be changed.

AlexTugarev: see `ContextURL` from https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-protocol/src/context-url.ts

components/gitpod-protocol/src/protocol.ts
```diff
@@ -968,6 +968,16 @@ export namespace WithReferrerContext {
     }
 }
 
+export type IDESpecifiedContext = WithReferrerContext;
```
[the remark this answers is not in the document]
AlexTugarev: as long as this doesn't introduce more specifics to the type, using WithReferrerContext should be fine. I'm not sure if aliasing it here make it more readable.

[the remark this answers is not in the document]
mustard-mh: Will add a new prefix context like `/#editor` later

components/server/src/workspace/gitpod-server-impl.ts
mustard-mh: It can be here, but the check will be more weak because prefix maybe failed in server regex

AlexTugarev: Which regex? 

The question is, can we test `User.isOnboardingUser(user) && !IDESpecifiedContext.is(context)` on frontend? 
If so, let's do it, and avoid an extra roundtrip to server.

components/gitpod-protocol/src/protocol.ts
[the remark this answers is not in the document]
AlexTugarev: Let's add things when they are used. 🙏🏻

components/server/src/workspace/gitpod-server-impl.ts
mustard-mh: regex like this https://github.com/gitpod-io/gitpod/blob/45f13cf8e9214db664adc23a418e53d92364fbb7/components/server/src/workspace/referrer-prefix-context-parser.ts#L13-L17

I will go with check URL prefix on frontend only

components/gitpod-protocol/src/protocol.ts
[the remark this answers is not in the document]
mustard-mh: updated🙈 cc @AlexTugarev

mustard-mh:
> [After Reset Account] Prefix with special IDE /#referrer but args not correctly fit referrer required will use default ide code to start workspace

Update PR description to add a new test, since our check conditions changed

components/dashboard/src/App.tsx
```diff
@@ -495,12 +498,26 @@ function App() {
...
+            toRender = (
+                <StartPage phase={StartPhase.Checking}>
+                    <SelectIDEModal onClose={() => setShowUserIdePreference(false)} />
+                </StartPage>
+            );
+        } else {
+            toRender = <CreateWorkspace contextUrl={hash} />;
+        }
     } else if (isWsStart) {
         toRender = <StartWorkspace {...parseProps(hash, window.location.search)} />;
```
mustard-mh, line 522: We don't need to show Modal while restarting a workspace, because `createWorkspace` covered most cases

mustard-mh:
Initially, this PR was created as a quick fix for this bug, as it was simple to modify and would be much sooner than the PR https://github.com/gitpod-io/gitpod/pull/9663 for review.

In reality, both PRs have been in the review state for a long time.

if https://github.com/gitpod-io/gitpod/pull/9663 can be merged first, we can close this one.

mustard-mh:
@AlexTugarev Any other input about this PR 🙏?

components/dashboard/src/App.tsx
```diff
@@ -495,12 +498,26 @@ function App() {
         );
         return <div></div>;
     }
+    // Prefix with `/#referrer` will specify an IDE for workspace
+    // We don't need to show IDE preference in this case
+    const shouldUserIdePreferenceShown = User.isOnboardingUser(user) && !hash.startsWith(ContextURL.REFERRER_PREFIX);
+    if (shouldUserIdePreferenceShown !== showUserIdePreference) {
+        setShowUserIdePreference(shouldUserIdePreferenceShown);
```
AlexTugarev, line 505: setting a state here will result in an additional render cycle, AFAIC.

you can simplify with a `const` here:
```
const showUserIdePreference = shouldUserIdePreferenceShown !== showUserIdePreference;
```

mustard-mh: @AlexTugarev Prev solution is because we have this one in App.tsx, maybe need to change it later?
https://github.com/gitpod-io/gitpod/blob/b29409611ef14ad4df628e24e309882391e90f8a/components/dashboard/src/App.tsx#L331-L334

mustard-mh:
/hold 

Wait for my final test after built

loujaybee:
@mustard-mh some [context on the decision to remove re-approve here](https://gitpod.slack.com/archives/C01KGM9AW4W/p1650874752728509)

Issue #9678, opened 2022-05-02
clean-slate-deployment should behave gracefully if there is nothing to clean


meysholdt:
## Is your feature request related to a problem? Please describe
Right now, the deploy-to-preview-environments on `main` fails, because the job has `clean-slate-deployment` enabled but there is no old preview environment to clean.

Example: https://werft.gitpod-dev.com/job/gitpod-build-main.3090

## Describe the behaviour you'd like
Don't fail if `clean-slate-deployment` is enabled but there is nothing to clean.

mads-hartmann:
Latest build on main succeeded https://werft.gitpod-dev.com/job/gitpod-build-main.3092 ☺️

PR #9681, opened 2022-05-02
Ensure deleteVM is idempotent

mads-hartmann:
## Description

Jobs using `with-clean-slate-deployments` would fail if there was no VM for the branch. This has caused preview environment deployments to fail on main as we don't have a VM-based preview environment there yet.

This PR adds `--ignore-not-found=true` so that `kubectl delete` will succeed if the resource doesn't exist. That's the behaviour we want in this case.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/9678

## How to test

I could reproduce this by deleting the entire preview-* namespace for my branch.

```
kubectl delete ns preview-mads-9678-830ab3680f
```

And then trigger a new job

```
werft job run github -a with-clean-slate-deployment=true
```

It failed with the same error as described in the original issue, see [here](https://werft.gitpod-dev.com/job/gitpod-build-mads-9678-clean-slate.2/raw).

I ran a job with the code on this branch and it succeeded, see [here](https://werft.gitpod-dev.com/job/gitpod-build-mads-9678-clean-slate.3).

I also triggered a 2nd `with-clean-slate-deployment` job ([link](https://werft.gitpod-dev.com/job/gitpod-build-mads-9678-clean-slate.5)) to verify it still performs the deletion when the VM does exist ☺️

## Release Notes

```release-note
NONE
```

## Documentation

N/A

### DIFF
--- a/.werft/vm/vm.ts
+++ b/.werft/vm/vm.ts
@@ -19,9 +19,9 @@ EOF
 /**
  * Convenience function to kubectl delete a manifest from stdin.
  */
-function kubectlDeleteManifest(manifest: string, options?: { validate?: boolean }) {
+function kubectlDeleteManifest(manifest: string) {
     exec(`
-        cat <<EOF | kubectl --kubeconfig ${HARVESTER_KUBECONFIG_PATH} delete -f -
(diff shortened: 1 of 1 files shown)


meysholdt (approved):
Code changes LGTM. Thank you!

PR #9689, opened 2022-05-02
[agent-smith] Remove egress check feature

aledbf:
## Description

In scenarios where bandwidth limit should be implemented, use the [CNI bandwidth plugin](https://www.cni.dev/plugins/current/meta/bandwidth/) and a custom default template in Gitpod:

```
  default.yaml: |
    metadata:
      annotations:
        kubernetes.io/ingress-bandwidth: 200M
        kubernetes.io/egress-bandwidth: 200M
```

## Release Notes
```release-note
[agent-smith] Remove egress check feature
```

Edit: the CNI approach is also more permissive because it does not terminate the workspace


aledbf:
the second commit is required to fix a CI issue:
```
level=error msg="Running error: 1 error occurred:\n\t* can't run linter goanalysis_metalinter: buildir: failed to load package : could not load export data: no export data for \"github.com/hashicorp/golang-lru\"\n\n"
```

Furisto (approved):

Pothulapati (approved):
LGTM, on the installer side.

easyCZ (approved):
Only an update to `go.mod` for webapp which looks to be transitive. LGTM

PR #9691, opened 2022-05-02
Extend cron job to handle VM-based preview envs

mads-hartmann:
## Description

This PR extends our platform-delete-preview-environments-cron job to also handle VM-based preview environments.

It introduces a lightweight modelling of our preview environments as `type PreviewEnvironment = CoreDevPreviewEnvironment | HarvesterPreviewEnvironment` and moves the related code to each `class`. This should simplify removing `CoreDevPreviewEnvironment` in the future.

This only checks for staleness of Harvester-based preview environments based on the git branch and not the DB activity in the preview environment for now.

Deleting preview environments seems to sometimes fail. I changed to job so that it will fail the slice for deleting a specific preview environments, but not the entire job. That means we're more likely to get through the job. Any deletion that fails might work on the 2nd or 3rd try (in my experience). In the future we can try to improve the reliability of preview environment deletion, but for now I think it's worth merging the PR as is - I created a quick Honeycomb board to help us visualise failure rates of preview environment deletions [here](https://ui.honeycomb.io/gitpod/board/E8HEt73Ao6W/Preview-Environments) and I can set up a quick experimental trigger if the failure rate is above e.g. 15% so we can be proactive about this.

I also I added a bit of span attributes so we know the counts of preview environments in core-dev and harvester. Thought this would be a fun way to get historic counts and also set up a simple threshold-based trigger in Honeycomb ([example query](https://ui.honeycomb.io/gitpod/datasets/werft/result/wftNQuiQ8)) so we can get a Slack message if we're running say 35 Harvester preview environments ☺️

Lastly, I ended up making the job a bit less noisy slice-wise as it was becoming very hard to make sense of the Werft logs. 

Sorry for a doing so much in one PR.

## Related Issue(s)

Part of https://github.com/gitpod-io/ops/issues/1713

## How to test

During development I ran it in DRY_RUN mode by modifying the variable and side-loading the file `werft run github -j .werft/platform-delete-preview-environments-cron.yaml -s .werft/platform-delete-preview-environments-cron.ts`

Once I had verified the preview environments were indeed stale I ran it without DRY_RUN using `werft run github -j .werft/platform-delete-preview-environments-cron.yaml `.

## Release Notes

```release-note
NONE
```

## Documentation

N/A

### DIFF
--- a/.werft/platform-delete-preview-environments-cron.ts
+++ b/.werft/platform-delete-preview-environments-cron.ts
@@ -6,6 +6,7 @@ import { exec } from './util/shell';
 import { previewNameFromBranchName } from './util/preview';
 import { CORE_DEV_KUBECONFIG_PATH, HARVESTER_KUBECONFIG_PATH } from './jobs/build/const';
 import {deleteDNSRecord} from "./util/gcloud";
+import * as VM from "./vm/vm";
 
 // for testing purposes
 // if set to 'true' it shows only previews that would be deleted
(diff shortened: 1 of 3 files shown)


meysholdt (approved):
Code changes LGTM

PR #9701, opened 2022-05-02
kots: Support s3 backend with incluster registry

Pothulapati:
## Description

Currently, There seems to be a configuration issue, around
s3 backend, and docker registry. The s3 option is available
when `incluster` is disabled, which is confusing as s3
backend is supported when `incluster` registry is enabled.

This PR fixes that by having this configured correctly.

## How to test

With `kots`, You should now see an Storage backend option when incluster
registry is enabled.

## Release Notes

```release-note
[kots] support s3 backend in incluster registry
```

### DIFF
--- a/install/kots/manifests/gitpod-registry-s3-backend.yaml
+++ b/install/kots/manifests/gitpod-registry-s3-backend.yaml
@@ -9,7 +9,7 @@ metadata:
     app: gitpod
     component: gitpod-installer
   annotations:
-    kots.io/when: '{{repl and (ConfigOptionEquals "reg_incluster" "0") (ConfigOptionEquals "reg_s3storage" "1") }}'
+    kots.io/when: '{{repl and (ConfigOptionEquals "reg_incluster" "0") (ConfigOptionEquals "reg_incluster_storage" "s3") }}'
 data:
-  s3AccessKey: '{{repl ConfigOption "reg_accesskey" | Base64Encode }}'
-  s3SecretKey: '{{repl ConfigOption "reg_secretkey" | Base64Encode }}'
+  s3AccessKey: '{{repl ConfigOption "reg_incluster_storage_s3_accesskey" | Base64Encode }}'
+  s3SecretKey: '{{repl ConfigOption "reg_incluster_storage_s3_secretkey" | Base64Encode }}'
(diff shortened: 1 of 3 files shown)


install/kots/manifests/gitpod-registry-s3-backend.yaml
```diff
@@ -11,5 +11,5 @@ metadata:
   annotations:
     kots.io/when: '{{repl and (ConfigOptionEquals "reg_incluster" "0") (ConfigOptionEquals "reg_s3storage" "1") }}'
```
MrSimonEmms, line 12: The `kots.io/when` annotation needs updating too

```suggestion
    kots.io/when: '{{repl and (ConfigOptionEquals "reg_incluster" "0") (ConfigOptionEquals "reg_incluster_storage" "s3") }}'
```

install/kots/manifests/kots-config.yaml
```diff
@@ -33,6 +33,39 @@ spec:
           help_text: You may either use an in-cluster container registry or configure your own external container registry for better performance. This container registry must be accessible from your Kubernetes cluster.
           recommended: false
 
+        - name: reg_incluster_storage
+          title: In-cluster Storage provider
+          type: select_one
+          when: '{{repl (ConfigOptionEquals "reg_incluster" "1") }}'
+          default: none
+          help_text: You may configure your docker registry to use an external storage backend.
```
MrSimonEmms, line 41:
```suggestion
          help_text: You may configure your Docker registry to use an external storage backend.
```

MrSimonEmms (commented):
I like this change. Just a couple of minor issues to fix and it'll be good to go

install/kots/manifests/kots-config.yaml
Pothulapati: It is worth! but this is still `in-cluster` registry right? Here we are just asking `incluster` registry to use an `S3` backend.

This configuration is passed down to the `registry` deployments ENV, but not replace it 🤔

Pothulapati: My worry is that the suggested second sentence could be confusing as it is suggesting to use AWS Elastic Container Registry but we are just using S3 right? 🤔

MrSimonEmms: As far as my understanding goes, AWS ECR doesn't work with Gitpod because ECR doesn't automatically create a new repo if it doesn't exist when pushing the image. For that reason, if you're using AWS as your platform, I believe that you have to use in-cluster registry with S3 backing (as you've done).

Maybe the sentence should be `This setting is recommended for AWS users instead of using Elastic Container Registry` and then (future PR) link to something in the docs that explains this in more detail.

Pothulapati: Yep, That makes sense! Updated

MrSimonEmms (approved):

PR #9703, opened 2022-05-02
[public-api] Extract BearerToken from request

easyCZ:
## Description

First phase of adding authentication to public API
* We check that a token exists, but otherwise do no extra validation. This is acceptable, as we're only returning mock data at this point.
* Eventually, this will move into a middleware and will automatically attach extracted user context into the request context, but we don't yet need this.

## Related Issue(s)

## How to test

* Unit tests

## Release Notes

```release-note
NONE
```

## Documentation

NONE


laushinka:
Approving! Use of the bearerToken will be in a follow-up PR.

laushinka (approved):

PR #9705, opened 2022-05-02
[server] Start XL workspaces with internal-xl class

csweichel:
## Description
This change starts workspaces which should get more resources using the `gitpodio-internal-xl` class.
Note: we have to deploy workspace clusters which support this class first.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/server/src/workspace/workspace-starter.ts
+++ b/components/server/src/workspace/workspace-starter.ts
@@ -1294,6 +1294,11 @@ export class WorkspaceStarter {
             ideImage = ideConfig.ideOptions.options[ideConfig.ideOptions.defaultIde].image;
         }
 
+        let workspaceClass: string = "";
+        if (await this.userService.userGetsMoreResources(user)) {
+            workspaceClass = "gitpodio-internal-xl";
+        }
+
         const spec = new StartWorkspaceSpec();
         await createGitpodTokenPromise;
         spec.setEnvvarsList(envvars);
@@ -1309,6 +1314,7 @@ export class WorkspaceStarter {
         spec.setWorkspaceImage(instance.workspaceImage);
         spec.setWorkspaceLocation(workspace.config.workspaceLocation || checkoutLocation);
         spec.setFeatureFlagsList(this.toWorkspaceFeatureFlags(featureFlags));
+        spec.setClass(workspaceClass);
         if (workspace.type === "regular") {
             spec.setTimeout(this.userService.workspaceTimeoutToDuration(await userTimeoutPromise));
         }
--- a/components/server/src/workspace/workspace-starter.ts
+++ b/components/server/src/workspace/workspace-starter.ts
@@ -1294,7 +1294,7 @@ export class WorkspaceStarter {
             ideImage = ideConfig.ideOptions.options[ideConfig.ideOptions.defaultIde].image;
         }
 
-        let workspaceClass: string = "";
+        let workspaceClass: string = "default";
         if (await this.userService.userGetsMoreResources(user)) {
             workspaceClass = "gitpodio-internal-xl";
         }


components/server/src/workspace/workspace-starter.ts
```diff
@@ -1251,6 +1251,11 @@ export class WorkspaceStarter {
             ideImage = ideConfig.ideOptions.options[ideConfig.ideOptions.defaultIde].image;
         }
 
+        let workspaceClass: string = "";
+        if (await this.userService.userGetsMoreResources(user)) {
+            workspaceClass = "gitpod.io/internal/xl";
```
aledbf, line 1256: We cannot use that as a node label

`May 10 00:13:11 headless-ws-ephemeral86-pool-internal-xl-vp2f k3s[2536]: E0510 00:13:11.540005    2536 kubelet_node_status.go:92] "Unable to register node with API server" err="Node \"headless-ws-ephemeral86-pool-internal-xl-vp2f\" is invalid: metadata.labels: Invalid value: \"gitpod.io/internal/xl\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')" node="headless-ws-ephemeral86-pool-internal-xl-vp2f"`

components/server/src/workspace/workspace-starter.ts
```diff
@@ -1251,6 +1251,11 @@ export class WorkspaceStarter {
             ideImage = ideConfig.ideOptions.options[ideConfig.ideOptions.defaultIde].image;
         }
 
+        let workspaceClass: string = "";
```
aledbf, line 1254: Can we set "default" instead of an empty string? Same issue with labels, we need a value :(

aledbf:
@csweichel if we introduce the label validation, then we don't need https://github.com/gitpod-io/gitpod/pull/9868
(only to change the default class to "default" - or something)

components/ws-manager/pkg/manager/testdata/cdwp_class.golden
```diff
@@ -8,7 +8,7 @@
                 "app": "gitpod",
                 "component": "workspace",
                 "gitpod.io/networkpolicy": "default",
-                "gitpod.io/workspaceClass": "foobar",
+                "gitpod.io/workspaceClass": "gitpod.io/internal/foobar",
```
aledbf, line 11: this is not a valid label value

csweichel: now I get it - sorry for that. I was under the impression that label values merely had a size limit, but was unaware of the character limitation 🤦

aledbf: I think we should use https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L167 to ensure we are setting a valid value

csweichel: 💯 will add

components/server/src/workspace/workspace-starter.ts
aledbf: `let workspaceClass: string = "default";` ?

atduarte: Would suggest `standard` as we have been using that name already

aledbf: > Would suggest standard as we have been using that name already

We already use that in [code (go)](https://github.com/gitpod-io/gitpod/blob/3939796825c604fe1e3fa6c9a59993a8abbb1ce6/components/ws-manager-api/go/config/config.go#L29) and also the [ops](https://github.com/gitpod-io/ops/blob/45b7d2ea558a4716e1f8fdbd149391f98096f4e9/deploy/workspace/manifests/workspace-templates.yaml#L12) repo

aledbf:
@csweichel I added a commit to set the default workspace class name to `default`

aledbf (approved):

aledbf:
@csweichel @geropl this is the same PR as https://github.com/gitpod-io/gitpod/pull/10351 deployed today in the two meta clusters.

AlexTugarev (approved):
LGTM

jankeromnes:
This needs a rebase, right?

```
[Validating configuration] STDERR: Error: error loading config: error unmarshaling JSON: while decoding JSON: json: unknown field "usage"
```

PR #9706, opened 2022-05-02
[supervisor] Improve IDE readiness probe

aledbf:
## Description

- Reduce test window
- Properly close body response
- Run a probe and wait instead of waiting for initial test
 
## Release Notes
```release-note
[supervisor] Improve IDE readiness probe
```

### DIFF
--- a/components/supervisor/pkg/supervisor/supervisor.go
+++ b/components/supervisor/pkg/supervisor/supervisor.go
@@ -934,6 +934,7 @@ func runIDEReadinessProbe(cfg *Config, ideConfig *IDEConfig, ide IDEKind) (deskt
 	if ide == DesktopIDE {
 		defaultProbePort = desktopIDEPort
 	}
+
 	switch ideConfig.ReadinessProbe.Type {
 	case ReadinessProcessProbe:
 		return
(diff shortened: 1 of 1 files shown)


components/supervisor/pkg/supervisor/supervisor.go
```diff
@@ -945,42 +945,47 @@ func runIDEReadinessProbe(cfg *Config, ideConfig *IDEConfig, ide IDEKind) (deskt
...
+			tick   = time.NewTicker(250 * time.Millisecond)
 		)
 		defer tick.Stop()
 
 		t0 := time.Now()
 
-		for {
-			<-tick.C
-
+		for ; true; <-tick.C {
```
jeanp413, line 954: I think this doesn't read nice, how about `for _ := range tick.C` ?

components/supervisor/pkg/supervisor/supervisor.go
```diff
@@ -945,42 +945,47 @@ func runIDEReadinessProbe(cfg *Config, ideConfig *IDEConfig, ide IDEKind) (deskt
...
+
+			err = json.Unmarshal(bodyBytes, &desktopIDEStatus)
+			if err != nil {
+				log.WithField("ide", ide.String()).WithError(err).WithField("body", bodyBytes).Debugf("Error parsing JSON body from IDE status probe.")
 				break
 			}
 
-			log.WithField("ide", ide.String()).WithField("status", resp.StatusCode).Info("IDE readiness probe came back with non-200 status code")
+			log.WithField("ide", ide.String()).Infof("Desktop IDE status: %s", desktopIDEStatus)
+			break
```
jeanp413, line 988: I think it would be better to move all this logic outside the loop so we don't have that many `breaks`

components/supervisor/pkg/supervisor/supervisor.go
aledbf: @jeanp413 that will wait for a tick. That said, if you prefer that, I can change the code

jeanp413: isn't `for ; true; <-tick.C` also waiting for the tick? if it's not waiting then why we need the tick at all :thinking:

components/supervisor/pkg/supervisor/supervisor.go
aledbf: fixed

components/supervisor/pkg/supervisor/supervisor.go
aledbf: changed

jeanp413:
Status readiness for vscode browser is broken now :sweat_smile: 
I think only this should be inside the loop, response parsing etc should be after breaking the loop
```
	client := http.Client{Timeout: 1 * time.Second}
	resp, err := client.Get(url)
	if err != nil {
		log(...)
	}

	if resp.StatusCode == http.StatusOK {
              break;
	}
```

aledbf:
@jeanp413 ready for another review

jeanp413 (approved):
Works for vscode browser, desktop and jetbrains

PR #9717, opened 2022-05-03
[installer]: Add server chargebee configuration to the installer

andrew-farries:
## Description

This PR makes it possible to configure the `server` component's chargebee configuration via new configuration surface in the `experimental` section of the installer config that allows specifying the name of a secret containing chargebee config. For installations on which payments should be enabled, we expect there to be a secret present in the cluster containing this configuration. For Gitpod SaaS, this is achieved via terraform as a cluster prerequisite:

https://github.com/gitpod-io/ops/blob/8fb50515adbfa90a3435f2148660f7de8d70c219/terraform/modules/gitpod/main.tf#L31-L38

## Related Issue(s)

Part of #9097 

## How to test

```yaml
experimental:
  webapp:
    server:
      chargebeeSecret: "chargebee-config"
```

```
```

Then invoke the installer as:

```
```

The `server` configmap will have `enablePayment` set to `true` and the `server` deployment will have an extra volume + volumemount:

```yaml
        - mountPath: /chargebee
          name: chargebee-config
          readOnly: true
```

```yaml
      - name: chargebee-config
        secret:
          secretName: chargebee-config
```

## Release Notes

```release-note
Allow chargebee payment config to be specified via the installer for SaaS installations.
```

## Documentation

None.

### DIFF
--- a/install/installer/pkg/components/server/configmap.go
+++ b/install/installer/pkg/components/server/configmap.go
@@ -74,6 +74,14 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		return nil
 	})
 
+	chargebeeSecret := ""
+	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
+		if cfg.WebApp != nil && cfg.WebApp.Server != nil {
+			chargebeeSecret = cfg.WebApp.Server.ChargebeeSecret
+		}
+		return nil
+	})
+
 	githubApp := GitHubApp{}
 	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
 		if cfg.WebApp != nil && cfg.WebApp.Server != nil && cfg.WebApp.Server.GithubApp != nil {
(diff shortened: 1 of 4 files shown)


install/installer/pkg/components/server/configmap.go
```diff
@@ -74,6 +74,14 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		return nil
 	})
 
+	chargebeeConfig := ""
```
andrew-farries, line 77: We've got rid of the default value for `ChargebeeProviderOptionsFile` here which was previously set to `"/chargebee/providerOptions"`, but since `EnablePayment` was always set to `false`, I don't think the value was ever actually used.

andrew-farries: My understanding is that this is a path on a volume (provided by the `chargebee-config` secret that we will create in the outer helm chart). We'll need to add the extra `volume` and `volumeMount` to the server deployment once we've added it.

install/installer/pkg/config/v1/experimental/experimental.go
```diff
@@ -115,6 +115,7 @@ type ServerConfig struct {
 	OAuthServer                       OAuthServer       `json:"oauthServer"`
 	Session                           Session           `json:"session"`
 	GithubApp                         *GithubApp        `json:"githubApp"`
+	ChargebeeSecret                   string            `json:"chargebeeSecret"`
```
andrew-farries, line 118: I'd like to use an [ObjectRef](https://github.com/gitpod-io/gitpod/blob/main/install/installer/pkg/config/v1/config.go#L176) here rather than a `string`, but that would require creating an import cycle `config -> experimental -> config`

geropl (approved):
Thx, tested and looks good! :+1:

install/installer/pkg/config/v1/experimental/experimental.go
corneliusludmann: Is it worth to refactor this? Maybe something for a follow-up issue?

corneliusludmann (approved):
Do you think a single commit instead of 3 for such a rather small change would be also sufficient?

Otherwise, change looks good so far. :rocket: 

Holding in case you want to squash the commits.

/hold

andrew-farries:
> Holding in case you want to squash the commits.

Thanks, I've squashed down to one commit.

install/installer/pkg/config/v1/experimental/experimental.go
andrew-farries: I think it's not worth following up on at this stage.

PR #9721, opened 2022-05-03
[licensor] force preinstall to build `liblicensor.a`

nandajavarma:
## Description

This PR fixes the current break in build when opening a workspace from `gitpod` main.

`licensor/typescript` currently do not rebuild since the build is triggered by the presence of a `.a` file. Since #9343 requires rebuild of the `liblicensor.a` file and `api.ts` (built using `genapi.go`), this PR deletes the `.a` file to force the rebuild.

## Related Issue(s)

Fixes #

## How to test

Open a workspace from this branch, the errors in the build should be resolved.

## Release Notes

```release-note
None
```

### DIFF
--- a/components/licensor/typescript/ee/src/api.ts
+++ b/components/licensor/typescript/ee/src/api.ts
@@ -15,7 +15,7 @@ export enum Feature {
 export interface LicenseData {
     type: LicenseType
     payload: LicensePayload
-    plan: string
+    plan: LicenseSubscriptionLevel
     fallbackAllowed: boolean
 }
 
@@ -31,6 +31,14 @@ export interface LicensePayload {
     seats: number
 }
 
+export enum LicenseSubscriptionLevel {
+    ReplicatedLicenseTypeCommunity = "community",
+    ReplicatedLicenseTypeDevelopment = "dev",
+    ReplicatedLicenseTypePaid = "prod",
+    ReplicatedLicenseTypeTrial = "trial",
+    CommunityLicense = "community",
+    ProfessionalLicense = "prod",
+}
 export enum LicenseType {
     LicenseTypeGitpod = "gitpod",
     LicenseTypeReplicated = "replicated",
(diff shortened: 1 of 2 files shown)


components/licensor/typescript/ee/src/api.ts
```diff
@@ -31,6 +31,14 @@ export interface LicensePayload {
     seats: number
 }
 
+export enum LicenseSubscriptionLevel {
```
MrSimonEmms, line 34: Should there be a duplicate of `community` and `prod`?

nandajavarma: `CommunityLicense` and `ProfessionalLicense` were introduced[ in the licensor code](https://github.com/gitpod-io/gitpod/pull/9343) for the `Gitpod` license levels. This duplication exists now, since I was trying to standardize the license level names and didn't want to change the replicated code too much. We can clean it up in the future, right now this is a requirement of the existing code.

corneliusludmann: Thanks for the quick fix, Nandaja. :partying_face: 

I know, it's less convenient, but I think having 2 separate pull requests (one PR per concern) would help us in general. In this specific case, the discussion about the LicenseSubscriptionLevel blocks merging the build fix. What do you think? :thinking:

corneliusludmann (approved):
Don't know how to test this but changes look good so far.

components/licensor/typescript/ee/src/api.ts
nandajavarma: @corneliusludmann I agree with one issue per PR comment. Unfortunately here, if we force the build, it will update `api.ts`(this file is auto-generated by `genapi`). Meaning that people will have a change to be staged in their workspace which is not very nice. This is a mistake on my part, not including it in the license UI PR. But the fact that changes in this PR regenerates `api.ts` makes this change related to the main issue.

AlexTugarev (approved):
Thanks for the fix!

PR #9741, opened 2022-05-03
[kots]: improve installer job failure recovery

MrSimonEmms:
## Description

If the `https-certificates` secret exists, it sets the `helm upgrade` timeout to 5 minutes. Otherwise, it keeps it at 1 hour. This means that an update has a shorter timeout than an fresh install and allows for users to regain control of their "(re)deploy" button sooner

Also set the `backoffLimit` to `1` to avoid it restarting out of control. If a restart is required, this can be achieved by the user pressing "(re)deploy)

## How to test

Install via KOTS

## Release Notes

```release-note
[kots]: improve installer job failure recovery
```


corneliusludmann (approved):
Awesome! :star_struck:

PR #9756, opened 2022-05-04
[installer] Make runDbDeleter configurable for the server component

andrew-farries:
## Description

This PR makes it possible to configure the `server` component's `runDbDeleter` config field, rather than hardcoding it to `true`.

## Related Issue(s)

Part of #9097 

## How to test

```yaml
experimental:
  webapp:
    server:
      runDbDeleter: false
```

```
```

Then invoke the installer as:

```
```

The `server` configmap will have `runDbDeleter` set to `false`. Without the experimental config, it defaults to `true` as before.

## Release Notes

```release-note
Make `runDbDeleter` for the server configurable via the installer
```

## Documentation

None.

### DIFF
--- a/install/installer/pkg/components/server/configmap.go
+++ b/install/installer/pkg/components/server/configmap.go
@@ -64,6 +64,14 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		return nil
 	})
 
+	runDbDeleter := true
+	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
+		if cfg.WebApp != nil && cfg.WebApp.Server != nil && cfg.WebApp.Server.RunDbDeleter != nil {
+			runDbDeleter = *cfg.WebApp.Server.RunDbDeleter
+		}
+		return nil
+	})
+
 	defaultBaseImageRegistryWhitelist := []string{}
 	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
 		if cfg.WebApp != nil && cfg.WebApp.Server != nil {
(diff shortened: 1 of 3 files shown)


easyCZ (approved):
LGTM

install/installer/pkg/components/server/configmap_test.go
```diff
@@ -30,6 +31,7 @@ func TestConfigMap(t *testing.T) {
 	expectation := Expectation{
 		EnableLocalApp:                    true,
 		DisableDynamicAuthProviderLogin:   true,
+		RunDbDeleter:                      false,
```
geropl, line 34: The default should be true. This is the default I assume?

install/installer/pkg/components/server/configmap.go
```diff
@@ -64,6 +64,14 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		return nil
 	})
 
+	runDbDeleter := true
+	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
+		if cfg.WebApp != nil && cfg.WebApp.Server != nil {
+			runDbDeleter = cfg.WebApp.Server.RunDbDeleter
```
geropl, line 70: IMO wee need to check for nil, and either:
 - default to true (again)
 - or (better) add the != nil check to the if check.

install/installer/pkg/components/server/configmap_test.go
andrew-farries: This is for the test code. I set it to `false` here because it's *not* the default, ie test that setting it explicitly has the desired effect.

 We should probably have multiple test cases that set each of these boolean flags differently.

geropl: Thx, got it!

install/installer/pkg/components/server/configmap.go
andrew-farries: Not sure I follow 😕 . `RunDbDeleter` is a `bool` field so can't be `nil`.

geropl: Woops, my bad.
What I meant is we should distinguish between false (default value in case not specified) and actively set to false. The former we need to identify and in ignore.
Maybe by making that field a `*bool`?

andrew-farries: Rebased to make it a pointer. The same also applies to `enableLocalApp` which should also default to `true` when unset (8d1af9f7c8e74645f40cf44e3daeb6969dd2c392).

geropl (approved):
Thank you! :pray:

corneliusludmann (approved):
:balloon:

PR #9759, opened 2022-05-04
[kots]: add database to preflight checks

### LINKED ISSUE #9931: KOTS database preflight checks

MrSimonEmms:
## Description

Add a database check to the KOTS preflight checks. This validates that the given configuration is able to make a connection to a MySQL database called `gitpod` and that the version is `5.7.x` - this version check is considered a warning because it's possible that we get anomalies with the versions

### Failed

image

### Pass

image

## Related Issue(s)

Fixes #9931

## How to test

Install with KOTS - the version to use is `sje-kots-config-check.11` in `sje/kots-config-check` channel

Now change the config values for database in KOTS:
 - in-cluster should always pass for both checks
 - external should pass with correct credentials (check that DB's firewall allows connections - can make public with `0.0.0.0/0`)
 - cloud SQL should pass with correct credentials (ensure firewall is closed)

You can create a support bundle and look in the file explorer for `database/database.log` to see the database that's output

## Release Notes

```release-note
[kots]: add database to preflight checks
```

### DIFF
--- a/components/BUILD.yaml
+++ b/components/BUILD.yaml
@@ -70,6 +70,7 @@ packages:
       - components/ws-manager:docker
       - components/ws-proxy:docker
       - components/ide-proxy:docker
+      - components/kots-config-check/database:docker
       - test:docker
       - dev/version-manifest:app
     config:
--- a/components/kots-config-check/database/entrypoint.sh
+++ b/components/kots-config-check/database/entrypoint.sh
@@ -1,0 +1,87 @@
+#!/bin/bash
+# Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+# Licensed under the GNU Affero General Public License (AGPL).
+# See License-AGPL.txt in the project root for license information.
+
+
+set -euo pipefail
+
+DB_IN_CLUSTER_ENABLED="${1:-""}"
+DB_CLOUDSQL_ENABLED="${2:-""}"
+DB_USERNAME="${3:-""}"
+DB_PASSWORD="${4:-""}"
+DB_HOST="${5:-""}"
+DB_PORT="${6:-""}"
+CSP_INSTANCES="${7:-""}"
+CSP_CREDENTIALS="${8:-""}"
+
+connection="false"
+version=""
+
+DB_TYPE="incluster"
+if [ "${DB_IN_CLUSTER_ENABLED}" == "0" ]; then
+  if [ "${DB_CLOUDSQL_ENABLED}" == "1"  ]; then
+    DB_TYPE="cloudsqlproxy"
+  else
+    DB_TYPE="external"
+  fi
+fi
+
+case "${DB_TYPE}" in
+  cloudsqlproxy | external)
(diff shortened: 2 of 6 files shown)


install/kots/manifests/kots-preflight.yaml
```diff
@@ -7,6 +7,19 @@ metadata:
   name: gitpod
 spec:
   collectors:
+    - run:
+        collectorName: database
+        image: eu.gcr.io/gitpod-core-dev/build/kots-config-check/database:sje-kots-config-check.9
```
corneliusludmann, line 12: Will this work in an air-gap setting as well? I guess that KOTS will handle this. However, we need to make sure it will before we can merge this.

corneliusludmann (commented):
Thanks for adding this check. That is very useful! :rocket: 

Added a few comments. The question, of whether this will work for air-gap is the most important one. I would expect that is works. However, we need to make sure before we release it.

The other comments can be handled in follow-up issues and are no blockers, in my opinion.

components/kots-config-check/database/entrypoint.sh
```diff
@@ -0,0 +1,87 @@
...
+    # Check the database version
+    version_query=$(mysql \
+      --connect-timeout=5 \
+      --database=gitpod \
+      --user="${DB_USERNAME}" \
+      --password="${DB_PASSWORD}" \
+      --host="${DB_HOST}" \
+      --port="${DB_PORT}" \
+      --execute="SELECT VERSION();" \
+      --silent \
```
corneliusludmann, line 62: Random thought: Do we still get enough log data with the `silent` flag for the support bundle or should we remove this flag?

install/kots/manifests/kots-preflight.yaml
MrSimonEmms: Yes it will. It's exactly the same as the [Installer job](https://github.com/gitpod-io/gitpod/blob/main/install/kots/manifests/gitpod-installer-job.yaml#L31)

components/kots-config-check/database/entrypoint.sh
MrSimonEmms: Yes. The `--silent` flag basically removes all the connection information. At this point, we're only interested in the version so we can assign it to the `version_query` variable. It is therefore correct that we have the `--silent` flag.

Also, there's actually no way of debugging a preflight check so we are limited to just putting the data in a known format so we can perform regex queries on it.

components/BUILD.yaml
```diff
@@ -70,6 +70,7 @@ packages:
       - components/ws-manager:docker
       - components/ws-proxy:docker
       - components/ide-proxy:docker
+      - components/kots-config-check/database:docker
```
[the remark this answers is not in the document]
MrSimonEmms: Yeah, that would be a good optimisation. Is there an example I can use to follow as I didn't know this was possible?

MrSimonEmms:
🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

PR #9773, opened 2022-05-04
[installer] Add `serviceAnnotations` config to `proxy`

andrew-farries:
## Description

This PR allows for extra annotations to configured on the `proxy` component so that we can configure GCP annotations for Gitpod SaaS.

## Related Issue(s)
Part of #9097 

## How to test

```yaml
experimental:
  webapp:
    proxy:
      serviceAnnotations:
        hello: world
```

```
```

Then invoke the installer as:

```
```

The `Service` resource for `proxy` will contain the extra annotation.

## Release Notes

```release-note
Allow setting `proxy` service annotations via the installer.
```

## Documentation

None.

### DIFF
--- a/install/installer/pkg/components/proxy/service.go
+++ b/install/installer/pkg/components/proxy/service.go
@@ -23,6 +23,14 @@ func service(ctx *common.RenderContext) ([]runtime.Object, error) {
 		return nil
 	})
 
+	var annotations map[string]string
+	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
+		if cfg.WebApp != nil && cfg.WebApp.ProxyConfig != nil {
+			annotations = cfg.WebApp.ProxyConfig.ServiceAnnotations
+		}
+		return nil
+	})
+
 	ports := map[string]common.ServicePort{
 		ContainerHTTPName: {
 			ContainerPort: ContainerHTTPPort,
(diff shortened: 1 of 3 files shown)


corneliusludmann:
@Pothulapati, just out of curiosity: Would these be the annotations we would need for the AWS/EKS load balancer as well?

install/installer/pkg/components/proxy/service.go
```diff
@@ -47,7 +55,12 @@ func service(ctx *common.RenderContext) ([]runtime.Object, error) {
 	return common.GenerateService(Component, ports, func(service *corev1.Service) {
 		service.Spec.Type = corev1.ServiceTypeLoadBalancer
 		service.Spec.LoadBalancerIP = loadBalancerIP
+
 		service.Annotations["external-dns.alpha.kubernetes.io/hostname"] = fmt.Sprintf("%s,*.%s,*.ws.%s", ctx.Config.Domain, ctx.Config.Domain, ctx.Config.Domain)
 		service.Annotations["cloud.google.com/neg"] = `{"exposed_ports": {"80":{},"443": {}}}`
+
+		for k, v := range annotations {
+			service.Annotations[k] = v
```
geropl, line 63: :100:

geropl (approved):
:100: Thx for tests!

install/installer/pkg/components/proxy/service_test.go
```diff
@@ -27,13 +27,28 @@ func TestServiceLoadBalancerIP(t *testing.T) {
...
+
+	ctx := renderContextWithProxyConfig(t, &experimental.ProxyConfig{ServiceAnnotations: annotations})
+
+	objects, err := service(ctx)
+	require.NoError(t, err)
+
+	require.Len(t, objects, 1, "must render only one object")
+
+	svc := objects[0].(*corev1.Service)
+	for k, v := range annotations {
```
easyCZ, line 41: Should be possible to use `require.EqualValues(t, annotations, svc.Annotations)`

MrSimonEmms (approved):

install/installer/pkg/components/proxy/service_test.go
andrew-farries: What we are checking here is slightly different; we need to check that the extra annotations are a *subset* of those on the rendered service. The installer hard-codes a couple of annotations too.

PR #9779, opened 2022-05-04
[installer] Minor alignments + fixes

geropl:
## Description
This PR contains 2 minor fixes:
 - set a node affinity for the cloudsql deployment
 - align server ports between service and deployment

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/install/installer/pkg/components/database/cloudsql/deployment.go
+++ b/install/installer/pkg/components/database/cloudsql/deployment.go
@@ -6,6 +6,7 @@ package cloudsql
 
 import (
 	"fmt"
+	"github.com/gitpod-io/gitpod/installer/pkg/cluster"
 	"github.com/gitpod-io/gitpod/installer/pkg/common"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


install/installer/pkg/components/database/cloudsql/deployment.go
```diff
@@ -43,7 +44,9 @@ func deployment(ctx *common.RenderContext) ([]runtime.Object, error) {
 						Labels:    labels,
 					},
 					Spec: corev1.PodSpec{
-						Affinity:                      &corev1.Affinity{},
+						Affinity: &corev1.Affinity{
```
easyCZ, line 47: Does it make sense to extend the tests to check this is set correctly?

easyCZ (approved):
LGTM, left a comment. Adding a hold in case you want to tackle it in this PR.

/hold

install/installer/pkg/components/database/cloudsql/deployment.go
geropl: Thx for finding this! Excellent idea, but I fear personally I won't time today, so opting for the "follow up" choice.

MrSimonEmms (approved):

PR #9780, opened 2022-05-04
[installer-diff] Filter out unwanted objects, and reduce noise

geropl:
## Description
This PR fine-tunes the installer-diff tool to reduce noise by:
 - filtering out unwanted objects
 - renames certain objects to for easier diffing
 - sorts Service ports by name

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/dev/installer-diff/cmd/filter.go
+++ b/dev/installer-diff/cmd/filter.go
@@ -34,20 +34,29 @@ var filterCmd = &cobra.Command{
 			log.Panic(err)
 		}
 
-		// sort by .kind and .metadata.name
-		sort.SliceStable(objs, func(i, j int) bool {
-			id := func(i int) string {
-				return fmt.Sprintf("%s:%s", objs[i].GetKind(), objs[i].GetName())
-			}
-			return id(i) < id(j)
-		})
-
 		var outObjs []unstructured.Unstructured
 		for _, obj := range objs {
 			// filter out generic stuff: .status, .metadata.annotations, etc.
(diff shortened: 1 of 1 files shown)


dev/installer-diff/cmd/filter.go
```diff
@@ -133,7 +153,7 @@ func filterGenericStuff(obj *unstructured.Unstructured) (err error) {
 	// (needs to be below the kind-specific stuff to override converter artifacts!)
 	obj.SetAnnotations(emptyMap)
 	obj.SetNamespace("")
-	obj.SetCreationTimestamp(v1.Time{})	// results in "null". How to set to "nil" w/o writing by hand?
+	obj.SetCreationTimestamp(v1.Time{}) // results in "null". How to set to "nil" w/o writing by hand?
```
easyCZ, line 156: Looks like it's not possible. It ends up with value of `null` because you're giving time of 0 (default value) but the API looks very determined to ensure this isn't set to an invalid value. I couldn't find a good way to do this.

easyCZ (approved):
Code-wise LGTM, I don't have intimate knowledge of the logic so if you're looking for context review, let's wait for @andrew-farries. Otherwise feel free to land.

/hold

dev/installer-diff/cmd/filter.go
geropl: > Looks like it's not possible

Jep. Setting to a common value works well for us. :+1:

PR #9297, opened 2022-04-13
Add experimental config values for the Server component

andrew-farries:
## Description

One of the Webapp team's [epics for Q2](https://github.com/gitpod-io/gitpod/issues/9097) is to use the Gitpod installer to deploy to Gitpod SaaS. In order to do that we may need to add additional configuration to the installer to make the output suitable for a SaaS deployment as opposed to a self-hosted deployment.

This PR adds configuration for a few values which are currently hard-coded in the installer but need to be configurable for SaaS:

* `workspaceDefaults.workspaceImage`
*  `oauthServer.jwtSecret`
* `session.secret`
* `githubApp`

## Related Issue(s)

Part of https://github.com/gitpod-io/gitpod/issues/9097

## How to test

Add an experimental section to the bottom of an installer config file like this:

```yaml
experimental:
  webapp:
    workspaceDefaults:
      workspaceImage: "someWorkspaceImage"
    oauthServer:
      jwtSecret: "someJwtSecret"
    session:
      secret: "someSessionSecret"
    githubApp:
      appId: 123
      authProviderId: 'someAuthProviderId'
      baseUrl: 'someBaseUrl'
      certPath: 'someCertpath'
      enabled: true
      logLevel: 'someLogLevel'
      marketplaceName: 'someMarketPlacename'
      webhookSecret: 'somewebhookSecret'
      certSecretName: 'certSecretName'

```

and invoke this installer:

```
installer render --config /path/to/config --use-experimental-config
```

The `server-config` configMap should then contain the values set in the installer config:
<details>
<summary>config.json</summary>

```yaml
      {
        "version": "main.2686",
        "hostUrl": "https://gitpod-staging.com",
        "installationShortname": "eu01",
        "stage": "production",
        "devBranch": "",
        "insecureNoDomain": false,
        "license": "",
        "licenseFile": "",
        "definitelyGpDisabled": false,
        "enableLocalApp": true,
        "builtinAuthProvidersConfigured": true,
        "disableDynamicAuthProviderLogin": false,
        "maxEnvvarPerUserCount": 4048,
        "maxConcurrentPrebuildsPerRef": 10,
        "makeNewUsersAdmin": false,
        "theiaPluginsBucketNameOverride": "",
        "defaultBaseImageRegistryWhitelist": [],
        "runDbDeleter": true,
        "contentServiceAddr": "content-service:8080",
        "imageBuilderAddr": "image-builder-mk3:8080",
[...]

PR #9795, opened 2022-05-05
[installer] Make the `server` mount the Github App secret

andrew-farries:
## Description

This PR adds an an extra volume and volume mount to the `server` component to allow it to mount the github app secret, when a github app is specified in the installer config. The ability to add github app config was added in https://github.com/gitpod-io/gitpod/pull/9297.

## Related Issue(s)
Part of #9097 

## How to test

```yaml
experimental:
  webapp:
      githubApp:
        appId: 123
        authProviderId: 'someAuthProviderId'
        baseUrl: 'someBaseUrl'
        certPath: 'someCertpath'
        enabled: true
        logLevel: 'someLogLevel'
        marketplaceName: 'someMarketPlacename'
        webhookSecret: 'somewebhookSecret'
        certSecretName: 'certSecretName'
```

```
```

Then invoke the installer as:

```
```

The `server` component will have an extra volume and volume mount

## Release Notes

```release-note
Make sure the server mounts the github app secret when an app is specified in the installer
```

## Documentation

None.

### DIFF
--- a/install/installer/pkg/components/server/constants.go
+++ b/install/installer/pkg/components/server/constants.go
@@ -15,6 +15,7 @@ const (
 	authProviderFilePath  = "/gitpod/auth-providers"
 	licenseFilePath       = "/gitpod/license"
 	chargebeeMountPath    = "/chargebee"
+	githubAppCertSecret   = "github-app-cert-secret"
 	PrometheusPort        = 9500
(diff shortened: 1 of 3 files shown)


geropl (approved):
Awesome - looks like this might be the last PR! :tada: :heart_eyes:

Issue #8704, opened 2022-03-10
Investigate reported performance of JetBrains 


loujaybee:
We've had several reports of the JetBrains experience feeling slow, but don't currently have good data on the situation. It seems that the Gateway implementation experience degrades quite significantly when on a low internet connection, we should look into this issue to figure out whether it's something we can improve on our Gitpod infrastructure, or whether it's something to report back to JetBrains. 

We also have the opportunity to leverage API's for the new performance indicator in Gateway

![image](https://user-images.githubusercontent.com/5528307/157661326-df51a299-7197-4b4b-8b8b-622dc8746679.png)

loujaybee:
@iQQBot - Can you add more details here about the performance issues you were having with JetBrains? Any logs, data, insight about your current bandwidth, machine specs etc would all be useful when reporting back to JetBrains.

iQQBot:
I think it now about Gitpod, Jetbrains Gateway itself is not a great experience, I tried to use a wired network to connect to my other server on my LAN, but the experience was still poor

It's hard for me to say what the detailed performance metrics differences are (ping is good ~0-10ms, cpu and memory usage is not high), but the experience does feel much worse compared to the local Jetbranins IDE, so it may be an overall problem, I'm currently using the 2021.3.2 version of Gateway
![image](https://user-images.githubusercontent.com/8299500/160588624-c396cab9-d50a-42b1-b8a2-035f1a28af11.png)

felladrin:
I can say that high latency makes the experience worse. The difference between having ping 60 and 350+ is noticeable.

felladrin:
Another two tests we can do:
- Compare performance on IntelliJ IDEA when opening a large repo (Gitpod) vs a small one (Spring PetClinic).
- Compare two relatively same-size repos where one is written in Java and the other is written in Kotlin, to check if Kotlin language server is the bottleneck.

iQQBot:
Another try! I use my self-hosted Gitpod with low latency, it works like local

felladrin:
> Another try! I use my self-hosted Gitpod with low latency, it works like local

Does it mean that the problems are due to high latency? Or Gitpod's server resources?

buggtb:
I've been seeing some issues, but mine are a little different. My performance is okay because I stuck in a sed line to my dotfiles to fix it :) :

```
sed -i 's/Xmx2048m/Xmx4096m/g' /ide-desktop/backend/plugins/remote-dev-server/bin/launcher.sh
[...]

PR #9796, opened 2022-05-05
jb: java perf monitoring

### LINKED ISSUE #9521: Expose perf details for the current workspace
When users are diagnosing challenges in a workspace, it would be good to expose container-specific performance information to help the user understand what is going on with their workspace and make necessary changes. 

Ideally, this would be exposed as part of `gp cli` as this command can then be consumed via plugins in the various IDE's for IDE specific experience (e.g. the JetBrains integrated performance monitor)

akosyakov:
## Description

See [internal doc](https://www.notion.so/gitpod/Java-Performance-e14d1acd4993410288b95e5687411172) for results of perf testing.

This PR adds tooling which can be used by us or end users to control and monitor resources consumptions of workspaces generally and particularly JB backends:
- adds a supervisor endpoint to fetch resources status (memory and cpu) with respect of cgroup v1
- provide internal CLI command `/.supervisor/supervisor top` to monitor resources status based on endpoint
- contributes env var for each JB product to control max heap size (XMX) via `[product]_XMX`, i.e. `INTELLIJ_XMX=4096M`
- monitor JB backends max/used memory in bytes and report it to prometheus for each workspace pod
- contributes a dev time script `node ./dev/ide/profile-workspace.js` which capture avg/max/90% percentile of memory and CPU usage as well as max/used JB backend memory for a single workspace running on VM

**Out of scope**
- https://github.com/gitpod-io/gitpod/issues/10043
- support of cgroup v2 for resources endpoint [(add a link to new issue here)]
- making it convenient to configure XMX for JB products via .gitpod.yml [[1](https://github.com/gitpod-io/gitpod/issues/8704#issuecomment-1123718534)]
- adding cgroup based stats to JB perf indicator [(add a link to new issue here)]

## Related Issue(s)

Fixes #9521 

## How to test

**Important** only one person can test at the time. So comment below that you are testing and when done remove a comment.

- Start a dev workspace [https://gitpod.io#https://github.com/gitpod-io/gitpod/tree/refs/heads/ak/java_perf](https://gitpod.io/#https://github.com/gitpod-io/gitpod/tree/refs/heads/ak/java_perf)
- Run `./dev/preview/install-k3s-kubeconfig.sh` to configure k8s context.
- Run `./dev/preview/portforward-monitoring-satellite.sh -c harvester` to port forward prometheus API endpoint.
- Run `node ./dev/ide/profile-workspace.js` to start monitoring resource usage of a workspace pod in prev envs.
- Configure prev envs with IntelliJ IDEA: https://ak-java-perf.preview.gitpod-dev.com/preferences
- Start a workspace in prev envs: https://ak-java-perf.preview.gitpod-dev.com/#INTELLIJ_XMX=4096m,JB_MEM_PROFILE=true/https://github.com/eclipse/xtext-core
- Connect with IntelliJ.
- Run `curl http://localhost:22999/_supervisor/v1/status/resources` to check resource status of supervisor.
- Use `/.supervisor/supervisor top` to see resources status of the current workspace via internal CLI
- Capture CPU, memory and heap metrics for the backend from a profile file in the dev workspace: `watch tail -n 1 /workspace/gitpod/dev/ide/perf.log`

## Release Notes

```release-note
NONE
```

## Documentation

- [x] /werft with-vm
- [ ] /werft without-vm


roboquat:

<details>

</details>

components/supervisor-api/status.proto
```diff
@@ -72,6 +72,13 @@ service StatusService {
         };
     }
 
+    // ResourcesStatus provides workspace resources status information.
+    rpc ResourcesStatus(ResourcesStatuRequest) returns (ResourcesStatusResponse) {
```
akosyakov, line 76: @csweichel Could you have a look at new status endpoint please?

In this PR it respects only cgroup v1. I will open an issue to respect cgroup v2 as a follow-up.

akosyakov: I added `top` only to `/.supervisor/supevisor`, not yet to `gp`.

https://github.com/gitpod-io/gitpod/issues/9521 mention though ideally we should have on `gp` too. Would it be alright with you to add it there?  cc @loujaybee

mustard-mh:
I'm testing this PR 🙌

mustard-mh:
Follow up How to test section, works well for me
|curl|top-cli|watch-log|
|-|-|-|
|curl|top-cli| watch-log|

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodManager.kt
```diff
@@ -59,6 +63,35 @@ class GitpodManager : Disposable {
...
     }
 
+    private val memProfileMode = System.getenv("JB_MEM_PROFILE").toBoolean()
+
+    init {
+        val memProfileJob = GlobalScope.launch {
+            if (application.isHeadlessEnvironment || !memProfileMode) {
+                return@launch
+            }
+            Files.newBufferedWriter(Paths.get("/tmp/jb_mem_profile.log")).use { writer ->
```
mustard-mh, line 73: `/tmp` folder is calc as memory for our workspace. If we keet appending content, mem will keep increase, once user open workspace for a long time, it'll be a bad experience. It's more terrible than process mem increase (OOM can kill it)

akosyakov: yeah, overwriting will be better 👍 

I actually would like to move it to prometheus reporting instead, i.e. look up it on supervisor metrics endpoint

Should I do it in this PR?

mustard-mh: - If history is not very useful, we can replace all content
- Or make this log rotate delete?

akosyakov: it was not useful to me, but in grafana dashboard we will be able to see history too.

mustard-mh: If you consider this feature as normal users not reachable, we can go with it👍

mustard-mh:
I'm testing this PR

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodManager.kt
mustard-mh: `prometheus` is good place 👍

mustard-mh:
How to check if `xmx=4096m` works? @akosyakov

akosyakov:
> How to check if xmx=4096m works? @akosyakov

It is a bit tricky:
- in performance indicator click ... and then select main window. It will render UI of backend in another window.
- In main window press double shift and then enable `Show Memory Usage` toggle
- Check max heap size in the status bar of this window.

mustard-mh:
Tested and works well (we don't need expected result in how to test section?), should we explain what `/workspace/gitpod/dev/ide/perf.log` and `/tmp/jb_mem_profile.log` use for?

akosyakov:
Do you mean a link to https://www.notion.so/gitpod/Java-Performance-e14d1acd4993410288b95e5687411172#a24a5c9f95e043edbbfb249eb086a66a?

components/ide/jetbrains/image/status/main.go
```diff
@@ -254,6 +258,20 @@ func remoteDevServerCmd(args []string) *exec.Cmd {
...
+	xmx := os.Getenv(strings.ToUpper(alias) + "_XMX")
+	if xmx == "" {
+		return nil
+	}
+	launcherPath := "/ide-desktop/backend/plugins/remote-dev-server/bin/launcher.sh"
+	content, err := ioutil.ReadFile(launcherPath)
+	if err != nil {
+		return err
+	}
+	newContent := strings.Replace(string(content), "2048m", xmx, 1)
```
mustard-mh, line 271: Should we change `2048m` to `-Xmx2048m`? It's like a magic number here if you not access `launcher.sh` file

mustard-mh: And add this one https://github.com/gitpod-io/gitpod/blob/a4f50fa74f24c6abe1fa04a6a9e1ea6d0fee3c62/components/ide/jetbrains/image/startup.sh#L13

components/supervisor/pkg/supervisor/top.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &api.ResourcesStatusResponse{
+		Memory: memory,
+		Cpu:    cpu,
+	}, nil
+}
+
+func resolveMemoryStatus() (*api.ResourceStatus, error) {
```
mustard-mh, line 36: Maybe we can use `procfs` package?

dev/ide/profile-workspace.js
```diff
@@ -0,0 +1,91 @@
...
+    if (sorted[base + 1] !== undefined) {
+        return sorted[base] + rest * (sorted[base + 1] - sorted[base]);
+    } else {
+        return sorted[base];
+    }
+};
+
+(async () => {
+    let wsPodName;
+    while (!wsPodName) {
```
mustard-mh, line 29: `wsPodName` only setup once. we need to restart script next time to get correct pod name.

You can ignore it since it's not important

mustard-mh (approved):
/hold

Code looks good

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodManager.kt
mustard-mh: Good for me

components/supervisor/pkg/supervisor/top.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	cpu, err := resolveCPUStatus()
+	if err != nil {
+		return nil, err
+	}
+	return &api.ResourcesStatusResponse{
+		Memory: memory,
+		Cpu:    cpu,
+	}, nil
+}
```
csweichel, line 34: We should harmonise this with the way ws-daemon does it.
Either copy over some of that code, or stick it in something like `common-go/cgroup`.

@Furisto @utam0k could you guide here how to proceed?

Furisto: At the moment the code in ws-daemon is not written to be used by other components. Factoring this out would be equivalent to starting our own cgroup library. Considering that the other options like using containerd's cgroup are awkward to use I still think this is the best way forward. This does not need to be done in this PR though.

akosyakov: Could you point me to relevant code please? 🙏 

There also can be slight difference, i.e. I subtract evict-able memory to give a user more usefull data, i.e. simlar to what GCP showing. I am not sure what ws-daemon does.

Furisto: CPU: 
v1; https://github.com/gitpod-io/gitpod/blob/main/components/ws-daemon/pkg/cpulimit/cfs.go
v2: https://github.com/gitpod-io/gitpod/blob/main/components/ws-daemon/pkg/cpulimit/cfs_v2.go

Memory: 
v1: https://github.com/gitpod-io/gitpod/blob/main/components/ws-daemon/pkg/cgroup/plugin_cachereclaim.go
v2: n.a.

IO:
v1: https://github.com/gitpod-io/gitpod/blob/main/components/ws-daemon/pkg/cgroup/plugin_iolimit_v1.go
v2: https://github.com/gitpod-io/gitpod/blob/main/components/ws-daemon/pkg/cgroup/plugin_iolimit_v2.go

akosyakov: > We should harmonise this with the way ws-daemon does it.

After looking at the code I don't think it can be harmonised without refactoring, ws daemon does not have some functions like getting used memory without evictable caches. Instead it compare page caches after trying to reclaim memory. It also has special cases for different state of workspaces which is not applicable inside a workspace.

I'm not sure refactoring will make such sense trying to create a reusable library right now. Code in ws-daemon is not really source of truth, i.e. there should not be a situation when you change one and need to change another.

akosyakov: I filed https://github.com/gitpod-io/gitpod/issues/10043 to look how to refactor it nicely later.

utam0k: @akosyakov Great PR. I have a question do you have a specific plan to support cgroup v2? As far as I know, In the case of cgroup v2, the supervisor may don't see the cgroup value of the pod.
https://github.com/gitpod-io/gitpod/blob/cf2c0aafa9d38973e3b599637e8c7256d8d3701b/components/ws-daemon/pkg/iws/iws.go#L915-L919
I recommend checking to be sure. You can build the gcp image with cgroup v2 by changing this line. And you can use a gcp image you build on workspace-preview.
https://github.com/gitpod-io/gitpod-packer-gcp-image/blob/ef5be5d24542142f39fc2aa6748761d2f2177124/setup.sh#L111-L112

utam0k: You can use `gitpod-k3s-202205162340` I built as a cgroup v2 version image.
on workspace
```console
root@gitpodio-templategolang-twmrukrpk05:/sys/fs/cgroup# cat /sys/fs/cgroup/memory.max 
max
root@gitpodio-templategolang-twmrukrpk05:/sys/fs/cgroup# cat /sys/fs/cgroup/memory.high
max
```
I make sure that the workspace's pod was limited
```console
    resources:
      limits:
        cpu: "2"
        memory: 4Gi
      requests:
        cpu: "1"
        memory: 2Gi
```

utam0k: So, I guess what I am saying is that in cgroup v2 the memory limit is not visible from the supervisor. This could be an issue with cgroup v2 as the JVM probably can't see it either.

akosyakov: @utam0k Is it always like that or it is how we implement it?

I asked already on other channels whether instead of relying on files we can actually ask container runtime endpoints to collect metrics and provide results via workspace kits somehow?

I wonder also why ws-daemon does not use container runtime endpoints but need to read from the disk?

akosyakov: Anyway we discussed with @iQQBot to land this PR as is today and then tomorrow start working on https://github.com/gitpod-io/gitpod/issues/10043 and support of cgroup v2 in supervisor. So I think we will need to research it as a part of it.

components/supervisor-api/status.proto
akosyakov: for now I will stick with internal cli, there are some uncertainties about implementation, so let's explore first

components/supervisor/pkg/supervisor/top.go
utam0k: > I wonder also why ws-daemon does not use container runtime endpoints but need to read from the disk?

Thanks for your feedback! Sorry, I couldn't get what `the disk` is.

akosyakov: > Thanks for your feedback! Sorry, I couldn't get what the disk is.

I mean container runtime has [such API](https://github.com/opencontainers/runc/blob/067aaf8548d78269dcb2c13b856775e27c410f9c/libcontainer/container.go#L94-L95) which can return usage/limit stats for cpu and memory: https://github.com/opencontainers/runc/blob/067aaf8548d78269dcb2c13b856775e27c410f9c/libcontainer/cgroups/stats.go#L158-L161

Supervisor cannot reach this endpoint, so we have to read cgroup from filesystem. I am asking whether ws-daemon can rely on it instead and provide this information to supervisor somehow?

Furisto: The API you have linked would mean to execute runc stat and then parse the output, this is not a service endpoint. We also need to write to the cgroup filesystem not only read from it and we should strive to do so in a runtime agnostic manner. We could ask this information from containerd but if we allow the option to use any CRI compatible runtime in the future (like CRI-O) this would require us to adapt the code again.

akosyakov: ok, sorry i don't really know it and though opencontainers runc is an abstraction over different runtimes and using this API oppositely will make things more stable.

akosyakov: @csweichel Would it be alright to continue as suggested [here](https://github.com/gitpod-io/gitpod/pull/9796#discussion_r874502070)? Could you approve it please if so? 🙏

Furisto (approved):

mustard-mh:
Code looks good, will test it with Prometheus

csweichel (approved):

mustard-mh (approved):
Tested good

|used|max|
|-|-|
|||

Another way to check xmx=4096m in thin client (related https://github.com/gitpod-io/gitpod/pull/9796#issuecomment-1125895833

|Img|
|-|
|image|

mustard-mh:
Feel free to unhold it 👍 @akosyakov

utam0k (approved):

Issue #9262, opened 2022-04-12
Epic: Improve experience using ports with Gitpod


loujaybee:
We will have almost parity of ports features between JetBrains, VS Code and CLI. 

Following the introduction of: 
1. https://github.com/gitpod-io/gitpod/issues/9700
1. https://github.com/gitpod-io/gitpod/issues/10581
1. https://github.com/gitpod-io/gitpod/issues/10835
1. https://github.com/gitpod-io/gitpod/issues/9286
1. https://github.com/gitpod-io/gitpod/issues/11616
1. https://github.com/gitpod-io/gitpod/issues/7408

Beyond those issues, we can look at introducing some "value add" features across all three main clients (JetBrains, VS Code and CLI) for things like grouping, hiding/showing ports) etc. For example:

1. Forwarded / exposed ports experience
1. Bi-directional port management (configuration drift)
1. https://github.com/gitpod-io/gitpod/issues/13777
1. https://github.com/gitpod-io/gitpod/issues/7764
1. https://github.com/gitpod-io/gitpod/issues/10294
1. https://github.com/gitpod-io/gitpod/issues/10275
1. https://github.com/gitpod-io/gitpod/issues/10408
1. https://github.com/gitpod-io/gitpod/issues/6687
1. https://github.com/gitpod-io/gitpod/issues/13069
1. https://github.com/gitpod-io/gitpod/issues/10448
1. https://github.com/gitpod-io/website/issues/2958

**Completed**
1. https://github.com/gitpod-io/gitpod/issues/9908
1. https://github.com/gitpod-io/gitpod/issues/6240
1. https://github.com/gitpod-io/gitpod/issues/10248
1. https://github.com/gitpod-io/gitpod/issues/9930
1. https://github.com/gitpod-io/gitpod/issues/6993
1. https://github.com/gitpod-io/gitpod/issues/10449
1. https://github.com/gitpod-io/gitpod/issues/11154
1. https://github.com/gitpod-io/gitpod/issues/7662

Related internal conversation(s) [[1](https://gitpod.slack.com/archives/C03372UP877/p1649788554906039)][[2](https://www.notion.so/gitpod/IDE-Discovery-Konrad-Lou-Vizlib-596bb89fe94c4d0f970f9dba53243b5e)] and feedback on Twitter [[1](https://twitter.com/mikepb78/status/1483705407075876864?s=20&t=jcSnhuRByuqQNexd53zMsw)]

konne:
[...]

PR #9797, opened 2022-05-05
Fix Supervisor sending empty port names and descriptions

felladrin:
## Description

Currently, supervisor send empty name and description of ports to editor(code) which make tooltip in code not works (see also https://github.com/gitpod-io/gitpod/issues/9262#issuecomment-1100738991).

This PR fix it

## Related Issue(s)

https://github.com/gitpod-io/gitpod/issues/9262#issuecomment-1100738991

## How to test

Note that port with range will work only when ports are opened

#### Init with .gitpod.yml

- Open a repo which contains ports define in `.gitpod.yml` with name and descripts like this one https://github.com/mustard-mh/test/tree/hw/port-name
- Choose VSCode as IDE in preference
- Start workspace with that repo https://vn-fix-emp45629038ff.preview.gitpod-dev.com/#https://github.com/mustard-mh/test/tree/hw/port-name
- Check if port explorer got name and description with port with tooltip (hover)

#### Watch .gitpod.yml

- Change `.gitpod.yml` in your workspace, the tooltip in `port explorer` should reflect to `yml` file

|Image|
|-|
|image|

## Release Notes

```release-note
Fix port name and description not showing up on Gitpod VS Code Extension sidebar.
```

### DIFF
--- a/components/supervisor/pkg/ports/ports.go
+++ b/components/supervisor/pkg/ports/ports.go
@@ -411,6 +411,12 @@ func (pm *Manager) nextState(ctx context.Context) map[uint32]*managedPort {
 
 		var public bool
 		config, kind, exists := pm.configs.Get(mp.LocalhostPort)
+		if exists {
+			mp.Name = config.Name
+			mp.Description = config.Description
+			mp.OnExposed = getOnExposedAction(config, mp.LocalhostPort)
+		}
+
 		configured := exists && kind == PortConfigKind
 		if mp.Exposed || configured {
 			public = mp.Visibility == api.PortVisibility_public
--- a/components/gitpod-protocol/go/gitpod-config-types.go
+++ b/components/gitpod-protocol/go/gitpod-config-types.go
@@ -72,7 +72,10 @@ type Image_object struct {
 // PortsItems
 type PortsItems struct {
 
-	// Port name (deprecated).
+	// A description to identify what is this port used for.
+	Description string `yaml:"description,omitempty"`
+
+	// Port name.
 	Name string `yaml:"name,omitempty"`
 
 	// What to do when a service on this port was detected. 'notify' (default) will show a notification asking the user what to do. 'open-browser' will open a new browser tab. 'open-preview' will open in the preview on the right of the IDE. 'ignore' will do nothing.
@@ -84,7 +87,7 @@ type PortsItems struct {
 	// The protocol to be used. (deprecated)
 	Protocol string `yaml:"protocol,omitempty"`
 
-	// Whether the port visibility should be private or public. 'public' (default) will allow everyone with the port URL to access the port. 'private' will only allow users with workspace access to access the port.
+	// Whether the port visibility should be private or public. 'private' (default) will only allow users with workspace access to access the port. 'public' will allow everyone with the port URL to access the port.
 	Visibility string `yaml:"visibility,omitempty"`
 }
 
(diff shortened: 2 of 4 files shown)


felladrin:
The current code is working, as seen in [here](https://github.com/gitpod-io/gitpod/pull/9802#issuecomment-1119063295). What is wrong is that it's not reading the the .gitpod.yml file as it should.

components/supervisor/pkg/ports/ports.go
```diff
@@ -360,8 +360,6 @@ func (pm *Manager) nextState(ctx context.Context) map[uint32]*managedPort {
 				state[port] = mp
 			}
 			mp.LocalhostPort = port
```
akosyakov, line 362: I think it is important to keep it here, step 2 should be about handling coinfgs

akosyakov: You can add some debug logs and enable them `#SUPERVISOR_DEBUG_ENABLE=true/...` prefix, then you can check workspace logs.

felladrin:
@mustard-mh, just to let you know I'll add more details to the PR description soon, to make it clear what's the problem and what I have already tried to fix it. Hopefully you can help to find the cause ^^
(It stopped working somewhen in the last 3 months.)

iQQBot (approved):
LGTM Thanks @mustard-mh

akosyakov:
Does it work for port ranges too?

mustard-mh:
> Does it work for port ranges too?

I think it will after the new push, waiting for new build to test 👍

mustard-mh:
Ranges will work only when ports are opened. cc @akosyakov 

image

akosyakov:
Hm, there is no name in a view. It is a bug in vscode web. Could you fit there as well? 🙏

mustard-mh:
> Hm, there is no name in a view. It is a bug in vscode web. Could you fit there as well? 🙏

What `name in a view` means?

`1111` should be `range: 1111` in the screen capture?

akosyakov:
but it is another follow up PR

components/supervisor/pkg/ports/ports.go
```diff
@@ -411,6 +411,11 @@ func (pm *Manager) nextState(ctx context.Context) map[uint32]*managedPort {
 
 		var public bool
 		config, kind, exists := pm.configs.Get(mp.LocalhostPort)
+		if exists {
```
akosyakov, line 414: I wonder why it is necessary here, configurations should be handled why earlier above, this section about served state

akosyakov: ah, it is for ranges?

mustard-mh: visibility (port private and public) is handled here https://github.com/gitpod-io/gitpod/blob/3013055d321c0eabff7216541f847d0df53c2c8f/components/supervisor/pkg/ports/ports.go#L430

akosyakov: Do we need to call `mp.OnExposed = getOnExposedAction(config, port)`?

mustard-mh: > Do we need to call mp.OnExposed = getOnExposedAction(config, port)? or it is handled already somewhere too?
added

I tested with `onOpen: open-browser` with ranges before and it works fine, but I don't know where is code. Add to confirm it

akosyakov: thank you, great 👍

akosyakov: maybe a commit in if will be nice, that we are handling configuration for port ranges here

PR #9799, opened 2022-05-05
[installer] Fix server GitHub app cert mount

geropl:
## Related Issue(s)

Fixes a path from https://github.com/gitpod-io/gitpod/pull/9795

## Release Notes

```release-note
NONE
```

### DIFF
--- a/install/installer/pkg/components/server/deployment.go
+++ b/install/installer/pkg/components/server/deployment.go
@@ -7,6 +7,7 @@ package server
 import (
 	"encoding/base64"
 	"fmt"
+	"path"
 
 	"github.com/gitpod-io/gitpod/installer/pkg/cluster"
 
@@ -201,7 +202,7 @@ func deployment(ctx *common.RenderContext) ([]runtime.Object, error) {
 
 			volumeMounts = append(volumeMounts, corev1.VolumeMount{
 				Name:      githubAppCertSecret,
-				MountPath: cfg.WebApp.Server.GithubApp.CertPath,
+				MountPath: path.Dir(cfg.WebApp.Server.GithubApp.CertPath),
 				ReadOnly:  true,
 			})
 		}


install/installer/pkg/components/server/deployment.go
```diff
@@ -201,7 +202,7 @@ func deployment(ctx *common.RenderContext) ([]runtime.Object, error) {
 
 			volumeMounts = append(volumeMounts, corev1.VolumeMount{
 				Name:      githubAppCertSecret,
-				MountPath: cfg.WebApp.Server.GithubApp.CertPath,
+				MountPath: path.Dir(cfg.WebApp.Server.GithubApp.CertPath),
```
easyCZ, line 205: Do we validate the path somewhere? If the path is empty, `path.Dir()` returns `.` which is unlikely what we want here. https://go.dev/play/p/P9N0NNBdfdq

geropl: Thought about that as well, but I could not come up with anything sensible here.

As a reminder, we have a single client for this anyway, and we need to ensure config stability in general at a higher level anyway, so: I don't think it make sense to add tests for properties on the strings be pipe through. :shrug:

easyCZ: For SaaS yes, but if this is also used by self-hosted, chances are someone will misconfigure it. Perhaps just checking that the path is not empty and erroring when it is makes sense here to avoid unexpected behaviour.

geropl: It's[ in the experimental section](https://github.com/gitpod-io/gitpod/blob/main/install/installer/pkg/config/v1/experimental/experimental.go#L114-L140), so [by definition only to be used by us](https://github.com/gitpod-io/gitpod/blob/main/install/installer/pkg/config/v1/experimental/experimental.go#L5-L10).

easyCZ: Ah, makes sense for now then. Thanks for the clarification.

easyCZ (approved):

Issue #9598, opened 2022-04-28
cannot initialize workspace: no backup found; last backup failed: workspace does not exist.


sagor999:
### Bug description

Sometimes workspace can fail during shutdown with this error:
'cannot initialize workspace: no backup found; last backup failed: workspace does not exist.'

This happens because `DisposeWorkspace` can be called multiple times.
First call removes workspace id from the store, and second one will fail.

### Steps to reproduce

it happens randomly. Look in logs for 'workspace does not exist'

Issue #8326, opened 2022-02-19
Prebuilds don't run due to `last backup failed: workspace does not exist.`


mikenikles:
### Bug description

The Prebuild logs display the following console output in the dashboard:

```
Connecting to workspace logs...

last backup failed: workspace does not exist.
```

<details>
  <summary>Screenshot</summary>
  <img src="https://user-images.githubusercontent.com/788827/154807097-4e146d08-41ac-4ea0-983b-289d45e7624b.png" />
</details>

However, the list of Prebuilds for that project lists the above Prebuild as "READY".

<details>
  <summary>Screenshot</summary>
  <img src="https://user-images.githubusercontent.com/788827/154807172-b31afcf8-ca4c-4399-8cf1-b4586dd5be65.png" />
</details>

### Steps to reproduce

1. Create a new repository in GitHub
2. Commit a `.gitpod.yml` file with the following content
    ```yaml
    image: gitpod/workspace-postgres

    tasks:
      - command: echo ""
    ```
3. Connect the project to Gitpod via https://gitpod.io/new
4. Check the Prebuild that was started automatically

### Workspace affected

N/A

### Expected behavior

The Prebuild runs, shows the workspace logs and is reported as READY on the relevant screens in the dashboard.

### Example repository

https://github.com/mikenikles/gitpod-test-postgresql

kylos101:
There are actually two problems here:
1. The prebuild failed, hinting `last backup failed: workspace does not exist`.
2. The prebuild is indicated as Ready, but, no workspace can be started from it. @geropl does this warrant a separate issue, or is it already solved with ongoing prebuild work you have in-flight?

kylos101:
This appears to heavily impact prebuilds, [753 in the last day.](https://cloudlogging.app.goo.gl/zcQEx67QCaMFXjxG7)

We'll want to see if we can recreate the problem (i.e. is it related or not to incremental prebuilds, and socialize next steps).

This is a research issue. The output isn't necessarily a code fix - but it could be - it depends what is found.

geropl:
> @geropl does this warrant a separate issue, or is it already solved with ongoing prebuild work you have in-flight?

@kylos101 That's something we just fixed. First part will roll out today, 2nd end of week (I think).

kylos101:
I am going to close this, in favor of https://github.com/gitpod-io/gitpod/issues/8326 (which is the same issue), and add that back to the webapp inbox.

[...]

PR #9803, opened 2022-05-05
[ws-manager] fix calling dispose multiple times

sagor999:
## Description

Correct fix for sometimes calling dispose multiple times, which can result in incorrect error about backup failed error.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/9598
Fixes https://github.com/gitpod-io/gitpod/issues/8326

## How to test

Start and stop workspaces. They should always shutdown correctly now, previously sometimes they would error out with failed backup error due to race condition.

## Release Notes

```release-note
[ws-manager] fix sometimes workspaces fail with backup not found error
```

### DIFF
--- a/components/ws-manager/pkg/manager/annotations.go
+++ b/components/ws-manager/pkg/manager/annotations.go
@@ -66,6 +66,10 @@ const (
 	// pvcWorkspaceFeatureAnnotation is set on workspaces which are using persistent_volume_claim feature
 	pvcWorkspaceFeatureAnnotation = "gitpod.io/pvcFeature"
 
+	// startedDisposalAnnotation sets to true when finalizeWorkspaceContent is called to prevent finalize from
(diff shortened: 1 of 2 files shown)


princerachit (approved):

csweichel:
@sagor999 why did this even happen? The finalizer map should have prevented that in the first place. Does ws-manager restart?

Also, this really needs tests. There's an `actOnPodEvent` test which should capture this behaviour in a fixture. I'm somewhat surprised that there are not fixture changes.

sagor999:
@csweichel I think it is a race issue.
Finalizer map is cleaned up when `doFinalize` has finished, if after that something triggers `actOnPodEvent`, it will attempt to run finalizer again. Hence adding annotation to make sure nothing will trigger it attempting to run finalizer.

Let me know if there is a better way to fix this.

PR #9805, opened 2022-05-05
[ws-manager] add cluster role needed for volume snapshots content access

sagor999:
## Description

add cluster role needed for volume snapshots content access

## Related Issue(s)

## How to test

Run installer and verify that it created cluster role and cluster role binding for ws-manager

## Release Notes

```release-note
NONE
```

### DIFF
--- a/install/installer/pkg/components/ws-manager/role.go
+++ b/install/installer/pkg/components/ws-manager/role.go
@@ -16,6 +16,32 @@ func role(ctx *common.RenderContext) ([]runtime.Object, error) {
 	labels := common.DefaultLabels(Component)
 
 	return []runtime.Object{
+		&rbacv1.ClusterRole{
+			TypeMeta: common.TypeMetaClusterRole,
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      Component,
+				Namespace: ctx.Namespace,
+				Labels:    labels,
+			},
+			Rules: []rbacv1.PolicyRule{
+				{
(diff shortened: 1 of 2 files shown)


aledbf (approved):

install/installer/pkg/components/ws-manager/role.go
```diff
@@ -16,6 +16,32 @@ func role(ctx *common.RenderContext) ([]runtime.Object, error) {
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      Component,
+				Namespace: ctx.Namespace,
+				Labels:    labels,
+			},
+			Rules: []rbacv1.PolicyRule{
+				{
+					APIGroups: []string{"snapshot.storage.k8s.io"},
+					Resources: []string{
+						"volumesnapshotcontents",
```
jenting, line 30: Question: do we need to grant `volumesnapshotcontents` as well?

sagor999: you mean volumesnapshot? that one is granted as regular role, since it is namespaced object.

jenting: Sorry, typo, I'm asking for the VolumeSnapshotClass.

sagor999: Ah. So far no need. If in the future we will need it, we can always add it (at least that my way of thinking on it).

PR #9807, opened 2022-05-05
[content-service] add prestop hook to extract git status

sagor999:
## Description

add prestop hook to extract git status when workspace is using pvc.
This is needed when workspace is using PVC and ws-daemon cannot access those files directly because:
they are on a dedicated pvc
finalize happens when container is stopped, and at that point pvc has been detached.

So we use prestop hook to do necessary work right before container will receive signal to stop (later prebuilds will extract build log in the same way for example).

## Related Issue(s)

## How to test

You can test this as part of https://github.com/gitpod-io/gitpod/pull/9475
But for this PR, verify that workspace starts and stops normally. This PR shouldn't affect normal workflow (without PVC) and that is what should be tested in this PR.

## Release Notes

```release-note
none
```


components/ws-daemon/pkg/content/service.go
```diff
@@ -240,6 +241,22 @@ func (s *WorkspaceService) InitWorkspace(ctx context.Context, req *api.InitWorks
 		}
 	}
 
+	if req.PersistentVolumeClaim {
+		// create a folder that is used to store data from running prestophook
+		deamonDir := fmt.Sprintf("%s-daemon", req.Id)
+		prestophookDir := filepath.Join(s.config.WorkingArea, deamonDir, "prestophookdata")
+		_, err = exec.CommandContext(ctx, "mkdir", "-p", prestophookDir).CombinedOutput()
```
princerachit, line 248: Perhaps using `os.MkdirAll()` is better to keep consistency.

components/content-service/pkg/layer/provider.go
```diff
@@ -465,15 +465,27 @@ func contentDescriptorToLayer(cdesc []byte) (*Layer, error) {
...
 func contentDescriptorToLayerPVC(cdesc []byte) (*Layer, error) {
-	return layerFromContent(
-		fileInLayer{&tar.Header{Typeflag: tar.TypeDir, Name: "/.workspace", Uid: initializer.GitpodUID, Gid: initializer.GitpodGID, Mode: 0755}, nil},
-		fileInLayer{&tar.Header{Typeflag: tar.TypeDir, Name: "/.workspace/.gitpod", Uid: initializer.GitpodUID, Gid: initializer.GitpodGID, Mode: 0755}, nil},
-		fileInLayer{&tar.Header{Typeflag: tar.TypeReg, Name: "/.workspace/.gitpod/content.json", Uid: initializer.GitpodUID, Gid: initializer.GitpodGID, Mode: 0755, Size: int64(len(cdesc))}, cdesc},
-	)
+	layers := []fileInLayer{
+		{&tar.Header{Typeflag: tar.TypeDir, Name: "/.workspace", Uid: initializer.GitpodUID, Gid: initializer.GitpodGID, Mode: 0755}, nil},
+		{&tar.Header{Typeflag: tar.TypeDir, Name: "/.workspace/.gitpod", Uid: initializer.GitpodUID, Gid: initializer.GitpodGID, Mode: 0755}, nil},
+		{&tar.Header{Typeflag: tar.TypeReg, Name: "/.supervisor/prestophook.sh", Uid: 0, Gid: 0, Mode: 0775, Size: int64(len(prestophookScript))}, []byte(prestophookScript)},
```
princerachit, line 483: Can you please explain why we are setting Uid and Gid to 0 but then permission to 0775?

sagor999: Uid/Gid to 0 to make sure no one else can edit\modify this file from inside the workspace (otherwise you can execute things from ring1).
775 to allow gitpod user to execute it as part of prestop hook.

utam0k:
Nice PR! I have a question if I use the `stash` command, this stash's information will disappear next time?

components/ws-daemon/pkg/content/service.go
sagor999: Fixed.

components/content-service/pkg/git/git.go
```diff
@@ -215,6 +215,54 @@ func (c *Client) Git(ctx context.Context, subcommand string, args ...string) (er
 	return nil
 }
 
+// GitStatusFromFiles same as Status but reads git output from preexisting files that were generated by prestop hook
+func GitStatusFromFiles(ctx context.Context, loc string) (res *Status, err error) {
```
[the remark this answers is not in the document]
sagor999: Tests added.

princerachit:
/hold

I am ok to merge this PR. @sagor999 please remove the hold label when you think it is ok to merge. I have added this as there were other reviewers who left comments.

princerachit (approved):

PR #9808, opened 2022-05-05
[content-service-api] add new snapshot initializer

sagor999:
## Description

This just adds a new from snapshot volume initializer that is used by https://github.com/gitpod-io/gitpod/pull/9475 to add support for initializing workspace from snapshot volume.

## Related Issue(s)

## Release Notes

```release-note
none
```


components/content-service-api/initializer.proto
```diff
@@ -14,6 +14,7 @@ message WorkspaceInitializer {
         CompositeInitializer composite = 5;
         FileDownloadInitializer download = 6;
         FromBackupInitializer backup = 7;
+        FromSnapshotVolumeInitializer snapshot_volume = 8;
```
jenting, line 17: If we rename it as `FromVolumeSnapshotInitializer`, would it better?
Since it aligns to the naming https://github.com/kubernetes-csi/external-snapshotter/tree/master/client/config/crd, WDYT?

jenting (commented):
The rest LGTM

components/content-service-api/initializer.proto
sagor999: You are right, especially that DB entity is correctly named VolumeSnapshot. I will fix.

sagor999: @jenting fixed! PTAL

csweichel:
How would this be used? I.e. who would decide how when to use this initializer? Server can hardly make that decision, because workspace content storage is opaque to webapp.

IMHO we should try and stick to the backup/prebuild/snapshot initializer we have today which rely on a "snapshot URL" (or in case of the backup initializer use their own convention to discover the backup).

sagor999:
@csweichel @geropl can you please review?

geropl:
> How would this be used? I.e. who would decide how when to use this initializer? Server can hardly make that decision, because workspace content storage is opaque to webapp.

> IMHO we should try and stick to the backup/prebuild/snapshot initializer we have today which rely on a "snapshot URL" (or in case of the backup initializer use their own convention to discover the backup).

@sagor999 Would be interested in this as well :point_up: So far I'm not aware WebApp is going to store information about how workspace content is stored; that was handled inside `content-service`. Is that going to change in any of the following PRs? Or is this a workspace-internal change, and just happens to expand the API surface? :thinking:

sagor999:
@csweichel this is used by https://github.com/gitpod-io/gitpod/pull/9475, if you want to see how decision is being made when to use it. 
It will be very cumbersome (I think) to add this into existing initializer (and possibly confusing)

sagor999:
@csweichel @geropl I extracted this from original PR at https://github.com/gitpod-io/gitpod/pull/9475 to make it easier to review those changes piece by piece, but I guess this was not very helpful. You can look at that PR to see how this initializer is used.

sagor999:
@csweichel @geropl just to add, if there is a concern that there are too many initializers, once PVC approach is adopted, FromBackup and Snapshot initializers will be removed, as they will be obsolete.

sagor999:
After a sync with @csweichel decided it will be best to re-use existing initializers so that it is easier to add PVC support into prebuilds later on.

PR #9825, opened 2022-05-06
Ensure workspace can be deleted even if context timed out

Furisto:
## Description
While we are waiting for the workspace to come up, the context can time out which means the removal of the finalizer will fail. Instead use a seperate context for deletion of the workspace.

```
{
insertId: "36gpzslgv8uop3jn"
jsonPayload: {
@type: "type.googleapis.com/google.devtools.clouderrorreporting.v1beta1.ReportedErrorEvent"
error: "context canceled"
instanceId: "<redacted>"
level: "error"
message: "was unable to remove finalizer"
pod.Name: "<redacted>"
pod.Namespace: "default"
serviceContext: {2}
userId: "<redacted>"
workspaceId: "<redacted>"
}
labels: {1}
logName: "projects/workspace-clusters/logs/stderr"
receiveTimestamp: "2022-05-06T08:27:22.832122356Z"
resource: {2}
severity: "ERROR"
timestamp: "2022-05-06T08:27:21Z"
}
```

## Related Issue(s)
n.a.

## How to test
n.a.

## Release Notes

```release-note
Ensure workspace can be deleted even if start workspace timeout has been exceeded
```

### DIFF
--- a/components/ws-manager/pkg/manager/manager.go
+++ b/components/ws-manager/pkg/manager/manager.go
@@ -270,8 +270,10 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
 			// We do below in loop as sometimes the pod might be already deleted or modified when we attempt to remove the finalizer
 			// so we first get the pod and then attempt to remove the finalizer
 			// in case the pod is not found in the first place, we return success
+			rmCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+			defer cancel()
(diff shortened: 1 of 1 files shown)


kylos101:
Hey @Furisto , in regards to how to test, can you think of a way to recreate the scenario you saw, where the workspaces never started properly? If no, would be good to pick the brain of Alejandro or the IDE team. For example, maybe there's an environment variable we can set in our Gitpod settings that will cause the workspace to self-destruct. This way, we can more easily test this code path/change.

PR #9829, opened 2022-05-06
[public-api] Convert server errors to appropriate RPC errors

easyCZ:
## Description

Server errors are returned largely as string errors into the library we're using. We need to parse them somewhat to know what to return to the caller.

In this PR, we first extract info from the server errors and wrap them with a corresponding gRPC status code, we then take this on the GetWorkspace handler to return meaningful user facing errors out, without the underlying protocol errors. In effect, we have `internal` and `public` errors, and the internal details shouldn't be returned the caller.

## Related Issue(s)

## How to test

Unit tests

## Release Notes

```release-note
NONE
```

## Documentation

NONE

### DIFF
--- a/components/public-api-server/pkg/apiv1/workspace_test.go
+++ b/components/public-api-server/pkg/apiv1/workspace_test.go
@@ -6,7 +6,7 @@ package apiv1
 
 import (
 	"context"
-	"fmt"
+	"errors"
 	"github.com/gitpod-io/gitpod/common-go/baseserver"
 	gitpod "github.com/gitpod-io/gitpod/gitpod-protocol"
 	v1 "github.com/gitpod-io/gitpod/public-api/v1"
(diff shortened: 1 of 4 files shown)


components/public-api-server/pkg/apiv1/workspace_test.go
```diff
@@ -109,7 +109,7 @@ func TestWorkspaceService_GetWorkspace(t *testing.T) {
 			resp, err := client.GetWorkspace(ctx, &v1.GetWorkspaceRequest{
 				WorkspaceId: scenario.WorkspaceID,
 			})
-			require.Equal(t, scenario.ErrorCode, status.Code(err), "status code must match")
+			require.Equal(t, scenario.ErrorCode.String(), status.Code(err).String(), "status code must match")
 			if diff := cmp.Diff(scenario.Response, resp, protocmp.Transform()); diff != "" {
```
csweichel, line 113: Existing could would introduce an `Expectation` type and compare that, rather than use `require`.

easyCZ: Unsure if I managed to parse the above correctly. Are you referring to line `if diff := cmp.Diff(scenario.Response, resp, protocmp.Transform()); diff != "" {` or the one above? Github is rendering this against L113.

If this is indeed against line 112, and I understand the above correctly, you're suggesting we do:
```
type Expectation struct {
	Code 			codes.Code
	Response	*v1.GetWorkspaceResponse
}
```
and we would then compare 
```
if diff := cmp.Diff(scenario.Expectation, Expectation{
	Code: status.Code(err),
	Response: resp
})...
```
If this is the case, then the reason for my change to use `.String()` doesn't fully address it. I used `.String()` to get a human readable status code instead of `0x4` or similar hex which is how the status code is represented by grpc.

Would you be able to please clarify what you mean?

csweichel: when using cmp.Diff you'll get a human readable representation of the status code, see https://go.dev/play/p/uv8P_7ZuZ9H

easyCZ: Ah, neat. Thanks!

easyCZ: Fixed.

csweichel (approved):

laushinka (approved):

PR #9830, opened 2022-05-06
Remove intellij to avoid generating large gradle

mustard-mh:
## Description

Right now, Gitpod repo prebuild will generate a .grandle folder with 12G of storage. This PR remove `intellij` in `.gitpod.yml` file to avoid pre-indexing of intellij, to save time of download rebuild and workspace storage

image

## Related Issue(s)

Fixes #

## How to test

1. Start Gitpod with this branch https://gitpod.io/#https://github.com/gitpod-io/gitpod/tree/hw/remove-intellij
2. Exec `du -d 1 -h /workspace`, you'll see something like these below
```
36M     /workspace/.cache
0       /workspace/.cargo
4.0K    /workspace/.config
396K    /workspace/.gitpod
5.9G    /workspace/.gradle
1.4G    /workspace/gitpod
30M     /workspace/go
446M    /workspace/.vscode-remote
7.8G    /workspace
```
Another part of .gradle is gen by https://github.com/gitpod-io/gitpod/blob/main/.gitpod.yml#L38-L42 we have other plan about it, like move JetBrains related to another repo, see [internal chat](https://gitpod.slack.com/archives/C01KGM9BH54/p1651844869166919?thread_ts=1651172017.770269&cid=C01KGM9BH54)

## Release Notes

```release-note
NONE
```

## Documentation

- [x] /werft without-preview=true

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -68,7 +68,3 @@ jetbrains:
   goland:
     prebuilds:
       version: stable
-  intellij:
-    prebuilds:
-      version: stable
-


felladrin:
@akosyakov / @mustard-mh
Should this PR be blocked till the end of https://github.com/gitpod-io/gitpod/issues/9874 investigation?

akosyakov:
I would rather close it for now. Ideally we should keep everything but make it perform.

akosyakov:
Sure we can remove everything, but what is the point if we cannot dogfood our own things and learn from it 🤷‍♂️

akosyakov:
@easyCZ Do you keep using GoLand and benefit from indexing?

easyCZ:
@akosyakov It has been faster to start-up, but not significantly because GoLand seems to re-index on workspace startup

akosyakov:
@easyCZ Do you use stable GoLand? we disabled indexing for latest.

easyCZ:
I'm on EAP, themes and a couple of other things are glitchy for me on Stable.

akosyakov:
@easyCZ new stable was released which is the same as EAP now

PR #9831, opened 2022-05-06
[installer] Allow the use of a static password for the message bus

andrew-farries:
## Description

This PR adds an `experimental.common.staticMessagebusPassword` field to set a message bus password, rather than having the installer render a fresh password on every invocation. cf [this comment](https://github.com/gitpod-io/gitpod/blob/66e60f0609b95af7dc1c35b019bbe51334cdb729/install/installer/pkg/common/render.go#L127) from @MrSimonEmms.

## Related Issue(s)
Part of #9097 

## How to test

```yaml
experimental:
  common:
    staticMessagebusPassword: "some-password"
```

```
```

Then invoke the installer as:

```
```

The message bus password will be the value set in the config. If the `experimental` config section is omitted, or the `staticMessagebusPassword` is set to an empty string, a password will be generated.

## Release Notes

```release-note
Add `staticMessagebusPassword` config flag to use a fixed message bus password in the installer
```

## Documentation

None.

### DIFF
--- a/install/installer/pkg/common/render.go
+++ b/install/installer/pkg/common/render.go
@@ -150,9 +150,18 @@ func (r *RenderContext) generateValues() error {
 	}
 	r.Values.InternalRegistryPassword = internalRegistryPassword
 
-	messageBusPassword, err := RandomString(20)
-	if err != nil {
-		return err
+	messageBusPassword := ""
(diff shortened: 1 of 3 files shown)


geropl (approved):
Nice, thank you for the tests! :orange_heart:

corneliusludmann (approved):

Issue #9659, opened 2022-04-29
[werft]: jobs fail often during "Flushing telemetry before stopping job"


kylos101:
### Bug description

Refer to [this build](https://werft.gitpod-dev.com/job/gitpod-build-main.3082) for an example. The build job fails often here. 

The [preview environment](https://main.staging.gitpod-dev.com/workspaces) for main is alive "now" (9 minutes after the build was done), therefore I'm not sure that this was an actual failure.

I ask because if this is not real breakage, we should fix the werft job, so that it doesn't look failed. Otherwise I feel folks might be less likely to respond to actual main build failures.

### Steps to reproduce

Not sure

### Expected behavior

Depends if this is an actual failure or not.

If its not really a failure, I'd argue it should exit with a zero status, so the job appears to have succeeded.

### Example repository

This one

### Anything else?

Thank yous!

meysholdt:
([mention in Slack](https://gitpod.slack.com/archives/C01KGM9EBD4/p1651250691079339))

mads-hartmann:
@kylos101 Thanks for the report Kyle. I have opened a PR that fixes a few things related to this, please see the PR for details https://github.com/gitpod-io/gitpod/pull/9837

PR #9837, opened 2022-05-06
Improve handling of preview deployment failures on main

mads-hartmann:
## Description

This PR improves how we handle preview deployment failures on main.

Rather than relying on catching errors everywhere and using `exit 0` in the implementation of preview environment deployments, this moves the `try/catch` all the way to the top. This ensure that preview deployment errors never
breaks the main build.

Additionally the this improves a few other things as well

- Using `exit 0` means that all of our shutdown logic like flushing traces doesn't get executed which might result in missing spans. So getting rid of those is nice.
- Always log the error that caused the job to fail. Previously, on the main build, the error would only be sent to Slack, and added to the span but wouldn't be visible in the actual Werft logs. This fixes that.
- Using `werft.fail` will no longer end all spans. We have a `finally` handler that does that anyway, and ending all the spans in the `fail` code means we close the root span before the job is actually done (e.g. our top catch/finally logic hasn't been executed) 

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/9659
Fixes https://github.com/gitpod-io/ops/issues/947

## How to test

I have not tried to inject errors, so we won't know for sure this does the right thing until this is merged and we see a preview environment deployment failure on main.

## Release Notes

```release-note
NONE
```

## Documentation

N/A

### DIFF
--- a/.werft/build.ts
+++ b/.werft/build.ts
@@ -60,6 +60,16 @@ async function run(context: any) {
         return
     }
 
-    await deployToPreviewEnvironment(werft, config)
+    try {
+        await deployToPreviewEnvironment(werft, config)
+    } catch (e) {
+        // We currently don't support concurrent deployments to the same preview environment.
+        // Until we do we don't want errors to mark the main build as failed.
+        if (config.mainBuild) {
+            return
+        }
+        throw e
+    }
+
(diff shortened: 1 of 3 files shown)


meysholdt (approved):
nice to see all the "if(!main" gone :)

Issue #9842, opened 2022-05-07
Delete the Helm chart


meysholdt:
Now that the WebApp and IDE components are deployed using the installer, we no longer need the helm chart. I did take a quick look at what needs to be done to delete it and found:

### Blockers
* [ ] https://github.com/gitpod-io/gitpod/issues/9844
* [ ] https://github.com/gitpod-io/gitpod/issues/9845

### Optional but nice cleanup:
* [ ] https://github.com/gitpod-io/gitpod/issues/9843
* [ ] https://github.com/gitpod-io/gitpod/issues/9846

### The Reward
* [ ] delete /chart/

Issue #9843, opened 2022-05-07
Move db-init-scripts out of Helm chart


meysholdt:
## Is your feature request related to a problem? Please describe
As a preparation to delete the helm chart (https://github.com/gitpod-io/gitpod/issues/9842) move the [chart/config/db/init](https://github.com/gitpod-io/gitpod/tree/main/chart/config/db/init) out of the helm chart because they are still needed elsewhere. 

They are needed by:
* The `db-migrations`:
https://github.com/gitpod-io/gitpod/blob/4e331313ad8787d6218639d94bf6dd4483c6aa7f/components/gitpod-db/BUILD.yaml#L50

* The installer also has references. Not sure if it consumes the scripts directly or just has references into the db-migrations component:
https://github.com/gitpod-io/gitpod/blob/4820e804bb86ecee9548ce9f1f9aca761d8436b9/install/installer/pkg/components/database/init/constants.go#L12
https://github.com/gitpod-io/gitpod/blob/4820e804bb86ecee9548ce9f1f9aca761d8436b9/install/installer/pkg/components/database/incluster/constants.go#L13
## Describe the behaviour you'd like
Remove files related to db-init from the chart/ folder so that the chart/folder can be safely deleted.

geropl:
This has already been done with the helm chart removal PR.

PR #9858, opened 2022-05-09
Remove helm chart

csweichel:
## Description
This PR removes the helm chart. Now that we're fully migrated to the installer, we can finally get rid of that thing.

This PR also (temporarily removes) the `contrib/docker` directory, as that was sitting on the helm chart. @corneliusludmann what should we do here?

## How to test
Use the installer instead :) 

## Release Notes

```release-note
Fully removed the deprecated helm charts.
```

## Documentation

https://github.com/gitpod-io/website/issues/2002


geropl:
@csweichel AFAIK this is blocked by https://github.com/gitpod-io/gitpod/issues/9843.

csweichel:
That's part of this PR - albeit a tad broken. Working on a fix.

meysholdt:
This fixes https://github.com/gitpod-io/gitpod/issues/9842

> This PR also (temporarily removes) the contrib/docker directory, as that was sitting on the helm chart. @corneliusludmann what should we do here?

Also asking that [question here](https://gitpod.slack.com/archives/C01KLC56NP7/p1651914204113709).

corneliusludmann:
`contrig/docker` can be removed entirely. It's sufficient to have this in the Git history.

cc @Pothulapati

Pothulapati:
Agree with @corneliusludmann, I can use it from the git history to build a similar thing with `kots` 👍🏼

aledbf (approved):

corneliusludmann (approved):

geropl:
Sorry for blocking, but I'd love to understand the inti script refactor. I have a time slot at 5PM to do so, and can :heavy_check_mark: then. :+1:

PR #9863, opened 2022-05-09
Laushinka/add metrics for server 9824

### LINKED ISSUE #9824: Add metrics for server connection
We want to keep track of how long it takes to create a connection

laushinka:
## Description
Sets up metric for how long a connection takes.

## Related Issue(s)

Fixes #9824 

## How to test
1. Under `/components/public-api-server` run `go run main.go`
2. Run `curl localhost:9500/metrics`
3. See the custom metrics exposed.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/public-api-server/pkg/proxy/conn.go
+++ b/components/public-api-server/pkg/proxy/conn.go
@@ -10,6 +10,7 @@ import (
 	gitpod "github.com/gitpod-io/gitpod/gitpod-protocol"
 	"github.com/grpc-ecosystem/go-grpc-middleware/logging/logrus/ctxlogrus"
 	"net/url"
+	"time"
 )
 
 type ServerConnectionPool interface {
@@ -26,6 +27,10 @@ type NoConnectionPool struct {
 func (p *NoConnectionPool) Get(ctx context.Context, token string) (gitpod.APIInterface, error) {
 	logger := ctxlogrus.Extract(ctx)
 
+	start := time.Now()
+	defer func() {
+		reportConnectionDuration(time.Since(start))
+	}()
 	server, err := gitpod.ConnectToServer(p.ServerAPI.String(), gitpod.ConnectToServerOpts{
 		Context: ctx,
 		Token:   token,
--- a/components/public-api-server/pkg/proxy/prometheusmetrics.go
+++ b/components/public-api-server/pkg/proxy/prometheusmetrics.go
@@ -1,0 +1,24 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package proxy
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"time"
+)
+
+func reportConnectionDuration(d time.Duration) {
+	proxyConnectionCreateDurationSeconds.Observe(d.Seconds())
+}
+
+var proxyConnectionCreateDurationSeconds = prometheus.NewHistogram(prometheus.HistogramOpts{
+	Namespace: "gitpod",
+	Name:      "public_api_proxy_connection_create_duration_seconds",
+	Help:      "Histogram of connection time in seconds",
+})
+
+func RegisterMetrics(registry *prometheus.Registry) {
+	registry.MustRegister(proxyConnectionCreateDurationSeconds)
+}
--- a/components/public-api-server/pkg/proxy/prometheusmetrics_test.go
+++ b/components/public-api-server/pkg/proxy/prometheusmetrics_test.go
@@ -1,0 +1,42 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package proxy
+
+import (
+	"context"
+	"github.com/gitpod-io/gitpod/common-go/baseserver"
+	v1 "github.com/gitpod-io/gitpod/public-api/v1"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/testutil"
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/grpc"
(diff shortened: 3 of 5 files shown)


components/public-api-server/pkg/proxy/conn.go
```diff
@@ -26,6 +27,10 @@ type NoConnectionPool struct {
 func (p *NoConnectionPool) Get(ctx context.Context, token string) (gitpod.APIInterface, error) {
 	logger := ctxlogrus.Extract(ctx)
 
+	start := time.Now()
+	defer func() {
+		ReportConnectionDuration(time.Since(start))
+	}()
```
andrew-farries, line 33: Nice use of `defer` 👍

andrew-farries (approved):
Looks good. Are the 'how to test' instructions correct though?

I did:

1. `go run main.go` (in `public-api-server`)
2. `curl localhost:9500/metrics`

which works as expected. I don't think the server takes subcommands like `api workspaces get`.

components/public-api-server/pkg/proxy/conn.go
laushinka: Just taught by @easyCZ 🤗

laushinka:
> Looks good. Are the 'how to test' instructions correct though?

Oops forgot a step! Thanks for the catch.

components/public-api-server/pkg/proxy/prometheusmetrics.go
```diff
@@ -0,0 +1,20 @@
...
+// See License-AGPL.txt in the project root for license information.
+
+package proxy
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"time"
+)
+
+func ReportConnectionDuration(d time.Duration) {
```
easyCZ, line 12:
```suggestion
func reportConnectionDuration(d time.Duration) {
```
doesn't need to be public

components/public-api-server/pkg/proxy/prometheusmetrics.go
```diff
@@ -0,0 +1,20 @@
...
+	ConnectionDurationHistogram.Observe(d.Seconds())
+}
+
+var ConnectionDurationHistogram = prometheus.NewHistogram(prometheus.HistogramOpts{
```
easyCZ, line 16:
```suggestion
var proxyConnectionCreateDurationSeconds = prometheus.NewHistogram(prometheus.HistogramOpts{
```
Should reflect the metric name

laushinka:
Thanks for helping review, @andrew-farries 🤗

components/public-api-server/pkg/proxy/prometheusmetrics.go
```diff
@@ -0,0 +1,20 @@
...
+	Namespace: "gitpod",
+	Name:      "grpc_connection_duration_seconds",
```
easyCZ, line 18:
```suggestion
	Name:      "public_api_proxy_connection_create_duration_seconds",
```
Metric should be specific, and should include units where relevant. See https://prometheus.io/docs/practices/naming/

components/public-api-server/pkg/server/server.go
```diff
@@ -37,7 +41,9 @@ func Start(logger *logrus.Entry, cfg Config) error {
 	return nil
 }
 
-func register(srv *baseserver.Server, cfg Config) error {
+func register(srv *baseserver.Server, cfg Config, registry *prometheus.Registry) error {
+	registry.MustRegister(proxy.ConnectionDurationHistogram)
```
easyCZ, line 45: The registry here shouldn't really know about the metric, and once you make it private as suggested, this won't be possible. Move this into the Proxy package, as something like `proxy.RegisterMetrics(registry)`

easyCZ (changes requested):
Nice, left a couple of comments mostly to make the metric private. As a challenge, can you write a test for this? By injecting the registry, you should be able to check that it got incremented after you invoked `Get`

components/public-api-server/pkg/proxy/prometheusmetrics_test.go
```diff
@@ -0,0 +1,38 @@
...
+	client := v1.NewWorkspacesServiceClient(conn)
+
+	// Call GetWorkspace
+	client.GetWorkspace(ctx, &v1.GetWorkspaceRequest{
+		WorkspaceId: "some-ID",
+	})
+
+	count, err := testutil.GatherAndCount(registry)
+	assert.NoError(t, err)
+	assert.Equal(t, 1, count)
```
laushinka, line 37: @easyCZ So this passes, but not sure if this is the right approach?

easyCZ: Nice! Yeah this is good, well done!

components/public-api-server/pkg/server/server.go
```diff
@@ -12,21 +12,25 @@ import (
...
 		baseserver.WithHTTPPort(cfg.HTTPPort),
 		baseserver.WithGRPCPort(cfg.GRPCPort),
+		baseserver.WithMetricsRegistry(registry),
 	)
 	if err != nil {
 		return fmt.Errorf("failed to initialize public api server: %w", err)
 	}
 
-	if registerErr := register(srv, cfg); registerErr != nil {
+	if registerErr := register(srv, cfg, registry); registerErr != nil {
```
easyCZ, line 33: Because of this part being quite not ergonomic, I've raised https://github.com/gitpod-io/gitpod/pull/9879 which allows us to do `srv.MetricsRegistry()` to get access to the registry. It just reduces the need to pass the registry along in a side-channel. We can update later if this PR lands first.

easyCZ (approved):
LGTM. Nice job on the tests!

mustard-mh:
/hold

/werft run

This PR block merge pool

mustard-mh:
License missing https://werft.gitpod-dev.com/job/gitpod-build-laushinka-add-metrics-for-server-9824.16/raw
image

laushinka:
> License missing 

Oops didn't check the job. Thank you, @mustard-mh 🤗

components/public-api-server/pkg/server/server.go
laushinka: Nice!

Issue #8441, opened 2022-02-24
installer: generate documentation for configuration


Pothulapati:
## Is your feature request related to a problem? Please describe

Currently, `Config` is the spec through which users pass configuration into the installer but it isn't
documented. This means users have to check the [golang types directly](https://github.com/gitpod-io/gitpod/blob/main/install/installer/pkg/config/v1/config.go) to understand what
all options are even allowed.

## Describe the behaviour you'd like

We should be generating the documentation directly from the golang structs. Probably, Having them
as comments and rendering them out into a table in `config.md` should be good.

## Additional context

Similar projects around this:

- [Helm Doc Generator from `values.yaml`](https://github.com/norwoodj/helm-docs)
- [Generating Yaml spec from go structs
](https://www.siderolabs.com/blog/documentation-as-code/)

PR #9867, opened 2022-05-09
[installer] Make blocked repositories configurable

andrew-farries:
## Description

This PR makes it possible to configure blocked repositories for the server component:

* Add a `blockedRepositories` field to the server configmap struct to match https://github.com/gitpod-io/gitpod/blob/main/components/server/src/config.ts#L164.
* Add a new `experimental.webapp.server.blockedRepositories` config field to the installer to set the field in the configmap.

## Related Issue(s)
Part of #9097 

## How to test

```yaml
experimental:
  webapp:
    server:
      blockedRepositories:
      - urlRegExp: "https://github.com/some-user/some-bad-repo"
        blockUser: true
      - urlRegExp: "https://github.com/some-other-user/another-bad-repo"
        blockUser: false
```

```
```

Then invoke the installer as:

```
```

The `blockedRepositories` field in the server config map will reflect the value set in the installer config.

## Release Notes

```release-note
Add `disableWorkspaceGarbageCollection` experimental installer config flag
```

## Documentation

None.

### DIFF
--- a/install/installer/pkg/components/server/configmap.go
+++ b/install/installer/pkg/components/server/configmap.go
@@ -98,6 +98,19 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		return nil
 	})
 
+	var blockedRepositories []BlockedRepository
+	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
+		if cfg.WebApp != nil && cfg.WebApp.Server != nil && len(cfg.WebApp.Server.BlockedRepositories) > 0 {
+			for _, repo := range cfg.WebApp.Server.BlockedRepositories {
+				blockedRepositories = append(blockedRepositories, BlockedRepository{
+					UrlRegExp: repo.UrlRegExp,
+					BlockUser: repo.BlockUser,
+				})
+			}
+		}
+		return nil
+	})
+
 	githubApp := GitHubApp{}
 	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
 		if cfg.WebApp != nil && cfg.WebApp.Server != nil && cfg.WebApp.Server.GithubApp != nil {
@@ -161,6 +174,7 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		MaxConcurrentPrebuildsPerRef:      10,
 		IncrementalPrebuilds:              IncrementalPrebuilds{CommitHistory: 100, RepositoryPasslist: []string{}},
 		BlockNewUsers:                     ctx.Config.BlockNewUsers,
+		BlockedRepositories:               blockedRepositories,
 		MakeNewUsersAdmin:                 false,
 		DefaultBaseImageRegistryWhitelist: defaultBaseImageRegistryWhitelist,
 		RunDbDeleter:                      runDbDeleter,
--- a/install/installer/pkg/components/server/configmap.go
+++ b/install/installer/pkg/components/server/configmap.go
@@ -6,6 +6,7 @@ package server
 
 import (
 	"fmt"
+	"regexp"
 
 	"github.com/gitpod-io/gitpod/installer/pkg/common"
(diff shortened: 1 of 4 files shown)


install/installer/pkg/config/v1/experimental/experimental.go
```diff
@@ -132,15 +132,21 @@ type WsManagerBridgeConfig struct {
...
+	WorkspaceDefaults                 WorkspaceDefaults   `json:"workspaceDefaults"`
+	OAuthServer                       OAuthServer         `json:"oauthServer"`
+	Session                           Session             `json:"session"`
+	GithubApp                         *GithubApp          `json:"githubApp"`
+	ChargebeeSecret                   string              `json:"chargebeeSecret"`
+	DisableDynamicAuthProviderLogin   bool                `json:"disableDynamicAuthProviderLogin"`
+	EnableLocalApp                    *bool               `json:"enableLocalApp"`
+	RunDbDeleter                      *bool               `json:"runDbDeleter"`
+	DefaultBaseImageRegistryWhiteList []string            `json:"defaultBaseImageRegistryWhitelist"`
+	BlockedRepositories               []BlockedRepository `json:"blockedRepositories,omitempty"`
```
easyCZ, line 144: [nit] Would it make sense to comment the usage of a particular field? I know it's slightly increasing the scope of the PR here but I believe it would make onboarding new people to the team easier if they could read something like:

```suggestion
    // BlockedRepositories define repositories (by regex) which can be blocked.
    // Repositories are often blocked due to abuse or ...
    // To block a repository foo/bar, specify:
    //  { UrlRegExp: "https://github.com/foo/bar", BlockUser: false }
    // ...
	BlockedRepositories               []BlockedRepository `json:"blockedRepositories,omitempty"`
```

This could also act as a good basis for documentation

install/installer/pkg/components/server/configmap.go
```diff
@@ -90,6 +90,19 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		return nil
 	})
 
+	var blockedRepositories []BlockedRepository
+	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
+		if cfg.WebApp != nil && cfg.WebApp.Server != nil && len(cfg.WebApp.Server.BlockedRepositories) > 0 {
+			for _, repo := range cfg.WebApp.Server.BlockedRepositories {
+				blockedRepositories = append(blockedRepositories, BlockedRepository{
```
easyCZ, line 97: Should we attempt to compile the regex to ensure it's a valid regex?

easyCZ (approved):
LGTM, added a couple of questions. Feel free to land without addressing them but wanted to raise them in this context.
/hold

install/installer/pkg/components/server/configmap.go
andrew-farries: Yes, see [d70c671](https://github.com/gitpod-io/gitpod/pull/9867/commits/d70c671c48fc6d742cf63266de2cee210735bbb0)

install/installer/pkg/config/v1/experimental/experimental.go
andrew-farries: At least in this case, this documentation already exists in the server config:

https://github.com/gitpod-io/gitpod/blob/8ce8f7a5d014b6726245218291ada85f17e9d155/components/server/src/config.ts#L160-L164

and there is a pretty close mapping between fields in this struct and those config fields.

geropl (approved):
LGTM

geropl:
/unhold

Because comments have been adressed above

install/installer/pkg/config/v1/experimental/experimental.go
Pothulapati: Just a heads up, There's #8441 to generate config documentation from the struct itself! Could be done together!

Pothulapati (approved):
:shipit:

PR #9879, opened 2022-05-09
[baseserver] Expose gRPC prometheus metrics

easyCZ:
## Description

Currently, prometheus metrics are not part of our default interceptors used across all gRPC servers. This is because every server handles is slightly differently. As part of ongoing WIP migration of existing gRPC servers onto base-server, it's necessary to be able to expose gRPC metrics with configurable prometheus registry.

This PR ensures that with, or without, a custom `prometheus.Registry`, gRPC metrics are exposed as long as the server serves gRPC traffic.

## Related Issue(s)

Fixes #

## How to test

Unit tests

or

1. Start public-api-server on this branch with `cd components/public-api-server && go run main.go`
2. Hit the server with a CLI `cd dev/gpctl && go run main.go api workspaces get --id 123 --token foo --address localhost:9501 --insecure` - the token here doesn't matter, we just need to hit it even with bad auth
3. Grab the server exposed metrics with `curl localhost:9500/metrics` and observe at least the following metrics: `"grpc_server_handled_total", "grpc_server_handling_seconds", "grpc_server_started_total"`

## Release Notes

```release-note
NONE
```

## Documentation

NONE

### DIFF
--- a/components/common-go/baseserver/server_test.go
+++ b/components/common-go/baseserver/server_test.go
@@ -5,11 +5,16 @@
 package baseserver_test
 
 import (
+	"context"
 	"fmt"
 	"github.com/gitpod-io/gitpod/common-go/baseserver"
 	"github.com/gitpod-io/gitpod/common-go/pprof"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/testutil"
 	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/health/grpc_health_v1"
 	"net/http"
 	"testing"
 )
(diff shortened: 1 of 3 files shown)


components/common-go/baseserver/server_test.go
```diff
@@ -77,3 +82,43 @@ func TestServer_ServesPprof(t *testing.T) {
 	require.NoError(t, err)
 	require.Equalf(t, http.StatusOK, resp.StatusCode, "must serve pprof on %s", pprof.Path)
 }
+
+func TestServer_Metrics_gRPC(t *testing.T) {
```
laushinka, line 86: Thanks for this test and the comments 🔥

laushinka:
Nice! Works as advertised 💪🏼

LGTM

laushinka (approved):

PR #9886, opened 2022-05-10
Change imagePullPolicy to IfNotPresent

jenting:
## Description
Change imagePullPolicy from Always to IfNotPresent.
We should minimize the traffic to pull container image from remote container image registry if it's already present in the cluster node.
This could avoid we hit the rate limit of the remote container image registry.

## Related Issue(s)

N/A

## How to test

N/A

## Release Notes

```release-note
None
```

## Documentation

N/A

### DIFF
--- a/.werft/build.yaml
+++ b/.werft/build.yaml
@@ -73,7 +73,7 @@ pod:
   - name: build
     image: eu.gcr.io/gitpod-core-dev/dev/dev-environment:cw-werft-cred.0
     workingDir: /workspace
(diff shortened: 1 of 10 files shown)


aledbf (approved):

meysholdt (approved):
Thank you! The change should be fine: It will not prevent new image versions from being pulled because new image versions will have a new image tag.

andreafalzetti (approved):

PR #9896, opened 2022-05-10
[public-api] Cleanup registration of metrics

easyCZ:
## Description

With https://github.com/gitpod-io/gitpod/pull/9879 merged, there is always a `prometheus.Registry` available on the server for use, so we can simplify our logic.

## Related Issue(s)

Fixes #

## How to test

Unit passes

## Release Notes

```release-note
NONE
```

## Documentation

NONE


laushinka:
Lol was just about to make the adjustments! #blazingfast

components/public-api-server/pkg/proxy/prometheusmetrics_test.go
```diff
@@ -36,7 +34,7 @@ func TestConnectionCreationWasTracked(t *testing.T) {
 		WorkspaceId: "some-ID",
 	})
 
-	count, err := testutil.GatherAndCount(registry)
+	count, err := testutil.GatherAndCount(srv.MetricsRegistry())
 	assert.NoError(t, err)
 	assert.Equal(t, 1, count)
```
laushinka, line 39: This now fails because it found more. Is it because now we're using the global registry, there are more metrics that went along with it?

easyCZ: That's because the `srv.MetricsRegistry()` has more metrics than just the 1. We can pass an argument to `GatherAndCount` to filter on only the ones we are testing. Will fix.

easyCZ:
To make the tests work, a significant refactor was needed. This feels unnecessary for the sake of testing a metric so will close this PR and cut a new one with just the refactor. We'll remove the test to reduce the need for indirection just for the sake of the test.

PR #9903, opened 2022-05-10
Validate scopes per client

jeanp413:
## Description

Requesting an oauth token with a scope not listed in the scopes list of a client scopes should fail

## How to test

1. Tested directly using gitpod-desktop extension by modifying source code and requesting for an invalid scope

## Release Notes

```release-note
NONE
```


akosyakov:
We need to test that local app and JB Gateway plugin still working with that too. @gitpod-io/engineering-ide Could someone test for regressions please?

geropl (approved):
Code looks good to me, but did _not_ test

/hold because I did not test @jeanp413 Please feel free to merge if you tested already

akosyakov (approved):
checked that all clients can auth

Issue #9874, opened 2022-05-09
investigate how to speed up prebuilds on gitpod repo


akosyakov:
### Bug description

Right now there is a suspicion that gradle projects and Intellij indexing of them takes a lot of time. We should confirm whether it is true by profiling prebuilds somehow and understand why incremental prebuilds simply does not reuse dependencies and caches from last prebuild. We change these projects very rare.

### Steps to reproduce

Run a prebuild on this repo.

felladrin:
I've noticed `components/ide/jetbrains/gateway-plugin:publish` is being triggered every time due to this `publishToJBMarketplace: true` in [WORKSPACE.yaml](https://github.com/gitpod-io/gitpod/blob/a766836727e7e9e0340332947d14a2a32b8dde8d/WORKSPACE.yaml). (Take [this Werft build log](https://werft.gitpod-dev.com/job/gitpod-build-main.3179/raw) as example)

Using GitHub Code Search you can see that `publishToJBMarketplace` is aways `true` which will cause Gateway Plugin to always rebuild:
- https://cs.github.com/gitpod-io/gitpod?q=jbMarketplacePublishTrigger
- https://cs.github.com/gitpod-io/gitpod?q=publishToJBMarketplace

<img width="1183" alt="image" src="https://user-images.githubusercontent.com/418083/167485983-96c86809-7ae9-41e6-b799-55bddb1f3041.png">

<img width="1239" alt="image" src="https://user-images.githubusercontent.com/418083/167486186-e9967823-9ede-4a5a-90d6-560fc012b2d4.png">

Explosion-Scratch:
Yeah prebuilds are taking a really long time for me (+10min). It would be nice to simply not wait for the prebuild but instead see the progress of it in the editor via terminal

akosyakov:
@felladrin Is it the same effect in prebuilds?

mustard-mh:
As you can see in this prebuild job https://hw-gradle-research.preview.gitpod-dev.com/projects/gitpod/c4d70b3b-fc76-4df4-8312-b2eed367b452

|Task|StartTime|EndTime| Cost|
|:-:|:-:|:-:|:-:|
|gradle|Tue 10 May 2022 03:43:35 PM UTC|Tue 10 May 2022 03:45:04 PM UTC| 1:30|
|yarn & build|Tue 10 May 2022 03:43:16 PM UTC|Tue 10 May 2022 03:48:07 PM UTC|5:00|
[...]

PR #9913, opened 2022-05-10
Enable build cache for gradle tasks and run them in parallel when possible, to reduce their duration

felladrin:
## Description

Enable build cache for gradle tasks and run them in parallel when possible, to reduce their duration.

Reference: [Gradle Docs: Enable the Build Cache](https://docs.gradle.org/current/userguide/build_cache.html#sec:build_cache_enable)

## Related Issue(s)

- https://github.com/gitpod-io/gitpod/issues/9874

## How to test

Observe the Prebuild logs on Gitpod Dashboard and compare with previous ones. The Java tasks should be a little bit faster.

## Release Notes

```release-note
NONE
```

## Documentation

None.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -37,9 +37,8 @@ tasks:
       exit 0
   - name: Java
     init: |
-      leeway exec --package components/supervisor-api/java:lib --package components/gitpod-protocol/java:lib -- ./gradlew build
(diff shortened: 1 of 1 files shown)


andreafalzetti (approved):
LGTM

PR #9914, opened 2022-05-10
[public-api] Add network policy to allow connections from proxy

easyCZ:
## Description

Without this, proxy cannot reach the public api over k8s SRV records which are used as a reverse proxy target by caddy

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Documentation

NONE


geropl (approved):
LGTM

PR #9915, opened 2022-05-10
ws-daemon: Remove the cpu bucket limit because we don't use

utam0k:
## Description

Remove the cpu bucket limit because we don't use it. now we are using only `FixedLimiter`. Saves testing and build time

## Related Issue(s)

No

## How to test

* pass the go tests

## Release Notes

```release-note
ws-daemon: Remove the bucket CPULimit
```

## Documentation

No


components/ws-daemon/pkg/cpulimit/cpulimit.go
```diff
@@ -222,79 +222,6 @@ func (f fixedLimiter) Limit(budgetSpent CPUTime) (newLimit Bandwidth) {
 	return f.FixedLimit
 }
 
-// Bucket describes a "pot of CPU time" which can be spent at a particular rate.
-type Bucket struct {
```
utam0k, line 226: This is proof that it has not been used because this part has been erased and the other parts have not been changed.

utam0k:
@csweichel Any reason why you left the BucketLimit behind when you refactored CPULimit?
https://github.com/gitpod-io/gitpod/pull/8036

nandajavarma:
The changes generally looks good! A small heads up to remove the [`dynamicLimits` reference from the installer config for the preview environment](https://github.com/gitpod-io/gitpod/blob/main/.werft/jobs/build/installer/installer.ts#L89).

install/installer/pkg/config/v1/config.go
```diff
@@ -218,11 +217,8 @@ const (
 
 type Resources struct {
 	// todo(sje): add custom validation to corev1.ResourceList
-	Requests      corev1.ResourceList `json:"requests" validate:"required"`
-	Limits        corev1.ResourceList `json:"limits,omitempty"`
-	DynamicLimits *struct {
-		CPU []cpulimit.Bucket // todo(sje): add custom validation
-	} `json:"dynamicLimits,omitempty"`
```
csweichel, line 225: Note: this is backwards incompatible change. We load the config "strictly", i.e. removing this field might break existing configuration. We don't want that.

Even though this field hasn't been used (had no effect), removing it now might break someone's config.

IMHO we should do something like this:
```
DeprecatedDyanmicLimits json.RawMessage `json:"dynamicLimits,omitempty" validate:"deprecated"``
```

and add a config validation function which returns `true` but emits a log message that the field is deprecated and has no effect.

PR #9916, opened 2022-05-10
[bridge] Fix logging

geropl:
## Description
Fixes unnecessary bloated logging.

## Related Issue(s)

Fixes #

## How to test
 - look at ws-manager-bridge, and grep for "done reconciling."
 - note how "allClusters" only contains names

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/ws-manager-bridge/src/bridge-controller.ts
+++ b/components/ws-manager-bridge/src/bridge-controller.ts
@@ -90,7 +90,7 @@ export class BridgeController {
                 const bridge = await this.createAndStartBridge(newCluster);
                 this.bridges.set(newCluster.name, bridge);
             }
-            log.info("done reconciling.", { allClusters: Array.from(allClusters.values()) });
+            log.info("done reconciling.", { allClusters: Array.from(allClusters.keys()) });
         });
     }
 


components/ws-manager-bridge/src/bridge-controller.ts
```diff
@@ -90,7 +90,7 @@ export class BridgeController {
                 const bridge = await this.createAndStartBridge(newCluster);
                 this.bridges.set(newCluster.name, bridge);
             }
-            log.info("done reconciling.", { allClusters: Array.from(allClusters.values()) });
+            log.info("done reconciling.", { allClusters: Array.from(allClusters.keys()) });
```
easyCZ, line 93:
```suggestion
            log.info("Done reconciling.", { allClusters: Array.from(allClusters.keys()) });
```
Logs tend to be nicer when written for humans - with first word capitalized. But up to you, I don't mind :)

easyCZ (approved):
/hold

components/ws-manager-bridge/src/bridge-controller.ts
geropl: > Logs tend to be nicer when written for humans - with first word capitalized

My strong personal preference as well - but as we do not have a guideline for this I don't feel it's relevant. :slightly_smiling_face:

Issue #9898, opened 2022-05-10
Epic: Require a Gitpod project for prebuilds 


jldec:
Now that projects are usable across git providers, Gitpod should limit prebuilds to repos with a project.

- [ ] Check for project before running prebuild
- [ ] Change documentation to explain that prebuilds require a project
- [ ] Post announcement about new prebuild requirement in changelog

gtsiolis:
+1000 for this.

PR #9936, opened 2022-05-11
[prebuilds] no prebuilds for inactive repos

svenefftinge:
## Description
This change prevents prebuilds for repositories that have been inactive (no regular workspace starts) for at least a week.
This would have prevented 98835 unnecessary prebuilds in the last 11 days (vs. 28174 prebuilds on repos without projects that are still active).

## Related Issue(s)

See also my comment here https://github.com/gitpod-io/gitpod/issues/9898#issuecomment-1123787585

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/gitpod-db/src/typeorm/workspace-db-impl.ts
+++ b/components/gitpod-db/src/typeorm/workspace-db-impl.ts
@@ -356,6 +356,22 @@ export abstract class AbstractTypeORMWorkspaceDBImpl implements WorkspaceDB {
         return workspaceRepo.find({ ownerId: userId });
     }
 
+    public async getWorkspaceCountByCloneURL(
+        cloneURL: string,
+        sinceLastDays: number = 7,
+        type: string = "regular",
+    ): Promise<number> {
+        const workspaceRepo = await this.getWorkspaceRepo();
+        const since = new Date();
+        since.setDate(since.getDate() - sinceLastDays);
+        return workspaceRepo
+            .createQueryBuilder("ws")
+            .where('context->"$.repository.cloneUrl" = :cloneURL', { cloneURL })
+            .andWhere("creationTime > :since", { since: since.toISOString() })
+            .andWhere("type = :type", { type })
+            .getCount();
+    }
+
     public async findCurrentInstance(workspaceId: string): Promise<MaybeWorkspaceInstance> {
         const workspaceInstanceRepo = await this.getWorkspaceInstanceRepo();
         const qb = workspaceInstanceRepo
--- a/components/server/ee/src/prebuilds/prebuild-manager.ts
+++ b/components/server/ee/src/prebuilds/prebuild-manager.ts
@@ -195,6 +195,11 @@ export class PrebuildManager {
                 prebuild.error =
                     "Project is inactive. Please start a new workspace for this project to re-enable prebuilds.";
                 await this.workspaceDB.trace({ span }).storePrebuiltWorkspace(prebuild);
+            } else if (!project && (await this.shouldSkipInactiveRepository({ span }, cloneURL))) {
+                prebuild.state = "aborted";
+                prebuild.error =
+                    "Repository is inactive. Please create a project for this repository to re-enable prebuilds.";
+                await this.workspaceDB.trace({ span }).storePrebuiltWorkspace(prebuild);
             } else {
                 span.setTag("starting", true);
                 const projectEnvVars = await projectEnvVarsPromise;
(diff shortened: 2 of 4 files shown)


components/gitpod-db/src/typeorm/workspace-db-impl.ts
```diff
@@ -356,6 +356,22 @@ export abstract class AbstractTypeORMWorkspaceDBImpl implements WorkspaceDB {
...
+    ): Promise<number> {
+        const workspaceRepo = await this.getWorkspaceRepo();
+        const since = new Date();
+        since.setDate(since.getDate() - sinceLastDays);
+        return workspaceRepo
+            .createQueryBuilder("ws")
+            .where('context->"$.repository.cloneUrl" = :cloneURL', { cloneURL })
+            .andWhere("creationTime > :since", { since: since.toISOString() })
+            .andWhere("type = :type", { type })
+            .getCount();
```
jankeromnes, line 372: Wow, many thanks for this clever fix! 👀

My only worry here is, how do we know that this query on all `d_b_workspace` entries won't be too slow/expensive? (Especially since we'll run it on every received webhook, which seems very frequent.)

svenefftinge: This is what EXPLAIN has to say:
Screenshot 2022-05-11 at 17 46 55

jankeromnes: Aha, that totally addresses the performance concern then, right? (Or, do we need to somehow test/qualify this?)

jankeromnes:
Briefly testing this, code looks good, so will approve if it works in preview.

components/server/ee/src/prebuilds/prebuild-manager.ts
```diff
@@ -356,4 +361,18 @@ export class PrebuildManager {
         const inactiveProjectTime = 1000 * 60 * 60 * 24 * 7 * 1; // 1 week
         return now - lastUse > inactiveProjectTime;
     }
+
+    private async shouldSkipInactiveRepository(ctx: TraceContext, cloneURL: string): Promise<boolean> {
+        const span = TraceContext.startSpan("shouldSkipInactiveRepository", ctx);
+        try {
+            return (
+                (await this.workspaceDB
```
AlexTugarev, line 369: Q: Would it make sense to add some caching here? Just in-memory, time based eviction – let's say 1d.

Thinking out loud:
* webhook events can be dense
* checking for a project seems to be a great eviction policy in general

components/server/ee/src/prebuilds/prebuild-manager.ts
```diff
@@ -195,6 +195,11 @@ export class PrebuildManager {
                 prebuild.error =
                     "Project is inactive. Please start a new workspace for this project to re-enable prebuilds.";
                 await this.workspaceDB.trace({ span }).storePrebuiltWorkspace(prebuild);
+            } else if (!project && (await this.shouldSkipInactiveRepository({ span }, cloneURL))) {
+                prebuild.state = "aborted";
+                prebuild.error =
+                    "Repository is inactive. Please create a project for this repository to re-enable prebuilds.";
```
jankeromnes, line 201: Nit: I think this isn't totally accurate (I agree that we want users to create a Project, but the minimal fix is actually just to open a new workspace for that repository).

However, since this error isn't actually shown anywhere in the UI, users can never see it. 😅 So, for something only a Gitpod admin can see while looking into the DB, that's good enough. 💯

components/gitpod-db/src/typeorm/workspace-db-impl.ts
AlexTugarev: @svenefftinge, the number of rows sent is also a perf issue sometimes. 

If we're just checking for `count === 0` down below, we could start out with a `limit`ed version.

jankeromnes (approved):
Works like a charm! 🎆 🎉 🎊 

Many thanks @svenefftinge for yet again shipping a clever small fix much faster than the more complicated solution we had in mind to solve this problem (i.e. enforcing projects for all prebuilds). 💯

Adding a hold in case you'd like to request changes @AlexTugarev.

(Will also briefly take a look at the prod DB to double-check there is a key on the workspace `creationTime`.)

/hold

components/gitpod-db/src/typeorm/workspace-db-impl.ts
svenefftinge: I would expect typeorm to translate to count(*) when using `getCount`. Doesn't it do that?

AlexTugarev: 👍🏻 

https://github.com/typeorm/typeorm/blob/62518ae1226f22b2f230afa615532c92f1544f01/src/query-builder/SelectQueryBuilder.ts#L2849

geropl:
@svenefftinge @AlexTugarev  Honstely when looking at the query, I doubt we should send that to prod.

Iterating all regular workspaces from the last 7 days _for every prebuild request_ is quite expensive, and I'm pretty sure will blow our DB budget - and I'd rather not find out. :slightly_smiling_face:

I don't have time to look into this until 10am, so maybe we exclude this one from the rollout.

geropl:
> Would in-memory caching as @AlexTugarev suggested be acceptable?

Ideally I'd like to use an indexed version of `context.repository.cloneUrl` by adding a separate field + index for it (the MySQL way :roll_eyes: ), even if that means we have to clone the content [on insert/update](https://github.com/gitpod-io/gitpod/blob/680932745a14a6453c5bb102ee595e5d9a107619/components/gitpod-db/src/typeorm/workspace-db-impl.ts#L138-L146).

svenefftinge:
Working on it [here](https://github.com/gitpod-io/gitpod/pull/9976).

PR #9938, opened 2022-05-11
Only end the span if it exists

mads-hartmann:
## Description

If you call werft.fail with a slice ID that doesn't exist we currently produce an unhelpful stack trace. This fixes that so we log a warning to the slice instead.

See example here: https://werft.gitpod-io-dev.com/job/ops-custom-main.1657

## Related Issue(s)

No issue

## How to test

N/A

## Release Notes

```release-note
NONE
```

## Documentation

N/A


meysholdt (approved):
🧹

PR #9939, opened 2022-05-11
[kots]: add storage to preflight checks

### LINKED ISSUE #9933: Storage - check connection, can upload a file to storage

MrSimonEmms:
## Description

Add a storage check to the KOTS preflight checks. This validates that the given configuration is able to create a bucket, upload a file, download a file and then delete the bucket. The bucket names include two randomly generated strings to avoid global name collisions. This checks against Azure, GCP and S3.

In-cluster connections always return as valid

Arguably, it may look a little strange how I've done the test. As the test is for the connection to be valid and we use the commands, I took the view that testing that it's able to fulfil the required requests was a good idea as there's no "are these credentials valid?" command. I'm open to a discussion on a better way if you can think of one

**NB** Azure doesn't test the region because this is defined on the storage account and the access key/secret are generated within the storage account.

## Related Issue(s)

Fixes #9933

## How to test

Install with KOTS - the version to use is `sje-kots-storage-check.10` in `sje/kots-storage-check` channel

## Release Notes

```release-note
[kots]: add storage to preflight checks
```

### DIFF
--- a/components/kots-config-check/storage/entrypoint.sh
+++ b/components/kots-config-check/storage/entrypoint.sh
@@ -1,0 +1,172 @@
+#!/bin/bash
+# Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+# Licensed under the GNU Affero General Public License (AGPL).
+# See License-AGPL.txt in the project root for license information.
+
+set -euo pipefail
+
+STORE_PROVIDER="${1:-""}"
+STORE_LOCATION="${2:-""}"
+AZURE_ACCOUNT_NAME="${3:-""}"
+AZURE_ACCESS_KEY="${4:-""}"
+GCP_PROJECT_ID="${5:-""}"
+GCP_SERVICE_ACCOUNT_KEY="${6:-""}"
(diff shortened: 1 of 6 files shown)


corneliusludmann (commented):
Great!

Putting this on hold for now until we handled my comments in #9759.

components/kots-config-check/storage/entrypoint.sh
```diff
@@ -0,0 +1,152 @@
...
+    "s3/${bucket_name}" || return 1
+
+  echo "Upload a file"
+  mc cp \
+    "${uploaded_file}" \
+    "s3/${bucket_name}/${file_name}" || return 1
+
+  echo "Download a file"
+  mc cp \
+    "s3/${bucket_name}/${file_name}" \
```
corneliusludmann, line 110: We need to consider the static bucket name introduced in https://github.com/gitpod-io/gitpod/pull/10188. This is a blocker. :disappointed:

nandajavarma (approved):
Tested this and the changed look good!

PR #9944, opened 2022-05-11
Remove contrib build

csweichel:
## Description
Fixes our main build

## Release Notes

```release-note
NONE
```


kylos101 (approved):
:wave:  bye to `contrib:all`

`leeway collect | grep contrib` shows `contrib:all` is no longer a package we depend on.

PR #9947, opened 2022-05-11
[installer] Small render improvements

csweichel:
## Description
This PR makes two small improvements to the `render` command:
- `--output-split-files` which produces individual files for each Kubernetes object the installer produces. This is helpful when needing to filter (i.e. remove) some of the generated objects, e.g. `NetworkPolicies`
- `installer render --config -` which reads the config from stdin. This is helpful when using the Docker image to render the config, because it removes the need to faff with volumes to pass in the config.

## How to test
```bash
#!/bin/bash

# prepare config
docker run --rm -it eu.gcr.io/gitpod-core-dev/build/installer:cw-misc-installer.0 init > cfg.yaml
curl -o yq -L https://github.com/mikefarah/yq/releases/download/v4.25.1/yq_linux_amd64 && chmod +x yq
./yq -i '.domain="foo.com"' cfg.yaml

# test output split files
mkdir -p output
cat cfg.yaml | docker run -v $PWD/output:/o --rm -i eu.gcr.io/gitpod-core-dev/build/installer:cw-misc-installer.0 render --config - --output-split-files /o
ls output

# test reading the config from STDIN
cat cfg.yaml | docker run --rm -i eu.gcr.io/gitpod-core-dev/build/installer:cw-misc-installer.0 render --config - > k8s.yaml
```

## Release Notes

```release-note
[installer] Miscellaneous `render` improvements
```

### DIFF
--- a/install/installer/cmd/render.go
+++ b/install/installer/cmd/render.go
@@ -6,6 +6,7 @@ package cmd
 
 import (
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


nandajavarma (approved):
Neat! Tested and everything works as extected.

PR #9959, opened 2022-05-12
Revert "[prebuilds] no prebuilds for inactive repos"

AlexTugarev:
Reverts gitpod-io/gitpod#9936

We're in doubt about perf implications of this change. So temporarily reverting and testing it again to make sure we're not risking an incident.

```release-notes
NONE
```


geropl (approved):
:+1:

PR #9965, opened 2022-05-12
[baseserver] Add rate limits to interceptors by default

easyCZ:
## Description

Out of the box, the server has rate-limiting middleware configured. The actual rate limits are configurable when starting the server. This is needed to be able to port `ws-manager` to baseserver.

## Related Issue(s)

Fixes #

## How to test

Unit tests

## Release Notes

```release-note
NONE
```

## Documentation

NONE

### DIFF
--- a/components/common-go/baseserver/options.go
+++ b/components/common-go/baseserver/options.go
@@ -7,6 +7,7 @@ package baseserver
 import (
 	"context"
 	"fmt"
+	gitpod_grpc "github.com/gitpod-io/gitpod/common-go/grpc"
 	"github.com/gitpod-io/gitpod/common-go/log"
 	"github.com/heptiolabs/healthcheck"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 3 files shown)


components/common-go/baseserver/options.go
```diff
@@ -34,6 +35,8 @@ type config struct {
 	healthHandler healthcheck.Handler
 
 	grpcHealthCheck grpc_health_v1.HealthServer
+
+	rateLimits map[string]gitpod_grpc.RateLimit
 }
 
 func defaultConfig() *config {
```
csweichel, line 42: IMHO we should move to a common config struct. The more things we add through functional options, the more translating between config structs we'll need to do.

easyCZ: I think we'll always need some level of translation. For example, to pass in a `grpc_health_v1.HealthServer` above, we need to do this during instantiation rather than during config. My thinking is that we define a YAML/JSON serializable config, which can be (auto) converted into Options which are then passed directly into the baseserver. This way, we can share standard config but retain the ability to override in code where necessary. 

There is already a reasonable amount of drift in components with respect to how they accept config and how they are configured at runtime (services etc). Firstly unifying the servers through (options) and then normalising these options into config feels like the more iterative and gradual migration

Issue #9923, opened 2022-05-11
Implement GetOwnerToken RPC on WorkspaceService


easyCZ:
Implement Owner Token retrieval.

To do this, we need to do the following:
* [x] Extend the protobuf definitions to add `GetOwnerToken` rpc in [protobuf definitions](https://github.com/gitpod-io/gitpod/blob/main/components/public-api/gitpod/v1/workspaces.proto#L17) - this should be a seperate PR and should include the generated code (see `generate.sh` in the `public-api` component) 
  * (https://github.com/gitpod-io/gitpod/pull/9966)
* [x] Implement a stub version of this RPC on the [WorkspaceService](https://github.com/gitpod-io/gitpod/blob/main/components/public-api-server/pkg/apiv1/workspace.go#L23) - this should return a stub token so we can have something running and test it
  * https://github.com/gitpod-io/gitpod/pull/9970
* [x] Extend `gpctl api workspaces` to add a sub-command `owner-token` which retrieves the owner token from the RPC
  * https://github.com/gitpod-io/gitpod/pull/9983
* [x] Implement GetOwnerToken properly, [see existing proxying to server](https://github.com/gitpod-io/gitpod/blob/main/components/public-api-server/pkg/apiv1/workspace.go#L35) for how to invoke the method against server.
  * https://github.com/gitpod-io/gitpod/pull/10068

Ensure there are tests for each of the above (except 1. which is just a proto definition).

PR #9966, opened 2022-05-12
[public-api] Add a new `getOwnerToken` RPC to the `WorkspaceService`

andrew-farries:
## Description

Add a new `getOwnerToken` RPC to the `WorkspaceService`.

This PR just adds the rpc to the `.proto` file and regenerates the client and server code.

## Related Issue(s)

Part of https://github.com/gitpod-io/gitpod/issues/9923

## Release Notes

```release-note
NONE
```

## Documentation
none


components/public-api/gitpod/v1/workspaces.proto
```diff
@@ -69,6 +72,13 @@ message GetWorkspaceResponse {
     Workspace result = 1;
 }
 
+message GetOwnerTokenRequest {
+    string workspace_id = 1;
+}
+message GetOwnerTokenResponse {
+    string result = 1;
```
easyCZ, line 79: Normally, `result` would be an object to allow for the returned set to be extended. Here it would look like:
```
message WorkspaceOwnerToken {
  string token = 1;
}
```
Or alternatively, I'd rename `result` to just `token`

andrew-farries: Renamed in `eab9f306c`

easyCZ (approved):

PR #9742, opened 2022-05-03
[baseserver] Dedicated debug server

easyCZ:
## Description

Adds a dedicated debug server to baseserver. This splits up the existing HTTP server into 2.

The debug server is always started, but the HTTP server (intended for serving non debug endpoints - ie actual traffic) is only started when configured. This aligns it more closely with existing components we run.

## Related Issue(s)

Fixes #

## How to test

Unit tests

## Release Notes

```release-note
NONE
```

## Documentation

NONE

/hold

PR #9973, opened 2022-05-12
[content-service] Port to use baseserver

easyCZ:
## Description

Refactors content service to use baseserver for server setup. Removes scattered config and allows us to normalize server setup across components.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Documentation

NONE

/uncc
/hold


components/content-service/cmd/run.go
```diff
@@ -32,106 +19,63 @@ var runCmd = &cobra.Command{
...
+		port, err := cfg.Service.Port()
+		if err != nil {
+			log.WithError(err).Fatal("Failed to get service port.")
+		}
 
-		grpcMetrics := grpc_prometheus.NewServerMetrics()
-		grpcMetrics.EnableHandlingTimeHistogram()
-		reg.MustRegister(grpcMetrics)
+		opts := []baseserver.Option{
+			baseserver.WithGRPCPort(port),
```
csweichel, line 29: Would this start the prometheus/debug endpoint, too?

easyCZ: Once https://github.com/gitpod-io/gitpod/pull/9742 lands, then every base-server will always have metrics, pprof, health and ready endpoints. We'll need to do some port standardization as well to drive more consistency but with that, this will have it out of the box so no need for custom config.

easyCZ:
From my testing on preview, this looks to be working but I'm lacking context on some flows which use the content service, so would appreciate help/direction on what needs to be validated.

components/content-service-api/go/config/config.go
```diff
@@ -113,48 +107,11 @@ type MinIOConfig struct {
...
 
 type ServiceConfig struct {
-	// Daemon  daemon.Config `json:"daemon"`
-	Service    Service       `json:"service"`
-	Prometheus Prometheus    `json:"prometheus"`
-	PProf      PProf         `json:"pprof"`
-	Storage    StorageConfig `json:"storage"`
-}
-
-type TLSConfig struct {
```
easyCZ, line 137: TLS config is now part of `baseserver` so using it directly.

install/installer/pkg/components/content-service/configmap.go
```diff
@@ -17,14 +18,8 @@ import (
 
 func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 	cscfg := config.ServiceConfig{
-		Service: config.Service{
-			Addr: fmt.Sprintf(":%d", RPCPort),
-		},
-		Prometheus: config.Prometheus{
-			Addr: fmt.Sprintf(":%d", PrometheusPort),
-		},
-		PProf: config.PProf{
```
easyCZ, line 26: Wasn't used anywhere besides as a value passed into the server as config.

install/installer/pkg/components/content-service/configmap.go
```diff
@@ -17,14 +18,8 @@ import (
 
 func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 	cscfg := config.ServiceConfig{
-		Service: config.Service{
-			Addr: fmt.Sprintf(":%d", RPCPort),
-		},
-		Prometheus: config.Prometheus{
```
easyCZ, line 23: There's a default metrics port for each baseserver so using that instead.

components/content-service/cmd/run.go
```diff
@@ -32,106 +19,48 @@ var runCmd = &cobra.Command{
...
-			log.Warn("no TLS configured - gRPC server will be unsecured")
+			log.WithError(err).Fatal("Failed to create server.")
 		}
 
-		server := grpc.NewServer(grpcOpts...)
-
 		contentService, err := service.NewContentService(cfg.Storage)
 		if err != nil {
-			log.WithError(err).Fatalf("cannot create content service")
+			log.WithError(err).Fatalf("Cannot create content service")
```
csweichel, line 32: please leave this lowercase. The vast majority of our error messages is lowercase, and we should keep things consistent

csweichel (approved):
Sans nit, lgtm

/hold

components/content-service/cmd/run.go
easyCZ: General practice in Go (but elsewhere) is to uppercase logs to make them sentences. They are designed to be consumed by humans and first letter uppercase is slightly easier to visually grep. Even our style guide uses first letter uppercase in logs https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

The change here is really just a drive by improvement to be more consistent with general practice.

Errors, on the other hand, should never start with an uppercase because they get wrapped and composed so don't really form sentences.

easyCZ: Here's another reference from [go code review style guide](https://github.com/golang/go/wiki/CodeReviewComments#error-strings)

csweichel: Fair point.

/hold cancel

mustard-mh (approved):
lgtm

sagor999 (approved):

PR #9976, opened 2022-05-12
[prebuilds] no prebuilds for inactive repos

svenefftinge:


roboquat:

<details>

</details>

components/gitpod-db/src/typeorm/migration/1652365883273-CloneUrlIndexed.ts
```diff
@@ -0,0 +1,33 @@
...
+        const COLUMN_NAME = "cloneURL";
+        const TYPE_INDEX_NAME = "d_b_workspace_cloneURL_idx";
+
+        if (!(await columnExists(queryRunner, TABLE_NAME, COLUMN_NAME))) {
+            await queryRunner.query(
+                `ALTER TABLE
+                    ${TABLE_NAME}
+                ADD COLUMN
+                    ${COLUMN_NAME} VARCHAR(256)
+                GENERATED ALWAYS AS (
```
AlexTugarev, line 22: Nice! 😎 

Just read on it in https://dev.mysql.com/blog-archive/indexing-json-documents-via-virtual-columns/ and was wondering if `Stored Generated Columns – 3)` may apply here as well.

> [...] and you anticipate large amounts of scanning

That definitely applies here.

components/gitpod-db/src/typeorm/migration/1652365883273-CloneUrlIndexed.ts
```diff
@@ -0,0 +1,33 @@
...
+                    context ->> "$.repository.cloneUrl"
```
geropl, line 23: Sorry @svenefftinge :cry:  this does not work: we learned the hard way that indexes on generated columns do not work :disappointed:

geropl (changes requested):
https://github.com/gitpod-io/gitpod/pull/9976/files#r872044707

components/gitpod-db/src/typeorm/migration/1652365883273-CloneUrlIndexed.ts
geropl: They basically appear to work, but queries on them return inconsistent results.

svenefftinge: oh, interesting. Where did we learn that?

geropl: `workspaceinstance.phase`. IMO we still have that column, but don't use it for that reason. We have phasePersisted instead.
To be clear: the field itself works. But building an index based on that does return the weird results. Maybe we overlooked options back then, so might be worth investigating a bit.

svenefftinge: > indexes on generated columns do not work 

I couldn't find any sign on the internet that this doesn't work, but multiple articles recommending the approach [[1](https://dev.mysql.com/doc/refman/8.0/en/create-table-secondary-indexes.html)], [[2](https://www.compose.com/articles/mysql-for-json-generated-columns-and-indexing/)], [[3](https://medium.com/1mgofficial/mysql-json-support-virtual-columns-and-indexing-json-31df4cc1aa31)], [[4](https://stackoverflow.com/questions/68118107/how-to-create-a-virtual-column-to-index-json-column-in-mysql)].

geropl: @svenefftinge We can test by writing queries against `phase`. i can do that later and share a loom :slightly_smiling_face:

svenefftinge: I tried the following which didn't reveal any discrepancies:
```
SELECT status->>'$.phase' as phase1, phase, count(id) from d_b_workspace_instance where creationTime>'2022-05-13' group by phase1, phase
```

AlexTugarev: We decided to not use the GENERATED COLUMN of mysql, but rather use the `phasePersisted` way, i.e. we will have a  a redundant `cloneUrl` column which is updated on save in typeorm. This is a rather boring solution, but we can move forward with our goals.

AlexTugarev:
I've added https://github.com/gitpod-io/gitpod/pull/9976/commits/ce6b91822b0864e67d3c9de71aae903d6161253c to decouple deployment from enablement. That means, we can migrate as much entries as needed and then reconfigure server  to enable this feature.

jankeromnes:
Thanks for the review request! And for adding the config flag.

I'm a bit worried now, because this Pull Request is growing more and more in complexity. At first it was a quick fix to disable prebuilds for inactive repo URLs, but now we have:
- some additional logic with a 255 character hash index (how well was this tested? has anyone calculated the collision risk?)
- a migration on `d_b_workspace` (that's likely going to take a long time and a lot of DB CPU, right?)
- a configuration change (which can lead to more bugs / ops incidents / headaches)

This is starting to look a bit less like a skateboard. 😕

How about, instead, we purely and simply disable all prebuilds that do not have a project? That's something we want to do anyway soon, and it resolves the same problem in a very simple, risk-free way.

svenefftinge:
> some additional logic with a 255 character hash index (how well was this tested? has anyone calculated the collision risk?)

It doesn't add complexity as it is well encapsulated in a single function and the max length we had in the past 3 months is 159.

> a migration on d_b_workspace (that's likely going to take a long time and a lot of DB CPU, right?)

Why is that migration problematic or taking a long time? It's adding an empty column.

> a configuration change (which can lead to more bugs / ops incidents / headaches)

I agree that is a bit over the top as this was intended an intermediate fix, but really no reason to not move this forward.

> How about, instead, we purely and simply disable all prebuilds that do not have a project? That's something we want to do anyway soon, and it resolves the same problem in a very simple, risk-free way.

There are way too many important active projects to "simply" do this without notice.

Can we please get this in now and move on?

svenefftinge:
> I've added https://github.com/gitpod-io/gitpod/commit/ce6b91822b0864e67d3c9de71aae903d6161253c to decouple deployment from enablement. That means, we can migrate as much entries as needed and then reconfigure server to enable this feature.

There is no need to be flexible about the activity time. 
We could instead update the last 7 days of regular workspaces within the migration.

jankeromnes:
Okay, giving this a closer look and a vigorous shake (i.e. testing it).

jankeromnes (approved):
Code is a bit more complicated but still looks correct. Also, I tested this a lot, and it still works as advertised.

It's nice to be able to disable this by setting `inactivityPeriodForRepos` to `0` in the server configmap (also tested this, works).

So, approving, but I'm very much looking forward to us simply disabling prebuilds for non-projects (and deleting the `/#prebuild/` manual prefix and all of its problems).

Thanks again! Ad astra 🚀

PR #9999, opened 2022-05-13
[baseserver] Support common config struct

csweichel:
## Description
This PR refactors baseserver so that it's easy to integrate into the existing components, all the while harmonising the RPC/debug/HTTP configuration surface.

There's a stack of follow-up PRs which demonstrate the use of these changes in the real world:
- https://github.com/gitpod-io/gitpod/pull/10005
- https://github.com/gitpod-io/gitpod/pull/10007

## How to test
See unit tests

## Release Notes

```release-note
NONE
```

/werft no-preview

### DIFF
--- a/components/common-go/baseserver/config.go
+++ b/components/common-go/baseserver/config.go
@@ -27,7 +27,7 @@ func (s *ServerConfiguration) GetAddress() string {
 }
 
 type TLSConfiguration struct {
-	CA   string `json:"ca" yaml:"ca"`
-	Cert string `json:"cert" yaml:"cert"`
-	Key  string `json:"key" yaml:"key"`
+	CAPath   string `json:"caPath" yaml:"caPath"`
+	CertPath string `json:"certPath" yaml:"certPath"`
+	KeyPath  string `json:"keyPath" yaml:"keyPath"`
 }
--- a/components/common-go/baseserver/config.go
+++ b/components/common-go/baseserver/config.go
@@ -9,9 +9,8 @@ type Configuration struct {
 }
 
 type ServicesConfiguration struct {
-	Debug *ServerConfiguration `json:"debug,omitempty" yaml:"debug,omitempty"`
-	GRPC  *ServerConfiguration `json:"grpc,omitempty" yaml:"grpc,omitempty"`
-	HTTP  *ServerConfiguration `json:"http,omitempty" yaml:"http,omitempty"`
+	GRPC *ServerConfiguration `json:"grpc,omitempty" yaml:"grpc,omitempty"`
+	HTTP *ServerConfiguration `json:"http,omitempty" yaml:"http,omitempty"`
 }
 
 type ServerConfiguration struct {
--- a/components/common-go/baseserver/config.go
+++ b/components/common-go/baseserver/config.go
@@ -1,0 +1,34 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package baseserver
+
+type Configuration struct {
+	Services ServicesConfiguration `json:"services" yaml:"services"`
+}
+
+type ServicesConfiguration struct {
+	Debug *ServerConfiguration `json:"debug,omitempty" yaml:"debug,omitempty"`
+	GRPC  *ServerConfiguration `json:"grpc,omitempty" yaml:"grpc,omitempty"`
+	HTTP  *ServerConfiguration `json:"http,omitempty" yaml:"http,omitempty"`
+}
+
+type ServerConfiguration struct {
+	Address string            `json:"address" yaml:"address"`
+	TLS     *TLSConfiguration `json:"tls" yaml:"tls"`
+}
+
+// GetAddress returns the configured address or an empty string of s is nil
+func (s *ServerConfiguration) GetAddress() string {
+	if s == nil {
+		return ""
+	}
+	return s.Address
+}
+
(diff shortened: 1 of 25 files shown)


MrSimonEmms (approved):

sagor999 (approved):

jenting (commented):
/lgtm

components/common-go/baseserver/options.go
```diff
@@ -7,29 +7,26 @@ package baseserver
...
-	// grpcPort is the port we listen on for gRPC traffic
-	grpcPort int
-	// httpPort is the port we listen on for HTTP traffic
-	httpPort int
+	config *Configuration
 
 	// closeTimeout is the amount we allow for the server to shut down cleanly
 	closeTimeout time.Duration
 
+	noBuiltinServices bool
```
easyCZ, line 27: Looks unused.

components/common-go/baseserver/options.go
```diff
@@ -38,111 +35,118 @@ type config struct {
...
+		opts.config = config
 		return nil
 	}
 }
 
-// WithHTTPPort sets the port to use for an HTTP server. Setting WithHTTPPort also enables an HTTP server on the baseserver.
-func WithHTTPPort(port int) Option {
-	return func(cfg *config) error {
-		cfg.httpPort = port
+func WithoutBuiltinServices() Option {
```
easyCZ, line 65: IMO, this feel like it shouldn't really be an option. I think that all of our servers should have a debug server running and until we've come across a use-case where we definitely do need it, I'd not even offer the option.

components/common-go/baseserver/config.go
```diff
@@ -0,0 +1,33 @@
...
+		return ""
+	}
+	return s.Address
+}
+
+type TLSConfiguration struct {
+	CA   string `json:"ca" yaml:"ca"`
+	Cert string `json:"cert" yaml:"cert"`
+	Key  string `json:"key" yaml:"key"`
+}
```
easyCZ, line 33: I'd personally add the `Path` suffix to these to indicate we want to consume them from files, not as data directly.

components/common-go/baseserver/testing.go
```diff
@@ -35,6 +35,33 @@ func NewForTests(t *testing.T, opts ...Option) *Server {
...
 
+func MustUseRandomLocalAddress(t *testing.T) *ServerConfiguration {
+	t.Helper()
+
+	return &ServerConfiguration{
+		Address: fmt.Sprintf("localhost:%d", MustFindFreePort(t)),
+	}
+}
+
+func MustFindFreePort(t *testing.T) int {
```
easyCZ, line 46: I think this is possibly racy. We're binding a socket to find out it's free, but then we release it immediately and again re-bind during server startup. During tests, we can have a number of servers running concurrently and do not have guarantees in which order they will execute.

I think here we should either make it so we always require an explicit port and never assign one automatically, it then becomes the Test's problem to find an available port, or support binding the socket only once during server startup.

components/common-go/baseserver/server.go
```diff
@@ -340,18 +269,121 @@ func (s *Server) initializeGRPC() error {
...
+		return s.Health.Serve(l)
+	})
+	return eg.Wait()
 }
 
-func (s *Server) isHTTPEnabled() bool {
-	return s.cfg.httpPort >= 0
+func (s *builtinServices) Close() error {
+	var eg errgroup.Group
+	eg.Go(func() error { return s.Debug.Close() })
```
easyCZ, line 385: Neat! This does however not ensure that the Debug server is closed last. In general, it's beneficial to ensure it's last as it allows you to get the most metrics out of the system before it shuts down, and that way it also does contain metrics from the http/grpc servers shutting down.

easyCZ (commented):
Nice, makes sense. Left a couple of comments.

In general, this PR could've been split over multiple PRs to make the review more digestible.

components/common-go/baseserver/options.go
csweichel: Agreed - it's only used for testing. I'd be happy to remove the option and just make the tests set the flag directly.

csweichel: Agreed - will remove

components/common-go/baseserver/testing.go
csweichel: This is indeed racy. Short of exposing the listener through API I fret there's no good way to not make it racy. 

> I think here we should either make it so we always require an explicit port and never assign one automatically

That's what's implemented now. In addition there's the helper func to try and find a free port. I'd vote for going with this for - if the raciness proves to be a problem in practice, we extend the API so that tests can hand in a bound listener.

csweichel:
I've addressed the comments, and rebased the stacked PRs.
https://github.com/gitpod-io/gitpod/pull/10005 is ready for review now.

components/common-go/baseserver/testing.go
easyCZ: Let's see how this goes. I think testing with a baseserver is already an improvement compared to what we have today for most things. We can always follow-up on this.

install/installer/pkg/components/public-api-server/service.go
```diff
@@ -10,10 +10,6 @@ import (
 
 func service(ctx *common.RenderContext) ([]runtime.Object, error) {
 	return common.GenerateService(Component, map[string]common.ServicePort{
-		DebugPortName: {
```
easyCZ, line 13: Thanks for removing this. It was the next for me to clean up after I landed the "debug/built-in server" last Friday.

easyCZ (approved):

PR #10009, opened 2022-05-13
[public-api-server] Harmonise configuration structure

csweichel:
## Description
This PR harmonises the structure of the `public-api-server` so that
- changes to its configuration require a change in the installer: before, we relied solely on unit tests on command line flags. Now, in addition, we rely on structural changes, too.
- it looks and feels a lot more like existing components, which helps folks navigate the codebase

## How to test
- ensure the public-api-server still works as intended
- Run ` go run main.go api workspaces get --id 123 --token foo --address api.cw-api-cfg.preview.gitpod-dev.com:443` to validate API is reachable - you will get unathorized due to token being dummy but this is the expected response and the service works.

## Release Notes

```release-note
NONE
```


install/installer/pkg/components/public-api-server/deployment.go
```diff
@@ -43,64 +57,76 @@ func deployment(ctx *common.RenderContext) ([]runtime.Object, error) {
...
-							ImagePullPolicy: corev1.PullIfNotPresent,
-							Resources: common.ResourceRequirements(ctx, Component, Component, corev1.ResourceRequirements{
-								Requests: corev1.ResourceList{
-									"cpu":    resource.MustParse("100m"),
-									"memory": resource.MustParse("32Mi"),
+						Containers: []corev1.Container{
+							{
+								Name:            Component,
+								Image:           ctx.ImageName(ctx.Config.Repository, Component, ctx.VersionManifest.Components.PublicAPIServer.Version),
+								Args:            []string{"run", "-j"},
```
easyCZ, line 64:
```suggestion
								Args:            []string{"run", "--json-log true"},
```

components/public-api-server/cmd/root.go
```diff
@@ -0,0 +1,71 @@
...
+// This is called by main.main(). It only needs to happen once to the rootCmd.
+func Execute() {
+	if err := rootCmd.Execute(); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+}
+
+func init() {
+	rootCmd.PersistentFlags().StringVar(&rootOpts.CfgFile, "config", "", "config file")
```
easyCZ, line 51: IMO, this shouldn't be a persistent flag on the root command as it doesn't really have any meaning for the root command. It's only relevant for the `run` sub-command. Due to this, it also prints somewhat misleading command help text:
```
go run main.go    
serves public API services

Usage:
  public-api-server [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  help        Help about any command
  run         starts the service

Flags:
      --config string   config file
  -h, --help            help for public-api-server
  -j, --json-log        produce JSON log output on verbose level (default true)
  -v, --verbose         Enable verbose JSON logging

Use "public-api-server [command] --help" for more information about a command.
```

components/public-api/go/config/config.go
```diff
@@ -2,13 +2,10 @@
...
+package config
 
-import "net/url"
-
-type Config struct {
-	GitpodAPI *url.URL
-
-	GRPCPort  int
-	DebugPort int
+type Configuration struct {
```
easyCZ, line 7: 👍

install/installer/pkg/components/public-api-server/configmap.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/gitpod-io/gitpod/installer/pkg/common"
+	"github.com/gitpod-io/gitpod/public-api/config"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
```
easyCZ, line 18: Please add a test that it renders the object under the `config.json` key in the configmap.

easyCZ (commented):
Nice, thanks for the change. Left a couple of comments.

This PR also changes the dev x principle I've been aiming for - starting it locally should work out of the box without needing you to specify config. Currently, a working config can be hard to discover and you'd have to hunt against a particular env configuration.

To this end, I'd like to propose that we include a `config.json` in the component directory which by default talks to `main.preview` or `staging` such that I can just `go run ..` and get something running. This would also be the default value of the `--config` flag.

IMO, this is a necessary step to incrementally move us away from reliance of telepresence and be able to iterate on components without needing to understand the components in a lot of depth.

csweichel:
> To this end, I'd like to propose that we include a `config.json` in the component directory which by default talks to `main.preview` or `staging` such that I can just `go run ..` and get something running. This would also be the default value of the `--config` flag.

Other components solve this by maintaining an `example-config.json` in their directory. We could do the same thing here. Only caveat: those example configs tend to rot quickly.

components/public-api-server/cmd/root.go
```diff
@@ -0,0 +1,71 @@
...
+	rootCmd.PersistentFlags().BoolVarP(&rootOpts.Verbose, "verbose", "v", false, "Enable verbose JSON logging")
+}
+
+func getConfig() *config.Configuration {
+	ctnt, err := os.ReadFile(rootOpts.CfgFile)
+	if err != nil {
+		log.WithError(err).Fatal("cannot read configuration. Maybe missing --config?")
+	}
+
+	var cfg config.Configuration
```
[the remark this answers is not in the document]
csweichel: Viper does a lot of things, and early iterations of our services did use it. 
However, viper is built for CLIs and most of its defaults/functionality don't make sense in a container context. They add complexity and additional behaviour that make services harder to debug.

components/public-api-server/cmd/root.go
```diff
@@ -0,0 +1,71 @@
...
+	if err := rootCmd.Execute(); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+}
+
+func init() {
+	rootCmd.PersistentFlags().StringVar(&rootOpts.CfgFile, "config", "", "config file")
+	rootCmd.PersistentFlags().BoolVarP(&rootOpts.JsonLog, "json-log", "j", true, "produce JSON log output on verbose level")
+	rootCmd.PersistentFlags().BoolVarP(&rootOpts.Verbose, "verbose", "v", false, "Enable verbose JSON logging")
```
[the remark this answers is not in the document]
csweichel: > we want this to work nicely for self-hosted and if one of our customers sees --j or --v it's a lot less informative 

Agreed - will remove.

> Additionally, I'd change json-log to log-format and supply the value of json into it. 

That makes 💯 sense once we introduce additional logging formats. For now I'd prefer to keep this flag as is for the sake of consistency with the other services.

easyCZ:
> Only caveat: those example configs tend to rot quickly.

That's true if you continue to use telepresence. But without it, it becomes the only way you can develop locally so there's direct motivation to maintain it.

corneliusludmann (approved):
Self-hosted related changes (`install/installer/BUILD.yaml` and `install/installer/go.mod`) look good so far. In case you need another self-hosted review after changing significant bits, feel free to re-request a review. Otherwise, I'll unblock this part of the PR and let the **Team WebApp** do the final review of the actual changes.

easyCZ:
I'm fixing up the PR. To make it a bit easier to digest, I'm splitting it up into the following:
* [Align the command structure](https://github.com/gitpod-io/gitpod/pull/10117)
* Introduce config in a file (and configmap)

easyCZ:
This is now ready for review again. Changes have been tested in preview.

easyCZ (approved):

PR #10020, opened 2022-05-14
[common-go] Add file watcher

aledbf:
## Description

Refactor watch of configuration files and reload and remove duplication

## How to test
- Change `ws-daemon` or `ws-daemon` configmap and check the reload message appears in the log

```
{"level":"info","message":"reloading file after change","path":"/config/config.json","serviceContext":{"service":"ws-daemon","version":"commit-5de52b5daf5c35fe22f46adcff90cbaca78b075b"},"severity":"INFO","time":"2022-05-16T05:23:37Z"}}
```

## Release Notes
```release-note
[common-go] Add file watcher
[registry-facade] Refactor watch of configuration file
[ws-daemon] Refactor watch of configuration file
```

### DIFF
--- a/components/common-go/watch/file.go
+++ b/components/common-go/watch/file.go
@@ -1,0 +1,127 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package watch
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/fsnotify/fsnotify"
+	"golang.org/x/xerrors"
+
+	"github.com/gitpod-io/gitpod/common-go/log"
+)
+
+type fileWatcher struct {
+	onChange func()
+
+	watcher *fsnotify.Watcher
+
+	hash string
+}
+
+func File(ctx context.Context, path string, onChange func()) error {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return xerrors.Errorf("unexpected error creating file watcher: %w", err)
+	}
+
+	fw := &fileWatcher{
+		watcher:  watcher,
+		onChange: onChange,
+	}
+
+	// initial hash of the file
+	hash, err := hashConfig(path)
+	if err != nil {
+		return xerrors.Errorf("cannot get hash of file %v: %w", path, err)
+	}
+
+	// visible files in a volume are symlinks to files in the writer's data directory.
+	// The files are stored in a hidden timestamped directory which is symlinked to by the data directory.
+	// The timestamped directory and data directory symlink are created in the writer's target dir.
+	// https://pkg.go.dev/k8s.io/kubernetes/pkg/volume/util#AtomicWriter
+	watchDir, _ := filepath.Split(path)
+	err = watcher.Add(watchDir)
+	if err != nil {
+		watcher.Close()
+		return xerrors.Errorf("unexpected error watching file %v: %w", path, err)
+	}
+
+	log.Infof("starting watch of file %v", path)
+
+	fw.hash = hash
+
+	go func() {
+		defer func() {
+			log.WithError(err).Error("stopping file watch")
+
+			err = watcher.Close()
+			if err != nil {
+				log.WithError(err).Error("unexpected error closing file watcher")
+			}
+		}()
+
+		for {
+			select {
+			case event, ok := <-watcher.Events:
+				if !ok {
+					return
+				}
+
+				if !eventOpIs(event, fsnotify.Create) && !eventOpIs(event, fsnotify.Remove) {
+					continue
+				}
+
+				currentHash, err := hashConfig(path)
+				if err != nil {
+					log.WithError(err).Warn("cannot check if config has changed")
+					return
+				}
+
+				// no change
+				if currentHash == fw.hash {
+					continue
+				}
+
+				log.WithField("path", path).Info("reloading file after change")
+
+				fw.hash = currentHash
+				fw.onChange()
+			case err := <-watcher.Errors:
+				log.WithError(err).Error("unexpected error watching event")
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
+
+	return nil
+}
+
+func hashConfig(path string) (hash string, err error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return "", err
+	}
+	defer f.Close()
+
+	h := sha256.New()
+
+	_, err = io.Copy(h, f)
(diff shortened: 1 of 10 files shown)


utam0k:
I love this feature. Do you have a plan to apply it to other components (e.g. ws-manager)?

aledbf:
> I love this feature. Do you have a plan to apply it to other components (e.g. ws-manager)?

Not really. The only two components that support partial configuration reload are ws-daemon and ws-registry. That said, if we add such a feature to any other component, like ws-manager, we can add the four lines required to detect changes.

utam0k (approved):
works fine

easyCZ (commented):
https://github.com/spf13/viper looks to solve this problem, perhaps we could simplify our usage with it.

components/registry-facade/cmd/run.go
```diff
@@ -150,7 +148,22 @@ var runCmd = &cobra.Command{
...
 
-		go watchConfig(configPath, reg)
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		err = watch.File(ctx, configPath, func() {
+			ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+			defer cancel()
+
+			err := reg.UpdateStaticLayer(ctx, cfg.Registry.StaticLayer)
```
csweichel, line 158: I don't see where the config is actually re-read from disk. The prior implementation had a `config.GetConfig(fn)` call to that end.

components/ws-daemon/cmd/run.go
```diff
@@ -128,7 +129,21 @@ var runCmd = &cobra.Command{
...
 
-		go config.Watch(configFile, dmn.ReloadConfig)
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		err = watch.File(ctx, configFile, func() {
+			ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+			defer cancel()
+
+			err := dmn.ReloadConfig(ctx, &cfg.Daemon)
```
csweichel, line 139: I don't see where the config is actually re-read from disk. The prior implementation had a `reloadConfig()` call in the `config` package to that end.

csweichel (commented):
Change makes a ton of sense.
Left two comments.

jenting (commented):
Having a Kubernetes controller and watches the ConfigMap change (resourceVersion change) is also a way.

components/common-go/watch/file.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	// initial hash of the file
+	hash, err := hashConfig(path)
+	if err != nil {
+		return xerrors.Errorf("cannot get hash of file %v: %w", path, err)
+	}
+
+	watchDir, _ := filepath.Split(path)
+	err = watcher.Add(watchDir)
```
jenting, line 47: If I understand correctly, this watch the directory rather than the config path?

components/common-go/watch/file.go
```diff
@@ -0,0 +1,123 @@
...
+		}()
+
+		for {
+			select {
+			case event, ok := <-watcher.Events:
+				if !ok {
+					return
+				}
+
+				if !eventOpIs(event, fsnotify.Create) && !eventOpIs(event, fsnotify.Remove) {
```
jenting, line 74: Should we skip [rename](https://github.com/fsnotify/fsnotify/blob/3eb7fa4b3213c056d4657a9776041e96e8fe73b6/fsnotify.go#L31) and [chmod](https://github.com/fsnotify/fsnotify/blob/3eb7fa4b3213c056d4657a9776041e96e8fe73b6/fsnotify.go#L32) as well?

sagor999: It does already. This skips everything except for `create` and `remove` ops.

jenting: Oops, I did not see the `!`.

So, if we update the Kubernetes ConfigMap, the behaviour of the file is recreating the file (delete->create)?

components/registry-facade/cmd/run.go
aledbf: Fixed

components/ws-daemon/cmd/run.go
aledbf: Fixed

components/common-go/watch/file.go
aledbf: Yes, this is a weird k8s thing. The keys in the configmap are not mapped to files but to a symlink. Because how fsnotify works, when we update the configmap, there's an atomic update that creates a new directory and swaps the mapped keys.
There was an attempt to fix this by adding the function fsnotify.AddRaw (watching the symlink) but that introduced some weird behaviors. 
For that reasons we watch the directory :(

aledbf:
> Having a Kubernetes controller and watches the ConfigMap change (resourceVersion change) is also a way.

That means we need to watch the configmap using client-go. I would prefer to not add that here.

aledbf:
> https://github.com/spf13/viper looks to solve this problem, perhaps we could simplify our usage with it.

Almost the same code https://github.com/spf13/viper/blob/v1.11.0/viper.go#L440

I am ok to use that but right now we are not using `viper` (another new dependency)

components/common-go/watch/file.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// visible files in a volume are symlinks to files in the writer's data directory.
+	// The files are stored in a hidden timestamped directory which is symlinked to by the data directory.
+	// The timestamped directory and data directory symlink are created in the writer's target dir.
+	// https://pkg.go.dev/k8s.io/kubernetes/pkg/volume/util#AtomicWriter
+	watchDir, _ := filepath.Split(path)
+	err = watcher.Add(watchDir)
+	if err != nil {
+		watcher.Close()
+		return xerrors.Errorf("unexpected error watching file %v: %w", path, err)
```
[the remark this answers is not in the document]
easyCZ: And in other places in this PR too

components/common-go/watch/file.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	log.Infof("starting watch of file %v", path)
+
+	fw.hash = hash
+
+	go func() {
+		defer func() {
+			log.WithError(err).Error("stopping file watch")
```
easyCZ, line 63: Similar to the [discussion in another PR](https://github.com/gitpod-io/gitpod/pull/9973#discussion_r874823408), logs should have first letter capital to be sentences as they are consumed by humans. Errors on the other hand should never be capitalized.

aledbf: fixed

components/common-go/watch/file.go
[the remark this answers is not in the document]
aledbf: fixed

jenting (approved):
/lgtm

Issue #8938, opened 2022-03-23
Support json keys as credentials to access private registries


princerachit:
## Is your feature request related to a problem? Please describe

As a gitpod user who wants to use private registries to pull images, I want to use my json key instead of plain username:password to configure registry access. i.e. my password could be a json e.g. service account key of google cloud.

With https://github.com/gitpod-io/gitpod/pull/8550 a support for using username:password was added but it fails when the auth config passed is built using a json key e.g.
```console
echo $key | docker login -u _json_key --password-stdin
```

## Describe alternatives you've considered

NA

### Failure
<img width="2558" alt="image" src="https://user-images.githubusercontent.com/32481722/159685945-6ec1a05a-1c51-4a62-9487-ac24fc76b779.png">

### Private image

```console
europe-docker.pkg.dev/prince-tf-experiments/dazzle/workspace-images:base
```

kylos101:
@princerachit can you add this to our inbox too?

princerachit:
The issue is due to the way we decode and split the auth string in our code. In the image-builder-mk3 we split by colon which results the json to get split into multiple strings.
https://github.com/gitpod-io/gitpod/blob/6913b4fb2b6bfa4e1149efd5105e005d9d380d4c/components/image-builder-mk3/pkg/auth/auth.go#L210-L215

Due to the check at line 211 the password is never set. This result in the auth string being used for bearer token authentication method instead of basic authentication which results in a 401 error.

When I change above lines to the following to check if this string can be split into at least two string and then treat `segs[1] `as password. image-builder-mk3 succeeds in creation of imagebuild pods.
```go
	if err == nil {
		segs := strings.SplitN(string(dec), ":", 2)
		if len(segs) == 2 {
			res.Username = segs[0]
			res.Password = segs[1]
		}
	}
```

While creating the imagebuild pods mk3 sets the `WORKSPACEKIT_BOBPROXY_ADDITIONALAUTH` env variable which contains both user passs as well as the auth string.

https://github.com/gitpod-io/gitpod/blob/6913b4fb2b6bfa4e1149efd5105e005d9d380d4c/components/image-builder-mk3/pkg/orchestrator/orchestrator.go#L384-L385 

I suspect our credentials setting login involving above env var is causing an issue where it does not use the user and password but instead tries to use auth string for bearer token authentication.

[...]

PR #10031, opened 2022-05-16
[image-builder] Fix error when secret is a json

princerachit:
## Description

The credentials from decoded auth string should be split in such a way that the first substring is treated as username and the rest of the string as password/secret. This is in conformance with what we have in [image-builder-bob](https://github.com/gitpod-io/gitpod/blob/39049b30572807d871a2fdbc6fe8c30e814cb33e/components/image-builder-bob/pkg/proxy/auth.go#L45-L51).
## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/8938

## How to test

Prefer testing in workspace-preview due to known issues (see following `Known Caveats` section)
The json key secret is used in GAR/GCR. It allows you to use your service account key as a json key secret for docker login. Read more [here](https://cloud.google.com/artifact-registry/docs/docker/authentication#json-key).
I am not aware of any other such registry services which accepts a json key as a secret. In any case this PR should solve any issues whether a registry service accepts a json key or a yaml or a plain string.

1. Create a Google Artifact Registry in us region (There is a conflict of keys in preview env with europe region, to be fixed by followup PR)
2. Create a service account with registry writer permission
3. Create a key for above service account
4. Rename the key to key.json
5. Run the following command: `cat key.json | docker login us-docker.pkg.dev -u _json_key --password-stdin`
6. Push an image to the registry (prefer retagging a workspace image e.g. gitpod/workspace-go:latest)
7. Configure your repo to use above image in `.gitpod.yml`
8. Open docker config (`~/.docker/config`) and copy the auth base64encoded string and use it as per instructions of [this PR](https://github.com/gitpod-io/gitpod/pull/8550). Your workspace should start now. 

### Working example
image

### Known Caveats
* The default behaviour of image-build is to use an existing auth if it exists. e.g. in preview env we configure `europe-docker.pkg.dev` and `eu.gcr.io` in our pull secrets. If you create a private registry in any of these registries then your image build will fail due to incorrect key.
* I will raise a follow up PR to solve this issue. Whereas, we can always ignore the existing creds and use the additional auth supplied by user, this will not work well with eu.gcr.io because that is where we pushed the built images.

## Release Notes

```release-note
Fix credential errors when json key is used as secret in image-builder-mk3 
```

### DIFF
--- a/components/image-builder-bob/pkg/proxy/auth.go
+++ b/components/image-builder-bob/pkg/proxy/auth.go
@@ -61,6 +61,7 @@ func (a MapAuthorizer) AddIfNotExists(other MapAuthorizer) MapAuthorizer {
 	}
 	for k, v := range other {
 		if _, ok := a[k]; ok {
+			log.Infof("Skip adding key: %s to MapAuthorizer as it already exists", k)
 			continue
 		}
 		res[k] = v
--- a/components/image-builder-mk3/pkg/auth/auth.go
+++ b/components/image-builder-mk3/pkg/auth/auth.go
@@ -208,9 +208,9 @@ func (a AllowedAuthFor) additionalAuth(domain string) *Authentication {
 	dec, err := base64.StdEncoding.DecodeString(ath)
 	if err == nil {
 		segs := strings.Split(string(dec), ":")
-		if len(segs) == 2 {
+		if len(segs) > 1 {
 			res.Username = segs[0]
-			res.Password = segs[1]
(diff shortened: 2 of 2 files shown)


roboquat:
@princerachit: Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://git.k8s.io/community/contributors/guide/release-notes.md) to remove it.

<details>

</details>

jenting (approved):
Did not test but code LGTM

Issue #10008, opened 2022-05-13
Revert Investigate #9924 and re-apply


geropl:
- [x] Temporary revert: https://github.com/gitpod-io/gitpod/pull/10006
 - [x] Investigate why #9924 breaks prebuilds, fix it, and re-apply

andrew-farries:
Working in preview environment https://af-test-noae1a9bf075.preview.gitpod-dev.com/. This is off main but with https://github.com/gitpod-io/gitpod/pull/10006 reverted (ie the bug should be reproducible there)

* Created a new project from https://github.com/andrew-farries/gitpod-playground
* A prebuild was created in `PENDING` state
* Watching the logs of the prebuild we saw the image build logs followed by the prebuild logs, but the state never left `PENDING`.
* After the prebuild was complete, the UI did not transition to the workspace.
* Reloading the page did not transition to the workspace either, just reloaded the prebuild view
* Starting a new workspace also just led to the prebuild logs view
* By hitting 'skip prebuild', we were able to get to the workspace

PR #10042, opened 2022-05-16
[prebuilds] Fix prebuild updates received by ws-manager-bridge

andrew-farries:
## Description

Un-revert https://github.com/gitpod-io/gitpod/pull/9924 and fix the bug it introduced that caused no prebuild updates to be recorded for any prebuilds.

Many thanks to @laushinka for helping track this down.

## Related Issue(s)
Fixes #https://github.com/gitpod-io/gitpod/issues/10008

## How to test

Create a prebuild in this preview environment and see that it transitions through to READY rather than being stuck in PENDING.

## Release Notes

```release-note
Fix prebuild updates
```

## Documentation
none

### DIFF
--- a/components/ws-manager-bridge/ee/src/bridge.ts
+++ b/components/ws-manager-bridge/ee/src/bridge.ts
@@ -5,15 +5,19 @@
  */
 
 import { WorkspaceManagerBridge } from "../../src/bridge";
-import { injectable } from "inversify";
+import { inject, injectable } from "inversify";
 import { TraceContext } from "@gitpod/gitpod-protocol/lib/util/tracing";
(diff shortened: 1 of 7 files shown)


components/ws-manager-bridge/ee/src/bridge.ts
```diff
@@ -83,7 +83,7 @@ export class WorkspaceManagerBridgeEE extends WorkspaceManagerBridge {
             if (update) {
                 const updatedPrebuild = {
                     ...prebuild,
-                    ...update,
+                    ...update.update,
```
geropl, line 86: 🙈

PR #9819, opened 2022-05-06
Pass quality to code build process 

andreafalzetti:
## Description
Companion PR of https://github.com/gitpod-io/openvscode-server/pull/347

The editor name is misconfigured and not showing correctly when users are using the Insider version.

| Before | After |
|--------|-------|
|  Screenshot 2022-05-12 at 16 49 03      |   Screenshot 2022-05-12 at 16 49 13

Note: Ignore the light/dark mode

## Related Issue(s)

Fixes #9380

## How to test
https://gitpod.slack.com/archives/C01KGM9BH54/p1652369461003149?thread_ts=1652353405.265159&cid=C01KGM9BH54

## Release Notes

```release-note
NONE
```

PR #10045, opened 2022-05-16
[code] Serve webview resources using blobserve

jeanp413:
## Description

Serve webview resources using blobserve, this will allow as to update gp-code nightly as there were some breaking changes for us upstream

## How to test

You need to test this for both stable and insiders vscode to ensure changes are backwards compatible
1. Open a workspace
2. Check webviews work, e.g simple browser
3. Install vim extension, check it activates sucessfully

## Release Notes

```release-note
NONE
```

### DIFF
--- a/WORKSPACE.yaml
+++ b/WORKSPACE.yaml
@@ -7,7 +7,7 @@ defaultArgs:
   jbMarketplacePublishTrigger: "false"
   publishToJBMarketplace: true
   localAppVersion: unknown
-  codeCommit: 9d98f338d0c8584b66195fceb555bb7223c755bc
+  codeCommit: 32d5576921d1fb2a4891adb67f5003c554b04125
   codeQuality: stable
   jetbrainsBackendQualifier: stable
   intellijDownloadUrl: "https://download.jetbrains.com/idea/ideaIU-2022.1.1.tar.gz"
--- a/components/ws-proxy/pkg/proxy/routes.go
+++ b/components/ws-proxy/pkg/proxy/routes.go
@@ -94,15 +94,6 @@ func installWorkspaceRoutes(r *mux.Router, config *RouteHandlerConfig, ip Worksp
 	})
 	routes.HandleSupervisorFrontendRoute(faviconRouter.NewRoute())
 
-	// Theia has a bunch of special routes it probably requires.
-	// TODO(cw): figure out if these routes are still required, and how we deal with specialties of other IDEs.
-	for _, pp := range []string{"/services", "/file-upload"} {
-		routes.HandleDirectIDERoute(r.Path(pp))
-	}
-	for _, pp := range []string{"/mini-browser", "/file", "/files", "/hostedPlugin", "/webview"} {
-		routes.HandleDirectIDERoute(r.PathPrefix(pp))
-	}
-
 	routes.HandleSupervisorFrontendRoute(enableCompression(r).PathPrefix("/_supervisor/frontend"))
 
 	routes.HandleDirectSupervisorRoute(r.PathPrefix("/_supervisor/v1/status/supervisor"), false)
--- a/components/ws-proxy/pkg/proxy/routes_test.go
+++ b/components/ws-proxy/pkg/proxy/routes_test.go
@@ -580,26 +580,6 @@ func TestRoutes(t *testing.T) {
 				Body:   "blobserve hit: /blobserve/gitpod-io/supervisor:latest/main.js\nreadOnly: true\n",
 			},
 		},
-		{
-			Desc: "extensions route GET",
-			Request: modifyRequest(httptest.NewRequest("GET", "https://extensions-foreign.test-domain.com/"+workspaces[0].WorkspaceID+"/extensions.js", nil),
-				addHostHeader,
-				addHeader("Origin", config.GitpodInstallation.HostName),
-				addOwnerToken(workspaces[0].InstanceID, workspaces[0].Auth.OwnerToken),
-			),
-			Expectation: Expectation{
-				Status: http.StatusOK,
-				Header: http.Header{
-					"Access-Control-Allow-Credentials": {"true"},
-					"Access-Control-Allow-Origin":      {"test-domain.com"},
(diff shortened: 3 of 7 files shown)


jeanp413:
Waiting on https://github.com/gitpod-io/gitpod/pull/9819 to be merged as it depends on it

components/ws-proxy/pkg/proxy/routes.go
```diff
@@ -94,15 +94,6 @@ func installWorkspaceRoutes(r *mux.Router, config *RouteHandlerConfig, ip Worksp
 	})
 	routes.HandleSupervisorFrontendRoute(faviconRouter.NewRoute())
 
-	// Theia has a bunch of special routes it probably requires.
```
andreafalzetti, line 97: Are we removing these lines in scope of fixing vscode or just because theia is deprecated now?

jeanp413: because is deprecated

components/ws-proxy/pkg/proxy/workspacerouter.go
```diff
@@ -95,18 +97,20 @@ func matchWorkspaceHostHeader(wsHostSuffix string, headerProvider hostHeaderProv
...
+				// REMOVE this once all workspaces have updated to new vscode version
+				// Don't match if path contains `/__files__/` (blobserve image separator)
+				return false
+			}
 			matches = foreignContentPathR.FindStringSubmatch(req.URL.Path)
 			if matchPort {
 				if len(matches) < 4 {
 					return false
 				}
-				// https://extensions-foreign.ws-eu10.gitpod.io/3000-coral-dragon-ilr0r6eq/index.html
```
andreafalzetti, line 109: What defined before this subdomain `extensions-foreign` and why is now `https://{{uuid}}.`? 🤔

andreafalzetti:
I've tested it and it works. I was able to start a new workspace in the prev env, and webviews work. I've tested the markdown preview and the vim extension.

Now, I am trying to review without much knowledge on the topic, so let's see if I understood something 😄 

My understanding of this change, is:

1. A recent upstream change in VS Code changed how some content related to webview should be served and it broke our integration. To fix it, we’re adjusting the proxy/import configuration for those resources to match the new approach
2. Previously some static assets were served from the workspace itself. Now we moved them to blobserve

Did I miss anything Is this a good description of the change?

Some questions:

1. Looking at [this commit](https://github.com/gitpod-io/openvscode-server/commit/e9a328220ac3a6f7216451151438b145437b1e44). Are these changes coming from upstream or are custom changes to make VS Code work in Gitpod?
2. Is the term “foreign content” a term coming from VS Code?
3. What do we consider foreign content? 
4. Was the content served from the workspace directly, before this change?
5. Why does serving from blobserve fix the issue?

Sorry for all the questions @jeanp413, thanks in advance 🙏

components/ws-proxy/pkg/proxy/routes_test.go
```diff
@@ -580,26 +580,6 @@ func TestRoutes(t *testing.T) {
 				Body:   "blobserve hit: /blobserve/gitpod-io/supervisor:latest/main.js\nreadOnly: true\n",
 			},
 		},
-		{
-			Desc: "extensions route GET",
```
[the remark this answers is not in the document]
jeanp413: removed because I removed the old `foreign` route

components/ws-proxy/pkg/proxy/workspacerouter.go
jeanp413: I think it was designed like this In the beginning because of theia theia and then in our vscode fork we adopted it, then vscode added a check so it has to be an uuid always

jeanp413:
> Looking at [this commit](https://github.com/gitpod-io/openvscode-server/commit/e9a328220ac3a6f7216451151438b145437b1e44). Are these changes coming from upstream or are custom changes to make VS Code work in Gitpod?

Some are custom, some are copied from src/vs/code/browser/workbench/workbench.ts

> Is the term “foreign content” a term coming from VS Code?

Nope, I think it comes from theia

> What do we consider foreign content?

Content served from a different origin 

> Was the content served from the workspace directly, before this change?

yes

> Why does serving from blobserve fix the issue?

Because the vscode server now have a check for `quality-commit` for static resources in the url path, if you run vscode from sources that's Ok as the server will add `quality-commit` in the base url when serving index.html
In production we serve static resources from blobserve so we don't need `quality-commit` in the path

easyCZ (approved):
LGTM

sagor999 (approved):

andreafalzetti (approved):

akosyakov:
@jeanp413 @andreafalzetti It would be nice to update the PR description with info about the cause and solution.

akosyakov:

for me envs are very unstable, i was not able to test, rerunning with clean slate

akosyakov:
> What do we consider foreign content?
> Content served from a different origin

@andreafalzetti Whatever is not coming from VS Code itself but from extensions should be run in a separate origin to disable access to VS Code resources like browser storages. Otherwise they can still sensitive information, like user tokens.  See https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy

akosyakov (approved):
code wise it looks good, but i could not test it, i rely on @andreafalzetti for testing :)

WORKSPACE.yaml
```diff
@@ -7,7 +7,7 @@ defaultArgs:
   jbMarketplacePublishTrigger: "false"
   publishToJBMarketplace: true
   localAppVersion: unknown
-  codeCommit: 64ebb59829c0d3998ebf6b69308aafd49362d29c
+  codeCommit: e9a328220ac3a6f7216451151438b145437b1e44
```
akosyakov, line 10: this one should go to gp-code/main before merging?

jeanp413: yep

jeanp413: Actually nope, we need blobserve to be deployed first before we can update gp-code/main

PR #10050, opened 2022-05-16
[baseserver] Initialize go metrics automatically

easyCZ:
## Description

Automatically initialize go metrics against the baseserver registry - removes extra thing to remember.

## Related Issue(s)

Fixes #

## How to test

Unit tests

## Release Notes

```release-note
NONE
```

## Documentation

NONE

### DIFF
--- a/components/common-go/baseserver/server.go
+++ b/components/common-go/baseserver/server.go
@@ -20,6 +20,7 @@ import (
 	grpc_logrus "github.com/grpc-ecosystem/go-grpc-middleware/logging/logrus"
 	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/collectors"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
@@ -42,6 +43,11 @@ func New(name string, opts ...Option) (*Server, error) {
 
 	server.httpMux = http.NewServeMux()
 	server.http = &http.Server{Handler: server.httpMux}
+
+	err = server.initializeMetrics()
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize metrics: %w", err)
+	}
 
 	err = server.initializeGRPC()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


components/common-go/baseserver/server.go
```diff
@@ -43,6 +44,11 @@ func New(name string, opts ...Option) (*Server, error) {
 	server.httpMux = http.NewServeMux()
 	server.http = &http.Server{Handler: server.httpMux}
 
+	metricsErr := server.initializeMetrics()
+	if metricsErr != nil {
+		return nil, fmt.Errorf("failed to initialize metrics: %w", err)
+	}
```
utam0k, line 50: Is there any reason that you don't use `err`?
```suggestion
	err := server.initializeMetrics()
	if metricsErr != nil {
		return nil, fmt.Errorf("failed to initialize metrics: %w", err)
	}
```

utam0k (commented):
Great PR! I left a comment.

components/common-go/baseserver/server.go
```diff
@@ -387,3 +396,14 @@ func (s *builtinServices) Close() error {
...
+func httpAddress(cfg *ServerConfiguration, l net.Listener) string {
+	if l == nil {
+		return ""
+	}
+	protocol := "http"
+	if cfg != nil && cfg.TLS != nil {
+		protocol = "https"
+	}
+	return fmt.Sprintf("%s://%s", protocol, l.Addr().String())
+}
```
jenting, line 409: It's not related to this PR, please rebase to the latest main branch.

components/common-go/baseserver/server.go
```diff
@@ -7,6 +7,7 @@ package baseserver
 import (
 	"context"
 	"fmt"
+	"github.com/prometheus/client_golang/prometheus/collectors"
```
jenting, line 10: It'd be good if we could couple the import package `"github.com/prometheus/client_golang` together.
```go
        "github.com/prometheus/client_golang/prometheus"
        "github.com/prometheus/client_golang/prometheus/collectors"
	"github.com/prometheus/client_golang/prometheus/promhttp"
```

components/common-go/baseserver/server.go
easyCZ: It is, this is just a move of the function to the bottom to not be in the middle of instance methods of the Server. The github diff renders funny in this case.

components/common-go/baseserver/server.go
easyCZ: If we do want this behaviour, we should ge a linter to automatically re-order the imports as this is purely cosmetic but creates noise and toil in dev cycle. I'll fix it here for now.

components/common-go/baseserver/server.go
easyCZ: Old habit. Normally I'd use 
```
if err := server.initializeMetrics(); err != nil {
	...
}
```
Which makes err local, and prevents shadowing. But since that's for some reason discouraged in our style guide, to pass a stricter version (than we have currently enabled) gometalinter you'd need to assign different error variables as error shadowing is not allowed. It is for a good reason, the exact mistake I made here where I check the error for nil, but wrap the error of `err` rather than `metricsErr`. I've now renamed to `err` but that will shadow the variable.

components/common-go/baseserver/server.go
jenting: > The github diff renders funny in this case.

How can we trust the GitHub diff 😆

jenting (approved):
/lgtm

PR #10052, opened 2022-05-16
[public-api] Publish to NPM

easyCZ:
## Description

Publish public-api typescript defs to NPM so they can be consumed by IDE.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
[public-api] Publish typescript definitions to NPM
```

## Documentation

NONE

### DIFF
--- a/components/public-api/typescript/BUILD.yaml
+++ b/components/public-api/typescript/BUILD.yaml
@@ -4,9 +4,9 @@ packages:
     deps:
       - components/gitpod-protocol:lib
     srcs:
-      - "src/*.ts"
-      - "src/*.js"
+      - src/**
       - package.json
+      - tsconfig.json
     config:
       packaging: library
       dontTest: true
@@ -14,3 +14,16 @@ packages:
         build: ["yarn", "build"]
       yarnLock: ${coreYarnLockBase}/../yarn.lock
       tsconfig: tsconfig.json
+
+  - name: publish
+    type: generic
+    env:
+      - DO_PUBLISH=${publishToNPM}
+    argdeps:
+      - npmPublishTrigger
+    deps:
+      - :lib
+      - components/gitpod-protocol:scripts
+    config:
+      commands:
+        - ["node", "components-gitpod-protocol--scripts/publish.js", "${version}", "components-public-api-typescript--lib/package"]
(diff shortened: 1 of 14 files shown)


components/public-api/typescript/package.json
```diff
@@ -1,10 +1,8 @@
 {
-  "private": true,
   "name": "@gitpod/public-api",
   "version": "0.1.5",
   "license": "UNLICENSED",
   "files": [
-    "client",
```
andrew-farries, line 7: why remove this?

components/public-api/typescript/BUILD.yaml
```diff
@@ -4,13 +4,26 @@ packages:
...
+    env:
+      - DO_PUBLISH=${publishToNPM}
+    argdeps:
+      - npmPublishTrigger
+    deps:
+      - :lib
+      - components/gitpod-protocol:scripts
+    config:
+      commands:
+        - ["node", "components-gitpod-protocol--scripts/publish.js", "${version}", "components-public-api-typescript--lib/package"]
```
andrew-farries, line 29: How is `node` able to find `publish.js` here (the `components-` don't look like valid paths?)

components/public-api/typescript/BUILD.yaml
```diff
@@ -4,13 +4,26 @@ packages:
...
+      - tsconfig.json
     config:
       packaging: library
       dontTest: true
       commands:
         build: ["yarn", "build"]
       yarnLock: ${coreYarnLockBase}/../yarn.lock
       tsconfig: tsconfig.json
+
+  - name: publish
```
andrew-farries, line 18: How is this triggered? We already have this:

[mp/public-api-publish-npm](https://github.com/gitpod-io/gitpod/blob/c887240be815c345efbba3fcc2660935637bd767/components/BUILD.yaml#L81-L89)

Does that need to be updated

components/public-api/typescript/BUILD.yaml
easyCZ: I've no idea! This whole thing looks to be way more complicated than it needs to be but not really well placed to simplify. The code here just mirrors what supervisor-api does

components/public-api/typescript/BUILD.yaml
easyCZ: Not sure, mirrors supervisor-api setup

andrew-farries (approved):

Issue #10070, opened 2022-05-17
Use a single object storage bucket per Gitpod installation


atduarte:
## Is your feature request related to a problem? Please describe
Currently, full object storage permission is required to install Gitpod as it creates and uses multiple storage buckets. This is a problem for many self-hosted customers. Additionally, AWS account default limit is 100 buckets [[1](https://docs.aws.amazon.com/AmazonS3/latest/userguide/BucketRestrictions.html)].

## Describe the behaviour you'd like
Self-hosted installations should only use one storage bucket, while SaaS in GCP, for egress optimization regions, should remains as is.

## Describe alternatives you've considered
None that I can think of.

## Additional context
[Internal slack thread 1](https://gitpod.slack.com/archives/C02F19UUW6S/p1652789401018709), and [2](https://gitpod.slack.com/archives/C01LDA28WV6/p1652728380421749)

sagor999:
@aledbf is working on this: https://github.com/gitpod-io/gitpod/pull/10073
Assigning to him, as I was about to start working on this... good that I checked slack threads prior to diving in.
Please update issues when you start working on them.

aledbf:
@sagor999 sorry about that.

sagor999:
@aledbf @kylos101 asked me to help out to take this issue\fix to the finish line, so going to assign it to myself.

PR #10073, opened 2022-05-17
[content-service] Add support to use existing S3 bucket

aledbf:
## Description
This PR introduces a new "single bucket mode" to the S3 storage implementation.

In some scenarios, we could not have enough permissions to create buckets. In those cases the "one-bucket per user" implementation fails.

If there's a bucket name configured (see installer/content-service config change), that bucket name is used and the user ID is prefixed to the object name. If no such bucket name is configured, the old "one bucket per user" behaviour kicks in. This way, users can maintain backwards compatibility for now. There is no automatic migration or "hybrid mode" where we fall back to the old behaviour. We don't have enough installations in the wild which were able to actually use the old behaviour to justify introducing such a mode.

We maintain the old mode in GCP (and maybe even S3) for now because 
- it maintains backwards compatibility
- it allows for region-local optimisations. That said, we could just as well introduce single regional buckets instead.
- moving away from it would require migration for gitpod.io

## How to test
Precursor: ensure the installation uses minio/S3, and that the `bucket` is configured
- start/stop a workspace, make sure the backup happens and is restored
- take a snapshot and open a workspace from it
- run a prebuild and open a workspace from it
- use the VS Code settings sync, make sure it still syncs
- check out prebuild logs in the dashboard

## Release Notes

```release-note
[content-service] Add support to use a single S3 bucket
```

Fixes https://github.com/gitpod-io/gitpod/issues/10070

### DIFF
--- a/components/content-service/pkg/storage/minio.go
+++ b/components/content-service/pkg/storage/minio.go
@@ -339,7 +339,11 @@ func (rs *DirectMinIOStorage) bucketName() string {
 }
 
 func (rs *DirectMinIOStorage) objectName(name string) string {
-	return minioWorkspaceBackupObjectName(rs.Username, rs.WorkspaceName, name)
+	var username string
+	if rs.MinIOConfig.BucketName != "" {
+		username = rs.Username
+	}
+	return minioWorkspaceBackupObjectName(username, rs.WorkspaceName, name)
 }
 
 func newPresignedMinIOAccess(cfg config.MinIOConfig) (*presignedMinIOStorage, error) {
@@ -532,13 +536,17 @@ func (s *presignedMinIOStorage) BlobObject(name string) (string, error) {
 }
 
 // BackupObject returns a backup's object name that a direct downloader would download
-func (s *presignedMinIOStorage) BackupObject(workspaceID, name string) string {
-	return minioWorkspaceBackupObjectName("", workspaceID, name)
+func (s *presignedMinIOStorage) BackupObject(ownerID string, workspaceID, name string) string {
+	var username string
+	if s.MinIOConfig.BucketName != "" {
+		username = ownerID
+	}
+	return minioWorkspaceBackupObjectName(username, workspaceID, name)
 }
 
 // InstanceObject returns a instance's object name that a direct downloader would download
-func (s *presignedMinIOStorage) InstanceObject(workspaceID string, instanceID string, name string) string {
-	return s.BackupObject(workspaceID, InstanceObjectName(instanceID, name))
+func (s *presignedMinIOStorage) InstanceObject(ownerID string, workspaceID string, instanceID string, name string) string {
+	return s.BackupObject(ownerID, workspaceID, InstanceObjectName(instanceID, name))
 }
 
 func translateMinioError(err error) error {
--- a/components/content-service/pkg/storage/minio.go
+++ b/components/content-service/pkg/storage/minio.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -129,6 +130,7 @@ func (rs *DirectMinIOStorage) Init(ctx context.Context, owner, workspace, instan
 	rs.Username = owner
 	rs.WorkspaceName = workspace
 	rs.InstanceID = instance
+
 	err = rs.Validate()
 	if err != nil {
 		return err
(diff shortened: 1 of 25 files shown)


aledbf:
Configuration example:
```
objectStorage:
  inCluster: false
  kind: minio
  s3:
    endpoint: minio.default.svc.cluster.local:9000
    accessKey: GlrlMt5e40dnW0GhveAX
    secretKey: 0L6XW0.GWcD8CBMseV31
    
    bucket: some-valid-bucket
```

components/content-service/pkg/storage/minio.go
```diff
@@ -310,17 +312,21 @@ func (rs *DirectMinIOStorage) Upload(ctx context.Context, source string, name st
...
+		return bucketName
+	}
+
 	return fmt.Sprintf("gitpod-user-%s", ownerID)
 }
 
-func minioWorkspaceBackupObjectName(workspaceID string, name string) string {
-	return fmt.Sprintf("workspaces/%s/%s", workspaceID, name)
+func minioWorkspaceBackupObjectName(ownerID, workspaceID, name string) string {
+	return filepath.Join(ownerID, "workspaces", workspaceID, name)
```
corneliusludmann, line 324: With this change, we change the name of the workspace bucket object name, correct? Wouldn't this introduce a backward-incompatible change? We would not be able to open workspace backups that has been created prior this change, right?

I think, we need to distinguish between the cases where a bucket name is given and where not.

components/content-service/pkg/storage/minio.go
```diff
@@ -526,8 +532,8 @@ func (s *presignedMinIOStorage) BlobObject(name string) (string, error) {
 }
 
 // BackupObject returns a backup's object name that a direct downloader would download
-func (s *presignedMinIOStorage) BackupObject(workspaceID string, name string) string {
-	return minioWorkspaceBackupObjectName(workspaceID, name)
+func (s *presignedMinIOStorage) BackupObject(workspaceID, name string) string {
+	return minioWorkspaceBackupObjectName("", workspaceID, name)
```
corneliusludmann, line 536: What is the reason that the username is not set here but in 
```
func (s *presignedMinIOStorage) Bucket(ownerID string) string {
```
above it is?!

install/installer/pkg/config/v1/config.go
```diff
@@ -157,6 +157,8 @@ type ObjectStorage struct {
 type ObjectStorageS3 struct {
 	Endpoint    string    `json:"endpoint" validate:"required"`
 	Credentials ObjectRef `json:"credentials" validate:"required"`
+
+	BucketName string `json:"bucket"`
```
corneliusludmann, line 161: NIT: Would you mind adding a comment to this config value that explains what happens when this bucket name is set (and more importantly what happens when this value is not set?

corneliusludmann (changes requested):
Awesome to see this change! :rocket: 

Added a few comments regarding backward compatibility. :thinking:

corneliusludmann:
Follow-up issue: Add config for S3 bucket to KOTS (Replicated) installer.

corneliusludmann:
Follow-up issue: Document S3 bucket config in self-hosted docs

components/content-service/pkg/storage/minio.go
aledbf: > We would not be able to open workspace backups that has been created prior this change, right?

Correct. There is no automatic migration or "hybrid mode" where we fall back to the old behavior.

install/installer/pkg/config/v1/config.go
aledbf: done

components/content-service/pkg/storage/minio.go
```diff
@@ -329,11 +335,11 @@ func (rs *DirectMinIOStorage) BackupObject(name string) string {
 }
 
 func (rs *DirectMinIOStorage) bucketName() string {
-	return minioBucketName(rs.Username)
+	return minioBucketName(rs.Username, rs.MinIOConfig.BucketName)
 }
 
 func (rs *DirectMinIOStorage) objectName(name string) string {
-	return minioWorkspaceBackupObjectName(rs.WorkspaceName, name)
+	return minioWorkspaceBackupObjectName(rs.Username, rs.WorkspaceName, name)
```
sagor999, line 342:
```suggestion
	var username string
	if rs.MinIOConfig.BucketName != "" {
		username = rs.Username
	}
	return minioWorkspaceBackupObjectName(username, rs.WorkspaceName, name)

```
that way if bucket name is set, then we default to new way of doing things, otherwise we maintain backwards compatibility to old way of accessing buckets.

components/content-service/pkg/storage/minio.go
sagor999: I think that is because in this case it is `presignedMinIOStorage`

sagor999 (changes requested):
can you please add unit tests that would ensure old code is working correctly, and new one also generates correct bucket and path.

components/content-service/pkg/storage/minio.go
sagor999: So this feature would only work with directMinioAccess, presignedMinIOStorage (that I think is used by prebuilds, log upload, etc), would not work correctly with dedicated bucket. It would still attempt to access bucket the old way. :thinking:

sagor999: I think presignedMinio would not be working correctly with dedicated bucket. So I committed a fix to it.

sagor999:
@corneliusludmann @aledbf I added tests and fixed directMinio and presignedMinio access.
Tested in preview env, and made sure that now all content indeed is being written into dedicated bucket if it was set in installer config.

Please double check tests that I added to ensure they are correct.

This also ensures backward compatibility (rolling out this update should not break any existing self hosted installations). 
But this feature should cannot be enabled on existing installations without losing all data or manually moving old buckets into new bucket.

components/content-service/pkg/storage/minio.go
sagor999: this should be fixed now.

sagor999 (approved):
with tests it looks good now (feels weird approving my own changes to PR :smiley: )

corneliusludmann (approved):
Looks good. Thanks, @aledbf and @sagor999! :rocket:

akosyakov (approved):
lgtm

mustard-mh (approved):
LGTM in supervisor mod 🙈

PR #10074, opened 2022-05-17
Add SSH connection analytics

iQQBot:
## Description

This PR introduce SSH connection analytics in `ws-proxy`

## Related Issue(s)

Fixes #

## How to test

1. open a workspace
2. use copy/paste SSH command to connect
3. watch segment event, you will see `ssh_connection ` event 

image
image

## Release Notes

```release-note
Add SSH connection analytics
```

## Documentation

### DIFF
--- a/components/ws-proxy/pkg/sshproxy/server.go
+++ b/components/ws-proxy/pkg/sshproxy/server.go
@@ -6,13 +6,16 @@ package sshproxy
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"strings"
 	"time"
 
+	"github.com/gitpod-io/gitpod/common-go/analytics"
 	"github.com/gitpod-io/gitpod/common-go/log"
 	supervisor "github.com/gitpod-io/gitpod/supervisor/api"
+	tracker "github.com/gitpod-io/gitpod/ws-proxy/pkg/analytics"
 	p "github.com/gitpod-io/gitpod/ws-proxy/pkg/proxy"
 	"golang.org/x/crypto/ssh"
 	"golang.org/x/xerrors"
@@ -20,6 +23,10 @@ import (
 )
 
 const GitpodUsername = "gitpod"
+
+var ErrWorkspaceNotFound = errors.New("not found workspace")
+var ErrAuthFailed = errors.New("auth failed")
+var ErrUsernameFormat = errors.New("username format is not correct")
 
 type Session struct {
 	Conn *ssh.ServerConn
@@ -51,11 +58,26 @@ func New(signers []ssh.Signer, workspaceInfoProvider p.WorkspaceInfoProvider, he
 
 	server.sshConfig = &ssh.ServerConfig{
 		ServerVersion: "SSH-2.0-GITPOD-GATEWAY",
-		PasswordCallback: func(conn ssh.ConnMetadata, password []byte) (*ssh.Permissions, error) {
+		PasswordCallback: func(conn ssh.ConnMetadata, password []byte) (perm *ssh.Permissions, err error) {
 			workspaceId, ownerToken := conn.User(), string(password)
-			err := server.Authenticator(workspaceId, ownerToken)
+			wsInfo, err := server.Authenticator(workspaceId, ownerToken)
+			defer func() {
+				server.TrackSSHConnection(wsInfo, "auth", err)
+			}()
 			if err != nil {
-				return nil, err
+				if err == ErrAuthFailed {
+					return
+				}
+				args := strings.Split(conn.User(), "#")
+				if len(args) != 2 {
+					return
+				}
+				workspaceId, ownerToken = args[0], args[1]
+				wsInfo, err = server.Authenticator(workspaceId, ownerToken)
+				if err == nil {
+					err = errors.New("miss private key")
+				}
+				return
 			}
 			return &ssh.Permissions{
 				Extensions: map[string]string{
@@ -67,10 +89,13 @@ func New(signers []ssh.Signer, workspaceInfoProvider p.WorkspaceInfoProvider, he
 			args := strings.Split(conn.User(), "#")
 			// workspaceId#ownerToken
 			if len(args) != 2 {
-				return nil, fmt.Errorf("username error")
+				return nil, ErrUsernameFormat
 			}
 			workspaceId, ownerToken := args[0], args[1]
-			err := server.Authenticator(workspaceId, ownerToken)
+			wsInfo, err := server.Authenticator(workspaceId, ownerToken)
+			defer func() {
+				server.TrackSSHConnection(wsInfo, "auth", err)
+			}()
 			if err != nil {
 				return nil, err
 			}
@@ -108,6 +133,7 @@ func (s *Server) HandleConn(c net.Conn) {
 	key, err := s.GetWorkspaceSSHKey(ctx, wsInfo.IPAddress)
 	if err != nil {
 		cancel()
+		s.TrackSSHConnection(wsInfo, "connect", err)
 		return
 	}
 	cancel()
@@ -121,6 +147,7 @@ func (s *Server) HandleConn(c net.Conn) {
 	remoteAddr := wsInfo.IPAddress + ":23001"
 	conn, err := net.Dial("tcp", remoteAddr)
 	if err != nil {
+		s.TrackSSHConnection(wsInfo, "connect", err)
 		log.WithField("instanceId", wsInfo.InstanceID).WithField("workspaceIP", wsInfo.IPAddress).WithError(err).Error("dail failed")
 		return
 	}
@@ -137,12 +164,15 @@ func (s *Server) HandleConn(c net.Conn) {
 		Timeout: 10 * time.Second,
 	})
 	if err != nil {
+		s.TrackSSHConnection(wsInfo, "connect", err)
 		log.WithField("instanceId", wsInfo.InstanceID).WithField("workspaceIP", wsInfo.IPAddress).WithError(err).Error("connect failed")
 		return
 	}
 	s.Heartbeater.SendHeartbeat(wsInfo.InstanceID, false)
 	client := ssh.NewClient(clientConn, clientChans, clientReqs)
 	ctx, cancel = context.WithCancel(context.Background())
+
+	s.TrackSSHConnection(wsInfo, "connect", nil)
 
 	go func() {
 		client.Wait()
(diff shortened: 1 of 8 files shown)


components/ws-proxy/pkg/sshproxy/server.go
```diff
@@ -161,15 +194,38 @@ func (s *Server) HandleConn(c net.Conn) {
...
+func (s *Server) TrackSSHConnection(wsInfo *p.WorkspaceInfo, staging string, err error) {
+	// if we didn't find an associated user, we don't want to track
+	if wsInfo == nil {
+		return
+	}
+	propertics := make(map[string]interface{})
+	propertics["workspaceId"] = wsInfo.WorkspaceID
+	propertics["instanceId"] = wsInfo.InstanceID
+	propertics["state"] = "success"
+	propertics["staging"] = staging
```
akosyakov, line 217: phase or stage?

components/ws-proxy/pkg/sshproxy/server.go
```diff
@@ -51,11 +58,29 @@ func New(signers []ssh.Signer, workspaceInfoProvider p.WorkspaceInfoProvider, he
 
 	server.sshConfig = &ssh.ServerConfig{
 		ServerVersion: "SSH-2.0-GITPOD-GATEWAY",
-		PasswordCallback: func(conn ssh.ConnMetadata, password []byte) (*ssh.Permissions, error) {
+		BannerCallback: func(conn ssh.ConnMetadata) string {
+			if strings.Contains(conn.User(), "#") {
+				return "Welcome use Gitpod. If you see a prompt to enter a password, fill in any value"
```
iQQBot, line 63: Can anyone provide some better guidewords?

components/ws-proxy/pkg/sshproxy/server.go
iQQBot: updated

components/ws-proxy/pkg/sshproxy/server.go
```diff
@@ -21,6 +24,10 @@ import (
 
 const GitpodUsername = "gitpod"
 
+var ErrWorkspaceNotFound = errors.New("not found workspace")
+var ErrAuthFailed = errors.New("auth failed")
```
akosyakov, line 28: it seems to be very generic, i.e. does not add any useful information, we already track `auth` as a stage and no that it failed. It would be interesting to know whether password is wrong or any other reasons

components/ws-proxy/pkg/sshproxy/server.go
```diff
@@ -51,11 +58,29 @@ func New(signers []ssh.Signer, workspaceInfoProvider p.WorkspaceInfoProvider, he
...
-				return nil, err
+			wsInfo, err := server.Authenticator(workspaceId, ownerToken)
+			defer func() {
+				server.TrackSSHConnection(wsInfo, "auth", err)
+			}()
+			if err != nil && err != ErrAuthFailed {
+				args := strings.Split(conn.User(), "#")
+				// workspaceId#ownerToken
+				if len(args) != 2 {
+					return nil, ErrAuthFailed
```
akosyakov, line 77: ErrUsernameFormat ?

components/ws-proxy/pkg/sshproxy/server.go
```diff
@@ -108,6 +136,7 @@ func (s *Server) HandleConn(c net.Conn) {
 	key, err := s.GetWorkspaceSSHKey(ctx, wsInfo.IPAddress)
 	if err != nil {
 		cancel()
+		s.TrackSSHConnection(wsInfo, "connect", err)
```
akosyakov, line 139: i see this function can exit above too, should we track such exits as well?

components/ws-proxy/pkg/sshproxy/server.go
iQQBot: This is only convenient to determine the type of error, track and this has no relationship

components/ws-proxy/pkg/sshproxy/server.go
iQQBot: what do you mean? you mean defer?

components/ws-proxy/pkg/sshproxy/server.go
iQQBot: The original error should be used here

iQQBot: In fact, this should be workspace not found

components/ws-proxy/pkg/sshproxy/server.go
akosyakov: How can I trigger it?

cc @gtsiolis

components/ws-proxy/pkg/sshproxy/server.go
akosyakov: I mean that HandleConn can exit before reaching any "connect" track either with error or no. It seems we don't track such neither in logs or with analytics.

components/ws-proxy/pkg/sshproxy/server.go
akosyakov: "Welcome to Gitpod. Press enter to continue if you see a password prompt." ?

components/ws-proxy/pkg/sshproxy/server.go
iQQBot: The above exit condition cannot be analyzed because there is no associated workspace information

akosyakov: Can we track it with prometheus? Another PR?

components/ws-proxy/pkg/sshproxy/server.go
iQQBot: > How can I trigger it?

open a workspace in gitpod.io (no privatekey default)

use copy/paste command to connect with this preview environment workspace

components/ws-proxy/pkg/sshproxy/server.go
iQQBot: Yes

akosyakov:
@iQQBot Could you add a new event in https://www.notion.so/gitpod/Tracking-Plan-ed27d1492da745b9b31711dd73dcbe57 please, explain properties, phases and possible values? 🙏

components/ws-proxy/pkg/sshproxy/server.go
iQQBot: This is related to SSH clients, some of which do not see this prompt （sorry this is not working in workspace 🤦‍♂️

gtsiolis: > This is related to SSH clients, some of which do not see this prompt

@iQQBot Using the built-in `Terminal` app from macOS, I am seeing the prompt when having a private key, but I'm _not_ seeing the prompt when not having any private key. ❗ 

> Welcome to Gitpod. Press enter to continue if you see a password prompt.

Agree with @akosyakov's copy suggestion. However, does this work? What's happening in the backstage? 🔮 

---

Also, the prompt is printed in line with the rest of the output when you connect to a workspace and merges the text from the workspace terminal output, see screenshot below. Could we use a new line to push the workspace terminal output further below?

Screenshot 2022-05-18 at 11 26 44 AM

components/ws-proxy/pkg/sshproxy/server.go
```diff
@@ -51,11 +58,32 @@ func New(signers []ssh.Signer, workspaceInfoProvider p.WorkspaceInfoProvider, he
 
 	server.sshConfig = &ssh.ServerConfig{
 		ServerVersion: "SSH-2.0-GITPOD-GATEWAY",
-		PasswordCallback: func(conn ssh.ConnMetadata, password []byte) (*ssh.Permissions, error) {
+		BannerCallback: func(conn ssh.ConnMetadata) string {
+			if strings.Contains(conn.User(), "#") {
+				return "Welcome use Gitpod. If you see a prompt to enter a password, fill in any value"
```
andreafalzetti, line 63: Consider adding new lines after the message, because in Windows it looks a bit messy:

Screenshot 2022-05-18 at 11 08 57

components/ws-proxy/pkg/sshproxy/server.go
akosyakov: @iQQBot Does it really worth it? If a user does not see, I afraid that it adds even more magic. Maybe we should improve docs first?

components/ws-proxy/pkg/sshproxy/server.go
akosyakov: Do you think it is worth it?

I had users asking already why do you need private key if there is no public. I afraid that next question will be why do you ask password if you don't need it.

akosyakov:
@iQQBot Could we extract an improvement in another PR? I think the improvement is questionable and need to be discussed. Analytics though looks good already. I would like analytics parts get merged.

components/ws-proxy/pkg/sshproxy/server.go
gtsiolis: Agree that this is already cryptic enough.

components/ws-proxy/pkg/sshproxy/server.go
iQQBot: already revert this change, now this PR focus on analytics

components/ws-proxy/pkg/sshproxy/server.go
```diff
@@ -161,15 +191,38 @@ func (s *Server) HandleConn(c net.Conn) {
...
+
+	if err != nil {
+		propertics["state"] = "failed"
+		propertics["cause"] = err.Error()
+	}
+
+	tracker.Track(analytics.TrackMessage{
+		Identity:   analytics.Identity{UserID: wsInfo.OwnerUserId},
+		Event:      "ssh_connection",
+		Properties: propertics,
```
akosyakov, line 224: Can we provide timestamp here to make sure that events are ordered properly? Otherwise I'm not sure whether they can race in Segment.

iQQBot: I can add it, but segement has added timestamp to every event by default, which may be a duplicate

iQQBot: https://github.com/segmentio/analytics-go/blob/v3.0/analytics.go#L152-L189

akosyakov (approved):
lgtm, thank you a lot!

akosyakov:
Could you also add a metric for prometheus? I don't think we need to track duration of ssh connections, but having a counter of auth/connection attempts and success rate will be helpful. We don't need to label workspaces or collect any error information but only use it for monitoring and alerts. If we see a spike then we should go to logs and check them.

andreafalzetti:
> watch segment event, you will see ssh_connection event

Where do I verify the date? From segment.com? 🤔

akosyakov:
> Ok, no problem, just one point needs to be clarified, as a public network to expose the IP of port 22, is bound to be a variety of hackers violent crack every day, I suggest that the success rate of the statistics here should ignore the user name length of less than 6 digits

Let's try without first?

sagor999 (approved):

utam0k (approved):

nandajavarma (approved):
LGTM from the installer side. Also, noting that this does not affect the self hosted user.

PR #10075, opened 2022-05-17
[server] Explicitely list teams getting XL resources

csweichel:
## Description
This PR makes the decision which team/user gets "XL resources" explicitly configurable, as compared to tying it to the subscription. This will help us steer that experiment better while keeping cost under control.

- needs a follow-up ops PR which adds the actual teams

Issue: https://github.com/gitpod-io/ops/issues/2346

## How to test
(all set up on this branch)
- create a team
- add the team UUID to the `xlTeams` server config
- make sure server restarts
- add a cluster with the `has-more-resources` constraint
- check that the workspace on that team starts on that workspace

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/server/ee/src/user/user-service.ts
+++ b/components/server/ee/src/user/user-service.ts
@@ -5,7 +5,14 @@
  */
 
 import { UserService, CheckSignUpParams, CheckTermsParams } from "../../../src/user/user-service";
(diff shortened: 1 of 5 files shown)


csweichel:
There's already a flag on the team plan - this change is not neccesary

PR #10076, opened 2022-05-17
Preview environment CLI

ArthurSens:
## Description

Following our [internal RFC](https://www.notion.so/gitpod/A-Go-based-CLI-for-interacting-with-Preview-Environments-1834aa90bc104a0b836dd523e22f9e93), the work done in this CLI was to make sure it is easily extendable with new commands and that the architecture makes it easy to test.

It also introduces the first command asked in Milestone 1(see [important notes](#Important-notes)) of that same RFC. Which makes it possible to install the kube-context of different preview environments.

### `install-k3s-context.sh` was not re-implemented in Go yet.

The implementation is not easy and I think it would delay this PR too much. Our current implementation in bash uses SSH tunneling and I'm not sure how to implement/replace that in Go. I'm also not sure how to write a good test for such implementation 😬.

If possible, I'd like to convert this into a separate issue and tackle the re-implementation and testing into a separate PR.

### `install-k3s-context.sh` was modified to install context from custom branches.

Since one of the requirements was the ability to install context from different branches, I extended our bash script to install context from custom branches 🙂 

### `install-k3s-context.sh` doesn't exit with code != 0 when preview doesn't exist.

And as a consequence, `previewctl install-context <unexisting-branch>` succeeds even with a broken context. I'm on the fence if this bug should be a blocker for this PR or not, but I've already spent a good time trying to solve this one without success 😅. The options that I see are:

* Since users of this CLI are mostly interested in their own previews(which works), this bug could be tackled in a separate issue/PR.

Or

* We shouldn't merge this PR until the bug is fixed, therefore I need some extra help 😝 

### Code walkthrough

https://www.loom.com/share/6da6cf3b6b61457e8c0f17736e716434

## How to test

A couple of options to try

* Start a workspace from this pull request - The right context should be already installed
* Try installing the context from another existing branch by running `previewctl install-context --branch <branch-name>`
* Create a new branch from this PR, commit something, open a new workspace from the new branch/commit, the context should be installed when starting the workspace.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/dev/preview/previewctl/cmd/get_name.go
+++ b/dev/preview/previewctl/cmd/get_name.go
@@ -1,0 +1,27 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/gitpod-io/gitpod/previewctl/pkg/preview"
+	"github.com/spf13/cobra"
+)
+
+func getNameCmd() *cobra.Command {
+
+	cmd := &cobra.Command{
+		Use:   "get-name",
+		Short: "Returns the name of the preview for the corresponding branch.",
+		Run: func(cmd *cobra.Command, args []string) {
+			p := preview.New(branch)
+
+			fmt.Println(p.GetPreviewName())
+		},
+	}
+
+	return cmd
+}
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -31,6 +31,11 @@ ports:
   # Dev Theia
   - port: 13444
 tasks:
+  - name: Install Preview Environment kube-context
+    command: |
+      (cd dev/preview/previewctl && go install .)
+      previewctl install-context
+      exit
   - name: Add Harvester kubeconfig
     command: |
       ./dev/preview/util/download-and-merge-harvester-kubeconfig.sh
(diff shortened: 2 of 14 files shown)


ArthurSens:
Pinging @gitpod-io/platform just for awareness and in case you want to provide early feedback

vulkoingim:
Looks good, I just have a few nits about the file names in the `cmd` package - in general the convention is not to use mixed caps in the file/package names. Usually I would go with a single lowercase word.

ArthurSens:
> Looks good, I just have a few nits about the file names in the cmd package - in general the convention is not to use mixed caps in the file/package names. Usually I would go with a single lowercase word.

Ah yes, thanks for pointing that out. Just renamed the files :)

dev/preview/previewctl/cmd/get_name.go
```diff
@@ -0,0 +1,28 @@
+/*
+Copyright © 2022 NAME HERE <EMAIL ADDRESS>
+
+*/
```
vulkoingim, line 4: This is in a few places and is unnecessary :) Unless there's actually a requirement to fill it with something relevant.

ArthurSens: Oh whoops, that was auto-generated and I forgot to clean it up 😅. Thanks!

It is required indeed, but not like that :P

vulkoingim (approved):

ArthurSens:
/hold

Looks like some changes made to main have broken this build :/, I'll try rebasing

Issue #9950, opened 2022-05-11
Use the published gitpod theme in gitpod


gtsiolis:
### Bug description

Following up from https://github.com/gitpod-io/gitpod/issues/4281, when you install the newly published **Gitpod Theme**[[1](https://open-vsx.org/extension/gitpod/gitpod-theme)], users end up with duplicate theme entries.

<img width="619" alt="Screenshot 2022-05-12 at 1 19 14 AM" src="https://user-images.githubusercontent.com/120486/167957286-62da961c-6921-499a-a8fe-6bf00f38b4f7.png">

### Steps to reproduce

1. Open any workspace
2. Search and install the new Gitpod Theme
3. Set Gitpod Light or Gitpod Dark Theme
4. Open Color Theme Preferences
5. Notice the duplicate Gitpod Light and Gitpod Dark entries

jeanp413:
Yes, I will remove the ones that are built-in inside gp-code

PR #10079, opened 2022-05-17
Install gitpod.gitpod-theme as builtin extension

jeanp413:
## Description

The extension is already bundled when building vscode, this will only update the extension to a newer version if any exist.
If there's no access to openvsx for some reason (self-hosted case) bundled version will be used.

Fixes https://github.com/gitpod-io/gitpod/issues/9950

## How to test

1. Select insiders
2. Open workspace
3. Check gitpod theme is installed

## Release Notes

```release-note
NONE
```


mustard-mh (approved):
Code lgtm

Works as expected

|ps| builtin| command|
|-|-|-|
|image|image| image |

Issue #10089, opened 2022-05-18
Image builder does not support private container registry when gitpod uses same registry


princerachit:
### Bug description

If gitpod installation contains image pull secret for a registry say `myreg.com` and user tries to configure the same registry provider to pull private images then gitpod fails to pull the image.
This happens because gitpod tries to use the existing credentials of imagepull secrets rather than the credential provided by the user.

e.g. We have europe-docker.pkg.dev access credentials configured in preview env as part of image pull secret. When I create a project in preview env and use europe-docker.pkg.dev for my private images, gitpod does not use my credentials.

### Steps to reproduce

This can be reproduced easily in preview env.

* Create a private google artifact registry in europe region such that the host is `europe-docker.pkg.dev`.
* Create a service account and give it registry writer permission to your repo.
* Upload an image in your registry.
* Configure your repo with a .gitpod.yml which uses this image.
* Follow steps in https://github.com/gitpod-io/gitpod/pull/8550 description.

The image build will fail.

NOTE: The image build will fail early if the changes of [this PR ](https://github.com/gitpod-io/gitpod/pull/10031)Iare not included in your preview env.

### Expected behavior

Image build should have worked and correct credentials should have been used.

princerachit:
There are a few ways we can solve this problem.

1. Maintain a list of auths for a registry combining both WORKSPACEKIT_BOBPROXY_AUTH and WORKSPACEKIT_BOBPROXY_ADDITIONALAUTH. Try the next auth in the list if the first one fails. -> This is expensive as their would be multiple calls (as we need to retry)
2. Modify the bob proxy code to maintain two maps -> One for Target ref which will use WORKSPACEKIT_BOBPROXY_AUTH and would use WORKSPACEKIT_BOBPROXY_ADDITIONALAUTH only if former doesn't have an entry. Another for Base ref which will WORKSPACEKIT_BOBPROXY_ADDITIONALAUTH and will back to WORKSPACEKIT_BOBPROXY_AUTH only if former does not have an entry.
3. Accept per repo/image auths from user -> Fallback to registry level credentials if repo level credential is missing. i.e. do a maximum prefix match against all the keys.

princerachit:
Reopening this as we reverted the PR.

PR #10094, opened 2022-05-18
[image-builder-bob] Use separate auth for target and base

princerachit:
## Description

Use terms such `base` and `target` to distinguish between image source (base) and built image (target).

Use separate auth for target and base so that when the target and base images are both in the same registry provider, correct auth is used.

e.g. User wants to use an image from their private gcr but we want to push the built image to gitpod's private gcr.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/10089

## How to test

Follow Reproduce steps from this issue https://github.com/gitpod-io/gitpod/issues/10089. Error should not be reproducible.

I already tested this in the preview env of this branch.

### Image builds

image

### Workspace Starts

image

## Release Notes

```release-note
Fix conflicting auth selection for image-builder-bob
```

### DIFF
--- a/components/image-builder-bob/cmd/proxy.go
+++ b/components/image-builder-bob/cmd/proxy.go
@@ -19,8 +19,8 @@ import (
 
 var proxyOpts struct {
 	BaseRef, TargetRef string
-	Auth               string
-	AdditionalAuth     string
+	BaseAuth           string
+	TargetAuth         string
 }
 
 // proxyCmd represents the build command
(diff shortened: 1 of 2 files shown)


components/image-builder-bob/cmd/proxy.go
```diff
@@ -92,6 +108,6 @@ func init() {
 	// These env vars start with `WORKSPACEKIT_` so that they aren't passed on to ring2
 	proxyCmd.Flags().StringVar(&proxyOpts.BaseRef, "base-ref", os.Getenv("WORKSPACEKIT_BOBPROXY_BASEREF"), "ref of the base image")
 	proxyCmd.Flags().StringVar(&proxyOpts.TargetRef, "target-ref", os.Getenv("WORKSPACEKIT_BOBPROXY_TARGETREF"), "ref of the target image")
-	proxyCmd.Flags().StringVar(&proxyOpts.Auth, "auth", os.Getenv("WORKSPACEKIT_BOBPROXY_AUTH"), "authentication to use")
-	proxyCmd.Flags().StringVar(&proxyOpts.AdditionalAuth, "additional-auth", os.Getenv("WORKSPACEKIT_BOBPROXY_ADDITIONALAUTH"), "additional authentication to use")
+	proxyCmd.Flags().StringVar(&proxyOpts.BaseAuth, "base-auth", os.Getenv("WORKSPACEKIT_BOBPROXY_AUTH"), "authentication to use for base ref")
```
sagor999, line 111: would this break any existing installs by any chance? (changing command line param names)

princerachit: I checked ops repo and gitpod repo but could not find any references. AFAIK image-builder-bob proxy is only run in the imagebuild pods which is started by image-builder-mk3. Since image-builder-mk3 it self does not use these flags, it is unlikely that these flags are used anywhere else.

sagor999 (approved):

princerachit:
/unhold

I am going to merge this PR as some members of the team are at kubecon and it can take them few more days to come back. If this breaks anything we can revert this.

PR #10101, opened 2022-05-18
Fix to stopping and snapshot state

### LINKED ISSUE #10054: [prebuild] prebuilds are marked as error even though they are in ready state
### Bug description

Here is a loom video showing an issue:
https://www.loom.com/share/55988106aa8a4fc39544a53a0611daaa

Here is a description from: https://github.com/gitpod-io/gitpod/issues/9710#issuecomment-1128221314
```
while trying to repro this case, I noticed an odd behaviour that I think is due to a fix for 
https://github.com/gitpod-io/gitpod/issues/9676 :
when prebuild has finished, and snapshot is being taken, prebuild will show up in dashboard in ERROR state. 
After snapshot was taken, it will get updated into READY state. But when I go into list of prebuilds,
it still shows up there as in ERROR state.
```

### Steps to reproduce

follow the steps in loom video

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

laushinka:
## Description
1. Removes error when state is available and snapshot is available.
2. Removes the assumption that a stopping phase without a snapshot is an error. This is most likely an intermediate state where it will turn into something else, therefore we do nothing here.

## Related Issue(s)

Fixes #10054 

## How to test
In the preview environment, add a project and run a prebuild.
Once `Ready`, it should not switch to `Failed`.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/ws-manager-bridge/src/prebuild-state-mapper.spec.ts
+++ b/components/ws-manager-bridge/src/prebuild-state-mapper.spec.ts
@@ -53,11 +53,8 @@ class TestPrebuildStateMapper {
                 },
             },
             {
-                name: "failed - no snapshot",
-                expected: {
-                    state: "failed",
-                    hasError: true,
-                },
+                name: "Stopping and no snapshot yet",
+                expected: undefined,
                 status: {
                     id,
                     phase: WorkspacePhase.STOPPING,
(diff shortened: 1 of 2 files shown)


AlexTugarev (approved):
LGTM!

PR #10107, opened 2022-05-18
[JetBrains] Show notification when port becomes available 🔔

### LINKED ISSUE #9908: Handle port notifications in JetBrains
### Summary

Show a notification in JetBrains IDEs when a port is ready and exposed

### Context

At the moment, `open-preview` and `open-browser` do not work automatically. The user can run `gp url <PORT>` and get the URL manually, however there is no information or notification about the port/URL being ready.

This is a feature that we have in VS Code, so we should handle supervisor notifications also via JetBrains Gateway

### Value

Users are aware that a port has been exposed and their preview URL can be opened

### Acceptance Criteria

A notification appears for each port being exposed

### Measurement

Not incoming support request about how to open the preview URL from JetBrains Gateway

### Out of scope

* Implementing a ports view, or updates to gp CLI to show port information

andreafalzetti:
## Description
This PR introduces ports notification in JetBrains IDEs. 🔔 

### Context
Currently, when ports become available (e.g. a web server is ready to handle requests and the port is exposed to the internet), in VS Code you get a toast notification within the IDE, to inform you. Additionally, you get an action within the notification to open the browser or the preview (specifically for VS Code, as it supports preview windows). While, for JetBrains IDEs, we didn't yet show any notification when ports became available.

Example notification in VS Code

Screenshot 2022-05-25 at 15 41 49

### Approach

The approach used is to listen to supervisor ports status updates (see [status.proto](https://github.com/gitpod-io/gitpod/blob/main/components/supervisor-api/status.proto)).

Gitpod's JetBrains backend-plugin already integrates with supervisor generic notification, so this feature leverages the existing integration and re-uses the same notification group.

After this change, the notifications will appear in the native notifications panel. See example below:

Screenshot 2022-05-25 at 16 35 29

Additionally, we offer to "Make Public" a port when its visibility is _private_.

Screenshot 2022-05-26 at 22 00 33

Notes
- To avoid noise and undesired notifications, certain ports are explicitly ignored. Most of those are ports used internally by the IDE itself.

## Related Issue(s)

Fixes #9908 

## How to test
1. Start a workspace using the following link: https://afalz-9908c5dfc24116.staging.gitpod-dev.com/#https://github.com/andreafalzetti/gitpod-experiments-b
1. Choose a JetBrains product as your preferred editor
1. Notice a prompt to open the browser on port 8080
1. Notice a notification for port 8081 being available
1. Notice a notification for port 8083 being available (set to be `notify` in `.gitpod.yml`)
1. Port 8082 should not trigger a notification (set to be ignored in `.gitpod.yml`)
1. Try exposing a new port, make it public, verify that it's accessible

## Release Notes

```release-note
[JetBrains] Show notification when port becomes available
```

## Documentation

--- 
- [ ] /werft with-vm
- [x] /werft without-vm

### DIFF
--- a/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodClientProjectSessionTracker.kt
+++ b/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodClientProjectSessionTracker.kt
@@ -4,6 +4,12 @@
 
 package io.gitpod.jetbrains.remote
 
+import com.intellij.codeWithMe.ClientId
+import com.intellij.ide.BrowserUtil
+import com.intellij.idea.StartupUtil
+import com.intellij.notification.NotificationAction
+import com.intellij.notification.NotificationType
+import com.intellij.openapi.Disposable
 import com.intellij.openapi.application.ApplicationInfo
 import com.intellij.openapi.client.ClientProjectSession
 import com.intellij.openapi.components.service
@@ -11,27 +17,178 @@ import com.intellij.openapi.diagnostic.thisLogger
 import com.intellij.openapi.fileEditor.FileEditorManagerEvent
 import com.intellij.openapi.fileEditor.FileEditorManagerListener
 import com.intellij.openapi.fileTypes.LanguageFileType
+import com.intellij.remoteDev.util.onTerminationOrNow
+import com.intellij.util.application
+import com.jetbrains.rd.util.lifetime.Lifetime
 import io.gitpod.gitpodprotocol.api.entities.RemoteTrackMessage
+import io.gitpod.gitpodprotocol.api.entities.WorkspaceInstancePort
 import io.gitpod.supervisor.api.Info
-import kotlinx.coroutines.GlobalScope
+import io.gitpod.supervisor.api.Status
+import io.gitpod.supervisor.api.Status.PortVisibility
+import io.gitpod.supervisor.api.Status.PortsStatus
+import io.gitpod.supervisor.api.StatusServiceGrpc
+import io.grpc.stub.ClientCallStreamObserver
+import io.grpc.stub.ClientResponseObserver
+import kotlinx.coroutines.*
 import kotlinx.coroutines.future.await
-import kotlinx.coroutines.launch
+import org.jetbrains.ide.BuiltInServerManager
+import java.util.concurrent.CancellationException
+import java.util.concurrent.CompletableFuture
 
 class GitpodClientProjectSessionTracker(
         private val session: ClientProjectSession
-) {
+) : Disposable {
 
     private val manager = service<GitpodManager>()
 
     private lateinit var info: Info.WorkspaceInfoResponse
     private val versionName = ApplicationInfo.getInstance().versionName
     private val fullVersion = ApplicationInfo.getInstance().fullVersion
+    private val lifetime = Lifetime.Eternal.createNested()
+
+    override fun dispose() {
+        lifetime.terminate()
+    }
 
     init {
         GlobalScope.launch {
             info = manager.pendingInfo.await()
             trackEvent("jb_session", mapOf())
             registerActiveLanguageAnalytics()
+        }
+    }
+
+    private fun isExposedServedPort(port: Status.PortsStatus?) : Boolean {
+        if (port === null) {
+            return false
+        }
+        return port.served && port.hasExposed()
+    }
+
+    private fun showOpenServiceNotification(port: PortsStatus, offerMakePublic: Boolean = false) {
+        val message = "A service is available on port ${port.localPort}"
+        val notification = manager.notificationGroup.createNotification(message, NotificationType.INFORMATION)
+
+        val openBrowserAction = NotificationAction.createSimple("Open Browser") {
+            openBrowser(port.exposed.url)
+        }
+        notification.addAction(openBrowserAction)
+
+        if (offerMakePublic) {
+            val makePublicLambda = {
+                runBlocking {
+                    makePortPublic(info.workspaceId, port)
+                }
+            }
+            val makePublicAction = NotificationAction.createSimple("Make Public", makePublicLambda)
+            notification.addAction(makePublicAction)
+        }
+
+        ClientId.withClientId(session.clientId) {
+            notification.notify(null)
+        }
+    }
+
+    private suspend fun makePortPublic(workspaceId: String, port: PortsStatus) {
+        val p = WorkspaceInstancePort()
+        p.port = port.localPort
+        p.visibility = io.gitpod.gitpodprotocol.api.entities.PortVisibility.PUBLIC.toString()
+        p.url = port.exposed.url
+
+        try {
+            manager.client.server.openPort(workspaceId, p).await()
+        } catch (e: Exception) {
+            thisLogger().error("gitpod: failed to open port ${port.localPort}: ", e)
+        }
+    }
+
+    private fun openBrowser(url: String) {
+        ClientId.withClientId(session.clientId) {
+            BrowserUtil.browse(url)
+        }
+    }
+
+    private val portsObserveJob = GlobalScope.launch {
+        if (application.isHeadlessEnvironment) {
+            return@launch
+        }
+
+        // Ignore ports that aren't actually used by the user (e.g. ports used internally by JetBrains IDEs)
+        val backendPort = BuiltInServerManager.getInstance().waitForStart().port
+        val serverPort = StartupUtil.getServerFuture().await().port
+        val ignorePorts = listOf(backendPort, serverPort, 5990)
+        val portsStatus = hashMapOf<Int, Status.PortsStatus>()
+
+        val status = StatusServiceGrpc.newStub(GitpodManager.supervisorChannel)
+        while (isActive) {
+            try {
+                val f = CompletableFuture<Void>()
+                status.portsStatus(
+                        Status.PortsStatusRequest.newBuilder().setObserve(true).build(),
+                        object : ClientResponseObserver<Status.PortsStatusRequest, Status.PortsStatusResponse> {
+
+                            override fun beforeStart(requestStream: ClientCallStreamObserver<Status.PortsStatusRequest>) {
+                                lifetime.onTerminationOrNow {
+                                    requestStream.cancel(null, null)
+                                }
+                            }
+
+                            override fun onNext(ps: Status.PortsStatusResponse) {
+                                for (port in ps.portsList) {
+                                    // Avoiding undesired notifications
+                                    if (ignorePorts.contains(port.localPort)) {
+                                        continue
+                                    }
+
+                                    val previous = portsStatus[port.localPort]
+                                    portsStatus[port.localPort] = port
+
+                                    val shouldSendNotification = !isExposedServedPort(previous) && isExposedServedPort(port)
+
+                                    if (shouldSendNotification) {
+                                        if (port.exposed.onExposed.number == Status.OnPortExposedAction.ignore_VALUE) {
+                                            continue
+                                        }
+
+                                        if (port.exposed.onExposed.number == Status.OnPortExposedAction.open_browser_VALUE || port.exposed.onExposed.number == Status.OnPortExposedAction.open_preview_VALUE) {
+                                            openBrowser(port.exposed.url)
+                                            continue
+                                        }
+
+                                        if (port.exposed.onExposed.number == Status.OnPortExposedAction.notify_VALUE) {
+                                            showOpenServiceNotification(port)
+                                            continue
+                                        }
+
+                                        if (port.exposed.onExposed.number == Status.OnPortExposedAction.notify_private_VALUE) {
+                                            showOpenServiceNotification(port, port.exposed.visibilityValue !== PortVisibility.public_visibility_VALUE)
+                                            continue
+                                        }
+                                    }
+                                }
+                            }
+
+                            override fun onError(t: Throwable) {
+                                f.completeExceptionally(t)
+                            }
+
+                            override fun onCompleted() {
+                                f.complete(null)
+                            }
+                        })
+                f.await()
+            } catch (t: Throwable) {
+                if (t is CancellationException) {
+                    throw t
+                }
+                thisLogger().error("gitpod: failed to stream ports status: ", t)
+            }
+            delay(1000L)
+        }
+    }
+    init {
+        lifetime.onTerminationOrNow {
+            portsObserveJob.cancel()
         }
     }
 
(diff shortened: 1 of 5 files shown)


roboquat:

<details>

</details>

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodManager.kt
```diff
@@ -199,6 +205,107 @@ class GitpodManager : Disposable {
...
+    private val portsObserveJob = GlobalScope.launch {
+        if (application.isHeadlessEnvironment) {
+            return@launch
+        }
+
+        val ignorePorts = listOf(5990, 6942)
+        val portsStatus = hashMapOf<Int, PortsStatus>()
+
+        val status = StatusServiceGrpc.newStub(supervisorChannel)
+        //while (isActive) {
```
andreafalzetti, line 217: @akosyakov if I keep this `while` - it creates for each loop a new subscription to supervisor and soon it starts throwing errors for too many concurrent subscriptions. Are we sure I need it?

akosyakov: you should call `f.await()` as here https://github.com/gitpod-io/gitpod/blob/641adda863bbe7c904f34f8d94b6e866f89d9feb/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodManager.kt#L186

akosyakov: Please make sure to align with how we listen to notifications 🙏 it is done right there

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodManager.kt
```diff
@@ -199,6 +205,105 @@ class GitpodManager : Disposable {
...
+                                val shouldSendNotification = isFirstUpdate || !wasServedExposed && isServedExposed
+
+                                portsStatus[port.localPort] = port
+
+                                if (shouldSendNotification) {
+                                    if (port.exposed.onExposed.number == OnPortExposedAction.ignore_VALUE) {
+                                        continue
+                                    }
+
+                                    if (port.exposed.onExposed.number == OnPortExposedAction.open_browser_VALUE) {
```
andreafalzetti, line 262: Unfortunately when the IDE is still loading, this api call to open the browser gets lost. If someone has an idea on how to delay/persist it - until the IDE is ready - let me know. Otherwise I would suggest to move forward and ask our friends @ JetBrains for suggestions.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodManager.kt
```diff
@@ -199,6 +205,105 @@ class GitpodManager : Disposable {
...
+                                    }
+
+                                    if (port.exposed.onExposed.number == OnPortExposedAction.open_preview_VALUE) {
+                                        BrowserUtil.browse(port.exposed.url)
+                                        continue
+                                    }
+
+                                    val message = "A service is available on port ${port.localPort}"
+                                    val notification = notificationGroup.createNotification(message, NotificationType.INFORMATION)
+                                    // TODO(andreafalzetti): add analytics event similar to https://github.com/gitpod-io/openvscode-server/blob/7c912399c70799e4e4452a63cf3c0f7cc8f5832c/extensions/gitpod-web/src/extension.ts#L582
```
andreafalzetti, line 274: I couldn't find a setup for analytics in this component. Am I missing something or we don't have it yet?

mustard-mh: It is here https://github.com/gitpod-io/gitpod/blob/a60fc2d2a632651aab4ace5532515e4cc02f30ae/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodClientProjectSessionTracker.kt#L62

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodManager.kt
```diff
@@ -199,6 +205,105 @@ class GitpodManager : Disposable {
         }
     }
 
+    private val portsObserveJob = GlobalScope.launch {
+        if (application.isHeadlessEnvironment) {
+            return@launch
+        }
+
+        // Ignore ports that aren't actually used by the user (e.g. ports used internally by JetBrains IDEs)
+        val ignorePorts = listOf(5990, 5991, 6679, 6942, 6943)
```
akosyakov, line 214: I think we should only ignore a port of JB backend itself.

You can get it like that: https://github.com/gitpod-io/gitpod/blob/fea2135ad5e85757b510665710c9a6bfdbe3c734/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodManager.kt#L106

you can add this port + 1 as well if JB_DEV is true

We can ignore only if we know something from us is running otherwise we should not exlude a case that a user app will do it.

Do you know who is owning 5990, 5991, 6679?

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodManager.kt
akosyakov: it sounds like we should listen for each project session (window) separately in https://github.com/gitpod-io/gitpod/blob/main/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodClientProjectSessionTracker.kt#L20

And execute actions in client closure using:
```
ClientId.withClientId(session.clientId) {
  send notification or open browser
}
```

andreafalzetti: it worked! thanks @akosyakov 🙌

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodManager.kt
andreafalzetti: oooh! thanks @mustard-mh :) I've added analytics now 🚀

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodManager.kt
andreafalzetti: @akosyakov I've added those to the ignore list because as soon as I start `./launch-dev-server.sh` and join the session, I get a notification for each of those ports. 

Screenshot 2022-05-26 at 08 38 19

6942 seems related to JetBrains too [[1]](https://intellij-support.jetbrains.com/hc/en-us/community/posts/360003708940-PyCharm-binds-to-a-port-that-I-need-to-use-)

I think 5990, 5991 are related to CodeWithMe [[1]](https://www.jetbrains.com/help/idea/faq-about-code-with-me-security.html#communication_ports)

`6679` indeed I wasn't able to find any reference today - so I've removed it already.

I have a feeling that most users will be annoyed by these notifications if we don't ignore them. We could suggest to add `onOpen: ignore` but it feels an extra effort for the user.

akosyakov: Do you get them in preview envs as well?

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodManager.kt
andreafalzetti: We agreed with @akosyakov to remove the analytics event for now. It costs us $$$ and if we are not actively using it, it only causes noise.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodClientProjectSessionTracker.kt
```diff
@@ -35,6 +58,140 @@ class GitpodClientProjectSessionTracker(
...
+    }
+
+    private suspend fun makePortPublic(workspaceId: String, port: PortsStatus) {
+        val p = WorkspaceInstancePort()
+        p.port = port.localPort
+        p.visibility = io.gitpod.gitpodprotocol.api.entities.PortVisibility.PUBLIC.toString()
+        p.url = port.exposed.url
+
+        try {
+            manager.client.server.openPort(workspaceId, p).await()
```
andreafalzetti, line 100: at the moment, pressing the button doesn't update the UI (in VS Code in the ports view you can see the text "private" change into "public"), so it looks a bit unfinished. We could consider adding an extra notification like:

```kotlin
val message = "The port ${port.localPort} is now public"
val notification = manager.notificationGroup.createNotification(message, NotificationType.INFORMATION)
ClientId.withClientId(session.clientId) {
   val project = RestService.getLastFocusedOrOpenedProject()
   notification.notify(project)
}
```

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodManager.kt
andreafalzetti: > Do you get them in preview envs as well?

No. Removed as agreed. I've only left the IDE backend port 🙏

akosyakov:
@felladrin I see these ports too:
Screenshot 2022-05-27 at 09 05 05

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodClientProjectSessionTracker.kt
```diff
@@ -35,6 +58,140 @@ class GitpodClientProjectSessionTracker(
...
+                runBlocking {
+                    makePortPublic(info.workspaceId, port)
+                }
+            }
+            val makePublicAction = NotificationAction.createSimple("Make Public", makePublicLambda)
+            notification.addAction(makePublicAction)
+        }
+
+        ClientId.withClientId(session.clientId) {
+            val project = RestService.getLastFocusedOrOpenedProject()
```
akosyakov, line 88: you need to find a project associated with session, see how it is done in cli

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodClientProjectSessionTracker.kt
akosyakov: Too many notification is also not nice. Is it possible to update an action? Maybe just make it expirable?

akosyakov:
If I close a client and open again, I don't see notifications it is something with state management.

akosyakov:
@andreafalzetti 5990 is a default port for SSH connection from JB GW. You can hardcore it as to ignore.

I'm not sure why built-in server port is not ignored. It seems you are waiting properly for it.

andreafalzetti:
> @andreafalzetti 5990 is a default port for SSH connection from JB GW. You can hardcore it as to ignore.
> 
> I'm not sure why built-in server port is not ignored. It seems you are waiting properly for it.

I think `6942` is also IntelliJ as I mentioned previously.. There is mention of it [here](https://github.com/JetBrains/intellij-community/blob/01310124331f179a8e872399b4805d4a861a5e95/platform/platform-impl/src/com/intellij/idea/SocketLock.java#L150) (not 100% sure it's the same port we are talking about tho)

Also, the backend port is `63343`. It's the port that I get from `BuiltInServerManager.getInstance().waitForStart().port`. And I never got notification about that one.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodClientProjectSessionTracker.kt
andreafalzetti: Let's do a step back, in [GitpodManager.kt#L175](https://github.com/gitpod-io/gitpod/blob/main/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodManager.kt#L175), the function `notification.notify(null)` receives `null` as `projectId`. 

Initially, I copied that approach, but given that we have the project, I thought it would be better to pass it, so that each session will receive the notification properly. I think it must be working anyway with `null` because we only use 1 session, perhaps.

Anyway, the retrieve of the project, I got it from [GitpodCLIService.kt](https://github.com/gitpod-io/gitpod/blob/main/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt#L72).

So, I am not sure what to change. 

1. Shall we keep `null`?
2. If not, can you point me to where I can get a project associated with session? I assumed `getLastFocusedOrOpenedProject` would do that 🤔

akosyakov: does `session.project` work? 

if not let's keep `null`

akosyakov:
> I think 6942 is also IntelliJ as I mentioned previously. T[here](https://github.com/JetBrains/intellij-community/blob/01310124331f179a8e872399b4805d4a861a5e95/platform/platform-impl/src/com/intellij/idea/SocketLock.java#L150) is mention of it here (not 100% sure it's the same port we are talking about tho)

then you should also add something like to ignore it
```
const serverPort = StartupUtil.getServerFuture().waitForStart().port
```

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodClientProjectSessionTracker.kt
andreafalzetti: `session.project` seems to work! Thanks 🙏

andreafalzetti:
> > I think 6942 is also IntelliJ as I mentioned previously. T[here](https://github.com/JetBrains/intellij-community/blob/01310124331f179a8e872399b4805d4a861a5e95/platform/platform-impl/src/com/intellij/idea/SocketLock.java#L150) is mention of it here (not 100% sure it's the same port we are talking about tho)
> 
> then you should also add something like to ignore it
> 
> ```
> const serverPort = StartupUtil.getServerFuture().await().port
> ```

Added 👍

PR #10109, opened 2022-05-18
[baseserver] Change default metrics port to 9502 to not clash with kube-rbac-proxy

easyCZ:
Normally, we expose `kube-rbac-proxy` for metrics on port 9500. With https://github.com/gitpod-io/gitpod/pull/10099 we removed listening on localhost to avoid a pitfall where listening on localhost would be inaccessible from outside the container.

In this PR, the default port is updated to not clash with 9500 which we commonly use for metrics through `kube-rbac-proxy`. Currently, only usage is by `public-api-server` (which crash-loops due to this) and `content-service` which doesn't have `kube-rbac-proxy` deployed so this change is safe.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Documentation

NONE

### DIFF
--- a/components/common-go/baseserver/server.go
+++ b/components/common-go/baseserver/server.go
@@ -332,7 +332,7 @@ func (s *Server) GRPCAddress() string { return s.options.config.Services.GRPC.Ge
 
 const (
 	BuiltinDebugPort   = 6060
-	BuiltinMetricsPort = 9500
+	BuiltinMetricsPort = 9502
 	BuiltinHealthPort  = 9501
 )
 


easyCZ:
@sagor999 for public api, the installer already uses the the cost that's changed in this PR so there's no change needed. For scraping, we parametrize the proxy url for Kube-rbac-proxy so there no change is needed either. This should work without requiring additional changes

aledbf:
We need to rollback this change. The issue here is not the collision but the Prometheus listening should happen in localhost, that way the traffic can go from `kube-rbac-proxy` to the localhost port

PR #10123, opened 2022-05-19
Terminate IDEs gracefully on workspace shutdown

### LINKED ISSUE #9955: terminate IDEs gracefully on workspace shutdown
### Bug description

Right now we send SIGKILL to all process of IDE group. Here is a feedback about it from JB folks:

> SIGKILL is the furthest from graceful shutdown one can get - there's no way to trap it and do some work before dying, it's forced process termination without giving it the chance to finish any work or clean up. I would suggest using the shell command, waiting 15 seconds for that, and then sending it SIGKILL to be sure it's dead

Instead we should call `remote-dev-server.sh exit /path/to/project`. Not doing can lead to broken indexes and it is most likely a reason why we see reindexing sometime on workspace restart.

cc @csweichel 

### Steps to reproduce

see above

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

mustard-mh:
## Description

Terminate IDEs gracefully on workspace shutdown.

Before this PR, we will send Interrupt (SIGINT) to ask IDEs to exit, and only give them 5 seconds to do it. This behavior may cause IDEs indexing failed sometimes.

To properly shutdown IDE server
- Increase time to force shutdown to 15 seconds
- For code ~~we use these `enable-remote-auto-shutdown` and `remote-auto-shutdown-without-delay` args~~ we pass SIGTERM
- For Gateway we use `exit` command

## Related Issue(s)

Fixes #9955

## How to test

For Gateway
- Preference IDE with IntelliJ IDEA
- Start workspace https://hw-ide-term-2.preview.gitpod-dev.com/#https://github.com/gitpod-io/spring-petclinic
- Exec `kubectl logs -f <workspace_pod_name> > ws.log` to get all logs of workspace
- Stop workspace after gateway index complete
- Restart that workspace
- Exec `kubectl logs -f <workspace_pod_name> > ws-2.log` to get all logs of workspace
- Export all logs of Gateway backend (see image below)

With `ws.log` we must see `asked IDE to terminate` before `IDE stopped, exiting`
With downloaded zipped log, unzip and unzip .zip file inside, you will get `idea.log`, and check indexing logs inside

image

## Release Notes

```release-note
Terminate IDEs gracefully on workspace shutdown
```

### DIFF
--- a/components/ide/jetbrains/image/startup.sh
+++ b/components/ide/jetbrains/image/startup.sh
@@ -18,4 +18,4 @@ if [ "${SUPERVISOR_DEBUG_ENABLE+}" = "true" ]; then
   export JAVA_TOOL_OPTIONS "-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:0"
 fi
 
-/ide-desktop/status "$@"
+exec /ide-desktop/status "$@"
--- a/components/ide/jetbrains/image/status/main.go
+++ b/components/ide/jetbrains/image/status/main.go
@@ -16,15 +16,18 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
+	"os/signal"
 	"path/filepath"
 	"reflect"
 	"regexp"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/hashicorp/go-version"
 	"golang.org/x/xerrors"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
 	yaml "gopkg.in/yaml.v2"
 
 	"github.com/gitpod-io/gitpod/common-go/log"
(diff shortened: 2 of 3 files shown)


jeanp413:
@mustard-mh Be sure to check if `remote-auto-shutdown-without-delay` is helpful or not, from what I've read in the code it will terminate the server immediately after closing the browser tab which may not be desired if user closed it by accident. Also check if it's possible to dispose the front-end workbench properly after the 30 min timeout

akosyakov:
@mustard-mh yes it will be good to have how to test section fro VS Code as well

mustard-mh:
> @mustard-mh yes it will be good to have how to test section for VS Code as well

I don't know how to reproduce it [[1]](https://github.com/gitpod-io/gitpod/issues/9955#issuecomment-1129854972), so don't know how to test it🥲 @akosyakov

akosyakov:
> I don't know how to reproduce it https://github.com/gitpod-io/gitpod/issues/9955#issuecomment-1129854972, so don't know how to test it🥲 @akosyakov

Could you take time to read through VS Code code and understand meaning of this options and how we can test?

akosyakov:
I don’t think `enable-remote-auto-shutdown` is useful for us. The extension host is a process which runs VS Code extensions for a given window. If we enable enable-remote-auto-shutdown  then a server will be shutdown after all windows are gone and there is no new windows within 5 mins. We don’t need this logic. We already handle it on level of workspaces and VS Code Server lifetime is bound to a workspace lifetime for us.

SIGTERM should work instead better and wait like 5s.

felladrin (approved):
Code reviewed and PR tested as described. Everything working fine! ✅

Issue #9875, opened 2022-05-09
Error: headless task failed, on imagebuild


axonasif:
### Bug description

I was trying to open this repo https://gitpod.io/#https://github.com/axonasif/wrover-experimental/tree/wokwi-test after adding a custom dockerfile
<img width="1060" alt="Screenshot 2022-05-09 at 9 47 51 PM" src="https://user-images.githubusercontent.com/39482679/167447779-5597c636-d1ca-4219-af77-b61ec55d81b4.png">

### Steps to reproduce

Open https://github.com/axonasif/wrover-experimental/tree/wokwi-test on Gitpod.

### Workspace affected

axonasif-wroverexperime-u9ik2sre8pl

### Example repository

https://github.com/axonasif/wrover-experimental/tree/wokwi-test

### Anything else?

A similar report on Discord today: https://discord.com/channels/816244985187008514/972590280446476308/973228246407733288

ChamMach:
+1, same for me 🤔

felipealencar:
Same problem here.

ChamMach:
Very surprised that there aren't more people complaining about it. Although it's about viewing the outputs/logs, for me, it becomes *complicated* not knowing exactly where the problem is

felipealencar:
Hello everyone, Idk who still is facing this issue, but I could see detailed information about my error by:
1. Selecting the option to run a default workspace;
2. Running my custom `.gitpod.Dockerfile` in a test:
`docker build -f .gitpod.Dockerfile -t gitpod-dockerfile-test .`
3. Finally, I could see the error messages related to the problems of my Dockerfile.

I hope it helps someone.

geropl:
Scheduled this, as the user experience feels really broken :confused: .

@felipealencar Thank you for sharing the workaround! This is indeed what's happening: The `headless task failed: exit status 1` error is always emitted if there's some user-error with the image build.
That could be anything from error pulling the image, to a failing Docker command inside the docker file.

ChamMach:
Yep, thanks for the workaround! I realized it too by simply building the *problematic Dockerfile* on another gitpod.

Thanks to the team for taking this issue into account too! 🙏

jankeromnes:
Maybe this error is related? (internal) https://console.cloud.google.com/errors/detail/CNif-a-gteWfNw;service=server;version=main.3268;time=PT1H?project=gitpod-191109

csweichel:
[...]

PR #10127, opened 2022-05-19
[installer, ws-manager]: Use installation shortname in URL templates

andrew-farries:
## Description

As described in https://github.com/gitpod-io/gitpod/issues/9875, failing image builds currently fail to produce any useful output. A possible cause, as described in [this comment](https://github.com/gitpod-io/gitpod/issues/9875#issuecomment-1129924819), is that the format of the `ws-manager` `WorkspaceURLTemplate` strings is wrong. 

When we deployed with the old helm chart, these values were set in the relevant configmap like this:

https://github.com/gitpod-io/gitpod/blob/2f988c93499926bd34d6728defbd9d9c0964797e/chart/templates/ws-manager-configmap.yaml#L74-L75

ie, they included a suffix after the `ws`. 

This PR tailors the installer output to produce these URLs in the same way the old helm chart did. Installation short name suffixes like `eu-02` are set here:

https://github.com/gitpod-io/ops/blob/main/deploy/production/meta-eu02/app/installer-config.yaml#L6

so the URLs will now look like `https://{{ .Prefix }}.ws-eu02.example.com`.

## Related Issue(s)
Fixes https://github.com/gitpod-io/gitpod/issues/9875

## How to test

I've tested this change in staging by editing the `ws-manager` configmap in staging to contain the correct URLs, restarting `ws-manager` and running an image build. I was able to see image build logs as expected:

image

## Release Notes
```release-note
[installer] Use installation shortname when constructing ws-manager URL templates
```

## Documentation
None

### DIFF
--- a/install/installer/pkg/components/ws-manager/configmap.go
+++ b/install/installer/pkg/components/ws-manager/configmap.go
@@ -113,6 +113,11 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		return nil, err
 	}
 
+	installationShortNameSuffix := ""
+	if ctx.Config.Metadata.InstallationShortname != "" {
+		installationShortNameSuffix = "-" + ctx.Config.Metadata.InstallationShortname
+	}
+
 	wsmcfg := config.ServiceConfiguration{
 		Manager: config.Configuration{
 			Namespace:      ctx.Namespace,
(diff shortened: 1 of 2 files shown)


sagor999 (approved):

sagor999:
@andrew-farries I think this change broke preview env. 
https://gitpod.slack.com/archives/C032A46PWR0/p1652996871215559

PR #10129, opened 2022-05-19
Wait the Guest Client to be connected to the workspace before executing "gp preview" on JetBrains IDEs.

### LINKED ISSUE #10118: [jb] gp preview fails if running from .gitpod.yml
### Bug description

See https://github.com/jenkinsci/design-library-plugin/pull/66#discussion_r876189440

The issue is that JB backend is not running yet. We should await always instead.

### Steps to reproduce

See above.

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

felladrin:
## Description

Wait the Guest Client to be connected to the workspace before executing "gp preview" on JetBrains IDEs.

## Related Issue(s)

Fixes #10118

## How to test

- Access the following URL to open Jenkins Design Library in Gitpod's Preview Environment in IntelliJ IDEA:
  https://vn-10118-j8f802b11df.preview.gitpod-dev.com/#referrer:jetbrains-gateway:intellij/https://github.com/jenkinsci/design-library-plugin/tree/73f12733d95932c333ba4155521ff78b4ed40dfc
- As soon as the IDE opens, it stats building the project. Meanwhile, if you want, you can use `gp tasks attach` to attach to "Preview" task. (See the project's [.gitpod.yml](https://github.com/jenkinsci/design-library-plugin/blob/73f12733d95932c333ba4155521ff78b4ed40dfc/.gitpod.yml))
- After the build completes, port 8080 will be available, and at this moment, the IDE should ask if you want to open the browser at the address `$(gp url 8080)/jenkins/`.
    image
    image
    Note: Sometimes the project build fails with messages like `java: error reading /workspace/m2-repository/jaxen/jaxen/1.2.0/jaxen-1.2.0.jar; zip file is empty` and the only solution is to manually delete the dependency folder (`rm -rf /workspace/m2-repository/jaxen/`) and then run `mvn hpi:run` again.
- Stop the workspace and restart it. Now the build will happen pretty quickly, as all dependencies are already downloaded. So as soon as you open the workspace on IntelliJ you should see the popup asking if you want to open Jenkins URL in the browser.

## Release Notes

```release-note
Fixed a bug that was preventing "gp preview" to work in JetBrains IDEs when executed as a task in ".gitpod.yml".
```

## Documentation

None.

### DIFF
--- a/components/ide/jetbrains/backend-plugin/launch-dev-server.sh
+++ b/components/ide/jetbrains/backend-plugin/launch-dev-server.sh
@@ -39,4 +39,14 @@ export IJ_HOST_SYSTEM_BASE_DIR=/workspace/.cache/JetBrains
 # Enable host status endpoint
 export CWM_HOST_STATUS_OVER_HTTP_TOKEN=gitpod
 
+# Build and move idea-cli, then overwrite environment variables initially defined by `components/ide/jetbrains/image/leeway.Dockerfile`
+IDEA_CLI_DEV_PATH=/ide-desktop/bin/idea-cli-dev
+(cd ../cli && go build -o $IDEA_CLI_DEV_PATH)
+export EDITOR="$IDEA_CLI_DEV_PATH open"
+export VISUAL="$EDITOR"
+export GP_OPEN_EDITOR="$EDITOR"
+export GIT_EDITOR="$EDITOR --wait"
+export GP_PREVIEW_BROWSER="$IDEA_CLI_DEV_PATH preview"
+export GP_EXTERNAL_BROWSER="$IDEA_CLI_DEV_PATH preview"
+
 $TEST_BACKEND_DIR/bin/remote-dev-server.sh run $TEST_DIR
--- a/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
+++ b/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
@@ -7,6 +7,7 @@ package io.gitpod.jetbrains.remote
 import com.intellij.codeWithMe.ClientId
 import com.intellij.ide.BrowserUtil
 import com.intellij.ide.CommandLineProcessor
+import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.client.ClientSession
 import com.intellij.openapi.client.ClientSessionsManager
 import com.intellij.openapi.diagnostic.thisLogger
@@ -20,7 +21,7 @@ import org.jetbrains.ide.RestService
 import java.nio.file.InvalidPathException
 import java.nio.file.Path
 
-
+@Suppress("UnstableApiUsage")
 class GitpodCLIService : RestService() {
 
     override fun getServiceName() = SERVICE_NAME
@@ -54,22 +55,34 @@ class GitpodCLIService : RestService() {
     }
 
     private fun withClient(request: FullHttpRequest, context: ChannelHandlerContext, action: (project: Project?) -> Unit): String? {
+        ApplicationManager.getApplication().executeOnPooledThread {
+            getClientSessionAndProjectAsync().let { (session, project) ->
+                ClientId.withClientId(session.clientId) {
+                    action(project)
+                    sendOk(request, context)
+                }
+            }
+        }
+        return null
+    }
+
+    private data class ClientSessionAndProject(val session: ClientSession, val project: Project?)
+
+    private tailrec fun getClientSessionAndProjectAsync(): ClientSessionAndProject {
         val project = getLastFocusedOrOpenedProject()
         var session: ClientSession? = null
         if (project != null) {
-            session = ClientSessionsManager.getProjectSessions(project, false).first()
+            session = ClientSessionsManager.getProjectSessions(project, false).firstOrNull()
         }
         if (session == null) {
-            session = ClientSessionsManager.getAppSessions(false).first()
+            session = ClientSessionsManager.getAppSessions(false).firstOrNull()
         }
-        if (session == null) {
-            return "no client"
+        return if (session != null) {
+            ClientSessionAndProject (session, project)
+        } else {
+            Thread.sleep(1000L)
+            getClientSessionAndProjectAsync()
         }
-        ClientId.withClientId(session.clientId) {
-            action(project)
-        }
-        sendOk(request, context)
-        return null
     }
 
     private fun parseFilePath(path: String): Path? {
--- a/components/ide/jetbrains/cli/cmd/root.go
+++ b/components/ide/jetbrains/cli/cmd/root.go
@@ -5,7 +5,11 @@
 package cmd
 
 import (
+	"errors"
+	"log"
+	"net/url"
 	"os"
+	"strconv"
 
 	"github.com/spf13/cobra"
 )
@@ -19,6 +23,18 @@ func Execute() {
 	if err != nil {
 		os.Exit(1)
 	}
+}
+
+func getCliApiUrl() *url.URL {
+	var backendPort = 63342
+	if _, fileStatError := os.Stat("/ide-desktop/bin/idea-cli-dev"); !errors.Is(fileStatError, os.ErrNotExist) {
+		backendPort = backendPort + 1
+	}
+	parsedUrl, urlParseError := url.Parse("http://localhost:" + strconv.Itoa(backendPort) + "/api/gitpod/cli")
+	if urlParseError != nil {
+		log.Fatal(urlParseError)
+	}
+	return parsedUrl
 }
 
 func init() {}
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -2,6 +2,15 @@
 <project version="4">
   <component name="CompilerConfiguration">
     <bytecodeTargetLevel>
+      <module name="io.gitpod.api.jetbrains-backend-plugin.gitpod-protocol" target="11" />
+      <module name="io.gitpod.api.jetbrains-backend-plugin.gitpod-protocol.main" target="11" />
+      <module name="io.gitpod.api.jetbrains-backend-plugin.gitpod-protocol.test" target="11" />
+      <module name="io.gitpod.api.jetbrains-backend-plugin.supervisor-api" target="11" />
+      <module name="io.gitpod.api.jetbrains-backend-plugin.supervisor-api.main" target="11" />
+      <module name="io.gitpod.api.jetbrains-backend-plugin.supervisor-api.test" target="11" />
+      <module name="io.gitpod.jetbrains.jetbrains-backend-plugin" target="11" />
+      <module name="io.gitpod.jetbrains.jetbrains-backend-plugin.main" target="11" />
+      <module name="io.gitpod.jetbrains.jetbrains-backend-plugin.test" target="11" />
       <module name="jetbrains-backend-plugin" target="11" />
       <module name="jetbrains-backend-plugin.gitpod-protocol" target="11" />
       <module name="jetbrains-backend-plugin.gitpod-protocol.main" target="11" />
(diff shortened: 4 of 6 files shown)


felladrin:
I've noticed that if I stop the workspace, start it again and run `gp tasks attach` and select the `Preview` task, I always see the "500 Internal Server Error" message:
image

felladrin:
> I've noticed that if I stop the workspace, start it again and run `gp tasks attach` and select the `Preview` task, I always see the "500 Internal Server Error" message: image

This has been fixed by the [changes on GitpodCLIService.kt](https://github.com/gitpod-io/gitpod/pull/10129/files#diff-81d1cb74ab7142c7fba14c9bb140bb6a38eedf7d9e72a5f10065b130f95d2801)

components/ide/jetbrains/cli/cmd/root.go
```diff
@@ -15,10 +17,21 @@ var rootCmd = &cobra.Command{
...
+	waitUntilBackendPluginIsAccessible()
 	err := rootCmd.Execute()
 	if err != nil {
 		os.Exit(1)
 	}
 }
 
+func waitUntilBackendPluginIsAccessible() {
+	for {
+		resp, httpError := http.Get("http://localhost:63342/api/gitpod/cli?op=ping")
```
akosyakov, line 29: It can race no? after successful ping a client can drop and then command will fail

why don't we do a request which does not timeout and resolve await on server only?

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
```diff
@@ -46,9 +52,20 @@ class GitpodCLIService : RestService() {
             if (url.isNullOrBlank()) {
                 return "url is missing"
             }
-            return withClient(request, context) { project ->
-                BrowserUtil.browse(url, project)
+            GlobalScope.launch {
```
akosyakov, line 55: all client based requests should await client, i.e open as well

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
```diff
@@ -46,9 +52,20 @@ class GitpodCLIService : RestService() {
...
+                val controllerClientId = getControllerClientIdAsync()
+                runInEdt {
+                    ClientId.withClientId(controllerClientId) {
+                        BrowserUtil.browse(url)
+                    }
+                }
             }
+            sendOk(request, context)
```
akosyakov, line 63: does it really work with open --wait?

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
```diff
@@ -85,7 +102,17 @@ class GitpodCLIService : RestService() {
         }
     }
 
+    private tailrec suspend fun getControllerClientIdAsync(): ClientId? {
```
akosyakov, line 105: we should follow logic in withClient to get last focused or opened client, why don't we poll like each 1s with it?

akosyakov:
Should not we just change withClient that it loops till client is found and waits for 1s? It should be cancellable in case if incoming request is cancelled.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
felladrin: I think we should look into "gp open" in another PR, otherwise the "How to Test" section will become gigantic.

We also need another repo to test, cause Jenkins doesn't use "gp open" in their .gitpod.yml tasks.

(I suspect we don't need to change the "gp open" command as the file can be opened on the backend before the Thin Client connects. But we really need to investigate and confirm it.)

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
felladrin: Cross-posting: https://github.com/gitpod-io/gitpod/pull/10129#discussion_r881477632

components/ide/jetbrains/cli/cmd/root.go
akosyakov: adding port as a parameter makes sense for testing, you don't need prev env for that

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
akosyakov: There are only 3 scenarios:
- open
- open --wait
- preview

I don't think we need a special handling for each but one holistic solution about how to wait for a client.

felladrin: Got it.

I'll create a test-repo that has all those command in gitpod-tasks and use in place of Jenkins. Should also allow us to test the PR faster.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
felladrin: Updated!

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
akosyakov: To make sure you are trying to open files within project? 
There is a bug that it is not possible to open files outside of the project.

components/ide/jetbrains/cli/cmd/root.go
felladrin: Added the support for it to work on Dev Server.
I've also added one line to `launch-dev-server.sh`.

components/ide/jetbrains/backend-plugin/launch-dev-server.sh
```diff
@@ -39,4 +39,7 @@ export IJ_HOST_SYSTEM_BASE_DIR=/workspace/.cache/JetBrains
 # Enable host status endpoint
 export CWM_HOST_STATUS_OVER_HTTP_TOKEN=gitpod
 
-$TEST_BACKEND_DIR/bin/remote-dev-server.sh run $TEST_DIR
\ No newline at end of file
+# Build and move ide-cli to its respective place.
+(cd ../cli && go build && mv cli /ide-desktop/bin/idea-cli)
```
akosyakov, line 43: why not update PATH instead?

felladrin: 💡 Good idea!
I'll update it.

I'll move it to `/ide-desktop/bin/idea-cli-dev` and then update the PATH, just for the file don't be listed as a newly-added-file-ready-to-commit to git.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
felladrin: Finally fixed the "gp open --wait" 🎉
Needed to wrap it with `ApplicationManager.getApplication().executeOnPooledThread`.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
```diff
@@ -53,22 +54,31 @@ class GitpodCLIService : RestService() {
...
+        } else {
+            Thread.sleep(1000L)
+            getClientSessionAsync()
         }
-        ClientId.withClientId(session.clientId) {
-            action(project)
+    }
+
+    private fun withClient(request: FullHttpRequest, context: ChannelHandlerContext, action: (project: Project?) -> Unit): String? {
+        ApplicationManager.getApplication().executeOnPooledThread {
```
akosyakov, line 75: @felladrin Could you clarify what it does and why it is necessary please? 🙏

akosyakov:
@felladrin open --wait still does not work for me

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
```diff
@@ -53,22 +54,31 @@ class GitpodCLIService : RestService() {
...
+            val session = getClientSessionAsync()
+            ClientId.withClientId(session.clientId) {
+                action(getLastFocusedOrOpenedProject())
```
akosyakov, line 78: you should pass here a project from getClientSessionAsync

felladrin:
> @felladrin open --wait still does not work for me
> 
> I tried to perform `git rebase -i origin/master` and cli hangs for me forever.

Seems to be an issue with JetBrains Remote Development, cause we need to explicit click "Release" (in the editor) for it to detect that the file was closed/release. Closing the file tab is not working.

image

Let me confirm it with JetBrains team.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
felladrin: Updated!

components/ide/jetbrains/cli/cmd/root.go
```diff
@@ -21,4 +25,16 @@ func Execute() {
 	}
 }
 
+func getCliApiUrl() *url.URL {
+	var backendPort = 63342
+	if _, fileStatError := os.Stat("/ide-desktop/bin/idea-cli-dev"); !errors.Is(fileStatError, os.ErrNotExist) {
+		backendPort = backendPort + 1
```
mustard-mh, line 31: What's this used for? If we make `backendPort` in func, it can only be `63342` or `63343`

felladrin: `63342` is the default port used by JetBrains "remote-dev-server.sh".

For developing the `components/ide/jetbrains/backend-plugin`, we need to use IntelliJ IDEA on Gitpod. Which means we'll be connecting to the remote-dev-server that is running in `63342`.

For testing changes on backend-plugin, we run `launch-dev-server.sh`.
This starts a new remote-dev-server on that workspace [[1](https://github.com/gitpod-io/gitpod/blob/8a1930a9ea5eeb955bb3ae0564f9e53e1a9d1c98/components/ide/jetbrains/backend-plugin/README.md#development)].
This new remote-dev-server will look for the availability of port `63342`, and if it's not available, it will increment the port by 1 until it finds an open port.
So we can say it's always going to pick `63343`.

Also note that `/ide-desktop/bin/idea-cli-dev` is created by `launch-dev-server.sh` right before launching the remote-dev-server. That's why we're using port `63343` if that file exists.

felladrin: Let me update the code with the addition of this comment explaining the magic number.

Thanks for asking!

felladrin:
> @felladrin Please open a follow up issue about broken `gp open --wait`, assign it to yourself and add to team sycn under port management for JB. We need to fix it with help of JB.

Issue created.

- https://github.com/gitpod-io/gitpod/issues/10299

PR #10147, opened 2022-05-20
Configures the pvc.spec.storageClassName when it's not empty

### LINKED ISSUE #9990: Fix broken workspace integration test TestMissingBackup(PVC)
### Bug description

Fix the broken workspace integration test TestMissingBackup of PVC test  https://github.com/gitpod-io/gitpod/blob/81c5a507fd14b27675a1907393539da0a885bc78/test/tests/components/ws-manager/content_test.go#L188

```
=== RUN   TestMissingBackup/CreateWorkspace/it_can_run_workspace_tasks/pvc_backup_init
    content_test.go:205: cannot start workspace: "rpc error: code = Internal desc = runtime error: invalid memory address or nil pointer dereference"
```

### Steps to reproduce

Run the integration test
```shell
./dev/preview/install-k3s-kubeconfig.sh
cd test
go test -v ./... \
  -kubeconfig=/home/gitpod/.kube/config \
  -namespace=default \
  -run=TestMissingBackup
```

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

### LINKED ISSUE #10017: Make sure PVC is working in dev environment
If we create the PVC with `spec.storageClassName=""`, the Kubernetes will search the storage class name "" rather than using the default storage class (which is local-path) in the cluster, ref to https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reserving-a-persistentvolume.

We should update the code then to not set storageClassName in that case. As that would be what users in self hosted would probably do as well, do not set (keep empty) storage class to use default one.

jenting:
## Description
Configures the `pvc.spec.storageClassName` when it's not empty. Otherwise, the PVC requests empty storage class.

## Related Issue(s)

Fixes #10017
Fixes #9990

## How to test

```go
./dev/preview/install-k3s-kubeconfig.sh
cd test
go test -v ./... -run=TestMissingBackup -kubeconfig=/home/gitpod/.kube/config -namespace=default
```

## Release Notes

```release-note
[experimental] Workspace persistent volume claim (PVC) uses a default storage class within the Kubernetes cluster if the user does not configure it
```

## Documentation

N/A

### DIFF
--- a/components/ws-manager/pkg/manager/create.go
+++ b/components/ws-manager/pkg/manager/create.go
@@ -237,22 +237,29 @@ func (m *Manager) createPVCForWorkspacePod(startContext *startWorkspaceContext) 
 	if startContext.Class != nil {
 		PVCConfig = startContext.Class.PVC
 	}
-	storageClassName := PVCConfig.StorageClass
-	return &corev1.PersistentVolumeClaim{
+
+	pvc := &corev1.PersistentVolumeClaim{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      fmt.Sprintf("%s-%s", prefix, req.Id),
 			Namespace: m.Config.Namespace,
 		},
 		Spec: corev1.PersistentVolumeClaimSpec{
-			AccessModes:      []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce},
-			StorageClassName: &storageClassName,
+			AccessModes: []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce},
 			Resources: corev1.ResourceRequirements{
 				Requests: corev1.ResourceList{
 					corev1.ResourceName(corev1.ResourceStorage): PVCConfig.Size,
 				},
 			},
 		},
-	}, nil
+	}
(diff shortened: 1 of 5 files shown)


components/ws-manager/pkg/manager/create.go
```diff
@@ -237,22 +237,29 @@ func (m *Manager) createPVCForWorkspacePod(startContext *startWorkspaceContext)
...
 			},
 		},
-	}, nil
+	}
+	if PVCConfig.StorageClass != "" {
+		// Specify the storageClassName when the storage class is non-empty.
+		// This way, the Kubernetes uses the default StorageClass within the cluster.
+		// Otherwise, the Kubernetes would try to request the PVC with no class.
+		pvc.Spec.StorageClassName = &PVCConfig.StorageClass
+	}
```
utam0k, line 260: I think this change makes sense and this change affects the users because we have self-hosted uses. So, I think you should write the Release Note section

jenting: Address, PTAL.

utam0k (approved):
It works fine

PR #10148, opened 2022-05-20
[gitlab] get rid of UnhandledPromiseRejectionWarning

AlexTugarev:
This PR fixes several issues with the GitLab webhook receiver:

* clean up of logs: less pollution, and no `UnhandledPromiseRejectionWarning`
* align with [webhook implementation recommendation](https://docs.gitlab.com/ee/user/project/integrations/webhooks.html#configure-your-webhook-receiver-endpoint): avoid receiving double events on 501 responses or requests taking too long to be processed

```release-notes
[gitlab] webhook receiver clean-up and fixing UnhandledPromiseRejectionWarning
```

### DIFF
--- a/components/server/ee/src/prebuilds/gitlab-app.ts
+++ b/components/server/ee/src/prebuilds/gitlab-app.ts
@@ -32,49 +32,55 @@ export class GitLabApp {
 
     @postConstruct()
     protected init() {
+        /**
+         * see https://docs.gitlab.com/ee/user/project/integrations/webhooks.html#configure-your-webhook-receiver-endpoint
+         * for recommendation on creating webhook receivers:
+         *
+         *  - ignore unrecognized event payloads
+         *  - never return 500 status responses if the event has been handled
+         *  - prefer to return 200; indicate that the webhook is asynchronous by returning 201
+         *  - to support fast response times, perform I/O or computationally intensive operations asynchronously
+         */
         this._router.post("/", async (req, res) => {
             const event = req.header("X-Gitlab-Event");
-            if (event === "Push Hook") {
-                const context = req.body as GitLabPushHook;
-                const span = TraceContext.startSpan("GitLapApp.handleEvent", {});
-                span.setTag("request", context);
-                log.debug("GitLab push hook received", { event, context });
-                let user: User | undefined;
-                try {
-                    user = await this.findUser({ span }, context, req);
-                } catch (error) {
-                    TraceContext.setError({ span }, error);
-                    log.error("Cannot find user.", error, {});
-                }
-                if (!user) {
-                    res.statusCode = 503;
-                    res.send();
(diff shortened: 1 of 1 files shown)


components/server/ee/src/prebuilds/gitlab-app.ts
```diff
@@ -33,28 +33,36 @@ export class GitLabApp {
...
-                    user = await this.findUser({ span }, context, req);
-                } catch (error) {
-                    log.error("Cannot find user.", error, {});
-                }
-                if (!user) {
-                    res.statusCode = 503;
-                    res.send();
-                    return;
+            try {
+                const event = req.header("X-Gitlab-Event");
```
andrew-farries, line 37: Does Gitlab send any kind of signature with the payload that we should verify, like [Github does](https://docs.github.com/en/developers/webhooks-and-events/webhooks/securing-your-webhooks#validating-payloads-from-github)?

AlexTugarev: Nope, see https://gitlab.com/gitlab-org/gitlab/-/issues/19367

easyCZ:
Thanks for the PR. Would it make sense to write a test case for this?

AlexTugarev:
>  Thanks for the PR. Would it make sense to write a test case for this?

@easyCZ, yes, sure. Is this a general question, do you want want to help, or do you want me to create one?
How far should we go? Extracting the operation for which this PR adds a guard, or creating a whole mocked http server.

easyCZ:
> > Thanks for the PR. Would it make sense to write a test case for this?
> 
> @easyCZ, yes, sure. Is this a general question, do you want want to help, or do you want me to create one? How far should we go? Extracting the operation for which this PR adds a guard, or creating a whole mocked http server.

Sorry, I should've been clearer. Normally, when you encounter a 3rd party integration behaviour which is unexpected, it's useful to add a regression test to ensure the special treatment doesn't get undone without necessary context. It also helps as documentation of the behaviour. 

My question was more how difficult it would be for you to add it, or whether you feel it is necessary for this case.

AlexTugarev:
@geropl, thanks for the pointer, while updating the PR description as requested rightly, I took a change to strictly follow the recommendations on webhook receiver implementations, thus rewrote the part again to make it clean.

components/server/ee/src/prebuilds/gitlab-app.ts
```diff
@@ -32,49 +32,55 @@ export class GitLabApp {
 
     @postConstruct()
     protected init() {
+        /**
+         * see https://docs.gitlab.com/ee/user/project/integrations/webhooks.html#configure-your-webhook-receiver-endpoint
```
geropl, line 36: :orange_heart:

geropl (approved):
:+1:

PR #10152, opened 2022-05-20
[installer, ws-manager]: Use installation shortname in URL templates

andrew-farries:
## Description

ℹ️ ℹ️ ℹ️ ℹ️ 

This is a second attempt at #10127, which was reverted (#10143) as it broke preview environments. See [this comment](https://github.com/gitpod-io/gitpod/issues/9875#issuecomment-1132803114) in #9875. The last commit on the branch is a workaround for the preview environment issue.

 ℹ️ ℹ️ ℹ️ ℹ️ 

As described in https://github.com/gitpod-io/gitpod/issues/9875, failing image builds currently fail to produce any useful output. A possible cause, as described in [this comment](https://github.com/gitpod-io/gitpod/issues/9875#issuecomment-1129924819), is that the format of the `ws-manager` `WorkspaceURLTemplate` strings is wrong. 

When we deployed with the old helm chart, these values were set in the relevant configmap like this:

https://github.com/gitpod-io/gitpod/blob/2f988c93499926bd34d6728defbd9d9c0964797e/chart/templates/ws-manager-configmap.yaml#L74-L75

ie, they included a suffix after the `ws`. 

This PR tailors the installer output to produce these URLs in the same way the old helm chart did. Installation short name suffixes like `eu-02` are set here:

https://github.com/gitpod-io/ops/blob/main/deploy/production/meta-eu02/app/installer-config.yaml#L6

so the URLs will now look like `https://{{ .Prefix }}.ws-eu02.example.com`.

## Related Issue(s)
Fixes https://github.com/gitpod-io/gitpod/issues/9875

## How to test

I've tested this change in staging by editing the `ws-manager` configmap in staging to contain the correct URLs, restarting `ws-manager` and running an image build. I was able to see image build logs as expected:

image

## Release Notes
```release-note
[installer] Use installation shortname when constructing ws-manager URL templates
```

## Documentation
None

### DIFF
--- a/install/installer/pkg/components/ws-manager/configmap.go
+++ b/install/installer/pkg/components/ws-manager/configmap.go
@@ -113,6 +113,11 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		return nil, err
 	}
 
+	installationShortNameSuffix := ""
+	if ctx.Config.Metadata.InstallationShortname != "" {
+		installationShortNameSuffix = "-" + ctx.Config.Metadata.InstallationShortname
+	}
+
 	wsmcfg := config.ServiceConfiguration{
 		Manager: config.Configuration{
 			Namespace:      ctx.Namespace,
@@ -137,8 +142,8 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 			InitProbe: config.InitProbeConfiguration{
 				Timeout: (1 * time.Second).String(),
 			},
-			WorkspaceURLTemplate:     fmt.Sprintf("https://{{ .Prefix }}.ws.%s", ctx.Config.Domain),
-			WorkspacePortURLTemplate: fmt.Sprintf("https://{{ .WorkspacePort }}-{{ .Prefix }}.ws.%s", ctx.Config.Domain),
+			WorkspaceURLTemplate:     fmt.Sprintf("https://{{ .Prefix }}.ws%s.%s", installationShortNameSuffix, ctx.Config.Domain),
+			WorkspacePortURLTemplate: fmt.Sprintf("https://{{ .WorkspacePort }}-{{ .Prefix }}.ws%s.%s", installationShortNameSuffix, ctx.Config.Domain),
 			WorkspaceHostPath:        wsdaemon.HostWorkingArea,
 			Timeouts: config.WorkspaceTimeoutConfiguration{
 				AfterClose:          timeoutAfterClose,
--- a/install/installer/pkg/components/ws-manager/configmap.go
+++ b/install/installer/pkg/components/ws-manager/configmap.go
@@ -114,7 +114,7 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 	}
 
 	installationShortNameSuffix := ""
-	if ctx.Config.Metadata.InstallationShortname != "" {
+	if ctx.Config.Metadata.InstallationShortname != "" && ctx.Config.Metadata.InstallationShortname != configv1.InstallationShortNameOldDefault {
 		installationShortNameSuffix = "-" + ctx.Config.Metadata.InstallationShortname
 	}
 
--- a/install/installer/pkg/components/ws-manager/configmap_test.go
+++ b/install/installer/pkg/components/ws-manager/configmap_test.go
@@ -5,12 +5,16 @@
 package wsmanager
 
 import (
+	"encoding/json"
 	"testing"
 
 	"github.com/gitpod-io/gitpod/installer/pkg/common"
+	"github.com/gitpod-io/gitpod/installer/pkg/config/v1"
 	configv1 "github.com/gitpod-io/gitpod/installer/pkg/config/v1"
+	"github.com/gitpod-io/gitpod/installer/pkg/config/versions"
 	wsmancfg "github.com/gitpod-io/gitpod/ws-manager/api/config"
 	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/require"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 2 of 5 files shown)


sagor999:
@meysholdt how do you deploy gitpod in preview env? 
Andrew placed a workaround to make sure this change doesn't break preview, but I feel like we should instead fix the way we are deploying in preview (remove `default` as shortname)

sagor999:
@gitpod-io/platform for comment above. can you provide insight before we will merge this.

mads-hartmann:
@andrew-farries @sagor999 The Werft job uses the Gitpod Installer to generate a config for the installer and then modifies it quite heavily for preview environments. The implementation is [here](https://github.com/gitpod-io/gitpod/blob/main/.werft/jobs/build/installer/installer.ts).

The metadata.shortname is set to `default` by the installer when generating the config. I verified by running the Installer from my workspace off main.

```
gitpod /workspace/gitpod/install/installer (main) $ go run main.go  init | grep -6 metadata
  inCluster: true
database:
  inCluster: true
disableDefinitelyGp: true
domain: ""
kind: Full
metadata:
  region: local
  shortname: default
objectStorage:
  inCluster: true
observability:
  logLevel: info
```

So the change you want to make (if you just want to remove `metadata.shortname` from the config used for preview environments) is something like this

```diff
diff --git a/.werft/jobs/build/installer/installer.ts b/.werft/jobs/build/installer/installer.ts
index 52edde6f6..ea2a83671 100644
--- a/.werft/jobs/build/installer/installer.ts
+++ b/.werft/jobs/build/installer/installer.ts
@@ -58,6 +58,7 @@ export class Installer {
         this.options.werft.log(slice, "Adding extra configuration");
         try {
             this.getDevCustomValues(slice)
+            this.configureMetadata(slice)
             this.configureContainerRegistry(slice)
             this.configureDomain(slice)
             this.configureWorkspaces(slice)
@@ -92,6 +93,10 @@ export class Installer {
         exec(`yq m -i ${this.options.installerConfigPath} ${WORKSPACE_SIZE_CONFIG_PATH}`, { slice: slice });
     }
 
+    private configureMetadata(slice: string): void {
+        exec(`yq d -i ${this.options.installerConfigPath} metadata.shortname `, { slice: slice });
+    }
+
     private configureContainerRegistry(slice: string): void {
         exec(`yq w -i ${this.options.installerConfigPath} certificate.name ${this.options.proxySecretName}`, { slice: slice });
         exec(`yq w -i ${this.options.installerConfigPath} containerRegistry.inCluster false`, { slice: slice });
```

andrew-farries:
I've removed the `installation.shortname` when deploying to a preview environment. Previews still appear to work.

sagor999:
@andrew-farries can you rebase your PR? There is a breaking change without which build fails now :(

corneliusludmann (approved):

andrew-farries:
Tests failing due to this: https://github.com/gitpod-io/gitpod/pull/10243. 

Will rebase again when that's resolved.

kylos101:
@sagor999 could you test the installer asset produced by this build in an ephemeral cluster? I'm worried this could impact our post processing for `ws-proxy`.

kylos101:
> @sagor999 could you test the installer asset produced by this build in an ephemeral cluster? I'm worried this could impact our post processing for `ws-proxy`.

@Furisto wanted to give you a heads up on this, as it could impact our ability to deploy workspace clusters.

sagor999:
@kylos101 tested ephemeral cluster creation using installer from this branch. All worked fine. Was able to start workspace in that cluster.

sagor999 (approved):

install/installer/pkg/components/ws-manager/configmap_test.go
```diff
@@ -133,3 +137,58 @@ func TestBuildWorkspaceTemplates(t *testing.T) {
...
+		InstallationShortname            string
+		ExpectedWorkspaceUrlTemplate     string
+		ExpectedWorkspacePortURLTemplate string
+	}{
+		{
+			Name:                             "With an installation shortname",
+			Domain:                           "example.com",
+			InstallationShortname:            "eu02",
+			ExpectedWorkspaceUrlTemplate:     "https://{{ .Prefix }}.ws-eu02.example.com",
+			ExpectedWorkspacePortURLTemplate: "https://{{ .WorkspacePort }}-{{ .Prefix }}.ws-eu02.example.com",
```
geropl, line 154: :+1:

install/installer/pkg/components/ws-manager/configmap_test.go
```diff
@@ -133,3 +137,58 @@ func TestBuildWorkspaceTemplates(t *testing.T) {
...
+		},
+		{
+			Name:                             "Without an installation shortname",
+			Domain:                           "example.com",
+			InstallationShortname:            "",
+			ExpectedWorkspaceUrlTemplate:     "https://{{ .Prefix }}.ws.example.com",
+			ExpectedWorkspacePortURLTemplate: "https://{{ .WorkspacePort }}-{{ .Prefix }}.ws.example.com",
```
geropl, line 161: :+1:

geropl:
> I'm more concerned about the consequences of removing the installation shortname rather than the mechanics of how to do it, given the comment here from @geropl:

@andrew-farries I added that line when making `installationShortname` editable (again) and kept the "default" instead of moving to "" (empty string), because I did not want to run into migration trouble with existing self-hosted installations.

The problem is that whatever we put into `installationShortname` is what new workspaceInstances get set as "region". And for the lifetime of a workspace, we need to make sure there is a ws-manager registered for exactly that "region"/"cluster".

> But I guess we can leave it as the default in the installer and remove it specifically for preview environments.

:100: This is what I would have done as well.

install/installer/pkg/components/ws-manager/configmap.go
```diff
@@ -113,6 +113,11 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		return nil, err
 	}
 
+	installationShortNameSuffix := ""
+	if ctx.Config.Metadata.InstallationShortname != "" {
```
geropl, line 117: @andrew-farries IMO we need to check for `!= "" && != "default", no? (cmp test cases below)

install/installer/pkg/components/ws-manager/configmap_test.go
geropl: @andrew-farries We need a test case "with 'default' as installation shortname" to cover all existing self-hosted installations out there I think. That will lead us to [the comment above](https://github.com/gitpod-io/gitpod/pull/10152/files#r883577546) I think.

geropl (changes requested):
https://github.com/gitpod-io/gitpod/pull/10152/files#r883578267

geropl (approved):
Fixes (my own) complaint with [this commit](https://github.com/gitpod-io/gitpod/pull/10152/commits/88230166a9495202802e844cd98fd4dcf9dde61f).

Issue #10028, opened 2022-05-16
Reduce memory request of minio and workspaces in preview envs


meysholdt:
## Is your feature request related to a problem? Please describe
Reduce memory request of minio and workspaces in preview envs.

This will help to fit more workspaces into a preview VM.

## Additional context
https://gitpod.slack.com/archives/C01KGM9EBD4/p1652294079424289

meysholdt:
Context: 
* The k8s YAML gets generated by the [installer](https://github.com/gitpod-io/gitpod/tree/main/install). It also holds the config for [minio](https://github.com/gitpod-io/gitpod/tree/main/install/installer/pkg/components/minio).
* The YAML for workspaces gets created by [ws-manager](https://github.com/gitpod-io/gitpod/tree/main/components/ws-manager), and IIRC it's controlled by a config. 
* The installer also receives a config. Potentially that's another option to adjust limits.
* Another option (last resort) is to [post-process the YAML from the installer](https://github.com/gitpod-io/gitpod/blob/main/.werft/jobs/build/installer/post-process.sh).

proposed solution: Preferably adjust the memory requests in configs that are only used for preview envs but not for production systems. If that's not possible, resort to post-processing.

PR #10158, opened 2022-05-20
Reduce memory requests of minio in preview envs

vulkoingim:
## Description

Average memory usage for `minio` seems to be ~100mb - I've left it a bit more just in case

```
kubectl top pods minio-859b8f94b5-xztxk --containers
POD                      NAME    CPU(cores)   MEMORY(bytes)   
minio-859b8f94b5-xztxk   minio   2m           87Mi   
```

```release-note
Reduce memory requests of minio in preview environments
```

## Related Issue(s)

Fixes # https://github.com/gitpod-io/gitpod/issues/10028

### DIFF
--- a/install/installer/pkg/config/v1/config.go
+++ b/install/installer/pkg/config/v1/config.go
@@ -49,6 +49,9 @@ func (v version) Defaults(in interface{}) error {
 	cfg.Metadata.Region = "local"
 	cfg.Metadata.InstallationShortname = "default" // TODO(gpl): we're tied to "default" here because that's what we put into static bridges in the past
 	cfg.ObjectStorage.InCluster = pointer.Bool(true)
+	cfg.ObjectStorage.Resources.Requests = corev1.ResourceList{
+		corev1.ResourceMemory: resource.MustParse("2Gi"),
+	}
 	cfg.ContainerRegistry.InCluster = pointer.Bool(true)
 	cfg.Workspace.Resources.Requests = corev1.ResourceList{
 		corev1.ResourceCPU:    resource.MustParse("1000m"),
@@ -152,6 +155,7 @@ type ObjectStorage struct {
 	Azure              *ObjectStorageAzure        `json:"azure,omitempty"`
 	MaximumBackupCount *int                       `json:"maximumBackupCount,omitempty"`
 	BlobQuota          *int64                     `json:"blobQuota,omitempty"`
+	Resources          Resources                  `json:"resources,omitempty"`
 }
 
 type ObjectStorageS3 struct {
--- a/.werft/jobs/build/installer/installer.ts
+++ b/.werft/jobs/build/installer/installer.ts
@@ -61,6 +61,7 @@ export class Installer {
             this.configureContainerRegistry(slice)
             this.configureDomain(slice)
             this.configureWorkspaces(slice)
+            this.configureObjectStorage(slice)
             this.configureIDE(slice)
             this.configureObservability(slice)
             this.configureAuthProviders(slice)
@@ -107,6 +108,10 @@ export class Installer {
     private configureWorkspaces(slice: string) {
         exec(`yq w -i ${this.options.installerConfigPath} workspace.runtime.containerdRuntimeDir ${CONTAINERD_RUNTIME_DIR}`, { slice: slice });
         exec(`yq w -i ${this.options.installerConfigPath} workspace.resources.requests.cpu "100m"`, { slice: slice });
+    }
+
+    private configureObjectStorage(slice: string) {
+        exec(`yq w -i ${this.options.installerConfigPath} objectStorage.resources.requests.memory "256Mi"`, { slice: slice });
     }
 
     private configureIDE(slice: string) {
@@ -181,7 +186,7 @@ export class Installer {
     }
 
     postProcessing(slice: string): void {
-        this.options.werft.log(slice, "Post processing YAML manfests");
+        this.options.werft.log(slice, "Post processing YAML manifests");
 
         this.configureLicense(slice)
         this.configureWorkspaceFeatureFlags(slice)
--- a/install/installer/example-config.yaml
+++ b/install/installer/example-config.yaml
@@ -18,6 +18,9 @@ metadata:
   shortname: default
 objectStorage:
   inCluster: true
+  resources:
+    requests:
+      memory: 2Gi
 observability:
   logLevel: info
 openVSX:
@@ -27,6 +30,7 @@ workspace:
   maxLifetime: 36h0m0s
   pvc:
     size: 30Gi
+    snapshotClass: ""
     storageClass: ""
   resources:
     requests:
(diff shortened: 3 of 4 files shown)


roboquat:
@vulkoingim: Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://git.k8s.io/community/contributors/guide/release-notes.md) to remove it.

<details>

</details>

install/installer/pkg/config/v1/config.go
```diff
@@ -49,6 +49,9 @@ func (v version) Defaults(in interface{}) error {
 	cfg.Metadata.Region = "local"
 	cfg.Metadata.InstallationShortname = "default" // TODO(gpl): we're tied to "default" here because that's what we put into static bridges in the past
 	cfg.ObjectStorage.InCluster = pointer.Bool(true)
+	cfg.ObjectStorage.Resources.Requests = corev1.ResourceList{
+		corev1.ResourceMemory: resource.MustParse("256Mi"),
```
sagor999, line 53: what was default prior to this change? or there was no requests at all?

sagor999 (approved):
/hold
if you want to address nit.

easyCZ (approved):
LGTM, approving for WebApp

install/installer/pkg/config/v1/config.go
```diff
@@ -152,6 +155,7 @@ type ObjectStorage struct {
 	Azure              *ObjectStorageAzure        `json:"azure,omitempty"`
 	MaximumBackupCount *int                       `json:"maximumBackupCount,omitempty"`
 	BlobQuota          *int64                     `json:"blobQuota,omitempty"`
+	Resources          Resources                  `json:"resources" validate:"required"`
```
mads-hartmann, line 158: I'm a bit hesitant to make this required as that's a breaking change (as far as I know). Can we make it optional instead?

install/installer/pkg/config/v1/config.go
mads-hartmann: I'd prefer to have exactly the same behaviour as before, so if it wasn't set before I'd prefer not to set it by default now either ☺️

mads-hartmann (changes requested):
Thanks for the PR @vulkoingim! I'd prefer for this change to be a no-op for anyone who doesn't want to set resource requests for minio. That would mean making the config optional, and if the user doesn't set the config, then don't provide defaults.

install/installer/pkg/config/v1/config.go
mads-hartmann: I realise there might have been some context missing about the installer, please see this comment where I elaborate a bit more ☺️ https://github.com/gitpod-io/gitpod/pull/10191#pullrequestreview-982693345

vulkoingim: By default `minio` gets 2G of memory, but it's not obvious where it is set ([have to dig to find the original yaml, which gets overwritten eventually](https://github.com/gitpod-io/gitpod/blob/main/install/installer/third_party/charts/minio/values.yaml#L13)) - that's why I opted for making it required as it makes it explicit. I realise I should have left a bit more context.

install/installer/pkg/config/v1/config.go
mads-hartmann: Thanks for clarifying! Setting it to 2gb sounds good then ☺️

install/installer/pkg/config/v1/config.go
vulkoingim: > When you have a default value, you don't need to set a value in the config. The user-given config from the file is merged with the default config in the code.

I thought it would be like that as well, but it seems that the merging is done at some later stage. 

1. Installer inits a default config
2. The custom configs in `installer.ts` get applied
3. Validation runs and fails if it wasn't set if it was marked as required

https://github.com/gitpod-io/gitpod/blob/main/.werft/jobs/build/deploy-to-preview-environment.ts#L290-L292

corneliusludmann: @vulkoingim: Sorry, I don't get it. :innocent: 

With this installer: `eu.gcr.io/gitpod-core-dev/build/installer:aa-minio-resources.20` (where it was required and not `omitempty`), you can run
1. `installer init > config.yaml`
2. Remove
   ```
   resources:
     requests:
       memory: 256Mi
   ```
   from `objectStorage` block in `config.yaml`
3. and `installer validate config --config config.yaml` success as expected!?


The only difference between `omitempty` and `validate:"required"` is whether it is generated by `installer init` or not. And with `omitempty` it's not part of the generated `config.yaml` but I think it should. That makes that config value easier to discover. That's why I would vote for settings is as required and not `omitempty`. What do you think?

mads-hartmann: > The only difference between omitempty and validate:"required" is whether it is generated by installer init or not. And with omitempty it's not part of the generated config.yaml

Sorry for leading you astray @vulkoingim - I thought that adding `validate:"required"` meant it had to be part of config.yaml.

vulkoingim: @corneliusludmann 
> The only difference between omitempty and validate:"required" is whether it is generated by installer init or not. And with omitempty it's not part of the generated config.yaml but I think it should.

The `required` part is also used for validation, not only for generating the config. So it will be generated, but without a default value (or overwritten in `installer.ts`) it will be set to null, so eventually it will fail the validation. It would look like this:

```
objectStorage:
  inCluster: true
  resources:
    requests: null
```

In general I also think that we should set it to `required` and give it a default value in the [config.go:53](https://github.com/gitpod-io/gitpod/pull/10158/files#diff-ed0dab0d7e4e73c6487f629ed17403feeb8ee3b8cf1601e5c940f7a9365ccf17R53), as I've done.

But what Mads also says is true - if we set it to `required` and someone is using an old config with this new version of the installer, where this part is not generated - it will fail.

```
{
  "valid": false,
  "fatal": [
    "Field 'Config.ObjectStorage.Resources.Requests' is required"
  ]
}

```

corneliusludmann: I just tested the behavior for both versions and both fail for this config:
```
objectStorage:
  inCluster: true
  resources:
    requests: null
```

However, when the new config value `resources` in `objectStorage` is missing entirely (that is what old config files would have, it works with both versions as well.

And: Both versions generate the resources block (I was wrong here and it makes sense since the value is not empty and thus not omitted).

Therefore, I'm fine with the current implementation.

corneliusludmann (approved):
Looks good. :rocket: 

A few random comments:
1. Please squash your commits into one commit with a meaningful message.
2. Since you already there, would you mind running the following command and pushing the updated `example-config.yaml`?
   ```
   cd install/installer && go run . init > example-config.yaml
   ```
4. You need to add a release-note block to your pull request description with a meaningful message like this:
   ~~~
   ```release-note
   Add your meaningful changelog message here!
   ```
   ~~~

When you are done, remove the `do-not-merge/hold` label from this pull request.

Thanks a lot! :ship: :skateboard:

mads-hartmann (approved):
Thanks 🎉

Issue #9667, opened 2022-04-30
VS Code Release 1.67


jeanp413:
Endgame: https://github.com/microsoft/vscode/issues/148074

Expected release date: 05/04 

DRI: @filiptronicek @jeanp413 
Plan:

- [x]  Create corresponding release branches for `main` and `gp-code/main`
    - [x] Create release branch `release/1.67`
    	- Execute the rebase from upstream: `./scripts/sync-with-upstream.sh upstream/release/1.67 release/1.67`
    - [x] Create release branch `gp-code/release/1.67`
     	- Execute the rebase from upstream: `./scripts/sync-with-upstream.sh upstream/release/1.67 gp-code/release/1.67`
- [x]  Switch nightly jobs in https://github.com/gitpod-io/openvscode-releases and Gitpod Code-Nightly GitHub actions to point to the release branches
    - [x] [insiders-gp.yml](https://github.com/gitpod-io/openvscode-releases/blob/1b60a53a1a34b61dfb41ca13b65bfbad4115bda4/.github/workflows/insiders-gp.yml#L19)
    - [x] [insiders.yml](https://github.com/gitpod-io/openvscode-releases/blob/1b60a53a1a34b61dfb41ca13b65bfbad4115bda4/.github/workflows/insiders.yml#L19)
    - [x] [code-nightly.yaml](https://github.com/gitpod-io/gitpod/blob/4bdee21961c5390e1dc61606b7070af3b0f65971/.github/workflows/code-nightly.yaml#L34) 
- [x]  Create and merge PR in Gitpod repo to generate stable image for VS Code https://github.com/gitpod-io/gitpod/pull/9761
    - Use the template from https://gist.github.com/filiptronicek/be19dcab639a1cdf08089cc762377a41
    - [x]  [Smoke test](https://www.notion.so/Gitpod-VS-Code-1aa1dfcfdc5147869ec5ffcf86f430a6) version 1.67 in VS Code Insiders
- [x]  Create and merge PR in Gitpod repo updating VS Code stable image tags to image generated in previous step https://github.com/gitpod-io/gitpod/pull/9777
    - [values.yaml](https://github.com/gitpod-io/gitpod/blob/14a9636f476fc69294c5d0d9e7f831f16b39074f/chart/values.yaml#L361)
    - [constants.go](https://github.com/gitpod-io/gitpod/blob/866357d3743c875c56883189778492a4fafcca03/install/installer/pkg/components/workspace/ide/constants.go#L9)
[...]

PR #10159, opened 2022-05-20
Update VS Code Stable to `1.67.2`

filiptronicek:
## Description
Updates the stable version of VS Code Web to https://github.com/microsoft/vscode/releases/tag/1.67.2.

## Related Issue(s)
Tracked in https://github.com/gitpod-io/gitpod/issues/9667

## How to test
1. In the preview environment, select VS Code Web stable as the IDE.
2. Start a workspace
3. Check that the about dialog in VS Code says that you are on `1.67.2`.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/install/installer/pkg/components/workspace/ide/constants.go
+++ b/install/installer/pkg/components/workspace/ide/constants.go
@@ -6,7 +6,7 @@ package ide
 
 const (
 	CodeIDEImage                = "ide/code"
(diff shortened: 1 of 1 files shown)


filiptronicek:
@jeanp413 is it so that we now only update `install/installer/pkg/components/workspace/ide/constants.go`, since `chart/values.yaml` is no longer a thing?

jeanp413 (approved):
yep, now only this line needs to be updated :smile:

PR #10167, opened 2022-05-20
[observability] add horizontal line for load avg panel

sagor999:
## Description

Currently when viewing load average per cluster, due to dynamic scaling, it is very difficult to see when the cluster is starting to experience high load avg, as you have to constantly look at the left axis to see the actual numbers. That is cumbersome.
I propose adding horizontal line at value `1`. Why `1? Because it will help greatly with cluster that have very low load, and comparing them to other clusters.
If we change node types, we would need to update this panel in the future.

Compare old way:
image

vs new way:
image

I believe it shows a much clearer picture.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
none
```

### DIFF
--- a/operations/observability/mixins/cross-teams/dashboards/gitpod-overview.libsonnet
+++ b/operations/observability/mixins/cross-teams/dashboards/gitpod-overview.libsonnet
@@ -82,6 +82,12 @@ local wsNodeLoadAverageGraph =
       topk(5, sum(nodepool:node_load1:normalized{%(clusterLabel)s=~"$cluster", nodepool=~".*workspace.*"}) by (node))
     ||| % _config, legendFormat='{{node}}'
   ))
+  .addTarget(prometheus.target(
+    |||
+      1
+    ||| % _config, legendFormat='Node Max Load Avg'
+  ))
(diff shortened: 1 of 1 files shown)


ArthurSens:
Hi pavel!

If I'm not mistaken, I this graph we already normalize the load average. By "normalize" I mean that we divide the load average of every node by the number of CPU cores that it has.

This means that everything above 1 is what actually describes the situation mentioned in your PR. And it works for all machine types, doesn't matter the size 🙂

sagor999:
Heh. You are right! 
And I was reading that graph wrong this entire time. :man_facepalming:

sagor999:
Updated PR. Instead of 16, lets add horizontal line at 1. Why? It still will help greatly with cluster that have very low load, and comparing them to other clusters.
Example:
image

vs
image

sagor999:
@ArthurSens actually even better option would be to:
1. set SoftMax to 1.0
2. Set Threshold to 1.0 and set show thresholds to `as filled regions and lines`
image

But I have no idea how to do it in jsonnet. :( Would you be able to update this PR with suggested change?

princerachit (approved):
We can add enhancement in another PR

PR #10175, opened 2022-05-22
[jb] configure vmoptions for intellij backend server

yaohui-wyh:
## Description

1. Specify JVM max heap usage with `-XX:MaxRAMPercentage` (follows #9796 ) for better resource usage in a containerized env
2. Move all vmoptions customizations into `/ide-desktop/status` and make it easier for future vmoptions optimizations

## Related Issue(s)

Related #8704

## How to test

Test manually at https://github.com/gitpod-io/gitpod/issues/8704#issuecomment-1133104647

## Release Notes

```release-note
- Allow `-XX:MaxRAMPercentage` VM setting for JetBrains backend server
```


akosyakov:
Maybe we should look into more generic solution to allow customisations for different JVMs than introducing an env var for everything? Maybe we could use one env var instead like `[option]_VM_OPTIONS`? `[option]_XMX` makes sense to have separate only because it treated specially by launcher script.

components/ide/jetbrains/image/status/main.go
```diff
@@ -322,6 +326,35 @@ func configureXmx(alias string) error {
 	return ioutil.WriteFile(launcherPath, []byte(newContent), 0)
 }
 
+func configureVMOptions(alias string) error {
```
akosyakov, line 329: Supervisor can run IDE startup scripts several times. Status program should make sure that it produces the same result each time, i.e. does not prepend options over and over.

yaohui-wyh: Good catch, I will replace the prepend with replace & deduplicate

akosyakov:
It sounds good

akosyakov:
@iQQBot do you think you can help @yaohui-wyh how to setup self-hosted solution to test PRs?

yaohui-wyh:
I test the PR on Gitpod SaaS. PTAL @akosyakov 🙏 

----

The test plan is to use a Gitpod workspace (with intellij backend server) to build the `status` binary and replace the `/ide-desktop/status`, and kill the `status` process from the workspace (`supervisor` would restart the `status` process), then the main logic can be tested in-place

1. Choose IntelliJ as the default editor <https://gitpod.io/preferences>
2. Add envvar `INTELLIJ_VMOPTIONS`, such as `-XX:MinRAMPercentage=25 -Xmx3g`
3. Start Gitpod workspace from this PR: <https://gitpod.io/#https://github.com/gitpod-io/gitpod/pull/10175>
4. Launch a new Terminal and build the `status` binary:  `cd /workspace/gitpod/components/ide/jetbrains/image/status && go build` (maybe better with `leeway build ... <some args>`)
6. Replace the `status` binary: `rm /ide-desktop/status; cp status /ide-desktop/status`
7. Check the IDE server Java Process VMOptions: `jps -ml | grep intellij; jinfo <pid> > ~/jinfo1.txt`
8. Kill the `status` process: `ps aux | grep status; kill -9 <pid>`
9. (Supervisor restart `status` process), and check step7 again: `jps -ml | grep intellij; jinfo <pid> > ~/jinfo2.txt`

Expected behavior:
1. `jinfo2.txt` (Step7) should contain VMOptions set by envvar (Step2), and no other VMOptions are touched
2. Connect to Gitpod workspace via Gateway, check the "backend control center" from JBThinClient, and the Maximum heap size should be 3GiB

----

image

image

image

p

components/ide/jetbrains/image/status/main.go
```diff
@@ -307,19 +307,65 @@ func handleSignal(projectPath string) {
...
+			result = append(result, left)
+		}
+	}
+	return result
+}
+
+func updateVMOptions(alias string, content string) string {
+	// inspired by how intellij platform merge the VMOptions
+	// https://github.com/JetBrains/intellij-community/blob/master/platform/platform-impl/src/com/intellij/openapi/application/ConfigImportHelper.java#L1115
+	filterFunc := func(l, r string) bool {
```
akosyakov, line 346: why it is not part of deduplicateVMOption?

yaohui-wyh: To make it easier to have finer control over the type of VMOptions to dedup. Currently, only memory-related VMOptions (`-Xmx` / `-Xms`, etc.) are handled, however, if we need to deal with VMOptions such as GC-related by another feature toggle (e.g. a separate env var). Putting all filter logic into deduplicateVMOption is a bit messy.

It's similar to `updateVMOptions` (https://github.com/JetBrains/intellij-community/blob/master/platform/platform-impl/src/com/intellij/openapi/application/ConfigImportHelper.java#L1068) from the Intellij codebase

akosyakov:
@yaohui-wyh Do you want to add it to .gitpod.yml?

akosyakov:
> Hi Anton! Do you mean adding to .gitpod.yml [user document](https://www.gitpod.io/docs/environment-variables) how to customize VMOptions for Gateway? Or modify the .gitpod.yml of some JB Gateway demo projects such as gitpod-io/spring-petclinic?

I mean first adding it to gitpod.yml schema and actually making use of it. Then everything else.

yaohui-wyh:
> I mean first adding it to gitpod.yml schema and actually making use of it. Then everything else.

Thanks, I get it. Yeah I would add the vmoptions config to `gitpod-config-type` (according to this reply https://github.com/gitpod-io/gitpod/pull/10175#issuecomment-1134951660)

One thing to mention, I think users can already make use of it now by specifying [variables](https://gitpod.io/variables) for a project. However, add the option to project-wide `.gitpod.yml` config does give users more flexibility.

akosyakov:
> One thing to mention, I think users can already make use of it now by specifying [variables](https://gitpod.io/variables) for a project. However, add the option to project-wide .gitpod.yml config does give users more flexibility.

.gitpod.yml is to share across all users of the project, it can be overridden by user env vars

PR #10177, opened 2022-05-22
[jb] enable connect button only when JetBrains Client not activated

yaohui-wyh:
## Description

Each time user clicked on the **"connect"** button, previous activated JetBrainsClient is **restarted**. Since JBClient takes quite a long startup time, disable the connect button to prevent unexpected restart might be reasonable.

This PR syncs the jbclient's state for workspace view: when a **JetBrainsClient** is launched for a Gitpod workspace (via JetBrains Gateway Plugin), update the **"connect"** button text as **"connected"**, and disable the connect button until the JetBrainsClient is closed

x

## How to test

Build & Install the plugin following
https://github.com/gitpod-io/gitpod/blob/main/components/ide/jetbrains/gateway-plugin/README.md#L8

## Release Notes

```release-note
JetBrains Gateway: The "Connect" button now gets disabled while a JetBrains Client is connected to the workspace.
```

- [x] /werft no-preview=true

### DIFF
--- a/components/ide/jetbrains/gateway-plugin/src/main/kotlin/io/gitpod/jetbrains/gateway/GitpodConnectionProvider.kt
+++ b/components/ide/jetbrains/gateway-plugin/src/main/kotlin/io/gitpod/jetbrains/gateway/GitpodConnectionProvider.kt
@@ -9,7 +9,9 @@ import com.fasterxml.jackson.databind.DeserializationFeature
 import com.fasterxml.jackson.databind.PropertyNamingStrategies
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.intellij.ide.BrowserUtil
+import com.intellij.openapi.Disposable
 import com.intellij.openapi.components.service
+import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.diagnostic.thisLogger
 import com.intellij.remote.RemoteCredentialsHolder
 import com.intellij.ssh.AskAboutHostKey
@@ -20,6 +22,7 @@ import com.intellij.ui.components.JBTextArea
 import com.intellij.ui.dsl.builder.panel
 import com.intellij.ui.dsl.gridLayout.HorizontalAlign
 import com.intellij.ui.dsl.gridLayout.VerticalAlign
+import com.intellij.util.EventDispatcher
 import com.intellij.util.application
 import com.intellij.util.io.DigestUtil
 import com.intellij.util.ui.JBFont
@@ -32,6 +35,8 @@ import com.jetbrains.gateway.ssh.ClientOverSshTunnelConnector
 import com.jetbrains.gateway.thinClientLink.ThinClientHandle
 import com.jetbrains.rd.util.URI
 import com.jetbrains.rd.util.lifetime.Lifetime
+import com.jetbrains.rd.util.lifetime.isAlive
+import com.jetbrains.rd.util.lifetime.onTermination
 import io.gitpod.gitpodprotocol.api.entities.WorkspaceInstance
 import io.gitpod.jetbrains.icons.GitpodIcons
 import kotlinx.coroutines.*
@@ -51,6 +56,26 @@ import kotlin.coroutines.coroutineContext
 class GitpodConnectionProvider : GatewayConnectionProvider {
 
     private val gitpod = service<GitpodConnectionService>()
+
+    companion object {
+        var jetbrainsClientMap: MutableMap<String, ThinClientHandle> = mutableMapOf()
+
+        private val dispatcher = EventDispatcher.create(Listener::class.java)
+
+        private interface Listener : EventListener {
+            fun didChange()
+        }
+
+        fun addListener(listener: () -> Unit): Disposable {
+            val internalListener = object : Listener {
+                override fun didChange() {
+                    listener()
+                }
+            }
+            dispatcher.addListener(internalListener);
+            return Disposable { dispatcher.removeListener(internalListener) }
+        }
+    }
 
     private val httpClient = HttpClient.newBuilder()
         .followRedirects(HttpClient.Redirect.ALWAYS)
@@ -234,18 +259,27 @@ class GitpodConnectionProvider : GatewayConnectionProvider {
                                     )
                                     val client = connector.connect()
                                     client.clientClosed.advise(connectionLifetime) {
+                                        thisLogger().d("clientClosed, ${client.prettyPrint()}")
                                         application.invokeLater {
+                                            connectionLifetime.onTermination {
+                                                thisLogger().d("clientTermination, clientClosed didChange fired, ${client.prettyPrint()}")
+                                                dispatcher.multicaster.didChange()
+                                            }
                                             connectionLifetime.terminate()
                                         }
                                     }
                                     client.onClientPresenceChanged.advise(connectionLifetime) {
+                                        thisLogger().d("presenceChanged, ${client.prettyPrint()}")
                                         application.invokeLater {
+                                            thisLogger().d("clientPresenceChanged didChange fired, ${client.prettyPrint()}")
+                                            dispatcher.multicaster.didChange()
                                             if (client.clientPresent) {
                                                 statusMessage.text = ""
                                             }
                                         }
                                     }
                                     thinClient = client
+                                    jetbrainsClientMap[update.workspaceId] = client
                                 } catch (t: Throwable) {
                                     if (t is CancellationException) {
                                         throw t
@@ -443,4 +477,12 @@ class GitpodConnectionProvider : GatewayConnectionProvider {
 
     private data class SSHHostKey(val type: String, val hostKey: String)
 
+}
+
+fun ThinClientHandle.prettyPrint(): String {
+    return "uid=${uid}, clientPresent=${clientPresent}, lifeTime={status=${lifetime.status}, alive=${lifetime.isAlive}}"
+}
+
+fun Logger.d(message: String) {
+    thisLogger().info("[jb-gw] $message")
 }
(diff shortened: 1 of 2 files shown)


felladrin:
/werft run

Nice addition!
I'm gonna review it soon.

akosyakov:
> Gateway SDK is a black box sadly, I'm quite curious how @akosyakov gets to know the extension points of Gateway & figures out which apis to use :-) If we have more docs about the Gateway SDK, we could avoid lots of unexpected behaviors

Logging and trying 😆

felladrin:
> I have pushed the fix at 4618773, and I will rebase the commits and force push again.

Thanks for fixing it @yaohui-wyh. I see it refreshing the list after I close the IDE, but a new problem came up: it updates the list a few milliseconds before `GitpodConnectionProvider.jetbrainsClientMap[info.workspace.id]?.clientPresent` becomes `false`. So I see the list updating (I see it blinking and repositioning), but the button is still "Connected" after I close the Thin Client. And if, right after the update, I click (the fastest I can) the "Refresh" button, it updates to "Connect". Are you able to reproduce it?

Maybe we could delay `dispatcher.multicaster.didChange()` in 1 second here:
https://github.com/gitpod-io/gitpod/blob/24efeeb96f71eec8e7714c593561eec4fdc9074a/components/ide/jetbrains/gateway-plugin/src/main/kotlin/io/gitpod/jetbrains/gateway/GitpodConnectionProvider.kt#L258-L263

yaohui-wyh:
Hi @felladrin, Thanks again for the thorough testing!

> ... the button is still "Connected" after I close the Thin Client.

Yeah, I can reproduce it. (And sorry for that)
Actually I found out the "Connected" state won't switch back to "Connect" after the ThinClient closed most of the time. It updates to "Connect" only when I click the "Refresh" button. 

So the previous attempt (add `dispatcher.multicaster.didChange()` in the `client.clientClosed` **did passed the event**, however the **state** between _ConnectionProvider_ and _WorkspaceView_ is **not synced**)

----

And after digging into the properties of `ThinClient`:

1. **Which callback listener to subscribe**: after logging and trying @akosyakov, I think we should move  `dispatcher.multicaster.didChange()` to `connectionLifetime.onTermination` callback (rather than `client.clientClosed`) 
2. **Which state to check for ThinClient's liveness**: ThinClient's `lifeTime.isAlive` is the reliable way to check if JBClient is terminated, rather than `clientPresent`

```
2022-05-24 11:31:53,515 [jb-gw] clientClosed, uid=7d88d4fa-1c88-44f2-91a1-148d9009f43c, clientPresent=true, lifeTime={status=Alive, alive=true}
2022-05-24 11:31:53,528 [jb-gw] clientTermination, clientClosed didChange fired, uid=7d88d4fa-1c88-44f2-91a1-148d9009f43c, clientPresent=true, lifeTime={status=Canceled, alive=false}
2022-05-24 11:31:53,530 [jb-gw] WorkspaceView refresh, [wsId=gitpodio-springpetclini-n6bq52vs609, uid=7d88d4fa-1c88-44f2-91a1-148d9009f43c, clientPresent=true, lifeTime={status=Canceled, alive=false}]
```

> Note that the **clientPresent** was always true, while **lifeTime** reflected the correct state of JBClient

In short, we should rely on `Lifetime` of RD protocol (which is still a myth by now), but not `ThinClient`'s internal state. `ThinClient` does hold a `lifeTime` property and we should use that.

yaohui-wyh:
The exact fix is only one line
```diff
@@ -256,13 +259,19 @@ class GitpodConnectionProvider : GatewayConnectionProvider {
                                     )
                                     val client = connector.connect()
                                     client.clientClosed.advise(connectionLifetime) {
                                         application.invokeLater {
-                                            dispatcher.multicaster.didChange()
+                                            connectionLifetime.onTermination {
+                                                dispatcher.multicaster.didChange()
+                                            }
                                             connectionLifetime.terminate()
                                         }
                                     }
```

However I think it would be better to keep the logging statements & extension functions, since they are not on the critical path and wouldn't jam the log.

components/ide/jetbrains/gateway-plugin/src/main/kotlin/io/gitpod/jetbrains/gateway/GitpodConnectionProvider.kt
```diff
@@ -444,3 +478,11 @@ class GitpodConnectionProvider : GatewayConnectionProvider {
     private data class SSHHostKey(val type: String, val hostKey: String)
 
 }
+
+fun ThinClientHandle.prettyPrint(): String {
+    return "uid=${uid}, clientPresent=${clientPresent}, lifeTime={status=${lifetime.status}, alive=${lifetime.isAlive}}"
+}
+
+fun Logger.d(message: String) {
```
yaohui-wyh, line 486: Make the logging statement an extension function to make it easier for log filtering, and also it would be easier if we need to cleanup someday (IDEA will help us finding all the referenced position)

components/ide/jetbrains/gateway-plugin/src/main/kotlin/io/gitpod/jetbrains/gateway/GitpodConnectionProvider.kt
```diff
@@ -234,18 +259,27 @@ class GitpodConnectionProvider : GatewayConnectionProvider {
...
                                     client.clientClosed.advise(connectionLifetime) {
+                                        thisLogger().d("clientClosed, ${client.prettyPrint()}")
                                         application.invokeLater {
+                                            connectionLifetime.onTermination {
+                                                thisLogger().d("clientTermination, clientClosed didChange fired, ${client.prettyPrint()}")
+                                                dispatcher.multicaster.didChange()
+                                            }
                                             connectionLifetime.terminate()
                                         }
                                     }
```
felladrin, line 270: One thing I've noticed, is that we could benefit from letting the `connectionLifetime.onTermination` set even before the client gets closed. Because there is another place where we call `connectionLifetime.terminate()`, and in the future we can have even more cases.
Something like the following, which can be done in a follow-up PR if needed:
```suggestion
                                    connectionLifetime.onTermination {
                                        thisLogger().d("clientLifetimeTermination, didChange fired, ${client.prettyPrint()}")
                                        dispatcher.multicaster.didChange()
                                    }
                                    client.clientClosed.advise(connectionLifetime) {
                                        thisLogger().d("clientClosed, ${client.prettyPrint()}")
                                        application.invokeLater {
                                            connectionLifetime.terminate()
                                        }
                                    }
```

felladrin (approved):
It's working great now! Thanks for this PR @yaohui-wyh!

components/ide/jetbrains/gateway-plugin/src/main/kotlin/io/gitpod/jetbrains/gateway/GitpodConnectionProvider.kt
```diff
@@ -52,6 +57,26 @@ class GitpodConnectionProvider : GatewayConnectionProvider {
 
     private val gitpod = service<GitpodConnectionService>()
 
+    companion object {
+        var jetbrainsClientMap: MutableMap<String, ThinClientHandle> = mutableMapOf()
```
akosyakov, line 61: When do we remove a client from this map?

akosyakov:
This PR is leaking thin clients and does not look very structured :(

components/ide/jetbrains/gateway-plugin/src/main/kotlin/io/gitpod/jetbrains/gateway/GitpodWorkspacesView.kt
```diff
@@ -293,7 +300,10 @@ class GitpodWorkspacesView(
                                     )
                                 }
                                 label(getRelativeTimeSpan(info.latestInstance.creationTime))
-                                button("Connect") {
+                                val jbClientConnected = GitpodConnectionProvider.jetbrainsClientMap[info.workspace.id]
```
akosyakov, line 303: Should not we rather focus existing client instead?

components/ide/jetbrains/gateway-plugin/src/main/kotlin/io/gitpod/jetbrains/gateway/GitpodConnectionProvider.kt
```diff
@@ -234,18 +259,27 @@ class GitpodConnectionProvider : GatewayConnectionProvider {
...
                                         application.invokeLater {
+                                            thisLogger().d("clientPresenceChanged didChange fired, ${client.prettyPrint()}")
+                                            dispatcher.multicaster.didChange()
                                             if (client.clientPresent) {
                                                 statusMessage.text = ""
                                             }
                                         }
                                     }
                                     thinClient = client
+                                    jetbrainsClientMap[update.workspaceId] = client
```
akosyakov, line 282: Should not it be like:
```java
jetbrainsClientMap[update.workspaceId] = client
dispatcher.multicaster.didChange()
connectionLifetime.onTermination {
    jetbrainsClientMap.remove(update.workspaceId)
    dispatcher.multicaster.didChange()
}
```

We could also create a private function `setCleint(id, client | null)` and then do here:
```java
setClient(id, client);
connectionLifetime.onTermination {
    setClient(id, null) 
}
```

components/ide/jetbrains/gateway-plugin/src/main/kotlin/io/gitpod/jetbrains/gateway/GitpodConnectionProvider.kt
akosyakov: (nit) Should not it also be private and we expose instead `getClient`?

akosyakov:
@yaohui-wyh Will you send a clean up PR? of @felladrin takes care about clean up?

yaohui-wyh:
> @yaohui-wyh Will you send a clean up PR? of @felladrin takes care about clean up?

Sure, I will make a clean up PR. Thanks for the comments:) @akosyakov @felladrin

components/ide/jetbrains/gateway-plugin/src/main/kotlin/io/gitpod/jetbrains/gateway/GitpodWorkspacesView.kt
yaohui-wyh: This is a good idea, however I haven't found a way to requestFocus for the ThinClient yet

Issue #10057, opened 2022-05-17
Remaining issues for simplified team plans


jankeromnes:
### Bug description

Remaining small problems before the new Team Billing can be un-hidden, to finish https://github.com/gitpod-io/gitpod/issues/7759 :

- [x] Staging does not handle the new Chargebee webhooks properly?

Potentially relevant error from the `payment` pod:

```
QueryFailedError: ER_DATA_TOO_LONG: Data too long for column 'userId' at row 1
```

- [x] Did pro-ration on team join break?
- [x] Polish UI (especially in Dark theme) https://github.com/gitpod-io/gitpod/issues/8352
- [x] Un-hide the new Team Billing

### Steps to reproduce

N/A

PR #10182, opened 2022-05-23
Fix remaining issues for simplified team plans

jankeromnes:
## Description

Improve and ship the simplified Team Billing. ✨

TODO:

- [x] Un-hide the new Team Billing menu entry
- [x] Polish the UI
- [x] Fix race condition on multiple team joins 🐞

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/10057

## How to test

1. Create a Team
2. Navigate to Team Settings > Team Billing -- it should look nice ✨
3. Upgrade your Team to either Professional or Unleashed
    - (provide bogus billing details, for example type "a" into every required field, and use the proposed fake credit card)
4. Invite two more members, who should join at almost exactly the same time (within ~30s)
5. Check Chargebee-test for new charges. There should be:
    - one "Pro-rated upgrade from 1 to 2"
    - one "Pro-rated upgrade from 2 to 3"

(If the bug isn't fixed, you'd see one "from 1 to 2" invoice, and another "from 1 to 3" invoice(!) that's double the amount)

Screenshot of a successful test:

Screenshot 2022-05-27 at 10 21 45

## Release Notes

```release-note
Fix, polish, and release new simplified Team Billing
```

## Documentation

### DIFF
--- a/components/dashboard/src/components/DropDown.tsx
+++ b/components/dashboard/src/components/DropDown.tsx
@@ -11,6 +11,7 @@ import ContextMenu from "./ContextMenu";
 export interface DropDownProps {
     prefix?: string;
     customClasses?: string;
+    renderAsLink?: boolean;
     activeEntry?: string;
     entries: DropDownEntry[];
 }
@@ -35,14 +36,20 @@ function DropDown(props: DropDownProps) {
             },
         };
     });
-    const font =
-        "text-gray-400 dark:text-gray-500 text-sm leading-1 group hover:text-gray-600 dark:hover:text-gray-400 transition ease-in-out";
+    const defaultFont = "text-gray-400 dark:text-gray-500 hover:text-gray-600 dark:hover:text-gray-400 ";
+    const asLinkFont = "text-blue-500 dark:text-blue-400 hover:text-blue-600 dark:hover:text-blue-500";
+    const asLinkArrowBorder =
+        "border-blue-500 dark:border-blue-400 group-hover:border-blue-600 dark:group-hover:border-blue-500";
     return (
         <ContextMenu menuEntries={enhancedEntries} customClasses={`${props.customClasses} right-0`}>
-            <span className={`py-2 cursor-pointer ${font}`}>
+            <span
+                className={`py-2 cursor-pointer text-sm leading-1 group ${
+                    props.renderAsLink ? asLinkFont : defaultFont
+                } transition ease-in-out`}
+            >
                 {props.prefix}
                 {current}
-                <Arrow up={false} />
+                <Arrow up={false} customBorderClasses={props.renderAsLink ? asLinkArrowBorder : undefined} />
             </span>
         </ContextMenu>
     );
--- a/components/dashboard/src/components/DropDown.tsx
+++ b/components/dashboard/src/components/DropDown.tsx
@@ -10,7 +10,7 @@ import ContextMenu from "./ContextMenu";
 
 export interface DropDownProps {
     prefix?: string;
-    contextMenuWidth?: string;
+    customClasses?: string;
     activeEntry?: string;
     entries: DropDownEntry[];
 }
@@ -38,7 +38,7 @@ function DropDown(props: DropDownProps) {
     const font =
         "text-gray-400 dark:text-gray-500 text-sm leading-1 group hover:text-gray-600 dark:hover:text-gray-400 transition ease-in-out";
     return (
-        <ContextMenu menuEntries={enhancedEntries} classes={`${props.contextMenuWidth} right-0`}>
+        <ContextMenu menuEntries={enhancedEntries} customClasses={`${props.customClasses} right-0`}>
             <span className={`py-2 cursor-pointer ${font}`}>
                 {props.prefix}
                 {current}
--- a/components/dashboard/src/teams/TeamBilling.tsx
+++ b/components/dashboard/src/teams/TeamBilling.tsx
@@ -143,13 +143,14 @@ export default function TeamBilling() {
             title="Billing"
             subtitle="Manage team billing and plans."
         >
-            <h3>{!teamPlan ? "No billing plan" : "Plan"}</h3>
+            <h3>{!teamPlan ? "Upgrade Team Plan" : "Team Plan"}</h3>
             <h2 className="text-gray-500">
                 {!teamPlan ? (
                     <div className="flex space-x-1">
-                        <span>Select a new billing plan for this team. Currency:</span>
+                        <span>Upgrade team plan to access unlimited workspace hours, and more. Currency:</span>
                         <DropDown
                             customClasses="w-32"
+                            renderAsLink={true}
                             activeEntry={currency}
                             entries={[
                                 {
@@ -172,12 +173,12 @@ export default function TeamBilling() {
             <div className="mt-4 space-x-4 flex">
                 {isLoading && (
                     <>
-                        <SolidCard>
+                        <SolidCard className="w-72 h-64">
                             <div className="w-full h-full flex flex-col items-center justify-center">
                                 <Spinner className="h-5 w-5 animate-spin" />
                             </div>
                         </SolidCard>
-                        <SolidCard>
+                        <SolidCard className="w-72 h-64">
                             <div className="w-full h-full flex flex-col items-center justify-center">
                                 <Spinner className="h-5 w-5 animate-spin" />
                             </div>
@@ -188,14 +189,22 @@ export default function TeamBilling() {
                     <>
                         {availableTeamPlans.map((tp) => (
                             <>
-                                <SolidCard
-                                    className="cursor-pointer hover:bg-gray-200 dark:hover:bg-gray-700"
-                                    onClick={() => checkout(tp)}
-                                >
+                                <SolidCard className="w-72 h-72">
                                     <div className="px-2 py-5 flex-grow flex flex-col">
-                                        <div className="font-medium text-base">{tp.name}</div>
-                                        <div className="font-semibold text-gray-500 text-sm">Unlimited hours</div>
-                                        <div className="mt-8 font-semibold text-sm">Includes:</div>
+                                        <div className="font-semibold text-gray-800 dark:text-gray-100 text-lg">
+                                            {tp.name}
+                                        </div>
+                                        <div className="font-semibold text-gray-400 dark:text-gray-600 text-sm">
+                                            Unlimited hours
+                                        </div>
+                                        <div className="mt-2">
+                                            <PillLabel type="warn" className="font-semibold normal-case text-sm">
+                                                {members.length} x {Currency.getSymbol(tp.currency)}
+                                                {tp.pricePerMonth} = {Currency.getSymbol(tp.currency)}
+                                                {members.length * tp.pricePerMonth} per month
+                                            </PillLabel>
+                                        </div>
+                                        <div className="mt-4 font-semibold text-sm">Includes:</div>
                                         <div className="flex flex-col items-start text-sm">
                                             {(featuresByPlanType[tp.type] || []).map((f) => (
                                                 <span className="inline-flex space-x-1">
@@ -204,12 +213,10 @@ export default function TeamBilling() {
                                                 </span>
                                             ))}
                                         </div>
-                                        <div className="flex-grow flex flex-col items-end justify-end">
-                                            <PillLabel type="warn" className="font-semibold normal-case text-sm">
-                                                {members.length} x {Currency.getSymbol(tp.currency)}
-                                                {tp.pricePerMonth} = {Currency.getSymbol(tp.currency)}
-                                                {members.length * tp.pricePerMonth} per month
-                                            </PillLabel>
+                                        <div className="flex-grow flex flex-col items-stretch justify-end">
+                                            <button className="m-0" onClick={() => checkout(tp)}>
+                                                Upgrade to {tp.name}
+                                            </button>
                                         </div>
                                     </div>
                                 </SolidCard>
@@ -219,10 +226,14 @@ export default function TeamBilling() {
                 )}
                 {!isLoading && teamPlan && (
                     <>
-                        <Card>
+                        <Card className="w-72 h-64">
                             <div className="px-2 py-5 flex-grow flex flex-col">
-                                <div className="font-bold text-base">{teamPlan.name}</div>
-                                <div className="font-semibold text-gray-500 text-sm">Unlimited hours</div>
+                                <div className="font-semibold text-gray-100 dark:text-gray-800 text-lg">
+                                    {teamPlan.name}
+                                </div>
+                                <div className="font-semibold text-gray-400 dark:text-gray-600 text-sm">
+                                    Unlimited hours
+                                </div>
                                 <div className="mt-8 font-semibold text-sm">Includes:</div>
                                 <div className="flex flex-col items-start text-sm">
                                     {(featuresByPlanType[teamPlan.type] || []).map((f) => (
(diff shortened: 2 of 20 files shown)


roboquat:

<details>

</details>

gtsiolis:
Looking at this now! 👀

components/dashboard/src/teams/TeamBilling.tsx
```diff
@@ -34,6 +34,17 @@ export default function TeamBilling() {
...
+        "members",
+        members.length,
+        "currency",
+        currency,
+        "teamSubscription",
+        teamSubscription,
+        "pendingTeamPlan",
+        pendingTeamPlan,
+    );
+
```
gtsiolis, line 47: **suggestion**: Probably not the best line I could comment about this, but what do you think of making the [section title](https://github.com/gitpod-io/gitpod/pull/10182/files#diff-fab4237d49423136bbbf60168ab7b05684291690d3b0679ae4fa02d51c858f55R157) more actionable and relatable to the action below?

From _No billing plan_ to _Upgrade team plan_. 💡

gtsiolis, line 47: **suggestion**: Another irrelevant-line-to-comment-on: Could we rename team plan names and drop the duplicate and unnecessary _Team_ prefix, following the specs?

From _Team Professional_ to  _Professional_, etc.

gtsiolis, line 47: **suggestion**: Another irrelevant-line-to-comment-on: Could we adjust font color and weight following the specs?
- I find it a bit hard to parse the information in the current state. I could not find the line of code to comment directly with the class names needed. Small changes can make a difference.
- Also, what do you think of moving the sum above the plan features as in the most recent design spec, to avoid implying that the label in the bottom right corner is an action button?

| BEFORE | AFTER |
|-|-|
| cards-before | cards-after |

gtsiolis, line 47: **question**: Thoughts on rephrasing the subtitle from:

> Select a new billing plan for this team.

To:

> Upgrade team plan to access unlimited workspace hours, and more.

gtsiolis, line 47: **question**: Last comment on irrelevant line of code: Instead of the hover state as used with the editor and theme in preferences, we could use here a clear action button for each case, but we'd need to drop the hover state on the card. WDYT?

| BEFORE | AFTER |
|-|-|
| Frame 1088 | Frame 1089 |

gtsiolis (commented):
Thanks, @jankeromnes! Overall looks great. ✨ 

I couldn't find any code changes in this PR related to the UX comments left below, which made it harder to gather all these comments and find a place to add them. I assume these visual changes were part of another pull request, right?

Left some first round review comments. Let me know what you think.

jankeromnes:
Wow, many thanks @gtsiolis for the in-depth UX review and spot-on suggestions! 🪄

> I couldn't find any code changes in this PR related to the UX comments left below, which made it harder to gather all these comments and find a place to add them. I assume these visual changes were part of another pull request, right?

Ah, whoops, indeed. Most of these rough visual changes were actually implemented in https://github.com/gitpod-io/gitpod/pull/8041, but were left hidden from users by removing the new menu entry. 🕶️

Thank you for still managing to do the in-depth review! 🙏

components/dashboard/src/teams/TeamBilling.tsx
jankeromnes: I can easily hide the word "Team" in this page, if that's nicer visually, but we'll probably want to keep the word "Team" in all official team plans (e.g. to more easily differentiate them from individual plans in Chargebee, the DB, the Admin dashboard, etc)

gtsiolis: Any option could suffice for this iteration. I'll leave this to you to decide, @jankeromnes!

jankeromnes: I tried it, but then felt it looked too confusing -- using the same name as individual plans doesn't feel right. 🙈

So back to "Team" everywhere for now. Can be adjusted later, but also we'll deprecate these in favor of Usage-Based Billing in the not-too-distant future. 🔮

gtsiolis: Thanks for giving this a try, @jankeromnes! Going with plan :a: with the prefix sounds good.

jankeromnes:
Random lint config fail? 🤔 Let's try this again:

components/dashboard/src/teams/TeamBilling.tsx
```diff
@@ -232,25 +243,31 @@ export default function TeamBilling() {
...
-                                    <div className="font-semibold text-base text-gray-600">{members.length}</div>
-                                    <div className="mt-8 font-medium text-base text-gray-400">Next invoice on</div>
-                                    <div className="font-semibold text-base text-gray-600">
+                                    <div className="font-medium text-base text-gray-400 dark:text-gray-600">
+                                        Members
+                                    </div>
+                                    <div className="font-semibold text-base text-gray-600 dark:text-gray-400">
+                                        {members.length}
+                                    </div>
+                                    <div className="mt-8 font-medium text-base text-gray-400 dark:text-gray-600">
```
gtsiolis, line 264: **nitpick**: Minor spacing fix.

```suggestion
                                    <div className="mt-4 font-medium text-base text-gray-400 dark:text-gray-600">
```

components/dashboard/src/teams/TeamBilling.tsx
```diff
@@ -188,14 +189,22 @@ export default function TeamBilling() {
...
-                                        <div className="font-semibold text-gray-500 text-sm">Unlimited hours</div>
-                                        <div className="mt-8 font-semibold text-sm">Includes:</div>
+                                        <div className="font-semibold text-gray-800 dark:text-gray-100 text-lg">
+                                            {tp.name}
+                                        </div>
+                                        <div className="font-semibold text-gray-400 dark:text-gray-600 text-sm">
+                                            Unlimited hours
+                                        </div>
+                                        <div className="mt-2">
+                                            <PillLabel type="warn" className="font-semibold normal-case text-sm">
```
gtsiolis, line 201: **praise**: Don't know what you changed here but thanks for fixing the pill element here to hug contents instead of filling the container width.

components/dashboard/src/components/DropDown.tsx
```diff
@@ -35,14 +36,20 @@ function DropDown(props: DropDownProps) {
...
-    const font =
-        "text-gray-400 dark:text-gray-500 text-sm leading-1 group hover:text-gray-600 dark:hover:text-gray-400 transition ease-in-out";
+    const defaultFont = "text-gray-400 dark:text-gray-500 hover:text-gray-600 dark:hover:text-gray-400 ";
+    const asLinkFont = "text-blue-500 dark:text-blue-400 hover:text-blue-600 dark:hover:text-blue-500";
+    const asLinkArrowBorder =
+        "border-blue-500 dark:border-blue-400 group-hover:border-blue-600 dark:group-hover:border-blue-500";
     return (
-        <ContextMenu menuEntries={enhancedEntries} classes={`${props.contextMenuWidth} right-0`}>
-            <span className={`py-2 cursor-pointer ${font}`}>
+        <ContextMenu menuEntries={enhancedEntries} customClasses={`${props.customClasses} right-0`}>
```
gtsiolis, line 44: **praise**: Thanks for changing this, @jankeromnes!

gtsiolis, line 44: **thought**: Wondering if we need the custom classes altogether here. 💭

components/dashboard/src/teams/TeamBilling.tsx
```diff
@@ -232,25 +243,31 @@ export default function TeamBilling() {
...
                         {!teamSubscription ? (
-                            <SolidCard>
+                            <SolidCard className="w-72 h-64">
                                 <div className="w-full h-full flex flex-col items-center justify-center">
                                     <Spinner className="h-5 w-5 animate-spin" />
                                 </div>
                             </SolidCard>
                         ) : (
-                            <SolidCard>
+                            <SolidCard className="w-72 h-64">
```
gtsiolis, line 256: **thought**: Hoping in the future we can drop these by introducing component variants that include card size, etc. instead of hard-coding sizes like this. We'll get there!

components/dashboard/src/components/DropDown.tsx
```diff
@@ -35,14 +36,20 @@ function DropDown(props: DropDownProps) {
...
+            <span
+                className={`py-2 cursor-pointer text-sm leading-1 group ${
+                    props.renderAsLink ? asLinkFont : defaultFont
+                } transition ease-in-out`}
+            >
```
gtsiolis, line 49: **thought**: This is interesting! Is this only about the color change? If yes, it could be fine to leave this without the color change to match the rest of the link dropdowns but it's fine to also leave this as is for for now. Your call.

| BEFORE | AFTER |
|-|-|
| link-before | link-after |

gtsiolis (approved):
UX LGTM, @jankeromnes! Left one last round of non-blocking comments.

gtsiolis:
> /hold

@jankeromnes Also, regarding the hold comment in https://github.com/gitpod-io/gitpod/pull/10182#issuecomment-1141122366, in case it helps, this still needs a review from `@gitpod-io/engineering-webapp` to be merged.

> At least 1 approving review is required to merge this pull request.

components/dashboard/src/teams/TeamBilling.tsx
jankeromnes: There are ways to achieve this with a single component, e.g. by using:
- `min-height` on the card (to allow it to grow arbitrary large depending on its content)
- `items-stretch` on the card container (to make all the cards the same height, even if one card has a lot of content but the others don't)

components/dashboard/src/components/DropDown.tsx
jankeromnes: I believe we do, because the context-menu sometimes need to be left-aligned or right-aligned, among other things.

components/dashboard/src/teams/TeamBilling.tsx
jankeromnes: I removed a hard-coded `ml-2` from the component. 😇

components/dashboard/src/components/DropDown.tsx
jankeromnes: Yes, all these were only for the color change. 😅 But I think it's worth it, because with @jldec we found out that the grey currency switcher was very hard to notice when not link-color.

I think this is a nice improvement. ✨

jankeromnes:
> At least 1 approving review is required to merge this pull request.

Oh. Sounds like a bug if it's true -- you should absolutely be able to approve at least dashboard fixes & improvements. 🚢

components/dashboard/src/components/DropDown.tsx
gtsiolis: It's better to  never use right-aligned items inside the dropdown menus. 💡 

Do you have a handy instance of a left-alignment you're referring to?

I've only seen the dropdown in the workspace start page when using a desktop editor that's using center alignment and this should also change, see https://github.com/gitpod-io/gitpod/issues/6910.

If this is about the alignment of the component itself with the rest of the elements around it, we could probably move the alignment control outside this component, right?

components/dashboard/src/components/DropDown.tsx
gtsiolis: DEAL. Let's keep it for now. Maybe we could update all link dropdowns to become blue in the long run. Or the other way around.

components/dashboard/src/teams/TeamBilling.tsx
jankeromnes: Also, it's good to keep in mind that Tailwind is so nice because it allows directly "hard-coding" style in a nice way (as opposed to hiding style under multiple levels of abstractions, making them really hard to adjust afterwards).

jankeromnes: > Long-term, to keep the component flexible I'd love to introduce a stepper property to increase the width or height without using Tailwind's classes. 🧡

But, even better would be if the Cards always look perfectly-sized, depending on their contents, _without_ needing a stepper or Tailwind classes, right? 😇 ("Default correct" is better than "lots of knobs and switches")

components/dashboard/src/components/DropDown.tsx
jankeromnes: > It's better to only use left-aligned items inside the dropdown menus. 💡

Ah, sorry, I meant how the collapsible elements horizontally align with respect to the control, e.g.:

<table><tr><td>
Screenshot 2022-05-30 at 16 01 13
</td><td>
Screenshot 2022-05-30 at 16 01 18
</td><td>
Screenshot 2022-05-30 at 16 01 31
</td></tr><tr><td>
Screenshot 2022-05-30 at 16 01 56
</td><td>
Screenshot 2022-05-30 at 16 02 41
</td></tr></table>

geropl:
Starting review now...

geropl (approved):
@jankeromnes Tested and fixed: Thx for that, and approve! :heavy_check_mark: 

I _think_ I still managed to break it:
image

I think the reasons is the "Quantity" dropdown:
 - during initial upgrade I choose 10 (or sth)
 - I got billed 175 € - without gaining anything :grimacing: 
 - the subsequent 1st quantity upgrade (10 -> 2) did not trigger an addition charge being created (:+1: )
  - the 2nd quantity upgrade (2 -> 3) incurred the charge (+35€) for the current month (:+1: )

IMO we should remove the quantity dropdown if possible.

/hold feel free to address this in this PR or let's discuss/fix in a follow up! :+1:

components/dashboard/src/teams/TeamBilling.tsx
gtsiolis: > Also, it's good to keep in mind that Tailwind is so nice because it allows directly "hard-coding" style in a nice way

I'm also in favor of duplication over abstraction as it comes with multiple benefits, but in the long run it makes sense to abstract _some_ component aspects to help a) avoid running into code reviews nitpicking dimensions or other attributes, b) avoid requesting UX reviews when possible to avoid, c) feeling more confident about the product UX, and more.

However, this issue will become more relevant as these components mature and are widely used. For now, as long as we have two component instances of the solid card component, any approach would suffice and does not worth keeping this PR from being merged. 🤝 

> But, even better would be if the Cards always look perfectly-sized, depending on their contents, without needing a stepper or Tailwind classes, right?

Debatable! 🙂 

In most cases, the card component would need to use a specified width or height, which could be challenging to maintain when having multiple cards side-by-side that from a design perspective have to use fixed and specific dimensions regardless if some contents are missing. For example, see editor preferences, etc.

However, there's still need for having flexible card layouts without fixed width that fill the parent container on smaller viewports, etc.

jankeromnes:
> @jankeromnes Tested and fixed: Thx for that, and approve! ✔️

Woohoo 🙌 many thanks!

> - during initial upgrade I choose 10 (or sth)
> - I got billed 175 € - without gaining anything 😬

"User buys too many seats on purpose, is then surprised to pay for too many seats" 😅

All joking aside, I agree that it would be best to hide this quantity selector that isn't needed. It's confusing at best, if not actively harmful.

Will see if there is an easy fix, but if there isn't, I'll just merge this as is and open a follow-up issue.

jankeromnes:
I think the quantity selector can only be disabled in Chargebee's settings:

| Chargebee plan settings | Disallow editing quantity |
| --- | --- |
| Screenshot 2022-05-30 at 18 10 56 | Screenshot 2022-05-30 at 18 10 20 |

Happy to do this, but since it's a follow-up issue, let's unblock this PR first.

/unhold

components/dashboard/src/components/DropDown.tsx
gtsiolis: Ah, thanks for clarifying, @jankeromnes!

PR #10193, opened 2022-05-23
[server] make sure to release websocket clients

AlexTugarev:
What was initially introduces as some experimental try, seems to be a reasonable way to mitigate server leaks.

The way websocket connections are bounds to jsonrpc servers, which themselves integrate with the rest of the system, makes it necessary to have a circuit breaker in place to guarantee that stale client connections do not cause leakage of resource.

### How to test
Unfortunately, it's hard to provoke the otherwise missing `close` event. 

```release-notes
Fixing memory leak in server.
```

### DIFF
--- a/components/server/src/express/ws-connection-handler.ts
+++ b/components/server/src/express/ws-connection-handler.ts
@@ -10,6 +10,7 @@ import { Disposable, DisposableCollection } from "@gitpod/gitpod-protocol";
 import { repeat } from "@gitpod/gitpod-protocol/lib/util/repeat";
 import { log } from "@gitpod/gitpod-protocol/lib/util/logging";
 import { WsNextFunction, WsRequestHandler } from "./ws-handler";
+import { ClientMetadata } from "../websocket/websocket-connection-manager";
 
 /**
  * This class provides a websocket handler that manages ping-pong behavior for all incoming websocket requests.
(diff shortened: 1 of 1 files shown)


roboquat:

<details>

</details>

components/server/src/express/ws-connection-handler.ts
```diff
@@ -30,19 +30,16 @@ export class WsConnectionHandler implements Disposable {
...
-                                }
+                            // (AT) for unknown reasons the expected `close` event was not emitted, otherwise this websocket would
+                            // no longer be contained in the list of clients iterated over. make sure we release all resources bound
+                            // to this client connection.
+                            const emitClose = (ws as any).emitClose;
+                            if (typeof emitClose === "function") {
+                                this.clients.delete(ws);
+                                emitClose();
+                                log.warn(
+                                    "websocket in strange state. forcefully emitting a close event to release resources.",
```
easyCZ, line 41: A metric would be better here than a log. Alternatively, enrich it with some data - ie Connection ID to make it useful for tracking down which connection.

AlexTugarev: Reading on https://github.com/websockets/ws/blob/master/lib/websocket.js#L245-L259, we shouldn't even end up in this situation. 

What kind of metric would you suggest? Assuming this is a bug, we want to log it. +1 for logging more details on the connection. Not sure which id is meant, I'll go with full `ClientMetadata`.

AlexTugarev: done

easyCZ: Basically it comes down to what is it you're trying to observe in the system. Presumably you're logging it so that you can get an idea that it is happening. For that use-case, a counter metric you add to a dashboard is a better choice - it's cheaper and doesn't require deep context.

If, however, this is just cleanup and we don't really care about how often, or details, of the problem, we might as well not log at all.

easyCZ (approved):

PR #10200, opened 2022-05-23
Allow configuration of s3 region for registry backend

### LINKED ISSUE #9800: KOTS incluster registry S3 dialog doesn't request region
### Bug description

The S3 region defaults to "local" in the installation if one is just using S3 for the the registry backend and not anything else. In order to fix this, the deployment can be edited by hand, but a subsequent redeployment by kots when changing other settings will remove the change.

Introduced by https://github.com/gitpod-io/gitpod/pull/9701

### Steps to reproduce

Install in AWS using the getting starting guide and provide S3 credentials for the incluster registry. 

`kubectl edit deployment registry -n gitpod` 

Change region to correct region, restart just the registry deployment and it works, but doing a redeployment if making another change (or updating) will revert the change.

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

nandajavarma:
## Description

This PR adds an option to configure the bucket region. Alongside it also ensures the secret is created even when the registry is set to `inCluster` since s3 backend can be set for `inCluster` registry.

## Related Issue(s)

Fixes #9800

## How to test

Use `S3` as the `In-cluster Storage provider` in KOTS config UI. You will see the option to set the `Region` for S3

## Release Notes

```release-note
[kots] Add option to set the `S3` bucket region used for registry backend
```


princerachit:
Hi @nandajavarma thanks for the PR. I think unlike self hosted team everyone would not have enough knowledge to operate using kots. e.g. Workspace team heavily relies on preview env and workspace-preview cluster for testing. For PRs like these we (at least me) would rely heavily on the testing performed by the author. Adding a screenshot or a short loom video can help me review it faster. What do you think?

PR #10216, opened 2022-05-24
[jb] enable connect button only when JetBrains Client not activated (cont.)

yaohui-wyh:
## Description

A cleanup PR for #10177

## Related Issue(s)

Fixes #

## How to test
 

Build & Install the plugin following
https://github.com/gitpod-io/gitpod/blob/main/components/ide/jetbrains/gateway-plugin/README.md#L8

## Release Notes

```release-note
NONE
```

## Documentation

- [x] /werft no-preview=true

### DIFF
--- a/components/ide/jetbrains/gateway-plugin/src/main/kotlin/io/gitpod/jetbrains/gateway/GitpodConnectionProvider.kt
+++ b/components/ide/jetbrains/gateway-plugin/src/main/kotlin/io/gitpod/jetbrains/gateway/GitpodConnectionProvider.kt
@@ -11,9 +11,9 @@ import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.intellij.ide.BrowserUtil
 import com.intellij.openapi.Disposable
 import com.intellij.openapi.components.service
-import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.diagnostic.thisLogger
 import com.intellij.remote.RemoteCredentialsHolder
+import com.intellij.remoteDev.util.onTerminationOrNow
 import com.intellij.ssh.AskAboutHostKey
 import com.intellij.ssh.OpenSshLikeHostKeyVerifier
 import com.intellij.ssh.connectionBuilder
@@ -35,7 +35,6 @@ import com.jetbrains.gateway.ssh.ClientOverSshTunnelConnector
 import com.jetbrains.gateway.thinClientLink.ThinClientHandle
 import com.jetbrains.rd.util.URI
 import com.jetbrains.rd.util.lifetime.Lifetime
-import com.jetbrains.rd.util.lifetime.isAlive
 import com.jetbrains.rd.util.lifetime.onTermination
 import io.gitpod.gitpodprotocol.api.entities.WorkspaceInstance
 import io.gitpod.jetbrains.icons.GitpodIcons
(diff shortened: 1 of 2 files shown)


roboquat:
@yaohui-wyh: Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://git.k8s.io/community/contributors/guide/release-notes.md) to remove it.

<details>

</details>

yaohui-wyh:
PTAL @akosyakov @felladrin

components/ide/jetbrains/gateway-plugin/src/main/kotlin/io/gitpod/jetbrains/gateway/GitpodConnectionProvider.kt
```diff
@@ -259,27 +270,22 @@ class GitpodConnectionProvider : GatewayConnectionProvider {
...
-                                            dispatcher.multicaster.didChange()
                                             if (client.clientPresent) {
                                                 statusMessage.text = ""
                                             }
                                         }
                                     }
                                     thinClient = client
-                                    jetbrainsClientMap[update.workspaceId] = client
+                                    setClient(update.workspaceId, client)
+                                    connectionLifetime.onTermination {
```
akosyakov, line 286: I wonder whether it can race that by now it is already terminated. Would not it be better to use `onTerminationOrNow`? Would will happen if `onTerminationOrNow` through not alive exception?

yaohui-wyh: > I wonder whether it can race that by now it is already terminated. Would not it be better to use `onTerminationOrNow`?

How about using `client.lifetime` instead of `connectionLifetime`? I think it might be more actuate to use each ThinClient's `lifetime` to handle termination event. I think there is no such race condition for `client.lifetime`
```suggestion
                                    client.lifetime.onTerminationOrNow { setClient(update.workspaceId, null) }
```

> Would will happen if `onTerminationOrNow` yield not alive exception?

I think `setClient(update.workspaceId, null)` in the nearest catch block would be fine. What do you think?

akosyakov:
I wonder is it really a solution? You can still get with multiple connection windows while workspace is starting or stopping?

yaohui-wyh:
> I wonder is it really a solution? You can still get with multiple connection windows while workspace is starting or stopping?
> I still was able to cause such conditions as from browser as from GW if i click several times while workspace is starting. 

Yeah, this PR (together with PR #10177) can only prevent the user from clicking the "Connect" button (unintentionally) when JBClient is alive, which is indeed not a perfect solution (not able to cover the transitioning states such as starting & stopping). Sorry for not discussing a solution before-head, feel free to close the PR if you reach a conclusion. ❤️

akosyakov:
We could for instance try to fail if one is calling a connection provider but we already handling such URL. Ideally we should bring to focus client or connection window in such case. Maybe we at least have a control over focus for the connection window?

yaohui-wyh:
> is it ok for you if we close this PR?

Yeah I agree that a better approach should be activating the running Thin Client, and since you have discussed it with the JetBrains Team, I think it's better to wait until such API is provided and do the enhancement then.
Thanks for all the suggestions @akosyakov @felladrin during the prev PR #10177 and this one. And again, sorry for the opinionated changes and for costing a lot of time for @felladrin back and forth. ❤️

PR #9786, opened 2022-05-05
Use a special domain name for SSH Gateway

iQQBot:
## Description

Use a special domain name for SSH Gateway, This has a big advantage for self-hosted users, who can set up another loadbalancer for this separate domain without breaking their original loadbalancer or ingress

## Related Issue(s)

Fixes #

## How to test

1. Download this special Gateway Plugin https://plugins.jetbrains.com/plugin/18438-gitpod-gateway/versions/Dev/172691
2. Open a workspace with any Jetbrains Client
3. Click More Action -> Connect via SSH, make sure hostname part of ssh command is `workspaceId.ssh.pd-simple-ssh-host.preview.gitpod-dev.com`
4. 
image

image

5. Click Open with xxx Jetbrains IDE to open Jetbrains Gateway to ensure everything should works as normal

## Release Notes

```release-note
Use a special domain name for SSH Gateway
```

PR #10221, opened 2022-05-24
[dev] Separate ssh IP and ws-proxy IP

iQQBot:
## Description

In https://github.com/gitpod-io/gitpod/pull/9786 we introduce a special domain for ssh gateway
This PR change harvester base preview environment use this special domain
This has some benefit
1. Save some money, before, accessing workspace pages also required the use of loadbalancer, which consumed some unnecessary traffic, after this PR, all non-SSH traffic will no longer go through loadbalancer
2. To improve some stability, `kubectl port-forward` is not very stable, for example, vm sometimes need to rebuild, then `kubectl` will time out, and the pod will restart, this restart time will have exponential growth, we do not want to bring this instability to the users who do not use SSH Gateway

## Related Issue(s)

Fixes #

## How to test

1. start a workspace, open browser IDE it should work
2. connect with ssh, it should be work too

## Release Notes

```release-note
NONE
```


mads-hartmann (approved):
Thanks!

PR #9337, opened 2022-04-14
[server] Use image-builder from workspace cluster

csweichel:
## Description
This PR makes server use the image builder from workspace cluster.

**Note: ** merge and deploy this only once #9335 is merged, deployed and running.

## Related Issue(s)

Fixes #9248

## How to test
Start a workspace which requires an image-build

## Release Notes

```release-note
Use image-builder from workspace cluster
```

 - [x] /werft run with-vm=true

PR #10235, opened 2022-05-25
[image-builder-bob] Prioritize using additional auth

princerachit:
## Description

<del>This implements the PR https://github.com/gitpod-io/gitpod/pull/10094 in a different way. Instead of changing the names of env vars, override existing vars with additional auth. I am not sure how this will impact gitpod installations. AFAIK additional auth is only used in the context of user supplied auth, so using it instead of existing configured auth should make sense.</del>

Prefer merging https://github.com/gitpod-io/gitpod/pull/9337 first as that will give us more control in testing and debugging issues related to this PR or any changes related to image-builder in general without impacting production env.

## Related Issue(s)

The initial revert was done due to an incident see: [INC-150](https://app.incident.io/incidents/150)
https://github.com/gitpod-io/gitpod/issues/10089

## How to test

Refer to original PR https://github.com/gitpod-io/gitpod/pull/10094

## Release Notes

```release-note
Fix conflicting auth selection for image-builder-bob
```

### DIFF
--- a/components/image-builder-bob/cmd/proxy.go
+++ b/components/image-builder-bob/cmd/proxy.go
@@ -39,7 +39,7 @@ var proxyCmd = &cobra.Command{
 		if err != nil {
 			log.WithError(err).WithField("auth", proxyOpts.Auth).Fatal("cannot unmarshal auth")
 		}
-		authP = authP.AddIfNotExists(authA)
+		authP = authP.OverrideWith(authA)
 
 		baseref, err := reference.ParseNormalizedNamed(proxyOpts.BaseRef)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


princerachit:
/hold

Prefer merging https://github.com/gitpod-io/gitpod/pull/9337 first

princerachit:
> Hi @princerachit If I remember correctly, `base` and `target` are both our registry, they are just used for different purposes, and base does not refer to FROM in the `dockerfile`

Let me correct the naming

princerachit:
This is the PR which configures using a private registry to pull image. The caveat is, you can not use it in dockerfile.

https://github.com/gitpod-io/gitpod/pull/8550

If you look at the PR the additional auth supplied should be used to pull the image. But in a special case when such config already exist in the pull secret then this causes a conflict of credential selection.

I think the changes I just pushed cannot really address it and would rather cause other issues when trying to push to container registry. We probably need to separate out the variables to distinguish user provided auth for image pulling vs auth for container registry used by gitpod.

sagor999:
not stale

princerachit:
Closing as this is not the right approach to fix the issue.

PR #10251, opened 2022-05-25
[dev-image] Add MySQL Server to workspace image

easyCZ:
## Description

Adds MySQL server to our workspace image. This allows us to run unit-tests against a local instance of MySQL instead of always depending on a preview environment (once tests are updated to use it).

The setup mirrors the container image in our [gitpodify guide](https://www.gitpod.io/blog/gitpodify#mysql). 
There may be a cleaner way, and I'm all ears :)

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Documentation

NONE

### DIFF
--- a/dev/image/BUILD.yaml
+++ b/dev/image/BUILD.yaml
@@ -9,6 +9,7 @@ packages:
     srcs:
       - gcloud-default-config
       - kubeconfig.yaml
+      - mysql/*
     config:
       dockerfile: Dockerfile
       image:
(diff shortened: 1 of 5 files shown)


mustard-mh:
why not using mysql docker?

some command like below maybe better
```bash
docker run --name mysql  -p 3306:3306 -e MYSQL_ROOT_PASSWORD=gitpod -d mysql:8
```

PR #10264, opened 2022-05-25
Fixing (missing) traces

mads-hartmann:
### TL;DR 
This PR resolves the mystery of the (missing) parent spans that we have seen in Honeycomb. If you don't close a span, the span will never be sent to the tracing backend (Honeycomb) and thus, for all children of the span, Honeycomb will see that the spans contains a parent ID that points to a span it never received and will render the parent as (missing). So to fix the issue, we "simply" have to ensure that we **always** close spans.

This PR fixes a lot of such cases but not all.

### Context
This has been a long-standing issue. Since migrating to Honeycomb from Jaeger we've noticed the problem more as Honeycomb displays these kinds of error more prominently, but this issue has existed since the start of time[^1], it's just that Jaeger didn't call it out as much - see comment with a screenshot comparing Jaeger and Honeycomb [here](https://github.com/gitpod-io/ops/issues/385#issuecomment-984633422).

### Details

This PR contains a lot of fixes for places where we forgot to close spans in some circumstances. I wasn't able to track down all the places but I'm opening up this PR now as this already produces considerable improvements to the "createWorkspace" trace.

See a [trace before](https://ui.honeycomb.io/gitpod/datasets/preview-environments/result/qE8ttJtvjBH/trace/BgMYToxuVJy?span=daa8a5408ac1549f) and [trace after](https://ui.honeycomb.io/gitpod/datasets/preview-environments/result/naqtGFVAwev/trace/uJ1jkUyqhWd?span=fca7b94d1f2681fe) -- you'll notice that the trace looks much cleaner now. That's because before Honeycomb didn't know where in the trace to inject the spans since the parent was missing, so they were all just being rendered as children of the root span. Now that we have the parent they can be placed correctly so we get trace that's much more meaningful and easier to navigate. One nice example of this is the "sub-tree" of `/ws-manager/startWorkspace` as shown below:

Before

Screenshot 2022-05-27 at 12 10 47

After

Screenshot 2022-05-27 at 12 10 40

### What (missing) spans are left

I wasn't able to track down a few missing spans still, mostly around the content-initializer, messagebus part of server, and one or two MySQL queries (see screenshots below). We can tackle these in a follow up PR.

Screenshot 2022-05-27 at 12 16 52
Screenshot 2022-05-27 at 12 16 56
Screenshot 2022-05-27 at 12 17 04

## Related Issue(s)

Fixes https://github.com/gitpod-io/ops/issues/2318

## How to test

I have been testing this in a regular preview environment.

## Release Notes

```release-note
NONE
```

## Documentation

N/A

[^1]: The start of time in our case is of course https://github.com/gitpod-io/gitpod/commit/24f7b609bf79b247d72d1841282639726bcd5891

### DIFF
--- a/components/ws-manager/pkg/manager/create.go
+++ b/components/ws-manager/pkg/manager/create.go
@@ -881,6 +881,7 @@ func (m *Manager) newStartWorkspaceContext(ctx context.Context, req *api.StartWo
 
 	workspaceSpan := opentracing.StartSpan("workspace", opentracing.FollowsFrom(opentracing.SpanFromContext(ctx).Context()))
 	traceID := tracing.GetTraceID(workspaceSpan)
+	defer tracing.FinishSpan(workspaceSpan, &err)
 
 	clsName := req.Spec.Class
 	if _, ok := m.Config.WorkspaceClasses[req.Spec.Class]; clsName == "" || !ok {
--- a/components/common-go/tracing/tracing.go
+++ b/components/common-go/tracing/tracing.go
@@ -110,7 +110,8 @@ func ApplyOWI(span opentracing.Span, owi logrus.Fields) {
 	}
 }
 
-// GetTraceID extracts the ueber-trace-id from the context
+// GetTraceID extracts the uber-trace-id from the context which encodes
+// {trace-id}:{span-id}:{parent-span-id}:{flags}
 func GetTraceID(span opentracing.Span) string {
 	var buf bytes.Buffer
 	err := opentracing.GlobalTracer().Inject(span.Context(), opentracing.Binary, &buf)
--- a/components/content-service/pkg/initializer/initializer.go
+++ b/components/content-service/pkg/initializer/initializer.go
@@ -60,8 +60,9 @@ func (e *EmptyInitializer) Run(ctx context.Context, mappings []archive.IDMapping
 type CompositeInitializer []Initializer
 
 // Run calls run on all child initializers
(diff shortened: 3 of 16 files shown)


geropl (approved):
Thanks a lot @mads-hartmann ! :orange_heart: Would be really awesome to have lints for these, as it's hard to come by it in code reviews.

Code LGTM, and workspace starts! :+1:

sagor999 (approved):

components/ws-manager/pkg/manager/create.go
```diff
@@ -881,6 +881,7 @@ func (m *Manager) newStartWorkspaceContext(ctx context.Context, req *api.StartWo
 
 	workspaceSpan := opentracing.StartSpan("workspace", opentracing.FollowsFrom(opentracing.SpanFromContext(ctx).Context()))
 	traceID := tracing.GetTraceID(workspaceSpan)
+	defer tracing.FinishSpan(workspaceSpan, &err)
```
csweichel, line 884: This not great - the intent of this span is to live over the course of the workspace lifecycle until the workspace hits `running`.

csweichel:
> I'm in favour of "Introduce a new span"

Very much agree. We should then remove the old (misguided) "try and persist the span until workspace is running" approach. Central to that code is [`traceWorkspace(...)`](https://github.com/gitpod-io/gitpod/blob/15cce5f092604dd30df24e86458222815bc73aca/components/ws-manager/pkg/manager/monitor.go#L518) and the [traceID annotation](https://github.com/gitpod-io/gitpod/blob/28635821880f37c5f26a8780380afe621456b4c3/components/common-go/kubernetes/kubernetes.go#L41).

PR #10267, opened 2022-05-25
volume snapshot delete and restore support

### LINKED ISSUE #10259: ws-manager: add support for deleting snapshotVolumes (to clean up old backups) on workspace delete

sagor999:
## Description

This PR adds two features:
1. New workspace volume snapshot is created. We need to delete old volume snapshot for that workspace now.
2. Open workspace in the cluster where volume snapshot doesn't exist yet. We need to restore volume snapshot.
3. Workspace is deleted: it will garbage collect all volume snapshots associated with that workspace.

Fixes #10259 

## How to test

Loom video showing this in action: 
https://www.loom.com/share/51b41802054d4de89902aaa358171624

## Release Notes

```release-note
none
```


roboquat:

<details>

</details>

components/ws-manager-api/core.proto
```diff
@@ -40,6 +40,8 @@ service WorkspaceManager {
 
     // controlAdmission makes a workspace accessible for everyone or for the owner only
     rpc ControlAdmission(ControlAdmissionRequest) returns (ControlAdmissionResponse) {}
+
```
jenting, line 43: Nit: it'd be good to have a comment describing this function.

components/ws-manager/pkg/manager/manager.go
```diff
@@ -346,6 +361,117 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
...
+				Namespace: m.Config.Namespace,
+			},
+			DeletionPolicy: "Delete",
+			Source: volumesnapshotv1.VolumeSnapshotContentSource{
+				SnapshotHandle: &handle,
+			},
+			Driver: volumeSnapshotClass.Driver,
+		},
+	}
+
```
jenting, line 425: Nit: We could consider add VolumeSnapshotClassName to VolumeSnapshotContent as well.
```go
if volumeSnapshotClassName != "" {
	volumeSnapshotContent.Spec.VolumeSnapshotClassName = &volumeSnapshotClassName
}
```

components/ws-manager/pkg/manager/manager.go
```diff
@@ -346,6 +361,117 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
 	return okResponse, nil
 }
 
+func (m *Manager) restoreVolumeSnapshotFromHandle(ctx context.Context, id, handle string) (err error) {
+	span, ctx := tracing.FromContext(ctx, "restoreVolumeSnapshotFromHandle")
+	defer tracing.FinishSpan(span, &err)
+	//log := log.WithField("func", "restoreVolumeSnapshotFromHandle")
```
jenting, line 367: nil: we could consider remove this line

components/ws-manager/pkg/manager/manager.go
```diff
@@ -346,6 +361,117 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
...
+	}
+
+	return nil
+}
+
+func (m *Manager) DeleteVolumeSnapshot(ctx context.Context, req *api.DeleteVolumeSnapshotRequest) (res *api.DeleteVolumeSnapshotResponse, err error) {
+	span, ctx := tracing.FromContext(ctx, "DeleteVolumeSnapshot")
+	tracing.LogRequestSafe(span, req)
+	defer tracing.FinishSpan(span, &err)
+	log := log.WithField("func", "DeleteVolumeSnapshot")
```
jenting, line 438: nit: we could consider remove this line

components/ws-manager/pkg/manager/manager.go
```diff
@@ -346,6 +361,117 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
...
+
+	okResponse := &api.DeleteVolumeSnapshotResponse{}
+
+	var volumeSnapshot volumesnapshotv1.VolumeSnapshot
+	err = m.Clientset.Get(ctx, types.NamespacedName{Namespace: m.Config.Namespace, Name: req.Id}, &volumeSnapshot)
+	if k8serr.IsNotFound(err) {
+		if !req.SoftDelete {
+			err = m.restoreVolumeSnapshotFromHandle(ctx, req.Id, req.VolumeHandle)
```
jenting, line 446: Not quite understand that _why_ do we need to support soft delete?
The purpose seems like we will restore the VolumeSnapshotContent as well as VolumeSnapshot from the GCP snapshots, then delete the VolumeSnapshot object. Therefore, the VolumeSnapshotContent remains.

Could you please elaborate more for the purpose?

sagor999: this is for a future PR:
when we need to remove workspace and all associated volume snapshots for it, that could be spread over multiple clusters. 
So I was thinking I could send request to soft delete to all clusters except one that will actually perform hard delete.

components/ws-manager/pkg/manager/manager.go
sagor999: Not sure I understood what you meant.

components/ws-manager/pkg/manager/manager.go
sagor999: why this one? I like to have a field with func name where the log entry is coming from, as sometimes our log output text can be repeated in multiple places and then it might be difficult at times to figure out where this specific entry is coming from.

jenting: Okay, that makes sense. Let's keep it as it is.

components/ws-manager/pkg/manager/manager.go
jenting: Because we add volumeSnapshotClassName to VolumeSnapshot object on lines 397-399.
I'm thinking that we could add volumeSnapshotClassName to VolumeSnapshotContent object as well. (Although if we don't specify the VolumeSnapshotClass, the Kubernetes uses the default one within the cluster).

geropl:
Starting to review now

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -371,11 +375,32 @@ export class WorkspaceManagerBridge implements Disposable {
...
                     .findVolumeSnapshotById(status.conditions.volumeSnapshot.volumeSnapshotName);
                 if (existingSnapshot === undefined) {
-                    await this.workspaceDB.trace(ctx).storeVolumeSnapshot({
+                    const storedVolumeSnapshot = await this.workspaceDB.trace(ctx).storeVolumeSnapshot({
                         id: status.conditions.volumeSnapshot.volumeSnapshotName,
                         creationTime: new Date().toISOString(),
                         volumeHandle: status.conditions.volumeSnapshot.volumeSnapshotHandle,
                     });
+
+                    // since we stored most recent volume snapshot now, delete any old volume snapshots
```
geropl, line 384: Uff. IMo the call _back_ to ws-manager-bridge for cleanup purposes is something we should _not_ handle here. :confused: It's already complicated and untestable as-is, but we should try hard to not make it worse.

Instead, I propose we have add a garbage-collection mechanism to `server`, in the same style that we use for workspaces, for instance. Optionally, this can be triggered by "workspace stopped" events. But has the additional benefit of being more reliable.

@sagor999 WDYT?

components/ws-manager/pkg/manager/manager.go
sagor999: @jenting there is an example of soft delete usage now in this PR.

components/ws-manager-bridge/src/bridge.ts
sagor999: @geropl feedback addressed!

sagor999:
/hold
to ensure I get review from webapp team.

components/ws-manager/pkg/manager/manager.go
```diff
@@ -346,6 +361,119 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
...
+				Name:      req.Id,
+				Namespace: m.Config.Namespace,
+			},
+		},
+	)
+	if err != nil && !k8serr.IsNotFound(err) {
+		log.WithError(err).Errorf("failed to delete volume snapshot `%s`", req.Id)
+		return nil, err
+	}
+	if !k8serr.IsNotFound(err) {
```
jenting, line 470: Should this one be `if k8serr.IsNotFound(err) {`?

components/server/src/workspace/garbage-collector.ts
```diff
@@ -124,4 +127,27 @@ export class WorkspaceGarbageCollector {
             span.finish();
         }
     }
+
+    // finds volume snapshots that have been superceded by newer volume snapshot and removes them
+    protected async deleteOutdatedVolumeSnapshots() {
+        const span = opentracing.globalTracer().startSpan("deleteOutdatedVolumeSnapshots");
+        try {
+            const workspaces = await this.workspaceDB.trace({ span }).findVolumeSnapshotWorkspacesForGC();
+            const volumeSnapshots = await Promise.all(
```
geropl, line 136: Note: This is potentially unbound, and might blow up the whole process as we grow.
The other deletion mechanisms feature a "limit" which I think we could re-use here.

components/server/src/workspace/workspace-deletion-service.ts
```diff
@@ -94,8 +99,79 @@ export class WorkspaceDeletionService {
...
+            TraceContext.setError({ span }, err);
+            throw err;
+        } finally {
+            span.finish();
+        }
         return true;
     }
+
+    /**
+     * Perform deletion of volume snapshot from all clusters and from gloud provider:
```
geropl, line 122: > gloud

cloud

components/server/src/workspace/workspace-deletion-service.ts
```diff
@@ -94,8 +99,79 @@ export class WorkspaceDeletionService {
...
+        try {
+            const allClusters = await this.workspaceManagerClientProvider.getAllWorkspaceClusters();
+            // we need to do two things here:
+            // 1. we want to delete volume snapshot object from all workspace clusters
+            // 2. we want to delete cloud provider source snapshot
+            let wasDeleted = false;
+            let index = 0;
+
+            let availableClusters = allClusters.filter((c) => c.state === "available");
+            for (let cluster of availableClusters) {
```
geropl, line 140: I expected this method to connect to a specific ws-manager based on a "region" field, either retrieved from `VolumeSnapshot`, `Workspace` or `WorkspaceInstance`. Iterating over all currently connected clients makes it impossible to detect cases where we are no longer connected to a certain ws-manager.

Or is this due to the fact that we might have multiple ws-manager (with different names) in the same GCloud region, and we do not know how to map those? :thinking:

geropl:
@sagor999 Regarding merging of this PR: Once the initial review is through, it'd would be awesome to decouple this PR into smaller chunks. For instance:
 - DB changes
 - ws-manager changes
 - server+bridge changes

components/gitpod-db/src/typeorm/migration/1654628106102-VolumeSnapshotAddWSId.ts
```diff
@@ -0,0 +1,21 @@
...
+import { MigrationInterface, QueryRunner } from "typeorm";
+import { tableExists } from "./helper/helper";
+
+export class VolumeSnapshotAddWSId1654628106102 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        if (await tableExists(queryRunner, "d_b_volume_snapshot")) {
+            await queryRunner.query("DROP TABLE `d_b_volume_snapshot`");
+        }
+        await queryRunner.query(
+            `CREATE TABLE IF NOT EXISTS d_b_volume_snapshot (  id char(36) NOT NULL, workspaceId char(36) NOT NULL, creationTime timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6), volumeHandle varchar(255) NOT NULL, PRIMARY KEY (id), KEY ind_5iziusbz752p4jv3e8erq3wc7n (workspaceId), KEY ind_dbsync (creationTime)) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;`,
```
geropl, line 16: > ind_5iziusbz752p4jv3e8erq3wc7n

We usually use readable names ala `ind_workspaceId` :+1:

components/gitpod-db/src/typeorm/workspace-db-impl.ts
```diff
@@ -760,6 +760,30 @@ export abstract class AbstractTypeORMWorkspaceDBImpl implements WorkspaceDB {
...
+        const results = (await qb.getRawMany()) as Pick<DBVolumeSnapshot, "workspaceId">[];
+        return results.map((vs) => vs.workspaceId);
+    }
+
+    // returns the list of all volume snapshots for specific workspace id
+    public async findVolumeSnapshotForGCByWorkspaceId(wsId: string): Promise<VolumeSnapshot[]> {
+        const volumeSnapshotRepo = await this.getVolumeSnapshotRepo();
+        const results = await volumeSnapshotRepo
+            .createQueryBuilder("vs")
+            .where("vs.workspaceId = :wsId", { wsId })
```
geropl, line 781: Checking this query it again it might be necessary to instead of introduce `ind_workspaceId`, make it a combined 2ndary key: `ind_workspaceId_creationTime (workspaceId, creationTime)`. This we guarantee it fits here, but can also be used for workspaceId alone. :+1:

components/ws-manager/pkg/manager/manager.go
sagor999: no. we want to report back if we indeed deleted snapshot volume. if it was not found, then nothing was actually deleted.

components/server/src/workspace/workspace-deletion-service.ts
sagor999: Not quite. Imagine following scenario:
EU cluster where WS was started from Volume Snapshot X
US cluster where WS was also started from same Volume Snapshot X (from prebuild for example)

Then we want to GC that volume snapshot.
We need to do two things now:
1. Ensure that we delete k8s object representing this snapshot from ALL clusters (to make sure we do not accumulate those over time, as it will affect k8s api server)
2. Ensure we delete volume snapshot object from cloud provider

So this code ensures that. That is why it attempts to talk to all existing clusters and clean it up from all of them. 
If we are no longer connected to some cluster, I think that is fine, as I would assume that cluster is getting decommissioned.

components/server/src/workspace/workspace-deletion-service.ts
sagor999: Fixed

components/server/src/workspace/garbage-collector.ts
sagor999: Good point. Added limit. Reusing existing config param for this, assuming it is fine.

components/gitpod-db/src/typeorm/workspace-db-impl.ts
sagor999: Fixed.

components/gitpod-db/src/typeorm/migration/1654628106102-VolumeSnapshotAddWSId.ts
sagor999: Fixed.

jenting (approved):
LGTM 🛹 🚀

components/server/src/workspace/workspace-deletion-service.ts
geropl: > We need to do two things now:

Ah, that explains it. Just to ensure we're on the same page: `ws-manager` does _not_ remove the PVC object from the control plane when the corresponding workspace is stopped?

> Ensure we delete volume snapshot object from cloud provider
I thought we just need to do this, and also wondered if this was something we put into content-service...? :thinking:

geropl: Sorry to "block" the merging of the PR here, but I feel I'm still missing some context to understand why the PVC lifecycle looks the way it does.

jenting: > Just to ensure we're on the same page: `ws-manager` does _not_ remove the PVC object from the control plane when the corresponding workspace is stopped?

No, when the workspace stopped, the PVC object would be removed after the VolumeSnapshot object is ready to use. (Which means the PVC snapshot is done).

> > Ensure we delete volume snapshot object from cloud provider
> 
> I thought we just need to do this, and also wondered if this was something we put into content-service...? 🤔

The lifecycle of PersistentVolumeClaim/VolumeSnapshot is managed by ws-manager.

sagor999: @geropl 
>Just to ensure we're on the same page: ws-manager does not remove the PVC object from the control plane when the corresponding workspace is stopped?

it does. There are two objects in play here. PVC is an actual storage object, it is only alive while workspace is running. When workspace is stopped, wsman will "convert" PVC into VolumeSnapshot object that is stored in cloud provider (outside of k8s control).

When we open workspace, PVC is created and restored from VolumeSnapshot object.

When it is time to delete\GC volume snapshot, we might have same VolumeSnapshot across multiple clusters (prebuild that was opened in different clusters for example). So we want to clean up that volume snapshot object from all clusters (as otherwise we will be accumulating a lot of them over the lifecycle of the cluster), and then also remove actual snapshot from cloud provider as well.

All this work is done by wsman currently.

It will be better to move that into content service indeed, but currently it is not possible, because content service still lives in webapp cluster, and should be migrated into workspace clusters first.

jenting:
Hey Pavel, I re-run this PR to test; I found that after the workspace stopped, I clicked `Delete Workspace` on the GUI. However, the VolumeSnapshot is still there. So I suspect the VolumeSnapshot should be deleted by the ws-manager as well. Correct? Also, I see the loom video doesn't demo this part.

aledbf (approved):

sagor999:
@jenting it will be deleted when GC runs and deletes workspace for real. 
I simulated that part manually by updating workspace delete timestamp to be one week old, then GC ran and cleaned everything up. 
But thank you for double checking! ❤️

components/server/src/workspace/workspace-deletion-service.ts
geropl: Thx @sagor999 for syncing! :+1: 
I understood that:
 - workspace takes care of deleting the actual [disk snapshots](https://cloud.google.com/compute/docs/disks/snapshots?hl=en)
 - we keep the cluster-local VolumeSnapshots and VolumeSnapshotContent in the cluster to optimize for fast workspace starts
 - but to avoid cluttering the clusters too much, we aim to remove those here whenever their disk snapshot get deleted

It would be nice if we added some comments along those lines to the GC. :pray:

components/server/src/workspace/workspace-deletion-service.ts
```diff
@@ -94,8 +99,81 @@ export class WorkspaceDeletionService {
...
+                const client = await this.workspaceManagerClientProvider.get(cluster.name);
+                const req = new DeleteVolumeSnapshotRequest();
+                req.setId(vs.id);
+                req.setVolumeHandle(vs.volumeHandle);
+
+                let softDelete = true;
```
geropl, line 148: nit: Already mentioned it, but just to highlight: Especially this part feels a tad brittle, and it would be nice to find a better way to handle/encapsulate that logic.

geropl (approved):
LGTM :+1:

PR #10274, opened 2022-05-26
[ws-daemon] Limit the number of concurrent workspace backups

aledbf:
## Description

The creation of workspace backups and upload to the object storage can consume

## How to test

- Start multiple workspaces
- Stop all the workspaces
- Check ws-daemon logs for messages related to `waiting time for concurrent backups to finish was XXXXX`

## Release Notes
```release-note
[ws-daemon] Limit the number of concurrent workspace backups to three to avoid excessive memory utilization
```

### DIFF
--- a/components/ws-daemon/pkg/content/service.go
+++ b/components/ws-daemon/pkg/content/service.go
@@ -396,6 +396,9 @@ func (s *WorkspaceService) DisposeWorkspace(ctx context.Context, req *api.Dispos
 	return resp, nil
 }
 
+// channel to limit the number of concurrent backups and uploads.
+var backupWorkspaceLimiter = make(chan bool, 3)
+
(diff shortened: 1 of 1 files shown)


jenting (commented):
Can we improve the release note to specify the number of concurrent limits?

The rest /lgtm

components/ws-daemon/pkg/content/service.go
```diff
@@ -396,6 +396,9 @@ func (s *WorkspaceService) DisposeWorkspace(ctx context.Context, req *api.Dispos
 	return resp, nil
 }
 
+// channel to limit the number of concurrent backups and uploads.
+var backupWorkspaceLimiter = make(chan bool, 3)
```
utam0k, line 400: I think it would be better to make the number be able to control with config. WDYT?

jenting (approved):
LGTM

PR #10283, opened 2022-05-26
Update x/sys for kubecdl

vulkoingim:
## Description
Update `golang.org/x/sys` so it builds on Mac with go 1.18+

## Release Notes

```release-note
NONE
```


ArthurSens (approved):
Hmmmm interesting that workspace-team is not set as codeowners. (If I'm not mistaken they are the ones who created and actively maintain this binary 🙂)


I'm approving because the change doesn't seem like it's gonna break anything, but tagging @gitpod-io/engineering-workspace just for awareness!

jenting (approved):

PR #10297, opened 2022-05-27
[dashboard] hide feedback form from self-hosted users 10296

### LINKED ISSUE #10296: Hide Feedback Form for self-hosted
Self-hosted are not sending tracking events, therefore we should hide the feedback form from those users.

laushinka:
## Description
Hides the feedback form from self-hosted users.

## Related Issue(s)
Fixes #10296

## Release Notes

```release-note
Feedback form only shows for SaaS gitpod-io users.
```


easyCZ (approved):

laushinka:
Thanks, @easyCZ!

laushinka:
I'm sure this code works and it's reviewed already too, but just to be sure, @lucasvaltl I assume you have a self-hosted instance that you use to test? Is it worth testing this PR with that? 🙏🏽 I'd also like to know how to test it for future reference..

PR #10308, opened 2022-05-27
Add a Stripe settings secret to server

jankeromnes:
## Description

Add a Stripe settings secret to server, in order to provide the secret key for API calls.

## Related Issue(s)

Fixes #

## How to test

- Builds **without** payment should succeed (and **not** have a Stripe volume mounted on `server` pod)
- Build **with** payment should succeed, **and** have the Stripe volume mounted:

```
$ kubectl describe pod server-5995b76459-llfwg
[...]
Volumes:
[...]
  stripe-config:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  stripe-config
    Optional:    false
```

## Release Notes

```release-note
NONE
```

## Documentation

### DIFF
--- a/.werft/jobs/build/payment/stripe-config-secret.yaml
+++ b/.werft/jobs/build/payment/stripe-config-secret.yaml
@@ -1,0 +1,8 @@
+apiVersion: v1
+data:
+  settings: eyJwdWJsaXNoYWJsZUtleSI6InBrX3Rlc3RfNTFLeHVyN0dhZFJYbTUwbzNJNXJKQTNvbnkxdGNmdTNkM0NOd3BUWFR6QURkWTJISmlvRk1XTGdTa2M1d2h0UkZRam85UG5kM3pYYUdlcktQcXRmN0REQ3kwMFhBb01kbjZhIiwic2VjcmV0S2V5Ijoic2tfdGVzdF81MUt4dXI3R2FkUlhtNTBvM0NtVFJWc1Q2Q0xqd0VlSlhsWWtmdjZHajREQm42aVlVeDJQWUlUNDhjVlI5dlNUS0s1b2hwQTVCdWdycU5NUU9WVzN0NVJIODAwS011T3lEZ1QifQo=
+kind: Secret
+metadata:
+  name: stripe-config
+  namespace: ${NAMESPACE}
+type: Opaque
--- a/install/installer/pkg/components/server/configmap.go
+++ b/install/installer/pkg/components/server/configmap.go
@@ -91,6 +91,14 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		return nil
 	})
 
+	stripeSecret := ""
+	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
+		if cfg.WebApp != nil && cfg.WebApp.Server != nil {
+			stripeSecret = cfg.WebApp.Server.StripeSecret
+		}
+		return nil
+	})
+
 	disableWsGarbageCollection := false
 	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
 		if cfg.WebApp != nil && cfg.WebApp.Server != nil {
@@ -208,8 +216,9 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		ImageBuilderAddr:             "image-builder-mk3:8080",
 		CodeSync:                     CodeSync{},
 		VSXRegistryUrl:               fmt.Sprintf("https://open-vsx.%s", ctx.Config.Domain), // todo(sje): or "https://{{ .Values.vsxRegistry.host | default "open-vsx.org" }}" if not using OpenVSX proxy
-		EnablePayment:                chargebeeSecret != "",
+		EnablePayment:                chargebeeSecret != "" || stripeSecret != "",
 		ChargebeeProviderOptionsFile: fmt.Sprintf("%s/providerOptions", chargebeeMountPath),
+		StripeSettingsFile:           fmt.Sprintf("%s/settings", stripeMountPath),
 		InsecureNoDomain:             false,
 		PrebuildLimiter: map[string]int{
 			// default limit for all cloneURLs
--- a/install/installer/pkg/components/server/deployment.go
+++ b/install/installer/pkg/components/server/deployment.go
@@ -189,6 +189,29 @@ func deployment(ctx *common.RenderContext) ([]runtime.Object, error) {
 	})
 
 	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
+		if cfg.WebApp != nil && cfg.WebApp.Server != nil && cfg.WebApp.Server.StripeSecret != "" {
+			stripeSecret := cfg.WebApp.Server.StripeSecret
+
+			volumes = append(volumes,
+				corev1.Volume{
+					Name: "stripe-config",
+					VolumeSource: corev1.VolumeSource{
+						Secret: &corev1.SecretVolumeSource{
+							SecretName: stripeSecret,
+						},
+					},
+				})
+
+			volumeMounts = append(volumeMounts, corev1.VolumeMount{
+				Name:      "stripe-config",
+				MountPath: stripeMountPath,
+				ReadOnly:  true,
+			})
+		}
+		return nil
+	})
+
+	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
 		if cfg.WebApp != nil && cfg.WebApp.Server != nil && cfg.WebApp.Server.GithubApp != nil {
 			volumes = append(volumes,
 				corev1.Volume{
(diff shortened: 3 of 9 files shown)


jankeromnes:
Aha, maybe it's because payment is disabled? Let's try:

/werft run with-payment=true

jankeromnes:
Aha, good point! 🎯 Many thanks. Will re-trigger Werft with the config from my branch.

jankeromnes:
Stripe secret successfully mounted! 🎉 

```
$ kubectl describe pod server-5995b76459-llfwg
[...]
Volumes:
[...]
  stripe-config:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  stripe-config
    Optional:    false
```

(Also, Chargebee still works. ✅)

Now testing a build without payment...

jankeromnes:
Without payment, `server` crash-loops with:

```
Unable to load ChargebeeProviderOptions from: /chargebee/providerOptions
```

and

```
Error: ENOENT: no such file or directory, open '/stripe/settings'
```

jankeromnes:
Alright, seems to work fine now! 🎉

## Without payment

https://werft.gitpod-dev.com/job/gitpod-build-jx-stripe-secret.6 Build succeeded ✅ 

| Logging in works ✅ / No payment UI ✅  | Volume is not mounted ✅ |
| --- | --- |
| Screenshot 2022-05-30 at 16 55 30 | Screenshot 2022-05-30 at 16 56 20 |

## With payment

https://werft.gitpod-dev.com/job/gitpod-custom-jx-stripe-secret.3 Build succeeded ✅ 

| Logging in works ✅ / Payment UI is there ✅  | Volume is mounted ✅ |
| --- | --- |
| 
Screenshot 2022-05-30 at 17 06 49 | Screenshot 2022-05-30 at 17 11 19 |

.werft/jobs/build/payment/stripe-config-secret.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+data:
+  settings: eyJwdWJsaXNoYWJsZUtleSI6InBrX3Rlc3RfNTFLeHVyN0dhZFJYbTUwbzNJNXJKQTNvbnkxdGNmdTNkM0NOd3BUWFR6QURkWTJISmlvRk1XTGdTa2M1d2h0UkZRam85UG5kM3pYYUdlcktQcXRmN0REQ3kwMFhBb01kbjZhIiwic2VjcmV0S2V5Ijoic2tfdGVzdF81MUt4dXI3R2FkUlhtNTBvM0NtVFJWc1Q2Q0xqd0VlSlhsWWtmdjZHajREQm42aVlVeDJQWUlUNDhjVlI5dlNUS0s1b2hwQTVCdWdycU5NUU9WVzN0NVJIODAwS011T3lEZ1QifQo=
```
easyCZ, line 3: Should we really have a JWT encoded secret defined here? This is readable by anyone.

install/installer/pkg/components/server/deployment.go
```diff
@@ -188,6 +188,29 @@ func deployment(ctx *common.RenderContext) ([]runtime.Object, error) {
...
+					Name: "stripe-config",
+					VolumeSource: corev1.VolumeSource{
+						Secret: &corev1.SecretVolumeSource{
+							SecretName: stripeSecret,
+						},
+					},
+				})
+
+			volumeMounts = append(volumeMounts, corev1.VolumeMount{
+				Name:      "stripe-config",
```
easyCZ, line 206:
```suggestion
stripeConfigVolumeName = "stripe-config"
// ...
					Name: stripeConfigVolumeName,
					VolumeSource: corev1.VolumeSource{
						Secret: &corev1.SecretVolumeSource{
							SecretName: stripeSecret,
						},
					},
				})

			volumeMounts = append(volumeMounts, corev1.VolumeMount{
				Name:      stripeConfigVolumeName,
```

Because they need to match, they should reference the same variable.

.werft/jobs/build/payment/stripe-config-secret.yaml
jankeromnes: Yes -- this is the "public" testing secret used for PR preview environments.

install/installer/pkg/components/server/configmap.go
```diff
@@ -208,8 +216,9 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		ImageBuilderAddr:             "image-builder-mk3:8080",
 		CodeSync:                     CodeSync{},
 		VSXRegistryUrl:               fmt.Sprintf("https://open-vsx.%s", ctx.Config.Domain), // todo(sje): or "https://{{ .Values.vsxRegistry.host | default "open-vsx.org" }}" if not using OpenVSX proxy
-		EnablePayment:                chargebeeSecret != "",
+		EnablePayment:                chargebeeSecret != "" || stripeSecret != "",
```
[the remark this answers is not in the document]
jankeromnes: I don't think we want to enable anything Stripe-related in Self-Hosted, where payment is disabled.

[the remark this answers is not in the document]
easyCZ: That makes sense. However, currently the stripe config is in experimental mode and only configured for preview. That alone should be enough for it to not be deployed in self-hosted. Am I missing something?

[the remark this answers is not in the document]
jankeromnes: This is actually the other way around, right? I.e. we set `EnablePayment` to true if either the `chargebeeSecret` or the `stripeSecret` is set. 💭

.werft/jobs/build/payment/stripe-config-secret.yaml
easyCZ: Okay. Is this an isolated Stripe **account** which will not be used for production use? My primary concerns are that we should expect someone will use this key to access stripe. By doing so, they can extract test data but also burn through a rate limit. For this use-case, we should have an entirely isolated Stripe account such that there's no way for any of this data to show up while using a production account.

jankeromnes: Good point. There is no production Stripe account yet, but I agree it would make sense to create an entirely separate Stripe account for production when the time comes. Cc @geropl

install/installer/pkg/components/server/configmap.go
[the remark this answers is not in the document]
easyCZ: Okay, that makes more sense. Would it make sense to separate them then? We may want to keep Chargbee enabled, but disable Stripe.

.werft/jobs/build/payment/stripe-config-secret.yaml
easyCZ: @mads-hartmann pointed out that we have a guide for adding secrets to Werft. IMO, it makes sense to follow that here https://www.notion.so/gitpod/Adding-a-secret-to-Werft-652d3cf3bd9d431e94da248901a64fe1.

If we do this, we should also burn the secret committed here.

geropl: @easyCZ  Thx for checking here. The chargebee/Stripe "secrets" are a) for test accounts only, and b) added by intention; this was done to unblock us.

That being said, if we have a nicer way to do things (e.g., the Google Secret Manager), we should try to use it (in a follow-up PR).

install/installer/pkg/components/server/configmap.go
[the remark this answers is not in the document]
geropl: I think the code makes sense as-is. "enablePayment" is a legacy field that we should aim to remove when we kill the chargbee integration. Until we're there, and especially in the context of this PR, it make sense to keep it.

Might make sense to try and remove it from the config surface altogether (and only keep it internall where it make sense) in a separate PR.

geropl:
> However, when payment is disabled (for example in Self-Hosted), I think it would be much nicer to not even try to load Chargebee or Stripe secrets, as opposed to trying + failing + logging (potentially confusing) warnings about payment-related stuff.

@jankeromnes in Self-Hosted neither stripe nor chargebee files are configured, so this is a non-issue. But am ok with the `if (payment && stripeFile)` pattern. :+1:

corneliusludmann (approved):
Approving changes in self-hosted owned file `install/installer/pkg/config/v1/experimental/experimental.go` to unblock this pull request. :skateboard:

geropl (approved):
:+1:

jankeromnes:
Many thanks @geropl @easyCZ and @corneliusludmann for the quick & very helpful reviews! 💯

> @jankeromnes in Self-Hosted neither stripe nor chargebee files are configured, so this is a non-issue. But am ok with the `if (payment && stripeFile)` pattern. 👍

FYI I added the `payment` check because in this Pull Request, when I disabled payment, both Chargebee and Stripe sections failed because of not being configured (Chargebee logged a warning, but Stripe caused a crash).

I assumed this would be the same behavior in Self-Hosted -- if I'm right, when you install Self-Hosted, you'll get a warning about Chargebee not being configured. (Not really a big issue though.)

PR #10343, opened 2022-05-30
[installer] Add flag to configure strict config parsing

easyCZ:
## Description

To make it possible to run config validation in non-strict mode. [Internal context](https://gitpod.slack.com/archives/C032A46PWR0/p1653896011997109).

## Related Issue(s)

Fixes #

## How to test

```bash
$ cd installer/install
$ go run main.go init > config.yaml

# set domain
$ yq w -i config.yaml '.domain = "foo.bar"' 

$ echo "foo: bar" >> config.yaml

$ go run main.go validate config --config ./config.yaml
# expected failure, the current and default behaviour
go run main.go validate config --config ./config.yaml
Error: error loading config: error unmarshaling JSON: while decoding JSON: json: unknown field "foo"

$  go run main.go validate config --config ./config.yaml --strict-parse=false 
{
  "valid": true
}

```

## Release Notes

```release-note
[installer] Command line flag to configure strict config parsing
```

## Documentation

NONE

### DIFF
--- a/install/installer/cmd/root.go
+++ b/install/installer/cmd/root.go
@@ -23,11 +23,13 @@ func Execute() {
 }
 
 var rootOpts struct {
-	VersionMF string
+	VersionMF         string
+	StrictConfigParse bool
 }
 
 func init() {
 	rootCmd.PersistentFlags().StringVar(&rootOpts.VersionMF, "debug-version-file", "", "path to a version manifest - not intended for production use")
+	rootCmd.PersistentFlags().BoolVar(&rootOpts.StrictConfigParse, "strict-parse", true, "toggle strict configuration parsing")
 }
 
 type kubeConfig struct {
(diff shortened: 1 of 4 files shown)


install/installer/cmd/root.go
```diff
@@ -23,11 +23,13 @@ func Execute() {
...
 
 var rootOpts struct {
-	VersionMF string
+	VersionMF         string
+	StrictConfigParse bool
 }
 
 func init() {
 	rootCmd.PersistentFlags().StringVar(&rootOpts.VersionMF, "debug-version-file", "", "path to a version manifest - not intended for production use")
+	rootCmd.PersistentFlags().BoolVar(&rootOpts.StrictConfigParse, "strict-parse", true, "toggle strict configuration parsing")
```
geropl, line 32: nit: being more explicit of what "strict" or "non-strict" means would be nice. E.g. I had to look up [the docs](https://pkg.go.dev/gopkg.in/yaml.v2#UnmarshalStrict) to learn that strict fails on a) unknown and b) _double_ fields.

geropl (approved):
Tested and works. :heavy_check_mark:

corneliusludmann (approved):

PR #10344, opened 2022-05-30
Use slices instead of maps for service ports to control ordering

mads-hartmann:
## Description

This changes the structure from `map[string]ServicePort` to `[]ServicePort` and then extends the `ServicePort` struct with a `Name`. This ensure that the ordering of ports is stable and that we have full control over the ordering.

We need to be able to control the ordering as the order of ports need to match how Kubernetes will order to ports to avoid creating constant "out of sync" applications in ArgoCD ([see internal ticket](https://github.com/gitpod-io/ops/issues/2472) for specific use-case).

This PR doesn't change the ordering yet. Once this is merged and applied in staging I will go over the resulting ArgoCD diff drift and open a new PR that fixes the ordering.

## Related Issue(s)

Part of https://github.com/gitpod-io/ops/issues/2472

## How to test

I used the Gitpod Installer from this branch on a branch in gitpod-io/ops and verified that the ArgoCD diff produced a valid diff (e.g. that I didn't somehow break the manifests generated by the Installer).

## Release Notes

```release-note
NONE
```

## Documentation

N/A

### DIFF
--- a/install/installer/pkg/common/objects.go
+++ b/install/installer/pkg/common/objects.go
@@ -40,17 +40,18 @@ func DefaultServiceAccount(component string) RenderFunc {
 }
 
 type ServicePort struct {
+	Name          string
 	ContainerPort int32
 	ServicePort   int32
 }
 
-func GenerateService(component string, ports map[string]ServicePort, mod ...func(spec *corev1.Service)) RenderFunc {
+func GenerateService(component string, ports []ServicePort, mod ...func(spec *corev1.Service)) RenderFunc {
 	return func(cfg *RenderContext) ([]runtime.Object, error) {
 		var servicePorts []corev1.ServicePort
-		for name, port := range ports {
+		for _, port := range ports {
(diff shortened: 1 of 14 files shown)


mustard-mh (approved):
indeed map has random seq, LGTM

install/installer/pkg/common/objects.go
```diff
@@ -40,17 +40,18 @@ func DefaultServiceAccount(component string) RenderFunc {
 }
 
 type ServicePort struct {
```
easyCZ, line 42: This struct is now very close to the underlying `corev1.ServicePort`. We could consider using that instead as the extra struct isn't adding much anymore.

easyCZ (approved):
LGTM

nandajavarma:
LGTM! :confetti_ball:

nandajavarma (approved):
LGTM! :confetti_ball:

jenting (approved):

PR #10365, opened 2022-05-30
Add installer tests for GKE and k3s

### LINKED ISSUE #7316: Epic: Automated Self-Hosted Testing
## Summary
Set up werft jobs that automatically install and test (running the [existing integration tests](https://github.com/gitpod-io/gitpod/tree/120872bfce3bbb7a7d1b3372282166745d9f9ec2/test)) Gitpod self-hosted from `main` on 4 representative infrastructure environments every night as well as manually on demand. Also the update path should be tested

## Context
Before we can publish a new Gitpod self-hosted release, we need to test the installation on all supported platforms. Doing these by hand is annoying, error-prone, and time-consuming. That's why we need to automate this. This will build our confidence in our releases, which in turn will increase trust with users and customers. 

## Value
Regular, automated tests ensure more reliable Gitpod releases, increasing our own confidence and our customer's confidence.

## Acceptance Criteria
- We have a werft job that tests the latest self-hosted from `main` and runs the existing integration tests.
- Test runs on 4 representative environments (tbc what these are, see below for first draft)
- Tests for fresh installations as well as updates from the previous version/release.
- Notifications about failed tests.

## Measurement
- Number of tests run per week.
- Number of different environments we test on on a nightly basis.

## Growth Area
- Onboarding

## Proposed Test Environments
1. AWS
2. GKE
3. AWS - airgapped
4. AWS - self-signed certs

## Milestones
- [x] Milestone 1 - MVP with a basic werft pipeline to run tests
- [x] Milestone 2 - Support all cloud providers
- [x] Milestone 3 - Additional testing dimensions
- [ ] Milestone 4: Unittests for terraform script

Full testing surface area defined [here](https://www.notion.so/gitpod/Self-Hosted-Testing-Resources-1ad123a1fb944768a1bd8ad55f611d2a#bcec048707ab442daba5fa08ff5c5b90) (internal)

nandajavarma:
## Description

This PR adds the first version of automated installer tests for `GKE` and `k3s`.
> TODO: add proper description

## Related Issue(s)

Fixes #7316

## How to test

To create a `GKE` cluster, setup `Gitpod` and run existing tests on it, run:
```sh
werft run github -f -s .werft/installer-tests.ts -j .werft/gke-installer-tests.yaml -a debug=true
```
Similarly, to run on a `k3s` cluster:
```sh
werft run github -f -s .werft/installer-tests.ts -j .werft/k3s-installer-tests.yaml -a debug=true
```
> Make sure that it is run from a workspace that has access to the `core-dev` cluster since it uses secrets already created there for authentication

## Release Notes

```release-note
```


roboquat:

<details>

</details>

install/tests/manifests/gcp-issuer.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
+metadata:
+  name: gitpod-issuer
+spec:
+  acme:
+    email: nandaja.varma@gmail.com
```
corneliusludmann, line 7: As far as I know, you can simply remove the email line. This is only needed to notify when a cert expires (which is not relevant for test setups).

corneliusludmann (commented):
As far as I can tell, it looks good so far. However, I would like to get a second pair of eyes for this from the platform team. @meysholdt :eyes:

install/tests/manifests/gcp-issuer.yaml
meysholdt: If you need an email address for internal use as a sink for all emails from LetsEncrypt - we already use `letsencrypt-throwaway@gitpod.io` for that. With that being said - it it's possible to not specify an email at all then that is the better solution.

install/infra/terraform/gke/main.tf
```diff
@@ -0,0 +1,191 @@
...
+    ]
+
+    labels = {
+      "gitpod.io/workload_meta" = true
+      "gitpod.io/workload_ide"  = true
+    }
+
+    preemptible  = var.pre-emptible
+    image_type   = "UBUNTU_CONTAINERD"
+    disk_type    = "pd-ssd"
```
meysholdt, line 114: a standrad hdd is chaper is should be sufficient.

install/infra/terraform/gke/main.tf
```diff
@@ -0,0 +1,191 @@
...
+      "https://www.googleapis.com/auth/cloud-platform"
+    ]
+
+    labels = {
+      "gitpod.io/workload_meta" = true
+      "gitpod.io/workload_ide"  = true
+    }
+
+    preemptible  = var.pre-emptible
+    image_type   = "UBUNTU_CONTAINERD"
```
meysholdt, line 113: The COS image is optimised for performance and security. Since we don't run workspaces on this node pool (that are not compatible with COS), I think we should use COS nodes for the service pool.

install/infra/terraform/gke/main.tf
```diff
@@ -0,0 +1,191 @@
...
+  subnetwork = google_compute_subnetwork.subnet.name
+}
+
+resource "google_container_node_pool" "services" {
+  name               = "services-${var.name}"
+  location           = google_container_cluster.gitpod-cluster.location
+  cluster            = google_container_cluster.gitpod-cluster.name
+  version            = var.kubernetes_version // kubernetes version
+  initial_node_count = 1
+  max_pods_per_node  = 110
```
meysholdt, line 100: 110 is a dangerously high number - if something in the autoscaler goes wrong 110 nodes will lead to high cloud costs. 
Proposal Limit to 6 nodes. That's two per region if this is a regional cluster.

install/infra/terraform/gke/variables.tf
```diff
@@ -0,0 +1,81 @@
...
+  default     = "b"
+}
+
+variable "kubernetes_version" {
+  type        = string
+  description = "Kubernetes version to be setup"
+  default     = "1.21.11-gke.1100"
+}
+
+variable "name" {
```
meysholdt, line 30: why make this configurable? Do you wan to support more than one cluster per GCP project?

install/infra/terraform/gke/variables.tf
```diff
@@ -0,0 +1,81 @@
...
+  default     = 1
+}
+
+variable "max_count" {
+  type        = number
+  description = "Maximum number of nodes in the NodePool. Must be >= min_node_count."
+  default     = 50
+}
+
+variable "disk_size_gb" {
```
meysholdt, line 60: Why make this configurable? I think the disk size can be decided statically based on what our services need.

install/infra/terraform/gke/variables.tf
```diff
@@ -0,0 +1,81 @@
...
+  type        = number
+  description = "Size of the node's disk."
+  default     = 100
+}
+
+variable "initial_node_count" {
```
meysholdt, line 66: Why make this configurable?

install/infra/terraform/k3s/variables.auto.tfvars
```diff
@@ -0,0 +1,5 @@
+kubeconfig = "./kubeconfig"
+
+gcp_project = "sh-automated-tests"
+gcp_region  = "europe-west1"
+gcp_zone    = "europe-west1-b"
```
meysholdt, line 5: Is this file helpful? I'm asking because `variables.tf` already has defaults and for the sake of simplicity it would be nice to keep the number if files low.

kylos101:
:wave: this is awesome to see! :pray: Thank you so much.

@corneliusludmann What is the plan to help teams own exclusive testing for their domains of expertise? For example, I do not see workspace sub-folders, or updates to the codeowners file for this PR. Is that something that we'd layer on later, in a follow-on PR?

For example, I see some workspace things now, and if we are to own them later, I would prefer that the workspace team is looped in as an explicit approver so we can have related dialog.

install/infra/terraform/gke/main.tf
```diff
@@ -0,0 +1,191 @@
...
+    ip_cidr_range = "10.0.0.0/12"
+  }
+
+  secondary_ip_range {
+    range_name    = "services-secondary-ip-range"
+    ip_cidr_range = "10.64.0.0/12"
+  }
+}
+
+resource "google_container_cluster" "gitpod-cluster" {
```
adrienthebo, line 41: Noting this for information sharing, no action needed.

This module handles a number of cases that are great for a production cluster (and thus would be great for a reference architecture module), but may be overkill for testing. Features like cluster autoscaling or regional clusters will provide better performance and resilience for production traffic, but for everyday system testing we shouldn't need a lot of nodes, autoscaling, or resilience in the case of a failed GCP zone.

Having a single, unified module for system tests (automated and manual) as well as reference architectures will be fantastic, and this will be great for that - so I don't think we need to pull out any features. However keeping an eye towards the overall cost of test clusters gives us an advantage; if we can drive down the cost of a single test run then we can run tests more cheaply and more frequently.

install/infra/terraform/gke/main.tf
adrienthebo: Agree that this is quite high, and could be expensive. 6 is pretty reasonable.

For testing purposes I don't think we even need the resilience provided by a regional cluster; if we get a failed CI run due to a GCP zone going offline (which would be an extraordinary event) we can move to a different zone/region. If we opt for a zonal cluster and set the default max to 3 nodes, we can potentially halve our spend for each CI run.

install/infra/terraform/gke/variables.tf
adrienthebo: I see the possibility of parallel CI jobs creating parallel clusters; parameterizing the cluster name is cheap to do now but can be a bit painful down the road. I'm in favor of this approach.

(Side note; if we have multiple parallel CI jobs then the terraform state bucket will need to have a [prefix](https://www.terraform.io/language/settings/backends/gcs#prefix) per run, but this is a yak that we can shave another day.)

install/infra/terraform/gke/variables.tf
```diff
@@ -0,0 +1,81 @@
...
+variable "region" {
+  type        = string
+  description = "The region to create the cluster."
+  default     = "europe-west1"
+}
+
+variable "zone" {
+  type        = string
+  description = "The zone to create the cluster."
+  default     = "b"
```
adrienthebo, line 21: I'm not sure this default will behave as expected; my understanding (rooted in the google provider circa 2019) of the google provider block is that the [zone](https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/getting_started) field is the fully qualified zone name, so `europe-west1-b`.

My understanding may be out of date though, so we'll need to check if this works as expected.

install/infra/terraform/gke/variables.tf
```diff
@@ -0,0 +1,81 @@
...
+variable "min_count" {
+  type        = number
+  description = "Minimum number of nodes in the NodePool. Must be >=0 and <= max_node_count."
+  default     = 1
+}
+
+variable "max_count" {
+  type        = number
+  description = "Maximum number of nodes in the NodePool. Must be >= min_node_count."
+  default     = 50
```
adrienthebo, line 57: Small comment: a default max_count of 50 for Gitpod services seems pretty high, especially for tests. If we have a reasonable estimate of the number of nodes spun up during an average test run then we may be able to constrain this lower.

In addition, it may be worthwhile to have separate max_count values for the services and workspaces node pools.

adrienthebo (commented):
Hello!

I'm dropping in to review this to get a feel for how y'all perform code reviews.

I'd like to start by noting that I don't yet know the culture or review practices for Gitpod; sincere apologies if I'm stepping on toes, dropping too many comments, or focusing on topics that might be out of scope for this PR. In addition I'm approaching this PR from the perspective of my role as a GCP consultant, which is an outside perspective driven by GCP best practices but isn't yet informed about Gitpod's standard GCP operations.  I hope to share my experience in this domain and offer useful insights in constructive and effective ways, and getting this down will take a few review cycles. Thanks for bearing with me!

My overall perspective is that this is a rock solid foundation for system testing the Gitpod installer for self hosting purposes. I've been talking with @mrzarquon regarding reference architectures and automation for self hosted customers and this is the type of module that could be the basis for that reference, which is really awesome.

In terms of correctness, I think I saw one place (the `zone` variable) that needed a tweak to behave as expected but it might be using a feature (the full zone name being derived based on `region + zone`) that I'm not familiar with. 👍 

Summarizing some of my inline comments, there's some functionality in this module that's well suited for a reference architecture but may benefit from adjustments for running in a CI system. If our installer systems tests are cheap and fast to run then we can run them more frequently, and lightweight test environments will allow upstream tests to validate changes impacting the installation process earlier.

All in all, I think this is really good and look forward to taking this for a spin!

install/infra/terraform/gke/main.tf
meysholdt: > I misread this, this is pods per node. I'm fine with having a high number of pods per node. 

me too :(  110 pods per node is perfectly fine and has zero impact on cost. Apologies for the confusion.

nandajavarma:
> wave this is awesome to see! pray Thank you so much.
> 
> @corneliusludmann What is the plan to help teams own exclusive testing for their domains of expertise? For example, I do not see workspace sub-folders, or updates to the codeowners file for this PR. Is that something that we'd layer on later, in a follow-on PR?
> 
> For example, I see some workspace things now, and if we are to own them later, I would prefer that the workspace team is looped in as an explicit approver so we can have related dialog.

@kylos101 Yes, we need more iterations to add component specific tests. In general, I have tried to move the current guides that we have for self-hosted setup on `gke` and `k3s` to `IaC`. As of the tests, currently I am running the tests the integrations tests run(the ones inside the `tests/` directory). Since the coverage is low, we have to do more iterations to add more component specific tests. The hope is that it will be easily pluggable with the current structure. We will empower the individual teams to add tests for their domains of expertise.

kylos101:
@nandajavarma that is super helpful to know, thank you! So this PR is more about provisioning infrastructure and installing Gitpod, and then running [our existing tests](https://github.com/gitpod-io/gitpod/tree/main/test)?

Is it safe to say that we'll see future iterations of this, that use different reference architectures to provision infrastructure and install Gitpod?

nandajavarma:
> @nandajavarma that is super helpful to know, thank you! So this PR is more about provisioning infrastructure and installing Gitpod, and then running [our existing tests](https://github.com/gitpod-io/gitpod/tree/main/test)?
> 
> Is it safe to say that we'll see future iterations of this, that use different reference architectures to provision infrastructure and install Gitpod?

@kylos101 yes, correct! We will add automated setup of different reference architectures(for eg: the WIP aks one [here](https://github.com/gitpod-io/gitpod/pull/10415)) including all available configurations(eg: self-signed certs or otherwise) to test on in following iterations.

install/infra/terraform/gke/variables.tf
nandajavarma: Like @adrienthebo  suggested, the main reason for considering configurable name parameter is parallel CI runs(it is highly likely since we are planning to test multiple configs). I was hoping that having the `default` name reduces the hassle a bit.
Regarding the state @adrienthebo , this is a great point. The way I have worked around it currently, is by using [terraform workspaces](https://www.terraform.io/language/state/workspaces) in the [Makefile](https://github.com/gitpod-io/gitpod/pull/10365/files/5832150fbf31b334b333493f86a2ad5b7ef0dcd1#diff-bd7b7f3a397df93111a6b6e2a4284b01bd57e04474bc9f91f1b02ee1960399a2R28) since I couldn't set the the `prefix` using variables. Each job is run in its own tf workspace and this means the state file will have the name `<workspace-name>.tfstate`.

install/tests/manifests/gcp-issuer.yaml
nandajavarma: turns out, I could just leave out the `email` key!

install/infra/terraform/gke/variables.tf
nandajavarma: This is true @adrienthebo. Like you can see [here](https://github.com/gitpod-io/gitpod/pull/10365/files/5832150fbf31b334b333493f86a2ad5b7ef0dcd1#diff-6fb78b207bf98868abc50e2625d643de068cd54c1de8eda56af56bc4a98d3dcbR43), I am appending this value(eg. `b`) to the `var.region` (eg. `europe-west1`). In retrospect, I understand this is particularly hacky. Will fix. :+1:

install/infra/terraform/gke/main.tf
```diff
@@ -0,0 +1,191 @@
...
+      resource_type = "memory"
+      minimum       = 4
+      maximum       = 64
+    }
+  }
+
+  # We can't create a cluster with no node pool defined, but we want to only use
+  # separately managed node pools. So we create the smallest possible default
+  # node pool and immediately delete it.
+  remove_default_node_pool = true
```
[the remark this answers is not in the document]
nandajavarma: This is a great point. The solution to re-use the default node as `service` node is something I thought too. But I couldn't see an option at least in the [container cluster terraform resource](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster) to set labels only to the default node(the labels being `meta` and `ide`). If I set the `label` field, it would apply the same label to every nodepool. For the testing purposes, I imagine this is probably okay. Not sure if this is the best course of action.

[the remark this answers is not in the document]
nandajavarma: Turns out, I couldn't have been more wrong. That option only applied the labels to the default node. Making this change! Thanks for the suggestion, Mo!

install/infra/terraform/gke/variables.tf
nandajavarma: The assumption was that, if the user decides to use more in-cluster storage, it would mean that they need more storage. So keeping it configurable made sense at the time.

install/infra/terraform/gke/variables.tf
nandajavarma: Out currently suggested `max_count` value in the guides are `100`, I wanted to bring it a bit down if we are using default but keeping it configurable. We can configure this value in the [`main.tf` used for tests](https://github.com/gitpod-io/gitpod/pull/10365/files/5832150fbf31b334b333493f86a2ad5b7ef0dcd1#diff-a39dd45b31fd7a552e7b9786c9bcc9939453bbbbd20dc31835760bc78a56f4eb). I believe it is safe to assume that `install/infra/terraform/` will have more production ready code and `install/tests` has test code.

install/infra/terraform/gke/variables.auto.tfvars
```diff
@@ -0,0 +1,5 @@
+kubeconfig = "./kubeconfig"
+
+project = "sh-automated-tests"
+region = "europe-west1"
+zone = "b"
```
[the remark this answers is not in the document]
nandajavarma: Good point. Removed.

install/infra/terraform/gke/main.tf
```diff
@@ -0,0 +1,191 @@
...
+
+  autoscaling {
+    min_node_count = var.min_count
+    max_node_count = var.max_count
+  }
+
+
+  management {
+    auto_repair  = true
+    auto_upgrade = false
```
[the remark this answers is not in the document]
kylos101: I am pretty sure prebuilds are not fault tolerant yet. If they were, we'd be running them on spot VMs now.

PR #10367, opened 2022-05-31
[usage] Add db.WorkspaceInstance model in golang

easyCZ:
## Description

Same as https://github.com/gitpod-io/gitpod/pull/10293 but for WorkspaceInstances.

## Related Issue(s)

* Part of https://github.com/gitpod-io/gitpod/issues/10317
* Depends on https://github.com/gitpod-io/gitpod/pull/10390

## How to test

Unit tests

## Release Notes

```release-note
NONE
```

## Documentation

NONE

### DIFF
--- a/components/usage/pkg/db/workspace_instance.go
+++ b/components/usage/pkg/db/workspace_instance.go
@@ -1,0 +1,44 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package db
+
+import (
+	"database/sql"
+	"github.com/google/uuid"
+	"gorm.io/datatypes"
+	"time"
+)
+
+type WorkspaceInstance struct {
+	ID                 uuid.UUID      `gorm:"primary_key;column:id;type:char;size:36;" json:"id"`
+	WorkspaceID        string         `gorm:"column:workspaceId;type:char;size:36;" json:"workspaceId"`
+	Configuration      datatypes.JSON `gorm:"column:configuration;type:text;size:65535;" json:"configuration"`
+	Region             string         `gorm:"column:region;type:varchar;size:255;" json:"region"`
+	ImageBuildInfo     sql.NullString `gorm:"column:imageBuildInfo;type:text;size:65535;" json:"imageBuildInfo"`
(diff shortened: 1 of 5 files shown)


roboquat:

<details>

</details>

components/usage/pkg/db/workspace_instance.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"database/sql"
+	"github.com/google/uuid"
+	"gorm.io/datatypes"
+	"time"
+)
+
+type WorkspaceInstance struct {
+	ID                 uuid.UUID      `gorm:"primary_key;column:id;type:char;size:36;" json:"id"`
+	WorkspaceID        string         `gorm:"column:workspaceId;type:char;size:36;" json:"workspaceId"`
```
geropl, line 16: nit: one ID column is UUID, the other string: Why not make both strings? (so far we only have one column type for IDs: string) What's the advantange of UUID?

geropl (approved):
code looks good, tests are green! :heavy_check_mark: 

/hold because of [this question](https://github.com/gitpod-io/gitpod/pull/10367/files#r886544711). I leave it up to you to handle. :pray:

components/usage/pkg/db/workspace_instance.go
easyCZ: UUID allows you to parse UUIDs and validate before you even hit the DB. UUID is the more appropriate type as it's more specific. For example, it prevents you from inserting something like `Workspace{ID: "my-random-string"}` because it won't even compile

easyCZ: And it is. `uuid.UUID` serializes as a string. It's the same as using the UUID_COLUMN type in our TypeORM implementation. 

https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-db/src/typeorm/entity/db-workspace-instance.ts#L24

geropl: Ah! The UUID_COLUMN_TYPE is only there to ensure the width/type of the field is the same everywhere.

The key question is: does `uuid.UUID` influence the runtime behavior (e.g., introduces new error modes during parsing? If no I'm happy! :+1:

easyCZ: It shouldn't, as long as what is stored in the DB is a valid UUID. I'm gonna run it against the RO replica of staging & prod to batch list all items to validate we don't have underlying violations to verify.

PR #10372, opened 2022-05-31
content-service: ignore the error of `user.overlay.(impure|origin)` attributes

### LINKED ISSUE #10108: Permission denied when accessing .docker-root/overlay2
### Bug description

I see a lot (40k in last 3 hours) errors like this in SaaS:
```
jsonPayload: {
@type: "type.googleapis.com/google.devtools.clouderrorreporting.v1beta1.ReportedErrorEvent"
error: "permission denied"
gid: 33333
level: "warning"
message: "cannot chown"
path: ".docker-root/overlay2/205df79f676c4bbc3e8af627dc2a78228e55f494d7b21c5725fcaeb808e285cb/diff/home/node/parse-server/"
uid: 33333
```
Link to logs: https://cloudlogging.app.goo.gl/mZg3N1J2w58wm7fLA

This might be related to an issue customer is having as well: 
https://app.frontapp.com/inboxes/teammates/9932874/inbox/open/41825492426

Potentially related issue: https://github.com/gitpod-io/gitpod/issues/9263

Would like to understand why we are getting those errors and what is the impact.

### Steps to reproduce

One potential repro case:
This seems to do with running `docker-compose pull `and `docker-compose build` in init task.

### Workspace affected

tracthq-tractweb-xtjbxw26gqh

### Expected behavior

We shouldn't be getting permission denied errors when accessing .docker-root I think.

### Example repository

_No response_

### Anything else?

_No response_

utam0k:
## Description

`user.overlay.(impure|origin)` attributes is a marker to match inodes for overlayfs, such as when an upper layer copies a lower layer file in overlayfs.
However, when restoring content, the container in the workspace is not always running, so there is no problem ignoring the failure.
https://github.com/hisilicon/overlayfs-progs/blob/e10ef686570d9c7eff42f52461593a5c15da56bd/README#L62

## Related Issue(s)

Fixes #10108

## How to test

No errors including logs with the following steps

1. Open https://to-impure-dir.preview.gitpod-dev.com/#https://github.com/utam0k/gitpod-playground/tree/prebuild-docker
2. Stop the workspace when opening the workspace is completely
3. Reopen the workspace

## Release Notes

```release-note
content-service: Ignore errors in attributes settings
```

## Documentation

No

### DIFF
--- a/components/content-service/pkg/archive/tar.go
+++ b/components/content-service/pkg/archive/tar.go
@@ -198,10 +198,7 @@ func remapFile(name string, uid, gid int, xattrs map[string]string) error {
 			if err == syscall.ENOTSUP || err == syscall.EPERM {
 				continue
 			}
-
-			log.WithField("name", key).WithField("value", value).WithField("file", name).WithError(err).Error("restoring extended attributes")
-
-			return err
+			log.WithField("name", key).WithField("value", value).WithField("file", name).WithError(err).Warn("restoring extended attributes")
 		}
 	}
 
--- a/components/content-service/pkg/archive/tar.go
+++ b/components/content-service/pkg/archive/tar.go
@@ -194,10 +194,19 @@ func remapFile(name string, uid, gid int, xattrs map[string]string) error {
 			continue
 		}
 
+		if strings.HasPrefix(key, "user.") {
+			// This is a marker to match inodes, such as when an upper layer copies a lower layer file in overlayfs.
+			// However, when restoring a content, the container in the workspace is not always running, so there is no problem ignoring the failure.
+			if strings.HasSuffix(key, ".overlay.impure") || strings.HasSuffix(key, ".overlay.origin") {
+				continue
+			}
+		}
+
 		if err := unix.Lsetxattr(name, key, []byte(value), 0); err != nil {
 			if err == syscall.ENOTSUP || err == syscall.EPERM {
 				continue
 			}
+
 			log.WithField("name", key).WithField("value", value).WithField("file", name).WithError(err).Warn("restoring extended attributes")
 		}
 	}


utam0k:
> @utam0k can we confirm dockerd/containerd is not running when we create the backup?

This issue was also reproduced by simply pulling the image.

utam0k:
@aledbf I had a suspicion that the container process was running, but is there a problem with the daemon itself?

utam0k:
@aledbf Sorry, but what did `temporal files` point out?
This is the error when I pull with docker and docker-compose with this repository and restart
https://github.com/utam0k/gitpod-playground/tree/prebuild-docker-compose
https://cloudlogging.app.goo.gl/3HRd2Pi9PEtTfW6d8

aledbf:
@utam0k what about 


```
for key, value := range xattrs {
	// do not set trusted attributes
	if strings.HasPrefix(key, "trusted.") {
		continue
	}

	// This is a marker to match inodes, such as when an upper layer copies a lower layer file in overlayfs.
	// However, when restoring a content, the container in the workspace is not always running, so there is no problem ignoring the failure.
	if strings.HasSuffix(key, ".overlay.impure") || strings.HasSuffix(key, ".overlay.origin") {
		continue
	}

	if err := unix.Lsetxattr(name, key, []byte(value), 0); err != nil {
		if err == syscall.ENOTSUP || err == syscall.EPERM {
			continue
		}

		log.WithField("name", key).WithField("value", value).WithField("file", name).WithError(err).Warn("restoring extended attributes")
		return err
	}
}
```

utam0k:
@aledbf Thanks for your good review! PTAL

Furisto (approved):

PR #10376, opened 2022-05-31
[ws-manager] Add workspace class to metrics

csweichel:
## Description
This PR adds the workspace class to some of the ws-manager metrics. This way, we can see how many workspaces of a given class are currently running.

image

## How to test
- start a workspace
- observe the `gitpod_ws_manager_workspace_phase_total` metric and ensure the `class` label is present

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/ws-manager/pkg/manager/monitor.go
+++ b/components/ws-manager/pkg/manager/monitor.go
@@ -791,8 +791,9 @@ func (m *Monitor) initializeWorkspaceContent(ctx context.Context, pod *corev1.Po
 	} else {
 		err = handleGRPCError(ctx, err)
 	}
-	wsType := pod.Labels[wsk8s.TypeLabel]
-	hist, errHist := m.manager.metrics.initializeTimeHistVec.GetMetricWithLabelValues(wsType)
+	wsType := strings.ToUpper(pod.Labels[wsk8s.TypeLabel])
+	wsClass := pod.Labels[workspaceClassLabel]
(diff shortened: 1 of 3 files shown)


csweichel:
> Why `type` is uppercase and lowercase?

no good reason - will fix

components/ws-manager/pkg/manager/monitor.go
```diff
@@ -791,8 +791,9 @@ func (m *Monitor) initializeWorkspaceContent(ctx context.Context, pod *corev1.Po
 	} else {
 		err = handleGRPCError(ctx, err)
 	}
-	wsType := pod.Labels[wsk8s.TypeLabel]
-	hist, errHist := m.manager.metrics.initializeTimeHistVec.GetMetricWithLabelValues(wsType)
+	wsType := strings.ToUpper(pod.Labels[wsk8s.TypeLabel])
```
jenting, line 794: Why convert to upper case?

csweichel: to make it consistent with the other metrics: https://github.com/gitpod-io/gitpod/pull/10376#issuecomment-1141888616

It's easier to make that one uppercase considering that the rest is, too

Furisto (approved):

components/ws-manager/pkg/manager/monitor.go
jenting: I see, I was confused by the screenshot in this PR description, probably missed to update it.

Issue #10326, opened 2022-05-29
Use Stripe Elements to capture CreditCard information then create a Stripe Customer and Subscription


easyCZ:
In order to be able to charge a customer, we need to first accept their Credit Card information and have a corresponding Stripe Subscription for this customer.

Use Stripe elements to setup the form for accepting credit cards. Ensure that no credit-card data information lands on Gitpod systems and the complete flow is done through Stripe.

PR #10378, opened 2022-05-31
Allow teams to sign up for Usage-Based Billing via Stripe

jankeromnes:
## Description

Allow teams to sign up for Usage-Based Billing via Stripe.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/10326

## How to test

1. Join the "Gitpod" team: [invite link](https://jx-stripe-service.preview.gitpod-dev.com/teams/join?inviteId=9d9cd3f1-bad9-4725-8e06-469ce0d955c1) (this enables the Usage-Based feature flag for you)
2. Create a new, different team ✨ 
3. Navigate to your Team Settings > Team Billing
4. Wait for the Usage-Based UI to appear above the Team Plans
5. Upgrade your team to usage-based by providing a fake credit card (e.g. 4242 4242 4242 4242, 4/24, 424)
6. This should successfully create a Stripe Customer for your team, visible in our [Stripe Dashboard](https://dashboard.stripe.com/test/customers) (credentials in 1Password)

| No Team Billing | Add Credit Card | Active Team Billing | Stripe Dashboard |
| --- | --- | --- | --- |
| Screenshot 2022-06-01 at 16 54 30 | Screenshot 2022-06-01 at 16 55 02 | Screenshot 2022-06-01 at 16 58 35 | Screenshot 2022-06-01 at 16 56 42 |

## Release Notes

```release-note
NONE
```

## Documentation


components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -1827,6 +1829,32 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
         return subscription;
     }
 
+    async getStripePublishableKey(ctx: TraceContext): Promise<string | undefined> {
+        this.checkAndBlockUser("getStripePublishableKey");
+        return this.config.stripeSettings?.publishableKey;
+    }
+
+    async getStripeSetupIntentClientSecret(ctx: TraceContext): Promise<string | undefined> {
```
easyCZ, line 1837: When interactions with 3rd party services fail, we shouldn't be propagating the errors directly to the client. We don't entirely control what is in the errors and may contain sensitive data. How is that handled here? I'd expect that we check for success/error and return appropriate obfuscated error messages upstream to the client.

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -1827,6 +1829,32 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
         return subscription;
     }
 
+    async getStripePublishableKey(ctx: TraceContext): Promise<string | undefined> {
```
easyCZ, line 1832: This method should also be behind a feature flag. The feature flag on the client is useless if we do not check it on the server side too.

components/server/ee/src/workspace/gitpod-server-impl.ts
jankeromnes: Agree -- we could avoid propagating to the user/client with a try/catch:

```ts
// Pseudo-code
try {
    await stripeThing();
} catch (error) {
    log.error("the full error for our logs", { error });
    throw new Error("Stripe thing failed for some reason ¯\_(ツ)_/¯");
}
```

components/server/ee/src/workspace/gitpod-server-impl.ts
jankeromnes: Thanks, but in this case I would disagree:
- Gating this behind the Usage-Based Feature flag would be over-engineered here (requires making yet another call to configcat from the server, and wait for the response which can take a while)
- We primarily want to hide Usage-Based Billing from users because it's work-in-progress, but if someone is dedicated enough to call the server API directly, I guess they're allowed to try out our Stripe integration (which we're planning to GA in a few weeks anyway)

easyCZ:
Behaviour wise: works well and tested.

The popup to add credit-card logs a number of errors when making a `POST https://play.google.com/log?format=json&hasfast=true&authuser=0 net::ERR_BLOCKED_BY_CLIENT`. Can this be removed? Why does it happen?

If you wanted to, this PR could've been split into the following:
1. UI to show upgrade button which does nothing
2. Server-side APIs (behind a feature flag) which return Unimplemented when part of team gitpod, NotAllowed otherwise.
3. Server-side APIs actually implement behaviour
4. UIs which hook up the behaviour.

That way, you'd be able to keep the PRs smaller, and individual steps smaller. For example, in this case you'd get the review feedback on needing the feature flag guard server-side on the API setup, not on the full PR.

components/server/ee/src/workspace/gitpod-server-impl.ts
easyCZ: > We primarily want to hide Usage-Based Billing from users because it's work-in-progress, but if someone is dedicated enough to call the server API directly, I guess they're allowed to try out our Stripe integration (which we're planning to GA in a few weeks anyway)

The client must not be authoritative. We are not just hiding UI features, we're gatting functinality to control how we roll it out. We don't want anyone without the correct permissions (or membership) to access this feature - this cannot be done on the UI and must be done on the server side.

> making yet another call to configcat from the server, and wait for the response which can take a while)

Feature flags are updated async, actual eval is non-blocking.

geropl: @jankeromnes I agree with @easyCZ : Deciding whether or not to enable a feature is a authorative (server-side) question. IMO it's the only way to serve this as "clean cut".
And whether or not we do another request against configcat should not guide any decision I think. :+1:

jankeromnes: Done ✅

components/dashboard/src/teams/TeamUsageBasedBilling.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+    if (!showUsageBasedUI) {
+        return <></>;
+    }
+
+    return (
+        <div className="mb-16">
+            <h3>Usage-Based Billing</h3>
+            <h2 className="text-gray-500">Manage usage-based billing, spending limit, and payment method.</h2>
+            <div className="max-w-2xl">
```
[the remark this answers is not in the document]
jankeromnes: Thanks! Fix implemented. ✅

components/server/ee/src/workspace/gitpod-server-impl.ts
easyCZ: Should also be added for the other API methods that are for usage based billing.

jankeromnes:
@gtsiolis Had a quick look into how we can style Stripe's PaymentElement.

[Here](https://stripe.com/docs/elements/appearance-api) is how they suggest we do it:

> 1. Start by picking a [theme](https://stripe.com/docs/elements/appearance-api#theme)

The 4 themes are:

| stripe | night | flat | none |
| --- | --- | --- | --- |
| Screenshot 2022-06-07 at 15 10 36 | Screenshot 2022-06-07 at 15 07 12 | Screenshot 2022-06-07 at 15 11 00 | Screenshot 2022-06-07 at 15 11 34 |

I think a good Minimum-Viable-Change could be:
- Use `stripe` or `flat` in Light theme
- Use `night` in Dark theme

> 2. Customize the theme using [variables](https://stripe.com/docs/elements/appearance-api#variables)
> 
> Set variables like `fontFamily` and `colorPrimary` to broadly customize components appearing throughout each Element.

We could do this to further align the look & feel of Stripe Elements with our dashboard in broad strokes. 💡

> 3. If needed, fine-tune individual components and states using [rules](https://stripe.com/docs/elements/appearance-api#rules)

With this, we can even drill down into specific styling issues that might remain after 1. and 2. 🔍

components/dashboard/src/teams/TeamUsageBasedBilling.tsx
```diff
@@ -0,0 +1,240 @@
...
+    if (!showUsageBasedUI) {
+        return <></>;
+    }
+
+    return (
+        <div className="mb-16">
+            <h3>Usage-Based Billing</h3>
+            <h2 className="text-gray-500">Manage usage-based billing, spending limit, and payment method.</h2>
+            <div className="max-w-xl">
+                <div className="mt-4 h-32 p-4 pt-6 pl-6 flex flex-col rounded-xl bg-gray-100 dark:bg-gray-800">
```
gtsiolis, line 125: **nitpick**: Minor spacing fix to match the design specs.

```suggestion
                <div className="mt-4 h-32 p-4 pt-4 pl-4 flex flex-col rounded-xl bg-gray-100 dark:bg-gray-800">
```

components/dashboard/src/teams/TeamUsageBasedBilling.tsx
```diff
@@ -0,0 +1,240 @@
...
+        const { server } = getGitpodService();
+        Promise.all([
+            server.getStripePublishableKey().then((v) => () => setStripePromise(loadStripe(v || ""))),
+            server.getStripeSetupIntentClientSecret().then((v) => () => setStripeSetupIntentClientSecret(v)),
+        ]).then((setters) => setters.forEach((s) => s()));
+    }, []);
+
+    const getStripeAppearance = (): Appearance => {
+        return {
+            theme: isDark ? "night" : "flat",
```
gtsiolis, line 174: **issue**: Although flat looks slightly better, should we default to the `stripe` theme so that UX looks similar with the rest of the product? 

| Variable Modal | Billing Modal |
|-|-|
| Screenshot 2022-06-07 at 5 39 49 PM (2) | Screenshot 2022-06-07 at 5 39 40 PM (2) |

gtsiolis (commented):
@jankeromnes I left some minor and non-blocking comments regarding UX. 🏓 

Regarding the comments in https://github.com/gitpod-io/gitpod/pull/10378#issuecomment-1148673742:

1. Going with stripe's components sounds fine. Could we try updating the element styles to match our own? Do you prefer to skip it for now?
2. Did you figure out whether we can drop the country selection in the billing modal?

Also, thanks for taking a closer look on how we can customize Stripe's elements and including all the helpful references above in https://github.com/gitpod-io/gitpod/pull/10378#issuecomment-1148673742. 🙏

components/dashboard/src/settings/Billing.tsx
```diff
@@ -18,7 +19,15 @@ export default function Billing() {
             title="Billing"
             subtitle="Usage-Based Billing."
         >
-            <h3>Billing</h3>
+            <h3>Usage-Based Billing</h3>
+            <h2 className="text-gray-500">Manage usage-based billing, spending limit, and payment method.</h2>
+            <p className="mt-8">
+                Hint:{" "}
```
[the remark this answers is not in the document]
jankeromnes: Thought that could be helpful to Gitpodders with the feature-flag enabled. 💭

(Also the weird code artifacts are mandated by our auto-formatter 🙄.)

jankeromnes:
> @jankeromnes I left some minor and non-blocking comments regarding UX. 🏓

Many thanks @gtsiolis! ❤️ Happy to opportunistically get them in when possible.

> 1. Going with stripe's components sounds fine. Could we try updating the element styles to match our own? Do you prefer to skip it for now?

Cool! And yes we could try that, but in a follow-up Pull Request. I think the initial Light/Dark switch I implemented is good enough for now, and we can easily extend it later with better custom styles/rules. 🎨

> 2. Did you figure out whether we can drop the country selection in the billing modal?

Yes, I believe it's documented somewhere here: https://stripe.com/docs/js/elements_object/create_payment_element

But, I'd love to leave this one as a follow-up too.

components/dashboard/src/teams/TeamUsageBasedBilling.tsx
gtsiolis: Ah, you're right. I knew I should be more careful with sharing sandbox design files and not final design specs that include only designs for one issue. Sorry for the confusion, @jankeromnes!

You probably checked out a previous iteration[[1](https://www.figma.com/file/8UO0eql1UPmI1B0zgf5Kcu/Plans-(Sandbox)?node-id=1021%3A3469)], not the latest in progress below[[2](https://www.figma.com/file/8UO0eql1UPmI1B0zgf5Kcu/Plans-(Sandbox)?node-id=1030%3A3958)].

| Early iteration | Last iteration |
|-|-|
| Team Plans | Team Plans-1 |

jldec:
Works for me as well - Is there a way to navigate manually to the Stripe portal to manage billing/payment details (as a Gitpod user - not as the GitPod Stripe admin) ?

jldec:
@easyCZ do you know why the feature flag feels slow ?

jankeromnes:
> Works for me as well

Thanks! I'm glad this works perfectly for everyone. 😊

> Is there a way to navigate manually to the Stripe portal to manage billing/payment details (as a Gitpod user - not as the GitPod Stripe admin) ?

I don't think so, but I'm planning to implement the Stripe portal in the next PR once this is merged.

easyCZ:
> @easyCZ do you know why the feature flag feels slow ?

To me it looks like react is not being told the values have changed so it doesn't propagate as soon as it should. I'm no expert on react either so not entirely sure where this delay lies.
The request for feature flags is made early and completes fast so it should be available at that point for rendering. I'll try to debug a bit when I have time but others more versed in React can probs debug this faster :)

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -1827,6 +1830,54 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
+        const user = this.checkAndBlockUser("getStripePublishableKey");
+        await this.ensureIsUsageBasedFeatureFlagEnabled(user);
+        return this.config.stripeSettings?.publishableKey;
+    }
+
+    async getStripeSetupIntentClientSecret(ctx: TraceContext): Promise<string | undefined> {
+        const user = this.checkAndBlockUser("getStripeSetupIntentClientSecret");
+        await this.ensureIsUsageBasedFeatureFlagEnabled(user);
+        const setupIntent = await this.stripeService.createSetupIntent();
```
easyCZ, line 1860: The service implementation doesn't catch errors, and neither does this handler. This means that 3rd party errors (from a library) would be propagated to the client. Could you please throw non-detailed error to the client and log the underlying error such that we don't accidentally leak details?

The same also goes for `getTeamStripeCustomerId` and `subscribeTeamToStripe`

Related thread: https://github.com/gitpod-io/gitpod/pull/10378/files#r887842876

easyCZ (changes requested):

geropl:
> If you refresh the dashboard, the Usage-Based feature flag takes a while to be re-enabled (so, if you reload, the Usage-Based UI will be hidden for a few seconds, but otherwise still work well)

This is not the case for me, works rather snappy now. :+1:

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -1827,6 +1830,54 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
         return subscription;
     }
 
+    protected async ensureIsUsageBasedFeatureFlagEnabled(user: User): Promise<void> {
+        const teams = await this.teamDB.findTeamsByUser(user.id);
+        const isUsageBasedBillingEnabled = await getExperimentsClient().getValueAsync(
```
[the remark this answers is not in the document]
jankeromnes: If feature flags don't work as expected in the `server`, I'd vote for not using them right now (a client-side feature flag is enough), and open a separate issue about fixing the server-side feature flags so that we can use them.

components/server/ee/src/workspace/gitpod-server-impl.ts
jankeromnes: I left this as is, because forwarding errors makes debugging a bit easier while this is still behind a feature flag (and we have the same problem in many other server API methods, so a more holistic fix would be a good idea).

However, it's pretty easy to add a try/catch to the Stripe methods I've added, so I'll do that now, hoping that we can then merge this and make progress.

components/server/ee/src/workspace/gitpod-server-impl.ts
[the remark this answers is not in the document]
easyCZ: @jankeromnes How about adding a pre-pr to make it work on server properly and then using it here? I'm happy to pair if that makes it easier. Boy/girl scout rules apply.

geropl (approved):
LGTM, tested and works.

I leave it up to you @jankeromnes to address the ["error leak"](https://github.com/gitpod-io/gitpod/pull/10378/files#r891990649) in a follow-up.

jankeromnes:
Adding a hold for the try/catches:

/hold

components/server/ee/src/workspace/gitpod-server-impl.ts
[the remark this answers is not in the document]
jankeromnes: @easyCZ That's the dashboard implementation, right?

The server-side implementation seems to be here: https://github.com/gitpod-io/gitpod/blob/main/components/server/src/experiments.ts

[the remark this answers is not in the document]
easyCZ: Interesting, wasn't aware this landed (and my IDE resolved the method to the dashboard one). This must be stray from another PR. It also lives under the root of server/src which it should not.

This will in fact work for server. However, there's a second problem with it as it hard-codes a single Client ID which is for the preview/staging environment, but not production.

Let's land as is, and fix it up in a follow-up PR. We'll need to move the file, but also inject the Client ID based on environment we're running in. The current version would also break self-hosted environments as it reaches out to an external source.

easyCZ:
@jankeromnes Let me know when API layer error handling is in and I'll re-review.

components/server/ee/src/workspace/gitpod-server-impl.ts
jankeromnes: Resolved in [[server] Sanitize Stripe error messages sent by server](https://github.com/gitpod-io/gitpod/pull/10378/commits/d31c9f0af53fdd4f088fd028f3506b1905be616b)

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -1857,25 +1857,40 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
         const user = this.checkAndBlockUser("getStripeSetupIntentClientSecret");
         await this.ensureIsUsageBasedFeatureFlagEnabled(user);
-        const setupIntent = await this.stripeService.createSetupIntent();
-        return setupIntent.client_secret || undefined;
+        try {
+            const setupIntent = await this.stripeService.createSetupIntent();
+            return setupIntent.client_secret || undefined;
+        } catch (error) {
+            log.error("Failed to create Stripe SetupIntent", error);
+            throw new ResponseError(500, "Failed to create Stripe SetupIntent");
```
easyCZ, line 1865: Odd we don't have a 500 status code in our list of `errors.ts`

easyCZ (approved):

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -1827,6 +1830,69 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
+        }
+    }
+
+    async getTeamStripeCustomerId(ctx: TraceContext, teamId: string): Promise<string | undefined> {
+        const user = this.checkAndBlockUser("getTeamStripeCustomerId");
+        await this.ensureIsUsageBasedFeatureFlagEnabled(user);
+        await this.guardTeamOperation(teamId, "update");
+        try {
+            const customer = await this.stripeService.findCustomerByTeamId(teamId);
+            return customer?.id || undefined;
```
easyCZ, line 1875: Also, IMO you should return a 404 error when customer.id does not exist. Returning undefined prevents you classifying your errors client side and actually having reasonable metrics on what worked and didn't. This goes for `setupIntent` also.

components/server/ee/src/workspace/gitpod-server-impl.ts
jankeromnes: Agree. We hard-code 500 in various places. Should probably fix that 👍 (in a follow-up 😅)

components/server/ee/src/workspace/gitpod-server-impl.ts
jankeromnes: Hmm, interesting point. I agree on the 404 semantics. However, returning `undefined` currently means that the customer doesn't exist yet, and is handled that way in the client (different from "the Stripe API failed"). If we switch this around to a 404, we'd need to update the client as well to handle 404 as a valid case (i.e. "the team is not yet a Stripe customer").

easyCZ: I think that's correct. If it doesn't exist, then it wasn't found and we should create one (if needed). In general, we should try to avoid OK responses with No data as much as possible, especially when we know why there is no data - in this case it doesn't exist in stripe.

easyCZ: (Follow-up PR/issue is fine)

components/server/ee/src/workspace/gitpod-server-impl.ts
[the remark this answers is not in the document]
jankeromnes: Follow-up issue: https://github.com/gitpod-io/gitpod/issues/10524

PR #10381, opened 2022-05-31
Change threshold for GitpodWsDaemonCrashLooping

mads-hartmann:
### Context

Looking at the number of PagerDuty alerts ("Incidents") each month over the last year our current development isn't looking good.

Screenshot 2022-05-31 at 13 59 41

I dug into the specific alerts in this PR https://github.com/gitpod-io/ops/pull/2518 and the alerts responsible for 2/3rd of the "incidents" in the last month were `GitpodWsDaemonCrashLooping (54)` and `GitpodNodeConntrackTableIsFull (50)`

### Implementation

This changes the threshold for GitpodWsDaemonCrashLooping to be more than 2 restarts in 10 minutes instead of the current value of 1 restart in 10 minutes. My reasoning is that if it just restarted once then the damage is done but there really isn't anything useful the on-caller can do at this point. Restarting more than 2 times might indicate that it's actually crash-looping and perhaps the on-caller should trigger an incident. I have introduced a warning alert for restarts too, in case the workspace team wants to investigate restarts during work hours.

Looking at the last 7 days we'd go from something like > 30 alerts to 3 alerts with this threshold.

Threshold > 0

Screenshot 2022-05-31 at 13 45 51

Threshold > 2

Screenshot 2022-05-31 at 13 47 00

## Related Issue(s)

No issue

## How to test

N/A

## Release Notes

```release-note
NONE
```

## Documentation

N/A

### DIFF
--- a/operations/observability/mixins/workspace/rules/components/ws-daemon/alerts.libsonnet
+++ b/operations/observability/mixins/workspace/rules/components/ws-daemon/alerts.libsonnet
@@ -20,6 +20,20 @@
               description: 'Pod {{ $labels.namespace }}/{{ $labels.pod }} ({{ $labels.container }}) is restarting {{ printf "%.2f" $value }} times / 10 minutes.',
             },
             expr: |||
+              increase(kube_pod_container_status_restarts_total{container="ws-daemon"}[10m]) > 2
+            |||,
+          },
+          {
+            alert: 'GitpodWsDaemonRestarted',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              runbook_url: 'https://github.com/gitpod-io/runbooks/blob/main/runbooks/GitpodWsDaemonCrashLooping.md',
+              summary: 'Ws-daemon just restarted.',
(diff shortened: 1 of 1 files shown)


mads-hartmann:
Adding hold as the build failed.

kylos101 (commented):
@mads-hartmann I'm not sure that we should absorb this change now, but perhaps soon, later in June.

I say this because `ws-daemon` really shouldn't be restarting this often, and we have a couple fixes to help address it. [[1](https://github.com/gitpod-io/gitpod/pull/10274)][[2](https://github.com/gitpod-io/gitpod/pull/10384)]

In other words, even though it's painful, it's ideal to page when `ws-daemon` restarts, to see if any action is needed. Even if that results in an escalation. For example, there might be workspaces that (1) need to have finalizers removed (2) have their status changed in the database (3) etc.

That said, I think once we're backing up user data with PVC snapshotter (hopefully later in June), we'd want to revisit this PR. In other words, `ws-daemon` will be stateless at that point (I believe) and running our resource limiters.

@sagor999 what do you think?

mads-hartmann:
This was solved in another PR a while ago now. Closing this.

PR #10390, opened 2022-05-31
[usage] Harden parsing of time from VarChar field

easyCZ:
## Description

A consequence of using VarChar to store datetimes is that the varchar can be empty as well. This is handled on the TS side as optional but in Go, we need something akin to [`sql.NullString`](https://pkg.go.dev/database/sql#NullString) to handle this case.

This PR extends the VarCharTime struct to handle this case by setting validity and exposes methods to check if the value was set.

## Related Issue(s)

Fixes #

## How to test

Unit tests

## Release Notes

```release-note
NONE
```

## Documentation

NONE

### DIFF
--- a/components/usage/pkg/db/types.go
+++ b/components/usage/pkg/db/types.go
@@ -6,54 +6,83 @@ package db
 
 import (
 	"database/sql/driver"
-	"errors"
 	"fmt"
 	"github.com/relvacode/iso8601"
 	"time"
 )
 
 func NewVarcharTime(t time.Time) VarcharTime {
-	return VarcharTime(t.UTC())
+	return VarcharTime{
+		t:     t,
+		valid: true,
+	}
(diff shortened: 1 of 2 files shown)


components/usage/pkg/db/types.go
```diff
@@ -6,54 +6,83 @@ package db
...
 		if err != nil {
 			return fmt.Errorf("failed to parse %v into ISO8601: %w", string(s), err)
 		}
-		*n = VarcharTime(parsed.UTC())
+		n.valid = true
+		n.t = parsed.UTC()
+		return nil
+	case string:
+		if len(s) == 0 {
+			n.valid = false
```
easyCZ, line 57: We wouldn't necessarily need to do this as false is the default value, but it's here mostly for ease of reading.

PR #10388, opened 2022-05-31
[gitpod-cli] Add command 'gp ports list'

felladrin:
## Description

Adds command `gp ports list` to gitpod-cli, which prints the ports state.

## Related Issue(s)

#9930

## How to test

- Open the Gitpod repo on the Preview Environment of this PR: https://vn-gp-portbbf4ebbdcd.preview.gitpod-dev.com/#https://github.com/gitpod-io/gitpod
- Change the `ports` section from `.gitpod.yml` to the following:
```yml
ports:
  - port: 1337
    onOpen: open-preview
    description: Et omnis id unde veniam.
  - port: 3000
    onOpen: ignore
    description: Alias reiciendis est quidem. Saepe in aut est illum ex. Eius molestias et vero rem.
  - port: 3001
    onOpen: ignore
  - port: 3306
    onOpen: ignore
  - port: 4000
    onOpen: ignore
  - name: VNC
    port: 5900
    onOpen: ignore
    description: Qui quia porro officia.
  - name: noVNC
    port: 6080
    onOpen: ignore
  - name: Werft
    port: 7777
    onOpen: ignore
  - port: 9229
    onOpen: ignore
  - name: Go proxy
    port: 9999
    onOpen: ignore
  - port: 13001
    onOpen: ignore
    description: Ut explicabo et qui. Velit tempore porro impedit.
  - name: Dev Theia
    port: 13444
```
- Run `gp ports list` and confirm if the command outputs something like this:
image

## Release Notes

```release-note
Gitpod CLI has a new command to list the ports from the workspace: gp ports list
```

## Documentation

None.

PR #10403, opened 2022-06-01
Refactor cmd `tasks` and `ports` supervisor related func

mustard-mh:
## Description

Currently command `gp tasks` and `gp ports` not handle errors properly, i.e.
- Command not stop when supervisor grpc connect failed (`supervisor.Dial` func)
- Log with Error not Fatal and without return
- Package of client of supervisor called `supervisor` can be confusing

This PR 
- Move command files into `package cmd`
- Rename truly supervisor package from `api` to `supervisor`
- Move client of supervisor into `supervisor-helper` folder
- Fix i.e. problem above

For `gp tasks` 

- We don't change its logic, in case break something. If we want to refactor it, another PR will be good
- Update `GetTasksList` func for the same [reason](https://github.com/gitpod-io/gitpod/pull/10388#discussion_r886344771)

## Related Issue(s)

Fixes #

## How to test

- Start a workspace with current PR in preview env
- cd /workspace/gitpod/components/gitpod-cli
- Check if `go run main.go tasks xxx` behave the same with `gp tasks xxx`
- Check if `gp ports list` behave the same with prev PR https://github.com/gitpod-io/gitpod/pull/10388

## Release Notes

```release-note
NONE
```


felladrin (approved):
Code reviewed and commands (ports/tasks) tested! ✅

PR #10413, opened 2022-06-01
[supervisor] improve error message around user group and uid

sagor999:
## Description

[supervisor] improve error message around user group and uid

## Related Issue(s)

Fixes #

## Release Notes

```release-note
[supervisor] improve error message around user group and uid
```

PR #10414, opened 2022-06-01
[ws-manager] show why pod entered completed state

sagor999:
## Description

Whenever workspace fails due to pod entered completed state, show an error message, so that it is easier for user to understand what went wrong.
Example:
image

## Related Issue(s)

Fixes #

## How to test

Try to open this workspace:
https://github.com/atduarte/test-new

You should see a clear error why it failed. 
Here is a PR that also improved that specific error to be more clear: https://github.com/gitpod-io/gitpod/pull/10413

## Release Notes

```release-note
[ws-manager] show why pod entered completed state
```

### DIFF
--- a/components/ws-manager/pkg/manager/status.go
+++ b/components/ws-manager/pkg/manager/status.go
@@ -595,7 +595,7 @@ func extractFailure(wso workspaceObjects) (string, *api.WorkspacePhase) {
 					// default way for headless workspaces to be done
 					return "", nil
 				}
-				return fmt.Sprintf("container %s completed; containers of a workspace pod are not supposed to do that. error msg: %s", cs.Name, terminationState.Message), nil
+				return fmt.Sprintf("container %s completed; containers of a workspace pod are not supposed to do that. Reason: %s", cs.Name, terminationState.Message), nil
 			} else if !isPodBeingDeleted(pod) && terminationState.ExitCode != containerUnknownExitCode {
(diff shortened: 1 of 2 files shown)


components/ws-manager/pkg/manager/status.go
```diff
@@ -595,7 +595,7 @@ func extractFailure(wso workspaceObjects) (string, *api.WorkspacePhase) {
 					// default way for headless workspaces to be done
 					return "", nil
 				}
-				return fmt.Sprintf("container %s completed; containers of a workspace pod are not supposed to do that", cs.Name), nil
+				return fmt.Sprintf("container %s completed; containers of a workspace pod are not supposed to do that. error msg: %s", cs.Name, terminationState.Message), nil
```
csweichel, line 598: Nit: considering that the error is user-facing, `error msg` doesn't sit quite right :)

```suggestion
				return fmt.Sprintf("container %s completed; containers of a workspace pod are not supposed to do that. Reason: %s", cs.Name, terminationState.Message), nil
```

csweichel:
Other than the nit, LGTM.

/hold

csweichel (approved):

components/ws-manager/pkg/manager/status.go
sagor999: Good point! :heart:

Issue #9247, opened 2022-04-12
Error response from daemon: OCI runtime create failed /proc/self/fd


konne:
### Bug description

#8070 follow up

It just happend again

![image](https://user-images.githubusercontent.com/1100969/162907847-18c42359-ecb0-4343-9551-6f3609962910.png)

### Steps to reproduce

see descriptions

### Workspace affected

vizlib-gucci-0daa9pimywm.ws-eu39axl.gitpod.io

kylos101:
@utam0k this appears to have manifested again. Could we ask you to take a peek?

utam0k:
@konne I investigated the logs of this error. However, I didn't find this error message in the last 30 days. It's possible that my research for errors is incorrect.  Are you still encountering this error?
We have been running the integration test for the previous issue #8070 measures every day since about last week.

konne:
@utam0k still happens:
at least one developer had this issue yesterday at 3:58 PM GMT+2 on vizlib-gucci-yom27jlp1k9 workspace

utam0k:
@konne cc: @kylos101 
Thanks for your information. I found the same error message. This problem is certainly happening.
> at least one developer had this issue yesterday at 3:58 PM GMT+2 on vizlib-gucci-yom27jlp1k9 workspace

I tried to reproduce this problem with https://github.com/kylos101/OCI-runtime-create-failed, but I couldn't. How often did you encounter it?


And, today I found a curious thing the proc mount sys-call has sometimes failed since 4/30. I guess it may relate to this problem. However, there was insufficient information in the logs to trace from those failed logs to workspace information. I'll try creating PR to add some information to these logs tomorrow.
https://cloudlogging.app.goo.gl/oqgziQvYjPS2A2Xw5 (sorry, it is internal link)

utam0k:
And this seems to be a different bug than the last one, I couldn't reproduce the problem after 1000 iterations of mount proc.

on workspace
```console
for i in {1..%d}; do echo $i; sudo unshare -m --propagation unchanged mount -t proc proc $(mktemp -d) || exit 1; done
```

konne:
@utam0k  this bug happens during a complex docker-compose up we start:
kes, nginx, postgres, nats and about 5 go and nodejs services.
This happens about once a day per developer and I would say the developer does that docker-compose up maybe 3 times a day.
If this happens we can just start it again and usually, it is working then, so this is the reason the devs only report it to me if I actively ask for.

[...]

PR #10420, opened 2022-06-02
workspackit: put the workspace id into logs to trace.

utam0k:
## Description

We can't trace the workspace kit logs because it doesn't have the workspace id in the logs. This PR put a workspace id into the workspacekit log.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/9247

## How to test
[example logs](https://cloudlogging.app.goo.gl/EjKhF94qHoei1rXW7)

1. Open the worksapce
2. Run `sudo mount -t proc proc /proc/11111/`
3. Confirm that the log is included `worksapceId`

## Release Notes

```release-note
workspackit: put the workspace id into logs to trace.
```

## Documentation

No


components/workspacekit/cmd/rings.go
```diff
@@ -113,6 +120,7 @@ var ring0Cmd = &cobra.Command{
 		cmd.Env = append(os.Environ(),
 			"WORKSPACEKIT_FSSHIFT="+prep.FsShift.String(),
 			fmt.Sprintf("WORKSPACEKIT_NO_WORKSPACE_MOUNT=%v", prep.FullWorkspaceBackup || prep.PersistentVolumeClaim),
+			"GITPOD_WORKSPACE_ID="+wsid,
```
csweichel, line 123: We don't have to set this envvar, because we're passing `os.Environ()`.
This way we're setting it twice (same below)

utam0k: @csweichel Good point! Thank you. I fixed that.

Issue #9395, opened 2022-04-19
Prebuilds are stuck in 'queued'


svenefftinge:
### Bug description

The DB contains +6k entries of prebuilds in 'queued' state. Would be good to understand what happens, clean things up and prevent these from being created further.

#### Updating prebuilt.state on failed prebuild start
During [this incident](https://www.gitpodstatus.com/incidents/cpr3xxsw7dl6) we discovered that there seems to be an ongoing issue with updating state for prebuilds whose instances never reach `RUNNING` phase (note the empty `startedTime`):
```
SELECT wsi.id, wsi.creationTime, wsi.phasePersisted, wsi.status, wsi.startedTime, wsi.stoppingTime, wsi.stoppedTime
   FROM d_b_workspace_instance AS wsi
   JOIN d_b_prebuilt_workspace AS pws
      ON pws.buildWorkspaceId = wsi.workspaceId
   WHERE pws.state = 'queued'
	AND wsi.phasePersisted = 'stopped'
	ORDER BY wsi.creationTime DESC;
```

Note: Especially older entries could also be fallout from other incidents + cleanups, but there seems to be a couple of fresh ones every day.

### Steps to reproduce

`select * from d_b_prebuilt_workspace where state = 'queued' order by creationTime;`

svenefftinge:
I noticed that the `stoppingTime`s are always around the same seconds and happen in chunks at the same time. So probably some clean-up mechanism. 👀

geropl:
> The cases have in common that the phase is unknown (which has a timeout of 600 secs vs. 3600 secs for preparing).

This sounds related to a fix @andrew-farries is working on [here](https://github.com/gitpod-io/gitpod/pull/9453). At some point we seem to have started using `"unknown"` as the initial phase - which does not make sense to me, at all. We're seeking to fix this now, and 

@andrew-farries Can you make sure we [introduce separate timeouts](https://github.com/gitpod-io/gitpod/blob/main/components/ws-manager-bridge/src/meta-instance-controller.ts#L43) for "building", and start out with the same value as "preparing"?

svenefftinge:
Yes, that explains why they are no signs in workspace-cluster the workspaces are probably building and timing out in the meantime (after 10min).

jldec:
Suspected [example](https://app.frontapp.com/open/cnv_j5h2rbe) of affected user (internal)
Is there a workaround or DB change we can use to mitigate?

geropl:
[...]

PR #10425, opened 2022-06-02
[bridge] Extracting updating of prebuilds into PrebuildUpdater (2/3)

geropl:
## Description

This PR extract all logic related to updating prebuilds from src/ee/bridge behind a new interface `PrebuildUpdater` with two implementations: Noop (non-ee version) and DB (ee version) (2/3 in the chain)

## Related Issue(s)

Context: https://github.com/gitpod-io/gitpod/issues/9395

## How to test
 - start a workspace on this branch
 - start a workspace in the preview env
 - check the logs `kubectl logs ws-manager-bridge- ws-manager-bridge`
 - note how:
   - everything is working as expected
   - bridge logs show no errors

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/ws-manager-bridge/src/bridge.ts
+++ b/components/ws-manager-bridge/src/bridge.ts
@@ -37,6 +37,7 @@ import { WorkspaceCluster } from "@gitpod/gitpod-protocol/lib/workspace-cluster"
 import { repeat } from "@gitpod/gitpod-protocol/lib/util/repeat";
 import { PreparingUpdateEmulator, PreparingUpdateEmulatorFactory } from "./preparing-update-emulator";
 import { performance } from "perf_hooks";
+import { PrebuildUpdater } from "./prebuild-updater";
 
 export const WorkspaceManagerBridgeFactory = Symbol("WorkspaceManagerBridgeFactory");
 
@@ -72,6 +73,9 @@ export class WorkspaceManagerBridge implements Disposable {
 
     @inject(IAnalyticsWriter)
     protected readonly analytics: IAnalyticsWriter;
+
+    @inject(PrebuildUpdater)
+    protected readonly prebuildUpdater: PrebuildUpdater;
 
     protected readonly disposables = new DisposableCollection();
     protected readonly queues = new Map<string, Queue>();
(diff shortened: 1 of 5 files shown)


AlexTugarev (approved):
LGTM

/hold because of merge conflicts after 1/3 got merged

components/ws-manager-bridge/src/prebuild-updater.ts
```diff
@@ -0,0 +1,35 @@
+/**
+ * Copyright (c) 2020 Gitpod GmbH. All rights reserved.
```
laushinka, line 2: Is this supposed to be 2020?

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -462,24 +466,11 @@ export class WorkspaceManagerBridge implements Disposable {
             instance.stoppedTime = new Date().toISOString();
             promises.push(this.workspaceDB.trace({}).storeInstance(instance));
             promises.push(this.onInstanceStopped({}, ri.workspace.ownerId, instance));
-            promises.push(this.stopPrebuildInstance(ctx, instance));
+            promises.push(this.prebuildUpdater.stopPrebuildInstance(ctx, instance));
```
svenefftinge, line 469: @geropl I don't think this does anything useful. Looks to me like it would update the workspace instance that got set to 'stopped' a few lines above now to 'aborted'. Am I missing something?

geropl: > Looks to me like it would update the workspace instance that got set to 'stopped' a few lines above now to 'aborted'. Am I missing something?

It updates the `PrebuiltWorkspace` that's corresponding that `WorkspaceInstance` to `aborted`, which would otherwise not bet updated.

In what context are you looking into this? Can I help?

svenefftinge: > It updates the PrebuiltWorkspace that's corresponding that WorkspaceInstance to aborted, which would otherwise not bet updated.

Why does it do that? And how can a prebuild not be finished when there is a workspace based on it?

> In what context are you looking into this? Can I help?

Just reading the code and trying to make sense of it 😇

svenefftinge: Ah, got it now. Forget it 😁

svenefftinge: What confused me is that the method is called `stopPrebuildInstance` while it only aborts the PrebuildWorkspace in the database.

geropl:
> Why does it do that? And how can a prebuild not be finished when there is a workspace based on it?

I think we went down the road of having an explicit state model for Prebuilds to better decouple tables and implementation.

> Just reading the code and trying to make sense of it 

:+1:

components/ws-manager-bridge/src/bridge.ts
geropl: All good :wink:

PR #10435, opened 2022-06-02
[prebuilds] add metrics for prebuilds 10383

### LINKED ISSUE #10383: Add metrics to be able to identify leaking prebuild
Came up in the context of:
 - https://github.com/gitpod-io/gitpod/issues/8195: prebuilds are stuck in running because of an (suspected) supervisor bug (@AlexTugarev please fill in issue id HERE once created :pray: )
 - https://github.com/gitpod-io/gitpod/issues/9395#issuecomment-1142058473: prebuilds are stuck in `queued` for multiple reasons

First idea:
 - have counters on "queued" ([WorkspaceFactory](https://github.com/gitpod-io/gitpod/blob/81a79af4ec3f9a5de53f40370ab455bf8a054c85/components/server/ee/src/workspace/workspace-factory.ts#L168-L178) in `server`), "running" and end states ("available", "failed", in [bridge](https://github.com/gitpod-io/gitpod/blob/bb0a42a48c39ab6c1457d0624518415a2c3be596/components/ws-manager-bridge/ee/src/bridge.ts#L108), while making sure to only capture flanks :+1: )
 - subtracting those should give a fairly good trend

laushinka:
## Description
Provides counter metrics for prebuilds with states as labels.
- gitpod_prebuilds_started_total ("queued")
- gitpod_prebuilds_completed_total ("available", "failed", "timeout", "aborted" )

## Related Issue(s)

Fixes #10383 

## How to test
1. Trigger prebuilds
2. Check that those prebuilds show up on this Prometheus query page[[1](https://9090-gitpodio-gitpod-wbcbfz9sl3s.ws-eu46.gitpod.io/graph?g0.expr=gitpod_prebuilds_total&g0.tab=0&g0.stacked=0&g0.show_exemplars=0&g0.range_input=1h)]

## Release Notes

```release-note
NONE
```

## Documentation

/werft run with-observability

### DIFF
--- a/components/server/ee/src/workspace/workspace-factory.ts
+++ b/components/server/ee/src/workspace/workspace-factory.ts
@@ -30,7 +30,7 @@ import { ErrorCodes } from "@gitpod/gitpod-protocol/lib/messaging/error";
 import { HostContextProvider } from "../../../src/auth/host-context-provider";
 import { UserDB } from "@gitpod/gitpod-db/lib";
 import { UserCounter } from "../user/user-counter";
-
+import { increasePrebuildsStartedCounter } from "../../../src/prometheus-metrics";
 @injectable()
 export class WorkspaceFactoryEE extends WorkspaceFactory {
     @inject(LicenseEvaluator) protected readonly licenseEvaluator: LicenseEvaluator;
@@ -176,6 +176,10 @@ export class WorkspaceFactoryEE extends WorkspaceFactory {
                 branch,
                 statusVersion: 0,
             });
+
+            if (pws) {
+                increasePrebuildsStartedCounter();
+            }
 
             log.debug(
                 { userId: user.id, workspaceId: ws.id },
--- a/components/server/src/prometheus-metrics.ts
+++ b/components/server/src/prometheus-metrics.ts
@@ -151,3 +151,13 @@ const instanceStartsFailedTotal = new prometheusClient.Counter({
 export function increaseFailedInstanceStartCounter(reason: "clusterSelectionFailed" | "startOnClusterFailed") {
     instanceStartsFailedTotal.inc({ reason });
 }
+
+const prebuildsStartedTotal = new prometheusClient.Counter({
+    name: "gitpod_prebuilds_started_total",
+    help: "Counter of total prebuilds started.",
+    registers: [prometheusClient.register],
+});
+
+export function increasePrebuildsStartedCounter() {
+    prebuildsStartedTotal.inc();
+}
--- a/components/ws-manager-bridge/ee/src/prebuild-updater-db.ts
+++ b/components/ws-manager-bridge/ee/src/prebuild-updater-db.ts
@@ -70,6 +70,7 @@ export class PrebuildUpdaterDB implements PrebuildUpdater {
             prebuild.statusVersion = status.statusVersion;
 
             const update = this.prebuildStateMapper.mapWorkspaceStatusToPrebuild(status);
+            const terminatingStates = ["available", "timeout", "aborted", "failed"];
             if (update) {
                 const updatedPrebuild = {
                     ...prebuild,
@@ -78,6 +79,10 @@ export class PrebuildUpdaterDB implements PrebuildUpdater {
 
                 span.setTag("updatePrebuildWorkspace.prebuild.state", updatedPrebuild.state);
                 span.setTag("updatePrebuildWorkspace.prebuild.error", updatedPrebuild.error);
+
+                if (writeToDB && updatedPrebuild.state && terminatingStates.includes(updatedPrebuild.state)) {
+                    this.prometheusExporter.increasePrebuildsCompletedCounter(updatedPrebuild.state);
+                }
 
                 if (writeToDB) {
                     await this.workspaceDB.trace({ span }).storePrebuiltWorkspace(updatedPrebuild);
--- a/components/ws-manager-bridge/src/prometheus-metrics-exporter.ts
+++ b/components/ws-manager-bridge/src/prometheus-metrics-exporter.ts
@@ -18,6 +18,7 @@ export class PrometheusMetricsExporter {
     protected readonly clusterCordoned: prom.Gauge<string>;
     protected readonly statusUpdatesTotal: prom.Counter<string>;
     protected readonly stalePrebuildEventsTotal: prom.Counter<string>;
+    protected readonly prebuildsCompletedTotal: prom.Counter<string>;
 
     protected readonly workspaceInstanceUpdateStartedTotal: prom.Counter<string>;
     protected readonly workspaceInstanceUpdateCompletedSeconds: prom.Histogram<string>;
@@ -72,6 +73,12 @@ export class PrometheusMetricsExporter {
             // and outcomes by read-only replicas.
             labelNames: ["db_write", "workspace_cluster", "workspace_instance_type", "outcome"],
             buckets: prom.exponentialBuckets(2, 2, 8),
+        });
+
+        this.prebuildsCompletedTotal = new prom.Counter({
+            name: "gitpod_prebuilds_completed_total",
+            help: "Counter of total prebuilds ended.",
+            labelNames: ["state"],
         });
     }
 
@@ -141,5 +148,9 @@ export class PrometheusMetricsExporter {
         this.workspaceInstanceUpdateCompletedSeconds
             .labels(String(dbWrite), workspaceCluster, WorkspaceType[type], outcome)
             .observe(durationSeconds);
+    }
+
+    increasePrebuildsCompletedCounter(state: string) {
+        this.prebuildsCompletedTotal.inc({ state });
     }
 }


components/ws-manager-bridge/ee/src/prebuild-updater-db.ts
```diff
@@ -78,6 +78,7 @@ export class PrebuildUpdaterDB implements PrebuildUpdater {
 
                 span.setTag("updatePrebuildWorkspace.prebuild.state", updatedPrebuild.state);
                 span.setTag("updatePrebuildWorkspace.prebuild.error", updatedPrebuild.error);
+                this.prometheusExporter.increasePrebuildsCounter(updatedPrebuild.state);
```
laushinka, line 81: @geropl I ended up using only one counter because here the updated state can be either starting ("building") or ending.

components/ws-manager-bridge/src/prometheus-metrics-exporter.ts
```diff
@@ -73,6 +74,12 @@ export class PrometheusMetricsExporter {
             labelNames: ["db_write", "workspace_cluster", "workspace_instance_type", "outcome"],
             buckets: prom.exponentialBuckets(2, 2, 8),
         });
+
+        this.prebuildsTotal = new prom.Counter({
```
laushinka, line 78: I created a duplicate one here because `prometheus-metrics.ts` isn't accessible from `ws-manager-bridge`. The metrics show up just fine as one in Prometheus, but please let me know if there is a better way 🙏🏽

components/ws-manager-bridge/ee/src/prebuild-updater-db.ts
easyCZ: Semantically, a counter is not correct for this. You're tracking state which changes over time. This means that if I do `sum(gitpod_prebuilds_total)` I will receive 1 for each state a prebuild has been in, not a total number of prebuilds.

For this case, you need to use multiple metrics - 1 gauge to track the state. If you also want to know things like # of started prebuilds vs # of succesful prebuilds, you may need extra metrics.

easyCZ (dismissed):
Left a comment about semantics of the metrics.

components/ws-manager-bridge/ee/src/prebuild-updater-db.ts
laushinka: Thanks for the comment, @easyCZ! Indeed that for the total number of prebuilds, we will need a gauge. However that means reading the DB and we weren't sure that would be the best way. Having a counter for the amount of each state a prebuild has been in would serve the purpose of comparing how many `queued` and how many ended up in an "end" state. With this we can see how many never got to an end state and debug from there. 
What do you think?

easyCZ: Also depending on the desired accuracy, you may have to read from the DB as the DB is the only source of truth. If we are happy with some amount of error, handling this from the ws-manager-bridge is fine (but expect that when ws-manager-bridge fails, or otherwise doesn't work, your metrics will be off)

laushinka: After discussing with @easyCZ, we'll have this PR to add the counter metrics and see how they help us. If in the future we deem a gauge necessary then we will follow-up with it. 
A fix will be made in this PR to separate the counter into two - started and completed. It is now a draft until the fix is pushed.

components/server/src/prometheus-metrics.ts
```diff
@@ -151,3 +151,25 @@ const instanceStartsFailedTotal = new prometheusClient.Counter({
 export function increaseFailedInstanceStartCounter(reason: "clusterSelectionFailed" | "startOnClusterFailed") {
     instanceStartsFailedTotal.inc({ reason });
 }
+
+const prebuildsStartedTotal = new prometheusClient.Counter({
+    name: "gitpod_prebuilds_started_total",
+    help: "Counter of total prebuilds started.",
+    labelNames: ["state"],
```
easyCZ, line 158: Why do you need the state when the metric is explicitly for "started" prebuilds?

components/server/src/prometheus-metrics.ts
```diff
@@ -151,3 +151,25 @@ const instanceStartsFailedTotal = new prometheusClient.Counter({
...
+});
+
+export function increasePrebuildsStartedCounter(state: string) {
+    prebuildsStartedTotal.inc({ state });
+}
+
+const prebuildsEndedTotal = new prometheusClient.Counter({
+    name: "gitpod_prebuilds_ended_total",
+    help: "Counter of total prebuilds ended.",
+    labelNames: ["state"],
```
easyCZ, line 169: We run multiple replicas of ws-manager-bridge, this means you'll get this metric reported from multiple pods. But not all are `governing` - that means actually affecting the database. This presents a challenge as your data here is not going to reflect reality, and will be `n` times larger depending on the number of replicas. One way to address that is to use the `governing` flag `ws-manager-bridge` has. You can for example add it as a label which will allow you to filter out the extra instances, or better, only increment the metric on governing instances.

components/ws-manager-bridge/ee/src/prebuild-updater-db.ts
```diff
@@ -79,6 +79,10 @@ export class PrebuildUpdaterDB implements PrebuildUpdater {
                 span.setTag("updatePrebuildWorkspace.prebuild.state", updatedPrebuild.state);
                 span.setTag("updatePrebuildWorkspace.prebuild.error", updatedPrebuild.error);
 
+                if (updatedPrebuild.state && updatedPrebuild.state !== "building") {
```
easyCZ, line 82: I'd recommend flipping the logic - list all the "terminal" states for which we do increment the metric. It's more resilient to new states being introduced in the future. 

A better alternative yet, add an instance method on the `prebuild` object to check if it's in terminal state. That way, you can co-locate the logic around the definition of state - higher chance we'll not forget to update something when we make changes.

components/server/src/prometheus-metrics.ts
easyCZ: The `writeToDB` flag being true is what makes an instance governing.

components/ws-manager-bridge/ee/src/prebuild-updater-db.ts
laushinka: > list all the "terminal" states for which we do increment the metric

Agree it would be more resilient 👍🏼

components/server/src/prometheus-metrics.ts
laushinka: > The `writeToDB` flag being true is what makes an instance governing.

TIL!

components/server/ee/src/workspace/workspace-factory.ts
```diff
@@ -177,6 +178,10 @@ export class WorkspaceFactoryEE extends WorkspaceFactory {
                 statusVersion: 0,
             });
 
+            if (pws) {
+                increasePrebuildsStartedCounter();
```
easyCZ, line 182: This is also affected by the multi-replica problem. So depending on which solution you go for, it needs to be done here too.

geropl:
@laushinka Starting to review now.

components/server/src/prometheus-metrics.ts
```diff
@@ -151,3 +151,14 @@ const instanceStartsFailedTotal = new prometheusClient.Counter({
 export function increaseFailedInstanceStartCounter(reason: "clusterSelectionFailed" | "startOnClusterFailed") {
     instanceStartsFailedTotal.inc({ reason });
 }
+
+const prebuildsStartedTotal = new prometheusClient.Counter({
+    name: "gitpod_prebuilds_started_total",
+    help: "Counter of total prebuilds started.",
+    labelNames: ["is_governing"],
```
geropl, line 158: The label is not necessary.

components/ws-manager-bridge/ee/src/prebuild-updater-db.ts
```diff
@@ -79,6 +80,10 @@ export class PrebuildUpdaterDB implements PrebuildUpdater {
                 span.setTag("updatePrebuildWorkspace.prebuild.state", updatedPrebuild.state);
                 span.setTag("updatePrebuildWorkspace.prebuild.error", updatedPrebuild.error);
 
+                if (updatedPrebuild.state && terminatingStates.includes(updatedPrebuild.state)) {
```
geropl, line 83: The `writeToDB`/`is_governing` label should not be in the metric but checked in this if statement:
`if (writeToDB && terminatingStates.includes(updatedPrebuild.state)) { .... }`

components/ws-manager-bridge/src/prometheus-metrics-exporter.ts
```diff
@@ -73,6 +74,12 @@ export class PrometheusMetricsExporter {
             labelNames: ["db_write", "workspace_cluster", "workspace_instance_type", "outcome"],
             buckets: prom.exponentialBuckets(2, 2, 8),
         });
+
+        this.prebuildsCompletedTotal = new prom.Counter({
+            name: "gitpod_prebuilds_completed_total",
+            help: "Counter of total prebuilds ended.",
+            labelNames: ["state", "is_governing"],
```
geropl, line 81: As mentioned before: `is_governing` is not needed here.

components/server/src/prometheus-metrics.ts
geropl: > We run multiple replicas of ws-manager-bridge

... but only ever one "bridge" is "governing" a given cluster. Thus we don't care about that flag, but should write the metric based on it. See [comment](https://github.com/gitpod-io/gitpod/pull/10435#discussion_r891991889).

easyCZ: Basically what we're saying is that we should have
```
if (writeToDB) {
	reportMetric()
}
```

PR #10436, opened 2022-06-02
[previewctl] Improve logging

ArthurSens:
## Description

To improve experience while debugging when previewctl is not behaving as it should, this PR does 2 things:
* Adopt a better logging library.
* Log when `install-context --watch` fails to install and goes for another loop

## How to test

* Create a new branch from this PR.
* Run `previewctl install-context --watch`
* Run a werft job from your new branch, so it creates a new preview

You'll notice that the watch still doesn't work as expected (for me it stops looping at some point), but at least now we know what is happening 😅 

## Release Notes

```release-note
NONE
```


dev/preview/previewctl/cmd/get_name.go
```diff
@@ -8,16 +8,17 @@ import (
 	"fmt"
 
 	"github.com/gitpod-io/gitpod/previewctl/pkg/preview"
+	"github.com/go-kit/log"
```
csweichel, line 11: All code in Gitpod uses `logrus` - we even tie `klog` (used by Kubernetes components) back to logrus.

What's the motivation to go astray? I.e. which requirements warrant the use of a different library?

ArthurSens: None really, just wanted to move away from the previous 🙂

I'll update the PR to logrus

csweichel: best use `common-go/log`, but using logrus directly is fine, too.

ArthurSens: I believe adopting a big lib just for one of the features may be too much. 

I've only switched to logrus for now, but if someday in the future we need tracing/metrics/pprof, then common-go sounds like a good idea 🙂

dev/preview/previewctl/main.go
```diff
@@ -5,14 +5,20 @@
...
+	"github.com/sirupsen/logrus"
 )
 
 func main() {
-	root := cmd.RootCmd()
+	logger := logrus.New()
+	logger.SetFormatter(&logrus.TextFormatter{
+		DisableColors:   true,
+		FullTimestamp:   true,
+		TimestampFormat: "2006-01-02 15:04:05",
```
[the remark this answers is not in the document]
ArthurSens: We're not running it in CI though 😅, it is a tool used by ourselves and colleagues. I think the original format adds noise that we humans always ignore 😬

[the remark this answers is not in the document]
ArthurSens: I believe this is only the first command (which had the highest priority). The following milestones and commands will be focused on interactions through the workspace terminal 🙂

vulkoingim (approved):

PR #10442, opened 2022-06-02
Revert "[baseserver] Change default metrics port to 9502 to not clash with kube-rbac-proxy"

aledbf:
Reverts gitpod-io/gitpod#10109

PR #10443, opened 2022-06-02
Simplify configuration of KubeRBACProxyContainerWithConfig and remove collision

aledbf:
## Description

Gitpod components listen for metrics on localhost port 9500. Simplify KubeRBACProxyContainerWithConfig and remove any chance to brake that definition.

## How to test
- All components should start without issues
- Prometheus metrics should listen on `127.0.0.1:9500`
 
## Release Notes
```release-note
Simplify configuration of KubeRBACProxyContainerWithConfig and remove collision
```

Requires https://github.com/gitpod-io/gitpod/pull/10442

### DIFF
--- a/components/common-go/baseserver/server.go
+++ b/components/common-go/baseserver/server.go
@@ -364,7 +364,7 @@ func newBuiltinServices(server *Server) *builtinServices {
 			Handler: server.healthEndpoint(),
 		},
 		Metrics: &http.Server{
-			Addr:    fmt.Sprintf(":%d", BuiltinMetricsPort),
+			Addr:    fmt.Sprintf("127.0.0.1:%d", BuiltinMetricsPort),
 			Handler: server.metricsEndpoint(),
 		},
 	}
--- a/install/installer/pkg/common/common.go
+++ b/install/installer/pkg/common/common.go
@@ -13,6 +13,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/gitpod-io/gitpod/common-go/baseserver"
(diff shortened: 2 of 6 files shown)


components/common-go/baseserver/server.go
```diff
@@ -364,7 +364,7 @@ func newBuiltinServices(server *Server) *builtinServices {
 			Handler: server.healthEndpoint(),
 		},
 		Metrics: &http.Server{
-			Addr:    fmt.Sprintf(":%d", BuiltinMetricsPort),
+			Addr:    fmt.Sprintf("127.0.0.1:%d", BuiltinMetricsPort),
```
aledbf, line 367: This was the reason of the collision

sagor999 (approved):

aledbf:
> @easyCZ ^ I remember you specifically changing it to :

Good point. Even if we run this locally, we should never expose metrics outside the local machine without an additional security layer.

components/common-go/baseserver/server.go
easyCZ: I've changed it away from localhost:port as that listened only on local host and couldn't be reached from outside the container. This should work also. Will check preview shortly.

easyCZ (approved):
Makes sense. This is indeed cleaner than what I had. I ran into problems on port collision because I was binding on all listeners (due to `:port`) rather than just the loopback traffic

Pothulapati (approved):
LGTM, and makes sense to use local host (i.e `127.0.0.1` now) for metrics while also making sure port collision does not happen with components like `public-api`. :shipit:

PR #10450, opened 2022-06-03
[dashboard] Move inactive workspaces out of sight

### LINKED ISSUE #1053: batch delete of workspaces
I would like to select several workspaces for deletion.

### LINKED ISSUE #4233: Make it easier to delete multiple or all workspaces
### Problem to solve

Although workspaces eventually time out and stopped workspaces are getting deleted according to the [garbage collection](https://www.gitpod.io/docs/life-of-workspace#garbage-collection), sometimes a user would like to manually delete multiple or all workspaces regarding running status.

See also [relevant discussion](https://gitpod.slack.com/archives/C01KGM9AW4W/p1621335034023700) (internal). /cc @princerachit @corneliusludmann 

### Proposal

TBD (To be discussed)

First MVC (Minimum Viable Change) could be to add replace the status indicator with a checkbox on hover so that users can select multiple workspaces and then click on the _Delete_ button. Additionally, a checkbox at the top of the list to select all and delete all could be also helpful.

An early design spec could be the following:

| Possible solution for multiple workspace deletion |
|-|
| image |

svenefftinge:
## Description
Many users are manually deleting workspaces, keeping their workspaces list clean. 
Users should feel less urge to manually clean up and cleaning up should be simpler.

This PR, makes the inactive workspaces section on the /workspaces list collapsable (collapsed by default) to give a cleaner impression. When expanded there is a button to delete them.

- [ ] TODO: Keep workspaces of the last 12 hours active

## Related Issue(s)

Fixes #1053
Fixes #4233

## How to test
Use the preview env and start a few workspaces, stop them. Check the workspaces list.

## Release Notes

```release-note
Added action to delete all inactive workspaces
```

## Documentation

Screenshot 2022-06-03 at 12 02 12
Screenshot 2022-06-03 at 12 02 18
Screenshot 2022-06-03 at 12 02 24

### DIFF
--- a/components/dashboard/src/workspaces/Workspaces.tsx
+++ b/components/dashboard/src/workspaces/Workspaces.tsx
@@ -18,6 +18,8 @@ import { User } from "@gitpod/gitpod-protocol";
 import { useLocation } from "react-router";
 import { StartWorkspaceModalContext, StartWorkspaceModalKeyBinding } from "./start-workspace-modal-context";
 import SelectIDEModal from "../settings/SelectIDEModal";
+import Arrow from "../components/Arrow";
+import ConfirmationModal from "../components/ConfirmationModal";
 
 export interface WorkspacesProps {}
 
@@ -25,6 +27,8 @@ export interface WorkspacesState {
     workspaces: WorkspaceInfo[];
     isTemplateModelOpen: boolean;
     repos: WhitelistedRepository[];
+    showInactive: boolean;
+    deleteModalVisible: boolean;
 }
 
 export default function () {
@@ -35,6 +39,8 @@ export default function () {
     const [activeWorkspaces, setActiveWorkspaces] = useState<WorkspaceInfo[]>([]);
     const [inactiveWorkspaces, setInactiveWorkspaces] = useState<WorkspaceInfo[]>([]);
     const [workspaceModel, setWorkspaceModel] = useState<WorkspaceModel>();
+    const [showInactive, setShowInactive] = useState<boolean>();
+    const [deleteModalVisible, setDeleteModalVisible] = useState<boolean>();
     const { setIsStartWorkspaceModalVisible } = useContext(StartWorkspaceModalContext);
 
     useEffect(() => {
@@ -49,6 +55,18 @@ export default function () {
     return (
         <>
             <Header title="Workspaces" subtitle="Manage recent and stopped workspaces." />
+
+            <ConfirmationModal
+                title="Delete Inactive Workspaces"
+                areYouSureText="You are about to permanently delete your inactive workspaces."
+                buttonText="Delete Inactive Workspaces"
+                visible={!!deleteModalVisible}
+                onClose={() => setDeleteModalVisible(false)}
+                onConfirm={() => {
+                    inactiveWorkspaces.forEach((ws) => workspaceModel?.deleteWorkspace(ws.workspace.id));
+                    setDeleteModalVisible(false);
+                }}
+            ></ConfirmationModal>
 
             {isOnboardingUser && <SelectIDEModal location={"workspace_list"} />}
 
@@ -128,27 +146,58 @@ export default function () {
                             })}
                             {activeWorkspaces.length > 0 && <div className="py-6"></div>}
                             {inactiveWorkspaces.length > 0 && (
-                                <div className="p-3 text-gray-400 bg-gray-50 dark:bg-gray-800 rounded-xl text-sm text-center">
-                                    Unpinned workspaces that have been inactive for more than 14 days will be
-                                    automatically deleted.{" "}
-                                    <a
-                                        className="gp-link"
-                                        href="https://www.gitpod.io/docs/life-of-workspace/#garbage-collection"
+                                <div className="pt-14">
+                                    <div className="border-t border-gray-200 dark:border-gray-800"></div>
+                                    <div
+                                        onClick={() => setShowInactive(!showInactive)}
+                                        className="flex cursor-pointer py-6"
                                     >
-                                        Learn more
-                                    </a>
+                                        <div className="flex flex-col">
+                                            <h2 className="">Inactive Workspaces</h2>
+                                        </div>
+                                        <div>
+                                            <Arrow
+                                                up={!!showInactive}
+                                                customBorderClasses="text-gray-400 dark:text-gray-500 hover:text-gray-600 dark:hover:text-gray-400"
+                                            />
+                                        </div>
+                                    </div>
+                                    {showInactive ? (
+                                        <>
+                                            <div className="flex flex-row p-3 text-gray-400 bg-gray-50 dark:bg-gray-800 rounded-xl">
+                                                <div className="text-sm flex-auto py-3">
+                                                    Unpinned workspaces that have been inactive for more than 14 days
+                                                    will be automatically deleted.{" "}
+                                                    <a
+                                                        className="gp-link"
+                                                        href="https://www.gitpod.io/docs/life-of-workspace/#garbage-collection"
+                                                    >
+                                                        Learn more
+                                                    </a>
+                                                </div>
+                                                <button
+                                                    onClick={() => setDeleteModalVisible(true)}
+                                                    className="ml-2 danger secondary w-64 p-2"
+                                                >
+                                                    Delete Inactive Workspaces
+                                                </button>
+                                            </div>
+                                            {inactiveWorkspaces.map((e) => {
+                                                return (
+                                                    <WorkspaceEntry
+                                                        key={e.workspace.id}
+                                                        desc={e}
+                                                        model={workspaceModel}
+                                                        stopWorkspace={(wsId) =>
+                                                            getGitpodService().server.stopWorkspace(wsId)
+                                                        }
+                                                    />
+                                                );
+                                            })}
+                                        </>
+                                    ) : null}
                                 </div>
                             )}
-                            {inactiveWorkspaces.map((e) => {
-                                return (
-                                    <WorkspaceEntry
-                                        key={e.workspace.id}
-                                        desc={e}
-                                        model={workspaceModel}
-                                        stopWorkspace={(wsId) => getGitpodService().server.stopWorkspace(wsId)}
-                                    />
-                                );
-                            })}
                         </ItemsList>
                     </>
                 ) : (
--- a/components/dashboard/src/workspaces/Workspaces.tsx
+++ b/components/dashboard/src/workspaces/Workspaces.tsx
@@ -58,7 +58,7 @@ export default function () {
 
             <ConfirmationModal
                 title="Delete Inactive Workspaces"
-                areYouSureText="You are about to permanently delete your inactive workspaces."
+                areYouSureText="Are you sure you want to delete all inactive workspaces?"
                 buttonText="Delete Inactive Workspaces"
                 visible={!!deleteModalVisible}
                 onClose={() => setDeleteModalVisible(false)}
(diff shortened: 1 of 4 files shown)


svenefftinge:
@gtsiolis Can you take a look?
Also, I wonder if we should not display the empty state, in case there are no active workspaces. I.e. the "inactive workspaces" section would move below that empty space?

components/dashboard/src/workspaces/workspace-model.ts
```diff
@@ -145,8 +145,12 @@ export class WorkspaceModel implements Disposable, Partial<GitpodClient> {
     }
 
     protected isActive(info: WorkspaceInfo): boolean {
+        const lastSessionStart = WorkspaceInfo.lastActiveISODate(info);
+        const twentyfourHoursAgo = new Date(Date.now() - 1000 * 60 * 60 * 24).toISOString();
```
easyCZ, line 149: In `timeutil.ts` there are helpers for this - `hoursBefore(lastSessionStart, 24)`

components/dashboard/src/workspaces/workspace-model.ts
```diff
@@ -145,8 +145,12 @@ export class WorkspaceModel implements Disposable, Partial<GitpodClient> {
...
         return (
-            (info.workspace.pinned || (!!info.latestInstance && info.latestInstance.status?.phase !== "stopped")) &&
+            (info.workspace.pinned ||
+                (!!info.latestInstance && info.latestInstance.status?.phase !== "stopped") ||
+                twentyfourHoursAgo.localeCompare(lastSessionStart) < 0) &&
```
easyCZ, line 153: Similarly, we may be able to use `isDateGreaterOrEqual(d1, d2)` for the comparison here. Unsure if this works with the locale however.

easyCZ (approved):
Code-wise looks good to me, thanks for the change. Approving but adding hold for George to have a look at designs.

/hold

components/dashboard/src/workspaces/Workspaces.tsx
```diff
@@ -128,27 +146,58 @@ export default function () {
...
+                                                        href="https://www.gitpod.io/docs/life-of-workspace/#garbage-collection"
+                                                    >
+                                                        Learn more
+                                                    </a>
+                                                </div>
+                                                <button
+                                                    onClick={() => setDeleteModalVisible(true)}
+                                                    className="ml-2 danger secondary w-64 p-2"
+                                                >
+                                                    Delete Inactive Workspaces
```
gtsiolis, line 182: **thought**: I also swayed between keeping the alert component with the nested action button and having a separate element for toggling the list of inactive workspaces. See [design specs](https://www.figma.com/file/wcSB2SfgmaJCAdIVjarEmp/Move-inactive-workspaces-out-of-sight-%2310450?node-id=0%3A1) for more explorations.

components/dashboard/src/workspaces/Workspaces.tsx
```diff
@@ -128,27 +146,58 @@ export default function () {
...
                             {activeWorkspaces.length > 0 && <div className="py-6"></div>}
                             {inactiveWorkspaces.length > 0 && (
-                                <div className="p-3 text-gray-400 bg-gray-50 dark:bg-gray-800 rounded-xl text-sm text-center">
-                                    Unpinned workspaces that have been inactive for more than 14 days will be
-                                    automatically deleted.{" "}
-                                    <a
-                                        className="gp-link"
-                                        href="https://www.gitpod.io/docs/life-of-workspace/#garbage-collection"
+                                <div className="pt-14">
+                                    <div className="border-t border-gray-200 dark:border-gray-800"></div>
```
gtsiolis, line 150: **suggestion**: The border helps, I'd suggest to remove it and find other ways to create the needed visual balance here. Reusing the same structure of the workspace entry element could help (see :a:) as this would feel like a natural continuation of the active workspaces list. In case this still feelsSince this could

| BEFORE | AFTER :a: | AFTER :b: |
|-|-|-|
| InactiveWorkspaces-0 | InactiveWorkspaces-a | InactiveWorkspaces-b |

Posting below some early iterations for future reference:

| Iteration 1 | Iteration 2 | Iteration 3 |
|-|-|-|
| InactiveWorkspaces-3 | InactiveWorkspaces-5 | InactiveWorkspaces-a |

components/dashboard/src/workspaces/Workspaces.tsx
```diff
@@ -128,27 +146,58 @@ export default function () {
...
+                                    <div
+                                        onClick={() => setShowInactive(!showInactive)}
+                                        className="flex cursor-pointer py-6"
                                     >
-                                        Learn more
-                                    </a>
+                                        <div className="flex flex-col">
+                                            <h2 className="">Inactive Workspaces</h2>
```
gtsiolis, line 156: **issue(non-blocking)**: While the heading here describes the logical separation from the active workspaces, there's no equivalent heading for the active workspaces, thus making this slightly confusing. Instead, using a clear verb-based action here could help.

```suggestion
                                            <h2 className="">Show Inactive Workspaces</h2>
```

components/dashboard/src/workspaces/Workspaces.tsx
gtsiolis, line 182: **suggestion**: Using the new collapsible element described above instead of the info alert component, could allow use to also be more flexible with other elements like the action button here.

| Link button | Normal button |
|-|-|
| InactiveWorkspaces-link | InactiveWorkspaces-button |

components/dashboard/src/workspaces/Workspaces.tsx
gtsiolis, line 156: **suggestion**: Regarding the question from https://github.com/gitpod-io/gitpod/pull/10450#issuecomment-1145817205, we can keep the existing behavior, that is, removing the empty state when there are no active workspaces.

| User with only inactive workspaces |
|-|
| InactiveWorkspaces |

gtsiolis (commented):
Thanks for picking this up and +1 for making this happen, @svenefftinge! 🍷 

Left some comments below containing bits from the design thinking while reviewing UX changes here.

Overall, I love the idea of hiding the inactive workspaces. Since we already have a block element for separating the active from inactive workspaces, I'd suggest keeping this block element instead of introducing a line separator. I'd expect more changes to land in the workspaces list soon, see [relevant discussion](https://gitpod.slack.com/archives/C01KPEPQYE6/p1654156584703959) (internal).

Also made [a prototype](https://www.figma.com/proto/wcSB2SfgmaJCAdIVjarEmp/Move-inactive-workspaces-out-of-sight-%2310450?page-id=0%3A1&node-id=1303%3A3139&viewport=-4932%2C-6396%2C0.63&scaling=min-zoom&starting-point-node-id=1303%3A3139) (internal) to visualize this better. Feedback is welcome.

components/dashboard/src/workspaces/Workspaces.tsx
gtsiolis, line 156: **suggestion(non-blocking)**: One more thing we could improve here is to make this element kind of dynamic so that when there are only inactive workspaces the dashboard still feel active, populated recent data, and ready to go.

| Approach :a: | Approach :b: |
|-|-|
| InactiveWorkspaces-t1 | InactiveWorkspaces-t2 |

This information (amount of inactive workspaces) could be also nice-to-have inside the deletion confirmation modal.

| Modal using approach :a: | Modal using approach :b: |
|-|-|
| InactiveWorkspaces | InactiveWorkspaces-1 |

components/dashboard/src/workspaces/Workspaces.tsx
```diff
@@ -50,6 +56,18 @@ export default function () {
         <>
             <Header title="Workspaces" subtitle="Manage recent and stopped workspaces." />
 
+            <ConfirmationModal
+                title="Delete Inactive Workspaces"
+                areYouSureText="You are about to permanently delete your inactive workspaces."
```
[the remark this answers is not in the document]
svenefftinge: I feel like "Delete Inactive Workspaces" and a big red button is pretty clear. Also, they are soft deleted. So in the unlikely case that someone accidentally clicks delete twice, we could help.

components/dashboard/src/workspaces/Workspaces.tsx
```diff
@@ -128,27 +146,58 @@ export default function () {
...
+                                                <div className="text-sm flex-auto py-3">
+                                                    Unpinned workspaces that have been inactive for more than 14 days
+                                                    will be automatically deleted.{" "}
+                                                    <a
+                                                        className="gp-link"
+                                                        href="https://www.gitpod.io/docs/life-of-workspace/#garbage-collection"
+                                                    >
+                                                        Learn more
+                                                    </a>
+                                                </div>
```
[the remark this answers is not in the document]
svenefftinge: Can you clarify what you want me to do here?

components/dashboard/src/workspaces/Workspaces.tsx
```diff
@@ -128,27 +146,58 @@ export default function () {
...
+                                            <Arrow
+                                                up={!!showInactive}
+                                                customBorderClasses="text-gray-400 dark:text-gray-500 hover:text-gray-600 dark:hover:text-gray-400"
+                                            />
+                                        </div>
+                                    </div>
+                                    {showInactive ? (
+                                        <>
+                                            <div className="flex flex-row p-3 text-gray-400 bg-gray-50 dark:bg-gray-800 rounded-xl">
+                                                <div className="text-sm flex-auto py-3">
```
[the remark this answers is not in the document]
svenefftinge: Sounds good. Also we can change more than one thing 😅

components/dashboard/src/workspaces/Workspaces.tsx
svenefftinge: What is the meaning of the circled arrow icon on the left?

gtsiolis: The purpose of this icon was to :a: mimic the element structure of the workspace list item to help to associate this element with the workspaces (see [principle of grouping](https://en.wikipedia.org/wiki/Principles_of_grouping)[^1]) and :b: resemble a [history icon](https://duckduckgo.com/?q=history+icon&atb=v314-1&iax=images&ia=images) to help convey the contents of this collapsible component.

However, this first iteration of the icon could be confusing due to similarities with an [undo icon](https://duckduckgo.com/?q=undo+icon&atb=v314-1&iax=images&ia=images).

Here are a few more variations we could try:

| Icon variations for the inactive workspaces element |
|-|
| InactiveWorkspaces |

| With the current history icon | With an improved history icon | 
|-|-|
| InactiveWorkspaces-a | InactiveWorkspaces-1 |

Alternatively, we could simply replace the icon with an arrow icon from the first line. Dropping the icon altogether could easily create visual imbalance I'd like to avoid.

| With an arrow icon | Without an icon |
|-|-|
| InactiveWorkspaces-b | InactiveWorkspaces-c |

[^1]: Elements that are visually similar will be perceived as related.

components/dashboard/src/workspaces/Workspaces.tsx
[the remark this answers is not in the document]
gtsiolis: DEAL—Let's drop the idea of the extra alert component inside the modal. 🤜 🤛

components/dashboard/src/workspaces/Workspaces.tsx
gtsiolis: Minor after thought here. If we're going with the more explicit verb action here, we would need to also change this to _Hide Inactive Workspaces_ when the collapsible element is expanded.

svenefftinge:
Screenshot 2022-06-09 at 14 23 22
Screenshot 2022-06-09 at 14 23 16

components/dashboard/src/workspaces/Workspaces.tsx
gtsiolis: Marking as resolved. Consensus was to drop this idea. 🤝

components/dashboard/src/workspaces/Workspaces.tsx
```diff
@@ -128,27 +146,58 @@ export default function () {
...
+                                        <div>
+                                            <Arrow
+                                                up={!!showInactive}
+                                                customBorderClasses="text-gray-400 dark:text-gray-500 hover:text-gray-600 dark:hover:text-gray-400"
+                                            />
+                                        </div>
+                                    </div>
+                                    {showInactive ? (
+                                        <>
+                                            <div className="flex flex-row p-3 text-gray-400 bg-gray-50 dark:bg-gray-800 rounded-xl">
```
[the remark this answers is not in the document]
gtsiolis: Marking as resolved. ✔️

components/dashboard/src/workspaces/Workspaces.tsx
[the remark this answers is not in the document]
gtsiolis: Marking as resolved. Moved forward with a new (collapsible) component. ✔️

components/dashboard/src/workspaces/Workspaces.tsx
gtsiolis: Marking as resolved. Moved forward with an arrow icon. ✔️

components/dashboard/src/workspaces/Workspaces.tsx
gtsiolis: Marking as resolved. Moved forward with adding the workspace length in a parenthesis at the end of the section header. ✔️

gtsiolis: Marking as resolved. ✔️

components/dashboard/src/workspaces/Workspaces.tsx
gtsiolis: Marking as resolved. Moved forward with a secondary danger button. ✔️

components/dashboard/src/workspaces/Workspaces.tsx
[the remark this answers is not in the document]
gtsiolis: Marking as resolved, as this comment is no longer relevant. FWIW, this was referring to the missing icon on the left of the element before the latest updates in the designs. 🍷

david-bakin:
(As a user) I'm fine with hiding (folding away) inactive workspaces - but there is still a distinction between pinned and not-pinned, isn't there?

There used to be a separation on the workspaces page between pinned and unpinned.  I found that useful.  I'd like it back!

(Another suggestion, inappropriate for this PR, but just to provide food for thought: Users are allowed to _name_ their workspaces - but the name appears at the _end_ of the repo name - and there isn't much space for it!  See here where the names are "latest" and "tapscript-&lt;something-to-long-to-show>":

image

Maybe put the name first - or on a separate line if it gets too long?  Names are useful - in some use cases (much) more useful than branch names to keep what's in different workspaces clear in your head - it would be nice to have longer ones.)

svenefftinge:
The pinned workspaces are considered "active" and therefore always in the upper section. This is the same behavior as before.
Sound like you would benefit from a clear indicator of the pinned status as well, since giving your workspaces names feels like you care and want to keep them for longer? We can certainly improve how the name is displayed as suggested. Can you file an issue for this?

david-bakin:
> Can you file an issue for this?

#10773 - ty for the suggestion!

Issue #8261, opened 2021-06-17
Epic: Two Workspace Classes on SaaS


svenefftinge:
## Summary

Provide users with the ability to select between two different workspace classes for all their workspaces. On SaaS, the options are:
* Standard (up to 4 cores, up to 8GB RAM, 30GB storage)
* ~XL~ Large (up to 8 cores, up to 16GB RAM, 50GB storage)

⚠️ Note that because of the storage size differences, a XL workspace can't be restarted as Standard, and neither can a Standard workspace be started from an XL prebuild.

## Context 

Different projects require different resources. To allow developers to fully leverage the power and scalability of the cloud, we want to allow for different workspace classes.

[Internal Signal](https://github.com/gitpod-io/customers/issues/47)

## Value

Users will have access to better performance, that supports heavier workloads.

## Acceptance Criteria

* SaaS users that don't have a "legacy" monthly subscription must be able to choose between the two options (Standard and XL) in the Account Settings.
* SaaS users that have "legacy" monthly subscriptions should continue to use standard or XL according to the `excludeFromMoreResources` as they do today.
* SaaS users must be billed according to the workspace class they are using.
* Workspaces must not fail to start because the backup is larger than the storage available in the chosen class.
* Users should be able to determine what is the class of a given workspace, and what was the class used to run a prebuild

## Measurement

* % of workspaces that were XL.

## Growth Area

Expansion.

## Persona(s)

Users that have heavier workloads.

## Out of Scope

* Self-hosted use case.
* Having different workspace classes share the same node pool, and similar optimizations.
* Per project or per workspace class.

See https://github.com/gitpod-io/gitpod/issues/10805 to check what will follow.

## Tasks
- [x] #10842 
- [x] #11357
- [x] #10843 
- [ ] ~#10963~
- [x] #10981
- [x] #11358
- [x] #11583
- [x] [Configure workspace classes for SaaS](https://github.com/gitpod-io/ops/issues/3414)
- [x] #11839 
- [x] https://github.com/gitpod-io/gitpod/issues/11972
- [x] https://github.com/gitpod-io/gitpod/issues/12048 

<img src="https://front.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_3ox56)

csweichel:
[...]

PR #10454, opened 2022-06-03
Add workspace class to workspace instance

### LINKED ISSUE #10430: Store workspace size used for a given workspace instance
We are unable to determine which workspace size (or class, as it is mentioned in the code) was used for a given workspace instance. This a problem as different workspace sizes may have different storage capacities, and impacts our reporting capabilities.

Having one more column on `d_b_workspace_instance` with the workspace size name.

N/A

## Additional context
N/A

atduarte:
## Description
Adds the workspace class, also known as workspace size, to the workspace instance. This will be relevant to determine usage costs and for our own reporting.

A workspace class determines what type of resources a given workspace instance gets.

## Related Issue(s)

Fixes #10430

## How to test

Manual test using preview environments, using @jankeromnes [suggested steps](https://github.com/gitpod-io/gitpod/pull/10454#pullrequestreview-994836924).

[Results](https://github.com/gitpod-io/gitpod/pull/10454#issuecomment-1147246962)

## Release Notes

```release-note
NONE
```

## Documentation

- [x] /werft with-payment


components/gitpod-db/src/typeorm/entity/db-workspace-instance.ts
```diff
@@ -96,4 +96,7 @@ export class DBWorkspaceInstance implements WorkspaceInstance {
 
     @Column("simple-json", { nullable: true })
     imageBuildInfo?: ImageBuildInfo;
+
+    @Column({ nullable: true })
+    workspaceClass?: string;
```
jankeromnes, line 101: If you add a field here, you'll also want to add a migration file:
- See [How to add a migration file](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-db/src/typeorm/migration/README.md) and [an example migration file](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-db/src/typeorm/migration/1653983151212-TS2MoreResources.ts)

At this point, you can choose whether you want this field to be possibly `NULL`, or if you want to set every row to a default value during the migration (e.g. to `'default'` or empty string `''`).

---

Another thing to consider is: How to we specify all the workspace classes that exist?
- We can either specify them as "any string" (current implementation)
- Or, we can specify them as an enum of only specific string values (for example, like [WorkspaceType](https://github.com/gitpod-io/gitpod/blob/aafb99b6ba90658679ee9c5febe762881cb71c59/components/gitpod-protocol/src/protocol.ts#L566))
  - the problem being that someone installing Gitpod (e.g. Self-Hosted) might define any number of classes with weird names (e.g. we currently define only [gitpodio-internal-xl](https://github.com/gitpod-io/ops/blob/b980015f8efeaf900c24970f2e1c095624793f80/deploy/workspace/installer-values.yaml#L40) but we could add more and rename them in the future 💭)
  - however, agreeing on a common set of values will be necessary e.g. to correctly aggregate usage hours in our Usage component 💡

Personally, I'd love to go with an enum similar to this one:

```ts
type WorkspaceClass = "" | "S" | "M" | "L" | "XL";
```

(with `""` being the default value that we set every row to in our migration) but feel free to make a call here. We can always adjust later.

components/server/src/workspace/workspace-starter.ts
```diff
@@ -501,6 +501,7 @@ export class WorkspaceStarter {
 
                 instance.status.phase = "pending";
                 instance.region = installation;
+                instance.workspaceClass = startRequest.getSpec()!.getClass();
```
jankeromnes, line 504: 👍

jankeromnes (commented):
Hey, many thanks for implementing this @atduarte! 🎉

Regarding how this can be tested, I'd suggest the following:

- Add `- [x] /werft with-payment` at the bottom of your PR description, and trigger a new Werft build (this will enable paid plans)
- Set up a Chargebee webhook for this PR (I can do that for you)
- Connect to the database like so:
  - Open the PR in Gitpod
  - Run `./dev/preview/install-k3s-kubeconfig.sh` (manually 🙄)
  - Run `kubectl port-forward mysql-0 3306 &` (to expose the database port)
  - Run `mysql -h 127.0.0.1 -pjBzVMe2w4Yi7GagadsyB gitpod` (to connect to the database)

Next, you'll be able to test this:
- Start a new workspace while on the free plan
  - in the database, the workspace instance should have `workspaceClass: 'default'`
- Grant yourself access to XL:
  - create a new team (e.g. by visiting `/teams/new`)
  - upgrade the team to Unleashed (in team settings > billing)
  - in the database, update `d_b_team_subscription2` `excludeFromMoreResources` to `0`
- Start the same workspace again, or a new workspace
  - this workspace should have `workspaceClass = 'gitpodio-internal-xl'` in the DB

Happy to help with any of this, and/or have a call. 😁

atduarte (commented):
Thanks for the thoughtful comments and guidance, @jankeromnes! 🧡 

I'm currently stuck at "upgrade the team to Unleashed (in team settings > billing)" as I'm required to pay. 🤔 Looked for ways on Slack and Notion without any luck.

jankeromnes:
Sure, more than happy to help, especially since you spontaneously started implementing this yourself. 💯 Many thanks for that.

> "Payment in Progress". I guess you need to activate the Chargebee webhook now?

Ah, yes indeed, forgot to set this up. I went into Chargebee and added a webhook for this PR, so payments should go through now. (Also, didn't share the exact steps, because we're considering moving away from Chargebee soon, but if you're curious I'm still happy to share more anyway)

> Regarding the enum, I understand your concerns but I'm not sure using an enum is a good idea. It will depend on the conversation regarding how server will learn about the available classes in the workspace-clusters—it could be dynamic, and we will surely go further than the S-XL 🤔
> 
> Given this, do you still have a strong opinion on using enums? Happy to do what you think is best :)

Hmm, that's a good point. I still think we'll need some kind of fixed mapping between particular classes and credits/prices, but maybe that could also be dynamic?

Since this is still a bit unclear and somewhat orthogonal to this contribution, I have no strong feelings about enum or not -- please feel free to go ahead with whatever is most convenient for this PR, and let's figure out the mapping in a future iteration. 👍

atduarte:
image

😄

components/gitpod-db/src/typeorm/entity/db-workspace-instance.ts
atduarte: As discussed offline, the main issue is that we would need to keep this enum code in sync with whatever is available in the accompanying workspace-cluster, that in self-hosted might have arbitrarily named workspace classes.

We agreed for now to keep this as is, schedule a meeting to discuss how will WebApp's dashboard learn about the available classes https://github.com/gitpod-io/gitpod/issues/8261, and reevaluate.

easyCZ: For more context, and also relevant context for usage/billing, we'll need a service to list available workspace types - this needs to differ per installation (or also possibly by a cohort/project which has access to a particular type). It makes perfect sense to store and define as string because it will vary beyond a fixed set of values.

easyCZ:
LGTM. I've 2 questions remaining:
1. Are we guaranteed to receive the workspace class from the cluster manager? As in, has this been completely rolled out?
2. Given the code, it's currently possible to end up with 1 Workspace which had multiple Workspace Instances where each instance was of a different class. Is this acceptable?

easyCZ:
LGTM. I've 2 questions remaining:
1. Are we guaranteed to receive the workspace class from the cluster manager? As in, has this been completely rolled out?
2. Given the code, it's currently possible to end up with 1 Workspace which had multiple Workspace Instances where each instance was of a different class. Is this acceptable?

geropl:
> Given the code, it's currently possible to end up with 1 Workspace which had multiple Workspace Instances where each instance was of a different class. Is this acceptable?

From my point of view: yes, absolutely. I imagine a extreme case where you start up a workspace in the past, and try to re-start after 1 month. However the config change, the new instance should be guided by that, and not what was configured in the past.

atduarte:
> Are we guaranteed to receive the workspace class from the cluster manager? As in, has this been completely rolled out?

@easyCZ The class is chosen by the server and currently is actually hard-coded here: https://github.com/gitpod-io/gitpod/blob/6a8034257a023aa4e34886b08a29a44725dec200/components/server/src/workspace/workspace-starter.ts#L1298

easyCZ (approved):

PR #10071, opened 2022-05-17
Laushinka/error form 10062

laushinka:
## Description
Shows feedback component when there are error messages on:
- login
- create workspace

| MINIMISED | EXPANDED | THANKS |
| - | - | - |
| Screenshot 2022-05-24 at 13 15 05 | Screenshot 2022-05-24 at 13 06 15 |  Screenshot 2022-05-24 at 13 19 55 |

## Related Issue(s)

Fixes #10062

## How to test
(A bit tedious to make sure to get the error message, sorry)
1. Go to your user in the /admin page and remove the `repo` feature that gives permission for private repos.
2. Start a workspace for that private repo, i.e. `https://laushinka-f9a883104c.preview.gitpod-dev.com/#[YOUR_PRIVATE_REPO]`
3. See the feedback form below the error message.

## Release Notes

```release-note
Feedback form under error messages on login and starting workspaces.
```

## Documentation

/werft analytics=segment|TEZnsG4QbLSxLfHfNieLYGF4cDwyFWoe

PR #10455, opened 2022-06-03
[analytics] add error field to feedback event

### LINKED ISSUE #10402: Send error message with feedback
Right now the feedback_submitted event sends the user's feedback, but we have no visibility into the actual error.
We should send the error message/object along with the event.

_@jldec Sep 13, 2022_
See #10455 - The error message text needs to be separated from the rendered error html page, and included in the feedback.

laushinka:
## Description
_[@jldec updated Aug 17, 2022]_  
The intent of this PR was to add the rendered error message along with the feedback tracking event, however, the object used in this implementation does not actually include the rendered error message, because that is derived in JSX.

| With error | Without error |
| - | - |
| Screenshot 2022-06-03 at 14 18 12 | Screenshot 2022-06-03 at 15 26 14 |

## Related Issue(s)

Fixes #10402 

## How to test
1. Trigger an error when creating a workspace (steps [here](https://github.com/gitpod-io/gitpod/pull/10071#issue-1238938630)).
2. Visit [Segment](https://app.segment.com/gitpod/sources/staging_untrusted/debugger) and filter for `feedback_submitted`.
3. See the `error_rendered` body.
4. Send a non-error feedback and see no `error_rendered` field on Segment.

## Release Notes

```release-note
NONE
```
- [x] /werft with-preview

### DIFF
--- a/components/dashboard/src/Analytics.tsx
+++ b/components/dashboard/src/Analytics.tsx
@@ -9,6 +9,7 @@ import { log } from "@gitpod/gitpod-protocol/lib/util/logging";
 import Cookies from "js-cookie";
 import { v4 } from "uuid";
 import { Experiment } from "./experiments";
+import { StartWorkspaceError } from "./start/StartPage";
 
 export type Event =
     | "invite_url_requested"
@@ -40,6 +41,8 @@ export interface TrackFeedback {
     feedback: string;
     href: string;
     path: string;
+    error_object?: StartWorkspaceError;
+    error_message?: string;
 }
 interface TrackDashboardClick {
     dnt?: boolean;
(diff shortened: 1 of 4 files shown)


jldec:
I think we should try to include the actual error text in the feedback report

Information about the context would be good too e.g. to tell the difference between 404 errors in say
a. starting a workspace from a repo URL context
vs. 
b. starting a workspace from a snapshot URL

jldec:
@laushinka I rebased and updated the werft flags in the description

jldec:
This implementation works, but it is incomplete because the error message text is not included with the feedback.
We either have to grab an image, or (better), extract the error message text as a string. 

Currently, error handling implementations e.g. in [CreateWorkspace.tsx](https://github.com/gitpod-io/gitpod/blob/main/components/dashboard/src/start/CreateWorkspace.tsx) render errors directly to HTML, instead of starting with plaintext first, and then rendering it.

laushinka:
Closing because too stale. Picked up [here](https://github.com/gitpod-io/gitpod/pull/12992).

PR #10480, opened 2022-06-06
[usage] Automatically cleanup records between test DB connections

easyCZ:
## Description

Add hooks to test DB connections to clean up records after tests run.

## How to test

* Unit tests (run on CI)

## Release Notes

```release-note
NONE
```

## Documentation

NONE

### DIFF
--- a/components/usage/pkg/db/conn.go
+++ b/components/usage/pkg/db/conn.go
@@ -72,12 +72,26 @@ func ConnectForTests(t *testing.T) *gorm.DB {
 	})
 	require.NoError(t, err, "Failed to establish connection to DB. In a workspace, run `leeway build components/usage:init-testdb` once to bootstrap the DB.")
 
+	conn = conn.Debug()
+
 	t.Cleanup(func() {
+		// Delete records for known models from the DB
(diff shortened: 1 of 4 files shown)


components/usage/pkg/db/conn.go
```diff
@@ -54,12 +55,25 @@ func ConnectForTests(t *testing.T) *gorm.DB {
...
 
+	conn = conn.Debug()
+
 	t.Cleanup(func() {
+		// Delete records for known models from the DB
+		log.Info("Cleaning up DB between connections.")
+		for _, model := range []interface{}{
+			&WorkspaceInstance{},
+			&Workspace{},
+			&Project{},
```
andrew-farries, line 66: Do you want to pass in a struct of the type to be deleted to the `ConnectForTests` method, rather than iterating over known types here? This could be overlooked when adding new table.

andrew-farries (changes requested):

components/usage/pkg/db/conn.go
easyCZ: As in hoist these structs into a variable?

andrew-farries (approved):

PR #10485, opened 2022-06-06
Add self-hosted upgrade tests

nandajavarma:
## Description

This PR adds a new werft flag to test the upgrade of gitpod self-hosted from one version to the latest. This is intended to be used in a release PR, to see how the upgrade would work from the previous version(we can specify the version) to the latest. The setup uses the `beta` channel of replicated, since this is where the release ends up before going to the stable channel according to our workflow. To run the test you can add the following comment to the PR:

```
/werft run no-preview with-upgrade-tests fromVersion=2022.4.2
```
The above comment will creates a GKE cluster, install `20224.2` version of Gitpod, makes sure it is up and running, upgrades the Gitpod version to latest and confirm it has the basic functioning. This is the first :skateboard: , we will add the integration tests and support for different cloud providers as the second iteration.

Apologies for the lint fixes.

## Related Issue(s)

Fixes #

## How to test

To test this functionality you can open a workspace from this PR and run the following command once you make sure you have `core-dev` access:

```sh
werft run github -j .werft/build.yaml -a no-preview=true -a with-upgrade-tests=true -a fromVersion="2022.5.0"
```

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.werft/installer-tests.ts
+++ b/.werft/installer-tests.ts
@@ -4,7 +4,10 @@ import { Werft } from "./util/werft";
 
 const testConfig: string = process.argv.length > 2 ? process.argv[2] : "STANDARD_K3S_TEST";
 // we can provide the version of the gitpod to install (eg: 2022.4.2)
-const version: string = process.argv.length > 3 ? process.argv[3] : "";
+// "-" is the default value which will install the latest version
+const version: string = process.argv.length > 3 ? process.argv[3] : "-";
+
+const channel: string = process.argv.length > 4 ? process.argv[4] : "unstable";
 
 const makefilePath: string = join("install/tests");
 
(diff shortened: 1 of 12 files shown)


roboquat:

<details>

</details>

nandajavarma:
/werft run --no-preview --with-upgrade-tests --from-version="2022.4.2"

nandajavarma:
/werft run --no-preview --with-upgrade-tests --from-version "2022.4.2"

nandajavarma:
/werft run no-preview with-upgrade-tests from-version="2022.4.2"

.werft/installer-tests.ts
```diff
@@ -144,8 +174,9 @@ function getKubeconfig() {
 export async function installerTests(config: TestConfig) {
     console.log(config.DESCRIPTION);
     for (let phase of config.PHASES) {
-        const phaseSteps = INFRA_PHASES[phase];
-        const ret = callMakeTargets(phaseSteps.phase, phaseSteps.description, phaseSteps.makeTarget);
+        const args = phase.split(" ");
```
adrienthebo, line 177: I don't understand which phase has spaces in the name and needs to be split. Can you point me to the phase that's depending on this logic?

.werft/installer-tests.ts
```diff
@@ -144,8 +174,9 @@ function getKubeconfig() {
...
+        const phaseSteps = INFRA_PHASES[args[0]];
+        const ret = callMakeTargets(phaseSteps.phase, phaseSteps.description, phaseSteps.makeTarget, args.slice(1));
```
adrienthebo, line 179: The way that we're splitting the phase to generate args, and then dropping the first element of the args is a little bit confusing. As we grow these tests it might be useful to extract an `args` list for phases that need arguments.

No need to change, but raising this as a potential point of friction in a few weeks/months.

.werft/installer-tests.ts
```diff
@@ -154,8 +185,8 @@ export async function installerTests(config: TestConfig) {
     }
 }
 
-function callMakeTargets(phase: string, description: string, makeTarget: string) {
-    werft.phase(phase, description);
+function callMakeTargets(phase: string, description: string, makeTarget: string, args: string[]) {
```
adrienthebo, line 188: My understanding is that `args` are somehow passed to the `make` invocation, but I'm not seeing that in this code. Am I missing something?

install/tests/Makefile
```diff
@@ -92,6 +97,10 @@ check-gitpod-installation: delete-cm-setup check-kots-app check-env-sub-domain
 run-tests:
 	./tests.sh ${KUBECONFIG}
 
+kots-uprgade:
```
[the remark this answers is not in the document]
nandajavarma: Should I be concerned that it took me a solid two minutes to understand what the change here was? :eye:

.werft/installer-tests.ts
nandajavarma: This is a good question. This is my bad. Actually along with the upgrade I ended up adding some of the changes necessary for supporting different combinations of configurations in this one. So the thing is, depending on the `TEST_CONFIGURATION`, we can have different combinations of dependencies. like GKE with GCR or EKS with s3 etc. These things matter only when create a kots config file because that's where we specify all these connections. So I thought, I can pass args like "storage=s3" to the make target for `generate-kots-config`. But the problem is since these depend on the platform, I had to go back one more step and provide these args in `PHASES`. There are no phases in this example that does it, I just wanted to add the setup for it. Going one step further, for GKE the storage can be either `s3` or `in-cluster` and this can be chosen randomly so eventually we test all combinations. So I decided to do something like `storage=sample["s3, "in-cluster"]`. This randomly chooses one of the two storage options. Again, I know this is not the cleanest of solutions, but I wanted to have a working solution in place.

install/tests/Makefile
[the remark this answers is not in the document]
adrienthebo: Not even a little bit! It's hard to catch typos in code that you've written, fresh eyes help. (That and I have an overactive spell checker in my head. I don't mean to pick on small issues here, only catch potential areas that could trip us up later.)

.werft/installer-tests.ts
nandajavarma: I have removed this `args` usage. You are right, ultimately super hacky!

.werft/installer-tests.ts
nandajavarma: Updated this!

.werft/installer-tests.ts
```diff
@@ -170,6 +201,10 @@ function callMakeTargets(phase: string, description: string, makeTarget: string)
     return response.code;
 }
 
+function sample(stages: string[]): string {
```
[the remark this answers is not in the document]
nandajavarma: I have updated its usage

PR #10488, opened 2022-06-06
[previews] Expose observability components through stable DNS

ArthurSens:
## Description

The developer experience when working with metrics is not great and leads to metrics breaking in production every now and then.

This PR improves the DevX a little bit by facilitating access to the Observability components through stable domain names.

## TODO before merging

This PR only implements the routing to Harvester/Core-dev ingresses, we still need to configure the ingresses to route traffic to the correct components/previews

## Related Issue(s)

Fixes https://github.com/gitpod-io/ops/issues/881

## How to test

Access werft results

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.werft/util/certs.ts
+++ b/.werft/util/certs.ts
@@ -10,9 +10,7 @@ export class IssueCertificateParams {
     gcpSaPath: string
     dnsZoneDomain: string
     domain: string
-    ip: string
     additionalSubdomains: string[]
-    bucketPrefixTail: string
     certName: string
     certNamespace: string
     withVM: boolean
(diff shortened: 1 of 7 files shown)


.werft/util/certs.ts
```diff
@@ -10,9 +10,7 @@ export class IssueCertificateParams {
     gcpSaPath: string
     dnsZoneDomain: string
     domain: string
-    ip: string
     additionalSubdomains: string[]
-    bucketPrefixTail: string
```
ArthurSens, line 15: both aren't used anywhere... it would be nice to have a linter for this 🤔

PR #10490, opened 2022-06-06
[usage] Handle empty value varchar time, use ISO8601

easyCZ:
## Description

Hardens handling of Varchar Time records in the DB. Here, we ensure that empty value of VarChar is equivalent to the empty string and therefore not a valid Timestamp, but a valid value to store in the DB.

## Related Issue(s)

* Follow-up from conversation in https://github.com/gitpod-io/gitpod/pull/10293#discussion_r885595480

## How to test

Unit tests (run by CI)

## Release Notes

```release-note
NONE
```

## Documentation

NONE

### DIFF
--- a/components/usage/pkg/db/types.go
+++ b/components/usage/pkg/db/types.go
@@ -39,35 +39,34 @@ func (n *VarcharTime) Scan(value interface{}) error {
 
 	switch s := value.(type) {
 	case []uint8:
-		// Null value - empty string mean value is not set
-		if len(s) == 0 {
-			n.valid = false
-			return nil
-		}
-
-		parsed, err := iso8601.ParseString(string(s))
-		if err != nil {
-			return fmt.Errorf("failed to parse %v into ISO8601: %w", string(s), err)
-		}
-		n.valid = true
-		n.t = parsed.UTC()
-		return nil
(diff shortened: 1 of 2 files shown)


components/usage/pkg/db/types.go
```diff
@@ -80,9 +79,21 @@ func (n VarcharTime) IsSet() bool {
 
 // Value implements the driver Valuer interface.
 func (n VarcharTime) Value() (driver.Value, error) {
```
easyCZ, line 81: This is what will be inserted into SQL statements. Here we use the ISO8601 format to be consistent with existing records.

components/usage/pkg/db/types.go
```diff
@@ -80,9 +79,21 @@ func (n VarcharTime) IsSet() bool {
...
-	return n.t.UTC().Format(time.RFC3339Nano), nil
+	if n.IsSet() {
+		return TimeToISO8601(n.t), nil
+	}
+	return "", nil
 }
 
 func (n VarcharTime) String() string {
```
easyCZ, line 88: This is what is used to print this field for humans, when we're converting the struct to a string. We align with the DB value and also print ISO8601

andrew-farries (approved):

Issue #10315, opened 2022-05-29
[ws-manager] workspaces failed to start


atduarte:
### Bug description

Tried opening workspaces for a side-project and it failed to start at least two times. 

Two errors I didn't have time to read showed up in quick succession, and the workspace stopped before it actually started. When I clicked in the "Open Workspace" button that showed up in the "Stopped" page, I got the following error: "cannot initialize workspace: cannot initialize workspace: no backup found".

### Steps to reproduce

It was happening for every workspace started, but not anymore.

### Workspace affected

atduarte-revispt-960nm6i1khz, atduarte-revispt-jsh4ofkpnyz

kylos101:
Thanks for the heads up, we initially addressed [in generation 44](https://github.com/gitpod-io/gitpod/pull/9600). Therefore, there's still some work to do, scheduled.

sagor999:
I believe this is due to failure to make prior backup, like when ws-daemon crashes.
To make it easier to debug later on, here is PR: https://github.com/gitpod-io/gitpod/pull/10491
that makes sure that we will surface if there was an actual error when attempting to download backup.

atduarte:
@sagor999 the example workspace ids provided are from fresh starts (of a project with prebuilds) that failed and a few minutes later, trying again, started working.

sagor999:
`atduarte-revispt-960nm6i1khz` failed with this error: 
```
cannot pull image: rpc error: code = FailedPrecondition desc = failed to pull and unpack image "reg.ws-eu44xl.gitpod.io:20000/remote/8bef4e73-367a-41b5-94c6-01c1f7d6c879:latest": failed commit on ref "manifest-sha256:844f36297db5efe4e646d7767ef139628f11ca3b0a0a2943b8b7df6226ae3295": "manifest-sha256:844f36297db5efe4e646d7767ef139628f11ca3b0a0a2943b8b7df6226ae3295" failed size validation: 10409 != 10331: failed precondition
```
which was due to redis\ipfs\cache issues that we saw before.

`atduarte-revispt-jsh4ofkpnyz` this one did fail with no backup found. Was this WS opened from previous WS that was failed due to that error above? I think in some cases we do not handle the case of failed WS properly and attempt to download backup from it.

sagor999:
@atduarte moving back into "scheduled" for now. Feel free to close this as well, as I think it might have been resolved since then.

atduarte:
@sagor999 I'm sorry I missed this.

[...]

PR #10339, opened 2022-05-30
[ws-daemon] Use a metric for backup waiting time instead logs

aledbf:
## How to test

- Open multiple workspaces
- Wait some minutes and close it all
- Check the metric returns values
```
gitpod_ws_daemon_concurrent_backup_waiting_seconds_bucket{le="10"} 5
gitpod_ws_daemon_concurrent_backup_waiting_seconds_bucket{le="30"} 5
gitpod_ws_daemon_concurrent_backup_waiting_seconds_bucket{le="60"} 6
gitpod_ws_daemon_concurrent_backup_waiting_seconds_bucket{le="120"} 6
gitpod_ws_daemon_concurrent_backup_waiting_seconds_bucket{le="180"} 6
gitpod_ws_daemon_concurrent_backup_waiting_seconds_bucket{le="300"} 6
gitpod_ws_daemon_concurrent_backup_waiting_seconds_bucket{le="600"} 6
gitpod_ws_daemon_concurrent_backup_waiting_seconds_bucket{le="1800"} 6
gitpod_ws_daemon_concurrent_backup_waiting_seconds_bucket{le="+Inf"} 6
```

## Release Notes
```release-note
[ws-daemon] Use a metric for backup waiting time instead of logs
```

PR #10491, opened 2022-06-06
[content-service] show error if failed to download backup file

sagor999:
## Description

Make sure to show if there was an error attempting to download backup, if any.

## Related Issue(s)

Related https://github.com/gitpod-io/gitpod/issues/10315

## Release Notes

```release-note
[content-service] make sure to show error if there was one when attempting to download backup
```

### DIFF
--- a/components/content-service/pkg/initializer/initializer.go
+++ b/components/content-service/pkg/initializer/initializer.go
@@ -191,7 +191,7 @@ func (bi *fromBackupInitializer) Run(ctx context.Context, mappings []archive.IDM
 
 	hasBackup, err := bi.RemoteStorage.Download(ctx, bi.Location, storage.DefaultBackup, mappings)
 	if !hasBackup {
-		return src, xerrors.Errorf("no backup found")
+		return src, xerrors.Errorf("no backup found: %w", err)
 	}
 	if err != nil {
 		return src, xerrors.Errorf("cannot restore backup: %w", err)


components/content-service/pkg/initializer/initializer.go
```diff
@@ -191,10 +191,10 @@ func (bi *fromBackupInitializer) Run(ctx context.Context, mappings []archive.IDM
 
 	hasBackup, err := bi.RemoteStorage.Download(ctx, bi.Location, storage.DefaultBackup, mappings)
 	if !hasBackup {
-		return src, xerrors.Errorf("no backup found")
+		return src, fmt.Errorf("no backup found: %v", err)
 	}
 	if err != nil {
-		return src, xerrors.Errorf("cannot restore backup: %w", err)
```
aledbf, line 197: why the `xerrors` change?

utam0k: If we use `xerrors`, we get a warning that we have to use `fmt` instead of it.

aledbf: @utam0k I made a similar change and got this comment https://github.com/gitpod-io/gitpod/pull/10339#discussion_r884669484

https://github.com/gitpod-io/go-style-guide/blob/master/style.md#errors

utam0k: @aledbf Thanks for your information. I haven't seen it. You are right.

sagor999: Hm. I was under the impression that we will be slowly switching to fmt instead of xerrors.
@easyCZ I think it was in one of your PRs that this was discussed, unless I am mistaken.

aledbf: > Hm. I was under the impression that we will be slowly switching to fmt instead of errors.

Same here. Maybe the guide is not clear about that?

sagor999: @csweichel ^

kylos101: [A related Slack thread](https://gitpod.slack.com/archives/C01KGM9AW4W/p1651840589492279)

csweichel: > Same here. Maybe the guide is not clear about that?

The guide is clear IMHO: it promotes `xerrors` (on purpose).


Thanks @kylos101 for linking the Slack thread. We didn't really land on a change, because
- `fmt.Errorf` neither carries, nor perpetuates a stack trace
- `github.com/pkg/errors` would produce a stack trace, but is deprecated and yet another third party library

xerrors is not ideal, but 
- it works
- it is reasonably consistent throughout our codebase. Until we decide to do something else we should keep it this way to ease migration "search + replace" style.
- it's not deprecated
- it comes from the Go team themselves.

All that said, `xerrors` also isn't exactly future proof, and we could look for alternatives **in an RFC**.

sagor999: updated PR.

aledbf (approved):

csweichel (approved):

PR #10495, opened 2022-06-07
[usage] List workspaces for each workspace instance in usage period

easyCZ:
## Description

We first compute all workspace instances in a given period, for each WorkspaceID, we fetch the underlying Workspace from DB to have the necessary information to proceed to determining the owner (cost center)

## Related Issue(s)

Fixes #

## How to test

Unit tests

## Release Notes

```release-note
NONE
```

## Documentation

NONE

### DIFF
--- a/components/usage/pkg/controller/reconciler.go
+++ b/components/usage/pkg/controller/reconciler.go
@@ -6,6 +6,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"github.com/gitpod-io/gitpod/common-go/log"
 	"github.com/gitpod-io/gitpod/usage/pkg/db"
@@ -38,6 +39,8 @@ type UsageReconcileStatus struct {
 
 	WorkspaceInstances        int
 	InvalidWorkspaceInstances int
+
+	Workspaces int
 }
 
 func (u *UsageReconciler) Reconcile() error {
@@ -71,9 +74,57 @@ func (u *UsageReconciler) ReconcileTimeRange(ctx context.Context, from, to time.
 	if len(invalidInstances) > 0 {
 		log.WithField("invalid_workspace_instances", invalidInstances).Errorf("Detected %d invalid instances. These will be skipped in the current run.", len(invalidInstances))
 	}
-
 	log.WithField("workspace_instances", instances).Debug("Successfully loaded workspace instances.")
+
+	workspaces, err := u.loadWorkspaces(ctx, instances)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load workspaces for workspace instances in time range: %w", err)
+	}
+	status.Workspaces = len(workspaces)
+
 	return status, nil
+}
+
+type workspaceWithInstances struct {
+	Workspace db.Workspace
+	Instances []db.WorkspaceInstance
+}
+
+func (u *UsageReconciler) loadWorkspaces(ctx context.Context, instances []db.WorkspaceInstance) ([]workspaceWithInstances, error) {
+	var workspaceIDs []string
+	for _, instance := range instances {
+		workspaceIDs = append(workspaceIDs, instance.WorkspaceID)
+	}
+
+	workspaces, err := db.ListWorkspacesByID(ctx, u.conn, toSet(workspaceIDs))
+	if err != nil {
+		return nil, fmt.Errorf("failed to find workspaces for provided workspace instances: %w", err)
+	}
+
+	// Map workspaces to corresponding instances
+	workspacesWithInstancesByID := map[string]workspaceWithInstances{}
+	for _, workspace := range workspaces {
+		workspacesWithInstancesByID[workspace.ID] = workspaceWithInstances{
+			Workspace: workspace,
+		}
+	}
+
+	// We need to also add the instances to corresponding records, a single workspace can have multiple instances
+	for _, instance := range instances {
+		item, ok := workspacesWithInstancesByID[instance.WorkspaceID]
+		if !ok {
+			return nil, errors.New("encountered instance without a corresponding workspace record")
+		}
+		item.Instances = append(item.Instances, instance)
+	}
+
+	// Flatten results into a list
+	var workspacesWithInstances []workspaceWithInstances
+	for _, w := range workspacesWithInstancesByID {
+		workspacesWithInstances = append(workspacesWithInstances, w)
+	}
+
+	return workspacesWithInstances, nil
 }
 
 func (u *UsageReconciler) loadWorkspaceInstances(ctx context.Context, from, to time.Time) ([]db.WorkspaceInstance, []invalidWorkspaceInstance, error) {
@@ -132,8 +183,8 @@ func trimStartStopTime(instances []db.WorkspaceInstance, maximumStart, minimumSt
 	var updated []db.WorkspaceInstance
 
 	for _, instance := range instances {
-		if instance.StartedTime.Time().Before(maximumStart) {
-			instance.StartedTime = db.NewVarcharTime(maximumStart)
+		if instance.CreationTime.Time().Before(maximumStart) {
+			instance.CreationTime = db.NewVarcharTime(maximumStart)
 		}
 
 		if instance.StoppedTime.Time().After(minimumStop) {
@@ -143,4 +194,17 @@ func trimStartStopTime(instances []db.WorkspaceInstance, maximumStart, minimumSt
 		updated = append(updated, instance)
 	}
 	return updated
+}
+
+func toSet(items []string) []string {
+	m := map[string]struct{}{}
+	for _, i := range items {
+		m[i] = struct{}{}
+	}
+
+	var result []string
+	for s := range m {
+		result = append(result, s)
+	}
+	return result
 }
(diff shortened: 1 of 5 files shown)


roboquat:

<details>

</details>

components/usage/pkg/controller/reconciler.go
```diff
@@ -71,11 +74,59 @@ func (u *UsageReconciler) ReconcileTimeRange(ctx context.Context, from, to time.
...
+	for _, w := range workspacesWithInstancesByID {
+		workspacesWithInstances = append(workspacesWithInstances, w)
+	}
+
+	return workspacesWithInstances, nil
+}
+
 func (u *UsageReconciler) loadWorkspaceInstances(ctx context.Context, from, to time.Time) ([]db.WorkspaceInstance, []invalidWorkspaceInstance, error) {
 	log.Infof("Gathering usage data from %s to %s", from, to)
 	instances, err := db.ListWorkspaceInstancesInRange(ctx, u.conn, from, to)
```
jankeromnes, line 132: Reading the code below, I see that we ignore currently running instances (no stoppedTime).

But what if an instance is kept running for 30 hours, when the user has only 2h of usage left?

Wouldn't it be better to set an "artificial" stoppedTime to "right now" on the provisional usage record, and thus ensure we also count hours for currently running workspaces?

easyCZ: > Reading the code below, I see that we ignore currently running instances (no stoppedTime).

Where do you see the logic for this? We list all, including currently running instances, validate that the records are sensible (start is before end, if set, and there is a start time) and then split the set into valid and invalid.

components/usage/pkg/controller/reconciler.go
```diff
@@ -71,11 +74,59 @@ func (u *UsageReconciler) ReconcileTimeRange(ctx context.Context, from, to time.
...
 	return status, nil
 }
 
+type workspaceWithInstances struct {
+	Workspace db.Workspace
+	Instances []db.WorkspaceInstance
+}
+
+func (u *UsageReconciler) loadWorkspaces(ctx context.Context, instances []db.WorkspaceInstance) ([]workspaceWithInstances, error) {
+	var workspaceIDs []string
```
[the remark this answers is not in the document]
easyCZ: fixed

components/usage/pkg/controller/reconciler.go
jankeromnes: The relevant code:

https://github.com/gitpod-io/gitpod/blob/3c63e7eb6bb3570d4dc5debf1c17938d6a0a3591/components/usage/pkg/controller/reconciler.go#L113-L123

My point: Still running instances are valid. They should create a usage record going from `creationTime` to `getCurrentTime()`, thus giving a correct idea of how many seconds have been used so far (even if the instance hasn't been stopped yet).

components/usage/pkg/controller/reconciler.go
```diff
@@ -71,11 +74,59 @@ func (u *UsageReconciler) ReconcileTimeRange(ctx context.Context, from, to time.
...
+	}
+	status.Workspaces = len(workspaces)
+
 	return status, nil
 }
 
+type workspaceWithInstances struct {
+	Workspace db.Workspace
+	Instances []db.WorkspaceInstance
+}
```
jankeromnes, line 91: **Thought**: I wonder if we actually want to load into memory the "fully-hydrated" shapes of both `Workspace` and `WorkspaceInstance`.

In the past, we've had issues with memory consumption, because we were keeping full shapes in memory, while only a few fields were actually relevant. This is especially true for workspace instances, because there are so many of them.

A usage record will probably look something like this:

```ts
export type WorkspaceUsageRecord = {
    instanceId: string;
    workspaceId: string;
    userId: string;
    projectId?: string;
    teamId?: string;
    fromTime: string;
    toTime: string;
};
```

So, in my mind, the most efficient flow to get these usage records is:
1. Query only `instanceId`, `workspaceId`, `creationTime`, `stoppedTime` from `d_b_workspace_instance` (for the current period) -- all the other fields are unnecessary
2. Trim the time bounds between `beginningOfMonth` and `min(getCurrentTime(), endOfMonth)` (also when there is no `stoppedTime` yet)
3. For each unique `workspaceId` query only the owner `userId` and possible `projectId` from `d_b_workspace`
4. For each `userId`, query only the possible `teamId` (randomly pick one team for now)

If we query and then load the full `Workspace` and `WorkspaceInstance` shapes into memory, and start passing them around a lot, I worry that we'll have a bad time (because many workspace instances are created every month, and that number is rapidly increasing over time).

easyCZ: Sure, this makes sense. But I don't think it makes sense to do this now.

For me, this is an optimisation once we have something working. For now, I'd prefer to waste memory to get to a working version and then measure what is in fact the cost of these queries.

components/usage/pkg/controller/reconciler.go
easyCZ: `if instance.StoppedTime.IsSet() {` will be false when the Stop time is not present (see comment on L113). So this skips instances which are currently running. We only mark it as invalid if it has a Stop time, and stop time is before the start time which is semantically invalid.

easyCZ: And I've also now added a unit tests to prove this is the case - https://github.com/gitpod-io/gitpod/pull/10495/files#diff-a27489be974eefc74571351b4369b7f8f37c870194369a0d6423362f5b56a67bR32

jankeromnes: Gah, sorry, I misread the comment and didn't fully read the code snippet I shared 🤦‍♂️ sorry for the noise, and thanks for adding a test!

components/usage/pkg/controller/reconciler.go
jankeromnes: Fair point, thanks. (I don't really see why it's better to ship a less optimized version first when we know how to easily write a more optimized one, but I also agree about shipping this sooner than later, and maybe my suggestion could already be considered premature optimization). So, fine to defer this to later. 👍

components/usage/pkg/controller/reconciler.go
```diff
@@ -71,11 +74,59 @@ func (u *UsageReconciler) ReconcileTimeRange(ctx context.Context, from, to time.
...
+
+func (u *UsageReconciler) loadWorkspaces(ctx context.Context, instances []db.WorkspaceInstance) ([]workspaceWithInstances, error) {
+	var workspaceIDs []string
+	for _, instance := range instances {
+		workspaceIDs = append(workspaceIDs, instance.WorkspaceID)
+	}
+
+	workspaces, err := db.ListWorkspacesByID(ctx, u.conn, toSet(workspaceIDs))
```
jankeromnes, line 99: Thanks for making the IDs unique! I would have preferred to make the collection a unique "set" _before_ all of the appends, in order to iterate just once over all instances (not once, then again in `toSet` to create the map, then another time over the unique IDs to convert them back to a list) -- but okay for quick iterations without worrying too much about complexity this early. 🛹

components/usage/pkg/controller/reconciler.go
```diff
@@ -71,11 +74,59 @@ func (u *UsageReconciler) ReconcileTimeRange(ctx context.Context, from, to time.
...
 		log.WithField("invalid_workspace_instances", invalidInstances).Errorf("Detected %d invalid instances. These will be skipped in the current run.", len(invalidInstances))
 	}
-
 	log.WithField("workspace_instances", instances).Debug("Successfully loaded workspace instances.")
+
+	workspaces, err := u.loadWorkspaces(ctx, instances)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load workspaces for workspace instances in time range: %w", err)
+	}
+	status.Workspaces = len(workspaces)
```
jankeromnes, line 83: So, at this point we have:
- a collection of all "fully hydrated" instances
- a collection of all "fully hydrated" workspaces, each with an attached sub-collection of all its corresponding "fully hydrated" instances (although these will just be references to a single instance in memory, right, not a full copy of the object?)

I'm still slightly uneasy with loading so many large objects into memory, and iterating over them multiple times. So, I'm looking forward to seeing exactly how we'll use these collections, and see if we can later reduce the amount of used memory and the number of iterations. 👀

components/usage/pkg/controller/reconciler.go
easyCZ: Largely because I think the current handling will go away anyway once we do add attribution to the WSI. If we optimized here, it would then be wasted work because it would also be replaced by the attribution work from WSI.

components/usage/pkg/controller/reconciler.go
easyCZ: That's okay. Code review is as much about learning how something works for the reviewer as it is to ensure we're doing the right thing.

jankeromnes (approved):
Many thanks @easyCZ!

With the unit tests, this seems to be working as intended. 🎉  I couldn't resist to add a few more thoughts on (premature?) optimization, mostly because I generally find it easier to design an efficient algorithm from scratch (minimal data footprint, limited iterations) than to optimize existing complex code, but you're right -- the current code might already go away when we attribute instances on start. Someone clever once said, "one of the most expensive mistakes engineers make is to optimize something that shouldn't exist in the first place".

Anyway, so many words to say, thanks for the progress here! Looking forward to the next iterations. 🚀

Issue #9891, opened 2022-05-10
Preview Envs on Harvester: Delete when inactive


meysholdt:
## Is your feature request related to a problem? Please describe
Our Harvester cluster keeps filling up close to its max capacity. We should delete preview environments when they're inactive, just as we do for core-dev. 

## Describe the behavior you'd like
Implement 
https://github.com/gitpod-io/gitpod/blob/539414d0e40d318de07abfc8b047f60a4f7969fc/.werft/platform-delete-preview-environments-cron.ts#L82-L87

By porting over the logic from core-dev:
https://github.com/gitpod-io/gitpod/blob/539414d0e40d318de07abfc8b047f60a4f7969fc/.werft/platform-delete-preview-environments-cron.ts#L139-L176

meysholdt:
It would be lice if the job would also log which db rows are the cause for considering an env active. This can help to understand if we're looking at the right signals. 

Example: Change 
https://github.com/gitpod-io/gitpod/blob/539414d0e40d318de07abfc8b047f60a4f7969fc/.werft/platform-delete-preview-environments-cron.ts#L174

to 
```
werft.log(sliceID, `${this.name} (${this.namespace}) - is-inactive=${isInactive}. Active DB-Rows: d_b_workspace_instance:${latestInstance.length}, d_b_user-created:${latestUser.length}, d_b_user-modified:${lastModified.length}, d_b_workspace_instance_user:${heartbeat.length}`);
```

meysholdt:
An important use-case for this feature: When a preview env gets created from a branch that only has old commits, it's being GCed immediately today. Implementing this feature would have prevented that.

Context: https://gitpod.slack.com/archives/C032A46PWR0/p1652431918110569?thread_ts=1652427026.649699&cid=C032A46PWR0

mads-hartmann:
> When a preview env gets created from a branch that only has old commits, it's being GCed immediately today. Implementing this feature would have prevented that.

Currently the heuristics in the job means that if either of the conditions is true, it will delete the preview environment. So in this case, the preview environment would still be deleted as the branch is stale.

I think it's a good idea to change that though, e.g. so that if there's any DB activity we don't delete the preview environment.

mads-hartmann:
@vulkoingim I'm reopening this as there's still a tiny bit missing. Please see my comment here https://github.com/gitpod-io/gitpod/pull/10379#pullrequestreview-991711020

vulkoingim:
[...]

PR #10498, opened 2022-06-07
Always keep preview envs on db activity

vulkoingim:
## Description
Always keep preview environments when there is DB activity. This addresses an edge case where forks that have a running preview env would get delete due to not having a "real" branch in the repo.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/9891#issuecomment-1148318363

## Release Notes

```release-note
NONE
```


mads-hartmann (approved):
Thanks! 🛹

PR #10501, opened 2022-06-07
Delete preview env harvester certs

vulkoingim:
## Description

Harvester certs have the following format: `harvester-PREVIEW_ENV_NAME`, therefore they were never getting cleaned up. This should make sure that also those certs gets deleted properly. I'll do a manual clean-up of the ones that exist but have no corresponding preview environment..

## Related Issue(s)

Fixes https://github.com/gitpod-io/ops/issues/2483

## Release Notes

```release-note
NONE
```


ArthurSens (approved):
Nice, very simple fix 🙂

PR #10507, opened 2022-06-07
feat: Clean out terminal buffer before running tasks commands

axonasif:
## Description

It gets kinda confusing to look at what's happening on a `tasks` terminal as gitpod spawns a bunch of vscode terminals with task shell commands.

For comparison, please open [this](https://gitpod.io/#https://github.com/gitpod-io/template-flutter/tree/test-demo) and [this](https://gitpod.io/#https://github.com/gitpod-io/template-flutter) to see. (I've been manually including the `printf` command to clean out my task terminals on `template-flutter` repo to make it intuitive and less confusing)

This change will run ` printf '\e[3J\033c\e[3J'`([ref](https://en.wikipedia.org/wiki/ANSI_escape_code#CSI_(Control_Sequence_Introducer)_sequences)) before task shell commands and thus cleaning out the terminal entirely.

| Previously  | Now |
| ------------- | ------------- |
| Screenshot 2022-06-08 at 1 20 53 AM|Screenshot 2022-06-08 at 1 19 25 AM |

## Related Issue(s)

Fixes #

## How to test

See description 🙏 

## Release Notes

```release-note
feat: Clean the excess scrollback buffer of task terminals
```

### DIFF
--- a/components/supervisor/pkg/supervisor/tasks.go
+++ b/components/supervisor/pkg/supervisor/tasks.go
@@ -378,9 +378,10 @@ func getHistfileCommand(task *task, commands []*string, contentSource csapi.Work
 		return ""
 	}
 
-	// the space at beginning of the HISTFILE command prevents the HISTFILE command itself from appearing in
+	// the printf command clears out the terminal buffer
+	// the space at beginning of the printf command prevents our injected commands itself from appearing in
 	// the bash history.
-	return " HISTFILE=" + histfile + " history -r"
+	return " printf '\\e[3J\033c\\e[3J'" + " && " + "HISTFILE=" + histfile + " history -r"
 }
 
 func getCommands(task *task, isHeadless bool, contentSource csapi.WorkspaceInitSource, storeLocation string) []*string {
--- a/components/supervisor/pkg/supervisor/tasks.go
+++ b/components/supervisor/pkg/supervisor/tasks.go
@@ -378,10 +378,10 @@ func getHistfileCommand(task *task, commands []*string, contentSource csapi.Work
 		return ""
 	}
 
-	// the printf command clears out the terminal buffer
+	// the printf command clears out the excess terminal scrollback buffer
 	// the space at beginning of the printf command prevents our injected commands itself from appearing in
 	// the bash history.
-	return " printf '\\e[3J\033c\\e[3J'" + " && " + "HISTFILE=" + histfile + " history -r"
+	return " printf '\033[3J\033c\033[3J'" + " && " + "HISTFILE=" + histfile + " history -r"
 }
 
 func getCommands(task *task, isHeadless bool, contentSource csapi.WorkspaceInitSource, storeLocation string) []*string {


andreafalzetti:
I quite like this idea.

I wonder if we could keep it easy and use `clear` or `reset`? Did you consider those?

Could you also help breaking down the command outputs? For example, `printf "\033c"` is already clearing the terminal, so I don't know what the other bits do, e.g. `e[3J`?

axonasif:
Hey @andreafalzetti, thanks for sharing your concerns 🧡 

> I wonder if we could keep it easy and use `clear` or `reset`? Did you consider those?

Yes! `clear` and `reset` basically sends(prints) a bunch of sequences, most of which we won't need here and they're slow to run since they're external commands unlike `printf`/`echo`, these commands just reset your visible buffer, the rest still stays up there.

> Could you also help breaking down the command outputs? For example, `printf "\033c"` is already clearing the terminal, so I don't know what the other bits do, e.g. `e[3J`?

That's true 🎯 I was also going to just propose it only, however I noticed that it wasn't always working for all circumstances (e.g tmux and some other scrollback buffers with special chars in them which I don't remember now 😅), so I had to try out something that I found to be working almost everywhere. The tricky part is, you have to print this twice and keep the `\033c` sequence in the middle. For ref, see the `Erase in Display`(`\033[3J`) part in [here](https://en.wikipedia.org/wiki/ANSI_escape_code#CSI_(Control_Sequence_Introducer)_sequences).

Also, here's a video to show happening 🙏 
https://www.loom.com/share/f29dcbde9d744d6e8f18a114809b77b5

andreafalzetti:
Thanks for the clarification @axonasif 🙏 

Sharing a few more concerns:

- Does this prevent looking at prebuild logs?
- Does the final message of the prebuild get displayed? Example
   ```
    🤙 This task ran as a workspace prebuild
    🎉 Well done on saving 5 minutes
```

felladrin:
That's great, @axonasif!

And I'm also interested in the answers for @andreafalzetti's questions above :)

axonasif:
Hey @andreafalzetti @felladrin 👋 , it shouldn't since it's only being called in one shell invocation for the commands block we're sending to the created terminals. It's basically being put before the `HISTFILE` keyword as you see in the code of this PR and in the screenshot(not visible in this old screenshot) below:
Screenshot 2022-06-11 at 12 15 50 AM

mustard-mh (commented):
I prefer to get those content, that makes `why this command/output happen` clear. 

And same with @andreafalzetti 's question, prebuild log will reset too

axonasif:
Not stale Mr. bot 😂  , I'll look into it soon 🙏

felladrin:
@axonasif, are you planning to change how it works? Should we set it as a Draft meanwhile?

axonasif:
> @axonasif, are you planning to change how it works? Should we set it as a Draft meanwhile?

Yes, thanks for asking. I am OOO so didn't post an update yet. I think we can close this PR. I will raise an issue describing an idea of introducing a `gp` CLI subcommand for completely cleaning up the terminal session scrollback buffer, this is in regard to @mustard-mh's comment above.

felladrin:
> Yes, thanks for asking. I am OOO so didn't post an update yet. I think we can close this PR. I will raise an issue describing an idea of introducing a `gp` CLI subcommand for completely cleaning up the terminal session scrollback buffer, this is in regard to @mustard-mh's comment above.

Thank you! Willing to see your take on the `gp` command!

PR #10522, opened 2022-06-08
Revert "Update gitpod dev image"

mads-hartmann:
Reverts gitpod-io/gitpod#10509

The build and workspaces are failing with `protoc-gen-js: program not found or is not executable`. There are two reasons the build on the original PR might not have caught this

- We have job protection enabled, so by default jobs triggered by Werft uses the job specs from `main`
- The build didn't rebuild all components. Notably `components/supervisor-api/typescript-grpc:lib` and `components/supervisor-api/typescript-grpcweb:lib` weren't build as part of [the job](https://werft.gitpod-dev.com/job/gitpod-build-aledbf-go183.5/logs) and they are the ones failing on main.

To get a successful build I used the following command to make sure that Werft runs using the changes off my branch and that it updates the commit check as part of the job. Using `no-test` and `no-preview` just to speed up the job.

```sh
werft job run github -a updateGitHubStatus=gitpod-io/gitpod -a no-test=true -a no-preview=true 
```

**NOTE**: I don't know how to trigger Leeway to build all components and my build doesn't contain the components that break on main. But when I open a workspace off this branch the build works, and it doesn't on main so I think we should merge this. The following two commands works in a workspace off this branch but not on main:

```
leeway build components/supervisor-api/typescript-grpcweb:lib
leeway build leeway build components/supervisor-api/typescript-grpc:lib
```

## Release Notes

```release-note
NONE
```

## Documentation

N/A

PR #10509, opened 2022-06-07
Update gitpod dev image

aledbf:
## Description

- Trigger build to get the new workspace-full image with go v1.18.3
- Update k8s binaries to v1.24.1
- Update cert-manager to v1.8.0
- Update google-cloud-sdk to 388.0.0
- Update prometheus to v2.36.0

## Release Notes
```release-note
NONE
```

PR #10526, opened 2022-06-08
Make it possible to skip cache

mads-hartmann:
## Description

As part of https://github.com/gitpod-io/gitpod/pull/10509 and the revert of it here https://github.com/gitpod-io/gitpod/pull/10522 it would have been nice to be able to trigger a full rebuild to validate changes. E.g. this we could add instructions like

> If you're changing the dev image then run a build from your branch (by default we use the job spec from main due to job protection) and rebuild all components like so: `werft job run github -a no-cache=true -a updateGitHubStatus=gitpod-io/gitpod`

### Alternatives

This is still a very manual way to solve the issue and people might forget to run the job manually when changing the build job spec or the CI image.

- Could we have Leeway invalidate the build cache if the dev image changes?
- Could we have Werft not use job protection if the job was changed and the owner is a Gitpod employee?

## Related Issue(s)

No issue

## How to test

From a workspace on this branch:

```
werft job run github -a no-cache=true -a updateGitHubStatus=gitpod-io/gitpod
```

## Release Notes

```release-note
NONE
```

## Documentation

N/A

### DIFF
--- a/.werft/jobs/build/build-and-publish.ts
+++ b/.werft/jobs/build/build-and-publish.ts
@@ -16,13 +16,13 @@ export async function buildAndPublish(werft: Werft, jobConfig: JobConfig) {
     const {
         publishRelease,
         dontTest,
-        withContrib,
         retag,
         version,
         localAppVersion,
         publishToJBMarketplace,
         publishToNpm,
         coverageOutput,
+        noCache,
     } = jobConfig;
 
     const releaseBranch = jobConfig.repository.ref;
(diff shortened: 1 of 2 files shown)


csweichel:
> Could we have Leeway invalidate the build cache if the dev image changes?

Leeway already does that to some extend. The package version includes a hash over the environment manifest:  image

> Could we have Werft not use job protection if the job was changed and the owner is a Gitpod employee?

Indeed, we could tie the job protection to the newfound OPA-based policy mechanism.

What's the motivation for wanting to trigger a "cacheless" rebuild?

mads-hartmann:
> Leeway already does that to some extend. The package version includes a hash over the environment manifest

Ah cool. Right now the environment didn't capture that a binary was removed from the environment which broke the build. In this case that protoc-gen-js was no longer installed ([original pr](https://github.com/gitpod-io/gitpod/pull/10509), [revert](https://github.com/gitpod-io/gitpod/pull/10522), [follow up pr](https://github.com/gitpod-io/gitpod/pull/10529)); capturing all the binaries that might be called directly or transitively might be a bit too much of an ask.

> What's the motivation for wanting to trigger a "cacheless" rebuild?

It's to validate that you didn't break the build when you're changing something that doesn't cause Leeway to trigger a rebuild. In this case updating our Docker image ☺️

PR #10536, opened 2022-06-08
add temporary testing setup self-hosted

nandajavarma:
## Description

This is a very early stage solution. Currently opened as PR to test another PR in self-hosted setup.

To setup an infrastructure, open a workspace from this PR. Make sure you have access to the `core-dev` cluster in the present workspace.

- **Scenario**: You want to test the latest changes in the replicated channel `dev-nan` which you published to by using the comment `werft no-preview publish-to-kots` in your PR. You have the license file of your channel locally saved as `install/licenses/dev-nan.yaml` in your workspace(note that the name of the license file should be same as the channel name, and you don't have to push this change).  - If you are testing[ this PR of mine
](https://github.com/gitpod-io/gitpod/pull/10537), this step is already done for you. The license file is a part of this PR.

Run the following command:

```sh
werft run github -s install/licenses/dev-nan.yaml -j .werft/k3s-installer-tests.yaml -a destroy=false -a channel=dev-nan
```
Wait for the werft job to finish. You can go to the werft page, you will find a phase called `get-results` which would have logged the URL in which your preview env is deployed to and the contents of `kubeconfig` file. You can find an example of this run: https://werft.gitpod-dev.com/job/gitpod-custom-nvn-tmp-test.6

Please copy and paste that to a file in the aforementioned workspace. Run the following commands:
```
export KUBECONFIG=/path/to/the/new/file
curl https://kots.io/install | bash
kubectl kots -n gitpod admin-console
```

You will see the running kots UI in port 8800. Password is `gitpod`.

P. S. your setup will be deleted after two hours of initiation. We are working on a cleaner cleanup policy here.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
```

PR #10537, opened 2022-06-08
[installer] support service type `ClusterIP` for proxy

### LINKED ISSUE #9981: Add Option to Set Proxy Service as ClusterIP Type in KOTS Admin UI

For users who don't use the Proxy service as a LoadBalancer and instead have their own LoadBalancer in place, it would be very helpful to have the ability to set the Proxy service as ClusterIP instead. I believe having to manually set this has caused several failed Gitpod updates for me.

Ideally a simple toggle in the admin UI would be excellent.

Manually editing the service for the moment (`kubectl edit svc proxy -n gitpod`).

nandajavarma:
## Description

Currently, there is no way to do a post-processing in `KOTS` like [we suggest to do for replacing service type of `proxy`](https://github.com/gitpod-io/gitpod/blob/main/install/installer/README.md#post-processing-the-yaml) to `ClusterIP`. This PR introduces a new experimental field to the Config so users can provide a patch via the KOTS UI. The patch file can have the following structure:

```yaml
experimental:
  webapp:
    proxy:
      serviceType: "ClusterIP" # allowed values are "ClusterIP", "LoadBalancer", "NodePort", "ExternalName" consistent with https://pkg.go.dev/k8s.io/api/core/v1#ServiceType
```
> There is slight redundancy in the validation procedure at the moment(in the `validation.go` and `service.go` we check the allowed types). This is because we do not validate the config in KOTS as of now. We will remove this redundancy when we have better validation in place.

## Related Issue(s)

Fixes #9981 

## How to test

Follow the instructions of[ this PR ](https://github.com/gitpod-io/gitpod/pull/10536) to setup a self-hosted installation with this change. Go to the `Config` tab of the KOTS UI. In the bottom you will see an `Enable additional options` section. Enable it, and copy paste the aforementioned `YAML` snippet to a file and upload it. Update the installment, and if you do a `kubectl get svc proxy -n gitpod` from the CLI you will see that it now has `ClusterIP` type.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/install/installer/pkg/components/proxy/service.go
+++ b/install/installer/pkg/components/proxy/service.go
@@ -14,11 +14,29 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+var allowedServiceTypes = map[corev1.ServiceType]struct{}{
+	corev1.ServiceTypeLoadBalancer: {},
+	corev1.ServiceTypeClusterIP:    {},
+	corev1.ServiceTypeNodePort:     {},
+	corev1.ServiceTypeExternalName: {},
+}
+
 func service(ctx *common.RenderContext) ([]runtime.Object, error) {
+	serviceType := corev1.ServiceTypeLoadBalancer
+
 	loadBalancerIP := ""
 	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
-		if cfg.WebApp != nil && cfg.WebApp.ProxyConfig != nil && cfg.WebApp.ProxyConfig.StaticIP != "" {
-			loadBalancerIP = cfg.WebApp.ProxyConfig.StaticIP
+		if cfg.WebApp != nil && cfg.WebApp.ProxyConfig != nil {
+			if cfg.WebApp.ProxyConfig.StaticIP != "" {
+				loadBalancerIP = cfg.WebApp.ProxyConfig.StaticIP
+			}
+			st := cfg.WebApp.ProxyConfig.ServiceType
+			if st != nil {
+				_, allowed := allowedServiceTypes[corev1.ServiceType(*st)]
+				if allowed {
+					serviceType = *st
+				}
+			}
 		}
 		return nil
 	})
@@ -57,8 +75,10 @@ func service(ctx *common.RenderContext) ([]runtime.Object, error) {
 	}
 
 	return common.GenerateService(Component, ports, func(service *corev1.Service) {
-		service.Spec.Type = corev1.ServiceTypeLoadBalancer
-		service.Spec.LoadBalancerIP = loadBalancerIP
+		service.Spec.Type = serviceType
+		if serviceType == corev1.ServiceTypeLoadBalancer {
+			service.Spec.LoadBalancerIP = loadBalancerIP
+		}
 
 		service.Annotations["external-dns.alpha.kubernetes.io/hostname"] = fmt.Sprintf("%s,*.%s,*.ws.%s", ctx.Config.Domain, ctx.Config.Domain, ctx.Config.Domain)
 		service.Annotations["cloud.google.com/neg"] = `{"exposed_ports": {"80":{},"443": {}}}`
--- a/install/installer/pkg/config/v1/experimental/experimental.go
+++ b/install/installer/pkg/config/v1/experimental/experimental.go
@@ -154,8 +154,9 @@ type BlockedRepository struct {
 }
 
 type ProxyConfig struct {
-	StaticIP           string            `json:"staticIP"`
-	ServiceAnnotations map[string]string `json:"serviceAnnotations"`
+	StaticIP           string              `json:"staticIP"`
+	ServiceAnnotations map[string]string   `json:"serviceAnnotations"`
+	ServiceType        *corev1.ServiceType `json:"serviceType,omitempty" validate:"omitempty,service_config_type"`
 }
 
 type PublicAPIConfig struct {
--- a/install/installer/pkg/config/v1/experimental/validation.go
+++ b/install/installer/pkg/config/v1/experimental/validation.go
@@ -5,6 +5,8 @@
 package experimental
 
 import (
+	corev1 "k8s.io/api/core/v1"
+
 	"github.com/gitpod-io/gitpod/installer/pkg/cluster"
 	"github.com/go-playground/validator/v10"
 )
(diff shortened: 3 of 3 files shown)


install/installer/pkg/components/proxy/service.go
```diff
@@ -15,6 +16,19 @@ import (
 )
 
 func service(ctx *common.RenderContext) ([]runtime.Object, error) {
+	serviceType := corev1.ServiceTypeLoadBalancer
+	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
+		if cfg.Common != nil && cfg.Common.ServiceConfig != nil {
+			st, ok := cfg.Common.ServiceConfig["proxy"]
+			if ok {
+				if strings.ToLower(st.ServiceType) == "clusterip" {
```
adrienthebo, line 24: I see this conditional doesn't have an else clause; if `ServiceType` is misspelled (eg `cluterip`) will this this code default to `ServiceTypeLoadBalancer`? Is there validation occurring in this pipeline?

adrienthebo:
Copying my comment from Slack;

My philosophy on pull request review is to provide constructive feedback and make notes where I see potential changes/improvements/risks, but to grant approval as quickly so the author can land the work when they're ready and adopt feedback at their discretion. I think this is ready to ship but there _might_ be a case where an additional conditional clause (or documentation) could be useful.

Since there are no blocking issues on this PR, is it following team policy to grant an approval here? And if we grant that approval, is that sufficient to merge the PR?

install/installer/pkg/components/proxy/service.go
nandajavarma: This is a great point Adrien! I agree that we can add a validation process. As of this implementation, it just defaults to `LoadBalancer` as set in [this line](https://github.com/gitpod-io/gitpod/blob/main/install/installer/pkg/config/v1/experimental/validation.go).

About the validation, it is a great tip! We can leverage the existing [`validate` cmd](https://github.com/gitpod-io/gitpod/blob/main/install/installer/pkg/config/v1/experimental/validation.go)! I will do a follow-up PR for this!

adrienthebo (approved):
LGTM! 👍

corneliusludmann:
> My philosophy on pull request review is to provide constructive feedback and make notes where I see potential changes/improvements/risks, but to grant approval as quickly so the author can land the work when they're ready and adopt feedback at their discretion. I think this is ready to ship but there _might_ be a case where an additional conditional clause (or documentation) could be useful.

100 % agree with this philosophy!


> Since there are no blocking issues on this PR, is it following team policy to grant an approval here? And if we grant that approval, is that sufficient to merge the PR?

Approving PR is fine! :heavy_check_mark: However, since we automatically merge PRs by a bot once all required approvals are given, you should add the hold label on this PR so that the author has the opportunity to address your comments. You can do this with this command in a comment:

/hold

The author just need to remove the hold label or comment with `/unhold` and the PR will be merged.

(In this case, the PR still needs approval from Team WebApp. That's because this PR has not been merged yet automatically.)

install/installer/pkg/config/v1/experimental/experimental.go
```diff
@@ -24,8 +24,17 @@ type Config struct {
...
-	StaticMessagebusPassword string                `json:"staticMessagebusPassword"`
+	PodConfig map[string]*PodConfig `json:"podConfig,omitempty"`
+	// ServiceConfig is only supported for "proxy" service at the moment
+	ServiceConfig            map[string]*ServiceConfig `json:"serviceConfig,omitempty"`
+	StaticMessagebusPassword string                    `json:"staticMessagebusPassword"`
+}
+
+// ServiceConfig enables modification of type of service to `ClusterIP` if needed,
+// currently supported only for `proxy` service
+type ServiceConfig struct {
```
geropl, line 35: If this is special to proxy: Why not add it to `ProxyConfig` ([link](https://github.com/gitpod-io/gitpod/pull/10537/files#diff-7480b9126452b70b6c97afd8fbdcf95a8b614a986c02c113aafc5f691fef5630R165)), then?

Mentioning this because we have tried to keep things simple and non-generic in the installer to start with. And only extend to general solutions where it solved multiple use-cases. But I don't think we have another service that needs the same type of variance in the near future.

geropl: Also happy to sync to unblock here! :slightly_smiling_face:

nandajavarma: Yikes, I acted on a whim :see_no_evil: Have updated the setup now. Sorry for the back-and-fro.

install/installer/pkg/components/proxy/service.go
```diff
@@ -14,11 +14,29 @@ import (
...
 func service(ctx *common.RenderContext) ([]runtime.Object, error) {
+	serviceType := corev1.ServiceTypeLoadBalancer
+
 	loadBalancerIP := ""
 	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
-		if cfg.WebApp != nil && cfg.WebApp.ProxyConfig != nil && cfg.WebApp.ProxyConfig.StaticIP != "" {
-			loadBalancerIP = cfg.WebApp.ProxyConfig.StaticIP
+		if cfg.WebApp != nil && cfg.WebApp.ProxyConfig != nil {
+			if cfg.WebApp.ProxyConfig.StaticIP != "" {
+				loadBalancerIP = cfg.WebApp.ProxyConfig.StaticIP
```
geropl, line 31: Do we need to tie this to the `ServiceType`? :thinking: Really not sure what happens if we set this field, but do not have `LoadBalancer` type.

nandajavarma: You mean the bit here right? https://github.com/gitpod-io/gitpod/pull/10537/files#diff-d9a72ab10ad2fef022eeb810eac2018092ca6dec333c9c38c5c5c8c5244bafc0R79
It doesn't misbehave if we leave the `loadbalancerIP` in, I just didn't want to make the service definition confusing. I can roll this back.

nandajavarma: :heavy_check_mark:

nandajavarma: Scratch my comment, now I get what you mean. https://github.com/gitpod-io/gitpod/pull/10537/files#diff-d9a72ab10ad2fef022eeb810eac2018092ca6dec333c9c38c5c5c8c5244bafc0R79 this saves from setting the `loadbalancerIP` if the `ServiceType` is not loadbalancer. This reduces any potential confusion, although shouldn't effect the service in any ways.

I put  the check logic here just because the `webapp` experimental section was being parsed anyway, wanted to make the bust of the if jump.

geropl: >  this saves from setting the loadbalancerIP if the ServiceType is not loadbalancer.

Ah, thanks, missed that! :bow:

install/installer/pkg/config/v1/experimental/experimental.go
geropl: Awesome, thx for addressing! :pray:

geropl (approved):
:heavy_check_mark:

PR #10547, opened 2022-06-09
[usage] Add listing of memberships by UserID

easyCZ:
## Description

Adds implementation of listing memberships by User ID, needed to attribute workspace instances to a team.
This is used in sub-sequent [PR](https://github.com/gitpod-io/gitpod/pull/10534/files)

## Related Issue(s)

Fixes #

## How to test

Unit tests

## Release Notes

```release-note
NONE
```

## Documentation

NONE

### DIFF
--- a/components/usage/pkg/db/team_membership_test.go
+++ b/components/usage/pkg/db/team_membership_test.go
@@ -5,6 +5,7 @@
 package db_test
 
 import (
+	"context"
 	"github.com/gitpod-io/gitpod/usage/pkg/db"
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


components/usage/pkg/db/team_membership_test.go
```diff
@@ -30,3 +31,70 @@ func TestTeamMembership_WriteRead(t *testing.T) {
...
+			TeamID: uuid.New(),
+			UserID: userWithTwoTeams,
+			Role:   db.TeamMembershipRole_Member,
+		},
+		{
+			ID:     uuid.New(),
+			TeamID: uuid.New(),
+			UserID: userWithTwoTeams,
+			Role:   db.TeamMembershipRole_Member,
+		},
```
jankeromnes, line 56: FYI, this looks good, but I'm not sure why this works, given that the specified type for `TeamMembership.SubscriptionID` is:

```
	SubscriptionID uuid.UUID          `gorm:"column:subscriptionId;type:char;size:36;" json:"subscriptionId"`
```

Does Go set this member to `nil` by default? Also FYI, I think we'll want to change this type specification, because `TeamMembership.SubscriptionID` can also be `""` (empty string -- its default value) [[1](https://github.com/gitpod-io/gitpod/blob/b1a36659fdccc6fa45771360a1672002d7c1312d/components/gitpod-db/src/typeorm/entity/db-team-membership.ts#L32-L37))

components/usage/pkg/db/team_membership_test.go
```diff
@@ -30,3 +31,70 @@ func TestTeamMembership_WriteRead(t *testing.T) {
...
+	}{
+		{
+			Name:     "no user ids return empty",
+			UserIDs:  nil,
+			Expected: 0,
+		},
+		{
+			Name:     "no matching user IDs returns empty",
+			UserIDs:  []uuid.UUID{uuid.New()},
+			Expected: 0,
```
jankeromnes, line 75: Over-nit: This test has a non-zero chance of misfiring. 😂

jankeromnes (approved):
Thanks! Tests look good, so good to merge if green. 🛹

components/usage/pkg/db/team_membership_test.go
```diff
@@ -30,3 +31,70 @@ func TestTeamMembership_WriteRead(t *testing.T) {
...
+			ID:     uuid.New(),
+			TeamID: uuid.New(),
+			UserID: uuid.New(),
+			Role:   db.TeamMembershipRole_Member,
+		},
+		{
+			ID:     uuid.New(),
+			TeamID: uuid.New(),
+			UserID: userWithTwoTeams,
+			Role:   db.TeamMembershipRole_Member,
```
jankeromnes, line 49: One role could also be owner, for some variety. 💭

components/usage/pkg/db/team_membership_test.go
easyCZ: Good spot. Will investigate what it stores the uninitialized values as. The nil value for a UUID is 0 as that's the nil value for a byte, it will depend on how the Value() interface for the DB is implemented. Will follow up on this.

components/usage/pkg/db/team_membership_test.go
easyCZ: > UUIDs have 122 bits of entropy so the chance of two random UUIDs colliding is about 10^-37.

> If you generate 2^46 UUIDs (approximately 1 petabyte of entropy) the chance of getting a collision is 1 in 50 billion.[1] You're 170 times more likely to win the jackpot in Powerball.[2]

https://www.quora.com/Has-there-ever-been-a-UUID-collision

PR #10551, opened 2022-06-09
[usage] Generate usage report

easyCZ:
## Description

Generate a usage report in the reconciler.

Broader unit tests will be added in subsequent PR, to enable getting this into staging sooner and perform first pass validation.

## Related Issue(s)

Fixes #

## How to test

Unit tests

## Release Notes

```release-note
NONE
```

## Documentation

NONE

### DIFF
--- a/components/usage/pkg/controller/reconciler.go
+++ b/components/usage/pkg/controller/reconciler.go
@@ -6,12 +6,15 @@ package controller
 
 import (
 	"context"
-	"errors"
+	"encoding/json"
 	"fmt"
 	"github.com/gitpod-io/gitpod/common-go/log"
 	"github.com/gitpod-io/gitpod/usage/pkg/db"
 	"github.com/google/uuid"
 	"gorm.io/gorm"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"time"
 )
 
@@ -26,11 +29,12 @@ func (f ReconcilerFunc) Reconcile() error {
 }
 
 type UsageReconciler struct {
-	conn *gorm.DB
+	nowFunc func() time.Time
+	conn    *gorm.DB
 }
 
 func NewUsageReconciler(conn *gorm.DB) *UsageReconciler {
-	return &UsageReconciler{conn: conn}
+	return &UsageReconciler{conn: conn, nowFunc: time.Now}
 }
 
 type UsageReconcileStatus struct {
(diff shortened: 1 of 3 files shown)


components/usage/pkg/controller/reconciler.go
```diff
@@ -57,10 +61,25 @@ func (u *UsageReconciler) Reconcile() error {
...
+		return fmt.Errorf("failed to create temporary file: %w", err)
+	}
+	defer f.Close()
+
+	enc := json.NewEncoder(f)
+	err = enc.Encode(status.Report)
+	if err != nil {
+		return fmt.Errorf("failed to marshal report to JSON: %w", err)
+	}
+
```
andrew-farries: Should we write this to stdout instead? What's the expected workflow for getting hold of the report? With `kubectl exec`?

components/usage/pkg/controller/reconciler_test.go
```diff
@@ -14,67 +14,102 @@ import (
...
-			ID:          uuid.New(),
-			WorkspaceID: workspace.ID,
-			StoppedTime: db.NewVarcharTime(time.Date(2022, 06, 1, 1, 0, 0, 0, time.UTC)),
-			Status:      instanceStatus,
-		},
-	}
+	instanceStatus := []byte(`{"phase": "stopped", "conditions": {"deployed": false, "pullingImages": false, "serviceExists": false}}`)
+
+	scenarios := []Scenario{
+		(func() Scenario {
```
andrew-farries, line 34: Why do we need the `func` here. It just returns one `Scenario` without taking any inputs.

andrew-farries (changes requested):

components/usage/pkg/controller/reconciler.go
easyCZ: Yep, for now it would be kubexec. It does also log it when it prints the status (which contains the report) but that may not be as easily retrievable as the file.

components/usage/pkg/controller/reconciler_test.go
easyCZ: To encapsulate the extra variables that are needed for this scenario outside of the scope of other scenarios. I've a follow-up PR which adds more complex tests to this.

andrew-farries (approved):

Issue #10327, opened 2022-05-29
Show whether a user has a valid subscription and allow managing it


easyCZ:
Related to https://github.com/gitpod-io/gitpod/issues/10326.

When a user has already subscribed through Stripe, we need to show that a user has a valid subscription.

jankeromnes:
Fixed in https://github.com/gitpod-io/gitpod/pull/10555

PR #10555, opened 2022-06-09
Implement Stripe portal to allow usage-based customers to manage their billing details

jankeromnes:
## Description

- [x] Implement Stripe portal to allow usage-based customers to manage their subscription.

## Related Issue(s)

Part of https://github.com/gitpod-io/gitpod/issues/10327

## How to test

1. Create a team called "Gitpod", or "Gitpod1", "Gitpod2" ...
2. Navigate to Team Settings > Team Billing
3. Wait for the Usage-Based Billing UI to appear
4. Upgrade to Usage-Based Billing by adding a credit card (e.g. 4242 4242 4242 4242, 4/24, 424)
5. Once upgraded, clicking on `Manage Billing` should open the Stripe Portal
6. (The Stripe Portal also has a link that takes you back to Gitpod)

| Click on Manage Billing | View Stripe Portal |
| --- | --- |
| Screenshot 2022-06-10 at 08 29 13 | Screenshot 2022-06-10 at 08 29 30 |

## Release Notes

```release-note
NONE
```

## Documentation

- [x] /werft with-payment

### DIFF
--- a/components/dashboard/src/teams/TeamUsageBasedBilling.tsx
+++ b/components/dashboard/src/teams/TeamUsageBasedBilling.tsx
@@ -27,6 +27,7 @@ export default function TeamUsageBasedBilling() {
     const [showBillingSetupModal, setShowBillingSetupModal] = useState<boolean>(false);
     const [pendingStripeCustomer, setPendingStripeCustomer] = useState<PendingStripeCustomer | undefined>();
     const [pollStripeCustomerTimeout, setPollStripeCustomerTimeout] = useState<NodeJS.Timeout | undefined>();
+    const [stripePortalUrl, setStripePortalUrl] = useState<string | undefined>();
 
     useEffect(() => {
         if (!team) {
@@ -45,6 +46,16 @@ export default function TeamUsageBasedBilling() {
             }
         })();
     }, [team]);
+
+    useEffect(() => {
+        if (!team || !stripeCustomerId) {
+            return;
+        }
+        (async () => {
+            const portalUrl = await getGitpodService().server.getStripePortalUrlForTeam(team.id);
+            setStripePortalUrl(portalUrl);
+        })();
+    }, [team, stripeCustomerId]);
 
     useEffect(() => {
         if (!team) {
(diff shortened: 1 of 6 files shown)


easyCZ:
@jankeromnes Thanks for the PR. I haven't had a chance to dig into the details of Stripe API yet but to give you a faster review, how does the authentication and authorization against Stripe work from our link? Specifically:
1. Does the user in effect need to create an account to access teh dashboard? Or are we generating a one-time sign-in link which we're exposing in our dashboard?
2. What guarantees that a user outside a team is not given access to another teams portal?

jankeromnes:
Thanks @easyCZ! Good questions. The answers are in the code, but to give you fast answers:

> 1. Does the user in effect need to create an account to access teh dashboard?

Yes, you need both a Gitpod account and a Stripe customer account before you can see the Stripe Portal button, let alone open it.

> Or are we generating a one-time sign-in link which we're exposing in our dashboard?

We are indeed generating a one-time sign-in link server-side to allow specifically one customer to open their own Stripe Portal.

> 2. What guarantees that a user outside a team is not given access to another teams portal?

We do not give a one-time sign-in link to users who are not an Owner of the Team that is the Stripe Customer (standard team-ownership access control check before generating the one-time sign-in link).

Also, FYI from the portal, you can only see a semi-censored version of your own customer details, and update them to new values.

easyCZ:
Thanks. And last one:
> We are indeed generating a one-time sign-in link server-side to allow specifically one customer to open their own Stripe Portal.

From the code, I understand we generate the portal link based on the user. Is this correct? Should it instead be the team? I'm asking because it's unclear if you're in multiple teams, which link to the portal you should receive.

jankeromnes:
> From the code, I understand we generate the portal link based on the user. Is this correct? Should it instead be the team? I'm asking because it's unclear if you're in multiple teams, which link to the portal you should receive.

No, we actually generate the portal link based on the customer, which currently can only be a team.

Note: The `getStripePortalUrlForTeam` API has a required `teamId` parameter.

easyCZ (approved):

easyCZ:
Thanks for the info @jankeromnes. LGTM

PR #10561, opened 2022-06-09
Add `AKS` terraform module 

nandajavarma:
## Description

This PR adds `AKS` to the cloud providers supported in the installer nightly tests. This is a very basic setup of `AKS` without external dependencies and such. That would be done in a follow-up PR.

## Related Issue(s)

Fixes #

## How to test

If you have access to the `core-dev` cluster on GCP, you can run the following command:
```
werft run github -a debug=true -j .werft/aks-installer-tests.yaml
```
The integration tests are still failing. This issue is being handled by the workspace team.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.werft/aks-installer-tests.yaml
+++ b/.werft/aks-installer-tests.yaml
@@ -1,0 +1,94 @@
+# debug using `werft run github -f -s .werft/installer-tests.ts -j .werft/aks-installer-tests.yaml -a debug=true`
+pod:
+  serviceAccount: werft
+  affinity:
+  nodeAffinity:
+    requiredDuringSchedulingIgnoredDuringExecution:
+      nodeSelectorTerms:
+      - matchExpressions:
+        - key: dev/workload
+          operator: In
+          values:
+          - "builds"
+  securityContext:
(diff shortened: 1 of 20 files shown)


MrSimonEmms (approved):
/hold

Looks good to me. I've put it on hold and you can resolve the suggestions however you like (future PRs are fine).

I'd suggest adding `.terraform.lock.hcl` to the repo. This is similar to `package-lock.json` and `go.sum` which makes the `terraform init` (marginally) faster.

I'd recommend adding a couple of tests to the Terraform files:
 - `terraform fmt -check -recursive /path/to/install/infra`
 - `terraform validate` (this will need `cd /path/to/install/infra/terraform/<aks,gke,k3s,etc>` first as it doesn't accept a path argument)

nandajavarma:
/unhold

@MrSimonEmms I am going to let this one in since [the next PR](https://github.com/gitpod-io/gitpod/pull/10777) adding external dependencies comes right after this. Changing in this PR means conflict resolution/rebase etc. on the next one. Makes more sense to add there.

PR #10472, opened 2022-06-04
[dashboard] Fix Prebuilds item truncate

flaming-codes:
## Description
This PR fixes a small truncation issue w/ text in the **Prebuilds** items.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/9103

## How to test
No specific tests for this change are available.

## Release Notes

```release-note
NONE
```

## Screenshots

| Description | Image |
| --- | --- |
| Before fix | Screenshot 2022-06-04 at 20 37 17 |
| After fix | Screenshot 2022-06-04 at 20 37 08 |

PR #10564, opened 2022-06-09
chore(dashboard): adjust prebuild table col width

andreafalzetti:
## Description
The Prebuilds page not always align correctly the columns, e.g. see screeenshot below:

Screenshot 2022-06-09 at 11 21 36 copy

I think this small CSS change could improve already some scenarios.

## Related Issue(s)

N/A

## How to test
1. Start the dashboard for gitpod.io

(I would also suggest Prev Env, but there are no prebuilds there)

| Before | After |
|--------|-------|
|     Screenshot 2022-06-09 at 11 21 36   |   Screenshot 2022-06-09 at 11 21 03    |

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/dashboard/src/projects/Prebuilds.tsx
+++ b/components/dashboard/src/projects/Prebuilds.tsx
@@ -190,7 +190,7 @@ export default function (props: { project?: Project; isAdminDashboard?: boolean 
                 </div>
                 <ItemsList className="mt-2">
                     <Item header={true}>
-                        <ItemField className="my-auto w-5/12">
+                        <ItemField className="my-auto md:w-3/12 xl:w-4/12">
                             <span>Prebuild</span>
                         </ItemField>
                         <ItemField className="my-auto w-5/12">
@@ -216,7 +216,7 @@ export default function (props: { project?: Project; isAdminDashboard?: boolean 
                             >
                                 <Item key={`prebuild-${p.info.id}`}>
                                     <ItemField
-                                        className={`flex items-center my-auto w-5/12 ${
+                                        className={`flex items-center my-auto md:w-3/12 xl:w-4/12 ${
                                             props.isAdminDashboard ? "pointer-events-none" : ""
                                         }`}
                                     >


geropl:
@AlexTugarev Could you review this PR (tomorrow)? :pray:

jankeromnes:
Happy to take a look 👀

components/dashboard/src/projects/Prebuilds.tsx
```diff
@@ -190,7 +190,7 @@ export default function (props: { project?: Project; isAdminDashboard?: boolean
                 </div>
                 <ItemsList className="mt-2">
                     <Item header={true}>
-                        <ItemField className="my-auto w-5/12">
+                        <ItemField className="my-auto md:w-3/12 xl:w-4/12">
```
AlexTugarev, line 193: It would be great to not overshoot at `md`.

AlexTugarev:
@jankeromnes, are you still looking at this? This seems to fix a bigger issue, and I'm keen to merge it as is. wdyt?

jankeromnes:
Ah, yes, thanks for the reminder. Triggered a build then forgot about this.

Resuming review now. 👀

components/dashboard/src/projects/Prebuilds.tsx
jankeromnes: Tip for the future: Using CSS grid is much more powerful and robust in these situation than trying to guess good fixed widths.

With grid, you can simply specify "I want three columns of exactly the same width", or specify custom relative column widths, but they'll almost always be correct regardless of screen size, margins, size of content, etc.

jankeromnes (approved):
Thanks a lot for improving our Prebuilds list! ✨

I tested this change at various resolutions, and with various commit messages / branch name lengths. I think the current implementation is still not perfect (e.g. might need an ellipsis + tooltip on the branch name) hence the hold, but please feel free to merge this when you like (this is easy to adjust later as well).

FYI, here are my test results:

| My normal screen size | Max screen size | Smartphone (>50% of web users) | iPad (portrait) |
| --- | --- | --- | --- |
| Screenshot 2022-06-14 at 11 25 40 | Screenshot 2022-06-14 at 11 26 09 | Screenshot 2022-06-14 at 11 26 43 | Screenshot 2022-06-14 at 11 27 01 |

**Tip**: To automatically add an ellipsis to a potentially super long cell (e.g. the branch name), you can add:
- the `truncate` CSS class
- a `title` attribute with the exact same text content as the element (so that users can still hover over the truncated text to see the full text)

See also how the commit message is truncated for inspiration.

/hold

jankeromnes:
Oh, wait, I wonder if https://github.com/gitpod-io/gitpod/pull/10472 maybe already fixes the same issue? 👀

Looks like we may need a `overflow-hidden` in addition to the `truncate` class. 💡

akosyakov:
should be merged or closed?

gtsiolis:
I'd suggest getting this merged as it's already a good MVC improvement, regardless if the rendering on small (mobile) viewports looks great, leaving this for https://github.com/gitpod-io/gitpod/issues/4050.

If that's useful, I'd like to help get https://github.com/gitpod-io/gitpod/pull/10472 also merged afterwards.

akosyakov:
@jankeromnes @AlexTugarev is it alright to remove hold?

jankeromnes:
Hi! Yes, and sorry for sitting on this for so long.

@gtsiolis's [comment](https://github.com/gitpod-io/gitpod/pull/10564#issuecomment-1167311730) makes a good case or merging this. 👍 Thus, removing the hold.

/unhold

PR #10570, opened 2022-06-10
feat(gitpod-cli): add top cmd

### LINKED ISSUE #9521: Expose perf details for the current workspace

andreafalzetti:
## Description
Introducing a new command for gitpod-cli: `$ gp top`

```shell
Command line interface for Gitpod

Usage:
  gp [command]

Available Commands:
...
  top                 Display workspace resource (CPU/Memory usage)
```

## Related Issue(s)

Fixes #9521

## How to test
1. Start this workspace in prev env"
2. Run `gp top`, observe the cpu/memory usage
3. To raise the CPU usage, run: `yes > /dev/null &` once,
4. Run `gp top` again and obsere the CPU usage. Run the command above a few times, until the color of CPU changes.
5. To get the CPU back to a lower value, run: `killall yes`

## Release Notes

```release-note
Added a new command to gitpod-cli: "gp top" which displays workspace resources (CPU/Memory usage)
```
Screenshot 2022-06-09 at 21 59 49

Screenshot 2022-06-09 at 21 59 20

Screenshot 2022-06-09 at 21 59 33


components/gitpod-cli/cmd/top.go
```diff
@@ -0,0 +1,70 @@
...
+		return tablewriter.FgYellowColor
+	default:
+		return tablewriter.FgHiGreenColor
+	}
+}
+
+var topCmd = &cobra.Command{
+	Use:   "top",
+	Short: "Display workspace resource (CPU and memory usage)",
+	Run: func(cmd *cobra.Command, args []string) {
```
andreafalzetti, line 55: In the future, we could add a flag `--watch` to retrieve resources every X seconds

components/gitpod-cli/cmd/top.go
```diff
@@ -0,0 +1,70 @@
...
+
+	cpuColor := getColor(cpuFraction)
+	memoryColor := getColor(memFraction)
+
+	table.Rich([]string{cpu, memory}, []tablewriter.Colors{{cpuColor}, {memoryColor}})
+
+	table.Render()
+}
+
+func getColor(value int64) int {
```
andreafalzetti, line 41: I went for a simple "semaphore" approach (green / yellow / red).

Threshold are randomly picked based on common sense - happy to adjust them :)

components/gitpod-cli/pkg/supervisor-helper/status.go
```diff
@@ -0,0 +1,26 @@
...
+
+	supervisor "github.com/gitpod-io/gitpod/supervisor/api"
+)
+
+func GetWorkspaceResources(ctx context.Context) (*supervisor.ResourcesStatusResponse, error) {
+	conn, err := Dial(ctx)
+	if err != nil {
+		return nil, err
+	}
+	client := supervisor.NewStatusServiceClient(conn)
```
andreafalzetti, line 18: In the future, we could expand supervisor to also be aware of the vm resources and return additional metadata. As mentioned [here](https://github.com/gitpod-io/website/pull/2207/files#r894121494), the available amounts might not be true, if the vm is running out of resources due to noisy neighbours.

components/gitpod-cli/cmd/top.go
akosyakov: one should be able to run `watch gp top` as well

components/gitpod-cli/pkg/supervisor-helper/status.go
akosyakov: Let's stick with the simple approach for now and assume that VM always has enough resources.

components/gitpod-cli/cmd/top.go
akosyakov: Can we provide a flag to disable coloring? I tried `watch gp top` and it does not look nice for some reason.

components/gitpod-cli/cmd/top.go
```diff
@@ -0,0 +1,70 @@
...
+	case value >= 65:
+		return tablewriter.FgYellowColor
+	default:
+		return tablewriter.FgHiGreenColor
+	}
+}
+
+var topCmd = &cobra.Command{
+	Use:   "top",
+	Short: "Display workspace resource (CPU and memory usage)",
```
akosyakov, line 54: `Display usage of workspace resources (CPU and memory)` ?

akosyakov (approved):
/lgtm

/hold
if you want to make any small adjustments, otherwise it looks good

components/gitpod-cli/cmd/top.go
andreafalzetti: `watch gp top` is perfect for now 🛹 👍

components/gitpod-cli/pkg/supervisor-helper/status.go
andreafalzetti: yep, that's the plan 👍

components/gitpod-cli/cmd/top.go
andreafalzetti: I am implementing an improvement to detect if `gp` is being called by another process or is piped with stdin/stdout and disable colors accordingly.

I will also provide the flag for completeness 👍 

PR moved to draft while I finish this change.

andreafalzetti: I've updated the PR, now `watch gp top` should just work (colors will be automatically disabled)

components/gitpod-cli/cmd/top.go
```diff
@@ -0,0 +1,80 @@
...
+		if err != nil {
+			log.Fatalf("cannot get workspace resources: %s", err)
+		}
+
+		outputTable(workspaceResources)
+	},
+}
+
+func init() {
+	topCmd.Flags().BoolVarP(&noColor, "no-color", "", false, "rewrites the host header of passing HTTP requests to localhost")
```
akosyakov, line 78: `rewrites the host header of passing HTTP requests to localhost"` ? :)

andreafalzetti: 🤦‍♂️

filiptronicek (commented):
Thanks a lot @andreafalzetti 🚀!

I just have two questions:
1. Would it be a common use case for users to run this with `watch`? Maybe we could add a way to run continuously to behave more like `top`?
2. Is there a way to consume the info via other means? Can I use these values in my app somehow? I would like to for instance output JSON so it can be consumed by tools like https://github.com/akosyakov/gitpod-monitor (still the best VS Code extension there is).

andreafalzetti:
> Thanks a lot @andreafalzetti 🚀!
> 
> I just have two questions:
> 
>     1. Would it be a common use case for users to run this with `watch`? Maybe we could add a way to run continuously to behave more like `top`?
> 
>     2. Is there a way to consume the info via other means? Can I use these values in my app somehow? I would like to for instance output JSON so it can be consumed by tools like https://github.com/akosyakov/gitpod-monitor (still the best VS Code extension there is).

These are good points, I've also mentioned them in this PR - see: [1](https://github.com/gitpod-io/gitpod/pull/10570#discussion_r894120724) and [2](https://github.com/gitpod-io/gitpod/pull/10570#discussion_r894121174) :)

andreafalzetti:
Thanks @akosyakov and @filiptronicek 

/unhold

Issue #9932, opened 2022-05-11
Epic: Upload users SSH keys to Gitpod


loujaybee:
### Context

Local companion dynamically creates SSH keys for users, to grant SSH access, and the copy/paste SSH method for Gitpod uses an access code from within the workspace, and you can even manually inject your SSH public key using dotfiles. Whilst both of these approaches work, it would be better to allow users to upload their own SSH public keys to Gitpod to access their workspace using a more conventional SSH approach. We may also want to consider re-using existing public keys from GitHub, or other providers, for instance. 

### Value

1. Improved security (as not using a token)
2. A simpler UX for SSH'ing into workspaces
3. A more familiar SSH pattern (matches with user expectations)

### In Scope

* Uploading custom keys

### Out of scope

* Pulling SSH keys from 3rd party providers

### Related Issues

- https://github.com/gitpod-io/gitpod/pull/10704
- https://github.com/gitpod-io/gitpod/pull/10617
- https://github.com/gitpod-io/gitpod/pull/10573
- https://github.com/gitpod-io/website/issues/2371
- https://github.com/gitpod-io/website/issues/2375
- https://github.com/gitpod-io/gitpod/issues/11265

### Public FAQ

<details>
<summary><strong>Will we allow users to add SSH keys manually?</strong></summary>
Yes.
</details>

<details>
<summary><strong>Will we keep the current owner token SSH solution in place?</strong></summary>
Yes, as there is value in the quick copy/paste, however SSH key upload should be recommended. 
</details>

<details>
<summary><strong>Will importing SSH keys from GitHub be the primary way add SSH keys?</strong></summary>
Unlikely. Uploading custom keys would be the simplest implementation, which we could then extend with ways to fetch existing keys as a future update or refinement. We will start by setting up the Gitpod infrastructure for using SSH keys, and then look at ways to import/re-use existing keys from other providers.  
</details>

<details>
<summary><strong>Can we fetch SSH keys from all other providers (GitHub, BitBucket, GitLab)?</strong></summary>
- GitHub has an API
- GitLab has an API
- BitBucket needs to be investigated
</details>

<details>
<summary><strong>Can you use an SSH Key from one provider to connect to workspaces for repositories in other providers?</strong></summary>
Needs investigation.
</details>

### Internal FAQ

<details>
<summary><strong>What happens if SSH keys get removed from a provider, can we detect that?</strong></summary>
Needs investigation
</details>

loujaybee:
[...]

PR #10573, opened 2022-06-10
Support ssh public keys configuration

mustard-mh:
## Description

This PR works on allowing users to upload their own ssh public key to Gitpod, make copy-paste ssh connect more conventional, and make sharing of ssh connect command more safety

[Link of Figma](https://www.figma.com/file/UTwRHTYhuF05IiaEWcL9MT/Upload-SSH-keys-to-Gitpod-%239932?node-id=1%3A2614)

## TODO
- [x] ‼️ Find out how to setup db-sync
- [x] ‼️ Merge PR https://github.com/gitpod-io/gitpod/pull/10617
- [x] ‼️ Deploy previous PR https://github.com/gitpod-io/gitpod/pull/10617
- [x] Dashboard ssh copy-paste improve
- [x] Make copy-paste ways can be switched
- [x] Properly content for users

~~Functions for late used time?~~ 

## Related Issue(s)

Relate https://github.com/gitpod-io/gitpod/issues/9932

## How to test

Modal of ssh copy-paste always shows `SSH Key` as the default selection, if the user has no public key uploaded, it will show a warning in this section.

- Start a workspace in prev env
- Wait until the workspace is ready
- Connect with ssh copy-paste, get `conn A`
    - Failed with ssh key conn
    - Success with access token conn
- add a public key via setting -> ssh keys
- Connect with ssh copy-paste use `ssh key`, get `conn B`
- Delete that public key you set in the settings
- Connection B should still work
- A new `conn C` which use the same command like conn B, will not works - ask a password

## Release Notes

```release-note
Allow users to add and remove SSH public keys in settings
Change UX of SSH copy-paste to support ssh key connection
```

## Documentation

- [ ] TODO

## Werft options:

- [x] /werft with-preview


roboquat:

<details>

</details>

components/dashboard/tailwind.config.js
```diff
@@ -5,96 +5,90 @@
...
             container: {
                 center: true,
             },
             outline: {
-                blue: '1px solid #000033',
+                blue: "1px solid #000033",
+            },
+            width: {
+                120: "28rem",
+                128: "32rem",
```
mustard-mh, line 43: This file only adds these two new definitions, since I tried with [`w-[28rem]`](https://tailwindcss.com/docs/width#arbitrary-values) but not worked.

I can revert changes and add these four lines then save without format if you want.

components/gitpod-db/src/typeorm/user-db-impl.ts
```diff
@@ -395,6 +402,48 @@ export class TypeORMUserDBImpl implements UserDB {
...
+        const repo = await this.getSSHPublicKeyRepo();
+        const fingerprint = SSHPublicKeyValue.getFingerprint(value);
+        const allKeys = await repo.find({ where: { userId } });
+        const prevOne = allKeys.find((e) => e.fingerprint === fingerprint);
+        if (!!prevOne) {
+            if (prevOne.deleted === false) {
+                throw new Error(`Duplicate public key with ${prevOne.name}`);
+            }
+            prevOne.deleted = false;
+            return repo.save(prevOne);
```
mustard-mh, line 425: Not sure if this will affect `db-sync` or not

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
```diff
@@ -55,39 +55,46 @@ interface SSHProps {
...
+            .server.hasSSHPublicKey()
+            .then((d) => {
+                setHasSSHKey(d);
+            })
+            .catch(console.error);
+    }, []);
+
+    const host = props.ideUrl.replace(props.workspaceId, props.workspaceId + ".ssh");
+    const sshPswCommand = `ssh '${props.workspaceId}#${props.ownerToken}@${host}'`;
+    const sshKeyCommand = `ssh '${props.workspaceId}@${host}'`;
```
mustard-mh, line 71: We should make sure user who has ssh key uploaded can switch between `sshPswCommand` and `sshKeyCommand` since that private key may not exists in the computer user is using now. 

This need design too? @gtsiolis

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
```diff
@@ -55,39 +55,46 @@ interface SSHProps {
...
-                <p className="mt-4 text-gray-500 whitespace-normal text-base">
+        <div>
+            <div className="space-y-4">
+                <Alert type="warning" className="whitespace-normal">
+                    <b>Anyone</b> on the internet with this command can access the running workspace. The command
+                    includes a generated access token that resets on every workspace restart.
+                </Alert>
+                <Alert type="info" className="whitespace-normal">
+                    Before connecting via SSH, make sure you have an existing SSH private key on your machine. You can
+                    create one using&nbsp;
```
mustard-mh, line 82: Display different `warning` `info` content for `sshPswCommand` and `sshKeyCommand`. cc @gtsiolis @loujaybee

components/dashboard/src/settings/SSHKeys.tsx
```diff
@@ -0,0 +1,257 @@
...
+export function DeleteSSHKeyModal(props: DeleteModalProps) {
+    const confirmDelete = async () => {
+        await getGitpodService().server.deleteSSHPublicKey(props.value.id!);
+        props.onConfirm();
+        props.onClose();
+    };
+    return (
+        <ConfirmationModal
+            title="Delete SSH Key"
+            areYouSureText="This action CANNOT be undone. This will permanently delete the SSH key and if you'd like to use it in the future, you will need to upload it again."
```
gtsiolis, line 136: **suggestion**: We could go with a less scary warning here, unless you think this is really needed here. Updating the copy later is also an option. Also, including the key title and last used timestamp could help associate the key about to be deleted. What do you think?

```suggestion
            areYouSureText="Are you sure you want to delete this SSH Key?"
```

| Delete SSH Key Modal |
|-|
| Screenshot 2022-06-16 at 7 02 00 PM |

components/gitpod-protocol/src/protocol.ts
```diff
@@ -356,6 +356,68 @@ export namespace UserEnvVar {
...
+    userId: string;
+    fingerprint: string;
+    creationTime: string;
+    lastUsedTime?: string;
+}
+
+export type UserSSHPublicKeyValue = Omit<UserSSHPublicKey, "key" | "userId">;
+
+export namespace SSHPublicKeyValue {
+    export function validate(value: SSHPublicKeyValue): string | undefined {
```
gtsiolis, line 375: **issue(non-blocking)**: What do you think of also improving the error message when trying to add the same key again?

| Duplicate Key Error |
|-|
| Screenshot 2022-06-16 at 7 06 37 PM |

gtsiolis (commented):
Thanks for adding this @mustard-mh! ✨ 

Left some first-round comments below. 🏓

components/dashboard/src/settings/SSHKeys.tsx
```diff
@@ -0,0 +1,257 @@
...
+                                    </p>
+                                    <div className="truncate my-1 text-xl text-gray-800 dark:text-gray-100 font-semibold">
+                                        {key.name}
+                                    </div>
+                                    <p className="truncate mt-4">
+                                        Added on {moment(key.creationTime).format("MMM D, YYYY, hh:mm A")}
+                                    </p>
+                                    {!!key.lastUsedTime && (
+                                        <p className="truncate">
+                                            Last used on {moment(key.lastUsedTime).format("MMM D, YYYY, hh:mm A")}
```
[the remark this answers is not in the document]
iQQBot: With the current implementation, we don't have a way to report usage yet, so this part should just be set aside

components/gitpod-db/src/typeorm/user-db-impl.ts
iQQBot: I think there is a certain possibility of problems, `GC` and `db-sync` should be independent, if one side of the row has been deleted by GC, and db-sync is suspended, then there may be problems

I think the rows that have been marked as deleted can be excluded from the query

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
iQQBot: I think this is worth adding

components/gitpod-db/src/typeorm/user-db-impl.ts
```diff
@@ -395,6 +402,48 @@ export class TypeORMUserDBImpl implements UserDB {
...
+
+    public async getSSHPublicKeys(userId: string): Promise<UserSSHPublicKey[]> {
+        const repo = await this.getSSHPublicKeyRepo();
+        return repo.find({ where: { userId, deleted: false }, order: { creationTime: "ASC" } });
+    }
+
+    public async addSSHPublicKey(userId: string, value: SSHPublicKeyValue): Promise<UserSSHPublicKey> {
+        const repo = await this.getSSHPublicKeyRepo();
+        const fingerprint = SSHPublicKeyValue.getFingerprint(value);
+        const allKeys = await repo.find({ where: { userId } });
```
iQQBot, line 418: does it make sense to add `deleted: false` as where condition?

components/gitpod-protocol/src/protocol.ts
```diff
@@ -356,6 +356,68 @@ export namespace UserEnvVar {
...
+            key: resultGroup.groups?.["key"] as string,
+            email: resultGroup.groups?.["email"] || undefined,
+        };
+    }
+
+    export function getFingerprint(value: SSHPublicKeyValue) {
+        const data = getData(value);
+        let buf = Buffer.from(data.key, "base64");
+        // gitlab style
+        const hash = createHash("md5").update(buf).digest("hex");
```
iQQBot, line 412: I would like to use github style, because when I using `ssh -vvv` to show debug output, I can see my `public key` fingerprint, it same with `Github`

image

image

components/gitpod-db/src/typeorm/user-db-impl.ts
```diff
@@ -395,6 +402,48 @@ export class TypeORMUserDBImpl implements UserDB {
...
+        const prevOne = allKeys.find((e) => e.fingerprint === fingerprint);
+        if (!!prevOne) {
+            if (prevOne.deleted === false) {
+                throw new Error(`Duplicate public key with ${prevOne.name}`);
+            }
+            prevOne.deleted = false;
```
iQQBot, line 424: also need use new `name`

components/dashboard/src/settings/SSHKeys.tsx
```diff
@@ -0,0 +1,257 @@
...
+    const save = async () => {
+        const tmp = SSHPublicKeyValue.validate(value);
+        if (tmp) {
+            setErrorMsg(tmp);
+            return false;
+        }
+        try {
+            await getGitpodService().server.addSSHPublicKey(value);
+        } catch (e) {
+            setErrorMsg(e.message);
```
iQQBot, line 56: image

Maybe we can do some improve about this error message

iQQBot (approved):
Thank you @mustard-mh 🎉  I test it, work like a charm 🙏 

left some comment

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
```diff
@@ -55,39 +55,46 @@ interface SSHProps {
...
+                    <a
+                        href="https://en.wikipedia.org/wiki/Ssh-keygen"
+                        target="_blank"
+                        rel="noopener noreferrer"
+                        className="gp-link"
+                    >
+                        ssh-keygen
+                    </a>
+                    .
+                </Alert>
```
iQQBot, line 92: We can remove this note after https://github.com/gitpod-io/gitpod/pull/10704 merge

components/dashboard/src/settings/SSHKeys.tsx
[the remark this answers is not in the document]
gtsiolis: @iQQBot Do you mean this element (_Last used on ..._) will never show up in the current implementation?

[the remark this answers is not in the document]
iQQBot: Yes

components/dashboard/src/settings/SSHKeys.tsx
gtsiolis: Thanks, @iQQBot! The comment in https://github.com/gitpod-io/gitpod/pull/10573#discussion_r899263105 was aslo about this error. I've updated the screenshot to depict this better.

components/gitpod-db/src/typeorm/user-db-impl.ts
mustard-mh: Any suggestion? Should we delete the UNIQUE INDEX for (userId and fingerprint) in DB and use DB transaction here? 
cc @jankeromnes

components/dashboard/src/settings/SSHKeys.tsx
mustard-mh: Good for me, the previous error message is a copy from GH

components/gitpod-db/src/typeorm/user-db-impl.ts
mustard-mh: If we consider record with `delete: true` should be reverted to `delete: false` then not make sense

components/gitpod-protocol/src/protocol.ts
mustard-mh: I prefer GH style too, that's why I leave it here🙈

components/dashboard/src/settings/SSHKeys.tsx
gtsiolis: > ... the previous error message is a copy from GH

Yes, I noticed that. 🙂 

I'd think it's ok to go with our own copy style found in the product for now and adjust later if needed.

Thank you, @mustard-mh! 🙏

components/dashboard/tailwind.config.js
jankeromnes: Hey, many thanks for contributing larger widths 💯 I think this will be very useful.

Also, no worries about the formatting -- looks good to me, thanks for doing this as well. 👍

Small nit: Shouldn't `120: "28rem"` actually be `112: "28rem"`? (Seems like in Tailwind v2, width/height units are usually `rem * 4`)

components/gitpod-db/src/typeorm/entity/db-user-ssh-public-key.ts
```diff
@@ -0,0 +1,58 @@
+/**
+ * Copyright (c) 2020 Gitpod GmbH. All rights reserved.
```
jankeromnes, line 2: Nit: Not very important, but when creating new files, the year should be the current year. 😋

```suggestion
 * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
```

components/gitpod-db/src/typeorm/transformer.ts
```diff
@@ -38,6 +38,18 @@ export namespace Transformer {
         },
     };
 
+    export const MAP_ISO_STRING_TO_TIMESTAMP_DROP_NULLABLE: ValueTransformer = {
+        to(value: any): any {
+            return MAP_ISO_STRING_TO_TIMESTAMP_DROP.to(value);
+        },
+        from(value: any): any {
+            if (value == null) {
+                return null;
```
jankeromnes, line 47: 👍

components/gitpod-db/src/typeorm/transformer.ts
```diff
@@ -38,6 +38,18 @@ export namespace Transformer {
         },
     };
 
+    export const MAP_ISO_STRING_TO_TIMESTAMP_DROP_NULLABLE: ValueTransformer = {
```
jankeromnes, line 41: Interesting. I know that we store times as both ISO strings and timestamps in MySQL. I'm just curious -- what is the benefit of converting ISO string to timestamp here? Is it more efficient to store timestamps in MySQL? 👀

components/gitpod-db/src/typeorm/user-db-impl.ts
jankeromnes: Hi, thanks for the ping @mustard-mh!

Oh, re-reading your question, I've just now understood the relationship between the UNIQUE INDEX (`UNIQUE KEY ind_unique_userId_fingerprint (userId, fingerprint)` in the migration) and why you're trying to handle rows with `deleted: true` here -- you're worried that, if someone adds a key with a fingerprint of a just-deleted key (but not yet fully removed), MySQL will throw an error because of the duplicate userId/fingerprint, right?

In that case, yes, I think removing the UNIQUE INDEX makes sense, as it allows you to delete & recreate keys without having to wait for full removal of deleted rows (or trying to "undeleted" keys that are in the process of being deleted).

Now, I just don't fully understand why a DB transaction would be useful here. 👀 Could you please explain?

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -1929,6 +1932,61 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
...
+            fingerprint: e.fingerprint,
+            creationTime: e.creationTime,
+            lastUsedTime: e.lastUsedTime,
+        }));
+    }
+
+    async addSSHPublicKey(ctx: TraceContext, value: SSHPublicKeyValue): Promise<UserSSHPublicKeyValue> {
+        const user = this.checkUser("addSSHPublicKey");
+        const data = await this.userDB.addSSHPublicKey(user.id, value);
+        this.updateSSHKeysForRegularRunningInstances(ctx, user.id);
```
jankeromnes, line 1955: Here, you could show that you're intentionally not waiting for this to finish by adding a comment like so:

```suggestion
        /** no await */ this.updateSSHKeysForRegularRunningInstances(ctx, user.id);
```

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -1929,6 +1932,61 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
...
+    }
+
+    async deleteSSHPublicKey(ctx: TraceContext, id: string): Promise<void> {
+        const user = this.checkUser("deleteSSHPublicKey");
+        await this.userDB.deleteSSHPublicKey(user.id, id);
+        this.updateSSHKeysForRegularRunningInstances(ctx, user.id);
+        return;
+    }
+
+    async updateSSHKeysForRegularRunningInstances(ctx: TraceContext, userId: string) {
```
jankeromnes, line 1972: This should not be publicly exposed, right?

```suggestion
    protected async updateSSHKeysForRegularRunningInstances(ctx: TraceContext, userId: string) {
```

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
mustard-mh: Hey @gtsiolis, do we need a design fot this? or it's good for us to use the similar style with image below from https://www.gitpod.io/ for switch?

|Switch| Button Style|
|-|-|
|image |image|

components/gitpod-protocol/src/protocol.ts
mustard-mh: I switch to GH style fingerprint, is it good for you? @gtsiolis

components/gitpod-db/src/typeorm/transformer.ts
mustard-mh: The main reason is that other DBs in the repo are set up like this🤣

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
iQQBot: https://github.com/gitpod-io/gitpod/pull/10704 already merged

components/server/src/workspace/gitpod-server-impl.ts
mustard-mh: `tslint` print this error, so I change it to. [source](https://stackoverflow.com/a/52620074)
```ts
/** no await */ void this.updateSSHKeysForRegularRunningInstances(ctx, user.id);
```

SCR-20220621-14p

mustard-mh:
This PR is ready for review, UX improvement can be another PR if geropl want to improve it.

Could you review again 🙏 ? cc @iQQBot @jankeromnes @loujaybee

mustard-mh:
Upvote.

I hate to delete Gitpod related content in `~/.ssh/known_hosts`  too
- When I want to see the confirm prompt
- Or want to clean the output message (The host key is known ....) shown in @iQQBot 's screen capture )

I need to find the correct one, or search `vim regex delete line` every time

akosyakov:
> We can change only the dashboard copy/paste part, without changing the style of the vscode desktop extensio

I guess it is alright only for copy/paste, but let's do it in another PR. That additional discussions does not delay this one.

components/gitpod-db/src/tables.ts
```diff
@@ -268,6 +268,11 @@ export class GitpodTableDescriptionProvider implements TableDescriptionProvider
             deletionColumn: "deleted",
             timeColumn: "_lastModified",
         },
+        {
+            name: "d_b_user_ssh_public_key",
+            primaryKeys: ["id"],
+            timeColumn: "_lastModified",
```
geropl, line 274: @mustard-mh I think we need `deletionColumn: "deleted",` here.
Sorry for the double-entry (cmp. `deleted-entry-gc.ts`). :cry: This is debt we need to get rid of...

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -1929,6 +1932,61 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
...
+            fingerprint: data.fingerprint,
+            creationTime: data.creationTime,
+            lastUsedTime: data.lastUsedTime,
+        };
+    }
+
+    async deleteSSHPublicKey(ctx: TraceContext, id: string): Promise<void> {
+        const user = this.checkUser("deleteSSHPublicKey");
+        await this.userDB.deleteSSHPublicKey(user.id, id);
+        /** no await */ void this.updateSSHKeysForRegularRunningInstances(ctx, user.id);
```
geropl, line 1968: nit: `void` does not have any meaning, here.

If we do not `await`, we still need to handle errors with `.catch(log.error)`, for example, to avoid `UnhandledPromiseRejection`s. See other cases in this file as well.

components/gitpod-db/src/typeorm/transformer.ts
geropl: I got three comments:
 1. `null` is the wrong value to return. Judging from the interface, [we expect](https://github.com/gitpod-io/gitpod/pull/10573/files#diff-957d9dd15b947e3969539ebd0dabcb8b40ce8b5564e5820cd9bb27d88be1dbc7R369) `undefined` here.
 2. the transformer is "asymmetric" (from returns `null`, to returns `undefined`). Which leads to weird bugs if you do write-read-write cycles. Thus please do stick to 3. :point_down: 
 3. the default pattern we use for "nullable" timestamp columns that is know to work correctly is this:
   1. string column type, NOT nullable, default to "" (empty string)
   2. use MAP_EMPTY_STR_TO_UNDEFINED ([example](https://github.com/gitpod-io/gitpod/blob/1420b9ca546bcc7bb140feb9bc201dcf24acf318/components/gitpod-db/src/typeorm/entity/db-workspace.ts#L88-L92))
 
 I know that we do it differently in other occasions. But if you ask me, because the underlying tech (mysql) is so unsatisfying, we should try hard to stick to one pattern how to work around it. and 3. is the way we recommend. :lotus_position:

geropl (changes requested):
@mustard-mh Sorry to hold up the PR, but I have some comments I'd love to address beforehand. Most specifically [DB related](https://github.com/gitpod-io/gitpod/pull/10573/files#r905772075).

I'm happy to chat in sync or Slack if you like to solve it quickly! :+1:

components/gitpod-db/src/typeorm/user-db-impl.ts
```diff
@@ -395,6 +402,43 @@ export class TypeORMUserDBImpl implements UserDB {
...
+            name: value.name,
+            key: value.key,
+            creationTime: new Date().toISOString(),
+            deleted: false,
+        });
+    }
+
+    public async deleteSSHPublicKey(userId: string, id: string): Promise<void> {
+        const repo = await this.getSSHPublicKeyRepo();
+        await repo.update({ userId, id }, { deleted: true });
```
geropl, line 439: This won't necessarily hit the PK index. Why not use `{ id }` alone? That should guarantee this to work as fast as possible. :+1:

components/gitpod-db/src/typeorm/user-db-impl.ts
```diff
@@ -395,6 +402,43 @@ export class TypeORMUserDBImpl implements UserDB {
...
     }
 
+    public async hasSSHPublicKey(userId: string): Promise<boolean> {
+        const repo = await this.getSSHPublicKeyRepo();
+        return !!(await repo.findOne({ where: { userId, deleted: false } }));
+    }
+
+    public async getSSHPublicKeys(userId: string): Promise<UserSSHPublicKey[]> {
+        const repo = await this.getSSHPublicKeyRepo();
+        return repo.find({ where: { userId, deleted: false }, order: { creationTime: "ASC" } });
```
geropl, line 412: Note that this query is not backed 100% by an index. It should still work, because:
 1. userId is indexed, and has the highest cardinality, so the query engine should go for this first
 2. creationtime has a low cardinality
 3. deleted has a low cardinality

components/gitpod-db/src/typeorm/user-db-impl.ts
mustard-mh: To make sure this public key belongs to the user

components/gitpod-db/src/typeorm/user-db-impl.ts
mustard-mh: Is it means we should refer an index here to find?

geropl: I think it's fine as-is, just wanted to share my thoughts. :+1:

components/gitpod-db/src/typeorm/user-db-impl.ts
geropl: > To make sure this public key belongs to the user

Ah, got it. :+1:

components/gitpod-db/src/tables.ts
mustard-mh: I use `env_var` as my baseline to change db related things, as we can see we lose a lot of `db-sync` feature (env_var have no `deletionColumn` too), not sure if it'll affect our data

https://github.com/gitpod-io/gitpod/blob/3c327432c8a22b93f3b3d46193bb1a2c84107a3a/components/gitpod-db/src/tables.ts#L171-L175

mustard-mh: We should properly change this file and baseline file [internal chat](https://gitpod.slack.com/archives/C02EN94AEPL/p1654843498128069?thread_ts=1654842003.970989&cid=C02EN94AEPL)

mustard-mh:
@geropl  I changed db and server, you can see diff [here](https://github.com/gitpod-io/gitpod/compare/1455d9844de4f5782f5f8c74fa38b9a71a98ee11..6edd014107dd302354b25b30a52e227cf8a4f7de)

components/gitpod-db/src/tables.ts
geropl: @AlexTugarev Can you verify my comment above? :point_up:

geropl: Synced with Alex, he confirmed. :+1: Will approve.

geropl (approved):
Nice, LGTM! :+1:

akosyakov:
@mustard-mh wanted to check out UI, but werft fails

mustard-mh:
Resolved merge conflict

gtsiolis:
@mustard-mh Could you clean deploy another preview environment or delete some users because we've reached the limit of 10 users and no new users can sign up.

mustard-mh:
> @mustard-mh Could you clean deploy another preview environment or delete some users because we've reached the limit of 10 users and no new users can sign up.

@gtsiolis Rebased to main again, since some commits may fix this problem (db have only two builtin users)

components/dashboard/src/settings/SSHKeys.tsx
```diff
@@ -0,0 +1,258 @@
...
+                    <Alert type="error" className="mb-2">
+                        {errorMsg}
+                    </Alert>
+                )}
+            </>
+            <div className="text-gray-500 dark:text-gray-400 text-md">
+                Add an SSH key for secure access workspaces via SSH.{" "}
+                <a href="/docs/configure/ssh" target="gitpod-ssh-doc" className="gp-link">
+                    Learn more
+                </a>
```
gtsiolis, line 85: **thought**: Do we need this help link here? Asking because we also link below to the same page.

mustard-mh:
works now @gtsiolis

components/gitpod-db/src/typeorm/user-db-impl.ts
```diff
@@ -395,6 +402,43 @@ export class TypeORMUserDBImpl implements UserDB {
...
+    }
+
+    public async addSSHPublicKey(userId: string, value: SSHPublicKeyValue): Promise<UserSSHPublicKey> {
+        const repo = await this.getSSHPublicKeyRepo();
+        const fingerprint = SSHPublicKeyValue.getFingerprint(value);
+        const allKeys = await repo.find({ where: { userId, deleted: false } });
+        const prevOne = allKeys.find((e) => e.fingerprint === fingerprint);
+        if (!!prevOne) {
+            throw new Error(`Duplicate public key with ${prevOne.name}`);
```
gtsiolis, line 421: **nitpick**: What do you think of making this easier to understand.

```suggestion
            throw new Error(`Key already in use);
```

components/gitpod-protocol/src/protocol.ts
```diff
@@ -357,6 +357,68 @@ export namespace UserEnvVar {
...
+        if (value.name.length > 255) {
+            return "Title too long. Maximum value length is 255 characters.";
+        }
+        if (value.key.length === 0) {
+            return "Key must not be empty.";
+        }
+        try {
+            getData(value);
+        } catch (e) {
+            return "Key is invalid. You must supply a key in OpenSSH public key format.";
```
gtsiolis, line 389: **issue(non-blocking)**: While we catch some cases, we still allow users adding invalid keys. Is this known? If needed, let's open a follow-up issue to resolve this.

components/dashboard/src/settings/SSHKeys.tsx
```diff
@@ -0,0 +1,258 @@
...
+            )}
+        </PageWithSubMenu>
+    );
+}
+
+function KeyItem(props: { sshKey: UserSSHPublicKeyValue }) {
+    const key = props.sshKey;
+    return (
+        <ItemField className="flex flex-col gap-y box-border overflow-hidden">
+            <p className="truncate text-gray-400 dark:text-gray-600">SHA256:{key.fingerprint}</p>
```
gtsiolis, line 250: **suggestion**: Since we're never going to have space to print the whole value here, shall we trim the text here and show only for example 16 characters or somehow encode this long string?

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
```diff
@@ -55,40 +57,80 @@ interface SSHProps {
...
+                    }}
+                />
+                <TabMenuItem
+                    key="access_token"
+                    name="Access Token"
+                    selected={!selectSSHKey}
+                    onClick={() => {
+                        setSelectSSHKey(false);
+                    }}
+                />
```
gtsiolis, line 94: **question**: Do we want to keep the access token option around? I thought this caused more confusion than helped. Maybe it's worth removing this altogether and instead nudge users to add an SSH key when they try to connect via SSH and there's no SSH key in their account. Cc @akosyakov @loujaybee

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
```diff
@@ -55,40 +57,80 @@ interface SSHProps {
...
-                            target="_blank"
-                            rel="noopener noreferrer"
-                            className="gp-link"
-                        >
-                            ssh-keygen
+                    </Alert>
+                )}
+                {!hasSSHKey && selectSSHKey && (
+                    <Alert type="warning" className="whitespace-normal">
+                        You don't have any public SSH keys in your Gitpod account. You can{" "}
```
gtsiolis, line 106: **praise**: Nice! I came back to this PR to make sure we have this warning here when needed. 👏

gtsiolis, line 106: **issue**: When you don't have an SSH key in place, you still see an SSH command below to connect, which asks for a password. Should we drop (hide) the SSH command below?

gtsiolis (commented):
Looks great, @mustard-mh! 🔮 

Took another look at this and left a few more comments below. 👀

components/dashboard/src/settings/SSHKeys.tsx
iQQBot: I think we can keep that, they are different part link

components/dashboard/src/settings/SSHKeys.tsx
iQQBot: These long strings have actually been encoded, and such characters are also used in ssh debug, so we can compare them to determine if they are the same, although the first dozen or so characters are basically the same

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
iQQBot: If this PR got deployed https://github.com/gitpod-io/gitpod/pull/10704 , using access token will become a very convenient practice, if they just want to temp connect workspace, they don't need think about ssh side things

akosyakov: I think it is alright to have, but ssh key should be always default.

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
akosyakov: If we should allow to copy it always. If a user adds keys following a warning then they can use it immediately. Otherwise they need to reopen the dialog

akosyakov:
UI looks good, but then I try first time user flow I cannot connect:
- i start a workspace
- then go to dashboard and select connect to SSH
- it tells me that I don't have public keys
- i generate a new key on my machine under `~/.ssh/test_rsa`
- i go to setting and add this key
- i go back to dialog, copy a command, paste it with `-i ~/.ssh/test_rsa`
- it asks me for password

Is it because newly installed keys are not automatically propagated to workspaces?

akosyakov:
@iQQBot I did everything again and it worked now. So probably some mistake on my side.

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
gtsiolis: > I think it is alright to have, but ssh key should be always default.

Fair point, let's keep it there then. 🤝 

If there's no reason to keep access tokens around, I'd vote for removing this option to limit the amount of options presented to users, as it can cause more confusion than value.

In case we keep this, does it make sense to drop the command printed when there are no keys in place?

Screenshot 2022-06-30 at 12 38 19 AM

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
gtsiolis: > If we should allow to copy it always. If a user adds keys following a warning then they can use it immediately. Otherwise they need to reopen the dialog

Ah, I see. Then the warning dialog becomes obsolete or no longer relevant after adding a key, right?

Maybe it's not worth doing so, but what do you think of rephrasing the copy then to:

##### BEFORE 
> The following shell command can be used to SSH into this workspace with a [ssh key](https://hw-ssh-keys-s.preview.gitpod-dev.com/keys).

##### AFTER
> The following shell command can be used to SSH into this workspace with a [ssh key](https://hw-ssh-keys-s.preview.gitpod-dev.com/keys), once you add an SSH key.

Otherwise, let's leave this as is and ignore the comment above in https://github.com/gitpod-io/gitpod/pull/10573#discussion_r910434472. ✔️

components/dashboard/src/settings/SSHKeys.tsx
gtsiolis: > ... although the first dozen or so characters are basically the same

Could we then trim the first characters and show only the last dozens ones? Makes sens or no?

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
akosyakov: I'm fine with both. I think they will read a warning when prompted for a password.

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
akosyakov: > In case we keep this, does it make sense to drop the command printed when there are no keys in place?

answered [here ](https://github.com/gitpod-io/gitpod/pull/10573#discussion_r909161251) right?

components/dashboard/src/workspaces/ConnectToSSHModal.tsx
gtsiolis: DEAL—Let's go with any option for now.

components/dashboard/src/settings/SSHKeys.tsx
gtsiolis: To clarify, the reason of asking this is because the encoded string is sometimes not visible to users in smaller viewports like a regular 13" screen. Although in this case, the information is not quite interesting or helpful, it's generally good for users to be able to see the whole information when possible.

| Zoom 100% | Zoom 90% |
|-|-|
| Screenshot 2022-06-30 at 1 17 33 PM | Screenshot 2022-06-30 at 1 17 39 PM |
 
However, I see that the string is always (?) visible on the delete confirmation modal so let's leave this as is.

| Delete Key (1) | Delete Key (2) |
|-|-|
| Screenshot 2022-06-30 at 1 17 52 PM | Screenshot 2022-06-30 at 1 18 04 PM |

Thanks, @iQQBot! Resolving this thread. 🤝

Issue #10533, opened 2022-06-08
[Usage-based] Attribute workspace instances on start


jankeromnes:
## Is your feature request related to a problem? Please describe

In the context of usage-based billing (#9036), we've identified that it's important to immediately "attribute" a workspace instance to a project, and to a "customer" (i.e. a Team or a User) when the instance starts.

This cannot be done asynchronously / later on, because any of these relationships from instance to project/customer can change at any time:
- Workspace (can be deleted)
- Project (can be deleted, can possibly change ownership in the future?)
- User (can be deleted, can join/leave teams, can upgrade/cancel)
- Team (can be deleted, can upgrade/cancel)

Thus, it's important to determine and persist the attributed final customer/account every time a workspace instance is started.

## Describe the behaviour you'd like

Attribution can be determined in various ways:
- Workspace Instance → Workspace → Project? → Owner(User/Team)
- Workspace Instance → Workspace → Owner(User) (→ Team?)

Once determined on instance start, this information could be persisted:
- As new fields in `d_b_workspace_instance` (e.g. `projectId?`, plus `userId?` and/or `teamId?`)
- In a new database (e.g. a DB reserved for Usage accounting)

jankeromnes:
Another realization, the "attribution" information is more complex that just one field.

A workspace instance needs to be attributed to:
- A `projectId` (so that we can compute project usage)
- A `userId` (so that we know who created the instance, and the user can possibly be the final "customer" as well)
- A `teamId` (if a team is the "cost center" for this workspace instance, e.g. the team has enabled usage-based billing and the user has selected this team as their "cost center")
- Possibly, the user's or the team's Stripe `customerId` if there is one (if the usage is billed, i.e. not free credits, then this is the `customerId` to which we bill the usage -- can be a team or user, and can change if users/teams change their billing or get deleted)

easyCZ:
I believe that the attribution must work as an immutable record at the start of the workspace. We mostly need to know the unique ID (if a team, or individual) who has promised to pay for this workspace. 

This information should be unrelated to Stripe as it would couple us to Stripe.

[Related (internal) conversation.](https://gitpod.slack.com/archives/C02EN94AEPL/p1654668705113199)

jankeromnes:
Great feedback from @svenefftinge -- rather than attributing to a Team (and defaulting to the owner User if there is no attributed Team), it would nicer to create a dedicated "Cost Center" object.

[...]

PR #10574, opened 2022-06-10
[Usage-based] Attribute workspace instances to a Team on start

jankeromnes:
## Description

When a workspace instance starts, we want to determine and persist which "cost center" that instance's usage should be attributed to:
- If `wsi.attributedTeamId` is set to a specific Team ID, then that Team "owns" the instance's usage
  - This means that the instance's usage time will be reflected in the Team's Usage dashboard, and in their Stripe invoices if usage-based billing is enabled
- If `wsi.attributedTeamId` is not set, then it's the workspace owner (User) who owns the instance's usage
  - This means that we default to user billing or their free credits when no Team is attributed

Reminder: We need to do this on instance start, because the usual attribution links can change over time.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/10533

## How to test

1. Should build
2. In the DB, `d_b_workspace_instance` should have a `attributedTeamId` column that defaults to empty string (`''`)
3. If you start a workspace **without any Project**:
    - its `attributedTeamId` should be empty (i.e. usage is attributed to you)
4. If you create a Project under your User, and start a workspace for that **User Project**:
    - the `attributedTeamId` should also be empty (i.e. usage still attributed to you)
5. If you create a Project under a Team, and then start a Workspace for that **Team Project**:
    - the `attributedTeamId` should now be set to the Team's ID (i.e. usage is attributed to the team)

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/gitpod-protocol/src/workspace-instance.ts
+++ b/components/gitpod-protocol/src/workspace-instance.ts
@@ -62,6 +62,13 @@ export interface WorkspaceInstance {
      * resources that are provided to the workspace.
      */
     workspaceClass?: string;
+
+    /**
+     * Identifies the team to which this instance's runtime should be attributed to
+     * (e.g. for usage analytics or billing purposes).
+     * If unset, the usage should be attributed to the workspace's owner (ws.ownerId).
+     */
+    attributedTeamId?: string;
 }
 
 // WorkspaceInstanceStatus describes the current state of a workspace instance
--- a/components/server/src/user/user-service.ts
+++ b/components/server/src/user/user-service.ts
@@ -16,7 +16,7 @@ import {
     WORKSPACE_TIMEOUT_EXTENDED,
     WORKSPACE_TIMEOUT_EXTENDED_ALT,
 } from "@gitpod/gitpod-protocol";
-import { TermsAcceptanceDB, UserDB } from "@gitpod/gitpod-db/lib";
+import { ProjectDB, TermsAcceptanceDB, UserDB } from "@gitpod/gitpod-db/lib";
 import { HostContextProvider } from "../auth/host-context-provider";
 import { log } from "@gitpod/gitpod-protocol/lib/util/logging";
 import { Config } from "../config";
@@ -63,6 +63,7 @@ export class UserService {
     @inject(Config) protected readonly config: Config;
     @inject(TermsAcceptanceDB) protected readonly termsAcceptanceDb: TermsAcceptanceDB;
     @inject(TermsProvider) protected readonly termsProvider: TermsProvider;
+    @inject(ProjectDB) protected readonly projectDb: ProjectDB;
 
     /**
      * Takes strings in the form of <authHost>/<authName> and returns the matching User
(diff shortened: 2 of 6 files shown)


jankeromnes:
TODO:
- [ ] In workspace starter, call `stripeService` to find a team which has a Stripe Customer ID?

components/server/src/user/user-service.ts
```diff
@@ -205,6 +206,28 @@ export class UserService {
...
 
+    /**
+     * Identifies the team to which a workspace instance's running time should be attributed to
+     * (e.g. for usage analytics or billing purposes).
+     * If no specific team is identified, the usage will be attributed to the user instead (default).
+     *
+     * @param user
+     * @param projectId
+     */
+    async getWorkspaceUsageAttributionTeamId(user: User, projectId?: string): Promise<string | undefined> {
```
jankeromnes, line 217: Placing this method in `UserService` is debatable. For example, this logic could be in-lined in `workspace-starter.ts` (which already has `projectDb` injected).

My reason for placing it here was to allow an extension point where ee/UserService could detect if payment is enabled, and use a different logic (i.e. attribute all usage to a "cost center" team, instead of the project's team).

components/gitpod-protocol/src/workspace-instance.ts
```diff
@@ -62,6 +62,13 @@ export interface WorkspaceInstance {
      * resources that are provided to the workspace.
      */
     workspaceClass?: string;
+
+    /**
+     * Identifies the team to which this instance's runtime should be attributed to
+     * (e.g. for usage analytics or billing purposes).
+     * If unset, the usage should be attributed to the workspace's owner (ws.ownerId).
+     */
+    attributedTeamId?: string;
```
andrew-farries, line 71: Rather than leaving this unset in this case, should we explicitly fill in the ownerId?

andrew-farries:
Tested this myself and saw the expected teamIds 👍

components/gitpod-protocol/src/workspace-instance.ts
jankeromnes: I believe leaving it unset makes more sense:
- It can only be a `teamId` (the billing is either delegated to a specific team, or it falls back to you -- delegating to a different user does not make sense, right? 🤔)
- Setting it to possibly a `userId` makes the design more complex (need a second column `attributedUserId`, or maybe a prefix to the ID e.g. `team-1324` vs `user-1234`)
- The team attribute link goes through instance > workspace > project > team (any of the links can change so persisting the endpoint makes sense). User attribution is more direct (instance > workspace > owner) -- I don't think we can even get to a state where we don't know which user an instance belongs to.

andrew-farries: > delegating to a different user would not make sense, right?
Yes, that's my understanding too.

components/server/src/user/user-service.ts
andrew-farries: Not strongly opinionated here either but your reasoning here makes sense; the subject of cost center vs team has already come up in discussions.

andrew-farries (approved):

jankeromnes:
Hmm, I just realized how similar this is to https://github.com/gitpod-io/gitpod/pull/10632 🤔

My initial reflex was to revert my PR again -- but then I realized my PR will allow to re-attribute usage to a different team (e.g. if you have a team with usage-based billing that wants to handle all your workspace usage regardless of project/team). So, maybe the two are actually complementary (and when we re-attribute, we might have to update the workspace metadata too 💡)

PR #10583, opened 2022-06-10
[usage] Fix report generation

easyCZ:
## Description

My testing dataset was not large enough, and we'd run into int64 overflows for very long running instances. Using Seconds as the baseline unit fixes that.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
```


roboquat:

<details>

</details>

components/usage/pkg/controller/reconciler.go
```diff
@@ -131,16 +131,16 @@ func (u *UsageReconciler) ReconcileTimeRange(ctx context.Context, from, to time.
 func generateUsageReport(teams []teamWithWorkspaces, maxStopTime time.Time) ([]TeamUsage, error) {
 	var report []TeamUsage
 	for _, team := range teams {
-		var teamTotalRuntime time.Duration
+		var teamTotalRuntime float64
```
jankeromnes, line 134: Why the overkill floating-point precision on seconds? 🤔 Rounded-up seconds would probably be precise enough, right?

jankeromnes: Credits are likely to map to hours anyway. I think rounding to the next second is extremely fine. 👍

easyCZ: Let's do that in a follow-up PR.

jankeromnes (approved):
Mmf, approving to unblock progress, but I really don't think we want `float64` precision on our seconds.

Please switch to rounding runtimes to the next second in a follow-up PR. 🙏

components/usage/pkg/db/workspace_instance.go
```diff
@@ -42,16 +42,16 @@ type WorkspaceInstance struct {
...
-// TotalRuntime computes how long this WorkspaceInstance has been running.
+// WorkspaceRuntimeSeconds computes how long this WorkspaceInstance has been running.
 // If the instance is still running (no stop time set), maxStopTime is used to to compute the duration - this is an upper bound on stop
-func (i *WorkspaceInstance) TotalRuntime(maxStopTime time.Time) time.Duration {
+func (i *WorkspaceInstance) WorkspaceRuntimeSeconds(maxStopTime time.Time) float64 {
 	start := i.StartedTime.Time()
 	stop := maxStopTime
 	if i.StoppedTime.IsSet() {
 		stop = i.StoppedTime.Time()
 	}
```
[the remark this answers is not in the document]
easyCZ: The trimming already happens [here](https://github.com/gitpod-io/gitpod/blob/e3f9ca58586af53c66f9afa6acc44710d94d7f9e/components/usage/pkg/controller/reconciler.go#L284), after the workspace instances are loaded and validated. 

Here, it is a general purpose method which works on the data on the struct, the trimming is specific to the usage report for billing and therefore done there.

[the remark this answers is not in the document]
jankeromnes: Aha, cool! 🎯

https://github.com/gitpod-io/gitpod/blob/e3f9ca58586af53c66f9afa6acc44710d94d7f9e/components/usage/pkg/controller/reconciler.go#L288-L294

I guess these should actually be called `minimumStart` and `maximumStop` 😅 but the logic looks correct. 👍

easyCZ:
Follow-up in https://github.com/gitpod-io/gitpod/pull/10591

PR #10591, opened 2022-06-10
[usage] Round workspace runtime

easyCZ:
## Description

As requested in https://github.com/gitpod-io/gitpod/pull/10583#pullrequestreview-1002851458.

## Related Issue(s)

Fixes #

## How to test

Unit tests

## Release Notes

```release-note
NONE
```

## Documentation

NONE

### DIFF
--- a/components/usage/pkg/controller/reconciler.go
+++ b/components/usage/pkg/controller/reconciler.go
@@ -131,7 +131,7 @@ func (u *UsageReconciler) ReconcileTimeRange(ctx context.Context, from, to time.
 func generateUsageReport(teams []teamWithWorkspaces, maxStopTime time.Time) ([]TeamUsage, error) {
 	var report []TeamUsage
 	for _, team := range teams {
-		var teamTotalRuntime float64
+		var teamTotalRuntime int64
 		for _, workspace := range team.Workspaces {
 			for _, instance := range workspace.Instances {
 				teamTotalRuntime += instance.WorkspaceRuntimeSeconds(maxStopTime)
@@ -312,6 +312,6 @@ func toSet(items []string) []string {
 }
 
 type TeamUsage struct {
-	TeamID           string  `json:"team_id"`
-	WorkspaceSeconds float64 `json:"workspace_seconds"`
+	TeamID           string `json:"team_id"`
+	WorkspaceSeconds int64  `json:"workspace_seconds"`
 }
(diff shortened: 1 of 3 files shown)


easyCZ:
I personally think that rounding at this layer is losing precision for no good reason. The rounding can always happen before it's inserted into the DB, or before it's synced into the stripe but the report should remain as accurate as possible. It is an intermediate representation.

components/usage/pkg/controller/reconciler.go
```diff
@@ -131,7 +131,7 @@ func (u *UsageReconciler) ReconcileTimeRange(ctx context.Context, from, to time.
 func generateUsageReport(teams []teamWithWorkspaces, maxStopTime time.Time) ([]TeamUsage, error) {
 	var report []TeamUsage
 	for _, team := range teams {
-		var teamTotalRuntime float64
+		var teamTotalRuntime int64
```
jankeromnes, line 134: Thought: This could even be `uint64` right? (I don't think we want to support "negative usage time", so we can double our max representable value 🙌)

jankeromnes (approved):
Thanks again!

> I personally think that rounding at this layer is losing precision for no good reason. The rounding can always happen before it's inserted into the DB, or before it's synced into stripe but the report should remain as accurate as possible. It is an intermediate representation.

Thanks for voicing these concerns. I do understand your point, however I still believe that the extra readability and sanity of integer values for seconds outweighs any benefits given by sub-second precision. Still, if we really do want sub-second precision one day, I think it will be easy to adjust. 😊

Approving, but holding due to one optional suggestion.

/hold

components/usage/pkg/controller/reconciler.go
geropl: If this really is only about "usage"/"runtime", I agree. If we want to represent "credits" here (which I'm not sure we want/should..?) then int64 it is.
Another thought: Not sure how (type) safe `uint64`/`int64` interactions are in Go: Is it easy to mess it up? If yes, that might be a good argument too keep one type everywhere (most likely `in64` if we need the "positive" side)

geropl:
> I do understand your point, however I still believe that the extra readability and sanity of integer values for seconds outweighs any benefits given by sub-second precision

My 2 cents:
 - we should definitely store time durations in integer values (upvoting the `uint64` [comment](https://github.com/gitpod-io/gitpod/pull/10591#discussion_r895514178) above!)
 - whether we want to store `ms` vs `s`:
   - for precisions/consistencies sake it's sometimes beneficial to keep whatever we get as input (ms here), e.g. if we need to be able to exactly identify time-range in the opposite direction (usage -> workspace instance)
   - but as all clients are not interested in anything below `s` precision, I think it's a great way to keep things simple, and use the same datatype everywhere :lotus_position:

jankeromnes:
Since Milan is on holidays, let's merge this as is and address the review comments in follow-up PRs.

/unhold

PR #10594, opened 2022-06-10
Use `xterm-256color` when creating terminal

jeanp413:
## Description

Use `xterm-256color` when creating pty terminal

## How to test

1. Run `echo $TERM` in a  task terminal it should output `xterm-256color`

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/supervisor/pkg/terminal/service.go
+++ b/components/supervisor/pkg/terminal/service.go
@@ -99,7 +99,7 @@ func (srv *MuxTerminalService) OpenWithOptions(ctx context.Context, req *api.Ope
 	if cmd.Dir == "" {
 		cmd.Dir = srv.DefaultWorkdir
 	}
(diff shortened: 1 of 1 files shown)


jeanp413:
I don't see any difference but It would be good to use the same value as vscode nonetheless
image

filiptronicek:
@jeanp413 is there any difference, or is it maybe just an alias? Is there a different amount of colors supported with `xterm-256color`? I was trying to find out the difference but was unlucky.

I agree that it's good to have it like VS Code does.

mustard-mh (approved):
Tested good.

It's quite different in MacOS for both browser and desktop, 

maybe you can test with `msgcat` too, cc @filiptronicek @jeanp413 

```
sudo apt-get update
sudo apt-get install -y gettext
```

image

jeanp413:
Cool, thanks @mustard-mh :rocket:

PR #10595, opened 2022-06-10
When using the Latest Release of JetBrains IDEs, if the workspace has tasks defined on .gitpod.yml, the IDE will start with one terminal opened for each task

### LINKED ISSUE #10579: Integrate multiple terminals into JetBrains IDE's
Currently, JetBrains doesn't read the commands/tasks functionality of Gitpod. 

This can be worked around by using `gp tasks` for now, but ideally this is a first-class feature in JetBrains. 

There is an issue to resolve for JetBrains https://youtrack.jetbrains.com/issue/CWM-6093 but this is reported now as fixed.

felladrin:
## Description

When using the Latest Release of JetBrains IDEs, if the workspace has tasks defined on .gitpod.yml, the IDE will start with one terminal opened for each task, behaving similar to VS Code on Gitpod.

When using the Stable Release the behavior remains the current one: one terminal will be displayed, with a welcome message and invoking `gp tasks attach`.

This separation of versions was required for backward compatibility, as there were breaking changes on the API between IntelliJ 2022.1 and 2022.2. Also, the fix for  [Shared Terminals Losing Reference](https://youtrack.jetbrains.com/issue/CWM-6093) was applied only to v2022.2.

## Known limitations

- It's not possible to resize the terminal (max line width) from Thin Client. Resize happens only if the terminal width is changed on the Backend IDE. [[1](https://youtrack.jetbrains.com/issue/CWM-5410/LINES-env-var-doesnt-update-when-terminal-changes-size)][[2](https://youtrack.jetbrains.com/issue/CWM-5248/Soft-wrap-doesnt-work-in-Terminal)][[3](https://youtrack.jetbrains.com/issue/CWM-5456)][[4](https://youtrack.jetbrains.com/issue/CWM-4314)]
- If we have more than two terminals opened at once, only two of them will be displaying correctly their title in the tab, while the other will be displaying "Local" until the user focuses on them - at this moment the info about the terminal tab is fetched and the tab title is updated.  Seems to be a limitation from the backend. I'm going to file an issue for JetBrains and leave the link here later.
- It's not possible to split shared terminals. This is not on plans to be supported, at this moment.
- When we stop a workspace, restart, and reconnect to it, if there were terminals running, they will show only the last few lines of the history. The number of history lines displayed is currently hardcoded on JetBrains RemoteDev. The only way to see the full history is by looking at the terminals from the Backend IDE.

## To be done in a follow-up PR
- Update the terminal title whenever there's a change on the title running under Supervisor
- Close the IDE Terminal tab if Gitpod Terminal was closed remotely and emit a 'close' signal for Gitpod Terminal if user clicks the "X" button to close the IDE Terminal.

## Related Issue(s)

Part of #10579

## How to test

- Make sure you're using the latest version from JetBrains Gateway EAP (v222.3048.1 or later). You can download the latest version at https://www.jetbrains.com/remote-development/gateway/ 
- Open, on the Preview Environment of this PR, a new workspace that contains tasks defined on .gitpod.yml.  Suggestion: Gitpod Repository, by clicking this URL: https://felladrin-ed4d29747a.preview.gitpod-dev.com/#referrer:jetbrains-gateway:intellij/https://github.com/gitpod-io/gitpod
- As by default the Stable version of the IDE is used, when the workspace starts, you should see the usual terminal with a welcome message:
  image
- Having confirmed that the Stable version of JetBrains IDEs are still using the current implementation of terminals, you can close that workspace.
- Now, open [Preferences on the Preview Environment of this PR](https://felladrin-ed4d29747a.preview.gitpod-dev.com/preferences) and select _IntelliJ_ and mark _Latest Release_ to use the EAP version of the IDE.
  image
- Open, on the Preview Environment of this PR, a new workspace, with the same repository as the previous step.
- When opening it, you should see 3 terminals opened (one for each Gitpod Task) and no more terminal displaying a welcome message:
  image

## Release Notes

```release-note
When using the Latest Release of JetBrains IDEs, if the workspace has tasks defined on .gitpod.yml, the IDE will start with one terminal opened for each task, behaving similar to VS Code on Gitpod.
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -46,7 +46,7 @@ tasks:
         read -r -p "Press enter to continue Java gradle task"
       fi
       leeway exec --package components/supervisor-api/java:lib --package components/gitpod-protocol/java:lib -- ./gradlew --build-cache build
-      leeway exec --package components/ide/jetbrains/backend-plugin:plugin --package components/ide/jetbrains/gateway-plugin:publish --parallel -- ./gradlew --build-cache buildPlugin
+      leeway exec --package components/ide/jetbrains/backend-plugin:plugin-latest --package components/ide/jetbrains/gateway-plugin:publish --parallel -- ./gradlew --build-cache buildPlugin
   - name: TypeScript
     before: scripts/branch-namespace.sh
     init: yarn --network-timeout 100000 && yarn build
--- a/components/ide/jetbrains/backend-plugin/.gitignore
+++ b/components/ide/jetbrains/backend-plugin/.gitignore
@@ -3,3 +3,4 @@
 .vscode
 bin
 build
+gradle-local.properties
--- a/components/ide/jetbrains/backend-plugin/BUILD.yaml
+++ b/components/ide/jetbrains/backend-plugin/BUILD.yaml
@@ -1,5 +1,5 @@
 packages:
-  - name: plugin
+  - name: plugin-stable
     type: generic
     deps:
       - components/supervisor-api/java:lib
@@ -8,11 +8,32 @@ packages:
       - "**/*.kt"
       - "build.gradle.kts"
       - "gradle.properties"
+      - "gradle-stable.properties"
       - "gradle/wrapper/*"
       - "gradlew"
       - "settings.gradle.kts"
       - "src/main/resources/*"
+      - "src/main/resources-stable/*"
     config:
       commands:
-        - ["./gradlew", "-PsupervisorApiProjectPath=components-supervisor-api-java--lib/", "-PgitpodProtocolProjectPath=components-gitpod-protocol-java--lib/", "runPluginVerifier"]
-        - ["./gradlew", "-PsupervisorApiProjectPath=components-supervisor-api-java--lib/", "-PgitpodProtocolProjectPath=components-gitpod-protocol-java--lib/", "buildPlugin"]
+        - ["./gradlew", "-PsupervisorApiProjectPath=components-supervisor-api-java--lib/", "-PgitpodProtocolProjectPath=components-gitpod-protocol-java--lib/", "-PenvironmentName=stable", "runPluginVerifier"]
+        - ["./gradlew", "-PsupervisorApiProjectPath=components-supervisor-api-java--lib/", "-PgitpodProtocolProjectPath=components-gitpod-protocol-java--lib/", "-PenvironmentName=stable", "buildPlugin"]
+  - name: plugin-latest
+    type: generic
+    deps:
+      - components/supervisor-api/java:lib
+      - components/gitpod-protocol/java:lib
+    srcs:
+      - "**/*.kt"
+      - "build.gradle.kts"
+      - "gradle.properties"
+      - "gradle-latest.properties"
+      - "gradle/wrapper/*"
+      - "gradlew"
+      - "settings.gradle.kts"
+      - "src/main/resources/*"
+      - "src/main/resources-latest/*"
+    config:
+      commands:
+        - ["./gradlew", "-PsupervisorApiProjectPath=components-supervisor-api-java--lib/", "-PgitpodProtocolProjectPath=components-gitpod-protocol-java--lib/", "-PenvironmentName=latest", "runPluginVerifier"]
+        - ["./gradlew", "-PsupervisorApiProjectPath=components-supervisor-api-java--lib/", "-PgitpodProtocolProjectPath=components-gitpod-protocol-java--lib/", "-PenvironmentName=latest", "buildPlugin"]
--- a/components/ide/jetbrains/backend-plugin/build.gradle.kts
+++ b/components/ide/jetbrains/backend-plugin/build.gradle.kts
@@ -10,18 +10,36 @@ fun properties(key: String) = project.findProperty(key).toString()
 plugins {
     // Java support
     id("java")
-    // Kotlin support
-    id("org.jetbrains.kotlin.jvm") version "1.5.10"
+    // Kotlin support - check the latest version at https://plugins.gradle.org/plugin/org.jetbrains.kotlin.jvm
+    id("org.jetbrains.kotlin.jvm") version "1.7.0"
     // gradle-intellij-plugin - read more: https://github.com/JetBrains/gradle-intellij-plugin
-    id("org.jetbrains.intellij") version "1.0"
+    id("org.jetbrains.intellij") version "1.6.0"
     // detekt linter - read more: https://detekt.github.io/detekt/gradle.html
     id("io.gitlab.arturbosch.detekt") version "1.17.1"
     // ktlint linter - read more: https://github.com/JLLeitschuh/ktlint-gradle
     id("org.jlleitschuh.gradle.ktlint") version "10.0.0"
+    // Gradle Properties Plugin - read more: https://github.com/stevesaliman/gradle-properties-plugin
+    id("net.saliman.properties") version "1.5.2"
 }
 
 group = properties("pluginGroup")
 version = properties("version")
+
+val environmentName = properties("environmentName")
+
+project(":") {
+    kotlin {
+        var excludedPackage = "stable"
+        if (environmentName == excludedPackage) excludedPackage = "latest"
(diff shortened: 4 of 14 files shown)


akosyakov:
Should we review it? build seems to be broken

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
```diff
@@ -73,24 +90,139 @@ class GitpodTerminalService(private val project: Project) : Disposable {
 
         messageBusConnection.subscribe(ToolWindowManagerListener.TOPIC, toolWindowManagerListener)
 
-        return completableFuture
+        completableFuture.get()
+    }
+
+    private tailrec fun waitUntilGuestClientIsConnected() {
```
akosyakov, line 96: I wonder should not we rather merge this class into `GitpodClientProjectSessionTracker`? and rename to `GitpodClientProjectSessionManager`? 

It has a project and a session, so no need for guessing?

felladrin: It's possible, indeed. There's just one difference, as far as I remember:
As `GitpodClientProjectSessionTracker` is marked with `client="guest"`, the class is being invoked everytime a client connects or reconnects.
While `GitpodTerminalService` currently runs only when the backend starts.
So it's necessary to add a flag to determine if the shared terminals have already been created and don't create them again in case of reconnection.

```
        <projectService serviceImplementation="io.gitpod.jetbrains.remote.GitpodClientProjectSessionTracker" client="guest" preload="true"/>
        <projectService serviceImplementation="io.gitpod.jetbrains.remote.GitpodTerminalService" preload="true"/>
```

But wouldn't it be better to do it in a follow-up PR as it's a refactoring?

akosyakov: ok, let's do later if we have problems with multiple clients connecting

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodTerminalService.kt
```diff
@@ -73,24 +90,139 @@ class GitpodTerminalService(private val project: Project) : Disposable {
...
+        }
     }
 
-    private suspend fun delayUntilControllerClientConnects() {
-        while (UnattendedHostManager.getInstance().controllerClientId == null) {
-            delay(1000L)
+    private fun createSharedTerminalAndExecuteCommand(name: String, command: String) {
+        val registeredTerminals = terminalView.widgets.toMutableList()
+
+        backendTerminalManager.createNewSharedTerminal(UUID.randomUUID().toString(), name)
```
felladrin, line 116: JetBrains team suggested we check `JBTerminalWidget.TerminalTitle.change` for changing the terminal tab title. I'm gonna try it now.

felladrin: I've updated the code.
It worked but will only update the title after the terminal window gains focus.

akosyakov:
@felladrin why it does not work in stable anymore? Should not we use conditional compilation to have one version for EAP and another for stable? see how it is done in Rust: https://github.com/intellij-rust/intellij-rust/blob/eefa4f43e19e5dc6e3923e99478b504bdec24028/build.gradle.kts#L388-L398

akosyakov:
@felladrin I was trying with https://github.com/akosyakov/parcel-demo and for some reasons the second terminal is `Local`. I need to click to see real terminal label. Can we update them dynamically?

I'm also not sure about welcome terminal. I'm kind of liking it, but at the same time maybe a user wants to see the last terminal.

akosyakov:
Also explicit closing of the frontend terminal still does not trigger closing of the underlying supervisor task. We should make sure that it works. Do we track it already for JB remote dev? @felladrin

felladrin:
> @felladrin I was trying with https://github.com/akosyakov/parcel-demo and for some reasons the second terminal is Local. I need to click to see real terminal label. Can we update them dynamically?

I tried. Checked every method available inside inside `terminalWidget.terminalTitle.change` and it didn't work. Seems to be a limitation from the backend, which only fetches the info from terminal when the Thin Client focus on it.

I've also tried using `terminalWidget.grabFocus()` (also ith `ClientId.withClientId()` to force executing on the Thin Client), but it didn't work to programmatically focus on the terminal to fetch the info.
I'm going to file an issue for JetBrains today.

> I'm also not sure about welcome terminal. I'm kind of liking it, but at the same time maybe a user wants to see the last terminal.

I agree! Let me remove it.

felladrin:
> Also explicit closing of the frontend terminal still does not trigger closing of the underlying supervisor task. We should make sure that it works. Do we track it already for JB remote dev? @felladrin

Hmm, it was not implemented this time.

I'll put this PR back to Draft state while I implement it with the following specs:

- If we run 'exit' we need to react to the Supervisor info that the terminal was closed, then we close JetBrians Terminal.
- If we click `X` button to close the terminal, we intercept the JetBrains Terminal Closing action and emit to Supervisor a `close` command.

felladrin:
> Also after clicking it shows outdated title, i.e. `Gitpod Task2: idea-cli` while a terminal does not run anything anymore.

Good point. I need to set up a listener for terminal title changes.
Let me add it.

felladrin:
> @felladrin why it does not work in stable anymore? Should not we use conditional compilation to have one version for EAP and another for stable? see how it is done in Rust: https://github.com/intellij-rust/intellij-rust/blob/eefa4f43e19e5dc6e3923e99478b504bdec24028/build.gradle.kts#L388-L398

Thanks for sharing! It could work for conditionally using `terminalWidget.terminalTitle.change` (which was only added on 2022.2).

But [terminals losing reference](https://youtrack.jetbrains.com/issue/CWM-6093) is not related to the plugin SDK, but to the Backend IDE compilation. And the fix is only available in Backend IDE 2022.2+ builds.
So even if we compile the backend-plugin using the 2022.2 SDK, the plugin can still run in 2022.1 IDE if we don't use any API added on 2022.2, but issues from the Backend IDE remain there.

akosyakov:
> But [terminals losing reference](https://youtrack.jetbrains.com/issue/CWM-6093) is not related to the plugin SDK, but to the Backend IDE compilation. And the fix is only available in Backend IDE 2022.2+ builds.
So even if we compile the backend-plugin using the 2022.2 SDK, the plugin can still run in 2022.1 IDE if we don't use any API added on 2022.2, but issues from the Backend IDE remain there.

I thought that we should have 2 different source sets one for stable backend without using the fix, and another for version with a fix. What I'm missing?

felladrin:
> I thought that we should have 2 different source sets one for stable backend without using the fix, and another for version with a fix. What I'm missing?

💡Got it!
The one for the stable version would still display the Welcome Message in a single shared terminal, while the other, for EAP, would display one terminal for each task (without the welcome message), right?
Let me try.

akosyakov:
> 💡Got it!
The one for the stable version would still display the Welcome Message in a single shared terminal, while the other, for EAP, would display one terminal for each task (without the welcome message), right?
Let me try.

Yes, we should avoid long standing PRs. We should have one plugin which can work in both versions and after new release. We just need to clean up old version code.

akosyakov:
We agreed that we should separate that this logic is only applied if latest is toggled for now, i.e. experimentation phase in [[1](https://www.notion.so/gitpod/Progressive-roll-out-of-IDE-features-e5efaf3b5e5645648c6265242c4029bc)]. After that it can be merged and worked on with follow-up PRs.

akosyakov (approved):
It is great work! Thank you for such nice separation.

> Resolves https://github.com/gitpod-io/gitpod/issues/10579

I don't think it does. We should do all follow-up PRs, right? And address outstanding usability issues?

PR #10377, opened 2022-05-31
Properly handle the workspace list when the workspace context ref is undefined

felladrin:
## Description

Properly handle the workspace list when the workspace context ref is undefined.

This fixes the issue on JetBrains Gateway preventing the workspace list from being displayed when a workspace had been created from a detached commit instead of a branch. For example, creating a workspace from https://github.com/jenkinsci/design-library-plugin/tree/73f12733d95932c333ba4155521ff78b4ed40dfc

This is the info that comes with this workspace, which does not contain the `workspace.context.ref`:
```json
{
    "jsonrpc": "2.0",
    "result": [
        {
            "workspace": {
                "context": {
                    "isFile": false,
                    "path": "",
                    "title": "jenkinsci/design-library-plugin - 73f12733:",
                    "revision": "73f12733d95932c333ba4155521ff78b4ed40dfc",
                    "repository": {
                        "cloneUrl": "https://github.com/jenkinsci/design-library-plugin.git",
                        "host": "github.com",
                        "name": "design-library-plugin",
                        "owner": "jenkinsci",
                        "private": false
                    },
                    "normalizedContextURL": "https://github.com/jenkinsci/design-library-plugin/tree/73f12733d95932c333ba4155521ff78b4ed40dfc",
                    "checkoutLocation": "design-library-plugin",
                    "referrer": "jetbrains-gateway",
                    "referrerIde": "intellij"
                }
            }
        }
    ]
}
```

## Related Issue(s)

None.

## How to test

- Create a workspace based on some detached commit from any repo. You can use this, for example: https://github.com/jenkinsci/design-library-plugin/tree/73f12733d95932c333ba4155521ff78b4ed40dfc
- Note that on the Gitpod Dashboard, it will display the workspace reference as "(detached"):
  image
- Now open JetBrains gateway using the [current stable version of the plugin](https://plugins.jetbrains.com/plugin/18438-gitpod-gateway/versions/stable/177166). You should get the workspace list stuck in "Loading" state:
  image
- Download [Gitpod Gateway 0.0.1-vn-fix-undefined-context-ref-on-jb-gateway.3](https://plugins.jetbrains.com/plugin/18438-gitpod-gateway/versions/Dev/180010) and manually install it on JetBrains Gateway
- Confirm if the Workspace list is displayed correctly and displaying the detached workspace correctly.
  image

## Extra Info

[...]

PR #10596, opened 2022-06-10
Refactor 'normalizedContextURL' property from Java class 'WorkspaceContext' as Optional

felladrin:
## Description

Refactor 'normalizedContextURL' property from Java class 'WorkspaceContext' as Optional, because it can come null from the server, as you can see on the _interface WorkspaceContext_ from `components/gitpod-protocol/src/protocol.ts`.

https://github.com/gitpod-io/gitpod/blob/40158aceb4c956f097b22253c040afa71d565c66/components/gitpod-protocol/src/protocol.ts#L885

## Related Issue(s)

Issue raised by @mustard-mh in https://github.com/gitpod-io/gitpod/pull/10377#pullrequestreview-990366026

## How to test

I haven't seen a case where the 'normalizedContextURL' of a workspace was null. And don't know if we can simulate it, but following the same `How to test` from https://github.com/gitpod-io/gitpod/pull/10377 should show it's working as expected. The difference is that we should use the [gateway-plugin build generated by this PR](https://plugins.jetbrains.com/plugin/18438-gitpod-gateway/versions/Dev/183936) instead.

## Release Notes

```release-note
NONE
```

## Documentation

- [x] /werft --no-preview=true

### DIFF
--- a/components/gitpod-protocol/java/src/main/java/io/gitpod/gitpodprotocol/api/entities/WorkspaceContext.java
+++ b/components/gitpod-protocol/java/src/main/java/io/gitpod/gitpodprotocol/api/entities/WorkspaceContext.java
@@ -10,8 +10,8 @@ public class WorkspaceContext {
     private String normalizedContextURL;
     private String ref;
 
-    public String getNormalizedContextURL() {
-        return normalizedContextURL;
+    public Optional<String> getNormalizedContextURL() {
+        return Optional.ofNullable(normalizedContextURL);
     }
 
     public void setNormalizedContextURL(String normalizedContextURL) {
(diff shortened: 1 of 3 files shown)


mustard-mh (approved):
Test good in Gateway 2022.1.3

|Img|
|-|
|image|

I build a prev env which makes server.getWorkspaces API respond without field `normalizedContextURL`, and it works well too

https://github.com/gitpod-io/gitpod/blob/5f12b12abb8cfdf66d2dab2c925bb3652836c303/components/server/src/workspace/gitpod-server-impl.ts#L834-L837

|Workspaces|Connection|
|-|-|
|image|image|

PR #10599, opened 2022-06-10
Git-ignore all files from the '.idea' folder, except gradle.xml

felladrin:
## Description

When opening Gitpod repository on different JetBrains IDEs cause the changes on files inside .idea folder. Currently we simply ignore those files by hand at the moment we're going to commit code changes.

But we actually don't need most of the files inside .idea folder. So this PR git-ignores all files from the '.idea' folder, except `gradle.xml` as it's required to initialize the Gradle projects from "backend-plugin" and "gateway-plugin" folders, allowing autocomplete to the Kotlin files and other integrations with the IDE.

## Related Issue(s)

NONE.

## How to test

- Open the branch from this PR in Gitpod IntelliJ IDEA and then in a locally running IntelliJ IDEA Community Edition: https://gitpod.io/#https://github.com/gitpod-io/gitpod/tree/felladrin/git-ignore-idea-everchanging-files
- See that no files are marked as changed (`git status`) after the project finishes indexing.
- Check also if files from `backend-plugin` and `gateway-plugin` (both from inside `components/ide/jetbrains`) have Intellisense working properly.

## Release Notes

```release-note
NONE
```

## Documentation

- [x] /werft --no-preview=true


akosyakov:
We committed .idea folder to import gradle modules on opening of IntelliJ. If we remove modules then we can rather delete entire .idea folder completely?

felladrin:
> We committed .idea folder to import gradle modules on opening of IntelliJ. If we remove modules then we can rather delete entire .idea folder completely?

I tried removing and git-ignoring the whole folder, but then when opening it on IntellIJ Idea it opens with the Project Explorer messed up:
image

akosyakov:
Should we close it then?

akosyakov (approved):
let's try if it does not cause any perf drawbacks

/hold

felladrin:
> I was thinking can we put parent gradle project in root instead? and completely remove .idea? See https://docs.gradle.org/current/userguide/multi_project_builds.html#sec:creating_multi_project_builds
> 
> Maybe IntelliJ will pick it up automatically then?

Let me try it!

felladrin:
@akosyakov, I tried adding settings.gradle.kts with the following content to the root, but it didn't link the projects in IntelliJ:

```kotlin
rootProject.name = "gitpod"

include("backend-plugin")
project(":backend-plugin").projectDir = file("components/ide/jetbrains/backend-plugin")

include("gateway-plugin")
project(":gateway-plugin").projectDir = file("components/ide/jetbrains/gateway-plugin")
```
image

I've asked JetBrains Team if there's a specific way to do it, but meanwhile, I'd like to merge it with the current approach of leaving just ".idea/gradle.xml" file committed.

akosyakov:
@felladrin yes, go ahead with current approach

PR #10563, opened 2022-06-09
Use new stripe secret in preview environments

andrew-farries:
## Description

Following on from #10552 which added the `stripe-api-keys` secret to preview environment clusters, this PR makes the `server` component use the secret and removes the old `stripe-config` secret from the repo.

## Related Issue(s)

Part of #9036 and https://github.com/gitpod-io/ops/issues/2554

## How to test

Manually trigger a werft job for this branch [(Notion doc)](https://www.notion.so/gitpod/Manually-triggering-a-Werft-job-4aaf0cacbfc54f21991be700746637c3#840790f17f7143009bf0164bcd0720c1), setting `with-payment=true`.

Check that `server` is in a running state in the preview environment cluster and that it has successfully mounted the `stripe-api-keys` secret:

```
> kubectl describe pod <server pod>
```

## Release Notes

```release-note
NONE
```

Issue #9036, opened 2022-03-31
Epic: Usage-Based Pricing


svenefftinge:
### Summary

Gitpod's SaaS pricing shall be based on usage rather than number of seats (current state).

### Context

We believe the value received by teams and individuals is correlated with the time spent using Gitpod which makes usage hours (not seat count) our value metric. While seat count serves as a proxy, we believe usage is a more accurate predictor of customer value. Hence, we believe our current pricing creates friction in turning users into paying customers and limiting the potential upside of large clients for Gitpod (fixed €35 for unleashed plan).

### Value

Align value users gain with what they pay.
Make it easier for new team members to join a team and grow into the product as they use it more often. 

Moving away from the existing user-based subscriptions to usage-based pricing has become more important because we need this to ship [Different Workspace Classes](https://github.com/gitpod-io/gitpod/issues/8261) as well as allowing for [flexible workspace timeouts](https://github.com/gitpod-io/gitpod/issues/9038).

### Acceptance Criteria

Users and teams can purchase credits and use them run workspaces. 

### Measurement

TBD

### Growth Area

Expansion

### Hypothesis

- Gitpod is most valuable in team settings (i.e. >1 contributor to the codebase).
- Ephemeral dev environments are mostly useful for projects with 2+ contributors. Single-person projects can easily live in a long-lived workspace.
- Seat-based pricing leads to situations in which the customer pays even though they are not using the product.
(Large) companies are familiar and comfortable with usage-based pricing.
- competition used usage-based pricing.

### Out of scope

- Automatic migration of existing subscriptions

### Skateboard (Alpha version)

**Generate invoices for Team Gitpod only (week of Jun 27)**
- [x] https://github.com/gitpod-io/gitpod/pull/10227
- [x] #10316
- [x] #10317
- [x] https://github.com/gitpod-io/gitpod/pull/10249
- [x] #10319
- [x] https://github.com/gitpod-io/gitpod/issues/10430
- [x] https://github.com/gitpod-io/ops/issues/2554
- [x] #10322
- [x] https://github.com/gitpod-io/gitpod/issues/10642
- [x] #10720
- [x] #10753
- [ ] Add integration test for Billing Controller which asserts against mock Stripe
- [x] #10851
- [x] https://github.com/gitpod-io/gitpod/pull/10302
- [x] #10326
- [x] #10327
- [x] #10655
- [x] #10533
- [x] #10856
- [x] #10822
- [x] #10901
- [x] #10973
[...]

PR #10631, opened 2022-06-13
Mount Stripe API keys into the usage component

andrew-farries:
## Description

Mount the Stripe API keys into the `usage` component so that they are available to report usage data to the Stripe API.

These are the same API keys that are mounted into the `server` component (#10563)

## Related Issue(s)
Part of #9036 

## How to test

Manually trigger a werft job for this branch [(Notion doc)](https://www.notion.so/gitpod/Manually-triggering-a-Werft-job-4aaf0cacbfc54f21991be700746637c3#840790f17f7143009bf0164bcd0720c1), setting with-payment=true.

In the preview environment for the branch, verify that the usage pod has a `/stripe-secrets` mount containing Stripe API keys.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/install/installer/pkg/components/usage/constants.go
+++ b/install/installer/pkg/components/usage/constants.go
@@ -5,5 +5,6 @@
 package usage
 
 const (
-	Component = "usage"
+	Component             = "usage"
+	stripeSecretMountPath = "stripe-secret"
 )
--- a/install/installer/pkg/components/usage/deployment.go
+++ b/install/installer/pkg/components/usage/deployment.go
@@ -7,6 +7,7 @@ import (
 	"github.com/gitpod-io/gitpod/common-go/baseserver"
 	"github.com/gitpod-io/gitpod/installer/pkg/cluster"
 	"github.com/gitpod-io/gitpod/installer/pkg/common"
+	"github.com/gitpod-io/gitpod/installer/pkg/config/v1/experimental"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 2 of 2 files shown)


install/installer/pkg/components/server/types.go
```diff
@@ -32,7 +32,7 @@ type ConfigSerialized struct {
 	ImageBuilderAddr                  string   `json:"imageBuilderAddr"`
 	VSXRegistryUrl                    string   `json:"vsxRegistryUrl"`
 	ChargebeeProviderOptionsFile      string   `json:"chargebeeProviderOptionsFile"`
-	StripeSettingsFile                string   `json:"stripeSettingsFile"`
+	StripeSecretsFile                 string   `json:"stripeSecretsFile"`
```
geropl, line 35: nit: Why is this called "StripeSecretsFile"? Given the rename, shouldn't it be the "StripeApiKeyFile" or so? :thinking:

.werft/jobs/build/helm/values.payment.yaml
```diff
@@ -15,7 +15,7 @@ components:
         secretName: chargebee-config
     - name: stripe-config
       secret:
-        secretName: stripe-config
+        secretName: stripe-api-keys
```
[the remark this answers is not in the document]
geropl: If not: Why to rename the to "stripe-api-keys" instead of "stripe-secret?"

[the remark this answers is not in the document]
andrew-farries: The idea is to put the rest of the config in a configmap as the data is not sensitive.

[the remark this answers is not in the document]
geropl: Jep, having it separate makes sense. :+1:

install/installer/pkg/components/server/types.go
andrew-farries: yes, perhaps.  I went with the more general name because I wasn't sure if we'd need other secrets in there besides the keys.

geropl: This is about the file, so "secret mount point" + "key inside the secret".
If we added another key in that secret it would still be another file.

geropl: Anyway, I just noticed it already got merged. But still think `StripeKeysFile` would make sense. :upside_down_face:

geropl (approved):
Tested, and volume is mounted :heavy_check_mark:

Issue #9374, opened 2022-04-18
Add Team and Project to OWI metadata in ws-manager, so we can inspect metrics by customer


atduarte:
Example use cases: 
1. Understand better the performance requirements of certain teams and project, and support research of #8261.
2. Provide context whenever we need to assist a customer.

cc @kylos101

atduarte:
@kylos101 added "Project". WDYT?

kylos101:
> @kylos101 added "Project". WDYT?

Good clarification, thanks @atduarte !

kylos101:
@atduarte good find ... scheduled!

jankeromnes:
@atduarte In the context of Usage-Based Billing, we would like to allow "re-attributing" workspace usage.

**Use case**: One of your teams upgrades to Usage-Based Billing, and wants to cover 100% of your Gitpod usage (whether that usage is on team projects, or personal projects, or no projects at all just random workspaces).

If we do "re-attribute" usage to a specific team/customer, I guess you'd like the OWI metadata in ws-manager to also show the re-attributed `teamId`, right? (I.e. if team Gitpod covers all my usage, and I open a workspace for a project that's under a different team -- the logs should show Gitpod's teamId, not the other project's teamId, right?)

PR #10632, opened 2022-06-13
[ws-manager] Add team and project to logs and traces

Furisto:
## Description
Add team and project to ws-manager logs and traces. 

Logs:
logs

Traces:
traces

## Related Issue(s)
Fixes https://github.com/gitpod-io/gitpod/issues/9374

## How to test
- Create a team and within that team a project
- Open a workspace within the project
- Check that the pod contains project and team labels
- Search for team id/projectId in logs explorer
- Search for  team id/projectId in honeycomb

## Release Notes

```release-note
NONE
```


components/ws-manager/pkg/manager/create.go
```diff
@@ -911,6 +911,14 @@ func (m *Manager) newStartWorkspaceContext(ctx context.Context, req *api.StartWo
 		workspaceClassLabel:    clsName,
 	}
 
+	if req.Metadata.Project != "" {
```
sagor999, line 914: nit: why not to just add it above into labels map as: 
`wsk8s.ProjectLabel: req.Metadata.Project` ?
If it is empty, it will be an empty label. Which I think is fine.

sagor999 (approved):

components/common-go/log/log.go
```diff
@@ -36,6 +40,35 @@ func OWI(owner, workspace, instance string) log.Fields {
 	}
 }
 
+// LogContext builds a structure meant for logrus which contains the owner, workspace and instance.
+// Beware that this refers to the terminology outside of wsman which maps like:
+//    owner = owner, workspace = metaID, instance = workspaceID
+func LogContext(owner, workspace, instance, project, team string) log.Fields {
```
jenting, line 46: I think we could extend this function https://github.com/gitpod-io/gitpod/blob/e1b335c029047debce6209485635c76b678a5e36/components/common-go/log/log.go#L31 to have the project and team as well.

Or to create another function
```go
func PT(project, team string) log.Fields {
	return log.Fields{
		Project: project,
		Team: team,
	}
}
```

components/common-go/tracing/tracing.go
```diff
@@ -100,7 +100,7 @@ func FromContext(ctx context.Context, name string) (opentracing.Span, context.Co
 
 // ApplyOWI sets the owner, workspace and instance tags on a span
 func ApplyOWI(span opentracing.Span, owi logrus.Fields) {
-	for _, k := range []string{log.OwnerField, log.WorkspaceField, log.InstanceField} {
+	for _, k := range []string{log.OwnerField, log.WorkspaceField, log.InstanceField, log.ProjectField, log.TeamField} {
```
geropl, line 103: Not necessarily in-scope for this PR, but: We started writing up some naming conventions for tracing, incl. [for OWI](https://www.notion.so/gitpod/Naming-conventions-af9084c4743349089cf2450a3c860733#5c8ee0a476904a9aa2abc637340aa747). (just extended it for projectId/teamId :innocent: )

components/ws-manager-api/core.proto
```diff
@@ -446,6 +446,12 @@ message WorkspaceMetadata {
     // Annotations are key/value pairs that gets attached to the workspace.
     // This is primarily intended for annotating headless workspace loads.
     map<string, string> annotations = 4;
+
+    // team the workspace belongs to
+    string team = 5;
+
+    // project the workspace belongs to
+    string project = 6;
```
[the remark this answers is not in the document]
Furisto: That's a good question! Here is how I understand it: In contrast to proto2, proto3 uses no presence semantics as default (i.e. it just does not serialize values that are not present) and does not use explicit presence semantics (i.e. it stores if a field has explicitly been set to a value). 

Tagging the fields with optional would switch it to explicit presence semantics.  Explicit presence therefore increases message size, even when no value is present, because of the tracking requirement. In return additional methods are generated that allow you to check if a value has been set. I do not foresee a need to do this, as with no presence semantics we get an empty string for team/project which should be good enough for us. What do you think?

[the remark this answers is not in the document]
geropl: > I do not foresee a need to do this, as with no presence semantics we get an empty string for team/project which should be good enough for us. What do you think?

Sometimes `optional` is nice for readability bc a) you get the `has_` accessor and b) for making it explicit in the proto file. (Or if you need the extra bit in case you need the default value in the value domain - not the case here),
But searching our repo we mention optional [mostly](https://github.com/gitpod-io/gitpod/blob/e3423ccb96223de8f0ac83ca9f51317a8f3e20b9/components/public-api/gitpod/v1/workspaces.proto#L268) on comments, so fine with not using it! :+1:  A comment would be nice, though. :slightly_smiling_face:

components/common-go/log/log.go
jenting: Well, my idea was has a another function like this
```go
func PT(project, team string) log.Fields {
	return log.Fields{
		Project: project,
		Team: team,
	}
}
```

So, the caller would be
```go
owi := log.OWI(req.Metadata.Owner, req.Metadata.MetaId, req.Id)
pt := log.PT(project, team)
clog := log.WithFields(owi).WithFields(pt)
```

Furisto:
@aledbf You are configured as one of the reviewers. Can you please take a look?

aledbf (approved):

Issue #9930, opened 2022-05-11
gp ports command


loujaybee:
Currently, in IDE's such as JetBrains, and when SSH'ing into Gitpod it's not (yet) possible to manage your Gitpod ports (e.g. list open ports). This issue is to introduce a `gp` CLI command which allows a user to modify and manipulate the ports of their workspace. Ideally, in the long run, the functionality available in each of the IDE's and `gp` CLI is mirrored / consistent.

loujaybee:
Relates to: https://github.com/gitpod-io/gitpod/issues/9286

akosyakov:
The first version should be 🛹 and allow following:
#### List ports
```
gp ports list
```

This command should print a table of ports with following columns:
- status: not served|failed to expose|exposing...|served
- label: port name + number
- visibility: public|private
- url
- description


#### Change visibility of ports
```
gp ports visibility <port>:{public|private}...
```

Example:
```
gp ports visibility 8000:public 3000:private
```

loujaybee:
Issue for docs: https://github.com/gitpod-io/website/issues/2178

akosyakov:
@felladrin is it resolved? 

I noticed that `watch gp ports` does not look nice. I think @andreafalzetti has some ideas how to improve.

andreafalzetti:
I've created a dedicated issue for `watch gp ports` because it's a side effect of colors and they are also used in other commands, so I think we can fix all-in-one PR. 

Link to dedicated issue: https://github.com/gitpod-io/gitpod/issues/10633

andreafalzetti:
I've raised https://github.com/gitpod-io/gitpod/pull/10638 to fix #10633 - which solves the output issue with `watch` for both commands `gp ports lists` and `gp tasks list`.

akosyakov:
@loujaybee @felladrin I removed it from scheduled it for now. Since we have gp ports now and it is better to improve other aspects of JB integration.

PR #10638, opened 2022-06-13
[GP CLI] Allow explicity and automated disabling of output colorization

### LINKED ISSUE #10633: [GP CLI] Automatically disable output colorization when not required
### Bug description

When using `gp` CLI in addition to other programs via stdin/stout, (e.g. `watch`), the output colorization ruins the user experience because the colors are not supported and are actually rendered as ASCII characters.

### Steps to reproduce

1. Launch a workspace
2. Run `watch gp ports list` or `watch gp tasks list`

It will look like:

Screenshot 2022-06-13 at 12 16 55
Screenshot 2022-06-13 at 12 16 48

### Workspace affected

_No response_

### Expected behavior

The output colorization is automatically disabled. We can now implement this thanks to https://github.com/gitpod-io/gitpod/pull/10570/files#diff-7d3a206fbf5ce6606b82d0f45db9ed5d3100ba0116e94a052047835e1ef8ea53

### Example repository

_No response_

### Anything else?

_No response_

andreafalzetti:
## Description
This PR updates the commands of gitpod-cli which use output colorization, making it explicitly disabled when requested by the user (via `--no-color`) or automatically (when stdin/stdout detected or for other community/standard conditions) [[1]](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-cli/pkg/utils/colors.go).

## Related Issue(s)

Fixes #10633
Related #9930 

## How to test
1. Open the prev env
2. Run `watch gp ports list` and `watch gp tasks list` notice how the output does not display any invalid character and the output is readable (and not colorized).

## Release Notes

```release-note
NONE
```

## Documentation

We should document this flag/feature: https://github.com/gitpod-io/website/issues/2235

### DIFF
--- a/components/gitpod-cli/cmd/ports-list.go
+++ b/components/gitpod-cli/cmd/ports-list.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	supervisor_helper "github.com/gitpod-io/gitpod/gitpod-cli/pkg/supervisor-helper"
+	"github.com/gitpod-io/gitpod/gitpod-cli/pkg/utils"
 	supervisor "github.com/gitpod-io/gitpod/supervisor/api"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -76,9 +77,14 @@ var listPortsCmd = &cobra.Command{
 				}
 			}
 
+			colors := []tablewriter.Colors{}
+			if !noColor && utils.ColorsEnabled() {
+				colors = []tablewriter.Colors{{}, {statusColor}, {}, {}}
+			}
+
 			table.Rich(
 				[]string{fmt.Sprint(port.LocalPort), status, port.Exposed.Url, nameAndDescription},
-				[]tablewriter.Colors{{}, {statusColor}, {}, {}},
+				colors,
 			)
 		}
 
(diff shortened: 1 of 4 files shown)


filiptronicek (approved):
Thanks a lot for taking this, @andreafalzetti!

I tested it and it's working just as advertised, but I have one question: could we abstract the logic for **all** of the `gp` commands so that for future commands we could just wrap the CLI in some function that would add the no-color options? This would help with many commands failing if you provide them with `--no-color`, like `gp ports`, which I think it would be better if it just ignored (if there was no color to display in the first place).

Holding the PR if you have any thoughts on this, feel free to unhold :)

/hold

andreafalzetti:
> Thanks a lot for taking this, @andreafalzetti!
> 
> I tested it and it's working just as advertised, but I have one question: could we abstract the logic for **all** of the `gp` commands so that for future commands we could just wrap the CLI in some function that would add the no-color options? This would help with many commands failing if you provide them with `--no-color`, like `gp ports`, which I think it would be better if it just ignored (if there was no color to display in the first place).
> 
> Holding the PR if you have any thoughts on this, feel free to unhold :)
> 
> /hold



> 

Thanks, @filiptronicek for the review 🙏 

We now have the `util.ColorsEnabled` which wraps all the logic to disable colors, the `noColor` flag variable has global scope, so it's accessible from all commands - if they need to have it. I thought about having a global `--no-color` flag for all commands, but for those which don't use colors, it would be confusing why it's there, so I've opted for having the 1-line copy/paste flag on each command that uses colors.

At the moment that's the best I can think of! 

If anyone has any idea for further improvements, I am all ears :) 

/unhold

PR #10648, opened 2022-06-14
Resolve prebuild updateables

AlexTugarev:
This PR ensures we resolve prebuild updatables even if there is an issue to create a commit status for.

Also to ensure we're not overloading this introduces proper limits to the periodic updater.


roboquat:

<details>

</details>

components/server/ee/src/prebuilds/prebuilt-status-maintainer.ts
```diff
@@ -243,14 +241,19 @@ export class PrebuildStatusMaintainer implements Disposable {
...
+                if (age > MAX_UPDATABLE_AGE && age < IGNORED_UPDATEABLE_AGE) {
                     log.info(
                         "found unresolved updatable that's older than MAX_UPDATABLE_AGE and is inconclusive. Resolving.",
                         updatable,
                     );
-                    await this.doUpdate(ctx, updatable, updatable.prebuild);
+                    try {
+                        await this.doUpdate(ctx, updatable, updatable.prebuild);
+                    } catch (error) {
+                        log.error("Failed to process prebuild updatable.", error, { updatable });
```
geropl, line 255: nit: is there opportunity to extract `logContext` info (owner, worspaceId, wsiId)from the updateable and pass it as first arg here?

components/server/ee/src/prebuilds/prebuilt-status-maintainer.ts
```diff
@@ -197,15 +199,11 @@ export class PrebuildStatusMaintainer implements Disposable {
...
-                        );
-                        found = true;
-                    } else {
-                        throw err;
-                    }
+                    log.info("Could not create commit status.", { updatable, errorMessage: err?.message });
+                    // e.g. this happens if a commit is removed from remote repository (force push)
+
+                    // let's just assume, we have not found the commit to create a status for
+                    found = false;
```
geropl, line 206: Do we need to keep it if we're not sure what it means? E.g., either we keep the extra-case for "Not found" and report "found", or we drop it altogether.
The span would benefit from a `TraceContext.setError({ span }, err);`, which is now not done anymore below.

AlexTugarev:
@geropl, I've added a test for `getUnresolvedUpdatables` as suggested.

components/server/ee/src/prebuilds/prebuilt-status-maintainer.ts
```diff
@@ -197,25 +198,10 @@ export class PrebuildStatusMaintainer implements Disposable {
...
-                            "Did not find repository while updating updatable. Probably we lost the GitHub permission for the repo.",
-                            { owner: updatable.owner, repo: updatable.repo },
-                        );
-                        found = true;
-                    } else {
-                        throw err;
-                    }
+                    log.info("Could not create commit status.", { updatable, errorMessage: err?.message });
+                    // (AT) this might happen e.g. when a commit is removed from remote repository (force push).
+                    // Note, we're ignoring those errors for tracing.
```
geropl, line 203: :+1:

components/server/ee/src/prebuilds/prebuilt-status-maintainer.ts
```diff
@@ -243,14 +229,19 @@ export class PrebuildStatusMaintainer implements Disposable {
         const ctx = TraceContext.childContext("periodicUpdatableCheck", {});
 
         try {
-            const unresolvedUpdatables = await this.workspaceDB.trace(ctx).getUnresolvedUpdatables();
+            const unresolvedUpdatables = await this.workspaceDB.trace(ctx).getUnresolvedUpdatables(MAX_UPDATABLES);
             for (const updatable of unresolvedUpdatables) {
-                if (Date.now() - Date.parse(updatable.workspace.creationTime) > MAX_UPDATABLE_AGE) {
+                const age = Date.now() - Date.parse(updatable.workspace.creationTime);
+                if (age > MAX_UPDATABLE_AGE && age < IGNORED_UPDATEABLE_AGE) {
```
geropl, line 235: :pray: 

Did you think about going without `IGNORED_UPDATEABLE_AGE` because we now should be marking all entries as "resolved"? I feel that would be a bit "cleaner".
Both would work, obviously.

AlexTugarev: That's a second barrier, to ensure we're not overloading. This is also ensures an immediate fix of the issue, no db migration required.

geropl: > This is also ensures an immediate fix of the issue, no db migration required.

Wouldn't that be the case as well if we didn't have it? It would doUpdate all entries, and mark them as resolved?

AlexTugarev: There is difference though. 

Without this kind of hard limit here, `getUnresolvedUpdatables` will continue to return VERY OLD stuff for which this updates will be tried to get applied, i.e. tons of unnecessary/failing commit status updates via GH App. THIS isn't what we want, do we?

geropl: I feel having the DB in a good shape is beneficial. If you think it takes too long, and we should include a migration, I'd rather do that then skipping old entries.
Having systems that automatically converge is always easier to reason about then time-window stuff.

geropl (approved):
Tests work, and changes LGTM! :+1:

PR #10659, opened 2022-06-14
Code clean-up

AlexTugarev:
Removing obsolete code
 * ToS path
 * TheiaPlugin
 * UserMessageViews

```release-notes
NONE
```


components/server/src/auth/generic-auth-provider.ts
```diff
@@ -443,22 +446,9 @@ export class GenericAuthProvider implements AuthProvider {
...
-                });
-
-                // attach the sign up info to the session, in order to proceed after acceptance of terms
-                await TosFlow.attach(request.session!, flowContext);
-
-                response.redirect(this.config.hostUrl.withApi({ pathname: "/tos", search: "mode=login" }).toString());
-                return;
+            if (TosFlow.WithIdentity.is(flowContext)) {
+                // This is the regular path on sign up.
+                await this.signUpNewUser(request, response, authFlow, flowContext);
```
geropl, line 451: This re-factoring is a bit out-of-scope I feel. The DB-cleanup is staight forward to review, but still has to be done carefully.

This is already another level, where I need to ask and understand what does (not) change to be able to juge if this is good or not.

Could you split this into two PRs? :pray:

geropl:
Thank you for caring and doing the cleanup here, @AlexTugarev ! :orange_heart: :pray:

components/server/src/auth/generic-auth-provider.ts
AlexTugarev: Will do, but for another reason. We need to fade out the client redirect to /tos first. 
Other than that, it's really just deleting code that's no longer evaluated until only what's left. And renaming to `signUpNewUser` in the end.

PR #10661, opened 2022-06-14
[prebuilds] make sure prebuild and branch commits are the same

### LINKED ISSUE #5908: Prebuild status indicator on "Branches" can be misleading
Situation: Th status indicator for the branch says "Ready" (because a prebuild for a former commit exists), but the current HEAD has none (yet). Because the branch details contain a commit (short) hash I as a user assume the commit and prebuild are connected (and wonder why I can a workspace build on the wrong commit), while they aren't.
image

Possible solutions:
 - shows status indicator for latest commit only (incl. none if for whatever reasons there is no prebuild for the latest commit, yet)
 - remove commit details, and make the view about branches only
 - somehow make transparent that here possibly is are a lot of prebuilds, one for each commit (no idea how to display that)

laushinka:
## Description
The issue in #5908 wasn't able to be reproduced, but this PR is to make sure that it doesn't happen.
When a prebuild's commit hash and branch's latest commit hash don't align, then we don't show the prebuild status in `/branches`.
Some prebuild namings are also improved here.

## Related Issue(s)
Fixes #5908

## How to test
1. Trigger a prebuild by pushing a commit.
2. Check that the commit hash info in the Commit column is consistent with your actual commit.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/dashboard/src/projects/Project.tsx
+++ b/components/dashboard/src/projects/Project.tsx
@@ -34,7 +34,7 @@ export default function () {
     const [isLoading, setIsLoading] = useState<boolean>(false);
     const [isLoadingBranches, setIsLoadingBranches] = useState<boolean>(false);
     const [branches, setBranches] = useState<Project.BranchDetails[]>([]);
-    const [lastPrebuilds, setLastPrebuilds] = useState<Map<string, PrebuildWithStatus | undefined>>(new Map());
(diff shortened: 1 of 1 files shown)


AlexTugarev (approved):
LGTM! Thanks @laushinka!

This change makes sense. There might be a gap between fetching latest commit via API and getting a push event via webhook to get processed. This will ensure, we're not rendering unrelated items.

PR #10669, opened 2022-06-14
[ws-daemon] Fix cgroups v2 I/O limiting and add support for Systemd cgroup

aledbf:
## Release Notes
```release-note
Fix I/O limiting when cgroups v2 is enabled
```


components/ws-daemon/pkg/daemon/daemon.go
```diff
@@ -58,12 +58,17 @@ func NewDaemon(config Config, reg prometheus.Registerer) (*Daemon, error) {
...
+
 	cgroupPlugins, err := cgroup.NewPluginHost(config.CPULimit.CGroupBasePath,
 		&cgroup.CacheReclaim{},
-		&cgroup.FuseDeviceEnablerV1{},
-		&cgroup.FuseDeviceEnablerV2{},
 		cgroupV1IOLimiter,
-		cgroup.NewIOLimiterV2(config.IOLimit.WriteBWPerSecond.Value(), config.IOLimit.ReadBWPerSecond.Value(), config.IOLimit.WriteIOPS, config.IOLimit.ReadIOPS),
+		cgroupV2IOLimiter,
+		&cgroup.FuseDeviceEnablerV1{},
+		//&cgroup.FuseDeviceEnablerV2{},
```
utam0k, line 71: How about using this function to determine which cgroup should use?
https://github.com/gitpod-io/gitpod/blob/e5c37ddbe1e8e994008f0ff12893d5e8c740848c/test/pkg/integration/common/cgroup.go#L13

aledbf: Fixed

csweichel (commented):
Code looks great - how I can test?

components/common-go/cgroups/cgroup.go
```diff
@@ -5,108 +5,21 @@
...
-}
-
-func IsLegacyCgroupSetup() (bool, error) {
-	setup, err := GetCgroupSetup()
-	if err != nil {
-		return false, err
-	}
-
-	return setup == Legacy, nil
+	return cgroups.Mode() == cgroups.Unified, nil
```
utam0k, line 13: We have discussed using containerd or runc cgroup libs, and we decided not to depend on any libs. WDYT?
https://github.com/gitpod-io/gitpod/pull/8471#discussion_r819542056

components/ws-daemon/pkg/dispatch/dispatch.go
```diff
@@ -223,13 +223,12 @@ func (d *Dispatch) handlePodUpdate(oldPod, newPod *corev1.Pod) {
...
-				l := l
-				go func() {
-					err := l.WorkspaceAdded(containerCtx, s.Workspace)
+				go func(listener Listener) {
+					err := listener.WorkspaceAdded(containerCtx, s.Workspace)
 					if err != nil {
 						log.WithError(err).WithFields(owi).Error("dispatch listener failed")
 					}
-				}()
+				}(l)
```
utam0k, line 231: 👍

components/common-go/cgroups/cgroup.go
aledbf: We already have that library as a dependency. Not sure that argument is valid?

aledbf:
@utam0k fuse is working correctly now. Please review again.

utam0k:
I checked
- [x] memory high and limit
```
root@ga8a2ae65be0d10d33ff12f:/sys/fs/cgroup/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod34261dc3_3529_427f_ad87_538c766e5335.slice/cri-containerd-d6ddb03d6d6091f03b5b2a6bc3b0f21be28da9a7b1f6b23fea04eb249100bca0.scope# cat memory.high 
3435970560
root@ga8a2ae65be0d10d33ff12f:/sys/fs/cgroup/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod34261dc3_3529_427f_ad87_538c766e5335.slice/cri-containerd-d6ddb03d6d6091f03b5b2a6bc3b0f21be28da9a7b1f6b23fea04eb249100bca0.scope# cat memory.max 
4294967296
```
- [x] CPU max with stress-ng
- [x] fuse
```
  58: (55) if r2 != 0x2 goto pc+4
  59: (55) if r4 != 0xa goto pc+3
  60: (55) if r5 != 0xe5 goto pc+2
  61: (b4) w0 = 1
  62: (95) exit
// Allow /dev/fuse(10:229) 
```
- [x] io limit `"ioLimit": { "writeBandwidthPerSecond": "300Mi", "readBandwidthPerSecond": "350Mi" },`
```
$ dd if=/dev/zero of=zero.dat bs=2G count=5
dd: warning: partial read (2147479552 bytes); suggest iflag=fullblock
0+5 records in
0+5 records out
10737397760 bytes (11 GB, 10 GiB) copied, 34.8157 s, 308 MB/s
```
```
# cat io.max 
8:0 rbps=367001600 wbps=314572800 riops=max wiops=max
```

components/ws-daemon/pkg/cgroup/plugin_iolimit_v2.go
```diff
@@ -6,138 +6,140 @@ package cgroup
...
-		log.WithField("cgroupPath", cgroupPath).Debug("starting io limiting")
-		// We are racing workspacekit and the interaction with disks.
-		// If we did this just once there's a chance we haven't interacted with all
-		// devices yet, and hence would not impose IO limits on them.
-		ticker := time.NewTicker(30 * time.Second)
-		defer ticker.Stop()
+		log.WithField("cgroupPath", cgroupPath).Info("starting io limiting")
 
-		ioMaxFile := filepath.Join(basePath, cgroupPath)
+		_, err := v2.NewManager(basePath, filepath.Join("/", cgroupPath), c.limits)
```
utam0k, line 62: We only want to write the value to `io.max` , right? Why do you use contained Manager?

utam0k: It was a little difficult to know what this code means because I had to assume some kinds of things.

utam0k: I actually mistakenly thought this was updating the device filter and changing the eBPF. However, I admit I was a little dumb.

utam0k:
The behavior itself is the charm. Thank you @aledbf

Furisto (approved):

akosyakov (approved):

PR #10674, opened 2022-06-15
[usage] Lookup stripe customers for each team in a usage report

andrew-farries:
## Description

Add a `stripe` package to the usage reconciler and use it to find all Stripe `Customers` who have registered usage in the preceding month.

By querying the Stripe API for all teamIds occurring in the usage report, we are able to find a list of Stripe `Customers` who should have their subscriptions updated to register usage.

This PR just finds and logs the `Customers` occurring in the usage report. In a subsequent PR these `Customers` will be updated to register their usage.

## Related Issue(s)
Part of #9036 

## How to test

* Get a kube context for the staging cluster (`gcloud auth login` then `gcloud container clusters get-credentials ...`) - the exact commands can be taken from the GCP console.

* Port forward to the staging database:

```
kubectl port-forward svc/cloudsqlproxy 3306:3306
```

* Get some Stripe test mode API keys from the Stripe dashboard and save them in `components/usage/apikeys`. The file should look like:

```json
{
  "publishableKey": "foo",
  "secretKey": "bar"
}
```

* Run the usage reconciler against the staging database:

```
DB_HOST=localhost DB_USERNAME=gitpod DB_PASSWORD=<> DB_PORT=3306 go run . run --api-key-file ./apikeys
```
(the env vars, including password, can be found in the `server` pod by running):

```
kubectl exec server-849f6f8895-8kct9 -- env | grep DB_
```

After 1 minute the reconciler will run and display output like:

```json
{"level":"info","message":"about to make query \"metadata['teamId']:'0163ce3d-d46d-4c71-9d23-80d5e014959b' OR metadata['teamId']:'2db3d0ae-9fa2-41ed-9781-acfc7e17294b' OR metadata['teamId']:'0d39eefc-d1b7-461d-a4e5-03826b3dd054' OR metadata['teamId']:'0a6486da-b080-4f66-ba86-98f5a03374b5' OR metadata['teamId']:'14f79ba1-6e24-46de-aa80-12f63e58ea5f' OR metadata['teamId']:'7981ba4e-68fe-4d0a-afa1-e21b428bd7d0' OR metadata['teamId']:'af870004-62e5-4b44-9395-cd37abb10e2e' OR metadata['teamId']:'58ba9114-49ee-4538-8550-6f2bace66ac4'\"","serviceContext":{"service":"usage","version":""},"severity":"INFO","time":"2022-06-15T06:36:28Z"}
{"level":"info","message":"found customer \"Test Customer (for invoice generation)\" for teamId \"2db3d0ae-9fa2-41ed-9781-acfc7e17294b\"","serviceContext":{"service":"usage","version":""},"severity":"INFO","time":"2022-06-15T06:36:28Z"}

```

Showing the query that was run against the Stripe API and the 1 customer that was found with a matching teamId.

## Release Notes

```release-note
NONE
```

- [x] /werft with-payment

### DIFF
--- a/components/usage/.gitignore
+++ b/components/usage/.gitignore
@@ -1,0 +1,1 @@
+apikeys
--- a/components/usage/pkg/stripe/stripe.go
+++ b/components/usage/pkg/stripe/stripe.go
@@ -10,7 +10,9 @@ import (
 	"os"
 	"strings"
 
+	"github.com/gitpod-io/gitpod/common-go/log"
 	"github.com/stripe/stripe-go/v72"
+	"github.com/stripe/stripe-go/v72/customer"
 )
 
 type stripeKeys struct {
@@ -32,6 +34,21 @@ func Authenticate(apiKeyFile string) error {
 
 	stripe.Key = stripeKeys.SecretKey
 	return nil
+}
+
+// FindCustomersForTeamIds queries the stripe API to find all customers with a teamId in `teamIds`.
+func FindCustomersForTeamIds(teamIds []string) {
+	queries := queriesForCustomersWithTeamIds(teamIds)
+
+	for _, query := range queries {
+		log.Infof("about to make query %q", query)
+		params := &stripe.CustomerSearchParams{SearchParams: stripe.SearchParams{Query: query}}
+		iter := customer.Search(params)
+		for iter.Next() {
+			customer := iter.Customer()
+			log.Infof("found customer %q for teamId %q", customer.Name, customer.Metadata["teamId"])
+		}
+	}
 }
 
 // queriesForCustomersWithTeamIds constructs Stripe query strings to find the Stripe Customer for each teamId
--- a/components/usage/pkg/stripe/stripe.go
+++ b/components/usage/pkg/stripe/stripe.go
@@ -1,0 +1,33 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package stripe
+
+import (
+	"encoding/json"
+	"os"
+
+	"github.com/stripe/stripe-go/v72"
+)
+
+type stripeKeys struct {
+	PublishableKey string `json:"publishableKey"`
+	SecretKey      string `json:"secretKey"`
+}
+
+func Authenticate(apiKeyFile string) error {
+	bytes, err := os.ReadFile(apiKeyFile)
+	if err != nil {
+		return err
+	}
+
+	var stripeKeys stripeKeys
+	err = json.Unmarshal(bytes, &stripeKeys)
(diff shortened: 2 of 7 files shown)


jankeromnes:
Many thanks! Will take a look. 👀

components/usage/.gitignore
```diff
@@ -0,0 +1 @@
+apikeys
```
jankeromnes, line 1: Interesting -- is placing an untracked, secret file next to the `cmd` the recommended way to provide credentials to a Go component? 👀 (I haven't worked on Go component secrets before.)

components/usage/pkg/stripe/stripe.go
```diff
@@ -0,0 +1,74 @@
...
+			log.Infof("found customer %q for teamId %q", customer.Name, customer.Metadata["teamId"])
+		}
+	}
+}
+
+// queriesForCustomersWithTeamIds constructs Stripe query strings to find the Stripe Customer for each teamId
+// It returns multiple queries, each being a big disjunction of subclauses so that we can process multiple teamIds in one query.
+// `clausesPerQuery` is a limit enforced by the Stripe API.
+func queriesForCustomersWithTeamIds(teamIds []string) []string {
+	const clausesPerQuery = 10
```
jankeromnes, line 58: Interesting. I guess you found this limit by testing. Is the value `10` documented somewhere? 👀

components/usage/.gitignore
andrew-farries: I think this way works well here. When the component runs in a pod it takes the credentials file in the same way (file on disk, location passed via `--api-keys-file`).

How would you expect the necessary api keys to be provided when running the usage component locally?

components/usage/pkg/stripe/stripe.go
andrew-farries: Not documented anywhere I could find. It appeared in an error message when trying larger queries.

components/usage/.gitignore
andrew-farries: Makes sense. I think @easyCZ is keen to move away from requiring preview environments to develop components. When working on the usage component I think the intention is to run it locally, port forwarding to a database as necesary.

components/usage/pkg/stripe/stripe.go
andrew-farries: Yes, I wondered about differences between live vs staging too.

We will likely need to rate limit our Stripe API calls in production in general. For this search endpoint the rate limit is:

> Rate limits
We apply a [rate limit](https://stripe.com/docs/rate-limits) of up to 20 read operations per second which applies for all search endpoints in both live mode and test mode. Live mode and test mode limits are separate.

jankeromnes (approved):
Works like a charm, thanks! 🎉

Tested by setting up usage-based billing for a team, running a few workspaces, and checking the `usage` component logs (it found the customer).

PR #10675, opened 2022-06-15
Improve license API

AlexTugarev:
```release-note
NONE
```


components/dashboard/src/App.tsx
```diff
@@ -183,14 +181,6 @@ function App() {
                     }
                 }
                 setTeams(teams);
-
-                if (user?.rolesOrPermissions?.includes("admin")) {
-                    const adminSettings = await getGitpodService().server.adminGetSettings();
-                    setAdminSettings(adminSettings);
-
-                    var license: LicenseInfo = await getGitpodService().server.adminGetLicense();
-                    setLicense(license);
```
geropl, line 192: We never used those? :exploding_head:

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -2650,6 +2650,10 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
     }
 
     async adminGetLicense(ctx: TraceContext): Promise<LicenseInfo> {
+        traceAPIParams(ctx, {});
```
geropl, line 2653: :100:

components/dashboard/src/App.tsx
geropl: Potential problem: https://werft.gitpod-dev.com/job/gitpod-build-at-admingetlicense.0#build:components/dashboard:app

AlexTugarev: They are, but in Settings.tsx/License.tsx, where there is another call already. 

Thanks for the pointer to the build issue. Fixed.

PR #10678, opened 2022-06-15
Add `--schedule` flag to `usage run` command

andrew-farries:
## Description

⚠️ This is based on https://github.com/gitpod-io/gitpod/pull/10674. Review and merge that one first ⚠️ 

Add a `--schedule` flag to the `usage run` command to allow setting the schedule on which the reconciler runs.

The main motivation here is to make local development easier (don't have to wait 1 min after invoking the command to get a reconciliation run), but we'll probably want to set this in-cluster via a configmap later on.

## Related Issue(s)
Part of #9036 

## How to test

```bash
DB_HOST=localhost DB_USERNAME=gitpod DB_PASSWORD=foo DB_PORT=3306 go run . run --api-key-file ./apikeys --schedule 5s
```

Reconciliation now runs after 5 seconds. Without the flag, reconciliation runs after 1m.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/usage/cmd/run.go
+++ b/components/usage/cmd/run.go
@@ -24,6 +24,7 @@ func run() *cobra.Command {
 	var (
 		verbose    bool
 		apiKeyFile string
+		schedule   time.Duration
 	)
 
 	cmd := &cobra.Command{
@@ -48,7 +49,7 @@ func run() *cobra.Command {
 				log.WithError(err).Fatal("Failed to initialize stripe client.")
 			}
 
-			ctrl, err := controller.New(1*time.Minute, controller.NewUsageReconciler(conn))
+			ctrl, err := controller.New(schedule, controller.NewUsageReconciler(conn))
 			if err != nil {
 				log.WithError(err).Fatal("Failed to initialize usage controller.")
 			}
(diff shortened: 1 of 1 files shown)


components/usage/cmd/run.go
```diff
@@ -72,6 +73,7 @@ func run() *cobra.Command {
 	}
 
 	cmd.Flags().BoolVar(&verbose, "verbose", false, "Toggle verbose logging (debug level)")
+	cmd.Flags().DurationVar(&schedule, "schedule", 1*time.Minute, "The schedule on which the reconciler should run")
```
jankeromnes, line 76: Side-note: Maybe also increase the default reconciliation loop period to 30min or 1h? (I'm a little worried about high-frequency logging noise)

jankeromnes:
@andrew-farries The `usage` component now crash-loops in preview environments -- that's not expected right?

```
{"@type":"type.googleapis.com/google.devtools.clouderrorreporting.v1beta1.ReportedErrorEvent","error":"open /stripe-secret/apikeys: no such file or directory","level":"fatal","message":"Failed to initialize stripe client.","serviceContext":{"service":"usage","version":"commit-0f56f84b454fb6e5a3daecf0fe4e0a0a1c3551db"},"severity":"CRITICAL","time":"2022-06-16T15:02:38Z"}
```

I wonder if we should do an extra step so that the Stripe secret gets mounted as an `apikeys` file in the component. 🤔

components/usage/cmd/run.go
andrew-farries: Rebased to change the default frequency to 1 hour.

andrew-farries:
The stripe secret is already mounted into the usage component (https://github.com/gitpod-io/gitpod/pull/10631)

The problem is that this PR doesn't set `with-payment=true`, without which the stripe secret isn't mounted.

jankeromnes (approved):
Looks good to me and works like a charm! ✨

Many thanks. 🙏

Issue #9075, opened 2022-04-01
Epic: Local preview installation method


lucasvaltl:
### Summary

Create a way to install a basic version of self-hosted Gitpod on a local machine for trial purposes. Should require minimal effort (time to wow) for most somewhat technical enterprise users.

### Context

The easiest way to try out Gitpod is via our SaaS offering (gitpod.io). However, there are users and companies where this is not possible because of internal compliance issues and other restrictions (e.g. can only use vetted providers of software if it is running somewhere else). For these, Self-Hosted is the only viable path. However, our [current self-hosted installation method](https://www.gitpod.io/docs/getting-started) requires you to set up an entire Kubernetes cluster just to try out Gitpod for the first time and see what it is all about. In a large corporation (exactly the ones that cannot use the SaaS version) this might mean going through internal hurdles to provision said cluster. Further this requires k8s experience.

We want to create a way for these corporate users that are technical but may not have deep Kubernetes expertise to experience Gitpod before they go to the next phase of investing more into it and getting for example a team to spin up a k8s cluster and install Gitpod self-hosted. 

### Value

- Enables those that cannot or do not want to run Gitpod SaaS to try out Gitpod, (hopefully) be amazed by it and potentially become full time users, and then potentially tell their friends / colleagues about it and start pushing for a wider implementation. 

### Acceptance Criteria

- Gitpod can be installed via simple and straight forward way that most somewhat technical enterprise users can successfully go through. It is a tradeoff: minimizing dependencies, minimizing installation steps vs. maximizing amount of addressable target users. 

### Measurement

- Amount of new leads we get from users that first used this installation method should be going up dramatically
- Amount of support questions we get related to basic K8s concepts should go down 

### Complexities

- Runtime: docker vs. vm 
- Certs
- DNS resolution: localhost, DNS masks etc.

### Related Issues / Work
- https://github.com/MrSimonEmms/gitpod-single-instance
- https://github.com/gitpod-io/gitpod/tree/main/contrib/docker
- https://github.com/gitpod-io/gitpod/issues/7315#issuecomment-998393009

csweichel:
Current work: https://github.com/gitpod-io/preview-install

meysholdt:
On the Platform team, we used a Hack Friday to explore solutions around DNS and certs. Find the notes here:
https://www.notion.so/gitpod/Platform-Hack-Friday-3341419375c0499891499da281294151

JohannesLandgraf:
[...]

PR #10695, opened 2022-06-16
[preview-install] Add user-friendly output

Pothulapati:
## Description

This PR adds user friendly output to the `preview` install
docker container.

## Related Issue(s)

Part of https://github.com/gitpod-io/gitpod/issues/9075

## Release Notes

```release-note
```


roboquat:

<details>

</details>

install/preview/entrypoint.sh
```diff
@@ -111,10 +115,16 @@ done
...
 /gitpod-installer render --config config.yaml --output-split-files /var/lib/rancher/k3s/server/manifests/gitpod
+
+# store files in `gitpod.debug` for debugging purposes
 for f in /var/lib/rancher/k3s/server/manifests/gitpod/*.yaml; do (cat "$f"; echo) >> /var/lib/rancher/k3s/server/gitpod.debug; done
+# remove NetowrkPolicy resources as they are not relevant here
 rm /var/lib/rancher/k3s/server/manifests/gitpod/*NetworkPolicy*
+# update PersistentVolumeClaim's to use k3s's `local-path` storage class
 for f in /var/lib/rancher/k3s/server/manifests/gitpod/*PersistentVolumeClaim*.yaml; do yq e -i '.spec.storageClassName="local-path"' "$f"; done
+# Set `volumeClassTemplate` so that each replica creates its own PVC
 yq eval-all -i ". as \$item ireduce ({}; . *+ \$item)" /var/lib/rancher/k3s/server/manifests/gitpod/*_StatefulSet_messagebus.yaml /app/manifests/messagebus.yaml
```
Pothulapati, line 126: Hmm, Guessing we might not actually need this, will try and revert back!

install/preview/entrypoint.sh
```diff
@@ -5,6 +5,10 @@
 
 set -ex
 
+if [ "$1" != "logging" ]; then
```
adrienthebo, line 8: We're leaking a small amount of logs with the `set -x` on line 6, can we move that `set -x` as an else clause to this conditional?

```
❯ docker run --privileged --name gitpod --rm -it -v /tmp/gitpod:/var/gitpod eu.gcr.io/gitpod-core-dev/build/preview-install:tar-preview-output.5
+ '['  '!=' logging ]
+ /entrypoint.sh logging
+ /prettylog
# [pretty logs follow]
```

install/preview/entrypoint.sh
```diff
@@ -5,6 +5,10 @@
...
+  $0 logging 2>&1 | /prettylog
```
adrienthebo, line 9: If I'm reading this correctly, we're re-launching this script in order to redirect all logs to `/prettylog`, correct?

If we are relaunching the script with the intent described above, we can use `exec` to redirect stdout/stderr, with something like this:

```
exec > >(/prettylog) 2>&1
```

This way we don't need to relaunch the script; we can launch `/prettylog` with [process substitution](https://tldp.org/LDP/abs/html/process-sub.html) and [redirect stderr/stdout with exec](https://tldp.org/LDP/abs/html/x17974.html).

This is another approach to the problem, but isn't necessarily better or worse. If you like the above approach we can adopt it; if what you have is more readable then let's keep it as-is. 😄

adrienthebo: Following up on this, we're also relying on the `set -e` to exit this script upon error, and we're relying on `/entrypoint.sh` and `prettylog` to exit with a non-zero exit value. When the user hits ctrl-c to kill the container, the forked processes exit non-zero then the behavior of `set -e` in the parent process kicks in and terminates the script.

If the `set -e` is removed for whatever reason, then when the forked processes exit we'll proceed to run this script again, which isn't intended. I think this code will be a little bit more correct to have an `exit` after `$0 logging`; that way changes in exit codes or other behaviors won't accidentally run the script twice.

Issue #10041, opened 2022-05-16
Improve prebuild-in-progress page layout and actions as well as make workspace start sequence clearer


gtsiolis:
### Problem to solve

Following up from https://github.com/gitpod-io/gitpod/pull/9904#discussion_r871268558, the **prebuild-in-progress** phase:
1. Happens before the workspace loading phases (Checking, Creating, & Starting), however, it's not clear if it takes place before or after the rest of the workspace start phases.
2. Does not re-use the prebuild log output component from the prebuild page, see https://github.com/gitpod-io/gitpod/issues/9132.
3. Currently hijacks the title element from the existing loading phases of the workspace start page to indicate it's about the prebuild-in-progress phase.
4. Does not feel connected with the rest of the workspace start phases.

### Proposal

Since the existing loading phases of the workspace start page aim to abstract the sub-states of the loading cycle to make the loading friendlier to the user, we could consider 🅰️ reusing the same element from the prebuilds pages (see https://github.com/gitpod-io/gitpod/issues/9132) where we could surface these sub-states of the prebuild phase and 🅱️ provide better visibility of the step sequence.

| BEFORE | AFTER |
|-|-|
| <img width="1440" alt="loading-0" src="https://user-images.githubusercontent.com/120486/168054711-fd0530ba-405f-421f-8811-8140cc4cc304.png"> | <img width="1440" alt="loading-5" src="https://user-images.githubusercontent.com/120486/168062587-be9cc612-0b66-4c29-ad86-36f82c9bd453.png"> |

In short, to break things down into smaller minimum viable changes (MVC) if needed:
1. Introduce a _Preparing_ loading phase for the prebuild-in-progress phase
2. Additionally, consider re-using the prebuild log output element for better associating prebuilds across the product as well as surface potential sub-states of the prebuild-in-progress phase as well as improve the UX of the actions available to users, see https://github.com/gitpod-io/gitpod/issues/9132.

atduarte:
Just sharing my 2 cents as I saw this being shared: I like the additional info 🙌  although I'm concerned about reducing the width of the window with logs.

gtsiolis:
Thanks for the feedback, @atduarte! The width aspect has been brought up before in https://github.com/gitpod-io/gitpod/issues/9132#issuecomment-1090827146. Keeping the existing width can be also a good way forward and will be possible without further design specs as the component is flexible to change its width. 🤝 

[...]

PR #10696, opened 2022-06-16
[prebuilds] prebuild detail view in create workspace

### LINKED ISSUE #9132: Use the same prebuild view for Workspace start as for Prebuild Details

Originally raised in https://github.com/gitpod-io/gitpod/pull/8831#discussion_r842711075 by @gtsiolis.

| BEFORE | AFTER |
|-|-|
| Screenshot 2022-04-05 at 2 00 24 PM | Screenshot 2022-04-05 at 2 09 51 PM |

Doing this would normalize the UI elements between prebuilds and workspace starts (which require a prebuild)

laushinka:
## Description
This PR gives a perceived consistency between the create workspace page and the prebuild detail view.  

Screenshot 2022-06-22 at 00 36 13

## Related Issue(s)
Fixes #9132

## How to test
1. Make a commit to a project with a prebuild.
2. Go to the project page's Branches tab. Once the prebuild is shown as "Running", start a new workspace.
3. See the logs running and the prebuild status showing below.

## Release Notes

```release-note
Prebuild status is shown under the logs when starting a workspace.
```

## Documentation

- [x] /werft with-preview

### DIFF
--- a/components/dashboard/src/components/PrebuildLogs.tsx
+++ b/components/dashboard/src/components/PrebuildLogs.tsx
@@ -13,13 +13,17 @@ import {
     WorkspaceImageBuild,
     HEADLESS_LOG_STREAM_STATUS_CODE_REGEX,
     Disposable,
+    PrebuildWithStatus,
 } from "@gitpod/gitpod-protocol";
 import { getGitpodService } from "../service/service";
+import { PrebuildStatus } from "../projects/Prebuilds";
 
 const WorkspaceLogs = React.lazy(() => import("./WorkspaceLogs"));
 
 export interface PrebuildLogsProps {
-    workspaceId?: string;
+    workspaceId: string | undefined;
+    onIgnorePrebuild?: () => void;
+    children?: React.ReactNode;
 }
 
 export default function PrebuildLogs(props: PrebuildLogsProps) {
@@ -27,6 +31,7 @@ export default function PrebuildLogs(props: PrebuildLogsProps) {
     const [workspaceInstance, setWorkspaceInstance] = useState<WorkspaceInstance | undefined>();
     const [error, setError] = useState<Error | undefined>();
     const [logsEmitter] = useState(new EventEmitter());
+    const [prebuild, setPrebuild] = useState<PrebuildWithStatus | undefined>();
 
     useEffect(() => {
         const disposables = new DisposableCollection();
@@ -37,9 +42,14 @@ export default function PrebuildLogs(props: PrebuildLogsProps) {
             }
             try {
                 const info = await getGitpodService().server.getWorkspace(props.workspaceId);
+                const pbws = await getGitpodService().server.findPrebuildByWorkspaceID(props.workspaceId);
                 if (info.latestInstance) {
                     setWorkspace(info.workspace);
                     setWorkspaceInstance(info.latestInstance);
+                }
+                if (pbws) {
+                    const foundPrebuild = await getGitpodService().server.getPrebuild(pbws.id);
+                    setPrebuild(foundPrebuild);
                 }
                 disposables.push(
                     getGitpodService().registerClient({
@@ -56,6 +66,11 @@ export default function PrebuildLogs(props: PrebuildLogsProps) {
                                 return;
                             }
                             logsEmitter.emit("logs", content.text);
+                        },
+                        onPrebuildUpdate(update: PrebuildWithStatus) {
+                            if (update.info && update.info.buildWorkspaceId === props.workspaceId) {
+                                setPrebuild(update);
+                            }
                         },
                     }),
                 );
@@ -91,9 +106,16 @@ export default function PrebuildLogs(props: PrebuildLogsProps) {
     }, [props.workspaceId, workspaceInstance?.status.phase]);
 
     return (
-        <Suspense fallback={<div />}>
-            <WorkspaceLogs classes="h-full w-full" logsEmitter={logsEmitter} errorMessage={error?.message} />
-        </Suspense>
+        <>
+            <Suspense fallback={<div />}>
+                <WorkspaceLogs logsEmitter={logsEmitter} errorMessage={error?.message} />
+            </Suspense>
+            <div className="h-20 px-6 border-gray-200 dark:border-gray-600 flex space-x-2">
+                {prebuild && <PrebuildStatus prebuild={prebuild} />}
+                <div className="flex-grow" />
+                {props.children}
+            </div>
+        </>
     );
 }
 
--- a/components/dashboard/src/components/PrebuildLogs.tsx
+++ b/components/dashboard/src/components/PrebuildLogs.tsx
@@ -106,16 +106,18 @@ export default function PrebuildLogs(props: PrebuildLogsProps) {
     }, [props.workspaceId, workspaceInstance?.status.phase]);
 
     return (
-        <>
-            <Suspense fallback={<div />}>
-                <WorkspaceLogs logsEmitter={logsEmitter} errorMessage={error?.message} />
-            </Suspense>
-            <div className="h-20 px-6 border-gray-200 dark:border-gray-600 flex space-x-2">
+        <div className="rounded-xl overflow-hidden bg-gray-100 dark:bg-gray-800 flex flex-col">
+            <div className="h-96 flex">
+                <Suspense fallback={<div />}>
+                    <WorkspaceLogs classes="h-full w-full" logsEmitter={logsEmitter} errorMessage={error?.message} />
+                </Suspense>
+            </div>
+            <div className="h-20 px-6 bg-gray-50 dark:bg-gray-800 border-t border-gray-200 dark:border-gray-600 flex space-x-2">
                 {prebuild && <PrebuildStatus prebuild={prebuild} />}
                 <div className="flex-grow" />
                 {props.children}
             </div>
-        </>
+        </div>
     );
 }
 
--- a/components/dashboard/src/projects/Prebuild.tsx
+++ b/components/dashboard/src/projects/Prebuild.tsx
@@ -13,7 +13,6 @@ import PrebuildLogs from "../components/PrebuildLogs";
 import Spinner from "../icons/Spinner.svg";
 import { getGitpodService, gitpodHostUrl } from "../service/service";
 import { TeamsContext, getCurrentTeam } from "../teams/teams-context";
-import { PrebuildStatus } from "./Prebuilds";
 import { shortCommitMessage } from "./render-utils";
 
 export default function () {
@@ -157,12 +156,7 @@ export default function () {
             <Header title={renderTitle()} subtitle={renderSubtitle()} />
             <div className="app-container mt-8">
                 <div className="rounded-xl overflow-hidden bg-gray-100 dark:bg-gray-800 flex flex-col">
-                    <div className="h-96 flex">
-                        <PrebuildLogs workspaceId={prebuild?.info?.buildWorkspaceId} />
-                    </div>
-                    <div className="h-20 px-6 bg-gray-50 dark:bg-gray-800 border-t border-gray-200 dark:border-gray-600 flex space-x-2">
-                        {prebuild && <PrebuildStatus prebuild={prebuild} />}
-                        <div className="flex-grow" />
+                    <PrebuildLogs workspaceId={prebuild?.info?.buildWorkspaceId}>
                         {["aborted", "timeout", "failed"].includes(prebuild?.status || "") || !!prebuild?.error ? (
                             <button
                                 className="flex items-center space-x-2"
@@ -195,7 +189,7 @@ export default function () {
                         ) : (
                             <button disabled={true}>New Workspace ({prebuild?.info.branch})</button>
                         )}
-                    </div>
+                    </PrebuildLogs>
                 </div>
             </div>
         </>
--- a/components/dashboard/src/projects/Prebuild.tsx
+++ b/components/dashboard/src/projects/Prebuild.tsx
@@ -155,42 +155,40 @@ export default function () {
         <>
             <Header title={renderTitle()} subtitle={renderSubtitle()} />
             <div className="app-container mt-8">
-                <div className="rounded-xl overflow-hidden bg-gray-100 dark:bg-gray-800 flex flex-col">
-                    <PrebuildLogs workspaceId={prebuild?.info?.buildWorkspaceId}>
-                        {["aborted", "timeout", "failed"].includes(prebuild?.status || "") || !!prebuild?.error ? (
-                            <button
-                                className="flex items-center space-x-2"
-                                disabled={isRerunningPrebuild}
-                                onClick={rerunPrebuild}
-                            >
-                                {isRerunningPrebuild && (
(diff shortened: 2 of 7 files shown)


roboquat:

<details>

</details>

components/dashboard/src/start/IntermediatePrebuildStatus.tsx
```diff
@@ -0,0 +1,22 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import StatusRunning from "../icons/StatusRunning.svg";
+
+function IntermediatePrebuildStatus() {
```
laushinka, line 9: To show the prebuild status, we need to find prebuilds related to a project. When showing the logs in the prebuild detail view, we already have the project information. This is not the case during workspace start. However, we do have a prebuild ID and status. 
This intermediate status component is a workaround so that we can show the prebuild status "Running" while waiting for the prebuild updates. Once the prebuild updates are received and we have a prebuild object, we show those accordingly.

laushinka: We don't need this intermediate status anymore because we are now fetching the prebuild from the beginning.

jldec:
From my experiments in this preview environment, the scrollbar in the project prebuild log view (1st screenshot below) is quite different from the scrollbar in the start workspace view with prebuild in progress. (2nd screenshot below)
 
Screenshot 2022-06-23 at 12 03 38

Screenshot 2022-06-23 at 12 04 06

jldec:
Other than the scrollbar difference, the behavior LGTM
/hold for more technical review

components/dashboard/src/components/PrebuildLogs.tsx
```diff
@@ -13,20 +13,27 @@ import {
...
 import { getGitpodService } from "../service/service";
+import { PrebuildStatus } from "../projects/Prebuilds";
 
 const WorkspaceLogs = React.lazy(() => import("./WorkspaceLogs"));
 
 export interface PrebuildLogsProps {
-    workspaceId?: string;
+    workspaceId: string | undefined;
+    prebuildId?: string | undefined;
+    logsEmitter?: EventEmitter;
```
geropl, line 26: This one looks like a copy'n'paste thing; it's not used (anymore).

laushinka: Thanks for the catch! Will remove it.

components/dashboard/src/components/PrebuildLogs.tsx
```diff
@@ -41,6 +48,10 @@ export default function PrebuildLogs(props: PrebuildLogsProps) {
                     setWorkspace(info.workspace);
                     setWorkspaceInstance(info.latestInstance);
                 }
+                if (props.prebuildId) {
```
geropl, line 51: We should always be able to retrieve the prebuild shape. For that we could use a method called `findPrebuildByWorkspaceId(workspaceId)`.

components/dashboard/src/components/PrebuildLogs.tsx
```diff
@@ -57,6 +68,11 @@ export default function PrebuildLogs(props: PrebuildLogsProps) {
                             }
                             logsEmitter.emit("logs", content.text);
                         },
```
geropl, line 70: This will get triggered by _any_ prebuild this user has access to.
Maybe a simple `if (update.info && update.info.buildWorkspaceId === props.workspaceId)` already is enough.

components/dashboard/src/projects/Prebuild.tsx
```diff
@@ -157,12 +156,7 @@ export default function () {
             <Header title={renderTitle()} subtitle={renderSubtitle()} />
             <div className="app-container mt-8">
                 <div className="rounded-xl overflow-hidden bg-gray-100 dark:bg-gray-800 flex flex-col">
-                    <div className="h-96 flex">
-                        <PrebuildLogs workspaceId={prebuild?.info?.buildWorkspaceId} />
-                    </div>
-                    <div className="h-20 px-6 bg-gray-50 dark:bg-gray-800 border-t border-gray-200 dark:border-gray-600 flex space-x-2">
-                        {prebuild && <PrebuildStatus prebuild={prebuild} />}
-                        <div className="flex-grow" />
+                    <PrebuildLogs workspaceId={prebuild?.info?.buildWorkspaceId} prebuildId={prebuild?.info?.id}>
```
geropl, line 159: :orange_heart:

components/dashboard/src/start/CreateWorkspace.tsx
```diff
@@ -490,17 +488,15 @@ function RunningPrebuildView(props: RunningPrebuildViewProps) {
...
-                }}
+            <PrebuildLogs
+                workspaceId={props.runningPrebuild.workspaceID}
+                onIgnorePrebuild={props.onIgnorePrebuild}
+                prebuildId={props.runningPrebuild.prebuildID}
             >
-                Don't Wait for Prebuild
-            </button>
+                <button className="secondary" onClick={() => props.onIgnorePrebuild && props.onIgnorePrebuild()}>
+                    Skip Prebuild
```
geropl, line 497: Why change the text here?
Personally I don't have an opinion, but these texts tend to have been checked by other people having a better overview over UX then myself. :wink:

components/dashboard/src/start/StartWorkspace.tsx
```diff
@@ -391,8 +392,8 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
 
     render() {
         const { error } = this.state;
-        const isHeadless = this.state.workspace?.type !== "regular";
-        const isPrebuilt = WithPrebuild.is(this.state.workspace?.context);
+        const isPrebuild = this.state.workspace?.type !== "regular";
```
geropl, line 395: Should be: `const isPrebuild = this.state.workspace?.type === "prebuild"`

components/dashboard/src/start/StartWorkspace.tsx
```diff
@@ -391,8 +392,8 @@ export default class StartWorkspace extends React.Component<StartWorkspaceProps,
...
+        const withPrebuild = WithPrebuild.is(this.state.workspace?.context);
```
geropl, line 396: :100:

components/dashboard/src/start/CreateWorkspace.tsx
laushinka: Good point! I just followed the text specified in the issue because I assumed that went through a design discussion already. Thoughts, @jldec?

components/server/src/projects/projects-service.ts
```diff
@@ -225,6 +225,18 @@ export class ProjectsService {
         return result;
     }
 
+    async getPrebuild(prebuildId: string): Promise<PrebuildWithStatus | undefined> {
```
geropl, line 228: This should not reside in ProjectService I feel. It has no connection to that, at all. I'd rather see it in GitpodServerImpl - or a PrebuildsService, if we were to introduce one.

components/dashboard/src/start/StartWorkspace.tsx
```diff
@@ -720,7 +731,7 @@ function ImageBuildView(props: ImageBuildViewProps) {
     );
 }
 
-function HeadlessWorkspaceView(props: { instanceId: string }) {
+function HeadlessWorkspaceView(props: { instanceId: string; workspaceId?: string }) {
```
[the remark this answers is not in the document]
laushinka: Oh right, that's a good point. Now everything it needs is handled already in PrebuildLogs.

components/server/src/projects/projects-service.ts
laushinka: My thought was because prebuilds are intended to only be used for projects, therefore I kept it here. But I can see why it seems out of place. I'll move it 👍🏼

components/dashboard/src/components/PrebuildLogs.tsx
laushinka: Currently the only PrebuildLogs needing a prebuild are CreateWorkspace and Prebuild, and both have this information already and it gets passed into the component as a prop. Do we need to make this extra call?

laushinka: Oh wait StartWorkspace also

geropl: > Do we need to make this extra call?

It's not an extra call. So far we doing 1 call (getPrebuild) sometimes, and I'm arguing for doing 1 call (findPrebuildByWorkspaceId) always.
If we wanted to optimize we could think about accepting a full `PrebuildInfo` shape to avoid rendering placeholders but we have to do that one call anyway to be sure we're up-to-date and avoid lost updates.

components/dashboard/src/start/CreateWorkspace.tsx
geropl: :bulb: Ah, didn't noticed!

components/server/src/projects/projects-service.ts
geropl: > I'll move it

:pray:

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -2238,11 +2238,17 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
...
 
-        const user = this.checkAndBlockUser("getPrebuilds");
+        const user = this.checkAndBlockUser("findPrebuilds");
         await this.guardProjectOperation(user, params.projectId, "get");
         return this.projectsService.findPrebuilds(params);
     }
 
+    public async getPrebuild(ctx: TraceContext, prebuildId: string): Promise<PrebuildWithStatus | undefined> {
+        traceAPIParams(ctx, { prebuildId });
+        this.checkUser("getPrebuild");
```
geropl, line 2248: One last - but important - thing: We need to guard access to a prebuild.

So far, we don't have an access guard for prebuilds specifically (I think we should have one).
As placeholder, we mostly used the [WorkspaceInstance one](https://github.com/gitpod-io/gitpod/blob/3389f9a91c2e9401c2e49d97cf5b7d241d4a37bf/components/server/src/workspace/gitpod-server-impl.ts#L1542-L1543), which is ok to re-use in the scope of this PR.

But we should create a follow-up issue to fix this. :+1:

components/dashboard/src/components/PrebuildLogs.tsx
laushinka: Incorporated `findPrebuildByWorkspaceId` and got rid of the `prebuildId` props 👍🏼

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -2238,11 +2238,34 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
...
-        const user = this.checkAndBlockUser("getPrebuilds");
+        const user = this.checkAndBlockUser("findPrebuilds");
         await this.guardProjectOperation(user, params.projectId, "get");
         return this.projectsService.findPrebuilds(params);
     }
 
+    public async getPrebuild(ctx: TraceContext, prebuildId: string): Promise<PrebuildWithStatus | undefined> {
+        traceAPIParams(ctx, { prebuildId });
+        this.checkAndBlockUser("getPrebuild");
+        const pbws = await this.workspaceDb.trace({}).findPrebuiltWorkspaceById(prebuildId);
```
geropl, line 2249: Thanks for adding this one! :pray:
Instead of `trace({})` we should forward the `ctx` I think.

components/dashboard/src/start/StartWorkspace.tsx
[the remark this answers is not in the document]
geropl: `HeadlessWorkspaceView` can be replace entirely I think. :+1:

components/server/src/workspace/gitpod-server-impl.ts
geropl: Ah, and we still need some form of access restriction, as mentioned [here](https://github.com/gitpod-io/gitpod/pull/10696#discussion_r904965697).

Copying over the lines from the headless-log method should be fine. :+1:

components/dashboard/src/start/StartWorkspace.tsx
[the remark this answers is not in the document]
laushinka: Ahhh I already removed this but then put it back for testing and mistakenly committed it

components/server/src/workspace/gitpod-server-impl.ts
laushinka: @geropl The existing access guards didn't work because they check for properties that the prebuildWithStatus didn't have, so I created a basic `GuardedPrebuild` resource. Tested the preview env and seems to work fine, but let me know what you think 🙏🏽

laushinka: Removed ✅

geropl:
This is basically done, and I'd love to approve. But as [we found out above](https://github.com/gitpod-io/gitpod/pull/10696#discussion_r906208239), we're missing proper resourceguards for prebuilds. So I'm going to hold this PR, and try to get it in before we land this.

/hold

geropl:
Now waiting for https://github.com/gitpod-io/gitpod/pull/10940 to get merged...

gtsiolis:
Looking at this now! 👀

components/server/src/workspace/gitpod-server-impl.ts
geropl: Fixed with https://github.com/gitpod-io/gitpod/pull/10696/commits/8f2b6c8be82dcff9c63830ef440e4f61f50464fc

gtsiolis (commented):
I never clicked submit but here's what I had in draft since yesterday. 🙈 

Since we do not have yet the capability to expand the logs into a full page component, going with the current logs width sounds ok. Any thoughts on https://github.com/gitpod-io/gitpod/issues/9132#issuecomment-1092072758?

Two issues that pop up:
1. The scrollbar position mentioned in https://github.com/gitpod-io/gitpod/pull/10696#issuecomment-1164275542. Moving forward with the current scrollbar styles sounds ok but the positioning the scrollbar in the middle of the element feels like something is broken.
2. The width of the element wrapping the bottom parts of the prebuilds logs differ per case. Could we adjust the width of the bottom elements to match the prebuilds logs element above? See screenshots below.

| Prebuild (Running) | Prebuild (Preparing) |
|-|-|
| in-progress-running | in-progress-preparing |

geropl (approved):
Tested and works! :+1: 

/hold in cases someone else wants to review

gtsiolis (approved):
Perfection achieved. ✨ 

Thanks, @geropl, @laushinka, @jankeromnes, and @jldec! 🏀

While I still think using a narrower width as described in https://github.com/gitpod-io/gitpod/issues/9132#issuecomment-1092072758 could be interesting to explore, :a: the changes here already significantly improve the current state of the prebuild-in-progress page and :b: we'd still need to add a function to expand the view to a full page component which would significantly increase the scope of this PR.

Looking forward for https://github.com/gitpod-io/gitpod/issues/10041 or whatever comes next.

Issue #10499, opened 2022-06-07
Keep admin menu active when opening a user, project, or team


gtsiolis:
### Problem to solve

Following up from https://github.com/gitpod-io/gitpod/pull/10309, currently when opening a user, project, or team in the `/admin` area the menu highlight disappears.

| `/admin/teams` | `/admin/teams/c5895528-xyz` |
|-|-|
| <img width="1105" alt="Screenshot 2022-06-07 at 12 18 46 PM (2)" src="https://user-images.githubusercontent.com/120486/172345240-0cdd9d95-a458-47bb-94cc-8565f46a7f8a.png"> | <img width="1105" alt="Screenshot 2022-06-07 at 12 20 14 PM (2)" src="https://user-images.githubusercontent.com/120486/172345246-9916b399-0b87-411e-b1e5-deb49662fb0e.png"> |

Pasting below a relevant line code that could help:

https://github.com/gitpod-io/gitpod/blob/0a6e84e0af629c3066e27726e599d417744d17a4/components/dashboard/src/Menu.tsx#L110

### Proposal

The menu item should remain active also when opening a user, project, or team in the `/admin` area.

Devansu-Yadav:
@gtsiolis Is this issue still prevalent, considering that we addressed this as part of #16307?

gtsiolis:
Thanks, @Devansu-Yadav! It definitely feels less critical and will be irrelevant if we merge the changes in https://github.com/gitpod-io/gitpod/pull/16423. Let's close this! 📕

PR #10698, opened 2022-06-16
fix(dashboard): Tab menu visible for urls trailing '/'

### LINKED ISSUE #10666: Visiting the workspaces list with a trailing slash character in the URL breaks the new navigation
### Bug description

See [relevant discussion](https://gitpod.slack.com/archives/C01KPEPQYE6/p1655189692684969) (internal). Thanks for reporting, @jankeromnes!

| WIth `/` :x: | Without `/` ✔️ |
|-|-|
| Screenshot 2022-06-14 at 08 53 31 | Screenshot 2022-06-14 at 08 53 43 |

### Steps to reproduce

Try going to :a: [`/workspaces`](https://gitpod.io/workspaces) and then :b: [`/workspaces/`](https://gitpod.io/workspaces/).

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

CuriousCorrelation:
## Description

This bug #10666 was due to this

https://github.com/gitpod-io/gitpod/blob/402894b63b604234899db345fca0a9ce3750cec1/components/dashboard/src/Menu.tsx#L109

failing if `url` contained trailing `'/'`.

Here's a test

``` js
var url = new URL('https://gitpod.io/workspaces/')
>> ["/workspaces"].includes(url.pathname)
<- false
```
So if that was true, it should also happen to 

https://github.com/gitpod-io/gitpod/blob/402894b63b604234899db345fca0a9ce3750cec1/components/dashboard/src/Menu.tsx#L108

which was indeed the case, see https://gitpod.io/new/ (notice trailing '/').

A simple solution would have been to include another element (`"workspaces/")` in the array or discard trailing '/' and check with `includes`. Adding a new element can potentially cause similar problem if we were to forget duplicating for some new path. For the second option, I first thought of regex but it got really unreadable, so instead I chose `split` and `filter`

``` js
url.pathname.split('/').filter(p => p).join('/')
```
This would work but what if we are on a resource but not exactly on base? For example

(Browser console)
``` js
const url1 = new URL('https://gitpod.io/workspaces/testing')
var testingSubpath = url1.pathname.split('/').filter(p => p).join('/')
testingSubpath
"workspaces/testing"
```

``` js
var testing = ["workspaces", "teams/new", "open"]
```
Tab menu shouldn't be displayed in path `workspace/testing` because it is `isWorkspacesUI` (a subdirectory)

``` js
testing.includes(testingSubpath)
false
``` 
A better way would be to test "subpath" against "rootpath"

(Browser console)
``` js
var url1 = new URL("https://gitpod.io/teams/new/testing")
var trimmedPath = url1.pathname.split('/').filter(Boolean).join('/')
trimmedPath
"teams/new/testing"
```
Now if we were to test against 

``` js
var testing = ["workspaces", "teams/new", "open"]
```
we would get an array

``` js
testing.map(r => trimmedPath.startsWith(r))
Array(3) [ false, true, false ]
```
on which we can do an exhaustive search like

``` js
testing.map(r => trimmedPath.startsWith(r)).some(Boolean)
true
```

This would also work for `isAdminUI` check

https://github.com/gitpod-io/gitpod/blob/402894b63b604234899db345fca0a9ce3750cec1/components/dashboard/src/Menu.tsx#L110

(Browser console)
``` js
const url = new URL('https://gitpod.io/admin/testing')
const isAdminUI = url.pathname.startsWith("/admin")
isAdminUI
true
var testing = ["admin"]
var trimmedPath = url.pathname.split('/').filter(Boolean).join('/')
trimmedPath
"admin/testing"
testing.map(r => trimmedPath.startsWith(r)).some(Boolean)
true
```
---
If this subpath check behaviour isn't exactly what we need, we can swap positions and get exact results

``` js
var testing = ["admin"]
testing.map(r => r.startsWith(testingSubpath)).some(Boolean)
false
testing.map(r => testingSubpath.startsWith(r)).some(Boolean)
true
```
## Related Issue(s)
Fixes  #10666

## How to test
Described above

## Release Notes
```release-note
Fixed Tab menu being visible for urls with trailing '/'
```

## Documentation
Does this PR require updates to the documentation at www.gitpod.io/docs?
* No

- [x] /werft with-preview

### DIFF
--- a/components/dashboard/src/Menu.tsx
+++ b/components/dashboard/src/Menu.tsx
@@ -25,7 +25,7 @@ import { getProjectSettingsMenu } from "./projects/ProjectSettings";
 import { ProjectContext } from "./projects/project-context";
 import { PaymentContext } from "./payment-context";
 import FeedbackFormModal from "./feedback-form/FeedbackModal";
-import { isGitpodIo } from "./utils";
+import { inResource, isGitpodIo } from "./utils";
 import { getExperimentsClient } from "./experiments/client";
 
 interface Entry {
@@ -105,9 +105,9 @@ export default function Menu() {
     }
 
     // Hide most of the top menu when in a full-page form.
-    const isMinimalUI = ["/new", "/teams/new", "/open"].includes(location.pathname);
-    const isWorkspacesUI = ["/workspaces"].includes(location.pathname);
-    const isAdminUI = window.location.pathname.startsWith("/admin");
+    const isMinimalUI = inResource(location.pathname, ["new", "teams/new", "open"]);
+    const isWorkspacesUI = inResource(location.pathname, ["workspaces"]);
+    const isAdminUI = inResource(window.location.pathname, ["admin"]);
 
     const [teamMembers, setTeamMembers] = useState<Record<string, TeamMemberInfo[]>>({});
     useEffect(() => {
--- a/components/dashboard/src/utils.test.ts
+++ b/components/dashboard/src/utils.test.ts
@@ -1,0 +1,29 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { inResource } from "./utils";
+
+test("inResource", () => {
+
+    // Given root path is a part of resources specified
+    expect(inResource("/app", ["new", "app", "teams"])).toBe(true);
+
+    // Given path is a part of resources specified
+    expect(inResource("/app/testing", ["new", "app", "teams"])).toBe(true);
+
+    // Empty resources
+    expect(inResource("/just/a/path", [])).toBe(false);
+
+    // Both resources starting with '/'
+    expect(inResource("/app", ["/app"])).toBe(true);
+
+    // Both resources ending with '/'
+    expect(inResource("app/", ["app/"])).toBe(true);
+
+    // Both resources containing path with subdirectories
+    expect(inResource("/admin/teams/someTeam/somePerson", ["/admin/teams"])).toBe(true);
+
+});
--- a/components/dashboard/src/utils.ts
+++ b/components/dashboard/src/utils.ts
@@ -57,3 +57,21 @@ export function isGitpodIo() {
         window.location.hostname.endsWith("gitpod-io-dev.com")
     );
 }
+
+function trimResource(resource: string): string {
+    return resource.split('/').filter(Boolean).join('/');
+}
+
+// Returns 'true' if a 'pathname' is a part of 'resources' provided.
+// `inResource("/app/testing/", ["new", "app", "teams"])` will return true
+// because '/app/testing' is a part of root 'app'
+//
+// 'pathname' arg can be provided via `location.pathname`.
+export function inResource(pathname: string, resources: string[]): boolean {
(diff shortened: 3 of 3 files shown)


laushinka:
Added werft with-preview in the description

laushinka:
Hi @CuriousCorrelation, sorry for the delay in reviewing. The build failed, and to make sure that we have everything up to date, would you mind rebasing to the latest in main? 🙏🏽

CuriousCorrelation:
@laushinka No worries at all! Branch should be updated now

laushinka:
Thank you for this contribution, @CuriousCorrelation!
It indeed fixed #10666. It doesn't yet resolve #10499. You're very welcome to pick that up for your next PR, or include it in this one :)

Before doing that, would you mind writing a simple test for `inResource()`? It should be very similar and as straightforward as [this test](https://github.com/gitpod-io/gitpod/blob/main/components/dashboard/src/App.test.ts).
This will also make it much easier to test the #10499 cases should you or someone else pick that one :)

Thanks again 🙏🏽

CuriousCorrelation:
Writing tests would be no trouble at all. I'll try to get it done as soon as possible.

CuriousCorrelation:
@laushinka Great idea for adding tests! Not only did it help me fix a couple of edge issues, I believe it also fixed #10499

The problem was the trim function's application on first argument and not on second; the list of resources, because both of them can have a trailing or leading '/'.

A couple of changes from the last commit:

1. Moved `inResource` to `utils.ts`. I figured that'd be a better location since it's not technically a 'component'.
2. Which meant we could add a `utils.test.ts` for placing `inResource`'s tests. Also easier to add more cases.

Let me know if these changed are fine, or if there's something I need to change.

laushinka:
> @laushinka Great idea for adding tests! Not only did it help me fix a couple of edge issues, I believe it also fixed #10499

That's awesome, @CuriousCorrelation! I'm on vacation right now, so let me tag my team @gitpod-io/engineering-webapp so someone could take over :) 
Otherwise I will pick this up again next Monday.

laushinka:
Hi @CuriousCorrelation, I'm back from vacation now - sorry that it's taking a long time. 

The build is failing because we're missing a license header for `utils.test.ts`. Would you mind doing the following:
1. Run `leeway run components:update-license-header` in the terminal
2. Rebase to main.

laushinka:
Connection got closed - running this again

/werft run

laushinka:
Not sure why builds are failing. Trying again..

/werft run

laushinka:
@CuriousCorrelation This looks good! I'm happy to approve this (finally). 

#10499 is not yet fixed, which I believe is due to [this line](https://github.com/gitpod-io/gitpod/blob/04b9a640240d01c243e397ef55bffae6a69c42c7/components/dashboard/src/components/PageWithSubMenu.tsx#L31). All that needs to be done I believe is replacing that check with your new logic 😉

I don't want to block this from being merged any further though because you've done a lot of work. You're very welcome to work on #10499 with the hint above.

Thanks for your contribution 🔥

laushinka (approved):

laushinka:
(Oops I should've asked you to squash. Note for next time!)

CuriousCorrelation:
> #10499 is not yet fixed, which I believe is due to this line.

Ah makes sense, thanks for the heads up!

PR #10717, opened 2022-06-17
[installer]: set minio azure image to last supported version

### LINKED ISSUE #10697: Minio gateway no longer supports Azure
### Bug description

When configuring the storage to use Azure, the Minio gateway errors with.

```
‘azure’ is not a minio sub-command. See ‘minio --help’.
```

This is a change in the Minio container brought in due to #10428. It appears that it no longer supports Azure https://github.com/bitnami/charts/issues/10258

### Steps to reproduce

Configure KOTS to use Azure as the object storage mechanism

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

mrsimonemms:
## Description

Azure support has been removed from the latest version of Minio gateway. If Azure is selected as the object store, this uses the last supported version of Minio gateway with Azure support.

This also means that the Azure gateway won't receive any updates. I think this is an acceptable compromise on the basis that the gateway and the object storage endpoints are separate binaries/codebases anyway. This will need a better fix in time (probably adding Azure as a native storage option in the `content-service`) as there will undoubtedly be divergences between the Helm chart and image. However, this will solve the problem in the short-term.

Proof that it's uploading to the blob store

image

## Related Issue(s)

Fixes #10697

## How to test

Deploy with an Azure object store

## Release Notes

```release-note
[installer]: set minio azure image to last supported version
```

### DIFF
--- a/install/installer/pkg/components/minio/azure/constants.go
+++ b/install/installer/pkg/components/minio/azure/constants.go
@@ -1,0 +1,8 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the MIT License. See License-MIT.txt in the project root for license information.
+
+package azure
+
+const (
+	ImageTag = "2022.4.29-debian-10-r0" // Last tag with Azure gateway support @link https://github.com/bitnami/charts/issues/10258
+)
--- a/install/installer/pkg/components/minio/azure/minio.go
+++ b/install/installer/pkg/components/minio/azure/minio.go
@@ -6,6 +6,7 @@ package azure
 
 import (
 	"fmt"
+
 	"github.com/gitpod-io/gitpod/installer/pkg/common"
 	"github.com/gitpod-io/gitpod/installer/pkg/helm"
 	"github.com/gitpod-io/gitpod/installer/third_party/charts"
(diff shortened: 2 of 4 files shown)


nandajavarma (approved):
Adding this change to the test environment fixes the issues!

khurrambilalaurecon:
Hi,
I'm still receiving the same error that azure is not a minio sub-command.
I'm running Minio in Azure App service (Web App), passed startup file as "gateway azure". Any hints here?


image

MrSimonEmms:
@khurrambilalaurecon what version of Gitpod are you using? This has been fixed for some time.

Also, please check the version of `minio` that is installed to your cluster

khurrambilalaurecon:
Hi @MrSimonEmms 
I'm deploying minio docker image on Azure App service (Web app). Tried doing the same on Azure container instance as well, I pulled the latest docker image still received the above error, any thoguhts?

Thanks for the feedback.

MrSimonEmms:
Are you running Gitpod? I don't mind helping, but this is the thread for Gitpod, not the Minio Gateway application - really, you should be on that project's issue page.

For the record, the problem is that you're running the latest Docker image - they removed it so you need to use the last version that has it in. Go through this PR's files and you'll see that I've changed to the last support version - use that and you'll be grand.

Issue #10720, opened 2022-06-17
Make the frequency at which the usage reconciler runs configurable


andrew-farries:
In https://github.com/gitpod-io/gitpod/pull/10678 we added the `--schedule` flag to the usage controller.

We should set this value to the desired schedule in preview/staging/prod by adding a configmap for the usage controller.

PR #10729, opened 2022-06-17
[usage] Add configmap for usage component to control run frequency

andrew-farries:
## Description

Make the frequency at which the usage component runs configurable via a configmap.

## Related Issue(s)
Fixes https://github.com/gitpod-io/gitpod/issues/10720
Part of #9036 

## How to test

* Open the preview environment.
* Edit the `usage-configmap` (`kubectl edit cm usage-config`) and change the `schedule` to `1m`.
* Kill the usage pod to restart it with the new config (`kubectl delete pod <usage pod>`).
* Tail the logs of the new usage pod, including timestamps (`kubectl logs -f --timestamps=true <usage-pod>`)
* Watch the logs of the usage component and see that it now runs at 1 minute intervals rather than the previous 1 hour.

## Release Notes

```release-note
NONE
```
* [x] /werft run with-payment=true

### DIFF
--- a/install/installer/pkg/components/usage/deployment.go
+++ b/install/installer/pkg/components/usage/deployment.go
@@ -4,6 +4,8 @@
 package usage
 
 import (
+	"fmt"
+
 	"github.com/gitpod-io/gitpod/common-go/baseserver"
 	"github.com/gitpod-io/gitpod/installer/pkg/cluster"
 	"github.com/gitpod-io/gitpod/installer/pkg/common"
@@ -77,6 +79,8 @@ func deployment(ctx *common.RenderContext) ([]runtime.Object, error) {
 							Image: ctx.ImageName(ctx.Config.Repository, Component, ctx.VersionManifest.Components.Usage.Version),
 							Args: []string{
 								"run",
+								"--schedule",
+								"$(RECONCILER_SCHEDULE)",
 							},
 							ImagePullPolicy: corev1.PullIfNotPresent,
 							Resources: common.ResourceRequirements(ctx, Component, Component, corev1.ResourceRequirements{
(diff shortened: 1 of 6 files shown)


install/installer/pkg/components/usage/deployment.go
```diff
@@ -77,6 +79,8 @@ func deployment(ctx *common.RenderContext) ([]runtime.Object, error) {
 							Image: ctx.ImageName(ctx.Config.Repository, Component, ctx.VersionManifest.Components.Usage.Version),
 							Args: []string{
 								"run",
+								"--schedule",
```
easyCZ, line 82: My 2 cents, we should move this into a config file just like the public API. I'm happy to approve here to unblock but longer term we're better off with most config in files.

easyCZ (approved):
/hold

In case you want to revisit how the config is provided (file instead of flag)

install/installer/pkg/components/usage/deployment.go
andrew-farries: You mean not have a `--schedule` flag at all but rather read the schedule from a config file?

I've tended to configure CLIs like this with multiple command line args, set via helm charts (or the installer here in the Gitpod case) in the past. What makes you prefer config files for this (not saying I disagree, just curious).

easyCZ: Generally the pattern used for other components. I think it also somewhat simplifies configuration for the installer and in general observing what a component is configured with (but that's debatable).

easyCZ: I think it's worth landing this is as is and align in subsequent PR.

andrew-farries: Made https://github.com/gitpod-io/gitpod/issues/10753 to track this.

corneliusludmann (approved):

Issue #10374, opened 2022-05-31
better .gitpod.yml generation


gengjiawen:
## Is your feature request related to a problem? Please describe
https://github.com/vuejs/core, when open this repo.
Generated `.gitpod.yml`
```yml
tasks:
  - init: npm install && npm run build
    command: npm run dev
```

## Describe the behaviour you'd like
Actually it's using `pnpm`, so it should be

```yml
tasks:
  - init: pnpm install 
    command: pnpm run dev
```

You may ask how should I know it's pnpm. It can infer from root `pnpm-lock.yaml`

Toanzzz:
> You may ask how should I know it's pnpm. It can infer from root pnpm-lock.yaml

Or from "packageManager" field in `package.json` file

```json

{
...
  "packageManager": "pnpm@7.1.0",
...
}
```

gengjiawen:
Not sure this implemented typescript or golang, if it's in typescript, I can help to create a PR. @pawlean

axonasif:
cc @jankeromnes

jankeromnes:
> Not sure this implemented typescript or golang, if it's in typescript, I can help to create a PR. @pawlean

@gengjiawen it is actually in TypeScript:

https://github.com/gitpod-io/gitpod/blob/9552b6008fd009a66914384311dcee00d9ca2a1a/components/server/src/config/config-inferrer.ts#L44-L47

I made a quick first attempt here: https://github.com/gitpod-io/gitpod/pull/10731 (but the build failed)

If you'd like to fix it @gengjiawen, please absolutely go ahead 😊 I'd be more than happy to review. ❤️

PR #10731, opened 2022-06-17
[.gitpod.yml generator] Use 'pnpm' package manager when there is a pnpm-lock.yaml file or the package.json specifies it

jankeromnes:
## Description

When auto-generating a `.gitpod.yml`, look for a `pnpm-lock.yaml` file in the root of the repository, or a `packageManager` entry in `package.json` that starts with `"pnpm"`. If it's a match, the wanted package manager is likely `pnpm` (instead of `npm` or `yarn`).

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/10374

## How to test

1. Open a repository using `pnpm` in Gitpod, which does not yet have a committed `.gitpod.yml`
2. The auto-generated `.gitpod.yml` should use `pnpm` (instead of `npm` or `yarn`)

## Release Notes

```release-note
[.gitpod.yml generator] Use 'pnpm' package manager when there is a pnpm-lock.yaml file or the package.json specifies it
```

### DIFF
--- a/components/server/src/config/config-inferrer.ts
+++ b/components/server/src/config/config-inferrer.ts
@@ -41,9 +41,12 @@ export class ConfigInferrer {
         if (!pckjsonContent) {
             return;
         }
-        let command: "yarn" | "npm" = "npm";
+        let command: "yarn" | "npm" | "pnpm" = "npm";
         if (await ctx.exists("yarn.lock")) {
             command = "yarn";
+        }
+        if (await ctx.exists("pnpm-lock.yaml") || pckjsonContent.packageManager.startsWith("pnpm")) {
+            command = "pnpm";
         }
         this.addCommand(ctx.config, command + " install", "init");
         try {


components/server/src/config/config-inferrer.ts
```diff
@@ -41,10 +41,13 @@ export class ConfigInferrer {
         if (!pckjsonContent) {
             return;
         }
-        let command: "yarn" | "npm" = "npm";
+        let command: "yarn" | "npm" | "pnpm" = "npm";
         if (await ctx.exists("yarn.lock")) {
             command = "yarn";
         }
+        if (await ctx.exists("pnpm-lock.yaml") || pckjsonContent.packageManager.startsWith("pnpm")) {
```
jankeromnes, line 48: FIXME:

```
src/config/config-inferrer.ts(48,66): error TS2339: Property 'packageManager' does not exist on type 'string'.
```

gengjiawen:
```ts
        const pckjsonContent = await ctx.read("package.json");
        if (!pckjsonContent) {
            return;
        }
        let command: "yarn" | "npm" | "pnpm" = "npm";
        if (await ctx.exists("yarn.lock")) {
            command = "yarn";
        }
        let pckjson;
        try {
            pckjson = JSON.parse(pckjsonContent);
        } catch (e) {
            console.log(e, pckjsonContent);
        }
        if ((await ctx.exists("pnpm-lock.yaml")) || pckjson?.packageManager?.startsWith("pnpm")) {
            command = "pnpm";
        }
        this.addCommand(ctx.config, command + " install", "init");
        if (pckjson?.scripts) {
            if (pckjson.scripts.build) {
                this.addCommand(ctx.config, command + " run build", "init");
            } else if (pckjson.scripts.compile) {
                this.addCommand(ctx.config, command + " run compile", "init");
            }
            if (pckjson.scripts.start) {
                this.addCommand(ctx.config, command + " run start", "command");
            } else if (pckjson.scripts.dev) {
                this.addCommand(ctx.config, command + " run dev", "command");
            } else if (pckjson.scripts.watch) {
                this.addCommand(ctx.config, command + " run watch", "command");
            }
        }
        this.addExtension(ctx, "dbaeumer.vscode-eslint");
```

components/server/src/config/config-inferrer.ts
```diff
@@ -41,29 +41,33 @@ export class ConfigInferrer {
...
             return;
         }
-        let command: "yarn" | "npm" = "npm";
+        let pckjson;
+        try {
+            pckjson = JSON.parse(pckjsonContent);
+        } catch (e) {
+            console.log(e, pckjsonContent);
+        }
+        let command: "yarn" | "npm" | "pnpm" = "npm";
```
andrew-farries, line 50: Is it worth extracting a type for this union now that it has 3 members?

andrew-farries (changes requested):
We should add a test for this to `config-inferrer.spec.ts` too.

gengjiawen:
> We should add a test for this to `config-inferrer.spec.ts` too.

Is there easy way mock ctx for now ?

gengjiawen:
@jankeromnes friendly ping

components/server/src/config/config-inferrer.ts
jankeromnes: Since this is used in a single place, I think it's okay to leave it like that for now.

andrew-farries (approved):

jankeromnes:
New tests run and pass:

```
[components/server:app]   config inferrer
[components/server:app]     ✓ [node] yarn
[components/server:app]     ✓ [node] npm
[components/server:app]     ✓ [node] pnpm
```

PR #10732, opened 2022-06-17
Introduce SLI for preview environment start

mads-hartmann:
NOTE: I have added the hold label because the build fails are we seem to have a problems with VMs right now.

## Description

This PR instruments our build job to include annotations on the root span that we can use as part of our ["Preview environments should start successfully" SLI](https://www.notion.so/gitpod/Preview-Environment-Service-Level-Indicators-SLIs-55b649ad17774f279c142af55ad2fec7#f483f73aa4974ad3b8b6825fa2f251fa).

These annotations are used to created a derived column in Honeycomb that represents our SLI. The structure of the `IF(condition, then-value, else-value)` - see [here](https://docs.honeycomb.io/working-with-your-data/use-advanced-operators/derived-columns/reference/#if) - additionally for SLIs in Honeycomb `true` means the event should count as a success, `false` means it should count as a failure, and `null` means the event isn't part of the SLI.

So in our case, we only count the event if `preview.gitpod_built_successfully` is `true` and the `preview.k3s_successfully_created` attribute exists. We consider it a success if the root span doesn't have an error set and `preview.k3s_successfully_created` is `true`. Otherwise it's a failure.

```
IF(
  AND(
    EQUALS($preview.gitpod_built_successfully, true),
    EXISTS($preview.k3s_successfully_created)
  ),
  AND(
    NOT($error),
    EQUALS($preview.k3s_successfully_created, true)
  ),
  null
)
```

## Related Issue(s)

Fixes https://github.com/gitpod-io/ops/issues/2728
Fixes https://github.com/gitpod-io/ops/issues/2729
Fixes https://github.com/gitpod-io/ops/issues/2731

## How to test

I started a job off the branch so it loaded my new TS code

```
 werft job run github -f
```

The VM happened to fail as we're overloaded right now, so that's great for testing 😅 See the trace [here](https://ui.honeycomb.io/gitpod/datasets/werft/trace/friXtpnmrq1?fields[]=c_name&fields[]=c_service.name&span=d46e2e9ac9fa9294) and screenshot of it below:
Screenshot 2022-06-17 at 15 37 12

Here is a [simple query](https://ui.honeycomb.io/gitpod/datasets/werft/result/FiNjDzu9f1i) showing the count of events grouped by the the SLI - there is a bunch of events that doesn't count as part of the SLI and one failure (no successes because Harvester is overloaded)

Screenshot 2022-06-17 at 15 48 27

Also created an SLO for the fun of it [here](https://ui.honeycomb.io/gitpod/datasets/werft/slo/zRm48xoKMXf/Preview-Start).

## Release Notes

```release-note
NONE
```

## Documentation

N/A

### DIFF
--- a/.werft/jobs/build/build-and-publish.ts
+++ b/.werft/jobs/build/build-and-publish.ts
@@ -16,7 +16,6 @@ export async function buildAndPublish(werft: Werft, jobConfig: JobConfig) {
     const {
         publishRelease,
         dontTest,
-        withContrib,
         retag,
         version,
         localAppVersion,
@@ -26,6 +25,9 @@ export async function buildAndPublish(werft: Werft, jobConfig: JobConfig) {
     } = jobConfig;
 
     const releaseBranch = jobConfig.repository.ref;
+
+    // We set it to false as default and only set it true if the build succeeds.
+    werft.rootSpan.setAttributes({ "preview.gitpod_built_successfully": false });
 
     werft.phase("build", "build running");
     const imageRepo = publishRelease ? "gcr.io/gitpod-io/self-hosted" : "eu.gcr.io/gitpod-core-dev/build";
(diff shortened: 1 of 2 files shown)


ArthurSens:
All code changes look good! Awesome to see SLOs already as well 😁 

I tried rebasing your PR to see if the Preview Problem was solved already, but I think we'll have to wait until next week to get this one merged 😕

meysholdt (approved):
🚀

PR #10735, opened 2022-06-17
Revert "Change disk bus from scsi to virtio to gain performance"

vulkoingim:
This reverts commit 9552b6008fd009a66914384311dcee00d9ca2a1a.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.werft/vm/manifests.ts
+++ b/.werft/vm/manifests.ts
@@ -73,7 +73,7 @@ spec:
             - name: system
               bootOrder: 1
               disk:
-                bus: virtio
+                bus: scsi
             - name: cloudinitdisk
               disk:
                 bus: virtio


vulkoingim:
I think this change caused preview envs to fail (well, the vm to fail to start)

vulkoingim:
/werft run github -j .werft/build.yaml -a updateGitHubStatus=gitpod-io/gitpod -a no-preview=true

jenting:
> I think this change caused preview envs to fail (well, the vm to fail to start)

🤔 I thought the change worked, but it seems to be no.

jenting (approved):

mads-hartmann:
I think what might have happened is that job protection tricked us again ☺️ the werft job we tested on in your PR would have run from main unless you manually triggered the job using the CLI. Could that have been what happened?

jenting:
Absolutely yes 🤦🏻🤦🏻🤦🏻

PR #10737, opened 2022-06-17
previews: Remote-write metrics to core-dev

ArthurSens:

## Description

Configure Prometheus in Preview Environments to remote-write metrics to core-dev, but only the ones needed for our SLIs.

## Related Issue(s)

Fixes https://github.com/gitpod-io/ops/issues/2764
Fixes https://github.com/gitpod-io/ops/issues/2766

## How to test

Unfortunately, we're still having some network issues between previews and core-dev, but once that is solved we should be able to query metrics from the core-dev Prometheus with the following query:

```
{environment="preview-environment"}
```

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/.werft/observability/monitoring-satellite.ts
+++ b/.werft/observability/monitoring-satellite.ts
@@ -82,6 +82,17 @@ export class MonitoringSatelliteInstaller {
                     requests: { memory: '200Mi', cpu: '50m' },
                 },
             },
+            remoteWrite: {
+                username: '${process.env.PROM_REMOTE_WRITE_USER}',
+                password: '${process.env.PROM_REMOTE_WRITE_PASSWORD}',
+                urls: ['https://prometheus.gitpod-dev.com/api/v1/write'],
+                writeRelabelConfigs: [{
+                    sourceLabels: ['__name__', 'job'],
+                    separator: ';',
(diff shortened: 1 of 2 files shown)


.werft/observability/monitoring-satellite.ts
```diff
@@ -82,6 +82,16 @@ export class MonitoringSatelliteInstaller {
                     requests: { memory: '200Mi', cpu: '50m' },
                 },
             },
+            remoteWrite: {
+                username: '${process.env.PROM_REMOTE_WRITE_USER}',
+                password: '${process.env.PROM_REMOTE_WRITE_PASSWORD}',
+                urls: ['https://prometheus.gitpod-dev.com/api/v1/write'],
+                writeRelabelConfigs: [{
+                    sourceLabels: ['__name__'],
+                    regex: 'probe_.*|rest_client_requests_total',
```
ArthurSens, line 91: I've just noticed that this regex drops the `up` metric, which is required [by SLI 2](https://github.com/gitpod-io/ops/issues/2765). 

The `up` metric is built from ALL targets, so I need to figure out a good regex that will only give us the up from probers...

mads-hartmann:
@ArthurSens Is this ready for review? I wasn't sure if the network problems were resolved ☺️

PR #10754, opened 2022-06-20
[usage] Make billing optional in the usage component

andrew-farries:
## Description
Make the payment processing part of the usage controller optional. If the secret containing Stripe API keys is configured then enable payments, otherwise disable it.

Usage metering and usage based billing are separate concepts that we will need to enable and disable independently of one another. In Gitpod SaaS we will want both, but for self-hosted installations, users will only care about usage metering.

At some point we will want to have the usage collection and billing integration in separate services but for now, while they both live in the usage component, let's at least be able to toggle billing integration on and off.

## Related Issue(s)
Part of #9036 

## How to test

* Trigger separate Werft builds, one setting `with-payment=true` and the other without.
* Observe that when payment is enabled, the usage component runs with billing integration enabled:
  ```
  {"level":"info","message":"payment integration with Stripe is set to true","serviceContext":{"service":"usage","version":"commit-b5ca1b4eebb51424345134f6847f9d180bd7c44f"},"severity":"INFO","time":"2022-06-20T05:56:38Z"}
  ```
  and when payment is not enabled, observe that it runs with billing integration disabled:
  ```
  {"level":"info","message":"payment integration with Stripe is set to false","serviceContext":{"service":"usage","version":"commit-b5ca1b4eebb51424345134f6847f9d180bd7c44f"},"severity":"INFO","time":"2022-06-20T05:56:38Z"}
  ```

## Release Notes

```release-note
NONE
```

## Documentation

None

### DIFF
--- a/components/usage/cmd/run.go
+++ b/components/usage/cmd/run.go
@@ -44,12 +44,17 @@ func run() *cobra.Command {
 				log.WithError(err).Fatal("Failed to establish database connection.")
 			}
 
-			err = stripe.Authenticate(apiKeyFile)
-			if err != nil {
-				log.WithError(err).Fatal("Failed to initialize stripe client.")
+			var billingController controller.BillingController = &controller.NoOpBillingController{}
+
+			if apiKeyFile != "" {
+				err = stripe.Authenticate(apiKeyFile)
+				if err != nil {
+					log.WithError(err).Fatal("Failed to initialize stripe client.")
+				}
+				billingController = &controller.StripeBillingController{}
 			}
 
-			ctrl, err := controller.New(schedule, controller.NewUsageReconciler(conn))
+			ctrl, err := controller.New(schedule, controller.NewUsageReconciler(conn, billingController))
 			if err != nil {
 				log.WithError(err).Fatal("Failed to initialize usage controller.")
 			}
@@ -74,7 +79,7 @@ func run() *cobra.Command {
 
 	cmd.Flags().BoolVar(&verbose, "verbose", false, "Toggle verbose logging (debug level)")
 	cmd.Flags().DurationVar(&schedule, "schedule", 1*time.Hour, "The schedule on which the reconciler should run")
-	cmd.Flags().StringVar(&apiKeyFile, "api-key-file", "/stripe-secret/apikeys", "Location of the stripe credentials file on disk")
+	cmd.Flags().StringVar(&apiKeyFile, "stripe-secret-path", "", "Location of the Stripe credentials file on disk")
 
 	return cmd
 }
--- a/components/usage/pkg/controller/billing.go
+++ b/components/usage/pkg/controller/billing.go
@@ -1,0 +1,26 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package controller
+
+import "github.com/gitpod-io/gitpod/usage/pkg/stripe"
+
+type BillingController interface {
+	Reconcile(report []TeamUsage)
+}
+
+type NoOpBillingController struct{}
+type StripeBillingController struct{}
+
+func (b *NoOpBillingController) Reconcile(report []TeamUsage) {}
+
+func (b *StripeBillingController) Reconcile(report []TeamUsage) {
+	// Convert the usage report to sum all entries for the same team.
+	var summedReport = make(map[string]int64)
+	for _, usageEntry := range report {
+		summedReport[usageEntry.TeamID] += usageEntry.WorkspaceSeconds
+	}
+
+	stripe.UpdateUsage(summedReport)
+}
--- a/components/usage/pkg/controller/reconciler.go
+++ b/components/usage/pkg/controller/reconciler.go
@@ -8,15 +8,15 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/gitpod-io/gitpod/common-go/log"
-	"github.com/gitpod-io/gitpod/usage/pkg/db"
-	"github.com/gitpod-io/gitpod/usage/pkg/stripe"
-	"github.com/google/uuid"
-	"gorm.io/gorm"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"time"
+
+	"github.com/gitpod-io/gitpod/common-go/log"
+	"github.com/gitpod-io/gitpod/usage/pkg/db"
+	"github.com/google/uuid"
+	"gorm.io/gorm"
 )
 
 type Reconciler interface {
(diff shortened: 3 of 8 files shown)


components/usage/cmd/run.go
```diff
@@ -44,12 +45,15 @@ func run() *cobra.Command {
 				log.WithError(err).Fatal("Failed to establish database connection.")
 			}
 
-			err = stripe.Authenticate(apiKeyFile)
-			if err != nil {
-				log.WithError(err).Fatal("Failed to initialize stripe client.")
+			log.Infof("payment integration with Stripe is set to %v", paymentEnabled)
+			if paymentEnabled {
```
easyCZ, line 49: PR title says that payment is enabled when the secret is set. But here, we also set an extra flag?

components/usage/cmd/run.go
```diff
@@ -44,12 +45,15 @@ func run() *cobra.Command {
 				log.WithError(err).Fatal("Failed to establish database connection.")
 			}
 
-			err = stripe.Authenticate(apiKeyFile)
-			if err != nil {
-				log.WithError(err).Fatal("Failed to initialize stripe client.")
+			log.Infof("payment integration with Stripe is set to %v", paymentEnabled)
```
easyCZ, line 48: Logs should always start with a capital letter - it's a sentence intended for humans. Errors on the other hand should always start with a lowercase because they are composed together.

components/usage/cmd/run.go
```diff
@@ -74,7 +78,8 @@ func run() *cobra.Command {
 
 	cmd.Flags().BoolVar(&verbose, "verbose", false, "Toggle verbose logging (debug level)")
 	cmd.Flags().DurationVar(&schedule, "schedule", 1*time.Hour, "The schedule on which the reconciler should run")
-	cmd.Flags().StringVar(&apiKeyFile, "api-key-file", "/stripe-secret/apikeys", "Location of the stripe credentials file on disk")
+	cmd.Flags().BoolVar(&paymentEnabled, "enable-payment", false, "Whether or not payment integration with Stripe should be enabled")
+	cmd.Flags().StringVar(&apiKeyFile, "api-key-file", "/stripe-secret/apikeys", "Location of the Stripe credentials file on disk")
```
easyCZ, line 82:
```suggestion
	cmd.Flags().StringVar(&apiKeyFile, "stripe-secret-path", "/stripe-secret/apikeys", "Location of the Stripe credentials file on disk")
```

easyCZ: There may be other API keys in the future.

components/usage/cmd/run.go
```diff
@@ -44,12 +45,15 @@ func run() *cobra.Command {
 				log.WithError(err).Fatal("Failed to establish database connection.")
 			}
 
-			err = stripe.Authenticate(apiKeyFile)
-			if err != nil {
-				log.WithError(err).Fatal("Failed to initialize stripe client.")
+			log.Infof("payment integration with Stripe is set to %v", paymentEnabled)
+			if paymentEnabled {
+				err = stripe.Authenticate(apiKeyFile)
```
easyCZ, line 50: Not strictly part of this PR but is there a way to not have the Authentication be global? This makes testing harder. 

I'd prefer if we constructed a new client with a set of credentials, and then injected that client to other components that need it.

components/usage/pkg/controller/reconciler.go
```diff
@@ -30,12 +30,13 @@ func (f ReconcilerFunc) Reconcile() error {
...
-	nowFunc func() time.Time
-	conn    *gorm.DB
+	nowFunc        func() time.Time
+	paymentEnabled bool
+	conn           *gorm.DB
 }
 
-func NewUsageReconciler(conn *gorm.DB) *UsageReconciler {
-	return &UsageReconciler{conn: conn, nowFunc: time.Now}
+func NewUsageReconciler(conn *gorm.DB, paymentEnabled bool) *UsageReconciler {
```
easyCZ, line 38: How about injecting a BillingController object, which if payment is disabled has a no-op implementation? That way, we don't need to push the decision logic on what to run with into the controller and can leave it at the instantiation level.

components/usage/cmd/run.go
andrew-farries: Yes, that's possible with the Go Stripe client. I'll address this in a follow up PR.

components/usage/cmd/run.go
andrew-farries: Which other keys do you forsee we'll need here? Can we make this change when if and when we need to?

components/usage/cmd/run.go
andrew-farries: Rebased to use the existing `--api-keys-file` flag rather than introducing another one.

components/usage/cmd/run.go
easyCZ: It's more the point of being more specific, when we can. `api-key-file` doesn't tell me which API it is for, while `stripe-something-something` does

components/usage/pkg/controller/billing.go
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package controller
+
+import "github.com/gitpod-io/gitpod/usage/pkg/stripe"
+
+type BillingController interface {
+	SubmitUsageToStripe(report []TeamUsage)
```
easyCZ, line 10: For it to remain a generic interface that makes sense, it shouldn't mention a concrete implementation. Here, it should probs be 
```suggestion
	SubmitReport(report []TeamUsage)
```

Or even just `Reconcile()`

andrew-farries: [5eafb9b](https://github.com/gitpod-io/gitpod/pull/10754/commits/5eafb9bab1435d646c77a2ee57393d474f96c3d0)

install/installer/pkg/components/usage/objects_test.go
```diff
@@ -61,3 +61,34 @@ func renderContextWithUsageConfig(t *testing.T, usage *experimental.UsageConfig)
 func renderContextWithUsageEnabled(t *testing.T) *common.RenderContext {
 	return renderContextWithUsageConfig(t, &experimental.UsageConfig{Enabled: true})
 }
+
+func renderContextWithStripeSecretSet(t *testing.T) *common.RenderContext {
+	ctx, err := common.NewRenderContext(config.Config{
```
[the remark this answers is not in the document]
andrew-farries: [4c0bcf1](https://github.com/gitpod-io/gitpod/pull/10754/commits/4c0bcf179c7214f77cb643c898c235fc99c13adc)

andrew-farries:
I'm not sure why Werft is failing here 😕 .

There are no obvious error message in the logs. The last line of the output is:

```
[components/usage:app] === RUN   TestListWorkspaceInstancesInRange_InBatches
```

easyCZ:
> I'm not sure why Werft is failing here 😕 .
> 
> There are no obvious error message in the logs. The last line of the output is:
> 
> ```
> [components/usage:app] === RUN   TestListWorkspaceInstancesInRange_InBatches
> ```

That test may output a rather long log line. It's possible it doesn't sit well with werft. Main (which contains the same test) is happy however.

install/installer/pkg/components/usage/objects_test.go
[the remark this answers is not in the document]
easyCZ: Thanks for the fix!

andrew-farries:
This looks to be more widespread than just this branch.

 https://werft.gitpod-dev.com/job/gitpod-build-af-add-timeout-show-command.8/raw now also failing at the same point.

easyCZ:
Fix for the debug logging in https://github.com/gitpod-io/gitpod/pull/10815

easyCZ:
@gitpod-io/engineering-self-hosted Would you be able to please help review this?

corneliusludmann (approved):
Went over the code and it looks good so far. Since we do not own code that is touched in this PR, let me know if you need me to review something in particular.

PR #10758, opened 2022-06-20
[usage] Batch lookup Workspaces to fix too many placeholders error

### LINKED ISSUE #10642: [Usage] workspace.go:68 Error 1390: Prepared statement contains too many placeholders
### Bug description

It looks like our `usage` component is logging an error containing tons of workspace IDs once every minute:

```
github.com/gitpod-io/gitpod/usage/pkg/db/workspace.go:68 Error 1390: Prepared statement contains too many placeholders\n[140.200ms] [rows:0] SELECT * FROM `d_b_workspace` WHERE `d_b_workspace`.`id` IN ('techlabapps-myjourney-mysicp2zuyi','antdesign-antdesign-j3ad1xm6oqt',.........
```

I believe the problem arises here:

https://github.com/gitpod-io/gitpod/blob/39557c349987d26bffc60529a2804197313243e8/components/usage/pkg/controller/reconciler.go#L204

when we try to get all the workspaces corresponding to all the workspace instances which ran some time this month.

Reminder: Gitpod runs _tons_ of workspaces and instances each month. We should be careful when querying them all into memory and/or dumping them all into logs. (And, currently, the usage component does both every minute.)

See also:
- [slack thread](https://gitpod.slack.com/archives/C02EN94AEPL/p1655158943465569) (internal)
- [cloud logging](https://cloudlogging.app.goo.gl/5wwBKCobgoEMqqDp6) (internal)

### Steps to reproduce

1. Check `usage` component logs in production

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

easyCZ:
## Description

Ensures that we do not include more than 2^16 placeholders in a `WHERE field IN (values)` prepared query.
https://bugs.mysql.com/bug.php?id=101630

## Related Issue(s)

Fixes #10642

## How to test

Unit tests

## Release Notes

```release-note
NONE
```

## Documentation

NONE

### DIFF
--- a/components/usage/pkg/db/workspace.go
+++ b/components/usage/pkg/db/workspace.go
@@ -11,6 +11,7 @@ import (
 	"github.com/google/uuid"
 	"gorm.io/datatypes"
 	"gorm.io/gorm"
+	"math"
 	"time"
 )
 
@@ -59,15 +60,33 @@ const (
 	WorkspaceType_Regular  WorkspaceType = "regular"
 )
 
+const maxListBatchSize = 65535 // 2^16 - 1
+
 func ListWorkspacesByID(ctx context.Context, conn *gorm.DB, ids []string) ([]Workspace, error) {
 	if len(ids) == 0 {
 		return nil, nil
 	}
 
 	var workspaces []Workspace
-	tx := conn.WithContext(ctx).Where(ids).Find(&workspaces)
-	if tx.Error != nil {
-		return nil, fmt.Errorf("failed to list workspaces by id: %w", tx.Error)
+
+	items := len(ids)
+	batches := int(math.Ceil(float64(items) / maxListBatchSize))
+	for i := 0; i < batches; i++ {
+		lower := i * maxListBatchSize
+		upper := (i + 1) * maxListBatchSize
+
+		if upper > items {
+			upper = items
+		}
+
+		batchIDs := ids[lower:upper]
+		var results []Workspace
+		tx := conn.WithContext(ctx).Where(batchIDs).Find(&results)
+		if tx.Error != nil {
+			return nil, fmt.Errorf("failed to list workspaces by id: %w", tx.Error)
+		}
+
+		workspaces = append(workspaces, results...)
 	}
 
 	return workspaces, nil
--- a/components/usage/go.mod
+++ b/components/usage/go.mod
@@ -63,6 +63,7 @@ require (
 	github.com/sirupsen/logrus v1.8.1
 	github.com/spf13/cobra v1.4.0
 	github.com/stretchr/testify v1.7.0
+	github.com/stripe/stripe-go/v72 v72.114.0
 	gorm.io/datatypes v1.0.6
 	gorm.io/driver/mysql v1.3.3
 	gorm.io/gorm v1.23.5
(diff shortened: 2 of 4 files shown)


jankeromnes:
> Ensures that we do not include more than 2^16 placeholders in a `WHERE field IN (values)` prepared query.
> https://bugs.mysql.com/bug.php?id=101630

🤯 is that limit really 2^16? I assumed it would be something like 1000. Anyway, I guess the unit tests will tell -- will take a look today. 👀

easyCZ:
> > Ensures that we do not include more than 2^16 placeholders in a `WHERE field IN (values)` prepared query.
> > [bugs.mysql.com/bug.php?id=101630](https://bugs.mysql.com/bug.php?id=101630)
> 
> 🤯 is that limit really 2^16? I assumed it would be something like 1000. Anyway, I guess the unit tests will tell -- will take a look today. 👀

Yeah I was also surprised.

components/usage/pkg/db/workspace.go
```diff
@@ -59,15 +60,33 @@ const (
...
 	}
 
 	var workspaces []Workspace
-	tx := conn.WithContext(ctx).Where(ids).Find(&workspaces)
-	if tx.Error != nil {
-		return nil, fmt.Errorf("failed to list workspaces by id: %w", tx.Error)
+
+	items := len(ids)
+	batches := int(math.Ceil(float64(items / listBatchSize)))
+	for i := 0; i <= batches; i++ {
```
jankeromnes, line 74: Won't this always process one more batch than necessary?

(`math.Ceil` already rounds up the number of necessary batches, then you go from `0` to `batches` included?)

```suggestion
	batches := int(math.Ceil(float64(items / listBatchSize)))
	for i := 0; i < batches; i++ {
```

jankeromnes (approved):
Code looks good and seems to work (according to the unit test).

I find it hard to trust code just based on unit tests (e.g. production workload might be significantly different than what the unit tests do, as we've seen for this bug), but in this case, it's okay to test in production since this component is not on the critical path yet. 😊

Adding a hold because of optional questions/remarks:

/hold

components/usage/pkg/db/workspace.go
```diff
@@ -59,15 +60,33 @@ const (
...
+		lower := i * listBatchSize
+		upper := (i + 1) * listBatchSize
+
+		if upper > items {
+			upper = items
+		}
```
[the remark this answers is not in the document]
easyCZ: In go, the upper bound from a slice is not safe for out of bounds, it panics https://goplay.tools/snippet/SgSiwcAPfXC

components/usage/pkg/db/workspace.go
easyCZ: I've simplified this a bit. The issue was with `float64(items / listBatchSize)` which actually performed integer division (div). What it needs to be is `float64(items) / listBatchSize` because the numerator determines the resulting type of the division operation, here it's floating point division. Which then works with Ceil to always run at least one batch.

easyCZ:
> I find it hard to trust code just based on unit tests (e.g. production workload might be significantly different than what the unit tests do, as we've seen for this bug)

How do you propose we solve this problem? In general there are a couple of constraints:
1. There's only one production, staging helps (but questionably it's planned to be removed) but still doesn't come close to production
2. Engineers shouldn't really have direct DB access to production. This should be used in emergency situations but not as part of a development cycle, it is a main source of compromised systems - attack on an employee.
3. If CD exists, and so does a fast rollback, then it should be encouraged to "test in production" - given there are sufficient safe-guards to prevent the system going crazy - feature flags or read only mode are good examples

PR #10775, opened 2022-06-20
Add tests for external dependencies in GCP

nandajavarma:
## Description

This PR adds the external dependencies test for GCP cloud platform. These dependencies include `cloudsql` as external database, `GCR` as external registry, and `cloud storage` as external storage. These dependencies are randomly selected for each test providing us a overall testing of every combination over a course of days. 

## Related Issue(s)

Fixes #

## How to test

You can run the following command to test one of the available combinations:
```
werft run github -j .werft/gke-installer-tests.yaml
```

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/.werft/installer-tests.ts
+++ b/.werft/installer-tests.ts
@@ -209,7 +209,7 @@ function callMakeTargets(phase: string, description: string, makeTarget: string)
 function randomize(resource: string, platform: string): string {
     // in the follow-up PR we will add `${platform}-${resource}` as an option here to
     // test against resource dependencies(storage, db, registry) for each cloud platform
-    const options = ["incluster"];
+    const options = [`${platform}-${resource}`, "incluster"];
     return options[Math.floor(Math.random() * options.length)];
 }
 
--- a/install/infra/terraform/gke/main.tf
+++ b/install/infra/terraform/gke/main.tf
@@ -152,6 +152,29 @@ resource "google_container_node_pool" "workspaces" {
   }
 }
 
+resource "google_sql_database_instance" "gitpod" {
+  name = "sql-${var.name}"
+  database_version = "MYSQL_5_7"
+  region = "${var.region}"
+  settings {
+      tier = "db-n1-standard-2"
+  }
+  deletion_protection = false
+}
+
+resource "google_sql_database" "database" {
+    name = "gitpod"
+    instance = "${google_sql_database_instance.gitpod.name}"
+    charset = "utf8"
+    collation = "utf8_general_ci"
+}
+
+resource "google_sql_user" "users" {
+    name = "gitpod"
+    instance = "${google_sql_database_instance.gitpod.name}"
+    password = "gitpod"
+}
+
 module "gke_auth" {
   depends_on = [google_container_node_pool.workspaces]
 
--- a/install/infra/terraform/k3s/main.tf
+++ b/install/infra/terraform/k3s/main.tf
@@ -171,6 +171,28 @@ resource "google_dns_record_set" "gitpod-dns-3" {
   rrdatas = [google_compute_instance.k3s_master_instance.network_interface[0].access_config[0].nat_ip]
 }
 
+resource "google_sql_database_instance" "gitpod" {
+  name = "sql-${var.name}"
+  database_version = "MYSQL_5_7"
+  region = "${var.gcp_region}"
+  settings {
+      tier = "db-n1-standard-2"
+  }
+  deletion_protection = false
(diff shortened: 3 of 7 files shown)


roboquat:

<details>

</details>

.werft/installer-tests.ts
```diff
@@ -41,14 +116,22 @@ const INFRA_PHASES: { [name: string]: InfraConfig } = {
         makeTarget: "managed-dns",
         description: "Sets up external-dns & cloudDNS config",
     },
+    GENERATE_KOTS_CONFIG: {
+        phase: "generate-kots-config",
+        makeTarget: `generate-kots-config storage=${randomize("storage", cloud)} registry=${randomize(
```
adrienthebo, line 121: Suggestion: performing randomization in this location means that loading this file isn't an idempotent operation. If a randomized test configuration fails we'll likely need to repeat that test case (either to rule out a transient failure or reproduce the failure). Because these settings are encoded during the file evaluation we don't have a means of injecting an exact configuration.

Phase parameterization can be kept out of scope for this PR (espcially because it'll take time to define how we configure phases), but we should definitely follow up on this in another issue/PR for Q3. If we defer to Q3 I'd love to capture this requirement in a ticket before we merge this PR.

**EDIT**: Leaving the above comment as it's still relevant, but less so.

In `install/tests/Makefile` we do provide mechanisms for setting the exact configuration so we should be able to reproduce a given test case if we omit werft. This reduces the challenge of randomizing these values 🎉  , but we'll still need to account for this in the future.

install/tests/Makefile
```diff
@@ -92,6 +124,10 @@ check-gitpod-installation: delete-cm-setup check-kots-app check-env-sub-domain
 run-tests:
 	./tests.sh ${KUBECONFIG}
 
+kots-uprgade:
```
adrienthebo, line 127:
```suggestion
kots-upgrade:
```

(Tiny nitpick, I think this may have already been addressed in a previous PR)

install/tests/Makefile
```diff
@@ -101,11 +137,8 @@ cleanup:
 
 get-results:
 	@echo "If you have gotten this far, it means your setup succeeded"
-	@echo
-	@echo "URL of your setup is: "https://$(TF_VAR_TEST_ID).gitpod-self-hosted.com""
-	@echo
+	@echo "The IP address of you setup is "https://$(TF_VAR_TEST_ID).gitpod-self-hosted.com""
```
adrienthebo, line 140: Note: will this subdomain need to change given the issues we've had with domain level wildcard letsencrypt issuance?

install/infra/terraform/gke/main.tf
```diff
@@ -152,6 +152,29 @@ resource "google_container_node_pool" "workspaces" {
   }
 }
 
+resource "google_sql_database_instance" "gitpod" {
+  name = "sql-${var.name}"
+  database_version = "MYSQL_5_7"
+  region = "${var.region}"
+  settings {
+      tier = "db-n1-standard-2"
```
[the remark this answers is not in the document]
nandajavarma: This is a good point, ideally we can make this configurable and use a cheaper one for tests. I was basically just translating the [reference architecture](https://www.gitpod.io/docs/self-hosted/latest/reference-architecture/production-ready-single-cluster) here. I was thinking that since our test setup is ephemeral the extra cost we pay here is quite small. May be I am wrong, and it adds up. I will make a note for this for a follow-up

install/infra/terraform/k3s/main.tf
```diff
@@ -171,6 +171,28 @@ resource "google_dns_record_set" "gitpod-dns-3" {
   rrdatas = [google_compute_instance.k3s_master_instance.network_interface[0].access_config[0].nat_ip]
 }
 
+resource "google_sql_database_instance" "gitpod" {
```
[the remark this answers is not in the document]
nandajavarma: You are right! I just wanted to avoid calling yet another module from the `main.tf` :see_no_evil:

install/tests/Makefile
nandajavarma: We will highly likely stick with the same subdomain, just we will try and re-use the certs issued across multiple sub-domains(max 200 subdomains can be added under a certificate, iirc). In reality, I don't think we will hit a limit even if we create new certs per domain. This week it so happened because I was running too many tests on my own

adrienthebo (approved):
LGTM!

PR #10782, opened 2022-06-21
Add `gp timeout show` command

andrew-farries:
## Description
Add a `gp timeout show` command to the `gp` CLI to go along with the recently added `gp timeout extend` command (see https://github.com/gitpod-io/gitpod/pull/10619).

- [x] ~~Need to hold until PR https://github.com/gitpod-io/gitpod/pull/10896 deployed~~
- [x] Maybe want to follow up @filiptronicek 's suggestion https://github.com/gitpod-io/gitpod/pull/10782#issuecomment-1162703143 

## How to test

1. Create a workspace in preview environment
2. Run `gp timeout show`
3. Output should be (value depends on user's plan):
  ```
  60m
  ```
4. Extend the workspace timeout with `gp timeout extend`. 
5. Rerun `gp timeout show`. Output should be:
  ```
  180m
  ```
6. Open another workspace, and extend its timeout `gp timeout extend`
7. Back to first one, `gp timeout show` should back to original one

## Release Notes

```release-note
Add command `gp timeout show` to show the timeout of current workspace
```

## Documentation

We'll need to add something to https://github.com/gitpod-io/website/pull/2218 either before or after it's merged.

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-cli/cmd/timeout-show.go
+++ b/components/gitpod-cli/cmd/timeout-show.go
@@ -1,0 +1,52 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	gitpod "github.com/gitpod-io/gitpod/gitpod-cli/pkg/gitpod"
+	"github.com/spf13/cobra"
+)
+
+// showTimeoutCommand shows the workspace timeout
+var showTimeoutCommand = &cobra.Command{
+	Use:   "show",
+	Short: "Show the current workspace timeout",
+	Run: func(_ *cobra.Command, _ []string) {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		wsInfo, err := gitpod.GetWSInfo(ctx)
+		if err != nil {
+			fail(err.Error())
+		}
+		client, err := gitpod.ConnectToServer(ctx, wsInfo, []string{
+			"function:getWorkspaceTimeout",
+			"resource:workspace::" + wsInfo.WorkspaceId + "::get/update",
+		})
+		if err != nil {
+			fail(err.Error())
+		}
+
+		res, err := client.GetWorkspaceTimeout(ctx, wsInfo.WorkspaceId)
+		if err != nil {
+			fail(err.Error())
+		}
+
+		// Try to use `DurationRaw` but fall back to `Duration` in case of
+		// old server component versions that don't expose it.
+		if res.DurationRaw != "" {
+			fmt.Println(res.DurationRaw)
+		} else {
+			fmt.Println(res.Duration)
+		}
+	},
+}
+
+func init() {
+	timeoutCmd.AddCommand(showTimeoutCommand)
+}
--- a/components/gitpod-protocol/go/gitpod-service.go
+++ b/components/gitpod-protocol/go/gitpod-service.go
@@ -1707,8 +1707,9 @@ type StartWorkspaceOptions struct {
 
 // GetWorkspaceTimeoutResult is the GetWorkspaceTimeoutResult message type
 type GetWorkspaceTimeoutResult struct {
-	CanChange bool   `json:"canChange,omitempty"`
-	Duration  string `json:"duration,omitempty"`
+	CanChange   bool   `json:"canChange,omitempty"`
+	DurationRaw string `json:"durationRaw,omitempty"`
+	Duration    string `json:"duration,omitempty"`
 }
 
 // WorkspaceInstancePort is the WorkspaceInstancePort message type


components/gitpod-cli/cmd/timeout-show.go
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) 2020 Gitpod GmbH. All rights reserved.
```
akosyakov, line 1: 2022

akosyakov:
@andrew-farries Don't forget about docs update. At least file an issue.

andrew-farries:
@mustard-mh Do you want to add the docs for this to your existing docs PR (https://github.com/gitpod-io/website/pull/2218) or would you rather I create a new one?

mustard-mh:
> @mustard-mh Should we add the docs for this to your existing docs PR (https://github.com/gitpod-io/website/pull/2218) or would you rather I create a new one?

@andrew-farries  You can do it after previous one https://github.com/gitpod-io/website/pull/2218 merged

mustard-mh (commented):
I tested with it. As we can see in image below, 

default value is `long`
after extend `180m`
extend another workspace, this workspace will reset to default `long`

If we can't format every output time to `180m`, maybe we can turn `180m` to `extended`? 

PS. As a user I prefer `180m` formated duration🙂

https://github.com/gitpod-io/gitpod/blob/7bf63df2e9f39bf80fd83e1ae0d1a864a2e69c9a/components/gitpod-protocol/src/gitpod-service.ts#L340-L349

image

mustard-mh:
https://github.com/gitpod-io/gitpod/pull/10782#pullrequestreview-1013290002 What do you think/prefer? @loujaybee

mustard-mh (commented):
If we want to turn all output into `180m` formatted, maybe we need a map/func in protocol, since we have multiple similar code like this below
https://github.com/gitpod-io/gitpod/blob/402894b63b604234899db345fca0a9ce3750cec1/components/server/src/user/user-service.ts#L173-L183
https://github.com/gitpod-io/gitpod/blob/402894b63b604234899db345fca0a9ce3750cec1/components/server/src/user/user-service.ts#L185-L196

mustard-mh:
Notice that if we need changes from server, this PR will be approved before next WebApp deploy, or it will block IDE deployment 🙏

mustard-mh:
> Thanks a lot @andrew-farries for the new command 🧡! I tested it and ti works as advertised.
> 
> Maybe we could add to the minutes just a "status" which would indicate whether the timeout is the default one or the extended one? Something like
> 
> ```
> $ gp timeout show
> 60m (Default)
> $ gp timeout extend
> $ gp timeout show
> 180m (Extended)
> ```

@filiptronicek  Remember that `60m` is not default for all users, value of default is depended on Plan of user/team. cc @andrew-farries 

|Image|
|-|
|image|

andrew-farries:
> Remember that 60m is not default for all users, value of default is depend on Plan of user/team.

I suggest we leave the output as is for now, ie without `(Default)` or `(Extended)`.

filiptronicek:
> Remember that 60m is not default for all users, value of default is depend on Plan of user/team.

@mustard-mh I agree, thanks for bringing this up. I wanted to say, that as a user, I would personally like to check `gp timeout show` just to see how many minutes my timeout is and whether I have successfully for example extended it for this repo. For checking the subscription, there could be something like `gp billing` (not sure if suggested before).

> I suggest we leave the output as is for now, ie without (Default) or (Extended).

@andrew-farries I see where you're coming from, it could perhaps be in some way programmatically consumed afterwards as well, but I would prefer to see a bit more context about my timeout, if we have it available. We could also ship it as is and afterwards provide a flag or similar to provide such information.

mustard-mh:
If a user cannot extend timeout, output will be like this @filiptronicek 

Only **Unleashed** plan can extend timeout 🙈

|Image|
|-|
|image|

mustard-mh:
Could you rebase and squash first and third commit since they are not worth split. cc @andrew-farries 

image

filiptronicek:
> If a user cannot extend timeout, output will be like this @filiptronicek

@mustard-mh yes, I meant like if you have a lot of workspaces open and want to check a certain one to see if 15 minutes ago you correctly extended its timeout, you can check if it's extended. I just feel like having that bit of extra info would be nice from this command - some new users may not know if 180 minutes is extended or standard.

mustard-mh (approved):
Tested and code LGTM

/hold

- [ ] ‼️ Need to deploy WebApp first, so this PR should unhold before next WebApp deploy
- [ ] Maybe want to follow up @filiptronicek 's suggestion https://github.com/gitpod-io/gitpod/pull/10782#issuecomment-1162703143 

I'll add them to PR description too

mustard-mh:
> > @mustard-mh Should we add the docs for this to your existing docs PR ([gitpod-io/website#2218](https://github.com/gitpod-io/website/pull/2218)) or would you rather I create a new one?
> 
> @andrew-farries You can do it after previous one [gitpod-io/website#2218](https://github.com/gitpod-io/website/pull/2218) merged, I can ping you after that

@andrew-farries  Previous doc PR has been merged, you can start it or ask @filiptronicek to help with it 👍

andrew-farries:
Docs PR is here: https://github.com/gitpod-io/website/pull/2292

mustard-mh:
Thank you @andrew-farries , could you split this PR into two PRs, to unblock deployment? `server` will approve first, and then `gp-cli`. 

See more context in [internal chat](https://gitpod.slack.com/archives/C01KGM9BH54/p1656053577087659?thread_ts=1656048950.670019&cid=C01KGM9BH54)

andrew-farries:
>  could you split this PR into two PRs, to unblock deployment? `server` will approve first, and then `gp-cli`.

I've done this. This PR now contains just the `gp` change and https://github.com/gitpod-io/gitpod/pull/10896 has the `server` change.

geropl (approved):
Thx, works as expected! :+1: 

image
nit: I think the `long` in the beginning might be a hint towards a conversion bug in `server`. But as it's a display issue only: :+1:

jeanp413:
Some feedback, not really important, but I would expect some more user friendly output other than just a number, same with `extend` a message telling me what was the side effect of running this command

mustard-mh (approved):
Thank you, LGTM

mustard-mh:
/unhold

> Some feedback, not really important, but I would expect some more user friendly output other than just a number, same with extend a message telling me what was the side effect of running this command

> I see where you're coming from, it could perhaps be in some way programmatically consumed afterwards as well, but I would prefer to see a bit more context about my timeout, if we have it available. We could also ship it as is and afterwards provide a flag or similar to provide such information.

You can do it in another PR if you want, to print more friendly 🙂 cc @jeanp413 @filiptronicek 

i.e. `extend` success don't print anything🙈, we can show current ws timeout

Issue #10757, opened 2022-06-20
[Usage-based] Implement a "Cost Center" entity


jankeromnes:
## Is your feature request related to a problem? Please describe

Until now, we attribute workspace usage hours to Users or Teams directly (which themselves might map to a Stripe Customer).

But, as explained in [this comment](https://github.com/gitpod-io/gitpod/issues/10533#issuecomment-1160159281), it would be good to create an "indirection", and attribute workspace usage to a dedicated "Cost Center" entity.

## Describe the behaviour you'd like

- A new "Cost Center" entity should be created (which can potentially map to a Stripe Customer if payments are enabled)
- All workspace usage hours should get pre-attributed to a "Cost Center" instead of a Team

## Describe alternatives you've considered

We currently attribute directly to a Team or a User. However, this makes some of the logic implicit (i.e. "if no attributed Team, default to workspace owner"), and does not allow future evolutions like sharing a Cost Center between multiple Teams, or having "Organization"-level Cost Centers (once "Organizations" exist)

geropl:
Hey @jankeromnes , thank you for creating an issues for this!

I feel we still need a bit of alignment before we dive into the implementation. I'm not sure about the naming, for instance. "Cost center" feels to me like a different meaning, and we came to use it because we put "Kostenstelle" into Google translate. :laughing: 
E.g., I like "billing account", because it's less ambiguous.
But anyway, we should align on this one. :slightly_smiling_face:

geropl:
> What I don't like about "billing account" is that this directly maps to payments -- e.g. I would expect a "billing account" to be something in Stripe. In Self-Hosted, you may want to track costs without any billing (e.g. just for accounting, or setting up per-team/per-project quotas, etc).

:bulb: Ah, that's a new perspective for me! I always thought about the concept to strictly be a "BillingService" (SaaS) concern. Have to think about the idea that it can also host quotas...

jankeromnes:
**Update**: As an initial simplification, we've started by implementing a unique `usageAttributionId` (new column on `d_b_workspace_instance`):

#10822 

This sort of represents a "cost center" for workspace usage (but without an entity), and can either point to a Team or to a User, e.g.:
- A `team:1234` attribution ID identifies the Team with ID `1234`
- A `user:5678` attribution ID identifies the User with ID `5678`

[...]

PR #10788, opened 2022-06-21
[db][protocol] Implement a CostCenter entity to attribute workspace usage to

jankeromnes:
## Description

Implement a CostCenter entity to attribute workspace usage to.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/10757

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-db/src/typeorm/entity/db-cost-center.ts
+++ b/components/gitpod-db/src/typeorm/entity/db-cost-center.ts
@@ -1,0 +1,22 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the Gitpod Enterprise Source Code License,
+ * See License.enterprise.txt in the project root folder.
+ */
+
+import { Entity, Column, PrimaryColumn } from "typeorm";
+import { CostCenter } from "@gitpod/gitpod-protocol";
+
+@Entity()
+// on DB but not Typeorm: @Index("ind_lastModified", ["_lastModified"])   // DBSync
+export class DBCostCenter implements CostCenter {
+    @PrimaryColumn()
+    id: string;
+
+    @Column()
+    spendingLimit: number;
+
+    // This column triggers the db-sync deletion mechanism. It's not intended for public consumption.
+    @Column()
+    deleted: boolean;
+}
--- a/components/gitpod-db/src/container-module.ts
+++ b/components/gitpod-db/src/container-module.ts
@@ -67,6 +67,8 @@ import { TypeORMBlockedRepositoryDBImpl } from "./typeorm/blocked-repository-db-
 import { BlockedRepositoryDB } from "./blocked-repository-db";
 import { WebhookEventDB } from "./webhook-event-db";
 import { WebhookEventDBImpl } from "./typeorm/webhook-event-db-impl";
+import { CostCenterDB } from "./cost-center-db";
+import { CostCenterDBImpl } from "./typeorm/cost-center-db-impl";
 
 // THE DB container module that contains all DB implementations
 export const dbContainerModule = new ContainerModule((bind, unbind, isBound, rebind) => {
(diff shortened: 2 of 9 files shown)


jankeromnes:
@geropl Apologies for rushing this 🙈 I actually implemented all this before I saw your request for further alignment, which I agree with: https://github.com/gitpod-io/gitpod/issues/10757#issuecomment-1161354950

So, I'm happy to further align on this concept and on naming, then reflect any changes back in this PR. 👍

components/gitpod-db/src/typeorm/migration/1655795038249-CostCenter.ts
```diff
@@ -0,0 +1,29 @@
...
+export class CostCenter1655795038249 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(
+            "CREATE TABLE IF NOT EXISTS `d_b_cost_center` (`id` char(36) NOT NULL, `deleted` tinyint(4) NOT NULL DEFAULT '0', `_lastModified` timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6), PRIMARY KEY (`id`), KEY `ind_dbsync` (`_lastModified`)) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;",
+        );
+        if (!(await columnExists(queryRunner, "d_b_workspace_instance", "costCenterId"))) {
+            await queryRunner.query(
+                "ALTER TABLE `d_b_workspace_instance` CHANGE `attributedTeamId` `costCenterId` char(36) NOT NULL DEFAULT '', ALGORITHM=INPLACE, LOCK=NONE",
+            );
+        }
```
jankeromnes, line 19: Obsolete and no longer needed:

```suggestion
```

components/gitpod-db/src/typeorm/migration/1655795038249-CostCenter.ts
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        if (!(await columnExists(queryRunner, "d_b_workspace_instance", "attributedTeamId"))) {
+            await queryRunner.query(
+                "ALTER TABLE `d_b_workspace_instance` CHANGE `costCenterId` `attributedTeamId` char(36) NOT NULL DEFAULT '', ALGORITHM=INPLACE, LOCK=NONE",
+            );
+        }
```
jankeromnes, line 27:
```suggestion
```

jankeromnes:
Whoops, looks like this suffered from bit-rot a bit.

If anyone would like to pick this up, please feel free to force-push to this branch, or salvage anything that looks useful from it (no need to attribute me 😊).

components/gitpod-db/src/typeorm/entity/db-cost-center.ts
```diff
@@ -0,0 +1,32 @@
...
+    @Column(TypeORM.UUID_COLUMN_TYPE)
+    @Index("ind_userId")
+    userId: string;
+
+    @Column({
+        ...TypeORM.UUID_COLUMN_TYPE,
+        default: "",
+        transformer: Transformer.MAP_EMPTY_STR_TO_UNDEFINED,
+    })
+    teamId?: string;
```
easyCZ, line 27: IMO, this mapping should be opaque. As in, the column should not state it maps to a Team, but stay as `costID`. For now, that can contain the current format of `team:<teamID>` or `user:<userID>` but it allows to keep the cost center idea a bit more general than just teams.

AlexTugarev: I've no problem with that, but we tend to use another pattern in other places and I'd prefer to stay consistent. E.g. Project?.teamId/Project?.userId

easyCZ: The point is that the value shouldn't necessarily represent just a teamID, it could represent some other structure, ie `userId` when the attribution is to a user.

AlexTugarev: Got that. If the `teamId` is null, the `userId` is the attribution target. 
If we plan to introduce a different type of attribution target, we'd add a new column for that type. This is the way we've extended entities so far.

geropl: @AlexTugarev We already use the [`AttributionId`](https://github.com/gitpod-io/gitpod/blob/c7cfb42d785897fec9a9c2e4d7514e6906f92802/components/gitpod-protocol/src/attribution.ts#L19) in various places and APIs now. I think it makes sense to re-use it here as well. :+1:

geropl (approved):
Concise and spot-on - nice! :orange_heart:

easyCZ:
How is no-spending limit represent here?
What will populate this table?

Issue #10762, opened 2022-06-20
Don't refresh tokens twice in parallel


AlexTugarev:
We see `Failed to refresh token!` errors in `server` logs for token refresh tries with GitLab.

This is caused by calling the API client "in parallel" e.g. to fetch context details. Each client call may result in a token refresh cycle, which is just not allowed for GitLab, thus a call may fail. 

Token refresh cycles should be managed internally in way that there are no two competing ones.

PR #10794, opened 2022-06-21
Fix concurrent token renewal issues

AlexTugarev:
## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/10762

## How to test
1. Use git with the gitlab.com integration for a while, no issues with tokens should occur.

## Release Notes

```release-note
Token renewals should not run concurrently.
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/server/src/user/token-service.ts
+++ b/components/server/src/user/token-service.ts
@@ -52,9 +52,9 @@ export class TokenService implements TokenProvider {
                 `No token found for user ${identity.authProviderId}/${identity.authId}/${identity.authName}!`,
             );
         }
-        const refreshTime = new Date();
-        refreshTime.setTime(refreshTime.getTime() + 30 * 60 * 1000);
-        if (token.expiryDate && token.expiryDate < refreshTime.toISOString()) {
+        const aboutToExpireTime = new Date();
+        aboutToExpireTime.setTime(aboutToExpireTime.getTime() + 5 * 60 * 1000);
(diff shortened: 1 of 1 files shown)


components/server/src/user/token-service.ts
```diff
@@ -52,9 +52,9 @@ export class TokenService implements TokenProvider {
                 `No token found for user ${identity.authProviderId}/${identity.authId}/${identity.authName}!`,
             );
         }
-        const refreshTime = new Date();
-        refreshTime.setTime(refreshTime.getTime() + 30 * 60 * 1000);
-        if (token.expiryDate && token.expiryDate < refreshTime.toISOString()) {
+        const aboutToExpireTime = new Date();
+        aboutToExpireTime.setTime(aboutToExpireTime.getTime() + 5 * 60 * 1000);
+        if (token.expiryDate && token.expiryDate < aboutToExpireTime.toISOString()) {
```
andrew-farries, line 57: Why was this renamed?

andrew-farries (approved):
How does this work between multiple `server` replicas? This promise cache is per-server - does that matter?

/hold

components/server/src/user/token-service.ts
AlexTugarev: Renaming was not necessary, but that reflect the purpose of this time, while `refreshTime` does not. 

Also note the change from 30 to 5 minutes before expiration.

PR #2529, opened 2020-12-14
[ws-daemon] don't fail if prebuilt is not found

AlexTugarev:
should fix #2525

PR #10795, opened 2022-06-21
remove core-dev preview environment options

liam-j-bennett:
## Description
Removes the core-dev preview environment options

## Related Issue(s)

Fixes [#2529](https://github.com/gitpod-io/ops/issues/2529)

## Release Notes
```
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/.werft/build.yaml
+++ b/.werft/build.yaml
@@ -23,9 +23,6 @@ pod:
   - name: gcp-sa-release
     secret:
       secretName: gcp-sa-gitpod-release-deployer
-  - name: gpsh-coredev-license
-    secret:
-      secretName: gpsh-coredev-license
   - name: prometheus-remote-write-auth
     secret:
       secretName: prometheus-remote-write-auth
@@ -94,9 +91,6 @@ pod:
       readOnly: true
     - name: gcp-sa-release
       mountPath: /mnt/secrets/gcp-sa-release
-      readOnly: true
-    - name: gpsh-coredev-license
-      mountPath: /mnt/secrets/gpsh-coredev
       readOnly: true
     - name: gpsh-harvester-license
       mountPath: /mnt/secrets/gpsh-harvester
--- a/.werft/jobs/build/deploy-to-preview-environment.ts
+++ b/.werft/jobs/build/deploy-to-preview-environment.ts
@@ -57,7 +57,6 @@ const vmSlices = {
 export async function deployToPreviewEnvironment(werft: Werft, jobConfig: JobConfig) {
     const {
         version,
-        withVM,
         analytics,
         cleanSlateDeployment,
         withPayment,
@@ -75,7 +74,7 @@ export async function deployToPreviewEnvironment(werft: Werft, jobConfig: JobCon
 
 
 
-    const domain = withVM ? `${destname}.preview.gitpod-dev.com` : `${destname}.staging.gitpod-dev.com`;
+    const domain = `${destname}.preview.gitpod-dev.com`;
     const monitoringDomain = `${destname}.preview.gitpod-dev.com`;
     const url = `https://${domain}`;
     const imagePullAuth = exec(`printf "%s" "_json_key:$(kubectl --kubeconfig ${CORE_DEV_KUBECONFIG_PATH} get secret ${IMAGE_PULL_SECRET_NAME} --namespace=keys -o yaml \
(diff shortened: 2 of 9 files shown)


roboquat:
@liam-j-bennett: Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://git.k8s.io/community/contributors/guide/release-notes.md) to remove it.

<details>

</details>

meysholdt:
hi @liam-j-bennett! 
It's great to see how much simpler our code becomes  - can you remove core-dev-related code from the cleanup job as well? see [here](https://github.com/gitpod-io/gitpod/blob/main/.werft/platform-delete-preview-environments-cron.ts#L155).

.werft/jobs/build/prepare.ts
```diff
@@ -62,33 +62,20 @@ function configureStaticClustersAccess() {
...
     werft.done(prepareSlices.BOOT_VM)
 }
 
-function shouldCreateVM(config: JobConfig) {
-    return config.withVM && (
-        !VM.vmExists({ name: config.previewEnvironment.destname }) ||
-        config.cleanSlateDeployment
-    )
-}
-
```
ArthurSens, line 91: `shouldCreateVM()` was doing more than just checking if the `withVM `annotation was set, notice that it also checks if the VM already exists. Was it intentional that we always try to create a VM? 

* If yes, I believe we should delete `decideHarvesterVMCreation` and just use the `createVM` method where the former was called.
* If not, then I believe this change was accidental 😅

ArthurSens (commented):
Nice clean-up 🧹!

Just like mo said, I believe we can clean up even more! I think we can also un-mount some secrets from `.werft/build.yaml` and also delete some terraform code that creates those secrets(terraform would be another PR in another repo of course 🙂)

.werft/jobs/build/deploy-to-preview-environment.ts
```diff
@@ -221,34 +216,14 @@ async function deployToDevWithInstaller(werft: Werft, jobConfig: JobConfig, depl
...
-            // re-create namespace
-            await cleanStateEnv(deploymentKubeconfig, metaEnv());
-
-        } else {
-            werft.log(installerSlices.CLEAN_ENV_STATE, "Clean the preview environment slate...");
-            createNamespace(namespace, deploymentKubeconfig, metaEnv({ slice: installerSlices.CLEAN_ENV_STATE }));
-        }
+
+        werft.log(installerSlices.CLEAN_ENV_STATE, "Clean the preview environment slate...");
+        createNamespace(namespace, deploymentKubeconfig, metaEnv({ slice: installerSlices.CLEAN_ENV_STATE }));
```
ArthurSens, line 221: Looks like a bug that already existed before your clean-up, but I'm not sure if we want to `always` create a new namespace in the "Clean slate deployment" phase.

That's even more relevant for harvester preview because we use the `default` namespace there, and there is absolutely no need to create such namespace 🤔 

Should we just delete this whole try-catch block?

.werft/jobs/build/prepare.ts
liam-j-bennett: Similarly to the namespace, this should be atomic, so this guard should be unnecessary. Doing another `apply` shouldn't cause any issues, but I can add the guards in again if we think it might cause issues

.werft/observability/monitoring-satellite.ts
```diff
@@ -68,7 +66,7 @@ export class MonitoringSatelliteInstaller {
                 domain: '${previewDomain}',
                 nodeExporterPort: ${nodeExporterPort},
             },
-            ${withVM ? "" : "nodeAffinity: { nodeSelector: { 'gitpod.io/workload_services': 'true' }, },"}
+            ${"nodeAffinity: { nodeSelector: { 'gitpod.io/workload_services': 'true' }, },"}
```
meysholdt, line 69: this line should be removed, shouldn't it? The code looks like it was only executed for `withVM == false` before.

.werft/util/kubectl.ts
```diff
@@ -134,7 +134,7 @@ export function createNamespace(namespace: string, kubeconfig: string, shellOpts
 
     // (re-)create namespace
     [
-        `kubectl --kubeconfig ${kubeconfig} create namespace ${namespace}`,
+        `kubectl --kubeconfig ${kubeconfig} create namespace ${namespace} --dry-run=client -o yaml | kubectl --kubeconfig ${kubeconfig} apply -f -`,
```
meysholdt, line 137: Interesting pattern. You're generating the YAML before you apply this. Is this something we should keep in the code because there are advantages or is this more like a temporary debug thing and we should remove it for the sake of simplicity?

meysholdt (changes requested):
This looks great already! I found one line that looks like it should be deleted, too and two nits. See comments.

meysholdt (approved):
🚀 🚀 🚀

meysholdt:
Thx for updating the PR! The code changes LGTM, but I see the [build failing via](https://werft.gitpod-dev.com/job/gitpod-build-ljb-2529-remove-core-dev-preview-env.9/raw):
> Error Error: kubectl --kubeconfig /workspace/gitpod/kubeconfigs/k3s apply -f k8s.yaml exit with non-zero status code.
> The Job "migrations" is invalid: spec.template: Invalid value:... (there are more details in the log)

.werft/build.yaml
```diff
@@ -89,15 +80,6 @@ pod:
       mountPath: /mnt/secrets/monitoring-satellite-stackdriver-credentials
     - name: monitoring-satellite-preview-token
       mountPath: /mnt/secrets/monitoring-satellite-preview-token
-    - name: gcp-sa
-      mountPath: /mnt/secrets/gcp-sa
-      readOnly: true
-    - name: gcp-sa-release
```
meysholdt, line 95: The secret is still used blow, let's not delete it:
https://github.com/gitpod-io/gitpod/blob/a365a2ade31ca94477484df214f3f0cf3b0ad7e6/.werft/jobs/build/build-and-publish.ts#L46

.werft/build.yaml
```diff
@@ -89,15 +80,6 @@ pod:
       mountPath: /mnt/secrets/monitoring-satellite-stackdriver-credentials
     - name: monitoring-satellite-preview-token
       mountPath: /mnt/secrets/monitoring-satellite-preview-token
-    - name: gcp-sa
```
meysholdt, line 92: The secret is still used in several places (search for `/mnt/secrets/gcp-sa`), let's not delete it.

meysholdt (changes requested):
After taking a second look, I think two of the secrets are still in use.

liam-j-bennett:
Build issue seems to have fixed itself

meysholdt (approved):
🚀 🚀 🚀

PR #10796, opened 2022-06-21
[supervisor] Make compatible with run-gp

csweichel:
## Description
Makes supervisor compatible with run-gp and cleans up some code.

## How to test
In Gitpod no behaviour should change.

To test this with run-gp:
1. open a workspace on https://github.com/gitpod-io/run-gp
2. run `./hack/update-images.sh cw-run-gp-compat.3`
3. run `./hack/update-assets.sh`
4. run `go run main.go`

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/supervisor/cmd/run.go
+++ b/components/supervisor/cmd/run.go
@@ -12,6 +12,10 @@ import (
 	"github.com/gitpod-io/gitpod/supervisor/pkg/supervisor"
 )
 
+var runOpts struct {
+	RunGP bool
+}
+
 var runCmd = &cobra.Command{
 	Use:   "run",
 	Short: "starts the supervisor",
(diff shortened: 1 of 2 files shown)


mustard-mh (approved):
Code LGTM

- Maybe we should fix this `gp preview` command so that user can access URL in vscode built-in browser. 
- Cannot access task port 5000 (defined in .gitpod.yml) in preview env
- Log workspace location of `run-gp` is not correct (always `http://localhost:8080/?folder=/workspace`, but we need `http://localhost:8080/?folder=/workspace/<folder_name>`

Are they `run-gp` repo part's issue?

/hold

image

csweichel:
> * Maybe we should fix this `gp preview` command so that user can access URL in vscode built-in browser.
> * Cannot access task port 5000 (defined in .gitpod.yml) in preview env
> * Log workspace location of `run-gp` is not correct (always `http://localhost:8080/?folder=/workspace`, but we need `http://localhost:8080/?folder=/workspace/<folder_name>`
> 
> Are they `run-gp` repo part's issue? Feel free to unhold

Those are run-gp issues indeed. I'll file them in the repo.
Thank you for checking up on those.

/hold cancel

csweichel:
filed https://github.com/gitpod-io/run-gp/issues/13 and https://github.com/gitpod-io/run-gp/issues/14

Issue #10524, opened 2022-06-08
Improve server-side feature flags


jankeromnes:
### Bug description

Currently, we have the following feature-flag implementation in `dashboard`:

https://github.com/gitpod-io/gitpod/blob/e63d89afca4694f89fcefbeea888b177875900d3/components/dashboard/src/experiments/client.ts#L38-L55

This is great, because it allows us to specify and control feature-flags separately for staging and production, with custom rules as to who should get the feature flag.

We have a similar implementation in the `server`:

https://github.com/gitpod-io/gitpod/blob/e63d89afca4694f89fcefbeea888b177875900d3/components/server/src/experiments.ts#L13-L22

However, this one is problematic, because it only uses the feature-flags configuration from staging (even in production), and the API is somewhat different.

See also https://github.com/gitpod-io/gitpod/pull/10378#discussion_r892030476:

> It also lives under the root of server/src which it should not.

> This will in fact work for server. However, there's a second problem with it as it hard-codes a single Client ID which is for the preview/staging environment, but not production.

> Let's land as is, and fix it up in a follow-up PR. We'll need to move the file, but also inject the Client ID based on environment we're running in. The current version would also break self-hosted environments as it reaches out to an external source.

### Steps to reproduce

1. Try to use feature flags in `dashboard` / in `server`

### Expected behavior

The `dashboard` and `server` APIs should be similar, and use different rule-sets for staging vs production

geropl:
@easyCZ I think you already addressed this right? I cannot find the PR/issues, though. :shrug:

easyCZ:
@geropl Yes, this fully landed this morning. Thanks for closing this one!

PR #10807, opened 2022-06-21
[experiments] Add abstraction for configcat to work in self-hosted

easyCZ:
## Description

Adds an interface in front of configcat. This allows the following:
1. Different SDK keys to be used for different environments (specified as ENV variable)
3. When not specified, uses a default implementation which always returns the default value. This is essential to not break air-gapped environments for self-hosted.

We use a shared wrapper interface for experiments implemented in `@gitpod-protocol`. We then either use a browser version of configcat client for dashboard, or `configcat-node` for backend.

## Related Issue(s)

Fixed: https://github.com/gitpod-io/gitpod/issues/10524

* Fixes air-gapped envs for self-hosted
* Depends on CONFIGCAT_SDK_KEY being added to the server pod, or any other pod that wants to use feature flags from TS on the server side - https://github.com/gitpod-io/gitpod/pull/10819

## How to test

1. Start a preview environment
2. Create team `Gitpod`
3. Try to configure usage based pricing - if you can see it, and configure it then this PR works

## Release Notes

```release-note
NONE
```

## Documentation

NONE

- [ ] /werft with-preview


components/gitpod-protocol/src/experiments/configcat-server.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    // Self-hosted installations do not set the ConfigCat SDK key, so always use a client which returns the default value.
+    if (sdkKey == undefined || sdkKey === "") {
+        client = newAlwaysReturningDefaultValueClient();
+        return client;
+    }
+
+    const configCatClient = configcat.createClient(sdkKey, {
+        logger: configcat.createConsoleLogger(LogLevel.Error),
+        maxInitWaitTimeSeconds: 0,
```
easyCZ, line 32: @jankeromnes I've pulled in your change into my refactor as well

jankeromnes: Thanks! Is this ready for review?

easyCZ: It is :)

easyCZ:
Can confirm this works in preview
Screenshot 2022-06-24 at 14 37 26

components/gitpod-protocol/src/experiments/configcat-server.ts
```diff
@@ -0,0 +1,37 @@
...
+    // We have already instantiated a client, we can just re-use it.
+    if (client !== undefined) {
+        return client;
+    }
+
+    // Retrieve SDK key from ENV Variable
+    const sdkKey = process.env.CONFIGCAT_SDK_KEY;
+
+    // Self-hosted installations do not set the ConfigCat SDK key, so always use a client which returns the default value.
+    if (sdkKey == undefined || sdkKey === "") {
```
geropl, line 25: nit: `==` always feels like a bug, and IMO we should never use it. `===` is equivalent in this case anyway.

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -1860,15 +1860,12 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
 
     protected async ensureIsUsageBasedFeatureFlagEnabled(user: User): Promise<void> {
         const teams = await this.teamDB.findTeamsByUser(user.id);
-        const isUsageBasedBillingEnabled = await getExperimentsClient().getValueAsync(
+        const isUsageBasedBillingEnabled = await getExperimentsClientForBackend().getValueAsync(
             "isUsageBasedBillingEnabled",
             false,
             {
-                identifier: user.id,
```
geropl, line 1867: What is the difference here? :thinking: 

Is this a free-form shape we :100: control?

geropl: (asking because the former shape looks like it followed a pattern/external guideline/external interface.

components/gitpod-protocol/src/experiments/configcat-server.ts
easyCZ: Thanks, this was a miss on my part.

components/server/ee/src/workspace/gitpod-server-impl.ts
easyCZ: User ID, Email and Region are first class, The rest are custom attributes we control

components/gitpod-protocol/src/experiments/configcat-server.ts
easyCZ: Fixed.

PR #10811, opened 2022-06-22
feature(supervisor): installation of dotfiles via Makefile

ghuntley:
## Description

I and many folks use a `Makefile` as a standard interface to dotfiles. This pull-request adds support for `Makefile`s

## How to test

* Change was directly applied using GitHub editor and is pending test
* Start instance, configure dotfiles repo with `Makefile` - does it work?

## Release Notes

```release-note
feature: installation of dotfiles via Makefile
```

## Documentation

Will require https://www.gitpod.io/docs/config-dotfiles to be updated once this ships.

### DIFF
--- a/components/supervisor/pkg/supervisor/supervisor.go
+++ b/components/supervisor/pkg/supervisor/supervisor.go
@@ -485,6 +485,7 @@ func installDotfiles(ctx context.Context, cfg *Config, tokenService *InMemoryTok
 			"install",
 			"bootstrap.sh",
 			"bootstrap",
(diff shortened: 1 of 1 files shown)


akosyakov:
@csweichel @iQQBot does it make sense? Could you review please?

andreafalzetti:
@ghuntley did you run some tests on this?

I think what @iQQBot is saying makes sense, you would not run `Makefile`instead you run the `make` cmd. 

Is that what you do for your dotfiles?

PR #10818, opened 2022-06-22
[usage] Fix flaky test for ListWorkspacesByID

easyCZ:
## Description

Sometimes (~15/100 from my repeated runs), the DB cleanup hook post tests would conflict with the data creation and would cause the tests to fail due to the exact missing IDs. Generating new random IDs hasn't failed in my test runs.

## How to test

Unit tests

## Release Notes

```release-note
NONE
```

## Documentation

NONE

### DIFF
--- a/components/usage/pkg/db/workspace_test.go
+++ b/components/usage/pkg/db/workspace_test.go
@@ -87,14 +87,10 @@ func stringToVarchar(t *testing.T, s string) db.VarcharTime {
 }
 
 func TestListWorkspacesByID(t *testing.T) {
-	conn := db.ConnectForTests(t)
-
 	workspaces := []db.Workspace{
-		dbtest.NewWorkspace(t, db.Workspace{ID: "gitpodio-gitpod-aaaaaaaaaaa"}),
-		dbtest.NewWorkspace(t, db.Workspace{ID: "gitpodio-gitpod-bbbbbbbbbbb"}),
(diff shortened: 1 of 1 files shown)


components/usage/pkg/db/workspace_test.go
```diff
@@ -90,8 +90,8 @@ func TestListWorkspacesByID(t *testing.T) {
 	conn := db.ConnectForTests(t)
 
 	workspaces := []db.Workspace{
-		dbtest.NewWorkspace(t, db.Workspace{ID: "gitpodio-gitpod-aaaaaaaaaaa"}),
-		dbtest.NewWorkspace(t, db.Workspace{ID: "gitpodio-gitpod-bbbbbbbbbbb"}),
+		dbtest.NewWorkspace(t, db.Workspace{}),
```
easyCZ, line 93: The change here is that it generates random IDs, rather than keeping them fixed. This doesn't matter as we reference the actually created IDs rather than the fixed values.

andrew-farries (approved):

easyCZ:
Found the problem. We created the connection at the start of the test, but re-used it for sub-tests. This meant the hooks didn't run properly.

andrew-farries:
What do you mean by hooks in this context?

easyCZ:
Hmm, main still fails on the 2^16 test. Trying to reproduce. https://werft.gitpod-dev.com/job/gitpod-build-main.3638

PR #10819, opened 2022-06-22
[experiments] Configure ConfigCat SDK through installer

easyCZ:
## Description

In https://github.com/gitpod-io/gitpod/pull/10807, we are adding support for configuring configcat through an ENV variable to target various ConfigCat environments.

The SDK Key is **NOT A SECRET**, it is a unique identifier of app just like Google Analytics has a unique identifier.

## Related Issue(s)

* Relates to #10807 

## How to test

Run installer with config:
```
experimental:
	webApp:
		configcatKey: "config-cat-key-here"
```
Observe it's in the resulting yaml

## Release Notes

```release-note
NONE
```

## Documentation

NONE


easyCZ:
@gitpod-io/engineering-self-hosted Would you be able to please help review this?

corneliusludmann (approved):

Issue #10043, opened 2022-05-16
extract cgroup library from ws-daemon in supervisor


akosyakov:
Supervisor and ws-daemon both need information provided by cgroups, but have slightly different implementation. We need to extract a common go library.

See [this discussion](https://github.com/gitpod-io/gitpod/pull/9796#discussion_r872432473) for more context.

Furisto:
@akosyakov I discussed this with @kylos101 and these are the tasks that need to be completed:
- Create the cgroup library
- Switch workspace components to the library
- Extend IWS with an endpoint that provides resource information that is using this library
- Supervisor uses this endpoint to retrieve resource information. That way the IDE team does not need to mess with linux and workspace implementation details.

Do you agree with the direction and is there anything else that you would need?

kylos101:
> It sounds good then the workspace team is going to take care about it?

@akosyakov the first 3 bullets, yes, I think so. For the 4th, for supervisor to use new IWS endpoint, what do you think? I thought IDE team may want to do consumption of the new IWS endpoint from supervisor.

akosyakov:
I think you need to refactor supervisor as well to make sure that library is aligned with our expectations.

Furisto:
Solved by https://github.com/gitpod-io/gitpod/pull/10836

PR #10836, opened 2022-06-22
Provide workspace resource information

### LINKED ISSUE #10839: gp top does not work with cgroup v2
### Bug description

gp top fails with the following error when executed in gen49 which is using cgroup v2:

`cannot get workspace resources: rpc error: code = Unknown desc = failed to read memory.limit_in_bytes: open /sys/fs/cgroup/memory/memory.limit_in_bytes: no such file or directory`

### Steps to reproduce

Execute gp top in gen49 cluster

### Workspace affected

n.a.

### Expected behavior

gp top should show cpu and memory consumption of the workspace

### Example repository

_No response_

### Anything else?

_No response_

Furisto:
## Description
Adds a workspace info service to the workspace, that allows the retrieval of information about the current workspace. For now resource limits and usage are supported. Supports systems with cgroup v1 and v2.

The service itself only exposes a single (ratelimited) operation that is then delegated to IWS. This indirection was introduced to limit the access of user code to IWS.

In addition I introduced a (minimal) cgroup library as discussed in #10043 in order to support cgroup v1 and v2. Moving other code in e.g. ws-daemon over to use this library is outside of scope for this PR.

## Related Issue(s)

Fixes #10839 and #10043

## How to test
- Open workspace in preview environment and execute gp top. Our preview environments are still on cgroup v1, if you want to test cgroup v2, you can use this [VM](https://g0dfc9476e946bdf8a8ce91.workspace-preview.gitpod-io-dev.com/workspaces) that I prepared or use workspace-preview yourself to test it.

## Release Notes

```release-note
Provide endpoint that allows retrieving information about the workspace from within the workspace
```

## Werft options:

- [x] /werft with-preview


components/workspacekit/cmd/rings.go
```diff
@@ -983,6 +999,60 @@ func connectToInWorkspaceDaemonService(ctx context.Context) (*inWorkspaceService
...
 }
 
+type workspaceInfoService struct {
+	socket net.Listener
+	server *grpc.Server
+	api.UnimplementedWorkspaceInfoServiceServer
+}
+
+func startInfoService(socketDir string) (func(), error) {
+	socketFN := filepath.Join(socketDir, "info.sock")
```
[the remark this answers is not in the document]
Furisto: Yes, one of the reasons it is still in draft :wink:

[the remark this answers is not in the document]
utam0k: 👍

components/ws-daemon/pkg/iws/iws.go
```diff
@@ -995,6 +1002,294 @@ func (wbs *InWorkspaceServiceServer) unPrepareForUserNS() error {
...
+	if err != nil {
+		return nil, xerrors.Errorf("failed to read memory limit: %w", err)
+	}
+
+	stats, err := memory.Stat()
+	if err != nil {
+		return nil, xerrors.Errorf("failed to read memory stats: %w", err)
+	}
+
+	if stats.InactiveFileTotal > 0 {
```
akosyakov, line 1176: it is not applicable to cgroup v2?

An idea is to inform users when they are really close to OOM kills. Inactive files should not be counted in used, since they can be released. Is it different for v2?

Furisto: It makes sense to bring it in line with cgroup v1. Looking through the documentation for memory.current it includes page cache, so we should subtract it if the goal is to show how close they are to OOM.

akosyakov: fyi I originally learned it from here: https://github.com/kubernetes/kubernetes/blob/d5e5f14508b02997b117ca9214b2fa349e57d0f9/test/e2e/node/node_problem_detector.go#L259-L293

I was investigating how GCP is showing, since they subtract evictable memory. And following the logic node problem detector I was able to get the same number like GCP:

aledbf (approved):

Furisto:
Hey @gitpod-io/engineering-webapp @gitpod-io/engineering-self-hosted could you please a look at this? We need it for the deployment of gen50 next week.

aledbf:
> We need it for the deployment of gen50 next week.

is today ;)

geropl (approved):
Took a while to get a workspace, but... works nicely! :tada:

Furisto:
@MrSimonEmms One of the installer tests [failed](https://werft.gitpod-dev.com/job/gitpod-build-fo-ws-info.38) because the order of the elements in the list did not match up. I do not think the order of the elements in the list is important for correctness so I [fixed](https://github.com/gitpod-io/gitpod/pull/10836/commits/39a5050fe160ca10fcff76d7e81964d811ef7937) the test by ensuring that the elements match up before the comparison and added the commit to this PR. If I am mistaken let me know.

Furisto:
@gitpod-io/engineering-ide Can you take a look please :pray:

mustard-mh (approved):
Works well 👍 

image

Issue #10461, opened 2022-06-03
Epic: Allow for custom annotations & env variables on K8s Resources for Self-Hosted Installations


lucasvaltl:
### Summary

Allow users to customise to adapt their deployment of self-hosted to the specifics of the environment/cluster it is applied to. 

### Context

Customers are finding it difficult to sometimes use KOTS because it does not allow them to add custom annotations or env variables to K8s resources. The installer produces a fixed manifest and that then gets applied to the kubernetes cluster. Any changes made forces them to re-apply those changes after an update to Gitpod, which can lead to confusion.

Specific reasons they need to customise / make changes:
- Annotate service to use internal [networking](quizlet)
- Add an environment variable to S3 when used as a registry storage backend (internal reference)

### Value

Allow Kots to work in more setups, because it can be customised to environment specific circumstances.

### Acceptance Criteria

- Customers can add generic annotations and env variables to resources via a configuration field in the KOTS UI. 
- This is well documented in our documentation.

### Measurement

- Even more customers start using kots

### Proposed solution
Allow for generic annotations on resources - have a way to set this in the KOTS UI & Installer Config. Then apply together with the installation. 

### Alternatives considered
1. Open issues for the missing configuration options customers are doing currently with post-processing and solve this properly
3. Rely on replicated git-ops flow (requires a feature request from us) for post-processing.

### Additional Context
- Likely related to https://github.com/gitpod-io/gitpod/issues/9981 and https://github.com/gitpod-io/gitpod/commit/0b599a83ace29230bab192231449fa55202a9e3a
- [Examples](https://github.com/gitpod-io/gitpod/issues/10461#issuecomment-1150003811) of what customers are currently customising: 

### Tasks

- [x] #10853
- [x] #10826
- [x] #10902
- [x] #10903
- [x] #10942
- [x] #10910
- [ ] #10908
- [x] #10840
- [x] #10941

corneliusludmann:
I doubt that we want to support post-processing in the near future. Instead, we should open issues for the missing configuration options customers are doing currently with post-processing and solve this properly. [See also this internal comment](https://github.com/gitpod-io/customers/issues/55#issuecomment-1146216048).

lucasvaltl:
Here is the current list of known example customisations that are being done by customers via post-processing: 

1. Change proxy service to a cluster IP instead of a loadbalancer type when using your own load balancer (this was changed via https://github.com/gitpod-io/gitpod/pull/10537): 

```
experimental:
[...]

PR #10841, opened 2022-06-22
[installer]: document the custom annotations

### LINKED ISSUE #10840: Document for Installer

MrSimonEmms:
## Description

Documentation for the custom annotations. Based upon https://github.com/gitpod-io/gitpod/issues/10461#issuecomment-1160504435

## Related Issue(s)

Fixes #10840

## How to test

Read

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/installer/README.md
+++ b/install/installer/README.md
@@ -142,6 +142,69 @@ defined in the spec and so will be deleted shortly after the jobs have run.
 
 # Advanced topics
 
+## Custom Annotations, Environment Variables and Labels
+
+There are times when it is desirable for custom annotations, environment
+variables and labels should be added to an installation.
+
+### Installer Config
+
+This can be added to the root of the `gitpod.config.yaml`. If nothing is passed in,
+no custom parameters are added in.
+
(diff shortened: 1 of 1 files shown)


nandajavarma:
Just wondering, should we document that the user will have to do a delete `--cascade=orphan` on `ws-manager` Deployment in the given example?

MrSimonEmms:
> Just wondering, should we document that the user will have to do a delete `--cascade=orphan` on `ws-manager` Deployment in the given example? Edit: If updating an existing installation, that is.

@nandajavarma I've had a think about it and I think, that in the README at least, we probably don't want to be adding complexities. These are mostly for our team anyway.

I've made the documentation [on the website](https://github.com/gitpod-io/website/pull/2316) more "how to get around" it. I decided against the `--cascade=orphan` as I don't think we should be adding that complexity if/when it goes wrong. I'd rather suggest something simple that works for customers and then they can come to us when they have specific needs

PR #10854, opened 2022-06-23
[usage] Use stripe clients rather than a singleton in the usage controller

andrew-farries:
## Description

Change the usage component's Stripe package to return client instances rather than using a Stripe singleton.

## Related Issue(s)

https://github.com/gitpod-io/gitpod/pull/10754#discussion_r901398982

## How to test

The usual test steps to set up usage based billing and generate a test invoice.

See eg https://github.com/gitpod-io/gitpod/pull/10801

## Release Notes

```release-note
NONE
```

## Werft options:
- [x] /werft with-preview

### DIFF
--- a/components/usage/pkg/stripe/stripe.go
+++ b/components/usage/pkg/stripe/stripe.go
@@ -12,9 +12,12 @@ import (
 
 	"github.com/gitpod-io/gitpod/common-go/log"
 	"github.com/stripe/stripe-go/v72"
-	"github.com/stripe/stripe-go/v72/customer"
-	"github.com/stripe/stripe-go/v72/usagerecord"
+	"github.com/stripe/stripe-go/v72/client"
 )
+
+type Client struct {
+	sc *client.API
+}
 
 type stripeKeys struct {
 	PublishableKey string `json:"publishableKey"`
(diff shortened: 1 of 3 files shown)


components/usage/pkg/stripe/stripe.go
```diff
@@ -12,35 +12,39 @@ import (
...
+}
+
 type stripeKeys struct {
 	PublishableKey string `json:"publishableKey"`
 	SecretKey      string `json:"secretKey"`
 }
 
 // Authenticate authenticates the Stripe client using a provided file containing a Stripe secret key.
-func Authenticate(apiKeyFile string) error {
+func Authenticate(apiKeyFile string) (*Client, error) {
```
easyCZ, line 28: I'd recommend passing in a struct with the credentials, instead of the Path. The path can be resolved during component initialization. For example, unit tests should need a path, just data.

components/usage/pkg/stripe/stripe.go
```diff
@@ -12,35 +12,39 @@ import (
...
 	var stripeKeys stripeKeys
 	err = json.Unmarshal(bytes, &stripeKeys)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	stripe.Key = stripeKeys.SecretKey
-	return nil
+	sc := &client.API{}
```
easyCZ, line 40: Would `client.New(key, backends)` be more appropriate here?

components/usage/pkg/stripe/stripe.go
easyCZ:
```suggestion
func New(apiKeyFile string) (*Client, error) {
```
So that you can do `stripe.New(..)`

easyCZ (commented):
How are we going to mock the Stripe client so that we can test what requests we send to Stripe?

easyCZ:
I think using `stripe-mock` but directly in code, rather than starting another process is our best option for now. The only other alternative is to expose a (small) interface which matches Stripe for the API methods we need and stub these out for our purposes.

Using a real account is not really an option for unit tests, they will run possibly in parallel (from multiple PRs) and we'd have a hard time ensuring they are not flaky.

andrew-farries:
Thanks 👍 We'll add tests for the Stripe integration in a later PR. Could you approve if you are happy with these changes 🙏 .

/hold

easyCZ:
I was waiting for the parent PR to land, otherwise it would've merged this one into the parent PR

easyCZ (approved):

PR #10857, opened 2022-06-23
[installer]: add customization functions to components

### LINKED ISSUE #10853: Apply dummy functions to all components
Add dummy functions to add in custom envvars, labels, annotations. This will allow the other teams to approve whilst we do the work on the actual functions in the background.

Helm components will be a future PR

MrSimonEmms:
## Description

Adds in customisation functions to annotations, environment variables and labels. The functions currently return what's input, but the functions will be completed in a future PR - it's been done this way to reduce the number of urgent PRs submitted to each team.

It's only applied to limited components at the moment to allow for self-hosted customers the flexibility they require in installing Gitpod in their own environments.

Things to check:
- I've put the correct `TypeMeta` in the functions - these should be the same as in the parent object (eg, it should be a deployment in a deployment, a configmap in a configmap etc). **NB** pods should have their parent `TypeMeta`
- I've correctly entered any existing key/value pairs

I have tested it and confirm that I can build an image and open a new workspace

## Context

Customers need to set custom annotations, env vars, and labels to some components. To make it easier for all teams to review, this PR does not implement this but simply extracts the setting of annotations, env vars, and labels to the functions `CustomizeAnnotation`, `CustomizeEnvvar`, and `CustomizeLabel` in `common/customize`. These functions do not change anything yet but allow us to implement this behaviour in a future PR without the need that every team has to review this again.

**Uff. These are a lot of changed files. :sweat:**

Indeed. But actually, the changes are:
- `common/customize.go` that introduces the customize functions that do not change something in the behavior yet but simply return the annotations/env vars/labels that are given via the `existing...` param ([identity function](https://en.wikipedia.org/wiki/Identity_function) :wave:).
- All other changes are only the replacement of setting the annotations/env vars/labels directly in favour of calling this function.

It's just a refactoring. No new logic yet.

In a follow-up PR, we will extend the customize functions to let customers set custom annotations/env vars/labels.

## Related Issue(s)

Fixes #10853

## How to test

Deploy to preview environment

## Release Notes

```release-note
[installer]: add customization to components
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/installer/pkg/common/customize.go
+++ b/install/installer/pkg/common/customize.go
@@ -1,0 +1,37 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the MIT License. See License-MIT.txt in the project root for license information.
+
+package common
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func CustomizeAnnotation(ctx *RenderContext, component string, typeMeta metav1.TypeMeta, existingAnnotations ...func() map[string]string) map[string]string {
+	annotations := make(map[string]string, 0)
+
+	for _, e := range existingAnnotations {
+		for k, v := range e() {
+			annotations[k] = v
+		}
+	}
+
+	return annotations
+}
+
+func CustomizeEnvvar(ctx *RenderContext, component string, existingEnvvars []corev1.EnvVar) []corev1.EnvVar {
+	return existingEnvvars
+}
+
+func CustomizeLabel(ctx *RenderContext, component string, typeMeta metav1.TypeMeta, existingLabels ...func() map[string]string) map[string]string {
+	labels := DefaultLabels(component)
+
+	for _, e := range existingLabels {
+		for k, v := range e() {
+			labels[k] = v
+		}
+	}
+
+	return labels
+}
--- a/install/installer/pkg/components/agent-smith/configmap.go
+++ b/install/installer/pkg/components/agent-smith/configmap.go
@@ -49,9 +49,10 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		&corev1.ConfigMap{
 			TypeMeta: common.TypeMetaConfigmap,
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      Component,
-				Namespace: ctx.Namespace,
-				Labels:    common.DefaultLabels(Component),
+				Name:        Component,
+				Namespace:   ctx.Namespace,
+				Labels:      common.CustomizeLabel(ctx, Component, common.TypeMetaConfigmap),
+				Annotations: common.CustomizeAnnotation(ctx, Component, common.TypeMetaConfigmap),
 			},
 			Data: map[string]string{
 				"config.json": string(fc),
--- a/install/installer/pkg/components/agent-smith/daemonset.go
+++ b/install/installer/pkg/components/agent-smith/daemonset.go
@@ -17,7 +17,7 @@ import (
 )
 
 func daemonset(ctx *common.RenderContext) ([]runtime.Object, error) {
-	labels := common.DefaultLabels(Component)
+	labels := common.CustomizeLabel(ctx, Component, common.TypeMetaDaemonset)
 
 	configHash, err := common.ObjectHash(configmap(ctx))
 	if err != nil {
@@ -30,16 +30,19 @@ func daemonset(ctx *common.RenderContext) ([]runtime.Object, error) {
 			Name:      Component,
 			Namespace: ctx.Namespace,
 			Labels:    labels,
-			Annotations: map[string]string{
-				common.AnnotationConfigChecksum: configHash,
-			},
+			Annotations: common.CustomizeAnnotation(ctx, Component, common.TypeMetaDaemonset, func() map[string]string {
+				return map[string]string{
+					common.AnnotationConfigChecksum: configHash,
+				}
+			}),
 		},
 		Spec: appsv1.DaemonSetSpec{
 			Selector: &metav1.LabelSelector{MatchLabels: labels},
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:   Component,
-					Labels: labels,
+					Name:        Component,
+					Labels:      labels,
+					Annotations: common.CustomizeAnnotation(ctx, Component, common.TypeMetaDaemonset),
 				},
 				Spec: corev1.PodSpec{
 					Affinity:                      common.NodeAffinity(cluster.AffinityLabelWorkspacesRegular, cluster.AffinityLabelWorkspacesHeadless),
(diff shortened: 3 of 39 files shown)


geropl (approved):
Reviewed WebApp components, and deployment LGTM :+1:

MrSimonEmms:
I've put a diff file in to show the difference. `orig.yaml` is run without these changes, `change.yaml` is run with these changes and the `yaml.diff` shows the difference.

It's the randomly generated strings causing changes in `gitpod.io/checksum_config`

[diff.zip](https://github.com/gitpod-io/gitpod/files/8966686/diff.zip)

corneliusludmann (approved):
Looks good.

mustard-mh (approved):
LGTM

/hold

Will we implement that mock function?

MrSimonEmms:
> Will we implement that mock function?

Yes, that's coming in a future PR #10827. We did it like this to avoid having to keep asking every team to rush and do an incremental change - do a single big change and then our team will implement the function later

MrSimonEmms:
@mustard-mh are you ok if I unhold then?

aledbf:
@MrSimonEmms are you going to "protect" the default labels? Changing those could [impact ws-manager](https://github.com/gitpod-io/gitpod/blob/5abab0393ec55f356dc96356aa40de3fbcc7edca/components/ws-manager/pkg/manager/manager.go#L1312-L1313) for instance

Furisto (approved):

MrSimonEmms:
> @MrSimonEmms are you going to "protect" the default labels? Changing those could [impact ws-manager](https://github.com/gitpod-io/gitpod/blob/5abab0393ec55f356dc96356aa40de3fbcc7edca/components/ws-manager/pkg/manager/manager.go#L1312-L1313) for instance

@aledbf No. By definition of the requirement, we're explicitly allowing users to change (and potentially break) things. The KOTS config will have a warning line, but it's definitely a risk.

As a future iteration, we could (and probably should) protect certain things, but the 🛹 is to allow for custom overrides on all labels/envvars/annotations in the supported resources

MrSimonEmms:
> If we merge this change and change `app` and/or `component` you break the connection between `ws-manager `and `ws-daemon`

Yes. I agree it's potentially problematic.

If you'd prefer for me to remove this from some components, I can do - would it just be labels in `ws-manager` and `ws-daemon`?

MrSimonEmms:
@aledbf if it's ok with you, I'm going to unhold this. You are absolutely correct in your observation, but it's something that a user would have to specifically do in order to achieve this - it's not something that they could do by accident.

This should be treated with the same caution that Firefox gives when going to `about:config` - and the KOTS form will explicitly say this
image

aledbf:
> I'm going to unhold this

I'm afraid I don't agree with that decision. I prefer to escalate to @csweichel

Keep in mind this is (potentially) support overhead for the self-hosted team or worse, break a Gitpod installation.

corneliusludmann:
As far as I can see (@MrSimonEmms correct me when I'm wrong), the decision of whether we protect existing annotations/env vars/labels or not is independent of this PR. Let's merge this PR and discuss whether we want to protect existing once in the issue.

MrSimonEmms:
@corneliusludmann That is correct. This PR applies the dummy functions only. We can remove/amend/add protection prior to merging the implementation of the functions.

I can also see why @aledbf is (rightly) protective of the `ws-daemon` and `ws-manager` because, once this is merged, we won't require any approval from @gitpod-io/engineering-workspace in the implementation of the functions.

As a compromise, shall I [remove the controversial sections](https://github.com/gitpod-io/gitpod/pull/10857#issuecomment-1164486925) from this PR? I don't believe we have any current customers wishing to make customisations here (@lucasvaltl can you confirm?) then then we can add this in when necessary?

corneliusludmann:
> As a compromise, shall I remove the controversial sections from this PR?

Just to be sure: To my understanding, the concern is about removing/overriding existing annotations and not about adding annotations to these components at all, right?

> once this is merged, we won't require any approval from teams/engineering-workspace in the implementation of the functions

Rising this concern here is totally fine and that we need to agree on something is obvious. However, we could agree on [not overriding existing annotations](https://github.com/gitpod-io/gitpod/issues/10461#issuecomment-1164553319) in the issue and anyone can trust us that we implement this accordingly. :innocent: In my opinion, reviews, the way we use them, are mainly an instrument to ensure quality (second pair of eyes) instead of getting the “permit” to change something.

That said, given the fact that the `CustomOverride...` functions don't do anything yet (in this PR), I don't think that it makes sense to implement something here to preserve existing annotations. This should be solved in [the PR](https://github.com/gitpod-io/gitpod/pull/10827) that implements the functions.

**tl;dr** – I would vote for: merging this PR now, agreeing on not overriding existing annotations in the issue, and implementing this accordingly in the follow-up PR.

csweichel:
> No. By definition of the requirement, we're explicitly allowing users to change (and potentially break) things. The KOTS config will have a warning line, but it's definitely a risk.

That's not how I understand what we're trying to solve here. We want to enable the addition of labels and annotations to enable things like attributing cost to departments, or to customise behaviour on certain cloud platforms. We do not want to endanger the functioning of Gitpod which heavily relies on the correct labels and annotations being present. I see no value in breaking these invariants which would lead to nothing but a broken Gitpod installation.

> Keep in mind this is (potentially) support overhead for the self-hosted team or worse, break a Gitpod installation.

I very much agree with this. We want to make it easy to do the right thing. Breaking Gitpod is not the right thing, hence should not be easy. @mrzarquon (and others) needs to have an easier time in adapting a Gitpod installation to the requirements of a customer's infrastructure.

> We can remove/amend/add protection prior to merging the implementation of the functions.

This is a really easy change with the potential to avoid a harm down the road. If we come across a situation where we will want to be able to override a Gitpod-set label/annotation, we should either find out if that label/annotation needs to be there - or if it really makes sense at that point to enable the override of such values.

> In my opinion, reviews, the way we use them, are mainly an instrument to ensure quality (second pair of eyes) instead of getting the “permit” to change something.

Indeed, ensuring quality is the primary function of those reviews. Part of this quality assurance is also that the teams who are responsible for certain functions of the system can ensure that functionality. It's not permission as much as it is stewardship.

In the end a decision needs to be made, and it looks like we're at an impasse. What we want to achieve here is that a Gitpod installation is easier to adapt to a customer's environment. I fail to see a use-case where overriding a Gitpod annotation/label which the system likely depends on makes sense today. If we come across such an (unlikely) use-case we can still fix that particular situation.
We want to make it easy to adapt Gitpod, not to break it.

I'd very much vote for not overriding annotations and labels for the reasons outlined above.

MrSimonEmms:
> In the end a decision needs to be made, and it looks like we're at an impasse

A decision _does_ need to be made, although I'm not sure we're at an impasse - I always think of an impasse as something where's there no way forward, whereas this is just a different understanding of the requirements/consequences that we'll solve with open dialogue.

How would you prefer me to proceed?
1. Remove the label override from `ws-manager`, `ws-daemon` and anything else?
2. Prevent overriding of existing labels/annotations/envvars?

If I understand the two options, the former this PR will need a change. If it's the latter, this PR is good as-is*. As a third option, I'm happy with any other suggestion on how we can change the implementation.

\* Except for the conflict

csweichel:
> It was the original idea of the PR opener to allow to override of existing ideas, the concerns that this can break Gitpod have been raised and acknowledged

👍 

> What do you all think? Are we on the same page?

Yes - let's do this :)

MrSimonEmms:
Changes made as proposed - I'll leave the removal of the "hold" label to @csweichel or @aledbf as a final approval

Issue #10822, opened 2022-06-22
[Usage-based] Adjust `d_b_workspace_instance.attributedTeamId` to be not team-specific


jankeromnes:
## Is your feature request related to a problem? Please describe

Adjust `d_b_workspace_instance.attributedTeamId` to be not Team-specific, so that it can also explicitly include attributed User IDs when relevant.

This way, the `usage` component can simply group by this "attribution ID" without needing to join with other tables.

## Additional context

Optionally, we can also back-fill attribution IDs in the production DB from when the column was first deployed (as an ad-hoc, one-time prod DB fix).

jankeromnes:
Relevant sync summary from [Slack](https://gitpod.slack.com/archives/C02EN94AEPL/p1655905211038299?thread_ts=1655897526.086109&cid=C02EN94AEPL) (internal):

> We discussed that:
> - Jan would drive the effort to fix this soon-ish
> - we decided to have one column name `usageAttributionId`
> - with prefixes (where the separator should be something not part of the ID, so not `-`)
> - the migration does _not_ backfill data. We handle prod manually by either:
>   - setting userIds from the point in time we started adding teamIds (safe), or
>   - remove all values that do not start with `team`

PR #10868, opened 2022-06-23
Refactor workspace instance `attributedTeamId` to an explicit, not-team-specific `usageAttributionId`

jankeromnes:
## Description

Refactor workspace instance `attributedTeamId` to an explicit, not-team-specific `usageAttributionId`.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/10822

## How to test

1. Should build
2. Workspace instances should always get a `usageAttributionId`
    - If the workspace is for a project that's owned by a Team, the attribution ID should look like `team:1234`
    - If the workspace is for a user project, or for no project at all, the attribution ID should look like `user:1234`

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-db/src/typeorm/migration/1655988518555-WorkspaceInstanceUsageAttributionId.ts
+++ b/components/gitpod-db/src/typeorm/migration/1655988518555-WorkspaceInstanceUsageAttributionId.ts
@@ -1,0 +1,33 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { MigrationInterface, QueryRunner } from "typeorm";
+import { columnExists, indexExists } from "./helper/helper";
+
+const TABLE_NAME = "d_b_workspace_instance";
+const COLUMN_NAME = "usageAttributionId";
+const INDEX_NAME = "ind_usageAttributionId";
+
+export class WorkspaceInstanceUsageAttributionId1655988518555 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        if (!(await columnExists(queryRunner, TABLE_NAME, COLUMN_NAME))) {
+            await queryRunner.query(
+                `ALTER TABLE ${TABLE_NAME} ADD COLUMN ${COLUMN_NAME} varchar(60) NOT NULL DEFAULT '', ALGORITHM=INPLACE, LOCK=NONE`,
+            );
+        }
+        if (!(await indexExists(queryRunner, TABLE_NAME, INDEX_NAME))) {
+            await queryRunner.query(`CREATE INDEX ${INDEX_NAME} ON ${TABLE_NAME} (${COLUMN_NAME})`);
+        }
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        if (await columnExists(queryRunner, TABLE_NAME, COLUMN_NAME)) {
+            await queryRunner.query(
+                `ALTER TABLE ${TABLE_NAME} DROP COLUMN ${COLUMN_NAME}, ALGORITHM=INPLACE, LOCK=NONE`,
+            );
+        }
+    }
+}
--- a/components/gitpod-db/src/typeorm/entity/db-workspace-instance.ts
+++ b/components/gitpod-db/src/typeorm/entity/db-workspace-instance.ts
@@ -107,5 +107,5 @@ export class DBWorkspaceInstance implements WorkspaceInstance {
         default: "",
         transformer: Transformer.MAP_EMPTY_STR_TO_UNDEFINED,
     })
-    attributedTeamId?: string;
+    usageAttributionId?: string;
 }
--- a/components/gitpod-db/src/workspace-db.spec.db.ts
+++ b/components/gitpod-db/src/workspace-db.spec.db.ts
@@ -66,7 +66,7 @@ class WorkspaceDBSpec {
             ideImage: "unknown",
         },
         deleted: false,
-        attributedTeamId: undefined,
+        usageAttributionId: undefined,
     };
     readonly wsi2: WorkspaceInstance = {
         workspaceId: this.ws.id,
@@ -89,7 +89,7 @@ class WorkspaceDBSpec {
             ideImage: "unknown",
         },
         deleted: false,
-        attributedTeamId: undefined,
+        usageAttributionId: undefined,
     };
     readonly ws2: Workspace = {
         id: "2",
@@ -127,7 +127,7 @@ class WorkspaceDBSpec {
             ideImage: "unknown",
         },
         deleted: false,
-        attributedTeamId: undefined,
+        usageAttributionId: undefined,
     };
 
     readonly ws3: Workspace = {
@@ -165,7 +165,7 @@ class WorkspaceDBSpec {
             ideImage: "unknown",
         },
         deleted: false,
-        attributedTeamId: undefined,
+        usageAttributionId: undefined,
     };
 
     async before() {
--- a/components/gitpod-protocol/src/workspace-instance.ts
+++ b/components/gitpod-protocol/src/workspace-instance.ts
@@ -64,11 +64,10 @@ export interface WorkspaceInstance {
     workspaceClass?: string;
 
     /**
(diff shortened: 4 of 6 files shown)


components/gitpod-db/src/typeorm/migration/1655988518555-WorkspaceInstanceUsageAttributionId.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+import { MigrationInterface, QueryRunner } from "typeorm";
+import { columnExists } from "./helper/helper";
+
+export class UsageAttributionId1655988518555 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        if (!(await columnExists(queryRunner, "d_b_workspace_instance", "usageAttributionId"))) {
+            await queryRunner.query(
+                "ALTER TABLE `d_b_workspace_instance` CHANGE `attributedTeamId` `usageAttributionId` varchar(255) NOT NULL DEFAULT '', LOCK=NONE",
```
jankeromnes, line 14: FYI, I tried `ALGORITHM=INPLACE`, but MySQL wasn't happy about it:

```
query failed: ALTER TABLE `d_b_workspace_instance` CHANGE `attributedTeamId` `usageAttributionId` varchar(255) NOT NULL DEFAULT '', ALGORITHM=INPLACE, LOCK=NONE
error: Error: ER_ALTER_OPERATION_NOT_SUPPORTED_REASON: ALGORITHM=INPLACE is not supported. Reason: Cannot change column type INPLACE. Try ALGORITHM=COPY.
    at Query.Sequence._packetToError (/tmp/build/components-gitpod-db--dbtest-init.2db8f1d34fe1fa534d23cb857822e9eb7e0f3398/mig/node_modules/mysql/lib/protocol/sequences/Sequence.js:47:14)
```

jankeromnes: Oh no, `LOCK=NONE` also doesn't work 🙈

```
error: Error: ER_ALTER_OPERATION_NOT_SUPPORTED_REASON: LOCK=NONE is not supported. Reason: Cannot change column type INPLACE. Try LOCK=SHARED.
```

jankeromnes:
It's working! 🎉

```
mysql> select id, workspaceId, creationTime, usageAttributionId from d_b_workspace_instance;
+--------------------------------------+-----------------------------------+--------------------------+-------------------------------------------+
| id                                   | workspaceId                       | creationTime             | usageAttributionId                        |
+--------------------------------------+-----------------------------------+--------------------------+-------------------------------------------+
| 3610efe0-28f3-43f9-a742-62cf2b15c361 | autofixdev-autofix-qdb4hnp3343    | 2022-06-23T15:01:14.804Z | user:03c5bb79-361d-4956-ad0e-58221b378484 |
| 7e62f39b-8e70-40ea-9fe3-6776e3d51c09 | jankeromnes-metachess-ne9ernqjojg | 2022-06-23T15:02:12.145Z | team:b2ee4f72-2dab-48b9-9b2b-903387e577f2 |
| f63a6dcd-04ca-4ca7-b4ca-d2bdf5ca70af | jankeromnes-metachess-u8zwfhjtiqq | 2022-06-23T15:01:59.035Z | team:b2ee4f72-2dab-48b9-9b2b-903387e577f2 |
+--------------------------------------+-----------------------------------+--------------------------+-------------------------------------------+
3 rows in set (0.01 sec)
```

components/gitpod-db/src/typeorm/migration/1655988518555-WorkspaceInstanceUsageAttributionId.ts
```diff
@@ -0,0 +1,29 @@
...
+import { columnExists } from "./helper/helper";
+
+const TABLE_NAME = "d_b_workspace_instance";
+const COLUMN_NAME = "usageAttributionId";
+
+export class WorkspaceInstanceUsageAttributionId1655988518555 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        if (!(await columnExists(queryRunner, TABLE_NAME, COLUMN_NAME))) {
+            await queryRunner.query(
+                `ALTER TABLE ${TABLE_NAME} ADD COLUMN ${COLUMN_NAME} varchar(255) NOT NULL DEFAULT '', ALGORITHM=INPLACE, LOCK=NONE`,
```
geropl, line 17: > varchar(255)

What about making it smaller in max size? Mentioning this because the bigger the key, the bigger the index and the more costly operations on it will be. :watch:

components/gitpod-db/src/typeorm/migration/1655988518555-WorkspaceInstanceUsageAttributionId.ts
```diff
@@ -0,0 +1,29 @@
...
+            );
+        }
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        if (await columnExists(queryRunner, TABLE_NAME, COLUMN_NAME)) {
+            await queryRunner.query(
+                `ALTER TABLE ${TABLE_NAME} DROP COLUMN ${COLUMN_NAME}, ALGORITHM=INPLACE, LOCK=NONE`,
```
geropl, line 25: :+1:

components/gitpod-db/src/typeorm/migration/1655988518555-WorkspaceInstanceUsageAttributionId.ts
jankeromnes: Good catch! But `varchar`'s exact size depends on the actual content, right? And with one address byte, you can go up to 255 characters. So, I think there is no advantage in using `varchar(50)` over `varchar(255)`, other than preventing ourselves from using larger sizes one day (either by accident or because we need to).

From the [Drupal docs](https://www.drupal.org/docs/7/api/schema-api/data-types/varchar)

> **Storage Information :**
> L + 1 bytes if column values require 0 − 255 bytes, L + 2 bytes if values may require more than 255 bytes where L is L is the byte length of the string that you are storing and not the maximum size that you have set.

geropl: That's true for the column, but not for the indexex: It has to assume the max size for a given column. And that directly affects query times, especially if the index becomes too big for RAM (as we see for this table already).

If we want to later `GROUP BY` on this, an index is required.

geropl:
Nice.
image

components/gitpod-db/src/typeorm/migration/1655988518555-WorkspaceInstanceUsageAttributionId.ts
jankeromnes: Ah, that's an excellent point, thank you @geropl!

So, should I:
- change the type to `char(41)` (exactly enough for `team:` or `user:` prefix followed by UUID)?
- or make it slightly bigger? (e.g. `char(50)`?)
- and add an index to it in this migration?

geropl: I'm fine with sth like `char(50)` or `char(60)` to anticipate slighty longer prefixes. :+1: 

> and add an index to it in this migration?

Let's leave that for a follow-up. :lotus_position:

jankeromnes: > I'm fine with sth like `char(50)` or `char(60)` to anticipate slighty longer prefixes. 👍

How about `varchar(60)`? 😁

> Let's leave that for a follow-up. 🧘

I think it's just a couple more characters in the migration, right? 😇 Maybe I can add it now.

geropl (approved):
Tested and works, approving to unblock.

/hold So for [reducing the column size](https://github.com/gitpod-io/gitpod/pull/10868/files#r905842410). :+1:

jankeromnes:
@geropl FYI, I've changed the type to `varchar(60)` and created an index: https://github.com/gitpod-io/gitpod/compare/303f5aa3e8ee03eeb22724b355e608e349d1fe1c..b3c03b94c67f550292645717393a112af1f8c054

jankeromnes:
Still working as intended! 🎉 Many thanks for the review @geropl 🙏

/unhold

PR #10876, opened 2022-06-23
Clean up unused `attributedTeamId` column from workspace instances post-deployment

jankeromnes:
## Description

Once we've merged and successfully deployed https://github.com/gitpod-io/gitpod/pull/10868, we should merge this Pull Request to clean up the left-over `attributedTeamId` column.

Optionally, we may also want to "save" the historical attribution data from that older column and back-fill the newer column with it.

## Related Issue(s)

Follow-up to https://github.com/gitpod-io/gitpod/pull/10868

## How to test

1. Should build

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


easyCZ:
LGTM, I guess in self-hosted, when it runs migrations it will first add it and then remove it later on but that should be okay.

easyCZ (approved):

PR #10877, opened 2022-06-23
Assume that preview envs without `mysql-0` are not active.

meysholdt:
## Description
We have may preview envs running without `mysql-0` DB. To ensure they get garbage collected eventually, this PR treats them as *not* active. 

## Related Issue(s)
Context: https://gitpod.slack.com/archives/C01KGM9EBD4/p1655991725949059

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/.werft/platform-delete-preview-environments-cron.ts
+++ b/.werft/platform-delete-preview-environments-cron.ts
@@ -111,12 +111,12 @@ class HarvesterPreviewEnvironment {
             const kubectclCmd = `KUBECONFIG=${PREVIEW_K3S_KUBECONFIG_PATH} kubectl --insecure-skip-tls-verify`
 
             werft.log(sliceID, `${this.name} (${this.k3sNamespace}) - Checking status of the MySQL pod`)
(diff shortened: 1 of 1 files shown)


mads-hartmann (approved):
LGTM. Would be nice to follow up with https://github.com/gitpod-io/ops/issues/2758 just in case this means we're deleting preview envs prematurely

ArthurSens:
Do we know if it is common to have db connection issues during this check?

I believe we made this assumption because we were afraid of small connection issues that could delete previews that were actually in use.

If we make sure that we only spin up VMs when a preview is required, wouldn't it solve this problem of trying to connect to a non-existing database?

meysholdt:
> If we make sure that we only spin up VMs when a preview is required, wouldn't it solve this problem of trying to connect to a non-existing database?

I think not, because even then there can be preview envs without gitpod installed. E.g. the install failed or somebody simply deleted Gitpod. Even in those scenarios we need to be 100% sure the VM gets GCed eventually.

Idea: If there is connection trouble, garbage-collect less aggressively. E.g. after 5 days. Or 7 days. maybe somebody is just re-installing Gitpod.

PR #10727, opened 2022-06-17
[bridge] Refactor Bridge.controlInstances and add tracing (4/5)

geropl:
## Description

This PR prepares for fixing a bug with prebuilds being stuck in `queued` indefinitely in rare cases. (4/5 in the chain)
To that end, if mostly refactors the method `controlInstances` in `Bridge`, and adds tracing to it.

## Related Issue(s)

Context: https://github.com/gitpod-io/gitpod/issues/9395
Follow-up: #10882 

## How to test
 - start a [workspace on this branch](https://gitpod.io/#https://github.com/gitpod-io/gitpod/pull/10727) and `./dev/preview/install-k3s-kubeconfig.sh`
 - start a workspace in the preview env
 - enable debug logs for `ws-manager-bridge`: `gpctl debug log ws-manager-bridge`
 - check the logs and notice this pattern:
   - "Controlling instances..."
   - "Controlling running instances..."
   - "Done controlling running instances."
   - "Done controlling instances."

## Release Notes

```release-note
fix prebuilds stuck in `queued` indefinitely
```

## Documentation

- [x] /werft with-preview

PR #10882, opened 2022-06-23
[bridge] Update prebuild status when controlling instance timeouts (5/5)

geropl:
## Description

This PR fixes a bug with prebuilds being stuck in `queued` indefinitely in rare cases. (5/5 in the chain)

This turned out to be bigger then anticipated, because the `MetaInstanceController` we had in place beforehand was ill-conceived, and did only work for self-hosted installations. With this change it's part of `Bridge`, so going to be executed per workspace cluster, so it's going to run in production as well (as well as Self-Hosted, of course).

For documenting this change we now:
 - have a [diagram](https://www.notion.so/gitpod/WebApp-Architecture-784cf14bc76440e3beb8de56f61e7340#6c9bd05eef5a44408171a95b14ae73a9)
 - [updated description](https://www.notion.so/gitpod/WebApp-Architecture-784cf14bc76440e3beb8de56f61e7340#ed5196fd4fdb47319c42ec4a781d232c) for `BridgeController` and `Bridge` reflecting this change

## Related Issue(s)

Fixes: https://github.com/gitpod-io/gitpod/issues/9395, #8274
Depends on: https://github.com/gitpod-io/gitpod/pull/10727

## How to test
 - start a [workspace on this branch](https://gitpod.io/#https://github.com/gitpod-io/gitpod/pull/10426) and `./dev/preview/install-k3s-kubeconfig.sh`
 - `kubectl edit cm ws-manager-bridge-config` and make the timeout for the `building` phase `60`s
 - [start a prebuild](https://gpl-9395-sccfe07d632.preview.gitpod-dev.com/workspaces) that is depending on an image build that takes longer than 1min
   - for instance: clone [this branch](https://github.com/geropl/gitpod-test-repo/blob/gpl/docker-build-10mins/.gitpod/Dockerfile)
   - create a project for it
   - update that Dockerfile
 - check how workspace and prebuild get updated and marked as stopped/aborted in the DB
   - watch the ws-manager-bridge logs for: "Marking workspace instance as stopped"
   - that log line contains the workspaceId of the prebuild workspace
 - check the DB tables `d_b_workspace_instance` and `d_b_prebuilt_workspace` to both show that those are stopped/aborted :heavy_check_mark: 

## Release Notes

```release-note
fix prebuilds stuck in `queued` indefinitely
```

## Documentation

- [x] /werft with-preview


components/ws-manager-bridge/src/bridge.ts
```diff
@@ -474,12 +474,7 @@ export class WorkspaceManagerBridge implements Disposable {
 
             for (const [instanceId, ri] of runningInstancesIdx.entries()) {
                 const instance = ri.latestInstance;
-                if (
-                    !(
-                        instance.status.phase === "running" ||
-                        durationLongerThanSeconds(Date.parse(instance.creationTime), maxTimeToRunningPhaseSeconds)
```
geropl, line 480: The `maxTimeToRunningPhaseSeconds` functionality is effectively covered by `contrlInstanceTimeouts` now, as it has specific timeouts for `preparing` and `building` phases.
The difference is that it no longer covers the other, ws-manager-goverened, pre-running phases (`pending`, `creating`, `initializing`). But as this was introduced a very long time ago as a bandaid to cover for a ws-manager bug, I think we're better of removing it.

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -503,6 +498,81 @@ export class WorkspaceManagerBridge implements Disposable {
...
+                installation: this.cluster.name,
+            });
+            TraceContext.setError(ctx, err);
+        } finally {
+            span.finish();
+        }
+    }
+
+    protected async controlInstanceTimeouts(parentCtx: TraceContext, info: RunningWorkspaceInfo) {
+        const logContext: LogContext = {
```
geropl, line 532: The following code is copied over from `MetaInstanceController`.

components/ws-manager-bridge/src/bridge.ts
geropl: compare this diagram: https://www.notion.so/gitpod/WebApp-Architecture-784cf14bc76440e3beb8de56f61e7340#6c9bd05eef5a44408171a95b14ae73a9

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -503,6 +498,75 @@ export class WorkspaceManagerBridge implements Disposable {
...
+            instanceId: info.latestInstance.id,
+        };
+        const ctx = TraceContext.childContext("controlDBInstance", parentCtx);
+        try {
+            const now = Date.now();
+            const creationTime = new Date(info.latestInstance.creationTime).getTime();
+            const timedOutInPreparing = now >= creationTime + this.config.timeouts.preparingPhaseSeconds * 1000;
+            const timedOutInBuilding = now >= creationTime + this.config.timeouts.buildingPhaseSeconds * 1000;
+            const timedOutInUnknown = now >= creationTime + this.config.timeouts.unknownPhaseSeconds * 1000;
+            const currentPhase = info.latestInstance.status.phase;
```
easyCZ, line 543: It is somewhat annoying that we have to do this here. Ideally, we'd always be able to ask a ws-manager about a particular WSI and it would tell us if it is in fact in a failed state based on these timeouts.

easyCZ (commented):
Code LGTM. Just running through the test scenario now.

easyCZ (approved):

Issue #10656, opened 2022-06-14
Fix briefly incorrect state when upgrading to Usage-Based Billing


jankeromnes:
### Bug description

Part of #9036

When you upgrade a team to Usage-Based Billing, you might see a briefly incorrect state in the UI before the upgrade shows as "pending":

> The UI embedded on the Gitpod side went through a slightly confusing transition where it went back to a 'not enabled' state after putting in card details, before transitioning to the 'manage billing' UI.

Source: https://github.com/gitpod-io/gitpod/pull/10630#pullrequestreview-1005751159

### Steps to reproduce

1. Upgrade a team to Usage-Based Billing by adding a credit card
2. After submitting, you might briefly see the UI in "Billing: Not enabled" state with an active "Add Payment Method" button (i.e. what you just did, which is confusing) before it eventually switched to the correct "pending" state, and then finally the "Billing: Enabled" state when the Stripe processing is done

PR #10890, opened 2022-06-24
Allow teams to cancel and renew their usage-based subscription in Stripe

jankeromnes:
## Description

Allow teams to cancel and renew their usage-based subscription in Stripe.

**Before**:
- We show subscription status based on whether the Stripe customer exists
- If the Stripe customer has a payment issue, or cancels their subscription, they are "stuck" in "Active Billing" status (but without a subscription and no way to renew it)

**After**:
- We show subscription status based on whether a valid Stripe subscription exists
- If the Stripe customer has a payment issue, or cancels their subscription, the UI goes back to the "Inactive Billing" (allowing an existing customer to create a new subscription)

## Related Issue(s)

Fixes #

## How to test

1. 

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/server/ee/src/workspace/gitpod-server-impl.ts
+++ b/components/server/ee/src/workspace/gitpod-server-impl.ts
@@ -1904,18 +1904,22 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
         }
     }
 
-    async findStripeCustomerIdForTeam(ctx: TraceContext, teamId: string): Promise<string | undefined> {
-        const user = this.checkAndBlockUser("findStripeCustomerIdForTeam");
+    async findStripeSubscriptionIdForTeam(ctx: TraceContext, teamId: string): Promise<string | undefined> {
+        const user = this.checkAndBlockUser("findStripeSubscriptionIdForTeam");
         await this.ensureIsUsageBasedFeatureFlagEnabled(user);
         await this.guardTeamOperation(teamId, "update");
         try {
             const customer = await this.stripeService.findCustomerByTeamId(teamId);
-            return customer?.id || undefined;
+            if (!customer?.id) {
+                return undefined;
+            }
+            const subscription = await this.stripeService.findUncancelledSubscriptionByCustomer(customer.id);
+            return subscription?.id;
         } catch (error) {
-            log.error(`Failed to get Stripe Customer ID for team '${teamId}'`, error);
+            log.error(`Failed to get Stripe Subscription ID for team '${teamId}'`, error);
             throw new ResponseError(
                 ErrorCodes.INTERNAL_SERVER_ERROR,
-                `Failed to get Stripe Customer ID for team '${teamId}'`,
+                `Failed to get Stripe Subscription ID for team '${teamId}'`,
             );
         }
     }
@@ -1931,7 +1935,10 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
         await this.guardTeamOperation(teamId, "update");
         const team = await this.teamDB.findTeamById(teamId);
         try {
-            const customer = await this.stripeService.createCustomerForTeam(user, team!, setupIntentId);
+            let customer = await this.stripeService.findCustomerByTeamId(team!.id);
+            if (!customer) {
+                customer = await this.stripeService.createCustomerForTeam(user, team!, setupIntentId);
+            }
             await this.stripeService.createSubscriptionForCustomer(customer.id, currency);
         } catch (error) {
             log.error(`Failed to subscribe team '${teamId}' to Stripe`, error);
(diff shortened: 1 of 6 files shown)


components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -1904,18 +1904,22 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
-        const user = this.checkAndBlockUser("findStripeCustomerIdForTeam");
+    async findStripeSubscriptionIdForTeam(ctx: TraceContext, teamId: string): Promise<string | undefined> {
+        const user = this.checkAndBlockUser("findStripeSubscriptionIdForTeam");
         await this.ensureIsUsageBasedFeatureFlagEnabled(user);
         await this.guardTeamOperation(teamId, "update");
         try {
             const customer = await this.stripeService.findCustomerByTeamId(teamId);
-            return customer?.id || undefined;
+            if (!customer?.id) {
+                return undefined;
```
easyCZ, line 1914: As in another PR, I think this should be returning NotFound and the client handling not found as No subscription exists.

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -1931,7 +1935,10 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
         await this.guardTeamOperation(teamId, "update");
         const team = await this.teamDB.findTeamById(teamId);
         try {
-            const customer = await this.stripeService.createCustomerForTeam(user, team!, setupIntentId);
+            let customer = await this.stripeService.findCustomerByTeamId(team!.id);
+            if (!customer) {
+                customer = await this.stripeService.createCustomerForTeam(user, team!, setupIntentId);
+            }
```
easyCZ, line 1941: It seems it is possible for 2 parallel requests to `subscribeTeamToStripe` to succeed and create 2 different subscriptions. Is this somehow transactional on the Stripe side?

easyCZ (commented):
How is the cancellation triggered by the user? I'm assuming it's using the customer portal.

jankeromnes:
Thanks @easyCZ for the initial feedback!

> How is the cancellation triggered by the user? I'm assuming it's using the customer portal.

Please read the [provided test instructions](https://github.com/gitpod-io/gitpod/pull/10890#issue-1283423051):

> 5. In the Stripe Portal, click on "Cancel" next to your subscription, and go back to Gitpod

components/server/ee/src/workspace/gitpod-server-impl.ts
jankeromnes: Thanks for identifying this pre-existing problem! I don't think it can realistically happen, but I'm happy to open a follow-up issue about this if it seems important. 💡

components/server/ee/src/workspace/gitpod-server-impl.ts
jankeromnes: Until now, I think we've had our API return `something | undefined` for such `findSomething` methods.

We could change this pattern, but then maybe it would make sense to change all these methods?

easyCZ: Up to you. I'd mostly suggest we do this opportunistically on apis we work on as part of features.

easyCZ:
Just ran through the steps:

easyCZ:
Just ran through the steps:

* After I added a payment method, for 2-3seconds, the view still continued to say `Inactive` and only then a loading screen indicator showed.
* I've also managed to create 2 subscriptions for my team. This then results in the following error (as implemented)
```
TeamUsageBasedBilling.tsx:43 Error: Failed to get Stripe Subscription ID for team 'TEAM_ID_FOO'
    at main.js:449:48
    at main.js:276:17
    at main.js:260:13
    at setImmediate.js:40:13
    at l (setImmediate.js:69:21)
    at n (setImmediate.js:109:17)
```
Here, the problem is the plan remains `Inactive` and I can still click Update payment info, add my credit card and create yet another subscription.

jankeromnes:
> - After I added a payment method, for 2-3seconds, the view still continued to say `Inactive` and only then a loading screen indicator showed.

Good catch! This is https://github.com/gitpod-io/gitpod/issues/10656, which I'm planning to fix soon. 📝

> - I've also managed to create 2 subscriptions for my team. This then results in the following error (as implemented)

Did you achieve this by quickly completing the billing setup again, as allowed by https://github.com/gitpod-io/gitpod/issues/10656? If so, when we fix the UI state, this bug will become much harder to reproduce.

However, I do see your point about the race condition / transaction, so I'll open a follow-up issue for this as well.

easyCZ (approved):
Works as advertised.v User can unsubscribe and then re-subscribe % modulo other related issues discussed.

Thanks for the change!

components/server/ee/src/workspace/gitpod-server-impl.ts
easyCZ: The controller already handles this case - https://github.com/gitpod-io/gitpod/pull/10854/files#diff-1f1047abea4ed15decb31dfdf94d84646a601fbacd5462dcef20f7f8b4de33dbR54 it will ignore any with multiple subscriptions and logs an error.

PR #10893, opened 2022-06-24
[Usage-based] When payment is enabled, attribute all workspace instance usage to an explicitly selected "billing account"

jankeromnes:
## Description

When payment is enabled, attribute all workspace instance usage to an explicitly selected "billing account".

TODO:

- [x] In the user's Billing settings, show all the teams that have usage-based billing enabled in a dropdown
- [x] Persist a user's selected "billing account"
- [x] When computing the `usageAttributionId` of a workspace instance, attribute based on the workspace owner's selected "billing account"

Follow-ups:
- Auto-select "billing account" when there is no ambiguity (e.g. when a team upgrades to usage-based billing, also update all its users' `usageAttributionId` where it makes sense)
- Auto-revert to user account when the team subscription is cancelled, or when the user leaves the team

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/10901

## How to test

1. Open a workspace with no project
2. Create a project under your user (this automatically triggers a prebuild, so no need to also start a workspace)
3. Create a Team A, with a Project A under it
4. Create a Team B, with a Project B under it
5. Verify attribution*:
  a. Workspace with no project --> your user
  b. Workspace with user project --> your user
  c. Workspace with Team A project --> your user
  d. Workspace with Team B project --> your user
6. Now, enable usage-based billing for Team A, by adding a credit card for it (e.g 4242 4242 4242 4242, 4/24, 424)
7. Next, in your user's Settings, go to the "Billing" page, and select "Team A" as a billing account (note: "Team B" should not be available, since it doesn't have usage-based billing enabled)
8. Open a new workspace with no project
9. Open a new workspace for your user project
10. Open a new workspace for Project A
11. Open a new workspace for Project B
12. Verify attribution*:
  a. Workspace with no project --> team A
  b. Workspace with user project --> team A
  c. Workspace with Team A project --> team A
  d. Workspace with Team B project --> team A

*To verify attribution, open this PR in Gitpod, then run:

```
kubectl port-forward mysql-0 3306 &
mysql -h 127.0.0.1 -pjBzVMe2w4Yi7GagadsyB gitpod
mysql> select workspaceId, usageAttributionId, creationTime from d_b_workspace_instance order by creationTime;
```

Example output:

```
+-----------------------------------+-------------------------------------------+--------------------------+
| workspaceId                       | usageAttributionId                        | creationTime             |
+-----------------------------------+-------------------------------------------+--------------------------+
| autofixdev-autofix-vyof8v6yj9i    | user:a7dcf253-f05e-4dcf-9a47-cf8fccc74717 | 2022-06-24T13:13:34.254Z |
| jankeromnes-metachess-g7xzjycgnqj | team:8eba990a-4f1c-4233-b80f-c45330f81dd6 | 2022-06-24T13:14:41.517Z |
| jankeromnes-metaboard-9uzkqymxtz0 | team:8eba990a-4f1c-4233-b80f-c45330f81dd6 | 2022-06-24T13:15:04.492Z |
| jankeromnes-metachess-zwbh1q0jis4 | user:a7dcf253-f05e-4dcf-9a47-cf8fccc74717 | 2022-06-24T13:15:38.521Z |
+-----------------------------------+-------------------------------------------+--------------------------+
```

You can also verify every user's `usageAttributionId` preference like so:

```
mysql> select name, additionalData->"$.usageAttributionId" from d_b_user;
+--------------------------+---------------------------------------------+
| name                     | additionalData->"$.usageAttributionId"      |
+--------------------------+---------------------------------------------+
| jankeromnes              | "team:8eba990a-4f1c-4233-b80f-c45330f81dd6" |
| agent-smith              | NULL                                        |
| builtin-workspace-prober | NULL                                        |
+--------------------------+---------------------------------------------+
```

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/dashboard/src/settings/Billing.tsx
+++ b/components/dashboard/src/settings/Billing.tsx
@@ -4,14 +4,48 @@
  * See License-AGPL.txt in the project root for license information.
  */
 
-import { useContext } from "react";
+import { Team } from "@gitpod/gitpod-protocol";
+import { useContext, useEffect, useState } from "react";
 import { Link } from "react-router-dom";
+import getSettingsMenu from "./settings-menu";
+import { ReactComponent as Spinner } from "../icons/Spinner.svg";
+import DropDown from "../components/DropDown";
 import { PageWithSubMenu } from "../components/PageWithSubMenu";
 import { PaymentContext } from "../payment-context";
-import getSettingsMenu from "./settings-menu";
+import { getGitpodService } from "../service/service";
+import { TeamsContext } from "../teams/teams-context";
+import { UserContext } from "../user-context";
 
 export default function Billing() {
+    const { user } = useContext(UserContext);
     const { showPaymentUI, showUsageBasedUI } = useContext(PaymentContext);
+    const { teams } = useContext(TeamsContext);
+    const [teamsWithBillingEnabled, setTeamsWithBillingEnabled] = useState<Team[] | undefined>();
+
+    useEffect(() => {
+        if (!teams) {
+            setTeamsWithBillingEnabled(undefined);
+            return;
+        }
+        const teamsWithBilling: Team[] = [];
+        Promise.all(
+            teams.map(async (t) => {
+                const subscriptionId = await getGitpodService().server.findStripeSubscriptionIdForTeam(t.id);
+                if (subscriptionId) {
+                    teamsWithBilling.push(t);
+                }
+            }),
+        ).then(() => setTeamsWithBillingEnabled(teamsWithBilling));
+    }, [teams]);
+
+    const setUsageAttributionTeam = async (team?: Team) => {
+        if (!user) {
+            return;
+        }
+        const additionalData = user.additionalData || {};
+        additionalData.usageAttributionId = team ? `team:${team.id}` : `user:${user.id}`;
+        await getGitpodService().server.updateLoggedInUser({ additionalData });
+    };
 
     return (
         <PageWithSubMenu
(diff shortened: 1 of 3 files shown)


components/server/src/user/user-service.ts
```diff
@@ -210,10 +210,31 @@ export class UserService {
...
+     * B. Project-based attribution: If payments are not enabled (e.g. Self-Hosted), we attribute:
+     *   - To the owner of the project (user or team), if the workspace is linked to a project
+     *   - To the user, iff the workspace is not linked to a project
+     *
      * @param user
      * @param projectId
      */
     async getWorkspaceUsageAttributionId(user: User, projectId?: string): Promise<string | undefined> {
+        // A. Billing-based attribution
+        if (this.config.enablePayment) {
```
geropl, line 226: Thx a lot for this PR, @jankeromnes ! :pray: 

However, this is sth I wanted to avoid: having the the attribution logic coupled with user/workspace services. :slightly_smiling_face: 

I'd rather move the (re-)attribution part over to the usage service. Sorry if I wasn't clear there - got a :exploding_head: after all those meetings this week.. :grimacing:

components/dashboard/src/settings/Billing.tsx
```diff
@@ -21,13 +55,45 @@ export default function Billing() {
...
             <h3>Usage-Based Billing</h3>
             <h2 className="text-gray-500">Manage usage-based billing, spending limit, and payment method.</h2>
-            <p className="mt-8">
-                Hint:{" "}
-                <Link className="gp-link" to="/teams/new">
-                    Create a team
-                </Link>{" "}
-                to set up usage-based billing.
-            </p>
+            <div className="mt-8">
```
geropl, line 58: :heart:

components/gitpod-protocol/src/protocol.ts
```diff
@@ -150,9 +150,10 @@ export interface AdditionalUserData {
     oauthClientsApproved?: { [key: string]: string };
     // to remember GH Orgs the user installed/updated the GH App for
     knownGitHubOrgs?: string[];
-
     // Git clone URL pointing to the user's dotfile repo
     dotfileRepo?: string;
+    // Identifies an explicit team or user ID to which all the user's workspace usage should be attributed to (e.g. for billing purposes)
+    usageAttributionId?: string;
```
geropl, line 156: I think this deserves a separate field on `User` (I'm aware of the overhead). This way we could easier consume it from the `usage` service.

components/server/src/user/user-service.ts
geropl: Happy to sync about this early next week to discuss and align. :calendar:

jankeromnes: If you don't like the `if (this.config.enablePayment)` check, we could also make this a `if (user.usageAttributionId)` check (i.e. only "force" the attribution if the user "wants" a specific attribution ID for all their instances)

components/server/src/user/user-service.ts
```diff
@@ -210,10 +210,31 @@ export class UserService {
...
+            if (!user.additionalData?.usageAttributionId) {
+                // No explicit user attribution ID yet -- attribute all usage to the user by default (regardless of project/team).
+                user.additionalData = user.additionalData || {};
+                user.additionalData.usageAttributionId = `user:${user.id}`;
+                await this.userDb.updateUserPartial(user);
```
geropl, line 231: @jankeromnes Sorry, one last thing: If feel that we should not update the `User` entity in this method. It should work without that update as well, right? :thinking:

jankeromnes: True -- the goal was to automatically default to the correct default value, but we can also use a default value without modifying the User. 💭

components/gitpod-protocol/src/protocol.ts
jankeromnes: Is this a must-have before merging this PR? (Note: This would also require new specific method(s) on the server API.)

components/server/src/user/user-service.ts
jankeromnes: As discussed, this can still be adjusted in follow-up PRs, for example when we implement the entitlement service. 👍

components/server/src/user/user-service.ts
jankeromnes: Fixed.

geropl (approved):
Thank you, LGTM!

Issue #10786, opened 2022-06-21
Slow evaluation of feature flags causes UI issues


andrew-farries:
Feature flag evaluation is currently very slow (in the order of a few seconds), which causes UI glitches when page components are rendered/activated on already loaded pages in response to feature flags becoming enabled.

Here the delay in evaluating the 'PVC' feature flag causes the corresponding checkbox to render disabled for several seconds before becoming enabled:

https://user-images.githubusercontent.com/8225907/174745252-9bd22413-431f-4967-a291-40221449b468.mov

jankeromnes:
Thanks for opening this issue @andrew-farries!

FYI, this issue also affects the Usage-Based Billing UI in Team Billing (also behind a feature flag).

> Feature flag evaluation is currently very slow (in the order of a few seconds)

Actually, the delay seems to be a bit random -- sometimes, it's consistently super fast (near instantaneous) on every page load, but other times it seems to take several seconds on each page load.

But, it seems indeed to be more often slow than fast.

PR #10897, opened 2022-06-24
Disable 5 second delay before returning feature flag value

jankeromnes:
## Description

While debugging with @easyCZ why getting feature flag values was sometimes so slow, we found that by default, our experiments client waits for 5 seconds before returning the initial value (see [all options](https://configcat.com/docs/sdk-reference/js/#auto-polling-default)):

Screenshot 2022-06-24 at 12 57 35

Disabling this delay makes getting feature flag values instantaneous. 🪄 ⚡

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/10786

## How to test

1. Create a team called "Gitpod"
2. Go to Team Billing
3. The Usage-Based UI should be instantly visible, even if you reload the page

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/dashboard/src/experiments/configcat.ts
+++ b/components/dashboard/src/experiments/configcat.ts
@@ -23,6 +23,7 @@ export function newProductionConfigCatClient(): Client {
     // clientKey is an identifier of our ConfigCat application. It is not a secret.
     const clientKey = "WBLaCPtkjkqKHlHedziE9g/TwAe6YyftEGPnGxVRXd0Ig";
     const client = configcat.createClient(clientKey, {
+        maxInitWaitTimeSeconds: 0,
         logger: configcat.createConsoleLogger(2),
     });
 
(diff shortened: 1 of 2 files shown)


components/dashboard/src/experiments/configcat.ts
```diff
@@ -35,6 +36,7 @@ export function newNonProductionConfigCatClient(): Client {
     // clientKey is an identifier of our ConfigCat application. It is not a secret.
     const clientKey = "WBLaCPtkjkqKHlHedziE9g/LEAOCNkbuUKiqUZAcVg7dw";
     const client = configcat.createClient(clientKey, {
+        maxInitWaitTimeSeconds: 0,
```
andrew-farries, line 39: What's the purpose of the default wait time of 5 seconds? Any downsides to setting it to 0?

andrew-farries (approved):
/hold

easyCZ (approved):

components/dashboard/src/experiments/configcat.ts
jankeromnes: I have no idea -- there seems to be no documentation for this feature apart from the one-line description shown in my PR description (see also ConfigCat's [source code](https://github.com/configcat/common-js/blob/14cab5585a4fc8bbe79c7c038ac0882b18b05ec9/src/ConfigCatClientOptions.ts#L127-L128))

Digging a bit into ConfigCat's history, it looks like this feature was already part of this [initial commit](https://github.com/configcat/common-js/commit/7d3cdc7d0796afbb434e0fbda7805581370298ff), which unfortunately doesn't provide much context as to why this feature is useful. 🙈

PR #10900, opened 2022-06-24
[bridge] Add log.error in case we are about to override a previous "failed" condition

geropl:
## Description
1 of 2 PRs that attempt to fix the bug described in #8596 
Explanation: https://github.com/gitpod-io/gitpod/issues/8596#issuecomment-1165542976

## Related Issue(s)

Context: #8596

## How to test
- start a workspace
- see that everything works as expected
- check ws-manager-bridge logs and search for "We received an empty "failed" condition overriding an existing one"

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


easyCZ (approved):
You could also consider using a metrics for that.

/hold

PR #10905, opened 2022-06-24
[db] Remove DeletedEntryGC and all references

### LINKED ISSUE #10904: [db] Get rid of outdated GC mechanism for Db objects
Remove tech debt: `DeletedEntryGC`

geropl:
## Description
Get rid of tech debt: The functionality is not used anymore (`DB_DELETED_ENTRIES_GC_ENABLED` is not set to `true` in any environment for months), and has been superseded by `PeriodicDBDeleter` for quite some time now (`runDbDeleter` defaults to `true` in self-hosted).

## Related Issue(s)

Fixes #10904 
Companion PR: https://github.com/gitpod-io/ops/pull/2946

## How to test
 - check if everything builds

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


easyCZ (approved):
LGTM.

Just to confirm my understanding, this would only be enabled in SaaS so we would not get into a situation where this config would be set on the ConfigMap in self-hosted but would fail to start as we wouldn't parse it. Hold for this Q, feel free to unhold.

/hold

geropl:
> so we would not get into a situation where this config would be set on the ConfigMap in self-hosted but would fail to start as we wouldn't parse it. 

:+1: In SaaS, `db-sync` performs these deletes for us. `PeriodicDBDeleter` has been introduced to do the same for self-hosted (using the same data source), and is already doing so the last couple of months.
I'm not sure when this DeletedEntryGC has been introduced, or when it got disabled again, but it served the same purpose, just required config duplication. :shrug: Time to :x: it.

PR #10911, opened 2022-06-24
[kots]: apply customization file to Installer

### LINKED ISSUE #10910: Configure KOTS input

MrSimonEmms:
## Description

Adds the config section to the KOTS dashboard for the customization file

**NB** this must be applied to a cluster which does not have Gitpod installed. If you have it installed, you must run `helm un -n gitpod gitpod` beforehand if you intend to change any labels with your customization file. This is because Kubernetes cannot change the labels of a deployment/daemonset/statefulset. As this is for a specific set of enterprise customers who have engaged us for support, this should be an acceptable compromise for the time being.

This may be fixed in a future PR.

This is not an issue if changing annotations

## Related Issue(s)

Fixes #10910

## How to test

Deploy via KOTS to a cluster, using this YAML as your customization file

```yaml
customization:
  - apiVersion: apps/v1
    kind: Deployment
    metadata:
      name: server
      annotations:
        hello: world
```

When you next run it, `kubectl get pods -n gitpod -l component=server -o jsonpath="{.items[0].metadata.annotations}"` should have the annotation `hello=word`

## Release Notes

```release-note
[kots]: apply customization file to Installer
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/kots/manifests/kots-config.yaml
+++ b/install/kots/manifests/kots-config.yaml
@@ -379,6 +379,21 @@ spec:
           default: "0"
           help_text: Enables additional customization options. Enable only when you know what you are doing!
 
+        - name: customization_patch
+          title: Gitpod customization patch (YAML file)
+          type: file
+          required: false
+          when: '{{repl ConfigOptionEquals "advanced_mode_enabled" "1" }}'
+          help_text: |
+            A file configured to override annotations, labels and environment variables in components. See our
+            [advanced documentation](https://www.gitpod.io/docs/self-hosted/latest/advanced/customization) for
+            more details and the format.
+
+            Labels are an immutable property in Kubernetes, so you may need to uninstall your Gitpod instance
+            before applying - run `helm uninstall -n {{repl Namespace }} gitpod`.
+
+            **WARNING** this should not be run if using in-cluster database and/or object storage.
+
         - name: config_patch
           title: Gitpod config patch (YAML file)
           type: file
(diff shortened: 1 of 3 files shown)


install/kots/manifests/kots-config.yaml
```diff
@@ -379,6 +379,13 @@ spec:
           default: "0"
           help_text: Enables additional customization options. Enable only when you know what you are doing!
 
+        - name: customization_patch
+          title: Gitpod customization patch (YAML file)
+          type: file
+          required: false
+          when: '{{repl ConfigOptionEquals "advanced_mode_enabled" "1" }}'
+          help_text: A file configured to override annotations, labels and environment variables in components. See our [advanced documentation](https://www.gitpod.io/docs/self-hosted/latest/advanced/customization) for more details and the format.
```
corneliusludmann, line 387: Can we have a statement here that changing this value lets the installation fail after the initial (first) installation?

corneliusludmann (approved):
Just a tiny comment. Feel free to address and merge it.

/hold

install/kots/manifests/kots-config.yaml
MrSimonEmms: We can, and I will do that.

Currently, the `help_text` isn't actually displayed - see https://github.com/replicated-collab/replicated-gitpod/issues/29 which I raised this morning

Issue #10766, opened 2022-06-20
Platform information in gitpod telemetry


Pothulapati:
## Is your feature request related to a problem? Please describe

Now that we have basic Telemetry of a Gitpod instance, It would be really useful to understand
the environment in which the cluster is running to service the users better i.e writing better
docs, integrations, etc.

The platform/distribution is one such important data to understand where the users
are. The values could be `gke`, `aks`, `eks`, etc.

This also relates to https://github.com/gitpod-io/gitpod/issues/9075, so that we have a way
to separate the preview install's telemetry from others.

## Describe the behaviour you'd like

Add a new `platform` attribute to the Gitpod telemetry. The identification can be
performed by using [Replicated's distribution static function
](https://docs.replicated.com/reference/template-functions-static-context#distribution)

To also accomodate the `preview` value, We can instead make this as a specific option
in `config.go`, and pass the right value during the `render` phase

## Additional context

Discussion thread in slack https://gitpod.slack.com/archives/C01KLC56NP7/p1655726762212639

PR #10925, opened 2022-06-27
[telemetry] Add new `telemetry.data` config option for telemetry

Pothulapati:
## Description

This PR adds a new `experimental.telemetry.data.platform` config option
that is then passed to `GITPOD_INSTALLATION_PLATFORM` env variable, which
is then transmitted by the `gitpod-telemetry` job as the `platform`
property into our telemetry store. The value of this env variable
is then overrided by each install method.

This env can have the following values:
- `installer` by default as when the installation is done directly
  through the installer, We don't have enough information
   about the environment.
-  [One of the `distribution` values](https://docs.replicated.com/reference/template-functions-static-context#distribution)
   when done through `kots`. This would have the distribution in which
   Gitpod is being installed, and is dependent on `kots`.
- `local-preview` when the installation is done through the new
   preview install method.

As the changes show in the PR, each installation method overrides
the value.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/10766

## How to test

Check the `gitpod-telemetry` cronjob to have the new env, that is set
to the appropriate value based on the installation method

## Release Notes

```release-note
[telemetry] Add new telemetry.data config option for telemetry
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/installation-telemetry/cmd/send.go
+++ b/components/installation-telemetry/cmd/send.go
@@ -45,6 +45,11 @@ var sendCmd = &cobra.Command{
 			return fmt.Errorf("GITPOD_INSTALLATION_VERSION envvar not set")
 		}
 
+		platform := os.Getenv("GITPOD_INSTALLATION_PLATFORM")
+		if platform == "" {
+			return fmt.Errorf("GITPOD_INSTALLATION_PLATFORM envvar not set")
+		}
+
 		client, err := analytics.NewWithConfig(segmentIOToken, analytics.Config{})
 		defer func() {
 			err = client.Close()
@@ -54,7 +59,8 @@ var sendCmd = &cobra.Command{
 			Set("version", versionId).
 			Set("totalUsers", data.TotalUsers).
 			Set("totalWorkspaces", data.TotalWorkspaces).
-			Set("totalInstances", data.TotalInstances)
+			Set("totalInstances", data.TotalInstances).
+			Set("platform", platform)
 
 		if data.InstallationAdmin.Settings.SendCustomerID {
 			properties.Set("customerID", data.CustomerID)
--- a/install/installer/pkg/components/gitpod/cronjob.go
+++ b/install/installer/pkg/components/gitpod/cronjob.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/gitpod-io/gitpod/installer/pkg/common"
 	"github.com/gitpod-io/gitpod/installer/pkg/config/v1"
+	"github.com/gitpod-io/gitpod/installer/pkg/config/v1/experimental"
 	batchv1 "k8s.io/api/batch/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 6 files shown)


roboquat:

<details>

</details>

components/installation-telemetry/cmd/send.go
```diff
@@ -56,6 +56,11 @@ var sendCmd = &cobra.Command{
 			Set("totalWorkspaces", data.TotalWorkspaces).
 			Set("totalInstances", data.TotalInstances)
 
+		platform := os.Getenv("GITPOD_INSTALLATION_PLATFORM")
+		if platform != "" {
+			properties.Set("platform", platform)
```
MrSimonEmms, line 61: Do we want to leave the `platform` as blank in the Segment data, or do we want something like `unknown` in there?

MrSimonEmms (approved):
/hold

Looks fine. I've approved with a comment - your choice what you do with it

install/kots/manifests/gitpod-installer-job.yaml
```diff
@@ -303,6 +303,11 @@ spec:
                 'del(select(.kind == "StatefulSet" and .metadata.name == "openvsx-proxy").status)' \
                 "${GITPOD_OBJECTS}/templates/gitpod.yaml"
 
+              echo "Gitpod: Update GITPOD_INSTALLATION_PLATFORM telemetry value"
+              yq eval-all --inplace \
+                '(select(.kind == "CronJob" and .metadata.name == "gitpod-telemetry") | (.spec.jobTemplate.spec.template.spec.containers[0].env[] | select(.name=="GITPOD_INSTALLATION_PLATFORM")).value |= "repl{{ Distribution }}"' \
+                "${GITPOD_OBJECTS}/templates/gitpod.yaml"
```
corneliusludmann, line 309: Could we maybe find a cleaner solution than this `yq` magic? This makes the installer job unnecessarily complicated, IMHO. Maybe a proper config value for the installer config?

install/installer/pkg/components/gitpod/cronjob.go
```diff
@@ -63,6 +63,10 @@ func cronjob(ctx *common.RenderContext) ([]runtime.Object, error) {
 												Name:  "GITPOD_INSTALLATION_VERSION",
 												Value: ctx.VersionManifest.Version,
 											},
+											{
+												Name:  "GITPOD_INSTALLATION_PLATFORM",
+												Value: "installer",
```
corneliusludmann, line 68: Similar to Simon's comment, I would propose to set `unknown` here as well instead of `installer` (which is not a platform in the strict sense). WDYT?

install/kots/manifests/gitpod-installer-job.yaml
MrSimonEmms: I guess we could stick it in the `experimental` config as we probably don't want to make this a publicly configurable parameter and default to `installer`. Only question is whether the Local Preview runs it in experimental mode

Pothulapati: We can update the `local-preview` to also run in experinmental mode. So, Wokring on adding this into the config

Pothulapati:
Tested this on both `local-preview` and `kots`, and the values are set as expected
- `local-preview`
```
Environment:
      GITPOD_INSTALLATION_VERSION:   tar-preview-telemetry.23
      GITPOD_INSTALLATION_PLATFORM:  local-preview
      SERVER_URL:                    http://server.default.svc.cluster.local:9000/
    Mounts:                          <none>
```
- `kots` on a `k3s` cluster
```
Environment:
      GITPOD_INSTALLATION_VERSION:   tar-preview-telemetry.23
      GITPOD_INSTALLATION_PLATFORM:  local-preview
      SERVER_URL:                    http://server.default.svc.cluster.local:9000/
    Mounts:                          <none>
```

Also, Verified that our telemetry store is receving the new property as expected!

corneliusludmann (approved):
Looks good! :rocket: Thanks for taking my comments into consideration! :orange_heart:

MrSimonEmms (approved):

Issue #10922, opened 2022-06-27
Use attributionId when listing workspace instances for usage


easyCZ:

easyCZ:
Implemented.

PR #10927, opened 2022-06-27
[usage] Add usageAttributionID to WorkspaceInstance model (in go)

easyCZ:
## Description

Adds model definition to consume usage attribution id

## Related Issue(s)

* Part of https://github.com/gitpod-io/gitpod/issues/10922

## How to test

Unit tests

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage/pkg/db/workspace_instance.go
+++ b/components/usage/pkg/db/workspace_instance.go
@@ -11,6 +11,7 @@ import (
 	"github.com/google/uuid"
 	"gorm.io/datatypes"
 	"gorm.io/gorm"
+	"strings"
 	"time"
 )
 
@@ -23,6 +24,7 @@ type WorkspaceInstance struct {
 	IdeURL             string         `gorm:"column:ideUrl;type:varchar;size:255;" json:"ideUrl"`
 	WorkspaceBaseImage string         `gorm:"column:workspaceBaseImage;type:varchar;size:255;" json:"workspaceBaseImage"`
 	WorkspaceImage     string         `gorm:"column:workspaceImage;type:varchar;size:255;" json:"workspaceImage"`
+	UsageAttributionID AttributionID  `gorm:"column:usageAttributionId;type:varchar;size:60;" json:"usageAttributionId"`
 
 	CreationTime VarcharTime `gorm:"column:creationTime;type:varchar;size:255;" json:"creationTime"`
 	StartedTime  VarcharTime `gorm:"column:startedTime;type:varchar;size:255;" json:"startedTime"`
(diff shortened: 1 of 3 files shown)


roboquat:

<details>

</details>

components/usage/pkg/db/workspace_instance.go
```diff
@@ -87,3 +89,20 @@ func ListWorkspaceInstancesInRange(ctx context.Context, conn *gorm.DB, from, to
...
+type AttributionID string
+
+func (a AttributionID) Values() (entity string, identifier string) {
+	tokens := strings.Split(string(a), ":")
+	if len(tokens) != 2 {
+		return "", ""
+	}
+
+	return tokens[0], tokens[1]
+}
```
jankeromnes, line 108: How does this `Values` function work, i.e. who calls it? 👀

components/usage/pkg/db/workspace_instance.go
```diff
@@ -87,3 +89,20 @@ func ListWorkspaceInstancesInRange(ctx context.Context, conn *gorm.DB, from, to
 
 	return instances, nil
 }
+
+func NewAttributionID(entity, identifier string) AttributionID {
```
jankeromnes, line 93: Nit: I guess Go can infer the type here, but it looks a bit inconsistent without the explicit type:

```suggestion
func NewAttributionID(entity string, identifier string) AttributionID {
```

jankeromnes:
Many thanks for this PR!

In general, I'm very much in favor of multiple smaller PRs. However (and maybe this is just me adjusting to this new paradigm), I still struggle with reviewing code that isn't used in any meaningful way.

Thus, I took a peek at your other PR: https://github.com/gitpod-io/gitpod/pull/10938 👀

- There, I saw how the `attribution.Values()` function is being used 👍 (helps me build confidence that it's indeed what we wwanat)
- However, I also saw there that the `NewAttributionID(entity string, identifier string)` is not the most useful method, proven by the fact that you've refactored it into `db.NewUserAttributionID(identifier string)` and `db.NewTeamAttributionID(identifier string)` -- would it make sense to hoist the better implementation in this PR?

components/usage/pkg/db/workspace_instance.go
easyCZ: This is not inference, this is standard style of give. When there are multiple arguments of the same type, you only need to specify the type on the last one.

easyCZ:
>  I still struggle with reviewing code that isn't used in any meaningful way.
Here, the constructor and the destructor are mostly for completness and symmetry. Agree that ideally the destructor wasn't used in this PR (and perhaps should've been used in tests) but it felt incomplete without it.

> However, I also saw there that the NewAttributionID(entity string, identifier string) is not the most useful method, proven by the fact that you've refactored it into db.NewUserAttributionID(identifier string) and db.NewTeamAttributionID(identifier string) -- would it make sense to hoist the better implementation in this PR?
This would also suffer from the same problem as above. I'd be creating a constructor for `NewTeamAttributionID` without having usage for it in this PR, so that felt even more of an over-kill to include in this one.
On the other hand, once I had practice usage, it longer made sense to keep the `NewAttributionID` function public so it got refectored. This mostly shows that you raise PRs as you go, rather than figuring out what the perfect final version is (and back-slicing it then).

easyCZ:
I've pulled in the two methods and added some tests for using the Value to showcase it better. Hopefully this addresses the above comment.

components/usage/pkg/db/workspace_instance.go
jankeromnes: Thanks! Today I learned. (I guess a few lines below, the type can also be factored out.)

components/usage/pkg/db/workspace_instance.go
jankeromnes: Resolved below. Thanks!

components/usage/pkg/db/workspace_instance.go
easyCZ: Interesting. This works (compiles) but I've not really noticed this pattern in other people's code before. Perhaps it's because it's less common to return two values of the same type from a function

PR #10933, opened 2022-06-27
[kots]: change to not install via Helm

MrSimonEmms:
## Description

A [regression by Replicated](https://github.com/replicated-collab/replicated-gitpod/issues/27) introduced a breaking change to how to deploy KOTS releases that included a Helm chart between v1.72.0 and v1.72.2. The [change](https://github.com/replicatedhq/kots/commit/77690e5cad4a1abe0bb63063129564059777d8f4) only affects manifests that use native Helm but the legacy implementation inside KOTS.

## How to test

Install with KOTS v1.72.0 and then with KOTS latest

```sh
# Install KOTS v1.72.0
curl https://kots.io/install/1.72.0 | bash
```

```sh
# Install KOTS latest
curl https://kots.io/install | bash
```

## Release Notes

```release-note
[kots]: change to not install via Helm
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/kots/manifests/gitpod-log-collector.yaml
+++ b/install/kots/manifests/gitpod-log-collector.yaml
@@ -10,7 +10,7 @@ spec:
     name: fluent-bit
     chartVersion: 0.20.2
   helmVersion: v3
-  useHelmInstall: true
+  useHelmInstall: false
   weight: 10
   values:
     fluent-bit:


nandajavarma (approved):
LGTM :shipit:

PR #10938, opened 2022-06-27
[usage] Use attribution ID to reduce DB queries for usage report

easyCZ:
## Description

Updates usage controller flow to do the following:

1. Find all Workspace Instances (with time bounds) which have an `usageAttributionId` set.
2. Groups these (in-code) by Attribution ID to create the report

We can't use SQL level group-by because it can only be used with summary functions (like count). The secondary reason is that the Billing controller (which runs after the usage controller) needs to given the full list of WorkspaceInstances to exclude any WorkspaceInstances which it may have already billed for (in another invoice, for example). For this, it has to get the raw data-set and can't receive an aggregation.

## Related Issue(s)

* Part of https://github.com/gitpod-io/gitpod/issues/10922

## How to test

Unit tests
Can be run against staging by port forwarding

## Release Notes

```release-note
NONE
```

## Documentation

NONE

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage/pkg/controller/reconciler.go
+++ b/components/usage/pkg/controller/reconciler.go
@@ -45,12 +45,6 @@ type UsageReconcileStatus struct {
 
 	WorkspaceInstances        int
 	InvalidWorkspaceInstances int
-
-	Workspaces int
-
-	Teams int
-
-	Report []TeamUsage
 }
 
 func (u *UsageReconciler) Reconcile() error {
@@ -60,7 +54,7 @@ func (u *UsageReconciler) Reconcile() error {
 	startOfCurrentMonth := time.Date(now.Year(), now.Month(), 1, 0, 0, 0, 0, time.UTC)
 	startOfNextMonth := startOfCurrentMonth.AddDate(0, 1, 0)
 
-	status, err := u.ReconcileTimeRange(ctx, startOfCurrentMonth, startOfNextMonth)
+	status, report, err := u.ReconcileTimeRange(ctx, startOfCurrentMonth, startOfNextMonth)
 	if err != nil {
 		return err
 	}
@@ -75,7 +69,7 @@ func (u *UsageReconciler) Reconcile() error {
 	defer f.Close()
 
 	enc := json.NewEncoder(f)
-	err = enc.Encode(status.Report)
+	err = enc.Encode(report)
 	if err != nil {
 		return fmt.Errorf("failed to marshal report to JSON: %w", err)
 	}
@@ -89,7 +83,7 @@ func (u *UsageReconciler) Reconcile() error {
 	return nil
 }
 
-func (u *UsageReconciler) ReconcileTimeRange(ctx context.Context, from, to time.Time) (*UsageReconcileStatus, error) {
+func (u *UsageReconciler) ReconcileTimeRange(ctx context.Context, from, to time.Time) (*UsageReconcileStatus, UsageReport, error) {
 	now := u.nowFunc().UTC()
 	log.Infof("Gathering usage data from %s to %s", from, to)
 	status := &UsageReconcileStatus{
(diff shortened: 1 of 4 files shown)


roboquat:

<details>

</details>

components/usage/pkg/db/workspace_instance.go
```diff
@@ -79,6 +79,7 @@ func ListWorkspaceInstancesInRange(ctx context.Context, conn *gorm.DB, from, to
 		).
 		Where("creationTime < ?", TimeToISO8601(to)).
 		Where("startedTime != ?", "").
+		Where("usageAttributionId != ?", "").
```
jankeromnes, line 82: How will we want to handle potentially un-attributed instances?

components/usage/pkg/controller/reconciler.go
```diff
@@ -302,20 +200,20 @@ func trimStartStopTime(instances []db.WorkspaceInstance, maximumStart, minimumSt
...
+	for _, instance := range instances {
+		if _, ok := result[instance.UsageAttributionID]; !ok {
+			result[instance.UsageAttributionID] = []db.WorkspaceInstance{}
+		}
+
+		result[instance.UsageAttributionID] = append(result[instance.UsageAttributionID], instance)
+	}
+
+	return result
+}
```
jankeromnes, line 219: Are you sure it makes sense performance-wise to re-implement a SQL "group by" in Go?

I wonder if maybe this should be part of the query. Relatedly, have you had a chance to chat with @geropl about the query's performance, and how to efficiently index and shard it? 👀

easyCZ: You can only use group-by with summary functions (like count). That's because a group-by takes multiple rows, and turns them into a single row by aggregating them.

In practice, we need the full dataset anyway because the billing controller needs to do another pass to drop WSI which have already been billed. 

Happy to go into more details on this.

components/usage/pkg/db/workspace_instance.go
easyCZ: Given the `usageAttributionId` should always be present (the premise of your PR). The check here mostly ensures we only run on "new" data which does have the attribution set

easyCZ:
I've moved this to Draft as I've added a bunch of fixes for flaky tests which I need to pull into a separate PR. The main logic is still reviewable, but I'll move it back to review once cleaned up.

easyCZ:
Fixed up and rebased, ready for review again.

geropl:
Starting to review now...

components/usage/pkg/controller/reconciler.go
```diff
@@ -108,129 +102,38 @@ func (u *UsageReconciler) ReconcileTimeRange(ctx context.Context, from, to time.
...
+	attributedUsage := map[string]int64{}
 
-	membershipsByUserID := map[uuid.UUID]db.TeamMembership{}
-	for _, membership := range memberships {
-		// User can belong to multiple teams. For now, we're choosing the membership at random.
-		membershipsByUserID[membership.UserID] = membership
-	}
+	for attribution, instances := range u {
+		entity, id := attribution.Values()
+		if entity != db.AttributionEntity_Team {
```
geropl, line 119: nit: A comment that we handle this in the future would be golden :+1:

components/usage/pkg/controller/reconciler.go
```diff
@@ -108,129 +102,38 @@ func (u *UsageReconciler) ReconcileTimeRange(ctx context.Context, from, to time.
...
-	// Finally, join the datasets
-	// Because we iterate over memberships, and not workspaces, we're in effect ignoring Workspaces which are not in a team.
-	// This is intended as we focus on Team usage for now.
-	var teamsWithWorkspaces []teamWithWorkspaces
-	for userID, membership := range membershipsByUserID {
-		teamsWithWorkspaces = append(teamsWithWorkspaces, teamWithWorkspaces{
-			TeamID:     membership.TeamID,
-			Workspaces: workspacesByOwnerID[userID],
-		})
+		attributedUsage[id] = int64(runtime)
```
geropl, line 128: This feels dangerous. Why not make attributedUsage accumulate in `uint64`?

easyCZ: Happy to update to unit64 everyone (but Stripe only accepts int64, not unit64). But the reason for using it here is to limit changes to the existing Usage Report reconcile logic with stripe in this PR. That happens here https://github.com/gitpod-io/gitpod/pull/10938/files#diff-b2499b7086d5733f081dcce586e0ff0e77206dd5d8fd6ede638e3fc8f284c798L25-L32 and it currently has int64 as the interface.
I'll update these in a follow-up PR (minus the Stripe API call which will have to be int64) if that's acceptable.

geropl: Ok, fine with this for now.

PR #10939, opened 2022-06-27
Add tests for GuardedResources: Workspace, WorkspaceInstance and WorkspaceLog (1/2)

geropl:
## Description

This PR adds tests for workspace-like GuardedResources: `Workspace`, `WorkspaceInstance` and `WorkspaceLog`.
It's a precursor for a follow-up PR.

## Related Issue(s)

Context: https://github.com/gitpod-io/gitpod/pull/10696

## How to test
 - `cd components/server && yarn test`

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

PR #10940, opened 2022-06-27
Introduce GuardedPrebuild to be used in #10696

geropl:
## Description

Introduces `GuardedPrebuild` to guard operations on Prebuilds.
This will be used in a follow-up PR (#10696).

## Related Issue(s)

Based on: #10939
Context: #10696

## How to test
 - `(components/server && yarn test)`

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/server/src/auth/resource-access.ts
+++ b/components/server/src/auth/resource-access.ts
@@ -217,7 +217,11 @@ export class OwnerResourceGuard implements ResourceAccessGuard {
                         return resource.members.some((m) => m.userId === this.userId && m.role === "owner");
                 }
             case "workspaceLog":
-                return resource.subject.ownerId === this.userId;
+                // Owners may do everything, team members can "get"
+                return (
+                    resource.subject.ownerId === this.userId ||
+                    (operation === "get" && !!resource.teamMembers?.some((m) => m.userId === this.userId))
+                );
             case "prebuild":
                 // Owners may do everything, team members can "get"
                 return (


components/server/src/auth/resource-access.ts
```diff
@@ -209,6 +218,8 @@ export class OwnerResourceGuard implements ResourceAccessGuard {
                 }
             case "workspaceLog":
                 return resource.subject.ownerId === this.userId;
+            case "prebuild":
+                return resource.workspace.ownerId === this.userId;
```
jankeromnes, line 222: Shouldn't this be accessible by the team too? (E.g. `resource.members.some(...)` as above)

jankeromnes: Ah sorry, I see that this is handled via the `RepositoryAccessGuard` below. Makes sense. 💡

jankeromnes (approved):
Looks good and makes sense to me, many thanks!

I guess my biggest question is -- why do we need this extra resource guard? Isn't it enough to guard on access to the `workspace` (or on the `workspaceLog`)? Do we expect access models for workspaces, logs, and prebuilds to differ significantly in the future?

/hold

components/server/src/auth/resource-access.spec.ts
```diff
@@ -833,6 +833,76 @@ class TestResourceAccess {
...
+                teamRole: undefined,
+                expectation: true,
+            },
+            {
+                name: "prebuild get other",
+                resourceKind: "prebuild",
+                workspaceType: "prebuild",
+                isOwner: false,
+                teamRole: undefined,
+                expectation: false,
```
AlexTugarev, line 851: @geropl, out of curiosity, what the expected effect of this change for a non-team-member to open a workspace on a configured repository? e.g. is this rule also enforced on workspace startup?

components/server/src/auth/resource-access.ts
geropl: Hm. No, you are right! It has the same effect (because on the happy path, if you are member of a team, you are likely to have read access to the repo.
I felt into the cargo-cult trap: We should have the same for workspaceLog. Will fix. :+1:

components/server/src/auth/resource-access.spec.ts
geropl: This PR does just introduce the GuardedPrebuild, but it is not used yet.

A follow up PR will introduce prebuild APIs (get/findByWorkspaceId), and it felt wrong to re-/ab-use workspace or workspaceinstance for it.

components/server/src/auth/resource-access.ts
geropl: Fixed in https://github.com/gitpod-io/gitpod/pull/10940/commits/0a37cf9f329b23d168d11fda8c80af479840ae0f

geropl:
> Do we expect access models for workspaces, logs, and prebuilds to differ significantly in the future?

@jankeromnes Not really, but I'd like to have it explicit. That way it's easier to either a) migrate to other implementations, or b) remove in the future. :lotus_position:

Issue #10921, opened 2022-06-27
Distinguish between pinned workspaces and recent workspaces in dashboard


jldec:
It would be helpful to show an indicator for pinned workspaces in the workspaces list.
![Screenshot 2022-06-27 at 12-33-40 Workspaces — Gitpod(1)](https://user-images.githubusercontent.com/39482679/175897571-64ac3786-9107-4423-97d5-ff0b4b47480f.png)

gtsiolis:
I just love how this is now hiding 44 workspaces for you, @jldec! ✨

gtsiolis:
Closing this in favor of UX-47.

PR #10947, opened 2022-06-27
Clarify meaning of "Inactive Workspaces"

atduarte:
## Description

Since a few weeks ago, we group and hide "Inactive Workspaces" but don't provide any explanation on what those are. Instead of a description, we only warn that they will be deleted in 14 days.

I propose to change that description to clarify what an "inactive workspace" is while keeping the information about the garbage collection.

| Before  | After |
| ------------- | ------------- |
| image | image  |

## Related Issue(s)

N/A

## How to test

If needed, running the preview.

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/dashboard/src/workspaces/Workspaces.tsx
+++ b/components/dashboard/src/workspaces/Workspaces.tsx
@@ -162,8 +162,8 @@ export default function () {
                                                 </span>
                                             </div>
                                             <div className="text-sm flex-auto">
-                                                Workspaces that were stopped more than 24 hours ago. Note that these
-                                                will be automatically deleted after 14 days.{" "}
+                                                Workspaces that have been stopped for more than 24 hours. Inactive 
+                                                workspaces are automatically deleted after 14 days.{" "}
                                                 <a
                                                     className="gp-link"
                                                     href="https://www.gitpod.io/docs/life-of-workspace/#garbage-collection"
--- a/components/dashboard/src/workspaces/Workspaces.tsx
+++ b/components/dashboard/src/workspaces/Workspaces.tsx
@@ -162,8 +162,8 @@ export default function () {
                                                 </span>
                                             </div>
                                             <div className="text-sm flex-auto">
-                                                Unpinned workspaces that have been inactive for more than 14 days will
-                                                be automatically deleted.{" "}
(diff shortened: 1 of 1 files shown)


components/dashboard/src/workspaces/Workspaces.tsx
```diff
@@ -162,8 +162,8 @@ export default function () {
                                                 </span>
                                             </div>
                                             <div className="text-sm flex-auto">
-                                                Unpinned workspaces that have been inactive for more than 14 days will
-                                                be automatically deleted.{" "}
+                                                Workspaces that were stopped more than 24 hours ago. Note that these
+                                                will be automatically deleted after 14 days.{" "}
```
gtsiolis, line 166: **suggestion**: In the spirit of using plain language with simpler copy and messaging that also favor brevity, we could rephrase this like the following, unless there are any strong feelings about keeping it as is. [WDYT](https://en.wiktionary.org/wiki/WDYT)?

> Workspaces that have been stopped for more than 24 hours. Inactive workspaces are automatically deleted after 14 days. 

**thought**: I wouldn't mind keeping the _Unpinned_ prefix to be more explicit, but we can remove it out for now and improve later in https://github.com/gitpod-io/gitpod/issues/10921, as changes here are already a good MVC improvement. Cc @jldec 

> Unpinned workspaces that have been stopped ... 

To clarify more and in case it helps, here's how I read this copy.

copy

gtsiolis (approved):
UX LGTM! Thanks, @atduarte! 🏀 

This will also need an approval from someone from the `@gitpod-io/engineering-webapp` team.

Thanks for also including BEFORE / AFTER screenshots. Left one minor suggestion but feel free to also ignore the suggestion as this is already a good MVC improvement.

components/dashboard/src/workspaces/Workspaces.tsx
gtsiolis: Alternative suggestion to avoid line break on expansion, etc.

> Unpinned workspaces that are stopped for more than 24 hours, automatically deleted after 14 days.

geropl:
@jldec Tagging you for review :+1:

geropl (approved):
Looks good to me! :+1:

PR #10962, opened 2022-06-28
[server] abort running prebuilds on same branch

### LINKED ISSUE #6569: Pushing new commits to a branch, does not stop a prebuild from a previous commit of the same branch.
### Bug description

I think a previously running prebuild should be cancelled if new commits are pushed to the same branch. 

I noticed it after I pushed changes to a branch, and realized I forgot to make a small change.
image

### Steps to reproduce

Create long init task in `.gitpod.yml`.
After pushing a commit that will trigger the prebuild, push another commit to the same branch, and go to Prebuilds page on your gitpod.io account.

### Workspace affected

_No response_

### Expected behavior

Seems like waste of resources to keep the previous prebuild running.
Not sure if this might cause other issues (like an old prebuild finish running after the latest prebuild, and perhaps override things?)

### Example repository

_No response_

### Anything else?

_No response_

svenefftinge:
## Description
Cancels all running and queued prebuilds for the same branch when a new commit is pushed. 
Adds a project setting to opt-out of this behavior.

Screenshot 2022-06-28 at 12 54 36

## Related Issue(s)

Fixes #6569

## How to test
- Add a project with a long prebuild task
- push multipüle times to the same branch
- see how the previous prebuilds get cancelled.
- push to a different branch and verify that running prebuilds on the other branch are not getting cancelled.
- disable the project setting and verify that the cancellation doesn't happen

## Release Notes

```release-note
Outdated prebuilds (i.e. new commits are pushed on a branch) are automatically canceled. This behavior can be disabled in the project's settings.
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/dashboard/src/projects/ProjectSettings.tsx
+++ b/components/dashboard/src/projects/ProjectSettings.tsx
@@ -14,7 +14,6 @@ import { PageWithSubMenu } from "../components/PageWithSubMenu";
 import PillLabel from "../components/PillLabel";
 import { ProjectContext } from "./project-context";
 import { getExperimentsClient } from "./../experiments/client";
-import { DeepPartial } from "@gitpod/gitpod-protocol/lib/util/deep-partial";
 
 export function getProjectSettingsMenu(project?: Project, team?: Team) {
     const teamOrUserSlug = !!team ? "t/" + team.slug : "projects";
@@ -56,19 +55,14 @@ export default function () {
     const { teams } = useContext(TeamsContext);
     const team = getCurrentTeam(location, teams);
 
-    const [isLoading, setIsLoading] = useState<boolean>(true);
-    const [isIncrementalPrebuildsEnabled, setIsIncrementalPrebuildsEnabled] = useState<boolean>(false);
-    const [isPersistentVolumeClaimEnabled, setIsPersistentVolumeClaimEnabled] = useState<boolean>(false);
     const [isShowPersistentVolumeClaim, setIsShowPersistentVolumeClaim] = useState<boolean>(false);
+    const [projectSettings, setProjectSettings] = useState<ProjectSettings>({});
 
     useEffect(() => {
         if (!project) {
             return;
         }
-        setIsLoading(false);
-        setIsIncrementalPrebuildsEnabled(!!project.settings?.useIncrementalPrebuilds);
-        setIsPersistentVolumeClaimEnabled(!!project.settings?.usePersistentVolumeClaim);
-
+        setProjectSettings({ ...project.settings });
         (async () => {
             const showPersistentVolumeClaim = await getExperimentsClient().getValueAsync(
                 "persistent_volume_claim",
@@ -84,49 +78,34 @@ export default function () {
         })();
     }, [project, team, teams]);
 
-    const updateProjectPartial = (settings: DeepPartial<ProjectSettings | undefined>) => {
+    const updateProjectSettings = () => {
         if (!project) {
             return;
         }
-
-        return getGitpodService().server.updateProjectPartial({ id: project.id, settings });
+        setProjectSettings({
+            ...projectSettings,
+        });
+        return getGitpodService().server.updateProjectPartial({ id: project.id, settings: projectSettings });
     };
 
     const toggleIncrementalPrebuilds = async () => {
-        if (!project) {
-            return;
-        }
-        setIsLoading(true);
-        try {
-            await updateProjectPartial({
-                useIncrementalPrebuilds: !isIncrementalPrebuildsEnabled,
-                usePersistentVolumeClaim: isPersistentVolumeClaimEnabled,
-            });
-            setIsIncrementalPrebuildsEnabled(!isIncrementalPrebuildsEnabled);
-        } finally {
-            setIsLoading(false);
-        }
+        projectSettings.useIncrementalPrebuilds = !projectSettings.useIncrementalPrebuilds;
+        updateProjectSettings();
+    };
+
+    const toggleCancelOutdatedPrebuilds = async () => {
+        projectSettings.keepOutdatedPrebuildsRunning = !projectSettings.keepOutdatedPrebuildsRunning;
+        updateProjectSettings();
     };
 
     const togglePersistentVolumeClaim = async () => {
-        if (!project) {
-            return;
-        }
-        setIsLoading(true);
-        try {
-            await updateProjectPartial({
-                useIncrementalPrebuilds: isIncrementalPrebuildsEnabled,
-                usePersistentVolumeClaim: !isPersistentVolumeClaimEnabled,
-            });
-            setIsPersistentVolumeClaimEnabled(!isPersistentVolumeClaimEnabled);
-        } finally {
-            setIsLoading(false);
-        }
+        projectSettings.usePersistentVolumeClaim = !projectSettings.usePersistentVolumeClaim;
+        updateProjectSettings();
     };
 
     return (
         <ProjectSettingsPage project={project}>
-            <h3>Incremental Prebuilds</h3>
+            <h3>Prebuilds</h3>
             <CheckBox
                 title={
                     <span>
@@ -146,9 +125,14 @@ export default function () {
                         </a>
                     </span>
                 }
-                checked={isIncrementalPrebuildsEnabled}
-                disabled={isLoading}
+                checked={!!projectSettings.useIncrementalPrebuilds}
                 onChange={toggleIncrementalPrebuilds}
+            />
+            <CheckBox
+                title={<span>Cancel Prebuilds on Outdated Commits </span>}
+                desc={<span>Cancels all pending and running prebuilds on the branch when a new commit is pushed.</span>}
+                checked={!projectSettings.keepOutdatedPrebuildsRunning}
+                onChange={toggleCancelOutdatedPrebuilds}
             />
             <br></br>
             <h3>Persistent Volume Claim</h3>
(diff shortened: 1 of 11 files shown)


jankeromnes:
Many thanks @svenefftinge for continuously improving & optimizing which prebuilds we should be running or not! 💯

I'll give this Pull Request a look. 👀

components/dashboard/src/projects/ProjectSettings.tsx
```diff
@@ -56,19 +55,14 @@ export default function () {
...
 
     useEffect(() => {
         if (!project) {
             return;
         }
-        setIsLoading(false);
-        setIsIncrementalPrebuildsEnabled(!!project.settings?.useIncrementalPrebuilds);
-        setIsPersistentVolumeClaimEnabled(!!project.settings?.usePersistentVolumeClaim);
-
+        setProjectSettings({ ...project.settings });
```
jankeromnes, line 65: What is the benefit of having an extra `projectSettings` state here, given that we already have a `project`? 🤔

Couldn't we use `project.settings` as the source of truth? (And modify that + call `updateProjectSettings` when things are toggled?)

svenefftinge: Because we are using react state and need to update the object for react to rerender. react doesn't check deep references

components/dashboard/src/projects/ProjectSettings.tsx
```diff
@@ -56,19 +55,14 @@ export default function () {
     const { teams } = useContext(TeamsContext);
     const team = getCurrentTeam(location, teams);
 
-    const [isLoading, setIsLoading] = useState<boolean>(true);
```
jankeromnes, line 59: ❤️ Many thanks for removing a `isLoading` boolean! (which I might have added 🙈)

I agree that maintaining a `isLoading` boolean is generally bad practice, and creates many opportunities for bugs. (And when a loading state is required, it's always better to explicit wait on what still needs to be loaded, instead of trying to code this into a `isLoading` boolean.) 👍

components/dashboard/src/projects/ProjectSettings.tsx
```diff
@@ -146,10 +125,15 @@ export default function () {
                         </a>
                     </span>
                 }
-                checked={isIncrementalPrebuildsEnabled}
-                disabled={isLoading}
```
jankeromnes, line 150: I wonder if we should keep the disabled state of checkboxes, e.g. for when `project` isn't loaded yet? 🤔 This could avoid having interactive checkboxes in the wrong state (e.g. everything seems disabled while we're loading the actual project settings.)

Not super important though. 👍

components/dashboard/src/projects/ProjectSettings.tsx
```diff
@@ -146,10 +125,15 @@ export default function () {
...
+                checked={!!projectSettings.useIncrementalPrebuilds}
                 onChange={toggleIncrementalPrebuilds}
             />
+            <CheckBox
+                title={<span>Cancel Prebuilds on Outdated Commits </span>}
+                desc={<span>Cancels all pending and running prebuilds on the branch when a new commit is pushed.</span>}
```
jankeromnes, line 133: In general, I think that this feature is super valuable and needed, and I'm happy that we're taking a first step towards it. 💯

Without having read the code much further or tested it, one initial thought that occurs is that, if one of your branches (e.g. `main`) takes a long time to build, and sees very frequent commits (e.g. during working hours), we might be always-cancelling running prebuilds without ever finishing one, for long stretches of time (e.g. maybe we only successfully finish running a prebuild over lunch break or at the end of the day).

components/dashboard/src/projects/ProjectSettings.tsx
jankeromnes: Ah, that's a good point, thanks.

So I guess getting `setProject` from the ProjectContext, and calling `setProject(project)` after modifying the `project.settings` wouldn't actually update the state, right? 🤔

In any case, the current implementation is fine by me assuming it works 😊 (will test it soon) so this is not a blocker.

components/server/ee/src/prebuilds/prebuild-manager.ts
```diff
@@ -58,6 +58,22 @@ export class PrebuildManager {
...
 
+    async abortPrebuildsForBranch(ctx: TraceContext, project: Project, user: User, branch: string): Promise<void> {
+        const span = TraceContext.startSpan("abortPrebuildsForBranch", ctx);
+        const prebuilds = await this.workspaceDB
+            .trace({ span })
+            .findActivePrebuiltWorkspacesByBranch(project.id, branch);
+        const results: Promise<any>[] = [];
+        for (const prebuild of prebuilds) {
+            for (const instance of prebuild.instances) {
+                results.push(this.workspaceStarter.stopWorkspaceInstance({ span }, instance.id, instance.region));
```
jankeromnes, line 69: I wonder what happens for instances which are not in a "stoppable" state, e.g. in image build? Will this fail for some not-yet-stoppable instances, or will this wait until it reaches a stoppable state? (I'm guessing the former.)

components/server/src/workspace/workspace-starter.ts
```diff
@@ -349,6 +351,20 @@ export class WorkspaceStarter {
         }
     }
 
+    public async stopWorkspaceInstance(
```
jankeromnes, line 354: Good idea moving this here. 👍

components/server/src/workspace/workspace-starter.ts
```diff
@@ -349,6 +351,20 @@ export class WorkspaceStarter {
...
+        ctx: TraceContext,
+        instanceId: string,
+        instanceRegion: string,
+        policy?: StopWorkspacePolicy,
+    ): Promise<void> {
+        const req = new StopWorkspaceRequest();
+        req.setId(instanceId);
+        req.setPolicy(policy || StopWorkspacePolicy.NORMALLY);
+
+        const client = await this.clientProvider.get(instanceRegion);
```
jankeromnes, line 364: If I remember correctly, some instances might not have a region yet (e.g. still in image build or so?) and this could cause `client` to be incorrect or undefined. Is this still the case? 🤔 I guess this needs to be tested.

jankeromnes (commented):
Thanks again! Had an initial pass through the code, and it looks good to me, modulo a few (mostly optional) questions.

Will now give the PR a go, and try to find some edge cases (e.g. image builds?)

jankeromnes:
Err, will actually test this PR later today (in ~1.5 hours) as I have upcoming meetings. Please feel free to push any changes if you want to (i.e. no need to necessarily keep the preview env alive until then).

geropl:
@svenefftinge Do you have any plans on announcing this change? This will certainly affect some heavy Gitpod users. :thinking:

svenefftinge:
> This will certainly affect some heavy Gitpod users. 🤔

How do you think it will affect users? In a negative way you mean?

svenefftinge:
The only path where users might be surprised is seeing prebuilds "CANCELED". We should probably tell them why

components/dashboard/src/projects/ProjectSettings.tsx
svenefftinge: I'd rather work on making loading a single project faster if that's really an issue

components/dashboard/src/projects/ProjectSettings.tsx
svenefftinge: Yes, it does not update the state.

jankeromnes:
>  I'm adding the status descriptions as a tooltip for the items in the list and also add the information about new commits being pushed on the branch:

Good idea! 👍 I wonder if it also makes sense to tell users that they can click on the prebuild in order to know the exact cancellation reason. (It should always be visible in the full error message below the prebuild logs.)

svenefftinge:
> Good idea! 👍 I wonder if it also makes sense to tell users that they can click on the prebuild in order to know the exact cancellation reason. (It should always be visible in the full error message below the prebuild logs.)

It is the same message

gtsiolis:
[FWIW](https://en.wiktionary.org/wiki/FWIW), auto-cancelling redundant builds is a common practice in CI systems[[1](https://docs.gitlab.com/ee/ci/pipelines/settings.html#auto-cancel-redundant-pipelines)][[2](https://circleci.com/docs/2.0/skip-build#auto-cancelling)][...].

In the future, we could introduce a preference option in project settings but going with a sensible default[[3](https://en.wikipedia.org/wiki/Convention_over_configuration)] sounds better.

svenefftinge:
> In the future, we could introduce a preference option in project settings but going with a sensible default[[3](https://en.wikipedia.org/wiki/Convention_over_configuration)] sounds better.

This PR already comes with that preference (default is ☑️ ):
Screenshot 2022-06-29 at 16 36 43

gtsiolis (approved):
>This PR already comes with that preference (default is ☑️ ) 

Perfect (Perfect)! ✨

Left one comment for updating the copy used in settings. Feel free to ignore or adjust the suggestion. 🏓

components/server/ee/src/prebuilds/prebuild-manager.ts
```diff
@@ -58,6 +58,22 @@ export class PrebuildManager {
...
+            }
+            prebuild.prebuild.state = "aborted";
```
jankeromnes, line 71: Please also specify the cancellation reason by setting a helpful `prebuild.error` value, similar to the [rate-limit error](https://github.com/gitpod-io/gitpod/blob/7d005012a82778e5b916bf494a2c64bd461fec7d/components/server/ee/src/prebuilds/prebuild-manager.ts#L188-L191):

```suggestion
            prebuild.prebuild.state = "aborted";
            prebuild.error = "A newer commit was pushed to the same branch.";
```

jankeromnes (approved):
Many thanks! In addition to the code looking good to me earlier, I've tested this PR and it works as advertised. ✨ 🎯

For what it's worth, I also tested the behavior with image builds (by pushing a custom Dockerile and then several commits on top) -- the Prebuilds actually get cancelled, even though the logs continue to receive new updates. 😅 I guess that's fine. 👍

Screenshot 2022-06-29 at 19 12 24

Adding a hold because I'd love the cancellation reason to be explicitly added as a `prebuild.error`, or in case someone else would like to review.

/hold

svenefftinge:
> I'd love the cancellation reason to be explicitly added as a prebuild.error,

Will do in a follow up PR.

PR #10964, opened 2022-06-28
[werft] Update Stripe product price IDs

jankeromnes:
## Description

The "graduated price with free tier in Stripe" experiment yielded interesting data. However, it doesn't align with our planned team pricing model, so I've adjusted the Stripe product/prices accordingly.

We can re-introduce the graduated price with free tier in Stripe when we implement individual user billing (either as two new Stripe prices under the same product, or as a separate product with 2 prices -- I personally think that having a single product with 4 prices is preferable).

## Related Issue(s)

Fixes #

## How to test

1. Should still build

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/.werft/jobs/build/payment/stripe-configmap.yaml
+++ b/.werft/jobs/build/payment/stripe-configmap.yaml
@@ -4,10 +4,10 @@ metadata:
   name: stripe-config
   namespace: ${NAMESPACE}
 data:
-  config: >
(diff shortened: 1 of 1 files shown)


geropl (approved):
Thx @jankeromnes ! :+1:

PR #10971, opened 2022-06-28
[usage] Fix flakes by deleting records created by each test, not deleting all

easyCZ:
Golang will run tests from different files (and packages) concurrently (in seperate go-routines). This presents a problem for our test cleanup logic which tries to execute `DELETE * FROM <table>` to cleanup a test table.

Follows guidance from https://kevin.burke.dev/kevin/fast-parallel-database-tests/

When tests are run, we use a helper from `dbtest.CreateWorkspaces` or `dbtest.CreateWorkspaceInstances` to create records which have a test cleanup hook. This hook deletes created records by ID instead of deleting all. This is necessary because when tests run concurrently, we don't want to drop records possibly created by other tests.

There still remain a problem for tests which try to do ListAll without any constraints but we don't have any of these in our usage component, yet.

## Related Issue(s)

Fixes #

## How to test

Unit tests

## Release Notes

```release-note
NONE
```

## Documentation

NONE

## Werft options:

- [ ] /werft with-preview


roboquat:

<details>

</details>

geropl:
@easyCZ Got a couple of these while testing:
```
    --- FAIL: TestListWorkspacesByID/not_found_id_returns_emtpy_results (0.00s)
        workspace_test.go:129: 
                Error Trace:    conn.go:73
                                                        workspace_test.go:129
                Error:          Received unexpected error:
                                Error 1040: Too many connections
                Test:           TestListWorkspacesByID/not_found_id_returns_emtpy_results
                Messages:       Failed to establish connection to DB. In a workspace, run `leeway build components/usage:init-testdb` once to bootstrap the DB.
```

easyCZ:
> @easyCZ Got a couple of these while testing:
> 
> ```
>     --- FAIL: TestListWorkspacesByID/not_found_id_returns_emtpy_results (0.00s)
>         workspace_test.go:129: 
>                 Error Trace:    conn.go:73
>                                                         workspace_test.go:129
>                 Error:          Received unexpected error:
>                                 Error 1040: Too many connections
>                 Test:           TestListWorkspacesByID/not_found_id_returns_emtpy_results
>                 Messages:       Failed to establish connection to DB. In a workspace, run `leeway build components/usage:init-testdb` once to bootstrap the DB.
> ```

Interesting, I wasn't able to reproduce these. I guess that happens when the tests create too many connection against the DB. Perhaps a connection pool on the app side can solve this.

Let me try with a larger sample (1k runs) to see if I can reproduce, to test a fix.

easyCZ:
Now the only ones which do not like to be run in parallel are the ones where we create raw objects, due to seed values of random generation of workspace name. I'm gonna drop these tests as they are no longer really relevant, they existed to validate we can read from the existing DB without issues but we've long been running in that state that the data in the unit tests are no longer up-to-date either.

components/usage/pkg/db/varchar_test.go
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package db_test
```
easyCZ, line 5: These tests were moved to a seperate file because they required using the `dbtest.ConnectForTest` but were previously in the `db` package. This would create a circular import. Nothing changed about the test.

easyCZ:
There remains an issue with the test `TestListWorkspaceInstancesInRange` which uses a fixed query arguments and will likely fail when that test is run concurrently (that test repeated concurrently). While annoying, there isn't a good fix that wouldn't blow up this PR. 
I'm willing to accept that concurrent runs of this test will fail for now and land the PR as is as it is already an improvement for the general level of flakiness.

Issue #10832, opened 2022-06-22
[preview envirionments] run with cgroup v2


kylos101:
## Is your feature request related to a problem? Please describe

`gen49` is switching to cgroup v2; preview environments should have parity with production.

## Describe the behaviour you'd like

Upgrade preview environments to use cgroup v2.

## Describe alternatives you've considered

n/a

## Additional context

The image has been updated in other issues [like so](https://github.com/gitpod-io/gitpod/issues/9684).

The latest image which supports cgroup v2 [is internal](https://github.com/gitpod-io/gitpod-packer-gcp-image/releases/tag/gitpod-k3s-202206202141). This uses K3s 1.24, and uses systemd as the cgroupDriver.

## How to test
```
gitpod /workspace/template-python-flask (main) $ ls /sys/fs/cgroup/cgroup.controllers
ls: cannot access '/sys/fs/cgroup/cgroup.controllers': No such file or directory
```

The above file should exist (thanks, @Furisto !), but is what we currently get in preview environments.

meysholdt:
reopening, because we needed to revert the PR here: https://github.com/gitpod-io/gitpod/pull/10989

reasons: 
* https://gitpod.slack.com/archives/C032A46PWR0/p1656448245444129
* https://github.com/gitpod-io/gitpod/pull/10987#issuecomment-1169389161

ArthurSens:
Now that we reverted it, how do we proceed here? Could you point us to a new image that you'd like to have in preview environments @kylos101 ?

kylos101:
Hey @ArthurSens , could you try a branch with [this image](https://github.com/gitpod-io/ops/blob/main/deploy/workspace/common.sh#L33), which is what we're using "now"? Also, is it possible to test that we can start workspaces using the image, prior to merging back to main?

cc: @aledbf

ArthurSens:
> Also, is it possible to test that we can start workspaces using the image, prior to merging back to main?

That is an interesting question, I believe I did test it... I tried opening workspaces and they started 🤔 

Not sure how to tell if my tests were correct or not though. I'll try to hold the merge until someone from workspace team made the tests as well, is that ok?

kylos101:
@ArthurSens adding you as an assignee on this because of the open PR. Is this okay? I'm not sure if this fits into your current week plan for Platform.

ArthurSens:
[...]

PR #10987, opened 2022-06-28
Update preview VM image

ArthurSens:

## Description

Updates VM image to [this one](https://github.com/gitpod-io/gitpod-packer-gcp-image/releases/tag/gitpod-k3s-202206202141)

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/10832

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/.werft/vm/manifests.ts
+++ b/.werft/vm/manifests.ts
@@ -35,7 +35,7 @@ kind: VirtualMachine
 metadata:
   namespace: ${namespace}
   annotations:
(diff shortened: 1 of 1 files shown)


ArthurSens:
I believe we need to run this and while making sure the preview runs with the updated image?

meysholdt (approved):
Code change LGTM and the preview env runs as well.

meysholdt:
CC @Furisto

Furisto:
That does not look right. The cgroup filesystem should be mounted at /sys/fs/cgroup. If it is mounted at /sys/fs/cgroup/unified then it points to a hybrid configuration (which in practice is basically cgroup v1).

PR #10993, opened 2022-06-29
[supervisor] Check if git repository is shallow before running --unshallow

aledbf:
## Description

https://git-scm.com/docs/git-rev-parse

## Related Issue(s)

xref: https://cloudlogging.app.goo.gl/iaodaYmM7HfTDpwq8

## How to test
- Start a workspace
- Stop the workspace and wait until is not running
- Reopen the workspace
- Check workspace logs and check it does not contain:
  `fatal: --unshallow on a complete repository does not make sense`
 
## Release Notes
```release-note
NONE
```

### DIFF
--- a/components/supervisor/pkg/supervisor/supervisor.go
+++ b/components/supervisor/pkg/supervisor/supervisor.go
@@ -357,6 +357,10 @@ func Run(options ...RunOption) {
 					log.Debugf("unshallow of local repository took %v", time.Since(start))
 				}()
 
+				if !isShallowRepository(repoRoot, childProcEnvvars) {
+					return
+				}
+
 				cmd := runAsGitpodUser(exec.Command("git", "fetch", "--unshallow", "--tags"))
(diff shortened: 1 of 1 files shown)


jeanp413:
Thanks @aledbf, but do you know why we get this error? As far as I know we always do a shallow clone and then unshallow so does this mean we are unshallowing twice (or more times) :thinking: ?

aledbf:
> Thanks @aledbf, but do you know why we get this error? As far as I know we always do a shallow clone and then unshallow so does this mean we are unshallowing twice (or more times) thinking ?

I think this happens when you restart a workspace. I can see `jsonPayload.source="from-backup"` (which makes sense)

jeanp413:
From the logs I see this error started happening on may 29th, maybe the root cause is somewhere else, will take a quick look just to be sure

jeanp413 (approved):

PR #10996, opened 2022-06-29
[usage] Expose metrics for reconciles started and duration

easyCZ:
## Description

So that we can setup dashboards and alerts on these

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Documentation

NONE

## Werft options:

- [x] /werft with-preview


jankeromnes (approved):
Many thanks! This works as advertised. ✨ 

Screenshot 2022-06-29 at 12 39 49

I can't comment on whether the code change is the best approach, as I don't have any experience with Go component metrics reporting best practices, but other than that this looks good to me. 👍

PR #11001, opened 2022-06-29
[installer]: order the custom envvars

### LINKED ISSUE #11000: The Gitpod Installer randomises ordering of environment variables for components
### Bug description

I think we have a regression in the Installer that makes the environment variables order for components random ([PR](https://github.com/gitpod-io/gitpod/pull/10906)). By converting the env vars to a map and then back to a slice I believe we will randomise ordering as the map doesn't maintain any kind of stable ordering. This causes drift in ArgoCD similar to the issue we had with ports previously ([PR](https://github.com/gitpod-io/gitpod/pull/10344))

### Steps to reproduce

We're seeing this in or SaaS deployments that are managed by ArgoCD.

### Workspace affected

_No response_

### Expected behavior

The YAML produced by the Installer should be the same given the same configuration.

### Example repository

_No response_

### Anything else?

_No response_

MrSimonEmms:
## Description

Changes how the customize envvars orders it's output. It keeps the existing envvars in the order that's given and the customized envvars are ordered in terms of first one received. This is to avoid unnecessary deploys in a GitOps setup.

Overrides the "fix failing test" commit in #10836 by @Furisto - it achieves the same effect, but does so in the application code rather than the test

## Related Issue(s)

Fixes #11000

## How to test

The unit tests provides the test coverage. @mads-hartmann should check that the order is as expected

## Release Notes

```release-note
[installer]: order the custom envvars
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/installer/pkg/common/customize_test.go
+++ b/install/installer/pkg/common/customize_test.go
@@ -5,7 +5,6 @@ package common_test
 
 import (
 	"reflect"
-	"sort"
 	"testing"
 
 	"github.com/gitpod-io/gitpod/installer/pkg/common"
@@ -390,7 +389,6 @@ func TestCustomizeEnvvar(t *testing.T) {
 		ExistingEnnvars []corev1.EnvVar
 		Expect          []corev1.EnvVar
 	}{
-
 		{
 			Name:          "no customization",
 			Customization: nil,
@@ -489,6 +487,10 @@ func TestCustomizeEnvvar(t *testing.T) {
 			},
 			ExistingEnnvars: []corev1.EnvVar{
 				{
+					Name:  "key4",
+					Value: "value",
+				},
+				{
 					Name:  "key3",
 					Value: "original",
 				},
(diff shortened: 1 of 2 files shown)


install/installer/pkg/common/customize_test.go
```diff
@@ -521,13 +527,6 @@ func TestCustomizeEnvvar(t *testing.T) {
 
 			result := common.CustomizeEnvvar(ctx, testCase.Component, testCase.ExistingEnnvars)
 
-			sort.Slice(result, func(a, b int) bool {
```
MrSimonEmms, line 524: Removing this ensures that the order generated by the function matches the expectation

corneliusludmann:
@MrSimonEmms: When it's true that the “map doesn't maintain any kind of stable ordering” as stated in the issue, then we would have the same issue with e.g. [annotations](https://github.com/gitpod-io/gitpod/blob/e2eced05879caff398fc292125d4e0990d501564/install/installer/pkg/common/customize.go#L56) (and maybe in several other places as well). I guess, in practice, we can assume that the order, in general, is the same in a map (even if that is not guaranteed), can't we?

cc @mads-hartmann

corneliusludmann (approved):
Besides my more general comment above, the implementation looks good so far.

MrSimonEmms:
@corneliusludmann yes, you are correct. However, as these are already `map[string]string` in the Kubernetes spec, the order is not specified (nor has it ever been).

My guess is, knowing this, ArgoCD makes allowances for `labels` and `annotations` in it's diff

mads-hartmann:
@MrSimonEmms I think your guess is right. ArgoCD uses their [GitOps Engine](https://github.com/argoproj/gitops-engine) which I believe tries to be clever about these things.

PR #10768, opened 2022-06-20
[jb] enable vmoptions config in .gitpod.yml

yaohui-wyh:
## Description

Add `vmoptions` in gitpod schema (`.gitpod.yml`).

```yaml
jetbrains:
  intellij:
    vmoptions: "-Xmx4g"
  goland:
    vmoptions: "-Xmx3g"
```

- This is a follow-up PR with https://github.com/gitpod-io/gitpod/pull/10175
- Schema description text is borrowed from https://www.jetbrains.com/help/idea/remote-development-a.html#rd_add_vm_options
- VMOptions specified by project level `.gitpod.yml` have higher priority than user's environment variables. Suppose we have a workspace with:
    - project level `.gitpod.yml`: `jetbrains.intellij.vmoptions = "-Xmx4g"`
    - Gitpod user preferences `INTELLIJ_VMOPTIONS` EnvVar set to `-Xmx3g -Didea.is.internal=true` 
    - the final effective VMOptions will be `-Didea.is.internal=true -Xmx4g`

## Related Issue(s)

Remedy for https://github.com/gitpod-io/gitpod/issues/8704

## How to test

> Build & test `status` from a Gitpod workspace

- Start Gitpod workspace from this PR: https://gitpod.io/#https://github.com/gitpod-io/gitpod/pull/10768
- **Modify project's `.gitpod.yml` and insert a `jetbrains.intellij.vmoptions` field with value `"-Xmx4g"`**
- Launch a new Terminal and build the `status` binary: `cd /workspace/gitpod/components/ide/jetbrains/image/status && go build`. Replace the `status` binary: `rm /ide-desktop/status; cp status /ide-desktop/status`
- Kill the `status` process: `ps aux | grep status; kill -9 <pid>`
- (`supervisor` restart `status` immediately) Check JetBrains IDE server VMOptions: `jps -mlv | grep intellij | grep -i xmx`

Expected result:

- IDE Server should be running with `-Xmx4g`

## Release Notes

```release-note

- Allow customize VMOptions for JetBrains backend server, by setting "vmoptions" in .gitpod.yml

```

PR #11002, opened 2022-06-29
[jb] check maxHeapSize of IDE server and guide users to set vmoptions in .gitpod.yml

### LINKED ISSUE #10924: [jb] add a warning if runtime Xmx is different from configured
An idea is to detect a user change to the actual max heap size in the backend plugin and suggest a user to put in .gitpod.yml to persist. So the flow would be like:
- a user starts using with default Xmx
- experience perf issues and follow notifications to upgrade Xmx of the running instance
- after JB backend restart there will be a warning on .gitpod.yml that Xmx of the running instance was changed, but it should be applied to .gitpod.yml to be persistent

@yaohui-wyh Maybe you can have a look how we can validate .gitpod.yml? Basically add a warning on .gitpod.yml or on a project if the file is missing if configured Xmx at runtime is different from default one. Later we can add a quick fix to add it into .gitpod.yml

yaohui-wyh:
## Description

Quote https://github.com/gitpod-io/gitpod/issues/10924

> An idea is to detect a user change to the actual max heap size in the backend plugin and suggest a user to put in .gitpod.yml to persist. So the flow would be like:
>
> - a user starts using with default Xmx
> - experience perf issues and follow notifications to upgrade Xmx of the running instance
> - after JB backend restart there will be a warning on .gitpod.yml that Xmx of the running instance was changed, but it should be applied to .gitpod.yml to be persistent
>
> Also check out https://github.com/gitpod-io/gitpod/pull/10768 for more background info

This PR implements Xmx config checks in JetBrains `gitpod-remote` plugin:
- Register a background activity (GitpodStartupActivity) for handling postStart events
- Add YAML plugin dependency for parsing & navigating PSI elements in `.gitpod.yml`
- Show warning notification when the runtime Xmx is different from the configured Xmx
  - Add edit/create-if-missing `.gitpod.yml` actions, and navigate caret to `vmoptions` YAML key if exists
  - Add help action for browsing gitpod.yml references
- Set `jvmArgs` for the `runIde` task (only affect local sandbox IDE instance with the developed plugin), which can be modified for local testing purpose

## Related Issue(s)

Resolve #10924

## How to test

Test locally:

- Run `runIde` task from IntelliJ plugin, and modify `jvmArgs = listOf("-Xmx4g")` to simulate the user's change to the -Xmx of JetBrains IDE server inside a Gitpod workspace. When sandbox IDE instance is started, a warning Notification should popup, indicating current runtime maxHeapSize is different with default/configured vmoptions from `.gitpod.yml`. Click on "Edit .gitpod.yml" would open the file in the editor
  - edit .gitpod.yml action

- If there is no `.gitpod.yml` in the project root, the notification popup would suggest the user to create a `.gitpod.yml` with generated content (with the current runtime Xmx configured)
  -  create .gitpod.yml action
  - .gitpod.yml created

## Release Notes

```release-note
- Detect user changes to the max heap size of IDE server when using JetBrains Gateway, and suggest a user to put in .gitpod.yml to persist.
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodStartupActivity.kt
+++ b/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodStartupActivity.kt
@@ -1,0 +1,129 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package io.gitpod.jetbrains.remote
+
+import com.intellij.diagnostic.VMOptions
+import com.intellij.ide.BrowserUtil
+import com.intellij.ide.actions.OpenFileAction
+import com.intellij.notification.NotificationAction
+import com.intellij.notification.NotificationGroupManager
+import com.intellij.notification.NotificationType
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.startup.StartupActivity
+import com.intellij.openapi.util.BuildNumber
+import com.intellij.openapi.vfs.LocalFileSystem
+import com.intellij.openapi.vfs.VfsUtil
+import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.psi.PsiFileFactory
+import com.intellij.psi.PsiManager
+import com.intellij.util.application
+import io.gitpod.jetbrains.remote.utils.GitpodConfig
+import io.gitpod.jetbrains.remote.utils.GitpodConfig.YamlKey.jetbrains
+import io.gitpod.jetbrains.remote.utils.GitpodConfig.YamlKey.vmOptions
+import io.gitpod.jetbrains.remote.utils.GitpodConfig.defaultXmxOption
+import io.gitpod.jetbrains.remote.utils.GitpodConfig.gitpodYamlFile
+import io.gitpod.jetbrains.remote.utils.GitpodConfig.gitpodYamlReferenceLink
+import org.jetbrains.yaml.YAMLFileType
+import org.jetbrains.yaml.YAMLUtil
+import org.jetbrains.yaml.psi.YAMLFile
+import org.jetbrains.yaml.psi.YAMLKeyValue
+import java.nio.file.Paths
+
+class GitpodStartupActivity : StartupActivity.Background {
+
+    override fun runActivity(project: Project) {
+        application.invokeLater { checkMemoryOption(project) }
+    }
+
+    private fun checkMemoryOption(project: Project) {
+        val productCode = BuildNumber.currentVersion().productCode
+        val productName = GitpodConfig.getJetBrainsProductName(productCode) ?: return
+        // it's ok for .gitpod.yml to not exist
+        var vmOptions = emptyList<String>()
+        getGitpodYamlVirtualFile(project)?.let {
+            val vmOptionsKeyValue = getGitpodYamlVMOptionsPsiElement(project, it, productName)
+            vmOptions = vmOptionsKeyValue?.valueText?.split("\\s".toRegex()) ?: emptyList()
+        }
+        // if there is no -Xmx option from .gitpod.yml, compare runtime maxHeapSize with default value
+        var xmxVmOptions = vmOptions.filter { it.startsWith("-Xmx") }
+        if (xmxVmOptions.isEmpty()) {
+            xmxVmOptions = listOf(defaultXmxOption)
+        }
+        // the rightmost -Xmx option is the one to take effect (after deduplication)
+        val finalXmx = xmxVmOptions.last()
+        // shift right 20 (xmxInBytes >> 20) to convert to MiB
+        val finalXmxValueInMiB = VMOptions.parseMemoryOption(finalXmx.substringAfter("-Xmx")).shr(20)
+        val runtimeXmxValueInMiB = Runtime.getRuntime().maxMemory().shr(20)
+        if (finalXmxValueInMiB != runtimeXmxValueInMiB) {
+            showEditVMOptionsNotification(project, runtimeXmxValueInMiB, finalXmxValueInMiB, productName)
+        }
+    }
+
+    private fun getGitpodYamlVirtualFile(project: Project): VirtualFile? {
+        val basePath = project.basePath ?: return null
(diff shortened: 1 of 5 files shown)


components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodStartupActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+    private fun getGitpodYamlVMOptionsPsiElement(
+        project: Project,
+        virtualFile: VirtualFile,
+        productName: String
+    ): YAMLKeyValue? {
+        val psiFile = PsiManager.getInstance(project).findFile(virtualFile) as? YAMLFile ?: return null
+        return YAMLUtil.getQualifiedKeyInFile(psiFile, jetbrains, productName, vmOptions)
+    }
+
+    private fun showEditVMOptionsNotification(project: Project, runtimeXmxMiB: Long, configXmxMiB: Long, productName: String) {
```
akosyakov, line 78: I don't think we need notificaiotns. They are too loud. Can we use inspections instead? https://plugins.jetbrains.com/docs/intellij/code-inspections.html

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodStartupActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+import io.gitpod.jetbrains.remote.utils.GitpodConfig.defaultXmxOption
+import io.gitpod.jetbrains.remote.utils.GitpodConfig.gitpodYamlFile
+import io.gitpod.jetbrains.remote.utils.GitpodConfig.gitpodYamlReferenceLink
+import org.jetbrains.yaml.YAMLFileType
+import org.jetbrains.yaml.YAMLUtil
+import org.jetbrains.yaml.psi.YAMLFile
+import org.jetbrains.yaml.psi.YAMLKeyValue
+import java.nio.file.Paths
+
+class GitpodStartupActivity : StartupActivity.Background {
```
akosyakov, line 34: We should perform validation whenever content of .gitpod.yml is changing even if dirty. Not only once. Should not it be rather local inspection?

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodStartupActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+        notification.addAction(helpAction)
+        notification.notify(project)
+    }
+
+    private fun editGitpodYamlAction(project: Project, gitpodYaml: VirtualFile, productName: String): NotificationAction {
+        return NotificationAction.createSimple("Edit .gitpod.yml") {
+            OpenFileAction.openFile(gitpodYaml, project)
+            val vmOptionsKeyValue = getGitpodYamlVMOptionsPsiElement(project, gitpodYaml, productName)
+            // navigate caret to "vmoptions" if exist
+            vmOptionsKeyValue?.navigate(true)
```
akosyakov, line 107: Cannot we automatically update it, instead of asking a user to edit?

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodStartupActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+    ): YAMLKeyValue? {
+        val psiFile = PsiManager.getInstance(project).findFile(virtualFile) as? YAMLFile ?: return null
+        return YAMLUtil.getQualifiedKeyInFile(psiFile, jetbrains, productName, vmOptions)
+    }
+
+    private fun showEditVMOptionsNotification(project: Project, runtimeXmxMiB: Long, configXmxMiB: Long, productName: String) {
+        val notificationGroup = NotificationGroupManager.getInstance().getNotificationGroup("Gitpod Notifications")
+        val title = "Gitpod memory settings"
+        val message = """
+            |Current maxHeapSize <code>-Xmx${runtimeXmxMiB}m</code> is not matched to configured <code>-Xmx${configXmxMiB}m</code>.<br/>
```
akosyakov, line 82: There should be a warning inspection:
`IDE's max heap size (-Xmx) is {runtimeValue}M, but not configured in .gitpod.yml` - if it is missing in .gitpod.yml
`IDE's max heap size (-Xmx) is {runtimeValue}M, but {configuredXmx}M in .gitpod.yml` - if it is defined in .gitpod.yml

There should be quick fixes:
- Add -Xmx{runtimeValue}M to .gitpod.yml (both cases) - It should automatically create .gitpod.yml if necessary and sections or update vmoptions string
- Apply default -Xmx{defaultValue} (if not in .gitpod.yml) - it should apply default value using VMOptions api
- Apply -Xmx{defaultValue} configured in .gitpod.yml (if not in .gitpod.yml) - it should apply configured value using VMOptions api

akosyakov (commented):
I think warning should not be notifications, but dynamically update inspections. And there should be automatic quick fixes.

yaohui-wyh:
Hi @akosyakov, thanks for the quick reply. About the notification part:

I do understand the Notification is a bit noisy and interrupting, so I was trying to make it a post-start activity once only. Inspection on `.gitpod.yml` might not fit into the memory settings scenario with:

- Suppose a user experience perf issues and upgrade Xmx of the running instance (e.g. from the [control center window](https://www.jetbrains.com/help/pycharm/remote-development-a.html#control_center) of JBClient). 
- After JB backend restart the IDE would reopen the recent files of the project and the user continued on their work, and is unlikely to open (maybe also unaware of) the `.gitpod.yml` until they stop the workspace.
- Due to the immutable nature of vmoptions, when they open the same project in a new Gitpod workspace they will encounter the perf issue again.
- Inspections on `.gitpod.yml` cannot help users familiarize themselves with how to persist the vmoptions settings.

JetBrains IDEs also use Warning Notification & Dialog for [low memory warning & settings](https://www.jetbrains.com/help/idea/increasing-memory-heap.html):

image
image

yaohui-wyh:
>  think having good diagnostics and automatic quick fixes which goes in both directions should be a starting point. If we see that it does not help we can be more noisy. ...

Agree with this. I will try adding local inspection and quick fix intention on .gitpod.yml

Issue #10983, opened 2022-06-28
Proposed documentation


MrSimonEmms:
## Proposed changes

Support for proxy service types is required for enterprise client

1. Remove `experimental.webapp.proxy.serviceType` from the config object
2. Add:

```yaml
components:
  proxy:
    service:
      serviceType: ClusterIP # allowed values are "ClusterIP", "LoadBalancer", "NodePort", "ExternalName" consistent with https://pkg.go.dev/k8s.io/api/core/v1#ServiceType
```

## KOTS

This will be a [select_one](https://docs.replicated.com/reference/custom-resource-config#select_one) under the "advanced" section.

## Rationale

We don't want to be supporting the same thing from multiple configuration parameters. That would require checks between the two and/or a hierarchy of importance - we should have a single source of truth.

Having a components section makes this extensible for the future. Unlike the `experimental` section which lives under our conventions, this new convention is ordered as per the resources - again, this is done with future extensibility in mind.

## Documentation

> This is written from the KOTS-only perspective

All web traffic enters the application through the `Proxy` component. By default, this is a `LoadBalancer` type which will expose the application through the cloud provider's load balancer.

Under certain circumstances, this may be undesirable. Under the "Advanced" section, you may select a different service type - all [Kubernetes service types](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types) are supported, except `ExternalName`. If you are using anything other than `LoadBalancer`, you are responsible for configuring your network to route traffic through to the `proxy` service.

MrSimonEmms:
@lucasvaltl very good question. My thought is that this would be a new page under the "advanced" docs on the website. We can talk about the component customization - this will be the first of (presumably) many things

But, as you said, you're the one who's got more of a global ownership of the docs so I'll defer to your thoughts

MrSimonEmms:
Update:

Added "except `ExternalName`" to the service types we support. Reason for this is that there's additional configuration required which will be awkward to add in and there's no driver for this yet.

This only removes it from KOTS - it's still a valid option in the Installer

PR #11006, opened 2022-06-29
[installer]: promote proxy service type from experimental

### LINKED ISSUE #11008: Move installer config value out of experimental to the main section.

### LINKED ISSUE #11009: Add a config option in KOTS.

### LINKED ISSUE #11025: Move loadbalancer-specific annotations to loadbalancer-only
See https://github.com/gitpod-io/gitpod/issues/10725#issuecomment-1160838536

### LINKED ISSUE #11033: Make sure it will be tested (could be: unit tests, part of an explicit dimension in the automated self-hosted tests, part of a reference architecture that will be tested by the automated tests, etc).

MrSimonEmms:
> Anyone relying upon this config in the `experimental` section should move to the `components`. See description for new format

## Description

Deprecated the proxy's service type in `experimental` and promotes it to the primary configuration. Implements the configuration as per #10983.

Includes the KOTS config section in a separate "Components" group - the reasoning behind this is that there may be additional component-specific configuration in the future. This should make it easier to manage/visualise. This is only displayed if the "enable additional options" is enabled.

Adds the ability to deprecate config parameters.

I've also renamed "additional options" to "advanced options"

image

## Related Issue(s)

Fixes #11008
Fixes #11009
Fixes #11025
Fixes #11033 

## Release Notes

```release-note
[installer]: promote proxy service type from experimental
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/installer/pkg/components/proxy/service.go
+++ b/install/installer/pkg/components/proxy/service.go
@@ -81,10 +81,10 @@ func service(ctx *common.RenderContext) ([]runtime.Object, error) {
 		service.Spec.Type = serviceType
 		if serviceType == corev1.ServiceTypeLoadBalancer {
 			service.Spec.LoadBalancerIP = loadBalancerIP
-		}
 
-		service.Annotations["external-dns.alpha.kubernetes.io/hostname"] = fmt.Sprintf("%s,*.%s,*.ws.%s", ctx.Config.Domain, ctx.Config.Domain, ctx.Config.Domain)
-		service.Annotations["cloud.google.com/neg"] = `{"exposed_ports": {"80":{},"443": {}}}`
+			service.Annotations["external-dns.alpha.kubernetes.io/hostname"] = fmt.Sprintf("%s,*.%s,*.ws.%s", ctx.Config.Domain, ctx.Config.Domain, ctx.Config.Domain)
+			service.Annotations["cloud.google.com/neg"] = `{"exposed_ports": {"80":{},"443": {}}}`
+		}
 
 		for k, v := range annotations {
 			service.Annotations[k] = v
--- a/install/installer/pkg/components/proxy/service.go
+++ b/install/installer/pkg/components/proxy/service.go
@@ -22,7 +22,6 @@ var allowedServiceTypes = map[corev1.ServiceType]struct{}{
 }
 
 func service(ctx *common.RenderContext) ([]runtime.Object, error) {
-	serviceType := corev1.ServiceTypeLoadBalancer
 
 	loadBalancerIP := ""
 	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
@@ -30,16 +29,20 @@ func service(ctx *common.RenderContext) ([]runtime.Object, error) {
 			if cfg.WebApp.ProxyConfig.StaticIP != "" {
 				loadBalancerIP = cfg.WebApp.ProxyConfig.StaticIP
 			}
-			st := cfg.WebApp.ProxyConfig.ServiceType
-			if st != nil {
-				_, allowed := allowedServiceTypes[corev1.ServiceType(*st)]
-				if allowed {
-					serviceType = *st
-				}
-			}
 		}
 		return nil
 	})
+
+	serviceType := corev1.ServiceTypeLoadBalancer
+	if ctx.Config.Components != nil && ctx.Config.Components.Proxy != nil && ctx.Config.Components.Proxy.Service != nil {
+		st := ctx.Config.Components.Proxy.Service.ServiceType
+		if st != nil {
+			_, allowed := allowedServiceTypes[corev1.ServiceType(*st)]
+			if allowed {
+				serviceType = *st
+			}
+		}
+	}
 
 	var annotations map[string]string
 	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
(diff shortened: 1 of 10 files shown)


install/installer/pkg/config/v1/config.go
```diff
@@ -113,6 +113,8 @@ type Config struct {
 
 	Customization *[]Customization `json:"customization,omitempty"`
 
+	Components *Components `json:"components,omitempty"`
```
geropl, line 116: I don't have a strong opinion here, but: did you think about moving this under an `advanced` section?
The idea is to mirror the KOTS config, including the "reduce service" that we offer for changes that go into this section: "only touch when advised or you know what you're doing".

MrSimonEmms: I didn't, although I think there's probably quite a few things which are in the config already that are "advanced" topics that we've not done so with - customization and the resource limits jump to mind.

My personal feeling is that we should keep the config as objective to the section, rather than differentiating between basic, advanced and experimental sections.

But, like you, I don't have a strong opinion here either. I'll let our friend @corneliusludmann have the final say

install/installer/pkg/config/v1/experimental/experimental.go
```diff
@@ -156,9 +156,8 @@ type BlockedRepository struct {
 }
 
 type ProxyConfig struct {
-	StaticIP           string              `json:"staticIP"`
-	ServiceAnnotations map[string]string   `json:"serviceAnnotations"`
-	ServiceType        *corev1.ServiceType `json:"serviceType,omitempty" validate:"omitempty,service_config_type"`
```
geropl, line 161: @MrSimonEmms What's the proper way to deprecate fields in the installer?

This is "experimental", but in prod we do rely on this particular field, for instance.
Is there a way to ensure the installer errors when seeing this field in the config? That would be awesome, because we could be sure to never break prod. :hugs:

geropl: > so I guess whatever we decide here is "the process".

Sorry to delay here. But I'm poking at this because we're in the process to switch to Continuous Deployment, and then every merge to main would have immediate impact on gitpod.io. 

> I put a message in the [Product Engineering](https://gitpod.slack.com/archives/C01KGM9AW4W/p1656581818013179) channel this morning and shared in Platform Team channel.

Thank you, that's what brought me here :wink: 

> If someone is using the experimental flag, this will break as Golang won't know what do to with that YAML key. It won't go through to the actual prod cluster because will have broken before it gets that far.

Hm, I'm not sure. To deploy/render for prod, we use the `--strict-parse=false` ([code](https://github.com/gitpod-io/gitpod/blob/4d9100ef8224aad4e5c96237486ab2a6c46cdcab/install/installer/cmd/root.go#L32)) parameter for not having to sync on config changes. But for that to "work" we have to make sure that non-backwards compatible changes are "breaking".

Can we maybe keep this flag around as **deprecated** and  `ServiceType        *corev1.ServiceType `json:"serviceType,omitempty"`, and error if it's set? Just to establish that habit?
(not saying we should do that for all experimental fields, but this one is kind of touchy. But also I wonder how many of your current customers do use this field in their POC? :thinking: )

geropl: > put the experimental feature back in

@MrSimonEmms  That sounds good!

> schedule this feature for removal. I propose the 2022.12.0 release,

As this is `experimental`, and you are sure that we do not have external clients for this, I would be happy to do it even quicker, as soon as we have our internal client (gitpod.io) moved. We will share an update on Slack! :+1:

corneliusludmann (changes requested):
I have a comment regarding overwriting the experimental config with the new config together with the KOTS UI change. I think that could break the installation when users have a config patch that set the `serviceProxy` in the experimental section because this will be overwritten by the new proxy config in any case (solely by updating Gitpod with KOTS). I think we need at least a warning, better a property way to handle this.

The remaining comments are my 2 cents only. Feel free to disagree and ignore.

install/installer/pkg/components/proxy/service.go
```diff
@@ -41,6 +41,16 @@ func service(ctx *common.RenderContext) ([]runtime.Object, error) {
 		return nil
 	})
 
+	if ctx.Config.Components != nil && ctx.Config.Components.Proxy != nil && ctx.Config.Components.Proxy.Service != nil {
+		st := ctx.Config.Components.Proxy.Service.ServiceType
+		if st != nil {
+			_, allowed := allowedServiceTypes[corev1.ServiceType(*st)]
+			if allowed {
+				serviceType = *st
```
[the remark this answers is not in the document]
geropl: I agree with the sentiment, especially if we replace "experimental" with "deprecated" (and do it similarly for all future deprecations)! :orange_heart: 

> During the validation, if only experimental is set, print warning and move the experimental value over to the regular config.

That's a neat trick to cope with this! :100:

MrSimonEmms:
Ok, I've updated this so it has @corneliusludmann's [elegant proposal](https://github.com/gitpod-io/gitpod/pull/11006#discussion_r913230709) of dealing with the experimental values.

I've also changed the KOTS config so that the `LoadBalancer` option doesn't actually set any value. This is so that it doesn't silently break the workflow for anyone upgrading who is using the `experimental` version already.

I want to refactor how the KOTS config is generated with the config patch file. Deprecations will come in that

corneliusludmann (approved):
Love it. :star_struck: 

Thanks for taking my suggestions into consideration. :pray:

geropl (approved):
Nice pattern, thank you @MrSimonEmms !

MrSimonEmms:
Team effort - couldn't have done it without you two

Eiffel Tower

Issue #10985, opened 2022-06-28
Differentiate usage of different workspace classes and types (prebuilds)


jldec:
Users viewing a usage report should recognize where that usage happened. This means differentiating between usage for different [workspace classes](https://github.com/gitpod-io/gitpod/issues/8261) (which are metered at different rates), as well as separating usage of workspaces like prebuilds, from regular workspaces.

easyCZ:
In https://github.com/gitpod-io/gitpod/pull/11043, we've added configuration to price different classes differently. The remaining work is to expose the WorkspaceType in the results. To do this, we'll need to join WorkspaceInstances with Workspaces.

We'll need to something like the following query to get the:
```
SELECT wsi.id, wsi.creationTime, wsi.startedTime, wsi.stoppedTime, wsi.usageAttributionId, ws.projectId, ws.type
    FROM d_b_workspace_instance AS wsi
    LEFT JOIN d_b_workspace AS ws
      ON wsi.workspaceId = ws.id
    LIMIT 100;
```

easyCZ:
This is now complete.

PR #11032, opened 2022-06-30
[usage] Refactor credit calculation into a WorkspacePricer

easyCZ:
## Description

As part of https://github.com/gitpod-io/gitpod/issues/10985, we need to compute Credit cost differently for different WorkspaceClasses. This refactor keeps the existing behavour the same, but allows us to extend it to specify cost per class differently.

In subsequent PRs, the cost will be made configurable through config (eventually it may move to DB as well).

## Related Issue(s)

Fixes #

## How to test

Unit tests

## Release Notes

```release-note
NONE
```

## Documentation

NONE

## Werft options:

- [ ] /werft with-preview


geropl (approved):
Tests are green, and code looks good! :pray:

Issue #11030, opened 2022-06-30
Move blocked repository configuration into the database and build admin UI to work with it


andrew-farries:
Gitpod allows for repositories to be blocked, so that workspaces can't be started with those contexts. Support for blocking repositories was added to `components/server` in #9762 where the implementation used server config to provide the list of blocked repos.

We'd like to move blocked repository configuration out of server config and into the application database for a few reasons:
* We can more easily build blocked repo configuration into the Gitpod admin UI (and/or `gpctl`?)
* Having this configurable in the admin UI makes on-call easier than having to change `server` config at runtime with `kubectl`.
* The ability to block repositories may be useful for self hosted users.

## Description

We don't need to use a feature flag for this; we can support blocked repository config in the database and in server config at the same time. We can then move the repositories that reside in server config into the database and delete the former.

## Tasks

* [ ] Add new table to the database to hold blocked repository data.
* [ ] Read new table at workspace startup and prevent workspace start/block user as necessary.
* [ ] Add CRUD API surface to work with blocked repositories.
* [ ] Build CRUD admin UI to work with blocked repositories.
* [ ] Move existing blocked repositories from server config into database.
* [ ] Remove server configuration of blocked repositories and code that reads it.

geropl:
@andrew-farries Do you mind if I take over here during my on-call week? :question:

andrew-farries:
Please do 😄

PR #11036, opened 2022-06-30
[server] Read blocked repositories from database

andrew-farries:
## Description

Add a `d_b_blocked_repository` table to the application database and use it at workspace startup to check whether a workspace should be blocked. Also checks if the user should be blocked for attempting to start such a workspace and blocks the user as necessary.

## Related Issue(s)
Part of https://github.com/gitpod-io/gitpod/issues/11030

## How to test

* Get a kube context for the preview environment for this branch:

  ```bash
  previewctl install-context
  ```

* Port forward to the database (username and password is in the `server` environment):

  ```bash
  ```

* Using `mysql` client or `mycli` (needs `brew install mycli` first) add an entry to the `d_b_blocked_repository` table.

  ```sql
  insert into d_b_blocked_repository (`urlRegexp`, `blockUser`, `deleted`) values ("github.com/<your github username>", true, false)
  ```

* Attempt to start a workspace from a github repository under your user.

* Observe that the workspace fails to start.

* Try to open another workspace and see that you are now blocked.

* Unblock yourself  by updating your entry in the `d_b_users` table:

  ```sql
  update d_b_user set blocked = 0 where id=<id field value for the row for your username>'
  ```

* Experiment by inserting/deleting other combinations of `urlRegexp` and `blockUser` in the `d_b_blocked_repository` table.

<insert image>

## Release Notes

```release-note
NONE
```

## Werft options

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-db/src/blocked-repository-db.spec.db.ts
+++ b/components/gitpod-db/src/blocked-repository-db.spec.db.ts
@@ -1,0 +1,67 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import * as chai from "chai";
+const expect = chai.expect;
+import { suite, test, timeout } from "mocha-typescript";
+
+import { testContainer } from "./test-container";
+import { TypeORMBlockedRepositoryDBImpl } from "./typeorm/blocked-repository-db-impl";
+import { TypeORM } from "./typeorm/typeorm";
+import { DBBlockedRepository } from "./typeorm/entity/db-blocked-repository";
+
+@suite
+class BlockedRepositoryDBSpec {
+    blockedRepositoryDb = testContainer.get<TypeORMBlockedRepositoryDBImpl>(TypeORMBlockedRepositoryDBImpl);
+
+    async before() {
+        await this.wipeRepo();
+    }
+
+    async after() {
+        await this.wipeRepo();
+    }
+
+    @test(timeout(10000))
+    public async checkRepositoryIsBlocked() {
+        const typeorm = testContainer.get<TypeORM>(TypeORM);
+        const manager = await typeorm.getConnection();
+        manager.getRepository(DBBlockedRepository).insert({
+            urlRegexp: "github.com/bob/.*",
+            blockUser: true,
+            deleted: false,
+        });
+
+        const blockedRepository = await this.blockedRepositoryDb.isRepositoryBlocked("github.com/bob/some-repo");
+
+        expect(blockedRepository).not.undefined;
+        expect(blockedRepository?.urlRegexp).to.eq("github.com/bob/.*");
+        expect(blockedRepository?.blockUser).to.eq(true);
+    }
+
+    @test(timeout(10000))
+    public async checkRepositoryIsNotBlocked() {
+        const typeorm = testContainer.get<TypeORM>(TypeORM);
+        const manager = await typeorm.getConnection();
+        manager.getRepository(DBBlockedRepository).insert({
+            urlRegexp: "github.com/bob/.*",
+            blockUser: true,
+            deleted: false,
+        });
+
+        const blockedRepository = await this.blockedRepositoryDb.isRepositoryBlocked("github.com/alice/some-repo");
+
+        expect(blockedRepository).undefined;
+    }
+
+    async wipeRepo() {
+        const typeorm = testContainer.get<TypeORM>(TypeORM);
+        const manager = await typeorm.getConnection();
+        await manager.getRepository(DBBlockedRepository).delete({});
+    }
+}
+
+module.exports = new BlockedRepositoryDBSpec();
--- a/components/gitpod-db/src/blocked-repository-db.spec.db.ts
+++ b/components/gitpod-db/src/blocked-repository-db.spec.db.ts
@@ -35,7 +35,7 @@ class BlockedRepositoryDBSpec {
             deleted: false,
         });
 
-        const blockedRepository = await this.blockedRepositoryDb.isRepositoryBlocked("github.com/bob/some-repo");
+        const blockedRepository = await this.blockedRepositoryDb.findBlockedRepositoryByURL("github.com/bob/some-repo");
 
         expect(blockedRepository).not.undefined;
         expect(blockedRepository?.urlRegexp).to.eq("github.com/bob/.*");
@@ -52,7 +52,9 @@ class BlockedRepositoryDBSpec {
             deleted: false,
         });
 
-        const blockedRepository = await this.blockedRepositoryDb.isRepositoryBlocked("github.com/alice/some-repo");
+        const blockedRepository = await this.blockedRepositoryDb.findBlockedRepositoryByURL(
+            "github.com/alice/some-repo",
+        );
 
         expect(blockedRepository).undefined;
     }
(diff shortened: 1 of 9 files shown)


components/gitpod-db/src/typeorm/entity/db-blocked-repository.ts
```diff
@@ -0,0 +1,35 @@
...
+    updatedAt: string;
+
+    @DeleteDateColumn()
+    deletedAt: string;
+
+    // TODO: not sure if we need this?
+    // TODO: should have a default value of false?
+    // This column triggers the db-sync deletion mechanism. It's not intended for public consumption.
+    @Column()
+    deleted: boolean;
```
andrew-farries, line 34: As the comment suggests, I'm not sure if we need this field on the new table.

easyCZ:
To be able to land this, we also need to udpate all the runbooks and point to how to establish a DB connection so that on-call engineers can block dynamically.

To make this less one-shot-change, can we first merge the config and the DB results when checking to enable gradual migration away? This would give us time to stage the change and communicate it.

andrew-farries:
> To be able to land this, we also need to udpate all the runbooks and point to how to establish a DB connection so that on-call engineers can block dynamically.

My plan (as described in https://github.com/gitpod-io/gitpod/issues/11030) was to allow the two mechanisms (reading from db and from config) to exist in parallel until we add the API and UI to interact with blocked repositories in the db. At that point we copy the blocked repos from config into the db. Until that point the blocked repository db table remains empty.

This PR isn't intended to replace the mechanism we currently use to block repos, just be a step towards that. So we don't need to update any runbooks etc.

easyCZ:
Nice, sorry didn't click through into the parent issue.

components/gitpod-db/src/blocked-repository-db.spec.db.ts
```diff
@@ -0,0 +1,67 @@
...
+
+        const blockedRepository = await this.blockedRepositoryDb.isRepositoryBlocked("github.com/bob/some-repo");
+
+        expect(blockedRepository).not.undefined;
+        expect(blockedRepository?.urlRegexp).to.eq("github.com/bob/.*");
+        expect(blockedRepository?.blockUser).to.eq(true);
+    }
+
+    @test(timeout(10000))
+    public async checkRepositoryIsNotBlocked() {
```
geropl, line 46: :orange_heart:

components/gitpod-db/src/blocked-repository-db.ts
```diff
@@ -0,0 +1,13 @@
...
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { BlockedRepository } from "@gitpod/gitpod-protocol/src/blocked-repositories-protocol";
+
+export const BlockedRepositoryDB = Symbol("BlockedRepositoryDB");
+
+export interface BlockedRepositoryDB {
+    isRepositoryBlocked(contextURL: string): Promise<BlockedRepository | undefined>;
```
geropl, line 12: nit: don't want to get into bike shedding, but wanted to share that I was confused by a method name starting with `is...` to return an object. As a result, I had to read the code twice, and have a look at the unit tests to make sure I understand how it works.

Did you think about calling it `findRepositoryBlockedByUrl` or similar?

andrew-farries: Not bikeshedding 👍 I agree it needs changing.

components/gitpod-db/src/typeorm/blocked-repository-db-impl.ts
```diff
@@ -0,0 +1,35 @@
...
+    async getBlockedRepositoryRepo(): Promise<Repository<DBBlockedRepository>> {
+        return (await this.getEntityManager()).getRepository<DBBlockedRepository>(DBBlockedRepository);
+    }
+
+    public async isRepositoryBlocked(contextURL: string): Promise<BlockedRepository | undefined> {
+        const blockedRepositoryRepo = await this.getBlockedRepositoryRepo();
+
+        const query = blockedRepositoryRepo
+            .createQueryBuilder("br")
+            .where(":pattern REGEXP br.urlRegexp", { pattern: contextURL });
```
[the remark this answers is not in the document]
andrew-farries: I didn't measure the performance but I did read something about how this works [[1]](https://stackoverflow.com/questions/12318083/mysql-optimization-for-regexp) [[2]](http://billauer.co.il/blog/2020/12/mysql-index-pattern-matching-performance/). 

It reads as though the `REGEXP` operation can't take advantage of any indexes on the `urlRegExp` column and has to do full table scan each time. I think this won't be a problem in practice given how little data we expect in this table but it's something to keep in mind.

components/gitpod-protocol/src/blocked-repositories-protocol.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+// TODO: Maybe this should be in protocol.ts?
```
geropl, line 7: > // TODO: Maybe this should be in protocol.ts?

Maybe it's fine to have one top-level file per (DB) entity. If anything I'd try to group it with other abuse-related shapes (if there is opportunity).

components/server/src/workspace/workspace-starter.ts
```diff
@@ -224,6 +226,7 @@ export class WorkspaceStarter {
         options = options || {};
         try {
             await this.checkBlockedRepository(user, workspace.contextURL);
+            await this.checkBlockedRepositoryInDB(user, workspace.contextURL);
```
geropl, line 229: Nice, like the approach! :+1:

components/gitpod-db/src/typeorm/entity/db-blocked-repository.ts
geropl: @andrew-farries For this DB table to be synced we also need an entry in `tables.ts`. Happy to chat about how that needs to look like.

components/gitpod-db/src/typeorm/blocked-repository-db-impl.ts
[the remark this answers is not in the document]
geropl: > I didn't measure the performance but I did read something about how this works

Cool! :+1: 

> I think this won't be a problem in practice given how little data we expect in this table but it's something to keep in mind.

Agree.

components/gitpod-db/src/typeorm/entity/db-blocked-repository.ts
andrew-farries: I've added an entry to `tables.ts` in c4aebe35ccc2540fc03e4ef8f547e56a8be1a652

components/gitpod-db/src/typeorm/entity/db-blocked-repository.ts
```diff
@@ -0,0 +1,33 @@
...
+    blockUser: boolean;
+
+    @CreateDateColumn()
+    createdAt: string;
+
+    @UpdateDateColumn()
+    updatedAt: string;
+
+    @DeleteDateColumn()
+    deletedAt: string;
```
geropl, line 28: I think we do not use this one.

geropl:
Testing...

geropl (approved):
Awesome, looking forward to have this! :rocket: 

/hold We should remove [this unused DB field](https://github.com/gitpod-io/gitpod/pull/11036#discussion_r911115003).

PR #11046, opened 2022-06-30
fix: Refactored server methods to return a 404 error instead of undefined when something cannot be found

### LINKED ISSUE #10892: Refactor  findStripeSubscriptionIdForTeam to return a 404 error instead of `undefined` when something cannot be found

Refactor ~all~ [findStripeSubscriptionIdForTeam](https://github.com/gitpod-io/gitpod/blob/26b4f555f4efd4b00f7912f63f930477e962c95d/components/server/ee/src/workspace/gitpod-server-impl.ts#L1913-L1915)  to return a 404 error instead of `undefined` when something cannot be found to be differentiate precisely between various cases.

E.g.

```ts
async findSomething(query): Promise<Something | undefined> {
    if (!something) {
        return undefined;
    }
    return something;
}
```

would become

```ts
async findSomething(query): Promise<Something> {
    if (!something) {
        throw new ResponseError(ErrorCodes.NOT_FOUND, "Could not find something");
    }
    return something;
}
```

This would of course also need to be handled by all clients, for example the dashboard (where a `undefined` check should become a 404 error catcher).

- Staying with `undefined` (status quo)

goverdhan07:
…ined when something cannot be found

## Description

Refactored server methods to return a 404 error instead of undefined when something cannot be found

## Related Issue(s)

Fixes #10892

## Release Notes

```release-note
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-protocol/src/gitpod-service.ts
+++ b/components/gitpod-protocol/src/gitpod-service.ts
@@ -276,7 +276,7 @@ export interface GitpodServer extends JsonRpcServer<GitpodClient>, AdminServer, 
 
     getStripePublishableKey(): Promise<string>;
     getStripeSetupIntentClientSecret(): Promise<string>;
-    findStripeSubscriptionIdForTeam(teamId: string): Promise<string | undefined>;
+    findStripeSubscriptionIdForTeam(teamId: string): Promise<string>;
     subscribeTeamToStripe(teamId: string, setupIntentId: string, currency: Currency): Promise<void>;
     getStripePortalUrlForTeam(teamId: string): Promise<string>;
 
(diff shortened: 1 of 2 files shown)


geropl:
Hey @goverdhan07 , thanks for your contribution! :pray:

I have a couple of remarks:
 1. When adjusting the implementation, we also need to adjust the [service definition](https://github.com/goverdhan07/gitpod/blob/refactor-server-methods/components/gitpod-protocol/src/gitpod-service.ts#L279). To check if it builds, you can run `cd components && yarn watch` inside your workspace
 2. If we adjust this method, we should also make sure that all call sites in the frontend handle this error in a sane and sensible way. You build it with `cd components/dashboard && yarn start`.


BTW I'll trigger our contributor lincense aggrement process in parallel so we can merge it in the end.

geropl (changes requested):
Compare [this comment](https://github.com/gitpod-io/gitpod/pull/11046#issuecomment-1172242206).

PR #11056, opened 2022-06-30
[kots]: add registry to preflight and support checks

### LINKED ISSUE #10337: KOTS registry pre-flight checks
Also check that can pull workspace images from Docker Hub

MrSimonEmms:
## Description

Adds a preflight and support checks to KOTS. As with the storage and database, this uses a container for this job. This extends the storage check so that the S3 backend can be checked as well

This uses Heroku's [Docker Registry Client](https://pkg.go.dev/github.com/heroku/docker-registry-client) to check the validity of the credentials. I've looked through the source code and, whilst it's not what we're using in the Image Builder, it's a much simpler structure than used elsewhere in the project. From what I could see, we'd have to write (and therefore maintain) a credential checker as I've not found that exact operation in our existing codebase. It does seem to work as expected for all the registries I've checked (Azure, GCP).

The GCP URLs add in a custom check to ensure that they're using the correct server address. This is two-fold:
1. the library always returns the server as authenticated if using `gcr.io/project/repo`, but uses the credentials if on `gcr.io`
2. we've found that you have to set that with GCP

It's arguably a bit of a hacky approach (I'm open to suggestions), but it does work for our needs.

As with all the preflight checks, they don't prevent the user from deploying but are there to highlight whether it'll work or not. Whilst we shouldn't get any false negatives, the risk is low.

## Related Issue(s)

Fixes #10337

## How to test

Deploy to a cluster and run tests against the config prefligths

## Release Notes

```release-note
[kots]: add registry to preflight and support checks
```

## Werft options:

- [ ] /werft with-preview


roboquat:

<details>

</details>

nandajavarma:
This is so useful!! Thanks for adding! :confetti_ball: 
registry_check

nandajavarma (approved):
works great!!

Issue #10418, opened 2022-06-02
[installer] remove need for YAML overrides / leverage config surface


kylos101:
## Is your feature request related to a problem? Please describe

Leverage the config surface for:
1. blobserve
2. workspace-templates
3. ws-manager
4. ws-daemon
5. ws-proxy

## Describe the behaviour you'd like

We use [YAML overrides](https://github.com/gitpod-io/ops/tree/main/deploy/workspace/manifests) for deploying the saas.

They are [applied here](https://github.com/gitpod-io/ops/blob/f675f178f36820bc9bca47110aebd02c94f2aff5/deploy/workspace/gitpod-up.sh#L231-L307).

Ideally we'd have the shape (structure) for the configuration that we want in the installer, some of which may be experimental, with logical corresponding defaults.

Consider continuing to use YAML overrides for`agent-smith` too, unless it is very easy to include as part of this effort. **Note**: the config values we use for agent-smith must still be applied "at deploy time" and persisted in the Ops repo.

## Describe alternatives you've considered

n/a

## Additional context

Related issues (close as needed):
https://github.com/gitpod-io/gitpod/issues/7748
https://github.com/gitpod-io/gitpod/issues/7747
https://github.com/gitpod-io/gitpod/issues/7746
https://github.com/gitpod-io/gitpod/issues/7006

kylos101:
This was scheduled so that we 1) treat the installer config surface as the source of truth and 2) stop using override manifests when we deploy. Why? Recently other teams were making changes to workspace components, and they were unaware that we use manifest overrides (like with `ws-proxy`).

sagor999:
all changes are done in PRs referenced above.
Once all of them are merged, this task can be marked as done.

sagor999:
all PRs have been merged.

PR #11058, opened 2022-06-30
[installer] allow to specify agent smith config

sagor999:
## Description

Allows to specify agent smith config via installer yaml, as currently we overwrite configmap instead when deploying in SaaS.

## Related Issue(s)

Related to https://github.com/gitpod-io/gitpod/issues/10418

## Release Notes

```release-note
none
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/installer/pkg/components/agent-smith/configmap.go
+++ b/install/installer/pkg/components/agent-smith/configmap.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/gitpod-io/gitpod/agent-smith/pkg/config"
 	"github.com/gitpod-io/gitpod/installer/pkg/common"
+	"github.com/gitpod-io/gitpod/installer/pkg/config/v1/experimental"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -29,10 +30,13 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		},
 	}
 
-	if ctx.Config.Experimental.AgentSmith != nil {
-		ascfg.Config = *ctx.Config.Experimental.AgentSmith
-		ascfg.Config.KubernetesNamespace = ctx.Namespace
-	}
+	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
+		if cfg.AgentSmith != nil {
+			ascfg.Config = *cfg.AgentSmith
+			ascfg.Config.KubernetesNamespace = ctx.Namespace
+		}
+		return nil
+	})
 
 	fc, err := common.ToJSONString(ascfg)
 	if err != nil {
--- a/install/installer/pkg/components/agent-smith/configmap.go
+++ b/install/installer/pkg/components/agent-smith/configmap.go
@@ -5,10 +5,8 @@
 package agentsmith
 
 import (
-	"encoding/base64"
 	"fmt"
 
-	"github.com/gitpod-io/gitpod/agent-smith/pkg/classifier"
 	"github.com/gitpod-io/gitpod/agent-smith/pkg/config"
 	"github.com/gitpod-io/gitpod/installer/pkg/common"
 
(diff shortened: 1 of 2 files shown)


install/installer/pkg/components/agent-smith/configmap.go
```diff
@@ -21,25 +19,20 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
...
+			Kubernetes:          config.Kubernetes{Enabled: true},
+			KubernetesNamespace: ctx.Namespace,
 			GitpodAPI: config.GitpodAPI{
 				HostURL: fmt.Sprintf("https://%s", ctx.Config.Domain),
 			},
 		},
 	}
 
+	if ctx.Config.AgentSmith != nil {
+		ascfg.Config = *ctx.Config.AgentSmith
```
aledbf, line 33: Can we overwrite `KubernetesNamespace` and `GitpodAPI` here?

sagor999: We can overwrite `KubernetesNamespace` but probably not `GitpodAPI` since it requires a token to be present as well. Unless it is already available somewhere in the installer yaml file? :thinking:

aledbf: I am sorry, I meant, to use the values from the installer as the source.

aledbf (approved):

utam0k (approved):

corneliusludmann (changes requested):
I really love the efforts to reduce the need for post-processing! That is awesome! :star_struck: 

---

**May I ask for a change in the Gitpod config surface?**

Although we use the Gitpod installer mainly internally, we want to make sure we maintain a stable Gitpod config struct and want to avoid that the public config surface getting too technical. Having the whole agent-smith config as part of the regular installer config feels quite intrusive to me. Every change in the agent-smith config has a direct effect on the Gitpod installer config and could introduce an incompatible config change. Community members that are using the installer directly would be affected by this.

If we want to make certain agent-smith config publicly available, we would ideally include a high-level config for it. E.g. a `block-list` config under the `workspace` section of this config. That would be more user-centric and does not need to understand what agent-smith actually is.

As long as we don't need to make this publically available (yet), but are using this for internal deployment only, my proposal is to simply move the config as is to the [`experimental` section](https://github.com/gitpod-io/gitpod/blob/d537b24994f67318dcadb9d29ebc132d36514f8c/install/installer/pkg/config/v1/experimental/experimental.go) instead. In this section, we have more freedom because we explicitly state that this can change at any point in time and users out there should not rely on this.

What do you think?

install/installer/pkg/components/agent-smith/configmap.go
```diff
@@ -21,25 +19,21 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
...
 			GitpodAPI: config.GitpodAPI{
 				HostURL: fmt.Sprintf("https://%s", ctx.Config.Domain),
 			},
 		},
 	}
 
+	if ctx.Config.Experimental.AgentSmith != nil {
+		ascfg.Config = *ctx.Config.Experimental.AgentSmith
+		ascfg.Config.KubernetesNamespace = ctx.Namespace
+	}
```
corneliusludmann, line 35:
```suggestion
	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
		if cfg.AgentSmith != nil {
			ascfg.Config = *cfg.AgentSmith
			ascfg.Config.KubernetesNamespace = ctx.Namespace
		}
		return nil
	})
```

corneliusludmann (changes requested):
Thanks, @sagor999 for changing the config value. Just a tiny change request to use the `WithExperimental` function. :pray: Then we are ready to merge! Thank you for your patience! :smile:

install/installer/pkg/components/agent-smith/configmap.go
sagor999: Doh! You are right, should have done it myself. Thanks for catching this. Fixed!

corneliusludmann (approved):
Thanks, @sagor999!

PR #11077, opened 2022-07-01
[preview] run the telemetry job right after gitpod is ready

Pothulapati:
## Description

This PR updates the `entry-point.sh` to run a parallel step
that waits for the gitpod pods to be ready, and manually executes
the cronjob once its necessary.

This is necessary as our telemetry `cronjobs` are only ran
daily in the midnight, and we can't expect this schedule to work
for local previews as they are short lived.

example data that we get:

```
client.Track(&analytics.Track{
  UserId: "334c379a-afee-4882-ba44-fbd7dfc0a8b0",
  Event: "Installation telemetry",
  Properties: map[string]interface{}{
    "customerID": "",
    "platform": "local-preview",
    "totalInstances": 0,
    "totalUsers": 0,
    "totalWorkspaces": 0,
    "version": "tar-preview-tel-init.4",
  },
})
```

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/11113

## How to test

Check if there is a new job called `gitpod-telemetry-init` is created once
gitpod pods are ready.

## Release Notes

```release-note
[preview] run the telemetry job right after a Gitpod install is ready
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/install/preview/entrypoint.sh
+++ b/install/preview/entrypoint.sh
@@ -136,6 +136,18 @@ yq eval-all -i 'del(.spec.template.spec.initContainers[0])' /var/lib/rancher/k3s
 for f in /var/lib/rancher/k3s/server/manifests/gitpod/*.yaml; do (cat "$f"; echo) >> /var/lib/rancher/k3s/server/manifests/gitpod.yaml; done
 rm -rf /var/lib/rancher/k3s/server/manifests/gitpod
 
+# waits for gitpod pods to be ready, and manually runs the `gitpod-telemetry` cronjob
+run_telemetry(){
+  # wait for the k3s cluster to be ready and Gitpod workloads are added
+  sleep 100
+  # indefinitely wait for Gitpod pods to be ready
+  kubectl wait --timeout=-1s --for=condition=ready pod -l app=gitpod,component!=migrations
+  # manually tun the cronjob
+  kubectl create job gitpod-telemetry-init --from=cronjob/gitpod-telemetry
+}
+
+run_telemetry 2>&1 &
+
 /bin/k3s server --disable traefik \
   --node-label gitpod.io/workload_meta=true \
   --node-label gitpod.io/workload_ide=true \


roboquat:

<details>

</details>

install/preview/entrypoint.sh
```diff
@@ -136,6 +136,18 @@ yq eval-all -i 'del(.spec.template.spec.initContainers[0])' /var/lib/rancher/k3s
 for f in /var/lib/rancher/k3s/server/manifests/gitpod/*.yaml; do (cat "$f"; echo) >> /var/lib/rancher/k3s/server/manifests/gitpod.yaml; done
 rm -rf /var/lib/rancher/k3s/server/manifests/gitpod
 
+# waits for gitpod pods to be ready, and manually runs the `gitpod-telemetry` cronjob
+run_telemetry(){
+  # wait for the k3s cluster to be ready and Gitpod workloads are added
+  sleep 100
```
corneliusludmann, line 142: Magic sleep time feels always like a hack. What if it takes longer than 100 for some reason? However, maybe good enough for this use case.

install/preview/entrypoint.sh
```diff
@@ -136,6 +136,18 @@ yq eval-all -i 'del(.spec.template.spec.initContainers[0])' /var/lib/rancher/k3s
...
+  # indefinitely wait for Gitpod pods to be ready
+  kubectl wait --timeout=7m --for=condition=ready pod -l app=gitpod,component!=migrations
```
corneliusludmann, line 144: Are 7 minutes long enough? What if it takes longer to pull the images due to bad internet connectivity?

corneliusludmann (approved):
Added 2 minor comments.

/hold in case you want to change something. Otherwise, feel free to remove the hold.

install/preview/entrypoint.sh
Pothulapati: Yep, It is a hack but Just for atleast one Gitpod pod to be just added seems like a valid wait time. Other hacks involve looping on `kubectl wait` command https://github.com/kubernetes/kubernetes/issues/83242#issuecomment-1029256777 🤔

install/preview/entrypoint.sh
Pothulapati: Makes sense. Updated it to be `-1s` which means a week, as users are not expecting feedback from this command, and we can wait more time. wdyt? 🤔

install/preview/entrypoint.sh
Pothulapati: Thinking more about this. A lot of these kind of hacks, seem to be cause of the fact that we reun `k3s` with manifests added. How about we run the cluster parallely, and apply manifests on our own. We can then also remove `prettylog` and instead print statements as they happen, and we'll have much more easy time handling the script 🤔

Pothulapati:
Also, Ignore the added changes. 
Those are now removed to limit this change to running the telemetry when Gitpod is ready, and only that. Exit telemetry will added in a separate PR.

install/preview/entrypoint.sh
corneliusludmann: Makes sense. :+1:

corneliusludmann:
Great! Feel free to remove the hold label when you are ready to merge.

PR #11080, opened 2022-07-01
[server] Add server API to list blocked repositories

andrew-farries:
## Description

Add `adminGetBlockedRepositories` API to `components/server` to fetch the list of blocked repositories.

This is a step towards using these endpoints to manage blocked repositories in the admin UI.

## Related Issue(s)

Part of #11030

## How to test

* Follow the first 3 steps on https://github.com/gitpod-io/gitpod/pull/11036 to add some blocked repositories.

* Open the developer tools on the admin page for this preview environment.

* Run:
```js
await window._gp.gitpodService.server.adminGetBlockedRepositories({offset: 0, limit: 10, orderBy: 'id', orderDir: "desc", searchTerm: "foo"})
```

setting `searchTerm` as appropriate (or omitting it entirely).

## Release Notes

```release-note
NONE
```

## Werft options

- [x] /werft with-preview

/hold

### DIFF
--- a/components/gitpod-db/src/blocked-repository-db.ts
+++ b/components/gitpod-db/src/blocked-repository-db.ts
@@ -9,5 +9,19 @@ import { BlockedRepository } from "@gitpod/gitpod-protocol/src/blocked-repositor
 export const BlockedRepositoryDB = Symbol("BlockedRepositoryDB");
 
 export interface BlockedRepositoryDB {
+    findAllBlockedRepositories(
+        offset: number,
+        limit: number,
+        orderBy: keyof BlockedRepository,
+        orderDir: "DESC" | "ASC",
+        searchTerm?: string,
+        minCreationDate?: Date,
+        maxCreationDate?: Date,
+    ): Promise<{ total: number; rows: BlockedRepository[] }>;
+
     findBlockedRepositoryByURL(contextURL: string): Promise<BlockedRepository | undefined>;
+
+    createBlockedRepository(urlRegexp: string, blockUser: boolean): Promise<BlockedRepository>;
+
+    deleteBlockedRepository(id: number): Promise<boolean>;
 }
--- a/components/gitpod-db/src/typeorm/blocked-repository-db-impl.ts
+++ b/components/gitpod-db/src/typeorm/blocked-repository-db-impl.ts
@@ -23,6 +23,50 @@ export class TypeORMBlockedRepositoryDBImpl implements BlockedRepositoryDB {
         return (await this.getEntityManager()).getRepository<DBBlockedRepository>(DBBlockedRepository);
     }
 
+    public async createBlockedRepository(urlRegexp: string, blockUser: boolean): Promise<BlockedRepository> {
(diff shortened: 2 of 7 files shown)


components/gitpod-db/src/blocked-repository-db.ts
```diff
@@ -9,5 +9,15 @@ import { BlockedRepository } from "@gitpod/gitpod-protocol/src/blocked-repositor
 export const BlockedRepositoryDB = Symbol("BlockedRepositoryDB");
 
 export interface BlockedRepositoryDB {
+    findAllBlockedRepositories(
+        offset: number,
+        limit: number,
+        orderBy: keyof BlockedRepository,
```
geropl, line 15: Here it might make sense to limit to the fields that actually are indexex, for instance by using `orderBy: keyof Pick<BlockedRepository, "id | "blockUser">` or similar.

components/gitpod-db/src/typeorm/blocked-repository-db-impl.ts
```diff
@@ -23,6 +23,37 @@ export class TypeORMBlockedRepositoryDBImpl implements BlockedRepositoryDB {
...
+        orderDir: "DESC" | "ASC",
+        searchTerm?: string,
+        minCreationDate?: Date,
+        maxCreationDate?: Date,
+    ): Promise<{ total: number; rows: BlockedRepository[] }> {
+        const blockedRepositoryRepo = await this.getBlockedRepositoryRepo();
+
+        const qBuilder = blockedRepositoryRepo.createQueryBuilder("br");
+        if (searchTerm) {
+            qBuilder.andWhere(`br.urlRegexp LIKE :searchTerm`, { searchTerm: "%" + searchTerm + "%" });
```
geropl, line 39: Note that this won't work: `andWhere` strictly requires a prior `where` this is a typeorm oddity... :see_no_evil: 
Usually it can be solved by having something general as first clause, maybe `.where("br.deleted = FALSE")`.

geropl:
Left some comments. Please squash before merging! :upside_down_face:

components/gitpod-db/src/blocked-repository-db.ts
andrew-farries: We have no indexes on the table besides the PK. As mentioned yesterday https://github.com/gitpod-io/gitpod/pull/11036#discussion_r910989383, mysql can't make use of indexes on `urlRegexp` when testing with `REGEXP`.

Do you think we should add indexes to support this admin search?

components/gitpod-db/src/typeorm/blocked-repository-db-impl.ts
geropl: > This does seem to work though; tests using the searchTerm pass and it behaves as expected when tested manually too.

Uh, that's nice! Might have been introduced with the typeorm version bump we did last October - thx for sharing!

geropl:
/werft run


:crossed_fingers:

geropl (approved):
Code LGTM, tested and works! :heavy_check_mark: 

/hold Please squash before merging! :upside_down_face:

PR #11081, opened 2022-07-01
Auto-forward all workspace open ports when using Latest JetBrains IDEs

### LINKED ISSUE #10783: [jb] auto forward all ports
JB backend plugin should listen to supervisor port updates and auto forward all ports using Remote Dev APIs [[1](https://gitpod.slack.com/archives/C02BRJLGPGF/p1655728987129749?thread_ts=1655709305.050799&cid=C02BRJLGPGF)].

UI elements like port notifications then should respect such mapping.

felladrin:
## Description

Auto-forward all workspace open ports when using Latest JetBrains IDEs.

Changes included:
- Adds a `GitpodPortsService`, which is used by both _Stable_ and _Latest_ versions of the Backend Plugin.
- Adds a `GitpodPortForwardingService`, which affects only Latest (Unstable) Version of JetBrains IDEs, and controls the auto-forwarding of all ports.
- Adds a `GitpodPortsServiceTest` with a function unit test. It can be tested via IDE UI or command line [[1](https://github.com/gitpod-io/gitpod/pull/11081#discussion_r938863805)].
- Updates `GitpodTerminalService`, as we moved the implementation to forward all ports to `GitpodPortForwardingService`.
- Updates `GitpodClientProjectSessionTracker`, which displays notifications about ports that have just been opened, to have the correct localhost URL if the port is being forwarded. (A few other small changes were made on that file, which were suggestions from IntelliJ to improve the code).
- Updates `GitpodCLIService` to rewrite URLs coming from `gp preview`, when the URL is pointing to localhost, for pointing to the correct forwarded port on the client.
  - For example, a user starts a service on port 3000, which is automatically forwarded from the server to the user machine. But the user had port 3000 already occupied by some other process. What JetBrains IDE does, in this case, is looking for the next free port (3001, for example). So whenever the user runs `gp preview http://127.0.0.1:3000`, it should open `http://127.0.0.1:3001` on the browser.
  - Note: It currently doesn't happen because of [this issue](https://github.com/gitpod-io/gitpod/issues/11849), which will be solved later. So this shouldn't be tested in this PR. But the class is ready for when the issue is fixed.
- Update of gradle-intellij-plugin to v1.8.0, which adds [support for JetBrains Runtime 2022.2 directories layout](https://github.com/JetBrains/gradle-intellij-plugin/issues/1016).

Note: We confirmed with JetBrains that the host used for Forwarded Ports on the client is 127.0.0.1.

## Related Issue(s)

Resolves #10783

## How to test

- Open [Preferences on the Preview Environment of this PR](https://felladrin-d4177edc98.preview.gitpod-dev.com/preferences), select any JetBrains IDE, and mark Latest Release to use the EAP version of the IDE.
- Start any workspace, and once connected, execute any HTTP server from any terminal (can be a terminal used by a Gitpod Task or a new terminal created via the IDE UI). (You can run lama.sh for example: `sh <(curl lama.sh) -p 8080`)
- Click on the "Open Browser" link from the notification and confirm it opens `http://127.0.0.1:<forwarded_service_port>/` on your browser.
  image

Note:
- If the port is already occupied by another process, JetBrains IDE will look for some random open port. (This random port unfortunately isn't displayed anywhere at this moment. It will be solved in a follow-up PR.)
- The links from notifications about open ports should point to localhost only when using the Latest (Unstable) version of the IDE. When using the Stable version, links will point to the remote URL, as usual.
- This PR should be affecting just the toast notification, not the dialog [[1](https://github.com/gitpod-io/gitpod/pull/11081#discussion_r935444156)]. The dialog will be resolved in [another PR](https://github.com/gitpod-io/gitpod/issues/11849).

## Release Notes

```release-note
All open ports from the workspace are automatically forwarded when using Latest JetBrains IDEs.
```

## Documentation

None.

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
+++ b/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
@@ -16,6 +16,10 @@ import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.io.BufferExposingByteArrayOutputStream
 import com.intellij.openapi.util.io.FileUtilRt
 import com.intellij.util.application
+import com.intellij.util.withFragment
+import com.intellij.util.withPath
+import com.intellij.util.withQuery
+import com.jetbrains.rd.util.URI
 import io.netty.buffer.Unpooled
 import io.netty.channel.ChannelHandlerContext
 import io.netty.handler.codec.http.FullHttpRequest
@@ -31,6 +35,7 @@ import java.nio.file.Path
 class GitpodCLIService : RestService() {
 
     private val manager = service<GitpodManager>()
+    private val portsService = service<GitpodPortsService>()
 
     override fun getServiceName() = SERVICE_NAME
 
@@ -68,11 +73,28 @@ class GitpodCLIService : RestService() {
             if (url.isNullOrBlank()) {
                 return "url is missing"
             }
+
+            val resolvedUrl = resolveExternalUrl(url)
+
             return withClient(request, context) { project ->
-                BrowserUtil.browse(url, project)
+                BrowserUtil.browse(resolvedUrl, project)
             }
         }
         return "invalid operation"
+    }
+
+    private fun resolveExternalUrl(url: String): String {
+        val uri = URI.create(url)
+        val optionalLocalHostUriMetadata = portsService.extractLocalHostUriMetaDataForPortMapping(uri)
+
+        return when {
+            optionalLocalHostUriMetadata.isEmpty -> url
+            else -> portsService.getLocalHostUriFromHostPort(optionalLocalHostUriMetadata.get().port)
+                    .withPath(uri.path)
+                    .withQuery(uri.query)
+                    .withFragment(uri.fragment)
+                    .toString()
+        }
     }
 
     private fun withClient(request: FullHttpRequest, context: ChannelHandlerContext, action: (project: Project?) -> Unit): String? {
--- a/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodClientProjectSessionTracker.kt
+++ b/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodClientProjectSessionTracker.kt
@@ -33,11 +33,13 @@ import org.jetbrains.ide.BuiltInServerManager
 import java.util.concurrent.CancellationException
 import java.util.concurrent.CompletableFuture
 
+@Suppress("UnstableApiUsage", "OPT_IN_USAGE")
 class GitpodClientProjectSessionTracker(
         private val session: ClientProjectSession
 ) : Disposable {
 
     private val manager = service<GitpodManager>()
+    private val portsService = service<GitpodPortsService>()
 
     private lateinit var info: Info.WorkspaceInfoResponse
     private val lifetime = Lifetime.Eternal.createNested()
@@ -54,19 +56,26 @@ class GitpodClientProjectSessionTracker(
         }
     }
 
-    private fun isExposedServedPort(port: Status.PortsStatus?): Boolean {
+    private fun isExposedServedPort(port: PortsStatus?): Boolean {
         if (port === null) {
             return false
         }
         return port.served && port.hasExposed()
     }
 
+    private fun getForwardedPortUrl(port: PortsStatus): String {
+        return when {
+            portsService.isForwarded(port.localPort) -> portsService.getLocalHostUriFromHostPort(port.localPort).toString()
+            else -> port.exposed.url
+        }
+    }
+
     private fun showOpenServiceNotification(port: PortsStatus, offerMakePublic: Boolean = false) {
         val message = "A service is available on port ${port.localPort}"
         val notification = manager.notificationGroup.createNotification(message, NotificationType.INFORMATION)
 
-        val openBrowserAction = NotificationAction.createSimple("Open Browser") {
-            openBrowser(port.exposed.url)
+        val openBrowserAction = NotificationAction.createSimple("Open browser") {
+            openBrowser(getForwardedPortUrl(port))
         }
         notification.addAction(openBrowserAction)
 
@@ -76,7 +85,7 @@ class GitpodClientProjectSessionTracker(
                     makePortPublic(info.workspaceId, port)
                 }
             }
-            val makePublicAction = NotificationAction.createSimple("Make Public", makePublicLambda)
+            val makePublicAction = NotificationAction.createSimple("Make public", makePublicLambda)
             notification.addAction(makePublicAction)
         }
 
@@ -113,7 +122,7 @@ class GitpodClientProjectSessionTracker(
         val backendPort = BuiltInServerManager.getInstance().waitForStart().port
         val serverPort = StartupUtil.getServerFuture().await().port
         val ignorePorts = listOf(backendPort, serverPort, 5990)
-        val portsStatus = hashMapOf<Int, Status.PortsStatus>()
+        val portsStatus = hashMapOf<Int, PortsStatus>()
 
         val status = StatusServiceGrpc.newStub(GitpodManager.supervisorChannel)
         while (isActive) {
@@ -147,7 +156,7 @@ class GitpodClientProjectSessionTracker(
                                         }
 
                                         if (port.exposed.onExposed.number == Status.OnPortExposedAction.open_browser_VALUE || port.exposed.onExposed.number == Status.OnPortExposedAction.open_preview_VALUE) {
-                                            openBrowser(port.exposed.url)
+                                            openBrowser(getForwardedPortUrl(port))
                                             continue
                                         }
 
@@ -157,7 +166,7 @@ class GitpodClientProjectSessionTracker(
                                         }
 
                                         if (port.exposed.onExposed.number == Status.OnPortExposedAction.notify_private_VALUE) {
-                                            showOpenServiceNotification(port, port.exposed.visibilityValue !== PortVisibility.public_visibility_VALUE)
+                                            showOpenServiceNotification(port, port.exposed.visibilityValue != PortVisibility.public_visibility_VALUE)
                                             continue
                                         }
                                     }
--- a/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodPortsService.kt
+++ b/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodPortsService.kt
@@ -1,0 +1,81 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package io.gitpod.jetbrains.remote
+
+import com.intellij.openapi.diagnostic.thisLogger
+import com.jetbrains.rd.util.URI
+import org.apache.http.client.utils.URIBuilder
+import java.util.Optional
+import java.util.regex.Pattern
+
+class GitpodPortsService {
+    companion object {
+        /** Host used by forwarded ports on JetBrains Client. */
+        const val FORWARDED_PORT_HOST = "127.0.0.1"
+    }
+    private val hostToClientForwardedPortMap: MutableMap<Int, Int> = mutableMapOf()
+
+    fun isForwarded(hostPort: Int): Boolean = hostToClientForwardedPortMap.containsKey(hostPort)
+
+    private fun getForwardedPort(hostPort: Int): Optional<Int> = Optional.ofNullable(hostToClientForwardedPortMap[hostPort])
+
+    fun setForwardedPort(hostPort: Int, clientPort: Int) {
+        hostToClientForwardedPortMap[hostPort] = clientPort
+    }
+
+    fun removeForwardedPort(hostPort: Int) {
+        hostToClientForwardedPortMap.remove(hostPort)
+    }
+
+    fun getLocalHostUriFromHostPort(hostPort: Int): URI {
+        val optionalForwardedPort = getForwardedPort(hostPort)
+
+        val port = if (optionalForwardedPort.isPresent) {
+            optionalForwardedPort.get()
+        } else {
+            thisLogger().warn(
+                    "gitpod: Tried to get the forwarded port of $hostPort, which was not forwarded. " +
+                    "Returning $hostPort itself."
+            )
+            hostPort
+        }
+
+        return URIBuilder()
+                .setScheme("http")
+                .setHost(FORWARDED_PORT_HOST)
+                .setPort(port)
+                .build()
+    }
+
+    interface LocalHostUriMetadata {
+        val address: String
+        val port: Int
+    }
+
+    fun extractLocalHostUriMetaDataForPortMapping(uri: URI): Optional<LocalHostUriMetadata> {
+        if (uri.scheme != "http" && uri.scheme != "https") return Optional.empty()
+
+        val localhostMatch = Pattern.compile("^(localhost|127(?:\\.[0-9]+){0,2}\\.[0-9]+|0+(?:\\.0+){0,2}\\.0+|\\[(?:0*:)*?:?0*1?])(?::(\\d+))?\$").matcher(uri.authority)
+
+        if (!localhostMatch.find()) return Optional.empty()
+
+        var address = localhostMatch.group(1)
+        if (address.startsWith('[') && address.endsWith(']')) {
+            address = address.substring(1, address.length - 2)
+        }
+
+        var port = 443
+        try {
+            port = localhostMatch.group(2).toInt()
+        } catch (throwable: Throwable){
+            if (uri.scheme == "http") port = 80
+        }
+
+        return Optional.of(object: LocalHostUriMetadata {
+             override val address = address
+             override val port = port
+        })
+    }
+}
--- a/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/latest/GitpodPortForwardingService.kt
+++ b/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/latest/GitpodPortForwardingService.kt
@@ -1,0 +1,129 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package io.gitpod.jetbrains.remote.latest
+
+import com.intellij.openapi.components.service
+import com.intellij.openapi.diagnostic.thisLogger
+import com.intellij.openapi.project.Project
+import com.intellij.remoteDev.util.onTerminationOrNow
+import com.intellij.util.application
+import com.jetbrains.codeWithMe.model.RdPortType
+import com.jetbrains.rd.platform.util.lifetime
+import com.jetbrains.rd.util.lifetime.LifetimeStatus
+import com.jetbrains.rdserver.portForwarding.ForwardedPortInfo
+import com.jetbrains.rdserver.portForwarding.PortForwardingManager
+import com.jetbrains.rdserver.portForwarding.remoteDev.PortEventsProcessor
+import io.gitpod.jetbrains.remote.GitpodManager
+import io.gitpod.jetbrains.remote.GitpodPortsService
+import io.gitpod.supervisor.api.Status
+import io.gitpod.supervisor.api.StatusServiceGrpc
+import io.grpc.stub.ClientCallStreamObserver
+import io.grpc.stub.ClientResponseObserver
+import io.ktor.utils.io.*
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.TimeUnit
+
+@Suppress("UnstableApiUsage")
+class GitpodPortForwardingService(private val project: Project) {
+    companion object {
+        const val FORWARDED_PORT_LABEL = "gitpod"
+    }
+
+    private val portsService = service<GitpodPortsService>()
+
+    init { start() }
+
+    private fun start() {
+        if (application.isHeadlessEnvironment) return
+
+        observePortsListWhileProjectIsOpen()
+    }
+
+    private fun observePortsListWhileProjectIsOpen() = application.executeOnPooledThread {
+        while (project.lifetime.status == LifetimeStatus.Alive) {
+            try {
+                observePortsList().get()
+            } catch (throwable: Throwable) {
+                when (throwable) {
+                    is InterruptedException, is CancellationException -> break
+                    else -> thisLogger().error(
+                            "gitpod: Got an error while trying to get ports list from Supervisor. " +
+                                    "Going to try again in a second.",
+                            throwable
+                    )
+                }
+            }
+
+            TimeUnit.SECONDS.sleep(1)
+        }
+    }
+
+    private fun observePortsList(): CompletableFuture<Void> {
+        val completableFuture = CompletableFuture<Void>()
+
+        val statusServiceStub = StatusServiceGrpc.newStub(GitpodManager.supervisorChannel)
+
+        val portsStatusRequest = Status.PortsStatusRequest.newBuilder().setObserve(true).build()
+
+        val portsStatusResponseObserver = object :
+                ClientResponseObserver<Status.PortsStatusRequest, Status.PortsStatusResponse> {
+            override fun beforeStart(request: ClientCallStreamObserver<Status.PortsStatusRequest>) {
+                project.lifetime.onTerminationOrNow { request.cancel("gitpod: Project terminated.", null) }
+            }
+            override fun onNext(response: Status.PortsStatusResponse) {
+                application.invokeLater { updateForwardedPortsList(response) }
+            }
+            override fun onCompleted() { completableFuture.complete(null) }
+            override fun onError(throwable: Throwable) { completableFuture.completeExceptionally(throwable) }
+        }
+
+        statusServiceStub.portsStatus(portsStatusRequest, portsStatusResponseObserver)
+
+        return completableFuture
+    }
+
+    private fun updateForwardedPortsList(response: Status.PortsStatusResponse) {
+        val portForwardingManager = PortForwardingManager.getInstance(project)
+        val forwardedPortsList = portForwardingManager.getForwardedPortsWithLabel(FORWARDED_PORT_LABEL)
+
+        for (port in response.portsList) {
+            val hostPort = port.localPort
+            val isServed = port.served
+
+            if (isServed && !forwardedPortsList.containsKey(hostPort)) {
+                val portEventsProcessor = object : PortEventsProcessor {
+                    override fun onPortForwarded(hostPort: Int, clientPort: Int) {
+                        portsService.setForwardedPort(hostPort, clientPort)
+                        thisLogger().info("gitpod: Forwarded port $hostPort to client's port $clientPort.")
+                    }
+
+                    override fun onPortForwardingEnded(hostPort: Int) {
+                        thisLogger().info("gitpod: Finished forwarding port $hostPort.")
+                    }
+
+                    override fun onPortForwardingFailed(hostPort: Int, reason: String) {
+                        thisLogger().error("gitpod: Failed to forward port $hostPort: $reason")
+                    }
+                }
+
+                val portInfo = ForwardedPortInfo(
+                        hostPort,
+                        RdPortType.HTTP,
+                        FORWARDED_PORT_LABEL,
+                        emptyList(),
+                        portEventsProcessor
+                )
+
+                portForwardingManager.forwardPort(portInfo)
+            }
+
+            if (!isServed && forwardedPortsList.containsKey(hostPort)) {
+                portForwardingManager.removePort(hostPort)
+                portsService.removeForwardedPort(hostPort)
+                thisLogger().info("gitpod: Stopped forwarding port $hostPort.")
+            }
+        }
+    }
+}
--- a/components/ide/jetbrains/backend-plugin/build.gradle.kts
+++ b/components/ide/jetbrains/backend-plugin/build.gradle.kts
@@ -13,7 +13,7 @@ plugins {
     // Kotlin support - check the latest version at https://plugins.gradle.org/plugin/org.jetbrains.kotlin.jvm
     id("org.jetbrains.kotlin.jvm") version "1.7.0"
     // gradle-intellij-plugin - read more: https://github.com/JetBrains/gradle-intellij-plugin
-    id("org.jetbrains.intellij") version "1.6.0"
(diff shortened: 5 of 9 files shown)


akosyakov:
It is not enough to forward. We should make use of them, in UI elements like notifications. Could you review please the backend for such places and there we use port URL translate them to localhost host if available.

felladrin:
> It is not enough to forward. We should make use of them, in UI elements like notifications. Could you review please the backend for such places and there we use port URL translate them to localhost host if available?

Understood. In this case, we'll see some files being duplicated (between "stable" and "latest" folders/packages), as port forwarding only works on "latest".

I'll leave this PR as a Draft while I update it.

felladrin:
Updated! It's ready for review again!

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
```diff
@@ -64,10 +65,30 @@ class GitpodCLIService : RestService() {
...
-            val url = getStringParameter("url", urlDecoder)
+            var url = getStringParameter("url", urlDecoder)
+
             if (url.isNullOrBlank()) {
                 return "url is missing"
             }
+
+            // When it's auto-forwarding ports, we need to consider Gitpod Tasks running "gp preview".
+            // For example: gp preview --external $(gp url 3000)
+            if (manager.isAutoForwardingPorts) {
```
akosyakov, line 76: if a user wants to access exposed API then we should not translate it to local.

We should mimic logic of VS Code [[1](https://github.com/gitpod-io/openvscode-server/blob/9d68ec908452a065efdd6e8430c67b2898e4f038/src/vs/gitpod/browser/workbench/workbench.ts#L908-L926)] and apply it in similar situations.

felladrin: Let me check.

felladrin: @akosyakov, I didn't understand the usage of tunnels and how I should use them here.

The port forwarding of JetBrains is not related to the tunnels used by the Local Companion App, right?

Could you clarify? Or we can sync tomorrow.

akosyakov: In this code, tunnel is basically like port forwarding in JB, so the logic is:
- if a user is trying to open URI which is local [1](https://github.com/gitpod-io/openvscode-server/blob/9d68ec908452a065efdd6e8430c67b2898e4f038/src/vs/platform/tunnel/common/tunnel.ts#L134)] then apply port forwarding mapping
- otherwise just open URI 

So only localhost URL can be translated, not localhost URIs are never translated.

Try to play with VS Code Browser + Local Companion for following use cases:
- `gp preview localhost:3000` when let's say 3000 is forwarded on 3001
- and then `gp preview $(gp url 3000)` 

it should work the same way for JB and VS Code

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodClientProjectSessionTracker.kt
```diff
@@ -54,19 +55,26 @@ class GitpodClientProjectSessionTracker(
...
+    private fun isExposedServedPort(port: PortsStatus?): Boolean {
         if (port === null) {
             return false
         }
         return port.served && port.hasExposed()
     }
 
+    private fun getForwardedPortUrl(port: PortsStatus): String {
+        return when {
+            manager.isAutoForwardingPorts -> "http://localhost:${port.localPort}"
```
akosyakov, line 67: Is it really correct? Port forwarding can happen on another port if such local port is occupied we should read info from manager?

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodManager.kt
```diff
@@ -62,6 +62,7 @@ class GitpodManager : Disposable {
     }
 
     val devMode = System.getenv("JB_DEV").toBoolean()
+    var isAutoForwardingPorts = false
```
akosyakov, line 65: I am not sure it is good idea, instead we should take the given local port and look up in JB prot manager whether there is a mapping for it and if so use actual local port

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
felladrin: ℹ️ Haven't had the time to update it yet, but I will.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodClientProjectSessionTracker.kt
felladrin: True. When I wrote it, I thought JetBrains didn't have the functionality of forwarding to a different port. But I saw in the last meeting that they do have it.
I'll update it to use the port from JB Port Manager.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodClientProjectSessionTracker.kt
```diff
@@ -54,19 +55,26 @@ class GitpodClientProjectSessionTracker(
...
+        }
+    }
+
     private fun showOpenServiceNotification(port: PortsStatus, offerMakePublic: Boolean = false) {
         val message = "A service is available on port ${port.localPort}"
         val notification = manager.notificationGroup.createNotification(message, NotificationType.INFORMATION)
 
-        val openBrowserAction = NotificationAction.createSimple("Open Browser") {
-            openBrowser(port.exposed.url)
+        val openBrowserAction = NotificationAction.createSimple("Open browser") {
```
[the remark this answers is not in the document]
felladrin: I was trying to leave the file fully compliant (without yellow/red marks). JetBrains linter throws a warning about capital letters in all words of a sentence. I checked other JetBrains IDE buttons and confirmed they follow this pattern of button labels having just the first word of the sentence with a capital letter.
I can revert the labels if it's a problem.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodManager.kt
felladrin: I did it like this because we currently need to support JB IDEs v2022.1, which doesn't have the port forwarding feature. Doing so wouldn't need to move the files to 'stable' and 'latest' folders. That was the best I could think of to keep the changes minimal.
But let me review it.

akosyakov: I see, so it is kind of a feature flag? I think you need an interface like:
```java
interface GitpodPortService {
   forward(localPort: int) Optional<int>
}
```

For stable you implement then `NullGitpodPortService` which always return null, so no tunnelling, for latest you are implementing the real one which is using JB Port Manager.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodPortsService.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    interface LocalHostUriMetadata {
+        val address: String
+        val port: Int
+    }
+
+    fun extractLocalHostUriMetaDataForPortMapping(uri: URI): Optional<LocalHostUriMetadata> {
+        if (uri.scheme != "http" && uri.scheme != "https") return Optional.empty()
+
+        val localhostMatch = Pattern.compile("^(localhost|127(?:\\.[0-9]+){0,2}\\.[0-9]+|0+(?:\\.0+){0,2}\\.0+|\\[(?:0*:)*?:?0*1?])(?::(\\d+))?\$").matcher(uri.authority)
```
felladrin, line 34: ℹ️ This regex can be tested [here](https://pl.kotl.in/FFo1Yypsv).

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodClientProjectSessionTracker.kt
felladrin: Done. Now it's using the Client Port informed by JB Port Manager.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodPortsService.kt
```diff
@@ -0,0 +1,55 @@
...
+// See License-AGPL.txt in the project root for license information.
+
+package io.gitpod.jetbrains.remote
+
+import com.jetbrains.rd.util.URI
+import java.util.Optional
+import java.util.regex.Pattern
+
+class GitpodPortsService {
+    private val hostToClientForwardedPortMap: MutableMap<Int, Int> = mutableMapOf()
```
akosyakov, line 12: Is not JB already exposes such map to us?

Duplicating and synching content is a recipe for out of sync and race conditions usually. Could we use JB services as a source of the truth instead?

felladrin: I've checked, but unfortunately, they don't.

Here are all 4 methods provided by their PortManager at the moment of writing:
image

And this is what they provide in the PortForwarding package (I've also looked in the sub-packages):
image

Note: This is in the snapshot of 2022.2 (which is used in 'io.gitpod.jetbrains.remote.latest' package).
I haven't checked 2022.3 snapshot yet, because EAP JB IDEs on Gitpod.io are still 2022.2.

felladrin:
Btw, I've confirmed that gp-cli's `preview` command already converts the 'localhost' to the workspace URL:
https://github.com/gitpod-io/gitpod/blob/bffbd3216205d568e9820a7eae30d2af8ea855ba/components/gitpod-cli/cmd/preview.go#L27-L32

So even now (without the changes from this PR), notifications display the remote address (e.g. `https://8080-gitpodio-gitpod-zufs0q98iqf.ws-eu54.gitpod.io`) when using `gp preview http://localhost:8080` or `gp preview $(gp url 8080)`:
image

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
felladrin: I've confirmed that gp-cli's `preview` command already converts the 'localhost' to the workspace URL:
https://github.com/gitpod-io/gitpod/blob/bffbd3216205d568e9820a7eae30d2af8ea855ba/components/gitpod-cli/cmd/preview.go#L27-L32

So even now (without the changes from this PR), notifications display the remote address (e.g. `https://8080-gitpodio-gitpod-zufs0q98iqf.ws-eu54.gitpod.io`) when using `gp preview http://localhost:8080` or `gp preview $(gp url 8080)`:
image

As it's never receiving a localhost URL, I think it doesn't worth changing [GitpodCLIService.kt](https://github.com/gitpod-io/gitpod/pull/11081/files#diff-81d1cb74ab7142c7fba14c9bb140bb6a38eedf7d9e72a5f10065b130f95d2801) in this PR.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodPortsService.kt
akosyakov: `getForwardedPortsWithLabel` should do?

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
akosyakov: let's create an issue for it, we should move translation to IDEs from gp cli

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodPortsService.kt
felladrin: That's what I was expecting too (the `ForwardedPortInfo` to have a `clientPort`), but it doesn't.
The objects from the `getForwardedPortsWithLabel` map have only these properties:
image

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodCLIService.kt
felladrin: Issue created:
- https://github.com/gitpod-io/gitpod/issues/11849

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodPortsService.kt
akosyakov: Could you check with JB folks why is it? For now we can go with your solution. But we should talk with them to refine APIs.

felladrin:
I've updated the description and "how to test" sections and the PR is now ready for review.

@mustard-mh, I'd love to have also your review in this PR. 🙏

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodPortsService.kt
```diff
@@ -0,0 +1,81 @@
...
+            thisLogger().warn(
+                    "gitpod: Tried to get the forwarded port of $hostPort, which was not forwarded. " +
+                    "Returning $hostPort itself."
+            )
+            hostPort
+        }
+
+        return URIBuilder()
+                .setScheme("http")
+                .setHost(FORWARDED_PORT_HOST)
```
mustard-mh, line 47: If the port was not forwarded, will it worked in `127.0.0.1`? If not (if they are works with workspace URL), the `scheme` should change to `https` for gitpod SaaS too.

(self-hosted with self build may make it possible to use http themselves)

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodPortsService.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    fun extractLocalHostUriMetaDataForPortMapping(uri: URI): Optional<LocalHostUriMetadata> {
+        if (uri.scheme != "http" && uri.scheme != "https") return Optional.empty()
+
+        val localhostMatch = Pattern.compile("^(localhost|127(?:\\.[0-9]+){0,2}\\.[0-9]+|0+(?:\\.0+){0,2}\\.0+|\\[(?:0*:)*?:?0*1?])(?::(\\d+))?\$").matcher(uri.authority)
+
+        if (!localhostMatch.find()) return Optional.empty()
+
+        var address = localhostMatch.group(1)
+        if (address.startsWith('[') && address.endsWith(']')) {
```
mustard-mh, line 65: How could this happen😂

mustard-mh: note: ipv6

mustard-mh: https://pl.kotl.in/ERWpCkfaW this case `"http://[::1]:8080"` is not correct? addr should be `::1` and port `8080`

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/latest/GitpodPortForwardingService.kt
```diff
@@ -0,0 +1,136 @@
...
+    init { start() }
+
+    private fun start() {
+        if (application.isHeadlessEnvironment) return
+
+        observePortsListWhileProjectIsOpen()
+    }
+
+    private fun observePortsListWhileProjectIsOpen() = application.executeOnPooledThread {
+        while (project.lifetime.status == LifetimeStatus.Alive) {
```
mustard-mh, line 45: We don't need `while` and `sleep` here since use observe from grpc below. Just observe one time, and it'll keep respond data.

See also https://github.com/gitpod-io/gitpod/blob/151ed2c0095136f8c604def16b819334378bb3be/components/supervisor/pkg/supervisor/services.go#L206-L210

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/latest/GitpodPortForwardingService.kt
```diff
@@ -0,0 +1,136 @@
...
+        statusServiceStub.portsStatus(portsStatusRequest, portsStatusResponseObserver)
+
+        return completableFuture
+    }
+
+    private fun handlePortStatusResponse(response: Status.PortsStatusResponse) {
+        val portNumberToServedStatusMap = mutableMapOf<Int, Boolean>()
+
+        for (port in response.portsList) {
+            val isServed = (port.exposed != null || port.tunneled != null) && port.served
```
mustard-mh, line 89: Why not check their value here but check with `!= null`?

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/latest/GitpodPortForwardingService.kt
```diff
@@ -0,0 +1,136 @@
...
+
+                val portInfo = ForwardedPortInfo(
+                        hostPort,
+                        RdPortType.HTTP,
+                        FORWARDED_PORT_LABEL,
+                        emptyList(),
+                        portEventsProcessor
+                )
+
+                portForwardingManager.forwardPort(portInfo)
```
mustard-mh, line 126: How about move forward behavior into `portsService.setForwardedPort`? Is it because of only latest can use it? Will we do it in the future (move code)?

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/latest/GitpodPortForwardingService.kt
mustard-mh: If we are going to serve all opened ports, use `port.served` is enough

mustard-mh:
Note that forward to `127.0.0.1` only, sometimes `localhost` will not work because of `ipv6`

mustard-mh:
My self-hosted Gitpod (with gitpod repo and IntelliJ) has no jave intelligence, I review this PR with plain text 🥲, am I miss something? (I have run Java Task in `.gitpod.yml`)

but we can trust with build process 😂

I'll test this PR later today

felladrin:
Thanks for your insights, @mustard-mh!

> My self-hosted Gitpod (with gitpod repo and IntelliJ) has no jave intelligence, I review this PR with plain text 🥲, am I miss something? (I have run Java Task in `.gitpod.yml`)
> 
> but we can trust with build process 😂
> 
> I'll test this PR later today

Hm, I think the autocomplete depends only on having the Gradle Project properly configured on IDEA.
Maybe during the startup, it failed to set up the JDK version for the Backend Plugin Project?
Check on the sidebar if the Gradle menu has 2 projects set up.

image

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/latest/GitpodPortForwardingService.kt
felladrin: Exactly! We will move it to the "common" package when the Stable IDE version is compatible with and a few remaining issues get resolved! ✌️

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodPortsService.kt
felladrin: Gotcha!
I confirm `getLocalHostUriFromHostPort` function is only used for getting the local URI after checking it's being forwarded (by using `isForwarded` function).
Outside this class, we're handling the non-forwarded cases and properly using `https`.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodPortsService.kt
felladrin: Nice catch! That proves this function deserves a unit test. Let me try it.

ℹ️ This function is a translation of this JavaScript function:

https://github.com/gitpod-io/openvscode-server/blob/8047dd89d347c79ca6e686dd1968f496b6da27bf/src/vs/platform/tunnel/common/tunnel.ts#L134-L151

felladrin: @akosyakov, that's a good question from @mustard-mh: why is it removing the last 2 characters instead of removing just the last one (`]`) [here](https://github.com/gitpod-io/openvscode-server/blob/8047dd89d347c79ca6e686dd1968f496b6da27bf/src/vs/platform/tunnel/common/tunnel.ts#L144)?

And wouldn't it be better to return the full hostname, like URL returns?
image

Currently the code is returning `{address: "::", port: 8080}` for `http://[::1]:8080/`.

Should we change it (in Kotlin code from this PR) to return "[::1]"?

felladrin: I've added the unit test 🎉 

They can be run with `./gradlew test` (when inside `ide/jetbrains/backend-plugin` folder) or via IDE UI:

image

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/latest/GitpodPortForwardingService.kt
felladrin: 🤔 That `while` has a different function. It's there to cover the case where the user loses connection temporarily. So while the IDE is opened, it should keep trying to reconnect to Supervisor.

Do you think we can conver this case using only PortStatus?
But even so, we'd need to keep a cooldown before trying to connect again, so I think 1sec interval is ok.
We've been using this logic in all other classes (from JetBrains plugin) that requires waiting for reconnection.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/latest/GitpodPortForwardingService.kt
felladrin: Well observed! Updated! Thank you!

mustard-mh:
I want to try with preview env but when clicking with `Open in IntelliJ` the auth page does not work, just pending and give failed json blow

```
{"error":"Error: Handshake inactivity timeout\nStack: Error: Handshake inactivity timeout\n    at Handshake.<anonymous> (/app/node_modules/mysql/lib/protocol/Protocol.js:160:17)\n    at Handshake.emit (node:events:390:28)\n    at Handshake.emit (node:domain:475:12)\n    at Handshake._onTimeout (/app/node_modules/mysql/lib/protocol/sequences/Sequence.js:124:8)\n    at Timer._onTimeout (/app/node_modules/mysql/lib/protocol/Timer.js:32:23)\n    at listOnTimeout (node:internal/timers:557:17)\n    at processTimers (node:internal/timers:500:7)\n    --------------------\n    at Protocol._enqueue (/app/node_modules/mysql/lib/protocol/Protocol.js:144:48)\n    at Protocol.handshake (/app/node_modules/mysql/lib/protocol/Protocol.js:51:23)\n    at PoolConnection.connect (/app/node_modules/mysql/lib/Connection.js:116:18)\n    at Pool.getConnection (/app/node_modules/mysql/lib/Pool.js:48:16)\n    at Pool.query (/app/node_modules/mysql/lib/Pool.js:202:8)\n    at MySQLStore.query (/app/node_modules/express-mysql-session/index.js:439:33)\n    at MySQLStore.get (/app/node_modules/express-mysql-session/index.js:162:8)\n    at session (/app/node_modules/express-session/index.js:485:11)\n    at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:95:5)\n    at trim_prefix (/app/node_modules/express/lib/router/index.js:323:13)\n    at /app/node_modules/express/lib/router/index.js:284:7\n    at Function.process_params (/app/node_modules/express/lib/router/index.js:341:12)\n    at next (/app/node_modules/express/lib/router/index.js:275:10)\n    at cookieParser (/app/node_modules/cookie-parser/index.js:71:5)\n    at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:95:5)\n    at trim_prefix (/app/node_modules/express/lib/router/index.js:323:13)"}
```

More info:
- Choose latest IntelliJ as IDE
- JetBrains Gateway 2022.2.1 Preview
Build #GW-222.3739.24, built on August 4, 2022
Runtime version: 17.0.3+7-b469.37 aarch64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
macOS 13.0
GC: G1 Young Generation, G1 Old Generation
Memory: 512M
Cores: 8
Metal Rendering is ON
Non-Bundled Plugins:
    io.gitpod.jetbrains.gateway (0.0.1-main.4067)

felladrin:
Ah, I guess it was lack of memory in the server, cause I was also testing it at this moment :)
I'll restart the Preview Environment and you can test it, I won't touch it anymore.

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodPortsService.kt
akosyakov: I don't remember no, but looking at your client it does not seem to be matter. You don't use address anyway.

What you should understand here though, that the same port can be served on ip4 and ip6 at the same time as far as I remember then on client machine depending on the host there should be different ports too. I think it is worth to  create a follow-up issue to consider it, and discuss with JB how to handle such cases. But it should NOT block this PR.

meysholdt (approved):

install/installer/pkg/components/ws-proxy/deployment.go
```diff
@@ -65,7 +65,7 @@ func deployment(ctx *common.RenderContext) ([]runtime.Object, error) {
 		PriorityClassName: common.SystemNodeCritical,
 		Affinity:          common.NodeAffinity(cluster.AffinityLabelWorkspaceServices),
 		TopologySpreadConstraints: []corev1.TopologySpreadConstraint{
-			corev1.TopologySpreadConstraint{
+			{
```
felladrin, line 68: Ah, this change was not mine. Just a moment, let me remove it.
cc @andreafalzetti
/hold

felladrin: Updated!
/unhold

andreafalzetti: I was fixing `gofmt` but then another PR was merged to main to fix it, so I rebased and it fixed the job. I thought this would go away automatically after rebase, sorry :)

felladrin: No worries! ✌️

andreafalzetti:
I've tested a couple of times but I always get the workspace URL instead of localhost. Am I doing something wrong?

1. Started a workspace in https://felladrin-d4177edc98.preview.gitpod-dev.com/workspaces
2. Selected IntelliJ Latest as preferred IDE

I have the latest JetBrains Gateway installed, with the latest gitpod plugin from the marketplace.
Screenshot 2022-08-09 at 11 57 44
Screenshot 2022-08-09 at 11 58 13
Screenshot 2022-08-09 at 11 58 27

Any idea?

felladrin:
> I've tested a couple of times but I always get the workspace URL instead of localhost. Am I doing something wrong?

Ah, well observed! But that's ok :) It's coming from `gp preview` command, so we decided we'll check it in a separate PR [[1](https://github.com/gitpod-io/gitpod/pull/11081#discussion_r935444156)][[2](https://github.com/gitpod-io/gitpod/issues/11849)].

andreafalzetti:
> > I've tested a couple of times but I always get the workspace URL instead of localhost. Am I doing something wrong?
> 
> Ah, well observed! But that's ok :) It's coming from `gp preview` command, so we decided we'll check it in a separate PR [[1](https://github.com/gitpod-io/gitpod/pull/11081#discussion_r935444156)][[2](https://github.com/gitpod-io/gitpod/issues/11849)].

when I click on the “Open Browser” link inside the toast notification, it opens the dialog to confirm, where the URL is visible, and the URL is the full workspace, even after confirm, it opens the full URL (Not localhost), so I am still confused.

felladrin:
Fixed by wrapping the handling of the status of the ports from GRPC in `application.invokeLater` to execute actions with _PortForwardingManager_ in the UI thread. Without it, it was being executed on the same background thread as the GRPC listener. ([More Info](https://plugins.jetbrains.com/docs/intellij/general-threading-rules.html#modality-and-invokelater))

https://github.com/gitpod-io/gitpod/blob/3e940b8ff7866312dcd828909d2c59020c19bce9/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/latest/GitpodPortForwardingService.kt#L75-L77

felladrin:
ℹ️  I'll quickly check it on Preview Environment before setting it ready for review again.

felladrin:
Setting it as Ready for Review again :)

And just to rectify [what I said before](https://github.com/gitpod-io/gitpod/pull/11081#issuecomment-1209344418):
When clicking "Open browser" it will indeed show the dialog box with the localhost URL (which is a different dialog that is displayed when we run `gp preview $(gp url 3000)` on terminal):

image

andreafalzetti (approved):
LGTM, tested and works as expected 🍪

PR #11096, opened 2022-07-02
[installer]: set a stable password for messagebus

### LINKED ISSUE #10459: Use stable value for rabbitmq-erlang-cookie
The Installer is currently generating a random string for the rabbitmq-erlang-cookie secret ([source](https://github.com/gitpod-io/gitpod/blob/main/install/installer/pkg/components/rabbitmq/secret.go#L25)) for every invocation of the Installer which is the source of drift in ArgoCD for us. We've worked around it by telling ArgoCD to ignore the secret when producing diffs, but is there any way we could make it stable between invocations so our customers don't have to do the same workaround?

Internal discussion [here](https://gitpod.slack.com/archives/C01KLC56NP7/p1654248658232499)

MrSimonEmms:
## Description

This sets both the `rabbitmq.auth.password` value and the `messagebus-erlang-cookie` secret with static values. This has the benefit of not reloading the message bus every time a redeployment happens - this makes deployments faster and existing workspaces don't lose connections.

The initial ticket only mentions the cookie secret, but discovered that the [password is hardcoded in the experimental section](https://github.com/gitpod-io/ops/search?q=staticMessagebusPassword) - this doesn't change the functionality of `experiments.common.staticMessagebusPassword` so the Ops repo will still work as expected.

## Related Issue(s)

Fixes #10459

## How to test

Deploy the application

## Release Notes

```release-note
[installer]: set a stable password for messagebus
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/installer/pkg/common/render.go
+++ b/install/installer/pkg/common/render.go
@@ -158,10 +158,7 @@ func (r *RenderContext) generateValues() error {
 		return nil
 	})
 	if messageBusPassword == "" {
-		messageBusPassword, err = RandomString(20)
-		if err != nil {
-			return err
-		}
+		messageBusPassword = "uq4KxOLtrA-QsDTfuwQ-"
 	}
 	r.Values.MessageBusPassword = messageBusPassword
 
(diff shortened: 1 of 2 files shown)


roboquat:

<details>

</details>

mads-hartmann (approved):
Looks good to me, thanks!

corneliusludmann (approved):

PR #11111, opened 2022-07-04
[kots]: add firewall check for the pull registry

### LINKED ISSUE #10244: [Self-Hosted] Preflight check for to check for firewall settings
Some customers have their firewall / proxy setup so that it just swallows invalid connections instead of outright denying it (this can also be called stealth mode, silent block / fail). This shows up as a container image taking forever to download, which is in fact blocked. 

A kots preflight or diagnostic that can tell if gitpod (and specifically the workspace nodes) can get https://eu.gcr.io/v2/  within N seconds would circumvent some of these problems

## Additional context
Internal [slack thread](https://gitpod.slack.com/archives/C01KLC56NP7/p1653471602985009)

MrSimonEmms:
## Description

Adds a simple check to ensure that a network connection can be made to the pull registry. The command `curl https://eu.gcr.io/v2/` does actually (correctly) return an unauthorized error as the check is that it can make an HTTP connection in the timeout. However, it still exits with a code 0.

If it fails to connect within the timeout, it will exit with a non-zero error code.

In airgapped, it runs against the pull repository.

## Related Issue(s)

Fixes #10244

## How to test

Install in a non-airgapped and airgapped environment

## Release Notes

```release-note
[kots]: add firewall check for the pull registry
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/kots/manifests/kots-preflight.yaml
+++ b/install/kots/manifests/kots-preflight.yaml
@@ -44,6 +44,32 @@ spec:
           - '{{repl ConfigOption "store_s3_access_key_id" }}' # S3_ACCESS_KEY_ID
           - '{{repl ConfigOption "store_s3_secret_access_key" }}' # S3_SECRET_ACCESS_KEY
           - '{{repl ConfigOption "store_s3_bucket" }}' # S3_BUCKET_NAME
+    - run:
+        # Check that the pull registry is accessible from the cluster
+        collectorName: ping-registry
+        image: alpine/curl
+        name: ping-registry
+        command:
+          - /bin/sh
+          - -c
+        args:
+          - |
+            CONNECTION="error"
+
+            if [ '{{repl HasLocalRegistry }}' = "true" ];
+            then
+              # Don't test for airgapped
+              CONNECTION="ok"
+            else
+              URL="https://eu.gcr.io/v2/"
+              echo "ping ${URL}"
+              if curl --silent --max-time 5 "${URL}" > /dev/null;
+              then
+                CONNECTION="ok"
+              fi
+            fi
+
+            echo "connection: ${CONNECTION}"
   analyzers:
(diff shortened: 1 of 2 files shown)


corneliusludmann (approved):
Implementation looks good. :+1:  Looking forward to having this released! :rocket: 

---

The only concern of my always paranoid mind is whether the `v2/` endpoint is actually reachable for all supported registries? Is this a common endpoint that is used in the image pull protocol? Or could it be that some registries do not answer properly on _this_ path but only on the actual _full image path_?

The second rather minor question is whether we need to check the status code as well? Or set the `-L` flag to follow redirects? What is, when the registry is behind a proxy? Could it happen that the proxy answers with an okayish status code (that lets curl run successfully) but the actual registry is still not reachable? Maybe in case of a redirect?

Don't know. Maybe I'm overthinking this.

/hold in case you want to test something after my comment. Otherwise, feel free to remove the hold label.

MrSimonEmms:
My understanding based upon the [API spec](https://docs.docker.com/registry/spec/api/) is that it's always on `/v2`. But, as we know from the AWS implementation, there's no guarantee that it's been implemented correctly just because it's in the spec.

For the second point, I don't think we need to follow the `-L` or check the status. In fact, if you run `curl https://eu.gcr.io/v2/`, you'll see that it actually returns a 401 error - the test is checking that something is terminating that call and not timing out.

The other way we could do this is to simply not test airgapped installations. The test is that they can access the pull registry - in an airgapped installation, they'd have had to connect to their pull registry to get the KOTS images anyway so maybe this test doesn't actually add anything to the party.

MrSimonEmms:
@corneliusludmann I've amended the PR so that it doesn't test in an airgapped environment. I'll leave for you to remove the "hold" if/when you're ready

PR #11115, opened 2022-07-04
[supervisor] Refactor dockerd exit logs

aledbf:
## Release Notes
```release-note
NONE
```

### DIFF
--- a/components/supervisor/pkg/supervisor/supervisor.go
+++ b/components/supervisor/pkg/supervisor/supervisor.go
@@ -1443,6 +1443,10 @@ func processesWithParent(ppid int) (map[int]int, error) {
 	return children, nil
 }
 
+var (
+	errSignalKilled = errors.New("signal: killed")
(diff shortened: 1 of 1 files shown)


components/supervisor/pkg/supervisor/supervisor.go
```diff
@@ -1478,7 +1482,7 @@ func socketActivationForDocker(ctx context.Context, wg *sync.WaitGroup, term *te
 			if !ok {
 				return errors.New("cannot find pty")
 			}
-			ptyCtx, cancel := context.WithCancel(context.Background())
+			ptyCtx, cancel := context.WithCancel(ctx)
```
iQQBot, line 1485: I think we did this on purpose, if ptyCtx is based on ctx, then once ctx is done, it will cause the following select to be chosen randomly, so there is a chance that kill will not be triggered, resulting in a deadlock

PR #11131, opened 2022-07-05
Fix workspace location when opening in vscode desktop

### LINKED ISSUE #11101: multi-repo: opening workspace file looses task's terminal window (desktop VSCode)
### Bug description

Using desktop VSCode I have a multi repo.  When it opens in VSCode desktop it only displays the main repo (in the explorer bar).  But I have a VSCode workspace file in the repo which VSCode finds and it offers to open that.  I click "Yes" and the VSCode window refreshes and now the explorer bar shows all the repos in my workspace, as it should. But the task terminal window disappears - I have to open a new one.

_In the past_ with this _same setup_ when I opened the workspace file the task terminal window came back _with the output from the task present_.

This gitpod.yml has _one_ task only, it's terminal window is what I'm talking about.

Unfortunately, I can't pinpoint when this change in behavior happened.  It definitely was the case that it worked _before_ I switched to the ssh remote (v.s. the gitpod remote desktop thing).  But I also think it worked for awhile _after_ I switched to the ssh remote

### Steps to reproduce

With VSCode desktop, open a multi repo workspace, e.g., https://gitpod.io/start/#bakinsbits-gitpodsetupf-x243p43jh48, based on https://gitlab.com/bakins-bits/gitpod-setup-for-bitcoin-core-dev

VSCode desktop opens, and after I agree that I trust this folder and its parents, the (single) `init`+`command` task terminal window is visible, and it asks to open the workspace file `bitcoin.code-workspace` which is at the root of the main repo:

0-with-terminal-window-asking-for-workspace-file

I click "yes" and the window refreshes and I see _no_ terminal window:

1-after-saying-to-open-workspace-window-refreshes-without-terminal

So I execute the menu **View > Terminal** and I see a brand new terminal window, and VSCode knows of no others:

3-after-opening-terminal-pane-get-a-new-terminal

### Workspace affected

bakinsbits-gitpodsetupf-x243p43jh48 (and others)

### Expected behavior

After opening the workspace file I should continue to see an _active_ terminal window that is the task that `.gitpod.yml` ran (`init` and `command` phases).

### Example repository

Snapshot url: https://gitpod.io/#snapshot/479b394b-ee03-4b7d-a065-bf15f52b7a77

Taken right after opening the workspace before it offered to open the VSCode workspace file.  If on restore it doesn't itself offer to open the VSCode workspace file you can do it yourself:  Just navigate to and then click `/workspace/gitpod-setup-for-bitcoin-core-dev/bitcoin.code-workspace` and then it offers you a "Open Workspace" button in the lower-right to click on.

### Anything else?

Could be a gitpod change, could be a VSCode change, not sure.

`gp version` reports `se-db-perf.0`

Gitpod extension (installed in desktop): `v0.0.401
Gitpod Remote extension (installed in desktop): `v0.0.32`

VSCode about box says:

```
Version: 1.68.1
Commit: 30d9c6cd9483b2cc586687151bcbcd635f373630
Date: 2022-06-14T12:48:58.283Z (2 wks ago)
Electron: 17.4.7
Chromium: 98.0.4758.141
Node.js: 16.13.0
V8: 9.8.177.13-electron.0
OS: Windows_NT x64 10.0.22000
```

jeanp413:
## Related Issue(s)

Fixes #11101

## How to test

1. Select vscode desktop in dashboard
1. Open https://gitlab.com/bakins-bits/gitpod-setup-for-bitcoin-core-dev in prev env
2. When vscode desktop connects to workspace check the workspace openup with three folders by default - `bitcoin` `btcdeb` `gitpod-setup-for-bitcoin-core-dev`

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/ide/code-desktop/status/main.go
+++ b/components/ide/code-desktop/status/main.go
@@ -63,10 +63,15 @@ func main() {
 		}
 		queryString := string(b)
 
+		workspaceLocation := wsInfo.GetWorkspaceLocationFile()
+		if workspaceLocation == "" {
+			workspaceLocation = wsInfo.GetWorkspaceLocationFolder()
+		}
+
 		link := url.URL{
 			Scheme:   schema,
 			Host:     "gitpod.gitpod-desktop",
-			Path:     wsInfo.CheckoutLocation,
+			Path:     workspaceLocation,
 			RawQuery: url.QueryEscape(queryString),
 		}
 
--- a/components/ide/code/codehelper/main.go
+++ b/components/ide/code/codehelper/main.go
@@ -41,15 +41,11 @@ func main() {
 	log.Init(ServiceName, Version, true, false)
 	startTime := time.Now()
 
-	log.Info("wait until content available")
-
-	// wait until content ready
-	contentStatus, wsInfo, err := resolveWorkspaceInfo(context.Background())
-	if err != nil || wsInfo == nil || contentStatus == nil || !contentStatus.Available {
-		log.WithError(err).WithField("wsInfo", wsInfo).WithField("cstate", contentStatus).Error("resolve workspace info failed")
+	wsInfo, err := resolveWorkspaceInfo(context.Background())
+	if err != nil || wsInfo == nil {
+		log.WithError(err).WithField("wsInfo", wsInfo).Error("resolve workspace info failed")
 		return
 	}
-	log.WithField("cost", time.Now().Local().Sub(startTime).Milliseconds()).Info("content available")
 
 	// code server args install extension with id
 	args := []string{}
@@ -117,8 +113,8 @@ func main() {
 	}
 }
 
-func resolveWorkspaceInfo(ctx context.Context) (*supervisor.ContentStatusResponse, *supervisor.WorkspaceInfoResponse, error) {
-	resolve := func(ctx context.Context) (contentStatus *supervisor.ContentStatusResponse, wsInfo *supervisor.WorkspaceInfoResponse, err error) {
+func resolveWorkspaceInfo(ctx context.Context) (*supervisor.WorkspaceInfoResponse, error) {
+	resolve := func(ctx context.Context) (wsInfo *supervisor.WorkspaceInfoResponse, err error) {
 		supervisorAddr := os.Getenv("SUPERVISOR_ADDR")
 		if supervisorAddr == "" {
 			supervisorAddr = "localhost:22999"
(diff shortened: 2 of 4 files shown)


akosyakov:
@jeanp413 It does not seem to work for me. It always opens `/` for me
Screenshot 2022-07-05 at 08 42 51

mustard-mh (commented):
I can't see any differences between `prev env` and `prod env` (I use `Open in VS Code` command in browser code)

BTW. I found a bug (`prod env`) that after a close Desktop VSCode, browser code's terminals will be terminated...Maybe I should file an issue?

akosyakov:
> BTW. I found a bug (prod env) that after a close Desktop VSCode, browser code's terminals will be terminated...Maybe I should file an issue?

It is known and fixed, but needs a new release of VS Code Desktop, please see https://github.com/gitpod-io/gitpod/issues/11101#issuecomment-1173548202

components/ide/code-desktop/status/main.go
```diff
@@ -63,10 +63,15 @@ func main() {
 		}
 		queryString := string(b)
 
+		workspaceLocation := wsInfo.GetWorkspaceLocationFile()
```
akosyakov, line 66: I think we should align with https://github.com/gitpod-io/openvscode-server/blob/57d3840d9d89487462a2004551d58563022a7324/extensions/gitpod-shared/src/features.ts#L405

workspace folder can be also empty then we should fall back to checkout location

jeanp413:
> I can't see any differences between prev env and prod env (I use Open in VS Code command in browser code)

@mustard-mh  You need to select vscode desktop in dashboard, using vscode web uses a different logic in gitpod-web

components/ide/code-desktop/status/main.go
jeanp413: fallback is already handled [here](https://github.com/gitpod-io/gitpod/blob/c43a6c9859754a06c425d8ec490d54e9978a8a6a/components/server/src/workspace/workspace-starter.ts#L1390) if I'm not mistaken

mustard-mh:
> @mustard-mh You need to select vscode desktop in dashboard, using vscode web uses a different logic in gitpod-web

Not works too

mustard-mh:
And why this PR can fix issue https://github.com/gitpod-io/gitpod/issues/11101 ?

mustard-mh:
We should remove `contentReady` in JetBrains too

https://github.com/gitpod-io/gitpod/blob/2c5cd4cd885f87a0c0d9948eb9cde477b545aa17/components/ide/jetbrains/image/status/main.go#L70-L77

mustard-mh:
Our docs about `workspaceLocation` are wrong [.gitpod.yml ref](https://www.gitpod.io/docs/references/gitpod-yml#workspacelocation) and [workspace location doc](https://www.gitpod.io/docs/checkout-location) we should change them. It means the pointer of `.code-workspace` file as [our schema](https://github.com/gitpod-io/gitpod/blob/1690eee42e4dd9e2aaed925a298dd7f0f98d403b/components/gitpod-protocol/data/gitpod-schema.json?q=repo%3Agitpod-io%2Fgitpod+workspaceLocation+#L168-L171) said

@loujaybee @jeanp413

mustard-mh (changes requested):
Nice job JP 🎉  Need to change https://github.com/gitpod-io/gitpod/pull/11131#issuecomment-1177238524

I tested this PR in prev env with `vscode desktop` selected, it works as expected - open up with three folders.

|Prod| Prev Env|
|-|-|
|image | image|

mustard-mh:
Also tested with JetBrains and it didn't break

image

PR #11003, opened 2022-06-29
Update k8s dependencies to v0.24.2

aledbf:
## Description

Update go dependencies

## Release Notes
```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

PR #11135, opened 2022-07-05
Revert "Update k8s dependencies to v0.24.2"

jenting:
## Description

Reverts gitpod-io/gitpod#11003

Bumping the k8s dependencies to v0.24.2, when the installer installs gitpod, it failed at
```console
error validating data: ValidationError(NetworkPolicy): unknown field "status" in io.k8s.api.networking.v1.NetworkPolicy
```

The root cause is that the NetworkPolicy status introduces from v1.24 (`kubectl explain networkpolicy`). However, our environment is still running v1.23.
Instead of asking the user manually remove the `status` field as we did before for [StatefulSet status](https://github.com/gitpod-io/gitpod/tree/main/install/installer#error-validating-statefulsetstatus), it'd be good for the installer to auto remove the `status` field.

## Release Notes
```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


geropl (approved):
:+1:

iQQBot (commented):
Why need revert whole PR? you can revert 1 commit

iQQBot (approved):

jenting:
> We can choose to remove all status items at the end of the yaml generation

Yes, we discussed in the incident channel, and we would like the installer to remove the status items rather than we ask the user to manually remove the status field.
After that, this approach applies to any environment 💪

mads-hartmann:
Preview environment works, removing the hold

PR #11138, opened 2022-07-05
Align access level of Workspaces of type "prebuild" with Prebuilds

### LINKED ISSUE #11127: Align permissions for Workspace-like GuardedResources so people can access prebuilds in all scenarios
So far in some cases, you might be presented with this screen when waiting for a prebuild:

We should ensure that we share the same logic between:
 a) the places where we decided someone is eligible to receive a prebuild with
 b) where we decide that people are able to access the required resources

image

geropl:
## Description

Align access level of Workspaces of type "prebuild" with Prebuilds

:information_source: It make sense to review this PR per commit:
 - the 1st commit fixes the test setup (and a bug)
 - the 2nd commit is the actual fix for this issue

## Related Issue(s)

Fixes #11127 

## How to test
 - `cd components/server`
 - `yarn test`

## Release Notes

```release-note
align access level of Workspaces of type "prebuild" with Prebuilds
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/server/src/auth/resource-access.spec.ts
+++ b/components/server/src/auth/resource-access.spec.ts
@@ -782,6 +782,43 @@ class TestResourceAccess {
                 teamRole: "owner",
                 expectation: true,
             },
+            // prebuild workspace with repo access
+            {
+                name: "prebuild workspace get owner",
+                resourceKind: "workspace",
+                workspaceType: "prebuild",
+                isOwner: true,
+                teamRole: undefined,
+                repositoryAccess: true,
+                expectation: true,
+            },
+            {
+                name: "prebuild workspace get other",
+                resourceKind: "workspace",
+                workspaceType: "prebuild",
+                isOwner: false,
+                teamRole: undefined,
+                repositoryAccess: true,
+                expectation: true,
+            },
+            {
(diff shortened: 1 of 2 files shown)


components/server/src/auth/resource-access.spec.ts
```diff
@@ -782,6 +782,43 @@ class TestResourceAccess {
...
+                teamRole: undefined,
+                repositoryAccess: true,
+                expectation: true,
+            },
+            {
+                name: "prebuild workspace get other",
+                resourceKind: "workspace",
+                workspaceType: "prebuild",
+                isOwner: false,
+                teamRole: undefined,
```
laushinka, line 800: What is this case where the user is not the owner but also doesn't have a team?

geropl: They have read access to the repository (see one line :point_down: ).

That's the qualifier based on which we share access to prebuilds, and thus should be the same for workspaces (at least for those of type `prebuild`). :+1: 

This fixes the UI issue we see currently.

laushinka: Aha, makes sense.

PR #11158, opened 2022-07-05
ws-manager: Replace backup/restore success with total metric

ArthurSens:
## Description

Hey Workspace-team!

We noticed that you want to level up your game towards success-criteria-related metrics, and there is momentum in the company to start adopting SLOs to get the right metrics.

We from the platform-team are being a little more proactive regarding implementing SLOs and while trying to get the first example up and running, we noticed that your metrics for workspace backups and workspace restores are a bit out of best practices for SLOs with Prometheus 😬 

When implementing success ratio SLOs, we usually look for metrics that represent "total amount of requests" and "amount of failed requests". So the calculation looks like this: `backup error ratio = amount of backup with errors / total amount of backups`

Would it be fine if we change the current metrics to be more SLO-friendly?

_PS: I also simplified the nested ifs by using `WithLabelValues` instead of `GetMetricWithLabelValues` as well_

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/components/ws-manager/pkg/manager/metrics.go
+++ b/components/ws-manager/pkg/manager/metrics.go
@@ -47,9 +47,9 @@ type metrics struct {
 	// Counter
 	totalStartsCounterVec         *prometheus.CounterVec
 	totalStopsCounterVec          *prometheus.CounterVec
-	totalBackupSuccessCounterVec  *prometheus.CounterVec
+	totalBackupCounterVec         *prometheus.CounterVec
 	totalBackupFailureCounterVec  *prometheus.CounterVec
-	totalRestoreSuccessCounterVec *prometheus.CounterVec
+	totalRestoreCounterVec        *prometheus.CounterVec
 	totalRestoreFailureCounterVec *prometheus.CounterVec
 
 	// Gauge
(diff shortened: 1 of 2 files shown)


components/ws-manager/pkg/manager/metrics.go
```diff
@@ -111,7 +111,7 @@ func newMetrics(m *Manager) *metrics {
 			Name:      "workspace_stops_total",
 			Help:      "total number of workspaces stopped",
 		}, []string{"reason", "type", "class"}),
-		totalBackupSuccessCounterVec: prometheus.NewCounterVec(prometheus.CounterOpts{
+		totalBackupCounterVec: prometheus.NewCounterVec(prometheus.CounterOpts{
 			Namespace: metricsNamespace,
 			Subsystem: metricsWorkspaceSubsystem,
 			Name:      "workspace_backups_success_total",
```
sagor999, line 117: should the name of the metric be updated as well then?

components/ws-manager/pkg/manager/metrics.go
```diff
@@ -123,7 +123,7 @@ func newMetrics(m *Manager) *metrics {
 			Name:      "workspace_backups_failure_total",
 			Help:      "total number of workspace backups failure",
 		}, []string{"type", "class"}),
-		totalRestoreSuccessCounterVec: prometheus.NewCounterVec(prometheus.CounterOpts{
+		totalRestoreCounterVec: prometheus.NewCounterVec(prometheus.CounterOpts{
 			Namespace: metricsNamespace,
 			Subsystem: metricsWorkspaceSubsystem,
 			Name:      "workspace_restores_success_total",
```
sagor999, line 129: same here.

components/ws-manager/pkg/manager/metrics.go
ArthurSens: very good catch!

sagor999 (approved):
LGTM, thank you!!

/hold
adding hold, can you please double check that we are not using that metric in any of grafana dashboards? As in that case they will become broken. I don't think we do, but if you could double check that would be appreciated. Feel free to remove the hold.
:heart:

ArthurSens:
I've looked at [Workspace success criteria](https://grafana.gitpod.io/d/jgjwvIc7k/success-criteria?orgId=1), and couldn't find anything 🤔

components/ws-manager/pkg/manager/monitor.go
```diff
@@ -1244,13 +1230,13 @@ func (m *Monitor) finalizeWorkspaceContent(ctx context.Context, wso *workspaceOb
 		BackupComplete: true,
 		GitStatus:      gitStatus,
 	}
+
+	if doBackup {
```
jenting, line 1234: We need to consider `doSnapshot` as well. @sagor999 Correct?
```go
if doBackup || doSnapshot {
```

sagor999: yeah, you are right!

ArthurSens: Thanks for catching that, I've added `doSnapshot` in the conditional 🙂

jenting (approved):
LGTM, thank you ❤️

jenting:
We only use the `workspace_backups_failure_total` in the [ws-daemon alert rules](https://github.com/gitpod-io/gitpod/blob/7800a21/operations/observability/mixins/workspace/rules/components/ws-daemon/alerts.libsonnet).

/unhold

PR #11194, opened 2022-07-07
[ws-manager-bridge] Use WS update status version

csweichel:
## Description
This PR makes ws-manager-bridge respect and record the status version reported by ws-manager.
Prior this behaviour was only used for prebuilds. We need to extend this behaviour to get rid of a host of heuristics which aim to identify the latest update, and to properly support the public API.

If I understand correctly, this change renders the prebuild status check ineffective because it preempts this check. If this is really the case, we want to consider removing the prebuild status check.

cc @easyCZ 

## How to test
1. Start and stop workspaces - things should work as normal
2. Inspect the database/instance updates. You should see the status.version increment

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-protocol/src/workspace-instance.ts
+++ b/components/gitpod-protocol/src/workspace-instance.ts
@@ -72,6 +72,11 @@ export interface WorkspaceInstance {
 
 // WorkspaceInstanceStatus describes the current state of a workspace instance
 export interface WorkspaceInstanceStatus {
+    // version is the current version of the workspace instance status
+    // Note: consider this value opague. The only guarantee given is that it imposes
+    //       a partial order on status updates, i.e. a.version > b.version -> a newer than b.
+    version?: number;
+
     // phase describes a high-level state of the workspace instance
     phase: WorkspaceInstancePhase;
 
--- a/components/ws-manager-bridge/src/bridge.ts
+++ b/components/ws-manager-bridge/src/bridge.ts
@@ -227,6 +227,7 @@ export class WorkspaceManagerBridge implements Disposable {
         const span = TraceContext.startSpan("handleStatusUpdate", ctx);
         span.setTag("status", JSON.stringify(filterStatus(status)));
         span.setTag("writeToDB", writeToDB);
+        span.setTag("statusVersion", status.statusVersion);
         try {
             // Beware of the ID mapping here: What's a workspace to the ws-manager is a workspace instance to the rest of the system.
             //                                The workspace ID of ws-manager is the workspace instance ID in the database.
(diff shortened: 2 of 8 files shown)


easyCZ:
Change looks good to me. 

To clarify my understanding, each WorkspaceInstance has a status version which allows us to apply partial ordering on the events. However, each status of a WorkspaceInstance is independent of another WorkspaceInstance for the same Workspace. So the unique partial order on Workspace Status is through the tuple of (WorkspaceInstanceID, status.version). Is the above correct?

components/gitpod-protocol/src/workspace-instance.ts
```diff
@@ -72,6 +72,11 @@ export interface WorkspaceInstance {
 
 // WorkspaceInstanceStatus describes the current state of a workspace instance
 export interface WorkspaceInstanceStatus {
+    // version is the current version of the workspace instance status
+    // Note: consider this value opague. The only guarantee given is that it imposes
+    //       a partial order on status updates, i.e. a.version > b.version -> a newer than b.
+    version: number;
```
geropl, line 78: This should be `version?` for backwards compatibility.

I'm fine with promoting at some point, but let's stick with optional for now to avoid `undefined` issues.

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -252,6 +253,14 @@ export class WorkspaceManagerBridge implements Disposable {
                 return;
             }
 
+            if (instance.status.version > 0 && instance.status.version >= status.statusVersion) {
+                // We've gotten an event which is older than one we've already processed. We shouldn't process the stale one.
+                span.setTag("statusUpdate.staleEvent", true);
+                this.prometheusExporter.recordStaleStatusUpdate();
+                log.debug(ctx, "Stale status update received, skipping.");
+                return;
```
geropl, line 261: :+1:

components/gitpod-protocol/src/workspace-instance.ts
geropl: /hold

to discuss this.

csweichel: done

csweichel: /hold cancel

components/ws-manager-bridge/src/bridge.ts
geropl: This request might sound strange, but: This interface is critical, and has shown in the past to quickly break in subtle ways. :lotus_position:

csweichel:
> Change looks good to me.
> 
> To clarify my understanding, each WorkspaceInstance has a status version which allows us to apply partial ordering on the events. However, each status of a WorkspaceInstance is independent of another WorkspaceInstance for the same Workspace. So the unique partial order on Workspace Status is through the tuple of (WorkspaceInstanceID, status.version). Is the above correct?

`(WorkspaceInstanceID, status.version)` is on the safe side, and `status.version` alone should also suffice. The status version is built from a hybrid logic clock which is (most of the time monotonically increasing)

components/ws-manager-bridge/src/bridge.ts
csweichel: I thought about this, too - and deemed the risk ok because we're doing this kind of check for prebuilds for a while now.

But yeah, the chance of breaking stuff is high here.
Return is removed.

csweichel: /hold

csweichel: /hold cancel

geropl (approved):
👍

Issue #11108, opened 2022-07-04
Automatically add contexts for preview envs


meysholdt:
## Is your feature request related to a problem? Please describe
When launching a new preview env, the k8s context should be added to the workspace automatically. 

## Describe the behaviour you'd like
in https://github.com/gitpod-io/gitpod/blob/0708df5e4dbe1634eab4dc7dbd8c6e6f0effb53c/.gitpod.yml#L35-L38
start something like `previewctl install-context --watch` so that it keeps running in the background forever and keeps the context in sync with the current previe env.

I belive this also requires a change in`previewctl install-context --watch` to run forever instead of just running until a context has been installed successfully. 

see also [the RFC](https://www.notion.so/gitpod/A-Go-based-CLI-for-interacting-with-Preview-Environments-1834aa90bc104a0b836dd523e22f9e93#fcadb7f6ed4941a4b8af5f1b1e99356b).

PR #11202, opened 2022-07-07
Keep preview environment context in sync

mads-hartmann:
## Description

This PR implements the `--watch` option and uses it in `.gitpod.yaml`. I considered the following requirements:

- It should deal with users changing branches
- It should deal with users using "with-clean-slate-deployment"; that is, it should re-install the kubectx if the underlying VM is recreated.

### Implementation

This changes the behaviour of "--watch" so that it runs forever and polls the current branch and VM periodically to see if a new preview environment context should be installed.

### Alternative implementation

If we wanted to avoid polling could've created a channel for changes to the preview environment. For changes to the VM we could (I think) have used an Informer. Branch changes would be a bit more tricky though but might be achievable using a git hook that sends a signal to the process. Or if the current branch is stored in the filesystem we might be able to be notified on changes to that file. 

I went with polling as that seemed like a smaller 🛹 

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/11108

## How to test

```
# Run manually during development (or start a workspace off this branch as gitpod.yaml runs --watch
go run main.go install-context --watch

# To test switching branches:
# In another terminal
# change between a few different branches. See that it installs the context
git checkout main
# wait
kubectx
git checkout - 
# wait
kubectx

# To test re-creating the VM
werft run github -a with-clean-slate-deployment=true

```

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
N/A

### DIFF
--- a/dev/preview/previewctl/cmd/install_context.go
+++ b/dev/preview/previewctl/cmd/install_context.go
@@ -19,25 +19,53 @@ var (
 
 func installContextCmd(logger *logrus.Logger) *cobra.Command {
 
+	// Used to ensure that we only install contexts
+	var lastSuccessfulPreviewEnvironment *preview.Preview = nil
+
+	install := func(timeout time.Duration) error {
+		p, err := preview.New("", logger)
+
+		if err != nil {
+			return err
+		}
+
+		if lastSuccessfulPreviewEnvironment != nil && lastSuccessfulPreviewEnvironment.Same(p) {
+			logger.Infof("The preview envrionment hasn't changed")
+			return nil
+		}
+
+		err = p.InstallContext(true, timeout)
+		if err == nil {
+			lastSuccessfulPreviewEnvironment = p
+		}
+		return err
+	}
+
 	cmd := &cobra.Command{
 		Use:   "install-context",
 		Short: "Installs the kubectl context of a preview environment.",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			p, err := preview.New(branch, logger)
-			if err != nil {
-				return err
-			}
 
-			err = p.InstallContext(watch, timeout)
-			if err != nil {
-				logger.WithFields(logrus.Fields{"err": err}).Fatal("Failed to install context.")
+			if watch {
+				for range time.Tick(15 * time.Second) {
+					// We're using a short timeout here to handle the scenario where someone switches
+					// to a branch that doens't have a preview envrionment. In that case the default
+					// timeout would mean that we would block for 10 minutes, potentially missing
+					// if the user changes to a new branch that does that a preview.
+					err := install(30 * time.Second)
+					if err != nil {
+						logger.WithFields(logrus.Fields{"err": err}).Info("Failed to install context. Trying again soon.")
+					}
+				}
+			} else {
+				return install(timeout)
 			}
 
 			return nil
 		},
 	}
 
-	cmd.Flags().BoolVar(&watch, "watch", false, "If wait is enabled, previewctl will keep trying to install the kube-context every 30 seconds.")
-	cmd.Flags().DurationVarP(&timeout, "timeout", "t", 10*time.Minute, "Timeout for the watch flag.")
+	cmd.Flags().BoolVar(&watch, "watch", false, "If watch is enabled, previewctl will keep trying to install the kube-context every 15 seconds.")
+	cmd.Flags().DurationVarP(&timeout, "timeout", "t", 10*time.Minute, "Timeout before considering the installation failed")
 	return cmd
 }
--- a/dev/preview/previewctl/pkg/k8s/vm.go
+++ b/dev/preview/previewctl/pkg/k8s/vm.go
@@ -54,33 +54,22 @@ func getVMInstanceIPAddress(vm virtv1.VirtualMachineInstance) (string, error) {
 }
 
 func parseVMInstanceIPAddress(obj *unstructured.Unstructured) (string, error) {
-	var vm virtv1.VirtualMachineInstance
-	err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj.UnstructuredContent(), &vm)
+	var vmi virtv1.VirtualMachineInstance
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj.UnstructuredContent(), &vmi)
 	if err != nil {
 		return "", errors.Wrap(err, ErrGettingVMIpAddress.Error())
 	}
 
-	return getVMInstanceIPAddress(vm)
+	return getVMInstanceIPAddress(vmi)
 }
 
 func (c *Config) GetVMStatus(ctx context.Context, name, namespace string) error {
-	_, err := c.coreClient.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
-	if err != nil && !kerrors.IsNotFound(err) {
+	vmi, err := c.getVMI(ctx, name, namespace)
+	if err != nil {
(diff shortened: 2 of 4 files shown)


dev/preview/previewctl/pkg/k8s/vm.go
```diff
@@ -163,3 +162,29 @@ func (c *Config) GetVMs(ctx context.Context) ([]string, error) {
...
+func (c *Config) getVMI(ctx context.Context, name, namespace string) (*virtv1.VirtualMachineInstance, error) {
+	_, err := c.coreClient.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
+	if err != nil && !kerrors.IsNotFound(err) {
+		return nil, errors.Wrap(ErrVmNotReady, err.Error())
+	}
+
+	if kerrors.IsNotFound(err) {
+		c.logger.Infof("namespace [%s] not found", namespace)
+		return nil, errors.Wrap(ErrVmNotReady, err.Error())
+	}
```
ArthurSens, line 175: I failed to understand what the first if is doing 🤔 

Your conditionals are contradictory but returning the same errors? 🤔

dev/preview/previewctl/pkg/preview/preview.go
```diff
@@ -55,18 +60,21 @@ func New(branch string, logger *logrus.Logger) (*Preview, error) {
...
+		namespace:       fmt.Sprintf("preview-%s", GetName(branch)),
+		name:            GetName(branch),
+		kubeClient:      harvesterConfig,
+		logger:          logEntry,
+		vmiCreationTime: nil,
 	}, nil
 }
 
-func (p *Preview) InstallContext(watch bool, timeout time.Duration) error {
+func (p *Preview) InstallContext(wait bool, timeout time.Duration) error {
```
ArthurSens, line 72: the flag is called `watch` right? Any reason to rename it to `wait`?

dev/preview/previewctl/pkg/preview/preview.go
```diff
@@ -55,18 +60,21 @@ func New(branch string, logger *logrus.Logger) (*Preview, error) {
...
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
+	p.logger.WithFields(logrus.Fields{"branch": p.branch, "timeout": timeout}).Infof("Installing context")
```
ArthurSens, line 76:
```suggestion
	p.logger.WithFields(logrus.Fields{"timeout": timeout}).Infof("Installing context")
```

On line 55 we're already adding the branch to all logs from this logger, there is no need to add it here 🙂

ArthurSens (commented):
Hey Mads, thanks for picking this up!

I like the decision of watching the VMI creation timestamp as the information we use to decide if a preview should be re-installed!

What I didn't understand was the need to implement the watch loops 🤔. 


@vulkoingim could probably elaborate more here, but my assumption is that when using informers (which we are), we're subscribing to specific events from kubernetes. 

The events we're subscribed to are ADD and UPDATE from virtualmachines, but we decide to quit once we successfully install a context. Couldn't we just not quit, but keep watching for UPDATE instead?

dev/preview/previewctl/pkg/k8s/vm.go
mads-hartmann: This is cut-pasted as part of the refactoring to extract shared code into a getVMI function so I didn't look at it too closely, but the only difference I see is the additional logging being done ☺️ I'm inclined to keep it as is and we can return to it when Aleks is back from vacation.

dev/preview/previewctl/pkg/preview/preview.go
mads-hartmann: Thanks, I have accepted the change ✅

dev/preview/previewctl/pkg/preview/preview.go
mads-hartmann: I renamed it to better reflect what it meant for the function. The current implementation would just wait for a single VM to become ready and then install the context. It didn't "watch" for changes to the VM after or changes to the branch after the first initial success.

I will have a look and see if I can extend this function to do that watching as you suggest in your PR comment - if I can it makes sense to keep the name as "watch" instead of "wait" ☺️

mads-hartmann:
> What I didn't understand was the need to implement the watch loops 

The "outer" watch loop serve two purposes

1. Detecting if the VM has changed (e.g. was deleted and a new one created)
2. Detecting if the branch has been changed

We could most likely fix (1) by changing `preview.InstallContext` so that it keeps listening for changes to the VM and proxy. But we can't add (2) without a larger refactoring. A "preview" (the struct) is currently associated with a single branch (which I think makes sense) so if we wanted to use that loop to install contexts for VMs from other branches it would get weird rather quickly.

The implementation in this branch has the following separation of concerns:

1. `preview.InstallContext` is responsbile for installing a single context for a single VM. It implements a `wait` behaviour so it can be called even if the VM is not ready when you invoke it.
2. The install context command implements a watch loop that monitors for changes to the current branch or VM and invokes `preview.InstallContext` if anything changed

We could improve it slightly by moving the "VM was changed" logic into "preview.InstallContext" so that the outer loop only looks for changes to the branch. I can do a quick time-boxed attempt at that and see if the result is nicer ☺️

dev/preview/previewctl/cmd/install_context.go
```diff
@@ -19,25 +19,53 @@ var (
...
+			return nil
+		}
+
+		err = p.InstallContext(true, timeout)
+		if err == nil {
+			lastSuccessfulPreviewEnvironment = p
+		}
+		return err
+	}
+
```
ArthurSens, line 43: This part looks a bit like an anti-pattern for the pattern we established 😬 

To keep every logic testable, I believe we that code inside the `cmd` package should do nothing more than declare the CLI commands and calling functions from other packages.

If we move this part to a separate function the `preview` package, we can create tests for it 🙂

dev/preview/previewctl/cmd/install_context.go
```diff
@@ -19,25 +19,53 @@ var (
...
+					// to a branch that doens't have a preview envrionment. In that case the default
+					// timeout would mean that we would block for 10 minutes, potentially missing
+					// if the user changes to a new branch that does that a preview.
+					err := install(30 * time.Second)
+					if err != nil {
+						logger.WithFields(logrus.Fields{"err": err}).Info("Failed to install context. Trying again soon.")
+					}
+				}
+			} else {
+				return install(timeout)
```
ArthurSens, line 61: Same as above 😅

ArthurSens (approved):
As agreed in our team sync, let's get this merged since the PR solves the problem!

I just found a new problem here, but feel free to leave this to another PR if you prefer. We should be including any logic in the `cmd` package, all it is supposed to do is initialize commands and call functions 😬 

Adding the /hold label so you can decide on tackling this now or in another PR

mads-hartmann:
@ArthurSens Thanks! I'll merge this and follow-up on this in another PR

1. Move logic to `cmd` package
2. See if it makes sense to only care about git in the "outer loop" and extend the inner-loop to deal with VMs being recreated

If I don't manage to get a PR open before I leave for the week I'll create an issue about this instead ☺️

mads-hartmann:
@ArthurSens haha, updated the comment to avoid confusion

mads-hartmann:
Follow-up issue created as I didn't get to it this week https://github.com/gitpod-io/ops/issues/3357

Issue #10328, opened 2022-05-29
Show usage details in usage view


easyCZ:
> The usage view needs to show the details about who did run what workspace when, so that when in doubt users understand where and why the cost has been created.

> Also such a view provides teams a great insight of how Gitpod is used and adopted within their team.

Note: The usage view should also be available when there is no billing configured, e.g. to see personal usage against your free credit, or to see self-hosted team usage.

- [x] #11257
- [x] Update components with real API
- [x] #11285

jankeromnes:
Thought: We can show raw usage per Project, per User or per Team. Which one do we want here?

I guess we'll want per User first (for the skateboard). Later, we can also show per Team (once we have delegated billing). Per Project could also be interesting, but I'd leave it out of scope for now.

easyCZ:
Excellent observation. I'd say go for the simplest one initially and we cut additional tickets for the subsequent ones.

jankeromnes:
FYI, sharing an exciting early draft concept from @gtsiolis of what the Usage UI could eventually look like:

<img width="821" alt="Screenshot 2022-06-24 at 16 21 51" src="https://user-images.githubusercontent.com/599268/175555635-683f3556-87a6-4ea5-8429-74bd1cbb9151.png">

(I guess it's sort of off-topic here, since this issue calls for just showing "raw" usage data, but I think it's still helpful to give an idea of approximately what we'll want to show eventually. 💭)

svenefftinge:
The draft is more like an invoice with aggregated usage. The usage view needs to show the details about who did run what workspace when, so that when in doubt users understand where and why the cost has been created.
Also such a view provides teams a great insight of how Gitpod is used and adopted within their team.

I see this as a first skateboard towards full blown Usage analytics, where teams can also get metrics that help understand how the team is spending it's time on top of this data.

In that light, let's remove the billing bits from the usage (besides the number of used credits) and show the list of sessions where for each session we show:
 - workspace id
 - user (shows `prebuild` when it's a prebuild, as 'user' is not relevant here)
 - project/repo
 - Start (absolute time stamp to the minute)
 - Duration
 - Credits
 - (optional) session id (workspace instance id) - to easily reference a session

FWIW I don't believe we need to make a decision if we start with user's usage or team's usage as both views would be more or less the same. In case I missed something it is clearly team usage that would have the priority IMO.

jldec:
[...]

PR #11208, opened 2022-07-07
Implement skateboard Usage view with dummy data

### LINKED ISSUE #11257: Have skateboard of table with dummy data

laushinka:
## Description
Skateboard of table of billable sessions with dummy data behind a feature flag.
We didn't have a design for this skateboard table (which will be updated according to the actual design), so the view here was my decision based on a combination of [this](https://gitpod.slack.com/archives/C01KPEPQYE6/p1657289918794709?thread_ts=1657137000.465299&cid=C01KPEPQYE6) and what dummy data we agreed on.

Screenshot 2022-07-10 at 21 37 51

## Related Issue(s)

Fixes #11257 
I didn't list #10328 because this is only a skateboard of it - so I added this PR as a task under it.

## How to test
1. Join team "Gitpod" with this [link](https://laushinka-7c7be98297.preview.gitpod-dev.com/teams/join?inviteId=d0842f9c-4a3a-48d0-b234-9453fd3afd80).
2. Create any team.
3. Go to the team's Settings.
4. Click on "Usage".

## Release Notes

```release-note
Users can see their billable sessions.
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/dashboard/src/teams/TeamUsage.tsx
+++ b/components/dashboard/src/teams/TeamUsage.tsx
@@ -1,0 +1,115 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { useContext, useEffect, useState } from "react";
+import { Redirect, useLocation } from "react-router";
+import { PageWithSubMenu } from "../components/PageWithSubMenu";
+import { getCurrentTeam, TeamsContext } from "./teams-context";
+import { getTeamSettingsMenu } from "./TeamSettings";
+import { PaymentContext } from "../payment-context";
+import { getGitpodService } from "../service/service";
+import { BillableSession, BillableWorkspaceType } from "@gitpod/gitpod-protocol/lib/usage";
+import { Item, ItemField, ItemsList } from "../components/ItemsList";
+import moment from "moment";
+import Property from "../admin/Property";
+import Arrow from "../components/Arrow";
+
+function TeamUsage() {
+    const { teams } = useContext(TeamsContext);
+    const { showPaymentUI, showUsageBasedUI } = useContext(PaymentContext);
+    const location = useLocation();
+    const team = getCurrentTeam(location, teams);
+    const [billedUsage, setBilledUsage] = useState<BillableSession[]>([]);
+
+    useEffect(() => {
+        if (!team) {
+            return;
+        }
+        (async () => {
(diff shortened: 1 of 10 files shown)


components/gitpod-protocol/src/usage.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    // Relevant for workspace type. When prebuild, shows "prebuild"
+    userId?: string;
+    teamId?: string;
+
+    instanceId: string;
+
+    workspaceId: string;
+
+    workspaceType: WorkspaceType;
```
andrew-farries, line 21: This type includes a "probe" type - not sure what that's for.

Is it work typing this as `Omit<WorkspaceType, "probe">` ?

components/dashboard/src/teams/TeamUsage.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    if (!showUsageBasedUI) {
+        return <Redirect to="/" />;
+    }
+
+    const getType = (type: string) => {
+        if (type === "regular") {
+            return "Workspace";
+        }
+        return "Prebuild";
```
andrew-farries, line 45: `type` is typed as `WorkspaceType` in `gitpod-protocol` but here it's just a `string`. Can we make this type more precise here too?

andrew-farries (changes requested):

components/gitpod-protocol/src/usage.ts
laushinka: Good point! 🙏🏽

components/dashboard/src/teams/TeamUsage.tsx
laushinka: Thanks for the catch 💪🏼

components/gitpod-protocol/src/usage.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    // Relevant for workspace type. When prebuild, shows "prebuild"
+    userId?: string;
+    teamId?: string;
+
+    instanceId: string;
+
+    workspaceId: string;
+
+    workspaceType: Omit<WorkspaceType, "probe">;
```
andrew-farries, line 21: maybe have an alias for this type since it's now also used in the dashboard, `BillingWorkspaceType` or something similar?

andrew-farries (changes requested):

components/gitpod-protocol/src/usage.ts
laushinka: Done 💪🏼

andrew-farries (approved):

PR #11217, opened 2022-07-07
[ws-manager] fix bad log entry

sagor999:
## Description

Fixed bad log entry that outputs data in byte instead of string.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
none
```

## Werft options:

- [ ] /werft with-preview


kylos101 (approved):
:+1:

PR #11221, opened 2022-07-08
[usage] Add initial gRPC API

andrew-farries:
## Description

Add an initial version of a `components/usage-api` package that defines a gRPC API for the `usage` component.

The layout of the package is taken from the analogous `public-api` package.

Only Go code generation is included here; typescript can follow as required later.

## Related Issue(s)

Part of https://github.com/gitpod-io/gitpod/issues/10324

## Release Notes

```release-note
NONE
```


easyCZ (approved):

easyCZ:
It's unclear why the installer changes are in this PR, do you know?

andrew-farries:
> It's unclear why the installer changes are in this PR, do you know?

It's because of the transitive dependency installer -> usage -> usage API. 

Arguably the changes to the `usage` component (adding the dependency on the usage API) shouldn't be in this PR, but in the follow up https://github.com/gitpod-io/gitpod/pull/11224 instead.

Happy to sort that out here if you think it's worth it.

PR #11230, opened 2022-07-08
[gpctl] add command to get last heartbeat time of workspace

mustard-mh:
## Related Issue(s)

Fixes #

## How to test

- Open a workspace in prev env
- Open this gitpod with this PR
- Get workspace instance id that you created
```
# from k8s
kubectl get pods | grep ws-

# from workspace terminal
env | grep INSTANCE_ID
```
- Exec `gpctl workspaces last-heartbeat <instance_id>` in gitpod repo workspace

Should output like
```
gitpod /workspace/gitpod/dev/gpctl (main) $ go run main.go workspaces last-heartbeat 3191e531-2c7c-42e5-8f28-57ab47fd33f9
2022-07-08T09:21:09.103544311Z
```

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/dev/gpctl/cmd/workspaces-last-heartbeat.go
+++ b/dev/gpctl/cmd/workspaces-last-heartbeat.go
@@ -1,0 +1,47 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package cmd
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/gitpod-io/gitpod/common-go/log"
+	"github.com/gitpod-io/gitpod/ws-manager/api"
+)
+
+// workspacesLastHeartbeatCmd get workspace last heartbeat time
+var workspacesLastHeartbeatCmd = &cobra.Command{
+	Use:   "last-heartbeat <instanceID>",
+	Short: "get workspace last heartbeat time",
+	Args:  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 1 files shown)


dev/gpctl/cmd/workspaces-last-heartbeat.go
```diff
@@ -0,0 +1,47 @@
...
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		conn, client, err := getWorkspacesClient(ctx)
+		if err != nil {
+			log.WithError(err).Fatal("cannot connect")
+		}
+		defer conn.Close()
+
+		instanceID := args[0]
```
utam0k, line 32: How about receiving get multiple instances IDs?

sagor999: That probably can be done as a separate PR. But good suggestion.

dev/gpctl/cmd/workspaces-last-heartbeat.go
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) 2020 Gitpod GmbH. All rights reserved.
```
sagor999, line 1: why the year is not correct?

sagor999 (approved):
LGTM
/hold
for minor fix

dev/gpctl/cmd/workspaces-last-heartbeat.go
mustard-mh: I'm not going to create a follow-up PR to support multiple instance ids since our commands like `snapshot` `stop` `backup` all only works for one instance. 

We can do it like for now
```
cat backup-pods.txt | xargs -n1 -I{} gpctl workspaces backup {}
```

Issue #10867, opened 2022-06-23
Per node label memory requirements & preflight checks


lucasvaltl:
## Is your feature request related to a problem? Please describe
Currently, we have a single memory requirement for every node - 16gb. However, not all of our components are created equal, and some require more memory than others. E.g. the main driver of the 16gb memory need are the workspace nodes (i.e. image-builder). Non-workspace nodes may require less. 

## Describe the behaviour you'd like
A more granular approach to memory requirements that respects the actual components that will be scheduled on a node. 
- get definitive memory/CPU requirements for all [5 of the node labels](https://github.com/gitpod-io/gitpod/blob/main/install/installer/pkg/cluster/affinity.go#L9) from the appropriate teams, mention this in our requirements and we apply the pre-flight checks on a per node label basis
- This should also be reflected in our reference architectures

## Describe alternatives you've considered
- Keep current one-size-fits-all - while it "works", it is not very resource efficient.

## Additional context
- Internal [slack conversation](https://gitpod.slack.com/archives/C01KLC56NP7/p1655987771833879?thread_ts=1655986020.829029&cid=C01KLC56NP7)

## Tasks

- [x] #11344
- [ ] #11347
- [ ] Update the reference architectures/docs

Pothulapati:
Should we close this? 🤔 Wondering if there is anything left here (to take up)!

lucasvaltl:
I believe we can close this yes

PR #11237, opened 2022-07-08
[kots]: add node CPU/memory check tests to workspace node only

### LINKED ISSUE #9307: Investigate KOTS Preflight Checks Reporting Cluster Memory Incorrectly
### Bug description

The current preflight check which checks the memory of the cluster appears not to be working correctly. Initially when I installed Gitpod this check had no issues, but since then it has reported an issue.

Screen Shot 2022-04-13 at 14 57 02PM

Screen Shot 2022-04-13 at 14 57 35PM

According to Rancher I have plenty of memory available:

Screen Shot 2022-04-13 at 14 01 41PM

I don't believe there was any changes to the preflight checks since `main.2860`, but I could be wrong.

### Steps to reproduce

At this time I have not replicated the issue myself, however I would think you should be able to use the following steps:

1. Spin up a Gitpod instance using the KOTS installer.
2. Visit the KOTS admin dashboard and go through the install process. At this point you _may_ receive the failed preflight check for memory.
3. If you didn't get an issue with the initial preflight checks, you may need to wait until there's a new update available.

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

If there is a command I could manually run that performs the memory check as done in the preflight check please let me know so I can run that to see what is returned.

MrSimonEmms:
## Description

Applies tests on the CPU/memory to the workspace nodes only. Our [requirements](https://www.gitpod.io/docs/self-hosted/latest/cluster-set-up) only stipulate the performance of these nodes, not all the nodes in the cluster.

This also changes the wording for the CPU tests to change "cluster" to "node" (a remnant of when we had the tests on a cluster basis).

## Related Issue(s)

Fixes #9307

## How to test

Run the preflight checks

## Release Notes

```release-note
[kots]: add node CPU/memory check tests to workspace node only
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/kots/manifests/kots-preflight.yaml
+++ b/install/kots/manifests/kots-preflight.yaml
@@ -193,18 +193,80 @@ spec:
           - pass:
               message: Cert-manager is installed and available.
     - nodeResources:
-        checkName: CPU Cores per node
+        checkName: CPU Cores per workload_workspace_services node
+        filters:
+          selector:
+            matchLabel:
+              gitpod.io/workload_workspace_services: "true"
         outcomes:
           - fail:
               when: "min(cpuCapacity) < 2"
-              message: The cluster must contain at least 2 cores
+              message: The nodes must contain at least 2 cores
           - warn:
               when: "min(cpuCapacity) < 4"
-              message: The cluster must contain at least 4 cores
+              message: The nodes must contain at least 4 cores
           - pass:
-              message: There are at least 4 cores in the cluster
+              message: There are at least 4 cores in the nodes
     - nodeResources:
-        checkName: Memory per node
+        checkName: CPU Cores per workload_workspace_regular node
+        filters:
+          selector:
+            matchLabel:
+              gitpod.io/workload_workspace_regular: "true"
+        outcomes:
+          - fail:
+              when: "min(cpuCapacity) < 2"
+              message: The nodes must contain at least 2 cores
+          - warn:
+              when: "min(cpuCapacity) < 4"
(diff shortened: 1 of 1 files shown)


roboquat:

<details>

</details>

lucasvaltl:
👋 Do we have a minimum requirement for the non-workspace nodes though? I fear that there might be a case where someone uses an underpowered node for the server component for example and then wonders why. We did have another related issue (https://github.com/gitpod-io/gitpod/issues/10867) asking for a per-node label hardware requirement. How difficult would that be to implement?

MrSimonEmms:
Honestly, I don't know what the requirements for the non-workspace nodes are. To the best of my knowledge, these aren't documented anywhere so one assumes that there aren't any explicit requirements.

I think these two tickets are pretty much a duplication of each other, albeit with a slightly different focus (#10867 is about the general required whereas #9307 is about how this is tested in KOTS).

The implementation for #10867 is very easy once we have the requirements. My feeling is that that ticket has three phases:
1. Get the requirements from the specific teams
2. Add this into the documentation
3. Add this into KOTS (this will end up looking very similar to this PR, except with other node labels)

nandajavarma (approved):
Changes look good! Didn't test it.

lucasvaltl:
Did we want to merge this already?

MrSimonEmms:
@lucasvaltl I think it's fine that this is merged in as it was technically incorrect (see #9307). We can raise the additional things as a future PR to make it even more correct

PR #11247, opened 2022-07-08
Improve  "GitpodWorkspaceTooManyRegularNotActive" alert

### LINKED ISSUE #11241: Improve regular not active alert, it should trigger sooner
We had an incident in production with workspaces staying in not-ready state for a very long period of time even though it was running. We realized this when a client reported the issue.
We need to improve the alerting in cases where regular workspaces are not ready for a very long period of time so the on-call team can catch this early.

----

_This action was created from [Incident 175](https://app.incident.io/incidents/175?ctx=jfTPvkW6WEqrvnElpmUpr55iIe990UVFLte%2F9SYfH%2F7oxXIAeeO61WIbGdN0xvbb0gy7o3Tpcz0b6GkJbW6YW%2FY77GT4jvSU3S9Vi%2BDehqHSMZQ0NuD8YwoZN3JDdCpCbEJvKXaa5tUTa94%3D) by Nandaja, using [incident.io](https://app.incident.io) 🔥_

nandajavarma:
## Description

Around the time Build one first reported this issue (see [internal](https://gitpod.slack.com/archives/C03145ZD98D/p1657005586660919)) at around 9:15 AM European central Time 7:15 UCT the
 `gitpod_workspace_regular_not_active_percentage` was quite high around 0.9 as can be seen from the picture (it got it's peak at 7AM which was still 0.135). So we definitely do need to reduce the `0.15` value we specify in the prom rule. I propose we set this to `0.10`

grafana png(1)

Now the second thing is, even at this hour, our total active workspaces were much lower, and out alert to go off we have set a limit of `100`. This needs to be reduced too. I propose we eliminate this condition to get alerted if the percentage of inactive workspaces get high enough.
 
grafana-2

According to the trend I see, the inactivity period would be very easily detectable if we reduce our interval to `10m`

## Related Issue(s)

Fixes #11241 

## How to test

The best way to test this would be to try this metric on grafana

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/operations/observability/mixins/workspace/rules/components/workspaces/alerts.libsonnet
+++ b/operations/observability/mixins/workspace/rules/components/workspaces/alerts.libsonnet
@@ -77,14 +77,14 @@
             labels: {
               severity: 'critical',
             },
-            'for': '15m',
+            'for': '10m',
             annotations: {
               runbook_url: 'https://github.com/gitpod-io/runbooks/blob/main/runbooks/GitpodWorkspaceTooManyRegularNotActive.md',
               summary: 'too many running but inactive workspaces',
               description: 'too many running but inactive workspaces',
             },
             expr: |||
-              gitpod_workspace_regular_not_active_percentage > 0.15 AND sum(gitpod_ws_manager_workspace_activity_total) > 100
+              gitpod_workspace_regular_not_active_percentage > 0.10
             |||,
           },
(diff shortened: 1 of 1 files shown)


jenting (approved):
LGTM

Issue #10651, opened 2022-06-14
Epic: Limit TCP connections per workspace


atduarte:
### Summary

Limit number of TCP connects a given workspace can establish, to avoid "conntrack table is full" issues and ensure workspaces don't starve neighbours.

### Context

Currently, we receive a significant number of _GitpodNodeConntrackTableIsFull_ alerts. Whenever this happens, the experience in all workspaces that live in that node are at risk, and the on-callers have to manually identify the abuser and act accordingly. 

Relates to https://github.com/gitpod-io/gitpod/issues/10756 and https://github.com/gitpod-io/gitpod/issues/10760.

### Value

Improve reliability and reduce operational burden.

### Acceptance Criteria

* A given workspace is unable to starve the conntrack table.

### Measurement

* Reduction of _GitpodNodeConntrackTableIsFull_ alert count.

### Tasks

- [x] https://github.com/gitpod-io/gitpod/issues/11632
- [x] #13041

atduarte:
Preliminary work, that relies on augmenting agent-smith with Tetragon capabilities, exists at: https://github.com/gitpod-io/gitpod/tree/fo/agent-smith-mk2 (from @Furisto).
If that works, epic's effort should be small (2 weeks).

utam0k:
@adrienthebo @kylos101 What is the status of this issue? I ask because https://github.com/gitpod-io/gitpod/pull/11255 processed this epic, right?

kylos101:
> @adrienthebo @kylos101 What is the status of this issue? I ask because #11255 processed this epic, right?

I think you meant @atduarte ? 

And no, it does not finish the epic yet. Instead of limiting, it'll record where it would have limited, and we'll be able to use a new graph @Furisto added to the `ws-daemon` dashboard, to determine how often workspaces would have been limited. We can then use that knowledge to determine if we need to tune the configuration, prior to actually limiting.

PR #11255, opened 2022-07-09
Limit connections

Furisto:
## Description
Introduce limiting of new connections created per minute. For more info see the [RFC](https://www.notion.so/gitpod/Rate-limit-network-connections-b3ac59e7db5d41198d63cc7f374c0333).

## Related Issue(s)
Related to https://github.com/gitpod-io/gitpod/issues/10651

## How to test
- Open a workspace in the preview environment
- Annotate the pod with gitpd.io/netConnLimit=true
- This should add additonal nftable rules to the network namespace of the pod (ssh into the node and then use `nsenter -t workspacekitPid -n` and `nft list ruleset`)
- Metrics can be obtained with `kubectl port-forward ds/ws-daemon 9500:9500,` then get the metrics with `curl -XGET localhost:9500/metrics`
- Run roblox scanner

## Release Notes

```release-note
Limit the rate at which network connections can be made by a workspace 
```
## Werft options:

- [x] /werft with-preview


components/ws-daemon/nsinsider/main.go
```diff
@@ -528,3 +562,17 @@ const (
...
 	flagAtRecursive = 0x8000
 )
+
+func addConnectionLimitRule(limit int) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	defer cancel()
+	rule := fmt.Sprintf("add rule ip gitpod ratelimit ip protocol tcp ct state new meter connmeter { ip daddr & 0.0.0.0 timeout 1m limit rate over %d/minute } counter drop", limit)
+
+	// the go library does not support meters, so use nft instead
+	cmd := exec.CommandContext(ctx, "/usr/sbin/nft", rule)
```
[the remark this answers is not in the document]
Furisto: Can you explain what kind of issues have been observed in the context of kubelet? I was not able to find anything.

MrSimonEmms (approved):

sagor999 (approved):
LGTM
adding hold in case if @aledbf comment requires resolving prior to merging
feel free to remove it.

/hold

components/ws-daemon/nsinsider/main.go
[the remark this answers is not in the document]
Furisto: @aledbf Are you ok with merging this or would you like to discuss it further?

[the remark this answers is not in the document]
aledbf: I am not sure we should merge this without knowing the behavior of different `nftable` versions and if there is any conflict with the CNI bandwidth plugin.

Did you test this change in an ephemeral cluster?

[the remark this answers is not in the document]
Furisto: workspace preview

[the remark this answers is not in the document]
kylos101: We used nftables to replace slirp4net, if there is an issue with versions, would we have more fundamental issues with workspace networking? 
* This makes me wonder if we should add related [runtime tests for networking](https://www.notion.so/gitpod/Common-runtime-tests-4e3b05c3b81e4ff6810572aee66ffad3). 
* I ask because I'm not sure what dimensions to consider.

@aledbf I like the idea of testing with an ephemeral cluster. Can you [add a related test here](https://www.notion.so/gitpod/Common-runtime-tests-4e3b05c3b81e4ff6810572aee66ffad3), so we can have a repeatable way to do in the future? 

@Furisto let's leave on-hold till we've had a chance to test in an ephemeral cluster (to test for impacts to CNI bandwidth plugin).

utam0k, line 572: Since this function runs before `pivot_root`, I believe it is safe to specify the `nftable` bin with the complete path at this stage. Am I right? If yes, please add a comment that why it is safe to run.

utam0k: I'm concerned that the user may be able to override this bin with a custom image.

components/ws-daemon/nsinsider/main.go
```diff
@@ -476,6 +478,38 @@ func main() {
...
+					&cli.IntFlag{
+						Name:     "limit",
+						Required: true,
+					},
+				},
+				Action: func(c *cli.Context) error {
+					nftcon := nftables.Conn{}
+					limitTable := nftcon.AddTable(&nftables.Table{
+						Family: nftables.TableFamilyIPv4,
+						Name:   "gitpod",
```
utam0k, line 494: I think it would be better to move ip-fowarding rules to this table. WDYT? Maybe it is good for another PR to take care of that.

components/ws-daemon/nsinsider/main.go
utam0k: cc: @mrzarquon 
We should install the nftables command, right? If yes, we have to change the image and notify the self-hosted team.

components/ws-daemon/nsinsider/main.go
```diff
@@ -528,3 +562,17 @@ const (
 	// FlagAtRecursive: Apply to the entire subtree: https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/fcntl.h#L112
 	flagAtRecursive = 0x8000
 )
+
+func addConnectionLimitRule(limit int) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	defer cancel()
+	rule := fmt.Sprintf("add rule ip gitpod ratelimit ip protocol tcp ct state new meter connmeter { ip daddr & 0.0.0.0 timeout 1m limit rate over %d/minute } counter drop", limit)
```
utam0k, line 569: Can we use set syntax instead of meter as official documentation says here?
https://wiki.nftables.org/wiki-nftables/index.php/Meters
> Note that the meter keyword is obsolete, the dynamic set and map syntax is now preferred for consistency.

components/ws-daemon/nsinsider/main.go
Furisto: We do not enter the mount namespace of the workspace, only the network namespace. The binary [is part of ws-daemon](https://github.com/gitpod-io/gitpod/blob/f4ef8c3457b3bd352554c3feb76cb8f1bcc373c1/components/ws-daemon/leeway.Dockerfile#L25).

components/ws-daemon/nsinsider/main.go
Furisto: Yes, I have thought about encapsulating all the network rules we have within the gitpod table, but I felt that for this PR it is out of scope.

components/ws-daemon/nsinsider/main.go
Furisto: I think we can. Thanks for the hint. I also will update this rule to use a named counter, so that we can scrape metrics about dropped packets more easily.

components/ws-daemon/nsinsider/main.go
utam0k: 🆗

components/ws-daemon/nsinsider/main.go
utam0k: 💯 Can I ask you to add a comment about that? Why? I think this is a very important security point that is difficult to see from the code.

Furisto:
- Limiting is now setup by ws-daemon instead of workspacekit
- We use an annotation (gitpod.io/netConnLimit) to detect workspaces that should be limited
- The go library for nftables recently gained support for dynsets, so using nft is not necessary anymore
- Metrics for dropped bytes/packets have been added
- The network limiting annotation is now set by ws-manager if the WORKSPACE_CONNECTION_LIMITING feature is send as part of the request. The decision to send the flag will be made based on [the plan that the user is on](https://github.com/gitpod-io/gitpod/pull/12162). This part is not yet fully implemented because we do need to deploy this PR first, so that the flag can be handled by ws-manager.

kylos101:
:wave: @Furisto I've added this to our project for team visibility, and changed the description from "fixes" to "related to" (assuming merging this PR won't necessarily close the epic).

Will this audit connections (which we would have limited), or, will it actually limit? I assume audit, but wanted to double check.

Furisto:
@kylos101 
> Will this audit connections (which we would have limited), or, will it actually limit? I assume audit, but wanted to double check.

It will only [audit](https://github.com/gitpod-io/gitpod/blob/a87d9e0b246092f78c2b3113b6776a6f5e50e535/components/ws-daemon/nsinsider/main.go#L616) them

components/common-go/kubernetes/kubernetes.go
```diff
@@ -69,6 +69,9 @@ const (
 
 	// workspaceCpuBurstLimit denotes the cpu burst limit of a workspace
 	WorkspaceCpuBurstLimitAnnotation = "gitpod.io/cpuBurstLimit"
+
+	// workspaceNetConnLimit denotes the maximum number of connections a workspace can make per minute
+	WorkspaceNetConnLimitAnnotation = "gitpod.io/netConnLimit"
```
utam0k, line 74: I felt it would be good that the unit is included in the annotation name.
```suggestion
	WorkspaceNetConnLimitAnnotation = "gitpod.io/netConnLimitPerMin"
```

components/ws-daemon/pkg/nsinsider/nsinsider.go
```diff
@@ -0,0 +1,125 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
```
utam0k, line 1: 👍 Good refactoring. I love these code.

components/ws-daemon/pkg/netlimit/netlimit.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/gitpod-io/gitpod/ws-daemon/pkg/dispatch"
+	"github.com/gitpod-io/gitpod/ws-daemon/pkg/nsinsider"
+	"github.com/google/nftables"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/vishvananda/netns"
+)
+
+type ConnLimiter struct {
+	mu             sync.RWMutex
+	limited        map[string]bool
```
utam0k, line 28: How about using `sync.Map` instead of `mu`? And we can use `struct{}{}` as mark instead of `bool`
```suggestion
	limited        sync.Map
```

utam0k (changes requested):
Almost are good. I left some comments, please check 🙏

geropl (approved):
Reviewed ws-manager-api per code owners: change LGTM

components/ws-daemon/pkg/netlimit/netlimit.go
Furisto: Using a sync.Map would introduce a race condition because we could get a WorkspaceAdded event immediately followed by a WorkspaceUpdated event. If WorkspaceAdded is still processed then the sync.Map would not contain an entry yet and we would try to limit twice. Using a mutex serializes the updates to the limiting rules. I have updated the map to use an empty struct.

Furisto:
@utam0k PTAL

components/ws-daemon/pkg/netlimit/netlimit.go
utam0k: AFAI, in that case, you said either event should be executed. What about doing a LoadOrStore first and Delete if it fails?　　e.g. https://github.com/gitpod-io/gitpod/pull/12215#discussion_r950433725
This is more efficient because the lock is not applied to the entire map.
I learned `sync.Map` from this article.
https://sreramk.medium.com/go-inside-sync-map-how-does-sync-map-work-internally-97e87b8e6bf

Furisto: The way I see it, this is not an operation that is called many thousands of times per second, but only rarely (at most 3 times per second in my experience, usually less). Any contention on the lock is likely to be short lived. I consider simplicity therefore more important than efficiency in this case.

utam0k: Make sense🙆

utam0k (approved):
Awesome

PR #11268, opened 2022-07-11
[installer]: add test for customization of proxy service

### LINKED ISSUE #11106: Installer customization YAML does not recognise 'apiVersion: "v1"'
### Bug description

The `apiVersion: "v1"` in a customization YAML gets rendered the Installer config correctly, but this is not picked up in the internals - it's omitted entirely when it's pushed through to the config map

`apiVersion: "*"` still works though

### Steps to reproduce

When applying the customization file:

```yaml
customization:
  - apiVersion: "v1"
    kind: "Service"
    metadata:
      name: "proxy"
      annotations:
        service.beta.kubernetes.io/aws-load-balancer-ip-address-type: ipv4
```

When looking at the config file (`kubectl get configmaps -n gitpod gitpod -o jsonpath='{.data.config\.yaml}'`), this is rendered as (note the missing `apiVersion`):

```yaml
customization:
- kind: Service
  metadata:
    annotations:
      service.beta.kubernetes.io/aws-load-balancer-ip-address-type: ipv4
    creationTimestamp: null
    name: proxy
  spec:
    env: null
```

### Workspace affected

_No response_

### Expected behavior

The `apiVersion` should be implemented as `v1`

### Example repository

_No response_

### Anything else?

[Internal discussion](https://gitpod.slack.com/archives/C01KLC56NP7/p1656924888317209)

MrSimonEmms:
## Description

This was raised as a specific issue in #11106, but cannot recreate this issue. Test added to ensure that no regression happens.

## Related Issue(s)

Fixes #11106

## How to test

Check werft runs

## Release Notes

```release-note
[installer]: add test for customization of proxy service
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/installer/pkg/common/customize_test.go
+++ b/install/installer/pkg/common/customize_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 
 	"github.com/gitpod-io/gitpod/installer/pkg/common"
+	"github.com/gitpod-io/gitpod/installer/pkg/components/proxy"
 	"github.com/gitpod-io/gitpod/installer/pkg/config/v1"
 	"github.com/gitpod-io/gitpod/installer/pkg/config/versions"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


nandajavarma (approved):
:shipit:

PR #11283, opened 2022-07-11
Basic usage API for TypeScript (1/3)

### LINKED ISSUE #10325: Implement usage APIs
Depends on https://github.com/gitpod-io/gitpod/issues/10324.

With the API definition in place, implement the API. Recommended approach:

 - [x] First define a stub API which returns mocked out data
 - [x] Second, implement necessary access controls and restrictions (including feature flag check)
 - [x] Third, implement actual logic for data retrieval.
 - [ ] Add a "time range filter" parameter

geropl:
## Description
This PR adds an TypeScript client for the `usage-api` and fleshes out a basic `GetBilledUsage` method.

## Related Issue(s)

Context: #10325

## How to test
 - have a look at the `usage.proto`
 - see that the build succeeds 

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


components/usage-api/typescript/BUILD.yaml
```diff
@@ -0,0 +1,29 @@
...
+    env:
+      - DO_PUBLISH=${publishToNPM}
+    argdeps:
+      - npmPublishTrigger
+    deps:
+      - :lib
+      - components/gitpod-protocol:scripts
+    config:
+      commands:
+        - ["node", "components-gitpod-protocol--scripts/publish.js", "${version}", "components-usage-api-typescript--lib/package"]
```
andrew-farries, line 29: Where is this publish task used?

components/usage-api/typescript/src/usage/v1/usage_grpc_pb.js
```diff
@@ -0,0 +1,52 @@
+/**
```
andrew-farries, line 1: Is it necessary to commit both the generated `.ts` and `.js` files?

components/usage-api/usage/v1/usage.proto
```diff
@@ -4,16 +4,35 @@ package usage.v1;
...
-    // GetUsage retrieves all usage for a team.
-    rpc GetUsage(GetUsageRequest) returns (GetUsageResponse) {}
+    // GetBilledUsage retrieves all usage for the specified attributionId
+    rpc GetBilledUsage(GetBilledUsageRequest) returns (GetBilledUsageResponse) {}
 }
 
-message GetUsageRequest {
-    Pagination pagination = 1;
+message GetBilledUsageRequest {
+    string attribution_id = 1;
```
andrew-farries, line 15: I guess it doesn't matter since we never used it, but the field number here normally shouldn't re-use something that was previously assigned to something else.

andrew-farries (approved):

components/usage-api/typescript/src/usage/v1/usage_grpc_pb.js
easyCZ: Yes. The JS provides the implementation, while the TS is only the type-def such that it type-checks nicely.

components/usage-api/typescript/BUILD.yaml
geropl: disclaimer: I did a simple copy, paste and rename from some other `api` package.

We use that somewhere in the build process to be able to consume the NPM package during yarn builds, e.g. in `server`.

components/usage-api/usage/v1/usage.proto
geropl: True, but we have to be conscious about what we carry around with us, because it tends to create veeeery long-living drag.
Removing a never-used attribute is a good choice I think. :+1:

Issue #11183, opened 2022-07-06
[content-service] cannot restart stopped workspace


kylos101:
### Bug description

The full error on start is:

```
cannot initialize workspace: cannot restore backup: tar /dst: tar /dst: exit status 2;tar: .docker-root/overlay2/dedcf720e850fe485571223ecd9959f159eb8eef5250e7ee42ac96f010a3e369/diff/root/.npm/_cacache/content-v2/sha512/2c/49/1e6cfe1a91b8ccb4030ecac14cfe4d1d4e557aaf369a254fd344a9a7e864c7f4cab9edb19e21e7313357cb8eaacc48e19d7fc2b1d2ed294af55846548846: Cannot open: Permission denied tar: .docker-root/overlay2/dedcf720e850fe485571223ecd9959f159eb8eef5250e7ee42ac96f010a3e369/diff/root/.npm/_cacache/content-v2/sha512/74/de/010104b1dac999964663a5780e77912dd860c4bf1089dabe1f9b8175af2aaed3b5e5cba0f24fe31bfc38f19bcc40ffd609a6d2ab6ea02561055f96ba53a6: Cannot open: Permission denied tar: Exiting with failure status due to previous errors
```
Logs:
https://cloudlogging.app.goo.gl/82CpwrNzW9oJWX1Q8

Log entry for the error:
https://console.cloud.google.com/logs/query;cursorTimestamp=2022-07-06T11:31:39Z;query=resource.labels.cluster_name%20:%20%2528%22eu51%22%2529%0A%22eb595f2b-7de4-4248-800d-7dfe0280f802%22%0Atimestamp%3D%222022-07-06T11:31:39Z%22%0AinsertId%3D%227hkbcwop4k5qd0t4%22;summaryFields=:false:32:beginning:false;timeRange=P1D?project=workspace-clusters

Trend over last 7 days:
https://cloudlogging.app.goo.gl/qdGdJ7CrD1zmLyrQ8

![image](https://user-images.githubusercontent.com/1272076/177650734-a91fd61c-126f-4e85-8beb-e0654599cccc.png)

### Steps to reproduce

1. Open this repository on Gitpod: https://gitlab.com/6uliver/gitpod-workspace-restart-error-repro
2. Wait for the workspace and Supertokens to be started, you should see this message in the console: "Started SuperTokens on 0.0.0.0:3567"
3. Stop the workspace and wait for it to be stopped
4. Open and start workspace

### Expected result

[...]

PR #11287, opened 2022-07-11
[supervisor] support shutdown task

### LINKED ISSUE #3966: Allow end-tasks, which will run when a workspace stops (timeout)
Similar to [start-tasks](https://www.gitpod.io/docs/config-start-tasks) I propose adding `end-tasks`.
`End-tasks` will get triggered and run when a workspace gets stopped (timeout)

## Why it's necessary:
* In a "regular" (non-gitpod) setup on a local machine, working on projects that store content and changes in a database is fairly straight forward. I can pull the database once, switch between different branches, make changes in the database, and continue working on the project for many days.
* In a Gitpod setup, I can pull the database once, and any changes I make in the database will persist through timeout and a restart of a workspace. The next day (or minute) I want to work on a different branch so I start a NEW workspace, and _poof!_ the database is gone.

## How it works / Why it's a good idea:
* Once `end-tasks` are available, I can make sure that a machine that is about to be stopped, can run a command that will store the important information / database / docker-image in a safe place. And in the next time I open a new workspace, I can choose fetching that database I saved.
* I am sure there are many more scenarios possible that will be helpful for people once `end-tasks` feature is available.

svenefftinge:
fixes #3966

## Related Issue(s)

Fixes #

## Release Notes

```release-note
```

## Werft options:

- [x] /werft with-preview

Front logo [Front conversations](https://app.frontapp.com/open/top_3p64q)

### DIFF
--- a/components/gitpod-protocol/data/gitpod-schema.json
+++ b/components/gitpod-protocol/data/gitpod-schema.json
@@ -86,6 +86,10 @@
                         "type": "string",
                         "description": "The main shell command to run after `before` and `init`. This command is executed last on every start and doesn't have to terminate."
                     },
+                    "shutdown": {
+                        "type": "string",
+                        "description": "A shell command to run before a workspace stops. This command needs to finish with `terminationGracePeriodSeconds` and will be force killed if it runs longer."
+                    },
                     "env": {
                         "type": "object",
                         "description": "Environment variables to set."
@@ -113,6 +117,10 @@
                 },
                 "additionalProperties": false
             }
+        },
+        "terminationGracePeriodSeconds": {
+            "type": "number",
+            "description": "The grace period for all processes to stop before beeing force killed. Default is 15 seconds."
         },
         "image": {
             "type": [
--- a/components/gitpod-protocol/src/protocol.ts
+++ b/components/gitpod-protocol/src/protocol.ts
@@ -827,6 +827,7 @@ export interface WorkspaceConfig {
     image?: ImageConfig;
     ports?: PortConfig[];
     tasks?: TaskConfig[];
+    terminationGracePeriodSeconds?: number;
     checkoutLocation?: string;
     workspaceLocation?: string;
     gitConfig?: { [config: string]: string };
@@ -993,6 +994,7 @@ export interface TaskConfig {
     init?: string;
     prebuild?: string;
     command?: string;
+    shutdown?: string;
     env?: { [env: string]: any };
     openIn?: "bottom" | "main" | "left" | "right";
     openMode?: "split-top" | "split-left" | "split-right" | "split-bottom" | "tab-before" | "tab-after";
--- a/components/ide/code/startup.sh
+++ b/components/ide/code/startup.sh
@@ -3,14 +3,6 @@
 # Licensed under the GNU Affero General Public License (AGPL).
 # See License-AGPL.txt in the project root for license information.
 
-
- # DO NOT REMOVE THE SPACES AT THE BEGINNING OF THESE LINES
- # The spaces at the beginning of the line prevent those lines from being added to
- # the bash history.
- set +o history
- history -c
- truncate -s 0 "$HISTFILE"
-
 # This is the main entrypoint to workspace container in Gitpod. It is called (and controlled) by the supervisor
 # container root process.
 # To mimic a regular login shell on a local computer we execute this with "bash -li" (interactive login shell):
--- a/components/server/src/bitbucket/bitbucket-file-provider.ts
+++ b/components/server/src/bitbucket/bitbucket-file-provider.ts
@@ -63,7 +63,7 @@ export class BitbucketFileProvider implements FileProvider {
             ).data;
             return contents as string;
         } catch (err) {
-            log.debug({ userId: user.id }, err);
+            log.error({ userId: user.id }, err);
         }
     }
 }
--- a/components/supervisor/pkg/supervisor/config.go
+++ b/components/supervisor/pkg/supervisor/config.go
@@ -294,6 +294,7 @@ type TaskConfig struct {
 	Init     *string                 `json:"init,omitempty"`
 	Prebuild *string                 `json:"prebuild,omitempty"`
 	Command  *string                 `json:"command,omitempty"`
+	Shutdown *string                 `json:"shutdown,omitempty"`
 	Env      *map[string]interface{} `json:"env,omitempty"`
 	OpenIn   *string                 `json:"openIn,omitempty"`
 	OpenMode *string                 `json:"openMode,omitempty"`
--- a/components/supervisor/pkg/supervisor/supervisor.go
+++ b/components/supervisor/pkg/supervisor/supervisor.go
@@ -5,6 +5,7 @@
 package supervisor
 
 import (
+	"bufio"
 	"context"
 	"crypto/rand"
 	"crypto/rsa"
@@ -50,6 +51,7 @@ import (
 	"github.com/gitpod-io/gitpod/content-service/pkg/executor"
 	"github.com/gitpod-io/gitpod/content-service/pkg/git"
 	"github.com/gitpod-io/gitpod/content-service/pkg/initializer"
+	"github.com/gitpod-io/gitpod/content-service/pkg/logs"
 	gitpod "github.com/gitpod-io/gitpod/gitpod-protocol"
 	"github.com/gitpod-io/gitpod/supervisor/api"
 	"github.com/gitpod-io/gitpod/supervisor/pkg/activation"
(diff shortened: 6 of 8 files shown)


roboquat:

<details>

</details>

csweichel:
Had a discussion with @svenefftinge about this:
- shutdown tasks are subject to the `terminationGracePeriod` of a workspace, hence that time should be configurable in the `.gitpod.yml`. We would introduce a `terminationGracePeriod` property on the top level of the file.
- We would introduce a new task type: `shutdown`. When a `shutdown` task is present, upon workspace stop, the child processes of the task executing terminal would receive `SIGTERM` followed by `SIGKILL` 10 seconds later. Then we'd execute the `shutdown` command in the same terminal. If the task terminal has been closed in the meantime we open a new terminal and execute the shutdown command there.
- If a task has no `shutdown` command, the tasks `command` will keep running until all shutdown tasks have completed.
- The log output of the `shutdown` tasks are stored in the workspace filesystem so that the logs are available when the workspace is restarted, or if we choose to upload them. Previously existing shutdown logs are overwritten.

Example `.gitpod.yml`:
```YAML
# give the workspace enough time to finish the shutdown tasks
terminationGracePeriodSeconds: 300

tasks:
  # Note that the shutdown command is NOT part of the same task so that the database
  # keeps running until the `foobardb flush` command has been executed.
  - command: foobardb --datadir /workspace/data/db
  - shutdown: foobardb flush

  # Say you wanted to carry an image across multiple workspaces, using the shutdown command
  # you could push that image to a registry prior before the workspace shuts down
  - before: echo "$DOCKER_PWD" | docker login -u $DOCKER_USER --password-stdin myreg.company.dev
    init: docker pull myreg.company.dev/devstate/$GITPOD_OWNER_ID/server:latest
    command: docker run -it --name server -v /workspace/data/server:/data myreg.company.dev/devstate/$GITPOD_OWNER_ID/server:latest
    shutdown: docker container commit server myreg.company.dev/devstate/$GITPOD_OWNER_ID/server:latest && docker push myreg.company.dev/devstate/$GITPOD_OWNER_ID/server:latest
```

MarcusSorealheis:
Is this PR still alive?

svenefftinge:
Yes, we got to prioritize some other topics for the time being but will pick this up soon again. Could you outline what problem you hope to colve with this?

metcalfc:
We have folks that are affected by https://github.com/gitpod-io/gitpod/issues/11183 which currently seems to be caused by not running `docker compose down`. 

This same set of folks would also use shutdown to push images to Docker registries and persist data to object store on workspace shutdown. They'd do the opposite on workspace startup load data from object store and pull images.

sagor999:
Hey @svenefftinge 
I see you recently pushed into this PR. Are you resuming work on this?
I am asking because we have this task in workspace: https://github.com/gitpod-io/gitpod/issues/11183 that seems like might get fixed by this PR. :pray:

svenefftinge:
@sagor999 there is real analysis of the underlying problem in #11183 can you elaborate a bit on how you hope this PR will fix those issues? Is the underlying problem that docker doesn't terminate gracefully on workspace stop and increasing the time would help?

sagor999:
Yeah, it seems that way indeed.

svenefftinge:
update: depends on https://github.com/gitpod-io/gitpod/pull/13495

akosyakov:
@csweichel Just to understand how I can get proper docker-compose running in the background similar how people are doing it with systemd for instance [[1](https://gist.github.com/r0l1/4039684a76c508aa097a89ae5d752ca4#file-docker-compose-service-L11-L14)]. I would do following?
```yaml
tasks:
  - init: docker-compose pull --quiet --parallel
    command: |
      docker-compose up &
      exit
    shutdown: docker-compose down
```

With this solution though task terminal still can pop up in VS Code while pulling for instance in a workspace without the prebuild. I wist we would have something like background property for tasks that it does not create pty service but only shell, for example:
```yaml
tasks:
  - init: docker-compose pull --quiet --parallel
    command: docker-compose up
    shutdown: docker-compose down
    background: true
```

> shutdown tasks are subject to the terminationGracePeriod of a workspace, hence that time should be configurable in the .gitpod.yml. We would introduce a terminationGracePeriod property on the top level of the file.

👍 

> We would introduce a new task type: shutdown. When a shutdown task is present, upon workspace stop, the child processes of the task executing terminal would receive SIGTERM followed by SIGKILL 10 seconds later. Then we'd execute the shutdown command in the same terminal. If the task terminal has been closed in the meantime we open a new terminal and execute the shutdown command there.

I'm not sure about the order consider another use case with docker-compose with terminal:
```yaml
tasks:
  - init: docker-compose pull --quiet --parallel
    command: docker-compose up
    shutdown: docker-compose down
```

Sending SIGTERM to `docker-compose up` is not the same like running `docker-compose down`. It is not going terminate gracefully. First shutdown command should be executed independently in own shell and only then SIGTERM should be sent to the terminal shell. I also not sure why we need terminal device for shutdown command at all.


> If a task has no shutdown command, the tasks command will keep running until all shutdown tasks have completed.

I'm not sure what is reasoning it, i.e. shutdown of own task can affect another task? Should not we rather design that it should be local to the task itself?

> The log output of the shutdown tasks are stored in the workspace filesystem so that the logs are available when the workspace is restarted, or if we choose to upload them. Previously existing shutdown logs are overwritten.

👍

svenefftinge:
> Sending SIGTERM to docker-compose up is not the same like running docker-compose down. It is not going to terminate gracefully. First shutdown command should be executed independently in own shell and only then SIGTERM should be sent to the terminal shell. I also not sure why we need terminal device for shutdown command at all.

In that case people would need to start the shutdown in a separate terminal:
```
tasks:
  - init: docker-compose pull --quiet --parallel
    command: docker-compose up
  - shutdown: docker-compose down
 ```

akosyakov:
> In that case people would need to start the shutdown in a separate terminal:

Is not it make it complex for a user there will be 2 different semantics depending whether shutdown is present or not? I wonder what use case won't be covered by **all tasks are closed in parallel, shutdown is executed before SIGTERM**? There is also `gp sync-await` if someone wants to couple command and shutdown between 2 tasks already?

svenefftinge:
Yes, makes sense. So you suggest to go back to the design where we don't have a shutdown lifecycle on terminals but one global one. I.e. what I propose [here](https://github.com/gitpod-io/gitpod/issues/3966#issuecomment-1180481517)?

akosyakov:
> Yes, makes sense. So you suggest to go back to the design where we don't have a shutdown lifecycle on terminals but one global one. I.e. what I propose https://github.com/gitpod-io/gitpod/issues/3966#issuecomment-1180481517?

I think it makes sense to have shutdown per a task, i.e. to reload docker-compose configuration. Having it under another task or globally won't allow it. I meant that the contract can be different:
- each task is self contained and can have own shutdown command
- shutdown command is executed first in separate shell
- then SIGTERM is send to the terminal task process if it is still alive

We have `gp tasks stop` already and discussing to add `gp tasks restart`. I would think running `gp tasks stop` for some task should execute it shutdown command regardless of other tasks.

svenefftinge:
I think instead of introducing a hard-coded concept of a lifecycle for which the contract are not really clear (i.e. `shutdown`). We could look into building (or use an existing), convenient CLI that allows me to run something and run something else on signal.

metcalfc:
> Users can already do this with bash and trap or writing their own tools.

This requires developers to be very experienced. Many developers do not live on the command line, do not use shell job management, don't know about the various signals, or how to trap them. I couldn't make this work, @axonasif had to give me a snippet to make it work. Many developers find Dockerfile challenging enough.

Most folks do not have to manage this today. They've got full VMs with real inits and systemd doing clean up.  Having to manage this is something for us and our abstraction.

The original thoughts around  `shutdown` came from a prospective customer, who wanted to manage db state, containers, etc at shutdown. In much the same way as they think of `before`, `init`, and `command`. 

Its easy to drill into the specifics of the particular docker compose example but that wasn't the original signal this came from.

mbrevoort:
I quite dislike that we even have a notion of "stop" and "start". I wish that "stop" was hibernete and "restart" was resume. I wish that we could snapshot the current state versus go through a full lifecycle of starting and stopping a container. Though we use containers it feels like a leaky abstraction.

There are things I would like to be able to do as a developer to distinguish between shutdown and workspace deletion. (but like I mentioned, I don't like that we have start/stop in the first place)

For instance, for a project, for each workspace I create, I want to provision an SQS queue. And when the workspace is deleted I want to deprovision the SQS queue. When the workspace is running I want to be polling the queue. 

I can see following the same lifecycle with a cloud based database as well. Provision a new database or keyspace, load with seed data and clean up when the workspace is deleted. Every "ephemeral" workspace gets it's own ephemeral cloud resources.

axonasif:
> - `@akosyakov`: then SIGTERM is send to the terminal task process if it is still alive

I would like to have an option within `.gitpod.yml` to disable that behavior for a task. This could be helpful when I want to have manually handle from `shutdown` of a task when necessary. As you mentioned, `docker-compose up` doesn't like SIGTERM as they just simply handle SIGTERM that way, I may not want to start it on the background as a daemon to avoid SIGTERM if that's the current implementation of where SIGTERM's are sent.

> - `@csweichel`: If a task has no shutdown command, the tasks command will keep running until all shutdown tasks have completed.

I assume this is close to what I'm asking.

svenefftinge:
> This requires developers to be very experienced.

That's why I suggested building or using a convenient CLI.

PR #11288, opened 2022-07-11
[installer]: create tests to verify render result

### LINKED ISSUE #11147: [installer] Allow setting the rand seed as CLI argument for testing purposes
The installer command `render` renders the manifests with different random generated values (e.g. passwords). For testing, it's pretty hard to compare the effect of code changes in the installer because comparing the output of 2 different runs leads to a lot of changes just due to new generated random values.

A CLI argument like `--random-seed 42` that leads to the call of `rand.Seed(42)` in the Go code of the installer would lead to the same random values across different runs. Of course, this is not recommended for production usage and should probably lead to a warning.

There are still values that are generated somewhere else (currently 3 values in total). It would be great if we would find a way to control these values as well. However, controlling the seed in the Go code would be a first good step, in my opinion.

MrSimonEmms:
## Description

This is a proposed change to allow for testing of the rendered YAML in the Installer.

The data is stored in `cmd/testdata/render`, with the test names being stored in `cmd/testdata/render/<name>/<config|output>.golden`. Files ending `.golden` are excluded from the hooks to make them testable. The intention is that there will be a suite of these tests and they verify that the known output is matched. When a change is made (even outside of this directory), the files will require changing and @gitpod-io/engineering-self-hosted is notified about potentially breaking changes to the rendered YAML - we can then approve (or not) the changes.

How to generate the files is documented.

Randomisation is controlled by specifying the `Seed` for `math/rand` and by giving the `crypto/rand` a known string. A warning is printed to stderr if this is enabled during the YAML rendering process, which is designed to act as a deterrent to anyone using this in production.

### Risks Introduced

It's not impossible for a PR to pass its tests in a branch and then fail when it's merged. This can be mitigated by getting into the habit of doing a final `git rebase main` before this is merged.

In the event of a failure in the `main` branch, the engineer who authored the original PR is responsible for ensuring that this is fixed in a timely fashion.

## Related Issue(s)

Fixes #11147

## How to test

Run tests in Werft. See also the [docs](https://github.com/gitpod-io/gitpod/blob/sje/render-testing/install/installer/cmd/testdata/render/README.md)

## Release Notes

```release-note
[installer]: create tests to verify render result
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,3 +50,5 @@ repos:
         entry: leeway run components:update-license-header
         language: system
         pass_filenames: false
+
+exclude: ^install/installer/.*/.*\.golden$
--- a/install/installer/.gitignore
+++ b/install/installer/.gitignore
@@ -1,3 +1,3 @@
 installer
 dist
-versions.yaml
+/versions.yaml
--- a/install/installer/BUILD.yaml
+++ b/install/installer/BUILD.yaml
@@ -5,6 +5,7 @@ packages:
       - go.mod
       - go.sum
       - "**/*.go"
+      - "**/testdata/**/*"
       - "cmd/versions.yaml"
       - "pkg/components/**/*.tpl"
       - "pkg/components/**/*.crt"
--- a/install/installer/Makefile
+++ b/install/installer/Makefile
@@ -1,5 +1,5 @@
 OUTPUT ?= dist
-VERSION_MANIFEST = ./pkg/config/versions/versions.yaml
+VERSION_MANIFEST ?= ./pkg/config/versions/versions.yaml
 
 all: clean deps versionManifest compile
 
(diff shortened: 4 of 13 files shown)


roboquat:

<details>

</details>

corneliusludmann (approved):
Awesome! Thanks for implementing this! :rocket:

MrSimonEmms:
Hi @iQQBot, to get through this you'll need to run `make deps` inside `/install/installer` as the tests don't download the Helm dependencies at runtime.

I'll create a PR to auto-download these as part of the workspace creation

MrSimonEmms:
It is random how it's generated, but we sort the generated YAML - see https://github.com/gitpod-io/gitpod/blob/main/install/installer/cmd/render.go#L219

We also have an option to set the random values to a deterministic state - in the tests, we just set that to `42`. See https://github.com/gitpod-io/gitpod/blob/main/install/installer/cmd/root.go#L42,L53

MrSimonEmms:
Fix created https://github.com/gitpod-io/gitpod/pull/11829

Issue #8133, opened 2022-02-10
Settings page on mobile layout


krystxf:
It would be nice to have a responsive settings page. Probably with some sort of drop down menu instead of the side menu on mobile. I'm willing to help with the styling, but I need some suggestions to get it the way you'd like.

pawlean:
Thanks for your feedback, @krystofex! I'm looping in @gtsiolis for information. 😊

jldec:
Moved to investigating/visual-design

gtsiolis:
Thanks for opening this, @krystofex! This is related to the responsive layout of the dashboard in general, see also https://github.com/gitpod-io/gitpod/issues/4050.

Thanks for the ping @pawlean! 🏓

gtsiolis:
Regarding settings layout specifically, let me cross-post below an early design from a [relevant discussion](https://gitpod.slack.com/archives/C01KGM9BUNS/p1625656466130800?thread_ts=1625644808.128800&cid=C01KGM9BUNS) (internal).

Cross-posting also a [relevant doc](https://www.notion.so/gitpod/Mobile-friendly-Gitpod-a383083509754d77a355d769dd699975#3ce5bf63a4ae418981af58b23daa9717) (internal) from @jankeromnes.

<img width="1137" alt="Screenshot 2021-07-07 at 2 12 47 PM" src="https://user-images.githubusercontent.com/120486/153859779-a81f52ec-bfaa-4b42-af2c-b2e19aad1627.png">

jankeromnes:
Awesome, many thanks for getting the ball rolling and offering to help here @krystofex! ♥️🚀📱

And thanks @gtsiolis for re-surfacing the initial designs!

I like many of their aspects.
- For the top-right menu, I guess `Admin`, `Docs` and `Help` could indeed get folded into the Avatar drop-down menu
- For the Settings menu (and other similar sub-menus), I'm not a big fan of the big drop-down selector. We used to have this on https://www.gitpod.io/docs/ and it felt a bit awkward/unfinished (now they have a nice "Go back to all topics" link / all topics page)
- For the top-left menu, I'm not sure whether grouping into a drop-down makes sense, especially with our new "personal/team account selector with tabs beneath" UI

In any case, I do think maybe the top-left and top-right menus are out of scope here, since the issue specifically mentions the Settings menu. Focusing on these sub-menus first makes sense.

I wonder how other apps with outstanding mobile UI have solved this problem.

Issue #7879, opened 2022-01-27
Remove the sidebar menu and use the tabs for the admin dashboard


gtsiolis:
### Problem to solve

Given the recent additions of the settings (https://github.com/gitpod-io/gitpod/pull/7732) and upcoming addition of projects (https://github.com/gitpod-io/gitpod/issues/7636) in the admin dashboard, now that we have tabs, which were not preset during the initial dashboard redesign, long-term we'd like to use the tabs and replace the pills we currently have in the admin dashboard as this pattern fits and scales better for this need.

### Proposal

Remove the sidebar (pills) menu and use the navigation at the top (tabs) for the admin dashboard navigation. Preferrably, re-using the order from teams: _Users_, _Workspaces_, _Projects_, _Settings_.

This will also match the pattern used for surfacing workspaces, projects, and settings for personal accounts and teams.

Cc @jldec based on our [discussion](https://www.notion.so/gitpod/WebApp-Design-Sync-8c61b5f0cd2840e394fbe15f8f25099d) (internal) earlier today and https://github.com/gitpod-io/gitpod/issues/7597.

See also [relevant discussion](https://gitpod.slack.com/archives/C02EN94AEPL/p1643298717497700?thread_ts=1643283947.483100&cid=C02EN94AEPL) (internal). Cc @laushinka 

| BEFORE | AFTER |
|-|-|
| <img width="1440" alt="Screenshot 2022-06-28 at 5 52 19 PM (2)" src="https://user-images.githubusercontent.com/120486/176211141-115726e7-bd9d-41be-97c5-28681d9b6812.png"> | <img width="1440" alt="Screenshot 2022-06-28 at 5 54 15 PM (2)" src="https://user-images.githubusercontent.com/120486/176211196-b72e221e-5a4f-407e-90ac-f545c557ce75.png"> |

See [design specs](https://www.figma.com/file/4zbw802MFwPUyHxAM8m27B/Remove-the-sidebar-menu-and-use-the-tabs-for-the-admin-dashboard-%237879?node-id=0%3A1).

SangameshItagi:
@gtsiolis Will work in this issue.

gtsiolis:
Closing this in favor of https://github.com/gitpod-io/gitpod/issues/8385.

gtsiolis:
Let's reopen this as a follow-up issue and leftover from #8385. 

[...]

PR #11304, opened 2022-07-12
Add feature flag for blocked repositories UI

andrew-farries:
## Description

Add a feature-flagged entry in the admin sidebar that will show a placeholder blocked repositories admin UI.

The feature flag is enabled for non-production environments (ie preview and staging).

## Related Issue(s)
Part of #11030 

## How to test

Start the preview environment and see the new `Repositories` menu item in the `Admin` menu:

image

The link takes you to an empty placeholder page:

image

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/dashboard/src/Menu.tsx
+++ b/components/dashboard/src/Menu.tsx
@@ -15,7 +15,7 @@ import { getGitpodService, gitpodHostUrl } from "./service/service";
 import { UserContext } from "./user-context";
 import { TeamsContext, getCurrentTeam } from "./teams/teams-context";
 import getSettingsMenu from "./settings/settings-menu";
-import { adminMenu } from "./admin/admin-menu";
+import { getAdminMenu } from "./admin/admin-menu";
 import ContextMenu from "./components/ContextMenu";
 import Separator from "./components/Separator";
 import PillMenuItem from "./components/PillMenuItem";
@@ -26,6 +26,7 @@ import { ProjectContext } from "./projects/project-context";
 import { PaymentContext } from "./payment-context";
 import FeedbackFormModal from "./feedback-form/FeedbackModal";
 import { inResource, isGitpodIo } from "./utils";
+import { FeatureFlagContext } from "./contexts/FeatureFlagContext";
 import { getExperimentsClient } from "./experiments/client";
 
 interface Entry {
@@ -48,6 +49,7 @@ export default function Menu() {
         setIsStudent,
         setIsChargebeeCustomer,
     } = useContext(PaymentContext);
+    const { isBlockedRepositoriesUIEnabled } = useContext(FeatureFlagContext);
     const { project, setProject } = useContext(ProjectContext);
     const [isFeedbackFormVisible, setFeedbackFormVisible] = useState<boolean>(false);
 
@@ -246,7 +248,7 @@ export default function Menu() {
                   {
                       title: "Admin",
                       link: "/admin",
-                      alternatives: adminMenu.flatMap((e) => e.link),
+                      alternatives: getAdminMenu(isBlockedRepositoriesUIEnabled).flatMap((e) => e.link),
                   },
               ]
             : []),
--- a/components/dashboard/src/admin/admin-menu.ts
+++ b/components/dashboard/src/admin/admin-menu.ts
@@ -4,29 +4,39 @@
  * See License-AGPL.txt in the project root for license information.
  */
 
-export const adminMenu = [
-    {
-        title: "Users",
-        link: ["/admin/users", "/admin"],
-    },
-    {
-        title: "Workspaces",
-        link: ["/admin/workspaces"],
-    },
-    {
-        title: "Projects",
-        link: ["/admin/projects"],
-    },
-    {
-        title: "Teams",
-        link: ["/admin/teams"],
-    },
-    {
-        title: "License",
-        link: ["/admin/license"],
-    },
-    {
-        title: "Settings",
-        link: ["/admin/settings"],
-    },
-];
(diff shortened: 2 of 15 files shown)


components/dashboard/src/Menu.tsx
```diff
@@ -182,6 +184,17 @@ export default function Menu() {
         })();
     }, [user, teams, team, project]);
 
+    useEffect(() => {
+        (async () => {
+            const isBlockedRepositoriesUIEnabled = await getExperimentsClient().getValueAsync(
+                "isblockedrepositoriesuienabled",
+                false,
+                {},
```
geropl, line 192: FYI: It's super convenient to just pass in `user` here, which allows us to enable this feature flag more selectively (e.g., for all Gitpodders).

components/dashboard/src/contexts/FeatureFlagContext.tsx
```diff
@@ -0,0 +1,32 @@
...
+const FeatureFlagContext = createContext<{
+    isBlockedRepositoriesUIEnabled: boolean;
+    setIsBlockedRepositoriesUIEnabled: React.Dispatch<boolean>;
+}>({
+    isBlockedRepositoriesUIEnabled: false,
+    setIsBlockedRepositoriesUIEnabled: () => {},
+});
+
+const FeatureFlagContextProvider: React.FC = ({ children }) => {
+    const [isBlockedRepositoriesUIEnabled, setIsBlockedRepositoriesUIEnabled] = useState(false);
```
geropl, line 18: Nice! Was thinking about whether this make sense as well.

:bulb: I it possible to `useUserContext()` here, and the the values with an `useEffect` in here? :thinking: 
(this is clearly something for a follow up!)

geropl:
@andrew-farries When loading `/workspace` I see a blank page in 50% of the cases, or errors like this:
```
![image](https://user-images.githubusercontent.com/32448529/178435439-698f7af8-3ff3-440b-8b70-96274bff1eed.png)
```
Other errors cite `FeatureFlagContext` and `AdminSubMenu`.

Also, I noticed `isblockedrepositoriesuienabled` is queried two times per load.

:cloud: As much as I love the "feature flag in context" approach: One way to move forward with this would be to do the feature flag it "as usual", and play around with the other approach separately. :100:

components/dashboard/src/Menu.tsx
andrew-farries: Thanks, done in 73eedce

andrew-farries:
> @andrew-farries When loading `/workspace` I see a blank page in 50% of the cases, or errors like this:

The errors you were seeing were due to me using a telepresence pod at the same time that Werft deployed another dashboard pod to the preview environment. I had the same problem. I've stopped messing with telepresence so it should be good now.

geropl (approved):
Code LGTM, tested and works.

/hold because of names, names, names... https://github.com/gitpod-io/gitpod/pull/11304#issuecomment-1181452155 :upside_down_face:

gtsiolis:
Looking at this now! 👀 

Thanks for the ping, @andrew-farries!  🏓

components/dashboard/src/admin/admin-menu.ts
```diff
@@ -4,29 +4,39 @@
...
+            link: ["/admin/workspaces"],
+        },
+        ...(isBlockedRepositoriesUIEnabled
+            ? [
+                  {
+                      title: "Blocked Repositories",
+                      link: ["/admin/blocked-repositories"],
+                  },
+              ]
+            : []),
```
gtsiolis, line 24: **suggestion(non-blocking)**: What do you think of moving this further below, as this is more related to abuse control of the instance than associated with the workspace, project, or team management?

Later on, we could rename this section to _Abuse_, _Safety_, or something similar that could contain more abuse controls. 💭 

| BEFORE | AFTER |
|-|-|
| Screenshot 2022-07-12 at 1 38 26 PM | Screenshot 2022-07-12 at 1 45 16 PM |

gtsiolis (approved):
Thanks again for the ping, @andrew-farries! 🏓 
Left one non-blocking comment regarding menu order. ❗ 

🍊 🍊 🍊 🍊 

The subtle `16px` change in the width of the sidebar sounds ok from UX perspective to leave as is for all sidebar instances, no need to introduce a prop for conditionally change the width. ✔️

This would also look ok even after replacing the sidebar menu with tabs for the admin dashboard #7879. ✔️

Long-term we can work of fine tuning breakpoints and column layout, see [relevant discussion](https://gitpod.slack.com/archives/C01KPEPQYE6/p1657294999865219) (internal) and some early designs around this in https://github.com/gitpod-io/gitpod/issues/8133#issuecomment-1039013543. 🗺️ 

🍋 🍋 🍋 🍋 

One aspect that is not clear but probably irrelevant and out of the scope of this PR at the moment is whether we'd like to also list **repositories**, introduce a repository view, and have an action to block from there. This could be relevant when we introduce better control for multi-repository support, etc.

🥝 🥝 🥝 🥝 

Regarding the menu label, I agree with @geropl in https://github.com/gitpod-io/gitpod/pull/11304#issuecomment-1181452155 on being more explicit here although the dashboard design so far has been using shorter verbs or labels when possible to be clearer to users but also follow our [design principles](https://www.notion.so/gitpod/Enabling-Team-Product-Design-33f7d217be0d4db5a0666e7c99aece19#b8608b9f94434b5cb7043a10dd29a9a0) (internal). 🥯

components/dashboard/src/admin/admin-menu.ts
andrew-farries: Thanks 👍 I've moved the new entry down in the list as you suggested.

Issue #11228, opened 2022-07-08
All requests from Gateway to server is blocked


mustard-mh:
### Bug description

All request from Gateway to server is missing, `sendHeartbeat` not works so that users will get timed out by their default time out duration if they only open JetBrains IDE.

See also user report in [discord](https://discord.com/channels/816244985187008514/994584933265846322), and [internal Chat](https://gitpod.slack.com/archives/C03LKE6AS93/p1657202967512559)

This bug come up since `2 pm, Thu, Jun 30, 2022`, maybe related to PR https://github.com/gitpod-io/gitpod/pull/10974

### Steps to reproduce

- Open workspaces with any JetBrains IDEs
- Check if heartbeat sent

felladrin:
ℹ️ The fix has just been deployed to Gitpod servers.

PR #11307, opened 2022-07-12
Clean code from GitpodServerLauncher and update JetBrains backend-plugin to work when user has proxies

### LINKED ISSUE #11291: Refactor the Gitpod Server Launcher [Follow-up from #11232]
To [fix the issue about JetBrains IDEs not signaling heartbeats](https://github.com/gitpod-io/gitpod/pull/11232), we've brought back code from a previous implementation. Now that we confirmed the problem is fixed, we need to [refactor the solution](https://github.com/gitpod-io/gitpod/pull/11232#discussion_r917586352) to avoid duplicated code and ensure it's covering all connection cases.

Also, we need to make backend-plugin work the same way as gateway-plugin regarding proxies. See below:
https://github.com/gitpod-io/gitpod/blob/aafcaf1bca46cb286cc9ef1bf05b8b4caee09315/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodManager.kt#L276-L282
https://github.com/gitpod-io/gitpod/blob/aafcaf1bca46cb286cc9ef1bf05b8b4caee09315/components/ide/jetbrains/gateway-plugin/src/main/kotlin/io/gitpod/jetbrains/gateway/GitpodConnectionService.kt#L52-L67

felladrin:
## Description

Clean code from GitpodServerLauncher and update JetBrains backend-plugin to work when the user has proxies.

This reverts the hotfix from https://github.com/gitpod-io/gitpod/pull/11232 and passes a default `SslContextFactory` to `HttpClient`, which the root cause of the issue https://github.com/gitpod-io/gitpod/issues/11228. For reference, this is the error message the server was triggering:
```
java.lang.IllegalStateException: HttpClient has no SslContextFactory, wss:// URI's are not supported in this configuration
at io.gitpod.gitpodprotocol.api.GitpodServerLauncher.listen
```

## Related Issue(s)

Resolves #11291

## How to test

- Open https://felladrin-11291.preview.gitpod-dev.com/#referrer:jetbrains-gateway:intellij/https://github.com/gitpod-io/empty
- Get the instance of the workspace by running `env | grep GITPOD_INSTANCE_ID`
- Check if [heartbeats are sent to set to the server](https://github.com/gitpod-io/gitpod/pull/11230).

## Release Notes

```release-note
Gitpod Plugin for JetBrains IDEs was updated to properly handle network proxies.
```

## Documentation

None.

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-protocol/java/src/main/java/io/gitpod/gitpodprotocol/api/GitpodServerLauncher.java
+++ b/components/gitpod-protocol/java/src/main/java/io/gitpod/gitpodprotocol/api/GitpodServerLauncher.java
@@ -22,7 +22,6 @@ import java.net.InetSocketAddress;
 import java.net.Proxy;
 import java.net.SocketAddress;
 import java.net.URI;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -55,60 +54,16 @@ public class GitpodServerLauncher {
             String origin,
             String userAgent,
             String clientVersion,
-            String token
-    ) throws Exception {
-        return listen(apiUrl, origin, userAgent, clientVersion, token, Collections.emptyList(), null);
-    }
-
-    public GitpodServerConnection listen(
-            String apiUrl,
-            String origin,
-            String userAgent,
-            String clientVersion,
             String token,
             List<Proxy> proxies,
             SSLContext sslContext
     ) throws Exception {
         String gitpodHost = URI.create(apiUrl).getHost();
-        HttpClient httpClient;
-        if (sslContext == null && proxies.size() == 0) {
-            GitpodServerConnectionImpl connection = new GitpodServerConnectionImpl(gitpodHost);
-            connection.setSession(ContainerProvider.getWebSocketContainer().connectToServer(new Endpoint() {
-                @Override
-                public void onOpen(Session session, EndpointConfig config) {
-                    session.addMessageHandler(new WebSocketMessageHandler(messageReader, jsonHandler, remoteEndpoint));
-                    messageWriter.setSession(session);
-                    client.notifyConnect();
-                }
-
-                @Override
-                public void onClose(Session session, CloseReason closeReason) {
-                    connection.complete(closeReason);
-                }
-
-                @Override
-                public void onError(Session session, Throwable thr) {
-                    GitpodServerConnectionImpl.LOG.log(Level.WARNING, gitpodHost + ": connection error:", thr);
-                    connection.completeExceptionally(thr);
-                }
-            }, ClientEndpointConfig.Builder.create().configurator(new ClientEndpointConfig.Configurator() {
-                @Override
-                public void beforeRequest(final Map<String, List<String>> headers) {
-                    headers.put("Origin", Arrays.asList(origin));
-                    headers.put("Authorization", Arrays.asList("Bearer " + token));
-                    headers.put("User-Agent", Arrays.asList(userAgent));
-                    headers.put("X-Client-Version", Arrays.asList(clientVersion));
-                }
-            }).build(), URI.create(apiUrl)));
-            return connection;
-        }
-        if (sslContext == null) {
-            httpClient = new HttpClient();
-        } else {
-            SslContextFactory ssl = new SslContextFactory.Client();
+        SslContextFactory ssl = new SslContextFactory.Client();
+        if (sslContext != null) {
             ssl.setSslContext(sslContext);
-            httpClient = new HttpClient(ssl);
         }
+        HttpClient httpClient = new HttpClient(ssl);
         for (Proxy proxy : proxies) {
             if (proxy.type().equals(Proxy.Type.DIRECT)) {
                 continue;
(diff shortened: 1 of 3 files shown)


components/gitpod-protocol/java/src/main/java/io/gitpod/gitpodprotocol/api/GitpodServerLauncher.java
```diff
@@ -70,77 +59,52 @@ public GitpodServerConnection listen(
...
-                    messageWriter.setSession(session);
-                    client.notifyConnect();
-                }
+        GitpodServerConnectionImpl connection = new GitpodServerConnectionImpl(gitpodHost);
 
-                @Override
-                public void onClose(Session session, CloseReason closeReason) {
-                    connection.complete(closeReason);
-                }
+        if (!proxies.isEmpty()) {
```
akosyakov, line 64: it is not correct, certificates and proxies are independent.

components/gitpod-protocol/java/src/main/java/io/gitpod/gitpodprotocol/api/GitpodServerLauncher.java
```diff
@@ -70,77 +59,52 @@ public GitpodServerConnection listen(
...
+                if (proxy.type().equals(Proxy.Type.HTTP)) {
+                    httpClient.getProxyConfiguration().getProxies().add(new HttpProxy(hostName, port));
+                } else if (proxy.type().equals(Proxy.Type.SOCKS)) {
+                    httpClient.getProxyConfiguration().getProxies().add(new Socks4Proxy(hostName, port));
                 }
-            }).build(), URI.create(apiUrl)));
-            return connection;
-        }
-        if (sslContext == null) {
-            httpClient = new HttpClient();
```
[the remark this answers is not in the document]
felladrin: I checked the logs, but there was no message around the point where it was failing.

components/gitpod-protocol/java/src/main/java/io/gitpod/gitpodprotocol/api/GitpodServerLauncher.java
```diff
@@ -70,77 +59,52 @@ public GitpodServerConnection listen(
...
+                try {
+                    container.stop();
+                } catch (Throwable t) {
+                    GitpodServerConnectionImpl.LOG.log(Level.WARNING, gitpodHost + ": failed to stop websocket container:", t);
+                }
+            });
+        }
 
-        connection.setSession(container.connectToServer(new Endpoint() {
+        connection.setSession(ContainerProvider.getWebSocketContainer().connectToServer(new Endpoint() {
```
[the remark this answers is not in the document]
felladrin: True! I'm updating the code now to bring the `container` back to this line.

But I'm starting to think that was the cause of the incident. Because the previous code was using `ContainerProvider.getWebSocketContainer()`.

Cold we do like the following?

- if hasProxies: `connection.setSession(container.connectToServer(new Endpoint() {`
- else: `connection.setSession(ContainerProvider.getWebSocketContainer().connectToServer(new Endpoint() {`

Or we shouldn't use `ContainerProvider.getWebSocketContainer()` for any case?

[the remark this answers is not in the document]
akosyakov: > Or we shouldn't use ContainerProvider.getWebSocketContainer() for any case?

there should not be need for it.

components/gitpod-protocol/java/src/main/java/io/gitpod/gitpodprotocol/api/GitpodServerLauncher.java
```diff
@@ -161,10 +116,10 @@ public void onError(Session session, Throwable thr) {
...
             @Override
             public void beforeRequest(final Map<String, List<String>> headers) {
-                headers.put("Origin", Arrays.asList(origin));
-                headers.put("Authorization", Arrays.asList("Bearer " + token));
-                headers.put("User-Agent", Arrays.asList(userAgent));
-                headers.put("X-Client-Version", Arrays.asList(clientVersion));
+                headers.put("Origin", Collections.singletonList(origin));
+                headers.put("Authorization", Collections.singletonList("Bearer " + token));
+                headers.put("User-Agent", Collections.singletonList(userAgent));
+                headers.put("X-Client-Version", Collections.singletonList(clientVersion));
```
felladrin, line 122: ℹ️ Updated by IntelliJ IDEA.

felladrin:
I wasn't able to check the heartbeats of the workspace on the preview environment, but I confirm the workspace was up for almost two hours, meaning the refactoring succeeded. So I'm setting it as Ready for Review.

akosyakov (approved):
i did not try, but code changes looks good.

/hold

if someone wants to test, if not please merge

akosyakov:
> I wasn't able to run gpctl workspaces last-heartbeat command

just look at logs and see what JB backend can connect to Gitpod API

akosyakov:
Screenshot 2022-07-13 at 13 40 48

/unhold

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodManager.kt
```diff
@@ -271,14 +274,20 @@ class GitpodManager : Disposable {
         val connect = {
             val originalClassLoader = Thread.currentThread().contextClassLoader
             try {
+                val proxies = CommonProxy.getInstance().select(URL(info.gitpodHost))
```
akosyakov, line 277: @felladrin Should not it be `"https://$gitpodHost/"`?

felladrin: I'll double-check it in a moment. But according to the variable description, `info.gitpodHost` should be the full URL of the host, including the protocol.

felladrin: Confirmed. These are the values when running a workspace on [gitpod.io](https://gitpod.io):
- `info.gitpodHost`: `https://gitpod.io`
- `info.gitpodApi.endpoint`: `wss://gitpod.io/api/v1`

Issue #10843, opened 2022-06-22
Customers should be able to select their preferred workspace class


Furisto:
## Is your feature request related to a problem? Please describe
Support https://github.com/gitpod-io/gitpod/issues/8261

## Describe the behaviour you'd like
As a user of Gitpod, I would like to select the workspace class that is used for regular and prebuild workspaces.

## Describe alternatives you've considered
n.a.

## Additional context
Affected components:
- Dashboard
- gitpod-protocol
- gitpod-db
- ws-manager-api
- server

kylos101:
@Furisto this issue is still open, but the status is awaiting deployment. 🤔 

Is there a PR that can be linked to this issue? 🙏 

I ask because I'm sanitizing our Awaiting Deployment column. 🧹

Issue #10805, opened 2022-06-21
Epic: Workspace Classes on Self-Hosted


atduarte:
### Summary

Productise workspace classes for self-hosted.

### Context

After https://github.com/gitpod-io/gitpod/issues/8261.

We are unsure what capabilities that previous epic will bring to self-hosted admins, and what the experience will look like. This epic will be the follow-up to ensure that those concerns are taken care of.

### Value

Self-hosted customer will be able to provide differentiated workspace resources for different people, optimizing the use of their resources.

### Acceptance Criteria

* Self-hosted admins should be able to define multiple workspace classes, at installation time. (Kots, self-hosted team)
* Self-hosted admins should be able to control the type of underlying node, node density and CPU/IO/network/RAM limits of each workspace class. (Docs, Workspace team)
* Provide self-hosted admin guidance on how what limits (Kubernetes and ws-daemon) and nodes to use, and how to optimize resource usage for the customers workload (which ultimately influences the density per node per workspace class). (Docs, Workspace team)

### Measurement

Number of customers using the feature.

ynedelec:
Hi,
This would be very, very valuable to us.

Without diving into too much details, our data scientists would use Gitpod for both software development and exploratory analysis. While the former doesn't require high specs environments, the later could have drastically different needs in term of CPU or RAM.

This is why having the possibility to trigger Gitpod workspaces with different specs is a requirement for us.

kylos101:
@atduarte closing because this was for self-hosted. Feel free to reopen and update if needed for Dedicated.

PR #11313, opened 2022-07-12
Set workspace class based on user preference

Furisto:
## Description
This PR allows user to select their preferred workspace class and start a workspace with it. The feature is hidden behind a feature flag which is segmented to only activate the feature for members of gitpod. This means that WebApp can deploy this feature without having to wait for team workspace to configure the new workspace classes in the workspace clusters.

Currently there are 4 workspace classes that are supported:
- default
- gitpod-internal-xl
- g1-standard
- g1-large

Default and gitpod-internal-xl are the old internal workspace classes that will be eventually removed from the code. G1 are the new general purpose workspace classes that we are introducing.

## Related Issue(s)
Fixes #https://github.com/gitpod-io/gitpod/issues/10843

## How to test
- Go to the dashboard and select your preferred workspace class in the settings
- Open a workspace
- You can check that you have the correct workspace class either by checking the labels on the pod (gitpod.io/workspaceClass) or checking the GITPOD_WORKSPACE_CLASS environment variable in the workspace. The available classes are g1-standard and g1-large.
- Stop the workspace, then change your workspace class selection. Start the workspace again and check that the workspace still has the previously selected workspace class (the workspace class on a workspace needs to stay the same as otherwise downgrading to a lower workspace class could mean that content init will fail due to storage size limit).
- Start a new workspace. The workspace should have new workspace class that you have configured.
- Turn off the feature flag, then start a workspace. The workspace class configuration in the settings menu should be gone and the workspace should have the "default" workspace class instead of  one of the new workspace classes. There should be no entry in the workspaceClass column of the workspace and workspace instance. 
- Stop the workspace, then turn on the feature flag again. Now start the workspace you just stopped. This tests if a workspace that has been created before workspace classes still starts correctly. The workspace will get assigned a class based on if the user "has more resources". That means either g1-standard will be assigned or g1-large. 

## Release Notes

```release-note
None
```

## Werft options:

- [x] /werft with-preview


roboquat:
@Furisto: Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://git.k8s.io/community/contributors/guide/release-notes.md) to remove it.

<details>

</details>

components/server/src/workspace/workspace-starter.ts
```diff
@@ -690,122 +689,149 @@ export class WorkspaceStarter {
...
+                // set the workspace class based on if the user "has more resources"
+                if (!workspace.workspaceClass) {
+                    if (await this.userService.userGetsMoreResources(user)) {
+                        workspaceClass = this.config.workspaceClasses.defaultMoreResources;
+                    } else {
+                        workspaceClass = this.config.workspaceClasses.default;
+                    }
+
+                    workspace.workspaceClass = workspaceClass;
+                    this.workspaceDb.trace({ span }).store(workspace);
```
geropl, line 796: nit: there should be a .updatePartial() method which you can call with just `updatePartial({workspaceClass})`. This is slightly safer, as you never now who else does what changes to the `workspace` shape at hand.

geropl: Also: please add a trailing `.catch(..)` - or `await` the call.

components/gitpod-db/src/typeorm/entity/db-workspace.ts
```diff
@@ -121,4 +121,10 @@ export class DBWorkspace implements Workspace {
     })
     @Index("ind_basedOnSnapshotId")
     basedOnSnapshotId?: string;
+
+    @Column({
+        default: "",
+        transformer: Transformer.MAP_EMPTY_STR_TO_UNDEFINED,
+    })
+    workspaceClass?: string;
```
geropl, line 129: We already have that field on the WorkspaceInstance. Why do we need to add it here, now, as well? :thinking:

install/installer/pkg/components/server/configmap.go
```diff
@@ -237,6 +237,10 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 			// default limit for all cloneURLs
 			"*": 50,
 		},
+		WorkspaceClasses: WorkspaceClasses{
+			Default:              "g1-standard",
```
geropl, line 241: Ok, now I'm absolutely confused :confused:
 - I know `gitpodio-internal-xl` is a thing: is those one of the old values?
 - I recall discussions where we were talking about "Standard" and "XL" (I bet those are the descriptions, right? But for which classes??)
 - I see the following values thrown around in code:
   - ""
   - "default"
   - "standard"
   - "g1-large"
   - "g1-standard"

:exploding_head: 

@Furisto Could you add an explanatory table to the top of the PR, please?

And for the code in this PR: We shoudl remove _all_ usages of concrete string values from `server` code, and push that variance into the `workspaceClasses` config of this shape (or similar):

```
// Types
type WorkspaceClassesConfig = [WorkspaceClassConfig]
interface WorkspaceClassConfig {
  // The technical string we use to identify the class with internally
  workspaceClass: string;
  
  // Is the "default" class. The config is validated to only every have exactly _one_ default class.
  isDefault: boolean;
 
  // The string we display to users in the UI
  description: string;

  // Whether or not to:
  //  - offer users this Workspace class for selection
  //  - use this class to start workspaces with. If a user has a class marked like this configured and starts
  deprecated: boolean;

  // The price for this workspace class in "credits per minute"
  creditsPerMinute: number;
}
```
This should be sufficient to push all the string values into that. :lotus_position:

Furisto:
@geropl PTAL

components/server/src/config.ts
```diff
@@ -266,6 +292,15 @@ export namespace ConfigFile {
                 inactivityPeriodForRepos = config.inactivityPeriodForRepos;
             }
         }
+
+        let defaultClasses = config.workspaceClasses
+            .map((c) => (c.isDefault ? 1 : 0))
+            .reduce((acc: number, isDefault: number) => (acc + isDefault) as number, 0);
+
+        if (defaultClasses != 1) {
+            throw new Error("exactly one default workspace class needs to be configured, not " + defaultClasses);
```
geropl, line 301: :100:

components/server/src/workspace/workspace-starter.ts
```diff
@@ -345,7 +346,7 @@ export class WorkspaceStarter {
                 instance,
                 workspace,
                 user,
-                lastValidWorkspaceInstanceId,
+                lastValidWorkspaceInstance?.id ?? "",
```
geropl, line 349: In TypeScript, passing around "undefined" is the way to denote "not present". So `lastValidWorkspaceInstance?.id` (which is `string | undefined`) is way more in line with expectations than `string` (empty or not).

install/installer/pkg/components/server/configmap.go
```diff
@@ -237,6 +237,22 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 			// default limit for all cloneURLs
 			"*": 50,
 		},
+		WorkspaceClasses: []WorkspaceClass{
```
geropl, line 240: Is this the config we want to deliver for Self-Hosted? :thinking:

components/dashboard/src/settings/selectClass.tsx
```diff
@@ -65,9 +65,9 @@ export default function SelectWorkspaceClass(props: SelectWorkspaceClassProps) {
                     </SelectableCardSolid>
                     <SelectableCardSolid
                         className="w-36 h-32"
-                        title="XL"
-                        selected={workspaceClass === "XL"}
-                        onClick={() => actuallySetWorkspaceClass("XL")}
+                        title="Large"
+                        selected={workspaceClass === "g1-large"}
```
geropl, line 69: We really need to source these values from config :confused:

components/server/src/workspace/workspace-starter.ts
```diff
@@ -684,127 +684,166 @@ export class WorkspaceStarter {
...
+                    }
+
+                    if (workspace.type == "prebuild") {
+                        if (user.additionalData?.workspaceClasses?.prebuild) {
+                            workspaceClass = user.additionalData?.workspaceClasses?.prebuild;
+                        }
+                    }
+
+                    if (!workspaceClass) {
+                        workspaceClass = this.config.workspaceClasses.find((cl) => cl.isDefault)?.id ?? "";
```
geropl, line 804: I feel this should be in a function inside a namesapce `WorkspaceClasses` so that we can test it etc.
@Furisto Let me find a couple of minutes and I can provide a commit to show what I mean.

components/dashboard/src/settings/selectClass.tsx
Furisto: Yes! All of this should come from config, but for the skateboard I left it at hard coded.

components/server/src/workspace/workspace-starter.ts
geropl: Ah, did not check that. And we even do untyped equality checks against it... :see_no_evil: 
Ok, fixing this is out of scope here. :cry:

install/installer/pkg/components/server/configmap.go
Furisto: Good question. @kylos101 @atduarte Are we ok with these workspace classes for self-hosted?

components/dashboard/src/settings/selectClass.tsx
geropl: > but for the skateboard I left it at hard coded.

Yes, fine in this context. We can have a follow up PR for API + dynamic rendering.

install/installer/pkg/components/server/configmap.go
atduarte: Don't think we would want to have the two workspace classes as default for Self-Hosted. It's an advanced feature and the initial installation should be as simple as possible imho. 

Would it be hard to have the default for self-hosted only be the standard?

cc @lucasvaltl as he is responsible for defining the self-hosted offering

geropl: I think the idea was to deliver what we currently do for SH. This config should then be exposed under `experimental.webapp.workspaceClasses`, and the concrete values set [here](https://github.com/gitpod-io/ops/blob/main/deploy/production/meta-eu02/app/installer-config.yaml#L124).

components/server/src/workspace/workspace-starter.ts
geropl: @Furisto Here is a commit showing what I mean: https://github.com/gitpod-io/gitpod/commit/7412eb0f4ebc99f71a08bcd49a68f2c7f8a4e72f

I think it make sense to incorporate that commit + [exposing of that config under "experimental"](https://github.com/gitpod-io/gitpod/pull/11313/files#r924324654) into this PR, and then we should be fine to merge. :heavy_check_mark: 

:cloud: Ideally we should also have tests for those functions.

install/installer/pkg/components/server/configmap.go
lucasvaltl: Yeah default for self-hosted for now should be a single workspace class - to keep installation simple. Right now even that class is not made obvious - but is rather just a default that sets a request but no limits. 

We should however think about how to properly allow for configuration of this (and https://github.com/gitpod-io/gitpod/issues/10805 is likely a better place to discuss this). Fwiw, I don't think it makes sense to make this part of the installation config - this might be something a user wants to change over time, specifically a user that is not the ["installer persona"](https://www.notion.so/gitpod/Persona-v0-2-Installer-c1329ac36ae74bf191f3d0d0a2b43756) but rather an ["project configurer"](https://www.notion.so/gitpod/Persona-v0-2-Project-Configurer-ffef632ab79a45cc8188c7cf7ffafd3d) / admin.

install/installer/pkg/components/server/configmap.go
geropl: Has been resolve :rocket:

install/installer/pkg/config/v1/experimental/experimental.go
```diff
@@ -210,6 +211,14 @@ type UsageConfig struct {
 	CreditsPerMinuteByWorkspaceClass map[string]float64 `json:"creditsPerMinuteByWorkspaceClass"`
 }
 
+type WebAppWorkspaceClass struct {
```
geropl, line 214: :100:

geropl (approved):
Code LGTM (as discussed), tested and work as expected. :tada: 

Note that:
 - I could not test "g1-large" because preview envs now contain a single class only :heavy_check_mark: (the UI still allows to select two classes, which we should clean up, by making it dynamic based on the config, in a follow up PR, before the SH release [as discussed here](https://github.com/gitpod-io/gitpod/pull/11313/files#r924176203) /cc @lucasvaltl )
  - I did test default -> g1-standard after enabling the feature flag and the other scenarios :heavy_check_mark: 
  - on workspace stop, I always got this error: `container workspace completed; containers of a workspace pod are not supposed to do that. Reason: ` But I assume it's independent of this PR

Really looking forward to have this in prod! :orange_heart:

Pothulapati (approved):
So, I tried this on a `self-hosted` instance with default settings (using `local-preview`. This part of the build so not that hard to test. Just using [the relevant build image should be enough](https://www.gitpod.io/docs/self-hosted/latest/local-preview#1-running-the-docker-container))

1. I get two instances on the preferences settings
Screenshot 2022-07-21 at 10 33 32 AM

2. But setting them/changing them does not seem to do anything and always give me a `default` workspace class
```bash
/ # kubectl describe pod -l workspaceType=regular | grep workspaceClass
              gitpod.io/workspaceClass=default
              gitpod.io/workspaceClass=default
```

So, We should not even be having the preferences option then right? seems like they seem to be set in the UI https://github.com/gitpod-io/gitpod/blob/main/components/dashboard/src/settings/selectClass.tsx#L50. Shouldn't it instead be based on the config?

This could be a separate PR. Approving this hence.

/hold just in case.

Furisto:
@Pothulapati That's right. There will be a separate PR for dynamically displaying the workspace classes.

PR #11321, opened 2022-07-12
Refactor cert to create and wait in the background

liam-j-bennett:
## Description
Refactor the way we create and wait for certs.

These changes should lead to:
- Less failures, as we retry and wait 5 times
- Earlier failures, as we fail out before we start deploying things

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/.werft/jobs/build/deploy-to-preview-environment.ts
+++ b/.werft/jobs/build/deploy-to-preview-environment.ts
@@ -521,7 +521,7 @@ export async function issueMetaCerts(
     metaClusterCertParams.ip = getCoreDevIngressIP();
     metaClusterCertParams.bucketPrefixTail = "";
     metaClusterCertParams.additionalSubdomains = additionalSubdomains;
-    return await issueCertificate(werft, metaClusterCertParams, { ...metaEnv(), slice });
+    return issueCertificate(werft, metaClusterCertParams, { ...metaEnv(), slice });
 }
 
 async function installMetaCertificates(
--- a/.werft/jobs/build/deploy-to-preview-environment.ts
+++ b/.werft/jobs/build/deploy-to-preview-environment.ts
@@ -509,7 +509,7 @@ export async function issueMetaCerts(
     certsNamespace: string,
     domain: string,
     slice: string,
-) {
+): Promise<boolean> {
     const additionalSubdomains: string[] = ["", "*.", `*.ws.`];
     var metaClusterCertParams = new IssueCertificateParams();
     metaClusterCertParams.pathToTemplate = "/workspace/.werft/util/templates";
@@ -521,7 +521,7 @@ export async function issueMetaCerts(
     metaClusterCertParams.ip = getCoreDevIngressIP();
     metaClusterCertParams.bucketPrefixTail = "";
     metaClusterCertParams.additionalSubdomains = additionalSubdomains;
-    await issueCertificate(werft, metaClusterCertParams, { ...metaEnv(), slice });
+    return await issueCertificate(werft, metaClusterCertParams, { ...metaEnv(), slice });
 }
 
 async function installMetaCertificates(
--- a/.werft/util/certs.ts
+++ b/.werft/util/certs.ts
@@ -24,7 +24,7 @@ export class InstallCertificateParams {
     destinationKubeconfig: string;
 }
 
-export async function issueCertificate(werft: Werft, params: IssueCertificateParams, shellOpts: ExecOptions) {
+export async function issueCertificate(werft: Werft, params: IssueCertificateParams, shellOpts: ExecOptions): Promise<boolean> {
     var subdomains = [];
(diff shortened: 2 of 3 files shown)


.werft/jobs/build/deploy-to-preview-environment.ts
```diff
@@ -521,7 +521,7 @@ export async function issueMetaCerts(
     metaClusterCertParams.ip = getCoreDevIngressIP();
     metaClusterCertParams.bucketPrefixTail = "";
     metaClusterCertParams.additionalSubdomains = additionalSubdomains;
-    await issueCertificate(werft, metaClusterCertParams, { ...metaEnv(), slice });
+    return await issueCertificate(werft, metaClusterCertParams, { ...metaEnv(), slice });
```
ArthurSens, line 524: Do I understand correctly that the `await` here forces the process to wait for the whole execution of `issueCertificate` before continuing?

Isn't it the opposite of what you want to achieve?

ArthurSens (commented):
👋 

I'm not really proficient with Typescript, so I think it would be better if Mads could give a proper review here. Reading through the changes there was only one thing that looked off to me 🤔

.werft/jobs/build/deploy-to-preview-environment.ts
liam-j-bennett: My knowledge of TypeScript's syntax might be failing me. I'm treating running an async function in a similar way to go routines or Rust's `async` (although it seems like sync and async are not mutually exclusive in TS).

As this is wrapped by an async function (`var certReady = issueCertificate(werft, config);`), which we don't await on, it doesn't matter what goes on inside (i.e. we basically treat this synchronously). From my understanding, this means this function runs in the background, whether we have an `await` internal to the function or not.

I could remove the awaits and pass the promise from the downstream functions, but, from my understanding, it wouldn't impact this function

.werft/util/certs.ts
```diff
@@ -33,7 +33,46 @@ export async function issueCertificate(werft: Werft, params: IssueCertificatePar
 
     exec(`echo "Domain: ${params.domain}, Subdomains: ${subdomains}"`, { slice: shellOpts.slice });
     validateSubdomains(werft, shellOpts.slice, params.domain, subdomains);
-    createCertificateResource(werft, shellOpts, params, subdomains);
+
+    const maxAttempts = 5
+    const timeout = "150s"
+    var i: number
+    var certReady: boolean
```
mads-hartmann, line 40: Tiny nit: Would prefer to initialise and then not specify the type rather than specifying the type and not the initial value:

```ts
var i = 0
var certReady = false
```

.werft/util/certs.ts
```diff
@@ -33,7 +33,46 @@ export async function issueCertificate(werft: Werft, params: IssueCertificatePar
...
+    while(!certReady || i < maxAttempts) {
+        exec(`echo Creating cert: Attempt ${i}`);
+        createCertificateResource(werft, shellOpts, params, subdomains);
+        exec(`echo Checking for cert readiness: Attempt ${i}`);
```
mads-hartmann, line 44: Same as the other exec echo

.werft/util/certs.ts
```diff
@@ -33,7 +33,46 @@ export async function issueCertificate(werft: Werft, params: IssueCertificatePar
 
     exec(`echo "Domain: ${params.domain}, Subdomains: ${subdomains}"`, { slice: shellOpts.slice });
     validateSubdomains(werft, shellOpts.slice, params.domain, subdomains);
-    createCertificateResource(werft, shellOpts, params, subdomains);
+
+    const maxAttempts = 5
+    const timeout = "150s"
```
mads-hartmann, line 38: Tiny nit: As this is only used in checkCertReadiness I'd just move it there. Then it's also more clear why it's a string

mads-hartmann (approved):
I left a few minor comments about changes I'd like to see, but they can be done in a follow up PR if you like so I'll approve ☺️ I have added the hold label in case you want to take care of them in this PR.

.werft/jobs/build/deploy-to-preview-environment.ts
liam-j-bennett: Yes, this was my understanding. As there is no plan to add further concurrency into the function it makes sense to just return the promises upstream, rather than awaiting here

.werft/util/certs.ts
```diff
@@ -33,7 +33,46 @@ export async function issueCertificate(werft: Werft, params: IssueCertificatePar
...
+    var i: number
+    var certReady: boolean
+    while(!certReady || i < maxAttempts) {
+        exec(`echo Creating cert: Attempt ${i}`);
```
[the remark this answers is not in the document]
mads-hartmann: If we write to stdout without specifying the sliceID Werft will group the output in the "default" slice for the phase which can be rather confusing - there's not way to write the stdout and tell Werft to ignore it AFAIK ☺️

PR #11322, opened 2022-07-12
[server] Hook up usage API (4/5)

geropl:
## Description

Configures `server` to consume the actual `UsageService` over gRPC.

## Related Issue(s)

Context #10325 

## How to test
 - create a team with billing
 - go to team -> Billed Usage and see that empty list

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment


components/gitpod-protocol/src/usage.ts
```diff
@@ -24,10 +24,10 @@ export interface BillableSession {
     workspaceClass: string;
 
     // When the workspace started
-    startTime: number;
+    startTime: string;
 
     // When the workspace ended
-    endTime: number;
+    endTime: string;
```
geropl, line 30: I did this to follow the pattern we have going in virtually every other place in our JSON API. :+1:

components/gitpod-protocol/src/usage.ts
```diff
@@ -36,111 +36,4 @@ export interface BillableSession {
     projectId?: string;
 }
 
-export type BillableWorkspaceType = Omit<WorkspaceType, "probe">;
-
-export const billableSessionDummyData: BillableSession[] = [
```
easyCZ, line 41: Shall we keep this here an only append to the (for now empty) response from server? It would mean we're showing something in the interim which can be useful for demos but also iterating on the UI.

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2102,6 +2112,29 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
 
         await this.guardAccess({ kind: "costCenter", /*subject: costCenter,*/ owner }, operation);
     }
+
+    protected mapBilledSession(s: usage.BilledSession): BillableSession {
+        function mandatory<T>(s: T, m: (s: T) => string = (s) => "" + s): string {
+            if (!s) {
+                throw new Error("empty value in usage.BilledSession");
```
easyCZ, line 2119: Would be nice to also return which Instance ID had empty value, for easier debugging

easyCZ (approved):
/hold
for comments, up to you to make changes or not

components/gitpod-protocol/src/usage.ts
easyCZ: I think this is fine. The gRPC protocol will use `google.Time` which can be converted into a string easily

components/gitpod-protocol/src/usage.ts
geropl: Ah, good idea!

install/installer/pkg/components/server/configmap.go
```diff
@@ -222,6 +223,7 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		},
 		ContentServiceAddr:           "content-service:8080",
 		ImageBuilderAddr:             "image-builder-mk3:8080",
+		UsageServiceAddr:             fmt.Sprintf("%s:%d", usage.Component, usage.GRPCServicePort),
```
easyCZ, line 226: FYI, there's a utility called [`JoinHostPort`](https://pkg.go.dev/net#JoinHostPort) which can do this for you

geropl: That's a tad nicer indeed - added.

geropl:
/unhold

Now with dummydata again:
image

PR #11323, opened 2022-07-12
[kots]: delete channel on branch delete

MrSimonEmms:
## Description

GitHub Action to delete a KOTS channel on branch deletion. This will cut down on the amount of manual work that @gitpod-io/engineering-self-hosted have to do in housekeeping in the Replicated vendor portal.

In most cases, this will actually do nothing. Most branches do not create a channel in Replicated - the only time is if a PR is run with `/werft run publish-to-kots` when it will publish to a channel with the same name as the branch. This Action exists to delete these channels when the branch is deleted.

If a channel has customers assigned to it, it will not delete the channel as the workflow is that channels must have zero customers before they can be deleted - that's a failsafe against customer channels such as `stable` or `enterprise` being deleted. Also, this Action can never error as the `continue-on-error: true` param is set.

/hold Before this is merged, a secret will need to be added to our GitHub settings.

## How to test

I have built this using a [noddy repo](https://github.com/MrSimonEmms/github-action-delete-test) and done various tests to prove it works:
 - [delete channel](https://github.com/MrSimonEmms/github-action-delete-test/runs/7305418313?check_suite_focus=true)
 - [no error if channel doesn't exist](https://github.com/MrSimonEmms/github-action-delete-test/runs/7305475011?check_suite_focus=true)

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/.github/workflows/delete-kots-channel.yml
+++ b/.github/workflows/delete-kots-channel.yml
@@ -1,0 +1,32 @@
+name: Branch Deleted
+on: delete
+env:
+  REPLICATED_CLI_VERSION: 0.40.1
+  REPLICATED_API_TOKEN: ${{ secrets.REPLICATED_API_TOKEN }}
+  REPLICATED_APP: ${{ secrets.REPLICATED_APP }}
+jobs:
+  delete:
+    if: github.event.ref_type == 'branch'
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    steps:
+      - name: Install Replicated CLI
+        run: |
+          curl -sL https://github.com/replicatedhq/replicated/releases/download/v${{ env.REPLICATED_CLI_VERSION }}/replicated_${{ env.REPLICATED_CLI_VERSION }}_linux_amd64.tar.gz -o replicated.tar.gz
+          tar xf replicated.tar.gz replicated && rm replicated.tar.gz
+          mv replicated /usr/local/bin/replicated
+
+      - name: Delete Replicated channel
+        run: |
+          CHANNEL_ID=$(replicated channel inspect ${{ github.event.ref }} \
+            | grep "ID:" \
+            | sed "s/ID://" \
+            | sed "s/ //g" || true)
+
+          if [ "${CHANNEL_ID}" = "" ]; then
+            echo "No channel found"
+            exit 0
+          fi
+
+          # Allowed to fail if customers on channel - this will need to be manually deleted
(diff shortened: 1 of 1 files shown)


MrSimonEmms:
/hold cancel

The secrets have been added to GitHub Actions

.github/workflows/delete-kots-channel.yml
```diff
@@ -0,0 +1,32 @@
...
+            | sed "s/ID://" \
+            | sed "s/ //g" || true)
+
+          if [ "${CHANNEL_ID}" = "" ]; then
+            echo "No channel found"
+            exit 0
+          fi
+
+          # Allowed to fail if customers on channel - this will need to be manually deleted
+          replicated channel rm "${CHANNEL_ID}" || true
```
Pothulapati, line 32: TIOLI: How about we also do the check manually? Just want to be on the safer side, as its a v important operation.

Pothulapati (approved):
LGTM, apart from that question!

/hold just on the discussion. Feel free to remove this if you don't think that's necessary. :)

.github/workflows/delete-kots-channel.yml
MrSimonEmms: To the best of my knowledge, there's no way of checking if a channel has customers using the CLI. Honestly, it's all a little undocumented so I've had to dig through the [source code](https://github.com/replicatedhq/replicated/tree/main/cli/cmd). The closest we have is `channel counts` but keeps coming back with `Error: This feature is not supported for Kots applications.` which makes no sense.

It is documented that you cannot delete a channel with customers and there's no option to force delete. And we wouldn't habitually use branches with the channel names `stable`, `beta`, `unstable`, `enterprise`.

Pothulapati: Thanks for the elaborate answer. It makes sense.

MrSimonEmms: If there's any issues, we can always get revert this commit (I'm going to create/delete a branch now)

MrSimonEmms:
This works:

 - [branch deleted, no KOTS channel, no error](https://github.com/gitpod-io/gitpod/runs/7403781999?check_suite_focus=true)
 - [branch deletes, KOTS channel, deleted](https://github.com/gitpod-io/gitpod/runs/7404189597?check_suite_focus=true)

Pothulapati:
NICE @MrSimonEmms

PR #11331, opened 2022-07-12
[ws-daemon] do not fail workspace if git status failed during dispose 

sagor999:
## Description

(This PR is stacked on top of https://github.com/gitpod-io/gitpod/pull/11330, please review commit: https://github.com/gitpod-io/gitpod/commit/8d2d1c6d686a03efb2be274fb3831f534fbf8ded)

Some workspaces fail with `cannot get git status`, which is not a critical error.
Observed in some cases it failed due to potentially corrupt\misconfigured .git folder (potentially due to user's actions).

This PR would make it so that we don't fail workspace when this happens and instead continue disposing it.
For example, we already do that if workspace folder does not contain .git folder in it:
https://github.com/gitpod-io/gitpod/blob/905be0afd12bc43d348591225727ae5c3181982a/components/ws-daemon/pkg/internal/session/workspace.go#L309-L312

Main reasoning is that even if we cannot get git status, all that means is that we not going to show the user in dashboard uncommitted changes, which should not be a reason for failing whole workspace.

## Related Issue(s)

Fixes #

## How to test

Open workspace and corrupt your .git folder
Observe that workspace still finishes correctly, albeit without showing any uncommitted changes in dashboard.

## Release Notes

```release-note
none
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/ws-daemon/pkg/content/service.go
+++ b/components/ws-daemon/pkg/content/service.go
@@ -422,9 +422,12 @@ func (s *WorkspaceService) DisposeWorkspace(ctx context.Context, req *api.Dispos
 	// Update the git status prior to deleting the workspace
 	repo, err = sess.UpdateGitStatus(ctx, sess.PersistentVolumeClaim)
 	if err != nil {
+		// do not fail workspace because we were unable to get git status
+		// which can happen for various reasons, including user corrupting his .git folder somehow
+		// instead we log the error and continue cleaning up workspace
+		// todo(pavel): it would be great if we can somehow bubble this up to user without failing workspace
 		log.WithError(err).Error("cannot get git status")
 		span.LogKV("error", err.Error())
-		return nil, status.Error(codes.Internal, "cannot get git status")
 	}
 	if repo != nil {
 		resp.GitStatus = repo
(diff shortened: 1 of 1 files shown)


sagor999:
/hold
would like to get input from @csweichel on this one, so holding to make sure he has a chance to review

aledbf:
@sagor999 I think we need to update the configuration and configure `/mnt/workingarea` as safe using
`git config --global --add safe.directory *` instead of individual directories https://github.blog/2022-04-18-highlights-from-git-2-36/
(file /home/gitpod/.gitconfig in ws-daemon)

sagor999:
@aledbf I was referring to this particular case: 
https://cloudlogging.app.goo.gl/8GkTztnkJ5MdpwAz8
(check ws-daemon error log there)

I think what you are referring to is a frequent similar error that seems to be coming from prebuilds (unsafe directory error).
I will try to take a look at it as well, but it is out of scope for this particular PR (even though it is kind of related)

jenting (commented):
I like this approach 👍 
I encountered a scenario with the error `git status failed`, but the volume snapshot is ready to use.

In the scenario, we enable the PVC feature flag and manually make the node into a NotReady state (by disabling the k3s-agent on the host).
Then, the workspace pod becomes a Terminating state—finally, the VolumeSnapshot object is created by ws-manager and taken the snapshot of the PVC.
After relaunching the workspace pod, the workspace pod launches at another Ready node, and the user's content is recovered from the VolumeSnapshot.

Without this PR change, the dashboard shows the error `cannot get git status`.

The pro is we keep the user's content by PVC w/o data loss, and the con is we can't see the git difference in the dashboard.

sagor999:
something funky with unit tests in this build. each time it is a different test that fails, but opened it and did a leeway build and everything built fine...
going to try to rebase.

PR #11337, opened 2022-07-13
Call finalizeWorkspaceContent if the workspace Pod in Terminating

### LINKED ISSUE #11336: We did not backup content when the node goes to NotReady + Pod goes to Terminating
### Bug description

We did not backup content when the node goes NotReady and the workspace Pod is terminating.

### Steps to reproduce

1. Create 2 nodes, 1 control plane node, 1 worker node (using the workspace-preview).
2. Launch a workspace, and the workspace Pod should be located on the worker node 🙏 .
3. Stop the worker node by disabling the k3s-agent `systemctl disable k3s-agent`.
4. Waits for the node in NotReady state `kubectl get node -w`.
5. Waits for the workspace pod in Terminating state `kubectl get pod -l component=workspace -w`.

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

jenting:
## Description

When the node goes NotReady, the workspace Pod goes into a Terminating state.
In this case, the workspace Pod `status.containerStatuses.state` is still `Running`.

We should try to backup the workspace content if the Pod is Terminating and the underlying node is not ready or even gone.

This is the node taint if the node turns into a NotReady state.
image

This is the current workspace Pod spec.toleration
image

Therefore, for the cases
- `node.kubernetes.io/disk-pressure` and `node.kubernetes.io/memory-pressure`: the workspace Pod keeps in tolerance indefinitely. -> We would not handle this case because the tolerance second is not configured.
- `node.kubernetes.io/network-unavailable`: the workspace Pod tolerance duration is 30 seconds. -> We handle this case
- `node.kubernetes.io/not-ready` and `node.kubernetes.io/unreachable`: the workspace Pod tolerance duration is 300 seconds. -> We handle this case

After the current time - the node's taint.timeAdded > the workspace pod tolerance time, the ws-manager starts back up the content.

https://www.loom.com/share/8e0e870e6bed40809d4ac8ac1159b1e2

## Related Issue(s)

Fixes #11336

## How to test

1. Create 2 nodes, 1 control plane node, 1 worker node (using the workspace-preview).
2. Launch a workspace, and the workspace Pod should be located on the worker node 🙏 .
3. SSH into the worker node, disabling the k3s-agent `systemctl disable k3s-agent`.
4. Waits for the node in NotReady state `kubectl get node -w`.
5. Waits for the workspace pod in Terminating state `kubectl get pod -l component=workspace -w`.
6. Check the workspace pod content is back up successfully.

Note: after the node removal, the terminating pod will be removed by Kubernetes after a while. (About 1 minutes)

## Release Notes

```release-note
Try to backup content when the node goes into the NotReady state
```

## Documentation

None

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/ws-manager/pkg/manager/monitor.go
+++ b/components/ws-manager/pkg/manager/monitor.go
@@ -273,7 +273,7 @@ func (m *Monitor) onPodEvent(evt watch.Event) error {
 	}()
 
 	m.writeEventTraceLog(status, wso)
-	err = actOnPodEvent(ctx, m.act, status, wso)
+	err = actOnPodEvent(ctx, m.act, m.manager, status, wso)
 
 	// To make the tracing work though we have to re-sync with OnChange. But we don't want OnChange to block our event
 	// handling, thus we wait for it to finish in a Go routine.
@@ -287,7 +287,7 @@ func (m *Monitor) onPodEvent(evt watch.Event) error {
 
 // actOnPodEvent performs actions when a kubernetes event comes in. For example we shut down failed workspaces or start
 // polling the ready state of initializing ones.
-func actOnPodEvent(ctx context.Context, m actingManager, status *api.WorkspaceStatus, wso *workspaceObjects) (err error) {
+func actOnPodEvent(ctx context.Context, m actingManager, manager *Manager, status *api.WorkspaceStatus, wso *workspaceObjects) (err error) {
 	pod := wso.Pod
 
 	span, ctx := tracing.FromContext(ctx, "actOnPodEvent")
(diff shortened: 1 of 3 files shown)


components/ws-manager/pkg/manager/monitor.go
```diff
@@ -440,20 +440,8 @@ func actOnPodEvent(ctx context.Context, m actingManager, status *api.WorkspaceSt
...
-				terminated = c.State.Running == nil
-				break
-			}
-		}
-
-		_, gone := wso.Pod.Annotations[wsk8s.ContainerIsGoneAnnotation]
 		_, alreadyFinalized := wso.Pod.Annotations[startedDisposalAnnotation]
-
-		if (terminated || gone) && !alreadyFinalized {
+		if !alreadyFinalized {
```
sagor999, line 444: I am concerned about removing `terminated` check here. 
We want to make sure that we start backup ONLY when all processes have shutdown and container is stopped, otherwise we are risking trying to make a backup while something is terminating, and thus potentially making unstable backup.

But I guess in the case when node is not ready, k8s has no way to check the status of the container to ensure that it was stopped, hence it is stuck in fake running state.
What if we check the status of node here instead? And if it is not ready, then ignore terminated check and proceed making backup?

sagor999:
/hold
to prevent auto merge

components/ws-manager/pkg/manager/monitor.go
sagor999: >You mean a case where the node is NotReady and the Pod is still in Running state (haven't changed to Terminating state), and the user manually stops the workspace? I could test this scenario as well.

No, I am referring to a normal case: workspace is stopping. We wait until container is stopped, and only then start backup. 
If you remove `terminated` check here, as far as I understand it would mean we will attempt to run a backup right after pod transitions into terminating state, even though container is still running and is terminating\shutting down.
Or am I wrong here?

jenting: Yes, you're right. The Pod is terminating; it executes the preStop hook (if it existed), and then sends a TERM signal to the container.

If during the workspace Pod graceful shutdown period, it generates more content, and either ws-daemon _or_ PVC snapshot finished the content backup, then the unstable backup happens.

sagor999 (approved):
:rocket:

PR #11311, opened 2022-07-12
Add migration for new usage table

andrew-farries:
## Description

Define the schema and add the migration for a new `d_b_workspace_instance_usage` table to store workspace usage data.

## Related Issue(s)
Part of https://github.com/gitpod-io/gitpod/issues/10323

## How to test

Connect to the database in the preview environment for this branch and see the new table in the database:

image

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

Issue #10323, opened 2022-05-29
Setup usage datastore and store usage


easyCZ:
Depends on https://github.com/gitpod-io/gitpod/issues/10320

We want to start persisting our usage report such that it can be viewed and queried quickly (from the Public API, and dashboard). In this task, we're defining the datastore and persisting the report into it.

* [x] Create table to store usage data (https://github.com/gitpod-io/gitpod/pull/11311)
* [x] Persist data on every reconcilation run (https://github.com/gitpod-io/gitpod/pull/11343)
  * [x] Populate `creditsUsed` columns (https://github.com/gitpod-io/gitpod/pull/11406)

andrew-farries:
We decided:

1. Store usage data in a new table in the existing db, called `workspace_instance_usage` or similar. This will have fields: `instanceId`, `attributionId`, `startTime`, `stopTime`, `creditsUsed` and some kind of `generationId` to show the last report in which the instance appeared.

2. Separately, we would also like to persist the raw usage report (as json), possibly using `content-service` so that we have an audit trail. The `generationId` field in the usage table could then be a link to this report.

geropl:
That sounds good to me. I'm not sure we need 2. (the document store), though, but happy to learn about the motivation once we ticked 1.

Regarding the schema: I did not check all fields, but even given a `generationId`, a `generationTimestamp` or so would greatly help with debugging, at the very least.

PR #11343, opened 2022-07-13
[usage] Persist usage reports in database

andrew-farries:
## Description

Store the usage report in the database table created in #11311 whenever usage reconciliation runs.

⚠️ This PR does not yet populate the `creditsUsed` or `generationId` fields ⚠️ 

## Related Issue(s)

Part of #10323 

## How to test

Run the usage component against the database in this PR's preview environment:

* Port forward to the database in the preview env
* Edit the local config `components/usage/config.json` for the usage controller to run every 10s.
* Run the usage component:
```bash
DB_USERNAME=gitpod DB_PASSWORD=<> DB_HOST=localhost DB_PORT=3306 go run . run
```
(get the password from `server` environment)
* Wait a while for reconciliation to run a couple of times.
* Connect to the port-forwarded database with eg `mycli` and observe the data in `d_b_workspace_instance_usage`.

Unit tests.

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage/pkg/controller/reconciler.go
+++ b/components/usage/pkg/controller/reconciler.go
@@ -213,7 +213,7 @@ func trimStartStopTime(instances []db.WorkspaceInstanceForUsage, maximumStart, m
 	return updated
 }
 
-func groupInstancesByAttributionID(instances []db.WorkspaceInstanceForUsage) map[db.AttributionID][]db.WorkspaceInstanceForUsage {
+func groupInstancesByAttributionID(instances []db.WorkspaceInstanceForUsage) UsageReport {
 	result := map[db.AttributionID][]db.WorkspaceInstanceForUsage{}
 	for _, instance := range instances {
 		if _, ok := result[instance.UsageAttributionID]; !ok {
--- a/components/usage/pkg/controller/reconciler.go
+++ b/components/usage/pkg/controller/reconciler.go
@@ -6,6 +6,7 @@ package controller
 
 import (
 	"context"
+	"database/sql"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -84,6 +85,11 @@ func (u *UsageReconciler) Reconcile() (err error) {
 		return fmt.Errorf("failed to get file stats: %w", err)
 	}
 	log.Infof("Wrote usage report into %s", filepath.Join(dir, stat.Name()))
+
+	err = db.CreateUsageRecords(ctx, u.conn, usageReportToUsageRecords(report))
+	if err != nil {
+		return fmt.Errorf("failed to write usage records to database: %s", err)
+	}
 
 	return nil
 }
@@ -224,4 +230,29 @@ func groupInstancesByAttributionID(instances []db.WorkspaceInstanceForUsage) Usa
 	}
 
 	return result
+}
+
+func usageReportToUsageRecords(report UsageReport) []db.WorkspaceInstanceUsage {
+	var usageRecords []db.WorkspaceInstanceUsage
+
+	for attributionId, instances := range report {
+		for _, instance := range instances {
+			var stoppedAt sql.NullTime
+			if instance.StoppedTime.IsSet() {
+				stoppedAt.Time = instance.StoppedTime.Time()
+			}
+
+			usageRecords = append(usageRecords, db.WorkspaceInstanceUsage{
+				WorkspaceID:   instance.ID,
+				AttributionID: attributionId,
+				StartedAt:     instance.CreationTime.Time(),
+				StoppedAt:     stoppedAt,
+				CreditsUsed:   0,
+				GenerationId:  0,
+				Deleted:       false,
+			})
+		}
+	}
+
+	return usageRecords
 }
--- a/components/usage/pkg/controller/reconciler.go
+++ b/components/usage/pkg/controller/reconciler.go
@@ -243,7 +243,7 @@ func usageReportToUsageRecords(report UsageReport) []db.WorkspaceInstanceUsage {
 			}
 
 			usageRecords = append(usageRecords, db.WorkspaceInstanceUsage{
-				WorkspaceID:   instance.ID,
+				InstanceID:    instance.ID,
 				AttributionID: attributionId,
 				StartedAt:     instance.CreationTime.Time(),
 				StoppedAt:     stoppedAt,
--- a/components/usage/pkg/controller/reconciler.go
+++ b/components/usage/pkg/controller/reconciler.go
@@ -239,7 +239,7 @@ func usageReportToUsageRecords(report UsageReport) []db.WorkspaceInstanceUsage {
 		for _, instance := range instances {
 			var stoppedAt sql.NullTime
 			if instance.StoppedTime.IsSet() {
-				stoppedAt.Time = instance.StoppedTime.Time()
+				stoppedAt = sql.NullTime{Time: instance.StoppedTime.Time(), Valid: true}
 			}
 
 			usageRecords = append(usageRecords, db.WorkspaceInstanceUsage{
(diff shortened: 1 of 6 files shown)


components/usage/pkg/db/workspace_instance_usage_test.go
```diff
@@ -0,0 +1,102 @@
...
+						},
+					},
+				},
+				{
+					teamAttributionID: {
+						db.WorkspaceInstanceForUsage{
+							// the same workspace again
+							ID:                 instanceID,
+							UsageAttributionID: teamAttributionID,
+							WorkspaceClass:     "default",
```
easyCZ, line 84: There should be a constant for this `default` value

components/usage/pkg/db/workspace_instance_usage_test.go
```diff
@@ -0,0 +1,102 @@
...
+	ctx := context.Background()
+	for _, scenario := range scenarios {
+		t.Run(scenario.Name, func(t *testing.T) {
+			require.NoError(t, db.CreateUsageRecords(ctx, conn, scenario.Report))
+		})
+	}
+}
+
+func TestCanHandleMultipleReportsWithDuplicateEntries(t *testing.T) {
+	conn := dbtest.ConnectForTests(t)
```
easyCZ, line 53: I'd recommend moving the `conn` into each test scenario, to ensure that after each test the data is deleted before the next test run

components/usage/pkg/controller/reconciler.go
```diff
@@ -85,6 +85,8 @@ func (u *UsageReconciler) Reconcile() (err error) {
 	}
 	log.Infof("Wrote usage report into %s", filepath.Join(dir, stat.Name()))
 
+	db.CreateUsageRecords(ctx, u.conn, report)
```
easyCZ: Should handle error

components/usage/pkg/db/workspace_instance_usage.go
```diff
@@ -0,0 +1,61 @@
...
+	GenerationId  int           `gorm:"column:generationId;type:int;" json:"generationId"`
+	Deleted       bool          `gorm:"column:deleted;type:tinyint;default:0;" json:"deleted"`
+}
+
+// TableName sets the insert table name for this struct type
+func (u *WorkspaceInstanceUsage) TableName() string {
+	return "d_b_workspace_instance_usage"
+}
+
+func CreateUsageRecords(ctx context.Context, conn *gorm.DB, report map[AttributionID][]WorkspaceInstanceForUsage) error {
```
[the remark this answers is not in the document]
easyCZ: Or actually, it should take `[]WorkspaceInstanceUsage`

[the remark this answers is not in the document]
andrew-farries: done. The conversion between reports and `[]WorkspaceInstanceUsage` is done in the reconciler.

components/usage/pkg/db/workspace_instance_usage_test.go
andrew-farries: done

components/usage/pkg/db/workspace_instance_usage_test.go
```diff
@@ -0,0 +1,102 @@
...
+				},
+			},
+		},
+	}
+
+	ctx := context.Background()
+	for _, scenario := range scenarios {
+		t.Run(scenario.Name, func(t *testing.T) {
+			for _, report := range scenario.Reports {
+				require.NoError(t, db.CreateUsageRecords(ctx, conn, report))
```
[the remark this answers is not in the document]
andrew-farries: I've added `t.Cleanup` calls to each subtest to delete the usage records between each subtest.

components/usage/pkg/db/workspace_instance_usage_test.go
```diff
@@ -0,0 +1,112 @@
...
+				Deleted:       false,
+			}},
+		},
+	}
+
+	for _, scenario := range scenarios {
+		conn := dbtest.ConnectForTests(t)
+		t.Run(scenario.Name, func(t *testing.T) {
+			t.Cleanup(func() {
+				require.NoError(t, conn.Where("1=1").Delete(&db.WorkspaceInstanceUsage{}).Error)
```
easyCZ, line 58: This was the cause of the problems with flaky tests. When more tests exist and do this, it deletes the data needed for other runs of the test. That's why the current approach adds a hook for each ID created to only deleted the IDs created in each test.

components/usage/pkg/db/workspace_instance_usage_test.go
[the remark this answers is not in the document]
easyCZ: Commented in https://github.com/gitpod-io/gitpod/pull/11343/files#r919996694

components/usage/pkg/db/workspace_instance_usage_test.go
andrew-farries: 🤔 

None of these tests call `t.Parallel`, so AIUI the tests in this file will run in series, and within each test each subtest will also run in series. So where is the risk of having a subtest clearing the table after it finishes?

Is it tests running in different packages (which will be run in parallel by `go test`) that is the concern here?

easyCZ: https://github.com/gitpod-io/gitpod/pull/10971 has extra context on this, including a link to some literature

andrew-farries: > But tests in different files will be run concurrently.

I don't think that's correct. See for example [here](https://engineering.mercari.com/en/blog/entry/20220408-how_to_use_t_parallel/):

> By default, execution of test code using the testing package will be done sequentially. However, note that it is only the tests within a given package that run sequentially.

> If tests from multiple packages are specified, the tests will be run in parallel at the package level. For example, imagine there are two packages, package a and package b. The test code in package a will be run sequentially, and the test code in package b will also be run sequentially. However, the tests for package a and package b will be run in parallel. Let’s look more closely into how these will be run in parallel.

So I guess I'm struggling to understand how test cleanup code within a package can interfere with other tests in the same package if there is no `t.Parallel` anywhere. Maybe there is a race in actually deleting the rows from the database?

easyCZ: The issues we observed were across different packages. And the same will happen here once we add proper tests to the controller package.

andrew-farries: test cleanup changed to only remove records that were created by each test.

components/usage/pkg/db/workspace_instance_usage.go
```diff
@@ -0,0 +1,39 @@
...
+	"database/sql"
+	"time"
+
+	"github.com/google/uuid"
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+)
+
+type WorkspaceInstanceUsage struct {
+	WorkspaceID   uuid.UUID     `gorm:"primary_key;column:workspaceId;type:char;size:36;" json:"workspaceId"`
```
geropl, line 18: Does it make sense to merge the other PR with the rename into this one...? I though it had already been merged.

:bulb: Or has the table defintion, but not the Go struct?

geropl:
Doing a quick test...

geropl:
Works as advertised. I understood that credits and generationId will be set later.

Only thing I noticed is that `stoppedAt` is not set, while the workspace_instance is definitely `stopped` (and has a `stoppedTime`).
image

@andrew-farries Any idea why that might be?

andrew-farries:
> Only thing I noticed is that stoppedAt is not set, while the workspace_instance is definitely stopped (and has a stoppedTime).

Thanks. I've fixed this and added tests for the conversion from a usage report to instance usage records. (f25999dcf and b9016e9de).

geropl (approved):
Note that the current preview env is in [an undefined state](https://werft.gitpod-dev.com/job/gitpod-build-af-store-usage-data.11).

But I applied the migration manually, ran the test and :tada: : it worked. :lotus_position: 
image

andrew-farries:
/werft run with-clean-slate-deployment=true with-preview=true

geropl:
@andrew-farries What about runnign `with-preview=false`? I already tested twice and am confident it works, so let's get this in! :ship:

PR #11346, opened 2022-07-13
Add phone verification

### LINKED ISSUE #11339: Add Phone Verification for new Users
In order to reduce the ever increasing amount of abuse on Gitpod, we'd like to make it hard for such people to get onto Gitpod. The hypothesis is that phone number verification would add an extra step that can reduce the amount.

The verification should happen before after signup but before they can start their first workspace. We could further reduce the friction for good users by whitelisting based on certain criteria (e.g. age of GH or GL account). 

The verification should only be active on SaaS and we should make use of an external service such as Twillio's Verify.

See **[design specs](https://github.com/gitpod-io/gitpod/issues/11339#issuecomment-1183387915)** below. 🥯

svenefftinge:
## Description
Adds phone number verification to make abusing Gitpod harder

## Related Issue(s)
Fixes #11339

## Release Notes

```release-note
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/WORKSPACE.yaml
+++ b/WORKSPACE.yaml
@@ -12,7 +12,7 @@ defaultArgs:
   intellijDownloadUrl: "https://download.jetbrains.com/idea/ideaIU-2022.2.1.tar.gz"
   golandDownloadUrl: "https://download.jetbrains.com/go/goland-2022.2.2.tar.gz"
   pycharmDownloadUrl: "https://download.jetbrains.com/python/pycharm-professional-2022.2.1.tar.gz"
-  phpstormDownloadUrl: "https://download.jetbrains.com/webide/PhpStorm-2022.2.tar.gz"
+  phpstormDownloadUrl: "https://download.jetbrains.com/webide/PhpStorm-2022.2.1.tar.gz"
 provenance:
   enabled: true
   slsa: true


roboquat:

<details>

</details>

components/server/src/config.ts
```diff
@@ -240,6 +248,20 @@ export namespace ConfigFile {
                 log.error("Could not load Stripe config", error);
             }
         }
+        let twilioConfig:
+            | {
+                  serviceName: string;
```
geropl, line 253: `let twilioConfig: ConfigSerialized["twilioConfig"] = undefined;`

would work here as well.

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -457,6 +460,37 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
...
+    ): Promise<boolean> {
+        const user = this.checkUser("verifyPhoneNumberVerificationToken");
+        const checked = await this.phoneVerificationService.verifyVerificationToken(phoneNumber, token);
+        if (!checked) {
+            return false;
+        }
+        if (!user.additionalData) {
+            user.additionalData = {};
+        }
+        if (!user.additionalData?.profile) {
```
geropl, line 486: nit: this cascade could be wrapped in a `User.updateProfile(partial: Partial<User.Profile>)` method.

components/server/src/auth/phone-verification-service.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public async needsVerification(user: User): Promise<boolean> {
+        if (!this.config.twilioConfig) {
+            return false;
+        }
+        if (!!user.additionalData?.profile?.verifiedPhoneNumber) {
+            return false;
+        }
+        return true;
```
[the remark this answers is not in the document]
svenefftinge: It's not going to be rolled out like this. Still a draft, please focus on the config juggling at this point 😁 . Plan is to only ask for users that haven't started a workspace before and also not ask new users if they are using an aged and used github account.

install/installer/pkg/components/server/deployment.go
```diff
@@ -235,6 +235,29 @@ func deployment(ctx *common.RenderContext) ([]runtime.Object, error) {
 		return nil
 	})
 
+	_ = ctx.WithExperimental(func(cfg *experimental.Config) error {
+		if cfg.WebApp != nil && cfg.WebApp.Server != nil && cfg.WebApp.Server.TwilioConfig != "" {
+			twilioConfig := cfg.WebApp.Server.TwilioConfig
+
+			volumes = append(volumes,
+				corev1.Volume{
+					Name: "stripe-config",
```
geropl, line 244: copy-n-paste: twilio-config

components/server/src/auth/verification-service.ts
```diff
@@ -0,0 +1,79 @@
...
+            const client = new Twilio(this.config.twilioConfig.accountSID, this.config.twilioConfig.authToken);
+            this.verifyService = client.verify.v2.services(this.config.twilioConfig.serviceID);
+        }
+    }
+
+    public async needsVerification(user: User): Promise<boolean> {
+        if (!this.config.twilioConfig) {
+            return false;
+        }
+        // if (!!user.additionalData?.lastVerificationTime) {
```
svenefftinge, line 35: I have disabled this code, so it is easier to test multiple times.

MrSimonEmms (approved):

components/server/src/auth/verification-service.ts
```diff
@@ -0,0 +1,79 @@
...
+        user.additionalData!.lastVerificationTime = new Date().toISOString();
+        await this.userDB.updateUserPartial(user);
+        return user;
+    }
+
+    // private async isTrusted(user: User): Promise<boolean> {
+    //     // we only want to ask new users, so if a user has ever started workspaces before, we trust them.
+    //     // there might be a few false positives because all workspaces might have been deleted.
+    //     const workspaces = await this.workspaceDB.findWorkspacesByUser(user.id);
+    //     return workspaces.length !== 0;
```
svenefftinge, line 58: @geropl I think it makes sense to check for usage instead?

geropl: Jep, that could work as an indicator, with a threshold > what we see from abusers.

lucasvaltl:
I'm assuming this is only "turned on" in SaaS?

geropl:
/hold Due to [this comment](https://github.com/gitpod-io/gitpod/pull/11346#discussion_r950251073).

@svenefftinge While (un-)commenting this part of the code is not harmful we (WebApp) decided after a recent incident that we hide all "dev commits" behind Feature Flags: either the whole feature, or the "dev" part.

geropl:
Starting to review now.

svenefftinge:
Thanks for the hold. FWIW the logic for "trusted" if going to get a bit more complex as we don't want to ask every new user. I'll try to judge trustworthiness also based on age of GitHub/GitLab account. The rest is ready for review.

geropl:
> FWIW the logic for "trusted" if going to get a bit more complex as we don't want to ask every new user.

Jep! I guess that's also something we need/want to fine-tune as we go.

svenefftinge:
I'm spinning up a preview env and will update the env var on the server deployment, so trillio works.

components/gitpod-protocol/src/util/logging.ts
```diff
@@ -6,7 +6,7 @@
 
 const inspect: (object: any) => string = require("util").inspect; // undefined in frontend
 
-let plainLogging: boolean = false; // set to true during development to get non JSON output
+let plainLogging: boolean = true; // set to true during development to get non JSON output
```
geropl, line 9: This should also be part of the "dev commit". Ideally, we should not need this, at all. 

:cloud: I think we had a de-jsonifier in the past. Could we use that together with `yarn telepresence` to avoid the need for this code change? Or maybe set an env var, or similar?

svenefftinge: Yes, I'll remove it as well, of course.

components/gitpod-protocol/src/gitpod-service.ts
```diff
@@ -85,6 +85,9 @@ export interface GitpodServer extends JsonRpcServer<GitpodClient>, AdminServer,
     getLoggedInUser(): Promise<User>;
     getTerms(): Promise<Terms>;
     updateLoggedInUser(user: Partial<User>): Promise<User>;
+    needsVerification(): Promise<boolean>;
```
geropl, line 88: Do we need the method `needsVerification()`? I don't see it being used in the frontend (anymore). Feels like a leftover from an ealier version?

components/server/ee/src/billing/entitlement-service.ts
```diff
@@ -31,13 +32,21 @@ export class EntitlementServiceImpl implements EntitlementService {
...
     @inject(EntitlementServiceUBP) protected readonly ubp: EntitlementServiceUBP;
+    @inject(VerificationService) protected readonly verificationService: VerificationService;
 
     async mayStartWorkspace(
         user: User,
         date: Date = new Date(),
         runningInstances: Promise<WorkspaceInstance[]>,
     ): Promise<MayStartWorkspaceResult> {
         try {
+            const verification = await this.verificationService.needsVerification(user);
```
geropl, line 43: :+1:

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -464,6 +467,30 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
         return user;
     }
 
+    public async needsVerification(ctx: TraceContext): Promise<boolean> {
+        const user = this.checkUser("needsVerification");
+        return this.verificationService.needsVerification(user);
+    }
+
+    public async sendPhoneNumberVerificationToken(ctx: TraceContext, phoneNumber: string): Promise<void> {
+        this.checkUser("sendPhoneNumberVerificationToken");
```
geropl, line 476: This whole concern feels like it should be part of the EE code-base, and this methods should be empty stubs (like https://github.com/gitpod-io/gitpod/blob/7f2ab4dd96ca9c5494b9fe859601ffbfefe91cfd/components/server/src/workspace/gitpod-server-impl.ts#L2744).

components/server/src/auth/verification-service.ts
```diff
@@ -0,0 +1,79 @@
...
+import { User } from "@gitpod/gitpod-protocol";
+import { log } from "@gitpod/gitpod-protocol/lib/util/logging";
+import { inject, injectable, postConstruct } from "inversify";
+import { Config } from "../config";
+import { Twilio } from "twilio";
+import { ServiceContext } from "twilio/lib/rest/verify/v2/service";
+import { UserDB, WorkspaceDB } from "@gitpod/gitpod-db/lib";
+
+@injectable()
+export class VerificationService {
```
geropl, line 16: Currently this class is part of the community edition, but only bound in the EE container.

As mentioned in GitpodServerImpl, I think it should move to the EE side entirely.

components/server/ee/src/container-module.ts
```diff
@@ -86,6 +87,7 @@ export const productionEEContainerModule = new ContainerModule((bind, unbind, is
     bind(BitbucketAppSupport).toSelf().inSingletonScope();
     bind(GitHubEnterpriseApp).toSelf().inSingletonScope();
     bind(BitbucketServerApp).toSelf().inSingletonScope();
+    bind(VerificationService).toSelf().inSingletonScope();
```
geropl, line 90: This was it's only bound for the EE edition. See other comments maybe adjust.

components/server/src/workspace/gitpod-server-impl.ts
svenefftinge: What is the reasoning you believe it should be EE? I have no concerns about making this open-source as I don't see how it could be something we want to monetize.

components/server/src/config.ts
```diff
@@ -240,6 +253,22 @@ export namespace ConfigFile {
                 log.error("Could not load Stripe config", error);
             }
         }
+        let twilioConfig: Config["twilioConfig"];
+        log.info("loading twilio config");
+        if (!!env.TWILIO_CONFIG) {
```
geropl, line 258: Why the fallback with the env var? If it's for dev purposes, I think the same can be achieved with pushing and manipulating a Secret.

:cloud: We invested a bit of effort last year to move to a central config file which is far easier to reason about (not part of config file? Does not influence the config).

svenefftinge:
I have updated the server deployment, but noticed that there are still styling issues with the phone input. Strangely they are not there when running in dev mode 🙄

geropl:
> I have updated the server deployment, but noticed that there are still styling issues with the phone input.

Reminds me of last week, where a werft job "missed" to pick-up the "with-preview" config: build was green, but was still looking at old code.

Update: [last job](https://werft.gitpod-dev.com/job/gitpod-build-sefftinge-add-phone-verification-11339.14) worked, though.

components/server/src/workspace/gitpod-server-impl.ts
geropl: Mainly because I thought it would not be that interesting. But have no hard feelings here!

:bulb: Then the binding should move to the other container: https://github.com/gitpod-io/gitpod/pull/11346#discussion_r951072047

components/server/src/config.ts
svenefftinge: Makes sense. I would prefer to inline the twillioConfig in the `server-config` instead of copying what I see with the twilio-config, where there's a reference to another configmap. WDYT?

components/server/src/workspace/gitpod-server-impl.ts
geropl: > My take is to put as much as possible into the open-source version

That has been our strategy so far, yes, and I agree that it applies to phone verification as well. Put it into the "billing" basket at first, but it's unrelated. :100: . Sorry for creating confusion here. :lotus_position:

PR #11353, opened 2022-07-13
[usage] Rename misnamed column in usage table

andrew-farries:
## Description

Change the name of the `workspaceId` column to `instanceId` in the `d_b_workspace_instance_usage` table as that's what it stores.

As there is no production data in this table the migration is simply to drop the table and recreate it.

/hold because this is based on https://github.com/gitpod-io/gitpod/pull/11343.

## Related Issue(s)
Part of https://github.com/gitpod-io/gitpod/issues/10323

## How to test

Connecting to the database in the preview env for this PR shows the table with the correct column name.

## Release Notes
```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-db/src/typeorm/migration/1657722262390-RenameWorkspaceIdColumn.ts
+++ b/components/gitpod-db/src/typeorm/migration/1657722262390-RenameWorkspaceIdColumn.ts
@@ -1,0 +1,27 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class RenameWorkspaceIdColumn1657722262390 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`DROP INDEX \`IDX_e759ab5fcf57350da51fcf56bc\` ON \`d_b_workspace_instance_usage\``);
+        await queryRunner.query(`DROP INDEX \`IDX_25d77dfa246b93672c317e26ad\` ON \`d_b_workspace_instance_usage\``);
+        await queryRunner.query(`DROP INDEX \`IDX_1358af969a29fd9e0c6cabf37c\` ON \`d_b_workspace_instance_usage\``);
+        await queryRunner.query(`DROP TABLE \`d_b_workspace_instance_usage\``);
+
+        await queryRunner.query(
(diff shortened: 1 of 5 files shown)


components/gitpod-db/src/typeorm/migration/1657722262390-RenameWorkspaceIdColumn.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class RenameWorkspaceIdColumn1657722262390 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`DROP INDEX \`IDX_e759ab5fcf57350da51fcf56bc\` ON \`d_b_workspace_instance_usage\``);
+        await queryRunner.query(`DROP INDEX \`IDX_25d77dfa246b93672c317e26ad\` ON \`d_b_workspace_instance_usage\``);
+        await queryRunner.query(`DROP INDEX \`IDX_1358af969a29fd9e0c6cabf37c\` ON \`d_b_workspace_instance_usage\``);
+        await queryRunner.query(`DROP TABLE \`d_b_workspace_instance_usage\``);
```
geropl, line 14: :bulb: I think you can also do a 3-way-move with just dropping the PRIMARY KEY constraint (but maintaining the invariant that a table has to have at least one all the time). Sharing in case it's helpful in future scenarios, but this works here as well. :+1:

components/gitpod-db/src/typeorm/migration/1657722262390-RenameWorkspaceIdColumn.ts
```diff
@@ -0,0 +1,27 @@
...
+
+        await queryRunner.query(
+            `CREATE TABLE \`d_b_workspace_instance_usage\` (\`instanceId\` char(36) NOT NULL, \`attributionId\` varchar(255) NOT NULL, \`startedAt\` timestamp(6) NOT NULL, \`stoppedAt\` timestamp(6) NULL, \`creditsUsed\` double NOT NULL, \`generationId\` int NOT NULL, \`deleted\` tinyint NOT NULL, INDEX \`IDX_1358af969a29fd9e0c6cabf37c\` (\`attributionId\`), INDEX \`IDX_25d77dfa246b93672c317e26ad\` (\`startedAt\`), INDEX \`IDX_e759ab5fcf57350da51fcf56bc\` (\`stoppedAt\`), PRIMARY KEY (\`instanceId\`)) ENGINE=InnoDB`,
```
geropl, line 17: Regarding indexes: We have the convention to give them human readable names, e.g. one on `attributionId` would be `ind_attributionId`, one on `attributionId` and `creationTime` `ind_attributionId_creationTime`.

components/gitpod-db/src/typeorm/migration/1657722262390-RenameWorkspaceIdColumn.ts
geropl: You can specify those in `[db-workspace-instance-usage.ts](https://github.com/gitpod-io/gitpod/pull/11353/files#diff-30691deb4aa982d2d0f5b5aa25e12296a1a308aa7a2681bbe92232199bfa945b)` `@Index(name)`, and they will get generated in here.

Issue #10764, opened 2022-06-20
Record a "prebuild trigger" for every webhook event


geropl:
Currently users cannot distinguish at-all/easily between:
 - no webhook received at all (installation misconfigured)
 - webhook received but not run due to global rate-limiter
 - webhook received but not run because of user config
 - what caused a particular prebuild to be executed

This issue is the first one in a series that seeks to enable the user to understand just that. To that end, we need to keep track of "prebuild triggers".

To illustrate the idea, this could be possible states for such a prebuild trigger:
 - `created`: webhook was received
 - `ignored_user`: user configured rate limit, not run due to user config, ouf of money, ...
 - `ignored_system`: sanity rate limit, ...
 - `enqueued`: prebuild entry was created (in `queued` state)

A schema could/should contain (at least):
 - trigger id (pk)
 - creationTime
 - triggeredBy: `manual` | `webhook` | `app` (?)
 - state: `created` | `dropped_user` | `dropped_system` | `enqueued`
 - message (string, details on the state)
 - context (JSON, whatever we need from the trigger to start the prebuild)
 - prebuildId (if in `enqueued` state)

AlexTugarev:
Before starting with manifesting the schema, let's take a look at the flow to be backed by it. We need to take into account, that there is a status quo and some iterations to reach the wanted state.


1. We have 4 kinds of webhook receivers for different Git services. Each of them are sending (push) events in a specific shape and level of details. (Some of those details can actually replace the following API call, which are currently done afterwards to gather information to be rendered.)
2. From a push event we derive the context details (context URL, project association, prebuild details, etc.); the context URL is fundamental for the subsequent queries. 
3. A prebuild manager takes care of the actual start requests. Ideally, it should not need to differentiate between provider specific details.
4. Currently, a received push event is directly translated to a prebuild manager call – except for basic filter criteria.


5. Goal: we want to be able to tell when a push event was registered (per context/branch/name-it).
6. Goal: we want to be able to tell if a push event caused a prebuild start.
7. (Future) goal: event reception to be decoupled from prebuild triggering. This should allow for more control.

svenefftinge:
Would be indeed great to generalize the event types and properties over the various gitproviders. @AlexTugarev are you making a proposal?

PR #11356, opened 2022-07-13
Record webhook events

AlexTugarev:
## Description
This PR adds recording of webhook events received for GH/GHE/GL/BB/BBS.

For each `push` event received for any configured repositories, a record will be stored in the DB containing the raw payload necessary to trigger a prebuild at any time. To start with a 🛹, the actual trigger remains untouched here, only the status is updated up the the point of a successful prebuild trigger (or any error case.) The status updates will be already beneficial in the follow up PR to render updates on the dashboard.

The proposed DB table is quite simple. 

```
mysql> describe d_b_webhook_event;
+------------------+--------------+------+-----+----------------------+--------------------------------+
| Field            | Type         | Null | Key | Default              | Extra                          |
+------------------+--------------+------+-----+----------------------+--------------------------------+
| id               | char(36)     | NO   | PRI | NULL                 |                                |
| creationTime     | varchar(255) | NO   | MUL | NULL                 |                                |
| type             | char(60)     | NO   |     | NULL                 |                                |
| authorizedUserId | char(36)     | YES  |     | NULL                 |                                |
| status           | char(60)     | NO   |     | NULL                 |                                |
| message          | text         | YES  |     | NULL                 |                                |
| rawEvent         | text         | NO   |     | NULL                 |                                |
| cloneUrl         | char(255)    | YES  | MUL | NULL                 |                                |
| branch           | char(255)    | YES  |     | NULL                 |                                |
| commit           | varchar(255) | YES  |     | NULL                 |                                |
| projectId        | char(36)     | YES  |     | NULL                 |                                |
| prebuildStatus   | char(60)     | YES  |     | NULL                 |                                |
| prebuildId       | char(36)     | YES  |     | NULL                 |                                |
| deleted          | tinyint(4)   | NO   |     | 0                    |                                |
| _lastModified    | timestamp(6) | NO   | MUL | CURRENT_TIMESTAMP(6) | on update CURRENT_TIMESTAMP(6) |
+------------------+--------------+------+-----+----------------------+--------------------------------+
15 rows in set (0.02 sec)
```

## Related Issue(s)

Related #10764

## How to test
Push a commit to a repository configured as a project with this preview environment and verify you see webhook events stored in the DB. 

```
mysql> select type, status, prebuildStatus from d_b_webhook_event;
+------+-----------+--------------------+
| type | status    | prebuildStatus     |
+------+-----------+--------------------+
| push | processed | prebuild_triggered |
+------+-----------+--------------------+
1 row in set (0.00 sec)
```

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-db/src/typeorm/entity/db-webhook-event.ts
+++ b/components/gitpod-db/src/typeorm/entity/db-webhook-event.ts
@@ -1,0 +1,86 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the Gitpod Enterprise Source Code License,
+ * See License.enterprise.txt in the project root folder.
+ */
+
+import { Entity, Column, PrimaryColumn, Index } from "typeorm";
+
+import { Transformer } from "../../typeorm/transformer";
+import { WebhookEvent } from "@gitpod/gitpod-protocol";
+
+@Entity()
+// on DB but not Typeorm: @Index("ind_lastModified", ["_lastModified"])   // DBSync
+export class DBWebhookEvent implements WebhookEvent {
+    @PrimaryColumn("uuid")
+    id: string;
+
+    @Column()
+    @Index("ind_creationTime")
+    creationTime: string;
+
+    @Column()
+    type: string;
+
+    @Column({
+        default: "",
+        transformer: Transformer.MAP_EMPTY_STR_TO_UNDEFINED,
+    })
+    authorizedUserId?: string;
+
+    @Column()
+    rawEvent: string;
+
+    @Column()
+    @Index("ind_status")
+    status: WebhookEvent.Status;
+
+    @Column({
+        default: "",
+        transformer: Transformer.MAP_EMPTY_STR_TO_UNDEFINED,
+    })
+    message?: string;
+
+    @Column({
+        default: "",
+        transformer: Transformer.MAP_EMPTY_STR_TO_UNDEFINED,
+    })
+    @Index("ind_prebuildStatus")
+    prebuildStatus?: WebhookEvent.PrebuildStatus;
+
+    @Column({
+        default: "",
+        transformer: Transformer.MAP_EMPTY_STR_TO_UNDEFINED,
+    })
+    @Index("ind_prebuildId")
+    prebuildId?: string;
+
+    @Column({
+        default: "",
+        transformer: Transformer.MAP_EMPTY_STR_TO_UNDEFINED,
+    })
+    projectId?: string;
+
+    @Column({
+        default: "",
+        transformer: Transformer.MAP_EMPTY_STR_TO_UNDEFINED,
+    })
+    @Index("ind_cloneUrl")
+    cloneUrl?: string;
+
+    @Column({
+        default: "",
+        transformer: Transformer.MAP_EMPTY_STR_TO_UNDEFINED,
+    })
+    branch?: string;
+
+    @Column({
+        default: "",
+        transformer: Transformer.MAP_EMPTY_STR_TO_UNDEFINED,
+    })
+    commit?: string;
+
+    // This column triggers the db-sync deletion mechanism. It's not intended for public consumption.
+    @Column()
+    deleted: boolean;
+}
--- a/components/gitpod-db/src/typeorm/migration/1657702361007-AddWebhookEvent.ts
+++ b/components/gitpod-db/src/typeorm/migration/1657702361007-AddWebhookEvent.ts
@@ -1,0 +1,19 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class AddWebhookEvent1657702361007 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(
+            "CREATE TABLE IF NOT EXISTS `d_b_webhook_event` ( `id` char(36) NOT NULL, `creationTime` varchar(255) NOT NULL, `type` char(60) NOT NULL, `authorizedUserId` char(36) NULL, `status` char(60) NOT NULL, `message` text NULL, `rawEvent` text NOT NULL, `cloneUrl` char(255) NULL, `branch` char(255) NULL, `commit` varchar(255) NULL, `projectId` char(36) NULL, `prebuildStatus` char(60) NULL, `prebuildId` char(36) NULL, `deleted` tinyint(4) NOT NULL DEFAULT '0', `_lastModified` timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6), PRIMARY KEY(`id`), KEY `ind_dbsync` (`_lastModified`), KEY `ind_cloneUrl` (`cloneUrl`), KEY `ind_status` (`status`), KEY `ind_prebuildStatus` (`prebuildStatus`), KEY `ind_prebuildId` (`prebuildId`), KEY `ind_creationTime` (`creationTime`) ) ENGINE = InnoDB DEFAULT CHARSET = utf8mb4;",
+        );
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query("DROP TABLE `d_b_webhook_event`;");
+    }
+}
--- a/components/gitpod-db/src/typeorm/webhook-event-db-impl.ts
+++ b/components/gitpod-db/src/typeorm/webhook-event-db-impl.ts
@@ -1,0 +1,54 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the Gitpod Enterprise Source Code License,
+ * See License.enterprise.txt in the project root folder.
+ */
+
+import { injectable, inject } from "inversify";
+import { Repository } from "typeorm";
+import { v4 as uuidv4 } from "uuid";
+
+import { TypeORM } from "./typeorm";
+import { WebhookEventDB } from "../webhook-event-db";
+import { DBWebhookEvent } from "./entity/db-webhook-event";
+import { WebhookEvent } from "@gitpod/gitpod-protocol";
+
+@injectable()
+export class WebhookEventDBImpl implements WebhookEventDB {
+    @inject(TypeORM) protected readonly typeORM: TypeORM;
+
+    protected async getEntityManager() {
+        return (await this.typeORM.getConnection()).manager;
+    }
+
+    protected async getRepo(): Promise<Repository<DBWebhookEvent>> {
+        return (await this.getEntityManager()).getRepository(DBWebhookEvent);
+    }
+
+    async createEvent(parts: Omit<WebhookEvent, "id">): Promise<WebhookEvent> {
+        const repo = await this.getRepo();
+        const newEvent: WebhookEvent = {
+            ...parts,
+            id: uuidv4(),
+            creationTime: new Date().toISOString(),
+        };
+        return await repo.save(newEvent);
+    }
+
+    async updateEvent(id: string, update: Partial<WebhookEvent>): Promise<void> {
+        const repo = await this.getRepo();
+        const safeUpdate: Partial<WebhookEvent> = { ...update, id };
+        delete safeUpdate.type;
+        delete safeUpdate.rawEvent;
+        await repo.save(safeUpdate);
+    }
+
+    async findByCloneUrl(cloneUrl: string, limit?: number): Promise<WebhookEvent[]> {
+        const repo = await this.getRepo();
+        const query = repo.createQueryBuilder("event");
+        query.where("event.cloneUrl = :cloneUrl", { cloneUrl });
+        query.orderBy("creationTime", "DESC");
+        query.limit(limit);
+        return query.getMany();
+    }
+}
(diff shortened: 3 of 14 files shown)


components/gitpod-db/src/typeorm/migration/1657702361007-AddWebhookEvent.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class AddWebhookEvent1657702361007 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(
+            "CREATE TABLE IF NOT EXISTS `d_b_webhook_event`(`id` char(36) NOT NULL, `creationTime` varchar(255) NOT NULL, `type` char(255) NOT NULL, `status` char(255) NOT NULL, `message` text NULL, `rawEvent` text NOT NULL, `cloneUrl` char(255) NULL, `projectId` char(36) NULL, `prebuildId` char(36) NULL, `deleted` tinyint(4) NOT NULL DEFAULT '0', `_lastModified` timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6), PRIMARY KEY(`id`), KEY `ind_dbsync` (`_lastModified`), KEY `ind_cloneUrl` (`cloneUrl`), KEY `ind_creationTime` (`creationTime`), KEY `ind_projectId` (`projectId`)) ENGINE = InnoDB DEFAULT CHARSET = utf8mb4;",
```
geropl, line 12: type, creationTime and status are hot candidates for an index and seem very big (255 chars): For those it makes sense to make them a tad short if possible (in the past we did 60chars?), because that directly affects index size.

Also, `message` is in here, but not the `DBWebhookEvent`: I think we should have it to explain details to users, what do you think?

components/gitpod-db/src/typeorm/webhook-event-db-impl.ts
```diff
@@ -0,0 +1,60 @@
...
+            id: uuidv4(),
+            creationTime: new Date().toISOString(),
+        };
+        return await repo.save(newEvent);
+    }
+
+    async updateEvent(id: string, update: Partial<WebhookEvent>): Promise<void> {
+        const repo = await this.getRepo();
+        const safeUpdate: Partial<WebhookEvent> = { ...update, id };
+        delete safeUpdate.type;
```
geropl, line 41: :bulb: :+1:

components/gitpod-db/src/typeorm/webhook-event-db-impl.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    async find(filter: { cloneUrl?: string | undefined; projectId?: string | undefined }): Promise<WebhookEvent[]> {
+        const repo = await this.getRepo();
+        if (!filter.cloneUrl && !filter.projectId) {
+            return [];
+        }
+        const query = repo.createQueryBuilder("event");
+        if (filter.projectId) {
+            query.where("event.projectId = :projectId", { projectId: filter.projectId });
+        } else if (filter.cloneUrl) {
```
geropl, line 54: If this really is exclusive I think we should have two methods: `findByProjectId` and `findByCloneUrl`.

:cloud: For both queries we should have indexes: `ind_projectId_creationTime` and `ind_cloneUrl_creationTime`

components/gitpod-db/src/webhook-event-db.spec.db.ts
```diff
@@ -0,0 +1,57 @@
...
+
+import * as chai from "chai";
+import { suite, test, timeout } from "mocha-typescript";
+import { testContainer } from "./test-container";
+import { TypeORM } from "./typeorm/typeorm";
+import { WebhookEventDB } from "./webhook-event-db";
+import { DBWebhookEvent } from "./typeorm/entity/db-webhook-event";
+const expect = chai.expect;
+
+@suite.only
```
geropl, line 15: Leftover :broom:

components/gitpod-db/src/webhook-event-db.spec.db.ts
```diff
@@ -0,0 +1,57 @@
...
+        await this.clear();
+    }
+
+    protected async clear() {
+        const connection = await this.typeORM.getConnection();
+        const manager = connection.manager;
+        await manager.clear(DBWebhookEvent);
+    }
+
+    @test public async testSafeUpdate() {
```
geropl, line 35: :orange_heart:

components/gitpod-protocol/src/webhook-event.ts
```diff
@@ -0,0 +1,39 @@
...
+    rawEvent: string;
+
+    status:
+        | "received"
+        | "dismissed_unauthorized"
+        | "ignored_incomplete"
+        | "prebuild_trigger_failed"
+        | "ignored_unconfigured"
+        | "prebuild_triggered"
+        | string;
```
geropl, line 24: Why leaving status open here? :thinking:

components/gitpod-protocol/src/webhook-event.ts
```diff
@@ -0,0 +1,39 @@
...
+    message?: string;
+
+    /**
+     * if `status` is `prebuild_triggered` we can expect this to point to a triggered prebuild.
+     */
+    prebuildId?: string;
+
+    projectId?: string;
+
+    cloneUrl?: string;
```
geropl, line 38: Just wondering if this is not always available...? :thinking:

geropl: Ah, scrape that - understood! :+1:

components/server/ee/src/prebuilds/bitbucket-app.ts
```diff
@@ -112,11 +121,16 @@ export class BitbucketApp {
...
             span.setTag("contextURL", contextURL);
+            const context = (await this.contextParser.handle({ span }, user, contextURL)) as CommitContext;
+            await this.webhookEvents.updateEvent(event.id, {
+                projectId: projectAndOwner?.project?.id,
+                cloneUrl: context.repository.cloneUrl,
+            });
             const config = await this.prebuildManager.fetchConfig({ span }, user, context);
             if (!this.prebuildManager.shouldPrebuild(config)) {
                 console.log("Bitbucket push event: No config. No prebuild.");
+                await this.webhookEvents.updateEvent(event.id, { status: "ignored_unconfigured" });
```
geropl, line 133: Hm. As the table is about WebhookEvents now this prebuild-specific status does not make sense now: Imagine a webhook triggered both, a preview and a prebuild, what status would it be in?

What about keeping the `status` strictly with the WebHook:
```
"received"
  | "dismissed_unauthorized"
  | "ignored_incomplete"
  | "processed" // or similar
```

The link to the prebuild exists via `prebuildId`. Still, we need a place to document the decisions, e.g. whether or not a prebuild got triggered... :thinking:

components/gitpod-db/src/typeorm/webhook-event-db-impl.ts
svenefftinge: Agree and introduce `findByProjectId` only once we need it.

geropl:
> Could we also already extract the pushing user and match it with our user? We plan to build a metric into the product that shows team activation and will be based on the difference of commits authored in total (the webhook events) vs. which of those were done from Gitpod. So would be great to have a user shape in this table that optionally has a userID if that user exists in d_b_user.

I think that is the intend, but this PR serves as a first step towards that. :footprints:  :+1:

components/server/ee/src/prebuilds/bitbucket-app.ts
AlexTugarev: > prebuild-specific status does not make sense now: Imagine a webhook triggered both, a preview and a prebuild, what status would it be in?

Good question. I think instead of generalizing to a minimal set of commons, I'd rather split the status by concerns. In fact, we need explicit states for the frontend to render. If we coerce different concerns into generic states, I fear it will be difficult to render meaningful states. 

What about adding explicit `prebuildStatus`? We can tag specific statuses. 

Also, is there something to read more on the new preview feature? I find it hard to reason about, as I've no details so far.

components/gitpod-db/src/typeorm/migration/1657702361007-AddWebhookEvent.ts
AlexTugarev: > Also, message is in here, but not the DBWebhookEvent: I think we should have it to explain details to users, what do you think?

Going to add it. 

@geropl, isn't an index on `creationTime` needed to speed up "fetch last N entries"?

geropl: > @geropl, isn't an index on creationTime needed to speed up "fetch last N entries"?

Jep, true! :bulb:

components/gitpod-db/src/typeorm/entity/db-webhook-event.ts
```diff
@@ -0,0 +1,83 @@
...
+        default: "",
+        transformer: Transformer.MAP_EMPTY_STR_TO_UNDEFINED,
+    })
+    authorizedUserId?: string;
+
+    @Column()
+    rawEvent: string;
+
+    @Column()
+    status: WebhookEvent.Status;
```
geropl, line 35: :+1: for adding an index here, too.

components/gitpod-db/src/typeorm/entity/db-webhook-event.ts
```diff
@@ -0,0 +1,83 @@
...
+        default: "",
+        transformer: Transformer.MAP_EMPTY_STR_TO_UNDEFINED,
+    })
+    message?: string;
+
+    @Column({
+        default: "",
+        transformer: Transformer.MAP_EMPTY_STR_TO_UNDEFINED,
+    })
+    prebuildStatus?: WebhookEvent.PrebuildStatus;
```
geropl, line 47: :+1: Another candidate for proactively adding an index.

components/gitpod-db/src/typeorm/entity/db-webhook-event.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    @Column({
+        default: "",
+        transformer: Transformer.MAP_EMPTY_STR_TO_UNDEFINED,
+    })
+    prebuildId?: string;
```
geropl, line 53: Same here :+1:

components/gitpod-protocol/src/webhook-event.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+export interface WebhookEvent {
+    id: string;
+    creationTime: string;
+    type: "push" | string;
+
+    authorizedUserId?: string;
```
geropl, line 12: A comment would be golden. :sparkles: 
I still recall from our discussion earlier today, but tomorrow-me will have forgotten for sure :grimacing:

components/server/ee/src/prebuilds/bitbucket-app.ts
```diff
@@ -35,7 +36,13 @@ export class BitbucketApp {
                 if (req.header("X-Event-Key") === "repo:push") {
                     const span = TraceContext.startSpan("BitbucketApp.handleEvent", {});
                     const secretToken = req.query["token"] as string;
+                    const event = await this.webhookEvents.createEvent({
+                        type: "push",
+                        status: "received",
+                        rawEvent: JSON.stringify(req.body),
+                    });
                     if (!secretToken) {
+                        await this.webhookEvents.updateEvent(event.id, { status: "dismissed_unauthorized" });
```
geropl, line 45: I'd actually set `message: "No token provided"` here.

components/server/ee/src/prebuilds/bitbucket-app.ts
```diff
@@ -132,12 +152,23 @@ export class BitbucketApp {
...
+            await this.webhookEvents.updateEvent(event.id, {
+                prebuildStatus: "prebuild_triggered",
+                status: "processed",
+                prebuildId: ws.prebuildId,
+            });
             return ws;
+        } catch (e) {
+            console.error("Error processing Bitbucket webhook event", e);
+            await this.webhookEvents.updateEvent(event.id, {
+                prebuildStatus: "prebuild_trigger_failed",
```
geropl, line 168: Note: atm this also gathers a lot of other (potential) errors, but this is a good first intermittent step to untangle this code. :lotus_position:

components/gitpod-db/src/typeorm/entity/db-webhook-event.ts
AlexTugarev: done

components/server/ee/src/prebuilds/bitbucket-app.ts
AlexTugarev: going to improve error message in a separate PR.

BTW, this case is more of a "Broken Webhook registration", as it should provide a hint to what needs to be fixed.

components/server/ee/src/prebuilds/bitbucket-app.ts
geropl: > I believe the webhook table should not contain any prebuild information at all.

It's certainly a trade off, but I feel that it's worth it. The goal here is to untangle the question "what happened to this prebuild event and why" from the `PrebuiltWorkspace` state machine (which feels broken atm).

We could certainly add another table, e.g. WebhookEvent -> PrebuildEvent -> PrebuiltWorkspace, but as we'd join that on primary key of the WebHook event and to avoid the additional complexity here, I think it's ok (for now) to have that inside the WebhookEvent table. :lotus_position:

PR #11360, opened 2022-07-13
Purge old Workspaces at some point

### LINKED ISSUE #11378: Wipe deleted Workspace (+ their WorkspaceInstances) at some point
# Motivation
 - getting rid of old shapes/variance: our code is by now riddled with optional fields (like this: `someField?: string`) just because we once in 2018 had a period where we would not set those field. This is an attempt to get rid of those variance to at least have a chance to streamline the code
 - dead weight: This is about removing old _deleted_ workspaces only: It seems there is no sense in keeping them around
 - reduce table size:
   - reduce migration times: running a migration on one of the big tables takes ~50mins by now, which does not go well with CD
   - reduce storage: This has become a concern, although not a problem, yet
   - query time: This is a nice benefit, although not a driver as we don't have problems here for now

# Solutions

## Add 3rd "purged" state
Currently our Workspace deletion model has two phases:
 - **soft deleted**: A workspace goes into this state either when a) a user deletes it or b) it is garbage collected after a # of days of inactivity. The effect is that is "hidden" from our API, and can easily be reversed e.g. on user request, or inspected in case of abuse, for instance.
 - **content deleted**: A workspace goes into this mode after a certain time being in "soft deleted" state. All workspace contents (backups, logs, etc.) is deleted, and the workspace record is scrubbed from all PII, but the skeleton and main data model stays intact. This is helpful for being able to trace back issues or incidents even after the actual workspace content has been deleted.

This issue  proposes to introduce a 3rd state:
 - **purged**: A workspace moves into this state after a certain amount of time being in "content deleted" state. All workspace entries (workspace, instance, prebuild) are completely wiped from the DB.

# Remarks

Beyond the pure technical aspect, this affects all of downstream systems as well: We need to make sure that they can handle not getting the total view over all workspaces. :lotus_position:

geropl:
## Description

Introduces a 3rd workspace deletion state "purged" as discussed in #11378 .

## Related Issue(s)

Fixes #11378

### Run tests
 - `(cd components/gitpod-db && yarn test)`

### Execute in preview environment
 - open a workspace on [this branch](https://gitpod.io/#github.com/gitpod-io/gitpod/pull/11360)
 - `kubectl port-forward mysql-0 3306 &`
 - `mysql -u gitpod -h 127.0.0.1 -p`, PW: `PW`
 - Insert these:
 ```
INSERT INTO `gitpod`.`d_b_workspace`
(`id`,
`creationTime`,
`ownerId`,
`contextURL`,
`description`,
`context`,
`config`,
`softDeleted`,
`softDeletedTime`,
`contentDeletedTime`,
`cloneURL`)
VALUES
('123',
'2019-09-25T21:44:13.560Z',
'123',
'https://example.org/my/repo',
'my-repo',
'{}',
'{}',
'gc',
'2019-09-25T21:44:13.560Z',
'2019-10-25T21:44:13.560Z',
'https://example.org/my/repo');
 ```
 ```
INSERT INTO `gitpod`.`d_b_workspace_instance`
(`id`,
`workspaceId`,
`region`,
`creationTime`,
`configuration`,
`ideUrl`,
`workspaceImage`,
`status`)
VALUES
('1231',
'123',
'dev',
'2019-09-25T21:44:13.560Z',
'{}',
'https://example.org/my/workspace',
'',
'{"phase": "stopped"}');
 ```
 - `kubectl edit cm server-config` and set `workspaceGarbageCollection.intervalSeconds` to `30` (or similar)
 - `kubectl rollout restart deployment server`
 - wait at least 30 mins until you observe this line in the `server` logs: `wsgc: successfully purged 1 workspaces`

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-db/src/typeorm/workspace-db-impl.ts
+++ b/components/gitpod-db/src/typeorm/workspace-db-impl.ts
@@ -52,6 +52,7 @@ import { DBPrebuiltWorkspace } from "./entity/db-prebuilt-workspace";
 import { DBPrebuiltWorkspaceUpdatable } from "./entity/db-prebuilt-workspace-updatable";
 import { BUILTIN_WORKSPACE_PROBE_USER_ID } from "../user-db";
 import { DBPrebuildInfo } from "./entity/db-prebuild-info-entry";
+import { daysBefore } from "@gitpod/gitpod-protocol/lib/util/timeutil";
 
 type RawTo<T> = (instance: WorkspaceInstance, ws: Workspace) => T;
 interface OrderBy {
@@ -592,6 +593,22 @@ export abstract class AbstractTypeORMWorkspaceDBImpl implements WorkspaceDB {
         return dbResults as WorkspaceAndOwner[];
     }
 
+    public async findWorkspacesForPurging(
+        minContentDeletionTimeInDays: number,
+        limit: number,
+        now: Date,
+    ): Promise<WorkspaceAndOwner[]> {
+        const minPurgeTime = daysBefore(now.toISOString(), minContentDeletionTimeInDays);
+        const repo = await this.getWorkspaceRepo();
+        const qb = repo
+            .createQueryBuilder("ws")
+            .select(["ws.id", "ws.ownerId"])
+            .where(`ws.contentDeletedTime != ''`)
+            .andWhere(`ws.contentDeletedTime < :minPurgeTime`, { minPurgeTime })
+            .limit(limit);
+        return await qb.getMany();
+    }
+
     public async findWorkspacesForContentDeletion(
         minSoftDeletedTimeInDays: number,
         limit: number,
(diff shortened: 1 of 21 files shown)


svenefftinge:
looking into this, now 👀

svenefftinge (approved):
Looks really good! ❤️ 
Would be great to also delete corresponding records from
 - d_b_prebuilt_workspace 
 - d_b_workspace_instance_user
 - d_b_layout_data
 
Adding hold in case you want to address this in this PR.

/hold

geropl:
> Would be great to also delete corresponding records from

Yes indeed... make sense to do it now I guess. Have to double-check with @jakobhero about which ones are safe, though.

geropl:
`d_b_prebuild_info` is another candidate as well.

geropl:
@easyCZ Yes, looks good to me as well (cannot approve as it's my PR :see_no_evil: )! :+1: 

Just to summarize:
 - we decided to go with WorkspaceInstance and Workspace first
 - d_b_prebuilt_workspace, d_b_workspace_instance_user, d_b_prebuild_info and d_b_layout_data are good candidates to add as well, but we exclude them for now to move forward. We can add them later, and handle the diff manually (e.g., identify entries without workspace, and schedule removal of those)

easyCZ:
/unhold

💥

Issue #10981, opened 2022-06-28
Make ws-daemon workspace class aware


Furisto:
## Is your feature request related to a problem? Please describe
Support https://github.com/gitpod-io/gitpod/issues/8261

## Describe the behaviour you'd like
ws-daemon needs to be aware of the workspace class of the workspace that it is limiting, so that it can apply the proper resource restrictions for that class.

## Describe alternatives you've considered
n.a.

## Additional context
Affected components:
- ws-daemon

PR #11374, opened 2022-07-14
Limit resources based on workspace class

Furisto:
## Description
This PR adds support for cpu limiting based on the workspace class of the workspace. The configuration for each workspace class contains a cpu limiting section which describes the minimal resources the workspaces gets and the burst limit. These values will be set as annotations on the workspace pod. ws-daemon will then use these annotations to distribute cpu resources

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/10981

## How to test
- Open workspace
- Check that the workspace pod contains two annotations: `gitpod.io/cpuMinLimit` and `gitpod.io/cpuBurstLimit`
- SSH into the harvester vm and find the cgroup for the workspace pod
- Check that the min limit has been applied
- Install stress-ng in the workspace
- Run `stress-ng --cpu 4`
- After a few seconds you should see that the burst limit has been applied to the cgroup of the workspace
- Use kubectl annotate to change the annotations
- The cgroup of the workspace should have been updated with the new values
- Deactivate feature flag in configcat
- Start workspace
- Check that no limit annotations are present
- Find the cgroup of the workspace
- Check that the limits configured in the ws-daemon configmap are applied

## Release Notes

```release-note
None
```

## Note
This relies on on a change in the werft files, that is why the main build failed. The custom build with the updated werft files (below) is successful however.

## Werft options:

- [x] /werft with-preview


roboquat:
@Furisto: Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://git.k8s.io/community/contributors/guide/release-notes.md) to remove it.

<details>

</details>

kylos101:
@Furisto what are your thoughts about limiting disk IO bandwidth? I ask because on XL (large) nodes, we "could" grant the workspaces a higher bandwidth because there are less workspaces and more available bandwidth. 

For now, I think we should keep this as a :skateboard:, and continue to impose the same disk IO bandwidth limit like we do now (regardless of the node type). But, I am interested in your thoughts. :thinking:  :smile: 

Further, even when we go to PVC, we could initially impose the same disk IO bandwidth limit. And later, experiment with removing the disk IO bandwidth limit for `/workspace`, and assuming it doesn't introduce any performance degradation - ship it.

kylos101:
:wave: @Furisto I was talking with @csweichel about adding a dependency to configcat on the workspace side like you shared, in `common-go`. He shared an alternative, which may be simpler to implement and leverage existing patterns.

Prereqs:
1. Add a new ConfigCat entry, call it something like `workspace_class_limiting`, this would be used by `server`
2. Add a new feature flag (leveraging our pre-configcat feature flag implementation), also call it `workspace_class_limiting`, this would be used by `ws-manager`

The flow might go like this:
1. Use ConfigCat on the `server` side to build a [StartWorkspaceRequest](https://github.com/gitpod-io/gitpod/blob/735a30899fdef2739e55d487f69a62ee037bc819/components/ws-manager-api/core.proto#L81). Depending on the ConfigCat rule for `workspace_class_limiting`, we would conditionally update the [spec](https://github.com/gitpod-io/gitpod/blob/735a30899fdef2739e55d487f69a62ee037bc819/components/ws-manager-api/core.proto#L94) to include the `workspace_class_limiting` [feature_flag](https://github.com/gitpod-io/gitpod/blob/735a30899fdef2739e55d487f69a62ee037bc819/components/ws-manager-api/core.proto#L527). 
2. If `workspace_class_limiting` is present as a feature_flag in the `StartWorkspaceRequest`, `ws-manager` would add related annotations to the pod (given the limits defined on the related workspace class).
3. `ws-daemon` CPU limiting would have a convention to check pods for the above annotations to get base and burst values, and if the annotations do not exist, then the values persisted in our config map would be used for CPU limiting.

CC: @easyCZ as a heads up

csweichel:
> 👋 @Furisto I was talking with @csweichel about adding a dependency to configcat on the workspace side like you shared, in `common-go`. He shared an alternative, which may be simpler to implement and leverage existing patterns.
> 
> Prereqs:
> 
> 1. Add a new ConfigCat entry, call it something like `workspace_class_limiting`, this would be used by `server`
> 2. Add a new feature flag (leveraging our pre-configcat feature flag implementation), also call it `workspace_class_limiting`, this would be used by `ws-manager`
> 
> The flow might go like this:
> 
> 1. Use ConfigCat on the `server` side to build a [StartWorkspaceRequest](https://github.com/gitpod-io/gitpod/blob/735a30899fdef2739e55d487f69a62ee037bc819/components/ws-manager-api/core.proto#L81). Depending on the ConfigCat rule for `workspace_class_limiting`, we would conditionally update the [spec](https://github.com/gitpod-io/gitpod/blob/735a30899fdef2739e55d487f69a62ee037bc819/components/ws-manager-api/core.proto#L94) to include the `workspace_class_limiting` [feature_flag](https://github.com/gitpod-io/gitpod/blob/735a30899fdef2739e55d487f69a62ee037bc819/components/ws-manager-api/core.proto#L527).
> 2. If `workspace_class_limiting` is present as a feature_flag in the `StartWorkspaceRequest`, `ws-manager` would add related annotations to the pod (given the limits defined on the related workspace class).
> 3. `ws-daemon` CPU limiting would have a convention to check pods for the above annotations to get base and burst values, and if the annotations do not exist, then the values persisted in our config map would be used for CPU limiting.
> 
> CC: @easyCZ as a heads up

benefits of this approach:
- Today we're super careful on the workspace side with the variance introduced by feature flags. That variance is grouped within a few hundred lines in ws-manager. Not (yet) introducing configcat widely would let us keep that.
- Because workspace feature flags are always part of the workspace config/`StartWorkspaceRequest` we have an easier time debugging misbehaving workspaces. It's always clear what kind of configuration a workspace ran on, because it's sent to the workspace cluster in its entirety.
- connecting configcat with workspace feature flags in server gives us a clear path to roll out/get rid of the `protected_secrets` feature flag. We have a precedent for this motion with PVC.

easyCZ:
In general, I'd agree that the boundary between control plane (webapp) and data plane (workspace clusters) is very often best served as explicit feature flags in the API request crossing the boundary. It allows the two planes to be decoupled from each other in terms of feature flagging behaviour and does indeed provide slightly saner semantics in terms of the lifecycle of a data plane resource.

I have no issues with continuing to use the explicit feature flags in the request for Control Plane. One can always choose to evaluate the feature flags just before issuing the API request which crosses the plane boundary to get very similar behaviour, but without the downsides of having to manage lifecycle with potentially dynamically changing flags.

kylos101:
@utam0k could you help @Furisto with a review and testing? :pray:

components/ws-daemon/pkg/cpulimit/cpulimit.go
```diff
@@ -172,15 +177,23 @@ func (d *Distributor) Tick(dt time.Duration) (DistributorDebug, error) {
 	var burstBandwidth Bandwidth
 	for _, id := range wsOrder {
 		ws := d.History[id]
-		limit := d.Limiter.Limit(ws.Usage())
+		limit, err := d.Limiter.Limit(ws)
+		if err != nil {
+			log.Errorf("unable to apply min limit: %v", err)
```
utam0k, line 182:
```suggestion
			log.WithError(err).Errorf("unable to apply min limit")
```

components/ws-daemon/pkg/cpulimit/cpulimit.go
```diff
@@ -172,15 +177,23 @@ func (d *Distributor) Tick(dt time.Duration) (DistributorDebug, error) {
...
 		// if we didn't get the max bandwidth, but were throttled last time
 		// and there's still some bandwidth left to give, let's act as if had
 		// never spent any CPU time and assume the workspace will spend their
 		// entire bandwidth at once.
 		var burst bool
 		if totalBandwidth < d.TotalBandwidth && ws.Throttled() {
-			limit = d.BurstLimiter.Limit(ws.Usage())
+			limit, err = d.BurstLimiter.Limit(ws)
+			if err != nil {
+				log.Errorf("unable to apply burst limit: %v", err)
```
utam0k, line 194:
```suggestion
				log.WithError(err).Errorf("unable to apply burst limit")
```

components/ws-daemon/pkg/cpulimit/cpulimit.go
```diff
@@ -206,7 +219,7 @@ func (d *Distributor) Reset() {
 
 // ResourceLimiter implements a strategy to limit the resurce use of a workspace
 type ResourceLimiter interface {
-	Limit(budgetSpent CPUTime) (newLimit Bandwidth)
+	Limit(wsh *WorkspaceHistory) (Bandwidth, error)
```
utam0k, line 222: I feel we must return some value without error because if the error is returned, we can't set any value, in other words, a workspace is unlimited.
```suggestion
	Limit(wsh *WorkspaceHistory) Bandwidth
```

components/ws-manager/pkg/manager/create.go
```diff
@@ -379,6 +379,18 @@ func (m *Manager) createDefiniteWorkspacePod(startContext *startWorkspaceContext
 		}
 		annotations[customTimeoutAnnotation] = req.Spec.Timeout
 	}
+
+	limits := startContext.Class.Container.Limits
+	if limits != nil && limits.CPU != nil {
+		if limits.CPU.MinLimit != "" {
+			annotations[kubernetes.WorkspaceCpuLimitAnnotation] = limits.CPU.MinLimit
```
utam0k, line 386: I got confused about the name of annotation(`WorkspaceCpuLimitAnnotation`) and config(`WorkspaceCpuLimitAnnotation`). How about unifying with or without Min? I would recommend adding a Min.

components/ws-daemon/pkg/cpulimit/limiter_test.go
```diff
@@ -37,10 +37,15 @@ func TestBucketLimiter(t *testing.T) {
...
-			}
+			//limiter := cpulimit.BucketLimiter(test.Buckets)
+			// ws := &cpulimit.WorkspaceHistory {
+
+			// }
+
+			// limit,_ := limiter.Limit(test.BudgetSpent)
+			// if limit != test.ExpectedLimit {
+			// 	t.Errorf("unexpected limit %d: expected %d", limit, test.ExpectedLimit)
+			// }
```
utam0k, line 48: You forgot to remove these codes ... 🤔

components/ws-daemon/pkg/cpulimit/cpulimit.go
Furisto: This is now taken care of by using the composite limiter which has a fallback that always returns a value (if cpu limiting is enabled and configured, but that is the same behavior as before)

Furisto:
/werft run github -j .werft/build.yaml  -a updateGitHubStatus=gitpod-io/gitpod -a with-preview

sagor999:
@Furisto I keep holding off on reviewing your PR because I keep seeing that build keeps failing. Is it ready for review?

Furisto:
@sagor999 It is. This relied on a change to the werft files and there is no way to force the github checks to use the custom build instead of the main build. I have moved this change out in its [own PR](https://github.com/gitpod-io/gitpod/pull/11560) and rebased this on top of the PR so that the build is green now.

components/common-go/kubernetes/kubernetes.go
```diff
@@ -64,6 +64,12 @@ const (
 
 	// WorkspaceSSHPublicKeys contains all authorized ssh public keys that can be connected to the workspace
 	WorkspaceSSHPublicKeys = "gitpod.io/sshPublicKeys"
+
+	// workspaceCpuLimit denotes the cpu limit of a workspace
+	WorkspaceCpuMinLimitAnnotation = "gitpod.io/cpuMinLimit"
```
sagor999, line 69: why annotation is called `cpuMinLimit`? It implies this is a minimum limit set on workspace? :thinking: 
Maybe you could clarify that in the comment above?

sagor999 (approved):
LGTM :rocket: 
One minor nit.

Do you have a PR for ops repo with installer changes?

/hold
(just in case to prevent accidental auto merge by bot)

Furisto:
> LGTM rocket One minor nit.
> 
> Do you have a PR for ops repo with installer changes?
> 
> /hold (just in case to prevent accidental auto merge by bot)

[Yes](https://github.com/gitpod-io/ops/pull/3431/files)

easyCZ (approved):

utam0k (approved):
Awesome

kylos101:
@aledbf may we ask for your review on this PR? This way @Furisto can react to feedback from you. This file filter view may help (Only files owned by you):
image

@Furisto I think it would be good to leave the `hold`, at least until we have `gen56` stable. @aledbf wdyt?

kylos101:
@Furisto as a heads up, there's a conflict with `components/server/src/workspace/workspace-starter.ts`

nandajavarma (approved):

kylos101:
@aledbf can you help with this review? This is needed for `gen60` and Usage based pricing. cc: @geropl

components/ws-daemon/pkg/cpulimit/cpulimit.go
```diff
@@ -218,8 +238,32 @@ type fixedLimiter struct {
...
+		Annotation: annotation,
+	}
+}
+
+type annotationLimiter struct {
+	Annotation string
+}
+
+func (a annotationLimiter) Limit(wsh *WorkspaceHistory) (Bandwidth, error) {
+	value, ok := wsh.LastUpdate.Pod.Annotations[a.Annotation]
```
aledbf, line 256: Why are we storing the Pod and not only the annotations?

components/ws-manager-api/go/config/config.go
```diff
@@ -453,10 +494,62 @@ func RenderWorkspacePortURL(urltpl string, ctx PortURLContext) (string, error) {
...
+// ResourceRequestConfiguration configures resources of a pod/container
+type ResourceRequestConfiguration struct {
 	CPU              string `json:"cpu"`
 	Memory           string `json:"memory"`
 	EphemeralStorage string `json:"ephemeral-storage"`
 	Storage          string `json:"storage,omitempty"`
 }
+
+type ResourceLimitConfiguration struct {
+	CPU              *CpuResourceLimit `json:"cpu"`
```
aledbf, line 506: Any reason why we don't set omitempty to all fields?

Furisto: I just kept it consistent with the previous behavior. It is also nice to not have type out the fields when editing the configmap in e.g. preview environments.

components/ws-daemon/pkg/cpulimit/cpulimit.go
Furisto: It was just convenient to do and having the pod object around can be useful when we want to do limiting based on other attributes besides annotations.

aledbf (approved):

akosyakov (approved):

PR #11377, opened 2022-07-14
Fix the check cert function

liam-j-bennett:
## Description
Fix the cert check to be true for Ready

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


ArthurSens (approved):
Good catch!

ArthurSens:
/hold

Adding the hold label so we don't block the merge queue while the build is still running :)

PR #11387, opened 2022-07-14
[usage] Implement `GetBilledUsage` RPC

andrew-farries:
## Description

Implement the usage API `GetBilledUsage` RPC. Given an attribution ID, return usage records corresponding to that ID.

⚠️ No date filtering yet - this just returns *all* usage records for the attribution ID regardless of age. ⚠️ 

## Related Issue(s)
Part of https://github.com/gitpod-io/gitpod/issues/9036

## How to test

Unit tests for the RPC. 

To test the API manually:

1. Run the usage service locally against the database in the preview environment for this PR:

```bash
cd components/usage
DB_HOST=localhost DB_USERNAME=gitpod DB_PASSWORD=<> DB_PORT=3306 go run . run
```

2. Use a gRPC client like `evans` to run against the usage API:

image

## Release Notes

```release-note
NONE
```

## Werft options:
- [x] /werft with-preview


components/usage/pkg/apiv1/usage.go
```diff
@@ -8,20 +8,53 @@ import (
...
+			UserId:         "",
+			TeamId:         "",
+			WorkspaceId:    "",
+			WorkspaceType:  "",
+			ProjectId:      "",
+			InstanceId:     usageRecord.InstanceID.String(),
+			WorkspaceClass: "",
+			StartTime:      timestamppb.New(usageRecord.StartedAt),
+			EndTime:        endTime,
+			Credits:        int64(usageRecord.CreditsUsed),
```
andrew-farries, line 48: Regarding the missing fields here, I'm not sure how many we want to store in the usage table ourselves, vs how much we want to obtain by joining with the existing workspace table.

The motivation for storing usage was so that we don't need to recompute it on every query to the usage API, but as long as we have the instanceIds, we should be able to recover the workspace details here with joins.

geropl: In general I agree that this is the strategy we should follow. But WorkspaceClass and WorkspaceType should definitely in here, because they are relevant for the billing parts.
And regarding the hierachy-tuple TeamId-ProjectId-UserId-WorkspaceId: I think being able to directly correlate this entry (without the indirection over WorkspaceInstance) is valuable as well, and helps to "place this session on the map" if you will. I think we should have those, too.

geropl:
@andrew-farries I tested with evans but got an SEGFAULT. I have not clue how to debug that.

Also, wrt to the fields discussion: If we ship less fields then currently defined we should adjust the [parsing code here](https://github.com/gitpod-io/gitpod/blob/d1c412edc2536e3c451afc4ded6b3858f68f6770/components/server/ee/src/workspace/gitpod-server-impl.ts#L2125-L2135) as well.

andrew-farries:
> @andrew-farries I tested with evans but got an SEGFAULT. I have not clue how to debug that.

This appears to be a regression between `evans` `v0.10.6` and `v0.10.7` released yesterday.

Using `v0.10.6` works as expected.

andrew-farries:
PR to add a working version to the dev image here: https://github.com/gitpod-io/gitpod/pull/11413

components/usage/pkg/apiv1/usage.go
geropl: FYI We talked in sync, and agreed to wait for @easyCZ and his opinion. :+1:

easyCZ: IMO, we should have first-class fields (these are mostly the ones we want to group-by in our views or are important enough to show in list views.

All the fields above meet this first-class definition so I would keem them as is. We should populate them directly from the usage report.

For any extra fields, we should call into WorkspaceInstance for details, or setup a service which pre-joins these for us.

components/usage/pkg/db/workspace_instance_usage.go
```diff
@@ -37,3 +38,14 @@ func CreateUsageRecords(ctx context.Context, conn *gorm.DB, records []WorkspaceI
 
 	return db.CreateInBatches(records, 1000).Error
 }
+
+func GetUsage(ctx context.Context, conn *gorm.DB, attributionId AttributionID) ([]WorkspaceInstanceUsage, error) {
```
easyCZ, line 42:
```suggestion
func ListUsage(ctx context.Context, conn *gorm.DB, attributionId AttributionID) ([]WorkspaceInstanceUsage, error) {
```

It returns a collection of records. Get is often used to mean a single record.

components/usage/pkg/apiv1/usage.go
```diff
@@ -8,20 +8,53 @@ import (
...
 )
 
 var _ v1.UsageServiceServer = (*UsageService)(nil)
 
 type UsageService struct {
+	conn *gorm.DB
 	v1.UnimplementedUsageServiceServer
 }
 
 func (us *UsageService) GetBilledUsage(ctx context.Context, in *v1.GetBilledUsageRequest) (*v1.GetBilledUsageResponse, error) {
```
easyCZ, line 25: I didn't spot this before but I'd call the RPC `ListUsage` because it returns a collection of records

andrew-farries: `ListUsage` or `LIstBilledUsage`?

easyCZ: I guess `ListBilledUsage` to keep it similar. But generally, a Service should only return one type of Resource. So here, we'd normally have a BilledUsageService, which returns `BilledUsage` records but no point changing now.

easyCZ (approved):

PR #11177, opened 2022-07-06
Update k8s dependencies to v0.24.3

aledbf:
## Description

- Update k8s dependencies to v0.24.3
- Remove empty `status` and `resource` definitions from manifests

## Related Issue(s)

fixes https://github.com/gitpod-io/gitpod/issues/8529
fixes https://github.com/gitpod-io/gitpod/issues/11136

## How to test
- Check workspaces work as usual
- Check the generated manifests do not contain empty `status` or `resource` definitions

```
docker run --entrypoint sh --rm \
  eu.gcr.io/gitpod-core-dev/build/installer:aledbf-k8s24.11 -c "cat /app/installer" >gitpod-installer

chmod +x gitpod-installer

cat <<EOF >values.yaml
apiVersion: v1
domain: fake.gitpod.io
EOF

./gitpod-installer render --use-experimental-config --config values.yaml > manifests.yaml
```

## Release Notes
```release-note
NONE
```

## Werft options:
- [X] /werft with-preview

PR #11391, opened 2022-07-14
[installer]: add in ability to post-process

MrSimonEmms:
## Description

Adds the ability to post-process the Installer's YAML natively in the Golang binary. This extends the [YQ library](https://github.com/mikefarah/yq), so any valid [YQ expression](https://mikefarah.gitbook.io/yq/operators) can be used to amend the generated YAML.

It also removes the `status` object from both Network Policies and Stateful Sets, both of which are problematic when supporting multiple versions of Kubernetes.

This also removes the problems associated with upgrades to the Kubernetes libraries (eg #11003) and allows for #11177 to be merged.

## How to test

Run the Installer, there should be no `status` in any `NetworkPolicy` or the OpenVSXProxy `StatefulSet` resources.

## Release Notes

```release-note
[installer]: add in ability to post-process the generated YAML
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/installer/pkg/postprocess/postprocess.go
+++ b/install/installer/pkg/postprocess/postprocess.go
@@ -10,9 +10,11 @@ import (
 	"strings"
 
 	"github.com/gitpod-io/gitpod/installer/pkg/common"
+	openvsxproxy "github.com/gitpod-io/gitpod/installer/pkg/components/openvsx-proxy"
 	"github.com/mikefarah/yq/v4/pkg/yqlib"
 	logging "gopkg.in/op/go-logging.v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
 )
 
 // Processors list of processes executed on each resource document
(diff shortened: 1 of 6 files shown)


install/installer/pkg/postprocess/postprocess.go
```diff
@@ -0,0 +1,119 @@
...
+	printerWriter := yqlib.NewSinglePrinterWriter(&writer)
+	encoder := yqlib.NewYamlEncoder(2, false, false, true)
+
+	printer := yqlib.NewPrinter(encoder, printerWriter)
+
+	decoder := yqlib.NewYamlDecoder()
+
+	streamEvaluator := yqlib.NewStreamEvaluator()
+
+	dir, err := os.MkdirTemp("", "post-processor")
```
Pothulapati, line 66: Any specific reason, on writing into a local FS and using that, instead of directly creating a reader and using `Evaluate`? 🤔 https://github.com/mikefarah/yq/blob/v4.26.1/pkg/yqlib/stream_evaluator.go#L92

MrSimonEmms: I originally tried that but I couldn't get it to work. We already use the "[create temp file, amend, move on](https://github.com/gitpod-io/gitpod/blob/main/install/installer/pkg/helm/helm.go#L102,L114)" approach in Helm so took view that it was ok to use it again

MrSimonEmms: /hold actually, just had a brainwave...

MrSimonEmms: @Pothulapati Thanks for questioning that. Believe it or not, your question actually made me realise the problem and the solution. I've now changed to using the `Evaluate` method 👍🏻 

/unhold

Pothulapati (approved):
LGTM :shipit:

PR #11392, opened 2022-07-14
[ws-manager] Remove temporal workaround for kubelet issue

aledbf:
## Description

Due to Kubernetes issue https://github.com/kubernetes/kubernetes/pull/108366 we had to add a workaround to bypass the issue. Now that upstream versions are available (since April 13th) we can remove it.
Affected Kubernetes version with the issue released a fix in:
- [v1.22.9](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.22.md#v1229)
- [v1.23.6](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.23.md#v1236)

We need to remove the workaround due to additional conditions that can affect the scheduling, like scale-up events.
Creating and deleting pods due to timeouts only add load to the API server.

Edit: example of side effect https://cloudlogging.app.goo.gl/SG2Qx1e21EykbRzNA

## Release Notes

```release-note
NONE
```

## Werft options:

- [X] /werft with-preview

### DIFF
--- a/components/ws-manager/pkg/manager/manager.go
+++ b/components/ws-manager/pkg/manager/manager.go
@@ -290,41 +290,24 @@ func (m *Manager) StartWorkspace(ctx context.Context, req *api.StartWorkspaceReq
 		}
 	}
 
-	// create the Pod in the cluster and wait until is scheduled
-	// https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.22.md#workloads-that-saturate-nodes-with-pods-may-see-pods-that-fail-due-to-node-admission
-	backoff := wait.Backoff{
-		Steps:    10,
-		Duration: 100 * time.Millisecond,
-		Factor:   2.5,
-		Jitter:   0.1,
-		Cap:      5 * time.Minute,
-	}
+	err = m.Clientset.Create(ctx, pod)
+	if err != nil {
+		m, _ := json.Marshal(pod)
+		safePod, _ := log.RedactJSON(m)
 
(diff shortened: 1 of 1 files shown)


sagor999 (approved):

jenting:
Would removing this workaround affect self-hosted if they are not using the Kubernetes version >= v1.22.9 _or_ v1.23.6?

aledbf:
> Would removing this workaround affect self-hosted if they are not using the Kubernetes version >= v1.22.9 _or_ v1.23.6?

Yes? That said, if you are running the affected versions, you will have the same issues as why we added the workaround.
This is not related to Gitpod, creating pods at the same volume that SaaS reproduces the error.

sagor999:
@gitpod-io/engineering-self-hosted FYI ^

lucasvaltl:
I am trying to wrap my head around the potential impact on self-hosted customers here. What is the likelihood that self-hosted customers will run into this? Is there a way to enable the workaround if customers are on the specific versions of K8s that are affected? 

From what I gathered: If you are spinning up nodes at a similar pace to what we do on SaaS AND you are on a K8s version <= v1.22.9 or v1.23.6  you will likely be affected by the bug now that we removed the workaround. I can't tell though how many customers meet both of these conditions. I do know that the default GKE version is `1.22.8-gke.202`and the default EKS version [_seems_ to be `1.22.10`](https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html)

PR #11405, opened 2022-07-15
Simple UI for Prebuild Events (sneak preview)

AlexTugarev:
## Description
Screen Shot 2022-07-15 at 11 21 42

## Related Issue(s)

Related https://github.com/gitpod-io/gitpod/issues/10764

## How to test

1. Verify you do not see "Prebuild Events" in the menu, as this is a 🛹  PR without any discussions about proper UI design.
2. Push a commit to a repository configured as a project with this preview environment and verify you see webhook events rendered on: https://at-prebuild-events.preview.gitpod-dev.com/projects/YOUR-PROJECT/events

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/dashboard/src/App.tsx
+++ b/components/dashboard/src/App.tsx
@@ -74,6 +74,7 @@ const NewProject = React.lazy(() => import(/* webpackPrefetch: true */ "./projec
 const ConfigureProject = React.lazy(() => import(/* webpackPrefetch: true */ "./projects/ConfigureProject"));
 const Projects = React.lazy(() => import(/* webpackPrefetch: true */ "./projects/Projects"));
 const Project = React.lazy(() => import(/* webpackPrefetch: true */ "./projects/Project"));
+const Events = React.lazy(() => import(/* webpackPrefetch: true */ "./projects/Events"));
 const ProjectSettings = React.lazy(() => import(/* webpackPrefetch: true */ "./projects/ProjectSettings"));
 const ProjectVariables = React.lazy(() => import(/* webpackPrefetch: true */ "./projects/ProjectVariables"));
 const Prebuilds = React.lazy(() => import(/* webpackPrefetch: true */ "./projects/Prebuilds"));
@@ -397,6 +398,9 @@ function App() {
                             path={projectsPathMainWithParams}
                             render={(props) => {
                                 const { resourceOrPrebuild } = props.match.params;
+                                if (resourceOrPrebuild === "events") {
+                                    return <Events />;
+                                }
                                 if (resourceOrPrebuild === "prebuilds") {
                                     return <Prebuilds />;
                                 }
@@ -445,6 +449,9 @@ function App() {
                                     }
                                     if (maybeProject === "usage") {
                                         return <TeamUsage />;
+                                    }
+                                    if (resourceOrPrebuild === "events") {
+                                        return <Events />;
                                     }
                                     if (resourceOrPrebuild === "prebuilds") {
                                         return <Prebuilds />;
--- a/components/gitpod-protocol/src/teams-projects-protocol.ts
+++ b/components/gitpod-protocol/src/teams-projects-protocol.ts
@@ -7,6 +7,7 @@
 import { PrebuiltWorkspaceState } from "./protocol";
 import { v4 as uuidv4 } from "uuid";
 import { DeepPartial } from "./util/deep-partial";
+import { WebhookEvent } from "./webhook-event";
 
 export interface ProjectConfig {
     ".gitpod.yml": string;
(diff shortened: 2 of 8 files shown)


components/gitpod-protocol/src/teams-projects-protocol.ts
```diff
@@ -152,3 +153,15 @@ export interface TeamMembershipInvite {
     /** This is a flag that triggers the HARD DELETION of this entity */
     deleted?: boolean;
 }
+
+export interface PrebuildEvent {
+    id: string;
+    creationTime: string;
+    status: WebhookEvent.Status | WebhookEvent.PrebuildStatus;
```
geropl, line 160: Why not use the same interface we use for the DB? I feel that would avoid a lot of confusion, as that is our usual pattern (for the JSON RPC API). Plus it would avoid the duplication here. :lotus_position:

components/dashboard/src/App.tsx
```diff
@@ -397,6 +398,9 @@ function App() {
                             path={projectsPathMainWithParams}
                             render={(props) => {
                                 const { resourceOrPrebuild } = props.match.params;
+                                if (resourceOrPrebuild === "events") {
```
geropl, line 401: Sths strange here:
 1. With this code I would have expected to find this view at `/t/myteam/myproject/events`, but when I do so it drops into "PrebuildDetails" (fallback below).
 2. When looking at the code below, it seems we have a global events view at `/events`. I wonder what this is for - the user's events? :thinking: But if I visit that it says it cannot find it and redirects (to gitpod.io).
 3. In the description it says `/projects/myproject/events`: This is where I see an empty list with no data. It seems it's missing the "team" part, because the UI renders weird (see below), and we seem to [require it in code](https://github.com/gitpod-io/gitpod/pull/11405/files#diff-c41e3988d6100a5fc475d119437bc25b92fd009b1c68208148f9f494fbb795e3R61). Plus I didn't even knew we had that prefix - and wonder if we should at that view there :-D 
image

components/gitpod-protocol/src/teams-projects-protocol.ts
AlexTugarev: I want to avoid pushing event payload over websocket too often. In some projects this may be a significant amount of data. Also, if's feels unnecessary to push user id's. 
In general, the pattern you are referring to caused some pain in the past where we lost control over what got serialized and transferred over the websocket.

components/dashboard/src/App.tsx
AlexTugarev: there is indeed a problem with `/t/team/project/events` location: 
https://at-prebuild-events.preview.gitpod-dev.com/t/team1/plain/events ✖️ 
https://at-prebuild-events.preview.gitpod-dev.com/projects/foobar/events ✔️ 

on 3.) that looks good. no team is required, the code reads ...

```
if (!teams || !projectSlug) {
            return;
        }
```
there is a project slug.
please be aware that the main menu is untouched for the 🛹

components/gitpod-protocol/src/teams-projects-protocol.ts
geropl: Ah, you're right, some fields we might definitely not want in here. :bulb: (e.g., authorizedUser)
Could we still make this a subshape with `Omit`/`Pick` then? really don't see the point in merging both status fields here. :upside_down_face: And it nicely expresses the relationship.

components/dashboard/src/App.tsx
AlexTugarev: fixed in https://github.com/gitpod-io/gitpod/compare/9031b74794fb79a4cf834f6f5cdb9ce4e3e5c8a6..262f9587439ee3350c5573c2889d1be54164bb63

components/gitpod-protocol/src/teams-projects-protocol.ts
AlexTugarev: Merging status here, as from prebuild UI perspective there is little need to deal with two different statuses at all. Actually we're mostly interested in prebuild status and for error cases in the generic ones. What's the point having it separated for the frontend?

geropl: That we can choose where to show it. E.g. I'm pretty sure we want a column status, and maybe a small area underneath for details if it's an error state (401, for instance). But I'm absolutely not sure we want to show an message related to prebuilds in that same area, maybe even not on the same page. :cloud:

geropl (approved):
image

:heart_eyes: This already solves 90% of our "SCM integration pain" on install. And offers so much room for future integration/enhancements.

I'd say we can already merge this as is, despite [our discussion here](https://github.com/gitpod-io/gitpod/pull/11405#discussion_r922110783), and improve in smaller :footprints: later! :ladder: 

/hold

geropl:
@AlexTugarev Can you ask @gtsiolis to have a look at this as well? It's not as urgent as UBP Paywall topics, but comes next on the list I feel. And with this being hidden in prod he can play around with it a bit with real data. :upside_down_face:

jldec:
I like that we're keeping the URL path generic `/events` instead of `preview-events`.
I see that you have already fixed the upper-case in the description.

OK merging this since it is on a hidden "sneak peek" endpoint at `/projects/YOUR-PROJECT/events` or `/t/YOUR-TEAM/YOUR-PROJECT/events`.

I would also be OK making this change GA by adding a new `Events` project tab between "Prebuilds" and "Settings". 

/unhold

jldec:
@gtsiolis - would you take a look at this 🙏  - would like to add to menu for general availability.

PR #11406, opened 2022-07-15
[usage] Persist credits used for each instance in the usage store

andrew-farries:
## Description

Calculate and persist credits used for each instance in the usage store. 

On each usage reconciliation run, we store the reconciled usage in the database. This information should include the number of credits consumed by each instance.

Previously, this value was set to 0 for all instances.

## Related Issue(s)

Part of https://github.com/gitpod-io/gitpod/issues/10323

## How to test

Unit tests.

To test manually, run the usage component locally against the database from a preview environment that has had at least one workspace started in it. When the reconciler runs, see that the value for `creditsUsed` is populated in the `d_b_workspace_instance_usage` table:

image

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/usage/pkg/controller/reconciler.go
+++ b/components/usage/pkg/controller/reconciler.go
@@ -33,11 +33,17 @@ func (f ReconcilerFunc) Reconcile() error {
 type UsageReconciler struct {
 	nowFunc           func() time.Time
 	conn              *gorm.DB
+	pricer            *WorkspacePricer
 	billingController BillingController
 }
 
-func NewUsageReconciler(conn *gorm.DB, billingController BillingController) *UsageReconciler {
-	return &UsageReconciler{conn: conn, billingController: billingController, nowFunc: time.Now}
+func NewUsageReconciler(conn *gorm.DB, pricer *WorkspacePricer, billingController BillingController) *UsageReconciler {
+	return &UsageReconciler{
+		conn:              conn,
+		pricer:            pricer,
+		billingController: billingController,
+		nowFunc:           time.Now,
+	}
 }
 
 type UsageReconcileStatus struct {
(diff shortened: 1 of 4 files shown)


components/usage/pkg/controller/reconciler.go
```diff
@@ -247,7 +248,7 @@ func usageReportToUsageRecords(report UsageReport) []db.WorkspaceInstanceUsage {
 				AttributionID: attributionId,
 				StartedAt:     instance.CreationTime.Time(),
 				StoppedAt:     stoppedAt,
-				CreditsUsed:   0,
+				CreditsUsed:   float64(pricer.CreditsUsedByInstance(&instance, now)),
```
geropl, line 251: about the float64: I faintly recall we report credits as ints to stripe, right? If yes, we probably should do the same here, and adjust the DB/gRPC model if they don't agree.

@andrew-farries WDYT?

geropl (approved):
Nice! Really cool to see the piece snap together :upside_down_face: 

/hold because of float vs int [question](https://github.com/gitpod-io/gitpod/pull/11406/files#r922071042). We can also do that as a follwo-up :+1:

components/usage/pkg/controller/reconciler.go
andrew-farries: The designs here [internal](https://gitpod.slack.com/archives/C027AUU7BC3/p1657709179313939) show credits displayed with fractions.

But I think you are right in that we should be consistent in allowing the credits per minute for a workspace class be fractional but always bill and display whole numbers of credits.

I'll update the types in a follow-up.

jldec: My understanding was that we have to report total usage to Stripe as an int, but we should collect usage per session in fractional units to avoid rounding errors.

andrew-farries: As explained [here](https://github.com/gitpod-io/gitpod/pull/11429#issuecomment-1187105119), @jldec is right in thinking that this should in fact remain a float.

Issue #11303, opened 2022-07-12
[local-preview] Reduce resource usage


Pothulapati:
## Is your feature request related to a problem? Please describe

While `local-preview` is for users who want to try Gitpod locally, it may not be as accessible yet
because there hasn't been work done to make it as light as possible. There are a bunch of obvious
 improvements that can be done:

- [x] Remove `cert-manager` dependency https://github.com/gitpod-io/gitpod/pull/11412
- [x] Reduce PVC storage limits https://github.com/gitpod-io/gitpod/pull/12761

This is a global issue for all those improvements that can be done

## Describe the behaviour you'd like

`local-preview` being as light as possible.

PR #11412, opened 2022-07-15
[local-preview] Remove `cert-manager` dependency

Pothulapati:
## Description

This PR removes the dependency of `cert-manager` and thus
reducing resource usage. This is replaced by the usage of
`mkcert` instead. 

Each certificate is induvidually generated on those domains
and stored into specific files for `k3s` to apply them which
is then used by Gitpod.

## Related Issue(s)

Part of #11303 

## Release Notes

```release-note
[local-preview] Remove `cert-manager` dependency
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/install/preview/entrypoint.sh
+++ b/install/preview/entrypoint.sh
@@ -54,45 +54,121 @@ cat "${HOME}"/.local/share/mkcert/rootCA.pem >> /etc/ssl/certs/ca-certificates.c
 # also send root cert into a volume
 cat "${HOME}"/.local/share/mkcert/rootCA.pem > /var/gitpod/gitpod-ca.crt
 
-cat << EOF > /var/lib/rancher/k3s/server/manifests/ca-pair.yaml
+FN_CACERT="./ca.pem"
+FN_SSLCERT="./ssl.crt"
+FN_SSLKEY="./ssl.key"
+
+cat "${HOME}"/.local/share/mkcert/rootCA.pem > "$FN_CACERT"
+mkcert -cert-file "$FN_SSLCERT" \
+  -key-file "$FN_SSLKEY" \
+  "*.ws.${DOMAIN}" "*.${DOMAIN}" "${DOMAIN}" "reg.${DOMAIN}" "registry.default.svc.cluster.local" "gitpod.default" "ws-manager.default.svc" "ws-manager" "ws-manager-dev" "registry-facade" "server" "ws-manager-bridge" "ws-proxy" "ws-manager" "ws-daemon.default.svc" "ws-daemon" "wsdaemon"
+
+CACERT=$(base64 -w0 < "$FN_CACERT")
+SSLCERT=$(base64 -w0 < "$FN_SSLCERT")
+SSLKEY=$(base64 -w0 < "$FN_SSLKEY")
+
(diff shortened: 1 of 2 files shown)


install/preview/entrypoint.sh
```diff
@@ -54,45 +54,121 @@ cat "${HOME}"/.local/share/mkcert/rootCA.pem >> /etc/ssl/certs/ca-certificates.c
...
 
-cat << EOF > /var/lib/rancher/k3s/server/manifests/ca-pair.yaml
+FN_CACERT="./ca.pem"
+FN_SSLCERT="./ssl.crt"
+FN_SSLKEY="./ssl.key"
+
+cat "${HOME}"/.local/share/mkcert/rootCA.pem > "$FN_CACERT"
+mkcert -cert-file "$FN_SSLCERT" \
+  -key-file "$FN_SSLKEY" \
+  "*.ws.${DOMAIN}" "*.${DOMAIN}" "${DOMAIN}" "reg.${DOMAIN}" "registry.default.svc.cluster.local" "gitpod.default" "ws-manager.default.svc" "ws-manager" "ws-manager-dev" "registry-facade" "server" "ws-manager-bridge" "ws-proxy" "ws-manager" "ws-daemon.default.svc" "ws-daemon" "wsdaemon"
```
Pothulapati, line 64: Looks like this is enough, and we don't have to generate induvidual certs for each purpose.

MrSimonEmms:
I'm using the command `docker run -p 443:443 --privileged --name gitpod --rm -it -v gitpod:/var/gitpod eu.gcr.io/gitpod-core-dev/build/local-preview:tar-lp-cert-manager.11` to test - I believe this is the correct command

Pothulapati:
Correct @MrSimonEmms

MrSimonEmms (dismissed):
Works beautifully. Only question is about the certificate name - if it's not possible or want to do later, feel free to remove the hold

Pothulapati:
It's a valid question, and looks like there is no direct `mkcert` way of setting the CA name as per https://github.com/FiloSottile/mkcert/issues/260.

I just remembered I used https://github.com/smallstep/cli for similar tasks, and can try moving to that unless there is some concern using it. 🤔

Pothulapati:
So, I saw a similar image build error initially on this same change, but couldn't repro it. It could be a real issue. As this is just an improvement, I'd err on making sure if its a real problem or not before merging.

@adrienthebo Can you also PTAL once? 👀 😄

adrienthebo:
I got a successful build when creating an environment against https://github.com/mrzarquon/gitpod-aws-toolkit (as that's where I'm doing most of my work right now). This should be good to go, but I'm running a test against gitpod-io/gitpod to see if I can reproduce the failure that you're seeing @MrSimonEmms.

adrienthebo (approved):
LGTM!

MrSimonEmms:
I'm getting the same issue on both this PR and `latest`, so it's almost certainly not a problem introduced by this PR (I've also upgraded from Ubuntu 21.10 to 22.04, so it's probably my computer that's broken it)

Pothulapati:
Removing the hold, but will continue testing and raise a separate issue if its something that we might have to fix.

/unhold

PR #11419, opened 2022-07-15
[ws-manager] do not overwrite failed state

sagor999:
## Description

Do not clear up any failed state, as that will trip server in thinking that workspace has not failed, when in reality it did.
Which in turn causes subsequent workspace start to fail as server will instruct to restore workspace from backup.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/10315

## Release Notes

```release-note
none
```

## Werft options:

- [ ] /werft with-preview


aledbf (approved):

sagor999:
/hold
to fix fixtures test

PR #11420, opened 2022-07-15
[alerts] change load avg alert to critical

sagor999:
## Description

Add critical alert for when load avg is high for too long.

First, why change to critical alert:
if load average is high for too long, it will affect all workspaces on that node, so we want to react to that fast, hence critical alert (otherwise on call person can potentially be working on something else and might miss this for some time).

Why change duration to 10 min:
We want to minimize amount of false positives as much as possible. We should alert\wake up someone only on sustained high load average. 

Why change load avg check to above 10:
Again, we want to be alerted only on legit signal. Load avg above 10 for more then 10 min should signal in almost all cases that something really bad is going on on that node. 

## Related Issue(s)

Fixes #

## Release Notes

```release-note
none
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/operations/observability/mixins/workspace/rules/components/nodes/alerts.libsonnet
+++ b/operations/observability/mixins/workspace/rules/components/nodes/alerts.libsonnet
@@ -13,14 +13,15 @@
             alert: 'GitpodWorkspaceNodeHighNormalizedLoadAverage',
             labels: {
               severity: 'warning',
+              team: 'workspace'
             },
-            'for': '2m',
+            'for': '10m',
             annotations: {
-              runbook_url: 'https://github.com/gitpod-io/observability/blob/main/runbooks/GitpodWorkspaceNodeHighNormalizedLoadAverage.md',
-              summary: "Workspace node's normalized load average is higher than 3 for more than 2 minutes. Check for abuse.",
+              runbook_url: 'https://github.com/gitpod-io/runbooks/blob/main/runbooks/GitpodWorkspaceNodeHighNormalizedLoadAverage.md',
+              summary: "Workspace node's normalized load average is higher than 10 for more than 10 minutes. Check for abuse.",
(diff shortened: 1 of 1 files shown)


sagor999:
/hold
wait for runbook to be merged first: https://github.com/gitpod-io/runbooks/pull/371

meysholdt:
in the platform sync, the concern came up that this alert would be too noisy and lead to animosity among on-callers. Adding @vulkoingim as a reviewer to share ideas on how to make it less noisy.

vulkoingim:
Heya, we discussed this a bit in the platform sync. I think that looking at the 1 min load average is not a good indicator of high load - it's too short of an interval and it's common to have high load over a short time so you'll be alerting on spikes. Ideally you should be looking at 5m or, even better 15m. There's no recording rule for the higher timeframe, so I would suggest to add one similar to: https://github.com/gitpod-io/gitpod/blob/main/operations/observability/mixins/workspace/rules/components/nodes/rules.libsonnet

operations/observability/mixins/workspace/rules/components/nodes/alerts.libsonnet
```diff
@@ -12,15 +12,15 @@
           {
             alert: 'GitpodWorkspaceNodeHighNormalizedLoadAverage',
             labels: {
-              severity: 'warning',
+              severity: 'critical',
```
kylos101, line 15: @sagor999 can you change this to:

```
 labels: {
              severity: 'warning',
              team: 'workspace'
            },
```

That'll route here, https://gitpod.slack.com/archives/C03QZ1NH517, and give us a chance to measure if this is okay to promote to PagerDuty, etc.

kylos101 (changes requested):
@sagor999 one suggestion that I think will make this more palatable

kylos101 (approved):
:rocket: :pray:

PR #11429, opened 2022-07-18
[usage] Store credits as integer values in the usage table

andrew-farries:
## Description

Workspace credits per minute can be configured as fractional values, but we always want to bill and display usage in whole numbers of credits. This PR changes the type of the field that stores credits consumed per workspace instance to be an integer.

## Related Issue(s)

Part of #9036 and [this comment](https://github.com/gitpod-io/gitpod/pull/11406#discussion_r922071042) on #11406.

## How to test

Connect to the database for the preview and describe the `d_b_workspace_instance_usage` table:

image

## Release Notes

```release-note
NONE
```

## Werft options:
- [x] /werft with-preview

### DIFF
--- a/components/gitpod-db/src/typeorm/entity/db-workspace-instance-usage.ts
+++ b/components/gitpod-db/src/typeorm/entity/db-workspace-instance-usage.ts
@@ -34,7 +34,7 @@ export class DBWorkspaceInstanceUsage {
     @Index("ind_stoppedAt")
     stoppedAt: string;
 
-    @Column("int")
+    @Column("bigint")
     creditsUsed: number;
 
     @Column()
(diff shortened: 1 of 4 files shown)


geropl:
`int(11)` sounds potentially smaller than `int64` which we use in code. Maybe a different length or `bigint` is more suitable? 🤔

andrew-farries:
/werft run with-preview=true with-clean-slate-deployment=true

andrew-farries:
Changed to `bigint` in 3c12d8c36f27444bdc8715d09a03be4badfca114.

I think this is unnecessarily large though, but it's now consistent with the type used elsewhere.

easyCZ:
Regular workspace costs 1 credit per 6 minutes, there's no way we'd be able to express workspace that ran for 5 minutes, or other fractions in this way.

andrew-farries:
> Indeed, the stored value should remain a float. This is because only the total billed value should be rounded, not each workspace instance.

In that case, I believe we are incorrectly rounding in the billing data sent to Stripe:

https://github.com/gitpod-io/gitpod/blob/a86efd21242fa99c342f98a33b6c989b26fb69f4/components/usage/pkg/controller/reconciler.go#L143-L146

Here we round each workspace instance's credits, rather than rounding the total once at the end.

Issue #11350, opened 2022-07-13
[local-preview] Allow for opt out of telemetry


lucasvaltl:
## Is your feature request related to a problem? Please describe
Currently we send a telemetry event upon successful installation (note that this does NOT include any personally identifyable inf). We should allow users to opt out of this if they wish. 

## Describe the behaviour you'd like
- A flag when running the command that disables analytics - ideally we follow known standards here, but I could not find any in a quick search. 
- A callout in our documentation mentioning the tracking and how to opt out

## Additional context
- Snyk does something similar [here](https://support.snyk.io/hc/en-us/articles/360000922058-What-analytics-do-you-track-How-can-I-opt-out-Check-)

PR #11430, opened 2022-07-18
[local-preview] honour `DO_NOT_TRACK` env variable

Pothulapati:
## Description

Currently, There is no way to disable telemetry for `local-preview`
as we run the cronjob manaully during init, when Gitpod is ready.

This fixes that by honouring [`DO_NOT_TRACK` envrionment variable](https://consoledonottrack.com/)
, which when set prevent the script from triggering the cronjob manually.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/11350

## How to test

Add `-e DO_NOT_TRACK=1` to the following docker run command
```
docker run -p 443:443 --privileged --name gitpod --rm -it -v gitpod:/var/gitpod eu.gcr.io/gitpod-core-dev/build/local-preview
```

## Release Notes

```release-note
[local-preview] honour `DO_NOT_TRACK` env variable.
```

## Werft options:

- [ ] /werft with-preview


install/preview/entrypoint.sh
```diff
@@ -143,8 +143,12 @@ run_telemetry(){
   sleep 100
   # indefinitely wait for Gitpod pods to be ready
   kubectl wait --timeout=-1s --for=condition=ready pod -l app=gitpod,component!=migrations
-  # manually tun the cronjob
-  kubectl create job gitpod-telemetry-init --from=cronjob/gitpod-telemetry
+  echo "Gitpod pods are ready"
+  # honour DO_NOT_TRACK if set
+  if [ -z "${DO_NOT_TRACK}" ]; then
+    # manually tun the cronjob
+    kubectl create job gitpod-telemetry-init --from=cronjob/gitpod-telemetry
```
Pothulapati, line 150: had to move this inside as `run_telemetry` also is useful for readyiness and showing that Gitpod is ready.

MrSimonEmms (approved):
Looks good. Just a thought, the `DO_NOT_TRACK` also works if you set `DO_NOT_TRACK=0` - is it worth (perhaps in a future PR) setting the if statement to be "IF variable is set AND is equal to 1"?

/hold approving and you can decide how you want to resolve this (if indeed you do)

Pothulapati:
@lucasvaltl Yep, You are right. It would still run daily. My reason for knowingly skipping was that `local-preview` is short-lived, but we can indeed disable the telemetry job. Will update the PR based on the complexity it takes.

Pothulapati:
@lucasvaltl Updated the PR to also disable the daily telemetry job when `DO_NOT_TRACK` is set.

PR #11433, opened 2022-07-18
[ssh-gateway] Reject ssh connection when workspace not found

### LINKED ISSUE #11428: Reject ssh connection when trying to connect to a stopped workspace
Currently if we try to connect to a stopped workspace using SSH it will ask to authenticate using a password which is misleading that could indicate that connection is possible, instead the server should reject the connection as it doesn't make sense to connect to a stopped workspace

iQQBot:
## Description

[ssh-gateway] Reject ssh connection when workspace not found

For reviewer: In general, the SSH authentication process is as follows
`NoClientAuthCallback` -> `PublicKeyCallback(if private key provider)` -> `PasswordCallback`
there are 2 special error type

1. `ssh.ErrNoAuth` This Error only exists in the `NoClientAuthCallback`, which means that this callback is skipped and not counted as attempts
2. `ssh.ErrDenied` This Error means that retries are no longer allowed and the validation process is complete.
All error created by `NewSSHErrorWithReject` is base on `ssh.ErrDenied`
The new authentication process is simple than before

`NoClientAuthCallback` responsible for checking if workspace exists, if not, close the connection, if it does, determine if the username is in the form of `workspaceId#ownerToken`, if so, verify ownerToken, if the authentication fails, close the connection
`PublicKeyCallback` responsible for verifying the private key, the username check is removed here because the `NoClientAuthCallback` has already done it for us, so there is only one possibility that the username is a exists workspaceID.
`PasswordCallback` is similar with `PublicKeyCallback` but verify password (ownerToken).

## Related Issue(s)

Fixes #11428

## How to test

1. start a workspace in preview environment
2. connect ssh via copy/paste, it should be work
3. stop workspace
4. try use the previous command connect, it should terminated with error `Connection closed` instead of ask your provider password

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/ws-proxy/pkg/sshproxy/server.go
+++ b/components/ws-proxy/pkg/sshproxy/server.go
@@ -43,18 +43,22 @@ var (
 )
 
 var (
-	ErrWorkspaceNotFound  = NewSSHError("WS_NOTFOUND", "not found workspace")
-	ErrWorkspaceIDInvalid = NewSSHError("WS_ID_INVALID", "workspace id invalid")
-	ErrAuthFailed         = NewSSHError("AUTH_FAILED", "auth failed")
-	ErrUsernameFormat     = NewSSHError("USER_FORMAT", "username format is not correct")
-	ErrMissPrivateKey     = NewSSHError("MISS_KEY", "missing privateKey")
+	ErrWorkspaceNotFound  = NewSSHErrorWithReject("WS_NOTFOUND", "not found workspace")
+	ErrWorkspaceIDInvalid = NewSSHErrorWithReject("WS_ID_INVALID", "workspace id invalid")
+	ErrUsernameFormat     = NewSSHErrorWithReject("USER_FORMAT", "username format is not correct")
+	ErrMissPrivateKey     = NewSSHErrorWithReject("MISS_KEY", "missing privateKey")
 	ErrConnFailed         = NewSSHError("CONN_FAILED", "cannot to connect with workspace")
 	ErrCreateSSHKey       = NewSSHError("CREATE_KEY_FAILED", "cannot create private pair in workspace")
+
+	ErrAuthFailed = NewSSHError("AUTH_FAILED", "auth failed")
+	// ErrAuthFailedWithReject is same with ErrAuthFailed, it will just disconnect immediately to avoid pointless retries
+	ErrAuthFailedWithReject = NewSSHErrorWithReject("AUTH_FAILED", "auth failed")
 )
 
 type SSHError struct {
 	shortName   string
 	description string
+	err         error
 }
 
 func (e SSHError) Error() string {
@@ -64,9 +68,16 @@ func (e SSHError) Error() string {
 func (e SSHError) ShortName() string {
 	return e.shortName
 }
+func (e SSHError) Unwrap() error {
+	return e.err
+}
 
 func NewSSHError(shortName string, description string) SSHError {
 	return SSHError{shortName: shortName, description: description}
+}
+
+func NewSSHErrorWithReject(shortName string, description string) SSHError {
+	return SSHError{shortName: shortName, description: description, err: ssh.ErrDenied}
 }
 
 type Session struct {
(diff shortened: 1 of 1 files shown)


components/ws-proxy/pkg/sshproxy/server.go
```diff
@@ -43,18 +43,20 @@ var (
...
-	ErrWorkspaceIDInvalid = NewSSHError("WS_ID_INVALID", "workspace id invalid")
-	ErrAuthFailed         = NewSSHError("AUTH_FAILED", "auth failed")
-	ErrUsernameFormat     = NewSSHError("USER_FORMAT", "username format is not correct")
-	ErrMissPrivateKey     = NewSSHError("MISS_KEY", "missing privateKey")
-	ErrConnFailed         = NewSSHError("CONN_FAILED", "cannot to connect with workspace")
-	ErrCreateSSHKey       = NewSSHError("CREATE_KEY_FAILED", "cannot create private pair in workspace")
+	ErrWorkspaceNotFound    = NewSSHErrorWithReject("WS_NOTFOUND", "not found workspace")
+	ErrWorkspaceIDInvalid   = NewSSHErrorWithReject("WS_ID_INVALID", "workspace id invalid")
+	ErrAuthFailed           = NewSSHError("AUTH_FAILED", "auth failed")
+	ErrAuthFailedWithReject = NewSSHErrorWithReject("AUTH_FAILED", "auth failed")
```
akosyakov, line 49: SHOULD we use another code?

components/ws-proxy/pkg/sshproxy/server.go
```diff
@@ -43,18 +43,20 @@ var (
 )
 
 var (
```
akosyakov, line 45: it would be nice to add comment to each code to specify meaning, i.e. auth with reject means that workspace is not running?

iQQBot: no auth with reject same with auth, but quick end, because at this time, none of password is correct

components/ws-proxy/pkg/sshproxy/server.go
iQQBot: No, it's same, just quickly terminate

components/ws-proxy/pkg/sshproxy/server.go
```diff
@@ -64,11 +68,18 @@ func (e SSHError) Error() string {
...
 }
+func (e SSHError) Unwrap() error {
+	return e.err
+}
 
 func NewSSHError(shortName string, description string) SSHError {
 	return SSHError{shortName: shortName, description: description}
 }
 
+func NewSSHErrorWithReject(shortName string, description string) SSHError {
```
iQQBot, line 79: This means that no more attempts are allowed and the connection is simply disconnected

akosyakov:
> I expectation we only support workspaceID@workspaceUrl for private key, and workspaceID#ownerToken@workspaceUrl for none private key
> currently we support workspaceID#wrongOwnerToken@workspaceUrl for private key, which is not necessary

ok, let's try, @jeanp413 @felladrin could you check that it is aligned with expectations in JB GW plugin and VS Code Desktop extension 🙏

jeanp413:
@iQQBot it works but I see in the logs that it still tries to authenticate with `gssapi-with-mic` something with kerberos mayeb we disbale it in server too using `GSSAPIAuthentication no` in `sshd_config`

jeanp413 (approved):

akosyakov:
@felladrin Assigned you to test how JB GW behaves with this change.

felladrin:
I'm checking it now.

felladrin (approved):
Besides testing the SSH connection, as instructed in the How to Test section, I checked JetBrains IntelliJ (Stable and EAP) and JetBrains Gateway (Stable and EAP).
- Tested the full connection flow (start, connect, stop) (start, connect, disconnect, stop). All good.
- Tested trying to connect simultaneously as I stopped the workspaces, and Gateway will simply indicate the workspace is stopping/stopped and won't try to connect.
  image
- Tested trying to connect to non-running workspaces, and I couldn't find any way to do it in JetBrains Gateway flow. It will always take the user to the workspace page where they need to re-start it.

aledbf (approved):

Issue #11381, opened 2022-07-14
[Local-Preview] In-product nudge towards next installation steps (i.e. PoV reference Architecture)


lucasvaltl:
## Is your feature request related to a problem? Please describe
When a user successfully installed and seems to be using / happy with their installation, we should nudge them towards the logical next step in the user journey: to install Gitpod on K8s, likely via our PoV Reference architecture (https://github.com/gitpod-io/gitpod/issues/10622). 

## Describe the behaviour you'd like
- When a user has successfully installed and started 1 workspace, there should be a toast that asks them to go to the next installation method. 
- The amount of workspaces needed until we nudge should be configurable
- The nudge should point to the community license form, however with attribution (we should be able to track that a user came from a preview installation) - e.g. using an UTM link
- The same nudge should also be in the documentation.
- This only happens in the local preview

## Describe alternatives you've considered
- Only do this in the documentation - which might be overlooked.

## Additional context
- Internal [planning document](https://www.notion.so/gitpod/Local-Preview-Plans-f9c200ebfa154bf7a6ca3c550204aac4)
- Internal [release plan](https://www.notion.so/gitpod/Release-Plan-Local-Preview-a635787b23a84649967a64819e6be79f)

## Implementation Ideas:
- Can show a goodbye message ("thanks for trying Gitpod, go to this link for next steps") on terminate signal in the CLI (with condition "has opened a workspace")
- AND show this within the dashboard - see [internal discussion ](https://gitpod.slack.com/archives/C02EN94AEPL/p1657872926410249)
  - We can hide this behind a feature flag such that it only gets shown in local preview mode - see [this function](https://github.com/gitpod-io/gitpod/blob/fdc44f4beab2c0cdcac56cc71b19f2b64553418b/components/dashboard/src/utils.ts#L52-L59)
  - From @gtsiolis: I’d suggest to use a non-closable, info variant of the alert component variant placed at as a global page alert (see examples[[1](https://github.com/gitpod-io/gitpod/issues/9974#issuecomment-1125023447)][[2](https://www.figma.com/file/jorFpdvydV22oRc0YqKWGo/Add-phone-verification-for-new-users-%2311339?node-id=1402%3A711)]).
    - The content of the alert could inform users they are running a preview that is not intended production usage, and provide a call-to-action (CTA) to install with a community license, contact sales for buying an enterprise license, or later on start a free trial of the enterprise license.

Pothulapati:
[...]

PR #11434, opened 2022-07-18
[dashboard] Next steps nudge for local-preview 

Pothulapati:
## Description

Currently, After `local-preview` is exited There are no
direct concrete steps for users to take.

This PR fixes this by adding a new `Alert` box to the
global dashboard if we notice that they are on a Gitpod
`local-preview` DOMAIN, which is `preview.gitpod-self-hosted.com`
(our own DOMAIN) from now.

## Related Issue(s)

Part of https://github.com/gitpod-io/gitpod/issues/11381

## How to test

Run
```
docker run -p 443:443 --privileged --name gitpod --rm -it -v gitpod:/var/gitpod eu.gcr.io/gitpod-core-dev/build/local-preview:tar-lp-out-nudge.19
````
and see the nudge being present.

## Release Notes

```release-note
[dashboard] Next steps nudge for local-preview
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/dashboard/src/App.tsx
+++ b/components/dashboard/src/App.tsx
@@ -47,7 +47,8 @@ import { StartWorkspaceModal } from "./workspaces/StartWorkspaceModal";
 import { parseProps } from "./start/StartWorkspace";
 import SelectIDEModal from "./settings/SelectIDEModal";
 import { StartPage, StartPhase } from "./start/StartPage";
-import { isGitpodIo } from "./utils";
+import { isGitpodIo, isLocalPreview } from "./utils";
+import Alert from "./components/Alert";
 import { BlockedRepositories } from "./admin/BlockedRepositories";
 import { AppNotifications } from "./AppNotifications";
 
@@ -347,6 +348,28 @@ function App() {
         <Route>
             <div className="container">
                 <Menu />
+                {isLocalPreview() && (
+                    <div className="app-container mt-2">
+                        <Alert type="warning" className="app-container rounded-md">
+                            You are using a <b>local preview</b> installation, intended for exploring the product on a
+                            single machine without requiring a Kubernetes cluster.{" "}
+                            <a
+                                className="gp-link hover:text-gray-600"
+                                href="https://www.gitpod.io/community-license?utm_source=local-preview"
+                            >
+                                Request a community license
+                            </a>{" "}
+                            or{" "}
+                            <a
+                                className="gp-link hover:text-gray-600"
+                                href="https://www.gitpod.io/contact/sales?utm_source=local-preview"
+                            >
+                                contact sales
+                            </a>{" "}
+                            to get a professional license for running Gitpod in production.
+                        </Alert>
+                    </div>
+                )}
                 <AppNotifications />
                 <Switch>
                     <Route path={projectsPathNew} exact component={NewProject} />
--- a/components/dashboard/src/utils.ts
+++ b/components/dashboard/src/utils.ts
@@ -58,8 +58,12 @@ export function isGitpodIo() {
     );
 }
 
+export function isLocalPreview() {
+    return window.location.hostname === "preview.gitpod-self-hosted.com";
+}
+
 function trimResource(resource: string): string {
-    return resource.split('/').filter(Boolean).join('/');
+    return resource.split("/").filter(Boolean).join("/");
 }
 
 // Returns 'true' if a 'pathname' is a part of 'resources' provided.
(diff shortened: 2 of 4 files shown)


roboquat:

<details>

</details>

components/dashboard/src/utils.ts
```diff
@@ -58,8 +58,12 @@ export function isGitpodIo() {
     );
 }
 
+export function isLocalPreview() {
+    return true;
```
Pothulapati, line 62: This will be replaced with `return window.location.hostname === "nip.io` before merging

Pothulapati:
Screenshot of what a user would see 
Screenshot 2022-07-18 at 3 49 51 PM

@gtsiolis Looking forward to your feedback 👀

components/dashboard/src/utils.ts
Pothulapati: Discussed with @MrSimonEmms, and feeling it would be better to be much more specific here, instead of relying on this.

gtsiolis:
Looking at this now! 👀

components/dashboard/src/Menu.tsx
```diff
@@ -489,6 +493,18 @@ export default function Menu() {
                         ))}
                     </nav>
                 )}
+                {isLP && (
```
gtsiolis, line 496: **issue**: Not sure where can we move this piece of code, but the alert component component should be placed below the top navigation bar, including the tabs and above the pager heading. The current placement is breaking the layout in some pages.

| BEFORE | AFTER |
|-|-|
| Screenshot 2022-07-18 at 6 13 55 PM (2) | Screenshot 2022-07-18 at 6 17 16 PM (2) |
| Screenshot 2022-07-18 at 6 57 40 PM (2) | Screenshot 2022-07-18 at 6 57 32 PM (2) |

components/dashboard/src/Menu.tsx
```diff
@@ -489,6 +493,18 @@ export default function Menu() {
...
+                    <Alert type="warning" className="app-container rounded-md">
+                        This is a local-preview instance. Visit{" "}
```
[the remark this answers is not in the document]
lucasvaltl: Very much agree we should polish the copy here. I really like your proposal for the wording in the "after" screenshot :)

components/dashboard/src/Menu.tsx
Pothulapati: @gtsiolis Can you take a look on this layout issue?

I've moved this into `App.tsx` where the new AppNotifications are, and you can access the instance at https://tar-lp-out-nudge.preview.gitpod-dev.com/workspaces

gtsiolis: Looks great, @Pothulapati! Two minor issues:

1. **Grid**: We're using a **page container** with a **fixed layout** for our **grid layout**. Using this approach the ideal maximum width of the page container should follow the fixed width of the page container.
2. **Spacing**: We've been using an **8-point grid system** with **spacing scale** to help us provide all the negative area needed between elements and components, see [relevant discussion](https://gitpod.slack.com/archives/C01KPEPQYE6/p1657294999865219) (internal). Every part of the user interface should be intentional about empty space between elements. The amount of space between items creates relationships and hierarchy. For example, `8px` is used to separate related elements while `16px` is used to separate unrelated elements.

Cc @gitpod-io/engineering-self-hosted for visibility and future reference.

One interesting point we may run into the future could be to see how this alert stacks with more global alerts coming through the `AppNotifications.txt`, which could be fine as long as we limit the amount of alerts we stack, but I'd assume there's little to zero changes of running into more alerts here as the local preview is not intended for production use.

To get back to the actual UX changes, using a parent `div` container for the alert with two classes (`app-container` and `mt-2`) should make all the necessary changes here. 🍵 

| BEFORE | AFTER |
|-|-|
| Screenshot 2022-08-01 at 3 56 30 PM | Screenshot 2022-08-01 at 3 56 25 PM |

install/preview/entrypoint.sh
```diff
@@ -4,9 +4,9 @@
 
 set -e
 
-# Set Domain to `127-0-0-1.nip.io` if not set
+# Set Domain to `preview.gitpod-self-hosted.com` if not set
 if [ -z "${DOMAIN}" ]; then
-  export DOMAIN="127-0-0-1.nip.io"
+  export DOMAIN="preview.gitpod-self-hosted.com"
```
MrSimonEmms, line 9: Excellent. Glad my fag-packet thought worked

MrSimonEmms (approved):

components/dashboard/src/Menu.tsx
Pothulapati: > I'd assume there's little to zero changes of running into more alerts here as the local preview is not intended for production use.

Exactly, and hence my reason for not using `AppNotifications`.

components/dashboard/src/App.tsx
```diff
@@ -157,6 +158,8 @@ function App() {
     const [isWhatsNewShown, setWhatsNewShown] = useState(false);
     const [showUserIdePreference, setShowUserIdePreference] = useState(false);
     const [isSetupRequired, setSetupRequired] = useState(false);
+    const isLP = isLocalPreview();
```
andrew-farries, line 161: Can we use `isLocalPreview` here instead of the abbreviation?

andrew-farries (approved):

PR #11439, opened 2022-07-18
[usage] Store user, project, workspace, type & class + migration

easyCZ:
## Description

Adds extra fields to workspace instance usage such that we can perform common groupings easily

## Related Issue(s)

Fixes #

## How to test

Preview 

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/1658144661198-WorkspaceInstanceUsageAddExtraFields.ts
+++ b/1658144661198-WorkspaceInstanceUsageAddExtraFields.ts
@@ -1,0 +1,26 @@
+import {MigrationInterface, QueryRunner} from "typeorm";
+import {columnExists} from "@gitpod/gitpod-db/lib/typeorm/migration/helper/helper";
+
+export class WorkspaceInstanceUsageAddExtraFields1658144661198 implements MigrationInterface {
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+
+        const columns = ['userId', 'workspaceId', 'projectId', 'workspaceType', 'workspaceClass'];
+
+        const statements = columns
+            .filter(async col => {
+                const exists = await columnExists(queryRunner, "d_b_workspace_instance_usage", col)
+                return !exists;
+            })
+            .map(col => {
+                return `ADD COLUMN ${col} varchar(255) NOT NULL DEFAULT ''`;
+            });
+
+        if (statements.length > 0) {
+            await queryRunner.query(`ALTER TABLE \`d_b_workspace_instance_usage\` ${statements.join(", ")}`);
+        }
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {}
+
+}
--- a/components/gitpod-db/src/typeorm/entity/db-workspace-instance-usage.ts
+++ b/components/gitpod-db/src/typeorm/entity/db-workspace-instance-usage.ts
@@ -7,6 +7,7 @@
 import { Entity, Column, PrimaryColumn, Index } from "typeorm";
 import { TypeORM } from "../typeorm";
 import { Transformer } from "../transformer";
+import { WorkspaceType } from "@gitpod/gitpod-protocol";
 
 @Entity()
 export class DBWorkspaceInstanceUsage {
@@ -16,6 +17,21 @@ export class DBWorkspaceInstanceUsage {
     @Column("varchar")
     @Index("ind_attributionId")
     attributionId: string;
+
+    @Column("varchar")
+    userId: string;
+
+    @Column("varchar")
+    workspaceId: string;
+
+    @Column("varchar")
+    projectId: string;
+
+    @Column("varchar")
+    workspaceType: WorkspaceType;
+
+    @Column("varchar")
+    workspaceClass: string;
 
     @Column({
         type: "timestamp",


roboquat:

<details>

</details>

1658144661198-WorkspaceInstanceUsageAddExtraFields.ts
```diff
@@ -0,0 +1,26 @@
+import {MigrationInterface, QueryRunner} from "typeorm";
+import {columnExists} from "@gitpod/gitpod-db/lib/typeorm/migration/helper/helper";
+
+export class WorkspaceInstanceUsageAddExtraFields1658144661198 implements MigrationInterface {
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+
+        const columns = ['userId', 'workspaceId', 'projectId', 'workspaceType', 'workspaceClass'];
+
+        const statements = columns
```
geropl, line 10: I like that pattern! :smile:

1658144661198-WorkspaceInstanceUsageAddExtraFields.ts
```diff
@@ -0,0 +1,26 @@
...
+                const exists = await columnExists(queryRunner, "d_b_workspace_instance_usage", col)
+                return !exists;
+            })
+            .map(col => {
+                return `ADD COLUMN ${col} varchar(255) NOT NULL DEFAULT ''`;
+            });
+
+        if (statements.length > 0) {
+            await queryRunner.query(`ALTER TABLE \`d_b_workspace_instance_usage\` ${statements.join(", ")};`);
+        }
```
geropl, line 21: Note: I think statements here may _not_ end with `;` and will fail if they do for reasons unknown to mankind. Please double check what other migrations do. :+1:

components/gitpod-db/src/typeorm/entity/db-workspace-instance-usage.ts
```diff
@@ -17,6 +17,21 @@ export class DBWorkspaceInstanceUsage {
...
+
+    @Column("varchar")
+    workspaceId: string;
+
+    @Column("varchar")
+    projectId: string;
+
+    @Column("varchar")
+    workspaceType: string;
+
```
andrew-farries, line 31: `workspaceType` has a type defined for it in `gitpod-protocol` so we should use that here rather than `string`.

https://github.com/gitpod-io/gitpod/blob/a86efd21242fa99c342f98a33b6c989b26fb69f4/components/gitpod-protocol/src/protocol.ts#L697

components/gitpod-db/src/typeorm/entity/db-workspace-instance-usage.ts
```diff
@@ -17,6 +17,21 @@ export class DBWorkspaceInstanceUsage {
...
+    userId: string;
+
+    @Column("varchar")
+    workspaceId: string;
+
+    @Column("varchar")
+    projectId: string;
+
+    @Column("varchar")
+    workspaceType: string;
```
andrew-farries, line 30: Is it important to match the column types with their occurences in other tables? I'm not sure they are always `varchar` in other tables.

1658144661198-WorkspaceInstanceUsageAddExtraFields.ts
```diff
@@ -0,0 +1,26 @@
...
+    }
```
andrew-farries, line 22: For migrations to this table before we release the feature, we can simply drop and recreate the table. This works too though - up to you.

1658144661198-WorkspaceInstanceUsageAddExtraFields.ts
easyCZ: https://www.sqltutorial.org/sql-add-column/ says it should always end with a `;`. However `1649107789640-PrebuildStatusVersionColumn.ts` doesn't have a semicolon. I'll drop mine.

geropl: Jep, it's a typeorm thing :grimacing:

1658144661198-WorkspaceInstanceUsageAddExtraFields.ts
easyCZ: I tend to think through tables as being forever. Yes, here it would work but it has the extra added brain space to remember that under some scenarios it will drop.

components/gitpod-db/src/typeorm/entity/db-workspace-instance-usage.ts
easyCZ: I had a look but couldn't find which one would not be a string. Do you know?

andrew-farries: * `workspaceId` is defined here like this: https://github.com/gitpod-io/gitpod/blob/60f4bad9db7a51f3479cfd1a53361f361f45305b/components/gitpod-db/src/typeorm/entity/db-workspace.ts#L27-L28
* `userId` like this: https://github.com/gitpod-io/gitpod/blob/60f4bad9db7a51f3479cfd1a53361f361f45305b/components/gitpod-db/src/typeorm/entity/db-user.ts#L16-L17
* `projectId` : https://github.com/gitpod-io/gitpod/blob/60f4bad9db7a51f3479cfd1a53361f361f45305b/components/gitpod-db/src/typeorm/entity/db-project.ts#L15-L16

andrew-farries: so they are all strings but with some extra annotations

andrew-farries (approved):

PR #11444, opened 2022-07-18
[usage] Add user, workspace, project, type, class to WorkspaceInstanceUsage in go

easyCZ:
## Related Issue(s)

Defined in https://github.com/gitpod-io/gitpod/pull/11439/files

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


components/usage/pkg/controller/reconciler_test.go
```diff
@@ -202,7 +201,6 @@ func TestUsageReportConversionToDBUsageRecords(t *testing.T) {
 				StartedAt:     creationTime.Time(),
 				StoppedAt:     sql.NullTime{},
 				CreditsUsed:   470,
-				GenerationId:  0,
```
andrew-farries, line 205: Why remove `GenerationId` here and on line 183?

andrew-farries (approved):
/hold

components/usage/pkg/controller/reconciler_test.go
easyCZ: 0 is the default value so it doesn't need to be set when constructing the object.

components/gitpod-db/src/typeorm/migration/1658144661198-WorkspaceInstanceUsageAddExtraFields.ts
```diff
@@ -1,18 +1,22 @@
-import {MigrationInterface, QueryRunner} from "typeorm";
-import {columnExists} from "@gitpod/gitpod-db/lib/typeorm/migration/helper/helper";
+/**
```
easyCZ, line 1: This file was added in the root of the repo in the previous PR, which is wrong, moved to the right dir.

components/usage/pkg/db/workspace_instance_usage.go
```diff
@@ -18,11 +18,22 @@ import (
...
+
+	CreditsUsed float64 `gorm:"column:creditsUsed;type:double;" json:"creditsUsed"`
+
+	StartedAt time.Time    `gorm:"column:startedAt;type:timestamp;default:CURRENT_TIMESTAMP(6);" json:"startedAt"`
+	StoppedAt sql.NullTime `gorm:"column:stoppedAt;type:timestamp;" json:"stoppedAt"`
+
+	GenerationID int `gorm:"column:generationId;type:int;" json:"generationId"`
+
+	// deleted is used by db-sync
+	_ bool `gorm:"column:deleted;type:tinyint;default:0;" json:"deleted"`
```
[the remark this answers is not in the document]
easyCZ: I was actually wrong. Because we're using the update with Conflict handling, SQL complains when there's no value set so we've gotta keep it as a Real field, we just don't need to set it.

components/gitpod-db/src/typeorm/migration/1658144661198-WorkspaceInstanceUsageAddExtraFields.ts
andrew-farries: I think we just duplicated each other's work 😊: https://github.com/gitpod-io/gitpod/pull/11458

easyCZ: I've landed yours, will rebase this one.

PR #11474, opened 2022-07-19
[usage] Add a UsageReportService to `content-service` API

andrew-farries:
## Description

As part of the move towards usage based pricing (#9036), we'd like for the usage aggregator (`components/usage`) to be able to upload its usage reports to cloud storage. This will provide an audit trail of usage reports, allowing us to cross reference usage entries in the database with the usage reports that provided the data. In future, we may also allow access to these reports to users directly.

The usage component:
* runs multiple times per hour.
* produces a report (in `json` format) of all workspace instance credit consumption for all instances that have any runtime in the current billing period (one month).
* produces reports ~60MB in size.

This PR creates a new `UsageReportService` gRPC service and registers it with `content-service`. I've manually tested getting a presigned S3 URL from the `GetUploadUrl`  API and uploaded some content to `minio` (in a preview environment) using the URL.

Following PRs will make the usage component request this pre-signed upload URL from the `content-service` and upload usage reports to it.

## Related Issue(s)
Part of #9036 

## How to test

* Port forward to minio in the preview env:
```
 kubectl port-forward svc/minio 9000:9000
```
* Run the `content-service` locally using the `config.json` from the `content-service` pod in the preview env (with `minio.endpoint` set to `localhost:9000`):
```
cd components/content-service
go run . run --config config.json 
```
* Use `evans` or another gRPC client to invoke the `uploadUrl` RPC on the new `UsageReportService`:
image

* Make an http `PUT` request to the URL in the response to upload some content to the bucket.

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


components/content-service/pkg/service/usagereport-service.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	info, err := us.s.SignUpload(ctx, req.Bucket, req.Name, &storage.SignedURLOptions{
+		ContentType: "*/*",
+	})
+	if err != nil {
+		log.WithField("name", req.Name).
+			WithField("bucket", req.Bucket).
+			WithError(err).
+			Error("error getting UsageReport SignUpload URL")
```
easyCZ, line 56:
```suggestion
			Error("Error getting UsageReport SignUpload URL")
```

components/content-service/pkg/service/usagereport-service.go
```diff
@@ -0,0 +1,63 @@
...
+	return &UsageReportService{cfg: cfg, s: s}, nil
+}
+
+// UploadURL provides a URL to which clients can upload the content via HTTP PUT.
+func (us *UsageReportService) UploadURL(ctx context.Context, req *api.UsageReportUploadURLRequest) (resp *api.UsageReportUploadURLResponse, err error) {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "UsageReport.UploadURL")
+	span.SetTag("name", req.Name)
+	defer tracing.FinishSpan(span, &err)
+
+	err = us.s.EnsureExists(ctx, req.Bucket)
```
easyCZ, line 44: Why is the bucket specified in the request? I would've thought that the UsageReportService owns the location and as such should not be passed in from the outside. In this form, it mostly provides an API for abitrary file upload which is not necessarily what we want.

easyCZ:
For the later PRs, we should gzip the report either through upload transcoding, or in the service before we upload it. That should shrink the size to about a tenth

components/content-service/pkg/service/usagereport-service.go
andrew-farries: Thanks, done 👍 (b9de21746)

components/content-service/pkg/service/usagereport-service.go
andrew-farries: Good point, thanks. I've removed the bucket name from the RPC and used a constant instead (ebba8d090)

components/content-service/pkg/service/usagereport-service.go
```diff
@@ -0,0 +1,67 @@
...
+func NewUsageReportService(cfg config.StorageConfig) (res *UsageReportService, err error) {
+	s, err := storage.NewPresignedAccess(&cfg)
+	if err != nil {
+		return nil, err
+	}
+	return &UsageReportService{cfg: cfg, s: s}, nil
+}
+
+// UploadURL provides a URL to which clients can upload the content via HTTP PUT.
+func (us *UsageReportService) UploadURL(ctx context.Context, req *api.UsageReportUploadURLRequest) (resp *api.UsageReportUploadURLResponse, err error) {
```
easyCZ, line 43: Is it possible to write a unit test for the service? Can we inject mock storage?

andrew-farries: I've added a test using mocked storage in 145519c53.

In general I'm sceptical of the value of these 'mockist' style tests. All we're doing with the test is is ensuring that we pass the right bucket and filenames through to the dependencies and return the upload URL unmodified from the dependency. 

There isn't any logic being tested there.

components/content-service/pkg/service/usagereport-service.go
```diff
@@ -0,0 +1,67 @@
...
+
+	info, err := us.s.SignUpload(ctx, bucketName, req.Name, &storage.SignedURLOptions{
+		ContentType: "*/*",
+	})
+	if err != nil {
+		log.WithField("name", req.Name).
+			WithField("bucket", bucketName).
+			WithError(err).
+			Error("Error getting UsageReport SignUpload URL")
+		if err == storage.ErrNotFound {
```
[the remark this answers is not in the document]
andrew-farries: Not sure. Digging around the various implementations (`minio`, `gcloud`), it doesn't look possible to return not found given that we've already ensured the bucket exists.

Check removed in 83fb70019.

components/content-service/pkg/service/usagereport-service.go
easyCZ: That's fair. Mocks tend to do that. For me the value is that next time we make a change, we don't have a reason to say we won't add a test because there wasn't one in the first place.

easyCZ (approved):

geropl (approved):

Furisto (approved):

PR #11484, opened 2022-07-19
[usage] Make endTime non-mandatory for billed session

easyCZ:
## Description

Because it won't be set when the instance is still running.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/server/ee/src/workspace/gitpod-server-impl.ts
+++ b/components/server/ee/src/workspace/gitpod-server-impl.ts
@@ -2132,7 +2132,7 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
             workspaceType: mandatory(s.getWorkspaceType()) as WorkspaceType,
             workspaceClass: mandatory(s.getWorkspaceClass()),
             startTime: mandatory(s.getStartTime(), (t) => t!.toDate().toISOString()),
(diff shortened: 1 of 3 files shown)


components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2132,7 +2132,7 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
             workspaceType: mandatory(s.getWorkspaceType()) as WorkspaceType,
             workspaceClass: mandatory(s.getWorkspaceClass()),
             startTime: mandatory(s.getStartTime(), (t) => t!.toDate().toISOString()),
-            endTime: mandatory(s.getEndTime(), (t) => t!.toDate().toISOString()),
+            endTime: s.getEndTime()?.toDate().toISOString() || undefined,
```
andrew-farries, line 2135: Do you need the `|| undefined`? . I think the optional chaining will ensure the expression has the value `undefined` when the end time is not set.

easyCZ: That makes sense. Let me remove. I wasn't sure exactly how the `?` works.

andrew-farries (approved):

PR #11489, opened 2022-07-19
  [ws-manager] fix incorrect handling of failure state for workspaces 

### LINKED ISSUE #10315: [ws-manager] workspaces failed to start

sagor999:
## Description

This PR fixes two issues:
1. As described in https://github.com/gitpod-io/gitpod/issues/10315#issuecomment-1185810931 it resolves an issue of sending sometimes incorrectly to server empty failed state, even though prior to that we might have sent failure error message. This breaks opening that workspace again, as server will incorrectly assume that there was a backup created (no failure error) and will attempt to restore second workspace from backup, which will fail, thus completely breaking opening that workspace for the customer.
2. We incorrectly mark completed container as an error.  When container is stopped, it is will be in a completed state. That is not an error. 
This two fixes need to go hand in hand.
For more info, see the link above for the comment describing the issue in greater detail.

## Related Issue(s)

Fixes #10315

## How to test

Open workspace in preview env
Close it
It should close normally.

## Release Notes

```release-note
none
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/ws-manager/pkg/manager/status.go
+++ b/components/ws-manager/pkg/manager/status.go
@@ -350,11 +350,7 @@ func (m *Manager) extractStatusFromPod(result *api.WorkspaceStatus, wso workspac
 		result.Phase = api.WorkspacePhase_STOPPING
 
 		_, podFailedBeforeBeingStopped := pod.Annotations[workspaceFailedBeforeStoppingAnnotation]
-		if !podFailedBeforeBeingStopped {
-			// While the pod is being deleted we do not care or want to know about any failure state.
-			// If the pod got stopped because it failed we will have sent out a Stopping status with a "failure"
-			result.Conditions.Failed = ""
-		} else {
+		if podFailedBeforeBeingStopped {
 			if _, ok := pod.Annotations[workspaceNeverReadyAnnotation]; ok {
 				// The workspace is never ready, so there is no need for a stopping phase.
 				result.Phase = api.WorkspacePhase_STOPPED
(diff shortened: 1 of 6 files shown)


roboquat:

<details>

</details>

sagor999:
/hold
I would like to get input from @csweichel and @aledbf for this change, to verify that will not break some other assumptions somewhere else inside the gitpod.

components/ws-manager/pkg/manager/status.go
```diff
@@ -591,11 +587,8 @@ func extractFailure(wso workspaceObjects) (string, *api.WorkspacePhase) {
 					return fmt.Sprintf("container %s ran with an error: exit code %d", cs.Name, terminationState.ExitCode), &phase
 				}
 			} else if terminationState.Reason == "Completed" {
-				if wso.IsWorkspaceHeadless() {
-					// default way for headless workspaces to be done
-					return "", nil
-				}
-				return fmt.Sprintf("container %s completed; containers of a workspace pod are not supposed to do that. Reason: %s", cs.Name, terminationState.Message), nil
+				// container terminated successfully - this is not a failure
```
csweichel, line 590: The original intent with this branch was to cover the case that a regular workspace stops without being deleted. An unintentional stop, so to speak. That said, the failure mode is actually nicer if your workspace just stops compared to get getting an inactionable error message ("containers of a workspace pod are not supposed to do that" 🤷).

The `failed` condition on workspace status indeed isn't stable (but should be). ws-manager-bridge should protect us against that by not resetting that condition if it ever was set. Because we have nothing like an at-least-once delivery semantic, we might just miss the "failed" message though.

csweichel (approved):
I think this will do more good than harm.
We could consider adding a metric to count unintentional pod termination - either in this PR (hence the hold), or in a separate one, or not at all :) 

/hold

sagor999:
Added metric to track unintentional stops

PR #11493, opened 2022-07-20
[usage] Request signed upload URL from `content-service`

andrew-farries:
## Description

https://github.com/gitpod-io/gitpod/pull/11474 added a new `UsageReportService` to `content-service` that allowed S3 signed upload URLs to be retrieved for the purpose of uploading usage reports to cloud storage.

This PR hooks up the `usage` component  with the `content-service` so that it requests a signed upload URL from `content-service` each time it runs.

 Subsequent PRs will use this signed URL to actually upload usage reports to cloud storage.

## Related Issue(s)
Part of #9036 
Builds on https://github.com/gitpod-io/gitpod/pull/11474

## How to test

Look at the logs for the `usage` component in the preview environment for this PR. When usage reconciliation runs it successfully requests a signed upload URL from the `content-service`:

image

## Release Notes

```release-note
NONE
```

## Werft options:
- [x] /werft with-preview

### DIFF
--- a/components/usage/pkg/contentservice/client.go
+++ b/components/usage/pkg/contentservice/client.go
@@ -1,0 +1,43 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package contentservice
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/gitpod-io/gitpod/content-service/api"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+type Interface interface {
+	GetSignedUploadUrl(ctx context.Context) (string, error)
+}
+
+type Client struct {
+	url string
+}
+
+func New(url string) *Client {
+	return &Client{url: url}
+}
+
+func (c *Client) GetSignedUploadUrl(ctx context.Context) (string, error) {
+	conn, err := grpc.Dial(c.url, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		return "", fmt.Errorf("failed to dial content-service gRPC server: %w", err)
+	}
+	defer conn.Close()
+
+	uc := api.NewUsageReportServiceClient(conn)
+
+	resp, err := uc.UploadURL(ctx, &api.UsageReportUploadURLRequest{Name: "some-name"})
+	if err != nil {
+		return "", fmt.Errorf("failed to obtain signed upload URL: %w", err)
+	}
+
+	return resp.Url, nil
+}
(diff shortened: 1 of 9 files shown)


components/usage/pkg/contentservice/client.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/gitpod-io/gitpod/content-service/api"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+type Interface interface {
```
easyCZ, line 16: Do we need this wrapper? Wouldn't using the `api.UsageReportServiceClient` directly not be sufficient?

easyCZ: In practice, you set this up once during server start and then inject the client into the controller which means this abstraction doesn't add much.

easyCZ (commented):
I'm surprised a network policy was not needed to talk to the content service

install/installer/pkg/components/usage/configmap.go
```diff
@@ -20,6 +21,7 @@ import (
 func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 	cfg := server.Config{
 		ControllerSchedule: time.Hour.String(),
+		ContentServiceUrl:  fmt.Sprintf("%s:%d", content_service.Component, content_service.RPCPort),
```
easyCZ, line 24: `net.JoinHostPort` instead

install/installer/pkg/components/usage/configmap_test.go
```diff
@@ -23,6 +23,7 @@ func TestConfigMap_ContainsSchedule(t *testing.T) {
 	require.JSONEq(t,
 		`{
        "controllerSchedule": "2m",
+       "contentServiceUrl": "content-service:8080",
```
easyCZ, line 26: Technically, this is not a URL, just an address

andrew-farries: Renamed (b50982f5b)

install/installer/pkg/components/usage/configmap.go
andrew-farries: Thanks. aed2c3fb6

components/usage/pkg/contentservice/client.go
andrew-farries: How would you then handle the no-op case where we don't want to upload to cloud storage when no `contentServiceUrl` is configured?

The reason for this `Interface` was just to be able to do that - have a real implementation and a no-op one.

easyCZ: Ah, good point. Makes perfect sense to have it then.

andrew-farries:
> I'm surprised a network policy was not needed to talk to the content service

Yes, that is surprising. The `content-service` has a network policy that allows ingress from anywhere:

https://github.com/gitpod-io/gitpod/blob/7d886004ee85556058c4d4f805be9b39d8bdffde/install/installer/pkg/components/content-service/networkpolicy.go#L15-L31

This should probably be more restrictive.

easyCZ (approved):

PR #11497, opened 2022-07-20
[usage] Tiny fix to use variable rather than literal value in test

andrew-farries:
## Description
Use variable instead of equivalent constant.

## How to test

unit tests.
## Release Notes

```release-note
NONE
```


geropl (approved):
I can do this! :muscle: :smile:

PR #11502, opened 2022-07-20
[installer]: add custom CA to container runtime if using a self-signed cert

### LINKED ISSUE #11005: Resolve Inter Pod Communication When Using Self-Signed TLS Certificates
### Bug description

When selecting the 'Use Self-Signed TLS Certificates' option on the KOTS configuration page it causes an issue where pods are seemingly unable to communicate between each other. This happens at least between the image building pod and registry pod. Further discussion on this can be seen in [this internal Slack thread](https://gitpod.slack.com/archives/C01KLC56NP7/p1655389575693299).

Relates to gitpod-io/customers#62.

### Steps to reproduce

1. Spin up a Gitpod instance.
2. In the configuration screen within KOTS, under the TLS Certificates section, select option 1 which should be to use a self-signed certificate.
3. Once the instance is configured and Gitpod is running, attempt to spin up a workspace.
4. Observe an error about certificate signing - `x509: certificate signed by unknown authority`. This should appear in the image-builder pod logs.

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

MrSimonEmms:
## Description

This is a fix for the problems with using self-signed certs. The fundamental problem here is that the generated CA cert isn't recognised in the Kubelet (this is also a problem with the Local Preview - as we control everything in that environment, it's easier to solve).

This takes the CA, adds to the `/etc/ssl/certs` directory and runs `update-ca-certificates` command. That adds the CA in. In a managed k8s system, this then will simply involve a restart of the `containerd` service - in k3s, this doesn't exist so will require a full restart of the `k3s` service (this doesn't appear to kill containers though).

The KOTS installation job restarts the `registry-facade` and `ws-daemon` daemonsets, which is required for them to pick up the new certs and allow the `supervisor` process to access them.

### K3s Notes

 - Run the Installer
 - On the node, run `sudo service k3s restart`
 - A redeploy will then need to occur

Now, you'll be able to create an image

## Related Issue(s)

Fixes #11005

## How to test

Deploy via KOTS selecting the self-signed cert option. If using K3s, you will need to manually restart `k3s` and redeploy.

The KOTS channel is `mrsimonemms/resolve-inter-pod-communication-11005` - promote this release into your dev channel so you can test it

## Release Notes

```release-note
[installer]: add custom CA to Kubelet if using a self-signed cert
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/installer/pkg/components/registry-facade/daemonset.go
+++ b/install/installer/pkg/components/registry-facade/daemonset.go
@@ -146,11 +146,14 @@ func daemonset(ctx *common.RenderContext) ([]runtime.Object, error) {
 		return nil, err
 	}
 
+	useHostPid := false
+
 	initContainers := []corev1.Container{
 		*common.InternalCAContainer(ctx),
 	}
 	// Load `customCACert` into Kubelet's only if its self-signed
 	if ctx.Config.CustomCACert != nil && ctx.Config.CustomCACert.Name != "" {
+		useHostPid = true
 		initContainers = append(initContainers,
 			*common.InternalCAContainer(ctx, func(c *corev1.Container) {
 				c.Name = "update-containerd-certificates"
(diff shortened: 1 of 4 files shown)


csweichel:
Restarting containerd as part of an init container does not work reliably, because this might break lifecycle expectations of Kubernetes. Also, this hack is only required in GKE, nowhere else.

I'd strongly vote not to make this default behaviour. It's even questionable if we want to include this kind of behaviour at all - although we might introduce a GKE specific fix after all.

MrSimonEmms:
@csweichel This is the only way I've been able to get self-signed certs working. I'm open to better ideas of course, but I don't have any.

I've successfully tested this in Azure, GKE and K3s so I'm not sure it's GKE-specific

csweichel:
@MrSimonEmms please engage directly with the @gitpod-io/engineering-workspace team here - they should produce and own a fix for this issue.

roshvin:
Hi , i am also facing this issue using slef-signed certificate option in k3s . could you pls suggest a workaround

MrSimonEmms:
@roshvin see #11666

roshvin:
Hi @MrSimonEmms  thanks a lot for your quick response. 
i went through the issue given and coudnt really understood what changes i need to make . my setup is gitpod Airgap with K3s cluster.

could you pls suggest the workaround step to fix this issue if possible. so that i can try those steps in my environment .

MrSimonEmms:
@roshvin it won't work until that PR is done. For details on how to make work with k3s, please see https://github.com/gitpod-io/gitpod/pull/11666#issuecomment-1199214157

PR #11519, opened 2022-07-21
[usage] Upload usage reports to cloud storage

andrew-farries:
## Description

This is the third part of the PR stack:

https://github.com/gitpod-io/gitpod/pull/11474: Add a means of getting signed S3 upload URLs from content-service.
https://github.com/gitpod-io/gitpod/pull/11493: Request the signed upload URLs from the usage component.
This PR: Use the upload URL to upload compressed usage reports to object storage and stop writing the reports to the container filesystem.

## Related Issue(s)
Part of #9036 

## How to test

* Port forward to minio in the preview environment:
```
kubectl port-forward svc/minio 9000:9000
```
* Get the access key and secret key for the preview minio instance:
```
kubectl exec deploy/content-service -- cat /config/config.json
```
* [Download the minio client](https://min.io/download#/kubernetes) `mc`, then add server credentials for the preview minio:
```
mc alias set mm http://localhost:9000
```
* Change the usage component to run reconciliation every 30 seconds or so:
```
kubectl edit cm usage
 # edit the `controllerSchedule` field
```
* Tail the logs of the usage component:
```
kubectl logs -f deploy/usage
```
* See that when the usage component runs reconciliation, the resulting usage report will be uploaded to minio in-cluster storage:
```
mc ls mm/usage-reports
```
* Download and unzip the usage report to see that it has the expected contents:
```
mc cp mm/usage-reports/<some report file> .
gunzip <filename>.gz
cat <filename>
```

## Release Notes
```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/usage/pkg/contentservice/client.go
+++ b/components/usage/pkg/contentservice/client.go
@@ -7,14 +7,17 @@ package contentservice
 import (
 	"context"
 	"fmt"
+	"io"
+	"net/http"
 
+	"github.com/gitpod-io/gitpod/common-go/log"
 	"github.com/gitpod-io/gitpod/content-service/api"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 )
 
 type Interface interface {
-	GetSignedUploadUrl(ctx context.Context) (string, error)
+	UploadFile(ctx context.Context, filename string, body io.Reader) error
 }
 
 type Client struct {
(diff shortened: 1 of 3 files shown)


geropl:
@andrew-farries Blocked by this PR: https://github.com/gitpod-io/gitpod/pull/11596

components/usage/pkg/contentservice/client.go
```diff
@@ -25,7 +28,33 @@ func New(url string) *Client {
...
+	if err != nil {
+		return fmt.Errorf("failed to obtain signed upload URL: %w", err)
+	}
+
+	req, err := http.NewRequest(http.MethodPut, url, body)
+	if err != nil {
+		return fmt.Errorf("failed to construct http request: %w", err)
+	}
+
+	req.Header.Set("Content-Encoding", "gzip")
```
easyCZ, line 42: Does this cause the content to be gzipped on the server? I would've expected you set this to indicate that the content you're uploading is `gzip` but I don't see us compressing in this implementation.

components/usage/pkg/controller/reconciler.go
```diff
@@ -76,36 +75,26 @@ func (u *UsageReconciler) Reconcile() (err error) {
...
-	if err != nil {
-		return fmt.Errorf("failed to create temporary file: %w", err)
-	}
-	defer f.Close()
-
-	enc := json.NewEncoder(f)
-	err = enc.Encode(report)
+	reportBytes := &bytes.Buffer{}
+	gz := gzip.NewWriter(reportBytes)
+	err = json.NewEncoder(gz).Encode(report)
```
andrew-farries, line 80: This is where the json report is gzipped.

components/usage/pkg/contentservice/client.go
andrew-farries: See t[his comment](https://github.com/gitpod-io/gitpod/pull/11519#discussion_r928709112).

components/usage/pkg/controller/reconciler.go
easyCZ: I'd recommend moving this into the Storage Client. As it stands it's confusing because the client sets the content type gzipped, but nothing in the API signature indicates it should be gzipped ahead of it. 

Ideally, the Client takes a bytes buffer and wraps it in a gzip writer, then writes the whole thing as a stream

easyCZ: Or even can take an `io.Writer` which is a much better interface for wrapping

andrew-farries: The `UploadFile` method really should only take usage reports - this object storage signed upload URL that it obtains will place the file in the `usage-report` bucket. 

As such, I've renamed the method on the interface and changed its argument to be a report rather than an `io.Reader` (aa9f2d4ca4686a4fef89dd43f1aaaf8a160eecb3)

easyCZ (approved):

PR #11522, opened 2022-07-21
Move usageAttributionId

AlexTugarev:
## Description
This PR implements validations for `setUsageAttribution` in `UserServer`.
Also `usageAttributionId` is moved to be a field of `User` entity.

A follow-up to https://github.com/gitpod-io/gitpod/pull/10893

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/gitpod-protocol/src/protocol.ts
+++ b/components/gitpod-protocol/src/protocol.ts
@@ -45,6 +45,9 @@ export interface User {
     markedDeleted?: boolean;
 
     additionalData?: AdditionalUserData;
+
+    // Identifies an explicit team or user ID to which all the user's workspace usage should be attributed to (e.g. for billing purposes)
+    usageAttributionId?: string;
 }
 
 export namespace User {
@@ -199,8 +202,6 @@ export interface AdditionalUserData {
     knownGitHubOrgs?: string[];
     // Git clone URL pointing to the user's dotfile repo
     dotfileRepo?: string;
-    // Identifies an explicit team or user ID to which all the user's workspace usage should be attributed to (e.g. for billing purposes)
-    usageAttributionId?: string;
     // preferred workspace classes
     workspaceClasses?: WorkspaceClasses;
     // additional user profile data
--- a/components/server/ee/src/workspace/gitpod-server-impl.ts
+++ b/components/server/ee/src/workspace/gitpod-server-impl.ts
@@ -1497,16 +1497,14 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
             this.stripeService.findCustomerByTeamId(teamId),
             this.userDB.findUserById(userId),
         ]);
-        if (teamCustomer && user && !user.additionalData?.usageAttributionId) {
+        if (teamCustomer && user && !user.usageAttributionId) {
             // If the user didn't explicitly choose yet where their usage should be attributed to, and
             // they join a team which accepts usage attribution (i.e. with usage-based billing enabled),
             // then we simplify the UX by automatically attributing the user's usage to that team.
             // Note: This default choice can be changed at any time by the user in their billing settings.
             const subscription = await this.stripeService.findUncancelledSubscriptionByCustomer(teamCustomer.id);
             if (subscription) {
-                user.additionalData = user.additionalData || {};
-                user.additionalData.usageAttributionId = `team:${teamId}`;
-                await this.userDB.updateUserPartial(user);
+                await this.userService.setUsageAttribution(user, AttributionId.render({ kind: "team", teamId }));
             }
         }
     }
@@ -1519,12 +1517,11 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
             await this.teamSubscription2Service.cancelTeamMemberSubscription(ts2, userId, teamMembershipId, now);
         }
         const user = await this.userDB.findUserById(userId);
-        if (user && user.additionalData?.usageAttributionId === `team:${teamId}`) {
+        if (user && user.usageAttributionId === AttributionId.render({ kind: "team", teamId })) {
             // If the user previously attributed all their usage to a given team, but they are now leaving this
             // team, then the currently selected usage attribution ID is no longer valid. In this case, we must
             // reset this ID to the default value.
-            user.additionalData.usageAttributionId = undefined;
-            await this.userDB.updateUserPartial(user);
+            await this.userService.setUsageAttribution(user, undefined);
         }
     }
 
@@ -2027,10 +2024,8 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
             await Promise.all(
                 members.map(async (m) => {
                     const u = await this.userDB.findUserById(m.userId);
-                    if (u && !u.additionalData?.usageAttributionId) {
-                        u.additionalData = u.additionalData || {};
-                        u.additionalData.usageAttributionId = `team:${teamId}`;
-                        await this.userDB.updateUserPartial(u);
+                    if (u && !u.usageAttributionId) {
+                        await this.userService.setUsageAttribution(u, `team:${teamId}`);
                     }
                 }),
             );
--- a/components/server/src/user/user-service.ts
+++ b/components/server/src/user/user-service.ts
@@ -16,7 +16,7 @@ import {
     WORKSPACE_TIMEOUT_EXTENDED,
     WORKSPACE_TIMEOUT_EXTENDED_ALT,
 } from "@gitpod/gitpod-protocol";
-import { CostCenterDB, ProjectDB, TermsAcceptanceDB, UserDB } from "@gitpod/gitpod-db/lib";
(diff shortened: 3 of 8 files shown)


roboquat:

<details>

</details>

components/gitpod-protocol/src/protocol.ts
```diff
@@ -200,7 +203,7 @@ export interface AdditionalUserData {
     // Git clone URL pointing to the user's dotfile repo
     dotfileRepo?: string;
     // Identifies an explicit team or user ID to which all the user's workspace usage should be attributed to (e.g. for billing purposes)
-    usageAttributionId?: string;
+    // usageAttributionId?: string;
```
AlexTugarev, line 206: left over

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -3195,6 +3195,10 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
         throw new ResponseError(ErrorCodes.SAAS_FEATURE, `Not implemented in this version`);
     }
 
+    async setUsageAttribution(ctx: TraceContext, usageAttributionId: string): Promise<void> {
+        throw new ResponseError(ErrorCodes.SAAS_FEATURE, `Not implemented in this version`);
```
easyCZ, line 3199: Why is this not implemented in non SaaS?

components/server/src/user/user-service.ts
```diff
@@ -248,6 +249,36 @@ export class UserService {
         return AttributionId.render({ kind: "team", teamId: project.teamId });
     }
 
+    async setUsageAttribution(user: User, usageAttributionId: string | undefined): Promise<void> {
```
AlexTugarev, line 252: validation in here are the main motivator to externalize `usageAttributionId` from the unchecked `additionalData`.

components/server/src/workspace/gitpod-server-impl.ts
AlexTugarev: right, that's mistake. it should be the default.

AlexTugarev: check

components/gitpod-protocol/src/protocol.ts
```diff
@@ -199,8 +202,6 @@ export interface AdditionalUserData {
     knownGitHubOrgs?: string[];
     // Git clone URL pointing to the user's dotfile repo
     dotfileRepo?: string;
-    // Identifies an explicit team or user ID to which all the user's workspace usage should be attributed to (e.g. for billing purposes)
-    usageAttributionId?: string;
```
easyCZ, line 203: What will happen with existing records which do have this field set on the User?

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -1519,12 +1517,11 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
             await this.teamSubscription2Service.cancelTeamMemberSubscription(ts2, userId, teamMembershipId, now);
         }
         const user = await this.userDB.findUserById(userId);
-        if (user && user.additionalData?.usageAttributionId === `team:${teamId}`) {
+        if (user && user.usageAttributionId === `team:${teamId}`) {
```
easyCZ, line 1520: There's helper logic to construct these IDs [here](https://github.com/gitpod-io/gitpod/blob/896f4b0f3e66aa053603c457bf82193cf833164e/components/server/src/user/user-service.ts#L247)

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -3195,6 +3195,11 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
         throw new ResponseError(ErrorCodes.SAAS_FEATURE, `Not implemented in this version`);
     }
 
+    async setUsageAttribution(ctx: TraceContext, usageAttributionId: string): Promise<void> {
```
easyCZ, line 3198: How do we roll out this API change? Currently Usage Based Pricing is enabled behind a feature flag.

components/server/src/user/user-service.ts
```diff
@@ -248,6 +249,36 @@ export class UserService {
...
+            return;
+        }
+        const attributionId = AttributionId.parse(usageAttributionId);
+        if (attributionId?.kind === "user") {
+            if (user.id === attributionId.userId) {
+                user.usageAttributionId = usageAttributionId;
+                await this.userDb.storeUser(user);
+                return;
+            }
+            throw new Error("Cannot attribute another user.");
```
easyCZ, line 265: Please classify the error as Permission Denied

components/server/src/user/user-service.ts
```diff
@@ -248,6 +249,36 @@ export class UserService {
...
+        if (attributionId?.kind === "team") {
+            const membership = await this.teamDB.findTeamMembership(user.id, attributionId.teamId);
+            if (!membership) {
+                throw new Error("Cannot attribute an unrelated team.");
+            }
+            user.usageAttributionId = usageAttributionId;
+            await this.userDb.storeUser(user);
+            return;
+        }
+        console.warn("Unexpected call arguments for UserService.setUsageAttribution", {
```
easyCZ, line 276: Here, we should throw BadRequest error

components/server/src/user/user-service.ts
easyCZ, line 252: Here we allow to set it to any team (or user) you're a member of. It doesn't matter if the team is a paying team or not. Is this the desired behaviour?

components/gitpod-protocol/src/protocol.ts
AlexTugarev: Briefly talked to @geropl about this. Given the small impact in terms of user already using this, a data migration seems to be not necessary. Please correct me, if this assumption is wrong, or if you see real problems on production.

components/server/src/user/user-service.ts
AlexTugarev: Agree.

components/server/src/user/user-service.ts
AlexTugarev: Let me check if there is a simple way to compute that state.

components/gitpod-protocol/src/protocol.ts
easyCZ: That's okay. As long as we won't fail to parse the JSON stored in `additionalData` of the user record then this seems okay.

components/server/src/workspace/gitpod-server-impl.ts
AlexTugarev: This PR adds a replacement to an unchecked API call, providing more validation. Do you see any value in pushing it behind the feature flag?  e.g. where can it break?

easyCZ: The previous way to set the attribution wasn't actually exposed to users in any way, it had to be done through a DB update manually. The difference here is that we're adding an API which is now accessible by users.

If we choose to not gate it behind a feature flag, users will already be able to set their attribution ID to something. That may not necessarily be a bad thing, but it can cause confusion with respect to how Chargebee plans will work with respect to Usage Based Pricing.

My recommendation would be that for now, we gate it behind the `isUsageBasedBillingEnabled` flag, like [here](https://github.com/gitpod-io/gitpod/blob/2946bee541dc1594bc8e80ed1b02f2e9f93eb4d0/components/server/ee/src/workspace/gitpod-server-impl.ts#L1955).

AlexTugarev: > The previous way to set the attribution wasn't actually exposed to users in any way, it had to be done through a DB update manually. The difference here is that we're adding an API which is now accessible by users.

This is not true. `User.additionalData` is open for every user. You may put in there what ever you want.

AlexTugarev: > users will already be able to set their attribution ID to something

Exactly this is what this PR is about. The value should be validated.

components/server/src/user/user-service.ts
AlexTugarev: checking for `stripeService.findCustomerByTeamId(attributionId.teamId)` now

easyCZ: So the usage attribution is set on the non-EE version. That means that we can't assume there will be a subscription. I'm actually okay with the version where we don't check for now. We can always add.

components/server/src/workspace/gitpod-server-impl.ts
easyCZ: > This is not true. User.additionalData is open for every user. You may put in there what ever you want

How does a user (either from dashboard, or the API) update this value? I wasn't aware of this being directly exposed to users.

AlexTugarev: > How does a user (either from dashboard, or the API) update this value?
Yes, this used to hold user preferences. I believe this data should only affect this user's resources, which is not true for usage/billing related resources.

easyCZ (approved):

PR #11524, opened 2022-07-21
[installer]: add secret template to certs

MrSimonEmms:
## Description

The support bundle secret collector looks for a label `app=gitpod`. As of v1.5, cert-manager supports a `secretTemplate` option on the certs which adds a label and/or annotation to the secrets - **NB** I haven't updated to latest version of cert-manager as that requires a change of name and lots of updates to packages, including a Helm one that would require renaming the internal charts to lower case values

This upgrades cert-manager to latest (included renaming the package), adds the labels to the secret and configures the `secretTemplate`. As this is still cert-manager v1, this does not seems to introduce any regressions and have tested both upgrades and fresh installs.

The reason for this change is because I was helping a Discord user debug a cert issue and had no way of knowing if the cert had been created from the support bundle. This solves this issue by included the secret metadata in the support bundle

## How to test

Install via KOTS or Installer and check for `app: gitpod` in the secret's labels

## Release Notes

```release-note
[installer]: add secret template to certs
```

## Werft options:

- [x] /werft with-preview


Pothulapati (approved):
LGTM! Like we discussed, We should be showcasing the cert-manager requirement `>=1.5`.

Furisto (approved):

Furisto:
Tested that labels have been applied to the certificates and I do not see any adverse effects this would have on workspace components.

PR #11536, opened 2022-07-21
[terraform] update the disk size in eks module

### LINKED ISSUE #11533: Sync AWS `terraform` module with reference architecture
TODOs:
- [ ] Update EKS terraform module to use s3 access policy in EKS reference architecture
- [ ] Update the security group permissions for EKS module
- [x] Update the disk size for the EKS module PR open here https://github.com/gitpod-io/gitpod/pull/11536

nandajavarma:
## Description

Currently, when using the EKS module, the `disk_size` field set doesn't change the value. This is because templates are getting created for both the node pools. To change the disk size in the template as well, we define `block_device_mappings` for both the managed node pools

## Related Issue(s)

Partly fixes #11533 (not to be closed)

## How to test

Documentation task is still TODO for terraform modules.

## Release Notes

```release-note
NONE
```

## Werft options:

- [X] /werft with-preview

### DIFF
--- a/install/infra/terraform/eks/kubernetes.tf
+++ b/install/infra/terraform/eks/kubernetes.tf
@@ -54,7 +54,6 @@ module "eks" {
   }
 
   eks_managed_node_group_defaults = {
-    disk_size                  = 150
     ami_type                   = "CUSTOM"
     iam_role_attach_cni_policy = true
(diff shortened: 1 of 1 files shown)


adrienthebo (approved):
LGTM!

PR #11546, opened 2022-07-21
[installer] Disable integration tests during upgrade tests

nandajavarma:
## Description

This PR temporarily disables integration tests during upgrade test pipeline. Since the main point of running these tests is to get a self-hosted preview of the upgraded version of Gitpod during release test, having the installation messed up due to integration tests cause problems. This PR disables the tests from this phase, assuming the crons run tests every night anyway we don't necessarily need this.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [X] /werft with-preview


adrienthebo (approved):
:shipit: 

Is there value in having a comment here to indicate why we're disabling tests, or is this likely to be obvious?

Issue #10531, opened 2022-06-08
ws-manager: ensure it can reconcile orphaned PVCs without VolumeSnapshot is created


jenting:
## Is your feature request related to a problem? Please describe

Find a way to make the orphaned PVC happen without the corresponding VolumeSnapshot object.
For example, will the following scenario trigger the orphaned PVC to happen without the corresponding VolumeSnapshot object?
- node crashes
- workspace pod dies
- ws-manager pod dies
- etc

## Describe the behaviour you'd like

- Find a way to make the orphaned PVC happen, and without the corresponding VolumeSnapshot object.
- If yes, make sure the ws-manager can reconcile the orphaned PVC, creating the VolumeSnapshot object and making sure the VolumeSnapshot object is ready to use, and remove the PVC after that.

## Describe alternatives you've considered

If we can't find a way to make the orphaned PVC happen, this issue is not needed.

## Additional context

The workspace Pod has the finalizer to protect the pod itself. The finalizer is added when the workspace pod enters the Running state. https://github.com/gitpod-io/gitpod/blob/e0579ef9388cee0c4c7a2c85588683167e0cb0f7/components/ws-manager/pkg/manager/monitor.go#L407

The finalizer is removed either
- the pod is never ready https://github.com/gitpod-io/gitpod/blob/e0579ef9388cee0c4c7a2c85588683167e0cb0f7/components/ws-manager/pkg/manager/monitor.go#L496
- the pod is stopped https://github.com/gitpod-io/gitpod/blob/e0579ef9388cee0c4c7a2c85588683167e0cb0f7/components/ws-manager/pkg/manager/monitor.go#L506

The PVC has a finalizer to protect it, the PVC would be removed only when the corresponding Pod is gone.
Therefore, in general, the Pod would be removed first and then the PVC.

So, the orphaned PVC would happen _only when_ the workspace Pod is removed (which means the workspace pod finalizer is removed before the VolumeSnapshot object is created).

jenting:
@kylos101 could you please describe some scenarios that the PVC could be orphaned? Thank you.

kylos101:
Hi @jenting, I defer to @sagor999 on helping build a list of scenarios (and which ones are deemed edge cases). 

He shared some thoughts, and you did too, perhaps you two can sync up? Another option might be for you to make a brainstorm/picture to 'see' the use cases, or a table of use cases (ranking most likely to least likely).

Some considerations for this exercise:
[...]

PR #11553, opened 2022-07-22
Delete PVC once the VolumeSnapshot is ready to use

jenting:
## Description

Directly delete the workspace Pod and remove the finalizer before the VolumeSnapshot is ready, the VolumeSnapshot is created, and the workspace Pod is removed as well. However, the PVC does not be removed after that.

## Related Issue(s)

Partially #10531

## How to test

1. Create a workspace-preview environment
2. Enable feature flag `persistent_volume_claim`.
3. Launch a workspace.
4. Manually delete the workspace pod `kubectl delete pod <pod-name>`.
5. Before the VolumeSnapshot is ready to use, remove the workspace pod finalizer `kubectl patch pod <pod-name> -p '{"metadata":{"finalizers":null}}' --type=merge`.

## Release Notes

```release-note
None
```

## Documentation

None

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/ws-manager/pkg/manager/annotations.go
+++ b/components/ws-manager/pkg/manager/annotations.go
@@ -108,6 +108,9 @@ func (m *Manager) markWorkspace(ctx context.Context, workspaceID string, annotat
 	err := retry.RetryOnConflict(backoff, func() error {
 		pod, err := m.findWorkspacePod(ctx, workspaceID)
 		if err != nil {
+			if isKubernetesObjNotFoundError(err) {
+				return nil
+			}
 			return fmt.Errorf("cannot find workspace %s: %w", workspaceID, err)
 		}
 		if pod == nil {
--- a/components/ws-manager/pkg/manager/annotations.go
+++ b/components/ws-manager/pkg/manager/annotations.go
@@ -111,7 +111,7 @@ func (m *Manager) markWorkspace(ctx context.Context, workspaceID string, annotat
 			if isKubernetesObjNotFoundError(err) {
 				return nil
 			}
-			return fmt.Errorf("cannot find workspace %s: %w", workspaceID, err)
+			return fmt.Errorf("cannot mark workspace %s: %w", workspaceID, err)
 		}
 		if pod == nil {
 			return fmt.Errorf("workspace %s does not exist", workspaceID)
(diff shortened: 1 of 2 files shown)


components/ws-manager/pkg/manager/annotations.go
```diff
@@ -108,6 +108,9 @@ func (m *Manager) markWorkspace(ctx context.Context, workspaceID string, annotat
 	err := retry.RetryOnConflict(backoff, func() error {
 		pod, err := m.findWorkspacePod(ctx, workspaceID)
 		if err != nil {
+			if isKubernetesObjNotFoundError(err) {
+				return nil
+			}
 			return fmt.Errorf("cannot find workspace %s: %w", workspaceID, err)
```
sagor999, line 114: if we filter not found error above, then error message below no longer makes sense.
```suggestion
			return fmt.Errorf("cannot mark workspace %s: %w", workspaceID, err)
```

components/ws-manager/pkg/manager/monitor.go
```diff
@@ -183,19 +183,20 @@ func (m *Monitor) onVolumesnapshotEvent(evt watch.Event) error {
 
 	// get the pod resource
 	var pod corev1.Pod
-	if err := m.manager.Clientset.Get(context.Background(), types.NamespacedName{Namespace: vs.Namespace, Name: podName}, &pod); err != nil {
-		log.WithError(err).Warnf("the pod %s/%s is missing", podName, vs.Namespace)
-		return nil
-	}
+	_ = m.manager.Clientset.Get(context.Background(), types.NamespacedName{Namespace: vs.Namespace, Name: podName}, &pod)
```
sagor999, line 186: hm. should we at least log any other error that might happen here except for not found error?

sagor999 (approved):
/hold
for minor nits.
otherwise LGTM

components/ws-manager/pkg/manager/annotations.go
jenting: fixed

components/ws-manager/pkg/manager/monitor.go
jenting: fixed

PR #11563, opened 2022-07-22
Retry the finalizer loop when the ws-daemon pod not found

### LINKED ISSUE #10682: [ws-manager] "last backup failed: no running ws-daemon pod found."
### Bug description

A user lost his changes of a session and the status of the instance in the DB shows the error:
```
"last backup failed: no running ws-daemon pod found.
```

This seems to be a single case as I couldn't find other instances in that time with a similar error message.

### Workspace affected

instanceid: e3c9d5ff-4492-464e-96fc-2cba13e31d45

[Logs for the last seven days](https://cloudlogging.app.goo.gl/KVDSJu1L5MRBRfZz5) (internal link)

jenting:
## Description

Retry the function `doFinalize` when the ws-daemon pod is not found so that we can reach the condition https://github.com/gitpod-io/gitpod/blob/444ccf07dd7ac17ea244f1bc1944d27bfbdaef95/components/ws-manager/pkg/manager/monitor.go#L1272-L1278

## Related Issue(s)

Fixes #10682

## How to test

1. Create a workspace.
2. Create a large file.
   ```console
    dd if=/dev/urandom of=4G bs=16M count=256
   ```
4. Stop the workspace `Gitpod: stop workspace` and immediately go to step 5.
5. During the workspace stopping, keep deleting the ws-daemon pod for at least 1 minute. Note, we probably would hit the issue #11606, but we'd like to tackle it in separate PR. 
   ```console
   while true; do kubectl delete pod -l component=ws-daemon; done
   ```
6. Cancel step 5 (ctrl+c).
7. Check the workspace changes to the stopped state and the backup content persistent.

## Release Notes

```release-note
None
```

## Documentation

None

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/ws-manager/pkg/manager/monitor.go
+++ b/components/ws-manager/pkg/manager/monitor.go
@@ -19,6 +19,7 @@ import (
 	tracelog "github.com/opentracing/opentracing-go/log"
 	"golang.org/x/xerrors"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	grpc_status "google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
 
@@ -53,7 +54,7 @@ const (
 
 var (
 	// wsdaemonMaxAttempts is the number of times we'll attempt to work with ws-daemon when a former attempt returned unavailable.
-	// We rety for two minutes every 5 seconds (see wwsdaemonRetryInterval).
+	// We rety for two minutes every 5 seconds (see wsdaemonRetryInterval).
 	//
 	// Note: this is a variable rather than a constant so that tests can modify this value.
 	wsdaemonMaxAttempts = 120 / 5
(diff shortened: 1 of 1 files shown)


components/ws-manager/pkg/manager/monitor.go
```diff
@@ -1242,7 +1243,11 @@ func (m *Monitor) finalizeWorkspaceContent(ctx context.Context, wso *workspaceOb
 		gitStatus   *csapi.GitStatus
 	)
 	t := time.Now()
-	for i := 0; i < wsdaemonMaxAttempts; i++ {
+
+	// since we're within the content backup loop, the maximum retry time should
+	// rely on the content finalization configuration
+	maxRetry := int(m.manager.Config.Timeouts.ContentFinalization / util.Duration(wsdaemonRetryInterval))
```
utam0k, line 1249: If time is taken for a process other than `time.Sleep`, will the timeout time(`m.manager.Config.Timeouts.ContentFinalization`) be exceeded?

utam0k: Can we use this?
https://gobyexample.com/timeouts

jenting: The `wsdaemonMaxAttempts` is an integer that was a value of 24 (120 seconds / 5 seconds), which means its maximum retry is 24 times. Here, I'd like to address a case where the ws-daemon was not reachable (Unavailable) for over 2 minutes.

What you mentioned is another implementation time wait way, but I don't want to bring the change in this PR.

utam0k: I think your implementation is a great improvement even now. My opinion was a bit of an edge case. I will mark the resolution.

utam0k (approved):

PR #11568, opened 2022-07-22
[installer-tests] Support tests against multiple version of k8s and ubuntu images

### LINKED ISSUE #11272: Customise operating system images - test on specific ubuntu version

### LINKED ISSUE #11349: Run automated self-hosted tests on all supported Kubernetes versions
Expand our test suite to run on all supported Kubernetes versions as defined in our [documentation](https://www.gitpod.io/docs/self-hosted/latest/cluster-set-up#supported-kubernetes-distributions).

nandajavarma:
## Description

This PR parameterize  the version of node image when possible and version of kubernetes. The node images are not configurable easily on managed self-hosted instances. We can only use the supported images corresponding to the Kubernetes version we are using. Hence the all supported ubuntu versions are tested only in the k3s setup.

## Related Issue(s)

Fixes #11349 
Fixes #11272 

## How to test

One can use any of the suggested commands in this [internal document](https://www.notion.so/gitpod/Self-hosted-automated-nightly-test-pipelines-fdc5a202e67b4197aa00d93b98d57927) to test.

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/infra/terraform/aks/kubernetes.tf
+++ b/install/infra/terraform/aks/kubernetes.tf
@@ -1,8 +1,3 @@
-data "azurerm_kubernetes_service_versions" "k8s" {
-  location        = azurerm_resource_group.gitpod.location
-  include_preview = false
-}
-
 resource "azurerm_role_assignment" "k8s" {
   count = var.dns_enabled ? 1 : 0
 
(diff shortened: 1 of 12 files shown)


install/infra/terraform/aks/kubernetes.tf
```diff
@@ -20,37 +15,37 @@ resource "azurerm_role_assignment" "k8s_reader" {
...
-
-  kubernetes_version               = data.azurerm_kubernetes_service_versions.k8s.latest_version
+  name                            = format(local.name_format, local.location, "primary")
+  location                        = azurerm_resource_group.gitpod.location
+  resource_group_name             = azurerm_resource_group.gitpod.name
+  dns_prefix                      = "gitpod"
+  tags                            = {}
+  api_server_authorized_ip_ranges = []
+
+  kubernetes_version               = var.cluster_version
```
MrSimonEmms, line 25: I have a feeling that this may break in time. Azure tend to remove versions of k8s at fairly short notice, which is why the `data "azurerm_kubernetes_service_versions"` is the recommended way of getting the versions. You can use a [version_prefix](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/data-sources/kubernetes_service_versions#version_prefix) as well

My suggestion is to keep this change as you've done it, but do so in the knowledge that we may have to change it in the future.

MrSimonEmms (approved):

PR #11577, opened 2022-07-22
[usage] By default sort results descending

easyCZ:
## Description

Adds default ordering to billing records

## Related Issue(s)

Fixes #

## How to test

Unit tests

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage/pkg/apiv1/usage_test.go
+++ b/components/usage/pkg/apiv1/usage_test.go
@@ -85,6 +85,9 @@ func TestUsageService_ListBilledUsage(t *testing.T) {
 				instanceIDs = append(instanceIDs, instance.InstanceID.String())
 			}
 
+			matchingIDs := instanceIDs[0:5]
+			reverse(matchingIDs)
+
 			return Scenario{
 				name:      "filters results to specified time range",
 				Instances: instances,
@@ -95,7 +98,7 @@ func TestUsageService_ListBilledUsage(t *testing.T) {
 				},
 				Expect: Expectation{
 					Code:        codes.OK,
-					InstanceIds: instanceIDs[0:5],
+					InstanceIds: matchingIDs,
 				},
 			}
 		})(),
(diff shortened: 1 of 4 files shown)


components/usage/pkg/db/workspace_instance_usage.go
```diff
@@ -49,12 +49,28 @@ func CreateUsageRecords(ctx context.Context, conn *gorm.DB, records []WorkspaceI
 	return db.CreateInBatches(records, 1000).Error
 }
 
-func ListUsage(ctx context.Context, conn *gorm.DB, attributionId AttributionID, from, to time.Time) ([]WorkspaceInstanceUsage, error) {
+type Order int
```
andrew-farries, line 52: I think this type doesn't need to be exported from the module, ie change to `order`.

Also, I'd move the type definition and consts up above the `func`s. Not sure if the style guide has anything to say about this. Feel free to ignore.

easyCZ: It needs to be set by the api package when invoking it (at least it will need to when the order is specified in the request arguments)

andrew-farries: ok, so keep it exported because it will need to be for a later PR?

easyCZ: You're correct that in this PR it doesn't need to be public, as it could move entirely into the ListUsage function. But because the direction is to expose in the `api` layer, it needs to be settable. 

But I think I just now understood what you mean, the type can be private, but the Consts can be public.

andrew-farries: > But I think I just now understood what you mean, the type can be private, but the Consts can be public.

💯

easyCZ: Done.

components/usage/pkg/apiv1/usage_test.go
```diff
@@ -134,3 +137,9 @@ func TestUsageService_ListBilledUsage(t *testing.T) {
 		})
 	}
 }
+
+func reverse[S ~[]E, E any](s S) {
+	for i, j := 0, len(s)-1; i < j; i, j = i+1, j-1 {
+		s[i], s[j] = s[j], s[i]
+	}
+}
```
andrew-farries, line 145: hopefully this kind of thing makes it into the standard library soon.

andrew-farries (approved):

PR #11582, opened 2022-07-22
[usage] Fix usage query

easyCZ:
## Description

Before, the extra range filters would eventually be appended as top level OR, which would lead to returning results which didn't actually match the Attribution ID

## Related Issue(s)

Fixes #

## How to test

Unit test
Test as a regression, remove the changes to the implementation and run the test

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


geropl (approved):
:bulb: Nice find!

andrew-farries (approved):

PR #11601, opened 2022-07-25
Introduce BillingModes

geropl:
### ToDo
 1. Use getBillingModeUser/Team to show/hide controls/UI (Ideally replace getShowPaymentUI!)
    - block all Chargebee-based UI once switched! (except: old invoices: GitpodService.isChargebeeCustomer should help here)
 2. Block Chargebee API
    - block all Chargebee-based API once switched! (except: old invoices. GitpodService.isChargebeeCustomer should help here)
 3. Implement (and decide details of) [EntitlementServiceUBB](https://github.com/gitpod-io/gitpod/blob/080d381dd01961022363c25e5cf917c3cc160f24/components/server/ee/src/billing/entintlement-service-ubb.ts#L22-L44)

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


components/server/ee/src/billing/entitlement-service.ts
```diff
@@ -0,0 +1,90 @@
...
+    @inject(UserDB) protected readonly userDb: UserDB;
+
+    async mayStartWorkspace(
+        user: User,
+        date: Date,
+        runningInstances: Promise<WorkspaceInstance[]>,
+    ): Promise<MayStartWorkspaceResult> {
+        const billingMode = await this.billingModes.getBillingModeForUser(user);
+        switch (billingMode.mode) {
+            case "none":
```
easyCZ, line 39: For my own (and future reviewers) sanity, The mode of `none` only gets set for `Self-hosted`, bacause it depends on the `paymentEnabled` config to be false.

components/gitpod-protocol/src/accounting-protocol.ts
```diff
@@ -199,6 +199,9 @@ export namespace Subscription {
     export function isActive(s: Subscription, date: string): boolean {
         return s.startDate <= date && (s.endDate === undefined || date < s.endDate);
     }
+    export function isCancelled(s: Subscription, date: string): boolean {
+        return s.cancellationDate !== undefined && s.cancellationDate < date;
```
jankeromnes, line 203: Nit: It would seem safer to check for anything "falsy" (for example, empty string `""` or `null`) rather than just specifically `undefined` here:

```suggestion
        return !!s.cancellationDate && s.cancellationDate < date;
```

components/server/ee/src/billing/billing-mode.ts
```diff
@@ -0,0 +1,121 @@
...
+        const now = new Date();
+
+        // Is Usage Based Billing enabled for this user or not?
+        const isUsageBasedBillingEnabled = await getExperimentsClientForBackend().getValueAsync(
+            "isUsageBasedBillingEnabled",
+            false,
+            {
+                user,
+            },
+        );
```
jankeromnes, line 39: Wait, isn't UBB flag controlled by whether you're part of the Gitpod team or not? In this case I believe you may need to also pass along the `teams` of the user.

components/server/ee/src/billing/billing-mode.ts
```diff
@@ -0,0 +1,121 @@
...
+                return { mode: "chargebee", tier: "paid_cancelled_and_ubb" };
+            }
+
+            // The user has at least one paid subscription
+            return { mode: "chargebee", tier: "paid", planId: cbSubscriptions[0].planId! };
+        }
+
+        // UBB enabled?
+        if (!isUsageBasedBillingEnabled) {
+            // No Chargebee subscription, and UBB is not enabled: CB free tier?
```
jankeromnes, line 58: Why the question mark? 🤔

components/server/ee/src/billing/billing-mode.ts
```diff
@@ -0,0 +1,121 @@
...
+
+        // Is Usage Based Billing enabled for this team?
+        const isUsageBasedBillingEnabled = await getExperimentsClientForBackend().getValueAsync(
+            "isUsageBasedBillingEnabled",
+            false,
+            {
+                teamId: team.id,
+                teamName: team.name,
+            },
+        );
```
jankeromnes, line 90: If I'm not mistaken, we don't yet have anything like this in ConfigCat -- we only enable UBB for users who happen to be part of the Gitpod team, and then everywhere in the UI (i.e. for all their teams).

But maybe it could be a good idea to selectively enable UBB for a specific team, and only in that team's UI 👍

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -1919,24 +1919,34 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
+            // Stripe is allowed when you either are on the usage-based side already, or you can switch to)
+            (m) => m.mode === "usage-based" || (m.mode === "chargebee" && m.tier === "paid_cancelled_and_ubb"),
         );
-        if (!isUsageBasedBillingEnabled) {
-            throw new ResponseError(ErrorCodes.PERMISSION_DENIED, "not allowed");
+    }
+
+    protected async ensureBillingMode(
+        sub: { user?: User; team?: Team },
+        pred: (m: BillingMode) => boolean,
```
jankeromnes, line 1934: What does `pred` mean? 👀

components/gitpod-protocol/src/accounting-protocol.ts
jankeromnes: Also, can there be any cases where you don't have a `cancellationDate` but you do have an `endDate`? 👀

jankeromnes:
Many thanks for implementing this!

This change makes me very nervous, because we're introducing so much complexity:
- Complicated billing modes with both a "mode" and a "tier" (would ideally prefer just one parameter / moving part)
- Feature flags, which get queried in multiple different ways throughout the code (with/without `teams`, with/without `teamId` and `teamName`, different APIs in client and server, etc) -- I'm keenly looking forward to deleting all the inconsistent and risky feature-flag logic once UBB is shipped and everyone uses it (</wishful-thinking>)

However, we are in a complicated situation (with the transition from Chargebee to UBB with many moving parts), and your solution looks like it can work. So, many thanks for powering through this. 💯

I would love to test this somehow. Could you please get this into a testable state (and complete the "TODOs"), and then provide "How to test" instructions?

components/server/ee/src/billing/billing-mode.ts
geropl: > But maybe it could be a good idea to selectively enable UBB for a specific team, and only in that team's UI +1 we'd just need to code this into ConfigCat.

That was the intend. `teamId` and `team_name` are already available in ConfigCat. :+1:

components/server/ee/src/workspace/gitpod-server-impl.ts
geropl: Predicate. Sorry, was a tad late when I wrote this :stuck_out_tongue: Will improve.

geropl:
> However, we are in a complicated situation (with the transition from Chargebee to UBB with many moving parts), and your solution looks like it can work. So, many thanks for powering through this. 100

:100: 

> I would love to test this somehow. Could you please get this into a testable state (and complete the "TODOs"), and then provide "How to test" instructions?

It might also make sense to introduce this in steps. Especially the parts that affect current behaviors should be rolled out selectively so we do not break current users! Happy to brainstorm on this.

components/gitpod-protocol/src/accounting-protocol.ts
geropl: Yeah, both valid points. It _should_ be undefined, and it _should_ always have a `cancellationDate`. But that's one of the tables that saw a lot of manual "fixes", so worth being more conservative here! :100:

components/server/ee/src/billing/billing-mode.ts
geropl: I think it make sense for the two of us to discuss the decision tree/scope of the decision made here. :+1:

components/server/ee/src/billing/billing-mode.ts
geropl: Yes, that's up to discussion. See [comment](https://github.com/gitpod-io/gitpod/pull/11601#discussion_r934295916) below.

components/gitpod-protocol/src/accounting-protocol.ts
geropl: Checked the prod DB: we actually have 6 cases where this is not set :see_no_evil: Made the code fail-safe to save us our :brain:s.

geropl:
Closing this one as it's replaced by https://github.com/gitpod-io/gitpod/pull/11812 and it's successors.

Issue #11526, opened 2022-07-21
Allow users to switch between months and update usage page sidebar designs


gtsiolis:
### Problem to solve

Following up from https://github.com/gitpod-io/gitpod/pull/11481, the sidebar in the new usage page does not support switching between other periods (months).

### Proposal

| Sidebar | Sidebar (Dropdown) |
|-|-|
| <img width="500" alt="Sidebar-1" src="https://user-images.githubusercontent.com/120486/180205291-08396083-a9f0-4370-a303-64fbbcd5998a.png"> | <img width="500" alt="Sidebar" src="https://user-images.githubusercontent.com/120486/180205297-e5238e6d-eca5-4c63-81f1-f61807f2893a.png"> |

laushinka:
Hope you don't mind I reordered the wording to show the main task first 🙏🏽

gtsiolis:
Thanks, @laushinka! PERMISSION TO EDIT: ALWAYS ON 🏀

gtsiolis:
For more clarification, the goal of the designs is to provide dropdown options only for the months that the team had active billing months.

This comes with a scalability issue as the list could grow a lot after a year or two, which can be tackled later by either introducing a scrolling area or trimming the list to the last 6 months.

However, in the future, I'd expect this to become clearer by providing a separate invoice view where billed months are listed in a better and more scalable way similar to what has been initially described for the use page (see early designs in https://github.com/gitpod-io/gitpod/issues/10328#issuecomment-1165627563) and instead rely on date range filtering on the usage page for drilling down on usage for a specific invoice (billed month), see screenshot below.

<img width="520" alt="Screenshot 2022-07-26 at 11 22 31 AM" src="https://user-images.githubusercontent.com/120486/180959578-8f8584d9-6f6f-4305-8f5b-4be124303283.png">

laushinka:
Thanks, @gtsiolis! Writing down comments here after a quick brainstorm on the implementation with @easyCZ:

> the goal of the designs is to provide dropdown options only for the months that the team had active billing months.

For simplicity in this skateboard, we will show a list/dropdown/buttons/something of the previous 6 months without knowing beforehand whether the team has billed sessions or not. If there is no data, then that is tackled by the empty data page. 

> trimming the list to the last 6 months.

👍🏼

A summary of the functionality for this skateboard:
- User is shown by default the current month.
[...]

PR #11612, opened 2022-07-25
[usage] Extend server args with time-filtering

laushinka:
## Description
Extends the server with time-filtering args.
The client sends `number` and then it gets converted in the server to `google.protobuf.Timestamp`.

## Related Issue(s)

Relates #11526

## How to test
[Preview](https://laushinka-c45de4dc15.preview.gitpod-dev.com/t/gitpod/usage) still renders and calls `listBilledUsage` with the time params (the websocket connection in the network tab).

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-protocol/src/gitpod-service.ts
+++ b/components/gitpod-protocol/src/gitpod-service.ts
@@ -293,7 +293,7 @@ export interface GitpodServer extends JsonRpcServer<GitpodClient>, AdminServer, 
     getSpendingLimitForTeam(teamId: string): Promise<number | undefined>;
     setSpendingLimitForTeam(teamId: string, spendingLimit: number): Promise<void>;
 
-    listBilledUsage(attributionId: string): Promise<BillableSession[]>;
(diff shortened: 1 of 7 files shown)


easyCZ (approved):

components/gitpod-protocol/src/gitpod-service.ts
```diff
@@ -293,7 +293,7 @@ export interface GitpodServer extends JsonRpcServer<GitpodClient>, AdminServer,
     getSpendingLimitForTeam(teamId: string): Promise<number | undefined>;
     setSpendingLimitForTeam(teamId: string, spendingLimit: number): Promise<void>;
 
-    listBilledUsage(attributionId: string): Promise<BillableSession[]>;
+    listBilledUsage(attributionId: string, from?: number, to?: number): Promise<BillableSession[]>;
```
easyCZ, line 296: FYI, you can also pass in arguments for ordering, in case you wanted to implement ascending/descending

Issue #11609, opened 2022-07-25
Use dashboard `FeatureFlagContext` to retrieve and make feature flags available


andrew-farries:
fbdaff0fea added a `FeatureFlagContext` to retrieve feature flags and make them available throughout the UI.

We should move all the feature flags into this context so that for all future feature flags there is a clear example to follow.

- [x] Refactor to extract a `PageWithSettingsMenu` component to make following changes easier (https://github.com/gitpod-io/gitpod/pull/11610)
- [x] Move usage based pricing feature flag into `FeatureFlagContext` (https://github.com/gitpod-io/gitpod/pull/11613)
- [x] Move the other feature flags too.
  - [x] #11639  
  - [x] https://github.com/gitpod-io/gitpod/pull/11672

PR #11613, opened 2022-07-25
[dashboard] Move usage based pricing feature flag into feature flag context

andrew-farries:
## Description

Move the UBP feature flag into the `FeatureFlagContext`. We want to consolidate all setting of feature flags in this context.

Stacked on:
* https://github.com/gitpod-io/gitpod/pull/11610

## Related Issue(s)
Part of https://github.com/gitpod-io/gitpod/issues/11609

## How to test

- [x] Usage based billing UI is still displayed/hidden correctly according to the value of the feature flag.

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/dashboard/src/contexts/FeatureFlagContext.tsx
+++ b/components/dashboard/src/contexts/FeatureFlagContext.tsx
@@ -4,19 +4,46 @@
  * See License-AGPL.txt in the project root for license information.
  */
 
-import React, { createContext, useContext, useEffect } from "react";
+import React, { createContext, useContext, useState, useEffect } from "react";
+import { useLocation } from "react-router";
+import { getExperimentsClient } from "../experiments/client";
+import { ProjectContext } from "../projects/project-context";
+import { getCurrentTeam, TeamsContext } from "../teams/teams-context";
 import { UserContext } from "../user-context";
 
-const FeatureFlagContext = createContext({});
+const FeatureFlagContext = createContext<{ showUsageBasedPricingUI: boolean }>({
+    showUsageBasedPricingUI: false,
+});
 
 const FeatureFlagContextProvider: React.FC = ({ children }) => {
     const { user } = useContext(UserContext);
+    const { teams } = useContext(TeamsContext);
+    const { project } = useContext(ProjectContext);
+    const location = useLocation();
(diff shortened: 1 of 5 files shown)


components/dashboard/src/contexts/FeatureFlagContext.tsx
```diff
@@ -4,19 +4,46 @@
...
+import { ProjectContext } from "../projects/project-context";
+import { getCurrentTeam, TeamsContext } from "../teams/teams-context";
 import { UserContext } from "../user-context";
 
-const FeatureFlagContext = createContext({});
+const FeatureFlagContext = createContext<{ showUsageBasedPricingUI: boolean }>({
+    showUsageBasedPricingUI: false,
+});
 
 const FeatureFlagContextProvider: React.FC = ({ children }) => {
```
easyCZ, line 18: Would this Provider now need to implement every feature flag?

My thinking was that the component would accept (flagName, params) and return the value without needing to explicitly implement each flag in the ContextProvider

andrew-farries: The context is used here: 

https://github.com/gitpod-io/gitpod/blob/655eca7d3a2a867299b863bc057d51a99ab981bd/components/dashboard/src/index.tsx#L34-L36

ie at the very top level of the component tree, so it can't usefully take props. The idea of the context is that it gets the value of all feature flags and makes them available so that components further down the tree can look up the value of a flag like this:

https://github.com/gitpod-io/gitpod/blob/655eca7d3a2a867299b863bc057d51a99ab981bd/components/dashboard/src/settings/PageWithSettingsSubMenu.tsx#L20

The advantage is consolidation of all the feature flag retrieval logic in one place (in the context).

easyCZ: That's fair, just may not be intuitive for users that they also need to go implement the fetching. I guess it feels like a bit too much work to use a feature flag but I get it, and perhaps there isn't a better pattern in React

easyCZ (approved):
/hold
As it has a dependency

PR #11621, opened 2022-07-25
[usage] Fix conversion of from and to pb.Timestamp

easyCZ:
## Description

Because setSeconds doesn't work the same for ts Date and pb.Timestamp

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


laushinka (approved):
Thank you!

PR #11624, opened 2022-07-25
[werft] support large vm

iQQBot:
## Description

In ide integration test, we need run many workspace in the same time, this PR introduces a werft annotation `with-large-vm`, if job with this annotation, it will create a vm that use 12 cpu, and 24GB memory.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/.werft/jobs/build/prepare.ts
+++ b/.werft/jobs/build/prepare.ts
@@ -110,7 +110,9 @@ function createVM(werft: Werft, config: JobConfig) {
     }
 
     werft.log(prepareSlices.BOOT_VM, "Creating  VM");
-    VM.startVM({ name: config.previewEnvironment.destname });
+    const cpu = config.withLargeVM ? 12 : 6;
+    const memory = config.withLargeVM ? 24 : 12;
+    VM.startVM({ name: config.previewEnvironment.destname, cpu, memory });
     werft.currentPhaseSpan.setAttribute("preview.created_vm", true);
 }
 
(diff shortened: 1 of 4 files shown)


.werft/jobs/build/prepare.ts
```diff
@@ -110,7 +110,9 @@ function createVM(werft: Werft, config: JobConfig) {
     }
 
     werft.log(prepareSlices.BOOT_VM, "Creating  VM");
-    VM.startVM({ name: config.previewEnvironment.destname });
+    let cpu = config.withLargeVM ? 12 : 6;
```
andreafalzetti, line 113:
```suggestion
    const cpu = config.withLargeVM ? 12 : 6;
```

I think we can make both `cpu` and `memory` const because we don't modify them in this scope

andreafalzetti:
This is cool 🙏

Is there a way to test this flag?

andreafalzetti:
> > Is there a way to test this flag?
> 
> #11627 this PR use this flag

yeah, I saw the PR but I am wondering if you were able to test the flag itself, or until it gets deployed, there is no way of testing it?

ArthurSens:
Do we have the infrastructure to support this, though? 😬

Im not sure if our bare metal VMs are strong enough for this. 

What do you think @meysholdt?

meysholdt (approved):
yes, our infra should be able to handle this. Code changes LGTM, too.

Issue #11629, opened 2022-07-25
broken 'new project' widget when octokit/gh-api returns with null item(s)


szab100:
### Bug description

GH Enterprise (experienced with v3.3.9) instances (or at least its octokit client library) can contain null elements within the pages returned, therefore this ui widget is broken (basically frozen with no error msg) and preventing self-hosted users from setting up prebuilds for their git repositories.

### Steps to reproduce

This might be related to https://github.com/gitpod-io/gitpod/issues/11628, so reproducing might need a huge gh enterprise server with 10s of thousands of repositories OR this may be related to one or more repositories configured in a certain way on our side when this can occur.

geropl:
Fixed in #11630

Issue #11628, opened 2022-07-25
slow 'new project' ui widget with large github enterprise instances


szab100:
### Bug description

The |Team or persona scope| / Projects / New Project widget has been super slow since the feature's debut when running against large Github Enterprise instances with 10s of thousands of repositories under many-100 git-orgs. Since the current implementation is filtering to show only the repos where the current user has 'admin' permissions only after all pages are fetched and it makes the UI wait for the entire result set to be fetched before returning, the UI is frozen for 5-10+ minutes.

### Steps to reproduce

To successfully reproduce this issue, one would probably need access to a super beefy GH Enterprise instance (1k+ git-orgs with 10k+ git repos).

andrew-farries:
Some good suggestions for performance/usability improvements here: 

https://github.com/gitpod-io/gitpod/pull/11630#issuecomment-1195727935

PR #11630, opened 2022-07-25
fix: new project widget broken if 'null' item(s) received from gh api

szab100:
## Description
The |Team or persona scope| / Projects / New Project widget has been super slow since the feature's debut when running against large Github Enterprise instances with 10s of thousands of repositories under many-100 git-orgs. Since the current implementation is filtering to show only the repos where the current user has 'admin' permissions only after all pages are fetched and it makes the UI wait for the entire result set to be fetched before returning, the UI is frozen for 5-10+ minutes.

Still, even as being frozen / super slow (added a TODO to fix / optimize), this feature remained usable, since it is a one-time project setup. However, since we upgraded to GH Enterprise v 3.x, it seems that some returned pages might contain one or more null items, where the current filter expression fails. This PR simply aims at fixing this blocking issue, while not giving a remedy yet for the slowness of this API call.

## Related Issue(s)
https://github.com/gitpod-io/gitpod/issues/11628
https://github.com/gitpod-io/gitpod/issues/11629

Fixes #
https://github.com/gitpod-io/gitpod/issues/11629

## How to test
The fix should be trivial, simply ensure the feature still works after applying this change.

## Release Notes
```release-note
fix: new project widget broken if 'null' item(s) received from gh api
```

### DIFF
--- a/components/server/ee/src/prebuilds/github-service.ts
+++ b/components/server/ee/src/prebuilds/github-service.ts
@@ -25,6 +25,10 @@ export class GitHubService extends RepositoryService {
     @inject(TokenService) protected tokenService: TokenService;
     @inject(GithubContextParser) protected githubContextParser: GithubContextParser;
 
+    // TODO: consider refactoring this to either use GH Search API w/ typeahead search only OR
+    // return results in a stream, appending pages as being fetched (via callback below) OR
+    // simply use octokit.request w/ basic paginated api funcs (return # of pages + return a single page),
+    // so the UI can remain interactive vs. currently frozen for 5-10+ mins until all pages are fetched.
     async getRepositoriesForAutomatedPrebuilds(user: User): Promise<ProviderRepository[]> {
         const octokit = await this.githubApi.create(user);
         const repositories = await octokit.paginate(
@@ -32,7 +36,8 @@ export class GitHubService extends RepositoryService {
             { per_page: 100 },
             (response) =>
                 response.data
-                    .filter((r) => !!r.permissions?.admin)
+                    // On very large GH Enterprise (3.3.9) instances, items can be null
+                    .filter((r) => !!r?.permissions?.admin)
                     .map((r) => {
                         return <ProviderRepository>{
                             name: r.name,


andrew-farries (approved):
Thanks @szab100. The change looks good. This feels like a bug in GHE or octokit - perhaps worth following up with those projects.

/hold for CLA

meysholdt:
hi @szab100, and thank you for your contribution! Can you please sign our CLA via [this DocuSign form](https://eu.docusign.net/Member/PowerFormSigning.aspx?PowerFormId=30e1aa46-cce5-4bc6-a933-ad02cd8d5ada&env=eu&acct=86335c2c-01b9-4edc-a187-014f5150a112&v=2)? If there are any questions, you can reach me via moritz@gitpod.io.

szab100:
> hi @szab100, and thank you for your contribution! Can you please sign our CLA via [this DocuSign form](https://eu.docusign.net/Member/PowerFormSigning.aspx?PowerFormId=30e1aa46-cce5-4bc6-a933-ad02cd8d5ada&env=eu&acct=86335c2c-01b9-4edc-a187-014f5150a112&v=2)? If there are any questions, you can reach me via [moritz@gitpod.io](mailto:moritz@gitpod.io).

done.

meysholdt:
Thx for signing the CLA!

szab100:
> Thanks @szab100. The change looks good. This feels like a bug in GHE or octokit - perhaps worth following up with those projects.
> 
> /hold for CLA

I agree, that octokit client lib should never really return 'null' values, perhaps it is due to failed / retried paginated requests.

About the other mentioned issue (performance): it would be nice to see an improvement on that too soon, a few more comments / suggestions for that:
- a pure typeahead search would probably provide a great user experience as it could completely eliminate waiting..
- we might want to return the current_user != admin repos as well, marking the objects, so the UI can show them as greyed out / unselectable (with a comment explaining this action needs admin) => less confusing than not seeing the repository the user is looking for.
- returning a stream could be good too, the UI could interactively load the results & show the progress (not sure if octokit returns the counter or # of total pages upfront to show a deterministic progress bar).
- I have not tested, but I believe even just adding the 'affiliation: collaborator' (server-side) filter to the current octokit call would make it much better than today as - although it is a lower permission level than we need - it should greatly decrease the number of pages to fetch & further filter on, skipping the prob 10s of thousands of repositories where the user is not even a collaborator.

andrew-farries:
> a few more comments / suggestions for that:

Thanks for these 👍. I've added them to https://github.com/gitpod-io/gitpod/issues/11628 so they don't get lost.

PR #11642, opened 2022-07-26
[local-preview] Add new Telemetry

### LINKED ISSUE #11301: Local Preview Telemetry: Beginning of Installation
For the local preview installation method we want to not only know how many users successfully installed it, but also how many tired in order to get an understanding of the success rate. This helps us understand how complicated or easy it is to install, and can potentially be used to measure the impact of any improvements. 

Get telemetry on the number of users who tried installing

- Wrapper bash script that sends data to segment
- Measure number of image downloads

## Additional context
- Internal [chat](https://gitpod.slack.com/archives/C01KLC56NP7/p1657603801141869)

Pothulapati:
## Description

Currently, The telemetry of `local-preview` is the same as any self-hosted install which
means that these telemetry cronjob is run daily, and also once when the installation is ready.
This information isn't as useful, as the properties aren't that important for a short-lived
installation.

Instead, It would make sense to have telemetry on each event of the `local-preview`
command itself so that we can measure the Installation success rate,  Time taken, etc
to improve the overall UX of `local-preview`. This will be more important for us
to understand the usage of `local-preview` itself.

This PR fixes that by attaching a telemetry event to be sent, for each status update
in `local-preview`. For this to work, A anonymous User ID is generated and used
for all status updates.

Event Schema:

```
analytics.track({
    userId: userId,
    event: "localpreview_status",
    properties: {
        status: "checking prerequisites" | "starting gitpod" | "gitpod ready"
    }
})
```

## Related Issue(s)

Fixes #11301 

## How to test

Run

```
docker run -p 443:443 --privileged --name gitpod --rm -it -v gitpod:/var/gitpod eu.gcr.io/gitpod-core-dev/build/local-preview:tar-lp-telemetry.26
```

## Release Notes

```release-note
[local-preview] Add separated anonymous telemetry
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/preview/prettylog/go.mod
+++ b/install/preview/prettylog/go.mod
@@ -1,10 +1,9 @@
-module github.com/gitpod-io/gitpod/olpi/prettylog
+module github.com/gitpod-io/gitpod/install/preview/prettylog
 
 go 1.18
 
-require github.com/pterm/pterm v0.12.41
-
 require (
+	github.com/pterm/pterm v0.12.41
 	github.com/atomicgo/cursor v0.0.1 // indirect
 	github.com/gookit/color v1.5.0 // indirect
 	github.com/mattn/go-runewidth v0.0.13 // indirect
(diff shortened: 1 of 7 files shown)


install/preview/prettylog/main.go
```diff
@@ -72,3 +85,22 @@ func main() {
 		panic(err)
 	}
 }
+
+func send_telemetry(event string) {
+	if segmentIOToken == "" {
+		panic("No segmentIOToken set during build")
+	}
+
+	client, _ := analytics.NewWithConfig(segmentIOToken, analytics.Config{})
```
Pothulapati, line 94: Should we panic on these err's? 🤔

install/preview/prettylog/go.mod
```diff
@@ -1,15 +1,21 @@
-module github.com/gitpod-io/gitpod/olpi/prettylog
+module github.com/gitpod-io/gitpod/install/prettylog
```
MrSimonEmms, line 1: Should this be?
```suggestion
module github.com/gitpod-io/gitpod/install/preview/prettylog
```

MrSimonEmms (approved):
/hold one nit question - if you want to ignore, feel free to unhold

PR #11643, opened 2022-07-26
[server] add http endpoint workspacePageClose for beacon

mustard-mh:
## Description

Add HTTP endpoint for supervisor frontend to send heartbeat when user's leaving the browser

In the current implementation of supervisor-frontend, we will send some requests before/during browser unload, which may cause some data loss since requests will be canceled after 1 second once user close the tab. 

https://github.com/gitpod-io/gitpod/blob/00d510825c93b01bc3865c81d909d80ac6cf38df/components/supervisor/frontend/src/ide/heart-beat.ts#L45-L58

We will move those requests to use [Navigator.sendBeacon](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon) to avoid data loss.

So this PR is going to add HTTP post endpoint for sendBeacon

see also [internal chat](https://gitpod.slack.com/archives/C01KGM9BH54/p1658742853684919?thread_ts=1658720614.040559&cid=C01KGM9BH54)

## Related Issue(s)

Fixes #

## How to test

- Open a workspace in preview env
- Get needed info from cookie (preview env logged in cookie and workspace auth cookie)
- CURL to see if API works
```bash
curl -X "POST" "https://hw-hb.preview.gitpod-dev.com/api/auth/workspacePageClose/<instance_id>" \
     -H 'Cookie: _hw_hb_preview_gitpod_dev_com_=****;_hw_hb_preview_gitpod_dev_com_ws_****_owner_=****' \
     -H 'Content-Type: application/json; charset=utf-8' \
     -d $'{
  "sessionId": "1"
}'
```
- Use `gpctl workspaces last-heartbeat <instance_id>` to see if heartbeat sent
- Use `kubectl get pods ws-<instance_id> -oyaml | yq4 .metadata.annotations | grep closed` to see if `wasClosed` works

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/server/src/user/user-controller.ts
+++ b/components/server/src/user/user-controller.ts
@@ -27,9 +27,12 @@ import { TosCookie } from "./tos-cookie";
 import { TosFlow } from "../terms/tos-flow";
 import { increaseLoginCounter } from "../../src/prometheus-metrics";
 import { v4 as uuidv4 } from "uuid";
-import { ScopedResourceGuard } from "../auth/resource-access";
+import { OwnerResourceGuard, ScopedResourceGuard } from "../auth/resource-access";
 import { OneTimeSecretServer } from "../one-time-secret-server";
 import { trackSignup } from "../analytics";
+import { WorkspaceManagerClientProvider } from "@gitpod/ws-manager/lib/client-provider";
+import { EnforcementControllerServerFactory } from "./enforcement-endpoint";
+import { ClientMetadata } from "../websocket/websocket-connection-manager";
 
 @injectable()
 export class UserController {
@@ -46,6 +49,9 @@ export class UserController {
     @inject(LoginCompletionHandler) protected readonly loginCompletionHandler: LoginCompletionHandler;
     @inject(OneTimeSecretServer) protected readonly otsServer: OneTimeSecretServer;
     @inject(OneTimeSecretDB) protected readonly otsDb: OneTimeSecretDB;
+    @inject(WorkspaceManagerClientProvider)
+    protected readonly workspaceManagerClientProvider: WorkspaceManagerClientProvider;
+    @inject(EnforcementControllerServerFactory) private readonly serverFactory: EnforcementControllerServerFactory;
 
     get apiRouter(): express.Router {
         const router = express.Router();
@@ -279,6 +285,64 @@ export class UserController {
                     domain: `.${this.config.hostUrl.url.host}`,
                 });
                 res.sendStatus(200);
+            },
+        );
+
+        router.post(
+            "/auth/workspacePageClose/:instanceID",
+            async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+                if (!req.isAuthenticated() || !User.is(req.user)) {
+                    res.sendStatus(401);
+                    log.warn("unauthenticated workspacePageClose", { instanceId: req.params.instanceID });
+                    return;
+                }
+
+                const user = req.user as User;
+                if (user.blocked) {
+                    res.sendStatus(403);
+                    log.warn("blocked user attempted to workspacePageClose", {
+                        instanceId: req.params.instanceID,
+                        userId: user.id,
+                    });
+                    return;
+                }
+
+                const instanceID = req.params.instanceID;
+                if (!instanceID) {
+                    res.sendStatus(400);
+                    log.warn("attempted to workspacePageClose without instance ID", {
+                        instanceId: req.params.instanceID,
+                        userId: user.id,
+                    });
+                    return;
+                }
+                const sessionId = req.body.sessionId;
+                const server = this.createGitpodServer(user);
+                try {
+                    await server.sendHeartBeat({}, { wasClosed: true, instanceId: instanceID });
+                    /** no await */ server
+                        .trackEvent(
+                            {},
+                            {
+                                event: "ide_close_signal",
+                                properties: {
+                                    sessionId,
+                                    instanceId: instanceID,
+                                    clientKind: "supervisor-frontend",
+                                },
+                            },
+                        )
+                        .catch((err) =>
+                            log.warn({ userId: user.id }, "workspacePageClose: failed to track ide close signal", err),
+                        );
+                    res.sendStatus(200);
+                } catch (e) {
+                    log.error("workspacePageClose failed", e);
+                    res.sendStatus(500);
+                    return;
+                } finally {
+                    server.dispose();
+                }
             },
         );
         if (this.config.enableLocalApp) {
@@ -704,5 +768,18 @@ export class UserController {
 
         log.debug({ sessionId: req.sessionID }, "The redirect URL does not match", { query: req.query });
         return;
+    }
+
+    private createGitpodServer(user: User) {
+        const server = this.serverFactory();
+        server.initialize(
+            undefined,
+            user,
+            new OwnerResourceGuard(user.id),
+            ClientMetadata.from(user.id),
+            undefined,
+            {},
+        );
+        return server;
     }
 }


components/server/src/user/user-controller.ts
```diff
@@ -281,6 +286,95 @@ export class UserController {
                 res.sendStatus(200);
             },
         );
+
+        router.post(
+            "/auth/sendHeartbeat/:instanceID",
```
akosyakov, line 291: not sure why `auth`? 

We should call it not `sendHeartbeat` but something `logWorkspacePageClose`. It is not about heartbeating, but recording a fact that the page was closed.

components/server/src/user/user-controller.ts
```diff
@@ -281,6 +286,95 @@ export class UserController {
...
+                const instanceID = req.params.instanceID;
+                if (!instanceID) {
+                    res.sendStatus(400);
+                    log.warn("attempted to send heartbeat without instance ID", {
+                        instanceId: req.params.instanceID,
+                        userId: user.id,
+                    });
+                    return;
+                }
+                const [workspace, instance] = await Promise.all([
```
akosyakov, line 320: is it copied from somewhere? can we extract it to some service and reuse in both places that code of sending heartbeating is not duplicated

components/server/src/analytics.ts
```diff
@@ -46,6 +46,32 @@ export async function trackSignup(user: User, request: Request, analytics: IAnal
     });
 }
 
+export async function trackIDECloseSignal(
```
akosyakov, line 49: I think it should be inlined in route, there is no need to reuse it somewhere

akosyakov:
> add http send heartbeat endpoint

Please make it sure to update the PR description to explain why we do it instead of using Gitpod API, i.e. it is not about general heartbeating, but making sure that we report close signal always. General heartbeating still happens using Gitpod API, but it is not reliable to use while page is unloading.

components/server/src/analytics.ts
```diff
@@ -46,6 +46,32 @@ export async function trackSignup(user: User, request: Request, analytics: IAnal
...
+    clientKind: string,
+    sessionId?: string,
+) {
+    // make new complete identify call for each login
+    await fullIdentify(user, request, analytics);
+
+    // track the ide close signal
+    analytics.track({
+        userId: user.id,
+        anonymousId: stripCookie(request.cookies.ajs_anonymous_id),
```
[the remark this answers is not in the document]
mustard-mh: TrackEvent will add them too
https://github.com/gitpod-io/gitpod/blob/20fe1e55df48361a3f74a746d5b75fc734d3ae39/components/server/src/workspace/gitpod-server-impl.ts#L2962-L2990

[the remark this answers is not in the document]
akosyakov: Can we extarct trackEvent to some common service and call it as well from here?

components/server/src/user/user-controller.ts
mustard-mh: It's a copy from code above, we can refactor the code but not in this PR?

components/server/src/analytics.ts
mustard-mh: Should we respect the original structure of the code? Make code align to previous is better
 
https://github.com/gitpod-io/gitpod/blob/20fe1e55df48361a3f74a746d5b75fc734d3ae39/components/server/src/user/user-controller.ts#L32

components/server/src/user/user-controller.ts
akosyakov: I think we should refactor in this PR. copy/pasting is not good.

mustard-mh: Refactor should go with add middlewares like userVerify, workspaceOwnedVerify etc,

It'll increase the difficulty of PR testing, and maybe break something, that all these APIs are very important

components/server/src/analytics.ts
akosyakov: I'm not sure what you mean?

The server is built with https://github.com/inversify/InversifyJS for configurations which resolves around classes. If we want to reuse some functions from GitpodSerevr then we should move them to some existing services or to some intermediate which have sense conceptually.

We can also create a new instance of server as it is done here and call it instead: https://github.com/gitpod-io/gitpod/blob/2f0b9462b0451f0974434fa36bd757420ef04e34/components/server/src/user/enforcement-endpoint.ts#L50-L52

components/server/src/user/user-controller.ts
akosyakov: We need to verify these things for this endpoint as well, no?

components/server/src/analytics.ts
mustard-mh: If we want to use server like this, we will copy-paste needed code in this file 😅

mustard-mh: I'd like to keep the code

mustard-mh: Reuse of `enforcement-endpoint` need to refactor `enforcement-endpoint.ts` file too, since it's not that reusable. And have some permission check

akosyakov: I don't think we need to touch encofcement endpoint.

We have a user and it should be enough:
```ts
@inject(EnforcementControllerServerFactory) private readonly serverFactory: EnforcementControllerServerFactory;

private createGitpodServer(user: User) {
        const server = this.serverFactory();
        server.initialize(undefined, user, new OwnerResourceGuard(user.id), ClientMetadata.from(user.id), undefined, {});
        return server;
    }
```

we can rename EnforcementControllerServerFactory to GitpodServerFactory?

components/server/src/user/user-controller.ts
```diff
@@ -281,6 +286,95 @@ export class UserController {
...
+                    this.workspaceDB.findByInstanceId(instanceID),
+                    this.workspaceDB.findInstanceById(instanceID),
+                ]);
+                if (!workspace || instance?.status.phase !== "running") {
+                    res.sendStatus(404);
+                    log.warn("attempted to send heartbeat for non-existent workspace instance", {
```
andreafalzetti, line 326: should this be `log.warn(<context>, <message>)`? [[1]](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-protocol/src/util/logging.ts#L53)

I think `userId`, `istanceId` are part of [LogContext](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-protocol/src/util/logging.ts#L14-L19)

components/server/src/analytics.ts
mustard-mh: Thank you @akosyakov 🙏

components/server/src/user/user-controller.ts
mustard-mh: @andreafalzetti  See here https://github.com/gitpod-io/gitpod/blob/9b46ff3a216d5e598b91abf832569b91bad3e10d/components/server/src/user/user-controller.ts#L205-L208

components/server/src/user/user-controller.ts
```diff
@@ -281,6 +287,60 @@ export class UserController {
...
+                            {
+                                event: "ide_close_signal",
+                                properties: {
+                                    sessionId,
+                                    instanceId: instanceID,
+                                    clientKind: "supervisor-frontend",
+                                },
+                            },
+                        )
+                        .catch((err) => log.warn({ userId: user.id }, "trackIDECloseSignal", err));
```
akosyakov, line 335: "workspacePageClose: failed to track ide close signal"

components/server/src/user/user-controller.ts
mustard-mh: I thought `log` support a lot of param pairs, and we used to make log string first param

components/server/src/user/user-controller.ts
```diff
@@ -281,6 +287,62 @@ export class UserController {
...
+                    res.sendStatus(400);
+                    log.warn("attempted to workspacePageClose without instance ID", {
+                        instanceId: req.params.instanceID,
+                        userId: user.id,
+                    });
+                    return;
+                }
+                const sessionId = req.body.sessionId;
+                try {
+                    const server = this.createGitpodServer(user);
```
akosyakov, line 321: I'm not sure, but maybe we need `try-finally` to call `server.dispose`.

components/server/src/user/user-controller.ts
mustard-mh: I'm not quite sure if the order of params will affect the query in gcloud, I unresolved this conversation and maybe webApp team can answer us

components/server/src/user/user-controller.ts
```diff
@@ -705,4 +769,17 @@ export class UserController {
         log.debug({ sessionId: req.sessionID }, "The redirect URL does not match", { query: req.query });
         return;
     }
+
+    private createGitpodServer(user: User) {
+        const server = this.serverFactory();
+        server.initialize(
+            undefined,
+            user,
+            new OwnerResourceGuard(user.id),
```
geropl, line 778: Note: This line ensures that only the owner can access the `sendHeartbeat` method for their instances.

geropl (approved):
Code LGTM

components/server/src/user/user-controller.ts
```diff
@@ -281,6 +287,64 @@ export class UserController {
...
+                                },
+                            },
+                        )
+                        .catch((err) =>
+                            log.warn({ userId: user.id }, "workspacePageClose: failed to track ide close signal", err),
+                        );
+                    res.sendStatus(200);
+                } catch (e) {
+                    log.error("workspacePageClose failed", e);
+                    res.sendStatus(500);
```
akosyakov, line 341: @mustard-mh I just realised that it is wrong, i.e. if a user tries to access a workspace to which they don't have access it will pop up in our metrics like 500 (i.e. our failure), than it should be 403. Could you have a look sendHeartbeat method and translate exceptions properly? 🙏 cc @gitpod-io/engineering-webapp

components/server/src/user/user-controller.ts
akosyakov: ouch, it is a valid point in case of live sharing cc @mustard-mh we need to check how construct proper guard similar how we do for web socket connections, i think there should be a service fro HTTP connections as well already

components/server/src/user/user-controller.ts
geropl: > I'm not quite sure if the order of params will affect the query in gcloud

@mustard-mh It does indeed.

> we used to make log string as first param in most cases

That's not good. The expected order is: logContext, message, error, payload.

PR #10085, opened 2022-05-18
Make the pods be removed when workspacekit fails.

utam0k:
## Description

There was a problem with workspace pendings due to finalizer not working properly when iws failed to start (or had a large delay). This PR avoids that by skipping the finalizer. Also, the workspace phase is now consistent with it.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/9673

## How to test

- workspaces can normally start. You can confirm [this preview environment](https://to-conn-prob.preview.gitpod-dev.com/workspaces)
- Intentionally break communication between iws and ws-daemon, and when a workspace failed to start, confirm that  the workspace phase is properly stopped(NOTE: in this case, the stopping phase is skipped because we don't need to backup.) You can confirm it on [this preview environment](https://to-failed-ws-start.preview.gitpod-dev.com/workspaces)

## Release Notes

```release-note
ws-manager: Make the pods be removed when workspacekit fails.
```

## Documentation

No

PR #11644, opened 2022-07-26
Remove the PVC object if the workspace pod is never been ready

### LINKED ISSUE #11635: Remove the PVC object if the workspace pod is never been ready
### Bug description

We found that in some scenarios (eg, the workspace pod in ErrImagePull), the workspace pod never goes into the Running state.
After that, the workspace is stopped, but the PVC object doesn't be deleted.

### Steps to reproduce

No idea, but I saw it in workspace preview that the workspace pod is in ErrImagePull state and then terminating state.

image

After that, the PVC is orphaned, and we don't have to create VolumeSnapshot object since this workspace doesn't have to take backup.

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

#7901

jenting:
## Description

Remove the PVC object if the workspace pod is never been ready; we still go through the stopping phase even if the workspace is not ready. However, we don't perform backup but we just remove the finalizer in the end (and remove the PVC object if it existed).

## Related Issue(s)

Fixes #11635

## How to test

I have no idea how to make the workspace pod has never been in a ready state. 🤔 
But you could use this cluster for testing this PR https://jenting-10531.preview.gitpod-dev.com/workspaces, and open the workspace ID `gitpodio-templatenix-kzpw34u`. This workspace always hits the ErrImagePull error which belongs to the workspace pod that has never been in a ready state.

image
image

## Release Notes

```release-note
None
```

## Documentation

None

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/ws-manager/pkg/manager/monitor.go
+++ b/components/ws-manager/pkg/manager/monitor.go
@@ -498,6 +498,11 @@ func actOnPodEvent(ctx context.Context, m actingManager, manager *Manager, statu
 					}
 				}
 
+				// delete PVC because the workspace pod is never ready
+				if err = manager.deleteWorkspacePVC(ctx, pod.Name); err != nil {
+					log.Error(err)
+					return err
+				}
 				return m.modifyFinalizer(ctx, workspaceID, gitpodFinalizerName, false)
 			} else {
 				// We start finalizing the workspace content only after the container is gone. This way we ensure there's
@@ -1183,16 +1188,10 @@ func (m *Monitor) finalizeWorkspaceContent(ctx context.Context, wso *workspaceOb
 			// backup is done and we are ready to kill the pod, mark PVC for deletion
 			if readyVolumeSnapshot && !deletedPVC {
 				// todo: once we add snapshot objects, this will be changed to create snapshot object first
-				pvcErr := m.manager.Clientset.Delete(ctx,
-					&corev1.PersistentVolumeClaim{
-						ObjectMeta: metav1.ObjectMeta{
-							Name:      pvcName,
-							Namespace: m.manager.Config.Namespace,
-						},
-					},
-				)
-				if pvcErr != nil && !k8serr.IsNotFound(pvcErr) {
-					log.WithError(pvcErr).Errorf("failed to delete pvc `%s`", pvcName)
+				err = m.manager.deleteWorkspacePVC(ctx, pvcName)
+				if err != nil {
+					log.Error(err)
+					return true, nil, err
 				}
 				deletedPVC = true
 			}
--- a/components/ws-manager/pkg/manager/status.go
+++ b/components/ws-manager/pkg/manager/status.go
@@ -349,14 +349,6 @@ func (m *Manager) extractStatusFromPod(result *api.WorkspaceStatus, wso workspac
 	if isPodBeingDeleted(pod) {
 		result.Phase = api.WorkspacePhase_STOPPING
 
-		_, podFailedBeforeBeingStopped := pod.Annotations[workspaceFailedBeforeStoppingAnnotation]
-		if podFailedBeforeBeingStopped {
-			if _, ok := pod.Annotations[workspaceNeverReadyAnnotation]; ok {
-				// The workspace is never ready, so there is no need for a stopping phase.
-				result.Phase = api.WorkspacePhase_STOPPED
-			}
-		}
-
 		var hasFinalizer bool
 		for _, f := range wso.Pod.Finalizers {
 			if f == gitpodFinalizerName {
--- a/components/ws-manager/pkg/manager/testdata/status_workspace_startup_failed.golden
+++ b/components/ws-manager/pkg/manager/testdata/status_workspace_startup_failed.golden
@@ -20,7 +20,7 @@
             },
             "class": "default"
         },
-        "phase": 6,
+        "phase": 5,
         "conditions": {
             "volume_snapshot": {}
         },
--- a/components/ws-manager/pkg/manager/manager.go
+++ b/components/ws-manager/pkg/manager/manager.go
@@ -724,6 +724,23 @@ func (m *Manager) stopWorkspace(ctx context.Context, workspaceID string, gracePe
 	return nil
 }
 
+func (m *Manager) deleteWorkspacePVC(ctx context.Context, pvcName string) error {
+	pvc := &corev1.PersistentVolumeClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      pvcName,
+			Namespace: m.Config.Namespace,
+		},
+	}
+	err := m.Clientset.Delete(ctx, pvc)
+	if k8serr.IsNotFound(err) {
+		err = nil
+	}
+	if err != nil {
+		return xerrors.Errorf("cannot delete workspace pvc: %w", err)
+	}
+	return nil
+}
+
 func (m *Manager) deleteWorkspaceSecrets(ctx context.Context, podName string) error {
 	secret := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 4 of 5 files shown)


components/ws-manager/pkg/manager/status.go
```diff
@@ -354,11 +354,6 @@ func (m *Manager) extractStatusFromPod(result *api.WorkspaceStatus, wso workspac
 			// While the pod is being deleted we do not care or want to know about any failure state.
 			// If the pod got stopped because it failed we will have sent out a Stopping status with a "failure"
 			result.Conditions.Failed = ""
-		} else {
-			if _, ok := pod.Annotations[workspaceNeverReadyAnnotation]; ok {
```
sagor999, line 358: why this part is being removed?

jenting: Because I think if the workspace pod is **never ready**, it still needs to go through the [Stopping](https://github.com/gitpod-io/gitpod/blob/093e8c6586183660a4bdb9f218e4afd12ca1ad9e/components/ws-manager/pkg/manager/monitor.go#L430) phase to do extra logic (e.g.: resource clean up). The idea is like the Kubernetes controller reconcile loop, and we do extra logic if the resource is Terminating with the finalizer protected, and then remove the finalizer.

From my testing, the scenario ErrImagePull makes the `status.Conditions.Failed != ""`, so the annotation `workspaceFailedBeforeStoppingAnnotation` be added.
https://github.com/gitpod-io/gitpod/blob/093e8c6586183660a4bdb9f218e4afd12ca1ad9e/components/ws-manager/pkg/manager/monitor.go#L333-L341

During the workspace pod in the Stopping state, it contains the annotation `workspaceFailedBeforeStoppingAnnotation` and `workspaceNeverReadyAnnotation`.
https://github.com/gitpod-io/gitpod/blob/093e8c6586183660a4bdb9f218e4afd12ca1ad9e/components/ws-manager/pkg/manager/monitor.go#L487-L501

jenting: @utam0k
Do you know how to simulate the workspacekit faiI?
I want to check for this PR changes, would it break your previous contribution #10085.

sagor999: Ah, it makes sense then.
Only concern is to do some extra testing, just to make sure there is no code that was made on the assumption of this code existing. 
I think you already tested regular workspaces, can you also test prebuilds and image builds? Just to be extra safe.

components/ws-manager/pkg/manager/testdata/status_workspace_startup_failed.golden
```diff
@@ -20,7 +20,7 @@
             },
             "class": "default"
         },
-        "phase": 6,
+        "phase": 5,
```
sagor999, line 23: Slightly concerned over this test case change. It shows that phase would be potentially stuck in "stopping", instead of being "stopped". We might have a regression here where some workspaces would not stop correctly?

sagor999 (changes requested):
I like this change :rocket: but asking to do a bit more testing to verify nothing else would break. Left comments.

components/ws-manager/pkg/manager/status.go
utam0k: @jenting 
- Make this time short
     https://github.com/gitpod-io/gitpod/blob/90b8032d1c78e8d21549dbd5a01eacf7d056d860/components/workspacekit/cmd/rings.go#L971
- Change this socket file name
    https://github.com/gitpod-io/gitpod/blob/90b8032d1c78e8d21549dbd5a01eacf7d056d860/components/ws-daemon/pkg/iws/iws.go#L156

jenting: Thank you @sagor999 for the suggestion and @utam0k for the hack way.

- prebuild
image
image
- image build
image
image

components/ws-manager/pkg/manager/testdata/status_workspace_startup_failed.golden
jenting: 🤔 I need to double check the `status_workspace_startup_failed.json`.
I just ran the golden file update `go test ./... -update -force`.

jenting: Okay, the root cause is that the status test tests the output of the function [getWorkspaceStatus](https://github.com/gitpod-io/gitpod/blob/77d57bcb0a08d881070291d0ba8cfa0a8b45fa83/components/ws-manager/pkg/manager/status_test.go#L114) matches the golden file or not, and only executes _once_.

Since we remove changing the state at [here](https://github.com/gitpod-io/gitpod/blob/347af9bff0b870590f4da5b8df49a98b7297800b/components/ws-manager/pkg/manager/status.go#L352-L358). Therefore, the function output of the `phase` field would be `stopping`, not `stopped`.

However, for this case, the code goes into [here](https://github.com/gitpod-io/gitpod/blob/77d57bcb0a08d881070291d0ba8cfa0a8b45fa83/components/ws-manager/pkg/manager/monitor.go#L489-L502), removing the finalizer, and reenter into the `getWorkspaceStatus` at [here](https://github.com/gitpod-io/gitpod/blob/77d57bcb0a08d881070291d0ba8cfa0a8b45fa83/components/ws-manager/pkg/manager/status.go#L371). In the end, the phase would be `stopped`, that's the server receives final state.

components/ws-manager/pkg/manager/status.go
sagor999: I am confused. Why those screenshots show errors? 
I would have expected that prebuilds\imagebuild work with this change? Or am I misreading something here?

sagor999: Or is this testing workspacekit failure?

components/ws-manager/pkg/manager/monitor.go
```diff
@@ -498,6 +498,11 @@ func actOnPodEvent(ctx context.Context, m actingManager, manager *Manager, statu
 					}
 				}
 
+				// delete PVC because the workspace pod is never ready
+				if err = manager.deleteWorkspacePVC(ctx, pod.Name); err != nil {
+					log.Error(err)
```
sagor999, line 503: can you change this back to WithError and provide a short blurb that error is due to failure to delete workspace pvc? otherwise it will be hard to understand where error might be getting logged and coming from.

components/ws-manager/pkg/manager/monitor.go
```diff
@@ -1183,16 +1188,10 @@ func (m *Monitor) finalizeWorkspaceContent(ctx context.Context, wso *workspaceOb
...
-							Name:      pvcName,
-							Namespace: m.manager.Config.Namespace,
-						},
-					},
-				)
-				if pvcErr != nil && !k8serr.IsNotFound(pvcErr) {
-					log.WithError(pvcErr).Errorf("failed to delete pvc `%s`", pvcName)
+				err = m.manager.deleteWorkspacePVC(ctx, pvcName)
+				if err != nil {
+					log.Error(err)
```
sagor999, line 1193: same here

components/ws-manager/pkg/manager/status.go
jenting: > Or is this testing workspacekit failure? 

Sport for missing add the context.

Yes, I was testing the workspacekit failure against regular, prebuild, and imagebuild cases, verify that the workspace pod is able to be terminated and removed, and also the pod phased is stopped in the end.

components/ws-manager/pkg/manager/monitor.go
jenting: Because if the error happens, the error message from the `deleteWorkspacePVC` would be `return xerrors.Errorf("cannot delete workspace pvc: %w", err)`. And I think it contains enough information. WDYT?

sagor999 (approved):
🚀

Issue #10785, opened 2022-06-21
[usage] Decouple billing controller from usage controller


easyCZ:
Fix up how we trigger the sending of the usage report into Stripe. Currently it's done inside the ReconcileTimeRange This currently causes tests to report
[ERROR] Request error from Stripe (status 401): {"status":401,"message":"You did not provide an API key. 
What we want is something akin to a pipeline.
Generate the report and return it. This completes the usage controller "data munching"
(later) Store it.
Pass it onto a Billing Controller.
In practice, it should be possible to succeed the usage report, but fail the Stripe update (for example for Stripe outage reasons)

easyCZ:
Investigating this further, the easiest seems to actually introduce the following Services and RPCs:
1. UsageService.CollectUsage()
2. BillingService.UpdateInvoices()
These two are then responsible for performing the respective parts of the control loops, can be invoked remotely but in the common case, are invoked periodically by the cron in the controller.

Doing so also gives us a couple of benefits:
1. Improved observability of either task, as we can piggy back on gRPC observability
2. When we need to sync explicitly, we can do so over an RPC
3. Makes it possible to in the future decouple the RPC trigger from the implementation - k8s cron

**How would this actually look like?**
* `billing.proto` - https://github.com/gitpod-io/gitpod/pull/11631/files#diff-82630443ae7d5700a98e48b4aeaf4b6ecbc55c2156f639e58aa64cb92bda934bR1
* `usage.proto` - https://github.com/gitpod-io/gitpod/pull/11631/files#diff-02538d53a95c532babc3fba3868ed35d20f3625383b5886e360fa48947db87b9L9

easyCZ:
This is now complete.

PR #11645, opened 2022-07-26
[usage] Remove pricer dependency from billing controller

easyCZ:
## Description

This wasn't needed any more, as we already compute the CreditsUsed in the usage controller
This also fixes a bug where the pricer was using a different `now` timestamp to what the usage controller used.

## Related Issue(s)

Relates to https://github.com/gitpod-io/gitpod/issues/10785

## How to test

Unit tests

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/usage/pkg/contentservice/client.go
+++ b/components/usage/pkg/contentservice/client.go
@@ -20,7 +20,7 @@ import (
 )
 
 type Interface interface {
-	UploadUsageReport(ctx context.Context, filename string, report map[db.AttributionID][]db.WorkspaceInstanceForUsage) error
+	UploadUsageReport(ctx context.Context, filename string, report []db.WorkspaceInstanceUsage) error
 }
 
 type Client struct {
@@ -31,7 +31,7 @@ func New(url string) *Client {
 	return &Client{url: url}
 }
 
-func (c *Client) UploadUsageReport(ctx context.Context, filename string, report map[db.AttributionID][]db.WorkspaceInstanceForUsage) error {
+func (c *Client) UploadUsageReport(ctx context.Context, filename string, report []db.WorkspaceInstanceUsage) error {
 	url, err := c.getSignedUploadUrl(ctx, filename)
 	if err != nil {
 		return fmt.Errorf("failed to obtain signed upload URL: %w", err)
--- a/components/usage/pkg/controller/reconciler.go
+++ b/components/usage/pkg/controller/reconciler.go
@@ -72,7 +72,7 @@ func (u *UsageReconciler) Reconcile() (err error) {
 	}
 	log.WithField("usage_reconcile_status", status).Info("Reconcile completed.")
 
-	err = db.CreateUsageRecords(ctx, u.conn, usageReportToUsageRecords(report, u.pricer, u.nowFunc().UTC()))
+	err = db.CreateUsageRecords(ctx, u.conn, report)
 	if err != nil {
 		return fmt.Errorf("failed to write usage records to database: %s", err)
 	}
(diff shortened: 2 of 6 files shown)


components/usage/pkg/server/server.go
```diff
@@ -65,7 +65,7 @@ func Start(cfg Config) error {
 			return fmt.Errorf("failed to initialize stripe client: %w", err)
 		}
 
-		billingController = controller.NewStripeBillingController(c, pricer)
+		billingController = controller.NewStripeBillingController(c)
```
easyCZ, line 68: This is the goal of the PR, the billing controller doesn't need the pricer as the WorkspaceInstanceUsage records already have credits used on them.

components/usage/pkg/contentservice/client.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 type Interface interface {
-	UploadUsageReport(ctx context.Context, filename string, report map[db.AttributionID][]db.WorkspaceInstanceForUsage) error
+	UploadUsageReport(ctx context.Context, filename string, report []db.WorkspaceInstanceUsage) error
```
easyCZ, line 23: By taking the `[]db.WorkspaceInstanceUsage` we ensure that what goes to the bill is the same as our records, but also we no longer need the pricing config. The credit price has already been computed so there's no need to compute it again.

components/usage/pkg/controller/reconciler.go
```diff
@@ -205,50 +241,3 @@ func trimStartStopTime(instances []db.WorkspaceInstanceForUsage, maximumStart, m
 	}
 	return updated
 }
-
-func groupInstancesByAttributionID(instances []db.WorkspaceInstanceForUsage) UsageReport {
```
easyCZ, line 209: Instead of grouping here, and then un-grouping to insert the WorkspaceInstanceUsage records into the DB, we just not group. 

The grouping is then done in the BillingController as part of GetCreditReport()

components/usage/pkg/contentservice/client.go
andrew-farries: The previous report format had the nice property that usage was keyed by attribution id, whereas now it's a flat list of instances. Does that matter?

components/usage/pkg/controller/reconciler.go
```diff
@@ -105,33 +105,69 @@ func (u *UsageReconciler) ReconcileTimeRange(ctx context.Context, from, to time.
...
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to reconcile billing: %w", err)
 	}
 
-	return status, instancesByAttributionID, nil
+	return status, usageRecords, nil
 }
 
-type UsageReport map[db.AttributionID][]db.WorkspaceInstanceForUsage
+func instancesToUsageRecords(instances []db.WorkspaceInstanceForUsage, pricer *WorkspacePricer, now time.Time) []db.WorkspaceInstanceUsage {
```
andrew-farries, line 119: These two types `db.WorkspaceInstanceForUsage` and `db.WorkspaceInstanceUsage` are very similarly named. Are there better names we can use?

easyCZ: Suggestions welcome, I couldn't come up with one. The `ForUsage` is a join between WorkspaceInstance and Workspace

components/usage/pkg/contentservice/client.go
easyCZ: I think there was a higher chance of error, especially in tests The previous form was:
```
{
   "team:<teamID>": [
       { InstanceID: "asdas", AttributionID: "asdad" }
   ]
}
```
We've moved the grouping into the billing controller which I think works better, and I don't see much harm in doing the grouping there.

andrew-farries (approved):

PR #11647, opened 2022-07-26
[supervisor-frontend] use beacon to send heartbeat to avoid data loss

mustard-mh:
## Description

In the current implementation of supervisor-frontend, we will send some requests before/during browser unload, which may cause some data loss since requests will be canceled after 1 second once user close the tab. 

We will move those requests to use [Navigator.sendBeacon](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon) to avoid data loss.

So this PR is going to add HTTP post endpoint for sendBeacon

see also https://github.com/gitpod-io/gitpod/pull/11643 and [internal chat](https://gitpod.slack.com/archives/C01KGM9BH54/p1658742853684919?thread_ts=1658720614.040559&cid=C01KGM9BH54)

👉 The first commit is cherry-picked from prev PR https://github.com/gitpod-io/gitpod/pull/11643

## Related Issue(s)

Fixes #

### With prepared env (part of code changed)

- Use preview env https://hw-hb-t.preview.gitpod-dev.com/workspaces which change `main` to `hw-hb-t` as SaaS verify condition
- Use desktop code as IDE
- Search chrome devTool log with `unloadListener` should only get `handleUnload.unloadListener.Beacon` according to code below
- Check with [segment](https://app.segment.com/gitpod/sources/staging_untrusted/debugger) should shown `ide_close_signal` if you close or refresh the tab

https://github.com/gitpod-io/gitpod/blob/a72f0a0873c087ee4845d78eb4e2374d0191e6cc/components/supervisor/frontend/src/ide/heart-beat.ts#L67-L83

image

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/supervisor/frontend/src/ide/gitpod-server-compatibility.ts
+++ b/components/supervisor/frontend/src/ide/gitpod-server-compatibility.ts
@@ -1,0 +1,51 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { serverUrl } from '../shared/urls';
+
+const currentHost = new URL(serverUrl.toString()).hostname;
+
+export const isSaaS = currentHost === 'gitpod.io'
+
+const versionRegex = new RegExp('main\.(\\d+)')
+
+function getVersionInfo(version: string) {
+    const result = versionRegex.exec(version);
+    if (!result) {
+        return;
+    }
+    return Number(result[1]);
+}
+
+const serverVersion = (async () => {
+    const url = serverUrl.withApi({ pathname: '/version' }).toString();
+    const fetchVersion = async (retry: number) => {
+        try {
+            const resp = await fetch(url);
+            const currentVersionStr = await resp.text();
+            return getVersionInfo(currentVersionStr);
+        } catch (e) {
+            if (retry - 1 <= 0) {
+                throw e;
+            }
+            fetchVersion(retry - 1)
+        }
+    }
+    try {
+        return await fetchVersion(3)
+    } catch (e) {
+        console.error('failed to fetch server verson:', e)
+    }
+})();
+
+export async function isSaaSServerGreaterThan (version: string) {
+    if (!isSaaS) {
+        return false;
+    }
+    const serverVersionNum = await serverVersion;
+    const versionNum = getVersionInfo(version);
+    return !!serverVersionNum  && !!versionNum  && serverVersionNum >= versionNum;
+}
--- a/components/supervisor/frontend/src/ide/heart-beat.ts
+++ b/components/supervisor/frontend/src/ide/heart-beat.ts
@@ -6,6 +6,8 @@
 
 import { DisposableCollection, Disposable } from '@gitpod/gitpod-protocol/lib/util/disposable';
 import { WorkspaceInfo } from "@gitpod/gitpod-protocol";
+import { isSaaSServerGreaterThan, isSaaS } from './gitpod-server-compatibility';
+import { serverUrl } from '../shared/urls';
 
 let lastActivity = 0;
 const updateLastActivitiy = () => {
@@ -15,6 +17,12 @@ export const track = (w: Window) => {
     w.document.addEventListener('mousemove', updateLastActivitiy, { capture: true });
     w.document.addEventListener('keydown', updateLastActivitiy, { capture: true });
 }
+
+let pageCloseCompatibile: boolean = false
+
+isSaaSServerGreaterThan("main.4124").then((r) => {
+    pageCloseCompatibile = r
+})
 
 let toCancel: DisposableCollection | undefined;
 export function schedule(wsInfo: WorkspaceInfo, sessionId: string): void {
@@ -41,22 +49,25 @@ export function schedule(wsInfo: WorkspaceInfo, sessionId: string): void {
         }
     }
     sendHeartBeat();
-    let unloadTimeout: any;
-    const beforeUnloadListener = () => {
-        unloadTimeout = setTimeout(() => {
-            // if unload was cancelled then resume heartbeating
-            sendHeartBeat();
-        }, 2000);
-        sendHeartBeat(true);
-    };
-    const unloadListener = () => {
-        if (unloadTimeout) {
-            clearTimeout(unloadTimeout);
-        }
+
+    const beaconWorkspacePageClose = () => {
+        const instanceId = wsInfo.latestInstance!.id;
+        const data = { sessionId };
+        const blob = new Blob([JSON.stringify(data)], { type: "application/json" });
+        const url = serverUrl.withApi({ pathname: `/auth/workspacePageClose/${instanceId}` }).toString();
+        navigator.sendBeacon(url, blob);
     }
-    window.addEventListener('beforeunload', beforeUnloadListener);
+
+    const workspacePageCloseCompatible = () => {
+        if (isSaaS && !pageCloseCompatibile) {
+            sendHeartBeat(true);
+            return;
+        }
+        beaconWorkspacePageClose();
+    }
+
+    const unloadListener = () => { workspacePageCloseCompatible() };
     window.addEventListener('unload', unloadListener);
-    toCancel.push(Disposable.create(() => window.removeEventListener('beforeunload', beforeUnloadListener)));
     toCancel.push(Disposable.create(() => window.removeEventListener('unload', unloadListener)));
 
     let activityInterval = 30000;


roboquat:

<details>

</details>

akosyakov:
Can we change base of this PR to https://github.com/gitpod-io/gitpod/pull/11643 that we can test the entire flow? We should consider backward compatibility here somehow. Maybe we check in advance version of server to be sure that we can send beacon at the end?

mustard-mh:
> We should consider backward compatibility here somehow

@akosyakov  Yes, I'll check `/api/version`, and this PR should be merged later than another one to confirm the version

For self-host, we don't need to consider the compatibility, just to notice self-hosted team to not split these two PRs 🙈

akosyakov:
@mustard-mh It is not ready for review yet, right? Should we convert it to draft?

mustard-mh:
> @mustard-mh It is not ready for review yet, right? Should we convert it to draft?

I thought it's ready

akosyakov:
> I thought it's ready

I was not sure, it has WIP in name,

akosyakov:
@mustard-mh I changed the PR base to hw/hb-server, can you drop unnecessary commit please

components/supervisor/frontend/src/ide/gitpod-server-compatibility.ts
```diff
@@ -0,0 +1,26 @@
...
+function getVersionInfo(version: string) {
+    const result = versionRegex.exec(version);
+    if (!result) {
+        return;
+    }
+    return Number(result[1]);
+}
+
+export async function isSaasServerGreaterThan (version: string) {
+    const url = serverUrl.withApi({ pathname: "/version" }).toString();
```
akosyakov, line 20: Should not we check that we are on gitpod.io?

This check does not makes sense in Self-Hosted.

components/supervisor/frontend/src/ide/heart-beat.ts
```diff
@@ -17,7 +19,7 @@ export const track = (w: Window) => {
 }
 
 let toCancel: DisposableCollection | undefined;
-export function schedule(wsInfo: WorkspaceInfo, sessionId: string): void {
+export async function schedule(wsInfo: WorkspaceInfo, sessionId: string) {
```
akosyakov, line 22: I think it should stay sync to push heartbeat as early as possible.

components/supervisor/frontend/src/ide/gitpod-server-compatibility.ts
```diff
@@ -0,0 +1,26 @@
...
+    const resp = await fetch(url);
```
akosyakov, line 21: we should handle it it fails case, i.e. retry like 3 times and then assume that everything is up-to-date with some warning

components/supervisor/frontend/src/ide/heart-beat.ts
```diff
@@ -40,23 +42,27 @@ export function schedule(wsInfo: WorkspaceInfo, sessionId: string): void {
...
+        const url = serverUrl.withApi({ pathname: `/auth/workspacePageClose/${instanceId}` }).toString();
+        navigator.sendBeacon(url, blob);
+    }
+
+    let unloadListener = () => {
+        beaconWorkspacePageClose();
+    }
+
+    // TODO: Change version to correct one
+    if (!(await isSaasServerGreaterThan("main.4063"))) {
```
akosyakov, line 59: Maybe if we do it earlier, i.e. then this file is loaded and store in some variable, something like
```
const serverVersion = (async () => {
   ....
}());
```

and isSaasServerGreaterThan awaits on it

components/supervisor/frontend/src/ide/heart-beat.ts
```diff
@@ -70,6 +68,21 @@ export function schedule(wsInfo: WorkspaceInfo, sessionId: string): void {
...
+        const saasCompatibility = async () => {
+            const result = await isSaaSServerGreaterThan("main.4091");
+            if (result) {
+                window.removeEventListener('unload', unloadListener);
+                unloadListener = () => {
+                    beaconWorkspacePageClose();
+                }
+                window.addEventListener('unload', unloadListener);
+            }
+            toCancel?.push(Disposable.create(() => window.removeEventListener('unload', unloadListener)));
```
akosyakov, line 82: it should be called even if isSaaS is not executed.

components/supervisor/frontend/src/ide/heart-beat.ts
```diff
@@ -70,6 +68,21 @@ export function schedule(wsInfo: WorkspaceInfo, sessionId: string): void {
...
+        }
+        saasCompatibility().then().catch(console.error);
```
akosyakov, line 84: (nit) not sure `.then()` is necessary

components/supervisor/frontend/src/ide/heart-beat.ts
```diff
@@ -70,6 +63,22 @@ export function schedule(wsInfo: WorkspaceInfo, sessionId: string): void {
...
+                window.addEventListener('unload', unloadListener);
+                toCancel?.push(Disposable.create(() => window.removeEventListener('unload', unloadListener)));
+                return;
+            }
+        }
+        const unloadListener = () => { sendHeartBeat(true) }
+        window.addEventListener('unload', unloadListener);
+        toCancel?.push(Disposable.create(() => window.removeEventListener('unload', unloadListener)));
+    }
+    handleUnload().catch(console.error);
```
AlexTugarev, line 81: _Just a comment as I was scanning open PRs but did not mean to provide an in-depth review now._ 

I've no experience with `sendBeacon`, so had to read on it on https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon. 

This implementation has two problems:
1. usage of `unload` is discouraged, see docs
2. handler awaiting on fetch results before actually calling `sendBeacong` won't be effective, I guess. I didn't test it, but I'd bet the response would never be processed.

akosyakov:
We need also another PR later to fix endpoint before we start using it, see https://github.com/gitpod-io/gitpod/pull/11643#discussion_r934283285 and https://github.com/gitpod-io/gitpod/pull/11643#discussion_r934525099

components/supervisor/frontend/src/ide/heart-beat.ts
mustard-mh: @AlexTugarev  Thank you for your review 🧡 

> 1. usage of unload is discouraged, see docs

I tried with `pagehide` but it just not works, and `visibilitychange` seems not good for our use case, since switch tab should not send close event. 

> 2. handler awaiting on fetch results before actually calling `sendBeacon` won't be effective, I guess. I didn't test it, but I'd bet the response would never be processed.

the `async` keyword is not necessary, I'll remove it later

mustard-mh:
/hold

To resolve https://github.com/gitpod-io/gitpod/pull/11647#issuecomment-1202640248 first

mustard-mh:
I created a preview env https://hw-hb-t.preview.gitpod-dev.com/workspaces which change `main` to `hw-hb-t` as SaaS verify condition

We should test this PR here, since prev env cannot be SaaS version

- Use desktop code as IDE
- Search chrome devTool log with `unloadListener` should only get `handleUnload.unloadListener.Beacon` according to code below
- Check with [segment](https://app.segment.com/gitpod/sources/staging_untrusted/debugger) should shown `ide_close_signal`

https://github.com/gitpod-io/gitpod/blob/a72f0a0873c087ee4845d78eb4e2374d0191e6cc/components/supervisor/frontend/src/ide/heart-beat.ts#L67-L83

image

Also updated PR description

iQQBot (approved):
Test it works as expected

with version `hw-hb-t.17`, it use sendbeacon

image

image


for `hw-hb-t.1` it use websocket send heartbeat

image

components/supervisor/frontend/src/ide/gitpod-server-compatibility.ts
```diff
@@ -0,0 +1,51 @@
...
+            if (retry - 1 <= 0) {
+                throw e;
+            }
+            fetchVersion(retry - 1)
+        }
+    }
+    try {
+        return await fetchVersion(3)
+    } catch (e) {
+        console.error(e)
```
akosyakov, line 40: please add better logging, something like `console.error("failed to fetch server verson:", e)` and explicitly return undefined

components/supervisor/frontend/src/ide/gitpod-server-compatibility.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import { serverUrl } from "../shared/urls";
+
+const currentHost = new URL(serverUrl.toString()).hostname;
+
+export const isSaaS = currentHost === "gitpod.io"
+
+const versionRegex = new RegExp("main\.(\\d+)")
+
+function getVersionInfo(version: string) {
```
akosyakov, line 15: `string | undefined`

components/supervisor/frontend/src/ide/heart-beat.ts
```diff
@@ -24,7 +26,7 @@ export function schedule(wsInfo: WorkspaceInfo, sessionId: string): void {
     toCancel = new DisposableCollection()
     const sendHeartBeat = async (wasClosed?: true) => {
         try {
-            await window.gitpod.service.server.sendHeartBeat({ instanceId: wsInfo.latestInstance!.id, wasClosed });
+            window.gitpod.service.server.sendHeartBeat({ instanceId: wsInfo.latestInstance!.id, wasClosed });
```
akosyakov, line 29: why should we change this line? try-catcth does not make sense anymore, and ide_close_signal can be fired even if sendHeartBeat failed?

components/supervisor/frontend/src/ide/heart-beat.ts
```diff
@@ -70,6 +63,22 @@ export function schedule(wsInfo: WorkspaceInfo, sessionId: string): void {
         sendHeartBeat();
     }, activityInterval);
     toCancel.push(Disposable.create(() => clearInterval(intervalHandle)));
+
+    const handleUnload = async () => {
+        if (isSaaS) {
```
akosyakov, line 68: For self-hosting we should always use a new approach? Only backward compatibility for SaaS?

components/supervisor/frontend/src/ide/heart-beat.ts
mustard-mh: Ahh, I should move .catch to `sendHeartBeat` function, I thought we don't have to send `ide_close_signal` after `sendHeartBeat` is done. 

And it's another reason the previous implement unload did not works, the browser only gives around one second, `await` makes the `ide_close_signal` tracking more missable

mustard-mh:
I update the code to align to https://github.com/gitpod-io/gitpod/pull/11647#discussion_r936280871

Could you review the code when you're free? 🙏  @iQQBot @felladrin

components/supervisor/frontend/src/ide/heart-beat.ts
```diff
@@ -70,6 +65,27 @@ export function schedule(wsInfo: WorkspaceInfo, sessionId: string): void {
...
+            unloadListener = () => { sendHeartBeat(true) };
+            window.addEventListener('unload', unloadListener);
+            toCancel?.push(Disposable.create(() => window.removeEventListener('unload', unloadListener)));
+            const result = await isSaaSServerGreaterThan("main.4091");
+            if (result) {
+                window.removeEventListener('unload', unloadListener);
+                unloadListener = () => { beaconWorkspacePageClose() };
+                window.addEventListener('unload', unloadListener);
```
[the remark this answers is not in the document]
mustard-mh:
```typescript
toCancel?.push(Disposable.create(() => window.removeEventListener('unload', unloadListener)))
```

The `unloadListener` is inside a closure, it'll always be the latest refer of function (and moved to global var), so it's ok

components/supervisor/frontend/src/ide/heart-beat.ts
```diff
@@ -70,6 +65,27 @@ export function schedule(wsInfo: WorkspaceInfo, sessionId: string): void {
         sendHeartBeat();
     }, activityInterval);
     toCancel.push(Disposable.create(() => clearInterval(intervalHandle)));
+
+    const handleUnload = async () => {
+        if (isSaaS) {
+            unloadListener = () => { sendHeartBeat(true) };
+            window.addEventListener('unload', unloadListener);
+            toCancel?.push(Disposable.create(() => window.removeEventListener('unload', unloadListener)));
```
akosyakov, line 73: what if it already cancelled? Should we check for cancelled after each await? or add listener sync always and move condition within a listener?

components/supervisor/frontend/src/ide/heart-beat.ts
```diff
@@ -70,6 +65,27 @@ export function schedule(wsInfo: WorkspaceInfo, sessionId: string): void {
...
+            const result = await isSaaSServerGreaterThan("main.4091");
```
akosyakov, line 74: should it be updated after latest PR was merged?

akosyakov (approved):
lgtm

I left some comment, the code to me looks tricky. I afraid there are race conditions with handling of canellation.

mustard-mh:
Code updated, could you review it? @iQQBot

iQQBot (approved):
Code LGTM, Thanks @mustard-mh

PR #11650, opened 2022-07-26
[usage] Add BillingService with UpdateInvoices RPC

easyCZ:
## Description

To move the logic into an RPC, and decouple it from the Usage logic

## Related Issue(s)

* Relates to https://github.com/gitpod-io/gitpod/issues/10785

## How to test

CI runs

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


components/usage-api/usage/v1/billing.proto
```diff
@@ -0,0 +1,22 @@
...
+
+service BillingService {
+  rpc UpdateInvoices(UpdateInvoicesRequest) returns (UpdateInvoicesResponse) {};
+}
+
+message UpdateInvoicesRequest {
+  google.protobuf.Timestamp start_time = 1;
+  google.protobuf.Timestamp end_time = 2;
+
+  repeated BilledSession sessions = 3;
```
andrew-farries, line 18: As a client of this RPC I would assume I can:

* Give a start time and an end time and have invoices updated for all instances that ran in that time (maybe I'd like to give attributionIds too, so that I only update invoices for that team/user).

but here I can pass a list of `BilledSessions` as inputs (presumably intended to be those returned by the `ListBilledUsage` RPC on the usage service. What happens if I pass billed sessions that lie outside the time range?

easyCZ: So this is not a general purpose RPC, in a way, it's an internal RPC. I can clarify in comments. 

You're correct, this only makes sense in the context of another thing which produces the sessions. I'm not sure how to rename or better convey this but the intended usage is not outside callers without context, but the "usage reconcile run"

andrew-farries: I think some comments above the service to that effect would be good 👍

easyCZ: When we do have authN in depth, I would've also encoded that the only caller of this RPC can be a service account which has the perms to generate usage reports, which I believe would've helped.

easyCZ: Added a comment on the proto

andrew-farries (approved):

PR #11651, opened 2022-07-26
[bitbucket] make less requests and fail on error

### LINKED ISSUE #11649: [BitBucket] Failed calls to fetch `.gitpod.yml` are silently ignored
### Bug description

https://github.com/gitpod-io/gitpod/blob/main/components/server/src/bitbucket/bitbucket-file-provider.ts#L65

silently ignores (debug) thrown errors and as a result the caller is interpreting it as no config and continues with the project config and eventually inferring one. This results in strange and surprising behavior. We should bail out if this happens.

I found this in the logs:

```
"HTTPError: Too Many Requests
    at /app/node_modules/bitbucket/lib/index.umd.js:15:6585
    at processTicksAndRejections (node:internal/process/task_queues:96:5)"
```

### Steps to reproduce

try often enough

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

svenefftinge:
This PR makes the logic for finding a gitpod.yml sequential.
Currently, we run the gitpod.yml inference on every workspace start, which sends lots of requests (around 20) to the git provider in order to check for build files. Since this is a fallback mode that is only relevant in those cases where there is no `.gitpod.yml` I have changed the logic so it is sequential and only runs when really needed.

fixes #11649

## How to test
Start workspaces on repositories with gitpod.yml and without.
Make sure the right gitpod.yml is used.

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/server/src/workspace/config-provider.ts
+++ b/components/server/src/workspace/config-provider.ts
@@ -182,15 +182,12 @@ export class ConfigProvider {
                 }
                 const services = hostContext.services;
                 const contextRepoConfig = services.fileProvider.getGitpodFileContent(commit, user);
-                const projectDBConfig = this.projectDB
-                    .findProjectByCloneUrl(commit.repository.cloneUrl)
-                    .then((project) => project?.config);
-                const definitelyGpConfig = this.fetchExternalGitpodFileContent({ span }, commit.repository);
-                const inferredConfig = this.configurationService.guessRepositoryConfiguration({ span }, user, commit);
-
                 customConfigString = await contextRepoConfig;
                 let origin: WorkspaceConfig["_origin"] = "repo";
                 if (!customConfigString) {
+                    const projectDBConfig = this.projectDB
+                        .findProjectByCloneUrl(commit.repository.cloneUrl)
+                        .then((project) => project?.config);
                     // We haven't found a Gitpod configuration file in the context repo - check the "Project" in the DB.
                     const config = await projectDBConfig;
                     if (config) {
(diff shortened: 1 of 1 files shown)


components/server/src/bitbucket/bitbucket-file-provider.ts
```diff
@@ -15,10 +15,7 @@ export class BitbucketFileProvider implements FileProvider {
     @inject(BitbucketApiFactory) protected readonly apiFactory: BitbucketApiFactory;
 
     public async getGitpodFileContent(commit: Commit, user: User): Promise<MaybeContent> {
-        const yamlVersion1 = await Promise.all([
-            this.getFileContent(commit, user, ".gitpod.yml"),
-            this.getFileContent(commit, user, ".gitpod"),
```
svenefftinge, line 20: I don't know where all the API calls are coming from, but asking for `.gitpod` every time is unnecessary.

easyCZ: Do we have any metrics that would allow us to trace this? Could we add the metrics first to understand the source of these requests?

svenefftinge: I agree we need to better understand why we run into the limit (atlassian is unclear about that). But we have customers sitting on a bad experience today and removing this call does make sense in anyway even if it doesn't help with the rate limit. So I'd like to go ahead with this change and follow up with investigation of the api calls.

svenefftinge:
Turns out that a lot of calls are coming from config inferrer

easyCZ (approved):
Change looks sensible, but I frankly don't have sufficient context around this part of config resolution to know if there are skeletons. Adding hold in case you want to wait for a review from someone with more context.

/hold

Issue #6894, opened 2021-11-25
Implement analytics on changes to the `.gitpod.yml`


loujaybee:
**Context:** This one is important as it will give us data around when a user is updating this file as part of their onboarding journey. e.g. are users updating the file on their first visit? second visit? etc

We would like to know: if / when a user is changing or updating their gitpod.yml. 

Ideally, we would gather fine-grained data on which properties the user is updating on the YAML file. This is so we can optimize how and when we present certain features to our users. 

Considerations: 
* We could possibly implement this via a keystroke event within VS Code, but we'll need to set a reasonable debounce/throttle on the event to ensure we do not send many events. 
* We will need to decide which attributes/aspects of the `gitpod.yml` that are interesting for us to track and why. 

**Note:** If necessary we can break this ticket down, simply knowing that the user has changed the file / added it would deliver value at this stage.

filiptronicek:
@loujaybee this is a very interesting concept. I wonder if we could do more things on the client by just adding a tailored extension, which would take care of the whole `gitpod.yml` walkthrough process. This would mean nothing is sent back - good for our servers and also 0 latency. I imagine in this process there won't be too many features which we couldn't implement via an extension, but just a thought 💡!

loujaybee:
This one is a little tricky, so probably best if @akosyakov is able to collaborate with whoever picks this up to come up with a plan. As with all of the analytics tickets, having something working initially is better than trying to overthink it, and come up with something super comprehensive.

loujaybee:
There is potentially some tracking added here [[1](https://github.com/gitpod-io/gitpod/blob/da3991c94b08365b6de71b9337ee69a338b21dfc/components/supervisor/pkg/supervisor/supervisor.go#L1306)], let's investigate if we have that data, and if those events are working as expected.

akosyakov:
> There is potentially some tracking added already here in the supervisor [1].

@jakobhero Could you check please whether we have `config-changed` in Segment, but which does not make to BigQuery for some reasons?

jakobhero:
> whether we have config-changed in Segment, but which does not make to BigQuery for some reasons?

Hi @akosyakov, i had another look at the events coming into Segment and `config-changed` never makes it there. Also, the BigQuery destination is not configured to block any of the events. Are we sending the events to the source with the write key of [production trusted](https://app.segment.com/gitpod/sources/production_trusted/overview)?

akosyakov:
[...]

PR #11653, opened 2022-07-26
[supervisor]: gitpod config changed analytics

akosyakov:
## Description

This PR adds analytics for .gitpod.yml changes. We are track when changes are persisted to a disk and debounce them for 5s. After last change we analyse changes to top-level elements and report an event for each. It allows us to gain insights in what users are changing on top-level, e.g `ports` or `tasks`,   and avoid a lot of traffic at the same time. Later we can do more fine grained analysis to see which object properties, like `ports.name`.

I had to resolve couple issues which were discovered during review or were blocking testing:
- enable analytics env inside workspace in preview environments to verify with segment
- don't fire change events for syntactically broken .gitpod.yml files, thank you @andreafalzetti for noticing it
- adding git hook which keeps goland types up to date with gitpod schema, otherwise we will miss new properties, thank you @mustard-mh for brining forward that yaml parser only see properties which are defined in goland

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/6894

## How to test

- Open staging untrusted.
- Start a workspace, open .gitpod.yml yet, wait for 5s.
- You should not yet see any `gitpod_config_changed` events.
- Now edit .gitpoy.yml, then you finish wait for 5s.
- Now you should see `gitpod_config_changed` events.
- Edit .gitpod.yml to add some syntax errors and semantic changes, then you finish wait for 5s.
- You should not yet see any `gitpod_config_changed` events.
- Edit .gitpod.yml to fix syntax errors, then you finish wait for 5s.
- Now you should see `gitpod_config_changed` events for semantic changes.

Check for regressions:
- Add new ports and check there are detected in the ports view.

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-protocol/data/gitpod-schema.json
+++ b/components/gitpod-protocol/data/gitpod-schema.json
@@ -261,140 +261,20 @@
                     }
                 },
                 "intellij": {
-                    "type": "object",
-                    "description": "Configure IntelliJ integration",
-                    "additionalProperties": false,
-                    "properties": {
-                        "plugins": {
-                            "type": "array",
-                            "description": "List of plugins which should be installed for users of this workspace. From the JetBrains Marketplace page, find a page of the required plugin, select 'Versions' tab, click any version to copy pluginId (short name such as org.rust.lang) of the plugin you want to install.",
-                            "items": {
-                                "type": "string"
-                            }
-                        },
-                        "prebuilds": {
-                            "type": "object",
-                            "description": "Enable warming up of IntelliJ in prebuilds.",
-                            "additionalProperties": false,
-                            "properties": {
-                                "version": {
-                                    "type": "string",
-                                    "enum": [
-                                        "stable",
-                                        "latest",
-                                        "both"
-                                    ],
-                                    "description": "Whether only stable, latest or both versions should be warmed up. Default is stable only."
-                                }
-                            }
-                        },
-                        "vmoptions": {
-                            "type": "string",
-                            "description": "Configure JVM options, for instance '-Xmx=4096m'."
-                        }
-                    }
+                    "$ref": "#/definitions/jetbrainsProduct",
+                    "description": "Configure IntelliJ integration"
                 },
                 "goland": {
-                    "type": "object",
-                    "description": "Configure GoLand integration",
-                    "additionalProperties": false,
-                    "properties": {
-                        "plugins": {
-                            "type": "array",
-                            "description": "List of plugins which should be installed for users of this workspace. From the JetBrains Marketplace page, find a page of the required plugin, select 'Versions' tab, click any version to copy pluginId (short name such as org.rust.lang) of the plugin you want to install.",
-                            "items": {
-                                "type": "string"
-                            }
-                        },
-                        "prebuilds": {
-                            "type": "object",
-                            "description": "Enable warming up of GoLand in prebuilds.",
-                            "additionalProperties": false,
-                            "properties": {
-                                "version": {
-                                    "type": "string",
-                                    "enum": [
-                                        "stable",
-                                        "latest",
-                                        "both"
-                                    ],
-                                    "description": "Whether only stable, latest or both versions should be warmed up. Default is stable only."
-                                }
-                            }
-                        },
-                        "vmoptions": {
-                            "type": "string",
-                            "description": "Configure JVM options, for instance '-Xmx=4096m'."
-                        }
-                    }
+                    "$ref": "#/definitions/jetbrainsProduct",
+                    "description": "Configure GoLand integration"
                 },
                 "pycharm": {
-                    "type": "object",
-                    "description": "Configure PyCharm integration",
-                    "additionalProperties": false,
-                    "properties": {
-                        "plugins": {
-                            "type": "array",
-                            "description": "List of plugins which should be installed for users of this workspace. From the JetBrains Marketplace page, find a page of the required plugin, select 'Versions' tab, click any version to copy pluginId (short name such as org.rust.lang) of the plugin you want to install.",
-                            "items": {
-                                "type": "string"
-                            }
-                        },
-                        "prebuilds": {
-                            "type": "object",
-                            "description": "Enable warming up of PyCharm in prebuilds.",
-                            "additionalProperties": false,
-                            "properties": {
-                                "version": {
-                                    "type": "string",
-                                    "enum": [
-                                        "stable",
-                                        "latest",
-                                        "both"
-                                    ],
-                                    "description": "Whether only stable, latest or both versions should be warmed up. Default is stable only."
-                                }
-                            }
-                        },
-                        "vmoptions": {
-                            "type": "string",
-                            "description": "Configure JVM options, for instance '-Xmx=4096m'."
-                        }
-                    }
+                    "$ref": "#/definitions/jetbrainsProduct",
+                    "description": "Configure PyCharm integration"
                 },
                 "phpstorm": {
-                    "type": "object",
-                    "description": "Configure PhpStorm integration",
-                    "additionalProperties": false,
-                    "properties": {
-                        "plugins": {
-                            "type": "array",
-                            "description": "List of plugins which should be installed for users of this workspace. From the JetBrains Marketplace page, find a page of the required plugin, select 'Versions' tab, click any version to copy pluginId (short name such as org.rust.lang) of the plugin you want to install.",
-                            "items": {
-                                "type": "string"
-                            }
-                        },
-                        "prebuilds": {
-                            "type": "object",
-                            "description": "Enable warming up of PhpStorm in prebuilds.",
-                            "additionalProperties": false,
-                            "properties": {
-                                "version": {
-                                    "type": "string",
-                                    "enum": [
-                                        "stable",
-                                        "latest",
-                                        "both"
-                                    ],
-                                    "description": "Whether only stable, latest or both versions should be warmed up. Default is stable only."
-                                }
-                            }
-                        },
-                        "vmoptions": {
-                            "type": "string",
-                            "description": "Configure JVM options, for instance '-Xmx=4096m'."
-                        }
-                    }
+                    "$ref": "#/definitions/jetbrainsProduct",
+                    "description": "Configure PhpStorm integration"
                 }
             }
         },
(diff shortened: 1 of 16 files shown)


roboquat:

<details>

</details>

akosyakov:
Important if you want to rebuild then you start a Gitpod workspace and do following:
`werft run github -s .werft/jobs/build/installer/installer.ts -s .werft/jobs/build/installer/post-process.sh -f`

Because of security concerns, changes to werft scripts are not picked by default, but scripts from main are used. We need to side load them.

components/supervisor/pkg/config/gitpod-config-analytics.go
```diff
@@ -0,0 +1,145 @@
...
+	return &ConfigAnalyzer{
+		log:    log,
+		delay:  delay,
+		report: report,
+		prev:   initial,
+	}
+}
+
+func (a *ConfigAnalyzer) Analyse(current *gitpod.GitpodConfig) <-chan struct{} {
+	a.log.Debug("gitpod config analytics: schedulling")
```
andreafalzetti, line 44:
```suggestion
	a.log.Debug("gitpod config analytics: scheduling")
```

components/supervisor/pkg/config/gitpod-config-analytics.go
```diff
@@ -0,0 +1,145 @@
...
+		WithField("fields", fields).
+		Debug("gitpod config analytics: processed")
+
+	a.prev = current
+}
+
+func (a *ConfigAnalyzer) computeFields(configs ...*gitpod.GitpodConfig) []string {
+	defer func() {
+		if err := recover(); err != nil {
+			a.log.WithField("error", err).Error("failed to compute gitpod config fields")
```
akosyakov, line 85: `gitpod config analytics:`

components/supervisor/pkg/config/gitpod-config-analytics.go
```diff
@@ -0,0 +1,145 @@
...
+			}
+		}
+	}
+	return fields
+}
+
+func (a *ConfigAnalyzer) valueByField(config *gitpod.GitpodConfig, field string) interface{} {
+	defer func() {
+		if err := recover(); err != nil {
+			a.log.WithField("error", err).WithField("field", field).Error("failed to retrieve value from gitpod config")
```
akosyakov, line 112: `gitpod config analytics:`

components/supervisor/pkg/supervisor/supervisor.go
```diff
@@ -1513,82 +1512,39 @@ func socketActivationForDocker(ctx context.Context, wg *sync.WaitGroup, term *te
...
-		h := sha256.New()
-		_, err = h.Write(b)
-		if err != nil {
-			return "", err
-		}
-		return fmt.Sprintf("%x", h.Sum(nil)), nil
+func analyseConfigChanges(ctx context.Context, wscfg *Config, w analytics.Writer, cfgobs config.ConfigInterface, gitpodAPI gitpod.APIInterface) {
+	info, err := gitpodAPI.GetWorkspace(ctx, wscfg.WorkspaceID)
+	if err != nil {
+		log.WithError(err).Error("failed to track config changes")
```
akosyakov, line 1518: `gitpod config analytics: failed to fetch workspace owner`

components/supervisor/pkg/supervisor/supervisor.go
```diff
@@ -1513,82 +1512,39 @@ func socketActivationForDocker(ctx context.Context, wg *sync.WaitGroup, term *te
...
-			return "", err
-		}
-		h := sha256.New()
-		_, err = h.Write(b)
-		if err != nil {
-			return "", err
-		}
-		return fmt.Sprintf("%x", h.Sum(nil)), nil
+func analyseConfigChanges(ctx context.Context, wscfg *Config, w analytics.Writer, cfgobs config.ConfigInterface, gitpodAPI gitpod.APIInterface) {
+	info, err := gitpodAPI.GetWorkspace(ctx, wscfg.WorkspaceID)
```
andreafalzetti, line 1516: Is there any benefit on calling the API instead of using `os.Getenv("GITPOD_WORKSPACE_ID")`?

andreafalzetti: never mind, I just noticed it's for the owner

akosyakov: yes, GITPOD_WORKSPACE_ID is internal implementation, Config abstracts away from it.

components/supervisor/pkg/config/gitpod-config-analytics.go
```diff
@@ -0,0 +1,145 @@
...
+
+	a.log.WithField("current", current).
+		WithField("prev", a.prev).
+		WithField("fields", fields).
+		Debug("gitpod config analytics: processed")
+
+	a.prev = current
+}
+
+func (a *ConfigAnalyzer) computeFields(configs ...*gitpod.GitpodConfig) []string {
```
mustard-mh, line 82: This function will always returns all keys of `.gitpod.yml` since they are init-ed as a struct, i.e. pointer->nil string->"" int->0.

Or this is as expected?

akosyakov: I don't think it is true, at least I don't see it while testing. I can add more tests.

components/supervisor/pkg/supervisor/supervisor.go
mustard-mh: Yes, @akosyakov always said that once we use `GITPOD_*` env var [[1](https://github.com/gitpod-io/gitpod/pull/8313#discussion_r810007160)] 😂 

But I'm not sure if we have a chance to deprecate it since many users are using Gitpod's env vars. So maybe it's safe somehow? Anyway it must be best if we don't use it

components/supervisor/pkg/config/gitpod-config-analytics.go
mustard-mh: This is from the first test case

image

components/supervisor/pkg/supervisor/supervisor.go
akosyakov: there is somethere an issue to provide `gp info` instead and remove all `GITPOD_*` from docs

mustard-mh:
We only get the diffs for the first layer of fields, is it enough? Maybe `inner-loop` can re-use this `ConfigAnalyzer` but it required more detail of diff (i.e. tasks.?.name don't need to trigger, but tasks.?.command need)

andreafalzetti:
When the YAML file is invalid, additional events are fired, loom: https://www.loom.com/share/e140a1decc6c4019bbb41785ec280c72

mustard-mh:
> When the YAML file is invalid, additional events are fired, loom: https://www.loom.com/share/e140a1decc6c4019bbb41785ec280c72

Yes, it treated this object as nil 🙈, so diff from the previous one, will return all previous fields.

akosyakov:
> We only get the diffs for the first layer of fields, is it enough? Maybe inner-loop can re-use this ConfigAnalyzer but it required more detail of diff (i.e. tasks.?.name don't need to trigger, but tasks.?.command need)

It is an idea to have deep structural analysis, but with another PR.

sagor999 (approved):
reviewed one file that workspace team is code owner of and it LGTM :smile:

geropl (approved):

akosyakov:
/hold

because it does not handle invalid files well and it turned out we have a bug generally in gitpod config watching, i.e. fire nil config for invalid

components/gitpod-protocol/go/gitpod-config-types.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Gitpod GmbH. All rights reserved.
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
```
akosyakov, line 1: This file auto-generated. I added git hook which regenerates it on commit, see [.pre-commit-config.yaml](https://github.com/gitpod-io/gitpod/pull/11653/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9)

components/gitpod-protocol/data/gitpod-schema.json
```diff
@@ -261,140 +261,20 @@
...
-                                    "description": "Whether only stable, latest or both versions should be warmed up. Default is stable only."
-                                }
-                            }
-                        },
-                        "vmoptions": {
-                            "type": "string",
-                            "description": "Configure JVM options, for instance '-Xmx=4096m'."
-                        }
-                    }
+                    "$ref": "#/definitions/jetbrainsProduct",
```
akosyakov, line 264: Refactored to extract a common properties to the definition and stabilize code generation.

felladrin:
ℹ️ Reviewing it now.

components/supervisor/pkg/supervisor/supervisor.go
```diff
@@ -1513,82 +1512,39 @@ func socketActivationForDocker(ctx context.Context, wg *sync.WaitGroup, term *te
...
-				continue
+			if analyzer != nil {
+				analyzer.Analyse(cfg)
+			} else {
+				analyzer = config.NewConfigAnalyzer(log.Log, 5*time.Second, func(field string) {
+					w.Track(analytics.TrackMessage{
+						Identity: analytics.Identity{UserID: info.Workspace.OwnerID},
+						Event:    "gitpod_config_changed",
+						Properties: map[string]interface{}{
+							"key":         field,
```
felladrin, line 1540: **Observation (non-blocking)**: Opening a workspace without .gitpod.yml and then running `gp init` triggered 6 `gitpod_config_changed` at once, one for each of the keys:
- MainConfiguration
- WorkspaceLocation
- Ports
- CheckoutLocation
- Tasks
- ExperimentalNetwork

image

The content of the generated .gitpod.yml was the following:

```yml
# List the start up tasks. Learn more https://www.gitpod.io/docs/config-start-tasks/
tasks:
  - init: echo 'init script' # runs during prebuild
    command: echo 'start script'

# List the ports to expose. Learn more https://www.gitpod.io/docs/config-ports/
ports:
  - port: 3000
    onOpen: open-preview
```

So I expected it would trigger just two events keys: "Tasks" and "Ports".

felladrin (approved):
Working as advertised 🎉 

I left just a comment about events triggered when the workspace has no `.gitpod.yml` and we run `gp init`, which is an edge case.

components/supervisor/pkg/supervisor/supervisor.go
akosyakov: rrr, it is a bug, let me have a look a bit later, what is going on

akosyakov:
> I left just a comment about events triggered when the workspace has no .gitpod.yml and we run gp init, which is an edge case.

let me verify first that it does not happen then a user creates an empty file as well, it does not look good

akosyakov:
@felladrin I pushed the fix when https://werft.gitpod-dev.com/job/gitpod-build-ak-identiy-config-changed.38 is ready you can try again.

andreafalzetti (approved):
Great work @akosyakov 🙏 

It works as expected when the `.gitpod.yml` is invalid. I've noticed that when creating or deleting the whole file it does detect changes for the entire file but I'd say it's kind of expected and I don't consider it a bug.

akosyakov:
@loujaybee @andreafalzetti Do you think it would be useful to add an attribute like `operation` to distnguish what happened, i.e. `added`, `updated`,  `removed`? I looked right now at analytics it can be misleading, i.e. you can see 2 modifications to tasks, but final maybe was removing so actually tasks are not used. For analytics it maybe interesting to focus only on `added` and `updated`? It is rather an easy fix.

akosyakov:
> @loujaybee @andreafalzetti Do you think it would be useful to add an attribute like operation to distinguish what happened, i.e. added, updated, removed? I looked right now at analytics it can be misleading, i.e. you can see 2 modifications to tasks, but final maybe was removing so actually tasks are not used. For analytics it maybe interesting to focus only on added and updated? It is rather an easy fix.

Actually I'm wrong we cannot distinguish `removed` from `updated` because our goland types using string types not pointers. It will be quite impactful change, let's leave it for later.

PR #11654, opened 2022-07-26
Route GitpodWorkspaceStuckOnStarting to #t_workspace_alerts

ArthurSens:
## Description

Updates `GitpodWorkspaceStuckOnStarting` to #t_workspace_alerts.

Since alerts with no user impact should never go to PagerDuty, what is written in the runbook is already clear enough:
* [This issue has little to no user impact, but will result in on-call being paged regularly due to stuck nodes.](https://github.com/gitpod-io/runbooks/blob/e24540d28700262a4b8e8331640c1d2cea449c02/runbooks/GitpodWorkspaceStuckOnStarting.md?plain=1#L26)

Hey @kylos101, I'm coming back to your suggestion to re-classify the alert to something less severe.
https://github.com/gitpod-io/runbooks/pull/373#discussion_r929248324

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/operations/observability/mixins/workspace/rules/components/workspaces/alerts.libsonnet
+++ b/operations/observability/mixins/workspace/rules/components/workspaces/alerts.libsonnet
@@ -12,7 +12,8 @@
           {
             alert: 'GitpodWorkspaceStuckOnStarting',
             labels: {
-              severity: 'critical',
+              severity: 'warning',
+              team: 'workspace',
             },
             'for': '20m',
             annotations: {


kylos101 (changes requested):
@ArthurSens sorry, but we need to continue paging for this, or we pay for unnecessary pods and nodes longer than intended.

vulkoingim:
I agree with Arthur here - spending some $ for a few hours of a node running is better than paging people just so they can remove a node.

Having said that, I would go a step further - what is the reason for this alert at all? We know when a node is stuck (hence this alert) - why not automate the node removal?

kylos101 (approved):
:+1: thanks for routing this at workspace, @ArthurSens , I missed earlier and thought it was just a downgrade to warning from critical.

PR #11657, opened 2022-07-26
Automatically create Pull Requests for updating the Platform Version from JetBrains Plugins when available

### LINKED ISSUE #11120: Implement a Nightly GitHub Action to update JetBrains SDKs
## Task

Implment a Nightly GitHub Action that checks the latest available version of JetBrains Gateway SDK and creates a PR if it is different from the currently [used in the gateway-plugin](https://github.com/gitpod-io/gitpod/blob/f166459010778c6d40eab7fd9a44c126f181328d/components/ide/jetbrains/gateway-plugin/gradle.properties#L17). This should be done also [for the backend-plugin](https://github.com/gitpod-io/gitpod/blob/f166459010778c6d40eab7fd9a44c126f181328d/components/ide/jetbrains/backend-plugin/gradle-latest.properties#L9), checking for the IntelliJ SDK.

## Related

- https://github.com/gitpod-io/gitpod/pull/11117

felladrin:
## Description

Add a GitHub Action that runs every day checking if there's some update the Platform Version from JetBrains Plugins. If so, it create a Pull Request.

## Related Issue(s)

Resolves #11120

## How to test

- It won't be able to be tested manually, but I've ran it once by setting it to run on this PR, and from it we can see it's working as expected: When the version is up to date, nothing happens, but if it's outdated, a PR will be created:
  - https://github.com/gitpod-io/gitpod/runs/7546306836?check_suite_focus=true
    image
  - https://github.com/gitpod-io/gitpod/runs/7546306657?check_suite_focus=true (You can ignore the red icon in this case, as it just failed because it can't create a PR from another PR without specific configuration)
    image
- But you can read the PR changes and confirm if the code looks good.

## Release Notes

```release-note
NONE
```

## Documentation

None.

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/.github/workflows/jetbrains-update-plugin-platform-template.yml
+++ b/.github/workflows/jetbrains-update-plugin-platform-template.yml
@@ -1,0 +1,130 @@
+on:
+    workflow_call:
+        inputs:
+            pluginName:
+                description: Name of the plugin.
+                type: string
+                required: true
+            pluginId:
+                description: ID of the plugin in lowercase and without spaces.
+                type: string
+                required: true
+            xpath:
+                description: Xpath for the latest platform version in https://www.jetbrains.com/intellij-repository/snapshots
+                type: string
+                required: true
+            gradlePropertiesPath:
+                description: Path for the gradle.properties file of the plugin.
+                type: string
+                required: true
+        secrets:
+            slackWebhook:
+                required: true
+jobs:
+    update-plugin-platform:
+        name: Update Platform Version from ${{ inputs.pluginName }}
+        runs-on: ubuntu-latest
+        env:
+            SNAPSHOTS_HTML_FILENAME: snapshots.html
+        steps:
+            - name: Checkout Repository
+              uses: actions/checkout@v3
+            - name: Save the snapshots page to an HTML file
+              run: curl -sL https://www.jetbrains.com/intellij-repository/snapshots > ${{ env.SNAPSHOTS_HTML_FILENAME }}
+            - name: Get Current Platform Version
+              id: current-version
+              run: |
+                  CURRENT_VERSION=$(cat ${{ inputs.gradlePropertiesPath }} | grep platformVersion= | sed 's/platformVersion=//')
+                  echo "::set-output name=result::$CURRENT_VERSION"
+            - name: Extract Major Version from Current Platform Version
+              id: major-version
+              run: |
+                  MAJOR_VERSION=$(cut -c 1-3 <<< ${{ steps.current-version.outputs.result }})
+                  echo "Major Version from Current Platform Version: $MAJOR_VERSION"
+                  echo "::set-output name=result::$MAJOR_VERSION"
+            - name: Replace Major Version Placeholder
+              id: update-xpath
+              run: |
+                  UPDATED_XPATH=$(echo "${{ inputs.xpath }}" | sed 's/MAJOR_VERSION_PLACEHOLDER/${{ steps.major-version.outputs.result }}/')
+                  echo "Updated xpath: $UPDATED_XPATH"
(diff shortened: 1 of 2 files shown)


.github/workflows/jetbrains-update-plugin-platform.yml
```diff
@@ -0,0 +1,21 @@
...
+    schedule:
+        # At 11:00 on every day-of-week from Monday through Friday.
+        - cron: "0 11 * * 1-5"
+jobs:
+    update-backend-plugin-platform:
+        uses: ./.github/workflows/jetbrains-update-plugin-platform-template.yml
+        with:
+            pluginName: JetBrains Backend Plugin
+            pluginId: backend-plugin
+            xpath: "(/html/body/table[preceding::h2/text()='com.jetbrains.intellij.idea'][1]/tbody/tr/td[contains(text(),'-EAP-CANDIDATE-SNAPSHOT') and contains(text(),'222')]/text())[1]"
```
akosyakov, line 13: Why 222 hardcoded?

akosyakov:
Does it send a notification to our ide jobs channel if it fails?

.github/workflows/jetbrains-update-plugin-platform-template.yml
```diff
@@ -0,0 +1,76 @@
...
+              if: ${{ steps.latest-version.outputs.result != steps.current-version.outputs.result }}
+              uses: peter-evans/create-pull-request@v4
+              with:
+                  title: "Update Platform Version from ${{ inputs.pluginName }}"
+                  body: |
+                      ## Description
+                      This PR updates the Platform Version from ${{ inputs.pluginName }} to the latest version.
+
+                      ## How to test
+                      1. Ensure you have the latest JetBrains Gateway installed: https://www.jetbrains.com/remote-development/gateway/
```
akosyakov, line 59: Is it all relevant for backend plugin update? Do we really need to deal with backend plugin in this PR?

.github/workflows/jetbrains-update-plugin-platform.yml
felladrin: Good question! I actually forgot it there! Let me refactor it.
I'll extract the major version of the current platform and set it in a variable to use there as a guard statement to avoid creating PRs with major version regression (e.g. 222 -> 221) or upgrade (e.g. 222 -> 223).

felladrin:
> Does it send a notification to our ide jobs channel if it fails?

Well remembered. Going to add this step now.

.github/workflows/jetbrains-update-plugin-platform-template.yml
felladrin: I thought about making different "How to test" messages for each component. But then it would remove 50% of the usefulness of the Reusable Workflow. I don't think it would take much time if the reviewer tested both components when updating some of them.

But let me update it to have individual "How to test" per component.
(As I also have to fix the [hardcoded major version issue](https://github.com/gitpod-io/gitpod/pull/11657#discussion_r930321611), I'll review the need for the template.yml.)

.github/workflows/jetbrains-update-plugin-platform.yml
felladrin: Done!

.github/workflows/jetbrains-update-plugin-platform-template.yml
felladrin: Yeah, it was too much for testing the backend plugin.
I've updated it to have individual "How to test" per component now.
What do you think?

felladrin:
> Does it send a notification to our ide jobs channel if it fails?

Now it does :)

akosyakov (approved):
let's try

PR #11659, opened 2022-07-26
[usage] Fix BilledSession credits precision

easyCZ:
## Description

Currently, each BilledSession has Credits attached to it. However, we've been using int64 which is not appropriate, we need `float64` or `double` in proto terms.

## Related Issue(s)

Fixes #

## How to test

CI runs

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


roboquat:

<details>

</details>

components/usage-api/usage/v1/usage.proto
```diff
@@ -49,5 +49,6 @@ message BilledSession {
     google.protobuf.Timestamp start_time = 9;
     google.protobuf.Timestamp end_time = 10;
 
-    int64 credits = 11;
+    int64 credits_old = 11 [deprecated=true];   // insufficient precision to represent credits for workspace time
```
easyCZ, line 52: A rename is a valid backwards compatible operation (that's due to the field numbering).

components/usage-api/usage/v1/usage.proto
```diff
@@ -49,5 +49,6 @@ message BilledSession {
     google.protobuf.Timestamp start_time = 9;
     google.protobuf.Timestamp end_time = 10;
 
-    int64 credits = 11;
+    int64 credits_deprecated = 11 [deprecated=true];   // insufficient precision to represent credits for workspace time
```
easyCZ, line 52: A proto field rename is backwards compatible, because the names are only used for nice API to access them but under the hood all fields are accessed by their field number (ie 11).

andrew-farries: Is there a need to deprecate here? The API hasn't seen production use yet so can we just remove the old field?

easyCZ: It's being queried and we can't guarantee the order or the components we're rolling out. Could probs suffer some downtime but best we get into the habit of not breaking backwards compatibility

andrew-farries (approved):

PR #11666, opened 2022-07-27
[registry] allow insecure TLS for self signed cert

### LINKED ISSUE #11005: Resolve Inter Pod Communication When Using Self-Signed TLS Certificates

sagor999:
## Description

Allows registry and image builder to use insecure TLS when using self signed certs whose CA cannot be verified.

## Related Issue(s)

Fixes #11005 

## How to test

Update your installer.yaml with:
Specify valid 'customCACert` field in your installer yaml file, as well as:
```
containerRegistry:
  skipTLSVerify: true
```
Deploy your gitpod and verify that it is using self signed cert and that you are able to open the workspace, see image builder logs, and workspace can be shutdown. 
(you might need to wrangle with your browser to let it use insecure site)

## How I tested
Using workspace preview environment, I deployed gitpod using self signed cert (via cert manager) and verified that above test instructions succeed.

## Release Notes

```release-note
[registry] allow to use insecure TLS for self signed certs
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/image-builder-mk3/pkg/orchestrator/orchestrator.go
+++ b/components/image-builder-mk3/pkg/orchestrator/orchestrator.go
@@ -111,7 +111,7 @@ func NewOrchestratingBuilder(cfg config.Configuration) (res *Orchestrator, err e
 		censorship:    make(map[string][]string),
 		metrics:       newMetrics(),
 	}
-	o.monitor = newBuildMonitor(o, o.wsman)
+	o.monitor = newBuildMonitor(o, o.wsman, cfg.SkipTLSVerify)
 
 	return o, nil
 }
@@ -535,7 +535,7 @@ func (o *Orchestrator) checkImageExists(ctx context.Context, ref string, authent
 	defer tracing.FinishSpan(span, &err)
 	span.SetTag("ref", ref)
 
-	_, err = o.RefResolver.Resolve(ctx, ref, resolve.WithAuthentication(authentication))
+	_, err = o.RefResolver.Resolve(ctx, ref, resolve.WithAuthentication(authentication), resolve.WithSkipTLSVerify(o.Config.SkipTLSVerify))
 	if errors.Is(err, resolve.ErrNotFound) {
 		return false, nil
 	}
@@ -556,7 +556,7 @@ func (o *Orchestrator) getAbsoluteImageRef(ctx context.Context, ref string, allo
 		return "", status.Errorf(codes.InvalidArgument, "cannt resolve base image ref: %v", err)
 	}
 
-	ref, err = o.RefResolver.Resolve(ctx, ref, resolve.WithAuthentication(auth))
+	ref, err = o.RefResolver.Resolve(ctx, ref, resolve.WithAuthentication(auth), resolve.WithSkipTLSVerify(o.Config.SkipTLSVerify))
 	if xerrors.Is(err, resolve.ErrNotFound) {
 		return "", status.Error(codes.NotFound, "cannot resolve image")
 	}
--- a/components/image-builder-api/go/config/config.go
+++ b/components/image-builder-api/go/config/config.go
@@ -82,6 +82,9 @@ type Configuration struct {
 
 	// BuilderImage is an image ref to the workspace builder image
 	BuilderImage string `json:"builderImage"`
+
+	// SkipTLSVerify skips TLS verification in case if using self-signed certificate.
+	SkipTLSVerify bool `json:"skipTLSVerify,omitempty"`
 }
 
 type TLS struct {
--- a/components/image-builder-mk3/pkg/orchestrator/monitor.go
+++ b/components/image-builder-mk3/pkg/orchestrator/monitor.go
@@ -6,6 +6,7 @@ package orchestrator
 
 import (
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -36,12 +37,13 @@ type orchestrator interface {
 	PublishLog(buildID string, message string)
 }
 
-func newBuildMonitor(o orchestrator, wsman wsmanapi.WorkspaceManagerClient) *buildMonitor {
+func newBuildMonitor(o orchestrator, wsman wsmanapi.WorkspaceManagerClient, skipTLSVerify bool) *buildMonitor {
 	return &buildMonitor{
 		O:             o,
 		wsman:         wsman,
 		runningBuilds: make(map[string]*runningBuild),
 		logs:          map[string]context.CancelFunc{},
+		skipTLSVerify: skipTLSVerify,
 	}
 }
 
@@ -53,6 +55,8 @@ type buildMonitor struct {
 	runningBuildsMu sync.RWMutex
 
 	logs map[string]context.CancelFunc
+
+	skipTLSVerify bool
 }
 
 type runningBuild struct {
@@ -141,7 +145,7 @@ func (m *buildMonitor) handleStatusUpdate(status *wsmanapi.WorkspaceStatus) {
 		if _, ok := m.logs[status.Id]; !ok {
 			// we don't have a headless log listener yet, but need one
 			ctx, cancel := context.WithCancel(context.Background())
-			go listenToHeadlessLogs(ctx, bld.Logs.IdeURL, bld.Logs.OwnerToken, m.handleHeadlessLogs(status.Id))
+			go listenToHeadlessLogs(ctx, bld.Logs.IdeURL, bld.Logs.OwnerToken, m.handleHeadlessLogs(status.Id), m.skipTLSVerify)
 			m.logs[status.Id] = cancel
 		}
 	} else {
@@ -284,7 +288,7 @@ func (m *buildMonitor) RegisterNewBuild(buildID string, ref, baseRef, url, owner
 
 type listenToHeadlessLogsCallback func(content []byte, err error)
 
-func listenToHeadlessLogs(ctx context.Context, url, authToken string, callback listenToHeadlessLogsCallback) {
+func listenToHeadlessLogs(ctx context.Context, url, authToken string, callback listenToHeadlessLogsCallback, skipTLSVerify bool) {
 	var err error
 	defer func() {
 		if err != nil {
@@ -297,7 +301,7 @@ func listenToHeadlessLogs(ctx context.Context, url, authToken string, callback l
 		noTerminalErr = fmt.Errorf("no terminal")
 	)
 	err = retry(ctx, func(ctx context.Context) error {
-		logURL, err = findTaskLogURL(ctx, url, authToken)
+		logURL, err = findTaskLogURL(ctx, url, authToken, skipTLSVerify)
 		if err != nil {
 			return err
 		}
@@ -362,7 +366,7 @@ func listenToHeadlessLogs(ctx context.Context, url, authToken string, callback l
 	}
 }
 
-func findTaskLogURL(ctx context.Context, ideURL, authToken string) (taskLogURL string, err error) {
+func findTaskLogURL(ctx context.Context, ideURL, authToken string, skipTLSVerify bool) (taskLogURL string, err error) {
 	ideURL = strings.TrimSuffix(ideURL, "/")
 	tasksURL := ideURL + "/_supervisor/v1/status/tasks"
 	req, err := http.NewRequestWithContext(ctx, "GET", tasksURL, nil)
@@ -389,6 +393,13 @@ func findTaskLogURL(ctx context.Context, ideURL, authToken string) (taskLogURL s
 			return true, nil
 		}
 		return retryablehttp.DefaultRetryPolicy(ctx, resp, err)
+	}
+
+	if skipTLSVerify {
+		tr := &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		}
+		client.HTTPClient.Transport = tr
 	}
 
 	resp, err := client.StandardClient().Do(req)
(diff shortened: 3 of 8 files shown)


roboquat:

<details>

</details>

sagor999:
@gitpod-io/engineering-self-hosted can you please deploy this change via KOTS to verify it works in self hosted scenario as well.

MrSimonEmms:
I've updated this PR to include the KOTS changes, however this is still coming back with an error on the `imagebuilder` pod. When I run `kubectl describe pod -n gitpod imagebuild-6e81e077-4d3b-4a0e-9784-1dea0caaf6c5` we see the following:

```shell
Events:
  Type     Reason     Age              From               Message
  ----     ------     ----             ----               -------
  Normal   Scheduled  2s               default-scheduler  Successfully assigned gitpod/imagebuild-6e81e077-4d3b-4a0e-9784-1dea0caaf6c5 to gitpod-1
  Normal   Pulling    2s               kubelet            Pulling image "reg.gitpod-k3s.simonemms.com:20000/remote/6e81e077-4d3b-4a0e-9784-1dea0caaf6c5"
  Warning  Failed     2s               kubelet            Failed to pull image "reg.gitpod-k3s.simonemms.com:20000/remote/6e81e077-4d3b-4a0e-9784-1dea0caaf6c5": rpc error: code = Unknown desc = failed to pull and unpack image "reg.gitpod-k3s.simonemms.com:20000/remote/6e81e077-4d3b-4a0e-9784-1dea0caaf6c5:latest": failed to resolve reference "reg.gitpod-k3s.simonemms.com:20000/remote/6e81e077-4d3b-4a0e-9784-1dea0caaf6c5:latest": failed to do request: Head "https://reg.gitpod-k3s.simonemms.com:20000/v2/remote/6e81e077-4d3b-4a0e-9784-1dea0caaf6c5/manifests/latest": x509: certificate signed by unknown authority
  Warning  Failed     2s               kubelet            Error: ErrImagePull
  Normal   BackOff    1s (x2 over 2s)  kubelet            Back-off pulling image "reg.gitpod-k3s.simonemms.com:20000/remote/6e81e077-4d3b-4a0e-9784-1dea0caaf6c5"
  Warning  Failed     1s (x2 over 2s)  kubelet            Error: ImagePullBackOff
```

I've checked the `image-builder-mk3-config` ConfigMap and that does include the `skipTLSVerify: true` section in the `orchestrator`

```yaml
apiVersion: v1
data:
  image-builder.json: |-
    {
      "orchestrator": {
        "wsman": {
          "address": "ws-manager:8080",
          "tls": {
            "ca": "/wsman-certs/ca.crt",
            "crt": "/wsman-certs/tls.crt",
            "key": "/wsman-certs/tls.key"
          }
        },
        "pullSecret": "builtin-registry-auth",
        "pullSecretFile": "/config/pull-secret.json",
        "baseImageRepository": "registry.gitpod-k3s.simonemms.com/base-images",
        "workspaceImageRepository": "registry.gitpod-k3s.simonemms.com/workspace-images",
        "builderImage": "eu.gcr.io/gitpod-core-dev/build/image-builder-mk3/bob:c2d2b559411d806702e8151ba4e1cedf122ba923",
        "skipTLSVerify": true
      },
      "refCache": {
        "interval": "6h0m0s",
        "refs": [
          "docker.io/gitpod/workspace-full:latest"
        ]
      },
      "service": {
        "address": ":8080",
        "tls": {
          "ca": "",
          "crt": "",
          "key": ""
        }
      },
      "prometheus": {
        "address": "127.0.0.1:9500",
        "tls": {
          "ca": "",
          "crt": "",
          "key": ""
        }
      },
      "pprof": {
        "address": "127.0.0.1:6060"
      }
    }
kind: ConfigMap
metadata:
  annotations:
    meta.helm.sh/release-name: gitpod
    meta.helm.sh/release-namespace: gitpod
  creationTimestamp: "2022-07-28T07:13:37Z"
  labels:
    app: gitpod
    app.kubernetes.io/managed-by: Helm
    component: image-builder-mk3
  name: image-builder-mk3-config
  namespace: gitpod
  resourceVersion: "254742"
  uid: 11a0ecd6-411a-45a2-be15-3652c86d6440
```

I also tried adding `docker.io` into the `containerRegistry.privateBaseImageAllowList` array, but that made no difference (this may still be required later, but will eliminate for the time being).

MrSimonEmms:
@sagor999 yes I am. It sets:

```yaml
customCACert:
  kind: secret
  name: ca-issuer-ca
```

sagor999:
@MrSimonEmms this is due to the fact that containerd is not configured in KOTS environment to read certs from `/etc/containerd/certs.d`.  
So to make it work, you need to pass config file for containerd that includes this:
```
[plugins]
   [plugins."io.containerd.grpc.v1.cri".registry]
      config_path = "/etc/containerd/certs.d"
```

That is what we are doing for SaaS and workspace preview env as well:
https://github.com/gitpod-io/gitpod-packer-gcp-image/blob/553f2b4439087de50b01b3601f3803f75b76bc43/linux/containerd/containerd.toml#L50

MrSimonEmms:
@sagor999 blimey, that could well be a problem for _MANY_ people. KOTS doesn't have anything to do with the containerd configuration so this is all to do with the cluster setup.

I'll have a look and see how easy it is to change it. I'll also check managed k8s and see what that does

MrSimonEmms:
@sagor999 Ok, I've had a play with this and think this PR works in it's current state.

## Managed K8s

Using Azure as the template (I will check with AWS/GCP as well), this works as-is because it's already using `/etc/containerd/certs.d`. No further work is necessary.

## K3s

I had a play around with changing the certs directory as you mentioned above but, truthfully, that was awkward. So I tried by setting the `registry-facade` as a private registry without verifying the TLS - that works fine. I think that this is a perfectly acceptable compromise for anyone using k3s and wanting self-signed certs (spoiler - we have no paying customers we have to support doing this, but it's good to know we have a way around it).

I've updated my [k3s guide](https://github.com/MrSimonEmms/gitpod-k3s-guide/pull/10) so that the private registry is configured like this.

mrzarquon:
@MrSimonEmms AWS/EKS's base images don't have "/etc/containerd/certs.d" in the settings, so we need to perform the bootstrapoverride in eksctl.

MrSimonEmms:
It also doesn't work for GKE either disappointingly. So, as it stands:

- [x] k3s
- [x] Azure
- [ ] GCP
- [ ] AWS

sagor999:
@MrSimonEmms thank you for testing!
regarding GKE, I think we can reach out to our GCP support and see if there is a way to tweak containerd setup, though I suspect they might not allow it due to security/managed service. 

But even so, can the user still use the workaround that you mentioned regarding use our registry as private registry?

MrSimonEmms:
Honestly, I don't know. The private registry is a k3s-specific thing I believe, where it creates a mirror for the name registry - in this case, it seems to be doing a simple pass-through and bypassing the TLS verification.

As far as I know, that does not exist in GKE.

sagor999:
@MrSimonEmms My guess is then this PR fixes issues for self hosted environments that can control their containerd config. So at least it fixes some of the cases.
For GKE and AKS, for now, we can put a note that self signed cert is not supported in those environments.
I am going to open an issue with containerd folks, to see why that folder is not added by default, as that seems odd (but maybe there are reasons).

sagor999:
There is https://github.com/containerd/containerd/pull/6488 that does exactly that. Hopefully it will actually get merged sometime soon. :pray:

MrSimonEmms:
Just for clarification, AKS (Azure) works - it's EKS (AWS) that doesn't.

I guess Chris and I will have to have a conversation on Monday as to what he can support, although if it's not supported at a containerd level then I imagine we're largely forced into a corner here.

sagor999:
Alejandro shared this repo that does that for GKE (modifies containerd config):
https://github.com/boredabdel/gke-private-registries/
Article explaining code: https://medium.com/google-cloud/gke-and-private-registries-with-self-signed-certificates-b37b5fd1f982

So for GKE we could share this as a workaround for someone using self signed certs. :shrug:

csweichel:
> @MrSimonEmms thank you for testing! regarding GKE, I think we can reach out to our GCP support and see if there is a way to tweak containerd setup, though I suspect they might not allow it due to security/managed service.
> 

We already did that. This is the response:

> Without the workaround we would have to wait for the official feature and there is no clear ETA, yet.
> Several customers are also using https://github.com/samos123/gke-node-ca-importer (from another Google Engineer), which uses the same approach described above, but might be easier to test for you in a development or staging setup as the building blocks are ready to go.
> 
> If testing does not build up enough confidence to push this into production, then I am wondering if there is a chance to use a trusted CA for the registry instead.
> For example you are already using Let's Encrypt for [gitpod.io](http://gitpod.io/) and using their DNS01 challenge you could create a certificate without exposing a HTTP service publicly.
> On GCP you can leverage a split horizon DNS setup to expose verification records for Let's Encrypt publicly, but keep all the other records private.
Public CAs are forbidden from issuing certificates for non-public domains, so you'll have to use a public domain for your registry.

csweichel:
The main question IMHO is if we want to support the GKE-specific workaround in the installer/KOTS installation, or ask folks to make this happen another way.

@sagor999 @aledbf @MrSimonEmms WDYT?

components/image-builder-mk3/pkg/orchestrator/orchestrator.go
```diff
@@ -556,7 +556,7 @@ func (o *Orchestrator) getAbsoluteImageRef(ctx context.Context, ref string, allo
 		return "", status.Errorf(codes.InvalidArgument, "cannt resolve base image ref: %v", err)
 	}
 
-	ref, err = o.RefResolver.Resolve(ctx, ref, resolve.WithAuthentication(auth))
+	ref, err = o.RefResolver.Resolve(ctx, ref, resolve.WithAuthentication(auth), resolve.WithSkipTLSVerify(o.Config.SkipTLSVerify))
```
csweichel, line 559: This should not be necessary because we mount the custom CA cert in the image-builder container. Digging into why this is still required.

csweichel:
I think we really have two separate issues mixed here:
- how to make containerd talk to registry-facade on GKE when using custom certs [[1](https://github.com/gitpod-io/gitpod/pull/11666#issuecomment-1200861230)]
- how to make image-builder resolve image refs against a registry using custom certs (the original fix in this PR)

For the former, is there an issue we could move that discussion to (@MrSimonEmms)?

For the latter, the effect is rather odd considering that we're mounting the [custom cert in the image-builder container](https://github.com/gitpod-io/gitpod/blob/91d6ced645f965c7e9a475ed728091e2b835ccaf/install/installer/pkg/components/image-builder-mk3/deployment.go#L107-L110). To ensure that that's actually enough I set up an experiment. I ran a "local" HTTPS server with a custom cert and made image-builder try and pull from it.

🔴 First, without the custom CA cert installed. This behaves as expected, because the TLS chain is broken.
image

🟢 Then, with the custom CA cert installed. This also behaves as expected, because now the TLS chain is valid.
image

I restarted the image-builder after moving the cert in place because [Go reads the certificates at startup](https://go.dev/src/crypto/x509/root.go).

We will want to understand why the custom CA cert mount has no effect on image-builder.
@sagor999 I'd be more than happy to hop on a pair-debugging session.

MrSimonEmms:
@csweichel no, I don't believe there's a separate issue but it makes sense to separate the two issues

MrSimonEmms:
We'd quite like to get this PR in a position so that it can be merged so that we can make a hotfix release this week. Can @sagor999 and @adrienthebo have a conversation about this and see where we are and what else needs to be done please?

@mrzarquon as you're the (proxy to the) customer, can you confirm whether this PR fixes the immediate requirement we have (ie, fixes for k3s and Azure with a documented workaround for AWS and GCP) or whether additional work needs to happen to meet your needs please?

sagor999:
So I am quite confused now. :sweat_smile:
I just tested main.4121 build in workspace preview environment with CustomCA enabled.... and everything worked(!).
No issues with cert at all. 
@MrSimonEmms @adrienthebo could you test self hosted install using latest? 
image builder worked, logs streaming worked, and workspace opened (albeit with a million warnings from firefox about unsecure SSL connection due to unverified CA).

corneliusludmann:
_(converted to draft since based on the comment of Chris above this change seems not to be the desired solution)_

corneliusludmann:
Closing this since the initial issue cannot be reproduced. See also https://github.com/gitpod-io/gitpod/issues/11005#issuecomment-1205882107.

MrSimonEmms:
Closing this issue based in [internal discussion](https://gitpod.slack.com/archives/C01KLC56NP7/p1659699154988929?thread_ts=1659698938.907199&cid=C01KLC56NP7). Thanks for all your help on this (especially @sagor999 and @adrienthebo) but it looks like the specific customer issue was due to a configuration error on their part.

Until absolutely unavoidable, we should support loading of the correct trust mechanisms

PR #11676, opened 2022-07-27
[usage] Add ReconcileUsage RPC

easyCZ:
## Description

Adds RPC definition for Collecting usage

## Related Issue(s)

* Part of https://github.com/gitpod-io/gitpod/issues/10785

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


components/usage-api/usage/v1/usage.proto
```diff
@@ -9,6 +9,9 @@ import "google/protobuf/timestamp.proto";
 service UsageService {
     // ListBilledUsage retrieves all usage for the specified attributionId
     rpc ListBilledUsage(ListBilledUsageRequest) returns (ListBilledUsageResponse) {}
+
+    // CollectUsage collects usage for the specified time period, and stores the usage records in the database, returning the records.
+    rpc CollectUsage(CollectUsageRequest) returns (CollectUsageResponse) {}
```
andrew-farries, line 14: There is potential for confusion between this new RPC and the `ListBilledUsage` RPC above.

AIUI, ListBilledUsage retrieves usage records (within a time range, for an attribution id) that have already been computed.

This new RPC actually triggers the usage collection that will populate usage records that are then available to `ListBilledUsage`.

Maybe calling this one `StartUsageReconciliation` or something similar, to emphasise its imperative nature would be better?

easyCZ: In theory, calling either should be fine as the end result is going to be some (accurate) records in the DB. The main difference is one is a pure read, while the other is an Upsert in a way.

If the RPC is called `Start..`, then I would expect it's an async job which it is not. Perhaps `UpdateUsage`?

easyCZ: Or ReconcileUsage, as that's probs most accurate

andrew-farries (approved):
/hold for rename

components/usage-api/usage/v1/usage.proto
easyCZ: Renamed in https://github.com/gitpod-io/gitpod/pull/11676/commits/71191636830faafa78b802c6706db738259b8c47

PR #11681, opened 2022-07-27
[usage] Implement CollectUsage

easyCZ:
## Description

Moves the Usage Controller into an RPC called `CollectUsage`. This change gives us the following:
* Better decoupling between usage and billing
* Better observability through existing gRPC method monitoring
* More flexible structure to move services around
* Simplifies scheduled job to compute usage and update billing

## Related Issue(s)

* Part of https://github.com/gitpod-io/gitpod/issues/10785

## How to test

Unit tests

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview with-billing


components/usage/pkg/apiv1/usage.go
```diff
@@ -88,6 +97,212 @@ func (us *UsageService) ListBilledUsage(ctx context.Context, in *v1.ListBilledUs
...
+		nowFunc: time.Now,
+	}
+}
+
+type ReportGenerator struct {
+	conn    *gorm.DB
+	pricer  *controller.WorkspacePricer
+	nowFunc func() time.Time
+}
+
```
andrew-farries, line 196: Should we remove this `ReportGenerator` to it's own file now.? This `usage.go` is now rather large.

easyCZ: Sure, that's easy enough :)

andrew-farries (approved):
/hold for splitting out `UsageReport`

components/usage/pkg/apiv1/usage.go
easyCZ: Done and rebased. Also moved the pricer into `apiv1` as it didn't make sense in the controller anymore

Issue #11688, opened 2022-07-27
Usage report upload fails with GCS 403


easyCZ:
When the usage component attempts to use the content service to upload a usage report, it fails with the following error:

```
{"@type":"type.googleapis.com/google.devtools.clouderrorreporting.v1beta1.ReportedErrorEvent","error":"failed to upload usage report: failed to obtain signed upload URL: failed RPC to content service: rpc error: code = NotFound desc = cannot ensure storage exists: googleapi: Error 403: stag-meta-eu02-storage@gitpod-staging.iam.gserviceaccount.com does not have storage.buckets.get access to the Google Cloud Storage bucket., forbidden","level":"error","message":"Reconciliation run failed.","serviceContext":{"service":"usage","version":"commit-66578f1793ce828eaa28c604c087036411aeb646"},"severity":"ERROR","time":"2022-07-27T10:53:13Z"}
```

Investigate why, and what needs to happen to fix it.

andrew-farries:
Copying in my message from Slack:

There's two parts to this:

* GCloud storage bucket names have to be globally unique - we were using a hard-coded bucket name of usage-reports that is not unique. The bucket name will also need to vary between staging and production. I'm working on a PR against the installer and content service to do this.
* A unique bucket name allows the content-service to ensure the bucket exists, creating it if needed, but uploading a file to the bucket still fails with a 403 Forbidden . Once I've made the bucket name configurable I'll look into why that's happening.

andrew-farries:
The error from gcloud when trying to use the signed upload URL is:
```http
HTTP/1.1 403 Forbidden
Content-Length: 226
Content-Type: application/xml; charset=UTF-8
Date: Wed, 27 Jul 2022 13:59:44 GMT
Server: UploadServer
Vary: Origin
X-GUploader-UploadID: ADPycdvIzwf09QkyLRewUuK3kZag-1uXMv4-mQW_HcbVGPZ0lX9cx19lUkjTorUP6r6-V-B-vbtvr3A6VYDyBu2VVjdd

<?xml version='1.0' encoding='UTF-8'?>
<Error>
  <Code>AccessDenied</Code>
  <Message>Access denied.</Message>
  <Details>Anonymous caller does not have storage.objects.create access to the Google Cloud Storage object.</Details>
</Error>
```

[...]

PR #11689, opened 2022-07-27
[content-service] Make the name of the usage report bucket configurable

andrew-farries:
## Description

Making the `content-service` able to upload usage reports was done in https://github.com/gitpod-io/gitpod/pull/11519.

This PR makes the usage report bucket name configurable.

The name of the object storage bucket to which usage reports are uploaded needs to be configurable for Gitpod SaaS. GCloud bucket names are globally unique, so the bucket name must vary between staging and production.

## Related Issue(s)
#11688

## How to test

Unit tests for the installer.

Hit the `content-service` `UsageReport` service with a gRPC client to generate a signed upload URL that includes the name of the bucket (edit the `content-service` configmap to change the name of the bucket):

image

## Release Notes
```release-note
NONE
```

## Werft options:
- [x] /werft with-preview


components/content-service/pkg/service/usagereport-service.go
```diff
@@ -18,25 +18,22 @@ import (
...
-	usageReportBucketName = "usage-reports"
-)
-
 // UsageReportService implements UsageReportServiceServer
 type UsageReportService struct {
-	cfg config.StorageConfig
-	s   storage.PresignedAccess
+	cfg        config.StorageConfig
+	s          storage.PresignedAccess
+	bucketName string
```
jenting, line 25: Question: any reason that we did not put the bucketName variable under the StorageConfig data struct?

jenting (commented):
nit: suggest rename the file `usagereport-service.go` to `usage-report-service.go` _or_ `usage-report-service_test.go` to `usagereport-service_test.go`. It makes the unit test filename matches the source code file and is easy to read.

components/content-service/pkg/service/usagereport-service.go
andrew-farries: It could probably go there but it doesn't really fit as it's not related to credentials or accessing the storage in general; it's only relevant for the usage report service.

andrew-farries:
> nit: suggest rename the file usagereport-service.go to usage-report-service.go or usage-report-service_test.go to usagereport-service_test.go

Thanks, renamed in 17e7c40370685d5c517bd08fde880acd6c493bdf

jenting (approved):
Code LGTM.

csweichel (approved):

nandajavarma (approved):

Issue #11692, opened 2022-07-27
Implement BillingService.GetLatestInvoice


easyCZ:
Currently, when showing spending notifications to users we use the `UsageService.ListUsage` RPC and we sum up all the records. This is suboptimal, but more importantly it is the wrong data. Instead, we need to be looking at the current value of an Invoice (unpaid).

To that end, we want to add a new RPC to the `BillingService`. The RPC should be able to do the following:
* Given a UserID, or TeamID, it find details about the latest invoice and retrieves the current value of that invoce (along with more info if needed)

easyCZ:
[Related (internal) conversation.](https://gitpod.slack.com/archives/C02EN94AEPL/p1658925473327019)

PR #11693, opened 2022-07-27
[billing] GetLatestInvoice RPC protos

easyCZ:
## Description

Adds RPC definitions

Part of https://github.com/gitpod-io/gitpod/issues/11692

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


components/usage-api/usage/v1/billing.proto
```diff
@@ -23,3 +26,14 @@ message UpdateInvoicesRequest {
...
+message GetLatestInvoiceRequest {
+  string team_id = 1;
+  string user_id = 2;
+}
+
+message GetLatestInvoiceResponse {
+  string invoice_id = 1;
+  string currency = 2;
+  double amount = 3;
+}
```
andrew-farries, line 39: Do we also want to send the `team_id` or `user_id` back as part of the response too?

components/usage-api/usage/v1/billing.proto
```diff
@@ -23,3 +26,14 @@ message UpdateInvoicesRequest {
 
 message UpdateInvoicesResponse {
 }
+
+message GetLatestInvoiceRequest {
+  string team_id = 1;
+  string user_id = 2;
```
andrew-farries, line 32: Should this be a [oneof](https://developers.google.com/protocol-buffers/docs/proto3#oneof) field?

easyCZ: updated to one of

components/usage-api/usage/v1/billing.proto
easyCZ: I'd leave it out for now. The caller knows the user_id or team_id so I think for this case they can re-use that.

andrew-farries (approved):

PR #11694, opened 2022-07-27
[usage] Extend method call with order arg

### LINKED ISSUE #11691: Incorporate reverse-sorting arg from the API
The Usage API now accepts sorting as an argument. This should be incorporated in the server too.

laushinka:
## Description
Adds the ordering argument to `listBilledUsage` calls.
View implementation will be in a separate issue and PR.

## Related Issue(s)

Fixes #11691

## How to test
1. Join [team](https://laushinka-1e1c9f7bce.preview.gitpod-dev.com/teams/join?inviteId=5e187f1c-c628-4741-b4b0-3ad9d15cde44)
2. Go to [/usage](https://laushinka-1e1c9f7bce.preview.gitpod-dev.com/t/gitpod/usage)
3. Check the network call - `listBilledUsage` should be called with the additional param '0'.

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/dashboard/src/teams/TeamUsage.tsx
+++ b/components/dashboard/src/teams/TeamUsage.tsx
@@ -8,7 +8,7 @@ import { useContext, useEffect, useState } from "react";
 import { Redirect, useLocation } from "react-router";
 import { getCurrentTeam, TeamsContext } from "./teams-context";
 import { getGitpodService, gitpodHostUrl } from "../service/service";
-import { BillableSession, BillableWorkspaceType } from "@gitpod/gitpod-protocol/lib/usage";
+import { BillableSession, BillableWorkspaceType, SortOrder } from "@gitpod/gitpod-protocol/lib/usage";
 import { AttributionId } from "@gitpod/gitpod-protocol/lib/attribution";
 import { Item, ItemField, ItemsList } from "../components/ItemsList";
 import moment from "moment";
@@ -34,6 +34,7 @@ function TeamUsage() {
     const [startDateOfBillMonth, setStartDateOfBillMonth] = useState(timestampStartOfCurrentMonth);
     const [endDateOfBillMonth, setEndDateOfBillMonth] = useState(Date.now());
     const [isLoading, setIsLoading] = useState<boolean>(true);
+    const [startedTimeOrder] = useState<SortOrder>(SortOrder.Descending);
 
     useEffect(() => {
         if (!team) {
@@ -41,12 +42,14 @@ function TeamUsage() {
         }
         (async () => {
             const attributionId = AttributionId.render({ kind: "team", teamId: team.id });
+            const request = {
+                attributionId,
+                startedTimeOrder,
+                startDateOfBillMonth,
+                endDateOfBillMonth,
+            };
             try {
-                const billedUsageResult = await getGitpodService().server.listBilledUsage(
-                    attributionId,
-                    startDateOfBillMonth,
-                    endDateOfBillMonth,
-                );
+                const billedUsageResult = await getGitpodService().server.listBilledUsage(request);
                 setBilledUsage(billedUsageResult);
             } catch (error) {
                 if (error.code === ErrorCodes.PERMISSION_DENIED) {
@@ -56,7 +59,7 @@ function TeamUsage() {
                 setIsLoading(false);
             }
         })();
-    }, [team, startDateOfBillMonth, endDateOfBillMonth]);
+    }, [team, startDateOfBillMonth, endDateOfBillMonth, startedTimeOrder]);
 
     if (!showUsageBasedPricingUI) {
         return <Redirect to="/" />;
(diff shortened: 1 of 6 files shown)


components/dashboard/src/teams/TeamUsage.tsx
```diff
@@ -34,6 +34,7 @@ function TeamUsage() {
     const [startDateOfBillMonth, setStartDateOfBillMonth] = useState(timestampStartOfCurrentMonth);
     const [endDateOfBillMonth, setEndDateOfBillMonth] = useState(Date.now());
     const [isLoading, setIsLoading] = useState<boolean>(true);
+    const [startedTimeOrder] = useState(0); // Descending
```
andrew-farries, line 37: Can we use a smaller type rather than `number` here? ie a type like `'ascending' | 'descending'`.

components/gitpod-protocol/src/gitpod-service.ts
```diff
@@ -294,7 +294,12 @@ export interface GitpodServer extends JsonRpcServer<GitpodClient>, AdminServer,
     getSpendingLimitForTeam(teamId: string): Promise<number | undefined>;
     setSpendingLimitForTeam(teamId: string, spendingLimit: number): Promise<void>;
 
-    listBilledUsage(attributionId: string, from?: number, to?: number): Promise<BillableSession[]>;
+    listBilledUsage(
+        attributionId: string,
+        startedTimeOrder: number,
```
andrew-farries, line 299: Same question about the type. I know that the value sent in the RPC is ultimately an integer, but we should be able to use something more expressive before we convert to an integer to make the call.

andrew-farries (commented):
Looks good, just wonder whether we can do better with the type for the new parameter.

components/gitpod-protocol/src/gitpod-service.ts
laushinka: You're absolutely right. Will make the change. Thanks 🙌🏽

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2099,6 +2099,7 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
     async listBilledUsage(
         ctx: TraceContext,
         attributionId: string,
+        startedTimeOrder: number,
```
laushinka, line 2102: I didn't use the `SortOrder` type in the server because TypeScript kept complaining that it's expecting something else.

components/gitpod-protocol/src/gitpod-service.ts
```diff
@@ -294,7 +294,12 @@ export interface GitpodServer extends JsonRpcServer<GitpodClient>, AdminServer,
     getSpendingLimitForTeam(teamId: string): Promise<number | undefined>;
     setSpendingLimitForTeam(teamId: string, spendingLimit: number): Promise<void>;
 
-    listBilledUsage(attributionId: string, from?: number, to?: number): Promise<BillableSession[]>;
+    listBilledUsage(
+        attributionId: string,
+        startedTimeOrder: SortOrder,
+        from?: number,
+        to?: number,
```
AlexTugarev, line 301: @laushinka, please introduce `ListBilledUsageParams` and wrap the args, for 2 reasons:
1. it's likely this args list may be extended, right?
2. optional args at the end are simply not working with jsonrpc, because behind the scene there will be another param added (I think it's a cancellation token) which leads to nasty bugs in JS evaluation.

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -3210,6 +3210,7 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
     async listBilledUsage(
         ctx: TraceContext,
         attributionId: string,
+        startedTimeOrder: number,
```
andrew-farries, line 3213: I think the type here can also be updated to use `SortOrder`.

components/server/ee/src/workspace/gitpod-server-impl.ts
andrew-farries, line 2102: I think the type here can also be updated to use `SortOrder`.

components/gitpod-protocol/src/gitpod-service.ts
laushinka: > optional args at the end are simply not working with jsonrpc, because behind the scene there will be another param added (I think it's a cancellation token) which leads to nasty bugs in JS evaluation.

@AlexTugarev Is there a solution for this or can nothing be optional?

components/server/src/workspace/gitpod-server-impl.ts
laushinka: I wrote a comment [here](https://github.com/gitpod-io/gitpod/pull/11694#discussion_r934211465) that after I changed it to `SortOrder`, TS complained about expecting a different type.

components/server/ee/src/workspace/gitpod-server-impl.ts
andrew-farries: Here's a comparison that shows how it can be used:

https://github.com/gitpod-io/gitpod/compare/laushinka/incorporate-reverse-sorting-11691...af/temp

Does that work for you? Perhaps the TS error was due to not rebuilding after changing the `gitpod-protocol` to add the enum?

laushinka: Aha `startedTimeOrder as number` is the key! Thank you, @andrew-farries!

andrew-farries (approved):

PR #11699, opened 2022-07-27
Fix openvsx-proxy metrics

vulkoingim:
## Description

This adds the `kube-rbac-proxy` to the `openvxs-proxy` statefulset, exposing the metrics endpoint through it, so prometheus is able to scrape it correctly.

## How to test

[Before and after updating the address](https://grafana.gitpod.io/d/HNOvmGpxgd/openvsx-proxy?orgId=1&from=1658913689599&to=1658925254366&query=ws-proxy&var-datasource=VictoriaMetrics&var-cluster=All&var-node=All&var-pod=All&var-pod_instance=All&var-container_instance=All)

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/installer/pkg/components/openvsx-proxy/statefulset.go
+++ b/install/installer/pkg/components/openvsx-proxy/statefulset.go
@@ -132,7 +132,7 @@ func statefulset(ctx *common.RenderContext) ([]runtime.Object, error) {
 							Name:      "redis-data",
 							MountPath: "/data",
 						}},
-					},
+					}, *common.KubeRBACProxyContainer(ctx),
 					},
 				},
 			},


aledbf:
@vulkoingim  Instead of this change, I would suggest to add the kube-rbac-proxy to the [statefulset](https://github.com/gitpod-io/gitpod/blob/main/install/installer/pkg/components/openvsx-proxy/statefulset.go#L74) `*common.KubeRBACProxyContainer(ctx)}`

install/installer/pkg/components/openvsx-proxy/statefulset.go
```diff
@@ -132,7 +132,7 @@ func statefulset(ctx *common.RenderContext) ([]runtime.Object, error) {
 							Name:      "redis-data",
 							MountPath: "/data",
 						}},
-					},
+					}, *common.KubeRBACProxyContainer(ctx),
```
andreafalzetti, line 135: Just to understand how this works, we didn't have this param specified here before. How did this impact? Prometheus wasn't able to scrape the metrics at all?

vulkoingim: Heya, there was a bit of a longer PR description (guess you can still see it in the history), but trimmed it, since I changed the solution on advise of Alejandro.

Long-story short is that there was a breaking change introduced in: https://github.com/gitpod-io/gitpod/commit/14095f470baeb73e81cfc6bbfb622d7a3e3851f7

It changed the prometheus listen address from any (`0.0.0.0:9500`) to `127.0.0.1:9500`. Listening on 127.0.0.1 doesn't work because it is not accessible from outside of the container.

The rest of the services in the cluster use `kube-rbac-proxy` to proxy the metrics endpoint, which this PR adds. The `*common.KubeRBACProxyContainer(ctx)` is just a function that generates the container spec and we add it to the list of containers of the StatefulSet.

There's a longer discussion on Slack about that, if you are interested: https://gitpod.slack.com/archives/C01KGM9EBD4/p1658907591252089

andreafalzetti (approved):

PR #11717, opened 2022-07-28
Add openvsx alert

jeanp413:
## Description

Add alert when openvsx registry is down

runbook: https://github.com/gitpod-io/runbooks/pull/292

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/operations/observability/mixins/IDE/rules/components/openvsx-proxy/alerts.libsonnet
+++ b/operations/observability/mixins/IDE/rules/components/openvsx-proxy/alerts.libsonnet
@@ -1,0 +1,32 @@
+/**
+ * Copyright (c) 2021 Gitpod GmbH. All rights reserved.
+ * Licensed under the MIT License. See License-MIT.txt in the project root for license information.
+ */
+
+{
+  prometheusAlerts+:: {
+    groups+: [
+      {
+        name: 'gitpod-component-openvsx-proxy-alerts',
+        rules: [
+          {
+            alert: 'GitpodOpenVSXRegistryDown',
+            labels: {
+              severity: 'critical',
+            },
+            'for': '20m',
+            annotations: {
+              runbook_url: 'https://github.com/gitpod-io/runbooks/blob/main/runbooks/GitpodOpenVsxRegistryDown.md',
+              summary: 'Open-VSX registry is possibly down',
+              description: 'Open-VSX registry is possibly down. We cannot pull VSCode extensions we don\'t have in our caches',
+            },
+            expr:
+              |||
+                sum(rate(gitpod_openvsx_proxy_requests_total{status=~"5..|error"}[5m])) / sum(rate(gitpod_openvsx_proxy_requests_total[5m])) > 0.01
+              |||,
+          },
+        ],
+      },
+    ],
+  },
+}
--- a/operations/observability/mixins/IDE/rules/components/openvsx-proxy/rules.libsonnet
+++ b/operations/observability/mixins/IDE/rules/components/openvsx-proxy/rules.libsonnet
@@ -3,5 +3,9 @@
  * Licensed under the MIT License. See License-MIT.txt in the project root for license information.
  */
 
-(import './openvsx-proxy/alerts.libsonnet') +
-(import './openvsx-proxy/rules.libsonnet')
+{
+  prometheusRules+:: {
+    groups+: [],
+    // There is no recording rules for this component.
+  },
+}


operations/observability/mixins/IDE/rules/components/openvsx-proxy/alerts.libsonnet
```diff
@@ -0,0 +1,34 @@
...
+            },
+            'for': '20m',
+            annotations: {
+              runbook_url: 'https://github.com/gitpod-io/runbooks/blob/main/runbooks/GitpodOpenVsxRegistryDown.md',
+              summary: 'Open-VSX registry is possibly down',
+              description: 'Open-VSX registry is possibly down. We cannot pull VSCode extensions we don\'t have in our caches',
+            },
+            expr:
+              |||
+                sum(rate(gitpod_openvsx_proxy_requests_total{status=~"5\\d\\d|error"}[5m])) > 0.5
```
akosyakov, line 25: we need to alert on error ration not rate i.e. something like
```
sum(
  rate(
    gitpod_openvsx_proxy_requests_total{status=~"5\\d\\d|error"}[5m]
  )
) /
sum(
  rate(
    gitpod_openvsx_proxy_requests_total[5m]
  )
)
```

Error rate is misleading, i.e. seen it spikes with growth of throughput. Ratio stays stable.

Why 0.5? It is quite critical, it should be around 99% successful, so if it is over 1% is already bad, i.e `> 0.01`

akosyakov: You can read up about alerting on SLIs in https://www.oreilly.com/library/view/sre-with-java/9781492073918/ch04.html Look for `Service Level Indicators for Every Java Microservice`. The book about Java, but it is not really about Java 😆

operations/observability/mixins/IDE/rules/components/openvsx-proxy/alerts.libsonnet
```diff
@@ -0,0 +1,34 @@
...
+                OR
+                histogram_quantile(0.75, sum(rate(gitpod_openvsx_proxy_duration_upstream_call_seconds_bucket[5m])) by(le)) > 8
```
akosyakov, line 27: I'm not sure about how to do it properly. I think just alerting on errors should be enough for now. Let's drop it.

operations/observability/mixins/IDE/rules/components/openvsx-proxy/alerts.libsonnet
```diff
@@ -0,0 +1,32 @@
...
+    groups+: [
+      {
+        name: 'gitpod-component-openvsx-proxy-alerts',
+        rules: [
+          {
+            alert: 'GitpodOpenVSXRegistryDown',
+            labels: {
+              severity: 'critical',
+            },
+            'for': '20m',
```
akosyakov, line 17: Do you think it is necessary? It seem in other alerts we don't do it. Would it be good to know it immediately? if 0.01 is too noisy we can increase it?

jeanp413: > Do you think it is necessary? It seem in other alerts we don't do it.

Do you mean to decrease this value, I think 20 min is fine to avoid any false positives alerts like [this one](https://grafana.gitpod.io/d/HNOvmGpxgd/openvsx-proxy?orgId=1&from=1658949649687&to=1658955872665)
I don't see the benefit to have an alert immediately if the only thing we can do is just tell eclipse folks to take a look and cross fingers

akosyakov:
@ArthurSens Could you review please? It will be good start for us.

operations/observability/mixins/IDE/rules/components/openvsx-proxy/alerts.libsonnet
akosyakov: I thought [5m] interval already counts it in? But better to ask @ArthurSens

operations/observability/mixins/IDE/rules/components/openvsx-proxy/alerts.libsonnet
```diff
@@ -0,0 +1,32 @@
...
+            annotations: {
+              runbook_url: 'https://github.com/gitpod-io/runbooks/blob/main/runbooks/GitpodOpenVsxRegistryDown.md',
+              summary: 'Open-VSX registry is possibly down',
+              description: 'Open-VSX registry is possibly down. We cannot pull VSCode extensions we don\'t have in our caches',
+            },
+            expr:
+              |||
+                sum(rate(gitpod_openvsx_proxy_requests_total{status=~"5\\d\\d|error"}[5m])) / sum(rate(gitpod_openvsx_proxy_requests_total[5m])) > 0.01
```
ArthurSens, line 25:
```suggestion
                sum(rate(gitpod_openvsx_proxy_requests_total{status=~"5..|error"}[5m])) / sum(rate(gitpod_openvsx_proxy_requests_total[5m])) > 0.01
```

I'm not sure `\\d\\d` works as expected with prometheus, I'm sure `..` works though

ArthurSens (approved):
Besides my review, it should be good to go!

I added the hold just so you can address the review, feel free to remove and get this PR merged afterward :)

operations/observability/mixins/IDE/rules/components/openvsx-proxy/alerts.libsonnet
ArthurSens: [5m] Is the amount of data that `rate()` will use to calculate the increase rate.

It can be that we see a huge spike that makes a rate over 5minutes to be higher than 0.01, even though it was only for 30 seconds.

The for in the alert makes sure that the rate was continuously higher than 0.01 for the entire duration, so will avoid alerts during strong but short spikes

akosyakov (approved):
Let’s try

PR #11719, opened 2022-07-28
Revert the ws-daemon max unavailable back to wsdaemonMaxAttempts

jenting:
## Description

Give the ws-daemon unavailable retry maximum be 2 mins.
We see in the production cluster, that once the ws-daemon is gone, probably due to the underlying node being gone.
We don't need to waste time to keep retrying it for 1 hour in such a case.

Somehow, the finalizeWorkspaceContent executes takes 3h14m, it not what we expected within 1 hour.
image

## Related Issue(s)

Fixes #

## Release Notes

```release-note
None
```

## Documentation

None

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/ws-manager/pkg/manager/monitor.go
+++ b/components/ws-manager/pkg/manager/monitor.go
@@ -1244,10 +1244,7 @@ func (m *Monitor) finalizeWorkspaceContent(ctx context.Context, wso *workspaceOb
 	)
 	t := time.Now()
 
-	// since we're within the content backup loop, the maximum retry time should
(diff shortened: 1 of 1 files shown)


utam0k:
How about checking if the underlying node is existing?

jenting:
> How about checking if the underlying node is existing?

Well, I want to decrease the time that ws-manager keeps in finalizeWorkspaceContent over 3 hours.

aledbf (approved):

Issue #11725, opened 2022-07-28
WaitForContainer timeout during scale-up event


aledbf:
### Bug description

![Screenshot from 2022-07-28 09-08-05](https://user-images.githubusercontent.com/161571/181513031-194a487d-e001-44e6-9cde-5b7f42b55860.png)

### Steps to reproduce

-

### Expected behavior

The timeout should be set with the StartWorkspace timeout.

PR #11726, opened 2022-07-28
[ws-daemon] Increase WaitForContainer timeout

aledbf:
## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/11725

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/ws-daemon/pkg/container/containerd.go
+++ b/components/ws-daemon/pkg/container/containerd.go
@@ -306,6 +306,7 @@ func (s *Containerd) handleNewTask(cid string, rootfs []*types.Mount, pid uint32
 func (s *Containerd) WaitForContainer(ctx context.Context, workspaceInstanceID string) (cid ID, err error) {
 	//nolint:ineffassign
 	span, ctx := opentracing.StartSpanFromContext(ctx, "WaitForContainer")
+	span.LogKV("workspaceInstanceID", workspaceInstanceID)
 	defer tracing.FinishSpan(span, &err)
 
 	rchan := make(chan ID, 1)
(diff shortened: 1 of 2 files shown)


components/ws-daemon/pkg/container/containerd.go
```diff
@@ -306,6 +306,7 @@ func (s *Containerd) handleNewTask(cid string, rootfs []*types.Mount, pid uint32
 func (s *Containerd) WaitForContainer(ctx context.Context, workspaceInstanceID string) (cid ID, err error) {
 	//nolint:ineffassign
 	span, ctx := opentracing.StartSpanFromContext(ctx, "WaitForContainer")
+	span.LogKV("workspaceInstanceID", workspaceInstanceID)
```
jenting, line 309: Can we consider use this function `tracing.ApplyOWI(span, log.OWI("", "", workspaceInstanceID)`?

components/ws-daemon/pkg/dispatch/dispatch.go
```diff
@@ -199,7 +199,8 @@ func (d *Dispatch) handlePodUpdate(oldPod, newPod *corev1.Pod) {
 			},
 		}
 
-		waitForPodCtx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+		// Important!!!!: ideally this timeout must be equal to ws-manager https://github.com/gitpod-io/gitpod/blob/main/components/ws-manager/pkg/manager/manager.go#L171
```
jenting, line 202: Could use change this URL? Because it's persistent.

https://github.com/gitpod-io/gitpod/blob/2d898bf91a7962ad0e707ef82ff85c006626d2c1/components/ws-manager/pkg/manager/manager.go#L171

components/ws-daemon/pkg/container/containerd.go
aledbf: In any other place, I would say yes, but here we only have one variable. Keeping

components/ws-daemon/pkg/dispatch/dispatch.go
aledbf: This will be correct once we merge the change

jenting (approved):

PR #11730, opened 2022-07-28
[usage] Ensure `Content-Type` headers are set for GCP object storage uploads

andrew-farries:
## Description

In order to be able to upload these reports to GCP Cloud Storage, the signed URL can be created for a specific `Content-Type`; only `PUT` requests to that URL that set the same `Content-Type` header will be accepted.

This PR sets the `Content-Type` header when generating the signed URL and sets it on the `usage` component `PUT` to that URL.

Minio object storage has no such requirements on the `Content-Type` header which is why we only hit this problem in staging/production.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/11688

## How to test

This is difficult to test as preview uses minio for object storage, not GCP.

Tested by taking the images built for this branch and hot-patching the images for `content-service` and `usage` in staging.

Reports are uploaded to the bucket:

image

## Release Notes
```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/content-service/pkg/service/usage-report-service.go
+++ b/components/content-service/pkg/service/usage-report-service.go
@@ -48,7 +48,7 @@ func (us *UsageReportService) UploadURL(ctx context.Context, req *api.UsageRepor
 	}
 
 	info, err := us.s.SignUpload(ctx, us.bucketName, req.Name, &storage.SignedURLOptions{
-		ContentType: "*/*",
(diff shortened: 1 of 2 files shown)


aledbf (approved):

AlexTugarev (approved):
Déjà-vu

Issue #11743, opened 2022-07-29
Race condition at usage component startup


andrew-farries:
There is a race condition at usage component startup between the billing gRPC service being available and the billing client trying to connect to it:

```
{"config":{"controllerSchedule":"10s","creditsPerMinuteByWorkspaceClass":{"default":0.1666666667,"gitpodio-internal-xl":0.3333333333},"server":{"services":{"grpc":{"address":":9001"}}}},"level":"info","message":"Starting usage component.","serviceContext":{"service":"usage","version":""},"severity":"INFO","time":"2022-07-29T08:43:30Z"}
{"level":"info","message":"Starting usage controller.","serviceContext":{"service":"usage","version":""},"severity":"INFO","time":"2022-07-29T08:43:30Z"}
{"component":"grpc","level":"warning","message":"2022/07/29 08:43:30 WARNING: [core] [Channel #2 SubChannel #3] grpc: addrConn.createTransport failed to connect to {","serviceContext":{"service":"usage","version":""},"severity":"WARNING","time":"2022-07-29T08:43:30Z"}
{"component":"grpc","level":"warning","message":"  \"Addr\": \":9001\",","serviceContext":{"service":"usage","version":""},"severity":"WARNING","time":"2022-07-29T08:43:30Z"}
{"component":"grpc","level":"warning","message":"  \"ServerName\": \"localhost:9001\",","serviceContext":{"service":"usage","version":""},"severity":"WARNING","time":"2022-07-29T08:43:30Z"}
{"component":"grpc","level":"warning","message":"  \"Attributes\": null,","serviceContext":{"service":"usage","version":""},"severity":"WARNING","time":"2022-07-29T08:43:30Z"}
{"component":"grpc","level":"warning","message":"  \"BalancerAttributes\": null,","serviceContext":{"service":"usage","version":""},"severity":"WARNING","time":"2022-07-29T08:43:30Z"}
{"component":"grpc","level":"warning","message":"  \"Type\": 0,","serviceContext":{"service":"usage","version":""},"severity":"WARNING","time":"2022-07-29T08:43:30Z"}
{"component":"grpc","level":"warning","message":"  \"Metadata\": null","serviceContext":{"service":"usage","version":""},"severity":"WARNING","time":"2022-07-29T08:43:30Z"}
[...]

PR #11747, opened 2022-07-29
[usage] Delay gRPC self connection

andrew-farries:
## Description

The billing client tries to connect to the usage component gRPC server before the gRPC server is ready. This is a non-fatal error as gRPC dialing has built-in backoff/retry logic but it does produce ugly warnings in the usage component logs.

Here we use a simple solution to delay the self-connection attempt for a short period to give the gRPC server time to start.

## Related Issue(s)
Fixes https://github.com/gitpod-io/gitpod/issues/11743

## How to test

Warning messages at `usage` component startup (as described in #11743) are no longer present.

## Release Notes
```release-note
NONE
```

## Werft options:
- [ ] /werft with-preview

### DIFF
--- a/components/usage/pkg/server/dialer.go
+++ b/components/usage/pkg/server/dialer.go
@@ -1,0 +1,25 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package server
+
+import (
+	"context"
+	"net"
+	"time"
+
+	"google.golang.org/grpc"
+)
+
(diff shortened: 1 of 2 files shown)


easyCZ:
How does one know what delay is sufficient? I think possibly instead of this passing in a string address of the target service and the service establishes the connection the first time a request arrives may be easier to to deal with

andrew-farries:
> How does one know what delay is sufficient?

It doesn't. There is no synchronization, just a 'long enough' one second delay to give the gRPC server some time to start before the client makes the first connection attempt.

> I think possibly instead of this passing in a string address of the target service and the service establishes the connection the first time a request arrives may be easier to to deal with.

Not sure I follow. You mean don't have the client connect at startup but somehow do the connection on demand?

easyCZ (approved):

PR #11766, opened 2022-08-01
Set KOTS/Replicated dependencies to latest

MrSimonEmms:
## Description

These have become very outdated. As a way of avoiding this in future, these are set to the latest. 

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,4 +1,4 @@
-image: eu.gcr.io/gitpod-core-dev/dev/dev-environment:aledbf-go.18
+image: eu.gcr.io/gitpod-core-dev/dev/dev-environment:sje-kots-latest-img.2
 workspaceLocation: gitpod/gitpod-ws.code-workspace
 checkoutLocation: gitpod
 ports:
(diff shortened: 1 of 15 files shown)


jenting (approved):

adrienthebo (approved):
Verification:

```
 kubectl-kots version
Replicated KOTS 1.78.0
```

Merge conflicts nonwithstanding, LGTM!

mads-hartmann (approved):
I'd personally prefer not to use latest and instead just bump the version to whatever is latest right now. I understand that we'll then slowly fall behind again but I'd prefer that over accidentally updating Replicated/KOTS the next time someone makes an unrelated change to the Dockerfile - but as this is self-hosted tooling that's your decision to make ☺️

Due to job protection, in order to test your new image in Werft you'll have to trigger a job manually (see related issue https://github.com/gitpod-io/ops/issues/3723)

andreafalzetti (approved):

MrSimonEmms:
@mads-hartmann yeah, normally I'd be a little reticent about using `latest` on things, but it's proving more painful having it on a (very old) version. This is primarily an experiment to see if using `latest` is better and we can always revert if not. The intention is indeed to let the image get rebuilt regularly so it keeps our tooling mostly up to date (we always recommend using the latest in the docs anyway)

I'll do the job trigger now 👍🏻

Issue #10937, opened 2022-06-27
Billing controller excludes WorkspaceInstances which are already included in previous invoices


easyCZ:
The billing controller will need to be given the full set of WorkspaceInstances (which the UsageController fetched) to perform it's run. This is because in a given billing period (May 1 - May 31), a user can have potentially a large number of subscriptions (not in parallel) - create then cancel and repeat. For the cancelled subscriptions, they will of course need to settle it but we must not charge them for any usage from a cancelled subscription (but within that month). And the Billing Controller is the only component which should have access to Stripe to retrieve this data.
In practice, the controller will need to exclude all WorkspaceInstances which match previously already billed WorkspaceInstances.

## Architecture

*System architecture*: 

![regular-operation](https://user-images.githubusercontent.com/8225907/183827307-940b6b2d-877d-4544-9c07-b53f4366ade4.png)

*Interactions with Stripe and the `content-service`*:

![invoice-created](https://user-images.githubusercontent.com/8225907/183827288-2d56eae6-7c4f-436d-afd8-b45c4607bfd2.png)

### Done
- [x] Set up a placeholder webhook endpoint on the public API (https://github.com/gitpod-io/gitpod/pull/11806)
- [x] Configure ingress to the webhook in preview and live environments. (https://github.com/gitpod-io/gitpod/pull/11866)
- [x] Ensure that the billing controller can exclude instances from Stripe invoices based on start date.
  - [x] https://github.com/gitpod-io/gitpod/pull/11882
  - [x] https://github.com/gitpod-io/gitpod/pull/11883
### Webhook registration
- [ ] Register webhook in Stripe ([Stripe docs](https://stripe.com/docs/webhooks/go-live))
- [ ] Add webhook signing secret as K8S secret and mount it into the `public-api` pod.
- [ ] Implement Stripe signature verification on the webhook. ([Stripe docs](https://stripe.com/docs/webhooks/signatures))
### Implement `FinalizeInvoice` RPC
- [x] Define `FinalizeInvoice` RPC on `usage` that the public api will invoke from the webhook handler. (https://github.com/gitpod-io/gitpod/pull/11949)
- [x] Connect webhook so that it invokes `FinalizeInvoice` RPC when an `invoice.finalized` event is received. (https://github.com/gitpod-io/gitpod/pull/11987)
- [ ] Retrieve Stripe invoice for the requested invoice id
- [ ] Download usage report from content-service (report to download is indicated by Stripe metadata)
[...]

PR #11776, opened 2022-08-01
[usage] Add a placeholder Stripe webhook

andrew-farries:
## Description

Add a placeholder [Stripe webhook](https://stripe.com/docs/webhooks) to the usage component. 

In later PRs, the webhook will receive `invoice.finalized` events from Stripe so that we can update instance usage records to show that they have been included in a particular invoice.

## Related Issue(s)

Part of #9036 and https://github.com/gitpod-io/gitpod/issues/10937

## How to test

* Install the[ `stripe` CLI](https://stripe.com/docs/stripe-cli) into the workspace and run `stripe login`.
* Port forward to the database in the preview environment:
```
```
* Run the `usage` component locally:
```
DB_USERNAME=gitpod DB_HOST=localhost DB_PORT=3306 DB_PASSWORD=<>  go run . run  
```
* Forward `stripe` events to the local webhook endpoint:
```
stripe listen --skip-verify --forward-to localhost:9002/webhook
```
* Trigger a series of Stripe events:
```
stripe trigger invoice.finalized
```
This should start a series of Stripe events (setting up a customer and payment methods), each of which should be handled with a `200 OK`  by the webhook:

image

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/usage/pkg/server/server.go
+++ b/components/usage/pkg/server/server.go
@@ -138,5 +138,5 @@ func registerGRPCServices(srv *baseserver.Server, conn *gorm.DB, stripeClient *s
 }
 
 func registerHttpHandlers(srv *baseserver.Server, h *stripe.WebhookHandler) {
-	srv.HTTPMux().HandleFunc("/webhook", h.Handle)
+	srv.HTTPMux().HandleFunc("/stripe/invoices/webhook", h.Handle)
 }
--- a/components/usage/pkg/server/server.go
+++ b/components/usage/pkg/server/server.go
@@ -111,6 +111,9 @@ func Start(cfg Config) error {
 		return fmt.Errorf("failed to register gRPC services: %w", err)
 	}
 
+	h := stripe.NewWebhookHandler()
+	registerHttpHandlers(srv, h)
+
 	err = controller.RegisterMetrics(srv.MetricsRegistry())
 	if err != nil {
 		return fmt.Errorf("failed to register controller metrics: %w", err)
@@ -132,4 +135,8 @@ func registerGRPCServices(srv *baseserver.Server, conn *gorm.DB, stripeClient *s
 		v1.RegisterBillingServiceServer(srv.GRPC(), apiv1.NewBillingService(stripeClient))
 	}
 	return nil
+}
+
+func registerHttpHandlers(srv *baseserver.Server, h *stripe.WebhookHandler) {
+	srv.HTTPMux().HandleFunc("/webhook", h.Handle)
 }
--- a/components/usage/pkg/server/server.go
+++ b/components/usage/pkg/server/server.go
@@ -21,6 +21,7 @@ import (
 	"github.com/gitpod-io/gitpod/usage/pkg/controller"
 	"github.com/gitpod-io/gitpod/usage/pkg/db"
 	"github.com/gitpod-io/gitpod/usage/pkg/stripe"
+	"github.com/gorilla/handlers"
 	"gorm.io/gorm"
 )
 
@@ -138,5 +139,5 @@ func registerGRPCServices(srv *baseserver.Server, conn *gorm.DB, stripeClient *s
 }
 
 func registerHttpHandlers(srv *baseserver.Server, h *stripe.WebhookHandler) {
-	srv.HTTPMux().HandleFunc("/stripe/invoices/webhook", h.Handle)
+	srv.HTTPMux().Handle("/stripe/invoices/webhook", handlers.ContentTypeHandler(h, "application/json"))
 }
--- a/components/usage/pkg/stripe/webhook.go
+++ b/components/usage/pkg/stripe/webhook.go
@@ -19,7 +19,7 @@ func NewWebhookHandler() *WebhookHandler {
 	return &WebhookHandler{}
 }
 
-func (h *WebhookHandler) Handle(w http.ResponseWriter, req *http.Request) {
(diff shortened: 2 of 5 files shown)


andrew-farries:
I've put the webhook on the `usage` component which means we will need to set up ingress (via `proxy`) to allow the `usage` component to be reachable by Stripe. 

An alternative, would be to put the hook on `server` as we already have other webhook handling there. `usage` feels like the right place for this though.

components/usage/pkg/server/server.go
```diff
@@ -133,3 +136,7 @@ func registerGRPCServices(srv *baseserver.Server, conn *gorm.DB, stripeClient *s
 	}
 	return nil
 }
+
+func registerHttpHandlers(srv *baseserver.Server, h *stripe.WebhookHandler) {
+	srv.HTTPMux().HandleFunc("/webhook", h.Handle)
```
easyCZ, line 141: I'd recommend prefixing this with something like `/stripe/invoices/webhook`. This makes it easier to debug when you see just a request HTTP path without knowing what it does.

components/usage/pkg/stripe/webhook.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+type WebhookHandler struct{}
+
+func NewWebhookHandler() *WebhookHandler {
+	return &WebhookHandler{}
+}
+
+func (h *WebhookHandler) Handle(w http.ResponseWriter, req *http.Request) {
+	const maxBodyBytes = int64(65536)
```
easyCZ, line 24: I believe there' a middleware for this

components/usage/pkg/stripe/webhook.go
```diff
@@ -0,0 +1,45 @@
...
+		w.WriteHeader(http.StatusServiceUnavailable)
+		return
+	}
+
+	event := stripe.Event{}
+	if err := json.Unmarshal(payload, &event); err != nil {
+		log.WithError(err).Error("Stripe webhook error while parsing event payload")
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
```
easyCZ, line 39:
```
dec := json.Unmarshaller(http.Body)
if err := dec.Unmarshal(&event); err != nil { ... }
```

This should work. You shouldn't need to read all of the body into memory first and only then start unmarshalling. By using the above, you can stream unmarshal and avoid some buffering.

components/usage/pkg/stripe/webhook.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/stripe/stripe-go/v72"
+)
+
+type WebhookHandler struct{}
+
+func NewWebhookHandler() *WebhookHandler {
+	return &WebhookHandler{}
+}
+
+func (h *WebhookHandler) Handle(w http.ResponseWriter, req *http.Request) {
```
easyCZ, line 23: Also worth wrapping in middleware which rejects non `application/json` content type.

easyCZ:
> I've put the webhook on the usage component which means we will need to set up ingress (via proxy) to allow the usage component to be reachable by Stripe.

I'd avoid this. We don't have AuthN on the usage component so it shouldn't be directly exposed to the internet.

> An alternative, would be to put the hook on server as we already have other webhook handling there. usage feels like the right place for this though.

Do this, but proxy from Server to usage - server becomes the authN ingress, but you can keep logic in usage.

easyCZ:
Thinking about this, I'd actually implement the Usage part as a gRPC method which the `server` would delegate. I think in practice all we would need is something like `BillingService.FinalizeInvoice(..)`

andrew-farries:
> > I've put the webhook on the usage component which means we will need to set up ingress (via proxy) to allow the usage component to be reachable by Stripe.
> 
> I'd avoid this. We don't have AuthN on the usage component so it shouldn't be directly exposed to the internet.

Without AuthN on the `usage` gRPC methods, we could still expose just the webhook though right?

My preference for putting the webhook on the `usage` component comes simply from not wanting to load yet more functionality onto `server`.

components/usage/pkg/server/server.go
andrew-farries: b0d3d891b

components/usage/pkg/stripe/webhook.go
andrew-farries: I couldn't find any middleware that does exactly this. Did you have something in mind?

components/usage/pkg/stripe/webhook.go
andrew-farries: 766430278

components/usage/pkg/stripe/webhook.go
andrew-farries: 354cb8351

easyCZ:
> Without AuthN on the usage gRPC methods, we could still expose just the webhook though right?

That's fair, my worry is the added complexity where you now have to reason through some parts of it being public, while others not. It increases the threat surface for the system but worse it causes extra cognitive load on the engineers. The current (unfavourable) case where only `server` is exposed at least keeps the assumptions constant

easyCZ (approved):
I'm okay to go with this for now, to unblock you on other PRs but I'd like to understand how others feel about this congitive overhead as they will be the ones on-call for this.

/hold

andrew-farries:
Closing, in favour of https://github.com/gitpod-io/gitpod/pull/11806 that we've agreed that the public-api is a better home for the webhook

Issue #11579, opened 2022-07-22
Adjust usage attribution to prompt users when attribution is ambiguous


easyCZ:
In order to correctly attribute usage to either user or a team, we need to update our Workspace Instance attribution logic.

An `usageAttributionid` should always be set on a `WorkspaceInstance`, irrespective if the installation is SaaS or Self-hosted.

The logic to determine the `usageAttributionId` is as follows (in order of priority):
1. The user has made an explicit choice in their user settings and have configured the `User.usageAttributionId`.
2. The user is a member of exactly one entity from (user, team 1, team 2) which has billing enabled. If there is exactly such choice, then the attribution is not ambiguous and the one entity is used.
3. The choice is ambigous. In this case the user must be prompted to set the entity to which to attribute usage for the WorkspaceInstance currently being started. On the server, this should be reported as a typed error and the user should be prompted to specify the `usageAttributionId` as part of the WorkspaceInstance create request.

gtsiolis:
@easyCZ Is this related to a different problem to solve from https://github.com/gitpod-io/gitpod/issues/11498?

easyCZ:
@gtsiolis You've found this before I finished writing the description :)

It's related to the same problem. This focuses more on the backend logic but I think we can merge, tho the backend is likely needed a lot sooner than we can realistically do the UI changes.

geropl:
Note: Also need to update the WebApp docs!

jankeromnes:
On a side note: https://github.com/gitpod-io/gitpod/pull/11584 started implementing this change (by no longer setting the user attribution override in some cases). https://github.com/gitpod-io/gitpod/pull/11777 will continue and finish this work.

PR #11576, opened 2022-07-22
Stop sign on "out of credits"

AlexTugarev:
## Description

Partially resolves #11404

Very limited UI 🛹  of a modal shown when you've reached the configured spending limit.

Screen Shot 2022-08-03 at 14 07 02

## Related Issue(s)

Related to #11404

## Release Notes

```release-note
Add Spending Limit Reached modal on workspace creation.
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

PR #11777, opened 2022-08-01
When starting a workspace but usage attribution is unclear, prompt for explicit user choice

jankeromnes:
## Description

When starting a workspace but usage attribution is unclear, prompt for explicit user choice.

Possible unclear situations:
- User is a member of multiple teams with usage-based billing enabled
- User explicitly chose to attribute to a team, but they've since left this team (or the team was deleted, or its subscription was cancelled)

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/11579

## How to test

1. Start a workspace on any repo (should work)
2. Create a team called "Gitpod" (or "Gitpod 1", "Gitpod 2"... whichever is available) to enable the Usage-Based Billing flag
3. Start a new workspace on any repo (should work)
4. Upgrade your team to Usage-Based Billing by adding a payment method (e.g. 4242 4242 4242 4242, 4/24, 424)
5. Start a new workspace on any repo (should still work)
6. Create a new team (called anything you want)
7. Start a new workspace (should work)
8. Upgrade your second team to usage-based billing
9. Start a new workspace -- this time it should fail (unclear which team to bill) and ask you for an explicit choice
10. Select your second team -- this should retry the workspace start, and succeed
11. Delete your second team
12. Start a new workspace -- this time it should fail again (explicit choice is no longer valid) and ask you for a new choice

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/dashboard/src/components/BillingAccountSelector.tsx
+++ b/components/dashboard/src/components/BillingAccountSelector.tsx
@@ -1,0 +1,92 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { useContext, useEffect, useState } from "react";
+import { Team } from "@gitpod/gitpod-protocol";
+import { AttributionId } from "@gitpod/gitpod-protocol/lib/attribution";
+import { getGitpodService } from "../service/service";
+import { TeamsContext } from "../teams/teams-context";
+import { UserContext } from "../user-context";
+import SelectableCardSolid from "../components/SelectableCardSolid";
+import { ReactComponent as Spinner } from "../icons/Spinner.svg";
+
+export function BillingAccountSelector(props: { onSelected?: () => void }) {
+    const { user, setUser } = useContext(UserContext);
+    const { teams } = useContext(TeamsContext);
+    const [teamsWithBillingEnabled, setTeamsWithBillingEnabled] = useState<Team[] | undefined>();
+
+    useEffect(() => {
+        if (!teams) {
+            setTeamsWithBillingEnabled(undefined);
+            return;
+        }
+        const teamsWithBilling: Team[] = [];
+        Promise.all(
+            teams.map(async (t) => {
+                const subscriptionId = await getGitpodService().server.findStripeSubscriptionIdForTeam(t.id);
+                if (subscriptionId) {
+                    teamsWithBilling.push(t);
+                }
+            }),
+        ).then(() => setTeamsWithBillingEnabled(teamsWithBilling));
+    }, [teams]);
+
+    const setUsageAttributionTeam = async (team?: Team) => {
+        if (!user) {
+            return;
+        }
+        const usageAttributionId = AttributionId.render(
+            team ? { kind: "team", teamId: team.id } : { kind: "user", userId: user.id },
+        );
+        await getGitpodService().server.setUsageAttribution(usageAttributionId);
+        setUser(await getGitpodService().server.getLoggedInUser());
+        if (props.onSelected) {
+            props.onSelected();
+        }
+    };
+    return (
+        <>
+            {teamsWithBillingEnabled === undefined && <Spinner className="m-2 h-5 w-5 animate-spin" />}
+            {teamsWithBillingEnabled && (
+                <div>
+                    <p>Bill all my usage to:</p>
+                    <div className="mt-4 flex space-x-3">
+                        <SelectableCardSolid
+                            className="w-36 h-32"
+                            title="(myself)"
+                            selected={
+                                !teamsWithBillingEnabled.find(
+                                    (t) =>
+                                        AttributionId.render({ kind: "team", teamId: t.id }) ===
+                                        user?.usageAttributionId,
+                                )?.name
+                            }
+                            onClick={() => setUsageAttributionTeam(undefined)}
+                        >
+                            <div className="flex-grow flex items-end p-1"></div>
+                        </SelectableCardSolid>
+                        {teamsWithBillingEnabled.map((t) => (
+                            <SelectableCardSolid
+                                className="w-36 h-32"
+                                title={t.name}
+                                selected={
+                                    !!teamsWithBillingEnabled.find(
+                                        (t) =>
+                                            AttributionId.render({ kind: "team", teamId: t.id }) ===
+                                            user?.usageAttributionId,
+                                    )?.name
+                                }
+                                onClick={() => setUsageAttributionTeam(t)}
+                            >
+                                <div className="flex-grow flex items-end p-1"></div>
+                            </SelectableCardSolid>
+                        ))}
+                    </div>
+                </div>
+            )}
+        </>
+    );
+}
--- a/components/dashboard/src/start/CreateWorkspace.tsx
+++ b/components/dashboard/src/start/CreateWorkspace.tsx
@@ -25,6 +25,7 @@ import PrebuildLogs from "../components/PrebuildLogs";
 import CodeText from "../components/CodeText";
 import FeedbackComponent from "../feedback-form/FeedbackComponent";
 import { isGitpodIo } from "../utils";
+import { BillingAccountSelector } from "../components/BillingAccountSelector";
 
 export interface CreateWorkspaceProps {
     contextUrl: string;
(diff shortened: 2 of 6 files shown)


components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2008,20 +2008,6 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
-            const members = await this.teamDB.findMembersByTeam(teamId);
-            await Promise.all(
-                members.map(async (m) => {
-                    const u = await this.userDB.findUserById(m.userId);
-                    if (u && !u.usageAttributionId) {
-                        u.usageAttributionId = attributionId;
-                        await this.userDB.storeUser(u);
-                    }
-                }),
-            );
```
andrew-farries, line 2024: Why is this removed?

components/server/src/user/user-service.ts
```diff
@@ -229,12 +229,16 @@ export class UserService {
...
+                // TODO(janx): Throw error if selected attribution ID is no longer valid (e.g. user left team, team deleted, team cancelled)
+                return user.usageAttributionId;
             }
-            // Return the user's explicit attribution ID.
-            return user.usageAttributionId;
+            // TODO(janx):
+            //   - If single team with Usage-Based billing enabled, auto-attribute to that team
+            //   - If multiple teams with Usage-Based, throw error to prompt for explicit user choice
+            // No explicit user attribution ID yet -- attribute all usage to the user by default (regardless of project/team).
+            return AttributionId.render({ kind: "user", userId: user.id });
```
andrew-farries, line 241: There is no change here right? Just inverting the `if/else` and adding TODOs.

andrew-farries (commented):
PR description doesn't seem to match the code 😄 

AIUI this PR just removes the automatic attribution when a team signs up for Stripe.

jankeromnes:
Whoops, this was meant to be a Draft. 😅 Thanks for the quick review though @andrew-farries! 🙏

components/server/ee/src/workspace/gitpod-server-impl.ts
jankeromnes: The logic of this change is:
1. Stop setting/unsetting `User.usageAttributionId` automatically, but give full explicit control of this value to the user
2. When attributing instances on start, we already look at the explicit user attribution, but if none is set and attribution, or it's set to an invalid choice, we should either guess what the sensible non-ambiguous attribution should be, or throw an error in the backend and show a prompt in the front-end in order to ask the user what their explicit choice is

Step 1. was partly implemented by https://github.com/gitpod-io/gitpod/pull/11584, but this particular location was forgotten so I'm removing it now

Step 2. is to be implemented in this PR (the whole non-ambiguous logic, error case, and Modal shown to the user)

components/server/src/user/user-service.ts
```diff
@@ -229,12 +292,18 @@ export class UserService {
     async getWorkspaceUsageAttributionId(user: User, projectId?: string): Promise<string | undefined> {
         // A. Billing-based attribution
```
jankeromnes, line 293: TODO: Handle UBB flag off (fall back to Project-based attribution? or always to user?)

jankeromnes:
@gtsiolis If you have some time, could you please provide some advice on the UX? 🙂

I went with a very basic, non-dismissible Modal, but it's still rough around the edges:

| Starting a Workspace | User Billing |
| --- | --- |
| Screenshot 2022-08-03 at 15 05 47 | Screenshot 2022-08-03 at 15 06 43 |
| Screenshot 2022-08-03 at 15 06 09 | Screenshot 2022-08-03 at 15 06 26 |

gtsiolis:
Looking at this now! 👀

components/dashboard/src/components/BillingAccountSelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { getGitpodService } from "../service/service";
+import { TeamsContext } from "../teams/teams-context";
+import { UserContext } from "../user-context";
+import SelectableCardSolid from "../components/SelectableCardSolid";
+import { ReactComponent as Spinner } from "../icons/Spinner.svg";
+
+export function BillingAccountSelector(props: { onSelected?: () => void }) {
+    const { user, setUser } = useContext(UserContext);
+    const { teams } = useContext(TeamsContext);
+    const [teamsWithBillingEnabled, setTeamsWithBillingEnabled] = useState<Team[] | undefined>();
```
AlexTugarev, line 18: @jankeromnes, I bet this is not the single place, where we'd need this kind of info. have you considered promoting this flag to the result of getTeams? `team.isBillingEnabled` would be nice and it shave off the additional requests.

components/dashboard/src/components/BillingAccountSelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+                }
+            }),
+        ).then(() => setTeamsWithBillingEnabled(teamsWithBilling));
+    }, [teams]);
+
+    const setUsageAttributionTeam = async (team?: Team) => {
+        if (!user) {
+            return;
+        }
+        const usageAttributionId = team ? `team:${team.id}` : `user:${user.id}`;
```
AlexTugarev, line 40: hint: there is a `AttributionId.render`

components/dashboard/src/start/CreateWorkspace.tsx
```diff
@@ -185,6 +186,12 @@ export default class CreateWorkspace extends React.Component<CreateWorkspaceProp
                     phase = StartPhase.Stopped;
                     statusMessage = <LimitReachedOutOfHours />;
                     break;
+                case ErrorCodes.INVALID_COST_CENTER:
+                    // HACK: Hide the error (behind the modal)
+                    error = undefined;
+                    phase = StartPhase.Stopped;
+                    statusMessage = <SelectCostCenterModal onSelected={() => this.createWorkspace()} />;
```
AlexTugarev, line 193: `onSelected => createWorkspace()` doesn't seem to invalidate error state, I think this might delay the visual effect, e.g. the transition from the modal to next state. maybe we should clear the error state therein, or try to reload when selection was done.

components/gitpod-protocol/src/messaging/error.ts
```diff
@@ -32,15 +32,18 @@ export namespace ErrorCodes {
     // 430 Repository not whitelisted (custom status code)
     export const REPOSITORY_NOT_WHITELISTED = 430;
 
+    // 450 Payment error
+    export const PAYMENT_ERROR = 450;
+
+    // 455 Invalid cost center (custom status code)
+    export const INVALID_COST_CENTER = 455;
```
AlexTugarev, line 39: can we agree on a common prefix for everything payment related, please?

also see https://github.com/gitpod-io/gitpod/pull/11576/files#diff-3bb6d890c14ac42bf9860578a9f6a85ffd0121dbf980055baa99f8b7731c46daR45

and feel free to comment on that BTW!

alternative way to approach this, we could use just `PAYMENT_ERROR` (in similar cases) and encode specific conditions into the payload. it feels like we'll have the needs to extends for more error modes. wdyt?

components/dashboard/src/components/BillingAccountSelector.tsx
jankeromnes: @AlexTugarev I think this would be premature, because:
- I can't imagine other places that will need the same information currently (note: this billing account selector component is already reusable, and in fact it's already used in two places)
- I don't think it's wise to make `getTeams` load all the team subscriptions on every call, even when most callers don't need that currently

components/gitpod-protocol/src/messaging/error.ts
jankeromnes: I don't see much value in aligning every payment-related error around a single prefix or error code range (apart from readability of the error codes list -- which we arguably don't read that often).

Personally, I'd prefer to leave this as is (unless you have a strong opinion towards a different error name or code), and you can use whichever convention seems best in your PR. We can always align later if you think it's valuable, but I suspect that aligning these is not important in the grand scheme of things. 😇

jankeromnes:
Thanks for the reviews!

I guess this is now ready to be merged, unless you have any final requests. 🙂

components/server/src/user/user-service.ts
```diff
@@ -213,6 +216,66 @@ export class UserService {
...
+            if (!subscriptionId) {
+                throw new ResponseError(
+                    ErrorCodes.INVALID_COST_CENTER,
+                    "The billing team you've selected has no active subscription.",
+                );
+            }
+        }
+    }
+
+    protected async findSingleTeamWithUsageBasedBilling(user: User): Promise<Team | undefined> {
```
geropl, line 255: FYI: This can also come out of `BillingMode`: we already gather the data there, and only need to loop it through.

But that's improvement after both got merged.

components/server/src/user/user-service.ts
```diff
@@ -231,12 +294,18 @@ export class UserService {
     async getWorkspaceUsageAttributionId(user: User, projectId?: string): Promise<string | undefined> {
```
geropl, line 294: nit, and out-of-scope here: It feels like this method (and it's descendants) do not necessarily belong into UserService, but more into `usage/Attribution` or so. But that's sth for later. :upside_down_face:

geropl:
@jankeromnes testing this now...

geropl:
On re-start, the chooser is not shown, but instead I see the plain error:
image

I think that's what we discussed earlier, and part of a follow-up PR.

geropl (approved):
Code LGTM, tested and works :skateboard: 

[Error message](https://github.com/gitpod-io/gitpod/pull/11777#issuecomment-1206304264) is not nice, but does not harm prod, and can be a follow-up. :100:

PR #11798, opened 2022-08-02
[dashboard] Fix rendering of the redirect URI on Git Integrations page

### LINKED ISSUE #11795: Git Integrations shows wrong URI if port number is included
Follow-up issue https://github.com/gitpod-io/gitpod/issues/5305#issuecomment-1202148508

Reported by @jmls

TODO: `Git Integrations` modal should replace colon `:` with underscore `_` to meet the processing of the URI on server.

AlexTugarev:
## Description
Redirect URI is used to be copy-pasted into OAuth applications of GHE/GL/BBS. When port numbers are included in host names, this was not handled properly in #11409.

## Related Issue(s)

Fixes #11795

## How to test
Try to create a Git Integration with port number in hostname and see the value of the redirect URI is rendered with colon before the port number.

## Release Notes

```release-note
Fix rendering of the redirect URI on Git Integrations page
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/dashboard/src/settings/Integrations.tsx
+++ b/components/dashboard/src/settings/Integrations.tsx
@@ -643,6 +643,12 @@ export function GitIntegrationModal(
                 newHostValue = host.replace("https://", "");
             }
 
+            // Negative Lookahead (?!\/)
+            // `\/` matches the character `/`
+            // "https://foobar:80".replace(/:(?!\/)/, "_")
+            // => 'https://foobar_80'
+            newHostValue = host.replace(/:(?!\/)/, "_");
+
             setHost(newHostValue);
             setRedirectURL(callbackUrl(newHostValue));
             setErrorMessage(undefined);


andrew-farries (approved):

AlexTugarev:
@jmls, I don't think you need to wait for the next deployment to get unblocked. This PR fixes the rendering of the redirect URI to be entered in your OAuth app on GHE/BBS/GL server, if port number is used. 

There is a workaround. You can fix the copy-pasted URL easily yourself. Go to the provider specific OAuth app settings and replace the colon by an underscore, e.g. `.../auth/domain.tld:port/callback` –> `.../auth/domain.tld_port/callback`.

jmls:
Our customer said that they tried this and it didn't work. I'll have to check again with them

MaxKienle:
Hi everybody, just checked with our customer this fix: When entering a URL in the "New Git Integration" Screen with Port and : in the URL the (example foo.bar:7990), the URL will adjusted directly to foo.bar_7990. When trying to activate this connection, we get the error "Host could not be reached"

Could you help us out here?

MicrosoftTeams-image (1)

jmls:
@AlexTugarev the behaviour we are seeing is this:

As soon as we try and add the port to the hostname, the colon gets converted to a _. This makes the *hostname* invalid, and gipod then complains about the host could not be reached. Obviously, the host *does not* exist with the _

Please could someone actually try this ? It's obvious from the moment you try and press "Aactivaate Integration" that this is flawed and borken

This has been a problem for over 4 weeks now - it's causing major embarrassment with our client

AlexTugarev:
@AlexTugarev, thanks for the feedback! Let me fix this.

Issue #11715, opened 2022-07-28
[ws-daemon] Unexpected error during workspace dispose 


aledbf:
### Bug description

![Jaeger-UI (1)](https://user-images.githubusercontent.com/161571/181407168-9551f5a9-c459-4cf1-be8d-b063c0d1e573.png)

### Steps to reproduce

- n/a

### Expected behavior

Log the error when its encountered and include context (instance ID) so its easier to track down what happened. 

### Anything else?

This is just the first step for now, get some insight to whether this condition is happening, with related context, so we can react later.

csweichel:
This suggests multiple `Dispose` calls. When the `workspace.json` file is gone a second `Dispose` call should be a successful no-op.

kylos101:
@jenting reopened this issue and moved back to in-progress because of https://github.com/gitpod-io/gitpod/pull/12018

jenting:
We need to check the jaeger tracing on the gen63 cluster to see if it still happens or not.

Issue #11369, opened 2022-07-14
`ws-daemon` sometimes didn't delete workspace file after backup


iQQBot:
### Bug description

`ws-daemon` sometimes didn't delete workspace file after backup

### Steps to reproduce

see workspace folder in node `/var/gitpod/workspaces`

### Anything else?

The current mitigation for this is to `kubectl cordon` the node, but, it'd be ideal not to have this condition occur in the first place, this way the alert doesn't page on-callers.

This can cause alerts such as [NodeFilesystemAlmostOutOfSpace](https://gitpod.pagerduty.com/incidents/Q21LINWL2K6GGU)

aledbf:
Query for context of the frequency https://cloudlogging.app.goo.gl/FL1gFeddSjJF5Asm6

csweichel:
This is almost "by design". We [don't collect leftover content directories](https://github.com/gitpod-io/gitpod/blob/csweichel/ws-daemon-unexpected-11715/components/ws-daemon/pkg/internal/session/store.go#L244-L261), but [still assume that](https://github.com/gitpod-io/gitpod/blob/csweichel/ws-daemon-unexpected-11715/components/ws-daemon/pkg/internal/session/workspace.go#L203-L208) when deleting workspaces.

csweichel:
> Query for context of the frequency https://cloudlogging.app.goo.gl/FL1gFeddSjJF5Asm6

How does the failure to move a dangling base mount lead to leaking workspace content?

jenting:
> don't collect leftover content directories

👋 @csweiche you shared links can't be open.

csweichel:
> > don't collect leftover content directories
> 
> 👋 @csweiche you shared links can't be open.

Thanks for the hint - that means that "Copy Head Link" from within Gitpod is broken :/ 
I'll update the links

csweichel:
Working on a PR which reinstates the relevant code - updated to today's ws-daemon behaviour

kylos101:
@csweichel can you assign yourself to this issue, and change the status from scheduled to in-progress? 🙏

PR #11799, opened 2022-08-02
Ensure workspace is only deleted once

Furisto:
## Description
Takes care of the following problem:
- We attempt to delete the workspace and delete the workspace.json file
- Workspace deletion fails at a later step for some reason
- We do not remove the session
- We reattempt deletion but because the workspace.json has already been deleted we fail again due to the missing workspace.json file.

So lets remove the workspace session even if deleting the workspace from disk fails because we will definitely fail the second time and it just generates noise. We need to focus on why it fails in the first place.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/11715

## Release Notes

```release-note
None
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/ws-daemon/pkg/internal/session/store.go
+++ b/components/ws-daemon/pkg/internal/session/store.go
@@ -137,13 +137,13 @@ func (s *Store) Delete(ctx context.Context, name string) (err error) {
 	if !exists {
 		return nil
 	}
+	defer delete(s.workspaces, name)
 
 	err = session.Dispose(ctx)
 	if err != nil {
-		return xerrors.Errorf("cannot delete session: %w", err)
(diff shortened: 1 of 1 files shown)


utam0k (approved):

jenting:
> even if deleting the workspace from disk fails 

What's the side effect that deleting the workspace from the disk fails?
I _guess_ some of the data on the overlaying disk will be orphaned, correct?

Furisto:
> > even if deleting the workspace from disk fails
> 
> What's the side effect that deleting the workspace from the disk fails? I _guess_ some of the data on the overlaying disk will be orphaned, correct?

@jenting Yes. We can track this with https://github.com/gitpod-io/gitpod/issues/11369

PR #11800, opened 2022-08-02
[usage] Improve pagination behavior

### LINKED ISSUE #11530: Improve pagination layout, design, and structure
### Problem to solve

Following up from https://github.com/gitpod-io/gitpod/pull/11481, we need to update the pagination component. Some UX issues currently involved include:
1. Lack of proper disabled button states, see Previous and Next at the first or last page.
2. Lack of proper default, hover, and active states on buttons.
3. Lack of truncation when pages are more than what we can show.
4. ...

Screenshot 2022-07-21 at 3 20 44 PM

laushinka:
## Description
Improves the pagination behavior.
Note: This has a /hold label because it has a test commit that needs to be removed later.

| 1 | 2 | 3 |
| - | - | - |
| Screenshot 2022-08-03 at 16 26 44 | Screenshot 2022-08-03 at 16 26 54 | Screenshot 2022-08-03 at 16 27 01 |

## Related Issue(s)

Fixes #11530 

## How to test
This PR includes a file `dummy.ts` so that the pagination behavior when there are too many pages can be tested.
The number of results per page is also made smaller.
1. Join [this team](https://lau-pagina7108ace06a.preview.gitpod-dev.com/teams/join?inviteId=ccd62582-06a3-4fd3-928c-ad8159670a30).
2. Go to https://lau-pagina7108ace06a.preview.gitpod-dev.com/t/gitpod/usage.
3. Play around with the pagination.

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


roboquat:

<details>

</details>

laushinka:
Due to a [test commit](https://github.com/gitpod-io/gitpod/pull/11800/commits/5aabd213ffff3f1a909b9d057fb48268f7b89e6d):

/hold

andrew-farries:
Are you still able to access the preview for this? It doesn't work for me:

image

laushinka:
> Are you still able to access the preview for this? It doesn't work for me:

Let me re-run it - that usually fixes it.

/werft run

laushinka:
@andrew-farries All good now https://lau-pagina7108ace06a.preview.gitpod-dev.com/t/gitpod/usage

components/dashboard/src/teams/TeamUsage.tsx
```diff
@@ -32,7 +32,7 @@ function TeamUsage() {
     const team = getCurrentTeam(location, teams);
     const [billedUsage, setBilledUsage] = useState<BillableSession[]>([]);
     const [currentPage, setCurrentPage] = useState(1);
-    const [resultsPerPage] = useState(10);
+    const [resultsPerPage] = useState(1);
```
laushinka, line 35: I removed the dummy file and instead set this to 1 per page. Will change to `15` before merging.

andrew-farries:
Can you make me an owner of the team? Otherwise: 

image

components/dashboard/src/custom-hooks/usePagination.ts
```diff
@@ -0,0 +1,56 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { useMemo } from "react";
+
+export const usePagination = (props: {
```
andrew-farries, line 9: I'm not sure why this needs to be a hook. AIUI this returns an array of page numbers based on its arguments. I think the `calculatedPagination` function could be inlined into the `Pagination` component and called there directly.

laushinka: The main reason in extracting it was for readability and ease of maintenance. The usePagination logic is quite complex and it would be much harder to maintain with other noise around it.

andrew-farries: completely agree to make it a separate function rather than doing it inline in the component. I just wonder if it can be an ordinary function rather than a hook.

laushinka: Aha. I think technically yes, but I did want to make use of the `useMemo` for better performance, and therefore it has to be either a Hook or a React component. Do you think it would be overhead to maintain?

andrew-farries: I'm thinking of converting the hook to just a function called `getPageNumbers` (or whatever), that takes the same arguments as  the hook does now and returns an array of page numbers (as the hook does now).

If the function is created outside the `Pagination` component, it won't need to be memoized because its identity won't change between renders of the `Pagination` component.

laushinka: Hmm, makes sense. The new function will then have to be called inside a useEffect, right? Whenever each of the args change. Also let me know if a quick sync would be faster 😬

andrew-farries: I think there is no need for a `useEffect` - the function is just a helper for turning current `props` into an array of page numbers - no synchronization with anything external that would need a `useEffect`.

essentially the pattern is:

```ts
const func getPageNumbersFromProps(...) => { return [1,2,3,4] }

const MyComponent = (props) => {
  const xs = getPageNumbersFromProps(props)

  return (<ul>
    xs.map(x => <li>x</li>
  </ul>)
}
```

laushinka: Ah sorry not sure why I said useEffect I meant useState for updating the values, but they do get updated in the parent component. Will give this a shot after lunch.

components/dashboard/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+import { getPaginationNumbers } from "./getPagination";
+import Arrow from "../components/Arrow";
+
+interface PaginationProps {
+    totalResults: number;
+    totalNumberOfPages: number;
+    currentPage: number;
+    setCurrentPage: any;
```
andrew-farries, line 14: nit: I think you can give this a proper type; hovering over a setter should give you the type.

andrew-farries (approved):
Nice. I like this more with the unit test and without the hook 👍 

/hold for the typing issue and this comment: https://github.com/gitpod-io/gitpod/pull/11800/files#r936443747

components/dashboard/src/Pagination/Pagination.tsx
laushinka: Lol totally forgot about this one. Done!

laushinka: Btw it's also on hold because I still need to change the `resultsPerPage` that I changed for this PR.

laushinka: > yeah, I linked your comment in the review summary comment.

My brain is so not working right now 🤣

components/dashboard/src/teams/TeamUsage.tsx
laushinka: Already put back and will unhold.

laushinka:
Thank you @andrew-farries for reviewing and making sure I don't over-engineer 🧡

PR #11802, opened 2022-08-02
Detach dangling base mounts

Furisto:
## Description
As a last ditch effort detach base mounts if they cannot be unmounted. That way new accesses are denied and it will be unmounted once all processes are done using it.

## Related Issue(s)

Related to https://github.com/gitpod-io/gitpod/issues/11369

## Release Notes

```release-note
None
```
## Werft options:

- [ ] /werft with-preview


jenting (approved):
LGTM

/hold it'd be good to be double reviewed by other team members.

PR #11809, opened 2022-08-02
[ws-daemon] Enable workspace content housekeeping

### LINKED ISSUE #11369: `ws-daemon` sometimes didn't delete workspace file after backup

csweichel:
## Description
This PR re-instates ws-daemon's content garbage collection which removes workspace content from disk once the state file is gone. We had originally disabled this behaviour because it interfered with the ws-daemon/ws-sync migration three years ago.

There's a grace period for content built-in which ensures we only delete content older than two hours. The motivation is to guard against accidental state mishaps.

>**Warning**
> This PR carries the risk that it might delete workspace content unintentionally. 

## Related Issue(s)

Fixes #11369

## How to test
- see unit test
- start a workspace, produce a file in the workspace content directory which cannot be deleted and stop the workspace
- make the workspace content directory deletable again and see it being deleted

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/ws-daemon/pkg/internal/session/store.go
+++ b/components/ws-daemon/pkg/internal/session/store.go
@@ -6,6 +6,8 @@ package session
 
 import (
 	"context"
+	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 	"sync"
@@ -186,6 +188,9 @@ func (s *Store) StartHousekeeping(ctx context.Context, interval time.Duration) {
 	log.Debug("stopping workspace housekeeping")
 }
 
+// For good measure, we only GC ontent directories older than this age.
+const minContentGCAge = 2 * time.Hour
+
 func (s *Store) doHousekeeping(ctx context.Context) (errs []error) {
 	//nolint:ineffassign,staticcheck
 	span, ctx := opentracing.StartSpanFromContext(ctx, "doHousekeeping")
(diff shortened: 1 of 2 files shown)


components/ws-daemon/pkg/internal/session/store.go
```diff
@@ -242,36 +244,49 @@ func (s *Store) doHousekeeping(ctx context.Context) (errs []error) {
 	// 	log.WithField("workspace", ws).Info("deleted workspaces without content directory")
```
Furisto, line 244: Can we remove the commented out code? If need be we can get it from history.

components/ws-daemon/pkg/internal/session/store.go
```diff
@@ -242,36 +244,49 @@ func (s *Store) doHousekeeping(ctx context.Context) (errs []error) {
...
+			continue
+		}
 
-	// 	log.WithField("directory", f.Name()).Info("deleted workspace content directory without corresponding state file")
-	// }
+		// We have found a workspace content directory without a workspace state file, which means we don't manage this folder.
+		// Within the working area/location of a session store we must be the only one who creates directories, because we want to
+		// make sure we don't leak files over time.
+
+		// For good measure we wait a while before deleting that direcotry.
```
sagor999, line 270:
```suggestion
		// For good measure we wait a while before deleting that directory.
```

components/ws-daemon/pkg/internal/session/store_test.go
```diff
@@ -0,0 +1,118 @@
+// Copyright (c) 2020 Gitpod GmbH. All rights reserved.
```
sagor999, line 1: copy paste copyright? (should be 2022)

sagor999 (approved):
LGTM
/hold
for nits

jenting (commented):
/lgtm

components/ws-daemon/pkg/internal/session/store.go
```diff
@@ -242,36 +244,49 @@ func (s *Store) doHousekeeping(ctx context.Context) (errs []error) {
...
+		nfo, err := f.Info()
+		if err != nil {
+			log.WithError(err).Warn("Found workspace content directory without a corresponding state file, but could not retrieve its info")
+			errs = append(errs, err)
+			continue
+		}
+		if time.Since(nfo.ModTime()) < 2*time.Hour {
```
utam0k, line 277: Why not give the 2h a name and make it const?

Issue #11402, opened 2022-07-15
Define & Implement Entitlement Service APIs


geropl:
This consists of several tasks:
 - [x] Setup EntitlementService in `server` / TypeScript
 - [x] Implement it for licenses
 - [ ] Implement it for Stripe
 - [x] Wire it up with existing code (e.g., Chargbee)

PR #11813, opened 2022-08-02
BillingMode: add Json RPC API methods (2/4)

geropl:
## Description
Introduces JSON RPC API to retrieve the current BillingModes for a given user/team.

## Related Issue(s)

Context: https://github.com/gitpod-io/gitpod/issues/11402

## How to test
 - see [previous PR](https://github.com/gitpod-io/gitpod/pull/11812)

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/server/ee/src/workspace/gitpod-server-impl.ts
+++ b/components/server/ee/src/workspace/gitpod-server-impl.ts
@@ -6,7 +6,7 @@
 
 import { injectable, inject } from "inversify";
 import { GitpodServerImpl, traceAPIParams, traceWI, censor } from "../../../src/workspace/gitpod-server-impl";
-import { TraceContext } from "@gitpod/gitpod-protocol/lib/util/tracing";
+import { TraceContext, TraceContextWithSpan } from "@gitpod/gitpod-protocol/lib/util/tracing";
 import {
     GitpodServer,
     GitpodClient,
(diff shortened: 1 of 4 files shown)


components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2190,6 +2194,22 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
         };
     }
 
+    async getBillingModeForUser(ctx: TraceContextWithSpan): Promise<BillingMode> {
+        traceAPIParams(ctx, {});
+
+        const user = this.checkUser("getBillingModeForUser");
+        return this.billingModes.getBillingModeForUser(user, new Date());
```
andrew-farries, line 2201: Passing `new Date()` here feels a little odd, in that I wouldn't expect a user or team's billing mode to depend on the current time.

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2190,6 +2194,22 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
+    }
+
+    async getBillingModeForTeam(ctx: TraceContextWithSpan, teamId: string): Promise<BillingMode> {
+        traceAPIParams(ctx, { teamId });
+
+        this.checkAndBlockUser("getBillingModeForTeam");
+        const team = await this.guardTeamOperation(teamId, "get");
+
+        return this.billingModes.getBillingModeForTeam(team, new Date());
```
andrew-farries, line 2210: Same comment about the `new Date()`

components/server/ee/src/workspace/gitpod-server-impl.ts
geropl: It does, and other methods take it as well.

Thing is, I tried to remove it here by making it optional (defaulting to `new Date()`), but that does **not** result in compile errors when you implement it wrongly - thank you, Javascript. :face_with_spiral_eyes:

geropl: I'm all-in for streamlinig the API, but considered it out-of-scope because of the extra-effort to avoid runtime issues.

andrew-farries: yeah, making it default to the current time was what I was getting at, but if that's more trouble than it's worth ✅

andrew-farries (approved):

PR #11814, opened 2022-08-02
BillingMode: Use BillingMode to restrict API access (3/4)

geropl:
## Description
This PR introduces the usage of the `getBillindMode` calculation across the code base.
It turned out quite big in terms of lines of code, but can be either reviewed commit-by-commit, or even split into multiple PRs if that increases confidence. It contains four main changes:
 - Guard calculation of BillingMode by new FF "isBillingModeEnabled": if false, getBillingMode always returns "chargebee" (in SaaS). This is meant to de-risk the rollout of "BillingModes", prior to even thinking about enabling "usage-based pricing" Feature Flag. This is meant to go away short-term (end of next week?) once we see that calling the actual impl does not break prod
 - some improvements and fixes to getBillingMode
 - Enforce that Chargebee/Stripe API is only usable with the respective BillingMode: There is one TODO: We currently guard access to the "Portal" as well, because it allows people to perform checkouts and upgrades as well, which we want to prohibit. This might block people from accessing their invoices, though. If this is a problem we should investigate in a follow-up issue.
 - Use BillingMode to guide what components are shown in the UI: here I might be wrong in some details, but the commit should convey the idea of how to use the BIllingMode for these decisions
 
### ToDos for follow-up PRs:
   - [ ] Align access to configuring WorkspaceClasses with `BillingMode.canSetWorkspaceClass`
   ~- [ ] Find a way to display "SpendingLimit" independent of "Billing" component ([here](https://github.com/gitpod-io/gitpod/blob/1e909bab0e075c2816118a001a435326bb1888f5/components/dashboard/src/settings/Billing.tsx#L14))~
   - [ ] Cleanup: Introduce a `TeamContext` and use it across dashboard
   - [ ] Allow access to old Chargebee invoices _without_ allowing them ti upgrade (Portal access)

## Related Issue(s)

Fixes #

## How to test
- `(cd components/server && leeway build .:dbtest)` : verify that all BillingMode tests are still running

### isBillingModeEnabled: false
- go to preview environment, sign up with test user 1 (U1) and see that Chargebee integration works as expected:
   1. start workspace and try to increase timeout: :x: 
   2. Upgrade to "Unleashed", start workspace and try to increase timeout: :heavy_check_mark:  
- however, you do _not_ see any usage-based UI

### isBillingModeEnabled: true
- repeat the test from above, with a new user (U2)
- notice how everything still works (now with more expensive computation, maybe)

:information_source:  Testing `isUsageBaseBilling` is actually out-of-scope for this PR

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/dashboard/src/Menu.tsx
+++ b/components/dashboard/src/Menu.tsx
@@ -155,10 +155,13 @@ export default function Menu() {
 
         // Refresh billing mode
         refreshUserBillingMode();
-        if (team) {
-            server.getBillingModeForTeam(team.id).then(setTeamBillingMode);
-        }
     }, []);
+
+    useEffect(() => {
+        if (team) {
+            getGitpodService().server.getBillingModeForTeam(team.id).then(setTeamBillingMode);
+        }
+    }, [team]);
 
     const teamOrUserSlug = !!team ? "/t/" + team.slug : "/projects";
     const leftMenu: Entry[] = (() => {
--- a/components/dashboard/src/Menu.tsx
+++ b/components/dashboard/src/Menu.tsx
@@ -26,7 +26,7 @@ import { ProjectContext } from "./projects/project-context";
 import { PaymentContext } from "./payment-context";
 import FeedbackFormModal from "./feedback-form/FeedbackModal";
 import { inResource, isGitpodIo } from "./utils";
-import { FeatureFlagContext } from "./contexts/FeatureFlagContext";
+import { BillingMode } from "@gitpod/gitpod-protocol/lib/billing-mode";
 
 interface Entry {
     title: string;
@@ -35,13 +35,12 @@ interface Entry {
 }
 
 export default function Menu() {
-    const { user } = useContext(UserContext);
+    const { user, userBillingMode, refreshUserBillingMode } = useContext(UserContext);
     const { teams } = useContext(TeamsContext);
     const location = useLocation();
     const team = getCurrentTeam(location, teams);
-    const { showPaymentUI, setShowPaymentUI, setCurrency, setIsStudent, setIsChargebeeCustomer } =
-        useContext(PaymentContext);
-    const { showUsageBasedPricingUI } = useContext(FeatureFlagContext);
+    const { setCurrency, setIsStudent, setIsChargebeeCustomer } = useContext(PaymentContext);
+    const [teamBillingMode, setTeamBillingMode] = useState<BillingMode | undefined>(undefined);
     const { project, setProject } = useContext(ProjectContext);
     const [isFeedbackFormVisible, setFeedbackFormVisible] = useState<boolean>(false);
 
@@ -146,7 +145,6 @@ export default function Menu() {
     useEffect(() => {
         const { server } = getGitpodService();
         Promise.all([
-            server.getShowPaymentUI().then((v) => () => setShowPaymentUI(v)),
             server.getClientRegion().then((v) => () => {
                 // @ts-ignore
                 setCurrency(countries[v]?.currency === "EUR" ? "EUR" : "USD");
@@ -154,6 +152,12 @@ export default function Menu() {
             server.isStudent().then((v) => () => setIsStudent(v)),
             server.isChargebeeCustomer().then((v) => () => setIsChargebeeCustomer(v)),
         ]).then((setters) => setters.forEach((s) => s()));
+
+        // Refresh billing mode
+        refreshUserBillingMode();
+        if (team) {
+            server.getBillingModeForTeam(team.id).then(setTeamBillingMode);
+        }
     }, []);
 
     const teamOrUserSlug = !!team ? "/t/" + team.slug : "/projects";
@@ -191,7 +195,7 @@ export default function Menu() {
                     link: `/t/${team.slug}/members`,
                 },
             ];
-            if (showUsageBasedPricingUI) {
+            if (teamBillingMode && teamBillingMode.mode === "usage-based") {
                 teamSettingsList.push({
                     title: "Usage",
                     link: `/t/${team.slug}/usage`,
@@ -201,7 +205,7 @@ export default function Menu() {
                 teamSettingsList.push({
                     title: "Settings",
                     link: `/t/${team.slug}/settings`,
-                    alternatives: getTeamSettingsMenu({ team, showPaymentUI }).flatMap((e) => e.link),
+                    alternatives: getTeamSettingsMenu({ team, billingMode: teamBillingMode }).flatMap((e) => e.link),
                 });
             }
 
@@ -216,7 +220,7 @@ export default function Menu() {
             {
                 title: "Settings",
                 link: "/settings",
-                alternatives: getSettingsMenu({ showPaymentUI, showUsageBasedPricingUI }).flatMap((e) => e.link),
+                alternatives: getSettingsMenu({ userBillingMode }).flatMap((e) => e.link),
             },
         ];
     })();
--- a/components/dashboard/src/settings/settings-menu.ts
+++ b/components/dashboard/src/settings/settings-menu.ts
@@ -4,6 +4,7 @@
  * See License-AGPL.txt in the project root for license information.
  */
 
+import { BillingMode } from "@gitpod/gitpod-protocol/lib/billing-mode";
 import {
     settingsPathAccount,
     settingsPathBilling,
@@ -17,7 +18,7 @@ import {
     settingsPathSSHKeys,
 } from "./settings.routes";
 
-export default function getSettingsMenu(params: { showPaymentUI?: boolean; showUsageBasedPricingUI?: boolean }) {
+export default function getSettingsMenu(params: { userBillingMode?: BillingMode }) {
     return [
         {
             title: "Account",
@@ -27,26 +28,7 @@ export default function getSettingsMenu(params: { showPaymentUI?: boolean; showU
             title: "Notifications",
             link: [settingsPathNotifications],
         },
-        ...(params.showPaymentUI
-            ? [
-                  ...(params.showUsageBasedPricingUI
-                      ? [
-                            {
-                                title: "Billing",
-                                link: [settingsPathBilling],
-                            },
-                        ]
-                      : []),
-                  {
-                      title: "Plans",
-                      link: [settingsPathPlans],
-                  },
-                  {
-                      title: "Team Plans",
-                      link: [settingsPathTeams],
-                  },
-              ]
-            : []),
+        ...renderBillingMenuEntries(params.userBillingMode),
         {
             title: "Variables",
             link: [settingsPathVariables],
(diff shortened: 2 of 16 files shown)


components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -1122,6 +1126,7 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
         return this.eligibilityService.isStudent(user);
     }
 
+    // **deprecated**
```
jankeromnes, line 1129: Is this really deprecated?

To me, the BillingModes are only there to help with the transition to Stripe, but I thought that, as soon as all customers are migrated to Stripe, we'd delete the BillingModes and fall back to `getShowPaymentUI` again?

A very simple, boolean switch to turn off any payment UI seems like a must have, while replacing it with a somewhat complicated state machine with possible new bugs sounds riskier.

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -1199,6 +1204,7 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
     async createPortalSession(ctx: TraceContext): Promise<{}> {
         const user = this.checkUser("createPortalSession");
         const logContext = { userId: user.id };
+        await this.ensureChargebeeApiIsAllowed({ user });
```
jankeromnes, line 1207: Playing devil's advocate for a bit: What is the value of purposefully making API methods fail when users are not in the "right mode"?

To me, it's more important the make the UI look exactly correct (e.g. show the relevant UI pages / controls / menu items for whichever mode you're in), but it's less important that every API call is completely fenced-off all the time.

For example:
- What if someone is already using Usage-Based, but they still have some Chargebee subscription? (I know we don't expect this to happen, but if it does, is there value in forbidding the "cancel" API call? Or even forbidding users from opening the Chargebee portal?)

jankeromnes: To clarify, I'm still happy to go ahead with this change, but I wonder if maybe it would make more sense to always allow certain API calls regardless of your billing mode (e.g. showing the Chargebee portal, and maybe cancelling, while the checkout methods are good to fence off to prevent buying new Chargebee plans once you're on the other side).

jankeromnes (commented):
Many thanks! On principle, this change looks good. However:

- I would be very keen to test it (manually). Could you please deploy it with payment, and with a Chargebee webhook? This would help go through the various billing flows and build confidence that this actually does what we want.

- I'd love to see a UI counterpart to these changes. Right now, it looks like we're just making various UI bits silently fail (e.g. for Gitpod team, we have UBB enabled, but still a Chargebee plan -- so I guess we'll see the UBB UI, but if we try to load it or click on any buttons, I expect this PR will make everything fail 😬)

components/server/ee/src/workspace/gitpod-server-impl.ts
geropl: Yes, that's debateable. 

> while replacing it with a somewhat complicated state machine with possible new bugs sounds riskier.

That's not what happens: When we got rid of Chargebee, it's a enum with two cases: "none" (which maybe should be named "license") and "usage-based". I think this is superior, because then we have a clear indicator as to what to show on the UI, for instance to deciding between showing self-hosted or SaaS concerns in the Admin dashboard.

But let's leave it as out-of scope for this PR.

components/server/ee/src/workspace/gitpod-server-impl.ts
geropl: > but it's less important that every API call is completely fenced-off all the time.

Never trust the client, basically. And as we want to be strict with enforcing that users cannot (accidentally) move back to Chargebee, this just make sense to me.

> is there value in forbidding the "cancel" API call? Or even forbidding users from opening the Chargebee portal?

Ah, those I planned to leave out in this PR. If I haven't, I will do another round to remove those. (Portal is debateable, because people can upgrade there, too!)

geropl:
> I would be very keen to test it (manually). Could you please deploy it with payment, and with a Chargebee webhook? This would help go through the various billing flows and build confidence that this actually does what we want.

Absolutely. This is one of the first things we should do next week. Will try to move this PR in shape until then :upside_down_face: 

> I'd love to see a UI counterpart to these changes

:100: Yeah, wondered if we should to that change in the same PR - to be able to better test it/avoid double testing.
I won't find time for this I think, maybe someone can take over on Monday? :thinking:

geropl:
@jankeromnes I re-did the whole PR, and think adressed all your points + completed it (incl. UI work).
There might be small issues for Usage-Based (I did not test everything, and check out the ToDos :point_up: ). But the point for this PR is to _not break Chargebee_. The additional feature flag is there to allow for safe rollout to prod, so we can continue polishing this _before_ even enabling UBP:

components/server/ee/src/workspace/gitpod-server-impl.ts
geropl: cmp [comment](https://github.com/gitpod-io/gitpod/pull/11814#issuecomment-1207714530).

components/server/ee/src/workspace/gitpod-server-impl.ts
geropl: cmp. [comment](https://github.com/gitpod-io/gitpod/pull/11814#issuecomment-1207714530).

jankeromnes:
> I re-did the whole PR, and think adressed all your points + completed it (incl. UI work).

Whoa 🤯 that's amazing. Thanks @geropl! I'll review & test this this morning. 🔍

> There is one TODO: We currently guard access to the "Portal" as well, because it allows people to perform checkouts and upgrades as well, which we want to prohibit. This might block people from accessing their invoices, though. If this is a problem we should investigate in a follow-up issue.

I've never seen or heard from checkouts or upgrades performed form the Chargebee portal 🤔 to me, it only serves to:
- get invoices
- update payment method
- view subscription details (no upgrade/downgrade, no purchasing new plans)
- cancel existing subscriptions

If I'm right, it would be 100% safe to always enable the Portal.

geropl:
> If I'm right, it would be 100% safe to always enable the Portal.

:tada: That would be awesome! I basically did not check, and really hope this to be a no-brainer. :lotus_position:

jankeromnes:
Hey Werft, hope you had a great weekend! 🌻 For when you've got a minute:

/werft run

jankeromnes:
> > If I'm right, it would be 100% safe to always enable the Portal.
> 
> :tada: That would be awesome! I basically did not check, and really hope this to be a no-brainer. :lotus_position: 

@geropl I had a look through the portal settings. Seems like I'm mostly right, but the only question is what "Allow customers to change (active) subscriptions" means. 🤔 Will try to find & use this feature while testing your PR.

<table><tr><td>
Screenshot 2022-08-08 at 09 52 42
</td><td>
Screenshot 2022-08-08 at 09 53 43
</td></tr></table>

jankeromnes:
Oh. First test of the PR, I sign in as a new user, but I get no billing items/UI at all in `/settings` 😬

Screenshot 2022-08-08 at 10 47 24

jankeromnes:
Also, I'm unable to enable Usage-Based Billing for myself -- when I create a team called "Gitpod", nothing changes in the UI (still Chargebee in Team Settings and still no payment UI in user). I guess something is broken somewhere. Heading back to the code.

geropl:
@jankeromnes Judging from your comment above, I think we're still mis-aligned on what we're trying to achieve here.
Can we sync to avoid further confusion?

components/dashboard/src/user-context.tsx
```diff
@@ -5,18 +5,33 @@
...
 
 import { User } from "@gitpod/gitpod-protocol";
+import { BillingMode } from "@gitpod/gitpod-protocol/lib/billing-mode";
 import React, { createContext, useState } from "react";
+import { getGitpodService } from "./service/service";
 
 const UserContext = createContext<{
     user?: User;
     setUser: React.Dispatch<User>;
+    userBillingMode?: BillingMode;
```
jankeromnes, line 15: I think this should be a `React.Dispatch`, in order for changes to make their way to all the context users (e.g. any page with settings menu):

```suggestion
    userBillingMode?: React.Dispatch<BillingMode>;
```

jankeromnes:
Hmm, I've just added a commit which might fix the bug, but I realized seconds too late that I could have used dashboard telepresence instead 🙄 (and then added debug logs to test other things).

jankeromnes:
Okay, that fixed it, and I also set up a Chargebee webhook. So far it looks like all the Chargebee flows are working. 👀

components/dashboard/src/settings/settings-menu.ts
```diff
@@ -65,3 +47,40 @@ export default function getSettingsMenu(params: { showPaymentUI?: boolean; showU
...
+                          {
+                              title: "Team Plans",
+                              link: [settingsPathTeams],
+                          },
+                      ]
+                    : []),
+                {
+                    title: "Billing",
+                    link: [settingsPathBilling],
+                },
```
jankeromnes, line 83: Why was this moved below `Team Plans`? To me it's more important, and should remain above it. 💭

components/dashboard/src/Menu.tsx
```diff
@@ -146,14 +145,19 @@ export default function Menu() {
...
             }),
             server.isStudent().then((v) => () => setIsStudent(v)),
             server.isChargebeeCustomer().then((v) => () => setIsChargebeeCustomer(v)),
         ]).then((setters) => setters.forEach((s) => s()));
+
+        // Refresh billing mode
+        refreshUserBillingMode();
+        if (team) {
+            server.getBillingModeForTeam(team.id).then(setTeamBillingMode);
+        }
```
jankeromnes, line 160: ⚠️ This should probably be run again every time the `team` changes (e.g. if you're viewing one team which has this off, then move to a different team -- currently the effect you wrote this in runs only once when the `Menu` component gets mounted).

A separate, team-dependent effect could look like this:

```ts
useEffect(() => {
    if (team) {
        server.get // ...
    }
}, [team]);
```

jankeromnes: Addressed in https://github.com/gitpod-io/gitpod/pull/11814/commits/0006b8d855ca60773cf3fd1b2b9601e5f407fc32

components/dashboard/src/settings/settings-menu.ts
jankeromnes: Addressed in https://github.com/gitpod-io/gitpod/pull/11814/commits/0006b8d855ca60773cf3fd1b2b9601e5f407fc32

jankeromnes:
Update on https://github.com/gitpod-io/gitpod/pull/11814#issuecomment-1207790306 -- I can confirm it's 100% safe to enable the Chargebee portal for everyone who wants it, as it doesn't allow purchasing new plans or upgrading/downgrading (only consulting invoices, managing customer/payment details, and cancelling existing plans).

The only uncertainty from https://github.com/gitpod-io/gitpod/pull/11814#issuecomment-1207790306 was "editing subscription", which is actually just the ability to add/remove coupons, which is 100% okay:

| Editing a subscription (you can only apply a coupon) |
| --- |
| Screenshot 2022-08-08 at 17 35 02 |

I'll de-gate these methods as a last touch-up, then re-read once and approve, but hold in case Alex wants to review as well.

jankeromnes (approved):
Thanks again @geropl!

Approving to unblock this PR, after having:
- Read the code (somewhat quickly) a few times
- Added a few minor fixes
- Tested that all Chargebee flows still work without interference 👍

However, adding a hold because:
- This slightly breaks upgrading to Usage-Based (e.g. when users have cancelled all their plans and enabled the UBB flag, they still won't get the Billing menu entry -- see comment)
- @AlexTugarev mentioned he might like to have a look as well, which could be cool (more eyes = more safety 😋)

/hold

AlexTugarev:
Tested a couple of times and the CB side worked without any issues so far. Gimme just some more time to understand some details before releasing the breaks on this. 🙏🏻

components/dashboard/src/settings/settings-menu.ts
```diff
@@ -65,3 +47,40 @@ export default function getSettingsMenu(params: { showPaymentUI?: boolean; showU
...
+                {
+                    title: "Plans",
+                    link: [settingsPathPlans],
+                },
+                {
+                    title: "Team Plans",
+                    link: [settingsPathTeams],
+                },
+            ];
+        case "usage-based":
```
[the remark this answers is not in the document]
AlexTugarev: Looks like `BillingMode.showUsageBasedBilling` should be used, indeed.

Issue #11823, opened 2022-08-03
Sometimes workspace attempts to start with PVC feature enabled


sagor999:
### Bug description

Sometimes workspace attempts to start with PVC feature enabled, even though user does not have flag enabled, nor the project has flag enabled.
StartWorkspace request example:
```
{
    "id": "fdec6ee9-73a1-4104-9490-a9de771db03f",
    "metadata": {
        "metaId": "matiasmillan-bc9-mcx26mbrw4b",
        "owner": "b2761e91-264e-454c-b811-f8b54367e4f2",
        "project": "a681734e-f4c4-46b2-89e0-de801d31262d"
    },
    "servicePrefix": "matiasmillan-bc9-mcx26mbrw4b",
    "spec": {
        "class": "default",
        "deprecatedIdeImage": "eu.gcr.io/gitpod-core-dev/build/ide/code:commit-079f161f0aaf95d6b5e28f0709b3e629337291e9",
        "envvars": [
            {
                "name": "GITPOD_IDE_ALIAS",
                "value": "intellij"
            },
            {
                "name": "GITPOD_WORKSPACE_CONTEXT_URL",
                "value": "<redacted>"
            },
            {
                "name": "GITPOD_WORKSPACE_CONTEXT",
                "value": "{"isFile":false,"path":"","title":"<redacted>","ref":"main","refType":"branch","revision":"<redacted>","repository":{"cloneUrl":"<redacted>,"host":"github.com","name":"<redacted>","owner":"<redacted>","private":false,"fork":{"parent":{"cloneUrl":"<redacted>","host":"github.com","name":"<redacted>","owner":"<redacted>","private":false}}},"normalizedContextURL":"<redacted>","checkoutLocation":"<redacted>","upstreamRemoteURI":"<redacted>","prebuildWorkspaceId":"9c8960b7-0839-4a29-aaf8-ee0d72fafb1e","wasPrebuilt":true}"
            },
            {
                "name": "GITPOD_TASKS",
                "value": "[{"init":"./gradlew build"}]"
            },
            {
                "name": "VSX_REGISTRY_URL",
                "value": "https://open-vsx.gitpod.io/"
            },
            {
                "name": "SUPERVISOR_DOTFILE_REPO"
            },
            {
                "name": "[redacted]",
                "value": "[redacted]"
            }
        ],
        "featureFlags": [
            "PERSISTENT_VOLUME_CLAIM"
        ],
        "git": {
            "email": "<redacted>",
[...]

PR #11824, opened 2022-08-03
add log when starting workspace with pvc

sagor999:
## Related Issue(s)

Related https://github.com/gitpod-io/gitpod/issues/11823

## Release Notes

```release-note
none
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/server/src/workspace/workspace-starter.ts
+++ b/components/server/src/workspace/workspace-starter.ts
@@ -1366,6 +1366,16 @@ export class WorkspaceStarter {
         let volumeSnapshotInfo = new VolumeSnapshotInfo();
         const volumeSnapshots = await this.workspaceDb.trace(traceCtx).findVolumeSnapshotById(volumeSnapshotId);
         if (volumeSnapshots !== undefined) {
+            log.info(
+                {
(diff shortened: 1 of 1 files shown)


components/server/src/workspace/workspace-starter.ts
```diff
@@ -1366,6 +1366,16 @@ export class WorkspaceStarter {
...
+            log.info(
+                {
+                    lastInstanceId: lastValidWorkspaceInstanceId,
+                    volSnapshotId: volumeSnapshotId,
+                    snapshotId: volumeSnapshots.id,
+                    volumeHandle: volumeSnapshots.volumeHandle,
+                    workspaceId: workspace.id,
+                },
+                "starting workspace with volume snapshot info",
+            );
```
andrew-farries, line 1378: would a metric be more useful than a log message or is there already something in place?

andrew-farries (approved):
/hold for question about log vs metric

components/server/src/workspace/workspace-starter.ts
sagor999: No, we kind of need log instead for debugging reasons.

PR #11825, opened 2022-08-03
observability: Add a alert for the network connections.

utam0k:
## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/operations/observability/mixins/workspace/rules/components/nodes/alerts.libsonnet
+++ b/operations/observability/mixins/workspace/rules/components/nodes/alerts.libsonnet
@@ -48,6 +48,22 @@
               increase(cluster_autoscaler_failed_scale_ups_total[1m]) != 0
             |||,
           },
+          {
+            alert: 'NetworkConnectionsTooHigh',
+            labels: {
+              severity: 'warning',
+              team: 'workspace'
+            },
+            'for': '10m',
+            annotations: {
+              runbook_url: 'https://github.com/gitpod-io/runbooks/blob/main/runbooks/NetworkConnectionsTooHigh.md',
+              summary: "Network connection numbers remain high for 5 minutes.",
+              description: 'Network connection numbers remain high for 5 minutes.',
+            },
+            expr: |||
+              node_nf_conntrack_entries{instance=~"workspace-.*", instance!~"serv.*"} > 20000
+            |||,
+          },
         ],
       },
     ],


operations/observability/mixins/workspace/rules/components/nodes/alerts.libsonnet
```diff
@@ -48,6 +48,21 @@
...
+              severity: 'critical',
+            },
+            'for': '5m',
+            annotations: {
+              runbook_url: 'https://github.com/gitpod-io/runbooks/blob/main/runbooks/NetworkConnectionsTooHigh.md',
+              summary: "Network connection numbers remain high for 5 minutes.",
+              description: 'Network connection numbers remain high for 5 minutes.',
+            },
+            expr: |||
+              node_nf_conntrack_entries{instance=~"workspace-.*"} > 5000
```
utam0k, line 63: [grafana](https://grafana.gitpod.io/explore?orgId=1&left=%7B%22datasource%22:%22VictoriaMetrics%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22node_nf_conntrack_entries%7Binstance%3D~%5C%22workspace-.*%5C%22%7D%20%3E%205000%22%7D%5D,%22range%22:%7B%22from%22:%221659473696772%22,%22to%22:%221659493263830%22%7D%7D)

Furisto: I fear that people will get paged needlessly if the alert already fires when the number of connections goes above 5000. Is this not very similar to the GitpodNodeConntrackTableIsFull and GitpodNodeConntrackTableGettingFull alert?

kylos101: @utam0k can you change this back to draft? :pray: 

Why draft? I think more testing needs to be done with the expression, prior to marking ready for review. For example, given the expression you want to use here, can you test it to see if it helps you find abusers on nodes? If no, you cannot find related abuse with it, then the # of connections is probably too low.

> I fear that people will get paged needlessly if the alert already fires when the number of connections goes above 5000.

Agreed

> Is this not very similar to the GitpodNodeConntrackTableIsFull and GitpodNodeConntrackTableGettingFull alert?

This is a bit different, because those are percentage based like ` (node_nf_conntrack_entries / node_nf_conntrack_entries_limit) > 0.95`. What we're trying to build here is a water mark, where, if it gets exceeded for a period of time, that's a signal that potential abuse needs to be investigated. I agree with @Furisto , we don't want it to fire excessively.

@utam0k @Furisto 

wdyt about something like `node_nf_conntrack_entries{node=~"workspace.*", instance!~"serv.*"} > 20000` where the duration is `10m`? If you [look at this now](https://grafana.gitpod.io/explore?orgId=1&left=%7B%22datasource%22:%22VictoriaMetrics%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22node_nf_conntrack_entries%7Bnode%3D~%5C%22workspace.*%5C%22,%20instance!~%5C%22serv.*%5C%22%7D%20%3E%2020000%22%7D%5D,%22range%22:%7B%22from%22:%22now-1h%22,%22to%22:%22now%22%7D%7D), you'll see there are three nodes, which could potentially have abusers.

image

Furisto: > This is a bit different, because those are percentage based like (node_nf_conntrack_entries / node_nf_conntrack_entries_limit) > 0.95. 

Isn't that the same thing, just expressed differently? A percentage of something resolves to a number in the end.

utam0k: GitpodNodeConntrackTableIsFull is an alert that detects node limits, this time for abuse users. I believe that is the difference. So I think the main feature is that node_nf_contrack_entries_limit is not related to the threshold in this alert.

utam0k: I create [a metrics](https://grafana.gitpod.io/explore?orgId=1&left=%7B%22datasource%22:%22VictoriaMetrics%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22min_over_time(node_nf_conntrack_entries%7Binstance%3D~%5C%22workspace-.*%5C%22,%20%20instance!~%5C%22serv.*%5C%22%7D%5B10m%5D)%20%3E%2020000%22,%22editorMode%22:%22code%22,%22range%22:true,%22instant%22:true,%22exemplar%22:false%7D%5D,%22range%22:%7B%22from%22:%221659594721315%22,%22to%22:%221659633928038%22%7D%7D). 
How about having the on-caller check this metric for abuse?


> wdyt about something like node_nf_conntrack_entries{node=~"workspace.*", instance!~"serv.*"} > 20000 where the duration is 10m? If you [look at this now](https://grafana.gitpod.io/explore?orgId=1&left=%7B%22datasource%22:%22VictoriaMetrics%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22node_nf_conntrack_entries%7Bnode%3D~%5C%22workspace.*%5C%22,%20instance!~%5C%22serv.*%5C%22%7D%20%3E%2020000%22%7D%5D,%22range%22:%7B%22from%22:%22now-1h%22,%22to%22:%22now%22%7D%7D), you'll see there are three nodes, which could potentially have abusers.

operations/observability/mixins/workspace/rules/components/nodes/alerts.libsonnet
```diff
@@ -48,6 +48,21 @@
               increase(cluster_autoscaler_failed_scale_ups_total[1m]) != 0
             |||,
           },
+          {
+            alert: 'NetworkConnectionsTooHigh',
+            labels: {
+              severity: 'critical',
+            },
```
kylos101, line 55:
```suggestion
labels: {
              severity: 'warning',
              team: 'workspace'
            },
```

This way it routes to https://gitpod.slack.com/archives/C03QZ1NH517, and we can gauge if this is too frequent.

operations/observability/mixins/workspace/rules/components/nodes/alerts.libsonnet
kylos101: @utam0k I'm not sure we should route this to on-callers, yet.

@aledbf tracks connections and finds abusers very fast. @aledbf , wdyt of `min_over_time(node_nf_conntrack_entries{instance=~"workspace-.*",  instance!~"serv.*"}[10m]) > 20000` versus the expression that you use now?

operations/observability/mixins/workspace/rules/components/nodes/alerts.libsonnet
kylos101: :point_up: because I don't want to send the alert to on-callers until we (on the workspace team) have felt it first in Slack.

kylos101 (changes requested):
Let's route to Slack, instead of PagerDuty, for this initial alert.

Also, consulting with @aledbf Re: the prometheus expression being used to trigger the alert. He catches abuse often when inspecting connections, so will be "closer" to what is "reasonable".

utam0k:
@kylos101 @Furisto Thanks for your review. I updated this PR.

operations/observability/mixins/workspace/rules/components/nodes/alerts.libsonnet
```diff
@@ -48,6 +48,22 @@
...
           },
+          {
+            alert: 'NetworkConnectionsTooHigh',
+            labels: {
+              severity: 'warning',
+              team: 'workspace'
+            },
+            'for': '10m',
+            annotations: {
+              runbook_url: 'https://github.com/gitpod-io/runbooks/blob/main/runbooks/NetworkConnectionsTooHigh.md',
```
kylos101, line 59: @utam0k can you remove the runbook or leave it empty for now? I ask because eventually I think we'll want to reuse [this runbook](https://github.com/gitpod-io/runbooks/pull/371) for high normalized load and high # of connections.

Aside from that, this looks great!

utam0k: Sorry, I couldn't get 🙇 Should I close this PR and make `runbook_url` empty?
https://github.com/gitpod-io/runbooks/pull/379

kylos101: @utam0k this PR is good, no need to close. I was just sharing the runbook doesn't exist, and for the future we can model it after the one I shared.

You are welcome to mark this ready for review, so we can approve. :)

kylos101 (approved):
Go go gadget alert

Issue #11710, opened 2022-07-28
[ws-daemon] Unexpected cannot find workspace during DisposeWorkspace


aledbf:
### Bug description

![Jaeger-UI](https://user-images.githubusercontent.com/161571/181394879-28cafc4d-ef42-4dbb-b7d6-00b683d2c0f1.png)

### Steps to reproduce

-

### Expected behavior

If it's not really an error condition, then change this to be a warning. The warning would essentially be an indicator that we're calling dispose 2 or more times, and the 1st time would have done disposal.

### Anything else?

When does this error happen? Is it really an error? Is there anything else we can log with it, to determine the state of the workspace on the node (`/workspace` still exist?)

kylos101:
👋 @jenting , may I ask you to focus on https://github.com/gitpod-io/gitpod/issues/11713, move this back to scheduled, and unassign yourself for now? This way you can focus on #11713. If you think there's a need to do both at the same time, can you share why? I'm not seeing the relationship, and am trying to limit work in progress.

kylos101:
👋 @jenting alternatively, if you already started #11710, please move #11713 back to scheduled, etc.

jenting:
From the Jaeger tracing, the observation is that we call the `finalizeWorkspaceContent` almost within the same time.

<img width="672" alt="image" src="https://user-images.githubusercontent.com/49380831/182526044-a5a4625c-79e6-4110-a10f-d9ac302d14a7.png">
<img width="673" alt="image" src="https://user-images.githubusercontent.com/49380831/182526239-bcb2de7c-dbc4-44a6-82dc-20e300d2c4fb.png">

kylos101:
I am still seeing in `us60`:

![Image](https://user-images.githubusercontent.com/1272076/183795529-3082b3c8-a004-4471-94f9-431151f9ca47.png)

utam0k:
In this case, it has already been erased from ws-daemon's store. In other words, it was terminated normally, but the disposal was canceled due to a timeout in the context. In other words, it was terminated normally but failed due to a timing issue.
![image](https://user-images.githubusercontent.com/13010913/183860503-b564f6f7-ac7f-44eb-91d3-db680f72a20c.png)



> I am still seeing in `us60`, search workspace logs for instanceId `eb3d43eb-18bc-47fe-83b0-0adf388dcfb8` for detail from August 9.
> 
[...]

PR #11826, opened 2022-08-03
Add annotation gitpod.io/startedDisposal ealier

jenting:
## Description

Add annotation gitpod.io/startedDisposal once we enter finalize workspace content.

## Related Issue(s)

Related #11710

## How to test

None

## Release Notes

```release-note
None
```

## Documentation

None

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/ws-manager/pkg/manager/monitor.go
+++ b/components/ws-manager/pkg/manager/monitor.go
@@ -984,6 +984,12 @@ func (m *Monitor) finalizeWorkspaceContent(ctx context.Context, wso *workspaceOb
 		return
 	}
 
+	err := m.manager.markWorkspace(ctx, workspaceID, addMark(startedDisposalAnnotation, util.BooleanTrueString))
+	if err != nil {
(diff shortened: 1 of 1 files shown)


utam0k (approved):

sagor999:
@jenting I kind of disagree with this change. 
We will end up marking workspace several times, no? 
Since `finalizeWorkspaceContent` will be called several times. 
Also, if we were not able to connect to ws-daemon, then we did not actually started finalize process. But this change would have marked workspace as already started finalize.

jenting:
> @jenting I kind of disagree with this change. We will end up marking workspace several times, no? Since `finalizeWorkspaceContent` will be called several times. Also, if we were not able to connect to ws-daemon, then we did not actually started finalize process. But this change would have marked workspace as already started finalize.

Thanks for the feedback @sagor999 
After rethinking, I agree with you. I create a PR to revert it.

utam0k:
I'm not sure, but we may remove this annotation if it fails.

PR #11827, opened 2022-08-03
[dashboard] Fix displaying of logs for finished Prebuilds

### LINKED ISSUE #11755: A Prebuild Page Shows Nothing Where I Believe Logs Should Show Up
### Bug description

When visiting a prebuilds page there is a large amount of whitespace where I believe logs should show up. This is when you view the prebuild after it has completed.

Screen Shot 2022-07-29 at 12 43 52

When a prebuild is in progress you actually do see logs:

hnflchahhbaamldp

### Steps to reproduce

1. Create a project within Gitpod.
2. Perform an action which would trigger a prebuild - usually just making a commit should do the trick.
3. You can check the prebuild while it's going or after. If you check while it's going you should see some logs. If you check after you should see nothing.

Of note I have only replicated this on gitpod.io, I have not tried replicating it on a self-hosted instance but I have a feeling it would surface there as well.

### Workspace affected

_No response_

### Expected behavior

I would expect that even after the prebuild has been completed that I see the logs from the prebuild OR that the giant block of whitespace where one would think the logs would show up be hidden if our intention is not to show logs for completed prebuilds.

### Example repository

N/A

### Anything else?

N/A

Front logo [Front conversations](https://app.frontapp.com/open/top_3p34a)

geropl:
## Description
Fix a bug where we do not load Prebuild logs when the prebuild has already finished.

## Related Issue(s)

Fixes #11755 

## How to test
 - join [this team](https://gpl-11755-logs.preview.gitpod-dev.com/teams/join?inviteId=439d46d2-f218-4752-a96c-1f566a2fe637)
 - modify [this line](https://github.com/geropl/gitpod-test-repo/blob/gpl/docker-build-2mins/.gitpod.yml#L4)
 - [check](https://gpl-11755-logs.preview.gitpod-dev.com/t/superhumans/gitpod-test-repo) that the prebuild is executed
 - once it's "READY", go to the PrebuildDetails page (click the status) and notice how the logs are shown as well

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/dashboard/src/components/PrebuildLogs.tsx
+++ b/components/dashboard/src/components/PrebuildLogs.tsx
@@ -121,14 +121,16 @@ export default function PrebuildLogs(props: PrebuildLogsProps) {
                 disposables.push(retryWatchWorkspaceImageBuildLogs(workspaceId));
                 break;
             // When we're "running" we want to switch to the logs from the actual prebuild workspace, instead
+            // When the prebuild has "stopped", we still want to go for the logs
             case "running":
+            case "stopped":
                 disposables.push(
                     watchHeadlessLogs(
                         workspaceInstance.id,
                         (chunk) => {
                             logsEmitter.emit("logs", chunk);
                         },
-                        async () => workspaceInstance?.status.phase === "stopped",
+                        async () => false,
                     ),
                 );
         }


geropl:
/werft run

with flag

geropl:
@laushinka Could you take a look and review this one? I think you have most context here. :pray:

laushinka:
Works as advertised! Thanks for the fix 🙌🏽

laushinka (approved):

PR #11830, opened 2022-08-03
Log warning instead of tracing error span

### LINKED ISSUE #11710: [ws-daemon] Unexpected cannot find workspace during DisposeWorkspace

### LINKED ISSUE #11713: [ws-daemon] cannot find workspace during WaitForInit
### Bug description

### Steps to reproduce

-

### Workspace affected

_No response_

### Expected behavior

### Example repository

_No response_

### Anything else?

jenting:
## Description

Log warning instead of tracing error span if the error message contains `cannot find workspace`.

## Related Issue(s)

Fixes #11710
Fixes #11713

## How to test

None

## Release Notes

```release-note
None
```

## Documentation

None

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/ws-daemon/pkg/content/service.go
+++ b/components/ws-daemon/pkg/content/service.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jenting:
I'm thinking about whether we should handle this kind of error within ws-manager, rather than ws-daemon.

We made a similar approach for `WaitForInit` function, special handling the gRPC not found error. 
https://github.com/gitpod-io/gitpod/blob/e40e43d76120e5de702522e6b816f28b86a219c6/components/ws-manager/pkg/manager/monitor.go#L684-L698

WDYT? @csweichel @aledbf @kylos101

csweichel:
> I'm thinking about whether we should handle this kind of error within ws-manager, rather than ws-daemon.
> 
> We made a similar approach for `WaitForInit` function, special handling the gRPC not found error.
> 
> https://github.com/gitpod-io/gitpod/blob/e40e43d76120e5de702522e6b816f28b86a219c6/components/ws-manager/pkg/manager/monitor.go#L684-L698
> 
> WDYT? @csweichel @aledbf @kylos101

Agreed - handling this in ws-manager is the better way. For one, from ws-daemon's perspective the lifecycle of the workspace has actually ended, and keeping info about that workspace beyond this point doesn't feel "natural". From ws-manaer's perspective it's easier to keep the Git status in memory because the workspace still exists.

Furisto (approved):

PR #11831, opened 2022-08-03
[server] Introduce EntitlementServiceChargbee and move relevant parts of EligibilityService into it

geropl:
## Description
This PR does two things, while **not touching current behavior**:
 - it introduces the `EntitlementService` interface, which is based on a subset of methods of `EligibilityService`
 - it add a Chargebee implementation, which is 100% the identical code copied over from `EligibilityService`

## Related Issue(s)
Context: https://github.com/gitpod-io/gitpod/issues/11402

## How to test
 - open a workspace on this PR
 - get the content of the `EligiblityService` [before this commit](https://raw.githubusercontent.com/gitpod-io/gitpod/e3452c71b2fe1e81bfb91b5cc25bc8089d4efbde/components/server/ee/src/user/eligibility-service.ts), and put it into a file called `eligibilityservice.ts`
 - run `diff ee/src/billing/entitlement-service-chargebee.ts eligibilityservice.ts | less`
 - inspect the output carefully; notice how there is no relevant diff between the two

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


components/server/ee/src/billing/entitlement-service-chargebee.ts
```diff
@@ -0,0 +1,222 @@
...
+
+        return subscriptions.filter((s) => eligblePlans.includes(s.planId!)).length > 0;
+    }
+
+    /**
+     * Returns the default workspace timeout for the given user at a given point in time
+     * @param user
+     * @param date The date for which we want to know the default workspace timeout (depends on active subscription)
+     */
+    async getDefaultWorkspaceTimeout(user: User, date: Date = new Date()): Promise<WorkspaceTimeoutDuration> {
```
geropl, line 147: This method received the only notable modification: It is a merge of `UserService.getDefaultWorkspaceTimeout` and `EligibilityService.getDefaultWorkspaceTimeout`

geropl:
Noticed two small changes that I need to handle better. Will notify again once done. :+1:

components/server/ee/src/billing/entitlement-service-chargebee.ts
geropl: Ok, reverted that part. It's a plain copy now.

geropl:
@AlexTugarev This is ready for review now! It should really not change any behavior the slightest, and is reduced to a mere move. :+1:

components/server/src/billing/entitlement-service.ts
```diff
@@ -0,0 +1,78 @@
...
+    async mayStartWorkspace(
+        user: User,
+        date: Date,
+        runningInstances: Promise<WorkspaceInstance[]>,
+    ): Promise<MayStartWorkspaceResult> {
+        return { enoughCredits: true };
+    }
+
+    async maySetTimeout(user: User, date: Date): Promise<boolean> {
+        return true;
```
AlexTugarev, line 64: really?

AlexTugarev: I see. Existing strategy for disabled payment is to allow everything.

components/server/ee/src/user/eligibility-service.spec.db.ts
```diff
@@ -147,26 +144,27 @@ class AccountServiceSpec {
...
-    async testUserGetsMoreResources_excludeFromMoreResources() {
-        await this.createTsSubscription(true);
-
-        const actual = await this.cut.userGetsMoreResources(this.user);
-        expect(
-            actual,
-            "user with Team Unleashed but excludeFromMoreResources set does not get 'more resources'",
-        ).to.equal(false);
-    }
+    // TODO(gpl) These should be moved over to EntitlementService.spec.ts
```
AlexTugarev, line 147: 🤞🏻

components/server/ee/src/user/user-service.ts
```diff
@@ -72,9 +73,10 @@ export class UserServiceEE extends UserService {
         }
     }
 
+    // TODO(gpl) Needs to fold into EntitlementService
     async userGetsMoreResources(user: User): Promise<boolean> {
         if (this.config.enablePayment) {
```
AlexTugarev, line 78: `if (this.config.enablePayment)` still needed?

components/server/src/billing/entitlement-service.ts
geropl: Exactly. This implementation gathers defaults that so far have been hidden in various places.

:bulb: But notice that this is the code for the _community_ distribution which we do not even hand out. Self-Hosted is based on the EE code, which is driven by the "new" `EntintlementServiceChargebee` (the 100% old code).

In follow-up steps we want to separate that into pure `Chargebee` and `License`, but: step-by-step :footprints:

components/server/ee/src/user/user-service.ts
```diff
@@ -17,24 +17,25 @@ import { inject } from "inversify";
...
     @inject(OssAllowListDB) protected readonly OssAllowListDb: OssAllowListDB;
     @inject(HostContextProvider) protected readonly hostContextProvider: HostContextProvider;
     @inject(Config) protected readonly config: Config;
 
+    // TODO(gpl) Needs to fold into EntitlementService
     async getDefaultWorkspaceTimeout(user: User, date: Date): Promise<WorkspaceTimeoutDuration> {
         if (this.config.enablePayment) {
             // the SaaS case
-            return this.eligibilityService.getDefaultWorkspaceTimeout(user, date);
+            return this.entitlementService.getDefaultWorkspaceTimeout(user, date);
```
AlexTugarev, line 38: `if (this.config.enablePayment)` still needed?

geropl: No, that is going to be handled by the BillingMode-guided implementation. But that's a follow-up PR (cmp. [comment](https://github.com/gitpod-io/gitpod/pull/11831/files/bd28cba0f9d730d84b730dc497832f900b245ef5#r937673793)).

For this PR it was more important to make the code 100% identical to the old impl in eligibility service.

components/server/ee/src/user/user-service.ts
geropl: https://github.com/gitpod-io/gitpod/pull/11831/files/bd28cba0f9d730d84b730dc497832f900b245ef5#r937675184

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -404,7 +406,7 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
      * gitpod.io Extension point for implementing eligibility checks. Throws a ResponseError if not eligible.
      */
     protected async maySetTimeout(user: User): Promise<boolean> {
-        return this.eligibilityService.maySetTimeout(user);
+        return this.entitlementService.maySetTimeout(user, new Date());
```
AlexTugarev, line 409: looking at `maySetTimeout`, the date param seems to be weird. it's used internally to express "until now", thus `new Date()`. but it doesn't seem to make any sense for `maySetTimeout`, does it? e.g. `maySetTimeout(user, yesterday)` is irrelevant. let's remove it, wdyt?

AlexTugarev (approved):
Awesome! Thanks for the first chunk!

/hold 
just in case you want to do something about https://github.com/gitpod-io/gitpod/pull/11831/files#r937678415

components/server/ee/src/workspace/gitpod-server-impl.ts
geropl: > but it doesn't seem to make any sense for maySetTimeout, does it? 

:information_source:  It's mainly there for testability.

But I agree, it should have a default to not bother clients. Will change in a follow-up! :+1:

geropl:
Thank you for your review @AlexTugarev ! :pray: 

/unhold
to move forward and triple-check on staging.

PR #11832, opened 2022-08-03
[installer] Support configuration of the public API HTTP port

andrew-farries:
## Description

As of https://github.com/gitpod-io/gitpod/pull/11806, the public api exposes an HTTP port for a Stripe webhook handler. This handler should be present on Gitpod SaaS but absent on self hosted installations.

This PR adds experimental config for the public api component to allow an HTTP port to be configured. If the port is not present in the configuration, `baseserver` ensures that HTTP is not served.

## Related Issue(s)

Part of https://github.com/gitpod-io/gitpod/issues/10937 and #9036 .

## How to test

Installer unit tests to check the rendered public api configmap.

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/installer/pkg/components/public-api-server/configmap.go
+++ b/install/installer/pkg/components/public-api-server/configmap.go
@@ -28,6 +28,9 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 				GRPC: &baseserver.ServerConfiguration{
 					Address: fmt.Sprintf(":%d", GRPCContainerPort),
 				},
+				HTTP: &baseserver.ServerConfiguration{
+					Address: fmt.Sprintf(":%d", HTTPContainerPort),
+				},
 			},
 		},
 	}
--- a/install/installer/pkg/components/public-api-server/configmap_test.go
+++ b/install/installer/pkg/components/public-api-server/configmap_test.go
@@ -29,6 +29,9 @@ func TestConfigMap(t *testing.T) {
 				GRPC: &baseserver.ServerConfiguration{
 					Address: fmt.Sprintf(":%d", GRPCContainerPort),
 				},
+				HTTP: &baseserver.ServerConfiguration{
+					Address: fmt.Sprintf(":%d", HTTPContainerPort),
+				},
 			},
 		},
 	}
--- a/install/installer/pkg/components/public-api-server/constants.go
+++ b/install/installer/pkg/components/public-api-server/constants.go
@@ -10,4 +10,7 @@ const (
 	GRPCPortName      = "grpc"
 	GRPCContainerPort = 9001
 	GRPCServicePort   = 9001
+	HTTPContainerPort = 9002
+	HTTPServicePort   = 9002
+	HTTPPortName      = "http"
 )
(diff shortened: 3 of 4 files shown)


MrSimonEmms (approved):

install/installer/pkg/config/v1/experimental/experimental.go
```diff
@@ -205,7 +205,8 @@ type ProxyConfig struct {
 }
 
 type PublicAPIConfig struct {
-	Enabled bool `json:"enabled"`
+	Enabled  bool  `json:"enabled"`
+	HttpPort int32 `json:"httpPort"`
```
easyCZ, line 209: Why should it be configurable? I think we want it on 9002 always

andrew-farries: We want the public API to serve HTTP in Gitpod SaaS but not in self-hosted installations. If the `httpPort` setting is omitted, then the public API won't start the HTTP server (because `baseserver` won't start listening without a configured port).

Would you prefer the setting to be a boolean (`serveHTTP`)? Then we can hardcode the port to `9002` if it's set.

easyCZ: Hmm, I think we should instead run the server and just have a No-OP implementation of the endpoint. Turning a whole HTTP server off in self-hosted is bound to bite us when we go to roll out. In practice. you'd need the same config on the proxy and other places and it just gets too complicated. It's better if it runs, just returns "Unvailable in Self-hosted".

andrew-farries: Not starting the HTTP server in self-hosted is cleaner as then there is less redundant, SaaS-only stuff floating around in a self-hosted instance.

I think the config we need is:
* This PR.
* Make the `proxy` config that we need to add to expose this HTTP endpoint optional (with another installer PR very similar to this one but for `proxy`).

I'm not sure it's inevitably going to go wrong at rollout, as we have good test coverage here with regards to the `yaml` rendered by the installer.

easyCZ: Currently, the Public API server is in experimental and not deployed in Self-Hosted anyway. I'd avoid adding extra complexity at this stage which further changes which parts of the server are running and are not - it's already hard to understand what is running and what is not.

Basically, I'd drop the config for the HTTP port and always run the HTTP server on 9002

csweichel: > Basically, I'd drop the config for the HTTP port and always run the HTTP server on 9002

Was just typing the same comment - 💯 agree

andrew-farries: Ok, so I'll drop the new config surface here and always have the public api open for HTTP on port 9002, even in self-hosted. That means closing https://github.com/gitpod-io/ops/pull/3637 and https://github.com/gitpod-io/gitpod/pull/11833.

We can think about returning a no-op response for self hosted separately.

Do you agree @easyCZ ?

easyCZ (approved):

PR #11844, opened 2022-08-03
Implement preview->certificate owner reference

ArthurSens:

## Description

A lot of certificates are left hanging around because it heavily ties preview and certificate deletion. In case for some random reason a preview is deleted and the certificate isn't, the certificate will stay in our cluster forever.

This PR aims to refactor the certificate deletion, adopting something similar to Kubernetes Owner references.

1. To delete a certificate the cron first deletes a certain preview and the certificate is left there untouched. 
2. A following cron will be responsible for checking that we now have a orphan certificate, and it should be deleted.

Opening it as a draft because I have no idea how to test that 😅 

## How to test

I'm not sure, and open to ideas 😅 

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/.werft/jobs/build/deploy-to-preview-environment.ts
+++ b/.werft/jobs/build/deploy-to-preview-environment.ts
@@ -505,6 +505,7 @@ export async function issueMetaCerts(
     certName: string,
     certsNamespace: string,
     domain: string,
+    branch: string,
     slice: string,
 ): Promise<boolean> {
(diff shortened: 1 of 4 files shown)


ArthurSens:
Superseeded by https://github.com/gitpod-io/gitpod/pull/11855 and https://github.com/gitpod-io/gitpod/pull/11856

PR #11855, opened 2022-08-03
previews: Add preview owner reference to certificates

ArthurSens:
Signed-off-by: ArthurSens <arthursens2005@gmail.com>

## Description

The same thing as https://github.com/gitpod-io/gitpod/pull/11844, but more carefully. I'm splitting adding owner reference to the certificate and the garbage collection logic into separate PRs. 

The main reason for this separation is that the garbage collection will delete any certificate without an owner's reference.

A chained PR will implement the garbage collection logic(https://github.com/gitpod-io/gitpod/pull/11856)

## How to test

Open a workspace from this PR and run:
```
werft run github -a with-preview -f
```

Check if the generated certificate has an owner annotation referencing the correct preview.
```
kubectl --context dev describe certificate -n certs <certificate-name>
```

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

PR #11856, opened 2022-08-03
previews: Implement certificate garbage collection

ArthurSens:

## Description

A follow-up from #11855.

This PR aims to refactor the certificate deletion, adopting something similar to Kubernetes Owner references.

1. To delete a certificate the cron first deletes a certain preview and the certificate is left there untouched.
2. A following cron will be responsible for checking that we now have an orphan certificate, and it should be deleted.

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

## Noteworthy mentions

We should merge #11855 first and wait a few days to merge this one.

The reason is that the garbage collection will delete certificates without owners, and we still have preview environments that don't have such reference. 

Once all active previews have the owner reference, then it is safe to merge this one.

### DIFF
--- a/.werft/platform-delete-preview-environments-cron.ts
+++ b/.werft/platform-delete-preview-environments-cron.ts
@@ -9,7 +9,7 @@ import * as VM from "./vm/vm";
 
 // for testing purposes
 // if set to 'true' it shows only previews that would be deleted
-const DRY_RUN = true;
+const DRY_RUN = false;
 
 const SLICES = {
     CONFIGURE_ACCESS: "Configuring access to relevant resources",
--- a/.werft/platform-delete-preview-environments-cron.ts
+++ b/.werft/platform-delete-preview-environments-cron.ts
@@ -1,7 +1,6 @@
 import { Werft } from "./util/werft";
 import * as Tracing from "./observability/tracing";
 import { SpanStatusCode } from "@opentelemetry/api";
-import { wipePreviewEnvironmentAndNamespace, helmInstallName, listAllPreviewNamespaces } from "./util/kubectl";
 import { exec } from "./util/shell";
 import { previewNameFromBranchName } from "./util/preview";
 import { CORE_DEV_KUBECONFIG_PATH, HARVESTER_KUBECONFIG_PATH, PREVIEW_K3S_KUBECONFIG_PATH } from "./jobs/build/const";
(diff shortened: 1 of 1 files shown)


.werft/platform-delete-preview-environments-cron.ts
```diff
@@ -10,7 +9,7 @@ import * as VM from "./vm/vm";
 
 // for testing purposes
 // if set to 'true' it shows only previews that would be deleted
-const DRY_RUN = false;
+const DRY_RUN = true;
```
mads-hartmann, line 12: Remember to remove this change before taking it out of draft ☺️

.werft/platform-delete-preview-environments-cron.ts
```diff
@@ -422,11 +422,44 @@ async function removePreviewEnvironment(previewEnvironment: PreviewEnvironment)
...
+    .stdout.trim()
+    .replace(/preview-/g, "")
+    .split("\n");
+
+    certificates.forEach(certificate => {
+        const owner = exec(
+            `kubectl --kubeconfig ${CORE_DEV_KUBECONFIG_PATH} get certificates -n ${certificatesNamespace} -o=custom-columns=:metadata.annotations.preview/owner`,
+            { slice: SLICES.DELETING_ORPHAN_CERTIFICATES, silent: true, async: false },
+        ).stdout.trim()
+
```
mads-hartmann, line 449: How about ignoring certificates with an empty owner so we can merge this without having to worry about "old" certificates from before we started tracking the owner. Something like:

```ts
if (!owner) {
  werft.log(SLICES.DELETING_ORPHAN_CERTIFICATES, `Certificate ${certificate} does not have an owner. Skipping.`)
  return
}

mads-hartmann (commented):
Two minor comments, otherwise LGMT. I won't approve as it is still in draft ☺️

.werft/platform-delete-preview-environments-cron.ts
ArthurSens: I'm looking at the current previews and certificates now, I believe there's no need to no merge at the moment 🙂. We only have orphan certificates, besides one that the owner preview is almost 2 weeks old (not sure why it wasn't deleted yet)

ArthurSens:
Here are the logs from the job run in `DRY_RUN` mode: https://werft.gitpod-dev.com/job/gitpod-custom-as-preview-garbage-collection.6

I believe it is good to go 🙂

mads-hartmann (approved):
🛹

PR #11863, opened 2022-08-04
[dashboard] Optimization: Don't wait before fetching a team's spending limit

jankeromnes:
## Description

Optimization: Don't wait for the portal URL before fetching a team's spending limit -- there is no dependency between the two, so they can be fetched in parallel.

## Related Issue(s)

Fixes #

## How to test

1. Subscribing a team to UBB and setting a spending limit should still work

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/dashboard/src/teams/TeamUsageBasedBilling.tsx
+++ b/components/dashboard/src/teams/TeamUsageBasedBilling.tsx
@@ -56,9 +56,11 @@ export default function TeamUsageBasedBilling() {
             return;
         }
         (async () => {
-            const portalUrl = await getGitpodService().server.getStripePortalUrlForTeam(team.id);
+            const [portalUrl, spendingLimit] = await Promise.all([
+                getGitpodService().server.getStripePortalUrlForTeam(team.id),
+                getGitpodService().server.getSpendingLimitForTeam(team.id),
+            ]);
             setStripePortalUrl(portalUrl);
-            const spendingLimit = await getGitpodService().server.getSpendingLimitForTeam(team.id);
             setSpendingLimit(spendingLimit);
         })();
     }, [team, stripeSubscriptionId]);


components/dashboard/src/teams/TeamUsageBasedBilling.tsx
```diff
@@ -56,9 +56,11 @@ export default function TeamUsageBasedBilling() {
             return;
         }
         (async () => {
-            const portalUrl = await getGitpodService().server.getStripePortalUrlForTeam(team.id);
+            const [portalUrl, spendingLimit] = await Promise.all([
```
easyCZ, line 59: Does using `Promise.all` limit us in terms of showing meaningful errors to the user? As per the docs, 
>  It rejects immediately upon any of the input promises rejecting or non-promises throwing an error, and will reject with this first rejection message / error.

That means we can't show what didn't work, but also may also not get the data for the other to at least make that part of it functional.

jankeromnes: > That means we can't show what didn't work, but also may also not get the data for the other to at least make that part of it functional.

This is already the case when you load things sequentially (e.g. if the portal URL call fails, we're already not fetching the limit).

---

Also, the likelihood of errors in these calls is small (and if you do get any error with these, it's likely that something is severely wrong with UBB in general and you should fix that first before doing anything else).

On the other hand, loading things sequentially makes the UI slow for every user on every single page load. An acceptable UI performance in 99.99% of page loads is more important than a tweaked error handling in 0.01% cases.

(And finally, loading things sequentially is not even the right way to properly surface errors to users -- instead, we could start by showing proper error states in the UI, instead of silently failing / logging to the console -- however, this is unrelated to this PR.)

easyCZ: Sorry, this comment wasn't about whether doing it sequentially is the right thing or not. It's not, and of course it can be loaded in parallel. My main point is that it's using a pattern that we more-less want to discourage, because it obscures the issue of not correctly showing an error to the user for each call.

easyCZ (approved):
Functionally stays the same

components/dashboard/src/teams/TeamUsageBasedBilling.tsx
jankeromnes: Aha, interesting point, thanks for clarifying!

I guess we'll want to eventually add a try/catch, and then properly report errors to the user. I'm still not fully convinced that we'd want to very precisely separate all the calls & whether they succeeded or failed, because I believe it's enough to show the first error, so that you can resolve it, and then show the next error etc.

PR #11870, opened 2022-08-04
[gpctl] Add workspace type to workspaces list

easyCZ:
## Description

Add Type and Pod to output of `workspaces list`. This allows on-callers responding to https://github.com/gitpod-io/runbooks/blob/main/runbooks/GitpodWorkspaceStuckOnStarting.md more easily understand the type of workspace

## Related Issue(s)

Fixes #

## How to test

1. [Get kube context](https://www.notion.so/gitpod/How-to-connect-to-a-cluster-2d059f96124a4976ab56599dc1a29c32#d584cca8fa194f70a20586165ca5a883)
2. Run `cd dev/gpctl && go run main.go workspaces list`
3. Observe the output columns containing type and pod. 

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/dev/gpctl/cmd/workspaces-list.go
+++ b/dev/gpctl/cmd/workspaces-list.go
@@ -6,7 +6,7 @@ package cmd
 
 import (
 	"context"
-
+	"fmt"
 	"github.com/spf13/cobra"
 
 	"github.com/gitpod-io/gitpod/common-go/log"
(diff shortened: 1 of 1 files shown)


Furisto:
Nice! I like the addition of the type, not so sure about including the pod name. The pod name is ws|prebuild + instanceid and we already have the instance id in the third column and now we also have the type, so the pod name does not convey any new information.

easyCZ:
@Furisto That's fair. And I'd expect that from workspace team engineers. But with engineers from other teams being on-call for these, they are not always aware of the mapping between a workspace type and instance. Even here it's non-obvious because `Regular` workspaces use `ws` while prebuilds do use the full `prebuild`.

For me personally, this would've saved me a bunch of time during on-call, and especially during the first time handling the alert as I wasn't directly aware of the mapping between ws type and pod.

Furisto:
I see. We should have a FAQ kind of document that explains these things. Until we have that I am fine with including the  pod name. cc @kylos101 for awareness

Furisto (approved):

PR #11873, opened 2022-08-04
Retry doFinalize if gRPC status code is Unavailable

### LINKED ISSUE #11861: If we failed within finalizeWorkspaceContent, the pod keeps in Terminating state
### Bug description

If we fail within the function finalizeWorkspaceContent, the pod keeps Terminating because the annotation `gitpod.io/startedDisposal` is there.

### Steps to reproduce

image
image
image

https://github.com/gitpod-io/gitpod/blob/63f3bb78ae473e34febebcb901094a5026166769/components/ws-manager/pkg/manager/monitor.go#L1275
https://github.com/gitpod-io/gitpod/blob/63f3bb78ae473e34febebcb901094a5026166769/components/ws-manager/pkg/manager/monitor.go#L1413-L1417

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

https://github.com/gitpod-io/gitpod/pull/11826#issuecomment-1204655217

jenting:
## Description

Retry doFinalize if gRPC status code is Unavailable

## Related Issue(s)

Fixes #11861

## How to test

None

## Release Notes

```release-note
None
```

## Documentation

None

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/ws-manager/pkg/manager/monitor.go
+++ b/components/ws-manager/pkg/manager/monitor.go
@@ -681,23 +681,34 @@ func (m *Monitor) waitForWorkspaceReady(ctx context.Context, pod *corev1.Pod) (e
 		_, err = snc.WaitForInit(ctx, &wsdaemon.WaitForInitRequest{Id: workspaceID})
 		return err
 	})
-	if st, ok := grpc_status.FromError(err); ok && st.Code() == codes.NotFound {
-		// Looks like we have missed the CREATING phase in which we'd otherwise start the workspace content initialization.
-		// Let's see if we're initializing already. If so, there's something very wrong because ws-daemon does not know about
-		// this workspace yet. In that case we'll run another desperate attempt to initialize the workspace.
-		m.initializerMapLock.Lock()
-		if _, alreadyInitializing := m.initializerMap[pod.Name]; alreadyInitializing {
-			// we're already initializing but wsdaemon does not know about this workspace. That's very bad.
-			log.WithFields(wsk8s.GetOWIFromObject(&pod.ObjectMeta)).Error("we were already initializing but wsdaemon does not know about this workspace (bug in ws-daemon?). Trying again!")
-			delete(m.initializerMap, pod.Name)
-		}
-		m.initializerMapLock.Unlock()
 
-		// It's ok - maybe we were restarting in that time. Instead of waiting for things to finish, we'll just start the
-		// initialization now.
-		err = m.initializeWorkspaceContent(ctx, pod)
-	} else {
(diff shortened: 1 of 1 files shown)


kylos101:
@jenting can you share in https://github.com/gitpod-io/gitpod/issues/11861 why this PR is related to but does not fix the issue, and maybe next steps for #11861, after this PR is merged?

jenting:
> @jenting can you share in #11861 why this PR is related to but does not fix the issue, and maybe next steps for #11861, after this PR is merged?

Well, I was thinking that should we handle a case that the backup thru. ws-daemon keeps failure https://github.com/gitpod-io/gitpod/blob/bd16387c15101d37763fa84a9707f49b4b9e3e9a/components/ws-manager/pkg/manager/monitor.go#L1326-L1335, should we remove the annotation `gitpod.io/startedDisposal` to make the ws-manager be able to call finalizerWorkspaceContent again? https://github.com/gitpod-io/gitpod/blob/bd16387c15101d37763fa84a9707f49b4b9e3e9a/components/ws-manager/pkg/manager/monitor.go#L495-L497

But it has a drawback that it keeps calling ws-daemon APIs even if the ws-daemon is gone.

sagor999:
>, should we remove the annotation gitpod.io/startedDisposal to make the ws-manager be able to call finalizerWorkspaceContent again?

Keep in mind, that ws-manager will act ONLY if there was some sort of an update to the pod. If there was none, then `actOnPodEvent` will not be called, and so ws-manager will not do anything.

components/ws-manager/pkg/manager/monitor.go
```diff
@@ -1254,6 +1254,17 @@ func (m *Monitor) finalizeWorkspaceContent(ctx context.Context, wso *workspaceOb
 		return true, gitStatus, err
 	}
 
+	handleGRPCError := func(ctx context.Context, err error) error {
```
sagor999, line 1257: This will make code harder to read.
We already have function `handleGRPCError` here: 
https://github.com/gitpod-io/gitpod/blob/a257b53a17f7fc0b9f9ea36d28f9ccc6f6087958/components/ws-manager/pkg/manager/monitor.go#L1414

So overloading it here with an inline function will be confusing later on.

Can we instead check if err is Unavailable prior to calling handleGRPCError and handle that case?

sagor999 (changes requested):

components/ws-manager/pkg/manager/monitor.go
jenting: fixed

sagor999 (approved):

PR #11876, opened 2022-08-04
[installer] Small render test improvements

corneliusludmann:
## Description
This PR updates the installer render tests:
- It resets the seed for every test case. Otherwise, adding a test case would lead to changes in other tests because the pseudo-random generator starts with a different state.
- It allows to create a golden file even when there is not file yet.

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


MrSimonEmms (approved):
Good catch

Issue #10359, opened 2022-05-30
Split the Gitpod app from the Workspace images in the KOTS installer


mrzarquon:
## Is your feature request related to a problem? Please describe
The latest airgap bundle is now 11GB, because of how the Replicated web installation works, users are required to upload the entire .airgap file to the UI even if they've already staged the images. This is because the application payload is in the same bundle as the images.

The massive airgap bundle size means that doing any deployment with airgap is slow and difficult, especially when using the web UI which is required if you're not familiar with the CLI only methods to perform an installation (or upgrade).

ie, intellij java image is 3gb, what use of this is to a vscode only installation?

Related to https://github.com/gitpod-io/gitpod/issues/10298 as we'd need to support having different registries for different purposes in Gitpod - one for workspaces vs one for hosting Gitpod's specific images.

## Describe the behaviour you'd like
Only the very minimal base image should ship with the installer, so that upgrades and installations don't require pushing > 8gb of unnecessary images. The configuration of the default image and registry should be exposed as well, so users can change this as they go.

Documentation on how to pull/push other workspace images should be provided as an alternative. Workspace images could be populated with other registry mirroring systems since in essence it is requiring the users to grab `latest` of the workspace images (which is what we do when we build an airgap anyway).

## Related Context
- Internal [RFC](https://www.notion.so/gitpod/RFC-Keep-Airgap-bundles-small-563b5695c48f4da98439a12da54af27b) related to this

mrzarquon:
Removing all IDEs except vscode and switching to workspace-base instead of full, I was able to re-compression the airgap bundle to 2.7gb. The VSCode web IDE image is 700mb while the VScode remote image is 14mb, so doesn't make sense to remove the remote image. That would get to a "hello $gitpod" level of use.

corneliusludmann:
Experimented with a script that creates an air-gap bundle with images that have changed between two releases only: https://github.com/gitpod-io/gitpod/compare/clu/diff-airgap-bundle

Still in an experimental state. However, looks quite promising. We could add something like this to our build pipeline to create airgap diff bundles for each release.

MrSimonEmms:
I had a bit of brainwave about this the other day and suspect this might be a good way to go.

[...]

PR #11878, opened 2022-08-04
[kots]: only add to base image allow list in airgapped mode

MrSimonEmms:
## Description

Change the allowlist so it only amends the `privateBaseImageAllowList` in airgapped mode. As the Replicated proxy URLs were getting added (pointlessly - the base image is never them), this only needs adding for airgapped mode.

`docker.io` is still added in when in "airgapped mode" as this would be activated if someone is just mirroring their container images (this is possible in KOTS even if the license doesn't allow for airgapped installations).

The [codebase in question](https://github.com/gitpod-io/gitpod/blob/main/components/image-builder-mk3/pkg/auth/auth.go#L120-L153)

### Non Airgapped
- Deploy via KOTS in non-airgapped mode
- Run `kubectl get configmaps -n gitpod gitpod -o jsonpath='{.data.config\.yaml}'`
- Should see the following:
```yaml
containerRegistry:
  privateBaseImageAllowList: []
```

### Airgapped
- Deploy via KOTS in airgapped mode
- Run `kubectl get configmaps -n gitpod gitpod -o jsonpath='{.data.config\.yaml}'`
- Should see the following:
```yaml
containerRegistry:
  privateBaseImageAllowList:
  - <airgapped-hostname>
  - docker.io
```

## Release Notes

```release-note
[kots]: add docker.io to registry allowlist
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/kots/manifests/gitpod-installer-job.yaml
+++ b/install/kots/manifests/gitpod-installer-job.yaml
@@ -144,19 +144,20 @@ spec:
 
               if [ '{{repl HasLocalRegistry }}' = "true" ];
               then
-                echo "Gitpod: configuring mirrored container registry"
+                echo "Gitpod: configuring mirrored container registry for airgapped installation"
 
                 yq e -i ".repository = \"{{repl LocalRegistryAddress }}\"" "${CONFIG_FILE}"
                 yq e -i ".imagePullSecrets[0].kind = \"secret\"" "${CONFIG_FILE}"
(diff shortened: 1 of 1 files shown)


mrzarquon:
This looks good for now - we need to decide how to handle defaultbaseimage / expose it better as a setting (see also: workspace-base as default in the future -> https://github.com/gitpod-io/gitpod/issues/10359 ) but since we know docker.io/gitpod/workspace-full:latest is our current defaultbaseimage, it should be allowed

nandajavarma (approved):

Issue #11854, opened 2022-08-03
fatal: 'origin/branch-name' is not a commit and a branch 'branch-name' cannot be created from it


kylos101:
### Bug description

In `gen59`, we see errors like:
![image](https://user-images.githubusercontent.com/1272076/182715613-e09d1c82-9066-4af4-ae04-e62d4687fff1.png)

We must be hitting [this error](https://github.com/gitpod-io/gitpod/blob/e4842b96a389c5a4ccd0d1d44f74fb6990e4fb7e/components/content-service/pkg/initializer/git.go#L147), which gets [traced here](https://github.com/gitpod-io/gitpod/blob/e4842b96a389c5a4ccd0d1d44f74fb6990e4fb7e/components/content-service/pkg/initializer/git.go#L141-L142).

### Steps to reproduce

Not sure. 

Maybe try starting a workspace, and delete the remote branch before the workspace starts, ideally after context parsing, before loading the content.

### Workspace affected

n/a

### Expected behavior

1. We should log this as an error, not debug, so we can see in the logs w/o enabling debug
2. We should return a clear message to the user when this happens.
3. We shouldn't retry so many times for so long

![image](https://user-images.githubusercontent.com/1272076/182716689-22397e95-2276-48f7-b47f-53435610300e.png)

jenting:
I check the spans and corresponding logs, a case happens in prebuild, and the remote branch does not exist anymore.

kylos101:
[...]

PR #11895, opened 2022-08-04
[content-init] do not hide error during git checkout

sagor999:
## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/11854

## Release Notes

```release-note
none
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/content-service/pkg/initializer/git.go
+++ b/components/content-service/pkg/initializer/git.go
@@ -144,7 +144,7 @@ func (ws *GitInitializer) realizeCloneTarget(ctx context.Context) (err error) {
 	if ws.TargetMode == RemoteBranch {
 		// create local branch based on specific remote branch
 		if err := ws.Git(ctx, "checkout", "-B", ws.CloneTarget, "origin/"+ws.CloneTarget); err != nil {
-			log.WithField("remoteURI", ws.RemoteURI).WithField("branch", ws.CloneTarget).Error("Remote branch doesn't exist.")
+			log.WithError(err).WithField("remoteURI", ws.RemoteURI).WithField("branch", ws.CloneTarget).Error("Remote branch doesn't exist.")
 			err = checkGitStatus(err)
 			return err
 		}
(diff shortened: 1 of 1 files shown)


components/content-service/pkg/initializer/git.go
```diff
@@ -144,12 +144,14 @@ func (ws *GitInitializer) realizeCloneTarget(ctx context.Context) (err error) {
 	if ws.TargetMode == RemoteBranch {
 		// create local branch based on specific remote branch
 		if err := ws.Git(ctx, "checkout", "-B", ws.CloneTarget, "origin/"+ws.CloneTarget); err != nil {
-			log.WithField("remoteURI", ws.RemoteURI).WithField("branch", ws.CloneTarget).Debug("Remote branch doesn't exist.")
-			return nil
+			log.WithField("remoteURI", ws.RemoteURI).WithField("branch", ws.CloneTarget).Error("Remote branch doesn't exist.")
+			err = checkGitStatus(err)
```
jenting, line 148: Not understand why we check git statu at here 🤔 
https://github.com/gitpod-io/gitpod/blob/7640f4baf33c6840e3f5739f60214622b7db9b3c/components/content-service/pkg/initializer/git.go#L186-L194

sagor999: My understanding is that any git checkout command can encounter that Cloudflare error code, so I think we should handle it in every case?

components/content-service/pkg/initializer/git.go
```diff
@@ -144,12 +144,14 @@ func (ws *GitInitializer) realizeCloneTarget(ctx context.Context) (err error) {
 	if ws.TargetMode == RemoteBranch {
 		// create local branch based on specific remote branch
 		if err := ws.Git(ctx, "checkout", "-B", ws.CloneTarget, "origin/"+ws.CloneTarget); err != nil {
-			log.WithField("remoteURI", ws.RemoteURI).WithField("branch", ws.CloneTarget).Debug("Remote branch doesn't exist.")
-			return nil
+			log.WithField("remoteURI", ws.RemoteURI).WithField("branch", ws.CloneTarget).Error("Remote branch doesn't exist.")
```
utam0k, line 147:
```suggestion
			log.WithError(err).WithField("remoteURI", ws.RemoteURI).WithField("branch", ws.CloneTarget).Error("Remote branch doesn't exist.")
```

components/content-service/pkg/initializer/git.go
jenting: Sounds reasonable. 💯

jenting (approved):

PR #11904, opened 2022-08-05
[public-api] Allow unknown fields in public-api-server config

andrew-farries:
## Description

Allow unknown fields in the `public-api-server` config file.

Disallowing unknown fields guards against typos in the config file but makes it harder to deploy config changes independently of code changes. IMO the latter is worth more than the former.

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/public-api-server/cmd/root.go
+++ b/components/public-api-server/cmd/root.go
@@ -60,7 +60,6 @@ func getConfig() *config.Configuration {
 
 	var cfg config.Configuration
 	dec := json.NewDecoder(bytes.NewReader(ctnt))
-	dec.DisallowUnknownFields()
 	err = dec.Decode(&cfg)
 	if err != nil {
 		log.WithError(err).Fatal("Cannot decode configuration. Maybe missing --config?")


csweichel:
How can we have invalid config considering the installer reuses the same config struct that the public-api-server uses?

geropl:
> How can we have invalid config considering the installer reuses the same config struct that the public-api-server uses?

It's not about installer vs comp, but gitpod-io/ops vs. gitpod-io/gitpod: Those are out-of-sync by nature. And not accounting for unknown fields blocks rollbacks :x: .

@andrew-farries Can we make it so that "unknown field" triggers at least a warning? That's the same mode we have in `installer`, and I think it's a good trade-off between "broken config" and backwards compatibility. :balance_scale:

andrew-farries:
> @andrew-farries Can we make it so that "unknown field" triggers at least a warning? That's the same mode we have in installer, and I think it's a good trade-off between "broken config" and "backwards compatibility".

That's a good idea. I'm not sure that Go's `json` decoder can do this 'out of the box' though - I'll have a look at how the installer does it.

andrew-farries:
Going to move this back to draft given that I'm away from tomorrow and I haven't found an easy way to trigger warnings on unknown fields during unmarshalling yet.

andrew-farries:
closing as I don't have time to investigate this further.

PR #11908, opened 2022-08-05
[usage] Fix max call send message size

easyCZ:
## Description

Reconciler fails with:
```
failed to update invoices: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (19548617 vs. 16777216)"
```

That's because we're sending the BilledSessions to the UpdateInvoices RPC. For now, we increase the size.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


geropl (approved):
LGTM

PR #11888, opened 2022-08-04
[usage] Implement markWorkspaceInstanceSessionAsBilled

laushinka:
## Description
Implements setBilledSession and calls it during workspace start.

## Related Issue(s)
Depends on #11999
Fixes #11884

## How to test
1. Do NOT create a "Gitpod" team to ensure that you are still under the Chargebee billing mode.
2. Run workspaces either as personal or in a team's project.
3. Check in the DB (I can also check it) that the session is added into `d_b_billed_session`. 

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x]  /werft with-payment

PR #11909, opened 2022-08-05
[usage] Proto changes for markBilledSession RPC

laushinka:
## Description
Proto changes for https://github.com/gitpod-io/gitpod/pull/11888

## Related Issue(s)

Relates ##11884

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


components/usage-api/usage/v1/billing.proto
```diff
@@ -50,3 +53,12 @@ message FinalizeInvoiceRequest {
 
 message FinalizeInvoiceResponse {
 }
+
+message MarkWorkspaceInstanceSessionAsBilledRequest {
+  string instance_id = 1;
+  google.protobuf.Timestamp instance_creation_time = 2;
```
easyCZ, line 59: What's this field used for?

components/usage-api/usage/v1/billing.proto
```diff
@@ -19,6 +19,9 @@ service BillingService {
   // FinalizeInvoice marks all sessions occurring in the given Stripe invoice as
   // having been invoiced.
   rpc FinalizeInvoice(FinalizeInvoiceRequest) returns (FinalizeInvoiceResponse) {};
+
+  // MarkWorkspaceInstanceSessionAsBilled marks an instance as billed with a billing system
+  rpc MarkWorkspaceInstanceSessionAsBilled(MarkWorkspaceInstanceSessionAsBilledRequest) returns (MarkWorkspaceInstanceSessionAsBilledResponse) {};
```
easyCZ, line 24: IMO, you could go with a simpler CRUD like approach make it:
```
UpdateBilledSession(...)
```
or if you'd like to make it more explicit, then I'd go for 
```
SetWorkspaceSessionsBilled()
```

I personally favour the simpler CRUD like APIs in the form of `Update<ResourceName>` where the payload specifies what you're updating.

easyCZ: The general principle in API design is that the API needs to be useful, but we should also limit the APIs surface. Fewer RPCs, which are more powerful (like Update<Resource>) make it easier to understand what you should use, and how to use it correctly.

components/usage-api/usage/v1/billing.proto
laushinka: The creation time is for the `From` field for the BillingSession.

components/usage-api/usage/v1/billing.proto
laushinka: The initial idea was to have this RPC call the setter, but I guess that's not very necessary 🤔

geropl: @laushinka I like @easyCZ suggestions as they foster loose coupling. 
More context from yesterday's discussion: https://gitpod.slack.com/archives/C02EN94AEPL/p1659600593291589?thread_ts=1659591950.914839&cid=C02EN94AEPL

laushinka: Will update it to one of your suggestions 👍🏼

easyCZ: The flip side is that this API is a private API. We can more-less do whatever we want. If you see value in limit scope and adding an rpc that has a very explicit purpose then that's also OK, but we should have a reason for it. It's a question what you're after, and the constraints imposed on you.

laushinka: Done, thanks!

components/usage-api/usage/v1/billing.proto
easyCZ: Ah, so a Session is identified by (id, from). If a second session exists, then the from on this second session will be the `to` for the first, correct?

Might make sense to comment in the proto with this as this may not be obivous. Or at least say that WorkspaceInstance can be split into multiple WorkspaceSession records, each with the above property

laushinka: > If a second session exists, then the from on this second session will be the `to` for the first, correct?

Correct. But I wonder if there is a better way for the `to`. Do you want to hop on Kumo sometime when you're free?

laushinka: Also I changed `instance_creation_time` to `from` since it will be converted before.

components/usage-api/usage/v1/billing.proto
```diff
@@ -50,3 +53,14 @@ message FinalizeInvoiceRequest {
 
 message FinalizeInvoiceResponse {
 }
+
+// If there are two billable sessions for this instance ID,
+// the second one's "from" will be the first one's "to"
+message SetBilledSessionRequest {
```
easyCZ, line 59: Here, we're more less encoding a BilledSession object, just as fields. I think we could do the following:

```
message BilledSession {
	 string instance_id = 1;
  	 google.protobuf.Timestamp from = 2;
  	 string system = 3;
}

message SetBilledSessionRequest {
  BilledSession session = 1;
}

message SetBilledSessionResponse {
  BilledSession session = 1;
}
```

Why? Because we're then re-using a domain object (BilledSession) for the Set (which also returns the object with it's values (and any possibly set by the creation process)

components/usage-api/usage/v1/billing.proto
```diff
@@ -50,3 +53,14 @@ message FinalizeInvoiceRequest {
...
+  string instance_id = 1;
+  google.protobuf.Timestamp from = 2;
+  string system = 3;
```
easyCZ, line 62: I think the billing system here should be a [oneof](https://developers.google.com/protocol-buffers/docs/proto3#oneof) to better encode the options there are.

components/usage-api/usage/v1/billing.proto
```diff
@@ -19,6 +19,9 @@ service BillingService {
   // FinalizeInvoice marks all sessions occurring in the given Stripe invoice as
   // having been invoiced.
   rpc FinalizeInvoice(FinalizeInvoiceRequest) returns (FinalizeInvoiceResponse) {};
+
+  // SetBilledSession marks an instance as billed with a billing system
+  rpc SetBilledSession(SetBilledSessionRequest) returns (SetBilledSessionResponse) {};
```
easyCZ, line 24: From reading the RPC, it's also unclear how `SetBillSession` will behave if such a record already exists, how if it doesn't. I think it will do an upsert. But it's fine to leave as is, it's an internal API.

easyCZ (commented):
The only blocking comment for is the use of a `oneof` for the `system` field. Feel free to follow the other suggestions, or not.

components/usage-api/usage/v1/billing.proto
laushinka: I'll leave this for now for when I continue with the implementation and have a better idea of what the response looks like 🙏🏽

components/usage-api/usage/v1/billing.proto
```diff
@@ -50,3 +53,17 @@ message FinalizeInvoiceRequest {
 
 message FinalizeInvoiceResponse {
 }
+
+// If there are two billable sessions for this instance ID,
+// the second one's "from" will be the first one's "to"
+message SetBilledSessionRequest {
+  string instance_id = 1;
+  google.protobuf.Timestamp from = 2;
+  oneof system {
```
laushinka, line 62: @easyCZ Is this correct?

easyCZ: And the value of either `chargebee` or `stripe` would be the Invoice ID, right?

easyCZ: If we do not care about the value then using an Enum may be simpler but I think we'll still want to capture the invoice ID

laushinka: > And the value of either `chargebee` or `stripe` would be the Invoice ID, right?

It would be the string "chargebee" or "stripe"

laushinka: > we'll still want to capture the invoice ID

The invoice ID won't be relevant for "chargebee" though because we will be excluding them anyway, right?

easyCZ: > It would be the string "chargebee" or "stripe"

Ah, that's not how the `oneof` works. It basically allows you to set at most one of the fields in it's block. But you're still setting them to a value.

I think use an Enum here for now, 
```
enum System {
    UNKNOWN = 0;
    CHARGEBEE = 1;
    STRIPE = 2;
 }
  System system = 4;
```

The reason for making Unknown 0 value is because the 0th entry will also be the default value and here we don't want to default to either of chargebee or stripe (we want it to be explicitly set by the caller)

easyCZ (approved):

PR #11936, opened 2022-08-05
EntitlementService Usage-Based Pricing

### LINKED ISSUE #11402: Define & Implement Entitlement Service APIs

geropl:
## Description
Based on: https://github.com/gitpod-io/gitpod/pull/11814

 1. Introduce EntitlementServiceUBP
    - [ ] TODO: Implement `hasPaidSubscription`: https://github.com/gitpod-io/gitpod/pull/11936/commits/55c01f7d01df375535df8048bdd4a625699aae91#diff-fd2c4594d5bd9aa04668bdce9e9c810d3c495192a0be434006c7181f4853260bR93
 2. Implement `BillingService.checkSpendingLimitReached ` (based on [usage](https://github.com/gitpod-io/gitpod/pull/11936/commits/b50bc8e9cb449e6574197c3f6b0d2d8766a6f6a3#diff-0caf79924a209df9577c140a9aefa82d74dab19ed1d20bd1f8597259c3d83316R61) until #11692 got merged)#
 3. Wiring, re-use of `BillingService` ([commit](https://github.com/gitpod-io/gitpod/pull/11936/commits/8d35e448211677fa762db3fde2adbef1dc43ff07))

## Related Issue(s)
Fixes #11402

## How to test
Actually, after https://github.com/gitpod-io/gitpod/pull/11814 has been tested thoroughly, this can be tested in prod by enableing `isUsageBasedPricingEnabled`, and testing the rollout flow. :upside_down_face: 

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/dashboard/src/start/CreateWorkspace.tsx
+++ b/components/dashboard/src/start/CreateWorkspace.tsx
@@ -372,8 +372,7 @@ function SpendingLimitReachedModal(p: { hints: any }) {
     const [attributedTeam, setAttributedTeam] = useState<Team | undefined>();
 
     useEffect(() => {
-        const attributionId: AttributionId | undefined =
-            p.hints && p.hints.attributionId && AttributionId.parse(p.hints.attributionId);
+        const attributionId: AttributionId | undefined = p.hints && p.hints.attributionId;
         if (attributionId) {
             // setAttributionId(attributionId);
             if (attributionId.kind === "team") {
--- a/components/server/ee/src/billing/entitlement-service-ubp.ts
+++ b/components/server/ee/src/billing/entitlement-service-ubp.ts
@@ -12,6 +12,7 @@ import {
     WORKSPACE_TIMEOUT_DEFAULT_LONG,
     WORKSPACE_TIMEOUT_DEFAULT_SHORT,
 } from "@gitpod/gitpod-protocol";
+import { AttributionId } from "@gitpod/gitpod-protocol/lib/attribution";
 import { inject, injectable } from "inversify";
 import {
     EntitlementService,
@@ -20,6 +21,7 @@ import {
 } from "../../../src/billing/entitlement-service";
 import { Config } from "../../../src/config";
 import { BillingModes } from "./billing-mode";
+import { BillingService } from "./billing-service";
 
 const MAX_PARALLEL_WORKSPACES_FREE = 4;
 const MAX_PARALLEL_WORKSPACES_PAID = 16;
@@ -32,6 +34,7 @@ export class EntitlementServiceUBP implements EntitlementService {
     @inject(Config) protected readonly config: Config;
     @inject(UserDB) protected readonly userDb: UserDB;
     @inject(BillingModes) protected readonly billingModes: BillingModes;
+    @inject(BillingService) protected readonly billingService: BillingService;
 
     async mayStartWorkspace(
         user: User,
@@ -50,19 +53,23 @@ export class EntitlementServiceUBP implements EntitlementService {
                 return undefined;
             }
         };
-        const [spendingLimitReached, hitParallelWorkspaceLimit] = await Promise.all([
-            this.checkSpendingLimitReached(user.id, date),
+        const [spendingLimitReachedOnCostCenter, hitParallelWorkspaceLimit] = await Promise.all([
+            this.checkSpendingLimitReached(user, date),
             hasHitParallelWorkspaceLimit(),
         ]);
 
         return {
-            spendingLimitReached,
+            spendingLimitReachedOnCostCenter,
             hitParallelWorkspaceLimit,
         };
     }
 
-    protected async checkSpendingLimitReached(userId: string, date: Date): Promise<boolean> {
-        return false;
+    protected async checkSpendingLimitReached(user: User, date: Date): Promise<AttributionId | undefined> {
+        const result = await this.billingService.checkSpendingLimitReached(user);
+        if (result.reached) {
+            return result.attributionId;
+        }
+        return undefined;
     }
 
     protected async getMaxParallelWorkspaces(user: User, date: Date): Promise<number> {
--- a/components/server/ee/src/billing/entitlement-service-ubp.ts
+++ b/components/server/ee/src/billing/entitlement-service-ubp.ts
@@ -1,0 +1,96 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { UserDB } from "@gitpod/gitpod-db/lib";
+import {
+    User,
+    WorkspaceInstance,
+    WorkspaceTimeoutDuration,
+    WORKSPACE_TIMEOUT_DEFAULT_LONG,
+    WORKSPACE_TIMEOUT_DEFAULT_SHORT,
+} from "@gitpod/gitpod-protocol";
+import { inject, injectable } from "inversify";
+import {
+    EntitlementService,
+    HitParallelWorkspaceLimit,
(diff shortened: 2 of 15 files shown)


geropl:
@jankeromnes @AlexTugarev This should build now. Can be moved to "ready for review" after https://github.com/gitpod-io/gitpod/pull/11814 has been merged, and this PR has been rebased on that! :+1:

AlexTugarev:
Got a inversify dependency cycle on 
```
Server 
--> GithubApp 
--> PrebuildManager 
--> WorkspaceStarter 
--> WorkspaceManagerClientProvider 
--> WorkspaceManagerClientProviderCompositeSource 
--> Symbol(WorkspaceManagerClientProviderSource) 
--> Symbol(WorkspaceManagerClientProviderSource)
```
which seems to be unrelated to this PR changes on first look. Quickly checked with current main's state to rule out issues from there, and updated once more. 
Not sure, if that comes from the multiInject thing (which happens to cause headaches sometimes) or a subtle change of imports which also causes module to be loaded differently.

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -256,39 +256,22 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
-        }
-
         const result = await this.entitlementService.mayStartWorkspace(user, new Date(), runningInstances);
         if (!result.enoughCredits) {
             throw new ResponseError(
                 ErrorCodes.NOT_ENOUGH_CREDIT,
                 `Not enough monthly workspace hours. Please upgrade your account to get more hours for your workspaces.`,
             );
         }
+        if (result.spendingLimitReachedOnCostCenter) {
```
AlexTugarev, line 266: need to verify this, but it seems to be wrong for EntitlementServiceUBP.mayStartWorkspace as it wont return `.enoughCredits` in result, thus the error mode will be mixed up. 

I suppose we need to get the signals aligned (enoughCredits -> notEnoughCredits, for instance) so that the pattern is clear.

AlexTugarev:
'Wasted 1h of debugging in the wrong direction. Turns out the problematic import is `import { BillingService } from "./billing-service"`, which is completely unrelated to the above mentioned cycle reported by inversify.

components/server/ee/src/billing/entitlement-service.ts
```diff
@@ -4,52 +4,103 @@
...
+                case "none":
+                    return this.license.mayStartWorkspace(user, date, runningInstances);
+                case "chargebee":
+                    return this.chargebee.mayStartWorkspace(user, date, runningInstances);
+                case "usage-based":
+                    return this.ubp.mayStartWorkspace(user, date, runningInstances);
+            }
+        } catch (err) {
+            log.error({ userId: user.id }, "EntitlementService error: mayStartWorkspace", err);
+            return {};
```
AlexTugarev, line 52: @jankeromnes, wdyt, if anything goes wrong (e.g. in UB computation) this basically means, no limits are enforced. the question is, do we want it to fail silently in error cases?

jankeromnes: I agree that it seems better to reject `mayStartWorkspace` if there is an error here, instead of failing silently.

---

Also, the `MayStartWorkspaceResult` type seems quite weird to me:

- Apparently it's an object that contains two keys, `enoughCredits: boolean` and `hasHitParallelWorkspaceLimit`? Are no other fields needed?

https://github.com/gitpod-io/gitpod/blob/678a6a2f0cd1a82471b5a6a75946adcce2eafcc6/components/server/ee/src/user/eligibility-service.ts#L18-L21

- Also, reading the code, it seems that we occasionally deviate from this type, e.g. by returning an empty object (`{}` here), or `undefined` (if `billingMode.mode` becomes `undefined` for some reason one day)

components/server/ee/src/workspace/gitpod-server-impl.ts
AlexTugarev: Screen Shot 2022-08-10 at 09 44 45

🎯  finally, was able to actually run the stuff and verified that's it's an issue indeed. 
going to apply the refactoring mentioned above, just because it will solve the problem in a readable way.

components/server/ee/src/billing/entitlement-service.ts
AlexTugarev: thanks for the overview @jankeromnes!

revising this shortly. 

also, the base (empty) implementation in non-ee folder is clear, which supports your comment.

AlexTugarev: @jankeromnes can you have a look at https://github.com/gitpod-io/gitpod/pull/11936/commits/bcb06f44877897302e5ec6b89d66a17d771dee48

I hope to have addressed the bullet points.

components/dashboard/src/start/CreateWorkspace.tsx
```diff
@@ -372,8 +372,7 @@ function SpendingLimitReachedModal(p: { hints: any }) {
     const [attributedTeam, setAttributedTeam] = useState<Team | undefined>();
 
     useEffect(() => {
-        const attributionId: AttributionId | undefined =
-            p.hints && p.hints.attributionId && AttributionId.parse(p.hints.attributionId);
+        const attributionId: AttributionId | undefined = p.hints && p.hints.attributionId && p.hints.attributionId;
```
jankeromnes, line 375: Below, there is a check on `attributionId.kind` -- meaning this should be an object, resulting from a `parse`, right?

I don't think using the `string` `p.hints.attributionId` makes sense (also you're using it twice here).

components/server/ee/src/billing/entitlement-service-ubp.ts
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+
+    async userGetsMoreResources(user: User, date: Date = new Date()): Promise<boolean> {
+        return this.hasPaidSubscription(user, date);
+    }
+
+    protected async hasPaidSubscription(user: User, date: Date): Promise<boolean> {
+        // TODO(gpl) UBP personal: implement!
+        return true;
```
jankeromnes, line 102: Note: As currently implemented, it reads like everyone on Usage-Based gets XL workspaces.

jankeromnes:
Problematic situation:
- I have a personal Chargebee plan + a legacy Chargebee team plan
- I also have a team that has Usage-Based enabled

In the team, I'm able to view the Usage-Based UI (e.g. Usage, Billing), but if I click on upgrade, the credit card form does not show up.

That's because the API methods `getStripePublishableKey` and `getStripeSetupIntentClientSecret` get rejected for my user, because my user is still in Chargebee mode (technically, these calls are made on behalf of a team that has Usage-Based enabled, but we're not sending through the `teamId` because it's not needed).

Possible solutions:
- Display a proper error message in the UI (e.g. "You need to cancel all your personal Chargebee plans before you can upgrade a team to Usage-Based")
- Authorize anyone to make the calls `getStripePublishableKey` (get the publicly-readable Gitpod front-end key) and `getStripeSetupIntentClientSecret` (create a session that allows you to submit credit card details)

jankeromnes (approved):
Ok, thanks @geropl and @AlexTugarev for completing this big work item.

The code looks good to me, and after testing it, it seems that all the Chargebee flows work as well as before (with one slight worry about the out-of-hour situation).

Approving so that this lands, and so that we can test all the recent changes together on staging before they go live tomorrow morning. 🤞

Allegory of this PR: 💣🛹 (a skateboard that makes me nervous 😅)

PR #11948, opened 2022-08-08
registry-facade: Remove unnecessary varilable declaration.

utam0k:
## Description

just small refactoring

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/registry-facade/pkg/registry/blob.go
+++ b/components/registry-facade/pkg/registry/blob.go
@@ -359,8 +359,7 @@ func (sbs ipfsBlobSource) HasBlob(ctx context.Context, spec *api.ImageSpec, dgst
 func (sbs ipfsBlobSource) GetBlob(ctx context.Context, spec *api.ImageSpec, dgst digest.Digest) (dontCache bool, mediaType string, url string, data io.ReadCloser, err error) {
 	log := log.WithField("digest", dgst)
 
-	var ipfsCID string
-	ipfsCID, err = sbs.source.Redis.Get(ctx, dgst.String()).Result()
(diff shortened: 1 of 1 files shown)


components/registry-facade/pkg/registry/blob.go
```diff
@@ -359,8 +359,7 @@ func (sbs ipfsBlobSource) HasBlob(ctx context.Context, spec *api.ImageSpec, dgst
 func (sbs ipfsBlobSource) GetBlob(ctx context.Context, spec *api.ImageSpec, dgst digest.Digest) (dontCache bool, mediaType string, url string, data io.ReadCloser, err error) {
 	log := log.WithField("digest", dgst)
 
-	var ipfsCID string
-	ipfsCID, err = sbs.source.Redis.Get(ctx, dgst.String()).Result()
+	ipfsCID, err := sbs.source.Redis.Get(ctx, dgst.String()).Result()
```
jenting, line 362: This is intended, because we declare the err variable at line 359.
Otherwise, we can't catch this line's error and update to line 359 err variable.

utam0k: > Otherwise, we can't catch this line's error and update to line 359 err variable.

Oh, I didn't know that. So you are saying that we can't use err for return value?

jenting: Oops, sorry my bad, I did not see at line 365, we set the error as `err = distv2.ErrorCodeBlobUnknown`.

jenting (approved):

PR #11950, opened 2022-08-08
ws-manager: Provide more detail of DiposalStatus.

utam0k:
## Description

The `StartDisposal` annotation so far has only shown that it has started, so we will provide more detailed information.

## Related Issue(s)

Fixes #

## How to test

1. Open a workspace
2. Stop a workspace with monitoring workspace pod's annotation

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/ws-manager/pkg/manager/annotations.go
+++ b/components/ws-manager/pkg/manager/annotations.go
@@ -183,13 +183,14 @@ func (m *Manager) markDisposalStatus(ctx context.Context, workspaceID string, di
 type DisposalStatus string
 
 const (
+	DisposalEmpty    DisposalStatus = ""
 	DisposalStarted  DisposalStatus = "started"
 	DisposalRetrying DisposalStatus = "retrying"
 	DisposalFinished DisposalStatus = "finished"
 )
 
 func (ds DisposalStatus) IsDisposing() bool {
-	return !(ds == "" || ds == DisposalFinished)
+	return !(ds == DisposalEmpty || ds == DisposalFinished)
 }
 
 // workspaceVolumeSnapshotStatus stores the status of volume snapshot
--- a/components/ws-manager/pkg/manager/annotations.go
+++ b/components/ws-manager/pkg/manager/annotations.go
@@ -6,6 +6,7 @@ package manager
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
 	"time"
@@ -63,10 +64,6 @@ const (
 
 	// pvcWorkspaceVolumeSnapshotAnnotation stores volume snapshot name when snapshot was created from pvc
 	pvcWorkspaceVolumeSnapshotAnnotation = "gitpod.io/volumeSnapshotName"
-
-	// startedDisposalAnnotation sets to true when finalizeWorkspaceContent is called to prevent finalize from
-	// being called more then once, which can happen due to race between disposalStatusAnnotation update and actOnPodEvent
-	startedDisposalAnnotation = "gitpod.io/startedDisposal"
 
 	// gitpodFinalizerName is the name of the Gitpod finalizer we use to clean up a workspace
 	gitpodFinalizerName = "gitpod.io/finalizer"
@@ -171,6 +168,32 @@ type workspaceDisposalStatus struct {
 	BackupComplete bool             `json:"backupComplete,omitempty"`
 	BackupFailure  string           `json:"backupFailure,omitempty"`
 	GitStatus      *csapi.GitStatus `json:"gitStatus,omitempty"`
+	Status         DisposalStatus   `json:"status,omitempty"`
+}
+
+func (m *Manager) markDisposalStatus(ctx context.Context, workspaceID string, disposalStatus *workspaceDisposalStatus) error {
+	b, err := json.Marshal(disposalStatus)
+	if err != nil {
+		return err
+	}
+
+	err = m.markWorkspace(ctx, workspaceID, addMark(disposalStatusAnnotation, string(b)))
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+type DisposalStatus string
+
+const (
+	DisposalStarted  DisposalStatus = "started"
+	DisposalRetrying DisposalStatus = "retrying"
+	DisposalFinished DisposalStatus = "finished"
+)
+
+func (ds DisposalStatus) IsDisposing() bool {
+	return !(ds == "" || ds == DisposalFinished)
 }
 
 // workspaceVolumeSnapshotStatus stores the status of volume snapshot
--- a/components/ws-manager/pkg/manager/annotations.go
+++ b/components/ws-manager/pkg/manager/annotations.go
@@ -177,11 +177,7 @@ func (m *Manager) markDisposalStatus(ctx context.Context, workspaceID string, di
 		return err
 	}
 
-	err = m.markWorkspace(ctx, workspaceID, addMark(disposalStatusAnnotation, string(b)))
-	if err != nil {
-		return err
-	}
-	return nil
+	return m.markWorkspace(ctx, workspaceID, addMark(disposalStatusAnnotation, string(b)))
 }
 
 type DisposalStatus string
--- a/components/ws-manager/pkg/manager/monitor.go
+++ b/components/ws-manager/pkg/manager/monitor.go
@@ -320,19 +320,8 @@ func actOnPodEvent(ctx context.Context, m actingManager, manager *Manager, statu
 		//               to update the disposal status to reflect this timeout situation.
 		if status.Conditions.Timeout != "" && strings.Contains(status.Conditions.Timeout, string(activityBackup)) {
 			span.LogKV("event", "timeout during backup")
-			err = func() error {
-				b, err := json.Marshal(workspaceDisposalStatus{BackupComplete: true, BackupFailure: status.Conditions.Timeout})
-				if err != nil {
-					return err
-				}
-
-				err = m.markWorkspace(ctx, workspaceID, addMark(disposalStatusAnnotation, string(b)))
-				if err != nil {
-					return err
-				}
-				return nil
-			}()
-			if err != nil {
+			ds := &workspaceDisposalStatus{Status: DisposalFinished, BackupComplete: true, BackupFailure: status.Conditions.Timeout}
+			if manager.markDisposalStatus(ctx, workspaceID, ds) != nil {
 				log.WithError(err).Error("was unable to update pod's disposal state - this will break someone's experience")
 			}
 		}
@@ -495,9 +484,16 @@ func actOnPodEvent(ctx context.Context, m actingManager, manager *Manager, statu
 			}
 		}
 
-		_, startedDisposal := wso.Pod.Annotations[startedDisposalAnnotation]
-		span.LogKV("startedDisposal", startedDisposal)
-		if terminated && !startedDisposal {
+		ds := &workspaceDisposalStatus{}
+		if rawDisposalStatus, ok := pod.Annotations[disposalStatusAnnotation]; ok {
+			err := json.Unmarshal([]byte(rawDisposalStatus), ds)
+			if err != nil {
+				log.WithError(err).Errorf("cannnot parse disposalStatusAnnotation: %s", rawDisposalStatus)
+			}
+		}
+
+		span.LogKV("disposalStatusAnnotation", ds)
+		if terminated && !ds.Status.IsDisposing() {
 			if wso.Pod.Annotations[workspaceFailedBeforeStoppingAnnotation] == util.BooleanTrueString && wso.Pod.Annotations[workspaceNeverReadyAnnotation] == util.BooleanTrueString {
 				// The workspace is never ready, so there is no need for a finalizer.
 				if _, ok := pod.Annotations[workspaceExplicitFailAnnotation]; !ok {
@@ -1017,17 +1013,10 @@ func (m *Monitor) finalizeWorkspaceContent(ctx context.Context, wso *workspaceOb
 			return
 		}
 
-		b, err := json.Marshal(disposalStatus)
+		err := m.manager.markDisposalStatus(ctx, workspaceID, disposalStatus)
 		if err != nil {
 			tracing.LogError(span, err)
-			log.WithError(err).Error("unable to marshal disposalStatus - this will break someone's experience")
-			return
-		}
-
-		err = m.manager.markWorkspace(ctx, workspaceID, addMark(disposalStatusAnnotation, string(b)))
-		if err != nil {
-			tracing.LogError(span, err)
-			log.WithError(err).Error("was unable to update pod's disposal state - this will break someone's experience")
+			log.WithError(err).Error("was unable to update pod's disposal status - this will break someone's experience")
 		}
 	}()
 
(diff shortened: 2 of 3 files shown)


components/ws-manager/pkg/manager/monitor.go
```diff
@@ -492,9 +492,18 @@ func actOnPodEvent(ctx context.Context, m actingManager, manager *Manager, statu
...
 
-		_, startedDisposal := wso.Pod.Annotations[startedDisposalAnnotation]
-		span.LogKV("startedDisposal", startedDisposal)
-		if terminated && !startedDisposal {
+		// _, startedDisposal := wso.Pod.Annotations[startedDisposalAnnotation]
+		var ds workspaceDisposalStatus
+		if rawDisposalStatus, ok := pod.Annotations[disposalStatusAnnotation]; ok {
+			err := json.Unmarshal([]byte(rawDisposalStatus), &ds)
+			if err != nil {
+				log.WithError(err).Errorf("cannnot paarse disposalStatusAnnotation: %s", rawDisposalStatus)
```
jenting, line 500: paarse -> parse.

components/ws-manager/pkg/manager/monitor.go
```diff
@@ -492,9 +492,18 @@ func actOnPodEvent(ctx context.Context, m actingManager, manager *Manager, statu
...
+				break
```
jenting, line 501: Should we break here?

components/ws-manager/pkg/manager/monitor.go
```diff
@@ -1113,10 +1115,13 @@ func (m *Monitor) finalizeWorkspaceContent(ctx context.Context, wso *workspaceOb
...
-		if err != nil {
-			tracing.LogError(span, err)
-			log.WithError(err).Error("was unable to update pod's start disposal state - this might cause an incorrect disposal state")
+		if disposalStatus.Status != "" {
+			disposalStatus.Status = DisposalStarted
+			err = m.manager.markDisposalStatus(ctx, workspaceID, disposalStatus)
+			if err != nil {
+				tracing.LogError(span, err)
+				log.WithError(err).Error("was unable to update pod's start disposal status - this might cause an incorrect disposal status")
+			}
```
jenting, line 1124: I think this part does not need, correct? Because we have this already https://github.com/gitpod-io/gitpod/blob/795b44e1d7e463394b080d5ee299a0fcae5a508f/components/ws-manager/pkg/manager/monitor.go#L1292-L1299

components/ws-manager/pkg/manager/annotations.go
```diff
@@ -169,10 +166,36 @@ func (a *annotation) Apply(dst map[string]string) (needsUpdate bool) {
 // workspaceDisposalStatus indicates the status of the workspace diposal
 type workspaceDisposalStatus struct {
 	BackupComplete bool             `json:"backupComplete,omitempty"`
+	Status         DisposalStatus   `json:"status,omitempty"`
 	BackupFailure  string           `json:"backupFailure,omitempty"`
 	GitStatus      *csapi.GitStatus `json:"gitStatus,omitempty"`
```
jenting, line 171:
```suggestion
	BackupComplete bool             `json:"backupComplete,omitempty"`
	BackupFailure  string           `json:"backupFailure,omitempty"`
	DisposalStatus DisposalStatus   `json:"disposalStatus,omitempty"`
	GitStatus      *csapi.GitStatus `json:"gitStatus,omitempty"`
```

jenting (commented):
/hold

jenting:
The unit tests fail.

utam0k:
@jenting PTAL

components/ws-manager/pkg/manager/monitor.go
```diff
@@ -492,9 +492,16 @@ func actOnPodEvent(ctx context.Context, m actingManager, manager *Manager, statu
...
+		var ds *workspaceDisposalStatus
+		if rawDisposalStatus, ok := pod.Annotations[disposalStatusAnnotation]; ok {
+			err := json.Unmarshal([]byte(rawDisposalStatus), ds)
+			if err != nil {
+				log.WithError(err).Errorf("cannnot parse disposalStatusAnnotation: %s", rawDisposalStatus)
+			}
+		}
+
+		span.LogKV("disposalStatusAnnotation", ds)
+		if terminated && (ds == nil || !ds.Status.IsWorking()) {
```
jenting, line 504: If the `ds.Status == DisposalFinished`, will we run the following code?

components/ws-manager/pkg/manager/monitor.go
```diff
@@ -1310,6 +1312,14 @@ func (m *Monitor) finalizeWorkspaceContent(ctx context.Context, wso *workspaceOb
 		if (err != nil && strings.Contains(err.Error(), context.DeadlineExceeded.Error())) ||
 			st.Code() == codes.Unavailable ||
 			st.Code() == codes.Canceled {
+			if disposalStatus.Status != DisposalRetrying {
+				disposalStatus.Status = DisposalRetrying
```
jenting, line 1316: Probably we don't need the `retrying` state, because in general, the behavior is the same as `started` state.

utam0k: Many errors have occurred most recently in the monitor. I believe that any information that can give us some clues should be given. In the case of retrying, at least once the finalizer makes it clear that it has failed. WDYT?

components/ws-manager/pkg/manager/annotations.go
```diff
@@ -171,6 +168,32 @@ type workspaceDisposalStatus struct {
...
+	Status         DisposalStatus   `json:"status,omitempty"`
+}
+
+func (m *Manager) markDisposalStatus(ctx context.Context, workspaceID string, disposalStatus *workspaceDisposalStatus) error {
+	b, err := json.Marshal(disposalStatus)
+	if err != nil {
+		return err
+	}
+
+	err = m.markWorkspace(ctx, workspaceID, addMark(disposalStatusAnnotation, string(b)))
```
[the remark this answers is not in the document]
utam0k: 👍

components/ws-manager/pkg/manager/annotations.go
```diff
@@ -171,6 +168,32 @@ type workspaceDisposalStatus struct {
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+type DisposalStatus string
+
+const (
```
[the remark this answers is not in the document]
utam0k: Sorry, may I ask you to give me more explanation?

components/ws-manager/pkg/manager/monitor.go
utam0k: fixed

components/ws-manager/pkg/manager/annotations.go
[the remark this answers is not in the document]
sagor999: I think JenTing meant that if instead of doing this:
`ds == ""` you could instead replace it with 
`ds == DisposalEmpty` which you would to const above as
`DisposalEmpty DisposalStatus = ""`

components/ws-manager/pkg/manager/annotations.go
```diff
@@ -171,6 +168,28 @@ type workspaceDisposalStatus struct {
...
+	DisposalStarted  DisposalStatus = "started"
+	DisposalRetrying DisposalStatus = "retrying"
+	DisposalFinished DisposalStatus = "finished"
+)
+
+func (ds DisposalStatus) IsWorking() bool {
```
sagor999, line 191: This is a weird function name, that I cannot quite understand.
What does it mean that it is working? Is it currently disposing? I assume that is what this function checks for due to `ds == DisposalFinished` in it, so maybe it would be better to rename this function to be more informative:
`IsDisposing` or `isCurrentlyDisposing`

components/ws-manager/pkg/manager/monitor.go
sagor999, line 504: This statement looks a bit hard to read now due to checking if ds is nil.
Would it make sense to instead above add something like this:
```
if ds == nil {
  ds = &workspaceDisposalStatus{}
}
```
then this if can become much cleaner:
`if terminated && !ds.Status.IsWorking() {`

components/ws-manager/pkg/manager/monitor.go
```diff
@@ -515,9 +522,11 @@ func actOnPodEvent(ctx context.Context, m actingManager, manager *Manager, statu
 				}
 				return m.modifyFinalizer(ctx, workspaceID, gitpodFinalizerName, false)
 			} else {
-				// We start finalizing the workspace content only after the container is gone. This way we ensure there's
-				// no process modifying the workspace content as we create the backup.
-				go m.finalizeWorkspaceContent(ctx, wso)
+				if ds == nil || ds.Status != DisposalFinished {
```
sagor999, line 525: same here with checking for `ds == nil`

components/ws-manager/pkg/manager/monitor.go
```diff
@@ -1284,6 +1280,14 @@ func (m *Monitor) finalizeWorkspaceContent(ctx context.Context, wso *workspaceOb
 	)
 	t := time.Now()
 
+	disposalStatus = &workspaceDisposalStatus{
+		Status: DisposalStarted,
+	}
+	err = m.manager.markDisposalStatus(ctx, workspaceID, disposalStatus)
```
sagor999, line 1286: shouldn't this be done on line 1116 instead?

components/ws-manager/pkg/manager/monitor.go
sagor999: Does the rest of the code work properly if it will encounter status `DisposalRetrying`? If yes, then it is probably a good idea to keep it.

components/ws-manager/pkg/manager/annotations.go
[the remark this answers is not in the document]
jenting: Correct, thanks Pavel's explanation 🙏

components/ws-manager/pkg/manager/monitor.go
jenting: I want to double confirm that do we need to add one more status? But looks like it will give us more information on this. So, I am okay to add the retrying status 💪

components/ws-manager/pkg/manager/annotations.go
jenting: Agree, I was confused by the function name until I read the code logic.

utam0k: 👍

components/ws-manager/pkg/manager/monitor.go
utam0k: this one?
https://github.com/gitpod-io/gitpod/pull/11950/files#diff-f7e47990d20d5101157fb2978f0171823bfd55a42a3ab755395b1df96911abf9R1116

sagor999: yeah. why did we need to move mark lower? :thinking: 
concerned about some unknown side effects of this.

utam0k: This is to change to a more accurate disposal start timing.

sagor999 (approved):

components/ws-manager/pkg/manager/monitor.go
```diff
@@ -1010,24 +1009,17 @@ func (m *Monitor) finalizeWorkspaceContent(ctx context.Context, wso *workspaceOb
...
-		}
-
-		b, err := json.Marshal(disposalStatus)
-		if err != nil {
-			tracing.LogError(span, err)
-			log.WithError(err).Error("unable to marshal disposalStatus - this will break someone's experience")
+		if disposalStatus.Status == DisposalEmpty {
+			span.LogKV("disposalStatus", "empty")
 			return
 		}
```
jenting, line 1017: We could consider checking if the disposalStatus be updated or not.
If not, return it. Otherwise, call the `markDisposalStatus`.
For example
```go
disposalStatus := &workspaceDisposalStatus{}
existingDisposalStatus := disposalStatus
if reflect.DeepEqual(existingDisposalStatus, disposalStatus) {
	return
}

err := m.manager.markDisposalStatus(ctx, workspaceID, disposalStatus)
if err != nil {
	tracing.LogError(span, err)
	log.WithError(err).Error("was unable to update pod's disposal status - this will break someone's experience")
}
```

components/ws-manager/pkg/manager/monitor.go
```diff
@@ -495,9 +485,16 @@ func actOnPodEvent(ctx context.Context, m actingManager, manager *Manager, statu
...
+		ds := &workspaceDisposalStatus{}
+		if rawDisposalStatus, ok := pod.Annotations[disposalStatusAnnotation]; ok {
+			err := json.Unmarshal([]byte(rawDisposalStatus), ds)
+			if err != nil {
+				log.WithError(err).Errorf("cannnot parse disposalStatusAnnotation: %s", rawDisposalStatus)
+			}
+		}
+
+		span.LogKV("disposalStatusAnnotation", ds)
+		if terminated && !ds.Status.IsDisposing() {
```
jenting, line 497: What if we change change the function as
```go
func (ds DisposalStatus) IsDisposed() bool {
	return ds == DisposalFinished
}
```

and change this line as
```go
if terminated && !ds.Status.IsDisposed() {
```

so we could remove the line 518 `if ds.Status != DisposalFinished {`

The code be simpler, WDYT?

components/ws-manager/pkg/manager/monitor.go
utam0k: Or rather, should we have it updated every time I make a change rather than mark it with a defer?

components/ws-manager/pkg/manager/monitor.go
utam0k: @jenting Thanks for your comments. https://github.com/gitpod-io/gitpod/pull/12249

components/ws-manager/pkg/manager/monitor.go
jenting: > Or rather, should we have it updated every time I make a change rather than mark it with a defer?

It depends on our code convention.

I'd vote to update the status with defer function because it makes the code simple and we only have to update it once before we exit the function. Otherwise, we might update the status within a function more than once.

utam0k: I don't know the conventions of the Go language, but is it correct to use defer that way? mark should be reflected immediately, and it seemed natural to have the code in a set with the status structure changes. Could you please tell me what the convention is?

jenting: Sorry, it's not the Go language code convention, it's more like the code convention within Gitpod, and I think we don't have the convention on when to update the status. Never mind for the defer function discussion 😄 

But, it's good to check whether the resource needs to be updated before calling the kubernetes client API to update it, ref https://github.com/kubernetes/kubernetes/blob/a1128e380c2cf1c2d7443694673d9f1dd63eb518/pkg/controller/volume/persistentvolume/pv_controller.go#L794

utam0k: Good point. I'm not sure I felt we should avoid using `DeepEqual` if possible, right?
> But, it's good to check whether the resource needs to be updated before calling the kubernetes client API to update it, ref

I wonder In what scenarios would `DeepEqual` be needed? I believe that the current comparison is sufficient. If you have a case in mind, please let me know. Thanks for the learning opportunity.

Issue #11852, opened 2022-08-03
Unexpected error loading prebuild


aledbf:
### Bug description

`rpc error: code = FailedPrecondition desc = cannot initialize workspace: prebuild initializer: Git fallback: git initializer gitClone: mkdir /dst/spring-petclinic: no such file or directory`

![Jaeger-UI (4)](https://user-images.githubusercontent.com/161571/182707827-6640bf26-9620-4d3d-b32a-6a9db15f0b3d.png)

### Workspace affected

gitpodio-springpetclini-8a38a5a57eu

### Expected behavior

1. Log when this happens (we don't now).
2. Ideally we'd wait longer until the file is ready, before trying to clone

### Example repository

None

### Anything else?

How long do we wait before the file system is ready?

sagor999:
We already have log entry for this:
https://github.com/gitpod-io/gitpod/blob/6390f2064394ebccf788f4ff5b57fd66e0313ce1/components/content-service/pkg/initializer/git.go#L75

Also looking at logs, it looks like workspace failed, and then 5 minutes later ws-daemon tried to run initializer for it (???):
![image](https://user-images.githubusercontent.com/18602811/182961665-fc02a4ee-f106-4058-931d-444c79ace3aa.png)

I cannot quite understand what exactly happened to that workspace lifecycle.

kylos101:
Thanks for looking at this one, @sagor999 ! @jenting prior to resuming PVC work, could you peek at this to see what you can find? It's the last of broken windows we found from `gen59` traces.

jenting:
🤔 According to the log, the `Location` is `/dst/spring-petclinic`. However, tracing the code, the Location should be start with the prefix `/workspace`.

https://github.com/gitpod-io/gitpod/blob/3d97d5552ec092938327c3813d53a04038c3db7f/components/supervisor/pkg/supervisor/supervisor.go#L1373

https://github.com/gitpod-io/gitpod/blob/3d97d5552ec092938327c3813d53a04038c3db7f/components/content-service/pkg/executor/executor.go#L66

https://github.com/gitpod-io/gitpod/blob/3d97d5552ec092938327c3813d53a04038c3db7f/components/content-service/pkg/initializer/initializer.go#L110
https://github.com/gitpod-io/gitpod/blob/3d97d5552ec092938327c3813d53a04038c3db7f/components/content-service/pkg/initializer/initializer.go#L124

[...]

PR #11961, opened 2022-08-08
Change mount flag from --bind to --rbind

jenting:
## Description

Change mount flag from `--bind` to `--rbind`.

## Related Issue(s)

Fixes #

## How to test

None

## Release Notes

```release-note
None
```

## Documentation

None

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/ws-daemon/pkg/content/initializer.go
+++ b/components/ws-daemon/pkg/content/initializer.go
@@ -192,7 +192,7 @@ func RunInitializer(ctx context.Context, destination string, initializer *csapi.
 		Destination: "/dst",
 		Source:      destination,
 		Type:        "bind",
-		Options:     []string{"bind", "rprivate"},
+		Options:     []string{"rbind", "rprivate"},
 	})
 
 	spec.Hostname = "content-init"


components/ws-daemon/pkg/content/initializer.go
```diff
@@ -192,7 +192,7 @@ func RunInitializer(ctx context.Context, destination string, initializer *csapi.
 		Destination: "/dst",
 		Source:      destination,
 		Type:        "bind",
-		Options:     []string{"bind", "rprivate"},
+		Options:     []string{"rbind", "rprivate"},
```
jenting, line 195: To be honest, I don't know if we should change to `rbind` here.
But I was hoping this could address the issue #11852

@csweichel For this mount point, should we need the `rbind` instead of `bind`?

sagor999: I think `bind` here is more then enough, since we should not have any other binds inside `destination` param that used as source for the bind command.

jenting: Agree with you @sagor999.

Close this PR first, and if issue #11852 still happens, we could revisit it to see what went wrong.

PR #11966, opened 2022-08-08
Replace deprecated seccomp.security.alpha.kubernetes.io/pod annotation

aledbf:
## Description

The annotation will be removed in k8s 1.25

## How to test

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


components/ws-manager/pkg/manager/create.go
```diff
@@ -514,7 +512,14 @@ func (m *Manager) createDefiniteWorkspacePod(startContext *startWorkspaceContext
 			SchedulerName:                m.Config.SchedulerName,
 			EnableServiceLinks:           &boolFalse,
 			Affinity:                     affinity,
-			SecurityContext:              &corev1.PodSecurityContext{},
+			SecurityContext: &corev1.PodSecurityContext{
+				// We're using a custom seccomp profile for user namespaces to allow clone, mount and chroot.
+				// Those syscalls don't make much sense in a non-userns setting, where we default to runtime/default using the PodSecurityPolicy.
```
jenting, line 517: The PSP is going to deprecated, correct? Then, I am not use why we keep the PSP wording within the comment.

aledbf: We don't have a replacement for PSP yet. I would prefer to keep that.

sagor999 (approved):

jenting (approved):

PR #11977, opened 2022-08-08
[IDE jobs] only post interesting notifications to Slack

### LINKED ISSUE #11771: Adjust IDE's GitHub Actions to send notifications to Slack only when they fail or get resolved
Currently GitHub Actions send notifications independently if they fail or succeed.

To reduce the number of notifications, we should post notifications only when an action fails.

filiptronicek:
## Description
This PR changes all IDE jobs which have previously posted a message to Slack every time they were run to now only post when they are interesting to us. 
This includes them failing or being fixed (going from failing to passing).

This is all achieved by this little snippet:
```yaml
- name: Get previous job's status
  id: lastrun
  uses: filiptronicek/get-last-job-status@main
  with:
    workflow-id: 18268462
- name: Slack Notification
  if: always()
  if: ${{ (success() && steps.lastrun.outputs.status == 'failed') || failure() }}
  ...
```

### Progress
- [x] https://github.com/gitpod-io/gitpod/blob/main/.github/workflows/code-nightly.yaml
- [x] https://github.com/gitpod-io/gitpod/blob/main/.github/workflows/jetbrains-auto-update-template.yml
- [x] https://github.com/gitpod-io/gitpod/blob/main/.github/workflows/jetbrains-update-plugin-platform-template.yml
- [x] https://github.com/gitpod-io/gitpod/blob/main/.github/workflows/jetbrains-updates-template.yml

## Related Issue(s)

Fixes #11771

## How to test
For the workflows that have manual triggers (can be run by a `Run Workflow` button on the [Actions page](https://github.com/gitpod-io/gitpod/actions)), you can just trigger it and wait for it to finish. If it succeeds then it should not output anything to the Slack channel and the Action step should be skipped.

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/.github/workflows/code-nightly.yaml
+++ b/.github/workflows/code-nightly.yaml
@@ -37,8 +37,6 @@ jobs:
       - name: Get previous job's status
         id: lastrun
         uses: filiptronicek/get-last-job-status@main
-        with:
-          workflow-id: 15751907
       - name: Slack Notification
         if: ${{ (success() && steps.lastrun.outputs.status == 'failed') || failure() }}
         uses: rtCamp/action-slack-notify@v2
--- a/.github/workflows/code-nightly.yaml
+++ b/.github/workflows/code-nightly.yaml
@@ -34,8 +34,13 @@ jobs:
           headCommit=$(curl -H 'Accept: application/vnd.github.VERSION.sha' https://api.github.com/repos/gitpod-io/openvscode-server/commits/gp-code/main)
           cd components/ide/code
           leeway build -Dversion=nightly -DimageRepoBase=eu.gcr.io/gitpod-core-dev/build -DcodeCommit=$headCommit -DcodeQuality=insider .:docker
+      - name: Get previous job's status
+        id: lastrun
+        uses: filiptronicek/get-last-job-status@main
+        with:
+          workflow-id: 15751907
       - name: Slack Notification
-        if: always()
+        if: ${{ (success() && steps.lastrun.outputs.status == 'failed') || failure() }}
         uses: rtCamp/action-slack-notify@v2
         env:
           SLACK_WEBHOOK: ${{ secrets.IDE_SLACK_WEBHOOK }}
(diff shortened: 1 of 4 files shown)


.github/workflows/code-nightly.yaml
```diff
@@ -34,8 +34,13 @@ jobs:
           headCommit=$(curl -H 'Accept: application/vnd.github.VERSION.sha' https://api.github.com/repos/gitpod-io/openvscode-server/commits/gp-code/main)
           cd components/ide/code
           leeway build -Dversion=nightly -DimageRepoBase=eu.gcr.io/gitpod-core-dev/build -DcodeCommit=$headCommit -DcodeQuality=insider .:docker
+      - name: Get previous job's status
+        id: lastrun
+        uses: filiptronicek/get-last-job-status@main
+        with:
+          workflow-id: 15751907
```
jeanp413, line 41: using the number id this looks tricky, it think it would be better to use the filename

andreafalzetti: I would have guessed the workflow id would appear in the GHA's URL but it doesn't. Where do you get it from?

For me it's non-blocking but I agree it's more friendly to use the filename (although I suspect the workflow ID will stay the same even if the filename changes)

filiptronicek: Thanks @jeanp413, I totally agree the file name should be used instead. The reason I chose the workflow ID is because that's what the API call wants, so I will need to first query what the ID of the corresponding file name is. Should be easy though, will work on it.

jeanp413: you can use the filename directly https://octokit.github.io/rest.js/v18#actions-list-workflow-runs

filiptronicek: Missed this: 🤦‍♂️
> You can also pass the workflow file name as a string.

filiptronicek: Should now work without specifying anything.

andreafalzetti (approved):
LGTM! 👏

PR #11993, opened 2022-08-09
[dashboard] fix host value for Git Integrations

AlexTugarev:
if `:` is included, just the preview of the callback URL should be replaced with `_`, but not actual host value.

## Related Issue(s)

Fixes #11798

## Release Notes

```release-note
Fixes entering of colons in host on Git Integration page.
```

## Werft options:

- [x] /werft with-preview


AlexTugarev:
@andrew-farries, thanks for the nudge! done.

/hold cancel

PR #11995, opened 2022-08-09
Sync terraform EKS module with the reference architecture 

### LINKED ISSUE #11533: Sync AWS `terraform` module with reference architecture

nandajavarma:
## Description

This PR adds a single-cluster AWS terraform module, corresponding to our reference architecture for the same cloud provider.

## Related Issue(s)

Fixes #11533 

## How to test

You can follow the README [added to this PR](https://github.com/gitpod-io/gitpod/tree/nvn/tf-doc/install/infra/single-cluster/aws#readme) to test this terraform module.

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


roboquat:

<details>

</details>

install/infra/modules/eks/database.tf
```diff
@@ -0,0 +1,54 @@
...
+}
+
+resource "aws_db_instance" "gitpod" {
+  count = var.enable_external_database ? 1 : 0
+
+  allocated_storage      = 10
+  max_allocated_storage  = 100
+  engine                 = "mysql"
+  engine_version         = "5.7"
+  instance_class         = "db.t3.micro"
```
lucasvaltl, line 44: I don't think the database setup here matches the reference architecture. This is what we have in the reference architecture: 
```
aws rds create-db-instance \
    --db-name gitpod \
    --db-instance-identifier gitpod-instance \
    --db-instance-class db.m5.large \
    --db-subnet-group-name gitpod-rds \
    --vpc-security-group-ids sg-0e538ccac25bb1387 \
    --multi-az \
    --engine mysql \
    --allocated-storage 20 \
    --max-allocated-storage 120 \
    --master-username gitpod \
    --master-user-password $MYSQL_GITPOD_PW \
    --engine-version 5.7 \
    --tags Key=project,Value=gitpod Key=Name,Value=Gitpod-MySQLDB,Key=department,Value=demo
```
([sourc](https://www.gitpod.io/docs/self-hosted/latest/reference-architecture/single-cluster-ref-arch#database)e)

lucasvaltl (commented):
Did not fully review all of the settings and resources used to make sure that they match the reference architecture, but I did already find that there is a deviation in the DB set up and wanted to already mention that :) 

--> For those thar review this: please make sure to review that the infrastructure created via this terraform matches the infrastructure described in the [reference architecture guide](https://www.gitpod.io/docs/self-hosted/latest/reference-architecture/single-cluster-ref-arch) exactly.

install/infra/modules/eks/kubernetes.tf
```diff
@@ -95,18 +129,18 @@ module "eks" {
...
       instance_types = [var.workspace_machine_type]
-      name = "ws-${var.cluster_name}"
-      subnet_ids   = module.vpc.public_subnets
-      min_size     = 1
-      max_size     = 10
-      block_device_mappings =  [{
+      name           = "ws-${var.cluster_name}"
+      subnet_ids     = module.vpc.public_subnets
+      min_size       = 1
+      max_size       = 10
```
lucasvaltl, line 135: We are planning to change the max node pool size here to 50, see https://github.com/gitpod-io/website/pull/2554

install/infra/modules/eks/kubernetes.tf
```diff
@@ -64,16 +98,16 @@ module "eks" {
...
+      name                       = "service-${var.cluster_name}"
+      subnet_ids                 = module.vpc.public_subnets
+      min_size                   = 1
+      max_size                   = 10
+      desired_size               = 2
+      block_device_mappings = [{
         device_name = "/dev/sda1"
 
-        ebs =  [{
+        ebs = [{
```
lucasvaltl, line 110: I think here we are recommending different settings for storage in the reg arch guide, namely:
```
    disableIMDSv1: false
    volumeSize: 300
    volumeType: gp3
    volumeIOPS: 6000
    volumeThroughput: 500
    ebsOptimized: true
    privateNetworking: true
    propagateASGTags: true
```
([source](https://www.gitpod.io/docs/self-hosted/latest/reference-architecture/single-cluster-ref-arch#kubernetes-cluster), see the gitpod-cluster.yaml file in the AWS tab of the cloud provider specific instructions)

--> I saw in the AWS cost that we are using GP2 storage and not GP3 storage in this current version of the terraform script. Not sure about the iops, throughput and the other stuff.

nandajavarma:
@lucasvaltl I have addressed the comments. Sorry on the back and forth on this. I kinda of went easy on the configurable settings considering it can/will be changed. I will keep this in mind and try to do an exact 1:1 config.

lucasvaltl:
👍 Thank you! And no worries - maybe the intent here was not clear enough from my side. We want this terraform to effectively create exactly what the guide describes :)

install/infra/modules/eks/database.tf
```diff
@@ -0,0 +1,57 @@
...
+  subnet_ids = [module.vpc.public_subnets[2], module.vpc.public_subnets[3]]
+}
+
+resource "aws_security_group" "rdssg" {
+  count = var.enable_external_database ? 1 : 0
+
+  name   = "dh-sg-${var.cluster_name}"
+  vpc_id = module.vpc.vpc_id
+
+  egress {
```
adrienthebo, line 22: Per the [terraform aws_security_group documentation](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group), it looks like this inline `egress` rule will be overwritten by the following `aws_security_group_rule`. If we drop this egress rule does the database still work?

Some further searching indicates that [inline rules are somewhat badly behaved](https://github.com/hashicorp/terraform-provider-aws/issues/1555), let's see if we can chip at this problem and keep all of the rule definitions in separate resources.

install/infra/modules/eks/database.tf
```diff
@@ -0,0 +1,57 @@
...
+  instance_class         = "db.m5.large"
+  vpc_security_group_ids = [aws_security_group.rdssg[0].id]
+  identifier             = "db-${var.cluster_name}"
+  name                   = "gitpod"
+  username               = "gitpod"
+  password               = random_password.password[0].result
+  parameter_group_name   = "default.mysql5.7"
+  db_subnet_group_name   = aws_db_subnet_group.gitpod_subnets[0].name
+  skip_final_snapshot    = true
+  publicly_accessible    = true
```
adrienthebo, line 56: Since we're restricting access to the DB to only the Gitpod node groups, does having the DB as `publicly_accessible` solve any problems for us?

install/infra/modules/eks/kubernetes.tf
```diff
@@ -1,24 +1,65 @@
...
+  to_port                  = 443
+  type                     = "ingress"
+}
+
+
 resource "aws_security_group" "nodes" {
   name   = "nodes-sg-${var.cluster_name}"
   vpc_id = module.vpc.vpc_id
 
   ingress {
```
adrienthebo, line 59: As mentioned earlier, are there any interactions of this inline rule and the separate `aws_security_group_rule` resources?

install/infra/modules/eks/storage.tf
```diff
@@ -0,0 +1,66 @@
...
+  bucket = aws_s3_bucket.gitpod-registry-backend[0].id
+  versioning_configuration {
+    status = "Enabled"
+  }
+}
+
+data "aws_iam_policy_document" "s3_policy" {
+  count = var.enable_external_storage ? 1 : 0
+  statement {
+    actions   = ["s3:*"]
```
adrienthebo, line 38: **Recommendation:** This permission set is very broad, can we restrict this to the permission set specified in the [Object Storage](https://www.gitpod.io/docs/self-hosted/latest/reference-architecture/single-cluster-ref-arch#object-storage) section? (We may need to have separate policy documents for the object storage bucket and registry bucket if they're both defined.)

install/infra/modules/eks/storage.tf
```diff
@@ -0,0 +1,66 @@
...
+    resources = ["*"]
+    effect    = "Allow"
+  }
+}
+
+resource "aws_iam_policy" "policy" {
+  count       = var.enable_external_storage ? 1 : 0
+  name        = "spolicy-${var.cluster_name}"
+  description = "s3 storage bucket policy"
```
adrienthebo, line 47: **Fixup:** users auditing their IAM policies may not be able to associate this policy with Gitpod, we can make this a bit more clear.

```suggestion
  description = "Gitpod ${var.cluster_name} object storage bucket policy"
```

install/infra/single-cluster/aws/README.md
```diff
@@ -0,0 +1,244 @@
+# Terraform setup for AWS Single-cluster Gitpod reference architecture
```
adrienthebo, line 1: This is _excellent_ documentation! Thinking longer term, how can we link up this documentation to the reference architecture docs so this gets a good level of visibility?

No action needed, resolve when you've seen this.

install/infra/single-cluster/aws/main.tf
```diff
@@ -0,0 +1,6 @@
+terraform {
+  backend "s3" {
+    bucket = "gitpod-tf"
```
adrienthebo, line 3: **Question:** If a user directly uses this module, will this backend definition conflict with their environment?

install/infra/single-cluster/aws/terraform.tfvars
```diff
@@ -0,0 +1,26 @@
+
```
adrienthebo, line 1: **Question:** If a user tries to use this module, will the presence of this `terraform.tfvars` file conflict with their environment?

Note that this question and any associated changes may be out of scope for this PR, as we're mainly focused on synchronizing the reference architecture and Terraform module. If it is then resolve this issue, we can deal with issues like this in a later PR.

adrienthebo (commented):
## Summary

Looks good overall, I see a few hot spots but nothing major.

1. Security group inline rules vs `aws_security_group_rule` instances - the docs say that inline rules and separate rule resources may not play well together. Can we check this out and see if we have any risk here? We have a couple of egress rules addressed to `0.0.0.0/0` which deviates from the reference architecture; I'd like to confirm that they're not in use so we're testing the strict policies defined in the reference architecture.
2. RDS `publicly_accessible` setting - can we drop this setting?
3. S3 security policies - can we restrict the policy to the object storage and container registry buckets?

As noted inline I have a handful of opportunistic or style comments; if any of them resonate with you then we can address them but can and should skip if the review process starts running long.

**Edit:** one more quick topic - it seems like the only way to keep track of AWS resources is with the copious use of labels on every resource. Are we using labels to identify associated resources? (If not, give me a yell and I'll spin up a ticket to add that later.)

install/infra/single-cluster/aws/main.tf
nandajavarma: This is a good point! I could leave the bucket field empty since I cannot use variables in bucket definition. But I am also thinking, the chances of them having a bucket named `gitpod-tf` is unlikely, so it will error out saying no such bucket is present. Not sure what the right route is, I might leave this as is.

install/infra/single-cluster/aws/terraform.tfvars
nandajavarma: Excellent point! I will leave those fields empty, so they will have to edit it in any case

install/infra/modules/eks/kubernetes.tf
nandajavarma: removed the inline `ingress` rule.

install/infra/modules/eks/database.tf
nandajavarma: Changed it to `false`

install/infra/modules/eks/database.tf
nandajavarma: You are absolutely right! this `egress` rule is basically unnecessary in this context. I have removed this.

install/infra/single-cluster/aws/README.md
nandajavarma: Thanks for pointing this out! We have an issue open for this here https://github.com/gitpod-io/gitpod/issues/11281

install/infra/modules/eks/storage.tf
nandajavarma: I have updated this! I think the storage file has ended up becoming a bit convoluted, but I will clean it up in the future

lucasvaltl:
👋 It seems that errors can still creep up during the `make plan` phase on the EKS side: see screenshot [here](https://gitpod.slack.com/archives/C01KLC56NP7/p1660694981508769?thread_ts=1660612821.340189&cid=C01KLC56NP7) (internal). If I remember correctly, these were the exact same errors I ran into as well. On my side, it was because I was not starting off of a clean slate. I assume that @mbrevoort was. Can we please take a look at this before we merge this?

Issue #11997, opened 2022-08-09
Fix gRPC message size error in `UpdateInvoices` call


andrew-farries:
The usage reconciler is currently failing with this error when invoking the `UpdateInvoices` RPC on the billing service:

```
{"@type":"type.googleapis.com/google.devtools.clouderrorreporting.v1beta1.ReportedErrorEvent","error":"failed to update invoices: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (33395302 vs. 16777216)","level":"error","message":"Reconciliation run failed.","serviceContext":{"service":"usage","version":"commit-8e227e13ef2caae8f56c26387a78546b9fa93b30"},"severity":"ERROR","time":"2022-08-09T11:33:09Z"}
```

PR #12003, opened 2022-08-09
[UBP/baseserver] Allow gRPC server to receive much larger messages

andrew-farries:
## Description

Increase baseserver's gRPC maximum receivable message size to 100MB. Change the `usage` client's maximum send and receive values to the same size.

The usage component is currently failing with this error:

```
{"@type":"type.googleapis.com/google.devtools.clouderrorreporting.v1beta1.ReportedErrorEvent","error":"failed to update invoices: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (33395302 vs. 16777216)","level":"error","message":"Reconciliation run failed.","serviceContext":{"service":"usage","version":"commit-8e227e13ef2caae8f56c26387a78546b9fa93b30"},"severity":"ERROR","time":"2022-08-09T11:33:09Z"}
```

This change should allow the `UpdateInvoices` RPC (and all other gRPC services registered with `baseserver`) to accept larger message sizes.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/11997

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/common-go/baseserver/server.go
+++ b/components/common-go/baseserver/server.go
@@ -298,6 +298,7 @@ func (s *Server) initializeGRPC() error {
 		opts = append(opts, grpc.Creds(credentials.NewTLS(tlsConfig)))
 	}
 
+	opts = append(opts, grpc.MaxRecvMsgSize(100*1024*1024))
 	s.grpc = grpc.NewServer(opts...)
 
 	reflection.Register(s.grpc)
--- a/components/usage/pkg/apiv1/size_test.go
+++ b/components/usage/pkg/apiv1/size_test.go
@@ -1,0 +1,59 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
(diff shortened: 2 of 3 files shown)


components/common-go/baseserver/server.go
```diff
@@ -298,6 +298,7 @@ func (s *Server) initializeGRPC() error {
 		opts = append(opts, grpc.Creds(credentials.NewTLS(tlsConfig)))
 	}
 
+	opts = append(opts, grpc.MaxRecvMsgSize(100*1024*1024))
```
jenting, line 301: Can we not change the gRPC base server default message size and let the caller overwrite the default message size? Because this is an optional for all gRPC server, WDYT?

andrew-farries: Yes, we can probably provide an option to set this. 

The question is whether a user of the baseserver would ever care that the maximum message size was too large. IMO, probably not in which case an option is not needed. Perhaps we should add the option when it's needed.

jenting (approved):

AlexTugarev (approved):

PR #12018, opened 2022-08-10
ws-daemon: ignore error when workspace persistent state location not exist

### LINKED ISSUE #11715: [ws-daemon] Unexpected error during workspace dispose

jenting:
## Description

The initial workspace failed at the beginning, so the workspace's persistent state location never exist.

## Related Issue(s)

Fixes #11715

## How to test

Not sure, but we did see some nodes path `/var/gitpod/workspaces` with the folders
- `<instance-id>`
- `<instance-id>-daemon`

but without the corresponding `<instance-id>.workspace.json`.

## Release Notes

```release-note
None
```

## Documentation

None

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/ws-daemon/pkg/internal/session/workspace.go
+++ b/components/ws-daemon/pkg/internal/session/workspace.go
@@ -204,7 +204,12 @@ func (s *Workspace) Dispose(ctx context.Context) (err error) {
 	// old workspace content we can garbage collect that content later.
 	err = os.Remove(s.persistentStateLocation())
 	if err != nil {
-		return xerrors.Errorf("cannot remove workspace persistent state location: %w", err)
+		if os.IsNotExist(err) {
+			log.WithError(err).Warn("workspace persistent state location not exist")
+			err = nil
+		} else {
+			return xerrors.Errorf("cannot remove workspace persistent state location: %w", err)
+		}
 	}
 
 	s.stateLock.Lock()


components/ws-daemon/pkg/internal/session/workspace.go
```diff
@@ -204,7 +204,12 @@ func (s *Workspace) Dispose(ctx context.Context) (err error) {
 	// old workspace content we can garbage collect that content later.
 	err = os.Remove(s.persistentStateLocation())
 	if err != nil {
-		return xerrors.Errorf("cannot remove workspace persistent state location: %w", err)
+		if os.IsNotExist(err) {
+			log.WithError(err).Warn("workspace persistent state location not exist")
+			err = nil
```
jenting, line 209: I'm not sure should we keep running the following code?
https://github.com/gitpod-io/gitpod/blob/5fb85e47e307e8b2acc1ea5ac990ccd07abaceeb/components/ws-daemon/pkg/internal/session/workspace.go#L210-L232
or directly return at here.

sagor999 (approved):
I think this is a good fix. LGTM.

PR #12028, opened 2022-08-10
ws-manager: Fix consistency of disposed ws between ws-daemon and ws-manager

utam0k:
## Description

Due to a retry timing issue, there were cases where the workspace had already been successfully backed up on the ws-daemon side, but ws-manager was trying to back it up again with an error.

[More detail](https://github.com/gitpod-io/gitpod/issues/11710#issuecomment-1210375041)

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/11710

## How to test

Is there any good way to do this?

## Release Notes

```release-note
Fixed a rare case in which prebuild was terminating successfully but failing
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/ws-manager/pkg/manager/monitor.go
+++ b/components/ws-manager/pkg/manager/monitor.go
@@ -1090,18 +1090,6 @@ func (m *Monitor) finalizeWorkspaceContent(ctx context.Context, wso *workspaceOb
 			return true, nil, status.Errorf(codes.Unavailable, "cannot connect to workspace daemon: %q", err)
 		}
 
-		// only build prebuild snapshots of initialized/ready workspaces.
-		if tpe == api.WorkspaceType_PREBUILD {
-			_, err = snc.WaitForInit(ctx, &wsdaemon.WaitForInitRequest{Id: workspaceID})
-			if st, ok := grpc_status.FromError(err); ok && st.Code() == codes.FailedPrecondition &&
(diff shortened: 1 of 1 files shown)


components/ws-manager/pkg/manager/monitor.go
```diff
@@ -1093,10 +1093,22 @@ func (m *Monitor) finalizeWorkspaceContent(ctx context.Context, wso *workspaceOb
...
+					// the workspace has gone some reason
+					// e.g. since it was a retry, it already succeeded the first time.
+					log.WithError(err).Warnf("skipping snapshot and disposing because the workspace has already gone")
+					val, ok := m.finalizerMap.LoadAndDelete(workspaceID)
+					if ok {
+						cancelReq := val.(context.CancelFunc)
+						cancelReq()
+					}
+					return false, &csapi.GitStatus{}, nil
+				}
```
jenting, line 1111: Can we move these lines https://github.com/gitpod-io/gitpod/blob/b8669f351eb66bac37b4a06a6753b4390232cba6/components/ws-manager/pkg/manager/monitor.go#L1115-L1126 upper?
Therefore, we can reuse the defer func.

utam0k: @jenting Thanks for your feedback. fixed

sagor999 (approved):

PR #12034, opened 2022-08-10
Increase threshold for unavailable targets

ArthurSens:
## Description

The goal is to avoid getting alerted due to small unavailabilities that don't represent a treat.

## Release Notes

```release-note
NONE
```


mads-hartmann (approved):
🛹

PR #12041, opened 2022-08-10
[ws-manager] Log pod name

Furisto:
## Description
Log pod name

## Related Issue(s)

Fixes #

## Release Notes

```release-note
None
```

## Werft options:

- [ ] /werft with-preview


jenting (commented):
Will the user see the error message with the pod name? If not, then LGTM.

kylos101:
:wave: @Furisto , 

Can you add how to test for this PR? For example, it seems like it may be related to OTS, as we're talking about creating secrets.

Would it make sense to log the error "as is" (without include the pod name in the message), and instead include the workspace pod name as metadata? This way there's no worry about potentially sending an error message to the user with a pod name that is meaningless.

PR #12043, opened 2022-08-10
[kots]: add useful links to the KOTS dashboard

MrSimonEmms:
## Description

Whilst scouring the KOTS documentation, I discovered we have the ability to add [custom links](https://docs.replicated.com/vendor/admin-console-adding-buttons-links#add-a-button-to-the-dashboard) to the KOTS dashboard. I've added a couple as these could be quite useful for users.

image

This is what occurred to me, but the content/ordering is entirely configurable. Probably best for @lucasvaltl to decide on what (if anything) goes here.

## How to test

Create a KOTS instance and this is on the homepage once installed

## Release Notes

```release-note
[kots]: add useful links to the KOTS dashboard
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/kots/manifests/kots-links.yaml
+++ b/install/kots/manifests/kots-links.yaml
@@ -1,0 +1,24 @@
+# Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+# Licensed under the MIT License. See License-MIT.txt in the project root for license information.
+
+apiVersion: app.k8s.io/v1beta1
+kind: Application
+metadata:
+  name: gitpod
+  labels:
+    app.kubernetes.io/name: gitpod
+    app.kubernetes.io/version: "1.0.0"
+spec:
+  selector:
+    matchLabels:
+     app.kubernetes.io/name: gitpod
+  descriptor:
+    links:
+      - description: Open {{repl ConfigOption "domain" }}
+        url: https://{{repl ConfigOption "domain" }}
+      - description: Gitpod Docs
+        url: https://www.gitpod.io/docs
+      - description: Gitpod Source Code
+        url: https://www.github.com/gitpod-io/gitpod
+      - description: Gitpod Website
+        url: https://www.gitpod.io


install/kots/manifests/kots-links.yaml
```diff
@@ -0,0 +1,24 @@
...
+     app.kubernetes.io/name: gitpod
+  descriptor:
+    links:
+      - description: Gitpod
+        url: https://www.gitpod.io
+      - description: Gitpod Docs
+        url: https://www.gitpod.io/docs
+      - description: Gitpod Source Code
+        url: https://www.github.com/gitpod-io/gitpod
+      - description: My Gitpod
```
lucasvaltl, line 23:
```suggestion
      - description: Your Gitpod Dashboard
```
Might be worth moving this to the very top as well.

install/kots/manifests/kots-links.yaml
```diff
@@ -0,0 +1,24 @@
...
+  labels:
+    app.kubernetes.io/name: gitpod
+    app.kubernetes.io/version: "1.0.0"
+spec:
+  selector:
+    matchLabels:
+     app.kubernetes.io/name: gitpod
+  descriptor:
+    links:
+      - description: Gitpod
```
lucasvaltl, line 17:
```suggestion
      - description: Gitpod Website
```

lucasvaltl (approved):
LGTM - only question I have - can we change what the word is that is at the top of the dropdown in the UI? RIght now it is `gitpod`, but maybe it should rather be `useful links` or something. Because just `gitpod` makes this not very discoverable, especially because the word Gitpod is already all over the UI. And all over this drop down hehe.

install/kots/manifests/kots-links.yaml
corneliusludmann: To be honest, I haven't known what “My Gitpod” resp. “Your Gitpod Dashboard” actually means until I looked at the URL. What about using the domain in the description like `Open https://{{repl ConfigOption "domain" }}` instead?

MrSimonEmms:
> LGTM - only question I have - can we change what the word is that is at the top of the dropdown in the UI

@lucasvaltl not that I can see. The support for this is limited, so I've had to go through the [CRD](https://github.com/kubernetes-sigs/application/blob/master/config/crd/bases/app.k8s.io_applications.yaml#L132,L146) and there's no additional configuration.

I've taken @corneliusludmann's suggestion to change the wording and put that at the top. That will (hopefully) draw the eye to it and then click it

image

PR #12080, opened 2022-08-11
[server] maintain workspace instance state

### LINKED ISSUE #12067: [server] Improve maintenance of workspace instance state
When there's an exception happening in `actuallyStartWorkspace` we catch it and treat the start request as failed by updating the workspace_instance in the DB and marking it as stopped. The request to ws-manager might have succeeded, and the error might have been thrown later (the concrete case was a protocol error with billingservice). 

We should try to keep the system consistent and try to cancel the workspace start with ws-manager in `failInstanceStart`.

Because the workspace instance was actually started, the stopped state will eventually be overwritten by `ws-manager-bridge`. It however only sets the status but doesn't reset the `stoppedTime`and `stoppingTime`fields. If they are set, we should log an error but also set those two properties to their default state again.

svenefftinge:
## Description
get closer to workspace instance state reflects actual state in cluster.

## Related Issue(s)

Fixes #12067

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/ws-manager-bridge/src/bridge.ts
+++ b/components/ws-manager-bridge/src/bridge.ts
@@ -346,6 +346,15 @@ export class WorkspaceManagerBridge implements Disposable {
                     }
 
                     instance.status.phase = "running";
+                    // let's check if the state is inconsistent and be loud if it is.
+                    if (instance.stoppedTime || instance.stoppingTime) {
+                        log.error("Resetting already stopped workspace to running.", {
+                            instanceId: instance.id,
+                            stoppedTime: instance.stoppedTime,
+                        });
+                        instance.stoppedTime = undefined;
+                        instance.stoppingTime = undefined;
+                    }
                     break;
                 case WorkspacePhase.INTERRUPTED:
                     instance.status.phase = "interrupted";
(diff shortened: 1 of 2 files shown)


components/ws-manager-bridge/src/bridge.ts
```diff
@@ -346,6 +346,15 @@ export class WorkspaceManagerBridge implements Disposable {
                     }
 
                     instance.status.phase = "running";
+                    // let's check if the state is inconsistent and be loud if it is.
+                    if (instance.stoppedTime || instance.stoppingTime) {
```
AlexTugarev, line 350: I didn't find any places where either `stoppedTime` or `stoppingTime` would be set without the status being set accordingly. 

OTOH there is a handling of wsman events where the RUNNING phase is being persisted blindly  https://github.com/gitpod-io/gitpod/blob/24f7b609bf79b247d72d1841282639726bcd5891/components/ws-manager-bridge/src/bridge.ts#L158

That said, I'm fine this adding this check here, but it seems, it should be done before the RUNNING phase is persisted bogusly.

cc. @geropl because he reimplemented the status mapping for prebuilds, which seems to be a better pattern for the whole update cascade of states.

AlexTugarev (approved):
Thanks! That change to workspace starter makes great sense 🥇 

I left a comment on the [change to the bridge](https://github.com/gitpod-io/gitpod/pull/12080/files#r944172951) – it seems to that it reverts a bogus status update in error cases, which in itself sounds ok. Actually we need to avoid having persisted the bogus state beforehand, not least because these status updates are used as base of prebuild status updates and client notifications via messagebus. That's definitely tracking as a separate issue, and perhaps solve it like it's done for the prebuild status updates.

PR #12082, opened 2022-08-12
redirect to desktop IDE on app domain

akosyakov:
## Description

An idea to redirect always on parent app domain, i.e. gitpod.io, via iframe to suppress popup on each new workspace start.

Screenshot 2022-08-12 at 09 09 44

**Important** We need to consider a case if a user does not have yet app, but allowed pop-up always for some reason, i.e. for another IDE.

## Related Issue(s)

Fixes #

## How to test

- Select VS Code Desktop in settings.
- Start one workspace, click allow always in pop-up.
- Start another workspace, check that pop-up does not come anymore. VS Code Desktop is opened automatically.

## Release Notes

```release-note
Enable allowing redirect to Desktop IDEs for all workspaces.
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/supervisor/frontend/src/ide/heart-beat.ts
+++ b/components/supervisor/frontend/src/ide/heart-beat.ts
@@ -19,7 +19,7 @@ export const track = (w: Window) => {
 }
 
 let pageCloseCompatibile: boolean = false
-
+// TODO(ak) remove
 isSaaSServerGreaterThan("main.4124").then((r) => {
     pageCloseCompatibile = r
 })
--- a/components/supervisor/frontend/src/index.ts
+++ b/components/supervisor/frontend/src/index.ts
@@ -158,16 +158,7 @@ const toStop = new DisposableCollection();
                         });
                         if (!desktopRedirected) {
                             desktopRedirected = true;
-                            try {
-                                const desktopLink = new URL(ideStatus.desktop.link)
-                                // redirect only if points to desktop application
-                                // don't navigate browser to another page
-                                if (desktopLink.protocol != 'http:' && desktopLink.protocol != 'https:') {
-                                    window.location.href = ideStatus.desktop.link;
-                                }
-                            } catch (e) {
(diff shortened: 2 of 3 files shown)


roboquat:

<details>

</details>

loujaybee:
Oh wow, tested, and works great. IMO this redirect works more as a user would expect, once they allow once for Gitpod, doesn't seem relevant to keep showing this for different workspaces. A couple thoughts: 

1. I cannot see an easy to get the prompt back (if I wanted to), without going to: `code ~/Library/Application\ Support/Google/Chrome/Default/Preferences`, and manually fiddling with the preferences, however that said we can easily document this anyway. 

2. On mac, this will "pull" the focus to the desktop where VS Code is already opened (if it is). This isn't really related to this issue though, and is something we can investigate separately.

akosyakov:
> Ensure that we are catching and handling any redirect errors where the external redirect fails.

yes, it is also my though if we won't suppress dialog then we should handle it and show own dialog. But from other side we already have a message in grayish which start with `if you **don't** see a dialog...`. So maybe not a blocker?

> In both cases we could provide direct download embeds to download the app:
 
I'm still uncomfortable with that. Because without toolbox users cannot upgrade Gateway and it will case an issue with each major release. We could insert direct links to toolbox downloadable instead? I think this can be done as another PR though.

loujaybee:
> I think this can be done as another PR though.

> So maybe not a blocker?

Ah yes, sorry my general sentiment was: "we have some proposals already lined up to catch/handle the failure states for the user" so we should be good to proceed with this PR and look at those UX niceties after for both JB + VS Code 🚀

akosyakov:
/hold

- [ ] land https://github.com/gitpod-io/gitpod/pull/12102
- [ ] update compatibility check with build version from main

akosyakov:
We need to think how we can pull the same trick for copy/paste feature.

felladrin (approved):

PR #12103, opened 2022-08-12
[supervisor] Make resource status request more resilient

### LINKED ISSUE #12083: extend supervisor rate-limiting on ResourceStatus
Currently, supervisor endpoints for retrieving [ResourceStatus](https://github.com/gitpod-io/gitpod/blob/main/components/supervisor-api/status.proto#L76-L80) calls directly an [upstream socket](https://github.com/gitpod-io/gitpod/blob/main/components/supervisor/pkg/supervisor/top.go#L26) to retrieve the resource status info, when a client requests them. There is requests rate-limiting upstream, so multiple clients request resource status to supervisor, and those requests are directly sent upstream, supervisor might fail to respond.

If we extend supervisor by fetching resource status asyncronously from incoming ResourceStatus requests, we can control the frequency of calls to the socket. When a new client requests data, we would just return the latest data received from upsteam.

Implementing a exponentioanl backoff retry strategy in each client. Currently, [gitpod-monitor](https://github.com/akosyakov/gitpod-monitor/blob/15bc5a39d8ab4f71a18ce7e0a38a70c8e9e0e526/src/extension.ts#L65-L77) already uses this approach but it would mean replicating this logicin in all clients (jetbrains backend-plugin, gitpod-cli, etc..).

andreafalzetti:
## Description
Makes the resource status request more resilient by avoiding calling syncronously upstream to fetch resource status at every requests. With more clients adding functionalities such as `gp top` or workspace cpu/memory in jetbrains ide control center, the risk of getting rate-limited is guaranteed. 

With the approached used in this PR, we retrieve data from upstream at every second and when clients request it, we return the latest available data. Additional, an exponential backoff strategy is in place to be more resilient.

## Related Issue(s)

Fixes #12083

## How to test
Compare how [this prev](https://afalz-1208328e4f3122.preview.gitpod-dev.com) env behaves in comparison with gitpod.io by running a workspace and run `watch -n 0.1 gp top`. In gitpod.io (prod) it will almost immediately fail by returning gRPC errors because ResourceStatus endpoint is pulled too frequently and we hit the rate-limit.

In the prev env, you should be able to run multiple `watch -n 0.1 gp top` without any errors.

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/supervisor/pkg/supervisor/services.go
+++ b/components/supervisor/pkg/supervisor/services.go
@@ -97,6 +97,7 @@ type statusService struct {
 	Tasks           *tasksManager
 	ideReady        *ideReadyState
 	desktopIdeReady *ideReadyState
+	topService      *TopService
 
 	api.UnimplementedStatusServiceServer
 }
@@ -933,5 +934,5 @@ func (s *portService) RetryAutoExpose(ctx context.Context, req *api.RetryAutoExp
 
 // ResourcesStatus provides workspace resources status information.
 func (s *statusService) ResourcesStatus(ctx context.Context, in *api.ResourcesStatuRequest) (*api.ResourcesStatusResponse, error) {
-	return Top(ctx)
+	return s.topService.data, nil
 }
--- a/components/supervisor/pkg/supervisor/supervisor.go
+++ b/components/supervisor/pkg/supervisor/supervisor.go
@@ -257,6 +257,9 @@ func Run(options ...RunOption) {
 		go analyseConfigChanges(ctx, cfg, analytics, gitpodConfigService, gitpodService)
 	}
 
+	topService := NewTopService()
+	topService.Observe(ctx)
+
 	termMux := terminal.NewMux()
 	termMuxSrv := terminal.NewMuxTerminalService(termMux)
 	termMuxSrv.DefaultWorkdir = cfg.RepoRoot
@@ -287,6 +290,7 @@ func Run(options ...RunOption) {
 			Tasks:           taskManager,
 			ideReady:        ideReady,
 			desktopIdeReady: desktopIdeReady,
+			topService:      topService,
 		},
 		termMuxSrv,
 		RegistrableTokenService{Service: tokenService},
--- a/components/supervisor/pkg/supervisor/top.go
+++ b/components/supervisor/pkg/supervisor/top.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"golang.org/x/xerrors"
@@ -17,9 +18,62 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 
 	linuxproc "github.com/c9s/goprocinfo/linux"
+	"github.com/gitpod-io/gitpod/common-go/log"
 	"github.com/gitpod-io/gitpod/supervisor/api"
 	daemonapi "github.com/gitpod-io/gitpod/ws-daemon/api"
 )
+
+type TopService struct {
+	data      *api.ResourcesStatusResponse
+	ready     chan struct{}
+	readyOnce sync.Once
+	top       func(ctx context.Context) (*api.ResourcesStatusResponse, error)
+}
+
+func NewTopService() *TopService {
+	log.Debug("gitpod top service: initialized")
+	return &TopService{
+		top: Top,
+	}
+}
+
+// Observe starts observing the resource status
+func (t *TopService) Observe(ctx context.Context) {
+	var (
+		delay                       = 1 * time.Second
+		reconnectionDelayGrowFactor = 1.5
+		minReconnectionDelay        = 1 * time.Second
+		maxReconnectionDelay        = 30 * time.Second
+	)
+
+	t.ready = make(chan struct{})
+
+	go func() {
+		for {
+			data, err := t.top(ctx)
+			if err != nil {
+				log.WithField("error", err).Errorf("failed to retrieve resource status from upstream, trying again in %d seconds...", uint32(delay.Seconds()))
+			} else {
+				delay = minReconnectionDelay
+				t.data = data
+
+				t.readyOnce.Do(func() {
+					close(t.ready)
+				})
+			}
+			select {
+			case <-ctx.Done():
+				log.Info("Resource Status observer stopped")
+				return
+			case <-time.After(delay):
+				delay = time.Duration(float64(delay) * reconnectionDelayGrowFactor)
+				if delay > maxReconnectionDelay {
+					delay = maxReconnectionDelay
+				}
+			}
+		}
+	}()
+}
 
 // Top provides workspace resources status information.
 func Top(ctx context.Context) (*api.ResourcesStatusResponse, error) {
--- a/components/supervisor/pkg/supervisor/top_test.go
+++ b/components/supervisor/pkg/supervisor/top_test.go
@@ -1,0 +1,94 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package supervisor
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/gitpod-io/gitpod/supervisor/api"
+	"golang.org/x/xerrors"
+)
+
+func TestTopServiceHappyPath(t *testing.T) {
+	ctx := context.Background()
+
+	topService := NewTopService()
+	topService.Observe(ctx)
+
+	<-topService.ready
+
+	if topService.data == nil {
+		t.Errorf("topService data should not be nil")
+	}
+	if topService.data.Memory.Used == 0 {
+		t.Errorf("Used Memory should not be zero")
+	}
+	if topService.data.Memory.Limit == 0 {
+		t.Errorf("Total Memory should not be zero")
+	}
+	if topService.data.Cpu.Used == 0 {
+		t.Errorf("Used CPU should not be zero")
+	}
+	if topService.data.Cpu.Limit == 0 {
+		t.Errorf("Total CPU should not be zero")
+	}
+}
+
+func TestTopServiceWithUpstreamFailure(t *testing.T) {
+	ctx := context.Background()
+
+	var isFirstRun = true
+
+	topService := NewTopService()
+	topService.top = func(ctx context.Context) (*api.ResourcesStatusResponse, error) {
+		if isFirstRun {
+			isFirstRun = false
+			return &api.ResourcesStatusResponse{
+				Memory: &api.ResourceStatus{
+					Used:  1,
+					Limit: 10,
+				},
+				Cpu: &api.ResourceStatus{
+					Used:  2,
+					Limit: 5,
+				},
+			}, nil
+		} else {
+			return nil, xerrors.Errorf("some error occurred with the sock file")
+		}
+	}
+	topService.Observe(ctx)
+	<-topService.ready
+
+	if topService.data.Memory.Used != 1 {
+		t.Errorf("Used Memory should be 1")
+	}
+	if topService.data.Memory.Limit != 10 {
+		t.Errorf("Total Memory should be 10")
+	}
+	if topService.data.Cpu.Used != 2 {
+		t.Errorf("Used Cpu should be 2")
+	}
+	if topService.data.Cpu.Limit != 5 {
+		t.Errorf("Total Cpu should be 5")
+	}
+
+	time.Sleep(2 * time.Second)
+
+	if topService.data.Memory.Used != 1 {
+		t.Errorf("Used Memory should be 1")
+	}
+	if topService.data.Memory.Limit != 10 {
+		t.Errorf("Total Memory should be 10")
+	}
+	if topService.data.Cpu.Used != 2 {
+		t.Errorf("Used Cpu should be 2")
+	}
+	if topService.data.Cpu.Limit != 5 {
+		t.Errorf("Total Cpu should be 5")
+	}
+}


roboquat:

<details>

</details>

components/supervisor/pkg/supervisor/top.go
```diff
@@ -17,10 +17,57 @@ import (
...
+
+// Observe starts observing the resource status
+func (t *TopService) Observe(ctx context.Context) {
+	var (
+		delay                       = 1 * time.Second
+		reconnectionDelayGrowFactor float64
+		maxReconnectionDelay        = 30 * time.Second
+	)
+
+	t.updateTopData(ctx)
```
andreafalzetti, line 50: I also call this immediately otherwise `Observe` waits 1 second before fetching the metrics

components/supervisor/pkg/supervisor/services.go
```diff
@@ -933,5 +934,6 @@ func (s *portService) RetryAutoExpose(ctx context.Context, req *api.RetryAutoExp
 
 // ResourcesStatus provides workspace resources status information.
 func (s *statusService) ResourcesStatus(ctx context.Context, in *api.ResourcesStatuRequest) (*api.ResourcesStatusResponse, error) {
-	return Top(ctx)
+	var data *api.ResourcesStatusResponse = s.topService.data
+	return data, nil
```
akosyakov, line 938: (nit) `return s.topService.data, nil` - top service should ensure consistency of data, clients don't need to worry

components/supervisor/pkg/supervisor/top.go
akosyakov: you can do it like that to avoid code duplications:
```go
for {
      data, err := Top(ctx)
      if err != nil {
            log.WithField("error", err).Error("Failed to retrieve resource status from upstream")
      } else {
           delay = minReconnectionDelay
	   t.data = data
      }
   select {
	case <-ctx.Done():
		log.Info("Resource Status observer stopped")
		return
	case <-time.After(delay):
		delay = time.Duration(float64(delay) * reconnectionDelayGrowFactor)
		if delay > maxReconnectionDelay {
			delay = maxReconnectionDelay
		}
	}
}

File: components/supervisor/pkg/supervisor/top.go [hunk 3]
```diff
@@ -17,10 +17,57 @@ import (
...
+	if err != nil {
+		log.WithField("error", err).Error("Failed to retrieve resource status from upstream")
+	}
+	t.data = data
+}
+
+// Observe starts observing the resource status
+func (t *TopService) Observe(ctx context.Context) {
+	var (
+		delay                       = 1 * time.Second
```
akosyakov, line 45: you need `minReconnectionDelay` and reset on it whenever retrieving data is successful

components/supervisor/pkg/supervisor/top.go
```diff
@@ -17,10 +17,57 @@ import (
...
+	var newData *api.ResourcesStatusResponse = t.data
+	return newData
+}
+
+func (t *TopService) updateTopData(ctx context.Context) {
+	data, err := Top(ctx)
+	if err != nil {
+		log.WithField("error", err).Error("Failed to retrieve resource status from upstream")
+	}
+	t.data = data
```
akosyakov, line 39: it should be in else block, otherwise it will product nil if failed because of rate limitting

akosyakov:
@andreafalzetti How do you test failure case?

components/supervisor/pkg/supervisor/top_test.go
```diff
@@ -0,0 +1,33 @@
...
+	"testing"
+)
+
+func TestTopService(t *testing.T) {
+	ctx := context.Background()
+
+	topService := &TopService{}
+	topService.Observe(ctx)
+
+	if topService.data == nil {
```
akosyakov, line 18: maybe add a ready channel to synchronise on first successful request

andreafalzetti: @akosyakov do you mean something the client will have to check to see if data is ready? 🤔

akosyakov: just for tests use internal channel in the same package
```go
<-topService.ready // returns on first successful request
if topService.data == nil {
...
```

components/supervisor/pkg/supervisor/top.go
andreafalzetti: Updated 🙏

components/supervisor/pkg/supervisor/top.go
andreafalzetti: Updated 🙏

components/supervisor/pkg/supervisor/top_test.go
andreafalzetti: Updated 🙏

components/supervisor/pkg/supervisor/services.go
andreafalzetti: I've updated the `return` (removed the unnecessary pointer), but I don't understand if you are suggesting any more change by saying: " top service should ensure consistency of data, clients don't need to worry" - I agree with it, but is there any actionable on it?

Unless you're regerring to the change to the `return` but I think that is purely aesthetic change 🤔

akosyakov: ok, i though we initially introduced it to create a new pointer and avoid a situation when top service updates it later, and it was left over

andreafalzetti:
> How do you test failure case?

That's what is left missing. If you have any suggestions or can point me to some similar tests, I would appreciate. In the meantime I will attempt some ideas.

akosyakov:
Run several watch gp top in parallel and then manipulate with sock file.

akosyakov:
@andreafalzetti I actually cannot start a workspace if gets stopped immediately.

andreafalzetti:
> @andreafalzetti I actually cannot start a workspace if gets stopped immediately.

@akosyakov Weird. I was able to reproduce. Before today's changes it was working fine. Maybe it's the prev env, I will try to do a clean deployment

akosyakov:
You can check workspace logs, maybe supervisor panics somehow.

andreafalzetti:
@akosyakov If I look for the file pathes used to read memory/cpu, in my workspace, I don't see them. What am I missing?

e.g. `/sys/fs/cgroup/memory/memory.limit_in_bytes` -> https://github.com/gitpod-io/gitpod/blob/main/components/supervisor/pkg/supervisor/top.go#L67

akosyakov:
> If I look for the file pathes used to read memory/cpu, in my workspace, I don't see them. What am I missing?

for cgroup v2 there is not such files only via `/.supervisor/info.sock`

andreafalzetti:
@akosyakov workspace starts now, it was re-closing a closed channel

akosyakov:
> @akosyakov workspace starts now, it was re-closing a closed channel

Which one? rate limitting you can do many `watch gp top`.

components/supervisor/pkg/supervisor/top_test.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	if topService.data.Cpu.Used == 0 {
+		t.Errorf("Used CPU should not be zero")
+	}
+	if topService.data.Cpu.Limit == 0 {
+		t.Errorf("Total CPU should not be zero")
+	}
+}
+
+// func TestTopServiceFailure(t *testing.T) {
```
akosyakov, line 40: left over?

akosyakov (approved):
works as advertised

/hold
if you want to clean up something

components/supervisor/pkg/supervisor/top_test.go
andreafalzetti: @akosyakov I am still trying to figure out how to test for failure, either mocking gRPC or injecting the `Top` func dependency.. I would like to add one more test before merging

components/supervisor/pkg/supervisor/top.go
```diff
@@ -10,17 +10,68 @@ import (
...
+func (t *TopService) GetTopData() *api.ResourcesStatusResponse {
+	var newData *api.ResourcesStatusResponse = t.data
+	return newData
+}
+
+// Observe starts observing the resource status
+func (t *TopService) Observe(ctx context.Context) {
+	var (
+		delay                       = 1 * time.Second
+		reconnectionDelayGrowFactor float64
```
jeanp413, line 41: @andreafalzetti Maybe a dump question buy where is this initialized before being used in line 66?

components/supervisor/pkg/supervisor/top_test.go
akosyakov: I think you will need to mock `Top` itself via interface: https://gobyexample.com/interfaces I'm not sure whether it is worth a trouble, I tested by removing sock file and gp top was given the last number. In practice I don't expect it to stuck but resolve after sometime to new number.

components/supervisor/pkg/supervisor/top.go
akosyakov: good point it should be assigned to 1.5

components/supervisor/pkg/supervisor/top_test.go
akosyakov: Actually you don't need interface,  `TopService` can have a property `top` which is function, by default you assign `Top` to it. In tests you can provide you own implementation.

components/supervisor/pkg/supervisor/top.go
```diff
@@ -10,17 +10,68 @@ import (
...
+		maxReconnectionDelay        = 30 * time.Second
+	)
+
+	t.ready = make(chan struct{})
+
+	go func() {
+		for {
+			data, err := Top(ctx)
+			if err != nil {
+				log.WithField("error", err).Error("Failed to retrieve resource status from upstream")
```
akosyakov, line 52: usually we will add something like `trying again in $d...` see [[1](https://github.com/gitpod-io/gitpod/blob/12017a58e8d4e173016af5166a722b4603c41b81/components/gitpod-protocol/go/reconnecting-ws.go#L227)]

components/supervisor/pkg/supervisor/top_test.go
andreafalzetti: I've added another test injecting the Top function and this allowed me to simulate failure, the metrics stay the same which demonstrates that works!

I will run some double checks on prev env just to make sure everything works fine there too.

components/supervisor/pkg/supervisor/top.go
andreafalzetti: good catch, updated, thanks 🤦 🙏

components/supervisor/pkg/supervisor/top.go
```diff
@@ -10,17 +10,73 @@ import (
...
+	readyOnce sync.Once
+	Top       func(ctx context.Context) (*api.ResourcesStatusResponse, error)
+}
+
+func NewTopService(
+	top func(ctx context.Context) (*api.ResourcesStatusResponse, error),
+) *TopService {
+	log.Debug("gitpod top service: initialized")
+	return &TopService{
+		Top: top,
```
akosyakov, line 38: (nit) fyi you could do here like `top: Top` and in test do `topSerivce.top = func() {}`

components/supervisor/pkg/supervisor/top.go
```diff
@@ -10,17 +10,73 @@ import (
...
+	"github.com/gitpod-io/gitpod/common-go/log"
 	"github.com/gitpod-io/gitpod/supervisor/api"
 	daemonapi "github.com/gitpod-io/gitpod/ws-daemon/api"
 )
 
+type TopService struct {
+	data      *api.ResourcesStatusResponse
+	ready     chan struct{}
+	readyOnce sync.Once
+	Top       func(ctx context.Context) (*api.ResourcesStatusResponse, error)
```
akosyakov, line 30: (nit) upper case means exporting it outside of package, i think in this case it is just internal variable and should be `top`

components/supervisor/pkg/supervisor/supervisor.go
```diff
@@ -257,6 +257,9 @@ func Run(options ...RunOption) {
 		go analyseConfigChanges(ctx, cfg, analytics, gitpodConfigService, gitpodService)
 	}
 
+	topService := NewTopService(Top)
```
akosyakov, line 260: (nit) I think for usual clients it should be just `NewTopService`

components/supervisor/pkg/supervisor/top.go
```diff
@@ -10,17 +10,73 @@ import (
...
+	log.Debug("gitpod top service: initialized")
+	return &TopService{
+		Top: top,
+	}
+}
+
+// Observe starts observing the resource status
+func (t *TopService) Observe(ctx context.Context) {
+	var (
+		delay                       = 1 * time.Second
```
akosyakov, line 45: (nit) `delay = minReconnectionDelay` if you declare it below after constants

components/supervisor/pkg/supervisor/top_test.go
```diff
@@ -0,0 +1,93 @@
...
+		t.Errorf("Total Memory should be 10")
+	}
+	if topService.data.Cpu.Used != 2 {
+		t.Errorf("Used Cpu should be 2")
+	}
+	if topService.data.Memory.Limit != 5 {
+		t.Errorf("Total Cpu should be 5")
+	}
+
+	time.Sleep(2 * time.Second)
```
akosyakov, line 79: you can use some channel here like to avoid relying on timing:
`<-failed`

akosyakov:
I left some nit, can be merged without addressing them as well. Not sure about the test, using timeouts can lead to flakiness.

components/supervisor/pkg/supervisor/supervisor.go
andreafalzetti: @akosyakov do you mean without parameters?

components/supervisor/pkg/supervisor/top.go
andreafalzetti: ok, it would be nicer if in `supervisor.go` I don't have to pass `Top` as func in the constructor but in this case the constructor I guess I don't need it anymore and only do what you suggest in the tests `topSerivce.top = func() {}`

akosyakov: I think you can keep it, since you encapsulated how top is implemented?

andreafalzetti: I must have misinterpreted your suggestion, if I do `top: Top` it will pick up the globally exported `Top` func. 🤔

akosyakov: I meant:
```go
func NewTopService() *TopService {
	log.Debug("gitpod top service: initialized")
	return &TopService{
		top: Top,
	}
}
```

components/supervisor/pkg/supervisor/top_test.go
andreafalzetti: We agreed to keep it as is for now

PR #12105, opened 2022-08-12
[server] fix using snapshotBucketId when it is undefined or empty

sagor999:
## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/11823

## Release Notes

```release-note
none
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/server/src/workspace/workspace-starter.ts
+++ b/components/server/src/workspace/workspace-starter.ts
@@ -1424,7 +1424,7 @@ export class WorkspaceStarter {
         }
 
         const volumeSnapshotId =
-            SnapshotContext.is(workspace.context) || WithPrebuild.is(workspace.context)
+            ((SnapshotContext.is(workspace.context) || WithPrebuild.is(workspace.context)) && workspace.context.snapshotBucketId !== undefined && workspace.context.snapshotBucketId != "")
                 ? workspace.context.snapshotBucketId
                 : lastValidWorkspaceInstanceId;
 
--- a/components/server/src/workspace/workspace-starter.ts
+++ b/components/server/src/workspace/workspace-starter.ts
@@ -1424,7 +1424,7 @@ export class WorkspaceStarter {
         }
 
         const volumeSnapshotId =
-            ((SnapshotContext.is(workspace.context) || WithPrebuild.is(workspace.context)) && workspace.context.snapshotBucketId !== undefined && workspace.context.snapshotBucketId != "")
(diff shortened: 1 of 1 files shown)


sagor999:
@kylos101 FYI :eyes:

components/server/src/workspace/workspace-starter.ts
```diff
@@ -1424,7 +1424,7 @@ export class WorkspaceStarter {
         }
 
         const volumeSnapshotId =
-            SnapshotContext.is(workspace.context) || WithPrebuild.is(workspace.context)
+            ((SnapshotContext.is(workspace.context) || WithPrebuild.is(workspace.context)) && workspace.context.snapshotBucketId !== undefined && workspace.context.snapshotBucketId != "")
```
kylos101, line 1427: Hey @sagor999 , here's a brief suggestion to also cover null:

```suggestion
((SnapshotContext.is(workspace.context) || WithPrebuild.is(workspace.context)) && !!workspace.context.snapshotBucketId)
```

[Context](https://www.typescriptlang.org/play?#code/MYewdgziA2CmB0sBOSRIBQEJNgK7WgEoBuAKFEhgWVQ21zABNYAzASzFkZPPCjkQo0WTACJRPCv2pC6YqAFtYAAgAuACw4BzCWSA)

kylos101: [Moar context](https://stackoverflow.com/a/7452731)

sagor999: Thank you!! :heart:

sagor999:
@geropl could you confirm is passing undefined into
`const volumeSnapshots = await this.workspaceDb.trace(traceCtx).findVolumeSnapshotById(volumeSnapshotId);`
which calls:
https://github.com/gitpod-io/gitpod/blob/14b392003764fa1883285a6c1a114998c9fd243f/components/gitpod-db/src/typeorm/workspace-db-impl.ts#L741-L744
`findOne(volumeSnapshotId)`
I suspect it does not handle undefined variable properly here. But maybe I am wrong.
Does it? I have a feeling it returns random record if undefined is passed into it.

geropl:
> I suspect it does not handle undefined variable properly here. But maybe I am wrong.
> Does it? I have a feeling it returns random record if undefined is passed into it.

You're right in that it is not handled properly. `undefined` should map to an "empty" value, which that is depends on the column definition/mapping code. So it could be that it pulls any record that is empty (haven't checked nor tested).

geropl (approved):
Code LGTM! :+1:

PR #12116, opened 2022-08-13
[gitpod-cli] New cmd `gp tasks stop` 🧪🔬

andreafalzetti:
## Description
I would like to introduce this new cmd `gp tasks stop` to stop 1 or all workspace tasks. It can be useful when someone wants to run some new tasks and don't care about all pre-defined ones or quickly kill some long-living process running in another task, for example.

I am not sure if the approach is too naive, but it seems to work.

## Related Issue(s)

There is no real issue as this is more of a "weekend idea". Perhaps it could be considered slightly related to [inner loop](https://github.com/gitpod-io/gitpod/issues/7671).

## How to test
1. Go to [this prev env](https://afalz-gp-tasks-stop.preview.gitpod-dev.com)
2. Open the gitpod's repo
3. Open a new terminal and run `gp tasks stop`, `gp tasks stop <id>` and `gp tasks stop --all`
4. Observe how all tasks are terminated

## Release Notes

```release-note
Gitpod CLI: New command `gp tasks stop`
```

## Documentation

Will create issue/PR if the idea moves forward

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-cli/cmd/tasks-stop.go
+++ b/components/gitpod-cli/cmd/tasks-stop.go
@@ -1,0 +1,136 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"time"
+
+	supervisor_helper "github.com/gitpod-io/gitpod/gitpod-cli/pkg/supervisor-helper"
+	"github.com/gitpod-io/gitpod/supervisor/api"
+	supervisor "github.com/gitpod-io/gitpod/supervisor/api"
+	"github.com/manifoldco/promptui"
+	"github.com/spf13/cobra"
+)
+
+var stopTaskCmdOpts struct {
+	All bool
+}
+
+// stopTaskCmd represents the stop task command
+var stopTaskCmd = &cobra.Command{
+	Use:   "stop <id>",
+	Short: "Stop a workspace task",
+	Args:  cobra.MaximumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		var (
+			terminalAliases []string
+			ctx             context.Context
+			cancel          context.CancelFunc
+		)
+
+		terminalClient, err := supervisor_helper.GetTerminalServiceClient(context.Background())
+
+		if err != nil {
+			log.Fatalf("cannot get terminal service: %s", err)
+			return
+		}
+
+		ctx, cancel = context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+
+		all, _ := cmd.Flags().GetBool("all")
+
+		if all {
+			ctx, cancel = context.WithTimeout(context.Background(), 5*time.Second)
+			defer cancel()
+
+			tasks, err := supervisor_helper.GetTasksListByState(ctx, supervisor.TaskState_running)
+
+			if err != nil {
+				log.Fatalf("cannot get task list: %s", err)
+				return
+			}
+
+			if len(tasks) == 0 {
+				fmt.Println("There are no running tasks")
+				return
+			}
+
+			for _, task := range tasks {
+				terminalAliases = append(terminalAliases, task.Terminal)
+			}
+		} else if len(args) > 0 {
+			_, err := terminalClient.Get(ctx, &api.GetTerminalRequest{
+				Alias: args[0],
+			})
+
+			if err != nil {
+				fmt.Printf("The selected task was not found or already stopped: %s.\nMake sure to use the correct task ID.\nUse 'gp tasks list' to obtain the task id or run 'gp tasks stop' to select the desired task\n", args[0])
+				return
+			}
+
+			terminalAliases = append(terminalAliases, args[0])
+		} else {
+			tasks, err := supervisor_helper.GetTasksListByState(ctx, supervisor.TaskState_running)
+			if err != nil {
+				log.Fatalf("cannot get task list: %s", err)
+			}
+
+			if len(tasks) == 0 {
+				fmt.Println("There are no running tasks")
+				return
+			}
+
+			var taskNames []string
+			var taskIndex int
+
+			if len(tasks) == 1 {
+				taskIndex = 0
+			} else {
+
+				for _, task := range tasks {
+					taskNames = append(taskNames, task.Presentation.Name)
+				}
+
+				prompt := promptui.Select{
+					Label:        "What task do you want to stop?",
+					Items:        taskNames,
+					HideSelected: true,
+				}
+
+				selectedIndex, selectedValue, err := prompt.Run()
+
+				if selectedValue == "" {
+					return
+				}
+
+				if err != nil {
+					log.Fatalf("error occurred with the input prompt: %s", err)
+				}
+
+				taskIndex = selectedIndex
+			}
+
+			terminalAliases = append(terminalAliases, tasks[taskIndex].Terminal)
+		}
+
+		for _, terminalAlias := range terminalAliases {
+			_, err := terminalClient.Shutdown(context.Background(), &supervisor.ShutdownTerminalRequest{Alias: terminalAlias})
+			if err != nil {
+				log.Fatalf("cannot stop task: %s", err)
+				return
+			}
+		}
+	},
+}
+
+func init() {
+	tasksCmd.AddCommand(stopTaskCmd)
+
+	stopTaskCmd.Flags().BoolVarP(&stopTaskCmdOpts.All, "all", "a", false, "stop all tasks")
+}
--- a/components/gitpod-cli/go.mod
+++ b/components/gitpod-cli/go.mod
@@ -12,8 +12,7 @@ require (
 	github.com/google/shlex v0.0.0-20181106134648-c34317bd91bf
 	github.com/google/tcpproxy v0.0.0-20180808230851-dfa16c61dad2
 	github.com/gorilla/handlers v1.5.1
-	github.com/manifoldco/promptui v0.3.2
-	github.com/nicksnyder/go-i18n v1.10.1 // indirect
+	github.com/manifoldco/promptui v0.9.0
 	github.com/olekukonko/tablewriter v0.0.5
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/procfs v0.7.3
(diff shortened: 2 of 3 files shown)


roboquat:

<details>

</details>

components/gitpod-cli/cmd/tasks-stop.go
```diff
@@ -0,0 +1,138 @@
...
+	All bool
+}
+
+func stopTask(ctx context.Context, terminalClient supervisor.TerminalServiceClient, terminalAlias string) {
+	terminal, err := terminalClient.Get(context.Background(), &supervisor.GetTerminalRequest{Alias: terminalAlias})
+	if err != nil {
+		panic(err)
+	}
+
+	stopCmd := fmt.Sprintf("%s kill %d%s", "\003\n", terminal.Pid, "\n")
```
andreafalzetti, line 30:
```suggestion
	stopCmd := fmt.Sprintf("%s kill %d%s", "\003\n", terminal.Pid, "\n")
```

What this does:

1. `CTRL + C ` to stop potential foreground processes (eg. long running processes, watch, compilers, servers, etc)
2. Kill the task process by it's ID

`\003` -> `CTRL + C `
`\n` -> `Enter / Return`

components/gitpod-cli/cmd/tasks-stop.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+var stopTaskCmdOpts struct {
+	All bool
+}
+
+func stopTask(ctx context.Context, terminalClient supervisor.TerminalServiceClient, terminalAlias string) {
+	terminal, err := terminalClient.Get(context.Background(), &supervisor.GetTerminalRequest{Alias: terminalAlias})
+	if err != nil {
+		panic(err)
```
akosyakov, line 27: I'm not sure about usage of panic, usually we are using logger with fatal exit. Although in cli it should not be so bad, could you check and align with how we do in other commands please

components/gitpod-cli/cmd/tasks-stop.go
akosyakov: I wonder why do we need to connect to it and cannot send a system signal instead using something like `syscall.Kill`?

components/gitpod-cli/cmd/tasks-stop.go
andreafalzetti: Some commands use `panic` others use `fail`

```go
func fail(msg string) {
	fmt.Fprintln(os.Stderr, msg)
	os.Exit(-1)
}
```

I think they are similar and both OK for CLI, but we should consider using a single approach. I will create an issue for it.

components/gitpod-cli/cmd/tasks-stop.go
andreafalzetti: @akosyakov I have replaced the approach from:

```go
stopCmd := fmt.Sprintf("%s kill %d%s", "\003\n", terminal.Pid, "\n")
_, err = terminalClient.Write(ctx, &api.WriteTerminalRequest{Alias: terminal.Alias, Stdin: []byte(stopCmd)})}
```

to:

```go
syscall.Kill(int(terminal.Pid), syscall.SIGINT)
syscall.Kill(int(terminal.Pid), syscall.SIGTERM)
```

I think code-wise looks better and more stable and and works similarly 👍

components/gitpod-cli/cmd/tasks-stop.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	syscall.Kill(int(terminal.Pid), syscall.SIGINT)
+	syscall.Kill(int(terminal.Pid), syscall.SIGTERM)
+	syscall.Kill(int(terminal.Pid), syscall.SIGKILL)
```
akosyakov, line 32: we should some time processes to process SIGTERM to finish gracefully, I'm actually not sure we need SIGINT at all

```go
syscall.Kill(int(terminal.Pid), syscall.SIGTERM)
<-time.After(15 * time.Second)
syscall.Kill(int(terminal.Pid), syscall.SIGKILL)
```

components/gitpod-cli/cmd/tasks-stop.go
akosyakov: it seem even in this file we use `log.Fatalf` somewhere. Let's use `log` everywhere?

filiptronicek (commented):
Hello hello, @andreafalzetti!

This feature is honestly great. I love it and it works swimmingly. Here's some things I took note of:
- I accidentally did `gp stop a01cefb2-edc6-4e3a-9d5c-9f6ae97b01a1` instead of `gp tasks stop a01cefb2-edc6-4e3a-9d5c-9f6ae97b01a1`, which I am embarrassed about 😆. Probably nothing actionable but I think it may happen to others as well. 
- When one provides an incorrect ID for a task (maybe someone who is not really familiar with task IDs or they copy-paste incorrectly), it looks like we don't handle that properly:
   ```
	$ gp tasks stop andrea
    panic: rpc error: code = NotFound desc = terminal not found

	  goroutine 1 [running]:
	  github.com/gitpod-io/gitpod/gitpod-cli/cmd.stopTask({0xb96a70?, 0xc00003a080?}, {0xb99c78, 0xc00026d080}, {0x7ffd7ed43406, 0x6})
	          /tmp/build/components-gitpod-cli--app.6087b2bc176a15865c74375746932db6c25af9c2/cmd/tasks-stop.go:27 +0x105
	  github.com/gitpod-io/gitpod/gitpod-cli/cmd.glob..func20(0xf731a0?, {0xc00026cc80, 0x1, 0x1?})
	          /tmp/build/components-gitpod-cli--app.6087b2bc176a15865c74375746932db6c25af9c2/cmd/tasks-stop.go:125 +0x567
	  github.com/spf13/cobra.(*Command).execute(0xf731a0, {0xc00026cc50, 0x1, 0x1})
	          /home/gitpod/go-packages/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:856 +0x663
	  github.com/spf13/cobra.(*Command).ExecuteC(0xf718a0)
	          /home/gitpod/go-packages/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:960 +0x39d
	  github.com/spf13/cobra.(*Command).Execute(...)
	          /home/gitpod/go-packages/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:897
	  github.com/gitpod-io/gitpod/gitpod-cli/cmd.Execute()
	          /tmp/build/components-gitpod-cli--app.6087b2bc176a15865c74375746932db6c25af9c2/cmd/root.go:35 +0x325
	  main.main()
	          /tmp/build/components-gitpod-cli--app.6087b2bc176a15865c74375746932db6c25af9c2/main.go:12 +0x17
	```

akosyakov:
@andreafalzetti sorry for putting you on wrong track, we should actually use supervisor API for that [[1](https://github.com/gitpod-io/gitpod/blob/30a5834443755da4271c5a2dd4be0e08be8b9c16/components/supervisor-api/terminal.proto#L18-L20)] as our other clients do [[2](https://github.com/gitpod-io/openvscode-server/blob/f8dd8cbaa13b9c50ebdc3be3fac2c4e3f8827e9e/extensions/gitpod-shared/src/features.ts#L967-L983)]

components/gitpod-cli/cmd/tasks-stop.go
akosyakov: we should do https://github.com/gitpod-io/gitpod/pull/12116#issuecomment-1214832759 instead

felladrin:
Is it ready for review, or will there be more changes related to @akosyakov's last comment?

andreafalzetti:
> Is it ready for review, or will there be more changes related to @akosyakov's last comment?

Moved to draft, sorry 🙏 I will refactor as soon as I have some spare time, considering that it's low-priority (not scheduled as part of our weekly work)

components/gitpod-cli/go.mod
```diff
@@ -12,7 +12,7 @@ require (
 	github.com/google/shlex v0.0.0-20181106134648-c34317bd91bf
 	github.com/google/tcpproxy v0.0.0-20180808230851-dfa16c61dad2
 	github.com/gorilla/handlers v1.5.1
-	github.com/manifoldco/promptui v0.3.2
+	github.com/manifoldco/promptui v0.9.0
```
andreafalzetti, line 15: The newer version allows hiding the message after selection using `HideSelected`

components/gitpod-cli/cmd/tasks-stop.go
```diff
@@ -0,0 +1,136 @@
...
+			for _, task := range tasks {
+				terminalAliases = append(terminalAliases, task.Terminal)
+			}
+		} else if len(args) > 0 {
+			_, err := terminalClient.Get(ctx, &api.GetTerminalRequest{
+				Alias: args[0],
+			})
+
+			if err != nil {
+				fmt.Printf("The selected task was not found: %s.\nMake sure to use the correct task ID.\nUse 'gp tasks list' to obtain the task id or run 'gp tasks stop' to select the desired task\n", args[0])
```
andreafalzetti, line 73: Is it too verbose, what do you guys think?

I like the idea from @filiptronicek to provide more info because the task ID could be confusing for some users.

Screenshot 2022-08-21 at 19 43 21

components/gitpod-cli/go.mod
andreafalzetti: I also ran `go mod tidy` that's why these changes

components/gitpod-cli/cmd/tasks-stop.go
filiptronicek: I like it this way, I think it makes for a message that users without an idea about task IDs can understand.

andreafalzetti:
Ready for review again

mustard-mh:
When there is no tasks defined in `.gitpod.yml`, exec `gp tasks stop` will terminate the initial terminal. 

That is because supervisor will create one default task here https://github.com/gitpod-io/gitpod/blob/3361ebffb0dc4cba55ad4f15d40d253912ce8599/components/supervisor/pkg/supervisor/tasks.go#L183

@akosyakov What this default task works for? If it's necessary, we can leave a comment here, if not, we can remove it to make the task stop works better

components/gitpod-cli/cmd/tasks-stop.go
andreafalzetti: Changed to use the `Shutdown` API.

akosyakov:
> @akosyakov What this default task works for? If it's necessary, we can leave a comment here, if not, we can remove it to make the task stop works better

I cannot remember already, probably to have at least on terminal if there are not task terminals. You can try to remove it and see how it works. I would probably try it in another PR thought?

mustard-mh:
Once the task state is `closed`, gp tasks stop will said it's not found. (but it exists in `gp tasks list` which noticed in not found logs). 

Maybe we should update not found logs or check if tasks exists in `running` list

image

mustard-mh (commented):
Code LGTM 

Nice works in `gp tasks` 👍

andreafalzetti:
> Once the task state is `closed`, `gp tasks stop` will said it's not found. (but it exists in `gp tasks list` which noticed in not found logs).
> 
> Maybe we should update not found logs or check if tasks exists in `running` list
> 
> image

@mustard-mh It's a fair point, I've updated the error message with "The selected task was not found or already stopped"

mustard-mh (approved):
Nice

Issue #12090, opened 2022-08-12
Usage list shows workspace start times 11 hours earlier than workspace list


jldec:
Probably a timezone issue. The workspace start times in the 2 areas of the UI should match.

<img width="1118" alt="Screenshot 2022-08-12 at 11 53 48" src="https://user-images.githubusercontent.com/849592/184340327-19057bfe-5228-45c6-bf6b-311693fe5369.png">
<img width="1093" alt="Screenshot 2022-08-12 at 11 53 35" src="https://user-images.githubusercontent.com/849592/184340333-4e8115fc-00e1-4791-8deb-9be3c04f125f.png">

PR #12130, opened 2022-08-15
Polish handling of BillingMode

### LINKED ISSUE #12098: Enable workspace_classes when Usage-based billing is enabled for a user

Right now, usage based pricing and workspace classes are enabled independently.

If Usage-Based Billing is enabled for a user, also enable workspace_classes. 
We should still be able to enable workspace_classes separately, for example, if an installation does not use usage based billing.

geropl:
## Description
This PR includes a couple of alignments around BillingMode. It's reviewable by commit. As all changes share the same context I think it does not make sense to further split this up. :lotus_position: 

 1. Improved tests + additional tests (no change in behavior)
 2. Make "workspace classes" usable if either of these are true:
    1. You have the "workspace_classes" FF enabled
    2. The method `BillingMode.canSetWorkspaceClass` returns `true` for your BillingMode
 3. Small fix to BillingModes: Only have _paid_ UBP team memberships override existing Chargebee memberships
 4. Handle bad "CostCenter" error, in case you have not yet set a team spending limit for whatever reason
 5. ~Make `moment` time offset work on Usage List View~ removed
 6. Fix wrong usage of `findCustomerByUserId` and replace with `findCustomerByTeamId`

## Related Issue(s)

Fixes #12098
Fixes https://github.com/gitpod-io/gitpod/issues/12090
Fixes https://github.com/gitpod-io/gitpod/issues/12149

## How to test
 - [signup](https://gpl-ubp.preview.gitpod-dev.com/workspaces)
 - create a Team called `Gitpod-heroes` or so
 - check:
   - user preferences: no workspace classes selector, yet :heavy_check_mark: 
   - user billing: no Usage-based stuff, yet :heavy_check_mark: 
   - user settings menu: still has entries for Plans/Team Subscriptions :heavy_check_mark: 
 - try to start a workspace: shows the "set spending limit" UI :heavy_check_mark: 
 - signup for a Usage-Based team plan
   - user preferences: see workspace classes selector, yet :heavy_check_mark: 
   - user billing: see the UBP attribution selector :heavy_check_mark: 
   - user settings menu: no entries for Plans/Team Subscriptions :heavy_check_mark: 

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/dashboard/src/teams/TeamBilling.tsx
+++ b/components/dashboard/src/teams/TeamBilling.tsx
@@ -28,7 +28,7 @@ import { UserContext } from "../user-context";
 type PendingPlan = Plan & { pendingSince: number };
 
 export default function TeamBilling() {
-    const { user } = useContext(UserContext);
+    const { user, userBillingMode } = useContext(UserContext);
     const { teams } = useContext(TeamsContext);
     const location = useLocation();
     const team = getCurrentTeam(location, teams);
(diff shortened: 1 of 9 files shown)


roboquat:
@geropl: Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://git.k8s.io/community/contributors/guide/release-notes.md) to remove it.

<details>

</details>

components/server/ee/src/billing/billing-mode.ts
```diff
@@ -135,12 +135,12 @@ export class BillingModesImpl implements BillingModes {
         // 3. Check team memberships/plans
         // UBB overrides wins if there is _any_. But if there is none, use the existing Chargebee subscription.
         const teamsModes = await Promise.all(teams.map((t) => this.getBillingModeForTeam(t, now)));
-        const hasUbbTeam = teamsModes.some((tm) => tm.mode === "usage-based");
+        const hasUbbPaidTeamSeat = teamsModes.some((tm) => tm.mode === "usage-based" && !!tm.paid);
```
jankeromnes, line 138: Nit: Calling this a "seat" is slightly inaccurate (it's a "membership", where you're either a member of a team or you're not; it's not a "seat", because there is no "seat" belonging to a team which can be filled by someone or be left vacant) and thus confusing. However, fixing this is not important.

components/server/ee/src/billing/billing-mode.ts
```diff
@@ -135,12 +135,12 @@ export class BillingModesImpl implements BillingModes {
...
         const hasCbTeam = teamsModes.some((tm) => tm.mode === "chargebee");
         const hasCbTeamSeat = cbTeamSubscriptions.length > 0;
 
-        if (hasUbbTeam || hasUbbPersonal) {
-            // UBB is gready: once a user has at least a team seat, they should benefit from it!
+        if (hasUbbPaidTeamSeat || hasUbbPersonal) {
+            // UBB is greedy: once a user has at least a team seat, they should benefit from it!
```
jankeromnes, line 143: Optional suggestion: Add "paid" team seat in the comment.

components/server/ee/src/billing/billing-mode.ts
geropl: Yeah, good point! Fill adjust.

components/server/ee/src/billing/billing-mode.ts
```diff
@@ -196,7 +196,15 @@ export class BillingModesImpl implements BillingModes {
...
 
-        // 3. If not: we don't even have to check for a team subscription
-        return { mode: "usage-based" };
+        // 3. Now we're usage-based. We only have to figure out whether we have a plan yet or not.
+        const result: BillingMode = { mode: "usage-based" };
+        const customer = await this.stripeSvc.findCustomerByUserId(team.id);
+        if (customer) {
+            const subscription = await this.stripeSvc.findUncancelledSubscriptionByCustomer(customer.id);
+            if (subscription) {
+                result.paid = true;
```
jankeromnes, line 205: Ah, interesting, so we _do_ potentially set it for users as well (assuming they somehow get a Stripe subscription). This somewhat contradicts earlier comments, which say `.paid` is only for teams. 💭

jankeromnes:
Thanks! Code looks good, now moving on to tests.

jankeromnes:
Question:

1. I have no paid plans, but am by default in Chargebee mode
2. I create a team called "Gitpod Something", which enables Usage-Based for both myself and my team
3. In my team, I now see the Usage-Based Billing UI, which is great 👍
4. However, in my user, I still only see the Chargebee UI

- Is this expected?
- How will I be able to upgrade my user to individual Usage-Based Billing (once that's supported)?

components/server/ee/src/billing/billing-mode.ts
geropl: This is only set for teams, because that's where we need to distinguish atm. But can imagine that it also helps for users in the future, just want to avoid more complexity.

components/server/ee/src/billing/billing-mode.ts
geropl: Done.

components/server/ee/src/billing/billing-mode.ts
jankeromnes: Ah, maybe you wanted to have `findCustomerByTeamId` instead of `findCustomerByUserId` above? 👀

geropl:
We discussed @jankeromnes [comment](https://github.com/gitpod-io/gitpod/pull/12130#issuecomment-1216699686) above: It's very well possible to not have UBP on your user. In this case, because the Team had not paid plan, yet.

components/server/ee/src/billing/billing-mode.ts
geropl: :scream: Thank you! Will fix here, as well as in the tests.

geropl:
/unhold

Commit removed.

@jankeromnes PR should be good to review - sorry for the UI glitch :bow:

jankeromnes:
Many thanks for finishing up this PR @geropl! I went through the entire "How to test" process, but found a few possible small problems.

>  - [signup](https://gpl-ubp.preview.gitpod-dev.com/workspaces)

Right after signup, and also after randomly refreshing `/` once (but it's hard to reproduce), I got this strange error:

Screenshot 2022-08-17 at 09 30 00

I believe this is when a backend service returns an error, but it's missing an HTTP error code (it's supposed to say something like `Error 500`, but instead it's just `Error ` with a space but no error code).

But maybe it's just a preview problem and not related to this PR?

>  - create a Team called `Gitpod-heroes` or so

✅ 

Additionally, this wasn't part of the test flow, but I notice that the Team Billing UI has some unfortunate flickering:

https://youtu.be/v-x4zhj25pA

Instead of briefly showing the Chargebee UI before the Usage-Based UI shows up, I think this initial state should be a loading indicator. Can be done in a follow-up though.

>  - check:
>    - user preferences: no workspace classes selector, yet :heavy_check_mark: 

❌ Actually, there is this:

Screenshot 2022-08-17 at 09 37 07

Is this not expected?

>    - user billing: no Usage-based stuff, yet :heavy_check_mark: 

✅ (although I think this should change in the future when we implement usage-based billing for individual users)

>    - user settings menu: still has entries for Plans/Team Subscriptions :heavy_check_mark: 

✅ 

>  - try to start a workspace: shows the "set spending limit" UI :heavy_check_mark: 

❌ It didn't show the "set spending limit" UI, but started a workspace successfully instead:

<table><tr><td>
</td><td>
</td></tr></table>

>  - signup for a Usage-Based team plan
>    - user preferences: see workspace classes selector, yet :heavy_check_mark: 

✅ (selector is still visible)

>    - user billing: see the UBP attribution selector :heavy_check_mark: 

❔ basically worked, although I needed to hard-refresh once in order to see the "Billing" menu entry (instead of the "Plans" and "Team Plans" entries which were still there for some reason).

Screenshot 2022-08-17 at 09 43 29

After that single refresh, everything is there, although the "Billing" menu entry now takes a long time before it shows up.

>    - user settings menu: no entries for Plans/Team Subscriptions :heavy_check_mark: 

✅ (same as above -- menu is eventually correct, but it takes a long time before the "Billing" menu entry shows up)

geropl:
Thank you for your thorough review, @jankeromnes ! :rocket: 

> Is this not expected?

The value of the `workspace_classes` FF changed in the meantime: :see_no_evil: 
image
So this is expected, yes - but makes it hard to test. :roll_eyes: We could add an exlusion case there... or try again with a non-@gitpod.io user.

> Instead of briefly showing the Chargebee UI before the Usage-Based UI shows up, I think this initial state should be a loading indicator. Can be done in a follow-up though.

> grey_question basically worked, although I needed to hard-refresh once in order to see the "Billing" menu entry (instead of the "Plans" and "Team Plans" entries which were still there for some reason).

Would you have time after 10:30 to help me fix these? :pray: Feels like it could be simple, but I'm too unfamiliar with the Dashboard to be confident.

> I believe this is when a backend service returns an error, but it's missing an HTTP error code (it's supposed to say something like Error 500, but instead it's just Error with a space but no error code).

Are there any logs on the `server` pod? :thinking:

jankeromnes:
> So this is expected, yes - but makes it hard to test. 🙄 We could add an exlusion case there... or try again with a [non-@gitpod.io](mailto:non-@gitpod.io) user.

I tried with a non-@gitpod.io user -- same problem:

1. Initially, preferences has no workspace class selector
2. After creating a "Gitpod (Something)" team, preferences _has_ a workspace class selector (contrary to test plan)
3. After upgrading team to usage-based subscription, workspace class selector remains in preferences

> Would you have time after 10:30 to help me fix these? 🙏 Feels like it could be simple, but I'm too unfamiliar with the Dashboard to be confident.

Would be more than happy to! 👍 🎨

> Are there any logs on the server pod? 🤔

I didn't check for server errors, but I'm pretty sure the `/sorry/` redirect happens here:

https://github.com/gitpod-io/gitpod/blob/aa2c51c80405c1d2da0d58840c27fa7fd0bb746b/components/proxy/conf/Caddyfile#L277-L279

jankeromnes:
Thanks @geropl!

Unfortunately, while doing one last check, I found another problem 🙈 it looks like the Chargebee team billing UI is now gone entirely:

https://youtu.be/oFNoS11vmuM

Notice that I have one team with UB flag enabled, and another team with UB flag disabled.

components/dashboard/src/teams/TeamBilling.tsx
```diff
@@ -151,163 +151,180 @@ export default function TeamBilling() {
...
+                    Team Billing automatically adds all members to the plan.{" "}
+                    <a href="https://www.gitpod.io/docs/team-billing" rel="noopener" className="gp-link">
+                        Learn more
+                    </a>
+                </div>
+            </>
+        );
+    }
+
+    const showUBP = BillingMode.showUsageBasedBilling(userBillingMode);
```
jankeromnes, line 311: In fact, why is this decision made on the user's billing mode? 🤔 Shouldn't it instead be made on the team billing mode?

geropl: This :dart: Fixed!

jankeromnes (approved):
Alright! I think we're there now. 🎉 Thanks for iterating until it worked!

Approving to unblock merging, but holding because there is still one problem that I raised in [this comment](https://github.com/gitpod-io/gitpod/pull/12130#issuecomment-1217609169):

> ❌ It didn't show the "set spending limit" UI, but started a workspace successfully instead:

The testing plan says that, when you create a "Gitpod (Something)" team, and try to start a workspace, you get blocked by a "Spending Limit" modal.

This is not the case.

Instead, I have to upgrade a team to Usage-Based, and then set a Spending Limit to very low (e.g. 0), in order to see the "Spending Limit" modal on workspace start.

However, this is likely a bug in the test scenario, and not a bug in the code, right? 😇

/hold

geropl:
> The testing plan says that, when you create a "Gitpod (Something)" team, and try to start a workspace, you get blocked by a "Spending Limit" modal.

> This is not the case.

> However, this is likely a bug in the test scenario, and not a bug in the code, right? innocent

I tested this before the rebase. As it's not critical, and more important that we don't show an arkward error message as before, I'd say let's merge this one (as it's big enough), and I will re-test and take care in a follow up.

Mentioned earlier, but here again: thank you for the through review @jankeromnes ! :pray:

Issue #12091, opened 2022-08-12
Spending limit input always inserts 1 after deleting value


jldec:
Auto-populating with 1 interferes with normal editing e.g. to replace existing value users typically delete and then type (it is less common to select and then type-over). After the deleted value has been auto-replaced with 1, the value 1 becomes impossible to delete.

<img width="1065" alt="Screenshot 2022-08-12 at 12 13 31" src="https://user-images.githubusercontent.com/849592/184343043-5c526105-ec90-4325-8b9b-e10ee653defe.png">

**Sidenote**: the little control to increase/decrease the limit does not seem useful (it's very rare to adjust up/down by just 1) - cc: @gtsiolis

PR #12148, opened 2022-08-16
[dashboard] Don't forcefully set spending limit to '1' when the input is empty

jankeromnes:
## Description

Don't forcefully set spending limit to '1' when the input is empty.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/12091

## How to test

1. Upgrade a team to usage-based billing
2. Update the spending limit
3. Delete everything in the input (in preparation for entering a new value)
4. The input should not forcefully reset to the value `1` when empty

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/dashboard/src/teams/TeamUsageBasedBilling.tsx
+++ b/components/dashboard/src/teams/TeamUsageBasedBilling.tsx
@@ -235,7 +235,9 @@ function UpdateLimitModal(props: {
     onClose: () => void;
     onUpdate: (newLimit: number) => {};
 }) {
-    const [newLimit, setNewLimit] = useState<number | undefined>(props.currentValue);
+    const [newLimit, setNewLimit] = useState<string | undefined>(
+        props.currentValue ? String(props.currentValue) : undefined,
+    );
 
     return (
         <Modal visible={true} onClose={props.onClose}>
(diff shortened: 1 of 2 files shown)


jankeromnes:
```
Error from server (InternalError): error when creating "k8s.yaml": Internal error occurred: failed calling webhook "webhook.cert-manager.io": failed to call webhook: Post "https://cert-manager-webhook.cert-manager.svc:443/validate?timeout=10s": EOF
```

Insert coin 🪙 to try again! 🎰

/werft run

components/dashboard/src/teams/TeamUsageBasedBilling.tsx
```diff
@@ -253,17 +255,22 @@ function UpdateLimitModal(props: {
...
                     className="secondary"
                     onClick={() => {
-                        if (typeof newLimit === "number") {
-                            props.onUpdate(newLimit);
+                        if (!newLimit) {
+                            return;
+                        }
+                        const n = parseInt(newLimit, 10);
+                        if (typeof n !== "number") {
+                            return;
```
geropl, line 271: Here we silently drop errors here, and actually happened to me: Had a hidden space in front :see_no_evil: . Atm we just keep the modal (no feedback)

:cloud:  We could also have an `<Alert />`, and shows parsing errors there. But I guess that's sth for a quick follow-up.

geropl (approved):
Tested and LGTM!

Added a [comment](https://github.com/gitpod-io/gitpod/pull/12148/files#r946788087), but that's more further improvement, and might need design input.

components/dashboard/src/teams/TeamUsageBasedBilling.tsx
jankeromnes: Agreed, some validation would be nice. We could also trim the input.

Also, I don't know whether we should allow setting `0` (currently reverts the limit to `undefined` it seems?) and negative values should probably not work.

PR #12157, opened 2022-08-16
feat(supervisor): add resource status severity

### LINKED ISSUE #12075: cpu/memory thresholds should be propagated by supervisor
Currently, to display high usage of CPU/Memory we replicated thresholds in [gitpod-cli](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-cli/cmd/top.go#L48-L57) and `jb backend plugin`. Ideally, supervisor should own the thresholds and propagate them to the clients.

- hardcoded cpu/memory thresholds are removed from all components
- thresholds are retrieved from supervisor like we retrieve the resource status

Places in the code to change:
- https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-cli/cmd/top.go#L48-L57
- TODO: add link to JetBrains IDE once https://github.com/gitpod-io/gitpod/pull/12053 gets merged

N/A

andreafalzetti:
## Description
- Extends `status.proto` with ResourceStatus severity to remove the duplicated severity calculation from each client
- Refactor `gp-cli` and `jetbrains backend-plugin` to consume the new severity property

## Related Issue(s)

Fixes #12075

## How to test
1. Launch a workspace https://afalz-1207846d967e0b.preview.gitpod-dev.com/#https://github.com/gitpod-io/gitpod
2. Select IntelliJ latest as preferred IDE
3. Stress the CPU/Memory and using both `gp top` and JB backend control center, see the cpu/memory thresholds change color to WARN/DANGER

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


components/supervisor-api/generate.sh
```diff
@@ -53,7 +53,7 @@ local_java_protoc() {
         --java_out=java/src/main/java \
         ./*.proto
     # remove trailing spaces
-    find components/supervisor-api/java/src/main/java/io/gitpod/supervisor/api -maxdepth 1 -name "*.java" -exec sed -i -e "s/[[:space:]]*$//" {} \;
```
andreafalzetti, line 56: Do we need this like anymore?

I had to comment it out, because it fails:

```
find: ‘components/supervisor-api/java/src/main/java/io/gitpod/supervisor/api’: No such file or directory
```

jeanp413 (commented):
To be honest I don't see the benefit of this change, APIs should provide the basic info/data  and users of the API can use/interpret it freely. Even though it seems repetitive to map from values and threshold it's fine and the code is really simple.
I also see you are doing a mapping either way in `GitpodMetricProvider.kt` it's the same in the end.

andreafalzetti:
Fair point 🤔 From my point of view, we already had inconsistency between `gp-cli` and JB IDEs having different thresholds [[1]](https://github.com/gitpod-io/gitpod/pull/12087), from a client perspective, it makes sense to me having only to map `severity` to `color` instead of `value` to `color`.

jeanp413:
> it makes sense to me having only to map severity to color instead of value to color

yeah I see that but that's more UI related so the API shouldn't be responsible of that.

Anyway if everybody else aproves just ignore my review :smile:

akosyakov:
The main motivation is to having one place to control thresholds, not updating each client if we figure out that danger should .9 not .95.

components/supervisor/pkg/supervisor/top.go
```diff
@@ -104,14 +115,19 @@ func Top(ctx context.Context) (*api.ResourcesStatusResponse, error) {
 			return nil, xerrors.Errorf("could not retrieve workspace info: %w", err)
 		}
 
+		cpuPercentage := int64((float64(resp.Resources.Cpu.Used) / float64(resp.Resources.Cpu.Limit)) * 100)
```
akosyakov, line 118: Code in if block above is missing

akosyakov (approved):
it works as advertized, but we don't compute it in supervisor in all branches, see https://github.com/gitpod-io/gitpod/pull/12157/files#r947179792

/hold

components/supervisor/pkg/supervisor/top.go
andreafalzetti: Fixed

andreafalzetti: Thanks!

PR #12162, opened 2022-08-16
[billing] Check if network connections should be limited

Furisto:
## Description
If a user is on the free tier, we want to limit the network connections that they can create.

## Related Issue(s)
Fixes https://github.com/gitpod-io/gitpod/issues/12161)

## Release Notes

```release-note
None
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/server/ee/src/billing/entitlement-service-chargebee.ts
+++ b/components/server/ee/src/billing/entitlement-service-chargebee.ts
@@ -189,4 +189,15 @@ export class EntitlementServiceChargebee implements EntitlementService {
 
         return true;
     }
+
+    /**
+     * Returns true if network connections should be limited
+     * @param user
+     */
+    async limitNetworkConnections(user: User, date: Date = new Date()): Promise<boolean> {
+        const subscriptions = await this.subscriptionService.getNotYetCancelledSubscriptions(user, date.toISOString());
+        const freePlans = [Plans.FREE, Plans.FREE_50].map((p) => p.chargebeeId);
+
+        return subscriptions.filter((s) => !freePlans.includes(s.planId!)).length > 0;
+    }
 }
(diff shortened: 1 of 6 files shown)


components/server/ee/src/billing/entitlement-service-ubp.ts
```diff
@@ -97,6 +97,14 @@ export class EntitlementServiceUBP implements EntitlementService {
         return this.hasPaidSubscription(user, date);
     }
 
+    /**
+     * Returns true if network connections should be limited
+     * @param user
+     */
+    async limitNetworkConnections(user: User, date: Date): Promise<boolean> {
+        return !this.hasPaidSubscription(user, date);
```
geropl, line 105: Hehe, this is a TS/JS subtility: This will always be "false".
This works, though:

`        return !(await this.hasPaidSubscription(user, date));`

components/server/ee/src/billing/entitlement-service-chargebee.ts
```diff
@@ -189,4 +189,15 @@ export class EntitlementServiceChargebee implements EntitlementService {
...
         return true;
     }
+
+    /**
+     * Returns true if network connections should be limited
+     * @param user
+     */
+    async limitNetworkConnections(user: User, date: Date = new Date()): Promise<boolean> {
+        const subscriptions = await this.subscriptionService.getNotYetCancelledSubscriptions(user, date.toISOString());
+        const freePlans = [Plans.FREE, Plans.FREE_50].map((p) => p.chargebeeId);
```
geropl, line 199: This is nearly correct. Let me give it a spin.

geropl: Ah, was correct indeed. Introduced a dedicated method `Plans.isFreeTier` nonetheless.

components/server/ee/src/billing/entitlement-service-ubp.ts
geropl: Fixed.

geropl (approved):
Thx @Furisto, code LGTM!

/hold
Took the liberty to push a commit on top, please unhold if you're fine with that.

PR #12178, opened 2022-08-17
[usage] Test for and fix negative credits in GenerateUsageReport

### LINKED ISSUE #12064: Negative credits in usage list view
Something is wrong with the minutes -> credits computation.

Screenshot 2022-08-11 at 12 42 04

geropl:
## Description
There was a disconnect between when we determined `now`, and when we used it in combination with start-/stopTimes that - based on runtime behavior - _could_ lead to negative usage being calculated.

## Related Issue(s)

Fixes #12064 

### Run unit tests
 - [open a workspace](https://gitpod.io/#https://github.com/gitpod-io/gitpod/pull/12178)
 - `leeway build components/usage:init-testdb && (cd components/usage && go test -v ./...)`

### Manual test
I did not manage to provoke this situation in preview envs, mainly due to data volume I guess.
It's still worth testing that usage still works as expected: 
 - signup
 - create a team starting with `Gitpod-...`
 - import a project
 - start a workspace on that project
 - go to the usage view and notice the correct usage being displayed

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/usage/pkg/apiv1/usage-report.go
+++ b/components/usage/pkg/apiv1/usage-report.go
@@ -7,10 +7,13 @@ package apiv1
 import (
 	"context"
 	"fmt"
+	"time"
+
 	"github.com/gitpod-io/gitpod/common-go/log"
 	"github.com/gitpod-io/gitpod/usage/pkg/db"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"gorm.io/gorm"
-	"time"
 )
 
 type InvalidSession struct {
@@ -44,11 +47,21 @@ type ReportGenerator struct {
 	nowFunc func() time.Time
 }
 
-func (g *ReportGenerator) GenerateUsageReport(ctx context.Context, from, to time.Time) (UsageReport, error) {
+func (g *ReportGenerator) GenerateUsageReport(ctx context.Context, from, to time.Time) (*UsageReport, error) {
 	now := g.nowFunc().UTC()
-	log.Infof("Gathering usage data from %s to %s", from, to)
 
-	report := UsageReport{
+	// Sanity check: from <= now
+	if now.Before(from) {
+		return nil, status.Errorf(codes.InvalidArgument, "Now must be before from")
+	}
+
+	// Enforce: to <= now
+	if now.Before(to) {
+		to = now
+	}
+	log.Infof("Gathering usage data from %s to %s (%s)", from, to, now)
+
+	report := &UsageReport{
 		GenerationTime: now,
 		From:           from,
 		To:             to,
@@ -70,7 +83,7 @@ func (g *ReportGenerator) GenerateUsageReport(ctx context.Context, from, to time
 
 	trimmed := trimStartStopTime(valid, from, to)
 
-	report.UsageRecords = instancesToUsageRecords(trimmed, g.pricer, now)
+	report.UsageRecords = instancesToUsageRecords(trimmed, g.pricer, to)
 	return report, nil
 }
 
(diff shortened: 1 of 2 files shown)


geropl:
/werft run

:crossed_fingers: npmjs.org...

components/usage/pkg/apiv1/usage_test.go
```diff
@@ -297,8 +299,132 @@ func TestReportGenerator_GenerateUsageReport(t *testing.T) {
 
 	require.Equal(t, nowFunc(), report.GenerationTime)
 	require.Equal(t, startOfMay, report.From)
-	require.Equal(t, startOfJune, report.To)
+	// require.Equal(t, startOfJune, report.To) TODO(gpl) This is not true anymore - does it really make sense to test for it?
```
jankeromnes, line 302: Hmm, you're right. If we're looking at the usage from May to June, and `now` is some time in May, then `to` should be equal to `now` (and not `startOfJune`, which would only be the case if `now` is some time in June or later).

components/usage/pkg/apiv1/usage_test.go
```diff
@@ -297,8 +299,132 @@ func TestReportGenerator_GenerateUsageReport(t *testing.T) {
...
+				// 	{
+				// 		InstanceID: instanceID,
+				// 		AttributionID: db.NewTeamAttributionID(teamID.String()),
+				// 		StartedAt: Timestamp("2022-08-17T09:40:53.115Z").Time(),
+				// 		StoppedAt: sql.NullTime{ Time: Timestamp("2022-08-17T09:43:04.874Z").Time(), Valid: true },
+				// 		WorkspaceClass: "default",
+				// 		CreditsUsed: 3.0,
+				// 	},
+				// },
+			},
```
jankeromnes, line 360: No expectation here? 🤔

components/usage/pkg/apiv1/usage-report.go
```diff
@@ -44,11 +47,21 @@ type ReportGenerator struct {
...
 
-func (g *ReportGenerator) GenerateUsageReport(ctx context.Context, from, to time.Time) (UsageReport, error) {
+func (g *ReportGenerator) GenerateUsageReport(ctx context.Context, from, to time.Time) (*UsageReport, error) {
 	now := g.nowFunc().UTC()
-	log.Infof("Gathering usage data from %s to %s", from, to)
 
-	report := UsageReport{
+	// Sanity check: from <= now
+	if now.Before(from) {
+		return nil, status.Errorf(codes.InvalidArgument, "Now must be before from")
```
jankeromnes, line 55: Error message is wrong:

```suggestion
		return nil, status.Errorf(codes.InvalidArgument, "Now must be after from")
```

components/usage/pkg/apiv1/usage_test.go
geropl: Yes, can make can more explicit with: `usageRecords: nil`. :+1:

jankeromnes:
Note: I've added the Werft flag `with-payment`, which triggered a new build. My intent is to test the credit-counting behavior (i.e. start workspaces in various billing states, then double-check how many credits we've reported to Stripe).

components/usage/pkg/apiv1/usage_test.go
geropl: There are other potential solutions, but this was the least invasive - and seems to be in line with what we expect at other places. Happy to turn this one around when Milan/Andrew come back (and do better filtering on the instances we retrieve, for example) - as long as the test passes! :wink:

jankeromnes (approved):
Code looks good to me, test works, and I've also tested the credit-counting/reporting behavior with a few instances. Works great!

/hold

geropl:
/unhold

Handled review comments, so good to go! :racehorse:

PR #12192, opened 2022-08-18
[server] Introduce CachedBillingModes

geropl:
## Description
This introduces a cache-layer on the `BillingModes` interface which caches BM calculations for 10s max. to prevent excessive impacts during the usage-based rollout.

Note: 

## Related Issue(s)

Fixes #

## How to test
 - signup
 - create a team starting with `Gitpod-...`
 - noticed that you can see the UBP billing, but _not_ the regular Team Subscription view (BM works)
 - upgrade to paid
 - noticed how your personal menu no longer shows

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/gitpod-protocol/src/util/garbage-collected-cache.ts
+++ b/components/gitpod-protocol/src/util/garbage-collected-cache.ts
@@ -32,9 +32,14 @@ export class GarbageCollectedCache<T> {
         });
     }
 
-    public get(key: string): T | undefined {
+    public get(key: string, now: Date = new Date()): T | undefined {
         const entry = this.store.get(key);
         if (!entry) {
+            return undefined;
+        }
+        // Still valid?
+        if (entry.expiryDate < now.getTime()) {
+            this.store.delete(entry.key);
             return undefined;
         }
         return entry.value;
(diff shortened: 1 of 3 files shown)


jankeromnes (changes requested):
As discussed, this caching seems like premature optimization:
- We're addressing a suspected performance problem (but we're not sure it's an actual problem, i.e. we have no data to back this suspicion up)
- We believe our fix makes the situation better (but, since we have no data, it could also have no performance impact, or worse, maybe it makes it even slower for some unforeseen reason?)

However, let's not throw this away entirely, since the code change looks good, and it might improve things.

Thanks @geropl for finding a good compromise -- maybe we can:
1. Enable BillingModes for everyone in production (and double-check this doesn't break anyone)
2. Look at tracing for `startWorkspace` durations
3. Then, we can do a follow-up deployment of this change after a few days, and compare tracing (if it makes things faster, great 👍 or, if it doesn't have an effect, or has bugs, maybe let's revert?)

geropl:
Good call @jankeromnes ! :100: 

I will:
 - rollout isBillindMode in prod (incrementally)
 - monitor:
   - startWorkspace duration (based on traces)
   - DB CPU usage
 - get back to this if we deem it to be too bad

Issue #12177, opened 2022-08-17
Validate we're calculcating Usage based on correct dates


geropl:
1. Check what we currently do for Chargebee, and align with what we _want_ to do: startedTime -> stoppingTime
 2. Check `usage` component and make it the same there

/cc @jankeromnes @jldec

PR #12195, opened 2022-08-18
[usage] Change instance runtime calculation: creationTime → startedTime, stoppedTime → stoppingTime

jankeromnes:
## Description

Change instance runtime calculation:
- Start: use startedTime instead of creationTime
  - don't count startup time at all, e.g. scale up, image build/pull, ...
  - don't double-count when waiting for a prebuild to finish
- Stop: use stoppingTime instead of stoppedTime
  - don't count stopping time at all, e.g. backup time

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/12177

## How to test

1. Unit tests

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/usage/pkg/apiv1/usage-report.go
+++ b/components/usage/pkg/apiv1/usage-report.go
@@ -104,8 +104,8 @@ func validateInstances(instances []db.WorkspaceInstanceForUsage) (valid []db.Wor
 		start := instance.CreationTime.Time()
 
 		// Currently running instances do not have a stopped time set, so we ignore these.
-		if instance.StoppedTime.IsSet() {
-			stop := instance.StoppedTime.Time()
+		if instance.StoppingTime.IsSet() {
+			stop := instance.StoppingTime.Time()
 			if stop.Before(start) {
 				invalid = append(invalid, InvalidSession{
 					Reason:  "stop time is before start time",
@@ -125,12 +125,12 @@ func trimStartStopTime(instances []db.WorkspaceInstanceForUsage, maximumStart, m
 	var updated []db.WorkspaceInstanceForUsage
 
 	for _, instance := range instances {
-		if instance.CreationTime.Time().Before(maximumStart) {
-			instance.CreationTime = db.NewVarcharTime(maximumStart)
+		if instance.StartedTime.Time().Before(maximumStart) {
+			instance.StartedTime = db.NewVarcharTime(maximumStart)
 		}
 
-		if instance.StoppedTime.Time().After(minimumStop) {
-			instance.StoppedTime = db.NewVarcharTime(minimumStop)
+		if instance.StoppingTime.Time().After(minimumStop) {
+			instance.StoppingTime = db.NewVarcharTime(minimumStop)
 		}
 
 		updated = append(updated, instance)
--- a/components/usage/pkg/apiv1/usage_test.go
+++ b/components/usage/pkg/apiv1/usage_test.go
@@ -177,9 +177,9 @@ func TestInstanceToUsageRecords(t *testing.T) {
 	teamAttributionID := db.NewTeamAttributionID(teamID)
 	instanceId := uuid.New()
 	creationTime := db.NewVarcharTime(time.Date(2022, 05, 30, 00, 00, 00, 00, time.UTC))
-	startedTime := db.NewVarcharTime(time.Date(2022, 05, 30, 00, 00, 00, 00, time.UTC))
+	startedTime := db.NewVarcharTime(time.Date(2022, 05, 30, 00, 01, 00, 00, time.UTC))
 	stoppingTime := db.NewVarcharTime(time.Date(2022, 06, 1, 1, 0, 0, 0, time.UTC))
-	stoppedTime := db.NewVarcharTime(time.Date(2022, 06, 1, 1, 0, 0, 0, time.UTC))
+	stoppedTime := db.NewVarcharTime(time.Date(2022, 06, 1, 1, 1, 0, 0, time.UTC))
 
 	scenarios := []struct {
 		Name     string
@@ -211,7 +211,7 @@ func TestInstanceToUsageRecords(t *testing.T) {
 				ProjectID:      "",
 				WorkspaceType:  db.WorkspaceType_Prebuild,
 				WorkspaceClass: defaultWorkspaceClass,
-				CreditsUsed:    470,
+				CreditsUsed:    469.8333333333333,
 				StartedAt:      startedTime.Time(),
 				StoppedAt:      sql.NullTime{Time: stoppingTime.Time(), Valid: true},
 				GenerationID:   0,
@@ -242,10 +242,10 @@ func TestInstanceToUsageRecords(t *testing.T) {
 				ProjectID:      projectID.String(),
 				WorkspaceID:    workspaceID,
 				WorkspaceType:  db.WorkspaceType_Regular,
-				StartedAt:      creationTime.Time(),
+				StartedAt:      startedTime.Time(),
 				StoppedAt:      sql.NullTime{},
 				WorkspaceClass: defaultWorkspaceClass,
-				CreditsUsed:    470,
+				CreditsUsed:    469.8333333333333,
 			}},
 		},
 	}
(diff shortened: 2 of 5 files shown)


svenefftinge:
I don't know the background of this decision, but I think counting in the starting and stopping time makes sense, especially until we bill usage for storage because the starting and stopping phase highly depend on the size of the workspace backup.

jankeromnes:
Thanks for your feedback @svenefftinge!

> I don't know the background of this decision, but I think counting in the starting and stopping time makes sense, especially until we bill usage for storage because the starting and stopping phase highly depend on the size of the workspace backup.

This is exactly the reason we initially went with `runtime = stoppedTime - creationTime`.

However, there is one problem with `creationTime` -- when you start a workspace, and there is a prebuild currently running, and you decide to wait for it, if we count the runtime from `creationTime`, we'll end up counting the "wait for prebuild" time twice (we bill you once for the full prebuild, and once for the full workspace runtime including the waiting time).

Regarding `stoppingTime` vs `stoppedTime`, I agree with you that it's roughly proportional to the size of the workspace. However, this will change once we roll out PVC mounts, and @geropl also mentioned that we sometimes don't reliably set `stoppedTime` (e.g. when instances get stuck in `stopping`) which wouldn't be fair to bill users for.

However, I do share your worries about the shorter runtime tracking -- by doing `stoppingTime - startedTime`, we don't account for:
- scale-up time
- image build time
- image pull time
- git clone or backup extraction time
- backup time / workspace size

That's maybe okay at first, but might come back to bite us in the future e.g. when we want to start billing any of these, thus making Gitpod more expensive for users (it's always easier to start expensive and make it cheaper, than to start cheap and regularly make it more expensive 😬). But since `creationTime` has the double-billing issue, I don't currently have a good solution to this. I feel like this is still very open to discussion.

geropl:
@jankeromnes Thanks for identifying the status quo, incl. the (potential) changes.

@svenefftinge First and foremost, I wanted to have the discussion and document it somewhere. Maybe I just wasn't part of the last one..? :shrug: 

For context: [here is](https://github.com/gitpod-io/gitpod/blob/c715b3521f2a2d68937cdd44ecfa35674b3c10c5/components/ee/payment-endpoint/src/accounting/account-service-impl.ts#L261-L262) what we do now (Chargebee): started -> stopping

Regarding "when does the session a) start and b) end":
 - start:
    - I feel we're already billing users for prebuilds (which is the where value added)
    - I don't see benefit in billing for scale-up time, because it adds no value for users
    - Things we might missing out on: a) image-build times, b) pull times. Especially if we plan to bill for a) separately later one, and find a way to exclude b), it might make sense to keep as is for now. But be aware that "time spent (waiting) on prebuilds" >>> "time spent (waiting) on image-builds"
 - end:
   - I see the backup argument, but that should really be gone with PVC
   - Also, in a lot of technical failure scenarios we set `stoppedTime` very late, which was one of the reasons to introduce `stoppingTime` (which we configure "on user intent") in the first place. If we move to `stoppedTime` again, I fear we'll be subject to more complaints here.

I have no hard feelings here, but feel that `end == stoppingTime` is a tad more important to me bc of the reliability argument.

svenefftinge:
Ok, thanks for sharing your thoughts. Makes sense 👍

components/usage/pkg/apiv1/usage_test.go
```diff
@@ -280,6 +287,7 @@ func TestReportGenerator_GenerateUsageReport(t *testing.T) {
 			ID:                 uuid.New(),
 			UsageAttributionID: db.NewTeamAttributionID(teamID.String()),
 			StartedTime:        db.NewVarcharTime(time.Date(2022, 06, 1, 1, 0, 0, 0, time.UTC)),
+			StoppingTime:       db.NewVarcharTime(time.Date(2022, 06, 1, 1, 0, 0, 0, time.UTC)),
 			StoppedTime:        db.NewVarcharTime(time.Date(2022, 06, 1, 1, 0, 0, 0, time.UTC)),
```
svenefftinge, line 291: the comment says: `No creation time, invalid record`. that should be no `startedTime` and `StartedTime` should be removed, I guess.

components/usage/pkg/apiv1/usage_test.go
```diff
@@ -266,6 +272,7 @@ func TestReportGenerator_GenerateUsageReport(t *testing.T) {
 			UsageAttributionID: db.NewTeamAttributionID(teamID.String()),
 			CreationTime:       db.NewVarcharTime(time.Date(2022, 05, 1, 00, 00, 00, 00, time.UTC)),
 			StartedTime:        db.NewVarcharTime(time.Date(2022, 05, 1, 00, 00, 00, 00, time.UTC)),
+			StoppingTime:       db.NewVarcharTime(time.Date(2022, 06, 1, 1, 0, 0, 0, time.UTC)),
```
svenefftinge, line 275: It would be great to use slightly different times on for `StoppingTime` and `StoppedTime` as well as `CreationTime` `StartedTime` respectively. So the test better communicates that we measure between `StartedTime` and `StoppingTime`

components/usage/pkg/apiv1/usage-report.go
```diff
@@ -125,12 +125,12 @@ func trimStartStopTime(instances []db.WorkspaceInstanceForUsage, maximumStart, m
 	var updated []db.WorkspaceInstanceForUsage
 
 	for _, instance := range instances {
-		if instance.CreationTime.Time().Before(maximumStart) {
-			instance.CreationTime = db.NewVarcharTime(maximumStart)
+		if instance.StartedTime.Time().Before(maximumStart) {
+			instance.StartedTime = db.NewVarcharTime(maximumStart)
```
svenefftinge, line 129: Unrelated to this PR, but manipulating instances here was surprising, especially since they are handed out through the usage report. I don't have enough context, yet, though. Just wanted to share.

components/usage/pkg/apiv1/usage_test.go
jankeromnes: Good point, thanks! There is one test with real numbers from a DB example, but I guess the other tests could also have small differences between create/start and stopping/stopped.

components/usage/pkg/apiv1/usage-report.go
jankeromnes: I agree, and think it would make more sense to manipulate usage records instead of instances.

components/usage/pkg/apiv1/usage_test.go
svenefftinge: Yes, it's also easier to check as a human when using artificial numbers rather than real numbers.

components/usage/pkg/apiv1/usage-report.go
geropl: They are not manipulated, as we create copies here (not working on references/pointers).

But had to read twice as well.. :laughing:  :see_no_evil:

components/usage/pkg/apiv1/usage_test.go
jankeromnes: Fixed! Thanks again

components/usage/pkg/apiv1/usage_test.go
jankeromnes: Hmm, is this actually the case? This part of the query hasn't changed:

https://github.com/gitpod-io/gitpod/blob/135bc0f7586658b0e8730a2ff67d19bc7322b393/components/usage/pkg/db/workspace_instance.go#L79-L80

And when I remove the `StartedTime`, this changes the test behavior from {3 raw sessions; 1 invalid session; 2 usage records} to {2 raw sessions; 0 invalid session; 2 usage records}.

For now I've rolled back this suggestion, but please let me know if the test / the query logic should change here.

geropl:
Had a discussion with @jankeromnes again.

We noticed that if we would (have and) use `deployedTime` instead of `startedTime`, we would have an even better cut. It would:
 - exclude image-builds :white_check_mark: 
 - exclude prebuilds :heavy_check_mark: 
 - exclude scale-up time :heavy_check_mark: 
 - include pull times :heavy_check_mark: 
 - include content init (with PVC) :heavy_check_mark: 

As we do not have `deployedTime` anymore, we decided to move forward with this PR as is for now.

I will document the discussion that happened here, and will ask Team Workspace how much effort it is to re-vive `deployedTime`.

geropl (approved):
Code LGTM and tests work.

Issue #12015, opened 2022-08-09
Validate gitpod config patch, rendered gitpod installer YAML in preflight checks


adrienthebo:
The gitpod config patch provides a very low level interface to modify Gitpod's installation configuration. A malformed config patch can produce invalid YAML, or valid YAML that doesn't follow the gitpod-installer config schema.

Malformed configuration should be caught at the preflight check and should ideally provide some level of diagnostics about what's failing and how to fix it.

- The gitpod config patch file is valid YAML
- The merge result of the gitpod config patch and the gitpod installer config is valid YAML
- The final gitpod config YAML follows the gitpod-installer schema (as validated with `/app/installer validate`)

## Steps to reproduce

The following YAML patch will fail on current versions of gitpod:

```yaml

experimental:
  webapp:
    proxy:
      # allowed values are "ClusterIP", "LoadBalancer", "NodePort", "ExternalName" consistent with https://pkg.go.dev/k8s.io/api/core/v1#ServiceType
      serviceType: "NodePort"
```

```shell
# /app/installer validate config -c config.yaml 
{
  "valid": false,
  "warn": [
    "Deprecated config parameter: experimental.webapp.proxy.serviceType=NodePort"
  ],
  "fatal": [
    "Cannot set proxy service type in both components and experimental"
  ]
}
Error: configuration invalid
Usage:
  gitpod-installer validate config [flags]
```

Pothulapati:
👋

Un-assigning myself as this is now planned to be done differently with the new installer env abstraction. Happy to take it up in the new path too if you haven't already made progress @MrSimonEmms

MrSimonEmms:
Thanks @Pothulapati. I've added this to the main refactoring ticket

PR #12202, opened 2022-08-18
[installer]: move the kots install script into a bash file in Installer

### LINKED ISSUE #12197: move the bash script from installer-job to Installer container image
The bash script in the KOTS installer job has become far bigger than was ever intended. Having this written in the `args` section of a Kubernetes job is now working against us and we lose code completion, shellcheck by having it in there.

Move the script in the job into `/scripts` directory in the Installer and copy these into the container image at build stage so these can be run.

The `{{repl }}` options that are rewritten by KOTS at runtime will need to be passed to the script as environment variables, ensuring that no sensitive values are passed in this way. All efforts should be taken to remove the need for these in the scripts entirely and use secrets for these if they cannot be avoided entirely.

MrSimonEmms:
## Description

Moves the bash script in the installer job into the installer image. This is done to make the application more reliable, testable and readable. This makes no fundamental changes to how the job works except for fixing ShellCheck errors and changing all the `{{repl }}` references to environment variables.

This also improves upon the wording and gives troubleshooting steps in an effort to reduce the (small, but not insignificant) number of "help, my installation-status pod is in an errored state" issues we see raised.

## Related Issue(s)

Fixes #12197
Fixes https://github.com/gitpod-io/gitpod/issues/12259

## How to test

Install via KOTS.

I've checked this against all permutations that I can work out (k3s, AKS, EKS, GKE) and works on all.

## Release Notes

```release-note
[installer]: move the kots install script into a bash file in Installer
```

## Werft options:

- [ ] /werft with-preview


install/kots/manifests/gitpod-kots-config.yaml
```diff
@@ -0,0 +1,60 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: gitpod-kots-config
+  labels:
+    app: gitpod
+    component: gitpod-installer
+data:
+  # General settings
+  CURSOR: repl{{ Cursor | quote }}
```
corneliusludmann, line 13: :bulb: (idea; non-blocking) Asking Replicated for a generic feature for this

---

Like this approach a lot.

I wonder if we should ask Replicated whether there is a built-in way to get the whole KOTS config as config map or if not, create a feature request. A template function that returns the whole KOTS config as YAML would probably be good enough.

I think, that creating a generic solution in KOTS should not be that hard for Replicated. It would make our lives much easier to maintain this.

What do you think?

_However, that does not block this pull request at all!_

MrSimonEmms: It is possible to get this as a configmap, which we indeed do in the [support bundle](https://github.com/gitpod-io/gitpod/blob/main/install/kots/manifests/kots-support-bundle.yaml#L109,L122).

The problem with this as an approach is the KOTS config will have many sensitive pieces of data (passwords for DB/Registry/Storage etc) which we don't actually need for the Installer, as these are sent into secrets that the Installer then references.

We could of course store this as a secret rather than a configmap, but this would still contain information that we don't actually need for the Installer. I do see the benefit to not having to manage this though, but this then has the danger of thinking we're exposing something that we're not and errors being caused.

I'd propose that, for this PR, we keep it as-is and continue to discuss this subject when doing the general refactoring that we're doing.

corneliusludmann: > I'd propose that, for this PR, we keep it as-is and continue to discuss this subject when doing the general refactoring that we're doing.

100 % agree.

MrSimonEmms:
This is ready for test/review. I will leave the "hold" status as think this would benefit from multiple testers/approvers

install/kots/manifests/gitpod-kots-config.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: gitpod-kots-config
+  labels:
+    app: gitpod
+    component: gitpod-installer
+data:
+  # General settings
```
Pothulapati, line 12: This looks good overall on a macro level, but my main concern is that these options are now publicly visible right? esps the registry, database values being here?

Can we use a secret instead, and then load the env in the `gitpod-installer-job` using a `secretReg`? :thinking: WDYT?

Pothulapati: AHHHH, Right! They aren't changed. They follow the existing method of loading through secrets! 👍🏼

MrSimonEmms: Yeah, I've tried to avoid changing anything significantly in this PR - that comes later :)

install/installer/scripts/kots-install.sh
```diff
@@ -0,0 +1,345 @@
...
+    echo "Gitpod: containerd socket detected as k3s"
+
+    yq e -i ".workspace.runtime.containerdSocket = \"${CONTAINERD_SOCKET_K3S}\"" "${CONFIG_FILE}"
+elif [ -S "/mnt/node0${CONTAINERD_SOCKET_AL}" ]; then
+    echo "Gitpod: containerd socket detected as ${CONTAINERD_SOCKET_AL}"
+
+    yq e -i ".workspace.runtime.containerdSocket = \"${CONTAINERD_SOCKET_AL}\"" "${CONFIG_FILE}"
+fi
+
+echo "Gitpod: Inject the Replicated variables into the config"
```
Pothulapati, line 74: *Definitely not asking in this PR*, but a next iteration of this change would be to move all the config settings into a separate script so that we can validate the end config as a pre-flight check right? Should be simple enough and hence solving https://github.com/gitpod-io/gitpod/issues/12015 :thinking:

install/installer/scripts/kots-install.sh
```diff
@@ -0,0 +1,345 @@
...
+    yq e -i ".imagePullSecrets[0].kind = \"secret\"" "${CONFIG_FILE}"
+    yq e -i ".imagePullSecrets[0].name = \"${IMAGE_PULL_SECRET_NAME}\"" "${CONFIG_FILE}"
+    yq e -i '.dropImageRepo = true' "${CONFIG_FILE}"
+
+    # Add the registry to the server allowlist - keep docker.io in case it's just using the mirrored registry functionality without being airgapped
+    yq e -i ".containerRegistry.privateBaseImageAllowList += \"${LOCAL_REGISTRY_HOST}\"" "${CONFIG_FILE}"
+    yq e -i ".containerRegistry.privateBaseImageAllowList += \"docker.io\"" "${CONFIG_FILE}"
+fi
+
+if [ "${REG_CUSTOM_DOCKER_ENABLED}" = "1" ];
```
Pothulapati, line 118: nit: Should it be called `REG_CUSTOM_DOCKER_CONFIG_ENABLED`? Would prefer to have `CONFIG` to be more specific.

install/installer/scripts/kots-install.sh
```diff
@@ -0,0 +1,345 @@
...
+
+# Output the local registry secret - this is proxy.replicated.com if user hasn't set their own
+echo "${LOCAL_REGISTRY_IMAGE_PULL_SECRET}" | base64 -d > /tmp/kotsregistry.json
+
+if [ "${REG_INCLUSTER_ENABLED}" = "0" ];
+then
+    echo "Gitpod: configuring external container registry"
+
+    # Get the external-container-registry secret so we can merge the external registry and KOTS registry keys
+    kubectl get secret external-container-registry \
```
Pothulapati, line 140: :thumsup: Like that this is now routed through a secret, instead of building it here!

install/installer/scripts/kots-install.sh
```diff
@@ -0,0 +1,345 @@
+#!/bin/sh
+# Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+# Licensed under the MIT License. See License-MIT.txt in the project root for license information.
+
+# shellcheck disable=SC2050,SC2153
+
+set -e
```
Pothulapati, line 7: Did a line by line run through, and it matches with the previous `gitpod-installer-job.yaml`

install/installer/scripts/kots-install.sh
```diff
@@ -0,0 +1,345 @@
...
+        echo "Gitpod: Uninstall application"
+        helm uninstall -n "${NAMESPACE}" gitpod --wait || true
+    else
+        echo "Gitpod: Rolling back application"
+        helm rollback -n "${NAMESPACE}" gitpod --wait || true
+    fi
+fi
+
+echo "Gitpod: Create a Helm template directory"
+rm -Rf "${GITPOD_OBJECTS}"
```
Pothulapati, line 30: nit: Maybe document somewhere in the file about GITPOD_OBJECTS? :thinking: While its straightforward with other env's as they are input it could be confusing on what `kots-install.sh` emits.

Pothulapati (commented):
Actually liking this approach, as it makes the kots -> install config a lot more simpler! The code looks good, will do a another pass with manual testing after which it should be ✅ 

Great Work! :)

install/installer/scripts/kots-install.sh
MrSimonEmms: That's exactly the next step - good thought

install/installer/scripts/kots-install.sh
MrSimonEmms: If it's ok with you, I won't fix in here - this won't be sticking around long as I've already started the next refactor and this probably won't survive. If it does, I'll fix it then

Pothulapati: Sounds good!

install/installer/scripts/kots-install.sh
MrSimonEmms: I've renamed it to `REG_DOCKER_CONFIG_ENABLED` to that keeps consistency with what I've done elsewhere (take the KOTS config name and captialise it). I've not changed the underlying KOTS config name (it's missing a `d`) in the expectation that someone could have started using it

Pothulapati (approved):
✅ from my side. Upgraded my Gitpod instance on AWS to use this version, and all the installer, status and preflight checks worked as expected.

PR #12205, opened 2022-08-18
[installer-tests] Consolidate slack alerts to reduce noise

nandajavarma:
## Description

This is a very small PR that reduces the number of alerts sent out for integration test failures. Now test failures for one slack channel(one team) will be sent out only once in a consolidated message.

## Related Issue(s)

Fixes #

## How to test

One can use this doc to [try and trigger tests](https://www.notion.so/gitpod/WIP-Self-hosted-automated-nightly-pipelines-fdc5a202e67b4197aa00d93b98d57927).

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/.werft/installer-tests.ts
+++ b/.werft/installer-tests.ts
@@ -205,51 +205,51 @@ const TESTS: { [name: string]: InfraConfig } = {
     WORKSPACE_TEST: {
         phase: "run-workspace-tests",
         makeTarget: "run-workspace-tests",
-        description: "Run the workspace tests test/tests/workspace",
+        description: "Workspace integration tests",
         slackhook: slackHook.get("workspace-jobs"),
     },
     VSCODE_IDE_TEST: {
         phase: "run-vscode-ide-tests",
         makeTarget: "run-vscode-ide-tests",
-        description: "Run the vscode IDE tests test/tests/ide/vscode",
+        description: "vscode IDE tests",
         slackhook: slackHook.get("ide-jobs"),
     },
     JB_IDE_TEST: {
         phase: "run-jb-ide-tests",
         makeTarget: "run-jb-ide-tests",
-        description: "Run the jetbrains IDE tests test/tests/ide/jetbrains",
+        description: "jetbrains IDE tests",
         slackhook: slackHook.get("ide-jobs"),
     },
     CONTENTSERVICE_TEST: {
         phase: "run-cs-component-tests",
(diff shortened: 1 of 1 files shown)


.werft/installer-tests.ts
```diff
@@ -345,6 +345,7 @@ export async function installerTests(config: TestConfig) {
 
 function runIntegrationTests() {
     werft.phase("run-integration-tests", "Run all existing integration tests");
+    var slackAlerts = new Map<string, string>([])
```
MrSimonEmms, line 348: nit: probably could be a `const` rather than a `var`

MrSimonEmms, line 348:
```suggestion
    const slackAlerts = new Map<string, string>([])
```
nit: probably shouldn't use `var`

.werft/installer-tests.ts
```diff
@@ -353,10 +354,16 @@ function runIntegrationTests() {
                 `werft log result -d "failed test" url "${testPhase.description}(Phase ${testPhase.phase}) failed. Please refer logs."`,
             );
 
-            sendFailureSlackAlert(testPhase.description, new Error("Integration test failed"), testPhase.slackhook)
+            const msg: string = slackAlerts.get(testPhase.slackhook) || ""
+            slackAlerts.set(testPhase.slackhook, msg + "\n" + testPhase.description)
```
MrSimonEmms, line 358:
```suggestion
            slackAlerts.set(testPhase.slackhook, `${msg}\n${testPhase.description}`)
```
nit: modern JS for the win

MrSimonEmms (approved):
/hold

Put a couple nits in, feel free to change or ignore

.werft/installer-tests.ts
nandajavarma: I am changing this Map by appending items to it. I tend to think `const` as immutable vars. isn't that the case?

MrSimonEmms: But this is JS so normal rules don't apply 🤣 . In JS a `const` represents the variable assignment, not the contents.

## Fail

https://jsfiddle.net/rcs2zkg9/

```js
const a = "hello"
a = "world"
```

## Pass

https://jsfiddle.net/wgeo6npu/

```js
const a = []
a.push('hello')
a.push('world')
```

This would now fail if I did:

```js
a = []
```

But not if I cleared the original array

```js
a.splice(0, a.length)
```

PR #12215, opened 2022-08-19
ws-manager: Replace initializerMapLock with sync.Map

utam0k:
## Description

There were several workspaces content-initializer was taking a very long time to start. [^log1]
As a result, it seemed that stop requests sometimes came before the content-initializer.[^log2]
If content-initializer runs on ws-daemon after the workspace pod has disappeared, it should not work well because there should be no workspace directory. The lock on the initializerMap was common in ws-manager. This could cause longer locks and slower content-initializer startup. This has been improved.

[^log1]: [internal log1](https://console.cloud.google.com/logs/query;query=jsonPayload.instanceId%3D%225663503e-a764-4d88-8657-0c7659594ee9%22;timeRange=2022-08-17T22:11:00.000Z%2F2022-08-18T03:49:00.000Z;summaryFields=jsonPayload%252Flocation:false:32:beginning;cursorTimestamp=2022-08-18T03:27:45Z?project=workspace-clusters)
[^log2]: [internal log2](https://console.cloud.google.com/logs/query;query=jsonPayload.instanceId%3D%2273b2749f-cf2d-4fca-a0b8-433ec126b273%22;timeRange=2022-08-18T03:04:00.000Z%2F2022-08-18T04:11:00.000Z;summaryFields=jsonPayload%252Flocation:false:32:beginning;cursorTimestamp=2022-08-18T03:58:13Z?project=workspace-clusters)

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/11852

## How to test

Do you know good way?

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/ws-manager/pkg/manager/monitor.go
+++ b/components/ws-manager/pkg/manager/monitor.go
@@ -818,7 +818,13 @@ func (m *Monitor) initializeWorkspaceContent(ctx context.Context, pod *corev1.Po
 	// The function below deliniates the initializer lock. It's just there so that we can
 	// defer the unlock call, thus making sure we actually call it.
 	err = func() error {
-		_, alreadyInitializing := m.initializerMap.Load(pod.Name)
+		_, alreadyInitializing := m.initializerMap.LoadOrStore(pod.Name, struct{}{})
+		defer func() {
+			if err != nil {
+				m.initializerMap.Delete(pod.Name)
+			}
+		}()
+
 		if alreadyInitializing {
 			return nil
 		}
@@ -872,9 +878,6 @@ func (m *Monitor) initializeWorkspaceContent(ctx context.Context, pod *corev1.Po
 		if err != nil {
 			return err
 		}
-
-		// mark that we're already initialising this workspace
-		m.initializerMap.Store(pod.Name, struct{}{})
 
 		return nil
 	}()
--- a/components/ws-manager/pkg/manager/monitor.go
+++ b/components/ws-manager/pkg/manager/monitor.go
@@ -77,10 +77,8 @@ type Monitor struct {
 	probeMap     map[string]context.CancelFunc
 	probeMapLock sync.Mutex
 
-	initializerMap     map[string]struct{}
-	initializerMapLock sync.Mutex
-
-	finalizerMap sync.Map
+	initializerMap sync.Map
+	finalizerMap   sync.Map
 
 	act actingManager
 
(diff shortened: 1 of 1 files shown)


components/ws-manager/pkg/manager/monitor.go
```diff
@@ -827,10 +818,7 @@ func (m *Monitor) initializeWorkspaceContent(ctx context.Context, pod *corev1.Po
 	// The function below deliniates the initializer lock. It's just there so that we can
 	// defer the unlock call, thus making sure we actually call it.
 	err = func() error {
-		m.initializerMapLock.Lock()
-		defer m.initializerMapLock.Unlock()
```
sagor999, line 831: @utam0k I am a bit concerned with this change. It changes behavior of this function a bit.
Previously we would lock the map for the duration of this whole function, and unlock only at the end. 
Now we potentially can have two parallel executions of initializeWorkspaceContent happening at the same time if they are both called one after another one. 

I suggest to change this `_, alreadyInitializing := m.initializerMap.Load(pod.Name)` into a call that would Load and Store at the same time? (not sure if there is a function for this already).
Basically we want the behavior of: 
1. Check if `initializerMap` already has that pod name in it. if yes, bail out.
2. If not, immediately store value in there, so that if another `initializeWorkspaceContent` call is made, it would not attempt to initialize content as well. 
3. If we fail during initialize, delete the value from `initializerMap`, so that another call to `initializeWorkspaceContent` can attempt to initialize again.

sagor999 (changes requested):

components/ws-manager/pkg/manager/monitor.go
utam0k: Good point. I was a bit concerned with the same thing during writing codes. I updated this code; please review it again.
https://github.com/gitpod-io/gitpod/pull/12215/commits/db92fe7628b5033055c0214da6b9dbf1cfa7e547

components/ws-manager/pkg/manager/monitor.go
```diff
@@ -716,9 +709,7 @@ func (m *Monitor) waitForWorkspaceReady(ctx context.Context, pod *corev1.Pod) (e
 	if err != nil {
 		return xerrors.Errorf("cannot wait for workspace to initialize: %w", err)
 	}
-	m.initializerMapLock.Lock()
-	delete(m.initializerMap, pod.Name)
-	m.initializerMapLock.Unlock()
+	m.clearInitializerFromMap(pod.Name)
```
jenting, line 712: We could consider remove the function `clearInitializerFromMap`, and use the `m.initializerMap.Delete(pod.Name)` instead.

utam0k: We can't do that. This is because there is not always an initializerMap field.

sagor999 (approved):

components/ws-manager/pkg/manager/monitor.go
utam0k: @jenting Good point!!! Thanks. I created a PR to address your review.
https://github.com/gitpod-io/gitpod/pull/12286

Issue #10257, opened 2022-05-25
Kots admin UI bug - in-cluster S3 fields don't go away


lucasvaltl:
### Bug description

In the KOTS installation Configuration section, under Container registry, you can get into a state where fields are being shown that should be hidden.

### Steps to reproduce

Check the in-cluster option, then select the S3 alternative. Uncheck the in-cluster option - the S3 specific fields stay there instead of going away.

https://www.loom.com/share/1cbe5fb5122b499797b41aeefbba8290

### Expected behavior
 The S3 specific fields should go away when unselecting In-cluster container registry.

Pothulapati:
Verified that this still happens, and raised https://github.com/gitpod-io/gitpod/pull/12218

PR #12218, opened 2022-08-19
[kots] show `registry_s3` options only when `incluster` is enabled

Pothulapati:
## Description

This PR updates the `reg_incluster_storage_s3*` options to
only be enabled both when `incluster` is enabled, and `s3` is
selected, instead of only the latter being used currently.

https://www.loom.com/share/2278f177783d4b9986fe3a2f2b316537

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/10257

## How to test

Follow steps in https://www.loom.com/share/1cbe5fb5122b499797b41aeefbba8290, and
see that it does not happen

## Release Notes

```release-note
[kots] show `registry_s3` options only when `incluster` is enabled
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/kots/manifests/kots-config.yaml
+++ b/install/kots/manifests/kots-config.yaml
@@ -49,7 +49,7 @@ spec:
           title: Storage region
           type: text
           required: true
-          when: '{{repl (ConfigOptionEquals "reg_incluster_storage" "s3") }}'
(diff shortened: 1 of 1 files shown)


MrSimonEmms (approved):

MrSimonEmms:
Good catch on this one

Issue #12107, opened 2022-08-12
VSCode in browser fails to load sometimes


mbrevoort:
### Bug description

When starting a new workspace, VSCode fails to load fully in browser sometimes. A browser refresh usually fixes the problem. I have not been able to reproduce this consistently but it typically happens the first time I load a gitpod on a new device.

See screenshot:
<img width="1249" alt="image" src="https://user-images.githubusercontent.com/30455/184448747-1f4345c1-90f5-4c7d-bcf2-0005e590f502.png">

### Steps to reproduce

Start a workspace in the browser.

### Workspace affected

pbpdnbu5hqm.ws-us60.gitpod.io

### Expected behavior

VSCode should open and be functional

### Example repository

https://github.com/gitpod-io/gitpod

foobar421:
Could this be caused by the recent load-balancer issues? I.e. some of the requests haven't been served properly?

metcalfc:
I hit this as well I believe. I wasn't able to reload to fix it. I had to close the tab completely and reopen the workspace. Happy to get more details the next time I see it if there is something I can grab.

loujaybee:
We've added some analytics and observability for this error, and we can see the error. However, we don't currently have a clear view of how many users this is affecting right now. The next step is to get a clearer picture on the % of users that this is affecting right now. 

@filiptronicek is going to take a closer look.

loujaybee:
We've instrumented this one with data, we'll drop this one from the current iteration and pick up again for investigation in another week or two.

PR #12222, opened 2022-08-19
Capture errors in supervisor frontend

filiptronicek:
## Description

This PR adds error counting of failed VS Code Web resources to track down https://github.com/gitpod-io/gitpod/issues/12107. We investigated one such case at level of GCP Load Balancer and found out that resources are successfully returned by blobserve. [[1](https://gitpod.slack.com/archives/C01KGM9BH54/p1660889332074709?thread_ts=1660861342.562669&cid=C01KGM9BH54)] Something goes wrong in the browser.

https://github.com/gitpod-io/openvscode-server/pull/417 is a PR to instrument critical resources with error handler.

## Related Issue(s)

Related to https://github.com/gitpod-io/gitpod/issues/12107

## How to test

- Open a preview environment workspace
- Cause a load error by for example blocking `workbench.web.main.css` as a source
- Check if error count increased (with label)
- Exec `throw new Error('1')` in browser devTool terminal
- Check if error count increased

### How to monitor Prometheus data

- Start a new workspace with this PR
- Run `./dev/preview/portforward-monitoring-satellite.sh -c harvester`
- Open in Browser

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/ide-metrics/pkg/server/server.go
+++ b/components/ide-metrics/pkg/server/server.go
@@ -236,8 +236,6 @@ func (s *IDEMetricsServer) Start() error {
 			return true
 		},
 		AllowedHeaders: []string{"*"},
-		// Enable Debugging for testing, consider disabling in production
-		Debug: true,
 	})
 
 	routes.Handle("/metrics-api/", http.StripPrefix("/metrics-api", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
@@ -247,7 +245,8 @@ func (s *IDEMetricsServer) Start() error {
 			websocket.IsWebSocketUpgrade(r) {
 			grpcWebServer.ServeHTTP(w, r)
 		} else {
-			restMux.ServeHTTP(w, r)
+			x := c.Handler(restMux)
+			x.ServeHTTP(w, r)
 		}
 	})))
 	go http.Serve(httpMux, routes)
--- a/components/supervisor/frontend/src/ide/ide-metrics-service-client.ts
+++ b/components/supervisor/frontend/src/ide/ide-metrics-service-client.ts
@@ -4,31 +4,40 @@
  * See License-AGPL.txt in the project root for license information.
  */
 
-import { MetricsServicePromiseClient } from '@gitpod/ide-metrics-api-grpcweb/lib/idemetrics_grpc_web_pb'
-import { AddCounterRequest, AddCounterResponse } from '@gitpod/ide-metrics-api-grpcweb/lib/idemetrics_pb'
 import { serverUrl } from '../shared/urls';
-
-const client = new MetricsServicePromiseClient(serverUrl.asIDEMetrics().toString())
 
 export enum MetricsName {
     SupervisorFrontendClientTotal = "gitpod_supervisor_frontend_client_total",
     SupervisorFrontendErrorTotal = "gitpod_supervisor_frontend_error_total"
 }
 
+const MetricsUrl = serverUrl.asIDEMetrics().toString();
+
+interface AddCounterParam {
+    value?: number;
+    labels?: Record<string, string>;
+}
+
 export class IDEMetricsServiceClient {
 
-    static async addCounter(metricsName: string, labels?: Map<string, string>, value?: number) : Promise<AddCounterResponse> {
-        const req = new AddCounterRequest()
-        req.setName(metricsName)
-        if (value) {
-            req.setValue(value)
-        }
-        if (labels) {
-            const m = req.getLabelsMap()
-            for (const [name, value] of labels) {
-                m.set(name, value)
+    static async addCounter(metricsName: MetricsName, labels?: Record<string, string>, value?: number) : Promise<boolean> {
+        const url = `${MetricsUrl}/metrics/counter/add/${metricsName}`
+        const params: AddCounterParam = { value, labels }
+        try {
+            const response = await fetch(url, {
+                method: "POST",
+                body: JSON.stringify(params),
+                credentials: "omit",
+            })
+            if (!response.ok) {
+                const data = await response.json() // { code: number; message: string; }
+                console.error(`Cannot report metrics with addCounter: ${response.status} ${response.statusText}`, data)
+                return false
             }
+            return true
+        } catch (err) {
+            console.error("Cannot report metrics with addCounter, error:", err)
+            return false
         }
-        return client.addCounter(req)
     }
 }
--- a/components/supervisor/frontend/src/index.ts
+++ b/components/supervisor/frontend/src/index.ts
@@ -10,9 +10,29 @@
  */
 import { IDEMetricsServiceClient, MetricsName } from "./ide/ide-metrics-service-client";
 IDEMetricsServiceClient.addCounter(MetricsName.SupervisorFrontendClientTotal).catch(() => {})
-window.addEventListener('error', () => {
-    IDEMetricsServiceClient.addCounter(MetricsName.SupervisorFrontendErrorTotal).catch(() => {})
-})
+
+//#region supervisor frontend error capture
+function isElement(obj: any): obj is Element {
+    return typeof obj.getAttribute === 'function'
+}
+
+window.addEventListener('error', (event) => {
+    const labels: Record<string, string> = {};
+    if (isElement(event.target)) {
+        // We take a look at what is the resource that was attempted to load;
+        const resourceSource = event.target.getAttribute('src') || event.target.getAttribute('href')
+        // If the event has a `target`, it means that it wasn't a script error
+        if (resourceSource) {
+            if (resourceSource.match(new RegExp(/\/build\/ide\/code:.+\/__files__\//g))) {
+                labels['resource'] = 'vscode-web-workbench';
+            }
+            labels['error'] = 'LoadError';
+        }
+    }
+    IDEMetricsServiceClient.addCounter(MetricsName.SupervisorFrontendErrorTotal, labels).catch(() => {});
+});
+//#endregion
+
 require('../src/shared/index.css');
 
 import { createGitpodService, WorkspaceInstancePhase } from "@gitpod/gitpod-protocol";
--- a/install/installer/cmd/testdata/render/aws-setup/output.golden
+++ b/install/installer/cmd/testdata/render/aws-setup/output.golden
@@ -3439,7 +3439,23 @@ data:
           {
             "name": "gitpod_supervisor_frontend_error_total",
             "help": "Total count of supervisor frontend client errors",
-            "labels": null
+            "labels": [
+              {
+                "name": "resource",
+                "allowValues": [
+                  "vscode-web-workbench"
+                ],
+                "defaultValue": "unknown"
+              },
+              {
+                "name": "error",
+                "allowValues": [
+                  "LoadError",
+                  "Unknown"
+                ],
+                "defaultValue": "Unknown"
+              }
+            ]
           },
           {
             "name": "gitpod_supervisor_frontend_client_total",
@@ -7211,7 +7227,7 @@ spec:
   template:
     metadata:
       annotations:
-        gitpod.io/checksum_config: 429b9479c206dc66164883793af80586bca4017bb59e108fa7501a530d24ef57
+        gitpod.io/checksum_config: 3e35ede3647b985d307718989309d2794fdcb8691dd27fb35f538496fb209d79
       creationTimestamp: null
       labels:
         app: gitpod
--- a/install/installer/pkg/components/ide-metrics/configmap.go
+++ b/install/installer/pkg/components/ide-metrics/configmap.go
@@ -20,6 +20,23 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		{
 			Name: "gitpod_supervisor_frontend_error_total",
 			Help: "Total count of supervisor frontend client errors",
+			Labels: []config.LabelAllowList{
+				{
+					Name: "resource",
+					AllowValues: []string{
+						"vscode-web-workbench",
+					},
+					DefaultValue: "unknown",
+				},
+				{
+					Name: "error",
+					AllowValues: []string{
+						"LoadError", // js script style of errors
+						"Unknown",
+					},
+					DefaultValue: "Unknown",
+				},
+			},
 		},
 		{
 			Name: "gitpod_supervisor_frontend_client_total",
--- a/WORKSPACE.yaml
+++ b/WORKSPACE.yaml
@@ -7,7 +7,7 @@ defaultArgs:
   jbMarketplacePublishTrigger: "false"
   publishToJBMarketplace: true
   localAppVersion: unknown
-  codeCommit: 663d7021754843f9f3532c556963e31d04fa5231
+  codeCommit: 422fc166adbf15fcc61b361b6e4b2cfd44fc7ecb
   codeQuality: stable
   intellijDownloadUrl: "https://download.jetbrains.com/idea/ideaIU-2022.2.1.tar.gz"
   golandDownloadUrl: "https://download.jetbrains.com/go/goland-2022.2.2.tar.gz"
--- a/components/supervisor/frontend/package.json
+++ b/components/supervisor/frontend/package.json
@@ -5,8 +5,7 @@
   "version": "0.0.0",
   "dependencies": {
     "@gitpod/gitpod-protocol": "0.1.5",
-    "@gitpod/supervisor-api-grpc": "0.1.5",
-    "@gitpod/ide-metrics-api-grpcweb": "0.0.1"
+    "@gitpod/supervisor-api-grpc": "0.1.5"
   },
   "devDependencies": {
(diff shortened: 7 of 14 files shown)


components/supervisor/frontend/src/index.ts
```diff
@@ -10,9 +10,37 @@
...
+window.addEventListener('error', (event) => {
+
+    const flags = new Map();
+
+    //@ts-ignore We take a look at what is the resource that was attempted to load;
+    const resourceSource = event?.target?.getAttribute('src') || event?.target?.getAttribute('href');
+
+    // If the event has a `target`, it means that it wasn't a script error
+    if (resourceSource) {
+        flags.set('resource', 'vscode-web-workbench');
```
akosyakov, line 29: it is only the case if resourceSource is matching to `https://ide.gitpod.io/blobserve/eu.gcr.io/gitpod-core-dev/build/ide/code:commit`, otherwise resource should be `Unknown`

components/supervisor/frontend/src/index.ts
```diff
@@ -10,9 +10,37 @@
...
+    } else {
+        flags.set('error', 'Unknown');
+    }
+
+    IDEMetricsServiceClient.addCounter(MetricsName.SupervisorFrontendErrorTotal, flags).catch(() => {});
+});
+
+window.addEventListener('load', async () => {
+    for (const error of window.gatheredErrors) {
+        await IDEMetricsServiceClient.addCounter(MetricsName.SupervisorFrontendErrorTotal, error).catch(() => {});
```
akosyakov, line 40: yon don't need to await

akosyakov:
@filiptronicek you need to change installer configuration to add `resource` and `error` labels here: https://github.com/gitpod-io/gitpod/blob/b4edca5b685f17f062d802ac6acd821e0998a884/install/installer/pkg/components/ide-metrics/configmap.go#L20-L23

components/supervisor/frontend/src/index.ts
akosyakov: please not use it as i give it to you, but test, the url can be relative so maybe you need to match for substring like `ide/code:commit-`

It should work also in SH, but we can test it separately.

components/supervisor/frontend/src/index.ts
```diff
@@ -10,9 +10,37 @@
  */
 import { IDEMetricsServiceClient, MetricsName } from "./ide/ide-metrics-service-client";
 IDEMetricsServiceClient.addCounter(MetricsName.SupervisorFrontendClientTotal).catch(() => {})
-window.addEventListener('error', () => {
-    IDEMetricsServiceClient.addCounter(MetricsName.SupervisorFrontendErrorTotal).catch(() => {})
-})
+
+declare global {
+    interface Window {
+        gatheredErrors: Map<string, string>[];
```
akosyakov, line 16: use gitpod in name of global variables, i.e. `gitpodIDEErrors`

akosyakov:
@filiptronicek there is a PR in openvscode server we need to review it as well?

components/supervisor/frontend/src/index.ts
filiptronicek: Thanks for the reminder :pray:

components/supervisor/frontend/src/index.ts
filiptronicek: Added a regular expression to match this.

components/supervisor/frontend/src/index.ts
```diff
@@ -10,9 +10,37 @@
...
+
+    //@ts-ignore We take a look at what is the resource that was attempted to load;
+    const resourceSource = event?.target?.getAttribute('src') || event?.target?.getAttribute('href');
+
+    // If the event has a `target`, it means that it wasn't a script error
+    if (resourceSource && resourceSource.match(new RegExp(/\/build\/ide\/code:.{1,100}\/__files__\//g))) {
+        flags.set('resource', 'vscode-web-workbench');
+        flags.set('error', 'LoadError');
+    } else {
+        flags.set('error', 'Unknown');
```
akosyakov, line 32: what if this is target but it does not match to vscode-web-workbench? Should we report it as `LoadError` or `Unknown`?

components/supervisor/frontend/src/index.ts
```diff
@@ -10,9 +10,37 @@
  */
 import { IDEMetricsServiceClient, MetricsName } from "./ide/ide-metrics-service-client";
 IDEMetricsServiceClient.addCounter(MetricsName.SupervisorFrontendClientTotal).catch(() => {})
-window.addEventListener('error', () => {
-    IDEMetricsServiceClient.addCounter(MetricsName.SupervisorFrontendErrorTotal).catch(() => {})
-})
+
+declare global {
+    interface Window {
+        gitpodIDEErrors: Map<string, string>[];
```
akosyakov, line 16: we will need to make sure that both stable and latest have these changes at the time of supervisor deployment or it will break

components/supervisor/frontend/src/index.ts
```diff
@@ -10,9 +10,37 @@
...
+    }
+
+    IDEMetricsServiceClient.addCounter(MetricsName.SupervisorFrontendErrorTotal, flags).catch(() => {});
+});
+
+window.addEventListener('load', async () => {
+    for (const error of window.gitpodIDEErrors) {
+        IDEMetricsServiceClient.addCounter(MetricsName.SupervisorFrontendErrorTotal, error).catch(() => {});
+    }
+});
```
akosyakov, line 42: please add `{ once: true }`, just to be sure

filiptronicek: ✅

components/supervisor/frontend/src/index.ts
akosyakov: hm, i don't see any regular expression 🤔

components/supervisor/frontend/src/index.ts
filiptronicek: I think it's best if we do LoadError with resource of unknown, so that we know a bit more about it

components/supervisor/frontend/src/index.ts
filiptronicek: Git did something weird, going to re-push this change.

components/supervisor/frontend/src/index.ts
filiptronicek: Looks like this change didn't get pushed as well, huh. Will re-add.

filiptronicek: There we go.

components/supervisor/frontend/src/index.ts
filiptronicek: Donezos

components/supervisor/frontend/src/index.ts
```diff
@@ -10,9 +10,46 @@
...
+
+    const flags = new Map();
+
+    //@ts-ignore We take a look at what is the resource that was attempted to load;
+    const resourceSource = event?.target?.getAttribute('src') || event?.target?.getAttribute('href');
+
+    // If the event has a `target`, it means that it wasn't a script error
+    if (resourceSource) {
+        let resource = 'unknown';
+        if (resourceSource.match(new RegExp(/\/build\/ide\/code:.{1,100}\/__files__\//g))) {
```
akosyakov, line 30: Do we really need to limit to 100? .+ won’t work? Why 100?

filiptronicek: 100 seemed just like a number that could cover this whole part, since it can vary in length. I acknowledge that it's not wise to get an abitrary number in here, so thanks for teaching me about the `+` operator in regex!

filiptronicek: Here's the regex (with simple tests): https://regexr.com/6s8pf

filiptronicek: And here's the fixing commit: [`6184f1b` (#12222)](https://github.com/gitpod-io/gitpod/pull/12222/commits/6184f1b06b3f9636af9e0ef8c2b6e66a193198f9).

filiptronicek:
It seems that if I block the CSS from responding the error gets captured and if I don't there are no errors to report. I have to check next whether they actually get reported. 
image

filiptronicek:
Updated the configmap with the labels in [`3cfa878` (#12222)](https://github.com/gitpod-io/gitpod/pull/12222/commits/3cfa8787cf5020edc6c592641de10cb5f2c337c0).

akosyakov:
@iQQBot is going to create a PR which allow to post metrics via usual fetch directly to IDE proxy.
@mustard-mh Coudl you please:
- change https://github.com/gitpod-io/openvscode-server/pull/417/files to use fetch instead of saving on global variable
- we still need this PR because some resources can be loaded async later

install/installer/pkg/components/ide-metrics/configmap.go
```diff
@@ -24,6 +24,22 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		{
 			Name: "gitpod_supervisor_frontend_client_total",
 			Help: "Total count of supervisor frontend client",
+			Labels: []config.LabelAllowList{
```
iQQBot, line 27: this labels cannot dynamically change, if you want to change please consider add another metrics

components/supervisor/frontend/src/index.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2020 Gitpod GmbH. All rights reserved.
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
```
iQQBot, line 2: no need change this line

iQQBot (changes requested):
prometheus don't allow dynamic change labels

akosyakov:
> prometheus don't allow dynamic change labels

@iQQBot I'm not sure I understand it. We for sure would like to add new labels as we learn about new error kinds we don't want to create a new metric each time. Can we restart metrics component to pick up new labels? I think it is rather deployment issue or metrics component, but not this PR.

akosyakov:
> This is a bit complicated, for example, if you add a label resource for now, then all subsequent clients must fill in this label when reporting this metrics, otherwise prometheus will not accept the request, unless we design a default value

default value like `unknown` sounds good to me for old clients

akosyakov:
> And this can also lead to a series of problems when merging queries, For example, if we add the label resource and restart the metrics-server, there will be two sets of data on prometheus, and when you do sum or something like that, there will be some strange problems, such as their meaning changing, especially when changing the name

Is not it the same with other components like  if someone add a new error code in server for http request?

akosyakov:
> Not very clear, can you elaborate?

It is me not understanding problems which you are explaining. Are they persistent though or after a week they won’t be matter?

i was wondering if someone adds another label to existing metrics in other components like server and SSH gateway will be there the same problems? If not why is it different for ide metrics?

components/supervisor/frontend/src/index.ts
```diff
@@ -10,9 +10,38 @@
...
+
+//#region supervisor frontend error capture
+declare global {
+    interface Window {
+        gitpodMetricsSupervisorFrontendErrorTotal(labels?: Map<string, string>): Promise<void>;
+    }
+}
+
+// Add this for other clients, i.e. openvscode-server
+window.gitpodMetricsSupervisorFrontendErrorTotal = async (labels?: Map<string, string>) => {
```
akosyakov, line 22: we don't need it and should use browser API in VS Code Workbench directly, this file won't be necessary loaded at all

mustard-mh:
Changed in both openvscode-srever PR and this PR, @akosyakov @iQQBot could you have a look?

components/ide-metrics/pkg/server/server.go
```diff
@@ -232,12 +232,11 @@ func (s *IDEMetricsServer) Start() error {
 	}))
 
 	c := cors.New(cors.Options{
+		AllowCredentials: true,
```
akosyakov, line 235: should it be deployed before?

components/supervisor/frontend/src/ide/ide-metrics-service-client.ts
```diff
@@ -4,31 +4,47 @@
...
-            for (const [name, value] of labels) {
-                m.set(name, value)
+            params.labels = labels
+        }
+        try {
+            const response = await fetch(url, {
+                method: "POST",
+                body: JSON.stringify(params),
+                headers: { "Content-Type": "application/json;" },
+                credentials: "include",
```
akosyakov, line 37: why do we need it? 🤔

components/supervisor/frontend/src/index.ts
```diff
@@ -10,9 +10,29 @@
...
 import { IDEMetricsServiceClient, MetricsName } from "./ide/ide-metrics-service-client";
 IDEMetricsServiceClient.addCounter(MetricsName.SupervisorFrontendClientTotal).catch(() => {})
-window.addEventListener('error', () => {
-    IDEMetricsServiceClient.addCounter(MetricsName.SupervisorFrontendErrorTotal).catch(() => {})
-})
+
+//#region supervisor frontend error capture
+window.addEventListener('error', (event) => {
+
+    const flags = new Map();
```
akosyakov, line 17: (nit) labels

install/installer/cmd/testdata/render/aws-setup/output.golden
```diff
@@ -3444,7 +3444,23 @@ data:
...
+                ],
+                "defaultValue": "unknown"
+              },
+              {
+                "name": "error",
+                "allowValues": [
+                  "LoadError",
+                  "Unknown"
+                ],
+                "defaultValue": "unknown"
```
akosyakov, line 3461: Unknown

install/installer/pkg/components/ide-metrics/configmap.go
```diff
@@ -24,6 +24,23 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
...
+					},
+					DefaultValue: "unknown",
+				},
+				{
+					Name: "error",
+					AllowValues: []string{
+						"LoadError", // js script style of errors
+						"Unknown",
+					},
+					DefaultValue: "unknown",
```
akosyakov, line 41: Unknown

components/supervisor/frontend/src/ide/ide-metrics-service-client.ts
mustard-mh: I'll try to remove this line in `ide-metrics` and `credentials: "include",` in fetch, to see if it works

```
	c := cors.New(cors.Options{
		AllowCredentials: true,
```

components/ide-metrics/pkg/server/server.go
mustard-mh: both `ide-metrics` and `supervisor` is belong to us, they will deploy together.

fetch in openvscode-server is wrapped by catch, will only affect red(failed) requests in DevTool Network section and console

mustard-mh:
I'll test and post images here after werft build succeed

mustard-mh:
Force push to use Record<string, string> instead of Map (tested and found labels not worked (not sent) in preview env, because Map stringify not works)

filiptronicek:
> Force push to use Record<string, string> instead of Map (tested and found labels not worked (not sent) in preview env, because Map stringify not works)

@mustard-mh have you investigated using `Map`s and then just converting to JSON with the following? I personally do not mind using objects, but this could be an alternative (this should work every time if the map keys are all strings, which in our case are [[1](https://stackoverflow.com/a/70461480/10199319)]). 

```ts
JSON.stringify(Object.fromEntries(labels))
```

mustard-mh:
> > Force push to use Record<string, string> instead of Map (tested and found labels not worked (not sent) in preview env, because Map stringify not works)
> 
> @mustard-mh have you investigated using `Map`s and then just converting to JSON with the following? I personally do not mind using objects, but this could be an alternative (this should work every time if the map keys are all strings, which in our case are [[1](https://stackoverflow.com/a/70461480/10199319)]).
> 
> ```ts
> JSON.stringify(Object.fromEntries(labels))
> ```

💡 Good to know this. Record is enough and easier for us

mustard-mh:
Force push to

- Remove content-type in fetch function to reduce an OPTION request (same with openvscode-server's PR)
- Update changes of configmap to correct position

filiptronicek (commented):
Codewise LGTM. Tested and it seems like the load errors get reported correctly (see screenshot). I am not sure whether the script errors get reported as well though, because it seems like from the devtools console they do not get caught - maybe I just looked in the wrong place. 

image

iQQBot (approved):
Tested and it works well 🎉 

image

image

/hold

in case somebody want to have a look again, feel free to unhold it

akosyakov:
Most important that we get errors on load of VS Code resources. If we can track it now. Let's bring it in. We need it in both insider and stable. It would be ideally to have a label as well whether error is from stable or latest, but for now let's skip it.

JS errors is interesting but I think it would be more important to report them to GCP first that we can categorize and do something about them. It is for later as well.

mustard-mh:
I'm going to unhold it

PR #12227, opened 2022-08-19
[kots]: give installer same node affinity as ws-daemon

MrSimonEmms:
## Description

As we're auto-detecting the containerd location in the installer job, we should be on a "workspace" node so this ensures that the installer job is run on the correct node.

This is an edge case and not one we've yet been caught by, but it's possible that the nodes are configured differently (as with SaaS) so it's something I think we should do.

## How to test

Install via KOTS.

## Release Notes

```release-note
[kots]: give installer same node affinity as ws-daemon
```

## Werft options:

- [ ] /werft with-preview


nandajavarma (approved):
LGTM!

PR #12236, opened 2022-08-19
Use `toLocaleString` with usage of credits

filiptronicek:
## Description
Make bigger amounts of used credits more easily readable by using the browsers locale to format the number. This also solved an issue where trailing zeros would appear if the number was for example exactly `.2`. This may be intentional though and if so, it is easy to fix.

This change does not change how values are rounded, both this method and the previous one do it in the same way.

## Related Issue(s)
None I am aware of.

## How to test

TBD. Easier to test in prod 😆 

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/dashboard/src/teams/TeamUsage.tsx
+++ b/components/dashboard/src/teams/TeamUsage.tsx
@@ -114,7 +114,7 @@ function TeamUsage() {
     const calculateTotalUsage = () => {
         let totalCredits = 0;
         billedUsage.forEach((session) => (totalCredits += session.credits));
-        return totalCredits.toLocaleString(undefined, { maximumFractionDigits: 2 });
+        return totalCredits.toLocaleString(undefined, { minimumFractionDigits: 2, maximumFractionDigits: 2 });
     };
 
     const handleMonthClick = (start: any, end: any) => {
--- a/components/dashboard/src/teams/TeamUsage.tsx
+++ b/components/dashboard/src/teams/TeamUsage.tsx
@@ -114,7 +114,7 @@ function TeamUsage() {
     const calculateTotalUsage = () => {
         let totalCredits = 0;
         billedUsage.forEach((session) => (totalCredits += session.credits));
-        return totalCredits.toFixed(2);
+        return totalCredits.toLocaleString(undefined, { maximumFractionDigits: 2 });
     };
 
     const handleMonthClick = (start: any, end: any) => {


components/dashboard/src/teams/TeamUsage.tsx
```diff
@@ -114,7 +114,7 @@ function TeamUsage() {
     const calculateTotalUsage = () => {
         let totalCredits = 0;
         billedUsage.forEach((session) => (totalCredits += session.credits));
-        return totalCredits.toFixed(2);
+        return totalCredits.toLocaleString(undefined, { maximumFractionDigits: 2 });
```
geropl, line 117: I think this was meant to be fixed, but was not part of those discussions.

Deferring review to @gtsiolis :basketball:

gtsiolis:
Spinning up a preview environment above[[1](https://github.com/gitpod-io/gitpod/pull/12236#issuecomment-1221948022)] and will try to test this there.

For visual changes like this, it would be really helpful in the future include :a: some BEFORE / AFTER screenshots and :b: clearer steps to test and reproduce inside the preview environment. 💡 

If testing this in a preview environment is impossible for some reason, listing the visual changes clearly with text could also suffice. ✔️ 

Cc @filiptronicek @geropl

filiptronicek:
Thanks @gtsiolis for the feedback 🙏!

I 💯 agree about your points on testing and screenshots, I didn't include them since I didn't find a proper way to test these changes myself (the billing I suppose is a bit harder to set up).

gtsiolis:
> I didn't include them since I didn't find a proper way to test these changes myself

@filiptronicek It's ok! I've also found out about this by accident while reviewing PRs relevant to the billing feature. We could certainly be more intentional and better at documenting such things. We already have a [**support matrix**](https://www.notion.so/gitpod/Preview-Environments-6debd359591b43688b52f76329d04010#44f8b4ff77984e01828d2e260adb51c2) in **Product Engineering / Preview Environments**, but I don't think this clearly lists the available commands for `/werft`.

gtsiolis (approved):
From what I understand by reading the PR description and after testing in the preview environment, this does not introduce any visual changes. @filiptronicek Let me know if this is not the case.

| Production | Preview Environment |
|-|-|
| credtis-production | credtis-preview |

components/dashboard/src/teams/TeamUsage.tsx
geropl, line 117:
```suggestion
        return totalCredits.toLocaleString(undefined, { minimumFractionDigits: 2, maximumFractionDigits: 2 });
```
This is what I wanted to get at: Money always has 2 digits (in our money system at least, which most humans are comfortable with).

Example code:
```
const n = 7777.52;
const n2 = 7777.5;

console.log(n.toLocaleString(undefined, { minimumFractionDigits: 2, maximumFractionDigits: 2 }));
console.log(n2.toLocaleString(undefined, { maximumFractionDigits: 2 }));
console.log(n2.toLocaleString(undefined, { minimumFractionDigits: 2, maximumFractionDigits: 2 }));
```
Output:
```
7,777.52
7,777.5
7,777.50
```

geropl: /hold because I'd love to see this in as well @filiptronicek

geropl (approved):
:heavy_check_mark: to unblock,

hold bc would love to see [this suggestion](https://github.com/gitpod-io/gitpod/pull/12236/files#r952248269) in.

components/dashboard/src/teams/TeamUsage.tsx
filiptronicek: Agreed, let's do it!

gtsiolis:
Thanks, @filiptronicek @geropl! Removing the hold.

/unhold

PR #12253, opened 2022-08-22
[terraform] Make AWS certificate manager optional

nandajavarma:
## Description

In the `EWS` terraform module we are by default creating `AWS cert manager` entries, which are not essentially needed since by default our reference architecture suggests usage of `cert-manager`. Hence this PR makes the creation of `AWS cert manager` optional using the flag `user_aws_cert_manager`.

As of the terraform module corresponding the reference architecture, in a future PR, I will provide an option to use `aws cert manager` if the user would rather use that instead of `cert-manager` in which case, in the output, we will provide config patch necessary to post process the service annotation to use `ACM`.

## Related Issue(s)

Fixes #

## How to test

Following the [README](https://github.com/gitpod-io/gitpod/tree/main/install/infra/single-cluster/aws#initialize-terraform-backend-and-confirm-the-plan) for the terraform reference architecture, you will now see that the unnecessary `ACM` creation does not happen anymore.

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/infra/modules/eks/dns.tf
+++ b/install/infra/modules/eks/dns.tf
@@ -1,17 +1,77 @@
 resource "aws_route53_zone" "gitpod" {
-  force_destroy = true
-  name = var.domain_name
+  count = local.domain_name_count
 
-  tags = {
-    Environment = "test"
-  }
+  force_destroy = true
+  name          = var.domain_name
+}
+
+resource "aws_iam_policy" "gitpod" {
+  count = local.domain_name_count
+
+  name = "role-${var.cluster_name}"
+
+  # Terraform's "jsonencode" function converts a
+  # Terraform expression result to valid JSON syntax.
+  policy = jsonencode({
+    Version = "2012-10-17",
+    Statement = [
+      {
+        Effect = "Allow",
+        Action = [
+          "route53:ChangeResourceRecordSets"
+        ],
+        Resource = [
+          "arn:aws:route53:::hostedzone/*"
+        ]
+      },
+      {
+        Effect = "Allow",
+        Action = [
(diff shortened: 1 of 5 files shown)


roboquat:

<details>

</details>

MrSimonEmms:
LGTM

/hold in case you want to change to my suggestion

---

It might be worth creating a `locals.tf` file in `modules/eks` to avoid having to write `count = var.domain_name == "" || var.use_aws_cert_manager == false ? 0 : 1` and `count = var.domain_name == "" ? 0 : 1` so many times.

It could look something like this:

```tf
locals {
  cert_manager_enabled : local.domain_name_enabled && var.use_aws_cert_manager == true
  cert_manager_count : local.cert_manager_enabled ? 1 : 0
  domain_name_enabled : var.domain_name != ""
  domain_name_count : local.domain_name_enabled ? 1 : 0
}
```

Then you could change all the `count` references to:

```tf
resource "aws_iam_access_key" "edns" {
  count = local.domain_name_count
}

resource "aws_acm_certificate" "gitpod" {
  count = local.cert_manager_count
}
```

**NB** I wrote this entirely in this comment box so the syntax is probably buggy/wrong, but it should be enough for you to get started. We can always chat if you want more help with this

MrSimonEmms (approved):

nandajavarma:
Great tip, Simon! I have updated the `dns.tf` files with the local vars you suggested. There are other files with similar conditions, I will replace them in another PR with local vars. Merging this after the test here passes https://werft.gitpod-dev.com/job/gitpod-custom-nvn-remove-aws-acm.0

PR #12255, opened 2022-08-22
Remove code cli from path

jeanp413:
## Description

Do not add code cli to path, vscode adds it to the terminals env internally

## How to test

1. Print `env` and check `/ide/bin/remote-cli` is not in the path

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


akosyakov:
What about task terminals? They are not created by VS Code.

jeanp413:
Right, forgot about this :facepalm:. 
But then I would use this as an excuse to ask, does it really make sense for task terminals to be writable? I think it makes more sense for them to be readonly and use the vscode task API instead. And then, it will complement with `gp task stop` [command](https://github.com/gitpod-io/gitpod/pull/12116#issuecomment-1220606290) and we also need to add a `gp task run` and corresponding vscode commands same in jetbrains

akosyakov:
> But then I would use this as an excuse to ask, does it really make sense for task terminals to be writable?

I don't think we can change it now. If there is some new consistent approach to consider I guess we need to start with RFC.

jeanp413:
closing this for now then

Issue #11469, opened 2022-07-19
Epic: Admin Audit Log


geropl:
### Summary

Have a log of all admin activities:
 - blocking users
 - adding/removing blocked repositories, emails, etc.
 - accessing of user's resources with admin privileges (e.g., "download workspace")

### Context

 - RFC: https://www.notion.so/gitpod/Audit-Log-for-Gitpod-Admin-Actions-630e57b049d348d79f9ba4aed2a0c70b
 - potentially overlapping Self-Hosted centric epic: https://github.com/gitpod-io/gitpod/issues/10560
 - related: https://github.com/gitpod-io/gitpod/issues/9456

### Value

This would replace and improve solutions we have for gitpod.io, and make them available to all Self-Hosted installations.

### Acceptance Criteria

When are able to observe an audit log from the Admin UI that is automatically filled by all actions listed above.

### Measurement

 - [ ] Can we replace our existing internal solutions (Slack)?

svenefftinge:
I'd love to have an audit lof for all operations and get rid of all the duplicate `admin*` APIs entirely but instead work with privileges and use the common APIs.

PR #12258, opened 2022-08-22
Add phone verification

### LINKED ISSUE #11339: Add Phone Verification for new Users

svenefftinge:
## Description
This PR adds phone number verification before untrusted users can start a workspace.

A user is considered trusted if
 - Their GitHub account is older than 30 days, or
 - Their GitLab account `confirmed_at` event is older than 30 days, or
 - Their Gitpod account has been created before 2022-08-24 (i.e. existing users)

## Related Issue(s)
Fixes #11339 

See also https://github.com/gitpod-io/ops/pull/4149

## How to test
*Test 1*
 - Create a fresh GitHub account and sign in (use admin to unblock). 
 - You should need to do phone number verification before you can start your first workspace.
 - Once the number is verified you should be able to work as usual.

*Test 2*
 - Use a GitHub account older than 30 days.
 - You should be able to start a workspace without verification.

*Test 1 ++*
 - delete the `lastVerificationTimestamp` in your user's additional Data.
 - if the account had started a workspace before, you should not need to verify
 - if the account doesn't have any workspaces, you should be asked to verify

## Release Notes

```release-note
Added phone number verification in SaaS, to help mitigate abuse.
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/dashboard/src/start/VerifyModal.tsx
+++ b/components/dashboard/src/start/VerifyModal.tsx
@@ -1,0 +1,241 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { useState } from "react";
+import Alert, { AlertType } from "../components/Alert";
+import Modal from "../components/Modal";
+import { getGitpodService } from "../service/service";
+import PhoneInput from "react-intl-tel-input";
+import "react-intl-tel-input/dist/main.css";
+import "./phone-input.css";
+
+interface VerifyModalState {
+    phoneNumber?: string;
+    phoneNumberValid?: boolean;
+    sent?: Date;
+    sending?: boolean;
+    message?: {
+        type: AlertType;
+        text: string;
+    };
+    token?: string;
+    verified?: boolean;
+}
+
+export function VerifyModal() {
+    const [state, setState] = useState<VerifyModalState>({});
+
+    if (!state.sent) {
+        const sendCode = async () => {
+            try {
+                setState({
+                    ...state,
+                    message: undefined,
+                    sending: true,
+                });
+                await getGitpodService().server.sendPhoneNumberVerificationToken(state.phoneNumber || "");
+                setState({
+                    ...state,
+                    sending: false,
+                    sent: new Date(),
+                });
+                return true;
+            } catch (err) {
+                setState({
+                    sent: undefined,
+                    sending: false,
+                    message: {
+                        type: "error",
+                        text: err.toString(),
+                    },
+                });
+                return false;
+            }
+        };
+        return (
+            <Modal
+                onClose={() => {}}
+                closeable={false}
+                onEnter={sendCode}
+                title="User Validation Required"
+                buttons={
+                    <div>
+                        <button className="ml-2" disabled={!state.phoneNumberValid || state.sending} onClick={sendCode}>
+                            Send Code via SMS
+                        </button>
+                    </div>
+                }
+                visible={true}
+            >
+                <Alert type="warning" className="mt-2">
+                    To use Gitpod you'll need to validate your account with your phone number. This is required to
+                    discourage and reduce abuse on Gitpod infrastructure.
+                </Alert>
+                <div className="text-gray-600 dark:text-gray-400 mt-2">
+                    Enter a mobile phone number you would like to use to verify your account.
+                </div>
+                {state.message ? (
+                    <Alert type={state.message.type} className="mt-4 py-3">
+                        {state.message.text}
+                    </Alert>
+                ) : (
+                    <></>
+                )}
+                <div className="mt-4">
+                    <h4>Mobile Phone Number</h4>
+                    {/* HACK: Below we are adding a dummy dom element that is not visible, to reference the classes so they are not removed by purgeCSS. */}
+                    <input type="tel" className="hidden intl-tel-input country-list" />
+                    <PhoneInput
+                        autoFocus={true}
+                        containerClassName={"allow-dropdown w-full intl-tel-input"}
+                        inputClassName={"w-full"}
+                        allowDropdown={true}
+                        defaultCountry={""}
+                        autoHideDialCode={false}
+                        onPhoneNumberChange={(isValid, phoneNumberRaw, countryData) => {
+                            let phoneNumber = phoneNumberRaw;
+                            if (!phoneNumber.startsWith("+") && !phoneNumber.startsWith("00")) {
+                                phoneNumber = "+" + countryData.dialCode + phoneNumber;
+                            }
+                            setState({
+                                ...state,
+                                phoneNumber,
+                                phoneNumberValid: isValid,
+                            });
+                        }}
+                    />
+                </div>
+            </Modal>
+        );
+    } else if (!state.verified) {
+        const isTokenFilled = () => {
+            return state.token && /\d{6}/.test(state.token);
+        };
+        const verifyToken = async () => {
+            try {
+                const verified = await getGitpodService().server.verifyPhoneNumberVerificationToken(
+                    state.phoneNumber!,
+                    state.token!,
+                );
+                if (verified) {
+                    setState({
+                        ...state,
+                        verified: true,
+                        message: undefined,
+                    });
+                } else {
+                    setState({
+                        ...state,
+                        message: {
+                            type: "error",
+                            text: `Invalid verification code.`,
+                        },
+                    });
+                }
+                return verified;
+            } catch (err) {
+                setState({
+                    sent: undefined,
+                    sending: false,
+                    message: {
+                        type: "error",
+                        text: err.toString(),
+                    },
+                });
+                return false;
+            }
+        };
+
+        const reset = () => {
+            setState({
+                ...state,
+                sent: undefined,
+                message: undefined,
+                token: undefined,
+            });
+        };
+        return (
+            <Modal
+                onClose={() => {}}
+                closeable={false}
+                onEnter={verifyToken}
+                title="User Validation Required"
+                buttons={
+                    <div>
+                        <button className="ml-2" disabled={!isTokenFilled()} onClick={verifyToken}>
+                            Validate Account
+                        </button>
+                    </div>
+                }
+                visible={true}
+            >
+                <Alert type="warning" className="mt-2">
+                    To use Gitpod you'll need to validate your account with your phone number. This is required to
+                    discourage and reduce abuse on Gitpod infrastructure.
+                </Alert>
+                <div className="pt-4">
+                    <button className="gp-link" onClick={reset}>
+                        &larr; Use a different phone number
+                    </button>
+                </div>
+                <div className="text-gray-600 dark:text-gray-400 pt-4">
+                    Enter the verification code we sent to {state.phoneNumber}.<br />
+                    Having trouble?{" "}
+                    <a className="gp-link" href="https://www.gitpod.io/contact/support">
+                        Contact support
+                    </a>
+                </div>
+                {state.message ? (
+                    <Alert type={state.message.type} className="mt-4 py-3">
+                        {state.message.text}
+                    </Alert>
+                ) : (
+                    <></>
+                )}
+                <div className="mt-4">
+                    <h4>Verification Code</h4>
+                    <input
+                        autoFocus={true}
+                        className="w-full"
+                        type="text"
+                        placeholder="Enter code sent via SMS"
+                        onChange={(v) => {
+                            setState({
+                                ...state,
+                                token: v.currentTarget.value,
+                            });
+                        }}
+                    />
+                </div>
+            </Modal>
+        );
+    } else {
+        const continueStartWorkspace = () => {
+            window.location.reload();
+            return true;
+        };
+        return (
+            <Modal
+                onClose={continueStartWorkspace}
+                closeable={false}
+                onEnter={continueStartWorkspace}
+                title="User Validation Successful"
+                buttons={
+                    <div>
+                        <button className="ml-2" onClick={continueStartWorkspace}>
+                            Continue
+                        </button>
+                    </div>
+                }
+                visible={true}
+            >
+                <Alert type="success" className="mt-2">
+                    Your account has been successfully verified.
+                </Alert>
+            </Modal>
+        );
+    }
+}
--- a/components/server/ee/src/billing/entitlement-service.ts
+++ b/components/server/ee/src/billing/entitlement-service.ts
@@ -12,6 +12,7 @@ import {
 } from "@gitpod/gitpod-protocol";
 import { log } from "@gitpod/gitpod-protocol/lib/util/logging";
 import { inject, injectable } from "inversify";
+import { VerificationService } from "../../../src/auth/verification-service";
 import { EntitlementService, MayStartWorkspaceResult } from "../../../src/billing/entitlement-service";
 import { Config } from "../../../src/config";
 import { BillingModes } from "./billing-mode";
@@ -31,6 +32,7 @@ export class EntitlementServiceImpl implements EntitlementService {
     @inject(EntitlementServiceChargebee) protected readonly chargebee: EntitlementServiceChargebee;
     @inject(EntitlementServiceLicense) protected readonly license: EntitlementServiceLicense;
     @inject(EntitlementServiceUBP) protected readonly ubp: EntitlementServiceUBP;
+    @inject(VerificationService) protected readonly verificationService: VerificationService;
 
     async mayStartWorkspace(
         user: User,
@@ -38,6 +40,13 @@ export class EntitlementServiceImpl implements EntitlementService {
         runningInstances: Promise<WorkspaceInstance[]>,
     ): Promise<MayStartWorkspaceResult> {
         try {
+            const verification = await this.verificationService.needsVerification(user);
+            if (verification) {
+                return {
+                    mayStart: false,
+                    needsVerification: true,
+                };
+            }
             const billingMode = await this.billingModes.getBillingModeForUser(user, date);
             let result;
             switch (billingMode.mode) {
--- a/components/server/src/auth/rate-limiter.ts
+++ b/components/server/src/auth/rate-limiter.ts
@@ -14,7 +14,7 @@ type GitpodServerMethodType =
     | keyof Omit<GitpodServer, "dispose" | "setClient">
     | typeof accessCodeSyncStorage
     | typeof accessHeadlessLogs;
-type GroupKey = "default" | "startWorkspace" | "createWorkspace";
+type GroupKey = "default" | "startWorkspace" | "createWorkspace" | "phoneVerification";
 type GroupsConfig = {
     [key: string]: {
         points: number;
@@ -46,11 +46,17 @@ function getConfig(config: RateLimiterConfig): RateLimiterConfig {
             points: 3, // 3 workspace creates per user per 10s
             durationsSec: 10,
         },
+        phoneVerification: {
+            points: 10,
+            durationsSec: 10,
+        },
     };
     const defaultFunctions: FunctionsConfig = {
         getLoggedInUser: { group: "default", points: 1 },
         getTerms: { group: "default", points: 1 },
         updateLoggedInUser: { group: "default", points: 1 },
+        sendPhoneNumberVerificationToken: { group: "phoneVerification", points: 1 },
+        verifyPhoneNumberVerificationToken: { group: "phoneVerification", points: 1 },
         getAuthProviders: { group: "default", points: 1 },
         getOwnAuthProviders: { group: "default", points: 1 },
         updateOwnAuthProvider: { group: "default", points: 1 },
--- a/components/server/src/auth/verification-service.ts
+++ b/components/server/src/auth/verification-service.ts
@@ -1,0 +1,76 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { User } from "@gitpod/gitpod-protocol";
+import { log } from "@gitpod/gitpod-protocol/lib/util/logging";
+import { inject, injectable, postConstruct } from "inversify";
+import { Config } from "../config";
+import { Twilio } from "twilio";
+import { ServiceContext } from "twilio/lib/rest/verify/v2/service";
+import { WorkspaceDB } from "@gitpod/gitpod-db/lib";
+import { ConfigCatClientFactory } from "@gitpod/gitpod-protocol/lib/experiments/configcat-server";
+
+@injectable()
+export class VerificationService {
+    @inject(Config) protected config: Config;
+    @inject(WorkspaceDB) protected workspaceDB: WorkspaceDB;
+    @inject(ConfigCatClientFactory) protected readonly configCatClientFactory: ConfigCatClientFactory;
+
+    protected verifyService: ServiceContext;
+
+    @postConstruct()
+    protected initialize(): void {
+        if (this.config.twilioConfig) {
+            const client = new Twilio(this.config.twilioConfig.accountSID, this.config.twilioConfig.authToken);
+            this.verifyService = client.verify.v2.services(this.config.twilioConfig.serviceID);
+        }
+    }
+
+    public async needsVerification(user: User): Promise<boolean> {
+        if (!this.config.twilioConfig) {
+            return false;
+        }
+        if (!!user.lastVerificationTime) {
+            return false;
+        }
+        // we treat existing users (created before we introduced phone vwerification) as verified
+        if (user.creationDate < "2022-08-22") {
+            return false;
+        }
+        const isPhoneVerificationEnabled = await this.configCatClientFactory().getValueAsync(
+            "isPhoneVerificationEnabled",
+            false,
+            {
+                user,
+            },
+        );
+        return isPhoneVerificationEnabled;
+    }
+
+    public markVerified(user: User): User {
+        user.lastVerificationTime = new Date().toISOString();
+        return user;
+    }
+
+    public async sendVerificationToken(phoneNumber: string): Promise<void> {
+        if (!this.verifyService) {
+            throw new Error("No verification service configured.");
+        }
+        const verification = await this.verifyService.verifications.create({ to: phoneNumber, channel: "sms" });
+        log.info("Verification code sent", { phoneNumber, status: verification.status });
+    }
+
+    public async verifyVerificationToken(phoneNumber: string, oneTimePassword: string): Promise<boolean> {
+        if (!this.verifyService) {
+            throw new Error("No verification service configured.");
+        }
+        const verification_check = await this.verifyService.verificationChecks.create({
+            to: phoneNumber,
+            code: oneTimePassword,
+        });
+        return verification_check.status === "approved";
+    }
+}
--- a/components/server/src/user/user-deletion-service.ts
+++ b/components/server/src/user/user-deletion-service.ts
@@ -65,6 +65,7 @@ export class UserDeletionService {
             this.anonymizeUser(user);
             this.deleteIdentities(user);
             await this.deleteTokens(db, user);
+            user.lastVerificationTime = undefined;
             user.markedDeleted = true;
             await db.storeUser(user);
         });
@@ -148,6 +149,9 @@ export class UserDeletionService {
         user.avatarUrl = "deleted-avatarUrl";
         user.fullName = "deleted-fullName";
         user.name = "deleted-Name";
+        if (user.verificationPhoneNumber) {
+            user.verificationPhoneNumber = "deleted-phoneNumber";
+        }
     }
 
     protected deleteIdentities(user: User) {
--- a/components/server/src/user/user-service.ts
+++ b/components/server/src/user/user-service.ts
@@ -32,6 +32,7 @@ import { AttributionId } from "@gitpod/gitpod-protocol/lib/attribution";
 import { StripeService } from "../../ee/src/user/stripe-service";
 import { ResponseError } from "vscode-ws-jsonrpc";
 import { ErrorCodes } from "@gitpod/gitpod-protocol/lib/messaging/error";
+import { VerificationService } from "../auth/verification-service";
 
 export interface FindUserByIdentityStrResult {
     user: User;
@@ -72,6 +73,7 @@ export class UserService {
     @inject(CostCenterDB) protected readonly costCenterDb: CostCenterDB;
     @inject(TeamDB) protected readonly teamDB: TeamDB;
     @inject(StripeService) protected readonly stripeService: StripeService;
+    @inject(VerificationService) protected readonly verificationService: VerificationService;
 
     /**
      * Takes strings in the form of <authHost>/<authName> and returns the matching User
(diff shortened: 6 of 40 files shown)


components/dashboard/src/start/VerifyModal.tsx
```diff
@@ -0,0 +1,201 @@
...
+                ) : (
+                    <></>
+                )}
+                <div className="mt-4">
+                    <h4>Verification Code</h4>
+                    <input
+                        autoFocus={true}
+                        className="w-full"
+                        type="text"
+                        placeholder="enter token send via sms"
```
gtsiolis, line 164: **issue**: Generally, placeholders are problematic for multiple reasons. Some older browsers don't support them[[1](https://caniuse.com/input-placeholder)] although this can be considered minor, some browsers hide them on focus, it's harder for users to check if their input complies with the placeholder hint, have lower contrast[[2](https://usecontra.st/a8a29e/fff)] which is easy to fail for color contrast accessibility, some screen readers don’t recognise them, and much more.

**suggestion**: Instead, we could use a help text below or above the input when possible. In this case, we can safely drop this placeholder text and rely on the input label and the instructions above for users to understand what action they need to take. That's the primary reason why the design specs in https://github.com/gitpod-io/gitpod/issues/11339#issuecomment-1183387915 do not include any placeholders.

```suggestion
```

components/dashboard/src/start/VerifyModal.tsx
```diff
@@ -0,0 +1,201 @@
...
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { useState } from "react";
+import Alert, { AlertType } from "../components/Alert";
+import Modal from "../components/Modal";
+import { getGitpodService } from "../service/service";
+import PhoneInput from "react-intl-tel-input";
```
gtsiolis, line 11: **question**: Unless there's a strong preference over this library, given the styling issues of the input, shall we consider switching to a different library like [`intl-tel-input`](https://github.com/jackocnr/intl-tel-input) which also happens to be sponsored by Twilio? Seems more popular and flexible, no?

> 270 vs. 6.4K ⭐

components/dashboard/src/start/VerifyModal.tsx
```diff
@@ -0,0 +1,201 @@
...
+                    <div>
+                        <button className="ml-2" disabled={!state.phoneNumberValid} onClick={sendCode}>
+                            Send Code via SMS
+                        </button>
+                    </div>
+                }
+                visible={true}
+            >
+                <Alert type="warning" className="mt-2">
+                    To use Gitpod you'll need to validate your account with your phone number. This is required to
```
gtsiolis, line 55: **nitpick**: Does it help us achieve our goal (abuse)? I'd rather go with a neutral tone but any option could suffice. Feel free to ignore this nitpick. 🙂 

```suggestion
                    To use Gitpod you'll need to validate your account with a mobile phone number. This is required to
```

gtsiolis (commented):
Left some early feedback below. 🥯

components/dashboard/src/start/VerifyModal.tsx
svenefftinge: We only support the current version of chrome, firefox and safari.

components/dashboard/src/start/VerifyModal.tsx
svenefftinge: It is the react version of that library.

components/dashboard/src/start/VerifyModal.tsx
```diff
@@ -0,0 +1,201 @@
...
+                </div>
+                {state.message ? (
+                    <Alert type={state.message.type} className="mt-4 py-3">
+                        {state.message.text}
+                    </Alert>
+                ) : (
+                    <></>
+                )}
+                <div className="mt-4">
+                    <h4>Verification Code</h4>
```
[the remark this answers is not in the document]
svenefftinge: Please try again. The authtoken was invalidated.

gtsiolis:
Looking at this now! 👀

components/dashboard/src/start/VerifyModal.tsx
```diff
@@ -0,0 +1,221 @@
...
+                    <input
+                        autoFocus={true}
+                        className="w-full"
+                        type="text"
+                        placeholder="enter token send via sms"
```
gtsiolis, line 184: Regarding https://github.com/gitpod-io/gitpod/pull/12258#discussion_r952581084:

> We only support the current version of chrome, firefox and safari.

Agree, makes sense. 🤝 Maybe worth being more specific about this also in [our docs](https://www.gitpod.io/docs/self-hosted/helm-deprecated/requirements#supported-web-browsers). Opened https://github.com/gitpod-io/website/pull/2617 as a follow-up PR but closed in https://github.com/gitpod-io/website/pull/2617#issuecomment-1224271963 as this is already covered by the upcoming compatibility matrix.

❗ **issue**: Still placeholders remain problematic from UX perspective. 

If there's a strong sense in keeping the placeholder, let's at least update the casing and replace the token term, which is a new term for users here.

```suggestion
                        placeholder="Enter code sent via SMS"
```

components/dashboard/src/start/VerifyModal.tsx
```diff
@@ -0,0 +1,221 @@
...
+                {state.message ? (
+                    <Alert type={state.message.type} className="mt-4 py-3">
+                        {state.message.text}
+                    </Alert>
+                ) : (
+                    <></>
+                )}
+                <div className="mt-4">
+                    <h4>Mobile Phone Number</h4>
+                    {/* HACK: Below we are adding a dummy dom element that is not visible, to reference the classes so they are not removed by purgeCSS. */}
```
gtsiolis, line 81: **issue**: Well, that's not so invisible over here! 🙈 

Screenshot 2022-08-23 at 7 10 16 PM

components/dashboard/src/start/VerifyModal.tsx
```diff
@@ -0,0 +1,221 @@
...
+                        </button>
+                    </div>
+                }
+                visible={true}
+            >
+                <Alert type="warning" className="mt-2">
+                    To use Gitpod you'll need to validate your account with your phone number. This is required to
+                    discourage and reduce abuse on Gitpod infrastructure.
+                </Alert>
+                <div className="mt-2 -mx-6 px-6 pt-4">
```
gtsiolis, line 69: **suggestion**: Sounds safe to keep only the top margin here.

```suggestion
                <div className="mt-2">
```

components/dashboard/src/start/VerifyModal.tsx
```diff
@@ -0,0 +1,221 @@
...
+                    <Alert type={state.message.type} className="mt-4 py-3">
+                        {state.message.text}
+                    </Alert>
+                ) : (
+                    <></>
+                )}
+                <div className="mt-4">
+                    <h4>Mobile Phone Number</h4>
+                    {/* HACK: Below we are adding a dummy dom element that is not visible, to reference the classes so they are not removed by purgeCSS. */}
+                    <input type="tel" className="invisible intl-tel-input country-list" />
```
gtsiolis, line 82: **suggestion**: Going with hidden will also remove the occupied space here. Hopefully, this won't case any issues with referencing the classes, right?

```suggestion
                    <input type="tel" className="hidden intl-tel-input country-list" />
```

gtsiolis, line 82: **issue**: Not sure which element is causing this, but would it be easy to remove the double border here?

Screenshot 2022-08-23 at 7 41 02 PM

components/dashboard/src/start/VerifyModal.tsx
```diff
@@ -0,0 +1,221 @@
...
+        return (
+            <Modal
+                onClose={() => {}}
+                closeable={false}
+                onEnter={sendCode}
+                title="User Validation Required"
+                buttons={
+                    <div>
+                        <button className="ml-2" disabled={!state.phoneNumberValid} onClick={sendCode}>
+                            Send Code via SMS
```
gtsiolis, line 59: **issue**: Using two Greek phone number returns the following error. Not sure if I did anything wrong here. 🤷 

Screenshot 2022-08-23 at 7 42 36 PM

gtsiolis, line 59: **issue**: When selecting a different country and using the sample number from the placeholder, the flow proceeds to the following screen that asks for the verification code, but still returns the same error.

Screenshot 2022-08-23 at 7 51 41 PM

gtsiolis (commented):
@svenefftinge I've left another round of comments. 🥊 

I could not go through all the steps of the flow. 😭

components/dashboard/src/start/VerifyModal.tsx
svenefftinge: I had to update the preferences in twilio.

svenefftinge: fixed

gtsiolis:
Looking at this now! 👀

components/dashboard/src/start/VerifyModal.tsx
```diff
@@ -0,0 +1,220 @@
...
+        };
+        return (
+            <Modal
+                onClose={() => {}}
+                closeable={false}
+                onEnter={verifyToken}
+                title="User Validation Required"
+                buttons={
+                    <div>
+                        <button className="ml-2" disabled={!isTokenFilled()} onClick={verifyToken}>
```
gtsiolis, line 147: **issue(non-blocking)**: The **disabled button** state should have the  "not-allowed" cursor when hovered and not relying on the CSS class for disabling the interaction. Also, relevant to the comment about the button component.

**suggestion**: Tailwind supports `cursor-not-allowed` which is something to consider when we build the button component, see [relevant docs](https://tailwindcss.com/docs/cursor).

components/dashboard/src/start/VerifyModal.tsx
```diff
@@ -0,0 +1,220 @@
...
+                            Validate Account
```
gtsiolis, line 148: **suggestion**: I noticed we've skipped the _verification in progress_ step from the initial design specs in https://github.com/gitpod-io/gitpod/issues/11339#issuecomment-1183387915 which is fine, but what do you think of injecting a loading state inside the button here with the spinner icon? FWIW, we've used this pattern for running prebuilds in the past, see relevant diff below.

**thought**: We could use the same pattern also for sending the SMS via code since we rely on an external service for this action.

**thought**: The loading state of the button is another great example use case where a **button component** could be used. 💭 

https://github.com/gitpod-io/gitpod/blob/31332971069a926428151aae2eacac947ed2db21/components/dashboard/src/projects/Prebuild.tsx#L160-L169

Re-posting the verification in progress design from https://github.com/gitpod-io/gitpod/issues/11339#issuecomment-1183387915 for visibility:

> | Verification (Progress) |
> |-|
> | UserValidationModalVerificationInProgress |

components/dashboard/src/start/VerifyModal.tsx
```diff
@@ -0,0 +1,220 @@
...
+        };
+        return (
+            <Modal
+                onClose={() => {}}
+                closeable={false}
+                onEnter={sendCode}
+                title="User Validation Required"
+                buttons={
+                    <div>
+                        <button className="ml-2" disabled={!state.phoneNumberValid} onClick={sendCode}>
```
gtsiolis, line 59: **thought**: I noticed we also skipped the resend code functionality which sounds ok for now. 🤝 

Re-posting the verification in progress design from https://github.com/gitpod-io/gitpod/issues/11339#issuecomment-1183387915 for visibility:

> | Verification (Resend) |
> |-|
> | UserValidationModalVerificationResent |

components/dashboard/src/start/VerifyModal.tsx
```diff
@@ -0,0 +1,220 @@
...
+                            Continue
+                        </button>
+                    </div>
+                }
+                visible={true}
+            >
+                <Alert type="message" className="mt-2">
+                    Your account has been successfully verified.
+                </Alert>
+            </Modal>
```
gtsiolis, line 217: **praise**: Thanks for giving this a go while against the [relevant discussion](https://gitpod.slack.com/archives/C01KPEPQYE6/p1657728367916349?thread_ts=1657727187.914319&cid=C01KPEPQYE6) (internal)!

components/dashboard/src/start/VerifyModal.tsx
```diff
@@ -0,0 +1,220 @@
...
+                }
+                visible={true}
+            >
+                <Alert type="warning" className="mt-2">
+                    To use Gitpod you'll need to validate your account with your phone number. This is required to
+                    discourage and reduce abuse on Gitpod infrastructure.
+                </Alert>
+                <div className="mt-2">Enter a mobile phone number you would like to use to verify your account.</div>
+                {state.message ? (
+                    <Alert type={state.message.type} className="mt-4 py-3">
```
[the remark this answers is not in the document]
gtsiolis: **issue(non-blocking)**: I've also run into this error but should be fine to ignore. Twilio probably hates me already. 😇 

Screenshot 2022-08-24 at 2 30 49 PM

components/dashboard/src/start/VerifyModal.tsx
```diff
@@ -0,0 +1,220 @@
...
+                    <div>
+                        <button className="ml-2" onClick={continueStartWorkspace}>
+                            Continue
+                        </button>
+                    </div>
+                }
+                visible={true}
+            >
+                <Alert type="message" className="mt-2">
+                    Your account has been successfully verified.
```
gtsiolis, line 215: **suggestion**: Now that we've tried this, we probably need a new **alert component variant** for successful messages like this that is using green-ish colors. Could be fine to leave this out of the scope of these changes for now. Your call.

| Alert / Info (IMPLEMENTED) | Alert / Success (NOT IMPLEMENTED) | Alert / Success (NOT IMPLEMENTED) 🌔  |
|-|-|-|
| Frame 381 | Frame 382 | Frame 385 |

If you'd like to add this here, here's the icon (SVG) and the colors used for light and dark themes:


#### LIGHT THEME ⛅  
> Background: `green-100`
> Icon: `green-500`
> Text: `green-700`

```svg
<svg width="20" height="20" viewBox="0 0 20 20" fill="none" xmlns="http://www.w3.org/2000/svg">
<path fill-rule="evenodd" clip-rule="evenodd" d="M10 18C14.4183 18 18 14.4183 18 10C18 5.58172 14.4183 2 10 2C5.58172 2 2 5.58172 2 10C2 14.4183 5.58172 18 10 18ZM13.7071 8.70711C14.0976 8.31658 14.0976 7.68342 13.7071 7.29289C13.3166 6.90237 12.6834 6.90237 12.2929 7.29289L9 10.5858L7.70711 9.29289C7.31658 8.90237 6.68342 8.90237 6.29289 9.29289C5.90237 9.68342 5.90237 10.3166 6.29289 10.7071L8.29289 12.7071C8.68342 13.0976 9.31658 13.0976 9.70711 12.7071L13.7071 8.70711Z" fill="#84CC16"/>
</svg>
```

####  DARK THEME 🌔  
> Background: `green-800`
> Icon: `green-50`
> Text: `green-100`

```svg
<svg width="20" height="20" viewBox="0 0 20 20" fill="none" xmlns="http://www.w3.org/2000/svg">
<path fill-rule="evenodd" clip-rule="evenodd" d="M10 18C14.4183 18 18 14.4183 18 10C18 5.58172 14.4183 2 10 2C5.58172 2 2 5.58172 2 10C2 14.4183 5.58172 18 10 18ZM13.7071 8.70711C14.0976 8.31658 14.0976 7.68342 13.7071 7.29289C13.3166 6.90237 12.6834 6.90237 12.2929 7.29289L9 10.5858L7.70711 9.29289C7.31658 8.90237 6.68342 8.90237 6.29289 9.29289C5.90237 9.68342 5.90237 10.3166 6.29289 10.7071L8.29289 12.7071C8.68342 13.0976 9.31658 13.0976 9.70711 12.7071L13.7071 8.70711Z" fill="#F7FEE7"/>
</svg>
```

components/dashboard/src/start/VerifyModal.tsx
```diff
@@ -0,0 +1,220 @@
...
+        const continueStartWorkspace = () => {
+            window.location.reload();
+            return true;
+        };
+        return (
+            <Modal
+                onClose={continueStartWorkspace}
+                closeable={false}
+                onEnter={continueStartWorkspace}
+                title="User Validation Successful"
```
gtsiolis, line 204: **thought**: We don't have a concrete pattern for user feedback on successful messages. So far, we've used the modal and the page for user feedback like this, see screenshots below.

| New project ✔️  | New feedback ✔️ |
|-|-|
| Screenshot 2022-08-24 at 1 15 28 PM (2) | Screenshot 2022-08-24 at 1 16 30 PM (2) |

gtsiolis, line 204: **suggestion**: Given it's not ideal to update the modal title based on the user action and taking into account the [relevant discussion](https://gitpod.slack.com/archives/C01KPEPQYE6/p1657728367916349?thread_ts=1657727187.914319&cid=C01KPEPQYE6) (internal), what do you think of moving the alert component on the loading screen directly where we later on show the warnings about the latest editor release and skip asking users to acknowledge the account validation?

| Latest Release on Workspace Start | Validation on Workspace Start | 
|-|-|
| Screenshot 2022-08-24 at 2 52 07 PM (2) | Screenshot 2022-08-24 at 2 51 18 PM (2) |

gtsiolis, line 204: **thought**: In case this could be interesting we could surface the account validation status on the account settings.

gtsiolis (commented):
Alright! Left a few general comments and questions below. 🔮 

Most of them can be ignored or used for opening follow up issues out of the scope of these changes. 🛹 

Thanks @svenefftinge for pushing this forward! 🏀

components/dashboard/src/start/VerifyModal.tsx
```diff
@@ -0,0 +1,220 @@
...
+                message: undefined,
+                token: undefined,
+            });
+        };
+        return (
+            <Modal
+                onClose={() => {}}
+                closeable={false}
+                onEnter={verifyToken}
+                title="User Validation Required"
```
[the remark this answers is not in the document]
svenefftinge: That is intended. the code is uncommented so you can review the UI flow easily.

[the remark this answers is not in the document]
gtsiolis: Ah, ok! Resolving.

components/dashboard/src/start/VerifyModal.tsx
```diff
@@ -0,0 +1,220 @@
...
+        return (
+            <Modal
+                onClose={() => {}}
+                closeable={false}
+                onEnter={sendCode}
+                title="User Validation Required"
+                buttons={
+                    <div>
+                        <button className="ml-2" disabled={!state.phoneNumberValid} onClick={sendCode}>
+                            Send Code via SMS
```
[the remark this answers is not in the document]
svenefftinge: aren't the two limits you posted below what you are asking for here?

components/dashboard/src/start/VerifyModal.tsx
```diff
@@ -0,0 +1,220 @@
...
+                        </button>
+                    </div>
+                }
+                visible={true}
+            >
+                <Alert type="warning" className="mt-2">
+                    To use Gitpod you'll need to validate your account with your phone number. This is required to
+                    discourage and reduce abuse on Gitpod infrastructure.
+                </Alert>
+                <div className="mt-2">Enter a mobile phone number you would like to use to verify your account.</div>
```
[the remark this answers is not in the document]
svenefftinge: Why would you want to limit it?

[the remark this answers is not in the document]
gtsiolis: No strong feelings, only bringing this up because I noticed this behavior while testing with two numbers. Maybe we can leave this as is and rephrase the copy as described in https://github.com/gitpod-io/gitpod/pull/12258#discussion_r952511755? 🤷

components/dashboard/src/start/VerifyModal.tsx
svenefftinge: fixed 😇

components/dashboard/src/start/VerifyModal.tsx
svenefftinge: I think an explicit, "you are verified" message with a 'continue' button is clearer. Especially because we are reloading the page.

svenefftinge: It's a SaaS-only concern and should really touch as few places as possible. Why would it be interesting to show it there?

gtsiolis: Good point! Not much interesting, indeed.

I probably had in mind how we could use the same flow to validate with a credit card if we need in the future so that users benefit from having a credit card in place when and if they enable billing for their personal account. Resolving for now. 💳

svenefftinge: We can mark users as verified when they take other actions that make them trustworthy. No UI needed.

gtsiolis: Yeah, maybe also something to consider surfacing in the admin area. It will be interesting to see if users manage to bypass somehow this verification process for abusing the infrastructure. 💡 

Cc @jldec because https://github.com/gitpod-io/gitpod/issues/7597
Cc @geropl because https://github.com/gitpod-io/gitpod/issues/11469

gtsiolis: **thought**: It is certainly clearer and gives users the time to acknowledge this.

**question**: What do you think of keeping modal, and updating the modal title to be the same on all cases, and rely on modal body (paragraph and alerts) to inform the user about validation errors or successful validation?

gtsiolis: Fair point. Resolving this. 🏀

svenefftinge:
I'll hide this behind a feature flag so we can carefully enable it in prod.

corneliusludmann (approved):
It seems that this PR does not have changes in Team Self-Hosted owned files anymore. However, from self-hosted perspective, it looks good so far.

:bangbang: _**Hint:** You should rebase shortly before you merge this and let golden files of the installer render tests be re-generated since there are new tests in `main` already that will let fail the `main` build after merging. Sorry for the inconvenience._

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -438,12 +442,20 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
...
 
-        const allowedFields: (keyof User)[] = ["avatarUrl", "fullName", "additionalData"];
+        const allowedFields: (keyof User)[] = ["avatarUrl", "fullName"];
         for (const p of allowedFields) {
             if (p in partialUser) {
                 (user[p] as any) = partialUser[p];
             }
         }
+        // special treatment of nested additionalData
+        if (partialUser.additionalData) {
```
AlexTugarev, line 452: Preferably, let's not use a field dedicated for user preferences/settings for something that actually controls the wheels or breaks. That check seems to work fine, I've tried to break without success, but I'd prefer not to have a this guarded places in a field which otherwise is free to change for users. 

There is an option that we can use to add field to tables without the downside of locking tables, we just need to add `NOT NULL DEFAULT '', ALGORITHM=INPLACE, LOCK=NONE` for the alter statement.

As we used this alternative the last time(s) discussed this matter, I'd prefer to stick to one pattern 🙏🏻

components/server/ee/src/billing/entitlement-service.ts
```diff
@@ -31,13 +32,21 @@ export class EntitlementServiceImpl implements EntitlementService {
...
     @inject(EntitlementServiceUBP) protected readonly ubp: EntitlementServiceUBP;
+    @inject(VerificationService) protected readonly verificationService: VerificationService;
 
     async mayStartWorkspace(
         user: User,
         date: Date = new Date(),
         runningInstances: Promise<WorkspaceInstance[]>,
     ): Promise<MayStartWorkspaceResult> {
         try {
+            const verification = await this.verificationService.needsVerification(user);
```
AlexTugarev, line 43: Nice integration! Fits in perfectly. 

That reminds me of the ToS flow integration, which intercepted the signup, but it would actually be better to use this kind of hook.

components/server/src/auth/rate-limiter.ts
```diff
@@ -51,6 +51,8 @@ function getConfig(config: RateLimiterConfig): RateLimiterConfig {
         getLoggedInUser: { group: "default", points: 1 },
         getTerms: { group: "default", points: 1 },
         updateLoggedInUser: { group: "default", points: 1 },
+        sendPhoneNumberVerificationToken: { group: "default", points: 1 },
```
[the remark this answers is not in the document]
AlexTugarev: I guess this also applies to `verifyPhoneNumberVerificationToken`.

components/server/src/auth/verification-service.ts
```diff
@@ -0,0 +1,78 @@
...
+        }
+        const verification = await this.verifyService.verifications.create({ to: phoneNumber, channel: "sms" });
+        log.info("Verification code sent", { phoneNumber, status: verification.status });
+    }
+
+    public async verifyVerificationToken(phoneNumber: string, oneTimePassword: string): Promise<boolean> {
+        if (!this.verifyService) {
+            throw new Error("No verification service configured.");
+        }
+        const verification_check = await this.verifyService.verificationChecks.create({
```
AlexTugarev, line 72: try-catch, otherwise it's handled as internal server error whenever one uses the happy finger on the send button.

AlexTugarev (commented):
The account verification works nicely 🎉 

I left some comment, and I'm happy to discuss the `additionalData` comment 👍🏻 
Ah, the only real issue so far is that old GH account are not count as verified, because the check is not performed after signup, but rather after the following re-login.

kylos101:
:wave:  @svenefftinge, would it be possible to include bitbucket with this too, or perhaps in a follow-on PR? I ask because of [this thread](https://gitpod.slack.com/archives/C024TUZ0MDJ/p1661457779555119).

svenefftinge:
The logic is opt-out. So all bitbucket users need to verify.

components/server/src/user/user-deletion-service.ts
```diff
@@ -66,6 +66,9 @@ export class UserDeletionService {
             this.deleteIdentities(user);
             await this.deleteTokens(db, user);
             user.markedDeleted = true;
+            if (user.additionalData) {
+                delete user.additionalData.lastVerificationTime;
```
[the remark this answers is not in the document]
svenefftinge: I intentionally left it in order to follow up in case the user did something bad.

components/server/src/workspace/gitpod-server-impl.ts
svenefftinge: Which discussion do you refer to?

components/server/src/auth/rate-limiter.ts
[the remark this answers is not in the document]
svenefftinge: Ah, yes. Probably worth limiting this per user on our side, although there is a rate limit per phone number on twilio as well.

components/server/src/user/user-service.ts
```diff
@@ -419,6 +422,11 @@ export class UserService {
         user.name = user.name || authUser.name || authUser.primaryEmail;
         user.avatarUrl = user.avatarUrl || authUser.avatarUrl;
 
+        if (authUser.created_at && isDateSmaller(authUser.created_at, daysBefore(new Date().toISOString(), 30))) {
```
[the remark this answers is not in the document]
svenefftinge: Good catch! 🙏

components/server/src/auth/verification-service.ts
svenefftinge: I don't get this comment. Why is there always an internal server error?

components/server/src/user/user-service.ts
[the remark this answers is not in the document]
AlexTugarev: Let's move this check to `createUser`.

components/server/src/workspace/gitpod-server-impl.ts
AlexTugarev: To avoid implementing the checks for nested attributes in the `additionalData` bucket we've moved the attribution override to `User` in https://github.com/gitpod-io/gitpod/pull/11522. The validation necessary is to disallow users setting the attribution target to a team id they are not a member of.

components/server/src/user/user-deletion-service.ts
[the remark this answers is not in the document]
AlexTugarev: That we definitely need to check with data privacy. To prevent recreating accounts it would be sufficient to store a hash of the phone number, no?

components/server/src/auth/verification-service.ts
AlexTugarev: as this is not wrapped in a `ResponseError` it'll be logged automatically as an internal server error, thus it's loud on the logs, and it seems not to be rendered properly on frontend, at least I couldn't see twillios rate limiting error rendered on frontend.

components/server/src/user/user-deletion-service.ts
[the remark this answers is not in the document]
svenefftinge: Ok, I'll store the hash instead and introduce `d_b_blocked_phone_hashes`that we update on user block/unblock. Also will check that table on first sign up.

components/server/src/auth/rate-limiter.ts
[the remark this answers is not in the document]
svenefftinge: I'm introducing a new category `phoneVerification` with 10 points per 10 seconds.

components/server/src/user/user-deletion-service.ts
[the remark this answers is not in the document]
svenefftinge: Update: I'm leaving the blocked_phone_hashes out of this PR, as it makes it even bigger (it already is big).
But I'm moving the two properties into `d_b_user`.

components/server/src/auth/verification-service.ts
svenefftinge: It is fine to be loud on the logs. It's not a user error when an error is thrown but something with the system. The error is thrown to frontend and I'll make it so that the user gets feedback about the error.

components/server/src/user/user-service.ts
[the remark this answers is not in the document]
svenefftinge: moved it to the callback that is passed to `createUser` because I needed access to the authUser object. Also looked like a good pace for it otherwise. Fine?

jldec:
🤞

components/server/src/user/user-deletion-service.ts
```diff
@@ -148,6 +149,9 @@ export class UserDeletionService {
         user.avatarUrl = "deleted-avatarUrl";
         user.fullName = "deleted-fullName";
         user.name = "deleted-Name";
+        if (user.verificationPhoneNumber) {
+            user.verificationPhoneNumber = "deleted-verificationPhoneNumber";
```
AlexTugarev, line 153: `ER_DATA_TOO_LONG: Data too long for column 'verificationPhoneNumber'`

we're almost there, just 2 chars too long.

svenefftinge: what did you type in?

svenefftinge: ah, got it.

AlexTugarev: That happens on deletion of an account, which actually breaks the process.

AlexTugarev: `deleted-phoneNumber` with a length of 20 should work just fine.

AlexTugarev (approved):
LGTM!
Tested works as advertised.

Logic for `Test 1 ++` isn't there is code, which I think is not needed at all ;-)

PR #12261, opened 2022-08-22
Ensure workspace classes to not break selfhosted customers

Furisto:
## Description
See https://gitpod.slack.com/archives/C01KLC56NP7/p1660828231371779
Lets walk through the scenarios:
- Self-hosted
  - The installer surface is the same as before workspace classes
  - MinLimit and BurstLimit will be set to the value specified in workspace.resources.limits.cpu
  - The BurstLimit will be used to set the Kubernetes limit (i.e. workspace.resources.limits.cpu as before)
  -- > most customers are done at this point
  - If customers also have the ws-daemon cpu limiting enabled
    - Limiting will be applied based on the configuration in ws-daemon
    - It will not use the annotation based limiting because the annotation will not be set on the workspace pod. This would require that the workspace_classes feature flag is enabled. This is not the case on self-hosted.

- SaaS
  - Customers for which workspace classes have been enabled
    - Workspaces will either get g1-standard or g1-large, default is not available to them, so this has no impact on them.
  - Customers for which workspace classes have _not_ been enabled
      - MinLimit and BurstLimit will be set to the value specified in workspace.resources.limits.cpu
      - The BurstLimit will be used to set the Kubernetes limit (i.e. workspace.resources.limits.cpu as before)
      - ws-daemon cpu limiting is enabled
      - Limiting will be applied based on the configuration in ws-daemon because annotation based limiting requries the WORKSPACE_CLASS_LIMITING feature flag which is only send to ws-manager if workspace classes are enabled for the customer

## Related Issue(s)

Fixes #

## How to test
Check golden files

## Release Notes

```release-note
None
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/installer/cmd/testdata/render/external-registry/output.golden
+++ b/install/installer/cmd/testdata/render/external-registry/output.golden
@@ -1551,14 +1551,6 @@ data:
         snapshotClass: ""
         storageClass: ""
       resources:
-        limits:
-          cpu:
-            buckets: null
-            burst: ""
-            min: ""
-          ephemeral-storage: ""
-          memory: ""
-          storage: ""
         requests:
           cpu: "1"
           memory: 2Gi
--- a/install/installer/cmd/testdata/render/minimal/output.golden
+++ b/install/installer/cmd/testdata/render/minimal/output.golden
@@ -1617,14 +1617,6 @@ data:
         snapshotClass: ""
         storageClass: ""
       resources:
-        limits:
-          cpu:
-            buckets: null
-            burst: ""
-            min: ""
-          ephemeral-storage: ""
-          memory: ""
-          storage: ""
         requests:
           cpu: "1"
           memory: 2Gi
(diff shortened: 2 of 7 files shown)


MrSimonEmms (changes requested):
I think this is as we expect it for Self-Hosted, thanks.

It would be helpful if a new golden file test were added with the workspace resource set. This would nicely demonstrate this change.

corneliusludmann:
> It would be helpful if a new golden file test were added with the workspace resource set. This would nicely demonstrate this change.

:100: %!

As I mentioned in the Slack thread, [I used those](https://github.com/gitpod-io/gitpod/pull/12206) to tests the changes. Feel free to copy them.

aledbf (approved):

Furisto:
@MrSimonEmms PTAL

MrSimonEmms (approved):
LGTM

MrSimonEmms:
@Furisto it looks like this has broken the main build https://werft.gitpod-dev.com/job/gitpod-build-main.4334

Can you open another PR with a new generation of the golden files and send it my way please to approve?

Furisto:
@MrSimonEmms On it

Furisto:
@MrSimonEmms https://github.com/gitpod-io/gitpod/pull/12270

PR #12300, opened 2022-08-23
[content-service] Define UsageReportService.DownloadURL rpc

easyCZ:
## Description

As part of marking Workspace Instances as already billed, we'll need to download the latest report an invoice is based on, and from that report mark all instances for that invoice as already billed for.

This PR adds the RPC definition without actually implementing it, that will come in subsequent PR.

**What is the security model for content-service?**
* Is it possible for Workspaces to interact with the content service directly?
* What prevents the caller from asking for a report without credentials?

## Related Issue(s)

* https://github.com/gitpod-io/gitpod/issues/10937

## How to test

NONE

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


components/content-service-api/usage.proto
```diff
@@ -11,6 +11,9 @@ option go_package = "github.com/gitpod-io/gitpod/content-service/api";
 service UsageReportService {
   // UploadURL provides a URL to which clients can upload the content via HTTP PUT.
   rpc UploadURL(UsageReportUploadURLRequest) returns (UsageReportUploadURLResponse) {}
+
+  // GetDownloadURL retrieves a URL which can be used to download a Usage Report.
+  rpc GetDownloadURL(GetDownloadURLRequest) returns (GetDownloadURLResponse) {}
```
jenting, line 16: Should we have a verb `Get` for DownloadURL? Because we use UploadURL at line 13, should we define it as `DownloadURL` instead?

easyCZ: I don't mind. IMO, the RPC above should've had it as it indicates the action being performed but I can drop it from here also..

jenting: References
- https://github.com/gitpod-io/gitpod/blob/b5a7a2ff91f97e78852e44bab4f54fd7aeb433da/components/content-service-api/blobs.proto#L11-L16
- https://github.com/gitpod-io/gitpod/blob/b5a7a2ff91f97e78852e44bab4f54fd7aeb433da/components/content-service-api/ideplugin.proto#L11-L16

easyCZ: I've updated the protos.

jenting (approved):

geropl (approved):
LGTM

PR #12302, opened 2022-08-23
GetUpcomingInvoice for BillingService

### LINKED ISSUE #11692: Implement BillingService.GetLatestInvoice

AlexTugarev:
## Description
This PR implements ~`GetLatestInvoice`~  `GetUpcomingInvoice`. The rename is done intentionally, to a) align with terminology used in stripe, and b) to distinguish between the next invoice and an already paid invoice. The upcoming invoice contains the the amount of credits currently used (🤞🏻 after discounts), which can be compared with the spending limit configured.

- [x] rename GetLatestInvoice to GetUpcomingInvoice

## Related Issue(s)

Fixes #11692 

## How to test
Setup UBB in for a team and lower the spending limit to 1 or 2 credits for a simple test, then start a workspace or two to hit that limit. After a while, cannot tell it precisely, you'll need to reload the dashboard to refresh the spending limit checks. In the `server` component you'll find a log message on spending limit events, and the known dashboard notification should be visible:
Screen Shot 2022-08-23 at 16 21 48

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment


roboquat:

<details>

</details>

components/server/ee/src/billing/billing-service.ts
```diff
@@ -40,50 +44,45 @@ export class BillingService {
...
-            {},
-            attributionId,
-            startedTimeOrder as number,
-            timestampFrom,
-            timestampTo,
-        );
-        const sessions = response.getSessionsList().map((s) => UsageService.mapBilledSession(s));
-        return sessions;
+    async getUpcomingInvoice(teamId: string): Promise<GetUpcomingInvoiceResponse> {
+        const response = await this.billingServiceClientProvider.getDefault().getUpcomingInvoice(teamId);
```
AlexTugarev, line 85: WDYT of using `attributionId` here, and using `ParseAttributionID` on the side of `BillingService`?

geropl: :point_up: Thought about the same. :brain: :sparkles: :+1:

easyCZ (commented):
This is a great example where splitting the PR up into the following blocks would greatly help with quality of review:
1. Proto changes
2. Implement GetUpcoming rpc
3. Update Server to use it.

components/server/ee/src/billing/billing-service.ts
easyCZ, line 85: When this rolls out, you can't guarantee that when this is executed, the `usage` component has been rolled out to a newer version and that the `getUpcomingInvoice` RPC actually exists.

In the happy case, this inconsistency won't happen, but in the bad case, where for example usage fails to start and remains on the old version, it will start failing here.

components/usage/pkg/stripe/stripe.go
```diff
@@ -122,6 +130,38 @@ func (c *Client) updateUsageForCustomer(ctx context.Context, customer *stripe.Cu
...
+	query := fmt.Sprintf("metadata['teamId']:'%s'", teamId)
+	searchParams := &stripe.CustomerSearchParams{
+		SearchParams: stripe.SearchParams{
+			Query:   query,
+			Expand:  []*string{stripe.String("data.subscriptions")},
+			Context: ctx,
+		},
+	}
+	iter := c.sc.Customers.Search(searchParams)
+	if !iter.Next() {
```
easyCZ, line 144: First check for errors:
```suggestion
	if iter.Error() != nil {
```

components/usage/pkg/stripe/stripe.go
```diff
@@ -122,6 +130,38 @@ func (c *Client) updateUsageForCustomer(ctx context.Context, customer *stripe.Cu
...
+		return nil, fmt.Errorf("failed to find customer data for team %s", teamId)
+	}
+	customer := iter.Customer()
+	invoiceParams := &stripe.InvoiceParams{
+		Customer: stripe.String(customer.ID),
+	}
```
easyCZ, line 150: What are the semantics when the search returns multiple customers? That should never happen, but we must be defensive here. I'd recommend collecting all the customers (or checking number of results), checking there's exactly one, and then returning.

components/usage/pkg/stripe/stripe.go
```diff
@@ -122,6 +130,38 @@ func (c *Client) updateUsageForCustomer(ctx context.Context, customer *stripe.Cu
...
+	searchParams := &stripe.CustomerSearchParams{
+		SearchParams: stripe.SearchParams{
+			Query:   query,
+			Expand:  []*string{stripe.String("data.subscriptions")},
+			Context: ctx,
+		},
+	}
+	iter := c.sc.Customers.Search(searchParams)
+	if !iter.Next() {
+		return nil, fmt.Errorf("failed to find customer data for team %s", teamId)
```
easyCZ, line 145: Worth returning a special error for this case. We want the API to return a useful error message to the client. In this case, we want to say that for a given TeamID, there's no customer in Stripe. The general error here can't be classified (or compared, aside from string comparisons).

The pattern for this is to do:

```
var CustomerNotFoundError = errors.New(...)
...

if errors.Is(err, CustomerNotFoundError) {
  return NotFound
}
```

components/usage/pkg/stripe/stripe.go
```diff
@@ -122,6 +130,38 @@ func (c *Client) updateUsageForCustomer(ctx context.Context, customer *stripe.Cu
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch the upcoming invoice for customer %s", customer.ID)
+	}
+
+	return &StripeInvoice{
+		ID:             invoice.ID,
+		SubscriptionID: invoice.Subscription.ID,
+		Amount:         invoice.AmountRemaining,
+		Currency:       string(invoice.Currency),
+		Credits:        invoice.Lines.Data[0].Quantity,
```
easyCZ, line 161: Check that Data has a 0 item, this would panic if it was empty.

components/usage/pkg/stripe/stripe_test.go
```diff
@@ -83,3 +85,25 @@ func TestCustomerQueriesForTeamIds_MultipleQueries(t *testing.T) {
 		})
 	}
 }
+
+func TestGetUpcomingInvoice(t *testing.T) {
+	secretKey := os.Getenv("TEST_KEY_STRIPE")
+	teamId := os.Getenv("TEST_TEAM_ID")
```
easyCZ, line 91: How are these injected during test runs on CI?

easyCZ: If it's not gonna run most of the time on CI, does it make sense to keep it here?

components/usage/pkg/apiv1/billing.go
```diff
@@ -58,6 +58,21 @@ func (s *BillingService) FinalizeInvoice(ctx context.Context, in *v1.FinalizeInv
 	return &v1.FinalizeInvoiceResponse{}, nil
 }
 
+func (s *BillingService) GetUpcomingInvoice(ctx context.Context, in *v1.GetUpcomingInvoiceRequest) (*v1.GetUpcomingInvoiceResponse, error) {
+	invoice, err := s.stripeClient.GetUpcomingInvoice(ctx, in.GetTeamId())
```
easyCZ, line 62: Check that TeamID is set, return InvalidArgument if not

easyCZ (changes requested):

components/server/ee/src/billing/billing-service.ts
easyCZ: The way to fix this is to roll-out the RPC first, and only then change the behaviour on server. Alternatively, you can use a feature flag to control this also.

AlexTugarev: Good point! Let me thinks about where to add the feature flag. First idea is to guard the spending limit check.

easyCZ: Personally, I think you can achieve the same thing (with fewer extra pieces) by splitting the PR up and first landing (and deploying) the changes which introduce the RPC on the proto level, but also implement it.

components/usage/pkg/stripe/stripe_test.go
AlexTugarev: removing...

components/usage/pkg/stripe/stripe.go
AlexTugarev: Good point. Copied from UpdateUsage without thinking about that further.

AlexTugarev:
GitHub, why? :-( 

> merged commit [1cd0fbf](https://github.com/gitpod-io/gitpod/commit/1cd0fbfd1d96ff1320d61cb78a31f5242ff7829e) into main

AlexTugarev:
Reopened as https://github.com/gitpod-io/gitpod/pull/12377

laushinka:
> GitHub, why? :-(

😢

easyCZ:
How did this PR get merged? I'm really confused.

Issue #12264, opened 2022-08-22
Cap reported Usage to max(Usage, SpendingLimit) before reporting to Stripe


easyCZ:
When creating an invoice, we need to cap the invoice to maximum of the usage limit. This is to ensure we conform to users expectations.

In practice, this means the following:
* When we are updating an invoice with a new value, we need to take into consideration the spending limit set.
* When a spending limit is greater than the current value, do nothing special.
* When the spending limit is lower or equal to current used value, cap the invoice to the spending limit.

Optionally:
* Tag the invoice (in Stripe) with metadata saying what the spending limit is, such that admins are not surprised when it is not increasing and do not need to hunt across multiple systems.

The number of credits used can exceed the spending limit due to eventual consistency in the system between reconciliation of usage starting of new workspaces. As such, this difference is where we are losing money. To monitor how much this diff is, we want to expose a metric which measures the diff between real usage and spending limit for each attribution.

* Add a metric which tracks "billing accuracy" or "credits lost"

Side change:
- Move ownership of the DB Cost Center to the Usage Component

geropl:
@laushinka could you document the outcome of the Slack discussion here? 🙏

PR #12304, opened 2022-08-23
[usage] Implement GetCostCenter RPC

laushinka:
## Description
Implements the GetCostCenter RPC.
The use of this RPC in UpdateInvoices() will be a separate PR.

## Related Issue(s)

Relates #12264

## How to test
Under `/components/usage/pkg/db` run `go test`

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage/pkg/apiv1/usage.go
+++ b/components/usage/pkg/apiv1/usage.go
@@ -7,6 +7,7 @@ package apiv1
 import (
 	"context"
 	"database/sql"
+	"errors"
 	"fmt"
 	"time"
 
@@ -134,6 +135,36 @@ func (s *UsageService) ReconcileUsage(ctx context.Context, req *v1.ReconcileUsag
 		ReportId: filename,
 	}, nil
 
+}
+
+func (s *UsageService) GetCostCenter(ctx context.Context, in *v1.GetCostCenterRequest) (*v1.GetCostCenterResponse, error) {
+	var attributionIdReq string
+
+	if in.AttributionId == "" {
+		return nil, status.Errorf(codes.InvalidArgument, "Empty attributionId")
+	}
+
+	attributionIdReq = in.AttributionId
+
+	attributionId, err := db.ParseAttributionID(attributionIdReq)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Failed to parse attribution ID: %s", err.Error())
+	}
+
+	result, err := db.GetCostCenter(ctx, s.conn, db.AttributionID(attributionIdReq))
+	if err != nil {
+		if errors.Is(err, db.CostCenterNotFound) {
+			return nil, status.Errorf(codes.NotFound, "Cost center not found: %s", err.Error())
+		}
+		return nil, status.Errorf(codes.Internal, "Failed to get cost center %s from DB: %s", in.AttributionId, err.Error())
+	}
+
+	return &v1.GetCostCenterResponse{
+		CostCenter: &v1.CostCenter{
+			AttributionId: string(attributionId),
+			SpendingLimit: result.SpendingLimit,
+		},
+	}, nil
 }
 
 func NewUsageService(conn *gorm.DB, reportGenerator *ReportGenerator, contentSvc contentservice.Interface) *UsageService {
--- a/components/usage/pkg/db/cost_center.go
+++ b/components/usage/pkg/db/cost_center.go
@@ -5,15 +5,20 @@
 package db
 
 import (
+	"context"
+	"errors"
+	"fmt"
 	"time"
 
-	"github.com/google/uuid"
+	"gorm.io/gorm"
 )
 
+var CostCenterNotFound = errors.New("CostCenter not found")
+
 type CostCenter struct {
-	ID            uuid.UUID `gorm:"primary_key;column:id;type:char;size:36;" json:"id"`
-	SpendingLimit int       `gorm:"column:spendingLimit;type:int;default:0;" json:"spendingLimit"`
-	LastModified  time.Time `gorm:"column:_lastModified;type:timestamp;default:CURRENT_TIMESTAMP(6);" json:"_lastModified"`
+	ID            AttributionID `gorm:"primary_key;column:id;type:char;size:36;" json:"id"`
+	SpendingLimit int32         `gorm:"column:spendingLimit;type:int;default:0;" json:"spendingLimit"`
+	LastModified  time.Time     `gorm:"->:column:_lastModified;type:timestamp;default:CURRENT_TIMESTAMP(6);" json:"_lastModified"`
 
 	// deleted is restricted for use by db-sync
 	_ bool `gorm:"column:deleted;type:tinyint;default:0;" json:"deleted"`
@@ -22,4 +27,20 @@ type CostCenter struct {
 // TableName sets the insert table name for this struct type
 func (d *CostCenter) TableName() string {
 	return "d_b_cost_center"
+}
+
+func GetCostCenter(ctx context.Context, conn *gorm.DB, attributionId AttributionID) (*CostCenter, error) {
+	db := conn.WithContext(ctx)
+	var costCenter CostCenter
+
+	result := db.First(&costCenter, attributionId)
+	if result.Error != nil {
+		if errors.Is(result.Error, gorm.ErrRecordNotFound) {
+			return nil, CostCenterNotFound
+		}
+		return nil, fmt.Errorf("failed to get cost center: %w", result.Error)
+
+	}
+
+	return &costCenter, nil
 }
--- a/components/usage/pkg/db/cost_center_test.go
+++ b/components/usage/pkg/db/cost_center_test.go
@@ -5,6 +5,7 @@
 package db_test
 
 import (
+	"context"
 	"testing"
 
 	"github.com/gitpod-io/gitpod/usage/pkg/db"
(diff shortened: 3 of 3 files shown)


laushinka:
Re-converting to draft while I tackle failing tests and another change

components/usage/pkg/apiv1/usage.go
```diff
@@ -136,6 +136,20 @@ func (s *UsageService) ReconcileUsage(ctx context.Context, req *v1.ReconcileUsag
 
 }
 
+func (s *UsageService) GetCostCenter(ctx context.Context, in *v1.GetCostCenterRequest) (*v1.GetCostCenterResponse, error) {
+	result, err := db.GetCostCenter(ctx, s.conn, db.AttributionID(in.AttributionId))
```
easyCZ, line 140: It's always worth checking the input you're getting. In this case, I think you can do the following:
1. Check `in.AttributionId` is not empty
2. Do `db.ParseAttributionID` and ensure it doesn't return an error

components/usage/pkg/apiv1/usage.go
```diff
@@ -136,6 +136,20 @@ func (s *UsageService) ReconcileUsage(ctx context.Context, req *v1.ReconcileUsag
...
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Failed to get cost center: %s", err)
```
easyCZ, line 142: Can we do better here? Can we identify if it's an error because it doesn't exist?

components/usage/pkg/db/cost_center.go
```diff
@@ -23,3 +25,18 @@ type CostCenter struct {
 func (d *CostCenter) TableName() string {
 	return "d_b_cost_center"
 }
+
+func GetCostCenter(ctx context.Context, conn *gorm.DB, attributionId AttributionID) (*CostCenter, error) {
+	db := conn.WithContext(ctx)
+	var costCenter CostCenter
+
+	result := db.
+		Where("ID = ?", attributionId).
```
easyCZ, line 34: Is the ID really capitalized in the DB?

You may be better of using the `db.First(&costCenter, attributionId)` function here

components/usage/pkg/db/cost_center_test.go
```diff
@@ -29,4 +30,28 @@ func TestCostCenter_WriteRead(t *testing.T) {
...
+
+	t.Cleanup(func() {
+		conn.Model(&db.CostCenter{}).Delete(costCenter)
+	})
+}
+
+func TestGetCostCenter(t *testing.T) {
+	conn := dbtest.ConnectForTests(t)
+
+	costCenter1 := &db.CostCenter{
```
easyCZ, line 42:
```suggestion
	costCenter := &db.CostCenter{
```

components/usage/pkg/server/server.go
```diff
@@ -151,12 +151,12 @@ func Start(cfg Config) error {
 	return nil
 }
 
-func registerGRPCServices(srv *baseserver.Server, conn *gorm.DB, stripeClient *stripe.Client, reportGenerator *apiv1.ReportGenerator, contentSvc contentservice.Interface, billInstancesAfter time.Time) error {
+func registerGRPCServices(srv *baseserver.Server, conn *gorm.DB, stripeClient *stripe.Client, reportGenerator *apiv1.ReportGenerator, contentSvc contentservice.Interface, billInstancesAfter time.Time, selfConnection grpc.ClientConnInterface) error {
```
easyCZ: Here, I'd change the interface to instead of taking a `selfConnection`, make it take an instance of `v1.UsageServiceClientInterface` (unsure if that's the exact type). This means that you no longer deal with how to establish the conenction, you just receive a client.

components/usage/pkg/db/cost_center.go
laushinka: > Is the ID really capitalized in the DB?

It's not. I was following the convention I saw in the other models in the package.

easyCZ: Do you have an example? I have feeling it wouldn't actually work in this case as `ID` would need to be the exact name of the column (unless it does some case ignore)

laushinka: `d_b_project` is `id` lowercase in the DB but `ID` in the model. `d_b_billed_session` is `instanceId` but `InstanceID` in the model. I saw these and assumed it ignores cases.

easyCZ: What matters for these are the annotations - the stuff in the backticks, look for something like 
```golang
`gorm:"column:usageAttributionId`
```

easyCZ: The field names (as in regular go Fields) can be renamed however you like, but the column names can't (or would require a migration)

easyCZ: > d_b_project is id lowercase in the DB but ID in the model. d_b_billed_session is instanceId but InstanceID in the model. I saw these and assumed it ignores cases.

In this case, it's `ID` as a field because the Go convention is to capitalize shorthands like id. But it still has lowercase `id` in the annotation, like `gorm:"primary_key;column:id;type:char;size:36;" json:"id"`

laushinka: Yess, the column annotations correspond to the columns

laushinka: Thanks good catch, the where clause was incorrect 🙌🏽 Have replaced it with the `db.First` syntax too.

components/usage/pkg/apiv1/usage.go
```diff
@@ -136,6 +136,25 @@ func (s *UsageService) ReconcileUsage(ctx context.Context, req *v1.ReconcileUsag
...
+	}
+
+	attributionID, err := db.ParseAttributionID(in.AttributionId)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to parse attribution ID: %w", err)
+	}
+
+	return &v1.GetCostCenterResponse{
+		CostCenter: &v1.CostCenter{
+			AttributionId: string(attributionID),
```
laushinka, line 152: Weirdly I still have to cast it to a string here. Otherwise I get `cannot use attributionID (variable of type db.AttributionID) as string value in struct literal`

components/usage/pkg/server/server.go
laushinka: This will be moved to the PR that tackles the use of the RPC.

components/usage/pkg/apiv1/usage.go
```diff
@@ -136,6 +136,36 @@ func (s *UsageService) ReconcileUsage(ctx context.Context, req *v1.ReconcileUsag
 
 }
 
+func (s *UsageService) GetCostCenter(ctx context.Context, in *v1.GetCostCenterRequest) (*v1.GetCostCenterResponse, error) {
+	var attributionIdReq string
+
+	if in.AttributionId != "" {
+		attributionIdReq = in.AttributionId
+	}
```
easyCZ, line 144: I think this can be removed

components/usage/pkg/apiv1/usage.go
```diff
@@ -136,6 +136,36 @@ func (s *UsageService) ReconcileUsage(ctx context.Context, req *v1.ReconcileUsag
...
+
+	if in.AttributionId == "" {
+		return nil, status.Errorf(codes.InvalidArgument, "Empty attributionId")
+	}
+
+	attributionId, err := db.ParseAttributionID(attributionIdReq)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Failed to parse attribution ID: %w", err)
```
easyCZ, line 152: Nice!

components/usage/pkg/apiv1/usage.go
```diff
@@ -136,6 +136,36 @@ func (s *UsageService) ReconcileUsage(ctx context.Context, req *v1.ReconcileUsag
...
+	}
+
+	result, err := db.GetCostCenter(ctx, s.conn, db.AttributionID(attributionIdReq))
+
```
easyCZ, line 156: [nit] I'd remove the empty here

components/usage/pkg/apiv1/usage.go
```diff
@@ -136,6 +136,36 @@ func (s *UsageService) ReconcileUsage(ctx context.Context, req *v1.ReconcileUsag
...
+	if err != nil {
+		return nil, status.Errorf(codes.NotFound, "Cost center not found: %s", err)
```
easyCZ, line 158: Here, you'll need to do the `errors.Is(db.CostCenterNotFoundError, err)` check and only return NotFound if the check evalutes to true, otherwise return `codes.Internal` for the status code

easyCZ (changes requested):

components/usage/pkg/db/cost_center_test.go
```diff
@@ -29,4 +30,28 @@ func TestCostCenter_WriteRead(t *testing.T) {
...
+
+func TestGetCostCenter(t *testing.T) {
+	conn := dbtest.ConnectForTests(t)
+
+	costCenter := &db.CostCenter{
+		ID:            db.NewTeamAttributionID(uuid.New().String()),
+		SpendingLimit: 300,
+	}
+
+	conn.Create(costCenter)
```
easyCZ, line 47:
```suggestion
	require.NoError(t, conn.Create(costCenter).Error)
```

easyCZ (approved):
/hold for last change

components/usage/pkg/db/cost_center_test.go
laushinka: Committed, thanks!

Issue #12169, opened 2022-08-17
Implement EntitlementServiceUBP.hasPaidSubscription


geropl:
This methods needs to be implemented, and respect either the user's personal subscription as well as team membership.

Could actually reused BillingModes.getBillingModeForUser if we would set the "BillingMode.paid" field for users as well.

PR #12305, opened 2022-08-23
[server] Implement EntitlementServiceUBP.hasPaidSubscription

jankeromnes:
## Description

- [x] Implement `hasPaidSubscription` logic in `EntitlementServiceUBP`.

Drive-by fix:
- [x] No longer "force" "more resources" in usage-based billing -- users can choose their resources, so we can deprecate `userGetsMoreResources` with usage-based billing (but it's still used by Chargebee customers)

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/12169

## How to test

1. Create a team called "Gitpod Something" to **enable the Usage-Based flag** for your user and team
2. Don't add a credit card just yet
3. ~~Try starting multiple workspaces in parallel -- at **4 parallel workspaces**, you should start seeing a paywall that prevents you from starting more (until the currently-running workspaces are stopped again)~~
4. ~~Once a workspace is running, you should **not be able to extend** the workspace timeout (i.e. no "clock" button in the IDE bottom bar)~~
5. Now, upgrade your team to paid by adding a credit card
6. Once upgraded, you should now be able to start up to **16 parallel workspaces** before seeing a paywall
7. You should also **be able to extend** the workspace timeout (using the "clock" button in the IDE bottom bar)

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/server/ee/src/billing/entitlement-service-ubp.ts
+++ b/components/server/ee/src/billing/entitlement-service-ubp.ts
@@ -95,8 +95,12 @@ export class EntitlementServiceUBP implements EntitlementService {
         }
     }
 
+    /**
+     * DEPRECATED: With usage-based billing, users can choose exactly how many resources they want to get.
+     * Thus, we no longer need to "force" extra resources via the `userGetsMoreResources` mechanism.
+     */
     async userGetsMoreResources(user: User, date: Date = new Date()): Promise<boolean> {
-        return this.hasPaidSubscription(user, date);
+        return false;
     }
 
     /**
--- a/components/server/ee/src/billing/entitlement-service-ubp.ts
+++ b/components/server/ee/src/billing/entitlement-service-ubp.ts
@@ -4,8 +4,9 @@
  * See License-AGPL.txt in the project root for license information.
  */
 
-import { UserDB } from "@gitpod/gitpod-db/lib";
+import { TeamDB, UserDB } from "@gitpod/gitpod-db/lib";
 import {
+    Team,
     User,
     WorkspaceInstance,
     WorkspaceTimeoutDuration,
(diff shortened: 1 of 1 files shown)


components/server/ee/src/billing/entitlement-service-ubp.ts
```diff
@@ -107,7 +110,30 @@ export class EntitlementServiceUBP implements EntitlementService {
...
+                if (subscriptionId) {
+                    teamSubscriptionIds.push(subscriptionId);
+                }
+            }),
+        );
+        if (teamSubscriptionIds.length > 0) {
+            return true;
+        }
+        // Paid user?
+        const customer = await this.stripeService.findCustomerByUserId(user.id);
```
geropl, line 132: :thought_balloon: Can we do this the other way around: check personal first, and then go to teams? The motivation is to try to reduce the number of checks required.

:cloud: Another idea: This impl checks all teams, where we could try to return after the first positive already. Maybe there is a simple way to get there? :thinking:

jankeromnes: Good ideas! 👍 Done and done. ✅

jankeromnes:
Hmm, I thought we'd be able to test the Usage-Based free tier, but since it's not implemented, we can't:

Screenshot 2022-08-30 at 09 18 04

Also, unrelatedly, while opening several workspaces, I got a few of these error pages too. Looks like this retryAfter mechanism isn't properly supported in `CreateWorkspace`:

Screenshot 2022-08-30 at 09 19 35

jankeromnes:
Hmm, I wasn't able to extend my workspace timeout (no clock button) 🤔 I guess this isn't quite ready yet & needs more debugging.

AlexTugarev (approved):
👍🏻  to unblock!

Maybe it would be good to simplify the promise races here ;-)

/hold

jankeromnes:
Thanks! Good suggestion -- adjusted the code, and tested again, still works as intended. 🚀

/unblock

geropl:
/unhold

I think that's what @jankeromnes wanted to do [here](https://github.com/gitpod-io/gitpod/pull/12305#issuecomment-1233931744).

Issue #7555, opened 2022-01-11
[Dashboard] refactor the CodeText utility component


trumbitta:
As detailed in the https://github.com/gitpod-io/gitpod/pull/7302#pullrequestreview-847622451 thread, the `CodeText`utility component seems to need a refactor replacing the `<span>` tag with a `<code>` tag which is the right semantic choice for such a component.

Beware:

- Are there any cases in the codebase where a `<span>` tag is actually the best choice?
- Will this change work seamlessly? Better check each occurrence.

jankeromnes:
Thanks for opening this issue! [100% agree with you](https://github.com/gitpod-io/gitpod/pull/7302#discussion_r790555966).

Additionally, instead of having a single-line React component with a bunch of CSS classes (that everyone might forget to use or not know about), I think we should just use `<code>` everywhere (more semantic and "natural"), and apply our custom CSS classes to the `code` element in [index.css](https://github.com/gitpod-io/gitpod/blob/main/components/dashboard/src/index.css) (so that they apply everywhere consistently).

PR #12306, opened 2022-08-23
[dashboard] Simplify <code> snippets by using CSS instead of a whole component

jankeromnes:
## Description

Simplify `<code></code>` snippets by using CSS instead of a whole extra component.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/7555

## How to test

1. Open `/settings`
2. The `<code>` snippets should look the same as before

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


jankeromnes:
| BEFORE | AFTER |
| --- | --- |
| Screenshot 2022-08-23 at 16 34 08 | Screenshot 2022-08-23 at 16 34 19 |
| Screenshot 2022-08-23 at 16 34 48 | Screenshot 2022-08-23 at 16 34 58 |
| Screenshot 2022-08-23 at 16 35 23 | Screenshot 2022-08-23 at 16 35 36 |

gtsiolis:
Looking at this now! 👀

gtsiolis (approved):
Thanks for another clean-up, @jankeromnes! 🧹 

Although this is in a different direction from the initial discussion in https://github.com/gitpod-io/gitpod/pull/7302#discussion_r781436360, I like the boring (simple) solution of using the `<code>` element for now. 🤝 

[FWIW](https://en.wiktionary.org/wiki/FWIW), the `CodeText` component aimed at becoming a flexible solution for supporting a) single-line, b) multi-line, and c) inline code text where users could copy to clipboard via a clipboard button or by simply clicking on the inline code. 

For example, to provide some more context, a user could: 
1. Go to `/account` and click directly on that `GIT_AUTHOR_NAME` variable to copy-to-clipboard.
2. Go to `/admin/settings` and click on a copy to clipboard button to copy the telemetry code.
3. ...

However, let's leave this as is for now and come back to this once we know we want to improve the interaction with the code text across the dashboard.

geropl (approved):
LGTM

Issue #12295, opened 2022-08-23
[ws-daemon] panics with `integer divide by zero` err


Pothulapati:
### Bug description

ws-daemon seems to have randomly panic'ed with the following logs

```
panic: runtime error: integer divide by zero

goroutine 20657432 [running]:
github.com/gitpod-io/gitpod/ws-daemon/pkg/iws.getCpuResourceInfoV2({0xc0002a9020?, 0xc1d919dd40?}, {0xc0926d0540?, 0x12bbdf3?})
	github.com/gitpod-io/gitpod/ws-daemon/pkg/iws/iws.go:1121 +0x1b2
github.com/gitpod-io/gitpod/ws-daemon/pkg/iws.getWorkspaceResourceInfo({0xc0002a9020, 0x11}, {0xc0926d0540, 0xbf}, 0x24?)
	github.com/gitpod-io/gitpod/ws-daemon/pkg/iws/iws.go:1055 +0x3a
github.com/gitpod-io/gitpod/ws-daemon/pkg/iws.(*InWorkspaceServiceServer).WorkspaceInfo(0xc135bde460, {0x2321520, 0xc02a409bf0}, 0xc02a409c20?)
	github.com/gitpod-io/gitpod/ws-daemon/pkg/iws/iws.go:1042 +0x8b6
github.com/gitpod-io/gitpod/ws-daemon/api._InWorkspaceService_WorkspaceInfo_Handler.func1({0x2321520, 0xc02a409bf0}, {0x1db82c0?, 0xc02a409c20})
	github.com/gitpod-io/gitpod/ws-daemon/api@v0.0.0-00010101000000-000000000000/workspace_daemon_grpc.pb.go:432 +0x78
github.com/gitpod-io/gitpod/ws-daemon/pkg/iws.ratelimitingInterceptor.UnaryInterceptor.func1({0x2321520, 0xc02a409bf0}, {0x1db82c0, 0xc02a409c20}, 0xc02f24c660, 0xc1f44fdbd8)
	github.com/gitpod-io/gitpod/ws-daemon/pkg/iws/iws.go:1349 +0x2e6
github.com/gitpod-io/gitpod/ws-daemon/api._InWorkspaceService_WorkspaceInfo_Handler({0x1ef5e20?, 0xc135bde460}, {0x2321520, 0xc02a409bf0}, 0xc1d919dce0, 0xc14c5d6600)
[...]

PR #12315, opened 2022-08-23
[ws-daemon] Prevent divide by zero error

Furisto:
## Description
Prevent divide by zero error

## Related Issue(s)
Fixes https://github.com/gitpod-io/gitpod/issues/12295

## Release Notes

```release-note
Prevent divide by zero error in workspace info service
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/common-go/cgroups/v2/cpu.go
+++ b/components/common-go/cgroups/v2/cpu.go
@@ -43,7 +43,7 @@ func (c *Cpu) Max() (quota uint64, period uint64, err error) {
 	path := filepath.Join(c.path, "cpu.max")
 	content, err := os.ReadFile(path)
 	if err != nil {
-		return 0, 0, nil
+		return 0, 0, err
 	}
 
 	values := strings.Split(strings.TrimSpace(string(content)), " ")
--- a/components/ws-daemon/pkg/iws/iws.go
+++ b/components/ws-daemon/pkg/iws/iws.go
@@ -7,6 +7,7 @@ package iws
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"math"
@@ -1104,13 +1105,16 @@ func getCpuResourceInfoV2(mountPoint, cgroupPath string) (*api.Cpu, error) {
 	used := cpuUsage / totalTime * 1000
 
 	quota, period, err := cpu.Max()
-	if err != nil {
+	if errors.Is(err, os.ErrNotExist) {
+		quota = math.MaxUint64
+	} else if err != nil {
 		return nil, err
 	}
 
 	// if no cpu limit has been specified, use the number of cores
 	var limit uint64
 	if quota == math.MaxUint64 {
+		// TODO(toru): we have to check a parent cgroup instead of a host resources
 		cpuInfo, err := linuxproc.ReadCPUInfo("/proc/cpuinfo")
 		if err != nil {
 			return nil, err
--- a/components/common-go/cgroups/v2/cpu_test.go
+++ b/components/common-go/cgroups/v2/cpu_test.go
@@ -5,6 +5,8 @@
 package v2
 
 import (
+	"fmt"
+	"math"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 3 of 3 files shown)


kylos101:
:wave: @Furisto hey there, as this already has an issue, we can 1) relate the PR to the issue and 2) relate the issue to the project (I just did those things). Thanks for the quick turnaround!

components/common-go/cgroups/v2/cpu.go
```diff
@@ -43,7 +43,7 @@ func (c *Cpu) Max() (quota uint64, period uint64, err error) {
 	path := filepath.Join(c.path, "cpu.max")
 	content, err := os.ReadFile(path)
 	if err != nil {
-		return 0, 0, nil
+		return 0, 0, err
```
kylos101, line 46: @Furisto can you share how to test? It looks like the code path will [exit here](https://github.com/gitpod-io/gitpod/blob/91fa12d2edf22ef51b89e1a774d8deeb990f751b/components/ws-daemon/pkg/iws/iws.go#L1108) now, instead of trying to do the [division here](https://github.com/gitpod-io/gitpod/blob/91fa12d2edf22ef51b89e1a774d8deeb990f751b/components/ws-daemon/pkg/iws/iws.go#L1121). I ask because I'm not sure what scenario would have caused `os.ReadFile(path)` to return an error in the first place.

Furisto:
@kylos101 I am not sure how exactly this happened either, but returning an error is better than a runtime panic and the error message could help us find out why it is happening.

components/common-go/cgroups/v2/cpu.go
kylos101, line 46: Could you build a preview environment for this PR, and confirm you can start workspaces? 

I wonder if max doesn't exist, briefly, when workspaces are created, until `ws-daemon` applies maximums on a rotational/regular/scheduled basis.

utam0k: The `cpu.max` file is automatically generated when the container is created. It is not generated by us. Perhaps for some reason the container was erased, and this code was executed after the cgroup was cleaned up.

utam0k:
@Furisto I put a commit into your branch for more accurate value handling

components/ws-daemon/pkg/iws/iws.go
```diff
@@ -1104,13 +1105,16 @@ func getCpuResourceInfoV2(mountPoint, cgroupPath string) (*api.Cpu, error) {
...
+	if errors.Is(err, os.ErrNotExist) {
+		quota = math.MaxUint64
+	} else if err != nil {
 		return nil, err
 	}
 
 	// if no cpu limit has been specified, use the number of cores
 	var limit uint64
 	if quota == math.MaxUint64 {
+        // TODO(toru): we have to check a parent cgroup instead of a host resources
```
jenting, line 1117: please gofmt your code

utam0k: ops, fixed

components/common-go/cgroups/v2/cpu.go
jenting: We can't manually update the `cpu.max` file, correct?
If yes, could we use a unit test to test it?

utam0k: @jenting Good point. we can turn off the cgroup directory by hand, but I think just covering it with unit test is a great idea.

jenting (approved):
Nice!

For next, we need to unit test `iws.go` with the mock function.

components/ws-daemon/pkg/iws/iws.go
```diff
@@ -1104,13 +1105,16 @@ func getCpuResourceInfoV2(mountPoint, cgroupPath string) (*api.Cpu, error) {
...
+	if errors.Is(err, os.ErrNotExist) {
+		quota = math.MaxUint64
+	} else if err != nil {
 		return nil, err
 	}
 
 	// if no cpu limit has been specified, use the number of cores
 	var limit uint64
 	if quota == math.MaxUint64 {
+		// TODO(toru): we have to check a parent cgroup instead of a host resources
```
aledbf, line 1117: @utam0k this logic is wrong. What we should do here is to set the lower limit of the class, and if that value is not set, use the `cpuLimits` value from `ws-daemon`

utam0k: @aledbf Thanks for your review. I have a question. this method is used for `gp top`, right? I think this method doesn't relate to the CPU limiting. Please let me know if I am wrong.

Furisto: This method is not involved in cpu limiting, it is only used for the endpoint that gp top/supervisor is calling to get resource information.

PR #12335, opened 2022-08-24
[content-service] Implement UsageReportService.DownloadURL

easyCZ:
## Description

Implements DownloadURL rpc. This is needed to be able to download a usage report form the usage component.

## Related Issue(s)

* Depends on https://github.com/gitpod-io/gitpod/pull/12300

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/content-service/pkg/service/usage-report-service.go
+++ b/components/content-service/pkg/service/usage-report-service.go
@@ -6,7 +6,7 @@ package service
 
 import (
 	"context"
-
+	"errors"
 	"github.com/opentracing/opentracing-go"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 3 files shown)


components/content-service/pkg/storage/storage.go
```diff
@@ -32,7 +32,7 @@ const (
 
 var (
 	// ErrNotFound is returned when an object is not found
-	ErrNotFound = xerrors.Errorf("not found")
+	ErrNotFound = fmt.Errorf("not found")
```
easyCZ, line 35: Drive-by: `xerrors.Errorf` is deperacted

components/content-service/pkg/service/usage-report-service.go
```diff
@@ -42,20 +42,64 @@ func (us *UsageReportService) UploadURL(ctx context.Context, req *api.UsageRepor
 	span.SetTag("name", req.Name)
 	defer tracing.FinishSpan(span, &err)
 
-	err = us.s.EnsureExists(ctx, us.bucketName)
+	if req.GetName() == "" {
+		return nil, status.Error(codes.InvalidArgument, "Name is required but got empty.")
+	}
+
+	logger := log.WithField("name", req.Name).
```
sagor999, line 49: usually we just name the variable `log` in cases like this. 
```suggestion
	log := log.WithField("name", req.Name).
```

sagor999 (approved):

components/content-service/pkg/service/usage-report-service.go
easyCZ: `log` is the package name where the logging is implemented. It's a name clash that we shouldn't allow.

easyCZ:
dependency landed
/unhold

PR #12173, opened 2022-08-17
[installer]: remove all podsecuritypolicies

MrSimonEmms:
## Description

This is deprecated in Kubernetes 1.21 and removed in 1.25 and it's not really used in this setup

## Related Issue(s)

Fixes #6176

## How to test

Deploy via KOTS

## Release Notes

```release-note
[installer]: remove all podsecuritypolicies
```

## Werft options:

- [ ] /werft with-preview

PR #12336, opened 2022-08-24
[installer]: conditionally remove pod security policies

### LINKED ISSUE #12314: Remove all PodSecurityPolcies, but allow for them to remain with experimental config
An [incident](https://gitpod.slack.com/archives/C03UQ58RG0M) was caused by #12173 - despite the testing, the SaaS cluster was configured to have the PodSecurityPolcies enabled and the GKE testing done did not (_mea culpa_ - I thought it did).

As Kubernetes v1.25 removes support for PodSecurityPolicies, it is not an option to maintain these in the Installer for the long-term.

The proposal is to add a `usePodSecurityPolicies` bool in the `experimental` section of the config (defaulting to `false`) and to conditionally remove the resources removed in #12173.

As this is a deprecated feature in Kubernetes, the bool should be marked as deprecated from the start

MrSimonEmms:
## Description

This is similar to #12173 for most users, however the current Gitpod.io clusters still have pod security policies enabled. It was decided in the incident post-mortem that, as we still need this change for self-hosted users, this change should be conditionally applied via an `experimental` flag.

## Related Issue(s)

Fixes #12314

## How to test

As per #12173. The `use-pod-security-policy` golden files are intended to prove no changes have been made when using the experimental flag.

## Release Notes

```release-note
[installer]: conditionally remove pod security policies
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/install/installer/pkg/components/ws-daemon/clusterrole.go
+++ b/install/installer/pkg/components/ws-daemon/clusterrole.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 
 	"github.com/gitpod-io/gitpod/installer/pkg/common"
+	"github.com/gitpod-io/gitpod/installer/pkg/config/v1/experimental"
 
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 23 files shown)


nandajavarma (approved):
I tested deploying this branch and code in general looks good to me. I will put a /hold so you can merge at your discretion or someone else wants to take a look. Kinda scared since it is a big PR.

install/installer/pkg/components/ws-daemon/clusterrole.go
```diff
@@ -17,33 +18,46 @@ import (
...
+				APIGroups: []string{"policy"},
+				Resources: []string{"podsecuritypolicies"},
+				Verbs:     []string{"use"},
+				ResourceNames: []string{
+					fmt.Sprintf("%s-ns-privileged-unconfined", ctx.Namespace),
+				},
+			})
+		}
+		return nil
+	})
```
jenting, line 35: nit: we could consider making it as a function since there are lots of components with similar code.

jenting (commented):
Code LGTM

jenting (approved):

PR #12337, opened 2022-08-24
[usage] Proto changes for GetCostCenter RPC

laushinka:
## Description
Proto changes for the GetCostCenter RPC

## Related Issue(s)

Relates #12264

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


components/usage-api/usage/v1/usage.proto
```diff
@@ -66,3 +69,16 @@ message ReconcileUsageResponse {
...
+    string attribution_id = 1;
+}
+
+message GetCostCenterResponse {
+    CostCenter cost_center = 1;
+}
+
+message CostCenter {
+    string attribution_id = 1;
+    int32 spending_limit = 2;
```
easyCZ, line 83: What's the precision stored in the DB? Is `int32` enough? I'd probably err on the side of caution here and go for int64

easyCZ (approved):
/hold for Q around precision

components/usage-api/usage/v1/usage.proto
laushinka: It's `int(11)` which I understand to be 32 bits?

laushinka: I worry if we set it to `int64` we might truncate something during the conversion? 

I will unhold for now to unblock, but we can discuss again what we should store.

Issue #10572, opened 2022-06-10
[gp cli] gp command to print/show workspace class


csweichel:
## Is your feature request related to a problem? Please describe
We recently [added an environment variable](https://github.com/gitpod-io/gitpod/pull/10562) which carries the name of the workspace class a workspace is running on.

## Describe the behaviour you'd like
Environment variables are not always stable - and we are trying to educate folks to not rely on them. Instead, we should plug this into the `supervisor` info endpoint, and add a `gp` command to access it.

akosyakov:
I think the logical place will be `gp top`. It should show workspace class, its configuration for CPU and MEMORY, usage, and provide hints to upgrade if necessary.

To consume programatically we can provide an option like `gp top --json`.

mustard-mh:
x-posting suggestions of `gp top` here https://github.com/gitpod-io/gitpod/pull/12338#issuecomment-1227696754 https://github.com/gitpod-io/gitpod/pull/12338#pullrequestreview-1086514903

cc @filiptronicek @andreafalzetti 

Another reason to add `class description` in `gp top` is, sometimes the limit of resources in `gp top` is not align to the workspace's limit but limit in a time window

akosyakov:
I think we maybe will need to have another command to get different info about workspace, that people can rely on it instead of env var as now. But then it should be rather one command like `gp into (workspaceId|instanceId|workspaceClass|...)?` I think there is already an issue to deprecate evn var and we could do it as a part later.

PR #12338, opened 2022-08-24
[gp-cli] support workspace class in top command

mustard-mh:
## Related Issue(s)

Fixes #

## Release Notes

```release-note
```

## Werft options:

- [ ] /werft with-preview


roboquat:

<details>

</details>

components/supervisor-api/status.proto
```diff
@@ -236,6 +236,8 @@ message ResourcesStatusResponse {
     ResourceStatus memory = 1;
     // Used CPU and limit in millicores.
     ResourceStatus cpu = 2;
+    // Class of current workspace
+    string workspace_class=3;
```
akosyakov, line 240: we should add it to info endpoint

components/gitpod-cli/cmd/top.go
```diff
@@ -22,7 +23,12 @@ import (
 	"github.com/olekukonko/tablewriter"
 )
 
+var opts struct {
+	Json bool
+}
+
 func outputTable(workspaceResources *supervisor.ResourcesStatusResponse) {
+	fmt.Printf("Workspace Class: %s\n", workspaceResources.WorkspaceClass)
```
akosyakov, line 31: It would be nice also to print its configuration in human readable form, i.e. minimal and max CPU. To avoid confusion why CPU limit is shifting. Not sure how to present it, some suggested to put it as limit instead of real cgroup limit, but it can be confusing if a user did not manage to get it.

I remember @Furisto mentioned that we can pass it to supervisor from worksapcekit via env var. Maybe can be another PR as well.

mustard-mh:
For comment https://github.com/gitpod-io/gitpod/pull/12338#issuecomment-1225798958 above 

cc @loujaybee @akosyakov

components/supervisor/cmd/top.go
```diff
@@ -52,7 +52,7 @@ var topCmd = &cobra.Command{
 			tw := tabwriter.NewWriter(os.Stdout, 6, 4, 3, ' ', 0)
 			defer tw.Flush()
 
-			fmt.Fprintf(tw, "CPU(millicores)\tMEMORY(bytes)\n")
+			fmt.Fprintf(tw, "Workspace Class: %s\nCPU(millicores)\tMEMORY(bytes)\n", status.WorkspaceClass)
```
akosyakov, line 55: it can be empty.

akosyakov:
> How to ask if they want to upgrade their plan (when CPU/MEM is nearly full?)

I don't think we need to ask. Maybe just add a link to the plan page for now? cc @loujaybee

components/gitpod-cli/cmd/top.go
akosyakov, line 31: It can be empty. Actually for now for most users it will be empty.

mustard-mh: I thought it have default value `default`? not looked into the codebase but try with a new account

akosyakov: We need to clarify it. We should show here `Standard`, `Large` or nothin if no workspace class. We should also return human readable representation from info endpoint as well, so all clients display the same, i.e. somethign like:
```
{
  workspaceClass: {
    id: string
    name: string
  }
}

Comment (akosyakov) [reply components/gitpod-cli/cmd/top.go:31]: @kylos101 @Furisto Could you clarify which values workspace_class can have now and what is human readable name for each please? 🙏

Comment (mustard-mh) [reply components/gitpod-cli/cmd/top.go:31]: supervisor responds with id and ask for details from server in gp-cli? 

https://github.com/gitpod-io/gitpod/blob/9a7119d05dbb721cfd7151a8dd3e68e00b3439ea/components/gitpod-protocol/src/gitpod-service.ts#L317

Comment (akosyakov) [reply components/gitpod-cli/cmd/top.go:31]: nice 👍

Comment (akosyakov) [reply components/gitpod-cli/cmd/top.go:31]: Maybe then show something like:
`Standard: Up to 4 vCPU, 8GB memory, 30GB disk`

I'm not sure now whether supervisor should return anything besides workspace class id, since clients can fetch rest as needed from Gitpod API already.

Comment (akosyakov) [reply components/gitpod-cli/cmd/top.go:31]: for `--json` i would assume it should be `worksapceClass: g1-standard`? cc @Furisto

Comment (Furisto) [reply components/gitpod-cli/cmd/top.go:31]: > @kylos101 @Furisto Could you clarify which values workspace_class can have now and what is human readable name for each please? pray

The workspace classes are g1-standard (Standard is the display name), g1-large (Large), default (Standard (old)) and gitpodio-internal-xl (XL (old)). After switching to workspace classes completely default and gitpodio-internal-xl will go away.

> for --json i would assume it should be worksapceClass: g1-standard? cc @Furisto

It should be the display name so 'Standard'. 

Would it help if we changed the WORKSPACE_CLASS environment variable, so that it contains a json document with the information that you need?

Comment (akosyakov) [reply components/gitpod-cli/cmd/top.go:31]: > Would it help if we changed the WORKSPACE_CLASS environment variable, so that it contains a json document with the information that you need?

I think it would be better that we use getSupportedWorkspaceClasses that UI is consistent across all interfaces including dashboard. Internally in supervisor it is alright to know the technical name.

Comment (mustard-mh) [reply components/gitpod-cli/cmd/top.go:31]: `getSupportedWorkspaceClasses` in golang is not supported yet (not generated), I'll create another PR to add it and ask WebApp team to review

File: components/supervisor-api/info.proto [hunk 4]
```diff
@@ -78,4 +78,26 @@ message WorkspaceInfoResponse {
 
     // ide_port is the port on which the IDE is to be run
     uint32 ide_port = 14;
+
+    message WorkspaceClass {
```
akosyakov, line 82: let's not copy all api from server, it is rather internal thing, and here we are quite public, only add what we really need for now: `id`, `name`, `description`

components/gitpod-cli/cmd/top.go
```diff
@@ -22,6 +24,75 @@ import (
...
+			fmt.Println(string(content))
+			return
+		}
+		outputWorkspaceClass(wsInfo)
+		outputTable(workspaceResources)
+	},
+}
+
+func outputWorkspaceClass(wsInfo *supervisor.WorkspaceInfoResponse) {
+	if wsInfo.WorkspaceClass.DisplayName == "" {
```
akosyakov, line 89: i won't show anything if there is no name

components/supervisor/pkg/supervisor/services.go
```diff
@@ -692,6 +704,38 @@ func (is *InfoService) WorkspaceInfo(context.Context, *api.WorkspaceInfoRequest)
 	return resp, nil
 }
 
+// getWorkspaceClass of current workspace
+func (is *InfoService) getWorkspaceClass(ctx context.Context) *api.WorkspaceInfoResponse_WorkspaceClass {
+	if is.gitpodService == nil {
+		gitpodService, err := createServerService(is.cfg, is.tokenService, []string{"function:getSupportedWorkspaceClasses"})
```
akosyakov, line 710: we should not do it, service should be provided from supervisor.go if it is missing then no workspace class

there should be only one place where we create server service

components/supervisor/pkg/supervisor/services.go
```diff
@@ -623,8 +624,11 @@ func (rt *remoteTokenProvider) GetToken(ctx context.Context, req *api.GetTokenRe
 
 // InfoService implements the api.InfoService.
 type InfoService struct {
-	cfg          *Config
-	ContentState ContentState
+	cfg            *Config
+	gitpodService  *gitpod.APIoverJSONRPC
+	tokenService   *InMemoryTokenService
+	workspaceClass *api.WorkspaceInfoResponse_WorkspaceClass
```
akosyakov, line 630: should not it be protected by mutex to prevent concurrent requests to server?

akosyakov: Maybe we should rather resolve it in supervisor.go and pass in services.go. It will be cleaner and does not need all this synchronisation.

akosyakov: We can also consider to pass this info from env var from server on startup instead to avoid call at all.

components/supervisor/pkg/supervisor/supervisor.go
```diff
@@ -592,37 +592,17 @@ func installDotfiles(ctx context.Context, cfg *Config, tokenService *InMemoryTok
...
-	tknres, err := tknsrv.GetToken(context.Background(), &api.GetTokenRequest{
-		Kind: KindGitpod,
-		Host: host,
-		Scope: []string{
-			"function:getToken",
-			"function:openPort",
-			"function:getOpenPorts",
-			"function:guessGitTokenScopes",
-		},
+	gitpodService, err := createServerService(cfg, tknsrv, []string{
```
akosyakov, line 595: please revert it, it should be the only place to create

components/supervisor/pkg/supervisor/services.go
mustard-mh: If we use the same server service, `git` token provider will not work because `function:getSupportedWorkspaceClasses` scope may not exists if server rollback.

akosyakov: Let's provide info via env var like here [[1](https://github.com/gitpod-io/gitpod/blob/d3147bcb557cb7db1a3815f11239624f58335cc5/components/server/src/workspace/workspace-starter.ts#L1339-L1342)], i.e. pass `GITPOD_WORKSPACE_CLASS`? and then you will have it via `config.WorkspaceClass` if env var is missing we just return nil and client should respect it

No need for additional remote call or caching.

components/supervisor/go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/gitpod-io/gitpod/supervisor
 go 1.18
 
 require (
-	github.com/Netflix/go-env v0.0.0-20200908232752-3e802f601e28
+	github.com/Netflix/go-env v0.0.0-20220526054621-78278af1949d
```
mustard-mh, line 6: upgrade pkg to support with custom env parse

filiptronicek (commented):
Can we omit either the short or long name of the class in the human-readable output?

Reading the output below, it just doesn't read well.

```
Default: Default workspace class
```

I would like to have one of these ommited, either like this:
```
Workspace class: Default
```

or using the full name like this: 
```
Default workspace class
```

filiptronicek:
suggestion (non-blocking): add an option to disable this

Because you have to always actively change the workspace class, wouldn't it make sense to have an option to disable this entirely? Maybe if you had an environment variable in your dotfiles like `GP_TOP_SHOW_CLASS`, it would make disabling per user possible which could be useful for some users.

I can see myself just using either class for most tasks and very rarely changing it, so for me `gp top` would just look bloated with info I do not necessarily need.

Maybe we can discuss this after this PR though and can implement later.

mustard-mh:
> Can we omit either the short or long name of the class in the human-readable output?
> 
> Reading the output below, it just doesn't read well.
> 
> ```
> Default: Default workspace class
> ```
> 
> I would like to have one of these ommited, either like this:
> 
> ```
> Workspace class: Default
> ```
> 
> or using the full name like this:
> 
> ```
> Default workspace class
> ```

@filiptronicek  In production, id `default` will display with `Standard (old): Up to 6 vCPU, 12GB memory, 30GB disk`

`Default: Default workspace class` is formatted by `${displayName}: ${description}`

components/gitpod-cli/cmd/top.go
```diff
@@ -57,23 +127,8 @@ func getColor(severity api.ResourceStatusSeverity) int {
...
-			log.Fatalf("cannot get workspace resources: %s", err)
-		}
-
-		outputTable(workspaceResources)
-	},
-}
-
 func init() {
 	topCmd.Flags().BoolVarP(&noColor, "no-color", "", false, "Disable output colorization")
+	topCmd.Flags().BoolVarP(&opts.Json, "json", "j", false, "print like json")
```
andreafalzetti, line 132:
```suggestion
	topCmd.Flags().BoolVarP(&opts.Json, "json", "j", false, "output in json format")
```

andreafalzetti: I would suggest `output in json format` or `output data as json`

components/server/src/workspace/workspace-starter.ts
```diff
@@ -1341,6 +1341,14 @@ export class WorkspaceStarter {
         contextEnv.setValue(JSON.stringify(workspace.context));
         envvars.push(contextEnv);
 
+        const info = this.config.workspaceClasses.find((cls) => cls.id === instance.workspaceClass);
+        if (!!info) {
```
andreafalzetti, line 1345: why do we need the `!!`?

andreafalzetti (approved):
Code LGTM 👍

Nice changes @mustard-mh 🙏 I see room of improvements in the future in regards to what @filiptronicek suggested and around the output (personally I am not 100% convinced about the table format mixed with plain text) - but I think we should ship this as it is now :)

I love the `--json` flag 😍 

/hold 

in case you want to make any final change

components/server/src/workspace/workspace-starter.ts
filiptronicek: I'd agree here, `if` is already checking truthyness / falsyness of the expression [[1](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/if...else)], so we are basically doing the boolean conversion twice.

components/gitpod-cli/cmd/top.go
filiptronicek: I would use uppercase JSON, since it is an abbreviation [[1](https://www.json.org/json-en.html)]

```suggestion
	topCmd.Flags().BoolVarP(&opts.Json, "json", "j", false, "output in JSON format")
```

Or even make it super-short

```suggestion
	topCmd.Flags().BoolVarP(&opts.Json, "json", "j", false, "output JSON")
```

filiptronicek:
> @filiptronicek In production, id default will display with Standard (old): Up to 6 vCPU, 12GB memory, 30GB disk. You can see all classes in cm server-config

Thanks for pointing this out, Huiwen. I am having some issues with the command you mentioned but I don't think it matters, it makes sense in how you implemented this.

components/gitpod-cli/cmd/top.go
```diff
@@ -22,6 +24,74 @@ import (
...
+			if err != nil {
+				log.Fatalf("cannot get workspace resources: %s", err)
+			}
+			wg.Done()
+		}()
+
+		go func() {
+			wsInfo, err = supervisor.NewInfoServiceClient(conn).WorkspaceInfo(ctx, &supervisor.WorkspaceInfoRequest{})
+			if err != nil {
+				log.Fatalf("cannot get workspace info: %s", err)
```
akosyakov, line 64: we should not fail top command in this case

components/supervisor/pkg/supervisor/config.go
```diff
@@ -182,6 +205,12 @@ type WorkspaceConfig struct {
 	// WorkspaceUrl is an URL for which workspace is accessed.
 	WorkspaceUrl string `env:"GITPOD_WORKSPACE_URL"`
 
+	// WorkspaceClass denotes the class of the workspace
+	WorkspaceClass string `env:"GITPOD_WORKSPACE_CLASS"`
```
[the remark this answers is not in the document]
akosyakov: ah got it

akosyakov:
@Furisto Could you review it please?

mustard-mh:
Fource push to

- Fix `workspaceClassInfo` not exists ( try with previous commit and run `go run main.go top` in gitpod-cli folder will get it)
- Change `json` flag's description

See diff [here](https://github.com/gitpod-io/gitpod/compare/114bd2d3f6dad16931b657854f663a7aa5ae3ae9..6f57d8053e77d404113d9f4fc5973ab4185bcb94)

akosyakov:
> suggestion (non-blocking): add an option to disable this

I think it is important for people to know why they have such limits. I don't see one line as a bloat.

filiptronicek:
> > suggestion (non-blocking): add an option to disable this
> 
> I think it is important for people to know why they have such limits. I don't see one line as a bloat.

@akosyakov touché. I meant with it that I personally would not find this much useful, and a one-off command like mentioned in https://github.com/gitpod-io/gitpod/issues/10572#issuecomment-1228712900.

jenting (approved):
LGTM

Furisto (approved):

akosyakov (approved):
lgtm

akosyakov:
@mustard-mh You can unhold it?

components/server/src/workspace/workspace-starter.ts
andreafalzetti: @mustard-mh did you see these comments?

mustard-mh: > @mustard-mh did you see these comments?

@andreafalzetti  Thanks for your reminder 🙏 . I have saw this, but want left it to webApp team to answer it 😂, I use `!!` because I see it exists in some typescript code

image

svenefftinge (approved):

PR #12339, opened 2022-08-24
[usage] Test for complex usage cap scenarios

geropl:
## Related Issue(s)

Fixes #

## Release Notes

```release-note
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage/pkg/apiv1/billing_test.go
+++ b/components/usage/pkg/apiv1/billing_test.go
@@ -125,3 +125,190 @@ func TestFinalizeInvoice(t *testing.T) {
 
 	require.NoError(t, err)
 }
+
+func TestCapUsageToSpendingLimit(t *testing.T) {
+	type ReportSummary struct {
+		// These are the inputs
+		CalculatedUsage int32
+		SpendingLimit   int32
+		// These are the outputs (also depending on the previous run)
+		EffectiveSpendingLimit int32
+		CappedUsage            int32
+	}
+
+	cap := func(v, limit int32) int32 {
+		if v > limit {
+			return limit
+		}
+		return v
+	}
+
+	max := func(a, b int32) int32 {
+		if a >= b {
+			return a
+		} else {
+			return b
+		}
+	}
+
+	min := func(a, b int32) int32 {
+		if a <= b {
+			return a
+		} else {
+			return b
+		}
+	}
+
+	/**
+	 * Takes two ReportSummaries, both from the _same invoice_ (!).
+	 */
+	capUsage := func(previous, current ReportSummary) (cappedUsage int32, effectiveLimit int32) {
+		// idea: if a user reduces spending limit during a billing cycle, we should:
+		//  - allow to reduce that threshold
+		//  - but not below what we currently plan to bill them
+		// This allows us to a) fix reported usage and b) avoids the exploit case.
+		lowerBoundEffectiveSpendingLimit := min(previous.EffectiveSpendingLimit, current.CalculatedUsage)
+		effectiveLimit = max(current.SpendingLimit, lowerBoundEffectiveSpendingLimit)
+		cappedUsage = cap(current.CalculatedUsage, effectiveLimit)
+		return cappedUsage, effectiveLimit
+	}
+
+	type Test struct {
+		Name    string
+		Reports []ReportSummary
+	}
+
+	tests := []Test{
+		{
+			Name: "simple test",
+			Reports: []ReportSummary{{
+				CalculatedUsage:        95,
+				SpendingLimit:          100,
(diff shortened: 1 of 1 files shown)


roboquat:
@geropl: Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://git.k8s.io/community/contributors/guide/release-notes.md) to remove it.

<details>

</details>

geropl:
@laushinka This PR has the commit with the capping that fixes the more advanced cases.

We basically always need the previous `ReportSummary` and the current.calculatedUsage and current.spendingLimit, and can calculate the rest. All the fields we need from previous we should store on/retrieve from the "upcoming invoice".

Does that make sense?

components/usage/pkg/apiv1/billing_test.go
```diff
@@ -125,3 +125,156 @@ func TestFinalizeInvoice(t *testing.T) {
...
+	}
+
+	/**
+	 * Takes two ReportSummaries, both from the _same invoice_ (!).
+	 */
+	capUsage := func(previous, current ReportSummary) (cappedUsage int32, effectiveLimit int32) {
+		// idea: if a user reduces spending limit during a billing cycle, we should:
+		//  - allow to reduce that threshold
+		//  - but not below what we currently plan to bill them
+		// This allows us to a) fix reported usage and b) avoids the exploit case.
```
easyCZ, line 169: How does it solve the following case?

1. User has spending limit 25
2. User reaches spending limit and cannot start a new workspace - they use up 26 units
3. User cancels their subscription
4. User re-subscribes before instances from previous subscription are marked as billed for (invoice finalization)
5. We run the controller and compute usage. We observe 26, cap it at 25
6. We update the new invoice with 25
7. The finalization of the invoice happens
8. On next run, we exclude instances from the previous run and identify the user has used up 1 credit
9. We cannot update the invoice to 1, because it's already at 25

geropl: Can you turn that into a test case? If not I can have a look later.

laushinka:
Taking a look! Tagging @easyCZ too here re the Report shape.

components/usage/pkg/apiv1/billing_test.go
laushinka: > 8. On next run, we exclude instances from the previous run and identify the user has used up 1 credit
> 9. We cannot update the invoice to 1, because it's already at 25

This is where I'm still processing about how this case can come to be.
With the finalization of the invoice, won't the user have been billed for the 25 usage?

geropl: Will try to turn this into a meaningful test case now. :+1:

geropl: @easyCZ I _tried_ to capture it [here](https://github.com/gitpod-io/gitpod/blob/f4d8814b1cf6c253a5953f96f37234f7f7b473c1/components/usage/pkg/apiv1/billing_test.go#L259-L291).

Doesn't mean it handle _all_ cases, but this one seems to work. :eyes: 

`cd components/usage && go test -v ./... -run TestCapUsageToSpendingLimit`

laushinka:
There was one case missing: 

```
CalculatedUsage:        150,
SpendingLimit:          130,
EffectiveSpendingLimit: 100,
CappedUsage:            ?,
```

In this case the UpcomingInvoice < SpendingLimit < CalculatedUsage.
What happens here?

components/usage/pkg/apiv1/billing_test.go
```diff
@@ -125,3 +125,190 @@ func TestFinalizeInvoice(t *testing.T) {
...
+		}
+	}
+
+	/**
+	 * Takes two ReportSummaries, both from the _same invoice_ (!).
+	 */
+	capUsage := func(previous, current ReportSummary) (cappedUsage int32, effectiveLimit int32) {
+		// idea: if a user reduces spending limit during a billing cycle, we should:
+		//  - allow to reduce that threshold
+		//  - but not below what we currently plan to bill them
```
laushinka, line 168: I made the assumption that `EffectiveSpendingLimit` in these test cases is referring to "what we currently plan to bill them" aka `upcomingInvoice`. Is that correct?

geropl: @laushinka No, that's `cappedUsage`. Sorry for not making that clearer: It's a value that we retrieve from the metadata of the `upcomingInvoide`, prior to actually updating it. It's a value we carry over (and manipulate) with every run.

geropl:
Closing this PR as it's ouf of focus right now.

PR #12352, opened 2022-08-24
[ws-daemon] Check if the shiftfs module already exists

aledbf:
## Description

Disable build of `shiftfs` module if the kernel in the node already provides it.

## Release Notes
```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


MrSimonEmms:
I like this - simple

sagor999 (approved):

PR #12360, opened 2022-08-24
[ws-daemon] when WaitForInit ensure that state is indeed ready

### LINKED ISSUE #11713: [ws-daemon] cannot find workspace during WaitForInit

### LINKED ISSUE #12357: Failed workspace initialization not handled properly
### Bug description

If workspace fails to init due to precondition failure during initializiation, we do not seem to abort such workspace.
It seems like the error is being swallowed and pod is left hanging around for 1 hour, after which is gets killed due to timeout.
Trace attached of initialization failure.
[21821e8c81ca6948.txt](https://github.com/gitpod-io/gitpod/files/9418908/21821e8c81ca6948.txt)
Logs of the affected workspace: https://cloudlogging.app.goo.gl/KWKFpmrtHtsSZNtK6

### Steps to reproduce

n\a

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

sagor999:
## Description

This was quite fun to hunt it down.

When calling WaitForInit we will wait for `operatingCondition` to broadcast to unlock us.
There are two places where broadcast is being done:
`MarkInitDone` that changes `state` to `WorkspaceReady`
`Dispose` that changes `state` to `WorkspaceDisposed`
If workspace fails during `InitWorkspace`, then its state will be disposed. 
We call InitWorkspace during `CREATING` phase, and then we `WaitForInit` in `INITIALIZING` state.
So if InitWorkspace failed, we incorrectly would mark workspace as ready, causing all types of issues down the line.

## Related Issue(s)

Fixes #12357
Fixes #11713

## Release Notes

```release-note
none
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/ws-daemon/pkg/internal/session/workspace.go
+++ b/components/ws-daemon/pkg/internal/session/workspace.go
@@ -122,7 +122,8 @@ func (s *Workspace) WaitForInit(ctx context.Context) (ready bool) {
 
 	s.operatingCondition.L.Lock()
 	s.operatingCondition.Wait()
-	ready = true
+	// make sure that state is indeed ready when done waiting
+	ready = s.state == WorkspaceReady
 	s.operatingCondition.L.Unlock()
 	return
 }


utam0k (approved):
Great insight 💯

jenting:
Amazing 🤩

Issue #12248, opened 2022-08-22
Fix the workspace integration tests


utam0k:
### Bug description

The workspace integration tests continue to fail
Now it failed:

- [x] Pass all tests in a row from local
- [x] Pass all tests in a row from werft
- [x] Ensure branch deleting to avoid remaining many preview env

### Steps to reproduce

Run the integration test

### Expected behavior

The workspace integration tests pass every day 👍 

### Anything else?

[Last failed run](https://gitpod.slack.com/archives/C03E52788SU/p1661162310890499)

kylos101:
Thanks @utam0k ! 

I updated the description to account for some additional recent failures, and added this to Scheduled.

kylos101:
👋 hey @utam0k , may I ask, is there a draft PR for which you can link to this issue? Let us know if you need any help? 🤗

utam0k:
@kylos101 It took me a while to realize I was falling short of preview env's resources, but so far I don't need any help. Thank you.

kylos101:
@utam0k may I ask a favor? Can you alter the script which triggers test runs, so that it deletes prior branches, prior to starting? In this way, we can be sure we do not start too many preview VMs. [Internal reference.](https://gitpod.slack.com/archives/C01KGM9EBD4/p1662793389381669?thread_ts=1662621986.692229&cid=C01KGM9EBD4)

kylos101:
Related: https://github.com/gitpod-io/gitpod/issues/12021#issuecomment-1243069224

utam0k:
> @utam0k may I ask a favor? Can you alter the script which triggers test runs, so that it deletes prior branches, prior to starting? In this way, we can be sure we do not start too many preview VMs. [Internal reference.](https://gitpod.slack.com/archives/C01KGM9EBD4/p1662793389381669?thread_ts=1662621986.692229&cid=C01KGM9EBD4)

[At the end of the integration script](https://github.com/gitpod-io/gitpod/blob/d4fe26f45f7c8ba281bc93e0f9e2d08e2a0d7a0a/.werft/workspace-run-integration-tests.sh#L63) runs delete a branch that the script creates. I wonder why deleting branch code sometimes isn't run. I guess it relates to this issue.
https://github.com/gitpod-io/gitpod/issues/12795

And I have added a workaround for this issue yet. Your suggestions are excellent. However, can we just take a look at the situation a bit?
https://github.com/gitpod-io/gitpod/pull/12793

[...]

PR #12367, opened 2022-08-25
image-builder-bob: Introduce backoff to connect to buildkit

utam0k:
## Description

The retry to connect to the buildkit was too fast. At least in the current preview environment resources sometimes gave up before the buildkit was up and running.

## Related Issue(s)

Relates #12248

## How to test

Pass `TestBaseImageBuild/database/it_should_build_a_base_image` on the preview env
(Maybe the test doesn't pass, but at least it fails not a connection error to buildkit)

## Release Notes

```release-note
More reliably connect to buildkit during image builds
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/image-builder-bob/pkg/builder/builder.go
+++ b/components/image-builder-bob/pkg/builder/builder.go
@@ -251,6 +251,7 @@ func StartBuildkit(socketPath string) (cl *client.Client, teardown func() error,
 }
 
 func connectToBuildkitd(socketPath string) (cl *client.Client, err error) {
+	backoff := 1 * time.Second
 	for i := 0; i < maxConnectionAttempts; i++ {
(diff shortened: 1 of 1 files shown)


components/image-builder-bob/pkg/builder/builder.go
```diff
@@ -251,6 +251,7 @@ func StartBuildkit(socketPath string) (cl *client.Client, teardown func() error,
 }
 
 func connectToBuildkitd(socketPath string) (cl *client.Client, err error) {
+	backoff := 1 * time.Second
```
jenting, line 254: nit: `backoff := time.Second`

jenting (approved):
The `maxConnectionAttempts` is 10, so the times it tries to connect buildkits are 1s, 3s, 7s, 15s, 31s, ...

PR #12376, opened 2022-08-25
[server][dashboard] Allow new Stripe customers to select their billing currency

jankeromnes:
## Description

Allow new Stripe customers to select their billing currency.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/12343

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/dashboard/src/components/ContextMenu.tsx
+++ b/components/dashboard/src/components/ContextMenu.tsx
@@ -35,7 +35,7 @@ function ContextMenu(props: ContextMenuProps) {
         setExpanded(!expanded);
     };
 
-    const handler = (evt: KeyboardEvent) => {
+    const keydownHandler = (evt: KeyboardEvent) => {
         if (evt.key === "Escape") {
             setExpanded(false);
         }
@@ -55,11 +55,11 @@ function ContextMenu(props: ContextMenuProps) {
     };
 
     useEffect(() => {
-        window.addEventListener("keydown", handler);
+        window.addEventListener("keydown", keydownHandler);
         window.addEventListener("click", clickHandler);
         // Remove event listeners on cleanup
         return () => {
-            window.removeEventListener("keydown", handler);
+            window.removeEventListener("keydown", keydownHandler);
             window.removeEventListener("click", clickHandler);
         };
     }, []); // Empty array ensures that effect is only run on mount and unmount
--- a/components/dashboard/src/teams/TeamUsageBasedBilling.tsx
+++ b/components/dashboard/src/teams/TeamUsageBasedBilling.tsx
@@ -6,15 +6,16 @@
 
 import React, { useContext, useEffect, useState } from "react";
 import { useLocation } from "react-router";
+import { BillingMode } from "@gitpod/gitpod-protocol/lib/billing-mode";
 import { Appearance, loadStripe, Stripe } from "@stripe/stripe-js";
 import { Elements, PaymentElement, useElements, useStripe } from "@stripe/react-stripe-js";
 import { getCurrentTeam, TeamsContext } from "./teams-context";
+import DropDown from "../components/DropDown";
 import Modal from "../components/Modal";
 import { ReactComponent as Spinner } from "../icons/Spinner.svg";
 import { PaymentContext } from "../payment-context";
 import { getGitpodService } from "../service/service";
 import { ThemeContext } from "../theme-context";
-import { BillingMode } from "@gitpod/gitpod-protocol/lib/billing-mode";
 
 type PendingStripeSubscription = { pendingSince: number };
 
@@ -22,7 +23,6 @@ export default function TeamUsageBasedBilling() {
     const { teams } = useContext(TeamsContext);
     const location = useLocation();
     const team = getCurrentTeam(location, teams);
-    const { currency } = useContext(PaymentContext);
     const [teamBillingMode, setTeamBillingMode] = useState<BillingMode | undefined>(undefined);
     const [stripeSubscriptionId, setStripeSubscriptionId] = useState<string | undefined>();
     const [isLoading, setIsLoading] = useState<boolean>(true);
@@ -87,7 +87,7 @@ export default function TeamUsageBasedBilling() {
                 JSON.stringify(pendingSubscription),
             );
             try {
-                await getGitpodService().server.subscribeTeamToStripe(team.id, setupIntentId, currency);
+                await getGitpodService().server.subscribeTeamToStripe(team.id, setupIntentId);
             } catch (error) {
                 console.error("Could not subscribe team to Stripe", error);
                 window.localStorage.removeItem(`pendingStripeSubscriptionForTeam${team.id}`);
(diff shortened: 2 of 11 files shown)


components/dashboard/src/teams/TeamUsageBasedBilling.tsx
```diff
@@ -326,6 +334,9 @@ function CreditCardInputForm() {
         }
         setIsLoading(true);
         try {
+            // Create Stripe customer for team & currency (or update currency)
+            await getGitpodService().server.createOrUpdateStripeCustomerForTeam(props.teamId, currency);
+            // FIXME(janx): The above returns too early, when the customer doesn't exist yet? 🤔
```
jankeromnes, line 339: This breaks the first sign-up attempt:
1. Add credit card
2. Fails after reload 💥 (server error "Error: No Stripe customer profile for team '9d19a7ed-0154-43f8-a4c5-2dea787d06d8'")
3. Add credit card again
4. Now it works

jankeromnes: Looks like a race condition:

- This call is completed by `08:11:40.128Z`:

```ts
this.getStripe().customers.create({
    name: userName ? `${userName} (${team.name})` : team.name,
    metadata: {
        teamId: team.id,
    },
});
```

- This call at `08:11:43.554Z` comes back empty (no results):

```ts
this.getStripe().customers.search({
    query: `metadata['teamId']:'${teamId}'`,
});
```

- The same call at `08:15:27.202Z` comes back with one result (the created customer)

Maybe we need to poll Stripe after creating a customer?

jankeromnes: Polling fixed it. ✅

components/dashboard/src/components/ContextMenu.tsx
```diff
@@ -56,11 +56,11 @@ function ContextMenu(props: ContextMenuProps) {
 
     useEffect(() => {
         window.addEventListener("keydown", handler);
-        window.addEventListener("click", clickHandler);
+        window.addEventListener("click", clickHandler, { capture: true });
         // Remove event listeners on cleanup
         return () => {
             window.removeEventListener("keydown", handler);
-            window.removeEventListener("click", clickHandler);
+            window.removeEventListener("click", clickHandler, { capture: true });
```
jankeromnes, line 63: This seems to fix context menus in modals, but has a weird side-effect that you need to click twice to hide a menu again. 🤔

I wonder if this renders the "only skip once" trick obsolete.

jankeromnes: Actually that doesn't really work.

The problem is that the Modal modal calls `e.stopPropagation()` on click events here:

https://github.com/gitpod-io/gitpod/blob/924b20eaee81481451a163f996cf7d7a8e4b51ab/components/dashboard/src/components/Modal.tsx#L82

This prevent any clicks inside the Modal window from propagating up to the `window` and thus from closing the ContextMenu when needed, because this handler never gets called:

https://github.com/gitpod-io/gitpod/blob/924b20eaee81481451a163f996cf7d7a8e4b51ab/components/dashboard/src/components/ContextMenu.tsx#L59

components/server/ee/src/user/stripe-service.ts
```diff
@@ -93,14 +88,33 @@ export class StripeService {
...
+        while (!(await this.findCustomerByTeamId(team.id))) {
+            await new Promise((resolve) => setTimeout(resolve, POLL_CREATED_CUSTOMER_INTERVAL_MS));
+            if (++attempts > POLL_CREATED_CUSTOMER_MAX_ATTEMPTS) {
+                throw new Error(`Could not confirm Stripe customer creation for team '${team.id}'`);
+            }
+        }
+        return customer;
+    }
+
+    async setPreferredCurrencyForCustomer(customer: Stripe.Customer, currency: string): Promise<void> {
```
easyCZ, line 102: How do we ensure that `currency` is a valid value? Does Stripe reject an invalid type of currency?

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2054,22 +2054,37 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
+    async createOrUpdateStripeCustomerForTeam(ctx: TraceContext, teamId: string, currency: string): Promise<void> {
+        const user = this.checkAndBlockUser("createOrUpdateStripeCustomerForTeam");
+        const team = await this.guardTeamOperation(teamId, "update");
+        await this.ensureStripeApiIsAllowed({ team });
+        try {
+            let customer = await this.stripeService.findCustomerByTeamId(team!.id);
+            if (!customer) {
+                customer = await this.stripeService.createCustomerForTeam(user, team!);
+            }
+            await this.stripeService.setPreferredCurrencyForCustomer(customer, currency);
```
easyCZ, line 2066: Would this allow me to change my currency after I've got a subscription in one currency? And if so, is that a problem?

components/server/ee/src/user/stripe-service.ts
jankeromnes: `currency` can be any string. Then, when we choose a Gitpod Product Price in Stripe, we see if one is available in our configuration for that specific string. If there is, great. If not, we pick the `USD` Price.

components/server/ee/src/workspace/gitpod-server-impl.ts
jankeromnes: > Would this allow me to change my currency after I've got a subscription in one currency?

Yes -- you can always cancel your Subscription, then buy a new Subscription in any available currency.

> And if so, is that a problem?

No -- this is a feature. If you're paying in USD, but want to change to EUR, you can.

jankeromnes:
ERRATUM: You actually cannot switch currencies without first recreating a new Stripe customer.

If you subscribe with one currency, then cancel, then upgrade with another currency, Stripe will refuse the upgrade with:

```
Error: You cannot combine currencies on a single customer. This customer has had a subscription, coupon, or invoice item with currency eur
```

easyCZ (approved):
LGTM from a backend side. I don't have as much context on the UI but looks good to me.

/hold in case you want someone else to take a look at UI side

jankeromnes:
Many thanks @easyCZ!

@AlexTugarev had a good follow-up suggestion -- for these cases where subscribing fails, we don't properly display errors in the UI. It would be good to show (maybe not the full Stripe errors, but at least some hint that things went wrong and that you may need to contact support).

AlexTugarev (approved):
LGTM!
The flow to set a currency worked nicely 👏🏻 

We catched https://github.com/gitpod-io/gitpod/pull/12376#issuecomment-1233944378
Can be handled as follow-up

jankeromnes:
Many thanks! 🚀 Let's improve in follow-ups.

/unhold

jankeromnes:
Filed this follow-up issue: https://github.com/gitpod-io/gitpod/issues/12580

PR #12377, opened 2022-08-25
GetUpcomingInvoice for BillingService (2)

### LINKED ISSUE #11692: Implement BillingService.GetLatestInvoice

AlexTugarev:
## Description
This PR implements ~`GetLatestInvoice`~  `GetUpcomingInvoice`. The rename is done intentionally, to a) align with terminology used in stripe, and b) to distinguish between the next invoice and an already paid invoice. The upcoming invoice contains the the amount of credits currently used (🤞🏻 after discounts), which can be compared with the spending limit configured.

- [x] rename GetLatestInvoice to GetUpcomingInvoice

## Related Issue(s)

Fixes #11692 

## How to test
Setup UBB in for a team and lower the spending limit to 1 or 2 credits for a simple test, then start a workspace or two to hit that limit. After a while, cannot tell it precisely, you'll need to reload the dashboard to refresh the spending limit checks. In the `server` component you'll find a log message on spending limit events, and the known dashboard notification should be visible:
Screen Shot 2022-08-23 at 16 21 48

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment


components/server/ee/src/billing/billing-service.ts
```diff
@@ -40,50 +46,53 @@ export class BillingService {
...
-        if (totalUsage >= costCenter.spendingLimit) {
+        if (!(await this.isSpendingLimitCheckEnabled(user))) {
+            return {
+                reached: false,
+                attributionId,
+            };
+        }
+
+        const upcomingInvoice = await this.getUpcomingInvoice(attributionId);
+        const currentUsage = upcomingInvoice.getCredits();
```
easyCZ, line 57:
```suggestion
        const currentInvoiceCredits = upcomingInvoice.getCredits();
```

components/usage/pkg/stripe/stripe.go
```diff
@@ -122,6 +137,47 @@ func (c *Client) updateUsageForCustomer(ctx context.Context, customer *stripe.Cu
 	}, nil
 }
 
+// GetUpcomingInvoice fetches the upcoming invoice for the given team or user id.
+func (c *Client) GetUpcomingInvoice(ctx context.Context, in *v1.GetUpcomingInvoiceRequest) (*StripeInvoice, error) {
```
easyCZ, line 141: The stripe client in this package shouldn't be taking on a dependency from the API. It's tightly coupling the Stripe client to the API when they should be each able to evolve independently. 

In this case, I'd supply either teamID or userID as a string to break this coupling.

components/usage/pkg/stripe/stripe.go
```diff
@@ -122,6 +137,47 @@ func (c *Client) updateUsageForCustomer(ctx context.Context, customer *stripe.Cu
...
+			Query:   query,
+			Expand:  []*string{stripe.String("data.subscriptions")},
+			Context: ctx,
+		},
+	}
+	iter := c.sc.Customers.Search(searchParams)
+	if iter.Err() != nil && !iter.Next() {
+		return nil, ErrorCustomerNotFound
+	}
+	customer := iter.Customer()
```
easyCZ, line 160: What if the iterator contains multiple results?

easyCZ:
What's the plan for dealing with the scenario when the `GetUpcomingInvoice` RPC is not implemented on the usage component when server rolls out?

components/usage/pkg/stripe/stripe.go
AlexTugarev: raising then

easyCZ (approved):

PR #12378, opened 2022-08-25
[server] track more startWorkspace failures

### LINKED ISSUE #12332: Create an alert for failing image-builds (in app clusters)
There are several ways we could do this. The most sustainable from our perspective would be if we could have a metric for the failure rate of all image builds we trigger from `server`. This way it remains operational even after we move image-builds into workspace clusters.

svenefftinge:
Counts additional reasons for startWorkspace failures.

## Related Issue(s)
fixes #12332

## Release Notes

```release-note
NONE
```
- [ ] /werft with-preview

### DIFF
--- a/components/server/src/workspace/workspace-starter.ts
+++ b/components/server/src/workspace/workspace-starter.ts
@@ -112,7 +112,11 @@ import { WithReferrerContext } from "@gitpod/gitpod-protocol/lib/protocol";
 import { IDEOption, IDEOptions } from "@gitpod/gitpod-protocol/lib/ide-protocol";
 import { Deferred } from "@gitpod/gitpod-protocol/lib/util/deferred";
 import { ExtendedUser } from "@gitpod/ws-manager/lib/constraints";
-import { increaseFailedInstanceStartCounter, increaseSuccessfulInstanceStartCounter } from "../prometheus-metrics";
+import {
+    FailedInstanceStartReason,
+    increaseFailedInstanceStartCounter,
+    increaseSuccessfulInstanceStartCounter,
+} from "../prometheus-metrics";
 import { ContextParser } from "./context-parser-service";
 import { IDEService } from "../ide-service";
 import { WorkspaceClusterImagebuilderClientProvider } from "./workspace-cluster-imagebuilder-client-provider";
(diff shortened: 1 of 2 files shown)


components/server/src/workspace/workspace-starter.ts
```diff
@@ -244,6 +248,12 @@ export async function getWorkspaceClassForInstance(
     }
 }
 
+class StartInstanceError extends Error {
+    constructor(public readonly reason: FailedInstanceStartReason, public readonly cause: Error) {
+        super("Starting workspace instance failed: " + cause.message);
```
easyCZ, line 253: How is the reason value assigned to this inherited type before it is used below?
 
```
if (err instanceof StartInstanceError) {
    failedReason = err.reason;
}
```
I don't have that much experience with Errors in JS/TS but here it looks like we're just assigning it to a possibly variable string. My worry with this is that for the metric in prometheus, we should ideally have a fixed cardinality of the label which means we wouldn't expect the variance that could come from this constructor taking on arbitrary `message`.

svenefftinge: The [type](https://github.com/gitpod-io/gitpod/pull/12378/files#diff-fc6040e0ca736fbc219a7234a6c4e6c1c9dfbf711210358df76dd57f45530532R151) is restricted to only three values.

easyCZ (approved):
Spoke on a call. Sven explained that `public readonly reason: FailedInstanceStartReason` actually makes `reason` a property of the object. I was not familiar with that syntax before.

This resolves my question.

Thanks for the changes!

components/server/src/workspace/workspace-starter.ts
```diff
@@ -414,6 +424,11 @@ export class WorkspaceStarter {
                 forceRebuild,
             );
         } catch (e) {
+            let failedReason: FailedInstanceStartReason = "other";
```
aledbf, line 427: @geropl this is the value for the alerts `rate(gitpod_server_instance_starts_failed_total{}[$__rate_interval])`.
Not sure this makes sense to have `other` if that's the default?

geropl: You're :100: right, it doesn't (with the current quality of errors we're receiving). PR to fix this: https://github.com/gitpod-io/gitpod/pull/12574

PR #12382, opened 2022-08-25
[usage] Refactor creditSummary to extend return data and allow immutability

### LINKED ISSUE #12264: Cap reported Usage to max(Usage, SpendingLimit) before reporting to Stripe

laushinka:
## Description
This PR extends the return type of creditSummaryForTeams so that we don't mutate the credits.
It is purposefully separated into different commits.

## Related Issue(s)

Relates #12264

## How to test
Unit tests

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage/pkg/apiv1/billing.go
+++ b/components/usage/pkg/apiv1/billing.go
@@ -6,6 +6,7 @@ package apiv1
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math"
 	"time"
@@ -21,11 +22,12 @@ import (
 	"gorm.io/gorm"
 )
 
-func NewBillingService(stripeClient *stripe.Client, billInstancesAfter time.Time, conn *gorm.DB) *BillingService {
+func NewBillingService(stripeClient *stripe.Client, billInstancesAfter time.Time, conn *gorm.DB, usageClient v1.UsageServiceClient) *BillingService {
 	return &BillingService{
 		stripeClient:       stripeClient,
 		billInstancesAfter: billInstancesAfter,
 		conn:               conn,
+		usageClient:        usageClient,
 	}
 }
 
@@ -33,12 +35,14 @@ type BillingService struct {
 	conn               *gorm.DB
 	stripeClient       *stripe.Client
 	billInstancesAfter time.Time
+	usageClient        v1.UsageServiceClient
+	ctx                context.Context
 
 	v1.UnimplementedBillingServiceServer
 }
 
 func (s *BillingService) UpdateInvoices(ctx context.Context, in *v1.UpdateInvoicesRequest) (*v1.UpdateInvoicesResponse, error) {
-	credits, err := s.creditSummaryForTeams(in.GetSessions())
+	credits, err := s.creditSummaryForTeams(ctx, in.GetSessions())
 	if err != nil {
 		log.Log.WithError(err).Errorf("Failed to compute credit summary.")
 		return nil, status.Errorf(codes.InvalidArgument, "failed to compute credit summary")
@@ -89,8 +93,10 @@ func (s *BillingService) GetUpcomingInvoice(ctx context.Context, in *v1.GetUpcom
 	}, nil
 }
 
-func (s *BillingService) creditSummaryForTeams(sessions []*v1.BilledSession) (map[string]int64, error) {
+func (s *BillingService) creditSummaryForTeams(ctx context.Context, sessions []*v1.BilledSession) (map[string]map[string]float64, error) {
 	creditsPerTeamID := map[string]float64{}
+	var spendingLimit float64
+	var upcomingInvoice float64
 
 	for _, session := range sessions {
 		if session.StartTime.AsTime().Before(s.billInstancesAfter) {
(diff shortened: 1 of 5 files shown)


components/usage/pkg/apiv1/billing.go
```diff
@@ -20,24 +21,31 @@ import (
...
+func NewBillingService(stripeClient StripeClient, billInstancesAfter time.Time, conn *gorm.DB, usageClient v1.UsageServiceClient) *BillingService {
 	return &BillingService{
 		stripeClient:       stripeClient,
 		billInstancesAfter: billInstancesAfter,
 		conn:               conn,
+		usageClient:        usageClient,
 	}
 }
 
+type StripeClient interface {
```
laushinka, line 33: The reason why this changed is so that the Stripe client can be "mocked" in `billing-test.go`. I couldn't find another way to do it. I understand this is probably not expected, so I'm happy to talk about how to mock `GetUpcomingInvoice` in a more isolated way.

components/usage/pkg/apiv1/billing_test.go
```diff
@@ -93,26 +146,52 @@ func TestCreditSummaryForTeams(t *testing.T) {
...
-			require.Equal(t, s.Expected, actual)
+			require.Equal(t, s.Expected["creditsUsed"], actual["creditsUsed"])
 		})
 	}
+	t.Cleanup(func() {
+		dbconn.Model(&db.CostCenter{}).Delete(costCenter, costCenter2)
+	})
+}
+
+type testStripeClient struct{}
```
laushinka, line 182: Referred to [here](https://github.com/gitpod-io/gitpod/pull/12382/files#r956462065). Would love to learn how to mock this better.

components/usage/pkg/apiv1/billing.go
```diff
@@ -32,8 +32,10 @@ func NewBillingService(stripeClient StripeClient, billInstancesAfter time.Time,
 }
 
 type StripeClient interface {
-	GetUpcomingInvoice(ctx context.Context, kind stripe.CustomerKind, id string) (*stripe.StripeInvoice, error)
+	GetUpcomingInvoice(ctx context.Context, customerID string) (*stripe.Invoice, error)
 	UpdateUsage(ctx context.Context, creditsPerTeam map[string]map[string]float64) error
+	GetCustomerByTeamID(ctx context.Context, teamID string) (*stripe.Customer, error)
+	GetCustomerByUserID(ctx context.Context, userID string) (*stripe.Customer, error)
```
laushinka, line 38: After merging changes, this `*stripe.Customer` gets undeclared errors inexplicably 😵

components/usage/pkg/apiv1/billing.go
```diff
@@ -21,24 +22,32 @@ import (
...
 		conn:               conn,
+		usageClient:        usageClient,
 	}
 }
 
+type StripeClient interface {
+	GetUpcomingInvoice(ctx context.Context, customerID string) (*stripe.Invoice, error)
+	UpdateUsage(ctx context.Context, creditsPerTeam map[string]map[string]float64) error
+	GetCustomerByTeamID(ctx context.Context, teamID string) (*stripesdk.Customer, error)
+	GetCustomerByUserID(ctx context.Context, userID string) (*stripesdk.Customer, error)
```
easyCZ, line 38: If these are not used the "abstracted" client layer, they don't need to be listed. Here, it looks like it's only used by the mock in the tests which leads me to believe that we can drop them from here.

In Go, the interface is satisfied if all the methods in the interface are implemented. This means that you don't have to list every method on the strict in the interface (unlike with some other languages). This in turn allows you to have super thin interfaces which often only specify one method. But you can then compose interfaces. A good example is the [`io.Reader`](https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/io/io.go;l=83) and [related composed interfaces](https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/io/io.go;l=146)

laushinka: > it looks like it's only used by the mock in the tests

Hmm, this new interface `StripeClient` replaces the previous `*stripe.Client` and is actually used here. At least that's my understanding because when I remove them, I get something like `s.stripeClient.UpdateUsage undefined (type StripeClient has no field or method UpdateUsage)`.

easyCZ: So yes, the `UpdateUsage` method is needed, but the other two I think are not needed. I'll take a look in the IDE later as well.

easyCZ:
I'm commandeering the PR as Laurie is OOO. Needs further refactoring so moving into draft.

PR #12384, opened 2022-08-25
[bridge] don't delete failed condition

### LINKED ISSUE #12355: [bridge] Ignored Error Message
This [error event](https://console.cloud.google.com/logs/query;cursorTimestamp=2022-08-24T18:45:44.984Z;query=%22410bb5a4-af36-432d-bab1-7f62fd304041%22%0Atimestamp%3D%222022-08-24T18:45:44.984Z%22%0AinsertId%3D%22kf4q9gw5t15i2b1c%22?project=gitpod-191109) is not identified as an error and therefore ws-manager-bridge doesn't flag the workspace as erroneous. As a result users try to restart the workspace, but it doesn't have a backup (because the failure happened on start). 

The error event should result in a proper error message flagging the workspace as broken.

svenefftinge:
Don't remove existing failed condition.

fixes #12355
## Release Notes

```release-note
NONE
```

- [ ] /werft with-preview

### DIFF
--- a/components/ws-manager-bridge/src/bridge.ts
+++ b/components/ws-manager-bridge/src/bridge.ts
@@ -285,11 +285,18 @@ export class WorkspaceManagerBridge implements Disposable {
                 // Do not override!
                 log.error(logContext, 'We received an empty "failed" condition overriding an existing one!', {
                     current: instance.status.conditions.failed,
+                    status,
                 });
-
-                // TODO(gpl) To make ensure we do not break anything big time we keep the unconditional override for now, and observe for some time.
-                instance.status.conditions.failed = status.conditions.failed;
             } else {
+                if (
+                    !!instance.status.conditions.failed &&
+                    instance.status.conditions.failed !== status.conditions.failed
+                ) {
+                    log.info(logContext, 'Overriding previous "failed" condition!', {
+                        previous: instance.status.conditions.failed,
+                        status,
+                    });
+                }
                 instance.status.conditions.failed = status.conditions.failed;
             }
             instance.status.conditions.pullingImages = toBool(status.conditions.pullingImages!);


components/ws-manager-bridge/src/bridge.ts
```diff
@@ -285,11 +285,18 @@ export class WorkspaceManagerBridge implements Disposable {
...
             } else {
+                if (
+                    !!instance.status.conditions.failed &&
+                    instance.status.conditions.failed !== status.conditions.failed
+                ) {
+                    log.info(logContext, 'Overriding previous "failed" condition!', {
+                        previous: instance.status.conditions.failed,
+                        status,
+                    });
+                }
```
easyCZ, line 299: Would you be able to provide a bit more context on why this is needed?

Normally in controller theory, the controller takes readings from the environment, reduces them to some data and updates the state with these. It should never look at previous events as that prevents the controller to self-correct and eventually converge to the correct solution.

Here, we're getting an update which says it's not failed. Why does the update contain a not-failed state after it reported a failed state? Did it correct itself? If so, why wouldn't we set our status to what we're getting from the observations?

svenefftinge: > Would you be able to provide a bit more context on why this is needed?

The original issue was that a workspace failed to start (initialize) and as a result didn't receive a backup. Usually, such workspaces have the failed condition set, so the workspace has an error indicator in the dashboard.

Screenshot 2022-08-25 at 14 09 53

In this case ws-manager would send a status that overrides the `failed` condition, so the workspace looks ok and startable. When a user starts it they see

Screenshot 2022-08-25 at 14 14 39

Thanks for helping me reflecting on this more and I agree with what you wrote. The core issue is that we don't have a clear signal of whether a workspace is broken (startable) or not. It's just less surprising/confusing when a workspace is already marked with an error. I'll think a bit more how we can better indicate this.

svenefftinge:
We should [make use of `final_backup_complete` workspace condition](https://github.com/gitpod-io/gitpod/issues/12387) instead.

components/ws-manager-bridge/src/bridge.ts
geropl: x-refing: https://github.com/gitpod-io/gitpod/issues/12355#issuecomment-1229885538

PR #12385, opened 2022-08-25
[terraform] Add random variables to terrraform resource names

nandajavarma:
## Description

This is basically the same changes @adrienthebo opened here https://github.com/gitpod-io/gitpod/pull/12321 
The PR was unfortunately merged to a different branch, and this is an effort to split up the PR again to ease the review since it ended up becoming really big.

The main goal of this change is to make sure re-using the same state continues to work.  So for starters things like database names cannot be re-used across most cloud providers for a week after the deletion, etc. Hence, the change @adrienthebo suggested adds a random variable to the end of resource names. Moreover, if a kubernetes cluster already exists by the name, the cluster creation is skipped and re-installation of KOTS is proceeded.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/infra/modules/eks/kubernetes.tf
+++ b/install/infra/modules/eks/kubernetes.tf
@@ -64,6 +64,13 @@ resource "aws_security_group" "nodes" {
   }
 }
 
+resource "random_string" "ng_role_suffix" {
+  upper   = false
+  lower   = true
+  special = false
+  length  = 4
+}
+
 module "eks" {
   source  = "terraform-aws-modules/eks/aws"
   version = "18.8.1"
@@ -91,6 +98,7 @@ module "eks" {
   eks_managed_node_group_defaults = {
     ami_type                   = "CUSTOM"
     iam_role_attach_cni_policy = true
+    iam_role_use_name_prefix   = false
     ami_id                     = var.image_id
     enable_bootstrap_user_data = true
     vpc_security_group_ids     = [aws_security_group.nodes.id]
@@ -102,6 +110,7 @@ module "eks" {
       enable_bootstrap_user_data = true
       instance_types             = [var.service_machine_type]
       name                       = "service-${var.cluster_name}"
+      iam_role_name              = format("%s-%s", substr("${var.cluster_name}-svc-ng", 0, 58), random_string.ng_role_suffix.result)
       subnet_ids                 = module.vpc.public_subnets
       min_size                   = 1
       max_size                   = 4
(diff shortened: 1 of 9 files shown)


install/infra/modules/eks/kubernetes.tf
```diff
@@ -102,6 +110,7 @@ module "eks" {
       enable_bootstrap_user_data = true
       instance_types             = [var.service_machine_type]
       name                       = "service-${var.cluster_name}"
+      iam_role_name              = format("%s-%s", substr("${var.cluster_name}-svc-ng", 0, 58), random_string.ng_role_suffix.result)
```
MrSimonEmms, line 113: Am I correct in reading that this it to make a string a max of 62 characters long, always ending `-svc-ng-<4-random-chars>`?

install/tests/Makefile
```diff
@@ -53,8 +53,8 @@ gcp-kubeconfig:
 ## azure-kubeconfig: Get the kubeconfig configuration for Azure AKS
 azure-kubeconfig:
 	export KUBECONFIG=${KUBECONFIG} && \
-	export resource=$$(echo "$$TF_VAR_TEST_ID" | sed "s/[\\W\\-]//") && \
-	az aks get-credentials --name gitpod-test-nor-primary-$$resource --resource-group gitpod-test-nor-$$resource --file ${KUBECONFIG} || echo "No cluster present"
+	export resource=$$(echo "$$TF_VAR_TEST_ID" | sed "s/[\\W\\-]//g") && \
+	az aks get-credentials --name test-cluster-$$resource --resource-group sh-test-$$resource --file ${KUBECONFIG} || echo "No cluster present"
```
MrSimonEmms, line 57: Can you use the `terraform output` command instead? Something like...

```suggestion
	az aks get-credentials --name $(terraform output -json <name> | jq -r '.') --resource-group $(terraform output -json <name> | jq -r '.') --file ${KUBECONFIG} || echo "No cluster present"
```

MrSimonEmms (approved):
/hold

Looks ok. Added a couple of questions, with a "hold" to see if you want to change anything. Remove hold at your convenience

install/infra/modules/eks/kubernetes.tf
nandajavarma: that is correct. I believe that is the max char limit.

install/tests/Makefile
nandajavarma: This would involve changing the AKS terraform module output too. Do you mind if I take this up on a follow-up? I have to refactor the AKS module soon once the reference architecture is out.

PR #12386, opened 2022-08-25
[server] 'abort' prebuilds

### LINKED ISSUE #11453: Calling StopWorkspace on a workspace that is not fully up yet
### Bug description

Currently if calling StopWorkspace on a workspace that has not fully started up yet will cause that workspace state to become failed. 
We need to better handle this case. This is currently used by webapp to stop already running prebuild if newer prebuild has started (commit is obsolete, etc).

It is negatively impacting our workspace success rate (~5-10% depending on prebuild volume).

### Steps to reproduce

https://gitpod.slack.com/archives/C02EN94AEPL/p1657828853695599

### Workspace affected

_No response_

### Expected behavior

`conditions` object is a child of `status` column in `d_b_workspace_instance` table. 

On the workspace side, for `ws-manager`:
Add a `wasCancelled` attribute to the `conditions` object, and inspect the field when setting the phase on stop workspace. 

On the webapp side, for `server`:
If the `wasCancelled` is true, set the phase to `stopped` (instead of `failed`). For workspace start, we then have to make sure we do not try and start a new workspace instance from a prior workspace instance where its conditions have `wasCancelled` set to true.

### Example repository

_No response_

### Anything else?

_No response_

svenefftinge:
Use the new `StopWorkspacePolicy.ABORTED` introduced with #12284 to avoid
spending resources on backing up discarded prebuilds.

## Related Issue(s)

Fixes #11453

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/server/ee/src/prebuilds/prebuild-manager.ts
+++ b/components/server/ee/src/prebuilds/prebuild-manager.ts
@@ -31,6 +31,8 @@ import { secondsBefore } from "@gitpod/gitpod-protocol/lib/util/timeutil";
 
 import { inject, injectable } from "inversify";
 import * as opentracing from "opentracing";
+import { StopWorkspacePolicy } from "@gitpod/ws-manager/lib";
+import { error } from "console";
 
 export class WorkspaceRunningError extends Error {
     constructor(msg: string, public instance: WorkspaceInstance) {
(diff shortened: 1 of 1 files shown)


kylos101:
:wave:  @svenefftinge , #12284 is deployed now, so if you wanted to proceed with this from a webapp standpoint, that woudl be doable.

cc: @sagor999 for posterity

components/server/ee/src/prebuilds/prebuild-manager.ts
```diff
@@ -66,16 +68,27 @@ export class PrebuildManager {
...
+                        log.info(
+                            { userId: user.id, instanceId: instance.id, workspaceId: instance.workspaceId },
+                            "Aborting Prebuild workspace because a newer commit was pushed to the same branch.",
+                        );
+                        results.push(
+                            this.workspaceStarter.stopWorkspaceInstance(
+                                { span },
+                                instance.id,
+                                instance.region,
+                                StopWorkspacePolicy.ABORT,
```
svenefftinge, line 82: The main change is, to use this new policy which makes ws-manager not care about backups.

andrew-farries (approved):

PR #12174, opened 2022-08-17
[kots] Allow users to upload a `.docker/config.json` file

Pothulapati:
## Description

This PR updates the `kots` UI to add a new config option to upload registry
credentials irrespective of the registry being used, which is then merged
into a single `config.json` file and passed as the registry secret which
is then used across the workspace image builder components.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/12136

## How to test

Upload a `dockerconfigjson` file through the new option, 
Screenshot from 2022-08-18 19-03-10

and see those credentials in the secret by running
```
kubectl -n gitpod get secret builtin-registry-auth --output="jsonpath={.data.\.dockerconfigjson}" | base64 -d
```

## Release Notes

```release-note
[kots] Allow users to upload a `.docker/config.json` file
```

## Werft options:

- [ ] /werft with-preview

PR #12398, opened 2022-08-25
[kots] load dockerConfigJson reigstry names into `privateBaseImageAllowList`

Pothulapati:
## Description

Follow upto https://github.com/gitpod-io/gitpod/pull/12174

This PR updates the installer logic to also load the auth's reigstry
URL's into `.containerRegistry.privateBaseImageAllowList`.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
[kots] load dockerConfigJson reigstry names into `privateBaseImageAllowList`
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/kots/manifests/gitpod-installer-job.yaml
+++ b/install/kots/manifests/gitpod-installer-job.yaml
@@ -178,6 +178,15 @@ spec:
                 yq e -i ".containerRegistry.privateBaseImageAllowList += \"docker.io\"" "${CONFIG_FILE}"
               fi
 
+              if [ '{{repl ConfigOptionNotEquals "reg_docker_config" "" }}' = "true" ];
+              then
+                DOCKER_CONFIG='{{repl ConfigOptionData "reg_docker_config" | Base64Encode }}'
+                echo "${DOCKER_CONFIG}" | base64 -d  > /tmp/userconfig.json
+                # Add the registries to the server allowlist
+                yq e -i ".containerRegistry.privateBaseImageAllowList += $(cat /tmp/userconfig.json | jq '.auths' | jq -rc 'keys')" "${CONFIG_FILE}"
+                yq e -i ".containerRegistry.privateBaseImageAllowList += \"docker.io\"" "${CONFIG_FILE}"
+              fi
+
               # Output the local registry secret - this is proxy.replicated.com if user hasn't set their own
               echo "{{repl LocalRegistryImagePullSecret }}" | base64 -d > /tmp/kotsregistry.json
 
@@ -355,9 +364,6 @@ spec:
                   "${GITPOD_OBJECTS}/templates/gitpod.yaml" \
                   | base64 -d \
                   > /tmp/currentconfig.json
-
-                DOCKER_CONFIG='{{repl ConfigOptionData "reg_docker_config" | Base64Encode }}'
-                echo "${DOCKER_CONFIG}" | base64 -d  > /tmp/userconfig.json
 
                 export REGISTRY_SECRET=$(jq -s '.[0] * .[1]' /tmp/userconfig.json /tmp/currentconfig.json | base64 -w 0)
 
--- a/install/kots/manifests/kots-config.yaml
+++ b/install/kots/manifests/kots-config.yaml
@@ -119,7 +119,7 @@ spec:
           when: '{{repl ConfigOptionEquals "reg_docker_config_enable" "1" }}'
           type: file
           required: true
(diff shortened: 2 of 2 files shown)


lucasvaltl:
We should likely also call out that these will be added to the allow list via a small disclaimer in the installation Ui. What happens if a user also adds them to the allowlist themselves - will it break if things are in there twice? :)

install/kots/manifests/gitpod-installer-job.yaml
```diff
@@ -178,6 +178,14 @@ spec:
                 yq e -i ".containerRegistry.privateBaseImageAllowList += \"docker.io\"" "${CONFIG_FILE}"
               fi
 
+              if [ '{{repl ConfigOptionNotEquals "reg_docker_config" "" }}' = "true" ];
+              then
+                DOCKER_CONFIG='{{repl ConfigOptionData "reg_docker_config" | Base64Encode }}'
+                echo "${DOCKER_CONFIG}" | base64 -d  > /tmp/userconfig.json
+                # Add the registies to the server allowlist
```
MrSimonEmms, line 185: Nit
```suggestion
                # Add the registries to the server allowlist
```

install/kots/manifests/gitpod-installer-job.yaml
```diff
@@ -178,6 +178,14 @@ spec:
...
+                yq e -i ".containerRegistry.privateBaseImageAllowList += $(cat /tmp/userconfig.json | jq '.auths' | jq -rc 'keys')" "${CONFIG_FILE}"
```
MrSimonEmms, line 186:
```suggestion
                yq e -i ".containerRegistry.privateBaseImageAllowList += $(cat /tmp/userconfig.json | jq '.auths' | jq -rc 'keys')" "${CONFIG_FILE}"
                yq e -i ".containerRegistry.privateBaseImageAllowList += \"docker.io\"" "${CONFIG_FILE}"
```

We also need to ensure `docker.io` is in the allow list - yes, this will duplicate this if it's either in the existing allowlist or they're also using airgapping, but that'll be ok

MrSimonEmms (changes requested):

MrSimonEmms:
This works well. I've tested using:
 - [a repo which uses a private registry](https://github.com/MrSimonEmms/gitpod-private-image-test)
 - [a repo which uses a public image that extends `gitpod/workspace-full`](https://github.com/MrSimonEmms/gitpod-template-brainfuck)
 - [a repo which uses a public image that doesn't extend `gitpod/workspace-full`](https://github.com/MrSimonEmms/gitpod-centos-node-example)
 - [a Gitpod repo](https://github.com/gitpod-io/template-typescript-node)

install/kots/manifests/kots-config.yaml
```diff
@@ -119,7 +119,7 @@ spec:
           when: '{{repl ConfigOptionEquals "reg_docker_config_enable" "1" }}'
           type: file
           required: true
-          help_text: Docker [config JSON file](https://docs.docker.com/engine/reference/commandline/cli/#sample-configuration-file) with auth credentials used to access private registries, for workspace images.
+          help_text: "Docker [config JSON file](https://docs.docker.com/engine/reference/commandline/cli/#sample-configuration-file) with auth credentials used to access private registries, for workspace images. **NB.** All of the registries in the config with be automatically added to the [`privateBaseImageAllowList`]()."
```
MrSimonEmms, line 122:
```suggestion
          help_text: "Docker [config JSON file](https://docs.docker.com/engine/reference/commandline/cli/#sample-configuration-file) with auth credentials used to access private registries, for workspace images. **NB.** This will allow base workspace images to be pulled from these registries."
```

MrSimonEmms:
```suggestion
          help_text: "Docker [config JSON file](https://docs.docker.com/engine/reference/commandline/cli/#sample-configuration-file) with auth credentials used to access private registries, for workspace images. **NB.** All of the registries in the config will be automatically added to the [`privateBaseImageAllowList`]()."
```

nandajavarma: @MrSimonEmms the reason why I mentioned `privateBaseImageAllowList` is because Lucas wanted to ensure that we make people aware of the implications of this change. He was going to work on a doc that explains this. Do you think maybe we can add it somehow?

MrSimonEmms (approved):
You may remove the hold at will

PR #12405, opened 2022-08-25
[installer]: add prometheus port to ws-manager service

### LINKED ISSUE #12404: Add monitoring ports to ws-manager service

MrSimonEmms:
## Description

 Add monitoring ports to ws-manager service

## Related Issue(s)

Fixes #12404

## How to test

Check port `9500` added to the `ws-manager` service.

## Release Notes

```release-note
[installer]: add prometheus port to ws-manager service
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/installer/cmd/testdata/render/aws-setup/output.golden
+++ b/install/installer/cmd/testdata/render/aws-setup/output.golden
@@ -5418,6 +5418,10 @@ spec:
     port: 8080
     protocol: TCP
     targetPort: 8080
+  - name: metrics
+    port: 9500
+    protocol: TCP
+    targetPort: 9500
   selector:
     app: gitpod
     component: ws-manager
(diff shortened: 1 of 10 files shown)


aledbf:
@MrSimonEmms this will expose the port in clusters using a service type=LoadBalancer

MrSimonEmms:
Well spotted @aledbf - I was wondering if you'd notice my deliberate mistake 🤣

MrSimonEmms:
How does this get exposed? This service doesn't receive any configuration for the service type and this pattern is already used elsewhere in the Installer (eg, [content-service](https://github.com/gitpod-io/gitpod/blob/main/install/installer/pkg/components/content-service/objects.go#L23,L27)).

The only two components that can be exposed as a load balancer are `proxy` and `ws-proxy` - AFAIK, changing those services wouldn't expose this service.

aledbf:
> How does this get exposed?

Right now, the installer doesn't support governed workspace clusters. Meaning a workspace cluster that exposes `ws-manager` to the Internet and uses `gpctl` to register the cluster against a meta cluster.


At the moment, we patch the `ws-manager` service to achieve this.

PR #12311, opened 2022-08-23
[stripe] Refactor finding of customers, make more composable

easyCZ:
## Description

This PR does the following to the stripe client:

* Re-uses a single method for searching for customers
* Increases composability of the methods by accepting CustomerID when searching for Invoices rather than TeamID or UserID
* Adds specific methods to look up a Customer from a TeamID or UserID
* Removed the `stripe.Kind`, this logic is handled by `GetCustomerByTeamID` or `GetCustomerByUserID`

These changes are needed for a subsequent PR which will be updating an Invoice with metadata

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

PR #12409, opened 2022-08-25
[stripe] Set reportId on invoices after updating credits

easyCZ:
## Description

Retrieves the upcoming invoice and sets report ID into metadata.

This is needed to be able to retrieve the reportID when the invoice finalized, such that we can mark the workspace instances as already billed.

## Related Issue(s)

* Depends on https://github.com/gitpod-io/gitpod/pull/12311

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


easyCZ:
dependency landed
/unhold

svenefftinge (approved):

PR #12424, opened 2022-08-26
[usage] Add _lastModified column to d_b_workspace_instance_usage

easyCZ:
## Description

Adds a required `_lastModified` field to `d_b_workspace_instance_usage` such that we can enable db_sync for it

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/gitpod-db/src/typeorm/migration/1661492836561-WorkspaceInstanceUsageLastModified.ts
+++ b/components/gitpod-db/src/typeorm/migration/1661492836561-WorkspaceInstanceUsageLastModified.ts
@@ -1,0 +1,23 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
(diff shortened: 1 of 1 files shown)


svenefftinge:
Is the reason you add this column to sync it between EU<>US? I noticed it is not listed [here](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-db/src/tables.ts).

easyCZ:
Yes, that's the reason for adding it - to be able to enable db-sync for the table.

svenefftinge:
> Yes, that's the reason for adding it - to be able to enable db-sync for the table.

Why do you create a separate PR for that?

easyCZ:
To ensure they are rolled out separately. The DB changes must happen before the db-sync changes and to ensure that, I need to not land the changes until we've rolled this out.

Yes, it is possible we could keep them together because we normally run migrations before a rollout, but in some cases a component is upgraded directly (in incidents etc) without running the migrations and that would break in this case.

svenefftinge (approved):

PR #12431, opened 2022-08-26
Fix AccountStatementProvider 

geropl:
## Description
Fixes caching issue withing AccountStatementProvider. The reasons for this PR is a refactoring 2 weeks back which slightly broke the credit limit check; this PR is meant to bring it back.

## Related Issue(s)

Fixes #

## How to test
 - [signup](https://gpl-fix-ac6f1a002ea6.preview.gitpod-dev.com/workspaces) with user A
 - start a workspace
 - signup with user B
 - add negative credits to the accounting table for user A so user A should no longer be able to start workspaces
    - `INSERT INTO d_b_account_entry SET uid = '<some-id>', userId = '<your-userId>', amount = -50, date = '2022-08-26T13:47:36.360Z', expiryDate ='2022-10-26T13:47:36.360Z', kind = 'credit';`
    @AlexTugarev Already did this for [your user](https://gpl-fix-ac6f1a002ea6.preview.gitpod-dev.com/admin/users/c8d4457f-ee0f-4220-9af7-e87360a59669)
 - check that:
   - user A can indeed no longer start workspaces :heavy_check_mark: 
   - user B still can :heavy_check_mark: 

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/gitpod-protocol/src/util/garbage-collected-cache.ts
+++ b/components/gitpod-protocol/src/util/garbage-collected-cache.ts
@@ -37,6 +37,11 @@ export class GarbageCollectedCache<T> {
         if (!entry) {
             return undefined;
         }
+        // Still valid?
+        if (entry.expiryDate < Date.now()) {
+            this.store.delete(entry.key);
+            return undefined;
+        }
         return entry.value;
     }
 
--- a/components/server/ee/src/billing/entitlement-service-chargebee.ts
+++ b/components/server/ee/src/billing/entitlement-service-chargebee.ts
@@ -79,7 +79,7 @@ export class EntitlementServiceChargebee implements EntitlementService {
         runningInstances: Promise<WorkspaceInstance[]>,
     ): Promise<boolean> {
         // As retrieving a full AccountStatement is expensive we want to cache it as much as possible.
-        const cachedAccountStatement = this.accountStatementProvider.getCachedStatement();
+        const cachedAccountStatement = this.accountStatementProvider.getCachedStatement(userId);
         const lowerBound = this.getRemainingUsageHoursLowerBound(cachedAccountStatement, date.toISOString());
         if (lowerBound && (lowerBound === "unlimited" || lowerBound > Accounting.MINIMUM_CREDIT_FOR_OPEN_IN_HOURS)) {
             return true;
@@ -108,7 +108,7 @@ export class EntitlementServiceChargebee implements EntitlementService {
             return "unlimited";
         }
 
-        const diffInMillis = new Date(cachedStatement.endDate).getTime() - new Date(date).getTime();
+        const diffInMillis = Math.max(0, new Date(cachedStatement.endDate).getTime() - new Date(date).getTime());
         const maxPossibleUsage = millisecondsToHours(diffInMillis) * MAX_PARALLEL_WORKSPACES;
         return cachedStatement.remainingHours - maxPossibleUsage;
     }
--- a/components/server/ee/src/billing/entitlement-service-chargebee.ts
+++ b/components/server/ee/src/billing/entitlement-service-chargebee.ts
@@ -57,12 +57,7 @@ export class EntitlementServiceChargebee implements EntitlementService {
             hasHitParallelWorkspaceLimit(),
         ]);
 
-        const result = enoughCredits && !hitParallelWorkspaceLimit;
-
-        console.log("mayStartWorkspace > hitParallelWorkspaceLimit " + hitParallelWorkspaceLimit);
-
         return {
-            mayStart: result,
             oufOfCredits: !enoughCredits,
             hitParallelWorkspaceLimit,
         };
@@ -87,7 +82,6 @@ export class EntitlementServiceChargebee implements EntitlementService {
         const cachedAccountStatement = this.accountStatementProvider.getCachedStatement();
         const lowerBound = this.getRemainingUsageHoursLowerBound(cachedAccountStatement, date.toISOString());
         if (lowerBound && (lowerBound === "unlimited" || lowerBound > Accounting.MINIMUM_CREDIT_FOR_OPEN_IN_HOURS)) {
-            console.log("checkEnoughCreditForWorkspaceStart > unlimited");
             return true;
         }
 
@@ -96,7 +90,6 @@ export class EntitlementServiceChargebee implements EntitlementService {
             date.toISOString(),
             runningInstances,
         );
-        console.log("checkEnoughCreditForWorkspaceStart > remainingUsageHours " + remainingUsageHours);
         return remainingUsageHours > Accounting.MINIMUM_CREDIT_FOR_OPEN_IN_HOURS;
     }
 
--- a/components/server/ee/src/user/account-statement-provider.ts
+++ b/components/server/ee/src/user/account-statement-provider.ts
@@ -8,6 +8,7 @@ import { injectable, inject } from "inversify";
 import { WorkspaceInstance } from "@gitpod/gitpod-protocol";
 import { AccountStatement } from "@gitpod/gitpod-protocol/lib/accounting-protocol";
 import { AccountService } from "@gitpod/gitpod-payment-endpoint/lib/accounting";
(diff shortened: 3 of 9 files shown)


components/server/src/billing/entitlement-service.ts
```diff
@@ -81,7 +78,7 @@ export class CommunityEntitlementService implements EntitlementService {
         date: Date,
         runningInstances: Promise<WorkspaceInstance[]>,
     ): Promise<MayStartWorkspaceResult> {
-        return { mayStart: true };
```
AlexTugarev, line 84: I'd prefer to keep a readable signal. 
the empty object is hard to understand throughout the chain of calls.

geropl: But it makes it very likely to mis-interpret. Looking at the interface definition I think makes it very clear that this is a bag of reasons to _not_ start this workspace.

Or do you mean this specific site? This would be `{ outOfCredits: false }`, for instances, which is not really relevant, because this is a purely Chargebee specific field. I'd rather keep it as simple as possible.

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -256,16 +256,16 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
-            throw new ResponseError(ErrorCodes.INTERNAL_SERVER_ERROR, `Error in Entitlement Service.`);
-        }
-        log.info("mayStartWorkspace", { result });
-        if (result.mayStart) {
-            return; // green light from entitlement service
+            TraceContext.addNestedTags(ctx, { mayStartWorkspace: { result } });
+        } catch (err) {
+            log.error({ userId: user.id }, "EntitlementSerivce.mayStartWorkspace error", err);
+            TraceContext.setError(ctx, err);
+            return; // we don't want to block workspace starts because of internal errors
```
AlexTugarev, line 266: this means, all negative conditions should produce a negative response. if anything unexpected happens, then no rules are enforced. is that the intention? 

what's the expected feedback loop for us?

geropl: The feedback loop should never involve the user. The error logged here should a) feed a log-rate metric, and b) we should have specific metrics for possible causes of this, e.g. the Stripe API.

components/server/src/billing/entitlement-service.ts
AlexTugarev: > Looking at the interface definition I think makes it very clear that this is a bag of reasons to not start this workspace.

Let walk though this by example to show what I mean by a missing readability about this signal
Given
```
let result = entitlementService.mayStartWorkspace()
```
we want to decide if the workspace is allowed to take off. Having no strong signal on this, we need to check the 3  optional attributes to be present. 
This is possible to do so – *now* after applying a [polarizing filter](https://github.com/gitpod-io/gitpod/pull/11936/commits/b0b4d88b9ac884ce8052439849b50c2af6f7d21d#diff-42937a09d5730082001f0989261d91bf1560753c97c60e29ce58c76f0ca7c603R269), which showed already how fragile that can be. 

But think of a future task to extend the `entitlementService.mayStartWorkspace` check with more negative signals. It feels to be error prone to extend the evaluation of a result object to meet the extended computation.

components/server/ee/src/user/account-statement-provider.ts
```diff
@@ -19,20 +20,18 @@ export type CachedAccountStatement = Pick<AccountStatement, "remainingHours" | "
...
+     */
+    protected readonly cachedStatements = new GarbageCollectedCache<CachedAccountStatement>(5 * 60, 10 * 60);
 
-    setCachedStatement(cachedStatement: CachedAccountStatement) {
-        this.cachedStatement = cachedStatement;
-    }
-
-    getCachedStatement(): CachedAccountStatement | undefined {
-        return this.cachedStatement;
+    getCachedStatement(userId: string, date: Date): CachedAccountStatement | undefined {
```
svenefftinge, line 28: I still haven't looked very closely into the usage code so please bar with me if this is a stupid question: Why do we pass around the `now` date. Is there another time a user might start a workspace than now resp. can we answer that question?

components/gitpod-protocol/src/util/garbage-collected-cache.ts
```diff
@@ -32,11 +32,16 @@ export class GarbageCollectedCache<T> {
         });
     }
 
-    public get(key: string): T | undefined {
+    public get(key: string, now: Date = new Date()): T | undefined {
```
svenefftinge, line 35: I find passing a date a bit strange because it suggests I can pass in an old date and get a correct answer.

components/gitpod-protocol/src/util/garbage-collected-cache.ts
```diff
@@ -32,11 +32,16 @@ export class GarbageCollectedCache<T> {
...
         const entry = this.store.get(key);
         if (!entry) {
             return undefined;
         }
+        // Still valid?
+        if (entry.expiryDate < now.getTime()) {
+            this.store.delete(entry.key);
+            return undefined;
+        }
```
svenefftinge, line 44: 💯  Invalidating on the fly makes a ton of sense.

components/server/ee/src/user/account-statement-provider.ts
geropl: > I still haven't looked very closely into the usage code

FYI: This the is old Chargebee-related code, not usage-based related. (see updated description above :point_up: )

> Why do we pass around the now date?

This mostly is for making sure we're making time-based decision based on the same "reference"/"now". Also, it's helpful for testing.

:cloud: Could be that we do not/no longer need it in all places now - but as this code's lifetime is 6 months max., I don't think it makes sense ~cleanup~ tidy here, now. :+1:

components/gitpod-protocol/src/util/garbage-collected-cache.ts
geropl: Mostly the same reasons posted here: https://github.com/gitpod-io/gitpod/pull/12431#discussion_r956185187

components/server/ee/src/billing/entitlement-service-chargebee.ts
```diff
@@ -115,7 +108,7 @@ export class EntitlementServiceChargebee implements EntitlementService {
             return "unlimited";
         }
 
-        const diffInMillis = new Date(cachedStatement.endDate).getTime() - new Date(date).getTime();
+        const diffInMillis = Math.max(0, new Date(cachedStatement.endDate).getTime() - new Date(date).getTime());
```
AlexTugarev, line 111: From reading this up and down, I'm pretty sure this new issue stems from the date passed down the line from https://github.com/gitpod-io/gitpod/blob/20106df07bb532ef45159dafa0cab5c3a05ef5f4/components/server/ee/src/billing/entitlement-service.ts#L37, these kind of timing issues are really nasty as they may occur in different places here. Be aware this is all a chain of async and await, and the date is set before awaiting a couple of db queries, right?

geropl: Yes, it does. But in this case I don't see a problem with that. This is more about making the "guessing" correct for more cases, and we should not care about what dates we have where, or where we pull "now" from.

components/server/ee/src/user/account-statement-provider.ts
geropl: :information_source:  This was just a propagation of [this case](https://github.com/gitpod-io/gitpod/pull/12431#discussion_r956186607) which got removed.

components/server/ee/src/billing/entitlement-service-chargebee.ts
AlexTugarev: Nice catch. It took a while to understand the issue here: any small positive or negative minus a large negative equals positive, which leads to growing negative remaining hours. I think we've seen these odd results in the past, but could not pin point the cause.

AlexTugarev (approved):
Just verified, this really does fix the core issue with the check for credits 💯

PR #12438, opened 2022-08-26
Blocked repositories page improvements

filiptronicek:
## Description
Fixes some issues on inside https://github.com/gitpod-io/gitpod/blob/main/components/dashboard/src/admin/BlockedRepositories.tsx. Details are outlined below this comment.

## Related Issue(s)
n/a

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/dashboard/src/admin/BlockedRepositories.tsx
+++ b/components/dashboard/src/admin/BlockedRepositories.tsx
@@ -135,7 +135,7 @@ export function BlockedRepositoriesList(props: Props) {
                         </div>
                         <input
                             type="search"
-                            placeholder="Search by URL Regex"
+                            placeholder="Search by URL RegEx"
                             onKeyDown={(ke) => ke.key === "Enter" && search()}
                             onChange={(v) => {
                                 setQueryTerm(v.target.value.trim());
@@ -148,8 +148,10 @@ export function BlockedRepositoriesList(props: Props) {
                 </div>
             </div>
 
-            <Alert type={"info"} closable={false} showIcon={true} className="flex rounded p-2 w-2/3 mb-2 w-full">
-                <span>Search entries by their Repositoriy URL regular expression (RegEx).</span>
+            <Alert type={"info"} closable={false} showIcon={true} className="flex rounded p-2 mb-2 w-full">
+                <span>
+                    Search entries by their repository URL <abbr title="regular expression">RegEx</abbr>.
+                </span>
             </Alert>
             <div className="flex flex-col space-y-2">
                 <div className="px-6 py-3 flex justify-between text-sm text-gray-400 border-t border-b border-gray-200 dark:border-gray-800 mb-2">
@@ -211,7 +213,7 @@ function AddBlockedRepositoryModal(p: AddBlockedRepositoryModalProps) {
         setError("");
     }, [p.blockedRepository]);
 
-    let save = (): boolean => {
+    const save = (): boolean => {
         const v = ref.current;
         const newError = p.validate(v);
         if (!!newError) {


components/dashboard/src/admin/BlockedRepositories.tsx
```diff
@@ -135,7 +135,7 @@ export function BlockedRepositoriesList(props: Props) {
                         </div>
                         <input
                             type="search"
-                            placeholder="Search by URL Regex"
+                            placeholder="Search by URL RegEx"
```
filiptronicek, line 138: Made it so that we have the same capitalization everywhere

components/dashboard/src/admin/BlockedRepositories.tsx
```diff
@@ -148,8 +148,10 @@ export function BlockedRepositoriesList(props: Props) {
                 </div>
             </div>
 
-            <Alert type={"info"} closable={false} showIcon={true} className="flex rounded p-2 w-2/3 mb-2 w-full">
-                <span>Search entries by their Repositoriy URL regular expression (RegEx).</span>
+            <Alert type={"info"} closable={false} showIcon={true} className="flex rounded p-2 mb-2 w-full">
```
filiptronicek, line 151: Here there was an extra class which was not used because `w-full` already changed the width.

components/dashboard/src/admin/BlockedRepositories.tsx
```diff
@@ -148,8 +148,10 @@ export function BlockedRepositoriesList(props: Props) {
...
+                <span>
+                    Search entries by their repository URL <abbr title="regular expression">RegEx</abbr>.
```
filiptronicek, line 153: Here besides fixing a typo, an abbreviation was added so that we both save space and utilize semantic HTML for abbreviations.

components/dashboard/src/admin/BlockedRepositories.tsx
```diff
@@ -211,7 +213,7 @@ function AddBlockedRepositoryModal(p: AddBlockedRepositoryModalProps) {
         setError("");
     }, [p.blockedRepository]);
 
-    let save = (): boolean => {
+    const save = (): boolean => {
```
filiptronicek, line 216: Use `const` where possible

easyCZ (approved):
LGTM. Adding hold in case you want @gtsiolis to review for UI.

/hold

gtsiolis:
Looking at this now! 👀

components/dashboard/src/admin/BlockedRepositories.tsx
gtsiolis: **thought**: Not particularly proud of how this is rendering by default (see screenshots below) or using this controversial abbreviation design pattern[[1](https://microformats.org/wiki/abbr-design-pattern)] as it comes with some accessibility concerns, but let's keep it for now until we have a better way to present this information. 🤝 

| Chrome | Firefox | Safari |
|-|-|-|
| a | b | c |

gtsiolis (approved):
Thanks @filiptronicek for making this change, @svenefftinge for the ping[[1](https://github.com/gitpod-io/gitpod/pull/12438#event-7267256831)], and @easyCZ for taking a look[[2](https://github.com/gitpod-io/gitpod/pull/12438#pullrequestreview-1086931394)]! 🏀  

LGTM!

Took longer to come back to this after today's [tech talk](https://gitpod.slack.com/archives/C019K7PND1A/p1661284094648809) (internal) because of some spaghetti I had to make. 🍝

components/dashboard/src/admin/BlockedRepositories.tsx
filiptronicek: Thanks for introducing me to Microformats! It's a really great guide from what I'm seeing.

PR #12445, opened 2022-08-26
code-sync: store editSessions

jeanp413:
## Description

Store editSessions in sync server

## How to test

1. Create workspace using latest vscode
2. Test editsessions is working https://code.visualstudio.com/updates/v1_70#_edit-sessions-across-vs-code-for-the-web-and-desktop, in our case it could used to store changes between new workspaces of the same repo 

## Release Notes

```release-note
[server] Support for storing vscode edit sessions in sync server
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/server/src/code-sync/code-sync-service.ts
+++ b/components/server/src/code-sync/code-sync-service.ts
@@ -49,9 +49,16 @@ export type CodeSyncConfig = Partial<{
     };
 }>;
 
-const objectPrefix = "code-sync/";
+function getNamePrefix(resource: ServerResource) {
+    if (resource === "editSessions") {
+        return "edit-sessions/";
+    } else {
+        return "code-sync/";
+    }
+}
+
 function toObjectName(resource: ServerResource, rev: string): string {
-    return objectPrefix + resource + "/" + rev;
+    return getNamePrefix(resource) + resource + "/" + rev;
 }
 
 const fromTheiaRev = "from-theia";
@@ -249,10 +256,11 @@ export class CodeSyncService {
                     resourceKey === "machines"
                         ? 1
                         : config.resources?.[resourceKey]?.revLimit || config?.revLimit || defaultRevLimit;
+                const isEditSessionsResource = resourceKey === "editSessions";
                 const userId = req.user.id;
                 const contentType = req.headers["content-type"] || "*/*";
-                let oldRevList: string[] = [];
-                const rev = await this.db.insert(
+                let oldRevList: string[] | undefined;
+                const newRev = await this.db.insert(
                     userId,
                     resourceKey,
                     async (rev, oldRevs) => {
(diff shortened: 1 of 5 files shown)


jeanp413:
Not sure why the `ci/werft/build — The build failed!` link above redirects to main build, but you can look for the correct build in werft https://werft.gitpod-dev.com/job/gitpod-build-jp-sync-profile.7

mustard-mh:
werft run

akosyakov:
@jeanp413 Does it bring any user facing changes for Gitpod users? if so please update release notes. 🙏

filiptronicek (commented):
@jeanp413 thanks a lot for taking time to support this cool thingy. I am a bit confused in how the user is supposed to work with them. 

I executed the command to store the edit session, then opened a workspace with the same repo and did `Continue Edit session`, but I got this:
image

When actioning on the only item here VS Code wants to open vscode.dev
image

jeanp413:
Good catch I'll remove that vscode.dev option xD

components/server/src/code-sync/code-sync-service.ts
```diff
@@ -316,6 +328,47 @@ export class CodeSyncService {
...
+            // This endpoint is used to delete edit sessions data only
+            const { resource, ref } = req.params;
+            if (resource !== "editSessions") {
+                res.sendStatus(400);
+                return;
+            }
+
+            const userId = req.user.id;
+            await this.db.deleteResource(userId, resource, ref, async (ref?: string) => {
+                const request = new DeleteRequest();
```
mustard-mh, line 346: Why not use `this.deleteObjects` here?

jeanp413: Pushed some :lipstick:

jeanp413:
@filiptronicek I removed the `open in vscode.dev` option but now the quick pick will be empty that's expected, we can't support vscode desktop as there's no way to configure the edit sessions server like we do with settings sync server

filiptronicek (approved):
Code + tests all look good, I can <kbd>Resume latest Edit Session</kbd> just fine (see screenshot). This feature is great - it's like a synced [`git stash`](https://www.git-scm.com/docs/git-stash)!
image

I am not sure @jeanp413 how you made the `vscode.dev` option disappear, but could we make the `Sign in`, `Sign out` actions disappear too (we do that automatically, don't we?)? Removing the whole `Continue Edit Session...` would probably be nice as well if we don't have any options now.

image

/hold for others to review + if you want to integrate some of my suggestions 😊

jeanp413:
>  could we make the Sign in, Sign out actions disappear too (we do that automatically, don't we?)? 

yeah experience is still rough in upstream vscode as is still experimental, will look into fixing it upstream if I find some free time

andrew-farries (approved):

Issue #12254, opened 2022-08-22
[code-desktop] Align on a way to ensure Gitpod API compatibility between IDE clients and SH installations.


akosyakov:
- We decided to change how ouath2 endpoint works. It should ignore unknown scopes, verify and return only subset of known scopes. Clients should use returned scopes as source of truth to check whether features are present. It is for instance how GitHub auth works.
- We will still need to adhere to `api/version` for 2 SH releases which has changes to ouath2 endpoint. After this check can be removed.

Internal RFC [[1](https://www.notion.so/gitpod/Gitpod-Server-API-compatibility-check-ee00fcfe60a84c1e8e018ab6c06c1f5d)]

loujaybee:
Current update: PR Is up:

- https://github.com/gitpod-io/gitpod/pull/12446

But we need to resolve some of the outstanding discussions, would be good to get @geropl and WebApp teams help to unblock this one 🙏

jeanp413:
Closing this as it got deployed

PR #12446, opened 2022-08-26
Add `/oauth/inspect` endpoint

jeanp413:
## Description

Add `/oauth/inspect` endpoint

## Related Issue(s)

Related https://github.com/gitpod-io/gitpod/issues/12254

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/server/src/oauth-server/oauth-controller.ts
+++ b/components/server/src/oauth-server/oauth-controller.ts
@@ -15,6 +15,7 @@ import { inject, injectable } from "inversify";
 import { URL } from "url";
 import { Config } from "../config";
 import { clientRepository, createAuthorizationServer } from "./oauth-authorization-server";
+import { inMemoryDatabase } from "./db";
 
 @injectable()
 export class OAuthController {
@@ -173,6 +174,17 @@ export class OAuthController {
                 handleExpressError(e, res);
                 return;
             }
+        });
+
+        router.get("/oauth/inspect", async (req: express.Request, res: express.Response) => {
+            const clientId = req.query.client as string;
+            if (typeof clientId !== "string" || !Object.keys(inMemoryDatabase.clients).includes(clientId)) {
+                return res.sendStatus(400);
+            }
+
+            const client = inMemoryDatabase.clients[clientId];
+            const scopes = client.scopes.map((s) => s.name);
+            return res.send(scopes);
         });
 
         return router;


components/server/src/oauth-server/db.ts
```diff
@@ -73,7 +73,7 @@ function createVSCodeClient(protocol: "vscode" | "vscode-insiders" | "vscodium")
             { name: "function:getWorkspace" },
             { name: "function:getWorkspaces" },
             { name: "function:getSSHPublicKeys" },
-            { name: "function:sendHeartBeat" },
+            // { name: "function:sendHeartBeat" },
```
jeanp413, line 76: For testing

components/server/src/oauth-server/oauth-controller.ts
```diff
@@ -175,6 +176,17 @@ export class OAuthController {
             }
         });
 
+        router.get("/oauth/inspect", async (req: express.Request, res: express.Response) => {
+            const clientId = req.query.client as string;
+            if (typeof clientId !== "string" || !Object.keys(inMemoryDatabase.clients).includes(clientId)) {
+                res.sendStatus(400);
```
akosyakov, line 182: return?

components/server/src/oauth-server/oauth-controller.ts
```diff
@@ -175,6 +176,17 @@ export class OAuthController {
             }
         });
 
+        router.get("/oauth/inspect", async (req: express.Request, res: express.Response) => {
+            const clientId = req.query.client as string;
+            if (typeof clientId !== "string" || !Object.keys(inMemoryDatabase.clients).includes(clientId)) {
```
AlexTugarev, line 181: this definitely has the potential to provide an unguarded endpoint to scan for random `clientId`s.

tagging as security relevant.

akosyakov: everything is already available in this repo? 🤔

jeanp413: yeah client ids are already public

akosyakov:
@jeanp413 Let's open it for review?

mustard-mh (approved):
/hold 

hold to squash commits

Code LGTM, tested in browser

components/server/src/oauth-server/oauth-controller.ts
easyCZ: I think the point Alex is making is that this allows you to enumerate all client IDs we may have. Therefore, you can estimate how big of an installation of gitpod this is, how many users it has and which exact IDs exist in the system. 

There's a difference between the client ID being public and knowing exactly how many client IDs there are.

akosyakov: What would be an alternative? cc @geropl

akosyakov: > Therefore, you can estimate how big of an installation of gitpod this is, how many users it has and which exact IDs exist in the system.

I'm not sure how you can do it though. You can only there Gitpod installation supports installing ssh keys or not. You cannot get any other information from this endpoint.

easyCZ: Would you be able to provide a use-case this is looking to solve? Is it to discover my own ClientID?

akosyakov: Please have a look at this [RFC](https://www.notion.so/gitpod/Gitpod-Server-API-compatibility-check-ee00fcfe60a84c1e8e018ab6c06c1f5d#5aed369a61cf46c7bf945ae017a6e125) It is a work around for how we do auth now.

geropl: @mikenikles  @AlexTugarev  I think there is a misunderstanding here: Clients !== API clients.

Client is a possible _kind_ of clients we support as defined [here](https://github.com/gitpod-io/gitpod/blob/79ef436f154158d4a4c78aa7e619cedfd52ee83d/components/server/src/oauth-server/db.ts#L38-L96).

> everything is already available in this repo?

:point_up: Enumerating those is fine I think. :slightly_smiling_face:

geropl (approved):
LGTM. Thx @jeanp413 ! :+1:

PR #12462, opened 2022-08-29
[public-api] Fix Stripe Webhook Secret mount path

easyCZ:
## Description

The current implementation would mount the secret under `/stripe-secret/stripe-webhook-secret` but would set the config path to `secrets/stripe-secret`. This fixes that issue by fixing the path.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/public-api/go/config/config.go
+++ b/components/public-api/go/config/config.go
@@ -16,7 +16,3 @@ type Configuration struct {
 
 	Server *baseserver.Configuration `json:"server,omitempty"`
 }
-
-type StripeSecret struct {
-	WebhookSigningKey string `json:"signingKey"`
-}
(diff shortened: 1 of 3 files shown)


components/public-api/go/config/config.go
```diff
@@ -16,7 +16,3 @@ type Configuration struct {
 
 	Server *baseserver.Configuration `json:"server,omitempty"`
 }
-
-type StripeSecret struct {
```
easyCZ, line 20: Removing this as we'll not be needing the struct for the secret

akosyakov (approved):
lgtm

geropl (approved):

PR #12468, opened 2022-08-29
[local-preview] Fail if M1 Mac

Pothulapati:
## Description

Currently, If its a M1 Mac, The local preview instance
just continues running as the failures are in k3s. We
should instead fail earlier so that the user understands
the problem.

We already have a warning for the same in the documentation
website, but having a check seems to be useful.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
[local-preview] Check and exit if M1 Mac
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/preview/entrypoint.sh
+++ b/install/preview/entrypoint.sh
@@ -18,11 +18,17 @@ if [ "$1" != "logging" ]; then
   exit
 fi
 
+# fail if its arm64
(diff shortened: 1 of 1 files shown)


corneliusludmann (approved):
Looks good code-wise. Cannot test on M1 Mac (do not have one) but guess it will work well.

PR #12472, opened 2022-08-29
Clean up webhook events

### LINKED ISSUE #12430: Clean up recorded webhook events
Followup to PR https://github.com/gitpod-io/gitpod/pull/11356

We should auto-delete recorded webhook events after a configured retain period.

AlexTugarev:
by running a periodic garbage collector. This is a pattern we use for other resources.

fixes #12430

## Description
The way GC works is admittedly tricky, as it's an interplay between the periodic GC task and the `db-sync`. Periodically this will mark a bunch of entries as `deleted`, and we rely on the `db-sync` to actually delete the entries.

Because it's proven unreliable to mark entries as delete in the thousands, leaving them in the db for ever after a failure, the settings to mark as deleted periodically are: 600 entries every 6 minutes, which is a rate of ~200K/day and should not produce a lot of DB CPU usage.  

## Related Issue(s)

Fixes #12430

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-db/src/typeorm/webhook-event-db-impl.ts
+++ b/components/gitpod-db/src/typeorm/webhook-event-db-impl.ts
@@ -51,4 +51,20 @@ export class WebhookEventDBImpl implements WebhookEventDB {
         query.limit(limit);
         return query.getMany();
     }
+
+    public async deleteOldEvents(ageInDays: number, limit?: number): Promise<void> {
+        const repo = await this.getRepo();
+        const d = new Date();
+        d.setDate(d.getDate() - ageInDays);
+        const expirationDate = d.toISOString();
+        const query = repo
+            .createQueryBuilder("event")
+            .update()
+            .set({ deleted: true })
+            .where("event.creationTime < :expirationDate", { expirationDate });
+        if (typeof limit === "number") {
+            query.limit(limit);
+        }
+        await query.execute();
+    }
 }
--- a/components/gitpod-db/src/webhook-event-db.ts
+++ b/components/gitpod-db/src/webhook-event-db.ts
@@ -11,4 +11,5 @@ export interface WebhookEventDB {
     createEvent(parts: Omit<WebhookEvent, "id" | "creationTime">): Promise<WebhookEvent>;
     updateEvent(id: string, update: Partial<WebhookEvent>): Promise<void>;
     findByCloneUrl(cloneUrl: string, limit?: number): Promise<WebhookEvent[]>;
+    deleteOldEvents(ageInDays: number, limit?: number): Promise<void>;
 }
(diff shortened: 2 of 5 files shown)


AlexTugarev:
back to draft. just tested the query and `explain` showed me a full table scan :-P

AlexTugarev:
I removed the problematic `delete <> 0` predicate. Also adding more description on how this will work in the end.

AlexTugarev:
@geropl, maybe you want to have a look at this one?

geropl:
@AlexTugarev Will review now.

components/gitpod-db/src/typeorm/webhook-event-db-impl.ts
```diff
@@ -51,4 +51,21 @@ export class WebhookEventDBImpl implements WebhookEventDB {
...
         return query.getMany();
     }
+
+    public async deleteOldEvents(ageInDays: number, limit?: number): Promise<void> {
+        const repo = await this.getRepo();
+        const d = new Date();
+        d.setDate(d.getDate() - ageInDays);
+        const expirationDate = d.toISOString();
+        const LIMIT = typeof limit === "number" ? ` LIMIT ${limit} ` : "";
+        await repo.query(
```
geropl, line 61: Would be great to use the ORM here. One similar example:
```
        const repo = await this.getWorkspaceRepo();
        const qb = repo
            .createQueryBuilder("ws")
            .select(["ws.id", "ws.ownerId"])
            .where(`ws.contentDeletedTime != ''`)
            .andWhere(`ws.contentDeletedTime < :minPurgeTime`, { minPurgeTime })
            .limit(limit);
        return await qb.getMany();
``

File: components/gitpod-db/src/typeorm/webhook-event-db-impl.ts [hunk 2]
```diff
@@ -51,4 +51,21 @@ export class WebhookEventDBImpl implements WebhookEventDB {
         query.limit(limit);
         return query.getMany();
     }
+
+    public async deleteOldEvents(ageInDays: number, limit?: number): Promise<void> {
+        const repo = await this.getRepo();
+        const d = new Date();
+        d.setDate(d.getDate() - ageInDays);
```
geropl, line 58: :cloud: Would it make a difference to use `setUTCDate` here? :thinking:

AlexTugarev: that line just applies a shift of days. as we alway initialize with date of server (you recall ;-)) it should be fine.

components/gitpod-db/src/typeorm/webhook-event-db-impl.ts
AlexTugarev: I see no point in it, but let's do it!

AlexTugarev: done

AlexTugarev:
build was failing for an unrelated issue with db-sync replication. restarted.

geropl (approved):
Code LGTM, tested and saw the GC doing it's job. Thank you! :tada:

PR #12498, opened 2022-08-30
[supervisor] support ssh gateway on not gitpod base image

### LINKED ISSUE #12143: SSH connection fails if workspace image does not contain "/home/gitpod" directory and "gitpod" user
### Bug description

We very recently ran into a weird issue when trying the new direct ssh connection with public/private keys. The behavior is that when trying to connect to the workspace (either directly via ssh command or via the vscode desktop integration) it "appears" to succeed authentication but immediately gets a closed by remote host error. After looking in the logs of the workspace pod itself, it turned out that the authentication was attempting to get the `/home/gitpod/authorized_keys` file despite the fact that the base image for this workspace did not actually include that file/user yet (we needed a tool that is only available internal and were using its build image). I think previously when starting a workspace, if that user/dir did not exist than it was being autocreated but it seems like that autocreation process happens after the ssh auth check?

### Steps to reproduce

1. Create a workspace using an image that is not natively setup for gitpod (some basic ubuntu image for example)
2. Add valid ssh public keys to user config if not already done
3. Attempt to use the ssh connection command to ssh
4. Get connection closed by remote host error

### Workspace affected

_No response_

### Expected behavior

SSH connection should succeed assuming the public/private key setup is valid

### Example repository

_No response_

### Anything else?

_No response_

### LINKED ISSUE #12265: [ide] Investigate the impossibility of connecting via SSH to some workspaces using custom Docker images
We've found one case where the use of a Custom Docker Image in .gitpod.yml caused the SSH connection to fail (which makes it also impossible to connect using JetBrains IDE): https://github.com/spring-projects/spring-petclinic/

In this specific repository, it seems it started failing after this change: https://github.com/spring-projects/spring-petclinic/commit/5aef5c380c564d9c22afef3e173b1ec6af2c54bd

_Originally discussed in https://github.com/gitpod-io/gitpod/issues/12163#issuecomment-1222253943_

iQQBot:
## Description

[supervisor] support ssh gateway on not gitpod base image

In gitpod base docker image, we write HOME environment, but if user use none gitpod base image, the HOME environment is not set or set other value, which causes public key didn't write in correct path
|not gitpod base image|gitpod base image|
|---|---|
|image|image|

This PR change use hard code `/home/gitpod` as HOME dir

## Related Issue(s)

Fixes #12265
Fixes #12143

## How to test

1. open this example repo in preview environment https://github.com/spring-projects/spring-petclinic
2. connect to workspace via ssh gateway (copy/paste ssh command or upload your public key)
3. it should be connected

## Release Notes

```release-note
[supervisor] support ssh gateway on not gitpod base image
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/supervisor/pkg/supervisor/services.go
+++ b/components/supervisor/pkg/supervisor/services.go
@@ -675,10 +675,7 @@ func (is *InfoService) WorkspaceInfo(context.Context, *api.WorkspaceInfoRequest)
 		}
 	}
 
-	resp.UserHome, err = os.UserHomeDir()
-	if err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
-	}
+	resp.UserHome = "/home/gitpod"
 
 	endpoint, host, err := is.cfg.GitpodAPIEndpoint()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


mustard-mh (commented):
Code LGTM, going to test it

PR #12501, opened 2022-08-30
Pagination on Usage page

### LINKED ISSUE #12388: Slowness loading usage list view
Loading the usage list for the Gitpod team take 10s or longer, even when toggling between current and previous months. 

If this only happened once, we might have a chance to mitigate e.g. by prefetching, but it's not usable if it happens on every click.

NOTES: 
- Preserving the ability to change the sort direction on the timestamp column is not required - feel free to drop.
- Showing the total usage for the period is required. 
- A spinner when changing the period (as per #11960) is helpful, but not unfortunately not sufficient.

AlexTugarev:
## Description

Adding server-side pagination to list usage.

❗ There is a test commit to set the `perPage` attribute to a value of `3`, which makes this PR handy to test. That should be removed before merging.

Screen Shot 2022-08-30 at 17 43 20

Notes
* sort by `Timestamp` is removed as agreed-on
* drive-by fix of `[ 1 ]` page links to make them clickable
* total credits used is computed on DB instead on frontend

## Related Issue(s)

Fixes #12388 

## Release Notes

```release-note
Speed up the Usage page for teams.
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment


roboquat:

<details>

</details>

components/gitpod-protocol/src/usage.ts
```diff
@@ -42,7 +42,6 @@ export interface ExtendedBillableSession extends BillableSession {
 
 export interface BillableSessionRequest {
     attributionId: string;
-    startedTimeOrder: SortOrder;
```
svenefftinge, line 45: Should we not keep ordering information in the protocol especially if we do server-side pagination?

AlexTugarev: given that this feature is currently not working at all I wanted to start with tidy first.

svenefftinge:
Hey Alex! I can look into this. Do you have a team set up (invite link?) that has a bit of usage shown already?

components/usage-api/usage/v1/usage.proto
```diff
@@ -34,10 +34,18 @@ message ListBilledUsageRequest {
     }
 
     Ordering order = 4;
+
+    int64 per_page = 5;
+    int64 page = 6;
```
easyCZ, line 39: Encapsulate the pagination in a message.

components/usage-api/usage/v1/usage.proto
```diff
@@ -34,10 +34,18 @@ message ListBilledUsageRequest {
...
 }
 
 message ListBilledUsageResponse {
     repeated BilledSession sessions = 1;
+    int64 per_page = 2;
+    int64 total_pages = 3;
+    int64 total = 4;
+    int64 page = 5;
```
easyCZ, line 47: This should also be encapsulated in a Pagination object

components/dashboard/src/teams/TeamUsage.tsx
```diff
@@ -91,6 +66,39 @@ function TeamUsage() {
...
+        } catch (error) {
+            if (error.code === ErrorCodes.PERMISSION_DENIED) {
+                setErrorMessage("Access to usage details is restricted to team owners.");
+            }
+        } finally {
+            setIsLoading(false);
+        }
+    };
+
+    const handleSetPage = async (page: number) => {
```
svenefftinge, line 98: Looks like this function is not doing anything. can we just call `loadPage(page)`?

easyCZ (commented):
Similar to your previous [PR](https://github.com/gitpod-io/gitpod/pull/12377), this PR would greatly benefit from being split into multiple. The following steps lean themselves very nicely.

1. Change proto
2. Implement the UsageService logic, defaulting to Page Size of "Everything" to maintain current behaviour when the number of results is not specified
3. Change Server/dashboard to specify number of results per page, and paginate

As it stands, the big PR causes any changes to the underlying API protocol to force you to change everything else. It's better to agree and commit to the API spec and only then progress with the changes.

The current version (all in one) also suffers from undefined behaviour as the changes rollout across dashboard/server/usage. We do not know if the UI will handle only one page of changes properly if it doesn't yet know about the request per page logic.

components/dashboard/src/teams/TeamUsage.tsx
```diff
@@ -111,12 +119,6 @@ function TeamUsage() {
...
 
-    const calculateTotalUsage = () => {
-        let totalCredits = 0;
-        billedUsage.forEach((session) => (totalCredits += session.credits));
-        return totalCredits.toLocaleString(undefined, { minimumFractionDigits: 2, maximumFractionDigits: 2 });
-    };
-
     const handleMonthClick = (start: any, end: any) => {
         setStartDateOfBillMonth(start);
         setEndDateOfBillMonth(end);
```
svenefftinge, line 124: This doesn't yet trigger a reload and rerender of usage data

components/gitpod-protocol/src/usage.ts
```diff
@@ -40,16 +40,24 @@ export interface ExtendedBillableSession extends BillableSession {
...
-export type BillableWorkspaceType = WorkspaceType;
-
-export enum SortOrder {
-    Descending = 0,
-    Ascending = 1,
+export interface ListBilledUsageResponse {
+    sessions: ExtendedBillableSession[];
+    totalCreditsUsed: number;
+    totalPages: number;
+    total: number;
```
svenefftinge, line 58: I was wondering what `total` is? Maybe call it `totalUsageEntriees` or `totalSessions`.

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2149,48 +2150,60 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
-                        profile = User.getProfile(user);
+        const sessions = await Promise.all(
+            response
+                .getSessionsList()
+                .map((s) => UsageService.mapBilledSession(s))
+                .map(async (session) => {
+                    const ws = await this.workspaceDb.trace(ctx).findWorkspaceAndInstance(session.workspaceId);
+                    let profile: User.Profile | undefined = undefined;
+                    if (session.workspaceType === "regular" && session.userId) {
+                        // TODO add caching to void repeated loading of same profile details here
```
svenefftinge, line 2186: I guess we all know that this is bad in terms of complexity. But for later I think we should add this data redundantly (as well as contextURL)to the actual usage entries instead of optimizing augmentation here.

components/usage-api/usage/v1/usage.proto
AlexTugarev: 👍🏻 thanks for the hint to encapsulate the pagination concerns.

components/usage-api/usage/v1/usage.proto
AlexTugarev: 👍🏻 thanks for the hint to encapsulate the pagination concerns.

components/usage/pkg/db/workspace_instance_usage.go
```diff
@@ -83,9 +111,11 @@ func ListUsage(ctx context.Context, conn *gorm.DB, attributionId AttributionID,
...
 		).
+		Offset(int(offset)).
+		Limit(int(limit)).
 		Find(&usageRecords)
 	if result.Error != nil {
-		return nil, fmt.Errorf("failed to get usage records: %s", result.Error)
+		return nil, 0, 0, fmt.Errorf("failed to get usage records: %s", result.Error)
 	}
-	return usageRecords, nil
+	return usageRecords, count, totalCreditsUsed, nil
```
svenefftinge, line 120: Four return values is a bit much. It would be great to return 2 values (a struct of three plus the err).

components/usage/pkg/db/workspace_instance_usage_test.go
```diff
@@ -178,11 +179,45 @@ func TestListUsageInRange(t *testing.T) {
...
+	results, _, _, err := db.ListUsage(context.Background(), conn, attributionID, start, end, db.DescendingOrder, int64(0), int64(100))
 	require.NoError(t, err)
 
 	require.Len(t, results, 5)
 	require.Equal(t, []db.WorkspaceInstanceUsage{
 		startedInsideFinishedOutside, startedInsideAndStillRunning, startInsideFinishInside, startBeforeFinishInside, startedBeforeAndStillRunning,
 	}, results)
 }
+
+func TestListUsagePagination(t *testing.T) {
```
svenefftinge, line 191: 🧡

AlexTugarev:
> The current version (all in one) also suffers from undefined behaviour as the changes rollout across dashboard/server/usage.

@easyCZ, thanks for the comment. I think we know about the impact pretty well. Considering the timeline, I'd prefer to solve this as a vertical.

>  It's better to agree and commit to the API spec and only then progress with the changes.

In theory this is great, I'm kind of skeptical this works out nicely in a three layered cake. Once you meet the needy requirements from implementing the dashboard, you'll likely start out with the next iteration. 

Let's move general discussions out of PRs, please!

components/usage/pkg/apiv1/usage_test.go
```diff
@@ -170,6 +170,149 @@ func TestUsageService_ListBilledUsage(t *testing.T) {
 	}
 }
 
+func TestUsageService_ListBilledUsage_Pagination(t *testing.T) {
```
svenefftinge, line 173: 🧡

svenefftinge (commented):
Looks super good 🧡 Just a few smaller things.

components/dashboard/src/teams/TeamUsage.tsx
AlexTugarev: indeed, thanks!

components/server/ee/src/workspace/gitpod-server-impl.ts
AlexTugarev: True. That's just a low-hanging perf optimization.

components/usage/pkg/db/workspace_instance_usage.go
AlexTugarev: Yeah, it felt already dirty.

components/dashboard/src/teams/TeamUsage.tsx
AlexTugarev: Actually it does the request which fails with:
> Request listBilledUsage failed with message: 13 INTERNAL: unable to retrieve billed usage

AlexTugarev: > 	usageRecords, total, totalCreditsUsed, err := db.ListUsage(ctx, s.conn, db.AttributionID(in.GetAttributionId()), from, to, order, offset, limit)

`sql.NullFloat64` it should be

components/usage-api/usage/v1/usage.proto
AlexTugarev: though this is nice for the reuse, it's really ugly on call site.
I'm not arguing agains that for the sake of the timeline, so let's do this.

AlexTugarev:
@svenefftinge the [navigation](https://github.com/gitpod-io/gitpod/pull/12501#discussion_r959231945) between months is fixed now.

jldec (approved):
LGTM - thank you for this @AlexTugarev 
/hold because of test commit

PR #12502, opened 2022-08-30
[public-api] Fix stripe webhook secret mounting

easyCZ:
## Description

Ensures that only the key `stripe-webhook-secret` from the secret `stripe-webhook-secret` is mounted as a file, not as a directory with all keys in the secret.

See https://stackoverflow.com/questions/65399714/what-is-the-difference-between-subpath-and-mountpath-in-kubernetes

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/public-api-server/pkg/server/server.go
+++ b/components/public-api-server/pkg/server/server.go
@@ -5,7 +5,6 @@
 package server
 
 import (
-	"encoding/json"
 	"fmt"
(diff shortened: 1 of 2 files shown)


geropl (approved):
Code LGTM, tested and works

PR #12558, opened 2022-08-31
[installer] change blobserve node to workload_ide

### LINKED ISSUE #12366: Workspaces stuck in Pulling state
### Bug description

Self-Hosted K3S installation, latest release (2022.7.2). After installation workspace is stuck in Pulling state. 

After some try and error, deleting network policy blobserve solved the issue. We would like to better understand why this didn't work in the first place.

### Steps to reproduce

Classic install and spin up new workspace, apparently is not something that always happens, so might be difficult to reproduce. See our internal discussion cc: @corneliusludmann 

### Workspace affected

N/A

### Expected behavior

Really this is mostly about understanding the reason behind this problem

### Example repository

_No response_

### Anything else?

Relevant[ internal discussion](https://gitpod.slack.com/archives/C020VCB0U5A/p1661352400563409)

iQQBot:
## Description

This PR move blobserve to `workload_ide` node, because traffic through `ide-proxy`

This PR also adjust blobserve network policy, in order to support some edge case see [internal discuss](https://gitpod.slack.com/archives/C020VCB0U5A/p1661964821598319?thread_ts=1661352400.563409&cid=C020VCB0U5A)

## Related Issue(s)

Fixes #12366

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/installer/cmd/testdata/render/aws-setup/output.golden
+++ b/install/installer/cmd/testdata/render/aws-setup/output.golden
@@ -30,19 +30,7 @@ metadata:
   namespace: default
 spec:
   ingress:
-    - from:
-        - podSelector:
-            matchLabels:
-              component: proxy
-        - podSelector:
-            matchLabels:
-              component: ws-proxy
-        - podSelector:
-            matchLabels:
-              component: ide-proxy
-      ports:
-        - port: 32224
-          protocol: TCP
+    - {}
   podSelector:
     matchLabels:
       app: gitpod
--- a/install/installer/cmd/testdata/render/aws-setup/output.golden
+++ b/install/installer/cmd/testdata/render/aws-setup/output.golden
@@ -6497,7 +6497,7 @@ spec:
           requiredDuringSchedulingIgnoredDuringExecution:
             nodeSelectorTerms:
             - matchExpressions:
-              - key: gitpod.io/workload_workspace_services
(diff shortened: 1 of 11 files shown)


akosyakov:
@iQQBot Should I fix it somehow?
or ask @gitpod-io/engineering-self-hosted to verify?

akosyakov (approved):
lgtm

/hold

please merge when you are comfortable

MrSimonEmms (commented):
I'm broadly ok with this change, but I don't know the implications of changing the network policies. My recollection from when we did this is that it was only implemented in GKE (not EKS, AKS or k3s).

It's worth running the test suites we have. You'll need to load this PR in a workspace, and then run:

```
werft run github -j .werft/gke-installer-tests.yaml -a skipTests=true
werft run github -j .werft/aks-installer-tests.yaml -a skipTests=true
werft run github -j .werft/k3s-installer-tests.yaml -a skipTests=true
werft run github -j .werft/eks-installer-tests.yaml -a skipTests=true
```

There's a possibility that these may fail due to these tests being fairly new, but this should give us an idea if this change creates any regressions.

corneliusludmann (approved):
I tested the PR and it seems to fix the issue. Thank you very much @iQQBot for investigating this! :rocket: 

And sorry that it took so long on my end.

/unhold

PR #12562, opened 2022-08-31
[usage] Fix test for listing usage with pagination

easyCZ:
## Description

There are test hooks for record deletion. However, these wouldn't run after each test when the test setup up was hooked up to the top level `*testing.T` (because the hook would only run after they finished). Instead, we need to bind the hooks to each subtest.

## Related Issue(s)

Fixes unit tests from https://github.com/gitpod-io/gitpod/pull/12562

## How to test

unit tests

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/usage/pkg/apiv1/usage_test.go
+++ b/components/usage/pkg/apiv1/usage_test.go
@@ -175,11 +175,8 @@ func TestUsageService_ListBilledUsage_Pagination(t *testing.T) {
 	ctx := context.Background()
 
 	type Expectation struct {
-		count      int64
 		total      int64
 		totalPages int64
-		page       int64
-		perPage    int64
 	}
 
 	type Scenario struct {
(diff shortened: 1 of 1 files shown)


components/usage/pkg/apiv1/usage_test.go
```diff
@@ -204,112 +201,87 @@ func TestUsageService_ListBilledUsage_Pagination(t *testing.T) {
 	}
 
 	scenarios := []Scenario{
-		(func() Scenario {
```
easyCZ, line 207: This wrapping in a self-invoking anonymous function isn't needed. It would be useful we wanted to do some imperative work, but we're doing only declarative statements.

components/usage/pkg/apiv1/usage_test.go
```diff
@@ -175,11 +175,8 @@ func TestUsageService_ListBilledUsage_Pagination(t *testing.T) {
 	ctx := context.Background()
 
 	type Expectation struct {
-		count      int64
```
easyCZ, line 178: these properties were not used in the test anywhere...

components/usage/pkg/apiv1/usage_test.go
AlexTugarev: 💡  thanks 💯

AlexTugarev (approved):
Awesome! 
Nice lesson.

PR #12563, opened 2022-08-31
[usage] Drop raw sessions from usage report to cut size in half

easyCZ:
## Description

Not really strictly needed. Would be nice, but we can always add later on once we're comfortable with size

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


svenefftinge (approved):
👍

PR #12568, opened 2022-08-31
feat(jetbrains): show workspace class in backend control center

### LINKED ISSUE #12160: Integrate JetBrains backend control center with workspace classes
We now show performance information of the workspace in JetBrains backend control center

- https://github.com/gitpod-io/gitpod/issues/10580
- https://github.com/gitpod-io/gitpod/pull/12053

This issue is to investigate if we can provide a user with a prompt to upgrade workspace class if their workspace performance is observed to be sub-optimal. 

**Note:** The following issue is similar, but not the same:

- https://github.com/gitpod-io/gitpod/issues/11042

Which is to provide a mechanism to actually update the class from within the workspace, rather than just prompting the user.

andreafalzetti:
## Description
This PR adds support in JetBrains Backend Control Center for displaying the Gitpod's Workspace Class, to be accessible and easily discoverable.

The PR also includes small changes to the layout of the Performance Tab to make it easier for the user to understand what metrics are related to the Workspace (relevant) and what metrics are related to the Node (less relevant for them).

| Current                                                                                                                                                              | After this PR                                                                                                                                                        | After this PR (Missing Class)                                                                                                                                        | After this PR (hover on shared node resources)                                                                                                                       |
|----------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Screenshot 2022-09-01 at 23 11 04 | Screenshot 2022-09-02 at 22 58 12 | Screenshot 2022-09-02 at 23 03 59 | Screenshot 2022-09-02 at 22 58 08 |

Considerations
- I also tried adding the class on its own row, below "Workspace", but it would [look odd](https://user-images.githubusercontent.com/2318450/188026745-ffbce5bc-6dca-4abd-9e09-ea1868b3130f.png)

## Related Issue(s)

Fixes #12160 

## How to test
1. Start a workspace with IntelliJ: https://andreafalz6b6c136ea3.preview.gitpod-dev.com/
3. Check the backend control center, you should see the workspace class as in the screenshots above

## Release Notes

```release-note
JetBrains: Provide workspace class info in Backend Control Center 
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodMetricControlProvider.kt
+++ b/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodMetricControlProvider.kt
@@ -5,8 +5,7 @@
 package io.gitpod.jetbrains.remote
 
 import com.jetbrains.ide.model.uiautomation.BeControl
-import com.jetbrains.rd.ui.bedsl.dsl.VerticalGridBuilder
-import com.jetbrains.rd.ui.bedsl.dsl.verticalGrid
+import com.jetbrains.rd.ui.bedsl.dsl.*
 import com.jetbrains.rd.util.lifetime.Lifetime
 import com.jetbrains.rd.util.reactive.Property
 import com.jetbrains.rdserver.diagnostics.BackendDiagnosticsService
@@ -17,10 +16,60 @@ import com.jetbrains.rdserver.unattendedHost.customization.controlCenter.perform
 class GitpodMetricControlProvider : MetricControlProvider {
     override val id: String = "gitpodMetricsControl"
     override fun getControl(lifetime: Lifetime): BeControl {
+        val backendDiagnosticsService = BackendDiagnosticsService.Companion.getInstance()
+
         return verticalGrid {
-            val backendDiagnosticsService = BackendDiagnosticsService.Companion.getInstance()
-            createCpuControl(this, backendDiagnosticsService, lifetime)
-            createMemoryControl(this, backendDiagnosticsService, lifetime)
+            row {
+                horizontalGrid {
+                    column {
+                        label("Workspace")
+                    }
+                }
+            }
+            createWorkspaceHeaderRow(this, backendDiagnosticsService, lifetime)
+            row {
+                verticalGrid {
+                    createCpuControl(this, backendDiagnosticsService, lifetime)
+                    createMemoryControl(this, backendDiagnosticsService, lifetime)
+                }.withMargin { margin(0, 15, 0, 25) }
+            }
+            row {
+                horizontalGrid {
+                    column {
+                        label("Shared Node Resources")
+                    }
+                }.withMargin { margin(0, 0, 0, 15) }.withHelpTooltip("Shared Node Resources", "The shared metrics represent the used and available resources of the cluster node on which your workspace is running")
+            }
+        }
+    }
+
+    private fun createWorkspaceHeaderRow(ctx: VerticalGridBuilder, backendDiagnosticsService: BackendDiagnosticsService, lifetime: Lifetime) {
+        val labelProperty = Property("")
+
+        val workspaceClassMetric = backendDiagnosticsService.getMetric("gitpod_workspace_class")
+        val workspaceClass = workspaceClassMetric.toString()
+
+        fun updateLabel() {
+            if (workspaceClass != "") {
+                labelProperty.set(workspaceClass)
+            }
+        }
+        updateLabel()
+
+        workspaceClassMetric.valueProperty.change.advise(lifetime) {
+            updateLabel()
+        }
+
+        if (workspaceClass == "") {
+            return
+        }
+
+        return ctx.row {
+            horizontalGrid {
+                column {
+                    label(workspaceClass)
+                }
+            }.withMargin { margin(0, 15, 0, 0) }
         }
     }
 
@@ -29,7 +78,7 @@ class GitpodMetricControlProvider : MetricControlProvider {
         val cpuTotal = backendDiagnosticsService.getMetric("gitpod_workspace_cpu_total")
         val cpuPercentage = backendDiagnosticsService.getMetric("gitpod_workspace_cpu_percentage")
         val cpuPercentageProperty = Property("$cpuPercentage %")
-        val label = "Workspace CPU"
+        val label = "CPU"
         val progressBar = createProgressBar(lifetime, cpuPercentage.valueProperty, cpuPercentageProperty)
         val labelProperty = Property("")
 
@@ -51,7 +100,7 @@ class GitpodMetricControlProvider : MetricControlProvider {
         val memoryTotal = backendDiagnosticsService.getMetric("gitpod_workspace_memory_total")
         val memoryPercentage = backendDiagnosticsService.getMetric("gitpod_workspace_memory_percentage")
         val memoryPercentageProperty = Property("$memoryPercentage %")
-        val label = "Workspace Memory"
+        val label = "Memory"
         val progressBar = createProgressBar(lifetime, memoryPercentage.valueProperty, memoryPercentageProperty)
         val labelProperty = Property("")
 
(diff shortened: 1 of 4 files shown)


roboquat:

<details>

</details>

roboquat:

<details>

</details>

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodMetricControlProvider.kt
```diff
@@ -7,20 +7,62 @@ package io.gitpod.jetbrains.remote
 import com.jetbrains.ide.model.uiautomation.BeControl
 import com.jetbrains.rd.ui.bedsl.dsl.VerticalGridBuilder
 import com.jetbrains.rd.ui.bedsl.dsl.verticalGrid
+import com.jetbrains.rd.ui.bedsl.dsl.*
```
andreafalzetti, line 10: This was suggested by IntelliJ but it looks a bit odd to me, having `.dsl.*` and other specific imports 🤔

akosyakov:
> I also tried adding the class on its own row, below "Workspace", but it would [look odd](https://user-images.githubusercontent.com/2318450/188026745-ffbce5bc-6dca-4abd-9e09-ea1868b3130f.png)

@andreafalzetti What if we add at very top?

akosyakov:
👍 I like the separation of Workspace/Node. 
- We are missing disk resource for Workspace, disk resource for Node is misleading.
- Load Average is related to overall Node, not to workspace. We should move it down too?

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/GitpodMetricControlProvider.kt
akosyakov: You can run optimize imports to resolve such things. Usually I don't care about import sections. Let IntellIJ does it thing.

andreafalzetti:
@akosyakov I also prefer the separation Workspace/Node and the class if currently displayed with the Workspace header. If the class displayName is not available, the header is `Workspace` otherwise it will be displayed as `Workspace (Name: Description).` See [screenshot](https://user-images.githubusercontent.com/2318450/188026395-eb024362-524e-48cf-8699-180bc63e79db.png).

> We are missing disk resource for Workspace, disk resource for Node is misleading.

I can look into this as a separate PR 👍 

> What if we add at very top?

Currently, it's not possible to add elements at the very top, because `pingAndLaControl` has `order="first"`

 > Also we don't need to prefix with Class:

There is not `Class:` prefix (that was a test screenshot but I didn't go for that option)

components/ide/jetbrains/backend-plugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -22,18 +22,26 @@
     </dependencies>
 
     <extensions defaultExtensionNs="com.intellij">
-        <applicationService serviceImplementation="io.gitpod.jetbrains.remote.services.HeartbeatService" preload="true"/>
+        <applicationService serviceImplementation="io.gitpod.jetbrains.remote.services.HeartbeatService"
+                            preload="true"/>
         <applicationService serviceImplementation="io.gitpod.jetbrains.remote.GitpodManager" preload="true"/>
         <applicationService serviceImplementation="io.gitpod.jetbrains.remote.GitpodPortsService" preload="true"/>
-        <notificationGroup id="Gitpod Notifications" displayType="BALLOON" isLogByDefault="false" />
+        <notificationGroup id="Gitpod Notifications" displayType="BALLOON" isLogByDefault="false"/>
```
andreafalzetti, line 29: I have enabled auto-formatting of the code on save which made these adjustments. There are no functional changes otherwise.

Happy to revert, but I prefer following the suggestions of IntelliJ :)

components/ide/jetbrains/backend-plugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -22,18 +22,26 @@
     </dependencies>
```
andreafalzetti, line 22: I have enabled auto-formatting of the code on save which made these adjustments. There are no functional changes otherwise.

Happy to revert, but I prefer following the suggestions of IntelliJ :)

components/ide/jetbrains/backend-plugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -22,18 +22,26 @@
...
 
     <extensions defaultExtensionNs="com.intellij">
-        <applicationService serviceImplementation="io.gitpod.jetbrains.remote.services.HeartbeatService" preload="true"/>
+        <applicationService serviceImplementation="io.gitpod.jetbrains.remote.services.HeartbeatService"
```
andreafalzetti, line 25: I have enabled auto-formatting of the code on save which made these adjustments. There are no functional changes otherwise.

Happy to revert, but I prefer following the suggestions of IntelliJ :)

andreafalzetti:
> Can we add it after pingAndLaControl before Workspace section?
@akosyakov it's a matter of preference, I've tried 3 layout and took a screenshot so we can vote/discuss.

| A | B | C |
|---|---|---|
| Screenshot 2022-09-02 at 15 24 19  | Screenshot 2022-09-02 at 15 25 34  |  Screenshot 2022-09-02 at 15 26 42 |

My thoughts:

A: It feels that we now have 2 sections (workspace/node) and the class is outside of the workspace section, so I am not inclined towards this layout very much. At the same time pretty much all data is around the workspace, so I am not against it.

B: The class is just below "workspace" re-enforcing that that value is related. Also, other controls have a label (CPU, Latency, etc). This is the first component without a label in this UI so the fact that is below workspace makes me think it's more obvious that we're referring to the class..

C: This is a bit more explicit that those specs are for the workspace. I've tested with very long text it will just make the widget width larger but it's still visible.

akosyakov:
> I prefer B and C, but I would go with B because it looks cleaner the UI with 2 lines instead of a long one. What about you @akosyakov?

Let's go with B.

felladrin:
> I think it's a matter of preference so I've tried 3 layouts and took a screenshot so we can vote/discuss.

I liked more the option B. The separation seems clearer.
If you could add an icon before the words "Workspace" and "Node", just to make the separation even more clear, it would be great. (Alternatively, I think that adding `:` behind those works would already help.) (Or making them bold.)

akosyakov:
@andreafalzetti I wonder maybe below `Node` we can add some note, like that are resources shared among all users of the node or something like that. cc @loujaybee

felladrin:
> I wonder maybe below Node we can add some note, like that are resources shared among all users of the node or something like that.

How about "Shared resources in this machine:" in place of the word "Node"?

andreafalzetti:
> @andreafalzetti I wonder maybe below Node we can add some note, like that are resources shared among all users of the node or something like that. cc @loujaybee

If we want to keep the UI clean and intuitive, we could consider:

1. Replacing "Node" for something more specific like @felladrin suggested
2. Use a tooltip on "Node" header to provide extra info and not have the text always visible. It won't be discovered by most users, but it will not add extra text always visible.

felladrin:
`Resources shared between workspaces running on the node` sounds good!

andreafalzetti:
@akosyakov @felladrin thanks both for the feedback 🙏 

This is my proposed solution, I went with something sort but still that gives up the idea. I also went for the tooltip because adding more text would make the UI more crowded and once the info is read, is no longer needed (we can re-enforce the concept with external docs, etc):

I've updated the screenshot variations in the main comment, please have a look at let me know if you would like to change anything 👂

felladrin (approved):
Code checked and tested! ✅ 

image

I'll put it on hold to give the chance to @akosyakov to complete his review.

/hold

akosyakov (approved):
lgtm

PR #12577, opened 2022-09-01
[single-cluster/eks] Add `cluster-autoscaler`

Pothulapati:
## Description

This commit adds the `cluster-autoscaler` component into
the terraform scripts through the latest helm chart. This also uses
the `cluster_autoscaler_irsa_role` module to create the relevant
policy, and role which is also created as a service account
and then attached to the cluster-atuoscaler component

## Related Issue(s)

Fixes #

## Release Notes

```release-note
[single-cluster/eks] Add `cluster-autoscaler`
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/infra/single-cluster/aws/Makefile
+++ b/install/infra/single-cluster/aws/Makefile
@@ -26,7 +26,11 @@ plan-cluster:
 	@terraform plan -target=module.eks
 
 .PHONY: plan-tools
-plan-tools: plan-cm-edns plan-cluster-issuer
+plan-tools: plan-cm-edns plan-cluster-issuer plan-cluster-autoscaler
+
+.PHONY: plan-cluster-autoscaler
+plan-cluster-autoscaler:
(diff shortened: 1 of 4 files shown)


install/infra/single-cluster/aws/Makefile
```diff
@@ -41,7 +45,11 @@ apply-cluster:
 	@terraform apply -target=module.eks --auto-approve
 
 .PHONY: apply-tools
-apply-tools: install-cm-edns install-cluster-issuer
+apply-tools: install-cm-edns install-cluster-issuer install-cluster-autoscaler
```
Pothulapati, line 48: @nandajavarma Can you take a look on why this fails? I followed the same pattern but still get the following error:
Screenshot 2022-09-02 at 9 44 27 AM



Seems to be because it was using a stale kubeconfig

nandajavarma (approved):
Changes looks good!

PR #12586, opened 2022-09-01
Align GCP Single Cluster TF Database Output with Installation UI

lucasvaltl:
## Description
Reorders the output of the `make output` command for the GCP single cluster terraform config. Specifically, the "tick Google Cloud SQL proxy" instruction is moved to the top of the database section, because this is also where it is in the UI itself.

## How to test
- Look at changed files :) 

## Release Notes

```release-note
NONE
```


MrSimonEmms (approved):
Not sure about approving such a massive change :)

PR #12602, opened 2022-09-01
fix(jetbrains): avoid connection loops

### LINKED ISSUE #9848: Refresh workspace page or revisit it and you end up in jetbrains spawning hell
### Bug description

If you leave the workspace page and return to it, and, I beleive also just refresh the page. It starts spawning new instances to which there is no escaping.

### Steps to reproduce

Leave open workspace page, return to open workspace page 

Screenshot 2022-05-08 015708

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

andreafalzetti:
## Description
Adds a check to make sure to avoid connecting multiple times to the same workspace. Not having this check in place is causing a loop of connections when trying to open a connection with a workspace that is already opened in another client.

The approach used is to show this message popup while we wait for a new API to request focus to the open IDE.

Screenshot 2022-08-31 at 22 24 05

## Related Issue(s)

Fixes #9848

## How to test
1. Open prev env: <WAITING FOR URL>
2. Select any JetBrains IDE as preferred editor
3. Start a workspace for any repo
4. Try opening multiple IntelliJ sessions ..[WIP]

## Release Notes

```release-note
Fixed JetBrains connection loop when connecting twice to the same workspace 
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/ide/jetbrains/gateway-plugin/src/main/kotlin/io/gitpod/jetbrains/gateway/GitpodConnectionProvider.kt
+++ b/components/ide/jetbrains/gateway-plugin/src/main/kotlin/io/gitpod/jetbrains/gateway/GitpodConnectionProvider.kt
@@ -12,6 +12,7 @@ import com.intellij.ide.BrowserUtil
 import com.intellij.openapi.components.service
 import com.intellij.openapi.diagnostic.thisLogger
 import com.intellij.openapi.progress.ProgressManager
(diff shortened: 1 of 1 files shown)


andreafalzetti:
Closing - Only used to override the prev env!

PR #12612, opened 2022-09-02
[UBP] Add `LastUsageReconciliationTime` RPC

andrew-farries:
## Description

Add a `LastUsageReconcilationTime` RPC to the usage service. The RPC returns the last time of the most recent usage reconciliation.

The intention is to make this information visible on the usage view in the UI in order to give an indication of the 'freshness' of the data on that page.

## Related Issue(s)

Part of https://github.com/gitpod-io/gitpod/issues/12064

## How to test

1. Port forward to the usage service gRPC server: `kubectl port-forward svc/usage 9001:9001`
2. Use `evans -r -p 9001` to connect to it.
3. Invoke the `LastUsageReconcilationTime` RPC on the UsageService.

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


components/usage/pkg/apiv1/usage.go
```diff
@@ -32,6 +32,8 @@ type UsageService struct {
 
 	reportGenerator *ReportGenerator
 
+	lastReconciliation time.Time
```
easyCZ, line 35: This assumes the service is a singleton and that the scheduled job will always run on the same instance. Is that a valid assumption?

Songyuwasd:
求助

components/usage/pkg/apiv1/usage.go
andrew-farries: I think it's safe to assume that the usage service will remain a singleton (ie we won't run multiple replicas of it).

> and that the scheduled job will always run on the same instance

I'm not sure what you mean by this.

An alternative is to store this last reconciliation time in the database - then it will persist across pod restarts. I think the 🛹  is to do it in memory like this as the information returned is intended to be extra information to help make sense of the usage view.

easyCZ: The controller doesn't run on the US clusters currently. How will this value be populated for US clusters?

easyCZ: And related, what happens when the instance starts but the controller hasn't yet run?

andrew-farries: > The controller doesn't run on the US clusters currently. How will this value be populated for US clusters?

That's a good point. I hadn't considered that we no longer run the usage service in the US. That suggests that we need to write this to the database and rely on `db-sync` to keep the value consistent between both EU and US.

andrew-farries: > And related, what happens when the instance starts but the controller hasn't yet run?

In that case the method returns the timestamp corresponding to the zero value of a `time.Time` struct.

andrew-farries:
Closing after [internal discussion](https://gitpod.slack.com/archives/C02EN94AEPL/p1662128497099469) in favour of https://github.com/gitpod-io/gitpod/pull/12624.

PR #12584, opened 2022-09-01
Added Usage table

svenefftinge:
## Description
Introduces a usage table to keep track of any account's credit balance.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

PR #12618, opened 2022-09-02
[usage] Usage ledger API protos

easyCZ:
## Description

Defines the following RPCs on the UsageService:
* ReconcileUsageWithLedger

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


components/usage-api/usage/v1/usage.proto
```diff
@@ -15,6 +15,66 @@ service UsageService {
 
     // GetCostCenter retrieves the spending limit with its associated attributionID
     rpc GetCostCenter(GetCostCenterRequest) returns (GetCostCenterResponse) {}
+
+    // Triggers reconciliation of usage with ledger implementation.
+    rpc ReconcileUsageWithLedger(ReconcileUsageWithLedgerRequest) returns (ReconcileUsageWithLedgerResponse) {}
+
+    // Creates, or updates, a Usage record
+    rpc UpsertUsage(UpsertUsageRequest) returns (UpsertUsageResponse) {}
```
svenefftinge, line 23: Why would you like to expose this on API level? I thought that would be something only the recociler does.

components/usage-api/usage/v1/usage.proto
```diff
@@ -15,6 +15,66 @@ service UsageService {
 
     // GetCostCenter retrieves the spending limit with its associated attributionID
     rpc GetCostCenter(GetCostCenterRequest) returns (GetCostCenterResponse) {}
+
+    // Triggers reconciliation of usage with ledger implementation.
+    rpc ReconcileUsageWithLedger(ReconcileUsageWithLedgerRequest) returns (ReconcileUsageWithLedgerResponse) {}
```
svenefftinge, line 20: Nit: Would be great to find a simpler name (e.g. `reconcileUsageEntries`?) as we are likely deprecate and remove the other usage reconcilation method eventually.

components/usage-api/usage/v1/usage.proto
```diff
@@ -15,6 +15,66 @@ service UsageService {
...
+
+    // Attribution ID for the usage record. Required.
+    string attribution_id = 2;
+
+    // Description of the Usage event for humans.
+    string description = 3;
+
+    // Positive, or negative amount of credits for the Usage event.
+    // Credits are accepted with full precision, but returned with two decimal point precision.
+    double credits = 4;
```
svenefftinge, line 44: Should be `sint64` and `creditCents` ([see](https://github.com/gitpod-io/gitpod/pull/12584#discussion_r960779141)).

components/usage-api/usage/v1/usage.proto
```diff
@@ -15,6 +15,66 @@ service UsageService {
...
+
+    // effective_time is the time when this record has taken effect.
+    google.protobuf.Timestamp effective_time = 5;
```
svenefftinge, line 47: As you know we have the (somewhat bad) habit of storing time as an iso string on application level fields. Which is why I did the same for `effectiveTime` on the DB level.

components/usage-api/usage/v1/usage.proto
easyCZ: The intention here is not to mirror the table. In the table we can store it as credits * 100 rounded to nearest int but for usability from the API, I'd keep it as credits here (also see the comment above). This will avoid translating the credits everywhere into credit cents and back.

components/usage-api/usage/v1/usage.proto
easyCZ: Yep, we'd of course continue to store it as a string on the DB level but on the API level, we get a number of benefits from using the `google.protobuf.Timestamp`. It has good mappings to standard time objects for most languages and provides a fairly standard way to specify the timestamp on the API level.

But can change this into a string for the API as well, it just means an extra layer of parsing on the API (into time, and then again into a VarcharTime)

components/usage-api/usage/v1/usage.proto
easyCZ: To be able to issue a refund for example. Ideally, humans shouldn't really touch the DB and we should go through the RPC. Having this RPC allows us to then expose a UI where you can issue a refund from the admin interface.

components/usage-api/usage/v1/usage.proto
easyCZ: Once we deprecate the old one, we can actually rename this to `ReconcileUsage` as we control all the clients of the API. Happy to rename for now as well but I expect we'll rename this once more down the road.

components/usage-api/usage/v1/usage.proto
easyCZ: Additionally, the `ReconcileUsageWithLedger` would actually call this RPC to update individual records.

svenefftinge: Since we have introduce an ID now and only want to update drafts but never non-drafts we should be explicit about insert (with onConflict do nothing) and updates.

easyCZ: I've dropped this RPC for now.

svenefftinge (approved):

Issue #8024, opened 2022-02-04
[jetbrains] add actions for common gitpod links/features


akosyakov:
Similar to what we have in VS Code:
<img width="533" alt="Screenshot 2022-02-04 at 08 49 52" src="https://user-images.githubusercontent.com/3082655/152491595-0920373d-baa2-4a4f-bd52-840a4a118b84.png">

<img width="655" alt="Screenshot 2022-02-04 at 08 50 50" src="https://user-images.githubusercontent.com/3082655/152491729-b67eb19a-cf7e-451e-a058-34cbd9458626.png">

Not Open in VS Code obviously :D

PR #12621, opened 2022-09-02
Add Gitpod-related actions to JetBrains IDEs

felladrin:
## Description

Add Gitpod-related actions to JetBrains IDEs, similar to [the ones from VS Code](https://github.com/gitpod-io/openvscode-server/blob/35504520bdbd3c4ffaf170ff6356e83caffa7ddb/extensions/gitpod-shared/src/features.ts#L335-L505).

| Actions in Control Center | Actions in Search Menu |
| --- | --- |
| image | Screenshot 2022-09-07 at 15 59 57 |
| Sorted by relevance. We define the order. | Sorted alphabetically. IntelliJ defines the order. |

### Actions added:

- Open Dashboard
- Open Context
- Open Settings
- Open Access Control
- Documentation
- Report Issue
- Follow Us On Twitter
- Open Community Chat
- Upgrade Subscription
- Stop Workspace
- Extend Workspace Timeout
- Share Workspace Snapshot

### Actions out of scope:
- Open in Browser (requires indicator in supervisor frontend to avoid re-opening the IDE)
- Share Running Workspace (currently not supported)

## Related Issue(s)

- #8024

## How to test
- Start a workspace: https://felladrin-6e3aa70a7f.preview.gitpod-dev.com/#https://github.com/gitpod-io/spring-petclinic
- Select any JetBrains IDE (it can be Stable or EAP).
- Open the control center and click the three-dot icon to open the Actions Menu.
- Confirm if the Gitpod actions displayed there work properly by clicking in all of them.
    Screenshot 2022-09-07 at 15 58 57
- Check if the analytics events are received, as you click the actions.
- Double-press Shift key to open the Search Menu, and confirm if searching for "Gitpod: " returns displays all the actions.
    image
- Confirm if the Release Note in this PR is expressing this change accordingly.

## Release Notes

```release-note
JetBrains IDEs now have actions related to Gitpod, which can be accessed via Control Center and via the Search Menu.
```

## Werft options:

- [x] /werft with-preview
- [x]  /werft analytics=segment|TEZnsG4QbLSxLfHfNieLYGF4cDwyFWoe

### DIFF
--- a/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/actions/AccessControlAction.kt
+++ b/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/actions/AccessControlAction.kt
@@ -1,0 +1,23 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package io.gitpod.jetbrains.remote.actions
+
+import com.intellij.openapi.actionSystem.AnAction
+import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.components.service
+import io.gitpod.jetbrains.remote.GitpodManager
+import org.apache.http.client.utils.URIBuilder
+
+class AccessControlAction : AnAction() {
+    private val manager = service<GitpodManager>()
+
+    override fun actionPerformed(event: AnActionEvent) {
+        manager.pendingInfo.thenAccept { workspaceInfo ->
+            URIBuilder(workspaceInfo.gitpodHost).setPath("integrations").build().toString().let { url ->
+                manager.openUrlFromAction(url)
+            }
+        }
+    }
+}
--- a/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/actions/DashboardAction.kt
+++ b/components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/actions/DashboardAction.kt
@@ -1,0 +1,20 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package io.gitpod.jetbrains.remote.actions
+
+import com.intellij.openapi.actionSystem.AnAction
+import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.components.service
+import io.gitpod.jetbrains.remote.GitpodManager
+
+class DashboardAction : AnAction() {
+    private val manager = service<GitpodManager>()
+
+    override fun actionPerformed(event: AnActionEvent) {
+        manager.pendingInfo.thenAccept { workspaceInfo ->
+            manager.openUrlFromAction(workspaceInfo.gitpodHost)
+        }
+    }
+}
--- a/components/ide/jetbrains/backend-plugin/src/main/resources/META-INF/plugin.xml
+++ b/components/ide/jetbrains/backend-plugin/src/main/resources/META-INF/plugin.xml
@@ -44,4 +44,78 @@
                      restartRequired="true"/>
     </extensions>
 
+    <actions>
+        <action id="io.gitpod.jetbrains.remote.actions.StopWorkspaceAction"
+                class="io.gitpod.jetbrains.remote.actions.StopWorkspaceAction"
+                text="Gitpod: Stop Workspace"
+                icon="AllIcons.Actions.Pause">
+            <add-to-group group-id="UnattendedHostDropdownGroup" anchor="last"/>
+        </action>
+        <action id="io.gitpod.jetbrains.remote.actions.DashboardAction"
+                class="io.gitpod.jetbrains.remote.actions.DashboardAction"
+                text="Gitpod: Open Dashboard"
+                icon="AllIcons.General.ProjectTab">
+            <add-to-group group-id="UnattendedHostDropdownGroup" anchor="last"/>
+        </action>
+        <action id="io.gitpod.jetbrains.remote.actions.ExtendWorkspaceTimeoutAction"
+                class="io.gitpod.jetbrains.remote.actions.ExtendWorkspaceTimeoutAction"
+                text="Gitpod: Extend Workspace Timeout"
+                icon="AllIcons.Vcs.History">
+            <add-to-group group-id="UnattendedHostDropdownGroup" anchor="last"/>
+        </action>
+        <action id="io.gitpod.jetbrains.remote.actions.ContextAction"
+                class="io.gitpod.jetbrains.remote.actions.ContextAction"
+                text="Gitpod: Open Context"
+                icon="AllIcons.General.FitContent">
+            <add-to-group group-id="UnattendedHostDropdownGroup" anchor="last"/>
+        </action>
+        <action id="io.gitpod.jetbrains.remote.actions.SettingsAction"
+                class="io.gitpod.jetbrains.remote.actions.SettingsAction"
+                text="Gitpod: Open Settings"
+                icon="AllIcons.General.Settings">
+            <add-to-group group-id="UnattendedHostDropdownGroup" anchor="last"/>
+        </action>
+        <action id="io.gitpod.jetbrains.remote.actions.ShareWorkspaceSnapshotAction"
+                class="io.gitpod.jetbrains.remote.actions.ShareWorkspaceSnapshotAction"
+                text="Gitpod: Share Workspace Snapshot"
+                icon="AllIcons.Actions.Dump">
+            <add-to-group group-id="UnattendedHostDropdownGroup" anchor="last"/>
+        </action>
+        <action id="io.gitpod.jetbrains.remote.actions.AccessControlAction"
+                class="io.gitpod.jetbrains.remote.actions.AccessControlAction"
+                text="Gitpod: Open Access Control"
+                icon="AllIcons.Actions.Properties">
+            <add-to-group group-id="UnattendedHostDropdownGroup" anchor="last"/>
+        </action>
+        <action id="io.gitpod.jetbrains.remote.actions.DocumentationAction"
+                class="io.gitpod.jetbrains.remote.actions.DocumentationAction"
+                text="Gitpod: Documentation"
+                icon="AllIcons.General.ReaderMode">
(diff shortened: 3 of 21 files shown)


components/ide/jetbrains/backend-plugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -22,18 +22,114 @@
...
+                        preload="true"/>
+        <gateway.customization.name
+                implementation="io.gitpod.jetbrains.remote.GitpodGatewayClientCustomizationProvider"/>
+        <gateway.customization.performance id="gitpodMetricsControl" order="before cpuControl"
+                                           implementation="io.gitpod.jetbrains.remote.GitpodMetricControlProvider"/>
+        <gateway.customization.metrics id="gitpodMetricsProvider"
+                                       implementation="io.gitpod.jetbrains.remote.GitpodMetricProvider"/>
+        <registryKey key="gitpod.autoJdk.disabled" defaultValue="false"
+                     description="Disable auto-detection of JDK for the project and its modules"
+                     restartRequired="true"/>
```
felladrin, line 44: ℹ️ Formatted automatically by IntelliJ (or by Githook). No changes other than formatting in this `<extensions>` block.

felladrin:
I've tested it on the Preview Environment and it's working as expected. So I'm now marking it as Ready for Review.

image

image

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/actions/DashboardAction.kt
```diff
@@ -0,0 +1,23 @@
...
+import com.intellij.openapi.components.service
+import io.gitpod.jetbrains.remote.GitpodManager
+import org.apache.http.client.utils.URIBuilder
+
+class DashboardAction : AnAction() {
+    private val manager = service<GitpodManager>()
+
+    override fun actionPerformed(event: AnActionEvent) {
+        manager.pendingInfo.thenAccept { workspaceInfo ->
+            URIBuilder(workspaceInfo.gitpodHost).setPath("workspaces").build().toString().let { url ->
```
andreafalzetti, line 18:
```suggestion
            URIBuilder(workspaceInfo.gitpodHost).build().toString().let { url ->
```

For [consistency with VS Code Browser](https://github.com/gitpod-io/openvscode-server/blob/gp-code/main/extensions/gitpod-shared/src/features.ts#L336-L343) I think we should load the root url of the dashboard, which eventually redirects to `/workspaces`

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/actions/AccessControlAction.kt
```diff
@@ -0,0 +1,23 @@
...
+import com.intellij.openapi.components.service
+import io.gitpod.jetbrains.remote.GitpodManager
+import org.apache.http.client.utils.URIBuilder
+
+class AccessControlAction : AnAction() {
+    private val manager = service<GitpodManager>()
+
+    override fun actionPerformed(event: AnActionEvent) {
+        manager.pendingInfo.thenAccept { workspaceInfo ->
+            URIBuilder(workspaceInfo.gitpodHost).setPath("integrations").build().toString().let { url ->
```
andreafalzetti, line 18: In [VS Code](https://github.com/gitpod-io/openvscode-server/blob/gp-code/main/extensions/gitpod-shared/src/features.ts#L353) the paths are retrieved from [gitpod-protocol](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-protocol/src/util/gitpod-host-url.ts#L108-L110), I would prefer if we could extend the Java binding instead of hardcoding the paths here, it will be painful to remember if they change. 

Did you consider this approach? 🙏

components/ide/jetbrains/backend-plugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -22,18 +22,114 @@
...
     </extensions>
 
+    <actions>
+        <action id="io.gitpod.jetbrains.remote.actions.DashboardAction"
```
andreafalzetti, line 48: Is there any specific logic around the order of the actions?

In VS Code all commands in the palette are sorted alphabetically, and it seems here we have a bit more flexibility to decide. 

I personally would either  go with alphabetical order or alternatively, with relevance. For example, 

- <Contextual to the workflow/workspace> (e.g. stop, go to dashboard)
- <Gitpod Generic> (e.g. docs, upgrade plan)
- <Social> (e.g. chat, twitter)

My preference is having relevant order, but I am interested to know other points of view. cc @felladrin @loujaybee

felladrin: I agree with ordering it by relevance. Currently it's ordered by what I thought was more relevant.
Let me know in which order do you think they should be?

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/actions/AccessControlAction.kt
felladrin: > I am curious to know what do you think are the challenges in going with that approach first?

It's because the I'll be away next week :P So the follow-up PR won't be coming soon. I thought it would be better to have those links already there in this first PR, and replaced in the follow-up (which in turn would be focused in bring the server APIs).

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/actions/DashboardAction.kt
felladrin: Agreed. I'll update it!

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/actions/AccessControlAction.kt
andreafalzetti: Ah, I see. I am not against merging as it is, as long as we allocate time next week. I am happy to work on the follow-up PR.

components/ide/jetbrains/backend-plugin/src/main/resources/META-INF/plugin.xml
andreafalzetti: I think the order you gave is quite good actually, I got confused by the screenshot in which you search for actions and _Follow us on Twitter_ is the second option 😄

components/ide/jetbrains/backend-plugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -22,18 +22,114 @@
...
+                icon="AllIcons.Gutter.ExtAnnotation">
+            <add-to-group group-id="UnattendedHostDropdownGroup" anchor="last"/>
+        </action>
+        <action id="io.gitpod.jetbrains.remote.actions.CommunityChatAction"
+                class="io.gitpod.jetbrains.remote.actions.CommunityChatAction"
+                text="Gitpod: Open Community Chat"
+                icon="AllIcons.General.Balloon">
+            <add-to-group group-id="UnattendedHostDropdownGroup" anchor="last"/>
+        </action>
+        <action id="io.gitpod.jetbrains.remote.actions.InBrowser"
```
andreafalzetti, line 96: The "Open in Browser" is a weird one, because it actually opend supervisor-frontend which then triggers again the opening of JetBrains itself.

It would be cool if we could offer "Open in VS Code Browser" 👀 

Otherwise, I think it's better to remove it, unless I am missing something? cc @akosyakov

felladrin: I agree. Also felt it was not fitting in JetBrains. And couldn't find a way to directly open VS Code when clicking the action. So I'm going to comment it out, and it can be removed in the follow-up PR.

components/ide/jetbrains/backend-plugin/src/main/resources/META-INF/plugin.xml
felladrin: That's a good point! It means it deserves a comment clarifying it on the PR description. I'll comment there, thanks!

components/ide/jetbrains/backend-plugin/src/main/kotlin/io/gitpod/jetbrains/remote/actions/AccessControlAction.kt
felladrin: Thanks, Andrea! I'd appreciate if you could take over this PR/branch and complete it during next week (besides working on the follow-up PR). I'll assign this PR to you too.

components/ide/jetbrains/backend-plugin/src/main/resources/META-INF/plugin.xml
akosyakov: I don't think we should have any mentions of VS Code Browser in IntelliJ. `Open in Browser` sounds good. We also can allow other IDEs in the future. I think for start we can also just drop it.

andreafalzetti: Opening the browser will trigger automatically the JB to open the thinclient, I don't think this action is useful at all, unless I am misunderstanding what it should do? cc @akosyakov

akosyakov: let's remove, if we do it indeed we need some indicators to prevent auto openning.

components/ide/jetbrains/backend-plugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -44,4 +44,94 @@
...
+                icon="AllIcons.Vcs.History">
+            <add-to-group group-id="UnattendedHostDropdownGroup" anchor="last"/>
+        </action>
+        <action id="io.gitpod.jetbrains.remote.actions.ShareWorkspaceSnapshotAction"
+                class="io.gitpod.jetbrains.remote.actions.ShareWorkspaceSnapshotAction"
+                text="Gitpod: Share Workspace Snapshot"
+                icon="AllIcons.Nodes.UpFolder">
+            <add-to-group group-id="UnattendedHostDropdownGroup" anchor="last"/>
+        </action>
+        <action id="io.gitpod.jetbrains.remote.actions.ShareRunningWorkspaceAction"
```
akosyakov, line 121: this one can be removed, live sharing like that does not work with JB, only owner can connect

andreafalzetti:
I've updated the order to reflect the data-driven order and also aligned the event name for opening URLs with VS Code.

/unhold

PR #12646, opened 2022-09-05
Added API call to fetch usage data

svenefftinge:
## Description
Added API call to fetch usage data

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


components/usage-api/usage/v1/usage.proto
```diff
@@ -10,6 +10,9 @@ service UsageService {
     // ListBilledUsage retrieves all usage for the specified attributionId
     rpc ListBilledUsage(ListBilledUsageRequest) returns (ListBilledUsageResponse) {}
 
+    // ListBilledUsage retrieves all usage for the specified attributionId
+    rpc FindUsage(FindUsageRequest) returns (FindUsageResponse) {}
```
easyCZ, line 14: If we followed CRUD more closely then the RPC would generally be `Get/List/Create/Delete/Update` but I'm not too fussed here. It mostly stands out given the RPC above uses `List`

components/usage-api/usage/v1/usage.proto
```diff
@@ -70,6 +73,51 @@ message PaginatedResponse {
...
+    }
+
+    Ordering order = 4;
+
+    PaginatedRequest pagination = 5;
+}
+
+message FindUsageResponse {
+    repeated BilledSession sessions = 1;
+    double total_credits_used = 2;
```
easyCZ, line 99: If I understand this correctly then this would return the total credits for the period in the request, correct? For scenarios like this, the proto is a good place to put comments with this as they are included in all of the code-generated clients so docs can be right there with the SDK

components/usage/pkg/apiv1/usage.go
```diff
@@ -118,6 +118,105 @@ func (s *UsageService) ListBilledUsage(ctx context.Context, in *v1.ListBilledUsa
...
+
+	if from.After(to) {
+		return nil, status.Errorf(codes.InvalidArgument, "Specified From timestamp is after To. Please ensure From is always before To")
+	}
+
+	if to.Sub(from) > maxQuerySize {
+		return nil, status.Errorf(codes.InvalidArgument, "Maximum range exceeded. Range specified can be at most %s", maxQuerySize.String())
+	}
+
+	var order db.Order = db.DescendingOrder
```
easyCZ, line 139: I think the type here is rendundant and could be removed 
```suggestion
	order := db.DescendingOrder
```

components/usage/pkg/apiv1/usage.go
```diff
@@ -118,6 +118,105 @@ func (s *UsageService) ListBilledUsage(ctx context.Context, in *v1.ListBilledUsa
...
+			Kind:                kind,
+			WorkspaceInstanceId: usageRecord.WorkspaceInstanceID.String(),
+			Draft:               usageRecord.Draft,
+			Metadata:            string(usageRecord.Metadata),
+		}
+		usageData = append(usageData, usageDataEntry)
+	}
+
+	usageSummary, err := db.GetUsageSummary(ctx, s.conn,
+		db.AttributionID(in.GetAttributionId()),
```
easyCZ, line 193: I think we should first attempt to parse the attribution using `db.ParseAttributionID` which does a sanity check that the ID is valid.

components/usage/pkg/apiv1/usage.go
```diff
@@ -118,6 +118,105 @@ func (s *UsageService) ListBilledUsage(ctx context.Context, in *v1.ListBilledUsa
...
+		from,
+		to,
+		true,
+	)
+
+	if err != nil {
+		logger.WithError(err).Error("Failed to fetch usage metadata.")
+		return nil, status.Error(codes.Internal, "unable to retrieve usage")
+	}
+	var totalPages = int64(math.Ceil(float64(usageSummary.NumRecordsInRange) / float64(limit)))
```
easyCZ, line 203: This will panic if `limit` is 0. We don't validate the pagination input from the request.

components/usage/pkg/apiv1/usage.go
```diff
@@ -118,6 +118,105 @@ func (s *UsageService) ListBilledUsage(ctx context.Context, in *v1.ListBilledUsa
...
+	if in.Order == v1.ListUsageRequest_ORDERING_ASCENDING {
+		order = db.AscendingOrder
+	}
+
+	var limit int64 = 1000
+	var page int64 = 0
+	var offset int64 = 0
+	if in.Pagination != nil {
+		limit = in.Pagination.PerPage
+		page = in.Pagination.Page
```
easyCZ, line 149: Here we should also validate the numbers are positive integers

components/usage/pkg/db/usage.go
```diff
@@ -69,22 +71,79 @@ func FindAllDraftUsage(ctx context.Context, conn *gorm.DB) ([]Usage, error) {
...
+	result := db.FindInBatches(&usageRecordsBatch, 1000, func(_ *gorm.DB, _ int) error {
+		usageRecords = append(usageRecords, usageRecordsBatch...)
+		return nil
+	})
 	if result.Error != nil {
 		return nil, fmt.Errorf("failed to get usage records: %s", result.Error)
 	}
 	return usageRecords, nil
 }
+
```
easyCZ, line 109: For implementing these, you may [find gorm scopes very applicable](https://gorm.io/docs/scopes.html)

components/usage/pkg/db/usage.go
```diff
@@ -69,22 +71,79 @@ func FindAllDraftUsage(ctx context.Context, conn *gorm.DB) ([]Usage, error) {
...
+	query2 := db.Table((&Usage{}).TableName()).
+		Select("sum(creditCents) as creditCentsBalanceInPeriod", "count(id) as numRecordsInRange").
+		Where("attributionId = ?", attributionId).
+		Where("? <= effectiveTime AND effectiveTime < ?", from, to)
+	if excludeDrafts {
+		query2 = query2.Where("draft = ?", false)
+	}
+	var creditCentsBalanceInPeriod sql.NullInt64
+	var numRecordsInRange sql.NullInt32
+	err = query2.Row().Scan(&creditCentsBalanceInPeriod, &numRecordsInRange)
```
easyCZ, line 140: I believe we shouldn't need to use the `Row().Scan()` API here, and instead should be able to do this with https://gorm.io/docs/advanced_query.html#Count

Happy to try and pair on this, I haven't used it myself yet.

components/usage/pkg/db/usage.go
svenefftinge: Not quite sure how you would like them to be used in this context.

components/usage/pkg/db/usage.go
svenefftinge: But it only gives me the count not the sum, does it?

components/usage/pkg/apiv1/usage.go
```diff
@@ -121,6 +122,116 @@ func (s *UsageService) ListBilledUsage(ctx context.Context, in *v1.ListBilledUsa
...
+
+	order := db.DescendingOrder
+	if in.Order == v1.ListUsageRequest_ORDERING_ASCENDING {
+		order = db.AscendingOrder
+	}
+
+	var limit int64 = 1000
+	var page int64 = 0
+	var offset int64 = 0
+	if in.Pagination != nil {
```
easyCZ, line 151: Grpc guarantees it won't be nil. If its not specified in the request, it will contain default values. For ints that's 0

components/usage/pkg/apiv1/usage.go
```diff
@@ -121,6 +122,116 @@ func (s *UsageService) ListBilledUsage(ctx context.Context, in *v1.ListBilledUsa
...
+	for _, usageRecord := range listUsageResult {
+		kind := v1.Usage_KIND_WORKSPACE_INSTANCE
+		if usageRecord.Kind == db.InvoiceUsageKind {
+			kind = v1.Usage_KIND_INVOICE
+		}
+		attributionId, err := db.ParseAttributionID(string(usageRecord.AttributionID))
+		if err != nil {
+			logger.WithError(err).Errorf("Ignoring usage entry (ID: %s) with invalid attributionID: %s.", usageRecord.ID, usageRecord.AttributionID)
+			continue
+		}
```
easyCZ, line 192: Persoanlly, I'd return an error here from the RPC. Yes, it would cause it to fail but it would more acutely show up and give us a chance to fix. There's a tendency that these logs are missed and never properly fixed.

I think I may have confused you in the previous comment on this. I think when we take the attribution ID from the DB, we can trust it to be the right shape so doing just `db.AttributionID(usageRecord.AttributionID)` should be fine. But I think we can do better. We can actually type the UsageRecord as an AttributionID directly, this removes this step.

But we should try to parse the Attribution ID from the incoming request. We currently use it on line 208 without parsing (and therefore at least some validation).

svenefftinge: Yes, I got indeed confused (somehow the comments show up on the wrong lines in VS Code sometimes). Makes much more sense now.

components/usage/pkg/apiv1/usage.go
```diff
@@ -121,6 +122,116 @@ func (s *UsageService) ListBilledUsage(ctx context.Context, in *v1.ListBilledUsa
...
+
+	var perPage int64 = 50
+	if in.Pagination.PerPage > 0 {
+		perPage = in.Pagination.PerPage
+	}
+	var page int64 = in.Pagination.Page
+	var offset int64 = perPage * page
+
+	listUsageResult, err := db.FindUsage(ctx, s.conn, &db.FindUsageParams{
+		AttributionId: db.AttributionID(in.GetAttributionId()),
```
easyCZ, line 169:
```suggestion
		AttributionId: attributionId,
```

easyCZ: Also, in Go it's a common style to capitalize shorthands. In this case, `ID` instead of `Id`

components/usage/pkg/db/usage.go
easyCZ: The intention was to use deserialize into the `UsageSummary` object directly. If you were to add a `gorm` annotation to the `UsageSummary`, you can have those annotations match the returned column names (creditCentsBalanceAtStart, ...)

easyCZ (approved):
LGTM, left a couple of last comments but these can happen in a follow-up PR

PR #12648, opened 2022-09-05
[gitpod-protocol] move generation of go types to prebuilds and werft

akosyakov:
## Description

We need to keep them up to date. But it slows down each commit if done in git hook, so instead we do it in prebuilds and in leeway job on werft.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONOE
```

## Werft options:

- [ ] /werft with-preview


csweichel (approved):
Thank you :)

geropl (approved):
:heart_eyes: Ah, this is where all the time went - THANK YOU!

PR #12663, opened 2022-09-05
Cronjob that creates artificial traffic for new preview environments jobs.

ArthurSens:
## Description

Once ready, this PR will create a cronjob that triggers new werft jobs which, in their turn, will create new preview environments.

The final goal is that we get a little more traffic for new preview environments, so we can be more proactive when they start failing.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/.werft/platform-trigger-artificial-job.sh
+++ b/.werft/platform-trigger-artificial-job.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+#
+# The easiest way to run this script is through Werft so you don't have to worry
+# about installing the appropraite service account etc. locally.
+#
+#   werft job run github -j .werft/platform-trigger-artificial-job.yaml -s .werft/platform-trigger-artificial-job.sh
+#
+
+sleep 1
+
+set -Eeuo pipefail
+
+werft log phase "Trigger new job with-preview" "Trigger new job with preview environment"
+sudo chown -R gitpod:gitpod /workspace
+git config --global user.name roboquat
+git config --global user.email roboquat@gitpod.io
+
+git checkout -b "platform-artificial-job-$(echo $RANDOM | md5sum | head -c 20; echo;)"
+
+werft run github -a with-preview=true
--- a/.werft/platform-trigger-artificial-job.yaml
+++ b/.werft/platform-trigger-artificial-job.yaml
@@ -45,4 +45,4 @@ pod:
         - .werft/platform-trigger-artificial-job.sh
 
 plugins:
-  cron: "15 */4 * * *"
+  cron: "*/30 * * * *" # Every 30 minutes
--- a/.werft/platform-trigger-artificial-job.yaml
+++ b/.werft/platform-trigger-artificial-job.yaml
@@ -1,0 +1,48 @@
+# This job is responsible for periodically start a new werft job that spins up
+# a preview environment. It is needed to create artificial traffic for the
+# Preview Environment starts SLO. It will help us to be more proactive
+# to identify failures in the pipeline.
+#
+#   werft job run github -j .werft/platform-trigger-artificial-job.yaml
+#
+pod:
+  serviceAccount: werft
+  affinity:
+    nodeAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        nodeSelectorTerms:
+          - matchExpressions:
+              - key: dev/workload
+                operator: In
+                values:
+                  - "builds"
+  volumes:
+    - name: github-token-gitpod-bot
+      secret:
+        defaultMode: 420
+        secretName: github-token-gitpod-bot
+  containers:
+    - name: build
+      image: eu.gcr.io/gitpod-core-dev/dev/dev-environment:aledbf-dl.1
+      workingDir: /workspace
+      imagePullPolicy: IfNotPresent
+      volumeMounts:
+        - mountPath: /mnt/secrets/github-token-gitpod-bot
+          name: github-token-gitpod-bot
+      env:
+        - name: ROBOQUAT_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: github-roboquat-automatic-changelog
+              key: token
+        # Used by the Werft CLI through werft-credential-helper.sh
+        - name: WERFT_GITHUB_TOKEN_PATH
+          value: "/mnt/secrets/github-token-gitpod-bot/token"
+        - name: WERFT_CREDENTIAL_HELPER
+          value: "/workspace/dev/preview/werft-credential-helper.sh"
+      command:
+        - bash
+        - .werft/platform-trigger-artificial-job.sh
+
+plugins:
+  cron: "15 */4 * * *"


ArthurSens:
@gitpod-io/platform , I'm currently failing to configure git in the [job](https://werft.gitpod-dev.com/job/gitpod-custom-main.166) 😭 Can you identify anything I'm during wrong?

.werft/platform-trigger-artificial-job.yaml
```diff
@@ -0,0 +1,48 @@
+# This job is responsible for cordoning Werft build nodes if their disk is almost full.
```
mads-hartmann, line 1: Nit: this comment is outdated ☺️

.werft/platform-trigger-artificial-job.sh
```diff
@@ -0,0 +1,30 @@
...
+# about installing the appropraite service account etc. locally.
+#
+#   werft job run github -j .werft/platform-trigger-artificial-job.yaml -s .werft/platform-trigger-artificial-job.sh
+#
+
+sleep 1
+
+set -Eeuo pipefail
+
+# Activate service account and install core-dev context
```
mads-hartmann, line 13: Do we need this? The werft CLI should be using gRPC so we shouldn't need access to the core-dev cluster for that anymore at least ☺️

.werft/platform-trigger-artificial-job.sh
```diff
@@ -0,0 +1,30 @@
...
+gcloud auth activate-service-account --key-file "/mnt/secrets/gcp-sa/service-account.json"
+gcloud container clusters get-credentials core-dev --zone europe-west1-b --project gitpod-core-dev
+
+echo "[Creating new branch|PHASE] Trigger new job with preview environment"
+echo "[Creating new branch] Setting up user config"
+git config --global user.name roboquat
+git config --global user.email roboquat@gitpod.io
+git remote set-url origin https://oauth2:"${ROBOQUAT_TOKEN}"@github.com/gitpod-io/gitpod.git
```
mads-hartmann, line 21: I think this fails because you're missing a `sudo chown -R gitpod:gitpod /workspace` - at least all other jobs seems to have this ☺️

.werft/platform-trigger-artificial-job.sh
```diff
@@ -0,0 +1,30 @@
...
+
+sleep 1
+
+set -Eeuo pipefail
+
+# Activate service account and install core-dev context
+gcloud auth activate-service-account --key-file "/mnt/secrets/gcp-sa/service-account.json"
+gcloud container clusters get-credentials core-dev --zone europe-west1-b --project gitpod-core-dev
+
+echo "[Creating new branch|PHASE] Trigger new job with preview environment"
```
mads-hartmann, line 17: This could be `werft log phase "Creating new branch" "Trigger new job with preview environment"`

.werft/platform-trigger-artificial-job.sh
```diff
@@ -0,0 +1,30 @@
...
+echo "[Creating new branch] Setting up user config"
+git config --global user.name roboquat
+git config --global user.email roboquat@gitpod.io
+git remote set-url origin https://oauth2:"${ROBOQUAT_TOKEN}"@github.com/gitpod-io/gitpod.git
+
+echo "[Creating new branch] Checking out branch"
+git checkout -b "artificial-job-$(echo $RANDOM | md5sum | head -c 20; echo;)"
```
mads-hartmann, line 24: maybe preview the branch name with something like "platform-" to make it easier for other people to undertand the ownership

.werft/platform-trigger-artificial-job.sh
```diff
@@ -0,0 +1,30 @@
...
+echo "[Creating new branch] Creating commit"
+git commit -m "Creating an artificial traffic for the Preview Environment start SLO" --allow-empty
+echo "[Creating new branch|DONE] Trigger new job with preview environment"
+
+echo "[Trigger new job with-preview|PHASE] Trigger new job with preview environment"
+werft run github -a with-preview=true
```
mads-hartmann, line 30: I think we need to poll for the job to finish so we can delete the branch again, otherwise we'll create a ton of branches ☺️

.werft/platform-trigger-artificial-job.sh
```diff
@@ -0,0 +1,30 @@
...
+echo "[Creating new branch] Setting up user config"
+git config --global user.name roboquat
+git config --global user.email roboquat@gitpod.io
+git remote set-url origin https://oauth2:"${ROBOQUAT_TOKEN}"@github.com/gitpod-io/gitpod.git
+
+echo "[Creating new branch] Checking out branch"
+git checkout -b "artificial-job-$(echo $RANDOM | md5sum | head -c 20; echo;)"
+echo "[Creating new branch] Creating commit"
+git commit -m "Creating an artificial traffic for the Preview Environment start SLO" --allow-empty
+echo "[Creating new branch|DONE] Trigger new job with preview environment"
```
mads-hartmann, line 27: Could be `echo "Trigger new job with preview environment" | werft log slice "Creating new branch" --done`

.werft/platform-trigger-artificial-job.sh
```diff
@@ -0,0 +1,30 @@
...
+
+echo "[Creating new branch|PHASE] Trigger new job with preview environment"
+echo "[Creating new branch] Setting up user config"
+git config --global user.name roboquat
+git config --global user.email roboquat@gitpod.io
+git remote set-url origin https://oauth2:"${ROBOQUAT_TOKEN}"@github.com/gitpod-io/gitpod.git
+
+echo "[Creating new branch] Checking out branch"
+git checkout -b "artificial-job-$(echo $RANDOM | md5sum | head -c 20; echo;)"
+echo "[Creating new branch] Creating commit"
```
mads-hartmann, line 25: could be `echo "Creating commit" | werft log slice "Creating new branch"`

.werft/platform-trigger-artificial-job.sh
```diff
@@ -0,0 +1,30 @@
...
+git commit -m "Creating an artificial traffic for the Preview Environment start SLO" --allow-empty
```
mads-hartmann, line 26: You'll need to add `| werft log slice "Creating commit"` if you want the output to be part of the slice.

.werft/platform-trigger-artificial-job.yaml
```diff
@@ -0,0 +1,48 @@
...
+      requiredDuringSchedulingIgnoredDuringExecution:
+        nodeSelectorTerms:
+          - matchExpressions:
+              - key: dev/workload
+                operator: In
+                values:
+                  - "builds"
+  volumes:
+    # Needed to talk to the core-dev cluster and SSH to underlying instances
+    - name: gcp-sa
```
mads-hartmann, line 19: I don't think we need this secret in this job ☺️

.werft/platform-trigger-artificial-job.sh
mads-hartmann: This is done a few other places, e.g [here's how](https://github.com/gitpod-io/gitpod/blob/main/.werft/workspace-run-integration-tests.sh#L99) to workspace integration test jobs currently does it - we'd need something slightly different as we don't care if the preview environment failed or not, we still want to delete the branch.

Alternatively we'd need some other mechanism of deleting old branches. E.g. this job could even just start by delete all branches that match the pattern that are more than x hours old (that might be a more robust way to deal with it actually)

mads-hartmann (commented):
I left a comment on how I think you can fix the issue you're seeing ☺️ I also left a few other pointers and nits

.werft/platform-trigger-artificial-job.sh
ArthurSens: aaaaah good point. It will just work out-of-the-box? No need to configure anything with grpc?

.werft/platform-trigger-artificial-job.sh
ArthurSens: > I think we need to poll for the job to finish so we can delete the branch again, otherwise we'll create a ton of branches ☺️

I am creating tons of branches but I'm never pushing them to `origin`, so it should be fine right? I don't want them to stick around anyway, so it will even get cleaned up faster if it doesn't get pushed

ArthurSens:
We've got an artificial job running 🙂 https://werft.gitpod-dev.com/job/gitpod-build-platform-artificial-job-7f30eab068977bb4e4d2.0

.werft/platform-trigger-artificial-job.sh
mads-hartmann: Oh nice, I'm surprised this works to be honest 😝 I guess Werft doesn't validate that the Ref actually exist (e.g. [refs/heads/platform-artificial-job-7f30eab068977bb4e4d2](https://github.com/gitpod-io/gitpod/tree/refs/heads/platform-artificial-job-7f30eab068977bb4e4d2)), and it only really cares about the Revision when checking out the code ☺️ so as long as we don't use the Ref in the build job we should be good.

The only potential downside I can think of is if we decide to GC the VM because the branch doesn't exist. I can't remember if our heuristics were updated to keep VMs alive for at least 2h ☺️

.werft/platform-trigger-artificial-job.sh
```diff
@@ -0,0 +1,20 @@
...
+#
+
+sleep 1
+
+set -Eeuo pipefail
+
+werft log phase "Trigger new job with-preview" "Trigger new job with preview environment"
+sudo chown -R gitpod:gitpod /workspace
+git config --global user.name roboquat
+git config --global user.email roboquat@gitpod.io
```
mads-hartmann, line 16: If we're not pushing to origin I wonder if we even need these three lines at all ☺️

mads-hartmann (approved):
Looks good to me. Added hold in case you want to try to remove those git config lines

.werft/platform-trigger-artificial-job.sh
ArthurSens: Unfortunately it is still needed :(

ArthurSens: image

PR #12665, opened 2022-09-06
[installer-tests] fix azure authentication

nandajavarma:
## Description

Currently, for the cleanup job, we try to get the kubeconfig file so as to delete the gitpod installation, external-dns etc. that will remove loadbalancers and such. In azure, the aws cli is currently throwing an error saying it is not logged in. It turns out, it isn't automatically picking up the env vars anymore, we have to explicitly login using them. This PR logs in using azure cli before getting the kubeconfig.

## Related Issue(s)

Fixes #

## How to test

You can try exporting the suggested env vars, and running `make azure-kubeconfig` from `install/tests/` directory.

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/tests/Makefile
+++ b/install/tests/Makefile
@@ -52,6 +52,7 @@ gcp-kubeconfig:
 
 ## azure-kubeconfig: Get the kubeconfig configuration for Azure AKS
 azure-kubeconfig:
+	az login --service-principal -u $$ARM_CLIENT_ID -p $$ARM_CLIENT_SECRET  --tenant $$ARM_TENANT_ID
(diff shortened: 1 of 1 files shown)


Pothulapati (approved):
Haven't specifically tested it, but manually logging in using the `az` cli before we can retrieve the kubeconfig doesn't sound wrong.

Looks like it is a [known issue](https://github.com/Azure/azure-cli/issues/10241) where the CLI has to write to disk, and cannot directly use the env variables.

Issue #10649, opened 2022-06-14
[Self-Hosted] Track activated users


lucasvaltl:
## Is your feature request related to a problem? Please describe
To better understand the utilisation of Gitpod within a specific installation (a.k.ka. activation rate), we want to enhance the current telemetry we send with the amount of `activated users`. This metric is defined as the number of users that have started a workspace in at least 3 of the last 7 days.

## Describe the behaviour you'd like
- Daily telemetry dump includes a total count of actiated users (as defined above). We are NOT interested in specific users, and especially not in personally identifiable info from a user. Just a total number
- This is well documented and clarified to the user: It is shown in the telemetry preview in the admin dashboard (should happen by default but worth checking)
- Need to ensure that there is no impact on performance due to the query

## Describe alternatives you've considered
- We could run the following query against the DB and add the result to the telemetry:
``` 
SELECT COUNT(DISTINCT(ws.ownerId))
FROM d_b_workspace AS ws
WHERE ws.id IN (
   SELECT wsi.workspaceId
   FROM d_b_workspace_instance AS wsi
   WHERE wsi.creationTime within last 7 days
   GROUP BY wsi.workspaceId
   HAVING COUNT(wsi.id) > 3
)
```

## Additional context
- See internal slack [thread](https://gitpod.slack.com/archives/C02EN94AEPL/p1655201623131729)
- Expected implementation time: 1 Day

lucasvaltl:
@Pothulapati I know that you were going to look into this, did you make any progress?

Pothulapati:
@lucasvaltl Not yet! Haven't spent anytime at all on this.

Unassigning hence, and will assign me back once I start working to prevent confusion.

Pothulapati:
Started working on this, and hence assigned myself

Pothulapati:
@lucasvaltl The query that you mentioned seems to calculate **users who created atleast 3 workspace instances of a single workspace in the last 7 days** i.e they could all be on the same day, and not **users who created workspace instances in 3 days of the last 7 days**.

I'm trying for the later now, Feel free to tell me otherwise if that's not what you need.

jakobhero:
> The query that you mentioned seems to calculate users who created atleast 3 workspace instances of a single workspace in the last 7 days i.e they could all be on the same day, and not users who created workspace instances in 3 days of the last 7 days.

i've had a look and agree, this query should do the trick instead:
```
select
    count(userId) as activatedUsers
from (
    select
        u.userId
[...]

PR #12668, opened 2022-09-06
[telemetry] Add Active users

Pothulapati:
## Description

This PR adds telemetry to track active users.
Active users are calculated as the ones who
created workspace instances in 3 of the last 7 days.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/10649

## Release Notes

```release-note
[telemetry] Add Active users
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-db/src/typeorm/workspace-db-impl.ts
+++ b/components/gitpod-db/src/typeorm/workspace-db-impl.ts
@@ -1055,6 +1055,24 @@ export abstract class AbstractTypeORMWorkspaceDBImpl implements WorkspaceDB {
         return await queryBuilder.getCount();
     }
 
+    public async getActiveUserCount(): Promise<number> {
+        const workspaceInstanceRepo = await this.getWorkspaceInstanceRepo();
+        const since = new Date();
+        since.setDate(since.getDate() - 7);
+
+        // get the owners of workspaces with new instances created at-least
+        // 3 days of the last 7 days
+        const queryBuilder = workspaceInstanceRepo
+            .createQueryBuilder("i")
+            .select("u.userId")
+            .where("i.startedTime  > :since", { since: since.toISOString() })
+            .innerJoinAndMapOne("i.user", DBWorkspaceInstanceUser, "u", "i.id = u.instanceId")
+            .groupBy("1")
+            .having("count(distinct date(i.startedTime)) >= 3");
+
+        return await queryBuilder.getCount();
+    }
+
     public async findAllWorkspaceAndInstances(
         offset: number,
         limit: number,
--- a/components/gitpod-db/src/typeorm/workspace-db-impl.ts
+++ b/components/gitpod-db/src/typeorm/workspace-db-impl.ts
@@ -1070,6 +1070,9 @@ export abstract class AbstractTypeORMWorkspaceDBImpl implements WorkspaceDB {
             .groupBy("1")
             .having("count(distinct date(i.startedTime)) >= 3");
 
+        // print query for debugging purposes
+        log.debug(queryBuilder.getSql());
+
         return await queryBuilder.getCount();
     }
 
--- a/components/gitpod-db/src/typeorm/workspace-db-impl.ts
+++ b/components/gitpod-db/src/typeorm/workspace-db-impl.ts
@@ -1055,20 +1055,25 @@ export abstract class AbstractTypeORMWorkspaceDBImpl implements WorkspaceDB {
         return await queryBuilder.getCount();
     }
 
-    public async getActiveUserCount(): Promise<number> {
-        const workspaceInstanceRepo = await this.getWorkspaceInstanceRepo();
+    public async countUsersWithActiveInstances(): Promise<number> {
+        const workspaceInstanceUserRepo = await this.getWorkspaceInstanceUserRepo();
         const since = new Date();
         since.setDate(since.getDate() - 7);
 
         // get the owners of workspaces with new instances created at-least
         // 3 days of the last 7 days
-        const queryBuilder = workspaceInstanceRepo
-            .createQueryBuilder("i")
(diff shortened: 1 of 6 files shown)


roboquat:

<details>

</details>

components/gitpod-db/src/typeorm/workspace-db-impl.ts
```diff
@@ -1032,6 +1032,24 @@ export abstract class AbstractTypeORMWorkspaceDBImpl implements WorkspaceDB {
...
+
+        // get the owners of workspaces with new instances created at-least
+        // 3 days of the last 7 days
+        const queryBuilder = workspaceInstanceRepo
+            .createQueryBuilder("i")
+            .select("u.userId")
+            .where("i.startedTime  > :since", { since: since.toISOString() })
+            .innerJoinAndMapOne("i.user", DBWorkspaceInstanceUser, "u", "i.id = u.instanceId")
+            .groupBy("1")
+            .having("count(distinct date(i.startedTime)) >= 3");
```
svenefftinge, line 1048: I think there's no need to join this over d_b_workspace_instance as `lastSeen` on `d_b_workspace_instance_usage` could be used.
E.g. something like:
```sql
SELECT count(distinct userid) from gitpod.d_b_workspace_instance_user where lastSeen> <since>
```

svenefftinge:
For metrics like this, IMHO it makes more sense to start a metric-central DB facade or directly use and access the DB instead of adding global queries to the ORM layer. (Disclaimer: This is not meant to block this PR, but more to share my general thoughts on this and maybe start a discussion.)

components/gitpod-db/src/typeorm/workspace-db-impl.ts
Pothulapati: @svenefftinge Hmm. So, Doing that would give us instances that are last seen > 7 days, but here we are looking for `users of workspaces with new instances created at-least 3 days of the last 7 days` which means we somehow need to relate the instance with `d_b_workspace_instance.startedTime` (to group by day and then count(unique days) >=3) and hence the join. Without `startedTime`, I'm not sure how we can do that? :thinking:

Pothulapati:
> IMHO it makes more sense to start a metric-central DB facade or directly use and access the DB instead of adding global queries to the ORM layer.

In the interest of time (i.e release next week), I've raised https://github.com/gitpod-io/gitpod/issues/13196 instead of changing that here.

components/gitpod-db/src/typeorm/workspace-db-impl.ts
Pothulapati: Hmm, We might be able to use `lastSeen` itself instead of `startedTime` which means "users with workspaces instances that were last seen in at-least 3 of the 7 days`. Is that what you are suggesting? 🤔

geropl: @Pothulapati This query looks veeeeery expensive. Can you validate the impact on DB CPU using [one of our read-only copies](https://www.notion.so/gitpod/Accessing-the-database-242e9fa0fef642128380a10d800d3b99#75a5c6647483496f8f085e43c8b9178d)? :thinking:

svenefftinge: This never runs in Saas, does it?

Pothulapati: Correct! The telemetry job and the setting page is disabled in SAAS. It's only aimed for self hosted instances.

Pothulapati: I've updated this query to be a lot more closer to what's in the Issue, and that works on BigQuery (takes around of 2 minutes there to execute but this won't be happening like mentioned above)

But I get the following error:

```
TypeError: Cannot read properties of undefined (reading 'getParameters')\n
```

Any idea why? 🤔

easyCZ:
Closing as inactive. If this is still required, please re-open.

lucasvaltl:
I believe this is still needed - but not sure what the current state is. @Pothulapati could you say what is needed from the webapp side?

Pothulapati:
Correct @lucasvaltl , I'm struck on the error that I [mentioned in the comment](https://github.com/gitpod-io/gitpod/pull/12668#discussion_r978269916).

Would really appreciate any help @easycz

easyCZ:
Looking at this, would it possibly be more useful to get active workspaces (with their user IDs) in the last 7 days rather than approximate active users based on workspaces? It feels misleading as an active user doesn't need to run workspaces.

How are you planning to use the active users?

We'd do this by returning a list of Workspace objects in the Telemetry data.

Pothulapati:
> Looking at this, would it possibly be more useful to get active workspaces (with their user IDs) in the last 7 days rather than approximate active users based on workspaces? It feels misleading as an active user doesn't need to run workspaces.

The decision to do this was made [after talking](https://github.com/gitpod-io/gitpod/issues/10649#issuecomment-1243899157) to @jakobhero. I think, We calculate it like this so that we can be similar to the SAAS active user metric. There's also a long chat about [this](https://gitpod.slack.com/archives/C01KLC56NP7/p1662372204290429).

> How are you planning to use the active users?

The plan is to use it as part of existing telemetry mode dashboards, to understand how self-hosted Gitpod is being used.

Pothulapati:
Updated the comments, but still need to fix the typeorm error that I'm getting

```
TypeError: Cannot read properties of undefined (reading 'getParameters')\n
```
🤔

Pothulapati:
Closing this in favour of https://github.com/gitpod-io/gitpod/pull/13814, because of some [issues with the recent changes in CI.](https://gitpod.slack.com/archives/C032A46PWR0/p1665596747954979)

PR #12673, opened 2022-09-06
[db-sync] sync d_b_usage

svenefftinge:
## Description
Adds `d_b_usage` to the list of synced tables.
I intentionally didn't specify/add a deleted column, because we don't have a deletion mechanism in place, yet.  I hope we will no longer use `db-sync` when we introduce it 😅

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


easyCZ (approved):
I've only skimmed the code but I'm assuming you know db-sync is happy without the deletion column specified.

PR #12319, opened 2022-08-23
Installer config refactoring

MrSimonEmms:
## Related Issue(s)

Fixes #

## Release Notes

```release-note
```

## Werft options:

- [ ] /werft with-preview

PR #12679, opened 2022-09-06
[installer]: create config command and deprecate init command

### LINKED ISSUE #12678: Create a `config` command
As part of the refactored installation process, a key part of the application will be the configuration generation.

To that end, there should be a new `config` root command which everything will stem from. This will be about manipulating a config file that is then submitted to the Installer for rendering into the generated YAML that can be deployed to the Kubernetes cluster.

As part of this, we should also deprecate the `init` command and replace with a new `config init`. This will generate the same config file as before, except that it will save it to the file

### LINKED ISSUE #12847: Update the SaaS deployment process to use the refactored `config` commands
As part of the parent task, the Installer's configuration commands have updated. These will need to be incorporated into the SaaS process for building the YAML files for deployment to our clusters.

- [ ] The `init` command is now deprecated in favour of `config init` #12678 
- [ ] There are detection commands for the ShiftFS status and containerd location (this may not be needed) #12198 
- [ ] There is a new `config build-from-envvars` command to build the config parameters (this may not be needed) #12199

MrSimonEmms:
## Description

Create a `config init` command and deprecate the `init` command.

The two commands are functionally identical, with the exception of the new command saving the file for you (the previous command just printed the output).

This is part of a major refactoring of the configuration so that it's simpler and just config generation. See #12319 for my notepad of how it'll look eventually.

## Related Issue(s)

Fixes #12678
Fixes #12847

## How to test

Run `go run . config init` and check `gitpod.config.yaml`

## Release Notes

```release-note
[installer]: create config command and deprecate init command
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/install/installer/cmd/config.go
+++ b/install/installer/cmd/config.go
@@ -1,0 +1,64 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the MIT License. See License-MIT.txt in the project root for license information.
+
+package cmd
+
+import (
+	"errors"
+	"os"
+	"path/filepath"
+
+	"github.com/gitpod-io/gitpod/common-go/log"
+	"github.com/gitpod-io/gitpod/installer/pkg/config"
+	"github.com/spf13/cobra"
+	"k8s.io/utils/pointer"
+)
+
+var configOpts struct {
+	ConfigFile string
+}
+
+// configCmd represents the validate command
+var configCmd = &cobra.Command{
+	Use:   "config",
+	Short: "Perform configuration tasks",
+}
+
+// configFileExists checks if the configuration file is present on disk
+func configFileExists() (*bool, error) {
+	if _, err := os.Stat(configOpts.ConfigFile); err == nil {
+		return pointer.Bool(true), nil
+	} else if errors.Is(err, os.ErrNotExist) {
+		return pointer.Bool(false), nil
+	} else {
+		return nil, err
+	}
+}
+
+// saveConfigFile converts the config to YAML and saves to disk
+func saveConfigFile(cfg interface{}) error {
+	fc, err := config.Marshal(config.CurrentVersion, cfg)
+	if err != nil {
+		return err
+	}
+
+	err = os.WriteFile(configOpts.ConfigFile, fc, 0644)
+	if err != nil {
+		return err
+	}
+
+	log.Infof("File written to %s", configOpts.ConfigFile)
+
+	return nil
+}
+
+func init() {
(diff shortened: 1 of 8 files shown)


ArthurSens:
Hey Simon, did you notice that this change is breaking deployments to preview environments? 😬

MrSimonEmms:
@ArthurSens is that [this instance](https://sje-instal2d24543f57.preview.gitpod-dev.com/workspaces)? This looks to be installing ok now

ArthurSens:
Ah yes, sorry for misreading the results :) 

@mads-hartmann, as simon points out, I don't think we need to change the installer plugin in ArgoCD/workspace clusters/preview right now, but should we add issues to the schedule?

MrSimonEmms:
@ArthurSens phew.

I've created #12847 to capture the SaaS deployment work

mads-hartmann:
@ArthurSens @MrSimonEmms For SaaS deployment we only use the render command. I left a comment [here](https://github.com/gitpod-io/gitpod/issues/12847#issuecomment-1242654116) ☺️

mads-hartmann:
> workspace clusters
They also don't use the init command as far as I can tell ([ref](https://github.com/gitpod-io/ops/blob/main/deploy/workspace/gitpod-up.sh#L215))
> previews
@MrSimonEmms could I persuade you to update [this line](https://github.com/gitpod-io/gitpod/blob/main/.werft/jobs/build/installer/installer.ts#L55) to use `config init` as part of this PR? Then we should be all good and not use the deprecated command anywhere that I can remember right now at least ☺️

MrSimonEmms:
@mads-hartmann ooooh, you drive a hard bargain. But, seeing as you asked so nicely 😃

mads-hartmann (approved):
Awesome, thanks Simon!

install/installer/cmd/config.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// configCmd represents the validate command
+var configCmd = &cobra.Command{
+	Use:   "config",
+	Short: "Perform configuration tasks",
+}
+
+// configFileExists checks if the configuration file is present on disk
+func configFileExists() (*bool, error) {
```
nandajavarma, line 28: Just out of curiosity, what made you return a pointer to Boolean here?

MrSimonEmms: Correct. I don't like returning a boolean if it can return an error - it's entirely stylistic, so can see why someone else might not like/use it

nandajavarma (approved):
Tested and code looks good to me! :shipit:

PR #12687, opened 2022-09-06
[installer]: build the config

### LINKED ISSUE #12199: create a `config build-from-envvars` command in the Installer that will build the `config.yaml` from envvars

MrSimonEmms:
## Description

Creates a `config build-from-envvars` command that replaces all of the bash script config generation with a Golang file. This ties the envvars to the config version. This means that the envvars are now part of our public contract.

This is part of a major refactoring of the configuration so that it's simpler and just config generation. ~See https://github.com/gitpod-io/gitpod/pull/12319 for my notepad of how it'll look eventually.~

## Related Issue(s)

Fixes #12199

## How to test

As per #12986

## Release Notes

```release-note
[installer]: refactor the Installer config to build from envvars
```

## Werft options:

- [ ] /werft with-preview


install/installer/pkg/config/v1/envvars.go
```diff
@@ -0,0 +1,418 @@
...
+	cfg.Domain = envvars.Domain
+
+	if licenseName := envvars.LicenseName; licenseName != "" {
+		log.Infof("Setting license name: %s", licenseName)
+		cfg.License = &ObjectRef{
+			Kind: ObjectRefSecret,
+			Name: licenseName,
+		}
+	} else {
+		cfg.License = nil
```
Pothulapati, line 154: So, I'm wondering why we have this else clause. Considering that we have `License` marked with the `omitempety` tag with a `*` type, This should be `nil` already right? when we are not setting.

MrSimonEmms: The reason I've done this is because it may be run multiple times. The first time the license envvar was detected, the second it was not - if we didn't have the `else`, then the value wouldn't be unset.

Granted, this is an edge case, but it is logically possible. Throughout this function, it explicitly sets things to the default values.

install/installer/pkg/config/v1/envvars.go
```diff
@@ -0,0 +1,418 @@
...
+			Name: envvars.ImagePullSecretName,
+		})
+		cfg.DropImageRepo = pointer.Bool(true)
+
+		cfg.ContainerRegistry.PrivateBaseImageAllowList = append(cfg.ContainerRegistry.PrivateBaseImageAllowList, envvars.LocalRegistryHost)
+	}
+
+	if envvars.RegistryDockerConfigEnabled {
+		log.Info("Extracting servers from the custom registry authentication")
+		var registryDockerConfig dockerConfigJson
```
Pothulapati, line 226: Was Wondering that this isn't useful, but looking at https://github.com/gitpod-io/gitpod/pull/12986/files#diff-2ab933b419a78565fb696be3db9ac5a3fc6156ac1556a717c52b865b7f087070R79, It does make sense

install/installer/pkg/config/v1/envvars.go
```diff
@@ -0,0 +1,418 @@
...
+	"github.com/gitpod-io/gitpod/common-go/log"
+	"github.com/gitpod-io/gitpod/installer/pkg/config"
+	"github.com/gitpod-io/gitpod/installer/pkg/config/v1/experimental"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/yaml"
+)
+
+// isTruthy performs a case-insensitive match for values parsed as true
+func isTruthy(input string) bool {
```
Pothulapati, line 22: 💯

install/installer/pkg/config/v1/envvars.go
```diff
@@ -0,0 +1,418 @@
...
+	LocalRegistryAddress                 string   `env:"LOCAL_REGISTRY_ADDRESS"`
+	LocalRegistryEnabled                 bool     `env:"HAS_LOCAL_REGISTRY"`
+	LocalRegistryHost                    string   `env:"LOCAL_REGISTRY_HOST"`
+	LocalRegistryImagePullConfig         string   `env:"LOCAL_REGISTRY_IMAGE_PULL_DOCKER_CONFIG_JSON"`
+	OpenVSXUrl                           string   `env:"OPEN_VSX_URL"`
+	RegistryDockerConfigEnabled          bool     `env:"REG_DOCKER_CONFIG_ENABLED"`
+	RegistryDockerConfig                 string   `env:"REG_DOCKER_CONFIG_JSON"`
+	RegistryInClusterEnabled             bool     `env:"REG_INCLUSTER_ENABLED" envDefault:"1"`
+	RegistryInClusterStorageType         string   `env:"REG_INCLUSTER_STORAGE"`
+	RegistryInClusterStorageS3BucketName string   `env:"REG_INCLUSTER_STORAGE_S3_BUCKETNAME"`
```
Pothulapati, line 67: Because the `env`'s are also a bit public facing, I think we should have the same consistency here i.e a `_` btw bucket and name.

install/installer/pkg/config/v1/envvars.go
```diff
@@ -0,0 +1,418 @@
...
+	RegistryInClusterStorageS3CertName   string   `env:"REG_INCLUSTER_STORAGE_S3_CERTIFICATE_NAME"`
```
Pothulapati, line 68: Should we expand `REG` to `REGISTRY`? Makes sense though if we are hitting some env name limits :p

install/installer/pkg/config/v1/envvars.go
```diff
@@ -0,0 +1,418 @@
...
+	StorageGCPServiceAccountName         string   `env:"STORE_GCP_SERVICE_ACCOUNT_NAME"`
+	StorageS3Bucket                      string   `env:"STORE_S3_BUCKET"`
+	StorageS3CredsName                   string   `env:"STORE_S3_CREDENTIALS_NAME"`
+	StorageS3Endpoint                    string   `env:"STORE_S3_ENDPOINT"`
+	TLSSelfSignedEnabled                 bool     `env:"TLS_SELF_SIGNED_ENABLED"`
+	TLSCertManagerEnabled                bool     `env:"CERT_MANAGER_ENABLED"`
+	TLSCustomCACertEnabled               bool     `env:"TLS_CUSTOM_CA_CRT_ENABLED"`
+	TLSCustomCACertCredsName             string   `env:"TLS_CUSTOM_CA_CRT_CREDENTIALS_NAME"`
+	UserManagementBlockEnabled           bool     `env:"USER_MANAGEMENT_BLOCK_ENABLED"`
+	UserManagementBlockPassList          []string `env:"USER_MANAGEMENT_BLOCK_PASSLIST"`
```
Pothulapati, line 88: nit: a `_` btw `PASS` and `LIST`?

install/installer/pkg/config/v1/envvars_test.go
```diff
@@ -0,0 +1,96 @@
...
+			require.NoError(t, err)
+
+			for k, v := range env.Envvars {
+				t.Setenv(k, v)
+			}
+
+			err = apiVersion.BuildFromEnvvars(cfg)
+			require.NoError(t, err)
+
+			expect, _, err := config.Load(string(expectContent), true)
```
Pothulapati, line 88: Nice way to get the same defaults for comparision! :👍

Pothulapati (changes requested):
Most changes are around questions/documenting the approach. Did a comparision of `envvars.go` with existing `kots-install.sh` and they seem to match!

install/installer/pkg/config/v1/envvars.go
MrSimonEmms: I'm not sure. Let's do it as a separate PR _AFTER_ release as this has to change the setter in #12986

MrSimonEmms:
@Pothulapati added the comment we talked about. Thanks for your help 🧡

install/installer/pkg/config/v1/envvars.go
Pothulapati: You mean BEFORE release right? as we won't have a chance otherwise? 🤔

MrSimonEmms: Yeah, sorry - after merging of this PR and before the September release next week

Pothulapati (approved):
:shipit:

Issue #12350, opened 2022-08-24
[integration tests] provide a means to run tests from PR via opt-in, non-blocking mechanism


kylos101:
## Is your feature request related to a problem? Please describe

Integration tests run nightly, or ad-hoc. 

## Describe the behaviour you'd like

As an initial skateboard, give developers an opt-in, non-blocking way to easily run the tests. Perhaps as an additional check/item in PRs, such as:

- [ ] /werft with-integration-tests

For a follow-on skateboard, consider something like

- [ ] /werft with-integration-tests workspace 

Where we could also do

- [ ] /werft with-integration-tests workspace ide (but not webapp, as an example)

## Additional context

The thought is having an easier way to run tests will help teams at Gitpod more easily adopt and rely on the tests.

The preview environment for these test environments should not be the same as the live one used for preview. In other words, it should be possible to leave `with-preview` unchecked but check `with-integration-tests`.

Failed tests should not block the PR from merging.

kylos101:
@meysholdt may we ask for your help in adding to Platform groundwork?

mads-hartmann:
@kylos101 what's the motivation for this requirement:

> The preview environment for these test environments should not be the same as the live one used for preview. In other words, it should be possible to leave with-preview unchecked but check with-integration-tests.

I ask because it complicates the solution a bit. We currently have a 1:1 mapping between branches and preview environments, so if we extend this with 1:* mapping we have to extend our GC of preview environments, for example ☺️

If the goal is to be able to run integration tests in a fresh environment, could we have an initial skateboard of having people use `with-clean-slate-deployment` when they want to ensure that they get a clean VM and Gitpod installation? E.g. you'd use `with-integration-tests with-clean-slate-deployment` (then `with-preview` is implied if you use `with-integration-tests`)

kylos101:
@mads-hartmann good question! @meysholdt and I were worried that the integration tests would be extra load on the 1 preview environment. But, in hindsight, that may be premature optimization. Amazing what a night of rest will do.

Therefore, using the existing should be fine. And, since this is opt-in, it won't be disruptive. :wave:

mads-hartmann:
@meysholdt I'm assigning myself and adding it to next week as I'd like to look as this next week - I think we can get a skateboard out quite quickly ☺️ Please reach out if you'd prefer otherwise

mads-hartmann:
[...]

PR #12693, opened 2022-09-06
Fix running integration tests as part of build

mads-hartmann:
## Description

This PR makes it possible to use `with-integration-tests` to run integration tests as part of the build job. This functionality has existed for a long time, but it was never updated to work with Harvester-based preview environments.

The old approach was to start a separate Werft job which used the `eu.gcr.io/gitpod-core-dev/build/integration-tests:{{ .Annotations.version }}` image in Werft. This meant it would run the compiled integration tests. However, this image doesn't have any of the utility scripts available we need to configure access to the the preview environment.

So I instead decided to delete the `.werft/run-integration-tests.yaml` job and instead introduce a small helper script - 
 `test/run.sh` - which uses `go` to run the tests. This is similar to the approach adopted by Workspace ([.werft/workspace-run-integration-tests.sh](https://github.com/gitpod-io/gitpod/blob/main/.werft/workspace-run-integration-tests.sh)) and IDE ([.werft/ide-integration-tests-startup.yaml](https://github.com/gitpod-io/gitpod/blob/main/.werft/ide-integration-tests-startup.yaml)). 

My intention is to use `test/run.sh` in both the Workspace and IDE integration test jobs too, so that we have a single way to run integration tests which is used in both Werft and Gitpod workspaces - but I'll look into this in a follow up PR.

---

This PR also adds a new checkbox for `with-integration-tests` to the PR template so spread awareness of the flag.

## Related Issue(s)

Part of https://github.com/gitpod-io/gitpod/issues/12350

## How to test

I ran the script manually from my Gitpod workspace

```sh
./test/run.sh
./test/run.sh ide
./test/run.sh webapp
./test/run.sh workspace
```

I ran it through Werft ([see job](https://werft.gitpod-dev.com/job/gitpod-build-mads-fix-with-integration-tests.46))

```sh
werft job run github -a with-preview="true" -a with-integration-tests="ide"
```

## Release Notes

```release-note
NONE
```

## Documentation

N/A
## Werft options:

- [ ] /werft with-preview
   Valid updates are `all`, `workspace`, `webapp`, `ide`


mads-hartmann:
@utam0k I'd like you to review this as you've been working on integration tests for workspace recently. Does the README capture how you have been working on integration tests? Are there any instructions or tips & tricks missing? ☺️
@mustard-mh I believe you're worked on integration tests for IDE. If so, I'd love your review on this PR too.
@geropl Wasn't sure who knew most about the previous setup but thought it might be you.

utam0k:
@mads-hartmann Thanks for your great PR.
I don't think we can't get slack notifications when we do a werft run this way, right?

test/run.sh
```diff
@@ -0,0 +1,87 @@
...
+esac
+
+args=()
+args+=( "-kubeconfig=${KUBECONFIG:-/home/gitpod/.kube/config}" )
+args+=( "-namespace=default" )
+args+=( "-timeout=60m" )
+args+=( "-p=1" )
+
+if [[ "${GITPOD_REPO_ROOT:-}" != "" ]]; then
+  echo "Running in Gitpod workspace. Fetching USERNAME adn USER_TOKEN" | werft log slice "test-setup"
```
geropl, line 51:
```suggestion
  echo "Running in Gitpod workspace. Fetching USERNAME and USER_TOKEN" | werft log slice "test-setup"
```

geropl (approved):
Thank you @mads-hartmann , tested and works! :rocket:

mads-hartmann:
> I don't think we can't get slack notifications when we do a werft run this way, right? I want to eventually run integration tests daily and receive notifications of them via slack.

Yeah that's right, currently it won't send Slack notifications. After this PR is merged I'll look into seeing how we can have the IDE and Workspace integration tests jobs use this script.

There are two approaches (at least)

1. Have the IDE and Workspace integration test jobs use `--with-integration-test=<...>` when trigger the build job on their integration test branches.
2. Have the IDE and Workspace integration test jobs use run.sh too

Either way, we'd have to figure out exactly how to trigger the Slack notifications. But let's take that after this PR is merged 🧡

mads-hartmann:
Rebase on main to remove conflict (an image version was bumped in run-integration-tests which this PR removes entirely)

.werft/jobs/build/job-config.ts
```diff
@@ -179,3 +181,23 @@ function parseVersion(context: any) {
     }
     return version;
 }
+
+export function parseWithIntegrationTests(werft: Werft, sliceID: string, value?: string): WithIntegrationTests {
+    switch (value) {
+        case null:
+        case undefined:
+            return "skip";
```
liam-j-bennett, line 189: Why do we default to `skip` here, but for undefined/unknown we default to `all`?

mads-hartmann: It's mainly about how we want to interpret how people use (or not use) the `with-integration-tests` job attribute, especially when used in GH comments/PR descriptions where you can use `/werft run with-integration-tests` or if you just set the attribute in your PR description (`/werft with-integration-tests` - in both cases without specifying a value -  which case the value is the empty string.

In the CLI it's less common but you could do `werft job run github -a with-integration-tests=""`.

So if it is null/undefined it means they didn't use the `with-integration-tests` attribute at all - if it's the empty string they did use it, but just didn't specify a value - I decided to interpret that as "they want to run all the tests"

liam-j-bennett (approved):
🚀

.werft/jobs/build/job-config.ts
ArthurSens: Would it be useful to add this explanation as a comment in the code, or do you think it is overkill?

mads-hartmann: I added a log line of

```
werft.log(sliceID, "with-integration-tests was not set - will use 'skip'");
```

As I think that will

- Capture the essence of the description above. Each of the three cases (not set, set to empty, set to unknown value) now have a log line that describe things a bit
- This might be useful as it will help people know if they made a typo in the attribute, e.g. "with-integration-test" (missing `s`)

iQQBot (approved):
Thanks! LGTM

PR #12695, opened 2022-09-06
[WIP] configcat

mustard-mh:
## Related Issue(s)

Fixes #

## How to test

With preview env of this PR

- Create a team, and get its team_id
- Configure [ConfigCat](https://app.configcat.com/08da1258-64fb-4a8e-8a1e-51de773884f6/08da1258-6541-4fc7-8b61-c8b47f82f3a0/08da1258-6512-4ec0-80a3-3f6aa301f853) `gitpod_experimental_portsView_enabled` key with team_ids labels appended with your team_id
- See if new PortsView works
- Process in new PortsView, see if events in segment reported

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/.werft/jobs/build/installer/installer.ts
+++ b/.werft/jobs/build/installer/installer.ts
@@ -72,9 +72,8 @@ export class Installer {
             this.configureSSHGateway(slice);
             this.configurePublicAPIServer(slice);
             this.configureUsage(slice);
-            this.configureConfigCat(slice);
-
             this.configureDefaultTemplate(slice);
+            this.configureConfigCat(slice);
 
             if (this.options.analytics) {
                 this.includeAnalytics(slice);
@@ -247,10 +246,13 @@ EOF`);
 
     private configureConfigCat(slice: string) {
         // This key is not a secret, it is a unique identifier of our ConfigCat application
+        const configcatKey = "WBLaCPtkjkqKHlHedziE9g/LEAOCNkbuUKiqUZAcVg7dw";
         exec(
-            `yq w -i ${this.options.installerConfigPath} experimental.webapp.configcatKey "WBLaCPtkjkqKHlHedziE9g/LEAOCNkbuUKiqUZAcVg7dw"`,
+            `yq w -i ${this.options.installerConfigPath} experimental.webapp.configcatKey "${configcatKey}"`,
             { slice: slice },
         );
+        exec(`yq w -i ${this.options.installerConfigPath} 'workspace.templates.default.spec.containers.(name==workspace).env[+].name' GITPOD_CONFIGCAT_SDK_KEY`);
+        exec(`yq w -i ${this.options.installerConfigPath} 'workspace.templates.default.spec.containers.(name==workspace).env.(name==GITPOD_CONFIGCAT_SDK_KEY).value' "${configcatKey}"`);
     }
 
     private includeAnalytics(slice: string): void {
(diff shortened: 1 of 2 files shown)


.werft/jobs/build/installer/installer.ts
```diff
@@ -247,10 +246,13 @@ EOF`);
...
     private configureConfigCat(slice: string) {
         // This key is not a secret, it is a unique identifier of our ConfigCat application
+        const configcatKey = "WBLaCPtkjkqKHlHedziE9g/LEAOCNkbuUKiqUZAcVg7dw";
         exec(
-            `yq w -i ${this.options.installerConfigPath} experimental.webapp.configcatKey "WBLaCPtkjkqKHlHedziE9g/LEAOCNkbuUKiqUZAcVg7dw"`,
+            `yq w -i ${this.options.installerConfigPath} experimental.webapp.configcatKey "${configcatKey}"`,
             { slice: slice },
         );
+        exec(`yq w -i ${this.options.installerConfigPath} 'workspace.templates.default.spec.containers.(name==workspace).env[+].name' GITPOD_CONFIGCAT_SDK_KEY`);
+        exec(`yq w -i ${this.options.installerConfigPath} 'workspace.templates.default.spec.containers.(name==workspace).env.(name==GITPOD_CONFIGCAT_SDK_KEY).value' "${configcatKey}"`);
```
jeanp413, line 255: what's the a reason to put the key in an env variable? I know the key is public but in general I think is not a good practice to put secrets in env variables and also pollutes the shell environment, is it just to avoid copying it for the different clients that may use it (vscode/jetbrains/supervisor)?

mustard-mh: This is how we did with segment https://github.com/gitpod-io/gitpod/blob/867b11fef5eb25bf669cfb96c35ecfa566cf6565/.werft/jobs/build/installer/installer.ts#L263-L264

jeanp413: yeah I see that, the same applies to it :slightly_smiling_face:

mustard-mh: > yeah I see that, the same applies to it 🙂

@jeanp413  I thought segment env var in workspace is not used by anyone. Maybe remove it after ask other teams if it affect something.

If we don't want configcat key to exists in env var, where should we put it in? 
- 👎 Inside build process will affect Self-Hosted images
- 👎 Inside env leak secret
- 👎 Inside API only make it a little bit harder to leak

mustard-mh: As the key is public, I thought it can be ok.

For segment's key `GITPOD_ANALYTICS_SEGMENT_KEY`, it's ok too because its value is refer to `Production Untrusted` project

mustard-mh: > 👍 Ask server to provide API about configcat (experimental) like trackEvent API

Should we change plan to add it to server API? It benefits other components like JetBrains Gateway to use it easier. 

cc @akosyakov @jeanp413

akosyakov: No we should not go through server for it, but used ConfigCat directly.

akosyakov: The good thing about env var is that it is available to all IDEs, so IDE can use it or fallback to non-production hardcoded inside. I think it is alright to continue with env var.

mustard-mh:
Close as we move configcat `config.json` into `proxy`, more detail see issue https://github.com/gitpod-io/gitpod/issues/12739, and its PR https://github.com/gitpod-io/gitpod/pull/12756

PR #12701, opened 2022-09-06
Install monitoring-satellite with obs-installer

vulkoingim:
## Related Issue(s)

Fixes https://github.com/gitpod-io/ops/issues/3585

## How to test

```bash
werft run github -j .werft/build.yaml -a with-preview=true
```

https://werft.gitpod-dev.com/job/gitpod-custom-aa-obs-installer-ss.11 (fixed the popd issue with the last commit)

```bash
NAME                                 READY   STATUS    RESTARTS   AGE
alertmanager-main-0                  2/2     Running   0          38m
grafana-8dc9c7bb9-x5525              1/1     Running   0          2m20s
http-prober-867c755dd-drgkc          1/1     Running   0          38m
kube-state-metrics-876578749-4g8t2   3/3     Running   0          38m
kubescape-bdd7f8776-lkqlw            1/1     Running   0          38m
node-exporter-xnwlh                  2/2     Running   0          38m
prometheus-k8s-0                     2/2     Running   0          38m
prometheus-operator-7f5444cc-v5x6m   2/2     Running   0          38m
pyrra-api-7944b86bfd-swrds           1/1     Running   0          38m
pyrra-kubernetes-7dd5d8dd46-xqnzd    1/1     Running   0          38m
```

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/.werft/observability/install-satellite.sh
+++ b/.werft/observability/install-satellite.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+
+set -euo pipefail
+
+SCRIPT_PATH=$(realpath "$(dirname "$0")")
+
+if [[ -z "${PREVIEW_NAME}" ]]; then
+  echo "Must set PREVIEW_NAME variable" 1>&2
+  exit 1
+fi
+
+if [[ -z "${KUBE_PATH}" ]]; then
+  echo "Must set KUBE_PATH variable" 1>&2
+  exit 1
+fi
+
+# exports all vars
+shopt -os allexport
+
+kubectl --kubeconfig "${KUBE_PATH}" create ns monitoring-satellite || true
+kubectl --kubeconfig "${KUBE_PATH}" create ns certmanager || true
+
+if ! command -v envsubst; then
+  go install github.com/a8m/envsubst/cmd/envsubst@latest
+fi
+
+obsDir="${SCRIPT_PATH}/observability"
+mkdir -p "${obsDir}"
+git clone https://roboquat:"$(cat /mnt/secrets/monitoring-satellite-preview-token/token)"@github.com/gitpod-io/observability.git "${obsDir}"
+cd "${obsDir}/installer"
+
+tmpdir=$(mktemp -d)
+
+envsubst <"${SCRIPT_PATH}/manifests/monitoring-satellite.yaml" | go run main.go render --output-split-files "${tmpdir}" --config -
+
+pushd "${tmpdir}"
+
+# we have to apply the CRDs first and wait until they are available before we can apply the rest
+find . -name "*CustomResourceDefinition*" -exec kubectl --kubeconfig "${KUBE_PATH}" apply -f {} --server-side \;
+
+# wait for the CRDs
+kubectl --kubeconfig "${KUBE_PATH}" -n monitoring-satellite wait --for condition=established --timeout=60s crd/servicemonitors.monitoring.coreos.com
+
+kubectl --kubeconfig "${KUBE_PATH}" apply --server-side -f .
+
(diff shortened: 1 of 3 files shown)


.werft/observability/install-satellite.sh
```diff
@@ -0,0 +1,47 @@
...
+
+kubectl --kubeconfig "${KUBE_PATH}" create ns monitoring-satellite || true
+
+if ! command -v envsubst; then
+  go install github.com/a8m/envsubst/cmd/envsubst@latest
+fi
+
+obsDir="${SCRIPT_PATH}/observability"
+mkdir -p "${obsDir}"
+git clone https://roboquat:"$(cat /mnt/secrets/monitoring-satellite-preview-token/token)"@github.com/gitpod-io/observability.git "${obsDir}"
```
vulkoingim, line 28: We should replace this with `go install`, if this fixes the issue https://github.com/gitpod-io/gitpod/pull/12703

ArthurSens:
Hey aleks, I couldn't really understand what problems we're solving by using shell instead of typescript here 🤔

Could you elaborate a bit?

meysholdt:
> Hey aleks, I couldn't really understand what problems we're solving by using shell instead of typescript here 🤔

I assume that's because it's easier to test: run the bash script from your workspace instead of running the full werft job. That's in the spirit of the discussion we had in the platform sync on Monday: Have reusable modules that can be executed individually rather than one looong spaghetti. With that being said, it does not need to be bash, it just needs to be decoupled from the job.

vulkoingim:
In essence the [`server-side`](https://github.com/gitpod-io/gitpod/pull/12701/files#diff-7b8ce28079ef48186aaae21fbffec02c904048a018415f20f7acf589b477c123R39) apply fixes the issue with the client responding that resources exist (as that check is done client-side).

As to why use shell, instead of typescript - there are a couple of reasons:
* As Mo said - small reusable pieces that we can execute and test individually
* Wrapping bash in typescript brings 0 benefits and makes no sense IMO
* ☝️ makes everything more difficult to read and follow - i.e. spaghetti
* There is a similar script in `ops` that spins up `monitoring-satellite` in ws clusters. If the issue I referenced above is fixed - then we can have that script only in a single place (here), and reduce the amount of things we have to maintain.

.werft/observability/install-satellite.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# we have to apply the CRDs first and wait until they are available before we can apply the rest
+find . -name "*CustomResourceDefinition*" -exec kubectl --kubeconfig "${KUBE_PATH}" apply -f {} --server-side \;
+
+# wait for the CRDs
+kubectl --kubeconfig "${KUBE_PATH}" -n monitoring-satellite wait --for condition=established --timeout=60s crd/servicemonitors.monitoring.coreos.com
+
+kubectl --kubeconfig "${KUBE_PATH}" apply --server-side -f .
+
+kubectl --kubeconfig "${KUBE_PATH}" patch deployments.apps -n monitoring-satellite grafana --type=json -p="[{'op': 'remove', 'path': '/spec/template/spec/nodeSelector'}]"
```
vulkoingim, line 46: Added hold due to the grafana issue. Will look into it later

vulkoingim: Fixed in https://github.com/gitpod-io/observability/pull/309

vulkoingim:
Here's the custom job: https://werft.gitpod-dev.com/job/gitpod-custom-aa-obs-installer-ss.16

/werft run

ArthurSens (approved):

PR #12710, opened 2022-09-06
[usage] Add listing of credit balance

easyCZ:
## Description

Implements retrieval of credit balance for all attribution IDs in the Usage table. Will be used when updating invoices against Stripe.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage/pkg/db/usage.go
+++ b/components/usage/pkg/db/usage.go
@@ -200,3 +200,33 @@ func GetUsageSummary(ctx context.Context, conn *gorm.DB, attributionId Attributi
 		CreditCentsBalanceAtEnd:   creditCentsBalanceAtStart.Int64 + creditCentsBalanceInPeriod.Int64,
 	}, nil
 }
+
+type Balance struct {
+	AttributionID AttributionID `gorm:"column:attributionId;type:varchar;size:255;" json:"attributionId"`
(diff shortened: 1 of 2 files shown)


components/usage/pkg/db/usage.go
```diff
@@ -200,3 +200,33 @@ func GetUsageSummary(ctx context.Context, conn *gorm.DB, attributionId Attributi
 		CreditCentsBalanceAtEnd:   creditCentsBalanceAtStart.Int64 + creditCentsBalanceInPeriod.Int64,
 	}, nil
 }
+
+type Balance struct {
+	AttributionID AttributionID `gorm:"column:attributionId;type:varchar;size:255;" json:"attributionId"`
+	CreditCents   CreditCents   `gorm:"column:creditCents;type:bigint;" json:"creditCents"`
```
andrew-farries, line 206: I may have missed this discussion elsewhere but what is the purpose of the `CreditsCents` type? When we wrote the other reconciler we always stored credit values as floating point values.

andrew-farries (approved):

components/usage/pkg/db/usage.go
easyCZ: Basically it stores credits * 100 rounded. This removes precision errors. Here, we use a custom type to represent that it's not credits, but credits * 100, or CreditCents

PR #12711, opened 2022-09-06
[billing] Add ReconcileInvoices RPC to BillingService protos

easyCZ:
## Description

Defines ReconcileInvoices RPC on the BillingService.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

PR #12712, opened 2022-09-06
[billing] Implement ReconcileInvoices

easyCZ:
## Description

Implements reconciliation of usage balances against Stripe.

## Related Issue(s)

* Depends on https://github.com/gitpod-io/gitpod/pull/12710
* Depends on https://github.com/gitpod-io/gitpod/pull/12711

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage/pkg/apiv1/billing.go
+++ b/components/usage/pkg/apiv1/billing.go
@@ -67,6 +67,37 @@ func (s *BillingService) UpdateInvoices(ctx context.Context, in *v1.UpdateInvoic
 	return &v1.UpdateInvoicesResponse{}, nil
 }
 
+func (s *BillingService) ReconcileInvoices(ctx context.Context, in *v1.ReconcileInvoicesRequest) (*v1.ReconcileInvoicesResponse, error) {
+	balances, err := db.ListBalance(ctx, s.conn)
+	if err != nil {
+		log.WithError(err).Errorf("Failed to reconcile invoices.")
+		return nil, status.Errorf(codes.Internal, "Failed to reconcile invoices.")
+	}
+
+	creditSummaryForTeams := map[string]stripe.CreditSummary{}
+	for _, balance := range balances {
+		entity, id := balance.AttributionID.Values()
+
+		// TODO: Support updating of user attribution IDs
+		if entity != db.AttributionEntity_Team {
+			continue
+		}
+
+		creditSummaryForTeams[id] = stripe.CreditSummary{
+			Credits:  int64(math.Ceil(balance.CreditCents.ToCredits())),
+			ReportID: "no-report",
+		}
+	}
+
+	err = s.stripeClient.UpdateUsage(ctx, creditSummaryForTeams)
+	if err != nil {
+		log.WithError(err).Errorf("Failed to udpate usage in stripe.")
+		return nil, status.Errorf(codes.Internal, "Failed to update usage in stripe")
+	}
+
+	return &v1.ReconcileInvoicesResponse{}, nil
+}
+
 func (s *BillingService) FinalizeInvoice(ctx context.Context, in *v1.FinalizeInvoiceRequest) (*v1.FinalizeInvoiceResponse, error) {
 	logger := log.WithField("invoice_id", in.GetInvoiceId())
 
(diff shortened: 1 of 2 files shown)


components/usage/pkg/stripe/stripe.go
```diff
@@ -152,18 +152,6 @@ func (c *Client) updateUsageForCustomer(ctx context.Context, customer *stripe.Cu
...
-		return nil, fmt.Errorf("failed to find upcoming invoice for customer %s: %w", customer.ID, err)
-	}
-
-	_, err = c.UpdateInvoiceMetadata(ctx, invoice.ID, map[string]string{
-		ReportIDMetadataKey: summary.ReportID,
-	})
-	if err != nil {
-		return nil, fmt.Errorf("failed to udpate invoice %s metadata with report ID: %w", invoice.ID, err)
-	}
-
```
easyCZ, line 166: Dropping this part as 
1. It already doesn't work as we can't update an UpcomingInvoice in Stripe with metadata
2. We're no longer planning to use the metadata for sharing the report

components/usage/pkg/apiv1/billing.go
```diff
@@ -67,6 +67,37 @@ func (s *BillingService) UpdateInvoices(ctx context.Context, in *v1.UpdateInvoic
...
+			continue
+		}
+
+		creditSummaryForTeams[id] = stripe.CreditSummary{
+			Credits:  int64(math.Ceil(balance.CreditCents.ToCredits())),
+			ReportID: "no-report",
+		}
+	}
+
+	err = s.stripeClient.UpdateUsage(ctx, creditSummaryForTeams)
```
geropl, line 92: If I read correctly [here](https://github.com/gitpod-io/gitpod/blob/34ef9cb2edb63a102d881ea48bb1c330e6a235ba/components/usage/pkg/stripe/stripe.go#L83-L86), we might fail to update if the very first query is for a costCenter without stripe account, no?

easyCZ: We fail if we fail to lookup customers from stripe. But we then iterate over the stripe customers which actually exist rather than over all the attribution IDs so in this case it would not fail on the first which doesn't have a stripe account.

components/usage/pkg/apiv1/billing.go
```diff
@@ -67,6 +67,37 @@ func (s *BillingService) UpdateInvoices(ctx context.Context, in *v1.UpdateInvoic
...
 }
 
+func (s *BillingService) ReconcileInvoices(ctx context.Context, in *v1.ReconcileInvoicesRequest) (*v1.ReconcileInvoicesResponse, error) {
+	balances, err := db.ListBalance(ctx, s.conn)
+	if err != nil {
+		log.WithError(err).Errorf("Failed to reconcile invoices.")
+		return nil, status.Errorf(codes.Internal, "Failed to reconcile invoices.")
+	}
+
+	creditSummaryForTeams := map[string]stripe.CreditSummary{}
```
svenefftinge, line 77: Can we generalize and talk about attributeId instead of teams?

components/usage/pkg/apiv1/billing.go
```diff
@@ -67,6 +67,37 @@ func (s *BillingService) UpdateInvoices(ctx context.Context, in *v1.UpdateInvoic
...
+	for _, balance := range balances {
+		entity, id := balance.AttributionID.Values()
+
+		// TODO: Support updating of user attribution IDs
+		if entity != db.AttributionEntity_Team {
```
svenefftinge, line 82: why do we need this exclusion here?

svenefftinge (approved):
LGTM. My concern regarding generalization to attributionIds (or accounts, or costcenters) can be discussed and tackled in a follow-up PR.

components/usage/pkg/apiv1/billing.go
easyCZ: Legacy of the focus on team based UBP first.

components/usage/pkg/apiv1/billing.go
easyCZ: Yes, this requires a couple of changes. For one we need to start setting the `attributionId: <value>` on stripe customers rather than the `teamId` we set now

PR #12727, opened 2022-09-07
[supervisor] add 0 bucket for perf analytics

akosyakov:
## Description

We need to track if users need less than 1 CPU and 1 Gb of memory.

## Related Issue(s)

Fixes #

## How to test

- Start a workspace in prev env
- Inspect gitpod_cpu_changed and gitpod_memory_changed events in segment untrusted. I would recommend to filter by instance id. Whenever perf metrics hit a higher bucket you should see a change. If it goes down though we don't report changes.

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x]  /werft analytics=segment|TEZnsG4QbLSxLfHfNieLYGF4cDwyFWoe

### DIFF
--- a/components/supervisor/pkg/supervisor/supervisor.go
+++ b/components/supervisor/pkg/supervisor/supervisor.go
@@ -1565,8 +1565,8 @@ func analysePerfChanges(ctx context.Context, wscfg *Config, w analytics.Writer, 
 		})
 	}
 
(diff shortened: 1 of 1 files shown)


akosyakov:
@mustard-mh you are right, so everything is covered already 🙏

PR #12732, opened 2022-09-07
[UBP] Extract common UBP configuration components

andrew-farries:
## Description

Extract common React components that will be used on both the team settings and user settings pages to configure usage based pricing. Specifically:

image

and:

image

These components will be re-used in a later PR when they are added to the individual user settings page.

## Related Issue(s)

Part of https://github.com/gitpod-io/gitpod/issues/12685

## How to test

* Create a new team in the preview environment whose name contains 'Gitpod'.
* Configure usage based pricing for the new team (via its settings page).

The setup should work correctly as before, with a new customer created in Stripe with the correct team id in the customer's metadata.

## Release Notes

```release-note
NONE
```

## Werft options:
- [x] /werft with-preview
- [x] /werft with-payment

Issue #12685, opened 2022-09-06
[UBP for individuals] Configure usage-based billing in personal settings


jldec:
This lives in the Billing page at https://gitpod.io/billing in your personal settings, where we already show your team billing account if active.

![Image](https://user-images.githubusercontent.com/849592/188619141-9a3b0040-6d10-4696-8ce0-bd97c06e2a8c.png)

---

A new section on the same page should show personal UBP billing as inactive when no payment method has been configured.

![Image](https://user-images.githubusercontent.com/849592/188620711-89df76ff-0a0a-4032-ab34-d69cdb076d10.png)

---

After configuring a payment mething the section  shows the following, similar to team billing.

![Image](https://user-images.githubusercontent.com/849592/188620733-11eee880-3427-4297-9191-e3a8f4f3ccdb.png)

### Server
- [x] Add new methods to `server` to support individual user signup: 
  - [x] Add `createOrUpdateStripeCustomerForUser` method (#12753)
  - [x] Add `findStripeSubscriptionId` (https://github.com/gitpod-io/gitpod/pull/12876)
  - [x] Add `getStripePortalUrl` (https://github.com/gitpod-io/gitpod/pull/12864)
  - [x] Add `(get|set)UsageLimit(attributionId)` (https://github.com/gitpod-io/gitpod/pull/12767)
  - [x] Add `subscribeUserToStripe` (https://github.com/gitpod-io/gitpod/pull/12887)
### UI
- [x] Extract components common to both team and individual user UBP configuration (#12732)
- [x] Add placeholder billing configuration UI to individual billing page (#12733)
- [x] Hook up placeholder billing UI on the user settings page with the new server methods. (https://github.com/gitpod-io/gitpod/pull/12942)
- [x] Remove team-specific Stripe APIs from `server`:
  - [x] `findStripeSubscriptionIdForTeam` (https://github.com/gitpod-io/gitpod/pull/12957)
    - [x] Need to remove one usage from `BillingAccountSelector` first (https://github.com/gitpod-io/gitpod/pull/12954)
  - [x] `subscribeTeamToStripe` (https://github.com/gitpod-io/gitpod/pull/12957)
  - [x] `getStripePortalUrlForTeam` (https://github.com/gitpod-io/gitpod/pull/12957)
  - [x] `getUsageLimitForTeam` (https://github.com/gitpod-io/gitpod/pull/12957)
[...]

Issue #12684, opened 2022-09-06
[UBP for individuals] Configure a usage limit for usage-based billing in personal settings


jldec:
The design and behavior of this UI should be very similar to the UBP usage limit input for teams.

- The limit value is pre-set to x credits, where x = the free credit allowance for individuals.
- Thie UI should allow the value to be set above (or equal to) the free credit allowance, but not below.

![Image](https://user-images.githubusercontent.com/849592/188622052-89620a39-f285-4b09-b628-c83a851a980b.png)

svenefftinge:
The usage limit is unrelated to the amount of credits we grant on for free. I.e. It should be always 0 when there is no pay-as-you-go plan active. We should think of the usage limit as a debit limit.

andrew-farries:
Removing my assignment from this now that https://github.com/gitpod-io/gitpod/pull/12942 is merged.

Users can now sign up for UBP, so this issue is unblocked.

AlexTugarev:
This needs to be verified. I think it's already solved.

AlexTugarev:
Verified in a test deployment, cf. https://github.com/gitpod-io/gitpod/pull/13076#issuecomment-1250836541
✅

PR #12733, opened 2022-09-07
[UBP] Add placeholder billing configuration UI to the user settings page

andrew-farries:
## Description

Building on https://github.com/gitpod-io/gitpod/pull/12732 which extracted the components to make them reusable, add the UBP configuration UI widgets to the individual user billing page:

image

The 'Upgrade billing' button doesn't work yet. This is fine to merge as-is because only people opted into usage based billing will see the UI (currently only Gitpod team members).

## Related Issue(s)

## How to test

Visit your user settings page and see the placeholder UI

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/dashboard/src/settings/Billing.tsx
+++ b/components/dashboard/src/settings/Billing.tsx
@@ -4,11 +4,18 @@
  * See License-AGPL.txt in the project root for license information.
  */
 
+import { useContext } from "react";
 import { PageWithSettingsSubMenu } from "./PageWithSettingsSubMenu";
 import { BillingAccountSelector } from "../components/BillingAccountSelector";
+import { UserContext } from "../user-context";
 import UsageBasedBillingConfig from "../components/UsageBasedBillingConfig";
+import { AttributionId } from "@gitpod/gitpod-protocol/lib/attribution";
 
 export default function Billing() {
+    const { user } = useContext(UserContext);
+
+    const attributionId: AttributionId = { kind: "user", userId: user?.id || "" };
+
     return (
         <PageWithSettingsSubMenu title="Billing" subtitle="Usage-Based Billing.">
             <div>
@@ -16,8 +23,8 @@ export default function Billing() {
                 <BillingAccountSelector />
                 <h3 className="mt-12">Usage-Based Billing</h3>
                 <UsageBasedBillingConfig
-                    userOrTeamId={""}
-                    showSpinner={false}
+                    attributionId={attributionId}
+                    showSpinner={!user}
                     showUpgradeBilling={true}
                     showManageBilling={false}
                     doUpdateLimit={function (_: number): Promise<void> {
--- a/components/dashboard/src/settings/Billing.tsx
+++ b/components/dashboard/src/settings/Billing.tsx
@@ -6,6 +6,7 @@
 
 import { PageWithSettingsSubMenu } from "./PageWithSettingsSubMenu";
 import { BillingAccountSelector } from "../components/BillingAccountSelector";
+import UsageBasedBillingConfig from "../components/UsageBasedBillingConfig";
 
 export default function Billing() {
     return (
@@ -13,6 +14,16 @@ export default function Billing() {
             <div>
                 <h3>Billing Account</h3>
                 <BillingAccountSelector />
+                <h3 className="mt-12">Usage-Based Billing</h3>
+                <UsageBasedBillingConfig
+                    userOrTeamId={""}
+                    showSpinner={false}
+                    showUpgradeBilling={true}
+                    showManageBilling={false}
+                    doUpdateLimit={function (_: number): Promise<void> {
+                        throw new Error("Function not implemented.");
+                    }}
+                />
             </div>
         </PageWithSettingsSubMenu>
     );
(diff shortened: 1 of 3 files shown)


components/dashboard/src/settings/Billing.tsx
```diff
@@ -6,13 +6,24 @@
...
 
 export default function Billing() {
     return (
         <PageWithSettingsSubMenu title="Billing" subtitle="Usage-Based Billing.">
             <div>
                 <h3>Billing Account</h3>
                 <BillingAccountSelector />
+                <h3 className="mt-12">Usage-Based Billing</h3>
+                <UsageBasedBillingConfig
+                    userOrTeamId={""}
```
jankeromnes, line 19: I wonder whether this shouldn't be either `userId` or `teamId` 🤔 seems a bit awkward to have an ID that could be either for a user or a team. Also, the component will eventually need to know, because it needs to call either `createOrUpdateStripeCustomerForTeam` for teams (as currently implemented) or `createOrUpdateStripeCustomerForUser` for users (still to be implemented).

andrew-farries: Yes, I agree that the API of the component is not great here. 

My intention with the parent PR (https://github.com/gitpod-io/gitpod/pull/12732 where this prop was introduced) was to do a rough extraction of the components to allow them to be used in two places without duplicating them. As we work out exactly how this component is going to call the server API from the individual settings page we should be able to make the small changes needed to clean this up.

I've added a note to https://github.com/gitpod-io/gitpod/issues/12685 to revisit this.

andrew-farries: @jankeromnes are you happy to approve this, given that we'll revisit the API of the component as part of #12685?

andrew-farries: Now that we are a bit further along with the `server` side changes needed to support individual UBP, I've changed the prop from `userOrTeamId` to an `attributionId`.

I think this addresses the concern about the awkward prop name.

components/dashboard/src/teams/TeamUsageBasedBilling.tsx
```diff
@@ -158,11 +159,13 @@ export default function TeamUsageBasedBilling() {
         }
     };
 
+    const attributionId: AttributionId = { kind: "team", teamId: team.id };
```
jankeromnes, line 162: Given the above, maybe it's okay to have a "garbage" `attributionId` until both `team` (and `subscriptionId`, but that's less relevant here) are loaded:

```suggestion
    const attributionId: AttributionId = { kind: "team", teamId: team?.id || "" };
```

jankeromnes: Alternatively, you could make `attributionId` of type `AttributionId?`, and leave it `undefined` until the `team` is loaded.

components/dashboard/src/teams/TeamUsageBasedBilling.tsx
```diff
@@ -135,7 +136,7 @@ export default function TeamUsageBasedBilling() {
         }
     }, [pendingStripeSubscription, pollStripeSubscriptionTimeout, stripeSubscriptionId, team]);
 
-    if (!BillingMode.showUsageBasedBilling(teamBillingMode)) {
+    if (!BillingMode.showUsageBasedBilling(teamBillingMode) || !team) {
         return <></>;
```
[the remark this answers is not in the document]
andrew-farries: Thanks. Fixed in favour of letting the team id be `""` until it loads.

components/dashboard/src/teams/TeamUsageBasedBilling.tsx
andrew-farries: Thanks. Applied your suggestion.

AlexTugarev:
@andrew-farries and @jankeromnes, what's the state of this PR? And more interestingly, how does this relate to https://github.com/gitpod-io/gitpod/issues/12684? Is it already covered in here? If yes, please assign accordingly. If not, are you going to continue with that?

andrew-farries:
> @andrew-farries and @jankeromnes, what's the state of this PR? And more interestingly, how does this relate to #12684? Is it already covered in here? If yes, please assign accordingly. If not, are you going to continue with that?

I'm still hoping to land this soon. All review comments have been addressed.

The usage limit for individuals (#12684) I consider to be part of https://github.com/gitpod-io/gitpod/issues/12685 as the UI for configuring billing for individuals also includes the UI to set usage limits. I've assigned #12684 to myself.

components/dashboard/src/settings/Billing.tsx
```diff
@@ -4,15 +4,37 @@
...
 import { BillingAccountSelector } from "../components/BillingAccountSelector";
+import { UserContext } from "../user-context";
+import UsageBasedBillingConfig from "../components/UsageBasedBillingConfig";
+import { AttributionId } from "@gitpod/gitpod-protocol/lib/attribution";
 
 export default function Billing() {
+    const { user } = useContext(UserContext);
+
+    if (!user) {
+        return <></>;
```
jankeromnes, line 18: Same here: Returning `<></>` early when something is still loading prevents showing a spinner (or anything at all), and thus makes the UI look broken temporarily.

How about, instead, using `user?.id || ""` everywhere a string ID is needed, and setting `showSpinner={!user}` below?

andrew-farries: Thanks; done as you suggest.

jankeromnes (approved):
Many thanks for iterating on this! 👍 Now looks perfect to me 🎯 and also works as advertised.

(It's even possible to create a customer as a User, however since the User Billing page doesn't yet handle the Stripe redirection and doesn't call `subscribeUserToStripe`, we're not attaching credit cards or creating subscriptions for users yet -- unrelated to this PR / good as follow-up)

PR #12734, opened 2022-09-07
[server] Ensure PVC workspace class does not change on restart

### LINKED ISSUE #12494: [PVC] Can't open from prebuilds if prebuilds with large workspace class but user prefers standard workspace class
### Bug description

Prebuilds was built with a large workspace class (50Gi), and the user prefers a standard workspace class (30Gi).
The user cannot open the workspace because the PVC disk size is 30Gi, not 50Gi.

### Steps to reproduce

- Force the prebuilds to build with a large workspace class (50Gi)
- After prebuilds are finished, the VolumeSnapshot size is 50Gi
- Change the user workspace class preference is standard workspace class (30Gi)
- Open the workspace from prebuilds, the workspace can't start because the disk size dos not be increased

### Workspace affected

_No response_

### Expected behavior

@atduarte @Furisto @jenting my understanding is, that in this scenario, we want to auto upgrade the workspace class to large, so that the regular workspace being started from a large prebuild is also large. Otherwise, it will not be able to start, because the restore will fail as @jenting described.

### Example repository

_No response_

### Anything else?

https://github.com/gitpod-io/gitpod/issues/7901

### LINKED ISSUE #12666: [PVC] Open a fresh new workspace, it uses large node pool. Relaunch the workspace, it uses standard node pool
### Bug description

The user workspace class preference is standard.
The prebuild was built with large workspace class.

Open a fresh new workspace from the prebuild project, and then the workspace uses the large node pool because the prebuild was built on a large node pool. Relaunch the workspace, and then it uses a standard node pool. So, the workspace pod can't be up.

This is because the VolumeSnapshot size is 50Gi, and the PVC size is 30Gi.

### Steps to reproduce

Prepare the prebuild project
- Set the user's workspace class preference as large
- Configure a prebuild project and run the prebuild

Open workspace from the prebuild project
- Set the user's workspace class preference as standard
- Enable the user's PVC feature flag
- Do not enable the project to prebuild PVC feature flag
- Open the new workspace from the prebuild project 
  image
- Make sure the workspace is using the large node pool `g1-large-pvc` by checking the label `kubectl get pod <pod-name> --show-labels | grep gitpod.io/workspaceClass`
- Check the corresponding PVC with size 50Gi
- Stop the workspace
- Check the VolumeSnapshot size is 50Gi
- Reopen the workspace. The workspace can't be open 
  - The workspace pod is using the standard node pool `g1-standard-pvc` by checking the label `kubectl get pod <pod-name> --show-labels | grep gitpod.io/workspaceClass`
  - The workspace pvc size is 30Gi
  - The workspace pvc restored from the VolumeSnapshot, the VolumeSnapshot size is 50Gi
    image

### Workspace affected

_No response_

### Expected behavior

The workspace should keep use the large node pool. So the workspace can be open.

### Example repository

_No response_

### Anything else?

#7901

Similar to #12494

Furisto:
## Description
- Ensure PVC workspace class does not change on restart
- Use pvc class if prebuild is pvc if user would get legacy pvc class

## Related Issue(s)
Fixes #12666
Fixes #12494

## How to test
- Select large workspace class in your user preferences
- Enable persistent_volume_claim feature for your user in the preview environment via administration
- Start prebuild -> prebuild should have g1-large-pvc
- Create regular workspace from prebuild (should have g1-large-pvc)
- Stop workspace
- Restart workspace (should have g1-large-pvc)

## Release Notes

```release-note
None
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/server/src/workspace/workspace-starter.ts
+++ b/components/server/src/workspace/workspace-starter.ts
@@ -903,9 +903,20 @@ export class WorkspaceStarter {
 
                 featureFlags = featureFlags.concat(["workspace_class_limiting"]);
             } else {
-                workspaceClass = "default";
-                if (await this.entitlementService.userGetsMoreResources(user)) {
-                    workspaceClass = "gitpodio-internal-xl";
+                // todo: remove this once pvc has been rolled out
+                const prebuildClass = await WorkspaceClasses.getFromPrebuild(
+                    ctx,
+                    workspace,
+                    this.workspaceDb.trace(ctx),
+                );
+                if (prebuildClass?.endsWith("-pvc")) {
+                    workspaceClass = prebuildClass;
+                    // ####
+                } else {
+                    workspaceClass = "default";
+                    if (await this.entitlementService.userGetsMoreResources(user)) {
+                        workspaceClass = "gitpodio-internal-xl";
+                    }
                 }
             }
 
--- a/components/server/src/workspace/workspace-classes.ts
+++ b/components/server/src/workspace/workspace-classes.ts
@@ -85,6 +85,11 @@ export namespace WorkspaceClasses {
             return getDefaultId(workspaceClasses);
         }
 
+        // todo: remove this once pvc has been rolled out
+        if (previousWorkspaceClass.endsWith("-pvc")) {
+            return previousWorkspaceClass;
+        }
+
         const config = workspaceClasses.find((c) => c.id === previousWorkspaceClass);
         if (!config) {
             log.error(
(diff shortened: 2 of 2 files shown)


sagor999 (approved):
LGTM :rocket: 

/hold
hold for others to review

jenting (approved):
Work like a charm, thank you.

components/server/src/workspace/workspace-starter.ts
```diff
@@ -903,9 +903,20 @@ export class WorkspaceStarter {
...
-                    workspaceClass = "gitpodio-internal-xl";
+                // todo: remove this once pvc has been rolled out
+                const prebuildClass = await WorkspaceClasses.getFromPrebuild(
+                    ctx,
+                    workspace,
+                    this.workspaceDb.trace(ctx),
+                );
+                if (prebuildClass?.endsWith("-pvc")) {
+                    workspaceClass = prebuildClass;
+                    // ####
```
easyCZ, line 914:
```suggestion
```

easyCZ (approved):
Doesn't feel particularly elegant but I lack context to suggest an alternative.

Looks safe enough.

PR #12736, opened 2022-09-07
[usage] store more data for in usage entry

svenefftinge:
## Description
Fetch additional data for workspace instances and store it in d_b_usage.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage/pkg/db/dbtest/user.go
+++ b/components/usage/pkg/db/dbtest/user.go
@@ -1,0 +1,77 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package dbtest
+
+import (
+	"testing"
+	"time"
+
+	"github.com/gitpod-io/gitpod/usage/pkg/db"
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+	"gorm.io/gorm"
+)
+
+type User struct {
+	ID           uuid.UUID      `gorm:"primary_key;column:id;type:char;size:36;"`
+	AvatarURL    string         `gorm:"column:avatarUrl;type:char;size:255;"`
+	Name         string         `gorm:"column:name;type:char;size:255;"`
+	FullName     string         `gorm:"column:fullName;type:char;size:255;"`
+	CreationDate db.VarcharTime `gorm:"column:creationDate;type:varchar;size:255;"`
+
+	// user has more field but we don't care here as they are just used in tests.
+}
+
+func (user *User) TableName() string {
+	return "d_b_user"
+}
+
+func NewUser(t *testing.T, user User) User {
+	t.Helper()
+
+	result := User{
+		ID:           uuid.New(),
+		AvatarURL:    "https://avatars.githubusercontent.com/u/9071",
+		Name:         "HomerJSimpson",
+		FullName:     "Homer Simpson",
+		CreationDate: db.NewVarcharTime(time.Now()),
+	}
+
+	if user.ID != uuid.Nil {
+		result.ID = user.ID
+	}
+
+	if user.AvatarURL != "" {
+		result.AvatarURL = user.AvatarURL
+	}
+	if user.Name != "" {
+		result.Name = user.Name
+	}
+	if user.FullName != "" {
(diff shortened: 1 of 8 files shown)


components/usage/pkg/db/dbtest/user.go
```diff
@@ -0,0 +1,77 @@
...
+	CreationDate db.VarcharTime `gorm:"column:creationDate;type:varchar;size:255;"`
+
+	// user has more field but we don't care here as they are just used in tests.
+}
+
+func (user *User) TableName() string {
+	return "d_b_user"
+}
+
+func NewUser(t *testing.T, user User) User {
```
easyCZ, line 31: It's bit of an anti-pattern to have a `testing.T` used in a non test package. Normally we'd pull this into a separate package - it also has the benefit of not being included in the resulting binary if it's in a different package (not referenced by "prod" code)

components/usage/pkg/db/dbtest/workspace_instance.go
```diff
@@ -23,7 +24,7 @@ func NewWorkspaceInstance(t *testing.T, instance db.WorkspaceInstance) db.Worksp
 	t.Helper()
 
 	id := uuid.New()
-	if instance.ID.ID() != 0 { // empty value
+	if instance.ID != uuid.Nil {
```
easyCZ, line 27: Cool! I didn't know about this.

components/usage/pkg/db/dbtest/user.go
```diff
@@ -0,0 +1,77 @@
...
+
+type User struct {
+	ID           uuid.UUID      `gorm:"primary_key;column:id;type:char;size:36;"`
+	AvatarURL    string         `gorm:"column:avatarUrl;type:char;size:255;"`
+	Name         string         `gorm:"column:name;type:char;size:255;"`
+	FullName     string         `gorm:"column:fullName;type:char;size:255;"`
+	CreationDate db.VarcharTime `gorm:"column:creationDate;type:varchar;size:255;"`
+
+	// user has more field but we don't care here as they are just used in tests.
+}
```
easyCZ, line 25: This should be in the `db` package, not the `dbtest` package

components/usage/pkg/db/workspace_instance_test.go
```diff
@@ -277,13 +288,7 @@ func TestListWorkspaceInstancesInRange(t *testing.T) {
...
 	dbtest.CreateWorkspaceInstances(t, conn, all...)
 
-	retrieved, err := db.ListWorkspaceInstancesInRange(context.Background(), conn, startOfMay, startOfJune)
-	require.NoError(t, err)
-
-	ids := []uuid.UUID{}
-	for _, ws := range retrieved {
-		ids = append(ids, ws.ID)
-	}
+	retrieved := dbtest.ListWorkspaceInstancesInRange(t, conn, startOfMay, startOfJune, workspace.ID)
```
easyCZ, line 291: The listing methods should not be in the `dbtest` package, should be in the `db` package instead.

components/usage/pkg/db/dbtest/user.go
svenefftinge: it's in a testing package

easyCZ: Sorry, I got confused. The User model definition should be moved out of the `dbtest` package into the `db` package

components/usage/pkg/db/dbtest/user.go
svenefftinge: I put it here, because we don't need this in production code, it's only used for setting up testing state in DB.

components/usage/pkg/db/workspace_instance_test.go
svenefftinge: It is a testing utility that requires to provide a workspaceId and filters the result on it to make tests more stable. Should I name it differently to make it clearer?

components/usage/pkg/db/dbtest/user.go
easyCZ: I see. I guess I can live with this, but I'd generally define it on the `db` package as someone else will come along and not find the User object there so they would go and define it again

easyCZ (approved):
Let's land it like this to unblock, we can move it around later if needed.

PR #12753, opened 2022-09-08
[UBP] Add server method to create a stripe customers for an individual user

andrew-farries:
## Description

Add a `createOrUpdateStripeCustomerForUser` method to `server` to allow the creation of Stripe customers for individual users. Previously we had only the `createOrUpdateStripeCustomerForTeam` method which would create a Stripe customer for a Gitpod team.

The only difference between the `createOrUpdateStripeCustomerForUser` and `createOrUpdateStripeCustomerForTeam` methods is the metadata in the resulting Stripe customer object. For teams we store `[teamId]: foo` and for users we store `[userId]: bar`.

## Related Issue(s)

## How to test
* Set the billing mode to UBP for your user in the preview environment by creating a team with `Gitpod` in the name.
* In the developer console run:
```js
await window._gp.gitpodService.server.createOrUpdateStripeCustomerForUser("EUR")
```
* Visit the Stripe dashboard and see the new customer (without payment method or subscription) with the userId in the metadata.

## Release Notes

```release-note
NONE
```

## Werft options:
- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/server/ee/src/workspace/gitpod-server-impl.ts
+++ b/components/server/ee/src/workspace/gitpod-server-impl.ts
@@ -2086,6 +2086,24 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
         }
     }
 
+    async createOrUpdateStripeCustomerForUser(ctx: TraceContext, currency: string): Promise<void> {
+        const user = this.checkAndBlockUser("createOrUpdateStripeCustomerForUser");
+        await this.ensureStripeApiIsAllowed({ user });
+        try {
+            let customer = await this.stripeService.findCustomerByUserId(user.id);
+            if (!customer) {
+                customer = await this.stripeService.createCustomerForUser(user);
+            }
+            await this.stripeService.setPreferredCurrencyForCustomer(customer, currency);
+        } catch (error) {
+            log.error(`Failed to update Stripe customer profile for user '${user.id}'`, error);
+            throw new ResponseError(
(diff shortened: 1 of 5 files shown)


components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2086,6 +2086,24 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
         }
     }
 
+    async createOrUpdateStripeCustomerForUser(ctx: TraceContext, currency: string): Promise<void> {
+        const user = this.checkAndBlockUser("createOrUpdateStripeCustomerForUser");
+        await this.ensureStripeApiIsAllowed({ user });
```
geropl, line 2091: :100:

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2086,6 +2086,24 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
         }
     }
 
+    async createOrUpdateStripeCustomerForUser(ctx: TraceContext, currency: string): Promise<void> {
```
geropl, line 2089: :cloud: Originally I thought about merging these two methods on `attributionId`. But I think it makes sense to have them separate as they have different semantics around access rights etc.

PR #12754, opened 2022-09-08
[server] Don't count "Professional Open Source" as personal subscription

### LINKED ISSUE #12672: BillingMode: ignore personal OSS subscription
Currently user on a UBP plan are not switching to BillingMode `usage-based` in case they own a OSS subscription.
We should filter that out, as UBP is strictly more powerful, and should never block anybody from using UBP.

geropl:
## Description
Don't count "Professional Open Source" as personal subscription when calculating billing mode.

## Related Issue(s)

Fixes #12672

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/server/ee/src/billing/billing-mode.ts
+++ b/components/server/ee/src/billing/billing-mode.ts
@@ -108,7 +108,9 @@ export class BillingModesImpl implements BillingModes {
         }
         const cbSubscriptions = await this.subscriptionSvc.getActivePaidSubscription(user.id, now);
         const cbTeamSubscriptions = cbSubscriptions.filter((s) => isTeamSubscription(s));
-        const cbPersonalSubscriptions = cbSubscriptions.filter((s) => !isTeamSubscription(s));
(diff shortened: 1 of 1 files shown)


components/server/ee/src/billing/billing-mode.ts
```diff
@@ -108,7 +108,9 @@ export class BillingModesImpl implements BillingModes {
         }
         const cbSubscriptions = await this.subscriptionSvc.getActivePaidSubscription(user.id, now);
```
jankeromnes, line 109: Side note: `getActivePaidSubscription` is wrongly named (because it also returns unpaid / free subscriptions 🙄). Thankfully, this is all going away soon. 👋

jankeromnes (approved):
Code looks good to me, thanks!

(Wonder if semantically, it mightn't have been better to filter out free subscriptions from `cbSubscriptions`, but functionally it's all the same so no blocker here. 👍)

Issue #12739, opened 2022-09-07
Proxy ConfigCat config in caddy


geropl:
Currently, we're using ConfigCat client libraries from a wide variety of environments. All those clients periodically go to https://configcat.com/ and fetch the config from there directly.

Now Team IDE plans to/already uses ConfigCat to rollout features of with different lifecycles then the rest of Gitpod (delivered via marketplace to clients with different upgrade strategies vs. single version of Gitpod delivered to SaaS/Self-Hosted).

We see multiple problems with this:
 - number of requests: We're on a certain plan with ConfigCat, and are close to/already exceeding their "number of requests" limits. So far this was not an issue, because we had close control over all clients, and could tweak the polling behavior. This changes when we use ConfigCat from clients with longer lifecycles that are not under our control.
 - scaling to more clients: so far, we taught all clients to tell Self-Hosted and SaaS apart. It would be nice to centralize this behavior
 - access to ConfigCat API: so far we had to build ConfigCat keys into multiple docker images (depending on client and runtime).

The proposed solution:
 - installation-specific ConfigCat proixy: create a path in the caddy-powered `proxy` component that delivers the `config.json` file
 - if there is not ConfigCat API key configured (Self-Hosted), this returns an empty JSON object (for now)
 - if there is a ConfigCat API key (SaaS): go to ConfigCat at a certain interval, fetch the `config.json` from there and cache it locally to respond to all requests
 - point all clients to this installation specific path

@mustard-mh @easyCZ

geropl:
BTW here's an example of a Caddy plugin in our codebase: https://github.com/gitpod-io/gitpod/blob/242cc634adf0c8ec2594259554965694fad4a08a/components/proxy/plugins/headlesslogdownload/headless_log_download.go#L42

easyCZ:
The URL for the config is `https://cdn-global.configcat.com/configuration-files/<PUBLIC_API_KEY>/config_v5.json`

I believe the client library understands `etag` and 302 responses which should be used in case the config has not changed since the last etag (etag is used as a hash of the contents more-less)

PR #12756, opened 2022-09-08
Add configcat plugin to proxy

mustard-mh:
## Description

Bring ConfigCat `config.json` file to `proxy`

### For other projects that want to use configcat

Thanks for recommending @easyCZ provided: Point to a domain which is closest to your "type of install" [[1](https://gitpod.slack.com/archives/C01KGM9BH54/p1662708345333729?thread_ts=1662564786.505599&cid=C01KGM9BH54)]

- For self-hosted, point to the domain in self-hosted
- For prod, point to `gitpod.io`
- For non-prod instances, without a preview env, point to `gitpod-staging`
- For non-prod instances with a preview env, point to preview env domain

configcat `baseUrl` should be `<host>/configcat`

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/12739

### CURL

```
# fetch config should response with correct json
curl -v "https://mustard-mh2ea42534cd.preview.gitpod-dev.com/configcat/configuration-files/gitpod/config_v5.json"

# with etag should response with 302
curl -v "https://mustard-mh2ea42534cd.preview.gitpod-dev.com/configcat/configuration-files/gitpod/config_v5.json" \
     -H 'If-None-Match: W/"63161d7f-771"'
```

### SDK
- Open Gitpod with https://github.com/mustard-mh/test/tree/hw/configcat
- Open URL with port 8080
- See if the second config fetch size reduced (etag test)
- Click the button, to see if result calculate correct

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


components/proxy/plugins/configcat/configcat.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	hash := c.updateCache(configVersion, t)
+	return t, hash
+}
+
+var sg = &singleflight.Group{}
+
+// fetchConfigCatConfig with different config version. i.e. config_v5.json
+func (c *ConfigCat) fetchConfigCatConfig(configVersion string) ([]byte, error) {
+	b, err, _ := sg.Do(fmt.Sprintf("fetch_%s", configVersion), func() (interface{}, error) {
```
easyCZ, line 165: Nice use of singleflight!

install/installer/pkg/config/v1/experimental/experimental.go
```diff
@@ -172,6 +172,9 @@ type WebAppConfig struct {
 	DisableMigration       bool                   `json:"disableMigration"`
 	Usage                  *UsageConfig           `json:"usage,omitempty"`
 	ConfigcatKey           string                 `json:"configcatKey"`
+	ConfigcatBaseUrl       string                 `json:"configcatBaseUrl"`
+	ConfigcatPollInterval  string                 `json:"configcatPollInterval"`
+	ConfigcatDefaultConfig string                 `json:"configcatDefaultConfig"`
```
easyCZ, line 177: I think we could make this a new `ConfigCat` struct with the individual properties. It would make it slightly easier to reason about the relationship of the properties.

install/installer/pkg/components/proxy/deployment.go
```diff
@@ -119,6 +119,30 @@ func deployment(ctx *common.RenderContext) ([]runtime.Object, error) {
...
+			}, corev1.EnvVar{
+				Name:  "CONFIGCAT_BASE_URL",
+				Value: cfg.WebApp.ConfigcatBaseUrl,
+			}, corev1.EnvVar{
+				Name:  "CONFIGCAT_POLL_INTERVAL",
+				Value: cfg.WebApp.ConfigcatPollInterval,
+			})
+		}
+		return nil
+	})
```
easyCZ, line 144: There's a function in `common.ConfigcatEnv()` which already sets the SDK key. Perhaps we can extend it with the extra properties and then re-use in other components which instantiate ConfigCat

components/proxy/plugins/configcat/configcat.go
easyCZ: Shame https://github.com/golang/go/issues/53427 has not yet landed, would make retrieval of the result nicer.

easyCZ (approved):
Looks awesome, nice work! Left a couple of comments but these are non-blocking (adding hold if you do want to go for them). Feel free to land when you're ready.

/hold

install/installer/pkg/components/proxy/deployment.go
mustard-mh: Updated to add it to `common`, see diff [here](https://github.com/gitpod-io/gitpod/compare/1eeb694d1875331e35342d688346075f9c9cfa78..f3dae22458e64ccbdfda6e34283788b30cf0d300)

.werft/jobs/build/installer/installer.ts
```diff
@@ -248,10 +248,9 @@ EOF`);
 
     private configureConfigCat(slice: string) {
         // This key is not a secret, it is a unique identifier of our ConfigCat application
-        exec(
-            `yq w -i ${this.options.installerConfigPath} experimental.webapp.configcatKey "WBLaCPtkjkqKHlHedziE9g/LEAOCNkbuUKiqUZAcVg7dw"`,
-            { slice: slice },
-        );
+        exec(`yq w -i ${this.options.installerConfigPath} experimental.webapp.configcatKey "WBLaCPtkjkqKHlHedziE9g/LEAOCNkbuUKiqUZAcVg7dw"`, { slice: slice });
+        exec(`yq w -i ${this.options.installerConfigPath} experimental.webapp.proxy.configcat.baseUrl "https://cdn-global.configcat.com"`,{ slice: slice });
+        exec(`yq w -i ${this.options.installerConfigPath} experimental.webapp.proxy.configcat.pollInterval "10m"`,{ slice: slice });
```
easyCZ, line 253: In staging & prod, I wouldn't worry about going to 1m refresh rate.

install/installer/pkg/components/proxy/deployment.go
easyCZ: LGTM!

mustard-mh:
Tested with empty response (for self-hosted) with SDK, configcat print error message and fallback to default `false`

image

mustard-mh:
Follow up ops PR https://github.com/gitpod-io/ops/pull/4464

components/proxy/plugins/configcat/configcat.go
```diff
@@ -0,0 +1,212 @@
...
+	"golang.org/x/sync/singleflight"
+)
+
+const (
+	configCatModule = "gitpod.configcat"
+)
+
+var (
+	DefaultConfig = []byte("{}")
+	pathRegex     = regexp.MustCompile(`^/configcat/configuration-files/gitpod/config_v\d+\.json$`)
```
jeanp413, line 34: In gitpod-desktop we use both keys production and non-production for testing, having `gitpod` harcoded only maps to one key at a time, maybe better to send the configcat key dinamically?

components/proxy/plugins/configcat/configcat.go
```diff
@@ -0,0 +1,212 @@
...
+	return t, hash
+}
+
+var sg = &singleflight.Group{}
+
+// fetchConfigCatConfig with different config version. i.e. config_v5.json
+func (c *ConfigCat) fetchConfigCatConfig(configVersion string) ([]byte, error) {
+	b, err, _ := sg.Do(fmt.Sprintf("fetch_%s", configVersion), func() (interface{}, error) {
+		url := fmt.Sprintf("%s/configuration-files/%s/%s", c.baseUrl, c.sdkKey, configVersion)
+		resp, err := c.httpClient.Get(url)
```
jeanp413, line 167: Send the "If-None-Match" header to avoid fetching config.json if it hasn't change, check https://github.com/configcat/go-sdk/blob/master/v7/config_fetcher.go#L300-L344

mustard-mh: `If-None-Match` located at request [header](https://github.com/configcat/go-sdk/blob/master/v7/config_fetcher.go#L311) for server to check with etag (from previous config fetch), client will check if it's 304, if so, [use previous one](https://github.com/configcat/go-sdk/blob/eaa908ee61da13a1786a9dcda18de41c53d6be66/v7/config_fetcher.go#L320-L323)

mustard-mh: I see () 😅

components/proxy/plugins/configcat/configcat.go
mustard-mh: I've thought about this before, we could 
- Create a new preview env to do this test. But it does make the steps more complicated
- Change `baseUrl` when creating configcat client and sdkKey if we want to connect to non-product env?

jeanp413: yeah neither of those are nice :disappointed: 
- option 1 is definitely a no-no for me
- option 2, so we would use the configcat cdn as we do right now, correct?

cc @akosyakov thoughts?

mustard-mh: - Use product env in `vscode-desktop` is not harmful because configcat is calculating locally.
- If we bring non-product env to product gitpod, it will fetch non-product data every minute, but it's not need to do so
- Option 2 (change code if needed) can be a solution, i.e. use staging Gitpod as `baseUrl`

components/proxy/plugins/configcat/configcat.go
mustard-mh: I pushed a commit to request with etag @jeanp413

mustard-mh:
Concurrent r/w cache map will make some data update delay for around 1 min (pollDuration), but it's not harmful.

We can change it to `sync.Map` if you want

components/proxy/plugins/configcat/configcat.go
mustard-mh: @jeanp413  see https://github.com/gitpod-io/gitpod/pull/12756#issuecomment-1241671207

nandajavarma (approved):

mustard-mh:
Rebased to squash commits. Going to unhold it

PR #12759, opened 2022-09-08
[usage] Instantiate now after we've collected all usage data

easyCZ:
## Description

In the original case, we'd compute time as `t`, but we'd then go fetch instances and possibly find ones which which only started running after `t`. This would result in `now - t.Started()` to be negative, and hence negative runtime

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage/pkg/apiv1/usage.go
+++ b/components/usage/pkg/apiv1/usage.go
@@ -190,8 +190,6 @@ func (s *UsageService) ReconcileUsageWithLedger(ctx context.Context, req *v1.Rec
 		return nil, status.Errorf(codes.InvalidArgument, "To must not be before From")
 	}
 
-	now := s.nowFunc()
-
 	var instances []db.WorkspaceInstanceForUsage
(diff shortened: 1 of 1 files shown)


components/usage/pkg/apiv1/usage.go
```diff
@@ -224,6 +222,8 @@ func (s *UsageService) ReconcileUsageWithLedger(ctx context.Context, req *v1.Rec
 	logger.Infof("Found %d workspaces instances for usage records in draft.", len(instancesWithUsageInDraft))
 	instances = append(instances, instancesWithUsageInDraft...)
 
+	// now has to be computed after we've collected all data, to ensure that it's always greater than any of the records we fetch
+	now := s.nowFunc()
```
geropl, line 226: I think this might be an improvement and will fix ti.

:cloud: Note that this alone is not a guarantee that startTimes are not greated than now in a distributed setting. But maybe reasonably safe to assume, yes.

geropl (approved):
:+1:

PR #12755, opened 2022-09-08
[stripe] Use attributionId in stripe queries in usage

easyCZ:
## Description

Customers in staging & prod have been udpated to additively contain `attributionId` metadata.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

PR #12762, opened 2022-09-08
[stripe] Use AttributionID metadata when querying for customers

easyCZ:
## Description

Customers in staging & prod have been updated with this metadata.

Counterpart PR for usage component is here https://github.com/gitpod-io/gitpod/pull/12755

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/server/ee/src/user/stripe-service.ts
+++ b/components/server/ee/src/user/stripe-service.ts
@@ -13,6 +13,8 @@ import { AttributionId } from "@gitpod/gitpod-protocol/lib/attribution";
 const POLL_CREATED_CUSTOMER_INTERVAL_MS = 1000;
 const POLL_CREATED_CUSTOMER_MAX_ATTEMPTS = 30;
 
+const ATTRIBUTION_ID_METADATA_KEY = "attributionId";
+
 @injectable()
 export class StripeService {
     @inject(Config) protected readonly config: Config;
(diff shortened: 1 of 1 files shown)


andrew-farries (approved):

components/server/ee/src/user/stripe-service.ts
```diff
@@ -58,9 +64,7 @@ export class StripeService {
             email: User.getPrimaryEmail(user),
             name: User.getName(user),
             metadata: {
-                // userId is deprecated, use attributionId where possible
-                userId: user.id,
-                attributionId: AttributionId.render({ kind: "user", userId: user.id }),
+                ATTRIBUTION_ID_METADATA_KEY: AttributionId.render({ kind: "user", userId: user.id }),
```
jankeromnes, line 67: I don't think this is going to work (you're creating an object key that's literally `"ATTRIBUTION_ID_METADATA_KEY"`).

If you really want to make `ATTRIBUTION_ID_METADATA_KEY` configurable with a global const in this file, you'd want to first create a `const customerSpec = { email, name, metadata: {} }` and then set `customerSpec[ATTRIBUTION_ID_METADATA_KEY] = ...`.

jankeromnes:
@easyCZ FYI, this breaks the creation of new customers (please see comment above). This could have been caught immediately with a deployed preview + minimal "How to test" instructions.

components/server/ee/src/user/stripe-service.ts
easyCZ: Revert is [here](https://github.com/gitpod-io/gitpod/pull/12765). Thanks for spotting this. I applied Go style thinking it would use it as a key, not as a literal..

PR #12764, opened 2022-09-08
[usage] make tests robust

svenefftinge:
Make tess robust against parallel DB changes

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage/pkg/db/usage_test.go
+++ b/components/usage/pkg/db/usage_test.go
@@ -144,8 +144,19 @@ func TestInsertUsageRecords(t *testing.T) {
 
 	drafts, err := db.FindAllDraftUsage(context.Background(), conn)
 	require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


components/usage/pkg/db/usage_test.go
```diff
@@ -167,9 +178,10 @@ func TestUpdateUsageRecords(t *testing.T) {
 	require.NoError(t, db.UpdateUsage(context.Background(), conn, usage))
 
 	drafts, err := db.FindAllDraftUsage(context.Background(), conn)
+	cleaned := filter(drafts, attributionID)
```
andrew-farries, line 181: We should check `require.NoError(t, err)` before we do this.

andrew-farries (changes requested):

andrew-farries (approved):

Issue #12741, opened 2022-09-07
[UBP] Model free plan and other subscriptions


svenefftinge:
The current free plan is 50h per month. We need to make sure that new users get a monthly credit entry for that.

### Design Proposal

We will enhance the `CostCenter` type to contain information about a `billingStrategy`. There will initially be two strategies: `stripe`, `other`. 

All cost centers with a `stripe` strategy are managed by stripe. So we will treat them as we do now (i.e. update the credits in every usage reconciliation and wait for Stripe to come up with an invoice finalization.

The `other` cost centers will be managed by a control loop, that inserts a credit into usage that is `min(balance, usageLimit)`. This happens monthly.

svenefftinge:
The expiration algorithm could work as follows:
`Max(0, <currentBalance - SumOfNonExpiredCredits>) = expiration debit`

Example (Imagine we are at 2023-03-01):

| Credits  | ExpirationDate | Description | Date |
| ------------- | ------------- | ---------- | ---- |
| 50  |  2023-02-01  | Free credits January 2023 | 2023-01-01 |
| -10  |     | Coding Session | 2023-01-03 |
| 100  |  2023-08-01  | Special Credits | 2023-01-13 |
| -10  |    |  Coding Session | 2023-01-18 |
| -20  |    |  Expired Credits | 2023-02-01 |
| 50  |  2023-03-01  |  Free credits February 2023 | 2023-02-01 |
| -45  |    |  Coding Session | 2023-02-22 |

The currentBalance is `115`. The sum of all non expired credits is `100`. So we will put in the following two entries

| Credits  | ExpirationDate | Description | Date |
| ------------- | ------------- | ---------- | --- |
| 50  |  2023-04-01  | Free credits March 2023 | 2023-03-01 |
| -15  |     | Expired Credits | 2023-03-03 |


The control loop could be driven by the subscription table's `renewalTime` which the loop updates on every renewal. We need to ensure that expiration of debits and issuing of new credits happens in one pass.

svenefftinge:
The subscription data that drives the loop could look like this:

```
type Subscription {
   ID
   AttributionID 
   Description 
   CreationTime 
   Credits
   NextRenewalTime // will be updated on renewal
   RenewalInterval // not sure about the type, as I believe we should support something like monthly, where the renewal happens on the same day of the month.
}
```

svenefftinge:
[...]

PR #12766, opened 2022-09-08
[usage] Ignore running instances before August 2022

easyCZ:
## Description

We've got bad data quality on workspace_instances from the past. This excludes these from our queries for running workspaces until we can fix the source of truth.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage/pkg/db/workspace_instance.go
+++ b/components/usage/pkg/db/workspace_instance.go
@@ -80,6 +80,8 @@ func FindRunningWorkspaceInstances(ctx context.Context, conn *gorm.DB) ([]Worksp
 	tx := queryWorkspaceInstanceForUsage(ctx, conn).
 		Where("wsi.stoppingTime = ?", "").
 		Where("wsi.usageAttributionId != ?", "").
+		// We cannot guarantee data quality before this date
+		Where("wsi.startedTime > ?", TimeToISO8601(time.Date(2022, 8, 1, 0, 0, 0, 0, time.UTC))).
 		FindInBatches(&instancesInBatch, 1000, func(_ *gorm.DB, _ int) error {
 			instances = append(instances, instancesInBatch...)
 			return nil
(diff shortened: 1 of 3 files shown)


svenefftinge (approved):

components/usage/pkg/db/workspace_instance.go
```diff
@@ -80,6 +80,8 @@ func FindRunningWorkspaceInstances(ctx context.Context, conn *gorm.DB) ([]Worksp
 	tx := queryWorkspaceInstanceForUsage(ctx, conn).
 		Where("wsi.stoppingTime = ?", "").
 		Where("wsi.usageAttributionId != ?", "").
+		// Exclude anything before August 2022, we cannot guarantee data quality before then
+		Where("wsi.startedTime > ?", TimeToISO8601(time.Date(2022, 8, 1, 0, 0, 0, 0, time.UTC))).
```
jankeromnes, line 84: Question: Wasn't this part of why we have a `billInstancesAfter`? 💭

svenefftinge: `billInstancesAfter` would be a check against the `effectiveTime`. Running Instances have their `effectiveTime` now or in the future.

easyCZ: The `billInstancesAfter` is/was used in the BillingService. We could re-use it here as well

svenefftinge: I don't think so. It should actually be named `billInstancesStoppedAfter` to make clear it has nothing to do with when a workspace was started.

svenefftinge: I think we should remove `billInstancesStoppedAfter` entirely. Doesn't make sense to have a global starting point. Instead we should have the information how usage gets billed associated with each cost center. I'm going to work on this in the context of https://github.com/gitpod-io/gitpod/issues/12741

PR #12767, opened 2022-09-08
[UBP] Add `getUsageLimit` and `setUsageLimit` methods to server

andrew-farries:
## Description

Add two new methods to the server API for getting and setting usage limits. Both new functions take an attributionId and work for both users and teams.

For backwards compatibility, leave the `getUsageLimitForTeam` and `setUsageLimitForTeam` methods as they are still used by the dashboard, but change them to be implemented in terms of the more general `get/set` methods.

## Related Issue(s)

## How to test

* Create a team with `Gitpod` in the name in the preview environment and sign the team up for usage based pricing.
* Run the following in the browser developer console:

```js
await window._gp.gitpodService.server.setUsageLimit('team:<teamId of the new team>', 90)
```

* Then:
```js
await window._gp.gitpodService.server.getUsageLimit('team:<teamId of the new team>')
```

The team usage limit can also be updated from the team billing page.

The usage limits for the team are correctly persisted. 

Setting usage limits from the the UI for individual users will be done in a later PR.

## Release Notes

```release-note
NONE
```

## Werft options:
- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/server/ee/src/workspace/gitpod-server-impl.ts
+++ b/components/server/ee/src/workspace/gitpod-server-impl.ts
@@ -2246,15 +2246,6 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
         }
 
         const user = this.checkAndBlockUser("setUsageLimit");
-        switch (attrId.kind) {
-            case "team":
-                const team = await this.guardTeamOperation(attrId.teamId, "update");
-                await this.ensureStripeApiIsAllowed({ team });
-                break;
-            case "user":
-                await this.ensureStripeApiIsAllowed({ user });
-                break;
-        }
         await this.guardCostCenterAccess(ctx, user.id, attrId, "update");
 
         const response = await this.usageService.getCostCenter({ attributionId });
--- a/components/server/ee/src/workspace/gitpod-server-impl.ts
+++ b/components/server/ee/src/workspace/gitpod-server-impl.ts
@@ -2218,42 +2218,71 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
         return this.getStripePortalUrl(ctx, AttributionId.render(attributionId));
     }
 
-    async getUsageLimitForTeam(ctx: TraceContext, teamId: string): Promise<number | undefined> {
-        const user = this.checkAndBlockUser("getUsageLimitForTeam");
-        const team = await this.guardTeamOperation(teamId, "get");
-        await this.ensureStripeApiIsAllowed({ team });
+    async getUsageLimit(ctx: TraceContext, attributionId: string): Promise<number | undefined> {
+        const attrId = AttributionId.parse(attributionId);
+        if (attrId === undefined) {
+            log.error(`Invalid attribution id: ${attributionId}`);
+            throw new ResponseError(ErrorCodes.BAD_REQUEST, `Invalid attibution id: ${attributionId}`);
+        }
 
-        const attributionId: AttributionId = { kind: "team", teamId };
-        await this.guardCostCenterAccess(ctx, user.id, attributionId, "get");
+        const user = this.checkAndBlockUser("getUsageLimit");
+        await this.guardCostCenterAccess(ctx, user.id, attrId, "get");
 
-        const costCenter = await this.usageService.getCostCenter({
-            attributionId: AttributionId.render(attributionId),
-        });
+        const costCenter = await this.usageService.getCostCenter({ attributionId });
         if (costCenter?.costCenter) {
             return costCenter.costCenter.spendingLimit;
         }
         return undefined;
     }
 
-    async setUsageLimitForTeam(ctx: TraceContext, teamId: string, usageLimit: number): Promise<void> {
-        const user = this.checkAndBlockUser("setUsageLimitForTeam");
-        const team = await this.guardTeamOperation(teamId, "update");
-        await this.ensureStripeApiIsAllowed({ team });
-        if (typeof usageLimit !== "number" || usageLimit < 0) {
-            throw new ResponseError(ErrorCodes.BAD_REQUEST, "Unexpected `usageLimit` value.");
+    async setUsageLimit(ctx: TraceContext, attributionId: string, usageLimit: number): Promise<void> {
+        const attrId = AttributionId.parse(attributionId);
+        if (attrId === undefined) {
+            log.error(`Invalid attribution id: ${attributionId}`);
+            throw new ResponseError(ErrorCodes.BAD_REQUEST, `Invalid attibution id: ${attributionId}`);
         }
-        const attrId: AttributionId = { kind: "team", teamId };
+        if (typeof usageLimit !== "number" || usageLimit < 0) {
+            throw new ResponseError(ErrorCodes.BAD_REQUEST, `Unexpected usageLimit value: ${usageLimit}`);
+        }
+
+        const user = this.checkAndBlockUser("setUsageLimit");
+        switch (attrId.kind) {
+            case "team":
+                const team = await this.guardTeamOperation(attrId.teamId, "update");
+                await this.ensureStripeApiIsAllowed({ team });
+                break;
+            case "user":
+                await this.ensureStripeApiIsAllowed({ user });
+                break;
+        }
         await this.guardCostCenterAccess(ctx, user.id, attrId, "update");
-        const attributionId = AttributionId.render(attrId);
-        const costCenter = await this.usageService.getCostCenter({ attributionId });
+
+        const response = await this.usageService.getCostCenter({ attributionId });
+        if (response?.costCenter?.billingStrategy !== CostCenter_BillingStrategy.BILLING_STRATEGY_STRIPE) {
+            throw new ResponseError(
+                ErrorCodes.BAD_REQUEST,
+                `Setting a usage limit is not valid for non-Stripe billing strategies`,
+            );
+        }
         await this.usageService.setCostCenter({
             costCenter: {
                 attributionId,
                 spendingLimit: usageLimit,
                 billingStrategy:
-                    costCenter?.costCenter?.billingStrategy || CostCenter_BillingStrategy.BILLING_STRATEGY_OTHER,
+                    response?.costCenter?.billingStrategy || CostCenter_BillingStrategy.BILLING_STRATEGY_OTHER,
             },
         });
+    }
+
+    async getUsageLimitForTeam(ctx: TraceContext, teamId: string): Promise<number | undefined> {
+        const attributionId: AttributionId = { kind: "team", teamId: teamId };
+
+        return this.getUsageLimit(ctx, AttributionId.render(attributionId));
+    }
+
+    async setUsageLimitForTeam(ctx: TraceContext, teamId: string, usageLimit: number): Promise<void> {
+        const attributionId: AttributionId = { kind: "team", teamId: teamId };
+        return this.setUsageLimit(ctx, AttributionId.render(attributionId), usageLimit);
     }
 
     async getNotifications(ctx: TraceContext): Promise<string[]> {
--- a/components/server/src/auth/rate-limiter.ts
+++ b/components/server/src/auth/rate-limiter.ts
@@ -219,6 +219,8 @@ const defaultFunctions: FunctionsConfig = {
     getIDEOptions: { group: "default", points: 1 },
     getPrebuildEvents: { group: "default", points: 1 },
     setUsageAttribution: { group: "default", points: 1 },
+    getUsageLimit: { group: "default", points: 1 },
+    setUsageLimit: { group: "default", points: 1 },
     getUsageLimitForTeam: { group: "default", points: 1 },
     setUsageLimitForTeam: { group: "default", points: 1 },
     getNotifications: { group: "default", points: 1 },
--- a/components/gitpod-protocol/src/gitpod-service.ts
+++ b/components/gitpod-protocol/src/gitpod-service.ts
@@ -293,6 +293,8 @@ export interface GitpodServer extends JsonRpcServer<GitpodClient>, AdminServer, 
     subscribeTeamToStripe(teamId: string, setupIntentId: string): Promise<void>;
     getStripePortalUrl(attributionId: string): Promise<string>;
     getStripePortalUrlForTeam(teamId: string): Promise<string>;
+    getUsageLimit(attributionId: string): Promise<number | undefined>;
+    setUsageLimit(attributionId: string, usageLimit: number): Promise<void>;
     getUsageLimitForTeam(teamId: string): Promise<number | undefined>;
     setUsageLimitForTeam(teamId: string, spendingLimit: number): Promise<void>;
 
--- a/components/server/src/workspace/gitpod-server-impl.ts
+++ b/components/server/src/workspace/gitpod-server-impl.ts
@@ -3158,6 +3158,14 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
         throw new ResponseError(ErrorCodes.SAAS_FEATURE, `Not implemented in this version`);
     }
 
+    async getUsageLimit(ctx: TraceContext, attributionId: string): Promise<number | undefined> {
+        throw new ResponseError(ErrorCodes.SAAS_FEATURE, `Not implemented in this version`);
+    }
+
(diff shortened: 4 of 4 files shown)


components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2161,6 +2161,20 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
         return undefined;
     }
 
+    async getUsageLimitForUser(ctx: TraceContext): Promise<number | undefined> {
```
geropl, line 2164: @andrew-farries Why can't we generalize this to `getUsageLimit(attributionId)`? :thinking:  I think that way it would be way less code and complexity. And there should be zero differences between the two, as access is abstracted away behind the `guardCostCenterAccess`.

andrew-farries: Do you mean keep `(get|set)UsageLimitFor(Team|User)` on the server API but have them call a new method `getUsageLimitFor(attributionId)`? 

Or remove the `get|set` methods from the server in favour of a new `getUsageLimitFor(attributionId)`?

geropl: Ultimately the API should have:
 - getUsageLimit(attributionId)
 - setUsageLimit(attributionId)

If it's easier to 1. introduce the new methods, and remove the old usage+methods in a 2. PR that's fine with me.

andrew-farries: Rebased to add the two new methods, and rewrote the PR title + description.

We'll leave the old `get|setUsageLimitForTeam` methods and update the dashboard to use the more general ones in another PR.

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2146,35 +2146,64 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
         }
     }
 
-    async getUsageLimitForTeam(ctx: TraceContext, teamId: string): Promise<number | undefined> {
-        const user = this.checkAndBlockUser("getUsageLimitForTeam");
-        const team = await this.guardTeamOperation(teamId, "get");
-        await this.ensureStripeApiIsAllowed({ team });
+    async getUsageLimitFor(ctx: TraceContext, attributionId: string): Promise<number | undefined> {
+        const attrId = AttributionId.parse(attributionId);
+        if (attrId == undefined) {
```
geropl, line 2151: nit: `==` might do the trick here, but it instantly reads like a (potential) bug. We tend to use `===` everywhere.

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2146,35 +2146,64 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
 
-        const attributionId = AttributionId.render({ kind: "team", teamId });
+        const user = this.checkAndBlockUser("getUsageLimitFor");
+        switch (attrId.kind) {
+            case "team":
+                const team = await this.guardTeamOperation(attrId.teamId, "get");
+                await this.ensureStripeApiIsAllowed({ team });
+                break;
+            case "user":
+                await this.ensureStripeApiIsAllowed({ user });
```
geropl, line 2163: :exclamation: We need a guardCostCenter here as well.

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2146,35 +2146,64 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
+        if (attrId == undefined) {
+            log.error(`Invalid attribution id: ${attributionId}`);
+            throw new ResponseError(ErrorCodes.BAD_REQUEST, `Invalid attibution id: ${attributionId}`);
+        }
 
-        const attributionId = AttributionId.render({ kind: "team", teamId });
+        const user = this.checkAndBlockUser("getUsageLimitFor");
+        switch (attrId.kind) {
+            case "team":
+                const team = await this.guardTeamOperation(attrId.teamId, "get");
```
geropl, line 2159: This should be a `guardCostCenter`, as this is an operation on that entity.

components/server/src/auth/rate-limiter.ts
```diff
@@ -220,6 +220,8 @@ const defaultFunctions: FunctionsConfig = {
     getIDEOptions: { group: "default", points: 1 },
     getPrebuildEvents: { group: "default", points: 1 },
     setUsageAttribution: { group: "default", points: 1 },
+    getUsageLimitFor: { group: "default", points: 1 },
```
geropl, line 223: nit: I would have expected `getUsageLimit` instead of `getUsageLimitFor`

components/server/ee/src/workspace/gitpod-server-impl.ts
andrew-farries: This is done below the `switch` on line 2166:

https://github.com/gitpod-io/gitpod/blob/3a730199be6aacbde3d999a6bba8cc50046e386f/components/server/ee/src/workspace/gitpod-server-impl.ts#L2166

geropl: Thx, noticed [here](https://github.com/gitpod-io/gitpod/pull/12767#discussion_r966694409), forgot to update. :+1:

components/server/src/auth/rate-limiter.ts
andrew-farries: Renamed

andrew-farries:
This has some conflicts to resolve now that https://github.com/gitpod-io/gitpod/pull/12776 is merged (putting costcenter access behind an API). 

Moving back to draft until that's done.

svenefftinge:
FWIW we can only allow updating usagelimit for costcenters that are on pay-as-you-go (billingStrategy === 'stripe').

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2146,39 +2146,87 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
+                await this.ensureStripeApiIsAllowed({ team });
+                break;
+            case "user":
+                await this.ensureStripeApiIsAllowed({ user });
+                break;
+        }
+        await this.guardCostCenterAccess(ctx, user.id, attrId, "get");
+
+        const costCenter = await this.usageServiceClientProvider.getDefault().getCostCenter(attrId);
+        if (costCenter?.billingStrategy !== "stripe") {
```
AlexTugarev, line 2169: @andrew-farries, if I got @svenefftinge correct, this would be still valid. We are going to map other strategies and assign usage limits internally. 

The guard in `setUsageLimit` looks good though!

svenefftinge: Yes, reading them is fine. Just not changing.

andrew-farries:
I'm confused 😕 

* I have a team in the preview environment with usage based billing enabled.
* The cost center in the db looks like:
image

* But attempting to update the spending limit for the team results in:
image

* Extra `server` logs show:

```
{"component":"server","severity":"INFO","time":"2022-09-09T12:32:21.857Z","message":"attributionId is team:9ae55607-8b8c-4654-a076-c6492579446b"}
{"component":"server","severity":"INFO","time":"2022-09-09T12:32:21.857Z","message":"attrId is {\"kind\":\"team\",\"teamId\":\"9ae55607-8b8c-4654-a076-c6492579446b\"}"}
{"component":"server","severity":"INFO","time":"2022-09-09T12:32:21.860Z","message":"costCenter is {\"id\":{\"kind\":\"team\",\"teamId\":\"9ae55607-8b8c-4654-a076-c6492579446b\"},\"spendingLimit\":100,\"billingStrategy\":\"other\"}"}
{"component":"server","severity":"INFO","time":"2022-09-09T12:32:21.861Z","context":{"userId":"c3f3b0d5-2e22-40c4-8ac1-cdb0c904158b"},"message":"Request setUsageLimitForTeam unsuccessful: 400/\"Setting a usage limit is not valid for non-Stripe billing strategies\"","payload":{"method":"setUsageLimitForTeam","args":["9ae55607-8b8c-4654-a076-c6492579446b",30,{"_isCancelled":false}]}}
```

Which shows that the costCenter retrieved has a `billingStrategy` of `other`, but in the db it's clearly `stripe`.

I suspected something is wrong with the `getCostCenter` RPC, but we have unit tests covering this case here:

 https://github.com/gitpod-io/gitpod/blob/8c31c8599ccc226cba3c7ec24b23e405302446d0/components/usage/pkg/apiv1/usage_test.go#L221-L262

andrew-farries:
Perhaps it relates to the fact the stripe billing strategy has the value `0` in the enum in the proto file:

https://github.com/gitpod-io/gitpod/blob/8c31c8599ccc226cba3c7ec24b23e405302446d0/components/usage-api/usage/v1/usage.proto#L109-L117

[gRPC docs ](https://developers.google.com/protocol-buffers/docs/proto3#enum)usually have the `0` value as a special 'unspecified' value.

andrew-farries:
The problem lies in the `usage-api` usage `sugar.ts` file:

https://github.com/gitpod-io/gitpod/blob/fc2b4422c6cb308ef48481e3883fcb48bd97faa7/components/usage-api/typescript/src/usage/v1/sugar.ts#L203

here, we are using `||` against an enum which has `0` as a possible value (for the stripe billing strategy). So the expression currently always returns the `other` billing strategy.

I really dislike these sugar files, they are an untested, handwritten layer over the generated gRPC client in which bugs like this can lurk.

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2146,39 +2146,83 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
         if (typeof usageLimit !== "number" || usageLimit < 0) {
-            throw new ResponseError(ErrorCodes.BAD_REQUEST, "Unexpected `usageLimit` value.");
+            throw new ResponseError(ErrorCodes.BAD_REQUEST, `Unexpected usageLimit value: ${usageLimit}`);
+        }
+
+        const user = this.checkAndBlockUser("setUsageLimit");
+        switch (attrId.kind) {
+            case "team":
+                const team = await this.guardTeamOperation(attrId.teamId, "update");
+                await this.ensureStripeApiIsAllowed({ team });
```
svenefftinge, line 2189: This checks for `m.mode === "usage-based" || (m.mode === "chargebee" && !!m.canUpgradeToUBB)` which seems like I could then change my spendinglimit although I haven't put my credit card into stripe, yet. 
Maybe we should reconsider the special 'chargebee' case? What was the idea of it? cc @geropl

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2146,39 +2146,83 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
-        const attributionId: AttributionId = { kind: "team", teamId };
-        await this.guardCostCenterAccess(ctx, user.id, attributionId, "get");
+        const user = this.checkAndBlockUser("getUsageLimit");
+        switch (attrId.kind) {
+            case "team":
+                const team = await this.guardTeamOperation(attrId.teamId, "get");
+                await this.ensureStripeApiIsAllowed({ team });
+                break;
+            case "user":
+                await this.ensureStripeApiIsAllowed({ user });
```
svenefftinge, line 2163: We should not guard read access but only write access to the cost center based on billingMode.

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2146,39 +2146,83 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
+                break;
+        }
+        await this.guardCostCenterAccess(ctx, user.id, attrId, "update");
+
+        log.info(`attributionId is ${attributionId}`);
+        log.info(`attrId is ${JSON.stringify(attrId)}`);
```
svenefftinge, line 2198: why do you log the `attributionId` twice here? Also we usually have static messages and use variable data in an additional object that we pass as second argument. Makes it easier to search for this kind of log messages.

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2146,39 +2146,83 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
+        const costCenter = await this.usageServiceClientProvider.getDefault().getCostCenter(attrId);
+        log.info(`costCenter is ${JSON.stringify(costCenter)}`);
+        if (costCenter?.billingStrategy !== "stripe") {
```
svenefftinge, line 2201: I think this function should expect a costCenter to exist and thwor an error if not.

components/server/ee/src/workspace/gitpod-server-impl.ts
andrew-farries: I'm not sure I fully understand the comment. Do you mean this check (line 2163 in this diff) should be removed?

components/server/ee/src/workspace/gitpod-server-impl.ts
andrew-farries: Logs lines removed. They were only really useful to help track down the issue from [this comment](https://github.com/gitpod-io/gitpod/pull/12767#issuecomment-1242053307)

components/server/ee/src/workspace/gitpod-server-impl.ts
andrew-farries: I think I agree, but I'd like to address it in a separate PR as there are a few other usages of the function besides this one.

components/server/ee/src/workspace/gitpod-server-impl.ts
svenefftinge: Yes, because we are using costcenters now in any case. It is enough to guard the write access.

andrew-farries: Check removed.

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2146,39 +2146,75 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
+            log.error(`Invalid attribution id: ${attributionId}`);
+            throw new ResponseError(ErrorCodes.BAD_REQUEST, `Invalid attibution id: ${attributionId}`);
+        }
 
-        const attributionId: AttributionId = { kind: "team", teamId };
-        await this.guardCostCenterAccess(ctx, user.id, attributionId, "get");
+        const user = this.checkAndBlockUser("getUsageLimit");
+        if (attrId.kind === "team") {
+            const team = await this.guardTeamOperation(attrId.teamId, "get");
+            await this.ensureStripeApiIsAllowed({ team });
```
svenefftinge, line 2159: Sorry if I wasn't clear. I meant to remove these calls for both attribution kinds (user and team).

components/server/ee/src/workspace/gitpod-server-impl.ts
svenefftinge: Let's remove the ènsureStripeApiISAllowed` checks here as well. The check below testing the `billingStrategy !== "stripe"` is doing what we need and is more explicit.

geropl: @svenefftinge Originally, Chargebee and Usage-Limit was strictly independent. After last weeks discussion, they are not anymore.

> Let's remove the ensureStripeApiIsAllowed checks here as well.

:100: Agreed!

andrew-farries:
Needs rebased again now that https://github.com/gitpod-io/gitpod/pull/12903 landed 😅

components/server/ee/src/workspace/gitpod-server-impl.ts
svenefftinge: @andrew-farries do you agree to remove the calls to `ensureStripeApiIsAllowed`?

andrew-farries: Yes; I've removed them in a650ae74308b32c2ff910a0e8a1bd3354b990a38

svenefftinge (approved):
👍

PR #12769, opened 2022-09-08
[biling] Remove UpdateInvoices - old reconciler

easyCZ:
## Description

No longer used, we only use the new reconciler - `ReconcileInvoices`. Protos will be removed in subsequent PR to isolate generated and actual changes.

## Related Issue(s)

Fixes #

## How to test

CI unit/build

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


components/usage/pkg/server/server.go
```diff
@@ -36,7 +35,8 @@ type Config struct {
 
 	StripeCredentialsFile string `json:"stripeCredentialsFile,omitempty"`
 
-	ContentServiceAddress string `json:"contentServiceAddress,omitempty"`
+	// Deprecated
```
andrew-farries, line 38: Why do we need to deprecate this? Because we're no longer uploading usage reports to object storage?

components/usage/pkg/apiv1/billing.go
```diff
@@ -23,50 +22,20 @@ import (
 	"gorm.io/gorm"
 )
 
-func NewBillingService(stripeClient *stripe.Client, billInstancesAfter time.Time, conn *gorm.DB, contentService contentservice.Interface) *BillingService {
+func NewBillingService(stripeClient *stripe.Client, conn *gorm.DB) *BillingService {
```
andrew-farries, line 25: Do we not want to keep the `billInstancesAfter` behaviour in the new reconciler?

andrew-farries (approved):
/hold to avoid blocking the merge queue while the build is failing.

components/usage/pkg/apiv1/billing.go
easyCZ: From the [discussion here](https://github.com/gitpod-io/gitpod/pull/12766#discussion_r965800503) it looks like not.

components/usage/pkg/server/server.go
easyCZ: Yep, that's correct. I want to ensure that the service doesn't fail to start with config which does set it in case we rollback the version but not the config.

PR #12776, opened 2022-09-08
[cost center] put all costcenter access behind API

svenefftinge:
## Description
This change has unfortunately gotten a little big (still practicing keeping things small 😇).
It does the following:
 - introduce a write API for cost center on the `usage`component
 - replace any direct DB access of cost center in `server` with this API
 - added `creationTime` and `billingStrategy` to cost center

## Related Issue(s)

Fixes #

## How to test

You can verify that everything still works by [joining the Gitpod team](https://sefftinge-22b683c9ef.preview.gitpod-dev.com/teams/join?inviteId=0a990e86-7d57-46b7-8cea-792caf3a0ffd) and changing the spending limit

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment


components/gitpod-db/src/typeorm/migration/1662639748206-CostCenterPaymentStrategy.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const D_B_COST_CENTER = "d_b_cost_center";
+const COL_CREATION_TIME = "creationTime";
+const COL_BILLING_STRATEGY = "billingStrategy";
+
+export class CostCenterPaymentStrategy1662639748206 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        if (!(await columnExists(queryRunner, D_B_COST_CENTER, COL_CREATION_TIME))) {
+            await queryRunner.query(
+                `ALTER TABLE ${D_B_COST_CENTER} ADD COLUMN ${COL_CREATION_TIME} varchar(30) NOT NULL, ALGORITHM=INPLACE, LOCK=NONE `,
```
andrew-farries, line 18: Why do we use a `varchar` to store a date here? Would [datetime](https://dev.mysql.com/doc/refman/8.0/en/timestamp-initialization.html) be a better choice?

andrew-farries (approved):

andrew-farries:
Added the hold for question about the data type.

components/usage/pkg/db/cost_center.go
```diff
@@ -15,13 +15,20 @@ import (
...
-	SpendingLimit int32         `gorm:"column:spendingLimit;type:int;default:0;" json:"spendingLimit"`
-	LastModified  time.Time     `gorm:"->:column:_lastModified;type:timestamp;default:CURRENT_TIMESTAMP(6);" json:"_lastModified"`
+	ID              AttributionID   `gorm:"primary_key;column:id;type:char;size:36;" json:"id"`
+	CreationTime    VarcharTime     `gorm:"primary_key;column:creationTime;type:varchar;size:255;" json:"creationTime"`
+	SpendingLimit   int32           `gorm:"column:spendingLimit;type:int;default:0;" json:"spendingLimit"`
+	BillingStrategy BillingStrategy `gorm:"column:billingStrategy;type:varchar;size:255;" json:"billingStrategy"`
 
-	// deleted is restricted for use by db-sync
-	_ bool `gorm:"column:deleted;type:tinyint;default:0;" json:"deleted"`
+	LastModified time.Time `gorm:"->:column:_lastModified;type:timestamp;default:CURRENT_TIMESTAMP(6);" json:"_lastModified"`
```
AlexTugarev, line 31: out of curiosity: are the gorm definitions manually crafted, or is there a tool to sync them?

components/gitpod-db/src/typeorm/migration/1662639748206-CostCenterPaymentStrategy.ts
svenefftinge: Because we generally use this as a date type (apart from `_lastModified` columns). 
Happy to re-discuss our general strategy. Maybe something to bring up in a team meeting?

AlexTugarev, line 18: they will be filled up with the implicit default value for the `varchar` type right? which is `''`, an empty string.

components/usage/pkg/db/cost_center.go
svenefftinge: manual

components/gitpod-db/src/typeorm/migration/1662639748206-CostCenterPaymentStrategy.ts
svenefftinge: Oh, true. That was not my intend, but I think it will not break.

svenefftinge: I'm actually not sure what MYSQL will do with existing records here 😬 The migration worked well against an empty table.

svenefftinge: > they will be filled up with the implicit default value for the varchar type right? which is '', an empty string.

Just verified that this is the case, indeed

components/usage/pkg/db/cost_center.go
easyCZ: There's a tool which can generate them from the DB schema. I used those initially to set up the definitions but often requires some manual tweaking.

svenefftinge: Something I actually forgot to mention in yesterday's retro is that I'm getting a little tired of maintaining/writing all the different technical representations (DB & protobuf for TS&Go) of the same shapes. Feels like a lot of code (e.g. this PR) is really just about shoveling data through the stack by copying it into different formats. 😓

easyCZ: That's fair. Codegen can help but it's also important to keep the API shape independent of the shape of the DB. It tends to start very similar, but over time changes. The DB is expected to evolve while the API is expected to retain compatibility.

svenefftinge: Yes, I think it would already help if we didn't need to duplicate API to route dashboard requests through server. Also, I think the manual `sugar.ts` code could maybe be generated.

Issue #12628, opened 2022-09-02
Preview environment takes too long to start, causing workspace integration tests to fail


kylos101:
### Bug description

Sometimes the preview envrionment takes too long to start, causing the initial workspace integration test job to fail.

### Steps to reproduce

Refer to this [internal conversation](https://gitpod.slack.com/archives/C03E52788SU/p1662051454479299) for context.

### Expected behavior

Ideally we'd wait longer for the environment to start, allowing for the tests to run, or the environment would start up quicker.

mads-hartmann:
Some quick thoughts
1. We should update your job to be resilient to errors from Werft. From a quick glance it looks like it's just failing if we get at 503 when polling the state of the jobs. In that case we can just ignore the error and poll again :relaxed:
1. We should have better insights the Werft API - currently we have no monitoring for that AFAIK

mads-hartmann:
And a quick clarification: The integration tests don't fail because the preview environments start too slowly, it's because we get an error when querying for the state of the Werft job for the preview environments. At least from looking at the logs from the job https://werft.gitpod-dev.com/job/gitpod-custom-to-stablization-test.19 mentioned in the thread ☺️

```
[build preview environment] Waiting for gitpod-build-wk-inte-test-20220901164435.0 to finish running. Current phase: PHASE_RUNNING. Sleeping 10 seconds.
Error: rpc error: code = Unavailable desc = Service Unavailable: HTTP status code 503; transport: received the unexpected content-type "text/html"
[slack notification and cleanup EXIT|PHASE] Slack notification and cleanup: EXIT
[slack notification] Sending Slack notificaition
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   507    0     2  100   505     10   2729 --:--:-- --:--:-- --:--:--  2740
```

mads-hartmann:
It looks like Werft is restarting quite often:

```
 kubectl -n werft get pods werft-0
W0906 12:52:20.714506  119023 gcp.go:120] WARNING: the gcp auth plugin is deprecated in v1.22+, unavailable in v1.25+; use gcloud instead.
To learn more, consult https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke
NAME      READY   STATUS    RESTARTS   AGE
[...]

PR #12784, opened 2022-09-08
previews: Several improvements to clean up logic

ArthurSens:
## Description

Turns out this PR got a long bigger than initially expected. That is because I ended up making several changes in a single PR. 

1) Introduce job that cleans-up individual previews.
2) Update preview clean-up cronjob to re-use the job for individial clean ups .
3) Update job that creates artificial preview to also delete the preview at the end.

Every change is done in separate commits, so please use them to make the review process easier 🙂 

## Related Issue(s)

Fixes #

## How to test

You can test each change with different werft commands:

1) You can manually create previews and delete it individually with the following commands:
```
# Create new preview
werft run github -a with preview
# Delete preview
werft run github -j platform-delete-preview-environment.yaml -a preview="your preview name"
```

2) When triggering the cron job, notice that now it will trigger several new jobs. One for each stale preview
```
werft run github -j platform-delete-preview-environments-cron.yaml
```

3) When triggering the job that creates an artificial preview, notice that it will also trigger the deletion job at the end
```
werft run github -j platform-trigger-artificial-job.yaml
```

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/.werft/platform-delete-preview-environment.ts
+++ b/.werft/platform-delete-preview-environment.ts
@@ -1,0 +1,99 @@
+import { Werft } from "./util/werft";
+import * as Tracing from "./observability/tracing";
+import { HarvesterPreviewEnvironment, PreviewEnvironment } from "./util/preview";
+import { SpanStatusCode } from "@opentelemetry/api";
+import { exec } from "./util/shell";
+import { CORE_DEV_KUBECONFIG_PATH, HARVESTER_KUBECONFIG_PATH } from "./jobs/build/const";
+import * as fs from "fs";
+
+// Will be set once tracing has been initialized
+let werft: Werft;
+
+const context: any = JSON.parse(fs.readFileSync("context.json").toString());
+const annotations = context.Annotations || {};
+const previewName = annotations["preview"] || "";
+// for testing purposes
+// if set to 'true' it shows only previews that would be deleted
+const DRY_RUN = annotations["dry-run"] in annotations || false;
+
+const SLICES = {
+    VALIDATE_CONFIGURATION: "Validating configuration",
+    CONFIGURE_ACCESS: "Configuring access to relevant resources",
+    INSTALL_HARVESTER_KUBECONFIG: "Install Harvester kubeconfig",
+    DELETING_PREVIEW: `Deleting preview environment: ${previewName}`,
+};
+
+Tracing.initialize()
+    .then(() => {
+        werft = new Werft("delete-preview-environment-cron");
+    })
+    .then(() => deletePreviewEnvironment())
+    .catch((err) => {
+        werft.rootSpan.setStatus({
+            code: SpanStatusCode.ERROR,
+            message: err,
+        });
+        console.error("Werft job failed with an error", err);
+        // Explicitly not using process.exit as we need to flush tracing, see tracing.js
+        process.exitCode = 1;
+    })
+    .finally(() => {
+        werft.phase("Flushing telemetry", "Flushing telemetry before stopping job");
+        werft.endAllSpans();
+    });
+
+async function deletePreviewEnvironment() {
+    // Fail early if no preview was passed through annotations.
+    werft.log(SLICES.VALIDATE_CONFIGURATION, "Validating annotations");
+    if (previewName == "") {
+        werft.fail(
+            SLICES.VALIDATE_CONFIGURATION,
+            "A preview name is required. Please inform the preview name with '-a preview=<name of the preview>'.",
+        );
+    }
+    werft.done(SLICES.VALIDATE_CONFIGURATION);
+
+    werft.phase("Configure access");
+    try {
+        const GCLOUD_SERVICE_ACCOUNT_PATH = "/mnt/secrets/gcp-sa/service-account.json";
+        exec(`gcloud auth activate-service-account --key-file "${GCLOUD_SERVICE_ACCOUNT_PATH}"`, {
+            slice: SLICES.CONFIGURE_ACCESS,
+        });
+        exec(
+            `KUBECONFIG=${CORE_DEV_KUBECONFIG_PATH} gcloud container clusters get-credentials core-dev --zone europe-west1-b --project gitpod-core-dev`,
+            { slice: SLICES.CONFIGURE_ACCESS },
+        );
+        werft.done(SLICES.CONFIGURE_ACCESS);
+    } catch (err) {
+        werft.fail(SLICES.CONFIGURE_ACCESS, err);
+    }
+
+    werft.phase("Install Harvester kubeconfig");
+    try {
+        exec(`cp /mnt/secrets/harvester-kubeconfig/harvester-kubeconfig.yml ${HARVESTER_KUBECONFIG_PATH}`, {
+            slice: SLICES.INSTALL_HARVESTER_KUBECONFIG,
+        });
+        werft.done(SLICES.INSTALL_HARVESTER_KUBECONFIG);
+    } catch (err) {
+        werft.fail(SLICES.INSTALL_HARVESTER_KUBECONFIG, err);
+    }
+
+    const preview = new HarvesterPreviewEnvironment(werft, previewName);
+    if (DRY_RUN) {
+        werft.log(SLICES.DELETING_PREVIEW, `Would have deleted preview ${preview.name}`);
+    } else {
+        removePreviewEnvironment(preview);
+    }
+}
+
+async function removePreviewEnvironment(previewEnvironment: PreviewEnvironment) {
+    werft.log(SLICES.DELETING_PREVIEW, `Starting deletion of all resources related to ${previewEnvironment.name}`);
+    try {
+        // We're running these promises sequentially to make it easier to read the log output.
+        await previewEnvironment.removeDNSRecords(SLICES.DELETING_PREVIEW);
+        await previewEnvironment.delete();
+        werft.done(SLICES.DELETING_PREVIEW);
+    } catch (e) {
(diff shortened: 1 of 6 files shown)


ArthurSens:
/hold because it wasn't tested enough

but getting some feedback regarding the approach would be nice 🙂

vulkoingim:
I think this makes the cleanup script more complex, and we want the opposite.

How about we create a job that is `delete-preview-environment` that accepts a preview env name as input and would just run https://github.com/gitpod-io/gitpod/blob/main/.werft/platform-delete-preview-environments-cron.ts#L412-L423

We can call it at the end of the `deploy-preview-env` with a flag.

Also this would allow us to do a few more nice things:
* Other jobs - e.g. the ws teams' integration tests can trigger it when it finishes so it cleans up the preview env
* We can trigger the job on branch delete (i.e. when PR gets merged)
* We can have the delete cron run that job for every individual preview env, instead of looping through them - this would simplify the script overall and would surface issues when deleting a single preview env, instead of burying it in the logs of the bigger job.

ArthurSens:
> * Other jobs - e.g. the ws teams' integration tests can trigger it when it finishes so it cleans up the preview env

I was expecting that we could add another type of preview created for integration tests actually, and they would be deleted with different premises

> * We can trigger the job on branch delete (i.e. when PR gets merged)

That sounds like a good idea, but it also sounds like a significant change 😅. 

> * We can have the delete cron run that job for every individual preview env, instead of looping through them - this would simplify the script, reduce the overall runtime, and would surface issues when deleting a single preview env, instead of burying it in the logs of the bigger job.

I do like the current strategy of deleting in batches tho... We have a collection of premisses, and we periodically purge previews that follow our premises. Sounds pretty simple and effective.

I don't follow how having multiple different scripts simplify anything here 🤔

vulkoingim:
> I was expecting that we could add another type of preview created for integration tests actually, and they would be deleted with different premises

And you'll require more logic to determine if any of those environments are stale, as they might have different requirements. By shifting the responsibility of the clean-up of those environments to the job that is responsible for them - you are certain that whatever they were used for is done and they can be deleted. 

E.g. - in its current form this PR will try to delete any env that matches that type. But the job that is spinning up the env might not have finished yet. If you want to be sure that it is finished - now you have to implement something else to check that.  (I imagine that you would like to change the order of evaluation, since now it won't have an effect and it won't get cleaned up earlier)

> That sounds like a good idea, but it also sounds like a significant change 😅.

It's going to be an additional way to delete a preview env, not instead of the cron job. They serve different purposes. 

I'm not saying to implement this now - but it will allow us to have this later very easily.

> I do like the current strategy of deleting in batches tho... We have a collection of premisses, and we periodically purge previews that follow our premises. Sounds pretty simple and effective.

Nothing will change there for the cron job, except where the logic for deletion will be handled. The cron job will still loop through all preview envs and determine which one need to be removed - only it will trigger a different job for each env that will handle the deletion. And it will be faster as then multiple delete jobs will be running in parallel.

> I don't follow how having multiple different scripts simplify anything here 🤔

The logic you introduce will not be needed anymore + we'll get the rest of the benefits that I described.

ArthurSens:
> And you'll require more logic to determine if any of those environments are stale, as they might have different requirements.

We'll just be adding another premise, right? I don't understand how complex this can be

> By shifting the responsibility of the clean-up of those environments to the job that is responsible for them - you are certain that whatever they were used for is done and they can be deleted.

It is not uncommon that jobs fail mid-execution. In such cases, we'll not be cleaning up those previews. With a garbage collection strategy, we're safe to fail and everything will be cleaned up eventually.

In a scenario where we have no room for any stale preview, I'd understand that periodic clean-ups wouldn't work, but that's not our case.


> E.g. - in its current form this PR will try to delete any env that matches that type. But the job that is spinning up the env might not have finished yet. If you want to be sure that it is finished

That's a good point! I haven't thought about it, I probably want to adjust my premise to account for preview age (30 mins should be totally fine).

> Now you have to implement something else to check that. If it were an integration test (e.g.) - it might get interrupted in the middle and fail. (I imagine that you would like to change the order of evaluation, since now it won't have an effect and it won't get cleaned up earlier)

Yep, for different types of preview we'd have different premises. I don't think we have that many to say that we're increasing complexity that much. This PR is introducing "regular" and "artificial", nothing more than that. It might make sense to add another for "integration-test", but that's it. Do you see us adding dozens of types here?

> > I do like the current strategy of deleting in batches tho... We have a collection of premisses, and we periodically purge previews that follow our premises. Sounds pretty simple and effective.
> 
> Nothing will change there for the cron job, except where the logic for deletion will be handled. The cron job will still loop through all preview envs and determine which one need to be removed - only it will trigger a different job for each env that will handle the deletion. And it will be faster as then multiple delete jobs will be running in parallel.

Ah ok, now I'm getting it, the logic to delete a single preview will be reused by the cron. Sounds nice, we'll get parallelism power and we won't have duplicated code.

> > I don't follow how having multiple different scripts simplify anything here 🤔
> 
> The logic you introduce will not be needed anymore + we'll get the rest of the benefits that I described.

What if the job fails midway and we have dangling previews uncleaned? I really think this logic is still needed 😬

ArthurSens:
/unhold

Finally tested and happy with the results 🙂

.werft/platform-delete-preview-environment.ts
```diff
@@ -0,0 +1,99 @@
...
+const SLICES = {
+    VALIDATE_CONFIGURATION: "Validating configuration",
+    CONFIGURE_ACCESS: "Configuring access to relevant resources",
+    INSTALL_HARVESTER_KUBECONFIG: "Install Harvester kubeconfig",
+    DELETING_PREVIEW: `Deleting preview environment: ${previewName}`,
+};
+
+Tracing.initialize()
+    .then(() => {
+        werft = new Werft("delete-preview-environment-cron");
```
mads-hartmann, line 28:
```suggestion
        werft = new Werft("delete-preview-environment");
```

.werft/platform-delete-preview-environments-cron.ts
```diff
@@ -411,58 +216,69 @@ async function determineStalePreviewEnvironments(options: {
...
 async function removePreviewEnvironment(previewEnvironment: PreviewEnvironment) {
     const sliceID = `Deleting preview ${previewEnvironment.name}`;
-    werft.log(sliceID, `Starting deletion of all resources related to ${previewEnvironment.name}`);
+    werft.log(sliceID, `Triggering job to delete ${previewEnvironment.name}. DRY_RUN=${DRY_RUN}`);
     try {
-        // We're running these promises sequentially to make it easier to read the log output.
-        await previewEnvironment.removeDNSRecords(sliceID);
-        await previewEnvironment.delete();
+        exec(
+            `sudo chown -R gitpod:gitpod /workspace && git config --global user.name roboquat && git config --global user.email roboquat@gitpod.io`,
```
mads-hartmann, line 222: Do we need this on every execution of removePreviewEnvironment? Could we move this to to the yaml file - that's where we have this "init" code in other jobs ☺️

.werft/platform-trigger-artificial-job.sh
```diff
@@ -15,6 +15,40 @@ sudo chown -R gitpod:gitpod /workspace
...
+git checkout -b "$BRANCH" | werft log slice "Creating new branch"
 
-werft run github -a with-preview=true
+BUILD_ID=$(werft run github -a with-preview=true)
+job_url="https://werft.gitpod-dev.com/job/${BUILD_ID}"
+echo "Triggered job $job_url" | werft log slice "Triggering new job"
+werft log result -d "build job" url "$job_url"
+
+while true; do
+    job_phase=$(werft job get "${BUILD_ID}" -o json | jq --raw-output '.phase')
```
mads-hartmann, line 27: Werft is currently a bit unreliable (see https://github.com/gitpod-io/gitpod/issues/12628#issuecomment-1239429502) so there is a good chance that this will fail every once in a while.

You could temporarily disable "fail on error" and check the exit code manually. Something like this (not tested)

```sh
set +e
while true; do
    job_phase=$(werft job get "${BUILD_ID}" -o json | jq --raw-output '.phase')
    if [[ "$?" != "0" ]]; then
        echo "Failed to get job details. Sleeping and trying again"
        sleep 10
    fi
    // .. the rest of your code
done
set -e
```

This is how workspace [worked around it for now](https://github.com/gitpod-io/gitpod/blob/main/.werft/workspace-run-integration-tests.sh#L113)

mads-hartmann (approved):
I left a few comments - approving so you can decide if you want to fix them now or in a follow up PR ☺️

.werft/platform-trigger-artificial-job.sh
ArthurSens: Ah nice catch

ArthurSens: Actually I looked it up and while set +e/-e doesn't make any difference in while loops 🤔 
image

mads-hartmann: @ArthurSens I think that only applies to commands executed in the "condition" part of those expressions. E.g. try to run this script

```sh
#!/usr/bin/env bash

set -euo pipefail

function fail {
    if [[ $1 == "yes" ]]; then
        exit 1
    else
        exit 0
    fi
}

while $(fail "yes"); do
    echo "This shouldn't run"
done

while $(fail "no"); do
    echo "Running"
    sleep 1
    fail "yes"
done
```

You'll see

```
gitpod /workspace/empty (main) $ ./test.sh 
Running
```

E.g. that `fail "yes"` does cause the entire script to fail ☺️

PR #12787, opened 2022-09-08
[stripe] Finalized invoce debits, do not set credits to negative when updating

easyCZ:
1. We need to debit the credit amount in usage, not credit the credit amount
2. We shouldn't try to update usage to negative values. This doesn't make sense (and isn't allowed by Stripe). Instead we set it to 0. This happens when a customer has for example bought a bundle or overpaid on an invoice.

## Related Issue(s)

* Depends on https://github.com/gitpod-io/gitpod/pull/12755

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage/pkg/apiv1/billing.go
+++ b/components/usage/pkg/apiv1/billing.go
@@ -108,7 +108,8 @@ func (s *BillingService) FinalizeInvoice(ctx context.Context, in *v1.FinalizeInv
 		ID:            uuid.New(),
 		AttributionID: attributionID,
 		Description:   fmt.Sprintf("Invoice %s finalized in Stripe", invoice.ID),
-		CreditCents:   db.NewCreditCents(float64(creditsOnInvoice)),
+		// Apply negative value of credits to reduce accrued credit usage
+		CreditCents:   db.NewCreditCents(float64(-creditsOnInvoice)),
 		EffectiveTime: db.NewVarcharTime(finalizedAt),
 		Kind:          db.InvoiceUsageKind,
 		Draft:         false,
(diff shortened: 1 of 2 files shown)


components/usage/pkg/apiv1/billing.go
```diff
@@ -118,7 +108,8 @@ func (s *BillingService) FinalizeInvoice(ctx context.Context, in *v1.FinalizeInv
 		ID:            uuid.New(),
 		AttributionID: attributionID,
 		Description:   fmt.Sprintf("Invoice %s finalized in Stripe", invoice.ID),
-		CreditCents:   db.NewCreditCents(float64(creditsOnInvoice)),
+		// Apply negative value of credits to reduce accrued credit usage
+		CreditCents:   db.NewCreditCents(float64(-creditsOnInvoice)),
```
svenefftinge, line 112: Ideally we'd do it the otherway around. Maybe we can create a PR with a migration that negates all credits and changes the respective code as well?

svenefftinge (approved):

components/usage/pkg/apiv1/billing.go
svenefftinge: Actually thinking about it again, it might be fine as is. After all the table is called usage so positive means used negative means gained. 🤷

easyCZ: We can still make the change and drop the usage records manually

easyCZ: Spoke directly, we'll be keeping it as is to more closely align with Stripe which also doesn't want negative credits, but positive credits for usage

PR #12794, opened 2022-09-09
[terraform] Add k3s setup for tests

### LINKED ISSUE #12866: [terraform] Move k3s module to reference architecture section

nandajavarma:
## Description

This PR adds a very rough single-cluster setup for k3s. We are adding this for consistency with Azure, AWS and GCP setups so we can consistently run the tests. We do not recommend users to use this because the underlying module still make use of `k3sup` and we have plans to move to `k3s` before suggesting this as a solution. Hence we have not included a Readme with this directory.

## Related Issue(s)

Fixes #12866

## How to test

* Go to the directory `install/infra/single-cluster/k3s`.
* Edit `main.tf` to have the right bucket name.
* Edit `terraform.tfvars` to right parameters for the cluster.
Then run:
```
make init
make plan
make apply
# After testing, make sure you cleanup with:
make destroy
```

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/infra/single-cluster/k3s/cluster.tf
+++ b/install/infra/single-cluster/k3s/cluster.tf
@@ -1,0 +1,16 @@
+module "k3s" {
+  source = "../../modules/k3s"
+
+  name             = var.name
+  gcp_project      = var.project
+  gcp_region       = var.region
+  gcp_zone         = var.zone
+  credentials      = var.credentials_path
+  kubeconfig       = var.kubeconfig
+  dns_sa_creds     = var.credentials_path
+  dns_project      = var.project
+  managed_dns_zone = var.managed_dns_zone
+  domain_name      = var.domain_name
+  cluster_version  = var.cluster_version
+  image_id         = var.image_id
+}
--- a/install/infra/single-cluster/k3s/tools.tf
+++ b/install/infra/single-cluster/k3s/tools.tf
@@ -1,0 +1,19 @@
+module "certmanager" {
+  source = "../../modules/tools/cert-manager"
+
+  kubeconfig  = var.kubeconfig
+}
+
+module "cluster-issuer" {
+  source              = "../../modules/tools/issuer"
+  kubeconfig          = var.kubeconfig
+  gcp_credentials     = local.credentials
+  issuer_name         = "cloudDNS"
+  cert_manager_issuer = {
+    project                 = var.project
+    serviceAccountSecretRef = {
+      name = "clouddns-dns01-solver"
+      key  = "keys.json"
+    }
+  }
+}
--- a/install/infra/single-cluster/k3s/variables.tf
+++ b/install/infra/single-cluster/k3s/variables.tf
@@ -1,0 +1,48 @@
+variable "kubeconfig" {
+  description = "The KUBECONFIG file path to store the resulting KUBECONFIG file to"
+  default     = "./kubeconfig"
+}
+
+variable "project" {
+  description = "Google cloud Region to perform operations in"
+}
+
+variable "region" {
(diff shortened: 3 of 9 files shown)


roboquat:

<details>

</details>

install/infra/single-cluster/k3s/tools.tf
```diff
@@ -0,0 +1,28 @@
...
+  kubeconfig  = var.kubeconfig
+}
+
+module "externaldns" {
+  source      = "../../modules/tools/cloud-dns-external-dns"
+  kubeconfig  = var.kubeconfig
+  credentials = var.credentials
+  project     = var.project
+  region      = var.region
+  zone        = var.zone
```
adrienthebo, line 13: **Issue(blocker):** The cloud-dns-external-dns module doesn't define `region` and `zone` variables, can we omit these?

```suggestion
```

```
gitpod /workspace/gitpod/install/infra/single-cluster/k3s (nvn/k3s-tf-mod) $ make apply
╷
│ Warning: Backend configuration ignored
│ 
│   on ../../modules/k3s/main.tf line 11, in terraform:
│   11:   backend "gcs" {
│ 
│ Any selected backend applies to the entire configuration, so Terraform expects provider configurations only in the root module.
│ 
│ This is a warning rather than an error because it's sometimes convenient to temporarily call a root module as a child module for testing purposes, but this backend configuration block will have no effect.
╵
╷
│ Error: Unsupported argument
│ 
│   on tools.tf line 12, in module "externaldns":
│   12:   region      = var.region
│ 
│ An argument named "region" is not expected here.
╵
╷
│ Error: Unsupported argument
│ 
│   on tools.tf line 13, in module "externaldns":
│   13:   zone        = var.zone
│ 
│ An argument named "zone" is not expected here.
╵
make: *** [Makefile:101: apply-cluster] Error 1
```

install/infra/single-cluster/k3s/variables.tf
```diff
@@ -0,0 +1,48 @@
...
+  description = "Google cloud Region to perform operations in"
+  default     = "europe-west1"
+}
+
+variable "zone" {
+  description = "Google cloud Zone to perform operations in"
+  default     = "europe-west1-b"
+}
+
+variable "credentials" {
```
adrienthebo, line 20: In a prior pull request we started passing credentials as a variable; this reduces the dependency on having files on-disk but this variable assumes that we're passing in the credentials directly.

Inserting the credentials into `terraform.tfvars` doesn't work well as the JSON strings need to have their double quotes escaped and terraform doesn't like the newlines in the JSON document.

WDYT of having this as `credentials_path` and then using something like the following:?

```
locals {
  credentials = "${file(var.credentials_path)}"
}
```

Is this a better approach or do you believe it'll be cheaper to switch back to passing around credentials by a file path?

nandajavarma: Great point!! has updated accordingly!

install/infra/single-cluster/k3s/cluster.tf
```diff
@@ -0,0 +1,16 @@
+module "k3s" {
+  source = "../../modules/k3s"
+
+  name             = var.name
+  gcp_project      = var.project
+  gcp_region       = var.region
+  gcp_zone         = var.zone
+  credentials      = var.credentials_path
+  kubeconfig       = var.kubeconfig
+  dns_sa_creds     = var.credentials_path
```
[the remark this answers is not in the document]
nandajavarma: The `k3s` module expects the file path and not the JSON content itself. This change would break the module

PR #12796, opened 2022-09-09
[ws-manager-bridge] Don't report workspace instance update failures as both a failure and a success

jankeromnes:
## Description

Currently, when we fail to update a workspace instance in ws-manager-bridge, we report it once as a failure (in the `catch`), then again as a success (in the `finally`, which always gets called regardless of whether errors happened):

https://github.com/gitpod-io/gitpod/blob/7a0efe9999c3cc31ffc63c79fe9d8334620bee37/components/ws-manager-bridge/src/bridge.ts#L190-L217

I think it would make more sense to report errors as errors, and only successes as successes.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/ws-manager-bridge/src/bridge.ts
+++ b/components/ws-manager-bridge/src/bridge.ts
@@ -205,16 +205,15 @@ export class WorkspaceManagerBridge implements Disposable {
             );
             log.error(logCtx, "Failed to complete WorkspaceInstance status update", e);
             throw e;
-        } finally {
-            const durationMs = performance.now() - start;
-            this.prometheusExporter.reportWorkspaceInstanceUpdateCompleted(
-                durationMs / 1000,
-                writeToDB,
-                this.cluster.name,
-                status.spec.type,
-            );
-            log.info(logCtx, "Successfully completed WorkspaceInstance status update");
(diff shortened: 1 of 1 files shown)


geropl (approved):
LGTM

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -205,16 +205,15 @@ export class WorkspaceManagerBridge implements Disposable {
...
-            log.info(logCtx, "Successfully completed WorkspaceInstance status update");
         }
+        const durationMs = performance.now() - start;
+        this.prometheusExporter.reportWorkspaceInstanceUpdateCompleted(
+            durationMs / 1000,
+            writeToDB,
+            this.cluster.name,
+            status.spec.type,
+        );
+        log.info(logCtx, "Successfully completed WorkspaceInstance status update");
```
easyCZ, line 216: Alternatively you could bind `e` inside the catch to a variable outside of the try scope and then access it inside finally. This would avoid (almost) duplicating the report call.

PR #12800, opened 2022-09-09
[dashboard] update feature flags only on demand

akosyakov:
## Description

Each workspace page is loading the dashboard in iframe to show loading screen while workspace is not running. It causes a lot of requests each 3 minutes to ConfigCat. This PR changes the strategy to update the settings values on demand and cache them for one minute. For the workspace page it would mean only one request. I don't think it makes big effect on other pages, but generally reduce amount of requests for the dashboard.

## Related Issue(s)

Fixes #

## How to test

-

## Release Notes

```release-note
```

## Werft options:

- [x] /werft with-preview


roboquat:

<details>

</details>

components/dashboard/src/experiments/client.ts
```diff
@@ -35,10 +35,9 @@ export function getExperimentsClient(): Client {
 function newProductionConfigCatClient(): Client {
     // clientKey is an identifier of our ConfigCat application. It is not a secret.
     const clientKey = "WBLaCPtkjkqKHlHedziE9g/TwAe6YyftEGPnGxVRXd0Ig";
-    const client = configcat.createClient(clientKey, {
+    const client = configcat.createClientWithLazyLoad(clientKey, {
         logger: configcat.createConsoleLogger(LogLevel.Error),
-        pollIntervalSeconds: 60 * 3, // 3 minutes
-        maxInitWaitTimeSeconds: 0,
+        cacheTimeToLiveSeconds: 60 * 1, // 1 minute
     });
```
easyCZ, line 41: I'm not sure if this is desired behaviour from a UX perspective. This would block a UI element until the flags have been fetched.

What's the driver for this change?

easyCZ: > What's the driver for this change?

Just read the description as well.

akosyakov: > Each workspace page is loading the dashboard in iframe to show loading screen while workspace is not running. It causes a lot of requests each 3 minutes to ConfigCat.

> I'm not sure if this is desired behaviour from a UX perspective. This would block a UI element until the flags have been fetched.

We can reduce timeout time to 1500 milliseconds as we did in VS Code Desktop.

akosyakov: I increased cache time to 3 mins and set timeout to 1500 milliseconds.

easyCZ: Makes sense. I think I'm okay with this. Will just play around in preview.

akosyakov: Also looking at the code it does not seem that it can block, since flags are compute in effect, so in worst case flags won't be applied.

easyCZ: Sorry, I should've been clearer. I meant block as in not show for example the Usage UI immediately, only after it's fetched.

akosyakov: We can call force update once on page load?

akosyakov: > Sorry, I should've been clearer. I meant block as in not show for example the Usage UI immediately, only after it's fetched.

I don't think timing is changing. The client is created only on demand in use effect.

easyCZ (approved):
Let's try this out.

PR #12806, opened 2022-09-09
[usage] Include drafts when listing usage summary

easyCZ:
## Description

We need to use the same draft flags between ListUsage and ListUsageSummary, else the metadata about pagination does not match.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage/pkg/apiv1/usage.go
+++ b/components/usage/pkg/apiv1/usage.go
@@ -78,8 +78,9 @@ func (s *UsageService) ListUsage(ctx context.Context, in *v1.ListUsageRequest) (
 	if in.GetPagination().GetPage() > 1 {
 		page = in.GetPagination().GetPage()
 	}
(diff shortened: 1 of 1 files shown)


components/usage/pkg/apiv1/usage.go
```diff
@@ -78,15 +78,17 @@ func (s *UsageService) ListUsage(ctx context.Context, in *v1.ListUsageRequest) (
 	if in.GetPagination().GetPage() > 1 {
 		page = in.GetPagination().GetPage()
 	}
-	var offset int64 = perPage * (page - 1)
+	var offset = perPage * (page - 1)
```
easyCZ, line 81: drive-by fix, the type is not needed

svenefftinge (approved):

PR #12810, opened 2022-09-09
[usage] added a test for ListUsage

svenefftinge:
## Description
adds a test for UsageService.ListUsage

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage/pkg/apiv1/usage_test.go
+++ b/components/usage/pkg/apiv1/usage_test.go
@@ -7,6 +7,7 @@ package apiv1
 import (
 	"context"
 	"database/sql"
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


components/usage/pkg/apiv1/usage_test.go
```diff
@@ -259,3 +260,84 @@ func TestGetAndSetCostCenter(t *testing.T) {
...
+					Page:    1,
+				},
+			})
+			require.NoError(t, err)
+
+			require.Equal(t, test.creditsAtStart, metaData.CreditBalanceAtStart)
+			require.Equal(t, test.creditsAtEnd, metaData.CreditBalanceAtEnd)
+			require.Equal(t, test.recordsInRange, metaData.Pagination.Total)
+		})
+	}
```
easyCZ, line 341:
```suggestion
	tests := []struct {
	    name string // test name
		start, end time.Time
		// expectations
		creditsAtStart float64
		creditsAtEnd   float64
		recordsInRange int64
	}{
		{start, end, 3, 10, 2},
		{end, end, 10, 10, 0},
		{start, start, 3, 3, 0},
		{start.Add(-200 * 24 * time.Hour), end, 0, 10, 4},
		{start.Add(-200 * 24 * time.Hour), end.Add(10 * 24 * time.Hour), 0, 20, 5},
	}

	for i, test := range tests {
		t.Run(test.name, func(t *testing.T) {
			metaData, err := usageService.ListUsage(context.Background(), &v1.ListUsageRequest{
				AttributionId: string(attributionID),
				From:          timestamppb.New(test.start),
				To:            timestamppb.New(test.end),
				Order:         v1.ListUsageRequest_ORDERING_DESCENDING,
				Pagination: &v1.PaginatedRequest{
					PerPage: 1,
					Page:    1,
				},
			})
			require.NoError(t, err)

			require.Equal(t, test.creditsAtStart, metaData.CreditBalanceAtStart)
			require.Equal(t, test.creditsAtEnd, metaData.CreditBalanceAtEnd)
			require.Equal(t, test.recordsInRange, metaData.Pagination.Total)
		})
	}
```

It makes it nicer for humans on what you're testing, but also reflects it into the test when it fails

easyCZ (approved):
/hold in case you want to add name to the tests

components/usage/pkg/apiv1/usage_test.go
svenefftinge: It makes sense generally. In this case, I believe printing the number of the test is fine as well. But will definitely consider this in future!

Issue #12453, opened 2022-08-27
Workspace Cluttered with core.* files


a0v0:
### Bug description

![image](https://user-images.githubusercontent.com/62926341/187032364-cb0cdb83-5308-42e1-853d-0c945b2d06f7.png)

### Steps to reproduce

Install anything and you will see these files

Run:
`brew install go-task/tap/go-task`

<img src="https://front.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_3q51m)

pawlean:
Hey @a0v0! I've just tried to replicate this but was unable to. Are you still seeing it?

axonasif:
Related https://github.com/gitpod-io/gitpod/issues/11760

axonasif:
I am able to reproduce it by running `brew install tmux`, note that my `pwd` was `$GITPOD_REPO_ROOT`.

axonasif:
Rolling back to old brew doesn't seem to help:
```bash
cd $GITPOD_REPO_ROOT

git -C "/home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/homebrew/homebrew-core" checkout 4e1a5bafc78c38ab37fb2a6ef05f3235428f64b3

git -C "/home/linuxbrew/.linuxbrew/Homebrew" checkout 3.4.0

rm -f core.*; brew uninstall tmux; brew install tmux; ls -la
```

sharebear:
I'm experiencing this too... I didn't see the issue when doing `brew install <thing>` interactively in the terminal, but it's definitely when calling `brew install <thing>` during `init` on a task similar to @a0v0's example.

kylos101:
:wave: hi, we might be able to resolve this issue by making a change to our `workspacekit` component, by setting soft limits for the workspace, using the [prlimit syscall](https://linux.die.net/man/2/prlimit). For example, set `RLIMIT_CORE` to 0. [Reference](https://linux.die.net/man/2/prlimit#:~:text=to%20this%20size.-,RLIMIT_CPU,-CPU%20time%20limit). 

I've added this issue to our team's backlog, so we can further discuss later this week. We'll share more updates soon! :pray:

Furisto:
[...]

PR #12814, opened 2022-09-09
Disable core dumps for workspaces 

Furisto:
## Description
Disable core dumps for workspaces 

## Related Issue(s)
Fixes https://github.com/gitpod-io/gitpod/issues/12453

## How to test
- Open workspace
- ` sleep 900&`
- `kill -s SIGSEGV [pid]`

## Release Notes

```release-note
Prevent core dumps from cluttering workspaces
```

## Werft options:

- [x] /werft with-preview


Furisto:
@utam0k This prevents core dumps from being created but it also means that the user is not allowed to change the core limit anymore with prlimit or ulimit (due to operation not permitted error) in case he actually wants to create a core dump for debugging purposes.

I have allowed prlimit64 and setrlimit in the seccomp profile (strace shows that both tools are using prlimit64) but this did not have any effect. Do you have any idea why what the problem could be?

david-bakin:
Am I understanding @Furisto 's comment above correctly:  I can't get coredumps now _at all_?  If so: 👎   This cannot be an acceptable outcome of this PR.  How is one supposed to do any work in C++ and other languages (Rust, possibly Go, Python when debugging C++ addins)?  Core dumps are a problem in github (see #13102) but there must be a way to _allow_ them, even if not by default.  (And that way should be _documented_ properly too.)

utam0k:
> @utam0k This prevents core dumps from being created but it also means that the user is not allowed to change the core limit anymore with prlimit or ulimit (due to operation not permitted error) in case he actually wants to create a core dump for debugging purposes.
> 
> I have allowed prlimit64 and setrlimit in the seccomp profile (strace shows that both tools are using prlimit64) but this did not have any effect. Do you have any idea why what the problem could be?

My first impression is a lack of capability. We need `CAP_SYS_PTRACE` to run `strace`.

csweichel:
> We definitely would like to preserve that ability.

What stops users from re-starting their workspace with the corresponding `.gitpod.yml` change?

Issue #9189, opened 2022-04-08
extract IDE service from server


akosyakov:
Right now IDE configuration of the workspace is controlled by the server. IDE configuration is maintained by @gitpod-io/engineering-ide, though the deployment of the server is maintained by @gitpod-io/engineering-webapp. Instead we should move IDE configuration (code and configmap) to a new component which can be deployed by the IDE team and establish the interface between server and such service to resolve IDE configuration.

akosyakov:
It came up again while working on JB dev loop optimization we had to change different APIs and code of ws-manager and server. We should rather encapsulate such logic in our component and generalize APIs & internals of ws-manager and server.

It will allow us more fine grain optimization without relying on other teams.

jeanp413:
This week we need to do migration

loujaybee:
This one should be an epic according to @iQQBot, let's discuss and break this one down 🙏 

Assigning to myself to work with Pudong on that!

PR #12815, opened 2022-09-09
Add ide-service server

mustard-mh:
## Description

Add ide-service to resolve all IDE related things. First implement to bring 
- Make `ide-service` a component and deploy
- Resolve image digest
- Resolve nightly/latest tag of image
- Verify config content
- API to get config content (string)

## Related Issue(s)

Relates https://github.com/gitpod-io/gitpod/issues/9189

## How to test

- Open this PR with Gitpod
- Exec commands below to start a test service
```
cd components/ide-service
go run main.go run --config ./example-config.json
```
- Exec test command to get parsed ide content
```
go run main.go test | jq
```
- `latestImage` should not contain `nightly` or `latest` tag, should be resolved

### ide-service in prev env
- Open this PR with Gitpod
- Wait until preview env installed
- Exec commands below

```
# exec sh
kubectl exec -it ide-service<press_tab_to_get> -c ide-service sh

# run test cmd
./app/ide-service test --config /config/config.json
```
- Output should be correct and align to server ide config response

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


mustard-mh:
@iQQBot  Could you help review this PR 🙏

components/ide-service-api/ide.proto
```diff
@@ -0,0 +1,20 @@
...
+option java_package = "io.gitpod.ideservice.api";
+
+service IDEService {
+    rpc GetConfig(GetConfigRequest) returns (GetConfigResponse) {}
+}
+
+message GetConfigRequest {}
+
+message GetConfigResponse {
+    string content = 1;
```
akosyakov, line 19: I think for first iteration is alright 👍  let's bring it to production and switch server to resolve from here, only after that we should move type definitions here and do parsing in our service

components/ide-service-api/typescript-grpcweb/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "version": "0.0.1",
+  "license": "UNLICENSED",
+  "scripts": {
+    "build": "bash build.sh"
+  },
+  "files": [
+    "lib"
+  ],
+  "dependencies": {
+    "grpc-web": "^1.3.1"
```
akosyakov, line 12: we should use https://github.com/improbable-eng/grpc-web as we do for other services, check out on main supervisor-api and ide-metrics-api

grpc-web does not support Node.js and server is Node.js

iQQBot: we don't need http support, I think we can delete this

akosyakov: I'm not sure. Server is using supervisor via grpcweb https://github.com/gitpod-io/gitpod/blob/19455a4a9a31a55d033fe95331cdfe9eacaeec8d/components/server/package.json#L38 How will we use?

components/ide-service-api/ide.proto
iQQBot: I think in the follow-up process, we should shield the implementation details of IDEConfig from other external components, they just need to pass the user settings and context information through an interface similar to IDEChoose, and then the ide-service will take care of all the ide selection logic and return the results to the caller

components/ide-service-api/typescript-grpcweb/package.json
iQQBot: this is because supervisor is external service, we don't support grpc in `ws-proxy`

akosyakov: ok, then let's drop all unnecessary packages

components/ide-service-api/ide.proto
akosyakov: I hoped that we could reuse types in installer if we define them here or just in go

iQQBot (approved):
Nice work 👍 , I left some comment

We need to add `ide-server` in here, so that installer can known this component need render
https://github.com/gitpod-io/gitpod/blob/ceb75ebb61b9793158d7a904eaaf2fd7c0a07a0c/install/installer/pkg/components/components-webapp/components.go#L31-L32

And we need to add `ide-service` to CODEOWNER file

and werft build failed, otherwise code LGTM 🎉 

/hold

components/ide-service-api/ide.proto
mustard-mh: Yes, we can move types to `go` folder under `ide-service-api`  in next PR, for other components to reuse them

mustard-mh:
Force pushed to

- Add CODEOWNERS
- Remove ide-service-api `typescript-grpcweb`
- Fix `scheduleUpdate` not called
- Add installer component to render

andrew-farries (approved):

components/ide-service/pkg/server/ideconfig_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"testing"
+
+	ctesting "github.com/gitpod-io/gitpod/common-go/testing"
+)
+
+func TestCreateDefiniteWorkspacePod(t *testing.T) {
```
geropl, line 15: nit: the test name reads like a copy'n'paste oversight :upside_down_face:

adrienthebo (approved):

components/ide-service/pkg/server/ideconfig_test.go
mustard-mh: 🙏 Rebased to change test name @geropl

install/installer/pkg/components/ide-service/configmap.go
```diff
@@ -0,0 +1,52 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
+
+	cfg := config.ServiceConfiguration{
+		Server: &baseserver.Configuration{
+			Services: baseserver.ServicesConfiguration{
+				GRPC: &baseserver.ServerConfiguration{
+					Address: fmt.Sprintf(":%d", GRPCServicePort),
```
MrSimonEmms, line 25: /hold this will break the new HTTP_PROXY

```suggestion
					Address: fmt.Sprintf("0.0.0.0:%d", GRPCServicePort),
```

mustard-mh: Addressed

MrSimonEmms: Thank you

PR #12831, opened 2022-09-09
[usage] Refactor controller package into scheduler

easyCZ:
## Description

Refactor existing controller package into the following:
* Scheduler, capable of scheduling a function with a specified schedule
* Scheduler now tracks metrics for all jobs it runs
* Job Mutex moves outside of the scheduler and into each job

## Related Issue(s)

Fixes #

## How to test

Unit

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


components/usage/pkg/scheduler/job.go
```diff
@@ -2,55 +2,72 @@
...
 
+	select {
+	// attempt a write to signal we want to run
+	case r.running <- struct{}{}:
+		// we managed to write, there's no other job executing. Cases are not fall through so we continue executing our main logic.
+	default:
+		// we could not write, so another instance is already running. Skip current run.
+		log.Infof("Skipping ledger run, another run is already in progress.")
+		return nil
+	}
```
easyCZ, line 61: Mutex logic moves into the job, away from the controller itself. This is needed to allow us to run multiple control loops with the same scheduler.

components/usage/pkg/scheduler/reporter.go
```diff
@@ -16,26 +16,26 @@ const (
 )
 
 var (
-	reconcileStartedTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
+	jobStartedSeconds = prometheus.NewCounterVec(prometheus.CounterOpts{
```
easyCZ, line 19: These are now general job metrics, rather than specific for the reconcile with ledger. If a job desires more metrics, those can be added in job direclty rather than on the scheduler level.

components/usage/pkg/scheduler/job.go
```diff
@@ -2,55 +2,72 @@
...
 }
 
-func NewLedgerReconciler(usageClient v1.UsageServiceClient, billingClient v1.BillingServiceClient) *LedgerReconciler {
-	return &LedgerReconciler{
+func NewLedgerTrigger(usageClient v1.UsageServiceClient, billingClient v1.BillingServiceClient) *LedgerJob {
+	return &LedgerJob{
 		usageClient:   usageClient,
 		billingClient: billingClient,
+
+		running: make(chan struct{}, 1),
```
andrew-farries, line 39: Why does the channel have a buffer size of `1`?

components/usage/pkg/scheduler/reporter.go
```diff
@@ -16,26 +16,26 @@ const (
...
 		Namespace: namespace,
 		Subsystem: subsystem,
-		Name:      "reconcile_started_total",
-		Help:      "Number of usage reconciliation runs started",
-	}, []string{})
+		Name:      "scheduler_job_started",
```
andrew-farries, line 22: Prometheus [metric naming conventions](https://prometheus.io/docs/practices/naming/)  suggest this might be better to have the unit as a suffix
```suggestion
		Name:      "scheduler_jobs_started_total",
```

andrew-farries (approved):
/hold for a couple of questions

components/usage/pkg/scheduler/reporter.go
easyCZ: thanks, good spot

components/usage/pkg/scheduler/job.go
easyCZ: Because we only want to run 1 at a time. This is in effect a mutex, either we can write into the channel (it's not full) and hence acquire the mutes, or we fail (and the `select` executes the `default` branch) in which case we didn't acquire the mutex.

andrew-farries: Ah, ok. I was getting confused. `1` is the default buffer size for a channel though, so it could be omitted here.

easyCZ: Correct, I've included it just to be explicit, I'll also add a comment to this for future readers :)

Issue #12828, opened 2022-09-09
Build speed: Install/installer:docker


mads-hartmann:
Builds of install/installer:docker is taking rather long (40s) even when the component source code isn't changed.

Looking at the [Dockerfile](https://github.com/gitpod-io/gitpod/blob/main/install/installer/leeway.Dockerfile) we see that the order commands can be improved with respect to layer caching. We should move line 9 just below `COPY --from=alpine/helm:3.8.0 /usr/bin/helm /usr/bin/helm`

mads-hartmann:
Reopening as PR was reverted in https://github.com/gitpod-io/gitpod/pull/12860

PR #12834, opened 2022-09-09
Improve layer caching by moving apk commands up

mads-hartmann:
## Description

This moves the `apk` above the `COPY` instructions so that we don't invalidate the layers whenever the contents of the files changes ☺️

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/12828

## How to test

The Werft build should still succeed.

## Release Notes

```release-note
NONE
```

## Documentation

N/A

## Werft options:

- [ ] /werft with-preview


MrSimonEmms (approved):
LGTM

akosyakov:
It is only 4 secs faster though even with caching.

PR #12843, opened 2022-09-09
[UBP] Fix usage gRPC `sugar.ts` file

andrew-farries:
## Description

Fix the `sugar.ts` file that wraps the usage component gRPC client for typescript clients.

Use a return value of the correct shape rather than explicitly casting an incorrect shape.

In particular avoid the use of `||` against an enum type that can be 0.

See https://github.com/gitpod-io/gitpod/pull/12767#issuecomment-1242053307 for the full context.

I really dislike these sugar files, they are an untested, handwritten layer over the generated gRPC client in which bugs like this can lurk. 

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage-api/typescript/src/usage/v1/sugar.ts
+++ b/components/usage-api/typescript/src/usage/v1/sugar.ts
@@ -9,7 +9,15 @@ import { BillingServiceClient } from "./billing_grpc_pb";
 import { TraceContext } from "@gitpod/gitpod-protocol/lib/util/tracing";
 import * as opentracing from "opentracing";
 import { Metadata } from "@grpc/grpc-js";
-import { CostCenter as ProtocolCostCenter, GetCostCenterRequest, GetCostCenterResponse, ListUsageRequest, ListUsageResponse, SetCostCenterRequest, SetCostCenterResponse } from "./usage_pb";
+import {
+    CostCenter as ProtocolCostCenter,
+    GetCostCenterRequest,
+    GetCostCenterResponse,
+    ListUsageRequest,
+    ListUsageResponse,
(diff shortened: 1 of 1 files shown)


geropl:
> I really dislike these sugar files, they are an untested, handwritten layer over the generated gRPC client in which bugs like this can lurk.

Definitely. 

:telescope: We could check out `connect-web` ([link](https://github.com/bufbuild/connect-web)): it promises to generate usable TS code from gRPC protos.

components/usage-api/typescript/src/usage/v1/sugar.ts
```diff
@@ -176,59 +181,63 @@ export class PromisifiedUsageServiceClient {
...
+                resolve(response);
+            });
         });
         if (!response.hasCostCenter()) {
             return undefined;
         }
-        return <CostCenter> {
-            id: AttributionId.parse(response.getCostCenter()!.getAttributionId()),
+
+        const attrId = AttributionId.parse(response.getCostCenter()!.getAttributionId());
```
svenefftinge, line 201: Nit: It would be a contract violation of the GRPC service to return a costCenter without an attributionID. We should enforce the contract in the API impl itself rather than on the TypeScript bindings.

svenefftinge (approved):
I wonder why there are so many formatted changes. I thought we have the same formatting config set up on the workspaces.

andrew-farries:
> I wonder why there are so many formatted changes. I thought we have the same formatting config set up on the workspaces.

It looked as though `prettier` had somehow not been run on the file, rather than a different `prettier` config.

Issue #12660, opened 2022-09-05
Not able to get OAuth-like flow to callback VS Code in browser


jeanp413:
> I'm using vscode.env.asExternalUri() and this works perfecty in GitHub Codespaces. The same flow fails in Gitpod VS Code with a Not Found page. The URL looks correct, the host is the same as the actual workspace. 

More context https://discord.com/channels/816244985187008514/1015881863929274378/1015881867301498880

<img src="https://front.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_3qgwa)

ptrthomas:
@jeanp413 thanks ! when would this fix be available on gitpod.io/workspaces ?

jeanp413:
@ptrthomas it will take some time as deployment of this fix depends on workspace team schedule maybe 1-2 weeks

ptrthomas:
I was just able to confirm that this works in gitpod.io/workspaces

PR #12844, opened 2022-09-09
Adds vscode auth callback route

jeanp413:
## Description

Adds vscode auth callback route

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/12660

## How to test

1. Open a workspace and install this test extension [gitpod-foo-0.0.58.zip](https://github.com/gitpod-io/gitpod/files/9538509/gitpod-foo-0.0.58.zip)
2. Update settings.json with `"gitpod.host": "https://jp-test-9.preview.gitpod-dev.com/"`
3. Run `Test FOO sign in` command
4. Check the callback page opens sucessfully, it says `You can close this page now`
5. On vscode tab login will fail, that's expected
6. Not required: If you want login to succeed, then go to this [branch](https://github.com/gitpod-io/gitpod/tree/jp/test-9) and change the redirect url to match your workspace, pointer [[1](https://github.com/gitpod-io/gitpod/commit/6b78c8c63f74956ebeba23c324f93159e41d0506)], wait for build, yeah that sucks :facepalm: 
7. Check you are logged in 
image

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/proxy/conf/Caddyfile
+++ b/components/proxy/conf/Caddyfile
@@ -296,19 +296,18 @@ https://*.*.{$GITPOD_DOMAIN} {
 	import ssl_configuration
 	import debug_headers
 
-	@workspace_blobserve header_regexp host Host ^blobserve.ws(?P<location>-[a-z0-9]+)?.{$GITPOD_DOMAIN}
-	handle @workspace_blobserve {
-		gitpod.cors_origin {
-			base_domain {$GITPOD_DOMAIN}
-		}
-
+	# foreign content route used by vscode to serve webview and webworker resources of the form
+	# https://{{hash_base_32}}.<cluster>.<gitpod_domain> or https://v--{{hash_base_32}}.<cluster>.<gitpod_domain>
+	# e.g:
+	# https://0d9rkrj560blqb5s07q431ru9mhg19k1k4bqgd1dbprtgmt7vuhk.ws-us34xl.gitpod.io (for webviews)
+	# https://v--0d9rkrj560blqb5s07q431ru9mhg19k1k4bqgd1dbprtgmt7vuhk.ws-us34xl.gitpod.io (for webworker)
+	@foreign_content2 header_regexp host Host ^(?:v--)?[0-9a-v]+.ws(-[a-z0-9]+)?.{$GITPOD_DOMAIN}
+	handle @foreign_content2 {
 		reverse_proxy https://ws-proxy.{$KUBE_NAMESPACE}.{$KUBE_DOMAIN}:9090 {
 			import workspace_transport
 			import upstream_headers
 
 			header_up X-WSProxy-Host       {http.request.host}
-
-			header_down -access-control-allow-origin
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


sagor999 (approved):

andrew-farries (approved):

akosyakov:
@jeanp413 is it backward compatible for workspaces which would be already running during ws deployment or later rollback?

akosyakov:
@mustard-mh @iQQBot Could you have a look please?

components/proxy/conf/Caddyfile
```diff
@@ -296,19 +296,19 @@ https://*.*.{$GITPOD_DOMAIN} {
...
-		}
-
+	# foreign content route used by vscode to serve webview and webworker resources of the form
+	# https://{{hash_base_32}}.<cluster>.<gitpod_domain> or https://v--{{hash_base_32}}.<cluster>.<gitpod_domain>
+	# e.g:
+	# https://0d9rkrj560blqb5s07q431ru9mhg19k1k4bqgd1dbprtgmt7vuhk.ws-us34xl.gitpod.io (for webviews)
+	# https://v--0d9rkrj560blqb5s07q431ru9mhg19k1k4bqgd1dbprtgmt7vuhk.ws-us34xl.gitpod.io (for webworker)
+	# origin should be decoupled from the workspace (port) origin but the workspace (port) prefix should be the path root for routing
+	@foreign_content2 header_regexp host Host ^(?:v--)?[0-9a-v]+.ws(-[a-z0-9]+)?.{$GITPOD_DOMAIN}
+	handle @foreign_content2 {
```
akosyakov, line 306: @jeanp413 Do I understand corretly it is just tech debt, i.e. all workspaces for long time were using this route anyway?

components/ws-proxy/pkg/proxy/routes.go
```diff
@@ -101,12 +101,16 @@ func installWorkspaceRoutes(r *mux.Router, config *RouteHandlerConfig, ip Worksp
...
-	routes.HandleDirectIDERoute(enableCompression(r).MatcherFunc(func(req *http.Request, m *mux.RouteMatch) bool {
-		// this handles all foreign (none-IDE) content
-		return m.Vars != nil && m.Vars[foreignOriginIdentifier] != ""
-	}))
-
-	routes.HandleRoot(enableCompression(r).NewRoute())
+	rootRouter := enableCompression(r)
+	rootRouter.Use(func(h http.Handler) http.Handler {
+		return http.HandlerFunc(func(resp http.ResponseWriter, req *http.Request) {
+			if req.URL.Path == "/vscode-extension-auth-callback" {
```
akosyakov, line 107: what was an issue with using callback.html? The PR description does not explain anything.

components/ws-proxy/pkg/proxy/routes.go
```diff
@@ -153,16 +157,6 @@ func (ir *ideRoutes) HandleSSHHostKeyRoute(route *mux.Route, hostKeyList []ssh.S
 	})
 }
 
-func (ir *ideRoutes) HandleDirectIDERoute(route *mux.Route) {
```
akosyakov, line 156: why do we remove it?

components/proxy/conf/Caddyfile
jeanp413: yep

components/ws-proxy/pkg/proxy/routes.go
jeanp413: This is no longer used, it was related to the old way of accessing webview resources if I'm not mistaken

components/ws-proxy/pkg/proxy/routes.go
jeanp413: We can use `callback` (not sure about callback.html it should be an endpoint not a resource) but I thought that's too generic, maybe in the future we add another web ide that has some type of callback too.

components/ws-proxy/pkg/proxy/routes.go
akosyakov: It is direct route to workspace for web sockets, extension resources and so on. If we remove it and someone get URL workspace they can access it.

components/ws-proxy/pkg/proxy/routes.go
akosyakov: Just to understand, we would like to remove cross checks and auth for auth callback? I do not yet understand what this PR is fixing.

components/ws-proxy/pkg/proxy/routes.go
jeanp413: Isn't that handled by `HandleRoot` if not it wouldn't be working, it redirects to vscode server port for extension resources :thinking:

akosyakov: you are right 👍

akosyakov (approved):
lgtm

/hold

kylos101:
>  is it backward compatible for workspaces which would be already running during ws deployment or later rollback?

Good question, @akosyakov!

@jeanp413 for you to test this, I think in the preview environment, you'll need to replace the version of `ws-proxy` with what we have running in production, and then do your functional test again. Depending on the result, we may need to coordinate a patch of `ws-proxy` code and/or config in `workspace-cluster`.

To test:

```bash
# assuming your kubectx is for a preview environment

# get a yaml file containing the versions for the build asset being in used for the workspace-cluster deployment as of Sept 12
docker run --rm eu.gcr.io/gitpod-core-dev/build/versions:main.4550 cat /versions.yaml > versions.yaml

# replace ws-proxy in your preview environment
kubectl set image deployment ws-proxy ws-proxy=eu.gcr.io/gitpod-core-dev/build/ws-proxy:$(yq e '.components.wsProxy.version' versions.yaml)

# run your functional test for this PR
```

PR #12846, opened 2022-09-09
[supervisor-frontent] add error reporting

iQQBot:
## Description

This PR add error reporting to ide-metrics for supervisor-frontend

## Related Issue(s)

Fixes #

## Release Notes

```release-note
```

## Werft options:

- [ ] /werft with-preview


roboquat:
@iQQBot: Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://git.k8s.io/community/contributors/guide/release-notes.md) to remove it.

<details>

</details>

components/supervisor/frontend/src/index.ts
```diff
@@ -30,6 +30,12 @@ window.addEventListener('error', (event) => {
             labels['error'] = 'LoadError';
         }
     }
+    if (event.error) {
+        (async () => {
+            const gitpodServiceClient = await pendingGitpodServiceClient;
```
akosyakov, line 35: It seems to be failure point if this method breaks itself. Could we avoid waiting and only provide info which we have at the current point of time, i.e. get workapceId from Url, and instanceid, userid later?

components/supervisor/frontend/src/ide/ide-metrics-service-client.ts
```diff
@@ -40,4 +49,32 @@ export class IDEMetricsServiceClient {
             return false
         }
     }
+
+    static async reportError(error: Error, workspaceId: string, instanceId: string, userId: string) : Promise<boolean> {
+        const url = `${MetricsUrl}/reportError`
+        const params: ReportErrorParam = {
+            errorStack: error.stack ?? "",
```
akosyakov, line 56: I'm not sure it is enough, there is also `message` even if `stack` is missing, also as a complete fallback there is `toString`. I am not sure there are these options are mutually exclusive.

akosyakov: just fyi browser error telemetry in VS Code: 
- https://github.com/gitpod-io/openvscode-server/blob/ffd06f4d409bae03df48a3ad2e61f1523605fd20/src/vs/platform/telemetry/browser/errorTelemetry.ts#L11
- https://github.com/gitpod-io/openvscode-server/blob/ffd06f4d409bae03df48a3ad2e61f1523605fd20/src/vs/platform/telemetry/common/errorTelemetry.ts#L47

they seem to track msg and stack separately, they can also track line and column, not sure why we don't get this info, maybe we are instrumenting wrong points

akosyakov: interesting that they are buffer, remove duplicates and flush after 5s, not sure whether we need to do it as well now

components/supervisor/frontend/src/index.ts
```diff
@@ -30,6 +30,12 @@ window.addEventListener('error', (event) => {
             labels['error'] = 'LoadError';
         }
     }
+    if (event.error) {
```
akosyakov, line 33: we should replace cases when error is missing as well, we can add info about source and rest of metadata with generic error messages like `LoadError` or `UnknownError`. It will help us to track them down.

components/supervisor/frontend/src/index.ts
iQQBot: OK

components/supervisor/frontend/src/ide/ide-metrics-service-client.ts
iQQBot: google error reporting only supports the stack form of output log, it can automatically compatible with many language, on the other hand, the stack includes message line number and other information, there is no need to repeat the collection

Since this error comes from window.addListener, so this is probably a standard error

components/supervisor/frontend/src/index.ts
iQQBot: will try later

components/supervisor/frontend/src/ide/ide-metrics-service-client.ts
iQQBot: I think no, stack include message

iQQBot:
```
a.stack

'Error: this is a message\n' +
  '    at REPL2:1:5\n' +
  '    at Script.runInThisContext (node:vm:129:12)\n' +
  '    at REPLServer.defaultEval (node:repl:566:29)\n' +
  '    at bound (node:domain:421:15)\n' +
  '    at REPLServer.runBound [as eval] (node:domain:432:12)\n' +
  '    at REPLServer.onLine (node:repl:893:10)\n' +
  '    at REPLServer.emit (node:events:539:35)\n' +
  '    at REPLServer.emit (node:domain:475:12)\n' +
  '    at REPLServer.[_onLine] [as _onLine] (node:internal/readline/interface:418:12)\n' +
  '    at REPLServer.[_line] [as _line] (node:internal/readline/interface:879:18)'
> 
```

akosyakov: I don't think browser guruantee that `stack` is always present and that `message` is first line. I think we must collect both. They can be repetition, but it won't be good if we don't collect or if message is different.

akosyakov: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Stack

iQQBot: Then add a message attribute to the proto, and we watch for a while to see if this attribute is necessary (if they are always equal)

components/ide-metrics-api/idemetrics.proto
```diff
@@ -56,6 +56,8 @@ message ReportErrorRequest {
     string workspace_id = 5;
     string instance_id = 6;
     map<string, string> properties = 7;
+    string name = 8;
```
akosyakov, line 59: I though we will just use properties? In this case should we change it to `error_name` or `error_message? the trouble here that we also getting very specific to JS in Java and golang it would be different?

components/supervisor/frontend/src/index.ts
```diff
@@ -30,6 +31,15 @@ window.addEventListener('error', (event) => {
...
+        IDEMetricsServiceClient.reportError(event.error).catch(() => {});
+    } else if (labels['error'] == 'LoadError') {
+        let error = new Error("LoadError")
```
akosyakov, line 37: Each event should end up with error, otherwise we can miss something like script was loaded but failed on execution.

iQQBot: > script was loaded but failed on execution.

So that their `event.error` is not null and we can catch it above, we have nothing to report for types that neither have `event.error` nor are LoadError

akosyakov: How do you know it there is nothing? Why when it is called? At least we should report that it is happening. It sounds more like we don't understand args. 

As far as I understand it can be `message: string, filename: string, line: number, column?: number, e?: any`?  so we need to check if first argument is string, then we should compose error structure from such, if everything else fails then at least report `UnknownErrror`

akosyakov: We should not swallow anything.

akosyakov (approved):
After discussions with @iQQBot and @mustard-mh I think the code looks good. I have not tried. @mustard-mh Could you help to verify? 🙏 

/hold

We agreed to ad reporting for unhandled errors and console.error as well. But it could be 2 separate PRs afterwards.

components/supervisor/frontend/src/ide/ide-metrics-service-client.ts
```diff
@@ -40,4 +60,41 @@ export class IDEMetricsServiceClient {
...
+        p.error_message = error.message
+
+        const url = `${MetricsUrl}/reportError`
+        const params: ReportErrorParam = {
+            errorStack: error.stack ?? String(error),
+            component: "supervisor-frontend",
+            version: commit,
+            workspaceId: this.workspaceId ?? "",
+            instanceId: this.instanceId ?? "",
+            userId: this.userId ?? "",
```
mustard-mh, line 76: `?? ""` They are duplicated with lines (getter) above

mustard-mh:
Tested with workspace restart

| WS | Restart|
|-|-|
|image|image|

mustard-mh:
Could you change commit message, typos `frontent` 😄

mustard-mh (approved):
Nice

PR #12864, opened 2022-09-12
[UBP] Add `getStripePortalUrl` method to server

andrew-farries:
## Description

Add a new `getStripePortalUrl` method to `server`'s JSON-RPC API.

The method generates a Stripe portal URL that is used during the billing signup flow for both teams and individual users - it replaces the existing `getStripePortalUrlForTeam` method which only worked for team signup.

## Related Issue(s)

## How to test

1. Create a team called `Gitpod-xxx`
2. Sign the team up for UBP.
3. Create a new Stripe customer for your individual user:
```js
await window._gp.gitpodService.server.createOrUpdateStripeCustomerForUser("EUR")
```
3. Get the portal URL for the user:
```js
await window._gp.gitpodService.server.getStripePortalUrl("user:<your user id>")
```

Using the returned URL should take you to some Stripe UI.

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/gitpod-protocol/src/gitpod-service.ts
+++ b/components/gitpod-protocol/src/gitpod-service.ts
@@ -295,6 +295,7 @@ export interface GitpodServer extends JsonRpcServer<GitpodClient>, AdminServer, 
     createOrUpdateStripeCustomerForTeam(teamId: string, currency: string): Promise<void>;
     createOrUpdateStripeCustomerForUser(currency: string): Promise<void>;
     subscribeTeamToStripe(teamId: string, setupIntentId: string): Promise<void>;
+    getStripePortalUrl(attributionId: string): Promise<string>;
     getStripePortalUrlForTeam(teamId: string): Promise<string>;
     getUsageLimitForTeam(teamId: string): Promise<number | undefined>;
     setUsageLimitForTeam(teamId: string, spendingLimit: number): Promise<void>;
--- a/components/server/ee/src/user/stripe-service.ts
+++ b/components/server/ee/src/user/stripe-service.ts
@@ -130,6 +130,18 @@ export class StripeService {
         return session.url;
     }
 
+    async getPortalUrlForUser(user: User): Promise<string> {
+        const customer = await this.findCustomerByUserId(user.id);
+        if (!customer) {
+            throw new Error(`No Stripe Customer ID found for user '${user.id}'`);
+        }
+        const session = await this.getStripe().billingPortal.sessions.create({
+            customer: customer.id,
+            return_url: this.config.hostUrl.with(() => ({ pathname: `/billing` })).toString(),
+        });
+        return session.url;
+    }
+
     async findUncancelledSubscriptionByCustomer(customerId: string): Promise<Stripe.Subscription | undefined> {
         const result = await this.getStripe().subscriptions.list({
             customer: customerId,
--- a/components/server/ee/src/workspace/gitpod-server-impl.ts
+++ b/components/server/ee/src/workspace/gitpod-server-impl.ts
@@ -2150,6 +2150,43 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
         }
     }
 
+    async getStripePortalUrl(ctx: TraceContext, attributionId: string): Promise<string> {
+        const attrId = AttributionId.parse(attributionId);
+        if (attrId === undefined) {
+            log.error(`Invalid attribution id: ${attributionId}`);
+            throw new ResponseError(ErrorCodes.BAD_REQUEST, `Invalid attibution id: ${attributionId}`);
+        }
+
+        const user = this.checkAndBlockUser("getStripePortalUrl");
+
+        let url: string;
+        if (attrId.kind === "user") {
+            await this.ensureStripeApiIsAllowed({ user });
+            try {
+                url = await this.stripeService.getPortalUrlForUser(user);
+            } catch (error) {
+                log.error(`Failed to get Stripe portal URL for user '${user.id}'`, error);
+                throw new ResponseError(
+                    ErrorCodes.INTERNAL_SERVER_ERROR,
+                    `Failed to get Stripe portal URL for user '${user.id}'`,
+                );
+            }
+        } else {
+            const team = await this.guardTeamOperation(attrId.teamId, "update");
+            await this.ensureStripeApiIsAllowed({ team });
+            try {
+                url = await this.stripeService.getPortalUrlForTeam(team);
+            } catch (error) {
+                log.error(`Failed to get Stripe portal URL for team '${team.id}'`, error);
+                throw new ResponseError(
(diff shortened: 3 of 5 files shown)


components/gitpod-protocol/src/gitpod-service.ts
```diff
@@ -294,6 +294,7 @@ export interface GitpodServer extends JsonRpcServer<GitpodClient>, AdminServer,
     createOrUpdateStripeCustomerForTeam(teamId: string, currency: string): Promise<void>;
     createOrUpdateStripeCustomerForUser(currency: string): Promise<void>;
     subscribeTeamToStripe(teamId: string, setupIntentId: string): Promise<void>;
+    getStripePortalUrl(attributionId: string): Promise<string>;
```
jankeromnes, line 297: Why make this name ambiguous?

```suggestion
    getStripePortalUrlForUser(attributionId: string): Promise<string>;
```

components/server/src/auth/rate-limiter.ts
```diff
@@ -210,6 +210,7 @@ const defaultFunctions: FunctionsConfig = {
     createOrUpdateStripeCustomerForTeam: { group: "default", points: 1 },
     createOrUpdateStripeCustomerForUser: { group: "default", points: 1 },
     subscribeTeamToStripe: { group: "default", points: 1 },
+    getStripePortalUrl: { group: "default", points: 1 },
```
jankeromnes, line 213:
```suggestion
    getStripePortalUrlForUser: { group: "default", points: 1 },
```

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -3233,6 +3233,9 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
     async getStripePortalUrlForTeam(ctx: TraceContext, teamId: string): Promise<string> {
         throw new ResponseError(ErrorCodes.SAAS_FEATURE, `Not implemented in this version`);
     }
+    async getStripePortalUrl(ctx: TraceContext, attributionId: string): Promise<string> {
```
jankeromnes, line 3236: Also, tiny problem, the order is wrong 😇

```suggestion
    async getStripePortalUrlForUser(ctx: TraceContext, attributionId: string): Promise<string> {
```

components/gitpod-protocol/src/gitpod-service.ts
jankeromnes: Ah sorry, just saw the `attributionId`

components/server/src/workspace/gitpod-server-impl.ts
andrew-farries: > Also, tiny problem, the order is wrong

I'm not sure what you mean by this 😕

jankeromnes: Sorry for the noise -- I was mistaken, which is why I immediately resolved this conversation. 🙈

components/server/ee/src/user/stripe-service.ts
```diff
@@ -130,6 +130,18 @@ export class StripeService {
...
     }
 
+    async getPortalUrlForUser(user: User): Promise<string> {
+        const customer = await this.findCustomerByUserId(user.id);
+        if (!customer) {
+            throw new Error(`No Stripe Customer ID found for user '${user.id}'`);
+        }
+        const session = await this.getStripe().billingPortal.sessions.create({
+            customer: customer.id,
+            return_url: this.config.hostUrl.with(() => ({ pathname: `/t/${user.id}/billing` })).toString(),
```
jankeromnes, line 140: This `return_url` won't work. Isn't it simply `/billing`?

```suggestion
            return_url: this.config.hostUrl.with(() => ({ pathname: "/billing" })).toString(),
```

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2130,20 +2130,46 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
-            return url;
-        } catch (error) {
-            log.error(`Failed to get Stripe portal URL for team '${teamId}'`, error);
-            throw new ResponseError(
-                ErrorCodes.INTERNAL_SERVER_ERROR,
-                `Failed to get Stripe portal URL for team '${teamId}'`,
-            );
+    async getStripePortalUrl(ctx: TraceContext, attributionId: string): Promise<string> {
+        const attrId = AttributionId.parse(attributionId);
+        if (attrId === undefined) {
```
jankeromnes, line 2135: I'd make this check more "paranoid" (e.g. we wouldn't want any `null`, `NaN`, `false`, etc here)

```suggestion
        if (!attrId) {
```

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2130,20 +2130,46 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
+            log.error(`Invalid attribution id: ${attributionId}`);
+            throw new ResponseError(ErrorCodes.BAD_REQUEST, `Invalid attibution id: ${attributionId}`);
+        }
+
+        const user = this.checkAndBlockUser("getStripePortalUrl");
```
jankeromnes, line 2140: Not really important, but shouldn't the block check be the very first call in each server method? 💭

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2130,20 +2130,46 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
+                    `Failed to get Stripe portal URL for team '${team.id}'`,
+                );
+            }
         }
+        return url;
+    }
+
+    async getStripePortalUrlForTeam(ctx: TraceContext, teamId: string): Promise<string> {
+        const attributionId: AttributionId = { kind: "team", teamId: teamId };
+        return this.getStripePortalUrl(ctx, AttributionId.render(attributionId));
```
jankeromnes, line 2172: Side note: You could easily remove this now-somewhat-useless method by updating the single call-site in the dashboard.

jankeromnes (approved):
Many thanks for the helpful test instructions! 🙏

This change looks good and works as advertized, except for the `return_url` (please see inline). The other comments are optional.

/hold

components/server/ee/src/workspace/gitpod-server-impl.ts
andrew-farries: thanks, moved

components/server/ee/src/workspace/gitpod-server-impl.ts
andrew-farries: I'll move the dashboard over to the new more general methods in a separate PR later.

Issue #12580, opened 2022-09-01
Better surface Stripe sign-up errors in the UI


jankeromnes:
### Bug description

Follow-up to https://github.com/gitpod-io/gitpod/issues/12343 / https://github.com/gitpod-io/gitpod/pull/12376

When testing the PR, we noticed two separate UX problems:

1. When you sign up with a given currency, then cancel, then try to sign up with a different currency, Stripe fails with this error:

```
Error: You cannot combine currencies on a single customer. This customer has had a subscription, coupon, or invoice item with currency eur
```

It would be cool to handle this problem specifically, e.g. by preventing existing customers from selecting a different currency, and/or by suggesting that they reach out to Gitpod's support (so that we can de-activate the Stripe customer account, allowing the user to create a new one in a different currency).

2. In general, when an error happens on Stripe sign-up, it is not properly shown in the UI, and is currently only visible in the browser console & in the server logs. It would be nice to show visually in the UI that something went wrong and what can be done about it.

### Steps to reproduce

1. Sign up a team up to usage-based in one currency
2. Cancel
3. Try to sign up the same team again but in a different currency (it fails, but shows no error in the UI)

### Expected behavior

It should not be possible to try and sign in again with a different currency. Instead, we could prevent the selection of a different currency, and/or suggest that customers reach out to support if they want to switch to a different currency.

PR #12872, opened 2022-09-12
[dashboard] Show error messages in the UI when upgrading with Stripe fails

jankeromnes:
## Description

Show error messages in the UI when upgrading with Stripe fails for any reason.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/12580

## How to test

1. Create a team called "Gitpod [Something]"
2. Try to upgrade using an invalid credit card number, or one of Stripe's [failing test cards](https://stripe.com/docs/testing#declined-payments) -- this should always show an error message
3. Upgrade using a regular test card in the currency **USD** (e.g. 4242 4242 4242 4242, 4/24, 424) -- this should work
4. Once it went through, manage and cancel your Stripe subscription
5. Upgrade again with a regular test card, but this time with the currency **EUR**
6. The payment method should go through, but on the next page you should see an error message

(Note: Currently that second step error message is redacted, and will just say that "upgrading failed" while the actual error message is hidden in `server` logs -- we can improve this UX in a follow-up step.)

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/dashboard/src/components/UsageBasedBillingConfig.tsx
+++ b/components/dashboard/src/components/UsageBasedBillingConfig.tsx
@@ -13,6 +13,7 @@ import { PaymentContext } from "../payment-context";
 import { getGitpodService } from "../service/service";
 import DropDown from "../components/DropDown";
 import Modal from "../components/Modal";
+import Alert from "./Alert";
 
 interface Props {
     userOrTeamId: string;
@@ -150,12 +151,14 @@ function CreditCardInputForm(props: { id: string }) {
     const elements = useElements();
     const { currency, setCurrency } = useContext(PaymentContext);
     const [isLoading, setIsLoading] = useState<boolean>(false);
+    const [billingError, setBillingError] = useState<string | undefined>();
 
     const handleSubmit = async (event: React.FormEvent) => {
         event.preventDefault();
         if (!stripe || !elements) {
             return;
         }
+        setBillingError(undefined);
         setIsLoading(true);
         try {
             // Create Stripe customer for team & currency (or update currency)
@@ -176,7 +179,7 @@ function CreditCardInputForm(props: { id: string }) {
             }
         } catch (error) {
             console.error(error);
-            alert(error?.message || "Failed to submit form. See console for error message.");
+            setBillingError(`Failed to submit form. ${error?.message || String(error)}`);
         } finally {
             setIsLoading(false);
         }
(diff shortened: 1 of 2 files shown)


components/dashboard/src/components/UsageBasedBillingConfig.tsx
```diff
@@ -176,14 +179,19 @@ function CreditCardInputForm(props: { id: string }) {
             }
         } catch (error) {
             console.error(error);
-            alert(error?.message || "Failed to submit form. See console for error message.");
+            setBillingError(error?.message || "Failed to submit form. See console for error message.");
```
andrew-farries, line 182: I get that there might be some extra context in the the console that might be helpful, but I think it's pretty unusual for a production site to refer a user to the developer console for further information.

Would it be better to remove that sentence and add the extra information to the UI in a subsequent PR?

components/dashboard/src/teams/TeamUsageBasedBilling.tsx
```diff
@@ -85,6 +87,7 @@ export default function TeamUsageBasedBilling() {
                 window.localStorage.removeItem(`pendingStripeSubscriptionForTeam${team.id}`);
                 clearTimeout(pollStripeSubscriptionTimeout!);
                 setPendingStripeSubscription(undefined);
+                setBillingError(error.message || "Could not subscribe team to Stripe. See console for error messages.");
```
andrew-farries, line 90: Same comment about referring users to the developer console.

andrew-farries (approved):
/hold for question about the error messages.

components/dashboard/src/components/UsageBasedBillingConfig.tsx
jankeromnes: Thanks for raising this! It's boilerplate code I kept from a Stripe snippet, but you're right about referring to the console. Instead, how about always stringifying the error? I think this works equally well with or without `error.message`:

```suggestion
            setBillingError(String(error));
```

components/dashboard/src/teams/TeamUsageBasedBilling.tsx
jankeromnes: Alternative proposal:

```suggestion
                setBillingError(`Could not subscribe team to Stripe. ${String(error)}`);
```

components/dashboard/src/components/UsageBasedBillingConfig.tsx
andrew-farries: Yes, that looks like it will return the `.message` in the case where the error has one [[1](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/toString)]

I think that's preferable to referring users to the console.

jankeromnes:
Many thanks for the reviews! 🙏

Nits addressed, confirmed still working as intended. Releasing the breaks.

/unhold

PR #12876, opened 2022-09-12
[UBP] Add `findStripeSubscriptionId` method to server

andrew-farries:
## Description

Add a general method `findStripeSubscriptionId` for finding the Stripe subscription id for a given attribution id. 

Implement the existing `findStripeSubscriptionIdForTeam` method in terms of the new method.

The more general method will be used to look up subscription ids for individual users in subsequent PRs.

## Related Issue(s)

## How to test

The flow to sign a team up for UBP still works:

1. Create a team with `Gitpod` in the name.
2. Sign the team up for UBP from the billing tab on the team settings page.

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview
- [ ] /werft with-payment

### DIFF
--- a/components/server/ee/src/workspace/gitpod-server-impl.ts
+++ b/components/server/ee/src/workspace/gitpod-server-impl.ts
@@ -114,6 +114,7 @@ import { BillingMode } from "@gitpod/gitpod-protocol/lib/billing-mode";
 import { BillingModes } from "../billing/billing-mode";
 import { getExperimentsClientForBackend } from "@gitpod/gitpod-protocol/lib/experiments/configcat-server";
 import { BillingService } from "../billing/billing-service";
+import Stripe from "stripe";
 
 @injectable()
 export class GitpodServerEEImpl extends GitpodServerImpl {
@@ -2047,23 +2048,42 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
         }
     }
 
-    async findStripeSubscriptionIdForTeam(ctx: TraceContext, teamId: string): Promise<string | undefined> {
-        this.checkAndBlockUser("findStripeSubscriptionIdForTeam");
-        await this.guardTeamOperation(teamId, "get");
+    async findStripeSubscriptionId(ctx: TraceContext, attributionId: string): Promise<string | undefined> {
+        const attrId = AttributionId.parse(attributionId);
+        if (attrId === undefined) {
+            log.error(`Invalid attribution id: ${attributionId}`);
+            throw new ResponseError(ErrorCodes.BAD_REQUEST, `Invalid attibution id: ${attributionId}`);
+        }
+
+        this.checkAndBlockUser("findStripeSubscriptionId");
+
+        let customer: Stripe.Customer | undefined;
         try {
-            const customer = await this.stripeService.findCustomerByTeamId(teamId);
+            if (attrId.kind == "team") {
+                await this.guardTeamOperation(attrId.teamId, "get");
+                customer = await this.stripeService.findCustomerByTeamId(attrId.teamId);
+            } else {
+                customer = await this.stripeService.findCustomerByUserId(attrId.userId);
+            }
+
             if (!customer?.id) {
                 return undefined;
             }
+
             const subscription = await this.stripeService.findUncancelledSubscriptionByCustomer(customer.id);
             return subscription?.id;
         } catch (error) {
-            log.error(`Failed to get Stripe Subscription ID for team '${teamId}'`, error);
+            log.error(`Failed to get Stripe Subscription ID for '${attributionId}'`, error);
             throw new ResponseError(
                 ErrorCodes.INTERNAL_SERVER_ERROR,
-                `Failed to get Stripe Subscription ID for team '${teamId}'`,
+                `Failed to get Stripe Subscription ID for '${attributionId}'`,
             );
         }
+    }
+
+    async findStripeSubscriptionIdForTeam(ctx: TraceContext, teamId: string): Promise<string | undefined> {
+        const attributionId: AttributionId = { kind: "team", teamId: teamId };
+        return this.findStripeSubscriptionId(ctx, AttributionId.render(attributionId));
     }
 
     async createOrUpdateStripeCustomerForTeam(ctx: TraceContext, teamId: string, currency: string): Promise<void> {
--- a/components/gitpod-protocol/src/gitpod-service.ts
+++ b/components/gitpod-protocol/src/gitpod-service.ts
@@ -290,6 +290,7 @@ export interface GitpodServer extends JsonRpcServer<GitpodClient>, AdminServer, 
 
     getStripePublishableKey(): Promise<string>;
     getStripeSetupIntentClientSecret(): Promise<string>;
+    findStripeSubscriptionId(attributionId: string): Promise<string | undefined>;
     findStripeSubscriptionIdForTeam(teamId: string): Promise<string | undefined>;
     createOrUpdateStripeCustomerForTeam(teamId: string, currency: string): Promise<void>;
     createOrUpdateStripeCustomerForUser(currency: string): Promise<void>;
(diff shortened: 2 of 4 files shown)


components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2047,25 +2048,44 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
+
+        this.checkAndBlockUser("findStripeSubscriptionId");
+
+        let customer: Stripe.Customer | undefined;
         try {
-            const customer = await this.stripeService.findCustomerByTeamId(teamId);
+            if (attrId.kind == "team") {
+                await this.guardTeamOperation(attrId.teamId, "get");
+                customer = await this.stripeService.findCustomerByTeamId(attrId.teamId);
+            } else {
```
easyCZ, line 2065: Perhaps worth being explcit about `user` in case we add more types of attribution ids.

easyCZ (approved):
We're adding quite a bit of Stripe interaction to server now. The original intent was to keep as much of it on `usage` as possible. Have we lifted that goal?

andrew-farries:
> We're adding quite a bit of Stripe interaction to server now. The original intent was to keep as much of it on usage as possible. Have we lifted that goal?

It's a good point - the API to interact with Stripe for the subscription creation flow was originally implemented on `server` and the goal of https://github.com/gitpod-io/gitpod/issues/12685 is to generalise that API to allow it work for individual users. I didn't really consider moving the API elsewhere as part of that tbh.

components/server/ee/src/workspace/gitpod-server-impl.ts
easyCZ: Would it? Here we just do a check against `== "team"` and then do an `else` for anything else. I think it wouldn't fail to compile in that case.

andrew-farries: If I extend the `AttributionId` union with a new `"foo"` type as below:

<summary>
<details>

```ts
export type AttributionId = UserAttributionId | TeamAttributionId | FooAttributionId;
export type AttributionTarget = "user" | "team" | "foo";

export interface UserAttributionId {
    kind: "user";
    userId: string;
}
export interface TeamAttributionId {
    kind: "team";
    teamId: string;
}

export interface FooAttributionId {
    kind: "foo";
    someOtherId: string;
}

export namespace AttributionId {
    const SEPARATOR = ":";

    export function parse(s: string): UserAttributionId | TeamAttributionId | FooAttributionId | undefined {
        if (!s) {
            return undefined;
        }
        const parts = s.split(":");
        if (parts.length !== 2) {
            return undefined;
        }
        switch (parts[0]) {
            case "user":
                return { kind: "user", userId: parts[1] };
            case "team":
                return { kind: "team", teamId: parts[1] };
            case "foo":
                return { kind: "foo", someOtherId: parts[1] };
            default:
                return undefined;
        }
    }

    export function render(id: AttributionId): string {
        switch (id.kind) {
            case "user":
                return `user${SEPARATOR}${id.userId}`;
            case "team":
                return `team${SEPARATOR}${id.teamId}`;
            case "foo":
                return `foo{SEPARATOR}${id.someOtherId}`;
        }
    }
}
```
</details>
</summary>

I get a type error in the `else` block here:

https://github.com/gitpod-io/gitpod/blob/015987aa4288fe23ff2fc846209da6671b06764d/components/server/ee/src/workspace/gitpod-server-impl.ts#L2066

because the type of `attrId` in the `else` block is now `UserAttributionId | FooAttributionId` rather than `UserAttributionId` and so can't be guaranteed to have a `userId` field.

andrew-farries: This is quite a neat feature of typescript's discriminated unions:

https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions

easyCZ: When I do this in a [playground](https://www.typescriptlang.org/play?#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCbvCwDKgU8JkY7p7ehCTkVDQS2E6gnPCxGcwmZqDSTgzxxWWVoASMFmgYkAAeRJTInN3ymj4d-jSCeNsMq-wuoPaOltigAKoASgAywhK7SbGQZIIz5VWCFzSeCrZagNYbChbHaxUDcCjJZLfSDbExIAgUdxkUBIursJzCFJtXydajBRSodqVVCoSjSdj1RoASVcAF4HoJvhUaXSGQ1OCzQAAfbweRDc2kUemMgWuEUAb1A8Qy3HyoAARMl4PB1aAAL4AbmCGQUZGSDCW905ZAlvJlgvlihKJWVPDV6vY1vVRudoE93xZasEks4zCNesUJu+5qWODFtqlfOZrkdvtdqo1CgYiG9TudWcQgauIbDigjimSXFgMpx2oAFHVaWqE9L+SyAJSgVMiVINnkBdOgVnDzNi9Wd7slBCcQTwdwBaDwZh19UF8d5vXg6Ccrt5qeNWfzxfLj0JAYSTjrkoR8sV+uK9PurU6-XtxRAA), it compiles

andrew-farries: But if you try to do anything with the value of `attr` inside your `else` block [like access `userId`](https://www.typescriptlang.org/play?#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCbvCwDKgU8JkY7p7ehCTkVDQS2E6gnPCxGcwmZqDSTgzxxWWVoASMFmgYkAAeRJTInN3ymj4d-jSCeNsMq-wuoPaOltigAKoASgAywhK7SbGQZIIz5VWCFzSeCrZagNYbChbHaxUDcCjJZLfSDbExIAgUdxkUBIursJzCFJtXydajBRSodqVVCoSjSdj1RoASVcAF4HoJvhUaXSGQ1OCzQAAfbweRDc2kUemMgWuEUAb1A8Qy3HyoAARMl4PB1aAAL4AbmCGQUZGSDCW905ZAlvJlgvlihKJWVPDV6vY1vVRudoE93xZasEks4zCNesUJu+5qWODFtqlfOZrkdvtdqo1CgYiG9TudWcQgauIbDigjimSXFgMpx2oAFHVaWqE9L+SyAJSgVMiVINnkBdOgVnDzNi9Wd7slBCcQTwdwBaDwZh92kBAsdvN68HQTldvNTxqz+eL5eNsgBf1kDclCPliv1xXp91anX69uKIA), you do get a type error because the type has not been sufficiently narrowed inside the `else` (it could still be either a `foo` or a `user`).

easyCZ: Sure, but that relies on the fact that we wouldn't overload the `userId` or another property. If my new type was for a `{ kind: "superuser", userId: "..."}` then it will compile, but won't work as expected.

andrew-farries: yeah, that's true. I assume that if we added a new attribution `kind` it would have a non-overlapping shape.

PR #12882, opened 2022-09-12
[usage] API for setting billingStrategy

svenefftinge:
## Related Issue(s)

Fixes #

## Release Notes

```release-note
```

## Werft options:

- [x] /werft with-preview


roboquat:

<details>

</details>

components/usage-api/usage/v1/usage.proto
```diff
@@ -106,12 +117,13 @@ message GetCostCenterResponse {
     CostCenter cost_center = 1;
 }
 
+enum BillingStrategy {
+    BILLING_STRATEGY_STRIPE = 0;
+    BILLING_STRATEGY_OTHER = 1;
+}
```
easyCZ, line 123: Because all fields in a protobuf message are optional, and protobuf defaults to the empty value, for Enums it will default to `BILLING_STRATEGY_STRIPE` in this struct, when not set. That's not always desirable, as if you forget to set it it will have adverse effects. For that reason, the UNKNOWN value is normally used in any enum (the RPC is then implemented to fail if the value is UNKNOWN.

```suggestion
enum BillingStrategy {
    BILLING_STRATEGY_UNKNOWN = 0
    BILLING_STRATEGY_STRIPE = 1;
    BILLING_STRATEGY_OTHER = 2;
}
```

components/usage/pkg/apiv1/usage.go
```diff
@@ -186,30 +186,52 @@ func (s *UsageService) GetCostCenter(ctx context.Context, in *v1.GetCostCenterRe
...
+		return nil, status.Errorf(codes.InvalidArgument, "The spending limit must be higher than the current usage %u", summary.CreditCentsBalanceAtEnd.ToCredits())
+	}
+
+	//TODO (se) - Check if the strategy has changed and was 'other' before. If so run the invoice finalization.
 	_, err = db.SaveCostCenter(ctx, s.conn, &db.CostCenter{
 		ID:              attributionID,
-		SpendingLimit:   in.CostCenter.SpendingLimit,
+		SpendingLimit:   in.SpendingLimit,
 		BillingStrategy: billingStrategy,
 	})
```
easyCZ, line 214: There's a possible race here due to write after read semantics, but I think in our case we don't actually need to handle it as we'll recoup it in the next billing cycle.

components/server/ee/src/billing/billing-service.ts
```diff
@@ -43,8 +44,8 @@ export class BillingService {
         }
 
         const upcomingInvoice = await this.getUpcomingInvoice(attributionId);
-        const currentInvoiceCredits = upcomingInvoice.getCredits();
-        if (currentInvoiceCredits >= costCenter.spendingLimit) {
+        const currentInvoiceCredits = upcomingInvoice.credits | 0;
+        if (currentInvoiceCredits >= (costCenter.spendingLimit || 0)) {
```
andrew-farries, line 48: rather than `|| 0` should we use the same `defaultSpendingLimit` as when the Stripe subscription is configured?

See [this comment](https://github.com/gitpod-io/gitpod/pull/12887#discussion_r968555099).

PR #12887, opened 2022-09-12
[UBP] Add `subscribeToStripe` server method

andrew-farries:
## Description

Add a general method `subscribeToStripe` to the server JSON-RPC API for  adding Stripe subscriptions to a Stripe customer, for customers representing teams or individual users.

Implement the existing `subscribeTeamToStripe` method  (which only worked for teams) in terms of the new method.

The more general method will be used to subscribe individual Stripe users in subsequent PRs.

## Related Issue(s)

## How to test

The flow to sign a team up for UBP still works:

1. Create a team with `Gitpod` in the name.
2. Sign the team up for UBP from the billing tab on the team settings page.

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/server/ee/src/workspace/gitpod-server-impl.ts
+++ b/components/server/ee/src/workspace/gitpod-server-impl.ts
@@ -2125,7 +2125,7 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
 
     protected defaultSpendingLimit = 100;
     async subscribeTeamToStripe(ctx: TraceContext, teamId: string, setupIntentId: string): Promise<void> {
-        this.checkAndBlockUser("subscribeUserToStripe");
+        this.checkAndBlockUser("subscribeTeamToStripe");
         const team = await this.guardTeamOperation(teamId, "update");
         await this.ensureStripeApiIsAllowed({ team });
         try {
--- a/components/server/ee/src/workspace/gitpod-server-impl.ts
+++ b/components/server/ee/src/workspace/gitpod-server-impl.ts
@@ -2124,30 +2124,48 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
     }
 
     protected defaultSpendingLimit = 100;
-    async subscribeTeamToStripe(ctx: TraceContext, teamId: string, setupIntentId: string): Promise<void> {
-        this.checkAndBlockUser("subscribeTeamToStripe");
-        const team = await this.guardTeamOperation(teamId, "update");
-        await this.ensureStripeApiIsAllowed({ team });
+    async subscribeToStripe(ctx: TraceContext, attributionId: string, setupIntentId: string): Promise<void> {
+        const attrId = AttributionId.parse(attributionId);
+        if (attrId === undefined) {
+            log.error(`Invalid attribution id: ${attributionId}`);
+            throw new ResponseError(ErrorCodes.BAD_REQUEST, `Invalid attibution id: ${attributionId}`);
+        }
+
+        const user = this.checkAndBlockUser("subscribeToStripe");
+
+        let customer: Stripe.Customer | undefined;
         try {
-            let customer = await this.stripeService.findCustomerByTeamId(team!.id);
-            if (!customer) {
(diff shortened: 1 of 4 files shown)


components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2123,33 +2123,51 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
+                throw new Error(`No Stripe customer profile for '${attributionId}'`);
             }
+
             await this.stripeService.setDefaultPaymentMethodForCustomer(customer, setupIntentId);
             await this.stripeService.createSubscriptionForCustomer(customer);
-
-            const attributionId: AttributionId = { kind: "team", teamId };
-
-            // Creating a cost center for this team
             await this.usageServiceClientProvider.getDefault().setCostCenter({
```
easyCZ, line 2151: Out of curiosity, how would we handle the case where we do setup the user in stripe but fail to set the cost center due to downstream issues?

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2123,33 +2123,51 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
-            throw new ResponseError(ErrorCodes.INTERNAL_SERVER_ERROR, `Failed to subscribe team '${teamId}' to Stripe`);
+            log.error(`Failed to subscribe '${attributionId}' to Stripe`, error);
+            throw new ResponseError(
+                ErrorCodes.INTERNAL_SERVER_ERROR,
+                `Failed to subscribe '${attributionId}' to Stripe`,
+            );
         }
     }
 
+    protected defaultSpendingLimit = 100;
```
easyCZ, line 2165: Is this still true? https://github.com/gitpod-io/gitpod/pull/12882/files#diff-0caf79924a209df9577c140a9aefa82d74dab19ed1d20bd1f8597259c3d83316R48 seems to be using 0 as a value as well.

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2123,33 +2123,51 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
+                const team = await this.guardTeamOperation(attrId.teamId, "update");
+                await this.ensureStripeApiIsAllowed({ team });
+                customer = await this.stripeService.findCustomerByTeamId(team!.id);
+            } else {
+                await this.ensureStripeApiIsAllowed({ user });
+                customer = await this.stripeService.findCustomerByUserId(user.id);
+            }
             if (!customer) {
-                throw new Error(`No Stripe customer profile for team '${team.id}'`);
+                throw new Error(`No Stripe customer profile for '${attributionId}'`);
```
easyCZ, line 2146:
```suggestion
                throw new Error(`Could not find Stripe customer profile for '${attributionId}'. Please reach out to our support team.`);
```
Just feels we can be friendlier when this does blow up.

easyCZ (approved):
LGTM. Left a couple of comments.

/hold if you want to tackle the comments

components/server/ee/src/workspace/gitpod-server-impl.ts
andrew-farries: have left a comment on that PR.

components/server/ee/src/workspace/gitpod-server-impl.ts
andrew-farries: Currently, I don't think we do handle it. I think we'd end up with a valid Stripe customer for which we aren't able to look up cost center information.

components/server/ee/src/workspace/gitpod-server-impl.ts
svenefftinge: This is an ok default spending limit for stripe-backed costCenters (ok, because it would be better to propose something reasonable based on past usage and/or number of users in a team).
For general costCenters we should assume 0, or better it should never be undefined on cost centers.

Issue #11397, opened 2022-07-15
workspaces which cannot be scheduled to a cluster get stuck in Pending


kylos101:
### Bug description

Workspaces which cannot be scheduled to a cluster get stuck in Pending.

### Steps to reproduce

In a preview environment, `kubectl cordon` the node. Try starting a workspace. It'll stay stuck in Pending for-ev-er!

### Workspace affected

n/a

### Expected behavior

The workspace should fail, and land in a permanent Failed phase

### Example repository

n/a

### Anything else?

This becomes a problem when you're trying to delete a cluster, because `ws-manager-bridge` reports that there are still active workspaces in the database for the cluster, even though there are no workspace pods in the governed workspace cluster.

This prevented the delete of `us53` (which was empty, and had no workspace clusters), until we ran the following queries:

```sql
-- find pending workspaces that aren't really pending
SELECT dbwi.id
FROM gitpod.d_b_workspace dbw  
inner join gitpod.d_b_workspace_instance dbwi 
on dbw.id = dbwi.workspaceId  
WHERE 
dbwi.phasePersisted not in ('stopped','failed') 
# change to match a region you're interested in
and dbwi.region = 'us53'
ORDER by dbwi.creationTime asc
;

-- force stop the pending workspaces
UPDATE gitpod.d_b_workspace_instance AS wsi
SET
			wsi.stoppedTime = IF(wsi.startedTime = '', wsi.creationTime, wsi.startedTime),
			wsi.stoppingTime = IF(wsi.startedTime = '', wsi.creationTime, wsi.startedTime),
			wsi.phasePersisted = 'stopped',			
			wsi.status = '{"phase": "stopped", "conditions": {}}'
WHERE wsi.id IN ('fb52614b-927b-4412-b583-dd3cc7bc3087',
'04eedb5b-1f35-4e4b-a8b3-56ea13c0f2dd',
'9b3d3928-59dd-46f0-9876-cc2bab905a2e',
'261bbde5-100f-4a19-8085-125ef82a0f0a',
'7878cc2a-544e-4f24-82ee-62a24c4f9c34',
'5dda17bf-1fe9-4cc6-b5ae-82aa6603049c',
'ad40bce6-a74c-47a9-b9a4-d9490afbfd41',
'028d826e-49b2-45ce-b829-8b4f693e93d3',
'99b74f2b-0a49-45c6-9c59-c4339a6dcb48',
'3695365a-0b4c-4df2-96f3-700598fe3baf')
[...]

PR #12902, opened 2022-09-13
[bridge] Logs stuck workspace instances to validate fix

laushinka:
## Description
This PR checks and logs if a workspace instance that the ws-manager does not know about is in `pending`, `creating`, `initializing` states, or if it is in `stopping` for longer than 10 seconds (a guard in case of a race condition).
We will monitor the logs and validate if we can deploy the planned fix [here](https://github.com/gitpod-io/gitpod/compare/main...lau/stuck-ws-11397).

## Related Issue(s)

Relates #11397
Relates #12955

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/ws-manager-bridge/src/bridge.ts
+++ b/components/ws-manager-bridge/src/bridge.ts
@@ -446,13 +446,13 @@ export class WorkspaceManagerBridge implements Disposable {
                     const installation = this.cluster.name;
                     log.debug("Controlling instances...", { installation });
 
-                    const runningInstances = await this.workspaceDB
+                    const nonStoppedInstances = await this.workspaceDB
                         .trace(ctx)
                         .findRunningInstancesWithWorkspaces(installation, undefined, true);
 
                     // Control running workspace instances against ws-manager
                     try {
-                        await this.controlRunningInstances(ctx, runningInstances, clientProvider);
+                        await this.controlNonStoppedWSManagerManagedInstances(ctx, nonStoppedInstances, clientProvider);
 
                         disconnectStarted = Number.MAX_SAFE_INTEGER; // Reset disconnect period
                     } catch (err) {
@@ -466,7 +466,7 @@ export class WorkspaceManagerBridge implements Disposable {
                     }
 
                     // Control workspace instances against timeouts
-                    await this.controlInstancesTimeouts(ctx, runningInstances);
+                    await this.controlInstancesTimeouts(ctx, nonStoppedInstances);
 
                     log.debug("Done controlling instances.", { installation });
                 } catch (err) {
@@ -485,17 +485,17 @@ export class WorkspaceManagerBridge implements Disposable {
      * This methods controls all instances that we have currently marked as "running" in the DB.
      * It checks whether they are still running with their respective ws-manager, and if not, marks them as stopped in the DB.
      */
-    protected async controlRunningInstances(
+    protected async controlNonStoppedWSManagerManagedInstances(
         parentCtx: TraceContext,
         runningInstances: RunningWorkspaceInfo[],
         clientProvider: ClientProvider,
     ) {
         const installation = this.config.installation;
 
-        const span = TraceContext.startSpan("controlRunningInstances", parentCtx);
+        const span = TraceContext.startSpan("controlNonStoppedWSManagerManagedInstances", parentCtx);
         const ctx = { span };
         try {
-            log.debug("Controlling running instances...", { installation });
+            log.debug("Controlling non-stopped instances that are managed by WS Manager...", { installation });
 
             const runningInstancesIdx = new Map<string, RunningWorkspaceInfo>();
             runningInstances.forEach((i) => runningInstancesIdx.set(i.latestInstance.id, i));
@@ -504,9 +504,27 @@ export class WorkspaceManagerBridge implements Disposable {
             const actuallyRunningInstances = await client.getWorkspaces(ctx, new GetWorkspacesRequest());
             actuallyRunningInstances.getStatusList().forEach((s) => runningInstancesIdx.delete(s.getId()));
 
+            // runningInstancesIdx only contains instances that ws-manager is not aware of
             for (const [instanceId, ri] of runningInstancesIdx.entries()) {
                 const instance = ri.latestInstance;
-                if (instance.status.phase !== "running") {
+                const phase = instance.status.phase;
+                if (phase !== "running") {
+                    // This below if block is to validate the planned fix
+                    if (
+                        phase === "pending" ||
+                        phase === "creating" ||
+                        phase === "initializing" ||
+                        (phase === "stopping" &&
+                            instance.stoppingTime &&
+                            durationLongerThanSeconds(Date.parse(instance.stoppingTime), 10))
+                    ) {
+                        log.info("Logging to validate #12902. Should mark as stopped in database.", {
+                            instanceId,
+                            workspaceId: instance.workspaceId,
+                            installation,
+                            phase,
+                        });
+                    }
                     log.debug({ instanceId }, "Skipping instance", {
                         phase: instance.status.phase,
                         creationTime: instance.creationTime,
@@ -516,9 +534,8 @@ export class WorkspaceManagerBridge implements Disposable {
                 }
 
                 log.info(
-                    { instanceId, workspaceId: instance.workspaceId },
                     "Database says the instance is running, but wsman does not know about it. Marking as stopped in database.",
-                    { installation },
+                    { instanceId, workspaceId: instance.workspaceId, installation, phase },
                 );
                 await this.markWorkspaceInstanceAsStopped(ctx, ri, new Date());
             }
@@ -603,6 +620,7 @@ export class WorkspaceManagerBridge implements Disposable {
         const nowISO = now.toISOString();
         info.latestInstance.stoppingTime = nowISO;
         info.latestInstance.stoppedTime = nowISO;
+        info.latestInstance.status.message = `Stopped by ws-manager-bridge. Previously in phase ${info.latestInstance.status.phase}`;
         info.latestInstance.status.phase = "stopped";
         await this.workspaceDB.trace(ctx).storeInstance(info.latestInstance);
 


roboquat:

<details>

</details>

AlexTugarev:
Hi @laushinka! Is this ready for review?

laushinka:
> Hi @laushinka! Is this ready for review?

@AlexTugarev I had issues where the previews would fail, but it looks like the builds succeed now. Would you like to review it? I can squash the changes later to avoid the build possibly breaking in the middle of a review.

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -506,7 +512,12 @@ export class WorkspaceManagerBridge implements Disposable {
 
             for (const [instanceId, ri] of runningInstancesIdx.entries()) {
                 const instance = ri.latestInstance;
-                if (instance.status.phase !== "running") {
+                if (
+                    !(
+                        instance.status.phase === "running" ||
+                        durationLongerThanSeconds(Date.parse(instance.creationTime), maxTimeToRunningPhaseSeconds)
```
svenefftinge, line 518: If I understood correctly you are adding a timeout for the phase `pending`.
We handle timeouts for the various phases in [this method](https://github.com/gitpod-io/gitpod/blob/e184f5a68637e59169848bfefdb5ffbc81086f5e/components/ws-manager-bridge/src/bridge.ts#L573-L611). Would be interesting to understand why the case in [L583](https://github.com/gitpod-io/gitpod/blob/e184f5a68637e59169848bfefdb5ffbc81086f5e/components/ws-manager-bridge/src/bridge.ts#L583-L584) doesn't do it.

geropl: @svenefftinge We have two controlling methods, which check instances in different phases, to clearly separate concerns. This was introduced after we detected outdated and broken logic in the old, single method. :upside_down_face: 

1. This one here is responsible for handling phases which [ws-manager](url) is responsible for, e.g. things that for whatever reason felt through the cracks, incl [running](url) and `pending`.
2. The one you linked is responsible for controlling phases WebApp is responsible for, e.g. `preparing` and `building`.

FYI: There is [a diagram](https://www.notion.so/gitpod/WebApp-Architecture-784cf14bc76440e3beb8de56f61e7340#6c9bd05eef5a44408171a95b14ae73a9) that tries to illustrate this a little bit.

svenefftinge: Ah, yes, that makes sense. I was confusing `pending` and `preparing`

geropl: > I was confusing pending and preparing

Yeah, I agree. All of this feels more complex then it should :roll_eyes: 


@laushinka I'd really love to have at least a comment on this line which explains why we think this is an instance we do (not) want to stop.

svenefftinge: Looking at the logic now (btw. it's a little hard to wrap my head around the negated or, tbh 😇) it seems like it would also handle all `stopping` cases. Which is generally great as we have leaking stopping states, that we should handle.
But applying we should apply the timeout on top of stoppingTime instead of creationTime in this case.
Also I think this code would now also trigger for the phases webapp is responsible for if the the timeout is smaller than the respective timeout for the phase.

How about we explicitly list the phases we want to handle in this function (i.e. ws-manager responsibilities) and use a single timeout against the `max(CreationTime, StartedTime, StoppingTime)` (i.e. the last signal we got). The timeout doesn't really need to be configurable I think, but can be rather small and static. We mainly want to make sure we don't act on anything that just in this moment got removed from ws-manager (i.e. races).

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -446,13 +446,18 @@ export class WorkspaceManagerBridge implements Disposable {
...
-                    const runningInstances = await this.workspaceDB
+                    const nonStoppedInstances = await this.workspaceDB
                         .trace(ctx)
                         .findRunningInstancesWithWorkspaces(installation, undefined, true);
 
                     // Control running workspace instances against ws-manager
                     try {
-                        await this.controlRunningInstances(ctx, runningInstances, clientProvider);
+                        await this.controlNonStoppedWSManagerManagedInstances(
+                            ctx,
```
laushinka, line 456: @svenefftinge What do you think about this name?

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -504,11 +510,21 @@ export class WorkspaceManagerBridge implements Disposable {
...
+
+                if (
+                    phase === "pending" ||
+                    phase === "creating" ||
+                    phase === "initializing" ||
+                    (phase === "stopping" &&
+                        instance.stoppingTime &&
+                        durationLongerThanSeconds(Date.parse(instance.stoppingTime), 10))
+                ) {
                     log.debug({ instanceId }, "Skipping instance", {
```
svenefftinge, line 526: I think we want to stop these instances as well as when `phase === 'running'`. So maybe you want to negate 😅 the logic or just move the lines 534-539 up into this if block (and also include phase===running).

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -603,6 +619,7 @@ export class WorkspaceManagerBridge implements Disposable {
         const nowISO = now.toISOString();
         info.latestInstance.stoppingTime = nowISO;
         info.latestInstance.stoppedTime = nowISO;
+        info.latestInstance.status.message = `Stopped by ws-manager-bridge. Previously in phase ${info.latestInstance.status.phase}`;
```
svenefftinge, line 622: 👍

components/ws-manager-bridge/src/config.ts
```diff
@@ -38,4 +38,7 @@ export interface Configuration {
 
     // clusterSyncIntervalSeconds configures how often we sync workspace cluster information
     clusterSyncIntervalSeconds: number;
+
+    // maxTimeToRunningPhaseSeconds is the timeout duration
+    maxTimeToRunningPhaseSeconds: number;
```
svenefftinge, line 43: I would love to not expose this as a property, but simply inline a magic number. I don't see why this should be configurable and the config surface is already rather large.

laushinka: Oops I missed to remove the whole part of this config which is not needed anymore 🙏🏽

components/ws-manager-bridge/src/bridge.ts
laushinka: My brain 🤦🏻‍♀️🤦🏻‍♀️🤦🏻‍♀️

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -504,23 +504,37 @@ export class WorkspaceManagerBridge implements Disposable {
...
+                    phase === "pending" ||
+                    phase === "creating" ||
+                    phase === "initializing" ||
+                    phase === "running" ||
+                    (phase === "stopping" &&
+                        instance.stoppingTime &&
+                        durationLongerThanSeconds(Date.parse(instance.stoppingTime), 10))
+                ) {
+                    log.info(
+                        { instanceId, workspaceId: instance.workspaceId },
```
geropl, line 525: nit: `phase` would be nice to have in here as well.

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -504,23 +504,37 @@ export class WorkspaceManagerBridge implements Disposable {
...
+                        "Database says the instance is running, but wsman does not know about it. Marking as stopped in database.",
+                        { installation },
```
geropl, line 527: nit: `phase` would be nice to have in here as well.

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -504,23 +504,37 @@ export class WorkspaceManagerBridge implements Disposable {
...
+                    );
+                    await this.markWorkspaceInstanceAsStopped(ctx, ri, new Date());
```
geropl, line 529: What about adding a `continue` here? That would avoid the additional info "skipping instance" below.

laushinka: Thanks for the catch! And I should've removed the continue in the line below.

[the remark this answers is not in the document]
laushinka: Definitely. I was thinking about how to test this better, because so far I tested it by cordoning the preview env node, but it feels lacking.

Logging first enable later could be a start. Could you elaborate on the separate branch idea?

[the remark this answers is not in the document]
laushinka: Also happy to have another pair of eyes when testing. So far my flow is:
1. Cordon the node
2. Run a workspace
3. See it stuck on pending
4. See stoppingTime and stoppedTime inserted to the DB ✅

But for some reason it takes a while until the phase is stopped.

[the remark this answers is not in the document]
svenefftinge: Maybe then you put in the old condition again and add the new if branch with just logging. We can review the log messages and then in another deploy activate the 'stopping'.

components/ws-manager-bridge/src/bridge.ts
svenefftinge: could you also put all the properties into one object?
I.e. as in: 
```ts
loginfo("my message ...", { all, the, properties })
```

components/ws-manager-bridge/src/bridge.ts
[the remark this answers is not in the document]
laushinka: YES let's do that 🚀

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -504,9 +504,27 @@ export class WorkspaceManagerBridge implements Disposable {
             const actuallyRunningInstances = await client.getWorkspaces(ctx, new GetWorkspacesRequest());
             actuallyRunningInstances.getStatusList().forEach((s) => runningInstancesIdx.delete(s.getId()));
 
+            // runningInstancesIdx only contains instances that ws-manager is not aware of
             for (const [instanceId, ri] of runningInstancesIdx.entries()) {
                 const instance = ri.latestInstance;
-                if (instance.status.phase !== "running") {
+                const phase = instance.status.phase;
+                if (phase !== "running") {
+                    // This below if block is to validate the planned fix
```
laushinka, line 512: I did it this way because we want to keep the previous check. Therefore I also had to remove the if running check below. Does this make sense? @svenefftinge @geropl

svenefftinge (approved):

components/ws-manager-bridge/src/bridge.ts
svenefftinge: perfect! let's see how it does in production 😅

components/ws-manager-bridge/src/bridge.ts
geropl: @svenefftinge There is a reason we have two bags: The first one is the OWI (owner-workspace-instance) shape. Having this typed means it's very easy to find all logs realted to a specific user or workspace or instance, across all components, across all tools (logs and tracing). Moving it our there breaks this. :confused: 
The second bag at the end is specifically for payload (arbitrary shape), which ends up in a different place in log entries.

svenefftinge: Oh sorry, I didn't know we have this special OWI object.

laushinka: I didn't know either. Will create a PR to fix it back!

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -504,9 +504,27 @@ export class WorkspaceManagerBridge implements Disposable {
...
+                    if (
+                        phase === "pending" ||
+                        phase === "creating" ||
+                        phase === "initializing" ||
+                        (phase === "stopping" &&
+                            instance.stoppingTime &&
+                            durationLongerThanSeconds(Date.parse(instance.stoppingTime), 10))
```
geropl, line 519: @laushinka One more comment on the timeout discussion (sorry for doing this out of band, but couldn't wrap my head around yesterday): We have to make sure that short disconnects between application and workspace clusters don't lead to falsely stopped workspaces. We've seen instances in the past where we had problems on and off for a couple of **minutes** (might have been due to bad gRPC connection options, might or might not be better by now). With this change, this would lead to a situation where we mark the workspace as stopped in the database, while users can use and work in their workspaces perfectly fine. :grimacing: 

I guess we just be very carefule with this change. Let's monitor this for now, and a couple of days. Plus double-check if we find hints in the logs on bridge -> ws-manager re-connects (this should be the [relevant line](https://github.com/gitpod-io/gitpod/blob/2823606369b667188187b1e8ee9a75cbe968633b/components/ws-manager-bridge/src/wsman-subscriber.ts#L38) in code).

PR #12903, opened 2022-09-13
[usage-api] Use better typescript generator

### LINKED ISSUE #12880: [protobuf] Use better TypeScript binding
The generated typescript bindings are supoptimal and there are way better more idiomatic code generators.
I have played around with https://github.com/stephenh/ts-proto which provides a nice API and is compatible with https://github.com/improbable-eng/grpc-web

svenefftinge:
## Description
This PR uses an alternative typescript generator for the protobuf bindings, that generates more ideomatic TypeScript.

## Related Issue(s)
Fixes #12880

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment


easyCZ (approved):
PR could benefit from a nicer description to explain that it's using [nice-grpc](https://www.npmjs.com/package/nice-grpc) which is a higher level abstraction on top of [improbable/grpc-web](https://www.npmjs.com/package/@improbable-eng/grpc-web) and uses modern TS to wrap the client into a more usable form.

Change looks great, thanks for digging into this. Let's see how it works for the usage service and we can then migrate the remaining grpc clients instances.

components/server/src/container-module.ts
```diff
@@ -264,19 +258,14 @@ export const productionContainerModule = new ContainerModule((bind, unbind, isBo
 
     bind(NewsletterSubscriptionController).toSelf().inSingletonScope();
 
-    bind(UsageServiceClientConfig).toDynamicValue((ctx) => {
+    bind<UsageServiceClient>(UsageServiceDefinition.name).toDynamicValue((ctx) => {
```
AlexTugarev, line 261: @svenefftinge, shouldn't that be bound in singleton scope?

Asking, because of that one of the changes being introduced/deployed after which we see a severe memory leak.

svenefftinge: Yes, I think it can be in singleton scope as well. But if this increases leakage it could be there are other components that leak as well and hold references to these instances.

Issue #12283, opened 2022-08-22
server: please log any call to StopWorkspace with the reason why it is called


sagor999:
## Is your feature request related to a problem? Please describe

On workspace side we only see `StopWorkspace` is being called. But looking at webapp logs, there is nothing in logs that would explain why StopWorkspace was called.

## Describe the behaviour you'd like

Can you please log a reason for anytime server is calling `StopWorkspace` to  ease with debug.
For example, if user requested to stop, or if some internal server logic decided to call it (like cancel prebuild because there is a newer commit).

kylos101:
@sagor999 wdyt of adding traces too?

We might even be able to ensure that this particular trace is sampled at 100%. @mads-hartmann is it possible for us to configure certain traces such that they are always sampled, instead of at the default (I think 5 or 10% is the default for traces).

mads-hartmann:
Ideally you would add both a log line and attach the information to the relevant span as the telemetry have different strengths depending on what kinds of questions you're trying to answer. For "system auditing" like this where we want to know why the system took a specific action against a specific workspace, logs are the best approach. If on the other hand you want to look at it more holistically, e.g. get a sense of what the distribution of "reasons" for stopping workspaces are, and you want to compute error rates, filter by cluster, etc. etc. then the tooling we have for traces are much better; and in that kind of analysis you really don't care about individual workspaces or requests, but the system as a whole.

I tried to capture some of this in [When to use what telemetry](https://www.notion.so/gitpod/When-to-use-what-telemetry-d38caafd594a45a598ad220149797e7d) but I haven't revised it since it was originally written so please comment if things are not clear 🧡 

Ideally we wouldn't have to worry about logs vs traces as a span is really just a structured log event with a fixed schema that captures the hierarchy of operations. I will create an issue to create a PoC that would allow us to set an attribute on a span (e.g. audit_log) that ensure that such spans are always sampled. With that in place we would be able to use spans for audit like logs, but I'm not a sure it's the right approach as I still think log lines are required in order for our self-hosted customers to understand why the system is behaving the way it is ☺️

easyCZ:
[...]

PR #12906, opened 2022-09-13
Add an explicit StopWorkspaceReason to all StopWorkspaceRequests

jankeromnes:
## Description

Add an explicit `StopWorkspaceReason` to all `StopWorkspaceRequest`s.

**Hint**: Reviewing each commit separately can help reduce the noise (a protoc regeneration was needed)

**Question**: This new `reason` field is not yet logged or added to tracing. Should it?

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/12283

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


components/ws-manager-api/core.proto
```diff
@@ -124,6 +127,29 @@ enum StopWorkspacePolicy {
     ABORT = 2;
 }
 
+enum StopWorkspaceReason {
+    // The user stopped the workspace
+    USER_STOPPED = 0;
+
+    // The user deleted the workspace
+    USER_DELETED = 1;
```
easyCZ, line 135: How is a STOP different to DELETE? I would've assumed that on the ws-manager API level they are one and the same, no?

components/ws-manager-api/core.proto
```diff
@@ -124,6 +127,29 @@ enum StopWorkspacePolicy {
     ABORT = 2;
 }
 
+enum StopWorkspaceReason {
+    // The user stopped the workspace
+    USER_STOPPED = 0;
```
easyCZ, line 132: 0 index enum should always be an UNKNOWN value. This is because when a field is not set in proto/grpc, it defaults to the 0 value. For enums, that's the value with the 0 index. In this case, if our system forgot to set the reason, it would default to USER_STOPPED which is not what we want - we want it to be UNKNOWN.

components/ws-manager-api/core.proto
```diff
@@ -124,6 +127,29 @@ enum StopWorkspacePolicy {
...
+
+    // The user deleted the workspace
+    USER_DELETED = 1;
+
+    // The user cancelled the running prebuild
+    USER_CANCELLED_PREBUILD = 2;
```
easyCZ, line 138: How is this different from USER_STOPPED? I'd say that if you stop a prebuild, you're effectively doing the same.

components/ws-manager-api/core.proto
```diff
@@ -124,6 +127,29 @@ enum StopWorkspacePolicy {
...
+
+    // An admin force-stopped the workspace
+    ADMIN_STOPPED = 3;
+
+    // An admin blocked the user
+    ADMIN_BLOCKED = 4;
```
easyCZ, line 144: IMO, we shouldn't encode if an Admin did it or not. This assumes that there's a very binary distinction between user and admin (which isnt' the case, and won't be in the future either). Instead, if we want the identity we should also specify it as a field in the request - `StopRequestedBy = milan`

components/ws-manager-api/core.proto
```diff
@@ -124,6 +127,29 @@ enum StopWorkspacePolicy {
...
+
+    // A new commit was pushed to a repository branch, so we cancelled all previously running prebuilds for that branch
+    SYSTEM_CANCELLED_BRANCH_PREBUILD = 5;
+
+    // The user exceeded their allowed usage limit (e.g. number of hours in month), so we stopped the workspace
+    SYSTEM_USER_LIMIT_REACHED = 6;
```
easyCZ, line 150: I'd generalize this. I'd go with `LIMIT_EXCEEDED` and an additional human message can specify if this is due to rate limiting prebuilds, or credits etc.

components/ws-manager-api/core.proto
jankeromnes: Yes, but the user intent is different:
- Stopped because user wanted it stopped
- Stopped automatically because user wants it deleted

components/ws-manager-api/core.proto
jankeromnes: Thanks! Good idea (I also felt there was a missing "unknown" state here, although there is no need for it yet in the current code)

components/ws-manager-api/core.proto
easyCZ: Sure, but that's the API that Server/Dashboard exposes, not the API that `ws-manager` exposes. And here, we're dealing with the ws-manager API. 

A delete is effectively a Stop followed by a Workspace Delete. To a ws-manager, it doesn't really make sense to broaden it's domain to a delete when it currently doesn't understand what a delete is (from it's API surface perspective)

components/ws-manager-api/core.proto
jankeromnes: True, but the code paths are slightly different. Still, I'm more than happy to make `USER_STOPPED` and `USER_CANCELLED_PREBUILD` the same thing.

components/ws-manager-api/core.proto
easyCZ: There is on the protocol level however. For example, if ws-manager implemented using the StopWorkspaceReason before server started setting it, it would always get the USER_STOPPED. There's a proto lint rule that should've caught this, I'll need to investigate why it didn't trigger here.

components/ws-manager-api/core.proto
easyCZ: Ultimately this is up to Workspace team, but my reasoning is based on this:

* We want to use the enum to categorise classes of reasons, not every single reason. This aids composability but also reduces decision paralysis for which category to use.
* We _should_ have an extra field for the reason which explains it for the human. This can also include information such as if an admin did it, or if it's due to lack of credits. The system wouldn't use this string directly to make decisions, it would only store it to show the user.
* The enums here should not account for every possible way that server can use the StopWorkspace RPC. Instead, it needs to consider the API domain from the `ws-manager` perspective. We do not want to leak concepts such as credits or billing into here, nor do we want to leak concepts like admin or regular user - those belong into the WebApp domain model but not necessarily into ws-manager.
* We can always add reasons here, so if there's a sufficient reason to expand, we always can. Removing is a lot more difficult.

jankeromnes:
Closing this as is for now, because such a change needs more discussion first.

PR #12911, opened 2022-09-13
Remove deprecated sendFeedback server method

jankeromnes:
## Description

Remove deprecated `sendFeedback` server method.

**Note**: This is completely unrelated to the new dashboard feedback mechanism. 💡

In the early days of Gitpod (I think when it was still in Beta, or fresh out of Beta) there was a simple mechanism to provide general feedback about Gitpod, and in exchange you would receive a credit of 50 bonus Gitpod hours.

This feature no longer exists, and we no longer even keep track of the provided feedback (it's logged at `info` level in server, which we currently discard), but the server method is still there in our API, potentially allowing clever users to credit themselves 50 hours whenever they want.

This PR removes the old, deprecated `sendFeedback` method. Also, removing it now is a good idea, because we have a new dashboard feedback mechanism in place, and it's too easy to accidentally mistake one for the other.

## Related Issue(s)

Fixes #

## How to test

1. Build should pass
2. Gitpod should still work

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


jankeromnes:
Werft was having too much trouble with this -- moving on to https://github.com/gitpod-io/gitpod/pull/12918.

Issue #6770, opened 2021-11-18
[ws-manager-bridge] Garbage-collect workspace instances whose workspace clusters are not available anymore


geropl:
context:
 - internal slack discussion: https://gitpod.slack.com/archives/C020VCB0U5A/p1636974437234400?thread_ts=1636963076.228900&cid=C020VCB0U5A
 - PR: https://github.com/gitpod-io/gitpod/pull/6682

JanKoehnlein:
I am trying to reverse-engineer the action for this issue form the given context. Please clarify: Is this about 
1) adding an additional (timeout-based?) mechanism on the `ws-manager-bridge` that also works when the cluster goes donw without deregistering, or
2) taking further action in the `ws-manager-bridge` when receiving a forced deregistration request
3) None of the above (please specify)

geropl:
Sorry for lagging details: This is about adding a mechanism that ensures we don't leak workspaces dangling in any state other than `stopped` once a workspace cluster is de-registered.

We could discuss whether this should be:
 a) timeout-based, and workspace may re-appear if we re-register a workspace quickly enough. This would make it a bit more fault-tolerant and safe. This is samewhat tricky, as we need to poll `d_b_workspace` by `region`. But it's impossible atm for `ws-manager-bridge` to distinguish between "I'm don't know this region but someone else does" and "no-one is governing this region".
 b) using the "forced de-registration" request as signal might serve as a first version. Although this might lead to unwanted fall-out in case we mis-used the `gptcl clusters` command.

a) will become possible once we have the changes required for simplified meta, hence I did not prioritize this, yet.

geropl:
Assigned this to epic "Simplified Multi-Meta"

geropl:
@jankeromnes Please ping if I can help with additional context/details 👍

jankeromnes:
Many thanks @geropl! Planning to start looking into either this or https://github.com/gitpod-io/gitpod/issues/12283 once https://github.com/gitpod-io/gitpod/issues/12580 is done.

jankeromnes:
Okay, I'm now blocked on [this question](https://github.com/gitpod-io/gitpod/issues/12283#issuecomment-1245172696) and have already started [one distraction](https://github.com/gitpod-io/gitpod/pull/12911), so I guess now is the time to pick this one up! 😆 🚀

jankeromnes:
[...]

PR #6682, opened 2021-11-12
[ws-manager-bridge] Support forced cluster dereg

csweichel:
## Description
This PR adds support for forced workspace cluster deregistration. 
When de-registering a cluster, ws-manager-bridge checks if there are still any workspaces which belong to this cluster and are marked as running. If we removed the cluster in this case, those instances would never again receive a status update - also the reconciliation in ws-manager-bridge would no longer act on those workspaces (the cluster is gone after all, what should we reconcile with).

With the newly added `--force` flag on gpctl, we can override this safety mechanism in exceptional cases, e.g. when introducing new workspace cluster types.

## How to test
1. Register the same ws-manager as dynamic cluster and give it a very high score
2. Start a workspace on that cluster
3. Deregister the newly added cluster - should not work without `--force`, but should work with it.

## Release Notes

```release-note
[gpctl] Support forceful cluster de-registration
```

Issue #12387, opened 2022-08-25
Make use of `final_backup_complete ` workspace condition


svenefftinge:
As a follow-up to https://github.com/gitpod-io/gitpod/issues/11453 this is about introducing a clear signal from ws-manager that a workspace hasn't gotten any backup. This is important information to signal to users, because such workspaces are broken and we should also forbid starting them (resulting in):
<img width="389" alt="Screenshot 2022-08-25 at 14 14 39" src="https://user-images.githubusercontent.com/372735/186668631-fada8282-7eb0-464c-95e1-a64e77eefafb.png">

svenefftinge:
@sagor999 Is the `final_backup_complete` WorkspaceCondition always sent when a backup happened? I.e. can we rely on treating workspace instances for which we never received this condition to have no backup?

sagor999:
Timely issue, thanks Sven!
I had a similar discussion with @geropl about this and we had a plan to introduce condition `hasBackup` to signal when backup was created. 
I was going to work on this today so this is a great reminder to implement this. 😀
More info to follow later today.

sagor999:
@svenefftinge you are right, we can totally use `final_backup_complete` condition.
This PR #12408 improved reliability of that condition by making sure it is set to true correctly.
So once that is merged, server instead of this code:
https://github.com/gitpod-io/gitpod/blob/0699f84aba13019cbdb3c13dce79b2d33541a56f/components/server/src/workspace/workspace-starter.ts#L334-L336
Should use something like this:
```
const hasValidBackup = pastInstances.some( 
     (i) => !!i.status && !!i.status.conditions && i.status.conditions.final_backup_complete, 
 ); 
```
We should probably either use it only for workspaces created after X timestamp (once #12408 is deployed) or maybe fix DB to ensure `final_backup_complete` is set correctly for old workspaces? Not sure which approach would be better. Or if there is a better alternative.

geropl:
Thank you @sagor999 for letting us now about https://github.com/gitpod-io/gitpod/pull/12408! From the data I saw I thought it has already been abandoned - good to have it back! :lotus_position:

kylos101:
@sagor999 looks like webapp has related work Scheduled for this issue, let's leave this In-Validation until that is deployed. 👍 

@geropl 🙏 thank you for Scheduling related work! As an FYI, we're tracking this in our In-Validation column, so we'll bump into and check/review it weekly.

jenting:
[...]

PR #12912, opened 2022-09-13
When de-registering a workspace cluster, mark any leftover running instances as stopped/failed

jankeromnes:
## Description

When de-registering a workspace cluster, mark any leftover running instances as stopped/failed in `ws-manager-bridge`.

Drive-by: Delete instance tokens and track workspace stop also in irregular cases.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/6770

## How to test

From https://github.com/gitpod-io/gitpod/pull/6682#issue-1051856697:

> 1. Register the same ws-manager as dynamic cluster and give it a very high score

E.g. by running:

```
gpctl clusters get-tls-config
gpctl clusters register --name temp --url dns:///ws-manager:8080 --tls-path ./wsman-tls/
gpctl clusters update --name temp score 1000
```

> 2. Start a workspace on that cluster
> 3. Deregister the newly added cluster - should not work without `--force`, but should work with it.

4. Your started workspace should be marked as stopped/failed, instead of remaining stuck in a non-final state like running or stopping.

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/ws-manager-bridge/src/bridge.ts
+++ b/components/ws-manager-bridge/src/bridge.ts
@@ -120,6 +120,7 @@ export class WorkspaceManagerBridge implements Disposable {
     }
 
     public stop() {
+        this.markAllRunningWorkspaceInstancesAsStopped();
         this.dispose();
     }
 
@@ -643,6 +644,35 @@ export class WorkspaceManagerBridge implements Disposable {
                 messageId: `bridge-wsstopped-${instance.id}`,
                 properties: { instanceId: instance.id, workspaceId: instance.workspaceId },
             });
+        } catch (err) {
+            TraceContext.setError({ span }, err);
+            throw err;
+        } finally {
+            span.finish();
+        }
+    }
+
+    protected async markAllRunningWorkspaceInstancesAsStopped(): Promise<void> {
+        const span = TraceContext.startSpan("markAllRunningWorkspaceInstancesAsStopped");
+        try {
+            const now = new Date();
+            const runningInstances = await this.workspaceDB
+                .trace({ span })
+                .findRunningInstancesWithWorkspaces(this.cluster.name, undefined, true);
+            await Promise.all(
+                runningInstances.map(async (info) => {
+                    const logContext: LogContext = {
+                        userId: info.workspace.ownerId,
+                        workspaceId: info.workspace.id,
+                        instanceId: info.latestInstance.id,
+                    };
+                    log.error(logContext, "Marking still-running instance as stopped in database.", {
+                        creationTime: info.workspace.creationTime,
+                        currentPhase: info.latestInstance.status.phase,
+                    });
+                    await this.markWorkspaceInstanceAsStopped({ span }, info, now);
+                }),
+            );
         } catch (err) {
             TraceContext.setError({ span }, err);
             throw err;


jankeromnes:
Hmm, in `bridge.ts`, it looks we have two different logics / flows for marking an instance as stopped. 🤔

## On status update

https://github.com/gitpod-io/gitpod/blob/a1df2b972d4195bca0d3767971e03a1092a54391/components/ws-manager-bridge/src/bridge.ts#L374-L383

https://github.com/gitpod-io/gitpod/blob/a1df2b972d4195bca0d3767971e03a1092a54391/components/ws-manager-bridge/src/bridge.ts#L621-L627

## In irregular cases

https://github.com/gitpod-io/gitpod/blob/a1df2b972d4195bca0d3767971e03a1092a54391/components/ws-manager-bridge/src/bridge.ts#L602-L611

This is called in the following cases:

- ws-manager doesn't know about this instance

https://github.com/gitpod-io/gitpod/blob/a1df2b972d4195bca0d3767971e03a1092a54391/components/ws-manager-bridge/src/bridge.ts#L518-L523

- the instance timed out in `preparing`, `building` or `unknown` phase:

https://github.com/gitpod-io/gitpod/blob/a1df2b972d4195bca0d3767971e03a1092a54391/components/ws-manager-bridge/src/bridge.ts#L583-L593

## Differences

|  | Status Update | Irregular Cases |
| --- | --- | --- |
| Set `stoppingTime`? | Only if not set | Always |
| Delete Gitpod tokens? | Yes | No |
| Track stopped in analytics? | Yes | No |

Are these differences intended? Do they make sense? 🤔

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -601,11 +602,17 @@ export class WorkspaceManagerBridge implements Disposable {
 
     protected async markWorkspaceInstanceAsStopped(ctx: TraceContext, info: RunningWorkspaceInfo, now: Date) {
         const nowISO = now.toISOString();
-        info.latestInstance.stoppingTime = nowISO;
+        if (!info.latestInstance.stoppingTime) {
+            info.latestInstance.stoppingTime = nowISO;
+        }
         info.latestInstance.stoppedTime = nowISO;
         info.latestInstance.status.phase = "stopped";
```
jankeromnes, line 609: Maybe `markWorkspaceInstanceAsStopped` should also set a `failed` condition, or something else that signals that there might not be a backup?

E.g. when I test the current code, any running instance is successfully marked as stopped. However, it's impossible to restart the instance:

Screenshot 2022-09-15 at 14 07 14

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -601,11 +602,17 @@ export class WorkspaceManagerBridge implements Disposable {
 
     protected async markWorkspaceInstanceAsStopped(ctx: TraceContext, info: RunningWorkspaceInfo, now: Date) {
         const nowISO = now.toISOString();
-        info.latestInstance.stoppingTime = nowISO;
+        if (!info.latestInstance.stoppingTime) {
+            info.latestInstance.stoppingTime = nowISO;
+        }
```
svenefftinge, line 607: ❤️

components/ws-manager-bridge/src/bridge.ts
geropl: @jankeromnes There is a condition that signals that there is a successful backup. But it only (re-)stabilized recently, and we do not use it, yet: https://github.com/gitpod-io/gitpod/issues/12387#issuecomment-1243071849

components/ws-manager-bridge/src/bridge.ts
geropl: :+1:

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -601,11 +602,17 @@ export class WorkspaceManagerBridge implements Disposable {
...
         info.latestInstance.stoppedTime = nowISO;
         info.latestInstance.status.phase = "stopped";
         await this.workspaceDB.trace(ctx).storeInstance(info.latestInstance);
 
+        // cleanup
+        // important: call this after the DB update
+        await this.onInstanceStopped(ctx, info.workspace.ownerId, info.latestInstance);
```
geropl, line 614: This feels off, really. Left me check in code/workspace...

geropl: It's legit and actually fixing it: nice! :upside_down_face:

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -120,6 +120,7 @@ export class WorkspaceManagerBridge implements Disposable {
     }
 
     public stop() {
+        this.markAllRunningWorkspaceInstancesAsStopped();
```
geropl, line 123: I'm not sure I like this approach.

So far, stopping a bridge was an operation that has absolutely no effect on workspaces. This made it a very cheap operation, and allowed for great operational complexity. E.g., if you had to fixup a DB entry, you could always remove/re-add an entry, with very limited downsides (delay of workspace updates for a handful of seconds). Or, if you wanted for a reconnect, you could remove and re-add a DB entry. Now, it has a very destructive side-effect.

:thought_balloon: I wonder what happens if we stop a `ws-manager-brige` pod during a rollout. It would stop all workspaces on that cluster, no? @jankeromnes

geropl (changes requested):
https://github.com/gitpod-io/gitpod/pull/12912/files#r971936170

components/ws-manager-bridge/src/bridge.ts
svenefftinge: What do you think of having a periodic clean-up where we check and stop instances for which no ws-manager exists?

easyCZ: Or a more general version: For all currently not-stopped instances, check it against ws-manager. This would catch a broader set of problems, but we already need to solve this problem anyway.

geropl: > What do you think of having a periodic clean-up where we check and stop instances for which no ws-manager exists?

This is something we need anyway, but with different time constraints: This PR is about unblocking Team Workspace in specific cases. This is a separate PR, but same issue.

> For all currently not-stopped instances, check it against ws-manager.

That's a layer above this discussion. Happy to provide context, outside of this PR. :upside_down_face:

jankeromnes: Many thanks for the great feedback here! 💯

> E.g., if you had to fixup a DB entry, you could always remove/re-add an entry, with very limited downsides (delay of workspace updates for a handful of seconds).

But why remove/re-add an entry when you can `gpctl clusters update`? 🤔 (E.g. to adjust the score)

> Or, if you wanted for a reconnect, you could remove and re-add a DB entry. Now, it has a very destructive side-effect.

Can't you kill the `ws-manager-bridge` pod, or `rollout restat` its deployment to achieve this?

> :thought_balloon: I wonder what happens if we stop a `ws-manager-brige` pod during a rollout. It would stop all workspaces on that cluster, no? @jankeromnes 

I wasn't sure, so I tested it:
- I ran `kubectl delete pod ws-manager-bridge-[tab]` several times in a row
- I also ran `kubectl rollout restart deployment ws-manager-bridge` a few times

My running workspace stayed alive and well all the time.

Only when I ran `gpctl clusters deregister --name temp --force` did it get marked as stopped.

jankeromnes: Sync feedback from @geropl: Marking all instances as stopped in this `bridge.stop()` lifecycle method is a change in behavior -- does it really make sense?

One the one hand ("my" view), `bridge.stop()` is only called:
- [here](https://github.com/gitpod-io/gitpod/blob/a921c1464bbe87d0772ce5db5c6cc852ce579aac/components/ws-manager-bridge/src/bridge-controller.ts#L74) when a cluster was de-registered (either via RPC with --force, or manually deleted from the DB) and 2. reconcile is called. In this case, the cluster is no longer connected, and we'll never hear again about any instances that were still running on it (so we can mark them as stopped/failed)
- [here](https://github.com/gitpod-io/gitpod/blob/a921c1464bbe87d0772ce5db5c6cc852ce579aac/components/ws-manager-bridge/src/bridge-controller.ts#L128) when we're disposing of the bridge controller (is this actually ever called? It [doesn't look like it](https://github.com/gitpod-io/gitpod/blob/88e5beb8d6942e7632f4237906b9319b71500376/components/ws-manager-bridge/src/main.ts#L46-L47), so maybe we could delete this dead / somewhat dangerous code)

So this means that `bridge.stop()` is only ever called when you're actually de-registering a cluster, and instead of leaving all the instances as they are without any further updates, maybe it makes more sense to mark them as stopped/failed.

On the other hand (@geropl's view), maybe we need to be clearer or more careful about the fact that calling `bridge.stop()` will also mark all of its instances as stopped -- i.e. this doesn't seem to be called currently when you try to reconnect a bridge, but we need to make sure that it also won't be called in the future with the assumption that instances will be kept running). Maybe this can be achieved with a comment, or by renaming the `stop` function. We could also make a bridge mark all its instances as stopped only when we receive a deregister --force RPC, but this seems a bit more complicated (need to extract/share code or somehow give access to the bridge to the RPC handler), and wouldn't handle the cases where a cluster is manually dropped from the DB.

My personal conclusion here would be to leave the PR as is, and simply add very explicit comments to stress the fact that calling `bridge.stop()` also marks all its instances as stopped in the DB. Or, we could even rename the method to `bridge.tearDown()` or `bridge.markAllInstancesAsStoppedAndDispose` or similar. 😊

svenefftinge: What you describe makes sense to me, it just seems that the lifecycle you are attaching the effect of stopping all instances is not well understood (I agree we should better understand it, and remove code that is not called in practice). Attaching the clean-up to the disconnect lifecycle also requires it to be always called, so that we don't leak/forget about workspaces.

It seems more general and less complicated to clean up instances for which we don't know a manager on a regular schedule instead of on disconnect events.

geropl:
@jankeromnes Given the mis-alignment [here](https://github.com/gitpod-io/gitpod/pull/12912#discussion_r973102566) (which is also due to other recent PRs improving in this area: I think it make sense to ship the fixes to `markWorkspaceInstanceAsStopped` now, and align on the lifetime issues outside of this PR. WDYT?

jankeromnes:
Thanks for the discussions! Moving this back to Draft, and splitting out the drive-by fix into https://github.com/gitpod-io/gitpod/pull/13074 so that it can be merged earlier.

PR #12913, opened 2022-09-13
[usage] Cost centers get `nextBillingTime`

svenefftinge:
## Description
This PR contains small steps towards having cost centers for all users and teams, even if they haven't signed up for stripe pay-as-you go, yet, but using the billing strategy `other`. 

In this PR I
 - introduce a property `nextBillingTime` on cost centers that will be used in an upcoming PR where a control loop uses the date to find cost centers that need to run finalization (i.e. create a synthetic invoice to balance out the ledger table).
 - make `GetCostCenter` always return a cost center

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment


roboquat:

<details>

</details>

components/usage-api/usage/v1/usage.proto
```diff
@@ -114,4 +114,7 @@ message CostCenter {
         BILLING_STRATEGY_OTHER = 1;
     }
     BillingStrategy billing_strategy = 3;
+
+    // next_billing_time specifies when the next billing cycle happens. Only set when billing strategy is 'other'. This property is readonly.
```
easyCZ, line 118: Nice use of proto comments to describe semantics!

components/usage/config.json
```diff
@@ -1,5 +1,7 @@
 {
   "controllerSchedule": "1h",
+  "defaultSpendingLimitForUsers": 5000,
+  "defaultSpendingLimitForTeams": 0,
```
easyCZ, line 4: 👍

components/usage/pkg/apiv1/usage.go
```diff
@@ -24,10 +24,17 @@ import (
 
 var _ v1.UsageServiceServer = (*UsageService)(nil)
 
+type UsageConfig struct {
+	DefaultSpendingLimitForUsers int32 `json:"defaultSpendingLimitForUsers,omitempty"`
+
+	DefaultSpendingLimitForTeams int32 `json:"defaultSpendingLimitForTeams,omitempty"`
```
easyCZ, line 30:
```suggestion
	DefaultSpendingLimitForUsers int `json:"defaultSpendingLimitForUsers,omitempty"`

	DefaultSpendingLimitForTeams int `json:"defaultSpendingLimitForTeams,omitempty"`
```
This should be sufficient. `int` is actually compiler target dependant but often it's easier to use `int` and let the compiler deal with it.

components/usage/pkg/apiv1/usage.go
```diff
@@ -24,10 +24,17 @@ import (
...
+}
+
 type UsageService struct {
 	conn    *gorm.DB
 	nowFunc func() time.Time
 	pricer  *WorkspacePricer
+	cfg     UsageConfig
```
easyCZ, line 37: I think passing in the `UsageConfig` is exposing too much, I think Maybe just passing in something like 
```
type SpendingLimitDefaults strut {
   Teams, Individuals  int
}
```
would be sufficient here.

components/usage/pkg/apiv1/usage.go
```diff
@@ -164,12 +171,30 @@ func (s *UsageService) GetCostCenter(ctx context.Context, in *v1.GetCostCenterRe
...
+				SpendingLimit:   defaultSpendingLimit,
+				NextBillingTime: db.NewVarcharTime(time.Now().AddDate(0, 1, 0)),
+			}
+			result, err = db.SaveCostCenter(ctx, s.conn, result)
+			if err != nil {
+				return nil, status.Errorf(codes.Internal, "Failed to store new cost center in DB: %s", err.Error())
+			}
+		} else {
+			return nil, status.Errorf(codes.Internal, "Failed to get cost center %s from DB: %s", in.AttributionId, err.Error())
 		}
```
easyCZ, line 197: How about we shift this logic into the DB by doing `GetOrCreate`?

components/usage/pkg/server/server.go
```diff
@@ -35,6 +36,8 @@ type Config struct {
 	StripeCredentialsFile string `json:"stripeCredentialsFile,omitempty"`
 
 	Server *baseserver.Configuration `json:"server,omitempty"`
+
+	apiv1.UsageConfig
```
easyCZ, line 40: I'd name this explicitly. It also allows you to be explicit about the `json` field name which guards against issues when you rename the `apiv1.UsageConfig` in the future.

install/installer/pkg/config/v1/experimental/experimental.go
```diff
@@ -246,6 +246,8 @@ type UsageConfig struct {
 	Enabled                          bool               `json:"enabled"`
 	Schedule                         string             `json:"schedule"`
 	BillInstancesAfter               *time.Time         `json:"billInstancesAfter"`
+	DefaultSpendingLimitForUsers     int32              `json:"defaultSpendingLimitForUsers"`
+	DefaultSpendingLimitForTeams     int32              `json:"defaultSpendingLimitForTeams"`
```
easyCZ, line 250:
```suggestion
	DefaultSpendingLimitForUsers     int              `json:"defaultSpendingLimitForUsers"`
	DefaultSpendingLimitForTeams     int              `json:"defaultSpendingLimitForTeams"`
```

components/usage/pkg/apiv1/usage.go
svenefftinge: `UsageConfig` is a new type I introduced for this.

easyCZ: Ah, my bad. I took as the UsageComponentConfig. Ignore then.

install/installer/pkg/config/v1/experimental/experimental.go
svenefftinge: We use int32 everywhere for spending limit. If we want to chaneg this I'd prefer to do it in a separate PR

components/usage/pkg/db/cost_center_test.go
```diff
@@ -30,28 +33,77 @@ func TestCostCenter_WriteRead(t *testing.T) {
 	require.NoError(t, tx.Error)
 	require.Equal(t, costCenter.ID, read.ID)
 	require.Equal(t, costCenter.SpendingLimit, read.SpendingLimit)
+}
+
+func TestGetCostCenter(t *testing.T) {
```
easyCZ, line 38:
```suggestion
func TestCostCenterManager_GetOrCreateCostCenter(t *testing.T) {
```

components/usage/pkg/db/cost_center_test.go
```diff
@@ -30,28 +33,77 @@ func TestCostCenter_WriteRead(t *testing.T) {
...
 	t.Cleanup(func() {
-		conn.Model(&db.CostCenter{}).Delete(costCenter)
+		conn.Model(&db.CostCenter{}).Delete(teamCC, userCC)
 	})
+	require.Equal(t, int32(0), teamCC.SpendingLimit)
+	require.Equal(t, int32(500), userCC.SpendingLimit)
 }
 
-func TestGetCostCenter(t *testing.T) {
+func TestSaveCostCenter(t *testing.T) {
```
easyCZ, line 59:
```suggestion
func TestCostCenterManager_SaveCostCenter(t *testing.T) {
```

components/usage/pkg/db/cost_center.go
```diff
@@ -27,36 +29,136 @@ type CostCenter struct {
...
 	costCenter := results[0]
-	return &costCenter, nil
+	return costCenter, nil
 }
 
-func SaveCostCenter(ctx context.Context, conn *gorm.DB, costCenter *CostCenter) (*CostCenter, error) {
-	db := conn.WithContext(ctx)
-	costCenter.CreationTime = NewVarcharTime(time.Now())
-	db = db.Save(costCenter)
+func (c *CostCenterManager) SaveCostCenter(ctx context.Context, costCenter CostCenter) (CostCenter, error) {
```
easyCZ, line 105: From a consumer of the `CostCenterManager` API perspective, it's quite unclear what the difference is between `GetOrCreate` and `SaveCostCenter`. Why would I use one over the other?

I think here, we can rename it to
```suggestion
func (c *CostCenterManager) UpdateCostCenter(ctx context.Context, costCenter CostCenter) (CostCenter, error) {
```
To better communicate that this is intended to work on an existing record(even if it needs to be created under the hood)

components/usage/pkg/db/cost_center.go
```diff
@@ -27,36 +29,136 @@ type CostCenter struct {
...
+	costCenter.CreationTime = NewVarcharTime(now)
+	db := c.conn.Save(&costCenter)
 	if db.Error != nil {
-		return nil, fmt.Errorf("failed to save cost center: %w", db.Error)
+		return CostCenter{}, fmt.Errorf("failed to save cost center for attributionID %s: %w", costCenter.ID, db.Error)
 	}
 	return costCenter, nil
 }
+
+func (c *CostCenterManager) ComputeFinalization(ctx context.Context, costCenter CostCenter) (Usage, error) {
```
easyCZ, line 144:
```suggestion
func (c *CostCenterManager) ComputeFinalization(ctx context.Context, costCenter CostCenter) (Usage, error) {
```
From this signature, I'd have expected that it will also store the Usage record in the DB. Perhaps renaming to `InvoiceUsageRecordForCostCenter` would convey this better.

Additionally, why this is method accept the `costCenter CostCenter` when it only uses the `costCenter.ID`? Could we supply only the ID here?

components/usage/pkg/db/cost_center_test.go
```diff
@@ -30,28 +33,77 @@ func TestCostCenter_WriteRead(t *testing.T) {
...
+	teamCC.BillingStrategy = db.CostCenter_Stripe
+	newTeamCC, err := mnr.SaveCostCenter(context.Background(), teamCC)
 	require.NoError(t, err)
-	require.Equal(t, costCenter.ID, results.ID)
+	require.Equal(t, db.CostCenter_Stripe, newTeamCC.BillingStrategy)
+	require.Equal(t, newTeamCC.CreationTime.Time().AddDate(0, 1, 0).Truncate(time.Second), newTeamCC.NextBillingTime.Time().Truncate(time.Second))
+}
 
+func cleanUp(t *testing.T, conn *gorm.DB, attributionIds ...db.AttributionID) {
 	t.Cleanup(func() {
```
easyCZ, line 103:
```suggestion
func cleanUp(t *testing.T, conn *gorm.DB, attributionIds ...db.AttributionID) {
    t.Helper()
	t.Cleanup(func() {
```
just for general practice

components/usage/pkg/db/usage.go
```diff
@@ -43,7 +43,7 @@ type Usage struct {
 	CreditCents         CreditCents    `gorm:"column:creditCents;type:bigint;" json:"creditCents"`
 	EffectiveTime       VarcharTime    `gorm:"column:effectiveTime;type:varchar;size:255;" json:"effectiveTime"`
 	Kind                UsageKind      `gorm:"column:kind;type:char;size:10;" json:"kind"`
-	WorkspaceInstanceID uuid.UUID      `gorm:"column:workspaceInstanceId;type:char;size:36;" json:"workspaceInstanceId"`
+	WorkspaceInstanceID *uuid.UUID     `gorm:"column:workspaceInstanceId;type:char;size:36;" json:"workspaceInstanceId"`
```
easyCZ, line 46: FWIW, I'd have split this into a separate pre-PR to isolate it from the rest of the main logic in this PR.

components/usage/pkg/db/workspace_instance.go
```diff
@@ -176,6 +177,11 @@ func (a AttributionID) Values() (entity string, identifier string) {
 	return tokens[0], tokens[1]
 }
 
+func (a AttributionID) IsUser() bool {
+	entity, _ := a.Values()
+	return entity == "user"
+}
```
easyCZ, line 183: [personal] I generally prefer not to introduce Binary methods on an object which can have more than a binary number of outcomes. This forces the caller to use 
```
switch a.Kind {
  case User:
    ...
  case Team:
    ...
  default:
    ...
}
```
This tends to be more resilient over time as you don't make binary choices anymore, but are forced to acknowledge existence of other values.

easyCZ (approved):
Looking good, left a couple of questions.
/hold

We can tackle these in a follow-up if we want.

I'd also recommend isolating DB migrations away from application changes which make use of those values. When we introduced https://github.com/gitpod-io/gitpod/pull/12873 and rolled it out, we actually saw requests fail as the migration was not compatible with the application changes - while this can happen with migrations being independent as well, it makes it easier to reason about it in the given PR.

components/usage/pkg/db/cost_center.go
svenefftinge: I generally use 'Compute...' to indicate that a function doesn't have aside-effects but just computes something. But I'm happy to change the name to make it clearer for you.
I'm passing in the costCenter because it contains information about `spendingLimit` which is likely to be used pretty soon, but we can fetch it within the function if needed.

components/usage/pkg/db/cost_center.go
svenefftinge: I aligned with the naming of gorm but have no problem changing this to 'Update..'.

install/installer/pkg/config/v1/experimental/experimental.go
```diff
@@ -246,6 +246,7 @@ type UsageConfig struct {
 	Enabled                          bool               `json:"enabled"`
 	Schedule                         string             `json:"schedule"`
 	BillInstancesAfter               *time.Time         `json:"billInstancesAfter"`
+	DefaultSpendingLimit             map[string]int32   `json:"defaultSpendingLimit"`
```
MrSimonEmms, line 249: Instead of using a `map`, my preference would be for this to use the [DefaultSpendingLimit](https://github.com/gitpod-io/gitpod/blob/05c183e208aad2e57dd1cc0af85db1df9879554b/components/usage/pkg/db/cost_center.go#L41) struct.

As this is in the `experimental` section, I'll approve this on the basis that it's not for public usage, but it would keep with the consistency of using structs in the configuration

MrSimonEmms (approved):
/hold

See comment for reasons for the `hold`. Happy to discuss further if necessary

PR #12916, opened 2022-09-13
[prebuild] Log 'not using incremental prebuild' as debug

easyCZ:
## Description

The log accounts for about 40% of log messages in a given hour. Given we've added it in 45f13cf8 and haven't really used the log for anything, it feels safe to downgrade. We can always change or add metrics for this if we need it.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/server/ee/src/workspace/workspace-factory.ts
+++ b/components/server/ee/src/workspace/workspace-factory.ts
@@ -125,7 +125,7 @@ export class WorkspaceFactoryEE extends WorkspaceFactory {
                             recentPrebuild.workspace,
                         ))
                     ) {
-                        log.info({ userId: user.id }, "Not using incremental prebuild base", {
+                        log.debug({ userId: user.id }, "Not using incremental prebuild base", {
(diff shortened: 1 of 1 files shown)


jankeromnes (approved):
Looks good to me, thanks!

But, aren't we already ignoring all messages at level INFO? 🤔

easyCZ:
We're not. We ship everything INFO and more server to GCP

easyCZ:
And to me it makes sense to ship INFO. I think we could make better use of fields and context to largely end up with ~1 message per request in logs. We could achieve this by enabling logging every request on a middleware level, attach request ID and other properties from the context that we populate as we process a request and then log it on the middleware as we respond. It would provide an easy extension for request tracing in logs. 

We do have honecomb tracing but that's sampled. What we'd get here is the ability to respond with a Request ID and ask users to quote it back to us for support cases.

PR #12921, opened 2022-09-13
[stripe] Use AttributionID metadata when querying for customers

easyCZ:
## Description

Use `attributionId` when querying for customers, instead team/user Id

## Related Issue(s)

Fixes #

## How to test

1. Go to preview
2. Create team with name Gitpod-something
3. Enable billing
4. See billing enables OK
5. Check in Test Stripe account customer has `attributionId:...` set

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/server/ee/src/user/stripe-service.ts
+++ b/components/server/ee/src/user/stripe-service.ts
@@ -13,6 +13,8 @@ import { AttributionId } from "@gitpod/gitpod-protocol/lib/attribution";
 const POLL_CREATED_CUSTOMER_INTERVAL_MS = 1000;
 const POLL_CREATED_CUSTOMER_MAX_ATTEMPTS = 30;
 
+const ATTRIBUTION_ID_METADATA_KEY = "attributionId";
+
 @injectable()
 export class StripeService {
     @inject(Config) protected readonly config: Config;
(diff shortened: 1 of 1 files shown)


components/server/ee/src/user/stripe-service.ts
```diff
@@ -34,11 +36,15 @@ export class StripeService {
     }
 
     async findCustomerByUserId(userId: string): Promise<Stripe.Customer | undefined> {
-        return this.findCustomerByQuery(`metadata['userId']:'${userId}'`);
+        return this.findCustomerByQuery(
+            `metadata['${ATTRIBUTION_ID_METADATA_KEY}']:'${AttributionId.render({ kind: "user", userId })}'`,
```
svenefftinge, line 40: do you plan to manually update existing customers?

svenefftinge (approved):
Thanks for cleaning this up! Would be great to also generalize the methods in StripeService.
I.e. have `findByAttributionID` and `createCustomerForAttributionID`.

components/server/ee/src/user/stripe-service.ts
easyCZ: They've already been updated. The previous PR changed the subscription queries to set the attribution ID alongside the previous team/user id, this now removes old I'd usage

PR #12942, opened 2022-09-14
[UBP] Allow individual users to sign up for usage based pricing

andrew-farries:
## Description

Allow individual users to sign up for usage based billing via their settings page:

https://user-images.githubusercontent.com/8225907/190082718-027dbeeb-15bc-4836-9bf8-55c35f258fde.mov

Previously, signup for usage based billing was only available to teams.

This PR builds on several previous PRs to add the necessary endpoints to the server JSON-RPC API.
* https://github.com/gitpod-io/gitpod/pull/12753
* https://github.com/gitpod-io/gitpod/pull/12876
* https://github.com/gitpod-io/gitpod/pull/12864
* https://github.com/gitpod-io/gitpod/pull/12767
* https://github.com/gitpod-io/gitpod/pull/12887

## Related Issue(s)

Part of https://github.com/gitpod-io/gitpod/issues/12685 and https://github.com/gitpod-io/gitpod/issues/12684.

## How to test

1. Create a `"Gitpod-something"` team in the preview env.
2. Sign the team up for UBP.
3. Go to the billing tab on your individual user settings page.
4. Sign yourself up for UBP.
5. See the individual user created in the Stripe dashboard with a payment method and subscription.

## Release Notes

```release-note
NONE
```

## Werft options:
- [x] /werft with-preview
- [x] /werft with-payment


components/dashboard/src/hooks/useStripe.ts
```diff
@@ -0,0 +1,176 @@
...
+    boolean,
+    string | undefined,
+    boolean,
+    boolean,
+    string | undefined,
+    number | undefined,
+    (newLimit: number) => Promise<void>,
+];
+
+const useStripe = (subject: hasId | undefined, attributionId: AttributionId, localStorageKey: string): returnType => {
```
jankeromnes, line 28: Suggestion: Instead of making `localStorageKey` a parameter, how about generating it automatically based on the `attributionId`?

E.g. instead of `pendingStripeSubscriptionForUser${user.id}` / `pendingStripeSubscriptionForTeam${team.id}`, we could have a unified `pendingStripeSubscriptionFor${attributionId}` (or the rendered version I guess)

components/dashboard/src/hooks/useStripe.ts
```diff
@@ -0,0 +1,176 @@
...
+
+type returnType = [
+    boolean,
+    string | undefined,
+    boolean,
+    boolean,
+    string | undefined,
+    number | undefined,
+    (newLimit: number) => Promise<void>,
+];
```
jankeromnes, line 26: Oof, this is a bit unreadable and looks a bit error prone 🙈 I'll try to track down what all these things are (and if they're all really needed), but if we do indeed need all of these, what do you think of returning a custom shape (with clearly named attributes) instead of an array with implicit positions?

components/dashboard/src/settings/Billing.tsx
```diff
@@ -10,11 +10,13 @@ import { BillingAccountSelector } from "../components/BillingAccountSelector";
...
 import UsageBasedBillingConfig from "../components/UsageBasedBillingConfig";
 import { AttributionId } from "@gitpod/gitpod-protocol/lib/attribution";
+import useStripe from "../hooks/useStripe";
 
 export default function Billing() {
     const { user } = useContext(UserContext);
-
     const attributionId: AttributionId = { kind: "user", userId: user?.id || "" };
+    const [showSpinner, _, showUpgradeBilling, showManageBilling, stripePortalUrl, usageLimit, doUpdateLimit] =
+        useStripe(user, attributionId, `pendingStripeSubscriptionForUser${user?.id || ""}`);
```
jankeromnes, line 19: I wonder if a more "idiomatic" solution to this problem could be a `StripeBillingContext` (similar to `UserContext`, `TeamContext`, `PaymentContext` etc).

Here are the changes that would be needed to achieve this:
- Refactor `useStripe.ts` to a `stripe-billing-context.tsx` (you can look at [theme-context.tsx](https://github.com/gitpod-io/gitpod/blob/0d757c4cd34111c46ffea402a3ed1c2a8d7e60c6/components/dashboard/src/theme-context.tsx) for inspiration on the structure)
- Import and use the `StripeBillingContextProvider` in [index.tsx](https://github.com/gitpod-io/gitpod/blob/6a1f7546c01e529cb35cddc88a3b7e739c09e415/components/dashboard/src/index.tsx)
- Use it in `Billing.tsx` (i.e. here) and `TeamUsageBasedBilling.tsx` like so:

```ts
const { showSpinner, showUpgradeBilling, showManageBilling, stripePortalUrl, usageLimit, doUpdateLimit } = useContext(StripeBillingContext);
```

jankeromnes: Ah, or, alternatively, it seems to me that you're getting a bunch of common variables out of one thing (`useStripe` hook or `StripeBillingContext`) but only to feed them unchanged and unused straight to another common component (`UsageBasedBillingConfig`).

Doesn't this mean that you could "simply" get all these common variables directly inside `UsageBasedBillingConfig`, and pass just the `subject` and `attributionId` to `UsageBasedBillingConfig`?

components/dashboard/src/hooks/useStripe.ts
andrew-farries: yes, that's a better approach.

`localStorageKey` is now calculated in `UsageBasedBillingConfig` .

components/dashboard/src/hooks/useStripe.ts
andrew-farries: Hook removed.

components/dashboard/src/hooks/useStripe.ts
```diff
@@ -0,0 +1,176 @@
...
+        billingError,
+        showUpgradeBilling,
+        showManageBilling,
+        stripePortalUrl,
+        usageLimit,
+        doUpdateLimit,
+    ];
+};
+
+export default useStripe;
```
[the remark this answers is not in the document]
andrew-farries: Hook removed

components/dashboard/src/components/UsageBasedBillingConfig.tsx
```diff
@@ -16,27 +17,157 @@ import DropDown from "../components/DropDown";
...
+        })();
+    }, [location.search, subject]);
+
+    useEffect(() => {
+        setPendingStripeSubscription(undefined);
+        if (!subject) {
+            return;
+        }
+        try {
+            const pendingStripeSubscription = window.localStorage.getItem(localStorageKey);
```
svenefftinge, line 111: Why is pendingStripeSubscription a global and not local state?

svenefftinge: I see it used to be like this before this change, just wonder what the reason is.

jankeromnes: @svenefftinge Can you do a local state that survives navigations and page reloads? 👀

components/dashboard/src/components/UsageBasedBillingConfig.tsx
```diff
@@ -16,27 +17,157 @@ import DropDown from "../components/DropDown";
...
     const [showBillingSetupModal, setShowBillingSetupModal] = useState<boolean>(false);
+    const [stripeSubscriptionId, setStripeSubscriptionId] = useState<string | undefined>();
+    const [isLoading, setIsLoading] = useState<boolean>(true);
+    const [stripePortalUrl, setStripePortalUrl] = useState<string | undefined>();
+    const [pollStripeSubscriptionTimeout, setPollStripeSubscriptionTimeout] = useState<NodeJS.Timeout | undefined>();
+    const [usageLimit, setUsageLimit] = useState<number | undefined>();
+    const [pendingStripeSubscription, setPendingStripeSubscription] = useState<PendingStripeSubscription | undefined>();
+    const [billingError, setBillingError] = useState<string | undefined>();
+
+    const localStorageKey = `pendingStripeSubscriptionFor${subject?.id || ""}`;
```
jankeromnes, line 43: Micro-nit: There is a super tiny ID collision risk if a user and a team both have the same ID. It's not super realistic, but using `AttributionId.render(attributionId)` instead of `subject?.id` would prevent that.

components/dashboard/src/components/UsageBasedBillingConfig.tsx
```diff
@@ -47,6 +178,11 @@ export default function UsageBasedBillingConfig({
         <div className="mb-16">
             <h2 className="text-gray-500">Manage usage-based billing, usage limit, and payment method.</h2>
             <div className="max-w-xl flex flex-col">
+                {billingError && (
+                    <Alert className="max-w-xl mb-4" closable={false} showIcon={true} type="error">
```
jankeromnes, line 182: Good idea moving this alert here, so that it's encapsulated in the common component. 👍

However, the margins need to be adjusted:

| BEFORE | AFTER |
| --- | --- |
| Screenshot 2022-09-15 at 09 21 26 | Screenshot 2022-09-15 at 09 23 14 |

```suggestion
                    <Alert className="max-w-xl mt-2" closable={false} showIcon={true} type="error">
```

jankeromnes (approved):
Many thanks! Code looks good to me, and works as intended. 🎯

/hold because of the margin fix

components/dashboard/src/components/UsageBasedBillingConfig.tsx
jankeromnes: Also, this code seems to use `AttributionId.render(attributionId)` (the string) more often than `attributionId` (the object). If that's true, maybe this code could be simplified a bit by refactoring `attributionId` to be a string and not a fully-hydrated shape. 💭 No strong opinion here though.

components/dashboard/src/components/UsageBasedBillingConfig.tsx
svenefftinge: Of course not. Why does it need to survive that?

jankeromnes: We've seen that users like to reload billing/subscription pages often.

The currently-persisted "pending subscription" allows the following UX:
1. Enter credit card details
2. They go through, we're subscribing you to Stripe
3. You reload the page -- we're still subscribing you to Stripe and waiting for confirmation

Without the persistence, it changes to:
1. Enter credit card details
2. They go through, we're subscribing you to Stripe
3. You reload the page -- it looks like it didn't work, so maybe you immediately try to add your credit card again

components/dashboard/src/components/UsageBasedBillingConfig.tsx
andrew-farries: I've applied both of these suggestions, thanks 👍

PR #12961, opened 2022-09-14
[usage] Track last compelted ledger job time

easyCZ:
## Description

Adds a Gauge metric which tracks the last time we completed a scheduled ledger reconciliation. We'll use this metric to setup an alert for when we haven't completed a run succesfully for more than 1h (which could cause us to have gaps in the data).

This metric is setup the same way as other standard kube metrics- [see example](https://grafana.gitpod.io/explore?orgId=1&left=%7B%22datasource%22:%22P4169E866C3094E38%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22kube_job_status_start_time%7Bjob_name%3D%5C%22cert-manager-startupapicheck%5C%22%7D%22%7D%5D,%22range%22:%7B%22from%22:%22now-7d%22,%22to%22:%22now%22%7D%7D)

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage/pkg/scheduler/job.go
+++ b/components/usage/pkg/scheduler/job.go
@@ -5,12 +5,9 @@
 package scheduler
 
 import (
-	"context"
 	"fmt"
 	"github.com/gitpod-io/gitpod/common-go/log"
-	v1 "github.com/gitpod-io/gitpod/usage-api/v1"
 	"github.com/robfig/cron"
-	"google.golang.org/protobuf/types/known/timestamppb"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


components/usage/pkg/scheduler/job.go
```diff
@@ -37,51 +34,6 @@ func NewPeriodicJobSpec(period time.Duration, id string, job Job) (JobSpec, erro
 	}, nil
 }
 
-func NewLedgerTriggerJobSpec(schedule time.Duration, job Job) (JobSpec, error) {
```
easyCZ, line 40: This is moved into separate file to make navigation a bit easier.

components/usage/pkg/scheduler/ledger_job.go
```diff
@@ -0,0 +1,63 @@
...
+
+type LedgerJob struct {
+	usageClient   v1.UsageServiceClient
+	billingClient v1.BillingServiceClient
+}
+
+func (r *LedgerJob) Run() (err error) {
+	defer func() {
+		reportLedgerCompleted(err)
+	}()
```
easyCZ, line 35: This is the only change in logic - reporting the metric

andrew-farries (approved):

PR #12963, opened 2022-09-14
[server] Every time we stop a workspace instance, log the reason why

jankeromnes:
## Description

Every time we stop a workspace instance, log the reason why. This should help with debugging.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/12283

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/server/src/user/user-deletion-service.ts
+++ b/components/server/src/user/user-deletion-service.ts
@@ -113,6 +113,10 @@ export class UserDeletionService {
             runningWorkspaces.map(async (info) => {
                 const wsi = info.latestInstance;
 
+                log.info({ workspaceId: info.workspace.id, instanceId: wsi.id }, "Stopping workspace instance", {
+                    reason: "deleting user",
+                });
+
                 const req = new StopWorkspaceRequest();
                 req.setId(wsi.id);
                 req.setPolicy(StopWorkspacePolicy.NORMALLY);
--- a/components/server/src/workspace/gitpod-server-impl.ts
+++ b/components/server/src/workspace/gitpod-server-impl.ts
@@ -748,7 +748,7 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
             await this.guardAccess({ kind: "workspace", subject: workspace }, "get");
         }
 
-        this.internalStopWorkspace(ctx, workspace).catch((err) => {
+        this.internalStopWorkspace(ctx, workspace, "stopped via API").catch((err) => {
             log.error(logCtx, "stopWorkspace error: ", err);
         });
     }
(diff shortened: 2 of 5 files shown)


components/server/src/user/user-deletion-service.ts
```diff
@@ -113,6 +113,12 @@ export class UserDeletionService {
             runningWorkspaces.map(async (info) => {
                 const wsi = info.latestInstance;
 
+                const reason = "deleting user";
+                log.info(
+                    { workspaceId: info.workspace.id, instanceId: wsi.id },
+                    `Stopping workspace instance (reason: ${reason})`,
```
geropl, line 119: :bulb: Could also be a payload, so it's easier to write log queries for it:

`..., "Stopping workspace instance", { reason });`

components/server/src/workspace/gitpod-server-impl.ts
```diff
@@ -748,14 +748,15 @@ export class GitpodServerImpl implements GitpodServerWithTracing, Disposable {
...
-        this.internalStopWorkspace(ctx, workspace).catch((err) => {
+        this.internalStopWorkspace(ctx, workspace, "stopped via API").catch((err) => {
             log.error(logCtx, "stopWorkspace error: ", err);
         });
     }
 
     protected async internalStopWorkspace(
         ctx: TraceContext,
         workspace: Workspace,
+        reason: string,
```
geropl, line 759: :cloud: Maybe this should be optional? :thinking:

components/server/src/workspace/workspace-starter.ts
```diff
@@ -439,8 +439,12 @@ export class WorkspaceStarter {
         ctx: TraceContext,
         instanceId: string,
         instanceRegion: string,
+        reason: string,
```
geropl, line 442: same as above: maybe optional? :thinking:

components/server/src/workspace/workspace-starter.ts
```diff
@@ -439,8 +439,12 @@ export class WorkspaceStarter {
...
         policy?: StopWorkspacePolicy,
     ): Promise<void> {
+        ctx.span?.setTag("stopWorkspaceReason", reason);
+        log.info({ instanceId }, `Stopping workspace instance (reason: ${reason})`);
```
geropl, line 446: same as [above](https://github.com/gitpod-io/gitpod/pull/12963/files#r971722712).

geropl (approved):
Thx @jankeromnes , LGTM.

/hold In case you want to tackle the nits, and to give @sagor999 the possibility to review as well.

components/server/src/workspace/gitpod-server-impl.ts
jankeromnes: Why make it optional if we always want to know the reason a workspace was stopped?

sagor999 (approved):
:heart:

jankeromnes:
Thanks for the reviews!

Nits addressed, go for launch 🚀

/unhold

Issue #12964, opened 2022-09-14
AWS Terraform Ref Arch is too permissive


mrzarquon:
### Bug description

Inheriting from our old EKS guide, we still set `"arn:aws:iam::aws:policy/AmazonS3FullAccess"` on the S3 user accounts, which grants global S3 access to the user.

There is no reason to do this now that we support specifying an S3 bucket for an object store. And it could cause this creation to fail in the instances where the user running the terraform code doesn't have permissions to grant this level of access.

### Steps to reproduce

See above

PR #12968, opened 2022-09-14
Correct AWS Terraform arch to contain correct S3 permissions

mrzarquon:
## Description
When I removed the global access to fix https://github.com/gitpod-io/gitpod/issues/12964 it exposed the fact that our IAM permissions weren't correct to just use the S3 bucket.

```release-note
Attaches correct permissions to the IAM user to have correct level of access to just the single S3 bucket we create for that user.
```


MrSimonEmms (approved):

nandajavarma:
This branch might need a rebase with `main` @mrzarquon

Issue #12970, opened 2022-09-14
Impossible to add some GitLab repositories as a Project when you're a maintainer of more than 100 repositories


jankeromnes:
### Bug description

When you want to add a repository as a new Project in Gitpod, Gitpod first gets the list of all the possible repositories you could add (grouped by account), and then lets you find the repository you want to add via scrolling or searching/filtering.

Unfortunately, in the case of GitLab, we only fetch the first 100 repositories that you could add:

https://github.com/gitpod-io/gitpod/blob/1d52870bfe3dbb675ae0028c2e1d7a441d2a1489/components/server/ee/src/gitlab/gitlab-app-support.ts#L54

This is a problem, because if you want to add a specific repository that is not present in the first 100 repositories returned by GitLab, you will never be able to see it or find it in the Gitpod UI -- thus adding it as a project is totally impossible.

### Steps to reproduce

1. Be a maintainer of more than 100 GitLab repositories
2. Try to create a new project for a repository that is not listed in the first 100 results returned by GitLab

### Expected behavior

- You should always be able to create a new project for a GitLab repository that you're a maintainer of (even if it's not in the first 100 results from GitLab)

### Anything else?

Possible solutions:
- Paginate the GitLab API call in order to fetch more than just the first 100 repositories
- Allow creating projects by manually entering a repository URL

gtsiolis:
Cross-posting in case it helps, there’s support for [offset-based pagination](https://docs.gitlab.com/ee/api/index.html#offset-based-pagination) which we could use to increase the page size here. But maybe this is not available through the API? 🤷 

For example:

```
https://gitlab.com/api/v4/projects?min_access_level=40
https://gitlab.com/api/v4/projects?min_access_level=40&per_page=2000
```

See [relevant discussion](https://gitpod.slack.com/archives/C02EN94AEPL/p1663172660241819?thread_ts=1663140993.635109&cid=C02EN94AEPL) (internal).

jankeromnes:
There is actually one workaround: Using the browser console to create a Project manually.

Say I want to add my repository https://gitlab.com/jankeromnes/test-before as a Project to one of my Teams, but for some reason, Gitpod's UI won't show it to me. Then, I can still do this:

1. Run `await window._gp.gitpodService.server.getTeams()` in order to find the team ID under which I want to create the Project
2. Then, run the following (just replace the `name`, `slug`, `cloneUrl` and `teamId` -- `appInstallationId` is useless but the API requires it):

```js
[...]

PR #12971, opened 2022-09-14
[dashboard][protocol] Remove unused params from CreateProjectParams

jankeromnes:
## Description

While reverse-engineering our `createProject` flow to identify [this workaround](https://github.com/gitpod-io/gitpod/issues/12970#issuecomment-1247024832), I noticed that we were populating two fields in `CreateProjectParams` that are not even read or stored anywhere. Thus, we can save ourselves the trouble.

## Related Issue(s)

Fixes #

## How to test

1. Creating Projects still works

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


geropl (approved):
Tested and LGTM - thx Jan :broom: :upside_down_face:

PR #12980, opened 2022-09-15
Redact the rest values once the previous one matches redact fields

jenting:
## Description

Redact the rest values once the previous one matches redact fields

## Related Issue(s)

Fixes https://github.com/gitpod-io/security/issues/64

## How to test

Check the unit test.
Check the issue to see how to reproduce and test it.

## Release Notes

```release-note
None
```

## Documentation

None

## Werft options:

- [ ] /werft with-preview


utam0k (approved):

svenefftinge (approved):
It seems to work but comes with a pretty hard-to-understand logic (at least for me 😇 ).
Adding some more tests would help. The core problem is that we want to redact based on property name not the value (as https://github.com/cloudfoundry/lager/blob/master/json_redacter.go does). So when we have properties like envs that are described on meta-level (name, value) it doesn't work so well anymore. This change "hacks" it in a way that it just redacts the rest of object (alphabetical property order), which might cause unwanted effects. 

I.e this test case would pass:
```
`{"env":[{"another":"value","name":"secret","specialproperty":"foo"}]}`,
`{"env":[{"another":"value","name":"[redacted]","specialproperty":"[redacted]"}]}`,
```

I'm still approving as it solves the issue.

svenefftinge:
> alphabetical property order

I think it is not alphabetical order but arbitrary as I've seen this test fail occasionally.

jenting:
> > alphabetical property order
> 
> I think it is not alphabetical order but arbitrary as I've seen this test fail occasionally.

Thanks for the heads up. Yes, the order is randomly not alphabetical. I will file a new PR to address it.

PR #12990, opened 2022-09-15
Add .gitattributes to mark installer golden files as generated

easyCZ:
## Description

This will automatically collapse the `*.golden` files in PRs and mark them as auto-generated.
Because it's causing PRs to be messy with changes which are auto-generated

See [docs](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github)

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+install/installer/cmd/testdata/render/*.golden linguist-generated=true


svenefftinge (approved):
Nice! Would love to see this for other generated code as well

easyCZ:
Currently, we don't have a deterministic pattern to use to mark all generated code (from grpc for example). in theory, we could take `components/*-api/go|typescript|...` as the path but in some cases, there's non-generated code in those packages also (see `components/public-api/go/config`).

I think a reasonable change we could make is to move generated code into a `generated/` prefixed directory, such as `components/public-api/go/generated/v1/...` which would then allow us to have a single pattern to do this.

easyCZ:
I've found a pattern that more-less works - https://github.com/gitpod-io/gitpod/pull/13078

PR #12992, opened 2022-09-15
[dashboard] Send error body with feedback

### LINKED ISSUE #10402: Send error message with feedback

laushinka:
## Related Issue(s)

Fixes #10402

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


components/dashboard/src/Analytics.tsx
```diff
@@ -40,6 +41,8 @@ export interface TrackFeedback {
     feedback: string;
     href: string;
     path: string;
+    error_object?: StartWorkspaceError;
+    error_message?: string;
```
andrew-farries, line 45: Is there a reason these props have to be snake-cased rather than camel-cased?

laushinka: Not really! Before it was `error_rendered` which I confirmed with @jakobhero. Let me see what the convention is in Segment.

laushinka: I see it's a mix 😅 Some are `ui_experiments`, `button_type`, while others are `workspaceInstanceId ` or `clientKind`.

laushinka: Asked [here](https://gitpod.slack.com/archives/C01KPEPKX0A/p1663243627662879)

andrew-farries (changes requested):

components/dashboard/src/Analytics.tsx
andrew-farries: Sorry, I thought this interface defined props for some component, but it doesn't 😊 It's fine how it is if Segment needs it that way.

components/dashboard/src/feedback-form/FeedbackComponent.tsx
```diff
@@ -35,6 +38,8 @@ function FeedbackComponent(props: {
                 feedback: text,
                 href: window.location.href,
                 path: window.location.pathname,
+                error_object: props.error_object || undefined,
+                error_message: props.error_message,
```
andrew-farries, line 42: Not introduced in this PR, but this object can be explicitly typed as `EventProperties` .

components/dashboard/src/feedback-form/FeedbackComponent.tsx
```diff
@@ -10,13 +10,16 @@ import happy from "../images/feedback/happy-emoji.svg";
...
+import { StartWorkspaceError } from "../start/StartPage";
 
 function FeedbackComponent(props: {
     onClose?: () => void;
     isModal: boolean;
     isError: boolean;
     message?: string;
     initialSize?: number;
+    error_object?: StartWorkspaceError;
+    error_message?: string;
```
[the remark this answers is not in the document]
laushinka: Apparently Segment converts these to snake case anyway so we can do it this way.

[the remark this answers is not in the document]
andrew-farries: I'd still change the props to `errorMessage` and `errorObject` because that's the convention for prop names in React.

andrew-farries (changes requested):

components/dashboard/src/feedback-form/FeedbackComponent.tsx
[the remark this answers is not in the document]
laushinka: I committed your suggestion but then it disappeared. GitHub confuses me sometimes

PR #12997, opened 2022-09-15
[usage] use credit balance for limit check

svenefftinge:
## Description
For checking the usage limit, instead of checking against the current stripe invoice we want to check against the current credit balance.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/server/ee/src/billing/billing-service.ts
+++ b/components/server/ee/src/billing/billing-service.ts
@@ -7,7 +7,6 @@
 import { User } from "@gitpod/gitpod-protocol";
 import { AttributionId } from "@gitpod/gitpod-protocol/lib/attribution";
 import { log } from "@gitpod/gitpod-protocol/lib/util/logging";
-import { BillingServiceClient, BillingServiceDefinition } from "@gitpod/usage-api/lib/usage/v1/billing.pb";
 import { UsageServiceClient, UsageServiceDefinition } from "@gitpod/usage-api/lib/usage/v1/usage.pb";
 import { inject, injectable } from "inversify";
 import { UserService } from "../../../src/user/user-service";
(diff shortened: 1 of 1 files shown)


components/server/ee/src/billing/billing-service.ts
```diff
@@ -43,8 +40,13 @@ export class BillingService {
             };
         }
 
-        const upcomingInvoice = await this.billingService.getUpcomingInvoice(attributionId);
-        const currentInvoiceCredits = upcomingInvoice.credits | 0;
+        const now = new Date();
+        const currentBalance = await this.usageService.listUsage({
```
easyCZ, line 44: We could also specify Page Size 1 or 0 to avoid getting any results in the list response, we don't need it

easyCZ (approved):
/hold if you want to set the page size

components/server/ee/src/billing/billing-service.ts
svenefftinge: I use `from: now, to: now` which does also make sure we don't fetch usage entries

easyCZ: Fair. We could change the usage implementation that if page size is 0, we don't even make the (list) query in the first place which would then be a bit faster.

svenefftinge: I'd rather introduce a new function theres too much information in the response that wouldn't make sense (e.g. balanceAtStart vs balanceAtEnd)

svenefftinge: your proposal makes sense once we need balance for a period but are not interested in the entries.

PR #12998, opened 2022-09-15
Move tsc to validate changes. Run validations in parallel

mads-hartmann:
## Description

To declutter the Werft phases for the build job I've moved the `tsc` check to the "Validating changes" phase. I also changed the implementation of `typecheckWerftJobs` and `preCommitCheck` so they are properly async, which makes it possible to run them in parallel (achieved through `Promise.all`)

I also had to make the actual build execution use `await` and add `exec(.., { async: true })` as traces otherwise broke. I'm not a 100% certain why this is needed - my hunch is that by running the build in a promise we're allowing other code to temporarily execute on the event loop which takes care of some trace management - as the change is harmless I'm okay keeping it in this PR and then we can investigate why the traces breaks without it some other time.

## Related Issue(s)

No issue, just noticed this when working on other things

## How to test

Here's a screenshot before/after that shows the slice has been moved to the validating phase (resulting in less clutter IMO)

Screenshot 2022-09-15 at 14 29 11

Screenshot 2022-09-15 at 14 29 26

And here's a screenshot of the trace before/after that shows that the slices are now running in parallel (not that there is a huge speed increase as the precommit hook takes much longer than tsc --noEmit, but nonetheless) 

Screenshot 2022-09-15 at 14 29 52

Screenshot 2022-09-15 at 14 29 35

```
werft job run github
```

## Release Notes

```release-note
NONE
```

## Documentation

N/A

## Werft options:

- [ ] /werft with-preview


.werft/jobs/build/validate-changes.ts
```diff
@@ -5,8 +5,7 @@ import { JobConfig } from "./job-config";
 export async function validateChanges(werft: Werft, config: JobConfig) {
     werft.phase("validate-changes", "validating changes");
     try {
-        branchNameCheck(werft, config);
-        preCommitCheck(werft);
+        await Promise.all([branchNameCheck(werft, config), preCommitCheck(werft), typecheckWerftJobs(werft)]);
```
ArthurSens, line 8: That's really nice, I bet there are more places that we could do the same thing 🙂

ArthurSens (approved):
Nice 💅

PR #13022, opened 2022-09-16
[baseserver] Add metric with version of the server, use in public api and usage

easyCZ:
## Description

To be able to track which version we're on in our dashboards. Will produce the following metric:

```
gitpod_server_version{version="<version>"} 1
```

Once validated, I can also add this to other instances of baseserver

The approach to the metric with version is standard, see [docs](https://www.robustperception.io/exposing-the-software-version-to-prometheus/)

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage/cmd/run.go
+++ b/components/usage/cmd/run.go
@@ -37,7 +37,7 @@ func run() *cobra.Command {
 				log.WithError(err).Fatal("Failed to get config. Did you specify --config correctly?")
 			}
 
-			err = server.Start(cfg)
+			err = server.Start(cfg, Version)
 			if err != nil {
 				log.WithError(err).Fatal("Failed to start usage server.")
 			}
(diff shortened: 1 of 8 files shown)


components/usage/cmd/run.go
```diff
@@ -37,7 +37,7 @@ func run() *cobra.Command {
 				log.WithError(err).Fatal("Failed to get config. Did you specify --config correctly?")
 			}
 
-			err = server.Start(cfg)
+			err = server.Start(cfg, Version)
```
andrew-farries, line 40: I see that this `Version` is set during build, but how and where it is set and what is the format of the version string?

andrew-farries (approved):

components/usage/cmd/run.go
easyCZ: I've observed either `commit-<hash>` or `main.<build-id>`

It's set by the CI job when it builds the images

PR #13042, opened 2022-09-16
Ensure error message for failed slices stay within the slice

mads-hartmann:
### Background

The Werft log cutter expects each line of output to follow the specific format ([docs](https://github.com/csweichel/werft#log-cutting)). If no slice is specified in a log line then it will be attached to the "default" slice for that phase (which is a slice with the same id as the phase).

As we're just `console.log`'ing strings it means it's out responsibility to ensure we don't log multiline strings.

### The change

This PR improves our handling of slices that we explicitly fail using `werft.fail` in two  important ways:

1. Previously it would not handle multi-line error messages well which meant that the error message for a slice might get split between the slice that failed and the default slice for the phase.
2. `werft.fail` throws an error which will eventually bubble up to our main `catch` handler. In that handler we would `console.log` the error which meant that the error would always end up in the default slice (see screenshot below)

With this PR any place where we use `werft.fail` we'll now ensure the error is contained to the slice, which makes it much easier to understand why a slice failed.

### Implementation details

I have introduced a new custom error type `SliceFailedError` which we use in `werft.fail`. In the main `catch` handler we then don't perform any additional logging. 

Some other notes:

- it appears that when you use `[<slice>|FAIL] <reason>` Werft won't display the reason. Because of that I updated the code so the last log output for the slice is "Failed. Expand to see why" and I have left the fail reason blank.
- You can't `fail` or `done` a phase, only slices. I've updated `validate-changes.ts` so it doesn't try to do so. There are still many other places where we can make similar changes.

## Related Issue(s)

No issue. Friday Tid(a)y

## How to test

To test this I introduced a TS error in `build.ts` and side-loaded the error to ensure that the "tsc --noEmit" slice would fail. I did this on this branch and on a branch off main to show the the difference.

```
werft job run github -s .werft/build.ts
```

Before (branch off main)

Screenshot 2022-09-16 at 20 54 03

After (this branch)

Screenshot 2022-09-16 at 20 56 02

## Release Notes

```release-note
NONE
```

## Documentation

N/A

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/.werft/build.ts
+++ b/.werft/build.ts
@@ -1,6 +1,6 @@
 import * as fs from "fs";
 import { SpanStatusCode } from "@opentelemetry/api";
-import { Werft } from "./util/werft";
+import { FailedSliceError, Werft } from "./util/werft";
 import { reportBuildFailureInSlack } from "./util/slack";
 import * as Tracing from "./observability/tracing";
 import * as VM from "./vm/vm";
@@ -27,7 +27,12 @@ Tracing.initialize()
             message: err,
         });
 
-        console.log("Error", err);
+        if (err instanceof FailedSliceError) {
+            // This error was produced using werft.fail which means that we
+            // already handled it "gracefully"
+        } else {
+            console.log("Error", err);
+        }
 
         if (context.Repository.ref === "refs/heads/main") {
             reportBuildFailureInSlack(context, err).catch((error: Error) => {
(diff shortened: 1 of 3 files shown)


ArthurSens (approved):
Awesome change. Multiline messages getting spread over slices has bothered me since forever!!

PR #13063, opened 2022-09-19
[usage] remove unused GetUpcomingInvoice

svenefftinge:
## Description
Removes no longer needed `getUpcomingInvoice` API.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage-api/go/v1/billing.pb.go
+++ b/components/usage-api/go/v1/billing.pb.go
@@ -100,158 +100,6 @@ func (*ReconcileInvoicesResponse) Descriptor() ([]byte, []int) {
 	return file_usage_v1_billing_proto_rawDescGZIP(), []int{1}
 }
 
-type GetUpcomingInvoiceRequest struct {
-	state         protoimpl.MessageState
-	sizeCache     protoimpl.SizeCache
-	unknownFields protoimpl.UnknownFields
-
-	// Types that are assignable to Identifier:
-	//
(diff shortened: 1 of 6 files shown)


easyCZ (commented):
There are existing calls to this service, I believe we can only remove once there are no more calls to this service (and in different chagesets).

Screenshot 2022-09-19 at 8 31 47

[Grafana](https://grafana.gitpod.io/d/fDH4_QMVk/grpc-server?orgId=1&var-service=usage.v1.BillingService&var-method=GetUpcomingInvoice&var-type=All)

In an incident on Friday September 16, 2022, we've had to patch version of server. Had we removed the implementation of `SetBilledInstance` in the same changeset, server would now fail on starting workspaces. The same applies here, we need to remove all usage before we land the removal.

svenefftinge:
I have removed the call site in https://github.com/gitpod-io/gitpod/pull/12997 on Friday. 
So once we have deployed that, we should be able to land this one.

/hold

easyCZ (approved):
There are no [more API calls ](https://grafana.gitpod.io/d/fDH4_QMVk/grpc-server?orgId=1&var-service=usage.v1.BillingService&var-method=GetUpcomingInvoice&var-type=All&from=now-30m&to=now)after the last rollout.

PR #13108, opened 2022-09-20
[server][github] fix file provider for self-managed GHE

### LINKED ISSUE #9253: Adjust URL Being Used To Fetch .gitpod.yml Files From GitHub Enterprise Instances
### Bug description

I currently have a self-hosted instance deployed which appears to be ignore the `image: image-name` line in my `.gitpod.yml` file within a repository. For example in my `.gitpod.yml` I've tried:

```yaml
image: node:16.14-buster
```

or

```yaml
image: jimmyb/workspace-laravel
```

but every time the `gitpod/workspace-full` image is loaded. I captured some logs from the build process and as you can see it's pulling the `workspace-full` image - https://ghostbin.linuxbox.ninja/paste/6jyt8

I've tested this on gitpod.io and I cannot replicate the issue there. @MrSimonEmms was also unable to replicate the issue on his self-hosted instance so it sounds like an issue specific to my instance.

Last night I tore down my Kubernetes cluster and rebuilt it from scratch, but the same issue occurs. I am currently on `main.2860` which I updated to from `main.2847` (where the issue existed for me) to no avail.

Is there some sort of setting I can look at that would force the `workspace-full` image to be used? I know if there is no `.gitpod.yml` or image set it will default to the `workspace-full` image, but that's definitely not the case here, I am setting the image to be used, it's just being ignored.

To further troubleshoot, I set my `.gitpod.yml` up as:

```yaml
image: jimmyb/workspace-laravel

vscode:
  extensions:
    - esbenp.prettier-vscode
```

I started a new workspace, it ignored the `image` line and used the `workspace-full` image, but it did install the extension. So it seems it's not completely ignoring the `.gitpod.yml` file.

My cluster is built from k3s, using 3 Ubuntu 20.04.x VMs (1 controller, 2 workers). If any other information about the cluster is needed, please don't hesitate to ask!

Any help on where to look next would be greatly appreciated!

### Steps to reproduce

So here's the thing, I/we can't seem to reproduce this 😩

### Workspace affected

_No response_

### Expected behavior

I would expect that the image specified in my `.gitpod.yml` file be used for my workspace instead of the `gitpod/workspace-full` image.

### Example repository

_No response_

### Anything else?

_No response_

AlexTugarev:
## Description
This PR removes URL construction for reading file contents, instead rely on supported REST API. 

## Related Issue(s)

Fixes #9253

## How to test
Though this claims to fix an GHE issue, it can be verified with GitHub SaaS as well, because the core issue was that we'd used to craft the download URL, but here we're changing to supported means of fetching contents via REST API. That said, if should be sufficient to test .gitpod.yml being read from any GitHub repo in a prebuild of a project.

## Release Notes

```release-note
Fix reading .gitpod.yml for self-managed GHE instances.
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/server/src/github/api.ts
+++ b/components/server/src/github/api.ts
@@ -156,7 +156,7 @@ export class GitHubRestApi {
     }
 
     protected get userAgent() {
-        return new URL(this.config.oauth!.callBackUrl).hostname;
+        return (this.config.oauth && new URL(this.config.oauth?.callBackUrl)?.hostname) || "GitPod unknown";
     }
 
     /**
--- a/components/server/src/github/file-provider.ts
+++ b/components/server/src/github/file-provider.ts
@@ -8,12 +8,11 @@ import { injectable, inject } from "inversify";
 
 import { FileProvider, MaybeContent } from "../repohost/file-provider";
 import { Commit, User, Repository } from "@gitpod/gitpod-protocol";
-import { GitHubGraphQlEndpoint, GitHubRestApi } from "./api";
+import { GitHubRestApi } from "./api";
 import { log } from "@gitpod/gitpod-protocol/lib/util/logging";
 
 @injectable()
 export class GithubFileProvider implements FileProvider {
-    @inject(GitHubGraphQlEndpoint) protected readonly githubGraphQlApi: GitHubGraphQlEndpoint;
     @inject(GitHubRestApi) protected readonly githubApi: GitHubRestApi;
 
     public async getGitpodFileContent(commit: Commit, user: User): Promise<MaybeContent> {
@@ -56,14 +55,27 @@ export class GithubFileProvider implements FileProvider {
         }
 
         try {
-            const contents = await this.githubGraphQlApi.getFileContents(
-                user,
-                commit.repository.owner,
-                commit.repository.name,
-                commit.revision,
-                path,
+            const response = await this.githubApi.run(user, (api) =>
+                api.repos.getContent({
+                    owner: commit.repository.owner,
+                    repo: commit.repository.name,
+                    path,
+                    ref: commit.revision,
+                    headers: {
+                        accept: "application/vnd.github.VERSION.raw",
+                    },
+                }),
             );
-            return contents;
+            if (response.status === 200) {
+                if (typeof response.data === "string") {
+                    return response.data;
+                }
+                console.warn("GithubFileProvider.getFileContent – unexpected response type.", {
+                    headers: response.headers,
+                    type: typeof response.data,
+                });
+            }
+            return undefined;
         } catch (err) {
             log.debug(err);
         }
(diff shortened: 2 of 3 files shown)


geropl:
@AlexTugarev During testing, while creating a project from my personal repo in GH, I see this error message:
image

As a result, the project seems to be there, but it does not receive any prebuild triggers... any idea why? :thinking:

AlexTugarev:
> As a result, the project seems to be there, but it does not receive any prebuild triggers... any idea why? 🤔

Unfortunately, not. 

I tried to reproduce with a fork of that repo, but it worked fine. 

@geropl, please repeat, as I just enabled debug logging for server. 

/hold

Because of this report, and I'm not able to tell if this is unrelate, though it seems. And because I forgot to remove dead code.

geropl:
@AlexTugarev seeing this in the logs:
```
{
  "@type": "type.googleapis.com/google.devtools.clouderrorreporting.v1beta1.ReportedErrorEvent",
  "serviceContext": {
    "service": "server",
    "version": "alex-adjust-url-being-used-to-fetch-9253.0"
  },
  "stack_trace": "GitHubApiError: GitHub API Error. Status: 422
    at GitHubRestApi.<anonymous> (/app/node_modules/@gitpod/server/dist/src/github/api.js:198:27)
    at Generator.throw (<anonymous>)
    at rejected (/app/node_modules/@gitpod/server/dist/src/github/api.js:20:65)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)",
  "component": "server",
  "severity": "ERROR",
  "time": "2022-09-20T12:33:47.225Z",
  "context": { "userId": "5c510231-73e4-4650-993d-644ba9b21745" },
  "message": "Request createProject failed with internal server error",
  "error": "GitHubApiError: GitHub API Error. Status: 422
    at GitHubRestApi.<anonymous> (/app/node_modules/@gitpod/server/dist/src/github/api.js:198:27)
    at Generator.throw (<anonymous>)
    at rejected (/app/node_modules/@gitpod/server/dist/src/github/api.js:20:65)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)",
  "payload": {
    "method": "createProject",
    "args": [
      {
        "name": "gitpod-test-repo",
        "slug": "gitpod-test-repo",
        "cloneUrl": "https://github.com/geropl/gitpod-test-repo.git",
        "teamId": "5b65b970-9b57-4857-a6f8-3c953f6d04d7",
        "appInstallationId": "undefined"
      },
      { "_isCancelled": false }
    ]
  }
}
```

AlexTugarev:
@geropl, could you please try again and use a different repo? I guess your test repo had see a lot of webhooks already 🤷🏻‍♂️ 

The error event is likely fired from there, thus not related to this PR. Let's tell the issues apart 🙏🏻 

Screen Shot 2022-09-20 at 16 00 52

geropl (approved):
With different repository I see prebuilds working as expected

/hold in case any other check is required :upside_down_face:

Issue #11968, opened 2022-08-08
Allow Users to Manage Workspace Snapshots


jimmybrancaccio:
## Is your feature request related to a problem? Please describe

As it seems to have implemented a limit (#11423) to the amount of snapshots a workspace can have we should implement a tool that allows user to manage (delete, get share URL, etc...) the snapshots of workspaces.

## Describe the behaviour you'd like

I would like to have a new page in the Gitpod workspaces UI (https://gitpod.io/workspaces) that allows users to manage snapshots for their workspaces. Perhaps getting to this page could be done by clicking on the 3 dots of the workspace and it would be a new selectable option on the dropdown menu.

## Describe alternatives you've considered

There are none that I can think of. I am not even sure how to manage them with access to the GCE bucket.

## Additional context

N/A

mbrevoort:
In summary, as a user, I should be able to:
* access a list of all snapshots URLs that I generated 
* delete a snapshot URL, invalidating it

Today it's possible to delete a workspace and still be able to access the snapshot URL. Is this desired? If not, snapshot URLs should be invalidated immediately when a workspace is deleted. If this is desired, as a user, I should be able to see a consolidated list of snapshot URLs across all workspaces (even those deleted) to manage them.

csweichel:
Today users can invalidate snapshots [by deleting their workspace](https://github.com/gitpod-io/gitpod/blob/sefftinge/supervisor-processes-13365/components/server/src/workspace/workspace-factory.ts#L64-L72).

PR #13121, opened 2022-09-20
[server] Always GC snapshots for workspaces

easyCZ:
## Description

Those snapshots would not be accessible anyway after we've garbage collected the Workspace.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/server/src/workspace/workspace-deletion-service.ts
+++ b/components/server/src/workspace/workspace-deletion-service.ts
@@ -69,8 +69,7 @@ export class WorkspaceDeletionService {
         const span = TraceContext.startSpan("garbageCollectWorkspace", ctx);
 
         try {
-            const deleteSnapshots = ws.softDeleted === "user";
-            const successfulDeleted = await this.deleteWorkspaceStorage({ span }, ws, deleteSnapshots);
+            const successfulDeleted = await this.deleteWorkspaceStorage({ span }, ws, true);
             await this.db.trace({ span }).updatePartial(ws.id, { contentDeletedTime: new Date().toISOString() });
             return successfulDeleted;
         } catch (err) {


components/server/src/workspace/workspace-deletion-service.ts
```diff
@@ -69,8 +69,7 @@ export class WorkspaceDeletionService {
         const span = TraceContext.startSpan("garbageCollectWorkspace", ctx);
 
         try {
-            const deleteSnapshots = ws.softDeleted === "user";
```
easyCZ, line 72: Look at git history, this has been there since the initial gitpod public repo commit.

geropl:
> Those snapshots would not be accessible anyway after we've garbage collected the Workspace.

Note that this is not true: As long as you have a snapshot link is still expected to work. :confused: 

A complete fix for this would be to promote the "snapshot" concept to be user-manageable: https://github.com/gitpod-io/gitpod/issues/11968

:bulb: Until then, we could try to work-around: we have the concept of pinning a workspace. Those are excluded from GC. Whenever we `createSanpshot`, we could pin the underlying workspace.
We could even think about marking it as "pinnedForSnapshot", for maybe that's not necessary.

svenefftinge:
In agree with @geropl that we need to surface snapshot handling more in the dashboard, but as it stands this change seem to make sense, as we can't start workspaces for snapshots where the workspace was deleted [[code](https://github.com/gitpod-io/gitpod/blob/mp/server-gc-workspace-storage/components/server/src/workspace/workspace-factory.ts#L115-L123)]. So whould we merge this change and continue the more general discussion in #11968?

geropl:
> as we can't start workspaces for snapshots where the workspace was deleted [[code](https://github.com/gitpod-io/gitpod/blob/mp/server-gc-workspace-storage/components/server/src/workspace/workspace-factory.ts#L115-L123)]

Whoops, that's news to me. Thanks for digging it out!

@easyCZ let's un-hold :+1:

jankeromnes (approved):
After the insightful discussion, this code change looks good to me, thanks! (Had already started to review yesterday, but held off due to being uncertain about snapshot vs deleted workspace behavior.)

> > as we can't start workspaces for snapshots where the workspace was deleted [[code](https://github.com/gitpod-io/gitpod/blob/mp/server-gc-workspace-storage/components/server/src/workspace/workspace-factory.ts#L115-L123)]
> 
> Whoops, that's news to me. Thanks for digging it out!

This was also news to me too. Many thanks for linking to the code. 🙏

PR #13123, opened 2022-09-20
[local-preview] Warn and Confirm from user before proceeding

### LINKED ISSUE #11792: [local preview] Check arch and fail fast on non-supported arch
The local preview image does not work on Macs with Apple Silicon (e.g. M1). We should check the arch at the beginning of the container run and fail fast when the arch is not supported (with a clear error message).

Pothulapati:
## Description

Currently, For Users with ARM CPU's, `local-preview` does not really work and its hard for the script to know as it runs inside docker (which could be a x86 VM).

This Updates the script to warn the users on the requirements before starting and running `local-preview`.

Screenshot 2022-09-20 at 6 27 14 PM

The default value is also `YES` which means even when the user presses enter, we proceed forward.

## Related Issue(s)

Fixes #11792 

## How to test

Run

```
docker run -p 443:443 --privileged --name gitpod -it  --mount type=volume,source=gitpod,destination=/var/gitpod eu.gcr.io/gitpod-core-dev/build/local-preview:tar-lp-warn-confirm.28
```

## Release Notes

```release-note
[local-preview] Warn and Confirm from user before proceeding
```

## Werft options:

- [x] /werft with-local-preview


install/preview/prettylog/main.go
```diff
@@ -38,6 +38,20 @@ var (
 )
 
 func main() {
+	// Warn and wait for user approval
+	pterm.FgBlue.Println(`
+Welcome to the local preview of Gitpod. Please note the following limitations:
```
lucasvaltl, line 43: I'd suggest we change this to: 
```
Welcome to the local preview of Gitpod. Please note the following limitations:
  - Performance is limited by the capabilities of your machine - a minimum of 4 cores and 6GB of RAM are required
  - ARM CPUs including Macs with Apple Silicon (e.g. M1) are currently not supported
  For more information about these limitation, please visit the [local preview documentation](https://www.gitpod.io/docs/self-hosted/latest/local-preview).
```
--> This adds a link & removes the periods at the end of the bullet points.
--> I am assuming we can do links in the CLI output here, but not sure how / if this works.

lucasvaltl (approved):
LGTM (did not test or review the implementation), but suggested a content change that would be good to add.

install/preview/prettylog/main.go
Pothulapati: Screenshot 2022-09-21 at 11 31 13 AM


It is rendered like this, with the http link being clickable (depends on the shell too)

lucasvaltl: In that case we should likely not try to make the http link a clickable link (with the [...](...) ) like on the web but instead have the link as full text: 
```
Welcome to the local preview of Gitpod. Please note the following limitations:
  - Performance is limited by the capabilities of your machine - a minimum of 4 cores and 6GB of RAM are required
  - ARM CPUs including Macs with Apple Silicon (e.g. M1) are currently not supported
  For more information about these limitation, please visit the local preview documentation: https://www.gitpod.io/docs/self-hosted/latest/local-preview

Comment (Pothulapati) [reply install/preview/prettylog/main.go:43]: Updated.

File: install/preview/prettylog/main.go [hunk 2]
```diff
@@ -38,21 +39,31 @@ var (
...
+	result, _ := pterm.DefaultInteractiveConfirm.WithDefaultText("Continue?").WithDefaultValue(true).Show()
+	if !result {
+		// send telemetry for user exit
+		send_telemetry("user exit")
+		return
 	}
-	defer dmp.Close()
 
-	r := io.TeeReader(os.Stdin, dmp)
+	file, err := tail.TailFile("logs.txt", tail.Config{Follow: true})
```
Pothulapati, line 56: @MrSimonEmms Like we discussed in our call. This was because while `logs.txt` gets updated, We just read at a point and are exiting prettylog as there is EOF.

Instead we need a way to continuously read `logs.txt` as it gets updated, and this is where the `tail` package [comes in ](https://stackoverflow.com/a/15328871). By using that we are continuously reading from the latest updates, and thus updating status based on that.

Pothulapati: Was also able to test that this works as expected!

MrSimonEmms (approved):
wORKS AS EXPECTED

PR #13126, opened 2022-09-20
[installer]: add validation rules to blockNewUsers in config block

### LINKED ISSUE #13114: Add validation check to ensure that, if limiting user registrations, there is at least one exception
If the "user limit" is enabled, you can end up in a world of pain if there are no exceptions. The application allows you to log in, but then it automatically blocks that user as not in the exceptions list.

MrSimonEmms:
## Description

If a user selects the option to limit their user registrations (which they should if on public internet), it is possible to get their Gitpod instance into a weird state where the initial admin user is blocked if they had configured no items in the passlist.

This adds a validation check to ensure that is not possible.

## Related Issue(s)

Fixes #13114

## How to test

Four scenarios to check:
1. If `blockNewUsers.enabled` is set to `false` and `blockNewUsers.passlist` is set to `[]` - this should pass
2. If `blockNewUsers.enabled` is set to `true` and `blockNewUser.passlist` is set to `[]` - this should fail
3. If `blockNewUsers.enabled` is set to `true` and `blockNewUser.passlist` is set to `["somedomain"]` - this should fail
3. If `blockNewUsers.enabled` is set to `true` and `blockNewUser.passlist` is set to `["somedomain.com"]` - this should pass

Use `go run . validate config -c /path/to/config.yaml` to execute without building the binary

## Release Notes

```release-note
[installer]: add validation rules to blockNewUsers in config block
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/installer/pkg/config/validation.go
+++ b/install/installer/pkg/config/validation.go
@@ -55,6 +55,8 @@ func Validate(version ConfigVersion, cfg interface{}) (r *ValidationResult, err 
 					res.Fatal = append(res.Fatal, fmt.Sprintf("Field '%s' is %s '%s'", v.Namespace(), tag, v.Param()))
 				case "startswith":
 					res.Fatal = append(res.Fatal, fmt.Sprintf("Field '%s' must start with '%s'", v.Namespace(), v.Param()))
+				case "block_new_users_passlist":
+					res.Fatal = append(res.Fatal, fmt.Sprintf("Field '%s' failed. If 'Enabled = true', there must be at least one fully-qualified domain name in the passlist", v.Namespace()))
 				default:
 					// General error message
 					res.Fatal = append(res.Fatal, fmt.Sprintf("Field '%s' failed %s validation", v.Namespace(), v.Tag()))
--- a/install/installer/pkg/config/v1/config.go
+++ b/install/installer/pkg/config/v1/config.go
@@ -403,8 +403,9 @@ const (
 )
 
 type BlockNewUsers struct {
-	Enabled  bool     `json:"enabled"`
-	Passlist []string `json:"passlist"`
+	Enabled bool `json:"enabled"`
+	// Passlist []string `json:"passlist" validate:"min=1,unique,dive,fqdn"`
+	Passlist []string `json:"passlist" validate:"block_new_users_passlist"`
 }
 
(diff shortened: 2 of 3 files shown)


Pothulapati (approved):
LGTM, and works as expected

install/installer/pkg/config/validation.go
```diff
@@ -55,6 +55,8 @@ func Validate(version ConfigVersion, cfg interface{}) (r *ValidationResult, err
 					res.Fatal = append(res.Fatal, fmt.Sprintf("Field '%s' is %s '%s'", v.Namespace(), tag, v.Param()))
 				case "startswith":
 					res.Fatal = append(res.Fatal, fmt.Sprintf("Field '%s' must start with '%s'", v.Namespace(), v.Param()))
+				case "block_new_users_passlist":
+					res.Fatal = append(res.Fatal, fmt.Sprintf("Field '%s' failed. If 'Enabled = true', there must be at least one fully-qualified domain name in the passlist", v.Namespace()))
```
mustard-mh, line 59: Maybe `Warning` is better, for the existing self-hosted

MrSimonEmms: No, I think fatal is correct. If you don't have any domains in the passlist then you end up in a broken state where the initial admin user gets blocked and you've basically bricked the installation. The only way I've found to resolve this is by deleting the database (I guess it's possible to do by deleting specific DB records, but I don't know those)

mustard-mh: It's good for new SH installation. But for existing SH, they need to LeanMore and change their `config.yaml`

And it blocks my case -- an empty passlist.

(My case also not allow me to use the same domain of email, so it's useless for my case like small company?)

mustard-mh: We can fix it by making first account available?

MrSimonEmms: Yeah, if there's an exception made to never block the first account then that would remove this problem completely.

I'd propose keeping the test as a fatal error, but it would remove the requirement for option `2` in the original test scenarios:
> If `blockNewUsers.enabled` is set to `true` and `blockNewUser.passlist` is set to `[]` - this should fail

PR #13145, opened 2022-09-20
[pre-commit] Split out pre-commit to separate task

easyCZ:
## Description

There's no reason for it to be bundled with Go

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -59,7 +59,7 @@ tasks:
   - name: Go
     init: |
       ./components/gitpod-protocol/go/scripts/generate-config.sh
-      leeway exec --filter-type go -v -- go mod verify
+      leeway exec --filter-type go -v -- go mod download
     openMode: split-right
   - name: Pre-commit
     init: pre-commit install --install-hooks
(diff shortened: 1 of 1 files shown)


.gitpod.yml
```diff
@@ -57,11 +57,12 @@ tasks:
...
     init: yarn --network-timeout 100000 && yarn build
   - name: Go
-    before: pre-commit install --install-hooks
     init: |
       ./components/gitpod-protocol/go/scripts/generate-config.sh
-      leeway exec --filter-type go -v -- go mod verify
+      leeway exec --filter-type go -v -- go mod download
     openMode: split-right
+  - name: Pre-commit
+    init: pre-commit install --install-hooks
```
AlexTugarev, line 65: @easyCZ, I wonder, could we install them *and* warm them um in dev image? They don't receive frequent update, do they?

easyCZ: Good idea. I'd need to investigate if the image itself has access to the `.pre-commit` config, or if we'd need to shift that into the image also.

AlexTugarev: happy to proceed with this PR as well. just to let you know, if you think this is ready, let us know.

PR #13161, opened 2022-09-21
[dashboard] add pagination to admin search

### LINKED ISSUE #8983: Populate Admin dashboard list views without first requiring a search

Showing nothing in the Admin Dashboard on the Workspaces page has resulted in some confusion that no workspaces exist on an instance or that there's a problem with the UI or a backend process that may fetch a list of workspaces, as one expects a list of workspaces to show up when the page is loaded. At the moment this page is only populated when a search is performed. This action which must be performed manually is not clearly identified on the page.

I'd like the page to display a list of workspaces perhaps 25 to 50 at a time in a paginated table.

None as there is no other UI in the admin dashboard which displays a listing of workspaces.

## Additional context

N/A.

svenefftinge:
## Description
This change adds pagination support for the various admin dashboard list views.

## Related Issue(s)

Fixes #8983

## Release Notes

```release-note
Admin dashboard now has pagination
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/dashboard/src/admin/UserSearch.tsx
+++ b/components/dashboard/src/admin/UserSearch.tsx
@@ -9,6 +9,7 @@ import moment from "moment";
 import { useEffect, useState } from "react";
 import { useLocation } from "react-router";
 import { Link } from "react-router-dom";
+import Pagination from "../Pagination/Pagination";
 import { getGitpodService } from "../service/service";
 import { PageWithAdminSubMenu } from "./PageWithAdminSubMenu";
 import UserDetail from "./UserDetail";
@@ -19,6 +20,11 @@ export default function UserSearch() {
     const [searchTerm, setSearchTerm] = useState("");
     const [searching, setSearching] = useState(false);
     const [currentUser, setCurrentUserState] = useState<User | undefined>(undefined);
+    const pageLength = 50;
+    const [currentPage, setCurrentPage] = useState(1);
+    useEffect(() => {
+        search();
+    }, [currentPage]);
 
     useEffect(() => {
         const userId = location.pathname.split("/")[3];
@@ -46,9 +52,9 @@ export default function UserSearch() {
         try {
             const result = await getGitpodService().server.adminGetUsers({
                 searchTerm,
-                limit: 50,
+                limit: pageLength,
                 orderBy: "creationDate",
-                offset: 0,
+                offset: (currentPage - 1) * pageLength,
                 orderDir: "desc",
             });
             setSearchResult(result);
@@ -103,6 +109,11 @@ export default function UserSearch() {
                         <UserEntry user={u} />
                     ))}
             </div>
+            <Pagination
+                currentPage={currentPage}
+                setPage={setCurrentPage}
+                totalNumberOfPages={Math.ceil(searchResult.total / pageLength)}
+            />
         </PageWithAdminSubMenu>
     );
 }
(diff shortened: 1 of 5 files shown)


components/dashboard/src/admin/UserSearch.tsx
```diff
@@ -46,9 +52,9 @@ export default function UserSearch() {
         try {
             const result = await getGitpodService().server.adminGetUsers({
                 searchTerm,
-                limit: 50,
+                limit: pageLength,
                 orderBy: "creationDate",
-                offset: 0,
+                offset: (currentPage - 1) * pageLength,
```
easyCZ, line 57: `currentPage` starts at value of 0, wouldn't this cause the offset to be negative?

components/dashboard/src/admin/WorkspacesSearch.tsx
```diff
@@ -73,9 +79,9 @@ export function WorkspaceSearch(props: Props) {
 
     const search = async () => {
         // Disables empty search on the workspace search page
-        if (!props.user && queryTerm.length === 0) {
-            return;
-        }
+        // if (!props.user && queryTerm.length === 0) {
+        //     return;
+        // }
```
easyCZ, line 84: I believe this was the cause of an incident when we originally added the WS search - it would cause too many queries (but may be okay now)

components/dashboard/src/admin/WorkspacesSearch.tsx
```diff
@@ -43,6 +44,11 @@ export function WorkspaceSearch(props: Props) {
     const [queryTerm, setQueryTerm] = useState("");
     const [searching, setSearching] = useState(false);
     const [currentWorkspace, setCurrentWorkspaceState] = useState<WorkspaceAndInstance | undefined>(undefined);
+    const pageLength = 2;
```
easyCZ, line 47: Why so low? Originally this was set to 100

components/dashboard/src/admin/UserSearch.tsx
svenefftinge: currentPage should start at 1.

components/dashboard/src/admin/WorkspacesSearch.tsx
svenefftinge: Yes, was this maybe before we introduced limits? cc @geropl Do you remember?
It's for sure a SaaS hack that makes the view less useful in smaller self-hosted installations. If we need to protect us we better do that on the server based on a saas-specific configuration.

components/dashboard/src/admin/WorkspacesSearch.tsx
svenefftinge: For testing. But I have updated it 50 again.

easyCZ: Out of curiosity, why is it needed to commit test code? I'm asking because I'm trying to understand the problem and because this is something worth digging into to eliminate. Committing it risks we actually land it (and we have done so in past and caused incidents).

components/dashboard/src/admin/UserSearch.tsx
easyCZ: Now it does, but before it was definitely initialized to 0.

components/dashboard/src/admin/WorkspacesSearch.tsx
svenefftinge: In this case, it was just an overlook by myself.
It could have been a conscious decision  to allow others testing the changes more easily. In that case I would have added a 'hold' and a comment.

components/dashboard/src/admin/UserSearch.tsx
svenefftinge: Yes on 3 out of 4 occasions 😬 (the four search pages are a bit copy and paste and of course I only tested on one of them, which also explains why the limit was set to `2` only on one of them)

easyCZ (approved):

PR #13162, opened 2022-09-21
[server] decouple get ide config logic from server

iQQBot:
## Description

Decouple get ide config logic from `server`

## Related Issue(s)

Fixes #

## How to test

1. open preview environment, it should  be able open a workspace
2. go to setting -> preference, it should be able see all ide selection
3. scaling down `ide-service`, the 1 and 2 step should work too (but with error output in server)
```shell
kubectl scale deployment ide-service --replicas 0
```
image

4. scaling up `ide-service`, and check again (no error in server)
```shell
kubectl scale deployment ide-service --replicas 1
```
5. use `kubectl port-forward svc/grafana -n monitoring-satellite 3000:3000` to open grafana and go to export use this query `grpc_client_handled_total{job="server", grpc_service="ide_service_api.IDEService"}` to see some metrics
image

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft withObservabilityBranch=pd/ide-service


roboquat:
@iQQBot: Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://git.k8s.io/community/contributors/guide/release-notes.md) to remove it.

<details>

</details>

install/installer/pkg/components/ide-service/ide_config_configmap.go
```diff
@@ -168,10 +167,10 @@ func ideConfigConfigmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		&corev1.ConfigMap{
 			TypeMeta: common.TypeMetaConfigmap,
 			ObjectMeta: metav1.ObjectMeta{
-				Name:        fmt.Sprintf("%s-ide-config", server.Component),
+				Name:        "server-ide-config",
```
iQQBot, line 170: keep this, because deploy order, and use hardcode `server` in order to prevent import cycle

components/server/src/ide-service.ts
```diff
@@ -5,10 +5,38 @@
  */
 
 import { JetBrainsConfig, TaskConfig, Workspace } from "@gitpod/gitpod-protocol";
-import { injectable } from "inversify";
+import { IDEOptions, IDEClient } from "@gitpod/gitpod-protocol/lib/ide-protocol";
```
akosyakov, line 8: I wonder whether we can use types from `@gitpod/ide-service-api`?

iQQBot: This is the first small step in the decouple, and it may not be a good idea to expose the specific structure of ide config to the public. Instead, we can shield the server's dependency on the configmap structure in a subsequent step, and replace it with a new IDEClient structure, which we will expose

iQQBot: In other words, these are temporary and are going to be deleted

akosyakov (approved):
I tried to start some IDEs and skimmed through the code. It looks good.

easyCZ (approved):
For PRs like this, I'd strongly encourage splitting them into multiple PRs along their component boundaries. Similarly, I'd recommend splitting config removal across multiple PRs - if a component is rolled back by patching it's version, it's often the case that the config is not patched also - causing non-backwards compatible issues.

Webapp changes look ok.

/hold in case there are changes from other reviewers.

components/server/src/container-module.ts
```diff
@@ -270,6 +270,14 @@ export const productionContainerModule = new ContainerModule((bind, unbind, isBo
         return createClient(BillingServiceDefinition, createChannel(config.usageServiceAddr));
     });
 
+    bind<IDEServiceClient>(IDEServiceDefinition.name).toDynamicValue((ctx) => {
```
AlexTugarev, line 273: @iQQBot could that be bound in singleton scope as well?

Asking, because of investigation of memory leak issue introduced very recently.

iQQBot: fine with me, this code is reference https://github.com/gitpod-io/gitpod/blob/071500408fd5a374e0317dca86effcdc23856f2c/components/server/src/container-module.ts#L263-L271, in my case I just want `config`

AlexTugarev: @iQQBot, I thought so ;-) 
see https://github.com/gitpod-io/gitpod/pull/13456

AlexTugarev: @iQQBot, would you mind adjusting in same way?

iQQBot: in another PR? or add an commit to https://github.com/gitpod-io/gitpod/pull/13456 ?

iQQBot: https://github.com/gitpod-io/gitpod/pull/13466 here is the PR

PR #13168, opened 2022-09-21
Validate the configuration as part of a pre-flight check

### LINKED ISSUE #12015: Validate gitpod config patch, rendered gitpod installer YAML in preflight checks

MrSimonEmms:
## Description

Enable validation of the configuration in KOTS in a preflight check.

This also pushes the result of the installation script to a configmap for users to easily read the logs. The reason for the "push to configmap" is because we cannot output dynamic data in an error message (or, at least, there's no documented example from Replicated). This has been noted on the "limitations of Replicated" page as one of the things to solve in the future.

This also removes the serviceaccount/clusterrolebinding that we define in favour of the [KOTS RBAC](https://docs.replicated.com/reference/custom-resource-application#requireminimalrbacprivileges) configuration - this was not available when we first did it. We need this change because we need a serviceaccount _BEFORE_ the application is deployed.

### Limitations

This tests every piece of data that is not stored in a secret. Currently, that is just the merging of the custom image pull secrets. Because the preflight checks are run _BEFORE_ anything is deployed, we cannot deploy secrets until _AFTER_ they're deployed - there is a [ticket open with Replicated](https://github.com/replicated-collab/replicated-gitpod/issues/51) that would solve this limitation, although there has been little traction on this as yet.

Theoretically, we could store the config values in plain text in the config map, but this would make them visible in the KOTS dashboard so I have elected to not do this. I'm happy to have discussions on this however.

The refactoring work has reduced this gap from pretty-much everything to a very specific limitation.

## Related Issue(s)

Fixes #12015

## How to test

1. Deploy your application
2. Set your configuration
3. Start the preflight checks
4. Check that it's valid/invalid as per your test
5. Get the config map (`kubectl get configmaps -n gitpod gitpod-installation-status -o jsonpath="{.data.gitpod-installer\.log}"`) and check the logs are "correct" as per your test

Adding a config patch as the following is a good way of enforcing a failure

```yaml
domain: ""
```

## Release Notes

```release-note
Validate the configuration as part of a pre-flight check
```

## Werft options:

- [ ] /werft with-preview


roboquat:

<details>

</details>

install/kots/Makefile
```diff
@@ -24,6 +24,31 @@ create_unstable_release:
 	replicated release create --lint --ensure-channel --yaml-dir ${YAML_DIR} --promote ${CHANNEL_UNSTABLE}
 .PHONY: create_unstable_release
 
+generate_preflight_checks:
```
Pothulapati, line 27: nit: Hmm. So Thinking more about this approach here i.e generating the kots-preflight pod manifest from the `installer-job` and changing it through `yq`, I was thinking its a bit too much but also writing a replica of `gitpod-installer-job` isn't a great idea too. So, Not too many opinions here. I think, we can move forward for now.

Pothulapati (approved):
Code looks good, and it works well as expected! 

Tried setting `observability.logLevel` to a random value
```
this may overwrite all settings: observability:\n  logLevel: random\n"
time="2022-09-22T08:35:52Z" level=info msg="File written to /tmp/gitpod-config.yaml"
Gitpod: Validate config
{
  "valid": false,
  "fatal": [
    "Field 'Config.Observability.LogLevel' failed log_level validation"
  ]
}
Error: configuration invalid
Usage:
```

image

And, Fixing it made the tests pass!

Really excited for all the useful validation checks to show up in `kots` now! 🚀

install/kots/Makefile
MrSimonEmms: The intention is that this goes away when Replicated does the [feature request](https://github.com/replicated-collab/replicated-gitpod/issues/51). I think the `yq` script is the least-bad way of doing it, but I'm very happy to be challenged on that

Pothulapati: > least-bad way of doing it,
Agree

Issue #13131, opened 2022-09-20
getNotifications is returning 500 code errors


easyCZ:
[Dashboard](https://grafana.gitpod.io/explore?orgId=1&left=%7B%22datasource%22:%22P4169E866C3094E38%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22editorMode%22:%22code%22,%22expr%22:%22sum%28increase%28gitpod_server_api_calls_total%7Bmethod%3D%5C%22getNotifications%5C%22%7D%5B7d%5D%29%29%20by%20%28statusCode%29%22,%22legendFormat%22:%22__auto%22,%22range%22:true,%22instant%22:true%7D%5D,%22range%22:%7B%22from%22:%22now-7d%22,%22to%22:%22now%22%7D%7D&query=gr)
<img width="1587" alt="Screenshot 2022-09-20 at 15 44 55" src="https://user-images.githubusercontent.com/1419286/191274338-bb27dee5-5ae6-4e52-b316-28c6a2dd523e.png">

query: `sum(increase(gitpod_server_api_calls_total{method="getNotifications"}[7d])) by (statusCode)`

easyCZ:
@jankeromnes Landing https://github.com/gitpod-io/gitpod/pull/13178 caused this issue to close. But I believe the fix introduced does not actually fix the API returning 500 errors - it handles them on the UI side but does not stop them from happening server side. Are you happy for me to re-open?

jankeromnes:
Hi @easyCZ! 👋

> it handles them on the UI side but does not stop them from happening server side

Actually, https://github.com/gitpod-io/gitpod/pull/13178 handles them server side 😇

<img width="1386" alt="Screenshot 2022-09-22 at 09 46 46" src="https://user-images.githubusercontent.com/599268/191688758-d7dbd96b-3c01-499e-8e92-fd5b8907e816.png">

And looking at the code, I'm pretty sure that this does fix the API returning 500 errors (unless `super` throws, but that [seems unlikely](https://github.com/gitpod-io/gitpod/blob/fff9ce3d5b7a2b69481b1284d0ad5a1f66de8a76/components/server/src/workspace/gitpod-server-impl.ts#L3229-L3232)).

But please do re-open this if you notice more 500 errors from this call post deployment. 👍

easyCZ:
[...]

PR #13178, opened 2022-09-21
[server] If getting usage-based notifications fails, log a warning instead of propagating the error to the dashboard

jankeromnes:
## Description

If getting usage-based notifications fails, log a warning in `server` instead of propagating the error to the dashboard.

## Related Issue(s)

Should fix https://github.com/gitpod-io/gitpod/issues/13131

by silencing this error which happens when you have multiple teams with usage-based billing enabled:

```
Request getNotifications unsuccessful: 455/"Multiple teams have billing enabled."
```

More context [in Slack](https://gitpod.slack.com/archives/C027AUU7BC3/p1663683874566519?thread_ts=1663683642.094259&cid=C027AUU7BC3) (internal):

> > Should we display the prompt when we see this error?
> 
> I'd rather catch it locally and ignore it for now.
> The user will be prompted on next workspace start, anyway. 🧘

## How to test

Hint: Reviewing without whitespace is easier: https://github.com/gitpod-io/gitpod/pull/13178/files?w=1

Screenshot 2022-09-21 at 16 43 37

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/server/ee/src/workspace/gitpod-server-impl.ts
+++ b/components/server/ee/src/workspace/gitpod-server-impl.ts
@@ -2294,15 +2294,20 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
         const result = await super.getNotifications(ctx);
         const user = this.checkAndBlockUser("getNotifications");
 
-        const billingMode = await this.billingModes.getBillingModeForUser(user, new Date());
-        if (billingMode.mode === "usage-based") {
-            const limit = await this.billingService.checkUsageLimitReached(user);
-            if (limit.reached) {
-                result.unshift("The usage limit is reached.");
-            } else if (limit.almostReached) {
-                result.unshift("The usage limit is almost reached.");
+        try {
+            const billingMode = await this.billingModes.getBillingModeForUser(user, new Date());
+            if (billingMode.mode === "usage-based") {
+                const limit = await this.billingService.checkUsageLimitReached(user);
+                if (limit.reached) {
+                    result.unshift("The usage limit is reached.");
+                } else if (limit.almostReached) {
+                    result.unshift("The usage limit is almost reached.");
+                }
             }
+        } catch (error) {
+            log.warn({ userId: user.id }, "Could not get usage-based notifications for user", { error });
         }
+
         return result;
     }
 


AlexTugarev (approved):

easyCZ:
Coming back to this, is this the correct approach? Just logging the error and returning nothing will only result in swalowing the error and not actually giving us a signal. Yes, we've got the log but that's not ultimately what we monitor so we'll not know (get alerted) when this is failing for users and they are not receiving their notifications.

What if we instead handled the notification subscription failing client side (not showing it) but we continued to throw the errors server side. That way, we can actually fix the underlying problem which seems errors to check the usage limit.

jankeromnes:
Thanks @easyCZ! See also how we reached this consensus [in Slack](https://gitpod.slack.com/archives/C027AUU7BC3/p1663683874566519?thread_ts=1663683642.094259&cid=C027AUU7BC3) (internal).

> > Should we display the prompt when we see this error?
> 
> I'd rather catch it locally and ignore it for now.
> The user will be prompted on next workspace start, anyway. 🧘

FYI, [this](https://gitpod.slack.com/archives/C027AUU7BC3/p1663684564043749?thread_ts=1663683642.094259&cid=C027AUU7BC3) (internal) was the identified underlying problem (that we're now logging as a warning instead of actually handling it):

> Found it: When you call `checkUsageLimitReached`, it (correctly) wants to know the user's `attributionId` -- but that can be ambiguous (e.g. no explicit choice and two paying teams)
> https://github.com/gitpod-io/gitpod/blob/c9aa5276c271fde089d75ba769ae3ffd6f62faef/components/server/ee/src/billing/billing-service.ts#L27
> So, `checkUsageLimitReached` throws when attribution is ambiguous

**Summary**: We've decided not to handle the attribution ambiguity in `getNotifications` (because it's sort of a nice to have, and it's not the end of the world when it doesn't work), but instead wait until the user tries to start a workspace -- at which point the same ambiguity error happens, and causes a blocking modal to appear, encouraging users to lift the ambiguity by making an explicit attribution choice. After this point, the error is solved (so `getNotifications` will work as intended).

However, maybe the catch-all should be more specific (i.e. log attribution ambiguity as a warning, but still throw any other error)?

easyCZ:
> However, I do see your point: Maybe the catch-all should be more specific (i.e. log attribution ambiguity as a warning, but still re-throw any other error)?

Indeed, that would make it more explicit (and also actually document the reason for this in code). As it stands, without that context I'd go and remove the try-catch-all

PR #13186, opened 2022-09-21
[server] don't allow more than three usages per phone number

### LINKED ISSUE #12883: Limit phone number re-use for verification
Followup to #12258

Phone number verification would probably be more effective if it did not allow re-use of phone numbers across accounts. There appear to be services which provide low-friction re-usable phone numbers for circumventing verification.

svenefftinge:
## Description
Adds checks for phone numbers being used by abusers or used too often (>3) 

## Related Issue(s)

Fixes #12883

## Release Notes

```release-note
Restrict reuse of phone numbers for verification
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-db/src/typeorm/migration/1663784254956-IndexPhoneNumber.ts
+++ b/components/gitpod-db/src/typeorm/migration/1663784254956-IndexPhoneNumber.ts
@@ -1,0 +1,23 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { MigrationInterface, QueryRunner } from "typeorm";
+import { columnExists } from "./helper/helper";
+
+const D_B_USER = "d_b_user";
+const COL_PHONE_NUMBER = "verificationPhoneNumber";
+
+export class IndexPhoneNumber1663784254956 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        if (!(await columnExists(queryRunner, D_B_USER, COL_PHONE_NUMBER))) {
+            await queryRunner.query(
+                `ALTER TABLE ${D_B_USER} ADD INDEX (${COL_PHONE_NUMBER}), ALGORITHM=INPLACE, LOCK=NONE `,
+            );
+        }
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {}
+}
--- a/components/server/src/auth/verification-service.ts
+++ b/components/server/src/auth/verification-service.ts
@@ -10,13 +10,16 @@ import { inject, injectable, postConstruct } from "inversify";
 import { Config } from "../config";
 import { Twilio } from "twilio";
 import { ServiceContext } from "twilio/lib/rest/verify/v2/service";
-import { WorkspaceDB } from "@gitpod/gitpod-db/lib";
+import { UserDB, WorkspaceDB } from "@gitpod/gitpod-db/lib";
 import { ConfigCatClientFactory } from "@gitpod/gitpod-protocol/lib/experiments/configcat-server";
+import { ErrorCodes } from "@gitpod/gitpod-protocol/lib/messaging/error";
+import { ResponseError } from "vscode-ws-jsonrpc";
 
 @injectable()
 export class VerificationService {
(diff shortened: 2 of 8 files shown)


roboquat:

<details>

</details>

components/server/src/auth/verification-service.ts
```diff
@@ -59,6 +60,14 @@ export class VerificationService {
         if (!this.verifyService) {
             throw new Error("No verification service configured.");
         }
+        const isBlockedNumber = this.userDB.countUsagesOfPhoneNumber(phoneNumber);
```
atduarte, line 63:
```suggestion
        const isBlockedNumber = this.userDB.isBlockedPhoneNumber(phoneNumber);
```

atduarte (commented):
This is awesome! Thank you, Sven 🙏 

Looking at the code out of curiosity found something you will notice also pretty quickly :) 

PS: Would it make sense and be sensible to also stop overriding the phone number on delete as part of this ([here](https://github.com/gitpod-io/gitpod/blob/main/components/server/src/user/user-deletion-service.ts#L157))? Asking because a potential way for an abuser to avoid both these checks is requesting the account to be deleted.

components/gitpod-db/src/typeorm/migration/1663784254956-IndexPhoneNumber.ts
```diff
@@ -0,0 +1,23 @@
...
+import { columnExists } from "./helper/helper";
+
+const D_B_USER = "d_b_user";
+const COL_PHONE_NUMBER = "verificationPhoneNumber";
+
+export class IndexPhoneNumber1663784254956 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        if (!(await columnExists(queryRunner, D_B_USER, COL_PHONE_NUMBER))) {
+            await queryRunner.query(
+                `ALTER TABLE ${D_B_USER} ADD INDEX (${COL_PHONE_NUMBER}), ALGORITHM=INPLACE, LOCK=NONE `,
```
easyCZ, line 17: What's the expected shape of the phone number here? Is it including spaces? Is it including country codes? Are the country codes encoded as + or 00? Would any of these formats cause issues when checking for the number of times a number has been used?

svenefftinge: There are a couple of allowed shapes, which could be exploited to reuse the same number. Is that what you are asking?

easyCZ: Basically. Could we in the least do the following transformations:

svenefftinge: We could store the number in https://en.wikipedia.org/wiki/E.164. I'll check later if that's easy enough.

easyCZ (changes requested):
Requesting changes as per [internal thread](https://gitpod.slack.com/archives/C01MLB7J24T/p1663830093846229)

components/server/src/auth/verification-service.ts
```diff
@@ -59,6 +60,14 @@ export class VerificationService {
         if (!this.verifyService) {
             throw new Error("No verification service configured.");
         }
+        const isBlockedNumber = this.userDB.isBlockedPhoneNumber(phoneNumber);
+        const usages = await this.userDB.countUsagesOfPhoneNumber(phoneNumber);
+        if (usages > 3) {
+            throw new Error("The given phone number has been used more than three times.");
+        }
+        if (await isBlockedNumber) {
+            throw new Error("The given phone number is blocked due to abuse.");
```
easyCZ, line 69: Can we throw this as a user error here? Otherwise it will count towards server errors in our metrics.

components/server/src/user/phone-numbers.ts
```diff
@@ -0,0 +1,23 @@
...
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+/**
+ * a simply cleaning method, that removes all non numbers and repaces a leading '00' with a leading '+' sign.
+ *
+ * @param phoneNumber
+ * @returns formatted phone number
+ */
+export function formatPhoneNumber(phoneNumber: string): string {
```
easyCZ, line 13: 👍

easyCZ (approved):
LGTM, one comment about correctly distinguishing user errors and server errors.

/hold

PR #13189, opened 2022-09-21
Add PortsStatus::on_open

jeanp413:
## Description

In case of forwarding we need to know configuration in advance. Right now it is possible to get to grasp of it only when port is exposed. We deprecated on open action in the exposed state (ExposedPortInfo::on_exposed) and move to PortStatus itself (use PortsStatus::on_open instead).

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


components/supervisor-api/status.proto
```diff
@@ -142,19 +142,27 @@ enum PortVisibility {
...
     notify = 3;
     notify_private = 4;
 }
+enum OnPortOpenAction {
+    _ignore = 0;
+    _open_browser = 1;
+    _open_preview = 2;
+    _notify = 3;
+    _notify_private = 4;
+}
```
jeanp413, line 159: Not sure about naming here as I can use the same names in OnPortExposedAction

mustard-mh:
Any context about why we need these changes?

jeanp413:
I sync @akosyakov today and talked about ports in vscode and we need this info even if the port is not exposed so we can update Machine/settings.json

akosyakov:
@mustard-mh I've updated the PR description to clarify for future readers.

akosyakov (approved):
lgtm

PR #13203, opened 2022-09-22
[server] Add a liveness probe for event loop lag

andrew-farries:
## Description

Add a liveness probe for `server`.

The probe fails iff the nodejs event loop lag (as reported by a prometheus metric) exceeds the value set in `server` config.

**Context**:

We have an alert and corresponding [runbook](https://github.com/gitpod-io/runbooks/blob/main/runbooks/ServerEventLoopLagTooHigh.md) which fires when the event loop lag exceeds a certain threshold. The runbook simply advises the operator to restart the affected pods, something that should not require manual intervention.

## How to test

1. Edit the `server-config` configmap to change the `maximumEventLoopLag` setting.
2. Restart the `server` pod after changing the configmap.
3. Hit `<preview-url>/api/live` to see the current event loop lag and see the response code change according to the current value of `maximumEventLoopLag`.

## Release Notes

```release-note
[server]: Add a liveness probe which fails when the nodejs event loop lag exceeds a certain threshold
```

## Werft options:

- [x] /werft with-preview
- [ ] /werft with-integration-tests=webapp

### DIFF
--- a/components/server/src/liveness/liveness-controller.ts
+++ b/components/server/src/liveness/liveness-controller.ts
@@ -1,0 +1,41 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { injectable, inject } from "inversify";
+import * as express from "express";
+import * as prom from "prom-client";
+import { Config } from "../config";
+import { log } from "@gitpod/gitpod-protocol/lib/util/logging";
+
+@injectable()
+export class LivenessController {
+    @inject(Config) protected readonly config: Config;
+
+    get apiRouter(): express.Router {
+        const router = express.Router();
+        this.addLivenessHandler(router);
+        return router;
+    }
+
+    protected addLivenessHandler(router: express.Router) {
+        router.get("/", async (_, res) => {
+            if (!this.config.maximumEventLoopLag) {
+                res.end();
+                return;
+            }
+            const metrics = await prom.register.getMetricsAsJSON();
+            const metric = metrics.find((m) => m.name === "nodejs_eventloop_lag_seconds");
+            if (!metric) {
+                log.error("unable to find nodejs event loop metric, liveness probe will not function");
+                res.end();
+                return;
+            }
+            const value = (metric as any).values[0].value;
+            res.status(value > this.config.maximumEventLoopLag ? 500 : 200);
+            res.end(`${value}`);
+        });
+    }
+}
--- a/install/installer/pkg/components/server/configmap.go
+++ b/install/installer/pkg/components/server/configmap.go
@@ -236,6 +236,7 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		ContentServiceAddr:           net.JoinHostPort(fmt.Sprintf("%s.%s.svc.cluster.local", contentservice.Component, ctx.Namespace), strconv.Itoa(contentservice.RPCPort)),
 		ImageBuilderAddr:             net.JoinHostPort(fmt.Sprintf("%s.%s.svc.cluster.local", common.ImageBuilderComponent, ctx.Namespace), strconv.Itoa(common.ImageBuilderRPCPort)),
 		UsageServiceAddr:             net.JoinHostPort(fmt.Sprintf("%s.%s.svc.cluster.local", usage.Component, ctx.Namespace), strconv.Itoa(usage.GRPCServicePort)),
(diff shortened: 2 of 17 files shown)


components/server/src/liveness/liveness-controller.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    get apiRouter(): express.Router {
+        const router = express.Router();
+        this.addLivenessHandler(router);
+        return router;
+    }
+
+    protected addLivenessHandler(router: express.Router) {
+        router.get("/", async (req, res, next) => {
+            const metrics = await prom.register.getMetricsAsJSON();
```
easyCZ, line 25: You can use `getSingleMetric` to get just one and avoid filtering.

easyCZ, line 25: Short-circuit the lookup if you do not have a `maximumEventLoopLag` value configured (< 0)

install/installer/pkg/components/server/configmap.go
```diff
@@ -236,6 +236,7 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		ContentServiceAddr:           net.JoinHostPort(fmt.Sprintf("%s.%s.svc.cluster.local", contentservice.Component, ctx.Namespace), strconv.Itoa(contentservice.RPCPort)),
 		ImageBuilderAddr:             net.JoinHostPort(fmt.Sprintf("%s.%s.svc.cluster.local", common.ImageBuilderComponent, ctx.Namespace), strconv.Itoa(common.ImageBuilderRPCPort)),
 		UsageServiceAddr:             net.JoinHostPort(fmt.Sprintf("%s.%s.svc.cluster.local", usage.Component, ctx.Namespace), strconv.Itoa(usage.GRPCServicePort)),
+		MaximumEventLoopLag:          0.35,
```
easyCZ, line 239: This would cause even self-hosted instances to restart if it reached this value. Are we happy with that? Should we document that?

easyCZ (commented):
Great! Like it.

It does have the potential to effectively take all instances out of the deployment pool, however. If on startup the value spikes and the liveness probe fails, it will not consider the instance ready. That can put more load on the other instances and cause a cascading failure.

andrew-farries:
> It does have the potential to effectively take all instances out of the deployment pool, however. If on startup the value spikes and the liveness probe fails, it will not consider the instance ready. That can put more load on the other instances and cause a cascading failure.

Looking at the dashboard for this metric, we do see regular spikes but none for longer than 15 seconds. By setting the `FailureThreshold` for the probe, we should be able to avoid probe failures due to these spikes.

I've bumped the `FailureThreshold` to `6`, at a period of `10s` which means we'd need to see ~1 minute of sustained lag before the probe fails.

install/installer/pkg/components/server/configmap.go
andrew-farries: Yes, and I think it's desirable that this applies to SaaS and self-hosted. 

The only point of contention is what this value should be set to, and should it be the same in self-hosted vs SaaS installations. This value is taken from the AlertManager, ie it's set to the value that currently triggers a page to the on-caller. I think it's appropriate to leave it the same for all installations right now.

I don't think any extra documentation beyond the changelog is necessary.

components/server/src/liveness/liveness-controller.ts
andrew-farries: done

andrew-farries: Unfortunately, `getSingleMetric` returns a different type, from which we can't retrieve the value of the metric.

There are some issues discussing the inability to easily get the value of a metric from `prom-client` [[1](https://github.com/siimon/prom-client/issues/367),[2](https://github.com/siimon/prom-client/issues/245)], but I think the current approach is ok.

install/installer/pkg/components/server/deployment.go
```diff
@@ -387,6 +388,20 @@ func deployment(ctx *common.RenderContext) ([]runtime.Object, error) {
...
+										Path: "/live",
+										Port: intstr.IntOrString{
+											Type:   intstr.Int,
+											IntVal: ContainerPort,
+										},
+									},
+								},
+								InitialDelaySeconds: 120,
+								PeriodSeconds:       10,
+								FailureThreshold:    6,
```
easyCZ, line 403: 👍

easyCZ (approved):
LGTM, this is a great improvement!

install/installer/pkg/components/server/configmap.go
easyCZ: Thanks!

Issue #13155, opened 2022-09-21
[billing] Show current credit balance


svenefftinge:
In the billing view of an account (individual or team) we currently only show the spending limit but not the current credit balance. We should show that as well.

jldec:
This will also be resolved by #13564

PR #13214, opened 2022-09-22
[usage] Introduce GetBalance API

svenefftinge:
## Description
Introduces a dedicated getBalance API that we can use to check against the spending limit in the billing-service. We currently use balance from ListUsage, which i a follow up change will be changed to only sum up the usage for the given range.

## Related Issue(s)

Related https://github.com/gitpod-io/gitpod/issues/13155

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment


components/usage/pkg/db/usage_test.go
```diff
@@ -322,3 +322,37 @@ func TestListBalance(t *testing.T) {
 		CreditCents:   -50,
 	})
 }
+
+func TestGetBalance(t *testing.T) {
```
easyCZ, line 326: Please add a test for when there are no usage records for an attributionID and it should return a 0

components/usage/pkg/db/usage.go
```diff
@@ -207,6 +207,30 @@ type Balance struct {
...
+	if !rows.Next() {
+		return 0, nil
+	}
+
+	var balance CreditCents
+	err = conn.ScanRows(rows, &balance)
+	if err != nil {
+		return 0, fmt.Errorf("failed to scan row: %w", err)
+	}
+	return balance, nil
```
easyCZ, line 231: It should be possible to do it this way, if you wanted
```
var balance Balance
	tx := conn.WithContext(ctx).
		Table((&Usage{}).TableName()).
		Select("attributionId, sum(creditCents) as creditCents").
		Where("attributionId = ?", string(attributionId)).
		Group("attributionId").
		First(&balance)
	if tx.Error != nil {
		if tx.Error == gorm.ErrRecordNotFound {
			return 0, nil
		}
		
		return 0, fmt.Errorf("failed to get rows for list balance query: %w", tx.Error)
	}

	return balance.CreditCents, nil
```

easyCZ (approved):
LGTM

/hold for test & if you want to follow the suggestion on db query

easyCZ:
Ugh, why did this get merged? A race between the approve and hold?

components/usage/pkg/db/usage.go
svenefftinge: Why would you prefer that? btw. I realized I unnecessarily added a group by clause. Will remove that and add the suggested test in a follow-up PR.

easyCZ: Just because it removes the need for the manual iteration over rows, which can hurt readability for those not familiar with gorm. It also handles the not found case a bit more explicitly which helps readability. But I don't mind, just wanted to offer an alternative in case you wanted to go with it.

components/usage/pkg/db/usage_test.go
svenefftinge: Added it [here](https://github.com/gitpod-io/gitpod/pull/13222/files#diff-fa1d370734494a0f6a33ba67a648028093ddd21743ef6df87ee8b9af4410c137R370-R373)

PR #13215, opened 2022-09-22
[kots]: delete workspace pods before installing Gitpod

### LINKED ISSUE #13147: KOTS: stop running workspaces prior to upgrading existing workspace for single cluster ref arch

We do not support live upgrades for the single cluster ref arch while workspaces are running.

Before KOTS begins a deployment:
1. Prompt the user if it is okay to proceed with the deploy to an existing cluster, share this should be done during an outage window planned with their business.
2. Stop workspaces, wait for them to backup/terminate, `kubectl delete pods -l component=workspace` may suffice
3. Then deploy Gitpod to the cluster (the assumption is KOTS deletes existing resources and then recreates them)

Additionally, as part of the monthly release cycle, a self-hosted test should be added, so that the upgrade flow with running workspaces is included as part of the testing.

N/A, this removes friction from the upgrade experience.

## Additional context

The deploy process should not start in a live cluster while workspaces are running.

As of the August KOTS release, when a deploy is done to an existing cluster, resources are deleted, however...because `ws-daemon` was deleted, the workspaces could not backup, and thus could not be deleted. Therefore, it is imperative that we wait for workspace pods to be deleted (including imagebuild and prebuild), before deleting the Gitpod installation.

Customers that experience this issue willl incur data loss, and to clean-up the pods, must remove the related finalizer from the regular and prebuild workspaces.

## Dependent Tasks

 - [x] #13329
 - [x] #13330

MrSimonEmms:
## Description

Adds a preflight check to see if there are any workspaces/image-builds running. It fails if there is one (the `textAnalyzer` doesn't allow for `warn`, which is very annoying as this isn't really a failure, but a warning).

It also stops any running workspaces/image-builds prior to the `helm upgrade` function being executed.

## Related Issue(s)

Fixes #13147

## How to test

Try installing via KOTS both with and without a workspace running.

## Release Notes

```release-note
[kots]: delete workspace pods before installing Gitpod
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/installer/scripts/kots-install.sh
+++ b/install/installer/scripts/kots-install.sh
@@ -30,6 +30,28 @@ catch() {
     exit "${1}"
 }
 
+stop_running_workspaces() {
+    echo "Gitpod: shut down any running workspaces/image-builders"
+
+    # Create a context for gpctl
+    kubectl config set-context "${NAMESPACE}"
+    kubectl config use-context "${NAMESPACE}"
+    # gpctl doesn't have a namespace flag
+    kubectl config set-context --current --namespace="${NAMESPACE}"
+
+    # Get list of workspace instances from gpctl
+    for instance in $(/app/gpctl workspaces list -o json | jq -r 'select(. != null) | .[] | .Instance'); do
+        echo "Gitpod: shutting down workspace ${instance}"
+        /app/gpctl workspaces stop "${instance}" || eval "$(echo "Gitpod: retrying shutting down workspace" && \
+            sleep 10 && \
+            /app/gpctl workspaces stop "${instance}")"
+    done
+
+    # Delete the context
+    kubectl config unset "contexts.${NAMESPACE}"
+    kubectl config unset current-context
+}
+
 main() {
     if [ "${INSTALLER_DRY_RUN}" != "true" ]; then
         echo "Gitpod: Killing any in-progress installations"
@@ -155,6 +177,8 @@ EOF
     if [ "${CERT_SECRET}" = "" ]; then
         HELM_TIMEOUT="1h"
     fi
+
+    stop_running_workspaces
 
     # The long timeout is to ensure the TLS cert is created (if required)
     echo "Gitpod: Apply the Kubernetes objects with timeout of ${HELM_TIMEOUT}"
--- a/install/installer/BUILD.yaml
+++ b/install/installer/BUILD.yaml
@@ -60,6 +60,7 @@ packages:
       - "scripts/*.sh"
     deps:
       - :app
+      - dev/gpctl:app
     argdeps:
       - imageRepoBase
     config:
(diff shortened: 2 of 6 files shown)


roboquat:

<details>

</details>

install/installer/scripts/kots-install.sh
```diff
@@ -156,6 +156,9 @@ EOF
         HELM_TIMEOUT="1h"
     fi
 
+    echo "Gitpod: shut down any running workspaces/image-builders"
+    kubectl delete pods -n "${NAMESPACE}" -l component=workspace --wait
```
nandajavarma, line 160: So this would mean that during upgrade the workspaces that are currently in use will get interrupted? I thought it was a nice feature that we had, allowing uninterrupted upgrades. And how would this affect potential backup jobs?

MrSimonEmms: As I understand it, this has been requested by @gitpod-io/engineering-workspace to guarantee that the workspaces are properly backed up. If you check out #13147, there's a whole discussion on this in there

nandajavarma: totally missed the context :see_no_evil: reading!

install/installer/scripts/kots-install.sh
```diff
@@ -30,6 +30,26 @@ catch() {
...
+    # Create a context for gpctl
+    kubectl config set-context "${NAMESPACE}"
+    kubectl config use-context "${NAMESPACE}"
+    # gpctl doesn't have a namespace flag
+    kubectl config set-context --current --namespace="${NAMESPACE}"
+
+    # Get list of workspace instances from gpctl
+    for instance in $(/app/gpctl workspaces list -o json | jq -r 'select(. != null) | .[] | .Instance'); do
+        echo "Gitpod: shutting down workspace ${instance}"
+        /app/gpctl workspaces stop "${instance}"
```
MrSimonEmms, line 45: This command can be a bit flakey if (presumably) the gRPC server communication errors - I've seen logs like:

```
time="2022-09-26T17:20:18Z" level=fatal msg="cannot connect" error="error upgrading connection: error dialing backend: EOF"
```

Three options as I see it:
1. Do nothing and let it error. A redeploy will almost always be fine
2. Add a `|| true` and never let it error. This allows for missing backups
3. Add a `|| sleep 10 && /app/gpctl workspaces stop "${instance}"` to allow it to retry after 10 seconds

adrienthebo: Option 3 sounds like the best balance of reliability and safety and should catch most of the spurious gRPC server errors; I'd favor that.

adrienthebo (approved):
/hold for adding a retry on `gpctl`.

Looks good to me; I haven't functionally verified this but we can do so during the RC release testing.

install/installer/scripts/kots-install.sh
MrSimonEmms: Ok, I'll make that change now so you can get the RC out

MrSimonEmms:
Retry added as described

PR #13222, opened 2022-09-23
[usage] show sum of usage not balance

### LINKED ISSUE #13067: [usage] Show the sum of usage, not the balance
We are currently showing the credit balance the user or team had at the end of the currently selected period.
We should instead show the sum of all usage in the given selected period.

svenefftinge:
## Description
On the usage view we now only show usage and the sum is the sum of all usage in the selected period.

## Related Issue(s)

fixes #13067

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment


components/usage/pkg/db/cost_center.go
```diff
@@ -158,19 +158,19 @@ func (c *CostCenterManager) UpdateCostCenter(ctx context.Context, costCenter Cos
...
-	summary, err := GetUsageSummary(ctx, c.conn, attributionID, now, now, false)
+	creditCents, err := GetBalance(ctx, c.conn, attributionID)
 	if err != nil {
 		return nil, err
 	}
-	if summary.CreditCentsBalanceAtEnd.ToCredits() <= 0 {
+	if creditCents.ToCredits() <= 0 {
 		// account has no debt, do nothing
 		return nil, nil
 	}
```
easyCZ, line 168: While we're here, we could also change the logic to do `creditCents <= 0` because the Credits conversion only multiplies by a 100, which won't change the polarity

components/usage-api/usage/v1/usage.proto
```diff
@@ -70,11 +70,8 @@ message ListUsageRequest {
 message ListUsageResponse {
     repeated Usage usage_entries = 1;
     PaginatedResponse pagination = 2;
-    // the amount of credits the given account (attributionId) had at the beginning of the requested period
-    double credit_balance_at_start = 3;
-
-    // the amount of credits the given account (attributionId) had at the end of the requested period
-    double credit_balance_at_end = 4;
+    // the amount of credits the given account (attributionId) has used during the requested period
+    double credits_used = 3;
```
easyCZ, line 74: This is not backwards compatible. The rollout order between server and usage is undefined, and therefore you're not guaranteed it won't try to pass this value. 
Please mark `credit_balance_at_start` as either `reserved` or `[deprecated=true]` and add a new field with ID 5.

easyCZ (changes requested):
LGTM - non-backwards compatible proto change

components/usage-api/usage/v1/usage.proto
svenefftinge: I'm aware this is not compatible. 
I expect the friction this would cause minimal (very short period between the two component deploys, where errors would be thrown for a feature that is not used by users, yet) compared to the extra turnaround.

svenefftinge: Do you expect any serious issues here or is this more a matter of principles?

easyCZ: Requests could actually fail to serialize on the usage component and return Unknown error. This is because each field is looked up by their index field. Here, it's kind of lucky that the field is the same type, but the removal of a field (4) would cause issues for the serializer as it expects there to be a field with index 4 (even if it's empty).

Even if we're happy with a temporary issue, we as a team (and organization) need to do better here in order to improve our reliability. It's a matter of getting into the habit of thinking (and dealing) with these else we're gonna struggle to uphold our standards as the systems grow.

svenefftinge: Ok, I agree with the principle behind it. Not sure going through unnecessary ceremony is helpful. I think acknowledging this explicitly, which is what we are doing here, is great. IMHO it is actually even better than blindly following a certain protocol, because we are talking about the real impact. FWIW just making it technically compatible wouldn't help users, because on the client side there would be a different result unless I roll out a full compatibility mode, which I would be happy to do if it was worth it.

Issue #12638, opened 2022-09-04
[integration tests] add prebuilds workspace PVC integration test from PVC test plans


jenting:
## Is your feature request related to a problem? Please describe

Manually go thru. the PVC test plans take about half a day, and it's too long for each release to manually test against all the PVC test plans.

## Describe the behaviour you'd like

Trying to add more integration tests from the PVC test plans if possible, and make the manual PVC tests as small as possible.
Therefore, we could minimize the deployer only need to test the manually PVC test plans. 

## Describe alternatives you've considered

None

## Definition of Done (DoD)
- Update existing tests with PVC
  - [x] TestPrebuildWorkspaceTaskSuccess https://github.com/gitpod-io/gitpod/pull/12859
- Create new test with Prebuild + PVC
  - [x] PVC test plan prebuild workspace case 1 https://github.com/gitpod-io/gitpod/pull/13024
  - [x] PVC test plan prebuild workspace case 2 https://github.com/gitpod-io/gitpod/pull/13371
  - [x] PVC test plan prebuild workspace case 3 https://github.com/gitpod-io/gitpod/pull/13371
  - [x] PVC test plan prebuild workspace case 4 https://github.com/gitpod-io/gitpod/pull/13223, https://github.com/gitpod-io/gitpod/pull/13268
  - [x] PVC test plan prebuild workspace case 5 https://github.com/gitpod-io/gitpod/pull/13223, https://github.com/gitpod-io/gitpod/pull/13268

## Additional context

#7901

jenting:
> @jenting could you update the issue description, so we can see which test cases are left? 🙏

We could check the Definition of Done (DoD) to see which test cases are left. If that's not enough, we can link the notion page.

kylos101:
The DoD is perfect, I see 2 tests have PRs, and 2 are left (need PRs). Thank you, @jenting !

jenting:
I have finished all the DoD, closing this issue.

PR #13223, opened 2022-09-23
preview env: create two workspace classes default and small

jenting:
## Description

Create two workspace classes in the preview environment so we can test workspace classes in the preview environment.
- Default workspace class with 30Gi disk
- Small workspace class with 20Gi disk

image

## Related Issue(s)

Related to #12638

## How to test

- Open the workspace preview.
- Enable the feature flag PVC.
- Open the workspace, and check the PVC size is 30Gi.
- Change your workspace class preference to 20Gi
- Open a new workspace, and check the PVC size is 20Gi.

## Release Notes

```release-note
None
```

## Documentation

None

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/.werft/jobs/build/installer/installer.ts
+++ b/.werft/jobs/build/installer/installer.ts
@@ -157,6 +157,39 @@ EOF`);
         exec(`yq w -i ${this.options.installerConfigPath} workspace.resources.requests.memory "256Mi"`, {
             slice: slice,
         });
+
+        // create two workspace classes (default and small) in server-config configmap
+        exec(`yq w -i ${this.options.installerConfigPath} experimental.webapp.workspaceClasses[+].id "default"`, { slice: slice });
+        exec(`yq w -i ${this.options.installerConfigPath} experimental.webapp.workspaceClasses[0].category "GENERAL PURPOSE"`, { slice: slice });
+        exec(`yq w -i ${this.options.installerConfigPath} experimental.webapp.workspaceClasses[0].displayName "Default"`, { slice: slice });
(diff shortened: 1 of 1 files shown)


jenting:
We need to consider extending the default disk from 30Gi to 50Gi https://github.com/gitpod-io/gitpod/blob/1b995ef47f6631c09c286103db281912d724fd1d/.werft/vm/manifests.ts#L42
Otherwise, if the developer uses the disk over 30Gi, no more data could be written into the disk.
But it also has a side effect that all the default disk sizes will increase to 50Gi.

Or, we could create a small workspace class that is smaller than 30Gi disk, for example, 20Gi, since it's for testing purposes.

WDYT? @gitpod-io/platform

mads-hartmann (approved):
> we could create a small workspace class that is smaller than 30Gi disk, for example, 20Gi, since it's for testing purposes.

@jenting I'd prefer the suggestion of using smaller classes and then keeping the storage definition at "30Gi".  We could even go as low as 10gb and 20gb for the workspace classes as I don't expect people to run any workspaces larger than that in preview environments ☺️

I added the hold label and I'm approving so you can merge once you've that change ☺️

jenting:
> > we could create a small workspace class that is smaller than 30Gi disk, for example, 20Gi, since it's for testing purposes.
> 
> @jenting I'd prefer the suggestion of using smaller classes and then keeping the storage definition at "30Gi". We could even go as low as 10gb and 20gb for the workspace classes as I don't expect people to run any workspaces larger than that in preview environments ☺️
> 
> I added the hold label and I'm approving so you can merge once you've that change ☺️

Thank you @mads-hartmann 

I change to using a smaller workspace class where the disk size is 20G.
I have updated the code.

/unhold

Issue #12823, opened 2022-09-09
Build speed: components/supervisor:docker and glitpod-cli


mads-hartmann:
We're seeing that `components/supervisor:docker` is being rebuild on every build even when no source code changes to the component have been made.

This is because `components/supervisor:docker` depends on `gitpod-cli` which in turn has an argdep dependency on `version` ([ref](https://github.com/gitpod-io/gitpod/blob/main/components/gitpod-cli/BUILD.yaml#L17)).

As `version` gets updated on every build by design, so does `components/supervisor:docker` and `gitpod-cli`.

`gitpod-cli` currently uses the version information to when implementing the `gp version` command. 

The solution we discussed during the call was

1. Remove the `gp version` command entirely for now, which means we can remove the `argdep` dependency on `version`
2. Still send version information from the CLI when invoking API calls, but just send the "hash" version rather than the "build" version.

PR #13228, opened 2022-09-23
Build speed: components/supervisor:docker and glitpod-cli

mads-hartmann:
## Description

This uses the Leeway package version when populating `versions.txt` - the motivation for this change is to improve build speed. When we use `argdeps` the package build cache will be invalidated whenever the argument changes. So having an argdeps on `version` means that we'll rebuild `components/gitpod-cli:version` and `components/gitpod-cli:app` on every CI build. Because `components/supervisor:docker` depends on `components/gitpod-cli:app` it means we'll rebuild that every time too.

This will reduce our build speeds by 38s (based on the example shown below).

**Discussion**: I went with using the Leeway package version. We could also use the git commit (`__git_commit`) or even use `version` but then just know it will only be bumped when the package is actually modified. We could also include all three. 
⬆️  **Update**: We decided on using the git commit instead [here](https://gitpod.slack.com/archives/C01KGM9BH54/p1663924153343129?thread_ts=1663924067.185089&cid=C01KGM9BH54). 

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/12823

## How to test

With the changes in this PR I triggered two commits to show that it no longer rebuilds components that didn't change. The arrows in the *before* images indicate what we'll no longer be rebuilding every time.

<table>
<tr>
<td>
<strong>Before</strong>: Logs

<strong>Before</strong>: Images

</td>
<td>
<strong>After</strong>: Logs

<strong>After</strong>: Images

</td>
</tr>
</table>

## Release Notes

```release-note
NONE
```

## Documentation

N/A

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/gitpod-cli/BUILD.yaml
+++ b/components/gitpod-cli/BUILD.yaml
@@ -13,17 +13,13 @@ packages:
       - components/supervisor-api/go:lib
       - components/gitpod-protocol/go:lib
       - components/common-go:lib
-    argdeps:
-      - version
     config:
       packaging: app
     prep:
       - ["cp", "_deps/components-gitpod-cli--version/version.txt", "pkg/gitpod/version.txt"]
   - name: version
     type: generic
-    argdeps:
-      - version
     config:
       commands:
-        - ["sh", "-c", "echo '${version}' > version.txt"]
-        - ["echo", "Gitpod-CLI Version: ${version}"]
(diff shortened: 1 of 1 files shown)


iQQBot (approved):
Thanks! 🙏

PR #13233, opened 2022-09-23
[kots]: ensure log-collector directory on node at startup

### LINKED ISSUE #13095: [self-hosted] 20022.08.0 support bundle does not contain pod logs
### Bug description

In Gitpod 2022.07.x, support bundles contained `logs` and `log-collector` directories containing logs of running pods and prior pods, respectively. Recent troubleshooting has shown that support bundles created from Gitpod 2022.08.0 don't contain these directories.

### Steps to reproduce

Create a support bundle based on 2022.08.0 and confirm that they are missing the expected log directories.

### Workspace affected

_No response_

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

Front logo [Front conversations](https://app.frontapp.com/open/top_3r0zu)

MrSimonEmms:
## Description

This ensures that the `/gitpod/log-collector` directory exists on every node at FluentBit startup. Previously, this directory was only created when a log was added which meant that the `log-collector` section in a support bundle was full of `configmap-errors` saying that the directory didn't exist.

Now, this only adds files into the support bundle if there are files to add, meaning that it's much easier to search for things to look through.

## Related Issue(s)

Fixes #13095

## How to test

Install via KOTS and generate a support bundle

## Release Notes

```release-note
[kots]: ensure log-collector directory on node at startup
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/kots/manifests/gitpod-log-collector.yaml
+++ b/install/kots/manifests/gitpod-log-collector.yaml
@@ -31,11 +31,10 @@ spec:
             - sh
             - -c
           args:
-            - while true; do echo "waiting" && sleep 60; done
+            - mkdir -p /gitpod/log-collector && while true; do echo "waiting" && sleep 60; done
           volumeMounts:
             - name: collector
               mountPath: /gitpod
-              readOnly: true
 
       labels:
         app: gitpod


Pothulapati (approved):
Haven't tested on GKE specifically but did it on k3s, and I don't see any config errors.

PR #13240, opened 2022-09-23
Cleanup after #13093

geropl:
## Description
Removes reference to `install/preview:docker` build target

## Related Issue(s)

Context: previous PR removing said target:  #13093

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/BUILD.yaml
+++ b/components/BUILD.yaml
@@ -13,7 +13,6 @@ packages:
       - :publish-api
       - dev:all-app
       - install/installer:docker
-      - install/preview:docker
       - install/kots:lint
       - components/gitpod-protocol:all
   - name: docker-versions


mads-hartmann:
@geropl added hold because we if we remove it from that list I don't think we'll ever build the package. Is this urgent or something we can chat about Monday?

easyCZ (approved):

geropl:
/unhold

@mads-hartmann Package was removed [here](https://github.com/gitpod-io/gitpod/pull/13093/files#diff-68e96e59b87c10d293c07a63d809a4d5d261b3d117c19cc49666e1d2a80c351c), I'm just removing the reference so `leeway` works again :+1:

geropl:
@mads-hartmann Ah-a! Sorry for not reading up on that context. :bow:  But also, hearing about "dynamic packages" the first time :confused: 
My problem was that I wanted to run `yarn watch` (which is: `leeway exec --package .:app --transitive-dependencies --filter-type yarn --components --parallel -- yarn build -w --preserveWatchOutput`) and that failed with "cannot find package" (or similar). Do we want to sync somewhere else (Slack)?

mads-hartmann:
@geropl Let's chat Monday (I have one foot out the door ☺️) and find a way to add the component back to the top level without breaking things - I thought I had tested the different use-cases but seems I overlooked some ☺️

PR #13244, opened 2022-09-23
[installer]: allow the s3 connection to be insecure

### LINKED ISSUE #6776: [Installer]: Cannot use S3 object storage in DigitalOcean
> See [example](https://gist.github.com/MrSimonEmms/a770d6e36e3fdf8222b54626f2ae9d7d) to demonstrate the problem

DigitalOcean provides "Spaces", which is their version of S3-compatible storage. Despite having an identical API to S3, this works slightly differently to AWS S3 due to how the regions and endpoints are configured - the `endpoint` is "<regions>.digitaloceanspaces.com" - examples use `ams3`, but is the same for all regions. This is different AWS S3 which use a common `s3.amazonaws.com` and then sets the region separately.

# Attempt 1 - remove region from the URL

`endpoint := "digitaloceanspaces.com"`

This results in `XML syntax error on line 2: invalid character entity & (no semicolon)`. Likely because nothing exists on this URL

# Attempt 2 - setting region in URL and MakeBucket `Region`

The `MakeBucket` call cannot calculate the SHA256 header correctly, resulting in `The provided 'x-amz-content-sha256' header does not match what was computed`. Likely this is because DigitalOcean don't calculate the region as part of the header on their side because you cannot create, for example, an `fra1` resource on the URL `ams3`

# What works - setting region URL but not MakeBucket

If you remove the `Region` from the `MakeBucket` options, this works fine.

However, every part of Gitpod (the Installer, the ContentService etc) expects the `Region` to be set.

---

We get the same problem using Minio Gateway

# Suggested solutions

1. make `Region` an optional parameter. Theoretically, this would work, but it may have unintended consequences elsewhere especially for non-DO S3 configuration
2. don't support DigitalOcean Spaces - this would limit our self-hosted options. Furthermore, it may also be an issue with other non-AWS S3-compatible providers.

In additional to these, we could ask DigitalOcean for help and see if there's anything they can suggest.

---

# Workaround for today

Use Minio in-cluster

### LINKED ISSUE #9698: [Self-Hosted] minio PVC defaults to 8Gi only
We should fix two things:
 - choose a sane default that does work for some minimal testing (maybe 50Gi?)
 - provide means for users to (optionally) configure it without `yq`ing the resulting yaml

### LINKED ISSUE #12416: Minio rapidly increasing size (> 160GB) due to prebuilds not being deleted as expected
### Bug description

In our gitpod instance, we are currently 3 users with about 6 workspaces in total.
Our server did just run out of free storage space... after investigation, I found that the minio volume is about 160 GB now. Inside there are 3 user folders with more that a 100(!) subfolders in  `/var/lib/k3s/storage/pvc-..._gitpod_minio/gitpod-user-.../workspaces`.

So what are those? Snapshots? Because there is no way we had such many workspaces over all... and we didn't intentionally creating snapshots either. 
Today I had only one workspace active but the number of folder in my minio user directory kept growing. 

### Steps to reproduce

Just use the system and monitor the needed storage space

### Workspace affected

_No response_

### Expected behavior

gitpod should not create those files or at least automatically remove them.

### Example repository

_No response_

### Anything else?

The recommended `calico` backend did not work in our environment, so we kept the default `flannel` backend. But that seems unrelated.

MrSimonEmms:
## Description

We often tell people that, if they're unable to use the configuration of the in-cluster Minio that they should deploy their own to their cluster and connect to it over the Kubernetes core DNS. However, this isn't currently possible because the S3 connection expects for it to be over HTTPS. When using a local Kubernetes endpoint, this is likely to be over HTTP.

This allows for an S3 endpoint to use an insecure URL.

The default expectation of it being over HTTPS has not changed.

Fixes #6776
Fixes #9698 
Fixes #12416

## How to test

Deploy the [KubeCon demo](https://github.com/MrSimonEmms/gitpod-self-hosted-demo) instance (run `make` on an Ubuntu machine). We need this for KubeCon as the default 8Gi PVC is not large enough, so need to configure a larger PVC.

## Release Notes

```release-note
[installer]: allow the s3 connection to be insecure
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/installer/pkg/common/storage.go
+++ b/install/installer/pkg/common/storage.go
@@ -51,7 +51,7 @@ func StorageConfig(context *RenderContext) storageconfig.StorageConfig {
 				Endpoint:            context.Config.ObjectStorage.S3.Endpoint,
 				AccessKeyIdFile:     filepath.Join(storageMount, "accessKeyId"),
 				SecretAccessKeyFile: filepath.Join(storageMount, "secretAccessKey"),
-				Secure:              true,
+				Secure:              !context.Config.ObjectStorage.S3.AllowInsecureConnection,
 				Region:              context.Config.Metadata.Region,
 				ParallelUpload:      100,
 
--- a/install/installer/cmd/testdata/render/insecure-s3-setup/config.yaml
+++ b/install/installer/cmd/testdata/render/insecure-s3-setup/config.yaml
@@ -1,0 +1,16 @@
+# Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+# Licensed under the MIT License. See License-MIT.txt in the project root for license information.
+
+apiVersion: v1
+domain: gitpod.example.com
+metadata:
+  region: local
+objectStorage:
+  inCluster: false
+  s3:
+    endpoint: s3-provider.com
+    allowInsecureConnection: true
+    bucket: s3-bucket
+    credentials:
+      kind: secret
+      name: s3-storage
--- a/install/installer/cmd/testdata/render/insecure-s3-setup/output.golden
+++ b/install/installer/cmd/testdata/render/insecure-s3-setup/output.golden
@@ -771,7 +771,7 @@ metadata:
   name: dashboard
   namespace: default
 ---
-# v1/ServiceAccount dbinit
+# v1/ServiceAccount db
 apiVersion: v1
 automountServiceAccountToken: true
 kind: ServiceAccount
@@ -779,8 +779,8 @@ metadata:
   creationTimestamp: null
   labels:
     app: gitpod
-    component: dbinit
-  name: dbinit
+    component: db
+  name: db
   namespace: default
 ---
 # v1/ServiceAccount docker-registry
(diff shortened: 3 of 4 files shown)


aledbf:
@MrSimonEmms why? we should not say yes to everything. Having SSL for S3 is not too much to ask.

MrSimonEmms:
@aledbf it's so we can use a Minio instance that's deployed in the cluster, but not the in-cluster configuration that we provide. For the longest time, we've said that people are able to configure their own Minio instance and then talk to it locally, but this is not actually the case - the [in-cluster Minio](https://github.com/gitpod-io/gitpod/blob/main/install/installer/pkg/common/storage.go#L70) connects via HTTP and this allows a user to configure their own Minio on their machine and route through there.

One example is that this will allow users to configure S3 instances that don't use regions in the URL (eg, DigitalOcean/Storj.io) or where they want to use an in-cluster solution, but an 8Gi PVC isn't big enough (eg #9698 or #12416)

install/installer/pkg/common/storage.go
```diff
@@ -51,7 +51,7 @@ func StorageConfig(context *RenderContext) storageconfig.StorageConfig {
 				Endpoint:            context.Config.ObjectStorage.S3.Endpoint,
 				AccessKeyIdFile:     filepath.Join(storageMount, "accessKeyId"),
 				SecretAccessKeyFile: filepath.Join(storageMount, "secretAccessKey"),
-				Secure:              true,
+				Secure:              !context.Config.ObjectStorage.S3.AllowInsecureConnection,
```
nandajavarma, line 54: I am imagining this would become `true` by default because in Go `bool` has default value `false`. Testing just to be sure

nandajavarma: seems to be the case!

MrSimonEmms: Yup. I added two new golden file tests to prove it as well

nandajavarma:
Tests looked good! https://werft.gitpod-dev.com/job/gitpod-k3s-installer-tests-head.28

nandajavarma (approved):

PR #13245, opened 2022-09-23
[installer-tests] make domain name configurable

nandajavarma:
## Description

This PR makes the Domain name configurable for self hosted installer tests. The main drive here is to still run the tests even after we hit the let's encrypt cert limits. This is a very tiny workaround. The domain we provide should be registered under CloudDNS under the project `dns-for-playgrounds`(the project will change in the future) with a zone name same as the domain but `.` replaced by `-`. For eg, if the domain is `tests.doptig.com`, it expects the domain to be registered under the name `tests-doptig-com`(this is how we currently do it under `dns-for-playgrounds` project.

## Related Issue(s)

Fixes #

## How to test

You can something like the following:
```
werft run github -j .werft/k3s-installer-tests.yaml -a skipTests=true -a domain=tests.doptig.com -a subdomain=testing-abc
```

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/.werft/aks-installer-tests.yaml
+++ b/.werft/aks-installer-tests.yaml
@@ -4,6 +4,10 @@ args:
   desc: "Subdomain to use. A terraform workspace of same name will be used"
   required: false
   default: ""
+- name: domain
+  desc: "domain to use. Make sure there is a DNS zone by the same name ('.' replaces by '-' under the project 'dns-for-playgrounds')"
+  required: false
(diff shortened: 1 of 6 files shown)


adrienthebo (approved):
Only applies to the eks job, but @nandajavarma and I have discussed this in slack. 

/hold for AKS/GKE support, good otherwise

nandajavarma:
Added support on all platforms. /unhold

Issue #10835, opened 2022-06-22
GP CLI command to control port visibility


loujaybee:
**e.g.**

```gp ports visibility <port>:{public|private}...```

Previously part of this issue: 
- https://github.com/gitpod-io/gitpod/issues/9930

PR #13253, opened 2022-09-23
[gp-cli] add command to change ports visibility

mustard-mh:
## Description

Add new gp-cli command to allow users to change ports visibility

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/10835

## How to test

Open workspace in preview env, check with portsView to see if port is locked(private) or unlocked(public)

### Public
- Run `curl lama.sh | sh` to open port 8080
- Run `gp ports visibility 8080:public` to make it public
- Check PortsView / Visit URL of 8080 in another browser to check if it's visitable

### Private
- Run `gp ports visibility 8080:private` to make it private
- Check PortsView / Visit URL of 8080 in another browser to check if it's not visitable

## Release Notes

```release-note
[gp-cli] add command to change ports visibility
```

## Documentation

- https://github.com/gitpod-io/website/issues/2804

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-cli/cmd/ports-visibility.go
+++ b/components/gitpod-cli/cmd/ports-visibility.go
@@ -1,0 +1,65 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"strconv"
+	"strings"
+	"time"
+
+	gitpod "github.com/gitpod-io/gitpod/gitpod-cli/pkg/gitpod"
+	serverapi "github.com/gitpod-io/gitpod/gitpod-protocol"
+	"github.com/spf13/cobra"
+)
+
+// portsVisibilityCmd change visibility of port
+var portsVisibilityCmd = &cobra.Command{
+	Use:   "visibility <port:{private|public}>",
+	Short: "Make a port public or private",
+	Args:  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		portVisibility := args[0]
+		s := strings.Split(portVisibility, ":")
+		if len(s) != 2 {
+			log.Fatal("cannot parse args, should be something like `3000:public` or `3000:private`")
+		}
+		port, err := strconv.Atoi(s[0])
+		if err != nil {
+			log.Fatal("port should be integer")
+		}
+		visibility := s[1]
+		if visibility != serverapi.PortVisibilityPublic && visibility != serverapi.PortVisibilityPrivate {
+			log.Fatalf("visibility should be `%s` or `%s`", serverapi.PortVisibilityPublic, serverapi.PortVisibilityPrivate)
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+
+		wsInfo, err := gitpod.GetWSInfo(ctx)
+		if err != nil {
+			log.Fatalf("cannot get workspace info, %s", err.Error())
+		}
+		client, err := gitpod.ConnectToServer(ctx, wsInfo, []string{
+			"function:openPort",
+			"resource:workspace::" + wsInfo.WorkspaceId + "::get/update",
+		})
+		if err != nil {
+			log.Fatalf("cannot connect to server, %s", err.Error())
+		}
+		if _, err := client.OpenPort(ctx, wsInfo.WorkspaceId, &serverapi.WorkspaceInstancePort{
+			Port:       float64(port),
+			Visibility: visibility,
+		}); err != nil {
+			log.Fatalf("failed to change port visibility: %s", err.Error())
+		}
+		fmt.Printf("port %v is now %s\n", port, visibility)
+	},
+}
+
+func init() {
+	portsCmd.AddCommand(portsVisibilityCmd)
+}
--- a/components/gitpod-protocol/go/gitpod-service.go
+++ b/components/gitpod-protocol/go/gitpod-service.go
@@ -1766,6 +1766,11 @@ type WorkspaceInstancePort struct {
 	Visibility string  `json:"visibility,omitempty"`
 }
 
+const (
+	PortVisibilityPublic  = "public"
+	PortVisibilityPrivate = "private"
+)
+
 // GithubAppConfig is the GithubAppConfig message type
 type GithubAppConfig struct {
 	Prebuilds *GithubAppPrebuildConfig `json:"prebuilds,omitempty"`


components/gitpod-cli/cmd/ports-visibility.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+		port, err := strconv.Atoi(s[0])
+		if err != nil {
+			log.Fatal("port should be integer")
+		}
+		visibility := s[1]
+		if visibility != serverapi.PortVisibilityPublic && visibility != serverapi.PortVisibilityPrivate {
+			log.Fatalf("visibility should be `%s` or `%s`", serverapi.PortVisibilityPublic, serverapi.PortVisibilityPrivate)
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
```
andrew-farries, line 39: 15 seconds is a longer timeout than all other `gp` commands. Does it need to be this long?

components/gitpod-cli/cmd/ports-visibility.go
```diff
@@ -0,0 +1,65 @@
...
+		if err != nil {
+			log.Fatalf("cannot connect to server, %s", err.Error())
+		}
+		if _, err := client.OpenPort(ctx, wsInfo.WorkspaceId, &serverapi.WorkspaceInstancePort{
+			Port:       float64(port),
+			Visibility: visibility,
+		}); err != nil {
+			log.Fatalf("failed to change port visibility: %s", err.Error())
+		}
+		fmt.Printf("port %v is %s now\n", port, visibility)
```
andrew-farries, line 59:
```suggestion
		fmt.Printf("port %v is now %s\n", port, visibility)
```

andrew-farries (changes requested):

andreafalzetti (approved):
Tested and works as expected 💯 

/hold

components/gitpod-cli/cmd/ports-visibility.go
mustard-mh: Yep, I create this PR to test `openPort` so I increased this timeout duration 😂. Will reduce it to 5s

mustard-mh:
Updated

components/gitpod-protocol/go/gitpod-service.go
```diff
@@ -1766,6 +1766,11 @@ type WorkspaceInstancePort struct {
 	Visibility string  `json:"visibility,omitempty"`
 }
 
+const (
+	PortVisibilityPublic  string = "public"
+	PortVisibilityPrivate        = "private"
```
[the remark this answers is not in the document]
andreafalzetti: Makes sense, thanks for clarifying, @mustard-mh 🙏

components/gitpod-cli/cmd/ports-visibility.go
```diff
@@ -0,0 +1,65 @@
...
+// portsVisibilityCmd change visibility of port
+var portsVisibilityCmd = &cobra.Command{
+	Use:   "visibility <port>[:private][:public]",
+	Short: "Make a port public or private",
+	Args:  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		portVisibility := args[0]
+		s := strings.Split(portVisibility, ":")
+		if len(s) != 2 {
+			log.Fatal("cannot parse args, should be something like `3000:public` or `3000:private`")
```
filiptronicek, line 29: nit(non-blocking): I wonder why we ended up with this format of `PORT:VISIBILITY`. It reminds me of port bindings in Docker or SSH port-forwarding, but what I expected would be just delimited with a space, like 
```
gp ports visibility 3000 public
```

what I would find handy and would love to have an alias to this visibility command (maybe even rename it altogether 😝) is to make the syntax as follows:
```
gp ports make 3000 public
```

## Why?

- the command reads well and if someone were to dictate it to you over the phone, it'd be hard to "mess it up"
- it's 6 characters shorter
- doesn't use any special characters

filiptronicek (approved):
Thanks a lot for working on this, @mustard-mh 🙏. It's a really great feature I cannot wait to use!

components/gitpod-cli/cmd/ports-visibility.go
filiptronicek: Tagging @akosyakov as you were the first I could find to propose this format 👀 [[1](https://github.com/gitpod-io/gitpod/issues/9930#issuecomment-1131230528)]

andreafalzetti: "make port 3000 public" is very readable however, `make` is also quite generic and less intuitive at a glance that it's related to setting the port's visibility.

I am happy with the current approach, but happy if we discuss alternatives. My thoughts so far have been:

We could consider `set-visibility` but I am not fully convinced because:
1. makes the cmd longer to type (which is not a deal breaker if the name makes sense)
2. it's not consistent with other cmds (such as `gp timeout`). For the `timeout` subcommand we've used the verbs `extend` and `show` which at the same time, would make me wonder if the command introduced by this PR, should be `gp ports visibility set` (although we'd be back at point 1 --> longer cmd).

I then wondered how codespaces named this command and found the [same approach](https://docs.github.com/en/codespaces/developing-in-codespaces/using-github-codespaces-with-github-cli#modify-ports-in-a-codespace) so I think @mustard-mh could have used that as inspiration?

`gp ports visibility 3000:public` is good enough in a way because is short and clear. 

The bit that I am not 100% convinced with it either, is that I feel that it's missing the "set" semantic and joining port + visibility feels a bit hacky.

If I had to provide an alternative that I both like and think it's very flexible and clear, it would be: `gp ports set-visibility --port 3000 --visibility public` - however it's way longer and harder to remember probably.

I am curious to see what other people think.

andreafalzetti: To recap, we need to agree on the name of the command and how to provide the values.

For the `name`, our options are:

1️⃣ `gp ports visibility`
2️⃣ `gp ports visibility set`
3️⃣ `gp ports set-visibility`

For providing values:

1️⃣ `3000:public`
2️⃣ `3000 public`
3️⃣ `--port 3000 --visibility public` 

Any other options or suggestions? Please add or express your preferred way 🙏

filiptronicek: Thanks for sharing, Andrea! Agreed `make` is very generic in commands, so may not be the best option. Just had another idea: `gp ports publish 3000` and `gp ports private/hide 3000` i.e. - make the publish states verbs.

andreafalzetti: I've updated my comment above with `publish`/`hide` 👍 

Naming is always hard! I think finding the best name for a command is important as much as creating a nice UI on the IDE plugin. Thanks for caring and sparking the discussion :)

jeanp413: Codespaces uses the colon format `gh codespace ports visibility codespace-port:private|org|public -c codespace-name` so maybe it makes sense
I was thinking of something like `gp ports update 3000 --visibility=public` if we want to change more attributes for a port in the future like adding a label to it you can set that in vscode

andrew-farries (approved):

components/gitpod-cli/cmd/ports-visibility.go
mustard-mh: I'd like to unhold this PR, created an issue for our discussion. https://github.com/gitpod-io/gitpod/issues/13443

🙏 Thanks folks for your thoughts

mustard-mh:
Updated help message (see [diff](https://github.com/gitpod-io/gitpod/compare/307748f155b2abfd56306c0ef476a32cde8fbaf4..f0ae58652e63f03ed3471d2c2936b93aa508d116))

/unhold

akosyakov:
@mustard-mh Why don't you use supervisor API?

akosyakov:
Actually I see we do the same in VS Code.

PR #13257, opened 2022-09-23
Unified server prototype

easyCZ:
## Description

Prototype of a Go server hosting multiple services, across various components.

Experimental

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


csweichel:
I like that it's nice and simple, it's not magical. [Clear is better than clever.](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=875s)

A few thoughts:
- We will need tooling that prevents accidental cross-talk between packages in the same server.  Without such automated checks we'll inevitably end up with a hotchpotch of dependencies.
- The pattern is easy to copy for other teams. At our current size it makes sense to forgo a framework in favour of convention (which is only instantiated twice).
- The name needs rethinking. Personally I'd favour `server-webapp` or `webapp-server`. I know it's boring, but descriptive.

easyCZ:
> prevents accidental cross-talk between packages in the same server
Would you be able to elaborate on what you considers cross-talk? Are you referring to loops, or services depending on each other for a piece of data? Or are you referring one service does not import another service directly into it's implementation?

In my mind, talking between services is not a problem as long as we don't create loops. I'd imagine over time services looking like a pyramid. 

For example, we have the following "building blocks":
* Images for Workspaces
* Auth
* VCS Integrations

Which we then compose into more powerful services, like:
* Workspaces
* Prebuilds
* Previews

Which are then again composed together into:
* Usage
* Billing


When it comes to talking between services, I'd always expect a `grpc.Client` to be used which is pointed against `service:<port>`. This ensures that we don't just talk to our current instance, but we actually load-balance across available instances.

We could end up with a bootstrap problem - we're unable to start the server because we can't establish a connection to "service:1234" but service is hosted by our current instance. For that, we can do a couple of things: Use non-blocking `grpc.Dial` which will eventually resolve once it comes up, or we defer construction of the dialer into the RPC handling, and re-use the connection afterwards.


> The name needs rethinking. Personally I'd favour server-webapp or webapp-server. I know it's boring, but descriptive.
I've deliberely picked `a name` to focus on what it does, rather than block on naming :) Suggestions seem reasonable.

PR #13269, opened 2022-09-25
ws-mananger: refactor volume snapshot watcher

### LINKED ISSUE #13280: [PVC] The workspace with PVC keeps in terminating state
### Bug description

At the gen67 cluster, the workspace with the PVC feature flag keeps terminating.

The symptom is the workspace pod starts the disposal process, and the volume snapshot is ready to use.
But somehow, the event does not receive by the receiver _or_ does not send to the sender.

### Steps to reproduce

We don't know yet but it looks like we need to do more loadgen testing.

https://gitpod.slack.com/archives/C02F19UUW6S/p1663981079108569

### Workspace affected

_No response_

### Expected behavior

The workspace should be stopped after the VolumeSnapshot is ready to use.

### Example repository

_No response_

### Anything else?

#7901

jenting:
## Description

Refactor volume snapshot watcher.

Using the Go channel to notify different event workers that the volume snapshot is ready seems risky.
We observe that at gen67 cluster, some workspaces with PVC can't be terminated. The symptom is the workspace pod starts the disposal process, and the volume snapshot is ready to use. But somehow, the event does not receive by the receiver _or_ does not send to the sender.

We rewrite the watcher with the volume snapshot client watch + retry watcher to make sure the watcher happens in each event worker. To minimize the chance that the event does not receive by the receiver _or_ does not send to the sender.
Moreover, we fall back to exponential retry to get the volume snapshot object if we can't create the retry watcher.

## Related Issue(s)

Fixes #13280

## How to test

Create an ephemeral cluster, and run loadgen test with 100 regular workspaces + 100 regular workspaces with PVC.
Ensure no workspace pod is terminating for a long time.

## Release Notes

```release-note
None
```

## Documentation

None

## Werft options:

- [x] /werft with-preview
- [ ] /werft with-integration-tests=workspace


jenting:
I use the us68 cluster to test loadgen with 100 regular workspaces + 100 regular workspaces with PVC, I can reproduce the workspace in the terminating state.

After I patch the ws-manager with this PR and rerun the same test (100 regular workspaces + 100 regular workspace with PVC), I can't reproduce the workspace in terminating state.

Therefore, I think it's good to review.

components/ws-manager/pkg/manager/monitor.go
```diff
@@ -1150,31 +1100,65 @@ func (m *Monitor) finalizeWorkspaceContent(ctx context.Context, wso *workspaceOb
...
+				var volumeSnapshotWatcher *watchtools.RetryWatcher
+				volumeSnapshotWatcher, err = watchtools.NewRetryWatcher("1", &cache.ListWatch{
+					WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+						return m.manager.VolumeSnapshotClient.SnapshotV1().VolumeSnapshots(m.manager.Config.Namespace).Watch(ctx, metav1.ListOptions{
+							FieldSelector: fields.OneTermEqualSelector("metadata.name", pvcVolumeSnapshotName).String(),
+						})
+					},
+				})
+				if err != nil {
+					log.Info("fall back to exponential backoff retry")
```
sagor999, line 1114: can we log the error here?

sagor999 (approved):
/hold

LGTM, but hold for minor nit.

jenting:
nit addressed.
/unhold

PR #13272, opened 2022-09-25
[installer]: only allow config patch if advanced mode enabled

MrSimonEmms:
## Description

Spotted a regression in the Installer refactoring where a config patch could be applied if advanced mode isn't enabled. This fixes it.

## Release Notes

```release-note
[installer]: only allow config patch if advanced mode enabled
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/installer/pkg/config/v1/envvars.go
+++ b/install/installer/pkg/config/v1/envvars.go
@@ -370,6 +370,18 @@ func (v version) BuildFromEnvvars(in interface{}) error {
 		}
 	}
 
+	if telemetryValue := envvars.Distribution; telemetryValue != "" {
+		log.Infof("Setting deployment platform: %s", telemetryValue)
+		if cfg.Experimental == nil {
+			cfg.Experimental = &experimental.Config{}
+		}
+
+		if cfg.Experimental.Telemetry == nil {
+			cfg.Experimental.Telemetry = &experimental.TelemetryConfig{}
+		}
+		cfg.Experimental.Telemetry.Data.Platform = telemetryValue
+	}
+
 	cfg.Components = nil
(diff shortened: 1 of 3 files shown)


install/installer/pkg/config/v1/envvars.go
```diff
@@ -418,7 +418,7 @@ func (v version) BuildFromEnvvars(in interface{}) error {
 		cfg.Experimental.Telemetry.Data.Platform = telemetryValue
 	}
 
-	if cfgPatch := envvars.ConfigPatch; cfgPatch != "" {
+	if cfgPatch := envvars.ConfigPatch; envvars.AdvancedModeEnabled && cfgPatch != "" {
```
Pothulapati, line 421: nit: I understand that the `ConfigPatch` has to be the last change to be applied (as the expectation here is that it can override anything). How about we [move the advanced mode toggle](https://github.com/gitpod-io/gitpod/blob/80d98f25fcde9e8629cb4d157bf3232a161964eb/install/installer/pkg/config/v1/envvars.go#L375) a bit below here so that we can have all the advanced mode settings at the same place? 🤔

Pothulapati (approved):
LGTM, and good find!

Holding on the nit, but feel free to move forward if you think we can do it later!
/hold

install/installer/pkg/config/v1/envvars.go
MrSimonEmms: You know what, you're right. I was 50/50 when I first did it (mainly because it makes the lines of code changed much larger) - your suggestion pushes me over 50/50 into one option

Pothulapati (approved):
:shipit:

PR #13241, opened 2022-09-23
[content-service] Fix git checkout

aledbf:
## Description

Avoid reverting the changes https://github.com/gitpod-io/gitpod/pull/13210 and https://github.com/gitpod-io/gitpod/pull/13053

## Test

Open different workspaces that triggered the changes

- https://aledbf-git.preview.gitpod-dev.com/#https://github.com/gitpod-io/dazzle
- https://aledbf-git.preview.gitpod-dev.com/#https://github.com/gitpod-io/dazzle/commit/ddbf82c1a4dc95bda26c9f299d8e3b8f84eb659d
- https://aledbf-git.preview.gitpod-dev.com/#prebuild/https://github.com/gitpod-io/dazzle
- https://aledbf-git.preview.gitpod-dev.com/#https://github.com/NixOS/nixpkgs
- https://aledbf-git.preview.gitpod-dev.com/#prebuild/https://github.com/NixOS/nixpkgs
- https://aledbf-git.preview.gitpod-dev.com/#https://github.com/gitpod-io/spring-petclinic/tree/ak/vmoptions_test
- https://gitpodio-empty-qkchjpk699u.ws.aledbf-git.preview.gitpod-dev.com/
- https://aledbf-git.preview.gitpod-dev.com#prebuild/https://github.com/gitpod-io/empty

## Release Notes
```release-note
NONE
```

## Werft options:

- [ ] /werft with-local-preview
      If enabled this will build `install/preview`
- [x] /werft with-preview
- [ ] /werft with-integration-tests=all
      Valid options are `all`, `workspace`, `webapp`, `ide`

PR #13277, opened 2022-09-26
content-service: revert git ls-remote with --exit-code flag

jenting:
## Description

Revert git ls-remote with --exit-code flag

## Related Issue(s)

Fixes #

## How to test

Running the integration test

## Release Notes

```release-note
None
```

## Documentation

None

## Werft options:

- [x] /werft with-preview
- [x] /werft with-integration-tests=workspace


jenting:
/hold

Waits for the integration test pass

kylos101:
:wave: @jenting can you share context for this PR? For example, the issue you linked to is closed. Is there a new issue or problem that was reported?

jenting:
> 👋 @jenting can you share context for this PR? For example, the issue you linked to is closed. Is there a new issue or problem that was reported?
> 
> edit: I assume you are referring to this comment? [#13241 (comment)](https://github.com/gitpod-io/gitpod/pull/13241#issuecomment-1257430882)

Yes, because we could fix it within 5 mins, so I did not bring up the issue originally.
I just create the issue and update the description.

jenting:
This PR is blocked by this issue #13284.

utam0k:
I have one question. Should this modify the test, or is this something that contentInitializer should handle (or ignore)?

jenting:
> I have one question. Should this modify the test, or is this something that contentInitializer should handle (or ignore)?

I use Jaeger to catch the initializer sent from the server, and it does not contain the _TargetMode_ and _CloneTarget_ from the start workspace request. That's why we can't reproduce it when we open the workspace from Gitpod GUI.

For the short term, we could fix it with an integration test.

But I am thinking about the long term. We should change the workspace integration test that sends requests from integration test client -> server -> ws-manager, rather than from integration test client-> ws-manager.
After that, we could test against the real code logic, WDYT?

utam0k:
> For the short term, we could fix it with an integration test.

> But I am thinking about the long term. We should change the workspace integration test that sends requests from integration test client -> server -> ws-manager, rather than from integration test client-> ws-manager.
After that, we could test against the real code logic, WDYT?

Make sense :100: Please put TODO comment about it to the integration test.

utam0k (approved):

PR #13292, opened 2022-09-26
[server][github] record "pull" webhook events

### LINKED ISSUE #12817: Webhook events are not registered properly for GitHub PRs
On GitHub PR updates we are missing proper `pull_request.synchronize` handling for webhook event registration. 

Also, it we're triggering two prebuilds for the same commit, one for the PR event (context URL is the PR,) and another for the `push` event (context URL is the branch.)

Follow-up to https://github.com/gitpod-io/gitpod/pull/11356

AlexTugarev:
## Description
Before this PR, only push events were recorded. Because Pull Requests updates are also triggers for prebuilds, they should be registered as well.

## Related Issue(s)

Fixes #12817

## How to test
Inspect `<PREVIEW-URL>/t/<TEAM>/<PROJECT>/events` for events recorded for updates of pull requests on `PROJECT` repository. 

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/server/ee/src/prebuilds/github-app.ts
+++ b/components/server/ee/src/prebuilds/github-app.ts
@@ -181,12 +181,17 @@ export class GithubApp {
             options.getRouter("/reconfigure").get(
                 "/",
                 asyncHandler(async (req: express.Request, res: express.Response) => {
-                    const gh = await app.auth();
-                    const data = await gh.apps.getAuthenticated();
-                    const slug = data.data.slug;
+                    try {
+                        const gh = await app.auth();
+                        const data = await gh.apps.getAuthenticated();
+                        const slug = data.data.slug;
 
-                    const state = req.query.state;
-                    res.redirect(`https://github.com/apps/${slug}/installations/new?state=${state}`);
+                        const state = req.query.state;
+                        res.redirect(`https://github.com/apps/${slug}/installations/new?state=${state}`);
+                    } catch (error) {
+                        console.error(error, { error });
+                        res.status(500).send("GitHub App is not configured.");
+                    }
                 }),
             );
         options.getRouter &&
@@ -398,6 +403,12 @@ export class GithubApp {
         const span = TraceContext.startSpan("GithubApp.handlePullRequest", {});
         span.setTag("request", ctx.id);
 
+        const event = await this.webhookEvents.createEvent({
+            type: ctx.name,
+            status: "received",
+            rawEvent: JSON.stringify(ctx.payload),
+        });
+
         try {
             const installationId = ctx.payload.installation?.id;
             const cloneURL = ctx.payload.repository.clone_url;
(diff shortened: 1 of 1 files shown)


components/server/ee/src/prebuilds/github-app.ts
```diff
@@ -423,14 +434,37 @@ export class GithubApp {
...
+                await this.webhookEvents.updateEvent(event.id, {
+                    prebuildStatus: "ignored_unconfigured",
+                    status: "processed",
+                });
             }
         } catch (e) {
             TraceContext.setError({ span }, e);
+            await this.webhookEvents.updateEvent(event.id, {
+                prebuildStatus: "prebuild_trigger_failed",
+                status: "processed",
```
easyCZ, line 466: Should the status be `processed` here? We failed to do something, do we know it will actually be created?

components/server/ee/src/prebuilds/github-app.ts
```diff
@@ -181,12 +181,17 @@ export class GithubApp {
...
+                    try {
+                        const gh = await app.auth();
+                        const data = await gh.apps.getAuthenticated();
+                        const slug = data.data.slug;
+
+                        const state = req.query.state;
+                        res.redirect(`https://github.com/apps/${slug}/installations/new?state=${state}`);
+                    } catch (error) {
+                        console.error(error, { error });
+                        res.status(500).send("GitHub App is not configured.");
```
easyCZ, line 193: Are all errors 500 errors? Can it happen that `const data = await gh.apps.getAuthenticated();` gives us a Not Found?

I'm trying to understand if all errors are in fact Server Errors, or if there are modes which fail due to user errors etc. Always returning a 500 has the obvious impact to our SLOs and may not always communicate the underlying problem to the user well.

easyCZ (commented):
Couple of questions to help me understand this PR better.

components/server/ee/src/prebuilds/github-app.ts
AlexTugarev: > Can it happen that const data = await gh.apps.getAuthenticated(); gives us a Not Found?

Don't think so. If `auth` succeed, it should be an in scope of an authenticated app, thus `getAuthenticated` should return its details. 
https://docs.github.com/en/rest/apps/apps#get-the-authenticated-app

AlexTugarev: > I'm trying to understand if all errors are in fact Server Errors

👍🏻 
In that case we're navigating to `/reconfigure`, but there is an issue with GH App settings. Example for this is in preview enviroments, or self-hosted installations.

easyCZ: What about if `auth` doesn't succeed? Does it still make sense to return a 500, even if the reason is that the user doesn't exist or it has an invalid access token?

I'd expect here that we only return a 500 if we fail for a reason that's not expected - Users not being found or not existing are expected reasons.

AlexTugarev: Actually, there was a GH issue to do exactly this. I've hard times to find it, though.

AlexTugarev: > What about if auth doesn't succeed? Does it still make sense to return a 500, even if the reason is that the user doesn't exist or it has an invalid access token?

@easyCZ, I see. `app.auth` is the request to use the configured GH App secrets to do the app authentication. Here it's the GH App of Gitpod which needs to do the handshake, in order to be auth'd for operations. That's not related to users. More on this: https://docs.github.com/en/developers/apps/building-github-apps/authenticating-with-github-apps#authenticating-as-a-github-app

If app authentication is failing, reason is most likely the app secrets aren't set up properly, or the app is not configured. Otherwise this endpoint should not be navigable from UI.

AlexTugarev: To sum up: the code 500 isn't new here.
The effect of this change:
1. readable error message to users, instead of error stack trace
2. logging of the actual error/reason in logs, instead of swallowing it

easyCZ (approved):
Thanks for the clarifications, appreciate transferring context and helping me understand the impact better. While I'd like that we report more accurate status codes, it's not a reason to block this PR.

PR #13295, opened 2022-09-26
[server][prebuild] assert no prebuild is running for same commit

### LINKED ISSUE #11757: Two prebuilds are created for one pushed commit
### Bug description

It seems like two prebuilds are created for each one commit, which results in two pods on the workspace cluster.

image

### Steps to reproduce

Push a commit to a Github repo that is configured as a project in Gitpod whereincremental prebuilds are disabled

### Workspace affected

n/a

### Expected behavior

Only one prebuild workspace should be configured for each commit

### Example repository

https://github.com/kylos101/prebuild-experiment

### Anything else?

You can see the two prebuilds for one commit here:
```sql
select * from gitpod.d_b_prebuilt_workspace dbpw 
where dbpw.`commit` = '2304e76f4b97c3850d97d83476e8f9f1cd0f84ed'
and dbpw.cloneURL = 'https://github.com/kylos101/prebuild-experiment.git'
limit 10
```

The workspaces:
* https://gitpod.io/admin/workspaces/kylos101-prebuildexperi-s3qb21p0t36
* https://gitpod.io/admin/workspaces/kylos101-prebuildexperi-8rbic993etc

AlexTugarev:
## Description
The process of triggering a prebuild is async and competing events may cause a race which results in two prebuilds being started for the same commit. This practical PR introduces repeated check for already started prebuild to reduce the likelihood of prebuilds started for the same commit.

## Related Issue(s)

Fixes #11757

## How to test
Note, this is easier to verify with https://github.com/gitpod-io/gitpod/pull/13292 been merged before, because one can inspect webhook events and compare with prebuild triggered.

Nevertheless, to verify a single prebuild is produced one needs:
* setup a project for a repo
* create a PR on that test repo which would produce a prebuild, see a prebuild is running
* verify that a PR update (force pushing a significant change, to produce a new build) is also triggering just a single prebuild per commit

## Release Notes

```release-note
Avoid second prebuild been triggered on same commit.
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/server/ee/src/workspace/workspace-factory.ts
+++ b/components/server/ee/src/workspace/workspace-factory.ts
@@ -89,18 +89,29 @@ export class WorkspaceFactoryEE extends WorkspaceFactory {
             const { project, branch } = context;
 
             const commitContext: CommitContext = context.actual;
-            const existingPWS = await this.db
-                .trace({ span })
-                .findPrebuiltWorkspaceByCommit(
-                    commitContext.repository.cloneUrl,
-                    CommitContext.computeHash(commitContext),
-                );
-            if (existingPWS) {
+
+            const assertNoPrebuildIsRunningForSameCommit = async () => {
+                const existingPWS = await this.db
+                    .trace({ span })
+                    .findPrebuiltWorkspaceByCommit(
+                        commitContext.repository.cloneUrl,
+                        CommitContext.computeHash(commitContext),
+                    );
+                if (!existingPWS) {
+                    return;
+                }
+                log.debug("Found existing prebuild in createForStartPrebuild.", {
+                    context,
+                    cloneUrl: commitContext.repository.cloneUrl,
+                    commit: CommitContext.computeHash(commitContext),
+                });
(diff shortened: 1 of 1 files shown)


geropl:
Testing now...

components/server/ee/src/workspace/workspace-factory.ts
```diff
@@ -89,18 +89,29 @@ export class WorkspaceFactoryEE extends WorkspaceFactory {
...
             const commitContext: CommitContext = context.actual;
-            const existingPWS = await this.db
-                .trace({ span })
-                .findPrebuiltWorkspaceByCommit(
-                    commitContext.repository.cloneUrl,
-                    CommitContext.computeHash(commitContext),
-                );
-            if (existingPWS) {
+
+            const asserNoPrebuildIsRunningForSameCommit = async () => {
```
geropl, line 93: Would be nice to have a comment here (or on usage-sites below) that summarizes _why_ we're calling it repeatedly (because we're being pragmatic).

AlexTugarev: done

PR #13308, opened 2022-09-26
[kots]: remove the "require min RBAC" boolean

MrSimonEmms:
## Description

This prevents the preflight checks from being run in some environments (eg, EKS) as there are elevated privileges required to interrogate nodes etc.

There was a misunderstanding from the KOTS docs made in #13168 which I thought created the `kotsadm` service account. This service account appears to always be created, but this boolean limits it's capabilities (in retrospect, that's obvious. But it wasn't at the time - we caught it before it was released to the public anyway).

## How to test

Deploy via KOTS on AWS. Preflight checks should run - previously, they were not running due to lack of cluster role permissions. With this fix, they should run.

## Release Notes

```release-note
[kots]: remove the "require min RBAC" boolean
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/kots/manifests/kots-app.yaml
+++ b/install/kots/manifests/kots-app.yaml
@@ -10,7 +10,6 @@ spec:
   icon: ""
   allowRollback: true
   kubectlVersion: ">= 1.21.0"
-  requireMinimalRBACPrivileges: true
   # daemonsets are not supported yet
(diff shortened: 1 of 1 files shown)


nandajavarma (approved):
Thanks for the quick fix!!

PR #13309, opened 2022-09-26
[supervisor]: add grpc observability

akosyakov:
## Related Issue(s)

Fixes #

## Release Notes

```release-note
```

## Werft options:

- [x] /werft with-preview


roboquat:

<details>

</details>

akosyakov:
> We deliberately turned off these metrics for the supervisor because it consumes a lot of monitoring storage and traffic

@iQQBot yes i remember, but it is our main interface for IDEs, so we need this data cc @ArthurSens @meysholdt just to heads up about possible costs implications

akosyakov:
@iQQBot As an alternative we report it via IDE proxy so it does not count workspaces?

components/supervisor/pkg/supervisor/supervisor.go
```diff
@@ -1107,6 +1108,15 @@ func startAPIEndpoint(ctx context.Context, cfg *Config, wg *sync.WaitGroup, serv
...
 	}
 
+	grpcMetrics := grpc_prometheus.NewServerMetrics()
+	grpcMetrics.EnableHandlingTimeHistogram(
+		grpc_prometheus.WithHistogramBuckets([]float64{.005, .025, .05, .1, .5, 1, 2.5, 5, 30, 60, 120, 240, 600}),
+	)
+	opts = append(opts,
+		grpc.StreamInterceptor(grpcMetrics.StreamServerInterceptor()),
+		grpc.UnaryInterceptor(grpcMetrics.UnaryServerInterceptor()),
+	)
```
easyCZ, line 1118: If you use the standard `grpc server` options defined in `common-go/grpc` you'll get this for free, and consistent with other servrs.
The alternative is to use `baseserver` for the gRPC server here to get the same benefits.

akosyakov: https://gitpod.slack.com/archives/C01KGM9BH54/p1664192086999129?thread_ts=1664192035.420599&cid=C01KGM9BH54

akosyakov:
> @iQQBot As an alternative we report it via IDE proxy so we don't have a label for workspace pods?

@iQQBot I would try to go with as it is now, since it is simpler on implementation side and if there is indeed significant cost reconsider it later.

components/supervisor/pkg/supervisor/supervisor.go
akosyakov: > If you use the standard grpc server options defined in common-go/grpc you'll get this for free, and consistent with other servrs.

@easyCZ Could you point me to the code please?

akosyakov: is it https://github.com/gitpod-io/gitpod/blob/3f76aaa0c9140dc7489f0430d1dee1510c11ee8d/components/common-go/grpc/grpc.go#L92 ?

easyCZ: Yes, indeed. Or you can also use the higher level function [here](https://github.com/gitpod-io/gitpod/blob/3f76aaa0c9140dc7489f0430d1dee1510c11ee8d/components/common-go/grpc/grpc.go#L86)

easyCZ: Let me know if you'd like to pair on this. Also fine to not do this if there's a reason not to re-use the common metrics setup.

akosyakov: Supervisor has different lifecycle compare to other services, i.e. it belongs to the workspace pod which is ephemeral and it can cause for instance high cardinality issues because of too many workspaces. I think we will most likely reconsider how we collect metrics, i.e. use push gateway. I would prefer to keep it separated for more control for now. If it turned out to be different, i.e. we need to port already solved issues from baseserver, we ca reconsider later.

easyCZ: Makes sense. Thanks for the context!

akosyakov:
/hold

to get green light from @ArthurSens and @meysholdt

ArthurSens:
I'd say let's go for it, if we find problems we can always revert!

akosyakov:
/unhold

@iQQBot could you review please?

iQQBot (approved):
LGTM

PR #12965, opened 2022-09-14
Fix: remove global S3 access policy

mrzarquon:
## Description
Fixes https://github.com/gitpod-io/gitpod/issues/12964

We no longer need to attach these policies to the S3 users we create in our terraform modules.

```release-note
No longer attaches `arn:aws:iam::aws:policy/AmazonS3FullAccess` to the IAM user accounts created in our AWS terraform reference infrastructure. This level of access is no longer required now that one can provide an S3 bucket to be used for object storage.
```

PR #13317, opened 2022-09-26
[terraform] Give AWSS3FullAccess to s3 IAM users

### LINKED ISSUE #13017: [bug] Fix the S3 backend registry pre-flight check failure for eks automated test

nandajavarma:
## Description

This PR essentially reverts [this one](https://github.com/gitpod-io/gitpod/pull/12965/files). Even though with AWS CLI the permission seems enough, `mc` cli that we use for preflights(probably for other ops in Gitpod, not sure) requires a lot more permissions to do S3 ops. This is made clean in their documentation [here](https://github.com/minio/mc#example---amazon-s3-cloud-storage).

## Related Issue(s)

Fixes #13017 

## How to test

You can run the pipeline using:
```
werft run github -j .werft/eks-installer-tests.yaml -a deps=external -a domain=tests.doptig.com
```

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/infra/modules/eks/storage.tf
+++ b/install/infra/modules/eks/storage.tf
@@ -21,45 +21,16 @@ resource "aws_s3_bucket_versioning" "storage" {
   }
 }
 
-data "aws_iam_policy_document" "s3_policy" {
-  count = var.create_external_storage ? 1 : 0
-  statement {
-    actions   = [
-      "s3:PutObject",
-      "s3:ListMultipartUploadParts",
-      "s3:GetObject",
-      "s3:DeleteObject",
-      "s3:AbortMultipartUpload"
-    ]
-    resources = ["${aws_s3_bucket.gitpod-storage[count.index].arn}/*"]
-    effect    = "Allow"
-  }
-      statement {
-    actions   = ["s3:ListBucket",
-                "s3:GetBucketLocation",
-                "s3:ListBucketMultipartUploads"]
-    resources = [aws_s3_bucket.gitpod-storage[count.index].arn]
-    effect    = "Allow"
-  }
-}
-
-resource "aws_iam_policy" "policy" {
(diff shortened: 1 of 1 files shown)


mrzarquon:
@nandajavarma since we are handing this out to customers as example reference architecture, I don't think we should be including extremely broad permissions that only need to exist for our CI jobs to work properly.

adrienthebo:
> @nandajavarma since we are handing this out to customers as example reference architecture, I don't think we should be including extremely broad permissions that only need to exist for our CI jobs to work properly.
> 
> Double checking - is this the S3 preflight check done in our CI or in KOTS? In either case, it these are the minimum required permissions for our services to function, requiring them to be more permissive just because of how we want to validate them isn't a good idea.

I believe that this is the [preflight check](https://github.com/gitpod-io/gitpod/blob/main/components/kots-config-check/storage/entrypoint.sh#L107-L139) in KOTS.

That said, I believe our reference architecture documentation uses a [reduced permission set](https://www.gitpod.io/docs/self-hosted/latest/reference-architecture/single-cluster-ref-arch#object-storage), will double check that.

nandajavarma:
Have created these follow up issues to fix the preflight check and reduce the permissions:
- https://github.com/gitpod-io/gitpod/issues/13332
- https://github.com/gitpod-io/gitpod/issues/13333

@adrienthebo @mrzarquon Do you think we can ship this now for the sake of the release and take up the follow-ups after?

adrienthebo (approved):
Since the Terraform module isn't shipped as part of the KOTS release and we're running on a schedule this week, let's merge this change, cut the release branch, and then clean up the IAM permissions right after the release. We take on minimal liability with this and merging this change keeps the release on track.

🚢

PR #13117, opened 2022-09-20
Fix propagation of PVC feature flag across prebuild/regular workspaces

geropl:
## Description
This ensures that:
 - prebuilds follow the `usePersistentVolumeClaim` project setting
 - on workspace start, `persistent_volume_claim` can be added (if user enabled that flag) but not removed (e.g. in case a prior instance had that feature-flag enabled)

:question: Is that what we want to achieve? :thinking: 

## Related Issue(s)

Fixes #12718 

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview
- [ ] /werft with-integration-tests=all
      Valid options are `all`, `workspace`, `webapp`, `ide`

PR #13341, opened 2022-09-26
allow to enable PVC feature flag for user via ConfigCat

### LINKED ISSUE #12745: Use ConfigCat to throttle adoption of PVC
Presently, we rely on a few things to determine if a workspace should start with PVC.

1. They are restoring [a workspace that previously used PVC](https://github.com/gitpod-io/gitpod/blob/1fe5162c9feb7adc088ca9ad1de4dd9ed1d35260/components/server/src/workspace/config-provider.ts#L130-L136) (either from a prebuild or prior run of a regular workspace).
2. A new workspace is being created, [the user has the `persistent_volume_claim` feature flag enabled, while also being enabled for workspace classes.](https://github.com/gitpod-io/gitpod/blob/1fe5162c9feb7adc088ca9ad1de4dd9ed1d35260/components/server/src/workspace/workspace-starter.ts#L897-L902)
3. A new workspace is being created from the context of a [prebuild workspace that was run with PVC](https://github.com/gitpod-io/gitpod/blob/1fe5162c9feb7adc088ca9ad1de4dd9ed1d35260/components/server/src/workspace/workspace-starter.ts#L865-L874) (which can happen at least two ways, 1. the project has PVC enabled or 2. the prebuild workspace has the `persistent_volume_claim` feature flag persisted)

For Case 2 (above), create a new configcat setting, and add a rule that uses `Target % of users`, which'll control the % of regular workspace starts that include the `persistent_volume_claim` feature flag. From a default perspective, for self-hosted, I think we'll want the new config setting to default to false, at least for now.

Case 1 should not need any changes. Case 3 is already covered by `persistent_volume_claim` (Prebuilds with Persistent Volume Claim) in configcat.

Then, in configcat, for both rules, we could be flexible with how we roll this out. 

[Internal conversation about how to configure rules in ConfigCat.](https://gitpod.slack.com/archives/C02EN94AEPL/p1662673410899989)

sagor999:
## Description

Instead of having to enable feature flag per user via admin panel, this allows to enable it via ConfigCat instead, as well as will allow us to roll out this feature to more users via configCat. 

## Related Issue(s)

Fixes #12745

## How to test

Open preview env
Add your user id to ConfigCat non production env
Open new workspace
In terminal: `ls -la /.workspace/.gitpod/` and you should see `pvc` file in there, indicating this workspace is running using PVC feature.

## Release Notes

```release-note
none
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/server/ee/src/workspace/workspace-factory.ts
+++ b/components/server/ee/src/workspace/workspace-factory.ts
@@ -32,12 +32,15 @@ import { UserDB } from "@gitpod/gitpod-db/lib";
 import { UserCounter } from "../user/user-counter";
 import { increasePrebuildsStartedCounter } from "../../../src/prometheus-metrics";
 import { DeepPartial } from "@gitpod/gitpod-protocol/lib/util/deep-partial";
+import { EntitlementService } from "../../../src/billing/entitlement-service";
+import { getExperimentsClientForBackend } from "@gitpod/gitpod-protocol/lib/experiments/configcat-server";
 
 @injectable()
 export class WorkspaceFactoryEE extends WorkspaceFactory {
     @inject(LicenseEvaluator) protected readonly licenseEvaluator: LicenseEvaluator;
     @inject(HostContextProvider) protected readonly hostContextProvider: HostContextProvider;
     @inject(UserCounter) protected readonly userCounter: UserCounter;
+    @inject(EntitlementService) protected readonly entitlementService: EntitlementService;
 
     @inject(UserDB) protected readonly userDB: UserDB;
 
@@ -332,6 +335,18 @@ export class WorkspaceFactoryEE extends WorkspaceFactory {
 
             // Special case for PVC: While it's a workspace-persisted feature flag, we support the upgrade path (non-pvc -> pvc), so we apply it here
             if (user.featureFlags?.permanentWSFeatureFlags?.includes("persistent_volume_claim")) {
+                config._featureFlags = (config._featureFlags || []).concat(["persistent_volume_claim"]);
+            }
+            const billingTier = await this.entitlementService.getBillingTier(user);
+            const userTeams = await this.teamDB.findTeamsByUser(user.id);
+            // this allows to control user`s PVC feature flag via ConfigCat
+            if (
+                await getExperimentsClientForBackend().getValueAsync("user_pvc", false, {
+                    user,
+                    teams: userTeams,
+                    billingTier,
+                })
+            ) {
                 config._featureFlags = (config._featureFlags || []).concat(["persistent_volume_claim"]);
             }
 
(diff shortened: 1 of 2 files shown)


components/server/src/workspace/workspace-starter.ts
```diff
@@ -867,6 +867,19 @@ export class WorkspaceStarter {
                 featureFlags = featureFlags.concat(["protected_secrets"]);
             }
 
+            if (
+                // ConfigCat already has feature flag named `persistent_volume_claim` that controls if we show
+                // the option to enable PVC for project prebuilds. So have to use a different name here.
+                await getExperimentsClientForBackend().getValueAsync("user_pvc", false, {
```
sagor999, line 873: webapp team, when reviewing, can you double check that this is indeed the only place where I need to add this code?

geropl:
Starting review now...

geropl (changes requested):
The places where we have to replace/OR the existing logic are these:
 - workspaces based on prebuilds: https://github.com/gitpod-io/gitpod/blob/9993ba6959dff56779aa3fe66991d2f102dbd358/components/server/ee/src/workspace/workspace-factory.ts#L334
 - on workspace creation: https://github.com/gitpod-io/gitpod/blob/f2656e83c3e42d61a9c1c612a5d445162ba3d088/components/server/src/workspace/config-provider.ts#L128

sagor999:
@geropl updated code. Hopefully I did it the right way, not sure if adding all those injects all over the place is the correct way of doing it though. :sweat_smile:

components/server/src/workspace/workspace-starter.ts
geropl: This change can actually go away, the other two are fine! :+1:

sagor999: Hm. Are you sure? Since I originally tested with this change only and it worked as expected. So want to double check before I remove this code. :thinking:

geropl: > Are you sure?

Yes. That's the core of the [PR from last week](https://github.com/gitpod-io/gitpod/pull/13117): `persistent_volume_claim` is a "workspace persistent" feature, so it should never be required to modify on start, only on creation.

This PR just changes the source from where we get the info from, so we should be able to simply follow all occurrences of `user.featureFlags?.permanentWSFeatureFlags?.includes("persistent_volume_claim")` and that check there (as you already did).

sagor999: @geropl ah, good to know. thank you! code removed. PTAL

jenting:
@geropl PTAL

easyCZ (approved):

Issue #9198, opened 2022-04-08
Epic: Get rid of db-sync


geropl:
### Summary

This is the epic for getting rid of `db-sync`, which was already discussed at length in the context of "Simplified Multi-Meta" in [this RFC](https://www.notion.so/gitpod/Simplified-Multi-Meta-9942f49ccfbe47b0a7e352970df13162).

### Context

 - basically summed up in [the RFC](https://www.notion.so/gitpod/Simplified-Multi-Meta-9942f49ccfbe47b0a7e352970df13162)
 - previous epic: #7435

### Value

 - enables more Gitpod clusters (e.g., asia) without adding yet another DB
 - enables delivering multi-cluster Gitpod to Self-Hosted customers (because it reduces setup complexity)

### Acceptance Criteria

 - we got rid of `db-sync`

### Measurement

 - when there is no `db-sync` deployment anymore, and we're still able to deploy to prod :upside_down_face: 

### Notion

Discussion related to this epic is [in Notion](https://www.notion.so/gitpod/WebApp-Epic-Get-rid-of-db-sync-a790d1d453fa482aa5041a8e0be39c28).

### Tasks

- [x] https://github.com/gitpod-io/gitpod/issues/14560
- [x] Create a sidecar container to configure Toxiproxy with a `latency` toxic.
- [x] Run the above sidecar in the Toxiproxy pod.
- [x] Deploy a second `slow-server` deployment which connects to `mysql` through Toxiproxy.
- [x] Allow `proxy` to route websocket connections with the `slow-database` protocol header to the `slow-server` deployment.
- [x] Allow setting of the `slow-database` websocket connection header from websocket connections from dashboard based on the value of a feature flag.
- [x] #14958
- [x] #14960
- [x] #15027
- [x] #14962
- [x] #14963
- [x] #14964
- [x] #14965
- [x] #14995
- [ ] #14996
- [ ] #14997
- [ ] #14998
- [x] #15297
- [x] #15298
- [x] #15328
- [ ] [db-sync] Look at slowest queries on Grafana and configure caching for them. (typeorm allows caching on a per-query basis).
- [x] #15655

geropl:
Old workspace cluster table PR: https://github.com/gitpod-io/gitpod/pull/8539

svenefftinge:
continues [here](https://github.com/gitpod-io/gitpod/issues/16265)

PR #13348, opened 2022-09-27
[db-sync] Sync `d_b_volume_snapshot` table

andrew-farries:
## Description

Make the `db_volume_snapshot` table able to be synced between clusters.

* Add a migration so that the table has `_lastModified` and `deleted` fields.

Records are currently hard-deleted from this table, so the `deleted` field will not be used in practice:

https://github.com/gitpod-io/gitpod/blob/99e86333318deb096044e0f25d6000293e46f26d/components/gitpod-db/src/typeorm/workspace-db-impl.ts#L776-L779

## Related Issue(s)
Part of https://github.com/gitpod-io/gitpod/issues/9198

## Release Notes
```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-db/src/typeorm/entity/db-volume-snapshot.ts
+++ b/components/gitpod-db/src/typeorm/entity/db-volume-snapshot.ts
@@ -13,6 +13,7 @@ import { Transformer } from "../transformer";
 @Entity()
 @Index("ind_dbsync", ["creationTime"]) // DBSync
 @Index("ind_workspaceId_creationTime", ["workspaceId", "creationTime"]) // findVolumeSnapshotForGCByWorkspaceId
+// on DB but not Typeorm: @Index("ind_lastModified", ["_lastModified"])   // DBSync
 export class DBVolumeSnapshot implements VolumeSnapshot {
     @PrimaryColumn(TypeORM.WORKSPACE_ID_COLUMN_TYPE)
     id: string;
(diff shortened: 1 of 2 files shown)


components/gitpod-db/src/tables.ts
```diff
@@ -47,6 +47,12 @@ export class GitpodSessionTableDescriptionProvider implements TableDescriptionPr
 export class GitpodTableDescriptionProvider implements TableDescriptionProvider {
     readonly name = "gitpod";
     protected readonly tables: TableDescription[] = [
+        {
```
easyCZ, line 50: Feels a bit risky to include this in the same PR as the migration. Are you happy with the risks involved in case the migration doesn't go through but db-sync gets rolled out?

easyCZ (approved):
LGTM

/hold for Q on risk

geropl:
@andrew-farries Did you engage with @sagor999 before doing this? We definitely need to do this before deploying this.

/hold

andrew-farries:
> @andrew-farries Did you engage with @sagor999 before doing this? 

I've requested a review from @sagor999.

components/gitpod-db/src/tables.ts
andrew-farries: Good point - I'll leave the migration + config change in this PR for ease of review but I can split it into two before merging.

sagor999 (approved):
LGTM
(I was under impression that this table already should have been db-synced but I guess it was not? So good catch! :pray: )

components/gitpod-db/src/tables.ts
andrew-farries: I've removed the `db-sync` config change from this PR and I'll add it in a follow up.

PR #13350, opened 2022-09-27
[bridge] Marks stuck stopping and pending instances as stopped

### LINKED ISSUE #11397: workspaces which cannot be scheduled to a cluster get stuck in Pending

laushinka:
## Description
After validating through logging of the past two weeks (since 15.09.2022), this change marks `stopping` and `pending` states after the timeout duration to be stopped. 
Findings from the [logging](https://github.com/gitpod-io/gitpod/pull/12902):
1. Most of the logs were for `stopping` state, and we could not find a clear explanation whether they were cases that we want to be stopped. Instances could be in a `stopping` state for longer than we expect (in the previous PR it was 10 seconds after `stoppingTime`) because instances with a big back-up size can take longer.
2. A lot of the `pending` states were stuck in that state for days, and we do want them stopped.
3. We found a few `creating` states that ended up running, and these we should not stop. 

We are still unsure why these instances end up in these states. Therefore we:
1. Will focus on only stopping stuck instances in `pending` and `stopping` states, and after the timeout duration. These we know for sure we want stopped.
2. To understand the why, we will create an issue. 

## Related Issue(s)

Fixes #11397

## How to test
1. `kubectl cordon` the preview node
2. Run a workspace. It should be stuck in pending.
3. Change the creationTime in `d_b_workspace_instance` to 1 hour before (or wait 1 hour). It should be marked as stopped with a message `"Stopped by ws-manager-bridge. Previously in phase pending`.
4. Once done, `kubectl uncordon` the node, otherwise builds will fail because the node can't be scheduled.

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/ws-manager-bridge/src/bridge.ts
+++ b/components/ws-manager-bridge/src/bridge.ts
@@ -452,7 +452,13 @@ export class WorkspaceManagerBridge implements Disposable {
 
                     // Control running workspace instances against ws-manager
                     try {
-                        await this.controlNonStoppedWSManagerManagedInstances(ctx, nonStoppedInstances, clientProvider);
+                        await this.controlNonStoppedWSManagerManagedInstances(
+                            ctx,
+                            nonStoppedInstances,
+                            clientProvider,
+                            this.config.timeouts.pendingPhaseSeconds,
+                            this.config.timeouts.stoppingPhaseSeconds,
+                        );
 
                         disconnectStarted = Number.MAX_SAFE_INTEGER; // Reset disconnect period
                     } catch (err) {
@@ -489,6 +495,8 @@ export class WorkspaceManagerBridge implements Disposable {
         parentCtx: TraceContext,
         runningInstances: RunningWorkspaceInfo[],
         clientProvider: ClientProvider,
+        pendingPhaseSeconds: number,
+        stoppingPhaseSeconds: number,
     ) {
         const installation = this.config.installation;
 
@@ -508,36 +516,33 @@ export class WorkspaceManagerBridge implements Disposable {
             for (const [instanceId, ri] of runningInstancesIdx.entries()) {
                 const instance = ri.latestInstance;
                 const phase = instance.status.phase;
-                if (phase !== "running") {
-                    // This below if block is to validate the planned fix
-                    if (
-                        phase === "pending" ||
-                        phase === "creating" ||
-                        phase === "initializing" ||
-                        (phase === "stopping" &&
-                            instance.stoppingTime &&
-                            durationLongerThanSeconds(Date.parse(instance.stoppingTime), 10))
-                    ) {
-                        log.info(
-                            { instanceId, workspaceId: instance.workspaceId },
-                            "Logging to validate #12902. Should mark as stopped in database.",
-                            { installation },
-                            { phase },
-                        );
-                    }
-                    log.debug({ instanceId }, "Skipping instance", {
-                        phase: instance.status.phase,
-                        creationTime: instance.creationTime,
-                        region: instance.region,
-                    });
+
+                // When ws-manager is not aware of the following instances outside of the timeout duration,
+                // they should be marked as stopped.
+                // pending states timeout is 1 hour after creationTime.
+                // stopping states timeout is 1 hour after stoppingTime.
+                if (
+                    phase === "running" ||
+                    (phase === "pending" &&
+                        durationLongerThanSeconds(Date.parse(instance.creationTime), pendingPhaseSeconds)) ||
+                    (phase === "stopping" &&
+                        instance.stoppingTime &&
+                        durationLongerThanSeconds(Date.parse(instance.stoppingTime), stoppingPhaseSeconds))
+                ) {
+                    log.info(
+                        { instanceId, workspaceId: instance.workspaceId },
+                        "Database says the instance is present, but ws-man does not know about it. Marking as stopped in database.",
+                        { installation, phase },
+                    );
+                    await this.markWorkspaceInstanceAsStopped(ctx, ri, new Date());
                     continue;
                 }
 
-                log.info(
-                    "Database says the instance is running, but wsman does not know about it. Marking as stopped in database.",
-                    { instanceId, workspaceId: instance.workspaceId, installation, phase },
-                );
-                await this.markWorkspaceInstanceAsStopped(ctx, ri, new Date());
+                log.debug({ instanceId }, "Skipping instance", {
+                    phase: phase,
+                    creationTime: instance.creationTime,
+                    region: instance.region,
+                });
             }
 
             log.debug("Done controlling running instances.", { installation });
--- a/components/ws-manager-bridge/src/config.ts
+++ b/components/ws-manager-bridge/src/config.ts
@@ -31,6 +31,8 @@ export interface Configuration {
         preparingPhaseSeconds: number;
         buildingPhaseSeconds: number;
         unknownPhaseSeconds: number;
+        pendingPhaseSeconds: number;
+        stoppingPhaseSeconds: number;
     };
 
     // emulatePreparingIntervalSeconds configures how often we check for Workspaces in phase "preparing" for clusters we do not govern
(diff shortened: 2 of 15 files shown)


roboquat:

<details>

</details>

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -508,36 +514,33 @@ export class WorkspaceManagerBridge implements Disposable {
...
-                        region: instance.region,
-                    });
+
+                // When ws-manager is not aware of the following instances outside of the timeout duration,
+                // they should be marked as stopped.
+                // pending states timeout is 1 hour after creationTime.
+                // stopping states timeout is 1 hour after stoppingTime.
+                if (
+                    (phase === "pending" &&
+                        durationLongerThanSeconds(Date.parse(instance.creationTime), maxInstanceDuration)) ||
```
geropl, line 524: Would be great to have separate timeouts per phase. This is something that does change, even if it's very rare.

laushinka: Aha, sure. So for now they would be something like `maxPendingInstanceDuration` and `maxStoppingInstanceDuration`? And both would be 3600 seconds for now.

geropl: Why not continue the pattern [here](https://github.com/gitpod-io/gitpod/pull/13350/files#diff-679f493f2181427cc1a110c353e46be2ac044f9f0648b7b89a84b0d0958101d6R31-R33): `pendingPhaseSeconds` and `stoppingPhaseSeconds`? Not caring so much about the particular names, more about consistency. :lotus_position:

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -508,36 +514,33 @@ export class WorkspaceManagerBridge implements Disposable {
...
-                        creationTime: instance.creationTime,
-                        region: instance.region,
-                    });
+
+                // When ws-manager is not aware of the following instances outside of the timeout duration,
+                // they should be marked as stopped.
+                // pending states timeout is 1 hour after creationTime.
+                // stopping states timeout is 1 hour after stoppingTime.
+                if (
+                    (phase === "pending" &&
```
geropl, line 523: The `running` case is missing, right? Or am I misinterpreting the diff/indentation? :see_no_evil:

laushinka: I guess I misinterpreted our sync yesterday 🤔 I thought we would only focus on `pending` and `stopping`, also because we didn't find any of the other states (other than `creating`) in the logs. Do you think the other states are also still relevant?

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -508,36 +516,33 @@ export class WorkspaceManagerBridge implements Disposable {
...
+                        { instanceId, workspaceId: instance.workspaceId },
+                        "Database says the instance is running, but ws-man does not know about it. Marking as stopped in database.",
+                        { installation },
+                        { phase },
+                    );
+                    await this.markWorkspaceInstanceAsStopped(ctx, ri, new Date());
                     continue;
                 }
 
-                log.info(
```
geropl, line 536: @laushinka Sorry for the long turnaround. I like the new layout of the loop! But we need it to include this case: If an instance in `runningInstancesIdx` is `running`, we still need to stop it unconditionally, as we do it at the moment.

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -508,36 +516,33 @@ export class WorkspaceManagerBridge implements Disposable {
...
+                if (
+                    (phase === "pending" &&
+                        durationLongerThanSeconds(Date.parse(instance.creationTime), pendingPhaseSeconds)) ||
+                    (phase === "stopping" &&
+                        instance.stoppingTime &&
+                        durationLongerThanSeconds(Date.parse(instance.stoppingTime), stoppingPhaseSeconds))
+                ) {
+                    log.info(
+                        { instanceId, workspaceId: instance.workspaceId },
+                        "Database says the instance is running, but ws-man does not know about it. Marking as stopped in database.",
```
geropl, line 533: `is running` is potentially misleading here. What do you think about `is present`?

components/ws-manager-bridge/src/bridge.ts
```diff
@@ -508,36 +516,33 @@ export class WorkspaceManagerBridge implements Disposable {
...
+                        { installation },
```
geropl, line 534:
```suggestion
                        { installation, phase },
```

Next line can be dropped, then.

components/ws-manager-bridge/src/config.ts
```diff
@@ -31,6 +31,8 @@ export interface Configuration {
         preparingPhaseSeconds: number;
         buildingPhaseSeconds: number;
         unknownPhaseSeconds: number;
+        pendingPhaseSeconds: number;
```
geropl, line 34: :+1:

components/ws-manager-bridge/src/bridge.ts
laushinka: Ah yes, good point. Feels like there could be a better one than "is present", but I can't come up with a better one and it's definitely better than "is running". Let's go with `is present`!

geropl:
Code LGTM, thx @laushinka ! :pray: 
Will test now...

laushinka:
> Code LGTM, thx @laushinka ! 🙏 Will test now...

@geropl Cool! Let me know if the steps I described were helpful, or if you test in a different way 🙏🏽

geropl (approved):
Code LGTM, tested and works as expected! :1st_place_medal:

Issue #13198, opened 2022-09-22
[usage] Move Stripe handling from server to usage


easyCZ:
## Is your feature request related to a problem? Please describe

We've accumulated a decent amount of Stripe logic in the `server` component, but we also have stripe interaction on the `usage` component. Ideally, we want to co-locate all Stripe interaction on a single component to reduce our mental overhead when dealing with Stripe problems.

We would introduce a StripeService on the usage component which would be responsible for handling our domain specific stripe interactions.

Example PRs which add Stripe interaction into server:
* https://github.com/gitpod-io/gitpod/pull/13002

- [x] [Proto changes](https://github.com/gitpod-io/gitpod/pull/13951)
- [ ] https://github.com/gitpod-io/gitpod/pull/14044
- [ ] Feature flag

laushinka:
CreateStripeSubscription RPC is working for a feature-flagged user on staging. Will now switch it on for all on staging.

PR #13351, opened 2022-09-27
[server] Refactor StripeService.getPortalUrl to take an attributionId instead of a userId or teamId

jankeromnes:
## Description

Refactor `StripeService.getPortalUrl` to take an `attributionId` instead of a `userId` or `teamId`.

Note: I know that we may want to eventually move `StripeService` over to the `usage` component (or, maybe better, to a dedicated `billing` component) as per https://github.com/gitpod-io/gitpod/issues/13198. However, this at least gives a cleaner API that could be migrated as is (i.e. no user- or team-specific methods, only attributionIds).

## Related Issue(s)

Fixes #

## How to test

1. Subscribe to Stripe
2. Opening the Stripe customer portal should still work

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/server/ee/src/user/stripe-service.ts
+++ b/components/server/ee/src/user/stripe-service.ts
@@ -6,9 +6,7 @@
 
 import { inject, injectable } from "inversify";
 import Stripe from "stripe";
-import { Team, User } from "@gitpod/gitpod-protocol";
 import { Config } from "../../../src/config";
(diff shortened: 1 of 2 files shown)


jankeromnes:
Tested, works 🎉

easyCZ (approved):
nice clean up

PR #13352, opened 2022-09-27
[ws-manager] config terminationGracePeriodSeconds

svenefftinge:
## Description
Allow extending the termination grace period of workspace pods through gitpod.yml configuration.

## Related Issue(s)

Fixes #

## How to test

- Start a workspace on this branch : https://github.com/svenefftinge/termination-test/tree/se-extended-race
- run `ignoreSigterm > out.txt` 
- stop workspace. It should stay in stopping close to 100 seconds.
- when starting again, the `out.txt` should contain log messages that show the process ran almost 100 seconds after receiving the SIGHUP

## Release Notes

```release-note
Allow configuration of 'terminationGracePeriodSeconds' to get more time for gracefully stopping services in workspaces
```

## Werft options:

- [x] /werft with-preview


roboquat:

<details>

</details>

components/gitpod-protocol/data/gitpod-schema.json
```diff
@@ -114,6 +114,10 @@
                 "additionalProperties": false
             }
         },
+        "terminationGracePeriodSeconds": {
```
easyCZ, line 117: In the protocol it's marked as optional, does it need to be marked optional here too?

easyCZ: By making this change, all requests will now be able to set this despite it not being actually implemented by workspaces. Is that intended? Do we want to control release of this in some way?

components/ws-manager-api/core.proto
```diff
@@ -567,6 +567,9 @@ message StartWorkspaceSpec {
 
     // ssh_public_keys is user's uploaded ssh public keys
     repeated string ssh_public_keys = 15;
+
+    // termination_grace_period_seconds sets the TerminationGracePeriodSeconds, defaults to 30 seconds
+    int64 termination_grace_period_seconds = 16;
```
easyCZ, line 572: Worth keeping in mind that `ws-manager` running in WS clusters will have a different rollout cadence to webapp. This change is backwards compatible so it shouldn't be a system problem, just an expectation.

easyCZ (approved):
What's the general release plan for this feature? Are we looking to control access to the termination grace period somehow?

/hold for Qs

components/ws-manager/pkg/manager/create.go
```diff
@@ -501,6 +501,11 @@ func (m *Manager) createDefiniteWorkspacePod(startContext *startWorkspaceContext
 		},
 	}
 
+	var terminationGracePeriodSeconds *int64
+	if req.Spec.TerminationGracePeriodSeconds != 0 {
```
aledbf, line 505: No upper limit? Even for free accounts?

svenefftinge: 👍 It makes sense to set an upper limit. Any suggestion for what is reasonable?
I wouldn't want to bind it to the account type, though.
We might want to reconsider measurement of credits , though, which is currently based on `stoppingTime`. cc @easyCZ

svenefftinge:
btw. the bash process we are starting doesn't seem to propagate SIGINT to the process it is waiting for.

components/gitpod-protocol/data/gitpod-schema.json
svenefftinge: > In the protocol it's marked as optional, does it need to be marked optional here too?

it is optional by default

easyCZ: Ah, it used to have `optional` in the v2 version of the spec. in `v4` it only has `required`. Thanks!

components/ws-manager/pkg/manager/create.go
csweichel: In this context the upper bound would only need to be sensible from an operations perspective, e.g. to not trip adverse behaviour or alerts.

For now, we could hardcode it, and if future need arrises plug it through config.

sagor999: Let's add 5 min limit? Feels like it might be a good starting point?

components/gitpod-protocol/data/gitpod-schema.json
```diff
@@ -114,6 +114,10 @@
...
+            "type": "number",
+            "description": "The grace period for all processes to stop before beeing force killed. Default is 15 seconds."
```
sagor999, line 119: Default is 30 seconds.

sagor999:
@csweichel 
https://github.com/kubernetes/apiserver/blob/710ad8da8e0ae8c5dcb4a12744d231c26c7e9905/pkg/registry/rest/delete.go#L123-L128
According to this only shorter grace period can be provided into delete. Which means our stopWorkspace call (and any other place where we do this) can effectively be a maximum grace period? :thinking: 
though it changes current behaviour quite a bit.

csweichel:
> @csweichel https://github.com/kubernetes/apiserver/blob/710ad8da8e0ae8c5dcb4a12744d231c26c7e9905/pkg/registry/rest/delete.go#L123-L128 According to this only shorter grace period can be provided into delete. Which means our stopWorkspace call (and any other place where we do this) can effectively be a maximum grace period? 🤔 though it changes current behaviour quite a bit.

Which indeed means that the value we use for `stopWorkspace` would take precedence, because it's either the default 30s or less than that.

We could either (in descending order of implementation complexity)
1. use the actual termination grace period of the pod,
2. use the configurable max termination grace period,
3. just use a very long constant time (`999 * time.Hour`).

svenefftinge:
I chose 3. ([see](https://github.com/gitpod-io/gitpod/pull/13352/files#diff-1036045c7294f75c4274cbea5ac095c4ac9c91b7a2994779fc262f9d46c06b9cR121))

components/ws-manager/pkg/manager/create.go
```diff
@@ -501,6 +501,15 @@ func (m *Manager) createDefiniteWorkspacePod(startContext *startWorkspaceContext
...
+		terminationGracePeriodSeconds = &req.Spec.TerminationGracePeriodSeconds
+		var maxGracePeriodSeconds int64 = 60 * 5 // 5 minutes
```
sagor999, line 507: can we use `stopWorkspaceNormallyGracePeriod` const from manager.go here? and add a comment in manager.go for this const that it is used as a maximum for user supplied grace period.

sagor999 (approved):

sagor999:
/hold
for conflicts and failed build

PR #13354, opened 2022-09-27
Use different Stripe Price IDs when subscribing individuals vs teams

jankeromnes:
## Description

Use different Stripe Price IDs when subscribing individuals vs teams.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/13302

## How to test

1. Subscribe to usage-based as a team
2. In the Stripe admin dashboard, your team's subscription should be linked to a "Team" price (pay-as-you-go)
3. Subscribe to usage-based as an individual
4. In the Stripe admin dashboard, your user's subscription should be linked to an "Individual" price (discounted flat rate + pay-as-you-go)

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/server/ee/src/user/stripe-service.ts
+++ b/components/server/ee/src/user/stripe-service.ts
@@ -8,6 +8,7 @@ import { inject, injectable } from "inversify";
 import Stripe from "stripe";
 import { Config } from "../../../src/config";
 import { log } from "@gitpod/gitpod-protocol/lib/util/logging";
+import { AttributionId } from "@gitpod/gitpod-protocol/lib/attribution";
 
 const POLL_CREATED_CUSTOMER_INTERVAL_MS = 1000;
 const POLL_CREATED_CUSTOMER_MAX_ATTEMPTS = 30;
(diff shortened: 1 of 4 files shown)


components/server/ee/src/user/stripe-service.ts
```diff
@@ -115,15 +116,25 @@ export class StripeService {
...
             throw new Error(`Stripe customer '${customerId}' could not be found`);
         }
+        const attrId = AttributionId.parse(attributionId);
+        if (!attrId) {
+            throw new Error(`Invalid attributionId '${attributionId}'`);
+        }
         const currency = customer.metadata.preferredCurrency || "USD";
-        const priceId = this.config?.stripeConfig?.usageProductPriceIds[currency];
+        const priceIds =
+            attrId.kind === "team"
```
easyCZ, line 130: Could we use a switch or if-if else - else here to catch the case when a different `kind` is added such that it doensn't silently fall into the `else` branch?

easyCZ (approved):
LGTM

/hold for explicitly handling kinds

components/server/ee/src/user/stripe-service.ts
jankeromnes: Good point! I agree that such silent fallbacks feel dangerous, so I really appreciate this nudge. 👍 Will fix.

jankeromnes:
Nits addressed, still works! 🎉

Many thanks again for the review 🙏

/unhold

PR #13356, opened 2022-09-27
[refarch] Update node size for GCP and AWS

Furisto:
## Description
Update the node size for GCP and AWS self hosted reference architectures as laid out in https://github.com/gitpod-io/website/pull/2790.

## Related Issue(s)
Related to https://github.com/gitpod-io/website/pull/2790

## How to test
Not sure, would appreciate some pointers.

## Release Notes

```release-note
Update node sizes for self hosted reference architectures
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/infra/single-cluster/aws/cluster.tf
+++ b/install/infra/single-cluster/aws/cluster.tf
@@ -11,7 +11,7 @@ module "eks" {
   create_external_database = var.create_external_database
   create_external_storage  = var.create_external_storage
   service_machine_type     = "m6i.xlarge"
-  workspace_machine_type   = "m6i.2xlarge"
+  workspace_machine_type   = "m6i.4xlarge"
 
   create_external_storage_for_registry_backend = var.create_external_storage_for_registry_backend
 }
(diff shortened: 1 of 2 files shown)


nandajavarma:
Hey @Furisto ! Thanks for this fix! I believe the following suggestion is an indicator that we should name the terraform modules better. The terraform setup corresponding to the Single cluster reference architecture is present [here](https://github.com/gitpod-io/gitpod/tree/main/install/infra/single-cluster/aws). In [cluster.tf](https://github.com/gitpod-io/gitpod/blob/main/install/infra/single-cluster/aws/cluster.tf#L13), you can see that the machine size is specified. We will have to update this value.

The `eks` module in the `cluster.tf` file essentially use the same module that you edited, but the variables are overridden in single-cluster ref arch and in the future from other ref arch modules.

Furisto:
@nandajavarma Thanks. Added the override for gke and eks.

nandajavarma (approved):
:shipit: 

/unhold

PR #13367, opened 2022-09-27
Add default value for SEGMENT_IO_TOKEN

mads-hartmann:
## Description

Specify a default value for `SEGMENT_IO_TOKEN` so that we can use `leeway build` in a Gitpod Workspace without specifying `-DSEGMENT_IO_TOKEN=""`

@gitpod-io/engineering-self-hosted Is is okay for this value to be empty? We always specify a value when building in CI so this is mainly if you run `leeway build` in a workspace to build the local-preview ☺️

## Related Issue(s)

Part of https://github.com/gitpod-io/ops/issues/5411

## How to test

```sh
leeway build install/preview:docker -Dversion="mads" -DwithLocalPreview=true
```

Should no longer error due to missing `SEGMENT_IO_TOKEN` build argument (`-DwithLocalPreview=true` is going away soon).

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/WORKSPACE.yaml
+++ b/WORKSPACE.yaml
@@ -15,8 +15,11 @@ defaultArgs:
   golandDownloadUrl: "https://download.jetbrains.com/go/goland-2022.2.3.tar.gz"
   pycharmDownloadUrl: "https://download.jetbrains.com/python/pycharm-professional-2022.2.2.tar.gz"
   phpstormDownloadUrl: "https://download.jetbrains.com/webide/PhpStorm-2022.2.2.tar.gz"
+  # Used by the kots installer
   REPLICATED_API_TOKEN: ""
   REPLICATED_APP: ""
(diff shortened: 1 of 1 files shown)


MrSimonEmms:
@mads-hartmann yes, it looks like it will be fine. The [command](https://github.com/gitpod-io/gitpod/blob/main/components/installation-telemetry/cmd/send.go#L40) will error, but it's only triggered as a [nightly cronjob](https://github.com/gitpod-io/gitpod/blob/main/install/installer/pkg/components/gitpod/cronjob.go) which shouldn't be happening in preview environments anyway I guess.

We need to ascertain that the cronjob fails nicely and doesn't spin out of control (looking at the spec, it doesn't appear to be the case but I wouldn't want to say for definite until we've tested it)

PR #13386, opened 2022-09-28
[ws-daemon] fix WaitForInit not found error in tracing

### LINKED ISSUE #11713: [ws-daemon] cannot find workspace during WaitForInit

sagor999:
## Description

Provide new gRPC method for ws-daemon to check if workspace exists prior to calling WaitForInit.
That is to clean up our traces from erroneous errors that are reported by WaitForInit call when workspace does not exists.

## Related Issue(s)

Fixes #11713

## How to test

Start and stop workspace

## Release Notes

```release-note
none
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-integration-tests=workspace


components/ws-daemon/pkg/content/service.go
```diff
@@ -743,15 +743,7 @@ func (s *WorkspaceService) WaitForInit(ctx context.Context, req *api.WaitForInit
...
-	defer func() {
-		// https://github.com/gitpod-io/gitpod/issues/11713
-		if err != nil && strings.Contains(err.Error(), "cannot find workspace") {
-			log.WithFields(log.OWI("", "", req.Id)).Warn(err)
-			tracing.FinishSpan(span, nil)
-		} else {
-			tracing.FinishSpan(span, &err)
-		}
-	}()
+	tracing.FinishSpan(span, &err)
```
WVerlaek, line 746: 👋 was scanning some PRs to build up context and noticed this, should `FinishSpan` stay `defer`ed? Same below

sagor999: Oops. You are :100: right. Good catch!

kylos101:
@sagor999 I believe gRPC methods for ws-daemon and ws-manager may be going away as part of the mkII effort. [[1](https://www.notion.so/gitpod/Pay-down-technical-debt-with-ws-manager-mk2-5f3942d7882849e9b8372eb8530f7b1c#0995e8880b95408b84faf169cc69a426)][[2](https://www.notion.so/gitpod/Milestones-ws-manager-mk2-fb471d7e01e4494b9bca9ad1cd50b3b2#905a0e63c72d4726b833628f8af4830b)]

If this cleans up traces only (so like, does not resolve a customer facing issue), perhaps it would make sense to wait to add this behavior until you're doing the CRD work mkII? In other words, we might be able to live with this error in traces for a few more months. I'm honestly not sure, might be good to socialize with @aledbf and @csweichel .

sagor999:
@kylos101 this is an easy fix to make tracing cleaner. Mk2 might take longer to roll out, and then we will constantly be seeing those errors from Mk1 as well, making it more difficult (hence error prone) to spot actual problems.
So I think this should go in without regard to Mk2 effort.

jenting:
I think you tried to fix the issue https://github.com/gitpod-io/gitpod/issues/11713#issuecomment-1244523673, correct?

sagor999:
@jenting yeah, it is related to that issue. only that not found is not really an error, but grpc logs it as such, which makes traces noisy when scanning for errors.

jenting:
> @jenting yeah, it is related to that issue. only that not found is not really an error, but grpc logs it as such, which makes traces noisy when scanning for errors.

👍 I'd prefer to reopen issue #11713 and link this PR to fix issue #11713.

sagor999:
@jenting done!

jenting (approved):
Code LGTM

kylos101:
@aledbf and @Furisto your review is needed as code owners, [this](https://www.notion.so/gitpod/Efficient-PR-reviews-14e2b608d4d74eeb9100580c66a99035#41cf97a0d4ba48d9b8670ef8ffeb51e5) will help you filter just to those files. :pray:

kylos101:
> @kylos101 this is an easy fix to make tracing cleaner. Mk2 might take longer to roll out, and then we will constantly be seeing those errors from Mk1 as well, making it more difficult (hence error prone) to spot actual problems. So I think this should go in without regard to Mk2 effort.

Perfect, thanks for the explanation, @sagor999 ! :smile_cat:

Furisto (approved):

PR #13387, opened 2022-09-28
[ws-manager] refactor ide image layer and envvar

iQQBot:
## Description

We want to decouple the IDE component from the `ws-manager`

In the past, if we needed to add an environment variable starting with `GITPOD`, we needed to modify the whitelist of `ws-manager` first, and then wait for the workspace team to deployed the new generation cluster before we could modify the server, in this PR, the `system variables` are distinguished from the `user-defined environment variables`, the `system variables` are not filtered, and No need to write secret

Another case is if we want to add another layer of images to the IDE, the situation is the same as above, we have to modify the `registry-facade` and `ws-manager` and wait for the workspace team to deploy it, in this PR, we add a new ide_layer_ref which is contain some ide image layer, later if we want to add layers later, we can dynamically add

## Related Issue(s)

Relate #9189

## How to test

1. open a workspace using desktop-ide in preview environment, it should work as normal
2. replace `server` image to `eu.gcr.io/gitpod-core-dev/build/server:main.4899` and test step 1, it should work too

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


components/registry-facade-api/imagespec.proto
```diff
@@ -8,16 +8,18 @@ option go_package = "github.com/gitpod-io/gitpod/registry-facade/api";
...
-    // desktop_ide_ref points to an image denotign the desktop IDE to use
+    // DEPRECATED desktop_ide_ref points to an image denotign the desktop IDE to use
     string desktop_ide_ref = 4;
     // supervisor_ref points to an image denotign the supervisor to use
     string supervisor_ref = 5;
-    // desktop_ide_plugin_ref points to an image denotign the desktop IDE plugin to use
+    // DEPRECATED desktop_ide_plugin_ref points to an image denotign the desktop IDE plugin to use
     string desktop_ide_plugin_ref = 6;
+    // dynamic_layer_ref contains all layers that ide need
+    repeated string dynamic_layer_ref = 7;
```
csweichel, line 22: This name is too generic. Please name something akin to `ide_layer_ref`.
Also in subsequent code in registry-facade.

components/registry-facade-api/imagespec.proto
```diff
@@ -8,16 +8,18 @@ option go_package = "github.com/gitpod-io/gitpod/registry-facade/api";
 message ImageSpec {
     // base_ref points to an image in another registry
     string base_ref = 1;
-    // ide_ref points to an image denoting the IDE to use
+    // DEPRECATED ide_ref points to an image denoting the IDE to use
     string ide_ref = 2;
```
csweichel, line 12: With the deprecation of `ide_ref`, how does blobserve decide from which image it should serve files from?

iQQBot: I remember I deleted it..

components/registry-facade-api/imagespec.proto
iQQBot: changed

components/registry-facade/pkg/registry/layersource.go
```diff
@@ -368,6 +368,9 @@ func (cs CompositeLayerSource) GetBlob(ctx context.Context, spec *api.ImageSpec,
 // RefSource extracts an image reference from an image spec
 type RefSource func(*api.ImageSpec) (ref string, err error)
 
+// ResfSource extracts an image list reference from an image spec
+type RefsSource func(*api.ImageSpec) (refs []string, err error)
```
csweichel, line 372: Better refactor `RefSource` to return `[]string` instead of introducing another one on the side

components/registry-facade-api/imagespec.proto
```diff
@@ -12,12 +12,14 @@ message ImageSpec {
...
     // content_layer describe the last few layers which provide the workspace's content
     repeated ContentLayer content_layer = 3;
-    // desktop_ide_ref points to an image denotign the desktop IDE to use
+    // DEPRECATED desktop_ide_ref points to an image denotign the desktop IDE to use
     string desktop_ide_ref = 4;
```
csweichel, line 16: Deprecating fields in `proto` is best done using `reserved` if we control both sides of the API (consumer and provider). Otherwise using the [deprecated field option](https://developers.google.com/protocol-buffers/docs/proto3) would be preferable.

```suggestion
    // Was used for desktop_ide_ref points to an image denotign the desktop IDE to use
    reserved 4;
```

components/registry-facade/pkg/registry/layersource.go
```diff
@@ -459,6 +462,130 @@ func (src *SpecMappedImagedSource) getDelegate(ctx context.Context, spec *api.Im
 	return lsrc, nil
 }
 
+// NewDynamicSpecMappedImageSource creates a new dynamic spec mapped image source
+func NewDynamicSpecMappedImageSource(resolver ResolverProvider, refsSource RefsSource) (*SpecDynamicMappedImagedSource, error) {
```
csweichel, line 466: How does the `DynamicSpecMappedImageSource` differ from the `SpecMappedImageSource`?

components/registry-facade-api/imagespec.proto
iQQBot: This would be safe if ws-manager didn't perform hot-patch, and I could remove unnecessary compatibility code

components/registry-facade/pkg/registry/layersource.go
iQQBot: `DynamicSpecMappedImageSource ` just like `CompositeLayerSource` + `SpecMappedImageSource` but the difference is that the exact number of layers is determined when the spec comes through

iQQBot: The main purpose is to support multiple image layers, i.e. https://github.com/gitpod-io/gitpod/pull/13387/files#diff-18042f8c1ba07aab6f1481f62e093bdadeb2042f52cbdef5a559605ec79f76efR372

easyCZ (approved):
LGTM WebApp side

/hold for other reviewers

aledbf (approved):

components/server/src/workspace/workspace-starter.ts
```diff
@@ -1569,16 +1571,20 @@ export class WorkspaceStarter {
...
         const startWorkspaceSpecIDEImage = new IDEImage();
         startWorkspaceSpecIDEImage.setWebRef(ideImage);
-        startWorkspaceSpecIDEImage.setDesktopRef(instance.configuration?.desktopIdeImage || "");
-        startWorkspaceSpecIDEImage.setDesktopPluginRef(instance.configuration?.desktopIdePluginImage || "");
-        startWorkspaceSpecIDEImage.setSupervisorRef(instance.configuration?.supervisorImage || "");
+        startWorkspaceSpecIDEImage.setSupervisorRef(instance.configuration?.supervisorImage ?? "");
         spec.setIdeImage(startWorkspaceSpecIDEImage);
-        spec.setDeprecatedIdeImage(ideImage);
+
+        instance.configuration?.desktopIdeImage && spec.addIdeImageLayers(instance.configuration?.desktopIdeImage);
```
geropl, line 1584: nit: We don't use `&&` chaining where an `if` would work as well.

iQQBot: NVM, this is a test commit, will delete it

csweichel (approved):

PR #13394, opened 2022-09-28
[installer-tests] Cleanup the self-hosted test crons

nandajavarma:
## Description

This PR fixes a couple of things that are breaking with the automated tests:
- I recently changed `skipTests` to `runTests` in the intention that we can make skipping tests the default behavior. But the thing I didn't consider is, this would mean that the crons running every night doesn't run the tests. So rolled this change back to set running tests the default behavior
- In the cleanup script, currently for cleaning up certificates, it checks if variable `self_signed` is set. But during cleanup crons, no such variables are being set but we have to cleanup anyways. Hence the variable check is removed
- Currently, in `get-kubeconfig` targets (the ones used to retrieve kubeconfig file either locally or during werft jobs) I am using a secret being set to check if it is a cron or local. This is a wrong behavior since that secret is not needed in EKS or AKS. I remove this dependency and set a custom env var in case of cron jobs.
- Fixing the preview env variable name
- Backing up the certificate secret in case of re-runs

## Related Issue(s)

Fixes #

## How to test

You can run the commands from [the internal doc](https://www.notion.so/gitpod/Self-hosted-automated-preview-and-test-pipelines-fdc5a202e67b4197aa00d93b98d57927)

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/.werft/jobs/build/self-hosted-upgrade-tests.ts
+++ b/.werft/jobs/build/self-hosted-upgrade-tests.ts
@@ -81,9 +81,9 @@ export async function triggerUpgradeTests(werft: Werft, config: JobConfig, usern
 export async function triggerSelfHostedPreview(werft: Werft, config: JobConfig, username: string) {
     const replicatedChannel =  config.replicatedChannel || config.repository.branch;
     const cluster =  config.cluster || "k3s";
-    const formattedBranch = config.repository.branch.replace("/", "-").slice(0,10)
+    const formattedBranch = config.repository.branch.replace("/", "-").slice(0,11)
     const phase = phases[cluster]
-    const subdomain =  `${formattedBranch}x-${phase.cloud}`
+    const subdomain =  `${formattedBranch}-${phase.cloud}`
 
     const replicatedApp = process.env.REPLICATED_APP
 
--- a/.werft/aks-installer-tests.yaml
+++ b/.werft/aks-installer-tests.yaml
@@ -20,8 +20,8 @@ args:
   desc: "Version of gitpod to install(in the case of upgrade tests, this is the initial install version and will later get upgraded to latest"
   required: false
   default: ""
-- name: runTests
-  desc: "Set this to true to run integration tests"
+- name: skipTests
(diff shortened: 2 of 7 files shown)


nandajavarma:
Passing test from this change https://werft.gitpod-dev.com/job/gitpod-custom-nvn-cleanup-sh-tests.2/logs

nandajavarma:
[Docs](https://www.notion.so/gitpod/Self-hosted-automated-preview-and-test-pipelines-fdc5a202e67b4197aa00d93b98d57927) have been updated to reflect this change

.werft/jobs/build/self-hosted-upgrade-tests.ts
```diff
@@ -81,9 +81,9 @@ export async function triggerUpgradeTests(werft: Werft, config: JobConfig, usern
 export async function triggerSelfHostedPreview(werft: Werft, config: JobConfig, username: string) {
     const replicatedChannel =  config.replicatedChannel || config.repository.branch;
     const cluster =  config.cluster || "k3s";
-    const formattedBranch = config.repository.branch.replace("/", "-").slice(0,10)
+    const formattedBranch = config.repository.branch.replace("/", "-").slice(0,11)
```
MrSimonEmms, line 84: What's the relevance of the `10`/`11`? Is that the length of a word that is being split by or something?

MrSimonEmms (approved):
/hold

Looks ok to me, just the one question. I haven't run it, but I'm happy to trust that you've tested it yourself

.werft/jobs/build/self-hosted-upgrade-tests.ts
nandajavarma: I had a `x` added to the line below because I was trying to test redeploy and I wanted to get a different subdomain. It caused for an [unfortunate subdomain yesterday ](https://werft.gitpod-dev.com/job/gitpod-k3s-installer-tests-head.33)so I am removing that and keeping the subdomain name length the same.

PR #13399, opened 2022-09-28
[supervisor] properly propagate signals to terminal processes

### LINKED ISSUE #13365: [supervisor] Processes started in terminals do not get a SIGTERM signal on stop

akosyakov:
## Related Issue(s)

Fixes #13365

## Release Notes

```release-note
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/supervisor/pkg/terminal/terminal.go
+++ b/components/supervisor/pkg/terminal/terminal.go
@@ -13,6 +13,7 @@ import (
 	"os/exec"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/creack/pty"
@@ -77,20 +78,23 @@ func (m *Mux) Start(cmd *exec.Cmd, options TermOptions) (alias string, err error
 	go func() {
 		term.waitErr = cmd.Wait()
 		close(term.waitDone)
-		_ = m.CloseTerminal(alias, 0*time.Second)
+
+		m.mu.Lock()
+		defer m.mu.Unlock()
+		_ = m.doClose(alias, unix.SIGKILL)
 	}()
 
 	return alias, nil
 }
 
-// Close closes all terminals with closeTerminaldefaultGracePeriod.
+// Close closes all terminals
 func (m *Mux) Close() error {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
 	var err error
 	for k := range m.terms {
-		cerr := m.doClose(k, closeTerminaldefaultGracePeriod)
+		cerr := m.doClose(k, unix.SIGTERM)
 		if cerr != nil {
 			log.WithError(cerr).WithField("alias", k).Warn("cannot properly close terminal")
 			if err != nil {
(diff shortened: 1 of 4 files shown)


roboquat:

<details>

</details>

components/supervisor/pkg/terminal/terminal.go
```diff
@@ -77,20 +78,23 @@ func (m *Mux) Start(cmd *exec.Cmd, options TermOptions) (alias string, err error
...
-// Close closes all terminals with closeTerminaldefaultGracePeriod.
+// Close closes all terminals
 func (m *Mux) Close() error {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
 	var err error
 	for k := range m.terms {
-		cerr := m.doClose(k, closeTerminaldefaultGracePeriod)
+		cerr := m.doClose(k, unix.SIGTERM)
```
svenefftinge, line 97: Would be good to run this in parallel for all terminals. So they all get max time to stop.

components/supervisor/pkg/terminal/terminal.go
```diff
@@ -146,35 +150,49 @@ func (m *Mux) doClose(alias string, gracePeriod time.Duration) error {
...
-func (term *Term) gracefullyShutdownProcess(gracePeriod time.Duration) error {
+func (term *Term) gracefullyShutdownProcess(sig syscall.Signal) (err error) {
 	if term.Command.Process == nil {
-		// process is alrady gone
+		// process is already gone
 		return nil
 	}
-	if gracePeriod == 0 {
+
+	if sig == unix.SIGKILL {
```
svenefftinge, line 159: we should either change the method (remove 'gracefully') name or not allow SIGKILL her but handle it on the call side.

components/supervisor/pkg/terminal/terminal.go
```diff
@@ -146,35 +150,49 @@ func (m *Mux) doClose(alias string, gracePeriod time.Duration) error {
...
-	case <-time.After(gracePeriod):
+	}
+	alive := syscall.Kill(pid, sig) == nil
+	if !alive {
+		return
 	}
 
-	// process did not exit in time. Let's kill.
+	// 2. wait to terminate
+	for i := time.Duration(0); alive && i < term.closeGracePeiod; i += time.Second {
```
svenefftinge, line 185: I think we can get rid of the `closeGracePeriod`, because there is an outer graceperiod that is imposed by kubernetes that will kill supervisor eventually.

components/supervisor/pkg/terminal/terminal.go
```diff
@@ -146,35 +150,49 @@ func (m *Mux) doClose(alias string, gracePeriod time.Duration) error {
...
+		if alive {
+			time.Sleep(time.Second)
+		}
+	}
+	if !alive {
+		return
+	}
+
+	// 3. process did not exit in time. Let's kill.
 	return term.shutdownProcessImmediately()
```
svenefftinge, line 196: getting rid of gracePeriod would also mean that we don't force shutdown the terminals here

components/supervisor/pkg/terminal/terminal.go
svenefftinge: I understood that we should send SIGHUP instead of SIGTERM because that is what bash would forward to their processes.

components/supervisor/pkg/terminal/terminal.go
svenefftinge: thinking about it, I think we should always send SIGHUP followed by SIGINT if the process is still alive. So the method name would be good, but we should not pass the signal?

components/supervisor/pkg/terminal/terminal.go
akosyakov: I also was thinking about it, but was not sure. From other side VS Code does it always like that when killing terminals. I would assume if you run a script which can be in executed in regular VS Code and super terminal then it should respect SIGHUP.

Issue #13392, opened 2022-09-28
Cannot delete teams in self-hosted RC build


jldec:
It appears that there is a dependency on Stripe configuration in team deletion

#### To reproduce
- Log in to a SH test deployment (see [notion](https://www.notion.so/gitpod/Self-Hosted-Release-2022-09-0-0534b6ba181b476cb1826e358a7149f9#60d436b917f3403da062a88301c31837) internal)
- Create a team
- Try to delete the team - nothing happens in the UI

Watching ws response in Network tab of browser dev-tools will show error
```json
{
	"jsonrpc": "2.0",
	"id": 24,
	"method": "deleteTeam",
	"params": [
		"243a1a6b-e50f-4ebe-8b98-9776ecddf1b7",
		"f4ee66e1-8b1f-4934-82fd-d8a362a7a45c"
	]
}
{
	"jsonrpc": "2.0",
	"id": 24,
	"error": {
		"code": -32603,
		"message": "Request deleteTeam failed with message: Stripe is not properly configured"
	}
}
```

PR #13402, opened 2022-09-28
[server] Fix team deletion for non-UBP enabled teams

andrew-farries:
## Description

Team deletion for non-UBP teams is currently broken. This PR fixes the issue by ensuring that we don't check for UBP specific details about the team (such as any currently active Stripe subscription) before ensuring that Stripe payments are enabled.

## Related Issue(s)
Fixes https://github.com/gitpod-io/gitpod/issues/13392

## How to test

Create a team in the preview environment and then delete it.

## Release Notes

```release-note
[server] Fix for the inability to delete teams that were not subscribed to usage based pricing
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/server/ee/src/workspace/gitpod-server-impl.ts
+++ b/components/server/ee/src/workspace/gitpod-server-impl.ts
@@ -1596,11 +1596,13 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
                 { teamId, chargebeeSubscriptionId },
             );
         }
-        const subscriptionId = await this.stripeService.findUncancelledSubscriptionByAttributionId(
-            AttributionId.render({ kind: "team", teamId: teamId }),
-        );
-        if (subscriptionId) {
-            await this.stripeService.cancelSubscription(subscriptionId);
+        if (this.config.enablePayment) {
+            const subscriptionId = await this.stripeService.findUncancelledSubscriptionByAttributionId(
+                AttributionId.render({ kind: "team", teamId: teamId }),
+            );
+            if (subscriptionId) {
+                await this.stripeService.cancelSubscription(subscriptionId);
+            }
         }
     }
 


components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -1596,11 +1596,14 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
                 { teamId, chargebeeSubscriptionId },
             );
         }
-        const subscriptionId = await this.stripeService.findUncancelledSubscriptionByAttributionId(
-            AttributionId.render({ kind: "team", teamId: teamId }),
-        );
-        if (subscriptionId) {
-            await this.stripeService.cancelSubscription(subscriptionId);
+        const billingMode = await this.billingModes.getBillingMode({ kind: "team", teamId: teamId }, now);
+        if (billingMode.mode === "usage-based") {
```
jankeromnes, line 1600: Just out of curiosity: wouldn't it be simpler (and safer) to check for "isPaymentEnabled" instead of invoking the complicated billingMode logic?

andrew-farries: I can't find any reference to `isPaymentEnabled`. Do you have a link to what you mean?

jankeromnes: Sure, sorry for the inaccuracy. It's just a simple config flag, which seems a bit more straightforward than the complicated state-dependent logic of billingModes:

https://github.com/gitpod-io/gitpod/blob/2f0b9462b0451f0974434fa36bd757420ef04e34/components/server/ee/src/user/user-deletion-service.ts#L28

jankeromnes: Alternatively (but it's pretty much the same thing), we could also check whether Stripe is configured, because that's the requirement here:

https://github.com/gitpod-io/gitpod/blob/17d7063b42a419a7e0f14766ad393a2394dbb211/components/server/ee/src/user/stripe-service.ts#L24

I'd personally be happy with any of these two approaches 😊

andrew-farries: Thanks 🙏 . I think you're right - a simple `enablePayment` check suffices here.

PR rebased to do that.

jankeromnes (approved):
Looks good to me, many thanks! 👍

Note: the check could technically also cover the above subscription check, but let's not "fix" that if it ain't broke 😁

PR #13427, opened 2022-09-28
fix network connections alert to fire only for workspace pods

sagor999:
## Description

This alert is specifically made to monitor workspaces that might be abusing conntrack table.
So update it to make sure it fires only for workspace pods, and not for any other pods.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
none
```

## Werft options:

- [ ] /werft with-preview


Furisto:
@sagor999 Hey Pavel, the pod refers to the pod that is exporting the metric which would be the node-exporter, so I do not think that alert will fire as it will find no matching entries. You can check that by using 'Explore' in Grafana and using this query: `node_nf_conntrack_entries{instance=~"workspace-.*", instance!~"serv.*"}`. When you check the values for `pod` in the output you will see that there are no workspace pods, only node exporter.

PR #13432, opened 2022-09-29
[supervisor] on shutdown send SIGINT to direct child processes of terminals

### LINKED ISSUE #13365: [supervisor] Processes started in terminals do not get a SIGTERM signal on stop

svenefftinge:
## Description
We currently send sigint to the terminal process directly, but bash doesn't propagate it to its child processes. 
With this change we first send a SIGINT to the direct children of the main terminal process and waiting for them to exit, before sending SIGINT to the main process itself. 
Also, we are waiting indefinitely for those processes to stop, relying on the graceperiod of Kubernetes.

## Related Issue(s)

Fixes #13365

## How to test
Start a workspace https://se-sighup.preview.gitpod-dev.com/#https://github.com/svenefftinge/termination-test

Run 
```
./ignoresigterm > out_disown.txt &
disown $1
./ignoresigterm > out_bg.txt &
./ignoresigterm > out.txt
```
and stop the workspace.
With this preview environment you should see log output like below in all three files
```
awaiting signal

Received signal
terminated

Ignored for 0 seconds.
Ignored for 1 seconds.
Ignored for 2 seconds.
Ignored for 3 seconds.
Ignored for 4 seconds.
Ignored for 5 seconds.
Ignored for 6 seconds.
Ignored for 7 seconds.
Ignored for 8 seconds.
Ignored for 9 seconds.
Ignored for 10 seconds.
Ignored for 11 seconds.
Ignored for 12 seconds.
Ignored for 13 seconds.
Ignored for 14 seconds.
Ignored for 15 seconds.
...
```

With gitpod.io you will see only (i.e. the processes never received SIGINT):
```
awaiting signal
```

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/supervisor/pkg/supervisor/supervisor.go
+++ b/components/supervisor/pkg/supervisor/supervisor.go
@@ -36,10 +36,8 @@ import (
 	grpcruntime "github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	"github.com/improbable-eng/grpc-web/go/grpcweb"
 	"github.com/prometheus/common/route"
-	"github.com/prometheus/procfs"
 	"github.com/soheilhy/cmux"
 	"golang.org/x/crypto/ssh"
-	"golang.org/x/sys/unix"
 	"golang.org/x/xerrors"
 	"google.golang.org/grpc"
 	"google.golang.org/protobuf/proto"
@@ -50,7 +48,6 @@ import (
 	csapi "github.com/gitpod-io/gitpod/content-service/api"
 	"github.com/gitpod-io/gitpod/content-service/pkg/executor"
 	"github.com/gitpod-io/gitpod/content-service/pkg/git"
-	"github.com/gitpod-io/gitpod/content-service/pkg/initializer"
 	gitpod "github.com/gitpod-io/gitpod/gitpod-protocol"
 	"github.com/gitpod-io/gitpod/supervisor/api"
 	"github.com/gitpod-io/gitpod/supervisor/pkg/activation"
@@ -395,10 +392,7 @@ func Run(options ...RunOption) {
 		log.WithError(err).Error("terminal closure failed")
 	}
 
-	// terminate all child processes once the IDE is gone
 	ideWG.Wait()
-	terminateChildProcesses()
-
 	wg.Wait()
 }
 
@@ -1414,77 +1408,6 @@ func startContentInit(ctx context.Context, cfg *Config, wg *sync.WaitGroup, cst 
 
 	log.WithField("source", src).Info("supervisor: workspace content init finished")
 	cst.MarkContentReady(src)
-}
-
-func terminateChildProcesses() {
-	parent := os.Getpid()
-
-	children, err := processesWithParent(parent)
-	if err != nil {
-		log.WithError(err).WithField("pid", parent).Warn("cannot find children processes")
-		return
-	}
-
-	for pid, uid := range children {
-		privileged := false
-		if initializer.GitpodUID != uid {
-			privileged = true
-		}
-
-		terminateProcess(pid, privileged)
-	}
-}
-
-func terminateProcess(pid int, privileged bool) {
-	var err error
-	if privileged {
-		cmd := exec.Command("kill", "-SIGTERM", fmt.Sprintf("%v", pid))
-		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stderr
-		err = cmd.Run()
-	} else {
-		err = syscall.Kill(pid, unix.SIGTERM)
-	}
-
-	if err != nil {
-		log.WithError(err).WithField("pid", pid).Debug("child process is already terminated")
-		return
-	}
-
-	log.WithField("pid", pid).Debug("SIGTERM'ed child process")
-}
-
-func processesWithParent(ppid int) (map[int]int, error) {
-	procs, err := procfs.AllProcs()
-	if err != nil {
-		return nil, err
-	}
-
-	children := make(map[int]int)
-	for _, proc := range procs {
-		stat, err := proc.Stat()
-		if err != nil {
-			continue
-		}
-
-		if stat.PPID != ppid {
-			continue
-		}
-
-		status, err := proc.NewStatus()
-		if err != nil {
-			continue
-		}
-
-		uid, err := strconv.Atoi(status.UIDs[0])
-		if err != nil {
-			continue
-		}
-
-		children[proc.PID] = uid
-	}
-
-	return children, nil
 }
 
 func socketActivationForDocker(ctx context.Context, wg *sync.WaitGroup, term *terminal.Mux) {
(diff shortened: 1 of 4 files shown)


roboquat:

<details>

</details>

components/supervisor/pkg/terminal/terminal.go
```diff
@@ -146,43 +147,56 @@ func (m *Mux) doClose(alias string, gracePeriod time.Duration) error {
...
-
-	err := term.Command.Process.Signal(unix.SIGTERM)
-	if err != nil {
-		return err
+	pid := term.Command.Process.Pid
+	var subwg sync.WaitGroup
+	processes, err := procfs.AllProcs()
+	if err == nil {
+		// terminate direct child processes
+		for _, p := range processes {
```
akosyakov, line 160: I'm not sure it is a way to go, the problem here is if someone has a process which start another process and they have a contract between each other based on signals. We will break it because we can terminate child process before parent.

The only real direct child process here is bash.

svenefftinge: You mean it is not ok to assume that the main terminal process is a shell?

components/supervisor/pkg/supervisor/supervisor.go
```diff
@@ -390,15 +387,12 @@ func Run(options ...RunOption) {
 
 	log.Info("received SIGTERM (or shutdown) - tearing down")
 	cancel()
-	err = termMux.Close()
+	err = termMux.Close(&wg)
```
csweichel, line 390: What makes it so that this call can no longer be synchronous? Particularly considering that the WaitGroup is awaited on immediately afterwards?

Also, passing `WaitGroups` into cross-package API is quite unusual.

components/supervisor/pkg/terminal/terminal.go
```diff
@@ -146,43 +147,56 @@ func (m *Mux) doClose(alias string, gracePeriod time.Duration) error {
...
-		return term.shutdownProcessImmediately()
-	}
-
-	err := term.Command.Process.Signal(unix.SIGTERM)
-	if err != nil {
-		return err
+	pid := term.Command.Process.Pid
+	var subwg sync.WaitGroup
+	processes, err := procfs.AllProcs()
+	if err == nil {
```
csweichel, line 158: I really struggle to parse the code - mainly because of the deep nesting and combination of Go routines and defers.

I'm currently trying to write an equivalent that's a bit more "left aligned". 
Would you be up for a 15min sync session to walk me through?

components/supervisor/pkg/terminal/terminal.go
csweichel: > You mean it is not ok to assume that the main terminal process is a shell? 

In case of regular workspace tasks that is safe to assume.
AFAIK supervisor executes `dockerd` in a terminal but not in a shell.

components/supervisor/pkg/terminal/terminal.go
svenefftinge: Yes, me too. I've rewritten / left-aligned it. Also added a test.

components/supervisor/pkg/supervisor/supervisor.go
svenefftinge: Good point. 👍  I clearly overused waitgroups here.

akosyakov:
I would test following things:
1. Have a process which spawns child processes and then trap SIGTERM waits sometime and then kill itself child processes, if child processes are terminated before by supervisor. I would think it is a bug.
2. Start regular VS Code terminal run code from `How to test`, kill it and see what happens with the processes. Then do the same with task terminal from VS Code. It should be consistent.

Does it make sense?

svenefftinge:
> Have a task terminal which spawns child processes and then trap SIGTERM waits sometime and then kill child processes, if child processes are terminated before by supervisor. I would think it is a bug.

This is what we discussed and where I suggest we treat this as expected. The current implementation would first SIGTERM children.

components/supervisor/pkg/terminal/terminal_test.go
```diff
@@ -313,6 +315,43 @@ func TestConcurrent(t *testing.T) {
 	}
 }
 
+func TestClose(t *testing.T) {
```
svenefftinge, line 318: This test fails in our infrastructure, but runs ok in my preview env. No idea yet, why.

akosyakov:
/werft run no-test=true

/hold 

I skipped tests to try. To merge we need to fix them

svenefftinge:
Discussed with @csweichel that we want to approach this on the `supervisor init` level.
The idea is that we keep sending `SIGTERM` to the terminal processes but without any waiting. In the case of a bash, this would reparent the child processes to the `supervisor init` process.
The process reaper in supervisor init would take care of sending `SIGTERM` to them after they have been reparented.
We need to wait for those processes for a certain amount of seconds that is passed in as an env var from ws-daemon, because apparently (does anyone know?) Kubernetes doesn't reliably SIGKILL containers.

components/supervisor/pkg/terminal/terminal.go
```diff
@@ -77,52 +78,54 @@ func (m *Mux) Start(cmd *exec.Cmd, options TermOptions) (alias string, err error
...
+// Close closes all terminals.
 func (m *Mux) Close() error {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
+	var wg sync.WaitGroup
 	var err error
 	for k := range m.terms {
-		cerr := m.doClose(k, closeTerminaldefaultGracePeriod)
+		cerr := m.doClose(&wg, k)
```
csweichel, line 95: Still stubbing my toe on the waitgroup. Alternative:

```Go
var eg errgrp.Group
for k := range m.terms {
    // capture loop variable - DO NOT REMOVE THIS LINE
    k := k
    eg.Go(func() error {
        return m.doCLose(k)
    })
}
return eg.Wait()
```

components/supervisor/pkg/terminal/terminal.go
```diff
@@ -146,44 +149,63 @@ func (m *Mux) doClose(alias string, gracePeriod time.Duration) error {
...
-	// process did not exit in time. Let's kill.
-	return term.shutdownProcessImmediately()
+	return nil
 }
 
-func (term *Term) shutdownProcessImmediately() error {
-	err := term.Command.Process.Kill()
-	if err != nil && !strings.Contains(err.Error(), "os: process already finished") {
-		return err
+func waitForChildProcessesToTerminate(PID int) {
```
csweichel, line 179: The function doesn't do what it says on the tin. It sends sigkill rather than wait for the termination. Is that behaviour intentional?

svenefftinge: where does it send sigkill?

components/supervisor/pkg/terminal/terminal.go
svenefftinge: ah, I see you would keep the asynchronicity local.

svenefftinge:
closing because of https://github.com/gitpod-io/gitpod/pull/13432#issuecomment-1263501898

PR #13438, opened 2022-09-29
[server] properly parse ports of clone URLs

AlexTugarev:
Otherwise this breaks the lookup of services mapped by host (including the port.)

This is the quite apparent aftermath of enabling ports in SCM locations.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Fix missing port in parsed clone URL.
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/server/src/repohost/repo-url.ts
+++ b/components/server/src/repohost/repo-url.ts
@@ -10,7 +10,7 @@ export namespace RepoURL {
         repoUrl: string,
     ): { host: string; owner: string; repo: string; repoKind?: string } | undefined {
         const u = new URL(repoUrl);
-        const host = u.hostname || "";
+        const host = u.host || "";
         const path = u.pathname || "";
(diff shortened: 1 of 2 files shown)


components/server/src/repohost/repo-url.ts
```diff
@@ -10,7 +10,7 @@ export namespace RepoURL {
         repoUrl: string,
     ): { host: string; owner: string; repo: string; repoKind?: string } | undefined {
         const u = new URL(repoUrl);
-        const host = u.hostname || "";
+        const host = u.host || "";
```
AlexTugarev, line 13: `host` contains the port, whereas `hostname` does not. 
that leads to a failure in lookup for services for host (which is taken from the SCM connection, thus needs to match)

geropl (approved):
Change LGTM, tests are green: :heavy_check_mark:

Issue #13239, opened 2022-09-23
High error rate on 'openPort'


geropl:
The last week we regularly see nigh rates on the `openPort` API call. Things that seem to happen:
 - client (IDE) calls `api.openPort`
 - `server` talks to `ws-manger`
 - `ws-manager` at some point returns error codes `RESOURCE_EXHAUSTED` and `DEADLINE_EXCEEDED`
 - client does retry

This leads to these questions:
 - What is the root cause for `DEADLINE_EXCEEDED`?
 - Could it be that `RESOURCE_EXHAUSTED` is a sign for clients to re-try too often (rate limited)? :thinking:

kylos101:
[High rate](https://grafana.gitpod.io/explore?orgId=1&left=%7B%22datasource%22:%22P4169E866C3094E38%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22editorMode%22:%22code%22,%22expr%22:%22sum%28rate%28grpc_server_handled_total%7Bservice%3D%5C%22gitpod-ws-manager%5C%22,%20grpc_method%3D%5C%22ControlPort%5C%22%7D%29%29%20by%20%28cluster,%20grpc_code%29%22,%22legendFormat%22:%22__auto%22,%22range%22:true,%22instant%22:true%7D%5D,%22range%22:%7B%22from%22:%22now-24h%22,%22to%22:%22now%22%7D%7D):
![image](https://user-images.githubusercontent.com/1272076/192064653-6df927cd-59ba-4af2-9a2f-5e7f529d0e47.png)

easyCZ:
<img width="1608" alt="Screenshot 2022-09-24 at 23 50 33" src="https://user-images.githubusercontent.com/1419286/192119862-151bd8a0-1893-48ab-bf77-40bf4597b837.png">

The above shows requests (and failures with codes) across our systems. Most notably, `WorkspaceManager.ControlPort` is returning ResourceExhausted. 

[...]

PR #13449, opened 2022-09-29
[supervisor] local exposed

mustard-mh:
## Description

Expose ports locally in supervisor (private is default available in ws-proxy)

**Context**

`ws-proxy` served every port in private, the user can access them or not is depending on if the ports are served.

This means auto `openPort` with private (call server API) in supervisor makes no sense (they are already in `private` state). So we can do it in supervisor locally

**Result**

With this PR, we can reduce 99% of useless requests, 99% comes from **Affect** section below

**Affect**

> _(Ignore `.gitpod.yml` ports defined with public)_

From Grafana, we can see today's `openPorts` calling total is around 60k, but users only change ports visibility (via vscode) 314 times today.

|Grafana|
|-|
|image|

|VSCode|
|-|
|image|

## Related Issue(s)

- Related https://github.com/gitpod-io/gitpod/issues/13239

## How to test

Example repo https://github.com/mustard-mh/test/tree/hw/logs-ports will open 100+ ports and mark port `8080` as public only

### Will not break default behavior
- Open workspace with a repo that provides both private and public ports in preview env and check if they work
- Change ports visibility by `gp-cli` or `PortsView` in VSCode (Do we provide more ways to do it?) and check if they work
- Check if ports status in `gp-cli` and `PortsView` are correct

### Will reduce the number of requests
- Open workspace with lots of ports defined/served with private and only one with public in `.gitpod.yml` on starting
- Check metrics of the num of `ControlPort` gRPC method called only one time

### How to check metrics of gRPC
Open Gitpod with this PR, exec command below
```
kubectl port-forward svc/prometheus-k8s -n monitoring-satellite 9090:9090
```
Search `grpc_client_handled_total{grpc_method='ControlPort'}` in graph tab

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/supervisor/pkg/ports/exposed-ports.go
+++ b/components/supervisor/pkg/ports/exposed-ports.go
@@ -6,6 +6,8 @@ package ports
 
 import (
 	"context"
+	"fmt"
+	"net/url"
 	"time"
 
 	backoff "github.com/cenkalti/backoff/v4"
@@ -57,9 +59,14 @@ func (*NoopExposedPorts) Expose(ctx context.Context, local uint32, public bool) 
 // GitpodExposedPorts uses a connection to the Gitpod server to implement
 // the ExposedPortsInterface.
 type GitpodExposedPorts struct {
-	WorkspaceID string
-	InstanceID  string
-	C           gitpod.APIInterface
+	WorkspaceID  string
+	InstanceID   string
+	WorkspaceUrl string
+	C            gitpod.APIInterface
+
+	localExposedPort   []uint32
+	localExposedNotice chan struct{}
+	lastServerExposed  []*gitpod.WorkspaceInstancePort
 
 	requests chan *exposePortRequest
 }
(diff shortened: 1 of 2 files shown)


akosyakov:
> Expose ports **locally**

@mustard-mh What does it mean?

mustard-mh:
> > Expose ports **locally**
> 
> 
> 
> @mustard-mh What does it mean?

make supervisor don't send openPort api if its private

components/supervisor/pkg/ports/exposed-ports.go
```diff
@@ -174,14 +220,19 @@ func (g *GitpodExposedPorts) doExpose(req *exposePortRequest) {
...
+		}
+		c := make(chan error)
+		close(c)
+		return c
 	}
 	req := &exposePortRequest{
 		port: &gitpod.WorkspaceInstancePort{
 			Port:       float64(local),
-			Visibility: v,
+			Visibility: "private",
```
jeanp413, line 235: shouldn't it be public?

```suggestion
			Visibility: "public",
```

components/supervisor/pkg/ports/exposed-ports.go
```diff
@@ -174,14 +220,19 @@ func (g *GitpodExposedPorts) doExpose(req *exposePortRequest) {
...
-	if public {
-		v = "public"
+	if !public {
+		if !g.existInLocalExposed(local) {
+			g.localExposedPort = append(g.localExposedPort, local)
+			g.localExposedNotice <- struct{}{}
+		}
+		c := make(chan error)
+		close(c)
+		return c
```
jeanp413, line 230: Should we remove from this list if we make a port public?

components/supervisor/pkg/ports/exposed-ports.go
```diff
@@ -105,22 +132,41 @@ func (g *GitpodExposedPorts) Observe(ctx context.Context) (<-chan []ExposedPort,
...
+					LocalPort: uint32(p.Port),
+					Public:    p.Visibility == "public",
+					URL:       p.URL,
+				}
+			}
+			exposedPort := make([]ExposedPort, 0, len(res))
+			for _, p := range res {
+				exposedPort = append(exposedPort, p)
+			}
+			return exposedPort
```
jeanp413, line 156: Why not push to array directly? is it because we are not filtering public ports in `localExposedPort`?

components/supervisor/pkg/ports/exposed-ports.go
mustard-mh: We make ports public/private in `vscode` and `gp-cli` by calling server API, so we don't need to handle them in `supervisor`.

In supervisor methods below will call it
- Private method `autoExpose`
  - (internal) Will be called for according to config (`.gitpod.yml`)
  - (internal) Will be called after ports are served
  - (exposed) Will be called by `RetryAutoExpose` has no definition for `visibility`

https://github.com/gitpod-io/gitpod/blob/bca27705c6499b2319c1fbd2673b7dda285fa245/components/supervisor-api/port.proto#L77-L79
- (exposed) Method `ExposePort` has no definition for `visibility`too

https://github.com/gitpod-io/gitpod/blob/bca27705c6499b2319c1fbd2673b7dda285fa245/components/supervisor-api/control.proto#L22-L27

mustard-mh:
Updated **How to Test** section. 

cc @jeanp413 @iQQBot Could you review and test this PR?

mustard-mh:
/hold

To ask WorkspaceTeam's confirm [internal chat](https://gitpod.slack.com/archives/C02F19UUW6S/p1665491591264149)

jeanp413 (approved):

PR #13470, opened 2022-09-29
[common-go] Add support to redact logs

aledbf:
## Description

Implement a custom JSON logrus Formatter to allow the redaction of fields using a [jq implementation in Go](https://github.com/itchyny/gojq)

## Related Issue(s)

xref https://github.com/gitpod-io/gitpod/pull/13419

## How to test
- Check tests
 
## Release Notes
```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

/hold


MrSimonEmms:
It's a simple change for Self-Hosted, so happy to approve but there's issues with running a workspace.

 - a [workspace](https://gitpodio-templatetypesc-hj7cx2rjodo.ws.aledbf-redact.preview.gitpod-dev.com/) ([source](https://github.com/gitpod-io/template-typescript-node)) built from `gitpod/workspace-full` a
image 
 - a [workspace](https://aledbf-redact.preview.gitpod-dev.com/#https://github.com/MrSimonEmms/gitpod-template-brainfuck) built from a custom Docker image doesn't build
image

csweichel:
I took the liberty and added benchmarks for this new redaction and the "old redaction" method:
image

What do these levels of allocation and time requirements do to our performance if we run either function for all our log statements?

aledbf:
> What do these levels of allocation and time requirements do to our performance

Good question.

but the benchmark is comparing two different things. The proposed approach here also removes the `envvars`, `env`, and `ssh_public_keys` arrays.

> (if has("envvars") then .envvars = [] else . end) |
> (if has("ssh_public_keys") then .ssh_public_keys = [] else . end) |
> (if has("env") then .env = [] else . end) |

that said, the question is still valid.

I would try to use this only in ws-manager, the affected component and use that first step to iterate.

easyCZ:
Moving back to draft as it does not look reviewable in the current state (needing rebase). Feel free to move back to in review once reviewable.

kylos101:
Removing from the workspace project for now.

PR #13471, opened 2022-09-29
[certificates]: validate that certs are valid for a Gitpod install

### LINKED ISSUE #11865: Add KOTS support bundle checks to validate certificate domain
We recently had a customer who had configured their deployment with the URL `db-incluser.company.com`, but had issued the certs against their corporate CA to `db-incluster.company.com` (note the `incluser` vs `incluster` - no 🍵).

We should add some checks into KOTS to see if we can validate that the certificate secret has been issued correctly - namely to `$DOMAIN`, `*.$DOMAIN` and `*.ws.$DOMAIN`.

It would be preferable to check this in the Support Bundle stage rather than the Pre-Flight checks. If the user is issuing them via cert-manager (or some other autogenerated method) then they may not exist at the Pre-Flight stage. It's unlikely that cert-manager would misconfigure them (especially if done via KOTS), all Gitpod cares about is that the `https-certificates` secret is correctly configured so this is a worthwhile generic check to do.

MrSimonEmms:
## Description

Add some rudimentary checks for the certificate - importantly, these are the skateboard not the finished article. These download and decode the certificate secret and perform the following checks:
1. that the secret exists and is able to be decoded as an OpenSSL x509 certificate
2. that `$DOMAIN`, `*.$DOMAIN` and `*.ws.$DOMAIN` exist in the secret (the subject alternate names seems to be non-standard, so this is a simple `grep`)
3. that the current timestamp is between the `notBefore` and `notAfter` timestamps

**NB** it does **NOT** validate that the certificate's CA is known and installed in the cluster. That's important but it's beyond the 🛹 

As with the other KOTS checks, these are perhaps a little rudimentary but they give an indication of whether the provided state is Gitpod-compatible. As there's a time-based test, it is plausible (however unlikely) that this could fail 5 minutes later if the cert expiry then occurs - again, this can be tweaked as a 🚲 

**IMPORTANT** this is not provided as a pre-flight check. As the majority of our users will be using LetsEncrypt and cert-manager to generate their certs, these are not triggered until _AFTER_ the first deployment. This would mean that the pre-flight checks are likely to fail. Instead, this is part of the troubleshooting to provide answers if the installation is not working.

## Related Issue(s)

Fixes #11865

## How to test

> Example support bundle - [file](https://vendor.replicated.com/troubleshoot/analyze/2022-09-30@11:18/contents/support-bundle-2022-09-30T11_15_35/certificate/certificate.log) and [analyzer](https://vendor.replicated.com/troubleshoot/analyze/2022-09-30@11:18)
 
Install via KOTS and go to "troubleshoot" to build a support bundle. This has been checked with both a LetsEncrypt and a self-signed cert built in-cluster. I do not have a valid, publicly-signed cert that's not provided by LetsEncrypt.

To check the edge-cases that KOTS smooths away for us, you can run the script in the image directly

```bash
docker run -it --rm \
  -v "${HOME}/.kube:/root/.kube" \
  eu.gcr.io/gitpod-core-dev/build/kots-config-check/certificate:mrsimonemms-add-kots-support-bundle-11865.1 \
  <domain> \
  gitpod \
  https-certificates
```

## Release Notes

```release-note
[certificates]: validate that certs are valid for a Gitpod install
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/kots-config-check/certificate/entrypoint.sh
+++ b/components/kots-config-check/certificate/entrypoint.sh
@@ -1,0 +1,105 @@
+#!/bin/bash
+# Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+# Licensed under the GNU Affero General Public License (AGPL).
+# See License-AGPL.txt in the project root for license information.
+
+set -euo pipefail
+
+DOMAIN="${1:-""}"
+NAMESPACE="${2:-""}"
+SECRET_NAME="${3:-""}"
+TLS_CRT_KEY="${4:-"tls.crt"}"
+TLS_KEY_KEY="${5:-"tls.key"}"
+
+cert_exists="false"
+domain="false"
+in_date="false"
+
+CRT_FILE="/tmp/tls.crt"
+CRT_CONTENTS_FILE="/tmp/tls.crt.txt"
+KEY_FILE="/tmp/tls.key"
+
+function get_cert() {
+  # Get certificate from secret
+  kubectl get secret \
+    -n "${NAMESPACE}" \
+    "${SECRET_NAME}" \
+    -o jsonpath="{.data.${TLS_CRT_KEY//./\\.}}" \
+    | base64 -d \
+    > "${CRT_FILE}" || return 1
+
+  kubectl get secret \
+    -n "${NAMESPACE}" \
+    "${SECRET_NAME}" \
+    -o jsonpath="{.data.${TLS_KEY_KEY//./\\.}}" \
+    | base64 -d \
+    > "${KEY_FILE}" || return 1
+
+  # Decode it as an x509 certificate
+  openssl x509 -in "${CRT_FILE}" -text -noout > "${CRT_CONTENTS_FILE}"
+
+  CRT_SIG="$(openssl x509 -noout -modulus -in "${CRT_FILE}" | openssl md5)"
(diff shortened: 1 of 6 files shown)


roboquat:

<details>

</details>

components/kots-config-check/certificate/entrypoint.sh
```diff
@@ -0,0 +1,105 @@
...
+  if cert_matches_domain_name; then
+    domain="true"
+  fi
+
+  if cert_in_date; then
+    in_date="true"
+  fi
+fi
+
+if [ "${cert_exists}" = "true" ]; then
```
Pothulapati, line 89: nit: can't we combine these if else clauses with the ones above and do the `echo` write there instead of here? 🤔

Pothulapati (approved):
LGTM! 👍🏼 

holding on the small nit, but feel free to remove it if it's not worth it.

/hold

components/kots-config-check/certificate/entrypoint.sh
MrSimonEmms: I've left it like that in case we do any future stuff. I also like to separate my business logic from my view logic - I treat any `echo` as view logic, even if it's a one-liner.

Are you ok if we merge as-is?

Pothulapati: Makes sense!

> are we ready to merge as-is?
Yep

PR #13484, opened 2022-09-30
[installer] Support enabling protected secrets

csweichel:
## Description
Adds support for enabling protected secrets in the installer.

## How to test
Run the installer with the following config:
```YAML
experimental:
  workspace:
    enableProtectedSecrets: true
```

and observe how the server configmap now contains a new default feature flag.

## Release Notes

```release-note
Add experimental support for protected secrets
```

### DIFF
--- a/install/installer/pkg/config/v1/experimental/experimental.go
+++ b/install/installer/pkg/config/v1/experimental/experimental.go
@@ -119,6 +119,8 @@ type WorkspaceConfig struct {
 		// Deprecated
 		UsageReportBucketName string `json:"usageReportBucketName"`
 	} `json:"contentService"`
+
+	EnableProtectedSecrets bool `json:"enableProtectedSecrets"`
 }
 
 type PersistentVolumeClaim struct {
--- a/install/installer/pkg/components/server/configmap.go
+++ b/install/installer/pkg/components/server/configmap.go
@@ -163,6 +163,14 @@ func configmap(ctx *common.RenderContext) ([]runtime.Object, error) {
 		return nil
 	})
 
+	defaultFeatureFlags := []NamedWorkspaceFeatureFlag{}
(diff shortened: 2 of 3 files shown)


easyCZ:
Currently it changes the rendering of the installer config when No flag is specified to `"defaultFeatureFlags": null`. Is that intentional? It used to be the empty list `[]`

Furisto:
Not intentional, but also not harmful I think. I will change it anyway, both to keep it consistent and to fix the test failure.

easyCZ (approved):
LGTM.

An aside, this is one thing which the `ws-director` would be able to help us solve - not needing to give DefaultWorkspaceFlags to `server` and only configure it in the director which is way closer to the problem at hand.

install/installer/pkg/config/v1/experimental/experimental.go
```diff
@@ -119,6 +119,8 @@ type WorkspaceConfig struct {
 		// Deprecated
 		UsageReportBucketName string `json:"usageReportBucketName"`
 	} `json:"contentService"`
+
+	EnableProtectedSecrets bool `json:"enabledProtectedSecrets"`
```
nandajavarma, line 123: nit: is `enabled` a typo? Just because in the example you gave in the description it was `enable` and kinda took me a minute to realize. :)

nandajavarma (approved):
Added a nit comment. But works well. Adding a hold for visibility.

/hold

install/installer/pkg/config/v1/experimental/experimental.go
csweichel: Yes, this is a typo. Great catch :)

nandajavarma:
/unhold 
(lol I made a typo on unhold)

PR #13495, opened 2022-09-30
[supervisor] shutdown processes gracefully

### LINKED ISSUE #13365: [supervisor] Processes started in terminals do not get a SIGTERM signal on stop

svenefftinge:
## Description
Changes the shutdown logic so that on SIGTERM or SIGINT on the `supervisor init` process we are sending SIGTERM to `supervisor` which in turn sends SIGTERM on all its terminal processes.
When supervisor finished we iterate over all of `init`'s child processes until the grace period is up.

## Related Issue(s)

Fixes #13365

## How to test
Start a workspace https://se-supervi88c18dbe3b.preview.gitpod-dev.com/#https://github.com/svenefftinge/termination-test

Run 
```
./ignoresigterm > out_disown.txt &
disown $1
./ignoresigterm > out_bg.txt &
./ignoresigterm > out.txt
```
and stop the workspace.
With this preview environment you should see log output like below in all three files
```
awaiting signal

Received signal
terminated

Ignored for 0 seconds.
Ignored for 1 seconds.
Ignored for 2 seconds.
Ignored for 3 seconds.
Ignored for 4 seconds.
Ignored for 5 seconds.
Ignored for 6 seconds.
Ignored for 7 seconds.
Ignored for 8 seconds.
Ignored for 9 seconds.
Ignored for 10 seconds.
Ignored for 11 seconds.
Ignored for 12 seconds.
Ignored for 13 seconds.
Ignored for 14 seconds.
Ignored for 15 seconds.
...
```

With gitpod.io you will see only (i.e. the processes never received SIGINT):
```
awaiting signal
```

## Release Notes

```release-note
All running processes receive a SIGTERM when a workspace shuts down.
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/common-go/process/process.go
+++ b/components/common-go/process/process.go
@@ -4,6 +4,14 @@
 
 package process
 
+import (
+	"context"
+	"errors"
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
 // IsNotChildProcess checks if an error returned by a command
 // execution is an error related to no child processes running
 // This can be seen, for instance, in short lived commands.
@@ -13,4 +21,37 @@ func IsNotChildProcess(err error) bool {
 	}
 
 	return (err.Error() == "wait: no child processes" || err.Error() == "waitid: no child processes")
+}
+
+var ErrForceKilled = errors.New("Process didn't terminate, so we sent SIGKILL")
+
+// TerminateSync sends a SIGTERM to the given process and returns when the process has terminated or when the context was cancelled.
+// When the context is cancelled this function sends a SIGKILL to the process and return immediately with ErrForceKilled.
+func TerminateSync(ctx context.Context, pid int) error {
+	process, err := os.FindProcess(pid)
+	if err != nil { // never happens on UNIX
+		return err
+	}
+	err = process.Signal(unix.SIGTERM)
+	if err != nil {
+		if err == os.ErrProcessDone {
+			return nil
+		}
+		return err
+	}
+	terminated := make(chan error, 1)
+	go func() {
+		_, err := process.Wait()
+		terminated <- err
+	}()
+	select {
+	case err := <-terminated:
+		return err
+	case <-ctx.Done():
+		err = process.Kill()
+		if err != nil {
+			return err
+		}
+		return ErrForceKilled
+	}
 }
--- a/components/supervisor/cmd/init.go
+++ b/components/supervisor/cmd/init.go
@@ -5,15 +5,23 @@
 package cmd
 
 import (
+	"context"
+	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"os/signal"
 	"strings"
+	"sync"
 	"syscall"
+	"time"
 
 	"github.com/gitpod-io/gitpod/common-go/log"
+	"github.com/gitpod-io/gitpod/common-go/process"
+	"github.com/gitpod-io/gitpod/supervisor/pkg/supervisor"
+	"github.com/prometheus/procfs"
+	reaper "github.com/ramr/go-reaper"
 	"github.com/spf13/cobra"
-	"golang.org/x/sys/unix"
 )
 
 var initCmd = &cobra.Command{
@@ -21,34 +29,14 @@ var initCmd = &cobra.Command{
 	Short: "init the supervisor",
 	Run: func(cmd *cobra.Command, args []string) {
 		log.Init(ServiceName, Version, true, false)
-		// Because we're reaping with PID -1, we'll catch the child process for
-		// which we've missed the notification anyways.
+		cfg, err := supervisor.GetConfig()
+		if err != nil {
+			log.WithError(err).Info("cannnot load config")
+		}
 		var (
-			sigInput      = make(chan os.Signal, 1)
-			sigReaper     = make(chan os.Signal, 1)
-			sigSupervisor = make(chan os.Signal, 1)
+			sigInput = make(chan os.Signal, 1)
 		)
-		signal.Notify(sigInput, syscall.SIGCHLD, os.Interrupt, syscall.SIGTERM)
-		go func() {
-			for s := range sigInput {
-				switch s {
-				default:
-					sigSupervisor <- s
-					// the reaper needs all signals so that it can turn into
-					// a terminating reaper if need be.
-					fallthrough
-				case syscall.SIGCHLD:
-					// we don't want to blob the SIGINT/SIGTERM behaviour because
-					// the reaper is still busy.
-					select {
-					case sigReaper <- s:
-					default:
-					}
-				}
-			}
-		}()
-
-		go reaper(sigReaper)
+		signal.Notify(sigInput, os.Interrupt, syscall.SIGTERM)
 
 		supervisorPath, err := os.Executable()
 		if err != nil {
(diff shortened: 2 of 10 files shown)


roboquat:

<details>

</details>

svenefftinge:
Adding hold because I want to make sure this gets thoroughly tested.

/hold

svenefftinge:
@csweichel One thing I noticed is that when I start the hanging go process directly in a terminal bash and then `kill -15 <PID>` the bash, the go process dies as well without a SIGTERM but is still listed as a zombie process in the proc table.
When I nest this in another bash and the SIGTERM the nested bash, the go process gets reparented and continues to run.

components/common-go/process/process.go
```diff
@@ -14,3 +22,36 @@ func IsNotChildProcess(err error) bool {
 
 	return (err.Error() == "wait: no child processes" || err.Error() == "waitid: no child processes")
 }
+
+var ErrForceKilled = errors.New("process didn't temrinate, so we send SIGKILL")
+
+// TerminateSync sends a SIGTERM to the given process and returns when the process has terminated but at most the given gracePeriod
+// if the gracePeriod ends we send SIGKILL to the process and return immediately with ErrForceKilled
+func TerminateSync(pid int, gracePeriodOver *time.Timer) error {
```
svenefftinge, line 30: bad API design. pass in a <-chan instead

components/supervisor/cmd/init.go
```diff
@@ -81,61 +61,58 @@ var initCmd = &cobra.Command{
...
+					err = g.Wait()
+					if err != nil {
+						log.WithError(err).Error("terminating child processes")
+					}
+				}
+
+			}()
+			// wait for either successful termination or the timeout
+			select {
+			case <-timer.C:
```
svenefftinge, line 109: give go routines a but of time to send the sigkills.

components/supervisor/cmd/init.go
```diff
@@ -81,61 +61,58 @@ var initCmd = &cobra.Command{
 		case <-supervisorDone:
 			// supervisor has ended - we're all done here
 			return
-		case s := <-sigSupervisor:
+		case <-sigInput:
 			// we received a terminating signal - pass on to supervisor and wait for it to finish
-			_ = runCommand.Process.Signal(s)
-			<-supervisorDone
+			cfg, err := supervisor.GetConfig()
```
svenefftinge, line 66: move up

components/supervisor/pkg/supervisor/config.go
```diff
@@ -410,6 +413,14 @@ func (c WorkspaceConfig) getCommit() (commit *gitpod.Commit, err error) {
 	return
 }
 
+func (c WorkspaceConfig) GetTerminationGracePeriod() time.Duration {
```
svenefftinge, line 416: todo validation

components/common-go/process/process.go
```diff
@@ -14,3 +21,36 @@ func IsNotChildProcess(err error) bool {
 
 	return (err.Error() == "wait: no child processes" || err.Error() == "waitid: no child processes")
 }
+
+var ErrForceKilled = errors.New("process didn't temrinate, so we send SIGKILL")
+
+// TerminateSync sends a SIGTERM to the given process and returns when the process has terminated but at most the given gracePeriod
+// if the gracePeriod ends we send SIGKILL to the process and return immediately with ErrForceKilled
+func TerminateSync[T any](pid int, gracePeriodOver <-chan T) error {
```
easyCZ, line 29: Why do you use generics here?

I think if all you're looking for is a signal from the outside then the idiomatic would be `chan struct{}`

components/supervisor/cmd/init.go
```diff
@@ -81,61 +65,56 @@ var initCmd = &cobra.Command{
 		case <-supervisorDone:
 			// supervisor has ended - we're all done here
 			return
-		case s := <-sigSupervisor:
+		case <-sigInput:
 			// we received a terminating signal - pass on to supervisor and wait for it to finish
-			_ = runCommand.Process.Signal(s)
-			<-supervisorDone
+			timer := time.After(cfg.GetTerminationGracePeriod())
+			terminationDone := make(chan int)
```
easyCZ, line 71:
```suggestion
			terminationDone := make(chan struct{})
```
is more idiomatic and uses fewer bytes

components/common-go/process/process.go
```diff
@@ -14,3 +21,36 @@ func IsNotChildProcess(err error) bool {
...
+	if err != nil { // never happens on UNIX
+		return err
+	}
+	err = process.Signal(unix.SIGTERM)
+	if err == os.ErrProcessDone {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
```
easyCZ, line 40: Normally I'd see this written as 
```
if err != nil {
  if err == os.ErrProcessDone { ..}
  otherwise..
}
```
but I don't think it matters much

components/common-go/process/process.go
easyCZ, line 29: Is it acceptable to go terminate the process even if at the start of this method the channel already has a value?

If not, you can do
```
select {
  case <- gracePeriodOver:
      return ..
   default:
}
```
Which will check the channel, but if no value is available it will continue

components/common-go/process/process.go
```diff
@@ -14,3 +21,36 @@ func IsNotChildProcess(err error) bool {
...
+	terminated := make(chan error, 1)
+	go func() {
+		_, err := process.Wait()
+		terminated <- err
```
easyCZ, line 44: If the process never stops, and we reach `gracePeriodOver` we will leak this go-routine and also the `terminated` channel.

What's the expected behaviour here? Is it that `process.Kill()` causes `process.Wait()` to unblock?

components/supervisor/cmd/init.go
```diff
@@ -81,61 +65,56 @@ var initCmd = &cobra.Command{
 		case <-supervisorDone:
 			// supervisor has ended - we're all done here
 			return
-		case s := <-sigSupervisor:
+		case <-sigInput:
```
easyCZ, line 68: For readability, I'd move this whole block into a separate function

components/supervisor/cmd/init.go
```diff
@@ -81,61 +65,56 @@ var initCmd = &cobra.Command{
...
+						log.WithError(err).Error("terminating child processes")
+					}
+				}
+
+			}()
+			// wait for either successful termination or the timeout
+			select {
+			case <-timer:
+				// Time is up, but we give all the goroutines a bit more time to react to this.
+				time.Sleep(time.Millisecond * 500)
```
easyCZ, line 111: This would leak the goroutine (and terminationDone channel) above. Does that matter?

easyCZ (commented):
Left a couple of Qs.

In general, I feel a lot of the code would become much more readable if we synchronised using a `context.Context`, in particular using the context.Done() property ([docs](https://pkg.go.dev/context#WithCancel)) than passing a channel as input to a function.

Also beware of the specific of `time.After(x)`. If the value is not read, it leaks - this can happen here when the processes do terminate within the period gracefully. See https://www.arangodb.com/2020/09/a-story-of-a-memory-leak-in-go-how-to-properly-use-time-after/

svenefftinge:
I decided against using context because of the side effect of kill the process when done. That seems surprising.

easyCZ:
> I decided against using context because of the side effect of kill the process when done. That seems surprising.

I'm not sure I follow. Could you explain, please?

I'd have expected that at the point we begin to handle the termination, we do the following:
```
ctx, cancel := context.WithDeadline(context.Background(), period)

// pass context into relevant methods, each async method does:
select {
  case <-ctx.Done():
	 exit..
  case <- something else
 
}
```

components/common-go/process/process.go
svenefftinge: I haven't thought of the leakage of the routine, but, yes, process.Kill() will unblock Wait().

components/supervisor/cmd/init.go
svenefftinge: The process ends here.

svenefftinge:
We don't exit but force kill the process, which I find a bit surprising if this was a side effect of a context cancellation:
```go
case <-gracePeriodOver:
		err = process.Kill()
		if err != nil {
			return err
		}
		return ErrForceKilled
	}
```

components/supervisor/cmd/init.go
easyCZ: In that case, should it contain `os.Exit(code)` then? It would at least communicate if it terminated gracefully, or not.

easyCZ (commented):
How would you feel about at least logging the PIDs being killed? I feel that for the amount of complexity this contains, we've got zero visibility into what it's actually doing.

components/supervisor/cmd/init.go
easyCZ: Also, I think it's worth flushing the logger at this point to guarantee any unwritten bytes are actually written.

easyCZ:
> which I find a bit surprising if this was a side effect of a context cancellation:

But the context doesn't kill the processes. It is just a signal that you've exceeded the grace period (or the time allowed for you to propagate the signal into) - in this respect, it's the same as a call stack of request. It's more less replacing the current argument you're passing in to indicate it's run out of time, just more standardized and wider used.

If you want, happy to pair to see what this would look like otherwise it matters not.

components/common-go/process/process.go
svenefftinge: It allows passing in `<- chan time.Time` (from time.After) but also other chans and communicate that we are only interested in the event.

svenefftinge:
Ok, I think if we put it like that it's ok to use context. API-wise it fits better and takes care of the leakage you mentioned

components/common-go/process/process.go
svenefftinge: Thanks, we do want run through the code including the effect of SIGKILL at the end.

components/supervisor/cmd/init.go
svenefftinge: > In that case, should it contain os.Exit(code) then? It would at least communicate if it terminated gracefully, or not.

But it just ends regularly here. So no reason to exit with an error code. Do I miss something?

easyCZ: Normally, when you do not terminate gracefully you signal that by setting a non-zero exit code to the shell. So in this case, we'd set it when the deadline is reached

svenefftinge: But it is not this process that doesn't terminate gracefully, but processes that were started by users during the workspace session. I don't think we should signal through error code on this level if such a process doesn't behave well

easyCZ: Good point. I guess what I'm after is to be able to tell the user (or even measure it ourselves) when we exit without completing the termination - this would then signal to users that they should possibly increase the grace termination period (in their config) but in general help them understand if the stop task did complete or not.

svenefftinge: Yes, that's important. The plan is to write the stdout and stderr of "shutdown" tasks [[pr](https://github.com/gitpod-io/gitpod/pull/11287)] to some file in `/workspace/.gitpod`. Do you think that is good enough? We could alternatively also write a shutdown protocol to that location.

svenefftinge: I think I like the idea of writing a termination protocol to a file in that folder. Will look into it.

components/supervisor/cmd/init.go
```diff
@@ -76,66 +64,113 @@ var initCmd = &cobra.Command{
...
 }
 
+// terminateAllProcesses terminates all processes but ours until there are none anymore or the context is cancelled
+// on context cancellation any still running processes receive a SIGKILL
+func terminateAllProcesses(ctx context.Context, slog shutdownLogger) {
+	for {
+		processes, err := procfs.AllProcs()
+		if err != nil {
+			log.WithError(err).Error("Cannot list processes")
+			slog.write(fmt.Sprintf("Cannot list processes: %s", err))
```
akosyakov, line 106: return is missing?

akosyakov:
@svenefftinge everything but `./ignoresigterm > out.txt` works for me. I tried from both supervisor and VS Code terminals.

akosyakov (approved):
ok, my bad, it turned out that one need to trap SIGHUP otherwise foreground processes don't receive a signal

PR #13503, opened 2022-10-01
New Default `.gitpod.yml` Template while using `gp init`

Siddhant-K-code:
> Educate More about Gitpod through `gp init`
Signed-off-by: Siddhant Khare <siddhant@gitpod.io>

## Description
This PR Change introduces the new `.gitpod.yml` template. Which Describes all the possible syntax/ features we support in `.gitpod.yml`. The old template only suggests the tasks & ports. And Misses the crucial things like educating more about Prebuilds, Extensions, Gitpod GitHub App badges, etc. & respective Docs. References will help users to learn about Gitpod, "How X Works"...

## Related Issue(s)

Fixes [Internal Discussion](https://www.notion.so/gitpod/Speed-up-the-Inner-Loop-for-Configuration-Changes-f1fa5e7ce80c4b0b8a044c1fa9046465#2205fcb3160d4fc39dfc0017f6893e80)

## How to test

- Open [preview dev environment](https://gitpod-cli559a93a73a.preview.gitpod-dev.com/workspaces)
- Open an empty workspace from [empty repository from preview env](https://gitpod-cli559a93a73a.preview.gitpod-dev.com/#https://github.com/gitpod-io/empty)
- Open Terminal & Run `gp init`

It will look something like this:

image

## Release Notes

```release-note
chore: New `.gitpod.yml` template while using `gp init`
```

## Werft options:

- [x] /werft with-preview
- [ ] /werft with-integration-tests=ide

### DIFF
--- a/components/gitpod-cli/cmd/init.go
+++ b/components/gitpod-cli/cmd/init.go
@@ -52,15 +52,139 @@ Create a Gitpod configuration for this project.
 			log.Fatal(err)
 		}
 		if !interactive {
-			d = []byte(`# List the start up tasks. Learn more https://www.gitpod.io/docs/config-start-tasks/
-tasks:
-  - init: echo 'init script' # runs during prebuild
-    command: echo 'start script'
+			d = []byte(`## Learn more about this file at 'https://www.gitpod.io/docs/references/gitpod-yml'
 
-# List the ports to expose. Learn more https://www.gitpod.io/docs/config-ports/
-ports:
-  - port: 3000
-    onOpen: open-preview
+## This '.gitpod.yml' file when placed at the root of a project instructs
+## Gitpod how to prepare & build the project, start development environments
+## and configure continuous prebuilds. Prebuilds when enabled builds a project
+## like a CI server so you can start coding right away - no more waiting for
+## dependencies to download and builds to finish when reviewing pull-requests
+## or hacking on something new.
+
+## With Gitpod you can develop software from any device (even iPads) via
+## desktop or browser based versions of VS Code or any JetBrains IDE and
+## customise it to your individual needs - from themes to extensions, you
+## have full control.
+
+## The easiest way to try out Gitpod is install the browser extenion:
+## 'https://www.gitpod.io/docs/configure/user-settings/browser-extension' or by prefixing
+## 'https://gitpod.io#' to the source control URL of any project.
+##
+## For example: 'https://gitpod.io#https://github.com/gitpod-io/gitpod'
+
+## The 'image' section defines which Docker image Gitpod should use.
+## By default, Gitpod uses a standard Docker Image called 'workspace-full'
+## which can be found at 'https://github.com/gitpod-io/workspace-images'
+
+## Workspaces started based on this default image come pre-installed with
+## Docker, Go, Java, Node.js, C/C++, Python, Ruby, Rust, PHP as well as
+## tools such as Homebrew, Tailscale, Nginx and several more.
+
+## If this image does not include the tools needed for your project then
+## a public Docker image or your own Docker file can be configured.
+##
+## Learn more about images at 'https://www.gitpod.io/docs/configure/workspaces/workspace-image'
+
+#image: node:buster                        # use 'https://hub.docker.com/_/node'
+#
+#image:                                    # leave image undefined if using a Dockerfile
+#  file: .gitpod.Dockerfile                # relative path to the Dockerfile from the
+#                                          # root of the project
+
+## The 'tasks' section defines how Gitpod prepares and builds this project
+## or how Gitpod can start development servers. With Gitpod, there are three
+## types of tasks:
+
+## - before: Use this for tasks that need to run before init and before command.
+## - init: Use this to configure prebuilds of heavy-lifting tasks such as
+##         downloading dependencies or compiling source code.
+## - command: Use this to start your database or application when the workspace starts.
+
+## Learn more about these tasks at 'https://www.gitpod.io/docs/configure/workspaces/tasks'
+
+#tasks:
+#  - before: |
+#      # commands to execute...
+#
+#  - init: |
+#      # sudo apt-get install python3     # can be used to install operating system
+#                                         # dependencies but these are not kept after the
+#                                         # prebuild completes thus Gitpod recommends moving
+#                                         # operating system dependency installation steps
+#                                         # to a custom Dockerfile to make prebuilds faster
+#                                         # and to keep your codebase DRY.
+#                                         # 'https://www.gitpod.io/docs/configure/workspaces/workspace-image'
+#
+#      # pip install -r requirements.txt  # install codebase dependencies
+#      # cmake                            # precompile codebase
+#
+#  - name: Web Server
+#    openMode: split-left
+#    env:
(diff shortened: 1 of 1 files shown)


components/gitpod-cli/cmd/init.go
```diff
@@ -52,15 +52,138 @@ Create a Gitpod configuration for this project.
...
-ports:
-  - port: 3000
-    onOpen: open-preview
+			d = []byte(`## Learn more about this file at 'https://www.gitpod.io/docs/references/gitpod-yml'
+
+## This '.gitpod.yml' file when placed at the root of a project instructs
+## Gitpod how to prepare & build the project, start development environments
+## and configure continuous prebuilds. Prebuilds when enabled builds a project
+## like a CI server so you can start coding right away - no more waiting for
+## dependencies to download and builds to finish when reviewing pull-requests
```
filiptronicek, line 61: suggestion (non-blocking): remove the dash

It seems like GitHub Docs point to the feature just being called `pull requests`

```suggestion
## dependencies to download and builds to finish when reviewing pull requests
```

components/gitpod-cli/cmd/init.go
```diff
@@ -52,15 +52,138 @@ Create a Gitpod configuration for this project.
...
+## or hacking on something new.
+
+## With Gitpod you can develop software from any device (even iPads) via
+## desktop or browser based versions of VS Code or any JetBrains IDE and
+## customise it to your individual needs - from themes to extensions, you
+## have full control.
+
+## The easiest way to try out Gitpod is install the browser extenion:
```
filiptronicek, line 69: suggestion (nit): `install` is either mising a `to` or a `-ing` + typo
```suggestion
## The easiest way to try out Gitpod is to install the browser extension:
```

```suggestion
## The easiest way to try out Gitpod is installing the browser extension:
```

components/gitpod-cli/cmd/init.go
```diff
@@ -52,15 +52,138 @@ Create a Gitpod configuration for this project.
...
-
-# List the ports to expose. Learn more https://www.gitpod.io/docs/config-ports/
-ports:
-  - port: 3000
-    onOpen: open-preview
+			d = []byte(`## Learn more about this file at 'https://www.gitpod.io/docs/references/gitpod-yml'
+
+## This '.gitpod.yml' file when placed at the root of a project instructs
+## Gitpod how to prepare & build the project, start development environments
+## and configure continuous prebuilds. Prebuilds when enabled builds a project
```
filiptronicek, line 59:
```suggestion
## and configure continuous prebuilds. Prebuilds (when enabled) build a project
```

components/gitpod-cli/cmd/init.go
```diff
@@ -52,15 +52,138 @@ Create a Gitpod configuration for this project.
...
+## like a CI server so you can start coding right away - no more waiting for
+## dependencies to download and builds to finish when reviewing pull-requests
+## or hacking on something new.
+
+## With Gitpod you can develop software from any device (even iPads) via
+## desktop or browser based versions of VS Code or any JetBrains IDE and
+## customise it to your individual needs - from themes to extensions, you
```
filiptronicek, line 66: suggestion: American spelling

```suggestion
## customize it to your individual needs - from themes to extensions, you
```

components/gitpod-cli/cmd/init.go
```diff
@@ -52,15 +52,138 @@ Create a Gitpod configuration for this project.
...
+#ports:
+#  - port: 8080 # alternatively configure entire ranges via '8080-8090'
+#    visibility: private # either 'public' or 'private' (default)
+#    onOpen: open-browser # either 'open-browser', 'open-preview' or 'ignore'
+
+## The 'vscode' section defines a list of Visual Studio Code extensions from
+## the OpenVSX.org registry to be installed upon workspace startup. OpenVSX
+## is an open alternative to the proprietary Visual Studio Code Marketplace
+## and extensions can be added by sending a pull-request with the extension
+## identifier to https://github.com/open-vsx/publish-extensions
```
filiptronicek, line 150: thought: The goal of the `vscode` section is not to describe extensions, but rather VS Code-specific config, which extensions are a part of. Would change this to something like `The vscode.extensions section...` or explaining that this is not its sole purpose - although being only useful for extensions at the moment.

filiptronicek, line 150: suggestion (non-blocking: could we please link to https://www.gitpod.io/docs/references/gitpod-yml/#vscodeextensions for further info instead? It would be good to have fewer places to have to maintain about this.

Siddhant-K-code: didn't got you on this. I think it is clearly specified that it defines a list that we want to use.

components/gitpod-cli/cmd/init.go
```diff
@@ -52,15 +52,138 @@ Create a Gitpod configuration for this project.
...
+
+## Learn more about ports at 'https://www.gitpod.io/docs/configure/workspaces/ports'
+
+#ports:
+#  - port: 8080 # alternatively configure entire ranges via '8080-8090'
+#    visibility: private # either 'public' or 'private' (default)
+#    onOpen: open-browser # either 'open-browser', 'open-preview' or 'ignore'
+
+## The 'vscode' section defines a list of Visual Studio Code extensions from
+## the OpenVSX.org registry to be installed upon workspace startup. OpenVSX
```
filiptronicek, line 147: suggestion: correct the public Open VSX deployment domain and the name of the project as well

```suggestion
## the open-vsx.org registry to be installed upon workspace startup. Open VSX
```

components/gitpod-cli/cmd/init.go
filiptronicek: I am trying to say that from the comment, it seems like we are saying that the section `vscode` of the `.gitpod.yml` file is used to define a list of VS Code extensions, and I think that's not the whole truth: I think we should mention that only the `extensions` section inside of the `vscode` section is used for this purpose.

Siddhant-K-code: @loujaybee WDYT?

svenefftinge (commented):
I'm voting against this change as it generates a lot of noise and boilerplate into the user's project. Generating a simple start config together with a link to the documentation seems much better.

Siddhant-K-code:
my take: `init` for every CLI/ template/ product. is like 404/homepage with max views. We should use it to educate more we can and in templates, we can show it as an example.

loujaybee:
> I'm voting against this change as it generates a lot of noise and boilerplate into the user's project. Generating a simple start config together with a link to the documentation seems much better - @svenefftinge 

Agree with @svenefftinge here. 

@Siddhant-K-code let's discuss this one. 

It appears we're trying to solve an education problem here with our users but we should consider: 

* **We should avoid "front-loading" our users with info** - When it comes to onboarding we want to reveal our users to features at the right time, and avoid putting everything in front of them at once, risking overload. 
* **Users don't always need documentation** - Sadly, as much as we like, users often don't read documentation. If we can put in guardrails for users to prevent them doing the wrong thing, or slowly reveal functionality bit-by-bit, this can help. 

There's some interesting aspects of this as a solution, let's figure out the right place to put this info, and the right time to surface the information to users 🙏

PR #13505, opened 2022-10-01
proposal: `gp prebuild-logs` Gitpod cli command

Siddhant-K-code:
**Important**

> **Warning** This PR is work in progress, and a proposal. It will not be merged without prior discussions, however please treat as **low priority**. 

---

> Signed-off-by: Siddhant Khare <siddhant@gitpod.io>

## Description
This PR Introduces the new command for Gitpod CLI, `gp prebuild-logs` which will open the Prebuild Logs File (_if it exists_)

## Related Issue(s)

Fixes #

## How to test

- Open preview dev environment
- Open an empty workspace from empty repository from preview env
- Open Terminal & Run `gp preuibld-logs`

## Release Notes

```release-note
feature: new gitpod cli command for prebuild logs - `gp prebuild-logs`
```

## Documentation

Yes. (will update it, after merge)

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-cli/cmd/prebuild-logs.go
+++ b/components/gitpod-cli/cmd/prebuild-logs.go
@@ -1,0 +1,60 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package cmd
+
+import (
+	"github.com/gitpod-io/gitpod/gitpod-cli/pkg/theialib"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"os"
+)
+
+// initCmd represents the init command
+var prebuildLogs = &cobra.Command{
+	Use:   "prebuild-logs",
+	Short: "Opens logs of Gitpod prebuilds",
+	Long: `
+Opens all logs of Gitpod prebuilds.
+Number of logs depends upon the init tasks you have configured.
+	`,
+	// Args:  cobra.MinimumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+
+		const prebuildFilePath = "/workspace/.gitpod/prebuild-log-*"
+		err := openPrebuildLogs(prebuildFilePath)
+
+		if err != nil {
+			log.WithError(err)
+			return
+		}
+
+		pcmd := os.Getenv("GP_OPEN_EDITOR")
+		if pcmd == "" {
+			log.Fatal("GP_OPEN_EDITOR is not set")
+			return
+		}
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(prebuildLogs)
+}
+
+func openPrebuildLogs(prebuildFilePath string) error {
+	service, err := theialib.NewServiceFromEnv()
+	if err != nil {
+		log.WithError(err)
+		return err
+	}
+
+	for {
+		_, err := service.OpenFile(theialib.OpenFileRequest{Path: prebuildFilePath})
+		if err != nil {
+			log.WithError(err).Println("Prebuild logs not found.\nMake sure you have configured prebuilds.\nLearn more about Prebuilds: https://www.gitpod.io/docs/configure/projects/prebuilds")
+			return nil
+		}
+	}
+
+}
--- a/components/gitpod-cli/cmd/prebuild-logs.go
+++ b/components/gitpod-cli/cmd/prebuild-logs.go
@@ -5,10 +5,10 @@
 package cmd
 
 import (
+	"fmt"
 	"github.com/gitpod-io/gitpod/gitpod-cli/pkg/theialib"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
-	"os"
 )
 
 // initCmd represents the init command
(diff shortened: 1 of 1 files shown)


svenefftinge:
Would it more sense for the command to print the prebuild logs to stdout instead of opening in an editor?

Siddhant-K-code:
> Would it make more sense for the command to print the prebuild logs to stdout instead of opening them in an editor?

Here is the scenario. The count of Prebuild Logs depends on the number of `init` tasks. So it might be not a good user flow to log out multiple files. 

image


And,  in some cases (mostly). Prebuild Logs are too long
image

loujaybee:
This looks interesting, @Siddhant-K-code 👀  and was an idea that we also had to support: 

- https://github.com/gitpod-io/website/issues/2424

To make it easier to iterate on the configurations. 

---

> Would it make more sense for the command to print the prebuild logs to stdout instead of opening them in an editor? - @svenefftinge 

* I agree with Sven, we should pick sensible defaults. printing to stdout is useful for text searching tools w/ `grep` (should we consider a flag to open in the text editor, e.g. `gp logs prebuild --edit`?)
* We are multi editor/IDE at Gitpod, let's be careful to not favour VS Code blindly. We should make sure our solution works with all the IDEs and for CLI users. That's one reason to favour printing to stdout as default / first iteration.

---

Have we considered `gp logs` as the namespace? Then we can extend in future with...

* `gp logs dotfiles`
* `gp logs prebuilds` 

...etc

* I'd appreciate @andreafalzetti's input here 
* @csweichel has also raised the need to improve our standardards for our `gp` CLI args - which I do agree with

---

Whilst promising, let's treat this PR as low(er) priority, so that the @gitpod-io/engineering-ide have appropriate time to give feedback and test the solution before merging. Related, but we should avoid PR's without issues, as we're discussing on an active PR, where ideally this could have been discussed as an issue to get more clarity on the solution first. 🙏

Siddhant-K-code:
> Related, but we should avoid PR's without issues, as we're discussing on an active PR, where ideally this could have been discussed as an issue to get more clarity on the solution first. 🙏

Ok, will remember that next time.
this is just the base of it because it was pending for a long time (& it was required) and there was no active discussion on it.
I have marked it as a draft, we can discuss the solution and will make the changes to it afterward.

akosyakov:
Although here is indeed many alternatives from user and implementation perspective. I wonder how we can skateboard which can be rolled back easily. How about contributing it only if latest version of IDE is selected and with experimental disclaimer in help. It does not need to be surfaced in docs and so on, but we could to use it internally or with customers via asking to choose latest and iterate.

components/gitpod-cli/cmd/prebuild-logs.go
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package cmd
+
+import (
+	"fmt"
+	"github.com/gitpod-io/gitpod/gitpod-cli/pkg/theialib"
```
akosyakov, line 9: We should ditch theialib. The official api is supervisor API, maybe we should think about how enrich it to return task output. We were discussing such API in the past with @geropl that dashboard could as well return output of prebuild task even if some task already finished. 

I think it would be in the right direction to make prebuild UX is more reliable. cc @csweichel

components/gitpod-cli/cmd/prebuild-logs.go
```diff
@@ -0,0 +1,54 @@
...
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+// initCmd represents the init command
+var prebuildLogs = &cobra.Command{
+	Use:   "prebuild-logs",
```
akosyakov, line 16: How about `gp prebuilds list` + `gp prebuilds log` to list human readable names of tasks and then fetch content.

components/gitpod-cli/cmd/prebuild-logs.go
```diff
@@ -0,0 +1,54 @@
...
+	Short: "Opens logs of Gitpod prebuilds",
+	Long: `
+Opens all logs of Gitpod prebuilds.
+Number of logs depends upon the init tasks you have configured.
+	`,
+	Run: func(cmd *cobra.Command, args []string) {
+
+		const prebuildFilePath = "/workspace/.gitpod/prebuild-log-*"
```
akosyakov, line 24: it is implementation detail of supervisor, i think it should be a secret of supervisor how to store and fetch content. Clients should not rely on it.

See https://en.wikipedia.org/wiki/Information_hiding

components/gitpod-cli/cmd/prebuild-logs.go
Siddhant-K-code: Currently `prebuild-logs-X` contains lot of things from binary characters to even process completion % (see [picture here](https://github.com/gitpod-io/gitpod/pull/13505#issuecomment-1264415276)). That's why I was saying to avoid printing in terminal.

akosyakov: I think they are special characters for shell and should be interpreted properly. We don't write anything else to these files besides what was produced by pty service for shell.

Issue #12345, opened 2022-08-24
Stuck on "Loading prebuild..." for 60 minutes


mikenikles:
### Bug description

Loading workspaces regularly gets stuck on the "Loading prebuild..." screen for 60 minutes, at which point it times out and a "Stopping workspace..." message shows up.

### Steps to reproduce

1. Load https://github.com/gitpod-io/website

### Workspace affected

gitpodio-website-yla6f7ogn1q

### Expected behavior

The workspace starts.

### Example repository

https://github.com/gitpod-io/website

It happens with any other repository too.

### Anything else?

It's a fairly common situation, but there are times when loading a workspace works. Over the last two weeks, I encountered the issue three times per day on average.

mikenikles:
After working with Firefox all day, it now too gets stuck on the "Loading prebuild..." screen for 60 minutes until it times out.

Prebuilds run successful and I've tried with "incremental prebuilds" on and off - same bug.

axonasif:
This has been happening lately but randomly.

mikenikles:
This was ok for ~5 days but it started to happen again yesterday & today.

sagor999:
Moved it into Breakdown phase for workspace  team.

kylos101:
I see the same problem with `gitpodio-gitpod-aditftfbhhh`.

```
gitpod /workspace/gitpod (main) $ gpctl workspaces list | grep 788c9d51-ea4e-4c78-864c-b8782ea54a18
8df3495b-685d-46e0-9820-009cc3b4afd8        gitpodio-gitpod-aditftfbhhh                 788c9d51-ea4e-4c78-864c-b8782ea54a18        INITIALIZING        REGULAR         ws-788c9d51-ea4e-4c78-864c-b8782ea54a18
```

```
kubectl describe pod ws-788c9d51-ea4e-4c78-864c-b8782ea54a18
...
  Warning  Unhealthy  4m34s (x897 over 19m)  kubelet  Readiness probe failed: Get "http://10.20.254.30:22999/_supervisor/v1/status/content/wait/true": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

Given these [logs](https://cloudlogging.app.goo.gl/HxVnWKs5uen1HyzM9), `supervisor` hasn't signaled that content-init is done.

No errors in the trace either:
![Image](https://user-images.githubusercontent.com/1272076/188238148-45dd19b9-bf55-4243-b006-2381cbc4be3d.png)

[...]

PR #13515, opened 2022-10-02
[ws-daemon] Improve download speed of snapshots

aledbf:
## Description

Improve the download of prebuild from the remote content storage using multiple connections.

fixes https://github.com/gitpod-io/gitpod/issues/12345

## How to test
- Add the role `new-workspace-cluster` to your account
- Check the `persistent_volume_claim` is not checked
- Deploy an ephemeral cluster
- Open a workspace using a prebuild
- Remove the role `new-workspace-cluster` from your account 
- Open a workspace using a the same prebuild
- Compare the times.
 
## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/ws-daemon/pkg/content/initializer.go
+++ b/components/ws-daemon/pkg/content/initializer.go
@@ -10,7 +10,6 @@ import (
 	"encoding/json"
 	"errors"
 	"io/ioutil"
-	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -390,13 +389,28 @@ func (rs *remoteContentStorage) Download(ctx context.Context, destination string
 	}
 
 	span.SetTag("URL", info.URL)
-	resp, err := http.Get(info.URL)
-	if err != nil {
-		return true, err
-	}
-	defer resp.Body.Close()
 
-	err = archive.ExtractTarbal(ctx, resp.Body, destination, archive.WithUIDMapping(mappings), archive.WithGIDMapping(mappings))
+	tempFile, err := os.CreateTemp("", "remote-content-")
+	if err != nil {
+		return true, xerrors.Errorf("cannot create temporal file: %w", err)
+	}
+	defer tempFile.Close()
+	defer os.Remove(tempFile.Name())
+
+	args := []string{
+		"-x16", "-j12",
+		info.URL,
+		"-o", tempFile.Name(),
+	}
+	cmd := exec.Command("aria2c", args...)
+	var out []byte
+	out, err = cmd.CombinedOutput()
+	if err != nil {
+		log.WithError(err).WithField("out", string(out)).Error("unexpected error downloading file")
+		return true, xerrors.Errorf("unexpected error downloading file")
+	}
+
+	err = archive.ExtractTarbal(ctx, tempFile, destination, archive.WithUIDMapping(mappings), archive.WithGIDMapping(mappings))
 	if err != nil {
 		return true, xerrors.Errorf("tar %s: %s", destination, err.Error())
 	}
--- a/components/ws-daemon/debug.Dockerfile
+++ b/components/ws-daemon/debug.Dockerfile
@@ -27,6 +27,7 @@ RUN apt update \
       apt-transport-https \
       python3-crcmod \
       gnupg \
+      aria2 \
   && echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list \
   && curl -sSL https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg add - \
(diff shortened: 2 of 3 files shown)


roboquat:
@aledbf: Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://git.k8s.io/community/contributors/guide/release-notes.md) to remove it.

<details>

</details>

utam0k:
This is an improvement by parallelization, right? I wonder how to impact a node by this change. How about doing loadgen?

components/ws-daemon/pkg/content/initializer.go
```diff
@@ -390,13 +389,28 @@ func (rs *remoteContentStorage) Download(ctx context.Context, destination string
...
+
+	tempFile, err := os.CreateTemp("", "remote-content-")
+	if err != nil {
+		return true, xerrors.Errorf("cannot create temporal file: %w", err)
+	}
+	defer tempFile.Close()
+	defer os.Remove(tempFile.Name())
+
+	args := []string{
+		"-x16", "-j12",
```
utam0k, line 401: Do we need to consider a self-hosted env?

aledbf:
> How about doing loadgen?

That would require access to prebuilds from the loadgen account and prepare a custom scenario.

components/ws-daemon/pkg/content/initializer.go
aledbf: What do you mean?

utam0k, line 401: Apparently, by default, retries are repeated 5 times. This seems a bit much.
https://aria2.github.io/manual/en/html/aria2c.html#cmdoption-m

utam0k:
> That would require access to prebuilds from the loadgen account and prepare a custom scenario.

Shall I put it as a note in the next release of the changelog? Will you have the deployer take a careful look at it?

components/ws-daemon/pkg/content/initializer.go
utam0k: Sorry for the lack of information :bow:  I think self-hosted environments may require less CPU cores than SaaS. I found it difficult to judge whether this degree of parallelism is appropriate at that time.

aledbf: We have the same sizes now as the default. We could choose values using the CPU core count, but I would prefer to do that in a different PR after we use it in SaaS and gather metrics.

utam0k: :+1:

kylos101:
> > That would require access to prebuilds from the loadgen account and prepare a custom scenario.
> 
> Shall I put it as a note in the next release of the changelog? Will you have the deployer take a careful look at it?

Luckily @aledbf is the deployer this week (on-call next week). :wink: 

**Note**: I think `loadgen` would require a code change to use a prebuild/snapshot initializer for benchmark. [Context, I think it assumes the git initializer](https://github.com/gitpod-io/gitpod/blob/df77406fc5d631e01978b23b653f0ce33e51aa9d/dev/loadgen/pkg/loadgen/generator.go#L215-L225). 

As an alternative to using `loadgen`, we could cordon the workspace nodes (but one) of an ephemeral cluster, and test a single node.

For example, start 8 standard Gitpod workspaces as fast as you can, have another person do the same, and observe how the node resources perform, do the workspaces start, etc. How does this perform while backups are happening concurrently? It'll be interesting to see how close we get to the memory limit for `ws-daemon`.

utam0k:
We are lucky!
> Hi @utam0k wave , luckily @aledbf is the deployer this week (on-call next week). wink

Ops, I forgot about it :bow: 
> Note: I think loadgen would require a code change to use a prebuild/snapshot initializer for benchmark. [Context, I think it assumes the git initializer](https://github.com/gitpod-io/gitpod/blob/df77406fc5d631e01978b23b653f0ce33e51aa9d/dev/loadgen/pkg/loadgen/generator.go#L215-L225).

Is this a limitation of ws-daemo itself? If I understand correctly, classic content-init should have very few restrictions since it runs on ws-daemon. If I am wrong, please let me know.
> It'll be interesting to see how close we get to the memory limit for ws-daemon.

Furisto:
> Is this a limitation of ws-daemo itself? If I understand correctly, classic content-init should have very few restrictions since it runs on ws-daemon. If I am wrong, please let me know.

No, it is just a limitation of loadgen at the moment.

utam0k:
> No, it is just a limitation of loadgen at the moment.
@Furisto 
I guess I don't know enough. May I ask you to elaborate a bit more?

sagor999 (approved):
LGTM
Added hold in case of, but I think discussion is not related to the changes. 
@aledbf feel free to unhold when you are ready, so that this change can go into next gen cluster. :pray:

mikenikles:
This was a community-reported bug via #12345. Without a changelog entry, how will I know when this is deployed?

utam0k:
> This was a community-reported bug via #12345. Without a changelog entry, how will I know when this is deployed?

If the schedule is as usual, maybe by the end of this week.

PR #13533, opened 2022-10-03
[installer-tests] Cleanup secret creation on Gitpod install

nandajavarma:
## Description

Previously we added a small feature that would backup the secret for TLS certificate so it can be reused at least for the same subdomain. But this feature was broken and was causing the test to fail. Along with this change, we broke the Github authentication secret creation as well. This PR basically cleans up the `make` target to `kots-install` a little bit by creating separate targets for both secret creations. 
Alongside, before destroying Gitpod, we back up the secret for certificate to a GCS bucket which is retrieved when recreating a setup with the same subdomain name. These secrets will get destroyed after 90 days.

## Related Issue(s)

Fixes #

## How to test

Try running a test using:
```
werft run github -j .werft/k3s-installer-tests.yaml -a subdomain=test-k3s
```

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/tests/Makefile
+++ b/install/tests/Makefile
@@ -7,6 +7,10 @@ TOPDIR=$(shell pwd)
 
 KUBECONFIG := "$(TOPDIR)/kubeconfig"
 
+cert_secret := ${TF_VAR_TEST_ID}-cert-secret.yaml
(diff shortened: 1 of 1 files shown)


adrienthebo (approved):
LGTM!

Issue #13125, opened 2022-09-20
Nearly 50% calls to joinTeam return 500 status code


easyCZ:
![Image](https://user-images.githubusercontent.com/1419286/191266737-6bb9b1e5-6cd5-4bff-935e-39132e87458b.png)

Query: `sum(increase(gitpod_server_api_calls_total{method="joinTeam"}[7d])) by (statusCode)`
[Dashboard](https://grafana.gitpod.io/explore?orgId=1&left=%7B%22datasource%22:%22P4169E866C3094E38%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22editorMode%22:%22code%22,%22expr%22:%22sum%28increase%28gitpod_server_api_calls_total%7Bmethod%3D%5C%22joinTeam%5C%22%7D%5B7d%5D%29%29%20by%20%28statusCode%29%22,%22legendFormat%22:%22__auto%22,%22range%22:true,%22instant%22:true%7D%5D,%22range%22:%7B%22from%22:%22now-7d%22,%22to%22:%22now%22%7D%7D)

jankeromnes:
Good catch! 💯 

Looking at the code, I have a strong suspicion that this is about trying to join a team where you're already a member:

https://github.com/gitpod-io/gitpod/blob/3ac0c02e2d62b0755177baa0591db8399c4caebe/components/gitpod-db/src/typeorm/team-db-impl.ts#L182-L184

(Otherwise, if a valid invite really wasn't found, it would be a 404 response:

https://github.com/gitpod-io/gitpod/blob/3ac0c02e2d62b0755177baa0591db8399c4caebe/components/server/src/workspace/gitpod-server-impl.ts#L2124-L2127

)

easyCZ:
It is, I looked at the logs for this. The API behaves weirdly here, it throws an error with "Already in a team" (or similar) which it then checks for with a regex on the client side. Here, we should probably use the 409 HTTP Status code to indicate that the membership already exists.

easyCZ:
Fix for returning appropriate status codes: https://github.com/gitpod-io/gitpod/issues/13125

easyCZ:
Since the rollout, the 500 errors have disappeared.


![Image](https://user-images.githubusercontent.com/1419286/194725737-d7505835-8ed9-4a6c-81f4-73e5958e6803.png)

Marking as complete.

PR #13534, opened 2022-10-03
[server] Throw specific errors when interacting with teams

easyCZ:
## Description

Throw more accurate errors such that they don't show up as generic 500 errors in metrics

## Related Issue(s)

* Relates to https://github.com/gitpod-io/gitpod/issues/13125

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/gitpod-db/src/typeorm/team-db-impl.ts
+++ b/components/gitpod-db/src/typeorm/team-db-impl.ts
@@ -15,6 +15,8 @@ import { DBTeam } from "./entity/db-team";
 import { DBTeamMembership } from "./entity/db-team-membership";
 import { DBUser } from "./entity/db-user";
 import { DBTeamMembershipInvite } from "./entity/db-team-membership-invite";
+import { ResponseError } from "vscode-jsonrpc";
+import { ErrorCodes } from "@gitpod/gitpod-protocol/lib/messaging/error";
 
 @injectable()
 export class TeamDBImpl implements TeamDB {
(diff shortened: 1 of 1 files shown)


components/gitpod-db/src/typeorm/team-db-impl.ts
```diff
@@ -122,27 +124,30 @@ export class TeamDBImpl implements TeamDB {
 
     public async createTeam(userId: string, name: string): Promise<Team> {
         if (!name) {
-            throw new Error("Team name cannot be empty");
+            throw new ResponseError(ErrorCodes.BAD_REQUEST, "Team name cannot be empty");
```
andrew-farries, line 127: As a general point about all of these changes, should we throwing `ResponseErrors` from the `gitpod-db` component? We don't throw these errors anywhere else in `gitpod-db`.

easyCZ: In general, I'd also be reluctant but we'd need to define a new set of errors, catch these and then re-throw. Given how intertwined the db model is with server, I see somewhat limited harm in doing this here tho. It still throws, just more specific.

easyCZ: But happy to go another route if we think it's worth it.

geropl: I think Andrew's comment is on-point: It feels like this method belongs more into `server`, or even `protocol`.

For (nearly?) all other entities we have those methods in `protocol` (cmp.: [TeamSubscription](https://github.com/gitpod-io/gitpod/blob/5c1074bbfcdfc07829f87c6a7085f3146eb1b3af/components/gitpod-protocol/src/team-subscription-protocol.ts#L26-L30)).

You could still argue whether there shouldn't be a "model-level" error in between. But would still feel better :upside_down_face:

easyCZ: So if I understand this correctly, you wan to move this method over to `server` or `protocol`. Once there, do we want to re-map the errors or are we okay to keep them as they are?

svenefftinge: I'm +1 on this change as I don't think having error translations all over the place helps us.

geropl: > error translations all over the place helps us.

Valid point.

Let's go with the change as-is, because:
 1. it fixes a concrete pain-point
 2. we don't really have a working counter proposal that is either a) established or b) we see any immediate benefit in

If we resolve 2. at some point in the future, we can always go back and adjust. :slightly_smiling_face:

Issue #13521, opened 2022-10-03
Usage view date range is off for previous month 


jldec:
For the non-current months, the usage view is querying for a date range which ends one day before the end of the month.

<img width="1294" alt="Screenshot 2022-10-03 at 10 22 10" src="https://user-images.githubusercontent.com/849592/193532538-f0a2563e-b7f3-4871-ae75-c3be7fb38097.png">

easyCZ:
[Fix PR is here](https://github.com/gitpod-io/gitpod/pull/13543)

easyCZ:
Re-assigning to @jankeromnes for validation.

PR #13543, opened 2022-10-03
[usage] Fix month boundaries in usage view

easyCZ:
## Related Issue(s)

* Relates to https://github.com/gitpod-io/gitpod/issues/13521
* Depends on https://github.com/gitpod-io/gitpod/pull/13548

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/dashboard/src/components/UsageView.tsx
+++ b/components/dashboard/src/components/UsageView.tsx
@@ -24,6 +24,7 @@ import { ReactComponent as UsageIcon } from "../images/usage-default.svg";
 import { toRemoteURL } from "../projects/render-utils";
 import { WorkspaceType } from "@gitpod/gitpod-protocol";
 import PillLabel from "./PillLabel";
+import dayjs from "dayjs";
 
 interface UsageViewProps {
     attributionId: AttributionId;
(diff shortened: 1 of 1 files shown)


svenefftinge:
Would be great to fix this without switching to the [deprecated](https://momentjs.com/docs/#/-project-status/) moment.js library. 
To fix the underlying issue we need to compute the month start using UTC timezone. 
We can use `Date#getTimezoneOffset()` for that.

easyCZ:
> Would be great to fix this without switching to the [deprecated](https://momentjs.com/docs/#/-project-status/) moment.js library. 

In that case here's removal of it from the entire codebase: https://github.com/gitpod-io/gitpod/pull/13548

andrew-farries (approved):

jankeromnes (commented):
Apologies for the conflict here with https://github.com/gitpod-io/gitpod/pull/13568 ! 🙈 I did the fix as a drive-by since I was already working on this code, and only found out about the issue after opening my PR.

However, maybe a one-line fix is preferable over importing yet another third-party library + a significant code refactoring? 😬

easyCZ:
Firstly, it doesn't just import yet another library, it actually [removes Moment](https://github.com/gitpod-io/gitpod/pull/13548) and replaces it with a much smaller dependency - there's already a win in that.
Secondly, I'd rather pull in a dependency to make future me (and others) not have to spend as much effort understanding the obtuse API that native Date provides when a much cleaner version exists.

jankeromnes:
Many thanks for explaining the context @easyCZ! (And sorry again for the collision 🙈)

I knew Moment was deprecated, but I assumed it continued to work for now as a way to provide nicer "human-readable times" (e.g. "3 months ago" instead of "2022-07-06T08:07:10.243Z").

I didn't know we had decided to replace it with Dayjs (today I learned). For what it's worth, it does seem like a nice drop-in replacement. I especially like that it doesn't pull in any extra dependencies 💯 (on the other hand, it does pull in [a lot of devDependencies](https://github.com/iamkun/dayjs/blob/c9370ea96bf420439ee7eaa4146e8ed643160312/package.json#L68-L98), but I guess that's okay as it impacts only the developers of Gitpod and not all the end users of Gitpod).

I'll do my best not to interfere with this transition again. 🙏

easyCZ:
> I didn't know we had decided to replace it with Dayjs (today I learned)

We didn't decide. I just did it because if usage of it was not gonna get a PR approved, well then we may as well remove it so that others don't fall in the same trap. Ask for forgivness, not permission.

PR #13548, opened 2022-10-03
[server] Removes momentjs and replaces it with dayjs

easyCZ:
## Description

Because we don't want to use it due it being deprecated: https://github.com/gitpod-io/gitpod/pull/13543#issuecomment-1265675920

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


components/ee/db-sync/package.json
```diff
@@ -21,7 +21,6 @@
     "@gitpod/gitpod-db": "0.1.5",
     "@gitpod/gitpod-protocol": "0.1.5",
     "argparse": "^2.0.1",
-    "moment": "^2.21.0",
```
easyCZ, line 24: Was specified here but not used.

laushinka:
What's the best way to test that this migration didn't break anything? Just make sure `dashboard` builds?

easyCZ:
Basically. `dayjs` is API compatible with `momentjs`. I can also run a preview env for this just to sanity check from the reviewer side too.

laushinka:
`dashboard` builds and I checked in the preview all the views that was affected by the change.
I can approve the implementation but will hold depending on whether we would like a further discussion on the use of `dayjs` in these discussions: [[1](https://github.com/gitpod-io/gitpod/pull/13568)][[2](https://github.com/gitpod-io/gitpod/pull/13543)].
I see the merit of using the Date object instead of being dependent on another library, but it was also a bit of a pain to use it for our logic. 

/hold

laushinka (approved):

easyCZ:
This change alone does not make the problem any worse. We still use a library to do this albeit a non-deprecated one. And `dayjs` is smaller than moment so there's already a win in doing this.

If we want to go to native Date, it should be independent of this change, in my opinion.

laushinka:
> This change alone does not make the problem any worse. We still use a library to do this albeit a non-deprecated one. 

Excellent point. Using an external library and using a deprecated one are two different things. If we decide to go with the Date object we'll create a new ticket.

/unhold

PR #13550, opened 2022-10-03
[public-api] Use buf for codegen

easyCZ:
## Description

Uses [buf generate](https://docs.buf.build/generate/usage) to generate gRPC stubs instead of manually invoking `protoc`. The output is equivalent, but the configuration is much nicer.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


components/public-api/typescript/package.json
```diff
@@ -12,19 +12,16 @@
...
   "dependencies": {
-    "@gitpod/gitpod-protocol": "0.1.5",
-    "@grpc/grpc-js": "^1.3.7",
-    "google-protobuf": "^3.19.1",
-    "opentracing": "^0.14.4"
+    "@bufbuild/connect-web": "^0.2.1",
+    "@bufbuild/protobuf": "^0.1.1",
+    "@bufbuild/protoc-gen-connect-web": "^0.2.1",
+    "@bufbuild/protoc-gen-es": "^0.1.1",
+    "@gitpod/gitpod-protocol": "0.1.5"
```
akosyakov, line 19: I wonder why this one is necessary?

easyCZ:
The dependency on `gitpod-protocol` is due to the NPM publish scripts which live in there. I'll add an example for how to use it with a client.

akosyakov:
@easyCZ it should go to dev deps then otherwise it going to be pulled on npm install by clients

easyCZ:
@akosyakov The way we use the client is here https://connect.build/docs/web/using-clients#promises

It's also implemented in the original Connect prototype [here](https://github.com/gitpod-io/gitpod/pull/13507/files#diff-d80ed6f5137a2285e475089fefd8470058c703b3f4c04c46e0da0df873d3beabR73-R80)

easyCZ:
> @easyCZ it should go to dev deps then otherwise it going to be pulled on npm install by clients

Moved it to dev dependencies.

geropl (commented):
Awesome, really looking forward to actually use `connect-web`! :upside_down_face:

easyCZ:
> I understood that we would use this API also from within node and go not only browsers, correct?

Yes, we can. It doesn't remove our ability to use gRPC but we can use buf to get slightly simpler clients with less "fluff" that gRPC has. It's made more user-friendly.

PR #13554, opened 2022-10-04
[ws-manager] fix tests

sagor999:
## Description

Fix panic in test when preview env is failing.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
none
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/ws-manager/pkg/manager/annotations_test.go
+++ b/components/ws-manager/pkg/manager/annotations_test.go
@@ -28,6 +28,10 @@ func TestMarkWorkspace(t *testing.T) {
 	for _, test := range tests {
 		t.Run(test.Description, func(t *testing.T) {
 			manager := forTestingOnlyGetManager(t)
+			if manager == nil {
+				t.Errorf("cannot create test manager; this is a bug in the unit test itself")
+				return
+			}
 			startCtx, err := forTestingOnlyCreateStartWorkspaceContext(manager, "foo", api.WorkspaceType_REGULAR)
 			if err != nil {
 				t.Errorf("cannot create test pod start context; this is a bug in the unit test itself: %v", err)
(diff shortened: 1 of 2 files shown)


components/ws-manager/pkg/manager/annotations_test.go
```diff
@@ -28,6 +28,10 @@ func TestMarkWorkspace(t *testing.T) {
 	for _, test := range tests {
 		t.Run(test.Description, func(t *testing.T) {
 			manager := forTestingOnlyGetManager(t)
+			if manager == nil {
```
utam0k, line 31: I think `forTestingOnlyGetManager` have to return the `err` and check `err != nil`. WDYT?

sagor999: It does not return `err` though, and I did not want to change a whole lot. So that seemed like a fairly viable fix, especially if manager failed to get created, you will get an actual error in the test, just no panic anymore, which was very confusing.

jenting (commented):
🤔 it means that we have a chance that the unit tests succeed but it did not run the test because we failed to start the testing only ws-manager.

sagor999:
i don't think so. it still fails the test, only that it also panics. 
now it no longer panics, but will still fail the test.

jenting (approved):
LGTM

It'd be good if we could know the reason why the test-only manager can't be created successfully.

/hold because Toru has the concern

sagor999:
it happened due to preview env failure

components/ws-manager/pkg/manager/annotations_test.go
utam0k: :ok: I'll fix it again when I get a chance.

utam0k (approved):

PR #13568, opened 2022-10-04
[dashboard] Fix off-by-one error on usage range end date

jankeromnes:
## Description

Fix off-by-one error on selected Usage range (end date).

Explanation:
- `new Date(year, month, 0)` results in the last day of the previous month (e.g. `2022-09-30T00:00:00.000Z`)
  - But, using this date as an "end" for a selected range omits all the data from the 24 hours between `2022-09-30T00:00:00.000Z` to `2022-10-01T00:00:00.000Z`
- A better "end" for the selected range is `new Date(year, month)`, i.e. the start of the next month, which results in `2022-10-01T00:00:00.000Z`

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/13521

## How to test

In a Node.js or browser console, run this to verify:

```js
const today = new Date();

const startDate = new Date(today.getFullYear(), today.getMonth() - 1);
// 2022-09-01T00:00:00.000Z -- OK

const oldEndDate = new Date(today.getFullYear(), today.getMonth(), 0);
// 2022-09-30T00:00:00.000Z -- NOT OK (this is missing 24h of data)

const newEndDate = new Date(today.getFullYear(), today.getMonth());
// 2022-10-01T00:00:00.000Z -- OK
```

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/dashboard/src/components/UsageView.tsx
+++ b/components/dashboard/src/components/UsageView.tsx
@@ -112,7 +112,7 @@ function UsageView({ attributionId }: UsageViewProps) {
         for (let i = 1; i < 7; i++) {
             const endDateVar = i - 1;
             const startDate = new Date(today.getFullYear(), today.getMonth() - i);
-            const endDate = new Date(today.getFullYear(), today.getMonth() - endDateVar, 0);
+            const endDate = new Date(today.getFullYear(), today.getMonth() - endDateVar);
             const timeStampOfStartDate = startDate.getTime();
             const timeStampOfEndDate = endDate.getTime();
             rows.push(


roboquat:

<details>

</details>

easyCZ:
https://github.com/gitpod-io/gitpod/pull/13543 I think we ended up doing the same work..

laushinka:
>   * But, using this date as an "end" for a selected range omits all the data from the 24 hours between `2022-09-30T00:00:00.000Z` to `2022-10-01T00:00:00.000Z`

Aha! Thanks for this catch 💪🏼

jankeromnes:
> https://github.com/gitpod-io/gitpod/pull/13543 I think we ended up doing the same work..

@easyCZ Oh no 🙈 sorry about that! I was working on the same code when I noticed this, and put out a one-line fix in case that's helpful.

easyCZ:
I'd still prefer to go with https://github.com/gitpod-io/gitpod/pull/13543 as it makes it easier to comprehend - working with native Date is bit of a pain.

> The core problem was actually that we asked for the month in the browser's timezone which in CEST would include 2 hours from the previous month and lack the last two hours of the month.
We were missing the 24hours of data irrespective of that.

jankeromnes:
Personally, I'd prefer not using something called `dayjs` -- JavaScript's own `Date` standard library works great, and I don't think we need to import lots of other libraries that claim to do a better job.

A one-line fix seems more sensible.

Also, if there is indeed a problem with time zone differences, we can use this UTC one-liner:

https://github.com/gitpod-io/gitpod/blob/83cb1c7a7419474a3a5372a90fc1d23bd62a18a2/components/server/ee/src/user/stripe-service.ts#L150-L151

easyCZ (approved):
As you wish. But I feel you're ignoring accessibility of the code for future maintainers. Irrespective of that, please follow up with the MomentJS removal - either refactor the rest of the API calls to use native Date, or land the change to use `dayjs` - we want to be good scouts after all.

PR #13573, opened 2022-10-04
[db-sync] Optionally read DB credentials from env vars

geropl:
## Description
Optionally read DB credentials from env vars ([commit without formatting](https://github.com/gitpod-io/gitpod/pull/13573/commits/15bb1f3857868132557550d326369bd1e889292a)).

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/ee/db-sync/src/commands.ts
+++ b/components/ee/db-sync/src/commands.ts
@@ -12,9 +12,9 @@ import { PeriodicReplicatorProvider } from "./replication";
 import { connect } from "./database";
 import { injectable, inject } from "inversify";
 import { Config } from "@gitpod/gitpod-db/lib/config";
-import * as path from 'path';
+import * as path from "path";
 
-export const ICommand = Symbol('ICommand');
+export const ICommand = Symbol("ICommand");
 
 export interface ICommand {
     name: string;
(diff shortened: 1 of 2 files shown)


components/ee/db-sync/src/commands.ts
```diff
@@ -39,40 +39,63 @@ export class RunCommand implements ICommand {
...
         const config = JSON.parse(readFileSync(args.config).toString()) as ReplicationConfig;
 
-        if(config.roundRobin) {
-            if(config.source) {
+        if (process.env.DB_PASSWORD) {
+            console.info("Found DB_PASSWORD env var, using it to establish connections.");
+            if (config.source) {
+                config.source.password = process.env.DB_PASSWORD;
+            }
+            for (const target of config.targets) {
```
easyCZ, line 54: Looks like this handling of multiple targets was not present in here before. This PR doesn't seem to introduce the targets either. Is this just an oversight?

easyCZ: And if it should be happening even when the var is not set through ENV, we could hoist it outside of the `if (process.env.DB_*)`

geropl: I'm not sure I follow. [Besides the formatting](https://github.com/gitpod-io/gitpod/pull/13573/commits/15bb1f3857868132557550d326369bd1e889292a), we just check both env vars now, and use them to override all other DB usernames/passwords.

We effectively only use `roundRobin`, so `source` is not strictly required, but found it cleaner to add it as well.
Does that make sense? :thinking:

easyCZ: I was confused that we override all the targets which was not done originally, but I think I've wrapped my head around it  - it was statically provided before and now it's dynamic due to the ENV var override

easyCZ (approved):

PR #13579, opened 2022-10-04
[public-api] Switch go codegen to buf

easyCZ:
## Description

Uses [buf gen](https://docs.buf.build/generate/usage) to generate go/grpc 

This makes the configuration easier to read and removes the need to move files around.

This also removes the vendored dependency on google/rpc as it's not needed.

We can also the same for other packages.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


andrew-farries (approved):

andrew-farries:
This appears to be blocking the merge pool for some reason even though it still has pending reviewers 😕 

/hold

scripts/protoc-generator.sh
```diff
@@ -13,14 +13,18 @@ install_dependencies() {
...
 
+    go install github.com/bufbuild/buf/cmd/buf@v1.8.0
+
     curl -sSo /tmp/protoc-gen-grpc-java https://repo1.maven.org/maven2/io/grpc/protoc-gen-grpc-java/1.49.0/protoc-gen-grpc-java-1.49.0-linux-x86_64.exe
     chmod +x /tmp/protoc-gen-grpc-java
 }
 
 lint() {
-    local PROTO_DIR=${1:-.}
+    buf lint || exit 1
```
csweichel, line 23: This change makes it necessary to make the `lint` call come after `install_dependencies`. Otherwise scripts fail:
image

easyCZ: Good spot. I fixed it in a later PR but will pull that here.

easyCZ: fixed

akosyakov (commented):
> This also removes the vendored dependency on google/rpc as it's not needed.

I don't really understand what is happening in the PR. I thought we are using google rpc because of error handling? Is it not needed anymore or it is still generated differently?

easyCZ:
The default libraries (TS, go, etc..) which use the code already embed the google RPC structs in their responses and in the protocol. We don't need to vendor them here (we're also not importing them into our protos).

akosyakov (approved):
lgtm

PR #13577, opened 2022-10-04
[db] Set `uid` column when writing to the `d_b_oauth_auth_code_entry` table

andrew-farries:
## Description

https://github.com/gitpod-io/gitpod/pull/13524 added a new `uid` column to the `d_b_oauth_auth_code_entry` table.

This PR ensures that all new records written to the table set this new field.

## Related Issue(s)

Part of #9198 

Follow up to #13524 

## How to test

1. Run the local-companion app `local-app` against the preview environment:
```bash
GITPOD_HOST=https://af-set-uid-column.preview.gitpod-dev.com ./local-app test
```
2. Run through the auth flow.
3. See that the new rows in the `d_b_oauth_auth_code_entry` table have their `uid` column set.

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-local-preview
      If enabled this will build `install/preview`
- [ ] /werft with-preview
- [ ] /werft with-integration-tests=all
      Valid options are `all`, `workspace`, `webapp`, `ide`

PR #13524, opened 2022-10-03
[db] Add `uid` column to `d_b_oauth_auth_code_entry` table

andrew-farries:
## Description

Add a new `uid` column to the `d_b_oauth_auth_code_entry` table. 

This table should be synced by `db-sync` but currently uses an auto-incrementing PK which will cause clashes between regions. To resolve, the plan is:

1. Add a new `uid` column and populate it (this PR).
2. Switch the PK over to the `uid` column and remove the old auto-incrementing`id` column (later PR).
3. Sync the table using `db-sync` (later PR).

## Related Issue(s)

Part of #9198 

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-local-preview
      If enabled this will build `install/preview`
- [ ] /werft with-preview
- [ ] /werft with-integration-tests=all
      Valid options are `all`, `workspace`, `webapp`, `ide`

PR #13583, opened 2022-10-04
[db] Change the primary key of the `d_b_oauth_auth_code_entry` table

andrew-farries:
## Description

Following on from https://github.com/gitpod-io/gitpod/pull/13524 which added the `uid` column to the `d_b_oauth_auth_code_entry` table and https://github.com/gitpod-io/gitpod/pull/13577 which ensured that all new records set the `uid` field, this PR changes to the primary key to the `uid` column (and renames it to `id`).

## Related Issue(s)

Part of #9198 

## How to test

1. Run the local-companion app `local-app` against the preview environment:
```bash
GITPOD_HOST=https://af-make-uib1daf36a03.preview.gitpod-dev.com ./local-app test
```
2. Run through the auth flow.
3. See that the new rows in the `d_b_oauth_auth_code_entry` table have their `id` column set.

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-db/src/typeorm/migration/1664799675276-MakeUidColumnPrimaryKeyInAuthCodeTable.ts
+++ b/components/gitpod-db/src/typeorm/migration/1664799675276-MakeUidColumnPrimaryKeyInAuthCodeTable.ts
@@ -1,0 +1,25 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+const TABLE_NAME = "d_b_oauth_auth_code_entry";
+const PK_COLUMN = "id";
+const UID_COLUMN = "uid";
+
+export class MakeUidColumnPrimaryKeyInAuthCodeTable1664799675276 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`UPDATE d_b_oauth_auth_code_entry SET uid=(SELECT uuid()) where uid='';`);
+
+        await queryRunner.query(`ALTER TABLE ${TABLE_NAME} DROP COLUMN ${PK_COLUMN}`);
+        await queryRunner.query(
+            `ALTER TABLE ${TABLE_NAME} CHANGE ${UID_COLUMN} ${PK_COLUMN} char(36) NOT NULL DEFAULT '';`,
+        );
+        await queryRunner.query(`ALTER TABLE ${TABLE_NAME} ADD PRIMARY KEY (${PK_COLUMN});`);
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {}
+}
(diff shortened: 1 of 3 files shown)


easyCZ (approved):
What do we expect to happen when the migrations take place but existing instances are still running with PK set previously? I believe this will work fine, unless the `@PrimaryGeneratedColumn()` affects how the values are determined (and possibly generated)

/hold for Q

components/gitpod-db/src/typeorm/migration/1664799675276-MakeUidColumnPrimaryKeyInAuthCodeTable.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const TABLE_NAME = "d_b_oauth_auth_code_entry";
+const PK_COLUMN = "id";
+const UID_COLUMN = "uid";
+
+export class MakeUidColumnPrimaryKeyInAuthCodeTable1664799675276 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`UPDATE d_b_oauth_auth_code_entry SET uid=(SELECT uuid()) where uid='';`);
+
+        await queryRunner.query(`ALTER TABLE ${TABLE_NAME} DROP COLUMN ${PK_COLUMN}`);
```
geropl, line 17: Hm, is this really possible? :thinking: Have to look it up, actually.

Did you try to run this against an existing DB in the previous state?

geropl: Looks like you can do this in one `ALTER TABLE` statement: https://stackoverflow.com/a/2342193

andrew-farries:
> What do we expect to happen when the migrations take place but existing instances are still running with PK set previously? I believe this will work fine, unless the @PrimaryGeneratedColumn() affects how the values are determined (and possibly generated)

Not sure I understand the question. Do you mean is it possible to perform this key switch against a real table with data? If so, this migration has been run against such a table as part of testing this.

easyCZ:
When the migrations run, the application that's running will still have the @PrimaryGeneratedColumn() annotation - is this a problem? This could result in a generated value being written to the `id` property.

andrew-farries:
I think the table is infrequently used enough that we can take the chance of a write to the table happening during the migration. I'm not sure how we would avoid the problem anyway, actually.

components/gitpod-db/src/typeorm/migration/1664799675276-MakeUidColumnPrimaryKeyInAuthCodeTable.ts
geropl: > In general it appears you can drop PRIMARY KEY columns without issue.

Just verified: that's the case indeed. :+1: 

Also, because we run this inside a transaction, the `ADD PRIMARY KEY` should work, because we fill `uid` with concrete values before the move. :heavy_check_mark:.

geropl:
> I think the table is infrequently used enough that we can take the chance of a write to the table happening during the migration.

I think I agree with this sentiment. But we have to take into account that things happen, e.g.:
 - one of our assumptions about transactionality of the migration is wrong, and it fails mid-flight, leaving it up to us to fix the situation manually
 - this happens to get rolled out together with another, later migration that takes a lot of time (say, ~15mins) to execute: Suddenly, old code is broken because of this

> I'm not sure how we would avoid the problem anyway, actually.

I think we could decouple the migration from the code change, no? In a way where we:
 1. fill both, `uid` and `id` from code
 2. roll out the PK change (by setting id = uid instead of rename), and stop filling `uid` in code
 3. drop the old `uid` column altogether

Might be a little bit more work, but would make me feel much safer. :+1:

Issue #13157, opened 2022-09-21
Stripe: Create observability into the type, rate and return code of API calls we issue against Stripe


geropl:
As it says in the title. An example is the metric `gitpod_server_api_calls_total` ([link](https://grafana.gitpod.io/d/Gj5DE-O7k/meta-overview?orgId=1&editPanel=36)) from which we get insight into error rates, repsonse times and type of method called.

easyCZ:
Implementation-wise, we should use a HTTP Client tripper to compose an HTTP client which also exports metrics. See https://dev.to/stevenacoffman/tripperwares-http-client-middleware-chaining-roundtrippers-3o00 for examples

jankeromnes:
FYI, Stripe Dashboard > Developers already has this:

<table><tr><td>
<img width="755" alt="Screenshot 2022-09-21 at 10 58 00" src="https://user-images.githubusercontent.com/599268/191461942-30a32ea8-187f-4beb-ba81-5f6207ed324a.png">
</td><td>
<img width="750" alt="Screenshot 2022-09-21 at 10 58 11" src="https://user-images.githubusercontent.com/599268/191461949-1976fff1-9ea2-41bc-8a78-fed20242d20c.png">
</td></tr></table>

geropl:
Still seems worth implement the low-hanging parts on our end. a) Grafana is more flexible in displaying data, and b) we can build alerts with our common tools.

easyCZ:
> Still seems worth implement the low-hanging parts on our end. a) Grafana is more flexible in displaying data, and b) we can build alerts with our common tools.

Stripe will only have server metrics - for requests that actually reached it. If we fail to contact Stripe for whichever reason, we'd be blind - we need client side metrics also.

easyCZ:
Client-side metrics from server [PR is here](https://github.com/gitpod-io/gitpod/pull/13645)

easyCZ:
Metrics from server are coming through nicely on staging


![Image](https://user-images.githubusercontent.com/1419286/194399417-34ad7939-ab84-41f7-9bf8-293790119096.png)


However, metrics from `usage` include the full path in their labels which also includes Subscription IDs - this would lead to high cardinality of these metrics and will need to be reworked. We'll need to strip the IDs and sub-resources.
[...]

PR #13604, opened 2022-10-05
[stripe] Report metrics from client calls in Go

easyCZ:
## Description

Report metrics when stripe client makes calls

## Related Issue(s)

* Relates to https://github.com/gitpod-io/gitpod/issues/13157

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage/pkg/stripe/stripe.go
+++ b/components/usage/pkg/stripe/stripe.go
@@ -8,9 +8,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"os"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/gitpod-io/gitpod/usage/pkg/db"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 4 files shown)


andrew-farries (approved):

components/usage/pkg/stripe/stripe.go
```diff
@@ -51,7 +54,14 @@ func ReadConfigFromFile(path string) (ClientConfig, error) {
 
 // New authenticates a Stripe client using the provided config
 func New(config ClientConfig) (*Client, error) {
-	return &Client{sc: client.New(config.SecretKey, nil)}, nil
+	sc := &client.API{}
+
+	sc.Init(config.SecretKey, stripe.NewBackends(&http.Client{
```
easyCZ, line 59: @andrew-farries Actually found a cleaner way to do this using https://github.com/stripe/stripe-go/blob/v70.15.0/stripe.go#L847

easyCZ: It is weird `Init()` itself doesn't use this method to set them (even if it just pointed to the default http client)

PR #13605, opened 2022-10-05
[ws-daemon] Stop logging cpu limit error

Furisto:
## Description
Stop logging cpu limit error when the cgroup cannot be found. This pis caused by a delay between the pod disappearing and the dispatch being informed about it. It just pollutes the log and does not help us identifying errors.

## Related Issue(s)
n.a.

## Release Notes

```release-note
None
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/ws-daemon/pkg/cpulimit/dispatch.go
+++ b/components/ws-daemon/pkg/cpulimit/dispatch.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


components/ws-daemon/pkg/cpulimit/dispatch.go
```diff
@@ -161,7 +162,7 @@ func (d *DispatchListener) sink(id string, limit Bandwidth, burst bool) {
 	d.workspacesBurstCounterVec.WithLabelValues("none").Inc()
 
 	changed, err := ws.CFS.SetLimit(limit)
-	if err != nil {
+	if err != nil && !strings.HasPrefix(err.Error(), "failed to read") {
```
utam0k, line 165: How about defining errors in const instead of error messages?

Furisto: Thought it is not possible to define constants if the error message contains dynamic content (like the path in this case), but it turns out you can: https://smyrman.medium.com/writing-constant-errors-with-go-1-13-10c4191617. Not sure if it is worth it in this case. I can do it if you think it's better.

jenting (approved):
LGTM

components/ws-daemon/pkg/cpulimit/dispatch.go
jenting: From my point of view, it's fine.

utam0k: You are right!

PR #13537, opened 2022-10-03
feat(gitpod-cli): Create `gp info` command

WVerlaek:
## Description
Adds a `gp info` command, to print workspace details. For now, this includes:
- Workspace ID
- Instance ID
- Workspace class
- Workspace URL
- Cluster host

The intention is for this to replace the use of the matching environment variables (see #10797 for more details).

Includes a `--json` flag to print as JSON, allowing e.g. for parsing the output in scripts.

## Related Issue(s)

Fixes #10797

## How to test

```bash
$ cd components/gitpod-cli

$ go run . info
  Workspace ID    : gitpodio-gitpod-4q1200dvfcf                            
  Instance ID     : 8a6188a3-4c7f-49fe-aa4d-db968c11e183                   
  Workspace class : Large: Up to 8 vCPU, 16GB memory, 50GB disk            
  Workspace URL   : https://gitpodio-gitpod-4q1200dvfcf.ws-eu67.gitpod.io  
  Cluster host    : ws-eu67.gitpod.io                                      

$ go run . info --json
{"workspace_id":"gitpodio-gitpod-4q1200dvfcf","instance_id":"8a6188a3-4c7f-49fe-aa4d-db968c11e183","workspace_class":{"id":"g1-large","display_name":"Large","description":"Up to 8 vCPU, 16GB memory, 50GB disk"},"workspace_url":"https://gitpodio-gitpod-4q1200dvfcf.ws-eu67.gitpod.io","cluster_host":"ws-eu67.gitpod.io"}
```

## Release Notes

```release-note
Create `gp info` command
```

## Werft options:

- [ ] /werft with-local-preview
      If enabled this will build `install/preview`
- [ ] /werft with-preview
- [ ] /werft with-integration-tests=all
      Valid options are `all`, `workspace`, `webapp`, `ide`

PR #13607, opened 2022-10-05
chore(gitpod-cli): Update top table output format

WVerlaek:
## Description

Updates `gp top` table output to match `gp info`, follow up from discussion in https://github.com/gitpod-io/gitpod/pull/13537#discussion_r987641981.

Also moves the workspace class to the table. Alternatively we could remove the workspace class from the output, as it's also shown in `gp info` now. Personally I think it'd be nice to keep it in `gp top` as well as it's relevant info, and it would be a breaking change to remove it from e.g. the `--json` output.

Original:
image

New:
image

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Gitpod CLI: `gp top` table output updated to match `gp info`
```

## Werft options:

- [ ] /werft with-preview


andreafalzetti (approved):
The output is now much easier to read IMO, thanks a lot! 🙌 

+1 also on your point about removing it 💯

PR #13613, opened 2022-10-05
[server] Fix network connections not limited

Furisto:
## Description
The result is not awaited and so the check if the user is on the  free tier is performed against the promise and not against the boolean which makes the returned result always false.

## Related Issue(s)
n.a.

## Release Notes

```release-note
None
```

## Werft options:

- [ ] /werft with-preview


geropl (approved):
:see_no_evil:  Thx for fixing @Furisto ! :pray:

kylos101:
Nice! Silly async :wink: !

PR #13616, opened 2022-10-05
[db] Change the type of the primary key in the `d_b_oauth_auth_code_entry` table

andrew-farries:
## Description

In order to de-risk the process of swapping the primary key in the `d_b_oauth_auth_code_entry` table, this PR stops the primary key from auto-incrementing and instead sets its value manually when persisting new rows.

We want to switch the primary key to one that is suitable for `db-sync` to be able to sync the table between regions.

See [this comment](https://github.com/gitpod-io/gitpod/pull/13583#issuecomment-1268320895) for more context.

## Related Issue(s)

Part of #9198 

## How to test

1. Run the local-companion app `local-app` against the preview environment:
```bash
GITPOD_HOST=<preview url> ./local-app test
```
2. Run through the auth flow.
3. See that the new rows in the `d_b_oauth_auth_code_entry` table have their `id` column set and that the type of the column is now a `char(36)` and no longer auto-increments.

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-db/src/typeorm/auth-code-repository-db.ts
+++ b/components/gitpod-db/src/typeorm/auth-code-repository-db.ts
@@ -57,6 +57,7 @@ export class AuthCodeRepositoryDB implements OAuthAuthCodeRepository {
     }
     public async persist(authCode: DBOAuthAuthCodeEntry): Promise<void> {
         const authCodeRepo = await this.getOauthAuthCodeRepo();
+        authCode.id = uuidv4();
         authCode.uid = uuidv4();
         authCodeRepo.save(authCode);
     }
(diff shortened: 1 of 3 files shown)


easyCZ (commented):
Looks sensible, but I've no idea how this will actually apply on cases where there are existing data in the table. Holding if you want a review from someone with SQL experience.

/hold

easyCZ (approved):

andrew-farries:
> Looks sensible, but I've no idea how this will actually apply on cases where there are existing data in the table. Holding if you want a review from someone with SQL experience.

With a small test dataset the change of column type worked without issue.

andrew-farries:
@geropl regarding [this comment](https://github.com/gitpod-io/gitpod/pull/13583#issuecomment-1268320895), I don't think any further 'primary key switching' is necessary beyond this PR right?

We now (in this PR) have a non-autoincrementing primary key column which is set to a uid like most of our other PKs. 

The only issue is that the existing rows in the table will have numerical values for their PK, and any new rows will have uids. Assuming we are ok with that, we can simply drop the `uid` column now - we don't need to switch over to it.

andrew-farries:
Added comments as suggested by @geropl in a sync call this morning.

/unhold

PR #13620, opened 2022-10-05
[code] sync profiles/collections

jeanp413:
## Description

Adds sync profile support to code-sync server

## How to test

1. Open a workspace using latest vscode in preview env https://jp-sync-collections.preview.gitpod-dev.com/workspaces
2. Update settings `"workbench.experimental.settingsProfiles.enabled": true`
3. Play around with [profiles](https://code.visualstudio.com/updates/v1_72#_settings-profiles), create, update, delete profiles and resources (settings, task, snippets) etc
4. open developer tools and check code-sync endpoint request return sucessfully
5. You can play in insider.vscode.dev and compare request responses
6. Check database migration works just fine
    - Use any other prev env and play with settings sync to add some data to the database
    - Connect to database and run DB migration manually, check it runs successfully

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


akosyakov:
Please clean up commit history and update release notes.

components/gitpod-db/src/typeorm/entity/db-code-sync-collection.ts
```diff
@@ -0,0 +1,18 @@
...
+
+@Entity()
+// on DB but not Typeorm: @Index("ind_dbsync", ["_lastModified"])   // DBSync
+export class DBCodeSyncCollection {
+    @PrimaryColumn(TypeORM.UUID_COLUMN_TYPE)
+    userId: string;
+
+    @PrimaryColumn(TypeORM.UUID_COLUMN_TYPE)
+    collection: string;
+}
```
easyCZ, line 18: I'd recommend adding a deleted column already, to allow for db-sync to delete records.

easyCZ: How will you manage size of this table? In other words, how will you identify what to clean up? In any case, it may be useful to include a CreatedTimestamp  on the table to help you track when records came into existence - helps with debugging.

components/gitpod-db/src/tables.ts
```diff
@@ -204,11 +204,17 @@ export class GitpodTableDescriptionProvider implements TableDescriptionProvider
             deletionColumn: "deleted",
             timeColumn: "_lastModified",
         },
+        {
+            name: "d_b_code_sync_collection",
+            primaryKeys: ["userId", "collection"],
```
easyCZ, line 209: If we added a UUID for an entry in the collection, you can then reference it as a single key in the `d_b_code_sync_resource` table rather than duplicating the `userId` and `collection` properties

jeanp413: `userId` was already in `d_b_code_sync_resource` so it would be the same to use this new uuid column or the collection column :thinking:

components/gitpod-db/src/typeorm/entity/db-code-sync-collection.ts
jeanp413: > I'd recommend adding a deleted column already, to allow for db-sync to delete records.

:+1: I'll add it, I wasn't sure as I saw other tables don't have it either in `tables.ts`, should all tables that support deleting and are synced need this column?

> it may be useful to include a CreatedTimestamp on the table to help you track when records came into existence

There's already an `_lastModified` column, it's just not declared in the ts file as I don't use it explicitly, should I add it either way?

mustard-mh: > should all delete operations need to set this value to true, meaning I cannot use .delete() method from typeorm?

Yes, and don't forget to filter `deleted=false` in `select`

> There's already an _lastModified column in the migration file, it's just not declared in the ts file as I don't use it explicitly, should I add it either way?

We need to define `_lastModified` while creating the table. but not necessary to have it in entity 

(This is what I did in db changed PR 
https://github.com/gitpod-io/gitpod/pull/10573/commits/f23c2bb8cddf5384424ebee6d416a7554ee33eed#diff-9ead78f16102163ca590f18a006f2c89ba38fcd9d929362687ee36c6f94acc1f )

jeanp413: @mustard-mh so this is a bug in [this line](https://github.com/gitpod-io/gitpod/blob/801de49c1fc521a4c206f41511ddb6cd29c99598/components/gitpod-db/src/typeorm/code-sync-resource-db.ts#L86) added in [this PR](https://github.com/gitpod-io/gitpod/commit/4f1c3922b4ad71b6601382068b8a33d9b7cb2f2b) :smile:

mustard-mh: > @mustard-mh so this is a bug in [this line](https://github.com/gitpod-io/gitpod/blob/801de49c1fc521a4c206f41511ddb6cd29c99598/components/gitpod-db/src/typeorm/code-sync-resource-db.ts#L86) added in [this PR](https://github.com/gitpod-io/gitpod/commit/4f1c3922b4ad71b6601382068b8a33d9b7cb2f2b) 😄

😇

components/gitpod-db/src/tables.ts
mustard-mh: @jeanp413  Where is the name of `collection` stored? No need?

components/gitpod-db/src/typeorm/entity/db-code-sync-collection.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Copyright (c) 2020 Gitpod GmbH. All rights reserved.
```
mustard-mh, line 2:
```suggestion
 * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
```

components/server/src/code-sync/code-sync-service.ts
```diff
@@ -410,12 +571,34 @@ export class CodeSyncService {
...
+    private async deleteCollection(userId: string, collection: string | undefined) {
+        try {
+            await this.db.deleteCollection(userId, collection, async (collection?: string) => {
+                try {
+                    const request = new DeleteRequest();
+                    request.setOwnerId(userId);
+                    request.setPrefix(toObjectName(SyncResource.GlobalState, "all", collection ?? "all"));
+                    const blobsClient = this.blobsProvider.getDefault();
+                    await util.promisify<DeleteRequest, DeleteResponse>(blobsClient.delete.bind(blobsClient))(request);
+                } catch (e) {
```
mustard-mh, line 592: We have duplicate code like this, why not make them inside this function? And it miss delete GCP (stored) resources too

components/gitpod-db/src/tables.ts
jeanp413: profile metadata is stored in `profile` server resource, not in the collection

mustard-mh:
It works for me.

But after starting another workspace synced user setting does not include `workbench.experimental.settingsProfiles.enabled`, which means I need to enable it every time I open a workspace 

(was it limited by VSCode? If we want to land this experimental feature, we may need to sync this setting too).

|Works 🎉 |
|-|
|Oct-13-2022 02-35-28|

jeanp413:
@easyCZ can you review db changes again :pray:

easyCZ (approved):
Looks reasonable. I don't have context on the collection domain to know if the DB schema is appropriate but code-wise this looks good.

@andrew-farries Wanna give this a look as well just to check it aligns with your DB sync work?
/hold

mustard-mh:
Will we merge this PR?

jeanp413:
yeah let me find some free time.
I want us to be ready/free if something goes wrong, though db change is just inserting a new column, better safe than sorry :slightly_smiling_face:

mads-hartmann:
@geropl From your screenshot it says "This branch cannot be rebased due to conflcits"? Was the PR in a state where it wasn't mergeable? Tide will try to merge anything that fits its query, regardless of whether or not the PR is mergable or not

easyCZ:
@mads-hartmann when I looked at the PR, Github did not indicate there would be a merge conflict. I did not investigate the build or anything else but the general signals I'd consider for this to be "mergeable" were satisfied.

mads-hartmann:
/unhold

Experiment to see what errors Tide gives up if any when trying to merge this.

mads-hartmann:
/hold

The PR is mergable, Tide tries to merge it. It doesn't say it errored but it keeps trying. I'm holding while I look at the logs ☺️
Screenshot 2022-11-10 at 11 04 59

mads-hartmann:
/unhold

From the Tide logs we see "PR is unmergable. Do the Tide merge requirements match the GitHub settings for the repo? This branch can't be rebased" but that's not what I saw in the UI just before.

Will try to merge the PR manually, see internal discussion [here](https://gitpod.slack.com/archives/C01KGM9EBD4/p1668075545157469?thread_ts=1668068143.677889&cid=C01KGM9EBD4)

PR #13627, opened 2022-10-05
[werft] Implement log streaming for subprocesses

adrienthebo:
## Description

This is a quick experiment to stream logs from subprocesses of a werft job.

At present the current subprocess exec function waits for a process to complete before emitting logs; this approach is viable for shorter tasks but Terraform runs can extend for 20+ minutes. In addition long waiting periods frequently mean that a job is in an error state, but there's no way for us to view the logs incrementally and halt a misbehaving job.

This commit alleviates this issue by adding a new async `execStream` function that will stream logs from the subprocess. When passed a slice logs will be streamed to werft; otherwise the subprocess stdout/stderr will be attached to the werft stdout/stderr.

> **note**: my experience with Typescript is limited and this is a "throw some code at the wall and see what sticks" type pull request. I expect that there's a number of stylistic improvements that could be made and invite vigorous criticism by someone that's written more than my hour or so of typescript. 😁

This may have an added benefit of preventing `werft job logs` from timing out. It appears that this command fails frequently when a job hangs with no output for extended periods of time; by eliminating the pauses found in our Terraform job logs this pain should be reduced.

<details>
<summary>Example werft job logs timeout</summary>

```
Error: rpc error: code = Internal desc = stream terminated by RST_STREAM with error code: INTERNAL_ERROR
tip: There seems to be a problem with your werft installation - please get in contact with whoever is operating this installation.
```
</details>

## Related Issue(s)

Fixes #

## How to test

```
werft run github -j k3s-installer-tests.yaml --follow -a domain=tests.doptig.com
```

[gitpod-custom-alt-stream-exec-logs.20](https://werft.gitpod-dev.com/job/gitpod-custom-alt-stream-exec-logs.20)

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/.werft/installer-tests.ts
+++ b/.werft/installer-tests.ts
@@ -1,7 +1,7 @@
 import * as fs from "fs";
 import * as https from "https";
 import { join } from "path";
-import { exec } from "./util/shell";
+import { exec, execStream } from "./util/shell";
 import { Werft } from "./util/werft";
 import { deleteReplicatedLicense } from "./jobs/build/self-hosted-upgrade-tests";
 
@@ -282,11 +282,15 @@ if (config === undefined) {
 }
 
 installerTests(TEST_CONFIGURATIONS[testConfig]).catch((err) => {
-    if(deleteOnFail == "true") {
-        cleanup();
-    }
     console.error(err);
-    process.exit(1);
+
+    if(deleteOnFail == "true") {
+        cleanup().finally(() => {
+            process.exit(1);
+        });
+    } else {
+        process.exit(1);
+    }
 });
 
 function logJobConfig(): void {
@@ -332,7 +336,7 @@ export async function installerTests(config: TestConfig) {
     werft.phase(majorPhase, phaseMessage);
     for (let phase of config.PHASES) {
         const phaseSteps = INFRA_PHASES[phase];
-        const ret = callMakeTargets(phaseSteps.phase, phaseSteps.description, phaseSteps.makeTarget);
+        const ret = await callMakeTargets(phaseSteps.phase, phaseSteps.description, phaseSteps.makeTarget);
         if (ret) {
             // there is not point in continuing if one stage fails for infra setup
             const err: Error = new Error("Cluster creation failed");
@@ -359,7 +363,7 @@ export async function installerTests(config: TestConfig) {
         // runIntegrationTests()
 
         const upgradePhase = INFRA_PHASES["KOTS_UPGRADE"];
-        const ret = callMakeTargets(upgradePhase.phase, upgradePhase.description, upgradePhase.makeTarget);
+        const ret = await callMakeTargets(upgradePhase.phase, upgradePhase.description, upgradePhase.makeTarget);
         if (ret) {
             sendFailureSlackAlert(
                 upgradePhase.description,
@@ -375,7 +379,7 @@ export async function installerTests(config: TestConfig) {
     if (skipTests === "true") {
         console.log("Skipping integration tests");
     } else {
-        runIntegrationTests();
+        await runIntegrationTests();
     }
 
     // if the preview flag is set to true, the script will print the result and exits
@@ -413,11 +417,11 @@ export async function installerTests(config: TestConfig) {
         werft.done("print-output");
     } else {
         // if we are not doing preview, we delete the infrastructure
-        cleanup();
+        await cleanup();
     }
 }
 
-function runIntegrationTests() {
+async function runIntegrationTests() {
     werft.phase(`run-${testSuite}-integration-tests`, `Run all ${testSuite} integration tests`);
 
     const componentTests = TestMap[testSuite.toLowerCase()]
@@ -431,7 +435,7 @@ function runIntegrationTests() {
     console.log(`Running ${testSuite} tests`)
     for (let test in componentTests) {
         const testPhase = componentTests[test];
-        const ret = callMakeTargets(testPhase.phase, testPhase.description, testPhase.makeTarget);
+        const ret = await callMakeTargets(testPhase.phase, testPhase.description, testPhase.makeTarget);
         if (ret) {
             exec(
                 `werft log result -d "failed test" url "${testPhase.description}(Phase ${testPhase.phase}) failed. Please refer logs."`,
@@ -449,33 +453,38 @@ function runIntegrationTests() {
     werft.done("run-integration-tests");
 }
 
-function callMakeTargets(phase: string, description: string, makeTarget: string, failable: boolean = false) {
-    werft.log(phase, `Calling ${makeTarget}`);
+/**
+ * Run a make target within a werft slice.
+ *
+ * @return The make target return code
+ */
+async function callMakeTargets(slice: string, description: string, makeTarget: string, failable: boolean = false): Promise<number> {
+    werft.log(slice, `Calling ${makeTarget}`);
     // exporting cloud env var is important for the make targets
     const env = `export TF_VAR_cluster_version=${k8s_version} cloud=${cloud} TF_VAR_domain=${baseDomain} TF_VAR_gcp_zone=${gcpDnsZone}`;
 
-    const response = exec(
+    const code = await execStream(
         `${env} && make -C ${makefilePath} ${makeTarget}`,
         {
-            slice: phase,
+            slice: slice,
             dontCheckRc: true,
         },
     );
 
-    if (response.code) {
-        console.error(`Error: ${response.stderr}`);
+    if (code !== 0) {
+        console.error(`Error: make target ${makeTarget} exited with code ${code}`);
 
         if (failable) {
-            werft.fail(phase, "Operation failed");
-            return response.code;
+            werft.fail(slice, "Operation failed");
+            return code;
         }
-        werft.log(phase, `'${description}' failed`);
+        werft.log(slice, `'${description}' failed`);
     } else {
-        werft.log(phase, `'${description}' succeeded`);
-        werft.done(phase);
+        werft.log(slice, `'${description}' succeeded`);
+        werft.done(slice);
     }
 
-    return response.code;
+    return code;
 }
 
 function randomize(options: string[]): string {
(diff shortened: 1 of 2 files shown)


roboquat:
@adrienthebo: Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://git.k8s.io/community/contributors/guide/release-notes.md) to remove it.

<details>

</details>

.werft/installer-tests.ts
```diff
@@ -387,12 +387,12 @@ function runIntegrationTests() {
     werft.done("run-integration-tests");
 }
 
-function callMakeTargets(phase: string, description: string, makeTarget: string, failable: boolean = false) {
+async function callMakeTargets(phase: string, description: string, makeTarget: string, failable: boolean = false) {
```
nandajavarma, line 390: Can we add an `await` in the [call to this fn ](https://github.com/gitpod-io/gitpod/pull/13627/files#diff-073c7685dea478fdf0c74dd43a01d49232bd3b941d8dac00658e9a885bbfb1e8R471) in `cleanup` function. Currently since it is being called async, we are getting weird behavior https://werft.gitpod-dev.com/job/gitpod-custom-nvn-test-ad-stream.1. The call to list the remaining state gets triggered which is not empty since deletion is still going on. And this sends quite a lot of alerts to slack.

nandajavarma:
This is extremely cool!! :star_struck:

.werft/installer-tests.ts
```diff
@@ -364,12 +364,12 @@ export async function installerTests(config: TestConfig) {
...
 }
 
-function runIntegrationTests() {
+async function runIntegrationTests() {
     werft.phase("run-integration-tests", "Run all existing integration tests");
     const slackAlerts = new Map<string, string>([]);
     for (let test in TESTS) {
         const testPhase = TESTS[test];
-        const ret = callMakeTargets(testPhase.phase, testPhase.description, testPhase.makeTarget);
+        const ret = await callMakeTargets(testPhase.phase, testPhase.description, testPhase.makeTarget);
```
nandajavarma, line 372: How about we remove the `await` here so multiple tests can run in parallel?

adrienthebo: I'd like to! We can't do that _yet_; test configurations run jobs in serial and depend on that for ordering, such as the ordering between the cert-manager helm chart and the cluster issuer. In the future we can change test definitions to take `Array<Phase>` to define steps that can be invoked in parallel.

Since Terraform already creates resources in parallel we probably won't see a significant speed boost from parallelism but this is an area we should generally explore.

vulkoingim:
Nice one! LGTM, but I'm in the same boat as you are with TS.... Let's wait for our resident typescript guru (Mads) to take a look at it :)

.werft/installer-tests.ts
```diff
@@ -268,10 +268,11 @@ if (config === undefined) {
 
 installerTests(TEST_CONFIGURATIONS[testConfig]).catch((err) => {
     if(deleteOnFail == "true") {
-        cleanup();
+        cleanup().finally(() => {
+            console.error(err);
+            process.exit(1);
+        });
     }
-    console.error(err);
```
nandajavarma, line 273: I feel like this should still be there. Even if we are not explicitly asking to fail, I believe if an error is quote it should respond with a process.exit (failed job)

nandajavarma: See this example https://werft.gitpod-dev.com/job/gitpod-custom-alt-stream-exec-logs.5

nandajavarma (changes requested):

.werft/util/shell.ts
```diff
@@ -62,6 +62,66 @@ export function exec(cmd: string, options?: ExecOptions): ChildProcess | shell.S
...
+
+    const child = shell.exec(command, {...options, async: true});
+
+    let stdout = '';
+    let stderr = '';
+
+    // note: the stdout/stderr event handlers aren't guaranteed to receive buffers that are always
+    // newline terminated. The original log messages can be preserved by finding the index of the
+    // last newline, printing up until that newline, buffering the remaining message, appending
+    // to that buffer on the next call, and finally flushing the buffers then the process exits.
```
mads-hartmann, line 88: I think we'd get nicer output in Werft if we did decide to go with the solution outlined in your comment. What made you decide not to? ☺️

.werft/util/shell.ts
```diff
@@ -62,6 +62,66 @@ export function exec(cmd: string, options?: ExecOptions): ChildProcess | shell.S
...
+    });
+
+    child.stderr.on('data', (data) => {
+        if (options.slice) werft.logOutput(options.slice, data.trim());
+        stderr += data;
+    });
+
+    const code = await new Promise<number>((resolve, reject) => {
+        child.on('close', (code) => {
+            if (typeof code === "number") {
```
mads-hartmann, line 101: I don't think we can separate these three if/else if/else conditions

If the exit code is non-zero I would expect it to `reject(code)` _unless_ `options.dontCheckRc` is set ☺️

.werft/util/shell.ts
```diff
@@ -62,6 +62,66 @@ export function exec(cmd: string, options?: ExecOptions): ChildProcess | shell.S
...
+
+    options = options || {};
+
+    if (options.slice) {
+        options.silent = true;
+    }
+
+    const child = shell.exec(command, {...options, async: true});
+
+    let stdout = '';
```
mads-hartmann, line 82: It's probably not bad, but for some commands I expect this could end up being a pretty big string. As we're not using the output anywhere at the moment anyway (as far as I can see), how about remove them for now? ☺️

.werft/installer-tests.ts
```diff
@@ -468,40 +474,41 @@ function randOsVersion(): string {
     return randomize(options);
 }
 
-function cleanup() {
+async function cleanup(): Promise<void> {
     const phase = INFRA_PHASES["DESTROY"];
     werft.phase(phase.phase, phase.description);
 
-    const ret = callMakeTargets(phase.phase, phase.description, phase.makeTarget);
+    return await callMakeTargets(phase.phase, phase.description, phase.makeTarget)
```
mads-hartmann, line 481: You don't need the `await` here as `callMakeTargets` is already a promise - it doesn't hurt either though so feel free to keep it

.werft/installer-tests.ts
mads-hartmann: I agree - but because cleanup is now async the old code won't work anymore though - so it could be something like the following

```ts
if(deleteOnFail == "true") {
    cleanup().finally(() => {
        console.error(err);
        process.exit(1);
    });
} else {
    console.error(err);
    process.exit(1);
}
```

mads-hartmann (commented):
This is neat, thanks @adrienthebo - I left a few comments but streaming the output sounds like an excellent improvement.

.werft/util/shell.ts
adrienthebo: Yep... this started out as a proof of concept and I've been glossing over the problem of potential memory exhaustion.

> As we're not using the output anywhere at the moment anyway (as far as I can see)

We are, but we should be able to remove it. [[1](https://github.com/gitpod-io/gitpod/blob/main/.werft/installer-tests.ts#L408)] 

<details>
<summary>Recorded for posterity: child_process buffering</summary>

Moving this to a collapsible block as it's not immediately pertinent.

Long story short, node's child_process.js is quite messy, and shelljs tacks on another layer of abstraction that makes it very difficult to reason about output buffering.

I just spent the last hour+ looking at [shelljs.exec](https://github.com/shelljs/shelljs/blob/v0.8.5/src/exec.js#L127), [node's child_process.js](https://github.com/nodejs/node/blob/bdf069ba0f6ee63056628af20d95cd82ca5ea736/lib/child_process.js#L319), and the [node child_process internals](https://github.com/nodejs/node/blob/0633e9a0b5dcbc6d3c250c9086e9386c5d5261b3/lib/internal/child_process.js), and to be honest I have _no idea_ how stdout/stderr is actually routed, when certain callbacks fire and in what order they can be guaranteed to fire, and how stderr/stdout buffering behaves when `child_process.exec` called by shelljs. With some calls to `exec` there's a [max buffer](https://nodejs.org/api/child_process.html#maxbuffer-and-unicode) applied, but I cannot figure out the context in which this is applied.

Looking at the node child_process internals has toasted my brain so I'll rephrase:

1. There's a possibility that very large stdout/stderr buffers could cause memory exhaustion;
2. It's possible that node's internals will catch this buffer exhaustion and terminate the process;
3. I can't tell if the buffer limiting will work as desired in this context;
4. There's likely a better way to collect, buffer, and return stderr/stdout than what I'm doing (and I'm almost certainly duplicating work);
5. Determining how it works will consume significantly more time than I can allocate;
6. This alleviates some major pain points in the installer integration tests and I would like to keep this moving forward so we can reap the benefits now.

</details>

I'll change it so that we simply return the exit code, will add in some comments leaving breadcrumbs, and will kick this can down the road.

.werft/util/shell.ts
adrienthebo: In short: doing it right is a bit complex, and I'm really low on time right now.

In long: The self-hosted test and preview environments run a lot of very long, very slow commands, so it can be upwards of 25 minutes to determine if a single step failed. Earlier on we were having issues with EKS tests where nodes would fail to attach - but the failure would happen 5 minutes in but you'd only find out after Terraform timed out after 30 minutes. The absence of streamed output is a major drain on velocity.

With behavior in this PR, you end up with slightly weird command output that's sprinkled with a few newlines; it's a minor irritation but pretty tolerable. Fixing this requires a bit of time and patience; it's doable but it's not critical. Since the self-hosted installer tests are the only systems that are impacted by this change we're not going to mangle the test output for anyone that won't expect it. I'd like to ship this as a skateboard, immediately reap the benefits of the improvements, and continue iterating on command execution quality of life fixups.

To emphasize, this is a _massive_ quality of life improvement for the self hosted tests. I don't like merging code that has known quirks like this but I'm painfully low on time right now. I'd like to spend a goodwill token on this if at all possible; 😇 incorporating this fix will really improve velocity in the pipelines and represents a small, well defined amount of technical debt.

.werft/installer-tests.ts
adrienthebo: Resolved in b84a84dfc3389e4a2d5c03e4c960a4e4ee22f65a

.werft/util/shell.ts
adrienthebo: Good catch, thanks! Also, I misread the docs; `on('close')` passes both the [exit code and signal](https://github.com/nodejs/node/blob/0633e9a0b5dcbc6d3c250c9086e9386c5d5261b3/lib/internal/child_process.js#L1098) as separate values, which simplifies this quite a bit.

Issue #13177, opened 2022-09-21
Store Stripe Customer ID internally, and make StripeService.findCustomerByAttributionId use the stored ID when available


jankeromnes:
## Is your feature request related to a problem? Please describe

This is a follow-up to https://github.com/gitpod-io/gitpod/pull/13002, but the bug was actually pre-existing:

- Sometimes (maybe around 1 in 10 times?) when you subscribe a team to Stripe, creating the subscription randomly fails:

<img width="600" alt="Screenshot 2022-09-21 at 16 21 40" src="https://user-images.githubusercontent.com/599268/191530013-fba72f30-bfc4-4706-b8bb-57ec9146d3ee.png">

- And in the backend, you'll see this error:

```
Error: No Stripe customer profile for 'team:123ff8cd-422f-4865-87eb-2d3767b2b16f'
```

- But if you try adding the same credit card / country / currency again, it works

This is because searching Stripe customers by `attributionId` seems a bit inconsistent right after customer creation (we poll Stripe Search to confirm that a new customer is available in search results, but then when running the same query a few seconds later, it might come up empty again).

## Describe the behaviour you'd like

Instead of polling Stripe Search like we do here:

https://github.com/gitpod-io/gitpod/blob/759e588e450f37131dd20fe46855cf77e01815be/components/server/ee/src/user/stripe-service.ts#L94-L101

we should:

1. Store the created Customer ID in the team/user or their cost center (but it should not be widely accessible, because querying via `attributionId` is preferred)
2. When trying to find a customer by `attributionId`, StripeService.findCustomerByAttributionId should check for a stored ID first, and fall back to search of there is no ID, or the customer for that ID was deleted or its metadata were modified

## Describe alternatives you've considered

- Accept that subscribing sometimes fails and needs to be retried once? 😬
- Wait for a few more seconds after polling Stripe Search? 😬

easyCZ:
Linking further context from [internal discussion](https://gitpod.slack.com/archives/C02EN94AEPL/p1664261553210779)

easyCZ:
Occurrence of this raised internally during Workspace Start - [details](https://gitpod.slack.com/archives/C02EN94AEPL/p1664978399985299)

easyCZ:
An initial draft of this is [here](https://github.com/gitpod-io/gitpod/pull/13622)

[...]

PR #13645, opened 2022-10-06
[server] Report Stripe client metrics

easyCZ:
## Description

Reports prometheus metrics on Stripe interaction.

## Related Issue(s)

* Part of https://github.com/gitpod-io/gitpod/issues/13157
* Needed to get data for https://github.com/gitpod-io/gitpod/issues/13177 when migrating

## How to test

Creating a UBP subscription works

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/server/ee/src/user/stripe-service.ts
+++ b/components/server/ee/src/user/stripe-service.ts
@@ -9,6 +9,10 @@ import Stripe from "stripe";
 import { Config } from "../../../src/config";
 import { log } from "@gitpod/gitpod-protocol/lib/util/logging";
 import { AttributionId } from "@gitpod/gitpod-protocol/lib/attribution";
+import {
+    observeStripeClientRequestsCompleted,
+    stripeClientRequestsCompletedDurationSeconds,
+} from "../../../src/prometheus-metrics";
 
 const POLL_CREATED_CUSTOMER_INTERVAL_MS = 1000;
 const POLL_CREATED_CUSTOMER_MAX_ATTEMPTS = 30;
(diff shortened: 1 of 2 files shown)


components/server/ee/src/user/stripe-service.ts
```diff
@@ -149,11 +180,27 @@ export class StripeService {
...
+                customer: customer.id,
+                items: [{ price: priceId }],
+                automatic_tax: { enabled: isAutomaticTaxSupported },
+                billing_cycle_anchor: Math.round(startOfNextMonth.getTime() / 1000),
+            });
         });
     }
 }
+
+async function reportStripeOutcome<T>(op: string, f: () => T) {
```
geropl, line 195:
```suggestion
async function reportStripeOutcome<T>(op: string, f: () => Promise<T>): T {
```

easyCZ: What's the difference? T can be either, no?

easyCZ: I guess it forces the return to be a promise

geropl: :+1: It avoids copy'n'paste errors where you insert an `await`, or forget what you're supposed to return.

easyCZ: That makes sense, thanks for the tip!

components/server/ee/src/user/stripe-service.ts
```diff
@@ -149,11 +180,27 @@ export class StripeService {
...
     }
 }
+
+async function reportStripeOutcome<T>(op: string, f: () => Promise<T>) {
+    const timer = stripeClientRequestsCompletedDurationSeconds.startTimer();
+    let outcome = "ok";
+    try {
+        return await f();
+    } catch (e) {
+        outcome = e.type;
```
andrew-farries, line 201: e has `any` type here. How do we know that it has a `type` field?

andrew-farries (approved):
/hold for question

components/server/ee/src/user/stripe-service.ts
geropl: We don't. But you're right, outcome should be `string | undefined`, and ` observeStripeClientRequestsCompleted` might need minimal extension to translate `undefined` to `"undefined"` (or whatever we want to call it).

easyCZ: Interesting, I'd have hoped TS has my back here but looks like no. 

I mostly followed https://stripe.com/docs/api/errors/handling?lang=node

Will add a guard

easyCZ: Added.

geropl (approved):
Tested and worked. :ballot_box_with_check:

PR #13649, opened 2022-10-06
[db] Reset primary keys in `d_b_oauth_auth_code_entry`

andrew-farries:
## Description

Reset the primary keys in the `d_b_oauth_auth_code_entry` table to random uuids.

The type of the primary key column was changed in https://github.com/gitpod-io/gitpod/pull/13616 to allow storing uids, but the field now contains a mix of integer values (for existing records) and uids (for new records). In order to be able to sync the table with `db-sync` we need to replace the integer values with uuids.

The value of the primary key for each record doesn't matter so long as it is unique across both EU and US databases, so resetting the primary key values of all rows will work fine.

## Related Issue(s)

Part of #9198 

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


geropl (approved):
Make sense, thank you! :pray:

PR #13652, opened 2022-10-06
Check if preview cert is ready before trying to create/destroy it

vulkoingim:
## Description

Check if preview cert is ready before trying to create/destroy it

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/.werft/util/certs.ts
+++ b/.werft/util/certs.ts
@@ -26,6 +26,11 @@ export class InstallCertificateParams {
 }
 
 export async function issueCertificate(werft: Werft, params: IssueCertificateParams, shellOpts: ExecOptions): Promise<boolean> {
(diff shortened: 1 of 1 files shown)


.werft/util/certs.ts
```diff
@@ -63,6 +68,15 @@ function isCertReady(certName: string): boolean {
     return rc == 0
 }
 
+function isCertReady(certName: string): boolean {
+    const rc = exec(
```
mads-hartmann, line 72: `rc` should probably be named `output` or something like that

mads-hartmann (approved):
LGTM - left a tiny comment, but we we can always change that later if we want

Issue #10581, opened 2022-06-10
Epic: JetBrains Ports information shown on terminals


loujaybee:
JetBrains are integrating a ports view into the terminals view [CWM-4821](https://youtrack.jetbrains.com/issue/CWM-4821/Offer-port-forwarding-for-the-built-in-terminal). 

This issue is to integrate with that view and expose port information. 

Related issues: 
- https://github.com/gitpod-io/website/issues/2958
- https://github.com/gitpod-io/gitpod/issues/14667
- https://github.com/gitpod-io/gitpod/issues/13761

Relates to: 
- https://github.com/gitpod-io/gitpod/issues/10579
- https://youtrack.jetbrains.com/issue/CWM-6196/Port-forwarding-for-Gitpod

felladrin:
[Auto-forward Task Terminals Ports on JetBrains IDEs #10986](https://github.com/gitpod-io/gitpod/pull/10986) got merged, but we need to keep track of this limitation:

- Forwarded ports on Terminals from Gitpod Tasks won't display the message "Your application is listening on ports" that you see when you start a server on a standard IntelliJ Terminal Widget.

| Open port on Gitpod Task Terminal | Open port on standard IntelliJ Terminal |
| ----------------------------------- | ------------------------------------- |
| ![image](https://user-images.githubusercontent.com/418083/176254937-63ca0146-5a83-482a-85e7-bb9c2c90233f.png) | ![image](https://user-images.githubusercontent.com/418083/176254985-c25924e7-1674-4f09-8706-c1fa180f3297.png) |

JetBrains team will continue working on Port Forwarding implementation, and we might have a solution for it soon.

loujaybee:
Waiting on a build from JetBrains for this one.

andreafalzetti:
> [Auto-forward Task Terminals Ports on JetBrains IDEs #10986](https://github.com/gitpod-io/gitpod/pull/10986) got merged, but we need to keep track of this limitation:
on for it soon.

I've captured the work mentioned here in this ticket: https://github.com/gitpod-io/gitpod/issues/14667

loujaybee:
**Current status:** Running processes within the terminal window shows ports information, but we cannot yet automate this process. @andreafalzetti / @felladrin to collate any YouTrack or related issues so that we can share also with JetBrains, and push for some more updates so that we can make progress on this.

loujaybee:
Still waiting to compile links for JetBrains on YouTrack.

[...]

PR #13653, opened 2022-10-06
Adjust JetBrains Backend Plugin to work with the new Port Forwarding API

felladrin:
## Description

Now that we have EAP IDEs build 223.6x available in gitpod.io, we need to update the back-end plugin with the support for the new Port Forwarding API.

In the current state of this PR, the plugin compiles and doesn't throw errors, but before setting it as Ready for Review, here's what we need to accomplish:
- [ ] Ensure the usage of port name/description columns is working.
- [ ] Remove the unused function overrides from `object: ForwardedPortListener`
- [ ] Confirm there's no other way to dispose the forwarded ports listeners (currently we use a map for it).
- [ ] Confirm if ports are removed from the UI when programatically removed.

## Related Issue(s)

- https://github.com/gitpod-io/gitpod/issues/10581

## How to test

1. Open the preview environment of this PR: https://felladrin-a8c1911845.preview.gitpod-dev.com/
2. In the account preferences, select the _Latest_ JetBrains IntelliJ IDEA as the editor
3. (...)

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-large-vm

### DIFF
--- a/components/ide/jetbrains/backend-plugin/gradle-latest.properties
+++ b/components/ide/jetbrains/backend-plugin/gradle-latest.properties
@@ -6,4 +6,4 @@ pluginUntilBuild=223.*
 # See https://jb.gg/intellij-platform-builds-list for available build versions.
 pluginVerifierIdeVersions=2022.3
 # Version from "com.jetbrains.intellij.idea" which can be found at https://www.jetbrains.com/intellij-repository/snapshots
-platformVersion=223.5756-EAP-CANDIDATE-SNAPSHOT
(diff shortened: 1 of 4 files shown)


felladrin:
Closed in favor of https://github.com/gitpod-io/gitpod/pull/13747, due to an issue with old Preview Environments [[1](https://gitpod.slack.com/archives/C032A46PWR0/p1665414537023909)].

PR #13659, opened 2022-10-06
[usage] Add GetStripeCustomer to BillingService

easyCZ:
## Description

This will allow us to move the lookup of customers to a single place and improve our ability to respond to it consistently.

## Related Issue(s)

* Part of https://github.com/gitpod-io/gitpod/issues/13177

## How to test

builds

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/usage-api/usage/v1/billing.proto
+++ b/components/usage-api/usage/v1/billing.proto
@@ -16,6 +16,10 @@ service BillingService {
   // CancelSubscription cancels a stripe subscription in our system
   // Called by a stripe webhook
   rpc CancelSubscription(CancelSubscriptionRequest) returns (CancelSubscriptionResponse) {};
+
+  // GetStripeCustomer retrieves a Stripe Customer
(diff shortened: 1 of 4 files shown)


components/usage-api/usage/v1/billing.proto
```diff
@@ -35,3 +39,19 @@ message CancelSubscriptionRequest {
...
+}
+
+message GetStripeCustomerResponse {
+  StripeCustomer customer = 1;
+  string attribution_id = 2;
+}
+
+message StripeCustomer {
+  string id = 1;
+}
```
andrew-farries, line 57: Is it worth inlining this into the `GetStripeCustomerResponse`?

andrew-farries (approved):
✅ just one question

components/usage-api/usage/v1/billing.proto
easyCZ: I'm keen for this to be extended in the future more fields, so would like to keep it as a dedicated object.

PR #13692, opened 2022-10-08
[public-api] Refactor to use connect handlers

easyCZ:
## Description

In this PR, we 1-1 map the old implementation in gRPC to the new implementation with Connect. Changes are deliberately not made to streamline the tests and simplify to keep the PR cleaner.

In general, the changes are the following (see [docs](https://connect.build/docs/go/grpc-compatibility#migration))
1. Use the generated ServiceHandlers
2. Wrap requests and responses in `connect.NewRequest` or `connect.NewResponse`
3. Use Connects error codes
4. Use HTTP server in tests

This PR also changes default routing of `api.<domain>` requests to the HTTP server, this has the following effect:
* Connect handlers handle both gRPC and HTTP traffic - this is fine, as Connect is designed for this
* Depending on the rollout sequencing, there can be a brief period where requests would fail. This is OK, as we don't get any traffic on this yet anyway.

## Related Issue(s)

* Depends on https://github.com/gitpod-io/gitpod/pull/13693

## How to test

Unit tests

1. Log in to preview and run the following in the console to get an access token:
```
	"function:getWorkspace",
	"function:getWorkspaces",
	"resource:default",]})
```
To make subsequent steps readily runnable, export the token in your shell:
```
export BEARER_TOKEN="<token>"
```

Validate we can reach the new APIs over cURL:
```
curl \
    --header 'Content-Type: application/json' \
    --header 'Authorization: Bearer $BEARER_TOKEN' \
    --data '{}' \
    https://api.mp-public-ca4417e0d2.preview.gitpod-dev.com/gitpod.v1.WorkspacesService/ListWorkspaces | jq

{}
```
We're doing a list call, and we don't have workspaces so an empty list response is expected.

Validate we can reach the APIs using a gRPC client
```
gpctl api workspaces list --address api.mp-public-ca4417e0d2.preview.gitpod-dev.com:443 --token $BEARER_TOKEN 

ID        Owner        ContextURL
```
Again, we're not getting any workspaces back because we've not started any but the important part is we're not getting an error.

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/public-api-server/pkg/server/integration_test.go
+++ b/components/public-api-server/pkg/server/integration_test.go
@@ -13,163 +13,111 @@ import (
 	"github.com/bufbuild/connect-go"
 	"github.com/gitpod-io/gitpod/common-go/baseserver"
 	"github.com/gitpod-io/gitpod/public-api-server/pkg/auth"
+	"github.com/gitpod-io/gitpod/public-api-server/pkg/proxy"
 	v1 "github.com/gitpod-io/gitpod/public-api/v1"
 	"github.com/gitpod-io/gitpod/public-api/v1/v1connect"
 	"github.com/stretchr/testify/require"
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/credentials/insecure"
-	"google.golang.org/grpc/metadata"
-	"google.golang.org/grpc/status"
 )
 
 func TestPublicAPIServer_v1_WorkspaceService(t *testing.T) {
-	ctx := metadata.AppendToOutgoingContext(context.Background(), "authorization", "some-token")
-	srv := baseserver.NewForTests(t,
-		baseserver.WithGRPC(baseserver.MustUseRandomLocalAddress(t)),
-	)
-
-	gitpodAPI, err := url.Parse("wss://main.preview.gitpod-dev.com/api/v1")
-	require.NoError(t, err)
-
-	require.NoError(t, register(srv, gitpodAPI))
-	baseserver.StartServerForTests(t, srv)
-
-	conn, err := grpc.Dial(srv.GRPCAddress(), grpc.WithTransportCredentials(insecure.NewCredentials()))
-	require.NoError(t, err)
-
-	workspaceClient := v1.NewWorkspacesServiceClient(conn)
-
-	_, err = workspaceClient.CreateAndStartWorkspace(ctx, &v1.CreateAndStartWorkspaceRequest{})
-	requireErrorStatusCode(t, codes.Unimplemented, err)
-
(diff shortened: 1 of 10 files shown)


components/public-api-server/pkg/server/integration_test.go
```diff
@@ -13,151 +13,101 @@ import (
...
-	_, err = listenToStatusStream.Recv()
-	requireErrorStatusCode(t, codes.Unimplemented, err)
-
-	listenToLogsStream, err := prebuildClient.ListenToPrebuildLogs(ctx, &v1.ListenToPrebuildLogsRequest{})
-	require.NoError(t, err)
-	_, err = listenToLogsStream.Recv()
-	requireErrorStatusCode(t, codes.Unimplemented, err)
-}
-
-func TestPublicAPIServer_WorkspaceServiceHandler(t *testing.T) {
```
easyCZ, line 104: Diff shows wierd. This test is removed as it's already tested in `TestPublicAPIServer_v1_WorkspaceService`

components/public-api-server/pkg/server/integration_test.go
```diff
@@ -13,151 +13,101 @@ import (
...
 }
 
-func requireErrorStatusCode(t *testing.T, expected codes.Code, err error) {
-	require.Error(t, err)
-	st, ok := status.FromError(err)
-	require.True(t, ok)
-	require.Equalf(t, expected, st.Code(), "expected: %s but got: %s", expected.String(), st.String())
+func requireErrorStatusCode(t *testing.T, expected connect.Code, err error) {
+	t.Helper()
+	if expected == 0 && err == nil {
```
easyCZ, line 105: Connect doesn't define the default Error Code value as OK, so we have to do this extra check..

dev/gpctl/cmd/public-api.go
```diff
@@ -51,7 +51,7 @@ func newPublicAPIConn() (*grpc.ClientConn, error) {
 	opts := []grpc.DialOption{
 		// attach token to requests to auth
 		grpc.WithUnaryInterceptor(func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
-			withAuth := metadata.AppendToOutgoingContext(ctx, "authorization", publicApiCmdOpts.token)
+			withAuth := metadata.AppendToOutgoingContext(ctx, "authorization", "Bearer "+publicApiCmdOpts.token)
```
easyCZ, line 54: Was not correctly adding the Bearer prefix, and we weren't also correctly checking it for it server side.

easyCZ:
Closing as werft got into an inconcistent state. Replacement PR is [here](https://github.com/gitpod-io/gitpod/pull/13736)

PR #13703, opened 2022-10-10
Provide pressure stall information for workspaces

Furisto:
## Description
Retrieves [pressure stall information](https://lwn.net/Articles/759781/) for workspaces. Followup to https://github.com/gitpod-io/gitpod/pull/13539 which retrieved PSI on node level. 

## Related Issue(s)
n.a.

## How to test
- Start workspace in preview environment
- `kubectl port-forward ds/ws-daemon 9500`
- `curl XGET localhost:9500/metrics`

## Release Notes

```release-note
Provide pressure stall information for workspaces
```

## Werft options:

- [x] /werft with-preview


components/common-go/cgroups/cgroup.go
```diff
@@ -85,3 +87,41 @@ func ReadFlatKeyedFile(path string) (map[string]uint64, error) {
...
+		}
+
+		if strings.HasPrefix(line, "some") {
+			psi.Some = total
+		}
+
+		if strings.HasPrefix(line, "full") {
+			psi.Full = total
+		}
+	}
```
easyCZ, line 119: You'll also want to call `scanner.Err()` after `scanner.Scan()` returns false. See [example](https://pkg.go.dev/bufio#example-Scanner-Lines)

components/common-go/cgroups/v2/io.go
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+// Licensed under the GNU Affero General Public License (AGPL).
+// See License-AGPL.txt in the project root for license information.
+
+package v2
```
easyCZ, line 5: I think you'll want to name it `cgroups_v2` because otherwise you'd import it as `v2.IO` which isn't very descriptive

kylos101:
/werft run with-integration-tests=workspace with-large-vm=true

kylos101:
Running tests again, the prior job failed setup.

kylos101:
@Furisto you might want to rebase with main and run another build? The integration tests are timing out after 2 hours.

Furisto:
Might be due to this: https://gitpod.slack.com/archives/C032A46PWR0/p1665414537023909

components/common-go/cgroups/cgroup.go
```diff
@@ -85,3 +87,41 @@ func ReadFlatKeyedFile(path string) (map[string]uint64, error) {
 
 	return kv, nil
 }
+
+func ReadPSIValue(path string) (PSI, error) {
+	file, err := os.Open(path)
+	if err != nil {
```
utam0k, line 93: If err doesn't mean that file doesn't exist, I recommend recording the error log.

components/ws-daemon/pkg/cgroup/plugin_psi.go
```diff
@@ -0,0 +1,98 @@
...
+				return
+			}
+		}
+	}()
+
+	return nil
+}
+
+func (p *PSIMetrics) scrape(cpu *cgroups.Cpu, memory *cgroups.Memory, io *cgroups.IO, instanceID string) {
+	if psi, err := cpu.PSI(); err == nil {
```
utam0k, line 84: Let us check `os.IsNotexist` and record the error log to fix future bugs.

components/common-go/cgroups/cgroup.go
```diff
@@ -85,3 +87,41 @@ func ReadFlatKeyedFile(path string) (map[string]uint64, error) {
 
 	return kv, nil
 }
+
+func ReadPSIValue(path string) (PSI, error) {
```
utam0k, line 91: May I ask you to add a test about this function?

Furisto:
/werft run with-integration-tests=workspace with-large-vm=true with-clean-slate-deployment=true

Furisto:
/werft run with-integration-tests=workspace with-large-vm=true with-clean-slate-deployment=true

Furisto:
/werft run with-integration-tests=workspace with-large-vm=true with-clean-slate-deployment=true

Furisto:
@easyCZ @utam0k PTAL

components/common-go/cgroups/cgroups_test.go
```diff
@@ -134,3 +134,56 @@ func TestReadSingleValue(t *testing.T) {
...
+		},
+	}
+
+	for _, s := range scenarios {
+		t.Run(s.name, func(t *testing.T) {
+			f, err := os.CreateTemp("", "cgroup_test*")
+			if err != nil {
+				t.Fatal(err)
+			}
+
```
jenting, line 176:
```suggestion
defer os.Remove(f.Name())
```

components/common-go/cgroups/cgroup.go
```diff
@@ -85,3 +87,45 @@ func ReadFlatKeyedFile(path string) (map[string]uint64, error) {
 
 	return kv, nil
 }
+
```
jenting, line 90: Could you add a comment that we only read the total value?

components/common-go/cgroups/v2/io.go
jenting: Why?
Shouldn't the other package import as `common-go/cgroups/v2` and calls either `v2.NewIOControllerWithMount` _or_ `v2.NewIOController`?

components/ws-daemon/pkg/cgroup/plugin_psi.go
```diff
@@ -0,0 +1,111 @@
...
+		p.cpu.WithLabelValues(p.nodeName, instanceID, "some").Set(float64(psi.Some))
+		p.cpu.WithLabelValues(p.nodeName, instanceID, "full").Set(float64(psi.Full))
+	} else {
+		if !os.IsNotExist(err) {
```
jenting, line 89:
```suggestion
	} else if !os.IsNotExist(err) {
```

components/common-go/cgroups/cgroup.go
```diff
@@ -85,3 +87,45 @@ func ReadFlatKeyedFile(path string) (map[string]uint64, error) {
...
+		}
+
+		if strings.HasPrefix(line, "some") {
+			psi.Some = total
+		}
+
+		if strings.HasPrefix(line, "full") {
+			psi.Full = total
+		}
+	}
```
easyCZ, line 123: You need to handle the scanner error
https://pkg.go.dev/bufio#example-Scanner-Lines

easyCZ: Nevermind, didn't see it already handled on 102. Thanks!

jenting (commented):
Other than the package naming, the rest LGTM

Furisto:
@jenting You mean you would like for it to be renamed back to v2 instead of cgroup_v2?

jenting:
> @jenting You mean you would like for it to be renamed back to v2 instead of cgroup_v2?

That's my preference, but @easyCZ has the [concern](https://github.com/gitpod-io/gitpod/pull/13703#discussion_r991216773). But I don't know why.

easyCZ (approved):

kylos101:
/werft run with-integration-tests=workspace with-large-vm=true

kylos101:
/hold

Please remove the hold once the tests are passing.

kylos101:
Tests are failing, please rebase this PR with main and try again :pray:

utam0k (approved):
Looks good

Furisto:
/werft run with-integration-tests=workspace with-large-vm=true

Furisto:
Had to wait for [this PR](https://github.com/gitpod-io/gitpod/pull/13890) to get merged.

kylos101:
Hey @Furisto , I was talking with @WVerlaek , and he mentioned a concern Re: collecting metrics at the workspace level as it relates to carnality. It could be expensive. What can we do to derisk this? 

Looping in @ArthurSens as a heads up, because I know this has been a concern in the past, too.

ArthurSens (changes requested):
Thanks for the ping Kyle!

Indeed, introducing metrics at the workspace level is complicated because of cardinality. It has a high chance that it will break our monitoring stack and, if not, it will certainly increase costs a lot.

I know what I'm requesting is a very big change, but would it be possible to follow team IDE approach and create a component that is responsible for collecting and exposing metrics of all workspaces at a single place?

Maybe it is even possible to re-use team IDE's component for it? cc @akosyakov

akosyakov:
Im out sick, but @iQQBot should be able to show what we are doing. It is possible that you can reuse IDE metrics endpoints. We could consider to move them behind the shared proxy if it is useful generally

kylos101:
:wave: @akosyakov thanks for the ping and looping in @iQQBot ! Can you share a picture or document that describes the approach you're using with supervisor?

@ArthurSens the new PSI metrics are available in our `ws-daemon` component (given the how to test nodes above). Perhaps that is enough to resolve your concern?

Here is how the metrics are defined:
https://github.com/gitpod-io/gitpod/pull/13703/files#diff-4eda78bbeca4334b0f918f771dac0fe377ceef23d3de98355935be8d01e990feR27-R40

ArthurSens:
> The idea behind this is not to aggregate it because it would allow us to investigate why a workspace exhibited a certain behavior.

Hmm I see, would it be possible to expose these metrics only for workspaces from paying customers? Thinking of self-hosted, maybe this metric could be turned on/off in the admin panel (Gitpod admins can choose teams that will have this metric exposed), so we can also choose our own customers there.

The problem with the current implementation is that every single workspace will introduce a new metric and we're definitely not analyzing PSI metrics for all of them, this means that we're wasting a huge amount of resources on something that we'll never use.

> I am struggling to understand why this would introduce so much additional load that it could break our metrics. Are we not already collecting a significant number of metrics for every pod on the cluster e.g. everything in [here](https://grafana.gitpod.io/d/6581e46e4e5c7ba40a07646395ef7b23/kubernetes-compute-resources-pod?orgId=1&refresh=10s&var-datasource=VictoriaMetrics&var-cluster=eu71&var-namespace=default&var-pod=ws-630dfa96-26e1-454a-8956-61f752cf820f).  In comparison to that, this PR does not introduce that much additional metrics from my point of view but I am happy to discuss this.

Indeed, compared to container metrics that are exposed by cAdvisor and Kubelet the costs are probably the same. This is the type of metric that we're really looking forward to removing 😅, they are indeed expensive.

The difference here is that we don't have much control over the kubelet nor cAdvisor metrics, while we do have control over our own metrics and we can be more conscious about them 🙂

Furisto:
> Hmm I see, would it be possible to expose these metrics only for workspaces from paying customers? Thinking of self-hosted, maybe this metric could be turned on/off in the admin panel (Gitpod admins can choose teams that will have this metric exposed), so we can also choose our own customers there.

Yes, that is certainly possible! I expect that this will be used more for paying customers anyway. 
Thank you for your explanation!

Furisto:
Metrics are now only retrieved for workspaces of paying users.

sagor999 (approved):

aledbf (approved):

atduarte:
Hey @Furisto! 👋 Removed release notes as this does not affect the Gitpod end user (developer).

utam0k:
> Hey @Furisto! wave Removed release notes as this does not affect the Gitpod end user (developer).

@atduarte 
You are right, but some users may be interested in that gitpod starting using PSI if it is on ChangeLog because PSI is lastest feature ;)  But it actually doesn't affect the user.

Furisto:
@atduarte I am actually unsure why I wrote a release note :laughing: As you said it does not affect the end user.

utam0k:
> @atduarte I am actually unsure why I wrote a release note laughing As you said it does not affect the end user.

I guess release notes are used creating monthly Changelog

PR #13706, opened 2022-10-10
[prebuild] Support opening a specfic prebuild

csweichel:
## Description
This PR enables users to open a specific prebuildm, and integrates this capability on the dashboard.

## How to test
1. Make sure there's a prebuild and a newer commit on the same branch
2. Go the prebuild view
3. Open the specific prebuild and observe it's the correct state

## Release Notes

```release-note
[prebuild] Add support for opening specific prebuilds
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/dashboard/src/projects/Prebuild.tsx
+++ b/components/dashboard/src/projects/Prebuild.tsx
@@ -181,7 +181,9 @@ export default function () {
                     ) : prebuild?.status === "available" ? (
                         <a
                             className="my-auto"
-                            href={gitpodHostUrl.withContext(`${prebuild?.info.changeUrl}`).toString()}
+                            href={gitpodHostUrl
+                                .withContext(`open-prebuild/${prebuild?.info.id}/${prebuild?.info.changeUrl}`)
+                                .toString()}
                         >
                             <button>New Workspace ({prebuild?.info.branch})</button>
                         </a>
--- a/components/server/src/workspace/open-prebuild-prefix-context-parser.ts
+++ b/components/server/src/workspace/open-prebuild-prefix-context-parser.ts
@@ -1,0 +1,37 @@
+/**
+ * Copyright (c) 2021 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { User, WorkspaceContext } from "@gitpod/gitpod-protocol";
+import { log } from "@gitpod/gitpod-protocol/lib/util/logging";
+import { OpenPrebuildContext } from "@gitpod/gitpod-protocol/src/protocol";
+import { inject, injectable } from "inversify";
+import { Config } from "../config";
+import { IPrefixContextParser } from "./context-parser";
+
+@injectable()
+export class OpenPrebuildPrefixContextParser implements IPrefixContextParser {
+    @inject(Config) protected readonly config: Config;
+    static PREFIX = /^\/?open-prebuild\/([^\/]*)\//;
+
+    findPrefix(user: User, context: string): string | undefined {
+        const result = OpenPrebuildPrefixContextParser.PREFIX.exec(context);
+        if (!result) {
+            return undefined;
+        }
+        return result[0];
+    }
+
+    public async handle(user: User, prefix: string, context: WorkspaceContext): Promise<WorkspaceContext> {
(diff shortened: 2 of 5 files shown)


components/server/src/workspace/open-prebuild-prefix-context-parser.ts
```diff
@@ -0,0 +1,37 @@
...
+import { log } from "@gitpod/gitpod-protocol/lib/util/logging";
+import { OpenPrebuildContext } from "@gitpod/gitpod-protocol/src/protocol";
+import { inject, injectable } from "inversify";
+import { Config } from "../config";
+import { IPrefixContextParser } from "./context-parser";
+
+@injectable()
+export class OpenPrebuildPrefixContextParser implements IPrefixContextParser {
+    @inject(Config) protected readonly config: Config;
+    static PREFIX = /^\/?open-prebuild\/([^\/]*)\//;
```
jankeromnes, line 17: Wow, many thanks @csweichel for implementing a new context URL prefix for this!

However, I'm curious why a new prefix is more beneficial than "just" pointing a Prebuild view's "New Workspace" button to the (already-supported) fully-qualified commit context URL -- if you open the exact commit that was used to create the prebuild, you're guaranteed to get this prebuild, right? 💭

jankeromnes:
Note: I'll give this PR a quick rebase, in the hopes of getting it to build. 🤞

jankeromnes:
Insert coin to try again 🎰

/werft run

jankeromnes:
And again

/werft run

components/server/src/workspace/open-prebuild-prefix-context-parser.ts
csweichel: > If you open the exact commit that was used to create the prebuild, you're guaranteed to get this prebuild, right?

Aye. Two reasons not to build it this way:
1. I did not see a clean way to construct the commit context URL for all SCM provider
2. It relies on the behaviour that we do actually open the prebuild for that commit given the correct context URL. Given the reliability of this feature I wanted to make this rock solid and identify the prebuild directly.

Another question would be: why use a prefix context parser rather than a regular context parser. This also has a couple of reasons:
- regular context parser are tied to the SCM system, i.e. assume they're part of the SCM inversify container. A `PrebuildContextParser` would break this assumption.
- permission checks can only happen in the server - if we wanted this to be a context parser, we'd need to make the permission checks part of that flow. Using a prefix context parser we can just rely on the permission checks already present in the SCM context parsing flow. This is an indicator to me that the current authorisation model has deficits we need to fix.

jankeromnes:
Indeed:

jankeromnes:
I guess that luck is permanently stuck. Maybe let's move this work to a different branch. Happy to do this now. 🚚

components/dashboard/src/projects/Prebuild.tsx
```diff
@@ -181,7 +181,9 @@ export default function () {
                     ) : prebuild?.status === "available" ? (
                         <a
                             className="my-auto"
-                            href={gitpodHostUrl.withContext(`${prebuild?.info.changeUrl}`).toString()}
+                            href={gitpodHostUrl
+                                .withContext(`open-prebuild/${prebuild?.info.id}/${prebuild?.info.changeUrl}`)
+                                .toString()}
                         >
                             <button>New Workspace ({prebuild?.info.branch})</button>
```
jankeromnes, line 188: I guess the `(branch)` part of the button label no longer makes sense when we're opening the prebuild.

Maybe let's remove it?

```suggestion
                            <button>New Workspace</button>
```

Or clarify it?

```suggestion
                            <button>New Workspace With This Prebuild</button>
```

jankeromnes:
Okay, I've tested this (over in https://github.com/gitpod-io/gitpod/pull/13768), but it doesn't see to do what I expected:

1. Added a repository as a new Project
2. Created a branch with a new "commit A" (and waited for the prebuild to be finished successfully)
3. Added a second "commit B" on top, in the same branch (also waited for successful prebuild)
4. When I navigate to "Prebuilds", I see two successful commits on my branch 👍
5. However, whichever I choose to open, I also seem to get the latest prebuild (with "commit B")

But, clicking on "New Workspace" in the prebuild of "commit A" should give me a workspace based on "commit A", right?

csweichel:
Let's move the conversation over to https://github.com/gitpod-io/gitpod/pull/13768. Will pick it up from there.

PR #13709, opened 2022-10-10
Delete cordoned node once it's empty during werft cleanup

vulkoingim:
## Description

We introduced a static build nodepool in https://github.com/gitpod-io/ops/pull/5281 but it stays cordoned off and the instance never gets terminated, since it cannot scale down. 

This PR waits until there are no more pods with werft jobs on the node (after it's cordoned) and deletes the GCE instance. This will force its recreation with a fresh local volume.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


mads-hartmann (approved):
Ah nice - good catch

PR #13716, opened 2022-10-10
[installer] Don't keep random trailing spaces when creating .golden files

jankeromnes:
## Description

Always output clean `.golden` files without random trailing spaces in them, by simply running a one-line post-processing script after the main generator.

Based on [this attempt](https://github.com/gitpod-io/gitpod/pull/11981#issuecomment-1210317909).

## Related Issue(s)

Fixes random trailing spaces in `.golden` files.

## How to test

1. Regenerate `.golden` files
2. They shouldn't contain random trailing spaces

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/install/installer/cmd/render_test.go
+++ b/install/installer/cmd/render_test.go
@@ -7,6 +7,7 @@ import (
 	"flag"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


install/installer/cmd/render_test.go
```diff
@@ -75,6 +76,8 @@ func TestRender(t *testing.T) {
 			f, err := os.OpenFile(goldenPath, os.O_RDWR, 0644)
 			defer f.Close()
 			require.NoError(t, err)
+			cmd := exec.Command("sed", "-i", "-e", "\"s/[[:space:]]*$//\"", goldenPath)
+                        defer cmd.Run()
```
jankeromnes, line 80: Finding: The code below,

```go
			// Compare
			require.Equal(t, string(content), got)
```

compares the live-generated, in-memory result with the golden file. So, fixing the golden file in post-processing does not work here.

We'd need to either:
- Save the live-generated, in-memory result to a file, then post-process, then compare with the (also clean) golden file
- Or, clean up trailing spaces in the live-generated, in-memory result (preferable)

PR #13722, opened 2022-10-10
[db] Add `applicationCluster` field to `d_b_workspace_cluster` table

andrew-farries:
## Description

As part of #9198  we want to start syncing the `d_b_workspace_cluster` table with `db-sync`.

Currently, the table differs between US and EU regions because each table contains only the data relevant to that region. For example, in the EU table, the `eu70` workspace cluster is marked as `available` and the `us70` cluster is `cordoned`.  In the US cluster `eu70` is `cordoned` and `us70` is `available`.

In order to sync the table we need to get to a point where there is no difference in the data in the table between EU and US regions.

To do that we will introduce a new field in the table called `applicationCluster` which records the name of the application cluster to which the record belongs. Thus, for each workspace cluster there will be two rows in Gitpod SaaS:

| name | applicationCluster | url | tls | state | ... |
| --- | --- | --- | --- | --- | --- |
| eu70 | eu02 | url | tls info | available | ... |
| eu70 | us02 | url | tls info | cordoned | ... |

Effectively the new `applicationCluster` column gives the table an extra dimension so that we can combine both tables (EU and US) into one.

As a first step towards this, this PR adds the column to the table and makes `gpctl` fill the value when `gpctl register`ing a new workspace cluster. The value is taken from the `GITPOD_INSTALLATION_SHORTNAME` environment variable in `ws-manager-bridge`.

⚠️ Needs https://github.com/gitpod-io/ops/pull/5758 merged first ⚠️ 

## Related Issue(s)

Part of #9198 

## How to test

I'd like to be able to register a workspace cluster from a preview environment to check that the value of `applicationCluster` is correctly populated.

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-db/src/typeorm/entity/db-workspace-cluster.ts
+++ b/components/gitpod-db/src/typeorm/entity/db-workspace-cluster.ts
@@ -84,4 +84,10 @@ export class DBWorkspaceCluster implements WorkspaceCluster {
         })(),
     })
     admissionConstraints?: AdmissionConstraint[];
+
+    @Column({
+        type: "varchar",
+        length: 255,
+    })
+    applicationCluster: string;
 }
--- a/components/gitpod-db/src/typeorm/entity/db-workspace-cluster.ts
+++ b/components/gitpod-db/src/typeorm/entity/db-workspace-cluster.ts
@@ -87,7 +87,7 @@ export class DBWorkspaceCluster implements WorkspaceCluster {
 
     @Column({
         type: "varchar",
-        length: 255,
+        length: 60,
     })
     applicationCluster: string;
 }
--- a/components/gitpod-db/src/typeorm/migration/1665071320428-AddColumnToWorkspaceClusterTable.ts
+++ b/components/gitpod-db/src/typeorm/migration/1665071320428-AddColumnToWorkspaceClusterTable.ts
@@ -1,0 +1,21 @@
+/**
+ * Copyright (c) 2022 Gitpod GmbH. All rights reserved.
+ * Licensed under the GNU Affero General Public License (AGPL).
+ * See License-AGPL.txt in the project root for license information.
+ */
+
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class AddColumnToWorkspaceClusterTable1665071320428 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        const installationShortname = process.env.GITPOD_INSTALLATION_SHORTNAME ?? "";
+
+        await queryRunner.query(
+            `ALTER TABLE d_b_workspace_cluster ADD COLUMN applicationCluster varchar(255) NOT NULL DEFAULT ''`,
+        );
+
+        await queryRunner.query(`UPDATE d_b_workspace_cluster SET applicationCluster = '${installationShortname}'`);
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {}
+}
--- a/components/gitpod-db/src/typeorm/migration/1665071320428-AddColumnToWorkspaceClusterTable.ts
+++ b/components/gitpod-db/src/typeorm/migration/1665071320428-AddColumnToWorkspaceClusterTable.ts
@@ -11,7 +11,7 @@ export class AddColumnToWorkspaceClusterTable1665071320428 implements MigrationI
         const installationShortname = process.env.GITPOD_INSTALLATION_SHORTNAME ?? "";
 
         await queryRunner.query(
-            `ALTER TABLE d_b_workspace_cluster ADD COLUMN applicationCluster varchar(255) NOT NULL DEFAULT ''`,
+            `ALTER TABLE d_b_workspace_cluster ADD COLUMN applicationCluster varchar(60) NOT NULL DEFAULT ''`,
         );
 
         await queryRunner.query(`UPDATE d_b_workspace_cluster SET applicationCluster = '${installationShortname}'`);
--- a/components/gitpod-protocol/src/workspace-cluster.ts
+++ b/components/gitpod-protocol/src/workspace-cluster.ts
@@ -15,6 +15,9 @@ export interface WorkspaceCluster {
     // Must be identical to the installationShortname of the cluster it represents!
     name: string;
 
+    // The name of the application cluster to which this cluster should be registered.
+    applicationCluster: string;
+
     // URL of the cluster's ws-manager API
     url: string;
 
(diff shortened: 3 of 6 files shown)


easyCZ:
@andrew-farries Would like to better understand the resulting semantics of this. It feels wrong to have the resulting rows in the table with the source cluster

geropl:
> @andrew-farries Would like to better understand the resulting semantics of this. It feels wrong to have the resulting rows in the table with the source cluster

@easyCZ You can imagine this to be a directed graph, where we want this table to contain all edges from application to workspace clusters. And each edge contains all parameters to configure the behavior of how the respective application cluster with the workspace cluster.

components/gitpod-protocol/src/workspace-cluster.ts
```diff
@@ -15,6 +15,9 @@ export interface WorkspaceCluster {
     // Must be identical to the installationShortname of the cluster it represents!
     name: string;
 
+    // The name of the application cluster to which this cluster should be registered.
```
geropl, line 18: :bulb: If we decide for non-255 chars we could document the different length-limit here.

components/gitpod-db/src/typeorm/migration/1665071320428-AddColumnToWorkspaceClusterTable.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class AddColumnToWorkspaceClusterTable1665071320428 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        const installationShortname = process.env.GITPOD_INSTALLATION_SHORTNAME;
+
+        await queryRunner.query(
+            `ALTER TABLE d_b_workspace_cluster ADD COLUMN applicationCluster varchar(255) NOT NULL DEFAULT ''`,
```
geropl, line 14: :+1: For the `NOT NULL DEFAULT ''`.

geropl:
@andrew-farries It would be nice if `gpctl` would display that field as well. But can easily be a follow up PR as long as the DB fields not carry meaning.

geropl:
When reviewing I noticed a mismatch between `bridge.cfg.installation`/`GITPOD_INSTALLATION_SHORTNAME` (`"dev"`)and the default local bridge `name` we create (`""`):
image

This:
 1. does _not_ block this change (only becomes relevant once we start to actively select by `applicationCluster`)
 2. might be either caused by:
    1. us misconfiguring preview environments
    2. a bug in `installer` that affects self-hosted as well (:warning: )

@andrew-farries  We should definitely investigate before the next Self-Hosted realease. :100:

geropl (approved):
Thank you @andrew-farries , the change LGTM!

/hold mainly for [this nit](https://github.com/gitpod-io/gitpod/pull/13722/files#r992234248), please decide whether you want to address or not

components/gitpod-db/src/typeorm/entity/db-workspace-cluster.ts
```diff
@@ -84,4 +84,10 @@ export class DBWorkspaceCluster implements WorkspaceCluster {
         })(),
     })
     admissionConstraints?: AdmissionConstraint[];
+
+    @Column({
+        type: "varchar",
+        length: 255,
```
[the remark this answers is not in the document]
andrew-farries: Thanks, shortened in b668b3e

andrew-farries:
> @andrew-farries It would be nice if `gpctl` would display that field as well. But can easily be a follow up PR as long as the DB fields not carry meaning.

Good point - will address this in a follow up PR.

andrew-farries:
> @andrew-farries We should definitely investigate before the next Self-Hosted realease. 💯

This is interesting. Looking at the installer code, they are both set to `InstallationShortName` so I'm not sure why one is empty.

https://github.com/gitpod-io/gitpod/blob/5025e1263a525950ae75f68315b6568594173a15/install/installer/pkg/components/ws-manager-bridge/configmap.go#L19

and 

https://github.com/gitpod-io/gitpod/blob/5025e1263a525950ae75f68315b6568594173a15/install/installer/pkg/components/ws-manager-bridge/objects.go#L38

sagor999:
What will happen to existing clusters after migration? I assume they will have an empty field for application cluster? Will this break anything?

Also, how does traffic routing happens with regards to this table? Does it have any affect on it? For example, in US, eu clusters are cordoned, so that no traffic from US gets sent to EU. (at least that is how I understand it). Want to make sure this change will not break anything anywhere else?

sagor999 (approved):
Approving since I am only an owner of one file, so do not want to block this PR.
Would love to receive response to the questions above as a sanity check.

geropl:
> What will happen to existing clusters after migration? I assume they will have an empty field for application cluster? Will this break anything?

@sagor999  In short: nothing, as a first step. As mentioned in the description, this PR is only about adding said field, without reading it anywhere.

> Also, how does traffic routing happens with regards to this table? Does it have any affect on it? For example, in US, eu clusters are cordoned, so that no traffic from US gets sent to EU. (at least that is how I understand it). Want to make sure this change will not break anything anywhere else?

The next step is to check that it's filled correctly (with `eu-02` and `us-02`, respectively). Once that's verified, we can start reading it, so that app cluster `eu-02` is only able to read all rows with that `applicationCluster='eu-02'`. This way, once we (later) start syncing that table, the code inside will still behave just the same as it does now. :+1: But instead of implicitly relying on the fact that we have a separate DB table per app cluster, we made the connection explicit with the `applicationCluster` column, and all can happily be stored in a single table. :lotus_position:

geropl:
> This is interesting. Looking at the installer code, they are both set to InstallationShortName so I'm not sure why one is empty.

@andrew-farries Maybe our preview env config code is broken/bad? :thinking:

andrew-farries:
> > This is interesting. Looking at the installer code, they are both set to InstallationShortName so I'm not sure why one is empty.
> 
> @andrew-farries Maybe our preview env config code is broken/bad? 🤔

The reason for this mismatch will be addressed by https://github.com/gitpod-io/gitpod/pull/13866/commits/97733893fe9580e455b5c115b87c1d0d5626726a

aledbf:
We don't have the shortName value in workspace clusters (yet). That is being [added here](https://github.com/gitpod-io/ops/pull/5795/files#diff-213153aee82c75bd8881fe14906e970ed7f14b8984ced98f3b863ea79a14a4fcR780)

Just to confirm, will this not break anything if we don't have such value now?

aledbf (approved):

PR #13724, opened 2022-10-10
Add a lock timeout for plan/apply and general timeout on apply

vulkoingim:
## Description

Adds a lock-timeout so jobs can wait a bit for the lock to get released, instead of failing immediately and a general timeout on `apply`. 

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


mads-hartmann (approved):
For clarification, does all workspaces share the same lock? I was hoping they had separate locks

vulkoingim:
> For clarification, does all workspaces share the same lock? I was hoping they had separate locks

No, every workspace has an individual lock. 

But if you do 2 quick pushes in succession, the 2nd one will fail as the first one would get the lock. This change makes the 2nd job a bit slower, but will wait until the lock is released instead of failing.

PR #13736, opened 2022-10-10
[public-api] Refactor to use connect handlers, route to HTTP server

easyCZ:
## Description

In this PR, we 1-1 map the old implementation in gRPC to the new implementation with Connect. Changes are deliberately not made to streamline the tests and simplify to keep the PR cleaner.

In general, the changes are the following (see [docs](https://connect.build/docs/go/grpc-compatibility#migration))
1. Use the generated ServiceHandlers
2. Wrap requests and responses in `connect.NewRequest` or `connect.NewResponse`
3. Use Connects error codes
4. Use HTTP server in tests

This PR also changes default routing of `api.<domain>` requests to the HTTP server, this has the following effect:
* Connect handlers handle both gRPC and HTTP traffic - this is fine, as Connect is designed for this
* Depending on the rollout sequencing, there can be a brief period where requests would fail. This is OK, as we don't get any traffic on this yet anyway.

## Related Issue(s)

* Depends on https://github.com/gitpod-io/gitpod/pull/13693

## How to test

Unit tests

1. Log in to preview and run the following in the console to get an access token:
```
	"function:getWorkspace",
	"function:getWorkspaces",
	"resource:default",]})
```
To make subsequent steps readily runnable, export the token in your shell:
```
export BEARER_TOKEN="<token>"
```

Validate we can reach the new APIs over cURL:
```
curl \
    --header 'Content-Type: application/json' \
    --header "Authorization: Bearer $BEARER_TOKEN" \
    --data '{}' \
    https://api.mp-public-72e473c7b0.preview.gitpod-dev.com/gitpod.v1.WorkspacesService/ListWorkspaces | jq

{}
```
We're doing a list call, and we don't have workspaces so an empty list response is expected.

Validate we can reach the APIs using a gRPC client
```
gpctl api workspaces list --address api.mp-public-72e473c7b0.preview.gitpod-dev.com:443 --token $BEARER_TOKEN 

ID        Owner        ContextURL
```
Again, we're not getting any workspaces back because we've not started any but the important part is we're not getting an error.

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/public-api-server/pkg/apiv1/prebuild.go
+++ b/components/public-api-server/pkg/apiv1/prebuild.go
@@ -6,23 +6,24 @@ package apiv1
 
 import (
 	"context"
+
+	"github.com/bufbuild/connect-go"
 	v1 "github.com/gitpod-io/gitpod/public-api/v1"
+	"github.com/gitpod-io/gitpod/public-api/v1/v1connect"
 )
 
 func NewPrebuildService() *PrebuildService {
-	return &PrebuildService{
-		UnimplementedPrebuildsServiceServer: &v1.UnimplementedPrebuildsServiceServer{},
-	}
+	return &PrebuildService{}
 }
 
 type PrebuildService struct {
-	*v1.UnimplementedPrebuildsServiceServer
+	v1connect.UnimplementedPrebuildsServiceHandler
 }
 
-func (p *PrebuildService) GetPrebuild(ctx context.Context, req *v1.GetPrebuildRequest) (*v1.GetPrebuildResponse, error) {
-	return &v1.GetPrebuildResponse{
+func (p *PrebuildService) GetPrebuild(ctx context.Context, req *connect.Request[v1.GetPrebuildRequest]) (*connect.Response[v1.GetPrebuildResponse], error) {
+	return connect.NewResponse(&v1.GetPrebuildResponse{
 		Prebuild: &v1.Prebuild{
-			PrebuildId: req.GetPrebuildId(),
+			PrebuildId: req.Msg.GetPrebuildId(),
 			Spec: &v1.PrebuildSpec{
 				Context: &v1.WorkspaceContext{
 					ContextUrl: "https://github.com/gitpod-io/gitpod",
@@ -32,5 +33,5 @@ func (p *PrebuildService) GetPrebuild(ctx context.Context, req *v1.GetPrebuildRe
 			},
 			Status: nil,
 		},
-	}, nil
+	}), nil
 }
--- a/components/public-api-server/pkg/apiv1/workspace.go
+++ b/components/public-api-server/pkg/apiv1/workspace.go
@@ -6,60 +6,48 @@ package apiv1
 
 import (
 	"context"
+	"fmt"
 
+	connect "github.com/bufbuild/connect-go"
 	protocol "github.com/gitpod-io/gitpod/gitpod-protocol"
+	"github.com/gitpod-io/gitpod/public-api-server/pkg/auth"
 	"github.com/gitpod-io/gitpod/public-api-server/pkg/proxy"
 	v1 "github.com/gitpod-io/gitpod/public-api/v1"
+	"github.com/gitpod-io/gitpod/public-api/v1/v1connect"
 	"github.com/grpc-ecosystem/go-grpc-middleware/logging/logrus/ctxlogrus"
 	"github.com/relvacode/iso8601"
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/metadata"
-	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
 func NewWorkspaceService(serverConnPool proxy.ServerConnectionPool) *WorkspaceService {
 	return &WorkspaceService{
-		connectionPool:                       serverConnPool,
-		UnimplementedWorkspacesServiceServer: &v1.UnimplementedWorkspacesServiceServer{},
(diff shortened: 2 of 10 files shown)


roboquat:

<details>

</details>

jenting (approved):
Changes for workspace team component LGTM

components/public-api-server/pkg/apiv1/prebuild.go
```diff
@@ -6,23 +6,24 @@ package apiv1
...
 func NewPrebuildService() *PrebuildService {
-	return &PrebuildService{
-		UnimplementedPrebuildsServiceServer: &v1.UnimplementedPrebuildsServiceServer{},
-	}
+	return &PrebuildService{}
 }
 
 type PrebuildService struct {
-	*v1.UnimplementedPrebuildsServiceServer
+	v1connect.UnimplementedPrebuildsServiceHandler
```
geropl, line 20: Asking for posterity: Is there an alternative to riddling our implementation with references to `connect`? It's not a big concern, as the changes look rather mechanical (near search-and-replace). Still wondering if there is a way to have a `ConnectProxy` that does the wrapping for us (I'm aware that this might be a tough ask for Go :upside_down_face: ).

components/public-api-server/pkg/apiv1/workspace.go
```diff
@@ -6,60 +6,48 @@ package apiv1
...
 		logger.WithError(err).Error("Failed to get connection to server.")
-		return nil, status.Error(codes.Internal, "failed to establish connection to downstream services")
+		return nil, connect.NewError(connect.CodeInternal, err)
 	}
 
-	workspace, err := server.GetWorkspace(ctx, r.GetWorkspaceId())
+	workspace, err := server.GetWorkspace(ctx, req.Msg.GetWorkspaceId())
 	if err != nil {
 		logger.WithError(err).Error("Failed to get workspace.")
-		converted := proxy.ConvertError(err)
```
geropl, line 51: Why don't we need this anymore? :thinking:

components/public-api-server/pkg/apiv1/workspace.go
```diff
@@ -73,72 +61,40 @@ func (w *WorkspaceService) GetWorkspace(ctx context.Context, r *v1.GetWorkspaceR
...
+		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("failed to establish connection to downstream services"))
 	}
 
-	ownerToken, err := server.GetOwnerToken(ctx, r.GetWorkspaceId())
+	ownerToken, err := server.GetOwnerToken(ctx, req.Msg.GetWorkspaceId())
 
 	if err != nil {
 		logger.WithError(err).Error("Failed to get owner token.")
-		converted := proxy.ConvertError(err)
-		switch status.Code(converted) {
```
geropl, line 97: Same question as above: Why is the switch superfluous?

geropl:
When testing as described I get a `bad handshake - 401 ` / `authorization header does not have a Bearer prefix: missing access token` (with both methods).

components/public-api-server/pkg/apiv1/workspace.go
easyCZ: Because the `proxy.ConvertError` method already does the switch

components/public-api-server/pkg/apiv1/prebuild.go
easyCZ: In general, we don't want to make this magic. We also want to stay as close to the official connect documentation as possible such that we don't need to re-teach things internally.

If it's just a wrapper, it's not adding match aside from importing the function from elsewhere (and that place imports connect).

geropl: > We also want to stay as close to the official connect documentation as possible

IMO the ideal situation would be if we could stick to the gRPC documentation. But I can definitely live with this. :100: It's not in the interest of connect, but maybe someone writes a wrapper at some point that we can use. :slightly_smiling_face:

easyCZ: Yes, but if we stuck with gRPC documentation then you'd not know you can invoke it with curl and use JSON as the payload - in Go principles we should not hide things, and the difference (or complexity) should be visible directly

geropl:
> You need to either re-install gpctl

Ah thx, my bad. :see_no_evil: `gpctl` works now :lotus_position: 

But I don't get why curl didn't work before but did now :cry: 

**Update**: Noticed you updated the description from `--header 'Authorization: Bearer $BEARER_TOKEN'` to `--header "Authorization: Bearer $BEARER_TOKEN"`: That explains it. :heavy_check_mark: 

> If you had opened the workspace from this branch

I did, but `gpctl` is only updated on dev image builds ([ref](https://github.com/gitpod-io/gitpod/blob/a9796bbdd5522e71c962370a11a1e95cd5cf5f85/dev/image/Dockerfile#L252)). :+1:

geropl (approved):
Changes LGTM, tests run and manual test works - nice work! :tada:

components/public-api-server/pkg/apiv1/prebuild.go
geropl: > in Go principles

I know, I know... but abstraction is a powerful concept (if it does not leak too much, hence I'm fine with as-is).

easyCZ:
> I did, but gpctl is only updated on dev image builds ([ref](https://github.com/gitpod-io/gitpod/blob/a9796bbdd5522e71c962370a11a1e95cd5cf5f85/dev/image/Dockerfile#L252)). 👍

Ah my bad, I should've made it more epxlicit.

Issue #11452, opened 2022-07-18
Prevent the only remaining team owner to change their role to member


gtsiolis:
### Bug description

While we are now preventing the last team owner to leave a team (see https://github.com/gitpod-io/gitpod/issues/4623), we're not preventing the only remaining team owner to change their role to member, leading edge cases like member-less teams after leaving the team (https://github.com/gitpod-io/gitpod/issues/6311) or owner-less teams.

See also [relevant discussion](https://gitpod.slack.com/archives/C02EN94AEPL/p1658155844284609) (internal). Cc @ jimmybrancaccio @jldec 

Cc @nidhinmanu because https://github.com/gitpod-io/gitpod/issues/6311#issuecomment-1186481212

From another [relevant discussion](https://gitpod.slack.com/archives/C02EN94AEPL/p1646217102883169?thread_ts=1646207958.538879&cid=C02EN94AEPL):

> The last owner can’t leave a team, but can change to a member and then leave. 

### Steps to reproduce

1. Create a team
2. Change your role to member
3. Leave a team
4. Try creating a new team with the same name ❗ 

<img src="https://front.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_3opl6)

easyCZ:
Fix is [here](https://github.com/gitpod-io/gitpod/pull/13744)

PR #13744, opened 2022-10-11
[server]  Disallow removing all Owners from a team

easyCZ:
## Description

Block changing role on teams of size 1 (or less).

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/11452

## Release Notes

```release-note
[teams] Disallow removing all Owners from a team
```

## Werft options:

- [x] /werft with-preview

### DIFF
--- a/components/gitpod-db/src/typeorm/team-db-impl.ts
+++ b/components/gitpod-db/src/typeorm/team-db-impl.ts
@@ -203,6 +203,18 @@ export class TeamDBImpl implements TeamDB {
             throw new ResponseError(ErrorCodes.NOT_FOUND, "A team with this ID could not be found");
         }
         const membershipRepo = await this.getMembershipRepo();
+
+        if (role != "owner") {
+            const ownerCount = await membershipRepo.count({
+                teamId,
+                role: "owner",
+                deleted: false,
+            });
+            if (ownerCount <= 1) {
+                throw new ResponseError(ErrorCodes.CONFLICT, "Team must retain at least one owner");
+            }
+        }
+
         const membership = await membershipRepo.findOne({ teamId, userId, deleted: false });
         if (!membership) {
             throw new ResponseError(ErrorCodes.NOT_FOUND, "The user is not currently a member of this team");


components/gitpod-db/src/typeorm/team-db-impl.ts
```diff
@@ -203,6 +203,12 @@ export class TeamDBImpl implements TeamDB {
             throw new ResponseError(ErrorCodes.NOT_FOUND, "A team with this ID could not be found");
         }
         const membershipRepo = await this.getMembershipRepo();
+
+        const teamSize = await membershipRepo.count({ teamId, deleted: false });
+        if (teamSize <= 1) {
```
jankeromnes, line 208: This isn't exactly the fix we need, right? (E.g. what if you're the last Owner, and there are 3 Members -- you shouldn't be able to change your role to Member)

easyCZ: Good spot. Fixed now. Will run through it once preview wakes up

components/gitpod-db/src/typeorm/team-db-impl.ts
```diff
@@ -203,6 +203,19 @@ export class TeamDBImpl implements TeamDB {
...
+        if (role != "owner") {
+            const ownerCountAfterChange = await membershipRepo.count({
+                teamId,
+                userId: Not(userId),
+                role: "owner",
+                deleted: false,
+            });
+            if (ownerCountAfterChange < 1) {
+                throw new ResponseError(ErrorCodes.CONFLICT, "Team must retain at least one owner");
+            }
```
jankeromnes, line 216: Interesting -- I've never seen typeorm's `Not` being used. Minor suggestion: Maybe it could be a bit simpler and less "magical" to just check the total number of owners? 😇 

```suggestion
            const ownerCount = await membershipRepo.count({
                teamId,
                role: "owner",
                deleted: false,
            });
            if (ownerCount <= 1) {
                throw new ResponseError(ErrorCodes.CONFLICT, "Team must retain at least one owner");
            }
```

easyCZ: Yep, this also works due to the `if (role == 'owner')` check. We could drop that check and do the Not exclusion but not sure if we gain much - it's just a bit more resilient to role spec changes.

easyCZ: Used your suggestion.

jankeromnes (approved):
Tested and works! The error is never shown in the UI (so you think you've successfully set yourself to Member until the next time you visit the page), however I'd say this was a pre-existing issue and can be fixed in follow-up. (The error is successfully shown in the console though.)

Issue #13389, opened 2022-09-28
Adjust cost centers and usage limits depending to spec


jankeromnes:
For un-upgraded cost centers (i.e. not Stripe customer), the usage limit should be:
- **500 credits** for individiual users (the free tier)
- **0 credits** for teams

These should also not be customizable (hard usage limit).

Then, for upgraded cost centers (i.e. Stripe customers), the usage limits should default to:
- **1000 credits** for individual users (customizable, but you can only set it to >= 1000 credits)
- **1000 credits * team members** for teams (customizable to any value)

jankeromnes:
In addition to the customizable / non-customizable / minimum value specs mentioned above, there is a suggestion by @svenefftinge to always ask new Stripe customers what limit they want when they sign up:

> Would it be better to ask the user what spending limit they want when they create a team? Hardcoding this in the backend doesn't seem to make so much sense, as e.g. the spending limit for teams greatly varies with the team size.
I've put the defaults into the usage component because of the get-or-create semantic of cost centers. But they will always be created as `billingStrategy: other` and Stripe cost centers are always created by users, so we can also pass the spendingLimit. Happy to discuss

I guess this requires further alignment on what we do by default; what we ask for and when; and how we enforce the minimum / non-customizable values.

easyCZ:
While doing this, it would be great to do it in a such a way that it complements the other scheduled issue in the current sprint - https://github.com/gitpod-io/gitpod/issues/13198 - that is move more of the handling logic to usage.

jankeromnes:
Fixed in https://github.com/gitpod-io/gitpod/pull/13746 and https://github.com/gitpod-io/gitpod/pull/13798

PR #13535, opened 2022-10-03
Configure default usage limits for new Stripe customers

jankeromnes:
## Description

Configure default usage limits for new Stripe customers, but don't use them anywhere yet (precursor to https://github.com/gitpod-io/gitpod/pull/13398)

## Related Issue(s)

Part of https://github.com/gitpod-io/gitpod/issues/13389

## How to test

1. Should build

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-local-preview
      If enabled this will build `install/preview`
- [ ] /werft with-preview
- [ ] /werft with-integration-tests=all
      Valid options are `all`, `workspace`, `webapp`, `ide`

PR #13746, opened 2022-10-11
[Usage-based] Adjust Stripe usage limits to spec

jankeromnes:
## Description

Adjust Stripe usage limits to spec

## Related Issue(s)

Part of https://github.com/gitpod-io/gitpod/issues/13389
~~Depends on https://github.com/gitpod-io/gitpod/pull/13798~~

## How to test

1. Create a team called "Gitpod [Something]"
2. Go to team billing
3. There shouldn't be a visible usage limit
4. Upgrade the team by adding a payment method
5. The default usage limit should be `1000 * team_size` (but you can set it to any value)
6. Go to your individual billing settings
7. The usage limit should be `500` (not editable)
8. Upgrade by adding a payment method
9. The usage limit should now be `1000`, and you should be able to set any value >= 1000 (but not < )
10. Cancel your individual user subscription
11. The limit should be back to `500` (not editable)

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview
- [x] /werft with-payment

### DIFF
--- a/components/dashboard/src/components/UsageBasedBillingConfig.tsx
+++ b/components/dashboard/src/components/UsageBasedBillingConfig.tsx
@@ -92,7 +92,15 @@ export default function UsageBasedBillingConfig({ attributionId }: Props) {
             setPendingStripeSubscription(pendingSubscription);
             window.localStorage.setItem(localStorageKey, JSON.stringify(pendingSubscription));
             try {
-                await getGitpodService().server.subscribeToStripe(attributionId, setupIntentId);
+                // Pick a good initial value for the Stripe usage limit (1000 * team_size)
+                // FIXME: Should we ask the customer to confirm or edit this default limit?
+                let limit = 1000;
+                const attrId = AttributionId.parse(attributionId);
+                if (attrId?.kind === "team") {
+                    const members = await getGitpodService().server.getTeamMembers(attrId.teamId);
+                    limit = 1000 * members.length;
+                }
+                await getGitpodService().server.subscribeToStripe(attributionId, setupIntentId, limit);
             } catch (error) {
                 console.error("Could not subscribe to Stripe", error);
                 window.localStorage.removeItem(localStorageKey);
--- a/components/dashboard/src/components/UsageBasedBillingConfig.tsx
+++ b/components/dashboard/src/components/UsageBasedBillingConfig.tsx
@@ -102,9 +102,10 @@ export default function UsageBasedBillingConfig({ attributionId }: Props) {
                     const members = await getGitpodService().server.getTeamMembers(attrId.teamId);
                     limit = BASE_USAGE_LIMIT_FOR_STRIPE_USERS * members.length;
                 }
-                await getGitpodService().server.subscribeToStripe(attributionId, setupIntentId, limit);
-                const newLimit = await getGitpodService().server.getUsageLimit(attributionId);
-                setUsageLimit(newLimit);
+                const newLimit = await getGitpodService().server.subscribeToStripe(attributionId, setupIntentId, limit);
+                if (newLimit) {
+                    setUsageLimit(newLimit);
+                }
             } catch (error) {
                 console.error("Could not subscribe to Stripe", error);
                 window.localStorage.removeItem(localStorageKey);
--- a/components/dashboard/src/components/UsageBasedBillingConfig.tsx
+++ b/components/dashboard/src/components/UsageBasedBillingConfig.tsx
@@ -407,6 +407,7 @@ export default function UsageBasedBillingConfig({ attributionId }: Props) {
             )}
             {showUpdateLimitModal && (
                 <UpdateLimitModal
+                    minValue={AttributionId.parse(attributionId || "")?.kind === "user" ? 1000 : 0}
                     currentValue={usageLimit}
                     onClose={() => setShowUpdateLimitModal(false)}
                     onUpdate={(newLimit) => updateUsageLimit(newLimit)}
@@ -537,12 +538,13 @@ function CreditCardInputForm(props: { attributionId: string }) {
 }
 
 function UpdateLimitModal(props: {
+    minValue?: number;
     currentValue: number | undefined;
     onClose: () => void;
     onUpdate: (newLimit: number) => {};
 }) {
     const [newLimit, setNewLimit] = useState<string | undefined>(
-        props.currentValue ? String(props.currentValue) : undefined,
+        typeof props.currentValue === "number" ? String(props.currentValue) : undefined,
     );
 
     return (
@@ -556,7 +558,7 @@ function UpdateLimitModal(props: {
                     <div className="w-full">
                         <input
                             type="number"
-                            min={0}
+                            min={props.minValue || 0}
                             value={newLimit}
                             className="rounded-md w-full truncate overflow-x-scroll pr-8"
                             onChange={(e) => setNewLimit(e.target.value)}
--- a/components/dashboard/src/components/UsageBasedBillingConfig.tsx
+++ b/components/dashboard/src/components/UsageBasedBillingConfig.tsx
@@ -101,6 +101,8 @@ export default function UsageBasedBillingConfig({ attributionId }: Props) {
                     limit = 1000 * members.length;
                 }
                 await getGitpodService().server.subscribeToStripe(attributionId, setupIntentId, limit);
+                const newLimit = await getGitpodService().server.getUsageLimit(attributionId);
+                setUsageLimit(newLimit);
             } catch (error) {
                 console.error("Could not subscribe to Stripe", error);
                 window.localStorage.removeItem(localStorageKey);
--- a/components/dashboard/src/components/UsageBasedBillingConfig.tsx
+++ b/components/dashboard/src/components/UsageBasedBillingConfig.tsx
@@ -20,6 +20,8 @@ import DropDown from "../components/DropDown";
 import Modal from "../components/Modal";
 import Alert from "./Alert";
 
+const BASE_USAGE_LIMIT_FOR_STRIPE_USERS = 1000;
+
 type PendingStripeSubscription = { pendingSince: number };
 
 interface Props {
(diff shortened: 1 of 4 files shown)


roboquat:

<details>

</details>

jankeromnes:
Issue: Immediately after upgrading, the limit shows up wrong. After a refresh it's okay. I guess we need to refresh that state somehow.

jankeromnes:
Okay, this now works almost as expected. 🎉

Except for the final step:

> 11. The limit should be back to `500` (not editable)

(When you cancel, the limit seems to remain unchanged? I think this should be fixed in the CostCenter as part of https://github.com/gitpod-io/gitpod/issues/13389)

components/dashboard/src/components/UsageBasedBillingConfig.tsx
```diff
@@ -92,7 +92,17 @@ export default function UsageBasedBillingConfig({ attributionId }: Props) {
...
+                // Pick a good initial value for the Stripe usage limit (1000 * team_size)
+                // FIXME: Should we ask the customer to confirm or edit this default limit?
+                let limit = 1000;
+                const attrId = AttributionId.parse(attributionId);
+                if (attrId?.kind === "team") {
+                    const members = await getGitpodService().server.getTeamMembers(attrId.teamId);
+                    limit = 1000 * members.length;
+                }
+                await getGitpodService().server.subscribeToStripe(attributionId, setupIntentId, limit);
+                const newLimit = await getGitpodService().server.getUsageLimit(attributionId);
```
easyCZ, line 104: Why would we get the usage limit now that we've set it to a given value (which we know) and the call hasn't failed because it proceeded to the next expression?

components/dashboard/src/components/UsageBasedBillingConfig.tsx
```diff
@@ -92,7 +92,17 @@ export default function UsageBasedBillingConfig({ attributionId }: Props) {
...
-                await getGitpodService().server.subscribeToStripe(attributionId, setupIntentId);
+                // Pick a good initial value for the Stripe usage limit (1000 * team_size)
+                // FIXME: Should we ask the customer to confirm or edit this default limit?
+                let limit = 1000;
+                const attrId = AttributionId.parse(attributionId);
+                if (attrId?.kind === "team") {
+                    const members = await getGitpodService().server.getTeamMembers(attrId.teamId);
+                    limit = 1000 * members.length;
+                }
+                await getGitpodService().server.subscribeToStripe(attributionId, setupIntentId, limit);
```
easyCZ, line 103: This would allow me to set the usage limit to an arbitrary value because we just pass that into the SetCostCenter call. Where do we validate this?

components/dashboard/src/components/UsageBasedBillingConfig.tsx
```diff
@@ -529,50 +540,64 @@ function CreditCardInputForm(props: { attributionId: string }) {
...
+    minValue?: number;
     currentValue: number | undefined;
     onClose: () => void;
     onUpdate: (newLimit: number) => {};
 }) {
+    const [error, setError] = useState<string>("");
     const [newLimit, setNewLimit] = useState<string | undefined>(
-        props.currentValue ? String(props.currentValue) : undefined,
+        typeof props.currentValue === "number" ? String(props.currentValue) : undefined,
     );
```
easyCZ, line 551: Why is the limit a string?

components/dashboard/src/components/UsageBasedBillingConfig.tsx
jankeromnes: Instead of assuming we know the current limit (based on our current understanding of the backend), I thought it's better to leave the backend (i.e. Cost Center logic) be the single authority on what limits are legal / acceptable / and when they should be reset / to what.

I feel like not asking, and proceeding on assumptions would make this code more fragile.

components/dashboard/src/components/UsageBasedBillingConfig.tsx
jankeromnes: While there is some front-end validation, I'm relying on the backend (Cost Center code) to be the final validator. The goal here is to eventually be able to ask users what sort of limit they would like when they sign up, but I think the backend should be the main authority on what is or isn't acceptable.

components/dashboard/src/components/UsageBasedBillingConfig.tsx
easyCZ: Shall we return the new cost center (and the assigned spending limit) from the `subscribeToStripe` call instead? That way, we push this logic down into the backend rather than having the client deal with this logic.

jankeromnes: @easyCZ I like this idea! (Returning a CostCenter object to the front-end, or maybe just the usage limit?)

easyCZ: For this, I'll extend the `SetCostCenter` API to actually return the updated CostCenter such that you don't have to do Update followed be a Get

components/dashboard/src/components/UsageBasedBillingConfig.tsx
easyCZ, line 103: Thinking about this, I think we can shift this to server also.

easyCZ:
The counterpart PR which adds validation is here https://github.com/gitpod-io/gitpod/pull/13798

geropl:
Having a look here as well.

components/dashboard/src/components/UsageBasedBillingConfig.tsx
```diff
@@ -399,6 +409,7 @@ export default function UsageBasedBillingConfig({ attributionId }: Props) {
             )}
             {showUpdateLimitModal && (
                 <UpdateLimitModal
+                    minValue={AttributionId.parse(attributionId || "")?.kind === "user" ? 1000 : 0}
```
geropl, line 412: Can we make the magic number `1000` a constant, please?

geropl:
More general feedback, which does not necessarily need to be solved with this PR:

### Usage Limit input type
image

Can we replace input type `number` with a text field? The up/down arrows are just weird and feel broken. :+1: 

### Error message on personal upgrade
image

(worked on retry)

geropl (approved):
Code LGTM, tested and:
 - step 11. does not work (yet) as per [this comment](https://github.com/gitpod-io/gitpod/pull/13746#issuecomment-1274887962)
 - there are [further potential improvements](https://github.com/gitpod-io/gitpod/pull/13746#issuecomment-1277158625), but those are out of scope IMO
So: :heavy_check_mark: 

/hold for [this nit](https://github.com/gitpod-io/gitpod/pull/13746/files#r994243797).

components/dashboard/src/components/UsageBasedBillingConfig.tsx
jankeromnes: Thanks! But I moved this out of server and into dashboard at Sven's request, because maybe in the future we'll want to ask users to define or confirm their usage limit while they're upgrading. I think we can leave this as is for now.

jankeromnes:
Many thanks for the reviews! Briefly taking this back to Draft in order to:
- address nits
- test https://github.com/gitpod-io/gitpod/pull/13798 by rebasing on top of it

components/dashboard/src/components/UsageBasedBillingConfig.tsx
easyCZ: IMO, we should move it. If we'd like to suggest a limit which the user then confirms, we can suggest it in the Intent flow (teamSize * X), and send along with the `subscribeToStripe` call for actual persistence. We make the UI unnecessarily complex with the extra logic and introduce more failure modes.

Would like to also understand the context of Sven's request, do you have a link? What held true then, may no longer be the case. In general, the less logic the client side has, the better and given we're not doing the suggestion now .. YAGNI

components/dashboard/src/components/UsageBasedBillingConfig.tsx
jankeromnes: Fixed ✅

components/dashboard/src/components/UsageBasedBillingConfig.tsx
jankeromnes: Hmm, if the code works as is, I'd be keen to merge it as is, then figure out how to improve it in follow-up PRs. Happy to file a follow-up issue for this! 😇 

> Would like to also understand the context of Sven's request, do you have a link?

https://github.com/gitpod-io/gitpod/pull/13535

> YAGNI

I can never remember what those acronyms mean, and I'm too lazy to look them up, so I just glaze over them and lose their meaning 😝 (also https://acronyms-suck.com 😇)

jankeromnes:
> Usage Limit input type

@geropl Fixed (it's now an `<input type="text">`)

> Error message on personal upgrade

@geropl Known issue: https://github.com/gitpod-io/gitpod/issues/13177

> Before landing this, we'll want to do a rebase to avoid having to do a get after set

@easyCZ Rebased and now using the new limit returned by `subscribeToStripe`

I've also uncoupled the Pull Requests again due to https://github.com/gitpod-io/gitpod/pull/13798/files -- happy to pair on that one if you'd like. Maybe it's a missing webhook? I see that the user cost center never goes back to `other` after cancel 👀

components/dashboard/src/components/UsageBasedBillingConfig.tsx
jankeromnes: Refactored to not do a Get after Set ✅ thanks!

jankeromnes:
Okay, majority of nits addressed, and this works 90% in preview environments (with a chance to work 100% in staging/production if webhooks work there -- in any case this problem pre-dates & is unrelated to this PR).

Happy to make any further adjustments as follow-up issues. I'll file these today, but please let me know if you think I forgot anything.

/unhold

components/dashboard/src/components/UsageBasedBillingConfig.tsx
jankeromnes: Many thanks for implementing the back-end validation in https://github.com/gitpod-io/gitpod/pull/13798! 👍

jankeromnes: Follow-up issue is here: https://github.com/gitpod-io/gitpod/issues/13835

components/server/ee/src/workspace/gitpod-server-impl.ts
```diff
@@ -2113,8 +2113,12 @@ export class GitpodServerEEImpl extends GitpodServerImpl {
...
     }
 
-    protected defaultSpendingLimit = 100;
-    async subscribeToStripe(ctx: TraceContext, attributionId: string, setupIntentId: string): Promise<void> {
+    async subscribeToStripe(
+        ctx: TraceContext,
+        attributionId: string,
+        setupIntentId: string,
+        usageLimit: number,
+    ): Promise<number | undefined> {
```
easyCZ, line 2121: FWIW, this API is now super awkward in terms of the response. A consumer wouldn't expect to get a number as a response to `subscribeToStripe`. `subscribe` should return a Subscription, which can have a property of `usageLimit`.

PR #13748, opened 2022-10-11
Fix preview name prefix

vulkoingim:
## Description

If there's a `preview` anywhere in the name of the branch/preview env, it will get replaced and the delete will fail. This handles the prefix properly.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


mads-hartmann (approved):
haha good catch

Issue #13574, opened 2022-10-04
Epic: Dashboard communicates with Public API


easyCZ:
### Summary

This epic focuses on removing technical blockers which prevent the dashboard (web application) to use the Public API to retrieve data.

This epic does not focus on implementing and migrating API calls that dashboard makes over to public api.

### Context

We want to give ourselves a stronger API interface which can be consumed by dashboard, but also external clients. The first step is to dog-food the API using our existing clients - dashboard.

To address the technical limitations, we need to tackle the following:
* Exposing endpoints on the Public API over HTTP1.1 - web clients cannot use HTTP2
* Generating & using clients in the dashboard to communicate with public api
* Laying the groundwork for experimental and stable APIs

### Value

Unblocking increase in traffic on Public API

### Acceptance Criteria

Dashboard webapp can communicate with Public API

## Tasks
- [x] Route HTTP 1.1 traffic from api.gitpod.io to public-api-server 
	- https://github.com/gitpod-io/gitpod/pull/13582
- [x] Refactor existing public-api protos into stable and experimental packages
	- https://github.com/gitpod-io/gitpod/pull/13579
- [x] Generate clients & interfaces for Connect.build for public API
	- https://github.com/gitpod-io/gitpod/pull/13550
- [x] Bind generated (unimplemented) Connect.build handler to Public API HTTP mux
   - https://github.com/gitpod-io/gitpod/pull/13595
- [x] Configure CORS for api.gitpod.io
	- https://github.com/gitpod-io/gitpod/pull/13743
- [x] Refactor WorkspaceService to implement connect.build handler
    - https://github.com/gitpod-io/gitpod/pull/13736
- [x] Ensure public-api has sufficient observability to show incoming requests
    - https://github.com/gitpod-io/gitpod/issues/13903
 - [x] Cache proxy connections to server
    - https://github.com/gitpod-io/gitpod/pull/13750
 - [x] #14114

easyCZ:
This is now complete.

PR #13750, opened 2022-10-11
[public-api] Implement connection pool with LRU cache

easyCZ:
## Description

Adds a connection pool implementation for `ServerConnectionPool`. This pool caches connections by their token. The cache is an LRU.

Initially, the LRU size is set to 3k which is derived from existing WS connection count on server.
There are metrics to track the cache hit rate, as well as how full the cache is.

## Related Issue(s)

* Part of https://github.com/gitpod-io/gitpod/issues/13574
* Depends on https://github.com/gitpod-io/gitpod/pull/14171

## How to test

Unit test

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/public-api-server/pkg/proxy/conn.go
+++ b/components/public-api-server/pkg/proxy/conn.go
@@ -7,10 +7,14 @@ package proxy
 import (
 	"context"
 	"fmt"
-	gitpod "github.com/gitpod-io/gitpod/gitpod-protocol"
-	"github.com/grpc-ecosystem/go-grpc-middleware/logging/logrus/ctxlogrus"
 	"net/url"
 	"time"
+
+	"github.com/gitpod-io/gitpod/common-go/log"
+	gitpod "github.com/gitpod-io/gitpod/gitpod-protocol"
+	"github.com/grpc-ecosystem/go-grpc-middleware/logging/logrus/ctxlogrus"
+
+	lru "github.com/hashicorp/golang-lru"
 )
 
 type ServerConnectionPool interface {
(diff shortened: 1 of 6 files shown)


components/public-api-server/pkg/proxy/conn.go
```diff
@@ -42,3 +46,69 @@ func (p *NoConnectionPool) Get(ctx context.Context, token string) (gitpod.APIInt
 
 	return server, nil
 }
+
+func NewConnectionPool(address *url.URL, poolSize int) (*ConnectionPool, error) {
+	cache, err := lru.NewWithEvict(poolSize, func(key, value interface{}) {
```
andrew-farries, line 51: `key` is unused here.

components/public-api-server/pkg/proxy/conn.go
```diff
@@ -42,3 +46,69 @@ func (p *NoConnectionPool) Get(ctx context.Context, token string) (gitpod.APIInt
...
+			return
+		}
+
+		closeErr := conn.Close()
+		if closeErr != nil {
+			log.Log.WithError(closeErr).Warn("Failed to close connection to server.")
+		}
+	})
+	if err != nil {
+		return nil, fmt.Errorf("faield to create LRU cache: %w", err)
```
andrew-farries, line 66:
```suggestion
		return nil, fmt.Errorf("failed to create LRU cache: %w", err)
```

components/public-api-server/pkg/proxy/conn.go
```diff
@@ -42,3 +46,69 @@ func (p *NoConnectionPool) Get(ctx context.Context, token string) (gitpod.APIInt
...
 	return server, nil
 }
+
+func NewConnectionPool(address *url.URL, poolSize int) (*ConnectionPool, error) {
+	cache, err := lru.NewWithEvict(poolSize, func(key, value interface{}) {
+		connectionPoolSize.Dec()
+
+		// We attempt to gracefully close the connection
+		conn, ok := value.(gitpod.APIInterface)
+		if !ok {
```
andrew-farries, line 56: should we at least log something here? We'd always expect to be able to cast to `gitpod.APIInterface` right?

components/public-api-server/pkg/proxy/conn.go
```diff
@@ -42,3 +46,69 @@ func (p *NoConnectionPool) Get(ctx context.Context, token string) (gitpod.APIInt
...
+
+	return &ConnectionPool{
+		cache: cache,
+		connConstructor: func(token string) (gitpod.APIInterface, error) {
+			return gitpod.ConnectToServer(address.String(), gitpod.ConnectToServerOpts{
+				// We want the connection to persist beyond the lifecycle of a single request
+				Context: context.Background(),
+				Token:   token,
+				Log:     log.Log,
+				CloseHandler: func(err error) {
```
andrew-farries, line 77: `err` is unused.

components/public-api-server/pkg/proxy/conn.go
```diff
@@ -42,3 +46,69 @@ func (p *NoConnectionPool) Get(ctx context.Context, token string) (gitpod.APIInt
...
+	})
+	if err != nil {
+		return nil, fmt.Errorf("faield to create LRU cache: %w", err)
+	}
+
+	return &ConnectionPool{
+		cache: cache,
+		connConstructor: func(token string) (gitpod.APIInterface, error) {
+			return gitpod.ConnectToServer(address.String(), gitpod.ConnectToServerOpts{
+				// We want the connection to persist beyond the lifecycle of a single request
```
andrew-farries, line 73: Not sure I understand the significance of the comment. Does it relate to the `Context` line below?

components/public-api-server/pkg/proxy/conn.go
```diff
@@ -42,3 +46,69 @@ func (p *NoConnectionPool) Get(ctx context.Context, token string) (gitpod.APIInt
...
+
+	// cache stores token to connection mapping
+	cache *lru.Cache
+}
+
+func (p *ConnectionPool) Get(ctx context.Context, token string) (gitpod.APIInterface, error) {
+	cached, found := p.cache.Get(token)
+	reportCacheOutcome(found)
+	if found {
+		connectionPoolCacheOutcome.WithLabelValues("true").Inc()
```
andrew-farries, line 98: I think this will increment the metric twice, given the call to `reportCacheOutcome` 2 lines above.

andrew-farries (changes requested):

components/public-api-server/pkg/proxy/conn.go
easyCZ: yep, not propagating the context from the request. I can clarify.

components/public-api-server/pkg/proxy/conn.go
easyCZ: Good spot. I refactored the metric but forgot to remove it here.

components/public-api-server/pkg/proxy/conn.go
easyCZ: Added. Thanks.

andrew-farries (approved):

PR #13680, opened 2022-10-07
fix(gitpod-cli): ports list safe access to exposed url

andreafalzetti:
## Description
This PR fix safe access to `port.Exposed.*` for both `Url` and `Visibility`.

Also, I've refactored the existing logic to determine the status, to be more readable and to follow the [vscode implementation](https://github.com/gitpod-io/openvscode-server/blob/gp-code/main/extensions/gitpod-shared/src/workspacePort.ts#L116-L140)

```
 gp ports list
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x97c5ae]

goroutine 1 [running]:
github.com/gitpod-io/gitpod/gitpod-cli/cmd.glob..func12(0xfb5fe0, {0xac9654, 0x0, 0x0})
        github.com/gitpod-io/gitpod/gitpod-cli/cmd/ports-list.go:86 +0x3ae
github.com/spf13/cobra.(*Command).execute(0xfb5fe0, {0xff6aa8, 0x0, 0x0})
        github.com/spf13/cobra@v1.1.3/command.go:856 +0x663
github.com/spf13/cobra.(*Command).ExecuteC(0xfb5ae0)
        github.com/spf13/cobra@v1.1.3/command.go:960 +0x39d
github.com/spf13/cobra.(*Command).Execute(...)
        github.com/spf13/cobra@v1.1.3/command.go:897
github.com/gitpod-io/gitpod/gitpod-cli/cmd.Execute()
        github.com/gitpod-io/gitpod/gitpod-cli/cmd/root.go:35 +0x325
main.main()
        github.com/gitpod-io/gitpod/gitpod-cli/main.go:12 +0x17
```

## Related Issue(s)

Fixes #11990

## How to test
- Start a workspace in prev env: https://afalz-fix-962bf6477c.preview.gitpod-dev.com/#https://github.com/gitpod-io/empty
- Start a workspace in production from this PR https://gitpod.io/#https://github.com/gitpod-io/gitpod/pull/13680
- in the prod workspace, check `kubectx` to make sure you are in the context of this prev env, otherwise run `previewctl install-context`
- in the prod workspace, scale down servers `kubectl scale --replicas=0 deployment server` 
- in the prev env workspace, try to trigger the port exposure with `curl lama.sh | sh -s -- -p 3333`
- in the prev env workspace, run `gp ports list` and make sure it does not throw any error
[...]

PR #13752, opened 2022-10-11
feat(gitpod-cli): report errors to ide-metrics-api

### LINKED ISSUE #13707: Integrate `gp cli` with `ide-metrics-api` for at least one command
## Description
If we integrate `gp cli` with `ide-metrics-api` we can collect errors happening in the go client.

andreafalzetti:
## Description
Introduce error reporting from `gp cli` (startin from `gp ports list`) to `ide-metrics-api`. Errors will be available in GCP.

## Related Issue(s)

Fixes #13707

## Follow up actions after merging
1. Add `util.logError` in all known errors 
1. Provide the `userId`
1. Find a way to capture unhandled exceptions at root level, e.g: [[1]](https://www.digitalocean.com/community/tutorials/handling-panics-in-go)

## How to test
- [Start an empty workspace in prev env](https://afalz-gitpb5e1db935e.preview.gitpod-dev.com/#https://github.com/gitpod-io/empty/) 
- in the prev env workspace, try to trigger the port exposure with `curl lama.sh | sh -s -- -p 3333`
- in the prev env workspace, break supervisor-api by replacing the following env var: `export SUPERVISOR_ADDR=localhost:22999x`
- in the prev env workspace, run `gp ports list` and make sure it does not throw any error
- [Start a workspace in prod for this PR](https://gitpod.io/#https://github.com/gitpod-io/gitpod/pull/13752)  then observe the logs for `kubectl logs ide-metrics-(...)` (you can also see the name of the ide-metrics pod via `kubectl get pods | grep ide-metrics`

## Release Notes

```release-note
NONE
```

## Werft options:

- [x] /werft with-preview


andreafalzetti:
Closed in favour of https://github.com/gitpod-io/gitpod/pull/13752 because prev env for this PR is broken

felladrin:
ℹ️ I'm reviewing it now!

components/gitpod-cli/cmd/ports-list.go
```diff
@@ -30,7 +29,7 @@ var listPortsCmd = &cobra.Command{
 		ports, portsListError := supervisor_helper.GetPortsList(ctx)
 
 		if portsListError != nil {
-			log.WithError(portsListError).Error("Could not get the ports list.")
+			utils.LogError(ctx, portsListError, "Could not get the ports list.")
```
iQQBot, line 32: This error cannot be triggered by following the `how to test` procedure, because the port information is obtained through the supervisor and has nothing to do with the server, so scaling the server to 0 does not generate any error here

If you change `SUPERVISOR_ADDR` to something else, it does trigger the error, but you can't do errorReporting at the same time, because errorReporting also relies on the supervisor passing back the gitpod host

felladrin: Does it mean it can't be tested, @iQQBot?
If so, can we merge it by just reviewing the code and confirming (on Preview Environment) it's not breaking anything?

felladrin (commented):
Are these two testing steps up to date?
- in the prev env workspace, break supervisor-api by replacing the following env var: export `SUPERVISOR_ADDR=localhost:22999x`
- in the prev env workspace, run `gp ports list` and make sure it does not throw any error

On the preview environment it throws error when `SUPERVISOR_ADDR=localhost:22999x`.
image

components/gitpod-cli/cmd/ports-list.go
iQQBot: Yes, we don't have a good way to test it, but it also shows that it's not very good to add here, where we actually have a high probability of not encountering problems

andreafalzetti: @iQQBot, for context we started here because we had another issue but eventually we'll add it to all errors

iQQBot: Do we have any reports or issues about this line? I think there is very little chance of error here

andreafalzetti: > we started here

I meant in this file. Because of https://github.com/gitpod-io/gitpod/pull/13680

iQQBot: make sense

components/gitpod-cli/pkg/utils/logError.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/go-errors/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+func LogError(ctx context.Context, errToReport error, errorMessage string) {
+	log.WithError(errToReport).Error(errorMessage)
+
+	conn, err := supervisor_helper.Dial(ctx)
+	if err != nil {
+		log.Fatal(err)
```
akosyakov, line 28: I don't think we should use Fatal in this method

andreafalzetti:
/hold

doing some tests

felladrin (approved):
Reviewed/tested these changes with Andrea during a pair programming session and it's working as described! ✅

components/gitpod-cli/pkg/utils/logError.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/go-errors/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+func LogError(ctx context.Context, errToReport error, errorMessage string) {
+	log.WithError(errToReport).Error(errorMessage)
+
+	file, err := os.OpenFile(os.TempDir()+"/gitpod-cli-errors.log", os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0666)
+	if err == nil {
+		log.SetOutput(file)
```
akosyakov, line 29: I'm not sure we should do it cc @iQQBot could you recommend?

felladrin: ℹ️ We've followed this pattern from other packages' logs in gitpod repo:

image

iQQBot: I don't think we're going to let users submit logs from here ....

felladrin: But this file wasn't supposed to be submitted.

The errors are actually sent to `ide-metrics-api` when they happen, via a POST request.

https://github.com/gitpod-io/gitpod/blob/4344ff0517c1453a940b8fcb9db1df9ebf805565/components/gitpod-cli/pkg/utils/logError.go#L69-L75

In the last mob session, we just discussed those errors should be written to some file, but we didn't discuss where. Please add [here](https://github.com/gitpod-io/gitpod/issues/13847) a reminder to change the file location in the follow-up PR.

PR #13766, opened 2022-10-11
[alerts] Group by cluster, where relevant, ahead of centralizing rule evaluation

easyCZ:
## Description

To ensure we continue to receive the alerts with Cluster information, we've gotta group by cluster.

Right now, the cluster label does not exist on the leaf Prometheus instances (in monitoring-satellite) but it doesn't affect the metrics - it's aggregated away

## Related Issue(s)

Fixes https://github.com/gitpod-io/ops/issues/5598

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/operations/observability/mixins/meta/rules/server.yaml
+++ b/operations/observability/mixins/meta/rules/server.yaml
@@ -15,28 +15,28 @@ spec:
   - name: server
     rules:
     - alert: WebsocketConnectionsNotClosing
-      expr: sum(server_websocket_connection_count) == 10000
+      expr: sum(server_websocket_connection_count) by (cluster) == 10000
       for: 10m
       labels:
         severity: critical
       annotations:
         runbook_url: https://github.com/gitpod-io/runbooks/blob/main/runbooks/WebsocketConnectionsNotClosing.md
-        summary: Open websocket connections are not closing for the last 10 minutes and accumulating.
+        summary: Open websocket connections are not closing for the last 10 minutes and accumulating in {{ $labels.cluster }}.
         description: We have accumulated {{ printf "%.2f" $value }} open websocket connections.
 
     - alert: ServerEventLoopLagTooHigh
-      expr: avg_over_time(nodejs_eventloop_lag_seconds{job="server"}[20m]) > 0.35
+      expr: avg_over_time(nodejs_eventloop_lag_seconds{job="server"}[20m]) by (cluster) > 0.35
       for: 5m
       labels:
         severity: critical
       annotations:
         runbook_url: https://github.com/gitpod-io/runbooks/blob/main/runbooks/ServerEventLoopLagTooHigh.md
-        summary: Server accumulated too much "event loop lag". The webapp will become unresponsive if we don't act here.
+        summary: Server accumulated too much "event loop lag" on {{ $labels.cluster }}. The webapp will become unresponsive if we don't act here.
         description: Server has accumulated {{ printf "%.2f" $value }}s event loop lag.
 
     - alert: InstanceStartFailures
       # Reasoning: 1 failure every 120s should not trigger an incident: 1/120 = 0.00833.. => 0.01
-      expr: sum (irate(gitpod_server_instance_starts_failed_total[2m])) by (reason) > 0.01
+      expr: sum(irate(gitpod_server_instance_starts_failed_total[2m])) by (reason) > 0.01
       for: 30s
       labels:
         severity: critical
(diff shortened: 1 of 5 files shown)


operations/observability/mixins/meta/rules/server.yaml
```diff
@@ -87,20 +87,20 @@ spec:
     #     description: db-sync pod not running
 
     - alert: MessagebusNotRunning
-      expr: up{job="messagebus"} < 1
+      expr: sum(up{job="messagebus"}) by (cluster) < 1
```
laushinka, line 90: We use `sum` now because with the centralization, the metric can be aggregated across all Webapp clusters?

easyCZ: Yep, with the centralization there would be 2 things happening:
1. There would be multiple series for the raw metric - 1 for each cluster
2. Aggregations can only be applied to summary metrics, so we need to sum anyway to be able to group by cluster

laushinka (approved):
Approving and holding in case @ArthurSens has comments.
Feel free to unhold anytime, @easyCZ

/hold

ArthurSens (approved):
Looks perfect from my side!

@easyCZ, we discussed the problem with the namespace during the call today, right? I think I have a better approach!

In ArgoCD we specify the namespace where we want to deploy things ([example](https://github.com/gitpod-io/ops/blob/74edc45270661f3a64eb75e8cfcb50495f1ed995/deploy/io-dev/argocd/monitoring-satellite.yaml#L18)), we could just delete the namespace from the YAML files here, and our ArgoCD App definitions would take care of choosing which namespace to deploy to 🙂 

Do you mind updating this PR to the move the namespace as well?

operations/observability/mixins/meta/rules/server.yaml
```diff
@@ -15,28 +15,28 @@ spec:
...
       annotations:
         runbook_url: https://github.com/gitpod-io/runbooks/blob/main/runbooks/ServerEventLoopLagTooHigh.md
-        summary: Server accumulated too much "event loop lag". The webapp will become unresponsive if we don't act here.
+        summary: Server accumulated too much "event loop lag" on {{ $labels.cluster }}. The webapp will become unresponsive if we don't act here.
         description: Server has accumulated {{ printf "%.2f" $value }}s event loop lag.
 
     - alert: InstanceStartFailures
       # Reasoning: 1 failure every 120s should not trigger an incident: 1/120 = 0.00833.. => 0.01
-      expr: sum (irate(gitpod_server_instance_starts_failed_total[2m])) by (reason) > 0.01
+      expr: sum(irate(gitpod_server_instance_starts_failed_total[2m])) by (reason) > 0.01
```
geropl, line 39: :question: Why not make this alert per-cluster as well? Not strictly necessary, but from what I see we make _all_ alerts cluster-aware, and it would keep the mental model easier.

easyCZ: I was actually looking at what it's doing and comparing with actual values in and having the threshold lower (per both clusters, rather than just one) looked reasonable.

easyCZ: There are other alerts (in modified files) which also stay global without being cluster specific.

easyCZ:
@ArthurSens Thanks for that pointer. I'm guessing the intent is to drop the namespace from our alerts and keep the `monitoring-satellite` [here](https://github.com/gitpod-io/ops/blob/74edc45270661f3a64eb75e8cfcb50495f1ed995/deploy/io-dev/argocd/monitoring-satellite.yaml#L18). When we go to switch it to central, we upate the namespace there?

If that's the case, how do we ensure that for the transitionary period (when we run both rule eval in central and in leafs) it gets deployed to both central and leafs? Would we we add a second Application definition?

operations/observability/mixins/meta/rules/server.yaml
geropl: Ok, makes sense! :+1:

geropl (approved):
Just for the record: :heavy_check_mark:

ArthurSens:
> @ArthurSens Thanks for that pointer. I'm guessing the intent is to drop the namespace from our alerts and keep the `monitoring-satellite` [here](https://github.com/gitpod-io/ops/blob/74edc45270661f3a64eb75e8cfcb50495f1ed995/deploy/io-dev/argocd/monitoring-satellite.yaml#L18). When we go to switch it to central, we upate the namespace there?

Correct! Not only namespace actually, but we'll also need to change `destination` too (i.e. cluster where it is deployed)

> If that's the case, how do we ensure that for the transitionary period (when we run both rule eval in central and in leafs) it gets deployed to both central and leafs? Would we we add a second Application definition?

I was planning to create a whole new app for central verify that it works, then start removing the satellite apps

easyCZ:
@ArthurSens I've now also removed the namespace definitions. If this looks correct to you, please unhold the PR otherwise let me know where I went wrong, please :)

easyCZ:
@ArthurSens ping on this. Could you please have a look and let me know if this PR is as expected? If so, please unhold.

ArthurSens:
> @ArthurSens ping on this. Could you please have a look and let me know if this PR is as expected? If so, please unhold.

Sorry for the ghosting, we got hit by a few problems with ArgoCD and I got distracted. I'll get back to it this week!

easyCZ:
@ArthurSens Ping on this.

ArthurSens:
Alright, things are very slowly progressing, but we can merge this one without problems now 

/unhold

PR #13768, opened 2022-10-11
[prebuild] Support opening a specfic prebuild

jankeromnes:
_(Werft was having too much trouble with https://github.com/gitpod-io/gitpod/pull/13706 hence the new PR 👇)_

## Description
This PR enables users to open a specific prebuildm, and integrates this capability on the dashboard.

## How to test
1. Make sure there's a prebuild and a newer commit on the same branch
2. Go the prebuild view
3. Open the specific prebuild and observe it's the correct state

## Release Notes

```release-note
[prebuild] Add support for opening specific prebuilds
```

## Werft options:

- [x] /werft with-preview

Front logo [Front conversations](https://app.frontapp.com/open/top_3rxsa)

### DIFF
--- a/components/gitpod-protocol/src/protocol.ts
+++ b/components/gitpod-protocol/src/protocol.ts
@@ -1200,6 +1200,16 @@ export namespace AdditionalContentContext {
     }
 }
 
+export interface OpenPrebuildContext extends WorkspaceContext {
+    openPrebuildID: string;
+}
+
+export namespace OpenPrebuildContext {
+    export function is(ctx: any): ctx is OpenPrebuildContext {
+        return "openPrebuildID" in ctx;
+    }
+}
+
 export interface CommitContext extends WorkspaceContext, GitCheckoutInfo {
     /** @deprecated Moved to .repository.cloneUrl, left here for backwards-compatibility for old workspace contextes in the DB */
     cloneUrl?: string;
--- a/components/server/ee/src/workspace/workspace-factory.ts
+++ b/components/server/ee/src/workspace/workspace-factory.ts
@@ -21,6 +21,7 @@ import {
     PrebuiltWorkspace,
     WorkspaceConfig,
     WorkspaceImageSource,
+    OpenPrebuildContext,
 } from "@gitpod/gitpod-protocol";
 import { log } from "@gitpod/gitpod-protocol/lib/util/logging";
 import { LicenseEvaluator } from "@gitpod/licensor/lib";
@@ -369,6 +370,18 @@ export class WorkspaceFactoryEE extends WorkspaceFactory {
                 })
             ) {
                 config._featureFlags = (config._featureFlags || []).concat(["persistent_volume_claim"]);
+            }
+
+            if (OpenPrebuildContext.is(context.originalContext)) {
+                // Because of incremental prebuilds, createForContext will take over the original context.
+                // To ensure we get the right commit when forcing a prebuild, we force the context here.
+                context.originalContext = buildWorkspace.context;
+
+                if (CommitContext.is(context.originalContext)) {
+                    // We force the checkout of the revision rather than the ref/branch.
+                    // Otherwise we'd the correct prebuild with the "wrong" Git status.
+                    delete context.originalContext.ref;
+                }
             }
 
             const id = await this.generateWorkspaceID(context);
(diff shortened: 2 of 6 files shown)


csweichel:
Change works now as intended. I've also changed the wording on the button.

I am not particularly fond of the hoops I had to introduce to make this happen!

jankeromnes:
Awesome news, many thanks @csweichel! 🎉

The Werft build failed with:

```
  1) AccountServiceSpec
       "before each" hook for "multiCreditSessionAfter":
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/tmp/build/components-ee-payment-endpoint--dbtest.8bbb983371d02e98c8c1e2bb25cf11d71d9a6e35/src/accounting/account-service.spec.db.ts)
```

we could raise this timeout to 30s. But meanwhile:

/werft run

geropl:
> I am not particularly fond of the hoops I had to introduce to make the overall change happen!

@csweichel While I see that this is a minimal change to enable opening workspaces arbitrary prebuilds: Is this really something we want to enable using a ContextParser? That somehow feels like a hack to me (and might cause some of the "hoops" you were referring to).

I'm ok with this if we want to have something to move forward and test quickly. But I'd rather not ask customers to use these ContextUrls. :confused: I'd rather go with something akin Jan is looking into and call that from the dashboard in appropriate places. WDYT?

csweichel:
> call that from the dashboard in appropriate places

How would that work if not using a context parser?

csweichel:
> Is this really something we want to enable using a ContextParser? That somehow feels like a hack to me (and might cause some of the "hoops" you were referring to).

Ideally I'd like to have an actual proper context for opening "named" prebuilds. Unfortunately `ContextParser` are tied to SCM systems and are hard to use outside of that context. This causes hoop #1: using a prefix context parser.

The way we currently integrate prebuilds in the workspace creation and context handling is a bit of a "slight of hand" style thing with a few implicit assumptions/behaviours sprinkled in, which e.g. enable the incremental nature of things today. That's what necessitates hoop #2: the copying of the original context and removal of `ref`, because we ignore the `refType`. I didn't want to enforce the `refType` for fear of adverse side effects.

Testing all this stuff is near impossible. I tried to introduce a unit test for `findPrebuildWorkspace` but gave up going down the `inversify` rabbit hole 30 minutes in. There are few (no?) tests around workspace creation which makes introducing changes safely difficult. That's hoop #3: I was unable to tie this behaviour down, and only through stepping through the flow as I able to understand what's even happening.

How do we ensure that users actually get the prebuild they selected, if not through context? Relying on the commit matching alone feels too brittle.

geropl:
> Ideally I'd like to have an actual proper context for opening "named" prebuilds

I agree here if by "named" you mean precisely define which prebuild you want. :+1: 

> How do we ensure that users actually get the prebuild they selected, if not through context?

I thought about this as more of a parameter to "createWorkspace" which controls where to initialize content from, rather than a Context. Content depends on the Context, but they are not the same.

> Testing all this stuff is near impossible.

> There are few (no?) tests around workspace creation which makes introducing changes safely difficult.

Absolutely agree. That's one area where we never found time yet to improve upon.

components/server/ee/src/workspace/workspace-factory.ts
```diff
@@ -373,6 +374,18 @@ export class WorkspaceFactoryEE extends WorkspaceFactory {
                 config._featureFlags = (config._featureFlags || []).concat(["persistent_volume_claim"]);
             }
 
+            if (OpenPrebuildContext.is(context.originalContext)) {
+                // Because of incremental prebuilds, createForContext will take over the original context.
```
geropl, line 378: :+1:

components/gitpod-protocol/src/protocol.ts
```diff
@@ -1200,6 +1200,16 @@ export namespace AdditionalContentContext {
     }
 }
 
+export interface OpenPrebuildContext extends WorkspaceContext {
```
geropl, line 1203: While I'm not super-fond of adding a ContextParser for this I understand that currently, it's the only accessible way that make things like this work.
Also, removing it later should not be too hard. :slightly_smiling_face:

geropl:
rebased to make the build green... :crossed_fingers:

components/gitpod-protocol/src/protocol.ts
svenefftinge: Why don't we use the commit context?

jankeromnes: See related discussion: https://github.com/gitpod-io/gitpod/pull/13706#discussion_r991574427

PR #13790, opened 2022-10-12
supervisor: Ignore the terminated signal

utam0k:
## Description

Until now, the start-up time of a workspace is slower than the start-up of an IDE but reserved. Therefore, if you requested to exit immediately after starting the workspace such as [the integration test](https://gitpod.slack.com/archives/C03E52788SU/p1665448602379589), the IDE was sometimes not ready!

## Related Issue(s)

None

## How to test

Pass the integration test

## Release Notes

```release-note
Ignore a noisy error in the supervisor
```

## Werft options:

- [ ] /werft with-preview

### DIFF
--- a/components/supervisor/pkg/supervisor/supervisor.go
+++ b/components/supervisor/pkg/supervisor/supervisor.go
@@ -821,10 +821,8 @@ func launchIDE(cfg *Config, ideConfig *IDEConfig, cmd *exec.Cmd, ideStopped chan
 		}()
 
 		err = cmd.Wait()
-		if err != nil {
-			if errSignalTerminated.Error() != err.Error() {
-				log.WithField("ide", ide.String()).WithError(err).Warn("IDE was stopped")
-			}
+		if err != nil && err.Error() != errSignalTerminated.Error() {
+			log.WithField("ide", ide.String()).WithError(err).Warn("IDE was stopped")
 
 			ideWasReady, _ := ideReady.Get()
 			if !ideWasReady {


utam0k:
/werft run with-large-vm with-preview with-integration-tests=workspace

kylos101:
@iQQBot are there IDE integration tests that we could run against this, too?

jenting (approved):

utam0k:
> Why could this happen?
> 
> https://github.com/gitpod-io/gitpod/blob/7d066874214db6a75bacd4e2a17138c707d67203/test/pkg/integration/workspace.go#L216-L217
> 
> we waiting the workspace to be ready, and ready means ide is ready
> 
> https://github.com/gitpod-io/gitpod/blob/7d066874214db6a75bacd4e2a17138c707d67203/components/ws-manager/pkg/manager/probe.go#L41-L49

Perhaps this is due to the IDE starting later than the time it takes to complete workspace startup.

utam0k:
@iQQBot Is it impossible for this code to mark the IDE as Ready?
https://github.com/gitpod-io/gitpod/blob/7d066874214db6a75bacd4e2a17138c707d67203/components/supervisor/pkg/supervisor/supervisor.go#L818-L821

utam0k:
@iQQBot ws-manager should only return RUNNING under the following conditions. Is there anything you can think of?
 https://github.com/gitpod-io/gitpod/blob/36beceb536e9c5f9e5204020ec64b685b2658c67/components/ws-manager/pkg/manager/status.go#L485-L496

utam0k:
> Hi @utam0k I found the root cause, it is because
> 
>     1. `ws-manager` will follow redirect when it run ide ready probe
> 
>     2. integration test uses the wrong servicePrefix, so `ws-proxy` will redirect it to `dashboard` workspace not found page, and this page status code is 200... This way `ws-manager` will incorrectly assume that ide is ready
> 
>     3. `ws-manager` didn't check ide ready response
> 
> 
> and another small thing is `ws-manager` request ide ready probe too frequently (every 500ms, and didn't use `/wait/true` API)
> 
> I fix this in #13828 this PR

:eyes: Thanks @iQQBot

PR #13791, opened 2022-10-12
[usage] Return CostCenter from UsageService.SetCostCenter protos

easyCZ:
## Description

This allows to use the returned value and avoid having to do a follow-up GetCostCenter call

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

## Werft options:

- [ ] /werft with-preview


components/usage/pkg/db/cost_center_test.go
```diff
@@ -57,6 +57,8 @@ func TestCostCenterManager_GetOrCreateCostCenter(t *testing.T) {
 }
 
 func TestCostCenterManager_UpdateCostCenter(t *testing.T) {
+	t.Run("m")
```
jankeromnes, line 60: Curious: What does this line do?

easyCZ: thanks, removed.

jankeromnes (approved):
Thanks! Code change looks relatively safe to me, and the build passes. 👀

